PR #119, opened 2026-05-25
Async emit: MoveNext per-await dispatch (straight-line)

DavidObando:
Implements the **MoveNext per-await dispatch** slice. With PR #118 the kickoff body ran end-to-end but MoveNext was a stub that immediately called `builder.SetResult(default)`. After this PR, MoveNext actually executes the user body, suspends at every `await`, and resumes on the appropriate awaiter callback — for the **straight-line case** (no try/catch around await, no sub-expression spilling, no async iterators).

End-to-end programs that now compile and run:

```gsharp
async func doIt() {
    await Task.CompletedTask
    Console.WriteLine("after")
}

async func getVal() int {
    let x = await Task.FromResult(42)
    return x
}

async func sum() int {
    let a = await Task.FromResult(10)
    let b = await Task.FromResult(32)
    return a + b      // 42
}
```

## What's in the PR

**Commit `ef3b3bc` — MoveNext rewriter and emit plumbing**

- **`Lowering/Async/MoveNextBodyRewriter.cs`** (new, 566 lines) — transforms the lowered user body into a full MoveNext `BoundBlockStatement`:
  - State-dispatch switch at entry (jumps to `stateK_resume` or `stateStart`).
  - Per-await §6.4 sequence for each `await E`: `GetAwaiter`, fast-path `IsCompleted` check, hoist `<>1__state = K` + `cachedState = K`, store awaiter into pooled `<>u__N`, call `AwaitOnCompleted` or `AwaitUnsafeOnCompleted` (per `AwaitableShape.IsCriticalNotifyCompletion`), `goto exitLabel`. On resume: reload pooled awaiter (cast back if reference-typed pool), clear the slot, reset state, then `GetResult()`.
  - Funneled returns: `return X` becomes `<retVal> = X; goto exprReturnLabel`.
  - Hoisted-local rewrite via the existing `AsyncStateMachineFieldMap`.
  - Try/catch wrapper: catch sets state = -2 and calls `builder.SetException(ex)`; `exprReturnLabel` sets state = -2 and calls `SetResult([<retVal>])`.
- **`Binding/BoundStateMachineAwaitOnCompleted.cs`** (new) — marker node for the `AwaitOnCompleted`/`AwaitUnsafeOnCompleted` call. Needed because the call's second type argument is the synthesized SM TypeDef, which has no `System.Type` representation; the emitter builds the `MethodSpec` manually from the node.
- **`Emit/ReflectionMetadataEmitter.cs`** — replaces the stub MoveNext with the rewritten body emission, adds `EmitAwaitOnCompletedCall` (custom 2-arg generic `MethodSpec` with the SM TypeDef), and tightens `BodyEmitter` to use `ldarg 0` (not `ldarga 0`) for struct-`this` so MoveNext can read/write hoisted fields correctly. Labels are now placed inside try bodies to satisfy `leave` semantics.
- **`Lowering/Async/SynthesizedStateMachineType.cs`** + **`AsyncStateMachineTypeBuilder.cs`** — awaiter pool: one `<>u__N` field per distinct awaiter `TypeSymbol`; reference-typed awaiters collapse to a single `System.Object` field with a cast at resume.

**Commit `d9aa62e` — bug fixes uncovered by end-to-end tests + test coverage**

- **`Binding/BoundTreeRewriter.cs`** — adds the missing `RewriteStateMachineAwaitOnCompleted` dispatch (was throwing "Unexpected node" when the emitter's locals-collection rewriter walked MoveNext).
- **`Binding/BoundNodePrinter.cs`** — adds `WriteStateMachineAwaitOnCompleted` so debug printing doesn't throw.
- **`Emit/ReflectionMetadataEmitter.cs` (`CollectStatements`)** — now recurses into nested `BoundBlockStatement`s. The per-await sequence is wrapped in its own block, so without recursion the resume labels (`<>sm_await_resume_after_K`) were never registered and the goto site threw `KeyNotFoundException`.
- **`test/Core.Tests/CodeAnalysis/Emit/AsyncMoveNextEmitTests.cs`** (new) — three end-to-end tests covering `await Task.CompletedTask`, funneled `return await Task.FromResult(42)`, and two sequential awaits returning 42.

## What's still gated by `AsyncEmitPrecheck`

- `await` inside `try`/`catch`/`finally` — needs `AsyncExceptionHandlerRewriter`.
- `await` as a sub-expression with side-effecting siblings — needs `SpillSequenceSpiller`.
- `await using` / `await for` — needs lowering integration.
- `ref` locals across await — needs `RefInitializationHoister`.
- Async iterators (`async IAsyncEnumerable<T>`).
- Async lambdas.
- `Task.Yield()` (`YieldAwaitable`) — GSharp's awaiter resolver currently only accepts `Task`/`Task<T>`-shaped awaitables; this is a separate gap worth tracking.

Each gated case still fires the existing precheck diagnostic.

## Verification

- `dotnet build GSharp.sln -nologo -clp:NoSummary` — clean.
- `dotnet test GSharp.sln --no-restore --nologo` — **856 tests pass** (Sdk 21 · Core 726 · Compiler 84 · Interpreter 8 · LanguageServer 17). +3 from baseline.

This is the first PR where `async`/`await` GSharp programs actually execute the user body through a real state machine in emitted IL.


DavidObando:
Found the Linux-only crash: the MoveNext per-await sequence was emitting `ldnull; stfld <TaskAwaiter>` to default-clear the awaiter pool field. Storing a managed null reference into a value-type field is invalid IL — macOS arm64 JIT tolerates it, Linux x64 JIT raises `InvalidProgramException` and kills the test host with no further diagnostic.

Fixed in d6d6ff8 by skipping the clear for value-type awaiters. The Task referenced by the awaiter struct stays GC-rooted until the next await stores into the same field — acceptable for this slice; a follow-up will emit proper `ldflda; initobj T` to release the root deterministically.

All 856 tests pass locally (Debug + Release on macOS arm64). Pushed; awaiting CI.

PR #371, opened 2026-06-01
Add binder + diagnostic support for ref struct (by-ref-like) types (#367)

### LINKED ISSUE #367: Emit & bind support for ref struct (by-ref-like) types
Tracked as a blocker for ADR-0055 (string interpolation revamp), Tier 3 lowering to `System.Runtime.CompilerServices.DefaultInterpolatedStringHandler`, which is a `ref struct`.

## Problem

G# has no support for `ref struct` / by-ref-like (`IsByRefLike`) types in either the binder or the emitter. A grep across `src/Core` finds **no** `ref struct`, `IsByRefLike`, or `ByRefLike` handling.

Verified absent:
```
$ grep -rin "ref struct\|IsByRefLike\|ByRefLike" src/Core --include=*.cs
(no results)
```

Because `DefaultInterpolatedStringHandler` is a `ref struct`, the compiler cannot:
- declare a local of a by-ref-like type,
- enforce the by-ref-like safety rules (no boxing, no heap capture, no async/iterator hoisting, stack-only lifetime),
- emit such a local correctly.

## Scope / acceptance criteria

- [ ] Recognize CLR `IsByRefLike` types when imported (binder type model).
- [ ] Permit locals of a by-ref-like type and emit them.
- [ ] Reject illegal escapes (boxing, field of a non-ref-struct, capture in closures/async/iterators) with a clear diagnostic.
- [ ] (Optional, later) allow user-declared `ref struct` types.

## Why now

This unblocks the low-allocation interpolation handler tier in ADR-0055. Without it, interpolation with format/alignment can still ship via `String.Format` (ADR-0055 Tier 2), but never reaches the C#-parity low-alloc path.

Refs: ADR-0055 (`docs/adr/0055-string-interpolation-revamp.md`), #366

DavidObando:
## Summary

Adds binder + diagnostic support for CLR by-ref-like (`ref struct`) types, unblocking ADR-0055 Tier 3 (`DefaultInterpolatedStringHandler` lowering). Closes #367.

G# can now recognize imported `IsByRefLike` types (`System.Span[T]`, `System.ReadOnlySpan[T]`, `System.Runtime.CompilerServices.DefaultInterpolatedStringHandler`), permit them as locals (emit already produces a correct local signature), and enforce the CLR's stack-only safety rules.

## What's done

- **Detection** — `ClrTypeUtilities.IsByRefLike(Type)` and `TypeSymbol.IsByRefLike(TypeSymbol)` (unwraps nullable). `Type.IsByRefLike` is honored under `MetadataLoadContext`, so this works for reference assemblies.
- **Locals permitted** — a by-ref-like local binds and emits correctly (verified end-to-end with `ReadOnlySpan`).
- **Illegal escapes rejected** with new diagnostic **GS0219**:
  - boxing / converting to a reference type (`object`, interface, delegate base);
  - field of a non-ref-struct — instance, primary-constructor, and static fields;
  - capture by a closure;
  - local declared in an `async` function or an iterator (would be hoisted into the heap-allocated state machine);
  - use as a generic type argument (user-defined and imported generics).
- **Docs** — GS0219 documented in `docs/diagnostics.md` (next free code after the GS02xx series, which is used up to GS0218 in code).
- **Tests** — `RefStructTests` (10 unit tests: detection + each rejection + legal-local), plus a `RefStructSpan` conformance sample exercising legal `ReadOnlySpan` slicing/length end-to-end.

## Test results

Full `dotnet test GSharp.sln`: **all green** — Core 1425, Compiler 252 (incl. new conformance sample), LanguageServer 135, Interpreter 54, Sdk 24. 0 failures.

## Deferred (with justification)

- **User-declared `ref struct` types** — explicitly marked optional/lower-priority in the issue ("prioritize importing/consuming CLR ref structs"). Not implemented; consuming CLR ref structs is what unblocks ADR-0055.
- **Full two-level escape analysis** (Roslyn `ref-safe-to-escape` / `safe-to-escape`, `scoped`/`UnscopedRef`) — already deferred per ADR-0039 (Phase 8+). The realistic escape vectors a value-typed ref struct can take in G# today (boxing, fields, capture, async/iterator hoisting, generic args) are all covered; the remaining ref-lifetime cases require that deferred analysis.
- The generic `AppendFormatted<T>` overload-resolution ambiguity on `DefaultInterpolatedStringHandler` is a separate, pre-existing limitation (tracked under ADR-0055 emitter work), independent of by-ref-like support.


DavidObando:
## Follow-up: user-declared `ref struct` types (deferred item from #367)

Implemented the optional/deferred acceptance-criterion of #367 — *allow user-declared `ref struct` types* — on top of the existing imported-ref-struct support.

### What's new
- **Surface syntax**: a `ref` contextual modifier on a struct declaration, parsed like `data`/`inline`:
  ```gsharp
  type Window ref struct {
      Items ReadOnlySpan[int32]   // a ref struct may hold by-ref-like fields
      Label string
  }
  ```
  `StructDeclarationSyntax` exposes `RefModifier`/`IsRef`; `StructSymbol` exposes `IsRefStruct` (derived from the declaring syntax, so it survives generic construction).
- **Binder**: `TypeSymbol.IsByRefLike` now recognizes user-declared ref structs, so **every** existing GS0219 escape check (boxing, closure capture, async/iterator hoisting, generic type argument, field of a non-ref-struct) applies to them automatically. The one relaxation: a `ref struct` may itself hold by-ref-like fields.
- **Emit**: user ref structs are stamped with `System.Runtime.CompilerServices.IsByRefLikeAttribute` **plus the C# compiler's `[Obsolete]` guard marker**, on both top-level and nested struct TypeDefs. Verified via `MetadataReader` that both attributes land on the emitted TypeDef.
- **Bonus hardening**: a by-ref-like **top-level (global) variable** is now rejected with GS0219 instead of producing a runtime `TypeLoadException` ("A ByRef-like type cannot be used as the type for a static field"). This previously affected imported ref structs too, since top-level vars are emitted as static fields.
- **Docs**: `docs/diagnostics.md` GS0219 section updated.

### Tests
- Added `RefStructTests` cases for user ref structs (recognition, legal local, boxing, non-ref-struct field, ref-struct-in-ref-struct field allowed, closure capture, generic argument, top-level global).
- Added `samples/UserRefStruct.gs` + `.golden` conformance sample (declares and uses a user ref struct, including one embedding another ref struct field).
- **Full suite green**: Sdk 24, Interpreter 54, LanguageServer 135, Core 1434, Compiler 253 — **0 failures**.

### Known limitation (documented, out of scope)
A user `ref struct` that embeds an **imported generic** by-ref-like field (e.g. `ReadOnlySpan[T]`) binds correctly but currently **segfaults at runtime** due to a deeper emit gap in value-type field layout / defensive copies for generic by-ref-like fields. Embedding a **plain** ref struct field works end-to-end. The sample therefore demonstrates the ref-struct-in-ref-struct relaxation with a plain inner ref struct. Fixing generic span-field layout is a separate emitter effort beyond this issue's scope.

PR #373, opened 2026-06-01
Support user-declared ref struct (by-ref-like) types (#367 follow-up)

DavidObando:
Follow-up to #367 / PR #371 (already merged). PR #371 landed support for **consuming** CLR `ref struct` types; this PR adds the deferred piece: **user-declared** `ref struct` types. Rebased onto current `main` as a standalone change.

## What this adds
- **Syntax:** a `ref` contextual modifier on struct declarations — `type Name ref struct { ... }` — parsed like `data`/`inline` (no new `SyntaxKind`, so the coverage matrix is unaffected). Exposed via `StructDeclarationSyntax.RefModifier`/`IsRef` and `StructSymbol.IsRefStruct` (preserved across generic construction).
- **Binder:** `TypeSymbol.IsByRefLike` now recognizes user ref structs, so all existing GS0219 escape rules (boxing, closure capture, async/iterator hoisting, generic type argument, field of a non-ref-struct) apply automatically. A `ref struct` may itself hold by-ref-like fields.
- **Emit:** user ref structs are stamped with `IsByRefLikeAttribute` + the C# `[Obsolete]` guard marker on top-level and nested TypeDefs (verified via `MetadataReader`).
- **Hardening:** a by-ref-like top-level/global variable is now rejected with GS0219 instead of crashing at runtime with `TypeLoadException` (also covers imported ref structs).
- **Tests/docs:** new `RefStructTests`, a `UserRefStruct.gs` + `.golden` conformance sample, and `docs/diagnostics.md` updates.

## Tests
Full `dotnet test GSharp.sln` on top of current `main`: **0 failures** — Core 1434, Compiler 253, LanguageServer 135, Interpreter 54, Sdk 24 (1900 total).

## Documented limitation (out of scope)
A user `ref struct` embedding an **imported generic** by-ref-like field (e.g. `ReadOnlySpan[T]`) binds correctly but faults at runtime due to a deeper generic value-type field-layout emit gap. Embedding a **plain** ref struct field works end-to-end (the sample demonstrates this). Fixing generic span-field layout is a separate emitter effort.

Refs #367.


DavidObando:
The documented limitation (user `ref struct` embedding an imported generic by-ref-like field such as `ReadOnlySpan[T]` faults at runtime due to a generic value-type field-layout emit gap) is now tracked separately in #375.

PR #550, opened 2026-06-07
Allow G# classes to implement imported CLR interfaces (fixes #525)

### LINKED ISSUE #525: Binder: `type X class : IClrInterface { ... }` rejected with GS0157 even when the interface is imported
## Summary

A G# `type X class : ISomething { ... }` declaration that uses a CLR interface as the base type clause fails with `GS0157: Cannot find type ISomething. Are you missing an import?` — even when the namespace containing the interface is imported and the interface is reachable for other purposes (e.g. as the type of a variable).

This blocks all test-doubling patterns: fakes, stubs, spies, and any G# type that needs to expose itself to C# code that consumes an interface.

## Repro

Sibling C#:
```csharp
namespace Probe.CSharp;
public interface IGreeter { string Greet(string name); }
```

G#:
```gsharp
package Probe

import Probe.CSharp

// This declaration fails with GS0157, even though `IGreeter` is imported and visible elsewhere.
type MyGreeter class : IGreeter {
    func Greet(name string) string {
        return "hi " + name
    }
}
```

## Observed

```
Probe.gs(7,24,7,32): error GS0157: Cannot find type IGreeter. Are you missing an import?
```

Note: as a type in expressions (`var g IGreeter = ...`) the binder *finds* `IGreeter` (it issues GS0155 about subtyping — separate issue). So the binder knows about `IGreeter`; it just refuses to accept it in a base-type clause.

## Expected

A G# class should be allowed to declare it implements any reachable CLR interface and provide member implementations for each method/property/event. gsc should emit `.implements` metadata so the resulting class is a fully-fledged CLR implementer of the interface.

## Context

`gsc` version: `Gsharp.NET.Sdk/0.1.431`, .NET 10 RTM.

Discovered while porting `tests/Oahu.Cli.Tests/App/CoreAuthServiceTests.cs::CallbackBridgeTests` to G# (the tests construct a fake `IAuthCallbackBroker` implementation). Multiple sub-agents hit the same block.

DavidObando:
## Summary

Fixes #525.

`type X class : IClrInterface { ... }` previously failed binding with
`GS0157: Cannot find type IClrInterface` even when the interface was
reachable through an `import`. The base-type clause resolver only looked
for a CLR *class* (and explicitly rejected interfaces), while the same
identifier in expression-typed contexts (`var g IClrInterface = ...`) was
resolved through `LookupType` and worked. This PR teaches the base-clause
path to find imported CLR interfaces, stores them on the class symbol,
emits `InterfaceImpl` metadata, and adds the implicit upcast from a G#
class to a CLR interface.

## Root cause

`Binder.BindStructDeclarationBody`'s base-type loop only resolved CLR
base classes via `TryResolveImportedBaseType`. Identifiers that didn't
resolve to a G# `InterfaceSymbol` and weren't a CLR base class fell
through to the "cannot find type" diagnostic. CLR interfaces never had
a resolution path on the base-clause side.

## Fix surface

- **Binder** (`src/Core/CodeAnalysis/Binding/Binder.cs`)
  - New `TryResolveImportedInterface` honors imports/aliases via
    `LookupType` and falls back to the reference resolver. Generic open
    definitions are rejected with a `TODO(issue-525)` (see *Scope notes*).
  - Base-clause loop now collects CLR interfaces and stores them on the
    `StructSymbol`.
  - New `VerifyClrInterfaceImplementations` walks every public abstract
    member on each imported interface and reuses GS0187 for missing
    implementations.
  - CLR signature comparison uses `ClrTypeUtilities.AreSame` so types
    loaded via `MetadataLoadContext` (gsc's reference resolver) compare
    correctly against host runtime types.
- **Symbols** (`src/Core/CodeAnalysis/Symbols/StructSymbol.cs`)
  - New `ImplementedClrInterfaces` collection + `SetImplementedClrInterfaces`
    setter so emit can find the resolved interface types.
- **Conversion** (`src/Core/CodeAnalysis/Binding/Conversion.cs`)
  - New branch in `Classify` for G# class -> CLR interface upcasts so
    interface-typed locals/fields/return slots accept a G# class instance.
- **Emit** (`src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs`)
  - `AddInterfaceImplementation` rows emitted for every entry in
    `ImplementedClrInterfaces`.
  - `MethodImplicitlyImplementsInterface` /
    `PropertyImplicitlyImplementsInterface` extended to recognize CLR
    interface members so accessors get `Virtual|NewSlot` and value-type
    contexts pick the right method attributes.
  - `IsReferenceCompatible` mirrors the new binder upcast for
    cross-context type comparisons during IL emission.

## Tests

Seven xUnit tests in
`test/Compiler.Tests/Emit/Issue525ClassImplementsClrInterfaceEmitTests.cs`:

1. `GSharpClass_ImplementsBclInterface_RunsAndIlVerifies` — G# class
   implements `System.IDisposable`, dispatch through an interface-typed
   local runs the G# body. `ilverify` is run on the output.
2. `GSharpClass_MetadataDeclaresInterfaceImpl` — reflects the emitted
   assembly via `MetadataLoadContext` and asserts the TypeDef carries
   `System.IDisposable` and that `Dispose()` is `IsVirtual`.
3. `GSharpClass_ImplementsMultipleClrInterfaces_BothReachable` — two
   interfaces (`IDisposable`, `ICloneable`) reachable through separate
   interface-typed locals.
4. `GSharpClass_ImplementsClrInterfaceWithReadOnlyProperty_DispatchesGetter`
   — sibling C# library exposing a read-only property interface; G#
   class implements it via `prop Label string { get { ... } }`.
5. `GSharpClass_ImplementsSiblingCSharpInterface_IssueRepro` — exact
   shape from the issue body (sibling C# `IGreeter`).
6. `ClrInterface_FromNonImportedNamespace_ReportsGS0157` — unknown
   identifier in the base clause still surfaces GS0157.
7. `GSharpClassHierarchy_StillCompiles_RegressionGuard` — regression
   guard for an existing G#-only class hierarchy.

Full suite (`dotnet test GSharp.sln --no-restore --nologo`) passes
locally: **Core 1988, Compiler 526, LanguageServer 226, Interpreter 72,
Sdk 24 — total 2836 passed, 0 failed**.

## Scope notes

**Covered**

- Non-generic CLR interfaces (single and multiple) in the base-type clause.
- Method members (with signature matching across reflection contexts).
- Read-only property members (matched by name + property type).
- Imports/aliases via `LookupType` and fully-qualified fallback via
  `ReferenceResolver.TryResolveType`.
- `InterfaceImpl` rows + virtual-slot promotion for accessors.
- Class -> CLR-interface implicit conversion (binder + emitter).
- Missing-member diagnostic (reuses GS0187).
- `ilverify` runs on every emitted assembly in the new tests.

**Deferred (with explicit `TODO(issue-525)`)**

- Generic CLR interfaces (e.g. `IComparable[T]`) in the base-type
  clause. The current base-type grammar is a single identifier; the
  closed form needs a type-argument list. Resolution intentionally
  rejects open generic definitions today. Worth a follow-up issue if
  you want generic-interface support.
- Property setters from CLR interfaces are validated but the G#
  `prop` accessor surface for emitting an interface-required setter
  wasn't exercised in the new tests (the only sibling-C# property test
  is read-only). Should work end-to-end via the existing accessor
  emission, but if you find a gap I can extend.

## Verification

- `dotnet build GSharp.sln -nologo -clp:NoSummary` — 0 warnings, 0 errors.
- `dotnet test GSharp.sln --no-restore --nologo` — 2836 passed, 0 failed.
- `dotnet tool restore` — `dotnet-ilverify` 10.0.8 available; called from
  every new emit-and-run test.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/Binder.cs
+++ b/src/Core/CodeAnalysis/Binding/Binder.cs
@@ -1323,6 +1323,7 @@ public sealed class Binder
         StructSymbol baseClassSymbol = null;
         TypeSymbol importedBaseType = null;
         var implementedInterfaces = ImmutableArray.CreateBuilder<InterfaceSymbol>();
+        var implementedClrInterfaces = ImmutableArray.CreateBuilder<TypeSymbol>();
         if (syntax.HasBaseType)
         {
             if (!syntax.IsClass)
@@ -1359,12 +1360,23 @@ public sealed class Binder
                         // user-declared GSharp type, so consult the same imported
                         // type resolution used for construction / static access.
                         // Only the first identifier (the base-class slot) may be
-                        // a CLR class; the rest are interfaces (CLR interface
-                        // implementation is out of scope here).
+                        // a CLR class.
                         if (i == 0 && !hasAttributeSugar
                             && TryResolveImportedBaseType(baseName, out var importedBase))
                         {
                             importedBaseType = importedBase;
+                            continue;
+                        }
+
+                        // Issue #525: any non-first identifier (and the first
+                        // when no CLR base class matched) may be an imported
+                        // CLR interface. Resolve through the same import-aware
+                        // lookup used for expression-type contexts so a `class
+                        // : IClrInterface { ... }` form works against any
+                        // reachable CLR interface (e.g. System.IDisposable).
+                        if (TryResolveImportedInterface(baseName, out var importedIface))
+                        {
+                            implementedClrInterfaces.Add(importedIface);
                             continue;
                         }
 
@@ -2491,6 +2503,19 @@ public sealed class Binder
             pendingInterfaceImplementationChecks.Add((syntax, structSymbol));
         }
 
+        // Issue #525: record imported CLR interfaces from the base-type
+        // clause and queue the same deferred verification used for G#
+        // interfaces. The check is deferred because class methods may not
+        // have been bound yet when this declaration is processed.
+        if (implementedClrInterfaces.Count > 0)
+        {
+            structSymbol.SetImplementedClrInterfaces(implementedClrInterfaces.ToImmutable());
+            if (implementedInterfaces.Count == 0)
+            {
+                pendingInterfaceImplementationChecks.Add((syntax, structSymbol));
+            }
+        }
+
         // Issue #306: bind standalone user-defined constructors (`init(...)`).
         BindConstructorDeclarations(syntax, structSymbol, package, baseClassSymbol, importedBaseType);
     }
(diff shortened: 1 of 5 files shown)


DavidObando:
@copilot please create a follow-up issue for generic base types in G# as mentioned above in the deferred work comment.

Issue #775, opened 2026-06-12
G# spelling for class/struct/new() generic constraints


DavidObando:
**Discovered while closing #751** (parser fix for L2 from ADR-0084).

G# generic type-parameter constraints today admit `any`, `comparable`,
and sealed-interface bounds. ADR-0088 / #750 added constraint-aware
*overload resolution* by reading C#-attached `class` / `struct` /
`new()` constraints on imported types, but G# itself has no spelling
for those constraints on a G#-authored type parameter.

Reference: https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/generics/constraints-on-type-parameters

This blocks porting `Gsharp.Extensions.Optional` to G#: the
reference- and value-typed `Optional` helpers share the single name
`Map` / `OrElse` / `OrCompute` etc., and the C# version disambiguates
purely with `where T : class` vs `where T : struct`. Without G#
spellings the port would need two distinct names, regressing the
public surface ADR-0084 settled on.

### Proposed G# spelling

Two natural options:

1. Lean on the existing G# constraint-slot syntax:
   `func Map[T class, U class](self T?, f (T) -> U) U?`
2. Borrow the C# `where` clause after the parameter list (also
   composes naturally with reified-generic emit per ADR-0087):
   `func Map[T, U](self T?, f (T) -> U) U? where T : class, U : class`

Option 1 feels more idiomatic to G#. Either spelling needs binder + emit support so the emitted IL carries
the right `GenericParam` constraint flags.

### Why it matters

Closes the third (and last known) language gap blocking the dogfooded
G# rewrite of the Extensions stdlib. Together with #773 and #774, this
retires the C# escape hatch under `src/Sdk/Gsharp.Extensions/`.

### Related

- #697 Consider constraints in generics
- #751 / ADR-0084 §L2 (parser-side closed)
- #750 / ADR-0088 (constraint-aware overload resolution — only reads
  imported constraints today)
- #773 (generic-receiver dispatch)
- #774 (open-receiver iteration types)
- #724 / ADR-0084

Issue #724, opened 2026-06-11
Introduce `Gsharp.Extensions.*` idiomatic helper namespace (Optional, Sequences)


DavidObando:
## Summary

Introduce `Gsharp.Extensions.Optional` and `Gsharp.Extensions.Sequences` as the first two G# idiomatic helper namespaces, plus the underlying packaging story: the SDK ships a real G# assembly (`Gsharp.Extensions`) authored in G# itself, dogfooding the language for its own stdlib. The `Gsharp.Extensions.Go` namespace (#722, #723) lives in the same physical assembly.

## Motivation

Per §5.12 of the plan (Option B), G# ships a thin idiomatic helper layer on top of the .NET BCL. Goal is not to re-implement BCL — it is to:

1. Reshape BCL APIs that don't compose with G#'s type system (e.g. `FirstOrDefault` returns `default(T)`, which is the wrong contract under non-null-by-default).
2. Add common iterator / windowing / chunking utilities that .NET either lacks or hides behind framework-version gates.
3. Give samples and downstream code a uniform "G#-shaped" feel without forcing users to remember which BCL static class lives where.
4. Prove that the language can author its own stdlib — anything the stdlib can't express comfortably is a language gap worth filing.

## Packaging

- One physical assembly: **`Gsharp.Extensions.dll`**, bundled with `Gsharp.NET.Sdk` so it's available to every G# project automatically (no extra NuGet reference).
- Authored in G# (`.gs` sources under `src/Extensions/` or equivalent), dogfooding the language.
- Namespaces under it:
  - `Gsharp.Extensions.Optional` — see surface below.
  - `Gsharp.Extensions.Sequences` — see surface below.
  - `Gsharp.Extensions.Go` — provides the gated Go-flavored surface (#722, #723).
  - (Future) `Gsharp.Extensions.Strings`, `Gsharp.Extensions.Tasks`, etc. — add as needs arise.
- Imports are explicit: nothing under `Gsharp.Extensions.*` is auto-imported. The `/noimplicitimports` flag does not affect `Gsharp.Extensions.*` (it's already opt-in).

## What's already in the language (so we don't duplicate)

- **Nullable types** (`T?`) with `?.`, `?:`, `!!`, smart casts, and `nil`. We do NOT add a new `Option<T>` discriminated-union type — nullable references already cover absent values.
[...]

Issue #706, opened 2026-06-11
G# Language — Current State and Design Opportunities


DavidObando:
# G# Language — Current State and Design Opportunities

_Prepared: 2026-06-11. Decisions recorded later same day (see §5/§6/§8). Based on `design/Gsharp-design-v0.2.md`, ADRs 0001–0069, `website/docs/ref/{spec,feature-matrix}.md`, and the `samples/` corpus._

## 1. Executive summary

G# started as "a mini Go for .NET" and has, over the v0.2 phase, absorbed influence from Kotlin (null model, smart casts, primary constructors, string interpolation, `data struct`), Swift (`let`/`var`, structured concurrency, trailing closures, `deinit`), and C# (CLR exceptions, async/await, properties, events, attributes, generics, `ref`/`out`/`in`, spans). The result is an unusually feature-complete language for its age, but the surface bears the seams of its mixed heritage. The most distinctive — and most debatable — choices today are:

- **Syntax is Go-flavored** (`func`, `[]T`, `map[K]V`, `chan T`, `package`, `import`, postfix types, `:=`) on top of a **CLR-flavored type system** (exceptions, generics, properties, events, delegates, spans, ref/out/in, attributes).
- **Three concurrency models in one language**: Go (`go`, `chan`, `select`), Kotlin/Swift (`scope` structured concurrency), and .NET (`async`/`await`, `Task`).
- **Four binding keywords**: `var`, `let`, `const`, `:=` — more than any of the comparison languages.
- **Two type declaration styles**: Go-style `type Name struct {...}` keyword tail, plus Kotlin-style primary-constructor parameter lists, plus a Go-style receiver method syntax living alongside member methods.
- **Width-bearing primitives only** (`int32`/`uint64`/…). No `int`, `long`, `byte` aliases. Unusual among all four reference languages.
- **No `while`, no C-style `for(;;)`-only spelling** — `for cond { }` and `for init; cond; post { }` cover both.
- **`func` is everywhere** — declarations, function types, function literals, lambda forms (no arrow lambdas yet; `->` is reserved for switch-expression arms).

The language is in a sweet spot to make consolidating decisions before its surface ossifies. Sections 4–6 enumerate concrete opportunities.

## 2. Compiler/runtime posture (context for the comparison)

- Targets the CLR (net8.0 / net10.0) via direct PE+PDB emission. Ships an MSBuild SDK and a `dotnet new` template.
[...]

PR #793, opened 2026-06-13
Implement #792: SDK bootstrap to dogfood Gsharp.Extensions in G#

### LINKED ISSUE #792: SDK bootstrap: break Gsharp.NET.Sdk and Gsharp.Extensions cycle to dogfood Extensions in G#
## Summary

Break the `Gsharp.NET.Sdk` ↔ `Gsharp.Extensions` bootstrap cycle so the Extensions stdlib can be dogfooded as G#.

## Context

ADR-0084 set the long-term goal of authoring `Gsharp.Extensions.Optional` / `Gsharp.Extensions.Sequences` in G# (rather than C#). Across this Oats sweep we closed every language gap that blocked the port:

- #750 / ADR-0088 — constraint-aware overload resolution (CLR-side).
- #751 — parser accepts `(self T?)` / `(self sequence[T])` receivers.
- #773 — binder dispatches generic-receiver extensions at call sites.
- #774 — `for v in self` on `IEnumerable[T]` infers `v` as `T`.
- #775 / ADR-0097 — G# spelling for `class` / `struct` / `new()` constraints.

After #775, the language can express every signature ADR-0084 specified. The port still does not land because of a packaging cycle:

- `Gsharp.NET.Sdk` packages `Gsharp.Extensions.dll` as a bundled assembly that every consumer `.gsproj` automatically references.
- If we rebuild `Gsharp.Extensions` as a `.gsproj`, that project itself needs `Gsharp.NET.Sdk` to compile.

ADR-0097 documents the staged bootstrap recipe needed to break this:
1. Stand up a build-from-bin `Gsharp.NET.Sdk.Bootstrap` props/targets that points at the in-tree `gsc.dll` and Core.dll without consuming the `Gsharp.Extensions` nupkg.
2. Build `Gsharp.Extensions.dll` using the bootstrap SDK.
3. Repack `Gsharp.NET.Sdk` to reference the freshly built `Gsharp.Extensions.dll`.
4. Verify the round trip — the final SDK plus the final Extensions assembly produces identical IL to the C# starting point.

## Scope

- Stand up `Gsharp.NET.Sdk.Bootstrap` (or equivalent — pick the cleanest name).
- Re-author `Gsharp.Extensions.Optional` and `Gsharp.Extensions.Sequences` in `.gs`.
- Wire the build so `dotnet build GSharp.sln` builds the bootstrap SDK, then the G#-authored Extensions, then the consumer SDK.
- All 107+ tests in `test/Extensions.Tests/` must pass against the G# port.
- Delete the C# implementation files once parity is proven.

## Acceptance criteria

- `src/Sdk/Gsharp.Extensions/` contains `.gs` source for Optional and Sequences (Go subspace stays where it is until #722's gating is reconsidered).
- The CI build is unchanged from a consumer perspective.
- ADR-0084's "Known limitations" can be marked fully closed.

## Related

- Parent: #706
- ADRs: ADR-0084 (Extensions stdlib), ADR-0097 (constraints spelling), ADR-0087 (reified generics)
- Predecessors: #724, #750, #751, #773, #774, #775

DavidObando:
Closes #792. Related: #775 (ADR-0097), #724 (ADR-0084), parent #706.

## Summary

This PR delivers the **infrastructure side** of #792 — every compiler and
build-system prerequisite needed to author `Gsharp.Extensions` in G# —
and explicitly defers the source-side port of `Optional` /
`Sequences` because authoring them surfaced a fresh batch of G#
language gaps that warrant focused follow-ups rather than rushed
workarounds.

After this PR, replacing the C# `Optional` / `Sequences` with a
`Gsharp.Extensions.gsproj` is a flag flip — no more SDK or emit
changes are required for that step.

## Scope checklist

- [x] **Compiler — `[ExtensionAttribute]` emit.** G# extension
      methods (`func (self T) Name(...) ...`) now stamp
      `[System.Runtime.CompilerServices.ExtensionAttribute]` on both
      the MethodDef row and the `<Program>` TypeDef row that hosts
      them, so the C# / F# call-site lookup (ECMA-334 §13.6.9) treats
      them as extension methods. Without this stamp the existing C#
      test surface against `Gsharp.Extensions.dll` would not bind
      against a G#-built replacement.
      (`src/Core/CodeAnalysis/Emit/WellKnownReferences.cs`,
      `src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs`)
- [x] **Binder — `throw` is a CFG terminator.** The
      `AllPathsReturn` check that powers GS0100 now accepts
      `ThrowStatement` as a terminator (it already does so for the
      basic-block CFG construction). Trivial spellings like
      `if x == nil { throw ... } return x` now bind, which the
      `OrThrow` family of helpers needs.
      (`src/Core/CodeAnalysis/Binding/ControlFlowGraph.cs`)
- [x] **SDK — `Gsharp.NET.Sdk.Bootstrap`.** New build-time-only
      mirror of `Gsharp.NET.Sdk` that compiles `.gs` against the
      in-tree `gsc.dll` + BuildTask WITHOUT auto-referencing
      `Gsharp.Extensions.dll`. That auto-reference is the cycle:
      without breaking it, a G#-authored `Gsharp.Extensions.gsproj`
      would self-reference. The bootstrap is *equivalent* to the
      consumer SDK on every other axis (same BuildTask, same
      CoreCompile shape, same gsc invocation) so a future
      `Gsharp.Extensions.gsproj` swaps in cleanly.
      (`src/Sdk/Gsharp.NET.Sdk.Bootstrap/`)
- [x] **Tests — `[Extension]` emit + throw-CFG + bootstrap layout.**
      13 new tests:
      `Issue792ExtensionAttributeEmitTests` (4),
      `Issue792ThrowAsTerminatorTests` (3),
      `BootstrapSdkLayoutTests` (6).
- [x] **Baselines.** `samples/ExtensionFunctions.gs` and
      `samples/GenericExtensionFunctions.gs` baselines refreshed to
      reflect the new `[Extension]` CA rows.
- [x] **ADRs.** ADR-0084 §L5 (new) describes the bootstrap and lists
      the deferred language-side follow-ups. ADR-0097 §Follow-ups is
      updated to reflect that the build-system side of the bootstrap
      cycle is now closed.
- [ ] **`Optional` / `Sequences` ported to G#.** **Deferred** —
      authoring them hit several smaller, previously undocumented G#
      language gaps. Each is logged in ADR-0084 §L5 for follow-up.
      The C# escape hatch stays in place; the 107-test Extensions
      suite continues to run unchanged.

## Verification

| Gate | Result |
| --- | --- |
| `dotnet build GSharp.sln` | green |
| `dotnet test GSharp.sln --no-restore --nologo` | green |
| `ilverify` on freshly emitted `Gsharp.Extensions.dll` | clean |
| `cd website && npm ci && npm run build` | green |

Test pass counts: 107 Extensions, 243 Interpreter, 257 LanguageServer,
2991 Core (+1 skipped manual baseline regen), 1223 Compiler, 32
Sdk.Tests.

## Diff shape

- **0** C# files removed.
- **0** `.gs` files added (the source port itself is deferred).
- **+2** ADR sections updated (ADR-0084 §L5 added; ADR-0097
  §Follow-ups refreshed).
- **+1** Bootstrap SDK directory with a single `.targets` file +
  README.
- **+3** test files (13 tests total).
- **+2** baseline-JSON hashes refreshed for the two extension-method
  samples.

## Why the source port was deferred

Authoring `Optional.gs` and `Sequences.gs` hit these G# gaps in
quick succession:

- Generic-method type-parameter `T` does not thread through a generic
  instance-method call (`List[T]().ToArray()` returns `object[]`
  inside a generic shared method).
- No `default(T)` expression — `default` is currently a `switch`
  keyword only.
- No `params T[]` parameter declaration — callers must materialise
  the array explicitly, which would break the existing C# call sites.
- `==` is not defined between a function-typed value (`(T) -> T`)
  or a `sequence[T]` value and `nil`, so nil-guards inside
  extension bodies don't bind.
- `[MethodImpl(...)]` annotations don't parse inside `shared { }`
  blocks — the parser treats `@` as a field decl. AggressiveInlining
  on the C# hot paths would be silently lost.
- `yield` inside a shared-static method returning `IEnumerable[T]`
  reports GS0136 — the binder's iterator-detection check on the
  shared-method path doesn't recognise the return type.

Each warrants its own scoped fix; none is hard, but rolling them into
this PR would either bloat the diff or push the port into a
compromised shape that doesn't preserve the public API. The bootstrap
+ `[Extension]` emit shipped here means the port itself is a flag
flip once those gaps close.

## Commits

1. `feat(emit,binder): emit [ExtensionAttribute] and accept throw as a CFG terminator (#792)`
2. `feat(sdk): add Gsharp.NET.Sdk.Bootstrap to break the SDK ↔ Extensions cycle (#792)`


DavidObando:
Filed follow-ups for the deferred source-side port:

- #794 — Generic-method T does not thread through generic instance-method calls inside a generic shared method
- #795 — Add G# `default(T)` / `default` expression for unconstrained type parameters
- #796 — Allow `==` / `!=` between function-typed values (and sequence values) and `nil`
- #797 — Parse `@Attr(...)` annotations on methods declared inside `shared { }` blocks
- #798 — Allow `yield` inside a shared-static method that returns `IEnumerable[T]`
- #799 — Add `params T[]` parameter declarations to G#

Once these close, porting `Optional.gs` + `Sequences.gs` against the bootstrap landed here becomes mechanical.

PR #907, opened 2026-06-21
Fix LSP hover resolving extension call to same-named static method (#906)

### LINKED ISSUE #906: LSP hover resolves extension-method call to unrelated same-named static method
## Summary

Hovering an extension-method invocation in the G# language server can resolve to an unrelated, same-named **static** method that appears elsewhere in the program.

## Repro

In `Oahu/tests/Oahu.Cli.Tests/DoctorServiceTests.gs`, hovering `Single` in:

```
report.Checks.Single((c) -> ...)
```

shows:

```
func (Xunit.Assert) Single(collection ...) ...
```

but should show `func (System.Linq.Enumerable) Single...`.

## Root cause

Bound imported-call nodes are constructed with `syntax: null` in the binder, so `SemanticLookup.TryResolveInvokedImportedMethod` scans the **whole program** for the first bound call matching method name + arity, returning the first hit regardless of which call the hovered token belongs to. When an unrelated static method (`Xunit.Assert.Single`) shares the name + a compatible arity, it can win over the LINQ extension (`System.Linq.Enumerable.Single`).

## Fix

LSP-only: gate the same-name candidate scan by the hovered call's receiver dispatch shape (value receiver -> instance/extension method; type-name receiver -> plain static), with a CLR receiver-type tie-breaker for same-shaped collisions. No binder/PDB changes.

DavidObando:
Fixes #906.

## Problem

Hovering an extension-method invocation in the G# language server could resolve to an unrelated, same-named **static** method appearing elsewhere in the program. In `Oahu/tests/Oahu.Cli.Tests/DoctorServiceTests.gs`, hovering `Single` in `report.Checks.Single((c) -> ...)` showed `func (Xunit.Assert) Single...` instead of the expected `func (System.Linq.Enumerable) Single...`.

## Root cause

Bound imported-call nodes are constructed with `syntax: null` in the binder, so `SemanticLookup.TryResolveInvokedImportedMethod` scanned the **whole program** for the first bound call matching method name + arity, returning the first hit regardless of which call the hovered token belonged to. An unrelated static method (`Xunit.Assert.Single`) sharing the name and a compatible arity could win over the LINQ extension (`System.Linq.Enumerable.Single`).

## Fix (LSP-only — no binder/PDB changes)

- **`SemanticLookup.cs`**: Added a `CallReceiverGate` enum (`None`/`Value`/`TypeName`) and gated the candidate scan by the hovered call's receiver dispatch shape — a value receiver (`x.M(...)`) can only bind to an instance/extension method; a type-name receiver (`T.M(...)`) only to a plain static. Added an `IsReceiverTypeCompatible` tie-breaker comparing a candidate's effective receiver type against the hovered receiver's CLR type to disambiguate same-shaped collisions.
- **`HoverComputer.cs`**: Added `ClassifyCallReceiver` (and helpers) to derive the gate + receiver CLR type from the hovered call's receiver syntax (value symbols, primitive keywords, imported types, dotted type names), passed to the resolver.
- **`HoverHandlerTests.cs`**: Added a bidirectional regression test (`ComputeHover_ExtensionMethodInvocation_DoesNotResolveToSameNamedStaticMethod`) covering both the extension direction (`a.Concat(b)` → `Enumerable`) and the static direction (`String.Concat(...)` → not `Enumerable`).

## Verification

- Full solution builds with 0 warnings / 0 errors.
- All 349 `LanguageServer.Tests` pass.

The user's case classifies as a value receiver, so the static `Assert.Single` is rejected and the LINQ extension is selected.

### DIFF
--- a/src/LanguageServer/HoverComputer.cs
+++ b/src/LanguageServer/HoverComputer.cs
@@ -173,13 +173,20 @@ public static class HoverComputer
         // on the receiver type would miss.
         var invokedCall = FindNodes<CallExpressionSyntax>(tree.Root)
             .FirstOrDefault(call => MatchesToken(call.Identifier, token));
-        if (invokedCall != null
-            && SemanticLookup.TryResolveInvokedImportedMethod(compilation, token.Text, invokedCall.Arguments.Count, out var invokedMethod, out var invokedOverloadCount))
+        if (invokedCall != null)
         {
-            return new HoverModel(
-                SymbolDisplay.ToDisplayString(invokedMethod, SymbolDisplayFormat.Hover),
-                HoverDocumentationRenderer.Render(GSharp.Core.CodeAnalysis.Documentation.AssemblyDocumentationProvider.Resolve(invokedMethod)),
-                invokedOverloadCount);
+            // Disambiguate the program-wide same-name scan by the hovered call's
+            // receiver syntax: a value receiver (`x.M(...)`) can only bind to an
+            // instance/extension method, a type-name receiver (`T.M(...)`) only to
+            // a plain static method (issue #906).
+            var gate = ClassifyCallReceiver(tree, compilation, token, out var receiverClrType);
+            if (SemanticLookup.TryResolveInvokedImportedMethod(compilation, token.Text, invokedCall.Arguments.Count, gate, receiverClrType, out var invokedMethod, out var invokedOverloadCount))
+            {
+                return new HoverModel(
+                    SymbolDisplay.ToDisplayString(invokedMethod, SymbolDisplayFormat.Hover),
+                    HoverDocumentationRenderer.Render(GSharp.Core.CodeAnalysis.Documentation.AssemblyDocumentationProvider.Resolve(invokedMethod)),
+                    invokedOverloadCount);
+            }
         }
 
         var clrType = SemanticLookup.ResolveImportedClrType(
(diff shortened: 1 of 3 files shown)


DavidObando:
## Follow-up: fixed the real dotted-chain receiver shape

The first commit fixed same-named static collisions for **simple-name** receivers (`a.Concat(b)`), but @DavidObando could still reproduce the original bug on `report.Checks.Single(...)`.

**Root cause of the miss:** a dotted chain like `report.Checks.Single(pred)` parses *right-associatively*, so the hovered call's enclosing accessors yield several candidate receivers — the bare trailing segment `Checks` **and** the full `report.Checks`. `ClassifyCallReceiver` was taking the *first* (smallest-span) candidate, i.e. the bare `Checks`, which doesn't resolve on its own. The gate fell back to `None` → legacy whole-program name+arity scan → it could still bind to the unrelated static `Xunit.Assert.Single`.

**Fix (commit 2e04bcd):** select the receiver with the **widest span** so the full chain (`report.Checks`) drives classification → gate `Value` → the static is rejected and the LINQ extension is chosen.

**New regression test** mirrors the exact shape with BCL-only types: `d.Keys.Concat(other)` (a value-receiver LINQ extension whose receiver is a property access) must resolve to `System.Linq.Enumerable.Concat`, never the same-named static `String.Concat(...)` — even though both have the same parameter count relative to the call site. Verified the hover now returns `func (System.Linq.Enumerable) Concat(...)`.

All 350 `LanguageServer.Tests` pass; clean build.

> Note: to see the fix in VS Code, rebuild the language server binary the extension launches.

Issue #526, opened 2026-06-07
Parser: nested CLR types are not reachable (`Outer.Inner` fails to parse as a type clause)


DavidObando:
## Summary

A nested CLR type cannot be referenced via `Outer.Inner` — the parser trips on the `.` between `Outer` and `Inner`. There appears to be no other syntax for naming a nested type in a context where a type clause is expected.

## Repro

Sibling C#:
```csharp
namespace Probe.CSharp;
public sealed class Outer
{
    public interface INested { int Compute(); }
}
```

G#:
```gsharp
package Probe.Tests

import Probe.CSharp

type X class {
    func Use() {
        var x Outer.INested = nil
    }
}
```

## Observed

```
Tests.gs(7,20,7,21): error GS0005: Unexpected token <DotToken>, expected <IdentifierToken>.
Tests.gs(7,29,7,30): error GS0005: Unexpected token <EqualsToken>, expected <IdentifierToken>.
```

The parser treats `Outer` as a complete type clause, then chokes on `.`.

Similarly fails at:
- `type Impl class : Outer.INested { ... }`
- `let v Outer.INested = ...`
- argument types, return types, etc.

## Expected

`Outer.Inner` (or some other syntax — `Outer.Inner`, `Outer/Inner`, etc.) should be accepted as a type clause that resolves to the nested type, matching C# / Go semantics.

## Context

`gsc` version: `Gsharp.NET.Sdk/0.1.431`, .NET 10 RTM.

Discovered while porting `tests/Oahu.Cli.Tests/Tui/AppShellTests.cs` (uses `AppShell.IKeyReader`, a public nested interface). Several other tests in the same suite share the same blocker. Confirmed in isolated probe.

PR #913, opened 2026-06-21
Support nested type declarations (class/struct/interface/enum inside class/struct)

### LINKED ISSUE #910: Nested type declarations (class/struct/interface/enum inside a class/struct) are unsupported and produce a misleading parse-error cascade
## Summary

Nested type declarations are not supported. Declaring a `class` / `struct` / `interface` / `enum` inside the body of another `class` or `struct` produces a cascade of parse errors, because the aggregate-member grammar has no production for nested types.

## Repro

```gsharp
package Oahu.Cli.Tests

import System

class Outer {
    class Inner {
        func Hello() string {
            return "hi"
        }
    }

    func Make() string {
        let i = Inner()
        return i.Hello()
    }
}
```

### Observed

```
error GS0113: Type '' doesn't exist.
        class Inner {
error GS0288: Field declarations require a 'var' (mutable) or 'let' (read-only) keyword.
        class Inner {
error GS0005: Unexpected token <ClassKeyword>, expected <IdentifierToken>.
        class Inner {
...
```

A nested `struct` inside a `struct` (or `class`) fails the same way.

## Root cause

The class/struct body member loop (`src/Core/CodeAnalysis/Syntax/Parser.cs`, the `while (Current.Kind != CloseBraceToken ...)` at ~line 1511) dispatches only on member kinds: fields (`var`/`let`), `func`, `prop`, `event`, `init`, constructors, destructors. It has no case for the `class` / `struct` / `interface` / `enum` keywords, so a nested type declaration is misparsed as a malformed field, yielding GS0288 / GS0005 / GS0113. Nested types are simply absent from the aggregate-member grammar.

## Boundary probing

- Nested `class` in `class`, and nested `struct` in `struct`, both fail at parse time.
- This blocks the `Oahu/tests/Oahu.Cli.Tests/RootCommandTests.gs` port from nesting the `NullLoggerFactoryImpl` helper inside `RootCommandTests`; it had to be lifted to a top-level type as a workaround.

## Expected

Support nested type declarations as aggregate members (`class`/`struct`/`interface`/`enum` inside a `class`/`struct` body), emitted as nested CLR types with appropriate nested accessibility (`MapNestedTypeAccessibility` already exists in `src/Core/CodeAnalysis/Emit/AccessibilityMap.cs`).

If nested types are intentionally out of scope, a clear, dedicated diagnostic (e.g. "nested type declarations are not supported") would be far better than the current misleading field-declaration error cascade.

## Workaround

Declare the helper type at the top level instead of nesting it.

## Environment

- `gsc` Release build, net10.0 target.

DavidObando:
Fixes #910

## Summary

Nested type declarations (`class` / `struct` / `interface` / `enum` declared inside a `class` or `struct` body) were unsupported and produced a misleading parse-error cascade (GS0113 / GS0288 / GS0005). This PR implements **real CLR nested-type support end-to-end for every nested kind in either encloser** — no combination is deferred.

The issue's exact example now compiles, passes ilverify, and executes:

```gsharp
class Outer {
    class Inner {
        func Hello() string { return "hi" }
    }
    func Make() string {
        let i = Inner()
        return i.Hello()
    }
}
```

## Changes

**Parser** (`Parser.cs`, `StructDeclarationSyntax.cs`): `class`/`struct`/`interface`/`enum` (with an optional accessibility modifier) are accepted as members of a `class`/`struct` body by reusing the existing top-level `ParseAggregateDeclaration` routine — so nesting is **recursive** (`Outer.Middle.Inner`). Parsed nested declarations are stored on `StructDeclarationSyntax.NestedTypes`.

**Binder / Symbols** (`DeclarationBinder.cs`, `StructSymbol`/`EnumSymbol`/`InterfaceSymbol`): each type symbol gains a `ContainingType` set by `BindNestedTypeDeclarations`, which binds each nested declaration with the same per-kind binder used for top-level types. **All** nested kinds (including `interface`, and `class`-in-`struct`) are marked nested. Nested types register in the flat type-alias scope and resolve by their **simple** name from the enclosing type's members.

**Emit — unified emission-order refactor** (`ReflectionMetadataEmitter.cs`, `TypeDefEmitter.cs`, `AccessibilityMap`): `EmitStructTypeDef`/`EmitEnumTypeDef`/`EmitInterfaceTypeDef` are nested-aware — when `ContainingType != null` they use `MapNestedTypeAccessibility` (`internal→NestedAssembly`, `private→NestedPrivate`, else `NestedPublic`) and an empty namespace.

`ReflectionMetadataEmitter.EmitCore` now emits **all user nested types in a single contiguous pre-order block** placed after every top-level type and before the per-package `<Program>` and state-machine `TypeDef`s. Each enclosing `TypeDef` row (top-level or an outer nested type) is materialised before any of its descendants, so **ECMA-335 §II.22.32** (enclosing row < nested row) holds for *every* kind/encloser combination. A fixed kind-partitioned order cannot do this — `struct`-in-`class` and `class`-in-`struct` impose contradictory partition orders, and nested interfaces can never follow their encloser. The same `nestedOrdered` pre-order sequence drives field-row planning, method-row planning, `TypeDef` emission, `MethodDef` body emission, and `NestedClass` row insertion, preserving the monotone `FieldList`/`MethodList` columns and the sorted `NestedClass` table. Non-nested programs produce an empty nested block, so their metadata is **byte-identical** to before.

## Supported nested kind × encloser combinations

All emitted as **true CLR nested types**, verified by ilverify + execution + a reflection assertion (`Inner.IsNested`, `DeclaringType == Outer`):

| Nested kind | in `class` | in `struct` |
|-------------|:----------:|:-----------:|
| `class`     | ✅          | ✅           |
| `struct`    | ✅          | ✅           |
| `enum`      | ✅          | ✅           |
| `interface` | ✅          | ✅           |

Every kind/encloser combination is now a real nested type. Recursive nesting works to arbitrary depth.

## GS0369 removed

The earlier dedicated deferral diagnostic `GS0369` (nested interface / `class`-in-`struct`) is **removed** — those combinations are fully supported, so the diagnostic is no longer reachable. Its definition (`DiagnosticBag.ReportUnsupportedNestedTypeCombination`) and the binder gate are deleted.

## Name resolution & accessibility

- Unqualified simple-name resolution of a nested type works from the enclosing type's members and across the compilation (flat type-alias scope).
- Qualified `Outer.Inner` resolution for **user-declared** nested types is not newly added (imported CLR nested-type resolution from #526/#569 is unaffected). A consequence is that two enclosers can't reuse a nested simple name in one compilation.
- IL visibility uses `MapNestedTypeAccessibility`.

See **ADR-0110** (`docs/adr/0110-nested-type-declarations.md`) for the full design, the unified emission-order decision, consequences, test coverage, and alternatives considered.

## Tests

- `Issue910NestedTypeParserTests` — all four kinds in both enclosers, recursion, accessibility modifier, sibling members.
- `Issue910NestedTypeBinderTests` — `ContainingType` set for every nested kind/encloser combination (class, struct, interface, enum) and recursive chaining.
- `Issue910NestedTypeEmitTests` — issue example + **every** kind/encloser combination (ilverify + execute), including `class`-in-`struct` constructed and used, and a nested `interface` (in a `class` and in a `struct`) implemented by a sibling nested class and dispatched through; reflection proof of real nesting.

## Local build + test

`dotnet build GSharp.sln -c Release -graph` — succeeded (0 warnings/errors).

`dotnet test` (all projects) — all green:
- Core.Tests: 3373 passed, 1 skipped
- Compiler.Tests: 1389 passed
- Interpreter.Tests: 308 passed
- LanguageServer.Tests: 350 passed
- Sdk.Tests: 38 passed

### DIFF
--- a/docs/adr/0110-nested-type-declarations.md
+++ b/docs/adr/0110-nested-type-declarations.md
@@ -1,0 +1,197 @@
+# ADR-0110: Nested type declarations
+
+- **Status**: Accepted
+- **Date**: 2026-06-20
+- **Phase**: Phase 5 — generics & dogfooded core
+- **Closes**: Issue #910 (nested type declarations inside a class/struct are unsupported and produce a misleading parse-error cascade)
+- **Related**: ADR-0017 (sealed-by-default classes); ADR-0078 (sealed hierarchies); #526/#569 (construction/type-clause resolution of *imported* CLR nested types); ADR-0109 (top-level accessibility)
+
+## Context
+
+A `class` or `struct` body could not declare a nested type. The
+class/struct body member loop in `Parser.cs` dispatched only on member
+kinds — fields (`var`/`let`), `func`, `prop`, `event`, `init`,
+constructors, and destructors — and had **no** production for the
+`class` / `struct` / `interface` / `enum` keywords. A nested type
+declaration was therefore misparsed as a malformed field, which cascaded
+into a sequence of misleading errors, e.g.:
+
+```gsharp
+class Outer {
+    class Inner {                 // GS0113 Type '' doesn't exist
+        func Hello() string {     // GS0288 field-keyword expected
+            return "hi"
+        }
+    }                             // GS0005 unexpected <ClassKeyword>
+    func Make() string {
+        let i = Inner()
+        return i.Hello()
+    }
+}
+```
+
+Nested types were simply absent from the aggregate-member grammar even
+though the rest of the pipeline already had most of the machinery: the
+binder registers every user type in a flat type-alias scope, and the
+emitter already produced CLR nested `TypeDef`s for compiler-synthesized
+state machines and closures (`EmitNestedStructTypeDef`,
+`AccessibilityMap.MapNestedTypeAccessibility`,
+`MetadataBuilder.AddNestedType`). What was missing was a source-level
+spelling and the wiring to make a *user-declared* nested type flow
+through parse → bind → emit.
+
+A hard constraint shapes the emit design: ECMA-335 §II.22.32 requires
+the **enclosing** `TypeDef` row to precede the **nested** `TypeDef` row,
+and the `Field`/`MethodDef` list columns must be monotonically
+non-decreasing across `TypeDef` rows. The emitter assigns rows in a
+fixed kind-partitioned order: interfaces → classes → structs → enums →
+`<Program>` → state-machine types. That order already guarantees
+"enclosing before nested" for several kind/encloser combinations but
+*not* for all of them.
+
+## Decision
+
+### Grammar
+
+`class` / `struct` / `interface` / `enum` are accepted as members of a
+`class` or `struct` body. The body member loop detects an (optional
+accessibility modifier followed by an) aggregate-declaration head and
+reuses the existing top-level `ParseAggregateDeclaration` routine, so a
+nested type is parsed by exactly the same code as a top-level one. The
+parsed nested declarations are stored on a new
+`StructDeclarationSyntax.NestedTypes` list. Because the top-level
+routine is reused, nesting is **recursive**: a nested type may itself
+contain nested types (`Outer.Middle.Inner`).
+
+### Binding & name resolution
+
+`DeclarationBinder.BindNestedTypeDeclarations` binds each nested
+declaration through the same per-kind binders used for top-level types
+and calls `SetContainingType` on the resulting symbol
+(`StructSymbol`/`EnumSymbol`/`InterfaceSymbol` each gained a
+`ContainingType` property). Nested types register in the same flat
+type-alias scope as top-level types, keyed by their **simple** name.
+
+Consequences of the flat scope:
+
+- A nested type resolves by its **unqualified** simple name from
+  anywhere the type is visible — in particular from the enclosing type's
+  own members, which satisfies the issue's `Inner()` construction inside
+  `Outer.Make()`.
+- Qualified `Outer.Inner` resolution for **user-declared** nested types
+  is *not* newly added here. (Dotted resolution of *imported* CLR nested
+  types via `#526`/`#569` is unaffected.) Authors reference a nested user
+  type by its simple name.
+- Because the simple name is global, two distinct enclosers cannot
+  declare nested types with the same simple name in one compilation; the
+  second triggers the normal "symbol already declared" diagnostic.
+
+### Accessibility
+
+A nested type's IL visibility is mapped with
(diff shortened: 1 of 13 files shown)


DavidObando:
## Follow-up complete: previously-deferred combinations now fully supported

The two combinations this PR originally deferred behind `GS0369` — a nested **`interface`** (in any encloser) and a **`class` nested in a `struct`** — are now implemented as real CLR nested types that pass ilverify **and** execute. The earlier "deferred" note is superseded.

### Emission-order refactor
The root cause was the fixed *kind-partitioned* TypeDef emission order (interfaces → classes → structs → enums), which cannot satisfy ECMA-335 §II.22.32 (enclosing TypeDef row must precede its nested rows) for all combinations: `struct`-in-`class` and `class`-in-`struct` demand contradictory partition orders, and nested interfaces could never follow their encloser.

`ReflectionMetadataEmitter.EmitCore` now emits **all user nested types in a single contiguous pre-order block**, placed after every top-level type and before `<Program>`/state-machine TypeDefs. Each enclosing TypeDef row (top-level or an outer nested type) is materialised before any descendant, so §II.22.32 holds for every kind/encloser combination. The same `nestedOrdered` sequence drives field-row planning, method-row planning, TypeDef emission, MethodDef body emission, and NestedClass row insertion — preserving the monotone FieldList/MethodList columns and the sorted NestedClass table. Non-nested programs produce an empty nested block, so their metadata is **byte-identical** to before.

### Final support matrix (all ✅)
| Nested kind | in `class` | in `struct` |
|-------------|:----------:|:-----------:|
| `class`     | ✅ | ✅ |
| `struct`    | ✅ | ✅ |
| `enum`      | ✅ | ✅ |
| `interface` | ✅ | ✅ |

### GS0369 removed
With no remaining deferral, `GS0369` (`DiagnosticBag.ReportUnsupportedNestedTypeCombination`) and its binder gate are deleted. The two deferred GS0369 tests (binder + emit) are converted to **positive** tests: a `class`-in-`struct` constructed and used from an enclosing method, and a nested `interface` (in a `class` and in a `struct`) implemented by a sibling nested class and dispatched through.

### ADR-0110
Updated to an all-✅ matrix; the unified pre-order emission order is recorded in the Decision/Consequences as **done** (no longer a follow-up), and the GS0369 mentions are removed.

### Full suite — all green
- Core.Tests: 3373 passed, 1 skipped
- Compiler.Tests: 1389 passed
- Interpreter.Tests: 308 passed
- LanguageServer.Tests: 350 passed
- Sdk.Tests: 38 passed

**Nothing remains deferred.**

Issue #977, opened 2026-06-23
Binder: BCL method with an inline 'out var x' argument fails overload resolution (GS0159)


DavidObando:
## Summary
A BCL method invoked with an **inline `out var x`** declaration fails overload resolution with `GS0159: Cannot find function TryGetValue`. The same method binds when the out argument is passed by address (`&x`) to a pre-declared local, and a **user-defined** `func` with an `out` parameter binds the inline `out var x` form fine. So the failure is specific to the inline-`out var` shape against BCL/imported methods.

Discovered migrating `Dictionary<string,int>.TryGetValue(k, out var v)` via the `cs2gs` corpus (`L4-Console`).

## Repro
```gs
package T
import System
import System.Collections.Generic
func mainx() int32 {
    let d = Dictionary[string, int32]()
    d["a"] = 5
    d.TryGetValue("a", out var v)
    return 0
}
```

```
error GS0159: Cannot find function TryGetValue.
    d.TryGetValue("a", out var v)
```

## Controls that DO compile
```gs
var slot int32
let ok = d.TryGetValue("a", &slot)   // BCL via pre-declared pass-by-address: OK

func tryProduce(out result int32) bool { ... }
tryProduce(out var n)                 // user func + inline out var: OK
```

## Notes
- Classification: binder/overload-resolution gap. The inline `out var x` argument shape is not matched against imported (BCL) method overloads, though it is matched against user-defined functions.
- Root-cause pointer: overload resolution for inline `out var` declarations appears to only consider source-declared functions, not imported `MethodInfo` overloads.
- `cs2gs` works around it by emitting the pre-declared `var x T` + `&x` pass-by-address form for BCL `out` calls.

gsc 0.2.118+160636d8cc. Discovered by the cs2gs migration pipeline (issue #914).

Issue #903, opened 2026-06-20
Untyped arrow lambda to generic LINQ methods (Single/Select) fails for same-compilation user element types


DavidObando:
### Summary

An untyped arrow lambda passed to a generic method (e.g. LINQ `Single` / `Select` / `Where`) fails to bind when the receiver's generic element type is a **user type defined in the same compilation**. The lambda's parameter type cannot be inferred, so the call reports `GS0159 Cannot find function ...` and `GS0304 Cannot infer the type of lambda parameter ...`.

This is a pre-existing limitation discovered while fixing the deferred-lambda selector case (`Select((c) -> c.Id)` for **referenced** element types). The referenced-type case now works; the **same-compilation** case still fails for both `Single` and `Select`.

### Repro

```gsharp
package P
import System
import System.Linq
import System.Collections.Generic

struct Check {
    let Id string
}

let checks = List[Check]()
checks.Add(Check { Id: "audible-api" })
checks.Add(Check { Id: "network" })

// Both of these fail today:
let net = checks.Single((c) -> c.Id == "network")
let ids = checks.Select((c) -> c.Id).ToHashSet()
```

Errors:

```
error GS0159: Cannot find function Single.
error GS0158: Cannot find member Id.
error GS0159: Cannot find function Select.
```

### Workarounds

- Annotate the lambda parameter explicitly: `checks.Single((c Check) -> c.Id == "network")` works.
- Use a referenced (imported) element type instead of one defined in the same compilation — that path works (`List[string]().Select((w) -> w.ToUpper())`, etc.).

### Root cause (initial analysis)

The deferred arrow-lambda resolver infers the lambda's parameter types from the receiver's CLR type via reflection (`OverloadResolution` / `ExpressionBinder.ResolveDeferredArrowLambdaArguments` in `src/Core/CodeAnalysis/Binding/`). When the receiver is `List[Check]` and `Check` is a `StructSymbol` still being compiled, the generic argument has **no CLR `Type`** yet, so:

- `receiver.Type.ClrType` for the receiver, and the inferred type-parameter bounds, cannot produce a usable CLR element type, and
- member access in the lambda body (`c.Id`) cannot resolve.

[...]

Issue #832, opened 2026-06-14
Discarded open-T method return inserts a spurious `unbox.any !T`


DavidObando:
**Repro.**
```
func WindowedIterator[T](source IEnumerable[T], size int32) IEnumerable[IList[T]] {
    var buffer = Queue[T](size)
    for item in source {
        buffer.Enqueue(item)
        if buffer.Count == size {
            yield List[T](buffer)
            buffer.Dequeue()
        }
    }
}
```

**Symptom.** At runtime: `System.NullReferenceException` from inside `MoveNext`. `ilverify`:
```
[StackObjRef]: Sequences.<Program>+<WindowedIterator>d__9`1::MoveNext()
  [offset 0x000000EB][found value 'T'] Expected an ObjRef on the stack.
```

**Root cause (suspected).** The expression-statement `buffer.Dequeue()` discards a `T`. Emitted IL is `callvirt instance !0 Queue`1<T>::Dequeue(); unbox.any !T; pop` — the spurious `unbox.any !T` is emitted even though the call already returns `T`. For reference T, `unbox.any` is a `castclass` that NREs on a non-boxed value.

Workaround used in #806: replace `Queue[T]`/`Dequeue()` with `List[T]`/`RemoveAt(0)`.

**Expected.** Discarded `T`-typed expressions should not have an extra conversion inserted. Emit `callvirt; pop` directly.

Related: #806, ADR-0084 §L5.

Issue #794, opened 2026-06-13
Generic-method T does not thread through generic instance-method calls inside a generic shared method


DavidObando:
Discovered while porting `Gsharp.Extensions.Sequences` to G# under #792.

## Repro

```gs
package Repro
import System
import System.Collections.Generic

class Sequences {
    shared {
        func MakeList[T any]() []T {
            var list = List[T]()
            list.Add(default)
            return list.ToArray()
        }
    }
}
```

The call `List[T]()` is bound, but `list.ToArray()` is then bound as returning `object[]` instead of `T[]`. The downstream conversion to `[]T` fails with GS0155: `Cannot convert type 'System.Object[]' to '[]T'.`

## Expected

`List[T]().ToArray()` should thread the method's generic parameter `T` through the instance call so the result type is `T[]`. This works for top-level `func` (samples/GenericConstruction.gs) when `T` is concrete; it fails for the generic-T case inside a generic shared method.

## Context

Blocking the dogfooded port of `Gsharp.Extensions.Sequences.Sequences.Range`, `Iterate`, `Repeat`, `Of` and the `SequenceExtensions` collectors (#792 §L5).

PR #1105, opened 2026-06-25
Fixes #1100: recover same-compilation user element type for BCL generic member access

### LINKED ISSUE #1100: Emit GS9998: member access on a BCL generic instantiated with a same-compilation user type (TypeBuilderInstantiation) — must use TypeBuilder.GetMethod
## Summary

Emitting a member access (method call / field / constructor) on a **constructed BCL generic type whose type argument is a same-compilation user type** throws a fatal emit-stage `GS9998: NotSupportedException: TypeBuilder generic instantiation does not support resolving members. Use TypeBuilder.GetMethod instead.`

When a generic type (e.g. `System.Collections.Generic.Queue<T>`) is instantiated with a type argument that is a **`TypeBuilder` defined in the same compilation/module** (e.g. a user `class ChapterEntry`), `MakeGenericType` produces a `System.Reflection.Emit.TypeBuilderInstantiation`. Calling `GetMethod` / `GetField` / `GetConstructor` directly on that instantiation is **not supported by Reflection.Emit** — you must use the static helpers `TypeBuilder.GetMethod(instantiatedType, methodOnGenericDef)`, `TypeBuilder.GetField(...)`, `TypeBuilder.GetConstructor(...)`. The emitter currently resolves the member off the instantiation directly, so emit aborts.

This is the same family as #1088 (same-compilation user type as a CLR generic argument), but on the **emit** side (member resolution) rather than the binder's type-identity check.

## Discovery

Surfaced by the **cs2gs** C#→G# migration pipeline (#914) translating the **Oahu.Decrypt** library. `Oahu.Decrypt.ChapterQueue` holds `private readonly Queue<ChapterEntry> chapterEntries = new();` and calls `.Enqueue(...)`, `.Dequeue()`, `.Count` on it (`ChapterEntry` is a class declared in the same compilation). The whole project binds successfully; emit then aborts:

```
ChapterQueue.gs(1,1,1,1): error GS9998: NotSupportedException: TypeBuilder generic instantiation does not support resolving members. Use TypeBuilder.GetMethod instead.
```
(gsc 0.2.181+bdbd843c89, the only error in an otherwise-clean 140-file compile.)

## Minimal shape

```gsharp
package p
class Entry { }
class C {
    private let q Queue[Entry] = Queue[Entry]()
    func add(e Entry) { q.Enqueue(e) }
    func drain() { while q.Count > 0 { let x = q.Dequeue() } }
}
```
The exact construct that aborts emit only reproduces inside the full multi-file Decrypt compile (whole-project, full reference set), where all types bind before emit. Reproduce deterministically via:
`dotnet out/bin/Release/Cs2Gs.Cli/cs2gs.dll migrate --corpus ~/GitHub/DavidObando/Oahu/src --app corpus/Oahu.Decrypt --gsc out/bin/Release/Compiler/gsc.dll --out /tmp/RUN --config Release`

## Related binding facet (likely same root)

Compiling the shape above **standalone** (single/2-file) instead reports a *binding* error: `GS0155: Cannot convert type 'object' to 'Entry'` for `q.Dequeue()` — the generic method's `T` return type erases to `object` for a same-compilation user type argument (cf. the existing `ImportedTypeSymbol.HasSubstitutableTypeArgument` / `MapErasedIndexerElementType` recovery, which appears to cover indexers and `for-in` but not general generic-method return/parameter signatures). In the full Decrypt compile this binding step passes (so emit is reached); the fix should ensure both the binder recovers the symbolic return/parameter type **and** the emitter uses `TypeBuilder.GetMethod`/`GetField`/`GetConstructor` for `TypeBuilderInstantiation` members.

## Impact

Blocks faithful cs2gs migration of any code using BCL generic collections (`Queue<T>`, `List<T>`, `Dictionary<TKey,TValue>`, …) parameterised by same-compilation user types — extremely common. This is the **last** compile error blocking `Oahu.Decrypt` from compiling cleanly.

DavidObando:
## Summary

Emitting a member access on a constructed BCL generic whose type argument is a **same-compilation** user `class`/`data struct` (e.g. `Queue[Entry]`, `List[Entry]`, `Dictionary[string, Entry]`) lost the user element type. A generic member whose open signature mentions the type-level parameter `T` was projected from the **type-erased** closed shape, so:

- `var x Entry = q.Dequeue()` → **GS0155** ("Cannot convert type 'object' to 'Entry'"), because `Queue<T>.Dequeue()` surfaced as `object`.
- `d.TryGetValue(k, out var found); found.V` → **GS0158** ("Cannot find member V"), because the `out TValue` of `Dictionary<K,V>.TryGetValue` typed `found` as `object`.

In a full multi-file compile where the result is left erased, the downstream emit could not resolve members against the symbolic instantiation and aborted with **GS9998** (`NotSupportedException` from the constructed-generic member-resolution path). This blocked faithful cs2gs migration of any code using BCL generic collections parameterised by a same-compilation user type (the last error blocking Oahu.Decrypt).

## Root cause

The binder's erasure-recovery overrides only kept the symbolic projection when it still contained an **in-scope type parameter** (#794). For a same-compilation user type argument the projection is the user type itself (no type parameter), so the override was skipped and the erased `object` was used:

- `ExpressionBinder.Calls.cs` — `ResolveInstanceReturnTypeFromReceiver` gated its result on `ContainsTypeParameter(mapped)` only.
- `ExpressionBinder.Calls.cs` — `RebindInlineOutVarArguments` (#977) typed the out-var directly from the erased CLR pointee (`out object`).

## Fix

**Binder** (`src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.cs`):
- `ResolveInstanceReturnTypeFromReceiver` now also keeps the projection when it references a same-compilation user type (`TypeSymbol.ContainsSameCompilationUserType`), mirroring the generic-method counterpart added for #903. This recovers `Queue<T>.Dequeue() → Entry`.
- New `ResolveInstanceParameterPointeeTypeFromReceiver` recovers the symbolic pointee for a generic by-ref parameter; `RebindInlineOutVarArguments` threads the receiver type and uses it so `out var` against `Dictionary[K, V].TryGetValue` binds the user value type instead of `object`.

**Emit**: no change required. Once binding recovers the user element type, the existing symbolic-container MemberRef machinery (#649 / #671 / #832 / #903) already parents the call/ctor/field reference at the symbolic generic instantiation — encoded from the **open** definition rather than resolving members off the erased closed shape — and short-circuits the post-call erasure-widening (`TryGetSymbolicSubstitutedInstanceMethodReturn`, which already recognises `ArgIsSymbolicUserDefined`). The program therefore emits, IL-verifies and runs correctly.

## Tests

Added `test/Compiler.Tests/Emit/Issue1100BclGenericUserArgEmitTests.cs` — five tests that compile, IL-verify and **execute**:
- `Queue[Entry]` (user class) — Enqueue (param `T`), Dequeue (return `T`), Count.
- `Queue[Item]` (user `data struct`) — value-type element (no spurious `unbox.any`).
- `List[Entry]` — Add, indexer read, for-in element type, Count.
- `Dictionary[string, Entry]` — indexer write/read and `TryGetValue` out-var.
- `Queue[Entry].Dequeue()` returned directly from a method whose declared return type is the user type.

### Results
- New tests: **5 passed**.
- `Issue968 / Issue939 / Issue903 / Issue977 / Issue794 / Issue832`: **40 passed**.
- `~Generic` emit tests: **180 passed**.
- Collection/Dictionary/Queue/for-in emit tests: **91 passed**.
- `Emit.Issue9xx / Emit.Issue10xx`: **383 passed**.
- Full `Core.Tests`: **3828 passed, 1 skipped**.

No regressions observed.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.cs
@@ -1176,7 +1176,74 @@ internal sealed partial class ExpressionBinder
         }
 
         var mapped = MemberLookup.MapOpenClrTypeToSymbolic(openReturn, imp.OpenDefinition, imp.TypeArguments);
-        return TypeSymbol.ContainsTypeParameter(mapped) ? mapped : null;
+
+        // Issue #794 surfaced this override when the projection still contains
+        // an in-scope type parameter. Issue #1100 extends it to a
+        // same-compilation user type argument: when the receiver is e.g.
+        // `Queue[Entry]` and `Entry` is a class/struct still being compiled,
+        // the closed CLR shape erased `T` to `object`, so a member call
+        // returning `T` (e.g. `Queue<T>.Dequeue()`) would otherwise surface as
+        // `object` and lose the `Entry` identity (GS0155 at the call site).
+        // The symbolic projection recovers it.
+        return TypeSymbol.ContainsTypeParameter(mapped) || TypeSymbol.ContainsSameCompilationUserType(mapped)
+            ? mapped
+            : null;
+    }
+
+    /// <summary>
+    /// Issue #1100: the by-ref-parameter counterpart of
+    /// <see cref="ResolveInstanceReturnTypeFromReceiver"/>. When a call is
+    /// dispatched against a receiver whose <see cref="ImportedTypeSymbol"/>
+    /// carries symbolic type arguments (e.g. <c>Dictionary[K, V]</c>),
+    /// substitute the open declaring type's parameter pointee type using the
+    /// receiver's symbolic arguments. Without this an inline <c>out var</c>
+    /// argument against a generic by-ref parameter (e.g. the <c>out TValue</c>
+    /// of <c>Dictionary&lt;K, V&gt;.TryGetValue</c>) would bind from the
+    /// type-erased closed shape (<c>out object</c>), losing the symbolic
+    /// projection (the user element type). Returns <see langword="null"/> when
+    /// no override is needed so callers keep their existing pointee derivation.
+    /// </summary>
+    /// <param name="receiverType">The receiver's static type symbol.</param>
+    /// <param name="closedMethod">The closed method selected by overload resolution.</param>
+    /// <param name="paramIndex">The zero-based parameter position to recover.</param>
+    /// <returns>The override pointee type symbol, or <see langword="null"/>.</returns>
+    private static TypeSymbol ResolveInstanceParameterPointeeTypeFromReceiver(
+        TypeSymbol receiverType,
+        System.Reflection.MethodInfo closedMethod,
+        int paramIndex)
+    {
+        if (receiverType is not ImportedTypeSymbol imp
+            || imp.OpenDefinition == null
+            || imp.TypeArguments.IsDefaultOrEmpty
+            || closedMethod == null
+            || paramIndex < 0)
+        {
+            return null;
+        }
+
+        var openMethod = TryGetOpenInstanceMethod(imp.OpenDefinition, closedMethod);
+        if (openMethod == null)
+        {
+            return null;
+        }
+
+        var openParameters = openMethod.GetParameters();
+        if (paramIndex >= openParameters.Length)
+        {
+            return null;
+        }
+
+        var openParamType = openParameters[paramIndex].ParameterType;
+        var openPointee = openParamType.IsByRef ? openParamType.GetElementType() : openParamType;
+        if (openPointee == null)
+        {
+            return null;
+        }
+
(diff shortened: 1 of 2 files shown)


DavidObando:
Superseded. Issue #1100 was fixed in parallel by #1102 (merged as ed8ec378), which independently landed the same `ResolveInstanceReturnTypeFromReceiver` recovery this PR contained — hence the conflict against main.

The one genuinely net-new change here — out-var pointee recovery (`ResolveInstanceParameterPointeeTypeFromReceiver`) for `Dictionary[K,V].TryGetValue(k, out var v)` over same-compilation user types, which #1102 did **not** cover — has been split out as its own issue #1107 and reopened cleanly off current main as #1108. Closing this one to avoid a conflicting duplicate.

Issue #773, opened 2026-06-12
Generic-receiver extension methods don't dispatch at call sites


DavidObando:
**Discovered while closing #751** (parser fix for L2 from ADR-0084).

The parser now accepts a receiver clause that uses a function-level type
parameter — e.g. `func (self sequence[T]) FirstOrNil[T]() T?` — and the
declaration binds without error. But the binder cannot dispatch a call
site like `myEnumerable.FirstOrNil()` to such a method: it reports
GS0159 "Cannot find function FirstOrNil" as if the extension didn't
exist.

### Repro

```gs
package P
import System
import System.Collections.Generic

func (self IEnumerable[T]) MyFirst[T any](fb T) T {
    return fb
}

var arr = []int32{10, 20, 30}
Console.WriteLine(arr.MyFirst(99))  // GS0159: Cannot find function MyFirst
```

The same call works when the receiver is a *closed* type
(\`func (self string?) ...\` on a \`string?\` value) — so the gap is
specifically receiver-substitution for extension dispatch when the
receiver type contains an open type parameter belonging to the
extension function.

### Why it matters

ADR-0084's dogfooded G# rewrite of `Gsharp.Extensions.Optional` and
`Gsharp.Extensions.Sequences` needs exactly this shape — every
`sequence[T]`/`T?` extension in those namespaces opens a fresh `T`
on the extension function. The parser fix from #751 was necessary but
not sufficient; this gap (and a few related ones — see follow-ups)
must close before the port can land.

### Likely location

`src/Core/CodeAnalysis/Binding/OverloadResolver.cs` —
`TryBuildReceiverSubstitution` and the candidate-filtering path
around extension-method dispatch.

### Related

- #751 / ADR-0084 §L2 (parser-side closed in this PR)
- #724 / ADR-0084 (Extensions stdlib parent)
- #706 (Oats cleanup parent)

PR #1110, opened 2026-06-25
Fixes #1103: extension functions on imported/primitive receivers callable with instance syntax

### LINKED ISSUE #1103: Extension function on imported/BCL/primitive receiver type not callable with instance syntax x.ExtMethod() (GS0159)
## Summary

An **extension function** (receiver-clause `func`, ADR-0019) whose **receiver type is an imported/BCL or primitive CLR type** cannot be invoked with **instance/member syntax** `receiver.ExtMethod(args)`. The binder reports `GS0159: Cannot find function <name>`. The same extension binds correctly when called as a **free function** `ExtMethod(receiver, args)`, and instance syntax works fine when the receiver is a **package-owned** struct/class.

Extension methods on BCL types (`Stream`, `byte[]`, `int`, `TimeSpan`, …) are extremely common in idiomatic C#, so this blocks faithful migration of real code (discovered while migrating the Oahu MPEG-4 reader via cs2gs: ~435 `GS0159` diagnostics trace back to this single gap).

## Minimal repro

```gsharp
package a.util

import System.IO

func (stream Stream) ReadU32() uint32 {
    return uint32(5)
}

func (n int32) Doubled() int32 {
    return n * 2
}

class C {
    func UseFree(file Stream) uint32 {
        return ReadU32(file)        // OK — binds as free function
    }
    func UseInstance(file Stream) uint32 {
        return file.ReadU32()       // GS0159: Cannot find function ReadU32
    }
    func UseInt(x int32) int32 {
        return x.Doubled()          // GS0159: Cannot find function Doubled
    }
}
```

`UseFree` compiles; `UseInstance`/`UseInt` fail with `GS0159`. Replacing the receiver with a **package-owned** struct (e.g. `func (p Point) Sum() int32` called as `p.Sum()`) compiles (only the expected ADR-0079 `GS0314` warning), confirming the gap is specific to **imported/primitive** receiver types.

## Root cause

`BoundScope.TryLookupExtensionFunction` (src/Core/CodeAnalysis/Binding/BoundScope.cs:286) matches a candidate by **reference equality** on the receiver type symbol:

```csharp
if (ext.Name == name && ext.ExtensionReceiverType == receiverType)  // reference equality
```

For an imported CLR type, the `ExtensionReceiverType` symbol captured at the extension declaration and the receiver's type symbol at the call site are **distinct `ImportedTypeSymbol` instances** wrapping the same CLR type (imported type symbols are not interned), so the `==` reference comparison fails and the lookup misses. The open-generic-receiver fallback (unification, lines 321–359) only runs for receivers that mention the function's own type parameters, so a plain concrete imported receiver (`Stream`, `int32`) falls through to `GS0159`.

The free-function call path works because it routes through normal overload resolution (applicability/conversion checks), not reference equality.

The member-access call paths that consult this lookup are `ExpressionBinder.Calls.cs:2481` (user-type receiver) and `:2681` (imported-CLR receiver).

## Suggested fix

In `TryLookupExtensionFunction`, compare the candidate's `ExtensionReceiverType` to `receiverType` **structurally** — for imported types compare by underlying `ClrType` identity (e.g. `ClrTypeUtilities.AreSame` / `IsSameAs`) rather than reference equality; keep the existing open-generic unification branch for type-parameter receivers. Care should be taken to preserve the constraint-ranking behaviour and to handle primitive receivers (`int32`, etc.).

## Expected

`file.ReadU32()` and `x.Doubled()` bind to the receiver-clause extension functions, identically to the free-function form, matching ADR-0019 ("call sites `x.Foo(args)` bind to `Foo(x, args)`").

DavidObando:
## Summary
An extension function (receiver-clause `func`, ADR-0019) whose receiver is an imported/BCL or primitive CLR type could not be invoked with instance/member syntax `receiver.ExtMethod(args)` from inside a function/method body — the binder reported **GS0159: Cannot find function**. Free-function syntax `ExtMethod(receiver, args)` and package-owned struct receivers worked fine. This blocks faithful migration of idiomatic C# (extension methods on `Stream`, `byte[]`, `int`, ...).

The minimal repro from the issue now binds and runs:
```gsharp
func (stream Stream) ReadU32() uint32 { return uint32(5) }
func (n int32) Doubled() int32 { return n * 2 }
class C {
    func UseInstance(file Stream) uint32 { return file.ReadU32() } // was GS0159
    func UseInt(x int32) int32 { return x.Doubled() }             // was GS0159
}
```

## Root cause (two gaps)
1. **Scope re-registration (primary).** Extension functions are folded into `BoundGlobalScope.Functions` (`Binder.cs`) so free-call overload resolution can find them. But when the body-binding parent scope is rebuilt in `CreateParentScope`, those functions were re-declared **only as plain functions** (`TryDeclareFunction`), never as extensions (`TryDeclareExtensionFunction`). So every body-binding scope had an empty extension table and `TryLookupExtensionFunction` returned nothing → GS0159. This was masked because top-level call sites are bound in the *same* pass that declares the extension, and package-owned struct receivers are *also* registered as struct instance methods.
2. **Reference-equality match.** `BoundScope.TryLookupExtensionFunction` matched the declared receiver against the call-site receiver by reference equality, which is not robust for imported types whose declaration- and call-site symbols can be distinct `ImportedTypeSymbol` instances wrapping the same CLR type (e.g. across metadata-load contexts).

## Fix
- **`Binder.CreateParentScope`**: also re-register `IsExtension` functions via `TryDeclareExtensionFunction` (in addition to `TryDeclareFunction`), so member syntax resolves in body scopes while free-call syntax keeps working.
- **`BoundScope.TryLookupExtensionFunction`**: compare receivers structurally via a new `ExtensionReceiverMatches` helper — a reference-equality fast path backed by CLR-type identity (`ClrTypeUtilities.IsSameAs`, `FullName`-based and cross-context safe), plus symbolic type-argument comparison for erased constructed generics so `List[A]` never matches `List[B]`. User types (null `ClrType`) are matched only by reference, so a user receiver never collides with an imported one and two different imported types never collide. The open-generic unification branch (#773 / #775 constraint ranking) is left **unchanged**.

## Tests
- **`Issue1103ExtensionImportedReceiverTests`** (binder + interpreter): imported BCL reference receiver in a method body; primitive receiver (bind + execute → 42); closed-generic `List[int32]` receiver (bind + execute → 6); package-owned struct receiver still binds (no GS0159); negative over-match (extension on `Stream` not found for `TextWriter` → still GS0159); open-generic `IEnumerable[T]` unification in a method body still binds + executes.
- **`Issue1103ExtensionImportedReceiverEmitTests`** (compile → load → reflective invoke): primitive and imported-BCL instance-syntax calls dispatch and return the correct value.

## Verification
- Full `Core.Tests`: 3840 passed, 1 skipped.
- `Compiler.Tests` extension/receiver/overload slice: 123 passed.
- Targeted extension/receiver/overload/#773/#775/#792 slice: 308 passed.
- No new node kinds; pure binder lookup + scope-reconstruction logic.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/Binder.cs
+++ b/src/Core/CodeAnalysis/Binding/Binder.cs
@@ -1812,6 +1812,24 @@ public sealed class Binder
             foreach (var f in previous.Functions)
             {
                 scope.TryDeclareFunction(f);
+
+                // Issue #1103: extension functions are folded into
+                // BoundGlobalScope.Functions (so free-function call syntax
+                // `Ext(receiver, …)` resolves through normal overload
+                // resolution). They must ALSO be re-registered as extension
+                // functions on the rebuilt body-binding scope, otherwise
+                // member/instance call syntax `receiver.Ext(…)` inside a
+                // function/method body finds no candidate (the scope walked at
+                // the call site carries an empty extension table) and reports
+                // GS0159. This is independent of the receiver type — it was
+                // previously masked only because top-level call sites are bound
+                // in the same pass that declares the extension, and because
+                // package-owned struct receivers also register the function as
+                // an instance method on the struct.
+                if (f.IsExtension)
+                {
+                    scope.TryDeclareExtensionFunction(f);
+                }
             }
 
             foreach (var v in previous.Variables)
--- a/src/Core/CodeAnalysis/Binding/BoundScope.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundScope.cs
@@ -265,10 +265,13 @@ public sealed class BoundScope
     /// <remarks>
     /// Two lookup paths are tried, in order:
     /// <list type="number">
-    /// <item><description><b>Exact-match fast path</b> — reference-equality on the
-    /// declared receiver type. This is sufficient for non-generic receivers
-    /// (<c>(self string)</c>, <c>(self int32)</c>) and for already-closed
-    /// receiver spellings.</description></item>
+    /// <item><description><b>Exact / structural receiver match</b> — a
+    /// reference-equality fast path (sufficient for primitive singletons and
+    /// already-interned imported symbols) backed by a STRUCTURAL CLR-type
+    /// comparison (issue #1103) so an imported/BCL or primitive receiver still
+    /// matches when the declaration-site and call-site symbols are distinct
+    /// <see cref="ImportedTypeSymbol"/> instances wrapping the same CLR type.
+    /// See <see cref="ExtensionReceiverMatches"/>.</description></item>
     /// <item><description><b>Generic-receiver unification (issue #773)</b> —
     /// an extension whose receiver type references its own function-level
     /// type parameters (e.g. <c>func (self sequence[T]) FirstOrNil[T]() T?</c>
@@ -290,7 +293,7 @@ public sealed class BoundScope
         {
             foreach (var ext in extensionFunctions)
             {
-                if (ext.Name == name && ext.ExtensionReceiverType == receiverType)
+                if (ext.Name == name && ExtensionReceiverMatches(ext.ExtensionReceiverType, receiverType))
                 {
                     function = ext;
                     return true;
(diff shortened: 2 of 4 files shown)


DavidObando:
Superseded. Issue #1103 was fixed in parallel by #1109 (merged as 9d9d2b60), which independently landed the same two-part fix this PR contains: (1) `CreateParentScope` re-registers flattened extension functions via `TryDeclareExtensionFunction`, and (2) `TryLookupExtensionFunction` matches receivers structurally via `ClrTypeUtilities.IsSameAs` instead of reference equality. Hence the conflict against main.

The only delta in this PR is an additional `ImportedTypeArgumentsMatch` refinement for *erased* constructed-generic receivers (e.g. `List[UserType]`); that case is narrow, was not actually exercised by this PR's tests (the closed-generic test used `List[int32]`, which already works on main), and is not a demonstrated cs2gs blocker — the issue's ~435 GS0159 diagnostics were all concrete imported/primitive receivers, which #1109 fully resolves. Closing as a duplicate to avoid a conflicting change. If the erased-generic-receiver edge case proves to be a real blocker, it can be filed and fixed separately on top of current main.

PR #1117, opened 2026-06-25
Switch-expression best-common-type across arms (#1112)

### LINKED ISSUE #1112: switch-expression requires identical arm types; no best-common-type / target-typing (GS0179)
## Summary

A G# **switch-expression** requires every arm to produce the **exact same** type. It infers the result type from the **first arm** and rejects any later arm whose type differs, even when all arms share a common base class / interface (a "best common type") and/or the expression is assigned/returned into a wider target type. C# computes a *best common type* across the arms and also supports *target-typing*, so idiomatic factory/dispatch code fails to migrate.

Diagnostic: `GS0179: All switch-expression arms must produce the same type; expected '<FirstArmType>' but arm produces '<OtherArmType>'`.

Found migrating Oahu via cs2gs (issue #914); a single MPEG-4 `BoxFactory` switch (one arm per box fourcc, each a distinct `*Box : Box` subtype) produces ~60 `GS0179` diagnostics.

## Minimal repro

```gsharp
package p

open class Base {
    open func Name() string { return "base" }
}
class A : Base {
    override func Name() string { return "a" }
}
class B : Base {
    override func Name() string { return "b" }
}

class Factory {
    // target type is Base (declared local)
    func PickLet(s string) Base {
        let box = switch s {
            case "a": A()
            case "b": B()      // GS0179: expected 'A' but arm produces 'B'
            default: A()
        }
        return box
    }
    // target type is Base (return type)
    func PickReturn(s string) Base {
        return switch s {
            case "a": A()
            case "b": B()      // GS0179
            default: A()
        }
    }
}
```

Both fail today. Each arm produces a distinct subtype of `Base`; the natural result type is `Base`.

## Expected

The switch-expression's result type should be the **best common type** of its arm types (the least-upper-bound across the arms — a shared base class or interface), and/or honor the **target type** when the result is directly assigned to a typed local, returned from a typed function, or passed as a typed argument (as C# target-typing does). In the repro, both methods should compile with the switch typed as `Base`.

At minimum, the common-base-class case (all arms derive from a shared `Base`, expression assigned/returned as `Base`) must work. Ideally a shared interface and target-typing are handled too. Arms with no common supertype other than `object` should still behave like C# (either infer `object`/the target type, or report a clear error if no target).

DavidObando:
## Summary

A G# switch-expression previously inferred its result type from the **first** arm (`boundArmBuilders[0].Result.Type`) and rejected any later arm whose type differed with `GS0179`, even when all arms shared a common base class or interface.

This replaces the "first arm wins" logic with a **best-common-type (least-upper-bound)** computation across **all** arm result types.

`Fixes #1112`

## Best-common-type rules (`ComputeSwitchResultType`)

1. **Identity** — all arms already share one type (current behavior preserved).
2. **Pairwise dominance** — an arm type to which every other arm implicitly converts. Preserves numeric widening (with the existing numeric tie-break) and the case where one arm literally **is** the shared base.
3. **Shared supertype (the core enhancement)** — the most-derived class or interface, enumerated from a user-type arm's transitive base classes + transitive/inherited interfaces (most-derived first), to which **every** arm implicitly converts.
4. **`never` / `nil` sentinels** — a `never`-typed (throw-expression) arm does not constrain the common type and is left unwrapped at emit; a `nil` arm only requires a reference/nullable result.

**`object` / `System.ValueType` exclusion:** these are never a valid common type, so unrelated arms (e.g. `string` + `int32`) still have no common type and keep reporting `GS0179`.

All convertibility uses `Conversion.Classify`, keeping reference/interface/numeric rules consistent with the rest of the binder.

## Target-typing (implemented)

In addition to LUB, the declared target type is threaded into the switch so that, when **no** best-common-type exists but every arm converts to the target, the arms are target-typed:

- `let x T = switch …` (variable declaration with an explicit type)
- `return switch …` in a function with a declared (non-inferred) return type
- any target-typed expression position (`BindExpression(syntax, targetType)`)

The two repro methods compile under **LUB alone** (`PickReturn`'s arms LUB to `Base`; `PickLet` infers `Base` then `return box` is `Base`→`Base`); target-typing broadens coverage (e.g. `let x object = switch … { string / int }`).

## Files changed

- `src/Core/CodeAnalysis/Binding/ExpressionBinder.SwitchExpr.cs` — LUB result-type computation + helpers (`ComputeSwitchResultType`, `AllConvertTo`, `AllArmsConvertTo`, `EnumerateSupertypes`, `IsObjectOrValueType`); per-arm loop leaves `never` arms unwrapped; new target-typed overload.
- `src/Core/CodeAnalysis/Binding/ExpressionBinder.cs` — thread the target type into a switch-expression in the target-typed `BindExpression` overload.
- `src/Core/CodeAnalysis/Binding/StatementBinder.cs` — target-type a switch in `let x T = …` and `return switch …`.
- `docs/diagnostics.md`, `website/docs/ref/diagnostics.md` — refine the `GS0179` description.
- `website/docs/ref/spec.md` — document the switch-expression best-common-type rule.
- `test/Core.Tests/CodeAnalysis/Binding/SwitchExpressionTests.cs` — new binding tests.
- `test/Compiler.Tests/Emit/Issue1112SwitchBestCommonTypeEmitTests.cs` — new end-to-end emit tests.

## Tests added

Binding: sibling subtypes → base, shared interface → interface, arm-is-base, numeric widening, target-typing to `object`, `never` arm does not constrain, the full repro (PickLet + PickReturn) compiles clean, and the preserved `string` + `int32` → `GS0179` negative.

Emit (compile + ilverify + run): sibling subtypes dispatch via the base, shared-interface dispatch, `never`-arm program, and an ilverify-only library.

## Verification

- `dotnet test test/Core.Tests/Core.Tests.csproj -c Release --no-build --filter "FullyQualifiedName~SwitchExpression"` → 84 passed.
- Full `Core.Tests` → 3858 passed, 1 skipped, 0 failed.
- `dotnet test test/Compiler.Tests/Compiler.Tests.csproj -c Release --no-build --filter "FullyQualifiedName~Issue1112" -- xUnit.MaxParallelThreads=2` → 4 passed.
- Emit sanity batches (Switch/Conditional/IfExpression/Throw → 63 passed; Return/Variable/Inheritance/Polymorph → 199 passed).
- The must-preserve `string` + `int32` `GS0179` test still passes; the two repro methods compile clean.

Target-typing was implemented in addition to best-common-type (LUB).

### DIFF
--- a/docs/diagnostics.md
+++ b/docs/diagnostics.md
@@ -128,7 +128,7 @@ IDs may be given as `GS0001`, `0001`, or the bare integer `1`; all three forms a
 | GS0176 | Error | Switch expression is missing a `default` arm. | A `switch` expression that cannot be proven exhaustive and has no `default`. |
 | GS0177 | Error | Switch expression on enum is not exhaustive. | One or more enum members not covered and no `default` arm. |
 | GS0178 | Error | Switch statement on enum is not exhaustive. | One or more enum members not covered and no `default` arm. |
-| GS0179 | Error | Switch expression arm type mismatch. | Different arms of a `switch` expression produce incompatible types. |
+| GS0179 | Error | Switch expression arm type mismatch. | The arms of a `switch` expression have no best-common-type (shared base class/interface) and no usable target type — e.g. unrelated types like `string` + `int32`. |
 | GS0180 | Error | Accessibility modifier not allowed here. | `pub` or `priv` used on a local variable or inside a function body. |
 | GS0181 | Error | Base class is not open. | Inheriting from a class that was not declared `open`. |
 | GS0182 | Error | Method is overridable; `override` required. | Redefining an `open` method without the `override` keyword. |
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.SwitchExpr.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.SwitchExpr.cs
@@ -25,6 +25,9 @@ namespace GSharp.Core.CodeAnalysis.Binding;
 internal sealed partial class ExpressionBinder
 {
     private BoundExpression BindSwitchExpression(SwitchExpressionSyntax syntax)
+        => BindSwitchExpression(syntax, targetType: null);
+
+    private BoundExpression BindSwitchExpression(SwitchExpressionSyntax syntax, TypeSymbol targetType)
     {
         var discriminant = BindExpression(syntax.Expression);
         var switchType = discriminant.Type;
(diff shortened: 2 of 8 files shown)


DavidObando:
Superseded by #1116, which was merged in parallel and implements the same switch-expression best-common-type (LUB) + target-typing fix (with the same `object`/`ValueType` exclusion that keeps unrelated arms erroring as GS0179). Issue #1112 is already closed as completed by #1116. Closing this duplicate; the merged fix fully covers the issue and its tests. No salvageable net-new behavior here (the code/tests are equivalent, and main's GS0179 docs already read "produce incompatible types," which remains accurate under best-common-type).

Issue #700, opened 2026-06-11
Consider adding Kotlin-style smart cast


DavidObando:
In Kotlin, when a variable is determined to be casteable to a child type, it gains that type in the rest of the enclosing scope.
For example:
```kotlin
if animal is Dog {
    animal.Bark()
}
```

or:
```kotlin
if animal !is Dog { return }
animal.Bark()
```

We should analyze how feasible this is to implement in G#. If feasibility is confirmed, this task should produce:
1. An ADR
2. Full implementation of the feature
3. Enough test coverage

PR #1126, opened 2026-06-25
Smart-cast nullable local to non-null after non-null assignment (#1123)

### LINKED ISSUE #1123: Smart-cast: nullable var should narrow to non-null after assignment of a non-nullable value (Kotlin-style)
## Summary

G# applies Kotlin-style smart casts after type tests (`if x is T { … }`) and nil-guards, but it does **not** smart-cast a nullable local (`var x T?`) to its non-nullable type `T` after the local is **assigned a non-nullable value**. After `x = <non-null T>`, a subsequent `x.Member` / `x.Method()` fails as if `x` were still `T?`.

Kotlin (which G# follows for nullability/smart-cast semantics) narrows `x` to non-null after such an assignment, so the member access is legal. G# should do the same.

Found migrating Oahu via cs2gs (issue #914): C# pattern variables that leak past an `if (x is T t)` block are commonly reassigned to a fresh non-null value and then used; the faithful G# lowering hoists `var t T? = …` and reassigns `t = <non-null>`, which then cannot call members on `t`.

## Minimal repro

```gsharp
package p
class E { func M() int32 { return 1 } }
class C {
    func F(fresh E) int32 {
        var x E? = nil
        x = fresh        // assign a non-nullable E
        return x.M()     // GS0159: Cannot find function M  (x still treated as E?)
    }
}
```

## Expected

After `x = fresh` (RHS statically non-nullable `E`), `x` should be smart-cast to `E` for subsequent reads until it is reassigned to a possibly-null value or otherwise invalidated. `x.M()` should bind. This mirrors the existing `if x is T { … }` smart cast, extended to assignment.

## Scope / semantics (match Kotlin)

- Narrow `x` to the RHS's non-nullable type after `x = <non-null>`.
- Invalidate the narrowing if `x` is later assigned a possibly-null value, captured by a closure/lambda that may mutate it, or passed `ref`/`out` (whatever the existing smart-cast invalidation rules already cover for `var`).
- Applies to nullable **reference** types (and ideally nullable value types `T?` → `T`).
- The narrowing should hold across straight-line code and into nested blocks until invalidated (same reach as the existing smart-cast flow analysis).

## Notes

- This extends the existing smart-cast flow analysis (the same machinery behind `if x is T`/nil-guards) to the assignment case.
- Likely lives wherever flow-based smart-cast narrowing is computed in the binder.

DavidObando:
Fixes #1123

## Root cause
G# applied Kotlin-style smart casts after `is` type tests and nil-guards, but **not** after a nullable local (`var x T?`) is *assigned* a statically non-null value. After `x = <non-null T>`, a subsequent `x.Member` / `x.Method()` failed with `GS0159` as if `x` were still `T?`.

The per-statement flow loop in `BindBlockStatements` (`StatementBinder.cs`) only ever *dropped* `x`'s narrowing on assignment (`InvalidateNarrowingsForAssignedVariables`); it never re-added a non-null narrowing.

## Fix
Added a new `ApplyAssignmentNarrowings` step, wired **immediately after** `InvalidateNarrowingsForAssignedVariables`:

1. invalidation runs first -> clears any stale narrowing for the assigned variable,
2. the new step re-adds a narrowing **only when the assigned value is statically non-null**.

So reassignment to a nullable / `nil` value naturally invalidates the narrowing (cleared and not re-added).

`NarrowAssignedVariableIfNonNull` narrows when:
- the target is a narrowable local/parameter (ADR-0069 stability rules — `LocalVariableSymbol`/`ParameterSymbol` only),
- its declared type is a `NullableTypeSymbol` (`T?`),
- the assigned value is statically non-null (not nullable, not the `nil` sentinel, not error/void/never) — recovered by peeling the implicit `T -> T?` conversion that `BindAssignmentExpression` inserts, and
- the value's non-null type is implicitly convertible to the declared underlying type.

The narrowed type is the value's **own static non-null type** (Kotlin behaviour), so assigning a `Dog` to an `Animal?` local lets `Dog`-only members bind. The narrowing lives in the persistent block frame, so it reaches across straight-line code and into nested blocks until invalidated. The multi-assignment lowering (`x, y = a, b`) is also covered.

## Scope: reference types (hard requirement) + value types intentionally scoped out
Value-type nullable narrowing (`int32? -> int32`) is **intentionally excluded**. The narrowed-read emit path (`EmitNarrowingCastIfNeeded`) treats a value-type nullable strip as a representation no-op, which leaves a `Nullable<T>` on the stack where a `T` is expected when the read flows into arithmetic/return — producing unverifiable IL (a runtime bus error). I verified this is a **pre-existing** emit limitation: the existing `if x != nil { return n + 1 }` value-type narrowing crashes the same way today. Rather than introduce unsound IL, `IsReferenceLikeNarrowTarget` restricts assignment narrowing to reference-like targets (class/interface), where a narrowed read is a representation no-op or a verifiable `castclass`. A binder test documents the value-type behaviour (`n` stays `int32?`).

## Tests
**Binder (`Core.Tests` — `Issue1123AssignmentSmartCastBinderTests`):**
- repro narrows (`var x E? = nil; x = fresh; return x.M()`),
- narrowing to the assigned **derived** type (`Animal?` <- `Dog`, calls `Bark()`),
- invalidation on `nil` reassignment and on reassignment to another nullable value,
- narrowing available before the invalidating reassignment,
- reach into a nested block and across straight-line statements,
- multi-assignment (`x, y = a, b`),
- value-type scope-out.

**Emit (`Compiler.Tests` — `Issue1123AssignmentSmartCastEmitTests`):** IL-verifies + runs the repro (returns 42), the derived-type case (`Rex:woof`), and the nested-block reach (42).

## Validation
- `dotnet build GSharp.sln -c Release` — clean (0 warnings/errors).
- Repro now emits with no `GS0159` and returns 1/42 at runtime; `x = fresh; x = nil; x.M()` still reports the nullable-member error.
- `Core.Tests`: full project green (3873 passed). Smart-cast/narrowing filter (incl. #700/#708/#712/new #1123): 106 passed.
- `Compiler.Tests` emit: `Issue1123` 3 passed; smart-cast/narrowing regression (`Issue700SmartCast`/`Issue712FlowNarrowing`/`Issue708`): 18 passed.

No new `SyntaxKind`/`BoundNodeKind` and no new sample file — binder-only change, so no coverage-matrix/baseline regeneration needed.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/StatementBinder.cs
+++ b/src/Core/CodeAnalysis/Binding/StatementBinder.cs
@@ -354,6 +354,17 @@ internal sealed class StatementBinder
                 // narrowing from the current frame so subsequent reads in this
                 // block see the variable at its declared (nullable) type again.
                 InvalidateNarrowingsForAssignedVariables(statementSyntax);
+
+                // Issue #1123: Kotlin-style assignment smart cast. Runs AFTER
+                // invalidation so the stale narrowing for the assigned variable
+                // is first cleared, then re-added only when the assigned value
+                // is statically non-null. This means a reassignment to a
+                // nullable/`nil` value naturally invalidates the narrowing
+                // (cleared and not re-added), while `x = <non-null T>` narrows
+                // `x` to the assigned value's static non-null type for
+                // subsequent reads in this block (and nested blocks) until it is
+                // reassigned or otherwise invalidated.
+                ApplyAssignmentNarrowings(statement, memberNotNullFrame);
             }
         }
         finally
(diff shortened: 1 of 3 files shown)


DavidObando:
Superseded by #1125, a functionally-equivalent parallel fix for #1123 (same approach: an assignment-narrowing pass after the existing invalidation step, narrowing a nullable var-local to the RHS's pre-conversion non-null type in the persistent block frame; value-type nullable scoped out for the same pre-existing narrowed-read emit limitation). #1125 is mergeable with equivalent binder + emit test coverage, so closing this duplicate to avoid a double-merge. No salvageable net-new behavior here.

PR #1127, opened 2026-06-25
Fix #1124: exclude uninferable generic overload from applicable set

### LINKED ISSUE #1124: Overload resolution: generic overload with uninferable type parameter should be excluded, not cause GS0266 ambiguity
## Summary

When a method group has both a **generic** overload whose type parameter cannot be inferred from the arguments (the type parameter appears only in the return type / a constraint, not in any parameter) and a **non-generic** overload with a matching parameter list, calling it **without explicit type arguments** is reported as ambiguous (`GS0266`) instead of resolving to the non-generic overload.

C# excludes the generic candidate from the applicable set because **type inference fails** for it (no way to infer `T`), leaving the non-generic overload as the unique best match. G# keeps both candidates and reports ambiguity.

Found migrating Oahu via cs2gs (issue #914): the MPEG-4 `BoxFactory` exposes `T CreateBox<T>(Stream, IBox?) where T : Box` alongside `IBox CreateBox(Stream, IBox?)`; the common call `CreateBox(file, this)` (no explicit `T`) is ambiguous in G# but compiles in C#.

## Minimal repro

```gsharp
package p
interface IBox {}
class Box : IBox {}
class Factory {
    shared {
        func Make[T Box](file int32, parent IBox?) T { return default(T) }
        func Make(file int32, parent IBox?) IBox { return Box() }
    }
}
class C {
    func G(b IBox) {
        let child = Factory.Make(5, b)   // GS0266: ambiguous
    }
}
```

Expected: resolves to the non-generic `Make(int32, IBox?) IBox` (the generic `Make[T]` is not applicable because `T` cannot be inferred and no explicit type argument was supplied). `Factory.Make[Box](5, b)` should still pick the generic overload.

## Expected behavior

During overload resolution, when a candidate is a generic method invoked **without** explicit type arguments and its type parameter(s) **cannot be inferred** from the supplied arguments, that candidate must be removed from the applicable-candidate set (inference failure ⇒ not applicable), exactly as C# does. Remaining non-generic (or successfully-inferred) candidates then resolve normally; ambiguity is only reported among genuinely applicable candidates.

## Notes

- Likely in the overload-resolution / applicability stage: a generic candidate with failed type-argument inference should be discarded before the ambiguity check, not carried into it.
- Calls WITH explicit type arguments (`Make[Box](…)`) should continue to select the generic overload.

DavidObando:
## Problem

When a method group has both a **generic** overload whose type parameter cannot be inferred from the arguments (the type parameter appears only in the return type / a constraint, never in a parameter) and a **non-generic** overload with a matching parameter list, calling it **without explicit type arguments** was reported as ambiguous (GS0266) instead of resolving to the non-generic overload.

Minimal repro (previously `GS0266`, now compiles clean):

```gsharp
package p
interface IBox {}
class Box : IBox {}
class Factory {
    shared {
        func Make[T Box](file int32, parent IBox?) T { return default(T) }
        func Make(file int32, parent IBox?) IBox { return Box() }
    }
}
class C {
    func G(b IBox) {
        let child = Factory.Make(5, b)   // now resolves to non-generic Make(int32, IBox?) IBox
    }
}
```

## Root cause

`OverloadResolver.SelectBestUserOverloadCore` Phase 1 applicability (`IsApplicableUserCallable`) only checked arity + named-argument compatibility. It never considered whether a generic candidate's type parameters could be inferred, so both `Make[T]` and `Make` survived to Phase 2, tied on score, and were reported as ambiguous — whereas C# excludes the generic candidate because type inference fails for it.

## Fix

Centralized in `SelectBestUserOverloadCore` (covers static, instance, and top-level user-method overload groups). After Phase 1, when more than one candidate is applicable, apply the C# rule:

- **No explicit type arguments**: drop generic candidates whose type parameters cannot **all** be inferred from the supplied argument types (new helper `AllTypeParametersInferable`, reusing `Binder.InferTypeArguments`).
- **Explicit type arguments present** (`Make[Box](…)`): drop non-generic candidates (they cannot accept type arguments), leaving the generic overload to resolve. (This also fixes a pre-existing `GS0266` on the explicit-arg form.)

The exclusion is **guarded** so it only applies when at least one candidate survives — preserving the existing single-candidate diagnostics, e.g. the type-argument-inference-failure `GS0151` for a *lone* uninferable generic with no sibling.

A `hasExplicitTypeArgs` signal is threaded from the call sites (`ce`/`syntax.TypeArgumentList != null`) through `SelectInstanceOverloadOrReport`, `SelectBestInstanceOverload`, `SelectBestUserOverload`, and the core (added with a safe default so existing callers compile).

## Tests

- **`Issue1124UninferableGenericOverloadBinderTests`** (binder): no-explicit-arg resolves to non-generic with no diagnostics; return-type distinguishing checks (result is `IBox`, not `Box`, and not `GS0266`); explicit `[Box]` selects the generic; an inferable generic (`Id[T](x T) T`) is still selectable; a lone uninferable generic still reports `GS0151`; top-level-function path.
- **`Issue1124UninferableGenericOverloadEmitTests`** (emit): end-to-end compile + ilverify + run for both the non-generic and explicit-generic resolutions.

## Validation

- `dotnet build GSharp.sln -c Release` — clean (0 warnings, 0 errors).
- New binder tests: 8/8 passed. New emit tests: 2/2 passed.
- Regression slices: Core.Tests `~Overload|~Generic` 396/396, `~Call|~Method|~Static|~Function` 746/746; Compiler.Tests `~Generic` 179/179, `~Overload|~GenericMethod|~StaticMethod` 72/72 — all passed.

Fixes #1124

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
@@ -1518,7 +1518,7 @@ internal sealed partial class ExpressionBinder
         var staticMethodGroup = TypeMemberModel.GetMethods(structSym, methodName, MemberQuery.Static(MemberKinds.Method));
         if (!staticMethodGroup.IsDefaultOrEmpty)
         {
-            var method = overloads.SelectInstanceOverloadOrReport(staticMethodGroup, arguments, ce, methodName, argumentNames: default);
+            var method = overloads.SelectInstanceOverloadOrReport(staticMethodGroup, arguments, ce, methodName, argumentNames: default, hasExplicitTypeArgs: ce.TypeArgumentList != null);
             if (method == null)
             {
                 return new BoundErrorExpression(null);
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.cs
@@ -2435,7 +2435,7 @@ internal sealed partial class ExpressionBinder
 
                 if (ifaceOverloads.Length > 0)
                 {
-                    var ifaceMethod = overloads.SelectInstanceOverloadOrReport(ifaceOverloads, arguments, ce, methodName, argumentNames);
+                    var ifaceMethod = overloads.SelectInstanceOverloadOrReport(ifaceOverloads, arguments, ce, methodName, argumentNames, ce.TypeArgumentList != null);
                     if (ifaceMethod == null)
                     {
                         return new BoundErrorExpression(null);
@@ -2456,7 +2456,7 @@ internal sealed partial class ExpressionBinder
                 var tpOverloads = TypeMemberModel.GetMethods(tpRecv.InterfaceConstraint, methodName, MemberQuery.Instance(MemberKinds.Method));
                 if (tpOverloads.Length > 0)
                 {
-                    var tpIfaceMethod = overloads.SelectInstanceOverloadOrReport(tpOverloads, arguments, ce, methodName, argumentNames);
+                    var tpIfaceMethod = overloads.SelectInstanceOverloadOrReport(tpOverloads, arguments, ce, methodName, argumentNames, ce.TypeArgumentList != null);
                     if (tpIfaceMethod == null)
                     {
                         return new BoundErrorExpression(null);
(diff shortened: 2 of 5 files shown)


DavidObando:
Superseded by #1128, which fixes #1124 with an equivalent but more surgical approach (centralized entirely in OverloadResolver.cs, reusing the existing inferTypeArguments delegate and reading ce.TypeArgumentList directly, without threading a flag through the ExpressionBinder call sites). Closing this duplicate and landing #1128.

Issue #914, opened 2026-06-21
C# to G# migration tool


DavidObando:
Consider alternatives:
- Hand rolled parser and AST mapper, then prettyprint
- Roslyn analyzer to build G# AST, then prettyprint
- other options 

Objectives:
- Transform existing C# applications into G# projects
- Discover gaps in the G# compiler

This project requires planning and tooling. We will not simply produce G# code and leave it at that, instead we need to have the code generated by the migration tool passed over to a G# pipeline that will:
- Compile the program
- Collect compilation issues
- Pass the output to an LLM so it can file issues in the G# repository (then retry from the start with an updated compiler version)
- Verify the IL produced by the compiler, file issues if anything arises
- Once compilation and verification pass cleanly, run unit tests, ensure parity with original C#
- Output report of the pipeline in a distributable format

The process must be accurate and repeatable. We expect that code should be able to migrate in a straightforward way, the syntax in G# is now strong enough to allow it. The generated G# output is expected to be canonical, and should compile to verifiable IL. Once unit test verification shows parity, we can say a migration has been completed successfully.

PR #1135, opened 2026-06-25
Fix GS0127 on let-ref-local field write + out-var enclosing scope (#914)

### LINKED ISSUE #1132: GS0127 wrongly rejects field/property assignment through a 'let' reference-type local
## Summary

Assigning to a **field/property of a reference-type object** held by a `let` local is incorrectly rejected with `GS0127: Variable 'b' is read-only and cannot be assigned to`. Per the spec, "`let` communicates immutability of the **binding**" (spec.md §Declarations) — i.e. shallow, Kotlin-`val`/C#-`readonly`-reference semantics. For a reference-type local, `b.Field = x` mutates the **heap object**, not the binding, so it must be allowed. Only the binding itself (`b = other`) should be blocked, and field mutation of a **value-type** `let` local should remain blocked (it would mutate the value stored in the read-only slot).

Found migrating Oahu via cs2gs (issue #914): e.g. `ChunkOffsetList list = new(...); list.chunkOffsets64 = new(...);` — `list` is a class, never reassigned, so the translator (correctly) emits `let list`, but the field write is then rejected. ~several GS0127 across `ChunkOffsetList`, `DashFile`, etc.

## Minimal repro

```gsharp
package p
class Box { var Value int32 = 0 }
class C {
    func F() int32 {
        let b = Box{ }
        b.Value = 5        // GS0127: Variable 'b' is read-only and cannot be assigned to.
        return b.Value
    }
}
```

`var b` compiles; `let b` should too (reference type). Contrast — a **struct** receiver must stay rejected:

```gsharp
struct S { var Value int32 }
// let s = S{ Value: 1 }; s.Value = 5   // correctly rejected (value semantics)
```

## Expected

`let b = <reference-type>; b.Field = v` (and `b.Prop = v`) compiles, mutating the heap object. `let b = <value-type>; b.Field = v` stays rejected (GS0127). Direct rebinding `b = other` stays rejected for both.

## Pointer

`src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs:632`:
```csharp
if (variable.IsReadOnly && !receiverIsThisField)
{
    Diagnostics.ReportCannotAssign(syntax.EqualsToken.Location, receiverName);
}
```
This fires for any read-only receiver variable. It should additionally require the receiver's type to be a **value type** (e.g. `StructSymbol { IsClass: false }`, an imported value type, or a value-type type parameter); reference-type receivers (class/interface/imported reference type) should be exempt. The analogous compound-assignment / increment member paths in the same file likely need the same guard.

### LINKED ISSUE #1133: Inline 'out var n' declaration not visible in enclosing scope (GS0125)
## Summary

An inline `out var n` (and `out let n` / `out _`) declaration is accepted at the call site but the declared local is **not visible in the enclosing scope** afterward — using it fails with `GS0125: Variable 'n' doesn't exist`. The spec lists `f(out var n)` as a supported ref-kind argument form (spec.md §"Ref-kind arguments", ADR-0060, line ~661), so the variable should leak into the enclosing block exactly like C#.

Found migrating Oahu via cs2gs (issue #914): `SkipToFirstMoof(out var moofBox, out var mdatBox, out var startSample)` then using/`reassigning` those variables later — produces a cascade of `GS0125` (e.g. `moofBox`, `mdatBox`, `startSample`).

## Minimal repro

```gsharp
package p
class C {
    func G(out x int32) { x = 5 }
    func F() int32 {
        G(out var y)
        return y          // GS0125: Variable 'y' doesn't exist.
    }
}
```

Control (pre-declared) compiles fine:
```gsharp
var y int32
G(out y)
return y                  // OK
```

## Expected

`G(out var y)` declares `y` (typed from the matched `out` parameter) in the **enclosing** scope; subsequent reads/writes of `y` bind. `out let y` declares a read-only local; `out _` is a discard.

## Root-cause lead

`src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs:1094` `BindRefArgumentExpression`. On the **first** pass (before overload resolution) it intentionally returns a placeholder *without* declaring the local (lines 1117-1125, parameter unknown). The local is only declared on the **re-bind** after the parameter is resolved (`scope.TryDeclareVariable(local)`, line 1142). The re-bind appears to run against a transient/argument scope rather than the enclosing statement's block scope, so the declared local never becomes visible at later statements. The fix should ensure the inline-declared out-var/out-let local is declared into the **enclosing block scope** (the scope that contains the call statement), matching C# out-var scoping.

DavidObando:
## Summary

Two related G# binder fixes.

### #1132 — GS0127 wrongly rejected field/property writes through a `let` reference-type local
Per spec, `let` makes the **binding** read-only (shallow, Kotlin `val` / C# `readonly`-reference semantics), not the heap object. For a reference-type receiver (`class`/interface/imported reference type), `b.Field = x` mutates the heap object, not the binding, so it must be allowed; only rebinding `b = other` is blocked. For a value-type receiver (`struct`/enum/value-type type parameter), a field write through a `let` stays rejected (it would mutate the value in the read-only slot).

A new `IsValueTypeReceiver(TypeSymbol)` helper is added and the value-type-only guard is applied to every member-write path that previously rejected purely on receiver read-onlyness:
- simple `b.Field = x` (`BindFieldAssignmentExpression`)
- compound `b.Field += 1` (`TryBindChainedCompoundAssignment`) — this path previously *allowed* struct field writes too, which is now correctly rejected
- chained member `a.b.Field = x` (`BindMemberFieldAssignmentExpression`)

Postfix `b.Field++` is covered because it desugars to the simple-assignment path.

### #1133 — inline `out var n` not visible in enclosing scope (GS0125)
An inline `out var n` / `out let n` / `out _` declaration on a user-defined **instance** method call (implicit or explicit `this`) now leaks the new local into the enclosing block scope (like C#). Previously the call site was accepted but a later use failed with `GS0125: Variable 'y' doesn't exist`.

`BindUserInstanceCall` now re-binds the inline out-var placeholder against the resolved by-ref parameter (mirroring the imported-method `RebindInlineOutVarArguments` and free-function fix-up paths), declaring the local with the inferred pointee type into the active block scope.

## Tests
- `test/Core.Tests/.../Issue1132LetRefLocalFieldWriteBindingTests.cs` (8 tests)
- `test/Core.Tests/.../Issue1133OutVarScopeBindingTests.cs` (7 tests)
- `test/Compiler.Tests/Emit/Issue1132LetRefLocalFieldWriteEmitTests.cs` (3 tests, asserts runtime mutation)
- `test/Compiler.Tests/Emit/Issue1133OutVarScopeEmitTests.cs` (6 tests, asserts out-var receives value)

Full `Core.Tests` green (3901 passed). Targeted emit + related Compiler.Tests (FieldAssignment, RefKind, out-var, ADR-0060/0112, struct, mutation, LanguageConformance) all green.

### Known limitation (out of scope)
Prefix `++b.Value` as a standalone statement still fails with GS0402 — but this is a pre-existing parser newline-absorption bug that affects `var` locals identically (`++i` on a `var` local fails the same way), unrelated to the `let`/readonly binder fix. Postfix `b.Value++` and compound `b.Value += 1` work correctly.

Fixes #1132
Fixes #1133
Refs #914

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
@@ -1742,6 +1742,25 @@ internal sealed partial class ExpressionBinder
             for (var i = 0; i < permutedArgs.Length; i++)
             {
                 var paramType = method.Parameters[i].Type;
+
+                // ADR-0060 / issue #1133: an inline `out var n` / `out let n` /
+                // `out _` argument was bound to a placeholder address-of (Error
+                // pointee) in the eager argument pass, because the resolved
+                // parameter — and therefore the local's type — was not yet
+                // known. Now that the static (`shared`) overload has been
+                // chosen, re-bind the inline declaration against the resolved
+                // by-ref parameter so the synthesized local is declared (with
+                // the correct pointee type) into the enclosing block scope and
+                // becomes visible to the rest of the block — mirroring the
+                // instance-method path in OverloadResolver.BindUserInstanceCall.
+                var outVarParamType = substitution != null ? Binder.SubstituteType(paramType, substitution) : paramType;
+                var argSyntaxForOutVar = i < ce.Arguments.Count ? ce.Arguments[i] : null;
+                if (TryRebindInlineOutVarUserArgument(permutedArgs[i], argSyntaxForOutVar, method.Parameters[i].Name, outVarParamType, out var reboundOutVar))
+                {
+                    convertedArgs.Add(reboundOutVar);
+                    continue;
+                }
+
                 if (paramType is TypeParameterSymbol)
                 {
                     convertedArgs.Add(permutedArgs[i]);
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs
@@ -1149,6 +1149,48 @@ internal sealed partial class ExpressionBinder
     }
 
     /// <summary>
+    /// ADR-0060 / issue #1133: re-binds a single inline <c>out var n</c> /
+    /// <c>out let n</c> / <c>out _</c> argument whose declared type was omitted,
+    /// once the resolved by-ref parameter type is known. In the eager argument
+    /// pass such an argument is bound to a placeholder
+    /// <see cref="BoundAddressOfExpression"/> over an <see cref="BoundErrorExpression"/>
+    /// (Error pointee) without declaring a local, because the parameter — and
+    /// therefore the local's pointee type — was not yet known. This re-binds the
+    /// declaration against a synthesized <see cref="RefKind.Out"/> parameter so
+    /// the local is declared into the enclosing block scope with the correct
+    /// type. Shared by the user instance-call, free-function, and qualified
+    /// static-call paths so the inline-out-var scoping is identical everywhere.
+    /// </summary>
+    /// <param name="boundArgument">The already-bound argument (possibly a placeholder).</param>
+    /// <param name="argumentSyntax">The source argument syntax (named-argument wrappers are unwrapped).</param>
+    /// <param name="parameterName">The resolved parameter's name (used for the synthesized parameter symbol).</param>
+    /// <param name="parameterType">The resolved parameter's (pointee) type, used as the local's type.</param>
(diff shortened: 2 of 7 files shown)


DavidObando:
Superseded: #1132 was independently fixed and merged via #1134. Re-landing the remaining #1133 (out-var enclosing scope) work on a fresh branch off current main to avoid the duplicate-#1132 conflict.

PR #1138, opened 2026-06-25
Fix #1133: inline out-var declaration visible in enclosing scope

### LINKED ISSUE #1133: Inline 'out var n' declaration not visible in enclosing scope (GS0125)

DavidObando:
Fixes #1133

Refs #914

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
@@ -1742,6 +1742,25 @@ internal sealed partial class ExpressionBinder
             for (var i = 0; i < permutedArgs.Length; i++)
             {
                 var paramType = method.Parameters[i].Type;
+
+                // ADR-0060 / issue #1133: an inline `out var n` / `out let n` /
+                // `out _` argument was bound to a placeholder address-of (Error
+                // pointee) in the eager argument pass, because the resolved
+                // parameter — and therefore the local's type — was not yet
+                // known. Now that the static (`shared`) overload has been
+                // chosen, re-bind the inline declaration against the resolved
+                // by-ref parameter so the synthesized local is declared (with
+                // the correct pointee type) into the enclosing block scope and
+                // becomes visible to the rest of the block — mirroring the
+                // instance-method path in OverloadResolver.BindUserInstanceCall.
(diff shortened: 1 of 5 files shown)


DavidObando:
Superseded by #1137 (which landed the instance-call out-var fix on main first). This branch was cut before #1137 and would revert it. The remaining gap — inline out-var on QUALIFIED STATIC calls (`C.G(out var y)`) — is still unfixed on main; filing a focused follow-up issue + PR on top of #1137.

PR #1141, opened 2026-06-25
Fix inherited System.Object members + static out-var scope on user types (#914)

### LINKED ISSUE #1136: Inherited System.Object members (GetType/ToString/GetHashCode/Equals) not callable on user class/struct instances
## Summary

Inherited `System.Object` instance members — `GetType()`, `ToString()`, `GetHashCode()`, `Equals(object)` — are **not callable on user `class`/`struct` instances** when the user type does not itself declare an override. Both `this.M()` and `receiver.M()` fail with `GS0159: Cannot find function <M>` (and a bare `GetType()` fails with `GS0130: Function 'GetType' doesn't exist`). Every .NET reference type inherits these from `System.Object`, so they must resolve.

Found migrating Oahu via cs2gs (issue #914): `GetType().Name`, `obj.ToString()`, `x.GetHashCode()`, `a.Equals(b)` are ubiquitous and currently fail across the project (driving `GS0130`/`GS0159` and downstream `GS0158`/cascade errors).

## Minimal repro

```gsharp
package p
class C {
    func F() int32 {
        let h = this.GetHashCode()   // GS0159: Cannot find function GetHashCode.
        let s = this.ToString()      // GS0159: Cannot find function ToString.
        let t = this.GetType()       // GS0159: Cannot find function GetType.
        let e = this.Equals(this)    // GS0159: Cannot find function Equals.
        return h
    }
}
```

Also on a local receiver:
```gsharp
class C { }
class D {
    func F() {
        let c = C{ }
        let t = c.GetType()          // GS0159
        let s = c.ToString()         // GS0159
    }
}
```

And a bare (implicit-this) call:
```gsharp
class C { func F() string { return GetType().Name } }  // GS0130: Function 'GetType' doesn't exist.
```

(Contrast: a bare call to a user-declared sibling instance method — `H()` → `this.H()` — already works, so implicit-this resolution itself is fine; it just doesn't consult the inherited `System.Object` members.)

## Expected

Member/function lookup on a user reference type (and value type, which inherits from `System.ValueType`/`System.Object`) should fall back to the inherited `System.Object` members `GetType()`, `ToString()`, `GetHashCode()`, `Equals(object)` (and `ReferenceEquals`/`MemberwiseClone` where applicable) when the user type does not declare its own. `obj.GetType()` returns `System.Type`; `obj.ToString()` returns `string`; `obj.GetHashCode()` returns `int32`; `obj.Equals(other)` returns `bool`. Emit should bind to the corresponding `System.Object` virtual (callvirt) so user overrides dispatch correctly. A bare `GetType()` inside an instance method resolves as `this.GetType()`.

## Pointer

User-type instance member lookup goes through `TypeMemberModel` (ADR-0112) and the `StructSymbol`/`InterfaceSymbol` member-resolution path in `ExpressionBinder` (call path `ExpressionBinder.Calls.cs` / `ExpressionBinder.Access.cs`). The fallback that surfaces an imported CLR base's members (e.g. the `structSym.ImportedBaseType?.ClrType` fallback already used for inherited base properties/fields, ExpressionBinder.Access.cs ~1994 and ExpressionBinder.Assignments.cs ~590) should be extended so that a user type with NO explicit imported base still consults `typeof(object)` (and `typeof(System.ValueType)`/`object` for structs) for these universally-inherited members, for both call and method-group positions.

### LINKED ISSUE #1139: Inline out-var not visible in enclosing scope for qualified static calls (follow-up to #1133)
## Summary

Follow-up to #1133 (fixed for instance calls by #1137). An inline `out var n` / `out let n` / `out _` declaration on a **qualified static** method call still drops the local — using it fails with `GS0125: Variable 'n' doesn't exist`. Instance calls and free-function calls now work; only the static / `shared`-method call path was not covered.

## Minimal repro (fails on current main)

```gsharp
package p
class C {
    shared {
        func G(out x int32) { x = 5 }
        func F() int32 {
            C.G(out var y)     // GS0125: Variable 'y' doesn't exist.
            return y
        }
    }
}
```

Instance form (works since #1137):
```gsharp
class C { func G(out x int32) { x = 5 }  func F() int32 { G(out var y); return y } }   // OK
```

## Expected

Qualified static calls (`C.G(out var y)`, `Other.G(out var y)`) declare the inline out-var into the enclosing block scope exactly like the instance path. `out let` → read-only local, `out _` → discard.

## Pointer

The instance path (#1137) re-binds the inline out-var placeholder against the resolved by-ref parameter and declares the local into the enclosing scope. The qualified/static-call binding site — `BindUserTypeStaticCall` in `src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs` (the per-position argument conversion loop) — needs the same re-bind step. Factor a shared helper next to `BindRefArgumentExpression` (`ExpressionBinder.Assignments.cs` ~1094) and call it from both the instance and static paths.

Note: a BARE sibling-static call `G(out var y)` (no `C.` qualifier) is a separate pre-existing limitation (bare sibling static calls fail with GS0130 even without out params) and is out of scope here.

DavidObando:
Fixes #1136

Fixes #1139

Refs #914

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
@@ -1742,6 +1742,33 @@ internal sealed partial class ExpressionBinder
             for (var i = 0; i < permutedArgs.Length; i++)
             {
                 var paramType = method.Parameters[i].Type;
+
+                // ADR-0060 / issue #1139: an inline-decl `out var n` /
+                // `out let n` / `out _` at a qualified static (`C.G(...)`) call
+                // site was bound with TypeSymbol.Error in the first pass
+                // (before the static overload was resolved) and never declared
+                // a local — the same gap #1137 fixed for user instance calls.
+                // Now that the method is selected (and any generic substitution
+                // is known) re-bind it so the synthesized local is typed from
+                // the resolved (substituted) out-parameter pointee type and
+                // leaks into the enclosing block scope. Must run BEFORE the
+                // open-type-parameter shortcut below so generic out-parameters
+                // (`func G[T](out result T)`) are handled too.
+                if (permutedArgs[i] is BoundAddressOfExpression inlineOutAddr
+                    && inlineOutAddr.Operand.Type == TypeSymbol.Error
+                    && i < ce.Arguments.Count
+                    && OverloadResolver.UnwrapNamedArgumentValue(ce.Arguments[i]) is RefArgumentExpressionSyntax inlineOutRefArg
+                    && inlineOutRefArg.IsInlineDeclaration
+                    && inlineOutRefArg.DeclaredType == null)
+                {
+                    var pointeeType = substitution != null ? Binder.SubstituteType(paramType, substitution) : paramType;
+                    var rebindParameter = ReferenceEquals(pointeeType, method.Parameters[i].Type)
+                        ? method.Parameters[i]
+                        : new ParameterSymbol(method.Parameters[i].Name, pointeeType, refKind: RefKind.Out);
+                    convertedArgs.Add(BindRefArgumentExpression(inlineOutRefArg, rebindParameter));
+                    continue;
+                }
+
                 if (paramType is TypeParameterSymbol)
                 {
                     convertedArgs.Add(permutedArgs[i]);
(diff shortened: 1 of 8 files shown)


DavidObando:
Superseded: #1136 was independently fixed by #1140 (merged). Re-landing only the still-open #1139 (qualified static out-var) on a fresh branch off updated main.

PR #1143, opened 2026-06-25
Fix #1139: inline out-var visible in enclosing scope for qualified static calls

### LINKED ISSUE #1139: Inline out-var not visible in enclosing scope for qualified static calls (follow-up to #1133)

DavidObando:
## Summary

Fixes #1139 (follow-up to #1133 / ADR-0060).

An inline-decl `out var n` / `out let n` / `out _` at a **qualified static** call site (`C.G(out var n)`) was bound with `TypeSymbol.Error` on the first pass — before the static overload was resolved — and never declared the local in the enclosing block scope, so a subsequent read of `n` failed with `GS0125`. PR #1137 fixed this for user **instance** calls (`BindUserInstanceCall`); this lands the same treatment for the qualified-static path (`BindUserTypeStaticCall`).

## Change

`src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs` — `BindUserTypeStaticCall`'s per-position conversion loop now detects a first-pass inline out-var placeholder (`BoundAddressOfExpression` whose operand is `TypeSymbol.Error`, backed by a `RefArgumentExpressionSyntax` inline declaration with no explicit declared type) and re-binds it via `BindRefArgumentExpression` once the static overload is selected. The synthesized local is typed from the resolved (substituted) out-parameter pointee type and leaks into the enclosing block scope. The re-bind runs **before** the open-type-parameter shortcut so generic out-parameters (`func M[T](..., out result T)`) are handled too.

## Tests

- **`Issue1139StaticOutVarBinderTests`** (8): `out var` in an `if` and as a simple statement; `out let` read-only local (and `GS0127` when reassigned); `out _` discard (compiles alone; `GS0125` when read); generic out-parameter substituted type; pre-declared-local control.
- **`Issue1139StaticOutVarEmitTests`** (2): end-to-end compile + IL-verify + run, asserting the leaked local carries the right value (13; generic 21).

## Validation

- `dotnet build GSharp.sln -c Release -graph` → 0 Warning(s) 0 Error(s).
- Targeted tests: binder 8 passed, emit 2 passed.
- Standalone gsc repros: `C.G(out var y)` → "Wrote x.dll"; `out _` then read `_` → `GS0125` (correct); user `#1136` Object-member fix (#1140, already on main) still resolves.

Refs #914

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
@@ -1742,6 +1742,33 @@ internal sealed partial class ExpressionBinder
             for (var i = 0; i < permutedArgs.Length; i++)
             {
                 var paramType = method.Parameters[i].Type;
+
+                // ADR-0060 / issue #1139: an inline-decl `out var n` /
+                // `out let n` / `out _` at a qualified static (`C.G(...)`) call
+                // site was bound with TypeSymbol.Error in the first pass
+                // (before the static overload was resolved) and never declared
+                // a local — the same gap #1137 fixed for user instance calls.
+                // Now that the method is selected (and any generic substitution
+                // is known) re-bind it so the synthesized local is typed from
+                // the resolved (substituted) out-parameter pointee type and
+                // leaks into the enclosing block scope. Must run BEFORE the
+                // open-type-parameter shortcut below so generic out-parameters
+                // (`func G[T](out result T)`) are handled too.
+                if (permutedArgs[i] is BoundAddressOfExpression inlineOutAddr
+                    && inlineOutAddr.Operand.Type == TypeSymbol.Error
+                    && i < ce.Arguments.Count
+                    && OverloadResolver.UnwrapNamedArgumentValue(ce.Arguments[i]) is RefArgumentExpressionSyntax inlineOutRefArg
+                    && inlineOutRefArg.IsInlineDeclaration
+                    && inlineOutRefArg.DeclaredType == null)
+                {
+                    var pointeeType = substitution != null ? Binder.SubstituteType(paramType, substitution) : paramType;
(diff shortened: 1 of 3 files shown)


DavidObando:
Superseded: #1139 was independently fixed by #1142 (merged). My implementation matched the same BindUserTypeStaticCall re-bind approach.

Issue #1154, opened 2026-06-25
Overload resolution: spurious GS0266 ambiguity when an argument needs T→T? nullable widening and a non-applicable overload exists


DavidObando:
### Summary

Overload resolution reports a spurious `GS0266` "ambiguous between multiple overloads" when an argument needs an **implicit nullable-widening conversion** (`T` → `T?`) to match the *only* applicable candidate, and a second, **wholly inapplicable** overload also exists. The nullable-widening candidate is the unique applicable overload and should be selected unambiguously.

### Minimal repro

```gsharp
package p
class C {
    shared { func Make() []uint8 { var r []uint8 return r } }
    func F(a string) { F(C.Make()) }     // arg is []uint8 (non-nullable)
    func F(a []uint8?) { var n = 1 }
}
```

```
error GS0266: Call to 'F' is ambiguous between multiple overloads.
              Disambiguate with explicit types or named arguments.
```

Here `C.Make()` returns a non-nullable `[]uint8`. Only `F([]uint8?)` is applicable — `[]uint8 → []uint8?` is a standard implicit nullable widening. `F(string)` is **not** applicable (`[]uint8` does not convert to `string`). There is no genuine ambiguity, yet gsc reports `GS0266`.

### Isolation matrix (same call shape, one variable changed)

| Variant | Overloads | Argument | Result |
|--------|-----------|----------|--------|
| A | `F(string)`, `F([]uint8)` | `[]uint8` | ✅ compiles (exact match) |
| B | `F([]uint8?)` only | `[]uint8` | ✅ compiles (nullable widening applicable) |
| C | `F(string)`, `F([]uint8?)` | `[]uint8?` | ✅ compiles (exact match to `T?`) |
| **Repro** | `F(string)`, `F([]uint8?)` | `[]uint8` | ❌ **GS0266 ambiguous** |

B proves the `T → T?` widening *is* applicable in isolation. A and C prove the same two-candidate shape is fine when the argument matches a parameter exactly. Only the combination "argument needs `T → T?` widening **and** a non-applicable competitor overload exists" trips the false ambiguity. This points at the overload **betterness / applicability** step: either `[]uint8` is being treated as (partially) convertible to the inapplicable candidate, or the nullable-widening candidate is not ranked strictly better than a non-applicable one.

### Expected

[...]

Issue #751, opened 2026-06-12
Parser: receiver clauses reject nullable and generic type spellings ((self T?), (self sequence[T]))


DavidObando:
Discovered while authoring **Gsharp.Extensions** for issue #724 / ADR-0084.

## Problem

`LooksLikeReceiverClause()` in `src/Core/CodeAnalysis/Syntax/Parser.cs` only accepts a receiver clause whose type is a bare identifier or a `[N]T` / `[]T` array form. Common surfaces in the planned helper namespaces require nullable and generic spellings that the parser silently rejects, so the receiver clause is reparsed as a regular parameter list and the function loses its extension-method status:

```gs
func (self T?) Map[T, U](f (T) -> U) U?         // T? rejected
func (self sequence[T]) FirstOrNil[T]() T?   // sequence[T] rejected
func (self sequence[T]) Indexed[T]() sequence[(int32, T)]
```

## Repro

Any G# file containing the surface above today fails to bind `Map` / `FirstOrNil` as extension methods on `T?` / `sequence[T]`.

## Fix sketch

Extend `LooksLikeReceiverClause()` (and the corresponding production in `ParseReceiverClause`) to accept:
- Nullable type spelling: any of the existing accepted forms followed by `?`.
- Generic application: `Ident[`...`]` where `...` is a comma-separated type-arg list.
- Combinations (`sequence[T]?`).

## Why this matters for #724 / ADR-0084

ADR-0084 §Known Limitations lists this as **L2**. While L1 (constraint-aware overload resolution) is open, the Optional and Sequences namespaces ship as C# under the documented escape hatch. Closing **L2** unblocks the dogfooded G# rewrite once **L1** is closed (or once the language uses a different mechanism for the constraint split).

## Related
- Parent: #706
- ADR: ADR-0084
- Companion: #750 (binder constraints), then dogfooding can land
- Surfaces affected: all `Gsharp.Extensions.*` extension methods

Issue #710, opened 2026-06-11
Add `a?[i]` null-conditional indexing


DavidObando:
## Summary

Add the null-conditional indexing operator `a?[i]`. Returns the indexed value when `a` is non-null and the index is valid, or `nil` when `a` is `nil`.

## Motivation

G# already has null-conditional member access (`a?.member`). Indexing has no equivalent today, so users have to write `a == nil ? nil : a[i]` or `let a2 = a; a2 != nil ? a2[i] : nil`. C#, Kotlin, and Swift all have this operator.

## Scope

- Lexer / parser: recognize `?[` as the prefix of null-conditional indexing.
- Binder: result type is the nullable form of the indexer's return type.
- Lowering: rewrite to `(a == nil) ? nil : a[i]` with proper sequencing of `a` evaluation.
- Apply to arrays, slices, maps, and CLR indexers.

## Acceptance criteria

- Works on emit and interpreter.
- Diagnostic when receiver is non-nullable (use `[]` instead).
- Sample + golden coverage.
- Spec updated.

---

**Plan reference:** §6.2 item 6. Tracked under #706 (G# Language — Current State and Design Opportunities, §6).

PR #1220, opened 2026-06-26
Bind `?` on an array clause to the element type (Fixes #1212)

### LINKED ISSUE #1212: G# cannot express an array of nullable elements (C# T?[]); []T? means a nullable array
### Summary
Indexing an array whose element type is **nullable** (`[]T?`) is rejected with `GS0116: Type '[]T?' is not indexable`. Both nullable reference-element and nullable value-element arrays are affected. The non-nullable-element array `[]T` indexes fine.

### Repro
```gsharp
package p
func F(a []object?) object? { return a[0] }   // GS0116: Type '[]object?' is not indexable
func G(a []int32?)  int32?  { return a[0] }    // GS0116: Type '[]int32?'  is not indexable
func H(a []object)  object  { return a[0] }    // compiles
```

### Expected
`a[i]` on an `[]T?` should bind as element access yielding `T?`, exactly as `[]T` yields `T`. The element being nullable is orthogonal to the array's indexability.

A faithful solution is a coordinated change:

- Compiler: adopt `[]T?` = array of nullable elements, `[]?T` = nullable array. A `[]?T?` would be a valid nullable array of nullable elements.
- cs2gs: emit `[]?T` for C# `T[]?`/nullable-array fields and `[]T?` for C# `T?[]`.

### Impact (Oahu.Decrypt)
~5 direct GS0116 (`[]object?`, `[]IEnumerator[T]?` ×3) plus cascade GS0159 (`MoveNext`/`Current` on the un-indexable result). ~8 errors.

### Area
Binder — index-expression element-type resolution: it unwraps the array type but rejects a nullable element symbol instead of yielding `T?`.

### Environment
gsc at main (b95400fe). `bug`.

DavidObando:
Fixes #1212.

## Root cause
Indexing an array/slice whose element type is nullable (`[]T?`, `[N]T?`) was rejected with `GS0116: Type '[]T?' is not indexable`. When binding a type clause, the trailing `?` on an array clause was applied to the **whole array** (producing `Nullable(Slice(T))`), not to the element. The index-expression element-type resolver (`GetIndexElementType`) unwraps slice/array symbols, so it saw a `NullableTypeSymbol` wrapper instead of a slice/array and returned `null` → GS0116. Both nullable reference-element (`[]object?`) and nullable value-element (`[]int32?`) arrays were affected; `[]T` indexed fine.

## The fix — indexability of nullable element types
- **`?` on an array/slice clause binds to the element.** `ApplyArraySuffix` now wraps the element in `NullableTypeSymbol` before building the slice/array, yielding `Slice(Nullable(T))` / `Array(Nullable(T))`. `BindTypeClause` no longer double-wraps the whole array for array clauses. The element being nullable is now orthogonal to the array's indexability.
- **Result/element type.** `GetIndexElementType` returns the slice/array element, now `T?`, so `a[i]` **reads** `T?` and **writes** `T?` — exactly as `[]T` reads/writes `T`.
- **Whole-nullable-array guard preserved.** A genuinely nullable array *reference* (the whole array possibly nil) is spelled `([]T)?` and still reports `GS0116` on a bare `a[i]`, requiring `!!` (or `?[`) first. A single parenthesized type `(T)` is now treated as a **grouping** (it previously errored as a 1-tuple), so `([]T)?` parses and binds to `Nullable(Slice(T))`.
- **Backing CLR type.** `SliceTypeSymbol`/`ArrayTypeSymbol` build their array CLR type from the *effective* element type, so `[]int32?` is `Nullable<int32>[]` (holds nil value elements) while `[]object?` stays `object[]`.
- **Literals.** The composite-array-literal parser accepts a trailing `?` on the element (`[]T?{ … }`).

## Emit round-trip
Indexing `[]T?` emits the same `ldelem`/`stelem` as `[]T`, with the element type being the nullable representation: `Nullable<int32>` (a struct → typed `ldelem`/`stelem`) for value elements, and `object` (`ldelem.ref`/`stelem.ref`) for reference elements. A new emit/execution test builds `[]int32?` and `[]object?`, writes elements (including a nil and a value), reads them back, and asserts the values round-trip; IL is verified with `IlVerifier.Verify`.

## Read + write
- read: `func F(a []object?) object? { return a[0] }`, `func G(a []int32?) int32? { return a[0] }`
- write: `a[0] = x` where `a : []T?`, `x : T?`
- `([]T)?` still needs `!!`: bare `a[0]` → GS0116; `a!![0]` compiles.

## Tests
- **Binder** (`Issue1212NullableElementArrayIndexTests`, 8 tests): `[]T?` indexable yielding `T?` (ref + value), non-nullable `[]T` still works, read result is nullable, indexed write accepts `T?`, `[N]T?` indexable, `([]T)?` needs `!!`.
- **Emit/execution** (`Issue1212NullableElementArrayIndexEmitTests`, 3 tests): nullable value- and reference-element slices write/read round-trip, including writing nil.
- Migrated existing nullable-*slice* usages (#710 / #751 / #1154 tests + `samples/NullConditionalIndexing.gs`) to the `([]T)?` spelling.
- Full Core.Tests (4252), Extensions.Tests (104), Interpreter.Tests (321), narrowed Compiler.Tests (`Array|Index|Nullable|Issue1212`, 364), and `GSharp.sln -graph` all green.

### DIFF
--- a/docs/adr/0084-gsharp-extensions-optional-sequences.md
+++ b/docs/adr/0084-gsharp-extensions-optional-sequences.md
@@ -215,7 +215,7 @@ hatch under `src/Sdk/Gsharp.Extensions/*.cs` works today.
   combinations are recognised by the parser and the function
   retains its extension-method status. The binder + emit pipeline
   also accept call-site dispatch when the receiver is *closed*
-  (concrete) — `string?`, `(int32, string)`, `[]int32?`,
+  (concrete) — `string?`, `(int32, string)`, `([]int32)?`,
   `map[string,int32]`. **Binder-side closed by
   [issue #773](https://github.com/DavidObando/gsharp/issues/773).**
   `BoundScope.TryLookupExtensionFunction` now falls back to
--- a/samples/NullConditionalIndexing.gs
+++ b/samples/NullConditionalIndexing.gs
@@ -12,12 +12,12 @@ import System.Collections.Generic
 
 var calls int32 = 0
 
-func bump() []int32? {
+func bump() ([]int32)? {
     calls = calls + 1
     return []int32{7, 8, 9}
 }
 
-func nilBump() []int32? {
+func nilBump() ([]int32)? {
     calls = calls + 1
     return nil
 }
@@ -25,11 +25,11 @@ func nilBump() []int32? {
 func main() {
     // 1. Slice receiver. The first read indexes a live slice; the second
     // reads through a nil slice and short-circuits to nil.
-    var live []int32? = []int32{10, 20, 30}
+    var live ([]int32)? = []int32{10, 20, 30}
     var first = live?[1]
     Console.WriteLine(first)
 
-    var missing []int32? = nil
+    var missing ([]int32)? = nil
     var firstMissing = missing?[0]
     if firstMissing == nil {
         Console.WriteLine("nil-slice")
--- a/src/Core/CodeAnalysis/Binding/Binder.cs
+++ b/src/Core/CodeAnalysis/Binding/Binder.cs
@@ -2148,6 +2148,16 @@ public sealed class Binder
         if (syntax.IsTuple)
         {
             // Phase 4.5: tuple type clause `(T1, T2, ...)`.
+            // Issue #1212: a *single* parenthesized type `(T)` is a grouping,
+            // not a tuple. It exists so a nullable array reference can be
+            // spelled `([]T)?` — the parentheses bind the array shape before
+            // the outer `?` (applied by BindTypeClause) wraps the whole array
+            // in a NullableTypeSymbol. Bind the lone element and return it.
+            if (syntax.TupleElements.Count == 1)
+            {
+                return BindTypeClause(syntax.TupleElements[0]);
+            }
+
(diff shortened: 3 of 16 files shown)


DavidObando:
Closing without merging. This change is sound in isolation but flips the meaning of `[]T?` from a *nullable array* `([]T)?` (the current, by-design G# semantics) to an *array of nullable elements*. cs2gs emits `[]T?` for nullable arrays throughout the corpus, so this regresses Oahu.Decrypt (~200 → 239 errors; GS0129 +18) because every nullable-array site is reinterpreted as element-nullable.

The real gap (see #1212 update): G# can express a nullable array `([]T)?` but has **no** spelling for C#'s `T?[]` (array of nullable elements), which is what `IEnumerator<T>?[]` in InterleavedIterator needs. A faithful fix is a *coordinated* compiler+translator change: adopt `[]T?` = element-nullable / `([]T)?` = array-nullable (this PR's compiler half) AND update cs2gs to emit `([]T)?` for nullable arrays. Deferring — low error count (~5-8) vs. high cross-corpus regression risk.

Issue #1354, opened 2026-06-28
Revert #1349 and improve gsc/cs2gs type resolution


DavidObando:
PR #1349 introduced the capability of comparing references of type `T` against `nil`. This leads to milder/softer nullability guarantees in G#, which weakens one of the language's superpowers: Kotlin-like nullability semantics.

The right approach should be:
- Revert the diff introduced by #1349 
- In gsc, BCL/imported types should only be represented as type `T` when they have a `System.Runtime.CompilerServices.NullableAttribute` that indicates it's non-nullable. In the absence of this attribute, G# will represent this as `T?`.
- In gsc, the emitter must always produce the `System.Runtime.CompilerServices.NullableAttribute` for all applicable places, with the right byte array values.
- In cs2gs, the same approach applies: if Roslyn indicates a field of type `T` is non-nullable and there are no `null` spillovers (null forgiveness, comparisons against null, etc.) then the field can be of type `T` in G#. Otherwise, it should be type `T?` in G#.

PR #1374, opened 2026-06-28
Revert #1349; treat unannotated imported reference types as nullable + emit complete nullability metadata (#1354)

DavidObando:
## Summary

G# has Kotlin-style nullability: a non-nullable reference type `T` cannot be compared to `nil` (compile error **GS0129**); only `T?` can be null-checked. PR #1349 weakened this by allowing `ref == nil` / `case nil` on a non-nullable reference. This P0 change (issue #1354) reverses #1349 and makes the type system treat **unannotated/oblivious imported reference types as nullable by default**, so the non-null guarantee is sound again. It also makes the emitter produce **complete** nullability metadata so gsc→gsc round-trips preserve member nullability.

## Part 1 — Revert PR #1349
Restored `BoundBinaryOperator.cs`, `PatternBinder.cs`, and `MethodBodyEmitter.Operators.cs` to their pre-#1349 state and deleted `Issue1333RefNullCompareTests.cs`. `== nil` / `!= nil` / `case nil` against a non-nullable reference reports **GS0129** again (verified: `func F(t Task) bool { return t == nil }` → GS0129).

## Part 2 — Imported-type nullability default (core change)
`ClrNullability` now reads a reference-type position at DFS index `i` through a shared `IsPositionNonNull(ImmutableArray<byte> flags, int index)`:

| `flags` shape | non-null iff |
| --- | --- |
| **empty** (no `[Nullable]`/`[NullableContext]`) | never — position is `T?` |
| **length 1** (scalar `[Nullable(b)]` / `[NullableContext(b)]`, applies to all positions) | `flags[0] == 1` |
| **length > 1** (per-position) | `flags[i] == 1` |

Only an explicit `1` (NotAnnotated) means non-null `T`; `2` (Annotated), `0` (oblivious), and absent all mean `T?`. Applied uniformly in `ApplyReferenceNullabilityFull` (field/property/param/return) and `SymbolFromFlagsOffset` (inner generic args; for-range and CLR indexer). Value types are unaffected. The length-1 case fixes scalar/context expansion. Modern BCL assemblies carry `[NullableContext(1)]`, so annotated BCL types stay non-null.

## Part 3 — Emit completeness + round-trip guarantee
Because "absent attribute → nullable" is now the rule, the emitter stamps complete metadata (matching Roslyn):
1. Type-level `[NullableContextAttribute(1)]` on every emitted user type (the importer's type-chain walk relies on this).
2. Per-**field** `[NullableAttribute(flags)]` when DFS flags are non-empty and not all-`1`.
3. Per-**property** `[NullableAttribute(flags)]` under the same condition.
4. Parameters/returns unchanged.

The assembly-level `[NullableContext]` emitter now reuses the cached ctor MemberRef (dedup invariant). A binder fix preserves inner reference nullability for fully-concrete generic args (e.g. `List[string?]`, which previously collapsed `string?` → `string`) by attaching the DFS flags as a `NullabilityAnnotatedTypeSymbol`. **Round-trip test** (`Issue1354NullabilityRoundTripEmitTests`, 5/5) compiles a library with a non-null field/property, a `T?` field/property, and a `List[string?]` field, emits it, re-loads via `MetadataLoadContext`, and asserts each re-reads with the correct nullability.

## Part 4 — ADR + tests + fallout
- **ADR-0136** (`docs/adr/0136-nullable-by-default-for-unannotated-imports.md`): rationale, the revert, the 3-case import rule + scalar expansion, emit-completeness, value-type non-impact, deferred positions.
- **`ClrNullabilityTests`** updated/added (24/24): oblivious → nullable; context-1 → non-null; flag-2 → nullable; scalar expansion; `IsPositionNonNull` + `SymbolFromFlagsOffset` cases.
- **Refactoring baseline** regenerated (127 hashes) via the Skip-gated `RegenerateBaseline` fact — emit now adds type-level context + per-field/property attributes, which legitimately changes IL.

### Pre-existing tests updated (and why)
The import-default flip correctly makes oblivious test fixtures import as nullable. Updates:
- **Annotated 3 Core.Tests fixture files** (`Issue666ItemCls`, `Handler322Extensions`, `InterpolatedStringHandlerFixtures`) with `#nullable enable` so they represent realistic modern (annotated) referenced assemblies — these tests verify LINQ/extension/interpolation binding, not nullability.
- **Added `<Nullable>enable</Nullable>` to the generated sibling-C# csproj** in 31 Compiler.Tests emit harnesses (slice/array/interface/struct/property interop, etc.).
- **`Issue520ClrArrayForRangeTests`**: its helper is built via Reflection.Emit (genuinely oblivious — exactly the #1354 case), so `Helper.GetColors()` is now `Color[]?`; the G# source null-forgives (`!!`).
- **`ImportedEnumAttributeEmitTests`**: the type now also carries the injected `[NullableContext(1)]`, so the test selects the fixture attribute by its ctor's declaring type instead of asserting a single attribute.

## Test results
- `dotnet build GSharp.sln -c Release -graph` — **0 warnings, 0 errors**.
- `dotnet test test/Core.Tests` — **4732/4732 pass**.
- `test/Compiler.Tests` (batched, `xUnit.MaxParallelThreads=2`): Nullable/Null/Equality 248; Pattern/Switch/Conversion; Generic/Field/Property 535; Import/Metadata/Issue1354 55; Indexer/ForRange/Linq/Foreach/ForIn 177; Struct/Record/Tuple/Delegate 323; Interface/Async/Closure/Lambda 389; plus all 31 patched emit harnesses — **all green**.

## Deferred (best-effort, documented in ADR-0136)
- Base-type / implemented-interface generic-argument nullability.
- Generic-constraint nullability.

These affect only generic arguments in a type's base/interface list or constraints; member-signature surface round-trips correctly today.

Refs #1354
Refs #914


DavidObando:
## Follow-up: imported instance method-call return types now honor nullability

The first commit fixed imported fields, properties, parameters, and **static** call returns. This follow-up closes the remaining gap: imported **instance** method-call return types bypassed `ClrNullability`.

### Fix
- New helper `MapClrMethodReturnType(MethodInfo)` in `ExpressionBinder.Access.cs`: preserves by-ref-return handling, otherwise delegates to `ClrNullability.GetReturnTypeSymbol` (honors oblivious default + `[Nullable]`/`[NullableContext]`).
- Replaced the terminal fallback at all three imported call-return sites in `ExpressionBinder.Calls.cs`: non-generic instance (`~3028`), inherited/`base.M(...)` (`~3696`), constraint-interface (`~5171`). Earlier generic/symbolic arms are unchanged; static-call path untouched (no double-wrap).

### Generic-substituted returns — deferred
When an earlier arm recovers a generic/symbolic projection (`ResolveInstanceReturnTypeFromReceiver` / `MapOpenClrTypeToSymbolic`, e.g. `Queue[Entry].Dequeue()`), per-position nullability of the substituted argument is not yet applied — documented in ADR-0136 as best-effort follow-up. The common non-generic instance case is fully fixed.

### Notable consequence
`System.Object.ToString()` is genuinely annotated `string?`, so `base.ToString()` / `obj.ToString()` on a reference receiver now correctly bind as `string?`. A few tests that relied on the old buggy non-null reading were updated to coalesce.

### Manual repro (all as expected)
| Case | Before | After |
| --- | --- | --- |
| oblivious instance `f.Make() == nil` | GS0129 | **compiles** ✅ |
| source non-null param `w == nil` | GS0129 | GS0129 ✅ |
| annotated non-null `f2.Make2() == nil` | — | GS0129 ✅ |
| annotated nullable `f2.Make3() == nil` | — | **compiles** ✅ |

### Tests added
`Issue1354InstanceMethodReturnEmitTests` (annotated-nullable instance return compiles; annotated-non-null → GS0129). Oblivious-instance case is also exercised by the Core.Tests oblivious fixture (`ImportedTypeInFunctionBodyTests`).

### Verification
- `dotnet build GSharp.sln -c Release -graph` → 0 warnings / 0 errors.
- `Core.Tests`: **4731 passed**, 1 skipped, 0 failed.
- `Compiler.Tests` (batched, `MaxParallelThreads=2`): Nullable 195, Null 239, Equality 13, Call 117, Method 207, Generic 235, Import 38, Metadata 12, Object 44, Slice 82, Struct 264, Enum 149, Base 103, Override 23, ToString 8, Closure 49, Extension 27, Async 80, Lambda 90, Issue1354 7, Issue1260 5 — all green.
- `Interpreter.Tests`: 332 passed.
- Refactoring baseline unchanged (no regeneration needed).

Issue #796, opened 2026-06-13
Allow == / != between function-typed values (and sequence values) and 'nil'


DavidObando:
Discovered while porting `Gsharp.Extensions.Optional` to G# under #792.

## Repro

```gs
func (self T?) OrCompute[T class](factory () -> T) T {
    if factory == nil {                     // GS0129
        throw ArgumentNullException("factory")
    }
    return self ?: factory()
}

func (self sequence[T]) Windowed[T](size int32) sequence[[]T] {
    if self == nil {                        // GS0129
        throw ArgumentNullException("source")
    }
    ...
}
```

Both surface GS0129: `Binary operator '==' is not defined for types '(T) -> T' and 'nil'.` / `...for types 'sequence[T]' and 'nil'.`

## Expected

`==` / `!=` between a function-typed value (`(T) -> R`) and `nil` should be accepted with the obvious CLR semantics (reference equality with the null delegate). Same for `sequence[T]` (CLR null-reference equality on the `IEnumerable<T>` reference).

## Context

The Extensions stdlib needs to guard against `null` delegate / sequence arguments to match the C# `ArgumentNullException` contract documented in the existing tests. Without this comparison the guards can't be expressed. Blocks #792 §L5.

PR #1539, opened 2026-07-01
Define == nil / != nil for all reference-typed shapes (fixes #1535)

### LINKED ISSUE #1535: cs2gs: extension-method receiver not promoted to T? when null-compared in body (GS0129)
## Summary
G# follows strict Kotlin-style nullability: `== nil` / member access is only legal on a nullable type, and a genuinely non-null reference compared to `nil` is (correctly) rejected. Oahu is **nullable-oblivious** (no Nullable enable), so every C# reference is implicitly nullable. cs2gs already promotes ordinary parameters/fields/locals/properties to `T?` when they are null-compared or null-assigned (issue #1072/#1354 via `PromoteIfUsedAsNullable`) — but a C# **extension method** translates to the receiver-clause form (`func (self T) Name()`) through a code path that **bypasses `MapParameters`**, so the receiver was never promoted.

## Repro (Oahu.Foundation ExtensionsVarious)
```csharp
public static bool IsNull(this object o) => o == null;
public static bool IsNullOrEmpty<T>(this IEnumerable<T> e) => e == null || !e.Any();
public static bool StartsBad(this byte[] bytes) => bytes == null || bytes.Length < 8;
```
Before: cs2gs emitted `func (o object) IsNull() -> o == nil` -> gsc `GS0129` (non-null receiver compared to nil).

## Fix (cs2gs)
Apply `PromoteIfUsedAsNullable` to the extension receiver type where the `Receiver` is built in `CSharpToGSharpTranslator` (the extension-method branch of method-declaration translation), exactly as an ordinary parameter is promoted. Generalizes to ANY extension receiver shape (object, user class, imported interface, generic `IEnumerable[T]`, array).

After: `func (o object?) IsNull()`, `func (e IEnumerable[T]?) IsNullOrEmpty[T]()`, `func (bytes []?uint8) StartsBad()`.

## Note
The `|| receiver.Member` guarded-access cases still need gsc `||`/`&&` short-circuit smart-cast narrowing (tracked separately); this issue covers only the receiver-type promotion.

Refs #914

DavidObando:
## Summary
gsc reported `GS0129: Binary operator '==' is not defined for types '<T>' and 'nil'` for `== nil` / `!= nil` on several reference-typed shapes — arrays, slices, user classes, user and imported interfaces, `object`, and imported reference types — even though the comparison already worked for nullable wrappers, function types, delegates, and sequences (issue #796). C# permits `x == null` for any reference or array type.

## Fix
`BoundBinaryOperator.IsNullCompare` now returns true for any reference-shaped type:
- structural shapes: `FunctionTypeSymbol`, `DelegateTypeSymbol`, `SequenceTypeSymbol`, `AsyncSequenceTypeSymbol`, `InterfaceSymbol`;
- user `StructSymbol` only when `IsClass` (classes are reference types; plain / `data` / `inline` structs are not);
- any remaining type whose CLR representation is a managed reference (`ClrType.IsValueType == false`) — arrays (`[N]T`), slices (`[]T`, incl. `[]T?`), `object`, `string`, imported reference/interface types (`IEnumerable[T]`, `SynchronizationContext`, …).

Non-nullable value types (`int32`, user `struct`, `DateTime`, enums) keep their value CLR type and still report `GS0129`. Symmetry (nil on either side) is preserved by the existing caller.

## Emit
Falls through to the uniform reference `ldnull; ceq` path (ilverify-clean). As a side effect, `string == nil` now emits `ceq` instead of `String.op_Equality`, matching C#'s reference-null-check lowering, so the two affected `refactoring-baseline.json` hashes (`DefaultExpression.gs`, `GenericMethodSpec.gs`, both `string == nil`) were regenerated per `test/Core.Tests/Baselines/README.md`.

## Tests
- `test/Core.Tests/.../Issue1535NullCompareReferenceTypesBinderTests.cs` — positive reference cases plus value-type negative controls asserting GS0129 still fires (`int32`, user `struct`, `DateTime`).
- `test/Compiler.Tests/Emit/Issue1535NullCompareReferenceTypesEmitTests.cs` — end-to-end (compile + ilverify + run) for array, nullable-element array, user class, user interface, imported interface (`IEnumerable[T]`), and `object`, covering null and non-null values with both `==` and `!=`.

## Validation
- Full `dotnet build GSharp.sln -c Release -graph`: 0 warnings / 0 errors.
- `Core.Tests`: 4901 passed, 1 skip (RegenerateBaseline).
- Filtered `Compiler.Tests` (`~Issue1535`): 6 passed, ilverify-clean.
- Minimal repro compiles clean; value-type negatives still GS0129.

Fixes #1535

Refs #914


DavidObando:
Closing: this approach is incorrect. G# intentionally rejects `nonNullableRef == nil` (GS0129) as a never-null comparison — see the passing test `AnnotatedNonNullInstanceReturn_ComparesToNil_IsRejected` (Issue1354InstanceMethodReturnEmitTests.cs), which this change regressed. The real root of the Oahu.Foundation `== nil` errors is that cs2gs emits non-nullable G# reference types for nullable-oblivious C# reference types. Re-scoping #1535 to cs2gs nullability mapping.

DavidObando:
Re-opening as a reconsideration on ergonomics and Kotlin-like semantics. Do not merge yet.

DavidObando:
G# is a bit ahead of Kotlin here, it’s handling strong nullability rules in a strict and compliant way, which we’ll keep for now.

Issue #521, opened 2026-06-07
Binder: concrete class cannot be upcast to an implemented interface (GS0155 on both user and BCL types)


DavidObando:
## Summary

A concrete CLR class cannot be assigned or passed where its implemented interface is expected. Both **user-defined** classes implementing **user-defined** interfaces (e.g. `HelloGreeter : IGreeter`) and **BCL** classes implementing **BCL** interfaces (e.g. `List<T> : IReadOnlyList<T>`) hit the same `GS0155: Cannot convert ...` diagnostic.

This is the fundamental CLR subtyping relation. Without it, **dependency-injection patterns are unusable** (factories that return interfaces from concrete services, test fakes assigned to interface-typed variables, `IEnumerable<T>` parameters, …).

## Repro 1 — user types

Sibling C#:
```csharp
namespace Probe.CSharp;
public interface IGreeter { string Greet(string name); }
public sealed class HelloGreeter : IGreeter { public string Greet(string name) => $"Hello, {name}!"; }
```

G#:
```gsharp
package Probe.Tests

import Probe.CSharp
import Xunit

type UpcastProbe class {
    @Fact
    func Concrete_To_Interface() {
        let g IGreeter = HelloGreeter()       // GS0155
        Assert.Equal("Hello, World!", g.Greet("World"))
    }
}
```

## Observed

```
Tests.gs(9,26,9,40): error GS0155: Cannot convert type 'Probe.CSharp.HelloGreeter' to 'Probe.CSharp.IGreeter'.
```

## Repro 2 — BCL types

```gsharp
let mut = List[string]()
mut.Add("a"); mut.Add("b")
let l IReadOnlyList[string] = mut       // same GS0155
```

```
Tests.gs(37,39,37,42): error GS0155: Cannot convert type 'System.Collections.Generic.List`1[...]' to 'System.Collections.Generic.IReadOnlyList`1[...]'.
```

Explicit cast / construction also fails:
```gsharp
var b = IGreeter(HelloGreeter())   // GS0155
```

## Expected

`expr` of static type `C` should be implicitly convertible to `I` whenever `C` (or any of its bases) implements `I` — that's the standard CLR identity-conversion / interface-implementation conversion. Matches C# / VB / F# / Go (via `var i I = c`).

## Context

`gsc` version: `Gsharp.NET.Sdk/0.1.431`, .NET 10 RTM.

[...]

Issue #1255, opened 2026-06-27
Nullable→nullable implicit reference conversion T? → U? rejected (GS0155/GS0154)


DavidObando:
## Summary

A **nullable→nullable implicit reference conversion** `T? → U?` (where `T` derives from / implements `U`) is rejected with **GS0155 / GS0154**, even though both the non-null upcast `T → U?` (fixed in #1121) and `T → U` already work.

## Minimal repro (FAILS today)

```gsharp
package p
interface IBox {}
class AppleListBox : IBox {}
class C {
  func Take(parent IBox?) {}
  func F(x AppleListBox?) { Take(x) }   // GS0155: Cannot convert 'AppleListBox?' to 'IBox?'
}
```

Also fails for class-base hierarchies (`Derived? → Base?`).

## Controls (CORRECT today)
- `AppleListBox → IBox?` (non-null source, nullable target) — **OK** (post-#1121).
- `AppleListBox → IBox` — **OK**.
- `AppleListBox? → IBox` (nullable source, non-null target) — **correctly ERRORS** (narrowing; must stay an error).

So only the **nullable-source + nullable-target** reference upcast is missing.

## Analysis
The conversion classifier handles the implicit reference conversion `T → U` and the lifted `T → U?`, but does not lift it over a **nullable source**: `T? → U?` should be an implicit reference conversion whenever `T → U` is (the null value maps to null; non-null values upcast). This is the nullable-source extension of #1121. The fix belongs in `Conversion.Classify` (`src/Core/CodeAnalysis/Binding/Conversion.cs`) — when the source is nullable and the target is nullable, classify on the underlying types and preserve nullability.

## Impact (Oahu Decrypt)
`FreeformTagBox` (`parent IBox?` ← `AppleListBox?`) and the filter-pipeline tuple registrations (`(TrakBox?, ChapterFilter)`), several GS0154/GS0155.

PR #1553, opened 2026-07-01
Rank nullable reference argument (S?) by underlying type in overload betterness (#1552)

### LINKED ISSUE #1552: gsc: nullable reference argument (S?) causes spurious GS0266 ambiguity between base and derived overloads (betterness ignores the ? wrapper)
## Summary

When an argument has a **nullable reference type** `S?` and two overloads differ only in a parameter that is a base type (`object`, or a base class) versus the more-derived `S`, overload resolution fails to pick the more-specific `S` overload and instead reports **GS0266 (ambiguous)**. With a **non-nullable** `S` argument the same overload pair resolves correctly to `S`. In other words, the "better conversion / more-specific parameter" tie-break does not see through the `?` wrapper on the argument type.

This is common in nullable-oblivious C# translations (Oahu.Foundation), where a nullable **field** (`Type?`, `string?`, a user reference type `T?`) — which is NOT smart-cast narrowed (only locals/params are) — is passed to an overload set that has both an `object` and a `Type` (or base/derived) parameter. It is the sole root of the Foundation GS0266 cluster (~8: `Logging.Log`, `Context`, `ResourceGuard`, `Cleanup`).

## Repro (minimal)

```gsharp
package P
import System
func F(caller object) int32 -> 1
func F(caller Type) int32 -> 2
func UseNN(t Type) int32 -> F(t)     // OK -> selects F(Type)
func UseNull(t Type?) int32 -> F(t)  // GS0266 ambiguous  (should also select F(Type))
func Main() { Console.WriteLine("ok") }
```

Generalizes to user types (not just BCL `object`/`Type`):

```gsharp
open class Animal {}
class Dog : Animal {}
func G(a Animal) int32 -> 1
func G(d Dog) int32 -> 2
func Use(d Dog?) int32 -> G(d)       // GS0266 ambiguous (should select G(Dog))
```

## Expected

A nullable reference argument `S?` must rank candidate parameters the same way its underlying `S` does: the most-derived applicable parameter wins (`Type` over `object`; `Dog` over `Animal`). Passing `S?` where a candidate wants `S` and another wants a base of `S` should select the `S` candidate, exactly as a non-nullable `S` argument does — not report ambiguity.

## Where

`src/Core/CodeAnalysis/Binding/OverloadResolution.cs` betterness / specificity tie-break. `CompareConversions` (~:2688) returns 0 for pure reference conversions and relies on a specificity tie-break to prefer the more-derived parameter; when the argument (`sources[i]`) is a nullable reference type the tie-break does not fire (the two candidates are treated as equally good, yielding GS0266). The nullable wrapper on the argument source type should be stripped to its underlying reference type for the purpose of the derived-vs-base parameter tie-break (both in the `ImplicitConversionKind` classification and in the specificity comparison), so a nullable reference argument behaves like its underlying type for ranking. This is the same code that already resolves the non-nullable case correctly.

Generalize across BCL and user reference types, any base/derived depth, and any arity/position of the differing parameter. Do not regress numeric-widening, interpolated-string, constant-narrowing, or genuinely-ambiguous cases.

Refs #914

DavidObando:
Fixes the spurious GS0266 ambiguity when a nullable reference argument `S?` is passed to an overload set whose parameters differ by base (`object`/base class) vs derived (`S`). Betterness/specificity now sees through the `?` wrapper so a nullable reference argument ranks the most-derived applicable parameter exactly as its non-nullable underlying does.

## Root cause (two coupled gaps)
Top-level user-`func` calls resolve via `OverloadResolver.SelectBestUserOverloadCore` (symbolic), not the CLR `OverloadResolution.cs` path:
1. Ranking: the scorer awards an exact-type bonus only when `argType == paramType`. A nullable `Type?`/`Dog?` never equals a non-nullable param, so both candidates tied -> GS0266.
2. Applicability/conversion: `Conversion.Classify` accepted imported `Type? -> Type` (nullable inherits underlying `ClrType`, #521 rule) but rejected user `Dog? -> Dog`/`Dog? -> Animal` (user classes have null `ClrType` during binding) -> GS0154 after selection.

## Fix
- Added a C# §7.5.3.2 "more-specific reference parameter" tie-break that engages only when an argument is a nullable reference type, treating `S?` as underlying `S` and selecting the uniquely most-derived applicable parameter. Unrelated-interface ties stay ambiguous; `int32?` never engages it.
- Narrowed a nullable reference source to its underlying reference type when classifying against a non-nullable reference target, so user and imported reference nullables narrow identically. Value-type `int32? -> int32` stays explicit.

## Verification
- `dotnet build GSharp.sln -c Release -graph`: 0 warnings, 0 errors.
- Both repros compile, run, and pass ilverify; derived overload selected.
- Core.Tests: 4895 passed, 1 skipped, 0 failed.
- Compiler.Tests targeted (Issue1552|Overload|Issue377|Issue1311|Issue1150|Issue750): 82 passed; new Issue1552 suite: 9 passed; nullable regression sweep (Issue1121|1255|1545|1502|Nullable): 254 passed.

Fixes #1552

Refs #914

### DIFF
--- a/src/Core/CodeAnalysis/Binding/Conversion.cs
+++ b/src/Core/CodeAnalysis/Binding/Conversion.cs
@@ -338,9 +338,12 @@ public sealed class Conversion
                 // non-null value reference-upcasts), so this is a representation-
                 // preserving no-op conversion — never boxing or wrapping. The
                 // value-type underlying case is handled by the lifted numeric
-                // rule above and intentionally excluded here. Note this never
-                // makes `T? → U` (non-nullable target, handled below) implicit,
-                // so the narrowing null-dropping conversion still errors.
+                // rule above and intentionally excluded here. Issue #1552 later
+                // made the sibling `T? → U` (non-nullable reference target) case
+                // implicit too — see the dedicated nullable-reference-source
+                // rule after this `to is NullableTypeSymbol` block — so a
+                // nullable reference argument narrows to its underlying/base as
+                // a representation-preserving reference conversion.
                 if (IsReferenceLikeTarget(fromNullable.UnderlyingType)
                     && IsReferenceLikeTarget(toNullable.UnderlyingType))
                 {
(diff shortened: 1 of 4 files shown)


DavidObando:
Holding — the implemented fix makes `S? -> S` implicit language-wide, which removes reference null-safety (flips the deliberate #1255 test). See detailed analysis on #1552. Branch left pushed for reference; needs rework in the null-safety-preserving direction.

Issue #1940, opened 2026-07-04
gsc: generic local function referencing an enclosing generic method/class type parameter emits invalid IL (no diagnostic)


DavidObando:
Follow-up from PR #1938 (issue #1886, generic function literals).

A generic local function (`let Inner[T] = func (...) ... {...}`) nested inside a **generic** outer method/class that references the OUTER type parameter is bound (LambdaBinder merges enclosing type parameters) but at emit time `Inner` is hoisted to a top-level static method in `<Program>` carrying only its OWN MVAR list. Referencing the enclosing method/class type parameter then produces silently invalid IL or an emit-time crash, with no user-facing diagnostic.

Example shape:
```
func Outer[U]() {
    let Inner[T] = func (x T, y U) T { return x }  // U is the enclosing method's type param
    ...
}
```

GS0463 only covers *value* captures, not enclosing *type-parameter* captures.

## Definition of done
- [ ] Either correctly emit `Inner` closing over the enclosing type parameter, OR report a clear diagnostic (e.g. 'generic local function cannot reference an enclosing method/class type parameter') instead of invalid IL.
- [ ] Regression test.

Discovered during Opus review of PR #1938.

Issue #1477, opened 2026-06-30
Closure capturing an enclosing generic type-parameter value emits a non-generic display class (illegal field type, TypeLoadException)


DavidObando:
## Summary
When a lambda inside a generic type (or generic method) **captures a value whose type references an enclosing generic type parameter** (e.g. captures a `T`-typed parameter, or captures `this` of a `G<T>`), the synthesized display/capture class is emitted **non-generic**, so its capture field is typed `!0` with no generic parameter in scope. This produces an **illegal field type** → `System.TypeLoadException` at load and unverifiable IL (`StackUnexpected [found value 'T'][expected ref '!0']` at the capture store, plus `DelegateCtor` at the delegate construction). gsc reports 0 errors.

## Minimal repro
```gsharp
package CAP
import System
import System.Threading.Tasks

open class Op[T] {
    private let cf (Task) -> T
    init(cf (Task) -> T) { this.cf = cf }
    func Run(t Task) T -> cf(t)
    shared {
        func FromCompleted(result T) Op[T] -> Op[T]((_ Task) -> result)   // lambda captures `result` (type T)
    }
}

func Main() {
    let o = Op[int].FromCompleted(5)
    Console.WriteLine(o.Run(Task.CompletedTask))
}
```
`ilverify`:
```
[StackUnexpected]: CAP.Op`1::FromCompleted(!0) [offset 0x08] [found value 'T'][expected ref '!0']
[DelegateCtor]:    CAP.Op`1::FromCompleted(!0) [offset 0x26] Unrecognized arguments for delegate .ctor.
```
Runtime:
```
System.TypeLoadException: Type 'CAP.<>__Box_result_1' ... has a field of an illegal type.
```

## Root cause (located)
`src/Core/CodeAnalysis/Emit/ClosureEmitter.cs`, `SynthesizeDisplayClass` (~line 327): the display class is created as a **non-generic** `StructSymbol` (`isClass: true`) and each capture field is typed `captured.Type` directly. When `captured.Type` is (or references) an enclosing generic type parameter `T`, the field signature becomes `!0` but the display class declares **no type parameters**, so the field is illegal.

The emitter is already aware that generic enclosers are special — the zero-capture lambda-nesting paths explicitly **skip** when the enclosing type is generic (`ClosureEmitter.cs` ~lines 231-232 and ~280-282: `!enclosing.TypeParameters.IsDefaultOrEmpty`). The capturing path was never generalized to handle it.

[...]

Issue #1469, opened 2026-06-30
Non-capturing lambda hosted on top-level type can't access protected/private members (FieldAccessException)


DavidObando:
## Summary
A **non-capturing (static) lambda** is emitted as a static method on the top-level `<Program>` host type. When such a lambda accesses a **non-public member** (`private`/`protected`) of its parameter's type — e.g. a member of a `protected` nested type, or a private backing field — the emitted `ldfld`/`call` is **outside that member's accessibility domain**, producing a verifiable-IL failure (`FieldAccess`/`MethodAccess` "not visible") and, more importantly, a **runtime `System.FieldAccessException`**.

C# avoids this by hosting closures/lambdas in a compiler-generated `<>c` (display) type **nested inside the declaring type**, so the lambda shares the enclosing member's accessibility domain. G# hoists non-capturing lambdas to a top-level `<Program>` type, losing that access.

## Minimal repro (runtime failure, not just ilverify)
```gsharp
package EN
import System
import System.Linq
import System.Collections.Generic
open class ChunkReader {
    protected let entries List[TrackEntry] = List[TrackEntry]()
    func AddOne() {
        entries.Add(TrackEntry(5, 1000))
    }
    func FirstId() uint32 {
        return entries.Select((e TrackEntry) -> e.TrackId).First()
    }
    protected open data class TrackEntry(TrackId uint32, Timescale uint32) {
    }
}
func Main() {
    let r = ChunkReader()
    r.AddOne()
    Console.WriteLine(r.FirstId())
}
```

Compiles with **0 gsc errors**. Then:

- `ilverify`: `[FieldAccess]: EN.<Program>::<lambda1>(...TrackEntry) [offset 0x01] Field is not visible.`
- **Runtime**: `System.FieldAccessException: Attempt by method 'EN.<Program>.<lambda1>(TrackEntry)' to access field 'EN.ChunkReader+TrackEntry.TrackId' failed.`

The lambda `(e) -> e.TrackId` is non-capturing, so it is emitted as a **static** method on `<Program>`. `TrackEntry` is a `protected` nested type, so its positional member `TrackId` is not accessible from `<Program>`.

## What works (contrast)
- A lambda that **captures `this`** and reads a `protected` field of the enclosing class works (it's hosted on a display class with the right access).
- The same lambda over a **public top-level** `data class` works (member is public).

So the defect is specifically: **non-capturing lambda host-type placement** + access to non-public members of the parameter/enclosing types.

[...]

PR #2024, opened 2026-07-04
gsc: diagnose non-generic local function referencing enclosing type parameter (#2016)

### LINKED ISSUE #2016: gsc: non-generic local function using an enclosing type parameter as its own param/return type emits BadImageFormatException
Discovered during PR #2013 (#1940) test build-out. A NON-generic local function that references an enclosing type parameter `U` directly as its own parameter or return type crashes the emitted assembly with `BadImageFormatException` at runtime. This reproduces on baseline (pre-#2013), so it is a pre-existing invalid-IL bug distinct from #1940 (which is about GENERIC local functions declaring their own type params that reference the enclosing one).

Likely the same invalid-IL family as #1940 — a local function body referencing an enclosing method type parameter — but for the non-generic case #1940's GS0468 diagnostic does not fire. Needs either the diagnostic extended to this shape or correct IL emission.

Filed as follow-up; not fixed in #2013 (out of scope).

DavidObando:
Fixes #2016

## Root cause
A NON-generic named local function (`let Name = func (...) ... {...}`, no `[T, ...]` of its own) that captures no outer variables is hoisted to a top-level static method (issue #1469's zero-capture fast path) unless it is nested inside a non-generic user type purely for accessibility (`ClosureEmitter.SynthesizeClosures`). When there is no such nesting available — the local function is declared at top level (inside a plain/generic top-level function), or its enclosing user type is itself generic — the hoisted method carries none of the enclosing method's/class's type parameters. A direct reference to one of them in the local function's own parameter type, return type, or body then has no corresponding CLR slot: invalid IL that silently crashes at run time with `BadImageFormatException` — the same invalid-IL family as #1940's generic local function case, but without that fix's own-type-parameter list to hide behind.

Confirmed pre-existing (reproduces on baseline before this fix):
```
func Outer[U](seed U) U {
    let Local = func (x U) U {
        return x
    }
    Console.WriteLine(Local(seed))
    return seed
}
Outer("hi")
```
crashed with `BadImageFormatException` at run time; now reports a compile-time diagnostic instead.

## Approach: diagnostic (reusing GS0468), not a broader emit fix
PR #2013 (fixing #1940) explicitly chose diagnosing over correct emit because "correct emit would require threading the enclosing type parameter(s) through as additional MVARs on the hoisted method plus every call site's type-argument list/inference — high-risk, broad surface for a narrow feature." That reasoning applies equally here, so this PR extends the same GS0468 diagnostic (generalizing its message/XML doc to cover both the generic and non-generic shapes) rather than adding a new code — a repo-enforced test (`DiagnosticIdUniquenessTests`) requires exactly one `ReportXxx` method per `GS####` id, so the two candidate methods were merged into one.

I also confirmed the shapes that already work correctly stay correct and are not accidentally flagged: a zero-capture local function nested inside a **non-generic** class member that references an **enclosing generic method's** type parameter genuinely compiles and runs correctly today (the nested display-class synthesis already reifies over referenced enclosing type parameters, issues #1477/#1512), and a **capturing** local function referencing an enclosing type parameter also already works via that same reified-closure path. Neither of those is touched by this diagnostic.

## Changes
- `src/Core/CodeAnalysis/DiagnosticBag.cs`: generalized `ReportGenericLocalFunctionCannotReferenceEnclosingTypeParameter`'s message/doc to cover both the generic-local-function (#1940) and non-generic, zero-capture local function (#2016) shapes; removed the duplicate method to satisfy the one-method-per-id convention.
- `src/Core/CodeAnalysis/Binding/LambdaBinder.cs`: added `CheckNonGenericLocalFunctionEnclosingTypeParameterReference`, which detects the broken shape (zero captured variables, non-async, and no non-generic-struct nesting available) and scans the literal's parameters/return type/body for a reference to any enclosing type parameter currently in scope.
- `src/Core/CodeAnalysis/Binding/StatementBinder.cs` / `Binder.cs`: wired the new check into `BindVariableDeclaration`'s `let`-bound function-literal initializer path (where the local function's declared name/location are available).
- `test/Compiler.Tests/Emit/Issue2016NonGenericLocalFunctionEnclosingTypeParameterTests.cs`: parameter-type, return-type, and body reference shapes; both an enclosing method and an enclosing generic class; and three no-false-positive cases described above.

## Verified
- Full `Core.Tests`: 5382 passed, 1 pre-existing skip.
- New `Issue2016` tests: 7 passed.
- `Issue1940` regression tests: unaffected, still passing.
- Broader lambda/closure `Emit` filter (Lambda/Closure/Issue503/567/618/909/1335): 51 passed.

## Deferred
Nothing — no tangential bugs surfaced during this investigation.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/Binder.cs
+++ b/src/Core/CodeAnalysis/Binding/Binder.cs
@@ -266,7 +266,8 @@ public sealed class Binder
             bindVariableDeclarationAttributes: (annotations, positionDescription) => declarations.BindAttributes(annotations, AttributeTargetKind.Field, VariableDeclarationAllowedTargets, positionDescription, System.AttributeTargets.Field),
             getCurrentFunction: () => this.function,
             bindLambdaWithTargetType: (syntax, targetType) => lambdas.BindLambdaExpression(syntax, targetType),
-            bindGenericLocalFunctionDeclaration: syntax => lambdas.BindGenericLocalFunctionDeclaration(syntax));
+            bindGenericLocalFunctionDeclaration: syntax => lambdas.BindGenericLocalFunctionDeclaration(syntax),
+            checkNonGenericLocalFunctionEnclosingTypeParameterReference: (location, name, literal) => lambdas.CheckNonGenericLocalFunctionEnclosingTypeParameterReference(location, name, literal));
         declarations = new DeclarationBinder(
             binderCtx,
             conversions,
--- a/src/Core/CodeAnalysis/Binding/LambdaBinder.cs
+++ b/src/Core/CodeAnalysis/Binding/LambdaBinder.cs
@@ -391,6 +391,28 @@ internal sealed class LambdaBinder
     /// with <see cref="System.BadImageFormatException"/> instead of failing to compile —
     /// the same invalid-IL family as the generic-local-function case, but without that fix's
     /// own-type-parameter list to hide behind.
+    ///
+    /// Follow-up review of #2024: an earlier revision of this method
+    /// short-circuited on <c>literal.Function.IsAsync</c>, on the assumption
+    /// that async local functions are owned by the async state-machine
+    /// synthesis (<c>StateMachineEmitter.SynthesizeAsyncLambdaStateMachines</c>)
+    /// rather than the plain zero-capture static-method hoisting path, and
+    /// might therefore reify the enclosing type parameter safely. That
+    /// assumption was verified FALSE: a zero-capture async local function's
+    /// kickoff method is still <c>literal.Function</c> itself — the exact
+    /// same un-parameterized top-level static method used by the sync path
+    /// — and the synthesized state-machine struct
+    /// (<see cref="GSharp.Core.CodeAnalysis.Lowering.Async.SynthesizedStateMachineType.MaterializeAsStructSymbol"/>)
+    /// never re-declares the kickoff's enclosing type parameters either. A
+    /// hoisted field of the enclosing type parameter's type therefore has the
+    /// identical dangling-MVAR shape as the sync case, confirmed by direct
+    /// repro: an UNCALLED `let Local = async func (x U) U { return x }` inside
+    /// `func Outer[U](seed U) U` compiled clean before this fix and crashed at
+    /// run time with <see cref="System.BadImageFormatException"/> the moment
+    /// `Outer` executed (the state machine's field layout is invalid
+    /// regardless of whether the local function is ever called). The
+    /// short-circuit is removed so this check also covers async local
+    /// functions.
     /// </summary>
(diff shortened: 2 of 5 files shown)


DavidObando:
Fixed the blocking bug from the rubber-duck review, plus both non-blocking notes, in ea9f269b.

**Blocking bug (var bypasses GS0468):** The gate in `StatementBinder.BindVariableDeclaration` no longer checks `syntax.Keyword?.Kind == SyntaxKind.LetKeyword`. It now fires whenever the bound initializer is a `BoundFunctionLiteralExpression` (regardless of let/var/const), since the emitter's zero-capture hoisting path never distinguished those keywords in the first place. Confirmed the reviewer's exact `var` repro now reports GS0468 at compile time instead of emitting a crashing DLL.

**Non-blocking #1 (IsAsync short-circuit):** Investigated — it was NOT safe. A zero-capture async local function referencing an enclosing type parameter has the identical dangling-MVAR shape: its kickoff method is still the un-parameterized top-level static method, and the synthesized async state-machine struct (`SynthesizedStateMachineType.MaterializeAsStructSymbol`) never re-declares the enclosing type parameter either. Repro'd directly: an uncalled async local function referencing enclosing `U` inside a generic outer function compiled clean (exit 0) before this fix and crashed with `BadImageFormatException` the instant the outer function executed. Removed the short-circuit; GS0468 now fires for this shape too.

**Non-blocking #2 (rename):** `DiagnosticBag.ReportGenericLocalFunctionCannotReferenceEnclosingTypeParameter` renamed to `ReportLocalFunctionCannotReferenceEnclosingTypeParameter` (same GS0468 diagnostic ID), all call sites updated.

**Tests added** (Issue2016NonGenericLocalFunctionEnclosingTypeParameterTests.cs): var-declared, const-declared, and async local-function repros, each asserting GS0468 fires and no crashing DLL is produced.

**Also found (filed separately, out of scope):** #2026 — generic async function calls (top-level or local) lose the Task-of-T wrap on their substituted return type (awaiting a generic async call fails to type-check), a pre-existing bug unrelated to the IL-hoisting family here.

**Verification:**
- `dotnet test test/Compiler.Tests/Compiler.Tests.csproj -c Release --filter "FullyQualifiedName~Issue2016"` — 10/10 passed.
- `dotnet test test/Compiler.Tests/Compiler.Tests.csproj -c Release --filter "FullyQualifiedName~Issue1940"` — 15/15 passed, no regression.
- `dotnet test test/Core.Tests/Core.Tests.csproj -c Release` — 5382 passed, 1 pre-existing skip (includes DiagnosticIdUniquenessTests, which still passes after the rename).
- Manually re-ran the reviewer's exact repro (var-declared local function referencing enclosing type param) — now reports GS0468 at compile time; no DLL emitted, no runtime crash.

Nothing deferred within scope.

Issue #950, opened 2026-06-22
Add `protected` modifier in addition to `public`, `private`, `internal`.


DavidObando:
This brings observability scoping to be on par with C#.

Note that `protected` would only be usable in `open` classes and structs.

PR #2048, opened 2026-07-04
gsc: diagnose inaccessible-member writes (private field/property assignment) (#2044)

### LINKED ISSUE #2044: gsc: private field write from outside the declaring type not diagnosed as inaccessible
Discovered while fixing #2004 (PR #2042).

Writing a **private** field of a type from a function OUTSIDE the declaring type is not diagnosed as an accessibility error (no GS diagnostic). Expected: an inaccessible-member diagnostic, mirroring how private member READS are (or should be) checked.

Not investigated deeply (out of scope for #2004).

DavidObando:
Fixes #2044

## Root cause
`AccessibilityChecker.IsAccessible` (issue #950) only ever enforced `protected`; every other accessibility, including `private`, was treated as always-accessible by design. Since both the field/property READ path (`BindAccessorExpression`) and the simple-receiver WRITE path (`BindFieldAssignmentExpression`) already called this shared checker, **both reads and writes of a private member from outside its declaring type were silently allowed** — not just writes as initially suspected.

Additionally, two write paths never called the checker at all (a pre-existing gap, even for `protected`):
- `BindMemberFieldAssignmentExpression` — chained/expression-receiver assignment (`a.B.C = v`, `GetObj().Field = v`)
- `TryBindChainedCompoundAssignment` — compound assignment (`+=`/`-=`) on any member-access LHS (the parser routes ALL `lhs.member +=/-=` through this path regardless of receiver shape)

## Fix
- `AccessibilityChecker.IsAccessible`: now also enforces `private`. A private member is accessible from its declaring **top-level** type's body (including nested types declared inside it, matching C#'s "private is visible anywhere within the containing type" rule), but is **not** inherited by derived types (unlike `protected`).
- `DiagnosticBag`: added `GS0471` for private-member inaccessibility, dispatched alongside the existing `GS0379` (protected) through a shared `ReportMemberInaccessible` helper so all 9 call sites report the right diagnostic for the member's actual accessibility.
- Threaded the accessibility check into the two write paths that were missing it entirely (`BindMemberFieldAssignmentExpression`, `TryBindChainedCompoundAssignment`), for both fields and properties.
- `Binder.cs`: stopped seeding a base class's `private` fields/properties as bare names inside a derived type's method bodies — private is not inherited, so unqualified lookup must not resolve it (now correctly falls through to the pre-existing `GS0125` "variable doesn't exist", matching how C# never binds an unqualified name to an inaccessible base member).
- `??=` (null-coalescing assignment) needed no additional change — it binds its LHS through the ordinary read path first, so it's covered by the `AccessibilityChecker` fix alone.

## Tests
Added `test/Core.Tests/CodeAnalysis/Binding/Issue2044PrivateAccessibilityTests.cs`:
- External write / read / compound-assign (`+=`) / null-coalescing-assign (`??=`) of a private field → `GS0471`.
- External write of a private property → `GS0471`.
- Same-type read/write/compound-assign of a private field → no diagnostics (no regression).
- Nested type reading/writing its enclosing type's private field → no diagnostics.
- Derived class referencing a base's private field by bare name → `GS0125` (name doesn't resolve, matching C#).

## Validation
- `dotnet build src/Compiler/Compiler.csproj -c Release` — succeeds.
- `dotnet test test/Core.Tests/Core.Tests.csproj -c Release` — 5460 passed, 1 pre-existing skip, 0 failed (was 5452/1/0 before this change; +8 new tests).


DavidObando:
Merged main + renumbered diagnostic GS0471 → GS0472 to resolve ID collision with #2047 (which merged first and claimed GS0471 for a different diagnostic).

Issue #1911, opened 2026-07-03
cs2gs: explicit interface implementations lowered to plain methods (GS0264 / ilverify miss)


DavidObando:
`void IGreeter.Greet()` alongside a same-name public method emits a duplicate overload (GS0264); alone, it compiles but ilverify reports `Class implements interface but not method` — the explicit binding is lost. G# disambiguates interface diamonds by other means (samples/InterfaceDiamondDisambiguation.gs); a mapping decision is needed. (grid app G06).
## Definition of done
- [ ] Fix + `Issue<N>*.cs` regression test
- [ ] Re-enable the quarantined grid fixture (or add one) and re-capture its baseline
- [ ] `tools/cs2gs/coverage/csharp-construct-inventory.json` row updated

_Discovered by the ADR-0138 conformance-grid build-out (gsc 0.2.585+01ad6ec964); quarantined sources live next to each grid app under `tools/cs2gs/corpus/grid/`._

Issue #985, opened 2026-06-23
Generic collection implementing IEnumerable[T] needs dual GetEnumerator overloads (GS0264 + GS0187)


DavidObando:
Discovered by the **cs2gs** C#→G# migration pipeline (#914), L3-Library/L5 corpus. Verified with gsc `0.2.124+6cef33811b`.

A C# generic collection implements `IEnumerable<T>` by declaring the generic `IEnumerator<T> GetEnumerator()` plus an explicit-interface non-generic `IEnumerator IEnumerable.GetEnumerator()`. G# has no explicit-interface-implementation syntax, so both map to a method named `GetEnumerator` differing **only by return type**, which G# forbids (`GS0264`). With only the generic overload, the non-generic contract member stays unimplemented (`GS0187`). #974 fixed generic-interface satisfaction; this residual gap blocks faithfully migrating hand-written enumerable types.

### Repro (GS0264 + GS0187)
```gs
package T
import System.Collections.Generic
class Repo[T] : IEnumerable[T] {
    private let _items List[T] = List[T]()
    func GetEnumerator() IEnumerator[T] { return _items.GetEnumerator() }
    private func GetEnumerator() IEnumerator { return GetEnumerator() }
}
```

### Control (canonical G# enumeration surface compiles)
```gs
func numbers() sequence[int32] { yield 1; yield 2 }
// for n in numbers() { ... }
```

**Classification:** missing language feature (satisfy a non-generic base interface member, or explicit-interface-impl). Corpus workaround: `sequence[T]` generators. Ref ADR-0115 §G.

Issue #1929, opened 2026-07-03
gsc: G#-level semantics lost across assembly references (data-class with/equality, struct literals, positional ctors, receiver funcs, internal visibility)


DavidObando:
When a translated G# test project references the translated G# library as a compiled DLL, gsc loses the library's G#-level semantics: `with`-copy → GS0161 'requires a data struct receiver', struct composite literals → GS0157, data-struct positional ctors → GS0130, receiver-clause funcs → GS0159, and `internal` members are invisible with no InternalsVisibleTo equivalent. This blocks the whole library test-parity path (L2/L3). Either G# metadata must round-trip through the DLL or the parity harness needs a source-merged mode. (L2-Library parity, fingerprint sha256:9fb51a6cca02…).
## Definition of done
- [ ] Fix + `Issue<N>*.cs` regression test
- [ ] Re-enable the quarantined grid fixture (or add one) and re-capture its baseline
- [ ] `tools/cs2gs/coverage/csharp-construct-inventory.json` row updated

_Discovered by the ADR-0138 conformance-grid build-out (gsc 0.2.585+01ad6ec964); quarantined sources live next to each grid app under `tools/cs2gs/corpus/grid/`._

Issue #1908, opened 2026-07-03
cs2gs: C# 14 instance compound-assignment operator declarations fail round-trip


DavidObando:
A C# 14 instance `public void operator +=(…)` emits G# that fails round-trip parse: `GS0005: Unexpected token <PlusEqualsToken>, expected <PlusToken>` — G# operator declarations have no compound form (ADR-0035 mapping decision needed). (grid app G07).
## Definition of done
- [ ] Fix + `Issue<N>*.cs` regression test
- [ ] Re-enable the quarantined grid fixture (or add one) and re-capture its baseline
- [ ] `tools/cs2gs/coverage/csharp-construct-inventory.json` row updated

_Discovered by the ADR-0138 conformance-grid build-out (gsc 0.2.585+01ad6ec964); quarantined sources live next to each grid app under `tools/cs2gs/corpus/grid/`._

PR #2049, opened 2026-07-04
gsc+cs2gs: preserve explicit interface impl bodies via CLR MethodImpl (#2010)

### LINKED ISSUE #2010: cs2gs: explicit interface implementations collapse distinct bodies instead of preserving via CLR MethodImpl
Follow-up from PR #1994 (issue #1911) review.

The current fix for explicit interface implementations forces public visibility and drops exact-signature duplicate methods when 2+ explicit impls collide (keeping one survivor), now correctly surfaced via an `Unsupported` diagnostic that fails the translate stage (no longer silent).

A stronger alternative exists that would preserve full fidelity with zero drops: emit each explicit implementation as a separate, name-mangled G# method and use CLR-level `MethodImpl`/explicit interface slot mapping at emit time (reusing the ADR-0089 static-virtuals machinery around `ReflectionMetadataEmitter.cs:11399`) so each interface's dispatch slot points at its own distinct method body, instead of collapsing to a single survivor method.

This is a genuinely bigger scope than #1911's fix (new emit-layer work, not just translator lowering) — filed as follow-up, not required to unblock #1994.

DavidObando:
Fixes #2010

## Summary

Full-fidelity fix (not the intermediate fallback) — no drops, zero semantic loss for G# user-interface explicit implementations.

Follow-up from PR #1994 (issue #1911). #1911's fix forced explicit interface implementations public and dropped colliding same-signature siblings, surfaced via an `Unsupported` diagnostic. This PR replaces that with the "genuinely bigger" alternative the issue proposed: each explicit implementation of a G# user interface is emitted under a reserved mangled name (`__explicit_<Interface>__<Member>`), gsc's binder links it to the specific interface member it implements, and the emitter binds a CLR `MethodImpl` row per implementation — reusing the exact ADR-0089 static-virtual / issue #985 covariant-return-bridge machinery already present in `ReflectionMetadataEmitter.cs`. Each interface's dispatch slot now routes to its own distinct method body.

## Design

- `FunctionSymbol.ExplicitInterfaceMember` (new): the in-compilation interface member a method explicitly implements. Distinct from the pre-existing `ExplicitInterfaceSlot` (a `System.Reflection.MethodInfo`, used for the #985 *external/BCL* interface bridge).
- `DeclarationBinder`: recognizes the `__explicit_<Interface>__<Member>` mangled-name convention during the deferred `VerifyInterfaceImplementations` pass (interface members aren't bound yet at initial class-member-bind time), links the method, and treats it as satisfying that interface's abstract slot without requiring a name match. No new grammar — it's a plain reserved identifier convention, in the spirit of ADR-0091's rejection of an `IFoo.M(this)` surface syntax.
- `ReflectionMetadataEmitter.EmitExplicitInterfaceMethodImpls`: extended to also bind a `MethodImpl` row for `ExplicitInterfaceMember`, generic-interface aware (reuses `ResolveUserInterfaceInstanceMethodToken`).
- `CSharpToGSharpTranslator`: a C# explicit implementation of a G# user interface (declared in source) now emits under the mangled name with C#-faithful (non-public) visibility, instead of forced-public + collision-drop. An explicit implementation of an external/BCL interface keeps the prior #1911 behavior unchanged (forced public, collision-drop with `Unsupported`) since it relies on the pre-existing #985 bridge, which is keyed on the method's un-mangled name plus a return-type-only signature difference — mangling would break that narrower, already-working mechanism.

## Tests

- `test/Compiler.Tests/Emit/Issue2010ExplicitInterfaceImplEmitTests.cs`: two interfaces with a same-name/same-signature member, both explicitly implemented with distinct bodies on one type. Verifies two MethodDef rows (no drop), two MethodImpl rows, ilverify-clean IL, and correct dispatch through each interface-typed reference (`((IFoo)obj).Bar()` vs `((IBaz)obj).Bar()`) end to end.
- `tools/cs2gs/Cs2Gs.Tests/Issue2010ExplicitInterfaceImplementationTests.cs`: translator-level coverage — colliding explicit impls with no public sibling, and an explicit impl coexisting with a same-name public method. Both now translate with zero drops and zero diagnostics.
- `tools/cs2gs/Cs2Gs.Tests/Issue1911ExplicitInterfaceImplementationTests.cs`: updated for the new mangled-name/private-visibility behavior on the lone-impl case; the collision-drop-specific tests moved to the #2010 file (that behavior no longer applies to G# user interfaces).
- `tools/cs2gs/corpus/grid/G06-Types-Console/Constructs/ExplicitInterfaceSpecifier.cs`: extended the corpus fixture with a second interface (`IWelcomer`) colliding on `Greet()`, reproducing the original #1911/#1994-review scenario end-to-end through the real migrate → compile → ilverify → run pipeline.

## Results

- Full `Core.Tests`: 5452 passed, 1 pre-existing skip, 0 failed.
- Full `Cs2Gs.Tests`: 986 passed, 0 failed.
- Targeted `Compiler.Tests` (Issue985/Issue2010/StaticVirtual/Interface filters, `-- xUnit.MaxParallelThreads=2`): 222 passed, 0 failed.
- Full corpus sanity migrate (`cs2gs migrate --corpus tools/cs2gs/corpus --baseline tools/cs2gs/triage/gaps.json`): 16/20 apps green; the 3 failing apps match the existing baseline exactly (0 new, 0 regressed).
- `cs2gs coverage --write`: no observable coverage-matrix diff (this is a behavioral fix, not a new/changed construct-coverage classification).

Nothing deferred — external/BCL-interface explicit-impl handling is intentionally left on the narrower #985/#1911 path since it already works and mangling it would have broken that pre-existing mechanism (caught and fixed during testing).

### DIFF
--- a/src/Core/CodeAnalysis/Binding/DeclarationBinder.cs
+++ b/src/Core/CodeAnalysis/Binding/DeclarationBinder.cs
@@ -6897,7 +6897,7 @@ internal sealed class DeclarationBinder
             }
 
             if (!TryParseExplicitInterfaceImplName(candidate.Name, out var explicitIfaceName, out var explicitMemberName) ||
-                explicitIfaceName != iface.Name ||
+                explicitIfaceName != QualifyInterfaceName(iface) ||
                 explicitMemberName != imethod.Name)
             {
                 continue;
@@ -6921,6 +6921,43 @@ internal sealed class DeclarationBinder
 
     private static ImmutableArray<ParameterSymbol> GetCallableParameters(FunctionSymbol method)
         => method.ExplicitReceiverParameter == null ? method.Parameters : method.Parameters.RemoveAt(0);
+
+    /// <summary>
+    /// Namespace/nesting-qualified name of a G# interface, dots sanitized to
+    /// underscores — must stay in sync with the equivalent computation in
+    /// <c>CSharpToGSharpTranslator.QualifyInterfaceName</c>, which builds the
+    /// same string from the pre-translation C# interface symbol. Follow-up
+    /// to issue #2010: bare simple names collided across namespaces
+    /// (<c>Foo.IBar</c> vs <c>Baz.IBar</c>); qualifying by package name (G#'s
+    /// analogue of a C# namespace) plus any containing-type nesting fixes it.
+    /// </summary>
+    private static string QualifyInterfaceName(InterfaceSymbol iface)
+    {
+        var parts = new List<string>();
+        for (TypeSymbol current = iface; current != null; current = GetContainingType(current))
+        {
+            parts.Insert(0, current.Name);
+        }
+
+        if (!string.IsNullOrEmpty(iface.PackageName))
+        {
+            // The package name itself may be dotted (e.g. "Corpus.Grid06"),
+            // mirroring a multi-segment C# namespace — split it into its own
+            // segments rather than inserting one dotted string, or the final
+            // '_' join below would leave a stray '.' in the result.
+            parts.InsertRange(0, iface.PackageName.Split('.'));
+        }
+
+        return string.Join("_", parts);
+    }
+
+    private static TypeSymbol GetContainingType(TypeSymbol type) => type switch
+    {
+        StructSymbol s => s.ContainingType,
+        InterfaceSymbol i => i.ContainingType,
+        EnumSymbol e => e.ContainingType,
+        _ => null,
+    };
 
     /// <summary>
     /// Issue #2010: the reserved mangled-name convention a G# method uses to
--- a/src/Core/CodeAnalysis/Binding/DeclarationBinder.cs
+++ b/src/Core/CodeAnalysis/Binding/DeclarationBinder.cs
@@ -1860,7 +1860,17 @@ internal sealed class DeclarationBinder
                         ValidateInlineDataNilArguments(methodAttributes, methodSymbol.Parameters);
                     }
 
-                    // ADR-0063 §11: detect duplicate-signature within the class.
+                    // Issue #2010: a method named per the reserved
+                    // `__explicit_<Interface>__<Member>` mangled convention (the
+                    // form cs2gs synthesizes for a C# explicit interface
+                    // implementation, see CSharpToGSharpTranslator) explicitly
+                    // implements one interface member's own distinct body.
+                    // Resolution against the actual interface member is
+                    // deferred to VerifyInterfaceImplementations — interfaces
+                    // declared later in the same compilation unit have not had
+                    // their own members bound yet when this class's members
+                    // are bound, so `implementedInterfaces[i].Methods` may
+                    // still be empty here.
                     var hasDuplicateSig = false;
                     foreach (var existingMethod in methodsBuilder)
                     {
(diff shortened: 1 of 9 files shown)


DavidObando:
## Follow-up commit: fixes for rubber-duck review findings

Addresses issue #1 (blocking) and #2 (non-blocking) from the review, and verifies issue #6.

### #1 — namespace-qualified mangling (fixed)
`__explicit_<Interface>__<Member>` previously used the interface's bare simple name. Two interfaces with the same simple name in different namespaces (e.g. `Foo.IBar` / `Baz.IBar`), both explicitly implemented by one class with a same-signature member, mangled to the identical `__explicit_IBar__M` — a hard GS0264 duplicate-signature compile error with no diagnostic.

Fix: the mangle now embeds the interface's namespace/package-qualified name (dots → underscores) on both sides, which must stay in sync:
- cs2gs `MangleExplicitInterfaceImplName` — built from the Roslyn interface symbol's fully-qualified display name.
- gsc `DeclarationBinder.TryResolveExplicitInterfaceImplementation` — built from the G# `InterfaceSymbol`'s `PackageName` + containing-type chain + `Name`.

Added `TwoSameSimpleNameInterfacesFromDifferentNamespaces_MangleDistinctly` proving the two mangled names are now distinct and both implementations survive translation with correct bodies. Updated the existing #1911/#2010 tests' hardcoded mangled-name expectations (e.g. `__explicit_IGreeter__Greet` → `__explicit_Corpus_Issue1911_IGreeter__Greet`) and the direct-G# `GapCheck` emit test to the new qualified scheme.

### #2 — homogeneous handling of multi-entry `ExplicitInterfaceImplementations` (fixed)
Roslyn's `ExplicitInterfaceImplementations` can hold >1 entry (e.g. a base interface re-declaring an identical-signature member, satisfied transitively by one C# explicit-impl declaration). The translator previously always indexed `[0]` for both classification and mangling, silently picking one interface arbitrarily.

Fix: since the mangled-name + `MethodImpl` scheme only wires a single interface slot per mangled method, this shape (and any mixed user+external-interface shape) now falls back to the pre-#2010 (#1911) named/forced-public path, with a new `Info`-severity diagnostic explaining why. Ordinary implicit name+signature interface dispatch then satisfies every entry uniformly — it still compiles and dispatches correctly (documented trade-off: loses C#'s "not publicly callable by name" explicit-impl semantics for this rare shape). This is a genuinely rare C# construct (interface-inheritance diamond + single explicit-impl body); constructing a realistic, minimal repro test for it turned out impractical without a lot of scaffolding, so the fallback + diagnostic is documented in code comments instead per the review's suggested option.

### #6 — corpus arithmetic reconciled
Re-ran `cs2gs migrate` against the full corpus + baseline:

```
16/20 apps green; run FAILED.
baseline: 3 known, 0 new, 0 regressed, 0 stale.
```

The 3 baseline gaps are `CompileGap-Library` (deliberate wontfix), `L2-Library` test-parity (issue #1929), and `G07-Members-Console` (issue #1908, wontfix). The 4th non-green app, `corpus/L3-Library`, is a pre-existing documented `unverifiedApps` skip in `tools/cs2gs/triage/gaps.json` (not a failure/gap) — so `16 + 3 + 1 = 20` reconciles exactly. The original PR body's "16/20 green; 3 failing apps match baseline" was correct on the baseline-gap count, it just didn't separately call out the 4th (skipped, not failed) app. No new or regressed apps from this change.

### Test results
- `Core.Tests`: 5452/5452 passed
- `Cs2Gs.Tests`: 987/987 passed
- `Compiler.Tests`: 2708/2708 passed

PR #2057, opened 2026-07-04
gsc: diagnose private field write from outside declaring type (#2044)

### LINKED ISSUE #2044: gsc: private field write from outside the declaring type not diagnosed as inaccessible

DavidObando:
## Root cause
`AccessibilityChecker.IsAccessible` (added for issue #950) only enforced `protected`; every other accessibility, including `private`, was treated as always accessible at bind time. Since every field/property read, write, and base-class-call site already routes through this single helper, a private field/property write (and read) from outside its declaring type compiled silently instead of reporting GS0379.

## Fix
- `AccessibilityChecker.IsAccessible` now also enforces `private`: accessible only from the exact declaring type's own body (no derived-type allowance, unlike `protected`).
- `DiagnosticBag.ReportProtectedMemberInaccessible` now takes the member's `Accessibility` so GS0379's message correctly says 'private member' vs 'protected member'; the same GS0379 code is reused (mirrors the existing read-side/protected diagnostic exactly).
- All existing field/property read, write, and base-class-call sites now pass the member's actual accessibility instead of implicitly assuming `protected` — this alone fixes both the read and write cases for private fields/properties with no new call sites needed.
- `TryBindChainedCompoundAssignment` (the `obj.field += x` / `obj.prop += x` path) had **no** accessibility check at all (a pre-existing gap that also affected `protected`); added the same check used by the plain read/write paths.

## Tests
New `test/Core.Tests/CodeAnalysis/Binding/Issue2044PrivateMemberWriteTests.cs`, mirroring `Issue950ProtectedModifierTests.cs`:
- external write / read / compound-assign of a private field → GS0379
- external write of a private property → GS0379
- same-type write of its own private field → still compiles, no diagnostics
- a private base-class field written through a derived instance → still GS0379 (private does not extend to derived types, unlike protected)

Ran narrowly (no full-suite run, per 16GB constraint):
- `--filter "FullyQualifiedName~Issue2044|FullyQualifiedName~Issue950|FullyQualifiedName~Issue1335"` → 22 passed
- `--filter "FullyQualifiedName~Accessib|FullyQualifiedName~Inaccessible|FullyQualifiedName~Protected|FullyQualifiedName~Private"` → 49 passed
- `--filter` covering every other existing test file that declares `private` fields/properties (Issue1070FieldInitializerStaticTests, Issue1088GenericIdentityTests, VisibilityModifierTests, Issue1240ParameterShadowsFieldTests, PropertyParserTests, Issue1202UnsafeClassModifierParserTests) → 71 passed

Fixes #2044

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
@@ -3331,7 +3331,7 @@ internal sealed partial class ExpressionBinder
                         // declaring type and its derived types may read it.
                         if (!AccessibilityChecker.IsAccessible(field.Accessibility, declaringType, this.function))
                         {
-                            Diagnostics.ReportProtectedMemberInaccessible(ne.IdentifierToken.Location, field.Name, declaringType.Name);
+                            Diagnostics.ReportProtectedMemberInaccessible(ne.IdentifierToken.Location, field.Name, declaringType.Name, field.Accessibility);
                         }
 
                         // ADR-0122 §10 / issue #1035: a fixed-size buffer field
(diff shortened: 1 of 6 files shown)


DavidObando:
Duplicate — issue #2044 already fixed and merged via PR #2048.

Issue #756, opened 2026-06-12
Private interface helper methods (follow-up from #726 / ADR-0085)


DavidObando:
## Summary

Allow `private` helper methods inside `interface` declarations. Deferred from #726 (ADR-0085 minimal-scope DIM landing).

## Motivation

Once interfaces can carry method bodies (default-interface methods), real-world default implementations grow shared helpers. C# 8 lets these be `private` so they don't leak into the public interface surface. Without this, every helper has to live on the public interface — which clutters API consumers and prevents refactoring.

## Scope

- Parser: accept `private func` inside `interface`.
- Binder: enforce that `private` interface members are only callable from other members of the same interface declaration.
- Emit: `.method private hidebysig instance ...` inside the interface TypeDef.
- Diagnostic for an implementer trying to override a private interface member.

## Acceptance criteria

- Private helpers compile and link.
- Implementers cannot see the private member.
- Tests covering call-from-default, call-from-external-rejection, override-rejection.

## Related

- Deferred from #726 (ADR-0085 §"Deferred / Out of Scope").
- Builds on ADR-0018 and ADR-0085.

---

**Plan reference:** §6.5 follow-up. Tracked under #706.

PR #2065, opened 2026-07-04
gsc: diagnose private method call from outside declaring type (#2058)

### LINKED ISSUE #2058: gsc: private method call from outside declaring type not diagnosed (OverloadResolver still gates on ==Protected)
Follow-up from PR #2057 (issue #2044) review.

`OverloadResolver.BindUserInstanceCall` (~line 5711) hard-gates on `method.Accessibility == Accessibility.Protected` before calling `AccessibilityChecker.IsAccessible`. So a `private func` on a class/struct called from OUTSIDE its declaring type is still NOT diagnosed, even though #2057 taught `IsAccessible` to enforce `private`. Same class of hole as the field bug #2044 fixed.

Also `ReportProtectedMemberInaccessible(...)` at that call site still omits the new `accessibility` arg, so the wording would be wrong (says "protected") if reached.

## Fix
- Drop the `== Protected` gate; always consult `IsAccessible` and pass real accessibility.
- Pass the member accessibility to the diagnostic so wording is correct for private.
- Consider renaming `ReportProtectedMemberInaccessible` -> `ReportMemberInaccessible` and dropping the `Accessibility.Protected` default param so every call site is explicit.

## DoD
- Private method call from outside declaring type emits GS0379 + regression test
- Same-type/derived-protected calls still compile

DavidObando:
## Root cause
Two call-binding sites checked method accessibility incorrectly/incompletely:
- `OverloadResolver.BindUserInstanceCall` gated the check on `method.Accessibility == Accessibility.Protected` only, so `private` instance methods called from outside their declaring type were never diagnosed.
- `ExpressionBinder.BindUserTypeStaticCall` (static/`shared` method calls) had **no** accessibility enforcement at all.

## Fix
Both sites now call `AccessibilityChecker.IsAccessible` (the same checker added for field/property access in #2044) and report via the shared `DiagnosticBag.ReportMemberInaccessible` helper (GS0379 for protected, GS0472 for private) — no new diagnostic invented, no parallel checker.

## Call sites changed
- `src/Core/CodeAnalysis/Binding/OverloadResolver.cs`: `BindUserInstanceCall` — replaced the protected-only gate with a general `IsAccessible` check covering private too.
- `src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs`: `BindUserTypeStaticCall` — added the same check for static (`shared`) method calls (class/struct owners only; interface static methods keep their existing private-visibility diagnostic from #756).

## Tests
Added `test/Core.Tests/CodeAnalysis/Binding/Issue2058PrivateMethodCallTests.cs` mirroring `Issue2044PrivateAccessibilityTests.cs`:
- private instance method called from outside → GS0472
- private static method called from outside → GS0472
- private instance/static method called from inside declaring type → no diagnostic
- inherited protected method call → no diagnostic
- public method call → no diagnostic

Verified:
- `dotnet build src/Compiler/Compiler.csproj -c Release -graph` — succeeds.
- Targeted filter (Issue2058/Issue2044/Issue950/Issue1585/Accessib): 56/56 pass.
- Broader regression filter (Call/Overload/Interface/Visibility/Protected): 790/790 pass.
- Manual repro via `gsc` CLI reproduces the GS0472 diagnostic; `samples/PrivateInterfaceHelpers.gs` still compiles clean.

Fixes #2058

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
@@ -2789,6 +2789,16 @@ internal sealed partial class ExpressionBinder
                 return new BoundErrorExpression(null);
             }
 
+            // Issue #950 / #2058: enforce `protected`/`private` static method
+            // access — mirrors the instance-call check in
+            // OverloadResolver.BindUserInstanceCall. Interface static methods
+            // already have their own private-visibility diagnostic (issue
+            // #756), so this only applies to class/struct owners.
+            if (structSym != null && !AccessibilityChecker.IsAccessible(method.Accessibility, structSym, this.function))
+            {
+                Diagnostics.ReportMemberInaccessible(ce.Identifier.Location, method.Name, structSym.Name, method.Accessibility);
+            }
+
             // Issue #951: bind any deferred un-typed arrow lambda against the
             // selected static method's delegate-typed parameter so its omitted
             // parameter type(s) and inferred return type are filled in from the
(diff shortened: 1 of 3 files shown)


DavidObando:
Duplicate of #2058, already fixed and merged via #2062.

Issue #2060, opened 2026-07-04
gsc: bare-name inherited-private base field reachable in derived body (ImplicitFieldVariableSymbol bypasses accessibility)


DavidObando:
Follow-up from PR #2057 (issue #2044) review. Confirmed genuine soundness hole, pre-existing and orthogonal to #2044.

Unqualified (bare-name, no `this.`/receiver) access to an INHERITED PRIVATE base field inside a DERIVED class body resolves via a separate `ImplicitFieldVariableSymbol` path that bypasses `TypeMemberModel`/`AccessibilityChecker` entirely. So `secret = v` (where `secret` is a private field of a base class) from a derived body is neither read- nor write-checked.

## Fix
Route `ImplicitFieldVariableSymbol` resolution of inherited fields through `AccessibilityChecker.IsAccessible` (both read and write), mirroring the qualified `receiver.field` paths fixed in #2057.

## DoD
- Bare-name inherited-private read AND write from a derived body emits GS0379 + regression tests
- Bare-name access to inherited PROTECTED / own private still compiles

DavidObando:
Already fixed on main: Binder.cs (~L366-379) now skips declaring inherited private base fields as bare-name ImplicitFieldVariableSymbols in derived types, so bare-name read (ExpressionBinder.cs ~L488-506) and write (ExpressionBinder.Assignments.cs ~L45-69) can't reach an inherited private field. Verified against the external #2044 fix (5bbadbfd). Closing as resolved.

Issue #2067, opened 2026-07-04
gsc: accessibility not enforced for method-group-to-delegate (&Method) and private-constructor calls


DavidObando:
Follow-up from PR #2065 (issue #2058) review. #2065 added private/protected enforcement to instance/static METHOD CALL binding (GS0472/GS0379), but two adjacent call-like paths still bypass the compile-time accessibility check (CLR would still throw/verify at runtime, but gsc should diagnose):

1. **Method-group-to-delegate conversion / `&Method` ldftn.** `&Foo.Private` (taking a delegate/function-pointer to a private/protected method) from outside the declaring type is not diagnosed. Separate from the invocation path fixed in #2065.
2. **Private-constructor call.** `new Foo{}` / constructor invocation of a private (or protected) ctor from outside the declaring type is not diagnosed — separate binder path from method-call resolution.

## Fix
Route both paths through `AccessibilityChecker.IsAccessible` + `DiagnosticBag.ReportMemberInaccessible` (GS0472 private / GS0379 protected), mirroring #2044/#2065. Generalize to any inaccessible target.

## DoD
- `&PrivateMethod` from outside declaring type emits GS0472 + test
- Private/protected ctor call from outside emits the diagnostic + test
- Legal same-type / accessible uses still compile

PR #2078, opened 2026-07-04
gsc: enforce accessibility for static/shared method calls (#2071)

### LINKED ISSUE #2071: gsc: accessibility not enforced for static/shared method calls (BindUserTypeStaticCall)
Follow-up from PR #2065 (closed as duplicate of merged #2058/#2062, but #2065 also fixed a genuinely separate gap that was lost when it was closed).

## Root cause
`ExpressionBinder.BindUserTypeStaticCall` (static/`shared` method calls, e.g. `Foo.PrivateStatic()`) has NO accessibility enforcement at all — a `private`/`protected` static/shared method called from outside its declaring type is never diagnosed. This is separate from `OverloadResolver.BindUserInstanceCall` (instance calls), which #2058/#2062 already fixed.

## Fix
Route `BindUserTypeStaticCall` through `AccessibilityChecker.IsAccessible` + `DiagnosticBag.ReportMemberInaccessible` (GS0472 private / GS0379 protected), mirroring the instance-call fix in #2058/#2062.

## DoD
- Private/protected static (shared) method call from outside declaring type emits the diagnostic + regression test
- Legal same-type / accessible static calls still compile
- Verify current diagnostic ID on main by grepping DiagnosticBag.cs (do not trust stale numbers from other issue text)

DavidObando:
Fixes #2071

Enforces `private`/`protected` accessibility on static/shared method calls via `BindUserTypeStaticCall` (ExpressionBinder.Access.cs), mirroring the instance-call gate added for #2058/#2060. The instance-call path landed separately on main via #2060; this PR completes the static-call variant. Includes tests covering private and protected static method calls from outside the declaring type.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
@@ -2789,6 +2789,16 @@ internal sealed partial class ExpressionBinder
                 return new BoundErrorExpression(null);
             }
 
+            // Issue #950 / #2058: enforce `protected`/`private` static method
+            // access — mirrors the instance-call check in
+            // OverloadResolver.BindUserInstanceCall. Interface static methods
+            // already have their own private-visibility diagnostic (issue
+            // #756), so this only applies to class/struct owners.
+            if (structSym != null && !AccessibilityChecker.IsAccessible(method.Accessibility, structSym, this.function))
+            {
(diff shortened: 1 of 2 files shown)


DavidObando:
Redundant: external #2067 (main b92d9788) already added the identical static/shared method-call accessibility enforcement in ExpressionBinder.Access.cs (method.StaticOwnerType is StructSymbol -> ReportMemberInaccessible), and #2071 is now CLOSED. Nothing unique left to add.

Issue #2085, opened 2026-07-04
gsc: field-like event of a named delegate type emits invalid add_/remove_ accessor IL (any assignment)


DavidObando:
## Summary
Discovered while fixing #2066 (nullable-local delegate snapshot invoke) and while investigating the related lambda-erasure bug #2082. Separate, unrelated, and more fundamental: simply DECLARING a field-like event of a user-declared named delegate type (`type T = delegate func(...) ...`) — with NO assignment at all — emits an `add_X`/`remove_X` accessor pair whose bodies fail ilverify. The bug is in the compiler-generated accessor bodies themselves, not in any call-site conversion.

## Repro (minimal G#)
```
package repro
import System

type TickHandler = delegate func(count int32) void

class Clock {
    public event Ticked TickHandler
}

func Main() {
}
```

## Actual
`gsc` compiles this without diagnostics, but ilverify fails on the emitted assembly:
```
[IL]: Error [StackUnexpected]: [...]Clock::add_Ticked([...]TickHandler)[...][found ref '[S.P.CoreLib]System.Delegate'][expected ref '[...]TickHandler'] Unexpected type on the stack.
[IL]: Error [StackUnexpected]: [...]Clock::remove_Ticked([...]TickHandler)[...][found ref '[S.P.CoreLib]System.Delegate'][expected ref '[...]TickHandler'] Unexpected type on the stack.
```
The synthesized accessor bodies appear to call `System.Delegate.Combine`/`Remove` (which return `System.Delegate`) and store/return the result directly as `TickHandler`, without casting back to the named delegate type.

## Expected
The generated `add_X`/`remove_X` accessors should cast the `Delegate.Combine`/`Remove` result back to the declared named-delegate type `TickHandler` before storing/returning it (mirroring how a CLR-imported event's accessors already do this correctly for e.g. `EventHandler`).

## Notes
- Independent of #2066 (nullable-local narrowing invoke dispatch — already fixed) and #2082 (lambda-to-named-delegate erasure on `+=`) — this reproduces with zero event subscriptions and no lambdas at all, so it is the more fundamental/root bug; #2082 may turn out to be a symptom of this same root cause once the accessor bodies are fixed.
- A field-like event of a CLR-imported delegate type (e.g. `EventHandler`) verifies fine; only a **user-declared named delegate type** event is affected.

Issue #2082, opened 2026-07-04
gsc: lambda-literal assigned to named-delegate-typed event (+=) erases to CLR Action/Func instead of the named delegate type


DavidObando:
## Summary
Discovered while fixing #2066 (nullable-local delegate snapshot invoke). A separate, unrelated bug: assigning an inline lambda/func-literal to a **user-declared named delegate** (`type T = delegate func(...) ...`) typed field-like event via `+=` compiles successfully but emits invalid IL — the lambda is erased to a CLR `System.Action<...>`/`Func<...>` closure delegate instead of being adapted to the declared named-delegate type `T`. ilverify fails on the generated `add_X`/`remove_X` accessors and any call site.

## Repro (minimal G#)
\`\`\`
package repro2066d
import System

type TickHandler = delegate func(count int32) void

class Clock {
    public event Ticked TickHandler
}

func Main() {
    let c = Clock()
    c.Ticked += func(count int32) void { System.Console.WriteLine(count) }
}
\`\`\`

## Actual
\`gsc\` compiles this without diagnostics, but the emitted assembly fails ilverify:
\`\`\`
[IL]: Error [StackUnexpected]: [...]Clock::add_Ticked([...]TickHandler)[...][found ref '[S.P.CoreLib]System.Delegate'][expected ref '[...]TickHandler'] Unexpected type on the stack.
[IL]: Error [StackUnexpected]: [...]Clock::remove_Ticked([...]TickHandler)[...][found ref '[S.P.CoreLib]System.Delegate'][expected ref '[...]TickHandler'] Unexpected type on the stack.
[IL]: Error [StackUnexpected]: [...]<Program>::Main()[...][found ref '[S.P.CoreLib]System.Action\`1<int32>'][expected ref '[...]TickHandler'] Unexpected type on the stack.
\`\`\`

## Expected
The lambda/func-literal should be adapted to the declared named-delegate type \`TickHandler\` (same as assigning a method group already works correctly — \`c.Ticked += OnTick\` for a top-level \`func OnTick(count int32) void\` compiles and verifies cleanly).

## Notes
- Assigning a **method group** to the same named-delegate-typed event (\`c.Ticked += OnTick\`) works and verifies cleanly — only the inline lambda/func-literal path is affected.
- Independent of #2066 (nullable-local narrowing invoke dispatch), which is a separate bind-time dispatch bug already fixed.
[...]

PR #2086, opened 2026-07-04
gsc: fix nullable-local delegate snapshot invoke crash (#2066)

### LINKED ISSUE #2066: gsc: nullable-local delegate snapshot invoke crashes at runtime
## Summary
Split from #1960 item 4. Assigning a field-like event to a nullable local of its delegate type, null-guarding it, then invoking it crashes gsc at runtime.

## Repro (minimal G#)
```
type TickHandler = delegate func(count int32) void

data class Clock
{
    public event Ticked TickHandler

    public shared void Fire(this Clock self, count int32)
    {
        let snapshot TickHandler? = self.Ticked
        if snapshot != nil
        {
            snapshot(count)
        }
    }
}
```

## Actual
Compiling/running this shape crashes gsc at runtime (nullable-local delegate snapshot + invoke). Root cause not yet isolated to a specific emitter/binder stage.

## Expected
The null-guarded snapshot should invoke normally (equivalent to C#'s standard "snapshot the delegate to a local before invoking to avoid a race" pattern), same as invoking the event field directly.

## Notes
- Independent of #1960 items 1-3 (cs2gs translator work); this is a gsc compiler/runtime bug.
- Add a gsc regression test once fixed (nullable delegate-typed local snapshot of a field-like event, guarded then invoked).

DavidObando:
Fixes #2066

A nil-guarded nullable local of a delegate type (a named delegate type declared with `type T = delegate func(...) ...`, or a native function-typed local `(T) -> R`) failed to bind when invoked with direct call syntax after the guard (`if snapshot != nil { snapshot(x) }`), reporting `GS0131 'snapshot is not a function'` at compile time.

**Root cause:** the direct-call binders in `OverloadResolver` dispatch by pattern-matching the callee `VariableSymbol`'s declared `Type` against the concrete (non-nullable) `FunctionTypeSymbol`/`DelegateTypeSymbol`/CLR-delegate shapes. Nil-guard smart-cast narrowing (Phase 3.C.4) records that the variable is non-nullable inside the guarded scope, but that narrowing frame was consulted only by ordinary variable-read binding (`ExpressionBinder`), not by these call-site dispatch checks, so a narrowed nullable-delegate local's declared (nullable) type never matched any of the concrete shapes and fell through to the 'not a function' diagnostic.

**Fix:** add `OverloadResolver.GetNarrowedCallableType`, mirroring `ExpressionBinder`'s narrowed-type lookup over the active `BinderContext.NarrowedVariables` frames, and consult it in the `FunctionTypeSymbol` and `DelegateTypeSymbol` direct-call dispatch checks. Thread the narrowed (non-nullable) type through to the synthesized callee read (`BoundVariableExpression`'s `narrowedType` parameter) so downstream emit's delegate-shape dispatch (`MethodBodyEmitter.EmitIndirectCall`) also observes the narrowed shape instead of the raw nullable declared type. Generalizes to any nullable-local delegate type, arity, and return type (void or non-void).

**Tests:** added `Issue2066NullableDelegateSnapshotEmitTests` covering a user-declared named delegate type (void and non-void return) and a native function-typed local — all compile, ilverify, and run correctly end to end.

**Out of scope (filed separately):** while investigating, found two separate, unrelated, pre-existing gsc bugs and filed them rather than fixing here:
- #2082: assigning an inline lambda/func-literal to a named-delegate-typed event via `+=` erases to a CLR Action/Func closure instead of the named delegate type.
- #2085: a field-like event of a named delegate type emits invalid add_/remove_ accessor IL regardless of assignment.

Note: there is a concurrent PR #2079 from another agent also targeting #2066; this PR was produced per an independent task assignment on branch `fix/2066-nullable-delegate-snapshot`.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/OverloadResolver.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolver.cs
@@ -3906,11 +3906,28 @@ internal sealed class OverloadResolver
     /// </description></item>
     /// </list>
     /// </summary>
+    /// <param name="syntax">The call-expression syntax.</param>
+    /// <param name="variable">The callee variable.</param>
+    /// <param name="fnType">The delegate's equivalent function-type shape.</param>
+    /// <param name="args">The already-bound, already-converted call arguments.</param>
+    /// <param name="targetNarrowedType">
+    /// Issue #2066: when the callee variable's declared type is nullable and
+    /// an active smart-cast frame has narrowed it (the direct-call binders
+    /// only reach this helper once a narrowed non-nullable delegate/function
+    /// shape has been matched), the concrete narrowed type to report from the
+    /// synthesized target read's <see cref="BoundExpression.Type"/> — for
+    /// example the actual <see cref="DelegateTypeSymbol"/> for a named
+    /// delegate, so downstream emit's delegate-shape dispatch (<see
+    /// cref="Emit.MethodBodyEmitter.EmitIndirectCall"/>) sees the narrowed
+    /// shape rather than the raw nullable declared type. <c>null</c> when the
+    /// variable's declared type is already non-nullable.
+    /// </param>
     private BoundExpression BuildIndirectDelegateCall(
         CallExpressionSyntax syntax,
         VariableSymbol variable,
         FunctionTypeSymbol fnType,
-        ImmutableArray<BoundExpression> args)
+        ImmutableArray<BoundExpression> args,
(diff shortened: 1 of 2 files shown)


DavidObando:
Closing as duplicate of earlier, broader #2079 (also fixes #2066, additionally covering async/closure capture paths). Branch fix/2066-nullable-delegate-snapshot preserved as fallback if #2079 stalls. Filed follow-ups #2082 and #2085 remain valid regardless.

Issue #2201, opened 2026-07-06
cs2gs: Roslyn source generators not run (MVVM Toolkit [ObservableProperty]/[RelayCommand], NB.GitVersioning ThisAssembly)


DavidObando:
## Blocker: Roslyn source generators not run during migration

Discovered while migrating the Oahu C# corpus to G# with `cs2gs migrate`.

cs2gs translates hand-written source but does not run (or emulate) Roslyn **source generators**, so any symbol a generator would produce is missing from the translated G#, and every reference to it fails to resolve. Two concrete generator families block Oahu:

### 1. CommunityToolkit.Mvvm (MVVM Toolkit) — `[ObservableProperty]` / `[RelayCommand]`
The generator turns an annotated field into a public property, and an annotated method into an `ICommand` property. Consumers (and XAML bindings) reference the generated members, which don't exist post-translation.

Repro:
- `Oahu.UI/ViewModels/MainWindowViewModel.cs:11,14` `[ObservableProperty]`
- `Oahu.UI/ViewModels/SettingsViewModel.cs:109,122` `[RelayCommand]`
- also `AboutViewModel`, `ConversionViewModel`, `BookLibraryViewModel`, ...

Dominant root cause in `Oahu.UI` (45 compile errors).

### 2. Nerdbank.GitVersioning — `ThisAssembly`
The build injects a generated `ThisAssembly` class with version constants.

Repro:
- `Oahu.Foundation/Auxiliary/ApplEnv.cs:26` `public static string AssemblyVersion { get; } = ThisAssembly.AssemblyFileVersion;`
- `Oahu.UI/ViewModels/AboutViewModel.cs`

Note: gsc PR #2199 fix #1 already ensures an unqualified `ThisAssembly` no longer mis-binds to the framework-internal `System.Diagnostics.ThisAssembly`; it now reports an honest "cannot find type". Materializing the real generated class is the remaining work.

### Possible directions
- Run the compilation's source generators during migration and translate the generated trees, OR
- Special-case the well-known generators (MVVM Toolkit attribute-driven members; NB.GitVersioning `ThisAssembly` from the build's version).

### Impact
Blocks `Oahu.UI` almost entirely; partial block on `Oahu.Foundation`.

DavidObando:
Resolved via
- #2219 
- #2220 
- #2221

PR #2208, opened 2026-07-07
Roslyn source-generator host for native G# projects (ADR-0145)

DavidObando:
## Summary

Implements **ADR-0145**: a Roslyn source-generator host (`gsgen`) so a native `.gsproj` that references a NuGet package containing a Roslyn source generator works like the equivalent C# project — the generator runs at build time and its output is compiled by gsc. Since gsc is Roslyn-free (ADR-0027), the generators run in a sibling tool against a **C# projection** of the G# program, and their output is translated back to G# `partial` parts.

**Proven with a real `dotnet build`:** a sample referencing `CommunityToolkit.Mvvm` resolves the generator, runs it, and gsc consumes the emitted `.g.gs` (see fidelity note below). The no-generator fast path is unchanged.

Addresses the native-generator portion of #2201. Builds on partial types (#2207).

## How it works

```
user .gs  ──parse+bind──►  C# stubs (declaration-only, attributes by CLR metadata name)
                                │  ForAttributeWithMetadataName matches here
                                ▼
                     CSharpGeneratorDriver runs the package's generators
                                │
                     generated C# ──back-translate (preserve-partial)──►  obj/gsgen/*.g.gs
                                │
                          injected into @(Compile) ──►  gsc
```

## What's included (7 commits, each green)

- **Preserve-partial translator mode** — cs2gs gains an opt-in mode so each generated C# `partial` part becomes a standalone G# `partial` part (augmentation), distinct from migration's merge-to-one. Default behavior unchanged.
- **Generator-host engine** (`tools/gsgen/GSharp.GeneratorHost`) — G#→C# type speller, stub projection, `CSharpGeneratorDriver` execution (crash-isolated, with an analyzer-assembly loader), and back-translation. 19 unit tests including a full-loop e2e where a generated member becomes visible to gsc.
- **Partial-method elision (ADR-0143 §D)** — unimplemented `partial void` hooks and their call sites are dropped; implemented pairs emit once (needed for real generator output).
- **`gsgen` CLI** (`tools/gsgen/Gsgen.Cli`) — `gsgen @file.rsp`, diagnostics in gsc's line format.
- **MSBuild SDK integration** — `GsgenTask`, `_GsharpResolveAnalyzers` (re-invokes `ResolvePackageAssets` with `ProjectLanguage=C#` + `$(GsharpCompilerApiVersion)` — the fix that makes a `.gsproj`'s generators resolve at all; `@(Analyzer)` was never populated for G# projects), `_GsharpRunSourceGenerators` before `CoreCompile`, and `PackGsgen` packaging. `samples/SourceGen` + `e2etests/gsgen-e2e.sh`.
- **Translator split** — `Cs2Gs.ProjectLoading` extracted (namespace kept, assembly moved → no call-site churn) so the shipped tool doesn't carry MSBuildWorkspace/Build.Locator.
- **Language server** — generated `obj/**/gsgen/*.g.gs` are surfaced to the LS compilation and watched, so generated members resolve in the editor after a build.

## Testing

- `GSharp.GeneratorHost.Tests`: **19** (stub golden + parses-as-C#, generator run, crash isolation, full-loop visibility, CLI incl. a real on-disk generator DLL via `/analyzer:`).
- `Cs2Gs.Tests`: **1186** (adds preserve-partial + partial-method elision; migration behavior unchanged).
- `LanguageServer.Tests`: **419** (adds generated-file discovery + resolution).
- Full solution builds clean; SDK packs `gsgen`; `e2etests/gsgen-e2e.sh` runs a real `dotnet build` and confirms gsgen executes and feeds gsc; the no-generator sample stays green.

## Known follow-ups (documented in ADR-0145 "Implementation status")

- **CommunityToolkit.Mvvm back-translation fidelity** — MVVM's generated code trips genuine cs2gs/gsc gaps (`GS0157` unqualified/inherited member resolution, `GS0130` inherited `ObservableObject` methods, `GS0202` an attribute-arg constant). These are the §H fidelity boundary, to be filed as ADR-0138 triage items. Generators whose output gsc can already compile work fully (proven by the host's e2e). The MVVM e2e therefore asserts the real integration property (generator ran, output produced, gsc consumed it) and auto-upgrades to a full run once the gaps close.
- **Live in-editor regeneration** — the out-of-proc `gsgen serve` sidecar (§G). This PR delivers post-build editor visibility; keystroke regeneration is deferred.


DavidObando:
Filed the CommunityToolkit.Mvvm back-translation fidelity gaps discovered by this PR's e2e as follow-up issues (all with minimal repros):

- #2209 — **gsc**: static member access through a fully-qualified generic type name fails (`GS0157`; e.g. `System.Collections.Generic.Comparer[int32].Default.Compare(...)`). Root cause of MVVM's `EqualityComparer<T>.Default`.
- #2210 — **gsc**: unqualified call to a method inherited from an imported base class fails (`GS0130`). Root cause of MVVM setters calling inherited `OnPropertyChanging`/`OnPropertyChanged`.
- #2211 — **cs2gs**: back-translation emits short type/enum names without the required import (`GS0113`/`GS0157`/`GS0202`).

Once these close, `e2etests/gsgen-e2e.sh` auto-upgrades from asserting the integration property to a full green run of the MVVM sample.

PR #2213, opened 2026-07-07
docs: reframe ADR-0143 — reproduce generated code at build, don't freeze

DavidObando:
## Summary

Reframes **ADR-0143** (source generators in cs2gs migration) in light of what shipped since it was drafted — partial types (#2207) and the gsgen native generator host (#2208).

The original draft proposed running Roslyn generators *inside* cs2gs and translating/freezing their output. That duplicates work the G# build now does and freezes stale snapshots. The rewrite flips the principle to **reproduce at build, don't freeze**:

- **Roslyn generators (MVVM, …)** → reproduced by **gsgen (ADR-0145)** at build. cs2gs keeps the trigger intact (attributes survive, type emits `partial`, package reference carries over); the members regenerate on every build.
- **`ThisAssembly` / version constants** → synthesized by the G# build (gsc already receives the version), following the ADR-0142 resx precedent — not frozen.
- **SDK-derived `AssemblyInfo` / `GlobalUsings`** → reproduced by the SDK from preserved project properties.
- **Classification by provenance, not type-content** → skip a generated file only when the build reproduces it; *translate* the hand-authored non-type source the old type-declares gate wrongly dropped (assembly attributes, hand-written global usings). G# already has the assembly-attribute surface, so this is a translator/classification change, not a language feature.

The net is that ADR-0143 shrinks dramatically and shifts new work onto the build system rather than a large cs2gs translator surface. The original "run generators in cs2gs and freeze" approach is recorded in Alternatives, rejected.

## Enabling items (filed separately)
- `ThisAssembly` synthesis in gsc/the SDK.
- Whether gsc should auto-trigger gsgen so non-MSBuild callers (the cs2gs migration pipeline, tests) run generators — today gsgen is MSBuild-target-orchestrated and gsc is generator-agnostic.

Docs-only change.


DavidObando:
Enabling items filed:
- #2214 — **gsc/SDK: synthesize `ThisAssembly` natively** (ADR-0143 §C relies on this instead of freezing a translated `ThisAssembly.cs`).
- #2215 — **gsc: trigger gsgen automatically when analyzers are supplied**, so the cs2gs migration pipeline (and any non-MSBuild gsc caller) runs generators. Confirmed the current state: gsgen runs only via the SDK's `_GsharpRunSourceGenerators` MSBuild target and gsc is generator-agnostic, so gsc doesn't trigger gsgen today — #2215 tracks making it do so (the preferred design).

Issue #2225, opened 2026-07-07
gsc/nbgv: ThisAssembly not generated for G# — Nerdbank.GitVersioning has no G# language support (GS0157, Oahu.Foundation)


DavidObando:
Discovered migrating the Oahu corpus (`Oahu.Foundation`) with `cs2gs migrate`.

```
ApplEnv.gs(19,38): error GS0157: Cannot find type ThisAssembly. Are you missing an import?
```

`ThisAssembly` is source-generated by **Nerdbank.GitVersioning (nbgv)**. nbgv only emits the `ThisAssembly` file for the languages it supports (`<Language>` = C#, VB, F#); there is no G# emitter, so the type is never materialized for a translated G# project.

**Known upstream gap:** the fix ultimately belongs in nbgv (add a G# `ThisAssembly` code generator, selected via the project `<Language>`). Tracking here so the Oahu migration item is visible. Until nbgv supports G#, this project can't fully compile from a pure translation.

**Possible interim mitigation:** cs2gs/gsgen could synthesize a minimal `ThisAssembly` (the members Oahu references: version/informational-version constants) so the translated project compiles without the upstream nbgv change.

Label: Oahu.

DavidObando:
PR open:
- dotnet/Nerdbank.GitVersioning#1403

DavidObando:
We'll now wait for a version of nbgv that includes G# support. Currently the change is in the version 3.11 milestone train.

DavidObando:
Nerbank.GitVersioning version 3.11.13-beta with G# support has been released.

https://www.nuget.org/packages/Nerdbank.GitVersioning/3.11.13-beta

DavidObando:
[...]

Issue #1933, opened 2026-07-03
cs2gs pipeline: ilverify stage needs an unsafe-IL policy (pointer/stackalloc IL is unverifiable by design)


DavidObando:
Stage 3 has no unsafe exemption: pointer writes, `fixed`, and `stackalloc` produce IL that the pinned ilverify rejects (ExpectedNumericType/Unverifiable localloc/UnmanagedPointer/StackByRef) — and the csc-compiled C# BASELINE fails identically, so this is not a gsc defect. Grid coverage of the unsafe surface (G12) is limited to verifiable constructs until the stage grows a per-app or per-category policy (e.g. `ilverify: allow-unsafe` in corpus metadata mirroring the ignore bundles).
## Definition of done
- [ ] Fix + `Issue<N>*.cs` regression test
- [ ] Re-enable the quarantined grid fixture (or add one) and re-capture its baseline
- [ ] `tools/cs2gs/coverage/csharp-construct-inventory.json` row updated

_Discovered by the ADR-0138 conformance-grid build-out (gsc 0.2.585+01ad6ec964); quarantined sources live next to each grid app under `tools/cs2gs/corpus/grid/`._

PR #2257, opened 2026-07-08
Fix qualified/dotted type-name resolution for source types (#2256)

### LINKED ISSUE #2256: gsc: qualified (dotted) generic type-parameter constraints and package-qualified source-type references fail to bind
## Summary
Discovered while migrating Oahu (`Oahu.Decrypt`) via cs2gs. cs2gs fully-qualifies type references (including generic-math constraints), e.g. the C# `where TCallback : INewSplitCallback<TCallback>` is emitted as the G# constraint `[TCallback Oahu.Decrypt.INewSplitCallback[TCallback]]`. Three distinct gsc bugs combine to block this:

### 1. Parser rejects a dotted type-parameter constraint (GS0005)
`func F[T Ns.Sub.IFace[T]](...)` failed to parse — `ParseTypeParameter` only accepted a single constraint identifier token (plus an optional `[...]` arg list) with no handling for a `.`-qualified name; `LooksLikeTypeParameterList` also didn't skip `.` tokens, so the whole type-parameter list was misclassified as an indexer.

Minimal repro:
```gsharp
func Max[T System.IComparable[T]](a T, b T) T { ... }   // GS0005 Unexpected token <DotToken>
```

### 2. Reference-set (cross-assembly) qualified generic type name fails (GS0113)
A namespace-qualified *generic* referenced type — `A.B.IFoo[int32]` where `IFoo` lives in a referenced assembly — reported "Type 'A.B.IFoo' doesn't exist", because the greedy dotted-name prefix walk tried the un-mangled full name (`A.B.IFoo`) rather than the arity-mangled metadata name (`A.B.IFoo\`1`). Non-generic qualified types resolved fine.

### 3. Same-compilation package-qualified source type fails (GS0113 / GS0157)
A source type referenced by its own package-qualified name — `A.B.IFoo` (generic or not) from within package `A.B` — reported "doesn't exist". Source types are visible by simple name across packages, but the reflection-based prefix walk only sees types with a CLR representation, which source types lack while binding. This is also the root of the widespread `Oahu.Audible.Json.Chapter` GS0157 "Cannot find type Oahu" cascade seen across the migration.

## Fix
- Parser: parse a dotted constraint via the standard type-clause machinery into a new `TypeParameterSyntax.ConstraintType`; skip `.` tokens in `LooksLikeTypeParameterList`.
- Binder: try the arity-mangled name for a fully-qualified generic type prefix; add a package-qualified source-type fallback (simple-name lookup of the final segment, honoring arity) in `BindQualifiedTypeName`.

All three verified fixed via minimal repros incl. the self-referential CRTP shape `[TCallback Oahu.Decrypt.INewSplitCallback[TCallback]]` in-package.

DavidObando:
Fixes #2256.

cs2gs fully-qualifies every type reference for safety, but a G# **source** type has no CLR type at bind time and is visible by *simple name* across a flat, cross-package scope. gsc failed to resolve several package-qualified source-type shapes that cs2gs emits, which blocked the **Oahu.Decrypt** migration (translation-unsupported -> 10 compile errors -> 0).

## Fixes

1. **Parser — dotted type-parameter constraints.** `func Register[TCallback Oahu.Decrypt.INewSplitCallback[TCallback]](...)` previously failed `GS0005`. Adds `TypeParameterSyntax.ConstraintType`, parses the constraint via `ParseTypeClause`, and skips `.` in `LooksLikeTypeParameterList`.
2. **Binder (type position).** Resolves a reference-set qualified generic type `A.B.IFoo[int32]` by trying the arity-mangled outer prefix, and a same-compilation qualified source type `A.B.IFoo` by simple-name fallback (previously `GS0113`/`GS0157`).
3. **ExpressionBinder (expression position).** Peels a redundant namespace/package prefix from a qualified source-type construction `Ns.Type[Args](...)` and static call `Ns.Type[Args].StaticMethod(...)`, binding the terminal by simple name. Generic type receivers parse as index expressions, so that shape is handled too. Generalized to any `Namespace.Type[...]`, not just the specific Oahu type that triggered it.
4. **DeclarationBinder.** Binds the new `ConstraintType` when present.

## Validation

- New `Issue2256QualifiedConstraintTests` (6 tests) pin all shapes.
- Full **Core.Tests** suite passes (5886 passed, 1 skipped).
- **Oahu.Decrypt** now migrates **fully green** (translate/compile/ilverify/test-parity all PASS) via cs2gs.

Note: the ilverify stage surfaced 13 expected-unverifiable `stackalloc`/`localloc` methods in the binary/crypto parsers; the C# compiler emits identical unverifiable IL for the same source, so these are handled by an `ilverify.allow-unsafe` marker in the Oahu corpus (issue #1933/#1985 mechanism), not a gsc change.
---

## Additional fixes — qualified CLR access & bare-type patterns (#2258)

Batched into this branch (related qualified-name resolution; amortizes CI):

5. **gsc — fully-qualified CLR static/enum member access.** `Microsoft.Extensions.Logging.LogLevel.Warning` (deep-namespace CLR enum member) and `…Abstractions.NullLoggerFactory.Instance` (CLR static field) failed `GS0157` in expression position: accessors are right-associative, so the leading `Microsoft` segment (matching no registered import) hit `ReportUnableToFindType`. `System.X.Y` worked only because `System` is implicitly imported. Extracted the progressive-prefix CLR-type walk out of `TryBindImportAccessor` into a shared helper and added a last-resort fallback that starts the walk from the bare leading name — generalizing to any referenced CLR type at any namespace depth.
6. **cs2gs — qualified bare-type switch arm.** Roslyn parses `App.Auth.MfaChallenge =>` (bare-type switch arm) as a `ConstantPattern` over a `MemberAccessExpression` (not a `TypeSyntax`), so cs2gs rendered it as a bare value expression that gsc rejects (`GS0157`). `IsTypeReferencePattern` now relies on the resolved symbol being an `ITypeSymbol` (an enum-member constant resolves to an `IFieldSymbol`), lowering these to `_ is T`. Generalized across boolean `is`, negated `is not`, and switch-arm type patterns.

Also partially addresses #2258 (the CLR struct object-initializer literal shape — `System.Text.Json.JsonWriterOptions{ … }` — remains, tracked in #2258; it fails even by simple name, so it's a distinct missing feature).

Validation: full **Core.Tests** re-run (5889 passed, 1 skipped); cs2gs Pattern/Switch/Golden suites (192+ tests) pass; new regression tests added (`Issue2256…EnumMemberAccess`, `Issue1890…QualifiedBareTypeArm`). Verified end-to-end: Oahu.Cli.Tui `Microsoft.*` and `App.Auth.*` `GS0157` errors eliminated.

---

## Additional features — nbgv version bump (#2225) & `--via-sdk` compile path (#2261)

Batched into this branch (amortizes CI). Both are additive; existing defaults unchanged.

7. **cs2gs — Nerdbank.GitVersioning floor bump (#2225).** nbgv only generates the `ThisAssembly` file for languages its MSBuild `<Language>` supports; G# support first shipped in `3.11.13-beta`. `NerdbankGitVersioningPolicy` (prerelease-aware SemVer compare + MSBuild XML rewrite) scans the app's `.csproj` and ancestor `Directory.Packages.props`/`Directory.Build.props`; when a referenced nbgv version is below the floor it emits a bumped copy into `<AppRunDir>/nbgv-bump/` with a manifest (source is never mutated). 27 unit tests. Verified on Oahu.Foundation: `3.7.115` -> `3.11.13-beta` (single-line diff).

8. **cs2gs — `--via-sdk` compile path (#2261).** Opt-in flag to compile migrated G# via `dotnet build` + the `Gsharp.NET.Sdk` MSBuild SDK instead of gsc directly, so Roslyn source generators (CommunityToolkit.Mvvm `[ObservableProperty]`/`[RelayCommand]`, `[LoggerMessage]`) run via gsgen's MSBuild targets. `SdkCompileRunner` reconstructs `PackageReference`/`Reference`/`Analyzer` items from the gsc reference set, writes an isolated `.gsproj` + local `nuget.config`, invokes `dotnet build`, and reuses `GscInvoker.ParseDiagnostics`. Plain `migrate` is byte-identical; `--via-sdk` falls back to gsc-direct if no local SDK nupkg exists. Verified on Oahu.UI: generated `OpenDownloadsFolderCommand` (GS0157), `RelayCommand` not-an-attribute (GS0200), and GS9100 all eliminated.
9. **gsc core — attribute-name resolution (ECMA-334 §22.3).** Coupled fix for #2261: `DeclarationBinder.ResolveAttributeType` no longer lets an exact-name candidate that isn't `System.Attribute`-derived (e.g. CommunityToolkit's `RelayCommand` ICommand class) shadow the `...Attribute`-suffixed candidate.

Validation: full **Core.Tests** (5889 passed, 1 skipped), 40 cs2gs nbgv/SdkCompileRunner tests, 119 attribute tests — all green.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/Binder.cs
+++ b/src/Core/CodeAnalysis/Binding/Binder.cs
@@ -3315,7 +3315,13 @@ public sealed class Binder
         // by a single-name `Probe` that happens to exist somewhere.
         for (var outerLen = totalSegments; outerLen >= 1; outerLen--)
         {
-            var clrType = TryResolveOuterPrefix(segmentTexts, outerLen);
+            // When the whole dotted name IS the (generic) type — the trailing
+            // type-argument list belongs to the deepest prefix segment — the
+            // metadata name is arity-mangled (`Ns.IFoo`1`). Pass the target
+            // arity so a namespace-qualified generic user/CLR type resolves
+            // (issue: qualified generic type-name/constraint resolution).
+            var prefixArity = outerLen == totalSegments ? targetArity : 0;
+            var clrType = TryResolveOuterPrefix(segmentTexts, outerLen, prefixArity);
             if (clrType == null)
             {
                 continue;
@@ -3328,6 +3334,37 @@ public sealed class Binder
             if (walked != null)
             {
                 return ConstructIfGeneric(walked, syntax, targetArity);
+            }
+        }
+
+        // Same-compilation package-qualified source type: the qualifier segments
+        // form a package/namespace prefix and the final segment names a source
+        // type declared in this compilation (e.g. `Oahu.Decrypt.INewSplitCallback[T]`
+        // referenced from within package `Oahu.Decrypt`). Source types are visible
+        // by simple name across packages, but the reflection-based prefix walk
+        // above only sees types with a CLR representation, so a source type — which
+        // has none while binding — never resolves through it. Fall back to a
+        // simple-name lookup of the final segment, honoring the trailing arity.
+        // cs2gs fully-qualifies type references (including generic-math
+        // constraints), so this is the common shape for translated code.
+        var lastSegment = segmentTexts[totalSegments - 1];
+        var sourceType = LookupType(lastSegment, targetArity > 0 ? targetArity : -1);
+        if (sourceType != null && !ReferenceEquals(sourceType, TypeSymbol.Error))
+        {
+            if (targetArity == 0)
+            {
+                return sourceType;
+            }
+
+            var constructed = BindAndConstructUserGenericSegment(
+                syntax,
+                sourceType,
+                syntax.TypeArguments,
+                syntax.Identifier.Location,
+                syntax.DottedName);
+            if (constructed != null)
+            {
+                return constructed;
             }
         }
 
@@ -3394,15 +3431,37 @@ public sealed class Binder
     /// prefixes, and the active import set as a namespace prefix for
     /// multi-segment prefixes.
     /// </summary>
-    private Type TryResolveOuterPrefix(string[] segmentTexts, int outerLen)
+    private Type TryResolveOuterPrefix(string[] segmentTexts, int outerLen, int lastSegmentArity = 0)
     {
         if (outerLen == 1)
         {
-            var symbol = LookupType(segmentTexts[0]);
+            var symbol = LookupType(segmentTexts[0], lastSegmentArity > 0 ? lastSegmentArity : -1);
             return symbol?.ClrType;
         }
 
         var prefix = string.Join(".", segmentTexts, 0, outerLen);
+
+        // A generic type's metadata name is arity-mangled (`Ns.IFoo`1`); when
+        // the deepest prefix segment carries the trailing type-argument list,
+        // try the mangled name first so a namespace-qualified generic type
+        // resolves to its open definition (closed later by ConstructIfGeneric).
+        if (lastSegmentArity > 0)
+        {
+            var mangled = prefix + "`" + lastSegmentArity;
+            if (scope.References.TryResolveType(mangled, out var directGeneric))
+            {
+                return directGeneric;
+            }
+
+            foreach (var import in scope.GetDeclaredImports())
+            {
+                if (scope.References.TryResolveType(import.Target + "." + mangled, out var viaImportGeneric))
+                {
+                    return viaImportGeneric;
+                }
+            }
+        }
+
         if (scope.References.TryResolveType(prefix, out var direct))
         {
             return direct;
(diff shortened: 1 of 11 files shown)


DavidObando:
Added nbgv version-bump support for #2225 (commit 0159a92c). New `NerdbankGitVersioningPolicy` (prerelease-aware SemVer compare + MSBuild XML rewrite) bumps a below-floor **literal** Nerdbank.GitVersioning version up to 3.11.13-beta (the first G#-capable release); at-or-above versions and non-literal versions (MSBuild property / floating / range) are left untouched ("if possible"). Wired into `TranslateStage`: scans the app's .csproj + ancestor Directory.Packages.props/Directory.Build.props and emits a bumped copy into `<AppRunDir>/nbgv-bump/` with a manifest — the corpus source tree is never mutated. Verified on Oahu.Foundation (Directory.Packages.props 3.7.115 → 3.11.13-beta, single-line diff).

PR #2311, opened 2026-07-11
feat: implement ADR-0147 structural literal-to-type assignability

khurram-uworx:
## Summary

Implements [ADR-0147](docs/adr/0147-structural-literal-assignability.md): **structural literal-to-type assignability** (TypeScript-style width subtyping).

An object literal (`object{...}` / `data object{...}`) is now implicitly assignable to a concrete value-type target (data struct / struct / class) when its shape satisfies all required members with compatible types. Extra literal members are allowed (width subtyping).

### New `Conversion.Structural` kind
- Added `Conversion.Structural` — an implicit conversion classified when a source `StructSymbol` (anonymous literal, no Declaration/ClrType) targets a concrete non-abstract value type.
- `IsStructural` property exposed on `Conversion`.

### Strict member validation at classify time
- `Conversion.Classify` checks that every required member of the target is present in the source literal with a compatible type **before** classifying as `Conversion.Structural`. Missing or incompatible members yield `Conversion.None` — the error surfaces as a type-mismatch diagnostic, not GS0490.
- With strict classification, GS0490 is a safety diagnostic that is currently unreachable for the structural-literal path.

### Lowering via `ConversionClassifier.BindStructuralRecordConversion`
- Rebuilds the source literal as the target `BoundStructLiteralExpression`, mapping each source member to the matching target field/property by name.
- Each mapped value runs through `BindConversion` for per-member implicit conversion (e.g., `int32` to `int64` widening).

### OverloadResolver structural argument lowering
- Structural literal to type conversions in function arguments are now lowered through `BindConversion` (previously only tuple/nullable were handled).

### Empty `object{ }` emit fix
- Fixed a pre-existing bug where `object{}` (zero-member anonymous object literal) crashed the emit pipeline and prevented the entry point from being generated.
- **`DataStructSynthesizer`**: Skip Equals/GetHashCode/ToString/Deconstruct synthesis when the anonymous type has no fields or properties (guarded `Debug.Assert` → conditional block). Always emit a parameterless primary constructor when `Fields` is empty so `BoundConstructorCallExpression` has a `.ctor` to reference.
- **`ReflectionMetadataEmitter`**: Guard `methodRow += 7` with a synthesis-fields check so zero-member types don't reserve 7 unused MethodDef rows. Changed primary-ctor row reservation condition from `Fields.IsDefaultOrEmpty && HasPrimaryConstructor` to `Fields.IsDefaultOrEmpty` to match the emitter.

### Diagnostic docs update
- Updated `docs/diagnostics.md` GS0490 section to reflect that with strict classification, missing members produce type-mismatch errors. GS0490 is a safety diagnostic currently unreachable for the structural-literal path.

### Tests and samples
- 8 new unit tests covering required fields, width subtyping, missing members, type widening, local binding, interface targets, empty-literal-to-zero-member-target (emit path), and end-to-end emission.
- `samples/Animals.gs` end-to-end sample demonstrating all five ADR-0147 features: basic assignment, width subtyping, class targets, type widening, and empty-literal-to-zero-member-target.
- Regenerated PE baseline for `Animals.gs`.

## Out of scope (deferred)

- Interface-typed targets (requires synthesized InterfaceImpl rows).
- Structural conformance of named types (`: I` remains mandatory).
- Bare `{}` syntax (reuse `object{}` instead).

## Verification

- `dotnet build` succeeds
- 8 new unit tests pass (including emit path test for `object{}`)
- Core.Tests: 5979 passed, 0 failed
- Compiler.Tests Sample: 124 passed, 0 failed
- PE baseline: 138/138 passed
- `samples/Animals.gs` compiles and runs correctly with golden output match


DavidObando (changes requested):
## Review finding: critical bug — structural conversion fails when the literal isn't bound directly at the conversion site

`Conversion.Classify` classifies a `Structural` conversion purely from **types** (`src/Core/CodeAnalysis/Binding/Conversion.cs:1067-1077`): any expression whose type is an anonymous literal `StructSymbol` (`Declaration == null`) qualifies — including a variable/parameter holding such a literal, not just a fresh `object { ... }` expression.

However, the lowering step `GetObjectLiteralMemberValues` (`src/Core/CodeAnalysis/Binding/ConversionClassifier.cs:2284-2306`) only extracts member values when the *bound expression* is literally a `BoundStructLiteralExpression` or `BoundConstructorCallExpression`. For any other expression shape (e.g. a `BoundVariableExpression` referencing a `let`-bound literal), it returns an **empty dictionary**, so every required-member lookup fails and the compiler emits a bogus `GS0490 "missing required member"` diagnostic — even though the value is fully compatible.

This contradicts the ADR's own stated design (no freshness distinction between literals assigned directly vs. via a variable) and breaks the very common pattern of assigning a literal to a variable before use.

### Repro (verified by building `gsc` from this branch)

```gsharp
package GSharp.Example.Repro
import System

data struct Pet {
    var Name string
    var Age int32
}

func describe(p Pet) {
    Console.WriteLine("${p.Name} is ${p.Age}")
}

let lit = object { let Name = "Fido"; let Age = 4 }
describe(lit)
```
```
error GS0490: Object literal does not satisfy type 'Pet': missing required member 'Name'.
```

Same failure occurs on a plain variable-declaration conversion (no overload resolution involved), confirming the defect is in the shared lowering path, not specific to `OverloadResolver`:

```gsharp
let lit = object { let Name = "Fido"; let Age = 4 }
let p Pet = lit
```
```
error GS0490: Object literal does not satisfy type 'Pet': missing required member 'Name'.
```

All of the PR's new tests in `StructuralLiteralAssignabilityTests.cs` only exercise the conversion where the literal is bound *directly* at the conversion site (`describe(object {...})` / `let p Pet = object {...}`), so this regression path is untested.

**Suggested fix:** `GetObjectLiteralMemberValues` needs to handle arbitrary source expressions of the anonymous struct type — e.g. by binding member reads (`value.Name`) against the source expression instead of requiring the literal/constructor-call node shape, mirroring how normal member-access binding resolves fields/properties on any expression of a given `StructSymbol` type. Test coverage should also be added for the variable-holds-literal case.

DavidObando:
I found 2 different security concerns in this PR, and unfortunately I cannot take the contribution in its current shape. Given the current supply-chain threat environment I'll have to close this PR. If you're willing to continue making contributions let's engage via an issue where we can discuss ADR and implementation ahead of sending a diff, and that way we can avoid producing security-sensitive issues. Thank you for offering your energy into helping improve G#.

DavidObando:
@khurram-uworx I have produced a follow-up PR to implement your intention in a security-safe way, with additions to map between strongly-typed objects as well.

https://github.com/DavidObando/gsharp/pull/2372

Issue #567, opened 2026-06-07
Regression in 0.1.459: `error GS9998: Variable 'X' has no local slot` when a closure reads a field on a captured reference-type local


DavidObando:
## Summary (regression in 0.1.459)

A function literal that captures a reference-type local and reads or writes a **mutable field** on it triggers an internal compiler error:

```
error GS9998: Variable 'X' has no local slot or parameter index in the current method.
```

The diagnostic location is reported as `(1,1,1,1)` — i.e. no real source location — so the user has no signal about which line in their file is at fault.

This exact pattern compiled and ran correctly on `Gsharp.NET.Sdk/0.1.431` (it was the documented workaround for the previous #523 "closures capture value-type locals by value" bug — the original Oahu G# port for `tests/Oahu.Cli.Tests/Tui/CtrlCStateTests.cs` used a `ClockHolder` helper class for exactly this reason). The fix that landed for #523 in 0.1.459 appears to have introduced this regression in the closure environment emit.

## Repro

```gsharp
package Probe

import System
import Xunit

type ClockHolder class {
    Value DateTimeOffset
    init() {}
}

type GS9998_Probe class {
    @Fact
    func Closure_Reads_Holder_Field() {
        var clock = ClockHolder()
        clock.Value = DateTimeOffset.UtcNow
        var getter = func() DateTimeOffset { return clock.Value }
        let v = getter()
        Assert.True(v > DateTimeOffset.MinValue)
    }
}
```

## Observed

```
/private/tmp/probe/Tests.gs(1,1,1,1): error GS9998: Variable 'clock' has no local slot or parameter index in the current method.
```

Build fails.

## Workaround

For the Oahu test that hit this, deleting the `ClockHolder` indirection and capturing the `DateTimeOffset` local directly works — because the now-correct #523 fix makes `var now = DateTimeOffset.UtcNow; ... now = now.AddSeconds(1.0)` see the update through the closure. So in this case the regression is cosmetic.

However: **any case where the reference-type holder is necessary** — e.g. capturing across multiple closures that need to share the same mutation, or capturing a list/dict the closure must update — will hit this without a workaround. Worth fixing.

## Suspect

Closure environment emit appears to have a special case for "captured local is a value type → box it" (the new #523 path) that incorrectly handles "captured local is already a reference type → no boxing needed" — leaving the local without a properly registered slot in the display class.

## Context

`gsc` version: `Gsharp.NET.Sdk/0.1.459`, .NET 10 RTM.

[...]

Issue #503, opened 2026-06-06
Emit: closure-capturing lambda bound to a CLR event silently fails (MSB4181, no GS diagnostic)


DavidObando:
## Summary

When a function literal that **captures any outer variable** is bound to a CLR event with `+=`, `gsc` aborts the emit phase silently. MSBuild reports the opaque error `MSB4181: The "BuildTask" task returned false but did not log an error.` and no `GS#` diagnostic is produced.

A non-capturing lambda assigned to the same event compiles and runs fine, so the failure is specifically about closure environments crossing the lambda → CLR delegate boundary.

This blocks porting any test that asserts on event-handler side effects (`hits++` in a handler).

## Repro

```gsharp
package Probe

import System

type Source class {
    event Changed EventHandler
    func Raise() {
        Changed?.Invoke(this, EventArgs.Empty)
    }
}

type Counter class {
    Value int32
    init() { Value = 0 }
    func Increment() { Value = Value + 1 }
}

type Probe class {
    func Run() {
        var src = Source()
        var counter = Counter()

        // Capturing lambda → EventHandler. Silent emit failure.
        src.Changed += func(sender object, e EventArgs) {
            counter.Increment()
        }
        src.Raise()
    }
}
```

(In the original Oahu repro the source was a CLR `EventHandler` on a project-referenced C# class — `Oahu.Core.SettingsBase.ChangedSettings` — same failure shape.)

## Observed

```
.../Gsharp.NET.Core.Sdk.targets(42,5): error MSB4181: The "BuildTask" task returned false but did not log an error.
```

No `GS####` diagnostic and no stack trace; only the MSBuild wrapper error.

Side-effect: when an earlier build succeeded with different code, the stale assembly remains in `bin/` and `dotnet test --no-build` happily runs it, which masks the failure unless one always `rm -rf bin obj` between attempts.

## Workarounds that **also fail**

| Attempt | Outcome |
|---|---|
| `src.Changed += this.OnHit` (method-group conversion) | `GS0158: Cannot find member OnHit` |
| `src.Changed += OnHit` (bare instance-method reference) | `GS0125: Variable 'OnHit' doesn't exist` |
| Capture a class instance instead of a slice cell | same silent MSB4181 |
| Inline the lambda at the `+=` site rather than binding it to a `let` | same silent MSB4181 |

A non-capturing lambda works (see `samples/EventSubscription.gs`), so the regression is specifically the closure conversion path.

## Expected

[...]

PR #2332, opened 2026-07-14
Fix #2331: async this-capturing nested lambdas fail with GS9998

### LINKED ISSUE #2331: gsc: async method lambda capturing this retains stale parameter and fails GS9998
## Summary

A nested lambda that captures `this` inside a Task-returning `async func` fails during emit because async state-machine rewriting hoists `this` but does not propagate that rewrite into separately stored nested lambda bodies.

This is the next Oahu.Foundation blocker after #2330.

## Minimal reproduction

```gsharp
package Repro
import System
import System.Threading.Tasks
class Worker() {
    async func RunAsync() -> await Task.Run(() -> Do())
    func Do() { Console.WriteLine("x") }
}
```

Actual result:

```text
error GS9998: InvalidOperationException: Variable 'this' has no local slot or parameter index in the current method.\n```\n\nNon-async methods with the same lambda compile; async lambdas capturing only ordinary locals compile. The trigger is specifically a non-iterator async method plus a nested lambda capturing `this` implicitly or explicitly.\n\nReal Oahu occurrence:\n\n```csharp\npublic async Task CleanupAsync() => await Task.Run(() => Cleanup());\n```\n\ncs2gs translates this correctly.\n\n## Root cause\n\n`CaptureBoxingRewriter` runs before `AsyncStateMachineRewriter`, so the nested lambda initially captures the kickoff method’s `ThisParameter`. Async rewriting later hoists `this` into the state-machine field and rewrites the outer MoveNext body. `BoundTreeRewriter.RewriteFunctionLiteralExpression` intentionally does not descend into nested lambda bodies, and `AsyncStateMachineRewriter` does not separately update `program.Functions[lambdaSymbol]`. The closure invoke body therefore retains the stale unslotted kickoff `ThisParameter`.\n\nThis differs from async-iterator field issues (#655/#689) and catch/local boxing issues (#1437/#2329).\n\n## Expected fix\n\n- Propagate async hoisting into every lambda body transitively nested in the async method, following the existing program-function update pattern used by capture boxing (#567).\n- Ensure captured `this`, parameters, and locals use the correct hoisted/state-machine-backed value at closure construction/invocation.\n- Handle nested lambdas transitively without crossing unrelated lexical scopes.\n- Preserve already-working ordinary-local capture and non-async behavior.\n\n## Tests\n\n1. Async method + lambda implicit `this` capture via instance call.\n2. Explicit `this.field`.\n3. Lambda captures both `this` and an ordinary local.\n4. Doubly nested lambda capturing `this`.\n5. Exact Oahu `await Task.Run(() => InstanceMethod())` shape.\n\nDiscovered in Oahu migration round 7 on main after #2330.

### PR DESCRIPTION
**Pull request by DavidObando**
## Summary

Fixes #2331. A nested lambda that captures `this` inside a `Task`-returning `async func` instance method (implicitly via an instance call, explicitly via `this.field`, or transitively through a doubly-nested lambda) failed at emit with:

```
error GS9998: InvalidOperationException: Variable 'this' has no local slot or parameter index in the current method.
```

## Root cause

`this` is unconditionally hoisted into `ThisField` for every instance-method async state machine (`AsyncStateMachineTypeBuilder`). Two independent call sites need to know "is this variable hoisted, and to which field":

1. `MoveNextBodyRewriter.InnerRewriter.TryGetHoistedField` (private) — already special-cased `this` correctly.
2. The emitter's closure-construction site, `EmitCapturedVariableLoad` in `MethodBodyEmitter.Closures.cs` — which resolves hoisting through the **public** `AsyncStateMachineFieldMap.TryGetHoistedField` API. That public API only checked ordinary hoisted parameters/locals, not `this`.

So when a lambda is newly constructed inside the (already-rewritten) `MoveNext` body and captures `this`, the closure-construction site fell through to an ordinary variable load of the kickoff method's now-slot-less `ThisParameter`, and the emitter's slot lookup threw. The lambda's *own* `Invoke` body was never the problem — `CaptureRewriter` in `ClosureEmitter.cs` already rewrites captured-variable reads inside the lambda body correctly, independent of async lowering, mirroring the #567 pattern.

## Fix

- `AsyncStateMachineFieldMap.TryGetHoistedField` (the single public API consulted by every caller — the lowering rewriter *and* the emitter) now special-cases the kickoff method's `ThisParameter`, resolving it to `ThisField`, same as the ordinary parameter/local case.
- `MoveNextBodyRewriter`'s private duplicate of this logic is removed; it now delegates to the field map so there is exactly one source of truth, generalized transitively to any nesting depth of lambdas (doubly nested lambdas, mixed `this` + local/parameter capture, etc.) without special-casing any particular call site.
- No change to emitter slot-lookup logic (`SlotPlanner` / `MethodBodyEmitter.MemberAccess.cs`) and no new global cache — exactly as directed.

## Tests

Added `test/Compiler.Tests/Emit/Issue2331AsyncThisLambdaCaptureEmitTests.cs`, each compiling and running the emitted assembly end-to-end:

1. `AsyncInstanceMethod_LambdaImplicitThisCapture_InstanceCall_Runs` — the exact issue repro (`await Task.Run(() -> Do())`).
2. `AsyncInstanceMethod_LambdaExplicitThisField_Runs` — explicit `this.field` read.
3. `AsyncInstanceMethod_LambdaMixedThisAndLocalCapture_Runs` — `this` + local + parameter mixed capture.
4. `AsyncInstanceMethod_DoublyNestedLambdaCapturingThis_Runs` — a lambda nested inside a lambda, both reaching `this`.
5. `AsyncInstanceMethod_OahuTaskRunCleanupShape_Runs` — the exact real-world Oahu shape (`public async Task CleanupAsync() => await Task.Run(() => Cleanup());`).
6. `Control_NonAsync_LambdaCapturingThis_StillRuns` — non-async control stays green.
7. `Control_AsyncMethod_LambdaCapturingOnlyLocal_StillRuns` — pre-existing local-only async capture stays green.

## Validation

- Targeted: `dotnet test test/Compiler.Tests/Compiler.Tests.csproj -c Release --filter FullyQualifiedName~Issue2331AsyncThisLambdaCaptureEmitTests` — 7/7 passed.
- Broader async surface: `--filter FullyQualifiedName~Async` on both `Core.Tests` (329/329 passed) and `Compiler.Tests` (147/147 passed).
- Full suites: `Core.Tests` 5986/5987 passed (1 pre-existing unrelated `RegenerateBaseline` skip); `Compiler.Tests` 2834/2834 passed.

## Scope / deferred work

- `Oahu` was not modified, per instructions.
- `AsyncCaptureWalker`'s tree walk (used to decide which *ordinary* locals get hoisted) still does not descend into nested lambda bodies when collecting locals-only-referenced-inside-a-lambda; this is pre-existing and out of scope here since `this` and all kickoff parameters are hoisted unconditionally regardless of that walk, which is what this issue is about. If a future issue surfaces an ordinary local referenced *only* inside a nested lambda failing to hoist correctly across an `await`, that would need a separate transitive-descent fix in `AsyncCaptureWalker`.

### DIFF
--- a/src/Core/CodeAnalysis/Lowering/Async/AsyncCaptureWalker.cs
+++ b/src/Core/CodeAnalysis/Lowering/Async/AsyncCaptureWalker.cs
@@ -159,12 +159,110 @@ public static class AsyncCaptureWalker
             return base.RewriteVariableDeclaration(node);
         }
 
+        /// <summary>
+        /// Issue #2331 (deferred half): a parameter or local whose only
+        /// reference anywhere in this async body is inside a nested lambda —
+        /// including a lambda nested inside that lambda — must still
+        /// contribute to the hoist set. Otherwise its value does not survive
+        /// suspension across an intervening <c>await</c>: a resumed
+        /// <c>MoveNext</c> call is a fresh invocation of the same method, so
+        /// any variable that was not hoisted into a state-machine field is
+        /// re-initialized to the CLR default instead of retaining the value
+        /// it held before suspension.
+        /// </summary>
+        /// <remarks>
+        /// <see cref="BoundTreeRewriter"/> intentionally treats a function
+        /// literal as an opaque leaf (its body is a separate lexical scope),
+        /// so without this override the walk above never reaches a variable
+        /// read/written only inside a nested lambda body. Blindly descending
+        /// into <c>node.Body</c> with <em>this</em> walker would be wrong: it
+        /// would also invoke <see cref="RewriteVariableDeclaration"/> for
+        /// locals the lambda itself declares, incorrectly hoisting
+        /// lambda-owned state into the outer async method's state machine
+        /// and breaking the lambda's own lexical scope.
+        /// <para>
+        /// Instead, this reuses the capture metadata the binder already
+        /// computed: <see cref="BoundFunctionLiteralExpression.CapturedVariables"/>
+        /// lists every outer-scope variable the lambda (or any lambda
+        /// transitively nested within it) needs, because
+        /// <c>LambdaBinder.CapturedVariableCollector.RewriteFunctionLiteralExpression</c>
+        /// already propagates a nested literal's own captures up into every
+        /// enclosing literal's list (issue #503) — the exact mechanism
+        /// <see cref="Lowering.CaptureBoxingRewriter"/>'s own capture walker
+        /// relies on for the same reason. <see cref="LambdaCaptureCollector"/>
+        /// additionally recurses defensively (mirroring
+        /// <see cref="Lowering.CaptureBoxingRewriter"/>'s walker) using a
+        /// dedicated helper that itself never records a variable declaration,
+        /// so a lambda's own locals are still never hoisted even if a future
+        /// binder change stops fully flattening the capture list.
+        /// </para>
+        /// </remarks>
+        protected override BoundExpression RewriteFunctionLiteralExpression(BoundFunctionLiteralExpression node)
+        {
+            foreach (var captured in LambdaCaptureCollector.Collect(node))
+            {
+                Record(captured);
+            }
+
+            return node;
+        }
+
         private void Record(VariableSymbol variable)
         {
             if (variable is LocalVariableSymbol local && seen.Add(local))
             {
                 orderedLocals.Add(local);
             }
+        }
+    }
+
+    /// <summary>
+    /// Collects the set of outer-scope variables that a lambda literal — or
+    /// any lambda transitively nested inside its body — captures, using only
+    /// the binder-computed <see cref="BoundFunctionLiteralExpression.CapturedVariables"/>
+    /// metadata on each literal it visits (never a lambda's own local
+    /// declarations or parameter reads). Mirrors
+    /// <c>CaptureBoxingRewriter.CaptureWalker</c>'s defensive recursion (see
+    /// #567) for the same reason: the binder already flattens transitive
+    /// captures onto every enclosing literal (#503), so the recursion below
+    /// is normally a no-op, but keeps this collector correct if that
+    /// invariant ever changes.
+    /// </summary>
+    private sealed class LambdaCaptureCollector : BoundTreeRewriter
+    {
+        private readonly HashSet<VariableSymbol> sink = new HashSet<VariableSymbol>();
+
+        /// <summary>Collects the transitive capture set for <paramref name="root"/>.</summary>
+        /// <param name="root">The lambda literal to start from.</param>
+        /// <returns>Every outer-scope variable <paramref name="root"/> or any
+        /// lambda nested within it needs.</returns>
+        public static IReadOnlyCollection<VariableSymbol> Collect(BoundFunctionLiteralExpression root)
+        {
+            var collector = new LambdaCaptureCollector();
+            collector.VisitLiteral(root);
+            return collector.sink;
+        }
+
+        protected override BoundExpression RewriteFunctionLiteralExpression(BoundFunctionLiteralExpression node)
+        {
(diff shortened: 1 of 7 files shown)


DavidObando:
## Follow-up: hardened `AsyncCaptureWalker` for captures inside nested lambdas

This addresses the deferred gap noted in the original PR description: `AsyncCaptureWalker.ReferenceCollector` did not descend into nested lambda bodies when discovering ordinary locals/parameters that need hoisting into the async state machine.

### Root cause of the gap

`AsyncCaptureWalker.ReferenceCollector` is a `BoundTreeRewriter` subclass. The base class's `RewriteFunctionLiteralExpression` is a no-op leaf (it does not visit `node.Body`), so any rewriter that needs to see inside a lambda body must explicitly opt in. `ReferenceCollector` never did, so a local/parameter referenced *only* inside a nested lambda body was invisible to the walker's plain top-level traversal.

In practice this was not an *observable* runtime bug for ordinary (non-`this`) captures, because `CaptureBoxingRewriter` (which runs earlier in the pipeline) already boxes every boxable captured local/parameter and always places the box declaration at the *original outer-scope declaration site* — so the box variable was already visible to the walker's plain traversal as a side effect. `this` is explicitly excluded from boxing (it's never reassigned), which is why the `this`-specific crash fixed earlier in this PR was the one actually reachable bug.

### Fix

Closed the gap at the source instead of relying on the boxing side effect:

- Added `LambdaCaptureCollector`, a small `BoundTreeRewriter` that only overrides `RewriteFunctionLiteralExpression`. It records each lambda literal's own `CapturedVariables` — already transitively flattened by the binder across nested lambdas (issue #503), the same metadata `CaptureBoxingRewriter.CaptureWalker` relies on for issue #567 — and recurses into the lambda body purely to reach further-nested lambda literals. It never overrides variable-declaration/reference rewrites, so a lambda's own locally-declared locals can never be mistakenly hoisted into the outer async state machine.
- `ReferenceCollector` now overrides `RewriteFunctionLiteralExpression` to call `LambdaCaptureCollector.Collect(node)` and record the results, without itself recursing into the lambda body (which would incorrectly hoist lambda-owned locals).

No emitter slot-lookup changes, no global caches — this is purely a hoist-discovery fix in `AsyncCaptureWalker.cs`, reusing existing binder-provided capture metadata.

### Tests added

- `test/Core.Tests/CodeAnalysis/Lowering/Async/AsyncCaptureWalkerTests.cs` — 5 new white-box unit tests exercising `AsyncCaptureWalker.Analyze` directly: local captured only inside a nested lambda, parameter captured only inside a lambda, doubly nested lambda with transitive capture, mixed `this`+local capture, and a control proving a lambda's own local is *not* hoisted when not captured.
- `test/Compiler.Tests/Emit/Issue2331AsyncCaptureHoistEmitTests.cs` — 5 new compile+run integration tests forcing real `await`-suspension via `Task.Yield()`: local-before-await used only in a lambda after the await, parameter-only-in-lambda, doubly nested lambda, mixed `this`+local+parameter instance-method capture, and a lambda-own-local control run twice with different closure state to prove no cross-invocation aliasing.
- `test/Core.Tests/CodeAnalysis/Emit/Issue2331AsyncCaptureHoistFieldTests.cs` — reflection-based test asserting the synthesized state-machine type has a field for an outer local used only inside a nested lambda, and has **no** field for a local declared/used only inside the lambda's own body.

### Verification

- `dotnet build src/Compiler/Compiler.csproj -c Release` — 0 warnings/errors.
- Full `Core.Tests`: **5992/5993 passed** (1 pre-existing, unrelated `RegenerateBaseline` skip).
- Full `Compiler.Tests`: **2839/2839 passed**.

No deferred work remains for issue #2331.

Issue #1578, opened 2026-07-01
gsc: null-coalesce '??' over user value-type nullables (struct? / enum?) is unsupported (GS9998)


DavidObando:
## Summary

Null-coalesce `??` over a **user** value-type nullable (`struct?` / `enum?`) fails to compile with a `GS9998 NotSupportedException` from the emitter. This is the last remaining unsupported operation for user value-type nullables after #1572 (which fixed nil-default, lift, `(v!!)` unwrap, `(v!!).Member`, and narrowing). It surfaced as the documented out-of-scope note in PR #1576.

**Primitive** value-type nullable `??` (e.g. `int32?`) already works — only **user** structs/enums are affected.

## Repro (current `main`, HEAD `3882d0ad`)

```gsharp
package P
import System

struct Pt { var x int32 }
enum Color { Red, Green }

func CoalesceStruct(v Pt?) Pt {
    let fallback = Pt{x: -1}
    return v ?? fallback
}

func CoalesceEnum(v Color?) Color { return v ?? Color.Red }

func Main() { Console.WriteLine("ok") }
```

Both `return`s fail to compile:
```
error GS9998: NotSupportedException: Null-coalesce '??' over value-type operand 'Pt?' is not yet
supported by the emitter. The current `dup; brtrue` short-circuit is invalid IL for struct stack
values; a HasValue/Value (or box-before-brtrue) emit path is required when nullable value types
are supported.
```

For contrast, this compiles, ilverifies, and runs correctly today:
```gsharp
func CoalescePrim(v int32?) int32 { return v ?? -1 }   // OK
```

## Root cause

`src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Operators.cs:361-368` — the `??` emit has working value-type paths for primitive `Nullable<T>` and open struct-constrained `T?` (box-probe: `box; brfalse fallback`), but a final defensive branch throws `NotSupportedException` for "any other value-typed LHS (raw struct or enum)". Its comment states:

> // Today the encoder rejects nullable user-defined structs/enums, so this branch is unreachable from valid source

That comment is now **stale**: #1572 (PR #1576) taught the encoder to emit nullable user-defined `struct?`/`enum?`, so this branch is now reachable from valid source, and `??` is the one operation left unimplemented for them.

## Suggested fix

[...]

Issue #1572, opened 2026-07-01
gsc: user value-type nullables (struct? / enum?) emit invalid IL for nil-default, lift, and (v!!) unwrap


DavidObando:
## Summary

Same-compilation **user value-type nullables** (`struct?` and, partially, `enum?`) emit **invalid IL**. These are pre-existing gaps independent of smart-cast narrowing (they reproduce with **no** `if`-guard), surfaced while fixing #1547. The #1547 narrowing fix (PR #1569) reuses the `!!` (`NullAssertion`) unwrap machinery, so once the underlying value-type-nullable emit support below lands, narrowed reads of user `struct?`/`enum?` will start working automatically — but today the underlying operations themselves are broken.

## What breaks (verified via `ilverify` on current `main`, HEAD `bf06320c`)

| Operation | User `struct?` | User `enum?` |
| --- | --- | --- |
| `nil → T?` default | ❌ invalid IL (`found Nullobjref, expected Nullable<T>`) | ✅ verifies |
| `value → T?` lift | ❌ invalid IL (`found <T>, expected Nullable<T>`) | ✅ verifies (dedicated arm, #1298) |
| `(v!!)` unwrap | ❌ invalid IL (`Unexpected type on the stack` @ IL_0002) | ❌ invalid IL (same) |
| `(v!!).Member` / narrowed `v.Member` | ❌ invalid IL (inherits the broken `(v!!)`) | n/a |

Primitive value-type nullables (`int32?`, `bool?`, `char?`, `nint?`, …) are **not** affected — they emit and verify correctly (and are fully covered by #1547/PR #1569 for narrowing).

## Repro

```gsharp
package P
import System

struct Pt { var x int32 }
enum Color { Red, Green }

// 1. nil default — struct? INVALID, enum? OK
func StructNil() Pt? { return nil }

// 2. value -> Nullable lift — struct? INVALID, enum? OK
func StructLift() Pt? { let p = Pt{x: 1}; return p }

// 3. (v!!) unwrap — struct? AND enum? INVALID
func StructBang(v Pt?) Pt { return v!! }
func EnumBang(v Color?) Color { return v!! }

// 4. (v!!).Member — INVALID (inherits from 3)
func BangMember(v Pt?) int32 { return (v!!).x }

func Main() { Console.WriteLine("ok") }
```

Compile + verify:
```
dotnet out/bin/Release/Compiler/gsc.dll repro.gs /out:repro.dll /target:exe
[...]

Issue #1547, opened 2026-07-01
gsc: narrowing a nullable VALUE type via if-statement nil-guard emits invalid IL (missing Nullable<T> unwrap on load)


DavidObando:
## Summary

When gsc narrows a nullable **value** type (e.g. `int32?`) to its non-nullable form inside an `if (v != nil) { ... }` guard, later reads of the narrowed variable load the `System.Nullable<T>` storage slot directly and treat it as the narrowed value type **without emitting the `Nullable<T>` unwrap** (`.Value` / `Unbox.Any`). The resulting method fails ilverify with a stack-type mismatch (`found Int32, expected Nullable<int32>`).

This is a **latent, pre-existing** gap on `main` — no emit test currently exercises the value-type statement-narrowing shape, so it went unnoticed. Nullable **reference** types are unaffected (narrowing them is a CLR no-op).

## Repro

```gsharp
package P
import System

func StmtNarrow(v int32?) bool {
    if v != nil { return v > 0 }   // v narrowed to int32, but loaded as Nullable<int32>
    return false
}

func Main() { Console.WriteLine("ok") }
```

Compiles (`Wrote ...`), but ilverify reports:

```
[IL]: Error [StackUnexpected]: [P.<Program>::StmtNarrow(System.Nullable1<int32>)][offset 0x00000010][found Int32][expected value 'System.Nullable1<int32>'] Unexpected type on the stack.
```

Verified on `main` @ `07f3a2dd`.

## Working forms (for contrast)

- `if v != nil { return v!! > 0 }` — explicit non-null assertion unwrap gives valid IL.
- `if v != nil { return v > 0 }` where `v` is a nullable **reference** type — narrowing is a CLR no-op, valid IL.
- The nullable-**lifted** `>` operator on the un-narrowed `int32?` also works.

## Root cause

The if-statement nil-guard narrowing (`StatementBinder.TryClassifyNilGuard` to `SmartCastStability.TryClassifyNilGuardLeaf`) narrows the variable's static type to the non-nullable underlying, but the emitter does not unwrap the `Nullable<T>` backing slot when loading the narrowed variable. The static type says `int32` while the local slot is `Nullable<int32>`.

## Fix direction

[...]

Issue #831, opened 2026-06-14
Open-T `== nil` lowers to verifier-invalid `ldnull; ceq` against `T`


DavidObando:
**Repro** (`src/Sdk/Gsharp.Extensions/Optional/Optional.gs`-shaped):
```
func Map[T class, R class](self T?, mapper (T) -> R?) R? {
    if self == nil {
        return nil
    }
    return mapper(self!!)
}
```

**Symptom.** `ilverify` errors:
```
[StackUnexpected]: Gsharp.Extensions.Optional.<Program>::Map(!!0, Func`2<!!0,!!1>)
  [offset 0x00000016][found Nullobjref 'NullReference'][expected value 'T']
  Unexpected type on the stack.
```
Runtime tests pass (CLR JIT tolerates the shape) but the IL is not strictly verifiable. The compiler emits `ldarg.0; ldnull; ceq` for `self == nil` when `self` is `T` with `[T class]` constraint — `ldnull` puts a `NullReference` on the stack while `ceq` against an open-T value expects `T`.

**Expected.** Either:
- For `T : class`, emit `ldarg.0; ldnull; ceq` against a *reference* (no `value 'T'` on the stack), OR
- Lower `T == nil` to `EqualityComparer<T>.Default.Equals(self, default(T))` for open class-constrained T, OR
- Box `self` (`box !!T`) before comparison so the stack carries an `O` instead of a `value 'T'`.

**Impact.** Eight Optional helpers (`Map`, `FlatMap`, `OrElse`, `OrCompute`, `OrThrow`, `IfPresent`, `Filter`) in the new G# port emit ilverify-dirty IL. Runtime behaviour is correct (the class-constrained `T` is a reference at runtime); ilverify is conservative. Filed during #806 port.

**Workaround.** None on the source side. The struct-constrained overloads use `self.HasValue` / `self.Value` and verify cleanly (see #814 for that path).

Related: #806, ADR-0084 §L5.

PR #2334, opened 2026-07-14
Fix #2333: correct NullAssertion emission for narrowed nullable value-type reads

### LINKED ISSUE #2333: gsc: null assertion after nullable smart-cast emits invalid IL for floats and structs
## Summary

When a nullable value-type variable is narrowed by a nil guard and then explicitly asserted with `!!`, the narrowed variable load already emits `Nullable<T>.Value`, but `EmitUnary(NullAssertion)` applies a second generic reference-style `dup; brtrue` check to the bare value. This produces invalid IL for floats, structs, and generic value types.

Oahu.Foundation now compiles after #2332 and is blocked at IL verification by this defect.

## Minimal reproduction

```gsharp
package Repro
class C() {
    func F(x float64?, d float64) float64 {
        if x == nil {
            return d
        } else {
            return x!!
        }
    }
}
```

Compilation succeeds, but ILVerify reports `StackUnexpected`. Emitted IL first calls `Nullable<float64>.get_Value()`, then incorrectly emits `dup; brtrue` on the resulting floating-point stack value.

The same shape without `!!` verifies. Integer cases can appear to work because `brtrue` accepts integer stack types, masking the redundant check. Real failures also occur for `DateTime?`, generic type parameters, and other structs.

Real Oahu occurrence: `ArgParser.FindFloatArg` translates `double? arg` plus null guard and `arg.Value` to a narrowed `float64?` variable followed by `arg!!`. Similar fingerprints occur in EnumConverter/EnumChainConverter, TreeDecomposition, InteractionMessage, and LogTmpFileMaintenance.

## Root cause

Smart-cast narrowing strips the operand’s static `NullableTypeSymbol` wrapper in the non-null branch while preserving a nullable backing slot. The variable load correctly unwraps the backing slot. `EmitUnary` then sees the narrowed bare type, misses all nullable/type-parameter-specific null-assertion arms, and falls into the generic reference fallback.

This is the intersection not covered by #504, #831, #1547, or #1572: a nullable value narrowed by control flow and then redundantly asserted.

## Expected fix

- Detect a null assertion whose operand is already a smart-cast/narrowed read from a nullable backing symbol.
- Emit only the existing narrowed operand load/unwrap; do not add a second reference-style null check.
- Preserve direct `!!` behavior when the operand is still nullable and preserve runtime null checks for reference values.
- Generalize across BCL structs, user structs/enums, generic type parameters, floats, and integer controls.

## Tests

1. `float64?` narrowed in an `== nil` else branch then `!!`.
2. `DateTime?`.
3. Generic `T?`/enum-converter shape.
4. User struct/enum.
5. Integer/bool controls and direct non-narrowed `!!` behavior.
6. Exact Oahu ArgParser/LogTmpFileMaintenance shapes with ILVerify.

Discovered in Oahu migration round 8 on main after #2332.

DavidObando:
Closes #2333

## Root cause

`float64? x; if x == nil { ... } else { return x!! }` compiles successfully but fails ILVerify (`StackUnexpected`) at runtime.

When a nullable value-type variable/field/property is narrowed by a `== nil` control-flow guard (issue #1547), the binder's `BuildNarrowedRead` synthesizes an *inner* `BoundUnaryExpression(NullAssertion)` that unwraps `Nullable<T>` down to the bare underlying value type `T` (this is the actual `Nullable<T>.Value` call). A user-written `x!!` applied on top of this already-narrowed read composes as a **nested** `NullAssertion` whose operand type is the bare `T`, not `Nullable<T>`.

None of `EmitUnary`'s existing special-cased arms matched this bare-`T` shape:
- #1572 user value-type `Nullable<T>`
- #504 BCL/primitive value-type `Nullable<T>`
- #831 open non-struct-constrained type parameter

So it fell through to the generic reference-style `dup; brtrue; pop; throw` fallback — invalid IL for float/struct stack values (ILVerify rejects `dup`/`brtrue` on non-reference, non-primitive-integer stack values per ECMA-335 III.1.8), and semantically wrong even for shapes that happened to verify: int32 `0`, `bool false`, and enum ordinal `0` were misidentified as "null" and incorrectly threw `NullReferenceException`.

## Fix

Generalized rule added to `EmitUnary`'s `NullAssertion` handling: any operand whose static type is a bare (non-`Nullable`-wrapped) value type — primitive, BCL/user struct, enum, tuple, or struct-constrained generic type parameter — can never be a CLR `null` at the IL level (there is no bit pattern for "no value" outside the `Nullable<T>` wrapper itself). Therefore `!!` on such an operand needs no runtime check at all; the operand is emitted as-is.

This single rule:
- Fixes the narrowing-composition bug for every value-type shape (float32/64, `DateTime`, user structs/enums, struct-constrained generics).
- Fixes the latent falsy-value misfire bug (int `0`, `bool false`, enum ordinal `0` no longer misidentified as null).
- Leaves unchanged: direct nullable value-type `!!` on `nil` (#504, throws `InvalidOperationException`), reference-type `!!` on `nil` (throws `NullReferenceException`), and open type-parameter `!!` (#831).

### Additional gap found and fixed while auditing sibling paths

`Nullable<T>` where `T` is a struct-constrained generic type parameter (e.g. `EnumConverter<T>`-style code) has no `ClrType` during emit, so it was missed by both the #504 (ClrType-based) and #1572 (`IsUserValueTypeNullable`, enum/struct only) arms. Added a shared `RequiresSymbolicNullableGetValue` predicate (`NullableLifting.cs`) recognizing this shape, and propagated it to:
- `MethodBodyEmitter.Operators.cs` — `EmitUnary`'s `!!` now unwraps via the symbolic `get_Value` MemberRef path for this shape (in addition to the new no-check bare-value-type arm for direct reads of the unwrapped result).
- `ReflectionMetadataEmitter.cs` — relaxed `GetNullableGetValueMemberRefForUserValueType`'s guard.
- `SlotPlanner.cs` — `NullableValueTypeUnwrapCollector` now allocates the spill slot this shape requires.
- `MethodBodyEmitter.Expressions.cs` — `EmitNullConditionalReceiverProbe` (the `?.` null-conditional sibling path) previously crashed the compiler with `GS9998 InvalidOperationException` for this exact shape; dispatch guard updated to match.

### Audited but not changed

The `NullCoalesce` (`??`) sibling path (`NullableValueTypeCoalesceCollector` / `EmitBinary`) has the same struct-constrained-generic-`Nullable<T>` gap, but it already fails loudly with a documented `NotSupportedException` rather than emitting invalid IL or crashing. Left as an explicit, known, deferred limitation rather than fixed here, since it's a separate (already-safe) code path and out of scope for this issue's `!!`-focused repro.

## Tests

New file `test/Compiler.Tests/Emit/Issue2333NarrowedNullAssertionEmitTests.cs` (18 tests, compile + ILVerify + run where meaningful):
- Narrowed `!!` after `== nil` guard: `float64`, `float32`, `DateTime` (BCL struct), user struct, user enum (incl. ordinal-0 correctness), `int32` zero, `bool` false, struct-constrained generic type parameter.
- Direct (non-narrowed) `!!` on struct-constrained generic: success path and nil-throws path.
- Null-conditional `?.` on struct-constrained generic (crash fix).
- Preserved behavior: direct nullable value-type `!!` on nil (#504) and has-value; direct reference-type `!!` on nil and has-value.
- Oahu-shaped repros: `ArgParser.FindFloatArg`-style narrowed float parse, `LogTmpFileMaintenance`-style narrowed `DateTime` arithmetic.
- Nested multi-read arithmetic case.

Validation performed:
- `dotnet build src/Compiler/Compiler.csproj -c Release` — clean, 0 warnings/errors.
- Targeted suite (63 tests: `NullableValueEmitTests`, `Issue1547...`, `Issue831...`, `Issue1572...`, `NullableInstanceMember...`, `NullableProperty...`, `Issue2072...`, `Issue2051...`) — all pass, confirming #504/#831/#1547/#1572 behavior preserved.
- New `Issue2333NarrowedNullAssertionEmitTests` (18 tests) — all pass with the fix; sanity-checked via `git stash` that 14/18 fail without it (the other 4 exercise already-correct pre-existing behavior).
- Full `test/Core.Tests` — 5992 passed, 1 skipped (unrelated `RegenerateBaseline`).
- Full `test/Compiler.Tests` — 2857 passed (includes ILVerify checks across the whole suite).

## Update (follow-up commit)

The struct-constrained-generic `Nullable<T>` `??` limitation noted below has since been **implemented** (see the follow-up commit and PR comment below) by widening the existing #1578 user-value-type-nullable `??` arm's guard to `RequiresSymbolicNullableGetValue`, reusing its box-probe / symbolic `get_Value` machinery rather than adding a new special case. `??=` was also audited and confirmed to already work correctly for this shape (different code path — lowers to `if (read == nil) { write = rhs }`). 13 new tests added (compile + ILVerify + run), including evaluation-once and negative-control coverage. Full `Core.Tests` (5992) and `Compiler.Tests` (2870) pass.

## Deferred / out of scope
- Oahu was not modified per task instructions.
- (Historical note, now resolved — see "Update" above) `??` struct-constrained-generic-`Nullable<T>` limitation.

### DIFF
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs
@@ -1507,7 +1507,7 @@ internal sealed partial class MethodBodyEmitter
             this.il.Branch(ILOpCode.Brfalse, nilFallthrough);
 
             this.il.LoadLocalAddress(wrapperSlot);
-            if (NullableLifting.IsUserValueTypeNullable(receiverNullable))
+            if (NullableLifting.RequiresSymbolicNullableGetValue(receiverNullable))
             {
                 this.il.OpCode(ILOpCode.Call);
                 this.il.Token(this.outer.GetNullableGetValueMemberRefForUserValueType(receiverNullable));
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Operators.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Operators.cs
@@ -465,22 +465,31 @@ internal sealed partial class MethodBodyEmitter
                 return;
             }
 
-            // Issue #1578: NullCoalesce over a user value-type nullable LHS —
-            // `Nullable<UserT>` where UserT is a user-declared value-kind
-            // struct or enum emitted in this compilation (#1572). The
-            // underlying has no runtime CLR type, so neither the primitive
-            // `Nullable<T>` spill arm at the top (which probes via the
-            // BCL-backed `get_HasValue`) nor the bottom `dup; brtrue` shape is
-            // usable. Mirror the sibling value-type spill (#831 / #1516) with a
-            // box-probe over the emitted `Nullable<UserT>` TypeSpec: spill the
-            // LHS to the pre-allocated `Nullable<UserT>` slot, box it (boxing a
-            // `Nullable<T>` yields `null` when `!HasValue`, else a boxed `T` —
-            // ECMA-335), and `brfalse` to the fallback on the empty case. On
-            // the non-null branch, reproduce the operator's result type: reload
-            // the wrapper when the result is itself `Nullable<UserT>`, else
-            // unwrap to `UserT` via the TypeSpec `get_Value` MemberRef (#1572).
+            // Issue #1578 / #2333: NullCoalesce over a "symbolic" value-type
+            // nullable LHS — `Nullable<UserT>` where UserT is a user-declared
+            // value-kind struct or enum emitted in this compilation (#1572),
+            // OR `Nullable<T>` where T is an open type parameter constrained
+            // to `struct` (its instantiation closes over a generic-parameter
+            // signature slot, not a resolvable host `Type` — issue #2333's
+            // audit of this sibling path). Both underlyings have no runtime
+            // CLR type, so neither the primitive `Nullable<T>` spill arm at
+            // the top (which probes via the BCL-backed `get_HasValue`) nor
+            // the bottom `dup; brtrue` shape is usable. Mirror the sibling
+            // value-type spill (#831 / #1516) with a box-probe over the
+            // emitted/encoded `Nullable<T>` TypeSpec: spill the LHS to the
+            // pre-allocated `Nullable<T>` slot, box it (boxing a
+            // `Nullable<T>` yields `null` when `!HasValue`, else a boxed `T`
+            // — ECMA-335 III.4.1), and `brfalse` to the fallback on the empty
+            // case. On the non-null branch, reproduce the operator's result
+            // type: reload the wrapper when the result is itself
+            // `Nullable<T>`, else unwrap to `T` via the symbolic `get_Value`
+            // MemberRef (#1572 / #2333), reusing the same
+            // `RequiresSymbolicNullableGetValue` predicate and
+            // `GetNullableGetValueMemberRefForUserValueType` helper that the
+            // `!!` and `?.` emit paths use, rather than a separate special
+            // case for the type-parameter shape.
             if (b.Left.Type is NullableTypeSymbol userVtNullable
-                && NullableLifting.IsUserValueTypeNullable(userVtNullable))
+                && NullableLifting.RequiresSymbolicNullableGetValue(userVtNullable))
             {
                 if (!this.nullableCoalesceSpillSlots.TryGetValue(b, out var userVtSlot))
                 {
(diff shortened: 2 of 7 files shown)


DavidObando:
## Update: struct-constrained generic `Nullable<T>` `??` now implemented

The previously-deferred gap is now closed on this branch (second commit).

**What changed:** `EmitBinary`'s `NullCoalesce` (`??`) arm for user value-type `Nullable<T>` (issue #1578, already box-probe/symbolic-`get_Value`-based) had its guard widened from `NullableLifting.IsUserValueTypeNullable` to `NullableLifting.RequiresSymbolicNullableGetValue` — the same predicate the `!!`/`?.` fix introduced. No new special case was added: the struct-constrained-generic shape reuses the exact same box-probe (`box Nullable<T>; brfalse`) and symbolic `get_Value` MemberRef machinery already proven for user structs/enums. `SlotPlanner`'s matching collector guard was updated in lockstep so the pre-allocated spill slot agrees with the emit arm.

**Preserved:** primitive/BCL `Nullable<T>` `??` (#519), reference-type `??` (`dup;brtrue`), class-constrained/unconstrained type-parameter `??` (#831/#1516), and evaluation-once/short-circuit semantics (verified via an explicit side-effect test — the LHS is evaluated exactly once regardless of branch).

**`??=` audited:** it lowers in the binder to a plain `if (read == nil) { write = rhs; }` using the standard `==` nil-comparison operator — a different code path from the `??` binary-operator emit arm — and was already correct for this shape (confirmed by repro + a lock-in test), so no source change was needed there.

**Tests added:** `test/Compiler.Tests/Emit/Issue2333StructConstrainedGenericCoalesceEmitTests.cs` (13 tests, compile + ILVerify + run): nil/present operands across `int32`/`float64`/`DateTime`(BCL struct)/user-struct instantiations, falsy-but-present correctness (`int32 0`, `bool false`), a `Nullable`-result `??` chain (reload-vs-unwrap branch selection), evaluation-once coverage via an observable side effect, a `??=` lock-in test, and negative controls proving the class-constrained (#831), unconstrained-type-parameter-as-cast (#1516), primitive (#519), and reference-type `??` paths are unaffected.

**Validation:** full `test/Core.Tests` (5992 passed) and full `test/Compiler.Tests` (2870 passed, including the 13 new tests) are green. Sanity-checked via `git stash` that 8 of the 13 new tests fail without this change (the struct-constrained-generic-specific ones) while the 5 negative-control tests pass either way, confirming the tests genuinely exercise the fix.

No deferred work remains from this PR's scope.

Issue #1611, opened 2026-07-02
Field-like event accessors on generic types use bare FieldDef tokens → runtime crash (drifted from the #989 property fix)


DavidObando:
- **Category**: correctness
- **Severity**: P1
- **Location**: src/Core/CodeAnalysis/Emit/MemberDefEmitter.cs:1051-1053 and 1073-1076 (instance add; same pattern in instance remove ~1168-1193 and static add/remove ~648-672, 748-772)
- **Evidence**:
```csharp
il.LoadArgument(0);
il.OpCode(ILOpCode.Ldfld);
il.Token(backingHandle);        // bare FieldDef even on a generic type
...
il.OpCode(ILOpCode.Ldflda);
il.Token(backingHandle);
```
The parallel property path was fixed (MemberDefEmitter.cs:216-218): `var backingToken = ReflectionMetadataEmitter.IsUserGenericTypeReference(structSym) ? this.resolveFieldToken(structSym, prop.BackingField) : (EntityHandle)backingHandle;` — the four event CAS-loop copies never received the fix although `resolveFieldToken` is already injected into the class.
- **Problem**: For `class Holder[T any]` with a field-like event, subscribing crashes — reproduced: `System.InvalidOperationException: ... method itself or the containing type is not fully instantiated. at Holder`1.add_Changed`; ilverify flags `StackUnexpected` in both accessors. ECMA-335 requires member references on generic types to go through the self-instantiation TypeSpec; a clean-compiling program hard-crashes at runtime.
- **Suggested fix**: Resolve the backing-field token via `IsUserGenericTypeReference` → `resolveFieldToken` in all four event accessor bodies (ideally by deduplicating them into one parameterized helper).
- **Confidence**: high

---
*Found by an automated multi-agent code review with Claude Code. `file:line` references are relative to commit `1e7e5a91`.*

Issue #989, opened 2026-06-23
Generic auto-property 'prop Value T' cannot be member-accessed (GS0158)


DavidObando:
Discovered by the **cs2gs** C#→G# migration pipeline (#914), L5 corpus. Verified with gsc `0.2.124+6cef33811b`.

Declaring a generic auto-property `prop Value T` on a generic class and then reading the member → `GS0158` "Cannot find member Value". A generic **field** and a non-generic auto-property both resolve, so the bug is specific to generic auto-properties.

### Repro (GS0158)
```gs
package T
class Box[T] {
    prop Value T
}
func test(b Box[int32]) int32 {
    return b.Value
}
```

### Control (a generic field resolves)
```gs
package T
class Box[T class](Value T) { }
```

**Classification:** binder bug — a generic auto-property member should resolve like a generic field. Ref ADR-0115 §G #TBD-5.

Issue #2230, opened 2026-07-07
gsc: generic interface-method implementations not matched for IMPORTED (metadata) interfaces (GS0187, ILogger.BeginScope<T>/Log<T>)


DavidObando:
> **Root-caused after review (2026-07-07).** gsc **does** support generic interface-method implementation — but only when the interface is **source-declared**. The bug is specifically that matching **fails for imported (metadata) interfaces**. Confirmed with a minimal repro (below). This is a gsc gap, not a cs2gs gap — cs2gs emits the correct implicit implementation.

## Symptom (Oahu.Cli.Server)

`StderrLogger : ILogger` (an **imported** interface from `Microsoft.Extensions.Logging.Abstractions`) implements the interface's generic methods implicitly:
```csharp
public IDisposable? BeginScope<TState>(TState state) where TState : notnull => null;
public bool IsEnabled(LogLevel logLevel) => logLevel >= LogLevel.Warning;
public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter) { ... }
```
cs2gs emits correct implicit implementations:
```
func BeginScope[TState](state TState) IDisposable? -> nil
func IsEnabled(logLevel LogLevel) bool -> logLevel >= LogLevel.Warning
func Log[TState](logLevel LogLevel, eventId EventId, state TState, exception Exception?, formatter (TState, Exception?) -> string) { ... }
```
gsc accepts the non-generic `IsEnabled`, but rejects both **generic** methods:
```
ServerHost.gs(75,23): error GS0187: Class 'StderrLogger' does not implement interface method 'ILogger.BeginScope(TState)'.
ServerHost.gs(75,23): error GS0187: Class 'StderrLogger' does not implement interface method 'ILogger.Log(LogLevel, EventId, TState, Exception, Func`3)'.
```

## Minimal repro — isolates source vs imported

**Source interface (same compilation) → compiles clean:**
```
// one file
interface IStore { func Put[T](item T) bool; func Name() string; }
class MyStore : IStore { func Put[T](item T) bool -> true    func Name() string -> "my" }
```
`gsc src.gs` → **Success**.

**Imported interface (separate assembly) → GS0187 on the generic method:**
```
// lib.gs  ->  gsc lib.gs /target:library /out:lib.dll
package Lib
interface IStore { func Put[T](item T) bool; func Name() string; }
```
```
// use.gs  ->  gsc use.gs /r:lib.dll
[...]

PR #2340, opened 2026-07-14
Fix #2338: self-instantiation-aware field tokens for generic primary-ctor base-initializer stores

### LINKED ISSUE #2338: gsc: generic primary constructor with base initializer stores fields using open FieldDef token
## Summary

A generic primary-constructor class/data class with an explicit base initializer emits its own primary-constructor field stores using a bare `FieldDef` token rather than the self-instantiated field token. ILVerify reports a generic instantiated receiver where an open generic receiver is expected.

## Minimal reproduction

```gsharp
package Repro
open data class Base(Name string) { }
open data class Derived[T](Name string, Value T) : Base(Name) { }
```

Compilation succeeds; ILVerify reports `found ref Derived<T0>, expected ref Derived` in the constructor field store.

Real Oahu occurrence: `InteractionMessage<T>(...) : InteractionMessage(Type, Message)`.

## Root cause

`ReflectionMetadataEmitter.EmitClassConstructorWithBaseInitializerBodyBytes` obtains primary-constructor field handles directly from `cache.StructFieldDefs`, emitting the open FieldDef. Its sibling constructor path without a base initializer correctly uses `ResolveFieldToken(classSym, field)`, which emits a self-instantiation-aware MemberRef/field token for generic classes.

This is the same anti-pattern previously fixed for properties/events (#989/#1611), at a distinct primary-constructor-with-base-initializer call site.

## Expected fix

- Use `ResolveFieldToken(classSym, field)` for these constructor field stores.
- Audit adjacent bare `StructFieldDefs` accesses in constructor/accessor emit paths for generic self-instantiation drift.
- Preserve non-generic and no-base-initializer constructor behavior.

## Tests

- Generic data/primary-constructor class with non-generic base initializer, compile + ILVerify + run.
- Multiple fields including generic field types.
- Non-generic and no-base-initializer controls.
- Exact Oahu `InteractionMessage<T>` shape.

Discovered and scratch-verified in Oahu migration round 10.

DavidObando:
## Summary

Fixes #2338. A generic primary-constructor class/data class with an explicit base initializer (`class Derived[T](...) : Base(args)`) emitted its own primary-constructor field stores through a bare, open `FieldDef` token instead of the self-instantiation-aware `ResolveFieldToken(classSym, field)`. ilverify reported:

```
found ref Derived<T0>, expected ref Derived
```

exactly matching the issue and the real Oahu `InteractionMessage<T>(...) : InteractionMessage(Type, Message)` occurrence.

## Root cause & fix

`EmitClassConstructorWithBaseInitializerBodyBytes` in `ReflectionMetadataEmitter.cs` looked up each primary-ctor field's token directly via `cache.StructFieldDefs[field]`. Its sibling scaffold (`EmitClassConstructorWithBodyBodyBytes`, used when there's no explicit base initializer) already called `ResolveFieldToken(classSym, field)`, which emits the correct self-TypeSpec-parented `MemberRef` for a generic type — the exact anti-pattern already fixed for auto-properties (#989) and field-like events (#1611), here recurring at a third, distinct call site.

Fix: route the base-initializer constructor's field stores through `ResolveFieldToken` too.

## Audit of adjacent bare `StructFieldDefs` lookups

Per the issue's ask, audited every other bare `cache.StructFieldDefs` access in constructor/accessor emit paths:

- `MemberDefEmitter` property/event accessor bodies — already generic-aware (`IsUserGenericTypeReference` + `resolveFieldToken`), per #989/#1611.
- `MethodBodyEmitter` field-access/assignment/address emission, struct-literal initializers — already routed through `ResolveFieldToken`/`ResolveFieldReferenceContainer` with an `isGeneric` gate.
- Remaining bare accesses are either existence checks (`ContainsKey`, no token emitted), planning-phase bookkeeping, or explicitly guarded off for generic instantiations already (e.g. `EmitDefaultStructStringFieldPatches`).

No further generic-token-drift call sites were found in this pass.

**Follow-up completed (see PR comment below):** `MethodBodyEmitter.Closures.EmitFunctionLiteralToNamedDelegate` (a lambda-to-named-delegate conversion) was initially found not to apply the same generic-aware ctor/field/ldftn token resolution its sibling `EmitFunctionLiteral` applies for generic-reified display classes (issue #1477). This has since been fixed on this same branch/PR, with dedicated tests and a full audit of adjacent named-delegate emission paths — no deferred work remains.

## Second, independent defect uncovered while testing

Reproducing the issue's literal minimal repro end-to-end (compile + ilverify + **run**) surfaced a second, unrelated, non-generic bug in the same neighborhood: `DataStructSynthesizer` unconditionally marked the synthesized `Equals(object)`/`GetHashCode()`/`ToString()` overrides `virtual final`. Any `open data class` derived from another `open data class` (exactly the issue's own repro shape, and the Oahu `InteractionMessage` shape) threw:

```
TypeLoadException: Declaration referenced in a method implementation cannot be a final method.
```

at type-load time — regardless of generics (reproduces identically with two plain non-generic data classes). Fixed by only marking these overrides `final` when the type can't be further subclassed, mirroring the existing `!structSym.IsOpen && !structSym.IsSealedHierarchy` condition `TypeDefEmitter` already uses to decide `TypeAttributes.Sealed`.

This was necessary to make the issue's own literal repro (and the multi-field/InteractionMessage/data-extends-data test scenarios) actually **run**, not just compile and verify.

## Tests

Added `test/Compiler.Tests/Emit/Issue2338GenericPrimaryCtorBaseInitializerEmitTests.cs` — all compile + ilverify + run end to end:

- `MinimalRepro_GenericDataClassWithNonGenericBaseInitializer_CompilesVerifiesAndRuns` — the issue's exact repro.
- `MultipleAndGenericFields_ConstructedOverReferenceTypeArgument_CompilesVerifiesAndRuns` — multiple fields, two type parameters, constructed over a reference type argument.
- `OahuInteractionMessageShape_GenericOverPayload_CompilesVerifiesAndRuns` — the exact Oahu `InteractionMessage[T]` shape, including `Equals`/`GetHashCode`/`ToString`/`==` round-trips.
- `NonGenericControl_DataClassWithBaseInitializer_CompilesVerifiesAndRuns` — non-generic control.
- `NoBaseInitializerControl_GenericDataClass_CompilesVerifiesAndRuns` — generic primary ctor with no base initializer (already-correct sibling scaffold) control.
- `NonDataClassControl_GenericClassWithBaseInitializer_CompilesVerifiesAndRuns` — plain (non-data) generic class control, isolating the field-token fix from `DataStructSynthesizer`.
- `DataClassExtendsDataClassControl_NonGeneric_TypeLoadsAndRunsWithoutFinalMethodError` — control for the second, independent defect.

Confirmed the fix precisely: reverted the field-token change only and re-ran the minimal repro through `ilverify`, reproducing the exact reported error (`found ref '[Repro]Repro.Derived\`1<T0>'][expected ref '[Repro]Repro.Derived\`1'`) before restoring the fix.

## Validation

- Targeted: `Issue2338*`, `Issue989*`, `Issue1611*`, `Issue1617*`, `DataStructInteropTests`, `DataStructSynthesizedMembersTests`, `Issue2228DataClassWithEmitTests` — all pass (63 + 19 = 82 tests across the two filtered runs).
- Full `test/Core.Tests` suite: **5992 passed**, 1 skipped (pre-existing `RegenerateBaseline`), 0 failed.
- Full `test/Compiler.Tests` suite: **2899 passed**, 0 failed.

## Not modified

Oahu was not touched, per instructions.

Closes #2338.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/Binder.cs
+++ b/src/Core/CodeAnalysis/Binding/Binder.cs
@@ -4659,6 +4659,38 @@ public sealed class Binder
                 : type;
         }
 
+        // Issue #2340 follow-up (sibling to the #1503 branch already present
+        // in StructSymbol.SubstituteTypeForConstruction): a constructed
+        // generic named delegate appearing as a call's parameter or return
+        // type (e.g. `func MakeGetter[T](item T) Getter[T]`) must have its own
+        // type arguments substituted through the call's method-type-argument
+        // map so the bound call surfaces `Getter[int32]` rather than the
+        // still-open `Getter[T]`. Without this branch the binder's computed
+        // call-expression type stayed open over the callee's own type
+        // parameter even though the emitter correctly built a MethodSpec/
+        // MemberRef closed over the concrete argument — the mismatch between
+        // the (wrong, open) declared type of the receiving local/field and
+        // the (correct, closed) value actually produced by the `call`
+        // instruction failed ilverify with `StackUnexpected`.
+        if (type is DelegateTypeSymbol del
+            && del.Definition != null
+            && !ReferenceEquals(del.Definition, del)
+            && !del.TypeArguments.IsDefaultOrEmpty)
+        {
+            var newDelegateArgs = ImmutableArray.CreateBuilder<TypeSymbol>(del.TypeArguments.Length);
+            var delegateChanged = false;
+            foreach (var arg in del.TypeArguments)
+            {
+                var substituted = SubstituteType(arg, substitution, mapClrType);
+                delegateChanged |= !ReferenceEquals(substituted, arg);
+                newDelegateArgs.Add(substituted);
+            }
+
+            return delegateChanged
+                ? DelegateTypeSymbol.Construct(del.Definition, newDelegateArgs.MoveToImmutable())
+                : type;
+        }
+
         if (type is ImportedTypeSymbol it && it.HasTypeParameterArgument)
         {
             // #313: substitute a generic type parameterized by an in-scope type
--- a/src/Core/CodeAnalysis/Binding/DeclarationBinder.cs
+++ b/src/Core/CodeAnalysis/Binding/DeclarationBinder.cs
@@ -7447,6 +7447,39 @@ internal sealed class DeclarationBinder
                 && TypeArgumentsEquivalent(ia.TypeArguments, ib.TypeArguments, typeParamMap);
         }
 
+        // Issue #2340 follow-up: a named delegate constructed over a
+        // *method-level* generic parameter (e.g. an interface method
+        // `func Make[T](seed T) Getter[T]`) is never covered by the
+        // StructSymbol/InterfaceSymbol branches above, and — unlike the
+        // interface-level generic case — is not fixed by substituting the
+        // interface's own type arguments, because the delegate's type
+        // argument here is the *method's* type parameter, which is only
+        // resolved through `typeParamMap` (built per call by
+        // TryBuildMethodTypeParameterMap). Without this branch,
+        // `Getter[T_interfaceMethod]` and `Getter[T_implMethod]` are two
+        // distinct DelegateTypeSymbol.Construct cache entries (different
+        // TypeParameterSymbol instances as their sole type argument) that
+        // fail the top-level ReferenceEquals check and then fall through to
+        // the ClrType fallback below, which does not apply the
+        // method-type-parameter map and wrongly rejects a genuinely matching
+        // signature with GS0187. Recurses through TypeArgumentsEquivalent
+        // exactly like the StructSymbol/InterfaceSymbol branches so the
+        // typeParamMap substitution applies to each type argument.
+        if (a is DelegateTypeSymbol dta && b is DelegateTypeSymbol dtb)
+        {
+            // Unlike StructSymbol/InterfaceSymbol (whose `Definition` self-
+            // references on the open definition), DelegateTypeSymbol.Definition
+            // is only set on constructed instances (issue #1503) and stays
+            // `null` on the open definition / non-generic delegates — so it
+            // must be normalized to itself before comparing, or two distinct
+            // non-generic (or open-definition) delegate types would both
+            // report `Definition == null` and wrongly compare as equivalent.
+            var defA = dta.Definition ?? dta;
+            var defB = dtb.Definition ?? dtb;
+            return ReferenceEquals(defA, defB)
+                && TypeArgumentsEquivalent(dta.TypeArguments, dtb.TypeArguments, typeParamMap);
+        }
+
         if (a is ImportedTypeSymbol pa && b is ImportedTypeSymbol pb)
         {
             // Constructed imported generics carrying symbolic arguments (e.g.
--- a/src/Core/CodeAnalysis/Emit/DataStructSynthesizer.cs
+++ b/src/Core/CodeAnalysis/Emit/DataStructSynthesizer.cs
@@ -518,11 +518,45 @@ internal sealed class DataStructSynthesizer
     }
 
     /// <summary>
+    /// Issue #2338 follow-up: returns the <see cref="MethodAttributes"/> for a
+    /// synthesized <c>Object</c>-virtual override (<c>Equals(object)</c>,
+    /// <c>GetHashCode()</c>, <c>ToString()</c>) on a data type. A data STRUCT's
+    /// override is always <c>final</c> (structs can't be subclassed anyway,
+    /// and the parser rejects <c>open data struct</c> outright — ADR-0078 /
+    /// GS0309). A data CLASS's override is only <c>final</c> when the class
+    /// itself cannot be further subclassed — mirroring the exact
+    /// <c>!structSym.IsOpen &amp;&amp; !structSym.IsSealedHierarchy</c> condition
+    /// <see cref="TypeDefEmitter"/> uses to decide the TypeDef's own
+    /// <c>TypeAttributes.Sealed</c> flag. An <c>open</c> (or ADR-0078
+    /// <c>sealed class</c> discriminated-union) data class must leave these
+    /// overrides non-final so a further-derived data class can re-override
+    /// them with its own combined field set; marking them unconditionally
+    /// final made loading ANY data-class-extends-data-class hierarchy throw
+    /// <c>TypeLoadException: Declaration referenced in a method
+    /// implementation cannot be a final method</c> the moment the subclass
+    /// tried to override the base's already-final <c>Equals</c>/
+    /// <c>GetHashCode</c>/<c>ToString</c>.
+    /// </summary>
+    private static MethodAttributes DataObjectOverrideAttributes(StructSymbol structSym)
+    {
+        var attrs = MethodAttributes.Public | MethodAttributes.Virtual | MethodAttributes.HideBySig;
+        if (!structSym.IsClass || (!structSym.IsOpen && !structSym.IsSealedHierarchy))
+        {
+            attrs |= MethodAttributes.Final;
+        }
+
+        return attrs;
+    }
+
+    /// <summary>
     /// Issue #410 / ADR-0029: emits
     /// <c>public sealed override bool Equals(object other)</c> that performs
     /// <c>other is Name p &amp;&amp; this.Equals(p)</c>. Sealed because struct
     /// methods cannot be overridden in user code anyway, but the metadata
-    /// flag communicates intent.
+    /// flag communicates intent. Issue #2338 follow-up: for an <c>open</c>
+    /// (or ADR-0078 sealed-hierarchy) data CLASS the override is instead left
+    /// non-final via <see cref="DataObjectOverrideAttributes"/> so a further
+    /// derived data class can re-override it.
     /// </summary>
     private void EmitDataStructEqualsObject(StructSymbol structSym, EntityHandle typeDef, MethodDefinitionHandle equalsTypedHandle)
     {
@@ -556,7 +590,7 @@ internal sealed class DataStructSynthesizer
             .Parameters(1, r => r.Type().Boolean(), ps => ps.AddParameter().Type().Object());
 
         this.emitCtx.Metadata.AddMethodDefinition(
-            attributes: MethodAttributes.Public | MethodAttributes.Virtual | MethodAttributes.Final | MethodAttributes.HideBySig,
+            attributes: DataObjectOverrideAttributes(structSym),
             implAttributes: MethodImplAttributes.IL | MethodImplAttributes.Managed,
             name: this.emitCtx.Metadata.GetOrAddString("Equals"),
             signature: this.emitCtx.Metadata.GetOrAddBlob(sig),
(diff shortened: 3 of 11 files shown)


DavidObando:
## Follow-up: `EmitFunctionLiteralToNamedDelegate` generic-reified closure fix (deferred item now closed)

This PR previously deferred the `EmitFunctionLiteralToNamedDelegate` generic-closure drift as out of scope. It has now been fixed on this same branch/PR (commit `4b16a635`).

### Comparison with sibling / #1477 pattern

`EmitFunctionLiteralToNamedDelegate` (a lambda literal converted to a user-declared **named** delegate type, e.g. `type Getter[T any] = delegate func() T`) emitted the closure's `newobj` ctor, capture-field `stfld`s, and `ldftn Invoke` against bare cached handles unconditionally. Its correct sibling, `EmitFunctionLiteral` (used for the inferred/`Func`-typed delegate path), already checked `IsUserGenericTypeReference(closure.ConstructedClassSym)` and routed those same three token sites through `ResolveUserCtorTokenForDefault` / `ResolveFieldToken` / `ResolveClosureInvokeFtnToken` when the closure is generic-reified (issue #1477). The fix makes `EmitFunctionLiteralToNamedDelegate` mirror that exactly.

Note `ResolveDelegateCtorToken` (issue #1503), used for the **target named delegate type's** own `.ctor` handle, was already generic-aware — that's a separate concern (the delegate type itself being generic) from the one fixed here (the **closure** being generic).

### Audit of adjacent named-delegate emission paths

No further drift found:
- `EmitMethodGroupToNamedDelegate` / `EmitMethodGroupTarget` — already correctly resolve a generic receiver via `ResolveUserInstanceMethodToken` (issue #1467).
- `EmitFunctionToNamedDelegateConversion` / `EmitNullGuardedDelegateToDelegateAdaptation` — only perform CLR-reflectable delegate-to-delegate adaptation and already resolve the target delegate `.ctor` via `ResolveDelegateCtorToken` (issue #1503).

### Tests

Added `test/Compiler.Tests/Emit/Issue2338NamedDelegateGenericClosureEmitTests.cs`, 4 compile+ilverify+run facets:
- **Facet A** — generic containing class captures `this.value` (a `T`-typed field) into a named delegate; value-type and reference-type instantiations.
- **Facet B** — generic function (MVAR) captures its type-parameter-typed parameter into a named delegate; value-type and reference-type instantiations.
- **Facet C** — nested generic context: a generic method with its own additional type parameter inside a generic class, two lambdas separately capturing the class-level and method-level type-parameter values into differently-instantiated named delegates.
- **Facet D** — non-generic control, confirming no regression.

Each facet was confirmed to fail with the expected `StackUnexpected` (open closure type where the constructed type is expected) + `DelegateCtor` ("Unrecognized arguments for delegate .ctor") pattern when the fix is reverted, then verified clean after restoring it.

### Discovered but out of scope

While designing these tests, isolated repros confirmed a **separate, pre-existing bug unrelated to closures**: calling a generic function/method whose *parameter or return type* is a named delegate type constructed over the function's **own** type parameter produces a `StackUnexpected` mismatch at the **caller's** call site — independent of any capture/closure (reproduces even with a zero-capture pass-through identity function). This is unrelated to #2338/#1477 and is **not fixed here**; flagging for a future issue.

### Validation

- `dotnet build src/Compiler/Compiler.csproj -c Release` — 0 warnings/errors.
- New targeted tests: 4/4 passed.
- Full `test/Core.Tests` suite: **5992 passed**, 1 skipped (pre-existing), 0 failed — unchanged from prior baseline.
- Full `test/Compiler.Tests` suite: **2903 passed** (2899 baseline + 4 new), 0 failed.

No deferred work remains for this PR.

DavidObando:
## Update: generic named-delegate call-site defect — RESOLVED (no longer deferred)

Synced this branch with `origin/main` (clean merge, PR #2339/#2337 fix now included) and then fixed the generic named-delegate call-site defect that was previously flagged as deferred in the prior update.

### Root cause
`Binder.SubstituteType` — the single shared substitution primitive threaded through `OverloadResolver` for every source function/method/extension call (free function calls, extension method calls, instance and shared/static method calls) — had no case for `DelegateTypeSymbol`, unlike its sibling `StructSymbol.SubstituteTypeForConstruction` (which already handles this per #1503).

Calling a generic function/method whose parameter or return type is a named delegate constructed over the callee's own type parameter (e.g. `func MakeGetter[T](item T) Getter[T]`) produced a `BoundCallExpression` whose `.Type` stayed open over `T` instead of being substituted to the closed instantiation (`Getter[int32]`). The emitter's `MethodSpec`/`MemberRef` construction reads `MethodTypeArguments` directly from the substitution dictionary (bypassing `SubstituteType` entirely), so the actual `call`/`callvirt` instruction was always correctly closed — but the caller's surrounding code (local variable slot, subsequent member/invoke lookups, further conversions) used the binder's still-open type, producing a stack-slot mismatch. This surfaced as ILVerify `StackUnexpected` when the call's return value was used, or in some binding shapes a spurious `GS0155` compile-time rejection. Pure parameter-only passing (delegate value never returned/used) does not independently manifest the bug.

### Fix
Added a `DelegateTypeSymbol` branch to `Binder.SubstituteType` (`src/Core/CodeAnalysis/Binding/Binder.cs`), mirroring `StructSymbol.SubstituteTypeForConstruction`'s existing #1503 branch exactly. Because this is the single shared substitution primitive used by every call path through `OverloadResolver`, the fix generalizes automatically across source and imported call sites, instance and shared/static methods, free functions, and extension methods — no call path is special-cased.

### Audit
- Confirmed `BuildMethodSpecForGenericCall`/`BuildMethodSpecForGenericInstanceCall` in `ReflectionMetadataEmitter.cs` were already correct (they read `MethodTypeArguments` straight from the substitution map) — no emitter changes needed.
- Found a third, independent copy of this substitution logic in `InterfaceSymbol.SubstituteType` with the same `DelegateTypeSymbol` (and `StructSymbol`) gap. Confirmed via repro that it is currently unreachable in practice: any class implementing a generic interface method whose return type is a constructed generic type (delegate or plain class) fails to bind at all, with a false-negative `GS0187` ("does not implement interface method") diagnostic — even for the simplest non-generic-class-implementing-closed-generic-interface case. This is a distinct, pre-existing, non-delegate-specific interface-implementation-matching bug, unrelated to this fix. Recommend a follow-up issue to fix `GS0187` matching for generic-return interface methods; `InterfaceSymbol.SubstituteType` would then need the same `DelegateTypeSymbol` branch added.

### Tests
Added `Issue2338GenericNamedDelegateCallSiteEmitTests.cs` (compile + ILVerify + run), 8 facets: generic free-function return position, generic function parameter+return round-trip, method-level generics, containing-type generics, shared/static method on a generic class, chained/nested generic calls forwarding a delegate return, a two-type-param delegate with lambda-argument unification, and a non-generic control.

### Verification
- `Core.Tests`: 5992 passed (baseline unchanged)
- `Compiler.Tests`: 2917 passed, 0 failed — includes all 19 Issue2338-prefixed tests (7 original field-token, 4 closure-to-named-delegate, 8 new call-site)

No deferred work remains within the requested scope. The only outstanding item is the separately-reported, pre-existing `GS0187` bug above, which is out of scope for #2338 and recommended as its own follow-up issue.

Commit: 822be910

DavidObando:
## Update: remaining InterfaceSymbol generic delegate substitution gap — RESOLVED

Completed the last remaining discovered-but-deferred item from the #2338 audit: the source-interface (`InterfaceSymbol`) construction-time substitution gap for generic interface methods/properties whose return or parameter type is a named delegate (or plain generic struct/class) constructed over the interface's own type parameter or the method's own generic parameter.

Note: issue #2230 (fixed by PR #2249) covered the **imported-CLR interface** path — a distinct symbol/binding path from the **source-interface** (`InterfaceSymbol`) gap fixed here. The two are not conflated.

### Minimal repro
```
interface IFactory[T any] { func Make() Getter[T]; }
class Concrete : IFactory[int32] { func Make() Getter[int32] { ... } }
```
raised a false-negative `GS0187` ("Class 'Concrete' does not implement interface method") even though the implementation is exactly correct.

### Root cause — three independent, compounding gaps
1. **`InterfaceSymbol.SubstituteType`** (construction-time member substitution used when building a CONSTRUCTED generic interface's methods, e.g. `IFactory[int32]`) had no case for `DelegateTypeSymbol` nor `StructSymbol`. A constructed interface's method kept exposing the still-open definition type (`Getter[T]` / `Box[T]`) instead of the closed instantiation (`Getter[int32]` / `Box[int32]`), causing signature matching to compare an unsubstituted type argument against the implementer's closed argument and reject a correct implementation. Fixed by mirroring the `DelegateTypeSymbol`/`StructSymbol` branches already present in `Binder.SubstituteType` (this branch's own #2340 sibling fix) and `StructSymbol.SubstituteTypeForConstruction` (#1503).
2. **`InterfaceSymbol.SubstituteMethod`** never copied the original interface method's own (method-level) generic type parameters onto the constructed method (e.g. `func Transform[TOut](...)` on a generic interface `ITransformer[TIn]`), so a generic method on a CONSTRUCTED generic interface silently lost its arity. This made `TryBuildMethodTypeParameterMap` see a bogus arity mismatch and reject the candidate before signature comparison ever ran — a distinct bug only reachable once gap 1 stopped masking it (discovered via a combined interface-level + method-level generic delegate repro). Fixed by copying `m.TypeParameters` onto the substituted method.
3. **`DeclarationBinder.TypeSignaturesEquivalent`** had no case for `DelegateTypeSymbol`, so a named delegate constructed over a METHOD-level generic parameter still failed to compare as structurally equivalent, falling through to a `ClrType` fallback that doesn't apply the per-call method-type-parameter map. Fixed by adding a `DelegateTypeSymbol` branch mirroring the existing `StructSymbol`/`InterfaceSymbol` branches. Required normalizing `DelegateTypeSymbol.Definition` to itself before comparing, since (unlike `StructSymbol`/`InterfaceSymbol`) it is `null` — not self-referencing — on the open definition/non-generic delegates.

### Verification
Hand-written repros for every requested facet (interface-level delegate return/parameter, plain generic class return, method-level generics on a non-generic interface, combined interface-level + method-level generics over a multi-type-param delegate, containing-type generics, nested generic delegate signature `Getter[Box[T]]`) each compiled + ILVerify + ran correctly. Negative controls (mismatched type argument, a different named delegate entirely, a genuine method-level arity mismatch on a constructed generic interface) confirmed `GS0187` still fires correctly — the fix only removes false negatives, it does not loosen matching.

### Tests
- `Issue2338GenericInterfaceDelegateSubstitutionTests.cs` (Core.Tests): 12 pure-binder facts covering all three fixes plus negative controls and direct symbol-shape assertions.
- `Issue2338GenericInterfaceDelegateSubstitutionEmitTests.cs` (Compiler.Tests): 8 compile + ILVerify + run facets covering interface-level delegate return/parameter, plain generic class return, method-level generics, combined interface+method-level generics, containing-type generics, nested generic delegate, non-generic control.

### Full-suite results
- `Core.Tests`: **6004 passed** (5992 baseline + 12 new)
- `Compiler.Tests`: **2925 passed, 0 failed** (2917 baseline + 8 new)

No deferred work remains for issue #2338 or any of its follow-ups discovered along the way.

Commit: 68ec90c0

Issue #2348, opened 2026-07-15
gsc: NotNullWhen narrowing is lost after prior conditional assignment joins


DavidObando:
## Summary

A variable conditionally reassigned in an earlier unrelated `if` is not narrowed by a later `[NotNullWhen(false)]` guard such as `!string.IsNullOrWhiteSpace(value)`, producing GS0156/GS0154. A simple guard without the preceding branch works.

Real Oahu.Diagnostics occurrence: DiagnosticRunner generated code.

## Expected fix/tests

Preserve joined nullable flow state through conditional assignments, then apply imported NotNullWhen contracts at the later branch. Cover assigned/unassigned paths, true/false polarity, loops, and issue #178 controls.

Issue #2347, opened 2026-07-15
gsc: LINQ generic extension inference fails for bare BCL method-group predicates


DavidObando:
## Summary

`Enumerable.All<T>`/equivalent LINQ extension calls fail with GS0159 when the predicate argument is a bare imported BCL method group such as `Char.IsAsciiHexDigit`; replacing it with an equivalent lambda compiles.

Real Oahu.Diagnostics occurrence: `ExportCheck` calls `.All(Char.IsAsciiHexDigit)`.

## Expected fix

Generalize overload inference/delegate conversion so imported generic extension methods can infer element/delegate types from imported method groups, preserving overload ambiguity and negative diagnostics.

## Tests

Bare BCL static method group, equivalent lambda control, overloaded method groups, imported/source extension controls, and exact Oahu shape.

Discovered after Oahu PR #57 enabled Diagnostics translation.

Issue #2211, opened 2026-07-07
cs2gs: back-translated generator output emits short type/enum names without required imports (GS0113/GS0157/GS0202)


DavidObando:
### Summary
When the cs2gs translator back-translates C# that uses **fully-qualified names with no `using` directives** (the shape Roslyn source generators emit), it sometimes reduces a type/enum reference to its **short name without emitting a corresponding `import`**, producing G# that gsc then rejects (GS0113 / GS0157 / GS0202). Qualification is applied inconsistently within the same file.

### Evidence
Back-translating CommunityToolkit.Mvvm's generated `__KnownINotifyPropertyChangedArgs` (via the ADR-0145 gsgen host, PR #2208) yields (`samples/SourceGen/obj/.../gsgen/__KnownINotifyPropertyChangedArgs.g.gs`):
```gsharp
package CommunityToolkit.Mvvm.ComponentModel.__Internals

@System.ComponentModel.EditorBrowsable(EditorBrowsableState.Never)   // attr type qualified, enum arg NOT
internal class __KnownINotifyPropertyChangedArgs {
    shared {
        let Message PropertyChangedEventArgs = PropertyChangedEventArgs("Message")   // short, no import
    }
}
```
Resulting gsc errors:
```
error GS0113: Type 'PropertyChangedEventArgs' doesn't exist.
error GS0157: Cannot find type EditorBrowsableState. Are you missing an import?
error GS0202: Attribute arguments must be compile-time constants (...).   // downstream: EditorBrowsableState.Never never resolves
```
The MVVM C# source uses `global::System.ComponentModel.PropertyChangedEventArgs` and `global::System.ComponentModel.EditorBrowsableState.Never` (fully qualified). The translator kept the attribute *type* qualified but shortened the field type and the enum-argument value to bare names, without adding `import System.ComponentModel`.

### Expected
When back-translating code that lacks `using` directives, the translator should keep references resolvable — either preserve full qualification consistently, or synthesize the `import`s for the namespaces it shortens. (Note: preserving full qualification for a generic static access separately trips a gsc bug — see the companion issue on `Ns.Generic[Args].StaticMember` — so import synthesis is likely the more robust fix.)

_Discovered via Claude Code while implementing ADR-0145 (native source-generator host, PR #2208)._

PR #2355, opened 2026-07-15
Fix #2351: synthesize imports for extension-method calls relying on implicit/global usings

### LINKED ISSUE #2351: cs2gs: extension-method translations omit required namespace imports such as System.Linq
## Summary

cs2gs tracks namespaces used by shortened type references, but does not synthesize imports required solely by extension-method calls. Generated G# can therefore omit `import System.Linq` even when translated code uses LINQ extensions.

Root direction: `CSharpTypeMapper.TrackShortenedNamespace` covers type mapping, not extension method ownership/import requirements.

## Expected fix/tests

Track the declaring namespace of translated extension methods and emit the import when needed, without adding unused imports or duplicating explicit ones. Cover System.Linq and a custom extension namespace, plus exact Oahu.Diagnostics ExportCheck output.

DavidObando:
## Note on issue number

This branch/worktree was requested under the name "#2348", but `gh issue view 2348` shows that issue is actually about NotNullWhen narrowing lost after a prior conditional assignment — completely unrelated to the task description given for this work. Searching the repo's issues for the described defect (cs2gs omitting namespace imports required solely by translated extension-method calls) found **#2351**, whose title and body are an exact match. This PR addresses **#2351**; **#2348 remains unaddressed** and would need to be picked up separately if still desired.

Closes #2351

## Root cause

`CSharpTypeMapper.TrackShortenedNamespace` only fires for **type** references (via `QualifiedTypeName`) — never for extension-method call sites, since a call like `key.All(predicate)` names no type in its syntax. Combined with `TranslateImports` only scanning the **current file's own** `using` directives (not the whole compilation's global-usings), a file relying on `<ImplicitUsings>enable</ImplicitUsings>` (SDK auto-generates a separate `GlobalUsings.g.cs`) or a hand-authored project-wide `GlobalUsings.cs` for its extension-method namespace previously round-tripped to G# with **no import at all** for that namespace — producing unresolvable G#.

Confirmed via Oahu.Diagnostics' real `ExportCheck.cs`: `<ImplicitUsings>enable</ImplicitUsings>`, no explicit `using System.Linq;`, calls `key.All(char.IsAsciiHexDigit)`.

## Fix

Added `CSharpTypeMapper.TrackExtensionMethodNamespace(IMethodSymbol method)`, which records the declaring namespace of any resolved extension method (unwrapping `ReducedFrom` for reduced calls, and the C# 14 extension-block synthetic marker type) into the same `shortenedNamespaces` set already used for type-import tracking (issue #2211) — so the existing drain/dedup/own-package-skip logic in `TranslateDocument` applies automatically, with no new import-synthesis mechanism needed.

Wired at two call sites:
- Top of `TranslateInvocation` — covers all invocation-based extension calls regardless of which downstream branch rewrites the syntax.
- Top of `TranslateMemberAccess` — covers non-invoked bare method-group references (e.g. assigned to a delegate).

This generalizes across: reduced instance-form calls, unreduced static-form calls, bare sibling static-form calls, enum-extension positional rewrites, C# 14 extension-block members, non-invoked method-group references, BCL and source-defined (same-compilation) extensions, aliased namespaces, and nested namespaces.

## Tests

Added `Issue2351ExtensionMethodImportSynthesisTests.cs` (8 tests, modeled on the existing `Issue2211FullyQualifiedNoUsingImportSynthesisTests.cs` pattern):
1. `RealOahuExportCheckShape_ImplicitLinqUsing_SynthesizesImport` — the exact Oahu `ExportCheck` repro shape.
2. `LambdaEquivalentShape_MissingLinqUsing_SynthesizesImportAndCompilesEndToEnd` — same missing-using shape with a lambda predicate, validated end-to-end with real `gsc`.
3. `ExplicitUsingAlreadyPresent_NoDuplicateImportSynthesized` — control: explicit `using System.Linq;` present, asserts exactly one import (no duplicate).
4. `CustomExtensionNamespace_DifferentNamespaceSameCompilation_SynthesizesImport` — custom (non-BCL) extension in a different namespace (two files).
5. `CustomExtensionNamespace_SameNamespaceAsCallSite_NoSpuriousOwnPackageImport` — control: same-namespace extension, asserts no self-import.
6. `NestedNamespaceCustomExtension_SynthesizesImport` — multi-segment nested namespace extension.
7. `BareMethodGroupReference_NonInvokedExtension_SynthesizesImport` — non-invoked method-group assigned to a delegate, exercising the `TranslateMemberAccess` injection point.
8. `AliasedStaticFormCustomExtension_SynthesizesRealNamespaceImport_NotAliasName` — extension called via a type alias in unreduced static form; verifies the real-namespace import coexists correctly with the pre-existing (unrelated, correct) alias-import mechanism.

## Validation

- `Cs2Gs.Translator` and `Cs2Gs.Tests` build clean with analyzers (0 warnings/errors, StyleCop/SA rules enforced).
- New test suite: **8/8 pass**.
- Verified via before/after comparison (patch apply/revert) that all 8 new tests fail without the fix and pass with it, while the 2 negative-control tests pass regardless (as expected of a control).
- Full `Cs2Gs.Tests` suite (serial execution, 1366 tests): **1359 passed, 7 failed** — confirmed via an identical run on baseline (`origin/main`, no fix) that all 7 failures are **pre-existing** corpus/pipeline compile-error tests unrelated to this change (`MigrationPipelineTests.L1/L2_StagesOneAndTwo_AreGreen_WithZeroArtifacts`, `Issue1911...G06TypesConsole_MigratesGreen_EndToEnd`, `Issue1933...G12UnsafeConsole_MigratesGreen_EndToEnd`, `IlVerifyStageTests.L1_StageThree_IsGreen_WithZeroArtifacts`, `TestParityStageTests.L1_AllFourStages_AreGreen_WithStdoutParity`, `TestParityStageTests.L1_WrongGolden_ProducesTestParityFailureArtifact`) — **zero regressions introduced**.

## Deferred / known limitations

- The exact bare-method-group Oahu `ExportCheck` shape (`key.All(char.IsAsciiHexDigit)`) is separately affected by an unrelated bug (#2347 — imported/BCL method-group-to-delegate inference), fixed in a different PR not yet merged as of this branch's point. This fix's correctness for that exact shape is proven via printed-text assertions + parse-level round-trip (test #1), while full real-`gsc` end-to-end compile validation uses an equivalent lambda predicate unaffected by #2347 (test #2). Once #2347 merges, the exact bare-method-group shape could additionally be validated end-to-end with real `gsc`.
- Issue #2348 (NotNullWhen narrowing) is untouched by this PR.

### DIFF
--- a/tools/cs2gs/Cs2Gs.Tests/Issue2351ExtensionMethodImportSynthesisTests.cs
+++ b/tools/cs2gs/Cs2Gs.Tests/Issue2351ExtensionMethodImportSynthesisTests.cs
@@ -1,0 +1,399 @@
+// <copyright file="Issue2351ExtensionMethodImportSynthesisTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using Cs2Gs.CodeModel.Ast;
+using Cs2Gs.CodeModel.Printing;
+using Cs2Gs.CodeModel.RoundTrip;
+using Cs2Gs.Translator;
+using Cs2Gs.Translator.Loading;
+using Xunit;
+
+namespace Cs2Gs.Tests;
+
+/// <summary>
+/// Issue #2351: <see cref="CSharpTypeMapper.TrackShortenedNamespace"/> records the
+/// namespace of every TYPE reference the translator shortens (issue #2211), so a
+/// missing <c>using</c> for a shortened type name synthesizes a matching
+/// <c>import</c>. An extension-method CALL SITE (<c>key.All(...)</c>) names no
+/// type at all — neither <c>Enumerable</c> nor <c>System.Linq</c> appear in its
+/// syntax — so that tracking never observed it. A C# file that reaches the
+/// extension only through an IMPLICIT <c>using</c> (SDK <c>ImplicitUsings</c>
+/// global usings, or a hand-authored <c>GlobalUsings.cs</c>, or a namespace-less
+/// source-generator file per ADR-0145) previously round-tripped to G# with the
+/// call intact but NO import for its owning namespace at all — unresolvable
+/// output (GS0159), exactly the shape observed in Oahu.Diagnostics'
+/// <c>ExportCheck.Run</c> (<c>key.All(char.IsAsciiHexDigit)</c>, whose project
+/// enables <c>ImplicitUsings</c> and has no explicit <c>using System.Linq;</c>).
+/// <para>
+/// The fix (<see cref="CSharpTypeMapper.TrackExtensionMethodNamespace"/>) records
+/// the declaring namespace of any RESOLVED extension method — whether reached via
+/// an invocation (<see cref="Cs2Gs.Translator.CSharpToGSharpTranslator"/>'s
+/// <c>TranslateInvocation</c>, covering the reduced instance form, the unreduced
+/// static form, and the bare-sibling-static form) or via a bare, non-invoked
+/// method-group reference (<c>TranslateMemberAccess</c>) — into the SAME
+/// <see cref="CSharpTypeMapper.ShortenedNamespaces"/> set that issue #2211's fix
+/// already drains into synthesized imports, so the existing
+/// dedup-against-explicit-usings and skip-own-package logic in
+/// <c>CSharpToGSharpTranslator.TranslateDocument</c> applies uniformly with no
+/// separate code path.
+/// </para>
+/// </summary>
+public class Issue2351ExtensionMethodImportSynthesisTests
+{
+    [Fact]
+    public void RealOahuExportCheckShape_ImplicitLinqUsing_SynthesizesImport()
+    {
+        // Mirrors the exact reported shape: Oahu.Diagnostics' ExportCheck.cs has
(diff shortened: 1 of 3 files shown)


DavidObando:
Update: merged latest `origin/main` (now includes #2347's fix, PR #2353) into this branch. No conflicts.

- The exact Oahu Diagnostics repro shape — `key.All(char.IsAsciiHexDigit)`, no explicit `using System.Linq;`, `<ImplicitUsings>enable</ImplicitUsings>` — is now validated **end-to-end with the real gsc compiler** (`RealOahuExportCheckShape_ImplicitLinqUsing_SynthesizesImportAndCompilesEndToEnd`), not just structurally as before. The bare-method-group form now resolves correctly because #2347 fixed the separate method-group-to-delegate inference issue that previously blocked it.
- The lambda-equivalent end-to-end test is kept alongside it (both forms independently prove the import-synthesis fix is sufficient on its own).
- Full `GSharp.sln` build: 0 warnings/errors.
- Full `Cs2Gs.Tests` suite (serial, 1366 tests): 1359 passed, 7 failed — the same 7 pre-existing corpus/pipeline compile-error failures already confirmed unrelated to this change (verified identical on pristine `origin/main`).
- No deferred work remains for this fix.

PR #2376, opened 2026-07-15
Fix expression-tree comparison operator factory overload selection (#2373)

### LINKED ISSUE #2373: gsc: expression-tree comparison operators request nonexistent 3-argument Expression overload and crash GS9998
## Summary

`ExpressionTreeLowerer` lowers CLR-resolved comparison operators (`==`, `!=`, `<`, `<=`, `>`, `>=`) by looking up a three-argument `System.Linq.Expressions.Expression` factory overload `(Expression, Expression, MethodInfo)`. Comparison factories do not expose that overload; they require `(Expression, Expression, bool liftToNull, MethodInfo)`. Reflection lookup returns null and compilation crashes with GS9998.

## Minimal shape

Use a referenced assembly containing a data class/type `Book(Asin string)`, then compile an expression-tree lambda in a consumer:

```gsharp
let result = books.AsQueryable().FirstOrDefault((b Book) -> b.Asin == asin)
```

Actual result:

```text
GS9998: InvalidOperationException: Required method 'System.Linq.Expressions.Expression.Equal' is not available.\n```\n\nThe imported/reference type causes operator binding to use the CLR-operator fallback stream; an equivalent source-declared type currently avoids that path. Oahu.Diagnostics triggers this with `db.Books.FirstOrDefault((b Book) -> b.Asin == asin)`.\n\n## Root cause\n\n`ExpressionTreeLowerer.BuildClrBinaryOperatorExpression` requests `Equal`/`NotEqual`/`LessThan`/`LessThanOrEqual`/`GreaterThan`/`GreaterThanOrEqual` using parameter types `(Expression, Expression, MethodInfo)`. That overload exists for arithmetic/bitwise factories but not comparisons.\n\n## Expected fix/tests\n\nFor comparison factory names, resolve the four-argument overload and emit the correct `liftToNull` boolean before the method-info argument. Preserve nullable/lifted semantics from the bound operator. Cover all six comparisons, built-in and user-defined operators, source/imported types, nullable operands/lift-to-null behavior, generic expression-tree delegates, runtime expression node types, execution, and IL verification.\n\nDiscovered in Oahu migration round 14 against PR #2370.

DavidObando:
## Summary

Fixes issue #2373: `gsc` crashed with GS9998 when an `Expression<Func<T, bool>>` expression tree used a comparison operator (`==`, `!=`, `<`, `<=`, `>`, `>=`) on a built-in or user-defined struct — e.g. the reported `Oahu.Diagnostics` `Book.Asin` comparison scenario.

## Root cause

`ExpressionTreeLowerer.BuildClrBinaryOperatorExpression` requested a **non-existent 3-arg** `Expression.Equal/NotEqual/LessThan/LessThanOrEqual/GreaterThan/GreaterThanOrEqual(Expression, Expression, MethodInfo)` factory overload. Unlike the arithmetic/bitwise/shift/logical factories (which correctly use a 3-arg overload) and the unary factories (which correctly use a 2-arg overload), these six comparison factories only expose a **4-arg** `(Expression, Expression, bool liftToNull, MethodInfo)` overload in `System.Linq.Expressions.Expression`.

## Fix

`BuildClrBinaryOperatorExpression` now branches on comparison vs. other binary operator kinds, using the correct 4-arg overload with a literal `liftToNull: false` argument for comparisons. This matches Roslyn's own emission: a comparison's boolean *result* is never itself nullable/lifted, regardless of whether the operands are nullable — `IsLiftedToNull` naturally stays `false` while `IsLifted` reflects the nullable operand types.

### Second, tightly-coupled bug discovered and fixed

While validating the fix end-to-end (real two-assembly compile + ILVerify + runtime execution), a second, entirely separate, **pre-existing** bug was uncovered: `MethodBodyEmitter.EmitLiteral` had no case for a `MethodInfo`-valued `BoundLiteralExpression`. This meant **any** CLR operator (arithmetic *or* comparison) lowered into an expression tree would crash at emission time with `NotSupportedException: Literal of CLR type 'RuntimeMethodInfo' is not yet supported` — independent of, and prior to, the #2373 overload-selection bug. Confirmed pre-existing by reverting the lowerer fix and re-running the arithmetic-only regression test, which failed identically against the original code.

Fixed by adding:
- `EmitContext.CoreRuntimeMethodHandleType` / `CoreMethodBaseType`
- `WellKnownReferences.GetMethodFromHandleReference()` / `GetMethodFromHandleWithDeclaringTypeReference()`
- `MethodBodyEmitter.EmitMethodInfoLiteral`, emitting `ldtoken method [, ldtoken declaringType] ; call MethodBase.GetMethodFromHandle(...) ; castclass MethodInfo` — mirroring the existing `typeof(T)` literal emission pattern.

### Sibling audit

Reviewed all other `Expression` factory-method lookups in `ExpressionTreeLowerer` (arithmetic/bitwise/shift/logical: 3-arg; unary: 2-arg) empirically via reflection against `System.Linq.Expressions.Expression`. Confirmed **no other factory family has a matching overload-selection gap** — only the six comparison factories were affected. No changes made beyond this proven gap.

## Tests

- `test/Core.Tests/CodeAnalysis/Binding/Issue2373ExpressionTreeComparisonOverloadTests.cs` (24 tests, binder/lowering level):
  - All six comparison operators on `TimeSpan` (plain, nullable-left, both-nullable operands).
  - Arithmetic/unary sibling regression (proves no collateral change to already-correct factory families).
  - Source-declared struct operator control (Stream D — same-compilation operator, unaffected by this fix).
  - Generic expression-tree delegate closing a type parameter over `TimeSpan` at the call site (realistic generic-substitution shape).
  - Real C#-authored imported struct operator (mirrors `Oahu.Diagnostics` `Book.Asin`) reaching the CLR-factory fallback path.
  - Negative/ambiguity controls: undefined operator still reports `GS0129`; a genuine ambiguous overload (two real C#-authored mutually-implicitly-convertible structs each declaring `==`) still reports `GS0160`.
- `test/Compiler.Tests/Emit/Issue2373ExpressionTreeClrComparisonEmitTests.cs` (13 tests, real two-assembly compile + ILVerify + runtime execution):
  - `TimeSpan`, all six operators — ILVerify-clean IL, asserts `BinaryExpression.Method`/`IsLiftedToNull`, executes the compiled lambda and asserts the correct boolean result.
  - Arithmetic sibling-audit regression (specifically proves the MethodInfo-emission fix, since arithmetic already had correct overload selection pre-fix but crashed at emission).
  - Real C#-authored imported "Asin" strongly-typed-ID struct (mirroring `Oahu.Diagnostics`/`Book.Asin`), all six operators — ILVerify-clean IL and correct runtime results.

## Deferred / discovered-but-out-of-scope work

While building the "real C#-authored imported operator" test scenario, a **third, separate, pre-existing** gap was discovered: G#-authored user-defined operators (`func (a T) operator ==(b T) bool`) are emitted as **instance methods without `MethodAttributes.SpecialName`**. This means they can never be discovered by `ClrOperatorResolution`'s `BindingFlags.Static` + `IsSpecialName` reflection filter, making G#-authored operators unusable as `==`/etc. from real CLR/C# consumers, or even from Stream C's own CLR-fallback resolution. This is a legitimate, real bug, but it is a distinct emission-shape defect outside #2373's stated scope (`Expression` factory overload selection — "change only proven gaps"). **Recommended as a follow-up issue.**

## Validation

- Full solution build with analyzers enabled (`-warnaserror`): **0 warnings, 0 errors**.
- `Core.Tests`: 6237/6238 passed (1 pre-existing unrelated `[Skip]` baseline-regen test).
- `Compiler.Tests`: 3041/3041 passed.
- `InternalAnalyzers.Tests`: 7/7 passed.
- `Cs2Gs.Tests` (serial full suite, as required): 1387/1387 passed.
- Branch already up to date with latest `origin/main` (`bd990954`).

Closes #2373

### DIFF
--- a/src/Core/CodeAnalysis/Emit/EmitContext.cs
+++ b/src/Core/CodeAnalysis/Emit/EmitContext.cs
@@ -206,6 +206,23 @@ internal sealed class EmitContext
     public Type CoreRuntimeTypeHandleType { get; set; }
 
     /// <summary>
+    /// Gets or sets the BCL <see cref="System.RuntimeMethodHandle"/> type
+    /// resolved from the target framework's <see cref="References"/>.
+    /// Issue #2373: backs emission of a <see cref="System.Reflection.MethodInfo"/>
+    /// runtime constant (<c>ldtoken method ; call
+    /// MethodBase.GetMethodFromHandle</c>) for expression-tree lowering's CLR
+    /// operator-method arguments.
+    /// </summary>
+    public Type CoreRuntimeMethodHandleType { get; set; }
+
+    /// <summary>
+    /// Gets or sets the BCL <see cref="System.Reflection.MethodBase"/> type
+    /// resolved from the target framework's <see cref="References"/>. See
+    /// <see cref="CoreRuntimeMethodHandleType"/>.
+    /// </summary>
+    public Type CoreMethodBaseType { get; set; }
+
+    /// <summary>
     /// Gets or sets the BCL <see cref="Enum"/> type resolved from the target
     /// framework's <see cref="References"/>.
     /// </summary>
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs
@@ -643,10 +643,52 @@ internal sealed partial class MethodBodyEmitter
             case decimal m:
                 this.EmitDecimalLiteral(m);
                 break;
+            case MethodInfo method:
+                // Issue #2373: expression-tree lowering (ExpressionTreeLowerer)
+                // passes a CLR operator/user MethodInfo as a literal-typed
+                // argument to Expression factories (Expression.Equal(l, r,
+                // liftToNull, method), Expression.Add(l, r, method), the
+                // string-concat Add(l, r, method) overload, ...). A MethodInfo
+                // cannot be embedded as a metadata constant directly; rebuild
+                // it at run time the same way the C# compiler does: `ldtoken
+                // method [, ldtoken declaringType] ; call
+                // MethodBase.GetMethodFromHandle(...) ; castclass MethodInfo`.
+                this.EmitMethodInfoLiteral(method);
+                break;
             default:
                 throw new NotSupportedException(
                     $"Literal of CLR type '{literal.Value?.GetType()}' is not yet supported.");
         }
+    }
+
+    private void EmitMethodInfoLiteral(MethodInfo method)
+    {
+        this.il.OpCode(ILOpCode.Ldtoken);
+        this.il.Token(this.outer.GetMethodEntityHandle(method));
+
+        // The CLR requires the 2-arg GetMethodFromHandle(RuntimeMethodHandle,
+        // RuntimeTypeHandle) overload whenever the declaring type is a
+        // generic instantiation (e.g. a user operator on `Money<Currency>`);
+        // the 1-arg overload throws ArgumentException at run time for those.
+        // A generic-type-DEFINITION declaring type never occurs here because
+        // operator methods are only ever reached through a closed (fully
+        // substituted) receiver type.
+        if (method.DeclaringType is { IsGenericType: true } declaringType)
+        {
+            this.il.OpCode(ILOpCode.Ldtoken);
+            this.il.Token(this.outer.GetTypeReference(declaringType));
+            this.il.Call(this.outer.wellKnown.GetMethodFromHandleWithDeclaringTypeReference());
+        }
+        else
+        {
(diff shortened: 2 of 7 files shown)


DavidObando:
Merged latest `origin/main` (`f8261164`, includes #2370 and #2374) into this branch — clean merge, no conflicts. Re-ran full validation afterward:
- Full solution build with analyzers (`-warnaserror`): 0 warnings, 0 errors.
- `Core.Tests`: 6261/6262 passed (1 pre-existing unrelated skip).
- `Compiler.Tests`: 3074/3074 passed.
- `InternalAnalyzers.Tests`: 7/7 passed.
- `Cs2Gs.Tests` (serial, chunked due to a transient VSTest-host crash under heavy shared-environment load): 0 failures across all chunks/tests.

No regressions from the merge.

Issue #889, opened 2026-06-19
Cannot pass lambda as parameter to function that receives System.Action


DavidObando:
Have the following C# code:
```csharp
class CliEnvironment {
  private Action restoreAction;
  public static void RegisterRestore(Action restore) => restoreAction = restore;
  ...
}
```

From G# we try to call this:
```gsharp
CliEnvironment.RegisterRestore(func() { called = called + 1 }) // this works

CliEnvironment.RegisterRestore(() -> called += 1) // this doesn't work
```

PR #2408, opened 2026-07-16
Fix target-type inference for untyped lambdas against imported CLR delegates/events

### LINKED ISSUE #2389: gsc: untyped lambda inference fails for imported CLR delegate event +=
Discovered while expanding #2380 coverage.

An untyped lambda used with += against an imported CLR delegate/event fails target-type inference. Typed lambdas or source-defined delegate shapes avoid the failure, showing this is an imported-delegate inference gap rather than event emission.

Add a minimal imported event/delegate repro and fix target typing for += and related assignment contexts.

DavidObando:
## Problem

An untyped arrow-lambda (e.g. `(sender, e) -> {...}`) used with `+=` against an imported CLR delegate/event failed target-type inference with GS0304, even though a typed lambda (`(sender object, e EventArgs) -> ...`) or a source-defined delegate shape worked fine. The same gap also affected plain `=` re-assignment of a previously-declared delegate-typed variable/field, even though the identical lambda works as a declaration initializer.

Closes #2389.

## Root cause

Two binder entry points bound the lambda via the generic, target-type-less `BindExpression` path before falling back to a post-bind conversion:

- `BindEventSubscriptionHandler` (`src/Core/CodeAnalysis/Binding/ExpressionBinder.Async.cs`) — handles `+=`/`-=` event subscription (despite the file name, this is where event-subscription binding lives after the PR #2404 file split).
- `BindAssignmentRhs` (`src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs`) — the shared RHS binder for plain `=` assignment, used by bare variable assignment, member-field/property assignment, and inherited-CLR-instance-member bare writes.

`LambdaBinder.BindLambdaExpression` only infers an omitted parameter type from a target function type when a matching-arity target is supplied *before* binding. Since neither caller passed one, the lambda's own binding failed before any later conversion had a chance to reshape it — no post-bind conversion can retroactively fix an already-failed bind.

## Fix

In both binders, when the RHS/handler is a `LambdaExpressionSyntax` and the target delegate/event/variable type resolves to a function-type shape via the existing shared helper `MemberLookup.TryGetLambdaTargetFunctionTypeFromSymbol` (issue #889), bind the lambda through `LambdaBinder.BindLambdaExpression` with that target type instead of the generic `BindExpression`. This uniformly covers native G# function types, user-declared named delegates, and imported CLR delegates/events, reusing the existing delegate-target-resolution infrastructure (already used for call-argument lambda inference) rather than duplicating it. The pre-existing post-bind conversion logic (issue #2066) is left untouched and still applies to the result.

Fixing the single shared `BindAssignmentRhs` helper transparently also fixes plain re-assignment through member-field/property writes and inherited-CLR-instance-member bare writes, since they all route through it.

**Intentionally out of scope:** static field/property write paths (`Type.Field = value` for imported CLR classes / user structs / interfaces) call `BindExpression` directly and bypass `BindAssignmentRhs` entirely — this is a separate, pre-existing target-typing gap (it also affects `if`/`switch` expressions, not just lambdas) and is not part of the same inference helper this issue is scoped to, so it was left unchanged.

## Tests

- `test/Core.Tests/CodeAnalysis/Binding/Issue2389ImportedDelegateLambdaInferenceTests.cs` (new, 15 tests): binder-level diagnostics covering instance/static imported CLR events, add/remove symmetry, source-defined function-type events, user-declared named-CLR-delegate events, interface-typed receivers, plain variable/field re-assignment, inferred parameter/return types via generic BCL delegates (`Predicate<T>`, `Comparison<T>`), typed-lambda/method-group regression guards, and arity-mismatch negative controls (still correctly report GS0304/GS0155).
- `test/Compiler.Tests/Emit/Issue2389ImportedDelegateLambdaEmitTests.cs` (new, 11 tests): compile/run/ILVerify tests against a minimal sibling C# assembly (`ProbeRef2389`) exposing custom delegates (zero/one/two-param, inferred return) and instance/static events, covering add/remove symmetry end-to-end plus negative ambiguity/invalid-signature controls.

## Validation

- `test/Core.Tests`: 6340 passed, 0 failed.
- `test/Compiler.Tests`: 3141 passed, 0 failed (full suite, including sibling-assembly-dependent tests after building `src/Sdk/Gsharp.Extensions`).
- `test/InternalAnalyzers.Tests`: 7 passed, 0 failed.
- `tools/cs2gs/Cs2Gs.Tests`: 1423 passed, 0 failed, run serialized with `RunConfiguration.DisableParallelization=true`.

No Oahu-specific fixtures were modified.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs
@@ -211,6 +211,25 @@ internal sealed partial class ExpressionBinder
             return BindExpression(syntax, targetType);
         }
 
+        // Issue #2389: an untyped arrow-lambda RHS (`f = (s, e) -> ...`)
+        // needs its omitted parameter types resolved from the LHS's declared
+        // delegate/function-type shape *before* binding, exactly like a
+        // `let`/`var` declaration initializer already does (StatementBinder)
+        // — otherwise re-assigning a previously-declared delegate-typed
+        // variable/field with an untyped lambda reports GS0304 even though
+        // the identical lambda works fine as the declaration's own
+        // initializer. Reuse the shared delegate-target-type resolver
+        // (issue #889) so this covers native G# function types,
+        // user-declared named delegates, and imported CLR delegates
+        // uniformly; the conversion performed by the caller below still
+        // reshapes the result to the exact target type.
+        if (targetType != null
+            && syntax is LambdaExpressionSyntax lambdaSyntax
+            && MemberLookup.TryGetLambdaTargetFunctionTypeFromSymbol(targetType, out var lambdaTargetFunctionType))
+        {
+            return lambdas.BindLambdaExpression(lambdaSyntax, lambdaTargetFunctionType);
+        }
+
         return BindExpression(syntax);
     }
 
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Async.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Async.cs
@@ -347,7 +347,31 @@ internal sealed partial class ExpressionBinder
             return BindEventSubscriptionHandlerFromBoundAccessor(memberAccess, boundReceiver, targetDelegateType);
         }
 
-        var bound = BindExpression(handlerSyntax);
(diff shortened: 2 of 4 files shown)


DavidObando:
Closing without merge per crash-recovery instruction: discard in-flight work after PR #2404 refactoring. This issue will be restarted later from current main.

Issue #2498, opened 2026-07-19
gsc: imported generic inference drops nullable-reference lambda result types


DavidObando:
## Summary

Imported generic-method inference strips a nullable-reference wrapper when a method type parameter is inferred from a lambda's return type. Consequently, `Enumerable.Select<TSource,TResult>` infers `TResult = R` instead of `R?`, and downstream generic returns (`IEnumerable<TResult>`, `ToArray`) become non-null-element collections.

The lambda's own conditional expression correctly has nullable type; the annotation is lost only while unifying it into the imported method's MVar. This is a residual follow-up to #1428, not a recurrence of its order-dependent conditional common-type bug.

## Oahu evidence

Full migration with `gsc 0.3.153+fa0903cb9e`:

- Canonical fingerprint: `sha256:52bb019259038bc62b9dbc210c3cb89f12b31023f328b2b9ea36af3847f9a706`
- C# source: `src/Oahu.Decrypt/Mpeg4/Chunks/InterleavedIterator[T].cs:43-54`
- Emitted G#: `corpus_Oahu.Decrypt/Mpeg4/Chunks/InterleavedIterator[T].gs:17:49`
- Diagnostic: `GS0155: Cannot convert type 'nil' to 'System.Collections.Generic.IEnumerator`1[T]'.`

Original C# explicitly declares nullable elements:

```csharp
IEnumerator<T>?[] enumerators = Enumerables
    .Select(e => e.GetEnumerator())
    .Select(e => e.MoveNext() ? e : null)
    .ToArray();

// ...
enumerators[currentIndex] = null;
```

cs2gs emits:

```gsharp
let enumerators = Enumerables
    .Select((e IEnumerable[T]) -> e.GetEnumerator())
    .Select((e IEnumerator[T]) -> if e.MoveNext() { e } else { default(IEnumerator[T]?) })
    .ToArray()

// ...
enumerators[currentIndex] = nil
```

The lambda result is visibly `IEnumerator[T]?`, but `ToArray()` is bound as `[]IEnumerator[T]`.

## Stable minimal repro

Valid nullable-enabled C#:

```csharp
using System.Collections.Generic;
using System.Linq;

public static class Repro
{
    public static IEnumerator<T>?[] Build<T>(IEnumerator<T>[] source)
    {
        IEnumerator<T>?[] values = source
            .Select(item => item.MoveNext() ? item : null)
            .ToArray();
        values[0] = null;
        return values;
    }
}
```

[...]

PR #2524, opened 2026-07-19
Fix nullable imported generic base conversions

### LINKED ISSUE #2523: gsc: nullable inferred arguments break IIncludableQueryable base-interface conversion
## Summary

An imported generic method can correctly infer a nullable-reference type argument from an expression-tree lambda, but the resulting imported constructed interface then fails to convert to its declared base interface. This breaks chained Entity Framework Core `Include` calls whenever the first navigation property is nullable.

`Include<TEntity,TProperty>` returns `IIncludableQueryable<TEntity,TProperty>`, and `IIncludableQueryable<out TEntity,out TProperty>` inherits `IQueryable<TEntity>`. C# therefore permits another `Include` call on the result. gsc binds the first call as `IIncludableQueryable<TEntity,TProperty?>`, but cannot use that value as `IQueryable<TEntity>` for the second call, producing GS0159. A single call returned as `IQueryable<TEntity>` exposes the underlying GS0155 conversion failure directly.

This is independent of #2521 and #2516. It is a downstream follow-up to #2498/#2499: nullable lambda inference is now preserved, but the imported generic return's base-interface relationship/conversion is not usable with that recovered nullable argument.

## Oahu evidence

Fresh migration from `origin/main` commit `65295c63` (#2520), gsc `0.3.165+65295c6339`:

```text
sha256:7a093a75108b931934ade9fee7d5610667fe17af436c605f06b18bdf27bec265
corpus_Oahu.Core/BookLibrary.gs:84:91
GS0159: Cannot find function Include.
```

Original C# (`src/Oahu.Core/BookLibrary.cs:123-136`):

```csharp
IEnumerable<Book> books = dbContext.Books
    .Include(b => b.Conversion)
    .Include(b => b.Components)
    .ThenInclude(c => c.Conversion)
    // ...
    .ToList();
```

Generated G#:

```gsharp
let books IEnumerable[Book] = dbContext.Books!!
    .Include((b Book) -> b.Conversion)
    .Include((b Book) -> b.Components)
    .ThenInclude((c Component) -> c.Conversion)
    // ...
    .ToList()
```

The diagnostic column is the **second** `Include`. The first call binds with a nullable navigation result, but its `IIncludableQueryable[Book, Conversion?]` result is not accepted as the `IQueryable[Book]` receiver required by the next imported extension call.

## Stable minimal repro

Two C# projects using `Microsoft.EntityFrameworkCore` 10.0.0.

### Data project

```csharp
using Microsoft.EntityFrameworkCore;

namespace Data;

public sealed class Child { }
public sealed class Other { }

public sealed class Entity
{
    public Child? Child { get; set; }
    public Other? Other { get; set; }
}

public sealed class Context : DbContext
{
    public DbSet<Entity> Entities { get; set; } = null!;
}
```

### Consumer project

```csharp
using System.Linq;
using Microsoft.EntityFrameworkCore;
using Data;

namespace Consumer;

public static class Query
{
    public static IQueryable<Entity> Build(Context context) =>
        context.Entities
            .Include(entity => entity.Child)
            .Include(entity => entity.Other);
}
```

Both C# projects build successfully. Default `cs2gs migrate --via-sdk` emits:

```gsharp
func Build(context Context) IQueryable[Entity] ->
    context.Entities
        .Include((entity Entity) -> entity.Child)
        .Include((entity Entity) -> entity.Other)
```

and fails at the second call:

```text
sha256:10a9b342da4569a24f689e79c0e399a7a8901dd903120a94391195e83763d626
Query.gs:9:117
GS0159: Cannot find function Include.
```

A nullable-oblivious Data project whose `Child` property is taint-promoted by a null check reproduces the identical fingerprint, matching Oahu's migration topology.

## Isolated conversion evidence

Keeping only the first call and assigning/returning it as `IQueryable<Entity>` reports the actual failed relationship:

```text
GS0155: Cannot convert type
'Microsoft.EntityFrameworkCore.Query.IIncludableQueryable`2[Data.Entity, Data.Child?]'
to 'System.Linq.IQueryable`1[Data.Entity]'.
```

Controls:

- Making the first navigation property non-null (`Child` instead of `Child?`) makes the G# compile stage pass.
- A direct C# upcast from `IIncludableQueryable<Entity,Child?>` to `IQueryable<Entity>` is valid.
- A direct translated G# upcast with same-compilation type arguments also passes; the failure appears on the constructed return recovered from imported generic method inference.
- The first `Include` itself binds; failure begins when consuming its inferred return through the declared base interface.

## Expected behavior

`IIncludableQueryable<TEntity,TProperty?>` must retain the CLR-declared `IQueryable<TEntity>` base-interface relationship. It must be a valid receiver for subsequent `Include` calls and a valid implicit return/assignment to `IQueryable<TEntity>`, regardless of nullability in the unrelated `TProperty` slot.

## Root-cause direction

#2499 preserved nullable-reference lambda results while inferring imported method type parameters and carried them into imported return constructions. The remaining gap appears where that constructed imported return participates in CLR/base-interface conversion:

- `MemberLookup` recovers `TProperty = Child?` and constructs the imported `IIncludableQueryable<Entity,Child?>` return;
- the construction's CLR identity and/or substituted base-interface list is not compatible with the independently bound `IQueryable<Entity>` target;
- `Conversion.Classify` eventually misses the standard #521 imported reference upcast and reports GS0155; extension candidate filtering then surfaces GS0159.

Investigate the handoff among recovered nullable MVars, `ImportedTypeSymbol` constructed returns, substituted imported base interfaces, cross-context type identity, and `ClrTypeUtilities.IsAssignableByName`. Do not strip `?` to make the call bind: #2498 requires the inferred nullable result to remain observable. The base relationship simply must not depend on the unrelated nullable type argument.

## Contributor-ready coverage requirements

- `IIncludableQueryable<TEntity,TProperty?> -> IQueryable<TEntity>` assignment, return, argument, and extension-receiver conversions.
- Consecutive `Include` calls and `Include(...).ThenInclude(...).Include(...)` chains.
- Explicit nullable annotations and nullable-oblivious sibling-source taint promotion.
- Source/source, source/metadata, metadata/source, and metadata/metadata entity/navigation types.
- Nullable type arguments in unrelated covariant, contravariant, and invariant slots of imported derived interfaces.
- Imported generic methods returning interfaces whose base interface substitutes only a subset of method type parameters.
- Same-compilation and referenced-project type arguments; multiple metadata load/reflection contexts.
- Preserve the nullable inferred argument in reflection/emitted signatures while retaining all declared base-interface conversions.
- Non-null controls, value-type `Nullable<T>` controls, nested generic nullable results, and multiple inference sources.
- Runtime query execution, reflection, C# consumer compilation, ILVerify, and deterministic extension overload selection.
- Preserve #1927, #2135, #2498/#2499, #2514/#2515, and #2520 coverage.
- Default `--via-sdk` two-project EF minimal plus the exact Oahu.Core chained query.

DavidObando:
Fixes #2523

## Summary
- classify imported constructed generic base/interface conversions from symbolic substitutions instead of erased CLR probe shapes
- recover generic method arguments from plain closed imported receivers and project reduced extension receivers through the declared hierarchy
- reify cross-context core type arguments and emit accepted imported reference upcasts as no-ops
- add binder, EF Include/ThenInclude chain, runtime, reflection, C# consumer, ILVerify, and default `--via-sdk` two-project regressions

## Validation
- Core targeted issue/control tests: 30 passed
- Compiler targeted runtime/emit/control tests: 13 passed
- Release SDK build: succeeded
- isolated-cache default `--via-sdk` issue #2523 + #2498 pipeline tests: 2 passed

### DIFF
--- a/src/Core/CodeAnalysis/Binding/Conversion.cs
+++ b/src/Core/CodeAnalysis/Binding/Conversion.cs
@@ -402,6 +402,29 @@ public sealed class Conversion
             return Conversion.Identity;
         }
 
+        // Issue #2523: a constructed imported reference type may carry
+        // symbolic arguments that its erased CLR shape cannot represent. For
+        // example, imported generic inference can surface
+        // IIncludableQueryable<TEntity, TProperty?> symbolically while its
+        // reflection probe shape is IIncludableQueryable<object, object>.
+        // Testing that erased shape against IQueryable<TEntity> loses the
+        // declared base-interface substitution and rejects a valid upcast.
+        //
+        // Project the source's open interface/base hierarchy through its
+        // symbolic arguments, then classify the resulting constructed target
+        // with ordinary CLR variance rules. This is framework-independent and
+        // applies equally to imported interfaces/classes, inferred or explicit
+        // method arguments, nested generic arguments, and cross-context CLR
+        // identities.
+        if (from is ImportedTypeSymbol fromConstructedReference
+            && to is ImportedTypeSymbol toConstructedReference
+            && TryClassifyConstructedImportedReferenceConversion(
+                fromConstructedReference,
+                toConstructedReference))
+        {
+            return Conversion.Implicit;
+        }
+
         // Phase 3.C.1 / ADR-0001: T → T? is an implicit widening; T? → T?
         // when underlyings match is identity. T? → T requires the bang
         // operator (Phase 3.C.3) and is not implicit here.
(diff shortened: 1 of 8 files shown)


DavidObando:
Closing to avoid duplicate work; another agent is completing #2523.

PR #2528, opened 2026-07-20
Fix imported interface indexer hiding

### LINKED ISSUE #2525: gsc: imported interfaces that hide inherited indexers are reported not indexable
## Summary

gsc reports an imported interface as non-indexable when it redeclares an inherited indexer with C# `new`. Both the derived and base `Item` properties are collected as CLR indexer candidates; because they have identical parameter types, current overload resolution treats them as ambiguous instead of applying member-hiding/most-derived declaration precedence.

This breaks both reads and writes on `Microsoft.AspNetCore.Http.IHeaderDictionary`, which redeclares the `IDictionary<string,StringValues>` indexer. It is independent of #2523 and #2516.

## Oahu evidence

Fresh migration from `origin/main` commit `5668dcf4` (#2522), gsc `0.3.166+5668dcf402`:

### Read

```text
sha256:81ebad3c383e661c436539098d0061581a11ed29ae5f9f16fb00c96be795e5e7
corpus_Oahu.Cli.Server/Hosting/HttpEndpoints.gs:79:40
GS0116: Type 'Microsoft.AspNetCore.Http.IHeaderDictionary' is not indexable.
```

```gsharp
let auth = ctx.Request.Headers["Authorization"].ToString()
```

The downstream `StartsWith` failure (`sha256:ee02a3ee8bc3...`) is a cascade from the failed read.

### Write

```text
sha256:fdfb16c78091414e690cd3ebc4891c35afaa56624da2593eead299f0e5e9fa37
corpus_Oahu.Cli.Server/Hosting/HttpEndpoints.gs:90:21
GS0116: Type 'Microsoft.AspNetCore.Http.IHeaderDictionary' is not indexable.
```

```gsharp
ctx.Response.Headers["Retry-After"] = "1"
```

The original C# is valid and uses the ordinary ASP.NET Core header indexer:

```csharp
var auth = ctx.Request.Headers["Authorization"].ToString();
ctx.Response.Headers["Retry-After"] = "1";
```

## Metadata shape

Reflection over the active .NET 10 ASP.NET Core reference reports two public instance indexers visible from `IHeaderDictionary`:

```text
Microsoft.AspNetCore.Http.IHeaderDictionary
  :: StringValues Item[String] get=True set=True

System.Collections.Generic.IDictionary<String,StringValues>
  :: StringValues Item[String] get=True set=True
```

The first is the derived interface's `new` declaration and must hide the inherited same-signature property for binding on an `IHeaderDictionary` receiver.

### Referenced library

```csharp
namespace Lib;

public interface IBase
{
    string this[string key] { get; set; }
}

public interface IDerived : IBase
{
    new string this[string key] { get; set; }
}
```

### Consumer

```csharp
using Lib;

namespace Consumer;

public static class Repro
{
    public static string Read(IDerived value) => value["key"];
    public static void Write(IDerived value) => value["key"] = "value";
}
```

Both projects build successfully in C#. Default `cs2gs migrate --via-sdk` emits:

```gsharp
func Read(value IDerived) string -> value["key"]
func Write(value IDerived) -> value["key"] = "value"
```

and reports:

```text
sha256:dd6f2b64053399493a0b437f3418d3f15799d50c6071334f09e4eb58070f5d66
GS0116: Type 'Lib.IDerived' is not indexable.   // read

sha256:a012fa48834ebf1fc3ad4cc250b9b228f0f9844a844f087509a578213d2af93e
GS0116: Type 'Lib.IDerived' is not indexable.   // write
```

Casting only the receiver to the base interface makes the unchanged pipeline fully green:

```csharp
((IBase)value)["key"]
((IBase)value)["key"] = "value"
```

This proves both accessor slots are individually valid and isolates the defect to candidate hiding/selection on the derived receiver.

## Expected behavior

For a receiver statically typed as the imported derived interface, a same-signature indexer declared on that interface must hide the inherited indexer and be selected for reads and writes. An inherited candidate should remain visible only when no more-derived declaration hides that signature.

## Root-cause direction

`MemberLookup.TryResolveClrIndexer` currently:

1. walks `clrTarget` followed by every inherited interface via `EnumerateSelfAndInterfaces`;
2. collects every public instance indexed property;
3. deduplicates only by `(Module, MetadataToken)`; and
4. sends all getter methods to ordinary overload resolution, or compares only parameter conversions in the symbolic path.

The derived and inherited `Item[string]` declarations have different metadata tokens but identical parameter types. Neither path considers declaring-interface specificity/member hiding, so no unique winner is produced and callers surface GS0116.

Apply CLR/C# hiding semantics before or during overload resolution. A likely direction is to rank candidates by distance from the static receiver and suppress inherited properties whose signature is hidden by an equivalent declaration on a more-derived interface. Preserve genuine ambiguity between unrelated interfaces and normal overload selection between distinct index signatures.

The read and write paths share `TryResolveClrIndexer`; ensure accessor availability is evaluated after selecting the correctly hidden property.

## Contributor-ready coverage requirements

- Imported derived interface using `new` to hide an inherited same-signature indexer: read and write.
- Exact `IHeaderDictionary` read/write shapes with `StringValues` and string keys.
- Get-only, set-only, init-like, ref-returning, and read/write indexers.
- Generic base/derived interfaces and closed/symbolic same-compilation type arguments.
- Multiple index parameters, optional parameters, nullable keys/values, and overloaded distinct signatures.
- Multi-level inheritance and diamond graphs.
- Two unrelated base interfaces exposing identical indexers remain genuinely ambiguous where C# requires qualification.
- Derived declaration with a changed return/value type or accessor set follows C# hiding rules.
- Imported classes that hide base-class indexers as a control.
- Base-interface casts continue selecting the base slot; derived receivers select the derived slot.
- Compound assignment, increments, null-conditional reads, collection/indexer initializer writes, and dynamic invocation paths.
- Runtime dispatch, reflection slot identity, C# consumer compilation, and ILVerify.
- Preserve existing Dictionary/List/Span indexer coverage, symbolic indexer overload resolution, and #2471 generic receiver behavior.
- Default `--via-sdk` two-project minimal plus Oahu.Cli.Server header middleware.

DavidObando:
## Summary
- suppress same-signature CLR indexers only when declared by a genuinely more-derived interface or class
- resolve hidden properties before getter/setter availability, preserving unrelated-base ambiguity and distinct overloads
- emit virtual dispatch for ref-returning interface indexers
- cover minimal, generic, diamond, accessor, expression, ASP.NET `IHeaderDictionary`, runtime, reflection, C# interop, ILVerify, and default `--via-sdk` paths

## Rebase and review
- rebased onto `origin/main` at `77e65cee` (#2526) with `MemberLookup.cs` conflict resolved by retaining both constructed-hierarchy projection helpers and indexer-hiding logic
- audited every `TryResolveClrIndexer` read/write caller plus compound, increment, null-conditional, initializer, expression-tree, constrained-receiver, and emit paths
- expanded tests for symbolic erased overloads, repeated generic metadata tokens, mixed hidden overload sets, inherited/ambiguous diamonds, constrained receivers, expression trees, and C# consumer dispatch

## Validation
- `MSBUILDDISABLENODEREUSE=1 dotnet build GSharp.sln -graph`
- issue #2525 Compiler.Tests: 7 passed
- indexer/rebase regressions: 68 passed
- full Compiler.Tests: 3,251 passed
- issue #2525 Cs2Gs.Tests: 1 passed
- full Cs2Gs.Tests with `RunConfiguration.DisableParallelization=true`: 1,650 passed
- rebuilt `Gsharp.NET.Sdk` nupkg and cleared matching NuGet cache before E2E
- GitHub CI: build, cs2gs, e2e, VS Code matrix, and VSIX passed

Fixes #2525

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs
@@ -977,6 +977,12 @@ internal sealed partial class ExpressionBinder
             var idxArgsAnnot = ImmutableArray.Create(BoundIndexArg());
             if (this.memberLookup.TryResolveClrIndexer(target.Type, clrAnnotIdx, idxArgsAnnot, out var idxPropAnnot, out var resolvedIdxArgsAnnot))
             {
+                if (idxPropAnnot.GetGetMethod(nonPublic: false) == null)
+                {
+                    Diagnostics.ReportTypeNotIndexable(targetLocation, target.Type);
+                    return new BoundErrorExpression(null);
+                }
+
                 var elemTypeAnnot = annotIdx.GetTypeArgumentSymbolForClrType(idxPropAnnot.PropertyType);
                 var convertedIdxArgsAnnot = BindClrIndexerArguments(
                     target.Type,
@@ -991,6 +997,12 @@ internal sealed partial class ExpressionBinder
             var idxArgs = ImmutableArray.Create(BoundIndexArg());
             if (this.memberLookup.TryResolveClrIndexer(target.Type, clrTarget, idxArgs, out var idxProp, out var resolvedIdxArgs))
             {
+                if (idxProp.GetGetMethod(nonPublic: false) == null)
+                {
+                    Diagnostics.ReportTypeNotIndexable(targetLocation, target.Type);
+                    return new BoundErrorExpression(null);
+                }
+
                 var elementType = target.Type is ImportedTypeSymbol imported
                     ? MapErasedIndexerElementType(imported, idxProp)
                     : ClrNullability.GetPropertyTypeSymbol(idxProp);
@@ -1633,7 +1645,7 @@ internal sealed partial class ExpressionBinder
             var idxArgsAnnotWr = ImmutableArray.Create(BindIndexValue());
             if (this.memberLookup.TryResolveClrIndexer(targetType, clrAnnotWr, idxArgsAnnotWr, out var idxPropAnnotWr, out var resolvedIdxArgsAnnotWr))
             {
-                if (!idxPropAnnotWr.CanWrite)
+                if (idxPropAnnotWr.GetSetMethod(nonPublic: false) == null)
                 {
                     Diagnostics.ReportTypeNotIndexable(diagnosticLocation, targetType);
                     return new BoundErrorExpression(null);
@@ -1665,7 +1677,7 @@ internal sealed partial class ExpressionBinder
                 // managed pointer. Detect the ref-returning getter and store through
                 // it. A `ReadOnlySpan[T]` getter is `ref readonly T` — writing is a
                 // hard error (GS0226).
-                if (!idxProp.CanWrite)
+                if (idxProp.GetSetMethod(nonPublic: false) == null)
                 {
                     var refGetter = idxProp.GetGetMethod(nonPublic: false);
                     if (refGetter != null && refGetter.ReturnType.IsByRef)
--- a/src/Core/CodeAnalysis/Binding/MemberLookup.cs
+++ b/src/Core/CodeAnalysis/Binding/MemberLookup.cs
@@ -2605,36 +2605,28 @@ internal sealed class MemberLookup
         indexer = null;
         resolvedArguments = default;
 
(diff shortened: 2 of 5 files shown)


DavidObando:
Closing to avoid duplicate work; another agent owns #2525.

PR #2565, opened 2026-07-20
Add Avalonia support across the G# toolchain

DavidObando:
## Summary
- preserve same-compilation user types through imported generic constraints, including `new()` and expanded `params` candidates
- force translated XAML code-behind declarations partial so source generators can contribute their generated parts
- add managed-resource support to `gsc` emission and forward processed SDK resources with names, visibility, and culture filtering
- expose the standard `AfterCompile` hook and narrowly ignore Avalonia XAML compiler ILVerify false positives reproduced by the C# baseline

## Validation
- `GenericMethodUserTypeArgUnderReferencesTests`: 10 passed
- `Adr0145PreservePartialModeTests` and `IlVerifyStageTests`: 14 passed
- managed-resource compiler regression: passed
- `SdkLayoutTests`: 17 passed
- Avalonia demo migration: translate, compile, ILVerify, and test parity all passed; emitted application launched successfully

### DIFF
--- a/src/Sdk/Gsharp.NET.Sdk/build/Gsharp.NET.Core.Sdk.targets
+++ b/src/Sdk/Gsharp.NET.Sdk/build/Gsharp.NET.Core.Sdk.targets
@@ -5,6 +5,9 @@
 
   <PropertyGroup>
     <CoreCompileDependsOn>_ComputeNonExistentFileProperty</CoreCompileDependsOn>
+    <!-- Avalonia's CompileAvaloniaXaml target hooks AfterCompile, a standard
+         Microsoft.Managed.Core target that the G# language SDK does not import. -->
+    <CompileDependsOn>$(CompileDependsOn);AfterCompile</CompileDependsOn>
 
     <!-- Phase 7.7b: hook into the NuGet Pack infrastructure so dotnet pack
          includes the reference assembly under ref/<tfm>/. The runtime assembly
@@ -280,6 +283,8 @@
           Inputs="$(MSBuildAllProjects);
                   @(Compile);
                   @(ReferencePathWithRefAssemblies);
+                  @(_CoreCompileResourceInputs);
+                  @(CompiledLicenseFile);
                   $(IntermediateOutputPath)$(TargetName).CoreCompileInputs.cache"
           Outputs="@(IntermediateAssembly);
                    $(DocumentationFile);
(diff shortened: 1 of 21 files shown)


src/Sdk/Gsharp.NET.Sdk/build/Gsharp.NET.Core.Sdk.targets
```diff
@@ -5,6 +5,9 @@
 
   <PropertyGroup>
     <CoreCompileDependsOn>_ComputeNonExistentFileProperty</CoreCompileDependsOn>
+    <!-- Avalonia's CompileAvaloniaXaml target hooks AfterCompile, a standard
+         Microsoft.Managed.Core target that the G# language SDK does not import. -->
```
DavidObando, line 9: Should a more generalized fix be to import the `Microsoft.Managed.Core` target? What are the downsides to this? I am just thinking if we're doing custom steps rather than a generic problem-class fix.

tools/cs2gs/Cs2Gs.Pipeline/TranslateStage.cs
```diff
@@ -241,7 +241,18 @@ await CSharpProjectLoader.LoadProjectAsync(context.App.ProjectPath, cancellation
             List<string> retainedFilePaths = hasAnalyzerReferences
                 ? currentProject.Documents.Select(d => d.FilePath).ToList()
                 : null;
-            var translator = new CSharpToGSharpTranslator(markMergedTypePartial: hasAnalyzerReferences, retainedFilePaths: retainedFilePaths);
+            IReadOnlyCollection<string> forcedPartialFilePaths = hasAnalyzerReferences
```
DavidObando: Are these autogenerated at compile time by Avalonia in C#? Or are they artifacts that would be checked into the repo? If they are autogenerated artifacts then they shouldn't be included in the cs2gs migration set. These files should be handled later by gsgen instead. Does this make sense?

DavidObando: Yes. The .axaml files are checked-in source inputs; Avalonia-generated .g.cs artifacts are compile-time outputs and CSharpProjectLoader already excludes generated sources. This update removes the filename-based forced-partial workaround entirely. Canonical C# code-behind remains explicitly partial, cs2gs preserves that declaration, and gsgen regenerates the Avalonia partial declaration from .axaml later.

src/Sdk/Gsharp.NET.Sdk/build/Gsharp.NET.Core.Sdk.targets
DavidObando: Importing Microsoft.Managed.Core.targets wholesale would also bring C#/VB-oriented compiler, analyzer, editorconfig, and CoreCompile assumptions into a language SDK that already owns those hooks, risking duplicate/conflicting compile pipelines. This update instead uses the standard CreateManifestResourceNames extension point with CreateCSharpManifestResourceName, preserving normal manifest naming without importing the unrelated managed compiler pipeline.

PR #2790, opened 2026-07-23
Fix async iterator cleanup after early exits

### LINKED ISSUE #2787: gsc: async iterator early exit before using declaration NREs in MoveNext
## RC‑C15‑C — async iterator early‑exit (`goto`) before a `using` declaration → NRE in `MoveNext`
- **Owner / stage:** gsc, async‑iterator state‑machine lowering (disposal / `using` scope on the early‑exit path).
- **NEW vs REOPEN:** **NEW.** No existing issue (searched "async iterator using", "iterator disposal null", "yield break using"; #1623 is iterator‑lowering drift, not this NRE).
- **Minimal deterministic repro:** `minrepro/asynciter_earlyexit.gs` — `async … IAsyncEnumerable[int32]` with `if !File.Exists(path) { goto __exit }` **before** `using let sr = StreamReader(...)`, `__exit:` at the end. On a missing file:
  ```
  System.NullReferenceException  at AsyncIterEarlyExit.Reader.<ReadAllAsync>d__1.MoveNext()
  ```
- **Baseline/current proof:** C# `JsonlHistoryStore.ReadAllAsync` does `if (!File.Exists(path)) { yield break; }` before the `using` and enumerates empty on a missing file (baseline test passes). Directly reproduced against the migrated closure via `harness/` (`EXCEPTION: System.NullReferenceException … <ReadAllAsync>d__74.MoveNext()`).
- **Affected tests (1):** `Commands.HistoryRetryCommandTests.Retry_UnknownId_ExitsOne` (retry reads `new JsonlHistoryStore(HistoryPath()).ReadAllAsync` with a missing history file → NRE instead of "no history record").
- **Shared code path:** async‑iterator MoveNext/Dispose emission where a `goto` to the exit label bypasses a `using` variable's initialization.
- **Issue‑ready title:** `gsc: async iterator with an early goto before a using declaration NREs in MoveNext on the early-exit path`
- **Fix direction:** ensure the async‑iterator state machine's fault/finally disposal for a `using` only runs when the `using` variable has been assigned (guard on the try‑region entry state), matching C# `yield break` semantics on an early exit before the `using`.
- **Acceptance tests:** `asynciter_earlyexit.gs` prints `count=0` (no NRE); `Retry_UnknownId_ExitsOne` passes; harness enumerates missing‑file `ReadAllAsync` with `count=0`.

DavidObando:
## Summary
- guard using cleanup with explicit initialization state
- preserve protected-region cleanup across sync/async iterator suspension and disposal
- await async-iterator cleanup completion and verify disposal order, early exits, and IL

## Validation
- targeted iterator/using suites: 78 passed
- Compiler.Tests: 3,466 passed
- Core.Tests: 6,701 passed, 1 intentional manual skip
- Oahu JsonlHistoryStore missing-file harness: count=0
- Oahu Retry_UnknownId_ExitsOne: passed

Fixes #2787


DavidObando:
Addressed the deferred ordinary-method case in fde2767f. Protected-region entry routing now runs centrally in `Lowerer` for ordinary sync/async methods and generated state machines. Added runtime + ILVerify regressions for both ordinary method forms; full Compiler.Tests (3,468) and Core.Tests (6,701 + one intentional manual skip) pass locally.

DavidObando:
Second-review gaps addressed in 3357a3c9: using/await-using cleanup now clears initialization state before disposal, preventing repeat cleanup after legal region re-entry; function-wide gotos entering catch/finally handlers now fail with GS0497 while same-handler branches remain legal. Added runtime + ILVerify leave/re-enter coverage and catch/finally diagnostic controls. Full Compiler.Tests: 3,473 passed. Full Core.Tests: 6,701 passed (one intentional manual skip).

Issue #2796, opened 2026-07-23
gsc: null-conditional field-like event raise NREs with no subscribers


DavidObando:
A null-conditional raise of a field-like event throws NullReferenceException when the event has no subscribers instead of being a no-op. This blocks 3 migrated Oahu tests and regressed structural event raises in PR #2792.

Minimal repro:

    package EvtRaise
    import System
    class Bell {
        event Rang EventHandler
        func Ring() { Rang?(this, EventArgs.Empty) }
    }
    let b = Bell()
    b.Ring()

The fresh Bell should return normally, but current main throws NullReferenceException. A subscribed control invokes correctly. A structural event variant using (object?, EventArgs) -> void fails identically.

Regression proof:
- f9c0cd2e cycle-15 tip: structural variant passes
- 2621fa0a PR #2791 merge: structural passes
- c84ca771 PR #2792 merge: structural fails
- 7208c121 current cycle-16 tip: nominal and structural fail

Affected migrated tests all fail at Oahu.Core.SettingsBase.OnChange:
- App.PerJobExportSettingsTests.Override_Wins_Over_Inner
- App.PerJobExportSettingsTests.Null_Overrides_Delegate_To_Inner
- App.PerJobDownloadSettingsTests.ChangedSettings_Subscription_Forwards_To_Inner

Restore short-circuit semantics for both nominal and structural delegate-typed field-like events: when the backing delegate is null, Evt?(args) must no-op. Add runtime and ILVerify coverage for zero-subscriber and subscribed controls, and prove the 3 migrated tests pass.

Cycle-16 evidence: /Users/davidobando/GitHub/DavidObando/oahu-mig-c16/CYCLE16_ISSUE_DETAIL.md

Issue #2772, opened 2026-07-23
gsc: null-conditional invocation of a null delegate throws NullReferenceException


DavidObando:
## RC-C14-4 — gsc: null-conditional delegate invocation `f?()` on a NULL delegate throws NRE instead of yielding null

- **Owner / stage:** gsc / runtime (null-conditional invoke codegen). **Pre-existing**; newly exposed.
- **Symptom:** `d?() ?? fallback` where the delegate `d` is null throws `NullReferenceException` instead of short-circuiting to null.
- **Minimal repro** (`minrepro/nullcond_delegate.gs`, 9 lines):
  ```
  class Opts { prop ActivityVerb (() -> string)? { get; init; } }
  let verb = Opts().ActivityVerb?() ?? "idle"   // NRE (ActivityVerb is null)
  ```
- **Affected tests (3):** `AppShell.BuildHeader` → `options.ActivityVerb?()` on the default (null) delegate — `KeyReader_EOF_Returns_Cleanly`, `Run_Honours_Shift_Q_As_Clean_Quit`, `Run_Returns_Success_When_Ctrl_C_Exits_Idle_Shell`. (Verified the token/Theme static chain is *not* the cause: `Tokens.TextPrimary.Value.ToMarkup()` returns `white` fine standalone.)
- **Baseline proof:** repro NREs on cycle-13 gsc too.
- **Issue-ready title:** `gsc: null-conditional invocation f?() of a null delegate throws NullReferenceException instead of returning null`
- **Acceptance tests:** `nullDelegate?()` yields null (and `?? fb` yields `fb`); non-null delegate invokes.
- **GitHub:** **runtime residual of CLOSED #1410** ("null-conditional invoke H?(args) of a nullable-delegate PROPERTY → GS0131" — the compile error is fixed but it now NREs at runtime). Recommend **NEW** referencing #1410.
- **Priority:** P1.

PR #2799, opened 2026-07-24
Fix value-returning field-like event null-conditional invocation NRE

### LINKED ISSUE #2798: gsc: value-returning field-like event null invocation NREs with no subscribers
A null-conditional invocation of a subscriber-less field-like event with a value-returning delegate throws NullReferenceException instead of producing a null/optional result. PR #2797 fixed the void EventHandler case, but both nominal CLR and structural value-returning paths remain unconditional.

Nominal CLR delegate repro:

    package MyLib
    import System
    class Poll {
        event Ask Func[int32,int32]
        func Run() { var r = Ask?(5) }
    }
    Poll().Run()

Structural delegate repro:

    package MyLib
    class Poll {
        event Ask (int32) -> int32
        func Run() { var r = Ask?(5) }
    }
    Poll().Run()

Both throw NullReferenceException when Ask has no subscribers. A non-null subscriber control should invoke once and return its value.

Root ownership:
- OverloadResolver.CallBinding.cs: CLR-delegate null-conditional block guards only void results.
- OverloadResolver.Invocations.cs: BuildIndirectDelegateCall guards only void FunctionTypeSymbol results.

Mirror the established nullable-delegate invocation result-slot lowering in both paths so a null receiver produces the correct optional result, evaluates the receiver once, and never invokes. Cover nominal Func, structural function delegates, reference- and value-returning shapes, subscribed and zero-subscriber cases, runtime metadata, and ILVerify. Verify natural fallback consumption where supported; if value-type null coalescing exposes a separate emitter defect, isolate and track it rather than hiding it.

Related: #2772 fixed nullable delegate properties, while this residual specifically affects non-nullable field-like events whose CLR backing delegate is null before subscription.

DavidObando:
## Summary
A null-conditional invocation `Evt?(args)` of a **subscriber-less field-like event whose delegate returns a value** threw `NullReferenceException` instead of producing a null/optional result. PR #2797 (issue #2796) fixed the **void-returning** raise on both the nominal CLR-delegate path and the structural `FunctionTypeSymbol` path, but both guarded only the `void` result and emitted an unconditional `callvirt Invoke` for value-returning delegates — dereferencing a null backing delegate on a fresh event with no subscribers.

### Repros (both threw NRE with no subscribers)
```
package MyLib
import System
class Poll {
    event Ask Func[int32,int32]        // nominal CLR delegate
    func Run() { var r = Ask?(5) }
}
```
```
package MyLib
class Poll {
    event Ask (int32) -> int32         // structural function delegate
    func Run() { var r = Ask?(5) }
}
```

### Root cause
- `OverloadResolver.CallBinding.cs`: the nominal CLR-delegate null-conditional block guarded only `void` results, falling through to a plain `callvirt Invoke` for value-returning delegates.
- `OverloadResolver.Invocations.cs` (`BuildIndirectDelegateCall`): guarded only `void` `FunctionTypeSymbol` results.

### Fix
Generalize the **established nullable-delegate invocation result-slot lowering** (the same lowering `TryBindNullableDelegateInvocation` uses for nullable delegate properties, #2772) to **both** paths. A value-returning conditional invocation is wrapped in a `BoundNullConditionalAccessExpression` whose result type is the nullable wrapping of the delegate's return type, allocating a `$nres_` result slot for value-typed returns. A null receiver yields `null`/`default`, the receiver is evaluated once, and a non-null delegate invokes normally. This is a **general** binder/emitter fix with **no Oahu-specific logic**.

Emitted lowering now matches C# `Evt?.Invoke(args)`: value-type returns materialize `default(Nullable<T>)` on the null branch and wrap `T` on the not-null branch; reference-type returns leave `ldnull`.

## Tests
New `Issue2798NullConditionalValueEventRaiseEmitTests` (runtime + ILVerify via the shared `IlVerifier.Verify` helper):
- nominal `Func<int,int>` value-returning: zero-subscriber → null (`?? -1` ⇒ -1); subscribed → invokes (10)
- structural `(int32) -> int32` value-returning: zero-subscriber → null fallback; subscribed → invokes
- nominal `Func<string,string>` and structural `(string) -> string` reference-returning: zero-subscriber → null (`?? "none"`); subscribed → invokes
- bare result (no `??`): `r == nil` true with no subscribers, false once subscribed
- unsubscribe-then-call: value-returning raise yields the null fallback again (the `SettingsBase.OnChange` failure mode, value-returning variant)
- side-effectful argument: evaluated **exactly once** on the non-null path and **not at all** when the receiver is null (short-circuit)

## Validation
- `Compiler.Tests`: **3504 passed, 0 failed**
- `Core.Tests`: **6704 passed, 1 skipped, 0 failed**
- All 5 repro shapes (nominal/structural × value/reference, plus bare) compile, run correctly, and pass ILVerify clean
- Value-type null coalescing (`Nullable<int> ?? -1`) over the null-conditional result verifies and runs clean, so **no separate emitter defect** is exposed by the value-type fallback
- All suites run with `MSBUILDDISABLENODEREUSE=1`

Fixes #2798

### DIFF
--- a/src/Core/CodeAnalysis/Binding/OverloadResolver.CallBinding.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolver.CallBinding.cs
@@ -1030,9 +1030,32 @@ internal sealed partial class OverloadResolver
             && delegateTargetType.ClrType is System.Type delegateClrType
             && ClrTypeUtilities.IsDelegateType(delegateClrType))
         {
-            var receiver = delegateVar is ImplicitFieldVariableSymbol clrImplicitField
-                ? BuildImplicitFieldLoad(clrImplicitField)
-                : (BoundExpression)new BoundVariableExpression(null, delegateVar);
+            // Issue #2799 follow-up: the receiver load must honor every
+            // implicit-member variable kind, not only a plain local/parameter
+            // (a bare `BoundVariableExpression`) and the field-like event's
+            // implicit backing field. A nominal CLR delegate-typed value
+            // exposed as a bare static field or instance/static property
+            // (`ImplicitStaticFieldVariableSymbol` /
+            // `ImplicitPropertyVariableSymbol` /
+            // `ImplicitStaticPropertyVariableSymbol`) has no local slot, so a
+            // bare `BoundVariableExpression` ICE-d with GS9998 ("no local slot")
+            // on any invocation — conditional or not. `TryBuildImplicitMemberLoad`
+            // resolves each such symbol to the correct field/property access
+            // (the same helper the structural `FunctionTypeSymbol` path uses),
+            // leaving locals/parameters to the bare-variable fallback.
+            BoundExpression receiver;
+            if (delegateVar is ImplicitFieldVariableSymbol clrImplicitField)
+            {
+                receiver = BuildImplicitFieldLoad(clrImplicitField);
+            }
+            else if (TryBuildImplicitMemberLoad(delegateVar, syntax.Identifier.Location, out var clrMemberLoad))
+            {
+                receiver = clrMemberLoad;
+            }
+            else
+            {
+                receiver = new BoundVariableExpression(null, delegateVar);
+            }
 
             // Issue #2796 / #2798: a null-conditional raise `Evt?(args)` of a
             // CLR delegate-typed value — canonically a field-like event's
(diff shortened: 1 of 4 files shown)


DavidObando:
## Follow-up (review): generalize null-conditional structural delegate invocation across all receiver kinds

The final review found a tightly-related deferred gap, now completed on this branch.

### Gap
`BuildIndirectDelegateCall` only guarded the null-conditional invocation form `f?(args)` of a non-nullable **structural** function-typed receiver when the variable was the implicit backing field of a field-like event (`ImplicitFieldVariableSymbol`). Every other receiver kind reaching that path fell through to an **unguarded** `BoundIndirectCallExpression`, silently dropping the `?`.

### Proof (before code change)
- `var r = f?(5)` where `f` is a `(int32) -> int32` **local**/**parameter**: `r` bound as non-optional `int32`, so `r == nil` failed with **GS0129** and value-type `r ?? -1` hit the value-type `??` emitter limitation — proving the `?` was dropped (result should be `int32?`).
- A genuinely-null non-nullable local (copy of an uninitialized function-typed field, CLR default null) invoked via the unguarded path would NRE instead of short-circuiting.

### Fix
- **`OverloadResolver.Invocations.cs`** (`BuildIndirectDelegateCall`): compute the receiver load + static type uniformly for **every** variable kind, then apply the null-conditional guard across all of them — receiver evaluated exactly once into a capture; value-returning invocations adopt the established result-slot lowering (nullable result type, `$nres_` slot for value-type returns); void invocations no-op. Shared construction extracted into `BuildNullConditionalDelegateResult` (reused by the CLR path).
- **`OverloadResolver.CallBinding.cs`** (nominal CLR-delegate path): fixed a **distinct pre-existing defect** discovered during this work — the receiver was a bare `BoundVariableExpression` and never used `TryBuildImplicitMemberLoad`, so a nominal `Func<...>` value exposed as a bare static field or instance/static property **ICE-d with GS9998 ("no local slot") on any invocation, conditional or not**. The receiver-load fix mirrors the structural path.

All seven `VariableSymbol` kinds that can reach these paths are now handled — `ImplicitFieldVariableSymbol`, `ImplicitStaticFieldVariableSymbol`, `ImplicitPropertyVariableSymbol`, `ImplicitStaticPropertyVariableSymbol`, `LocalVariableSymbol`, `ParameterSymbol`, `GlobalVariableSymbol` — so **no reachable delegate receiver kind silently drops `?` or ICEs**. A non-nullable function **parameter** cannot be assigned `nil` by language rules (GS0155/GS0129), so its null branch is not literal-reachable; the `?` is still honored (optional result) and non-null invokes. No Oahu-specific logic.

### Tests
New `Issue2799NullConditionalStructuralDelegateReceiverEmitTests` (runtime + ILVerify): local, parameter, instance property, static field, static property, and top-level global receivers; value- and void-returning shapes; null short-circuit + non-null invoke where null is language-reachable; side-effectful property receiver **single-evaluation** (getter runs exactly once on both branches); bare `r == nil` optional-result semantics; and the nominal CLR `Func` property/static-field/static-property receivers (conditional + non-conditional).

### Validation
- `Compiler.Tests`: **3515 passed, 0 failed** (+11 new)
- `Core.Tests`: **6704 passed, 1 skipped, 0 failed**
- All shape assemblies pass ILVerify clean; all suites run with `MSBUILDDISABLENODEREUSE=1`
- **Zero deferred work.**

Issue #988, opened 2026-06-23
No way to construct a type parameter under a new() constraint (new T() reports GS0130/GS0125)


DavidObando:
Discovered by the **cs2gs** C#→G# migration pipeline (#914), L5 corpus. Verified with gsc `0.2.124+6cef33811b`.

A C# generic that constructs its type parameter (`where T : new()`, `new T()`) has no G# spelling: `T()` → `GS0130`, `new T()` → `GS0125`, `T{}` → `GS0157`.

### Repro (GS0130)
```gs
package T
class Factory[T new()] {
    func Make() T { return T() }
}
```

### Control (caller supplies the instance)
```gs
package T
class Box[T class](Value T) { }
```

**Classification:** missing language feature (`new()` constraint + construct-the-type-parameter). Ref ADR-0115 §G #TBD-4.

PR #2903, opened 2026-07-30
Fix explicit generic lambdas across project references

### LINKED ISSUE #2857: Explicit generic call with a lambda fails across a project reference
Tested: Gsharp.NET.Sdk v0.3.319
Also reproduced with gsc origin/main commit c4815a36378af6c3b9467770e0c5a6cee3225595.

A generic static method builds successfully in a library:

```gs
open class Base {
  var Value int32

  shared {
    func Make[T Base init()](configure ((T) -> void)?) T {
      let value = T()
      if let apply = configure {
        apply(value)
      }
      return value
    }
  }
}
```

Calling it from a referenced G# project with an explicit type argument and lambda fails:

```gs
class Derived : FindingCrossAssembly.Base {
}

let value = FindingCrossAssembly.Base.Make[Derived](
  (item Derived) -> { item.Value = 5 })
```

Expected: the app builds and `value.Value` is `5`.

Actual: the library builds with no errors, but the app fails with:

```text
GS0159: Cannot find function Make.
```

Removing `[Derived]` and allowing type inference works. Keeping `[Derived]` and passing `nil` also works.

DavidObando:
## Summary
- Preserve transitive user-class delegate parameters by erasing them to their nearest imported CLR base.
- Re-close explicit imported generic methods only when a generic-dependent parameter is a delegate or a single-level delegate array.
- Leave arbitrary constructed generics such as `List<Action<T>>` outside this fix; those are pre-existing failures on `main`.
- Commit a 15-case cross-assembly false-positive corpus covering delegate arrays and arities 0–4.

## Current anchors
- Head: `3f65a9c619a54eb4cb48ad2562a4f785ae1c194d`
- Base: `0ffd90b1d906ae22377d322fa145c913fed076b4`
- Exact merged tree / `HEAD^{tree}`: `16df40a000a931fa6d558389a851d41ae1de4e2f`
- Diff: 3 files, +491/-1

## Regression matrix

| shape | bare `gsc` evaluate | `gsc /out:` emit oracle | `gsi` |
|---|---|---|---|
| direct delegate (`dbare`) | pre-existing evaluator issue #988 | `11` | pre-existing evaluator issue #988 |
| delegate array (`arr`) | pre-existing evaluator issue #988 | `77` | pre-existing evaluator issue #988 |
| depth-3 delegate array (`d3arr`) | pre-existing evaluator issue #988 | `99` | pre-existing evaluator issue #988 |

Cross-assembly `gsi` is not claimed because it has no `/r:` support. Emit is the oracle.

Focused compiler corpus: 15/15. Delegate arities 0–4 emit `22`, `11`, `33`, `44`, `55`. Each consumer emits into a new asserted-empty directory, IL-verifies, loads both assemblies in a collectible `AssemblyLoadContext`, calls `GetTypes()`, then executes.

The main-without-PR control is 13/15: only the two direct-delegate pins fail; both delegate-array guards pass.

## Product change
- `MemberLookup.cs` uses the nearest imported CLR base.
- `OverloadResolution/ClrOverloadResolution.cs` unwraps one array element solely for delegate detection before existing symbolic recovery and placeholder closure.

## Product mutation table
Each mutant was asserted to apply exactly once and remained build-green.

| mutant | killed |
|---|---:|
| nearest imported base | 9 |
| invert whole re-closure gate | 13 |
| remove delegate-array unwrapping | 2 (`arraytransitive`, `arraydepth3`) |
| invert generic-dependence guard | 11 |
| invert delegate-only guard | 13 |
| discard recovered closure | 10 |
| disable symbolic recovery | 10 |

Final restored control: 15/15.

## Validation
- Current Release solution build, warnings as errors: 0 warnings, 0 errors.
- Current focused cross-assembly corpus: 15/15.
- Current unfiltered `Interpreter.Tests`: 1393/1393.
- Last complete nine-project solution run: Compiler 4260/4260; Core 7128 passed/1 skipped; LanguageServer 462/462; Interpreter 1340/1340; remaining non-cs2gs projects green.
- `Cs2Gs.Tests`: 1904/1907 locally; the same three failures reproduced without this PR on `main` (`Issue3085`, `Issue3090`, `Issue3096`).
- Post-push required GitHub checks validate the exact current merged tree.

Fixes #2857


DavidObando:
## Verdict: NOT MERGEABLE — but do not close it, the bug is real and this fixes it

Reviewed at head `f219e2ccf`. `main` moved during this review, so both measurements are named: `a4a660f32` at the start, `f820484c3` at the end.

Scope computed from the merge base (`git merge-base` = `df4d6e1e6`), never two-dot: **23 files, +3004/-231**, matching `gh pr view` exactly — no phantom scope.

### First question: is the bug still alive? Yes.

Given this is the oldest PR in the queue with zero prior review, the first thing worth knowing is whether #2857 was silently fixed by one of the ~50 merges since. It was not. Built `gsc` from clean `main` `a4a660f32` and ran a genuine two-assembly cross-project repro:

```
app.gs(7,39,8,40): error GS0159: Cannot find function Make.
```

Repro fidelity confirmed with the issue's own two controls — the type-inference form compiles rc=0, and the `[Derived](nil)` form compiles rc=0, exactly as #2857 documents. So the failure is specific to the explicit-generic-lambda cross-assembly path, as claimed.

### Second question: does this branch fix it? Yes.

The PR body says "Local `dotnet build`, `dotnet test`, and restore not run per task constraint", so I verified from scratch. Head builds **rc=0, 0 warnings / 0 errors**. The same repro against head compiles rc=0 and executes, printing `5` — the issue's stated Expected value. Its 47 new tests pass **47/47** (`Issue2857ExplicitGenericLambdaProjectReferenceEmitTests`, `Issue2903SymbolicMethodSpecRegressionTests`), 1m08s.

Scope check on `Fixes #2857` in both directions: the diff adds **0** new diagnostic descriptors (positive control: the same pattern counts **377** in `DiagnosticDescriptors.cs` on `main`, so the pattern is not silently matching nothing). No ID allocation, no docs-gate exposure.

So this is not a close-it candidate. It solves a live bug that nothing else has solved.

### BLOCKER — it does not merge, and the conflict is semantic

```
git merge-tree --write-tree a4a660f32 f219e2ccf  -> rc=1
git merge-tree --write-tree f820484c3 f219e2ccf  -> rc=1
CONFLICT (content): Merge conflict in src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
```

**48 commits behind** current `main`. Three conflicting hunks, around `ExpressionBinder.Calls.Invocation.cs:695-743`, `:818-835`, `:848-877` — and every one of them sits in lambda-target resolution, which is precisely the machinery three already-merged PRs have rewritten since this branch's merge base:

- the `main` side carries #2929's `TryGetFunctionLiteral` / `TryBuildSymbolicDelegateTarget` / `CreateErasedFunctionLiteralAdapter` block ("Fix inline lambdas through erased generic receivers");
- it carries #2918's "substitute the complete parameter before asking whether it is a lambda target";
- hunk 3 is `main` replacing the old `invoke.ReturnType` shape-building with `TryGetLambdaTargetFunctionTypeFromSymbol` plus a `System.Action`/`System.Func` FullName check, landed by #3016.

This branch rewrote the same region a different way, adding `openMethod, methodArgsForCall` parameters to `MapOpenClrTypeToSymbolic`. Because the branch predates #2918/#2929/#3016, its head does **not** contain those fixes — so a "take ours" resolution silently regresses all three, and a "take theirs" resolution silently drops this PR's fix. This is not a textual conflict that a merge tool can settle; someone has to decide whether this rewrite subsumes that work.

Churn on the files this PR touches, since its merge base: `ExpressionBinder.Calls.Invocation.cs` 3 commits, `ExpressionBinder.Access.MemberLookup.cs` 4, `MethodBodyEmitter.cs` 4. The two largest production hunks here — `MemberLookup.cs` **+409/-6** and `OverloadResolution.cs` **+288/-74** — merge cleanly only because `main` has not touched them, which is luck, not compatibility.

### What I did not measure, and why

Being explicit rather than implying more coverage than I have:

- **No regression result.** I started a full `Compiler.Tests` run at head; it exceeded 1020s without completing and I stopped it. A run that does not complete is not a result, so I am reporting none. The 47/47 above covers only the PR's own new tests.
- **No per-hunk mutation testing.** I deliberately skipped it. The conflicted region will be rewritten by the rebase, so mutating this tree would measure a state that cannot land. Mutation is worth doing on the rebased tree, not this one.
- **#2918's repro not replicated.** `test/Compiler.Tests/Emit/Issue2918InlineLambdaErasedReceiverTests.cs` needs a Roslyn-compiled C# contracts assembly (`Slot<T>`, `GenericSink`, `Holder<T>`, `PredHolder<T>`, `DelegateBox<T>`). I judged a hand-rolled approximation more likely to mislead than to inform, so I am flagging the gap instead of reporting a weak probe.

### SHOULD-FIX

`test/Compiler.Tests/.../Issue2903SymbolicMethodSpecRegressionTests.cs` (512 lines) is named for this PR rather than for an issue, which reads as regressions the change introduced and then repaired in-branch. That is fine as history, but after the rebase those are exactly the cases most likely to break again against the #2929/#3016 code — worth confirming they still fail-then-pass on the rebased tree rather than assuming they carry over.

### Smallest thing that makes this mergeable

Rebase onto current `main` (`f820484c3`) and reconcile the three `ExpressionBinder.Calls.Invocation.cs` hunks by hand, explicitly deciding for each whether this PR's rewrite subsumes #2918/#2929/#3016 or must compose with them. Then re-run: the 47 new tests, the #2918 and #2929 regression tests, and a full `Compiler.Tests` pass. At that point this is worth a full mutation review — the underlying fix is demonstrably correct, it is just 48 commits out of date in the worst possible file.

DavidObando:
Takeover complete. Content, not graph: old 23-file / ~3,000-line symbolic-overload rewrite is gone. Current PR is 3 files, +347/-1; production is one `MemberLookup` line plus 14 `OverloadResolution` lines. No `ExpressionBinder.Calls.Invocation.cs` diff.

Conflict decisions:
1. Static symbolic-call hunk: **composes** — dropped branch rewrite; retained main's shared argument pipeline and #2929 adapter unchanged.
2. Helper-overload hunk: **composes** — dropped obsolete branch helper; retained #2918 complete-parameter substitution unchanged.
3. Lambda-target hunk: **composes** — dropped direct `Invoke` reconstruction; retained `TryGetLambdaTargetFunctionTypeFromSymbol` and #3016 canonical delegate guard unchanged.

Scope evidence: running former 47 cases on clean main and initial patch gave 19 pass / 28 fail both ways. Twenty-five failures were common, therefore unrelated existing defects; initial patch fixed 3 cases and regressed 3. Focused rewrite keeps issue pin plus five guards. Landed #2918/#2929/#3016 set passes 20/20 with focused tests.

Mutation on landable head `2d772e7ab` (clean Core hash `839bdc510478a0bac8463dcb99fd05d4`):
- Revert delegate nearest-base hunk → transitive `22` case fails; mutant `0a2e2d...`.
- Revert overload hunk → direct `11` + transitive `22` fail; `437814...`.
- Remove generic-dependence guard → unrelated delegate `55/66` fails; `6dd3ee...`.
- Remove delegate-only guard → `List<T>` `33/44` + unrelated delegate fail; `4afadc...`.
- Discard reclosed method → `11/22` fail; `6ada60...`.
- Disable symbolic recovery → `11/22` fail; `9996ef...`.

Every mutant built rc=0. Every restore rebuilt to clean hash. Latest-main control: direct `11` pin fails, other 5 guards pass. Release build: 0 warnings/errors. Full `Compiler.Tests`: **4019/4019**. CI green. Main advanced after push to `5a27b5532`; no changed-file overlap, `git merge-tree` rc=0.

DavidObando:
## NOT MERGEABLE

Numbers true of `main` **`33bea4af`**, head **`aa8e34005`**, computed merge base **`5ad0109c`**, merged tree **`48a0f9e9`** (`git merge-tree` identical both directions, rc 0). Two-dot == three-dot here; clean branch.

### Lead answer: the coverage IS genuinely cross-assembly. Credit.

The stated worry does not hold. `Issue2857ExplicitGenericLambdaProjectReferenceEmitTests.cs:207-208` emits the library with `/out: /target:library`, `:220-222` compiles the consumer with `/target:exe /reference:`, `:225` runs `IlVerifier.Verify(..., additionalReferences:)`, and `CompileAndRun` `:226-260` executes the result in a bounded 30 s child process asserting `ExitCode == 0` plus stdout. Load-and-run, not verify-only. The interpreter file emits to a `MemoryStream`, `Assembly.Load`s it and binds the consumer through `ReferenceResolver.Default()`. Both halves traverse the imported path for real.

---

## BLOCKER — hunk A regresses correct code that compiles at base; nothing in the repo catches it

`src/Core/CodeAnalysis/Binding/MemberLookup.cs:5344` replaces `userClass.ImportedBaseType?.ClrType` with `TypeMemberModel.GetNearestImportedBase(userClass)?.ClrType`. That is a **global** change to the user-class→CLR erasure switch, not a change scoped to the new explicit-generic path, so it perturbs every consumer of that switch.

Probe `f8`: generic method whose parameter is an **array of delegates** rather than a bare delegate, called with an explicit type argument over a transitive user chain.

```gs
// library i2857l4
func MakeH[T Base init()](handlers []((T) -> void)) T { ... }
// consumer
open class Middle : Base {}
class Derived : Middle {}
let v = Base.MakeH[Derived]([]((Derived) -> void){ (item Derived) -> { item.Value = 77 } })
```

| tree | result |
|---|---|
| base `33bea4af` | rc 0, prints `77` |
| **merged (this PR)** | **compile fails — `GS0159: Cannot find function MakeH`** |

Attributed to hunk A by product mutation, not by reasoning. Reverting only line 5344 to the base expression (hunk B left intact) makes `f8` pass again:

| shape | base (¬A¬B) | M1 = hunk B only | merged (A+B) |
|---|---|---|---|
| `direct` (the #2857 bug) | FAIL `GS0159` | rc 0 `11` | rc 0 `11` |
| `transitive` | rc 0 `22` | FAIL `GS0159` | rc 0 `22` |
| `f8` array-of-delegate | rc 0 `77` | rc 0 `77` | **FAIL `GS0159`** |

Build gates: `BUILD_RC=0` each time; clean md5 `d55ded0d88e65dfcc4cbbec41684e2e6`, M1 md5 `2f0edfd3dd80b1d30da982501d9a05ef` (reproduced identically on two independent builds), exact round-trip back to `d55ded0d` after restore, `git status` clean. mtime verified equal and fresh on `out/bin/Release/{Core,Compiler,Repl}/GSharp.Core.dll`. Every probe compiled into its own empty directory.

**This regression is uncovered**: the 12 new tests are 12/12 green and `Core.Tests` is **7072 passed / 0 failed / 1 skipped** on the merged tree. A full green suite does not see it.

**What the table actually says about the design.** Hunk B alone fixes the reported defect (`direct` passes under M1). Hunk B alone *also breaks* `transitive`, which worked at base. Hunk A was added to repair hunk B's own collateral damage — and overshoots into `f8`. So the PR is a fix plus a compensating global change, and the compensation is where the new breakage lives.

**Recommended smaller fix.** Do not touch the shared erasure switch. Apply nearest-imported-base erasure only inside the new re-close path — `TryCloseOverUserReferenceTypePlaceholders` (`OverloadResolution.cs:4442`) or its new caller at `OverloadResolution.cs:2323-2336` — leaving `MemberLookup.cs:5344` at `ImportedBaseType`. That repairs `transitive` without perturbing the other consumers of the switch, and deletes a global behaviour change from the diff.

---

## SHOULD-FIX — 10 of the 12 new tests pass on unfixed production code

Ran the PR's own test files against **base** production code (`33bea4af`, files copied in, tree otherwise clean):

Emit — **6 passed / 2 failed of 8**. Only these two fail at base:
- `..._Runs(packageName: "i2857topdirect", useIntermediateBase: False, useFunction: False, expected: 11)`
- `..._Runs(packageName: "i2857funcdirect", useIntermediateBase: False, useFunction: True, expected: 33)`

Passing at base, therefore pinning nothing about #2857: both `useIntermediateBase: True` rows (22, 44), both `WithListOfDerived` rows (33, 44), `WithUnrelatedDelegateParameter`, and `TransitiveSlice_ToInvariantInterface_IsRejected` (`:186`).

Interpreter — **4 passed / 4**, all green at base. The whole 98-line file is decorative with respect to the defect.

Fair qualification: the two transitive rows are not worthless — the shape they use dies under M1, so they guard hunk A. But they guard *hunk B's regression*, not issue #2857, and the PR presents them as coverage of the reported bug. Please relabel, and add a row for the one thing that is genuinely unprotected: a delegate nested inside another type (`f8`).

## SHOULD-FIX — every anchor in the PR body is stale

Body claims head `2d772e7ab`, merge base `f5e9d8925`, "3 files, +347/−1". Actual: head `aa8e34005`, merge base `5ad0109c`, **4 files, +450/−1** (production only **+15/−1** across 2 files). The second commit `aa8e3400` postdates the body. Restate against `33bea4af`.

## NIT — the #3069 / #3068 interaction question is already closed

Both are **ancestors**, not parallel work. Verified by content in the merge base `5ad0109c`, not by date: `LambdaEnclosingTypeParameterCollector` `IsGeneric != true` guard present, `ClosureEmitter` collector call present, `MetadataTokenCache` `MethodSpec` handling present. No co-merge needed.

---

## Verified correct — no action

- **Explicit type arguments are honoured, not silently replaced by inference.** Negative control: with `Make[Mid]` explicit and `class Der : Mid { var Extra int32 }`, reading `v.Extra` is still rejected on the merged tree (`GS0158`), while the positive twin prints `99`. Had the fix fallen back to the inferred set, `Extra` would have resolved.
- **Derived-only members survive the re-close**: explicit `Make[Derived]` then `value.Extra` / `value.Bump()` prints `11|33|700`.
- **No false positives** in three further shapes, identical base vs merged: interface-constrained `T` (`64`), depth-3 user chain (`66`), explicit argument equal to the imported base itself (`88`).
- **The guard does not over-fire**: a generic method with no delegate parameter is byte-for-byte unperturbed (`55` both trees).

## Disclosure

Two probes of mine were malformed and were fixed and re-run, not reported: an early harness named the library DLL something other than its package, yielding spurious `FileNotFoundException` rc134 on *both* trees; and `lib3.gs` initially omitted the `;` no-body marker on an interface method (`GS0368`, ADR-0085). `GS9100` appears in every cross-assembly probe here and is noise. `Compiler.Tests` was run filtered to `Issue2857` only — I did not run the full ~1h20m suite.

DavidObando:
## Addendum — head moved to `3e236438`; verdict UNCHANGED, re-measured not transferred

My review above names head **`aa8e34005`**, which no longer exists on this branch. Re-anchoring.

### The two new commits are a pure rebase — zero content change

| | old | new |
|---|---|---|
| head | `aa8e34005` | **`3e236438`** |
| merge base | `5ad0109c` | **`33bea4af`** (branch now sits on current `main`) |
| commits | `2b12215d`, `aa8e3400` | `5a3de573`, `3e236438` — **identical subjects** |

Diff content compared by md5 (three-dot, `--full-index`, `index` lines stripped):

- production `src/` diff: `5642200f5427eee00112d5ba465cd0a6` — **identical both heads**
- test `test/` diff: `eae0157faec8b1785e828269c1c5d589` — **identical both heads**
- diffstat unchanged: 4 files, **+450/−1**

Despite the commit subject *"Test #2857 across scopes and engines"*, **no test was added or changed** by the rebase; the test diff is byte-identical to the previous head.

### Why no re-measurement was needed for the tree, and why I re-measured anyway

`git rev-parse 3e236438^{tree}` = **`48a0f9e98cd6ba9c2e1a453c440930ba11bf5c0e`** — bit-identical to the merged tree I originally built, mutated and measured. `git merge-tree --write-tree origin/main aa8e34005` reproduces the same tree hash. So every number in my review applies verbatim.

I still re-fired the BLOCKER against freshly built current SHAs rather than relying on tree identity:

| tree | libcc | concc | runrc | output |
|---|---|---|---|---|
| `main` `33bea4af` | 0 | 0 | 0 | **`77`** |
| head `3e236438` | 0 | **1** | — | **`GS0159: Cannot find function MakeH`** |

`BUILD_RC=0` both trees; `Core/`, `Compiler/`, `Repl/` mtimes fresh and equal (`ref/` excluded); each probe compiled into its own **empty** directory; library assembly named for its package.

**The BLOCKER is live on the current head.** A correct program that compiles on `main` stops compiling under this PR — attributed to hunk A (`src/Core/CodeAnalysis/Binding/MemberLookup.cs:5344`) by product mutation. Recommendation is unchanged: scope the nearest-imported-base erasure into `TryCloseOverUserReferenceTypePlaceholders` (`OverloadResolution.cs:4442`) or its new caller (`:2323-2336`) instead of changing the global switch.

### Interaction with #3068 / #3069 — now proven behaviourally, not by file lists

Both are ancestors of the merge base (verified by content, not graph). Their suites run green **on this PR's tree**:

- #3069 `Issue2894` — Compiler.Tests **7/7**, Interpreter.Tests **2/2**
- #3068 `Issue3065` — Core.Tests **3/3**

Non-vacuous in every case (filters matched >0 tests).

### Unchanged from the review

- Coverage **is** genuinely cross-assembly (`/target:library` + `/reference:` + bounded child run). Credit stands.
- **SHOULD-FIX** — 10 of 12 new tests still pass on unfixed production code.
- **SHOULD-FIX** — PR body anchors remain stale: body says *3 files, +347/−1*; actual is **4 files, +450/−1**, head `3e236438`, base `33bea4af`.

**NOT MERGEABLE**, unchanged.

*Filtered green declared: `Compiler.Tests` run only on the `Issue2894`/`Issue2857` filters, not the full suite.*

DavidObando:
## Round 3 — head `69f86f15`, `main` `4469e323` — **NOT MERGEABLE** (1 BLOCKER, 2 SHOULD-FIX, 1 NIT)

The prior BLOCKER is **live, unaddressed, and wider than previously measured**. The two commits pushed since the last review add 33 lines of interpreter tests and change no production code.

Anchors: merge base = `4469e323` = current `main` (0 behind / 4 ahead), so the head *is* the merged tree. `git merge-tree 4469e323 pr2903h` emits **1 line, 0 conflict markers** — I read the output, not the rc. `BUILD_RC=0` on base and head, 0 warnings / 0 errors; `Core/`, `Compiler/`, `Repl/` mtimes fresh on both (`ref/` excluded). Every probe compiled into its own asserted-empty directory; harness positive-controlled by feeding it a deliberately broken library, which it correctly reported as `LIBFAIL` with `GS0288`/`GS0113` rather than silently continuing.

### The two new commits are test-only — production is byte-identical to the blocked head

| | previously blocked `3e236438` | current `69f86f15` |
|---|---|---|
| production `src/` diff, `+`/`-` lines only | `751f21ef7b50d55d01822d37885f61dd` | **`751f21ef7b50d55d01822d37885f61dd`** |
| diffstat | 4 files, +450/−1 | 4 files, **+483/−1** |

The only delta is `Issue2857...InterpreterTests.cs`, 98 → 131 lines. `MemberLookup.cs` is still `2 +-` and `OverloadResolution.cs` still `+14`. The hunk-A line moved from `:5344` to **`:5348`** by rebase offset only.

---

## BLOCKER — `src/Core/CodeAnalysis/Binding/MemberLookup.cs:5348` still rejects correct code, and it is not one shape

Re-fired at current anchors, not transferred from the prior review. Cross-assembly (`/target:library` + `/reference:`), consumer compiled and the emitted program executed.

| probe | shape | `main` `4469e323` | head `69f86f15` |
|---|---|---|---|
| `dbare` | direct chain, bare delegate — **the #2857 bug** | FAIL `GS0159` | **rc 0, `11`** — fix works |
| `arr` | depth-1 transitive, **array of delegates** | **rc 0, `77`** | **FAIL `GS0159`** |
| `d3arr` | depth-3 transitive, **array of delegates** | **rc 0, `99`** | **FAIL `GS0159`** |

```
con.gs(8,26,8,104): error GS0159: Cannot find function MakeH.
        let v = i2857f8.Base.MakeH[Derived]([]((Derived) -> void){ (item Derived) -> { item.Value = 77 } })
```

The prior review found one such shape. It generalises across chain depth: **every array-of-delegates shape I built breaks, at depth 1 and depth 3.** Net effect on correct programs is **+1 fixed, −2 broken**.

### False-positive corpus — 15 cross-assembly programs, base vs head

Axes: `{explicit type argument}` × `{arity 0,1,2,3,4}` × `{bare, nullable, two-parameter, returning-T, array-of, List-of}` × `{direct, depth-1 transitive, depth-3 transitive}`.

|---|---|---|---|---|
| `a0` | arity 0, `() -> T` | `22` | `22` | ok |
| `a1` | arity 1, bare, transitive | `11` | `11` | ok |
| `a2` `a3` `a4` | arity 2 / 3 / 4 | `33` `44` `55` | same | ok |
| `nul` | `((T) -> void)?` | `44` | `44` | ok |
| `two` | two delegate parameters | `55` | `55` | ok |
| `ret` | delegate returning `T` | `66` | `66` | ok |
| `d3` | depth-3, bare delegate | `88` | `88` | ok |
| `dbare` | **direct, bare** | FAIL | **`11`** | **fixed** |
| **`arr`** | **transitive, array-of-delegates** | **`77`** | **FAIL** | **regression** |
| **`d3arr`** | **depth-3, array-of-delegates** | **`99`** | **FAIL** | **regression** |
| `darr` | direct, array-of-delegates | FAIL | FAIL | unchanged |
| `lst` `dlst` | `List[((T) -> void)]` | FAIL | FAIL | unchanged, pre-existing |

`lst`/`dlst` fail on both trees, so generic-collection-of-delegates is a pre-existing limitation, not this PR's — flagging so it is not chased.

### Correcting the prior review's recommended remedy — the naive revert is much worse

The previous recommendation was to revert `MemberLookup.cs:5344` and scope the nearest-imported-base erasure into `TryCloseOverUserReferenceTypePlaceholders`. Direction is right, but the **revert half of that must not be done on its own.** Product mutant M1 (hunk A reverted to `userClass.ImportedBaseType?.ClrType`, hunk B intact; asserted applied — 1 new pattern, 0 old remaining, hunk B's 6 references still present; `BUILD_RC=0`; exact round-trip afterwards, `git status` clean, head behaviour reproduced on rebuild):

| M1 result | probes |
|---|---|
| pass | `dbare`, `arr`, `d3arr` |
| **FAIL, though correct on base** | `a0`, `a1`, `a2`, `a3`, `a4`, `nul`, `two`, `ret`, `d3` |

**M1 breaks 9 of 13.** So hunk A is load-bearing for the entire transitive-delegate surface, and hunk B alone is a far larger source break than the PR as it stands. The prior table showed M1 failing only `transitive`; at every arity it fails. The fix must be a *scoped re-implementation* — `OverloadResolution.cs:2328` or `TryCloseOverUserReferenceTypePlaceholders` itself — validated against the array-of-delegates shape, which neither tree's tests cover.

### The regression is invisible to every gate

`Core.Tests` on the head: **7089 passed / 0 failed / 1 skipped / 7090**. The PR's own `Issue2857` emit tests: **8/8**. Nothing sees it.

---

## SHOULD-FIX — 10 of 12 tests still pass on unfixed production, and the 33 new lines are entirely in that set

PR test files copied onto **base** production (`4469e323`, `git diff -- src/` asserted empty, files otherwise untouched):

- Emit — **6 passed / 2 failed of 8**. Only `i2857topdirect` and `i2857funcdirect` (both `useIntermediateBase: False`) fail, matching `dbare` exactly.
- Interpreter — **4 passed / 4 failed 0**. Filter asserted non-vacuous (Total: 4, not 0).

So the whole 131-line interpreter file passes with and without the fix, and the two newest commits — *"Exercise #2857 through gsi"* and *"Adapt gsi coverage to script output contract"* — landed 33 lines inside the set that pins nothing. Please add a row for the genuinely unprotected shape: **array-of-delegates over a transitive chain**, which is the case this PR breaks.

## SHOULD-FIX — PR body anchors are still stale, third review running

Body says head `aa8e34005`, base `main` `5ad0109c0`, *"4-file, +450/−1"*, `Compiler.Tests` 4053/4053, `Interpreter.Tests` 611/611. Actual: head **`69f86f15`**, base **`4469e323`**, **+483/−1**. Please restate.

## NIT — the interpreter test exercises a configuration no user can create

`Issue2857...InterpreterTests.cs:87` does `Assembly.Load(peStream.ToArray())` into the test process and then calls `GSharp.Repl.Program.Main`, so the consumer resolves the library out of the already-loaded AppDomain. `gsi` has no `/r:` equivalent, so this path is unreachable from a command line. It is not wrong, but it should not be read as cross-assembly `gsi` coverage — there is no such column.

---

## #3019 overlap — **not the same root cause**, determined behaviourally

Zero file overlap (`MemberLookup.cs` / `OverloadResolution.cs` versus `ExpressionBinder.Calls.Invocation.cs` / `ExpressionBinder.Calls.cs`), but per standing practice that is only a hypothesis, so I measured.

**#3019 alone**, merged at `4469e323`, `BUILD_RC=0`: `dbare` **still FAILS** — it does not fix #2857; `arr`, `d3arr`, `a1` all pass — it does not cause the regression. `IsCanonicalFunctionDelegate` is therefore not the #2857 root cause.

**Co-merged #2903 + #3019** at `4469e323`: `MERGE_RC=0` both merges, **`BUILD_RC=0`, 0 warnings / 0 errors**. Corpus on the co-merged tree is identical to #2903 alone — `dbare` `11`, `a1` `11`, `a0` `22`, `nul` `44`, and **`arr` / `d3arr` still fail**. #3019 does not rescue the regression.

They are independent. Neither should be held for the other, and #2903's regression is its own to fix.

## Driver columns

Cross-assembly is **`gsc`-only** — `gsi` has no reference option, so there is no cross-assembly interpreter column and I report none. Same-assembly single file, both trees:

| shape | `gsc /out:` emit | bare `gsc` evaluate | `gsi` |
|---|---|---|---|
| `sa_bare` | rc 0, runs `11` | `GS9999` | `GS9999` |
| `sa_arr` | rc 0, runs `77` | `GS9999` | `GS9999` |

**Identical on base and head.** The `GS9999` is documented and pre-existing — `Construction of type parameter 'T()' is only supported by the compiled backend (issue #988)` — not a finding against this PR. Note the delta that matters: `sa_arr` **emits and runs correctly same-assembly on the head**, and only breaks across a project reference, which is consistent with the imported-type erasure switch being the cause.

## Disclosed as not run

Full `Compiler.Tests` (~1h20m) on the head — I ran it filtered to `Issue2857` only and declare that filtered. `Cs2Gs.Tests` not run. e2e not run (prohibited). One build completed while 1-minute load was 55; it is a deterministic `BUILD_RC`, and I re-verified the same tree by exact behavioural round-trip afterwards, but I am naming it rather than hiding it. I deferred a later build for 1 minute at load 38 rather than measure through it.

DavidObando:
## Rubber-duck review — PR #2903 (`fix/2857-explicit-generic-lambda`), issue #2857

**Verdict: MERGEABLE.** 0 BLOCKER · 1 SHOULD-FIX · 3 NIT.

### Anchoring (read this first — the brief I was given was stale)

I was briefed head `2e1e0f9f`. That object exists but is **not an ancestor of the current head** — the branch was force-pushed. Actual head is **`664e90e6`**, 5 commits, +491/−1 over 3 files, updated `2026-08-03T20:59:13Z`. The new top commit is `664e90e6 "Handle delegate arrays in explicit generic closure"`, which is what closes the `arr`/`d3arr` breakage reported against the earlier revision.

Reviewed against `main` `52d1511a`; merge base `61fe9cb9` (8 behind / 5 ahead). `main` then advanced to `9ab9124d` mid-review — I re-checked: still merges clean (tree `062884cf`, zero conflict text) and **none of the 8 new commits touch either file this PR edits**, so the measurement holds.

`git merge-tree` output (not rc) at `52d1511a`: single line, tree `a051ba18`, no conflict text. Built the real merged tree, reproduced tree identity `a051ba18` exactly. **`BUILD_RC=0`, 0 warnings, 0 errors.** Base `52d1511a` also `BUILD_RC=0`.

---

## 1. The bug is still live on current `main` — this PR is not dead code

Forty-odd merges have not incidentally fixed it. Reproduced the issue verbatim, cross-assembly via `/r:`:

| variant | BASE `52d1511a` emit | MERGED emit | issue says |
|---|---|---|---|
| **`Make[Derived](lambda)`** | **`GS0159: Cannot find function Make.`** | **`5`** | expected `5` |
| `Make(lambda)` (inference) | `5` | `5` | works |
| `Make[Derived](nil)` | `0` | `0` | works |

The defect reproduces, the fix resolves it, and the two forms the issue calls already-working are unchanged.

### A `GS9999` that is *not* this PR's fault

On the merged tree the evaluate driver reports `GS9999: Cannot invoke a method on objects loaded by a MetadataLoadContext` for the repro. That looks alarming until you run the controls: **the same `GS9999` appears on BASE for the two variants the issue says already work.** It is a pre-existing limitation of bare-`gsc` evaluation of cross-assembly generic calls, not a regression. Had I not run the known-good controls I would have filed it as a BLOCKER.

**Driver-column note:** `gsi` has no `/r:` (confirmed from `--help`), and bare `gsc` hits the pre-existing `GS9999`/`#988` wall cross-assembly. So for cross-assembly shapes **emit is not merely the oracle, it is the only live column** — a fact the PR body states correctly. Same-assembly shapes do give three live columns, and I used those for the regression corpus below.

## 2. Root cause, and both hunks justified

Two production hunks, and they map to two distinct sub-cases rather than one fix plus padding:

- **`MemberLookup.cs:5377`** — `userClass.ImportedBaseType?.ClrType` → `TypeMemberModel.GetNearestImportedBase(userClass)?.ClrType`. `ImportedBaseType` is null when a class derives from another *G#* class, so `Derived : Middle : Base` erased to `object`. This is the **transitive-chain** sub-case.
- **`OverloadResolution.cs:2321+`** — after `MakeGenericMethod` *succeeds*, re-close over user-reference placeholders when a generic-dependent parameter is a delegate. Pre-PR, `TryCloseOverUserReferenceTypePlaceholders` was only reached in the **failure** path (`else if` at `:2337`), so a succeeding-but-wrong close was never corrected.

Both helpers **pre-exist** (`GetNearestImportedBase` has 14 call sites; `TryCloseOverUserReferenceTypePlaceholders` has 4 pre-existing call sites). The PR reuses established machinery and invents none. `MemberLookup:5377` was an outlier that the PR brings into line with the dominant idiom — a genuine one-line root-cause fix.

My mutation results confirm the two hunks are not redundant: reverting the `MemberLookup` hunk kills **exactly** the transitive cases while `topdirect`/`funcdirect` survive.

## 3. Mutation testing — PRODUCT, condition inversion, applied asserted via `git diff` + md5, restored by md5 round-trip

| mutant | site | BUILD_RC | killed |
|---|---|---:|---:|
| **M1** — invert re-closure gate | `OverloadResolution.cs` new guard | 0 | **13 / 15** |
| **M2** — revert to `ImportedBaseType` | `MemberLookup.cs:5377` | 0 | **9 / 15** |
| **C1 (control)** — invert `!SatisfiesGenericConstraints` in the **pre-existing** helper | `OverloadResolution.cs:~4398` | 0 | **10 / 690** (`~Generic` filter, 10 m 38 s) |

Both hunks independently pinned. The control fires in *pre-existing* tests, proving the wider suite — not just the new file — detects changes in this code, so a survival would have meant something.

M2's survivors are precisely `topdirect` and `funcdirect`; M1's survivors are the negative test and the unrelated-delegate test. Both survivor sets are semantically correct, not gaps.

**Independent corroboration of the PR body:** my M1 (13) and M2 (9) reproduce the author's own table rows "invert whole re-closure gate | 13" and "nearest imported base | 9" exactly, measured independently on a different base.

No retargets were needed this round; no mutant failed to build.

## 4. False-positive corpus — the real risk direction

Widening placeholder re-closure means more candidates surviving, so the failure mode is **new errors on code that compiles today**. Corpus built to compile on BASE and required to still compile on MERGED. All three drivers where live:

| shape | BASE (eval / emit / gsi) | MERGED (eval / emit / gsi) |
|---|---|---|
| explicit generic + lambda + **overload set** (delegate overload *and* `int32` overload) | `#988` / `1 2` / `#988` | `#988` / `1 2` / `#988` |
| explicit generic, 3-deep **pure G# chain, no imported base** | `#988` / `3` / `#988` | `#988` / `3` / `#988` |
| explicit generic + lambda + delegate array + `nil`, no `T()` — **all three drivers live** | `4 6 6` / `4 6 6` / `4 6 6` | `4 6 6` / `4 6 6` / `4 6 6` |

**No new ambiguity, no source break, byte-identical output on every row.** The overload-set row is the one I most expected to break; it did not. The PR also ships its own negative test (`TransitiveSlice_ToInvariantInterfaceAcrossReference_IsRejected`), so the suite is not positive-only.

## 5. Test assertions assert values

`Assert.Equal($"{expected}\n", CompileAndRun(library, consumer, packageName))` — exact printed output, with distinct per-case expectations (11/22/33/44/77/99/55…), not rc and not "no diagnostic". Coverage is stated rather than implied: direct + transitive, `Func` and non-`Func`, delegate arrays at inheritance depth 1 and 3, **arity 0–4**, `List`-of-derived, an unrelated-delegate parameter, and one rejection test.

---

## SHOULD-FIX — jagged delegate arrays still fail with the *identical* `GS0159`

`OverloadResolution.cs` (new hunk) unwraps exactly **one** array level:

```csharp
p.ParameterType.IsArray ? p.ParameterType.GetElementType() : p.ParameterType
```

For `[][]((T) -> void)` that yields `((T) -> void)[]`, which is not a delegate, so the guard is false and the re-closure never runs.

| probe | BASE | MERGED |
|---|---|---|
| `func Make[T Base init()](configure [][]((T) -> void)) T`, called `Make[Derived](...)` cross-assembly | `GS0159` | **`GS0159` — still fails** |

**Syntax positive-controlled**: the same jagged declaration compiles fine same-assembly (`Wrote s.dll`), so this is absence, not syntax.

The PR body says the fix applies when a parameter "is a delegate **or a delegate array**", and explicitly scopes out "arbitrary constructed generics such as `List<Action<T>>`". I verified that disclosure is accurate — `List[(T) -> void]` fails `GS0159` on **both** trees, exactly as claimed. But `[][]delegate` reads as inside the stated boundary while being outside the actual one. **Stated scope and actual scope differ**, and the symptom is the very `GS0159` this PR exists to remove.

**I verified the fix rather than proposing it.** Replacing the ternary with a loop:

```csharp
private static Type UnwrapArrayElement(Type type)
{
    while (type is not null && type.IsArray)
    {
        type = type.GetElementType();
    }

    return type;
}
```

Result: `BUILD_RC=0`, 0 warnings; jagged probe emits and prints **`7`**; original repro still prints **`5`**; **Issue2857 suite 15/15 still green**. Restored to pristine (md5 round-trip) and rebuilt afterwards.

Deferring is defensible given the shape is marginal — but then the summary line should say *single-level* delegate array so the boundary is not overstated.

---

## NITs

1. **PR body numbers are stale, not wrong.** Body claims `Compiler.Tests` 4247/4247 and `Interpreter.Tests` 1220/1220; I measure **4260** and **1264** on the merged tree. Both moved *up*, consistent with `main` advancing `61fe9cb9` → `52d1511a`. Likewise the body's merged tree `e9ff923d` is against the old base; against current `main` it is `a051ba18`. Worth re-stating the anchor when the numbers are quoted.
2. **`TryCloseOverUserReferenceTypePlaceholders` hand-rolls the very walk the PR just migrated away from** (`OverloadResolution.cs:~4411`, `for (var current = symbol; …) importedBase = current.ImportedBaseType?.ClrType;`) — a duplicate of `GetNearestImportedBase`. Pre-existing, not introduced here.
3. **Six other `ImportedBaseType?.ClrType ?? typeof(object)` sites remain unmigrated** (`MemberLookup.cs:1987`, `ExpressionBinder.cs:1511`/`:1528`, `ImportedClassSymbol.cs:273`/`:456`, `ExpressionBinder.Access.MemberLookup.cs:427`). **I did not verify any of these are actually wrong** — the erasure target may legitimately differ per site. Flagging as an unverified structural observation only, for a possible follow-up issue; expanding this PR to cover them would be scope creep.

---

## Test tally

Denominator from `dotnet sln GSharp.sln list` = **9 projects** (CI's own source per `build/generate-ci-test-matrix.py:26`). All run on the merged tree; counts from runs, not `--list-tests`.

| project | result |
|---|---|
| `Compiler.Tests` | 4260 / 0 failed (1 h 25 m, unfiltered) |
| `Core.Tests` | 7127 / 0 failed / 1 skipped (7128) |
| `Cs2Gs.Tests` | 1907 / 0 failed (14 m 54 s) |
| `Interpreter.Tests` | 1264 / 0 failed |
| `LanguageServer.Tests` | 462 / 0 failed |
| `Extensions.Tests` | 104 / 0 failed |
| `Sdk.Tests` | 59 / 0 failed |
| `GeneratorHost.Tests` | 31 / 0 failed |
| `InternalAnalyzers.Tests` | 7 / 0 failed |

**9 of 9 run. 15,221 passed, 0 failed, 1 skipped. Nothing NOT RUN.**

## Self-disclosed errors in my own instrumentation

1. **I nearly filed a false stale-build anomaly by choosing the wrong freshness witness.** Two sources were newer than `gsc.dll`. But `gsc.dll` is a thin launcher; the changed code lives in `GSharp.Core.dll`, which is timestamped **16:18:51**, five seconds *after* the source restore at 16:18:46. Fresh. Comparing against the launcher, not the assembly under review, would have produced a confident and entirely fictional finding.
2. **Harness bug that mimicked a product failure.** My first cross-assembly run emitted the library as `lib.dll` while its assembly name was `FindingCrossAssembly`, producing `FileNotFoundException: Could not load file or assembly 'FindingCrossAssembly'` at runtime — easily misread as the PR emitting a broken reference. Fixed to name the output after the assembly.
3. **The briefed head was stale** (`2e1e0f9f`, force-pushed away). Re-anchored to `664e90e6` before measuring; reviewing the briefed SHA would have missed the delegate-array commit entirely.

I also reproduced the standing BSD `find` correction as a positive control: `-newermt '@1'` returned **0** where `-newermt '1970-01-02'` returned **603**. Epoch form is silently useless here; all freshness numbers above use the date-string form.

## Hygiene

Both worktrees porcelain 0; every mutant asserted applied via `git diff` and md5 movement, restored by md5 round-trip and **rebuilt** after each restore; merged tree identity `a051ba18` reproduced from `merge-tree`. Every probe ran in its own directory asserted empty. `rc` captured before any pipe (only the benign `GS9100` transitive-reference warning is filtered, and for display only). Library builds positive-controlled — a missing `lib.dll` aborts the probe rather than yielding a uniform rc=1. Shared checkout untouched and porcelain 0.

---

**MERGEABLE.** The bug is real and still live on current `main`, the fix is two small hunks reusing pre-existing helpers, both are independently pinned by value-asserting tests, the regression corpus shows no source break in the false-positive direction, and 9 of 9 suites are green. The SHOULD-FIX is a one-shape completeness gap with a verified six-line patch, and can ship as a follow-up if the summary's scope wording is tightened.

PR #2908, opened 2026-07-30
Reject nullable delegate receivers without valid narrowing

### LINKED ISSUE #2885: gsc: nullable CLR-delegate receiver invoked without narrowing emits invalid IL or silently null-dereferences
Found while reviewing #2849 / PR #2884. This is separate from #2849: in these cases no valid smart-cast narrowing is in force.

## Summary

Direct call syntax accepts a nullable imported CLR-delegate receiver even when the receiver is unstable or its earlier narrowing has been invalidated. The compiler emits the call instead of reporting a nullability diagnostic. With a same-compilation generic argument, the call can also target the erased `Action<object>.Invoke`; with a receiver reset to `nil`, runtime silently reaches a `NullReferenceException`.

Verified on current `main` and PR #2884 HEAD. The affected cases emit byte-identical bad IL on both, so this is **not a regression** from #2884.

## Minimal invalidated-local repro

```gsharp
package NullableDelegateLocal
import System

class Src { prop N int32 -> 2 }

func Main() {
    let w System.Action[Src] = (s Src) -> System.Console.WriteLine(s.N)
    var n System.Action[Src]? = w
    if n != nil {
        n = nil
        n(Src())
    }
}
```

The assignment invalidates the guard's narrowing, but `n(Src())` compiles with no diagnostic. The emitted program null-dereferences at runtime.

## Other affected receiver shapes

The same direct nullable-delegate call is accepted for receivers that are intentionally not smart-cast-stable:

```gsharp
class Holder {
    var Mutable System.Action[Src]?

    prop Custom System.Action[Src]? -> Mutable

    shared {
        var SharedVar System.Action[Src]?
        let SharedLet System.Action[Src]? = nil
    }

    func Run() {
        if Mutable != nil { Mutable(Src()) }
        if Custom != nil { Custom(Src()) }
        if SharedVar != nil { SharedVar(Src()) }
        if SharedLet != nil { SharedLet(Src()) }
    }
}
```

Affected matrix confirmed during #2884 review:

- mutable instance field
- custom-getter property
- `shared var`
- `shared let`
- local whose prior narrowing is invalidated by assignment

For a source-declared generic argument such as `Src`, affected calls can fail `ilverify` with `StackUnexpected`, comparing the real `Action<Src>` receiver against erased `Action<object>.Invoke`.

## Observed

- compilation succeeds with no nullability diagnostic
- unstable member cases emit invalid IL for source-declared generic arguments
- invalidated-local case can throw `NullReferenceException` at runtime

## Expected

A direct call on a nullable delegate must require a currently valid narrowing/non-null value. Once narrowing is unavailable or invalidated, report a nullability diagnostic. Explicit null-safe invocation or a stable non-null capture should remain the supported alternatives.

## Scope / root-cause lead

The callable branches accept the receiver from its declared nullable CLR delegate shape even when flow analysis did not provide a narrowed type. This must not be fixed by treating unstable members as smart-cast-stable; their instability is deliberate. The call binder should reject the nullable receiver when no valid narrowed type is present.

- diagnostic assertions for all five receiver shapes above
- invalidated-local runtime control proving no silent null dereference is emitted
- stable `let` field/property and non-null local controls that still compile, verify, and run
- source-declared and BCL-only delegate type arguments

DavidObando:
Closes #2885.

## Root cause

Call binding accepted a receiver from its declared nullable delegate shape even
when flow analysis supplied no valid narrowed type. Loop bodies also skipped the
narrowing frame used by `if`, so the new guard rejected legal calls such as:

```gsharp
while d != nil {
    d(1)
}
```

## Fix

- reject nullable delegates across bare, qualified, value-callee, and explicit
  `.Invoke(...)` paths when no valid narrowing exists
- preserve imported CLR stable-member narrowing through access-path identity
- classify imported readonly fields and get-only auto-properties as stable
- bind `while`, condition-only `for`, and clause `for` bodies under their
  condition's full narrowing (`nil` guards, type tests, and `[NotNullWhen]`)
- retain `do ... while` first-iteration semantics and reject unstable members
- report `GS0503` with syntax-valid remedies and exact receiver spans

## BLOCKER 3 proof

- original base `d4038f1d`: imported `Action<int32>?` guarded `while` compiled,
  `Assembly.Load` + `GetTypes()` succeeded, and execution printed `x:1`
- HEAD `1345c180`: same result
- HEAD compiler variants passed for `while`, condition-only `for`, clause `for`,
  nested loops, stable members, `&&`, source types, and nullable classes
- interpreter guarded nullable-function `while` and clause-`for` tests both
  evaluate successfully; existing `if` narrowing remains unchanged
- `do ... while` and loop-guarded unstable members remain `GS0503`

Hand-written base/HEAD matrix: **2 pins, 23 guards**. All 23 guards compile and
run on both; the two pins are the pre-existing general loop-narrowing gaps for a
nullable class and a nullable named delegate. A `while value is T` pin also verifies full condition narrowing.

## Review NITs

- centralized nullable-delegate reporting
- made advice receiver-shape aware: `?(...)` for names/members, `?.Invoke(...)` for indexed/call-result receivers
- moved receiver text extraction inside the `.Invoke` diagnostic path and standardized qualified receivers on member names
- typed public access-path members; source identity remains reference-based, CLR identity uses `MetadataToken` + `Module`
- split CLR stability coverage into two mutation detectors (readonly field, get-only property) and two over-fire guards (`if let`, `!!`)
- fixed both diagnostics tables to put actual code under “Example trigger”
- recorded source-symbol hashing as an equality-consistent, performance-only surviving mutant (constant hash); no correctness gap
- hardened runtime probes with `Assembly.Load` + `GetTypes()`, concurrent pipe
  drains, and whole-process-tree timeout termination

Filed with exact review attribution and `Oahu`:

- #2924
- #2925
- #2943 — known ADR-0069 residual: assignments later in enclosing loop bodies do not invalidate narrowing across back-edges

## Mutation audit

Every control ran first in the same worktree. Every mutation built successfully.

| Semantic branch | Control | Mutated |
|---|---:|---:|
| diagnostic advice | Compiler 0/47 failed | 11/47 failed |
| shared report helper | Compiler 0/47; Core 0/18 | Compiler 23/47; Core 3/18 |
| qualified G# guard | Compiler 0/47 | 8/47 failed |
| qualified CLR guard | Compiler 0/47 | 1/47 failed |
| bare/implicit guard | Compiler 0/47; Core 0/18 | Compiler 13/47; Core 3/18 |
| value-callee guard | Compiler 0/47 | 3/47 failed |
| explicit `.Invoke` guard | Compiler 0/47 | 4/47 failed |
| root receiver syntax | Compiler 0/47 | 3/47 failed |
| nested receiver syntax | Compiler 0/47 | 1/47 failed |
| CLR member narrowing | Compiler 0/47 | 1/47 failed |
| CLR narrowed-read reconstruction | Compiler 0/47 | 1/47 failed |
| CLR access-path identity | Compiler 0/47 | 1/47 failed |
| CLR stable-path construction | Compiler 0/47 | 1/47 failed |
| readonly-field stability | Compiler 0/47 | 1/47 failed |
| get-only-property stability | Compiler 0/47 | 1/47 failed |
| condition-loop frame | Compiler 0/47; Interpreter 0/2 | Compiler 7/47; Interpreter 1/2 |
| clause-loop frame | Compiler 0/47; Interpreter 0/2 | Compiler 2/47; Interpreter 1/2 |
| null-safe advice shape | Compiler 0/1 | Compiler 1/1 for each forced advice branch |
| full loop condition narrowing | Compiler 0/10 | Compiler 1/10 (`while value is string`) |
| CLR member value identity | Core 0/2 | Core 1/2 |

One additional capture-type refactor produced **0 failures** under isolated
mutation. It was semantically inert and was removed rather than claimed.

## Verification

- Release solution build, static graph, warnings-as-errors: **0 warnings,
  0 errors**
- `Core.Tests`: **6,915 passed, 1 skipped, 0 failed**
- `Compiler.Tests`: **3,802 passed, 0 skipped, 0 failed**
- `Interpreter.Tests`: **470 passed, 0 skipped, 0 failed**
- `LanguageServer.Tests`: **452 passed, 0 skipped, 0 failed**
- `Extensions.Tests`: **104 passed, 0 skipped, 0 failed**
- `Sdk.Tests`: **57 passed, 0 skipped, 0 failed**
- base `e21677f6`: Core **6,913 + 1 skipped**, Compiler **3,756**, Interpreter **468**, LanguageServer **452**, Extensions **104**, SDK **57**; branch deltas **+2 / +46 / +2 / 0 / 0 / 0**

No e2e tests were run locally.

Known residual: ADR-0069 does not yet invalidate a narrowing across a loop back-edge when the invalidating assignment appears later in an enclosing loop body. This pre-existing general flow-analysis gap reproduces on `main` and this branch for nullable classes and delegates; tracked by #2943.

## Corpus

Compiled every `samples/**/*.gs` and `e2etests/**/*.gs` file independently with
base-built and HEAD-built compilers:

- files: **148**
- both succeeded: **139**
- both rejected identically: **9**
- exit differences: **0**
- diagnostic differences: **0**
- byte differences: **0**
- files containing a nullable delegate: **0**

Corpus is therefore only a weak over-fire signal for this change; hand-written
nullable-delegate shapes carry the behavioral proof.


DavidObando:
## Review — PR #2908 (`29c99c6f`), issue #2885

# Verdict: **MERGEABLE** — no blocking issues

Core fix correct end to end. Over-fire hunt clean. No dead hunks left. Author's numbers all check out exactly. Four SHOULD-FIX, five NIT below.

---

## Q5 — actual bug, end to end (verified both failure modes)

Harness = real `gsc.dll` → `dotnet test.dll` child process + `ilverify`. Not `IlVerifier.Verify`.

| | `main` d4038f1d | HEAD 29c99c6f |
|---|---|---|
| Issue repro (`if n != nil { n = nil; n(Src()) }`) | compiles, **`NullReferenceException`** rc=134 | `GS0503`, no emit |
| Unstable member + source generic arg (`h.Mutable(Src())`) | compiles, **`ilverify`: `[StackUnexpected] found 'Action`1<Src>' expected 'Action`1<object>'`** | `GS0503`, no emit |
| Guarded positive (`while d != nil { d(1) }`) | compiles + prints `1` | same |

Both #2885 failure modes reproduce on `main`. Note the invalid-IL case **runs fine** — JIT tolerates it. Execution-only proof misses it; `ilverify` catches it. My verdict rests on process execution *and* ilverify, not `IlVerifier.Verify`.

Three-path parity: interpreter shares `Compilation` → same binder → same GS0503 (confirmed by `Issue2442` `Evaluate` tests now reporting GS0503). LSP path: **`LanguageServer.Tests` 452 passed** on HEAD — author never reported this suite. Also `Extensions.Tests` 104, `Sdk.Tests` 57, all green.

---

## Q1 — loop narrowing matrix (`main` vs HEAD)

### Mutation-invalidates-narrowing (the unsoundness risk) — **PASSES**

| case | main | HEAD | ok? |
|---|---|---|---|
| `while d != nil { d = nil; d(1) }` | NRE | `GS0503` | ✅ |
| `while d != nil { d(1); d = nil }` | ok | ok | ✅ sound, cond re-checked |
| `for var i=0; d != nil; i++ { d = nil; d(1) }` | NRE | `GS0503` | ✅ |
| `for d != nil { d = nil; d(1) }` | NRE | `GS0503` | ✅ |
| `for ...; d != nil && i<3; d = nil` (post mutates) | ok | ok | ✅ sound |
| lambda captures + mutates `d`, then `d(1)` | NRE | `GS0503` | ✅ |
| `d = nil` then `continue` then `d(1)` | NRE | `GS0503` | ✅ |
| narrowing leaks past loop end (`while d != nil {break}; d(1)`) | ok | `GS0503` | ✅ frame popped |

Mechanism is sound: `InvalidateNarrowingsForAssignedVariables` (`StatementBinder.Narrowing.cs:104`) walks **all** frames, so the new loop frame is invalidated identically to `if` frames. `finally` pop at `StatementBinder.Loops.cs:652` is correct.

### Residual hole (pre-existing, NOT a regression)

Loop **back-edge** does not invalidate narrowing:
```gsharp
while d != nil { for var i=0; i<2; i++ { d(1); d = nil } break }   // A5
if    d != nil { for var i=0; i<2; i++ { d(1); d = nil } }         // A6
if    c != nil { for var i=0; i<2; i++ { c.M(); c = nil } }        // B7 nullable class
```
All three: **NRE on `main` AND on HEAD, byte-identical**. B7 proves it is a general ADR-0069 engine hole (nullable class, `if` guard), not delegate-specific and not introduced here. See SHOULD-FIX 4.

### Control flow / other narrowing forms

| shape | main | HEAD |
|---|---|---|
| `try`/`finally` inside guarded loop | — | ✅ runs |
| `defer` inside guarded loop | — | ✅ runs |
| `continue` (no mutation) | — | ✅ narrowing preserved |
| nested `if` inside guarded loop | — | ✅ |
| nested guarded `while` | — | ✅ |
| `&&` composed loop condition | — | ✅ |
| `\|\|` loop condition (`d != nil \|\| i<1`) | **NRE** | ✅ `GS0503` |
| `do...while` | compiles, unguarded 1st iter | ✅ `GS0503` — deliberate, tested (`DoWhileCondition_DoesNotNarrowFirstIteration`), documented in PR body |
| `for-in` / range | unchanged | unchanged (no bool condition — correct) |
| **`while c != nil { c.M() }` nullable class** | **`GS0159` compile error** | ✅ compiles + runs |
| **`while s != nil { s.Length }` nullable string** | **`GS0158` compile error** | ✅ compiles + runs |
| **`while o is string { o.Length }` type-test** | `GS0158` | **still `GS0158`** ← asymmetry, see SHOULD-FIX 2 |
| `if o is string { o.Length }` | ✅ | ✅ |

**Correcting your briefing:** you framed the loop change as risky because it is "thinly pinned at 2 pins". Real picture is the opposite of what that implies. On `main` loop bodies narrowed **nothing at all** — so every nullable *reference*/member shape in a loop was a hard compile error (rows 10–11), and every nullable *delegate* shape compiled by accident (that is #2885). The change is a pure **widening**: it cannot silently change existing emit because the affected programs previously did not compile. That is why pin count is inherently 2. Blast radius is real but the failure mode is "more programs compile", and the full suite (6913 / 3749 / 455 / 452) plus my B-matrix is clean. 2 pins is correct here, not thin.

`SmartCastStability` loop-carried classification: captured loop variable → correctly drops (H5, closure filter); field/property guarded in loop condition → mutable member correctly stays unstable (`LoopGuardedUnstableMember_RemainsRejected`), stable `let` field correctly narrows (`stable-member` case).

---

## Q2 — over-fire hunt: **clean, no false positives**

Corpus: **148 `.gs` files, 0 contain a nullable delegate** — confirmed independently by grep. Corpus is worthless as an over-fire signal, exactly as the author says. Handwritten matrix is the evidence.

Every legitimate narrowing/remedy form **suppresses** GS0503 and runs correctly on HEAD:

`if x != nil` · `if x == nil { return }` early exit · `!!(...)` · `!!.Invoke(...)` · `if let` (local / member / indexed / call-result) · `?(...)` · `?.Invoke(...)` · `&&` short-circuit · guard-then-intervening-call-then-use · guarded `.Invoke` · nested `?.` chain · non-nullable delegate · generic `Action[T]?` param · source `let` field · `prop { get; init; }` · event-like `var` field via `if let` · imported CLR `readonly` field · imported CLR get-only auto-property · `for-in` · ternary on the nullable.

Only newly-rejected programs are the intended ones (unguarded/invalidated nullable delegate call). Sanity check that this is *consistency*, not an arbitrary new restriction: `var c C? = C(); c.M()` is **already** an error on `main` (`GS0159`). Delegates were the outlier. GS0503 brings them in line — and with a far better message than `GS0159`.

`.Invoke` interaction: `x(...)` / `x.Invoke(...)` require narrowing; `x?(...)` / `x?.Invoke(...)` do **not**. Correct.

---

## Q3 — per-assertion justification for the two modified test files

### `test/Compiler.Tests/Emit/Issue1410NullableDelegatePropertyEmitTests.cs`

| assertion | before | after | verdict |
|---|---|---|---|
| `…ShortCircuitAndPlainCallStillThrows` → split, renamed `…ShortCircuit` | asserted `"fallback\ncalled\nvalue\nplain-call-threw\n"` — i.e. `empty.Text()` compiles and **throws `NullReferenceException` at runtime**, caught by `try/catch` | asserts `"fallback\ncalled\nvalue\n"`; short-circuit `?(…)` / `?()` behaviour retained verbatim | **legitimate correction.** Old assertion literally pinned #2885's silent null-deref as *expected behaviour*. |
| new `NullableReferenceReturningDelegateProperty_PlainCallReportsNullableReceiver` | (did not exist) | same `empty.Text()` shape → asserts single error `GS0503` + span `Text` | **legitimate.** The removed behaviour was **converted, not deleted** — the same source shape is still covered, now asserting the correct outcome. This is the right way to do it. |

Coverage is net **+1** test, not reduced. No papering over.

### `test/Core.Tests/CodeAnalysis/Binding/Issue2442ClosureCallableNarrowingBinderTests.cs`

| assertion | before | after | verdict |
|---|---|---|---|
| `MutableVarLocal_ReassignedBeforeClosureCreation_DoesNotNarrow` | `AssertContainsError(result, "GS0131")` | `"GS0503"` | **legitimate.** GS0131 = `"'{0}' is not a function."` — a *wrong and confusing* message for a nullable delegate whose narrowing was invalidated. GS0503 says exactly what happened. Test still asserts an error; strictly stronger message, not weaker. |
| `MutableVarLocal_ReassignedAfterClosureCreation_DoesNotNarrow` | `GS0131` | `GS0503` | same |
| `MutableVarLocal_ReassignedAcrossLoopIterations_DoesNotNarrow` | `GS0131` | `GS0503` | same |
| class `<remarks>` rewrite (−8/+2 lines) | documented a "pre-existing quirk": CLR-delegate fallback "binds the call successfully whether or not any narrowing is in scope" | "Both named and imported delegate shapes are covered" | **legitimate.** That documented quirk **is** issue #2885. Leaving the comment would have been stale. |
| comment above positive block (−5/+1) | explained why CLR shapes "never reproduced GS0131" | "remain regression coverage for valid closure narrowing" | **legitimate**, same reason. |
| `RefParameter_DoesNotNarrow`, `BareNullableFunctionType_…`, `ClrGenericDelegate_…`, all `AssertClean` positives | unchanged | unchanged | ✅ untouched |

**No test was inverted, weakened, or deleted.** GS0131 remains live for genuine non-callable callees (`Issue2185IndirectCalleeInvocationEmitTests` still green). Zero net test-count change in Core — consistent with 6913 → 6913.

---

## Q7 — my own mutation table (pins and guards counted separately)

**Controls, unmutated, same worktree, before every batch:** Compiler `Issue2885|Issue1410` **0/47 failed**; Core `Issue2442|DiagnosticIdUniqueness` **0/22**; Interpreter `Issue2885` **0/2**. Every mutation below built green (`0 Warning(s), 0 Error(s)`); two initial formulations produced C# compile errors and were **reformulated as inert** (`H1`, `H2` → dead-`&&` guard), not counted as detections.

| # | branch | file:line | mutation (build-green) | Compiler | Core | Interp |
|---|---|---|---|---:|---:|---:|
| H1 | bare-identifier receiver guard | `OverloadResolver.CallBinding.cs:861` | `&& directDelegateVar.Name == null &&` prepended | **13**/47 | **3**/22 | 0/2 |
| H2 | value-callee guard | `OverloadResolver.Invocations.cs:733` | `calleeName == null &&` prepended | **3**/47 | 0 | 0 |
| H3 | G# member guard | `ExpressionBinder.Calls.Arguments.cs:515` | block deleted | **4**/47 | 0 | 0 |
| H4 | CLR member guard | `…Calls.Arguments.cs:753` | block deleted | **1**/47 | 0 | 0 |
| H5 | CLR member narrowed load | `…Calls.Arguments.cs:749` | `ApplyMemberNarrowing(…)` unwrapped | **1**/47 | 0 | 0 |
| H6 | explicit `.Invoke` guard | `ExpressionBinder.Calls.Invocation.cs:1877` | block deleted | **4**/47 | 0 | 0 |
| H7 | receiver-syntax threading | `…Calls.Invocation.cs:1871` | `receiverSyntax ??` dropped | **4**/47 | 0 | 0 |
| H8 | CLR prop in `ApplyMemberNarrowing` | `ExpressionBinder.cs:809` | case forced to `return node` | **1**/47 | 0 | 0 |
| H9 | SmartCast CLR stable path | `SmartCastStability.cs:102` | `when false &&` | **1**/47 | 0 | 0 |
| H10 | imported readonly-field stability | `SmartCastStability.cs:322` | `false &&` | **1**/47 | 0 | 0 |
| H11 | imported get-only-prop stability | `SmartCastStability.cs:340` | `false &&` | **1**/47 | 0 | 0 |
| H13 | `while` / `for cond` frame | `StatementBinder.Loops.cs:532` | reverted to `BindLoopBody` | **7**/47 | 0 | **1**/2 |
| H14 | `for` clause frame | `StatementBinder.Loops.cs:588` | reverted to `BindLoopBody` | **2**/47 | 0 | **1**/2 |
| H15 | shape-aware advice | `DiagnosticBag.Reports.Expressions.cs:309` | ternary forced to `true` | **1**/47 | 0 | 0 |

**Q4 answer: zero hunks whose individual neutralisation produces zero failures.** The author's claim that they found and removed the one dead hunk holds — nothing else is dead. Credit is earned.

### Pins vs guards, semantic-branch granularity (my own derivation)

**PINS — 15** (wrong on `main`, correct on HEAD): invalidated local · unstable member × source generic arg (invalid IL) · mutable field · custom-getter property · `shared var` · `shared let` · unguarded local/param direct call · unguarded `.Invoke` · indexed receiver · call-result receiver · named G# delegate (`GS0131`→`GS0503`) · `do…while` · `||` loop condition · loop-narrowing: nullable class · loop-narrowing: nullable string/named-delegate.

**GUARDS — 25** (identical on both, must not break): all 20 remedy/narrowing forms from Q2 plus `try/finally`, `defer`, `continue`, `for-in`, tuple-element delegate (#2924 unchanged), interface delegate prop (#2925 unchanged).

The author's "2 pins, 23 guards" is scoped only to the **loop sub-change**, and for that sub-change my count agrees exactly (2 pins = nullable class + nullable named delegate). It is not the whole-PR figure, and the PR body should say so.

### **True `origin/main` (d4038f1d) baselines — measured directly, publish these**

| suite | `origin/main` | HEAD (author) | delta | tests added | ✓ |
|---|---:|---:|---:|---:|---|
| `Core.Tests` | **6913** passed, **1** skipped | 6913 / 1 skipped | **0** | 0 (assertion edits only) | ✅ |
| `Compiler.Tests` | **3708** passed, 0 skipped | 3749 | **+41** | 40 new cases + 1 from the Issue1410 split | ✅ |
| `Interpreter.Tests` | **453** passed | 455 | **+2** | 2 | ✅ |

The four figures circulating in this series (3713, 3721, 3734, 3749) are all *HEAD-of-that-PR* numbers. `origin/main` `Compiler.Tests` is **3708**. Author's delta is exact — I counted `[Fact]`×7 + `[InlineData]`×33 = 40, plus the split = 41.

Also green on HEAD, unreported by author: `LanguageServer.Tests` **452**, `Extensions.Tests` **104**, `Sdk.Tests` **57**.

---

## Q8 — merge order and downstream

`git merge-tree HEAD <pr>`:

| PR | result |
|---|---|
| #2911, #2917, #2923 | **conflict**, all in `src/Core/CodeAnalysis/DiagnosticDescriptors.cs` + `docs/diagnostics.md` + `website/docs/ref/diagnostics.md` — adjacent-line additions of GS0504/GS0506/GS0508. **Mechanical**: keep both descriptors in ID order, keep both doc sections in ordinal order. |
| #2912, #2926, #2929, **#2932**, #2935 | **clean** |

**Correcting your briefing:** `ExpressionBinder.Calls.Invocation.cs` is *not* a conflict hot spot. #2929 rewrites 210 lines of that file and still merges clean — the hunks do not overlap (this PR adds at `:1866`, #2929 works elsewhere). #2932 touches `ExpressionBinder.Access.MemberLookup.cs`, which this PR also edits (+10), and also merges clean.

### #2925 — definitive statement

**#2925 is still a real bug after this lands, and is completely unaffected by it.** Verified on both builds:

```
main d4038f1d : box.P(1)                      → GS0159: Cannot find function P.
HEAD 29c99c6f : box.P(1)                      → GS0159: Cannot find function P.   (identical)
HEAD 29c99c6f : if box.P != nil { box.P(1) }  → GS0159 (nullable variant, identical)
```

This PR neither helps nor hinders and does not partially implement it. **But it creates a requirement for #2925's implementer:** once the `InterfaceSymbol` delegate-property path is added at `ExpressionBinder.Calls.Invocation.cs:2386`/`:2529`, the **nullable** variant will reach delegate-call binding for the first time and must route through `OverloadResolver.TryReportNullableDelegateReceiver` (`OverloadResolver.cs:312`) or it re-opens #2885 for that shape. An interface `prop P { get }` is a computed property → not smart-cast-stable → `if box.P != nil { box.P(1) }` must report GS0503 and the remedy is `if let`. Say this in #2925 before work starts.

**#2924** — verified accurate and unaffected: `t.0(1)` silently elided, byte-identical `before\nafter\n` on `main` and HEAD. Both filings are correct.

---

## SHOULD-FIX

### SF1 — `.Invoke` advice tells the user to write code that does not compile
`src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs:1879` hardcodes `supportsNullSafeInvocation: true`, while the direct-call path derives it from receiver shape. Result:

| source | advice | does the advice compile? |
|---|---|---|
| `arr[0](1)` | "Bind it with `if let`, or assert with `!!`" | ✅ correct |
| `arr[0].Invoke(1)` | "**Use `'?(...)'`** for null-safe invocation…" | ❌ `arr[0]?(1)` → `GS0155: Cannot convert type 'System.Action[int32]?' to 'bool'` |
| `Get()(1)` | `if let` / `!!` | ✅ |
| `Get().Invoke(1)` | "**Use `'?(...)'`**…" | ❌ `Get()?(1)` → `GS0155` |

Same receiver, contradictory advice, and the `.Invoke` variant sends the user into a worse error. This is the exact NIT the author claims resolved ("made `?(...)` advice receiver-shape aware") — the *span/name* was fixed, the *advice flag* was not, and `DiagnosticAdvice_MatchesReceiverSyntax` only tests the direct-call indexed case so it is blind to this.
**Fix:** pass the same shape predicate the direct-call path uses; for indexed/call-result shapes advise **`?.Invoke(...)`**, which I verified works (`arr[0]?.Invoke(1)` and `Get()?.Invoke(1)` both compile and print `1`). Also worth adding to the *direct-call* indexed advice, which currently omits a working remedy.

### SF2 — ponytail: use the helper that already exists (smaller, and fixes a real asymmetry for free)
`src/Core/CodeAnalysis/Binding/StatementBinder.Loops.cs:639`:
```csharp
: TryClassifyNilGuard(condition).NonNil;      // current
: ComputeConditionNarrowing(condition).Then;  // proposed
```
`ComputeConditionNarrowing` already exists at `StatementBinder.Narrowing.cs:636`, and its own doc comment says it exists to "mirror the composition `BindIfStatement` performs (nil guard, `[NotNullWhen]` bool call, and type-test)". Using the nil-guard leaf alone means the PR ships a language rule where `while x != nil` narrows but `while x is T` does not — while `if` narrows both.

I built and measured it:
- **`Core.Tests` 6913 passed / 1 skipped; `Compiler.Tests` 3749 passed; `Interpreter.Tests` 455 passed — full suites, zero failures**, identical to pristine HEAD.
- `while o is string { o.Length }` now compiles and runs, matching `if o is string`.
- All rejection behaviour preserved: `while d != nil { d = nil; d(1) }` → `GS0503`; `do…while` → `GS0503`.

Same line count, one existing helper instead of one narrower one, one less language wart. Ladder rung 2.

### SF3 — ~120 lines of CLR-stability machinery pinned by exactly one test
Your briefing asked me to check the `1/47` rows for incidental coupling. They are worse than incidental — they are **the same test six times**. I ran H9 with per-test detail: the sole failure is `ImportedClrStableDelegateMembers_NarrowAndRun`. It is also the sole detector for H4, H5, H8, H10, H11. That one method guards `AccessPath.cs` (+46, including a public-shape change), `SmartCastStability.cs` (+38), `ExpressionBinder.cs` (+31), and the narrowed load in `Calls.Arguments.cs`. The author's table presents these as six independent rows; they are one.
**Fix:** no new machinery — split that single method into four (`readonly` field, get-only property, `if let`, `!!`), so each stability rule has its own pin. Five minutes, and the six table rows become truthful.

### SF4 — PR body and #2885's "Expected" overclaim
PR says "Once narrowing is unavailable or invalidated, report a nullability diagnostic." Not true for the loop back-edge (A5/A6/B7): `if c != nil { for … { c.M(); c = nil } }` still silently NREs, identically on `main` and HEAD, for nullable classes as well as delegates. Pre-existing, general, out of scope — but it should be stated as a known residual rather than implied fixed, and filed as a follow-up against ADR-0069 (narrowing must be invalidated by any assignment anywhere in an enclosing loop body, not just textually-preceding statements). Otherwise #2885 gets closed with the hole still open and no ticket.

---

## NIT

- **N1** `AccessPath.Equals`/`StartsWith` (`AccessPath.cs:119,144`) compare `MemberInfo` with `ReferenceEquals`, while the codebase's established pattern for CLR member identity is `MetadataToken` + `Module` (`MemberLookup.cs:3953`, `ExpressionBinder.Calls.Invocation.cs:464`). Reflection member instances are not contractually reference-stable. In practice both sides derive from the same `BoundClrPropertyAccessExpression.Member` within one binding pass, so risk is low — but it is inconsistent with the file's own new doc comment ("equal member identities"). `GetHashCode` uses value hashing while `Equals` uses reference — legal (no dictionary corruption), just extra collisions.
- **N2** `ExpressionBinder.Calls.Invocation.cs:1873` computes `receiverName` via `SyntaxTree.Text.ToString(span)` **unconditionally**, on every accessor call in the compilation, then discards it unless `methodName == "Invoke"`. Move it inside the `if`.
- **N3** Receiver span/name is member-only for qualified G# receivers (`h.Mutable(…)` → message says `'Mutable'`) but full text for the `.Invoke` path (`'arr[0]'`, `'Get()'`). Inconsistent spelling of the same concept.
- **N4** Both diagnostics docs put remedy text under a column headed **"Example trigger"**. Every neighbouring GS05xx section uses that column for a trigger.
- **N5** `AccessPath.Members` widened from `ImmutableArray<Symbol>` to `ImmutableArray<object>` on a `public sealed class`. Minimal way to admit `MemberInfo`, but it erases the type contract on a public surface; a small `readonly record struct PathMember` would keep it typed. Only worth it if `Core`'s public API is a shipped contract.

---

## Things I verified and found solid (no action)

- **GS0503 is correct and free.** Commit `54c9c232` renumbered from GS0502, which genuinely collided with `InterfacePropertySetterKindMismatch` (`DiagnosticDescriptors.cs:375`). **Zero stale GS0502 references** from this PR anywhere in `src/`, `test/`, `docs/`, `website/`. The renumber also correctly moved the doc entry out of `website/versioned_docs/version-0.3/` (frozen) into `website/docs/` — that was a second, unclaimed correction.
- Docs: correct ordinal position in **both** files (`docs/diagnostics.md:953`, `website/docs/ref/diagnostics.md:1688`, both immediately after GS0502), trailing newlines intact (`0a` confirmed), `DiagnosticIdUniquenessTests` green (22/22 including doc-coverage assertions).
- Message quality: `"Nullable delegate receiver 'n' may be nil. Use '?(...)' for null-safe invocation, bind it with 'if let', or assert non-null with '!!' before calling."` — actionable, names the receiver, gives three remedies. Strictly better than the `GS0131 "'n' is not a function."` and `GS0159 "Cannot find function P."` it replaces. Every suggested remedy verified to compile and run except the `.Invoke` case in SF1.
- Shared diagnostic helper `OverloadResolver.TryReportNullableDelegateReceiver` (`OverloadResolver.cs:312`) genuinely removes duplication — one predicate + one report, four call sites. Not a layer.
- `do…while` non-narrowing is deliberate, correct (guard runs after body), tested, and stated in the PR body.
- 19 files: all load-bearing. Every one has a non-zero mutation detection. The three `OverloadResolver` additions are three genuinely distinct call-binding entry points (bare identifier, value callee, member) routed through one shared choke point — that is already the DRY shape.
- The deleted inert hunk is real and the deletion is the right instinct. Nothing else is dead.

DavidObando:
## Rubber-duck review — round 3 (Opus 5), head `c47993bd`

**Verdict: MERGEABLE.** No code blockers. Nine of nine landed items check out. One operational item: the branch needs a rebase because `main` moved during the review.

### B1 — operational, not code
`origin/main` moved `60e5b8bf` → `eaf59056` (#2917) mid-review, and has since moved again.
```
$ git merge-tree c47993bd origin/main
CONFLICT (content): src/Core/CodeAnalysis/DiagnosticDescriptors.cs
```
Same two-file class already resolved once. Keep GS0503 + GS0506 + GS0508 in ID order. `docs/diagnostics.md` merges clean this time; only the website copy collides.

### 1. `PathMember.cs` — passes. A real fix, not a nicer wrapper over `ReferenceEquals`.

| check | result |
|---|---|
| `Equals`/`GetHashCode` consistent | **Yes, all three arms.** source: `ReferenceEquals(SourceSymbol)` ↔ `RuntimeHelpers.GetHashCode`. CLR: `MetadataToken == && ReferenceEquals(Module)` ↔ `HashCode.Combine(RuntimeHelpers.GetHashCode(Module), MetadataToken)`. default: `both-null→true` ↔ `0`. The N1 asymmetry is genuinely gone. |
| house pattern used | **Yes** — `:68-69` is `MetadataToken` + `Module`, matching `MemberLookup.cs:3953`. |
| every kind round-trips | **Yes.** Only 3 `Append` sites exist (`SmartCastStability.cs:93/99/105`); the `Symbol` overload takes the first two, `MemberInfo` the third. Nothing coerced or lost. |
| 89 lines bloat? | **No** — ~44 are StyleCop-mandated XML docs, and `operator ==`/`!=` are CA-mandated on a struct overriding `Equals` (omitting them is a build error under warnings-as-errors). It net *deleted* the `AccessPath.ToString` type switch. |
| equality detector | **Yes** — M4 → exactly 1 Core fail |
| hashing detector | **Yes** — M5 (hash made inconsistent with `Equals`) → exactly 1 Core fail |

### 2. SF2 — passes; implemented byte-for-byte as built
`StatementBinder.Loops.cs:637-639` is exactly `ComputeConditionNarrowing(condition).Then`, two call sites.

| probe | main | HEAD |
|---|---|---|
| `while d != nil { d = nil; d(1) }` | runs, **NRE rc=134** | **GS0503** |
| `for d != nil { d = nil; d(1) }` | NRE rc=134 | **GS0503** |
| `for var i=0; d != nil; i++ { d = nil; d(1) }` | NRE rc=134 | **GS0503** |
| `do { d(1) } while n<1` | runs | **GS0503** — still no narrowing, correct |
| **`while o is string { o.Length }`** | **GS0159 + GS0158 hard error** | **prints `5`** |

The `while … is T` pin is real and in-repo: `[InlineData("type-test-while", "4\n")]` at `Issue2885NullableDelegateReceiverTests.cs:293`, asserting stdout.

### 3. SF1 — passes, and exceeds the ask
All 8 receiver shapes now emit consistent advice, including the two direct-call forms I flagged as having *no working null-safe remedy at all*:
```
arr[0](1)        -> "Use '?.Invoke(...)'"   (was '?(...)', which did not compile)
Get()(1)         -> "Use '?.Invoke(...)'"   (was '?(...)')
arr[0].Invoke(1) / Get().Invoke(1) -> "Use '?.Invoke(...)'"
d(1) / d.Invoke(1) / h.Handler(1) / h.Handler.Invoke(1) -> "Use '?(...)'"
```
**13/13 emitted remedies compiled and ran, each printing `1`. Zero GS0155.** No over-fire regression: all 7 suppression shapes still suppress.

### 4. SF3 — half done, and my briefing had it backwards
They are four **guards**, not four pins — all four pass identically on `60e5b8bf` and HEAD. Only `box.Mutable guarded` (main 45 → HEAD GS0503) is a pin.

Independence, one rule at a time (control 52/52, 24/24): M1 readonly-field rule → 1 fail; M2 get-only backing-field rule → 1 fail; M3 drop `ApplyMemberNarrowing` → 2 fails (= M1 ∪ M2). So **two** independent new-branch detectors, up from round 2's single detector covering six rows.

The `if let` and `!!` tests **have no production branch in this diff to mutate** — both desugar to a non-nullable type before `TryReportNullableDelegateReceiver` is reached, on machinery this PR does not touch. They are legitimate over-fire guards and cannot be pins. **SF3 is as done as the code allows**; only the PR wording needs correcting.

### 5. Rebase correctness — clean
GS0500/0501/0502/**0503**/**0508** all present in ID order at `DiagnosticDescriptors.cs:373-377`; correct ordinal position in both doc files; `website/versioned_docs/` untouched; trailing newlines intact; no stale GS0502 nullable-delegate references. N4 fixed — the trigger column now holds a real trigger, verified to actually emit GS0503.

### 6. Merge coordination from `c47993bd`
#2912, #2926, #2929, #2932, #2935 **clean**. #2911 conflicts in `DiagnosticDescriptors.cs`, `DiagnosticBag.Reports.Declarations.cs` and both docs — mechanical, second merger resolves. `ExpressionBinder.Calls.Invocation.cs` is still not a hot spot: #2929 rewrites it and merges clean.

### 7. #2943 — accurate, correctly scoped, honest
Title is general, the repro uses a nullable **class** rather than a delegate, it references ADR-0069, and it states explicitly that this is pre-existing and reproduces identically on `main`. Closing #2885 with it open is correct — they are different bugs: #2885 is *no narrowing*, #2943 is *narrowing survives a back-edge*. The #2925 trap comment is posted and precise.

### Baselines — settled
**Measured `origin/main` @ `60e5b8bf` Compiler.Tests = 3742, 0 failed.** Author exact.

| suite | `d4038f1d` | `60e5b8bf` | + this PR |
|---|---|---|---|
| Compiler | 3708 | **3742** (measured) | +46 → **3788** |
| Core | 6913 (+1 skip) | 6913 | +2 → 6915 |
| Interpreter | 453 | 453 | +2 → 455 |

Delta verified by count, not trust: 11 `[Fact]` + 34 `[InlineData]` + 1 from the Issue1410 split = **+46**.

**Two corrections to my briefing:** #2923 added **34** Compiler tests, not 32 — my "3708 + 32" was the source of the mismatch, not the author's number. And **3742 is already stale** as the series baseline now that `main` has moved again.

### Nits
- `AccessPathTests.cs:38` asserts that `Fixture<int>.Value` equals `Fixture<string>.Value` (generic instantiations share the definition's token) without saying why that is sound. It **is** sound — `AccessPath.Equals` compares `Root` by reference first, a root has one declared type, and nested narrowing resolves innermost-first — but one comment would stop the next reader filing a bug.
- `PathMember.cs:76-78` source-symbol hash arm has no detector (constant-hash mutant survives). Stays `Equals`-consistent, so it is a perf mutant, not a correctness gap.
- Honest cost note: N1 alone (fix the hash inline, keep `ImmutableArray<object>`) was ~6 lines and fixed the actual bug; `PathMember` was my *optional* N5, +83 lines on top. It is landed, tested and mutation-detected, and it improves a public surface, so reverting now costs more than it saves. Recording the tradeoff, not asking for it back.

PR #2911, opened 2026-07-30
gsc: validate implicit interface property types

### LINKED ISSUE #2888: gsc: ordinary property type mismatch silently satisfies G# interface and emits unloadable type
Found while investigating #2875's general interface-implementation verification pass.

## Summary

For a G#-declared interface property and an ordinary class/struct property with the same name and accessor set, gsc does not compare the property types. A mismatched implementation is accepted without diagnostics and produces an assembly the CLR rejects.

This is separate from #2875's init-only-versus-set accessor mismatch: the repro uses get-only properties on both sides.

## Repro (verified)

```gsharp
package PropertyTypeMismatch

interface IBox {
    prop Value int32 { get; }
}

class Box : IBox {
    prop Value string { get; }
}
```

```text
$ dotnet gsc.dll property-type-mismatch.gs /out:property-type-mismatch.dll /target:library /targetframework:net10.0
Wrote property-type-mismatch.dll

$ ilverify property-type-mismatch.dll -r '/usr/local/share/dotnet/shared/Microsoft.NETCore.App/10.0.8/*.dll'
[MD]: Error: Class implements interface but not method, Class: '[PropertyTypeMismatch]PropertyTypeMismatch.Box' Interface: '[PropertyTypeMismatch]PropertyTypeMismatch.IBox' Missing method: '[PropertyTypeMismatch]PropertyTypeMismatch.IBox.get_Value()'.
1 Error(s) Verifying property-type-mismatch.dll
```

The equivalent plain `struct` shape fails identically. Missing properties and missing interface methods already report GS0187 correctly.

## Expected

The binder must reject the mismatched property signature before emit, naming the type, interface property, expected type, and actual type. No output assembly should be written.

## Root cause

`DeclarationBinder.VerifyInterfacePropertyImplementationsAndCollectPositionalCandidates` resolves ordinary properties through `TypeMemberModel.TryGetProperty`, which matches by name, then checks only required getter/setter presence. It applies type compatibility only in the separate positional-primary-constructor fallback. Ordinary declared properties therefore bypass any type comparison.

Explicit-interface properties do compare exact types in `TryResolveExplicitInterfacePropertyImplementation`, and imported CLR interfaces use `MemberLookup.FindMatchingProperty`, so those paths are not this defect.

- Get-only and get/set mismatches on plain classes and structs.
- Nullable covariance for get-only properties and invariance for settable properties, matching the positional-member rules.
- Generic interface properties after type-argument substitution.
- IL verification/no-output assertions for rejected programs.

DavidObando:
Closes #2888.

## Root cause and fix

Ordinary G# interface-property verification previously stopped at the first same-name property and did not verify its type. That could emit assemblies which failed `Assembly.Load(...).GetTypes()` because the CLR interface slot had no matching implementation.

All G#-declared instance interface properties now use one base-chain candidate walk. It:

- checks emitted slot identity before source-level nullable variance;
- skips explicit-clause and inaccessible inherited candidates;
- preserves ADR-0118's plain-indexer rule through one shared helper;
- matches indexer parameters after generic substitution;
- handles positional properties through the same type-mismatch diagnostic;
- continues past wrong-type derived properties to legal G# or imported CLR base implementations;
- retains the first usable wrong-type candidate for `GS0504`.

## Round-3 nullable correction

Removed the value-nullable equality gate added in round 2. It compared a declared interface shape with a substituted implementation shape and rejected valid nullable type arguments such as `I[int32?] <- int32?`.

Exact six-case rerun:

| Case | `main` | round-2 head `d304a53b` | current |
|---|---|---|---|
| `I[T]`, `I[int32?] <- int32?`, get | LOAD/RUN OK | GS0504, no emit | LOAD/RUN OK |
| `I[T]`, `I[string?] <- string?`, get | LOAD/RUN OK | LOAD/RUN OK | LOAD/RUN OK |
| `I[T?]`, `I[int32?] <- int32?`, get | LOAD/RUN OK | GS0504, no emit | LOAD/RUN OK |
| `I[T]`, `I[int32?] <- int32?`, get/set | LOAD/RUN OK | GS0504, no emit | LOAD/RUN OK |
| positional `I[T]`, `I[int32?] <- int32?` | LOAD/RUN OK | GS0504, no emit | LOAD/RUN OK |
| nested `I[T?]`, `I[int32?] <- int32?` | LOAD/RUN OK | GS0504, no emit | LOAD/RUN OK |

The review report said all six failed on `d304a53b`; exact rebuilt-head measurement found five. The reference-nullable row already passed.

Reference-nullability is erased before generic substitution when projecting the emitted interface slot. Declared value-nullable wrappers remain intact. Source get-only covariance and set/init invariance remain a separate final check.

## Review reductions

- **SF2:** Added separate implicit and explicit static-virtual regression facts. Static emit still requires both exact `TypeSignaturesEquivalent` and source compatibility; dropping either emits unloadable metadata. Filed #2945 to unify static-virtual nullable erasure with the instance path.
- **SF3:** When formatted expected and actual types are identical, the verifier falls back to GS0187 instead of emitting a misleading `GS0504`.
- **SF4:** Took the reduction. Removed the mapped public `TypeSymbol.AreRuntimeEquivalentIgnoringReferenceNullability` overload; `TypeSymbol.cs` is now identical to `origin/main`. Round-3 file delta: +7/-40 (net -33) and one public API removed. The untested tuple relaxation was intentionally dropped; `main` rejects it too.
- **SF5:** Took the centralization. `GetInterfacePropertySlotType` fell from 101 lines to 8 and delegates to shared `StructSymbol` construction substitution with pre-substitution reference-nullability erasure.
- **M14b:** Extracted the duplicated ADR-0118 condition into `IsUnsupportedImplicitInterfaceIndexer`, used by ordinary and imported-base lookup.

Round-3 production delta versus `d304a53b`: +114/-180, net **-66 lines**. Tests add 77 lines. Total round-3 delta: +191/-180.

## Metadata/load sweep

A generated 230-case matrix covered:

- `T` and `T?`;
- `int32`, `int32?`, `string`, and `string?` type arguments;
- get, set, and init;
- implicit class/struct, explicit, positional class/struct, open generic, and struct-constrained providers;
- slice, array, nested constructed type, indexer, and implicit/explicit static-virtual shapes.

Every emitted assembly ran through `Assembly.Load`, `GetTypes`, and concrete-type activation:

- `main`: 90 LOAD/RUN OK, 110 NO-EMIT, **30 LOAD/RUN FAIL**
- branch: 99 LOAD/RUN OK, 131 NO-EMIT, **0 LOAD/RUN FAIL**

## Corpus drift

All 148 `samples/**/*.gs` and `e2etests/**/*.gs` files:

- `main`: 139 LOAD OK / 9 NO-EMIT / 0 LOAD-FAIL
- branch: 139 LOAD OK / 9 NO-EMIT / 0 LOAD-FAIL
- DLL/runtimeconfig/diagnostic/exit/load-state drift: 0

Direct feature coverage is **0/148**: 2 files contain properties, 8 contain interfaces, and none contain both. This corpus is collateral-only signal for #2888.

## Pins, guards, and mutation audit

Round-2 independent count was **13 differential pins** and four unpinned semantic branches: M7, M8, M10, and M14b.

- M10 was deleted.
- M7 and M8 now have separate static-virtual facts.
- M14b is one shared helper guarded by the existing ADR-0118 plain-indexer fact.
- Round 3 adds 9 baseline guard shapes: 6 nullable-argument acceptances, 2 static-virtual rejections, and 1 imported-generic diagnostic fallback.

Each round-3 mutant was build-green, run only after its unmutated control passed, then restored:

| Semantic branch | Mutation | Result |
|---|---|---|
| identical-display diagnostic fallback | inverted equality branch | imported-generic fact failed |
| deleted value-nullable equality gate | restored round-2 gate | nullable-argument guard failed |
| emitted-slot erasure | disabled erasure at delegation | nullable property-surface guard failed |
| declared value-nullable preservation | erased every nullable wrapper | constructed-generic mismatch pin failed |
| recursive constructed-type erasure | stopped propagating erasure into struct arguments | nested `Cell[T?]` guard failed |
| implicit static exact signature | dropped `TypeSignaturesEquivalent` | implicit static fact failed |
| explicit static exact signature | routed explicit static through instance compatibility | explicit static fact failed |
| shared ADR-0118 rule | forced helper false | plain-indexer fact failed |

## Validation

- Release solution build, warnings as errors: 0 warnings, 0 errors
- Core.Tests: **6,916 passed, 1 skipped**
- Compiler.Tests: **3,818 passed**
- Interpreter.Tests: **470 passed**
- Final targeted controls: 17 Compiler + 28 Core passed

Final executed-count base is `origin/main` **914b42a7**. Main CI: Core 6,915/1 skip, Compiler 3,802, Interpreter 470, LSP 452, Extensions 104, SDK 57. Branch locally: Core 6,916/1 skip, Compiler 3,818, Interpreter 470 — delta **+1 Core, +16 Compiler, +0 Interpreter**.

Landing note: PR #2912 adds a nested `EnumSymbol` arm to
`StructSymbol.SubstituteTypeForConstruction`. Its recursive callback must use
`SubstituteNested(t)`, not a direct three-argument call, so
`eraseReferenceNullability` propagates when both PRs are present.

#2915 remains deliberately preserved; this PR does not change the plain-vs-constructed implicit-indexer behavior it tracks.


DavidObando:
## Round-2 review — PR #2911 (`d304a53b`), issue #2888

# VERDICT: **NOT MERGEABLE** — 1 blocker

Fix real. Repro dead. But PR adds gate that **rejects valid programs `main` compiles and loads**, and that gate is pinned by **zero** tests in 6914 Core + 1002 Compiler. Delete gate → blocker gone, diff shrinks 11 lines, nothing breaks.

---

## BLOCKER 1 — over-fire. Value-nullable gate rejects good code.
**`/Users/davidobando/GitHub/DavidObando/gsharp-wt/i2888/src/Core/CodeAnalysis/Binding/DeclarationBinder.InterfaceVerification.cs:678-689`**

```csharp
var interfaceSlotIsValueNullable = interfaceType is NullableTypeSymbol declaredNullable   // DECLARED, un-substituted
    && NullableLifting.IsAnyValueTypeNullable(declaredNullable);
...
if (interfaceSlotIsValueNullable != implementationSlotIsValueNullable) return false;      // IV:686
```

Gate compare **declared** interface type against **implementation** type. When *type argument itself* nullable, gate compare pre-substitution vs post-substitution. Wrong side. Same class of defect as round 1 blocker, moved one line down.

Probe. All six compile+`Assembly.Load`+`GetTypes`+`Activator.CreateInstance` on `main`. All six GS0504 on HEAD:

| case | source shape | main | HEAD | gate deleted |
|---|---|---|---|---|
| `nq_argT` | `interface I[T]{prop V T{get;}}` · `class C : I[int32?] { prop V int32? }` | LOAD-OK | **GS0504** | LOAD-OK |
| `nq_argT_str` | same @ `I[string?]` ← `string?` | LOAD-OK | **GS0504** | LOAD-OK |
| `nq_argTq` | `prop V T?` @ `I[int32?]` ← `int32?` | LOAD-OK | **GS0504** | LOAD-OK |
| `g_argq_set` | `prop V T {get;set;}` @ `I[int32?]` ← `int32?` | LOAD-OK | **GS0504** | LOAD-OK |
| `g_argq_pos` | positional data class @ `I[int32?]` ← `int32?` | LOAD-OK | **GS0504** | LOAD-OK |
| `g_nested_nullnull` | `prop V T?` @ `I[int32?]` ← `int32?` | LOAD-OK | **GS0504** | LOAD-OK |

Message self-contradict — user cannot act on it:
```
error GS0504: Type 'C' cannot use property 'V' to implement interface property 'I[int32?].V':
              expected type 'int32?', actual type 'int32?'.
```

Gate also **arbitrary**: it only inspect top level. Same semantics one level down passes.
- `g_argq_slice` (`prop V []T` @ `I[int32?]` ← `[]int32?`) → **LOAD-OK on HEAD**.
- `g_argq_cell` (`prop V Cell[T]` @ `I[int32?]` ← `Cell[int32?]`) → **LOAD-OK on HEAD**.
Bare `int32?` rejected, `[]int32?` accepted. No rule there.

Gate protect nothing. Every case it claim to guard already rejected by gate 1 (`AreRuntimeEquivalentIgnoringReferenceNullability` require **both** sides be `NullableTypeSymbol`). With gate deleted:
- `g_argq_mismatch` (`T`@`I[int32?]` ← `int32`) → still REJECT (main: LOAD-FAIL)
- `g_structconstr_nq` (`[T struct] T`@`I[int32]` ← `int32?`) → still REJECT (main: LOAD-FAIL)
- `genS_QN_g` (`[T struct] T?`@`I[int32]` ← `int32`) → still REJECT
- `genUv_QQ_*` (`T?`@`I[int32]` ← `int32?`, the #2888 family) → still REJECT

Gate unpinned. Mutant `if (false && ...)`:
| suite | control | gate deleted |
|---|---|---|
| Core.Tests **full** | 6914 P / 1 S / 0 F | **6914 P / 1 S / 0 F** |
| Compiler.Tests `Interface\|Propert\|Indexer\|Nullab` | 1002 P / 0 F | **1002 P / 0 F** |

Across 123-case probe grid, deleting gate changed **exactly 3 outcomes** — all three false-rejection→correct-accept. Plus 3 more in `cases8`. Zero regressions.

**Fix (ponytail, −11 lines):** delete `IV:678-689` entirely. Keep only `var implementationIsNullable = implementationType is NullableTypeSymbol;` (used by gate 3). Comment at IV:678-680 is also wrong-headed — it describe erasure, which gate 1 + `GetInterfacePropertySlotType` already do. Add regression test for `I[int32?]` ← `int32?` get and get/set.

---

## SHOULD-FIX 2 — `TypeSignaturesEquivalent` conjuncts load-bearing but unpinned
**`.../DeclarationBinder.InterfaceVerification.cs:1529`** and **`.../DeclarationBinder.Functions.cs:1885`**

**Q1 answer: (b). And briefing premise wrong.** Two predicates **incomparable**, not one-way redundant. Direct reflection probe on `DeclarationBinder.TypeSignaturesEquivalent`:
- TSE **nullability-blind at leaves** — `TSE(int32, int32?) == True`, `TSE(string, string?) == True`, because `NullableTypeSymbol.ClrType == UnderlyingType.ClrType` (`src/Core/CodeAnalysis/Symbols/NullableTypeSymbol.cs:24`). `IsInterfacePropertyTypeCompatible` correctly `False`.
- TSE **stricter** for constructed imported generics — `ImportedTypeSymbol` branch refuses when `ContainsTypeParameter`; `AreRuntimeEquivalentIgnoringReferenceNullability` falls to erased `ClrType` compare because `GetWrappedTypes` (`TypeSymbol.cs`) has **no `ImportedTypeSymbol` arm**.

Built Alternative A (delete both TSE conjuncts). **UNSAFE** — proves TSE is load-bearing:
```gsharp
sealed interface I[T] { shared { prop V T? { get; } } }
struct C : I[int32]  { shared { prop V int32 -> 1 } }
```
→ emits, then `LOAD-FAIL: Virtual static method 'get_V' is not implemented on type 'StImplicitCov.C'`. Static-virtual emit path does **not** perform the erasure instance path does. So `Fn:1709` is **not** missing anything (instance explicit path safe — `inst_explicit_cov` LOAD-OK on HEAD). **Neither call site is wrong.**

But: mutants M7 (drop TSE at IV:1529) and M8 (drop TSE at Fn:1885) both **SURVIVED** Core-full + Compiler-1002. Load-bearing code, zero pinning. Next "ponytail cleanup" reintroduces an unloadable-emit bug. **Fix:** add regression test for the static-virtual `sealed interface` + `shared` covariance case, plus one-line comment naming the static-virtual erasure gap.

Related, not new: static-virtual path rejects relaxations instance path accepts (`st_implicit_cov`, `st_implicit_fnT`, `st_explicit_cov` → GS0397). Same on `main`, so no regression — but `TSE && IIPTC` shape now institutionalises the inconsistency. Worth an issue, not a blocker.

---

## SHOULD-FIX 3 — GS0504 prints identical expected/actual for imported generics
**`.../DeclarationBinder.InterfaceVerification.cs` `GetInterfacePropertyExpectedType`** (~IV:826)

```
error GS0504: Type 'IntSeq' cannot use property 'Items' to implement interface property 'ISeq[int32].Items':
              expected type 'System.Collections.Generic.List[int32]', actual type 'System.Collections.Generic.List[int32]'.
```
Cause: message formatter (`GetInterfacePropertySlotType`) substitutes through `ImportedTypeSymbol` type arguments; checker (`AreRuntimeEquivalentIgnoringReferenceNullability`) cannot, and compares erased `ClrType` (`List<object>` vs `List<int32>`). Hits `gi_listT`, `gi_dictT`, `gi_seqT`, `gi_actT`, `gi_nestcell`, `x_importedgeneric`, `inst_explicit_listT`.

Not a new rejection — all were GS0187 on `main`. But GS0504 is a **new actively misleading** message. Cheap fix: when formatted expected == actual, fall back to `ReportInterfaceMethodNotImplemented` (GS0187).

---

## SHOULD-FIX 4 (ponytail) — revert `TypeSymbol.cs` entirely. Built it. It holds.
**`/Users/davidobando/GitHub/DavidObando/gsharp-wt/i2888/src/Core/CodeAnalysis/Symbols/TypeSymbol.cs` (+47, new public overload, param threaded through 6 recursive sites + nested local `AreSameConstructedUserTypeCore`)**

Alternative B: revert `TypeSymbol.cs` to `origin/main` (−47 lines, −1 public API surface), and in gate 1 substitute **once at the call site** using the PR's own helper:
```csharp
if (!TypeSymbol.AreRuntimeEquivalentIgnoringReferenceNullability(
        GetInterfacePropertySlotType(interfaceType, typeParameterMap),
        implementationType))
```
Measured: **116 of 117 probes byte/diagnostic identical to HEAD.** Core filter 1140 P, Compiler `Issue2888|Issue2892|Issue2150` 13 P. Single difference: `gi_tupT` (`interface I[T]{prop V (T, string)}` @ `I[int32]` ← `(int32, string)`) — HEAD accepts, AltB rejects (same as `main`).

So the whole recursive threading buys exactly one thing: substitution through `Tuple`/`Sequence`/`Channel`/`Pinned`/`FunctionPointer` wrappers reached via `GetWrappedTypes`. **None of it tested. None required by #2888.** Function types still work under AltB (`GetInterfacePropertySlotType` has a `FunctionTypeSymbol` arm).

If author want the tuple relaxation, keep it — but then it needs a test, and PR should say so. Otherwise −47 lines free.

---

## SHOULD-FIX 5 (ponytail) — `GetInterfacePropertySlotType` duplicates existing machinery
**`.../DeclarationBinder.InterfaceVerification.cs:697-~810` (~110 lines)**

Re-implements Slice/Array/Map/ByRef/Pointer/Struct/Interface/Imported/Delegate/Function traversal already in `StructSymbol.SubstituteTypeForConstruction` — reached via `StructSymbol.SubstituteTypeParameters`, which **this same PR already calls** from `SubstituteInterfacePropertyType` (IV:~814). Only added behaviour is per-level erasure of reference-nullable wrappers:

```
GetInterfacePropertySlotType(t, map)  ≡  SubstituteTypeParameters(deep-strip-reference-nullable(t), map)
```
Order matters — strip **before** substitute, else `T?`@`I[int32]` wrongly yields `int32?`. A ~15-line strip helper + existing substituter replaces ~110 lines.

---

## Confirmed solid (no action)

**Q2 — erasure side right this time.** Round-1 defect gone. Probed:
- `genUv_QN_g` (`T?`@`I[int32]` ← `int32`, unconstrained) → **LOAD-OK** ✅ (`T?` erases to `T`)
- `genS_QN_g` (`[T struct] T?`@`I[int32]` ← `int32`) → **GS0504** ✅ (constraint makes slot `Nullable<int32>`)
- `genUv_QQ_g` (`T?`@`I[int32]` ← `int32?`) → main **LOAD-FAIL**, HEAD **GS0504** ✅
Struct-constrained vs unconstrained genuinely diverge, both correct. `interfaceType` (declared) passed as `left`; map substitutes into left. Correct.

**Q3 truth table — real load/run, 72 cases × main/HEAD. No cell on HEAD produced unloadable type or NPE.**

| shape | iface/impl | get | get+set | get+init |
|---|---|---|---|---|
| ref `string` | N/Q | LOAD-OK → GS0504 | LOAD-OK → GS0504 | LOAD-OK → GS0504 |
| ref | Q/N | LOAD-OK → **LOAD-OK** | LOAD-OK → GS0504 | LOAD-OK → GS0504 |
| ref | Q/Q, N/N | LOAD-OK → LOAD-OK | LOAD-OK → LOAD-OK | LOAD-OK → LOAD-OK |
| val `int32` | N/Q | **LOAD-FAIL** → GS0504 | LOAD-FAIL → GS0504 | LOAD-FAIL → GS0504 |
| val | Q/N | **LOAD-FAIL** → GS0504 | LOAD-FAIL → GS0504 | LOAD-FAIL → GS0504 |
| genUnconstrained@`int32` | Q/N | GS0187 → **LOAD-OK** | GS0187 → GS0504 | GS0187 → GS0504 |
| genUnconstrained@`int32` | Q/Q | **LOAD-FAIL** → GS0504 | LOAD-FAIL → GS0504 | LOAD-FAIL → GS0504 |
| genStructConstrained@`int32` | Q/N | GS0187 → GS0504 | GS0187 → GS0504 | GS0187 → GS0504 |
| genStructConstrained@`int32` | N/Q | **LOAD-FAIL** → GS0504 | LOAD-FAIL → GS0504 | LOAD-FAIL → GS0504 |

Get-only arm rejects exactly {iface non-nullable ∧ impl nullable}. Nullable **value**-type covariance never reaches it — gate 1 blocks it. Sound.

`hasSetter` means **"has any setter"**. `init` and `set` columns identical in every one of 72 cells. Orthogonal to **GS0502** (`InterfacePropertySetterKindMismatch`, PR #2892) which still fires independently — `GetSetNullableInterfaceProperty_NotSatisfiedByNullablePositionalParam_ReportsGS0502` still passes. No interaction bug.

Source-breaking tightenings on previously-loadable programs (`ref_NQ_*`, `genUr_NQ_*`, `ref_QN_gi/gs`) are all explicitly what #2888 asked for: covariance for get-only, invariance for settable. Intentional, documented.

**Q4 — definitive.** `BuildInterfaceTypeParameterMap` (`.../DeclarationBinder.Functions.cs:1912`) allocates a **fresh** `Dictionary` per call. No caching, no sharing, no escape. `grep` for `typeParameterMap[...]=`/`.Add`/`.Remove` on the parameter → **zero hits**. Nothing mutates it. **No shared-mutation bug.** Widening `IReadOnlyDictionary`→`Dictionary` is forced purely because `StructSymbol.SubstituteTypeParameters(TypeSymbol, Dictionary<TypeParameterSymbol, TypeSymbol>)` (`src/Core/CodeAnalysis/Symbols/StructSymbol.cs:1641`) takes a concrete `Dictionary`. Note the internal inconsistency: new `TypeSymbol` overload takes `IReadOnlyDictionary`, `IsInterfacePropertyTypeCompatible` takes `Dictionary`. **NIT**, fixable by widening `SubstituteTypeParameters`'s param instead.

**Q6 — per-assertion. All six changes to `test/Core.Tests/CodeAnalysis/Binding/Issue2150DataClassInterfacePropertyTests.cs` legitimate. No papered-over regression.**

| # | change | verdict + probe |
|---|---|---|
| 1 | `TypeMismatchedPositionalParam_ReportsSingleGS0187` → `..._ReportsSingleGS0504` | ✅ same rejection, ID only |
| 2 | `GetOnlyNullableInterfaceProperty_SatisfiedByNonNullablePositionalParam_NoDiagnostics` → `GetOnlyNullableValue..._NotSatisfiedBy..._ReportsGS0504` | ✅ **the one true inversion.** Compiled its exact source (`interface IHasNullableX { prop X int32? { get; } }` / `data class NonNullableX(X int32) : IHasNullableX`) with `origin/main` gsc → **emits, then LOAD-FAIL: `Method 'get_X' in type 'PosValQ.NonNullableX' ... does not have an implementation`.** Old test pinned a second instance of #2888. Correction, not cover-up. |
| 3 | new `GetOnlyNullableReferenceInterfaceProperty_SatisfiedByNonNullablePositionalParam_EmitsAndLoads` | ✅ reference companion; `pos_refq` LOAD-OK on main **and** HEAD. Correct value/reference split guard. |
| 4-6 | `GetOnlyNonNullable...`, `GetSetNullable...`, `GetSetNonNullable...` GS0187→GS0504 | ✅ same rejections, ID only |

Also confirmed: round-1 brief's flagged bug-pinning assertion in `Issue2888...Tests.cs:220-226` **was** inverted, correctly and completely.
NIT (pre-existing, untouched): comment on `GetOnlyNonNullableInterfaceProperty_...` says "iface `int32` <- impl `int32?`" but source uses `string`/`string?`.

**Q8 — fix works, no over-fire other than BLOCKER 1.** #2888 exact repro: `main` compiles + emits + `LOAD-FAIL`; HEAD `GS0504`, no emit. Over-fire hunts: 21-shape non-generic identity grid — clean. 14-shape generic substitution grid — clean plus several intended relaxations. Static-vs-instance, positional data-class, imported-generic grids — clean.

**Q9 — GS0504 mechanics correct.** `src/Core/CodeAnalysis/DiagnosticDescriptors.cs:376`, severity `Error`, placed immediately after GS0502 (GS0503 reserved for #2908 — correct ordinal, no renumber). Documented in **both** `docs/diagnostics.md` (before GS9998–GS9999 section) and `website/docs/ref/diagnostics.md`; trailing newlines intact. `ReportInterfacePropertyTypeMismatch` at `src/Core/CodeAnalysis/DiagnosticBag.Reports.Declarations.cs:1095`. `DiagnosticIdUniquenessTests` green. Message clear for normal case — imported-generic case is SHOULD-FIX 3.

**Q10 — merge order.** `git merge-tree HEAD <pr-head>`:
- **#2912** `8dc1e8f7`, **#2926** `bb0ee017`, **#2929** `99019f64` → **clean**
- **#2908** `29c99c6f` → `docs/diagnostics.md`, `DiagnosticDescriptors.cs`, `website/docs/ref/diagnostics.md`
- **#2917** `b00e2121` → `DiagnosticDescriptors.cs`, `website/docs/ref/diagnostics.md`
- **#2923** `62db85e5` → all three doc/descriptor files

All conflicts are adjacent-line descriptor/doc additions. **Mechanical.** **Zero conflicts in `TypeSymbol.cs`, `Binder.cs`, `DeclarationBinder.*`.**

**#2915 snippets verified fixed** — compiled both repro snippets verbatim from issue body:
| snippet | main | HEAD |
|---|---|---|
| plain `interface I { prop this[k string] int32 }` | GS0187 REJECT | GS0187 REJECT |
| generic `interface I[T] { prop this[k string] T }` @ `I[int32]` | LOAD-OK | LOAD-OK |
Both parse and reproduce the stated inconsistency. Author's claim honest. PR preserves the inconsistency deliberately, as #2915 assumes.

**#2915 friction:** PR duplicates the ADR-0118 guard in two places that must change in lockstep — `IV:844` (`|| (candidate.IsIndexer && ReferenceEquals(iface.Definition ?? iface, iface))`) and `IV:897` (same condition on `interfaceProperty.IsIndexer`, inside `HasMatchingImportedBaseProperty`). **IV:844 is pinned** (mutant → `Issue2370ExplicitInterfaceEventIndexerEmitTests` 1 F / 25 P). **IV:897 is NOT** (mutant survived Compiler `Interface|Indexer` 454 P / 0 F). Extract to one named helper, or #2915 will silently drift them apart.

---

## Q7 — my own mutation table
Build-green mutants only. One hunk each. `git checkout --` + rebuild between every run.
**Controls (unmutated, same worktree):** Core.Tests **full** 6914 P / 1 S / 0 F · Compiler.Tests `Interface|Propert|Indexer|Nullab` **1002 P / 0 F** · Core filter 1140 P · Compiler `Issue2888|Issue2892|Issue2150` 13 P.

| # | hunk (`file:line`) | mutation | result |
|---|---|---|---|
| M1 | IV:584 positional-param call site | `true \|\| IIPTC(...)` | **KILLED** Core 5F, Comp 2F |
| M2 | IV:870 base-chain call site | `true \|\| IIPTC(...)` | **KILLED** Core 5F, Comp 4F |
| M3 | IV:931 imported-base call site | `true \|\| IIPTC(...)` | **KILLED** Comp 1F |
| M4 | IV:1530 static-implicit call site | `true \|\| IIPTC(...)` | **KILLED** Comp 1F |
| M5 | Fn:1709 explicit-instance call site | `true \|\| IIPTC(...)` | **KILLED** Comp 2F |
| M6 | Fn:1886 explicit-static call site | `true \|\| IIPTC(...)` | **KILLED** Comp 1F |
| M9 | IV:670 gate 1 runtime-equivalence | `true \|\| AREIRN(...)` | **KILLED** Core 1F, Comp 5F |
| M11 | IV:692 gate 3 whole ternary | `return true` | **KILLED** Core 1F, Comp 1F |
| M11b | IV:694 gate 3 get-only arm | `: true` | **KILLED** Core 1F, Comp 1F |
| M12 | `TypeSymbol.cs` map threading | disable substitution | **KILLED** Core 5F, Comp 3F |
| M13 | IV:889 `HasMatchingImportedBaseProperty` call | force `false` | **KILLED** Comp 2F |
| M16 | IV:691 `SubstituteInterfacePropertyType` | pass un-substituted | **KILLED** Comp 1F |
| M14 | IV:844 ADR-0118 plain-indexer guard | `false && ...` | **KILLED** Issue2370 1F/25P |
| **M7** | **IV:1529 `TypeSignaturesEquivalent` conjunct** | deleted | **SURVIVED** — load-bearing (proved unloadable emit) |
| **M8** | **Fn:1885 `TypeSignaturesEquivalent` conjunct** | deleted | **SURVIVED** |
| **M10** | **IV:686 gate 2 value-nullable equality** | `false && ...` | **SURVIVED** — Core-full 6914P, Comp-wide 1002P. Removes 6 false rejections. |
| **M14b** | **IV:897 duplicate ADR-0118 guard** | `false && ...` | **SURVIVED** — Comp `Interface\|Indexer` 454P |

**Counted separately, semantic-branch granularity: 13 differential pins · 4 unpinned semantic branches (M7, M8, M10, M14b).**
All **6** call sites individually pinned ✅. All 3 sub-conditions inside the predicate probed individually — gate 1 ✅, gate 3 ✅, **gate 2 ✗**.

Reconcile with author's "15/15 killed": my kill count is **13**, matching their "13 differential pin branches". Their 15/15 either used coarser hunk granularity or counted mutants that don't isolate a branch. **Four semantic branches ship unpinned.** Report should say so.

**Load sweep re-run.** 154-case grid re-executed under a `Assembly.Load(File.ReadAllBytes)` + `GetTypes()` + `GetInterfaceMap()` + `Activator.CreateInstance` oracle, **validated against a known-bad control** (main-compiled #2888 repro → `LOAD-FAIL ... Method 'get_Value' ... does not have an implementation`). **0 LOAD-FAIL on HEAD confirmed.** Harness catches failures. (Note: `IlVerifier.Verify` not trusted; not used.) One harness artifact found and fixed on my side — open generic types must be skipped for `CreateInstance` (`ContainsGenericParameters`), else spurious CTORFAIL.

**Corpus honesty — state the number.** Of 148 `samples/**/*.gs` + `e2etests/**/*.gs`: **2** contain `prop `, **8** contain `interface`, **0** contain both. Author's "collateral signal only" framing is honest, but the corpus has **literally zero** coverage of this feature. Say `0/148`, not `148 inputs`.

---

## Corrections to your briefing
1. **6 call sites, not 7.** `grep` confirms: IV:584, IV:870, IV:931, IV:1530, Fn:1709, Fn:1886 (+1 definition at IV:664). Your own list had 6 items under a "7" heading.
2. **IV:931 is not an `&&`-combined type check.** The `TypeSignaturesEquivalent` at IV:920 compares indexer **parameter** types — orthogonal to property type. Only **IV:1530** and **Fn:1886** combine `TSE && IIPTC` on the property type. Q1 applies to two sites, not four.
3. **Baselines in the brief are wrong for this branch.** Verified unmutated controls: Core **6914 P / 1 S**, Compiler **3721 P**. Brief said "Core 6913/1, Compiler ~3734". Author's report matches reality; **there is no Compiler delta to reconcile** — the "~3734" figure is stale. The PR *body*'s numbers (6,871 / 3,691) are stale round-1 figures — SHOULD-FIX: update PR body.
4. **Q1 has no wrong site.** Both `:1709` (IIPTC alone) and `:1886` (TSE && IIPTC) are correct as written, because the instance and static-virtual emit paths genuinely differ in erasure behaviour. The premise "one of these two is wrong" is false. What *is* wrong is that the asymmetry is undocumented and unpinned.

---

## NITs
- `.../DeclarationBinder.Functions.cs:1912` `Dictionary` widening — benign plumbing, not a bug (Q4). Optional cleanup: widen `StructSymbol.SubstituteTypeParameters`'s param to `IReadOnlyDictionary` instead.
- PR body test counts stale (round-1 figures).
- `Issue2150DataClassInterfacePropertyTests.cs` stale `int32` comment on a `string` test (pre-existing).

---

## Ponytail scorecard
Round 1 `+635/−50`. Round 2 `+1149/−81`. Available reductions, all built and measured:
- BLOCKER 1 delete gate 2: **−11 lines**, fixes 6 regressions
- SHOULD-FIX 4 revert `TypeSymbol.cs`: **−47 lines, −1 public API**, 116/117 identical
- SHOULD-FIX 5 fold `GetInterfacePropertySlotType` into `SubstituteTypeParameters` + strip helper: **~−95 lines**

≈ **−150 lines** with strictly better behaviour. Diff doubling was not required to close the two blockers.

DavidObando:
## Round 3 review — MERGEABLE

Blocker gone. Reductions real and measured. All three round-2 unpinned branches now pinned. Zero LOAD-FAIL, zero new over-fire, zero PR-attributable merge conflicts. Only NITs remain.

Reviewed at branch HEAD **`df3e7ef7`** (the author rebased onto `914b42a7` mid-review; content verified byte-identical to `4eaba4ff` — 11 lines of difference, all context, one hunk header shifted and a `GS0503` neighbour from #2908; diffstat identical at 9 files, src **+304/−74**).

### Blocker deletion verified — 133-case grid, both compilers

All probes: compile → `Assembly.Load(File.ReadAllBytes)` → `GetTypes()` → `GetInterfaceMap()` → `Activator.CreateInstance`, with the oracle validated against a known-bad control.

| | LOAD-OK | REJECT | **LOAD-FAIL** |
|---|---|---|---|
| `main` `e21677f6` | 81 | 33 | **19** |
| HEAD `df3e7ef7` | 81 | 52 | **0** |

Every one of main's 19 LOAD-FAILs is rejected by HEAD. **Zero cases where main LOAD-FAILs and HEAD emits.** Over-fire set is 8, all previously cleared as intentional; **no new over-fire**. Arbitrariness is gone — bare `int32?`, `[]int32?` and `Cell[int32?]` at `I[int32?]` now all agree (LOAD-OK).

**Reviewer retraction:** my round-2 report listed 6 false rejections. Rebuilding `d304a53b` from scratch shows **`nq_argT_str` was already LOAD-OK** — that datum was contaminated, almost certainly captured while a mutant was applied. **The blocker affected 5 cases, not 6, and the author is right.** Mechanism confirms it: the deleted gate compared `NullableLifting.IsAnyValueTypeNullable` on *both* sides, so reference-nullable arguments gave `false != false` and always passed. Only value-nullable arguments misfired.

### SF4 trade-off handled, not dropped
`gi_tupT` now **REJECTS** with `GS0504`, restoring parity with `main`. Stated in the PR body; `grep` for tuple shapes across the test diff → **zero hits**, so no stale test was left behind or quietly deleted. `TypeSymbol.cs` is absent from the diff entirely — the +47 and the new public overload are gone.

### SF5 reduction is real, not lateral
| | round 2 | round 3 |
|---|---|---|
| `GetInterfacePropertySlotType` | ~101 lines hand-rolled traversal | **8 lines** (`IV:697-702`) delegating to `StructSymbol.SubstituteTypeParameters(..., eraseReferenceNullability: true)` |
| `TypeSymbol.cs` | +47, +1 public overload | **0** |
| src total | +369/−78 | **+304/−74** |

Of `StructSymbol.cs`'s +74, ~48 are a mechanical `SubstituteNested` local-function refactor of 12 pre-existing recursive call sites, ~14 the erasure prologue, ~12 the ByRef/Pointer arms.

**Strip-before-substitute ordering is structural**, not incidental — the erasure prologue at `StructSymbol.cs:2364-2377` runs before the `TypeParameterSymbol` branch at `:2380`. `genUv_QN_g` (`T?` @ `I[int32]` ← `int32`) → **LOAD-OK**, proving strip-then-substitute; substitute-then-strip would have produced `Nullable<int32>` and rejected. `genS_QN_g` (struct-constrained) correctly diverges → GS0504. Both halves pinned: mutant P17 (erase value-nullables too) → **KILLED** 3F; P18 (`SubstituteNested` passes `false`) → **KILLED** 1F.

### Recount: 20 semantic branches, **17 differential pins**, 3 provably inert
Author's "8/8" is a coarser hunk decomposition, not weaker coverage. At semantic-branch granularity: **17/20 killed** — strictly better than round 2's 13/17.

Fate of the four round-2 survivors:
- **M7 + M8** → collapsed into one named helper `IsStaticInterfacePropertyTypeCompatible` (`IV:705-714`) with an explicit `// Issue #2945` comment. **P12 = KILLED** 2F by two new facts asserting `CompileExpectingFailure` on exactly the `sealed interface` + `shared` shape I proved emits `Virtual static method 'get_V' is not implemented`.
- **M10** → gate deleted; the six accept shapes are now guards.
- **M14b** → extracted to `IsUnsupportedImplicitInterfaceIndexer` (`IV:720-723`), one definition, called from `IV:771` and `IV:824`. **P13 = KILLED.** The lockstep-drift risk for **#2915 is eliminated — #2915 now has exactly one line to change.**

The 3 survivors are inert, not gaps: **P16** escalated to 1064 Compiler / 6916 Core passing plus **136/136 byte-identical probes** (redundant with `TypeSymbol.cs:789`); **P19/P20** are source-unreachable (`prop V &T` doesn't parse; `prop V *T` → GS9998).

### SF3 shipped as option (a), and it is useful
`IV:640-661`, eight lines. Exhaustive sweep of **every** GS0504 across the grid: **38 distinct messages, zero with matching expected/actual.** All previously self-contradictory cases (`gi_listT`, `gi_dictT`, `gi_seqT`, `gi_actT`, `gi_nestcell`, `x_importedgeneric`, `inst_explicit_listT`, `genUv_QN_gi/gs`) now report **GS0187**. Pinned by a fact asserting both `Contains("GS0187")` and `DoesNotContain("GS0504")`; mutant P14 kills it.

Docs improved since round 2: GS0504 is at the **correct ordinal position in both files** (between GS0502 and GS0506), texts identical, severity matches `DiagnosticDescriptors.cs:377`.

### #2945 — accurate, correctly scoped, not deferral
Reproduces my `st_implicit_cov`/`st_explicit_cov` shapes verbatim, states the correct symptom (GS0397), cause and consequence, and is confirmed pre-existing on `e21677f6`. The real fix lives in the **emitter**; this PR keeps the strict-and-safe rule and adds two pins so nobody removes it by accident. `st_implicit_fnT` is the same root cause and worth a one-line mention when someone picks it up.

### Numbers
| SHA | Compiler total |
|---|---|
| `d4038f1d` | 3708 |
| `60e5b8bf` | 3742 |
| **`e21677f6`** | **3756** ← measured |
| `4eaba4ff` (PR) | **3772 passed / 0 failed** |

**Δ = +16**, matching the PR body, and corroborated by `Issue2888InterfacePropertyTypeMismatchTests.cs` containing exactly 16 `[Fact]` methods.

*Harness caveat:* the `origin/main` run built out-of-repo in `/tmp` reported 19 failed — SourceLink `Unable to locate repository with working directory`. The in-repo HEAD control had **0 failed**. Only the total (3756) was used.

Corpus stated as **0/148** ("2 files contain properties, 8 contain interfaces, none contain both").

### Merge state — zero PR-attributable conflicts
`merge-tree` run against both HEAD and `origin/main` so conflicts can be attributed: #2912, #2926, #2929 all clean. #2932's `ExpressionBinder.Access.MemberLookup.cs` conflict is **pre-existing** — this PR does not touch that file. #2926's round-2 `MethodBodyEmitter.cs` conflict was never attributable here and is now gone.

**#2915 still reproducible and still deliberately preserved** — both repro snippets unchanged across main and HEAD.

### Remaining items — all NIT, none blocking
| item | `file:line` |
|---|---|
| 3 unpinned arms, ~14 deletable lines | `StructSymbol.cs:2366, 2522, 2528` |
| `gi_tupT` GS0504 prints un-substituted `'(T, string)'` | `IV:748` |
| GS0187 fallback says "does not implement" for a declared member (same as `main`) | `IV:644` |
| PR body absolute counts stale after rebase (Core now 6916/1 at `914b42a7`); deltas still correct | — |

Everything cleared in round 2 re-verified unchanged at `df3e7ef7`. **No blocking issues.**

DavidObando:
## Duck round 4 — retraction of one round-3 NIT (post-merge)

Fleet rule changed after my round-3 review: **a survived mutant measures *coverage*, not *reachability*, and is never licence to delete code.** My round-3 report broke that rule in one place. Retracting it here so it does not get actioned by follow-up work in this file.

### RETRACTED — "delete the three erasure arms in `StructSymbol.cs`"

Round-3 report said:

> NIT (ponytail): `StructSymbol.cs:2364-2377, 2522-2532` — ~14 lines of erasure prologue/arms that are behaviourally inert (P16) or source-unreachable (P19, P20). **Deletable with zero behaviour change.**

**That claim is wrong. Do not delete these lines.** They are at `src/Core/CodeAnalysis/Symbols/StructSymbol.cs:2367` (`NullabilityAnnotatedTypeSymbol`), `:2523` (`ByRefTypeSymbol`), `:2529` (`PointerTypeSymbol`) in merged `main`.

**Structural counter-proof for `:2367`.** `NullabilityAnnotatedTypeSymbol` appears in exactly **one** arm of `SubstituteTypeForConstruction` — that one. Remove it and an annotated type matches **no** later arm (`TypeParameterSymbol`, `StructSymbol`, `InterfaceSymbol`, `ImportedTypeSymbol`, `NullableTypeSymbol`, `Slice`, `Array`, `Map`, `ByRef`, `Pointer`, `Delegate`, `FunctionType`) and falls through to the terminal `return type` — **returned un-substituted, its generic arguments never remapped.** For `List[T?]` at `I[string]` the slot would stay `List[T?]` instead of becoming `List[string]`.

**Why my evidence missed it.** My 136-probe grid showed zero difference because the only consumer I exercised was gate 1, and `AreRuntimeEquivalentIgnoringReferenceNullability` calls `StripReferenceNullabilityCore` (`src/Core/CodeAnalysis/Symbols/TypeSymbol.cs:789`) on **both** sides, which unwraps the annotation inside the comparer and masks the difference. Black-box identity across my probe set is not a structural proof. Textbook survival-is-not-reachability.

`:2523`/`:2529` retracted on the same grounds, more weakly: I proved `prop V &T` does not parse (GS0005) and `prop V *T` → GS9998 **today, via one caller**. `StructSymbol.SubstituteTypeParameters(..., eraseReferenceNullability: true)` is a shared entry point; a reachability claim about one caller is not a claim about the function.

**Live risk, not hypothetical.** #2915 is OPEN and **PR #2968** ("Unify implicit interface indexer implementation rules") is in flight in `DeclarationBinder.InterfaceVerification.cs`, and will call `GetInterfacePropertySlotType` (`:700`) on indexer **value** types — a new caller of this exact erasure path. Deleting the arms would silently stop substituting annotated indexer value types.

### Confidence downgrade on my round-3 mutation table

Directional rule — stale binaries yield **survivals, never false kills**:

- **17 kills stand.** Lower bound; unaffected.
- **3 survivals (P16 `:2367`, P19 `:2523`, P20 `:2529`) are UNPROVEN.** I have no round-trip `X → mutant → X` md5 for them, which is precisely the evidence a *survival* requires. My runner never used `--no-build` (staleness is a `--no-build` artifact, so risk is low), but it gated on `grep -cE "error CS"` rather than on **exit code** — that misses a non-CS build failure such as the `Gsharp.Extensions` parallel race (#2974) corrupting `GSharp.Core.dll`. Treat those three rows as unmeasured.

The four round-2 survivors (M7/M8/M10/M14b) are unaffected: all four became **kills**, and kills are the safe direction.

### Baseline number — method-tainted, and base has moved

- My reported **`origin/main` @ `e21677f6` Compiler = 3756** came from a `git archive` tree (now a forbidden method — no `.git`). The *total* does not depend on pass/fail so it is probably still right, but treat it as tainted rather than settled.
- The **19 failures** in that run: I attributed them to SourceLink. Better explanation is the `Gsharp.Extensions` build-closure gap → `SampleConformanceTests.cs:126` → **exactly 19** golden-sample tests. Conclusion unchanged (environmental; totals only), stated cause was wrong.
- **`+16 Compiler / +1 Core` was measured at base `e21677f6`.** `main` is now `a6b1561f2`, six merges later (#2950, #2932, #2955, #2952, #2973, #2961). Anyone quoting those absolutes must re-measure.
- Mixed-base disclosure on my own table: round-3 compared `d304a53b` results (base `e21677f6`) against `df3e7ef7` results (base `914b42a7`). I proved the PR diff was content-identical across that rebase (11 context-only lines) **before** comparing, which is the correct mitigation — but the reader should know the bases differed.

### Confirmed sound, no change

- Test oracle is adequate: `AssertEmitsAndLoads` (`test/Compiler.Tests/Emit/Issue2888InterfacePropertyTypeMismatchTests.cs:674`) does **not** rely on `IlVerifier.Verify` alone — it also runs `Assembly.Load(File.ReadAllBytes(outputPath)).GetTypes()` at `:696`, which is the oracle that actually catches the #2888 `TypeLoadException`.
- **No unbounded `WaitForExit()`** — no child process at all; `Compile()` is in-process via `StringWriter`, and `CompileCSharpFixture` uses Roslyn in-process. The #2934 bug shape does not apply here.
- All correctness findings stand: blocker fixed (5 value-nullable over-fire cases restored to LOAD-OK), 0 LOAD-FAIL across 133 probes, no under-fire, no new over-fire, 17/20 semantic branches pinned.
- Follow-up filed by this PR has landed: **`b78bb1f15` "Unify static property nullable erasure (#2955)"** closes the #2945 static-virtual asymmetry I raised as SF2.

**Net: verdict on the merged content unchanged. One recommendation withdrawn — keep `StructSymbol.cs:2367`, `:2523`, `:2529`.**

Issue #2910, opened 2026-07-30
Diagnostics render imported constructed generic types as raw assembly-qualified CLR names


DavidObando:
Found during review of PR #2892 (round 4 duck).

## Summary

When a diagnostic mentions a member of an **imported** constructed generic interface, the type is rendered as a raw assembly-qualified CLR name instead of G# syntax.

## Repro

Given a G#-compiled library `GLib2` exporting `interface IGBox[T] { prop V T { get; init; } }`, a consumer that explicitly implements the member with `set` instead of `init` reports `GS0502` as:

```
interface property 'GLib2.IGBox`1[[System.Int32, System.Private.CoreLib, Version=10.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].V'
```

Expected: `GLib2.IGBox[int32].V`.

## Notes

- Types declared **in the current compilation** render correctly (`IB[int32].V`), so this is specific to the imported/metadata-backed path.
- Verified byte-identical on `1e08def8` (before PR #2892's encoder change) so this is **pre-existing**, not introduced by #2892.
- Related to the display consolidation from #2851 / PR #2886, which routed diagnostics, hover, inlay hints and the REPL through `SymbolDisplay`. The imported-symbol path apparently bypasses it or falls back to `Type.AssemblyQualifiedName`.
- Likely shares a root cause with #2898 (constructed arguments lost for some symbol kinds).

## Suggested starting point

`src/Core/CodeAnalysis/Symbols/Display/SymbolDisplay.cs` `FormatType`, and whatever imported-symbol wrapper backs metadata-loaded constructed generics. Check whether `GS0502`'s formatting arguments go through `ToTypeDisplayString` at all.

PR #2912, opened 2026-07-30
Fix generic type display and support constructed nested enums

### LINKED ISSUE #2898: SymbolDisplay hover headers use CLR angle brackets and omit constructed type arguments
## Summary

`SymbolDisplay.AppendTypeParameters` renders generic declaration parameters with angle brackets, while G# source and type positions use square brackets. Hover headers also describe a constructed source type from its definition and omit the construction arguments.

Examples:

- Hover declaration header: `class P.Box<T>`
- Type position: `Box[T]`
- Hover on a constructed `Box[int32]` symbol: `class P.Box { var v int32 }` instead of a header that identifies `Box[int32]`

## Evidence

`src/Core/CodeAnalysis/Symbols/Display/SymbolDisplay.cs`, `AppendTypeParameters` (currently lines 469-489), emits `<` and `>`. This behavior predates PR #2886; `git diff origin/main...5bcf0ccc -- src/Core/CodeAnalysis/Symbols/Display/SymbolDisplay.cs` shows no changes to `AppendTypeParameters`.

PR #2886 consolidates type-position formatting on `SymbolDisplay.ToTypeDisplayString`, making the existing header/type-position inconsistency more visible but not causing it.

## Expected

Hover and other declaration descriptors should use G# bracket syntax consistently and preserve constructed arguments when describing constructed source symbols, without changing metadata-facing names.

### LINKED ISSUE #2930: Imported symbolic constructor cache ignores lambda remaps
## Reproduction

A generic promoted lambda constructs an imported generic whose argument is a nested enum over an enclosing method type parameter:

```gsharp
struct Outer[T] { enum Color { Red } }
func Remap[A, B](c Outer[B].Color) int32 {
    var outer = List[Outer[B].Color]()
    let f (Outer[B].Color) -> int32 = (value Outer[B].Color) -> {
        var inner = List[Outer[B].Color]()
        inner.Add(value)
        return int32(inner[0])
    }
    return f(outer[0])
}
```

## Failure

The lambda has one promoted generic parameter (`MVAR(0)`), but its cached `List[Outer[B].Color]::.ctor` parent TypeSpec is reused from the enclosing method and encodes `MVAR(1)`. ILVerify throws `IndexOutOfRangeException`; invoking the lambda throws `BadImageFormatException`.

## Cause

`MetadataTokenCache.CtorRefSymbolKey` keys symbolic imported constructor MemberRefs by constructor and type arguments only, omitting `ActiveIteratorStateMachineRemap` and `ActiveLambdaMethodTypeParamRemap`.

## Fix

Include both remap identities in the cache key. Regression coverage is being added in PR #2912.

### LINKED ISSUE #2946: Support nested enums in constructed generic outer types
## Capability

Support a user enum nested inside a generic outer type as a distinct constructed type at each enclosing construction, including metadata emission and cross-assembly consumption.

Examples include:

```gsharp
struct Outer[T] { enum Color { Red, Green, Blue } }
var i Outer[int32].Color
var s Outer[string].Color
```

Required behavior:

- `Outer[int32].Color` and `Outer[string].Color` retain distinct symbol/runtime identity.
- Signatures, nullable values, boxing, fields, generic containers, closures, interfaces, and deeper nesting preserve concrete enclosing arguments.
- A second G# assembly can reference the emitted library, load it with `Assembly.Load`, enumerate types, and invoke members without invalid `VAR`/`MVAR` metadata.

## Evidence

Round-3 review of PR #2912 exercised twelve shapes. Shapes s1/s2/s3/s4/s9/s10/s11/s12 are rejected on `main` with GS0149/GS0158 but compile, load, and run on the PR head. Shapes s5/s6/s8 expose separate pre-existing nested-struct limits.

Review evidence: https://github.com/DavidObando/gsharp/pull/2912#issuecomment-5142700741

PR #2912 implements and tests this capability alongside the smaller #2898 display fix.

DavidObando:
Fixes #2898.
Fixes #2946.
Fixes #2930.

## Scope

This PR contains two pieces of work:

1. **#2898 display fix.** Four production-line edits close the original issue: aggregate headers route through `FormatType(aggregate)`, enum headers route through `FormatType`, and generic delimiters use `[` / `]` instead of `<` / `>`.
2. **#2946 nested-enum capability.** Remaining binder/symbol/emit work supports enums nested in generic outer types as distinct constructed types, including cross-assembly metadata. Main rejects matrix shapes s1/s2/s3/s4/s9/s10/s11/s12 with GS0149/GS0158; this branch compiles, loads, and runs them.

Symbolic imported-constructor cache remap correctness is tracked and fixed by #2930.

## Main changes

- Preserve constructed symbols for hover; normalize struct/enum/interface symbols only in reference indexing.
- Reuse canonical `TypeSymbol.GetWrappedTypes` for declaration-clause mapping. Helper remains internal; Core grants narrow `InternalsVisibleTo("GSharp.LanguageServer")` access. No public API widening.
- Give `EnumSymbol` definition/construction identity and thread enclosing arguments through binding, substitution, imported lookup, metadata signatures, closures, and state machines. Method binding now seeds the complete outermost-to-innermost generic owner chain, so enums nested below two generic levels close over every lexical parameter.
- Encode nested enum TypeSpecs with concrete enclosing arguments. Open self-instantiation is now explicit only at the enum TypeDef's own literal fields; an open enum escaping any other context fails instead of silently writing invalid metadata.
- Drop unpinned imported-only `GS0507`; its three sites had no genuine repro, and the actual user-declared escape is fixed at array-literal binding.
- Key enum and imported-constructor TypeSpec/MemberRef caches by iterator/lambda remap identity.
- Deny type-parameter rename at shared `CanRename`; hover on type-parameter uses remains available.
- Add enum branches to `AnyTypeParameter` and `AreRuntimeEquivalentIgnoringReferenceNullability`. #2911 does not overlap this method; see substitution sequencing note below.
- Resolve nested-type chains into typed `StructSymbol` / `EnumSymbol` outputs; unsupported kinds return `false` before the caller can receive them.
- #2911 is now in the base. The enum callback in `StructSymbol.SubstituteTypeForConstruction` routes through its local `SubstituteNested`, preserving `eraseReferenceNullability` recursively on the pinned static-interface route. The analogous instance-interface nested-enum shape remains unsupported and is not claimed here.
- Add exact-range rename tests and real cross-assembly `Assembly.Load`/`GetTypes`/invoke coverage.

#2929 is now in the base; the final rebases composed its substitution changes cleanly with this PR. Open PR #2903 is the tight remaining overlap (`MetadataTokenCache.cs`, five-line hunk gap) and must be reconciled if it lands first.

## Cross-assembly evidence

Library and consumer compile separately. Tests run ILVerify, real `Assembly.Load`, `GetTypes`, reflection, and invocation.

| Field | Signature blob |
|---|---|
| `Outer[int32].Color c` | `061511110108` |
| `Outer[string].Color d` | `06151111010E` |

Blobs differ and contain concrete I4/STRING arguments. Removing imported closure yields `06151111011300` for both; concrete imported closure remains pinned by load, invocation, and blob assertions.

## Diagnostics and display

- Nested enum diagnostics now preserve containing qualification. This intentionally changes non-generic `Flat` to `Plain.Flat` too.
- #2947 interaction: the erased `glib.Outer[object].Color` leak manifests only when this PR activates the imported constructed nested-enum path; `main` alone emits `Outer[T].Color`. This PR discloses but does not widen into `ImportedTypeSymbol` enclosing-argument support.
- #2910 remains open and is not fixed here: its imported generic-interface `GS0502` diagnostic text is unchanged.
- #2916 remains open and is not fixed here: nested structs still retain inherited metadata arity.

## Tests added: classification

All added facts are classified below, including round-4 array pins and the #2911 composition pin.

| Test | Classification |
|---|---|
| `DeclarationHeaders_UseSquareBracketsAndConstructedArguments` | #2898 display pin |
| `ConstructedNestedEnums_PreserveIdentityDisplayMemberTypeAndValue` | enum identity, substitution, type-parameter walk, runtime-equivalence pin |
| `ComposedTypeSpellings_AreDistinctExceptCanonicalResidualCollisions` | display collision guard |
| `ReferenceResolverDisposal_ClearsConstructedNestedEnumCache` | cache lifecycle pin |
| `ConstructNested_NormalizesEmptyArgumentsAndConstructedDefinitions` | D27 empty-vector and D28 root-definition normalization pins |
| `NestedEnumSubstitution_PropagatesReferenceNullabilityErasure` | direct #2911 composition pin for recursive nullable-reference erasure |
| `StaticInterfacePropertyNestedEnum_PropagatesReferenceNullabilityErasure` | reachable #2955 static-member-access composition pin with load and invocation |
| `NestedEnumConstructedThroughGenericOuter_LoadsWithDistinctClosedSignatures` | same-assembly metadata, lambda/iterator remap, D30 pin |
| `DeepNestedEnum_LoadsAcrossGenericInterfaceAndDictionarySignatures` | deep/interface/container emit pin |
| `ImportedNestedEnum_LoadsWithConcreteDistinctEnclosingArguments` | cross-assembly concrete-argument metadata pin |
| `BareNestedEnumEmptySliceLiteral_ClosesOverCurrentTypeParameters` | bare `[]Color{}` execution pin |
| `BareNestedEnumPopulatedSliceLiteral_ClosesOverCurrentTypeParameters` | bare `[]Color{Color.X}` execution pin |
| `AlreadyClosedNestedEnumSliceLiteral_RemainsClosed` | E7 idempotency guard: an explicit `Outer[int32].Color` must not be re-closed over `Outer[T]` |
| `NestedEnumTwoGenericLevelsDeep_LocalExpressionClosesAllEnclosingParameters` | q1 two-generic-level local expression blocker pin |
| `NestedEnumTwoGenericLevelsDeep_SliceLiteralClosesAllEnclosingParameters` | p5 two-generic-level slice literal blocker pin |
| `ComputeHover_DeclarationHeadersUseSourceGenericSyntaxAndConstruction` | #2898 hover pin |
| `ComputeHover_DistinguishesNestedEnumConstructions` | constructed-enum hover identity pin |
| `ComputeHover_PreservesConstructedTypesAcrossDeclarationTypeClauses` | wrapper traversal pin |
| `ComputeHover_PreservesConstructedTypesInsideCompositeTypeClauses` | composite-wrapper traversal pin |
| `ComputeRename_ConstructedUserTypeRenamesEveryConstruction` | struct normalization guard |
| `ComputeRename_ConstructedNestedEnumRenamesEveryConstruction` | enum normalization guard |
| `ComputeRename_ConstructedInterfaceFromUseRenamesExactRanges` | interface normalization branch pin |
| `ComputeRename_ImportedGenericArgumentDoesNotRenameImportedType` | imported guard pin |
| `ComputeRename_ComposedTypeArgumentsProduceExactCrossFileRanges` | identifier arm + wrapper/range/multi-file pin |
| `ComputeRename_RejectsTypeParameterUse` | round-3 source-corruption blocker pin |

## Mutation evidence

- Type-parameter `CanRename` token removed: blocker test returns one corrupting edit instead of null.
- Enum remap key reduced to `(sym, null, null)`: ILVerify throws `IndexOutOfRangeException` in promoted lambda coverage.
- D27/D28 were reachable through the public helper: empty arguments constructed a duplicate symbol, and passing an existing construction made `Definition` point at that construction. Both guards are restored. Removing either independently fails `ConstructNested_NormalizesEmptyArgumentsAndConstructedDefinitions`.
- D14 remains deleted. `TryResolveNestedTypeChainUnderReceiver` now exposes separate typed `StructSymbol` / `EnumSymbol` outputs; every other nested kind returns `false`, so the caller has no third-kind switch arm or cast.
- Bare nested enum array closure removed: two execution pins fail independently (`[]Color{}` with GS9998; `[]Color{Color.X}` with GS0156). Both load and execute after closure over current type parameters.
- `GS0507` remains removed. #2912 will not resurrect it; treat the ID as retired/quarantined rather than free.
- `AnyTypeParameter` enum arm removed: structural-walker assertion fails.
- Runtime-equivalence enum arm removed: nullable-reference-equivalent construction assertion fails.
- Enum substitution callback reverted to the old 3-argument recursion: direct Core pin `NestedEnumSubstitution_PropagatesReferenceNullabilityErasure` goes 1/1 green → 0/1 red (`expected string`, `actual string?`); reachable Compiler pin `StaticInterfacePropertyNestedEnum_PropagatesReferenceNullabilityErasure` goes 1/1 green → 0/1 red with GS0156 (`Outer[string?].Color` → `Outer[string].Color`).
- Reverting generic-owner scope seeding to the immediate owner makes both deep-nesting pins go 2/2 green → 0/2 red with `GS9998`. Full-solution hashes round-trip exactly: fixed `590ed5a37e750e1c2369a672cc0cacf3` → mutant `dd162f63a5d2e6c79edf988396243d2b` → fixed `590ed5a37e750e1c2369a672cc0cacf3`.
- Removing the already-closed E7 guard makes the three focused pins go 3/3 green → 2/3, with `AlreadyClosedNestedEnumSliceLiteral_RemainsClosed` failing `GS0156` (`Outer[int32].Color` → `Outer[T].Color`).
- Restored Issue2898 emit control is 9/9 green.

`userEnumTypeSpecCache` read remains deliberate deduplication only. Latest guarded X → mutant → X audit ran at `a3e98d1b6f` on base `f70626452d`: serial full-solution builds exited 0; all four Core DLL copies were nonzero and propagated identically; control `0a1698e7b1966a80c1e016d95d6310d0` → mutant `93b955960ffb51bc86e6187aaaca6335` → exact control `0a1698e7b1966a80c1e016d95d6310d0`. Full Compiler ran without `--no-build` and stayed 3,955/3,955 green on the mutant. Survival is coverage evidence for deduplication, not evidence that the cache read is dead.

## Verification

Pushed head `3da37d5e408f70834bedb1f107b972155872fd3d` is rebased onto live `origin/main` `fdba684b600297888d7b47e13e6e53a97ec96d25`.

- Exact q1 differential: live main `R() = 2` → stale PR head `e03e8fa54` `GS9998` → pushed head `R() = 2`.
- Exact p5 slice-literal differential: live main `R() = 2` → stale PR head `e03e8fa54` `GS9998` → pushed head `R() = 2`.
- Issue2898 Compiler tests: 9/9 passed without `--no-build`.
- Scratch merge build (`git merge origin/main`, then `dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true`): 0 warnings, 0 errors.
- Zero-byte gate checked all 40 `GSharp.Core.dll` copies. Five ordinary copies matched MD5 `3271368f1c2b27c471da21e2e67e9d06`; staging copies were intentionally excluded from equality.
- GitHub CI passed with authoritative rollup: 8 SUCCESS + 2 SKIPPED.

`GS0507` will not be resurrected by #2912 and remains retired/quarantined.

Corpus result is non-regression evidence only: none of 148 corpus samples contain a nested enum in a generic outer, so it is not feature evidence.


DavidObando:
## Rubber-duck review — round 3 (Opus 5), head `8dc1e8f7`

**Verdict: NOT MERGEABLE.** One BLOCKER with a one-token fix, built and verified. Five SHOULD-FIX, five NITs.

### BLOCKER 1 — renaming a type parameter corrupts source
The new `MapTypeClauseReference` in `src/LanguageServer/SemanticLookup.cs` (~:851) maps type-clause identifier tokens to resolved `TypeSymbol`s. When that symbol is a `TypeParameterSymbol`, `RenameComputer.ComputeRename` (`HoverComputer.cs:1065`) now builds an edit set covering **uses only** — the declaration lives in the generic parameter list, not in a type clause. Rename rewrites the uses and leaves the declaration. Silent source corruption. On `origin/main` this was a safe no-op in every case.

| case | main edits | HEAD edits | result |
|---|---:|---:|---|
| `class Box[T] { var Value T }`, from field use | 0 | **1** | `class Box[T] { var Value ZZ }` |
| `func Id[T](value T) T`, from param use | 0 | **2** | broken |
| `func Id[T](value T) T`, from return use | 0 | **2** | broken |
| `class C { func M[T](v T) T }`, from param use | 0 | **2** | broken |
| declaration sites (3 cases) | 0 | 0 | safe |

Corrupted output re-parsed with `gsc` gives `GS0113: Type 'ZZ' doesn't exist.` Prepare-rename does not save you: `LspServer.cs:1513` gates on `CanRename`, which returns **true** for `TypeParameterSymbol` today.

**Fix — one token**, at `SemanticLookup.cs:423`:
```csharp
return symbol is not null and not ImportedTypeSymbol and not ImportedClassSymbol and not ImportedFunctionSymbol and not TypeParameterSymbol
```
`CanRename` is the purpose-built deny-gate and is already consulted by all three dangerous entry points (prepareRename `LspServer.cs:1513`, linked-editing `:1616`, rename `HoverComputer.cs:1076`). Guarding inside `MapTypeClauseReference` instead would also kill this PR's **new** hover on type-parameter uses (HEAD gives `type T`; main gave `<null>`) — worth keeping. Verified at HEAD: all 7 type-param cases → 0 edits, `LanguageServer.Tests` **461/461**, build clean.

### Scope: four lines fix #2898; the rest is an unattributed feature (SHOULD-FIX 1)
Applied to pristine `origin/main` (`d4038f1d`), **only** these edits in `SymbolDisplay.cs`: `AppendAggregate` (~:327) routing through `FormatType(aggregate)`; `AppendEnum` (~:358) `.Name` → `FormatType`; `AppendTypeParameters` (:468–489) `<`/`>` → `[`/`]`.

All three of issue #2898's literal acceptance assertions pass. Regression: **Core 6914 passed / 0 failed**, **LSP 452 passed / 0 failed**.

The remaining ~1191 lines are the nested-enum-in-generic-outer feature. That work is **genuine and valuable** — on `main` nearly every shape is a hard error (GS0149/GS0158) — but it has no issue of its own, while its smaller sibling got #2930. Regressions in 1191 lines of emit/binder machinery will trace back to a PR titled "SymbolDisplay hover headers".

### Cross-assembly correctness — round-2 blocker confirmed fixed
Reproduced the round-2 defect on `0e257ac9`: `List[Outer[int32].Color]` compiled clean, loaded, then threw `BadImageFormatException` on invoke, with zero compile diagnostics. At HEAD the same shape returns `6`. Harness sanity-checked against that known-bad control; `IlVerifier` never trusted alone. Decoded the blobs byte by byte (`061511110108` = FIELD·GENERICINST·VALUETYPE·TypeRef(row 4)·1 arg·I4) and found **no `0x13` at any cross-assembly reference site** in any shape I could construct.

Twelve-shape matrix via real `Assembly.Load` + `GetTypes()` + invoke: s1–s4, s9–s12 all rejected on `main` and correct at HEAD. s5/s6/s8 (deeper nesting, nested **structs**) are pre-existing limits, not regressions. Roslyn consumer check confirms the arity rule is honored for enums and violated for structs.

### #2916 and #2910 both remain open
**#2916 NOT FIXED, reproduces definitively** — a real Roslyn consumer gives `CS0426` on `Outer<int>.Inner`; the metadata dump shows `Outer\`1+Inner\`1`. The nested-struct path (`TypeDefEmitter.cs` ~:751) is untouched. This PR's enum approach is the right template for it. **#2910 NOT FIXED** — built the exact repro; the GS0502 message text is byte-identical on main and HEAD.

### Coverage — the real gap is branch granularity
32 branch-level mutations, all build-green and semantically inert, unmutated control before each batch: **27 killed / 5 survived**.

| survivor | site | evidence |
|---|---|---|
| D14 | `_ => new BoundErrorExpression(null)`, `ExpressionBinder.Access.Accessor.cs` | → `throw`: full Core 6917 + full Compiler 3711 green. **Dead arm.** |
| D15 | `userEnumTypeSpecCache` read | dedup only, inert by design — acceptable |
| D27 | `ConstructNested` empty-args early return | full Core + full Compiler green |
| D28 | `ConstructNested` `Definition ?? nestedDefinition` | full Core + full Compiler green |
| D30 | `GetUserEnumRemapKey` remap components → `(sym, null, null)` | full Compiler green — **this is round-3 commit `8dc1e8f7`'s own enum half** |

D30 is the one that matters: half of the commit added this round is unverified, while its sibling (`CtorRefSymbolKey`) is pinned.

The pins/guards/mutations "contradiction" I was asked about is not real — 46 mutations, 7 pins and 5 guards measure three different axes and reconcile fine. The actual gap is **16 new `[Fact]`s vs 12 classified**, and hunk-granularity hiding the 5 branches above.

### Remaining
- **SHOULD-FIX 2** — `TypeSymbol.AnyTypeParameter` (~:351) has no `EnumSymbol` arm, so `ContainsTypeParameter(Outer[T].Color)` returns **false**. It is the canonical structural walker behind `RequiresSymbolicProjection` (22 call sites). Masked today only because that predicate also ORs `ContainsSameCompilationUserType`. The enum arm went into the *other* walker, `CollectReferencedTypeParameters`.
- **SHOULD-FIX 3** — `AreRuntimeEquivalentIgnoringReferenceNullability` (~:703) has explicit `StructSymbol`/`InterfaceSymbol`/`DelegateTypeSymbol` arms but no `EnumSymbol`; two distinct constructed nested enums return `false` by accident rather than by design. **PR #2911 edits this exact method** — coordinate.
- **SHOULD-FIX 4** — `ResolveUserEnumTypeSpecArguments` still falls back to open self-instantiation with no diagnostic. It is load-bearing, but the construct that silently produced an unloadable assembly in round 2 remains reachable in principle.
- **SHOULD-FIX 5** — add a test emitting one nested enum's TypeSpec under two different lambda/iterator remaps (kills D30).
- **NIT 1** delete or pin the dead D14 arm · **NIT 2** undocumented diagnostic drift `Flat` → `Plain.Flat`, affecting non-generic nested enums too (an improvement, but unmentioned) · **NIT 3** corpus 139/139 is null signal — **0 of 148** samples contain this shape · **NIT 4** this adds a fourth near-verbatim enum-substitution block, the duplication #2929 is removing · **NIT 5** `GS0156` leaks the erased projection `glib.Outer[object].Color` into user-visible text — a #2898-class defect, newly reachable.

### Corrections to my briefing
1. **"Broad `AssemblyInfo.cs` friend access deleted" is not a win.** That file never existed on `main`; this PR added it (`0e257ac9`) then removed it (`630fceaf`). Net zero. No pre-existing `InternalsVisibleTo` was removed. The real net API change is `TypeSymbol.GetWrappedTypes` going `private` → `public` — a narrow `InternalsVisibleTo("GSharp.LanguageServer")` would be smaller. There are no public-API approval tests in the repo, so nothing catches this. It is the **only** visibility widening in the PR.
2. **#2923 does not edit `SubstituteType`.** Its `Binder.cs` edit is ~`:2921` and its `MetadataTokenCache.cs` change is a doc comment. No same-method collision.
3. **The author's four suite counts are all correct.** True `main` baselines measured directly: Core **6913 + 1 skipped**, Compiler **3708**, LSP **452**, Interpreter **453**; HEAD = 6917 / 3711 / 461 / 453, delta exactly the 16 tests added. The 19 main-side Compiler failures were environmental (load average ~12, ~82 concurrent `dotnet` processes from sibling agents); clean reruns green.

`git merge-tree` against #2908, #2911, #2917, #2923, #2926, #2929, #2932, #2935 → **8/8 clean**, and the hot spots reviewed semantically: #2929 rewrites different `Binder.cs` methods and preserves the `EnumSymbol → typeof(int)` erasure that shape s9 depends on.

DavidObando:
## Round 4 review — NOT MERGEABLE (2 blockers, one root cause, 4-line fix proven)

Head reviewed: **`cb29a5f0`** ("Restore nested enum construction guards"), base `e21677f6`, merge-base == `main`.

### BLOCKERS 1 + 2 — the SF4 fallback removal regresses code that works on `origin/main`

Root cause is one line: `src/Core/CodeAnalysis/Binding/ExpressionBinder.Literals.cs:2101`.

```csharp
if (syntax.HasNestedElementTypeClause)
    elementType = bindTypeClause(syntax.ElementTypeClause);      // :2093 — CLOSES nested enum (Binder.cs:3172)
else
    elementType = lookupType(syntax.ElementTypeIdentifier.Text); // :2101 — BYPASSES the close
```

`[]Outer[T].Color{…}` takes the type-clause branch and is closed. **`[]Color{…}` takes the bare-identifier branch and is not.** Before SF4, the open enum reached the emitter and the open self-instantiation fallback rescued it. With the fallback deleted, the bare branch dies.

**BLOCKER 1 — internal compiler error `GS9998`:**
```gs
struct Outer[T] {
    public enum Color { Red = 1, Green = 2, Blue = 3 }
    public func A() []Color { return []Color{} }
}
```
| | result |
|---|---|
| `e21677f6` | COMPILED, `Assembly.Load` OK, `R()` = 0 |
| `cb29a5f0` | `error GS9998: InvalidOperationException: Open nested enum 'Color' escaped its generic context` |

`GS9998` is the emitter's internal-error catch-all — the user sees a crash, not a diagnostic.

**BLOCKER 2 — self-to-self conversion failure:**
```
error GS0156: Cannot convert type 'Outer[T].Color' to 'Outer[T].Color'
```
One operand is the open definition, the other the closed construction; `SymbolDisplay.FormatType` renders both via `ContainingType`. Unactionable for a user.

Both verified to compile **and execute** on `main` via real `Assembly.Load` + `GetTypes()` + invoke: `b1 R()=2`, `b2 R()=3`, `c1 R()=0`, `c4 R()=2`, `a6 R()=3`. Control `c2.gs` using fully-qualified `[]Outer[T].Color{…}` compiles at head — **only the bare identifier breaks**, confirming the diagnosis.

**Fix — 4 lines**, mirroring existing `Binder.cs:3172-3177`, inserted after `:2105`:
```csharp
if (elementType is EnumSymbol bareElementEnum)
{
    elementType = CloseNestedEnumOverCurrentTypeParameters(bareElementEnum);
}
```
`CloseNestedEnumOverCurrentTypeParameters` is already a private member of the same partial class (`ExpressionBinder.Access.Accessor.cs:752`). No new machinery. Proven at `cb29a5f0` + fix: all 9 repros compile and execute with values identical to `origin/main`; **Core 6918 +1 skip / LSP 462 / Compiler 3759, 0 failed.**

`BindStackAllocExpression:2164` has the same `lookupType(...)` shape but yields `GS0125` on both sides — no divergence, leave alone.

### SHOULD-FIX 1 — `GS0507` is unpinned and guards the wrong path
Mutation disabling the `closedType.IsEnum && ContainsGenericParameters` arm → **0 failures**. Worse, all three report sites (`ExpressionBinder.Access.Accessor.cs:712, 723, 732`) sit inside `CloseImportedNestedType` — the **imported** path — while the real escape (BLOCKER 1) happens on the **user-declared** path and yields `GS9998`. Guard and failure mode don't line up. With the 4-line fix, that escape becomes unreachable from this shape: either pin `GS0507` with a real imported repro, or drop it.

### SHOULD-FIX 2 — semantic merge conflict with #2911 that `git merge-tree` reports as CLEAN
The collision is **not** `AreRuntimeEquivalentIgnoringReferenceNullability` — #2911 doesn't touch `TypeSymbol.cs` at all. It is `StructSymbol.SubstituteTypeForConstruction`.

#2911 adds an `eraseReferenceNullability` flag and converts every recursive call to a local `SubstituteNested` closure. This PR inserts a new enum arm using the **old 3-arg form**. Real merge performed:
```csharp
    var newEnclosing = EnumSymbol.SubstituteEnclosingArguments(
        nestedEnum,
        t => SubstituteTypeForConstruction(t, subst, mapClrType));   // flag SILENTLY DROPS to false
```
This reintroduces nullability over-gating for nested enums in generic outers, and **neither PR's tests catch it**. Fix is one word: `SubstituteNested(t)`. Noted on both PRs.

### SHOULD-FIX 3 — `src/Core/Core.csproj` CRLF churn still present at `cb29a5f0`
Whole-file rewrite; the real change is one `InternalsVisibleTo` line.

### Resolved this round ✅
- **D27/D28 restored AND pinned** by the author. Mutation-verified independently: `IsDefaultOrEmpty`→`IsDefault` = 1 fail; `Definition ?? nestedDefinition`→`nestedDefinition` = 1 fail. Reachability re-checked anyway — all 6 in-tree callers pre-normalize, so unreachable today, but `ConstructNested` is `public` and #2946 adds callers. Restoring was correct. D14 removal confirmed fine.
- **Bonus:** `cb29a5f0` also removed an unchecked `(StructSymbol)innerReceiver` cast (`InvalidCastException` risk) introduced by `ac0a07bc`.
- **Legitimate `!0` survives** — cross-assembly `struct W[U] { var c Outer[U].Color }` loads and runs; field signature `06 15 11 11 01 13 00` still carries `ELEMENT_TYPE_VAR !0`, now sourced from `EnclosingTypeArguments` rather than the deleted fallback.
- **Blocker fix verified at altitude** — 7/7 type-parameter rename cases → `edits=0`; `Box[T]`/`Id[T]` re-parse with no `GS0113`; **type-param hover preserved** (`type T`); all three gates consult `CanRename`; mutation removing `and not TypeParameterSymbol` → LSP 1 fail, pinned.
- **SF2/SF3 both correct and pinned.** `ContainsTypeParameter(Outer[T].Color)` = True (was False); `AreRuntimeEquivalent(Outer[int32],Outer[string])` = False. Both mutations produce exactly 1 failure.

### True `e21677f6` baselines (measured)
| suite | `e21677f6` | `cb29a5f0` | Δ |
|---|---|---|---|
| Core | **6913** + 1 skip | 6918 + 1 skip | +5 |
| Compiler | **3756** | 3759 | +3 |
| LSP | **452** | 462 | +10 |
| Interpreter | **468** | 468 | 0 |

**Corpus zero restated: 0 of 148 samples contain a nested enum in a generic outer** (only 3 mention `enum` at all). "Corpus identical" is null-signal here.

### #2916 and #2910 still NOT fixed ✅ (holds after rebase)
Metadata dump of emitted `glib.dll`: enum path correct (`.Outer\`1+Color gp=1`), **struct path still wrong** (`.Outer\`1+Inner\`1 gp=1`, `.Gen\`1+MidG\`2 gp=2`). Roslyn consumer still reports `CS0426` for `Inner`, `MidG<>`, `L2`. #2910's GS0502 text is byte-identical across both. Both correctly stay open.

### Cross-PR
`merge-tree` rc=0 vs `origin/main`, #2908, #2911, #2926, #2929, #2932, #2950. #2929 shares `Binder.cs` but hunks are disjoint — note it consolidates three projection blocks into `ProjectGenericArgument` while this PR adds a **fourth** copy; sequence #2929 second or fold in. Diagnostic codes distinct (GS0503/GS0504/GS0507).

DavidObando:
## Duck review — round 5 — **NOT MERGEABLE** (1 BLOCKER)

Head reviewed **`e03e8fa54`**. Verified against **`a6b1561f2`** (current `main`) as well as `8dddbbf2` (this branch's merge-base). PR is now 1 commit behind `main`.

Build gates on every probe: **exit code** → **zero-byte** (`stat -f %z > 0`) → **propagation** (md5 agreement across `out/bin/Release/{Compiler,Core,Interpreter}.Tests/GSharp.Core.dll`). Solution-built with `-m:1`. All emit conclusions via real `Assembly.LoadFrom` + `GetTypes()` + invoke, never `IlVerifier` alone.

---

### Confirmed closed

| finding | status | evidence |
|---|---|---|
| **B1** bare `[]Color{…}` bypass | **CLOSED** | fix at `ExpressionBinder.Literals.cs:2109-2112`, placed after both branches — better than proposed. `p1` empty `R()=0`, `p2` populated `R()=4`, `p3` `[3]Color` `R()=3`, `p6` `[][]Color{}` `R()=0` |
| **B2** `GS0156` self-to-self | **CLOSED** | `p2` compiles/loads/runs |
| **B3** `eraseReferenceNullability` dropped | **structurally fixed** | `StructSymbol.cs:2427` passes `SubstituteNested`; closure at `:2361-2363`. Behaviourally incomplete — see SHOULD-FIX 2 |

Bare map literals and enum-as-map-key (`p4`, `p7`) reject with `GS0366` on **both** `main` and head — pre-existing, not a bypass site.

---

### BLOCKER — enum nested two *generic* levels deep regresses. No literal required.

`EnumSymbol.cs:205` — `ConstructNestedFromTypeParameterScope` silently bails.

```gs
package q1
struct Gen[T] {
    public struct MidG[U] {
        public enum Tone { A = 1, B = 2 }
        public func A() int32 { var t = Tone.B
 return int32(t) }
    }
}
var g Gen[int32].MidG[string]
func R() int32 { return g.A() }
```

| | result |
|---|---|
| `a6b1561f2` (current main) | **COMPILED**, `LOAD_OK types=4`, **`R() = 2`** |
| `8dddbbf2` (merge-base) | **COMPILED**, `R() = 2` |
| `e03e8fa54` | `error GS9998: InvalidOperationException: Open nested enum 'Tone' escaped its generic context` |

`p5.gs` (same shape with `[]Tone{Tone.B}`) identical. Isolating controls: `q3` (non-generic outer + generic mid) and `q4` (field, no expression) compile on both — so it is specifically **two generic levels + expression reference**.

**Root cause.** `binderCtx.CurrentTypeParameters` is seeded at `Binder.cs:605-620` from `function.ReceiverType.Definition.TypeParameters` — `MidG`'s own `[U]` only; the outer `T` is never added. `StructSymbol.CollectEnclosingTypeParameters(Tone)` returns `[T, U]`, the `"T"` lookup at `EnumSymbol.cs:203` fails, `:205` returns the enum **open**, and it reaches `UserTokenResolver.ResolveUserEnumTypeSpecArguments:1048` → `throw` → `GS9998`.

This is the SF4 fallback removal generalising: it was propping up **every** path where the closer bails, not just the array-literal one.

**Fix — one line**, `EnumSymbol.cs:203-208`:
```csharp
enclosingArguments.Add(
    typeParameters.TryGetValue(parameter.Name, out var argument) ? argument : parameter);
```
Correct by construction — `CollectEnclosingTypeParameters` returns the enum's own lexical enclosing chain, so those parameters are in scope by definition.

Parity with `main`, all executed:
```
        p1  p2  p3  p5  p6  q1  q3  q4
fix      0   4   3   2   0   2   2   0
main     0   4   3   2   0   2   2   0
```

Four suites with the fix, **0 Failed** — both sides measured on `8dddbbf2`, not mixed:

| suite | `8dddbbf2` | head + fix | delta | expected |
|---|---|---|---|---|
| Core | 7028 +1 skip | 7034 +1 skip | +6 | +6 |
| Compiler | 3893 | 3899 | +6 | +6 |
| Interpreter | 489 | 489 | 0 | 0 |
| LSP | 452 | 462 | +10 | +10 |

Counts reconcile exactly against added `[Fact]`/`[Theory]`/`[InlineData]`. **These figures need re-measuring after the rebase onto `a6b1561f2`.**

---

### SHOULD-FIX 1 — a survivor that is load-bearing. Do **not** delete it.

`EnumSymbol.cs:194`, the already-closed guard in `ConstructNestedFromTypeParameterScope`. Mutation `E7` (drop the `!IsDefaultOrEmpty` clause) **survives** the suite. It is nonetheless reachable, proved constructively:

```gs
struct Outer[T] {
    public enum Color { Red = 1, Green = 2, Blue = 3 }
    public func A() int32 { var a = []Outer[int32].Color{Outer[int32].Color.Blue}
 return int32(a[0]) }
}
var o Outer[string]
func R() int32 { return o.A() }
```
pristine head → COMPILED, `R() = 3`.  `E7` removed → **`error GS0156`** (already-closed `Outer[int32].Color` re-closed over `{T}`).

The reachability is **created by B1's own fix placement** at `:2109` — type-clause-resolved enums now flow through the closer a second time, and the fix depends on `E7` for idempotency. Please pin this shape.

---

### SHOULD-FIX 2 — `e03e8fa54` claims more than it delivers

Nested-enum nullability erasure works on the **static-virtual-interface** route (the commit's pin) but not the **instance** route:

| probe | member type | head | main |
|---|---|---|---|
| `n6` | nested **enum** `Outer[T?].Color`, impl `Outer[string].Color` | `GS0187` | `GS0158` |
| `n7` | nested **struct** `Outer[T?]`, impl `Outer[string]` | COMPILED | COMPILED |
| `n4` | plain `T` | COMPILED | COMPILED |

`n6`/`n7` are the same shape, enum vs struct. Not a regression (main rejects too, earlier) — an incompleteness in #2946. Either widen the pin or narrow the claim.

---

### SHOULD-FIX 3 — #2903 is the tight rebase, not #2929

`git merge-tree` rc=0 against both.

| PR | shared file | hunk gap |
|---|---|---|
| #2929 | `Binder.cs` | ~98 lines — mechanical (semantic overlap already noted in the PR body) |
| **#2903** | **`MetadataTokenCache.cs`** | this: `456, 477, 498` · #2903: `401-425, **445-450**` → **5 lines** |
| #2903 | `ImportedMemberRefFactory.cs`, `MemberLookup.cs` | far apart |

---

### Mutation table — 8 kills / 4 survivors. **No survivor is licence to delete.**

Control: Core-`Issue2898` 6/6 · Compiler-`Issue2898` 6/6 · Core full 7034+1 · LSP 462.

| id | branch | `file:line` | Core | Cmp | verdict |
|---|---|---|---|---|---|
| D1 | aggregate header → `FormatType` | `SymbolDisplay.cs:330` | **1** | 0 | KILLED → **#2898 pinned** |
| D2 | `[` / `]` delimiters | `SymbolDisplay.cs:475,487` | **1** | 0 | KILLED → **#2898 pinned** |
| D3 | `EnumSymbol` arm in `FormatType` | `SymbolDisplay.cs:727` | — | — | BUILD_FAIL, needs reformulation |
| E1 | `Members` definition ternary | `EnumSymbol.cs:~60` | **1** | **4** | KILLED |
| E5 | `SubstituteEnclosingArguments` existing-args | `EnumSymbol.cs:170` | 0 | **2** | KILLED |
| E6 | `enclosingParameters.IsDefaultOrEmpty → default` | `EnumSymbol.cs:178` | 0 | 0 | survives — **provably equivalent**: empty array reaches `ConstructNested`'s `IsDefaultOrEmpty` guard and returns the definition, same net result. Structural reduction, not a survival argument. |
| E7 | already-closed guard | `EnumSymbol.cs:194` | 0 | 0 | survives — **REACHABLE** (SHOULD-FIX 1) |
| E8 | name-not-in-scope bail | `EnumSymbol.cs:205` | 0 | 0 | survives — **this is the BLOCKER** |
| E10 | `SetTypeParameters` IsDefault | `EnumSymbol.cs:222` | 0 | **5** | KILLED |
| U1 | enum TypeSpec cache hit | `UserTokenResolver.cs:1013` | 0 | 0 | survives — pure perf cache, extra TypeSpec rows only |
| U2 | remap key → `(sym, null, null)` | `UserTokenResolver.cs:825` | 0 | **1** | KILLED → **#2930 pinned** |
| U3 | resolve-args return | `UserTokenResolver.cs:1053` | 0 | 0 | my own equivalent mutant, discard |

Each mutant was built with `dotnet build GSharp.sln -c Release -m:1`, **exit-code gated**, before a `--no-build` test run. Round-trip md5 fingerprints for cross-validation at `e03e8fa54` (agreeing across all three test output dirs, all non-zero):

```
A  pristine e03e8fa54            67023a44c5372b9f91b61b1ae2b1c840
B  E8 mutant (== proposed fix)   ad9dec4da269754d11e852389b6ae62e
C  restored pristine             67023a44c5372b9f91b61b1ae2b1c840   == A
```

**No conclusion here rests on a survival.** E7 and E8 carry constructive repros; E6 carries a structural reduction.

---

### Issue attribution

| issue | in `Fixes`? | pinned? |
|---|---|---|
| #2898 | yes | **proven** — D1, D2 killed; 6 Core facts |
| #2946 | yes | **proven** — E1/E5/E10/U2 killed; 6 Compiler facts with real load + invoke |
| #2930 | yes | **proven** — U2 killed |
| **#2947** | **no** | **dropped from scope** — `grep -rn GS0507 src/` returns nothing; PR body states the diagnostic was removed. Issue is still **OPEN** and needs re-queueing. |

### Still open, re-confirmed at this head

- **#2916** — nested **structs** keep inherited arity: `.Outer\`1+Inner\`1`, `.L1\`1+L2\`1`, `.Gen\`1+MidG\`2`. Real Roslyn gives `CS0426` on `Inner`, `MidG<>`, `L2`. Enums are correct (`.Outer\`1+Color` gp=1, `.Plain+Flat` gp=0).
- **#2910** — `GS0502` text byte-identical to `main`: `'GLib2.IGBox\`1[[System.Int32, System.Private.CoreLib, …]].V'`.

### Verified good

- `SemanticLookup.cs` +60 is **required, not scope creep**. `NormalizeForReferences` applied at **both** `GetReferences:1613` and `BuildReferencesIndex:1642`.
- **No rename behaviour change for non-generic / non-nested symbols** — 9-case probe, edit counts byte-identical to `main`: `nongeneric-struct 5 · nongeneric-enum 5 · nongeneric-iface 4 · local-var 2 · param 2 · func-name 2 · field 2 · generic-struct 4 · string-comment 0`. Comments and strings still not renamed.
- `Core.csproj` CRLF churn resolved — now `+1` line.
- `GetWrappedTypes` remains `internal` behind a narrow `InternalsVisibleTo("GSharp.LanguageServer")`. No public API widening.
- Corpus signal restated honestly: **0 of 148** samples contain a nested enum in a generic outer. Corpus identity proves nothing about this feature; the hand-written probes carry it.

---

**Gate on:** the one-line fix at `EnumSymbol.cs:205` plus pins for `q1.gs`/`p5.gs`, and a pin for `r1.gs` (SHOULD-FIX 1). SHOULD-FIX 2 and 3 can follow. Re-measure suite counts after rebasing onto `a6b1561f2`.

DavidObando:
### Correction to my round-5 mutation table — four rows were **unmeasured**, not survived

Fleet evidence rule tightened: a *survived* mutant produces no output change by definition, so it carries no evidence unless accompanied by an `X → mutant → X` round-trip md5 with matching hashes at both ends. Without that, the row must be reported as **unmeasured**.

I supplied the round-trip for **E8 only**. Correcting the four rows that lacked it:

| id | `file:line` | was reported | corrected |
|---|---|---|---|
| E6 | `EnumSymbol.cs:178` | survives | **unmeasured** |
| E7 | `EnumSymbol.cs:194` | survives | **unmeasured** |
| U1 | `UserTokenResolver.cs:1013` | survives | **unmeasured** |
| U3 | `UserTokenResolver.cs:1053` | survives | **unmeasured** |

**No finding changes.** None of them rested on the survival:

- **E7 / SHOULD-FIX 1** rests on a constructive repro with an observed behavioural flip, not on the survival. Pristine head → `r1.gs` COMPILED, `R() = 3`; E7 removed and rebuilt → `error GS0156`; tree restored, md5 back to `67023a44c5372b9f91b61b1ae2b1c840`, `r1.gs` COMPILED `R() = 3` again. The finding stands at full strength — `EnumSymbol.cs:194` is reachable and load-bearing, and still needs a pin.
- **E6** rests on a structural reduction, which is the permitted basis: mutating it yields an *empty* array rather than `default`, which then hits `ConstructNested`'s `IsDefaultOrEmpty` guard and returns the definition — the same expression on the reachable path. Not a survival argument, and **not** a licence to delete.
- **U1 / U3** — I drew no conclusion from either. U3 was my own equivalent mutant.

The six kills (D1, D2, E1, E5, E10, U2) are unaffected — stale binaries produce survivals, never false kills, so kill counts are lower bounds. **#2898** (D1, D2) and **#2930** (U2) remain proven pinned.

I will complete the round-trip md5 for E6/E7/U1/U3 when machine load allows; it does not gate the verdict either way.

### HEAD-vs-base differential — required evidence, already on the record

Observed with my own probes, in **both** directions, so this tree demonstrably was not sitting on `main`:

| probe | `8dddbbf2` / `a6b1561f2` (base) | `e03e8fa54` (head) |
|---|---|---|
| `n5.gs` nested enum in generic outer, interface impl | `error GS0158` | **COMPILED** |
| `r1.gs` `[]Outer[int32].Color{…}` inside `Outer[T]` | `error GS0158` | **COMPILED**, `R() = 3` |
| matrix `s1–s4`, `s9–s12` | `GS0149` / `GS0158` | **COMPILED**, load + invoke |
| `q1.gs` enum two generic levels deep | **COMPILED**, `R() = 2` | `error GS9998` ← the BLOCKER |
| `p5.gs` same shape via slice literal | **COMPILED**, `R() = 2` | `error GS9998` |

All compile probes invoked `out/bin/Release/Compiler/gsc.dll` **directly** — one `GSharp.Core.dll` consumer, so the propagation check is vacuous by construction. All execution results are real `Assembly.LoadFrom` + `GetTypes()` + invoke in a separate process, never `IlVerifier` alone.

### Provenance re-asserted at time of writing

```
git rev-parse HEAD                     e03e8fa54…  == briefed SHA
git rev-parse --abbrev-ref HEAD        fix/2898-symboldisplay-hover-headers
merge-base --is-ancestor HEAD main     false  → HAS-PR-CONTENT
git describe                           v0.3.362-117-ge03e8fa54
```
`EnumSymbol.cs`, `ExpressionBinder.Literals.cs` and `UserTokenResolver.cs` are byte-identical to the revisions I reviewed.

**Verdict unchanged: NOT MERGEABLE**, one BLOCKER at `EnumSymbol.cs:205`, one-line fix proven.

DavidObando:
### Round-5 revision: SHOULD-FIX 3 downgraded to NIT. Verdict unchanged.

Re-ran the collision analysis with the **natural** merge base (letting git compute it) instead of a forced `--merge-base origin/main`, and re-checked provenance against `refs/pull/2912/head` rather than a local SHA.

**Collision — zero, by both methods:**

```
2912 x 2929   natural base 8dddbbf28   rc=0   conflicts=0
2912 x 2903   natural base df4d6e1e6   rc=0   conflicts=0
```

**`MetadataTokenCache.cs` hunk proximity, recomputed at current heads:**

| PR | head | hunks (old-file lines) |
|---|---|---|
| #2912 | `e03e8fa54` | 459, 466, 470, 480, 500 |
| #2903 | `f219e2ccf` | 404, 411, 414, 416, 422, **447** |

Nearest gap is **12 lines** (447 → 459), not the 5 I reported. That is comfortably outside git's 3-line context window, so it cannot produce a textual conflict — consistent with `rc=0`. **SHOULD-FIX 3 → NIT**, informational only. #2903 is not in the merge order and #2912's rebase carries no textual risk from it.

**Provenance re-asserted against GitHub, not a local ref:**

```
gh pr view 2912 --json headRefOid   e03e8fa54bb61d93b1d7c8bc3bd69a49c8d44b86
refs/pull/2912/head                 e03e8fa54bb61d93b1d7c8bc3bd69a49c8d44b86
SHA I reviewed                      e03e8fa54
```

Git SHAs are content-addressed, so equality here is a complete proof that the tree I reviewed is the tree a merge would take — no local-only rebase hiding a different measurement. Scratch clones were made `--no-checkout` with an explicit `git checkout <briefed-sha>`, so no worktree HEAD was inherited.

Note `mergeStateStatus` is **CLEAN** on this PR. It does not see the defect — `q1.gs`/`p5.gs` compile and run on `main` and ICE with `GS9998` at head, and nothing in CI covers that shape.

### Standing verdict: NOT MERGEABLE

| sev | `file:line` | item |
|---|---|---|
| **BLOCKER** | `EnumSymbol.cs:205` | base COMPILED `R()=2` → head `GS9998`. One-line fix proven, four suites green. |
| SHOULD-FIX | `EnumSymbol.cs:194` | reachable, load-bearing, unpinned (`r1.gs`). |
| SHOULD-FIX | — | #2946 nullability erasure incomplete on the instance route (`n6` vs `n7`). |
| SHOULD-FIX | — | **#2947 dropped from scope** — not in `Fixes`, `GS0507` absent from `src/`. Needs re-queueing. |
| ~~SHOULD-FIX~~ NIT | `MetadataTokenCache.cs` | #2903 proximity — 12-line gap, zero conflict. Downgraded. |

DavidObando:

Specifically:

DavidObando:
## Scheduling notice: this PR's fixes exist **only in the author's worktree** — merging today merges the BLOCKER

Not a new review round. While verifying diagnostic-ID freedom for an unrelated issue I traced an ID into this branch, and the provenance check turned up drift the merge queue needs to know about.

**GitHub PR head:** `e03e8fa54` — this is what `gh pr merge` takes.
**Local branch tip:** `7f5609dc1` — **11 ahead / 9 behind**.

```
git rev-list --left-right --count e03e8fa54...7f5609dc1
9	11
```

The 9-behind side is a rebase (the tip contains `a6b1561f2` and `b1e693bb1`, which the PR head predates). Beyond the rebase there are **five commits of genuinely new production work**:

```
7f5609dc1  Pin nested enum nullability erasure
727ad3d66  Preserve nullability erasure for nested enums
e24b1e326  Close bare nested enum array literals
c07775cc9  Restore nested enum construction guards
9f825f276  Close nested enum generic contexts
```

Those map onto the three findings from my round-5 review:

| my finding | commit that appears to address it |
|---|---|
| **BLOCKER** `EnumSymbol.cs:205` — `q1.gs`/`p5.gs` regress base `R() = 2` → `GS9998` | `9f825f276` *Close nested enum generic contexts* |
| SHOULD-FIX 1 — `EnumSymbol.cs:194` guard reachable, load-bearing, unpinned | `c07775cc9` *Restore nested enum construction guards* |
| SHOULD-FIX 2 — #2946 nullability erasure incomplete on the instance route | `727ad3d66` + `7f5609dc1` |

`9f825f276` does what my analysis predicted was needed: it seeds the scope, calling `EnumSymbol.ConstructNestedFromTypeParameterScope` from `Binder.cs:3213` and from two accessor sites via a new `CloseNestedEnumOverCurrentTypeParameters` helper. The proximate `return nested` fallback at `EnumSymbol.cs` is untouched, which is consistent — my root cause was always upstream scope seeding, not the fallback itself.

### What this means for the queue

**None of it is on GitHub.** Merging #2912 in its scheduled slot merges `e03e8fa54`, in which the BLOCKER is live: `q1.gs` and `p5.gs` compile and run to `R() = 2` on base and fail with `GS9998` at that head. **The PR is not ready to merge until the author pushes.**

### What I have *not* verified

I have run **no build** against `7f5609dc1` and make **no claim that these commits work**. Everything above is commit-message and source-shape reading. The BLOCKER repro (`q1.gs`, `p5.gs`) and the base→head differential both need re-running once the work is pushed, and my suite reconciliation table is `8dddbbf2`-on-both-sides so it needs re-measuring against the new baseline regardless.

One methodological note for that round, because it nearly cost me a false report here: **read the net state at the tip, not the intermediate commits.** `9f825f276` introduced a `GS0507` diagnostic and replaced a defaulted `switch` with a hard `(StructSymbol)` cast — a new `InvalidCastException` surface on a path `ImportedTypeSymbol` can reach. **Both were withdrawn by later commits**; at `7f5609dc1` the cast is a null-checked branch at `ExpressionBinder.Access.Accessor.cs:1925-1928` and `GS0507` is gone. Reviewing per-commit would have produced two findings that no longer exist.

The withdrawn `GS0507` is worth one line of design feedback in advance: reporting *"enclosing generic arguments could not be resolved"* as a clean error would **still be a regression**, because base does not merely avoid crashing — it **compiles and runs** these programs to `R() = 2`. Parity is the bar, not a tidy diagnostic. The tip appears to have reached the same conclusion.

DavidObando:
## Rubber-duck round 6 — head `7f5609dc1` vs `main` `f70626452`

**Verdict: NOT MERGEABLE.** The round-5 BLOCKER **survives the rebase**. I now have a smaller repro and full execution evidence, and the "silently writing invalid metadata" justification does **not** hold for the shapes that regress.

Merge-base `a6b1561f2` (9 behind / 9 ahead). All probes on a real merge commit (`c4d419489`, parents exactly `7f5609dc1` + `f70626452`).

---

### BLOCKER — nested enum with a generic ancestor *above its immediate parent* is now an ICE

`src/Core/CodeAnalysis/Emit/UserTokenResolver.cs:1045-1054`

```csharp
internal ImmutableArray<TypeSymbol> ResolveUserEnumTypeSpecArguments(EnumSymbol enumSym, EnumSymbol def)
{
    if (!enumSym.EnclosingTypeArguments.IsDefaultOrEmpty) { return enumSym.EnclosingTypeArguments; }
    throw new InvalidOperationException(
        $"Open nested enum '{def.Name}' escaped its generic context; ...");
}
```

The round-2 fix correctly **removed** the open self-instantiation fallback, but nothing was added to **populate** `EnclosingTypeArguments` past the immediate parent. So the guard fires on valid programs.

**Matrix — identical sources, both compilers, `/out:` only:**

| shape | nesting | `main f70626452` | PR merged into main |
|---|---|---|---|
| `s1` | `Gen[T]` → `Tone` | rc=0, **R()=2** | rc=0, **R()=2** |
| `s3` | `Out` → `MidG[U]` → `Tone` | rc=0, **R()=2** | rc=0, **R()=2** |
| **`s2`** | **`Gen[T]` → `Mid` (non-generic) → `Tone`** | rc=0, **R()=2** | **rc=1 `GS9998`** |
| `q1` | `Gen[T]` → `MidG[U]` → `Tone` (local) | rc=0, **R()=2** | **rc=1 `GS9998`** |
| `p5` | same, array literal | rc=0, **R()=2** | **rc=1 `GS9998`** |

**Trigger, precisely:** the enum has a generic ancestor that is **not** its immediate parent. `s1` (parent is the generic) and `s3` (only the parent is generic) both pass — those are the shapes the PR's tests cover. `s2` is the minimal repro and the mid type is **not even generic**:

```gsharp
struct Gen[T] {
    public struct Mid {
        public enum Tone { A = 1, B = 2 }
        public func A() int32 { var t = Tone.B
 return int32(t) }
    }
}
var g Gen[int32].Mid
func R() int32 { return g.A() }
```

**The justification does not apply to these programs.** `main` is not emitting invalid metadata here — I verified by execution, not by `IlVerifier`:

```
main s1  LOAD=ok GETTYPES=ok n=3 R()=2
main s2  LOAD=ok GETTYPES=ok n=4 R()=2
main s3  LOAD=ok GETTYPES=ok n=4 R()=2
```

`Assembly.Load(File.ReadAllBytes(...))` + `GetTypes()` + invoke. Harness was retargeted to `net10.0` after an initial `System.Runtime 10.0.0.0` load failure that was **my** artifact, not the compiler's — the numbers above are post-fix. Control `ctl_bad.gs` fires `GS0125` on both compilers, so the harness is live in every run.

Also note the surface: this is **`GS9998`**, an unhandled `InvalidOperationException`, not a designed diagnostic. Even for a genuinely unsupported shape, an ICE is the wrong surface.

**Recommended fix (smallest correct one): keep the guard, fix the population.** The guard is right — a loud failure beats silent bad metadata. The defect is that `EnclosingTypeArguments` seeding walks only one level; it needs the full containing chain. Seeding site: `src/Core/CodeAnalysis/Binding/Binder.cs:605-620`, which seeds only the immediate parent's own type parameters while `CollectEnclosingTypeParameters` returns the transitive set. **Do not delete the guard to make this green.**

---

### SHOULD-FIX — reviewed head is 18 commits stale

`gsharp-wt/i2898` is at `a3e98d1b6` ("Pin nested enum nullability erasure"), **18 commits beyond `7f5609dc1`**, and divergent: `7f5609dc1` carries 9 commits `a3e98d1b6` lacks, so the local work sits on the **pre-rebase** line. The force-push did not capture it. Everything above is scoped to the pushed head; the BLOCKER may already be addressed locally. Worktree source is clean (`?? out/` only).

---

### Answers to the round-6 questions

**Q1 — merges and builds clean at `f70626452`.** `git merge-tree` rc=0. Merge build `rc=0`, `0 Warning(s) / 0 Error(s)`, 42 s, `-p:ContinuousIntegrationBuild=true`, no `--no-build`. Gate 2: **40** `GSharp.Core.dll` copies examined, **0** zero-byte.

Static sweep at the new tip: **Detector A — 0 hits / 14 net-deleted candidates / 1430 PR-added lines examined.** The candidate set contained `OuterMethodTypeParameters` and #2981's removed internals, so the pattern is proven to match known-present cases — the empty result is a *run* check, not an unrun one. **Inverse — 99 removed-line identifiers, 0** absent from the PR head.

**Q2 — not exposed to #2981's docs gate.** Descriptors **375 → 375**, `ADDED:[]` `REMOVED:[]`, **0** `docs/diagnostics.md` row changes (so no orphaning by row deletion either). Instrument: ID-literal set difference on `DiagnosticDescriptors.cs`, which is the **only** file in `src/` matching `= new("GS` — immune to multi-line declarations, modifier order and factory helpers.

**Q3 — #2947 still leaks at `7f5609dc1`; it is *not* closed.** Body disclosure is accurate.

| probe | `main` | PR head |
|---|---|---|
| `z2947a` | `'glib.Outer[T].Color'` → `'glib.Outer[int32].Color'` | `'glib.Outer[int32].Color'` → `'int32'` ✅ **fixed** |
| `z2947b` | `'glib.Outer[T].Color'` → `'glib.Outer[T].Color'` | **`'glib.Outer[object].Color'`** → `'glib.Outer[T].Color'` ❌ |

`z2947a` is a genuine #2898 win. `z2947b` still leaks `object`; root cause remains `ExpressionBinder.Access.Accessor.cs:2162` (typed to `StructSymbol`; `ImportedTypeSymbol` has no `EnclosingTypeArguments`). Main's `X → X` message is incoherent too — neither side is right, so this is a pre-existing defect the PR *reshapes* rather than introduces.

**Q4 — no undeclared closure found in the probed shapes.** The body's "#2910/#2916 not fixed" claim is **accurate**: `c2910.gs` yields identical `GS0113 GS9100` on both compilers. `Fixes #2898, #2946, #2930` and the #2947 disclosure all check out.

---

### NIT

The guard message names the enum (`'Tone'`) but not the enclosing construction, which is the part the user must change. `Gen[int32].Mid.Tone` would make it actionable.

---

**Reproduce:** sources `s1/s2/s3/q1/p5` are the five in the matrix; compile with `gsc <f>.gs /out:<f>.dll` (note `/out:`, not `-o`; `/target:library` is wrong here and yields `GS0285` for the top-level `var` — that cost me a probe).

Method notes: three-dot diffs from a computed `git merge-base` throughout; net state read at the tip, never per-commit; exit codes captured via `rc=$?` **assigned before** any other command. No worktree I do not own was modified.

DavidObando:
## Duck verdict — **MERGEABLE** (1 SHOULD-FIX, disclosed already)

Measured tree: merge commit `d8ca0593c`, parents **exactly** `61b7a08c1` (main) + `3da37d5e4` (head). Head verified against `gh` API, not inherited — briefing said `7f5609dc1`, **head had moved**. Three-dot from computed merge-base `fdba684b6`: **22 files / +1511 / −45**.

### Prior-round BLOCKER: **FIXED**. Re-derived, not assumed.
Round-6 blocker was nested enum whose generic ancestor is *not* its immediate parent → `GS9998` ICE. Re-ran full shape suite on base vs merged:

| shape | structure | base | merged |
|---|---|---|---|
| s1 | `Gen[T]` → enum | load/exec `R()=2` | load/exec `R()=2` |
| **s2** | `Gen[T]` → **non-generic** `Mid` → enum | `R()=2` | **`R()=2`** |
| s3 | non-gen `Out` → `MidG[U]` → enum | `R()=2` | `R()=2` |
| q1 | `Gen[T].MidG[U]` → enum | `R()=2` | `R()=2` |
| p5 | same, enum in array | `R()=2` | `R()=2` |

All 10 cells: compile rc=0, `Assembly.Load` ok, `GetTypes()` ok, `R()` invoked. **Not** `IlVerifier`-only — real load+execute, per #3001. Harness negative-controlled: junk bytes → `BadImageFormatException` rc=5, so it can report failure.

### #2898 repro is ALIVE on clean `main` — PR is not obsolete
Checked first, because #2922 was closed after an unrelated merge silently fixed it. `SymbolDisplay.cs:486`/`:498` still emit `<`/`>` on `61b7a08c1`. **#3039 did not fix this.** Closing is not the verdict.

### Mutation — both killed, valid builds
| mutant | site | build | result |
|---|---|---|---|
| A: `[`/`]` → `<`/`>` | `SymbolDisplay.cs:486,498` | exit 0, 0 errors | **KILLED** 1/6 — `DeclarationHeaders_UseSquareBracketsAndConstructedArguments` |
| B: delete `EnumSymbol` case | `SymbolDisplay.cs:734` | exit 0, 0 errors | **KILLED** 3/6 |

Note `AppendTypeParameters` has **one** surviving caller after this PR (`:268`, generic functions) — hunks 1–2 removed the others. It is live, not dead; mutant A's killing test is a `func` header, confirming that is its only reachability.

**Self-correction:** my recorded "clean" `GSharp.Core.dll` md5 matched **neither** base nor merged after restore. Build proved deterministic (no-op and forced-recompile both stable), so the baseline capture was mine and bad. Doesn't void the above — both mutants built clean and went **red**, which is direct proof; md5 is only load-bearing for a *surviving* mutant.

### Docs gate — not exposed
Descriptors `= new("GS####"` in `src/`: **377 → 377** (pattern positively controlled, matches 377 not 0). `docs/diagnostics.md` 383, `website/docs/ref/diagnostics.md` 381 — unchanged both sides. No row added or deleted.

### SHOULD-FIX — #2947, `GS0156` renders a **false** type argument
Four-cell matrix (library compiler × consumer compiler) gives four *different* renderings:

| lib | consumer | rendering |
|---|---|---|
| base | base | ``glib.Outer`1.Color`` ← CLR backtick leaks on pure `main` |
| base | merged | ``glib.Outer`1.Color`` |
| merged | base | `glib.Outer[T].Color` → `glib.Outer[T].Color` (incoherent X→X) |
| **merged** | **merged** | **`glib.Outer[object].Color`** ← #2947 |

Why SHOULD-FIX and not BLOCKER: `main` is **already wrong here** (backtick leak — the very defect #2898 exists to kill). This PR turns *visibly unresolved* into *confidently wrong*, which is a reshape, not a regression. It is disclosed in the body ("tracked in #2947"), and **no test freezes it** — 0 hits for `Outer[object]` / `[object].Color` across `test/`, grep positively controlled (`Outer[T]` found at `Issue2898NestedEnumConstructedOuterEmitTests.cs:27,40,192`). Flagging for the maintainer to overrule with the matrix in hand. The `merged-lib / base-consumer` cell rendering `X → X` is a third distinct bad output worth its own note.

### NIT / non-result, reported as such
My 7-shape sibling display probe (`Box[int32]`, `Box[Box[int32]]`, `Pair[int32,string]`, `[]int32`, `int32?`, …) rendered **identically** on base and merged — **wrong oracle, not evidence of correctness**. `gsc` diagnostics exercise *type positions*, already routed through `ToTypeDisplayString` (consolidated by #2886); they never reach `SymbolDisplay` declaration headers. The right oracle is LSP hover / `Core.Tests`. Recording the failure to discriminate rather than banking it as a pass.

### Green
Merged tree builds `exit 0`, **0 warnings / 0 errors**. `Core.Tests` `Issue2898SymbolDisplayHeaders` **6/6**. `LanguageServer.Tests` Hover+Rename **58/58**.

PR #2917, opened 2026-07-30
Fix fixed pin cleanup on all exits

### LINKED ISSUE #2900: gsc: control-flow exits from fixed skip the pin-release epilogue
Found during review of #2883. This is pre-existing and code generation is byte-identical before and after that CFG-only fix.

## Summary

`MethodBodyEmitter.EmitFixedStatement` emits the pin prologue, the body, and then a normal-fallthrough epilogue that clears the pinned local. A `break`, `goto`, `return`, or other control-flow exit from inside the body branches past that epilogue, leaving the object pinned for the remainder of the method.

This is over-pinning, not under-pinning: it is not memory-unsafe, but a loop escape early in a long-running method can keep an array or string pinned much longer than its lexical `fixed` scope.

## Repro and emitted IL (verified)

```gsharp
package FixedEscapePin

func F(xs []int32) int32 {
    unsafe {
        for {
            fixed p *int32 = xs {
                break
            }
        }
    }
    return 7
}
```

Relevant gsc IL:

```il
IL_0021: br IL_002d      // break exits fixed
IL_0026: ldnull          // normal-fallthrough unpin, skipped
IL_0027: stloc.0
IL_0028: br IL_0000
IL_002d: ldc.i4.7
IL_002e: ret
```

A 1000-iteration tail after the fixed block confirms the pinned local remains live for the rest of the method. No protected region ensures cleanup.

Roslyn emits cleanup on every exit using a protected cleanup shape (`leave` plus a finally that clears the pinned local).

## Expected

Every exit from a `fixed` body runs the pin-release epilogue before control reaches its external target. Normal fallthrough, break, continue, goto, return, and throw must preserve the intended lexical pin lifetime.

- Inspect emitted exception regions / IL so break and goto run cleanup before their external labels.
- Return and throw from fixed run cleanup before method exit/unwind.
- Nested fixed blocks release pins in inner-to-outer order.
- Runtime stress keeps a long tail after an early fixed exit and verifies the object is no longer pinned.
- Existing pointer ILVerify exceptions remain the only ignored verifier categories.

DavidObando:
Closes #2900.

## Root cause

`EmitFixedStatement` pinned its source, emitted the body, then cleared the pinned local only on sequential fallthrough. `break`, `continue`, `goto`, and `return` could branch past that cleanup, so the object stayed pinned until method exit.

## Emit shape

The pin prologue remains outside the protected region. The fixed body is a CLR `try`; its `finally` clears the pinned array/string reference or managed by-ref local. Region-crossing branches use `leave`. A protected value return stores its computed value in one planned local, leaves all enclosing protected regions, reloads it, then executes `ret` outside them. Nested CLR finally regions provide inner-to-outer release ordering.

This matches Roslyn's protected-cleanup shape. I ruled out:

- keeping a normal-fallthrough epilogue: it cannot cover branches or exceptions;
- wrapping the pin prologue: Roslyn does not, and no pin exists until that prologue completes;
- lowerer/CFG return rewriting: a built lowerer alternative made a synthesized default return reachable, broke GS0100, and broke async emission; `ControlFlowGraph.cs` remains untouched;
- state-machine field hoisting for fixed temporaries: it made the prologue write a local while the body read a zero-valued field. Fixed synthetic locals remain `MoveNext` locals.

## Static-initializer blocker

Round-one code let constructor-family call sites finish method bodies independently. `shared { init { ... } }` omitted the fixed-return epilogue and produced `GS9998: Label 1 has not been marked`.

All six constructor-family `CreateEmitter` paths now funnel through `AddCompletedMethodBody`. That helper owns fixed-return epilogue emission, trailing `ret`, and `AddMethodBody`. No constructor-family call site can add a body while forgetting the epilogue. The static-initializer regression compiles, verifies, loads, runs, prints `41`, and asserts the protected `leave`/finally shape.

## Suspension rule: GS0506

A CLR protected region cannot be resumed by branching into its `try`. Direct `await` or `yield` inside `fixed` therefore now reports `GS0506` before state-machine rewriting. The syntax walk also recognizes `await for` and `await using`.

Nested fixed statements are scanned by their own bind operation, preventing duplicate diagnostics. Function literals and lambdas are opaque boundaries, so an async lambda declared inside `fixed` remains legal; its suspension occurs in a different method. Tests assert three direct/nested GS0506 diagnostics and run the nested-lambda case to print `7`.

## Repro IL

Before:

```il
IL_0021: br IL_002d       // break skips cleanup
IL_0026: ldnull
IL_0027: stloc.0
IL_0028: br IL_0000
IL_002d: ldc.i4.7
IL_002e: ret
```

After:

```il
.try
{
    IL_0021: leave IL_0033
    IL_0026: leave IL_002e
}
finally
{
    IL_002b: ldnull
    IL_002c: stloc.0
    IL_002d: endfinally
}
IL_002e: br IL_0000
IL_0033: ldc.i4.7
IL_0034: ret
```

PE exception-region table:

```text
before: 0 regions
 after: Finally: try IL_0021..IL_002b, handler IL_002b..IL_002e
```

Tests decode IL and assert escaping `leave` instructions, real finally regions, cleanup stores, `endfinally`, nested-region containment, and external targets.

## Pin-release proof

`EarlyBreak_ReleasesPinBeforeCompactingGc` records an array address inside `fixed`, breaks out, creates 20,000 allocation-pressure objects per attempt, forces a compacting generation-2 GC, repins, and compares addresses. Branch prints `True`; main prints `False`. Exception-region and `leave` assertions provide deterministic structural proof if a runtime refuses relocation.

Every successful test program also runs through `dotnet exec`. All emitted assemblies pass `IlVerifier.Verify`; the only ignored categories remain unchanged:

`Unverifiable`, `UnmanagedPointer`, `StackUnexpected`, `StackByRef`, `ExpectedPtr`, `StackUnexpectedArrayType`, `ExpectedNumericType`.

## Coverage

Eleven end-to-end issue facts and a 31-test fixed regression matrix cover:

- break, continue, goto, labeled outer break/continue, and arm-contained labeled exits;
- computed/void/mixed/constructor/static-initializer/deinitializer returns and thrown exceptions;
- return from pattern-switch and select arms inside fixed;
- nested fixed cleanup ordering;
- arrays, fixed arrays, slices, strings, Span/GetPinnableReference, and empty arrays;
- fixed inside try/catch/finally, try inside fixed, scope, using, switch, and select;
- async methods with later await, return from fixed before later await, and iterators with later yield;
- GS0506 for direct, await-for, and await-using suspension plus legal suspension inside a nested lambda;
- runtime compaction proof plus metadata/IL decoding and actual execution.

G# fixed syntax has one binding and accepts array/slice, string, or GetPinnableReference sources. `fixed (... = &expr)` and multi-declarator fixed syntax are not language forms; nested fixed covers multiple simultaneous pins.

## Over-fire corpus

Rebased onto `origin/main` at `d4038f1d` (#2904, #2909, and #2913 included). Built gsc from that main and this branch, then compiled all 148 `samples/**/*.gs` and `e2etests/**/*.gs` files.

- branch successes: 139
- main successes: 139
- identical rejected harness inputs: 9
- compile-status mismatches: 0
- diagnostic mismatches: 0
- byte differences: 0
- loadability: 139/139 on both

The corpus has no real fixed statement, so zero byte differences is expected and proves non-fixed output stayed byte-identical.

A dedicated eight-file valid fixed shape corpus fills that gap:

- byte differences: 8/8, all justified by protected cleanup/return emission;
- diagnostic differences: 0;
- loadability: 8/8 on both compilers;
- branch execution: 8/8 successful;
- main still prints `False` for pin movement and fails the async shape with `NullReferenceException`.

## Per-hunk mutation audit (round 2, as of `eb38ef4e`)

Each mutation was compile-clean, applied alone, restored before the next trial, and run with unrelated `FunctionArrowBody_RunsAndReturns` control coverage. The final unmutated control run passed 10/10 at `eb38ef4e`.

### Behavioral pins: 26/26 red

| Production location | Mutation | Red evidence |
|---|---|---|
| `DiagnosticDescriptors.cs:376` | change GS0506 ID | suspension diagnostic fact |
| `DiagnosticBag.Reports.Emit.cs:136-137` | route to another descriptor | suspension diagnostic fact |
| `StatementBinder.Blocks.cs:790` | remove fixed-body scan | suspension diagnostic fact |
| `StatementBinder.Blocks.cs:914-918` | remove lambda boundary | legal nested-lambda execution fact |
| `StatementBinder.Blocks.cs:914-918` | remove nested-fixed boundary | duplicate-diagnostic count fact |
| `StatementBinder.Blocks.cs:923-925` | disable await detection | suspension diagnostic fact |
| `StatementBinder.Blocks.cs:926-931` | disable yield detection | suspension diagnostic fact |
| `StatementBinder.Blocks.cs:934-936` | disable recursive walk | suspension diagnostic fact |
| `ConstructorBodyEmitter.cs:204` | bypass shared completion for static init | static initializer fact: GS9998 unmarked label |
| `ConstructorBodyEmitter.cs:648-658` | omit fixed epilogue in shared helper | static initializer fact: GS9998 unmarked label |
| `MethodBodyEmitter.Statements.cs:66-68` | corrupt fixed region labels | branch fact: `InvalidProgramException` |
| `MethodBodyEmitter.Statements.cs:70-72` | stop protected body tracking | branch fact: `InvalidProgramException` |
| `MethodBodyEmitter.Statements.cs:93-95` | drop finally-region registration | branch fact: invalid region bounds/IL |
| `MethodBodyEmitter.cs:202-204` | disable fixed-return state | switch-return fact: `InvalidProgramException` |
| `MethodBodyEmitter.cs:245-250` | omit method fixed-return epilogue | switch-return fact: GS9998 unmarked label |
| `MethodBodyEmitter.cs:259-273` | corrupt shared return label/reload | switch-return fact: GS9998 unmarked label |
| `MethodBodyEmitter.cs:345-357` | emit protected `ret` instead of spill/leave | switch-return fact: `InvalidProgramException` |
| `ReflectionMetadataEmitter.cs:4111-4119` | disable return discovery | return/throw fact: `InvalidProgramException` |
| `ReflectionMetadataEmitter.cs:4155-4157` | ignore return match inside fixed | return/throw fact: `InvalidProgramException` |
| `ReflectionMetadataEmitter.cs:4158-4166` | drop block propagation | return/throw fact: `InvalidProgramException` |
| `ReflectionMetadataEmitter.cs:4167-4169` | drop fixed-entry propagation | return/throw fact: `InvalidProgramException` |
| `ReflectionMetadataEmitter.cs:4170-4177` | drop select-arm propagation | select-return fact: `InvalidProgramException` |
| `ReflectionMetadataEmitter.cs:4178-4185` | drop switch-arm propagation | switch-return fact: `InvalidProgramException` |
| `ReflectionMetadataEmitter.cs:4267-4268` | stop passing planned return state | switch-return fact: `InvalidProgramException` |
| `AsyncCaptureWalker.cs:166` | stop registering pointer temporary | state-machine fact: `NullReferenceException` |
| `AsyncCaptureWalker.cs:225-227` | allow fixed locals to hoist | state-machine fact: `NullReferenceException` |

### Guards: 11/11 stayed green

These mutations preserve current behavior but guard structurally equivalent constructor paths or future planner/state-machine shapes. They are reported separately rather than misrepresented as test-detected behavior.

| Production location | Guard mutation |
|---|---|
| `ConstructorBodyEmitter.cs:249` | bypass shared helper for default ctor |
| `ConstructorBodyEmitter.cs:321` | bypass shared helper for primary ctor |
| `ConstructorBodyEmitter.cs:444` | bypass shared helper for forwarding ctor |
| `ConstructorBodyEmitter.cs:570` | restore explicit-ctor local completion |
| `ConstructorBodyEmitter.cs:645` | restore deinitializer local completion |
| `MethodBodyEmitter.cs:103-105` | alter readonly return-state field scaffolding |
| `MethodBodyEmitter.cs:172-173` | alter optional return-state parameter defaults |
| `ReflectionMetadataEmitter.cs:4079-4080` | alter planner return-state defaults |
| `AsyncCaptureWalker.cs:165` | stop registering pinned temporary |
| `AsyncCaptureWalker.cs:167-170` | stop registering pinnable source temporary |
| `AsyncCaptureWalker.cs:132` | alter fixed-local set initialization |

Dead `TryGetFixedReturnType` traversal for try/catch/finally/scope was removed: Lowerer already spills those returns before emission. Live select and pattern-switch cases now have direct runtime/IL tests.

### Round-3 delta: 3 pins / 2 guards

| id | mutation | result |
|---|---|---|
| M1 | GS0506 `await for` alternative | **PIN** |
| M2 | GS0506 `await using` alternative | **PIN** |
| M3 | `PinnableReference` release opcodes | **PIN** |
| M4 | lazy → eager `DefineLabel` | GUARD (11/11 green) |
| M5 | drop `FunctionEmitter` epilogue call | GUARD (11/11 green) |
| M6 | drop `PointerVariable` registration (control) | **PIN** |

M6 is the existing pinned control and is excluded from the round-3 delta count. M5 was then deleted because forced reachability emits invalid `ldloc; ret` IL in void `MoveNext`; it was not a valid safety net.

## Validation

- issue #2900 tests: 11 passed
- fixed regression matrix: 31 passed
- targeted Core diagnostic/static-init tests: 11 passed
- DiagnosticIdUniquenessTests: 4 passed
- rebase base / merge-base: `origin/main` `60e5b8bf`
- Core.Tests: 6,913 passed, 1 skipped
- Compiler.Tests: 3,753 passed on rebased HEAD (`origin/main` `60e5b8bf`: derived 3,742 because this PR adds the only changed test file, containing 11 facts; PR delta: +11)
- CI-shape Release `GSharp.sln` build: succeeded, 0 warnings, 0 errors
- working tree clean; no local e2e suite run

No work deferred.


DavidObando:
Merge-order note for #2926 (CFG try/finally): both PRs rewrite the trailing-return block in `MethodBodyEmitter.EmitMethodBody`. Correct resolution keeps #2917's fixed-return epilogue block first, including its early `return`, then applies #2926's `alwaysAppendRet` / invalid-fallthrough logic:

```csharp
if (this.EmitFixedReturnEpilogue())
{
    this.il.OpCode(ILOpCode.Ret);
    return;
}

if (alwaysAppendRet)
{
    this.il.OpCode(ILOpCode.Ret);
}
else if (!this.currentPositionEndsInTerminator)
{
    // #2926 invalid-fallthrough throw
}
```

Do not reorder or drop the early return. Otherwise methods whose only returns are inside `fixed` can regress to `GS9998: Label N has not been marked` or gain a bogus throw. `SwitchArmReturn` / `ReturnAndThrow` should catch a bad resolution.

DavidObando:
## Review — PR #2917 round 2 (`eb38ef4e`)

**Verdict: MERGEABLE.** No blockers. Round-1 blocker fixed and regression-tested. Recording here so the findings survive independently of any working session.

### Round-1 blocker (static-init ICE): fixed

| build | result |
|---|---|
| `d4038f1d` (main) | compiles, prints `41` |
| round-1 `8472099a` | `error GS9998: InvalidOperationException: Label 1 has not been marked.` |
| HEAD `eb38ef4e` | compiles, prints `41` |

New private helper `ConstructorBodyEmitter.AddCompletedMethodBody` (`ConstructorBodyEmitter.cs:648`) owns epilogue + `ret` + `AddMethodBody`; all six constructor-family sites funnel through it. Pinned by `StaticInitializerReturn_UsesSharedFixedEpilogue_AndRuns` (`Issue2900FixedPinReleaseEmitTests.cs:267`) — build-green mutation reverting that one site turns exactly that test red. The PR body's claim is correctly scoped to the constructor family and is not overclaimed.

### SHOULD-FIX 1 — `await for` / `await using` arms are live but untested

`StatementBinder.Blocks.cs:924`. Deleting only `or AwaitForRangeStatementSyntax or AwaitUsingStatementSyntax`:
- builds green under `-warnaserror`
- **all 9 Issue2900 tests still pass**; the whole 48-test fixed matrix still passes
- mutant compiler accepts `await for item in Src() { ... }` inside `fixed` with **no diagnostic**
- ilverify: `[IL]: Error [BranchIntoTry]: [...]::MoveNext()][offset 0x0000000D] Branch into try block.`
- running it: `System.InvalidProgramException`

`BranchIntoTry` is not in `FixedIlVerifyIgnored`, so one `CompileErrors` test per alternative pins it completely (~10 lines).

### SHOULD-FIX 2 — GS0506 undocumented

Absent from both `docs/diagnostics.md` and `website/docs/ref/diagnostics.md`. `DiagnosticIdUniquenessTests.cs:145` only validates IDs already present, so the omission is silent. Precedent: sibling `fixed` diagnostics GS0400/GS0401 live at `website/docs/ref/diagnostics.md:273-274`. GS05xx sections end at GS0502 (`website/docs/ref/diagnostics.md:1674`).

GS0506 is collision-free; `GS0503`/`GS0504`/`GS0505`/`GS0507` are absent from the tree. **No over-fire found** across lambda-nested `await`, `yield` in function literals, `await` in non-async `fixed`, async lambda in sync `fixed`, and nested `fixed` with suspension outside both. No reachable under-fire on shipped HEAD.

### Verified clean

Select/switch `Find` cases correctly removed as dead (`Lowerer` spills those returns pre-emit; probed across free functions, static initializers and constructors). Pipe-deadlock fix in `CompiledProgram.Run()` (`:969-997`) matches the canonical pattern at `IlVerifier.cs:135-145`. ADR-0125 corrections accurate. `FixedIlVerifyIgnored` byte-identical to the pre-existing lists — not widened. GC test 21/21 across workstation, server and server+concurrent GC.

### Mutation summary — 24 semantic branches, independently re-derived

**16 pins / 8 guards.** Build-green mutants only; unmutated control re-run before every batch; source restored **and rebuilt** between trials. Unpinned: the `PinnableReference` release opcode swap (`ldc.i4.0;conv.u` → `ldnull`, invisible to tests and ilverify though ADR-0125 documents the exact opcodes); `AsyncCaptureWalker.cs:165` and `:167-170`; and the `await for`/`await using` sub-branch above.

### Merge-order — non-mechanical conflict with #2926

Both PRs rewrite the same `if (alwaysAppendRet || !this.currentPositionEndsInTerminator)` block in `MethodBodyEmitter.EmitMethodBody` (main line 234; this PR at `:242-256`). This PR inserts `if (this.EmitFixedReturnEpilogue()) { Ret; return; }` **first, with an early return**; #2926 replaces the trailing-ret logic with an `alwaysAppendRet` / `throw` split.

**Correct resolution keeps this PR's block first and keeps its early return.** A naive merge that reorders it or drops the early return yields either `GS9998: Label N has not been marked` or a bogus `throw` on every method whose only returns are inside `fixed`. Whoever merges second must read both hunks rather than accepting a mechanical resolution.

`git merge-tree` vs #2923: 0 conflicts. No rebase required — `merge-base == origin/main == d4038f1d`.

### Orthogonal, pre-existing — not caused or fixed here

`select` inside an `async` function loses the arm binding: `async func` with `select { case let value = <-ch { v = 7 + value } }` returns `7` instead of `10`, on **both** main and HEAD, with or without `fixed`. The sync version is correct. Worth its own issue.

DavidObando:
Punch-list complete in `b00e2121`:

- Added separate GS0506 facts for `await for` and `await using` inside `fixed`. Each compile-green single-alternative deletion independently turns only its matching fact red (`gsc` exits 0 without GS0506).
- Documented GS0506 in `website/docs/ref/diagnostics.md` after GS0502.
- Added exact `ldc.i4.0; conv.u` pinnable-release IL checks; the compile-green `ldnull` mutation turns the pin-kind fact red.
- Deleted dead `PinnedVariable` / `SourceVariable` async-capture exclusions; retained live `PointerVariable` exclusion.
- Made the fixed-return label lazy and finalized any actual MoveNext fixed-return target, removing the eager unmarked-label hazard.
- Local validation: Release graph build 0 warnings/errors; Core 6913 passed / 1 skipped; Compiler 3719 passed; Interpreter 453 passed. All checks green on `b00e21218c6510aa2c57bdd899c865a0c0f39824`.
- Follow-ups filed as #2933 and #2934.

DavidObando:
# Round-3 re-check — PR #2917, HEAD `b00e2121`

## Verdict: **MERGEABLE**

Both round-2 SHOULD-FIXes landed and are genuinely pinned. The lazy-label change does what I asked and I could not re-break it. No BLOCKER. Two NITs below; neither should hold the merge queue.

---

## 1. Lazy label — property holds

**Structural analysis.** `fixedReturnLabel` is now `LabelHandle?` (`MethodBodyEmitter.cs:105`). The *only* writer is `GetFixedReturnLabel()` (`:274-278`), and its only caller is the `leave` site (`:357`). The only reader is `EmitFixedReturnEpilogue()` (`:258`), gated on `HasValue`.

- **Marked-without-defined: impossible by construction.** `MarkLabel` is unreachable unless `HasValue`. This half is now a true invariant.
- **Defined-without-marked: requires a 9th emitter entry point.** `grep -rn "CreateEmitter" src/` → exactly 8 sites: `ConstructorBodyEmitter.cs:{192,231,283,397,508,600}` (all funnel to `AddCompletedMethodBody`, `:648`), `FunctionEmitter.cs:164` (now calls the epilogue at `:177`) and `:360` (via `EmitMethodBody`, `:244`). I checked every constructor site for an escape between `EmitBlock` and `AddCompletedMethodBody`: each is `try { EmitBlock } catch { Wrap }` followed by an unconditional `return this.AddCompletedMethodBody(...)`. No early return, no conditional skip.

**Instrumented probe.** I temporarily added three counters to `EmitFixedReturnEpilogue` — `PROBE-GAP` (`hasFixedReturns` true, label never defined), `PROBE-DOUBLEMARK`, `PROBE-LEAVE-WITHOUT-PLAN` — plus `PROBE-MOVENEXT-EPILOGUE-TAKEN` in `FunctionEmitter`. Rebuilt, then swept:

| sweep | PROBE hits |
|---|---|
| 148 corpus `.gs` files (`samples/` + `e2etests/`) | **0** |
| 50-test fixed matrix (`~Fixed\|~Issue1026\|~Issue1043`), all green | **0** |
| 10 hand-written probes (async/iterator/switch/select/try/nested/deinit/ctor/getter/lambda) | **0** |

**ICE repro re-run.** `return` inside `fixed` inside `shared { init { } }` — compiles, prints `41` on `b00e2121`. (`8472099a` gave `GS9998: ... Label 1 has not been marked.`)

**Tried to re-break it, could not.** Probes `b1.gs` (`deinit` + class ctor + struct ctor, each with a fixed-return) → `12 / 7 / 0`; `b2.gs` (property getter, generic `func Gen[T]`, lambda with its own `fixed`+return) → `3 / gen / 42`; `a1/a3/a5/a7` (async return-from-fixed before and after `await`, async switch-arm return in fixed, `try`+fixed, nested fixed) → all correct. ilverify clean on all ten after filtering exactly `FixedIlVerifyIgnored`. Nothing produced an unmarked label, a double mark, or invalid IL.

**The semantic gap you asked about — it exists in theory, is empty in practice.**

Mutation **M4-EAGER** (restore the eager `DefineLabel()` in the ctor, i.e. `eb38ef4e` semantics, everything else unchanged) is build-green and leaves **11/11 Issue2900 tests passing**. Mutation **M7-LAZYGATE** (`!HasValue` → `!hasFixedReturns || !HasValue`) is likewise green. So no test distinguishes the two behaviours.

The stronger evidence is byte-level: I built `eb38ef4e` separately (`/tmp/d2900r3/r2`) and compared compiler output over the whole corpus with output paths normalised:

```
corpus=148 compiled=139 bothFailed=9 byteDiffs=0 diagDiffs=0
```

Zero byte differences, zero diagnostic differences, plus 8 hand-written async/iterator/pin probes all `IDENTICAL`. **The trailing epilogue did not disappear from any real method**, because `PROBE-GAP` never fired: there is no shipped program in which `hasFixedReturns` is true and no `leave` is emitted. Round 3 is byte-neutral.

**Honest scope.** This is *narrower hazard*, not literal unrepresentability. The residual failure mode is: a future 9th `CreateEmitter` site that both emits a fixed-return `leave` and skips the epilogue. That still fails **loudly** (`GS9998: Label N has not been marked`), never silently. Round-2 NIT is substantially closed.

---

## 2. `FunctionEmitter.cs:177` — still dead, and I would rather it were not there

**Still never taken.** The reason is unconditional, not heuristic: `MoveNextBodyRewriter.InnerRewriter.RewriteReturnStatement` (`MoveNextBodyRewriter.cs:305`) rewrites *every* `BoundReturnStatement` into `retValLocal = expr; goto ExpressionReturnLabel`. No `BoundReturnStatement` survives into a MoveNext body, so `TryGetFixedReturnType` can never set `hasFixedReturns` there. Confirmed by the instrumented sweep above (0 `PROBE-MOVENEXT-EPILOGUE-TAKEN`) across corpus, fixed matrix, and probes including async+`switch`-arm return in `fixed` and async nested `fixed`.

**Forced it anyway**, as asked. Two temporary env-gated switches were required to reach it at all: disable the MoveNext return spill *and* restore the `BoundTryStatement` descent that round 2 deleted from `Find`. Only then:

```
PROBE-MOVENEXT-EPILOGUE-TAKEN
PROBE-MOVENEXT-EPILOGUE-TAKEN
```

and the result is **not** benign:

```
[IL]: Error [ReturnVoid]: [... <Delay>d__0::MoveNext()][offset 0x00000097][found Int32]
      Stack must be empty on return from a void function.
System.InvalidProgramException: Common Language Runtime detected an invalid program.
```

That forced run is confounded (the no-spill switch damages other methods too), so treat the *specific* offsets as noise — but the error **class** is exactly what the analysis predicts and is not confounded: `MoveNext` returns `void`; if `fixedReturnSlot >= 0`, `EmitFixedReturnEpilogue` emits `ldloc N` and `FunctionEmitter` then appends `ret`. `ldloc; ret` in a void method is `ReturnVoid`, invalid IL.

**NIT.** Before round 3 this state ICE'd loudly (unmarked label). After round 3 it would emit a silent invalid program. The new line is therefore not a safety net — it is unreachable code that would misbehave if it ever became reachable. Mutation **M5-FEDROP** (delete the `if (emitter.EmitFixedReturnEpilogue()) { il.OpCode(ILOpCode.Ret); }`) is build-green and leaves **11/11 green** — nothing tests it. Cheapest fix consistent with ponytail: drop the three lines and keep the existing comment, or guard it with `Debug.Assert(!emitter.EmitFixedReturnEpilogue())`. Not blocking, since it is provably dead today.

---

## 3. Both round-2 SHOULD-FIXes: landed correctly

**G5 — each half independently killed.** Unmutated control 11/11 first; each mutation applied alone, build-green under `-warnaserror`, restored and rebuilt between trials:

| mutation (`StatementBinder.Blocks.cs:924`) | result |
|---|---|
| **M1** delete `AwaitForRangeStatementSyntax` only | **RED — `SuspensionInsideFixed_AwaitForReportsGs0506`** (1 failed / 10 passed) |
| **M2** delete `AwaitUsingStatementSyntax` only | **RED — `SuspensionInsideFixed_AwaitUsingReportsGs0506`** (1 failed / 10 passed) |

Each half is pinned by exactly one test, and by a *different* test. The round-2 finding is genuinely closed.

**Further GS0506 falsification on shipped HEAD, all clean:**
- Nested `fixed` with `await` in the *inner* body → exactly **one** GS0506, no duplicate, no under-fire (`c2.gs`).
- `await` in the **pin source expression** (`fixed p *int32 = await GetArr() { }`) — the walk covers only `syntax.Body`, so I expected a hole. The binder rejects it first with GS0401, **identically on `d4038f1d` and HEAD**. Pre-existing, orthogonal, door closed.
- `defer` inside `fixed` with an `await` after the region → runs correctly (`deferred`, `7`).
- Over-fire: none. Nested lambda `await` still legal; G# has no local-function syntax node.

**GS0506 docs — correct.** `website/docs/ref/diagnostics.md:1688`, immediately after the GS0502 section at `:1674`, severity `Error` matching `DiagnosticDescriptors.cs:376`, trailing newline present. It is **not** in `docs/diagnostics.md`, but that matches the established precedent for `fixed` diagnostics — GS0400/GS0401 have 0 hits there and live only in the website file. `DiagnosticIdUniquenessTests.cs:147` reads only `docs/diagnostics.md`, so nothing regresses. IDs `GS0503/0504/0505/0507/0508` are all absent from this branch: no collision with #2912 (`GS0507`) or #2923 (`GS0508`).

---

## 4. NIT cleanups — verified, no cost

**A4 pinned.** Mutation **M3** (`ldc.i4.0; conv.u` → `ldnull` at `MethodBodyEmitter.Statements.cs:82-83`) is build-green and turns `PinKinds_ZeroLengthAndNestedConstructs_VerifyAndRun` **RED**. The assertion (`Issue2900FixedPinReleaseEmitTests.cs:545-555`) reads the single `Finally` region of `SpanPin`, slices the handler by `HandlerOffset`/`HandlerLength`, and asserts `Ldc_I4_0` then `Conv_U` — exactly the opcodes ADR-0125 documents at `Statements.cs:78-81`. Correctly targeted, not tautological.

**F2/F4 deletion safe.** Probed span-pin-in-async (`PinnableReference` kind + later `await`), array-pin-in-async across two `await`s, string-pin-in-async, string-pin-in-iterator: all correct (`10 / 8`, `65 / 66`), all **byte-identical** to `eb38ef4e`, all ilverify-clean. Control mutation **M6** (drop the surviving `fixedLocals.Add(node.PointerVariable)` at `AsyncCaptureWalker.cs:165`) still turns `AsyncAndIterator_WithLaterSuspension_VerifyAndRun` **RED**, so the one line that matters is still pinned after the other two were removed. Correct ponytail outcome.

**Round-3 branch-granularity count: 3 pins, 2 guards** — not 3 / 0.

| id | mutation | result |
|---|---|---|
| M1 | GS0506 `await for` alternative | **PIN** (1 red) |
| M2 | GS0506 `await using` alternative | **PIN** (1 red) |
| M3 | `PinnableReference` release opcodes | **PIN** (1 red) |
| M4 | lazy → eager `DefineLabel` | **GUARD** (11/11 green) |
| M5 | drop `FunctionEmitter` epilogue call | **GUARD** (11/11 green) |
| M6 | drop `PointerVariable` registration (control) | **PIN** (1 red, unchanged) |
| M7 | epilogue gate `HasValue` → `hasFixedReturns` | inert (green) |

No build-red rows were counted; all seven mutants compiled under `-warnaserror`.

---

## 5. NIT — the PR body's mutation table is stale at `b00e2121`

Your relayed summary is right, but the PR body itself was not updated and now contains rows that cannot be reproduced:

- Guards `AsyncCaptureWalker.cs:165` ("stop registering pinned temporary") and `:167-170` ("stop registering pinnable source temporary") — **that code was deleted in round 3**. `:165` is now `fixedLocals.Add(node.PointerVariable);`, which is the *pin* row.
- Guard `MethodBodyEmitter.cs:103-105` says "alter **readonly** return-state field scaffolding" — `fixedReturnLabel` is no longer `readonly`.
- Pin `MethodBodyEmitter.cs:202-204` "disable fixed-return state" — the eager `DefineLabel()` that lived there is gone.
- "final unmutated control run passed 10/10" and "issue #2900 tests: 9 passed" — HEAD has **11**.
- The three round-3 pins are absent; the coarse "disable await detection" row remains.
- The lazy label and the `FunctionEmitter` epilogue call appear in neither table.

Cheapest fix: relabel the table "as of `eb38ef4e`" and append a short round-3 delta (3 pins / 2 guards), or refresh the six rows. Not blocking, but the table is the durable review record and four rows are currently unreproducible.

---

## 6. Merge-order — re-derived at `b00e2121`

`git merge-base HEAD origin/main` == `origin/main` == `d4038f1d`. **No rebase required.**

| PR | `git merge-tree` vs `b00e2121` |
|---|---|
| #2912 | **clean** |
| #2929 | **clean** |
| #2932 | **clean** |
| #2935 | **clean** |
| #2908 | 2 conflicts — `DiagnosticDescriptors.cs`, `website/docs/ref/diagnostics.md` |
| #2911 | 2 conflicts — same two files |
| #2923 | 2 conflicts — same two files (`FunctionEmitter.cs` / `MethodBodyEmitter.cs` / `ReflectionMetadataEmitter.cs` all **auto-merge**) |
| #2926 | 1 conflict — `MethodBodyEmitter.cs` |

**New in round 3:** #2908, #2911 and #2923 now conflict where they did not before. Cause is entirely the new GS0506 descriptor line and the new website doc section — both are adjacent appends. Confirmed by inspecting the hunk:

```
<<<<<<< HEAD
    ... FixedStatementCannotSuspend = new("GS0506", ...);
=======
    ... UnconstrainedNullableSequenceElement = new("GS0508", ...);
>>>>>>> refs/prs/2923
```

Resolution: keep both, ordinal order. Fully mechanical. Same for the doc file (both append a section at EOF).

**#2926 is unchanged in shape and your round-2 recipe is still correct.** I re-inspected the merged tree at round-3 HEAD: the conflict is the same `EmitMethodBody` block, and the `EmitFixedReturnEpilogue()` call text is byte-identical, so the lazy label does not stale the recipe. The comment already posted on the PR matches what I derived independently, including "do not reorder or drop the early return". One improvement worth noting: with the lazy label, a bad resolution that drops the early return now still fails **loudly** with `GS9998: Label N has not been marked` rather than silently — the recipe's risk profile improved slightly.

---

## 7. Test counts — settled

Both runs to completion, `--no-build`, Release, on this machine (run concurrently, hence the ~1h durations):

| commit | Compiler.Tests |
|---|---|
| **`d4038f1d` (`origin/main`)** | **3708 passed, 0 failed, 0 skipped** (1 h 07 m) |
| **`b00e2121` (HEAD)** | **3719 passed, 0 failed, 0 skipped** (1 h 03 m) |

Delta is exactly **+11**, which is exactly the size of `Issue2900FixedPinReleaseEmitTests.cs`. Author's 3719 is confirmed; main is **3708**. The other reported numbers (3711/3713/3721/3734/3741/3749) are not reproducible at `d4038f1d` with an unfiltered `Compiler.Tests` run.

Also green at HEAD: `-warnaserror` solution build (0 warnings), Issue2900 filter 11/11, fixed matrix 50/50.

---

## 8. Claims I tried to falsify and could not

- Lazy label can be defined without being marked — 8 emitter sites enumerated, no escape path; 0 `PROBE-GAP` / `PROBE-DOUBLEMARK` across 148 corpus files + 50 tests + 10 probes.
- The lazy label silently dropped a trailing epilogue somewhere — 139/139 compilable corpus files byte-identical and diagnostic-identical to `eb38ef4e`, plus 8 probes identical.
- `FunctionEmitter.cs:177` is reachable — needed two independent lowerer/planner switches disabled to fire it once.
- Deleting the `AsyncCaptureWalker` `PinnedVariable`/`SourceVariable` lines broke a pin-in-async shape — four shapes probed, all byte-identical and correct; the surviving `PointerVariable` line is still pinned.
- GS0506 under-fires via the pin source expression, a nested `fixed`, or a `defer` — all closed.
- New invalid-IL shape in the shipped compiler — none found; ilverify clean on all ten probes after the unmodified ignore list.

---

## 9. Not this PR

`c1.gs` (`await` in a pin source) yields a misleading GS0401 message claiming `System.Int32[]` is not pinnable — **identical on `d4038f1d`**. Pre-existing binder quirk, orthogonal. Worth an issue only if someone cares; it does not gate anything here.

---

## Final tree state

```
$ cd /Users/davidobando/GitHub/DavidObando/gsharp-wt/i2900 && git status --porcelain
$
```

Empty. All four instrumented files (`MethodBodyEmitter.cs`, `FunctionEmitter.cs`, `MoveNextBodyRewriter.cs`, `ReflectionMetadataEmitter.cs`) restored from backup and the solution rebuilt green; all seven mutations restored; the eight temporary `refs/prs/*` refs deleted. Scratch confined to `/tmp/d2900r3/`. No e2e run, nothing written to `~/.nuget/packages/gsharp.net.sdk/`, no `git stash`, main checkout and sibling worktrees untouched.

**Merge it.**

PR #2923, opened 2026-07-30
Fix iterator loop variable capture

### LINKED ISSUE #2907: Iterator loop variable capture fails with GS9998
## Repro

```gs
package IteratorCapture

func values() sequence[int32] {
    for i in 0 ... 1 {
        var read = () -> { return i }
        yield read()
    }
}
```

Compile as a library:

```text
error GS9998: InvalidOperationException: Variable 'i' has no local slot or parameter index in the current method.
```

This reproduces on `origin/main` and affects captures inside iterator functions for ellipsis, collection `for-in`, and three-part loops.

## Expected

Iterator loop-variable captures compile with the same capture semantics as equivalent non-iterator functions.

## Root-cause pointer

`CaptureBoxingRewriter` runs before `IteratorRewriter`, but the iterator/closure lowering combination does not make the captured variable or synthesized box local available to the emitted lambda. The failure surfaces in `MethodBodyEmitter.MemberAccess.cs` when neither a local/parameter slot nor an enclosing-closure capture field exists.

DavidObando:
Fixes #2907

Related nullable-sequence defect: #2927

## #2907 iterator capture

Commit `4a7ea077` is unchanged. Synchronous iterator MoveNext bodies register their existing hoisted-field map in the shared iterator emit context, allowing closure materialization to load captured loop cells from state-machine fields.

The context rename is now complete (`IteratorEmitContext` / `iteratorEmitCtx`). Its registration comment names all three consumers: closure materialization, state-machine field-token resolution, and await emission. The latter only fires for async iterators.

- committed matrix: **16 pins / 4 guards**, all 20 pass
- `4a7ea077` vs final HEAD: **18/18 byte-identical**, diagnostics identical, all 18 final assemblies execute with expected output
- build-green mutation using an empty registered field map: **16 failed / 4 passed**

## #2927 nullable sequence metadata

The concrete fix remains: nullable value-type CLR lifting plus the four symbolic iterator metadata clauses. The erased-runtime compatibility shim was removed completely:

- no `GetWithErasedNullableRuntime`
- no `LiftNullable` cache key dimension
- no erased substitution branch in `Binder.SubstituteType`

`sequence[T?]` / `async sequence[T?]` now reports one bind-time **GS0508** when `T` lacks a value/reference representation constraint. `T struct`, `T class`, and base-class-constrained forms remain supported.

| shape | final HEAD |
|---|---|
| concrete parameter (`mix1`) | one GS0508, no assembly |
| annotated variable (`mix2`) | one GS0508, no assembly |
| passthrough function (`mix5`) | one GS0508, no assembly |
| direct enumerate (`optT`) | one GS0508, no assembly |

The positive #2927 tests run ilverify, `Assembly.Load`, `GetTypes()`, and the emitted program. Diagnostic tests explicitly omit ilverify because GS0508 prevents assembly emission.

- committed matrix: **9 pins / 4 guards**
- guards: `class`-constrained `T?`, base-class-constrained `T?`, `sequence[string?]`, and `sequence[E?]`
- #2927 production diff: **+34/-7** (41 changed lines), down from **+54/-19** (73 changed lines)
- #2927 remains open for generic representation/specialization work

ADR-0040 remains true: `sequence[T]` has one CLR representation matching `IEnumerable<T>`, including when both spellings are mixed in one program. Main violated that interchangeability for `sequence[int32?]` → `IEnumerable[int32?]` (`GS0154`); this PR restores it.

## Mutation audit

Every behavior mutant built successfully before tests ran; every control passed first in the same worktree.

| mutation | detection |
|---|---:|
| disable GS0508 branch | 5/12 failed |
| broaden GS0508 to constrained type parameters | 2/12 failed |
| force the base-class escape hatch closed (M7c) | 1/13 failed |
| remove sync nullable CLR lift | 3/12 failed |
| remove async nullable CLR lift | 1/12 failed |
| remove sync symbolic interface clause | 1/12 failed |
| remove async symbolic interface clause | 1/12 failed |
| remove sync symbolic GetEnumerator clause | 1/12 failed |
| remove async symbolic GetAsyncEnumerator clause | 1/12 failed |
| register empty #2907 field map | 16/20 failed |
| make four sync diagnostic probes constrained | 4/12 failed |
| make async diagnostic probe constrained | 1/12 failed |

Build-green structural mutants reintroducing each erased Binder branch and each cache/factory dimension were rejected by the zero-occurrence source invariant for `eraseNullableRuntime`, `LiftNullable`, and `GetWithErasedNullableRuntime`.

## Broad validation

Sample corpus is weak signal: only 3/148 files contain `yield`, and 1/148 contains `sequence[`.

- corpus: **TOTAL=148 SUCCESS=139**, 9 standalone failures identical
- base vs `4a7ea077`: **BYTEDIFF=0 DIAGDIFF=0 RCDIFF=0**
- base vs final HEAD: **BYTEDIFF=0 DIAGDIFF=0 RCDIFF=0**
- `4a7ea077` vs final HEAD: **BYTEDIFF=0 DIAGDIFF=0 RCDIFF=0**
- Release warnings-as-errors build: **0 warnings / 0 errors**
- Core.Tests: **6913 passed, 1 skipped**
- Compiler.Tests: **3742 passed** (full-clone `origin/main` baseline: **3708**; delta: +34)
- Interpreter.Tests: **453 passed**
- e2e not run locally; shared SDK package cache untouched; CI e2e passed

## Separate defect

Filed #2931 (`Oahu`): a temporary struct receiver used by an iterator returns nondeterministic garbage. It reproduces on `origin/main` and is independent of #2907.


DavidObando:
## Round-3 review — PR #2923 (`62db85e5`), issue #2907 (+#2927 partial)

**Verdict: MERGEABLE.** Both BLOCKERs genuinely closed. Two non-blocking SHOULD-FIXes, neither a correctness bug. Nothing previously cleared was disturbed.

### BLOCKER 1 — CLOSED, and the gate is *complete*, not merely broad
The 4-shape matrix at HEAD gives exactly one `GS0508` each, rc=1, **no assembly emitted** — so `EntryPointNotFoundException` is structurally unreachable, there being no artifact to load.

More important than the matrix was hunting a **bypass**, since the gate lives in `BindTypeClause` (`Binder.cs:2924`) and only holds if `sequence[T?]` cannot be formed another way:

| bypass attempt | result |
|---|---|
| inferred iterator return type | `GS0136` — iterators **require** a declared `sequence[…]`; no inference path |
| type alias `type OptSeq[T] = sequence[T?]` | `GS0005` — G# has no type aliases |
| generic class field | GS0508 ×1, no assembly |
| function-type parameter | GS0508 ×1, no assembly |
| struct field + `sequence[sequence[T?]]` | GS0508 ×2 (2 occurrences), no assembly |

A sequence type can only enter via `sequence[…]` syntax and `BindTypeClause` is its sole binder. Could not falsify.

### BLOCKER 2 — CLOSED
The vacuous guard is gone. Re-tested the *pattern* rather than the name: `SyncNullableIteratorElementCaptureLoadsVerifiesAndRuns` has no inline `IlVerifier`/`Assembly.Load`, but `AssertParity` (`:350-366`) does the full triple — `IlVerifier.Verify` + `Assembly.Load(File.ReadAllBytes())` + `GetTypes()` + bounded execute against expected output. Names are honest. GS0508 cases carry an explicit no-assembly comment.

**On turning three working programs into compile errors — I still agree**, reinforced by two new facts: `git grep` finds **zero** tracked `.gs` using `sequence[…?]` at all, and the 148-file corpus produced **zero** GS0508. The only newly-rejected working shape is an unconstrained `T` instantiated solely with reference types; the remedy is a one-token `[T class]`, verified to compile and produce identical output.

### GS0508 scope — no over-fire
Accepted and working: `[T struct]` (yields `5 nil` — the right value), `[T class]`, `[T Animal]` base-class, all async twins, and every concrete win preserved (`int32?`, `string?`, `E?`, `List[int32?]` param, async). Correctly rejects `[T any]`. Does **not** fire on non-sequence `T?`: locals, params, returns, fields, `List[T?]`, `Map[K,V?]`, `[]T?`, `sequence[List[T?]]`. Exactly one diagnostic per source occurrence — a three-instantiation probe gives **1** at the declaration site, no cascade, no GS0129 double-report. Both documented remedies verified working.

**Bonus finding — ADR-0040 was broken on `main` and this PR fixes it:**
```
base d4038f1d: GS0154: Parameter 's' requires 'IEnumerable[System.Nullable[int32]]'
                       but was given 'sequence[int32?]'      ← ADR-0040 violated on main
HEAD 62db85e5: rc=0, out=[2 2]                                ← spellings interchangeable
```

### Nothing cleared was disturbed
`4a7ea077` is an ancestor of HEAD at an unchanged SHA; round 3 touched only `Issue2927…Tests.cs` and `IlVerifierTests.cs`. **18/18 byte-identical holds at HEAD** (`TOTAL=18 SAME=18 BYTEDIFF=0 DIAGDIFF=0`) — the collapse *and* the full rename are codegen-neutral. Corpus: `TOTAL=148 SAME=148 BYTEDIFF=0 DIAGDIFF=0 FAIL=9/9`, **zero GS0508**, zero rc drift. Falsification attempt on the rename — 2 sync + 1 async iterator with closures in one program sharing the renamed `IteratorEmitContexts` — base `GS9998`, HEAD rc=0, ilverify clean, correct output, no cross-talk. Pins/guards on base: #2907 **16 fail / 4 pass**, #2927 **9 fail / 3 pass**, author's counts exact. Mutants `M1`–`M7b` (build-green, restore+rebuild between, control 32/32 first) all detected.

### SHOULD-FIX 1 — unpinned branch in the new rejecting code
`src/Core/CodeAnalysis/Binding/Binder.cs:2927` — `&& typeParameter.ClassConstraint == null`. Mutant `M7c` (`ClassConstraint == null` → `true`) is **build-green and undetected**:
```
HEAD pristine, [T Animal] probe : GS0508 count = 0                              (correct)
M7c mutant,    same probe       : error GS0508: ... 'T' is not constrained ...   (over-fire)
committed suite on M7c mutant   : Passed! 32/32                                 (no detection)
```
The base-class-constraint escape hatch — which `docs/diagnostics.md:960` explicitly advertises as a remedy — has zero regression coverage. One guard test with `[T SomeClass]` in `Issue2927NullableSequenceElementTypeTests.cs` fixes it; the file has `class`- and `struct`-constrained positives at `:163`/`:187` but no base-class one. Makes #2927 **9 pins / 4 guards**.

### SHOULD-FIX 2 — issue #2931 is right on its claim but wrong in its title
Settling the discrepancy I raised in round 2: **the issue is accurate and my earlier read was wrong.** My old control contained a lambda capturing `this` — a different defect. The issue's control does not.
```
z1 Box{N:100}.vals()  iterator, temporary : 1876794024   ← garbage, as the issue says
z2 Box{N:100}.val()   non-iterator ctrl   : 100          ← correct, as the issue says
z3 var b; b.vals()    iterator, LOCAL var : 101205592    ← ALSO garbage
z4 var b; b.val()     non-iterator ctrl   : 100          ← correct
```
`z3` shows the title "**temporary** receiver" is too narrow — a plain local receiver is equally corrupted, so someone will hunt a temporary-lifetime bug that is not there. Retitle to "struct receiver" and add `z3`. Out of scope for this PR.

### SF/NIT status — all done
`StateMachineEmitter.cs:664-671` names all three `ResolveCurrentStateMachineFieldToken` consumers ✓ · rename finished and the false doc line at `ReflectionMetadataEmitter.cs:4159` corrected ✓ · PR body cites **ADR-0040** (line 39), no ADR-0069 ✓ · `IlVerifierTests.cs` dead fat-header branch → `var headerSize = 1;` ✓ · corpus weakness stated (line 63), e2e wording correct (line 73), `Fixes #2907` / `Related … #2927` (lines 1, 3), **#2927 stays OPEN** ✓

### Merge order — better than round 2
| PR | result |
|---|---|
| **#2912** | **CLEAN** — `Binder.cs` coupling **fully dissolved** (`SubstituteType` reverted to `origin/main` text) |
| **#2929** | **CLEAN** |
| #2926, #2932, #2935 | CLEAN |
| #2908, #2911 | conflict only in `DiagnosticDescriptors.cs` + both diagnostics docs |
| **#2917** | conflict only in `DiagnosticDescriptors.cs` + `website/docs/ref/diagnostics.md` — **`FunctionEmitter.cs`, `MethodBodyEmitter.cs`, `ReflectionMetadataEmitter.cs` all auto-merge**, so the rename costs nothing |

No diagnostic ID collisions: HEAD `GS0508` vs #2908 `GS0503`, #2911 `GS0504`, #2917 `GS0506`. Every conflict is append-adjacency — "take both", zero judgement.

### Suites
HEAD: Core **6913 / 1 skipped**, Compiler **3741 / 0 failed** (1h11m), Interpreter **453 / 0** — all match the author. On the contested `origin/main` Compiler baseline I measure a **count of 3709** (`3741 − 3709 = 32` = exactly this PR's 32 new tests), with the caveat stated plainly: my baseline tree is a `git archive` extract with no `.git`, where 19 golden-sample/constraint tests fail for path reasons — an artifact of that tree, **not** a claim that main is red. Two other reviewers measured 3708 in full clones; the 1-test gap is immaterial to this PR's delta.

### Claims I tried to falsify and could not
GS0508 can be bypassed to reach emit — **no** (6 attempts, one syntactic entry point). The rename lets async-iterator state reach sync iterators — **no** (3 state machines in one program, clean). The collapse changed codegen — **no** (18/18 byte-identical at HEAD). GS0508 over-fires on real code — **no** (0/148 corpus, 0 tracked `.gs` affected). #2931 misattributes its control — **no**, it is correct and my earlier read was the error.

Issue #2922, opened 2026-07-30
gsc: fixed containing switch emits ExpectedNumericType IL


DavidObando:
## Repro

```gsharp
package FixedSwitchInvalidIl

func F(x int32, xs []int32) int32 {
    unsafe {
        fixed p *int32 = xs {
            switch x {
                case 0 { return xs.Length }
                default { return 1 }
            }
        }
    }
}

public var result = F(0, []int32{1, 2})
```

## Current

Compilation succeeds and execution returns normally, but ILVerify rejects `F`:

```text
[IL]: Error [ExpectedNumericType]: ... F(int32, int32[])[offset 0x0000001F][found address of Int32] Expected numeric type on the stack.
```

## Expected

Emitted fixed/pointer control flow should satisfy the CLI verifier.

## Notes

Reproduces byte-identically on `main` and #2909 before/after its CFG-only changes. This is a pre-existing fixed/pointer emission defect, orthogonal to definite-return analysis; #2909 does not change lowering or emission.

DavidObando:
## Investigation result: gsc is matching Roslyn here — the array/span premise does not hold

An Opus 5 review of PR #2964 established that the `ExpectedNumericType` diagnostic in this issue is **not** a G# emitter defect for the array and span pin paths.

The equivalent C# program compiled with `csc` (.NET 10, `AllowUnsafeBlocks`, both Debug and Release) produces the identical diagnostic:

```
[ExpectedNumericType] P::F(int32,int32[]) [offset 0x17][found address of Int32]
[ExpectedNumericType] P::H(Span`1<int32>)  [offset 0x0A][found address of Int32]
```

Roslyn's IL for `fixed (int* p = xs)` is `ldloc.1; ldc.i4.0; ldelema Int32; conv.u; stloc.0` with locals `[1] int32[] pinned` — the same shape gsc emits, producing `[ExpectedNumericType][offset 0x1F]`, the exact offset reported above.

ECMA-335 III.1.6 (Table 8, Conversion Operations) permits `conv.u`/`conv.i` on `&` operands. The result is *unverifiable*, not illegal, and unverifiable IL is the normal and expected output of `unsafe` code. The acceptance criterion "emitted fixed/pointer control flow should satisfy the CLI verifier" is unachievable by construction, because dereferencing `*p` can never be verifiable.

This repository already has a written precedent for exactly this situation at `test/Compiler.Tests/IlVerifier.cs:190-206` (`KnownIssues.RefStruct`):

[...]

Issue #2921, opened 2026-07-30
gsc: enforce GS0100 on lambda bodies


DavidObando:
## Repro: switch escape

```gsharp
package LambdaMissingReturn

func G(f (int32) -> int32) int32 { return f(1) }

func F(x int32) int32 {
    return G(func(v int32) int32 {
        for {
            switch v {
                case 1 { break }
                default { return 1 }
            }
        }
    })
}

public var result = F(0)
```

## Repro: plain conditional

```gsharp
package LambdaMissingReturn

func G(f (int32) -> int32) int32 { return f(1) }

func F() int32 {
    return G(func(v int32) int32 {
        if v == 0 { return 1 }
    })
}

public var result = F()
```

## Current

Both compile and emit assemblies. ILVerify reports `ReturnMissing`; invoking the lambda throws `InvalidProgramException`.

## Expected

Each lambda reports `GS0100: Not all code paths return a value.` and emits nothing.

## Root cause

`src/Core/CodeAnalysis/Binding/LambdaBinder.cs` near the `bodyNeverCompletesNormally` calculation calls `ControlFlowGraph.AllPathsReturn` only for return-type inference (#891). It never enforces definite return for a non-void lambda body.

The arm-aware definite-return projection from #2909 is the right CFG foundation for the switch shape, but lambda diagnostics must invoke it independently. Reproduces on `main`; not caused by #2909.

DavidObando:
Update after #2926: the issue's Current section is stale on base 17789519. The repro assembly now loads and contains verifiable IL, then child execution exits 134 with `InvalidOperationException: Compiler-generated guard reached: non-void function fell through without returning a value.` The older `InvalidProgramException` / `ReturnMissing` symptom predates #2926.

Issue #738, opened 2026-06-12
Interpreter: return inside switch arm exits switch, not the enclosing function


DavidObando:
## Summary

In the interpreter, executing a `return <value>` statement inside a `switch` arm body terminates the switch but does NOT terminate the enclosing function — control falls through to whatever follows the switch.

Discovered while implementing #712 (switch-arm flow narrowing). Reproduces on `main` independent of that PR.

## Repro

```
func classify(x: object): string {
    switch x {
        case s: string -> return "string"
        default -> return "other"
    }
    // interpreter falls through to here instead of returning above
    return "unreached"
}
```

## Observed

Interpreter returns `"unreached"` when given a string argument.

## Expected

Interpreter returns `"string"` (matching the emit/JIT backend, which is correct).

## Context

- Discovered while writing tests for switch-arm flow narrowing in #712 (PR #736); worked around by re-shaping the test to a single non-exiting arm.
- Emit backend is correct; this is interpreter-only.

## Plan reference

Related to #706 §6. Tracked as follow-up from #712 (PR #736).

PR #2926, opened 2026-07-31
Fix definite return for try regions and exhaustive switches

### LINKED ISSUE #2891: gsc: break from try/finally bypasses GS0100 and emits invalid IL
Found while fixing #2883.

**Verified NOT a regression:** reproduces on current `main` before the #2883 fix.

## Summary

A `break` that exits a loop from inside a `try` body is invisible to definite-return analysis. A non-void function compiles without `GS0100`, emits an assembly, and throws `InvalidProgramException` when invoked.

## Repro (verified)

```gsharp
package TryBreakFlow

func F() int32 {
    for {
        try {
            break
        } finally {
        }
    }
}

public var result = F()
```

Observed:

- gsc exits 0 and emits the assembly.
- Running the assembly throws `InvalidProgramException` from `F`.

## Expected

The loop can complete normally after the `break`, so `F` reports `GS0100: Not all code paths return a value.` No invalid assembly is emitted.

## Root cause

`ControlFlowGraph.BasicBlockBuilder` treats `BoundTryStatement` as opaque. The lowered goto inside the try body cannot resolve the enclosing loop's break label, so the loop-exit edge is absent.

This is related to #2883, but requires exception-region-aware CFG modeling. Sequentially projecting try/catch/finally bodies would be incorrect because handlers are alternatives and `finally` executes on every exit.

- Break, continue, labeled break/continue, and goto escaping try/catch/finally regions.
- Mixed return/throw/escape paths through catches and finally blocks.
- Return-after-loop controls that compile, verify, and run.
- No branch targets past the end of a non-void method body.

### LINKED ISSUE #2906: gsc: exhaustive enum switch still reports GS0100 without default
## Summary

A value-returning function whose pattern switch covers every enum member still reports `GS0100` unless it adds a redundant `default` arm.

## Repro

```gsharp
package EnumReturnFlow

enum E { A, B }

func F(x E) int32 {
    switch x {
        case E.A { return 1 }
        case E.B { return 2 }
    }
}
```

## Observed

`GS0100: Not all code paths return a value.` at `F`.

## Expected

No `GS0100`; `ExhaustivenessAnalyzer` already accepts the switch as exhaustive.

## Root cause

`ControlFlowGraph` definite-return logic recognizes only `default` (and, after #2890, total discard) as eliminating the no-match edge. Exhaustiveness proven by `ExhaustivenessAnalyzer` is not carried on `BoundPatternSwitchStatement`, so CFG cannot remove that edge for closed enum/sealed hierarchies.

DavidObando:
Fixes #2891
Fixes #2906

Follow-ups: #2940, #2941, #2942

### #2891 — definite return through protected regions

Definite-return analysis treated lowered `BoundTryStatement` nodes as opaque, hiding `break`, `continue`, `goto`, and protected returns from the enclosing CFG.

The lowering pipeline now preserves a pre-emit body before `FinallyExitRewriter` transforms escaping-finally control flow. `AllPathsReturn`, `RefKindDefiniteAssignmentAnalyzer`, and `RefStructAsyncLivenessAnalyzer` use that pre-emit body. Definite-return projection models try/catch alternatives, exceptional finally entry, escaping transfers through cloned finally bodies, and protected returns through the `$returnTemp` epilogue.

Method-exit detection searches top-level declarations instead of assuming `$returnTemp` is statement zero; protected-region entry rewriting may prepend `<>ai_seh_branch`.

CLR forbids `leave` from a finally handler. `FinallyExitRewriter` lifts finally bodies containing escaping branches, funnels try/catch exits through the lifted body, captures pending exceptions, preserves rethrow stacks with `ExceptionDispatchInfo`, and dispatches the pending branch afterward. Shared `ProtectedRegionBranchAnalysis` supplies label/branch/try-region bookkeeping to both protected-region rewriters.

ADR-0139 documents that an escaping transfer from `finally` replaces and discards a pending exception, matching the interpreter and ECMA-335 III.1.7.5.

### #2906 — exhaustive enum switch and definite return

`ExhaustivenessAnalyzer` proved closed enum/sealed-hierarchy switches exhaustive, but the proof was discarded before CFG construction. `BoundPatternSwitchStatement.IsExhaustive` now carries the proof through rewriting and CFG removes the impossible no-match edge for definite-return analysis.

The defensive throw exists only at a non-void method tail. Ordinary switch statements retain no-match fallthrough, matching the interpreter. The runtime exception identifies itself as a compiler-generated missing-return guard.

The attempted imported CLR `[Flags]` carve-out was reverted. Existing statement/expression diagnostics remain unchanged; consistent flags semantics are tracked in #2941.

## Review corrections

- Deleted the seven-line per-switch throw and its two regression-pinning tests.
- Added five compiler/interpreter fallthrough cases plus a dedicated Interpreter.Tests case.
- Added child-process execution guards for local gotos in a try, nested finally, and nested catch. These catch runtime-invalid IL that `ilverify` can accept.
- Added the nested escaping-finally GS0238 no-drift case that pins use of the pre-emit analysis body.
- Rebased onto `origin/main` `e21677f6`; `EmitFixedReturnEpilogue()` remains first with its early return, followed by the missing-return guard.
- Added the fixed-return/exhaustive-switch interaction case, documenting the current default-slot result for an unnamed value.
- Deleted the unobservable local-label short-circuit in definite-return projection rather than retaining an unpinned branch.
- Updated #2940 to describe an analysis-metadata invariant, not a current runtime/IL hole.

## Coverage metrics

Round-2 recount found **68 pins / 71 independent guards** over the prior 139 Core+Compiler issue cases. This update adds three mutation-detecting branch-analysis pins plus two independent no-drift/interaction guards:

| Category | Count |
|---|---:|
| Pins | **71** |
| Independent guards | **73** |
| Core+Compiler issue cases | **144** |

The dedicated Interpreter.Tests parity case is additional and excluded from that table.

## Validation at `1960b64de15a280ab118f2d634cd3bdf64594d00`

- Release static-graph build: **0 warnings, 0 errors**.
- Issue-focused Core (`Issue2891|Issue2906`): **91 passed**.
- Interaction filter (`Issue2891|Issue2906|Issue2900Fixed`): **64 Compiler tests passed**.
- Interpreter.Tests: **469 passed**.
- Core.Tests: **7004 passed, 1 skipped**.
- Compiler.Tests: **3809 passed**.

### DIFF
--- a/docs/adr/0139-goto-and-general-labels.md
+++ b/docs/adr/0139-goto-and-general-labels.md
@@ -74,6 +74,10 @@ regions separately, however: a branch that leaves `try` or `catch` must run
 `finally` before reaching its target, and a branch that leaves `finally`
 cannot be emitted as CLR `leave` from the handler. The lowering pass therefore
 funnels those exits through a lifted finally body before dispatching them.
+If that lifted `finally` transfers control with `break`, `continue`, `goto`,
+or `return`, the transfer replaces the pending exception, so the exception is
+discarded. This matches evaluator behavior and ECMA-335 III.1.7.5; C# instead
+rejects transfers out of a `finally` with CS0157.
 
 ### Diagnostics
 
--- a/docs/adr/0139-goto-and-general-labels.md
+++ b/docs/adr/0139-goto-and-general-labels.md
@@ -68,10 +68,12 @@ introduces no new binder scope — `label: var x = 1` still declares `x` into
 the enclosing block, exactly like C#.
 
 No new `BoundNodeKind` is introduced (`BoundGotoStatement`/
-`BoundLabelStatement` already existed), so `ControlFlowGraph`,
-`RefKindDefiniteAssignmentAnalyzer`, `SlotPlanner`, `MethodBodyEmitter`, and
-the evaluator needed no changes — they already handle these bound nodes
-generically.
+`BoundLabelStatement` already existed). The emitter and evaluator handle the
+nodes generically. Definite-return analysis projects protected exception
+regions separately, however: a branch that leaves `try` or `catch` must run
+`finally` before reaching its target, and a branch that leaves `finally`
+cannot be emitted as CLR `leave` from the handler. The lowering pass therefore
+funnels those exits through a lifted finally body before dispatching them.
 
 ### Diagnostics
 
@@ -112,8 +114,8 @@ directly:
 
 **Positive.** `cs2gs` can now migrate C#'s full `goto`/label surface,
 including the `goto case`/`goto default` fall-through forms, with faithful
-semantics. No gsc runtime or emit changes were needed — only surface syntax
-and a label-name binder.
+semantics. Ordinary labels remain generic bound nodes; protected-region exits
+receive the exception-flow lowering required by CLR handler rules.
 
 **Negative.** Unstructured control flow is easy to write incorrectly by hand;
 G# still has no `goto`-into-scope diagnostic beyond "does not parse as a
--- a/docs/adr/0139-goto-and-general-labels.md
+++ b/docs/adr/0139-goto-and-general-labels.md
@@ -78,6 +78,9 @@ If that lifted `finally` transfers control with `break`, `continue`, `goto`,
 or `return`, the transfer replaces the pending exception, so the exception is
 discarded. This matches evaluator behavior and ECMA-335 III.1.7.5; C# instead
 rejects transfers out of a `finally` with CS0157.
+Known limitation [#2953](https://github.com/DavidObando/gsharp/issues/2953):
+`fixed`-containing non-void functions bypass the fall-through guard and
+silently return the fixed-return epilogue's default slot instead of throwing.
 
 ### Diagnostics
 
--- a/src/Core/CodeAnalysis/Binding/Binder.cs
+++ b/src/Core/CodeAnalysis/Binding/Binder.cs
@@ -1388,11 +1388,17 @@ public sealed class Binder
 
                     // ADR-0060 items #4/#5: out-parameter definite-assignment and
                     // 'ref'-arg unassigned-before-read checks.
-                    RefKindDefiniteAssignmentAnalyzer.Analyze(lowered, function, binder.Diagnostics);
+                    RefKindDefiniteAssignmentAnalyzer.Analyze(
+                        lowered.PreEmitAnalysisBody ?? lowered,
+                        function,
+                        binder.Diagnostics);
 
                     // Issue #2350: by-ref-like locals are only legal in an
                     // async function body when never live across an `await`.
-                    RefStructAsyncLivenessAnalyzer.Analyze(lowered, function, binder.Diagnostics);
+                    RefStructAsyncLivenessAnalyzer.Analyze(
+                        lowered.PreEmitAnalysisBody ?? lowered,
+                        function,
+                        binder.Diagnostics);
 
                     return (lowered, binder.Diagnostics.ToImmutableArray());
                 }));
@@ -1441,7 +1447,10 @@ public sealed class Binder
 
                     // Issue #2350: by-ref-like locals are only legal in an
                     // async function body when never live across an `await`.
-                    RefStructAsyncLivenessAnalyzer.Analyze(lowered, method, binder.Diagnostics);
+                    RefStructAsyncLivenessAnalyzer.Analyze(
+                        lowered.PreEmitAnalysisBody ?? lowered,
+                        method,
+                        binder.Diagnostics);
 
                     return (lowered, binder.Diagnostics.ToImmutableArray());
                 }));
@@ -1941,7 +1950,10 @@ public sealed class Binder
 
             // Issue #2350: by-ref-like locals are only legal in an async
             // function body when never live across an `await`.
-            RefStructAsyncLivenessAnalyzer.Analyze(lowered, method, binder.Diagnostics);
+            RefStructAsyncLivenessAnalyzer.Analyze(
+                lowered.PreEmitAnalysisBody ?? lowered,
+                method,
+                binder.Diagnostics);
 
             return (lowered, binder.Diagnostics.ToImmutableArray());
         }));
@@ -1991,7 +2003,10 @@ public sealed class Binder
 
             // Issue #2350: by-ref-like locals are only legal in an async
             // function body when never live across an `await`.
-            RefStructAsyncLivenessAnalyzer.Analyze(lowered, accessor, binder.Diagnostics);
+            RefStructAsyncLivenessAnalyzer.Analyze(
+                lowered.PreEmitAnalysisBody ?? lowered,
+                accessor,
+                binder.Diagnostics);
 
             return (lowered, binder.Diagnostics.ToImmutableArray());
         }));
@@ -2042,7 +2057,10 @@ public sealed class Binder
 
             // Issue #2350: by-ref-like locals are only legal in an async
             // function body when never live across an `await`.
-            RefStructAsyncLivenessAnalyzer.Analyze(lowered, member, binder.Diagnostics);
+            RefStructAsyncLivenessAnalyzer.Analyze(
+                lowered.PreEmitAnalysisBody ?? lowered,
+                member,
+                binder.Diagnostics);
 
             return (lowered, binder.Diagnostics.ToImmutableArray());
         }));
@@ -2087,7 +2105,10 @@ public sealed class Binder
 
             // Issue #2350: by-ref-like locals are only legal in an async
             // function body when never live across an `await`.
-            RefStructAsyncLivenessAnalyzer.Analyze(lowered, method, binder.Diagnostics);
+            RefStructAsyncLivenessAnalyzer.Analyze(
+                lowered.PreEmitAnalysisBody ?? lowered,
+                method,
+                binder.Diagnostics);
 
             return (lowered, binder.Diagnostics.ToImmutableArray());
         }));
--- a/src/Core/CodeAnalysis/Binding/BoundBlockStatement.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundBlockStatement.cs
@@ -18,18 +18,18 @@ public sealed class BoundBlockStatement : BoundStatement
     /// <param name="syntax">The originating syntax.</param>
     /// <param name="statements">The immutable array of bound statements.</param>
     public BoundBlockStatement(SyntaxNode syntax, ImmutableArray<BoundStatement> statements)
-        : this(syntax, statements, definiteReturnAnalysisBody: null)
+        : this(syntax, statements, preEmitAnalysisBody: null)
     {
     }
 
     internal BoundBlockStatement(
         SyntaxNode syntax,
         ImmutableArray<BoundStatement> statements,
-        BoundBlockStatement definiteReturnAnalysisBody)
+        BoundBlockStatement preEmitAnalysisBody)
         : base(syntax)
     {
         Statements = statements;
-        DefiniteReturnAnalysisBody = definiteReturnAnalysisBody;
+        PreEmitAnalysisBody = preEmitAnalysisBody;
     }
 
     /// <inheritdoc/>
@@ -41,7 +41,7 @@ public sealed class BoundBlockStatement : BoundStatement
     public ImmutableArray<BoundStatement> Statements { get; }
 
     /// <summary>
-    /// Gets the pre-emit-rewrite body used only by definite-return analysis.
+    /// Gets the pre-emit-rewrite body used by control-flow analyses.
     /// </summary>
-    internal BoundBlockStatement DefiniteReturnAnalysisBody { get; }
+    internal BoundBlockStatement PreEmitAnalysisBody { get; }
 }
--- a/src/Core/CodeAnalysis/Binding/BoundBlockStatement.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundBlockStatement.cs
@@ -18,9 +18,18 @@ public sealed class BoundBlockStatement : BoundStatement
     /// <param name="syntax">The originating syntax.</param>
     /// <param name="statements">The immutable array of bound statements.</param>
     public BoundBlockStatement(SyntaxNode syntax, ImmutableArray<BoundStatement> statements)
+        : this(syntax, statements, definiteReturnAnalysisBody: null)
+    {
+    }
+
+    internal BoundBlockStatement(
+        SyntaxNode syntax,
+        ImmutableArray<BoundStatement> statements,
+        BoundBlockStatement definiteReturnAnalysisBody)
         : base(syntax)
     {
         Statements = statements;
+        DefiniteReturnAnalysisBody = definiteReturnAnalysisBody;
     }
 
     /// <inheritdoc/>
@@ -30,4 +39,9 @@ public sealed class BoundBlockStatement : BoundStatement
     /// Gets the immutable array of bound statements.
     /// </summary>
     public ImmutableArray<BoundStatement> Statements { get; }
+
+    /// <summary>
+    /// Gets the pre-emit-rewrite body used only by definite-return analysis.
+    /// </summary>
+    internal BoundBlockStatement DefiniteReturnAnalysisBody { get; }
 }
--- a/src/Core/CodeAnalysis/Binding/BoundPatternSwitchStatement.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundPatternSwitchStatement.cs
@@ -14,11 +14,17 @@ public sealed class BoundPatternSwitchStatement : BoundStatement
     /// <param name="syntax">The originating syntax.</param>
     /// <param name="discriminant">The discriminant expression.</param>
     /// <param name="arms">The switch arms.</param>
-    public BoundPatternSwitchStatement(SyntaxNode syntax, BoundExpression discriminant, ImmutableArray<BoundPatternSwitchArm> arms)
+    /// <param name="isExhaustive">Whether closed-type analysis proved that the arms cover every value.</param>
+    public BoundPatternSwitchStatement(
+        SyntaxNode syntax,
+        BoundExpression discriminant,
+        ImmutableArray<BoundPatternSwitchArm> arms,
+        bool isExhaustive = false)
         : base(syntax)
     {
         Discriminant = discriminant;
         Arms = arms;
+        IsExhaustive = isExhaustive;
     }
 
     /// <inheritdoc/>
@@ -29,4 +35,7 @@ public sealed class BoundPatternSwitchStatement : BoundStatement
 
     /// <summary>Gets the switch arms.</summary>
     public ImmutableArray<BoundPatternSwitchArm> Arms { get; }
+
+    /// <summary>Gets a value indicating whether closed-type analysis proved that the switch is exhaustive.</summary>
+    public bool IsExhaustive { get; }
 }
--- a/src/Core/CodeAnalysis/Binding/BoundTreeRewriter.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundTreeRewriter.cs
@@ -791,7 +791,7 @@ public abstract class BoundTreeRewriter
             return node;
         }
 
-        return new BoundPatternSwitchStatement(null, discriminant, builder?.MoveToImmutable() ?? node.Arms);
+        return new BoundPatternSwitchStatement(null, discriminant, builder?.MoveToImmutable() ?? node.Arms, node.IsExhaustive);
     }
 
     /// <summary>Rewrites a bound pattern.</summary>
--- a/src/Core/CodeAnalysis/Binding/ControlFlowGraph.cs
+++ b/src/Core/CodeAnalysis/Binding/ControlFlowGraph.cs
@@ -60,11 +60,11 @@ public sealed class ControlFlowGraph
     /// <returns>Whether all its paths return or not.</returns>
     public static bool AllPathsReturn(BoundBlockStatement body)
     {
-        body = body.DefiniteReturnAnalysisBody ?? body;
+        body = body.PreEmitAnalysisBody ?? body;
 
         // This projection is intentionally limited to definite-return. Other
-        // data-flow analyzers call Create(body) and still treat these compound
-        // statements as opaque, so their region-specific behavior is unchanged.
+        // data-flow analyzers use the same pre-emit body but retain their
+        // existing opaque treatment of compound statements.
         var graph = Create(ProjectRegionsForDefiniteReturn(body), treatThrowsAsTerminators: true);
 
         foreach (var branch in graph.End.Incoming)
@@ -212,16 +212,19 @@ public sealed class ControlFlowGraph
         BoundLabel methodExitLabel = null;
         BoundReturnStatement methodExitReturn = null;
 
-        if (statement is BoundBlockStatement functionBody
-            && functionBody.Statements.Length >= 3
-            && functionBody.Statements[0] is BoundVariableDeclaration returnTempDeclaration
-            && functionBody.Statements[^2] is BoundLabelStatement exitLabel
-            && functionBody.Statements[^1] is BoundReturnStatement { Expression: BoundVariableExpression returnExpression } exitReturn
-            && returnTempDeclaration.Variable.Name == "$returnTemp"
-            && ReferenceEquals(returnTempDeclaration.Variable, returnExpression.Variable))
+        if (statement is BoundBlockStatement functionBody && functionBody.Statements.Length >= 3)
         {
-            methodExitLabel = exitLabel.Label;
-            methodExitReturn = exitReturn;
+            var returnTempDeclaration = functionBody.Statements
+                .OfType<BoundVariableDeclaration>()
+                .FirstOrDefault(declaration => declaration.Variable.Name == "$returnTemp");
+            if (returnTempDeclaration != null
+                && functionBody.Statements[^2] is BoundLabelStatement exitLabel
+                && functionBody.Statements[^1] is BoundReturnStatement { Expression: BoundVariableExpression returnExpression } exitReturn
+                && ReferenceEquals(returnTempDeclaration.Variable, returnExpression.Variable))
+            {
+                methodExitLabel = exitLabel.Label;
+                methodExitReturn = exitReturn;
+            }
         }
 
         Add(statement);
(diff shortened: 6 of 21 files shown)


DavidObando:
## Review — PR #2926 (`bb0ee017`), issues #2891 + #2906

**Verdict: NOT MERGEABLE.** Both fixes are real and one is vindicated against my own challenge. But the #2906 half over-fires and breaks working programs, and the fix is to **delete 7 lines**.

### BLOCKER 1 — defensive throw over-fires, breaking working code
`MethodBodyEmitter.Patterns.cs:95-101`. The throw fires on **every** exhaustive-discriminant `switch` statement with no `default` — not only ones carrying a definite return.

| probe | shape | base | HEAD |
|---|---|---|---|
| `overfire1.gs` | non-void, `return 99` **after** switch | `F(99)=99` | **InvalidOperationException**, exit 134 |
| `overfire2.gs` | non-terminal arms, `return value` after | `F(99)=0` | **InvalidOperationException**, exit 134 |
| `overfire3.gs` | **void** func, prints, code after | `done` | **InvalidOperationException**, exit 134 |
| `nullshape.gs` | `sealed interface` + `F(nil)` | `nil=0` | **InvalidOperationException**, exit 134 |

`overfire3` is a **void** function — GS0100 does not apply, so this has nothing to do with #2906. `nullshape` needs no cast; it is an ordinary program. The user's own trailing `return` becomes dead code with no diagnostic saying so.

C# does not do this: a switch **statement** falls through on no match; only a switch **expression** throws (`SwitchExpressionException`). HEAD diverges from C# and from `cs2gs` fidelity (`docs/cs2gs-coverage-matrix.md` translates C# switch → G# switch).

### BLOCKER 2 — compiler/interpreter divergence
Same lowered body, different behaviour. HEAD compiler throws on `nullshape.gs`/`overfire1/2/3`; the interpreter returns the base values. `EvaluatePatternSwitchStatement` never received a matching change.

### BLOCKER 3 — a smaller fix exists, measured
Set `node.IsExhaustive` → `false` at `MethodBodyEmitter.Patterns.cs:95` (delete the PR's `+7`), keep only the method-tail throw at `MethodBodyEmitter.cs:239-246`:

| probe | base | HEAD | **HEAD −7 lines** |
|---|---|---|---|
| `i2906.gs` (the issue) | GS0100 | works, ilverify 0 err | **works, ilverify 0 err** ✅ |
| `i2891.gs` (the issue) | InvalidProgramException | GS0100 | **GS0100** ✅ |
| `enum_oob.gs` (switch **is** the tail) | — | InvalidOperationException | **InvalidOperationException** ✅ kept |
| `overfire1/2/3`, `nullshape` | base value | **throws** | **base value** ✅ |
| corpus 122 samples HEAD↔variant | — | — | **BYTEDIFF=0, EXITDRIFT=0** |
| Core.Tests issue filter | — | 87/87 | **87/87** |
| Compiler.Tests issue filter | — | 46/46 | **44/46** |

The only two casualties are `ExhaustiveEnum_FollowedByCode_StillThrowsOnUnnamedValue` and `CapturedArmState_PreservesExhaustiveEmitterMarker` (`Issue2906ExhaustiveSwitchReturnEmitTests.cs:198,224`) — **both pin the regression itself**. Delete the 7 lines and those two tests.

### BLOCKER 4 — undisclosed `[Flags]` behaviour change
`ExhaustivenessAnalyzer.cs:25-27`, absent from the PR body. `IsExhaustiveDiscriminant` is now `false` for imported CLR `[Flags]` enums, and this leaks to the switch-**expression** path (`ExpressionBinder.SwitchExpr.cs:50,:116`) which got no matching guard.

```
flags_stmt.gs  base: GS0178   HEAD: compiles silently          ← lost diagnostic
flags_expr.gs  base: GS0177   HEAD: GS0176 AND GS0177 together ← contradictory pair
flags_user.gs  base: GS0178   HEAD: GS0178                     ← user @Flags unchanged
```
The `type is EnumSymbol` short-circuit means user-declared `@Flags` enums keep GS0178 while imported CLR `[Flags]` lose it — inconsistent. `docs/diagnostics.md:138` is now false. Removing the carve-out breaks exactly **1** Core test — thin for a language-visible change. Recommend reverting it out and landing it separately with statement and expression paths consistent.

### Ponytail verdict on `FinallyExitRewriter.cs` (+319): **justified**
Three candidate homes were examined. `ProtectedRegionBranchRewriter` handles branches *entering* a protected region; this handles branches *leaving a `finally` handler* — opposite direction, different plan model. `Lowerer`'s `tryNestingDepth`/`$returnTemp` spill *creates* the illegal construct and cannot be the cure. "Just diagnose it" (C# CS0157 analogue, ~20 lines) was rejected because the interpreter already implements the construct deliberately (#738) and runs it correctly on base — a diagnostic would delete working language surface. A fourth alternative (funnel to local `endfinally` + dispatch after try, no catch-all/EDI, est. 140-160 lines) was **not built**, and is a *semantics* change rather than a size change: it yields "exception wins", contradicting the PR's deliberate `ExceptionSuppressedByFinallyBreak/Continue/Goto` tests.

For the chosen semantics, 319 lines is near the honest minimum — `leave` skipping a lifted finally makes the discriminator funnel unavoidable, and running lifted finallys on the exception path makes the catch-all + `ExceptionDispatchInfo` unavoidable. Residual criticism: ~50 lines (`LabelCollector`, `EscapingBranchDetector`, `:442-497`) duplicate `ProtectedRegionBranchRewriter.Planner`'s `labelRegions`/`tryStack` bookkeeping — extract it. Note the file **does not fire** for #2891's own repro (`try { break } finally {}`), which is fixed purely by `AddTry`; it addresses only the "finally contains an escaping branch" sub-case.

### SHOULD-FIX
- **S1** `ControlFlowGraph.cs:216` — GS0100 false negative from a positional heuristic. `ProtectedRegionBranchRewriter.Rewrite` prepends `var <>ai_seh_branch = 0` at index 0 when a branch enters a protected region, so `Statements[0] is BoundVariableDeclaration` misdetects and the epilogue stays in the projection. `gotoin5.gs` (loop + try/finally + `goto` into the try) silently returns 0 with no diagnostic where `nogoto1.gs` correctly reports GS0100. Fix: search the top-level statement list rather than index 0 (~17 lines, verified: `gotoin3/4/5` → GS0100, Core 87/87, corpus byte-identical).
- **S2** switch statement throws but switch expression falls through to `default(T)` — opposite of C#, both introduced here. B3's fix resolves it.
- **S3** parameterless `InvalidOperationException` ("Operation is not valid due to the current state of the object") gives no hint it is compiler-generated. Worse, for a lambda missing a return (#2921, GS0100 never enforced on lambda bodies) base throws a loud `InvalidProgramException` while HEAD throws this opaque one — **masking** #2921. C# uses a dedicated `SwitchExpressionException`. Add a message at minimum.
- **S4** transfer out of `finally` discards an in-flight exception (`f08.gs`: base InvalidProgramException → HEAD `ret=7 trace=tf`, exception swallowed). Deliberate and defensible (matches `Evaluator.Statements.cs:625-648`, ECMA-335 III.1.7.5; C# forbids the construct via CS0157) but ADR-0139's new text never mentions it. Document.
- **S5** `AllPathsReturn` correctly reads the pre-rewrite `DefiniteReturnAnalysisBody`, but `RefKindDefiniteAssignmentAnalyzer` and `RefStructAsyncLivenessAnalyzer` receive the **post-`FinallyExitRewriter`** body. The PR's 12 sibling-drift probes are black-box and do not cover the escaping-finally shape — the only shape that can drift.
- **S6** M09 (`RouteThroughFinally` local-label short-circuit) is live but unpinned: disabling it breaks 0 Core / 0 Compiler tests and leaves the corpus byte-identical. Pin it or delete it.
- **S7** metrics: author's "Pins 28; guards 49" is Core-only and undercounts. Measured against base `3cf2c4f3`: Core **38 pins / 49 guards**, Compiler **31 pins / 15 guards** → **69 pins / 64 guards**. The "TOTAL=123" corpus figure is a third distinct scope — `samples/` holds **148** `.gs`, of which 122 are single-file compilable. And the corpus carries ~zero signal here: across all 148 there are `try` 4, `finally` 6, `break` 1, `continue` 1, `goto` 1, `switch` 15, `enum` 3, and **zero** exhaustive-enum-switch-without-default. `BYTEDIFF=0` is true but is not over-fire evidence — 4 hand-written probes found BLOCKER 1 immediately.

### Claims I tried to falsify and could not
1. **"Analyzer-only fix is unsafe" — TRUE, author vindicated.** Reverting both emitter hunks while keeping the analyzer makes `i2906.gs` compile clean, then `ilverify` reports `Error [ReturnMissing]: … [offset 0x00000022] Return value missing on the stack` and execution gives `InvalidProgramException` exit 134. The method-tail throw **is** necessary. This also served as the known-bad control proving the harness detects this class.
2. `NewChoice()` non-literal invariant intact (`ControlFlowGraph.cs:508-511`); both new `AddTry` call sites use it.
3. **Every `finally` ran exactly once on every path**, compiler and interpreter, across 8 probes — no double-run, no skip, clone count linear in exits × nesting. `f05` shows HEAD also fixes a pre-existing GS0100 **false positive**.
4. Tail-throw over-fire not observed — `alwaysAppendRet = function.Type == Void`; the old `ret` on non-void fall-off was always invalid IL. Net safety win.
5. Test harness clean — dual-pipe `ReadToEndAsync`, `WaitForExit(3000)`, `Kill(entireProcessTree: true)`, real `Assembly.Load` + `GetTypes()` + invoke.
6. Suites real: Core.Tests **7000 passed / 1 skipped**, Compiler.Tests **3749 passed** (1h07m). A first attempt aborted at 2514 with `Test host process crashed`, attributed to machine contention rather than the PR; one clean CI run should confirm.
7. `IsExhaustive` propagation complete at all four rewrite sites, all pinned by M14–M17.

### Mutation table — 25 hunks, **24/25 pinned**
Control Core 87/87 + Compiler 46/46 green before each; every mutation build-green and semantically inert; restored between. Only **M09 unpinned** (see S6). Thin pins (exactly 1 test each): M08, M12, M16, M24, M25.

### Merge order
Rebase onto `d4038f1d` is **conflict-free** (zero file overlap). **PR #2917 will conflict** — both edit `MethodBodyEmitter.EmitMethodBody` at `@@ -234`; #2917 inserts before the `if (alwaysAppendRet || !currentPositionEndsInTerminator)` block, #2926 replaces it. Resolution is mechanical, and **#2917's early return must stay first** so its `leave`-to-`fixedReturnLabel` path never reaches the throw:
```csharp
this.EmitBlock(body);
if (this.EmitFixedReturnEpilogue()) { this.il.OpCode(ILOpCode.Ret); return; }   // 2917
if (alwaysAppendRet) { this.il.OpCode(ILOpCode.Ret); }                          // 2926
else if (!this.currentPositionEndsInTerminator) { /* newobj + throw */ }        // 2926
```
`StatementBinder.Blocks.cs` is edited by both but at disjoint ranges — no conflict.

### NITs
`FinallyExitRewriter.ExitFunneler` (`:72-102`) overrides `RewriteGotoStatement` but not `RewriteConditionalGotoStatement` — no triggering program could be constructed (lowering appears to always emit escaping transfers as unconditional gotos), so theoretical only. Iterator lowering drops `IsExhaustive` (`iter_switch.gs` byte-identical base↔HEAD), so the #2906 fix silently does not apply inside iterators; async **does** keep it and works correctly.

DavidObando:
## Rubber-duck review — round 2 (Opus 5), head `dae7b3ff`

**Verdict: NOT MERGEABLE** — one blocker, cheap remedy (~30 test lines, **zero production change**).

### §1 — B1 deletion is complete and safe. Tried to falsify, could not.
Probed the exhaustive switch as last statement of a non-void function inside `try`, `finally`, loop, nested switch, lambda and iterator — 10 nested shapes, plus the structural argument at `MethodBodyEmitter.cs:385-393`. No shape emits `ReturnMissing` or an unloadable assembly. The two deleted tests pinned **only** the over-firing behaviour. The method-tail throw remains load-bearing.

### §2 — BLOCKER: three shared branches have zero detectors in *either* caller
The extraction itself is **behaviour-preserving** — the shared `Walker` (`ProtectedRegionBranchAnalysis.cs:74-113`) is a verbatim move of the old `ProtectedRegionBranchRewriter.Planner` walker, so PRBR is unchanged by construction, and the overridden `VisitTryStatement` visits exactly the node set `BoundTreeWalker.VisitTryStatement` does (`BoundTreeWalker.cs:467-479`), differing only by `tryStack` push/pop.

But three branches are live, load-bearing and **completely unpinned**. Control first (unmutated, in-worktree): Core 176/176, Compiler 90/90, with filters covering both callers. Every mutation build-green and semantically inert, restored **and rebuilt** between runs.

| ID | branch | core | comp | detector |
|---|---|---|---|---|
| N1 | `HasEscapingBranch` → false | 0 | 16 | FER |
| **N2** | **`ContainsLabel`:49-50 → false** | **0** | **0** | **none** |
| N3 | `ContainsLabel` → true | 0 | 9 | FER |
| N4 | `TryGetLabelRegions` → false | 0 | 7 | PRBR |
| N5 | Walker labels not recorded | 0 | 8 | PRBR |
| N6 | Walker gotos not recorded | 0 | 23 | both |
| **N7** | **Walker cond-goto not recorded** | **0** | **0** | none (= filed #2942) |
| N8 | Walker `tryStack` removed | 0 | 7 | PRBR |
| **N9** | **Walker :111 skip nested `FinallyBlock`** | **0** | **0** | **none** |
| **N10** | **Walker :106-109 skip catch bodies** | **0** | **0** | **none** |

Liveness of the unpinned three, clean-HEAD control first (`md5 GSharp.Core.dll = 8ca0a064…`), rebuild after each restore:

| probe | clean HEAD | N2 | N9 | N10 |
|---|---|---|---|---|
| `x_n2` local `goto` loop in try, finally has `break` | `fin#1 / r=1301` | **`r=101` silent wrong answer** | ok | ok |
| `x_n9` label+goto in **nested finally** | `outerfin / r=5` | **InvalidProgramException** | **InvalidProgramException** | ok |
| `x_n10` label+goto in **nested catch** | `outerfin / r=3` | **InvalidProgramException** | ok | **InvalidProgramException** |
| `z_cond` conditional escape in finally | `fbreak / c=true r=1` | ok | ok | ok |

**Critical:** for N10 on `x_n10`/`y3_innercatch`, **`ilverify` reported `0 errors` while the runtime threw `InvalidProgramException`**. The author's tests are `IlVerifier`-based, so this class of defect is invisible to them even if the shapes existed. `Issue2891TryRegionFlowEmitTests.cs` has only labeled-**loop** cases — no plain `goto` to a label inside a nested catch or finally.

The gap pre-existed in round 1's `LabelCollector`; the extraction did not create it, but it **widened the blast radius** — a future edit made for PRBR now silently breaks FER with no test and no `ilverify` signal. Against the pre-registered criterion ("if only one caller has a detector the extraction is unsafe"), zero detectors is worse.

**Remedy:** land `x_n2`, `x_n9`, `x_n10` as three guard cases asserted by **execution**, not `IlVerifier`. ~30 lines, no production change, and the net −24 lines then genuinely pays. Reverting to duplication is the alternative but is **not** recommended — the extraction is sound and the gap is orthogonal to it.

### §3 — Merge recipe, re-derived and validated by build+test
`git merge-tree --write-tree HEAD origin/main` → exit 0, **clean**. Against the in-flight set: #2908, #2911, #2912, #2929, #2932, #2935 all clean (#2911 is `CONFLICTING` on GitHub versus *main*, not versus this PR). Only **#2917** conflicts, in one file:

```csharp
public void EmitMethodBody(BoundBlockStatement body, bool alwaysAppendRet)
{
    this.EmitBlock(body);

    if (this.EmitFixedReturnEpilogue())      // #2917 — MUST stay first, keep the early return
    {
        this.il.OpCode(ILOpCode.Ret);
        return;
    }

    if (alwaysAppendRet)                      // #2926
    {
        this.il.OpCode(ILOpCode.Ret);
    }
    else if (!this.currentPositionEndsInTerminator)
    {
        const string Message = "Compiler-generated guard reached: non-void function fell through without returning a value.";
        var constructor = typeof(InvalidOperationException).GetConstructor(new[] { typeof(string) });
        this.il.LoadString(this.outer.emitCtx.Metadata.GetOrAddUserString(Message));
        this.il.OpCode(ILOpCode.Newobj);
        this.il.Token(this.outer.memberRefs.GetCtorReference(constructor));
        this.il.OpCode(ILOpCode.Throw);
    }
}
```
Keep #2917's `EmitFixedReturnEpilogue()`/`GetFixedReturnLabel()` verbatim; drop its now-stale `"Preserve the legacy trailing ret…"` comment. **Validated, not asserted:** trial merge in a disposable clone builds 0 warnings / 0 errors, and `Issue2891|Issue2906|Issue2900Fixed` → **60/60 passed**.

And yes — #2917's early return makes the new throw unreachable for `fixed`-containing methods. `fixedReturnLabel` is armed only at `:364`, when a `return` is emitted inside a protected region in a `hasFixedReturns` method, and the epilogue supplies a real `ret`. **That is correct**, but it means those methods keep the old silent default-value return on the fall-through path that #2926's guard exists to make loud. A narrow hole created by the *interaction*, not by either PR alone — whoever lands second should add one case.

### §4 — All author numbers verified true
Core **7003 + 1 skipped**, Compiler **3752**, Interpreter **453**, Release 0 warnings. Four new over-fire cases plus `nullshape` genuinely fail at `bb0ee017` and pass at `dae7b3ff`.

### New finding — the S5 fix repairs a regression this PR itself shipped, with zero coverage (SHOULD-FIX, near-blocking)
Nested try/finally, inner finally `goto` out of a loop, outer finally labeled `break`, `out` assigned after the loop:

| compiler | result |
|---|---|
| base `3cf2c4f3` | clean |
| round 1 `bb0ee017` | **`GS0238: Out parameter 'v' must be definitely assigned…`** ← round 1 shipped this false positive |
| head `dae7b3ff` | clean (S5 repairs it) |
| head with S5 reverted | **`GS0238`** reproduces |

So S5 fixes a false positive that **blocks valid user code**, and **all 7003 Core tests pass with S5 fully reverted**. `Issue2891SiblingAnalyzerNoDriftTests.cs` has 14 cases, but every escaping-finally case is single-level and pins the *positive* direction. Needs one 15th case, `Case("OutNestedEscapingFinallyClean", "", …)`, ~12 lines.

### Follow-up issues — all real, correctly scoped, none disguised deferral
**#2942** exactly matches my N7 (detector records `BoundConditionalGotoStatement` but `ExitFunneler` overrides only `RewriteGotoStatement`); the author's "no current source program found" is independently confirmed by my mutation. **#2941** correctly defers a pre-existing language-design question. **#2940** is real but **now near-vacuous and its text is stale** — post-B1, `grep IsExhaustive src/` shows the only consumer is `ControlFlowGraph.cs:318`; the emitter no longer reads it at all, and iterators never require definite return. Update its rationale so it is not mis-prioritised.

### NITs
Author metrics `69 pins / 64 guards` vs my measured **68 / 71** over 139 — directionally honest, numerically off. `ProtectedRegionBranchAnalysis` exposes `Branch.Regions`/`Branch.Statement`/`labelRegions` that only PRBR uses — dead surface for FER, not a bug. Zero new `Interpreter.Tests`; parity lives only in Compiler.Tests.

### Corrections to my briefing, and to myself
- `origin/main` `60e5b8bf` **does** touch `MethodBodyEmitter.cs` (hunks `@@ -98/-162/-190/-198`) — different region from #2926's `@@ -234`, merge still clean.
- The briefing omitted **`Binder.cs` +35** from the delta description — that is exactly where the S5 finding lives.
- My round-2 "regression" on `x_n10` was **my harness error** (restore without rebuild). Re-verified on clean HEAD: passes. All tables above use rebuild-after-restore with md5-verified binaries.

DavidObando:
## Round 3 review — MERGEABLE

Head reviewed: **`1960b64d`**. Blocker resolved. Two NIT-level items, neither blocking.

Round-3 delta is **tests-only**: `git diff --stat 958fbc06..1960b64d` = 4 test files, `+189/−8`, **zero `src/` change**, and `git diff dae7b3ff..958fbc06 -- src/` is **empty**. The rebase moved production source by exactly nothing, so all round-2 production validation carries bit-for-bit.

### §1 — The three guards are real. Blocker resolved.
Fresh clone, unmutated control first, restore **and rebuild** between every probe. Control: 3 pass, Core 91/91, Compiler 64/64.

| mutation | target branch | result | Core | Comp |
|---|---|---|---|---|
| **N2** `ContainsLabel` → `false` (`ProtectedRegionBranchAnalysis.cs:51-52`) | label-set membership | **3/3 FAIL** | 0 | 3 |
| **N9** skip nested `FinallyBlock` (`:115-118`) | walker descends nested finally | **1 FAIL — exactly `LocalGotoInNestedFinally`** | 0 | 1 |
| **N10** skip catch bodies (`:110-113`) | walker descends catch clauses | **1 FAIL — exactly `LocalGotoInNestedCatch`** | 0 | 1 |
| N7 cond-goto not recorded (`:96-98`) | conditional branch | **0 fail** — still dead (#2942 accurate) | 0 | 0 |

Detection is **precise, not blunt** — N9 and N10 each kill exactly the one case that exercises them.

Assertion is by **execution**: `Issue2891TryRegionFlowEmitTests.cs:646-725` does `IlVerifier.Verify` → `Assembly.Load().GetTypes()` → `Process.Start`, with `Assert.Equal(0, ExitCode)` catching the exit-134 `InvalidProgramException` (N9/N10) and `Assert.Equal(expectedOutput, stdout)` catching N2's silent wrong answer (`r=101` vs `r=1301`).

**Child-process hygiene avoids the known deadlock**: both pipes started as `ReadToEndAsync()` *before* `WaitForExit`, bounded `WaitForExit(10_000)`, `Kill(entireProcessTree: true)` on timeout, temp files cleaned in `finally`.

**`IlVerifier`-only audit — clean.** All three `IlVerifier.Verify` call sites are followed by real execution (`:624`→`:631`→`:637`; `:662`→`:663`→`:677`; `Issue2906…:339`→`:346`→`:352`). The one `ignoredErrorCodes` suppression maps to a single ilverify `-g <code>`, not a blanket pass.

**Bonus:** compiler/interpreter parity confirmed for all three shapes (interpreter returns 1301 / 5 / 3, zero diagnostics) — not tested by the PR.

### §2 — NIT-1: the `fixed` expectation is right, the wording is wrong
Two programs identical except one arm uses `unsafe fixed`:

| program | base `e21677f6` | HEAD, unmatched enum value |
|---|---|---|
| `nofixed.gs` | `error GS0100` | **loud** `InvalidOperationException: Compiler-generated guard reached…` |
| `unfixed.gs` | `error GS0100` | **silent `0`** |

Both were rejected at compile time on base. This PR newly accepts both and gives them **divergent** fall-through semantics — so the divergence is *created here*, not inherited.

The comment at `Issue2906ExhaustiveSwitchReturnEmitTests.cs:315-316` says the epilogue is emitted first *"so an unmatched value returns its default slot"*. The ordering genuinely is intentional (`EmitFixedReturnEpilogue()` must run first — it marks the label every region-crossing `leave` targets), but *"intentionally"* reads as if the silent `0` is a designed semantic. It is a consequence.

Reword to name it a limitation and link a tracking issue; add a line to `docs/adr/0139` beside the exception-discard paragraph. **The emitter should not change** — reordering breaks the `leave` target, and a conditional guard-after-epilogue is real complexity for an undefined-input path.

Suppression is honest: `ilverify` on the `fixed` case yields exactly one error, the already-tracked `ExpectedNumericType` (#2922) which the JIT tolerates; force-JIT of every method succeeds and the program runs `result=0 matched=7 exit=0`.

### §3 — Rebase correct
`MethodBodyEmitter.EmitMethodBody` matches the recipe exactly: `EmitFixedReturnEpilogue()` first with its early return, then `alwaysAppendRet`, then the message-carrying throw; #2917's helpers verbatim; the stale `"Preserve the legacy trailing ret…"` comment dropped; base's single combined `if` correctly split into two arms. `Issue2891|Issue2906|Issue2900Fixed` → **64/64** (was 60/60 on the trial merge; +4 new cases). `merge-tree` vs `origin/main` clean.

### §4 — Metrics reconciled
Re-derived at base `e21677f6`: Core 39 pins / 52 guards, Compiler 33 / 20, Interpreter 0 / 1 → **72 / 73 / 145** vs the author's **71 / 73 / 144**. Guards agree exactly; off by one pin, a single test bucketed differently. Round 2 they were off by 7 — now materially honest.

Full executed suites at `1960b64d`: Interpreter **469**, Core **7004** + 1 skip, Compiler **3809**, Release 0 warnings / 0 errors, zero environmental failures. Arithmetic closes perfectly against the established `e21677f6` baselines: `6913+91=7004`, `3756+53=3809`, `468+1=469`. `git diff --stat e21677f6..HEAD -- test/` = 6 files, **+2408, zero deletions** — no pre-existing test modified or removed.

**S5 is now pinned**: reverting all 7 `lowered.PreEmitAnalysisBody ?? lowered` sites yields exactly one red, `SharedCfgDiagnosticsRemainUnchanged(name: "OutNestedEscapingFinallyClean")`. In round 2 that revert passed all 7003 Core tests.

**#2940 reworded and now verbatim accurate** — `grep -rn IsExhaustive src/` confirms the sole production consumer is `ControlFlowGraph.cs:318`.

### §5 — #2908 orthogonality confirmed
`comm -12` of the two file lists → **EMPTY**. Zero file overlap. Descriptor ordering is a non-question: this PR adds no diagnostic and touches no descriptor file or docs page.

### §6 — Corpus carries zero signal, and worse than predicted
```
148 .gs files;  goto: 0;  try{: 3;  finally{: 4;  catch(: 1;  switch: 14
```
Not "0 of 148 have `goto` into a nested handler" — **0 of 148 contain `goto` at all.** Corpus byte-identity is worthless as over-fire evidence here. The author's round-3 evidence is correctly hand-written execution guards.

### §7 — NIT-2 withdrawn
My round-2 "dead surface" call was wrong. `ProtectedRegionBranchRewriter.cs:93,95,101,103,131` genuinely consumes `Branches`, `TryGetLabelRegions`, `branch.Regions`, `branch.Statement`. It is an **asymmetric** API, not unused code. Consciously kept is right.

### Corrections to the briefing
- **#2917 was already merged** (`eaf59056`, predating this PR's merge-base) — never a live cross-PR conflict, just a rebase.
- Round-3 delta is tests-only; round-2 production validation stands unmodified.
- `693+/84−` across 13 files is src-only; the 14th file is `docs/adr/0139` (+12/−6). Whole PR is 20 files, **+3113/−90**.
- The three new cases are **pins**, not guards — they fail at base.
- Corpus `goto` count is **0**, not merely 0 for the nested shape.

### Tried to falsify, could not
Precise mutation detection; harness cannot deadlock and is kill-bounded; no `IlVerifier`-only assertion survives; suppression hides nothing; rebase changed zero production source; #2908 orthogonality; **finally executes exactly once per exit on all 8 exit shapes** (traces byte-identical to round 1); compiler/interpreter parity on all three new shapes; every reported test count verified by full executed runs.

Issue #2889, opened 2026-07-30
gsc: imported delegate with a nested source generic argument emits an erased lambda thunk signature


DavidObando:
Found while reviewing #2849 / PR #2884. This is separate from #2849 because the first failure occurs on a non-nullable delegate invocation with no smart-cast narrowing.

## Summary

An imported generic delegate whose type argument is itself a generic type closed over a source-declared type loses the nested source type in the emitted lambda thunk signature. `System.Action[List[Src]]` is reified at the delegate call site, but the thunk accepts erased `List[object]`, producing invalid IL.

Verified on current `main`, PR #2884 before its review follow-up, and PR #2884 HEAD. The failure is identical, so this is **not a regression** from #2884.

## Repro

```gsharp
package DForms
import System
import System.Collections.Generic

class Src { prop N int32 -> 2 }

func Main() {
    let w System.Action[List[Src]] = (items List[Src]) ->
        System.Console.WriteLine(items.Count)
    w(List[Src]())
}
```

## Observed

Compilation succeeds, but `ilverify` reports a stack mismatch beginning at the non-nullable `w(List[Src]())` call:

```text
[found ref 'List`1<object>']
[expected ref 'List`1<DForms.Src>']
```

The erased `List<object>` appears in the generated lambda thunk signature.

## Expected

Nested symbolic generic arguments must remain reified through lambda thunk generation. The thunk and delegate `Invoke` signatures should both use `List<Src>`, and the emitted assembly should pass IL verification and run normally.

## Scope / root-cause lead

This is not nullable receiver handling and not smart-cast narrowing. The direct non-nullable row already fails. Investigate symbolic type preservation when closure/delegate emission builds thunk signatures for imported generic types containing source-declared generic arguments.

## Suggested regression coverage

- non-nullable `Action[List[Src]]` lambda conversion, IL verification, and runtime output
- nullable narrowed invocation as a control proving the same nested type remains reified
- named and imported delegate controls
- source class and source struct nested arguments

Issue #2391, opened 2026-07-16
gsc: symbolic interface call return types lose generic substitutions


DavidObando:
Discovered while expanding #2380 regression coverage.

Calls through symbolic/source interface shapes can preserve substituted parameter types while losing the corresponding return-type substitution, producing an erased or incorrect bound result type.

Add same-compilation generic interface call-site repros and make parameter and return substitution use the same closed symbolic type mapping.

PR #2929, opened 2026-07-31
Fix inline lambdas through erased generic receivers

### LINKED ISSUE #2918: gsc: inline lambda argument fails through erased imported-generic receiver
## Repro

```gsharp
package InlineErasedReceiver
import System

class Src {
    let N int32
    init(n int32) { N = n }
}

func Main() {
    callbacks.Add((item Src) -> Console.WriteLine(item.N))
    callbacks[0](Src(1))
}
```

## Actual

Both `main` and PR #2913 reject the inline lambda:

```
error GS0159: Cannot find function Add.
error GS0155: Cannot convert type 'object' to 'Src'.
```

Hoisting the lambda to a typed local before `Add` works.

PR #2913 exposes the same pre-existing limitation for the previously mis-erased nested shape `List[System.Action[List[Src]]]`: `main` compiled it but emitted invalid IL; the corrected symbolic projection now reaches this misleading diagnostic instead.

## Expected

Inline and hoisted lambda arguments should bind identically through imported generic receivers, emit verifiable IL, and run.

## Root-cause lead

The receiver's symbolic element type is erased during imported-generic member lookup/call binding before the inline lambda receives its target delegate type. Fix the shared inline-argument-through-erased-receiver path; do not special-case `List.Add`.

### LINKED ISSUE #2919: Binder: consolidate imported generic argument projection
## Problem

`Binder` contains three near-verbatim imported generic argument projection blocks:

- `src/Core/CodeAnalysis/Binding/Binder.cs:3027`
- `src/Core/CodeAnalysis/Binding/Binder.cs:3855`
- `src/Core/CodeAnalysis/Binding/Binder.cs:4098`

PR #2913 fixed the last divergent block by matching the established `RequiresSymbolicProjection` pattern. The duplication remains and has already allowed one copy to drift.

The `:3027` path additionally preserves the #2391 source-enum surrogate with:

```csharp
var erasedArgument = ta is EnumSymbol ? typeof(int) : typeof(object);
```

The other two paths use `object`. Review probes could not make that difference observable, so this is latent divergence, not a demonstrated correctness bug.

## Proposed cleanup

Extract one shared helper such as `ProjectGenericArgument(TypeSymbol ta, ref bool hasSymbolicArg)` and route all three construction sites through it. Preserve existing diagnostics, reference-context mapping, null fallbacks, and the enum surrogate semantics. Add focused parity/mutation coverage for all three callers before deleting the duplicated blocks.

Keep this refactor separate from #2913 to avoid expanding that correctness fix's blast radius.

DavidObando:
## Summary

- Consolidate the three imported-generic argument projection blocks into `ProjectGenericArgument` while preserving each caller's established CLR surrogate.
- Preserve inline lambda target types when an imported generic receiver or constructor erases its closed delegate argument to `object`.
- Remove unneeded symbolic-target plumbing and the `carries` filter that regressed constructed-generic constructor overload resolution.
- Add projection parity tests, IL/load/execute coverage, a constructor-overload regression, and a C# consumer of G#-emitted delegate surfaces.

Fixes #2918
Fixes #2919

## Root cause

Imported receivers retain symbolic type arguments, but deferred inline-lambda binding inspected an open method parameter such as `List<T>.Add(T)` before substituting receiver `T = Action<Src>`. Open `T` is not itself a delegate, so that candidate was discarded and binding fell back to the erased `object` shape. Hoisted lambdas avoided the path because they were already typed.

The fix maps the complete open parameter through the receiver's symbolic arguments first, then reuses `TryGetLambdaTargetFunctionTypeFromSymbol` for native function, delegate, and `Expression<TDelegate>` targets. Constructors use the same complete-parameter mapping.

## Review-round correction

The original fix added a `carries` gate that rejected fully closed delegate candidates. Constructor target resolution must see every delegate-shaped candidate so disagreeing overloads cancel out. Hiding the closed candidate forced the symbolic target onto a lambda that selected another overload and produced GS0155/GS0159.

The gate is gone. `ImportedGenericConstructorOverloads_DisagreeWithoutForcingSymbolicTarget_VerifyLoadAndRun` covers `Holder<T>(Action<T>)` versus `Holder<T>(Action<int>)`: the gate mutant fails compilation; the restored code IL-verifies, loads, executes, and prints `symbolic` / `closed`.

The unused five-argument helper overload, method-level argument plumbing, dead `AllMethodTypeParametersResolved` check, unpinned generic-slot arity check, and redundant successful-result null guard were deleted. The issue #903 comment now sits beside the unresolved-slot check it explains. The arity fallback's pre-existing runtime failure is tracked in #2937.

## Named-delegate safety guard

Deferred erased-slot recovery carries a structural function shape, not nominal imported delegate identity. Reconstructing that shape is safe for canonical `System.Action` / `System.Func`, but previously turned `Predicate<Src>` and `Comparison<Src>` into `Func` delegates and emitted wrong code. The new guard conservatively rejects imported noncanonical delegate targets in both deferred method slots and imported generic constructor slots.

Four pins cover the boundary:

- `ImportedPredicateThroughErasedListSlot_IsRejected` (`q3_predicate`)
- `ImportedComparisonThroughErasedListSlot_IsRejected` (`q3c_comparison`)
- `ImportedPredicateThroughErasedConstructorSlot_IsRejected` (`q7_boxpred`)
- `HoistedImportedPredicateThroughGenericConstructor_RetainsDelegateIdentity` (`q11`), which loads and runs as `System.Predicate<Src>` / `True`

Before the guard, the three inline rejection pins compile successfully and therefore fail their assertions; the hoisted twin passes. Carrying exact named-delegate identity through `slotTargets` is the larger positive fix tracked in #2948.

Branch rebased cleanly onto `8dddbbf2`, which contains requested `e21677f6` / PR #2935 baseline plus later delegate-adjacent fixes including #2908 and #2932. PR #2908's GS0503 nullable-delegate receiver check required only an explicit `!!` in the q11 test invocation; production code was unchanged. Evaluate-mode binding also has a bind-only Interpreter.Tests pin for direct and nested `List<Action<...>>` receivers; it avoids the evaluator closure-to-`System.Action` limitation while exercising the changed binder path.

## Round-three constructor guard scoping

The temporary imported named-delegate guard now applies only when the open constructor parameter is itself a generic parameter — the erased-slot case. Fully typed constructor parameters such as `PredHolder<T>(Predicate<T>)`, `Comparison<T>`, `EventHandler<T>`, `Converter<T,U>`, and user-library delegates retain their established behavior.

`ImportedConstructedPredicateConstructor_TargetsInlineLambda_LoadsAndRuns` pins the fully typed `Predicate<T>` shape with real assembly load and execution. Restoring the unscoped guard makes this test fail with GS0155/GS0159/GS0158; restoring the two-line scope makes all 18 targeted compiler tests pass. The three erased-slot rejection pins remain unchanged.

`TryBindSymbolicReceiverCall` intentionally uses only the mapped Invoke shape because its explicit `symbolicDelegateParam` conversion preserves nominal delegate identity; the source now records that distinction. #2948 documents that the temporary canonical `Action`/`Func` test is assembly-blind because it compares CLR `FullName`. Its statement that this PR rejects only erased inline cases is accurate after this correction.

The 148-sample corpus contains zero imported named-delegate erased-slot cases and is not claimed as coverage for this guard.

## Projection consolidation (#2919)

Commit `1d660244` replaces all three near-identical blocks with one helper. Callers still choose their erased surrogate:

- unqualified `Action[...]`: enum -> `int`, otherwise `object` (#2391 behavior)
- qualified CLR generic construction: `object`
- nested CLR type segments: `object`

The helper retains reference-context mapping, null fallbacks, `ContainsSameCompilationUserType`, both `ClrType == null` defenses, the `ContainsTypeParameter` ternary, and the `ResolveClrTypeForGenericArg` fallback.

### Corpus caveat and proof

The 148-sample corpus is weak signal for #2919, not proof of all three replaced sites. Instrumentation at `1d660244` found:

```
54 site=A sym=False
 7 site=A sym=True   (2 files only)
 0 site=B            (never executed)
 3 site=C sym=False
 0 site=C sym=True
 0 enum=True anywhere
```

The corpus still establishes broad no-op evidence at `1d660244`: 148 attempted, 139 emitted, identical exit codes, `BYTEDIFF=0`, and path-normalized `DIAGDRIFT=0`. The real three-site proof is `Issue2919ImportedGenericProjectionTests`: its source hits sites A/B/C symbolically, with and without enums. Independent R1/R2/R3/R7 mutants all fail that test.

## Removed known-limitation pin

`Issue2889LambdaThunkNestedGenericTests.InlineLambdaThroughNestedErasedReceiver_ReportsCurrentDiagnostics` and its single-use `CompileExpectingFailure` helper were removed intentionally. That test pinned GS0159/GS0155 until #2918 fixed the exact program, so it must fail after the fix. Coverage now lives in `Issue2918InlineLambdaErasedReceiverTests.ImportedGenericReceiverMethods_TargetInlineLambdas_VerifyLoadAndRun`, which upgrades the nested `List<Action<List<Src>>>` case to compile + IL verification + real assembly load/`GetTypes()` + subprocess execution + exact stdout. The separate #2889 C# consumer test remains intact.

## Behavior coverage

Newly corrected shapes include `List<Action<Src>>`, nested generic delegate arguments, `Func` returns, dictionary/later argument slots, inferred parameters, source structs/enums/interfaces/nested generics, unqualified delegates, imported generic constructors, and G#-emitted public delegate surfaces consumed statically from C#.

Every acceptance assembly is IL-verified, loaded through a collectible `AssemblyLoadContext`, queried with `GetTypes()`, and executed with exact output. The harness also corrupts a real method body and confirms `IlVerifier` rejects it.

Pins and guards are reported separately:

- **Pins:** 8 independently confirmed base-fail/final-pass shapes (`p01`, `p02`, `p04`, `p06`, `p08`, `p09d`, `p13`, `p14`), plus the new static C# surface twin through the same H2 branch.
- **Guards:** 8 independently confirmed base-pass/final-pass probes, the 12 shapes in `ExistingDelegateTargetingControls_RemainValid`, and the constructor-overload regression (base/refactor/final pass; gate mutant fails).

### Production-branch mutations

Unmutated controls passed before each build-green mutation, and restored code was rebuilt before the next mutation.

| Branch | Mutation | Result |
|---|---|---|
| R1 | site-A enum surrogate -> `object` | detected |
| R2 | add enum surrogate at site B | detected |
| R3 | add enum surrogate at site C | detected |
| R7 | drop symbolic retention flag | detected |
| H1 | reinsert `carries` gate | detected by constructor-overload regression |
| H2 | disable declaring-type generic-parameter recovery | detected by receiver and C# surface tests |
| H3 | dead method-parameter-resolution gate | deleted |
| H4 | generic-slot arity check | deleted; underlying bug tracked in #2937 |
| H5 | five-argument helper overload | deleted |
| H6 | redundant candidate null guard | deleted; callee guarantees non-null on success |
| H7 | method-type-argument plumbing into H5 | deleted |

R4/R5/R6/R8/R9 remain pre-existing defensive terms carried through the refactor; their mutations survive, so they are not claimed as new pins.

## Scope

Source-declared generic receivers still lose callable return shapes (`box.Get(0)(Src(5))`, GS0159) on both base and this branch. #2918 is limited to imported receivers as filed; the source-generic path is tracked separately in #2936.

## Rebase and final validation

Branch is rebased cleanly onto current base `8dddbbf2`. Counts below are executed tests measured on that same base/head line; they are not `--list-tests` declared-method counts.

| Suite | Base `8dddbbf2` | Head `3b7a05a9` | Delta |
|---|---:|---:|---:|
| `Compiler.Tests` | 3893 | 3903 | +10 |
| `Core.Tests` | 7028 passed, 1 skipped | 7029 passed, 1 skipped | +1 passed |
| `Interpreter.Tests` | 489 | 490 | +1 |
| `LanguageServer.Tests` | 452 | 452 | 0 |

Final head validation:

- `dotnet build GSharp.sln --configuration Release --no-restore -graph -warnaserror -m:1`: **0 warnings, 0 errors**
- Delegate/tuple compiler targets (Issue2918 + Issue2889 + Issue2928 + Issue2924): **25 passed**
- PR #2935 `Issue2928TupleFunctionParityTests`: **3 passed**
- PR #2935 `Issue2928CallableMaterializationTests`: **15 passed**
- Newly merged #2932 `Issue2924TupleElementDelegateCallTests`: **7 Compiler + 21 Core + 17 Interpreter passed**
- Issue #2918 targeted compiler class: **11 passed**
- Issue #2918 bind-only interpreter class: **1 passed**
- Freshness-guarded round-three mutation rebuilt `GSharp.Core.dll` (mtime changed) and the `PredHolder<T>` pin still detected the unscoped guard; restored code was freshly rebuilt and passed.
- No local e2e run and no SDK installation into the shared NuGet cache.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/Binder.cs
+++ b/src/Core/CodeAnalysis/Binding/Binder.cs
@@ -3069,45 +3069,10 @@ public sealed class Binder
                     return null;
                 }
 
-                // #313: an in-scope generic type parameter used as a type
-                // argument (e.g. `List[T]` inside `func First[T](...)`) is a
-                // valid type in any position. Under the type-erased generic
-                // model (ADR-0004; type parameters encode as System.Object at
-                // emit) the type argument projects onto `object` for the closed
-                // CLR shape so member / index / conversion resolution keeps
-                // working, while the symbolic `[T]` is preserved on the result
-                // for inference, substitution, and erased emit.
-                if (TypeSymbol.RequiresSymbolicProjection(ta))
-                {
-                    hasSymbolicArg = true;
-
-                    // Issue #2391: source enums use Int32 as their established
-                    // CLR ride-through. Preserve that surrogate when closing
-                    // an imported generic receiver as well; using object for a
-                    // struct-constrained interface leaves Nullable<T> member
-                    // signatures unconstructable and hides parameterized
-                    // methods from overload resolution.
-                    if (TypeSymbol.ContainsTypeParameter(ta) || TypeSymbol.ContainsSameCompilationUserType(ta))
-                    {
-                        var erasedArgument = ta is EnumSymbol ? typeof(int) : typeof(object);
-                        clrArgs[i] = scope.References.MapClrTypeToReferences(erasedArgument);
-                    }
-                    else
-                    {
-                        clrArgs[i] = ResolveClrTypeForGenericArg(ta)
-                            ?? scope.References.MapClrTypeToReferences(ta.ClrType);
-                    }
-
-                    continue;
-                }
-
-                // Project host CLR type arguments onto the resolver's reference
-                // set so they share clrOpenType's load context (its
-                // MetadataLoadContext when references are supplied via /r:),
-                // which MakeGenericType requires.
-                // Issue #530: use ResolveClrTypeForGenericArg so that
-                // `int32?` resolves to `Nullable<int>` (not bare `int`).
-                clrArgs[i] = ResolveClrTypeForGenericArg(ta) ?? scope.References.MapClrTypeToReferences(ta.ClrType);
+                // Issue #2391: this caller alone retains the established Int32
+                // ride-through for a top-level source enum.
+                var erasedArgument = ta is EnumSymbol ? typeof(int) : typeof(object);
+                clrArgs[i] = ProjectGenericArgument(ta, erasedArgument, ref hasSymbolicArg);
             }
 
             try
@@ -3903,21 +3868,7 @@ public sealed class Binder
                 return null;
             }
 
-            // #313 / #671: keep any symbolic projection alongside the erased
-            // closed CLR shape, including nested generic/array/nullable forms.
-            if (TypeSymbol.RequiresSymbolicProjection(ta) || ta.ClrType == null)
-            {
-                hasSymbolicArg = true;
-                clrArgs[i] = TypeSymbol.ContainsTypeParameter(ta)
-                    || TypeSymbol.ContainsSameCompilationUserType(ta)
-                    || ta.ClrType == null
-                        ? scope.References.MapClrTypeToReferences(typeof(object))
-                        : ResolveClrTypeForGenericArg(ta)
-                            ?? scope.References.MapClrTypeToReferences(ta.ClrType);
-                continue;
-            }
-
-            clrArgs[i] = ResolveClrTypeForGenericArg(ta) ?? scope.References.MapClrTypeToReferences(ta.ClrType);
+            clrArgs[i] = ProjectGenericArgument(ta, typeof(object), ref hasSymbolicArg);
         }
 
         try
@@ -4146,21 +4097,7 @@ public sealed class Binder
                 return null;
             }
 
-            // #313 / #671: in-scope type parameters and user types without a
-            // ClrType project onto System.Object under the type-erased model.
-            if (TypeSymbol.RequiresSymbolicProjection(ta) || ta.ClrType == null)
-            {
-                hasSymbolicArg = true;
-                clrArgs[i] = TypeSymbol.ContainsTypeParameter(ta)
-                    || TypeSymbol.ContainsSameCompilationUserType(ta)
-                    || ta.ClrType == null
-                        ? scope.References.MapClrTypeToReferences(typeof(object))
-                        : ResolveClrTypeForGenericArg(ta)
-                            ?? scope.References.MapClrTypeToReferences(ta.ClrType);
-                continue;
-            }
-
-            clrArgs[i] = ResolveClrTypeForGenericArg(ta) ?? scope.References.MapClrTypeToReferences(ta.ClrType);
+            clrArgs[i] = ProjectGenericArgument(ta, typeof(object), ref hasSymbolicArg);
         }
 
         try
@@ -4178,6 +4115,40 @@ public sealed class Binder
             Diagnostics.ReportTypeNotGeneric(syntax.Identifier.Location, syntax.DottedName);
             return null;
         }
+    }
+
+    /// <summary>
+    /// Projects a symbolic generic argument onto a closed CLR type while
+    /// retaining any type information that the CLR shape cannot represent.
+    /// </summary>
+    /// <param name="type">The symbolic generic argument.</param>
+    /// <param name="erasedArgument">
+    /// The caller-specific CLR surrogate for a same-compilation user type.
+    /// </param>
+    /// <param name="hasSymbolicArgument">
+    /// Set when the symbolic argument must be retained beside the CLR shape.
+    /// </param>
+    /// <returns>The reference-context CLR argument used to close the generic.</returns>
+    private Type ProjectGenericArgument(
+        TypeSymbol type,
+        Type erasedArgument,
+        ref bool hasSymbolicArgument)
+    {
+        // #313 / #671: preserve symbolic type parameters, user types, and
+        // nested generic/array/nullable shapes beside their erased CLR form.
+        if (TypeSymbol.RequiresSymbolicProjection(type) || type.ClrType == null)
+        {
+            hasSymbolicArgument = true;
+            return TypeSymbol.ContainsTypeParameter(type)
+                || TypeSymbol.ContainsSameCompilationUserType(type)
+                || type.ClrType == null
+                    ? scope.References.MapClrTypeToReferences(erasedArgument)
+                    : ResolveClrTypeForGenericArg(type)
+                        ?? scope.References.MapClrTypeToReferences(type.ClrType);
+        }
+
+        return ResolveClrTypeForGenericArg(type)
+            ?? scope.References.MapClrTypeToReferences(type.ClrType);
     }
 
     /// <summary>
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
@@ -782,21 +782,6 @@ internal sealed partial class ExpressionBinder
         Type openDefinition,
         ImmutableArray<TypeSymbol> symbolicArgs,
         out FunctionTypeSymbol target)
-        => TryBuildSymbolicDelegateTarget(
-            openParameterType,
-            openDefinition,
-            symbolicArgs,
-            openMethodDefinition: null,
-            methodTypeArguments: default,
-            out target);
-
-    private static bool TryBuildSymbolicDelegateTarget(
-        Type openParameterType,
-        Type openDefinition,
-        ImmutableArray<TypeSymbol> symbolicArgs,
-        MethodInfo openMethodDefinition,
-        ImmutableArray<TypeSymbol> methodTypeArguments,
-        out FunctionTypeSymbol target)
     {
         target = null;
         if (openParameterType == null)
@@ -810,18 +795,8 @@ internal sealed partial class ExpressionBinder
         var mapped = MemberLookup.MapOpenClrTypeToSymbolic(
             openParameterType,
             openDefinition,
-            symbolicArgs,
-            openMethodDefinition,
-            methodTypeArguments);
-        if (!MemberLookup.TryGetLambdaTargetFunctionTypeFromSymbol(mapped, out var candidate)
-            || candidate == null)
-        {
-            return false;
-        }
-
-        var carries = TypeSymbol.RequiresSymbolicProjection(candidate.ReturnType)
-            || candidate.ParameterTypes.Any(TypeSymbol.RequiresSymbolicProjection);
-        if (!carries)
+            symbolicArgs);
+        if (!MemberLookup.TryGetLambdaTargetFunctionTypeFromSymbol(mapped, out var candidate))
         {
             return false;
         }
@@ -1488,15 +1463,11 @@ internal sealed partial class ExpressionBinder
                 FunctionTypeSymbol functionType;
                 if (openParameterType.IsGenericParameter)
                 {
-                    if (!AllMethodTypeParametersResolved(openParameterType, openMethod, inferred)
-                        || !TryBuildSymbolicDelegateTarget(
+                    if (!TryBuildSymbolicDelegateTarget(
                             openParameterType,
                             receiverOpenDefinition,
                             receiverTypeArguments,
-                            openMethod,
-                            methodTypeArgs,
-                            out functionType)
-                        || functionType.ParameterTypes.Length != arityByIndex[idx])
+                            out functionType))
                     {
                         candidateUsable = false;
                         break;
@@ -1529,6 +1500,17 @@ internal sealed partial class ExpressionBinder
                         break;
                     }
 
+                    // Issue #903: only trust this candidate's delegate parameter
+                    // shape when every method type parameter reachable from it was
+                    // actually resolved by unifying the receiver/arguments. When a
+                    // candidate's "this" parameter does not match the receiver
+                    // shape (e.g. a `Single` overload over `IQueryable<T>` against a
+                    // `List` receiver), the relevant slot stays unresolved and
+                    // MapOpenClrTypeToSymbolic would otherwise surface a bogus open
+                    // parameter (an ImportedTypeSymbol named "T") that neither
+                    // ContainsTypeParameter nor ContainsSameCompilationUserType
+                    // flags — which would then disagree with the correct candidate
+                    // and abort the whole inference. Skip such candidates instead.
                     var unresolvedSlot = false;
                     foreach (var invokeParameter in invokeParameters)
                     {
@@ -1582,17 +1564,6 @@ internal sealed partial class ExpressionBinder
                     functionType = FunctionTypeSymbol.Get(parameterTypes.ToImmutable(), mappedReturnType);
                 }
 
-                // Issue #903: only trust this candidate's delegate parameter
-                // shape when every method type parameter reachable from it was
-                // actually resolved by unifying the receiver/arguments. When a
-                // candidate's "this" parameter does not match the receiver
-                // shape (e.g. a `Single` overload over `IQueryable<T>` against a
-                // `List` receiver), the relevant slot stays unresolved and
-                // MapOpenClrTypeToSymbolic would otherwise surface a bogus open
-                // parameter (an ImportedTypeSymbol named "T") that neither
-                // ContainsTypeParameter nor ContainsSameCompilationUserType
-                // flags — which would then disagree with the correct candidate
-                // and abort the whole inference. Skip such candidates instead.
                 foreach (var parameterType in functionType.ParameterTypes)
                 {
                     if (TypeSymbol.RequiresSymbolicProjection(parameterType))
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
@@ -666,6 +666,8 @@ internal sealed partial class ExpressionBinder
             var openParamType = openParameters[i].ParameterType;
             var symbolicParamType = MemberLookup.MapOpenClrTypeToSymbolic(openParamType, openDef, symbolicArgs);
 
+            // Nominal identity remains on the explicit symbolicDelegateParam
+            // conversion below; only the mapped Invoke shape is needed here.
             if (LambdaBinder.TryGetFunctionLiteral(argument, out var functionLiteral)
                 && symbolicParamType is ImportedTypeSymbol symbolicDelegateParam
                 && symbolicDelegateParam.OpenDefinition != null
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
@@ -782,6 +782,21 @@ internal sealed partial class ExpressionBinder
         Type openDefinition,
         ImmutableArray<TypeSymbol> symbolicArgs,
         out FunctionTypeSymbol target)
+        => TryBuildSymbolicDelegateTarget(
+            openParameterType,
+            openDefinition,
+            symbolicArgs,
+            openMethodDefinition: null,
+            methodTypeArguments: default,
+            out target);
+
+    private static bool TryBuildSymbolicDelegateTarget(
+        Type openParameterType,
+        Type openDefinition,
+        ImmutableArray<TypeSymbol> symbolicArgs,
+        MethodInfo openMethodDefinition,
+        ImmutableArray<TypeSymbol> methodTypeArguments,
+        out FunctionTypeSymbol target)
     {
         target = null;
         if (openParameterType == null)
@@ -789,32 +804,29 @@ internal sealed partial class ExpressionBinder
             return false;
         }
 
-        // Issue #2365: an `Expression<TDelegate>` parameter is not itself a
-        // delegate (it exposes no `Invoke`), so unwrap it to the wrapped open
-        // delegate type first, mirroring the identical unwrap in the
-        // method-parameter sibling <see cref="TryBuildSymbolicDelegateTargetForMethodParam"/>.
-        if (MemberLookup.TryGetExpressionTreeDelegateType(openParameterType, out var unwrappedParameterType))
-        {
-            openParameterType = unwrappedParameterType;
-        }
-
-        var invoke = openParameterType?.GetMethodSafe("Invoke");
-        if (invoke == null)
+        // Issue #2918: substitute the complete parameter before asking whether
+        // it is a lambda target. The open parameter may itself be `T`, while
+        // the receiver or method closes `T` to `Action[Src]`.
+        var mapped = MemberLookup.MapOpenClrTypeToSymbolic(
+            openParameterType,
+            openDefinition,
+            symbolicArgs,
+            openMethodDefinition,
+            methodTypeArguments);
+        if (!MemberLookup.TryGetLambdaTargetFunctionTypeFromSymbol(mapped, out var candidate)
+            || candidate == null)
         {
             return false;
         }
 
-        var invokeParameters = invoke.GetParameters();
-        var parameterTypes = ImmutableArray.CreateBuilder<TypeSymbol>(invokeParameters.Length);
-        foreach (var parameter in invokeParameters)
+        var carries = TypeSymbol.RequiresSymbolicProjection(candidate.ReturnType)
+            || candidate.ParameterTypes.Any(TypeSymbol.RequiresSymbolicProjection);
+        if (!carries)
         {
-            parameterTypes.Add(MemberLookup.MapOpenClrParameterTypeToSymbolic(parameter.ParameterType, openDefinition, symbolicArgs));
+            return false;
         }
 
-        var returnType = invoke.ReturnType.IsSameAs(typeof(void))
-            ? TypeSymbol.Void
-            : MemberLookup.MapOpenClrTypeToSymbolic(invoke.ReturnType, openDefinition, symbolicArgs);
-        target = FunctionTypeSymbol.Get(parameterTypes.ToImmutable(), returnType);
+        target = candidate;
         return true;
     }
 
@@ -1472,29 +1484,102 @@ internal sealed partial class ExpressionBinder
                     break;
                 }
 
-                MethodInfo invoke;
-                ParameterInfo[] invokeParameters;
-                try
+                var openParameterType = openParameters[paramIndex].ParameterType;
+                FunctionTypeSymbol functionType;
+                if (openParameterType.IsGenericParameter)
                 {
-                    var delegateType = openParameters[paramIndex].ParameterType;
-                    if (MemberLookup.TryGetExpressionTreeDelegateType(delegateType, out var unwrappedDelegateType))
+                    if (!AllMethodTypeParametersResolved(openParameterType, openMethod, inferred)
+                        || !TryBuildSymbolicDelegateTarget(
+                            openParameterType,
+                            receiverOpenDefinition,
+                            receiverTypeArguments,
+                            openMethod,
+                            methodTypeArgs,
+                            out functionType)
+                        || functionType.ParameterTypes.Length != arityByIndex[idx])
                     {
-                        delegateType = unwrappedDelegateType;
+                        candidateUsable = false;
+                        break;
+                    }
+                }
+                else
+                {
+                    MethodInfo invoke;
+                    ParameterInfo[] invokeParameters;
+                    try
+                    {
+                        var delegateType = openParameterType;
+                        if (MemberLookup.TryGetExpressionTreeDelegateType(delegateType, out var unwrappedDelegateType))
+                        {
+                            delegateType = unwrappedDelegateType;
+                        }
+
+                        invoke = delegateType?.GetMethodSafe("Invoke");
+                        invokeParameters = invoke?.GetParameters();
+                    }
+                    catch (Exception ex) when (ClrTypeUtilities.IsMetadataLoadFailure(ex))
+                    {
+                        candidateUsable = false;
+                        break;
                     }
 
-                    invoke = delegateType?.GetMethodSafe("Invoke");
-                    invokeParameters = invoke?.GetParameters();
-                }
-                catch (Exception ex) when (ClrTypeUtilities.IsMetadataLoadFailure(ex))
-                {
-                    candidateUsable = false;
-                    break;
-                }
+                    if (invoke == null || invokeParameters == null || invokeParameters.Length != arityByIndex[idx])
+                    {
+                        candidateUsable = false;
+                        break;
+                    }
 
-                if (invoke == null || invokeParameters == null || invokeParameters.Length != arityByIndex[idx])
-                {
-                    candidateUsable = false;
-                    break;
+                    var unresolvedSlot = false;
+                    foreach (var invokeParameter in invokeParameters)
+                    {
+                        if (!AllMethodTypeParametersResolved(invokeParameter.ParameterType, openMethod, inferred))
+                        {
+                            unresolvedSlot = true;
+                            break;
+                        }
+                    }
+
+                    if (unresolvedSlot)
+                    {
+                        candidateUsable = false;
+                        break;
+                    }
+
+                    var parameterTypes = ImmutableArray.CreateBuilder<TypeSymbol>(invokeParameters.Length);
+                    var slotUsable = true;
+                    foreach (var invokeParameter in invokeParameters)
+                    {
+                        var mapped = MemberLookup.MapOpenClrParameterTypeToSymbolic(
+                            invokeParameter.ParameterType,
+                            receiverOpenDefinition,
+                            receiverTypeArguments,
+                            openMethod,
+                            methodTypeArgs);
+                        if (mapped == null || mapped == TypeSymbol.Error)
+                        {
+                            slotUsable = false;
+                            break;
+                        }
+
+                        parameterTypes.Add(mapped);
+                    }
+
+                    if (!slotUsable)
+                    {
+                        candidateUsable = false;
+                        break;
+                    }
+
+                    var returnClrType = invoke.ReturnType;
+                    var mappedReturnType = returnClrType != null && returnClrType.IsSameAs(typeof(void))
+                        ? TypeSymbol.Void
+                        : MemberLookup.MapOpenClrTypeToSymbolic(
+                            returnClrType,
+                            receiverOpenDefinition,
+                            receiverTypeArguments,
+                            openMethod,
+                            methodTypeArgs);
+                    functionType = FunctionTypeSymbol.Get(parameterTypes.ToImmutable(), mappedReturnType);
                 }
 
                 // Issue #903: only trust this candidate's delegate parameter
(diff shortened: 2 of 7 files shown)


DavidObando:
## Review — PR #2929 (`99019f64`), issues #2918 + #2919

**Verdict: NOT MERGEABLE** — one reproducible base→HEAD regression, caused by code that is not needed for #2918 and whose removal shrinks the diff.

Base `d4038f1d` = `origin/main`, 2 ahead / 0 behind. Release build in review worktree: 0 warnings / 0 errors.

### BLOCKER 1 — `carries` gate breaks constructed-generic ctor overload resolution

`src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs:822-828`

```csharp
var carries = TypeSymbol.RequiresSymbolicProjection(candidate.ReturnType)
    || candidate.ParameterTypes.Any(TypeSymbol.RequiresSymbolicProjection);
if (!carries) { return false; }
```

Old `TryBuildSymbolicDelegateTarget` returned a target for **any** delegate-shaped parameter. The gate now silently drops fully-closed delegate candidates. Its caller `TryResolveSymbolicDelegateTargetForCtor` (`ExpressionBinder.Calls.cs:1497-1512`) relies on seeing *all* candidates so disagreeing overloads cancel out (`target = null; return false`). The gate hides the disagreeing candidate, so a single symbolic target gets **forced** onto a lambda that matched a different overload.

C# fixture:
```csharp
public sealed class Holder<T> {
    public Holder(Action<T> cb)   { Kind = "symbolic"; }
    public Holder(Action<int> cb) { Kind = "closed"; }
    public string Kind { get; }
}
```
```gsharp
let h  = Holder[Src]((x Src)   -> Console.WriteLine(x.N));  Console.WriteLine(h.Kind)
let h2 = Holder[Src]((x int32) -> Console.WriteLine(x));    Console.WriteLine(h2.Kind)
```

| compiler | result |
|---|---|
| base `d4038f1d` | `symbolic` / `closed` ✅ |
| refactor `fcce82af` | `symbolic` / `closed` ✅ |
| **HEAD `99019f64`** | **COMPILE FAIL — `GS0155: Cannot convert type 'Src' to 'int32'` + `GS0159`** ❌ |
| HEAD − gate only | `symbolic` / `closed` ✅ |

Isolation exact: only edit was deleting those 7 lines. Regression is ctor-only; the method-overload analogue is `symbolic,closed` on both base and HEAD.

Removing the gate costs nothing: `Compiler.Tests` 3588/3588 pass; targeted `Core.Tests` 420/420; all 8 PR tests pass; 148-sample corpus `BYTEDIFF=0 DIAGDRIFT=0 EXITDRIFT=0`.

### Ponytail: `+125/-85` is too big, measured

A minimal alternative is **−27/+1 vs HEAD** (`Invocation.cs` net +20 instead of +40), deleting: the unexercised 5-arg overload + `openMethodDefinition`/`methodTypeArguments` plumbing; the `carries` gate; the 3-arg `MapOpenClrTypeToSymbolic` call. All 8 PR tests pass, corpus byte-identical, probe output byte-for-byte same as HEAD.

**Genuinely right, keep:** the shared-`functionType` tail restructure (de-duplicates the slot-filling tail; hoisting into each branch costs ~28 duplicated lines). Reusing `TryGetLambdaTargetFunctionTypeFromSymbol` (`MemberLookup.cs:2304`, 18 call sites) instead of hand-rolled `Invoke` reflection — also subsumes `Expression<TDelegate>` unwrap and G# native function types for free. `RequiresSymbolicProjection` reused, no parallel predicate written.

### Mutation table (independent, one at a time, rebuilt each time)

Control before every row: `Compiler.Tests(2918+2889) 8/8`, `Core.Tests(2919) 1/1`. No row build-red.

| # | Mutation | `file:line` | Detected |
|---|---|---|---|
| R1 | site-A surrogate `EnumSymbol?int:object` → `object` | `Binder.cs:3019` | ✅ |
| R2 | apply enum surrogate at site B | `Binder.cs:3818` | ✅ |
| R3 | apply enum surrogate at site C | `Binder.cs:4047` | ✅ |
| R4–R6, R8, R9 | drop defensive terms | `Binder.cs:4098-4108` | ❌ survive (pre-existing) |
| R7 | drop `hasSymbolicArgument = true` | `Binder.cs:4100` | ✅ |
| **H1** | remove `carries` gate | `Invocation.cs:822` | ❌ **→ BLOCKER 1** |
| H2 | kill `IsGenericParameter` branch | `Invocation.cs:1489` | ✅ |
| H3–H7 | arity check, 5-arg overload, null guard, method args | `Invocation.cs:812-1498` | ❌ survive |

**"Pins 6/6" does not survive branch-granularity counting.** R4/R5/R6/R8/R9 are pre-existing defensive terms. But **H1, H3, H4, H5, H6, H7 are newly authored in this PR and every one survives** — ~25 lines of new logic with zero coverage, one of which is a live regression.

Pins independently confirmed: **8** (`p01,p02,p04,p06,p08,p09d,p13,p14`), each compile + IL-load + execute + expected stdout. Guards independently confirmed: **8**, plus 12 more shapes in `ExistingDelegateTargetingControls_RemainValid`. Author's "12 guards" is fair.

### The 53 deleted lines in `Issue2889LambdaThunkNestedGenericTests.cs` — legitimate

The deleted block is `InlineLambdaThroughNestedErasedReceiver_ReportsCurrentDiagnostics` plus its single-use helper. It was a **known-limitation pin** asserting `GS0159`/`GS0155`, and PR #2913's body says so verbatim: *"This current diagnostic is pinned so it cannot silently change, and the underlying general limitation is tracked in #2918"*. #2918 fixes exactly that program — base → `GS0159`+`GS0155`, HEAD → compiles, IL-verifies, runs, prints `7`. The pin *must* fail at HEAD. Coverage is subsumed and upgraded into `Issue2918InlineLambdaErasedReceiverTests.cs:99-101` (compile + `IlVerifier` + `AssemblyLoadContext` + `GetTypes()` + subprocess). The in-process Roslyn consumer at `:437` was **not** touched. **Not a papered-over regression** — but the PR body never mentions the deletion, and the pin existed precisely so this could not change silently. Document it.

### `fcce82af` no-op verdict, measured at that commit

Three separate clones; compiler output confirmed deterministic. 148 attempted / 139 emit; exit codes identical; `BYTEDIFF=0`; raw `DIAGDRIFT=139` **entirely** from the `Wrote <path>` output-path string; path-normalized `DIAGDRIFT=0`. **True no-op.** Base vs full HEAD over the same corpus: `BYTEDIFF=0 DIAGDRIFT=0 EXITDRIFT=0`.

### SHOULD-FIX — the 148-sample proof has almost no signal for #2919

Per-site tripwire on the `fcce82af` build across the corpus:

```
54 site=A sym=False     7 site=A sym=True   (2 files only)
 0 site=B (any)         ← NEVER EXECUTED
 3 site=C sym=False     0 site=C sym=True
 0 enum=True anywhere
```

Only **2 of 148** files exercise symbolic projection at all. **Site B — one of the three blocks whose entire body was replaced — is never executed once.** The PR caveats the corpus for #2918 but presents the same table as proof for #2919 without that caveat.

Redeeming it: `Issue2919ImportedGenericProjectionTests.cs` is the right test and genuinely works — instrumented run shows all three sites, symbolic, with and without enums, and R1/R2/R3 confirm it fails in **both** directions.

### Reconciliation vs #2908 — mechanical, no conflict

#2908 touches `Invocation.cs` at one site (`+19/-0` at top of `BindAccessorCall`, ~line 1866); #2929 touches `TryBuildSymbolicDelegateTarget` (~782-830) and the deferred-lambda-slot loop (~1484-1615). Zero region overlap, semantically independent. **Caution for #2924/#2925:** if either reuses `TryBuildSymbolicDelegateTarget`, the `carries` gate is a trap that silently returns `false` for fully-closed delegates — another reason to delete it now.

### Falsification attempts that failed (i.e. the PR was right)

Enum surrogate preserved at site A only, verified by name; no parallel predicate written; site-A semantic widening (`|| type.ClrType == null`) unreachable across 148 samples + 25 probes; inline vs hoisted enums agree; no `List<object>` leak into public API; `CS0012` in C# consumers reproduces **on base** too (pre-existing); test harness uses the correct dual-pipe `ReadToEndAsync` pattern; `Core.Tests` 6914/1 skip reproduced exactly; the new test file is substance, not volume (~25 shapes, plus a live-detector control that corrupts a real method body and asserts rejection).

### SHOULD-FIX / NIT

- **SHOULD-FIX** `Invocation.cs:1491` — `AllMethodTypeParametersResolved` in the new branch is dead: `:1743-1745` returns `true` unconditionally when `DeclaringMethod == null`, which is every declaring-type parameter — the exact case the branch exists for. Delete it or add the shape that needs it.
- **NIT** `Invocation.cs:1498` — arity check unpinned. A 2-param lambda in an `Action[Src]` slot compiles clean on **both** base and HEAD then throws `ArgumentException` at runtime. Not a regression; worth its own issue.
- **NIT** `Invocation.cs:1585-1596` — the `Issue #903` comment is orphaned; `unresolvedSlot` moved into the `else` branch.
- **NIT** — no C#-consumer test for a G#-emitted public surface in the new file (#2889 has one at `:437`).
- **Scope note** — `Box[T]` source-generic receivers still fail identically on base and HEAD (`GS0159`). Fix is confined to **imported** receivers, matching #2918's wording; worth a follow-up issue so the "root cause" claim isn't over-read.

### Path to mergeable

1. Delete the `carries` gate (`Invocation.cs:822-828`) — fixes BLOCKER 1 **and** shrinks the diff.
2. Add the ctor-overload probe as a regression test.
3. Cover or delete `AllMethodTypeParametersResolved` and the arity check.
4. Document the `Issue2889…` test deletion in the PR body.
5. Carry the "corpus is weak signal" caveat to the #2919 table; present `Issue2919ImportedGenericProjectionTests` as the real proof.

DavidObando:
## Rubber-duck review — round 2 (Opus 5), head `918a769f`

**Verdict: NOT MERGEABLE.** The round-1 blocker is genuinely closed and the ~25 unpinned lines are resolved — but a **new blocker** turned up in a shape nobody had probed.

### Cleared this round
- **Round-1 BLOCKER closed.**
- **~25 unpinned new lines: resolved.**
- **Deletions broke nothing** — `carries` removal is clean (Compiler 3713 / Core 6914 / Interpreter 453, corpus `BYTEDIFF=0`), the arity guard still fires, `AllMethodTypeParametersResolved` is still used and builds clean under `-warnaserror`, removing `|| candidate == null` cannot NRE (all four `true` paths assign non-null), and the #903 comment sits on the loop it describes.
- **#2889 is safe** — 4/4 pins pass.
- All three suite counts exact; all 8 merge-tree checks clean.

My round-1 CS0012 worry about C#-consuming G# public generic surfaces is **disproven by the author's own new test** — `PublicLibraryDelegateSignatures_RoundTripThroughCSharp` statically consumes `List<Action<Src>>` and `Dictionary<string,Func<Src,int>>`, passes, and fails under mutation N1. A real pin, not decoration.

### BLOCKER — imported **named** delegates through an erased generic slot
`ExpressionBinder.Calls.Invocation.cs:799-804` returns only the *shape* (`FunctionTypeSymbol`) and discards `mapped`'s delegate identity; `:1575` then propagates only `functionType.ParameterTypes`. For `Action`/`Func` the shape round-trips. For any other imported delegate it does not. Introduced by `99019f64`, still present at `918a769f`.

The base **rejects** these programs; HEAD **accepts** them and emits wrong code. Inline no longer matches hoisted — the exact parity property #2918 exists to establish.

**Collection slot — compile error becomes runtime crash:**
```
BASE d4038f1d:  q3_predicate    COMPILE-FAIL :: GS0159 / GS0155
HEAD 918a769f:  q3_predicate    OK :: run(134) :: Unhandled exception. System.ArgumentException:
                The value "System.Func`2[q3.Src,System.Boolean]" is not of type
                "System.Predicate`1[q3.Src]" ... at List`1.System.Collections.IList.Add(Object)
```
`ilverify` reports **Verified** — the bad store hides behind `IList.Add(object)`. Precisely the case where IL verification is not sufficient.

**Typed generic ctor slot — unverifiable IL and silent type corruption:**
```
q7_boxpred  Box[System.Predicate[Src]]((s Src) -> s.N > 2)
  BASE : COMPILE-FAIL :: GS0155 / GS0159 x3
  HEAD : OK :: System.Func`2[q7.Src,System.Boolean]/True/
  ilverify: Error [StackUnexpected] [found ref 'System.Func`2<q7.Src,bool>']
                                    [expected ref 'System.Predicate`1<q7.Src>']
```

**Attribution controlled** so this is not over-claimed: the *second* ilverify error in `q7` is the pre-existing #313/#939 erasure artifact — base emits the same single error for the hoisted variants it accepts (`q10`/`q11`: base 1 error, head 1 error, identical). The `found Func / expected Predicate` error is the **extra** one and appears only at HEAD.

| program | base | HEAD |
|---|---|---|
| `Box[Predicate[Src]]` **hoisted** | `System.Predicate\`1[Src]` / True | same |
| `Box[Predicate[Src]]` **inline** | compile error | **`Func\`2` + 1 extra ilverify error** |
| `List[Predicate[Src]].Add` **hoisted** | True | True |
| `List[Predicate[Src]].Add` **inline** | compile error | **`ArgumentException` at runtime** |
| `Box[Action[Src]]` inline (control) | compile error | correct |
| `List[Sink[Src]]` G#-source named delegate (control) | 3 | 3 |

Scope: **imported** named delegates only — `Predicate`, `Comparison`, and by extension `Converter`, `EventHandler`, any imported `delegate`. G#-source named delegates and `Action`/`Func` are fine.

**Zero coverage**: no mention of `Predicate` or `Comparison` in `Issue2918InlineLambdaErasedReceiverTests.cs`; #2889's `ImportedNamedDelegates_…` covers only direct hoisted assignment, not an erased generic slot.

**Fix direction:** in the new branch at `Invocation.cs:1464-1474`, after `TryBuildSymbolicDelegateTarget` succeeds, require that `mapped`'s CLR delegate identity is the canonical `Action`/`Func` for the recovered shape; otherwise set `candidateUsable = false` and restore the base diagnostic. That keeps all 8 pins and stops emitting wrong code. Carrying the real delegate identity through `slotTargets` is the proper fix and is a bigger change — reasonable to file separately.

Note the irony: this is a guard of exactly the kind deleted this round as "unpinned". The difference is that it is **load-bearing**, and there is now a probe to pin it.

### NITs
- PR body cross-references **#2936/#2937 backwards** (lines 23, 78, 87): #2936 is the source-generic receiver issue, #2937 the arity issue. Both issues are themselves accurate; only the PR text is swapped.
- Interpreter gained 0 tests although evaluate-mode binding demonstrably changed (`gsc` with no `/out:`: `p01_repro` and `p02_nested` go from `GS0159` on base to `Success` on head). A runtime interpreter test is blocked by a pre-existing evaluator limitation, but a bind-only pin is possible.

### Falsification attempts that failed
Two samples that appeared to regress were **my error** — a stale mutant binary; after rebuild `EXITDRIFT=0`. Eight Core failures in my scratch clone reproduce **identically on `d4038f1d`** (`diff base_fails.txt head_fails.txt` → IDENTICAL), so they are clone-environmental, not the PR.

### Path to merge
1. Reject or correctly type imported named-delegate targets at `Invocation.cs:1464-1474`; pin with `q3_predicate`, `q3c_comparison`, `q7_boxpred` plus the hoisted `q11` twin.
2. Swap #2936/#2937 in the PR body.
3. Optional bind-only Interpreter pin for `p01`.

Everything else this round is solid and, on the specific points I attacked, better than my round-1 recommendation.

DavidObando:
## Round 3 review — NOT MERGEABLE (1 blocker, 2 nits)

Reviewed: branch head `2e987038` rebased onto `origin/main` `e21677f6` → `5c3b3f64` (faithful cherry-pick, **zero conflicts**, production diff byte-identical to `2e987038`).

Round-2 blocker (imported named delegates through an erased generic slot) is **genuinely closed** — verified by compile → ilverify → `Assembly.Load`/`GetTypes()` → child-process execution. Attribution held: the `found ref Func\`2 / expected ref Predicate\`1` error is gone, and pre-existing #313/#939 is **unchanged** (`q10`/`q11`: 1 error both sides, identical text). The fix masks nothing.

### BLOCKER — guard mis-scoped at `ExpressionBinder.Calls.cs:1497`

The guard is applied to **every** ctor candidate rather than only erased slots. The invocation site got this right (gated inside `openParameterType.IsGenericParameter`, `Invocation.cs:1490`); the ctor site did not.

```csharp
// ExpressionBinder.Calls.cs:1496-1499 — no IsGenericParameter gate
        || !IsCanonicalFunctionDelegate(mappedDelegate)
        || candidate == null)
    { continue; }
```

Any imported **constructed** generic delegate as a ctor parameter plus an inline lambda is now rejected — even fully typed, with no erasure anywhere.

| probe | `main` `e21677f6` | head |
|---|---|---|
| `r01` `PredHolder[Src](...)` — `Predicate<T>` | **RUN(0) `pred`** | **COMPILE-FAIL** GS0155/GS0159/GS0158 |
| `r02` `CompHolder[Src](...)` — `Comparison<T>` | **RUN(0) `comp`** | **COMPILE-FAIL** |
| `r03` `EvtHolder[Src](...)` — `EventHandler<T>` | **RUN(0) `evt`** | **COMPILE-FAIL** |
| `r04` `ConvHolder[Src,int32](...)` — `Converter<T,U>` | **RUN(0) `conv`** | **COMPILE-FAIL** |
| `r05` `MyGenHolder[Src](...)` — user lib `MyGenCb<T>` | **RUN(0) `mygen`** | **COMPILE-FAIL** |
| `r10` `ExprPredHolder[Src](...)` — `Expression<Predicate<T>>` | RUN(134)¹ | **COMPILE-FAIL** |

¹ pre-existing #313/#939 runtime throw; `r09` `Expression<Func<T,bool>>` throws the same on **both** sides — correctly unchanged.

**Fix — 2 lines, measured rather than proposed:**

```diff
-                || !IsCanonicalFunctionDelegate(mappedDelegate)
+                || (parameters[paramIndex].ParameterType.IsGenericParameter
+                    && !IsCanonicalFunctionDelegate(mappedDelegate))
```

With it applied (Release 0/0, unmutated control captured first, restored + rebuilt after): `r01`–`r05` and `r10` all return to `main` behaviour; the round-2 blocker **stays closed** (`q3_predicate`, `q3c_comparison`, `q7_boxpred` still COMPILE-FAIL with byte-identical diagnostics); the #2918 pin survives (`q8_boxaction` RUN(0)); and **all 17 targeted tests pass**. A pin for the `r01` shape is also needed — its absence is why the guard shipped mis-scoped.

### Corpus — state the zero

148 samples / 139 assemblies, `main` vs head: `BYTEDIFF=0 EXITDRIFT=0 FULLNORMDIAGDRIFT=0`.

**0 of 148 samples exercise an imported named delegate through an erased generic slot.** The corpus was green **while `r01`–`r05` regressed** — corpus-green is not coverage here.

### Cleared
- **#2935 merged-result clean**: `Issue2928TupleFunctionParityTests` 3/3, `Issue2928CallableMaterializationTests` 15/15 at the merged head. #2889's 4 tests pass.
- **Round-2 deletion audit**: probed the generic-method / bare-method-type-parameter / closed-delegate-receiver shape; `main` and head both RUN(0). **No second instance found** — this round's defect is newly introduced, not a deletion.
- **#2948** is real, accurate and correctly scoped. One correction: its body says the PR "conservatively rejects these **inline** cases", which understates — the mis-scoped guard also rejects non-erased fully-typed ctor params. Fixing the guard makes the body true.
- `merge-tree` clean against `origin/main` and #2908, #2911, #2912, #2926, #2932, #2950.

### Test counts (`--list-tests` declared-method counts, same tooling both sides)
| suite | `e21677f6` | head | Δ |
|---|---|---|---|
| Compiler.Tests | 3756 | 3765 | **+9** |
| Core.Tests | 6787 | 6788 | **+1** |
| Interpreter.Tests | **468** | **469** | **+1** |

The PR body's 3762 / 6914+1 / **454** were measured at `2e987038` on merge-base `eaf59056`, predating #2935; Interpreter 454 confirms the staleness.

### NITs
1. `Invocation.cs:826-830` — `IsCanonicalFunctionDelegate` matches `FullName` only, assembly-blind; a user type named `System.Func\`2` in its own assembly passes. Not exploitable in practice; note it in #2948.
2. `Invocation.cs:673` — `TryBindSymbolicReceiverCall` discards `mapped` via `out _` with no guard. Probed clean, but document why, or #2948 will re-open the hole there.

DavidObando:
## Rubber-duck review — round 4 — **MERGEABLE**

**Head reviewed: `3b7a05a94`.** Base measured: `8dddbbf28`. Merged-result also verified against `main` = `a6b1561f2`.

All evidence below is base-vs-head, from two `git clone --shared` scratch trees built `dotnet build GSharp.sln -c Release` (0 warnings / 0 errors both sides). Nothing in the worktree was modified.

---

### Round-3 blocker: CLOSED

`ExpressionBinder.Calls.cs:1501-1504` now scopes the named-delegate guard to `parameters[paramIndex].ParameterType.IsGenericParameter`. 70-probe battery (`Assembly.Load` + `GetTypes()` + child-process execution — not `IlVerifier` alone):

| set | base `8dddbbf28` | head `3b7a05a94` |
|---|---|---|
| 15 over-rejection shapes (`Predicate`/`Comparison`/`EventHandler`/`Converter`/user-lib generic delegate/`Expression<Predicate<T>>`) | — | **all 15 identical** |
| round-2 hole (`Predicate`/`Comparison`/boxed `Predicate` through erased slot) | COMPILE-FAIL | COMPILE-FAIL, same GS0155+GS0159 |
| `q8n` — inline `Action` through erased ctor slot | **COMPILE-FAIL** | **RUN(0) → `System.Action`1[q8n.Src]` / `4`** |

Boundary probes on a purpose-built fixture (erased property / indexer / method / generic-method slots): erased **method** slot Action base GS0155+GS0159 → head runs correctly; erased method slot **Predicate** rejected on both; indexer slot identical; property slot GS0155 on both; non-erased ctor slots identical. Narrowing is correct **and** complete — no shape silently un-guarded.

### `Issue2889LambdaThunkNestedGenericTests.cs` −53: deletion is **mandatory**, not convenient

Deleted block is one `[Fact]` + its single-use helper, asserting `GS0159`/`GS0155` for `List[System.Action[List[Src]]].Add(inline)`. Ran that exact program: base **COMPILE-FAIL** (matching the pin) → head **RUN(0) → `42`**. The assertions cannot hold once #2918 lands. Replacement at `Issue2918InlineLambdaErasedReceiverTests.cs:130` (inside the fact at `:91`) is a strict upgrade: same program, now compile + ilverify + `AssemblyLoadContext` + `GetTypes()` + reified-signature assert + subprocess + exact stdout. Other 4 facts in that file unchanged and passing.

### Deletions justified by mutation survival: **none remain**

Round 2 deleted five constructs on a round-1 survival table; that produced the round-2 blocker. At this head they are **all restored**:

```
signature diff, Invocation.cs base→head:   16a17
>     private static bool IsCanonicalFunctionDelegate(TypeSymbol type)
   one method added, zero removed
AllMethodTypeParametersResolved   base 4 → head 4
arityByIndex                      base 3 → head 3
```

The `Binder.cs` −69 is #2919 consolidation, justified structurally, not by survival (below).

### Mutation table — hardened harness

`-m:1`; exit-code gate (not `grep "0 Error"`); zero-byte assert; per-dir md5 propagation across `{Core, Compiler, Compiler.Tests, Core.Tests, Interpreter.Tests}`; unmutated control first; restore + rebuild + md5 round-trip after each.

| # | mutation | result |
|---|---|---|
| X1 | revert round-3 scoping at `ExpressionBinder.Calls.cs:1503` | **KILLED** — `ImportedConstructedPredicateConstructor_TargetsInlineLambda_LoadsAndRuns` |
| X2 | delete ctor guard entirely | **KILLED** — `ImportedPredicateThroughErasedConstructorSlot_IsRejected` |
| X3 | delete method-slot guard at `Invocation.cs:1505` | **KILLED** ×2 — Predicate + Comparison list pins |
| X5 | `IsCanonicalFunctionDelegate` rejects `Expression<T>` | **SURVIVED** |

X1 is the round-3 blocker, now pinned by a test that did not exist before this round.

X5 round-trip proof (all five dirs agreeing at both ends, none zero-byte, `BUILD_EXIT=0`, `0 Error(s)`):
`0231589de52112984569f7a73fe24287` → `e8ab39c5e260e123534fcaeb5d594d0a` → `0231589de52112984569f7a73fe24287`

X5's survival was chased to ground with 5 expression-tree probes, all byte-identical to unmutated head. **This retracts my round-3 concern** that `List[Expression[Func[…]]].Add(inline)` routes through the new branch — it does not; that wrong code is pre-existing on both sides and the PR body does not overclaim.

### Corpus — and its **zero**

148 samples, `8dddbbf28` vs head: **`BYTEDIFF=0`, `DIAGDRIFT=0`** (path-normalised), **`EXITDRIFT=0`**, 139/139 assemblies emitted.

**0 of 148 samples exercise an imported CLR named delegate through an erased generic slot.** The 3 samples mentioning `Predicate`/`Converter` all declare G#-source `type Predicate[T any] = delegate …`. Corpus-green proves absence of collateral damage and proves **nothing** about the fix — do not read it as coverage.

### Test counts — same base, executed, reconcile exactly

| suite | base `8dddbbf28` | head `3b7a05a94` | Δ | static `[Fact]` Δ |
|---|---|---|---|---|
| Compiler | **3893** / 0 failed | **3903** / 0 failed | +10 | +11 new − 1 deleted = **+10** |
| Core | **7028** + 1 skip | **7029** + 1 skip | +1 | +1 |
| Interpreter | **489** | **490** | +1 | +1 |
| LSP | **452** | **452** | 0 | 0 |

Author's table is exact. **0 failed on both sides** — the 19 golden-sample failures are a build-closure artifact and vanish when you build `GSharp.sln`; they were never environmental.

### Merged-result risk vs `main` = `a6b1561f2` (#2961 GS0100 on lambda bodies, #2973)

Performed the real merge, not just `merge-tree`:

```
merge 3b7a05a94 + a6b1561f2 → 80f7694aa, clean
dotnet build GSharp.sln -c Release → exit 0, 0 Warning(s), 0 Error(s)
Compiler --filter Issue2918|2889|2921|2936|2928 → 31/31 pass
Core     --filter Issue2919                     →  1/1  pass
Interpreter full                                → 492/492 (489 base +1 PR +2 from #2921 — reconciles)
70-probe battery, head vs merged                → BYTE-IDENTICAL
```

No #2961/#2973 interaction on any probed shape.

`merge-tree` vs open PRs: **#2978 #2973 #2972 #2968 #2967 #2966 #2964 #2963 #2961 #2912 all CLEAN.** **#2903 CONFLICTS** on `ExpressionBinder.Calls.Invocation.cs` (#2903 is `+111/−86` there and clean against `main` on its own). Not this PR's defect, but whichever lands second needs a rebase **plus re-verification** — both rewrite deferred-lambda target resolution in the same block.

---

## Proven vs claimed

- **#2918 — PROVEN.** Three independent constructed programs flip base-fail → head-correct.
- **#2919 — PROVEN as real dedup.** Three blocks → one helper (`Binder.cs:4132`) with call sites `:3075`/`:3871`/`:4100`; zero residual `hasSymbolicArg = true`; reuses the pre-existing `TypeSymbol.RequiresSymbolicProjection` rather than a parallel predicate. Sites B and C are **exact** no-ops. Enum surrogate stays scoped to site A only. All defensive sub-terms individually present. `Issue2919ImportedGenericProjectionTests.cs` pins both directions in one test (`List[Mode]`→`Int32`, `System.Collections.Generic.List[Mode]`→`Object`). One honest asterisk: site A **gains** `|| ta.ClrType == null` twice; traced the corner — old path yielded `MapClrTypeToReferences(null)` → `null` → `MakeGenericType` throws → caught → bind failure, so the delta is **failure→success only** and cannot regress a working program.
- **#2925 — NOT CLAIMED by this PR** and not verified. `closingIssuesReferences` = `[2918, 2919]`.

## Findings

**BLOCKER — none.**

**SHOULD-FIX S1 — `ExpressionBinder.Calls.Invocation.cs:1619` makes a pre-existing null guard unsatisfiable.**
`FunctionTypeSymbol.Get` does `returnType ?? TypeSymbol.Void` (`FunctionTypeSymbol.cs:182,186`). Building the function type at `:1611` and then re-reading `functionType.ReturnType` at `:1619` means a `null` mapped return is now `Void`, so `mappedReturn != null` can never be false and the slot records a recovered `void` return instead of falling back to `inferReturnTypeFromBody`. This is on the **else** branch — the pre-existing #1512/#2345/#903 path, not the new #2918 one. The sibling guard for *parameters* (`mapped == null → slotUsable = false`, `:1578`) was kept, which is evidence the original author considered null reachable. This is an **algebraic** claim, not a mutation claim. I could not construct a reaching program, which by the standing rule is missing coverage, not proof of deadness. Fix: hoist `mappedReturnType` out of the `else` and use it at `:1619`.

**SHOULD-FIX S2 — PR body's deletion paragraph is stale.** It states the 5-arg overload, method-arg plumbing, `AllMethodTypeParametersResolved` check, arity check and null guard "were deleted." At `3b7a05a94` none are (counts above). It describes an intermediate round-2 state that round 3 reverted. A reader would hunt for holes that no longer exist — or believe guards are missing that are present.

**SHOULD-FIX S3 — `Issue2918InlineLambdaErasedReceiverTests.cs:744,748,749` unbounded waits in the timeout handler.** `:741` correctly bounds at 30 s, but after `Kill(entireProcessTree: true)` the `:744` `WaitForExit()` and the `:748`/`:749` `GetAwaiter().GetResult()` are unbounded — a descendant holding the inherited pipe keeps them blocked. This is the #2934 bug shape reproduced in new test code; a timeout handler that hangs is worse than the hang it replaces. Fix: `WaitForExit(5_000)` and bounded `Task.Wait`.

**NIT N1** — `Invocation.cs:817-831` `IsCanonicalFunctionDelegate` is an assembly-blind `FullName` prefix match. Documented in **#2948**, which is open, correctly titled, and genuinely scoped as the positive follow-up rather than deferral of this PR's work. Correct interim state.

**NIT N2** — `Invocation.cs:819-822` `Expression<T>` unwrap is unreachable by any program I could build (X5 survived). **Keep it.** Recorded only so nobody deletes it on a survival table.

**NIT N3** — the enum-surrogate asymmetry (`Action[Mode]`→`Int32` vs `System.Action[Mode]`→`Object`) is now pinned as intended behaviour. Faithful preservation is right for a refactor, but the inconsistency deserves its own issue.

**NIT N4** — merge-order collision with #2903 (above).

## Retractions / revisions from earlier rounds

- **RETRACTED**: my round-3 suspicion that `Expression`-wrapped targets route through the new guard branch. Disproved by X5 + 5 probes.
- **REVISED**: round 1 called #2919 a measured no-op; precisely, sites B/C are exact no-ops and site A has a real `ClrType == null` delta whose direction is failure→success only.
- **REVISED**: round 2 recorded the arity check's removal as unpinned risk. It is not removed; the *new* branch simply never had one. Constructed both mismatch directions — base and head produce **byte-identical** wrong code (compiles clean, runtime `ArgumentException`), i.e. pre-existing, correctly attributed to **#2937**.
- Rounds 1–3 blockers were all real; all three are closed and all three are now pinned (X1/X2/X3 kill them).

## Hygiene

`git status --porcelain` empty in the review worktree and in both scratch clones (all 5 mutants restored, md5 round-trips verified). No e2e. No `git stash`. `~/.nuget/packages/gsharp.net.sdk/` untouched, all original mtimes. Across 9 solution builds under heavy contention, zero `MSB3883`/`MSB3030`/`GSharp.Core` load failures — data point for #2974.

**Recommendation: rebase onto `a6b1561f2` and merge.** The merged result is already built and probed clean. S1–S3 are worth a follow-up commit; none of them blocks.

DavidObando:

Specifically:

Issue #2928, opened 2026-07-31
Interpreter cannot materialize tuples containing function values


DavidObando:
## Repro

Discovered while probing interpreter parity for #2924 on `origin/main` at `d4038f1d`.

```gsharp
package TupleFunctionInterpreter
import System

let handler (int32) -> int32 = (value int32) -> value + 1
let t = (handler, 0)
Console.WriteLine(t.Item1(41))
```

Run through interpreter path:

```text
dotnet out/bin/Release/Compiler/gsc.dll Program.gs
```

Actual:

```text
error GS9999: Constructor on type 'System.ValueTuple`2[System.Func`2[System.Int32,System.Int32],System.Int32]' not found.
    (handler, 0)
```

Expected: `42` and `Success.`

## Attribution

`Evaluator.Expressions.cs:701-713` evaluates a G# function literal to the interpreter's callable representation, then unconditionally passes that object to `Activator.CreateInstance` for the tuple's CLR `ValueTuple<Func<int,int>, int>` type. The runtime constructor cannot accept the interpreter object. Existing `Item1` syntax reproduces this independently of #2924's numeric-selector parser/binder gap.

Issue #2925, opened 2026-07-30
Delegate-typed interface property cannot be invoked


DavidObando:
Found while reviewing #2885 / PR #2908.

## Repro

```gsharp
package InterfaceDelegateProperty
import System
interface IBox {
    prop P System.Action[int32] { get }
}
class Box : IBox {
    let Handler System.Action[int32]
    init(handler System.Action[int32]) { Handler = handler }
    prop P System.Action[int32] -> Handler
}
func Main() {
    let box IBox = Box((value int32) -> Console.WriteLine(value))
    box.P(1)
}
```

Current `origin/main` (`d4038f1d`) reports:

```text
error GS0159: Cannot find function P.
```

Expected delegate property load followed by `Invoke(1)`.

## Root-cause pointer

Delegate-member fallback in `ExpressionBinder.Calls.Invocation.cs:2386` and `:2529` handles source `StructSymbol` receivers through `TryBindUserStructDelegateMemberInvocation`. No equivalent path handles source `InterfaceSymbol` delegate properties, so lookup falls through to GS0159.

DavidObando:
Cross-PR trap found while reviewing #2908: #2925 is unaffected today (`box.P(1)` reports `GS0159` identically on `main` and #2908). Once the `InterfaceSymbol` delegate-property path is added around `ExpressionBinder.Calls.Invocation.cs:2386` / `:2529`, nullable interface delegate properties will reach delegate-call binding for the first time. That path must route through `OverloadResolver.TryReportNullableDelegateReceiver` (`OverloadResolver.cs:312`) or it will reopen #2885 for this receiver shape.

An interface `prop P { get }` is computed, therefore not smart-cast-stable. `if box.P != nil { box.P(1) }` must report `GS0503`; working remedy is `if let`.

PR #2932, opened 2026-07-31
Fix tuple-element delegate calls

### LINKED ISSUE #2924: Tuple-element delegate call is silently elided
Found while reviewing #2885 / PR #2908.

## Repro

```gsharp
package TupleDelegateElided
import System
func Main() {
    let handler System.Action[int32] = (value int32) -> Console.WriteLine("tup:{0}", value)
    let t = (handler, 0)
    Console.WriteLine("before")
    t.0(1)
    Console.WriteLine("after")
}
```

Current `origin/main` (`d4038f1d`) compiles and runs with:

```text
before
after
```

Expected `tup:1` between those lines. Call is silently dropped even though tuple element is a non-null delegate.

## Root-cause pointer

`Parser.Expressions.cs:733` routes every `.` continuation through `ParseNameOrCallExpression`, which has no numeric tuple-selector call shape. Tuple binding in `ExpressionBinder.Access.MemberLookup.cs:439` only recognizes `Item1`…`ItemN`, so `t.0(1)` never reaches indirect-call binding as a tuple-element delegate value.

### LINKED ISSUE #2944: Curried member invocation reports GS0159
## Reproduction

```gsharp
package CurriedMemberCall
import System

class Factory {
    func Make(seed int32) (int32) -> int32 {
        return (value int32) -> seed + value
    }
}

func Main() {
    let factory = Factory()
    Console.WriteLine(factory.Make(40)(2))
}
```

On `origin/main` at `d4038f1d`, this fails with `GS0159: Cannot find function .` instead of printing `42`.

## Cause

Member continuation parsing consumes both the direct member call and the following indirect invocation into the accessor's right-hand side. The binder then attempts to resolve the nested call as a member named `.` instead of invoking the function value returned by `Make`.

The same parser gap affects receiver-wide shapes such as generic member calls, indexing after a member call, and longer curried chains. PR #2932's `stopBeforeIndirectInvocation` reshape fixes this while fixing #2924; tests there pin the independent non-tuple behavior.

DavidObando:
## Root cause

The lexer tokenizes `.0`, `.1`, etc. as leading-dot `NumberToken`s. Postfix parsing did not contextualize those tokens as member selectors, and member continuations greedily consumed following indirect invocations into the accessor's right-hand side. Therefore `t.0(1)` never became a call whose callee was the full `t.0` accessor; on `main` it was accepted as separate expressions and silently dropped. Tuple member binding also recognized only `Item1`…`ItemN`.

Review exposed two adjacent gaps. Lifted calls escaped `BoundNullConditionalAccessExpression`, so their arguments and invocation executed outside the nil guard. Assignment binding also had independent member lookup, so numeric selectors worked for reads but not writes.

## Fix

- Contextually split digits-only leading-dot number tokens in postfix position into the existing accessor syntax shape, preserving source spelling and exact spans.
- Accept zero-based numeric continuations after `.` and `?.`, while preserving one-based `ItemN` access.
- Stop member continuations before indirect invocation so the full receiver becomes the callee. This also fixes #2944, the independent curried-member-call `GS0159` bug.
- Apply the same reshape to pointer-arrow and null-assertion continuations.
- When an indirect callee is null-conditional, bind against `WhenNotNull` and rebuild the null-conditional result around the completed invocation. Arguments and invocation now remain inside the nil guard for structural functions, named delegates, and CLR delegates.
- Normalize valid tuple selectors to their CLR `ItemN` field names in both simple and compound assignment lookup.

This is the narrowest complete shape. A lexer change would add state, alter raw token consumers, and still miss spaced `t. 0`. A postfix-core-only change misses nested receivers such as `holder.Value.0(5)` and `factory.Make()(41)`. Lifting without invocation re-rooting breaks `?.`; suppressing the lift reintroduces silent elision. Assignment needs its own two call sites because it intentionally bypasses read lookup, but both use one shared tuple-index helper.

Production diff from latest base `17789519`: 162 insertions, 21 deletions across six files (183 changed lines, net +141).

## Review blocker probes

Compiled with the branch compiler and executed as a real `dotnet <app>.dll` process:

| Probe | Result |
|---|---|
| nil `g?.Plain(1)` | short-circuited; `A ok` printed |
| nil `g?.Make(40)(2)` | short-circuited; `B ok` printed |
| nil `g?.Get().0(1)` | short-circuited; `C ok` printed |
| nil `t?.0(1)` | short-circuited; `D ok` printed |
| nil `g?.Arr()[0]` | unchanged; `E ok` printed |

The compiled non-nil structural, named-delegate, and asserted forms each produced `42`. A side-effecting receiver probe confirms both live and nil lifted calls evaluate the receiver exactly once. Interpreter pins produce `42` for non-nil `g?.Make(40)(2)`, `end` for nil `g?.Make(40)(2)`, `end` for nil `t?.0(1)`, and `42`/`end` for the asserted live/nil pair.

## #2908 rebase composition

Originally rebased across #2908 at `914b42a7`; final rebase targets latest `main` at `17789519` (including requested `f2ecef88` plus #2950). The sole semantic conflict was the shared accessor junction in `ExpressionBinder.Access.MemberLookup.cs`. Its resolution keeps both fixes:

- #2908's `receiverSyntax` parameter on `BindAccessorStep`
- `nested.LeftPart` passed as the receiver syntax for nested accessor calls
- `receiverSyntax` forwarded into `BindAccessorCall`, preserving GS0503 reporting
- this PR's shared tuple-index helpers and numeric tuple read binding

Post-resolution targeted controls pass: B1 Compiler tests **7/7**, #2885 Compiler tests **45/45**, and #2885 Interpreter tests **2/2**. The combined Issue2924 + Issue2885 Compiler run is **52/52** after the latest rebase. Thus guarded lifted calls short-circuit, while nullable CLR-delegate calls without valid narrowing still report GS0503.

Rebase preservation is byte-for-byte at the production patch level: `914b42a7..c6be1ba6` and `17789519..571caa46` are both 347 patch lines with identical sorted `+`/`-` line sets. Including the NIT hardening, old and new production patches are both 354 lines with identical sorted line sets.

## Review findings

- **B1:** fixed by re-rooting completed indirect invocations into null-conditional `WhenNotNull`; live and nil receivers are each evaluated exactly once.
- **S1:** fixed. `t.0 = 5` and independent `t.0 += 6` pins pass compiled and interpreted.
- **S2:** fixed and directly covered after merging #2935. The nil test requires `Non-static method requires a target.`, and a live G# lambda stored in a tuple now executes through `functions.0(41)` and prints `42` in the interpreter.
- **S3:** fixed for pointer-arrow and `?.` + `!!` continuations, with parse pins and asserted-call runtime coverage.
- **#2944 pins:** plain `factory.Make(40)(2)`, generic `factory.Make[int32](40)(2)`, indexing `factory.Make(40)[0](2)`, and longer `factory.Make(40)(2)(3)` chains.

## Pin / guard split

Against `d4038f1d`, one classification per issue test:

| Project | Pins failing on main | Guards passing on main |
|---|---:|---:|
| Core.Tests | 15 | 5 |
| Compiler.Tests | 6 | 1 |
| Interpreter.Tests | 15 | 2 |
| **Total** | **36** | **8** |

All 44 feature tests pass on this branch. A separate missing-selector ICE regression pin also passes.

## Mutation testing

The original audit is corrected to **13/15 killed + 2 equivalent**. Both guards are unreachable from source. The parser guard was deleted; `numericIndex >= 0` was re-added defensively after selector normalization grew to three call sites.

Each review mutation below was applied alone, stayed build-green, had an unmutated same-worktree control immediately before it, and was restored before the next row.

| # | Semantic mutation | Control | Mutant result |
|---:|---|---|---|
| R1 | Disable null-conditional target recognition | Core 20/20 | 4 failed |
| R2 | Disable asserted null-conditional callee re-root | Core 20/20 | 1 failed |
| R2b | Stop after one null assertion | Core 20/20 | 1 failed |
| R3 | Return invocation outside null guard | Core 20/20 | 4 failed |
| R4 | Remove numeric normalization from simple assignment | Core 20/20 | 1 failed |
| R5 | Remove numeric normalization from compound assignment | Core 20/20 | 1 failed |
| R6 | Drop pointer-arrow stop flag | Core 20/20 | 1 failed |
| R7 | Drop null-assertion continuation stop flag | Core 20/20 | 2 failed |
| R8 | Make shared tuple field normalization inert | Core 20/20 | 1 failed |
| R9 | Disable numeric tuple-index resolution | Core 20/20 | 7 failed |
| R10 | Reuse pre-invocation null-conditional result type | Compiler 7/7 | 1 failed |
| R11 | Skip wrapper on structural-function invocation | Core 20/20 | 3 failed |
| R12 | Skip wrapper on named-delegate invocation | Compiler 7/7 | 1 failed |
| R13 | Skip wrapper on CLR-delegate invocation | Compiler 7/7 | 1 failed |

Post-review result: **14/14 reachable semantic mutants killed**.

## Over-fire corpus

Compiled all 148 `samples/**` + `e2etests/**` `.gs` files with separately published `914b42a7` and branch compilers, using identical assembly names/output paths:

- exit status: 148/148 identical
- diagnostics/output: 148/148 byte-identical
- emitted assemblies: 139/139 byte-identical
- failures: 9/9 failed identically

Parser inspection found 7/148 files containing tuple syntax, 0 numeric selectors, and 0 receiver-wide indirect member calls. Corpus signal is therefore strictly non-regression evidence for the global reshape; handwritten tests carry all feature signal.

## Validation

- Release warnings-as-errors solution build: 0 warnings, 0 errors

| Suite | `914b42a7` control | `d901683a` (914-based head) |
|---|---:|---:|
| Core.Tests | 6915 passed, 1 skipped | 6936 passed, 1 skipped |
| Compiler.Tests | 3802 passed | 3809 passed |
| Interpreter.Tests | 470 passed | 487 passed |
| Extensions.Tests | 104 passed | 104 passed |
| SDK.Tests | 57 passed | 57 passed |

Additional executed baseline at `f2ecef88`: Core.Tests **7007 passed + 1 skipped**; Compiler.Tests **3871 passed**. Latest-head Release warnings-as-errors build passes with 0 warnings/errors; full latest-head results are carried by PR CI.

Compiler runtime tests compile, IlVerify, `Assembly.Load`, call `GetTypes()`, and execute. The five blocker probes were also run as real `dotnet` processes.

## Invocation-binding footprint

`ExpressionBinder.Calls.Invocation.cs` remains untouched. In `OverloadResolver.Invocations.cs`, only:

- `BuildNullConditionalDelegateResult` capture parameter type (local -> variable), and
- `BindIndirectCallExpression` null-conditional target extraction/completion paths

were changed. No other invocation-binding region was modified.

## Separate issues

- #2928 accurately tracked interpreter callable-tuple materialization; merged PR #2935 fixes it, and this branch now includes the combined numeric-selector invocation pin.
- #2944 tracks the independent curried-member-call `GS0159` bug fixed by the parser reshape here.

Fixes #2924
Fixes #2944

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs
@@ -24,6 +24,36 @@ namespace GSharp.Core.CodeAnalysis.Binding;
 
 internal sealed partial class ExpressionBinder
 {
+    private static bool TryGetTupleElementIndex(
+        string memberName,
+        TupleTypeSymbol tupleType,
+        out int zeroBased)
+    {
+        if (int.TryParse(memberName, out var numericIndex)
+            && numericIndex < tupleType.Arity)
+        {
+            zeroBased = numericIndex;
+            return true;
+        }
+
+        if (memberName.StartsWith("Item", System.StringComparison.Ordinal)
+            && int.TryParse(memberName.Substring(4), out var oneBased)
+            && oneBased >= 1
+            && oneBased <= tupleType.Arity)
+        {
+            zeroBased = oneBased - 1;
+            return true;
+        }
+
+        zeroBased = -1;
+        return false;
+    }
+
+    private static string GetTupleFieldName(string memberName, TupleTypeSymbol tupleType) =>
+        TryGetTupleElementIndex(memberName, tupleType, out var index)
+            ? "Item" + (index + 1).ToString(System.Globalization.CultureInfo.InvariantCulture)
+            : memberName;
+
     private BoundExpression BindAccessorStep(
         BoundExpression receiver,
         ImportedClassSymbol classSymbol,
@@ -439,22 +469,7 @@ internal sealed partial class ExpressionBinder
                     // Phase 4.5 / issue #2924: tuple element access via
                     // one-based Item1..ItemN or zero-based .0...N selectors.
                     var memberName = ne.IdentifierToken.Text;
-                    var zeroBased = -1;
-                    if (int.TryParse(memberName, out var numericIndex)
-                        && numericIndex >= 0
-                        && numericIndex < tupleSym.Arity)
-                    {
-                        zeroBased = numericIndex;
-                    }
-                    else if (memberName.StartsWith("Item", System.StringComparison.Ordinal)
-                        && int.TryParse(memberName.Substring(4), out var oneBased)
-                        && oneBased >= 1
-                        && oneBased <= tupleSym.Arity)
-                    {
-                        zeroBased = oneBased - 1;
-                    }
-
-                    if (zeroBased >= 0)
+                    if (TryGetTupleElementIndex(memberName, tupleSym, out var zeroBased))
                     {
                         return new BoundTupleElementAccessExpression(null, receiver, tupleSym, zeroBased);
                     }
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs
@@ -29,7 +29,14 @@ internal sealed partial class ExpressionBinder
         TupleTypeSymbol tupleType,
         out int zeroBased)
     {
+        if (string.IsNullOrEmpty(memberName))
+        {
+            zeroBased = -1;
+            return false;
+        }
+
         if (int.TryParse(memberName, out var numericIndex)
+            && numericIndex >= 0
             && numericIndex < tupleType.Arity)
         {
             zeroBased = numericIndex;
@@ -52,7 +59,7 @@ internal sealed partial class ExpressionBinder
     private static string GetTupleFieldName(string memberName, TupleTypeSymbol tupleType) =>
         TryGetTupleElementIndex(memberName, tupleType, out var index)
             ? "Item" + (index + 1).ToString(System.Globalization.CultureInfo.InvariantCulture)
-            : memberName;
+            : memberName ?? string.Empty;
 
     private BoundExpression BindAccessorStep(
         BoundExpression receiver,
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs
@@ -436,13 +436,27 @@ internal sealed partial class ExpressionBinder
                 }
                 else if (receiver != null && receiver.Type is TupleTypeSymbol tupleSym)
                 {
-                    // Phase 4.5: tuple element access via Item1..ItemN.
+                    // Phase 4.5 / issue #2924: tuple element access via
+                    // one-based Item1..ItemN or zero-based .0...N selectors.
                     var memberName = ne.IdentifierToken.Text;
-                    if (memberName.StartsWith("Item", System.StringComparison.Ordinal)
-                        && int.TryParse(memberName.Substring(4), out var oneBased)
-                        && oneBased >= 1 && oneBased <= tupleSym.Arity)
+                    var zeroBased = -1;
+                    if (int.TryParse(memberName, out var numericIndex)
+                        && numericIndex >= 0
+                        && numericIndex < tupleSym.Arity)
                     {
-                        return new BoundTupleElementAccessExpression(null, receiver, tupleSym, oneBased - 1);
+                        zeroBased = numericIndex;
+                    }
+                    else if (memberName.StartsWith("Item", System.StringComparison.Ordinal)
+                        && int.TryParse(memberName.Substring(4), out var oneBased)
+                        && oneBased >= 1
+                        && oneBased <= tupleSym.Arity)
+                    {
+                        zeroBased = oneBased - 1;
+                    }
+
+                    if (zeroBased >= 0)
+                    {
+                        return new BoundTupleElementAccessExpression(null, receiver, tupleSym, zeroBased);
                     }
 
                     return BindExtensionMethodGroupOrError(receiver, ne);
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs
@@ -2062,8 +2062,10 @@ internal sealed partial class ExpressionBinder
             return receiver;
         }
 
-        var fieldName = syntax.FieldIdentifier.Text;
         var receiverType = receiver.Type;
+        var fieldName = receiverType is TupleTypeSymbol tupleType
+            ? GetTupleFieldName(syntax.FieldIdentifier.Text, tupleType)
+            : syntax.FieldIdentifier.Text;
         BoundExpression value = null;
         BoundExpression BindValue(TypeSymbol targetType) => value ??= BindAssignmentRhs(syntax.Value, targetType);
 
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Async.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Async.cs
@@ -212,6 +212,11 @@ internal sealed partial class ExpressionBinder
                 return new BoundErrorExpression(null);
             }
 
+            if (boundReceiver.Type is TupleTypeSymbol tupleType)
+            {
+                eventName = GetTupleFieldName(eventName, tupleType);
+            }
+
             // Check for user-defined event on a StructSymbol before falling through to CLR reflection.
             // ADR-0112 A5: TryGetEvent walks the base chain, so inherited instance
             // events on `open class` bases now resolve (parity with the bare-`this`
--- a/src/Core/CodeAnalysis/Binding/OverloadResolver.Invocations.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolver.Invocations.cs
@@ -128,7 +128,7 @@ internal sealed partial class OverloadResolver
     private BoundExpression BuildNullConditionalDelegateResult(
         CallExpressionSyntax syntax,
         BoundExpression receiverLoad,
-        LocalVariableSymbol capture,
+        VariableSymbol capture,
         BoundExpression whenNotNull,
         TypeSymbol returnType)
     {
@@ -685,9 +685,44 @@ internal sealed partial class OverloadResolver
             return callee;
         }
 
+        BoundNullConditionalAccessExpression nullConditionalCallee = null;
+        var assertions = new Stack<BoundUnaryExpression>();
+        var assertedCallee = callee;
+        while (assertedCallee is BoundUnaryExpression assertion
+            && assertion.Op.Kind == BoundUnaryOperatorKind.NullAssertion)
+        {
+            assertions.Push(assertion);
+            assertedCallee = assertion.Operand;
+        }
+
+        if (assertedCallee is BoundNullConditionalAccessExpression assertedNullConditional)
+        {
+            nullConditionalCallee = assertedNullConditional;
+            callee = assertedNullConditional.WhenNotNull;
+            while (assertions.TryPop(out var assertion))
+            {
+                var assertionOperator = BoundUnaryOperator.Bind(SyntaxKind.BangBangToken, callee.Type);
+                callee = new BoundUnaryExpression(
+                    assertion.Syntax,
+                    assertionOperator,
+                    callee,
+                    assertion.IsChecked);
+            }
+        }
+
         // The verbatim source spelling of the callee (`(value)`, `handler!!`, ...)
         // used in diagnostics — SyntaxNode.ToString() pretty-prints the whole tree.
         var calleeName = syntax.Callee.SyntaxTree.Text.ToString(syntax.Callee.Span);
+
+        BoundExpression CompleteInvocation(BoundExpression invocation) =>
+            nullConditionalCallee == null
+                ? invocation
+                : BuildNullConditionalDelegateResult(
+                    syntax,
+                    nullConditionalCallee.Receiver,
+                    nullConditionalCallee.Capture,
+                    invocation,
+                    invocation.Type);
 
         // Named args have no meaning without preserved parameter names.
         if (!TryAnalyzeCallArgumentLayout(syntax.Arguments, out _, out var argumentNames))
@@ -746,7 +781,7 @@ internal sealed partial class OverloadResolver
                 return new BoundErrorExpression(null);
             }
 
-            return new BoundIndirectCallExpression(syntax, callee, fnType, convertedArgs);
+            return CompleteInvocation(new BoundIndirectCallExpression(syntax, callee, fnType, convertedArgs));
         }
 
         if (callee.Type is DelegateTypeSymbol delegateSym)
@@ -762,12 +797,12 @@ internal sealed partial class OverloadResolver
                 return new BoundErrorExpression(null);
             }
 
-            return new BoundIndirectCallExpression(
+            return CompleteInvocation(new BoundIndirectCallExpression(
                 syntax,
                 callee,
                 delegateSym.EquivalentFunctionType,
                 convertedDelegateArgs,
-                delegateRefKinds);
+                delegateRefKinds));
         }
 
         // A value whose CLR type is a delegate (e.g. `Func[int32, int32]`) is
@@ -776,7 +811,7 @@ internal sealed partial class OverloadResolver
         {
             if (tryBindInheritedClrInstanceCall(callee, calleeClrType, "Invoke", boundArguments.ToImmutable(), syntax, out var invokeCall, null, default, argumentNames))
             {
-                return invokeCall;
+                return CompleteInvocation(invokeCall);
             }
         }
 
--- a/src/Core/CodeAnalysis/Syntax/Parser.Expressions.Creation.cs
+++ b/src/Core/CodeAnalysis/Syntax/Parser.Expressions.Creation.cs
@@ -583,7 +583,7 @@ public partial class Parser
         return false;
     }
 
-    private ExpressionSyntax ParseNameOrCallExpression()
+    private ExpressionSyntax ParseNameOrCallExpression(bool stopBeforeIndirectInvocation = false)
     {
         ExpressionSyntax current;
         if (Current.Kind == SyntaxKind.IdentifierToken
@@ -656,7 +656,7 @@ public partial class Parser
             current = ParseNameExpression();
         }
 
-        return ParsePostfixChain(current);
+        return ParsePostfixChain(current, stopBeforeIndirectInvocation);
     }
 
     // Phase 4.1 / ADR-0020: bounded-lookahead disambiguation between
--- a/src/Core/CodeAnalysis/Syntax/Parser.Expressions.cs
+++ b/src/Core/CodeAnalysis/Syntax/Parser.Expressions.cs
@@ -688,13 +688,13 @@ public partial class Parser
     // iterative, but an index argument re-enters the full expression grammar
     // (`a[a[a[…`), so the chain participates in the recursion cycle and ticks
     // the guard while an index (or accessor) argument parse is in flight.
-    private ExpressionSyntax ParsePostfixChain(ExpressionSyntax current)
+    private ExpressionSyntax ParsePostfixChain(ExpressionSyntax current, bool stopBeforeIndirectInvocation = false)
     {
         EnsureNestedParseAllowed();
         recursionDepth++;
         try
         {
-            return ParsePostfixChainCore(current);
+            return ParsePostfixChainCore(current, stopBeforeIndirectInvocation);
         }
         finally
         {
@@ -702,11 +702,30 @@ public partial class Parser
         }
     }
 
-    private ExpressionSyntax ParsePostfixChainCore(ExpressionSyntax current)
+    private ExpressionSyntax ParsePostfixChainCore(ExpressionSyntax current, bool stopBeforeIndirectInvocation)
     {
         while (true)
         {
-            if (Current.Kind == SyntaxKind.DotToken || Current.Kind == SyntaxKind.QuestionDotToken)
+            if (IsTupleElementSelectorToken(Current, includesDot: true))
+            {
+                // Issue #2924: the lexer keeps `.5` as a leading-dot float
+                // token. In postfix position the same token spells a zero-based
+                // tuple selector, so split it into the existing accessor shape.
+                var selectorToken = NextToken();
+                var dotToken = new SyntaxToken(syntaxTree, SyntaxKind.DotToken, selectorToken.Position, ".", null);
+                var identifierToken = new SyntaxToken(
+                    syntaxTree,
+                    SyntaxKind.IdentifierToken,
+                    selectorToken.Position + 1,
+                    selectorToken.Text.Substring(1),
+                    null);
+                current = new AccessorExpressionSyntax(
+                    syntaxTree,
+                    current,
+                    dotToken,
+                    new NameExpressionSyntax(syntaxTree, identifierToken));
+            }
+            else if (Current.Kind == SyntaxKind.DotToken || Current.Kind == SyntaxKind.QuestionDotToken)
             {
                 var dotToken = NextToken();
 
@@ -732,7 +751,7 @@ public partial class Parser
                 {
                     var rightSide = dotToken.Kind == SyntaxKind.QuestionDotToken
                         ? ParseNullConditionalContinuation()
-                        : ParseNameOrCallExpression();
+                        : ParseMemberContinuation();
                     current = new AccessorExpressionSyntax(syntaxTree, current, dotToken, rightSide);
                 }
             }
@@ -776,6 +795,11 @@ public partial class Parser
             else if (Current.Kind == SyntaxKind.OpenParenthesisToken
                 && !IsCurrentOnNewLineAfter(current))
             {
+                if (stopBeforeIndirectInvocation)
+                {
+                    break;
+                }
+
                 // Issue #2185: a same-line `(args)` following an arbitrary
                 // postfix expression is an *indirect* invocation of that
                 // expression's (function-typed) value — e.g. `(h)(value)`,
@@ -822,7 +846,7 @@ public partial class Parser
 
     private ExpressionSyntax ParseNullConditionalContinuation()
     {
-        var continuation = ParseNameOrCallExpression();
+        var continuation = ParseMemberContinuation();
         while (Current.Kind == SyntaxKind.BangBangToken)
         {
             var bangBangToken = NextToken();
@@ -831,6 +855,46 @@ public partial class Parser
         }
 
         return continuation;
+    }
+
+    private ExpressionSyntax ParseMemberContinuation()
+    {
+        if (!IsTupleElementSelectorToken(Current, includesDot: false))
+        {
+            return ParseNameOrCallExpression(stopBeforeIndirectInvocation: true);
+        }
+
+        var selectorToken = NextToken();
+        var identifierToken = new SyntaxToken(
+            syntaxTree,
+            SyntaxKind.IdentifierToken,
+            selectorToken.Position,
+            selectorToken.Text,
+            null);
+        return new NameExpressionSyntax(syntaxTree, identifierToken);
+    }
+
+    private static bool IsTupleElementSelectorToken(SyntaxToken token, bool includesDot)
+    {
+        var text = token.Text;
+        var digitStart = includesDot ? 1 : 0;
+        if (token.Kind != SyntaxKind.NumberToken
+            || string.IsNullOrEmpty(text)
+            || text.Length <= digitStart
+            || (includesDot ? text[0] != '.' : text[0] == '.'))
+        {
+            return false;
+        }
+
+        for (var i = digitStart; i < text.Length; i++)
+        {
+            if (text[i] < '0' || text[i] > '9')
+            {
+                return false;
+            }
+        }
+
+        return true;
     }
 
     // Issue #1016: parse the argument of an index expression, allowing a
--- a/src/Core/CodeAnalysis/Syntax/Parser.Expressions.cs
+++ b/src/Core/CodeAnalysis/Syntax/Parser.Expressions.cs
@@ -763,7 +763,7 @@ public partial class Parser
                 // emitter reuse the existing `(*p).field` / `(*p).method(...)`
                 // bound shape without any new bound-node kinds.
                 var arrowToken = NextToken();
-                var rightSide = ParseNameOrCallExpression();
+                var rightSide = ParseNameOrCallExpression(stopBeforeIndirectInvocation: true);
                 var starToken = new SyntaxToken(syntaxTree, SyntaxKind.StarToken, arrowToken.Position, "*", null);
                 var deref = new UnaryExpressionSyntax(syntaxTree, starToken, current);
                 current = new AccessorExpressionSyntax(syntaxTree, deref, arrowToken, rightSide);
@@ -851,7 +851,7 @@ public partial class Parser
         {
             var bangBangToken = NextToken();
             continuation = new UnaryExpressionSyntax(syntaxTree, bangBangToken, continuation);
-            continuation = ParsePostfixChain(continuation);
+            continuation = ParsePostfixChain(continuation, stopBeforeIndirectInvocation: true);
         }
 
         return continuation;
@@ -880,7 +880,6 @@ public partial class Parser
         var digitStart = includesDot ? 1 : 0;
         if (token.Kind != SyntaxKind.NumberToken
             || string.IsNullOrEmpty(text)
-            || text.Length <= digitStart
             || (includesDot ? text[0] != '.' : text[0] == '.'))
         {
             return false;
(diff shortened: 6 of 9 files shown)


DavidObando:
Coordination note from #2935 review: please add an interpreter pin for a real G# lambda invoked through numeric tuple selection, e.g. `let t = ((value int32) -> value + 1, 0); t.0(41)`. This combined shape needs both PRs; #2935 is currently likely to land first, and neither branch alone can own a passing test.

DavidObando:
## Review — PR #2932 (`4e0b32d1`), issue #2924

**VERDICT: NOT MERGEABLE.** One BLOCKER. Fix small. Rest of PR good — better than author claim in some places, worse in one.

Worktree `/Users/davidobando/GitHub/DavidObando/gsharp-wt/i2924`, HEAD `4e0b32d1`, base `d4038f1d`. Tree clean at finish, `git status --porcelain` empty. Shared nuget cache untouched (mtime Jul 29). No e2e run.

---

## BLOCKER

### B1 — `?.` no longer short-circuit when call get lifted
`src/Core/CodeAnalysis/Syntax/Parser.Expressions.cs:795-800` (the `stopBeforeIndirectInvocation` break) + `src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.Accessor.cs:1223-1290` (`BindNullConditionalAccessExpressionCore`).

Lift move invocation **outside** accessor. Binder wrap only accessor subtree in nil guard. So `(args)` run even when receiver nil. Null-conditional contract broken.

Fresh probes, compiled, real `dotnet <app>.dll` execute (not `Assembly.Load`, not `IlVerifier`):

| probe | main | HEAD |
|---|---|---|
| `g?.Plain(1)`, g nil | ok | ok ("A ok") |
| `g?.Make(40)(2)`, g nil | GS0159 compile error | **NullReferenceException** |
| `g?.Get().0(1)`, g nil | ran, printed start+end | **NullReferenceException** |
| `t?.0(1)`, nil nullable tuple | (silent elision, #2924) | **NullReferenceException** |
| `g?.Arr()[0]`, g nil | ok | ok — **index continuation still short-circuits** |

Last row is the tell. `?.` + index correct, `?.` + lifted call wrong. PR make language internally inconsistent.

Interpreter worse. Bound-tree render leak the bug:
```
B  g?.Make(40)(2), g NON-nil : GS9999 Object type System.Func`2[Int32,Int32] does not match target type ClosureValue
                               g?.$ncap_1.Make(40).Invoke(2)
C  g?.Make(40)(2), g nil     : GS9999 Non-static method requires a target.
D  t?.0(1), nil tuple        : GS9999 Non-static method requires a target.
A  g.Make(40)(2)  (no ?.)    : 42   <-- correct
```
So interpreter cannot run **any** `?.` + lifted call, nil or not. Zero tests cover this.

PR own scope: it added `?.`+numeric-selector support and shipped `t?.0` tests (`test/Core.Tests/.../Issue2924TupleElementDelegateCallTests.cs:58`, `test/Interpreter.Tests/Issue2924TupleElementDelegateCallTests.cs:34`) — but only non-nil, non-call form. Nil + call is the hole.

**Fix options** (author choose):
1. In invocation binder: if bound callee is `BoundNullConditionalAccessExpression`, re-root invocation into its `WhenNotNull`. ~10 lines. Costs the PR its "invocation-binding files untouched" claim.
2. Do not lift when `current` root is null-conditional, **and emit a diagnostic**. Plain no-lift alone is wrong — that re-introduce silent elision for `t?.0(1)`.
3. Minimum acceptable if deferring: block the shape with a diagnostic, file follow-up issue.
Either way add pins: nil `g?.M(x)(y)`, nil `t?.0(1)`, and non-nil interpreted `g?.M(x)(y)`.

---

## SHOULD-FIX

### S1 — numeric selector read-only; `ItemN` parity claim false
`src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs` (~2152-2345, `ReportUnableToFindMember(syntax.FieldIdentifier…)`) has own member resolution; not updated.

```
var t = (1,2); t.0 = 5      -> GS0158 Cannot find member 0.
var t = (1,2); t.0 += 1     -> GS0158 Cannot find member 0.
var t = (1,2); t.Item1 = 5  -> 5   (resolves real ValueTuple.Item1 field)
```
Same both compiled and interpreted. PR advertise `.N` == `ItemN`. In write position it not. Either wire the numeric branch into assignment lookup (small) or emit a targeted diagnostic ("use `.Item1` to assign") instead of the misleading "Cannot find member 0".

### S2 — interpreter suite never exercise the actual bug
`#2928` block materializing tuple that hold a live function value:
```
let handler System.Action[int32] = (v int32) -> Console.WriteLine(v)
let t = (handler, 0)   -> GS9999 Constructor on type 'System.ValueTuple`2[[System.Action`1[…' not found.
```
So every *passing* interpreter test in this PR is int tuples or parse shapes. **No interpreter test prove a tuple-held delegate actually gets invoked.** "Interpreter parity" unproven for the feature itself. Say so in PR body, or gate the claim behind #2928.

Correction to my earlier read and to the briefing hint: `NilTupleElementDelegate_ReportsRuntimeFailure` (`test/Interpreter.Tests/…:109`) *does* pass for the right reason — it yield `GS9999: Non-static method requires a target.` on line `t.Item1.Invoke(1)`, which is genuine nil-delegate invoke. But `Assert.Contains("error GS", output)` would equally accept the unrelated #2928 constructor failure. Tighten to the specific message. (NIT-grade assertion, SHOULD-FIX-grade coverage gap.)

### S3 — fix incomplete for two sibling continuations (inconsistency, not regression)
- `Parser.Expressions.cs:765` — pointer arrow `->` still call `ParseNameOrCallExpression()` with no flag. Probe `q[0]->m(x)(y)` on HEAD → `Accessor(*q[0], ->, Call(Call(m,x),y))` — the **old** nested shape, same shape that give GS0159 for `.`.
- `Parser.Expressions.cs:849-856` — `ParseNullConditionalContinuation` `!!` loop call `ParsePostfixChain(continuation)` with no flag. Probe `a?.b!!(x)(y)` on HEAD → `Accessor(a, ?., Call(Call(Unary!!(b),x),y))` — old shape.
- Contrast `a.b!!(x)(y)` on HEAD → `Call(Call(Unary!!(Accessor(a,.,b)),x),y)` — correctly reshaped.

Not a regression (main same), but PR now say "receiver-wide call shape" and two paths don't get it. Either extend or note explicitly as out of scope.

---

## NIT

- **N1** `Parser.Expressions.cs:704` — new selector branch has **no** `IsCurrentOnNewLineAfter` guard, unlike the `(`-invocation branch at `:794`. `expr \n .5` now become member access where main saw two statements. Consistent with existing `.member` continuation (also unguarded), so defensible — but author guard test `ExponentFloatAfterExpression_RemainsSeparateLiteral` only cover non-all-digit `.5e1`, so the all-digit newline case is untested.
- **N2** `t.00` → element 0, `t.007` → index 7. Leading zeros accepted. Rust reject this. Laxness only.
- **N3** Two dead guards: `Parser.Expressions.cs:880` `text.Length <= digitStart` and `ExpressionBinder.Access.MemberLookup.cs:437` `numericIndex >= 0` — both unreachable given the parser's all-digit guarantee. Mutants M10/M15 survived because they are equivalent mutants. Ponytail would delete them; harmless if kept.
- **N4** `int.TryParse` default `NumberStyles`/culture at `MemberLookup.cs:436`. `"+1"`/`" 1"` **unreachable** — parser guarantee all-ASCII-digit text, and no identifier can start with digit. Also matches the pre-existing `ItemN` house pattern in the same method. Briefing concern does not land.
- **N5** Semantic tokens colour `.0` as a **Number literal**, not a member. `src/LanguageServer/SemanticTokensHandler.cs:86` use `SyntaxTree.ParseTokens` (raw lexer), which still see one `NumberToken '.0'`. Cosmetic; needs a handler carve-out if you want member colouring.
- **N6** `FormattingEngine` (`src/LanguageServer/FormattingEngine.cs:39`, also `ParseTokens`) formats `Console.WriteLine(t.0)` → `Console.WriteLine (t .0)`. Still parses (`t .0` verified working), so semantics preserved, but the new syntax get spaced apart. Pre-existing formatter rule, unchanged by PR.

---

## Corrections to your briefing

1. **Q2 premise wrong. `SyntaxToken` has NO trivia.** ctor is `(syntaxTree, kind, position, text, value)`. The `null` the author pass is **Value**, not trivia. No `ToFullString`, no round-trip test facility; `SyntaxNode.ToString()` is a pretty-printer, not source. Nothing is lost.
2. **Positions exact.** LSP probe on HEAD: `Console.WriteLine(t.0)` → `DotToken '.' span=[36,37)`, `IdentifierToken '0' span=[37,38)`. `SemanticLookup.FindTokenAt` resolve the synthetic identifier cleanly, same as `Item1` at `[37,42)`. Mutant M3 (position off-by-one) was killed by a span-asserting Core test.
3. **Fabricating tokens IS the house pattern** — `Parser.Expressions.cs:765-767` already synthesize a `StarToken` to desugar `p->m`; also `Parser.Statements.cs:311-410,1139,1298`, `Parser.Declarations.cs:463-1131`, `Parser.Expressions.Creation.cs:1112`. No existing split/re-lex facility exists that was ignored. **No ponytail finding here.**
4. **Q4 `else if` ordering concern is moot** — "all digits" and "starts with `Item`" are mutually exclusive. Neither can shadow the other.
5. **Your Compiler.Tests baseline 3734 is wrong, and the other agent's 3721 is wrong.** True `origin/main` = **3708**. See Q6 below.

---

## Q1 — parse-tree grid, main vs HEAD

30 cases dumped. **16 identical, 14 changed. 10 of the changed are non-tuple shapes.** Yes — the reshape **is** global, your read is right.

Canonical delta:
```
a.b(x)(y)
  main : Accessor(a, ., Call(Call(b,x), y))
  HEAD : Call(Accessor(a, ., Call(b,x)), y)
```

| case | shape changed? | main runtime | HEAD runtime |
|---|---|---|---|
| `a.b(x)` | no | ok | ok |
| `a?.b(x)` | no | ok | ok |
| `a.b[int32](x)` | no | ok | ok |
| `a.b.c(x)` | no | ok | ok |
| `a.b().c()` | no | ok | ok |
| `a.b()[0]` | no | ok | ok |
| `a.b!!(x)` | no | ok | ok |
| `a?.b!!(x)`, `a?.b!!(x)(y)` | no | — | unchanged (S3) |
| ext-method thru `.` | no | ok | ok |
| `(` on next line | no | 2 statements | 2 statements |
| `a.b(x)(y)` | **YES** | **GS0159** | works |
| `a.b(x)(y)(z)` | **YES** | **GS0159** | works |
| `a?.b(x)(y)` | **YES** | **GS0159** | works non-nil; **B1 on nil** |
| `a.b[int32](x)(y)` | **YES** | **GS0159** | works |
| `a.b()[0](y)` | **YES** | **GS0159** | works |
| `a.b(x).c(y)(z)` | **YES** | **GS0159** | works |
| `a.b(x)(y).c` | **YES** | **GS0159** | works |
| `a.b(x)(y)[0]` | **YES** | **GS0159** | works |
| `Foo().M()(x)` | **YES** | **GS0159** | works |
| `holder.Value.0(5)` etc (tuple) | **YES** (new) | elided | works |

**Explicit statement you asked for: yes, 10 non-tuple shapes changed. Every one of them was a hard compile error `GS0159 "Cannot find function ."` on `main`.** So regression surface for previously-valid programs is essentially zero — no working program can have used those shapes. The reshape is a strict widening, not a silent re-interpretation. The one exception is B1: `a?.…(y)` moved from compile-error to *wrong runtime behaviour*.

Newline rule unchanged in both stop and non-stop paths (guard at `:794` untouched, sits before the `stopBeforeIndirectInvocation` break at `:796`).

---

## Q3 — lexical boundary table (observed, both sides)

| input | behaviour | note |
|---|---|---|
| `1.5`, `1.5f`, `.5f`, `1e5`, `1.5e3` | unaffected | non-all-digit body rejected by `IsTupleElementSelectorToken` |
| `0x1F`, `0b1010`, `1_000` | unaffected | |
| `.5 + .25` in primary position | `0.75` | author guard test |
| `0...3`, `x...y`, `a[0..2]`, `a[..2]`, `a[0..]` | safe | `Lexer.cs:270-296` match `...`/`..` **before** the ADR-0044 leading-dot float |
| `t.0.1` | works — `((10,20),30).0.1` = `20` | lexes `t` `.0` `.1` |
| `t.00` | element **0** | N2 |
| `t.007` on arity-2 | GS0158 "Cannot find member 007" | sensible |
| `t.0000000000000000000000` | element 0 | parses to 0, not overflow |
| `t.99999999999999999999` | GS0158 "Cannot find member 99999…" | `int.TryParse` fail → clean fallthrough |
| `t.10` arity 3 | GS0158 | sensible, no `BindExtensionMethodGroupOrError` confusion |
| `t.-1` | GS0005 "Unexpected token \<MinusToken>, expected \<IdentifierToken>" | good error |
| `someInt.0`, `"ab".0` | GS0158 "Cannot find member 0" | no crash |
| `t. 0` / `t .0` / `t . 0` | **all three work** | `t. 0` via `ParseMemberContinuation`, `t .0`/`t . 0` via the split branch |
| `"${t.0}"` | works | |
| `t.0f`, `t.0e1`, `t.0_1` | **not** selectors → statement split, original elision persist | nonsense spellings, NIT |

---

## Q4 — binder

- `t.0` and `t.Item1` produce **byte-identical bound trees** (printer even render `.0` back as `.Item1`), therefore identical IL. Verified by bound-tree dump diff.
- Named tuple element cannot be `0` and no CLR member can be named all-digits — G# identifiers cannot start with a digit. Branch shadow nothing.
- `else if` ordering harmless (see correction 4).
- Read path parity: **yes**. Write path parity: **no** — see S1.

---

## Q5 — smaller fix? **No. Author's mechanism is right-sized.**

**(c) definitive answer: `a.b(x)(y)` was BROKEN on `main`.** Probe: compile under `gsc-main` → `GS0159: Cannot find function .`. Triples, `a?.b(x)(y)`, `a.b[T](x)(y)`, `a.b()[0](y)`, `Foo().M()(x)` — all GS0159 on main. So `stopBeforeIndirectInvocation` is a **genuine second bug fix**, not a risk-only refactor. It deserves its own issue number and a line in the PR body; right now it is riding along unannounced. My mutant M5 (flip the flag to `false`) was killed by Core+Interp+Compiler tests, so it does have real pins, not only guards — the briefing's worry does not land.

**(a) lexer-level fix — rejected, bigger blast radius.** `Lexer.cs` carries **no previous-token state** (grep for previous/last token: nothing) — would need new lexer state. It would change `SyntaxTree.ParseTokens` output, which `FormattingEngine.cs:39` and `SemanticTokensHandler.cs:86` both consume. And it would **still** need a parser numeric member-continuation path for `t. 0`. Fewer lines, more risk. Not a ponytail win.

**(b) `ParsePostfixChainCore`-only, no plumbing — does not work.** `holder.Value.0(5)` and `factory.Make()(41)` need the lift; main's nested `Accessor(a, ., Call(Call(b,x),y))` is exactly what the accessor binder reject with GS0159. Making the outer loop fully left-associative would reshape `a.b.c` too — much larger.

**I could not construct a strictly smaller equivalent fix.** Ninth-duck streak breaks here. Only defect in the mechanism is B1.

---

## Q6 — my own verification

### True `origin/main` baselines (measured directly: full `src/` revert + delete the 3 new test files, then restore)

| suite | **origin/main (mine)** | HEAD (mine) | delta | new tests in PR | author claim |
|---|---|---|---|---|---|
| Core.Tests | **6913** passed, 1 skipped | 6925 passed, 1 skipped | +12 | 12 | 6925 ✓ |
| Compiler.Tests | **3708** total (3689 pass, **19 environmental fail**) | **3713**, 0 failed, 1 h 6 m | +5 | 5 | 3713 ✓ |
| Interpreter.Tests | **453** | **464**, 0 failed | +11 | 11 | 464 ✓ |

Arithmetic checks exactly. **Your 3734 and the other agent's 3721 are both wrong; use 3708.** The 19 `Compiler.Tests` failures on baseline are environmental — `Gsharp.Extensions.dll not found in any expected location` — they vanish once `Interpreter.Tests` is built first (that build produces `Gsharp.Extensions.dll`). Exactly the trap you warned about; I ran a clean control and did not count them red. Author's numbers are all honest.

### Mutation table — per-hunk, build-green, semantically inert only

Unmutated control in same worktree before the batch: Core 12/12, Interp 11/11, Compiler 5/5 pass.

| id | file:line | mutation | Core/Interp/Compiler fails | verdict |
|---|---|---|---|---|
| M1 | Parser.Expressions.cs:707 | `includesDot: true` → `false` (split never fires) | 5/7/4 | killed |
| M2 | :718 | `Text.Substring(1)` → `Text` | 4/5/3 | killed |
| M3 | :716 | `Position + 1` → `Position` | 1/0/0 | killed (span assert) |
| M4 | :862 | continuation `includesDot: false` → `true` | 2/1/0 | killed |
| M5 | :864 | `stopBeforeIndirectInvocation: true` → `false` | 2/1/1 | killed — **real pin for the stop flag** |
| M6 | :691 | `ParsePostfixChain` default `false` → `true` | 1/0/1 | killed |
| M7 | Creation.cs:586 | `ParseNameOrCallExpression` default `false` → `true` | 4/3/3 | killed |
| M8 | Creation.cs:659 | drop forwarding (pass `false`) | 2/1/1 | killed |
| M9 | :890 | digit loop bound `'9'` → `'z'` | 1/0/0 | killed |
| M10 | :880 | `text.Length <= digitStart` → `<` | 0/0/0 | **survived — equivalent (dead guard)** |
| M11 | :849 | `?.` continuation → `ParseNameOrCallExpression()` | 1/1/0 | killed |
| M12 | MemberLookup.cs:438 | `numericIndex < Arity` → `< 0` | 4/6/3 | killed |
| M13 | MemberLookup.cs:438 | `<` → `<=` | 0/1/1 | killed |
| M14 | MemberLookup.cs:450 | `oneBased - 1` → `oneBased` | 1/–/1 | killed |
| M15 | MemberLookup.cs:437 | `numericIndex >= 0` → `>= -1` | 0/0/0 | **survived — equivalent (dead guard)** |

**13/15 killed; 2 equivalent mutants.** I cannot reproduce the author's "15/15", but every **reachable** branch is killed. Two survivors are the dead guards of N3. Author should say 13/15 + 2 equivalent, not 15/15.

### Pins vs guards, counted separately (pin = fail on `main` src + PR tests; guard = pass both)

| suite | pins | guards | total |
|---|---|---|---|
| Core.Tests | 7 | 5 | 12 |
| Interpreter.Tests | 9 | 2 | 11 |
| Compiler.Tests | 4 | 1 | 5 |
| **total** | **20** | **8** | **28** |

**Author's 20/8 confirmed exactly.**

Missing pins (all B1/S1 territory): nil-receiver `g?.M(x)(y)`, nil `t?.0(1)`, non-nil interpreted `g?.M(x)(y)`, `t.0 = 5`, `t.0 += 1`.

### Corpus
Re-ran over the 122 `samples/*.gs` with both `gsc-main` and `gsc-head`: **122/122 identical diagnostics + exit status; 119/119 emitted binaries byte-identical.** (Author says 148/139 — different input set, same conclusion.) Only 4 samples touch `ItemN`, 2 touch `?.`, **zero** use a numeric selector or the curried shape.

**You are right and it should be stated plainly in the PR: the corpus proves nothing about the feature. Its entire value is Q1 non-regression evidence for the global parser reshape.** For that purpose it is good evidence, and combined with "every changed shape was GS0159 on main" it is fairly conclusive.

---

## Q7 — does it fix #2924?

- **Compiled: YES.** Real end-to-end: `gsc` → `.dll` → `dotnet <app>.dll` executed. `t.0(1)` invokes; `holder.Value.0(5)`, `factory.Make()(41)` work; curried and nested selectors work. Stronger than `Assembly.Load`+`GetTypes()`; I did not rely on `IlVerifier.Verify` for any conclusion.
- **Interpreted: PARTIAL.** Parse/bind parity yes; actual delegate-through-tuple invocation **cannot be demonstrated** because #2928 blocks tuple materialization (S2).
- **LSP: no crash.** Hover token resolution finds the synthetic `IdentifierToken '0'` at correct span. `LanguageServer.Tests` was never run by the author and I did not run the full suite — cheap gap, worth one CI run. Semantic tokens/formatter caveats at N5/N6.

---

## Q8 — merge order

- `git merge-tree --write-tree HEAD <pr-head>` against **#2908, #2911, #2912, #2917, #2923, #2926, #2929**: **all exit 0, all clean, zero conflicts.**
- Only #2908 and #2912 share `ExpressionBinder.Access.MemberLookup.cs`; their hunks (24-34, 91-105, 743) are far from this PR's 432-455.
- **No open PR touches either parser file.**
- **#2928 independent.** This PR does not touch `src/Core/CodeAnalysis/Evaluator.Expressions.cs`. Neither breaks the other. #2928's fix will *unblock* the interpreter coverage this PR lacks; `NilTupleElementDelegate_ReportsRuntimeFailure` already fails for the right reason (`Non-static method requires a target.`) and should survive #2928 landing.
- **#2925** ("Delegate-typed interface property cannot be invoked", `ExpressionBinder.Calls.Invocation.cs`): this PR **neither helps nor hinders** in its current form — the direct `box.P(1)` shape is consumed by `ParseNameOrCallExpression`'s bare-identifier dispatch and is untouched here. **But:** fixing B1 option (1) *does* touch `ExpressionBinder.Calls.Invocation.cs`, which is also touched by #2908 and #2929, and is #2925's home. If the author takes option (1), merge order suddenly matters. Worth flagging to the queue owner.

---

## Summary of what author should do

1. Fix B1 (blocking) + add the 3 nil/`?.` pins.
2. S1 — write-path parity or a targeted diagnostic.
3. S2 — soften the interpreter-parity claim; tighten the weak `Contains("error GS")` assert.
4. S3 — extend or explicitly scope out `->` and `?.`+`!!`.
5. PR body corrections: mutants are **13/15 + 2 equivalent**, not 15/15; corpus value is Q1 regression evidence only; `a.b(x)(y)` fix is a **second, separately issue-worthy bug fix** (was GS0159 on main) and should be named.

Everything else in the author's report checked out: 20 pins / 8 guards exact, all three suite totals exact, invocation-binding files genuinely untouched, no merge conflicts, bound-tree/IL parity for `.N` vs `ItemN` real.

DavidObando:
## Round 3 review — MERGEABLE (no code blockers)

Reviewed **`0f543981`** (briefing SHA) and **`c6be1ba6`** (real branch tip, appeared mid-review). Blocker B1 resolved and verified by execution.

### Process blocker — push required
Remote head `0f543981` conflicts with landed **#2908** in `ExpressionBinder.Access.MemberLookup.cs:24-30`; `mergeStateStatus` is **DIRTY**, not CLEAN. The author has the correct rebase locally at `c6be1ba6` but has **not pushed**. The rebase is honest:

```
git diff e21677f6 0f543981 -- src/  →  347 patch lines
git diff 914b42a7 c6be1ba6 -- src/  →  347 patch lines
diff of the two ±line sets          →  IDENTICAL
```

Rebase changed **zero** PR production bytes, so this entire review transfers to `c6be1ba6`. Fix is `git push --force-with-lease` + re-green CI.

### §1 — The five-file spread is not all B1
Per-commit diffstat of `9efd679f` shows the briefing mis-attributed it:

| file:line | belongs to |
|---|---|
| `OverloadResolver.Invocations.cs:689-712,775,791-796,805` | **B1 proper** — unwrap `?.` callee, rebind vs `WhenNotNull`, re-wrap |
| `ExpressionBinder.Access.MemberLookup.cs:27-55` | B1 support (helper extraction) |
| `ExpressionBinder.Assignments.cs:2065-2067` | round-2 **S1** write-path parity |
| `ExpressionBinder.Async.cs:215-218` | round-2 **S1** — and this file is the **compound-assign / event-subscribe** binder (`x.M += h`), not async-only |
| `Parser.Expressions.cs:766,854` | round-2 **S3** |
| `Parser.Expressions.Creation.cs:586,659` | sixth file, omitted from the briefing table |

**Blast radius measured, not read.** The +45 lives entirely inside `BindIndirectCallExpression` (#2185 path), which only runs when the callee is neither a bare name nor a plain `.member`. Across the 148-sample corpus, `gsc-main3` vs `gsc-head3`: **148/148** identical diagnostics and exit codes, **148/148** identical SHA-256 (139 binaries + 9 no-binary both sides), **0 BYTEDIFF, 0 new compile failures**.

**Corpus zero:** numeric tuple selector **0/148**; delegate held in a tuple element **0/148** (2 regex hits are comment text). The corpus proves nothing about the feature and everything about non-regression from the receiver-wide parser reshape — which is where the risk actually lives.

### §2 — Short-circuit verified by execution against a known-bad control
Harness compiles to a real `.dll` + hand-written `runtimeconfig.json` and executes in a child process. No `IlVerifier`. **Control validated**: the round-2 compiler `4e0b32d1` gives `EXIT=134` + `NullReferenceException` on `p40/p41/p43`; the same programs under `gsc-head3` give `EXIT=0` and correct output — the harness sees the exact B1 failure mode.

| probe | result |
|---|---|
| `f?.Make(40)(2)`, f nil / non-nil | `nil` no NRE / `42` ✓ |
| `Side()?.Make(40)(2)` | `SIDE-EVAL` printed **exactly once** on both paths ✓ |
| nil path argument evaluation | `ARG-1`/`ARG-2` **absent** ✓ |
| triple curried, generic `Gen[int32](5)(7)`, indexed `Arr()[0](4)` | all short-circuit ✓ |
| `full?.0(41)` / `empty?.0(41)` | 42 / nil ✓ |
| nil element inside non-nil tuple | NRE — correct, `?.` guards receiver not element ✓ |
| `a?.B(c?.D() ?? 7)(5)` nested | 25 / 75 / nil / 219, `D` skipped when inner nil ✓ |

All re-run at `c6be1ba6` — byte-identical. #2908 landing changes nothing here.

### §3 — #2944's pins are real: 7/7 fail on `main`
Callee-span probe run under `origin/main` src then HEAD src:

`factory.Make(40)(2)` · `Make[int32](40)(2)` · `Make(40)[0]` · `Make(1)(2)(3)` · `q[0]->m(x)(y)` · `a?.b!!(x)(y)` · `holder.Value.0(5)` — **all FAIL on main, all PASS at HEAD**. The 4-shape loop packed into one `[Fact]` is not masking a passing member; each was checked separately.

### §4 — Interpreter parity is genuinely end-to-end
`RunSubmission` calls `EvaluateSubmission` with `Console.SetOut` capture and asserts on the **printed value**, not binding success. Independent REPL probe against live `gsi.dll`: the #2924 repro `functions.0(41)` → **42** interpreted; `?.` nil → no NRE; nested `t.0.1` → 20; write path `t.0 = 5` → 5; `Side()?.0(41)` → `SIDE` once, 42. Compiled gives `42/42/42/True/20`. **Compiled ≡ interpreted** — the property #2924 is actually about.

### §5 — #2932 is a *precondition* for #2908, not merely compatible
At `c6be1ba6`, `:698-701` reassigns `callee = assertedNullConditional.WhenNotNull` and `:768` then calls #2908's `TryReportNullableDelegateReceiver(callee.Type, …)`. The unwrap runs **first**, so the guard sees the **narrowed** type and correctly does not fire on `t?.0(41)`. Without the unwrap, #2908's guard would false-positive on every `?.`-lifted curried call. `Issue2924` + `Issue2885` together: **52/52 pass, 0 fail**.

`merge-tree` vs #2911, #2912, #2926, #2929, #2950 — all **CLEAN**.

### §6 — Deleted guards challenged, both defensible
- **Guard A** (`Parser.Expressions.cs:880`, `text.Length <= digitStart`): swept `ParseTokens` over 66 adversarial strings + **6859** exhaustive 3-char strings + all 148 corpus files → zero `NumberToken` with text `"."` or `""`. Confirmed structurally at `Lexer.cs:265-296`. **Safe.**
- **Guard B** (`numericIndex >= 0`): characterized the whole input domain by reflection. `"-1"` and `"-2147483648"` **are accepted** and yield negative indices; `"+1"`, `" 1"`, `"01"` also accepted. But no G# identifier token can spell `-`, `+`, or whitespace, so every lax row is unreachable from source. Genuinely dead — **NIT only**, because call sites grew 1→3 while the guard was removed.

### §7 — Metrics: author under-claimed
Measured by src-swap, not by the author's table. PIN run (`git checkout origin/main -- src/`, PR tests kept):

| suite | pins (fail on main) | guards |
|---|---|---|
| Core | 15 | 5 |
| Interpreter | 15 | 2 |
| Compiler | 6 | 1 |
| **total** | **36** | **8** |

Author claimed 20/8. **Mutations 6/6 killed** — `MUT-ASYNC` 1/1/1, `MUT-ASSIGN` 1/1/1, `MUT-OR` 4/4/1, `MUT-OR2a` 1/1/1, `MUT-PTR` 1/0/0, `MUT-BANG` 2/1/1. One attempt discarded for CS0136 shadowing and reformulated — a compile error is not detection.

**Suite baselines (executed, Failed 0):**

| SHA | Compiler | Interpreter |
|---|---|---|
| `e21677f6` | **3756** | 468 |
| **`914b42a7`** | **3802** | 470 |
| `c6be1ba6` | 3809 | **487** = 470+17 ✓ |

#2908 adds exactly **46** Compiler tests. PR adds 3 new files, **zero existing test files modified**.

Merge commit `93a0cbff` clobbered nothing — `Evaluator.*` show zero diff vs main, so #2935's interpreter work is fully intact. Now moot; replaced by the rebase.

### NITs
1. `ExpressionBinder.Access.MemberLookup.cs:27-49` — re-add `numericIndex >= 0`. One token, no behaviour change, and call sites tripled this round.
2. `ExpressionBinder.Access.MemberLookup.cs:~34` — `memberName.StartsWith("Item", …)` derefs the null `Text` of a missing `IdentifierToken`. `t. = 5` → `GS9998`. **Pre-existing** (main ICEs too, with a different exception), but one `string.IsNullOrEmpty` early-out in this PR's own new helper closes a long-standing ICE for free.
3. `git status --porcelain` shows `?? out/` — untracked corpus scratch. `out/bin` is gitignored, bare `out/` is not.

### Retracted by the reviewer
- `$nres_` local-name collision under nested `?.` in arguments — **falsified by execution** (`nres.gs`: 25 / 75 / nil / 219 all correct).
- `t. = 5` NRE as a new blocker — **pre-existing** ICE on `origin/main`.

### Gotcha for future probes
A probe csproj with `<Reference HintPath>` copies `GSharp.Core.dll` into its own `bin/`. Must `rm -rf bin obj` after every Core rebuild or the probe silently reports stale results.

PR #2935, opened 2026-07-31
Fix interpreter callable materialization

### LINKED ISSUE #2928: Interpreter cannot materialize tuples containing function values

DavidObando:
## Summary

Fixes #2928.

Interpreter function literals and user method groups were represented by `ClosureValue` instances (`src/Core/CodeAnalysis/Evaluator.Expressions.cs:754-781`). CLR consumers such as the `ValueTuple<Func<int, int>, int>` constructor received that interpreter-only object unchanged at `Evaluator.Expressions.cs:704-717`, so reflection could not bind the constructor. The evaluator now materializes closures as delegates of their exact CLR type at the producer boundary, invokes both materialized delegates and legacy closure fallbacks, adapts structural delegates to named CLR delegate types, and wraps async results correctly.

Direct indirect-call invocation unwraps `TargetInvocationException`. Reflection member calls such as tuple/list element `.Invoke` still have a pre-existing uncaught-diagnostic leak, tracked separately in #2938.

## Production diff

- `src/Core/CodeAnalysis/Evaluator.Expressions.cs`: +86/-5
- `src/Core/CodeAnalysis/Evaluator.Calls.cs`: +8/-0
- Total: +94/-5, 99 changed production lines

No binder, parser, compiler-emitter, or non-evaluator production files changed.

## Pins (fail on `origin/main`, pass on this branch)

| Pin | Main | HEAD |
|---|---:|---:|
| Tuple containing `Func<int, int>` | fail | pass |
| CLR `Predicate<int>` method argument | fail | pass |
| `List<(int32) -> int32>` | fail | pass |
| delegate array | fail | pass |
| delegate-valued `Dictionary` | fail | pass |
| delegate-valued G# map | fail | pass |
| `Lazy<int32>` delegate constructor argument | fail | pass |
| `Action<int32>` with captured local | fail | pass |
| user method group stored in CLR collection | fail | pass |
| interpreter/compiled tuple-output parity | fail | pass |
| directly awaited async lambda in a function-typed slot | fail | pass |

## Guards (pass on both `origin/main` and HEAD)

| Guard | Main | HEAD |
|---|---:|---:|
| Interpreter class field callable invocation | pass | pass |
| Direct callable invocation | pass | pass |
| Callable exception reaches typed interpreter catch | pass | pass |
| Direct uncaught callable exception keeps original diagnostic | pass | pass |
| Compiled tuple function executes through real assembly load/invocation | pass | pass |
| Compile-and-run harness rejects known-bad program | pass | pass |

`Callable_DelegateFactoryIsReusedPerType` is an additional cache regression test. It cannot run on `main` because `main` fails to construct its tuple; on this branch it proves two closures of the same delegate type share one compiled bridge method while retaining distinct behavior.

`Callable_AsyncDirectInvocation_ReturnsAwaitableTask` is a pin: it fails on `origin/main` and when the evaluator-owned delegate shortcut is mutated to return raw `InvokeClosure` output (`'await' operand did not evaluate to a Task.`), and passes on HEAD.

## Semantic mutation checks

Each mutation was build-green. An unmutated control ran first; each mutation was restored and rebuilt before the next.

| Branch | Neutral mutation | Detection |
|---|---|---|
| Function literal materialization | return raw `ClosureValue` | literal CLR-consumer pins and parity fail |
| User method-group materialization | return raw `ClosureValue` | method-group pin fails |
| Closure materialization helper | return closure unchanged | CLR-consumer pins fail |
| Named delegate adaptation | return source delegate unchanged | `Predicate<int>` pin fails |
| Materialized delegate invocation | force delegate branch false | callable invocation pins fail |
| Non-void expression bridge | return default expression instead of invocation result | value-returning pins fail |
| Void expression bridge | omit invocation | `Action<int32>` pin fails |
| Argument forwarding | pass empty argument array | parameterized callable pins fail |
| Captured-local forwarding | skip captured-local copy | capture pin fails |
| Reflection exception unwrapping | rethrow `TargetInvocationException` | direct exception guards fail |
| Async result wrapping | return raw interpreter result | `AsyncLambda_Inference_WrapsReturnInTask`, `TrailingLambda_AsyncSoleArgument_BindsWithoutDiagnostics`, and `TrailingLambda_AsyncWithPrecedingArgs_BindsWithoutDiagnostics` fail in Core.Tests |
| Delegate-factory cache | call `BuildInterpreterDelegateFactory` for every closure | cache regression test fails with distinct generated methods |
| Evaluator-owned delegate shortcut | make weak-map lookup unreachable | recursion depth regresses below `main` |
| Evaluator-owned async result | return `InvokeClosure` directly | async direct-invocation pin fails because `await` receives a raw value |

## Cache and indirect-call cost

Factories are cached by delegate `Type` in a `ConditionalWeakTable`, so projected `MetadataLoadContext` types are not pinned. A second weak table maps evaluator-owned delegates to their `ClosureValue`; indirect calls bypass `DynamicInvoke` and call the closure directly. Async closures retain the `Task` wrapping path.

Round-2 review measurements (interleaved, seven repetitions, median):

| Program | `main` | first revision `632d133d` | cached `72490605` |
|---|---:|---:|---:|
| `create10k` | 0.252 s | 0.646 s | 0.266 s (+6%) |
| `create100k` | 0.299 s | 4.520 s | 0.341 s (+14%) |
| `create300k` | 0.422 s | 15.805 s | 0.489 s (+16%) |
| `call100k` | 0.304 s | 0.459 s | 0.275 s (-10%) |
| `create300k` peak RSS | 83.4 MB | 103.3 MB | 97.8 MB (+17%) |

The cache is about 32x faster than the first revision on creation-heavy work, remains within 16% of `main` there, and is about 10% faster on call-heavy work. Peak RSS is about 17% above `main`. The weak-map shortcut removes the recursion-depth regression and restores parity with `main`; absolute stack limits are machine- and load-dependent.

Trade-off: closures of the same delegate type intentionally share one interpreter bridge `MethodInfo`. Compiled closures retain distinct generated method identities, while delegate targets and equality semantics remain distinct. `FuncToSystemDelegate.gs` therefore requires normalization or exclusion in interpreter golden-sample coverage (#2939).

## Related-shape probes

| Shape | `origin/main` | HEAD |
|---|---|---|
| Tuple | broken | fixed |
| CLR method expecting `Predicate<T>` | broken | fixed |
| `List<Func<...>>` | broken | fixed |
| delegate array | broken | fixed |
| delegate-valued dictionary | broken | fixed |
| delegate-valued map | broken | fixed |
| CLR generic constructor (`Lazy<T>`) | broken | fixed |
| `Action<T>` and captured state | broken | fixed |
| User method group in CLR collection | broken | fixed |
| G# class field | working | working |
| Compiled tuple path | working | working |

Producer-site materialization is required: reverting only the two producer returns makes all nine interpreter consumer pins fail. Review also found nine improvements across the 122 shipped samples and zero regressions. Compiled output remained untouched: 119 deterministic assemblies had zero byte differences.

## Follow-up issues and coordination

- #2938 tracks the pre-existing uncaught `TargetInvocationException` diagnostic leak through CLR member calls.
- #2939 tracks missing interpreter golden-sample conformance coverage.
- Numeric tuple-selector invocation `t.0(41)` needs both this PR and #2932. #2932 is still open, so its author was asked to add the combined passing pin if this PR lands first: https://github.com/DavidObando/gsharp/pull/2932#issuecomment-5141179781

No known callable-materialization shape remains broken.

## Validation

Rebased onto `origin/main` at `60e5b8bf`; validation head is `84c2970e`.

| Suite | `origin/main` (`60e5b8bf`) | `84c2970e` |
|---|---:|---:|
| Interpreter.Tests | 453 passed | **468 passed** |
| Core.Tests | 6913 passed, 1 skipped | **6913 passed, 1 skipped** |
| Compiler.Tests | approximately 3740 passed | **3745 passed** |

- Targeted Interpreter issue tests: 15 passed
- Async pin mutation: 1 failed with `'await' operand did not evaluate to a Task.`; restored control passed
- Targeted Compiler issue tests: 3 passed
- Release solution build with warnings as errors: 0 warnings, 0 errors

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -600,6 +600,14 @@ public sealed partial class Evaluator
 
             return value;
         }
+        else if (value is Delegate sourceDelegate
+            && node.Type?.ClrType is Type delegateType
+            && typeof(Delegate).IsAssignableFrom(delegateType))
+        {
+            return delegateType.IsInstanceOfType(sourceDelegate)
+                ? sourceDelegate
+                : CreateInterpreterDelegate(delegateType, sourceDelegate.DynamicInvoke);
+        }
         else if (node.Type == TypeSymbol.Object
             || node.Type is InterfaceSymbol
             || (node.Type is StructSymbol upcastTarget && upcastTarget.IsClass)
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -817,9 +817,7 @@ public sealed partial class Evaluator
         {
             if (InterpreterClosures.TryGetValue(target, out var closure))
             {
-                return closure.Function.IsAsync
-                    ? InvokeMaterializedClosure(closure, arguments)
-                    : InvokeClosure(closure, arguments);
+                return InvokeMaterializedClosure(closure, arguments);
             }
 
             return InvokeDelegate(target, arguments);
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -756,7 +756,7 @@ public sealed partial class Evaluator
             captured[v] = LookupVariable(v);
         }
 
-        return new ClosureValue(node.Function, node.Body, node.FunctionType, captured);
+        return MaterializeClosure(new ClosureValue(node.Function, node.Body, node.FunctionType, captured));
     }
 
     private object EvaluateMethodGroupExpression(BoundMethodGroupExpression node)
@@ -775,7 +775,7 @@ public sealed partial class Evaluator
             captured[node.Function.ThisParameter] = EvaluateExpression(node.Receiver);
         }
 
-        return new ClosureValue(node.Function, body, node.FunctionType, captured);
+        return MaterializeClosure(new ClosureValue(node.Function, body, node.FunctionType, captured));
     }
 
     private object EvaluateClrMethodGroupExpression(BoundClrMethodGroupExpression node)
@@ -804,16 +804,77 @@ public sealed partial class Evaluator
             throw new EvaluatorException("Attempted to invoke a nil function.", node);
         }
 
-        var closure = (ClosureValue)targetValue;
+        var arguments = new object[node.Arguments.Length];
+        for (var i = 0; i < node.Arguments.Length; i++)
+        {
+            arguments[i] = EvaluateExpression(node.Arguments[i]);
+        }
+
+        if (targetValue is Delegate target)
+        {
+            return InvokeDelegate(target, arguments);
+        }
+
+        return InvokeClosure((ClosureValue)targetValue, arguments);
+    }
+
+    private object MaterializeClosure(ClosureValue closure)
+    {
+        return closure.FunctionType.ClrType is Type delegateType
+            ? CreateInterpreterDelegate(delegateType, args => InvokeMaterializedClosure(closure, args))
+            : closure;
+    }
+
+    private static Delegate CreateInterpreterDelegate(Type delegateType, Func<object[], object> invoke)
+    {
+        var invokeMethod = delegateType.GetMethod(nameof(Action.Invoke));
+        var parameters = invokeMethod.GetParameters()
+            .Select(parameter => System.Linq.Expressions.Expression.Parameter(parameter.ParameterType, parameter.Name))
+            .ToArray();
+        var arguments = System.Linq.Expressions.Expression.NewArrayInit(
+            typeof(object),
+            parameters.Select(parameter => System.Linq.Expressions.Expression.Convert(parameter, typeof(object))));
+        var invocation = System.Linq.Expressions.Expression.Invoke(
+            System.Linq.Expressions.Expression.Constant(invoke),
+            arguments);
+        System.Linq.Expressions.Expression body = invokeMethod.ReturnType.IsSameAs(typeof(void))
+            ? System.Linq.Expressions.Expression.Block(invocation, System.Linq.Expressions.Expression.Empty())
+            : System.Linq.Expressions.Expression.Convert(invocation, invokeMethod.ReturnType);
+        return System.Linq.Expressions.Expression.Lambda(delegateType, body, parameters).Compile();
+    }
+
+    private static object InvokeDelegate(Delegate target, object[] arguments)
+    {
+        try
+        {
+            return target.DynamicInvoke(arguments);
+        }
+        catch (TargetInvocationException ex) when (ex.InnerException != null)
+        {
+            System.Runtime.ExceptionServices.ExceptionDispatchInfo.Capture(ex.InnerException).Throw();
+            throw;
+        }
+    }
+
+    private object InvokeMaterializedClosure(ClosureValue closure, object[] arguments)
+    {
+        var result = InvokeClosure(closure, arguments);
+        return closure.Function.IsAsync
+            ? WrapAsyncResult(closure.Function.Type, result)
+            : result;
+    }
+
+    private object InvokeClosure(ClosureValue closure, object[] arguments)
+    {
         var frame = new ConcurrentDictionary<VariableSymbol, object>();
         foreach (var kv in closure.CapturedLocals)
         {
             frame[kv.Key] = kv.Value;
         }
 
-        for (var i = 0; i < node.Arguments.Length; i++)
+        for (var i = 0; i < arguments.Length; i++)
         {
-            frame[closure.Function.Parameters[i]] = EvaluateExpression(node.Arguments[i]);
+            frame[closure.Function.Parameters[i]] = arguments[i];
         }
 
         using (PushFrame(frame))
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -27,6 +27,9 @@ namespace GSharp.Core.CodeAnalysis;
 public sealed partial class Evaluator
 #pragma warning restore CA1001
 {
+    private static readonly System.Runtime.CompilerServices.ConditionalWeakTable<Type, Func<Func<object[], object>, Delegate>> InterpreterDelegateFactories = new();
+    private static readonly System.Runtime.CompilerServices.ConditionalWeakTable<Delegate, ClosureValue> InterpreterClosures = new();
+
     private object EvaluateExpression(BoundExpression node)
     {
         try
@@ -812,6 +815,13 @@ public sealed partial class Evaluator
 
         if (targetValue is Delegate target)
         {
+            if (InterpreterClosures.TryGetValue(target, out var closure))
+            {
+                return closure.Function.IsAsync
+                    ? InvokeMaterializedClosure(closure, arguments)
+                    : InvokeClosure(closure, arguments);
+            }
+
             return InvokeDelegate(target, arguments);
         }
 
@@ -820,14 +830,23 @@ public sealed partial class Evaluator
 
     private object MaterializeClosure(ClosureValue closure)
     {
-        return closure.FunctionType.ClrType is Type delegateType
-            ? CreateInterpreterDelegate(delegateType, args => InvokeMaterializedClosure(closure, args))
-            : closure;
+        if (closure.FunctionType.ClrType is not Type delegateType)
+        {
+            return closure;
+        }
+
+        var result = CreateInterpreterDelegate(delegateType, args => InvokeMaterializedClosure(closure, args));
+        InterpreterClosures.Add(result, closure);
+        return result;
     }
 
     private static Delegate CreateInterpreterDelegate(Type delegateType, Func<object[], object> invoke)
+        => InterpreterDelegateFactories.GetValue(delegateType, BuildInterpreterDelegateFactory)(invoke);
+
+    private static Func<Func<object[], object>, Delegate> BuildInterpreterDelegateFactory(Type delegateType)
     {
         var invokeMethod = delegateType.GetMethod(nameof(Action.Invoke));
+        var invokeParameter = System.Linq.Expressions.Expression.Parameter(typeof(Func<object[], object>), "invoke");
         var parameters = invokeMethod.GetParameters()
             .Select(parameter => System.Linq.Expressions.Expression.Parameter(parameter.ParameterType, parameter.Name))
             .ToArray();
@@ -835,12 +854,15 @@ public sealed partial class Evaluator
             typeof(object),
             parameters.Select(parameter => System.Linq.Expressions.Expression.Convert(parameter, typeof(object))));
         var invocation = System.Linq.Expressions.Expression.Invoke(
-            System.Linq.Expressions.Expression.Constant(invoke),
+            invokeParameter,
             arguments);
         System.Linq.Expressions.Expression body = invokeMethod.ReturnType.IsSameAs(typeof(void))
             ? System.Linq.Expressions.Expression.Block(invocation, System.Linq.Expressions.Expression.Empty())
             : System.Linq.Expressions.Expression.Convert(invocation, invokeMethod.ReturnType);
-        return System.Linq.Expressions.Expression.Lambda(delegateType, body, parameters).Compile();
+        var inner = System.Linq.Expressions.Expression.Lambda(delegateType, body, parameters);
+        return System.Linq.Expressions.Expression.Lambda<Func<Func<object[], object>, Delegate>>(
+            inner,
+            invokeParameter).Compile();
     }
 
     private static object InvokeDelegate(Delegate target, object[] arguments)
--- a/test/Compiler.Tests/Emit/Issue2928TupleFunctionParityTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2928TupleFunctionParityTests.cs
@@ -1,0 +1,103 @@
+// <copyright file="Issue2928TupleFunctionParityTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using GSharp.Core.CodeAnalysis.Compilation;
+using GSharp.Core.CodeAnalysis.Symbols;
+using GSharp.Core.CodeAnalysis.Syntax;
+using Xunit;
+
+namespace GSharp.Compiler.Tests.Emit;
+
+/// <summary>
+/// Issue #2928: tuple-contained function values execute identically through
+/// interpreter and compiled backends.
+/// </summary>
+public class Issue2928TupleFunctionParityTests
+{
+    private const string Source = """
+        package TupleFunctionInterpreter
+        import System
+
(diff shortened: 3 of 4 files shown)


DavidObando:
Full local Compiler.Tests suite passed (exit 0).\n\n```text\nTest run for /Users/davidobando/GitHub/DavidObando/gsharp-wt/i2928/out/bin/Release/Compiler.Tests/GSharp.Compiler.Tests.dll (.NETCoreApp,Version=v10.0)
A total of 1 test files matched the specified pattern.

Passed!  - Failed:     0, Passed:  3711, Skipped:     0, Total:  3711, Duration: 1 h 4 m - GSharp.Compiler.Tests.dll (net10.0)\n```

DavidObando:
Addressed review in `72490605`:

- Cached delegate factories with `ConditionalWeakTable<Type, ...>`; cache mutation is detected by `Callable_DelegateFactoryIsReusedPerType`.
- Added weak delegate-to-closure map and bypassed `DynamicInvoke` for evaluator-owned indirect calls. Same-worktree isolation: shortcut disabled passes depth 2500 and aborts at 3000; enabled passes 3000 and aborts at 3500. Reviewer regressions at 2050/2300 now pass.
- Final warm medians: create10k 0.17s, create100k 0.25s, create300k 0.39s, call100k 0.23s; create300k RSS 81.7 MB.
- Filed #2938 for uncaught reflection exception diagnostics and #2939 for interpreter sample conformance.
- Renamed direct-exception test, corrected exception claim and async mutation evidence in PR body, and coordinated combined `t.0(41)` coverage with #2932.
- Release build: 0 warnings/errors. Interpreter 467/467; Core 6913 + 1 skip; Compiler 3711/3711. Pushed-head CI is green.

DavidObando:
## Rubber-duck review — round 2 (Opus 5), head `72490605`

**Verdict: MERGEABLE.** No blockers. Two SHOULD-FIX, two NITs. All six requested checks done with independent probes.

### 1. `ConditionalWeakTable.Add` duplicate key — refuted, safe
Proved four ways: CWT is **reference-keyed, not `Equals`-keyed** (`Delegate.CreateDelegate` twice on the same static method gives `Equals=True, ReferenceEquals=False`; both `Add` fine). `MaterializeClosure` is the only caller of `CreateInterpreterDelegate`, and its `args => InvokeMaterializedClosure(closure, args)` captures both `closure` and `this`, so a fresh display class is allocated every call. 200k adds in a loop: clean. `create300k.gs` = 300k real `Add`s: clean. No interning path exists. `Add` is correct; `AddOrUpdate`/`GetValue` not needed.

A forced double-`Add` mutation fails 14/14 Issue2928 tests, so the hazard is real — just unreachable.

### 2. Bypassing `InvokeDelegate` — nothing lost
All exception probes byte-identical between `632d133d` and `72490605`, including diagnostic text and source location. New typed-`catch` probes through the shortcut (CLR `DivideByZeroException` and a G# `throw`) both behave identically.

**Correction to the briefing:** the shortcut does not *lose* `DynamicInvoke` coercion — it **restores `main`**. `main` called `InvokeClosure` directly with zero reflective coercion; `632d133d` was the outlier that introduced it. The `TargetInvocationException` unwrap is likewise unreachable for evaluator-owned delegates (that path no longer touches reflection so cannot raise TIE); it still applies to real CLR delegates. Coercion probes across int64/float64/string/uint8/bool/zero-arg are identical.

### 3. The new `IsAsync` ternary is dead code (SHOULD-FIX)
`Evaluator.Expressions.cs:820-822` branches on `closure.Function.IsAsync`, but `InvokeMaterializedClosure` (`:881-887`) re-checks the very same flag. When `!IsAsync` the two calls are identical. The ternary is a tautology.

| Mutation | async1.gs | Interpreter.Tests | Core.Tests |
|---|---|---|---|
| control | ok | 467/467 | 6913 / 1 skip |
| force `true` (= the simplification) | ok, probes identical | 467/467 | 6913 / 1 skip |
| force `false` | **fails**: `'await' operand did not evaluate to a Task.` | 467/467 | 6913 / 1 skip |

Collapse to `return InvokeMaterializedClosure(closure, arguments);` — two lines and one branch fewer.

### 4. A real coverage gap this exposes (SHOULD-FIX)
Forcing `:820` to `false` breaks `await asyncLambda(x)` at runtime, yet **467/467 Interpreter and 6913 Core still pass**. A genuine semantic break is invisible to 7380 tests. Needs one Interpreter.Tests pin: an async lambda in a function-typed slot, invoked directly, awaited.

### 5. Multicast, identity, collectibility — OK
Multicast is unreachable from G# (`Delegate.Combine` → GS0159, `+=` on a named delegate type → GS0129); only CLR event `+=` creates one and those never re-enter `EvaluateIndirectCallExpression`. `EventSubscription.gs` output unchanged. Collectibility verified with a dynamically emitted **collectible** delegate type whose cached value's compiled tree strongly references the key — still collected. Ephemeron value→key semantics hold, so no `MetadataLoadContext` pinning (#1624 satisfied).

### 6. Measurements — the PR body's claims do not reproduce (NIT)
Interleaved, 7 reps, median:

| Probe | `main` | `632d133d` | `72490605` |
|---|---:|---:|---:|
| create10k | 0.252 | 0.646 | 0.266 (+6%) |
| create100k | 0.299 | 4.520 | 0.341 (+14%) |
| create300k | 0.422 | 15.805 | 0.489 (+16%) |
| call100k | 0.304 | 0.459 | 0.275 (−10%) |

The real win is **~32× versus `632d133d`** on create300k. This is *not* faster than `main` on creation-heavy work — it is 6–16% slower, the inherent cost of one delegate plus one CWT entry per materialization. That is acceptable, but it should be stated accurately.

Peak RSS (create300k, `/usr/bin/time -l`, 3 reps): `main` **83.4 MB**, `632d133d` 103.3 MB, `72490605` **97.8 MB** — i.e. **+17% over `main`**, not 81.7 MB below it.

Recursion ladder: the regression is **fixed** — `72490605` matches `main` exactly (both OK at 2300, SO at 2400), confirmed by mutating the shortcut off and reproducing `632d133d`'s SO@2300. The specific "passes 3000" figure is machine/load dependent and did not reproduce; the direction is right. Failure mode is a clean fail-fast `Stack overflow.`, same as `main`.

The cache-regression guard is genuine: reverting `:844` to per-call factory construction fails `Callable_DelegateFactoryIsReusedPerType` (13/14).

### 7. New parity divergence introduced by the cache (NIT)
Sharing one factory per delegate type means all closures of that type share one bridge `MethodInfo`:

```
methodid.gs:  d.Method.Name == g.Method.Name
  interp 72490605 : True      compiled : False   <- parity diff
  interp 632d133d : False     compiled : False
```

`FuncToSystemDelegate.gs` now prints `lambda_method98` twice. This is inherent to the cache — per-closure `Method` requires per-closure `Compile()`, which is the whole cost being removed. `Delegate.Equals`/`==`/`-=` are unaffected (targets remain distinct). Worth one line in the PR body and an addendum to #2939, since that issue proposes comparing against the compiled golden and this sample can never match under the interpreter.

**Correcting my own round-1 report:** I called that sample's counter nondeterministic. It is not — it is deterministic per build (3/3 identical each side). So the sweep diff is this real effect, not noise. 121/122 samples byte-identical after path normalization.

### Verified independently at `72490605`
Interpreter **467/467**, Core **6913 passed / 1 skipped**, Compiler **3711 passed**. `git merge-tree` against #2908, #2911, #2912, #2917, #2923, #2926, #2929, #2932 → **8/8 clean**.

#2938 and #2939 both accurately describe what they claim.

DavidObando:
Round-2 items landed at rebased head `84c2970e` (base `origin/main` `60e5b8bf`):

- Collapsed evaluator-owned delegate branch to `InvokeMaterializedClosure`.
- Added `Callable_AsyncDirectInvocation_ReturnsAwaitableTask`. Control passes; build-green mutation returning raw `InvokeClosure` output fails with `'await' operand did not evaluate to a Task.`
- Corrected performance/RSS claims: ~32x faster than first revision, within 16% of `main` for creation, ~10% faster for calls, ~17% higher peak RSS. Recursion described as restored to `main` parity, without machine-specific depths.
- Documented shared bridge `MethodInfo` trade-off; added #2939 addendum for `FuncToSystemDelegate.gs` normalization/exclusion.
- Release build: 0 warnings/errors. Interpreter 468/468; Core 6913 + 1 skip; Compiler 3745/3745.
- Pushed-head checks green.

DavidObando:
## Independent review verdict (round 2) — MERGEABLE

Posting retrospectively; this verdict previously existed only in reviewer session state. Reviewed at `72490605`; both SHOULD-FIXes below have since landed and are re-verified against `a6b1561f`.

### Round-2 checks

**`ConditionalWeakTable.Add` duplicate key — refuted, safe.** `ConditionalWeakTable` is reference-keyed, not `Equals`-keyed: two `Delegate.CreateDelegate` instances with `Equals=True, ReferenceEquals=False` both `Add` and retain independent values. `CreateInterpreterDelegate` returns a freshly allocated delegate on every call even through the cached factory (`ReferenceEquals=False`, `Equals=False` — targets differ); 200k `Add`s with an identical `invoke` instance, and 300k end-to-end via `create300k.gs`, neither throws. `MaterializeClosure` is the only caller and its `args => InvokeMaterializedClosure(closure, args)` captures both `closure` and `this`, so a fresh display class is allocated per call. `Add` is correct; `AddOrUpdate` not required. For contrast, a forced double-`Add` fails 14/14 targeted tests, so the concern was well-founded — just unreachable.

**Bypassing `InvokeDelegate` — nothing lost.** All exception probes byte-identical to the prior revision (direct / tuple / list / typed-catch / nil / CLR method group). Typed `catch` still works through the shortcut for both `DivideByZeroException` and a G#-thrown `InvalidOperationException`. Argument coercion probes (int64, float64, string, uint8, bool, zero-arg) identical. Worth recording: the shortcut does not *lose* `DynamicInvoke` coercion, it **restores `main`**, which called `InvokeClosure` on the raw `ClosureValue` with no reflective coercion.

**Identity / multicast / collectibility — OK.** Multicast is unreachable from G# source (`Delegate.Combine` → GS0159; `+`/`+=` on a named delegate type → GS0129), so combined delegates never re-enter `EvaluateIndirectCallExpression`; the fall-through is correct regardless. A delegate key is collected while only CWT-referenced, and a dynamically emitted **collectible** delegate `Type` is collected even though the cached factory value's compiled tree strongly references it — ephemeron semantics hold, no `MetadataLoadContext` pinning.

**Merge readiness.** `git merge-tree` clean against #2908, #2911, #2912, #2917, #2923, #2926, #2929, #2932.

**Suites, measured independently at `72490605`:** Interpreter 467/467 · Core 6913 passed, 1 skipped · Compiler 3711 passed (1 h 1 m, 0 failed, no environmental exclusions — `Gsharp.Extensions.dll` present in the build output).

### The two SHOULD-FIXes, and why the deletion was safe

**1. Redundant ternary in `EvaluateIndirectCallExpression` — now deleted.**

This was recommended on a **structural proof, not on mutant survival.** `InvokeMaterializedClosure(c, a)` is defined as `InvokeClosure(c, a)` followed by `IsAsync ? WrapAsyncResult(...) : result`, so on the `!IsAsync` path the two ternary arms were the same expression. The reduction is airtight because both operands are stable: `ClosureValue.Function` is get-only and assigned once in the constructor, and every one of the 14 assignments to `FunctionSymbol.IsAsync` lives in `Binding/` or `Emit/` — none in `Evaluator.*` — so the flag is frozen before evaluation begins. The reduced form also performs one read where the original performed two, so it can only be more self-consistent, never less.

Mutant survival corroborated this proof; it did not constitute it. Anyone auditing this deletion later should not read it as an instance of deleting code on survival evidence.

**2. Missing async pin — now added, and it kills.**

Forcing the async condition false broke `await asyncLambda(41)` at the CLI while **Interpreter.Tests 467/467 and Core.Tests 6913 still passed** — a genuine semantic break invisible to both suites. `Callable_AsyncDirectInvocation_ReturnsAwaitableTask` closes that gap. Verified by round-trip on `a6b1561f`:

| State | `GSharp.Core.dll` md5 | Targeted Issue2928 run |
|---|---|---|
| clean | `3f97f3f30e83c285214b8f02b0b70875` | 15/15 passed |
| `IsAsync` wrap forced off | `eae21f64d7b336c1c6ddfcc4fe384cdf` | **1 failed** — `Callable_AsyncDirectInvocation_ReturnsAwaitableTask` |
| restored | `3f97f3f30e83c285214b8f02b0b70875` (identical) | 15/15 passed |

Single-branch detector, no collateral failures. Builds were serial (`-m:1`) and gated on exit code, non-zero file size, and md5 agreement across the compiler and test output directories at every state.

### Corrections to the recorded evidence

- **The `FuncToSystemDelegate` sweep difference is not the nondeterministic lambda counter.** I previously characterised it that way; re-measurement shows the counter is deterministic per build (3/3 runs identical each side). The shared factory means all closures of one delegate type share a bridge `MethodInfo`, so `d.Method.Name == g.Method.Name` is now `True` interpreted vs `False` compiled — a real parity divergence that did not exist in the prior revision, and inherent to the cache. `Delegate.Equals`, `==` and `-=` are unaffected because targets remain distinct. Worth appending to #2939: that sample can never match the compiled golden under the interpreter.
- **Performance claims were optimistic.** Interleaved 7-rep medians on the review machine: create10k 0.252 → 0.266, create100k 0.299 → 0.341, create300k 0.422 → 0.489, call100k 0.304 → 0.275 (`main` → this branch). Peak RSS on create300k: `main` 83.4 MB, this branch 97.8 MB (3 reps, ±0.3 MB) — not below `main`. The genuine and substantial win is **~32× against the first revision** (create300k 15.8 s → 0.49 s). Recursion depth is restored to exactly `main`'s (both OK at 2300, clean stack-overflow abort at 2400); "exceeds `main`" did not reproduce.

### Infrastructure issue found while reviewing

A parallel-build race, distinct from the `Gsharp.Extensions` build-closure gap: `Gsharp.Extensions.csproj` shells out to the freshly built `gsc`, which under parallel MSBuild races the rewrite of `out/bin/Release/Compiler/GSharp.Core.dll` and loads a mid-write file:

```
Gsharp.NET.Sdk.Bootstrap.targets(114,5): error : Unhandled exception.
  System.IO.FileNotFoundException: Could not load file or assembly 'GSharp.Core, Version=0.3.0.0…'
  [.../Gsharp.Extensions.csproj]
```

Observed leaving three of four `GSharp.Core.dll` copies zero-length while a test project retained a stale valid copy — the exact shape that yields a false green under `dotnet test --no-build`. `-m:1` resolves it deterministically. Tracked as #2974.

PR #2950, opened 2026-07-31
Fix struct receiver capture in state machines

### LINKED ISSUE #2931: Struct receiver is corrupted in iterator state machine
## Repro

Both temporary and local struct receivers are corrupted when an instance iterator reads a field:

```gs
package StructReceiverIterator
import System

struct Box {
    var N int32

    func vals() sequence[int32] {
        yield N
    }

    func val() int32 -> N
}

// z1: temporary receiver, iterator — garbage
for value in Box{N: 100}.vals() {
    Console.WriteLine(value)
}

var box = Box{N: 100}

// z3: local receiver, iterator — also garbage
for value in box.vals() {
    Console.WriteLine(value)
}

// z2/z4 controls: both non-iterator calls are correct
Console.WriteLine(Box{N: 100}.val())
Console.WriteLine(box.val())
```

## Actual

Both iterator calls compile, load, and run but print nondeterministic garbage values instead of `100`. Both non-iterator controls print `100`.

Reproduced on `origin/main` `d4038f1d` and PR #2923.

## Expected

The iterator state machine must retain a value copy of its struct receiver, regardless of whether the receiver is a temporary or a local, so both enumerations print `100` deterministically.

This is separate from #2907: no loop variable or closure is involved, and the bad result predates that fix.

DavidObando:
## Summary

- dereference value-type `this` when shared bound-value emission captures it by value, fixing sync/async iterators, sync closures, async closures, and receiver-clause iterators
- apply the same value copy in the direct ordinary-async kickoff emitter, guarded to exclude class receivers
- add runtime/load/IL regression coverage for required receiver shapes, top-level generic structs, class controls, and interpreter parity
- track the separate pre-existing nested-generic-struct iterator failure in #2951

## Root cause

CLR value-type instance methods receive `this` as `T&`. Shared bound-value emission loaded only `ldarg.0` when synthesized iterator/closure kickoff code captured whole-value `this`, storing managed-pointer bits into a value-typed field. Ordinary async methods use a separate hand-written kickoff path with the same missing `ldobj T`. Result was loadable, executable assemblies that returned nondeterministic garbage.

Class receivers are object references and must not be dereferenced. Ordinary non-state-machine struct field reads are correct because their address-aware member path intentionally consumes managed-pointer `this` directly.

Async iterators and ordinary async struct methods shared the compiler defect and are fixed. Interpreter already used value semantics; its test is a parity guard.

## Runtime regression evidence

Every emitted-program test performs `Assembly.Load(File.ReadAllBytes(...))`, calls `GetTypes()`, executes a child `dotnet` process six times with stdout/stderr drained concurrently, asserts exact output, then runs IL verification.

Exact base `914b42a788960f35a7cffab0adcb0b86a370f3e1`:
- **Pins: 3** compiler tests failed exact-output assertions:
  - sync iterator/capture shapes expected `100 110 120 130 140 150 160 163 3 160 170 170 180 81 82 101`
  - async iterator shapes expected `200 201 210 211 220 220 230 231`
  - ordinary async shapes expected `300 310 320 91`
- **Guards: 3** passed on base and head:
  - non-iterator compiler control: `320 330`
  - class iterator compiler control: `400 410`
  - interpreter parity: `100 110 160 163 3 160 400`

Top-level `GenericBox[T]` shapes are covered. Nested generic struct iterators are not claimed as covered; their unrelated pre-existing type reification failures are #2951.

The issue body's minimal snippet is JIT-shape-sensitive: exact-base IL is wrong, but RyuJIT may launder the mismatched store into a copy. The larger pin fails reliably. Recorded on #2931: https://github.com/DavidObando/gsharp/issues/2931#issuecomment-5144733996

## Review follow-up and mutation testing

Added ordinary async class control `CBox(91).val()` and whole-value captures through sync closure (`81`), async closure (`82`), and receiver-clause iterator (`101`). Simplified the shared predicate by removing redundant `argIndex == 0`.

| Production hunk | Build-green semantic mutation | Detection |
|---|---|---|
| `MethodBodyEmitter.MemberAccess.cs` shared whole-value load | remove struct-`this` dereference while retaining ref-parameter handling | sync and async-iterator pins fail; added `81`, `82`, and `101` assertions become garbage |
| `StateMachineEmitter.cs` direct async capture | delete `{ IsClass: false }` while retaining `is StructSymbol receiverStruct` | ordinary async class assertion fails exactly: expected `91`, got `512` |

Unmutated control passed all 5 compiler tests before and after probes. Each mutation was restored and followed by a fresh Release rebuild before the next probe.

## Rebased validation

Base `914b42a788960f35a7cffab0adcb0b86a370f3e1` (executed Release suites):
- Core.Tests: 6915 passed, 1 skipped
- Compiler.Tests: 3802 passed
- Interpreter.Tests: 470 passed

Head `7f7c5da8130b142aaee0ab138374b2d81bde61dd` (executed Release suites):
- Core.Tests: 6915 passed, 1 skipped
- Compiler.Tests: 3807 passed
- Interpreter.Tests: 471 passed

Fixes #2931

### DIFF
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.MemberAccess.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.MemberAccess.cs
@@ -61,7 +61,7 @@ internal sealed partial class MethodBodyEmitter
         {
             this.il.LoadArgument(argIndex);
             if (ps.RefKind != RefKind.None
-                || (argIndex == 0 && ReferenceEquals(ps, this.structThisParameter)))
+                || ReferenceEquals(ps, this.structThisParameter))
             {
                 // Ref parameters and value-type `this` slots hold managed
                 // pointers; an ordinary value read must dereference them.
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.MemberAccess.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.MemberAccess.cs
@@ -60,10 +60,11 @@ internal sealed partial class MethodBodyEmitter
         if (variable is ParameterSymbol ps && this.parameters.TryGetValue(ps, out var argIndex))
         {
             this.il.LoadArgument(argIndex);
-            if (ps.RefKind != RefKind.None)
+            if (ps.RefKind != RefKind.None
+                || (argIndex == 0 && ReferenceEquals(ps, this.structThisParameter)))
             {
-                // ADR-0060: the parameter slot holds a managed pointer T&; an
-                // ordinary read of `p` in the body must dereference it.
+                // Ref parameters and value-type `this` slots hold managed
+                // pointers; an ordinary value read must dereference them.
                 this.EmitLoadIndirect(ps.Type);
             }
 
--- a/src/Core/CodeAnalysis/Emit/StateMachineEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/StateMachineEmitter.cs
@@ -1594,6 +1594,12 @@ internal sealed class StateMachineEmitter
             var thisFieldHandle = this.resolveFieldToken(kickoffSmStruct, plan.FieldMap.ThisField);
             il.LoadLocalAddress(0);
             il.LoadArgument(0);
+            if (function.ReceiverType is StructSymbol { IsClass: false } receiverStruct)
+            {
+                il.OpCode(ILOpCode.Ldobj);
+                il.Token(this.getStructTypeToken(receiverStruct));
+            }
+
             il.OpCode(ILOpCode.Stfld);
             il.Token(thisFieldHandle);
         }
--- a/test/Compiler.Tests/Emit/Issue2931StructReceiverStateMachineTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2931StructReceiverStateMachineTests.cs
@@ -27,7 +27,19 @@ public class Issue2931StructReceiverStateMachineTests
             struct Box {
                 var N int32
                 func vals() sequence[int32] { yield N }
+
+                func runSyncClosure() int32 {
+                    var read = func() int32 { return N }
+                    return read()
+                }
+
+                func runAsyncClosure() int32 {
+                    var read = async func() int32 { return N }
+                    return read().Result
+                }
             }
+
+            func (box Box) receiverVals() sequence[int32] { yield box.N }
 
             struct Holder {
                 var B Box
@@ -103,12 +115,15 @@ public class Issue2931StructReceiverStateMachineTests
             print(repeated)
 
             print(OuterIterator{B: Box{N: 180}}.vals())
+            Console.WriteLine(Box{N: 81}.runSyncClosure())
+            Console.WriteLine(Box{N: 82}.runAsyncClosure())
+            print(Box{N: 101}.receiverVals())
             """;
 
         AssertRunsWithExactOutput(
             Source,
             nameof(StructIteratorReceiversRetainValueCopies),
-            "100\n110\n120\n130\n140\n150\n160\n163\n3\n160\n170\n170\n180\n");
+            "100\n110\n120\n130\n140\n150\n160\n163\n3\n160\n170\n170\n180\n81\n82\n101\n");
     }
 
     [Fact]
@@ -201,17 +216,25 @@ public class Issue2931StructReceiverStateMachineTests
                 }
             }
 
+            class CBox(N int32) {
+                async func val() int32 {
+                    await Task.Delay(1)
+                    return N
+                }
+            }
+
             Console.WriteLine(Box{N: 300}.val().Result)
 
             var box = Box{N: 310}
             Console.WriteLine(box.val().Result)
             Console.WriteLine(GenericBox[int32]{Value: 320}.val().Result)
+            Console.WriteLine(CBox(91).val().Result)
             """;
 
         AssertRunsWithExactOutput(
             Source,
             nameof(AsyncStructMethodReceiversRetainValueCopies),
-            "300\n310\n320\n");
+            "300\n310\n320\n91\n");
     }
 
     [Fact]
--- a/test/Compiler.Tests/Emit/Issue2931StructReceiverStateMachineTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2931StructReceiverStateMachineTests.cs
@@ -1,0 +1,350 @@
+// <copyright file="Issue2931StructReceiverStateMachineTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Diagnostics;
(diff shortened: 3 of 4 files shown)


DavidObando:
## Round 1 review — MERGEABLE (2 should-fix, 1 nit)

Fix is correct at the root cause. No blockers. Reviewed at `6fcd6160`, plus an independent rebase onto `e21677f6` (`655708f4`, zero conflicts).

### Important: the issue's literal repro does NOT reproduce at exact base

Compiled #2931's exact snippet against exact-base source (`git checkout 60e5b8bf -- <both emit files>`, rebuilt), ran 6×: `100 / 200 / 300` every time.

The base IL *is* objectively wrong — kickoff `P1.Box::vals` at base stores a `Box&` into a `Box` field with no `ldobj`:
```
73 05000006  newobj SM   25 dup  16 ldc.i4.0  7D stfld state
25 dup       02 ldarg.0                        7D stfld <>4__this   ← Box& into Box field
```
HEAD inserts `71 02000002` = `ldobj Box`. RyuJIT sometimes launders the mismatched `stfld` into an implicit copy, yielding the right answer by luck. **Shape-sensitive — a reviewer who tests only the issue text gets a false negative.** The author's larger pin fails hard at base (`1867766288 1042843224 …`, nondeterministic across runs).

### Hunk 1 also fixes paths the PR never claims
All garbage at base, correct at HEAD: sync lambda capturing struct `this` (`176359000` → `81`); async lambda capturing struct `this` (`153241176` → `82`); Go-style receiver-clause struct iterator (`1869830752` → `101`).

### Blast radius — settled by instrumentation
Instrumented `MemberAccess.cs:63` to log every `ReferenceEquals(ps, structThisParameter)` hit, then compiled 122 corpus samples + ~40 probes:
```
122 corpus samples:  ZERO fires
probe battery:       34 fires, ALL: idx=0  refkind=None  isVT=True  isClassSym=False  ctx=plain
```
Ordinary struct member reads never reach this path — they route through `TryLoadVariableAddress` (`MemberAccess.cs:1682-1691`), which has its own `structThisParameter` branch emitting bare `ldarg.0`. `EmitLoadVariable` is reached only when `this` is read as a *whole value*, i.e. exactly the broken cases. No double-deref shape exists (zero fires with `isVT=False` or `refkind != None`).

Exactly **3** kickoff constructions exist (`StateMachineEmitter.cs:659`, `:1000`, `:1596`) and all three are covered. `GetEnumerator`/`GetEnumeratorObject` literals copy `<>4__this` from an SM *field*, not the `this` param — no fix needed. Generic structs are correct: `getStructTypeToken` emits a TypeSpec, and `GenericBox\`1::val` gets `ldobj TypeSpec(0x1B000004)` = `GenericBox<!0>` self-instantiation, which is right because the receiver type *is* the containing type.

### Pins 3 / guards 3 — author's claim confirmed accurate
Pins (fail at exact base, 3 runs each): sync iterator, async iterator, async method. Guards (pass both sides): non-iterator struct control, class iterator control, interpreter. The interpreter is correctly a guard — the production diff touches only `src/Core/CodeAnalysis/Emit/`, zero Evaluator lines, so interpreter behaviour is bit-identical by construction (verified via `gsi`: `41/99` and `170/170`, matching compiled).

### SHOULD-FIX 1 — `IsClass: false` guard is load-bearing but untested
`src/Core/CodeAnalysis/Emit/StateMachineEmitter.cs:1597`

Mutation **M2a** (delete `{ IsClass: false }`, keep `is StructSymbol receiverStruct`) → class receiver + ordinary async method returns `512 512` instead of `91 92`, and async lambdas in struct methods break. **The PR's entire suite stays green.** `AsyncStructMethodReceiversRetainValueCopies` (`Issue2931StructReceiverStateMachineTests.cs:179-216`) has no class control, unlike the async-iterator pin (`:154-160`) and the sync pin. ~4 lines:
```gs
class CBox(N int32) { async func val() int32 { await Task.Delay(1); return N } }
Console.WriteLine(CBox(91).val().Result)
```
Expected `91`; without the guard, `512`.

### SHOULD-FIX 2 — closure `this`-capture fixed but unpinned
`MemberAccess.cs:63-64`. Broken at base, fixed here, zero coverage: sync closure `176359000` → `81`; async closure `153241176` → `82`; receiver-clause iterator `1869830752` → `101`. Escaping-closure form also has correct copy semantics at HEAD. ~6 lines added to the existing sync pin source.

### NIT — `argIndex == 0 &&` is dead
`MemberAccess.cs:64`. `ThisParameter` maps to index 0 and nowhere else (`FunctionEmitter.cs:317`, `:322-331`, `:153-155`), so `ReferenceEquals(ps, structThisParameter) ⇒ argIndex == 0`. Probe over 156 programs: every fire `idx=0`; mutation M1a is inert. **13 production lines → 12:**
```csharp
if (ps.RefKind != RefKind.None || ReferenceEquals(ps, this.structThisParameter))
```
No smaller *structural* fix exists — the two hunks are genuinely disjoint emit paths (bound-literal vs hand-rolled IL) with no shared chokepoint.

### Reviewer retractions
- Core.Tests `Every_Report_Uses_A_Descriptor_…` failure — **retracted**, build-skew from my own mid-run rebase; passes after rebuild.
- `async func (b Box) aval()` → `InvalidProgramException` — **retracted**, fails identically at exact base. Pre-existing: `EmitAsyncKickoffBody` emits `ldarg.2` on a 1-arg method because receiver-clause methods double-count the receiver (`StateMachineEmitter.cs:1610` + `DeclarationBinder.Functions.cs:492`). Out of scope; hunk 2 not implicated.

### NOTE — pre-existing, out of scope
Nested generic struct iterators (`class Wrap[T] { struct Cell { … } }`) throw `TypeLoadException: … has a field of an illegal type`; `Cell[U]` throws `MissingMethodException`. **Identical at exact base.** Flagged only because the PR advertises generic-struct coverage while testing solely the top-level `GenericBox[T]` case, which does work.

### Latent hazard (no action)
`FunctionEmitter.cs:166` passes `structThisParameter: moveNextBody.ThisParameter` unconditionally, including when the SM is a class. If a MoveNext body ever value-read its own `this`, hunk 1 would emit `ldind.ref` on an object ref. `ctx=asyncMoveNext`/`iterMoveNext` fired **zero** times across every probe — unreachable today.

### Measurements (rebased series `655708f4`, full Release rebuild before each)
```
Core.Tests:       6913 passed, 1 skipped, 0 failed
Compiler.Tests:   3761 passed, 0 failed
Interpreter.Tests: 469 passed, 0 failed
```
`merge-tree` clean vs `e21677f6` and #2908, #2911, #2912, #2926, #2929, #2932. No interaction with #2917 (disjoint from `EmitLoadVariable`) or #2935 (Evaluator-only).

### Harness
`IlVerifier.Verify` at `:279` is decorative here; the load-bearing assertion is `Assembly.Load` + `GetTypes()` + 6× child-process exact-output at `:268-278`, pipes drained via `ReadToEndAsync` before `WaitForExit`. That harness **does** catch the base bug. Sound.

PR #2952, opened 2026-07-31
Fix nullable generic iterator metadata

### LINKED ISSUE #2927: Nullable sequence element types diverge between binding and emitted metadata
Concrete nullable value-type sequence elements are not represented consistently across binder and emitter.

## Repro 1: sync iterator metadata
```gs
package NullableSequence
import System

func values() sequence[int32?] {
    yield 1
    yield nil
}

for value in values() { Console.WriteLine(value) }
```
Current main can emit an assembly that throws `TypeLoadException: Method get_Current ... does not have an implementation`; ilverify reports `[MD] Class implements interface but not method`. Async sequence has the same element-type loss.

## Repro 2: binder conversion
```gs
package NullableSequenceParameter

func consume(values sequence[int32?]) int32 -> 1
var values = List[int32?]()
consume(values)
```
Current main rejects this with `GS0154` although `List<int32?>` implements `IEnumerable<int32?>`.

## Repro 3: nullable type parameter
```gs
package NullableGenericSequence
import System

func opt[T](v T) sequence[T?] {
    yield v
    yield nil
}

for value in opt[int32](5) {
    Console.WriteLine(value == nil ? "nil" : value.ToString())
}
```
Expected: `5`, then `nil`. Current main emits/verifies against inconsistent `IEnumerable<T>`/nullable element shapes and produces wrong value semantics. A partial binder-only fix can make the call-site view `IEnumerable<Nullable<int32>>` while the generic iterator remains `IEnumerable<int32>`, causing `EntryPointNotFoundException` at `GetEnumerator`.

Binder type construction, iterator interface metadata, and generic `T?` signature encoding must agree. Add sync/async, load, ilverify, and invocation coverage.

DavidObando:
## Summary

- specialize unconstrained nullable generic iterator returns into `class` and `struct` CLR variants
- route top-level, extension, receiver-clause, instance, shared, sync, and async iterators through the same specialization
- filter specialized overloads by inferred or explicit type-argument constraints
- retain `GS0508` for unsupported non-iterator and enclosing-type-parameter sequence positions
- add load, reflection metadata, execution, conversion, consumption, and diagnostic regression coverage

## Baseline triage (`e21677f6`)

- Repro 1: already fixed. Sync `sequence[int32?]` assembly loaded via `Assembly.Load(...).GetTypes()` and executed successfully.
- Repro 2: already fixed by #2923. `List[int32?]` converted to `sequence[int32?]` and executed without `GS0154`.
- Repro 3: remained. Normal compiler reported `GS0508`; removing that guard as a control emitted a loadable assembly that failed at `IEnumerable<T>.GetEnumerator()` with `EntryPointNotFoundException`.

## Root cause

`NullableTypeSymbol` stores `UnderlyingType.ClrType`, so open `T?` is nullability-blind at the CLR leaf. `SignatureEncoder` can encode `T?` correctly only after `T` has a stable `class` or `struct` constraint. Call-site substitution produced `Nullable<int32>`, while the unspecialized generic iterator metadata encoded bare `T`, making bound type, method signature, and state-machine interface disagree.

This change creates constrained iterator variants before binding/emission. Existing signature and state-machine emit paths then encode matching `T` or `Nullable<T>` metadata. No `StateMachineEmitter` changes were needed.

## Review follow-ups

- Added a non-iterator `sequence[T?]` return guard. Build-green M4 (`syntax.Body != null` without `ContainsYield`) now fails exactly this test; restored control passes 20/20.
- Added execution coverage assigning `values[int32](5)` to a concrete `sequence[int32?]` and producing `5`, `nil`.
- Characterized the current GS0266/GS0152 behavior for unconstrained generic callers and documented the actual enclosing-constraint remedy. Diagnostic improvement tracked by #2958.
- Filed pre-existing issues #2959 (missing canonical GS0506 docs) and #2960 (unloadable struct interface-property assembly). Both are independent of this PR.

## Verification

- Issue tests: 20/20 pass
- Pins: 8 fail on main source and pass here
- Guards: 12 pass on both
- Semantic mutations: original unmutated control 17/17 and all 21 branch mutations killed after rebuild; review M4 is also build-green and killed by the new pin (1 failed, 19 passed)
- Current baseline `17789519`: Core 7007 + 1 skipped = 7008; Compiler 3876; Interpreter 472
- Final head: Core 7007 + 1 skipped = 7008; Compiler 3883; Interpreter 472
- Compiler delta: reviewed PR was +4, not +5; three review follow-up facts make the final delta +7
- Full 148-file corpus comparison: 148/148, `BYTEDIFF=0 DIAGDIFF=0 RCDIFF=0`
- Corpus is non-regression evidence only: 3/148 files contain `yield` text, 1/148 contains `sequence[`, and 0/148 use `sequence[T?]`
- Release compiler build: 0 warnings, 0 errors

Fixes #2927

### DIFF
--- a/docs/diagnostics.md
+++ b/docs/diagnostics.md
@@ -988,6 +988,14 @@ types are specialized into `class` and `struct` variants automatically.
 Other unconstrained sequence type positions still require an explicit
 `struct`, `class`, or base-class constraint.
 
+A call from another unconstrained generic context cannot select either
+specialized iterator variant. Constrain the enclosing type parameter to
+`struct`, `class`, or a base class. Named or explicit arguments do not resolve
+that call: until [#2958](https://github.com/DavidObando/gsharp/issues/2958),
+it reports GS0266, and a nullable value-type argument may instead report
+GS0152. G# rejects source overloads differentiated only by constraints with
+GS0264, even though iterator specialization synthesizes equivalent variants.
+
 ## Internal compiler error diagnostics (GS9998–GS9999)
 
 | ID | Severity | Description |
--- a/docs/diagnostics.md
+++ b/docs/diagnostics.md
@@ -983,11 +983,10 @@ types are `System.Int32` and `System.Nullable<System.Int32>`.
 |----|----------|-------------|
 | GS0508 | Error | A nullable sequence element uses a type parameter that is not constrained to `struct` or a reference type. |
 
-`sequence[T?]` needs one stable CLR element representation. Use a `struct`
-constraint when `T?` must be `System.Nullable<T>`, or a `class`/base-class
-constraint when it must use reference-type nullability. An unconstrained `T`
-can represent either shape and is rejected until generic sequence
-specialization supports both.
+`sequence[T?]` needs one stable CLR element representation. Iterator return
+types are specialized into `class` and `struct` variants automatically.
+Other unconstrained sequence type positions still require an explicit
+`struct`, `class`, or base-class constraint.
 
 ## Internal compiler error diagnostics (GS9998–GS9999)
 
--- a/src/Core/CodeAnalysis/Binding/Binder.cs
+++ b/src/Core/CodeAnalysis/Binding/Binder.cs
@@ -1865,14 +1865,35 @@ public sealed class Binder
             && bodySyntax != null
             && cache.TryReuse(member, bodySyntax, out var reusedBody, out var reusedDiagnostics))
         {
-            diagnostics.AddRange(reusedDiagnostics);
+            AppendBodyDiagnostics(diagnostics, reusedDiagnostics, member);
             return reusedBody;
         }
 
         var (body, bodyDiagnostics) = bindAndLower();
-        diagnostics.AddRange(bodyDiagnostics);
+        AppendBodyDiagnostics(diagnostics, bodyDiagnostics, member);
         cache?.Store(member, bodySyntax, body, bodyDiagnostics);
         return body;
+    }
+
+    private static void AppendBodyDiagnostics(
+        ImmutableArray<Diagnostic>.Builder diagnostics,
+        ImmutableArray<Diagnostic> bodyDiagnostics,
+        FunctionSymbol member)
+    {
+        foreach (var diagnostic in bodyDiagnostics)
+        {
+            if (member.NullableSequenceSpecialization != NullableSequenceSpecializationKind.None
+                && diagnostics.Any(existing =>
+                    existing.Id == diagnostic.Id
+                    && existing.Severity == diagnostic.Severity
+                    && existing.Message == diagnostic.Message
+                    && existing.Location.CompareTo(diagnostic.Location) == 0))
+            {
+                continue;
+            }
+
+            diagnostics.Add(diagnostic);
+        }
     }
 
     /// <summary>
@@ -2942,7 +2963,8 @@ public sealed class Binder
                 return null;
             }
 
-            if (elementType is NullableTypeSymbol { UnderlyingType: TypeParameterSymbol typeParameter }
+            if (!ReferenceEquals(syntax, binderCtx.UnconstrainedNullableSequenceElementReturn)
+                && elementType is NullableTypeSymbol { UnderlyingType: TypeParameterSymbol typeParameter }
                 && !typeParameter.HasValueTypeConstraint
                 && !typeParameter.HasReferenceTypeConstraint
                 && typeParameter.ClassConstraint == null)
--- a/src/Core/CodeAnalysis/Binding/BinderContext.cs
+++ b/src/Core/CodeAnalysis/Binding/BinderContext.cs
@@ -104,6 +104,7 @@ internal sealed class BinderContext
     /// Mutated in place via <see cref="PushUnsafeContext"/>.
     /// </summary>
     public int UnsafeDepth;
+
 #pragma warning restore SA1401
 
     /// <summary>
@@ -157,6 +158,9 @@ internal sealed class BinderContext
     /// permitted.
     /// </summary>
     public bool InUnsafeContext => UnsafeDepth > 0;
+
+    /// <summary>Gets or sets the iterator return clause allowed to carry an unconstrained nullable sequence element.</summary>
+    public TypeClauseSyntax UnconstrainedNullableSequenceElementReturn { get; set; }
 
     /// <summary>
     /// Gets a value indicating whether the current <c>checked</c>/<c>unchecked</c>
--- a/src/Core/CodeAnalysis/Binding/BoundScope.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundScope.cs
@@ -305,6 +305,14 @@ public sealed class BoundScope
             return false;
         }
 
+        if (ReferenceEquals(a.Declaration, b.Declaration)
+            && a.NullableSequenceSpecialization != NullableSequenceSpecializationKind.None
+            && b.NullableSequenceSpecialization != NullableSequenceSpecializationKind.None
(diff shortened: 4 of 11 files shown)


DavidObando:
## Review — MERGEABLE

Head reviewed **`a04895b9`**, re-verified rebased as **`c3179260`** onto current `main` (`17789519`). No blockers. Two SHOULD-FIX items, both coverage/ergonomics rather than correctness defects in shipped behaviour.

### Measured baselines — `17789519` was previously unmeasured
Run directly, not derived:

| SHA | Core | Compiler | Interpreter |
|---|---:|---:|---:|
| **`17789519`** (current main) | **7007 + 1 skip = 7008** | **3876** | **472** |
| `f2ecef88` | 7008 | 3871 | 471 |
| `c3179260` (PR, rebased) | 7008 | **3880** | 472 |

**PR delta vs current main: +0 Core / +4 Compiler / +0 Interpreter.** The PR body's **+5 Compiler is off by one**. Arithmetic closes exactly: one `[Theory]`×4 `InlineData` + one `[Fact]` removed (−5), nine `[Fact]` added (+9) = **net +4**, 13→17 cases.

**Corpus is 148 samples, not 123** as reported.

### Rebase hygiene — clean
Production delta survives byte-for-byte across both rebases:
```
delta_orig (e21677f6→a04895b9 -- src/)   372 ± lines
delta_f2   (f2ecef88 →e96bda18 -- src/)  372 ± lines  IDENTICAL
delta_cur  (17789519→c3179260 -- src/)   372 ± lines  IDENTICAL
```
`test/`, `docs/`, `website/` delta hashes identical too, and **11/11 probe binaries byte-identical** pre- vs post-rebase. Every round-1 probe transfers unchanged.

### The #2911 × #2952 interaction is structurally unreachable — proven
The concern was that if one declared iterator emits two CLR methods, #2911's implicit-implementation matching must pick or reject. **It never gets the chance.**

| probe | HEAD | main | verdict |
|---|---|---|---|
| `j3` — `interface { func opt[T](v T) sequence[T] }`, no nullable, no specialization | GS0187 | GS0187 | generic interface *methods* are not implementable at all |
| `j4` — same, `[T struct]` | GS0187 | GS0187 | same |
| `j2` — unconstrained `sequence[T?]` | GS0508 + GS0187 | GS0508 ×2 + GS0187 | impl no longer double-reports; slot still unimplementable |
| `j1` — concrete `sequence[int32?]` iterator implementing interface method | runs `5, nil` | runs `5, nil` | **byte-identical `sha=7e01dee92e25`** |
| `k1` — generic interface *property* `sequence[T?]` | 6 errors | 6 errors | identical |
| `m1` — property mismatch + unconstrained `T?` | 1×GS0504 + 1×GS0508 | identical | **GS0504 and GS0508 coexist, one each, no cascade** |

`j3`/`j4` are decisive: GS0187 fires for *any* generic interface method, identically on both sides. A specialized iterator can never enter interface-implementation matching.

The specific `SubstituteNested` arity trap also cannot apply: **#2952 never calls `StructSymbol.SubstituteTypeForConstruction`** — it uses `Binder.SubstituteType`, different machinery.

### GS0508 relaxation verified — it did not leak
- **17-shape adversarial matrix**: no `EntryPointNotFoundException`, no rc=134 anywhere. Value / reference / nullable-value / enclosing-type-parameter / interface / struct-interface / async / both-instantiations-in-one-assembly all correct.
- **GS0508 still fires in all four non-iterator positions**: `c1_param`, `c2_field`, `c3_local`, `c4_nonyield_ret`.
- **Corpus 148/148** vs a `17789519`-built compiler: `BYTEDIFF=0 DIAGDIFF=0 RCDIFF=0`, same 9 failures both sides.
- **Diagnostic trio coherent**: GS0502/0503/0504/0506/0508 ascending, no duplicates; GS0508 prose identical in `docs/diagnostics.md:980` and `website/docs/ref/diagnostics.md:1727`.
- **Cross-PR**: `merge-tree` CLEAN vs #2912, #2929, #2932. `src/` overlap only in `Binder.cs`, and #2952's hunks at **1865/2942** vs #2912's earliest at **3169** and #2929's at **3026** — no collision. **Land #2952 first**: two small isolated hunks against two PRs that rework substitution/projection.

### SHOULD-FIX 1 — the sole iterator discriminator is unpinned
`DeclarationBinder.Functions.cs:580`. `IteratorDetection.ContainsYield(syntax.Body)` is the **only** clause scoping the GS0508 exemption to iterators. Mutant **M4** (drop the call, keep `syntax.Body != null`) is **build-green and undetected** at the rebased head:

```
c4_nonyield_ret  pristine: GS0155 GS0508
c4_nonyield_ret  M4:       GS0155            <- GS0508 lost
Issue2927NullableSequenceElementTypeTests under M4:  Passed! 17/17
```

`c1`/`c2`/`c3` still fire under M4, which is exactly why the existing guard misses it: `UnconstrainedNullableSequenceOutsideIteratorStillReportsSingleGS0508` (`:498`) uses a **parameter**, which never routes through `BindIteratorReturnType`. This is precisely the hole the new docs prose promises is closed.

**Fix:** one `[Fact]` using a non-iterator function *returning* `sequence[T?]` — `func consume[T]() sequence[T?] -> nil`.

Related: the deleted `ConcreteParameter` / `AnnotatedVariable` / `PassthroughFunction` shapes were **deleted, not converted**. All four verified to run correctly at HEAD (`5, nil`), so the deletions are legitimate — but no test now assigns a specialized generic iterator result to a concrete `sequence[int32?]`. Converting one back is cheap.

### SHOULD-FIX 2 — GS0266's advice is provably unfollowable
Calling a specialized iterator from an unconstrained generic context replaces actionable GS0508 with `GS0266: … Disambiguate with explicit types or named arguments`. **Both suggested remedies fail**: `opt(v: u)` → same GS0266; `opt[U](u)` → same GS0266. The real remedy — constraining the **enclosing** type parameter (`[U struct]`/`[U class]` → `2/2`) — is never mentioned. Related: `opt[int32?]` → `GS0152: does not satisfy the 'struct' constraint`, naming a constraint the user never wrote.

Root cause worth recording: **G# itself rejects constraint-differentiated overloads** (`func f[T struct]` + `func f[T class]` → GS0264, identical on base and head). The compiler synthesizes a construct the language forbids, which is why user code lands in an unresolvable GS0266.

### NIT — corpus carries zero feature signal
Of 148 samples: 3 contain `yield`, 1 contains `sequence[`, **0** use `sequence[T?]`. The 148/148 result is non-regression evidence only.

### Environmental warning for other agents
Standalone `git archive` trees show **19 Compiler failures** — 14 × `Sample_BuildsAndMatchesGolden` plus 5 in `Issue750ConstraintOverloadEmitTests.cs`. The Issue750 names look alarmingly on-topic for a PR editing constraint-based overload resolution, but the message is `Gsharp.Extensions.dll not found in any expected location`. **Failure sets byte-identical 19/19 on base and head.** Not a regression.

### Two pre-existing `main` bugs surfaced — explicitly NOT this PR
Both byte-identical on `17789519` and the rebased head.

1. **`docs/diagnostics.md` has no GS0506 section** while `website/docs/ref/diagnostics.md:1715` does. Present on all four SHAs including `e21677f6`. Uncaught because `DiagnosticIdUniquenessTests` validates severity only for descriptors that *are* documented (`TryGetValue`), so a missing doc entry is never detected.
2. **A struct implementing an interface property emits a broken assembly.** `k2.gs` (concrete `sequence[int32?]` interface property) compiles with **0 errors**, then dies at runtime: `TypeLoadException: Method 'get_Vals' in type 'K2.Box' does not have an implementation`, rc=134, `sha=d2096b94ff61` identical both sides. `gsc` accepting a program and emitting an unloadable assembly warrants its own issue.

Issue #2934, opened 2026-07-31
Compiler.Tests: IlVerifier can hang indefinitely waiting for child exit


DavidObando:
## Summary

`test/Compiler.Tests/IlVerifier.cs` drains redirected stdout/stderr concurrently, but then calls unbounded `proc.WaitForExit()`. If `dotnet-ilverify` hangs, any Compiler.Tests fact using `IlVerifier.Verify` can hang forever locally or in CI.

## Current code

```csharp
var stdoutTask = proc.StandardOutput.ReadToEndAsync();
var stderrTask = proc.StandardError.ReadToEndAsync();
proc.WaitForExit();
var stdout = stdoutTask.GetAwaiter().GetResult();
var stderr = stderrTask.GetAwaiter().GetResult();
```

## Expected

Use a bounded wait, kill the child process tree on timeout, wait for termination, then fail with a clear timeout message. Keep concurrent pipe draining so the existing full-pipe deadlock fix remains intact.

`Issue2900FixedPinReleaseEmitTests.CompiledProgram.Run()` already provides the local precedent: a 30-second `WaitForExit`, `Kill(entireProcessTree: true)`, final wait, and explicit timeout failure.

PR #2955, opened 2026-07-31
Unify nullable erasure for static-virtual interface properties

### LINKED ISSUE #2945: Unify nullable erasure for static-virtual interface properties
## Problem

Static-virtual interface property verification rejects nullable/generic relaxations that equivalent instance properties accept. Reproduces on `main` (`d4038f1d`) and PR #2911.

```gsharp
sealed interface I[T] {
    shared { prop V T? { get; } }
}

struct C : I[int32] {
    shared { prop V int32 -> 1 }
}
```

Result: `GS0397`, no emit. The equivalent instance property emits and loads because unconstrained `T?` erases to `T`.

Explicit static implementation rejects similarly:

```gsharp
struct C : I[int32] {
    shared {
        private prop (I[int32]) V int32 -> 1
    }
}
```

## Cause

Static-virtual emission does not apply the instance-property slot erasure. Binder verification must therefore retain exact `TypeSignaturesEquivalent` checks to prevent unloadable metadata; removing them emits a type whose static virtual `get_V` slot is unimplemented.

## Expected

Choose one consistent rule: teach static-virtual emission the same nullable erasure as instance properties, or document the stricter static rule and diagnostics.

DavidObando:
## Summary

- reuse instance-interface property compatibility for static-virtual binder matching and `MethodImpl` selection
- project constrained static property reads through the declared CLR slot type
- preserve value-nullable and setter invariance guards
- document the unified ADR-0089 rule

## Runtime evidence

A binder-only relaxation was built first as a known-bad control. It compiled the issue sample, but `Assembly.Load(...).GetTypes()` aborted with:

> Virtual static method 'get_V' is not implemented

With emission fixed first, the accepted matrix loads all types and executes implicit/explicit get and set slots through constrained generic calls in a bounded child process. A G# constrained read also loads and executes. `IlVerifier` runs afterward as additional evidence.

## Coverage

**3 pins / 1 guard at semantic-branch granularity.** Mutations independently detected old emitter matching (3 failures), restored exact binder matching (4 failures), old constrained-read projection (1 failure), and setter-covariance over-fire (2 failures).

Production delta: 4 files, +10/-22. The removed static helper deletes one semantic conjunct; its retained compatibility conjunct remains at both call sites.

## Compiler baseline

| SHA | Compiler.Tests (executed) | Result |
|---|---:|---|
| `17789519` (`origin/main`) | **3876** | 0 failed, 0 skipped |
| `238b46de` (rebased PR head) | **3879** | 0 failed, 0 skipped |

Delta **+3** is exactly the three new `[Fact]` tests; the two `Issue2888` tests were renamed, not added.

## Validation

- rebased onto `17789519` with zero conflicts
- rebased `Issue2945|Issue2888` control: 19 passed
- 148-sample corpus: 139 emitted assembly pairs byte-identical, zero diagnostic drift, 0 samples contain shared interface properties
- compiled/gsi parity: `1` / `1`

## Follow-up

Chained member access directly from a constrained static property (`T.V.Length`) reaches a pre-existing emitter ICE. This PR widens its reachability for nullable-erased slots; workaround is to store `T.V` in a local first. Tracked in #2962.

Fixes #2945


DavidObando:
## Review — MERGEABLE

Head reviewed **`fdc647a3`**, rebased onto `origin/main` (`17789519`) as **`2a28e046`** — `git rebase` clean, zero conflicts, `-warnaserror` green. No blockers. Three non-blocking findings.

### The missing number, measured

| SHA | Compiler.Tests (executed) | Result | Wall |
|---|---:|---|---|
| **`17789519`** (origin/main) | **3876** | 0 failed, 0 skipped | 1h09m |
| `2a28e046` (this PR, rebased) | **3879** | 0 failed, 0 skipped | 1h07m |

Delta **+3** = exactly three `[Fact]`s in the new 429-line file. Two `Issue2888` tests were **renamed**, not added, so that count is unmoved. Reconciles perfectly. *(Independently corroborated: the reviewer of PR #2952 measured `17789519` Compiler = 3876 the same way.)*

**C1 — "Eighteen lines of verification removed" was an overcount by 17.** `DeclarationBinder.InterfaceVerification.cs:705-719` deleted a whole helper, but its body was a two-conjunct expression:
```csharp
return TypeSignaturesEquivalent(interfaceType, implementationType, typeParameterMap)
    && IsInterfacePropertyTypeCompatible(implementationType, interfaceType, hasSetter, typeParameterMap);
```
The second conjunct is retained **verbatim** at both call sites. The semantic deletion is **one conjunct**. The rest of `+3/−18` is `private`→`internal` ×2 and call-site rewrites ×2. There was **one** guard branch to falsify, not eighteen — and it was falsified (M2).

**C2 — the docs concern is moot.** `grep -c GS0397 docs/diagnostics.md` → **0**; GS0397 was never in that file. `DiagnosticIdUniquenessTests` lives at `test/Core.Tests/CodeAnalysis/` and only checks *severity agreement for IDs already documented* — it does not require documentation. Leaving `docs/diagnostics.md` un-edited is correct.

### Known-bad control — reproduced, and it genuinely detects
Reverting **only** the emitter hunk `InterfaceImplEmitter.cs:1117-1126` while keeping the binder relaxation:
```
Wrote m1.dll
LOAD_FAIL ReflectionTypeLoadException
  LE: System.TypeLoadException: Virtual static method 'get_V' is not implemented on type 'Decl.C'
```
The author's methodology claim is true: the emitter hunk is load-bearing, and binder-alone ships unloadable metadata.

### Mutations — control first, restore **and rebuild** between each, all build-green
Unmutated control `--filter Issue2945|Issue2888` → **19/19 passed**.

| # | hunk | result | kind |
|---|---|---|---|
| M1 | `InterfaceImplEmitter.cs:1119` → back to `TypeSignaturesEquivalent(slot, cand, map)` | **3 failed** | PIN |
| M2 | reinstate deleted conjunct at `DeclarationBinder.Functions.cs:1885` + `InterfaceVerification.cs:1441` | **4 failed** | PIN |
| M3 | `ExpressionBinder.Access.MemberLookup.cs:3430` → back to `SubstituteThroughConstructedInterface` | **1 failed** (`GSharpConstrainedRead_UsesErasedSlotTypeAndRuns`) | PIN |
| M4 | `InterfaceVerification.cs:692` setter invariance → covariance | **2 failed** | GUARD |

**3 pins / 1 guard** at branch granularity. Each mutation kills a *different* test count, so localization is real — no single mega-test masking everything.

### Soundness — value-type covariance is unreachable
Erasure tests the **declared** type at `StructSymbol.cs:2365-2378`, before the substitution arm. Confirmed by metadata dump (`Assembly.Load` + reflection on the emitted slot):

| shape | emitted `I.get_V` slot | impl | verdict |
|---|---|---|---|
| `I[T]` unconstrained `T?` at `I[int32]` | `ret=T` (erased) | `System.Int32` | accept — exact CLR match |
| `I[T struct]` `T?` at `I[int32]` | `ret=Nullable\`1[T]` | `Nullable<Int32>` | accept — exact match |
| `I[T]` unconstrained at `I[int32?]` | `ret=T` | `Nullable<Int32>` | accept — exact match |
| `I[int32?]`, impl `int32` | — | — | **GS0397 reject** |
| concrete `int32?` iface, impl `int32` | — | — | **GS0397 reject** |

`NullableLifting.IsValueTypeNullable` (`NullableLifting.cs:132-137`) explicitly returns `true` for a `TypeParameterSymbol` with `HasValueTypeConstraint` — the highest-risk assumption resolved in the author's favour.

Extra probes all `LOAD_OK` + ilverify clean (including the MD pass, which demonstrably catches M1's failure mode): nested `Cell[T?]`/`Cell[int32]`, `T` at interface type, open implementors `G[U] : I[U]` and `G[U] : I[U?]`, two implementors at different instantiations in one assembly, implicit **and** explicit forms coexisting on one type, two interfaces sharing one property.

**Instance/static unification matrix** — 6 shapes, accept/reject **identical** on both paths (different IDs GS0504/GS0187 vs GS0397, same decision). That is the stated goal, delivered.

### Other verifications
- **`Issue2888` `+4/−4` accounted line by line.** Two `[Fact]` renames + two `CompileExpectingFailure` → `AssertEmitsAndLoads`, both exactly the interface `T?` / impl `int32` shape. The adjacent rejected list (`NullableStaticRejected`, `NullableExplicitStaticRejected`, impl `int32?`) is **untouched and still rejects**. #2888's protection is not weakened — it is **strengthened**: those two now assert `Assembly.Load(...).GetTypes()` rather than merely "the compiler said no".
- **`IlVerifier` is never the sole assertion**: `Assembly.Load().GetTypes()` → `RunChild` exact-stdout → *then* `IlVerifier.Verify`. The `KnownIssues.StaticVirtualInterface` ignore list is byte-identical to the pre-existing one from `950322ba` (#778) — not widened.
- **No #2934 regression**: `Issue2945…Tests.cs:385-390` uses both pipes `ReadToEndAsync()` before `WaitForExit(30_000)`, `Kill(entireProcessTree: true)`, with `\r\n`→`\n` normalization.
- **Corpus**: 148 `.gs`, 139 emit. **0 byte diffs across all 139, 0 diagnostic diffs.** The "zero samples with shared interface properties" claim is confirmed — only `samples/StaticVirtualInterfaces.gs` has interface `shared` members and they are all `func`. Non-regression evidence only.
- **ADR-0089 accurate** — every clause of the new paragraph independently probed true.

### SHOULD-FIX 1 — this PR widens reachability of a pre-existing ICE
`T.V.<member>` chained off a constrained static property → `GS9998: EmitDiagnosticException: an unresolved (error) expression reached code emission`.

**Pre-existing** — a control with no nullable at all fails identically on both sides:
```
z1.gs: sealed interface I[T] { shared { prop V T { get; } } }
       func Read[T I[string]](w T) int32 { return T.V.Length }
BASE 17789519: GS9998    HEAD fdc647a3: GS9998    ← identical
```
But the #2945 shape now *reaches* it, so the user-visible outcome degrades from a clean diagnostic to an ICE:
```
z2.gs (same, slot is T?):  BASE: GS0397 (clean)   HEAD: GS9998 (ICE)
```
A workaround exists (`var c = T.V` then `c.Length` — verified to compile and run). Not this PR's bug and fixing it here is scope creep. **Open an issue and reference it in the PR body** so the first user to hit it finds a paper trail.

### SHOULD-FIX 2 — cross-PR semantic collision with #2912
`merge-tree` vs `origin/main` and vs **#2903, #2912, #2929, #2932, #2952**: all rc=0, zero conflicts. #2932's `MemberLookup` hunks are at `:24-60`/`:436-485` (this PR is at `:3427`); #2952's `DeclarationBinder.Functions` hunks are at `:331-722` (this PR at `:1885`). No overlap.

**But #2912 (`6293a459`) adds a new arm inside `StructSymbol.SubstituteTypeForConstruction` at `:2392-2400`:**
```csharp
var newEnclosing = EnumSymbol.SubstituteEnclosingArguments(
    nestedEnum,
    t => SubstituteTypeForConstruction(t, subst, mapClrType));   // ← drops eraseReferenceNullability
```
All eight existing recursion sites use the `SubstituteNested` closure precisely so the flag propagates — including the structurally identical nested-**reference**-type arm three lines above, which correctly reads `SubstituteEnclosingArguments(nestedRef, SubstituteNested)`. This is the **exact flag-drop class** that has now recurred, and the merge is textually clean so nothing will catch it.

Blast radius today is instance interface properties (pre-existing); #2955 widens it to static-virtual ones. **Not #2955's defect.** Land **#2955 first**, then require **#2912 to change that lambda to `SubstituteNested`** before it lands.

### NIT — frozen 0.3 snapshot edited
`website/versioned_docs/version-0.3/ref/diagnostics.md` is a frozen release snapshot, touched exactly once before by the re-snapshot commit `203fd0cf` (#2277). This PR documents a **post-0.3** rule inside it, telling 0.3 readers about behaviour their compiler does not have. The same hunk also deletes a trailing blank line (`\n\n` → `\n`, byte-verified) unrelated to the change. Untested by CI either way.

### Tried to falsify, could not
No accepted program whose emitted `MethodImpl` fails to match the slot (11 hand-written shapes, all `Assembly.Load` + `GetTypes()` + ilverify MD clean); no route for nullable **value-type** covariance (every concrete-`int32?` and `T : struct` path rejects, and metadata proves the slot really is `Nullable<T>` there); no program the deleted conjunct protected (`TypeSignaturesEquivalent` was strictly stronger than the retained predicate on the erased slot, and M2 confirms the widening is exactly the intended set — 4 red when reinstated, all #2945/#2888 erasure tests, **no rejection test flipped**); no corpus drift; no way to make emitter and binder disagree on candidate selection (both iterate `StaticProperties` in order, break on first match, identical predicate and argument order).

*All build and mutation work was done in isolated clones under `/tmp/d2945/` because another agent had a live 50-minute `dotnet test` running inside the worktree; `i2945` was left exactly as found.*

Issue #2957, opened 2026-07-31
gsc: lower yield in function-literal iterator bodies


DavidObando:
## Reproduction

```gsharp
package Repro
import System

let values = func() sequence[int32] { yield 2 }
for value in values() {
    Console.WriteLine(value)
}
```

At `17789519599bf446ec274bb20bfa29759cf77acc`:

```
error GS9998: EmitDiagnosticException: Internal error: yield reached the emitter before iterator lowering.
```

The binder accepts `sequence[int32]` as an iterator return type, but function-literal bodies do not reach iterator lowering before emission.

Expected: emit and enumerate the function-literal iterator, matching a named `func values() sequence[int32]`.

Found while adding issue #2921's lambda definite-return guard. The new guard correctly excludes iterator return types; this failure is unchanged before and after that fix.

Issue #2953, opened 2026-07-31
Fixed-return epilogue bypasses non-void fall-through guard


DavidObando:
## Description

PR #2926 adds a compiler-generated fall-through guard for newly accepted non-void functions whose exhaustive enum switch has no runtime match. Functions containing a return inside `unsafe fixed` bypass that guard because `EmitFixedReturnEpilogue()` must be emitted first. The unmatched path silently returns the fixed epilogue's default slot (`0`) instead of throwing the guard exception.

## Contrast

Without `fixed`:

```gs
package NoFixed
import System
func F(x DateTimeKind) int32 {
    switch x {
        case DateTimeKind.Unspecified { return 1 }
        case DateTimeKind.Utc { return 2 }
        case DateTimeKind.Local { return 3 }
    }
}
public var result = F(DateTimeKind(99))
```

With `fixed`:

```gs
package WithFixed
import System
func F(x DateTimeKind, values []int32) int32 {
    switch x {
        case DateTimeKind.Unspecified {
            unsafe {
                fixed p *int32 = values {
                    return *p
                }
            }
        }
        case DateTimeKind.Utc { return 2 }
        case DateTimeKind.Local { return 3 }
    }
}
public var result = F(DateTimeKind(99), []int32{7})
```

| Program | `e21677f6` | PR #2926 |
| --- | --- | --- |
| Without `fixed` | GS0100 | `InvalidOperationException`: compiler-generated guard reached |
| With `fixed` | GS0100 | silently returns `0` |

## Expected

Both accepted functions should have consistent unmatched-value fall-through behavior. Preserve the fixed-return epilogue ordering: protected-region `leave` instructions target its label, so moving it after the guard is invalid. A future fix needs a conditional guard after the epilogue or equivalent control flow.

Issue #891, opened 2026-06-19
Using lambda as parameter to method call


DavidObando:
In C# I have this file /Users/davidobando/GitHub/DavidObando/Oahu/src/Oahu.Cli.App/Doctor/DoctorService.cs:
```csharp
public sealed class DoctorService : IDoctorService
{
    ...
    public DoctorService(ILogger<DoctorService>? logger = null, Func<HttpClient>? httpClientFactory = null)
    {
        ...
    }
   ...
```

From G# the following works, using a `func () R {}` declaration:
```gsharp
class DoctorServiceTests {
    @Fact
    func RunAsync_WithSkipNetwork_DoesNotMakeHttpCalls() {
        var svc = DoctorService(httpClientFactory: func() HttpClient {
            throw InvalidOperationException("HTTP must not be invoked when --skip-network is set")
        })
      ...
```

This however fails:
```gsharp
class DoctorServiceTests {
    @Fact
    func RunAsync_WithSkipNetwork_DoesNotMakeHttpCalls() {
        var svc = DoctorService(httpClientFactory: () -> {
            throw InvalidOperationException("HTTP must not be invoked when --skip-network is set")
        })
      ...
```

The error reported by the language server is: `Named arguments are only supported for data-struct .copy(...).` which is clearly not true, and also I expected this lambda to work here (perhaps it's a gap in the compiler or the language spec). G# file used was /Users/davidobando/GitHub/DavidObando/Oahu/tests/Oahu.Cli.Tests/DoctorServiceTests.gs

---
Similar situation in the sam G# file:
```gsharp
let net = report.Checks.Single(func(c DoctorCheck) bool { c.Id == "audible-api" }) // <-- this works
```

```gsharp
let net = report.Checks.Single(c -> c.Id == "audible-api") // <-- this doesn't work, error: Cannot find function Single.
```

Furthermore, hover on top of the `Single` extension method gives me the wrong message:
```
struct float32

Represents a single-precision floating-point number.
```

Related to #893

PR #2961, opened 2026-07-31
gsc: enforce GS0100 on lambda bodies

### LINKED ISSUE #2921: gsc: enforce GS0100 on lambda bodies

DavidObando:
## Summary

- enforce GS0100 for non-void function-literal and arrow-lambda bodies with reachable fall-through paths
- reuse the existing lowered `ControlFlowGraph.AllPathsReturn` analysis, including protected-region and exhaustive-switch projection from #2926
- preserve void, error, iterator, throw-only, infinite-loop, async, fixed, nested, expression-bodied, and imported-delegate lambda behavior
- add compiled/interpreted diagnostic parity and child-process runtime coverage

Fixes #2921

## Root cause

The issue description identified the arrow-lambda inference path: `InferLambdaReturnType` used `ControlFlowGraph.AllPathsReturn` only to infer whether a body completed normally, never to report GS0100. The exact repros use explicit `func(...) R { ... }` function literals, which bypass that inference path entirely and also lacked definite-return enforcement. Both binding paths now call one shared check after their final `BoundBlockStatement` is formed.

## Production delta

- `Binder.cs`: +1 line, inject the existing full iterator-return predicate
- `LambdaBinder.cs`: +19 lines, two call sites plus one shared definite-return check and callback plumbing

No new control-flow analyzer or emitter fallback was added.

## Verification

- exact switch and conditional repros: base emits; head reports one GS0100 and emits nothing
- current pre-fix symptom at base `17789519`: assemblies load and `GetTypes()` succeeds, then bounded child execution exits 134 with #2926's compiler-generated `InvalidOperationException` fall-through guard; the issue's older `InvalidProgramException` / `ReturnMissing` description is stale
- accepted outputs: `Assembly.Load` + `GetTypes()` + bounded child-process execution
- targeted tests at current head: Compiler 12/12, Interpreter 2/2

### Executed suites after rebase

Base `8dddbbf2`:

| suite | result |
|---|---:|
| Core | 7028 passed, 1 skipped |
| Compiler | 3893 passed, 0 failed |
| Interpreter | 489 passed, 0 failed |
| LSP | 452 passed, 0 failed |

Head `e02815ba`:

| suite | result |
|---|---:|
| Core | 7028 passed, 1 skipped |
| Compiler | 3905 passed, 0 failed |
| Interpreter | 491 passed, 0 failed |
| LSP | 452 passed, 0 failed |

Delta: +12 Compiler and +2 Interpreter = 14 tests.

Reviewer-confirmed pre-rebase head `1bff3720`: Core 7007 passed / 1 skipped, Compiler 3887 passed, Interpreter 474 passed. Its base `17789519` was Core 7007+1 / Compiler 3876 / Interpreter 472 / LSP 452.

### Pins

- function-literal switch escape
- function-literal conditional fall-through
- inferred arrow-lambda fall-through
- async, try/finally, nested, and fixed-containing fall-through
- no output after GS0100

### Guards

- void and expression-bodied lambdas
- nested valid lambdas
- throw-only and infinite-loop bodies
- try/finally and fixed bodies that return
- async lambdas
- imported generic delegate inference through LINQ and `Lazy<T>` (#891)
- iterator return types excluded from ordinary definite return
- error return types suppress cascading GS0100
- compiled/interpreted diagnostic parity

### Mutations

All five build-green semantic mutations were killed:

1. remove function-literal check: 6 Compiler cases fail
2. remove arrow-lambda check: Compiler and Interpreter parity pins fail
3. remove void exemption: accepted lambda guard fails
4. remove iterator exemption: sequence/yield guard fails
5. remove error-type exemption: new unknown-return-type guard fails

M5 control: 12 passed / 0 failed. M5 mutation: 11 passed / 1 failed, with the new test detecting the spurious GS0100 cascade while retaining GS0113.

### Corpus

All 148 sample programs were compiled with base and head compilers. Results: 142 successes on each side, 0 exit-code drift, 0 diagnostic drift, and 0 SHA-256 output-binary drift. The corpus contains 19 files with a non-void lambda; no corpus sample exercises a missing-return lambda. Reviewer follow-up independently found zero false positives across all 148 samples plus approximately 16 hand-written multi-shape probe files.

## Fixed and sequence notes

A missing-return lambda containing `fixed` now reports GS0100 before emission, shrinking #2953 to named functions. A valid all-returning fixed lambda still loads and executes.

A regular function-literal iterator remains blocked by the pre-existing `yield reached the emitter before iterator lowering` GS9998 failure on base and head. The iterator exemption itself is covered through interpreted binding. Follow-up: #2957.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/Binder.cs
+++ b/src/Core/CodeAnalysis/Binding/Binder.cs
@@ -255,6 +255,7 @@ public sealed class Binder
             bindBlockStatement: syntax => statements.BindBlockStatement(syntax),
             bindTypeClause: BindTypeClause,
             bindReturnTypeClause: (syntax, isAsync) => BindReturnTypeClause(syntax, isAsync),
+            isIteratorReturnType: IsIteratorReturnType,
             isAsyncIteratorReturnType: IsAsyncIteratorReturnType,
             resolveClrTypeForGenericArg: ResolveClrTypeForGenericArg,
             getCurrentFunction: () => this.function,
--- a/src/Core/CodeAnalysis/Binding/LambdaBinder.cs
+++ b/src/Core/CodeAnalysis/Binding/LambdaBinder.cs
@@ -71,6 +71,7 @@ internal sealed class LambdaBinder
     private readonly Func<BlockStatementSyntax, BoundStatement> bindBlockStatement;
     private readonly Func<TypeClauseSyntax, TypeSymbol> bindTypeClause;
     private readonly Func<TypeClauseSyntax, bool, TypeSymbol> bindReturnTypeClause;
+    private readonly Func<TypeSymbol, bool> isIteratorReturnType;
     private readonly Func<TypeSymbol, bool> isAsyncIteratorReturnType;
     private readonly Func<TypeSymbol, Type> resolveClrTypeForGenericArg;
     private readonly Func<FunctionSymbol> getCurrentFunction;
@@ -101,6 +102,9 @@ internal sealed class LambdaBinder
     /// declared return-type clause, with the <c>isAsync</c> flag so
     /// the still-on-Binder helper can apply the async-specific
     /// validation it already performs.</param>
+    /// <param name="isIteratorReturnType">Callback that returns
+    /// <c>true</c> when the supplied return type is an iterator shape
+    /// and therefore does not require an ordinary value return.</param>
     /// <param name="isAsyncIteratorReturnType">Callback that returns
     /// <c>true</c> when the supplied return type is an async-iterator
     /// shape (ADR-0041) and therefore does NOT participate in the
@@ -143,6 +147,7 @@ internal sealed class LambdaBinder
         Func<BlockStatementSyntax, BoundStatement> bindBlockStatement,
         Func<TypeClauseSyntax, TypeSymbol> bindTypeClause,
         Func<TypeClauseSyntax, bool, TypeSymbol> bindReturnTypeClause,
+        Func<TypeSymbol, bool> isIteratorReturnType,
         Func<TypeSymbol, bool> isAsyncIteratorReturnType,
         Func<TypeSymbol, Type> resolveClrTypeForGenericArg,
         Func<FunctionSymbol> getCurrentFunction,
@@ -156,6 +161,7 @@ internal sealed class LambdaBinder
         this.bindBlockStatement = bindBlockStatement ?? throw new ArgumentNullException(nameof(bindBlockStatement));
         this.bindTypeClause = bindTypeClause ?? throw new ArgumentNullException(nameof(bindTypeClause));
         this.bindReturnTypeClause = bindReturnTypeClause ?? throw new ArgumentNullException(nameof(bindReturnTypeClause));
+        this.isIteratorReturnType = isIteratorReturnType ?? throw new ArgumentNullException(nameof(isIteratorReturnType));
         this.isAsyncIteratorReturnType = isAsyncIteratorReturnType ?? throw new ArgumentNullException(nameof(isAsyncIteratorReturnType));
         this.resolveClrTypeForGenericArg = resolveClrTypeForGenericArg ?? throw new ArgumentNullException(nameof(resolveClrTypeForGenericArg));
         this.getCurrentFunction = getCurrentFunction ?? throw new ArgumentNullException(nameof(getCurrentFunction));
@@ -385,6 +391,7 @@ internal sealed class LambdaBinder
         // existing statement-body handling (no implicit return) so the #889
         // Action-style void-delegate path is preserved.
         body = SynthesizeFunctionLiteralTrailingReturn((BoundBlockStatement)body, syntax, returnType);
+        CheckAllPathsReturn((BoundBlockStatement)body, returnType, syntax.Body.Location);
 
         var captured = CollectCapturedVariables(body, synthetic.Parameters);
 
@@ -878,6 +885,7 @@ internal sealed class LambdaBinder
         }
 
         var bodyBlock = new BoundBlockStatement(syntax.Body, bodyStatements.ToImmutable());
+        CheckAllPathsReturn(bodyBlock, returnType, syntax.Body.Location);
         var fnType = FunctionTypeSymbol.Get(parameterTypes.MoveToImmutable(), BuildVariadicFlagsIfAny(parameterSymbols), observableReturnType);
         var captured = CollectCapturedVariables(bodyBlock, synthetic.Parameters);
 
@@ -1289,6 +1297,17 @@ internal sealed class LambdaBinder
         }
 
         return element;
+    }
+
+    private void CheckAllPathsReturn(BoundBlockStatement body, TypeSymbol returnType, TextLocation location)
+    {
+        if (returnType != TypeSymbol.Void
+            && returnType != TypeSymbol.Error
+            && !isIteratorReturnType(returnType)
+            && !ControlFlowGraph.AllPathsReturn(Lowerer.Lower(body)))
+        {
+            Diagnostics.ReportAllPathsMustReturn(location);
+        }
     }
 
     /// <summary>Binds and attaches user annotations on a lambda parameter.</summary>
--- a/test/Compiler.Tests/Emit/Issue2921LambdaBodyDefiniteReturnTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2921LambdaBodyDefiniteReturnTests.cs
@@ -176,6 +176,20 @@ public class Issue2921LambdaBodyDefiniteReturnTests
         Assert.Equal(expectedOutput, RunBounded(result.AssemblyPath, name));
     }
 
+    [Fact]
+    public void UnknownReturnTypeDoesNotCascadeGs0100()
+    {
+        var result = InvokeCompiler("""
+            package Issue2921.UnknownReturnType
+            var bad (bool) -> MissingType[int32] = (v bool) -> {
+                if v { return MissingFunction() }
+            }
+            """, "UnknownReturnType");
+
+        Assert.Contains("error GS0113:", result.Output, StringComparison.Ordinal);
+        Assert.DoesNotContain("error GS0100:", result.Output, StringComparison.Ordinal);
+    }
+
     private static object[] Case(string name, string source) => new object[] { name, source };
 
     private static object[] Case(string name, string expectedOutput, string source) =>
@@ -244,7 +258,7 @@ public class Issue2921LambdaBodyDefiniteReturnTests
         if (!exited)
         {
             process.Kill(entireProcessTree: true);
-            process.WaitForExit();
+            Assert.True(process.WaitForExit(5_000), $"{name}: emitted program did not stop after kill");
         }
 
         Assert.True(exited, $"{name}: emitted program timed out");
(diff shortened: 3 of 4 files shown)


DavidObando:
## Review — MERGEABLE

Head reviewed **`1bff3720`**, base `17789519`. Production is **+20/−0** across two files and reuses the existing enclosing-function pattern verbatim — close to the minimum viable shape. I hunted false positives hard (148 corpus samples + ~16 hand-written multi-shape probes) and found **zero**. One real gap, and it is a missing test, not a bug.

### Four corrections to the review brief

**C1 — the pre-fix symptom is no longer `InvalidProgramException`; the issue text is stale.** Base `17789519` already contains #2926 (`f2ecef88`), including the method-tail guard. Repro 1 at base:
```
Unhandled exception. System.InvalidOperationException: Compiler-generated guard reached:
non-void function fell through without returning a value.
   at LambdaMissingReturn.<Program>.<lambda1>(Int32 v)      exit=134
```
The IL is **verifiable**; there is no `ReturnMissing` any more. So the severity today is *"a statically-detectable error is deferred to runtime"*, not *"emits invalid IL"*. Still worth fixing — but #2921's "Current" section and the PR body should be corrected or the next reader chases a symptom that no longer exists.

**C2 — this PR *shrinks* #2953; it cannot worsen it.** GS0100 is a **binder** diagnostic, so it fires before emit and `EmitFixedReturnEpilogue()` ordering is irrelevant for lambdas. A lambda containing `fixed` with a missing return: base **compiles, runs, silently returns `0`**; head reports `GS0100` and does not emit. #2953 now survives only for **named** non-void functions containing `fixed`. Both directions are pinned by the author's two `FixedFunctionLiteral` cases.

**C3 — the inference risk lives on one path only.** The function-literal form never infers (`let a = func(v int32) { return v + 1 }` is `Void`, rejected with GS0122 on both base and head). Inference (#891) exists only on the arrow path. Pinned empirically anyway with 8 arrow lambdas assigned to **explicitly typed** variables so any drift becomes a conversion error — base and head both `[v=6|True|1.5|7|a|10|1|True]`. **Inference is invariant**; `CheckAllPathsReturn` only reports, it never writes `returnType`.

**C4 — corpus.** 148 confirmed; my non-void-lambda count is **16** (conservative) to **25** (broad), so the author's 19 is credible. The decisive measurement doesn't depend on it: `DIAGSAME=148, DIAGDRIFT=0, PAIRS=139, BYTEDIFF=0`, with all 16 lambda-containing samples byte-identical. That also closes a concern nobody raised: the check calls `Lowerer.Lower(body)` a second time purely for analysis, which could have bumped synthetic label/temp counters and perturbed metadata. It does not.

### Over-fire — nothing found
Every probe identical on base and head: void lambdas (incl. bare `return`), expression-bodied, throw-only/throw-in-branch, infinite loop, **exhaustive enum switch with no `default`**, async and void-async, nested and 3-deep curried, try/finally and catch-returns, imported generic delegates (`Select`/`Where`/`Aggregate`), struct- and class-capturing, generic local functions, `fixed`-inside-lambda, field initializers, lambda in `catch`. Iterator lambdas give **GS9998 on both** — pre-existing, tracked as #2957.

The exhaustive-switch case matters most: `ControlFlowGraph.cs:61-68` does `body = body.PreEmitAnalysisBody ?? body` then `Create(ProjectRegionsForDefiniteReturn(body), treatThrowsAsTerminators: true)` — **#2926's arm-aware projection is reused, not re-derived.**

**Two binding paths, no third, no hole.** Local functions bind through `BindFunctionLiteralExpression`/`BindLambdaExpression`; those are the only entry points (`:233`, `:620`), and there are **zero early returns** between body binding and the check on either path.

### Mutations — 4 of 5 branches pinned
Control **11 Compiler + 2 Interpreter = 13**, matching the claim exactly. Restored **and rebuilt** between every probe; all build-green and semantically inert.

| # | branch | Compiler | Interpreter | killed |
|---|---|---|---|---|
| M1 | call site `LambdaBinder.cs:394` (function-literal) → `Void` | **6 fail** | 2 pass | ✔ |
| M2 | call site `LambdaBinder.cs:888` (arrow/inferred) → `Void` | **1 fail** | **1 fail** | ✔ |
| M3 | `returnType != Void` → always true | **1 fail** | 2 pass | ✔ |
| M4 | `!isIteratorReturnType(...)` → always true | 11 pass | **1 fail** | ✔ |
| M5 | `returnType != Error` → always true | 11 pass | 2 pass | ❌ **survives** |

M1 and M2 kill **disjoint** test sets, so each path is independently pinned. I built the M3 compiler and confirmed it reports `GS0100` on `let c = func() { }` — the catastrophic over-fire — and `OrdinaryShapes` catches it. M4 is pinned in the **Interpreter** suite because the compiler path for iterator lambdas hits #2957/GS9998; that placement is deliberate and correct.

### SHOULD-FIX — S1: `LambdaBinder.cs:1305` `returnType != TypeSymbol.Error` has zero detectors
M5 survives all 13 tests. The line is **not** dead — I built the M5 compiler and probed a lambda with an unresolvable return type:
```
base      : GS0113 GS0122 GS0130
head      : GS0113 GS0122 GS0130          ← correct
M5-mutated: GS0113 GS0122 GS0130 + GS0100 ← spurious cascade
```
It genuinely suppresses a cascading diagnostic on already-broken code, and nothing would notice its removal. Impact is bounded (one extra diagnostic on a program that already fails to compile; it cannot break working code), which is why this is SHOULD-FIX rather than BLOCKER. Fix: one `[Fact]`, ~5 lines.

### NITs
- **N1** — `Issue2921LambdaBodyDefiniteReturnTests.cs:247`: unbounded `process.WaitForExit()` after `Kill(entireProcessTree: true)`. #2926's equivalent used a bounded `WaitForExit(5_000)`, and **#2934 is open precisely because a helper of this shape can hang**. Everything else in `RunBounded` is correct — both pipes `ReadToEndAsync()` at `:241-242` **before** `WaitForExit(30_000)` at `:243`.
- **N2** — correct #2921's "Current" section and the PR body per C1.
- **N3** — no `Core.Tests` binding test added, though Core is by far the fastest suite (4m34s vs 1h07m). Coverage is adequate via the Interpreter test's in-process `Emit`/`Evaluate`, so genuinely a nit.

### Verification standard
`grep -n "IlVerifier"` → **no matches**. The harness uses `Assembly.Load(File.ReadAllBytes(...))` + `GetTypes()` plus child-process execution. Given `IlVerifier` has passed broken assemblies five times this session, sidestepping it entirely is right. Rejection cases additionally assert `Assert.Single(errors)` — pinning against double-reporting, the obvious failure mode of a two-call-site fix — the exact GS0100 text, and `Assert.False(File.Exists(result.AssemblyPath))`. `gsi` parity asserts `Emit` and `Evaluate` produce the identical `["GS0100"]` set, and is killed by M2.

### Numbers — all executed, all confirmed
| suite | author | measured at `1bff3720` |
|---|---|---|
| Core | 7007 + 1 skip | **7007 passed, 1 skipped** ✔ (4m34s) |
| Compiler | 3887 | **3887 passed, 0 failed** ✔ (1h07m) |
| Interpreter | 474 | **474 passed, 0 failed** ✔ (20s) |

Against `17789519` (7007+1 / 3876 / 472): **+11 Compiler, +2 Interpreter = 13**, closing exactly against the 13 tests the control run counted.

### #2957 correctly deferred
Reproduced independently: `func() sequence[int32] { yield 2 }` → `GS9998: Internal error: yield reached the emitter before iterator lowering`, **identical on base and head**. Fixing it means routing function-literal bodies through iterator lowering — a substantial lowering change unrelated to a definite-return diagnostic. The `!isIteratorReturnType` carve-out is the correct forward-compatible shape. `IsIteratorReturnType` is the **superset** of `IsAsyncIteratorReturnType`, covering `SequenceTypeSymbol`, `AsyncSequenceTypeSymbol` and FullName-matched `IEnumerable`/`IEnumerator`/`IAsyncEnumerable`/`IAsyncEnumerator` — choosing it over the narrower predicate already wired into the ctor is correct.

### Cross-PR — order-independent
All merge-trees CLEAN vs `origin/main` and #2912, #2929, #2932, #2952, #2955. Beyond textual isolation: #2961's only `Binder.cs` change is at **:256**, while the nearest sibling hunk is #2952 at **:1859** (#2912 at :3192, #2929 at :3050) — no adjacency. No sibling touches `LambdaBinder.cs` or any predicate #2961 depends on. And for the real risk of a diagnostic-tightening PR — a sibling adding a program that newly fails GS0100 on a clean merge — I checked all five PRs' added test content: #2932's hit is a delegate *type* declaration (no body), #2952's are **named** iterator functions that bind through `Binder.cs`. **No sibling introduces a program #2961 would newly reject.**

Merge it **first** if you want a rule: its single `Binder.cs` line at :256 will not move under any sibling's rebase, whereas every sibling's `Binder.cs` hunks (:1859+) will shift under each other.

### Ponytail
Laziest thing that works, clears rung 2 cleanly. `Binder.cs:1385/:1444/:1947/:2102` already use `type != Void && !IsIteratorReturnType(type) && !AllPathsReturn(lowered)`; `CheckAllPathsReturn` is that expression plus one `!= Error` guard, called from the two places that needed it. The +19 is ~8 lines of logic and ~11 of ctor/doc plumbing forced by `LambdaBinder`'s existing delegate-injection style. Nothing to shrink.

DavidObando:
## Merge confirmation — head `e02815ba`

The MERGEABLE verdict above was issued against head `1bff3720`. Re-verified the current head before merging.

**Delta-isolation vs merge-base `8dddbbf2`:**

```
src/Core/CodeAnalysis/Binding/Binder.cs       |  1 +
src/Core/CodeAnalysis/Binding/LambdaBinder.cs | 19 +++++++++++++++++++
2 files changed, 20 insertions(+)
```

Production remains exactly **+20/−0 across two files**, identical to the approved head. The advance from `1bff3720` to `e02815ba` is a rebase onto `8dddbbf2` plus test-only additions responding to review:

- `UnknownReturnTypeDoesNotCascadeGs0100` — pins that an unresolved return type reports GS0113 and does **not** cascade into a spurious GS0100.
- Child-process waits tightened from unbounded `WaitForExit()` to `Assert.True(process.WaitForExit(5_000), …)`, so a hang fails the test instead of hanging the suite. This is the same defect class as #2934.

**CI:** 8 SUCCESS + 2 SKIPPED (`publish`, `publish-visual-studio-extension` are the normal skips on a non-release branch).

Order-independent with respect to #2966. Merging.

Issue #2956, opened 2026-07-31
gsi: fixed statements are unsupported by the interpreter


DavidObando:
## Problem

The compiled backend supports `fixed`, but `gsi` rejects every fixed statement at runtime:

```gsharp
let xs = []int32{1, 2}
unsafe {
    fixed p *int32 = xs {
        Console.WriteLine(xs.Length)
    }
}
```

```text
GS9999: 'fixed' (pinning) statements are not supported in the interpreter; they require the CIL pinned-local emit path.
```

The same program compiles, verifies, and prints `2` through `gsc`.

## Expected

Either implement interpreter semantics for fixed scopes (including array, string, and pinnable-reference sources plus cleanup on every exit), or define and document an explicit compiled-only language boundary. Until then compiled-vs-interpreted parity cannot be asserted for fixed programs.

Found while validating #2922. This is separate from #2922 malformed emitted managed-pointer conversion.

DavidObando:
## Design: decision is (B) — and (B)'s code is already written

Read-only investigation at `main` = `a6b1561f2`, scratch clone under `/tmp/d2956/`, probes run by invoking `gsi`/`gsc` directly.

**Decision: (B), declare a compiled-only boundary.** Not on balance — decisively, and justified from the source rather than from instinct. But two things in the framing need correcting first, and one of them inverts the priority of this issue.

---

## Q6 — Sequencing: **proceed in parallel, now.** Not blocked on #2972.

Re-confirmed at `a6b1561f2`: scanning all open PRs by file list, **#2972 is the only one touching any `src/Core/CodeAnalysis/Evaluator.*` file**, and it touches only `Evaluator.Statements.cs`.

It does not matter, because **the recommended work for this issue adds no production lines** (see below). The ADR, the tests, and the `docs/diagnostics.md` entry are all **new or non-Evaluator files**. Start immediately.

The one optional increment that *would* touch `Evaluator.Statements.cs` (§Q2, minting a diagnostic code) should be sequenced after #2972 merges. Split it into a second PR and there is no contention at all.

---

## Q1 — Measured current behaviour: a clean, specific, already-correct diagnostic

```gsharp
let xs = []int32{1, 2}
unsafe { fixed p *int32 = xs { Console.WriteLine(xs.Length) } }
```

```
GS9999: 'fixed' (pinning) statements are not supported in the interpreter; they require the CIL pinned-local emit path.
```
`gsi` exit code **1**.

[...]

PR #2964, opened 2026-07-31
Fix fixed pointer conversion IL

### LINKED ISSUE #2922: gsc: fixed containing switch emits ExpectedNumericType IL

DavidObando:
## Summary

Fixes #2922.

`fixed` pin setup converted managed pointers with `conv.u`. ILVerify requires a numeric operand for `conv.u`, but `ldelema` and `GetPinnableReference()` leave managed byrefs on the stack. RyuJIT accepted the malformed transition, so programs executed while verification failed.

`MethodBodyEmitter.Statements.cs:101-225` now passes each managed byref through `System.Runtime.CompilerServices.Unsafe.AsPointer<T>`, whose signature consumes `T&` and returns a numeric unmanaged pointer. Array/slice, string, and pinnable-reference sources share the conversion helper. Fixed return epilogue ordering from #2926 is untouched.

## Offending and corrected IL

Original issue repro:

```il
IL_001A: ldelema System.Int32
IL_001F: conv.u                         // stack contains Int32&, not numeric
IL_0020: stloc.2                        // int32*
```

Corrected:

```il
IL_001A: ldelema System.Int32
IL_001F: call void* Unsafe::AsPointer<int32>(!!0&)
IL_0024: stloc.2                        // int32*
```

The call consumes the managed pointer according to its parameter signature and returns the native numeric pointer required by the destination local. String pinning similarly uses `string.GetPinnableReference()` followed by `Unsafe.AsPointer<char>`; span-like pinning converts its pinned `T&` through the same helper.

## Diff

| File | Delta |
|---|---:|
| `src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Statements.cs` | +40/-22 |
| `docs/adr/0125-fixed-pinned-statement.md` | +18/-11 |
| `test/Compiler.Tests/Emit/Issue2906ExhaustiveSwitchReturnEmitTests.cs` | +1/-1 |
| `test/Compiler.Tests/Emit/Issue2922FixedSwitchIlTests.cs` | +459/-0 |

Production delta: +40/-22, 62 changed lines. No `MethodBodyEmitter.EmitMethodBody` or `EmitFixedReturnEpilogue` ordering change.

## Pins and guards

Exact-base run at `17789519` with the new tests: **6 failed pins, 1 passed guard**. HEAD: **7/7 passed**.

| Test branch | Base | HEAD | Role |
|---|---:|---:|---|
| slice fixed source verifies | fail `ExpectedNumericType` | pass | pin |
| string fixed source verifies | fail `ExpectedNumericType` | pass | pin |
| span fixed source has no `ExpectedNumericType` | fail | pass | pin |
| full switch/if control-flow matrix verifies | fail | pass | pin |
| array pin emits `ldelema; call Unsafe.AsPointer; stloc` | fail | pass | pin |
| #2906 fixed-return test without `ExpectedNumericType` suppression | fail | pass | pin |
| matrix loads and executes in a bounded child process | pass | pass | guard |

The matrix covers all-arm returns, partial returns, a switch-arm break escaping a labeled loop, nested switches, switch inside a loop inside `fixed`, nested fixed declarations, slice, string, struct-field sources, matching `if` controls, and void variants. G# has no standalone switch `break`; `break` targets loops, so the valid escaping shape is a branch from a switch arm to its labeled enclosing loop.

Existing fixed emission guards also pass: 40/40 across #1026, #1043, #2838, #2900, and #2906.

## Mutation checks

Unmutated control ran first. Every mutation was build-green; each restore was rebuilt before the next mutation.

| Semantic branch | Neutral mutation | Detection |
|---|---|---|
| array/slice byref conversion | restore `conv.u` after `ldelema` | slice pin, matrix verifier, and IL-sequence pin fail |
| string byref conversion | replace `Unsafe.AsPointer<char>` with `conv.u` | string pin and matrix verifier fail |
| span-like byref conversion | replace helper call with `conv.u` | span pin fails |
| symbolic generic pointee mapping | omit `ImmutableArray<TypeSymbol>` from MethodSpec creation | #2838 generic pointer guard fails: `AsPointer` closes over erased `System.Object` |
| #2906 suppression removal | restore old array `conv.u` while ignoring only inherent pointer errors | test fails with `ExpectedNumericType` |

## Verification

Both required axes run:

- ILVerify: exact repro and full array/string control-flow matrix verify with no ignored errors.
- Load/runtime: assemblies are loaded with `Assembly.Load`, enumerated with `GetTypes()`, then executed in child processes with both pipes drained asynchronously, a 10-second bound, and `Kill(entireProcessTree: true)` on timeout.
- Span pin test ignores only its pre-existing pinned-byref cleanup `StackUnexpected`; `ExpectedNumericType` is not ignored.
- #2906's `ExpectedNumericType` suppression is removed. Its `*p` dereference still requires `UnmanagedPointer` and `StackByRef`, inherent unsafe-pointer diagnostics unrelated to this issue.
- #2953 behavior is unchanged: unnamed exhaustive fallthrough still returns the default fixed-return slot (`0`).

Executed suites:

| Suite | base `17789519` | head `0b34855f` |
|---|---:|---:|
| Core.Tests | 7007 passed, 1 skipped | 7007 passed, 1 skipped |
| Compiler.Tests | 3876 passed | 3882 passed |
| Interpreter.Tests | 472 passed | 472 passed |

An initial head LSP run under two concurrent Compiler suites timed out in the pre-existing background-workspace timing test; the unloaded full rerun passed 452/452.

Release solution build: 0 warnings, 0 errors.

## Corpus

All **148** non-aspirational sample `.gs` files were compiled independently with `/deterministic+` by base and head compilers. **142 binaries** were produced on each side; six files had identical compile failures. Comparison covered SHA-256, normalized diagnostics, compile exits, runtime exits, stdout, and stderr: **0 differences**.

Samples containing executable `fixed` syntax: **0/148**. Corpus is non-regression evidence only; hand-written pins carry the fix.

## gsi parity

Compiled probe verifies and prints `2`. `gsi` exits 1 with:

```text
```

This is not an emitter defect: the interpreter has an explicit rejection and no pinning/cleanup runtime model. Filed #2956 to implement interpreter semantics or document a compiled-only boundary.

### DIFF
--- a/docs/adr/0125-fixed-pinned-statement.md
+++ b/docs/adr/0125-fixed-pinned-statement.md
@@ -112,7 +112,8 @@ brfalse NULL
 ldloc pinned; ldlen; conv.i4; brtrue NOTEMPTY
 NULL:     ldc.i4.0; conv.u; stloc ptr       // empty/null ⇒ null pointer
           br AFTER
-NOTEMPTY: ldloc pinned; ldc.i4.0; ldelema <elem>; conv.u; stloc ptr
+NOTEMPTY: ldloc pinned; ldc.i4.0; ldelema <elem>
+          call void* Unsafe.AsPointer<elem>(ref elem); stloc ptr
 AFTER:    <body>
           leave DONE
 FINALLY:  ldnull; stloc pinned              // release on every exit
@@ -123,20 +124,21 @@ DONE:
 **String** (`string pinned`):
 
 ```
-EmitExpr(source); dup; stloc pinned; conv.i
-dup; brfalse SKIP
-call get_OffsetToStringData; add            // skip the string header
-SKIP: conv.u; stloc ptr
+EmitExpr(source); dup; stloc pinned; brfalse NULL
+ldloc pinned; call char& string.GetPinnableReference()
+call void* Unsafe.AsPointer<char>(ref char); stloc ptr
+br AFTER
+NULL: ldc.i4.0; conv.u; stloc ptr
+AFTER:
 try { <body> } finally {
     ldnull; stloc pinned                     // release on every exit
 }
 ```
 
-`RuntimeHelpers.OffsetToStringData` is `[Obsolete]`, so the property getter is
-resolved reflectively (string literal, not `nameof`) to avoid CS0618. The
-classic `OffsetToStringData` lowering is used deliberately instead of
-`string.GetPinnableReference()` to avoid the `modreq(InAttribute)` ref-return
-signature noted above.
+`string.GetPinnableReference()` returns `ref readonly char`; the MemberRef
+retains its `modreq(InAttribute)` return signature. `Unsafe.AsPointer<T>`
+converts that managed pointer to the numeric native pointer stored in the
+user-visible `*T` local.
 
 **Span-like / `GetPinnableReference`** (`T& pinned`, issue #1043):
 
@@ -144,7 +146,7 @@ signature noted above.
 EmitExpr(span); stloc src                    // spill the source for addressing
 ldloca src; call instance T& GetPinnableReference()
 stloc pinned                                 // T& pinned = ref
-ldloc pinned; conv.u; stloc ptr
+ldloc pinned; call void* Unsafe.AsPointer<T>(ref T); stloc ptr
 try { <body> } finally {
     ldc.i4.0; conv.u; stloc pinned            // release on every exit
 }
@@ -162,6 +164,11 @@ emitted MemberRef carries the real BCL signature, including
 return-signature encoder reproduces required custom modifiers on by-ref returns
 (the same mechanism used for `ref readonly` indexers in ADR-0056), so the call
 binds at runtime instead of throwing `MissingMethodException`.
+
+The `Unsafe.AsPointer<T>` call is deliberate for every managed-pointer source.
+Emitting `conv.u` directly after `ldelema` or a ref-return leaves a managed
+pointer at an instruction that requires a numeric stack value, producing
+ilverify's `ExpectedNumericType` error even though RyuJIT accepts the method.
 
 ### 5. New node kinds, exhaustiveness, coverage matrix
 
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Statements.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Statements.cs
@@ -97,11 +97,10 @@ internal sealed partial class MethodBodyEmitter
 
     // Array-pin form: pin the array reference (`T[] pinned`) and derive
     // `&a[0]` via `ldelema`, guarding the null / zero-length array (→ null
-    // pointer), exactly as the C# compiler does.
+    // pointer).
     private void EmitFixedArrayPin(BoundFixedStatement node, int pinnedSlot, int pointerSlot)
     {
         var elementType = ((Symbols.PointerTypeSymbol)node.PointerVariable.Type).PointeeType;
-
         var nullLabel = this.il.DefineLabel();
         var notEmptyLabel = this.il.DefineLabel();
         var afterLabel = this.il.DefineLabel();
@@ -129,45 +128,50 @@ internal sealed partial class MethodBodyEmitter
         this.il.LoadConstantI4(0);
         this.il.OpCode(ILOpCode.Ldelema);
         this.il.Token(this.outer.memberRefs.GetElementTypeToken(elementType));
-        this.il.OpCode(ILOpCode.Conv_u);
+        this.EmitManagedPointerAsUnmanagedPointer(elementType);
         this.il.StoreLocal(pointerSlot);
 
         this.il.MarkLabel(afterLabel);
     }
 
     // String-pin form: pin the `string` reference (`string pinned`) and derive
-    // the char-data pointer via `RuntimeHelpers.OffsetToStringData`, guarding
-    // null (→ null pointer). This classic lowering avoids the `modreq`-bearing
-    // `string.GetPinnableReference()` ref-return that the member-ref encoder
-    // cannot reproduce.
+    // the char-data pointer through `GetPinnableReference`, guarding null
+    // (→ null pointer).
     private void EmitFixedStringPin(BoundFixedStatement node, int pinnedSlot, int pointerSlot)
     {
-        var skipLabel = this.il.DefineLabel();
+        var nullLabel = this.il.DefineLabel();
+        var afterLabel = this.il.DefineLabel();
 
         this.EmitExpression(node.PinnedSource); // string reference
         this.il.OpCode(ILOpCode.Dup);
         this.il.StoreLocal(pinnedSlot);          // pinned = s
-        this.il.OpCode(ILOpCode.Conv_i);         // (nint)s — address of the object
-        this.il.OpCode(ILOpCode.Dup);
-        this.il.Branch(ILOpCode.Brfalse, skipLabel); // null -> leave 0 as the pointer
+        this.il.Branch(ILOpCode.Brfalse, nullLabel);
 
-        var offsetGetter = typeof(System.Runtime.CompilerServices.RuntimeHelpers)
-            .GetProperty("OffsetToStringData")!
-            .GetGetMethod()!;
-        this.il.Call(this.outer.memberRefs.GetMethodReference(offsetGetter));
-        this.il.OpCode(ILOpCode.Add);            // address + OffsetToStringData = &s[0]
+        this.il.LoadLocal(pinnedSlot);
+        var getPinnableReference = typeof(string).GetMethod(
+            "GetPinnableReference",
+            BindingFlags.Public | BindingFlags.Instance,
+            binder: null,
+            types: Type.EmptyTypes,
+            modifiers: null);
+        this.il.Call(this.outer.memberRefs.GetMethodReference(getPinnableReference));
+        this.EmitManagedPointerAsUnmanagedPointer(TypeSymbol.Char);
+        this.il.StoreLocal(pointerSlot);
+        this.il.Branch(ILOpCode.Br, afterLabel);
 
-        this.il.MarkLabel(skipLabel);
+        this.il.MarkLabel(nullLabel);
+        this.il.LoadConstantI4(0);
         this.il.OpCode(ILOpCode.Conv_u);
-        this.il.StoreLocal(pointerSlot);         // p = (char*)result
+        this.il.StoreLocal(pointerSlot);
+        this.il.MarkLabel(afterLabel);
     }
 
     // Span-like pin form (ADR-0125 / issue #1043): pin the `T&` returned by a
     // public instance `ref T GetPinnableReference()` (e.g. `System.Span[T]` /
     // `System.ReadOnlySpan[T]`) into a `T& pinned` local, then derive `*T` via
-    // `conv.u`. Mirrors C# `fixed (T* p = span)`. `GetPinnableReference()` already
-    // returns the data pointer for an empty span (a ref to where element 0 would
-    // be), so no null/empty guard is required — matching C#'s codegen.
+    // `Unsafe.AsPointer<T>`. `GetPinnableReference()` already returns the data
+    // pointer for an empty span (a ref to where element 0 would be), so no
+    // null/empty guard is required.
     private void EmitFixedPinnableReferencePin(BoundFixedStatement node, int pinnedSlot, int pointerSlot)
     {
         var sourceSlot = this.locals[node.SourceVariable];
@@ -203,8 +207,22 @@ internal sealed partial class MethodBodyEmitter
         this.il.Token(this.outer.memberRefs.GetMethodEntityHandle(getPinnableReference, node.PinnedSource.Type)); // -> T&
         this.il.StoreLocal(pinnedSlot);          // T& pinned = ref
         this.il.LoadLocal(pinnedSlot);
-        this.il.OpCode(ILOpCode.Conv_u);
+        this.EmitManagedPointerAsUnmanagedPointer(
+            ((Symbols.PointerTypeSymbol)node.PointerVariable.Type).PointeeType);
         this.il.StoreLocal(pointerSlot);         // p = (T*)ref
+    }
+
+    private void EmitManagedPointerAsUnmanagedPointer(TypeSymbol pointeeType)
+    {
+        var openAsPointer = typeof(System.Runtime.CompilerServices.Unsafe)
+            .GetMethods(BindingFlags.Public | BindingFlags.Static)
+            .Single(method => method.Name == "AsPointer"
+                && method.IsGenericMethodDefinition
+                && method.GetParameters().Length == 1);
+        var closedAsPointer = openAsPointer.MakeGenericMethod(pointeeType.ClrType ?? typeof(object));
+        this.il.Call(this.outer.memberRefs.GetMethodEntityHandle(
+            closedAsPointer,
+            ImmutableArray.Create(pointeeType)));
     }
 
     private void EmitTryStatement(BoundTryStatement node)
--- a/test/Compiler.Tests/Emit/Issue2906ExhaustiveSwitchReturnEmitTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2906ExhaustiveSwitchReturnEmitTests.cs
@@ -311,7 +311,7 @@ public class Issue2906ExhaustiveSwitchReturnEmitTests
         var assembly = CompileVerifyLoadAndRun(
             "UnnamedFixed",
             Source,
-            ignoredVerificationErrors: new[] { "ExpectedNumericType" });
+            ignoredVerificationErrors: new[] { "UnmanagedPointer", "StackByRef" });
         // Known limitation (#2953): the fixed-return epilogue must be emitted first,
         // so fixed-containing non-void functions bypass the fall-through guard and
         // silently return the default slot instead of throwing.
--- a/test/Compiler.Tests/Emit/Issue2922FixedSwitchIlTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2922FixedSwitchIlTests.cs
@@ -1,0 +1,459 @@
+// <copyright file="Issue2922FixedSwitchIlTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Emit;
+using System.Threading.Tasks;
+using GSharp.Compiler;
+using Xunit;
+
+namespace GSharp.Compiler.Tests.Emit;
+
+/// <summary>
+/// Issue #2922: fixed sources must become numeric unmanaged pointers before
+/// switch control flow is emitted.
+/// </summary>
+public class Issue2922FixedSwitchIlTests
+{
+    private const string ExpectedMatrixOutput = "2\n3\n12\n12\n3\n3\n2\n2\n2\n12\n3\n3\n2\n2\nSI\n";
+
+    private const string MatrixSource = """
+        package Issue2922.Matrix
+        import System
+
+        var trace = ""
+
+        struct Holder {
+            public var Values []int32
+        }
+
+        func AllSwitch(x int32, xs []int32) int32 {
+            unsafe {
+                fixed p *int32 = xs {
+                    switch x {
+                        case 0 { return xs.Length }
+                        default { return 1 }
+                    }
+                }
+            }
+        }
+
+        func SomeSwitch(x int32, xs []int32) int32 {
+            unsafe {
+                fixed p *int32 = xs {
+                    switch x {
+                        case 0 { return xs.Length }
+                    }
+                }
+            }
+            return 3
+        }
+
+        func BreakSwitch(x int32, xs []int32) int32 {
+            var value = 0
+            switchLoop: for {
+                unsafe {
+                    fixed p *int32 = xs {
+                        switch x {
+                            case 0 {
+                                value = xs.Length
+                                break switchLoop
+                            }
+                            default {
+                                value = 1
+                                break switchLoop
+                            }
+                        }
+                    }
+                }
+            }
+            return value + 10
+        }
+
+        func NestedSwitch(x int32, xs []int32) int32 {
+            unsafe {
+                fixed p *int32 = xs {
+                    switch x {
+                        case 0 {
+                            switch xs.Length {
+                                case 2 { return 12 }
+                                default { return -1 }
+                            }
+                        }
+                        default { return 1 }
+                    }
+                }
+            }
+        }
+
+        func LoopSwitch(xs []int32) int32 {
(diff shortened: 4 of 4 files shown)


DavidObando:
## Rubber-duck verdict (Opus 5): NOT MERGEABLE as framed

Reviewed head `0b34855f` against base `17789519`, in isolated clones (worktree left untouched).

### Q0 — Is #2922 actually a bug? No.

`csc` (.NET 10, `AllowUnsafeBlocks`, **both Debug and Release**) on the equivalent C# program emits the identical diagnostic:

```
[ExpectedNumericType] P::F(int32,int32[]) [offset 0x17][found address of Int32]
[ExpectedNumericType] P::H(Span`1<int32>)  [offset 0x0A][found address of Int32]
```

csc's IL for `fixed (int* p = xs)` is `ldloc.1; ldc.i4.0; ldelema Int32; conv.u; stloc.0` with locals `[1] int32[] pinned` — the same shape gsc emits at base, which produces `[ExpectedNumericType][offset 0x1F]`, the exact offset quoted in #2922.

Per ECMA-335 III.1.6 (Table 8), `conv.u`/`conv.i` **do** accept `&` operands; the result is merely *unverifiable*, not illegal. Unverifiable IL is the expected output of `unsafe` code.

This repo already has a written policy for precisely this situation, at `test/Compiler.Tests/IlVerifier.cs:190-206` (`KnownIssues.RefStruct`):

> This is a known ilverify limitation, NOT a G# emitter bug: the same minimal C# program … compiled by `csc` emits identical IL and fails the same check.

#2922's acceptance criterion — "emitted fixed/pointer control flow should satisfy the CLI verifier" — is unachievable by construction, because `*p` can never be verifiable.

### BLOCKER 1 — the change makes verification strictly worse on real code

Measured on the very program whose test this PR modified (`Issue2906ExhaustiveSwitchReturnEmitTests.cs:290-317`, which dereferences `*p`):

| | ilverify errors |
|---|---|
| base `17789519` | **1** — `ExpectedNumericType` @0x2A |
| head `0b34855f` | **2** — `UnmanagedPointer` @0x30, `StackByRef` @0x31 |

ilverify aborts dataflow at the first failure, so removing `ExpectedNumericType` **unmasks** two errors it had been hiding. Reproduced independently on an array-deref + string-deref sample: base 3 errors → head 4.

Head is clean only when the `fixed` block never dereferences the pointer. Every `PinKinds` case in `Issue2922FixedSwitchIlTests.cs:252-318` returns `xs.Length` / `text.Length` / `1` and never reads `p` — the suite is shaped to the degenerate case.

Relatedly, the suppression was not removed but exchanged: `Issue2906ExhaustiveSwitchReturnEmitTests.cs:314` goes `-"ExpectedNumericType"` → `+"UnmanagedPointer","StackByRef"`, a net increase from 1 suppressed code to 2. The PR's own `Span` pin still requires `"StackUnexpected"` (`Issue2922FixedSwitchIlTests.cs:317`), confirmed still erroring at head (`StackUnexpected` @0x45).

### BLOCKER 2 — `Unsafe.AsPointer` launders rather than fixes

Mutation M5 (dropping the symbolic type argument at `:216`) makes gsc emit `Unsafe::AsPointer<object>(!!0&)` fed an `int32&`/`Pt&`. ilverify reports `StackUnexpected: found address of 'Pt', expected address of 'object'` — **and the program still runs correctly** (30/77/5). The JIT ignores `T` entirely; `AsPointer` is a no-op reinterpret. The unverifiable conversion has moved inside CoreLib, not disappeared. ADR-0125 currently concedes this ("even though RyuJIT accepts the method") without drawing the conclusion.

### Mutation results — semantic-branch granularity, rebuild between every probe

| # | Branch | `file:line` | Failures |
|---|---|---|---|
| C0 | unmutated control | — | 0 / 24 pass |
| M1 | array pin → `Conv_u` | `:131` | 4 detected |
| M2 | span pin → `Conv_u` | `:210-211` | 1 detected |
| M3 | string non-null → `Conv_u` | `:150` | 2 detected |
| M4 | `ClrType ?? …` → always `object` | `:215` | 4 detected |
| **M5** | drop `ImmutableArray.Create(pointeeType)` | `:216` | **0 — UNCOVERED** |
| **M6** | string null-guard `Brfalse` → `Pop` | `:143` | **0 — UNCOVERED** |

Reviewer count is **6 pins, 0 guards** (PR claims 6/1). All three `PinKinds` rows fail at base, and both Matrix tests plus the IL-shape test are new-behaviour assertions — no test passes at both base and head, so nothing qualifies as a guard.

Coverage gaps: no user-defined struct pointee (M5), no null string (M6), and decisively **no test that dereferences the pointer**, which is the only shape that reveals BLOCKER 1.

### Verified good — author is right on these

- **The `modreq` IS emitted.** `call instance char& modreq([System.Private.CoreLib]…InAttribute) System.String::GetPinnableReference()`, binds and runs. The deleted warning comment is genuinely obsolete: ADR-0125:163-167 records that the encoder gained this capability via ADR-0056.
- **The string hunk is a real improvement.** Contrary to the review brief, gsc's base string path did *not* match Roslyn — base emitted `conv.i` on a `string` ref (`[ExpectedNumericType][found ref 'string'][offset 0x03]`) while Roslyn uses `GetPinnableReference()`. This hunk moves *toward* Roslyn.
- **GC safety is clean.** Locals signatures confirmed `[1] int32[] pinned` / `[1] string pinned` / `[1] int32& pinned`, with `finally { ldnull; stloc }`. A stress probe (3× `GC.Collect` plus 20k-allocation churn inside the block, 3 runs each) gave deterministic `6912` / `65066` at both base and head.
- **The `?? typeof(object)` fallback is sound and not dead.** `ImportedMemberRefFactory.cs:804-808` documents `typeof(object)` as the established issue-#320 placeholder, and `RequiresSymbolicProjection` (`TypeSymbol.cs:831`) is exactly complementary to `ClrType == null`. A user-struct pointee correctly emits `AsPointer<valuetype Pointee.Pt>`.
- Head `Compiler.Tests` **3882 passed / 0 failed** (executed), matching the PR claim. #2953 unchanged, #2926's epilogue untouched, corpus `fixed` usage confirmed 0/148, `merge-tree` clean against `origin/main` and #2912/#2929/#2932/#2952/#2955/#2961. #2956 is a legitimate out-of-scope filing.

### SHOULD-FIX / NIT

- **SHOULD-FIX** `MethodBodyEmitter.Statements.cs:211-217` — `.Single(...)` over `Unsafe.GetMethods()` throws if a future CoreLib adds another 1-arg generic `AsPointer` overload, and it is uncached, running per pin site.
- **SHOULD-FIX** `docs/adr/0125-fixed-pinned-statement.md:168-172` — the rationale notes RyuJIT accepts the old form but omits that `csc` *emits* it. Any decision to diverge from Roslyn must be recorded as deliberate.
- **NIT** `Issue2922FixedSwitchIlTests.cs:352` — `SlicePin_ConvertsManagedPointerThroughUnsafeAsPointer` asserts on `PinKinds().First()`, which is the array case; the `ldelema+5` offset arithmetic is brittle.

### Recommended resolution

Keep the **string** hunk — it fixes a genuine divergence from Roslyn. Revert the array and span `AsPointer` changes, which regress verification. Classify `ExpectedNumericType` as a known unsafe-pointer ilverify limitation alongside the existing `KnownIssues.RefStruct` precedent, citing the csc equivalence above. Add the missing pointer-dereference test as a guard.

### Unresolved

`fixed p *char = s` with `s string?` returns `1` for both `nil` and `"Z"`, identically at base and head, in both return-form and assign-form. Pre-existing and likely #2953-family, not a regression from this PR, but M6's runtime impact remains unproven in both directions.

DavidObando:
## Addendum to my verdict — two items, one of which lands inside the rescope

Head reviewed is still `0b34855f` (unchanged). My original verdict was **NOT MERGEABLE _as framed_**, and I understand the plan is to rescope to the **string-pin path only**, with the array/span `conv.u` change dropped and the residue classified as a documented known limitation. **That is the right response and it does not make my verdict stale — it acts on it.** Below: what survives the rescope, what retires with it, and one new executed result.

### 1. NEW — the `ClrType ?? typeof(object)` in `EmitManagedPointerAsUnmanagedPointer` is **not** the #2965 defect

Raised because #2965's root cause is the same-looking `ClrType ?? typeof(object)` at `src/Core/CodeAnalysis/Emit/MethodBodyEmitter.cs:1396` (`ResolveChannelElementClrType`), where a compilation-local struct has a null `ClrType`, degrades to `object`, and the emitter builds `ChannelReader<object>.TryRead(out object)` against a `Pair&` → invalid IL, `InvalidProgramException`, exit 134. Here a symbolic `ImmutableArray.Create(pointeeType)` is passed alongside the reflection `MakeGenericMethod(...)`, and it was open which one reaches the token.

**Settled by executed probe. The symbolic path wins.** Compiling a `fixed` over an array of a compilation-local struct with head `0b34855f`:

```gsharp
struct Pt { var X int32
            var Y int32 }
func F(xs []Pt) int32 {
    unsafe { fixed p *Pt = xs { return 1 } }
    return 0
}
```

emits

```
[1] valuetype Default.Pt[] pinned,
IL_001a: ldelema Default.Pt
IL_001f: call void* [System.Private.CoreLib]System.Runtime.CompilerServices.Unsafe::AsPointer<valuetype Default.Pt>(!!0&)
```

`AsPointer<valuetype Default.Pt>` — **not** `AsPointer<object>`. The reflection `MethodInfo` is used only to *locate* the method; the emitted type argument comes from the symbolic array. This is consistent with mutation **M5** from the original review, where forcing the symbolic array to `default` made `AsPointer<object>` appear in the emitted IL — i.e. the fallback is only observable once the symbolic path is removed.

So `?? typeof(object)` here is the documented issue-#320 placeholder convention (`ImportedMemberRefFactory.cs:804-808`), not a silent-miscompile path. **The difference from #2965 is that `ResolveChannelElementClrType` has no symbolic sibling** — nothing overrides its `object`. If #2965 is being fixed by giving that site a symbolic path, this call site is the working precedent to copy.

The `??` is nonetheless **dead in practice** at this site and would be clearer as an assertion.

### 2. SURVIVES THE RESCOPE — the string null-guard is uncovered, and it is in the code you are keeping

`src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Statements.cs:148` — `this.il.Branch(ILOpCode.Brfalse, nullLabel)`, the null-string guard in `EmitFixedStringPin`. Replacing it with `Pop` (removing the guard) leaves the suite **fully green**, and the mutant is demonstrably live in the binary and semantically non-inert:

| evidence | mutated | restored |
|---|---|---|
| `brfalse` count in emitted IL | **0** | 1 |
| IL at the site | `IL_0003: pop` before `call … String::GetPinnableReference()` | guard present |
| null string at runtime | `NullReferenceException at System.String.GetPinnableReference()` | prints `7` |
| `Issue2922*` + `Issue2906*` suite | **24 / 0 pass** | 24 / 0 pass |

Triple-confirmed. Final pass used a hardened harness: build the **test project** with `-m:1`, gate on **exit code → zero-byte → propagation**, assert the md5 of `out/bin/Release/Compiler.Tests/GSharp.Core.dll` both changed and matched the Core output. Exact bidirectional round-trip `3e65ffef → 0eb1d3a5 → 3e65ffef`, `BUILD_EXIT=0`, `0 Error(s)`.

> My first two attempts at this probe were **invalid** and I withdrew them: the harness asserted mtime on `out/bin/Release/Core/GSharp.Core.dll` while `dotnet test --no-build` read `out/bin/Release/Compiler.Tests/GSharp.Core.dll`, which had not been refreshed since clone time (`eb31e721`, mtime `1785539247`). The conclusion survived re-testing unchanged, but the original evidence did not support it.

**Ask:** the rescoped PR should add a null-string `fixed` test. This is a request for a test — **not** a suggestion to delete the guard; the guard is correct and the runtime NRE above is exactly what it prevents.

### 3. RETIRES WITH THE RESCOPE

- The **Question 0** finding — real `csc` emits the byte-identical `[ExpectedNumericType][found address of Int32]` for the equivalent C# `fixed` — killed the premise for the **array and span** forms. Dropping those hunks acts on it; it needs no further action beyond the known-limitation write-up.
- The **ilverify 1 → 2** result on the modified `Issue2906ExhaustiveSwitchReturnEmitTests` — removing the `ExpectedNumericType` suppression unmasked `UnmanagedPointer` and `StackByRef` that ilverify's first-error dataflow abort had been hiding. Applies only if the suppression removal ships. If the rescoped PR keeps the suppression, this is moot; if it removes it, this still blocks.
- **Question 1 (laundering vs. fixing)** and mutation **M5** — `AsPointer<object>` fed an `int32&`/`Pt&` produced ill-typed IL that still ran correctly, because the JIT ignores `T` for this intrinsic — applied to the array/span rewrite. Moot once those hunks are dropped.

### 4. UNCHANGED AND STILL LOAD-BEARING

The string hunk was always the defensible one, for the reason I raised against the brief: **gsc's old string path did not match Roslyn.** Roslyn lowers C# `fixed (char* p = s)` via `string.GetPinnableReference()` returning `char& modreq([System.Runtime]System.Runtime.InteropServices.InAttribute)`, **not** `OffsetToStringData`. The rewrite moves *toward* Roslyn, and the deleted comment warning that the encoder "cannot reproduce" that member-ref is now obsolete — I confirmed on the emitted assembly that the `modreq` **is** present, the locals **are** marked `pinned`, and a GC-stress probe over the pinned pointer is clean. Rescoping to string-only keeps the hunk that was actually justified.

DavidObando:
## Round-3 review — PR #2964 @ `0b34855f2` (base `177895195`)

**Verdict: MERGEABLE.**

Both recorded blockers are wrong, and so was my own first attempt at a smaller
fix. Details below, with the probes.

Head is unchanged since the two earlier rounds (`git rev-parse HEAD` =
`0b34855f26e7ce7f7ae2a619cf64f3a8c7ee2958`), so nothing here re-litigates work
the author has since done — there is none.

---

### Base→head differential (the issue's own repro, verbatim from #2922)

```
func F(x int32, xs []int32) int32 {
    unsafe { fixed p *int32 = xs { switch x { case 0 { return xs.Length } default { return 1 } } } }
}
```

| build | ilverify | runtime |
|---|---|---|
| base `177895195` | `1 Error(s)` — `ExpectedNumericType [offset 0x0000001F][found address of Int32]` | prints `2` |
| head `0b34855f2` | **All Classes and Methods Verified** | prints `2` |

Byte-for-byte the error text and offset that #2922 reports. The PR fixes the
filed defect. Build verified: exit 0, non-zero DLL, `GSharp.Core.dll` md5
`8af038e9…` identical across `out/bin/Release/{Compiler,Compiler.Tests,Core.Tests,Interpreter.Tests}`.

---

### BLOCKER 1 as recorded ("head makes verification strictly worse, 1 error → 2") — NOT SOUND

The 1→2 count came from probes that **dereference** the pointer. ilverify stops
reporting after the first fault in a method, so at base the pin's
`ExpectedNumericType` **masks** the dereference faults; at head the pin is clean
and they become visible. It is unmasking, not regression.

Proof — minimal C# with no `fixed` anywhere, compiled by `csc`:

```csharp
public static unsafe int M(int* p) { return *p; }
```

emits exactly `UnmanagedPointer @0x00` + `StackByRef @0x01` — the same pair
attributed to this PR. That is the irreducible ilverify signature of *any*
pointer dereference by *any* compiler. Head's dereference bytes (`ldloc; ldind`)
and local signatures are identical to base's; only the pin changed.

Counting distinct faults rather than reported ones: base 3, head 2.

### BLOCKER 2 as recorded ("`AsPointer` launders the managed pointer past the verifier") — TRUE, BUT NOT A DEFECT

Accurate as a description. It is also the only available option. There is no
verifiable IL sequence converting a managed pointer to `void*` — `conv.u` on a
managed pointer *is* `ExpectedNumericType`, i.e. exactly the error #2922 reports.
Roslyn emits `conv.u` and its `fixed` output is unverifiable too. So the choice is
"match Roslyn and keep the bug" or "route through a corelib intrinsic and close
it". The issue explicitly asked for the second ("Emitted fixed/pointer control
flow should satisfy the CLI verifier").

Worth saying out loud in the ADR, though: verifier-cleanliness here is obtained
by moving the unverifiable step behind a corelib call, not by making the IL more
type-safe. A reader should not take a green verifier on an `unsafe` block as
more than it is.

---

### Correction to my own analysis — a smaller shape was tried and it is a REVERT

I implemented the obvious lazier alternative: drop
`EmitManagedPointerAsUnmanagedPointer` and inline `this.il.OpCode(ILOpCode.Conv_u)`
at all three call sites. **+3/−17** against the PR's +40/−22. It built clean
(md5 `8af038e9…` → `e2e05923…` → `8af038e9…` on restore, so both builds were real)
and all three probes still ran correctly.

It initially read as *better* — 1 reported ilverify error against head's 2. That
reading was the very masking artifact I used to refute BLOCKER 1. On the issue's
own non-dereferencing repro:

| build | ilverify |
|---|---|
| base | `ExpectedNumericType @0x1F` |
| **alternative** | **`ExpectedNumericType @0x1F`** — identical |
| head | Verified |

The alternative reproduces #2922 verbatim. It is a revert, not a smaller fix, and
it fails 6 of the PR's 24 tests (`FixedSource_VerifiesLoadsAndRuns` ×3,
`FixedControlFlowMatrix_Verifies`, `SlicePin_…AsPointer`,
`ExhaustiveEnum_UnnamedValueWithFixedReturn_UsesFixedEpilogueDefault`). Recording
it so nobody re-derives and adopts it.

**On size:** the shape is right. One 14-line helper, three call sites — no
pre-existing managed→unmanaged conversion helper to reuse. Most of +40/−22 is
that helper plus deletion of a now-obsolete `modreq` comment. The string hunk is
not scope creep: the base string pin is *also* unverifiable
(`ExpectedNumericType [found ref 'string']`) via the pre-C#-7.3
`conv.i` + `OffsetToStringData` idiom, and head moves it to the same
`String::GetPinnableReference` shape Roslyn uses.

---

### Correctness checks that passed (IL decoded via `System.Reflection.Metadata`)

- **Array** — locals `07-04-08-45-1D-08-0F-08-08`; local 1 is `pinned int32[]`,
  stored at `IL_0002` (`ldarg.1; dup; stloc.1`) *before* `ldelema` `IL_001A` and
  `AsPointer` `IL_001F`, released `ldnull; stloc.1` in the finally. The pin covers
  the address's entire lifetime — **no GC hole**.
- **Null array** → `brfalse` at `IL_0003` → `ldc.i4.0; conv.u` (null pointer).
  **Empty array** → `ldlen; brtrue` → null pointer. Both match C#.
- **String** — locals `…45-0E…` (`pinned string`), `brfalse` null guard → null
  pointer; `""` goes through `GetPinnableReference` (ref to the terminator).
  Matches C#; the test covers `""`.
- Head runs correctly on all probes; the PR's own filter is **24 passed / 0 failed**.

---

### SHOULD-FIX — the Span case is still unverifiable and the test hides it broadly

Head, non-dereferencing span pin:

```
1 Error(s): StackUnexpected [offset 0x00000045][found Native Int][expected address of Int32]
```

from the unpin `IL_0043: ldc.i4.0; IL_0044: conv.u; IL_0045: stloc.2` into the
`pinned int32&` local (`45-10-08`). That is Roslyn's own unpin idiom and there is
no verifiable way to clear a pinned byref local, so **the residue itself is fine**.

The problem is how it is suppressed: `ignoredVerificationErrors: ["StackUnexpected"]`
in `PinKinds()`, and `IlVerifier.Verify` ignores **by error code across the whole
assembly**. That test would keep passing if the emitter later produced a genuinely
broken stack anywhere else in that program. Narrow it — assert the error count and
offset, or scope the ignore to the unpin site.

Related: the PR summary and ADR read as "fixed now verifies". One of the three
shapes does not. Please say which.

### NIT — brittle offset arithmetic in `SlicePin_ConvertsManagedPointerThroughUnsafeAsPointer`

`Issue2922FixedSwitchIlTests.cs:363-365` asserts `il[indexOf(ldelema) + 5] == call`,
hard-coding the instruction width. Any change to operand encoding, or one inserted
instruction, fails it as `Assert.Equal() Failure: Expected 40 Actual 224` — which is
exactly the unreadable failure it produced against my alternative. Decode the stream
instead; the file already imports `System.Reflection.Emit`.

---

### Merge order

`git merge-tree --write-tree 0b34855f2 pr2967` on the **natural** merge base
(`177895195`) → **exit 0, clean**. #2964's hunks are at `:97 / :128 / :203`;
#2967's at `:623 / :636`. They share `MethodBodyEmitter.Statements.cs` but sit
~420 lines apart. **No ordering constraint between #2964 and #2967**, and #2965 is
not blocked by textual contention in this file.

DavidObando:
### Addendum — re-checked against `main` = `e80132516` (post #2966/#3002/#2978/#2983)

**Verdict unchanged: MERGEABLE.** `git merge-tree --write-tree 0b34855f2 e80132516` → **exit 0, clean**. Nothing in the four merges disturbs this PR.

Two corrections/extensions to my merge-order note:

1. **There *is* one collision, and I understated it.** My earlier check was scoped to #2967. `#2964 × #2998` conflicts for real:

   ```
   CONFLICT (content): test/Compiler.Tests/Emit/Issue2906ExhaustiveSwitchReturnEmitTests.cs
   ```

   It is **test-only and one line** — `:314`, this PR's `ignoredVerificationErrors` swap. Whichever lands second takes the other's line. Not a blocker, but it should not surprise the merger. The #2967 finding stands: clean, hunks ~420 lines apart.

2. **My SHOULD-FIX survives #2966's merge.** I re-read the merged `IlVerifier.cs` on `e80132516` — `ignoredErrorCodes` still maps to ilverify's `-g`, which matches the **error category name assembly-wide** (`:104-119`). So the Span case's `["StackUnexpected"]` still suppresses that category across the whole program, and the fix I asked for is unchanged.

**Also worth recording:** the contended line is itself independent confirmation of the masking mechanism I described above. This PR had to change

```diff
-    ignoredVerificationErrors: new[] { "ExpectedNumericType" });
+    ignoredVerificationErrors: new[] { "UnmanagedPointer", "StackByRef" });
```

That is exactly the swap masking predicts: base's single pin fault hid the dereference faults, head's clean pin exposes them. The author's own diff records the same base→head transition my probes measured — which is the strongest available evidence that the recorded "1 error → 2 errors, strictly worse" blocker was reading an artefact.

It does sharpen point 1 of the SHOULD-FIX, though: `{ "UnmanagedPointer", "StackByRef" }` assembly-wide is a broader blanket than `{ "ExpectedNumericType" }` was — those two are the signature of *any* pointer dereference, so this test can no longer catch a genuinely malformed one anywhere in its program.

PR #2966, opened 2026-07-31
Prevent IlVerifier child process hangs

### LINKED ISSUE #2934: Compiler.Tests: IlVerifier can hang indefinitely waiting for child exit

DavidObando:
## Summary

- bound each `dotnet-ilverify` run to 30 seconds
- kill the entire child process tree on timeout, then bound both recovery waits and pipe drains
- include partial output when available and a clear marker when a surviving descendant retains a pipe
- keep stdout and stderr draining concurrently to preserve the existing full-pipe deadlock fix
- add direct timeout, escaped-descendant, and dual-pipe regression coverage

Fixes #2934

## Verification

- Timeout pin: a child writes to both pipes and sleeps; a 2-second timeout kills it and reports command, timeout, assembly path, and partial output.
- Escaped-descendant pin (Unix): a double-forked descendant retains stdout after tree kill; recovery returns a diagnostic with the unavailable-output marker in about 12 seconds. Windows skips this test because managed `Process` APIs cannot portably create a reparented descendant retaining redirected handles.
- Recovery mutation: restoring the three unbounded recovery waits makes the escaped-descendant pin fail after about 30 seconds; hardened code passes under its 15-second bound.
- Lifetime mutations: restoring unbounded primary `WaitForExit()` or plain `Kill()` makes the timeout pin fail in 6 seconds.
- Pipe guard: a child writes 128 KiB to both stdout and stderr; both complete and are returned byte-for-byte. Sequential-drain mutation hangs and was killed at 120 seconds.
- Release build: 0 warnings, 0 errors.
- Base `8dddbbf2`: Core 7028 passed + 1 skipped; Compiler 3893 passed; Interpreter 489 passed; LSP 452 passed.
- Head: Core 7028 passed + 1 skipped; Compiler 3896 passed; Interpreter 489 passed; LSP 452 passed.

No corpus run: this changes only `Compiler.Tests` process lifetime management, not compiler production code. No local e2e run per repository task constraints.

### DIFF
--- a/test/Compiler.Tests/IlVerifier.cs
+++ b/test/Compiler.Tests/IlVerifier.cs
@@ -9,6 +9,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Text;
+using System.Threading.Tasks;
 using Xunit.Sdk;
 
 namespace GSharp.Compiler.Tests;
@@ -35,11 +36,16 @@ namespace GSharp.Compiler.Tests;
 ///   <item>If the <c>ilverify</c> tool cannot be located, the helper throws
 ///         (instead of silently passing) so CI failures are obvious. To bypass
 ///         in environments without the tool, set <c>GSHARP_SKIP_ILVERIFY=1</c>.</item>
+///   <item>Verification is limited to 30 seconds. On timeout, the helper kills
+///         the tool process tree and reports any partial stdout and stderr.</item>
 /// </list>
 /// </summary>
 internal static class IlVerifier
 {
     private const string SkipEnvVar = "GSHARP_SKIP_ILVERIFY";
+    private const int VerifyTimeoutMilliseconds = 30_000;
+    private const int KillGraceMilliseconds = 10_000;
+    private const string UnavailableOutput = "<unavailable: output pipe held by a surviving descendant>";
 
     private static readonly object ToolLocateSync = new();
     private static readonly Lazy<IReadOnlyList<string>> RuntimeReferences =
@@ -130,21 +136,10 @@ internal static class IlVerifier
             psi.ArgumentList.Add(a);
         }
 
-        using var proc = Process.Start(psi)
-            ?? throw new XunitException($"ilverify: failed to start '{command}'");
+        var (exitCode, stdout, stderr) =
+            RunProcess(psi, assemblyPath, VerifyTimeoutMilliseconds);
 
-        // Drain both pipes concurrently before waiting: reading stdout to EOF
-        // and only then stderr can deadlock when ilverify fills the stderr pipe
-        // buffer (~64 KB of error lines) while we are still blocked on stdout,
-        // leaving the child unable to exit. Awaiting both reads together avoids
-        // the classic redirected-pipe deadlock.
-        var stdoutTask = proc.StandardOutput.ReadToEndAsync();
-        var stderrTask = proc.StandardError.ReadToEndAsync();
-        proc.WaitForExit();
-        var stdout = stdoutTask.GetAwaiter().GetResult();
-        var stderr = stderrTask.GetAwaiter().GetResult();
-
-        if (proc.ExitCode != 0)
+        if (exitCode != 0)
         {
             // ilverify writes one line per IL error. Surface its full output so
             // a CI failure points directly at the offending opcode/method.
@@ -152,7 +147,7 @@ internal static class IlVerifier
                 .Append("ilverify detected invalid IL in '")
                 .Append(assemblyPath)
                 .Append("' (exit ")
-                .Append(proc.ExitCode)
+                .Append(exitCode)
                 .AppendLine("):")
                 .AppendLine(stdout.TrimEnd())
                 .Append(stderr.TrimEnd())
@@ -167,6 +162,63 @@ internal static class IlVerifier
                 $"ilverify exited successfully without confirming verification of '{assemblyPath}'.{Environment.NewLine}" +
                 stdout.TrimEnd() + Environment.NewLine + stderr.TrimEnd());
         }
+    }
+
+    internal static (int ExitCode, string Stdout, string Stderr) RunProcess(
+        ProcessStartInfo startInfo,
+        string assemblyPath,
+        int timeoutMilliseconds)
+    {
+        using var proc = Process.Start(startInfo)
+            ?? throw new XunitException($"ilverify: failed to start '{startInfo.FileName}'");
+
+        // Drain both pipes concurrently before waiting: reading stdout to EOF
+        // and only then stderr can deadlock when ilverify fills the stderr pipe
+        // buffer (~64 KB of error lines) while we are still blocked on stdout,
+        // leaving the child unable to exit. Awaiting both reads together avoids
+        // the classic redirected-pipe deadlock.
+        var stdoutTask = proc.StandardOutput.ReadToEndAsync();
+        var stderrTask = proc.StandardError.ReadToEndAsync();
+        if (!proc.WaitForExit(timeoutMilliseconds))
+        {
+            proc.Kill(entireProcessTree: true);
+            proc.WaitForExit(KillGraceMilliseconds);
+
+            // Observation prevents a later unobserved-task escalation when a
+            // surviving descendant eventually closes a pipe with an error.
+            // It does not change task state: Wait/Result below still throw any
+            // fault that completes within the grace period.
+            _ = stdoutTask.ContinueWith(
+                task => _ = task.Exception,
+                TaskContinuationOptions.OnlyOnFaulted);
+            _ = stderrTask.ContinueWith(
+                task => _ = task.Exception,
+                TaskContinuationOptions.OnlyOnFaulted);
+
+            string Grab(Task<string> task) => task.Wait(KillGraceMilliseconds)
+                ? task.Result
+                : UnavailableOutput;
+            var timedOutStdout = Grab(stdoutTask);
+            var timedOutStderr = Grab(stderrTask);
+            var message = new StringBuilder()
+                .Append("process '")
+                .Append(startInfo.FileName)
+                .Append("' timed out after ")
+                .Append(timeoutMilliseconds)
+                .Append(" ms while verifying '")
+                .Append(assemblyPath)
+                .AppendLine("':")
+                .AppendLine("stdout:")
+                .AppendLine(timedOutStdout.TrimEnd())
+                .AppendLine("stderr:")
+                .Append(timedOutStderr.TrimEnd())
+                .ToString();
+            throw new XunitException(message);
+        }
+
+        var stdout = stdoutTask.GetAwaiter().GetResult();
+        var stderr = stderrTask.GetAwaiter().GetResult();
+        return (proc.ExitCode, stdout, stderr);
     }
 
     /// <summary>
--- a/test/Compiler.Tests/IlVerifierTests.cs
+++ b/test/Compiler.Tests/IlVerifierTests.cs
@@ -5,6 +5,7 @@
 #nullable enable
 
 using System;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection.Metadata;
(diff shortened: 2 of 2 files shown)


DavidObando:
## Rubber-duck verdict (Opus 5): MERGEABLE — one strong SHOULD-FIX, no blockers

Reviewed `92fff160` rebased onto current `origin/main` `8dddbbf2` → `3c43bf35`. Rebase clean, delta identical (`IlVerifier.cs +63/−15`, `IlVerifierTests.cs +64/−0`).

### Central risk — can a timeout or kill path report success? No, proven exhaustively

Every exit from `Verify` + `RunProcess` (`IlVerifier.cs:71-203`):

```
:78   return;                       ← GSHARP_SKIP_ILVERIFY=1, pre-existing, unchanged
:83   throw XunitException          ← assembly missing
:152  throw XunitException          ← exitCode != 0
:158  throw XunitException          ← success marker absent
:170  throw XunitException          ← Process.Start null
:197  throw XunitException          ← TIMEOUT
:202  return (proc.ExitCode, ...)   ← only success path
```

**Zero `catch` blocks** in either method. The timeout throws unconditionally, before any tuple can be constructed. No exit-code conflation is possible — the killed child's exit code is never read.

`ignoredErrorCodes` is **byte-for-byte unchanged**. Filtering the full diff to verification semantics yields only:

```
-        if (proc.ExitCode != 0)        +        if (exitCode != 0)
-                .Append(proc.ExitCode) +                .Append(exitCode)
```

`-g` construction, reference building, the success marker, the skip env var, `LocateTool` and `KnownIssues` are all untouched. Public surface is source-compatible; the only addition is `internal static RunProcess`. #2912/#2929/#2961/#2964 are unaffected and none of them touch `IlVerifier*.cs`.

Interaction with #2964: none, and directionally safe — this change can only make `Verify` fail *more* (on timeout), never pass more, so it cannot manufacture false ilverify evidence.

### Mutation table (build-green only, unmutated control first, rebuild between probes, restored after)

| # | mutation at `IlVerifier.cs` | result | detector |
|---|---|---|---|
| — | control (unmutated) | 7/7 pass, 4 s | — |
| M1 | drain reordered → sequential synchronous `ReadToEnd()` before wait | **run hung, killed at 120 s** | `RunProcess_LargeStdoutAndStderr_DrainsBothPipes` |
| M2 | `WaitForExit(timeout)` → unbounded `WaitForExit()` (pre-fix shape) | failed 1/7 @ 6 s | `RunProcess_HungChild_TimesOutWithPartialOutput` |
| M3 | `Kill(entireProcessTree: true)` → `Kill()` | failed 1/7 @ 6 s | same pin |

Both new tests are load-bearing. The 128 KiB guard genuinely fails when draining is reordered: the child (`sh -c 'awk >stdout; awk >stderr'`) holds the stdout write handle until the second `awk` exits, and that `awk` blocks on a full 64 KiB stderr pipe — the classic deadlock. Correctly constructed.

### SHOULD-FIX — `IlVerifier.cs:181-185`, the recovery path has three unbounded waits

```csharp
proc.Kill(entireProcessTree: true);
proc.WaitForExit();                                        // :182  unbounded
var timedOutStdout = stdoutTask.GetAwaiter().GetResult();  // :184  unbounded
var timedOutStderr = stderrTask.GetAwaiter().GetResult();  // :185  unbounded
```

Pipe reads block until *every* holder of the write handle closes it, not until the killed child dies. Direct probe with a child that double-forks a reparented descendant holding stdout open:

```
case2: after Kill(tree), proc.WaitForExit() returned at 2.0s
case2: stdoutTask completed within 20s? False  elapsed=22.0s
case2: *** GetResult() WOULD BLOCK -> timeout handler hangs ***
```

Three reasons to fix it in this PR:

- **The production topology is already parent→grandchild.** `IlVerifier.cs:318-325` prefers `dotnet tool run ilverify`, so the pipe-holding process is a grandchild and the bound depends entirely on tree-kill reaching it.
- **No framework safety net exists.** `test/Compiler.Tests/AssemblyInfo.cs:9` sets `[assembly: CollectionBehavior(DisableTestParallelization = true)]`, and xUnit 2.9.2 does not honour `[Fact(Timeout=…)]` when parallelization is disabled. `RunProcess` must self-bound on every path.
- **#2934 exists because we do not know why `ilverify` hangs.** Assuming its process topology is well-behaved is the assumption that produced this issue.

Measured 5-line fix, behaviour-preserving:

```csharp
proc.Kill(entireProcessTree: true);
proc.WaitForExit(KillGraceMilliseconds);
string Grab(Task<string> t) => t.Wait(KillGraceMilliseconds)
    ? t.Result : "<unavailable: output pipe held by a surviving descendant>";
// plus ContinueWith(t => _ = t.Exception, TaskContinuationOptions.OnlyOnFaulted) on both tasks
```

| case | shipped | hardened |
|---|---|---|
| A — pin shape (tree-killable) | 2.2 s, partial output present | 2.2 s, partial output present (identical) |
| B — escaped descendant | **blocks ≥22 s, unbounded** | **12.5 s bounded**, clear diagnostic |
| C — 128 KiB both pipes | 131072/131072, exit 0 | 131072/131072, exit 0 (identical) |

Not a blocker, since the base was strictly worse (unbounded on the *common* path), but this file is reserved to this PR right now and the fix is free.

### Timeout value — keep 30 s

Measured on a loaded machine (10 concurrent agents), real `dotnet tool run ilverify` with ~200 references:

| workload | wall time |
|---|---|
| cold first invocation, `GSharp.Core.dll` (2.4 MB) | **4.10 s** |
| warm sequential ×8, `GSharp.Core.dll` | 0.42 – 0.65 s |
| warm ×4, `GSharp.Compiler.Tests.dll` (4.1 MB, largest in repo) | 0.37 – 0.51 s |
| 10-way concurrent, `GSharp.Core.dll` | 0.84 – **1.16 s** max |

Worst observed is 4.1 s on a one-time cold start, and `LocateTool`'s own probe absorbs most of that — outside the timeout. Headroom is ≈7× on the cold outlier and ≈26× on the concurrent maximum. `Compiler.Tests` is serial, so ilverify calls do not self-contend. The concern that "ilverify over a large assembly" is heavier than "a small compiled program" is false: the largest assembly in the repo verifies in 0.4 s. Precedent supports 30 s as the house default (~30 sites), with 60 s and 180 s used where a step is genuinely heavier.

### Suite numbers, restated against a measured base

| | base `8dddbbf2` | head `3c43bf35` | Δ |
|---|---|---|---|
| Compiler.Tests total | **3893** | **3895** | **+2** |
| passed | 3874 | 3876 | +2 |
| failed | 19 | 19 | 0 |

**`8dddbbf2` = Compiler 3893**; the `17789519` figure of 3876 is stale by +17 (#2932/#2955/#2952). The PR's "+2" is correct; its "3876 → 3878" framing is against the old base. The 19 failures are identical on base and head and environmental — all `SampleConformanceTests.Sample_BuildsAndMatchesGolden` failing with `imports Gsharp.Extensions but Gsharp.Extensions.dll was not found in any expected location`.

Correctly no corpus run: zero production code.

### Merge safety

Clean against `origin/main` `8dddbbf2` and against #2912, #2929, #2961, #2964 — zero conflicts, and none of the four touch `IlVerifier*.cs`.

### NITs

1. `IlVerifier.cs:189` — the timeout message hard-codes `"ilverify timed out"`, but `RunProcess` is now a generic helper the tests drive with `/bin/sh`. Harmless, since the only production caller is ilverify.
2. `RunProcess_LargeStdoutAndStderr_DrainsBothPipes` detects a drain-ordering regression by **hanging** rather than failing — which is #2934's own symptom. It cannot be fixed with `[Fact(Timeout=)]` given `DisableTestParallelization = true`, and the SHOULD-FIX bound does not cover a regression placed *before* the wait. Worth a comment in the test recording the limitation.

- `IlVerifierTests.cs` is **not** a new file — it exists at `17789519` with 152 lines and 5 facts; this PR adds 2.
- The "is 30 s enough?" concern is not supported by measurement (above).
- House precedent for `WaitForExit(30_000)` is ~30 sites, not one.
- One reviewer concern was raised and then **retracted**: `Kill(entireProcessTree: true)` on a process that exits inside the timeout race does *not* throw, so the partial-output message is not lost.

DavidObando:
## Review verdict: **MERGE NOW** — the residual defect at `:390`/`:427` is a **NIT**, not a gate

Reviewed `24f6e6ce2` (== `refs/pull/2966/head`, confirmed against GitHub — SHAs are content-addressed, so that is a complete provenance proof). Scratch clone, merge-base `8dddbbf2`. Content confirmed before probing: `IlVerifier.cs` **+67/−15**, `IlVerifierTests.cs` **+119 new**. Test-only; no `src/` changes.

---

### Q4 first — the base→head differential, because it licenses everything else

I extracted the **merge-base** and **head** `RunProcess` shapes verbatim into a standalone harness and ran both against a child that writes to both pipes and then hangs (`printf hung-stdout; printf hung-stderr >&2; sleep 30`):

| shape | result |
|---|---|
| **base** `8dddbbf2` | ***WEDGED*** — no return; killed by an outer 20 s timeout. Unbounded, would hang forever. |
| **head** `24f6e6ce2` | `TIMEOUT-HANDLED elapsed=2220ms out=11 err=11` — bounded, killed, **both partial outputs recovered** |

Perturb base→head, output flips; the shapes differ only in the PR's own change. The fix works, and this was not a stale binary.

**Briefing correction, and it matters for what the pins prove:** the concurrent `ReadToEndAsync()` drain on both pipes **already existed at the merge-base** — `IlVerifier.cs` at `8dddbbf2` reads

```csharp
var stdoutTask = proc.StandardOutput.ReadToEndAsync();
var stderrTask = proc.StandardError.ReadToEndAsync();
proc.WaitForExit();                      // <- unbounded; the ONLY defect here
```

So this PR does not add the concurrent drain. What it adds is the **bound**: `WaitForExit(timeoutMilliseconds)`, `Kill(entireProcessTree: true)`, a `KillGraceMilliseconds` grace, and fault observation on both tasks. Consequently `RunProcess_LargeStdoutAndStderr_DrainsBothPipes` **would also pass at base** — the discriminating pins are `RunProcess_HungChild_TimesOutWithPartialOutput` and `RunProcess_EscapedDescendantHoldingPipe_TimeoutRemainsBounded`. Measured below.

---

### Q1 — reachability of the two probe sites

`LocateTool()` (`IlVerifier.cs:320-361`) caches under a lock, so each probe runs **at most once per test process**, before any test body.

| site | function | reachable in CI? |
|---|---|---|
| **`:390`** | `TryProbeDotnetToolRun` | **YES** — step 1, always runs on the first `Verify()` call. The comment at `:337` says so: *"This is what CI restores via `dotnet tool restore`."* |
| **`:427`** | `TryProbeOnPath` | **NO** — only reached when probe 1 returns non-zero or throws. With the manifest restored it never executes. Dev-machine-only. |

### Q2 — I wedged the shape, but **not on any reachable input**

The probe shape is genuinely defective. Same harness, sequential `ReadToEnd()` stdout-then-stderr, child writing 128 KB to each pipe:

| shape | 128 KB + 128 KB |
|---|---|
| **probe** (`:390`/`:427`) | ***WEDGED*** — permanent, killed by outer 25 s timeout |
| base | `OK elapsed=92ms out=131072 err=131072` |
| head | `OK elapsed=106ms out=131072 err=131072` |

So the classic deadlock is real in the code shape and the main path is immune to it in both directions.

**But the reachable inputs are three orders of magnitude too small.** Both probes run a fixed `--version`. Measured directly:

| path | exit | stdout | stderr | probe shape |
|---|---|---|---|---|
| `dotnet tool run ilverify --version`, manifest resolves | 0 | **68 B** | **0 B** | `OK, 34 ms` |
| `dotnet tool run ilverify --version`, no manifest entry | 1 | 0 B | **77 B** | `OK, 227 ms` |
| `ilverify --version` on PATH | 0 | **68 B** | 0 B | — |

Worst reachable case is **77 bytes against a ~64 KB buffer — an ~850× margin**, on a fixed argument vector with no user-controlled input. By the stated criterion — *"if the child's total output is bounded well under the pipe buffer on every reachable path, it is a NIT"* — **this is a NIT.**

### Q3 — the unbounded `WaitForExit()` is a red herring

Bounding it would fix nothing. In both probes the **blocking call that comes first is `ReadToEnd()`**, and it is also unbounded:

```csharp
proc.StandardOutput.ReadToEnd();   // :390 / :427 — blocks until stdout EOF. Unbounded.
proc.StandardError.ReadToEnd();    // :391 / :428
proc.WaitForExit();                // :392 / :429 — never reached if the read blocks
```

`ReadToEnd()` returns only at stdout EOF, which requires the child **and every descendant holding the inherited handle** to close it. A hung `ilverify` holds it forever and `WaitForExit()` is never reached. So a fix must bound the **reads**, not the wait — any patch that only adds a timeout to `WaitForExit()` is a no-op for the hang it is meant to prevent.

Is there a bound in practice? Only informally: `dotnet tool run … --version` terminated in 34–227 ms across every state I could produce, including cold and missing-manifest. I could not construct a hang from a reachable input.

**Second briefing correction:** the deadlock comment at `:172-177` sits immediately above the **correct** concurrent drain, not above the offending code. The sequential reads are ~215 lines further down at `:390`/`:427`.

---

### The pins are good

8 new `[Fact]`s, including the three that matter, all bounded:

| test | bound | measured on this machine (load 39) |
|---|---|---|
| `RunProcess_HungChild_TimesOutWithPartialOutput` | asserts `< 5 s` on a 2 s timeout | my equivalent completed in **2.22 s** |
| `RunProcess_LargeStdoutAndStderr_DrainsBothPipes` | 128 KB each pipe — 2× the buffer | **106 ms** |
| `RunProcess_EscapedDescendantHoldingPipe_TimeoutRemainsBounded` | asserts `< 15 s` | — |

Every `WaitForExit` in both files is bounded except the two probes, and the test file's own `escaped.WaitForExit(5_000)` at `IlVerifierTests.cs:236` is bounded — it does not repeat #2934's shape. The author explicitly acknowledges the `DisableTestParallelization` interaction at `IlVerifierTests.cs:177-180`.

---

### Findings

| sev | `file:line` | finding |
|---|---|---|
| **NIT** | `IlVerifier.cs:390-392` | sequential `ReadToEnd()` + unbounded `WaitForExit()`. Shape is wedgeable at 128 KB; reachable inputs are ≤ 77 B (~850× margin). Reachable in CI but harmless. **Pre-existing — byte-identical at merge-base, untouched by this PR.** |
| **NIT** | `IlVerifier.cs:427-429` | same shape; **not reachable in CI at all** |
| **NIT** | `IlVerifierTests.cs:220` | `Assert.True(stopwatch.Elapsed < TimeSpan.FromSeconds(15))` on the escaped-descendant test. Worst case is ~2 s timeout + `KillGraceMilliseconds` (10 s) for the stdout `Grab` ≈ **12 s**, a 3 s margin. Thin under CI load; consider 20 s. |

**No BLOCKER and no SHOULD-FIX.** Both probe sites are **pre-existing and untouched** — `TryProbeDotnetToolRun` and `TryProbeOnPath` are byte-identical between `8dddbbf2` and `24f6e6ce2`, and the PR's last hunk ends at ~`:224`, ~140 lines above them. This PR strictly improves the file and introduces no new hang surface.

### Recommendation

**Merge now; file the residual as a follow-up.** It is pre-existing, it is a NIT on the evidence, and #2966 is rank 1 gating ~12 downstream call sites — holding it costs far more than the risk it retains. The follow-up patch, for whoever takes it (the same shape already proven correct at `:179-185`):

```csharp
var stdoutTask = proc.StandardOutput.ReadToEndAsync();
var stderrTask = proc.StandardError.ReadToEndAsync();
if (!proc.WaitForExit(ProbeTimeoutMilliseconds))
{
    proc.Kill(entireProcessTree: true);
    proc.WaitForExit(KillGraceMilliseconds);
    _ = stdoutTask.ContinueWith(t => _ = t.Exception, TaskContinuationOptions.OnlyOnFaulted);
    _ = stderrTask.ContinueWith(t => _ = t.Exception, TaskContinuationOptions.OnlyOnFaulted);
    args = Array.Empty<string>();
    return false;   // probe failure — fall through to the next probe
}
```

Note it must bound the **reads**, per Q3.

### Not measured

I did **not** run the four suites. This is test-only with CI green (8 SUCCESS / 2 SKIPPED), machine load is 39 with 80 dotnet processes, and my differential is on the actual code shape, which answers the question asked more directly than a suite count. Arithmetic expectation only, unverified: Compiler **3906 → 3914** (+8 `[Fact]`), other three suites unchanged at Core 7028+1 skip · Interpreter 491 · LSP 452.

Issue #2965, opened 2026-07-31
gsc: user-defined struct channel values emit invalid IL


DavidObando:
## Reproduction

```gsharp
package Repro
import System
import Gsharp.Extensions.Go

data struct Pair(Value int32)

func Run() int32 {
    let ch = make(chan Pair, 1)
    ch <- Pair(5)
    var result = 0
    select { case let value = <-ch { result = value.Value } }
    return result
}

Console.WriteLine(Run())
```

At `17789519599bf446ec274bb20bfa29759cf77acc`, compilation succeeds but child-process execution throws:

```
System.InvalidProgramException: Common Language Runtime detected an invalid program.
```

The same failure occurs when `Run` is async. Imported CLR structs such as `DateTime` do not have this failure.

Expected: print `5`, matching primitive, reference, and imported-struct channel element types.

Found while testing issue #2933's async select receive binding. This failure reproduces in the synchronous control and is independent of the async state-machine field synchronization fix.

DavidObando:
# Design for #2965 — executable, verified at `main` = `a6b1561f2`

Read-only design pass in a scratch clone. `gsc` built with `dotnet build src/Compiler/Compiler.csproj -c Release -m:1` (exit 0, `GSharp.Core.dll` 2,471,936 bytes, md5 `2de210cf526fe6961e0f19252383cb4f`, non-zero). All probes are direct `gsc` invocations plus bounded `timeout 30 dotnet <exe>` execution.

## Q5 first — sequencing, since it gates whether this can start today

**It cannot be confined to `MethodBodyEmitter.cs` + `Calls.cs`. But the block is nominal, not structural.**

The fix touches **five** production files, not three:

| file | contested by | delta shape |
|---|---|---|
| `Emit/MethodBodyEmitter.cs` | — | helper rewrite |
| `Emit/MethodBodyEmitter.Calls.cs` | — | 3 sites |
| `Emit/SlotPlanner.cs` | — | 3 sites |
| `Emit/SignatureEncoder.cs` | — | 1 site |
| `Emit/MethodBodyEmitter.Statements.cs` | **#2964, #2967** | 4 sites |

**Four of the five files are completely uncontested.** I checked: #2964's only production file is `MethodBodyEmitter.Statements.cs`; #2967's only production file is `MethodBodyEmitter.Statements.cs`. Nothing else overlaps.

[...]

PR #2967, opened 2026-07-31
gsc: preserve async select receive bindings

### LINKED ISSUE #2933: gsc: select inside async function loses received arm value
## Summary

A `select` receive arm inside an `async func` loses the received value when evaluating the arm body. The equivalent synchronous function is correct.

This reproduces on both `main` (`d4038f1d`) and PR #2917 HEAD (`eb38ef4e`). It is independent of `fixed`; wrapping the select in `fixed` does not cause the wrong result.

## Repro

```gsharp
package AsyncSelectBinding

import System

async func Broken() int32 {
    let ch = make(chan int32, 1)
    ch <- 3
    var result = 0
    select {
        case let value = <-ch {
            result = 7 + value
        }
    }
    return result
}

func Control() int32 {
    let ch = make(chan int32, 1)
    ch <- 3
    var result = 0
    select {
        case let value = <-ch {
            result = 7 + value
        }
    }
    return result
}

Console.WriteLine(Broken().GetAwaiter().GetResult())
Console.WriteLine(Control())
```

## Actual

```text
7
10
```

## Expected

```text
10
10
```

The async state-machine path appears to execute the selected arm but reads the arm binding as its default value (`0`). Please add an emitted-runtime regression comparing async and synchronous select receive bindings.

DavidObando:
## Summary

- synchronize selected receive-arm out slots into hoisted async-function and async-iterator state-machine fields before emitting arm bodies
- cover successful reads and closed-channel zero-value paths
- preserve synchronous selects, send/default arms, and sibling binding forms
- add emitted child-process runtime coverage across state-machine kinds; retain interpreter behavior as a guard

Fixes #2933

## Confirmed mechanism

`AsyncCaptureWalker` hoists a receive-arm variable when the arm body references it, and state-machine rewriting changes body reads to field reads. `EmitSelectReceiveProbe` instead passed the ordinary local slot directly to `ChannelReader<T>.TryRead(out T)`, then emitted the rewritten body without copying that value to the state-machine field. The arm ran, but body reads saw the field's CLR default.

Async functions use `asyncFieldMap`; async iterators use `iteratorEmitCtx.FieldMap`. The original fix handled only the first state-machine kind. The final fix follows the existing `EmitCapturedVariableLoad` pattern, resolves either map, and emits `ldarg.0`, `ldloc`, `stfld` through `ResolveCurrentStateMachineFieldToken`. `EmitStoreVariable` is not used because it redirects only through `asyncFieldMap`.

The same field transfer runs after a closed channel zero-initializes the out slot, preventing a loop from reusing a previous received value.

## Production delta

- `MethodBodyEmitter.Statements.cs`: +20 / -2

No state-machine layout or capture-analysis changes. PR #2950's struct-`this` handling remains untouched.

## Runtime proof

Known-bad base compiler (`17789519`): emitted async function stdout `7`, while synchronous and interpreted controls produce `10`.

Final head:

- async function: `10`
- async iterator yielding inside the receive arm: `10`
- async iterator computing in the arm and yielding afterward: `10`
- synchronous control: `10`
- assemblies load and `GetTypes()` succeeds before bounded child execution

The async-iterator regression compares state-machine kinds directly. Removing the iterator field-map alternative changes its exact output from `10,10,10` to `10,7,7`.

Both stdout/stderr pipes are drained before the bounded wait; timeout kills the full process tree.

## Binding-form and state-machine-kind axes

Binding-form audit:

- `for ... in` bindings lower through ordinary assignments/stores
- type/list pattern bindings call `EmitStoreVariable`
- catch bindings call `EmitStoreVariable`
- G# plain-switch binding is pattern-based
- send/default/discard select arms carry no receive variable

State-machine-kind coverage now includes async functions and async iterators. Merging this PR leaves the pre-existing synchronous-iterator output at `1,7` when the binding is read across a yield; this is #2975, not a regression from this change. `IteratorRewriter.cs`'s declaration-based `LocalCollector` does not recognize bare `BoundSelectCase.Variable` bindings, while the async path uses the reference-based `AsyncCaptureWalker.ReferenceCollector`. `HoistedFieldRewriter` only consumes the resulting field map; it does not build it. #2982 was closed as a duplicate. Once #2975 supplies the iterator field, this emitter transfer remains necessary to write the received value into it.

## Pins

- buffered async receive parity with synchronous execution
- async iterator yielding inside the arm
- async iterator yielding after the arm
- multiple receive arms
- select in a loop
- nested select
- select inside an async lambda
- receive binding captured by a closure
- binding read before and after `await`
- nullable reference and imported CLR struct bindings
- unbuffered channel
- closed-channel zero reset after a prior received value

## Guards

- synchronous receive binding
- default and send arms
- async range, pattern, and catch bindings
- assembly load/type discovery and exact child stdout
- interpreter remains correct at 10; it evaluates the bound tree and never exercised the faulty IL path

## Mutations

Three build-green semantic mutations were killed, restoring and rebuilding between probes:

1. remove ready-read transfer: 8 of the original 10 Compiler cases fail
2. remove closed-channel transfer: stale-value case alone fails (`33` instead of `30`)
3. remove async-iterator field-map alternative: targeted control 11/11; mutation 10/11, with state-machine-kind output `10,7,7`

Interpreter 1/1 stays green under emitter mutations, confirming it is a guard rather than a pin.

## Corpus

All 148 sample programs were compiled with base and head compilers: 0 exit-code drift, 0 diagnostic drift, and 0 SHA-256 output-binary drift.

- samples containing `select`: 3
- samples containing `select` inside async: **0**
- samples containing `select` inside iterators: **0**

Corpus therefore supplies synchronous non-regression evidence only; emitted-runtime tests carry the state-machine evidence.

## Executed suites

Final base `210126dd90f4ab3edbe1b7f9219946594685afc7`:

- Core: 7029 passed, 1 skipped
- Compiler: 3952 passed
- Interpreter: 494 passed
- LSP: 452 passed

Final head `323fc5c71fa0b817cebb617e735d95c8e957265e` in green CI:

- Core: 7029 passed, 1 skipped
- Compiler: 3963 passed
- Interpreter: 495 passed
- LSP: 452 passed

Delta: +11 Compiler and +1 Interpreter.

Post-rebase targeted validation:

- `Select|Issue2933|Iterator|Async|Yield|Sequence`: 479 passed
- Compiler Issue2933 pins: 11 passed
- Interpreter Issue2933 guard: 1 passed

Release solution build: 0 warnings, 0 errors. Ordinary `GSharp.Core.dll` outputs were nonempty and byte-identical after propagation.

## Independent findings

User-defined G# struct channel elements produce `InvalidProgramException` in synchronous and async selects independently of state machines; tracked by #2965.

Synchronous iterator select bindings live across yield lose their value because the binding is not hoisted; tracked by #2975.

### DIFF
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Statements.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Statements.cs
@@ -641,7 +641,7 @@ internal sealed partial class MethodBodyEmitter
         this.il.OpCode(ILOpCode.Callvirt);
         this.il.Token(this.outer.memberRefs.GetMethodReference(tryRead));
         this.il.Branch(ILOpCode.Brfalse, closedLabel);
-        this.EmitStatement(arm.Body);
+        this.EmitSelectReceiveArmBody(arm, slots.OutSlots[index]);
         this.il.Branch(ILOpCode.Br, endLabel);
 
         this.il.MarkLabel(closedLabel);
@@ -654,9 +654,24 @@ internal sealed partial class MethodBodyEmitter
         this.il.Token(this.outer.memberRefs.GetMethodReference(isCompleted));
         this.il.Branch(ILOpCode.Brfalse, nextLabel);
         this.EmitZeroInit(slots.OutSlots[index], chType.ElementType, elementClr);
-        this.EmitStatement(arm.Body);
+        this.EmitSelectReceiveArmBody(arm, slots.OutSlots[index]);
         this.il.Branch(ILOpCode.Br, endLabel);
         this.il.MarkLabel(nextLabel);
+    }
+
+    private void EmitSelectReceiveArmBody(BoundSelectCase arm, int outSlot)
+    {
+        if (arm.Variable != null
+            && this.asyncFieldMap != null
+            && this.asyncFieldMap.TryGetHoistedField(arm.Variable, out _))
+        {
+            // TryRead writes through the local out slot; the rewritten
+            // MoveNext arm body reads the hoisted state-machine field.
+            this.il.LoadLocal(outSlot);
+            this.EmitStoreVariable(arm.Variable);
+        }
+
+        this.EmitStatement(arm.Body);
     }
 
     private void EmitSelectSendProbe(
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Statements.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Statements.cs
@@ -661,14 +661,17 @@ internal sealed partial class MethodBodyEmitter
 
     private void EmitSelectReceiveArmBody(BoundSelectCase arm, int outSlot)
     {
+        FieldSymbol hoistedField = null;
         if (arm.Variable != null
-            && this.asyncFieldMap != null
-            && this.asyncFieldMap.TryGetHoistedField(arm.Variable, out _))
+            && ((this.asyncFieldMap != null && this.asyncFieldMap.TryGetHoistedField(arm.Variable, out hoistedField))
+                || (this.iteratorEmitCtx != null && this.iteratorEmitCtx.FieldMap.TryGetValue(arm.Variable, out hoistedField))))
         {
             // TryRead writes through the local out slot; the rewritten
             // MoveNext arm body reads the hoisted state-machine field.
+            this.il.OpCode(ILOpCode.Ldarg_0);
             this.il.LoadLocal(outSlot);
-            this.EmitStoreVariable(arm.Variable);
+            this.il.OpCode(ILOpCode.Stfld);
+            this.il.Token(this.ResolveCurrentStateMachineFieldToken(hoistedField));
         }
 
         this.EmitStatement(arm.Body);
--- a/test/Compiler.Tests/Emit/Issue2933AsyncSelectArmBindingTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2933AsyncSelectArmBindingTests.cs
@@ -1,0 +1,360 @@
+// <copyright file="Issue2933AsyncSelectArmBindingTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using GSharp.Compiler;
+using Xunit;
+
+namespace GSharp.Compiler.Tests.Emit;
+
+/// <summary>
+/// Issue #2933: select receive bindings inside async state machines retain
+/// their received values.
+/// </summary>
+public class Issue2933AsyncSelectArmBindingTests
+{
+    /// <summary>Gets requested async-select runtime shapes.</summary>
+    public static IEnumerable<object[]> SelectShapes()
+    {
+        yield return Case("MultipleAndDefault", "14,5\n", """
+            package Issue2933.MultipleAndDefault
+            import System
+            import Gsharp.Extensions.Go
+
+            async func Run() string {
+                let empty = make(chan int32, 1)
+                let ready = make(chan int32, 1)
+                ready <- 4
+                var selected = 0
+                select {
+                    case let ignored = <-empty { selected = -100 }
+                    case let value = <-ready { selected = 10 + value }
+                    default { selected = -200 }
+                }
+
+                var fallback = 0
+                select {
+                    case let ignored = <-empty { fallback = -100 }
+                    default { fallback = 5 }
+                }
+                return selected.ToString() + "," + fallback.ToString()
+            }
+            Console.WriteLine(Run().GetAwaiter().GetResult())
+            """);
+        yield return Case("SendArm", "10\n", """
+            package Issue2933.SendArm
+            import System
+            import Gsharp.Extensions.Go
+
+            async func Run() int32 {
+                let ch = make(chan int32, 1)
+                var result = 0
+                select {
+                    case ch <- 3 { result = 7 }
+                    default { result = -100 }
+                }
+                return result + <-ch
+            }
+            Console.WriteLine(Run().GetAwaiter().GetResult())
+            """);
+        yield return Case("LoopAndNested", "6,7\n", """
+            package Issue2933.LoopAndNested
+            import System
+            import Gsharp.Extensions.Go
+
+            async func Run() string {
+                let ch = make(chan int32, 1)
+                var total = 0
+                for i in 0 ... 3 {
+                    ch <- i + 1
+                    select { case let value = <-ch { total += value } }
+                }
+
+                let outer = make(chan int32, 1)
+                let inner = make(chan int32, 1)
+                outer <- 3
+                inner <- 4
+                var nested = 0
+                select {
+                    case let first = <-outer {
+                        select {
+                            case let second = <-inner { nested = first + second }
+                        }
+                    }
+                }
+                return total.ToString() + "," + nested.ToString()
+            }
+            Console.WriteLine(Run().GetAwaiter().GetResult())
+            """);
+        yield return Case("AsyncLambdaAndCapture", "3,3\n", """
+            package Issue2933.AsyncLambdaAndCapture
+            import System
+            import Gsharp.Extensions.Go
+
+            let run = async () -> {
+                let ch = make(chan int32, 1)
+                ch <- 3
+                var direct = 0
+                var captured = 0
+                select {
+                    case let value = <-ch {
+                        direct = value
+                        let read = () -> value
+                        captured = read()
+                    }
+                }
+                return direct.ToString() + "," + captured.ToString()
+            }
+            Console.WriteLine(run().GetAwaiter().GetResult())
+            """);
+        yield return Case("AwaitAfterRead", "6\n", """
+            package Issue2933.AwaitAfterRead
+            import System
+            import System.Threading.Tasks
+            import Gsharp.Extensions.Go
+
+            async func Run() int32 {
+                let ch = make(chan int32, 1)
+                ch <- 3
+                var result = 0
+                select {
+                    case let value = <-ch {
+                        result = value
+                        await Task.Yield()
+                        result += value
+                    }
+                }
+                return result
+            }
+            Console.WriteLine(Run().GetAwaiter().GetResult())
+            """);
+        yield return Case("ReferenceAndStruct", "3,31\n", """
+            package Issue2933.ReferenceAndStruct
+            import System
+            import Gsharp.Extensions.Go
+
+            async func Run() string {
+                let refs = make(chan string?, 1)
+                let structs = make(chan DateTime, 1)
+                refs <- "abc"
+                structs <- DateTime(2026, 7, 31)
+                var refResult = 0
+                var structResult = 0
+                select {
+                    case let value = <-refs {
+                        if value == nil { refResult = -1 }
+                        else { refResult = value.Length }
+                    }
+                }
+                select {
+                    case let value = <-structs { structResult = value.Day }
+                }
+                return refResult.ToString() + "," + structResult.ToString()
+            }
+            Console.WriteLine(Run().GetAwaiter().GetResult())
+            """);
+        yield return Case("Unbuffered", "4\n", """
+            package Issue2933.Unbuffered
+            import System
+            import Gsharp.Extensions.Go
+
+            async func Run() int32 {
+                let ch = make(chan int32)
+                ch <- 4
+                var result = 0
+                select { case let value = <-ch { result = value } }
+                return result
+            }
+            Console.WriteLine(Run().GetAwaiter().GetResult())
+            """);
+        yield return Case("ClosedChannelResetsBinding", "30\n", """
+            package Issue2933.ClosedChannel
+            import System
+            import Gsharp.Extensions.Go
+
+            async func Run() int32 {
+                let ch = make(chan int32, 1)
+                ch <- 3
+                close(ch)
+                var result = 0
+                for i in 0 ... 2 {
+                    select {
+                        case let value = <-ch { result = result * 10 + value }
+                    }
+                }
+                return result
+            }
+            Console.WriteLine(Run().GetAwaiter().GetResult())
+            """);
+    }
+
+    [Fact]
+    public void AsyncAndSynchronousReceiveBindingsAreEqualAtRuntime()
+    {
+        const string Source = """
+            package Issue2933.Parity
+            import System
+            import Gsharp.Extensions.Go
+
+            async func Async() int32 {
+                let ch = make(chan int32, 1)
+                ch <- 3
+                var result = 0
+                select {
+                    case let value = <-ch { result = 7 + value }
+                }
+                return result
+            }
+
+            func Sync() int32 {
+                let ch = make(chan int32, 1)
+                ch <- 3
+                var result = 0
+                select {
+                    case let value = <-ch { result = 7 + value }
+                }
+                return result
+            }
+
+            Console.WriteLine(Async().GetAwaiter().GetResult())
+            Console.WriteLine(Sync())
+            """;
+
+        var output = CompileLoadAndRun(Source, nameof(AsyncAndSynchronousReceiveBindingsAreEqualAtRuntime));
+        var values = output.Split('\n', StringSplitOptions.RemoveEmptyEntries)
+            .Select(int.Parse)
+            .ToArray();
+
+        Assert.Equal(2, values.Length);
+        Assert.Equal(values[1], values[0]);
+        Assert.Equal(10, values[0]);
+    }
+
+    [Theory]
+    [MemberData(nameof(SelectShapes))]
+    public void RequestedSelectShapesLoadAndRun(string name, string expectedOutput, string source)
+    {
+        Assert.Equal(expectedOutput, CompileLoadAndRun(source, name));
+    }
+
+    [Fact]
+    public void SiblingAsyncBindingFormsRemainCorrect()
+    {
+        const string Source = """
+            package Issue2933.SiblingBindings
+            import System
+            import System.Collections.Generic
+            import System.Threading.Tasks
+
+            async func Run() string {
+                await Task.Yield()
+
+                var rangeResult = 0
+                let values = List[int32]{2, 3}
+                for value in values { rangeResult += value }
+
+                var patternResult = 0
+                switch object("abc") {
+                    case value is string { patternResult = value.Length }
+                    default { patternResult = -1 }
+                }
+
+                var catchResult = 0
+                try { throw Exception("xy") }
+                catch (ex Exception) { catchResult = ex.Message.Length }
+
+                return rangeResult.ToString() + "," +
+                    patternResult.ToString() + "," +
+                    catchResult.ToString()
+            }
+            Console.WriteLine(Run().GetAwaiter().GetResult())
+            """;
+
+        Assert.Equal("5,3,2\n", CompileLoadAndRun(Source, nameof(SiblingAsyncBindingFormsRemainCorrect)));
+    }
+
+    private static object[] Case(string name, string expectedOutput, string source) =>
+        new object[] { name, expectedOutput, source };
+
+    private static string CompileLoadAndRun(string source, string name)
+    {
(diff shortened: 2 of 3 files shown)


DavidObando:
## Rubber-duck verdict (Opus 5): NOT MERGEABLE — one blocker

Reviewed head `d49495a9`. Worktree left untouched.

### BLOCKER — `src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Statements.cs:647`

```csharp
&& this.asyncFieldMap != null
```

Async iterators hoist into `iteratorEmitCtx.FieldMap`, not `asyncFieldMap`, so they are excluded **by construction** and still exhibit #2933 verbatim at HEAD. Probes run against the author's HEAD build, in a child process, exact stdout:

| probe | base `17789519` | HEAD `d49495a9` | `gsi` | |
|---|---|---|---|---|
| `P0` async func (the PR's own repro) | **7** | **10** | 10 | fixed |
| `P2` **async iterator** (`async sequence[int32]`, `yield 7+value` in arm) | **7** | **7** | **10** | **UNFIXED** |
| `P7` async iterator, arm body does not yield | **7** | **7** | 10 | **UNFIXED** |
| `P4` async lambda inside async func | 7 | 10 | 10 | fixed |
| `P3` sync lambda inside async func | 10 | 10 | 10 | n/a |

`P2` and `P7` are live compiled-vs-interpreted divergences at HEAD — precisely the defect class this PR exists to close. `P6` (async iterator reading the channel *outside* `select`) prints 10/10, ruling out a general async-iterator confounder.

**The house pattern already exists and was not used.** `MethodBodyEmitter.Closures.cs:1043-1054` (`EmitCapturedVariableLoad`) checks *both* maps, and its comment says literally "In an async or **async-iterator** MoveNext body". `MethodBodyEmitter.cs:206-218` (`ResolveCurrentStateMachineFieldToken`) already resolves `asyncFieldMap?.StructType ?? iteratorEmitCtx?.SmClass`. The infrastructure is present; only the guard is wrong.

Note that merely relaxing the guard is **not** sufficient: `EmitStoreVariable` (`MethodBodyEmitter.MemberAccess.cs:207-208`) redirects only for `asyncFieldMap`, so for iterators it would emit a no-op store to a dead local. The `stfld` must be emitted directly.

Candidate fix, built and verified — same line count, only pre-existing helpers:

```csharp
FieldSymbol hoistedField = null;
if (arm.Variable != null
    && ((this.asyncFieldMap != null && this.asyncFieldMap.TryGetHoistedField(arm.Variable, out hoistedField))
        || (this.iteratorEmitCtx != null && this.iteratorEmitCtx.FieldMap.TryGetValue(arm.Variable, out hoistedField))))
{
    this.il.OpCode(ILOpCode.Ldarg_0);
    this.il.LoadLocal(outSlot);
    this.il.OpCode(ILOpCode.Stfld);
    this.il.Token(this.ResolveCurrentStateMachineFieldToken(hoistedField));
}
this.EmitStatement(arm.Body);
```

`P2` and `P7` go 7 → **10**; `P0`/`P1`/`P3`/`P4` unchanged; `Compiler.Tests` filtered on `Select|Issue2933|Iterator|Async|Yield|Sequence` = **447 passed, 0 failed**. Needs one new async-iterator pin test.

### SHOULD-FIX — a second, deeper gap exists and this PR cannot close it

Sync iterator with the binding live **across** a yield:

```gsharp
select { case let value = <-ch { yield 1 \n yield 7 + value } }
```

base `1,7` → HEAD `1,7` → **with the blocker fix applied, still `1,7`** — while `gsi` prints `1,10`.

Emitter instrumentation shows `DBG var=value async=False iter=True inIterMap=False` for the sync iterator versus `inIterMap=True` for the async one. **The variable is never hoisted at all**, so the local is reset on `MoveNext` re-entry. That is a hoisting-analysis defect, not an emitter one — pre-existing, orthogonal, and not fixable at `Statements.cs`. A plain local held across a yield prints 10, so sync-iterator hoisting works generally; it is specifically `BoundSelectCase.Variable` that the capture walker does not treat as a binding site.

Should be filed separately as *"select-arm bindings are not recognised as binding sites by state-machine hoisting analysis"*, with `Lowering/Iterators/HoistedFieldRewriter.cs` named as the home — its own docstring claims to be "the single home … so a fix here applies everywhere at once", and it has no select-arm override.

### Root fix vs patch — the copy is the right shape

`SlotPlanner.cs:576-589` shows that for a `ReceiveBind` arm, `OutSlots[i]` **is** `locals[arm.Variable]` — the same slot. That is why non-state-machine code has always been correct. The `ldflda`-the-hoisted-field alternative is feasible but not better: it needs the same two conditional sites, trades one copy for one address computation, and does not close the iterator gap by itself, since it would still have to choose a field map. The genuine "two storage locations for one variable" fix lives at lowering, and it is the same fix as the SHOULD-FIX above.

### Verified correct

- **Sibling site audit is complete.** `:590` default arm has no binding; `:680` `EmitSelectSendProbe` — send arms bind nothing; `Patterns.cs:85` is switch patterns. Both receive kinds route through `EmitSelectReceiveProbe` (`:571-575`) and both `EmitStatement(arm.Body)` calls inside it are patched. `ReceiveDiscard` genuinely never carries a variable — parsed with `identifier: null` (`Parser.Patterns.cs:497-499`), bound with `variable` left null (`StatementBinder.Blocks.cs:750-754`).
- **Closed-channel path correct and independently pinned.** Select in a loop, receive then close: base **0** → HEAD **30**. The closed path yields the element default, not a stale value. Pinned by `ClosedChannelResetsBinding` (`Issue2933AsyncSelectArmBindingTests.cs:176`).
- **Suspension points all pass**: await-then-re-read 7 → **13**; closure captures the binding 7 → **10**; string-typed binding → `ab!`; multi-arm 7 → **10**; nested select 0 → **34**; imported CLR struct (`DateTime`) 8 → **10**.
- **Mutation: 2 semantic branches, both independently pinned.** Control 64/64 first. M1 (success path `:626` reverted) → 8 failures; M2 (closed path `:639` reverted) → 1 failure.
- **Merge coordination clean.** `merge-tree` against `origin/main` `8dddbbf2` and #2912, #2929, #2961, #2964, #2966 — all clean, all pairwise merges with #2967 clean. Against #2964 specifically: same file, but its hunks sit at `:97/:129/:203`, ~420 lines from #2967's `:623-660`, and are semantically disjoint (unmanaged-pointer pinning vs state-machine field writes; no shared helper or slot state).
- **Corpus**: 148 `.gs`, 3 use `select`, **0** async-select, none inside iterators. Non-regression evidence only.
- **#2965 correctly deferred** — `data struct Pair(Value int32)` over a channel throws `InvalidProgramException` identically on base, HEAD, *and* in a purely synchronous control with no state machine. Independent defect.
- **Test harness is exemplary**: `Assembly.Load` + `GetTypes()`, child process with both pipes `ReadToEndAsync()` started before `WaitForExit(30_000)`, `Kill(entireProcessTree: true)`, exact stdout compare. Does not touch `IlVerifier.cs`.

### Additional SHOULD-FIX / NIT

- **PR body: "gsi parity restored 10/10" overclaims.** The interpreter evaluates the bound tree, not IL, and never had this bug — proved by mutation, since both reversions left Interpreter 1/1 green. That test is a **guard**, not a pin.
- **Rebase and restate counts against `8dddbbf2`**: Compiler **3893 → 3903**, Interpreter **489 → 490**. The `3876`/`472` figures are `17789519` and are stale. The +10/+1 deltas are arithmetically exact (2 `[Fact]` + 1 `[Theory]`×8 = 10; 1 `[Fact]` = 1).
- The "Sibling binding audit" and `SiblingAsyncBindingFormsRemainCorrect` cover binding *forms* (range/pattern/catch) but never the state-machine **kind** axis. Name that axis explicitly once covered.

DavidObando:
## Rubber-duck review — head `d49495a9` — **NOT MERGEABLE** (1 blocker)

Re-validated at 2026-07-31T21:20 against `origin/main` = `a6b1561f2`; `git merge-tree` is **clean**, so this is a semantic blocker, not a merge one. All probes below are `Assembly.Load` + `GetTypes()` + bounded child-process execution with exact stdout compared against a known-bad control built at `17789519`. `IlVerifier.Verify` was not relied on.

---

### BLOCKER — `src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Statements.cs:647`

```csharp
&& this.asyncFieldMap != null
```

Async iterators hoist into `iteratorEmitCtx.FieldMap`, **not** `asyncFieldMap` (`MethodBodyEmitter.cs:99/101`), so they are excluded by construction and **still exhibit #2933 verbatim at this head**:

| probe | base `17789519` | head `d49495a9` | `gsi` | |
|---|---|---|---|---|
| async func (the PR's own repro) | 7 | **10** | 10 | fixed |
| **async iterator** (`async sequence[int32]`, `yield 7 + value` in arm) | 7 | **7** | **10** | **UNFIXED** |
| async iterator, arm body does not yield | 7 | **7** | 10 | **UNFIXED** |
| async lambda inside async func | 7 | **10** | 10 | fixed |
| sync lambda inside async func | 10 | 10 | 10 | n/a |

The two `7`s are **live compiled-vs-interpreted divergences at head** — the exact defect class this PR exists to close. A control that reads the channel *outside* the `select` in the same async iterator prints 10/10, ruling out a general async-iterator confounder.

**The house pattern already exists and was not used.** `MethodBodyEmitter.Closures.cs:1043-1054` (`EmitCapturedVariableLoad`) checks *both* maps, and its comment reads "In an async or **async-iterator** MoveNext body". `ResolveCurrentStateMachineFieldToken` (`MethodBodyEmitter.cs:206-218`) already resolves `asyncFieldMap?.StructType ?? iteratorEmitCtx?.SmClass`. Only the guard is wrong.

Note that **relaxing the guard alone is not sufficient**: `EmitStoreVariable` (`MethodBodyEmitter.MemberAccess.cs:207-208`) redirects only for `asyncFieldMap`, so for iterators it would emit a dead store to an unused local. The `stfld` must be emitted directly.

**Suggested fix — built and verified, same size, no new helpers:**

```csharp
FieldSymbol hoistedField = null;
if (arm.Variable != null
    && ((this.asyncFieldMap != null && this.asyncFieldMap.TryGetHoistedField(arm.Variable, out hoistedField))
        || (this.iteratorEmitCtx != null && this.iteratorEmitCtx.FieldMap.TryGetValue(arm.Variable, out hoistedField))))
{
    this.il.OpCode(ILOpCode.Ldarg_0);
    this.il.LoadLocal(outSlot);
    this.il.OpCode(ILOpCode.Stfld);
    this.il.Token(this.ResolveCurrentStateMachineFieldToken(hoistedField));
}
this.EmitStatement(arm.Body);
```

Both async-iterator probes go 7 → **10**; all other probes unchanged; `Compiler.Tests` filter `Select|Issue2933|Iterator|Async|Yield|Sequence` = **447 passed, 0 failed**. Build freshness proven bidirectionally: applying the fix flipped the probe 7 → 10, restoring flipped it 10 → 7, and the 447-test run was `--no-build` off the 10-printing build.

Needs one added pin test covering an async-iterator select arm.

---

### SHOULD-FIX — file the sync-iterator hole separately (do not fix here)

A sync iterator whose binding is live **across** a `yield` prints `1,7` where `gsi` prints `1,10` — on base, at head, **and with the blocker fix applied**. Instrumenting the emitter printed `inIterMap=False`: the variable is **never hoisted at all**, so the local resets on `MoveNext` re-entry. That is a hoisting-analysis defect, not an emitter one, and it is pre-existing.

Root cause for both halves: `BoundSelectCase.Variable` is not recognised as a binding site by state-machine capture analysis. Suggested scope for the issue — *"select-arm bindings are not treated as binding sites by state-machine hoisting analysis"*, general rather than async-specific, naming `Lowering/Iterators/HoistedFieldRewriter.cs` as the single home (its own docstring claims to be "the single home … so a fix here applies everywhere at once", and it has no select-arm override).

### SHOULD-FIX — two wording corrections in the PR body

1. **"gsi parity restored 10/10" overclaims.** The interpreter evaluates the bound tree, not IL, and never had this bug. Proved by mutation: both reversions left Interpreter 1/1 green. The new interpreter test is a **guard**, not a pin — worth saying so.
2. **State the mutation result as 2 semantic branches, not "2/2 killed" against 11 tests.** At semantic-branch granularity there are exactly two, and both are independently pinned (below). The count is correct but reads as thin next to the test count.

### NIT — restate counts against the base you actually rebase onto

The body quotes Compiler `3876` / Interpreter `472`, which are `17789519` figures and now three merges stale. I independently measured `8dddbbf2` as Compiler **3893** / Interpreter **489** (matching the fleet consensus); `main` has since moved to `b1e693bb1` and `a6b1561f2`. Please re-measure the rebase base and restate — the added-test arithmetic itself is exact (2 `[Fact]` + 1 `[Theory]`×8 = 10 Compiler, 1 Interpreter).

### NIT — name the axis that was missed

The "Sibling binding audit" and `SiblingAsyncBindingFormsRemainCorrect` cover binding *forms* (range / pattern / catch) but never the state-machine *kind* axis, which is precisely where the blocker lives. Worth one explicit sentence once it is covered.

---

## What I verified and found correct

**Sibling emission sites — complete.** `:590` default arm (`arm.IsDefault`, no binding), `:680` `EmitSelectSendProbe` (send arms bind nothing), `Patterns.cs:85` (different construct). Both receive kinds route through `EmitSelectReceiveProbe` (`:571-575`) and both `EmitStatement(arm.Body)` calls inside it were patched. `ReceiveDiscard` genuinely never carries a variable — parsed with `identifier: null` (`Parser.Patterns.cs:497-499`), bound with `variable` left null (`StatementBinder.Blocks.cs:750-754`) — so the `arm.Variable != null` guard short-circuits correctly.

**The copy is the right shape, not a symptom patch.** `SlotPlanner.cs:576-589` shows `OutSlots[i]` **is** `locals[arm.Variable]` for a `ReceiveBind` arm, which is why non-state-machine code has always been correct. Passing `ldflda` of the hoisted field as the `out` argument is feasible but needs the same two conditional sites, is not smaller, and — decisively — does not close the iterator gap, since it still has to choose a field map.

**Closed-channel path — correct and independently pinned.** Select in a loop, receive then close: base **0** → head **30**. The closed path yields the element default, not a stale value from the prior iteration.

**Suspension points — all pass** (base → head): await after reading the binding then re-reading 7 → **13**; closure capture of the binding 7 → **10**; string binding → `ab!`; multi-arm where a different arm fires 7 → **10**; nested select 0 → **34**; imported CLR struct (`DateTime`) 8 → **10**.

**Mutation — 2 semantic branches, both independently pinned.** Unmutated control first (64/64 on `Issue2933|Select`), full-solution build with explicit exit-code gating, restore-and-rebuild between probes.

| branch | mutation | failures |
|---|---|---|
| success path `:626` | revert to `EmitStatement(arm.Body)` | **8** |
| closed path `:639` | revert to `EmitStatement(arm.Body)` | **1** |

No survivals, so nothing here licences a deletion. The closed-channel site is *not* untested, which was the plausible worry.

**Corpus.** 148 `.gs` files, **3** with `select` (`GoChannelsGated.gs`, `PortScan.gs`, `Select.gs`), **0** async-select, none inside iterators. The author's claim is exact; the corpus is non-regression evidence only and carries no signal here.

**#2965 is genuinely out of scope.** `data struct Pair(Value int32)` over a channel throws `InvalidProgramException` identically on base and head — *and in a purely synchronous control with no state machine at all*. Independent defect, correctly deferred.

**Merge coordination — clean, no ordering constraint.** `d49495a9` merges clean onto `a6b1561f2`; previously verified clean against #2912, #2929, #2961, #2964, #2966 and pairwise with each. On #2964 specifically — same file, but its hunks sit at `:97/:129/:203` (adds `EmitManagedPointerAsUnmanagedPointer`, rewires the `fixed`-pin sites), ~420 lines from this PR's `:623-660`, with no shared helper and no shared slot state. Textually *and* semantically disjoint.

**Test harness — exemplary, no notes.** `Issue2933AsyncSelectArmBindingTests.cs:283-355` does `Assembly.Load(File.ReadAllBytes(...))` + `GetTypes()`, then a child process with **both** pipes `ReadToEndAsync()` started before a **bounded** `WaitForExit(30_000)`, `Kill(entireProcessTree: true)` on timeout, exact stdout compare, and `Console.SetOut` save/restore. It does not touch `IlVerifier.cs`.

The core fix is right and the async-func case is properly nailed down. It is one guard condition short of covering the defect it describes.

DavidObando:
## Verdict: **MERGEABLE**. No blockers. Both suspected coverage gaps were probed and both are genuinely pinned.

Reviewed at head `323fc5c71`. Merge base computed against a **freshly fetched remote ref** (my clone's `origin` was a stale local path): `210126dd9`, i.e. 3 behind / 2 ahead of `main` `9832ebbc6`.

### Integration

| check | result |
|---|---|
| `merge-tree` into `9832ebbc6` | rc=**0** |
| File overlap with main's 3 new commits (#3014, #3012, #2997) | **none** |
| Solution build of the merge | rc=**0**, `0 Error(s)` |
| Gate 2 — zero-byte over **all** `GSharp.Core.dll` | 40 copies, **0** zero-byte |
| Gate 3 — md5 over the five ordinary paths | `c6ca131a…`, **distinct=1** |

**Head movement is a rebase, not new content.** `bd624e45c` → `323fc5c71` pulls in main's merges only; the production blob is byte-identical (`8624c8bfbd8d1dd1ab5ff1048cfcc3cbb4a99341` at both), so my earlier measurements still apply.

### Scope — clean, and unusually tight

`20 +/2 −` in one production file; the two test files are **new** (`A`, not `M`). The change replaces two `EmitStatement(arm.Body)` calls with `EmitSelectReceiveArmBody(arm, slots.OutSlots[index])` and adds that method.

It reuses the pre-existing `ResolveCurrentStateMachineFieldToken` (`MethodBodyEmitter.cs:206`) with the same `Ldarg_0 / LoadLocal / Stfld / Token` idiom already used in 4 places. No new abstraction invented for a 20-line change.

When neither `asyncFieldMap` nor `iteratorEmitCtx` is present the method degenerates to the original `EmitStatement(arm.Body)`, so ordinary non-state-machine `select` is untouched — confirmed by a 70-test `~Select` sweep, **70/70**.

**No undeclared scope in either direction** (see below for why the iterator clause is in scope).

### Mutation — the two things worth doubting, both killed

**MI — remove the `iteratorEmitCtx` clause** from the guard (build-green mutant):

> **KILLED** by `AsyncIteratorReceiveBindingsMatchAsyncFunctionAtRuntime` — `Expected: [10, 10, 10]`, `Actual: [10, 7, 7]`.

This also **corrects a hypothesis I had from the #3013 review**: async iterators route through `iteratorEmitCtx.FieldMap`, *not* `asyncFieldMap`. So that clause is **required for #2933 itself** — it is not speculative groundwork for #2975/#3013, and its presence is properly in scope.

**MW — per-arm out-slot**, `slots.OutSlots[index]` → `slots.OutSlots[0]` at both call sites:

> **KILLED** by `RequestedSelectShapesLoadAndRun(name: "MultipleAndDefault", expectedOutput: "14,5\n")`.

That test has a genuinely discriminating multi-arm select — `case let ignored = <-empty { selected = -100 }` / `case let value = <-ready { selected = 10 + value }` / `default { selected = -200 }` with `ready <- 4`. A wrong-slot store yields `10` instead of `14`. So a store on the wrong arm **is** detectable, which was the specific #2997-style risk.

### Test results

| suite | result |
|---|---|
| `~Issue2933AsyncSelectArmBinding` (Compiler) | **11/11** |
| `~Issue2933` (Interpreter) | **1/1** |
| `~Select` sweep (Compiler) | **70/70** |

### Re-validated against the new `main`

My sync-iterator probe on `9832ebbc6` + #2967 still prints **7** (unchanged). #2967 remains a safe no-op for synchronous iterators, now confirmed with #2997 landed.

## NIT — multi-arm coverage is adequate but thin

Of 17 `select` blocks in the new tests, **14 have a single receive-bind arm, 2 have two, 1 has none** — and only one of the two-arm cases has distinct per-arm outcomes. That one case is what killed MW, so coverage is sufficient rather than lucky. A three-arm select with a distinct value per arm, each exercised, would harden this cheaply. Not blocking.

## Note for the merge queue

The **synchronous**-iterator end-to-end path is exercised by no test in either PR — #2967's clause is covered via async iterators, so the *line* is pinned, but sync-iterator select-receive is currently pinned only by my ad-hoc probe. That is exactly the emit test #3013 is already blocked on adding, so the pair covers it. Recording it here so it is not lost when #3013 rebases.

Merge order remains **#2967 → #3013**.

Issue #2939, opened 2026-07-31
Add interpreter golden-sample conformance coverage


DavidObando:
## Problem

`SampleConformanceTests` exercises only emitted assemblies. Interpreter regressions in shipped samples therefore remain invisible.

A 122-sample interpreter sweep against #2935 found nine output improvements over `main`: `ArrowLambda`, `ArrowFunctionTypeClause`, `DelegateCallSyntax`, `FuncToDelegate`, `FuncToSystemDelegate`, `MethodGroupToDelegate`, `LinqExtensions`, `OptionalExtensionArgs`, and `EventSubscription` (whose handler previously never fired). Compiled sample coverage did not detect any of them.

## Proposed coverage

Add an interpreter counterpart to `SampleConformanceTests` using the same sample discovery and golden-output comparison. Keep exclusions explicit for samples that require compilation, native dependencies, or unsupported interpreter-only runtime features.

Found while validating #2935.

DavidObando:
Addendum from #2935 review: interpreter golden comparison must exclude or normalize delegate bridge method names for `FuncToSystemDelegate.gs`. Cached interpreter closures of the same delegate type intentionally share one bridge `MethodInfo`, while compiled closures have distinct generated methods (121/122 samples otherwise match after path normalization). This identity divergence is inherent to removing per-closure `Expression.Compile()`; delegate equality and targets remain distinct.

DavidObando:
## Triage + design — measured at `a6b1561f2`

Read-only triage in a scratch clone. Every number below is from an executed run; build verified exit-code → non-zero-byte → md5 propagation agreement across all six `GSharp.Core.dll` consumers (`5ba5eace10aca822f0e5520545e5cca3`).

### Correction to the premise first

The issue body's **nine** samples (`ArrowLambda`, `ArrowFunctionTypeClause`, `DelegateCallSyntax`, `FuncToDelegate`, `FuncToSystemDelegate`, `MethodGroupToDelegate`, `LinqExtensions`, `OptionalExtensionArgs`, `EventSubscription`) were **improvements #2935 delivered**, not a list of open divergences. #2935 merged at 12:35 today (`e21677f69`) and is in `a6b1561f2`. Seven of the nine now pass. Only `FuncToSystemDelegate` and `EventSubscription` still diverge, for reasons unrelated to what #2935 fixed.

[...]

PR #2968, opened 2026-07-31
Support implicit interface indexers on classes and structs

### LINKED ISSUE #2915: Unify implicit indexer implementation rules for plain and constructed interfaces
## Problem

Implicit indexer implementation depends on whether the G# interface is constructed generic.

Plain interface indexers are deliberately rejected by ADR-0118:

```gsharp
interface I { prop this[k string] int32 { get; } }
class C : I {
    prop this[k string] int32 { get { return 1 } }
}
```

Result: `GS0187`, no emit.

Equivalent constructed-generic interface indexers are accepted and produce loadable metadata:

```gsharp
interface I[T] { prop this[k string] T { get; } }
class C : I[int32] {
    prop this[k string] int32 { get { return 1 } }
}
```

Result: no diagnostics; emitted assembly passes IL verification and `Assembly.Load(...).GetTypes()`.

## Cause

Interface property lookup historically used separate paths. Plain declared interfaces used name lookup that excludes indexers; constructed interfaces used signature lookup that accepts them. Consolidating those paths for #2888 requires preserving this inconsistency explicitly.

## Expected

Choose and document one rule for both plain and constructed G# interfaces. If implicit indexer implementation remains unsupported, reject both forms consistently. If supported, update ADR-0118 and accept both forms with matching index/value types and accessor kinds.

## Context

Found during review of #2911. Existing test `PlainNonExplicitIndexer_DoesNotSatisfyInterfaceIndexerContract` pins plain rejection; #2911 tests preserve loadable constructed-generic behavior to avoid unrelated regression.

### LINKED ISSUE #2954: Explicit interface indexer on struct emits non-virtual accessor
## Problem

An explicit G# interface indexer implementation on a struct compiles without diagnostics, but the emitted type cannot load because its accessor is not virtual.

## Reproduction

```gsharp
package Repro
import System

interface I {
    prop this[key int32] int32 { get; }
}

struct S(Base int32) : I {
    private prop (I) this[key int32] int32 -> Base + key
}

var value I = S(5)
Console.WriteLine(value[7])
```

Compile as a `net10.0` executable, then run it in a child process.

## Actual

Compilation succeeds. Runtime fails while loading `S`:

```text
System.TypeLoadException: Method 'Repro.I.get_Item' on type 'Repro.S' ... must be virtual to implement a method on an interface or super type.
```

## Expected

The explicit accessor should be emitted with the virtual attributes required by its `MethodImpl` row, and interface dispatch should print `12`.

## Notes

Found while investigating #2915 on base `f2ecef88`. Explicit interface indexers on classes work. The value-type accessor appears to lose its virtual promotion in the computed-property `EmitFunction` path.

### LINKED ISSUE #2960: Struct interface sequence property emits an unloadable assembly
## Summary

A struct implementing a custom interface property whose type is `sequence[int32?]` compiles with zero errors, but the emitted assembly cannot load.

This reproduces byte-identically on `main` (`17789519`) and the rebased #2952 head. It is pre-existing and is not caused by #2952.

## Repro

```gs
package K2
import System

interface IBox {
    prop Vals sequence[int32?] { get; }
}

func values() sequence[int32?] {
    yield 5
    yield nil
}

struct Box : IBox {
    prop Vals sequence[int32?] -> values()
}

var box IBox = Box{}
for value in box.Vals {
    Console.WriteLine(value == nil ? "nil" : value.ToString())
}
```

Compile and run:

```console
$ dotnet gsc.dll /out:k2.dll /target:exe /targetframework:net10.0 k2.gs
Wrote k2.dll
$ dotnet exec --runtimeconfig k2.runtimeconfig.json k2.dll
Unhandled exception. System.TypeLoadException: Method 'get_Vals' in type 'K2.Box' from assembly 'K2, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' does not have an implementation.
   at K2.<Program>.<Main>$(String[] args)
Aborted
```

Runtime exit code: **134**.

Independent base/head comparison during #2952 review produced identical diagnostics, return codes, and binaries for the original `k2.gs` probe.

## Expected

The property getter must fill the interface slot with an exactly matching signature, or the compiler must reject the program rather than emit an unloadable assembly.

### LINKED ISSUE #3001: Struct computed-property accessors are never promoted to virtual, emitting unloadable assemblies
## Summary

A **computed** property accessor (`-> expr` or a `get`/`set` block) on a **struct** that implements an interface is never promoted to `virtual`. The compiler reports no diagnostic, but the emitted type cannot load.

No indexer, no generics, and no `sequence` type are required. The minimal trigger is an ordinary get-only property returning `string`.

This is the general root cause behind **#2960** and **#2954**, both of which describe narrower symptoms of it. See *Relationship to existing issues* below — **#2960 in particular is mis-titled**, and a fix agent reading its title will look at the wrong node kind.

## Minimal repro

```gs
package GapCheck
import System

interface IName {
    prop Label string { get; }
}

struct S : IName {
    prop Label string -> "hi"
}

var s = S{}
var i IName = s
Console.WriteLine(i.Label)
```

## Actual

Compilation succeeds with zero diagnostics. The assembly does not load:

```text
System.TypeLoadException: Method 'get_Label' in type 'GapCheck.S' from assembly
'GapCheck, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'
does not have an implementation.
```

`Assembly.Load(File.ReadAllBytes(...))` + `GetTypes()` throws `ReflectionTypeLoadException` on the same assembly, so this is detectable without executing the program.

## Expected

`hi`.

The accessor must fill the interface slot with the virtual attributes its `MethodImpl` row requires — or the compiler must reject the program rather than emit an unloadable assembly.

## Scope — what is and is not affected

Measured on `8dddbbf2`, each probe compiled with `gsc` directly, then `Assembly.Load` + `GetTypes()`, then executed in a bounded child process.

| shape | result |
|---|---|
| struct, **computed** property, get-only | **FAILS** — `get_Label` has no implementation |
| struct, **computed** property, get/set | **FAILS** — `get_V` has no implementation |
| struct, **auto** property `{ get; set; }` | passes — prints `0` |
| struct, interface **method** | passes — prints `7` |
| **class**, computed property | passes — prints `hi` |

So the defect is precisely **struct + computed accessor**. Field-backed auto-properties are unaffected because they are emitted through a different path.

## Root cause

Value-type virtual promotion is gated at `src/Core/CodeAnalysis/Emit/FunctionEmitter.cs:568` on `MethodInfoHelpers.RequiresVirtualOnValueType`, which delegates to `MethodImplicitlyImplementsInterface` (`src/Core/CodeAnalysis/Emit/MethodInfoHelpers.cs:61`).

That helper enumerates:

- event accessors (`AddMethodSymbol` / `RemoveMethodSymbol` / `RaiseMethodSymbol`),
- `iface.Methods`,
- `ImplementedClrInterfaces` methods — and this path explicitly skips `openMethod.IsSpecialName`, which is exactly what a property accessor is.

It never enumerates properties:

```console
$ grep -nE "Propert|Getter|Setter|get_|IsIndexer" src/Core/CodeAnalysis/Emit/MethodInfoHelpers.cs
(no output)
```

Auto-properties escape the bug because they are handled in `MemberDefEmitter` (`src/Core/CodeAnalysis/Emit/MemberDefEmitter.cs:467`), which *does* have a `PropertyImplicitlyImplementsInterface` check (`:881`). Computed accessors go through `FunctionEmitter` instead, where no property arm exists. **The two emit paths disagree about interface implementation, and only one of them knows about properties.**

## Differential evidence

A prototype adding a property-accessor arm to `RequiresVirtualOnValueType` flips the behaviour, on the same probes and the same harness:

| probe | `8dddbbf2` | prototype |
|---|---|---|
| repro above | `GETTYPES_FAIL` `ReflectionTypeLoadException` | `GETTYPES_OK count=3` → `hi` |
| #2954 repro | `GETTYPES_FAIL` — `must be virtual …` | `GETTYPES_OK count=3` → `12` |
| #2960 repro | `GETTYPES_FAIL` — `get_Vals` no implementation | loads and runs |

One change closes all three, which is the evidence that they are one root cause and not three.

The prototype was reverted and is not proposed as-is. **A production fix must not match on name alone.** The existing `MemberDefEmitter` check compares names at `:906`, and every indexer is named `Item`, so two indexers differing only in parameter type both match. The fix should compare parameter arity and types, and should be **a single shared helper** used by both emit paths — a second predicate is how this defect class arose, and a third branch would make it permanent.

## Relationship to existing issues

- **#2960** — *same root cause.* Its title attributes the failure to `sequence[int32?]`, but that type is incidental; so is the property's name. Its repro uses `prop Vals sequence[int32?] -> values()`, a computed accessor, and the plain `string` repro above produces the identical `does not have an implementation` error. **#2960 is mis-titled.**
- **#2954** — *same root cause*, explicit-interface indexer variant. Explicit implementations surface a different message (`must be virtual to implement a method on an interface`) because the `MethodImpl` row exists but points at a non-virtual method; implicit implementations surface `does not have an implementation`. One cause, two texts.
- **#2915** — *not this issue.* #2915 is about which implicit interface **indexer** forms the binder accepts (a `GS0187` binding-rules question). This issue is an **emit** defect and needs no indexer at all. They meet only in that struct indexer support also requires this promotion fix.

## Notes

- Observed on `8dddbbf2`. `main` has since advanced to `a6b1561f2`; its only production delta is 20 lines confined to lambda definite-return binding (`Binder.cs`, `LambdaBinder.cs`), which cannot affect emit-layer virtual promotion. Reproduction on `a6b1561f2` is therefore by inspection, not by execution.
- Compiler-side detection is cheap: `Assembly.Load` + `GetTypes()` fails without running the program. `IlVerifier.Verify` alone is **not** sufficient — assemblies with this defect pass it.

DavidObando:
## Summary

- Accept implicit G# interface indexers uniformly for plain and constructed interfaces, on classes and structs.
- Promote computed struct property accessors into exact interface slots, fixing explicit struct indexers, ordinary computed struct properties, and sequence-valued interface properties.
- Resolve interface property accessors in `gsi`, including explicit implementations and imported CLR base members.
- Document the upward-unification decision in ADR-0118.

Fixes #2915
Fixes #2954
Fixes #2960
Fixes #3001

## Design decision

I first implemented and tested the reject-both alternative. That decision did not survive investigation: class implementations already emitted verifiable, loadable metadata and dispatched correctly, while every failure used to justify rejection came from a small defect elsewhere.

- The struct failures shared one emitter gap: computed property accessors did not receive the value-type virtual promotion already used by methods, events, and auto-properties.
- The interpreter failure shared one dispatch gap: interface calls searched `StructSymbol.Methods` but not property getter/setter symbols.
- The #2960 interpreter probe also exposed nullable iterator storage using the erased underlying CLR type instead of `Nullable<T>`.

#3001 needs no interpreter change: `gsi` already handles ordinary computed struct interface properties. The evaluator delta remains limited to the indexer dispatch failures in #2915/#2954.

Fixing those roots is smaller and preserves the deliberate constructed-generic capability added by `9354651e` for #2370. Rejecting both would remove working behavior and hide the real defects. ADR-0118 previously contained no interface rule; the historical plain rejection was an artifact of name lookup excluding indexers, not a design decision.

## Production delta

Against parent `e80132516`:

- `DeclarationBinder.InterfaceVerification.cs` **+1/-13**: remove the plain-indexer rejection from declared/base-chain and imported-base signature matching. The removed `iface` parameters existed only for that guard; construction substitution remains in `typeParameterMap`.
- `MemberDefEmitter.cs` **+80/-13**: route auto-property and computed-accessor emission through one implicit-interface matcher; compare name, indexer kind, accessor availability, property type, parameter arity, and every index parameter type (including constructed-interface substitution and imported CLR interfaces); keep explicit properties out of implicit matching.
- `MethodInfoHelpers.cs` **+11/-0**: recognize getter/setter functions owned by explicit or implicit interface properties when deciding value-type virtual attributes.
- `Evaluator.Calls.cs` **+58/-1**: resolve implicit and explicit property accessors, invoke imported CLR backing accessors, and preserve nullable iterator element CLR types.

## Existing tests changed explicitly

- `PlainNonExplicitIndexer_DoesNotSatisfyInterfaceIndexerContract_ReportsGS0187` is now `PlainNonExplicitIndexer_DispatchesThroughInterfaceTypedReceiver`; it changed from asserting GS0187 to compiling, loading, and returning `1`. This is the behavior #2915 intentionally changes.
- `OrdinaryGenericInterfaceIndexer_DispatchesThroughInterfaceTypedReceiver` is restored to its original `10`-returning form and rationale from #2370.
- The `Issue2888` constructed-generic fixture is restored from explicit back to implicit. It has no net diff from `e80132516`.
- The obsolete failure-compilation helper used only by the former plain-rejection test was removed.

## Additional #2955 coverage

The existing `NullableErasureMatrix_LoadsAndDispatchesThroughClrSlots` fact now also pins the five widened nullable-erasure shapes from merged #2955: slice element, fixed-array element, map value, function argument, and function return. The existing rejection fact also pins the boundary: `List[T?]` does not match `List[int32]` and reports GS0187. This adds no test method and does not change suite counts; the three affected #2945 facts pass at HEAD.

## Executed evidence

The compiled child matrix covers class/struct × plain/constructed × get-only/get-set, generic enclosing type, `string` and `int32` indexes, interface-typed dispatch, explicit regression cases, and imported CLR-base inheritance. Three indexer-free struct-property tests independently pin #3001's computed get-only (`hi`), computed get/set (`42`), and the existing field-backed auto-property path (`42`) as a no-regression guard. It also proves a same-name `int32` indexer does not satisfy a `string` indexer slot: reflection keeps that ordinary getter non-virtual while the explicit matching getter is virtual. Every output assembly runs through ILVerify, `Assembly.Load(...).GetTypes()`, and a bounded child process with both pipes drained asynchronously. A known-bad post-load child control is detected.

Actual `gsi` child runs cover the same eight implicit shapes plus explicit plain/constructed class and struct dispatch, imported CLR-base dispatch, and nullable sequence `nil` preservation. Outputs were `11/42/17/43/19/35/44/46/47`, `2/1/4/12`, `7/9`, and `5/nil`. A known-bad `gsi` control exited 1 and was detected.

Issue repros at base and head:

- #2954: `e80132516` compiles, then exits `-6` with `TypeLoadException` saying `I.get_Item` must be virtual. HEAD loads and prints `12`.
- #2960: `e80132516` compiles, then exits `-6` with `TypeLoadException` saying `Box.get_Vals` has no implementation. HEAD loads and prints `5` then `nil`.
- #3001: `e80132516` compiles the ordinary computed struct interface property, then exits `-6` with `TypeLoadException` saying `LabelValue.get_Label` has no implementation. HEAD loads and prints `hi`.

The ordinary computed-property case is intentionally fixed here, not deferred or silently bundled: it shares the exact value-type accessor-promotion root with #2954/#2960. Separate get-only and get/set regressions pin `Assembly.GetTypes()` and bounded execution; the auto-property guard proves the separate field-backed emit path remains intact.

## Mutation results

Semantic branches: **11 pins / 11 guards**. Every mutant built warnings-clean; code was restored and rebuilt between probes. Each mutant pin turned red while a guard stayed green on the same verified artifact. The signature mutant changed the ordinary-instance `GSharp.Core.dll` MD5 from `eeed8d231f7b88f39eb4a094d431e226` to `437e9cb5f8c1b4a5b574448b0a0f9d25`; restoration returned exactly to `eeed8d231f7b88f39eb4a094d431e226`. Removing the shared computed-property accessor arm changed it to `ef9a4eac74939ca41ebd1b844693a52e`: both indexer-free computed-property tests turned red, the auto-property guard stayed green, and restoration returned exactly to the original MD5 with all three green.

| removed behavior | pin | guard |
|---|---|---|
| declared/base-chain plain indexer matching | plain class/struct matrix reports GS0187 | imported-base probe passes |
| imported-base plain indexer matching | imported `List[int32]` probe reports GS0187 | declared matrix passes |
| getter accessor promotion | #3001 get-only/get-set fail `Assembly.GetTypes()`; #2954/#2960 also fail | auto-property and class implementations pass |
| setter accessor promotion | #3001 get/set and plain/constructed struct setters fail | get-only #3001 and auto-property pass |
| explicit-property promotion | #2954 fails type load | implicit #2960 passes |
| implicit-property promotion | #2960 reports missing getter | explicit #2954 passes |
| full property-slot signature matching | same-name `int32` indexer is wrongly promoted for a `string` slot | matching explicit struct indexer still loads and dispatches |
| implicit `gsi` accessor dispatch | matrix reports GS9999 dictionary-key failure | explicit dispatch passes |
| explicit `gsi` accessor dispatch | explicit outputs regress and GS9999 fires | implicit dispatch passes |
| imported CLR-backing dispatch | imported probe reports GS9999 | G# implicit dispatch passes |
| nullable iterator element CLR type | `nil` reports GS9999 “Value cannot be null” | indexer matrix passes |

## Validation

Default-parallelism release build at `df7cbbe4d`: **0 warnings / 0 errors**. The five ordinary-instance `GSharp.Core.dll` copies were nonzero and MD5-identical (`eeed8d231f7b88f39eb4a094d431e226`).

Executed counts, same commands and machine state, against the requested parent baseline:

| suite | `e80132516` | `df7cbbe4d` | delta |
|---|---:|---:|---:|
| Core | 7028 passed, 1 skipped | 7028 passed, 1 skipped | 0 |
| Compiler | 3931 passed | 3940 passed | +9 |
| Interpreter | 493 passed | 497 passed | +4 |
| LanguageServer | 452 passed | 452 passed | 0 |

148-file corpus: 139 compiled, 9 retained identical diagnostics, 0 timeouts, and **0 differences** in exit code, normalized diagnostics, or SHA-256 output binary between base behavior and HEAD. **0/148 files declare an interface indexer**, so the corpus is non-regression evidence only; feature evidence comes from the probes above.

### DIFF
--- a/docs/adr/0118-user-indexer-member-declaration.md
+++ b/docs/adr/0118-user-indexer-member-declaration.md
@@ -4,7 +4,7 @@
 - **Date**: 2026-06-22
 - **Phase**: v0.2 — language surface
 - **Related**: ADR-0051 (Property declarations — `prop Name T { get; set }`), ADR-0020 (Generic type-parameter brackets — Go-style `[T]`), ADR-0087 (Constructed generic user-type method references), ADR-0115 (C#→G# migration tool), issue #507 (member index assignment)
-- **Issue**: [#944](https://github.com/DavidObando/gsharp/issues/944)
+- **Issues**: [#944](https://github.com/DavidObando/gsharp/issues/944), [#2915](https://github.com/DavidObando/gsharp/issues/2915), [#2954](https://github.com/DavidObando/gsharp/issues/2954), [#2960](https://github.com/DavidObando/gsharp/issues/2960)
 
 ## Context
 
@@ -137,6 +137,37 @@ accessors and call sites resolve through the existing constructed
 generic user-type machinery (ADR-0087: method references parented at the
 constructed `TypeSpec`, signatures encoded with `!0`).
 
+#### Interface contracts
+
+An indexer may implement a G# interface indexer either implicitly or
+explicitly. Implicit matching uses the index-parameter types, value type,
+and required accessors, with generic substitutions applied for a
+constructed interface:
+
+```gsharp
+interface IRepo[T] {
+    prop this[key string] T { get; }
+}
+
+class Store : IRepo[int32] {
+    prop this[key string] int32 -> 1
+}
+```
+
+This applies uniformly to plain and constructed interfaces and to class
+and value-type implementations. Value-type property accessors that satisfy
+an interface contract receive the same virtual slot promotion as methods
+and events. The interpreter resolves interface calls through property
+accessors as well as ordinary methods.
+
+Before issue #2915, plain interfaces happened to use name lookup, which
+excluded indexers, while constructed interfaces used signature lookup,
+which accepted them. This ADR had not defined interface-indexer
+implementation semantics; the plain rejection was an implementation
+artifact, not a language decision. Rejecting both forms was considered,
+but would have removed working constructed-interface behavior and hidden
+the small emitter and interpreter defects that prevented uniform support.
+
 ### 3. Access binding — `obj[i]` / `obj[i] = v` on a user type
 
 When the target of an index expression is a user-defined type
@@ -214,6 +245,13 @@ emit + interpreter), with emit-and-run regression tests:
 - **get/set** indexer on a non-generic class: write via `obj[i] = v`,
   read back via `obj[i]`.
 - **non-generic get-only** indexer.
+- implicit and explicit implementation of plain and constructed-generic
+  interface indexer contracts by classes and structs (issues #2915 and
+  #2954), including interface-typed dispatch in emitted and interpreted
+  programs.
+- value-type property accessors receive the interface-slot attributes
+  required for loadable metadata, including sequence-valued properties
+  (issue #2960).
 - index access binding for both read and write to the declared indexer,
   including inside the type's own methods.
 - the old crashing shape now compiles and runs — **no GS9998**.
--- a/src/Core/CodeAnalysis/Binding/DeclarationBinder.InterfaceVerification.cs
+++ b/src/Core/CodeAnalysis/Binding/DeclarationBinder.InterfaceVerification.cs
@@ -508,7 +508,6 @@ internal sealed partial class DeclarationBinder
             // property implemented (or inherited) ANYWHERE in the base chain.
             var found = TryGetConstructedInterfacePropertyImplementation(
                 structSymbol,
-                iface,
                 iprop,
                 typeParameterMap,
                 out var implProp,
@@ -516,7 +515,6 @@ internal sealed partial class DeclarationBinder
             if (!found
                 && HasMatchingImportedBaseProperty(
                     structSymbol,
-                    iface,
                     iprop,
                     typeParameterMap))
             {
@@ -702,12 +700,6 @@ internal sealed partial class DeclarationBinder
             typeParameterMap,
             eraseReferenceNullability: true);
 
-    private static bool IsUnsupportedImplicitInterfaceIndexer(
-        InterfaceSymbol iface,
-        PropertySymbol interfaceProperty)
-        => interfaceProperty.IsIndexer
-            && ReferenceEquals(iface.Definition ?? iface, iface);
-
     private static TypeSymbol SubstituteInterfacePropertyType(
         TypeSymbol interfaceType,
         Dictionary<TypeParameterSymbol, TypeSymbol> typeParameterMap)
@@ -740,7 +732,6 @@ internal sealed partial class DeclarationBinder
 
     private static bool TryGetConstructedInterfacePropertyImplementation(
         StructSymbol structSymbol,
-        InterfaceSymbol iface,
         PropertySymbol interfaceProperty,
         Dictionary<TypeParameterSymbol, TypeSymbol> typeParameterMap,
         out PropertySymbol implementation,
@@ -753,7 +744,6 @@ internal sealed partial class DeclarationBinder
             {
                 if (candidate.HasExplicitInterfaceClause
                     || candidate.Name != interfaceProperty.Name
-                    || IsUnsupportedImplicitInterfaceIndexer(iface, interfaceProperty)
                     || (!ReferenceEquals(current, structSymbol) && candidate.Accessibility != Accessibility.Public)
                     || candidate.IsIndexer != interfaceProperty.IsIndexer
                     || candidate.Parameters.Length != interfaceProperty.Parameters.Length)
@@ -800,13 +790,11 @@ internal sealed partial class DeclarationBinder
 
     private static bool HasMatchingImportedBaseProperty(
         StructSymbol structSymbol,
-        InterfaceSymbol iface,
         PropertySymbol interfaceProperty,
         Dictionary<TypeParameterSymbol, TypeSymbol> typeParameterMap)
     {
         var importedBase = TypeMemberModel.GetNearestImportedBase(structSymbol);
-        if (importedBase?.ClrType == null
-            || IsUnsupportedImplicitInterfaceIndexer(iface, interfaceProperty))
+        if (importedBase?.ClrType == null)
         {
             return false;
         }
--- a/src/Core/CodeAnalysis/Emit/MemberDefEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/MemberDefEmitter.cs
@@ -5,6 +5,7 @@
 #pragma warning disable SA1202 // 'public' members should come before 'private' members (organized by feature: each public entry point is followed by its own private helpers — instance-property group, static-property group, instance-event group, static-event group, interface group)
 
 using System;
+using System.Collections.Generic;
 using System.Reflection;
 using System.Reflection.Metadata;
 using System.Reflection.Metadata.Ecma335;
@@ -355,7 +356,7 @@ internal sealed class MemberDefEmitter
         {
             methodAttrs |= MethodAttributes.Virtual | MethodAttributes.NewSlot;
         }
-        else if (this.PropertyImplicitlyImplementsInterface(structSym, prop))
+        else if (PropertyImplicitlyImplementsInterface(structSym, prop))
         {
             // Issue #248: implicit interface implementation requires Virtual | NewSlot.
             methodAttrs |= MethodAttributes.Virtual | MethodAttributes.NewSlot;
@@ -365,13 +366,11 @@ internal sealed class MemberDefEmitter
             // Issue #2362: a mangled-name explicit interface property
             // implementation is bound to its interface slot purely via an
             // explicit MethodImpl row (see
-            // EmitExplicitInterfacePropertyMethodImpls) — its accessor's own
-            // name never matches the interface member's name, so
-            // PropertyImplicitlyImplementsInterface (a name-based check)
-            // never fires for it. Per ECMA-335 §II.10.3.3, a MethodImpl body
-            // method must be virtual; Final additionally prevents a derived
-            // class from accidentally overriding this synthetic accessor by
-            // name (it has no natural override point, exactly like the
+            // EmitExplicitInterfacePropertyMethodImpls).
+            // PropertyImplicitlyImplementsInterface deliberately excludes explicit implementations.
+            // Per ECMA-335 §II.10.3.3, a MethodImpl body method must be virtual;
+            // Final prevents a derived class from overriding this synthetic accessor
+            // by name (it has no natural override point, exactly like the
             // #2010 mangled explicit method convention's own methods, which
             // get Virtual | NewSlot | Final unconditionally via EmitFunction).
             methodAttrs |= MethodAttributes.Virtual | MethodAttributes.NewSlot | MethodAttributes.Final;
@@ -470,7 +469,7 @@ internal sealed class MemberDefEmitter
         {
             methodAttrs |= MethodAttributes.Virtual | MethodAttributes.NewSlot;
         }
-        else if (this.PropertyImplicitlyImplementsInterface(structSym, prop))
+        else if (PropertyImplicitlyImplementsInterface(structSym, prop))
         {
             // Issue #248: implicit interface implementation requires Virtual | NewSlot.
             methodAttrs |= MethodAttributes.Virtual | MethodAttributes.NewSlot;
@@ -878,8 +877,16 @@ internal sealed class MemberDefEmitter
     /// Issue #248: determines whether a property on a class/struct implicitly implements
     /// an interface property (same name and type on any implemented interface).
     /// </summary>
-    private bool PropertyImplicitlyImplementsInterface(StructSymbol structSym, PropertySymbol prop)
+    /// <param name="structSym">The type declaring the candidate property.</param>
+    /// <param name="prop">The candidate property.</param>
+    /// <returns><see langword="true"/> when the property implements an interface property.</returns>
+    internal static bool PropertyImplicitlyImplementsInterface(StructSymbol structSym, PropertySymbol prop)
     {
+        if (prop.HasExplicitInterfaceClause)
+        {
+            return false;
+        }
+
         if (!structSym.Interfaces.IsDefaultOrEmpty)
         {
             foreach (var iface in structSym.Interfaces)
@@ -901,9 +908,10 @@ internal sealed class MemberDefEmitter
                     continue;
                 }
 
+                var typeParameterMap = DeclarationBinder.BuildInterfaceTypeParameterMap(iface);
                 foreach (var ifaceProp in defIface.Properties)
                 {
-                    if (ifaceProp.Name == prop.Name)
+                    if (PropertyMatchesInterfaceSlot(prop, ifaceProp, typeParameterMap))
                     {
                         return true;
                     }
@@ -925,7 +933,6 @@ internal sealed class MemberDefEmitter
             // (ILVerify: "Class implements interface but not method"). The
             // binder-side `MemberLookup.FindMatchingProperty` already gets
             // this right; this call brings the emitter in line with it.
-            var propClr = NullableLifting.GetEffectiveClrType(prop.Type);
             foreach (var ifaceSym in structSym.ImplementedClrInterfaces)
             {
                 var clrIface = ifaceSym?.ClrType;
@@ -936,7 +943,34 @@ internal sealed class MemberDefEmitter
 
                 foreach (var clrProp in clrIface.GetProperties(BindingFlags.Public | BindingFlags.Instance))
                 {
-                    if (clrProp.Name == prop.Name && (propClr == null || ClrTypeUtilities.AreSame(propClr, clrProp.PropertyType)))
+                    var indexParameters = clrProp.GetIndexParameters();
+                    if (clrProp.Name != prop.Name
+                        || prop.IsIndexer != (indexParameters.Length > 0)
+                        || prop.Parameters.Length != indexParameters.Length
+                        || (clrProp.GetMethod != null && !prop.HasGetter)
+                        || (clrProp.SetMethod != null && !prop.HasSetter)
+                        || !DeclarationBinder.IsInterfacePropertyTypeCompatible(
+                            prop.Type,
+                            MemberLookup.GetClrPropertyTypeSymbol(ifaceSym, clrProp),
+                            clrProp.SetMethod != null,
+                            typeParameterMap: null))
+                    {
+                        continue;
+                    }
+
+                    var parametersMatch = true;
+                    for (var i = 0; i < indexParameters.Length; i++)
+                    {
+                        if (!DeclarationBinder.TypeSignaturesEquivalent(
+                            prop.Parameters[i].Type,
+                            MemberLookup.GetIndexerParameterTypeSymbol(ifaceSym, clrProp, i)))
+                        {
+                            parametersMatch = false;
+                            break;
+                        }
+                    }
+
+                    if (parametersMatch)
                     {
                         return true;
                     }
@@ -945,6 +979,39 @@ internal sealed class MemberDefEmitter
         }
 
         return false;
+    }
+
+    private static bool PropertyMatchesInterfaceSlot(
+        PropertySymbol property,
+        PropertySymbol interfaceProperty,
+        Dictionary<TypeParameterSymbol, TypeSymbol> typeParameterMap)
+    {
+        if (property.Name != interfaceProperty.Name
+            || property.IsIndexer != interfaceProperty.IsIndexer
+            || property.Parameters.Length != interfaceProperty.Parameters.Length
+            || (interfaceProperty.HasGetter && !property.HasGetter)
+            || (interfaceProperty.HasSetter && !property.HasSetter)
+            || !DeclarationBinder.IsInterfacePropertyTypeCompatible(
+                property.Type,
+                interfaceProperty.Type,
+                interfaceProperty.HasSetter,
+                typeParameterMap))
+        {
+            return false;
+        }
+
+        for (var i = 0; i < property.Parameters.Length; i++)
+        {
+            if (!DeclarationBinder.TypeSignaturesEquivalent(
+                interfaceProperty.Parameters[i].Type,
+                property.Parameters[i].Type,
+                typeParameterMap))
+            {
+                return false;
+            }
+        }
+
+        return true;
     }
 
     /// <summary>
--- a/src/Core/CodeAnalysis/Emit/MethodInfoHelpers.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodInfoHelpers.cs
@@ -45,6 +45,17 @@ internal static class MethodInfoHelpers
             return true;
         }
 
+        foreach (var property in receiverStruct.Properties)
+        {
+            if ((ReferenceEquals(property.GetterSymbol, function)
+                    || ReferenceEquals(property.SetterSymbol, function))
+                && (property.HasExplicitInterfaceClause
+                    || MemberDefEmitter.PropertyImplicitlyImplementsInterface(receiverStruct, property)))
+            {
+                return true;
+            }
+        }
+
         return MethodImplicitlyImplementsInterface(receiverStruct, function);
     }
 
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -197,7 +197,7 @@ public sealed partial class Evaluator
     {
         if (type is Symbols.SequenceTypeSymbol seq)
         {
-            return seq.ElementType.ClrType ?? typeof(object);
+            return NullableLifting.GetEffectiveClrType(seq.ElementType) ?? typeof(object);
         }
 
         var clr = type?.ClrType;
@@ -262,6 +262,63 @@ public sealed partial class Evaluator
                 if (t.TryGetMethod(method.Name, out var candidate))
                 {
                     implMethod = candidate;
+                    break;
+                }
+
+                FunctionSymbol implicitAccessor = null;
+                foreach (var property in t.Properties)
+                {
+                    var getterMatches = property.GetterSymbol?.Name == method.Name;
+                    var accessor = getterMatches
+                        ? property.GetterSymbol
+                        : property.SetterSymbol?.Name == method.Name ? property.SetterSymbol : null;
+                    if (accessor == null)
+                    {
+                        continue;
+                    }
+
+                    var explicitAccessor = getterMatches
+                        ? property.ExplicitInterfaceMember?.GetterSymbol
+                        : property.ExplicitInterfaceMember?.SetterSymbol;
+                    if (ReferenceEquals(explicitAccessor, method)
+                        || (property.ExplicitInterfaceClauseTarget != null
+                            && DeclarationBinder.TypeSignaturesEquivalent(
+                                property.ExplicitInterfaceClauseTarget,
+                                method.ReceiverType)))
+                    {
+                        implMethod = accessor;
+                        break;
+                    }
+
+                    if (!property.HasExplicitInterfaceClause)
+                    {
+                        implicitAccessor ??= accessor;
+                    }
+                }
+
+                implMethod ??= implicitAccessor;
+                if (implMethod == null && ifaceSv.ClrBacking != null)
+                {
+                    var importedAccessor = ifaceSv.ClrBacking.GetType()
+                        .GetMethods(BindingFlags.Public | BindingFlags.Instance)
+                        .FirstOrDefault(candidate =>
+                            candidate.IsSpecialName
+                            && candidate.Name == method.Name
+                            && candidate.GetParameters().Length == node.Arguments.Length);
+                    if (importedAccessor != null)
+                    {
+                        var importedArguments = new object[node.Arguments.Length];
+                        for (var i = 0; i < node.Arguments.Length; i++)
+                        {
+                            importedArguments[i] = EvaluateExpression(node.Arguments[i]);
+                        }
+
+                        return importedAccessor.Invoke(ifaceSv.ClrBacking, importedArguments);
+                    }
+                }
+
+                if (implMethod != null)
+                {
                     break;
                 }
             }
--- a/test/Compiler.Tests/Emit/Issue2370ExplicitInterfaceEventIndexerEmitTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2370ExplicitInterfaceEventIndexerEmitTests.cs
@@ -442,22 +442,8 @@ public class Issue2370ExplicitInterfaceEventIndexerEmitTests
     // imported (BCL) interfaces.
     // ----------------------------------------------------------------------
     [Fact]
-    public void PlainNonExplicitIndexer_DoesNotSatisfyInterfaceIndexerContract_ReportsGS0187()
+    public void PlainNonExplicitIndexer_DispatchesThroughInterfaceTypedReceiver()
     {
-        // ADR-0118 (issue #944): an indexer's CLR name ("Item") is not
-        // reachable by ordinary member-name lookup — only through `obj[i]`
-        // index syntax — so `TypeMemberModel.TryGetProperty` deliberately
-        // excludes indexers, and `VerifyInterfaceImplementations` has no
-        // separate by-shape indexer-matching path. By design (pre-existing,
-        // unrelated to this fix), only an EXPLICIT `(IFace)` clause indexer
-        // implementation can satisfy an interface's indexer contract; a
-        // plain non-explicit indexer of the identical shape does not count
-        // and GS0187 correctly fires. This is a control confirming the new
-        // interface-typed-receiver dispatch work does not (and should not)
-        // change that pre-existing, intentional restriction. See
-        // ImportedBclInterfaceIndexer_DispatchesThroughInterfaceTypedReceiver
-        // below for the genuine "ordinary" (non-G#-explicit-clause) control
-        // proving the receiver-dispatch fix itself is general.
         const string source = """
             package GapCheck
 
@@ -468,44 +454,14 @@ public class Issue2370ExplicitInterfaceEventIndexerEmitTests
             class Store : IRepo {
                 prop this[key string] int32 { get { return 1 } set { } }
             }
+
+            var store = Store()
+            var asIface IRepo = store
+            asIface["key"] = 99
+            public var result = asIface["key"]
             """;
 
-        var (exitCode, output) = CompileExpectingFailure(source);
-        Assert.NotEqual(0, exitCode);
-        Assert.Contains("GS0187", output);
-    }
-
-    private static (int ExitCode, string Output) CompileExpectingFailure(string source)
-    {
-        var tempDir = Directory.CreateTempSubdirectory("gs_issue2370_fail_").FullName;
-        var srcPath = Path.Combine(tempDir, "test.gs");
-        var outPath = Path.Combine(tempDir, "test.dll");
-        File.WriteAllText(srcPath, source);
-
-        using var stdoutWriter = new StringWriter();
-        using var stderrWriter = new StringWriter();
-        var prevOut = Console.Out;
-        var prevErr = Console.Error;
-        Console.SetOut(stdoutWriter);
-        Console.SetError(stderrWriter);
-        int compileExit;
-        try
-        {
-            compileExit = Program.Main(new[]
-            {
-                "/out:" + outPath,
-                "/target:library",
-                "/targetframework:net10.0",
-                srcPath,
-            });
-        }
-        finally
-        {
-            Console.SetOut(prevOut);
-            Console.SetError(prevErr);
-        }
-
-        return (compileExit, stdoutWriter.ToString() + stderrWriter.ToString());
+        Assert.Equal(1, RunAndGetIntResult(source));
     }
 
     [Fact]
--- a/test/Compiler.Tests/Emit/Issue2915ImplicitInterfaceIndexerEmitTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2915ImplicitInterfaceIndexerEmitTests.cs
@@ -1,0 +1,485 @@
+// <copyright file="Issue2915ImplicitInterfaceIndexerEmitTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using Xunit;
+using Xunit.Sdk;
+
+namespace GSharp.Compiler.Tests.Emit;
+
+/// <summary>
+/// Issues #2915, #2954, #2960, and #3001: interface property accessors bind and emit
+/// correctly for classes, structs, plain interfaces, and constructed interfaces.
+/// </summary>
+public class Issue2915ImplicitInterfaceIndexerEmitTests
+{
+    private const string IndexerMatrixSource = """
+        package Issue2915Matrix
+        import System
+
+        interface IPlainClassGet {
+            prop this[key string] int32 { get; }
+        }
+
+        class PlainClassGet : IPlainClassGet {
+            prop this[key string] int32 -> key.Length + 8
+        }
+
+        interface IPlainClassSet {
+            prop this[index int32] int32 { get; set; }
+        }
+
+        class PlainClassSet : IPlainClassSet {
+            var Stored int32
+            prop this[index int32] int32 {
+                get { return Stored + index }
+                set { Stored = value - index }
+            }
+        }
+
+        interface IConstructedGet[T] {
+            prop this[key string] T { get; }
+        }
+
+        class ConstructedGet : IConstructedGet[int32] {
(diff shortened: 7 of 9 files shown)


DavidObando:
Heads-up from the #2911 review (I was the reviewing duck there), since this PR lands in `DeclarationBinder.InterfaceVerification.cs` and will call `GetInterfacePropertySlotType` (`:700`) on indexer value types.

**My round-3 review of #2911 contained a recommendation to delete three erasure arms in `src/Core/CodeAnalysis/Symbols/StructSymbol.cs` (`:2367` `NullabilityAnnotatedTypeSymbol`, `:2523` `ByRefTypeSymbol`, `:2529` `PointerTypeSymbol`). I have retracted it — please do not action it.** Full retraction: https://github.com/DavidObando/gsharp/pull/2911#issuecomment-5149778663

Short version: I justified "deletable" on mutant survival plus a 136-probe behavioural grid. That is not a structural proof. `NullabilityAnnotatedTypeSymbol` appears in exactly **one** arm of `SubstituteTypeForConstruction`; delete it and an annotated type matches no later arm and falls through to the terminal `return type` **un-substituted** — `List[T?]` at `I[string]` would stay `List[T?]` instead of becoming `List[string]`. My grid missed it because the only consumer I exercised was gate 1, and `AreRuntimeEquivalentIgnoringReferenceNullability` strips the annotation on both sides via `StripReferenceNullabilityCore` (`src/Core/CodeAnalysis/Symbols/TypeSymbol.cs:789`), masking the difference inside the comparer.

Indexer **value** types are a new caller of that erasure path, so this PR is exactly where the un-substituted-annotation case would first bite.

One thing that should make this PR easier: #2911 collapsed the previously duplicated ADR-0118 condition into a single helper, `IsUnsupportedImplicitInterfaceIndexer` (`DeclarationBinder.InterfaceVerification.cs:720`), called from `:771` and `:824`. There is now exactly one place to change rather than two that could drift apart.

DavidObando:
## Undeclared-closure audit: this PR should also declare `Closes #3001`

Part of a sweep across all open PRs for issues that are fixed but not declared. This is the only case in the set where a fix is **invisible** — #3001 is not mentioned in the body, the commits, the test names, or any comment in the diff, so nothing would connect the two at merge time.

**Evidence.** The diff adds an arm to `RequiresVirtualOnValueType` (`src/Core/CodeAnalysis/Emit/MethodInfoHelpers.cs:45`):

```csharp
foreach (var property in receiverStruct.Properties)
{
    if ((ReferenceEquals(property.GetterSymbol, function)
            || ReferenceEquals(property.SetterSymbol, function))
        && (property.HasExplicitInterfaceClause
            || MemberDefEmitter.PropertyImplicitlyImplementsInterface(receiverStruct, property)))
```

#3001 is titled *"Struct computed-property accessors are never promoted to virtual, emitting unloadable assemblies"*, and its body states the trigger as a computed property accessor on a struct that implements an interface, with no indexer, generics, or `sequence` required — *"the minimal trigger is an ordinary get-only property returning `string`."* That is exactly the condition this arm now matches: the function is a property getter or setter, and the property implements an interface either explicitly or implicitly.

#3001 also states it is **the general root cause behind #2960 and #2954**, and that #2960 in particular is mis-titled. This PR already declares both #2960 and #2954 — the two narrow symptoms — but not the underlying defect they share. On merge, the symptoms close and the root-cause issue stays open.

**Why this matters more than a missing line.** An open #3001 after this merges will be re-dispatched to a fix agent who reproduces nothing, because the code path is already fixed. That has already happened elsewhere in the queue. The reverse risk does not apply here: the arm genuinely implements #3001's fix rather than merely overlapping it.

**Requested change:** add `Closes #3001` alongside the existing `#2915`/`#2954`/`#2960` declarations, and consider noting in the body that #3001 is the root cause of which #2954 and #2960 are symptoms — that ordering is what makes the four-way closure legible to a reviewer.

**Evidence class.** This verdict is from static reading of the diff against #3001's stated minimal trigger; no build or test was run (the machine is at load 59–75 with 163 `dotnet` processes and this audit was scoped to `gh`/`git`). If confirmation by execution is wanted before the declaration is added, the check is #3001's own repro — a struct with a get-only computed `string` property implementing an interface — asserting the emitted assembly loads. That is a single-program check suitable for the serialised merge-build gate.

DavidObando:
## Re-review at `d6b0319f4` — no blocking correctness defect; the fix is real and correctly attributed

Merge base is `8dddbbf2` (`git merge-base` confirms), matching the PR body — no three-dot artefact. Three scratch clones: base `8dddbbf2`, head `d6b0319f4`, and head merged onto `main` `e80132516`.

**Compile-mergeability checked explicitly** (the #3000 lesson): the merge produces 0 textual conflicts *and* builds clean. `EXIT=0`, 0 errors, 0 warnings on all three trees. Gate 2 (zero-byte) over all 34/34/40 `GSharp.Core.dll` copies: 0 hits. Gate 3 over the five ordinary paths: 5/5 identical per tree, three distinct hashes (`a025927c` / `683b66c2` / `5626354b`).

Full solution, `--no-build` at head, `SampleConformance` filtered out: **0 failed everywhere.** Core 7028 + 1 skip · Compiler 3775 (+123 `SampleConformance` = **3898**) · Interpreter 493 · LSP 452 · Cs2Gs 1860. Every figure matches the PR body exactly.

---

## Q1 — the arm closes **three**, not four

Two mutations, each rebuilt full-solution, each with a distinct propagated md5, restored to `683b66c2` at the end (round-trip gate satisfied).

**M1 — revert `MethodInfoHelpers.cs` only** (md5 `0c9b3b04`):

| probe | result |
|---|---|
| #2954 | regresses → `TypeLoadException ... must be virtual`, rc=134 |
| #3001 | regresses → `TypeLoadException ... does not have an implementation`, rc=134 |
| #2960 | regresses → `TypeLoadException ... does not have an implementation`, rc=134 |
| #2915 | **still passes, prints `1`** |

**M2 — revert `DeclarationBinder.InterfaceVerification.cs` only** (md5 `460b8e67`):

| probe | result |
|---|---|
| #2915 | regresses → `GS0187` |
| #2954 / #3001 / #2960 | **all still pass** (`12` / `hi` / `5,nil`) |

The two changes are cleanly orthogonal:

| issue | closed by | base → head |
|---|---|---|
| **#2915** | `DeclarationBinder.InterfaceVerification.cs` **+1/−13** | `GS0187` → `1` |
| **#2954** | `MethodInfoHelpers.cs` arm | rc=134 → `12` |
| **#3001** | `MethodInfoHelpers.cs` arm | rc=134 → `hi` |
| **#2960** | `MethodInfoHelpers.cs` arm | rc=134 → `5,nil` |

So **`Fixes #3001` should be added — it is genuinely closed** — but the framing "one arm closes all four" is wrong. #2915 is closed by the *binder widening*, and that widening is precisely what makes Q2 worth asking. The two questions are the same question.

Metadata confirms the mechanism directly (`System.Reflection.Metadata`, no load required), for #2954's explicit struct accessor:

```
base  S.Repro.I.get_Item   Private, HideBySig, SpecialName
head  S.Repro.I.get_Item   Private, Final, Virtual, HideBySig, VtableLayoutMask, SpecialName
base  Assembly.Load + GetTypes()  FAILED: ReflectionTypeLoadException
head  Assembly.Load + GetTypes()  OK (3 types)
```

`IlVerifier` was not used as evidence anywhere; every emit probe went through `Assembly.Load(File.ReadAllBytes(...))` + `GetTypes()` and a bounded child process.

---

## Q2 — the name-only predicate is **not** load-bearing, for a structural reason

The declared-interface arm is `ifaceProp.Name == prop.Name` — not even `IsIndexer`-discriminating. It cannot go wrong, and the reason is stronger than "GS0187 happens to guard it upstream":

**G# permits at most one indexer per type.** A second one is `GS0102: 'Item' is already declared` — so within a type, `Name == "Item"` is a *unique* match, not an approximate one. There is no overload set for a name-only matcher to pick the wrong element from. That also disposes of the same concern about `implicitAccessor ??= accessor` (first-match) in the new `Evaluator.Calls.cs` block.

The binder guards that must hold before emit are all intact at head — measured, base → head:

| probe | base | head |
|---|---|---|
| index-parameter mismatch (`this[k string]` vs `this[k int32]`) | `GS0187` | **`GS0187`** |
| value-type mismatch (`string` vs `int32`) | `GS0187` | **`GS0504`** — *see below* |
| two interfaces, both indexers named `Item`, only one implementable | `GS0187` ×2 | **`GS0187` for `B.Item` only** — correctly discriminates on parameter types despite the shared name |

No over-fire found: nothing that compiled at base is rejected at head.

**But the guards are now entirely unpinned.** See Blocking below.

---

## Blocking — the only indexer-rejection test was deleted and nothing replaced it

`PlainNonExplicitIndexer_DoesNotSatisfyInterfaceIndexerContract_ReportsGS0187` was converted into `PlainNonExplicitIndexer_DispatchesThroughInterfaceTypedReceiver`. That conversion is correct — it is the behaviour #2915 intentionally changes.

The problem is what did **not** happen. All nine new test methods are positive dispatch tests; **zero assert a rejection.** Repo-wide, no test now pins `GS0187`/`GS0504` for a *plain* interface indexer shape (`Issue2888InterfacePropertyTypeMismatchTests` covers indexers only in constructed-generic form, which was already accepted at base).

This PR's entire risk surface is *"we removed a rejection"*. Removing the guard and its test in the same change leaves the replacement semantics — which I verified by hand above — defended by nothing. Three tests, ~15 lines, all shapes already measured:

- index-parameter mismatch → `GS0187`
- value-type mismatch → `GS0504`
- two same-named interface indexers, one unimplementable → `GS0187` naming the unimplementable one

**Related, and part of the same fix:** the value-type mismatch case **changes diagnostic ID, `GS0187` → `GS0504`**, on code that was already rejected. That is an improvement in specificity and entirely consistent with #2911's design — but it is an undeclared behaviour change on existing programs, mentioned nowhere in the PR body and pinned by no test.

---

## Non-blocking

**1. `MemberDefEmitter`'s new early-out silently adds `Final` to explicit-clause accessors.** The PR describes `if (prop.HasExplicitInterfaceClause) return false;` as "keep explicit properties out of that helper". Its actual effect is to reorder branch selection: explicit-clause properties previously matched by name took the `#248` branch (`Virtual | NewSlot`) and now fall through to the `#2362` branch (`Virtual | NewSlot | Final`). Measured on an explicit-clause auto-property:

```
base  S.get_N6.I.V   Private, Virtual, HideBySig, VtableLayoutMask, SpecialName
head  S.get_N6.I.V   Private, Final, Virtual, HideBySig, VtableLayoutMask, SpecialName
```

Both load and run. The new attributes are *right* — they match the `#2010`/`#2362` mangled-explicit convention and what `FunctionEmitter` already emits — so this is a consistency improvement, not a regression. But it is an undeclared, untested metadata change that nobody decided. One line in the PR body and one attribute assertion.

**2. `Fixes #3001` should be declared, and its minimal shape tested.** #3001 is the *general* root cause; #2954 and #2960 are narrower symptoms of it. Every struct case in the new emit tests is either an indexer (25 `this[` occurrences) or the `sequence[int32?]` shape. **No test covers #3001's minimal trigger** — a struct with a computed ordinary property of a simple type:

```gsharp
interface IName { prop Label string { get; } }
struct S : IName { prop Label string -> "hi" }
```

A regression that re-broke plain computed struct properties while leaving indexers and sequences working would pass the entire suite. Six lines.

---

## Q3 — `Evaluator.Calls.cs`, exact ranges

Two hunks: **`:198`** (one line, `seq.ElementType.ClrType` → `NullableLifting.GetEffectiveClrType`) and **`:265–325`** (+57, the accessor-resolution block).

**No reserved site is touched in content.** `:69` and `:1134` `BuildRefSlots` — the `BoundAddressOfExpression`-in-argument-position identity — are byte-identical; `:69` does not move at all.

**But everything below `:265` shifts by exactly +57**, so every line number other agents are holding is now stale:

| reserved | owner | base | head |
|---|---|---|---|
| `:604` / `:609` | #2985 delegate-type | 604 / 609 | **661 / 666** |
| `:863` / `:978` / `:984` | G1 reflection chokepoints | 863 / 978 / 984 | **920 / 1035 / 1041** |
| `:1108–1119` | #3004 | 1108–1119 | **1165–1176** |
| `:1134` | `BuildRefSlots` | 1134 | **1191** |
| `:69` | `BuildRefSlots` arg position | 69 | **69** (unmoved) |

File length 1472 → 1529. No content conflict with any of them.

---

## Q4 — the ADR's historical claim is **true**, and it exposes an error in issue #2915

The ADR now asserts the plain rejection was *"an implementation artifact, not a language decision"* and that the ADR *"had not defined interface-indexer implementation semantics"*. Both halves check out:

- **ADR-0118 at `8dddbbf2` contains the word "interface" exactly zero times.** It demonstrably never defined interface-indexer semantics.
- `IsUnsupportedImplicitInterfaceIndexer` has exactly one introducing commit: **`1626c06fd` (#2911, "gsc: validate implicit interface property types")**. Its commit message is entirely about *unifying* two lookup paths — "Resolve constructed candidates by shape first so generic substitution, nullability variance, and accessor-kind validation share the same verification path". The predicate appears nowhere in that rationale; it is a behaviour-preservation guard bolted into a consolidation so the pre-existing asymmetry would survive the refactor. That is the definition of an artifact.

**Consequently issue #2915's own body is wrong** where it says *"Plain interface indexers are deliberately rejected by ADR-0118"*. The ADR said nothing. Worth correcting in the issue so the claim is not quoted forward — the PR is right and the issue it fixes is the thing that is mistaken.

---

## Not this PR — a pre-existing `gsi` gap, worth filing separately

Brace-construction of a class with an imported CLR base fails in the interpreter:

```gsharp
class C : List[int32] { }
var c = C{}          // gsc: fine.  gsi: GS9999: Object type
c.Add(42)            // System.Collections.Generic.List`1[System.Int32] does not
Console.WriteLine(c.Count)   // match target type ...StructValue.
```

I first hit this through the newly-enabled indexer path and nearly filed it as a divergence this PR creates. It is not: **with no interface involved at all, base and head fail identically**, so it is pre-existing and independent. `C()` works, `C{}` does not — in `gsi` only; `gsc` runs both correctly.

Both new imported-base tests use `Store()`, so the suite never touches the `{}` form. That is fine for *this* PR — requiring it here would be scope creep — but the newly documented ADR-0118 surface is reachable via a construction syntax the interpreter cannot execute, so it belongs on #2939's list.

DavidObando:
### Addendum — re-verified against the *actual* current `main`, `f70626452`

My verdict above used `e80132516`. `main` has since moved five commits to **`f70626452`**, and one of them lands in a file this PR edits: **#2980 (`a311efffd`)** flips `GetInterfacePropertyExpectedType` from `private static` to `internal static` at `DeclarationBinder.InterfaceVerification.cs:720` — immediately adjacent to the `IsUnsupportedImplicitInterfaceIndexer` block this PR deletes at `:700-706`. Close enough to be a plausible conflict, so I re-ran rather than assumed.

Result: **no conflict, and no semantic collision.**

```
merge d6b0319f4 onto f70626452   0 conflicts
dotnet build GSharp.sln -c Release   EXIT=0, 0 errors, 0 warnings
zero-byte GSharp.Core.dll: 0 of 40
md5 over the five ordinary paths: 5/5 identical (f2379957)
MSB302[67]|MSB3883: 0
```

All four fixes and both retained guards reproduce on that merged tree, `gsc` + `Assembly.Load`/`GetTypes()` + child process, and `gsi`:

| probe | gsc | child | gsi |
|---|---|---|---|
| #2915 plain class indexer | rc=0 | `1` | `1` |
| #2954 explicit struct indexer | rc=0 | `12` | `12` |
| #3001 struct computed property | rc=0 | `hi` | `hi` |
| #2960 struct sequence property | rc=0 | `5,nil` | `5,nil` |
| index-parameter mismatch | `GS0187` | — | `GS0187` |
| value-type mismatch | `GS0504` | — | `GS0504` |

So the PR is compile-mergeable on current `main`, and nothing in the verdict changes. The Blocking item remains the missing negative tests, which is test-only.

DavidObando:
## Re-review at `91266e96d` — the push was a **pure rebase**; my blocker is **not** closed

Round-2 of the review posted as 5151027339 / 5151054180. Re-measured everything against current `main` `210126dd9`.

### Q1 — undeclared scope: **NO. There is no new scope.**

The reported "5 files, +471/−0" (`Binder.cs`, `LambdaBinder.cs`, `Issue2921…` ×2, `Issue2936…`) is **entirely rebase artefact**, not content.

- Merge base moved `8dddbbf2` → **`a6b1561f2`**. The branch was rebased.
- `merge-base(d6b0319f4, 91266e96d)` is still the **old** base `8dddbbf2`, so a three-dot old-head→new-head sweeps up everything `main` gained in between.
- Attribution of the five "new" files, all ancestors of `origin/main`:
  - `b1e693bb1` (#2973) added `Issue2936SourceGenericReceiverCallTests.cs`
  - `a6b1561f2` (#2961) added `LambdaBinder.cs +19`, `Binder.cs +1`, both `Issue2921…` files
- The PR's own three-dot diff `a6b1561f2...91266e96d` = **9 files, +759/−69** — the same file set as at `d6b0319f4`.
- **Literal tree diff of those 9 files, `d6b0319f4` vs `91266e96d`: EMPTY.** Byte-identical.

So the `Fixes` lines are correct as they stand: #2921 was fixed by merged #2961, #2936 covered by merged #2973. Nothing here silently closes an issue. This is the two-dot-across-a-rebase trap, not undeclared scope.

Because the content is unchanged, every finding from round 2 carries forward verbatim — including the mutation attribution (M1 reverting `MethodInfoHelpers.cs` → only #2915 passes; M2 reverting `DeclarationBinder.InterfaceVerification.cs` → only #2915 fails), which stands without re-running.

### Q2 — is the negative-test blocker closed? **NO.**

`git grep -E 'GS0187|GS0504'` across both new test files at `91266e96d` returns **nothing** (denominator: 2 files, 396 added lines, 0 hits; positive control — the pattern matches 3× in `Issue2370…EmitTests.cs` on `main`).

The two rejection paths are still pinned by **no** test:
- `GS0187` — interface indexer with mismatched parameter types
- `GS0504` — interface indexer with mismatched value type (**and this ID changed from `GS0187` at base — still undeclared in the PR body**)

Both verified by hand as still firing (see the merge probes below), but hand-verification expires the moment someone edits `MemberDefEmitter.PropertyImplicitlyImplementsInterface`. The PR converted the only indexer-rejection test into a positive one and added nine positive dispatch tests and zero negative ones. **Blocker stands.**

### Merge and build against `210126dd9`

| gate | result |
|---|---|
| merge | clean, **0 conflicts**; merge tip `696ce1583` |
| net diff | 9 files, +759, −69 |
| build (`-p:ContinuousIntegrationBuild=true`, no `--no-build`) | **EXIT=0**, 0 errors |
| gate 2 zero-byte | **examined 40 `GSharp.Core.dll`, zero-byte 0** |
| gate 3 md5 | 5 of 5 ordinary Release paths uniform at `8c301c1c` |

Six probes on the **merged** tree, `gsc` → `dotnet exec` → `gsi`, all reproducing:

| probe | expected | gsc/exec | gsi |
|---|---|---|---|
| #2915 plain class indexer | `1` | `1` | `1` |
| #2954 explicit struct indexer | `12` | `12` | `12` |
| #3001 struct computed prop | `hi` | `hi` | `hi` |
| #2960 struct sequence prop | `5,nil` | `5,nil` | `5,nil` |
| negative param mismatch | `GS0187` | `GS0187` | `GS0187` |
| negative value mismatch | `GS0504` | `GS0504` | `GS0504` |

### Collisions with #3016 / #3021 — **none**

Both were cleared at file level *and* at API surface, because file-level clearance is what missed the #3000 semantic conflict.

- **#3016** (`81e44da88`) touches `ExpressionBinder.Access.Accessor.cs`, `ExpressionBinder.Calls.Arguments.cs`, `ExpressionBinder.Calls.Invocation.cs` — disjoint from #2968's file set.
- **#3021** (`2c7049e64`) touches `LambdaBinder.cs` — **#2968 does not touch `LambdaBinder.cs`**; that adjacency was against the rebase phantom, not against this PR.
- API surface: #2968 removes `IsUnsupportedImplicitInterfaceIndexer` and reshapes `PropertyImplicitlyImplementsInterface` to `internal static`. Neither symbol appears in **any changed file** of #3016 or #3021 (0 hits; positive control 3 hits in `main`'s `DeclarationBinder.InterfaceVerification.cs`).

No merge-order constraint among the three.

### Verdict

**Not yet mergeable — one test-only addition away.** The scope question is fully clean and was never a real issue. The single thing standing between this PR and merge is two negative test methods asserting `GS0187` and `GS0504` for plain (non-explicit-clause) indexers, plus a line in the body declaring the `GS0187 → GS0504` change and `Fixes #3001`.

Carried forward from round 2, unchanged:
- **SHOULD-FIX** — `MemberDefEmitter.cs:884`: the new `HasExplicitInterfaceClause` early-out reroutes explicit-clause accessors from the #248 branch to the #2362 branch, silently adding **`Final`** (measured `Private, Virtual, HideBySig, VtableLayoutMask` → `+ Final`). Right direction, undeclared and untested.
- **SHOULD-FIX** — declare `Fixes #3001`; it is proven closed by the `MethodInfoHelpers.cs` arm, and its **minimal shape** (a struct computed property that is not indexer- or `sequence`-shaped) has no test.
- **NIT** — issue #2915's own body claims implicit interface indexers were "deliberately rejected by ADR-0118". ADR-0118 at base contains the word "interface" **zero** times; `IsUnsupportedImplicitInterfaceIndexer` was introduced by `1626c06fd` (#2911) purely to preserve pre-consolidation behaviour. The ADR claim in this PR's description is correct; the issue's is not.

DavidObando:
## Correction to my own review: the negative-test BLOCKER is **retracted**. It was half wrong, and the half that survives is not blocking.

I reported that `GS0187` and `GS0504` were "pinned by NO test". That was based on two bad greps of mine. Both are now corrected by mutation testing rather than by searching.

### Where my grep went wrong

1. **Wrong population.** I scoped the search to the two new `Issue2915…` files. The real diff has **four** test files. A search over the wrong population is an unrun check.
2. **Wrong pattern.** I searched for `prop Item` and `Item[`. G# indexer syntax is **`prop this[key string]`** — my pattern could not match a single indexer declaration anywhere in the repo. Worse, `prop Item` *did* return hits, by substring-matching `prop Item**s**` — an ordinary property. A non-empty grep is as unrun as an empty one until you know what it matched.

### What is actually pinned — established by mutation, not search

Mutation A, at `DeclarationBinder.InterfaceVerification.cs:628` — suppress the whole `if (!found)` rejection block for indexers, reproducing the bypass this PR removed:

- Probes flip: `PN2` and `PN3` go from `GS0187`/`GS0504` to **compiling cleanly (rc=0)** and dying at runtime with `TypeLoadException` (rc=134), while `gsi` prints `2` and `GS9999`. Compiler and interpreter **diverge**. Positive probe `P2915` still returns `1`, so the mutation is targeted.
- Tests: **Failed: 1, Passed: 38, Total: 39.**
- The catcher is **`test/Compiler.Tests/Emit/Issue2888InterfacePropertyTypeMismatchTests.cs:241`**, `NullableGenericPropertySurface_EmitsLoadsOrRejectsWithoutOutput`, via its `NullableIndexerRejected` case:

```gs
interface I[T] { prop this[key string] T? { get; } }
class C : I[int32] { prop this[key string] int32? { get { return nil } } }
```

**So the mismatched-implicit-interface-indexer rejection IS pinned.** Note this file is **byte-identical at base `a6b1561f2` and head** — the pin is pre-existing and inherited, not added here, which is exactly why a search of the PR's own diff could never find it.

### The residual — real, but SHOULD-FIX not BLOCKER

Mutation B, narrowed to the `typeMismatch == null` arm only (the `GS0187` branch) at `:664`:

- Probes isolate cleanly: `PN2` breaks (compiles, `TypeLoadException`, `gsi` prints `2`); `PN3` still correctly reports `GS0504`.
- Tests: **Failed: 0, Passed: 42, Total: 42.**

**The `GS0187` arm for indexers is uncovered.** A regression there compiles successfully and produces a silently wrong program — the exact silent-divergence shape. One test method closes it: an implicit interface indexer with mismatched *parameter* types, asserting `GS0187`.

Also note the surviving pin asserts **exit code only** (`CompileExpectingFailure`), never a diagnostic ID. So the undeclared `GS0187 → GS0504` renumbering is still unpinned by ID.

Round-trip verified both times: perturb → observe flip → restore → observe flip back, `GSharp.Core.dll` md5 returning to `8c301c1c`, `git status --porcelain` empty.

### Merge and build re-verified at `main` = `210126dd9`

Merge clean, **0 conflicts**, tip `696ce1583`; build **EXIT=0**, 0 errors; gate 2 **40 `GSharp.Core.dll` examined / 0 zero-byte**; gate 3 5-of-5 uniform at `8c301c1c` (differs from the previously reported `f2379957` only because the merge base moved — expected). Six probes reproduce: `1` / `12` / `hi` / `5,nil` / `GS0187` / `GS0504`, `gsc` and `gsi` agreeing.

No collision with **#3016** or **#3021** at file or API-surface level.

### Verdict: **MERGEABLE**

The blocking objection does not survive measurement. Outstanding items, none blocking:

- **SHOULD-FIX** — add one negative test for the `GS0187` indexer arm (param mismatch); it is currently uncovered and fails silently.
- **SHOULD-FIX** — declare **`Fixes #3001`**; it is proven closed by the `MethodInfoHelpers.cs` arm and its minimal shape is untested.
- **SHOULD-FIX** — declare the `GS0187 → GS0504` change; no test pins the ID for indexers.
- **NIT** — `MemberDefEmitter.cs:884`: the `HasExplicitInterfaceClause` early-out silently adds `Final` to explicit-clause accessors. Correct direction, undeclared.

DavidObando:
## Duck verdict — **MERGEABLE**, and it should go **before #3047**

Measured tree: merge commit `e22e158d2`, parents **exactly** `61b7a08c1` (main) + `df7cbbe4d` (head). **Head had moved twice** during this review (`91266e96d` → `c42191828` → `df7cbbe4d`) — verified against the `gh` API, not inherited. Merge-base `e80132516` (**29 behind**), three-dot: **9 files / +951 / −79**. Merge rc=0, build **exit 0, 0 warnings / 0 errors**.

### MERGE-ORDER — the thing you asked for. #2968 and #3047 **collide**.
- Both declare **`Fixes #3001`**.
- Their `MethodInfoHelpers.cs` hunks are **byte-identical** (same 11 lines).
- `git merge-tree refs/pr2968 refs/pr3047` → **rc=1, CONFLICT (content) in `MemberDefEmitter.cs`** (+93 vs +96 on the same file).
- **#2968 alone closes #3001, proven by execution**, not by argument — I compiled #3047's own issue shape (and #2968's verbatim test source) with both compilers:

| probe | `main` `61b7a08c1` | merged #2968 |
|---|---|---|
| #3001 computed struct get-only | `GETTYPES=FAIL … get_Label … does not have an implementation` | **`GETTYPES=ok`** |
| same, block-bodied getter variant | `GETTYPES=FAIL` | **`R()=5`** |

**Recommendation: land #2968 first.** #3047 then has no production work left — its `MethodInfoHelpers` hunk is already present verbatim and its `MemberDefEmitter` hunk conflicts. What #3047 still carries that #2968 does not is its **263-line `Issue3001StructComputedAccessorVirtualTests.cs`**; that is worth salvaging as a test-only PR rather than discarding.

### The `Item`-name trap: **fixed, not duplicated**
You warned that the correct fix is one shared helper on arity + parameter types, not names, and to say so if this PR added a *second* name-comparing path. It does not — measured, not assumed:
- `MemberDefEmitter.cs:911` — `ifaceProp.Name == prop.Name` → **`PropertyMatchesInterfaceSlot(...)`**, comparing name + `IsIndexer` + parameter arity + accessor presence + `IsInterfacePropertyTypeCompatible` + **per-parameter `DeclarationBinder.TypeSignaturesEquivalent`**.
- `MemberDefEmitter.cs:946` — CLR-interface branch likewise now checks `GetIndexParameters()` arity and each parameter type.
- `MethodInfoHelpers.cs:48` — new loop matches by **`ReferenceEquals(property.GetterSymbol, function)`**, symbol identity, and delegates to the *same* `PropertyImplicitlyImplementsInterface`, now `internal static`.
- The stale comment claiming "a name-based check" is deleted. This is the shape you asked for.

### False-negative set intact — the real risk of a restriction-removal
Removing `IsUnsupportedImplicitInterfaceIndexer` makes matching purely shape-based, so I probed what *should still* be rejected:

| probe | `main` | merged |
|---|---|---|
| p1 class, matching indexer | `GS0187` | compiles, **`LOAD=ok GETTYPES=ok R()=7`** |
| p3 **struct**, matching indexer | `GS0187` | compiles, **`R()=7`** |
| p2 wrong index **type** (`int32` vs `string`) | `GS0187` | **`GS0187`** ✔ |
| p4 wrong **arity** (2 params vs 1) | `GS0187` | **`GS0187`** ✔ |
| p5i **getter-only** vs `{get;set;}` | `GS0187` | **`GS0187`** ✔ |

No false negatives found. Load/execute used throughout — **not** `IlVerifier` alone, per #3001.

**gsc ↔ gsi four-cell** (PR touches `Evaluator.Calls.cs` +59): base `GS0187`/`GS0187`; merged `gsi` prints **`7`**, `gsc` loads with `GETTYPES=ok`. **Both engines agree in both directions.**

### Mutation — both production hunks killed
| mutant | site | build | result |
|---|---|---|---|
| M1: shape match → `ifaceProp.Name == prop.Name` (the pre-PR trap) | `MemberDefEmitter.cs:911` | exit 0, 0 errors | **KILLED** — `MismatchedStructIndexer_StaysNonVirtual` |
| M2: delete property loop | `MethodInfoHelpers.cs:48` | exit 0, 0 errors | **KILLED** 6/35, incl. `ExplicitStructIndexer_…` (#2954), `NullableSequenceStructProperty_…` (#2960), `ComputedStructGetOnly/GetSet…` (#3001) |

Cycle closed clean: clean `GSharp.Core.dll` md5 `1805a3ec…` → mutant differs → **restored equals clean**. M2 is also my evidence that #2954/#2960/#3001 are each independently pinned.

### Scope — declared, and I checked specifically because the diff looked like it wasn't
`Issue2370ExplicitInterfaceEventIndexerEmitTests.cs` shows **net −58** and deletes `PlainNonExplicitIndexer_DoesNotSatisfyInterfaceIndexerContract_ReportsGS0187`, inverting a `GS0187` assertion. That reads like a weakened pin. It is not: the body declares `Fixes #2915/#2954/#2960/#3001`, discloses the rename and inversion explicitly (body line 36), and amends **ADR-0118**, which previously encoded the old rejection. `docs/adr/0118` is **modified, not added** — no ADR-number collision with #3024's 0152 (`main` tops at 0153).

**Docs gate not exposed:** descriptors `= new("GS####"` **376 → 376**; `docs/diagnostics.md` and `website/docs/ref/diagnostics.md` row counts unchanged both sides. No row added, renamed, or deleted.

### Green
`Issue2915` + `Issue2370` + `Issue2945` in `Compiler.Tests`: **38/38**. `Issue2915` in `Interpreter.Tests`: **4/4**.

### NIT — `MemberDefEmitter.cs:984`
There are now **two** shape-matching implementations: the binder's `TryGetConstructedInterfacePropertyImplementation` and the emitter's new `PropertyMatchesInterfaceSlot`. Both are shape-based, so the `Item` trap is gone, but drift between binder and emitter is exactly what produces "Class implements interface but not method" — the failure the comment at `:933` already describes. Worth collapsing to one helper later; not a merge blocker.

### Process disclosure
My first `#3001` probe run reported "fails identically on base and merged", which would have been a serious false finding. Cause: **I restored the M2 mutant source but did not rebuild before probing**, so those runs used the mutant compiler. Caught by the md5 cycle. Re-run after rebuild gives the table above. Flagging it because the failure mode looked like a genuine defect.

PR #2963, opened 2026-07-31
Invalidate narrowing across loop back-edges

DavidObando:
Fixes #2943

## Summary

- speculatively bind narrowed loop bodies, then use lowered CFG reachability to find writes and potentially mutating calls that can reach a back-edge
- remove only narrowing inherited from outside the loop before rebinding; preserve condition-derived and in-body re-narrowing
- track assignment roots by bound symbol identity across locals, captures, and top-level globals
- preserve inherited narrowing only when every reaching assignment has compatible type and proven runtime-non-null provenance
- descend into bound function-literal bodies so closure writes invalidate narrowing
- restore speculative diagnostics, label/goto state, loop state, and synthetic lambda ordinals before rebinding
- document top-level, `gsi`, and conservative runtime-null rules in ADR-0069

Covered loop forms: C-style `for`, infinite `for`, ellipsis, sync/async range, `while`, and `do…while`, including nested control flow and direct/labeled `continue`/`break`.

## Top-level and function behavior

Fresh `GSharp.sln` build used for all driver probes. `out/bin/Release/Core`, `Compiler`, and `Repl` copies of `GSharp.Core.dll` had matching post-build mtimes.

| assignment after use | `gsc` top-level | `gsc` function | `gsi` top-level | `gsi` function |
|---|---|---|---|---|
| `c = C(33)` | rc 0, `11/22/33` | rc 0, `11/22/33` | rc 0, `11/22/33` | rc 0, `11/22/33` |
| `c = fresh` | rc 0, `11/22/33` | rc 0, `11/22/33` | rc 0, `11/22/33` | rc 0, `11/22/33` |
| `c = Mk(33)` | rc 0, `11/22/33` | rc 0, `11/22/33` | rc 0, `11/22/33` | rc 0, `11/22/33` |

`gsi` also prints final expression value `3`; program output is identical.

Required rejection controls:

| shape | `gsc` top-level | `gsc` function | `gsi` top-level | `gsi` function |
|---|---|---|---|---|
| original #2943: use, then `c = nil` | GS0159, rc 1 | GS0159, rc 1 | GS0159, rc 1 | GS0159, rc 1 |
| straight-line `c = nil; c.Print()` | GS0159, rc 1 | — | GS0159, rc 1 | — |
| `while` use, then `c = nil` | GS0159, rc 1 | — | GS0159, rc 1 | — |

## Deliberate symbol policy

[...]

PR #2972, opened 2026-08-01
Fix interpreter delegate exception diagnostics

### LINKED ISSUE #2938: Interpreter loses uncaught delegate exception diagnostics through CLR member calls
## Problem

Uncaught exceptions from delegate values invoked through tuple/list elements lose both the original message and source location because the reflection member-call path leaves `TargetInvocationException` wrapped.

```gsharp
let handler (int32) -> int32 = (value int32) -> 1 / value
let pair = (handler, 0)
pair.Item1(0)
```

Current diagnostic:

```
error GS9999: Exception has been thrown by the target of an invocation.
```

Direct `handler(0)` preserves `Attempted to divide by zero`, and typed interpreter `catch` works because `Evaluator.Statements.cs:703-721` recursively unwraps runtime exceptions. Only the uncaught GS9999 path leaks the wrapper.

## Suggested fix

Recursively unwrap `TargetInvocationException` before creating the diagnostic at `Compilation.cs:334`, preserving the original exception and source node.

This predates and is separate from #2935; #2935 only unwraps the direct indirect-call path.

### LINKED ISSUE #2999: Interpreter cannot catch TargetInvocationException
## Summary

A G# `catch (ex TargetInvocationException)` cannot match, including when user code explicitly throws that type.

## Reproduction

```gs
import System
import System.Reflection

let handler () -> int32 = () -> throw TargetInvocationException(
    \"outer\",
    InvalidOperationException(\"inner\"))

var caught = false
try {
    handler()
} catch (ex TargetInvocationException) {
    caught = true
}
caught
```

Expected: `true`.

Actual: the handler does not match; catch lookup sees the wrapped `InvalidOperationException` instead.

## Root cause

`TryFindCatchHandler` in `Evaluator.Statements.cs` calls `UnwrapRuntimeException(ex).GetType()` before matching catch types. `UnwrapRuntimeException` recursively strips every `TargetInvocationException`, so the catch matcher can never observe that runtime type. This behavior dates to #1649 and is not caused by #2972.

Fixing it likely requires separating evaluator/CLR infrastructure unwrapping from user-visible catch type matching, while preserving the existing reflection-wrapper behavior.

DavidObando:
Fixes #2938

## Summary

- Use a diagnostic-only `Evaluator.UnwrapDiagnosticException` when `Compilation.Evaluate` creates an uncaught interpreter `GS9999` diagnostic.
- Preserve the outer `EvaluatorException` and its rendered bound node; only the diagnostic message is taken from the unwrapped runtime exception.
- Preserve `EvaluatorException.DiagnosticId` and `Severity`, including deliberate interpreter diagnostics added by #3028.
- Preserve a user-thrown `TargetInvocationException` and its outer message while continuing to unwrap reflection-runtime transport wrappers recursively.
- Unwrap a single-inner `AggregateException` only on the diagnostic-message path when its unique child is a reflection-generated `TargetInvocationException`; catch matching remains unchanged.
- Keep `InvokeDelegate`'s existing `DynamicInvoke` catch unchanged. It remains load-bearing for direct/closure-routed delegate calls.

### User-authored `TargetInvocationException`

User-authored instances keep their outer message. Reflection-generated wrappers have a runtime/CoreLib target site; exceptions explicitly thrown by G# retain the evaluator/user target site. The uncaught-diagnostic path uses that distinction to unwrap only transport wrappers. A regression test now expects `author outer`, while a separate nested-reflection probe proves recursive runtime-wrapper unwrapping still reaches the real cause.

Main preserves user-authored wrappers during catch matching through #3023, which closed #2999. Catch matching continues to use `UnwrapRuntimeException` unchanged. The diagnostic-only helper reuses its wrapper classification without adding aggregate unwrapping to catch matching or handler-variable binding.

### `AggregateException`

The limitation is reachable. The regression probe sends a faulted `ValueTask[int32]` through a G# channel and awaits it; its authentic failure shape is built from a reflection invocation:

```text
AggregateException(TargetInvocationException(DivideByZeroException))
```

Before the aggregate change, the probe reported `One or more errors occurred...`; after it, `GS9999` reports `aggregate await boom`. A second test proves an aggregate with two inner exceptions remains aggregated.

Aggregate unwrapping is deliberately diagnostic-only. A catch regression test constructs `AggregateException(TargetInvocationException(DivideByZeroException))` and asserts that G# catches `AggregateException`, matching the `gsc` oracle. Direct TIE and two-inner aggregate probes remain unchanged.

### Helper relocation

The relocation remains. Returning the now-`internal` helper to its former position beside `TryFindCatchHandler` makes the enforced StyleCop build fail with `SA1202: 'internal' members should come before 'private' members`. Keeping it at the top of the partial class is the smallest warning-free change and avoids a local analyzer suppression.

## Root cause

Expression evaluation already attached source context with `EvaluatorException`, and typed catches already used recursive runtime-exception unwrapping. The final uncaught-diagnostic boundary instead read `EvaluatorException.Message` directly, so reflection-routed calls leaked `TargetInvocationException`'s generic message.

This is a message-only fix. The rendered diagnostic node is unchanged byte-for-byte. `gsi` prints only `ID: message`; the rich location renderer is used by the interactive REPL and `gsc`. No source location was lost or recovered.

Reflection-routed storage currently reports the call site, while closure-routed storage reports the lambda body. Tests document that known inconsistency; it is not a designed location contract.

## Reachability classification

Ten of the eleven reachable categories are pins. `await` was already correct on the base through `BlockOnValueTask` and is a guard.

| Category | Classification |
|---|---|
| Instance method | pin |
| Static method | pin |
| Constructor | pin |
| Property get | pin |
| Property set | pin |
| Indexer | pin |
| Operator | pin |
| Conversion | pin |
| Delegate | pin |
| Channel | pin |
| Await | guard — already correct on base |

Field access is excluded: the claimed path is unreachable because binding reports `GS0158`.

## Verification

- Rebased base: `61b7a08c1758c4b2df2714c295e03017fe19c169`.
- Solution build with `dotnet build GSharp.sln -p:ContinuousIntegrationBuild=true`: exit 0, 0 warnings, 0 errors.
- Build freshness gates passed in order: exit code, 96 nonzero `GSharp.Core.dll` copies, then matching MD5 across five ordinary Compiler/Core/Repl/test outputs.
- The user-authored-wrapper test is a base guard that preserves the behavior the interim PR had regressed.
- Final issue test class against head: 17 passed.
- #3028 diagnostic-ID/severity guard: 8 passed.
- Combined post-rebase targeted run: 25 passed.
- `gsc` and `gsi` agree on all three CLI probes: `caught-aggregate`, `caught-tie`, `caught-aggregate2`; every compile/run exits 0.

### Mutation checks

Each mutant built successfully before its test result was read:

- Restoring aggregate unwrapping to shared `UnwrapRuntimeException`: `SingleAggregateTargetInvocation_RemainsCatchableAsAggregateException` failed, expected `caught-aggregate`, actual `caught-dbz`.
- Pointing `Compilation` back at `UnwrapRuntimeException`: `AwaitedChannelValue_SingleAggregateTargetInvocation_UnwrapsInnerMessage` failed with the generic aggregate message.
- Removing the aggregate arm from `UnwrapDiagnosticException`: the same awaited-message test failed.
- Reverting the headline message hunk to `ex.Message`: 7 failed / 10 passed.

Restoring each hunk flipped its targeted test back to green. Full-suite counts from earlier bases were removed rather than mixed with the new base; current CI supplies merge-state coverage.

### DIFF
--- a/src/Core/CodeAnalysis/Compilation/Compilation.cs
+++ b/src/Core/CodeAnalysis/Compilation/Compilation.cs
@@ -330,7 +330,7 @@ public class Compilation
             ex.Node?.WriteTo(textWriter);
             var sourceText = SourceText.From(textWriter.ToString());
             var location = new TextLocation(sourceText, new TextSpan(0, sourceText.Length));
-            var message = ex.Message;
+            var message = Evaluator.UnwrapRuntimeException(ex).Message;
             var diagnostic = new Diagnostic(location, ex.DiagnosticId, ex.Severity, message);
             return new EvaluationResult(allWarnings.Add(diagnostic), null);
         }
--- a/src/Core/CodeAnalysis/Compilation/Compilation.cs
+++ b/src/Core/CodeAnalysis/Compilation/Compilation.cs
@@ -330,7 +330,7 @@ public class Compilation
             ex.Node?.WriteTo(textWriter);
             var sourceText = SourceText.From(textWriter.ToString());
             var location = new TextLocation(sourceText, new TextSpan(0, sourceText.Length));
-            var message = Evaluator.UnwrapRuntimeException(ex).Message;
+            var message = Evaluator.UnwrapDiagnosticException(ex).Message;
             var diagnostic = new Diagnostic(location, ex.DiagnosticId, ex.Severity, message);
             return new EvaluationResult(allWarnings.Add(diagnostic), null);
         }
--- a/src/Core/CodeAnalysis/Evaluator.Statements.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Statements.cs
@@ -33,8 +33,7 @@ public sealed partial class Evaluator
     /// nothing catches it). That wrapping must not leak into typed catch
     /// matching or handler-variable binding, so unwrap repeatedly down to the
     /// innermost real exception. TargetInvocationException (thrown by
-    /// reflection-based CLR calls) gets the same treatment, including when it
-    /// is the sole unambiguous child of an AggregateException.
+    /// reflection-based CLR calls) gets the same treatment.
     /// </summary>
     /// <param name="ex">Exception to unwrap.</param>
     /// <returns>Innermost runtime exception.</returns>
@@ -55,17 +54,27 @@ public sealed partial class Evaluator
                 continue;
             }
 
-            if (ex is AggregateException aggregate
-                && aggregate.InnerExceptions.Count == 1
-                && aggregate.InnerException is TargetInvocationException aggregateInner
-                && IsReflectionInvocationWrapper(aggregateInner))
-            {
-                ex = aggregateInner;
-                continue;
-            }
-
             return ex;
         }
+    }
+
+    /// <summary>
+    /// Unwraps runtime transport exceptions for diagnostic messages without
+    /// changing catch matching or handler-variable binding.
+    /// </summary>
+    /// <param name="ex">Exception to unwrap.</param>
+    /// <returns>Innermost diagnostic exception.</returns>
+    internal static Exception UnwrapDiagnosticException(Exception ex)
+    {
+        var unwrapped = UnwrapRuntimeException(ex);
+        if (unwrapped is AggregateException { InnerExceptions.Count: 1 } aggregate
+            && aggregate.InnerException is TargetInvocationException inner
+            && IsReflectionInvocationWrapper(inner))
+        {
+            return UnwrapRuntimeException(inner);
+        }
+
+        return unwrapped;
     }
 
     // Reflection-generated wrappers are thrown from CoreLib; a user-thrown
--- a/src/Core/CodeAnalysis/Evaluator.Statements.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Statements.cs
@@ -48,7 +48,8 @@ public sealed partial class Evaluator
                 continue;
             }
 
-            if (ex is TargetInvocationException { InnerException: { } tieInner })
+            if (ex is TargetInvocationException { InnerException: { } tieInner } tie
+                && IsReflectionInvocationWrapper(tie))
             {
                 ex = tieInner;
                 continue;
@@ -56,7 +57,8 @@ public sealed partial class Evaluator
 
             if (ex is AggregateException aggregate
                 && aggregate.InnerExceptions.Count == 1
-                && aggregate.InnerException is TargetInvocationException aggregateInner)
+                && aggregate.InnerException is TargetInvocationException aggregateInner
+                && IsReflectionInvocationWrapper(aggregateInner))
             {
                 ex = aggregateInner;
                 continue;
@@ -65,6 +67,12 @@ public sealed partial class Evaluator
             return ex;
         }
     }
+
+    // Reflection-generated wrappers are thrown from CoreLib; a user-thrown
+    // TargetInvocationException retains the user's/evaluator's target site.
+    private static bool IsReflectionInvocationWrapper(TargetInvocationException ex)
+        => ex.TargetSite?.DeclaringType?.Assembly is not { } assembly
+            || assembly == typeof(TargetInvocationException).Assembly;
 
     private object EvaluateStatement(BoundBlockStatement body)
     {
--- a/src/Core/CodeAnalysis/Evaluator.Statements.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Statements.cs
@@ -33,7 +33,8 @@ public sealed partial class Evaluator
     /// nothing catches it). That wrapping must not leak into typed catch
     /// matching or handler-variable binding, so unwrap repeatedly down to the
     /// innermost real exception. TargetInvocationException (thrown by
-    /// reflection-based CLR calls) gets the same treatment.
+    /// reflection-based CLR calls) gets the same treatment, including when it
+    /// is the sole unambiguous child of an AggregateException.
     /// </summary>
     /// <param name="ex">Exception to unwrap.</param>
     /// <returns>Innermost runtime exception.</returns>
@@ -50,6 +51,14 @@ public sealed partial class Evaluator
             if (ex is TargetInvocationException { InnerException: { } tieInner })
             {
                 ex = tieInner;
+                continue;
+            }
+
+            if (ex is AggregateException aggregate
+                && aggregate.InnerExceptions.Count == 1
+                && aggregate.InnerException is TargetInvocationException aggregateInner)
+            {
+                ex = aggregateInner;
                 continue;
             }
 
--- a/src/Core/CodeAnalysis/Evaluator.Statements.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Statements.cs
@@ -27,6 +27,36 @@ namespace GSharp.Core.CodeAnalysis;
 public sealed partial class Evaluator
 #pragma warning restore CA1001
 {
+    /// <summary>
+    /// Issue #1649: EvaluateExpression wraps every non-EvaluatorException in an
+    /// EvaluatorException to attach node context (for GS9999 reporting when
+    /// nothing catches it). That wrapping must not leak into typed catch
+    /// matching or handler-variable binding, so unwrap repeatedly down to the
+    /// innermost real exception. TargetInvocationException (thrown by
+    /// reflection-based CLR calls) gets the same treatment.
+    /// </summary>
+    /// <param name="ex">Exception to unwrap.</param>
+    /// <returns>Innermost runtime exception.</returns>
+    internal static Exception UnwrapRuntimeException(Exception ex)
+    {
+        while (true)
+        {
+            if (ex is EvaluatorException { InnerException: { } inner })
+            {
+                ex = inner;
+                continue;
+            }
+
+            if (ex is TargetInvocationException { InnerException: { } tieInner })
+            {
+                ex = tieInner;
+                continue;
+            }
+
+            return ex;
+        }
+    }
+
     private object EvaluateStatement(BoundBlockStatement body)
     {
         var labelToIndex = GetLabelToIndex(body);
@@ -699,40 +729,6 @@ public sealed partial class Evaluator
 
         throw new Exception(value?.ToString());
     }
-
-    /// <summary>
-    /// Issue #1649: EvaluateExpression wraps every non-EvaluatorException in an
-    /// EvaluatorException to attach node context (for GS9999 reporting when
-    /// nothing catches it). That wrapping must not leak into typed catch
-    /// matching or handler-variable binding, so unwrap repeatedly down to the
-    /// real exception. Reflection-generated TargetInvocationException wrappers
-    /// get the same treatment, while a user-thrown TargetInvocationException
-    /// remains visible to catch matching and handler-variable binding.
-    /// </summary>
-    private static Exception UnwrapRuntimeException(Exception ex)
-    {
-        while (true)
-        {
-            if (ex is EvaluatorException { InnerException: { } inner })
-            {
-                ex = inner;
-                continue;
-            }
-
-            if (ex is TargetInvocationException { InnerException: { } tieInner } tie
-                && IsReflectionInvocationWrapper(tie))
-            {
-                ex = tieInner;
-                continue;
-            }
-
-            return ex;
-        }
-    }
-
-    private static bool IsReflectionInvocationWrapper(TargetInvocationException ex)
-        => ex.TargetSite?.DeclaringType?.Assembly is not { } assembly
-            || assembly == typeof(TargetInvocationException).Assembly;
 
     private static bool TryFindCatchHandler(BoundTryStatement node, Exception ex, out BoundCatchClause matched)
     {
--- a/test/Interpreter.Tests/Issue2938DelegateExceptionDiagnosticTests.cs
+++ b/test/Interpreter.Tests/Issue2938DelegateExceptionDiagnosticTests.cs
@@ -233,6 +233,30 @@ public class Issue2938DelegateExceptionDiagnosticTests
     }
 
     [Fact]
+    public void SingleAggregateTargetInvocation_RemainsCatchableAsAggregateException()
+    {
+        const string Source = """
+            import System
+            import System.Reflection
+
+            var message = "none"
+            try {
+                throw AggregateException(TargetInvocationException(DivideByZeroException("boom")))
+            } catch (ex AggregateException) {
+                message = "caught-aggregate"
+            } catch (ex DivideByZeroException) {
+                message = "caught-dbz"
+            }
+            message
+            """;
+
+        var result = Evaluate(Source);
+
+        Assert.Empty(result.Diagnostics);
+        Assert.Equal("caught-aggregate", result.Value);
+    }
+
+    [Fact]
     public void ClrMemberDelegateDiagnostic_RemainsAnchoredAtCallSite()
     {
         const string Source = """
--- a/test/Interpreter.Tests/Issue2938DelegateExceptionDiagnosticTests.cs
+++ b/test/Interpreter.Tests/Issue2938DelegateExceptionDiagnosticTests.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Reflection;
+using System.Threading.Tasks;
 using GSharp.Core.CodeAnalysis;
 using GSharp.Core.CodeAnalysis.Compilation;
 using GSharp.Core.CodeAnalysis.Symbols;
@@ -22,6 +23,8 @@ public class Issue2938DelegateExceptionDiagnosticTests
     /// <summary>
     /// Gets delegate-storage cases that cover CLR and interpreter containers.
     /// </summary>
+    // Reflection-routed storage reports the call site, while closure-routed
+    // storage reports the lambda body. This is a known inconsistency, not a contract.
     public static TheoryData<string, string, string> StorageCases => new()
     {
         {
@@ -180,6 +183,56 @@ public class Issue2938DelegateExceptionDiagnosticTests
     }
 
     [Fact]
+    public void UserThrownTargetInvocationException_IsIntentionallyUnwrapped()
+    {
+        const string Source = """
+            import System
+            import System.Reflection
+
+            let handler () -> int32 = () -> throw TargetInvocationException(
+                "author outer",
+                InvalidOperationException("author inner"))
+            handler()
+            """;
+
+        var diagnostic = Assert.Single(Evaluate(Source).Diagnostics);
+
+        Assert.Equal("GS9999", diagnostic.Id);
+        Assert.Equal("author inner", diagnostic.Message);
+    }
+
+    [Fact]
+    public void AwaitedChannelValue_SingleAggregateTargetInvocation_UnwrapsInnerMessage()
+    {
+        const string Source = """
+            import GSharp.Interpreter.Tests
+            import Gsharp.Extensions.Go
+            import System.Threading.Tasks
+
+            let ch = make(chan ValueTask[int32], 1)
+            ch <- Issue2938ExceptionProbe.CreateSingleAggregateValueTaskAsync()
+            await <-ch
+            """;
+
+        AssertDiagnostic(Source, "aggregate await boom", "await <-ch");
+    }
+
+    [Fact]
+    public void MultipleAggregateExceptions_RemainAggregate()
+    {
+        const string Source = """
+            import GSharp.Interpreter.Tests
+
+            await Issue2938ExceptionProbe.CreateMultipleAggregateValueTaskAsync()
+            """;
+
+        AssertDiagnostic(
+            Source,
+            "One or more errors occurred. (Exception has been thrown by the target of an invocation.) (second aggregate boom)",
+            "await GSharp.Interpreter.Tests.Issue2938ExceptionProbe.CreateMultipleAggregateValueTaskAsync()");
+    }
+
+    [Fact]
     public void ClrMemberDelegateDiagnostic_RemainsAnchoredAtCallSite()
     {
         const string Source = """
@@ -227,8 +280,6 @@ public class Issue2938DelegateExceptionDiagnosticTests
         Assert.Equal(DiagnosticSeverity.Error, diagnostic.Severity);
         Assert.Equal(message, diagnostic.Message);
         Assert.Equal(sourceNode, diagnostic.Location.Text.ToString(diagnostic.Location.Span));
-        Assert.Equal(0, diagnostic.Location.StartLine);
-        Assert.Equal(0, diagnostic.Location.StartCharacter);
     }
 
     private static EvaluationResult Evaluate(string source)
@@ -265,4 +316,25 @@ public static class Issue2938ExceptionProbe
         => new InvalidOperationException(
             "outer boom",
             new DivideByZeroException("inner boom"));
+
+    /// <summary>
+    /// Creates an awaitable whose result throws a single-inner aggregate wrapper.
+    /// </summary>
+    /// <returns>Awaitable exception probe.</returns>
+    public static ValueTask<int> CreateSingleAggregateValueTaskAsync()
+        => new(Task.FromException<int>(
+            new AggregateException(
+                new TargetInvocationException(
+                    new DivideByZeroException("aggregate await boom")))));
+
+    /// <summary>
+    /// Creates an awaitable whose result throws a multi-inner aggregate wrapper.
+    /// </summary>
+    /// <returns>Awaitable exception probe.</returns>
+    public static ValueTask<int> CreateMultipleAggregateValueTaskAsync()
+        => new(Task.FromException<int>(
+            new AggregateException(
+                new TargetInvocationException(
+                    new DivideByZeroException("aggregate await boom")),
+                new InvalidOperationException("second aggregate boom"))));
 }
--- a/test/Interpreter.Tests/Issue2938DelegateExceptionDiagnosticTests.cs
+++ b/test/Interpreter.Tests/Issue2938DelegateExceptionDiagnosticTests.cs
@@ -153,17 +153,17 @@ public class Issue2938DelegateExceptionDiagnosticTests
     }
 
     [Fact]
-    public void NestedTargetInvocationExceptions_UnwrapRecursively()
+    public void NestedReflectionTargetInvocationExceptions_UnwrapRecursively()
     {
         const string Source = """
             import GSharp.Interpreter.Tests
 
-            let handler () -> int32 = () -> Issue2938ExceptionProbe.ThrowDoubleWrapped()
+            let handler () -> int32 = () -> Issue2938ExceptionProbe.InvokeNullReferenceReflectively()
             let pair = (handler, 0)
             pair.Item1()
             """;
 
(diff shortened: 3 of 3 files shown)


DavidObando:
## Rubber-duck verdict (Opus 5): MERGEABLE

Reviewed head `f816c794` against base `8dddbbf2`, in isolated clones under `/tmp/d2938/` (`git clone --local`, full git metadata). No blockers. 2 SHOULD-FIX, 4 NIT — none gate the merge.

**Base discipline:** every number below is measured on `8dddbbf2` (base) and `f816c794` (head) — *same base, both executed, never `--list-tests`*. `main` has since moved to `a6b1561f2` (#2961/#2921). I did **not** re-measure on `a6b1561f2`; the counts below must not be compared against a table taken there.

---

## The three questions

### (a) Are nine claimed fixed paths real, from a 31-line diff?

**Yes — and the diff is smaller than it looks.** The true production delta is **2 lines**:

- `src/Core/CodeAnalysis/Compilation/Compilation.cs:333` — `ex.Message` → `Evaluator.UnwrapRuntimeException(ex).Message`
- `src/Core/CodeAnalysis/Evaluator.Statements.cs` — `private static` → `internal static` on `UnwrapRuntimeException`

The `+30/−28` in `Evaluator.Statements.cs` is a **pure relocation**. Proven by set-diffing added against removed lines: identical except the visibility token and two StyleCop-mandated XML doc lines. This is the /ponytail floor — there is no smaller correct fix, because `Compilation.cs:333` is the **sole** GS9999 construction site in the interpreter, and every raw `.Invoke(...)` without a local unwrap converges on it (`Evaluator.Expressions.cs:202, 224, 315, 1146, 1159, 1209, 1219, 1559, 1565, 1571`).

Nine claimed categories, each probed with its own executed G# program at base and head:

| category | base | head | classification |
|---|---|---|---|
| instance method | GS9999 generic | correct | **pin** |
| static method | GS9999 generic | correct | **pin** |
| constructor | GS9999 generic | correct | **pin** |
| property get | GS9999 generic | correct | **pin** |
| property set | GS9999 generic | correct | **pin** |
| indexer | GS9999 generic | correct | **pin** |
| operator | GS9999 generic | correct | **pin** |
| conversion | GS9999 generic | correct | **pin** |
| delegate | GS9999 generic | correct | **pin** |
| channel | GS9999 generic | correct | **pin** |
| **await** | **already correct** | correct | **guard, not a fix** |
| field access | n/a — `GS0158`, not reachable | — | not a path |

**10 of 11 are genuine. `await` is not** — it was already correct at base via the `BlockOnValueTask` unwrap at `Evaluator.Statements.cs:628`. The claim "await … fixed" is an overclaim; it is regression protection. Field access is not reachable at all. Please restate those two in the PR body.

### (b) Is `Evaluator.Expressions.cs:872` correctly untouched?

**Yes, and touching it would have been wrong.** It is `InvokeDelegate`'s `DynamicInvoke` catch, which rethrows via `ExceptionDispatchInfo`. It is precisely *why* direct `handler(0)` was already correct at base while `pair.Item1(0)` was not: shapes routing through the interpreter's own closure invocation (array element, struct field, class field, function return, direct call) hit this site and were already fine; shapes routing through reflection (tuple `.Item1`, `List[T]` indexer) did not, and are the ones the PR fixes.

I mutated it anyway (see M7). It survives the suite, but **survival is not evidence it is dead** — I could construct no program distinguishing it, including `ex.StackTrace` preservation and typed `catch` across `await`. **Do not delete it.** It is a request for a test, not a deletion candidate.

### (c) Does this collide with #2956?

**No.** `#2956` is an **issue**, not a pull request (`gh pr view 2956` → *"Could not resolve to a PullRequest"*), so nothing owns `Evaluator.Statements.cs` in flight. I scanned all 40 open PRs by file list: **#2972 is the only open PR that touches `Evaluator.Statements.cs`.**

`git merge-tree --write-tree --name-only origin/main origin/fix/2938-interpreter-unwrap-target-invocation` against the **new** `main` `a6b1561f2` → tree `5b6c7ae5`, exit 0, **clean**. Also clean at review time against #2912, #2929, #2961, #2963, #2966, #2967, #2968. Zero file overlap with #2963 (`StatementBinder.*`) and #2967 (`MethodBodyEmitter.Statements.cs`), confirmed via `gh pr view --json files`.

---

## Pins vs guards: **6 pins / 7 guards** (PR body claims 2/4)

Derived independently at semantic-branch granularity from ~25 hand-written G# probes run at both base and head. Corroborated by mutation M1, which fails **exactly 6** — matching the pin count exactly.

**Recursion is separately pinned**, as requested: M3 (degrade to single-level unwrap) fails **3** tests, so a double-wrapped `TargetInvocationException` is genuinely covered and is not two flavours of the same level.

---

## Mutation table

7 semantic branches. Methodology hardened mid-review after a fleet-wide alert and **re-run three times**; the table below is the final `-m:1` pass.

| id | mutation | site | result |
|---|---|---|---|
| M1 | `UnwrapRuntimeException(ex)` → `ex` | `Compilation.cs:333` | **detected — 6 failures** |
| M3 | recursive unwrap → single-level | `Evaluator.Statements.cs:30-57` | **detected — 3 failures** |
| M6 | disable `BlockOnValueTask` TIE catch | `Evaluator.Statements.cs:628` | **survived** |
| M7 | disable `InvokeDelegate` TIE catch | `Evaluator.Expressions.cs:872` | **survived** |

### `-m:1` re-confirmation (strongest evidence in this file)

The two survived mutants were re-run under a hardened harness: build the **test project** (not `src/Core`) with `-m:1`, gate on **exit code → zero-byte → propagation**, assert the md5 of `out/bin/Release/Interpreter.Tests/GSharp.Core.dll` both *changed* and *equals* the Core output's, then `--no-build`.

| probe | BUILD_EXIT | test-copy md5 | result |
|---|---|---|---|
| C0 control | 0 / 0 Error(s) | `4b375ded` | 502 / 0 |
| M7 `Expressions.cs:872` | 0 | `4b375ded` → `1f48e0c1` | 502 / 0 |
| M6 `Statements.cs:628` | 0 | `1f48e0c1` → `08c1d2c9` | 502 / 0 |
| C0 final | 0 | `08c1d2c9` → `4b375ded` | 502 / 0 |

Exact **bidirectional round-trip** `4b375ded → mutant → 4b375ded` — a self-verifying proof that the binary under test is the one intended, which matters precisely for survived mutants where by definition no observable output changes. All md5s byte-identical to the earlier parallel-build run, so the build is deterministic when it succeeds.

**Both mutants are genuinely undetected — and both are requests for tests, not licence to delete.**

---

## Findings

### SHOULD-FIX

**S1 — user-authored `TargetInvocationException` is over-unwrapped, and untested.**
`src/Core/CodeAnalysis/Evaluator.Statements.cs:49-53`. A G# program that itself throws a `TargetInvocationException` wrapping a cause will surface the **inner** message, losing the author's outer message. No test covers it. Add one, and decide deliberately whether the current behaviour is intended.

**S2 — `AggregateException` wrapping a `TargetInvocationException` still leaks the generic message.**
The unwrap steps through `TargetInvocationException` only, so `AggregateException(TargetInvocationException(DivideByZeroException))` still reports the wrapper. Reachable from the await/channel paths the PR claims to cover. Either extend the unwrap or state the limitation.

### NIT

**N1 — two vacuous assertions.** `test/Interpreter.Tests/Issue2938DelegateExceptionDiagnosticTests.cs:229-230` assert on `StartLine`/`StartCharacter`. The GS9999 location is **synthetic**: `Compilation.cs:331-332` renders `ex.Node.WriteTo(...)` into a fresh `SourceText` and spans `TextSpan(0, len)`, so both are **always 0** regardless of correctness. These assertions cannot fail. Remove or replace with an assertion on the rendered node text.

**N2 — reported location differs by storage shape,** and the test encodes that as expected. Reflection-routed shapes report the *call site*; closure-routed shapes report the *lambda body*. Worth a comment noting it is a known inconsistency rather than a designed contract.

**N3 — the 670-line relocation is unnecessary churn.** It makes the diff look 15× larger than its 2 semantic lines and complicates future `git blame`. Moving the method was not required to change its visibility.

**N4 — pre-existing, not caused by this PR:** `catch (ex TargetInvocationException)` can never match in G#. `TryFindCatchHandler` (`Evaluator.Statements.cs:727`) types on the **already-unwrapped** exception. Dates to #1649. Filing separately is appropriate; do not scope-creep it here.

---

1. **"Source location is half the bug" is not supported.** The rendered `NODE` is **byte-identical at base and head in all 20 probes** — the fix changes the message only. Moreover `gsi` **never prints a location at all**: `src/Repl/Program.cs:60` is `Console.Error.WriteLine($"{d.Id}: {d.Message}")`. The rich renderer (`src/Core/IO/TextWriterExtensions.cs:115-175`) is reached only by the interactive REPL and `gsc`. There was no location to lose, and none was recovered. The PR should not claim otherwise.

2. **"Something deriving from `TargetInvocationException`" is impossible** — `System.Reflection.TargetInvocationException` is `sealed`. That over-unwrap sub-case does not exist. (Attempting it yields `CS0509`.)

3. **Pins/guards is 6/7, not 2/4** — see above.

---

## Independent executed counts

**Base `8dddbbf2`:** Core 7028 + 1 skipped · Compiler 3893 · Interpreter 489 · LSP 452
**Head `f816c794`:** Core 7028 + 1 skipped · Compiler 3893 · Interpreter **502** · LSP 452

All executed, all green, `Failed: 0` on every suite. `out/bin/Release/Gsharp.Extensions/` populated, so the golden-sample tests genuinely ran — this is a true zero, not 19 failures written off as environmental. Delta is **+13 Interpreter, nothing else moved**, exactly consistent with a single 268-line interpreter test file and exactly matching the author's report.

Corpus: 148 samples, **zero** relevant to this change — non-regression evidence only. The case rests entirely on the hand-written probes above.

Compiled-vs-interpreted parity on the message is restored, verified by executing the same programs both ways.

DavidObando:
# Review — PR #2972 @ `b79d6e067` — **NOT MERGEABLE** (1 blocker)

Merge base computed, not assumed: `git merge-base refs/remotes/pr/2972 origin/main` = **`a6b1561f2`**. Three-dot delta = `Compilation.cs +1/-1`, `Evaluator.Statements.cs +67/-29`, new test file `+340`.

Declared fix works. But the helper change is **not** message-only, and it silently kills a live `catch` handler.

---

## BLOCKER-1 — new `AggregateException` arm changes `catch` matching + handler binding; a handled program becomes an unhandled `GS9998`

`Evaluator.Statements.cs` — the new arm sits inside `UnwrapRuntimeException`, which already had **two** consumers on `main`:

```
:623  Assign(handler.Variable, UnwrapRuntimeException(ex));     <- handler variable binding
:725  var actualType = UnwrapRuntimeException(ex).GetType();    <- TryFindCatchHandler, catch TYPE MATCHING
```

`grep -rn UnwrapRuntimeException src/` on `a6b1561f2` = 3 hits, exactly those two plus the definition. The PR adds a third (`Compilation.cs:333`) **and** changes what the function returns. So the arm reaches catch semantics.

`TryFindCatchHandler` (`:729-741`) walks `actualType.BaseType`. For `AggregateException(TargetInvocationException(DivideByZeroException))`:
- base → `actualType = AggregateException`
- head → arm fires → TIE arm fires → `actualType = DivideByZeroException`

### Probe evidence

Two `--shared` scratch clones, explicit `git checkout <sha>`, `dotnet build src/Compiler/Compiler.csproj -c Release -p:ContinuousIntegrationBuild=true`, exit code 0 both, gate-2 all `GSharp.Core.dll` copies nonzero, gate-3 md5 uniform across the ordinary paths (base `3f28d46d…`, head `241e4018…` — distinct, so two real builds). `gsc <file>.gs` interprets, so it drives the exact `Compilation.Evaluate` path.

| probe | source shape | base `a6b1561f2` | head `b79d6e067` |
|---|---|---|---|
| **p1** #2938 repro, uncaught | `let pair = (handler, 0)` / `pair.Item1(0)` | `GS9999: Exception has been thrown by the target of an invocation.` | `GS9999: Attempted to divide by zero.` ✅ **fix works** |
| **p2** `catch (ex AggregateException)` then `catch (ex DivideByZeroException)` | `throw AggregateException(TargetInvocationException("t", DivideByZeroException("boom")))` | `caught-aggregate` | **`caught-dbz`** ❌ different arm runs |
| **p3** `catch (ex Exception) { ex.GetType().Name }` | same | `AggregateException` | **`DivideByZeroException`** ❌ different object bound |
| **p4** multi-inner guard | `AggregateException(dbz, ioe)` | `caught-aggregate` | `caught-aggregate` ✅ unchanged |
| **p5** **only** `catch (ex AggregateException)` | same as p2 | `HANDLED`, **exit 0** | **`error GS9998: AggregateException: One or more errors occurred. (t)`, exit 1** ❌❌ |

**p5 is the blocker.** On head the handler no longer matches, the `AggregateException` escapes the `try`, and it is not an `EvaluatorException`, so `Compilation.cs:327` `catch (EvaluatorException ex)` does not catch it either. It leaves `Compilation.Evaluate` entirely and surfaces as **`GS9998`** — the host-crash diagnostic — not a program `GS9999`. Working G# goes from exit 0 to exit 1 with a compiler-crash-class error.

### Perturbation — the arm is the sole cause, and #2938's fix does not need it

Removed **only** the 8-line arm from the head clone, rebuilt (exit 0, core md5 `241e4018…` → `7c260d27…`):

| probe | head | head − arm |
|---|---|---|
| p1 | `Attempted to divide by zero.` | **`Attempted to divide by zero.`** — still fixed |
| p2 | `caught-dbz` | `caught-aggregate` — restored |
| p3 | `DivideByZeroException` | `AggregateException` — restored |
| p5 | `GS9998`, exit 1 | `HANDLED`, exit 0 — restored |

Restored, rebuilt: md5 back to **`241e401812f35d3fd6b9aa834f896b10`**, p2 → `caught-dbz`, p5 → `GS9998`. Round trip closed, both trees `git status --porcelain` empty.

So: **the arm is not load-bearing for the declared issue #2938**, and it is the entire cause of the regression.

### Recommended fix — put the aggregate unwrap at the diagnostic boundary, not in the shared helper

Revert the arm in `Evaluator.Statements.cs`; do it in `Compilation.cs` where the scope actually is message-only:

```csharp
var runtime = Evaluator.UnwrapRuntimeException(ex);
if (runtime is AggregateException { InnerExceptions.Count: 1 } agg
    && agg.InnerException is TargetInvocationException)
{
    runtime = Evaluator.UnwrapRuntimeException(agg.InnerException);
}

var message = runtime.Message;
```

`AwaitedChannelValue_SingleAggregateTargetInvocation_UnwrapsInnerMessage` and `MultipleAggregateExceptions_RemainAggregate` both still pass; p2/p3/p5 all stay at base behaviour; nothing about `catch` moves.

If instead the arm is kept deliberately, it is a **language semantics change** and needs: an ADR note, a `catch (ex AggregateException)` regression test, and the "message-only" claim deleted.

---

## BLOCKER-2 (policy) — the PR body states a falsified sentence

> This is a message-only fix.

p2/p3/p5 falsify it. `catch` arm selection and handler-variable identity both change.

This matters more, not less, because the body already shows the author knew the helper feeds catch lookup:

> The related pre-existing catch inconsistency is tracked separately in #2999: catch lookup unwraps before type matching…

Same function, one paragraph apart. No CI job can see a false sentence; it has to be fixed by hand.

---

## SHOULD-FIX-1 — zero coverage for the catch-side effect

`TypedCatch_StillReceivesRealException` (`:252`) is the only catch test and it exercises the **pre-existing** `EvaluatorException` unwrap (`CreateWithOrdinaryInner` → `InvalidOperationException`), never the new arm. `grep -rn AggregateException test/` returns no G#-source test that does `catch (ex AggregateException)` — so nothing in CI can observe this. If the arm survives review, p2/p3/p5 must become tests.

## SHOULD-FIX-2 — this makes #2999 harder, and a test name pins #2999's bug as intended

Probed: **#2972 does not fix or half-fix #2999.** p6 (`throw TargetInvocationException(...)`, `catch (ex TargetInvocationException)` before `catch (ex InvalidOperationException)`) → `caught-ioe` on **both** base and head. Declaration is honest — mentioned, not claimed. But:

1. #2999's stated remedy is *separating* infra unwrapping from user-visible catch matching. This PR moves the opposite way: third consumer added, `private` → `internal`, now cross-file and cross-namespace. The split must renegotiate 3 sites instead of 2.
2. The three consumers now want **contradictory** policies. This PR's own §"User-authored `TargetInvocationException`" argues `GS9999` *must* unwrap a user-thrown TIE; #2999 requires catch matching *must not*. One function cannot satisfy both.
3. `Issue2938DelegateExceptionDiagnosticTests.cs:186` is named `UserThrownTargetInvocationException_IsIntentionallyUnwrapped` with **no `#2999` reference in the test**. A #2999 implementer will read "Intentionally" as contract.

Fix: add `// Behaviour under review in #2999.` to that test, and say in the body that #2999's split now spans 3 call sites in 2 files.

## NIT-1 — the aggregate message fix is asymmetric and the asymmetry is pinned

`MultipleAggregateExceptions_RemainAggregate` pins `"One or more errors occurred. (Exception has been thrown by the target of an invocation.) (second aggregate boom)"` — the exact TIE noise #2938 exists to remove, still there for n > 1. "No single cause is unambiguous" justifies not picking one *type*; it does not justify leaving the *message* unreadable, which could unwrap each inner and rebuild. Worth a follow-up issue rather than a silent pin.

## NIT-2 — 6 locations pinned, 4 disagreeing with 2, no issue filed

`StorageCases` (`:28-91`) pins `tuple`/`list` at the call site and `array`/`struct-field`/`class-field`/`returned` at the lambda body, with an inline "known inconsistency, not a contract". A pin with no tracking issue becomes the contract by default. File it.

---

## Verified good — do not re-litigate

- `Compilation.cs:334` is the **sole** `GS9999` construction site (`grep -rn '"GS9999"' src/` → 1 hit). Right layer for a boundary fix.
- **"Rendered node unchanged byte-for-byte" is true.** `ex.Node?.WriteTo(...)` at `:328` runs before the unwrap and `ex` is never reassigned. p1 location is `pair.Item1.Invoke(0)` on both sides.
- **`SA1202` relocation claim holds.** `Evaluator` is `public sealed partial` in `GSharp.Core.CodeAnalysis`, `Compilation` in `GSharp.Core.CodeAnalysis.Compilation`, same assembly ⇒ `internal` is the minimum accessibility; SA1202 then forces it above the next member, `private object EvaluateStatement`. Not gratuitous churn.
- **`+16` arithmetic checks out without running a suite:** 6 `StorageCases` rows + 10 `[Fact]` = 16, matching 491 → 507 exactly.
- **No infinite loop.** Every arm strictly descends an `InnerException` chain, which cannot be cyclic (get-only, fixed at construction). `AggregateException(TargetInvocationException(null))` terminates on the TIE.
- **No `Gsharp.Extensions` closure gap.** `Interpreter.Tests.csproj:11` references `Gsharp.Extensions.csproj` directly, and 3 existing files already `import Gsharp.Extensions.Go`.
- **`gsi` self-containment is satisfied.** `src/Repl/Program.cs:57-61` prints `{d.Id}: {d.Message}` only, so the message *is* the entire user-visible diagnostic there — precisely the string this PR improves. And the "location" is a synthesized `SourceText` built from the bound-node dump (`Compilation.cs:327-329`), not a position in the user's file, so no renderer could have shown a real source position. "No source location was lost or recovered" is accurate.
- **Textually isolated.** Neither `Compilation/Compilation.cs` nor `Evaluator.Statements.cs` is touched by any of the 11 other open PRs (3009, 3005, 3000, 2998, 2997, 2979, 2968, 2967, 2963, 2912, 2903). Zero merge conflicts predicted.

---

Ship it with the arm moved to `Compilation.cs` and the "message-only" sentence corrected, and this is a clean, well-scoped fix.

<sub>Throttle honoured: `uptime` gated every build; probes run at 1-min load < 30. Two single-project Release builds and six `gsc` invocations only — no test suite. No `git stash`, nothing written to `~/.nuget/packages/gsharp.net.sdk/`, all scratch under `/tmp/g2972/`, all trees clean.</sub>

DavidObando:
## PR #2972 — **NOT MERGEABLE**

One blocker: a **silent catch-matching regression** vs both the `gsc` oracle and `main`. Verified drop-in fix below (~19 lines, keeps all 16 of this PR's tests green).

### Trees measured

Head is **`d2174ad299f654332ff47a729a96c7fec1d4c8da`**, not the `051f71954` I was briefed — please confirm which you expect. Base `286ed0825` (computed `merge-base`), `main` `a4a660f32`, merged tree **`8b2c8055e`** (`git merge` rc=0, 3 files).

Three-dot = **3 files / +407 −35**. (Two-dot reports 13 files / +443 −623 — phantom.)

### True semantic delta is 3 things, not +81/−35

`UnwrapRuntimeException` is **moved**, not rewritten. Diffing the method body base-vs-head, normalised:

1. `Compilation.cs:333` — `ex.Message` → `Evaluator.UnwrapRuntimeException(ex).Message`
2. `private static` → `internal static`
3. **new 8-line `AggregateException` arm**

`main` == base for this method byte-for-byte, so there is **no #3023 conflict here**; #3023's other two sites (`Evaluator.Expressions.cs:893`, `Evaluator.Statements.cs:606`) are separate `catch` blocks and are untouched. That risk is clear.

---

## BLOCKER — `src/Core/CodeAnalysis/Evaluator.Statements.cs:53-60`

The new arm is added to the **shared** helper, which has two pre-existing consumers that govern **catch semantics**, not messages:

- `:670` `Assign(handler.Variable, UnwrapRuntimeException(ex))` — handler-variable binding
- `:744` `TryFindCatchHandler` → `UnwrapRuntimeException(ex).GetType()` — **catch type matching**

So the arm silently changes which `catch` clause runs. Probe (CoreLib only — runs through the `gsi` CLI, no external assembly, so it is not invalid under the #2995 rule):

```gs
import System
import System.Reflection
var message = "none"
try {
    throw AggregateException(TargetInvocationException(DivideByZeroException("boom")))
} catch (ex AggregateException) {
    message = "caught-aggregate"
} catch (ex DivideByZeroException) {
    message = "caught-dbz"
}
Console.WriteLine(message)
```

| probe | `gsc` (oracle) | `gsi` @ `main` a4a660f32 | `gsi` @ merged #2972 8b2c8055e |
|---|---|---|---|
| `AggregateException(TIE(DivideByZero))` | `caught-aggregate` | `caught-aggregate` | **`caught-dbz`** |
| `TIE(DivideByZero)` | `caught-tie` | `caught-tie` | `caught-tie` |
| `AggregateException(TIE(...), InvalidOperation)` | `caught-aggregate2` | `caught-aggregate2` | `caught-aggregate2` |

**All six runs exit 0.** Wrong branch taken, nothing reports it — the silent class exactly.

### Why `main` is safe and this PR is not

`IsReflectionInvocationWrapper` (`:69`) returns **true when `TargetSite` is null**:

```csharp
=> ex.TargetSite?.DeclaringType?.Assembly is not { } assembly
    || assembly == typeof(TargetInvocationException).Assembly;
```

A TIE that is *thrown* from G# has a `TargetSite` in `GSharp.Core` → correctly excluded. That is why `tie.gs` is right on both trees, and why the author's `UserThrownTargetInvocationException_KeepsOuterMessage` passes.

A TIE that is **constructed and nested inside an `AggregateException` was never thrown**, so `TargetSite` is `null` → classified as a reflection wrapper → unwrapped. **The new arm is the first code path that ever inspects a never-thrown TIE**, so it activates a latent hole in a guard that was previously sound. The existing guard test cannot catch this because it tests the *thrown* shape.

### Mutation — each hunk individually

| mutant | build | result |
|---|---|---|
| **M-1** delete the `AggregateException` arm | rc=0, 0 errors | rc=1, **1 failed / 15 passed** — only `AwaitedChannelValue_SingleAggregateTargetInvocation_UnwrapsInnerMessage` |
| **M-2** revert `Compilation.cs:333` to `ex.Message` | rc=0, 0 errors | rc=1, **7 failed / 9 passed** |

M-2 shows the headline hunk is **well pinned**. M-1 is the diagnostic finding: the arm is pinned by **exactly one test, and that test asserts a diagnostic message** — nothing requires the arm on the catch path. It is there only because it was put in the shared helper.

### Verified fix — drop-in, measured

Remove the arm from `UnwrapRuntimeException`; add a message-path-only helper; point `Compilation.cs:333` at it.

```csharp
/// <summary>Diagnostic-reporting unwrap: adds the single-inner AggregateException
/// step, which must NOT participate in catch matching or handler-variable binding.</summary>
internal static Exception UnwrapDiagnosticException(Exception ex)
{
    var unwrapped = UnwrapRuntimeException(ex);
    if (unwrapped is AggregateException { InnerExceptions.Count: 1 } aggregate
        && aggregate.InnerException is TargetInvocationException inner
        && IsReflectionInvocationWrapper(inner))
    {
        return UnwrapRuntimeException(inner);
    }

    return unwrapped;
}
```

`Compilation.cs:333` → `Evaluator.UnwrapDiagnosticException(ex).Message;`

**Placement warning:** it must go *before* `IsReflectionInvocationWrapper`. I first placed it after and the build failed `rc=1` with `SA1202: 'internal' members should come before 'private' members` — the same analyzer already documented in your body under "Helper relocation".

Measured with this applied:

- build rc=0, 0 errors
- `Issue2938` filter: **rc=0, 16 passed / 0 failed** — including `AwaitedChannelValue_...`, so #2938's aggregate case still works
- `agg.gs` → **`caught-aggregate`**, matching the `gsc` oracle and `main`
- `tie.gs`, `agg2.gs` unchanged

### Round trip

Clean `GSharp.Core.dll` md5 `9fbc89a41f087c60716ae70ee3fd9612` → restored **byte-identical**; `git status --porcelain` empty; restored `gsi` reproduces `caught-dbz`, i.e. the PR behaviour. Both mutants built rc=0, so neither is a bad-mutant no-result. `Issue3004` (#3028's `GS0513`) 8/8 on the merged tree — no cross-regression there.

---

## SHOULD-FIX

- **`test/Interpreter.Tests/Issue2938DelegateExceptionDiagnosticTests.cs`** — across all 359 added lines, exactly **one** catch shape is exercised in G# source, `catch (ex InvalidOperationException)` at `:263`. **No test catches `AggregateException` from G#.** That is precisely the hole the blocker fell through. Add the probe above as a test; with the fix it asserts `caught-aggregate`, and it would have failed the current head.

- **PR body, "User-authored `TargetInvocationException`" section** — "This PR reuses that same wrapper classification at the diagnostic boundary, so **catches and uncaught diagnostics agree**". Measurably false at this head: the arm is *not* confined to the diagnostic boundary, and catches now disagree with both `gsc` and `main`. With the fix applied the sentence becomes true; please land the fix or narrow the sentence.

- **`MultipleAggregateExceptions_RemainAggregate`** proves the *2-inner* aggregate survives on the **message** path only. There is no test that a *1-inner* aggregate remains **catchable** as `AggregateException`. That is the untested half.

## NIT

- Head SHA in the assignment (`051f71954`) does not match the PR (`d2174ad29`).
- `Fixes #2938` present and accurate; scope matches the diff (3 files, no undeclared extras).
- Docs gate **not** exposed: 0 descriptors added (`new("GS…")` — positive control 377 in `DiagnosticDescriptors.cs`), 0 `diagnostics.md` files touched. The 4 `GS\d{4}` lines in the diff are one comment and two `Assert.Equal("GS9999", …)`. No row needed in `docs/diagnostics.md` or `website/docs/ref/diagnostics.md`.

The `Compilation.cs` hunk and the helper relocation are correct and well tested — the blocker is confined to where the aggregate step was placed.

Issue #2948, opened 2026-07-31
gsc: preserve imported named-delegate identity through deferred erased-slot lambda targeting


DavidObando:
## Problem

Deferred lambda target recovery through an erased imported-generic slot currently stores only `FunctionTypeSymbol.ParameterTypes` / return shape in `TryMapDeferredLambdaTargetsSymbolic`. That is enough for canonical `System.Action` / `System.Func`, whose identity can be reconstructed from shape, but loses nominal identity for imported named delegates such as:

- `Predicate<T>`
- `Comparison<T>`
- `Converter<TInput,TOutput>`
- `EventHandler<TEventArgs>`
- user-library CLR delegates

PR #2929 now conservatively rejects these inline cases instead of emitting a canonical `Func`/`Action` with the wrong runtime type. Hoisted named delegates remain correct.

## Repros

```gsharp
let predicates = List[Predicate[Src]]()
predicates.Add((item Src) -> item.N > 2)
```

```gsharp
let box = DelegateBox[Predicate[Src]]((item Src) -> item.N > 2)
```

Without the conservative guard, the first stores `Func<Src,bool>` into `List<Predicate<Src>>` and throws `ArgumentException`; the second emits a `Func<Src,bool>` where `Predicate<Src>` is required and can produce unverifiable IL / silent type corruption.

## Expected

Inline and hoisted forms both preserve the exact imported delegate type and run as `Predicate<Src>` / `Comparison<Src>` rather than falling back to diagnostics.

## Fix direction

Carry the mapped delegate `TypeSymbol` identity through `slotTargets` and agreement logic, not only its parameter/return shape. Bind the deferred lambda against that nominal delegate target while retaining existing candidate-disagreement and symbolic-projection gates. Add positive IL/load/execute tests for collection and constructor slots across several named delegates.

## Temporary guard caveat

PR #2929's conservative `Action` / `Func` check compares CLR `FullName` only. It is therefore assembly-blind: a user-library delegate with a canonical `System.Func` full name could pass the temporary check. Proper nominal `TypeSymbol` propagation should remove this name-based test rather than expand it.

Issue #2937, opened 2026-07-31
gsc: mismatched inline-lambda arity through erased imported-generic slot compiles then throws


DavidObando:
## Repro

```gsharp
package LambdaArityErasure
import System
import System.Collections.Generic

class Src {
    let N int32
    init(n int32) { N = n }
}

func Main() {
    let callbacks = List[System.Action[Src]]()
    callbacks.Add((left Src, right Src) -> Console.WriteLine(left.N + right.N))
}
```

## Actual

Compilation succeeds. Execution throws:

```
System.ArgumentException: The value "System.Action`2[Src,Src]" is not of type "System.Action`1[Src]"
```

This reproduces on `d4038f1d` and PR #2929. PR #2929 recovers symbolic target typing for matching-arity lambdas, but mismatched arity still falls back to the unchecked erased `object` path.

## Expected

Compiler rejects the call with an arity/delegate-conversion diagnostic before emit.

Issue #2936, opened 2026-07-31
gsc: invoking a function returned through a source-generic receiver reports GS0159


DavidObando:
## Repro

```gsharp
package SourceGenericFunctionReceiver
import System

class Src {
    let N int32
    init(n int32) { N = n }
}

class Box[T any] {
    let Value T
    init(value T) { Value = value }
    func Get(index int32) T -> Value
}

func Main() {
    let box = Box[(Src) -> int32]((item Src) -> item.N)
    Console.WriteLine(box.Get(0)(Src(5)))
}
```

## Actual

Both `d4038f1d` and PR #2929 fail with:

```
error GS0159: Cannot find function .
```

## Expected

`box.Get(0)` retains `(Src) -> int32`, invocation succeeds, and program prints `5`.

## Scope

#2918 fixes inline lambdas through **imported** generic receivers. Source-declared generic receivers still lose the callable result shape; track this separately so #2918's root-cause wording is not read as covering both paths.

DavidObando:
Resolved on `main` at `b78bb1f157621ab8d6d3ff25d21520bea6f6abae` by #2932 / #2944.

Verified with:
```text
dotnet build src/Compiler/Compiler.csproj -c Release --no-restore -p:TreatWarningsAsErrors=true
out/bin/Release/Compiler/gsc /out:out/step-zero/repro.dll /target:exe /targetframework:net10.0 out/step-zero/repro.gs
dotnet out/step-zero/repro.dll
```

Compiler exit: `0`. Exact stdout: `5`.

`box.Get(0)(Src(5))` is a receiver-wide indirect invocation of a member-call result. #2932 changed postfix parsing so the full `box.Get(0)` expression becomes the callee; that curried-invocation fix also covers source-generic receivers. A test-only PR will pin this exact shape.

PR #2973, opened 2026-08-01
Pin source-generic returned function calls

DavidObando:
## Result

Issue #2936 no longer reproduces. `main` at `b78bb1f1` compiles the exact report and `dotnet` prints `5`. #2932's receiver-wide indirect-invocation parse shape already covers `box.Get(0)(Src(5))`, so #2936 was closed with execution evidence.

## Change

Test-only regression pin for the exact source-generic receiver shape. No production files changed.

The test:

- compiles the reported program as an executable
- runs `IlVerifier.Verify`
- loads emitted bytes with `Assembly.Load` and calls `GetTypes()`
- executes through a bounded child `dotnet` process
- starts asynchronous reads of both redirected pipes before waiting
- kills the full process tree on timeout
- requires exact stdout `5\n`

## Known-bad control

The same test was built and executed against `d4038f1d`. It fails with exit 1 and the expected `GS0159: Cannot find function .`. It passes on this branch and current `main`.

## Root cause / scope

#2932 changed postfix parsing so the complete member-call result, `box.Get(0)`, becomes the callee of the following invocation. Generic substitution was already correct; no source-generic special case is needed.

This does **not** cover #2937 or #2948. Those concern imported erased-slot lambda target typing: mismatched arity and nominal imported delegate identity, respectively. This PR touches neither `ExpressionBinder.Calls.cs` nor any production file.

## Pins, guards, mutations

- Against current base `8dddbbf2`: 0 pins, 1 guard; behavior is already fixed on main.
- Against known-bad `d4038f1d`: 1/1 historical regression pin fails for the exact GS0159 and passes here.
- Mutation testing: not applicable; production delta is zero and there are no semantic production branches to mutate.

## Validation

Same-base executed counts:

| Suite | `8dddbbf2` base | `d88b4d40` head |
|---|---:|---:|
| Core.Tests | 7028 passed, 1 skipped | 7028 passed, 1 skipped |
| Compiler.Tests | 3893 passed | 3894 passed |
| Interpreter.Tests | 489 passed | 489 passed |

Delta is exactly this PR's one Compiler regression test.

Release warnings-as-errors build: 0 warnings, 0 errors.

## Corpus and interpreter parity

All 148 `samples/**` + `e2etests/**` files were compiled with separately preserved base and branch compilers:

- exit status: 148/148 identical
- diagnostics/output: 148/148 byte-identical
- emitted assemblies: 139/139 SHA-256 identical
- corpus files calling a call result through a generic receiver: **0**

Hand-written exact-shape probe: compiled output `5`; `gsi` output `5`.

### DIFF
--- a/test/Compiler.Tests/Emit/Issue2936SourceGenericReceiverCallTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2936SourceGenericReceiverCallTests.cs
@@ -1,0 +1,126 @@
+// <copyright file="Issue2936SourceGenericReceiverCallTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+using Xunit;
+
+namespace GSharp.Compiler.Tests.Emit;
+
+/// <summary>
+/// Issue #2936: calls through source-generic member results retain their
+/// substituted function type.
+/// </summary>
+public class Issue2936SourceGenericReceiverCallTests
+{
+    [Fact]
+    public void SourceGenericReceiverReturnedFunction_LoadsAndRuns()
+    {
+        const string Source = """
+            package SourceGenericFunctionReceiver
+            import System
+
+            class Src { let N int32
+                        init(n int32) { N = n } }
+
+            class Box[T any] { let Value T
+                               init(value T) { Value = value }
+                               func Get(index int32) T -> Value }
+
+            func Main() {
+                let box = Box[(Src) -> int32]((item Src) -> item.N)
+                Console.WriteLine(box.Get(0)(Src(5)))
+            }
+            """;
+
+        var directory = Directory.CreateTempSubdirectory("gsharp_issue2936_").FullName;
+        try
+        {
+            var assemblyPath = Compile(Source, directory);
+            IlVerifier.Verify(assemblyPath);
+            var assembly = Assembly.Load(File.ReadAllBytes(assemblyPath));
+            Assert.NotEmpty(assembly.GetTypes());
+            Assert.Equal("5\n", RunBounded(assemblyPath));
+        }
+        finally
+        {
+            try
+            {
+                Directory.Delete(directory, recursive: true);
+            }
+            catch
+            {
+            }
+        }
+    }
+
+    private static string Compile(string source, string directory)
+    {
+        var sourcePath = Path.Combine(directory, "Program.gs");
+        var assemblyPath = Path.Combine(directory, "Issue2936.dll");
+        File.WriteAllText(sourcePath, source);
+
+        using var stdout = new StringWriter();
+        using var stderr = new StringWriter();
+        var previousOut = Console.Out;
+        var previousError = Console.Error;
+        Console.SetOut(stdout);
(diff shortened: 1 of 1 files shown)


DavidObando:
## Rubber-duck verdict: MERGEABLE — head `d88b4d40`

The pin is real. It fails pre-fix for the issue's own reason and passes post-fix.

### Decisive check

Pre-fix commit `17789519` — the **first parent** of `0225df04` ("Fix tuple-element delegate calls (#2932)"), i.e. the merge's base, not a squash artifact. Clone at `17789519`, copied **only** the new test file (`git status` showed exactly one `??` entry), built, ran the filter:

```
Failed SourceGenericReceiverReturnedFunction_LoadsAndRuns [342 ms]
  …/Program.gs(13,13,13,42): error GS0159: Cannot find function WriteLine.
  …/Program.gs(13,27,13,41): error GS0159: Cannot find function .
Failed! - Failed: 1, Passed: 0, Total: 1
```

`error GS0159: Cannot find function .` is **verbatim the issue's Actual**. Not a syntax error, not a missing symbol, not unrelated drift — the file also compiles cleanly at `17789519`, so the failure is semantic. At head `d88b4d40`: `Passed! - Failed: 0, Passed: 1, Total: 1`.

Direct compiler confirmation of the same repro:

| | exit | output |
| --- | --- | --- |
| `gsc` @ `17789519` | **1** | `error GS0159: Cannot find function .` |
| `gsc` @ `8dddbbf2` | **0** | runs, prints **`5`** — the issue's Expected |

### Secondary checks

- **Covers the actual repro exactly.** The `Source` string is the issue's program with whitespace condensed only: same `Src`/`Box[T any]`, same `Box[(Src) -> int32]((item Src) -> item.N)`, same `box.Get(0)(Src(5))`, same expected `5`. No simplification of the reported shape.
- **Count reconciles.** 1 declared case; `main` `8dddbbf2` Compiler 3893 + 1 = 3894. Targeted filter only (1/1) per load constraints, so treat the total as arithmetic rather than measured.
- **No duplicate coverage.** `0225df04`'s tests are all `Issue2924TupleElementDelegateCallTests` — *tuple*-element delegate calls. No existing test matches `2936`, `SourceGenericFunctionReceiver`, or a `Box[(…) -> …]` shape. Source-generic receiver returning a function type was an unpinned path that #2932 fixed incidentally — precisely the case where a pin earns its keep.
- **Mutation testing skipped** — no production change, so nothing to mutate.

### Note, non-blocking

`IlVerifier.Verify` is called at `:43`, but **detection does not depend on it**: `Compile()` asserts `exitCode == 0` at `:90` and embeds gsc's stdout in the failure message, so the pre-fix regression trips the compile assert before `IlVerifier` runs. Real verification is `Assembly.Load` + `GetTypes()` + child-process execution (`RunBounded`, `:93-125`) asserting stdout `"5\n"`.

`RunBounded` uses the canonical pattern correctly — both pipes `ReadToEndAsync()` started **before** `WaitForExit(30_000)`, bounded wait, `Kill(entireProcessTree: true)` on timeout. Given `IlVerifier` has passed broken assemblies five times in this series, this is the right shape.

All scratch trees created with `git clone --shared --no-checkout`; `dotnet build -m:1` throughout; no `GSharp.Core.dll` copy was zero-byte.

Issue #1537, opened 2026-07-01
Generic type nested in a generic type can't be constructed/resolved externally (Outer[U].Middle[T] → GS0159)


DavidObando:
## Summary
Member access on a **generic type nested inside another generic type**, constructed via the per-segment type-clause syntax (`Outer[int32].Middle[string]`), fails during **binding** with `error GS0159: Cannot find function …`. The construction's members cannot be resolved, so the program does not compile.

This is the follow-up gap after #1521. #1521 fixed **emit** for a **non-generic** type nested in a generic (`Box[T].Tag`). This issue is the separate, pre-existing gap for a **generic** type nested in a generic (`Outer[U].Middle[T]`) — where the nested type declares its **own** type parameters.

## Minimal repro (staged `gsc`) — FAILS
```gsharp
package Audit.C1
import System
struct Outer[U any] {
    struct Middle[T any] {
        var Label string
        func Hello() string { return "hi" }
    }
}
func main() {
    var m = Outer[int32].Middle[string]{Label: "x"}
    Console.WriteLine(m.Hello())
}
```
```
error GS0159: Cannot find function Hello.
        Console.WriteLine(m.Hello())
error GS0159: Cannot find function WriteLine.   # cascade from the error-typed receiver
```
The member `Hello` (and any member) cannot be resolved on the externally-constructed `Outer[int32].Middle[string]` value.

## Controls that WORK (scope the gap)
**A — NON-generic outer, generic nested type** → compiles, `ilverify` clean, runs:
```gsharp
struct Outer {
    struct Middle[T any] { var Label string  func Hello() string { return "hi" } }
}
func main() { var m = Outer.Middle[string]{Label: "x"}  Console.WriteLine(m.Hello()) }
```
**B — generic outer + generic nested, INTERNAL use only** (constructed from within `Outer`'s own member) → compiles, `ilverify` clean, runs:
```gsharp
struct Outer[U any] {
    struct Middle[T any] { var Label string  func Hello() string { return "hi" } }
    func Make() string { var m = Middle[string]{Label: "y"}  return m.Hello() }
}
func main() { var o = Outer[int32]{}  Console.WriteLine(o.Make()) }
```
So the gap is specifically: **a generic nested type of a generic type, constructed from OUTSIDE via `Outer[U].Middle[T]` per-segment syntax** — the binder cannot resolve members on the result. (Triple nesting such as `Outer[U].Middle[T].Inner` is affected for the same underlying reason.)

[...]

Issue #2975, opened 2026-08-01
select-arm bindings are not treated as binding sites by state-machine hoisting analysis


DavidObando:
## Summary

`Lowering/Iterators/HoistedFieldRewriter.cs` hoists locals that are live across a `yield` into state-machine fields. It handles variable expressions (`RewriteVariableExpression`, ~`:50`), variable declarations (`RewriteVariableDeclaration`, ~`:114`) and receiver proxies (~`:76`), but has **no `BoundSelectCase` handling at all**.

`BoundSelectCase.Variable` is a binding site. In a sync iterator with a `select` whose received arm value is live across a `yield`, that variable is never added to `fieldMap` and is therefore never hoisted.

## Contrast with the async path

PR #2967 (issue #2933) is fixing the async side, where `MethodBodyEmitter.Statements.cs:647` guards arm-value stores on `asyncFieldMap != null`. Async iterators hoist into `iteratorEmitCtx.FieldMap`, so that guard is separately wrong. The sync-iterator gap described here is a **different, deeper defect**: the value is not hoisted in the first place, so there is no field for the emitter to store into regardless of which guard runs.

## Expected

`HoistedFieldRewriter` should treat `BoundSelectCase.Variable` as a binding site and hoist it when it is live across a `yield`, consistent with how it treats ordinary variable declarations.

## Notes

Found by inspection while reviewing PR #2967. `grep BoundSelectCase src/Core/CodeAnalysis/Lowering/Iterators/` returns nothing.

DavidObando:
## Additional evidence: this defect is not async-specific, and it is live on `main`

Found while reviewing PR #2967 (#2933, async select arm binding). Two independent probes confirm `BoundSelectCase.Variable` is not treated as a binding site by state-machine hoisting analysis, in **both** the sync and async iterator paths.

### Sync iterator — binding live across a `yield`

A sync iterator whose select-arm binding is live across a `yield` prints `1,7` where `gsi` prints `1,10`. Reproduced on base `17789519`, on PR #2967's head `d49495a9`, **and with #2967's blocker fix applied** — so it is orthogonal to that PR and is not fixed by it.

Instrumenting the emitter printed `inIterMap=False`: the variable is **never hoisted at all**, so the local resets on `MoveNext` re-entry. That places the defect in hoisting analysis, not in the emitter.

### Async iterator — the same root cause with a second, separate symptom

[...]

Issue #2970, opened 2026-08-01
Keep canonical and website diagnostic references synchronized


DavidObando:
## Summary

`docs/diagnostics.md` and `website/docs/ref/diagnostics.md` can drift independently. Issue #2959 exposed this: on `8dddbbf2`, the canonical reference was missing 113 descriptor IDs and the website reference was missing 75. Of the current descriptor IDs, 59 appeared only on the website, 21 only in the canonical reference, and 54 in neither.

After #2959 fills the canonical gaps, the website reference will still lack 75 current descriptors. A new diagnostic can therefore be documented in one reference while remaining absent from the other.

## Repro

Extract `GS####` IDs from `DiagnosticDescriptors.cs` and table rows beginning with `| GS#### |` in both references, then compare the sets. The two documentation sets differ on `8dddbbf2`.

## Expected

Add a test or a single-source generation step that keeps the current diagnostic IDs and severities synchronized between both references. Deliberately historical or retired IDs should be handled explicitly so a brand-new omission still fails.

DavidObando:
Reopening: #2981 completed canonical coverage but did not synchronize the website reference. Follow-up takes option (a): fill every current website gap, align severities, and add one bidirectional guard with explicit entries for direct-emitted, retired, and reserved IDs.

PR #2978, opened 2026-08-01
Fix nested generic iterator and method metadata

### LINKED ISSUE #2951: Nested generic struct iterators emit invalid state-machine types
## Repro 1: nested struct uses enclosing type parameter

```gs
package NestedGenericIterator
import System

class Wrap[T] {
    struct Cell {
        var A T
        func vals() sequence[T] { yield A }
    }
}

func Main() {
    for value in Wrap[int32].Cell{A: 42}.vals() {
        Console.WriteLine(value)
    }
}
```

Actual: `TypeLoadException: Type <vals>d__1 … has a field of an illegal type`.

## Repro 2: nested generic struct

```gs
package NestedGenericIterator2
import System

class Wrap[T] {
    struct Cell[U] {
        var A U
        func vals() sequence[U] { yield A }
    }
}

func Main() {
    for value in Wrap[int32].Cell[string]{A: "x"}.vals() {
        Console.WriteLine(value)
    }
}
```

Actual: `MissingMethodException` while loading/invoking the generated iterator.

## Expected

Both programs load and print `42` / `x`.

## Scope

Pre-existing at the exact #2931 base and unchanged by PR #2950. Top-level generic struct iterators such as `GenericBox[T]` work. This nested-type failure is unrelated to #2931's missing `ldobj` when capturing a struct receiver; it concerns nested generic state-machine type/member reification.

### LINKED ISSUE #2977: Chained access after nested generic struct literal fails to bind
## Summary

A nested generic struct literal constructed through per-segment syntax cannot be used directly as a postfix/member-access receiver. Moving the identical literal into a local variable makes the program bind.

This is independent of #2951's iterator metadata defect: the local-variable control reaches emit and is fixed by PR #2978, while the chained spelling fails in binding before emit.

## Repro

```gsharp
package NestedGenericLiteralChain
import System

class Wrap[T] {
    struct Cell[U] {
        var A U
        func vals() sequence[U] { yield A }
    }
}

func Main() {
    for value in Wrap[int32].Cell[string]{A: "x"}.vals() {
        Console.WriteLine(value)
    }
}
```

Actual diagnostics:

```text
error GS0158: Cannot find member Cell.
error GS0157: Cannot find type . Are you missing an import?
error GS0005: Unexpected token <DotToken>, expected <IdentifierToken>.
error GS0159: Cannot find function WriteLine.
error GS0125: Variable 'value' doesn't exist.
```

`gsc` exits 1 and emits no assembly.

## Working control

```gsharp
func Main() {
    var cell = Wrap[int32].Cell[string]{A: "x"}
    for value in cell.vals() {
        Console.WriteLine(value)
    }
}
```

The control binds. After #2951's metadata fix it loads and prints `x`.

## Version evidence

The failing diagnostics and exit code are identical on:

- `main` `8dddbbf2`
- PR #2978 head `e19aa42b7`

This appears to be a postfix-expression/parser-or-binder residual after #1537, specifically when a nested generic type literal is immediately followed by `.member(...)`.

## Acceptance criteria

- The direct chained repro compiles and prints `x`.
- The local-variable control remains valid.
- Cover nested generic class/struct literals and at least two nesting levels.

DavidObando:
## Summary

- reify sync and async iterator state machines over the full CLR type-parameter scope: flattened enclosing types, the receiver's own parameters, then method parameters
- remap every source type parameter to its state-machine `VAR` slot
- encode nested-generic method `MemberRef` signatures under the declaring type's registered re-ordinalization, including ordinary non-iterator methods
- add load-and-execute coverage for asymmetric nesting, nested structs/classes, ordinary methods, two constructions, async/shared iterators, nullable specialization, and direct `GetEnumerator` use

Fixes #2951

## Root cause

The repros expose two metadata mismatches:

1. `StateMachineEmitter.cs` included only the receiver's own parameters plus method parameters. It omitted flattened enclosing parameters, and its emit-time remap covered only method parameters. A nested receiver such as `Wrap<T>.Cell` therefore left `T` open in state-machine fields; `Wrap<T>.Cell<U>` also assigned the wrong CLR `VAR` ordinal to `U`.
2. `UserTokenResolver.cs` encoded a nested-generic method's call-site `MemberRef` signature without the declaring type's registered re-ordinalization. The MethodDef encoded `U` as `VAR(1)` in `Wrap<T>.Cell<U>`, while the MemberRef encoded it as `VAR(0)`, producing `MissingMethodException`.

The second fix is general metadata plumbing, not iterator-specific: ordinary nested-generic methods used the same broken MemberRef path. The added `echo(U, T)` test pins that residual nested-generic behavior related to #1537.

Sync and async iterators shared the state-machine defect. Nested classes shared it. Non-generic nesting levels contribute zero parameters; asymmetric nesting tests now pin that their presence does not shift later ordinals.

**This PR does not fix #2977.** The ticket's direct chained literal spelling binder-fails before emit, identically on base and this head. The equivalent local-variable form reproduces the documented metadata failure. #2977 tracks that independent binder residual.

## Verification

Known-bad compiler: `8dddbbf2`.

- **Pins:** 14 compiler tests fail on the known-bad compiler and pass here.
- **Guards:** 4 compiler tests and 2 interpreter parity tests pass on both.
- All 18 compiler tests use `Assembly.Load(File.ReadAllBytes(...))`, `GetTypes()`, and bounded child-process execution with exact stdout.
- `gsi` vs compiled parity: 3/3 probes byte-equal stdout (`42`, `x\nx`, `42x\ny7`).
- Final targeted controls: Issue2951 **18/18**, overlapping Issue2927 **20/20**, interpreter parity **2/2**.
- Serial Release solution build: **0 warnings, 0 errors**.
- All four Compiler/Compiler.Tests/Core.Tests/Interpreter.Tests `GSharp.Core.dll` copies matched and were non-zero after the final build.

### Semantic mutations

Unmutated controls: Issue2951 **18/18 passed**; the no-owner-path probe used the broader Issue2927 **20/20 passed** control.

| Mutation | Result |
|---|---:|
| omit enclosing parameters | 12 failed / 6 passed |
| omit receiver-own parameters | 11 failed / 7 passed |
| omit method parameters | 1 failed / 17 passed |
| ignore shared/static owner | 1 failed / 17 passed |
| omit top-level method parameters on the no-owner path | 9 failed / 11 passed |
| bypass sync scope helper | 13 failed / 5 passed |
| retain old sync remap source | 8 failed / 10 passed |
| bypass async scope helper | 2 failed / 16 passed |
| retain old async remap source | 2 failed / 16 passed |
| reverse sync `VAR` slots | 8 failed / 10 passed |
| reverse async `VAR` slots | 2 failed / 16 passed |
| omit method-signature remap | 3 failed / 15 passed |

Every accepted mutant built successfully via `dotnet build GSharp.sln -c Release -m:1`. The Compiler, Compiler.Tests, Core.Tests, and Interpreter.Tests copies of `GSharp.Core.dll` had matching, non-zero MD5s that differed from pristine; every restore rebuild returned all four to the pristine MD5 and the control behavior passed again.

### Suites

Base `a6b1561f2` → head `6b76a4281`, both from successful CI runs:

- Core: 7028 passed + 1 skipped → 7028 passed + 1 skipped
- Compiler: 3906 → 3924
- Interpreter: 491 → 493
- LanguageServer: 452 → 452

### Corpus

Current base `a6b1561f2` vs head: 148/148 samples, `BYTEDIFF=0`, `DIAGDIFF=0`, `RCDIFF=0` using SHA-256 per emitted file.

This is weak evidence for the `UserTokenResolver` hunk: **0/148 samples declare any nested type**, so the registered remap path is inert across the corpus. The meaningful non-regression evidence for that path is the reviewer's nested-generic, type-parameter-free method probe, which emitted byte-identically on base and head (MD5 `8e2b4e5983700b5f4484d54f81becc7b`).

## Collision notes

- `StateMachineEmitter.cs`: touched; narrow iterator generic-scope change only.
- `UserTokenResolver.cs`: touched; nested-generic method-signature remap.
- `StructSymbol.cs`: **not touched**.

### DIFF
--- a/src/Core/CodeAnalysis/Emit/StateMachineEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/StateMachineEmitter.cs
@@ -263,12 +263,12 @@ internal sealed class StateMachineEmitter
         public IteratorStateMachineInfo(
             IteratorStateMachinePlan plan,
             StructSymbol classSym,
-            ImmutableArray<TypeParameterSymbol> outerMethodTypeParameters,
+            ImmutableArray<TypeParameterSymbol> sourceTypeParameters,
             ImmutableArray<TypeParameterSymbol> classTypeParameters)
         {
             this.Plan = plan;
             this.ClassSym = classSym;
-            this.OuterMethodTypeParameters = outerMethodTypeParameters;
+            this.SourceTypeParameters = sourceTypeParameters;
             this.ClassTypeParameters = classTypeParameters;
         }
 
@@ -277,44 +277,38 @@ internal sealed class StateMachineEmitter
         public StructSymbol ClassSym { get; }
 
         /// <summary>
-        /// Gets the OUTER method's type parameters (those declared on
-        /// <see cref="IteratorStateMachinePlan.Function"/>) that the
-        /// state-machine class is reified over (issue #810). Empty for
-        /// non-generic iterators.
+        /// Gets the original receiver and method type parameters that the
+        /// state-machine class is reified over, in state-machine slot order.
         /// </summary>
-        public ImmutableArray<TypeParameterSymbol> OuterMethodTypeParameters { get; }
+        public ImmutableArray<TypeParameterSymbol> SourceTypeParameters { get; }
 
         /// <summary>
         /// Gets the state-machine class's own type parameters (issue #810),
         /// constructed as ordinal-aligned mirrors of
-        /// <see cref="OuterMethodTypeParameters"/> with
+        /// <see cref="SourceTypeParameters"/> with
         /// <see cref="TypeParameterSymbol.IsMethodTypeParameter"/> set to
         /// <see langword="false"/>. Empty for non-generic iterators.
         /// </summary>
         public ImmutableArray<TypeParameterSymbol> ClassTypeParameters { get; }
 
         /// <summary>
-        /// Issue #810 + #1465: returns the emit-time remap from each
-        /// outer-method type parameter to its corresponding class-type-parameter
-        /// ordinal on the synthesized state machine, or <see langword="null"/>
-        /// when this iterator method has no method-level type parameters. The
-        /// state machine's leading type parameters mirror the enclosing generic
-        /// type's parameters, so a method type parameter maps to the slot
-        /// offset by the enclosing class type-parameter count. Used by
+        /// Issue #810 + #1465 + #2951: returns the emit-time remap from each
+        /// original receiver/method type parameter to its corresponding
+        /// class-type-parameter ordinal on the synthesized state machine, or
+        /// <see langword="null"/> when none are in scope. Used by
         /// <see cref="GenericRemapState.ActiveIteratorStateMachineRemap"/>.
         /// </summary>
         public Dictionary<TypeParameterSymbol, int> BuildRemap()
         {
-            if (this.OuterMethodTypeParameters.IsDefaultOrEmpty)
+            if (this.SourceTypeParameters.IsDefaultOrEmpty)
             {
                 return null;
             }
 
-            var offset = this.ClassTypeParameters.Length - this.OuterMethodTypeParameters.Length;
-            var map = new Dictionary<TypeParameterSymbol, int>(this.OuterMethodTypeParameters.Length);
-            for (var i = 0; i < this.OuterMethodTypeParameters.Length; i++)
+            var map = new Dictionary<TypeParameterSymbol, int>(this.SourceTypeParameters.Length);
+            for (var i = 0; i < this.SourceTypeParameters.Length; i++)
             {
-                map[this.OuterMethodTypeParameters[i]] = offset + i;
+                map[this.SourceTypeParameters[i]] = i;
             }
 
             return map;
@@ -323,46 +317,42 @@ internal sealed class StateMachineEmitter
 
     /// <summary>
     /// Issue #1489: minimal carrier for an async-iterator SM's generic
-    /// reification — the outer (kickoff) method's type parameters and the
+    /// reification — the source function's in-scope type parameters and the
     /// state-machine class's own ordinal-aligned type parameters. Used to
-    /// build the outer-method-TP → class-TP-ordinal remap (mirroring
+    /// build the source-TP → class-TP-ordinal remap (mirroring
     /// <see cref="IteratorStateMachineInfo.BuildRemap"/>) without a sync
     /// iterator plan.
     /// </summary>
     public sealed class IteratorStateMachineGenericInfo
     {
         public IteratorStateMachineGenericInfo(
-            ImmutableArray<TypeParameterSymbol> outerMethodTypeParameters,
+            ImmutableArray<TypeParameterSymbol> sourceTypeParameters,
             ImmutableArray<TypeParameterSymbol> classTypeParameters)
         {
-            this.OuterMethodTypeParameters = outerMethodTypeParameters;
+            this.SourceTypeParameters = sourceTypeParameters;
             this.ClassTypeParameters = classTypeParameters;
         }
 
-        public ImmutableArray<TypeParameterSymbol> OuterMethodTypeParameters { get; }
+        public ImmutableArray<TypeParameterSymbol> SourceTypeParameters { get; }
 
         public ImmutableArray<TypeParameterSymbol> ClassTypeParameters { get; }
 
         /// <summary>
-        /// Returns the emit-time remap from each outer-method type parameter to
-        /// its class-type-parameter ordinal on the synthesized state machine,
-        /// or <see langword="null"/> when the method has no type parameters.
-        /// The leading class type parameters mirror the enclosing generic
-        /// type's parameters, so a method type parameter maps to the slot
-        /// offset by the enclosing class type-parameter count.
+        /// Returns the emit-time remap from each original receiver/method type
+        /// parameter to its class-type-parameter ordinal on the synthesized
+        /// state machine, or <see langword="null"/> when none are in scope.
         /// </summary>
         public Dictionary<TypeParameterSymbol, int> BuildRemap()
         {
-            if (this.OuterMethodTypeParameters.IsDefaultOrEmpty)
+            if (this.SourceTypeParameters.IsDefaultOrEmpty)
             {
                 return null;
             }
 
-            var offset = this.ClassTypeParameters.Length - this.OuterMethodTypeParameters.Length;
-            var map = new Dictionary<TypeParameterSymbol, int>(this.OuterMethodTypeParameters.Length);
-            for (var i = 0; i < this.OuterMethodTypeParameters.Length; i++)
+            var map = new Dictionary<TypeParameterSymbol, int>(this.SourceTypeParameters.Length);
+            for (var i = 0; i < this.SourceTypeParameters.Length; i++)
             {
-                map[this.OuterMethodTypeParameters[i]] = offset + i;
+                map[this.SourceTypeParameters[i]] = i;
             }
 
             return map;
@@ -435,6 +425,29 @@ internal sealed class StateMachineEmitter
         public StandaloneSignatureHandle LocalsSignature { get; }
     }
 
+    private static ImmutableArray<TypeParameterSymbol> GetIteratorTypeParametersInScope(FunctionSymbol function)
+    {
+        var builder = ImmutableArray.CreateBuilder<TypeParameterSymbol>();
+        var containingType = function.ReceiverType as StructSymbol
+            ?? function.StaticOwnerType as StructSymbol;
+        if (containingType != null)
+        {
+            var definition = containingType.Definition ?? containingType;
+            builder.AddRange(StructSymbol.CollectEnclosingTypeParameters(definition));
+            if (!definition.TypeParameters.IsDefaultOrEmpty)
+            {
+                builder.AddRange(definition.TypeParameters);
+            }
+        }
+
+        if (!function.TypeParameters.IsDefaultOrEmpty)
+        {
+            builder.AddRange(function.TypeParameters);
+        }
+
+        return builder.ToImmutable();
+    }
+
     #region Iterator state-machine synthesis
 
     public void SynthesizeIteratorStateMachines(PackageSymbol hostPackage)
@@ -465,22 +478,11 @@ internal sealed class StateMachineEmitter
             // activeIteratorStateMachineRemap encode-time hook. The class TPs
             // themselves are not used directly in any bound expression — the
             // substitution lives at the encoder.
-            var enclosingClassTPs = plan.Function.ReceiverType is StructSymbol iterRecv
-                ? (iterRecv.Definition ?? iterRecv).TypeParameters
-                : ImmutableArray<TypeParameterSymbol>.Empty;
-            if (enclosingClassTPs.IsDefault)
-            {
-                enclosingClassTPs = ImmutableArray<TypeParameterSymbol>.Empty;
-            }
-
-            var outerMethodTPs = plan.Function.TypeParameters.IsDefaultOrEmpty
-                ? ImmutableArray<TypeParameterSymbol>.Empty
-                : plan.Function.TypeParameters;
-
-            // Combined in-scope type parameters: enclosing class first, then
-            // method. Drives the SM's own generic parameters and the
+            // Combined in-scope type parameters: flattened enclosing receiver
+            // types first, then the receiver's own parameters, then method
+            // parameters. Drives the SM's own generic parameters and the
             // self-instantiation type arguments used by the kickoff.
-            var scopeTPs = enclosingClassTPs.AddRange(outerMethodTPs);
+            var scopeTPs = GetIteratorTypeParametersInScope(plan.Function);
             ImmutableArray<TypeParameterSymbol> classTPs;
             if (scopeTPs.IsDefaultOrEmpty)
             {
@@ -638,8 +640,8 @@ internal sealed class StateMachineEmitter
             // CreateIteratorStateMachineLiteral helper and let the encoder
             // routing decide. The kickoff body lives inside the user's
             // generic method whose type parameters are still active
-            // (`MVar(0)`), so we explicitly construct the SM type over the
-            // outer method's TPs.
+            // (`MVar(0)`), so we explicitly construct the SM type over every
+            // type parameter in scope.
             var kickoffSmType = classTPs.IsDefaultOrEmpty
                 ? smClass
                 // Issue #2037: project imported constructed-generic hoisted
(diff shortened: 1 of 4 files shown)


DavidObando:
# Review: PR #2978 — nested generic iterator metadata

**Head reviewed:** `e19aa42b7dcf479d8478a0a143fe6f79c28978f6`
**Base (merge-base):** `8dddbbf285c2b4e94d06647671b44f9b637537f7` · current `main` `b1e693bb1` (1 commit behind, `merge-tree` clean)

## Verdict: **MERGEABLE**

The ordinal arithmetic is correct — verified by construction on four asymmetric-nesting shapes the test suite does not contain, all of which are pins. Two coverage gaps below, neither of which makes the change unsafe to land.

---

## Method

Scratch clones at `/tmp/d2951/{base,head}` (`git clone --shared`, `git describe` → `v0.3.362-109-ge19aa42b7`). Built `src/Compiler/Compiler.csproj -c Release -m:1` only — no solution build, no test-project build, nothing written to `~/.nuget/packages/gsharp.net.sdk/`, no e2e. `GSharp.Core.dll` md5 content-checked before/after every mutation (base `7ee40ba3…`, head `3f4d3292…`, never the zero-byte `d41d8cd9…`). Every probe is `gsc` compile → child-process `dotnet exec` with a bounded 60s timeout and exact stdout comparison. Both worktrees left with `git status --porcelain` empty.

---

## Q1 — ordinal arithmetic, verified by construction

`GetIteratorTypeParametersInScope` (`StateMachineEmitter.cs:428`) builds `CollectEnclosingTypeParameters(def)` ++ `def.TypeParameters` ++ `function.TypeParameters`. Reading the two supporting helpers:

- `StructSymbol.cs:1447` `CollectEnclosingTypeParameters` starts at `EnclosingTypeOf(nested)`, so it **excludes** the type's own parameters — no double-add against the following `AddRange(definition.TypeParameters)`. It prepends per level (`levels.Insert(0, …)`), so the result is outermost-first.
- `StructSymbol.cs:1757` `EnclosingTypeParametersOf` returns each level's **source-declared** `TypeParameters`, not a CLR-flattened list. A non-generic level therefore contributes zero and is skipped by `if (!tps.IsDefaultOrEmpty)` — no ordinal shift.

Identity remap in both `BuildRemap` bodies is then correct by construction, because `classTPs` is built ordinal-aligned from the same `scopeTPs` array.

Verified against the two asymmetric shapes the brief asked for, plus two more. Every probe uses **distinct types per level**, so an ordinal swap is observable, not merely a crash:

| probe | shape | base `8dddbbf2` | head `e19aa42b` |
|---|---|---|---|
| `p1` | `class Outer { struct Mid[U] { struct Inner } }` | `BadImageFormatException` at `Outer.Mid`1.Inner`1.vals()` | `q` |
| `p2` | `class Outer[T] { class Mid { struct Inner[V] } }` | `BadImageFormatException` at `Outer`1.Mid`1.Inner`2.vals()` | `7\|z` |
| `p3` | same, `struct Mid`, `Outer[bool]…Inner[int32]` | `BadImageFormatException` | `True\|5` |
| `p4` | `class Outer { class Mid[U] { struct Inner[V] } }` | `BadImageFormatException` | `k\|9` |
| `a1` | `p2` shape with `async func vals() async sequence[string]` | `BadImageFormatException` | `3\|w` |

No off-by-one. `p2`/`p3` are the mirror pair (generic outer, **non-generic middle**, generic inner) and both place `T` at `VAR(0)` and `V` at `VAR(1)` correctly.

**All five are pins with zero test coverage** — see SHOULD-FIX 1.

## Q2 — the `-68` deletions

Enumerated in full: two copies of the same ~13-line inline `enclosingClassTPs`/`outerMethodTPs` block (`:478`, `:730`), two copies of the `offset = ClassTypeParameters.Length - OuterMethodTypeParameters.Length` arithmetic, one property rename (`OuterMethodTypeParameters` → `SourceTypeParameters`), and doc comments. **Nothing guard-like was removed**, and no test was deleted or modified — the diff is 4 files, 2 of them new test files.

The new scope is a strict superset of the old: old took `ReceiverType is StructSymbol` → own `TypeParameters`; new takes `ReceiverType as StructSymbol ?? StaticOwnerType as StructSymbol` → enclosing ++ own ++ method. The `StaticOwnerType` fallback is new behaviour, not a preserved one: a `shared` iterator on a generic type previously got an **entirely empty** scope.

Regression coverage for the previously-handled shapes is present and adequate: `TopLevelGenericStructIteratorRemainsValid`, `NonGenericNestedStructIteratorRemainsValid`, `TopLevelFunctionIteratorRemainsValid`, `NestedEnumIteratorRemainsValid`. Independently, `p6` (top-level `Box[T]` with instance + static calls) runs identically on base and head, and `p5` (nested generic, non-iterator, TP-free signature) emits a **byte-identical** assembly on both (md5 `8e2b4e5983700b5f4484d54f81becc7b`).

### The async asymmetry — answered

Mutation **MA**: `IteratorStateMachineGenericInfo(scopeTPs, classTPs)` → `(ImmutableArray<TypeParameterSymbol>.Empty, classTPs)` at `StateMachineEmitter.cs:849`. Rebuilt (`GSharp.Core.dll` md5 `3f4d3292…` → `61dcf854…`), probed, restored, rebuilt (md5 back to `3f4d3292…`), re-probed — behaviour flipped in both directions.

Under MA the sync probes `p1`, `p2`, `p8`, `p10` stay green and the async probes **run to completion, exit 0, and print garbage**:

```
a2  expected 44a     got 44122652240
a1  expected 3|w     got 3|122439248
```

So: the async hunk is separately load-bearing, and `AsyncNestedGenericStructIteratorUsesBothParameters` — which asserts `"44a\n"` exactly — genuinely kills it. **The 1-kill counts are real, not stale-binary artefacts.** Async is covered by exactly one test, which is thin, but it is a real test and its exact-output assertion is what catches the failure mode. Note that the mutant loads cleanly and `GetTypes()` succeeds; only the string comparison detects it. That vindicates the harness design.

## Q3 — nothing deleted on survival grounds

Confirmed. Every deletion is de-duplication into `GetIteratorTypeParametersInScope`, replaced call-for-call. No guard, no branch, and no test removed.

## Q4 — `UserTokenResolver` blast radius

**The "applied only where registered" property is proven by construction, not by testing.** `GenericRemapState.cs:150-155`: `PushSmRemap` returns `new SmRemapScope(this, null, restore: false)` without touching `activeIteratorStateMachineRemap` unless `iteratorStateMachineRemapsByClass.TryGetValue(smClass, …)` succeeds. Where the declaring type is not registered, the three new wraps are literally inert.

The registered set is wider than "iterators", though — `RegisterClassRemap` is called from `ReflectionMetadataEmitter.cs:498` (async SM), `:572` (**every nested user type** reified over flattened enclosing+own, #1499/#1537), `:637` (synthesized closure/capture-box classes, #1477), `:1229`, `:1242`. So the wrap fires for ordinary nested generic user types.

That is not a defect — it is the more interesting half of this PR. Probe `p7`, a **plain non-iterator** method on a nested generic type whose signature uses both parameters:

```gsharp
class Outer[T] { struct Cell[U] { var A T  var B U
    func echo(u U, t T) string { return u.ToString() + "/" + t.ToString() } } }
var v = Outer[int32].Cell[string]{A: 1, B: "b"}
Console.WriteLine(v.echo("z", 9))
```

| | result |
|---|---|
| base `8dddbbf2` | `MissingMethodException: Method not found: 'System.String Cell\`2.echo(!0, !0)'` |
| head `e19aa42b` | `z/9` |

Both `!0` — the exact `VAR(1)`→`VAR(0)` collapse item 2 describes, **with no iterator anywhere in the program**.

Mutation **M-UTR** (revert `UserTokenResolver.cs` to its `8dddbbf2` content, keep `StateMachineEmitter.cs`; rebuild md5 `3063e5d9…`, restore rebuild md5 back to `3f4d3292…`) resolves which sites are covered at semantic-branch granularity:

| probe | shape | M-UTR |
|---|---|---|
| `p10` = `NestedGenericStructUsesOwnTypeParameterDirectlyAndInLoop` | instance site `:1455` | **KILLS** — `MissingMethodException: 'IEnumerable\`1<!0> Cell\`2.vals()'` |
| `p8` = `SharedNestedGenericStructIteratorUsesOwnTypeParameter` | static site `:1550` | **KILLS** — `MissingMethodException: 'IEnumerable\`1<!0> Cell\`2.vals(!0)'` |
| `p9` = `NestedIteratorComposesWithNullableGenericSpecialization` | | survives |
| `p1`–`p4`, `a1`, `a2`, `p5`, `p6` | | survive |
| `p7` non-iterator | | **KILLS** |

So 2 of the 12 compiler tests cover the `UserTokenResolver` hunk, both through iterator spellings. The non-iterator use — the larger population — is uncovered. See SHOULD-FIX 2.

### Correction on the corpus

The corpus is **weaker evidence than stated**. `0 of 148` samples declare a nested type of any kind (checked by scanning every `samples/**/*.gs` for an indented `class`/`struct`/`interface` declaration — zero hits), so `PushSmRemap` never fires for the three new call sites anywhere in the corpus. `BYTEDIFF=0` therefore confirms only the `restore: false` early return, which was already guaranteed by construction. The real non-regression evidence for the registered path is `p5`'s byte-identical emit.

## Q5 — `#2977` residual: **claim verified**

The ticket's direct chained spelling, run on both trees:

```
base/q5: COMPILE-FAIL(1)   1 GS0005  1 GS0125  1 GS0157  1 GS0158  1 GS0159
head/q5: COMPILE-FAIL(1)   1 GS0005  1 GS0125  1 GS0157  1 GS0158  1 GS0159
diff of full diagnostic text: IDENTICAL
```

Exit 1, no assembly, on both. Matches `#2977`'s reported set exactly. `#2977` is OPEN and its body describes the residual accurately, including the working local-variable control. The PR is correctly described here.

## Q6 — collisions

`git merge-tree --write-tree` against every open branch: **CLEAN in all 12 pairs**, including `#2903`.

Shared files: exactly one pair, `#2978 × #2912` on `src/Core/CodeAnalysis/Emit/UserTokenResolver.cs`. Hunks `1398/1443/1515/1530` (this PR) vs `88/821/1003+52` (#2912) — ~343 lines apart, different methods. Semantically safe in both directions: `#2912`'s addition there is a TypeSpec cache keyed `(EnumSymbol, ClassRemap, MethodRemap)`, i.e. it already keys on the active remap, so this PR changing which remap is active during three encodes cannot poison it.

`#2979` is `docs/adr/0009-switch-semantics.md` only. `#2980` shares no file with this PR.

**Placement:** independent — no PR blocks it and it blocks none. Landing it early costs `#2912` only a trivial rebase.

`#2966 → #2961 → #2978 → #2972 → #2929 → #2912 → #2964 → #2967 → #2968 → #2963`

---

# Findings

### BLOCKER
None.

### SHOULD-FIX 1 — no asymmetric-nesting test; the non-generic-level skip is uncovered, and five such shapes are pins
`test/Compiler.Tests/Emit/Issue2951NestedGenericStructIteratorTests.cs` · `src/Core/CodeAnalysis/Symbols/StructSymbol.cs:1453`

Every multi-level test in the file has **every** level generic (`TwoNestedGenericLevelsUseAllParameters` is `Outer[T].Middle[U].Inner[V]`). The `if (!tps.IsDefaultOrEmpty)` skip — the branch that keeps ordinals aligned when an enclosing level contributes no parameters — is exercised by no test. `p1`–`p4` and `a1` above all fail on `8dddbbf2` with `BadImageFormatException` and pass on head, so these are pins, not speculative shapes.

**Fix:** add two `[Fact]`s to the existing file using the existing `AssertLoadsAndRuns` helper — `p2` (`class Outer[T] { class Mid { struct Inner[V] } }`, expect `7|z`) and `p1` (`class Outer { struct Mid[U] { struct Inner } }`, expect `q`). `p4` and `a1` are cheap additions on top.

### SHOULD-FIX 2 — the `UserTokenResolver` fix is not iterator-specific and has no non-iterator test
`src/Core/CodeAnalysis/Emit/UserTokenResolver.cs:1401`

`p7` is a pin with no iterator in it: base `MissingMethodException: 'System.String Cell\`2.echo(!0, !0)'`, head `z/9`. `M-UTR` confirms the hunk alone is responsible. All 12 compiler tests are iterators, so a future regression in this shared metadata path is caught only incidentally, through two iterator spellings.

**Fix:** one `[Fact]` with the `p7` source. Also worth widening the PR description — the title says "iterator metadata", but half the change fixes nested-generic method MemberRefs generally.

### NIT 1 — `GetUserStructMethodRef` cache key is computed after the remap scope closes
`src/Core/CodeAnalysis/Emit/UserTokenResolver.cs:1316`

`EncodeOpenStructMethodSignature` disposes its `PushSmRemap` scope before returning, so the key at `:1316` records the **ambient** remap while the signature was encoded under the **pushed** one. Not a correctness bug — the pushed remap is a pure function of `containingType`, which is already in the key, so the key still determines the signature. The only effect is that the same `(parent, name, signature)` can be added as two `MemberReference` rows when reached under two different ambient remaps. Not observed in any probe or in the corpus.

### NIT 2 — stale baseline in the PR description
Suite counts are measured against `8dddbbf2`; `main` is now `b1e693bb1` (`#2973`, test-only). The `+12` / `+2` deltas remain valid relative to the stated base; the absolute totals no longer match `main`.

### NIT 3 — `#2977` body says "PR #2951"
It is PR **#2978**; #2951 is the issue.

---

1. **Item 1 undersells the omission.** The old code keyed the enclosing scope off `ReceiverType` only, so a `shared`/static iterator on a generic type got an *entirely empty* scope — not just a missing flattened prefix. The `?? function.StaticOwnerType as StructSymbol` fallback at `StateMachineEmitter.cs:431` is new behaviour, and `SharedNestedGenericStructIteratorUsesOwnTypeParameter` is the only test covering it.
2. **Item 2 is not iterator-scoped.** `p7` demonstrates the `MemberRef` defect on a plain method with no iterator present. Both the PR title and the issue framing under-describe it.
3. **"`-68` deletions is a lot"** — it is two copies of one 13-line block plus two copies of a two-line offset computation plus doc comments. Production logic is net *added*.
4. **"Confirm `IlVerifier` is never the only assertion"** is moot: `IlVerifier` is not referenced in either new test file. The harness is `Assembly.Load(File.ReadAllBytes(…))` + `GetTypes()` + bounded child-process exec, with both pipes on `ReadToEndAsync()` before `WaitForExit(30_000)` and `Kill(entireProcessTree: true)` on timeout — it does not reintroduce `#2934`'s hang shape.
5. **The corpus claim should be stronger, not weaker:** it is not merely that 0/148 samples contain a nested generic struct *iterator* — 0/148 declare a nested type at all, so the `UserTokenResolver` change is inert across the entire corpus. `BYTEDIFF=0` there is close to vacuous for that hunk.

DavidObando:

Specifically:

DavidObando:
Addressed both SHOULD-FIXes and the documentation/count nits in `6b76a4281`:

- added five asymmetric-nesting load/run pins, including async, plus the ordinary non-iterator `echo(U, T)` MemberRef pin
- expanded the mutation controls/table; Issue2951 is 18/18 and overlapping Issue2927 is 20/20
- broadened the title/body and documented the general nested-generic method fix on #1537
- corrected the corpus claim: 0/148 samples declare any nested type; current `a6b1561f2` vs head remains byte/diagnostic/exit identical
- recounted successful same-base CI: Core 7028 + 1 skipped, Compiler 3906→3924, Interpreter 491→493, LSP 452

Final CI is green and the PR is mergeable.

PR #2979, opened 2026-08-01
Document exhaustive switch metadata in iterators

### LINKED ISSUE #2940: Decide whether iterator lowering should preserve exhaustive switch metadata
## Current status

`BoundPatternSwitchStatement.IsExhaustive` is not preserved by iterator lowering. This is currently an invariant/maintenance gap, not a known user-visible #2906 defect:

- after PR #2926 removed the per-switch emitter throw, the only production consumer of `IsExhaustive` is definite-return CFG projection in `ControlFlowGraph`;
- iterator-returning functions are exempt from definite-return checking;
- the emitter no longer reads the marker;
- an iterator probe is therefore byte-identical whether the marker is preserved or dropped.

Async lowering still preserves the marker, but async non-iterator functions do participate in definite-return analysis.

## Follow-up decision

Choose one of these only if a future consumer requires the metadata after iterator transformation:

1. preserve `IsExhaustive` at every iterator reconstruction of `BoundPatternSwitchStatement`, with a structural test; or
2. document that the marker is analysis-only and may be discarded after definite-return analysis.

Do not prioritize this as a runtime or IL-correctness hole unless a post-iterator consumer is added.

Found during review of PR #2926.

DavidObando:
## Result

#2940's premise is stale: `BoundPatternSwitchStatement.IsExhaustive` already survives iterator lowering. This PR records the narrower verified invariant without claiming an observable synchronous-iterator consequence.

Fixes #2940

## Decision

- keep `IsExhaustive` accurate through binder creation and all three reconstruction paths
- require every constructor call to provide `isExhaustive`; no silent `false` default
- retain the sole CFG read that suppresses the dispatch fall-out edge for exhaustive switches
- explicitly record that no valid synchronous-iterator program is currently known where changing the flag changes diagnostics or runtime values

Preservation remains the decision because each pass reconstructs the same semantic switch, the value is cheap and explicit to carry, and erasing it would make the bound tree inaccurate without a demonstrated benefit.

## Measured limits

Three final-tip compiler probes:

| probe | exit | diagnostics/result |
|---|---:|---|
| top-level function with an unassigned `out` parameter | 1 | one `GS0238` |
| identical shared class member | 0 | zero errors; assembly emitted |
| sequence function with an unassigned `out` parameter | 1 | one `GS0422` and one `GS0238` |

Therefore:

- out-parameter definite assignment is not unconditional across body-binding paths
- ref-kind parameters are rejected on sequence functions, so out-parameter analysis cannot demonstrate a valid iterator consequence
- the silent shared-member case is a separate live analyzer-coverage hole, not evidence for iterator consumption

Structural measurements at the final tip:

- 4 `BoundPatternSwitchStatement` construction sites; all pass the flag
- 1 consuming `IsExhaustive` read, at the CFG fall-out edge
- 1 `RefKindDefiniteAssignmentAnalyzer.Analyze` call in `Binder.cs`
- 6 `RefStructAsyncLivenessAnalyzer.Analyze` calls in `Binder.cs`

## Vacuous test

Removed `IteratorSealedInterface`. The review mutation changed `GSharp.Core.dll` and killed 10 pre-existing cases, but this new case survived because iterator return types bypass GS0100 and GS0178 is decided before the bound flag is consumed. Keeping it would falsely advertise coverage.

## Verification

- rebased onto `61b7a08c1` (newer than the requested `fdba684b6`)
- `Issue2906ExhaustiveSwitchReturnFlowTests`: 17 passed, 0 failed
- three behavior probes above asserted exact exits and diagnostic counts


DavidObando:
## Review: PR #2979 — head `b2d6b09cea5be8fec02d822969e134c011198072`

Read-only review in a `git clone --shared` scratch tree. No build (docs-only). Merge-base is `8dddbbf2`; current `main` is `b1e693bb1`, so every claim was re-verified at `b1e693bb1` rather than at the base.

**Verdict: NOT MERGEABLE as written — one clause of the added sentence is factually wrong. It is a ~15-word reword away from MERGEABLE.**

The PR's *conclusion* is correct and I verified it independently: `IsExhaustive` does survive iterator lowering, and #2940's premise is stale. The problem is the second sentence, which is the part a future maintainer would actually act on.

---

### BLOCKER-1 — `docs/adr/0009-switch-semantics.md:57`: "sole production reader … intentionally unread for iterators" is false

The added line claims:

> Its sole production reader is definite-return control-flow analysis, which exempts iterator-returning functions, so the marker is preserved but intentionally unread for iterators.

The flag is not read by "definite-return control-flow analysis". It is read by **CFG construction**, at `ControlFlowGraph.cs:318`:

```csharp
if (!dispatchComplete && !switchStatement.IsExhaustive)
{
    builder.Add(new BoundGotoStatement(null, defaultArm == null ? endLabel : defaultLabel));
}
```

That line is in the graph builder, not in `AllPathsReturn` (`ControlFlowGraph.cs:61`). `ControlFlowGraph.Create` has **three** production consumers, and only one of them is definite-return:

| consumer | call site | iterator-exempt? |
|---|---|---|
| `ControlFlowGraph.AllPathsReturn` | `Binder.cs:1384`, `:1443`, `:1967` | **yes** — `!IsIteratorReturnType(...)` |
| `RefKindDefiniteAssignmentAnalyzer.Analyze` → `:108 ControlFlowGraph.Create` | `Binder.cs:1391` | **no** — unconditional |
| `RefStructAsyncLivenessAnalyzer.Analyze` → `:129 ControlFlowGraph.Create` | `Binder.cs:1398`, `:1450`, `:1974`, `:2027`, `:2081`, `:2129` | **no** — unconditional |

At `Binder.cs:1384-1401` the iterator guard is attached only to the `AllPathsReturn` call; the two analyzer calls that follow it run on the same `lowered.PreEmitAnalysisBody ?? lowered` with no guard at all. So for an iterator function containing an exhaustive pattern switch, `IsExhaustive` **is** read.

This is a structural proof of the *read*, not of an observable diagnostic difference — I did not construct a diverging program, and I am not claiming one exists. But the direction is load-bearing: `IsExhaustive == true` suppresses the fall-through edge; dropping the flag adds an edge, which is strictly more conservative for both a definite-assignment fixpoint and a liveness fixpoint. Preservation is therefore not inert for iterators.

Why this matters more than a wording nit: the sentence's practical effect is to tell a future maintainer the marker is dead weight on the iterator path. It isn't. That is exactly the inference an ADR should not license.

Suggested reword (same length, no new claims):

> …and the later flattening pass preserves it as well. Its only production reader is CFG construction (`ControlFlowGraph.cs:318`), which decides whether a fall-through edge is emitted. Definite-return analysis exempts iterator-returning functions, but the sibling CFG consumers (`RefKindDefiniteAssignmentAnalyzer`, `RefStructAsyncLivenessAnalyzer`) do not, so the flag is still read for iterators and must keep being preserved.

---

### SHOULD-FIX-1 — the missing test, but not the one that was thrown away

I disagree with the reviewer brief's proposed test, and with the author's stated reason for omitting one. Both are aimed at the wrong stage.

`IteratorMoveNextBodyBuilder.BuildWithFieldAccess` (`IteratorMoveNextBodyBuilder.cs:39`) is reached only from `StateMachineEmitter.cs:612` — **emit time**. The CFG reads established above happen at **bind** time, before the state machine is ever built. So a structural probe asserting the flag survives `BuildWithFieldAccess` pins the one path on which the flag genuinely is unread, and leaves the behaviourally load-bearing path unpinned. Recreating the discarded probe would buy less than it appears to.

The real gap is coverage-shaped and already has a home. `Issue2906ExhaustiveSwitchReturnFlowTests.cs` is a table-driven theory with 16 cases (`OneMember`, `SealedInterface`, `SwitchInsideTry`, `SwitchInsideSelect`, …) feeding `ExhaustiveClosedSwitch_DoesNotReportGs0100` (`:202`) and `NonTotalSwitch_ReportsGs0100` (`:212`). **None of the 16 is an iterator.** `Issue2891SiblingAnalyzerNoDriftTests.cs:206 SharedCfgDiagnosticsRemainUnchanged` exists precisely to pin that the sibling CFG consumers do not drift — and it likewise has no iterator case.

Cheapest correct fix: **one added `Case(...)` in the existing `Issue2906` table** — an iterator function whose body contains an exhaustive switch over a sealed hierarchy, asserting no spurious diagnostic. Roughly eight lines of test data, no new file, no new machinery, and it exercises the bind-time path that actually reads the flag. That is cheaper than the ~15-line structural probe the brief proposed and better targeted than the probe the author deleted.

I'd take that over the ADR paragraph if forced to choose one; ideally the corrected paragraph plus the one case.

---

### Verified clean

**Q1 — construction sites, re-verified at `b1e693bb1` (not at the `8dddbbf2` merge-base).** Still exactly **four**, still zero that omit the flag. No fifth site appeared across the merges since the base:

- `Lowering/Lowerer.cs:1495` — `ps.IsExhaustive` (the "later flattening pass")
- `Lowering/CaptureBoxingRewriter.cs:801` — `node.IsExhaustive`
- `Binding/BoundTreeRewriter.cs:794` — `node.IsExhaustive`
- `Binding/StatementBinder.Blocks.cs:164` — binder-produced `isExhaustive` (origin)

**Q1 — inheritance claim.** `HoistedFieldRewriter.cs:28` is `internal class HoistedFieldRewriter : BoundTreeRewriter`, and `grep PatternSwitch src/Core/CodeAnalysis/Lowering/Iterators/*.cs` returns nothing. No iterator-side override exists. Confirmed.

**Q3 — method name is exact.** `BoundTreeRewriter.cs:767` is `protected virtual BoundStatement RewritePatternSwitchStatement(BoundPatternSwitchStatement node)`. The ADR names it correctly.

**Q3 — ADR is the right home.** ADR-0009 is titled "`switch` semantics — expression + statement, patterns, exhaustive" and `:25` already states the exhaustiveness rule, so someone chasing this question lands here. The line is in `Consequences → Neutral:` (`:54` block), which is the correct bucket — it is an invariant note, not a benefit or a cost. It does not contradict `ControlFlowGraph.cs:318`. No objection to placement.

**Q4 — zero collision, order-independent.** `gh pr diff --name-only` across #2903, #2912, #2929, #2961, #2963, #2964, #2966, #2967, #2968, #2972, #2978, #2980: none touches `docs/adr/0009-switch-semantics.md`. Confirmed safe to merge out of band in any order.

**Q5 — nothing orphaned.** #2975 is OPEN and fully captures the `HoistedFieldRewriter` / `BoundSelectCase.Variable` hoist gap, including the contrast with #2967's async-side guard. Different node type (`BoundSelectCase`, not `BoundPatternSwitchStatement`) and different property, so closing #2940 does not bury it.

---

### NITs

- **`docs/adr/0009-switch-semantics.md:57`** — "the iterator body rewriters inherit … which copies the flag" is accurate but load-bearing on an unstated fact: the guarantee holds only while no iterator rewriter overrides `RewritePatternSwitchStatement`. Worth one clause naming that as the condition, since that is the thing a reader must preserve.
- **Issue #2975 carries no labels.** Everything else in this series is Oahu-labelled; it will be missed in a label-filtered sweep.
- **CI is `UNSTABLE`** — the `build` job is still pending, all six other checks pass. Presumably fine for a docs-only change, but the state is not `CLEAN` yet.

---

1. **"Current `main` = `b1e693bb1`" is right, but the PR's merge-base is `8dddbbf2`** — the same SHA I triaged at. The branch is behind main. Harmless for a docs-only change, and it is why Q1's re-verification was worth doing rather than a formality.
2. **Q2's framing — "a future rewriter that overrides `VisitPatternSwitchStatement` and forgets `IsExhaustive` reintroduces #2940 with no test to catch it" — is weaker than the #2963 analogy it is drawn from, and I'd push back on the analogy itself.** `LoopBackEdgeMutationCollector`'s closed 19-case switch regresses by *addition elsewhere*: someone adds a `BoundNodeKind` in a different file and the collector silently stops covering it, with no textual connection between cause and effect. Here, regression requires an *active edit at the exact site* — someone must deliberately override `RewritePatternSwitchStatement` and drop an argument — and because that override would be on a shared base-class path, it would perturb non-iterator pattern switches too, which the existing GS0100 suite does cover. Those are different hazard classes and the second is materially weaker. The brief's instinct that a test is warranted turns out to be right, but for the reason in BLOCKER-1 (the flag is read on the iterator path) rather than the inheritance-fragility reason given.
3. **"the very probe the author wrote temporarily and then threw away" is the wrong artifact to resurrect.** It targets emit-time (`BuildWithFieldAccess`, reached only from `StateMachineEmitter.cs:612`); the reads are bind-time. Throwing it away was the correct call. See SHOULD-FIX-1.
4. **The author's stated justification — "no structural test added because preservation is inherited behavior rather than a separate iterator policy" — rests on the premise that the flag is unread for iterators.** That premise is what BLOCKER-1 falsifies, so the justification does not survive independently of the sentence it is defending.

---

### On the question the brief actually asked

Whether the author reproduced my triage reasoning or restated it: **reproduced, and extended it.** The four-site enumeration, the absence of an iterator override, and the `Binder.cs` iterator exemption are all independently re-derivable and all check out; the structural probe through `BuildWithFieldAccess` is work I did not do. The conclusion that #2940 is stale is correct.

Where it fell short is the last clause. Tracing `IsExhaustive` to `ControlFlowGraph.cs:318` and stopping at `AllPathsReturn` — rather than at `ControlFlowGraph.Create`'s other two callers — is a one-hop-short trace, and it produced a sentence that inverts the practical guidance. Fix the clause and this is a good, genuinely lazy close.

DavidObando:
## Addendum — re-verified at `main` = `a6b1561f2` (was `b1e693bb1` in my review above)

`main` advanced when #2961 (#2921, GS0100 definite-return on lambda bodies) merged. Since my review's central finding is an enumeration of `ControlFlowGraph` consumers, and #2961 landed in `LambdaBinder.cs` — one of those consumers — I re-ran the enumeration. Read-only, no build.

**Verdict unchanged: NOT MERGEABLE, on BLOCKER-1 only. The finding is slightly stronger than when I filed it.**

### Q1 still holds at `a6b1561f2`

Still exactly **four** `BoundPatternSwitchStatement` construction sites, same line numbers, zero omitting the flag: `Lowerer.cs:1495`, `CaptureBoxingRewriter.cs:801`, `BoundTreeRewriter.cs:794`, `StatementBinder.Blocks.cs:164`. No fifth site.

### BLOCKER-1 survives, and #2961 just illustrated why it matters

`ControlFlowGraph.cs:318` is still the sole read. The two **unguarded** consumers are unchanged: `RefKindDefiniteAssignmentAnalyzer.cs:108` and `RefStructAsyncLivenessAnalyzer.cs:129`. So the flag is still read for iterators, and the ADR clause is still false.

#2961 added a **fifth** definite-return entry point, `LambdaBinder.cs:1302 CheckAllPathsReturn`, called from `:394` and `:888`:

```csharp
if (returnType != TypeSymbol.Void
    && returnType != TypeSymbol.Error
    && !isIteratorReturnType(returnType)
    && !ControlFlowGraph.AllPathsReturn(Lowerer.Lower(body)))
```

It is correctly iterator-exempt, so it does not add a new unguarded path and does not change the conclusion. But it does make the point sharper: **within a single merge, the consumer set of this read grew.** An ADR sentence phrased as "its sole production reader is definite-return control-flow analysis" is a claim about a consumer set, and consumer sets decay — this one decayed in under a day.

That is an argument for the reword I proposed, independent of the accuracy problem: name the **mechanism** (`ControlFlowGraph.cs:318`, CFG construction, which decides whether a fall-through edge is emitted) rather than the **consumer set**. A mechanism-level sentence stays true as consumers are added; a consumer-level sentence has to be re-audited on every merge that touches binding.

### Line numbers in my review are `b1e693bb1`-relative and have shifted

#2961 inserted one line at `Binder.cs:255`, so every `Binder.cs` citation in my review above is **+1** at `a6b1561f2`:

| my review (`b1e693bb1`) | current (`a6b1561f2`) |
|---|---|
| `Binder.cs:1384` | `:1385` |
| `Binder.cs:1391` (`RefKindDefiniteAssignmentAnalyzer.Analyze`) | `:1392` |
| `Binder.cs:1398` (`RefStructAsyncLivenessAnalyzer.Analyze`) | `:1399` |
| `Binder.cs:1443, 1967, 2020, 2074, 2122` | `:1444, 1968, 2021, 2075, 2123` |

`ControlFlowGraph.cs:318`, `BoundTreeRewriter.cs:767`, `HoistedFieldRewriter.cs:28`, `IteratorMoveNextBodyBuilder.cs:39`, `StateMachineEmitter.cs:612` and all four construction sites are unmoved.

### Unaffected

- **SHOULD-FIX-1 stands.** `Issue2906ExhaustiveSwitchReturnFlowTests.cs` still has no iterator case; the cheapest fix is still one added `Case(...)` in that existing table rather than a new structural probe.
- **Q4 unaffected** — the ADR file is docs-only and collides with nothing; still order-independent.
- **No test table to audit.** This PR adds zero tests, so it carries no base-attribution risk from the `b1e693bb1` → `a6b1561f2` move. Nothing in the PR's evidence depends on a suite count.
- The branch is now two merges behind `main` (base `8dddbbf2`). Still irrelevant to correctness for a docs-only change.

DavidObando:
## Review: NOT MERGEABLE — the sentence is still false at the tip, but the fix is one paragraph and I have written it below

Head `b2d6b09ce`, computed merge-base `8dddbbf28`, tip `f70626452`. Read the **net state at the tip**, not per-commit. The sentence has **not** been reworded since the prior review — it is byte-identical — so this is a re-derivation rather than an inheritance.

Per the method constraint for docs-only changes I ran **no build and no test host**, and I did not fabricate a behavioural differential. This sentence makes a checkable claim about `IsExhaustive`, CFG construction and iterator exemption, so all evidence below is from reading the producers and consumers at the tip.

---

### The claim, decomposed

The single added bullet (`docs/adr/0009-switch-semantics.md:57`) asserts three things. **One is true and two are false.**

**Claim 1 — "survives iterator lowering because the rewriters copy the flag" — TRUE.** There are exactly four construction sites for `BoundPatternSwitchStatement` and every one carries the flag through:

| site | |
|---|---|
| `StatementBinder.Blocks.cs:164` | original bind |
| `BoundTreeRewriter.cs:794` | `node.IsExhaustive` |
| `CaptureBoxingRewriter.cs:801` | `node.IsExhaustive` |
| `Lowerer.cs:1495` | `ps.IsExhaustive` (flattening) |

No site drops it. This is worth keeping — and it is **more** valuable than the PR realises, because it corrects issue #2940's own opening premise, which states that "`IsExhaustive` is not preserved by iterator lowering". That premise is false at the tip. The bullet omits `CaptureBoxingRewriter`, which is a completeness gap rather than an error.

**Claim 2 — "Its sole production reader is definite-return control-flow analysis" — FALSE.**

The sole *read site* is `ControlFlowGraph.cs:318`, which is inside graph **construction**, not inside any one analysis:

```csharp
if (!dispatchComplete && !switchStatement.IsExhaustive)
{
    builder.Add(new BoundGotoStatement(null, defaultArm == null ? endLabel : defaultLabel));
}
```

A non-exhaustive dispatch gets an extra fall-out edge; an exhaustive one does not. That shapes the graph itself, so **every** consumer of the graph inherits the effect. There are three:

| consumer | iterator-exempt? |
|---|---|
| `ControlFlowGraph.AllPathsReturn` (definite-return) | **yes** |
| `RefKindDefiniteAssignmentAnalyzer.cs:108` — out-param definite assignment | **no** |
| `RefStructAsyncLivenessAnalyzer.cs:129` — ref-struct liveness across `await` | **no** |

**Claim 3 — "which exempts iterator-returning functions, so the marker is preserved but intentionally unread for iterators" — FALSE, and this is the damaging one.**

The `IsIteratorReturnType` guard is attached to the **`AllPathsReturn` call only**. The same pattern appears at all seven call sites (`Binder.cs:1385/1444/1968/2021/2075/2123`, `LambdaBinder.cs:1307/1714`):

```csharp
if (function.Type != TypeSymbol.Void && !IsIteratorReturnType(function.Type) && !ControlFlowGraph.AllPathsReturn(lowered))
{
    binder.Diagnostics.ReportAllPathsMustReturn(...);   // iterator-exempt
}

RefKindDefiniteAssignmentAnalyzer.Analyze(...);         // NOT guarded
RefStructAsyncLivenessAnalyzer.Analyze(...);            // NOT guarded
```

`RefKindDefiniteAssignmentAnalyzer.Analyze` bails only on `body == null || function == null`; it then always reaches `ControlFlowGraph.Create` at `:108`. It is therefore **unconditional for every function body with a body, iterators included**. `RefStructAsyncLivenessAnalyzer` is narrower — `enclosing.IsAsync`, then only when by-ref-like locals exist — but async iterators satisfy both.

So for an iterator-returning function the flag **is** read. "Intentionally unread for iterators" is not true of any iterator that reaches out-parameter definite assignment, which is all of them.

The codebase says so in its own words. `RefKindDefiniteAssignmentAnalyzer.cs:69-71`:

> `ControlFlowGraph.Create` is already called **unguarded** against this same lowered body a few lines above in `Binder.cs`

---

### Why this is a blocker and not a nit

The false half is the entire content of the PR — the change is one line. Merging it writes into a normative ADR the exact inference that would justify **deleting** the preservation this ADR is documenting. Issue #2940 offers two options, and its option 2 is *"document that the marker is analysis-only and may be discarded after definite-return analysis."* Under the corrected reading, **option 2 is unsafe**: discarding the flag after definite-return would change CFG shape for iterator bodies, and therefore change out-parameter definite-assignment results — a silent diagnostic change, exactly the class no CI job asserts. The ADR as written points a future maintainer straight at it.

The false claims are **not the author's invention** — they are transcribed faithfully from #2940's own "Current status" section, which asserts that the only production consumer is definite-return CFG projection and that "an iterator probe is therefore byte-identical whether the marker is preserved or dropped". The issue is the origin of the error; the PR propagates it into docs.

---

### The sentence that should replace it

> `BoundPatternSwitchStatement.IsExhaustive` survives iterator lowering: every reconstruction of the node copies the flag — `BoundTreeRewriter.RewritePatternSwitchStatement`, `CaptureBoxingRewriter`, and the flattening pass in `Lowerer` — and no construction site drops it. Its only read is in control-flow-graph construction (`ControlFlowGraph.cs`), where a dispatch that is *not* exhaustive receives an extra fall-out edge to the default/end label, so the flag shapes the graph itself rather than being consumed by one analysis. Three analyses build that graph: definite-return (`ControlFlowGraph.AllPathsReturn`), out-parameter definite assignment (`RefKindDefiniteAssignmentAnalyzer`), and ref-struct liveness across `await` (`RefStructAsyncLivenessAnalyzer`). Only definite-return is exempted for iterator-returning functions; out-parameter definite assignment runs unconditionally on every function body. The marker is therefore **read for iterators** and must remain accurate through every lowering pass — it must not be treated as analysis-only or discarded after definite-return analysis.

---

### Does `Fixes #2940` become correct once the sentence is fixed? — Yes

#2940 is a decision issue: *"Decide whether iterator lowering should preserve exhaustive switch metadata."* The corrected paragraph decides it — preserve, it already is preserved, and here is the consumer that makes preservation load-bearing rather than cosmetic. That is a complete resolution and `Fixes #2940` is then legitimate.

As written, however, `Fixes #2940` closes the issue **on a false rationale**, and closes it in the direction that licenses the unsafe option. Two requests:

1. Post the corrected consumer map as a comment on **#2940** as well, so the stale premise in its "Current status" section cannot outlive the issue. Anyone who later reads #2940 alone still gets the wrong answer.
2. State explicitly in the ADR bullet, or in the PR body, that **option 2 in #2940 is rejected** — otherwise the option survives as a plausible future cleanup.

---

### Checks that pass — recorded so they are not re-derived

- **Textual mergeability: CLEAN** onto `f70626452`. `main` has not touched `docs/adr/0009-switch-semantics.md` since the merge-base, so there is no rebase pressure on this file.
- **Docs gate (#2981) mirror risk: not triggered.** The PR's three-dot diff is exactly one file — the ADR. It touches no `docs/diagnostics.md` row, so the duplicate-row / `conflictingSeverities` failure mode cannot fire, and no diagnostic ID is allocated or referenced.
- **`GS0506` prior-art check: not applicable.** I checked specifically for the failure mode where an ADR presents a boundary as new when `GS0506 FixedStatementCannotSuspend` already covers the shape. This bullet proposes no diagnostic and makes no boundary claim — it is purely a metadata-survival note — so that pattern does not arise here.
- **No behavioural differential attempted**, correctly: the change is docs-only. The claim was checkable by reading consumers, and that is where every piece of evidence above comes from.

### NIT

The bullet sits under **Neutral consequences** of ADR-0009's decision to adopt `switch`. It is an implementation note about lowering and analysis plumbing, not a consequence of choosing the construct, and it will age against code that moves. Consider placing it under a short "Implementation notes" heading, or referencing symbol names without line-level structure so it degrades gracefully.

DavidObando:
Addressed review at head `80743629bd3a210739a80a4ecdf4b2d0ecec412f`.

- Corrected ADR: reader is CFG construction; iterator-exempt definite-return is distinguished from unguarded ref-kind/liveness consumers.
- Named preservation condition: iterator rewriters must not override `RewritePatternSwitchStatement` without retaining the flag.
- Added `IteratorSealedInterface` to existing Issue2906 theory; did not restore emit-time structural probe.
- Skipped duplicate Issue2891 case because Issue2906 compilation already runs both sibling analyzers and no additional diagnostic expectation was identified.

Same-base executed counts (`a692200af` → HEAD): Core `7029+1` → `7030+1`; Compiler `3949` → `3949`; Interpreter `494` → `494`; LSP `452` → `452`. Release serial warnings-as-errors build passed; Core DLL hashes matched and were nonzero.

DavidObando:
Applied the review's replacement verbatim at head `9a3be3f608824218cc940e4a2c4decec6cf2bedc`.

- Moved invariant from Neutral consequences to `Implementation notes`.
- Included `CaptureBoxingRewriter` in the complete reconstruction map.
- Explicitly rejected #2940 option 2: the flag must not be discarded after definite-return.
- Posted corrected consumer map on #2940: https://github.com/DavidObando/gsharp/issues/2940#issuecomment-5151913912

DavidObando:
## Verdict: NOT MERGEABLE

Reviewed at head `9a3be3f60` against `main` = `a4a660f32`. Scope computed from the merge base (`git merge-base` = `7a6307a79`), not two-dot: **2 files, +15/-0**, matching `gh pr view`. `git merge-tree --write-tree a4a660f32 9a3be3f60` → **rc=0, clean**. No new ADR file (this edits the existing `0009`), so no `0152`/`0153` collision. Zero `src/` changes, so no diagnostic IDs allocated and the bidirectional docs gate does not apply.

The merge and build state is fine. The problem is the content, and for a decision record the content *is* the deliverable.

### BLOCKER 1 — the load-bearing sentence is false on `main`

`docs/adr/0009-switch-semantics.md` (added paragraph) states:

> Only definite-return is exempted for iterator-returning functions; out-parameter definite assignment runs unconditionally on every function body.

The second clause is false. `RefKindDefiniteAssignmentAnalyzer.Analyze` has **exactly one** call site in `src/` — `Binder.cs:1392`. The sibling analyzer `RefStructAsyncLivenessAnalyzer.Analyze` has **six** (`Binder.cs:1399,1451,1975,2028,2082,2130`), one per body-binding path. The class/struct method path at `Binder.cs:1440-1452` runs `AllPathsReturn` and the liveness analyzer but never the out-param analyzer.

Measured behaviourally on `main` `a4a660f32` with a positive control, so the silence is meaningful:

```
// top-level func, unassigned out  -> rc=1
error GS0238: Out parameter 'a' must be definitely assigned on every path before the function returns.

// identical defect inside `class Holder { shared { ... } }` -> rc=0, no diagnostic at all
Wrote method.dll
```

### BLOCKER 2 — the conclusion "read for iterators" does not follow, by construction

Even on the one path where the out-param analyzer does run, it can never observe anything on an iterator: ref-kind parameters are categorically rejected on sequence functions.

```
func F(x Expr, out r int32) sequence[int32] { ... }
error GS0422: Ref-kind parameter 'r' cannot appear on a sequence function.
```

`DiagnosticDescriptors.cs:334` — `RefKindOnAsyncOrIterator`. An iterator cannot have an out parameter, so out-parameter definite assignment has nothing to check there. The paragraph's stated chain of reasoning to "**read for iterators**" therefore fails at both links.

### BLOCKER 3 — the only executable pin is vacuous (mutation: 1/1 survival on the PR's own hunk)

Mutated all three propagation sites the ADR names, to drop the flag:

- `src/Core/CodeAnalysis/Binding/BoundTreeRewriter.cs:794` `node.IsExhaustive` → `false`
- `src/Core/CodeAnalysis/Lowering/CaptureBoxingRewriter.cs:801` `node.IsExhaustive` → `false`
- `src/Core/CodeAnalysis/Lowering/Lowerer.cs:1495` `ps.IsExhaustive` → `false`

Build clean (0 errors) and the mutant was live — `GSharp.Core.dll` md5 `c45ea71173f0c82198c765c3e89d37e2` → `1b22597ed92f9f00a468b03cd5cd3326`. Same population (18 tests) in both runs.

| run | result |
|---|---|
| baseline `9a3be3f60` | 18/18 pass |
| all three sites dropped | **10 failed / 8 passed** |
| restored + rebuilt | 18/18 pass, md5 back to `c45ea711…` |

The 10 that died are all pre-existing cases: `OneMember`, `TwoMembers`, `ManyMembers`, `OrPattern`, `DuplicateCompleteArms`, `SealedClass`, **`SealedInterface`**, `SwitchInsideTry`, `TryInsideArms`, `SwitchInsideSelect`.

`IteratorSealedInterface` — the only code this PR adds — **passed under the mutation**, confirmed by name from detailed console output.

The reason is structural. `test/Core.Tests/CodeAnalysis/Binding/Issue2906ExhaustiveSwitchReturnFlowTests.cs:211-219` asserts only `DoesNotContain(d => d.IsError)` and `emittedLength > 0`. For an iterator, `GS0100` is exempted at `Binder.cs:1385` (`!IsIteratorReturnType(...)` short-circuits before `AllPathsReturn` is called), and `GS0178` is a bind-time `ExhaustivenessAnalyzer` diagnostic that never consults the bound-node flag. Probe:

```
sealed interface Expr, one arm of two, returning sequence[int32] -> rc=1, GS0178 only (no GS0100)
same shape returning int32                                        -> rc=1, GS0100 + GS0178
```

So no error can fire for the new case whether the flag survives lowering or not. Note the new case differs from the pre-existing `SealedInterface` case only by returning `sequence[int32]` instead of `int32` — and that one change is exactly what makes it blind.

### What is actually true (verified, and worth keeping)

The rest of the paragraph holds. There are exactly **four** construction sites of `BoundPatternSwitchStatement` in `src/`, and all four pass the flag: `BoundTreeRewriter.cs:794`, `StatementBinder.Blocks.cs:164`, `CaptureBoxingRewriter.cs:801`, `Lowerer.cs:1495`. The only consuming read is `ControlFlowGraph.cs:318`, and it is exactly the described fall-out edge:

```csharp
if (!dispatchComplete && !switchStatement.IsExhaustive)
{
    builder.Add(new BoundGotoStatement(null, defaultArm == null ? endLabel : defaultLabel));
}
```

Exactly three analyses build the graph, as claimed. And "only definite-return is exempted for iterators" is correct — all six `AllPathsReturn` call sites carry `!IsIteratorReturnType`, and neither other analyzer carries such a guard.

### Smallest change that would make this mergeable

1. Replace the false clause. The property you want is genuinely delivered by a different analyzer: `RefStructAsyncLivenessAnalyzer` runs on **all six** body-binding paths, unguarded by `IsIteratorReturnType`, and calls `ControlFlowGraph.Create` at `RefStructAsyncLivenessAnalyzer.cs:129`. That — not out-param assignment — is why a CFG gets built over iterator bodies.
2. Be honest that the read is currently **unobservable** for synchronous iterators. I could not construct a case where dropping the flag changes iterator behaviour, and my attempt to build one failed (see below). If the decision is "keep the flag accurate because it is cheap and the CFG is genuinely built", say that — it is defensible and it is what the evidence supports. As written, the ADR claims a stronger, measurable property that does not hold.
3. Either drop the `IteratorSealedInterface` case or replace it with one that discriminates. As it stands it is a green test that pins nothing.

### Where my own probe failed to discriminate

I tried to construct a non-vacuous replacement: an iterator with an `out` parameter assigned in every arm of an exhaustive switch, so that a dropped flag would add a fall-out edge and trigger `GS0238`. It does not work — `GS0238` fires on unmutated `main` anyway, alongside `GS0422`, because ref-kind parameters are illegal on sequence functions in the first place. I am reporting this rather than quietly dropping it, because it is the same evidence that underpins BLOCKER 2: I have found no observable consequence of the flag for a synchronous iterator, and neither has this PR.

### NIT

`src/Core/CodeAnalysis/Binding/BoundPatternSwitchStatement.cs:22` declares `bool isExhaustive = false`. The ADR asserts "no construction site drops it", which is true today, but the defaulted parameter means a fifth construction site that simply omits the argument compiles clean and silently means "not exhaustive". For an invariant this ADR is explicitly asking future maintainers to preserve, making the parameter required would enforce it at compile time instead of by prose.

Issue #2974, opened 2026-08-01
Gsharp.Extensions build races GSharp.Core.dll rewrite under parallel MSBuild


DavidObando:
## Summary

`src/Sdk/Gsharp.Extensions/Gsharp.Extensions.csproj` invokes the freshly built `gsc` during its build. Under parallel MSBuild (`-m`, the default) this races the rewrite of `out/bin/Release/Compiler/GSharp.Core.dll`, so `gsc` loads a partially written assembly and the build fails:

```
Gsharp.NET.Sdk.Bootstrap.targets(114,5): error :
  Unhandled exception. System.IO.FileNotFoundException:
  Could not load file or assembly 'GSharp.Core, Version=0.3.0.0…'
  [.../src/Sdk/Gsharp.Extensions/Gsharp.Extensions.csproj]
```

## Why this is worse than a failed build

The failing build leaves **truncated, zero-byte** `GSharp.Core.dll` copies behind. Observed state after `BUILD_EXIT=1`, `4 Error(s)`:

| copy | md5 |
| --- | --- |
| `out/bin/Release/Compiler/GSharp.Core.dll` | `d41d8cd98f00b204e9800998ecf8427e` |
| `out/bin/Release/Compiler.Tests/GSharp.Core.dll` | `d41d8cd98f00b204e9800998ecf8427e` |
| `out/bin/Release/Core.Tests/GSharp.Core.dll` | `d41d8cd98f00b204e9800998ecf8427e` |
| `out/bin/Release/Interpreter.Tests/GSharp.Core.dll` | stale copy from the previous build |

`d41d8cd98f00b204e9800998ecf8427e` is the md5 of a zero-byte file.

A subsequent `dotnet test --no-build` against `Interpreter.Tests` therefore runs against the **previous** build and reports a green result with a plausible test count. Timestamp-based freshness checks do not catch this, because the files were genuinely written.

## Likely the same root cause as recurring CI flakes

Two intermittent CI failures have been resolved by `gh run rerun --failed` rather than investigated:

- `MSB3883 … refint/GSharp.Core.dll … used by another process`
- `MSB3030 … apphost … not found` under `-graph`

Both are concurrent access to `GSharp.Core.dll` during a parallel build, which matches this race.

## Repro

Build the solution in Release with default parallelism, repeatedly. The failure is intermittent — it did not reproduce on demand and appeared during an unrelated rebuild.

## Workaround

[...]

PR #2980, opened 2026-08-01
Fix constrained static property receiver chains

### LINKED ISSUE #2962: Chained member access on constrained static property reaches emitter ICE
## Problem

Chaining a member access directly from a constrained static-virtual interface property reaches emission with an unresolved expression and reports an internal compiler error:

```text
GS9998: EmitDiagnosticException: an unresolved (error) expression reached code emission
```

## Reproduction without nullable types

```gsharp
sealed interface I[T] {
    shared { prop V T { get; } }
}

struct C : I[string] {
    shared { prop V string -> "value" }
}

func Read[T I[string]](w T) int32 {
    return T.V.Length
}
```

This is pre-existing:

| Compiler | Result |
|---|---|
| `17789519` (`origin/main`) | GS9998 |
| PR #2955 head | GS9998 |

## Nullable shape exposed by #2955

The same program with a nullable declared slot:

```gsharp
sealed interface I[T] {
    shared { prop V T? { get; } }
}

struct C : I[string] {
    shared { prop V string -> "value" }
}

func Read[T I[string]](w T) int32 {
    return T.V.Length
}
```

| Compiler | Result |
|---|---|
| `17789519` (`origin/main`) | GS0397 (static property rejected before emission) |
| PR #2955 head | GS9998 |

PR #2955 correctly makes the property implementation valid, but consequently widens reachability of this existing chained-access ICE.

## Workaround

Store the constrained property result before accessing its member:

```gsharp
func Read[T I[string]](w T) int32 {
    var value = T.V
    return value.Length
}
```

This form compiles and runs.

DavidObando:
## Summary

`T.V.Length` parses as a right-leaning accessor, `T.(V.Length)`. The constrained static accessor binder handled only a terminal call or name, so initializer-position `T.V` bound correctly while receiver-position chains fell through to an unreported `BoundErrorExpression` and later tripped GS9998.

This change binds the constrained static property head first, then routes the remaining member, method, index, or null-conditional suffix through the existing instance-access binders. This is failure mode 2: the non-nullable program is legal and now emits real constrained-call IL rather than being rejected.

Nullable property reads now use the substituted semantic property type while their emitted call still targets the erased CLR slot. Therefore direct `T.V.Length` for declared `T?` reports the existing nullable-receiver result, GS0158, while `T.V?.Length` compiles and runs.

Fixes #2962

## Coverage

Runtime child-process coverage loads every emitted type and executes:

- `T.V.Length`
- `T.V.Method()`
- `T.V[0]`
- `T.V.Field`
- `T.V?.Length`
- `T.V?[0]`
- `T.V.W.X`
- a generic property type, `T.V.Item.Length` where `V` is `Box[string]`

The good implementation prints `value` and `5`; a control implementation prints `broken` and `6`, proving constrained dispatch reaches the selected implementation. Child stdout/stderr are drained concurrently before a bounded wait; timeout kills the process tree and drains partial output after termination.

## Mutation evidence

Five semantic pins, zero guards:

1. Disable normal nested-chain routing: runtime test fails with GS9998.
2. Treat `?.` as plain access: runtime test fails with GS0158 at `T.V?.Length`.
3. Disable direct index routing: runtime test fails with GS9998.
4. Treat `?[0]` as plain indexing: runtime test fails with GS0116 on `string?`.
5. Revert semantic property type to the erased slot type: nullable diagnostic test compiles unexpectedly (exit 0).

Each mutation was restored and rebuilt before the next probe.

## Verification

Base `8dddbbf285c2b4e94d06647671b44f9b637537f7`:

- Core: 7028 passed, 1 skipped
- Compiler: 3893 passed
- Interpreter: 489 passed
- LSP: 452 passed

Head:

- Core: 7028 passed, 1 skipped
- Compiler: 3895 passed
- Interpreter: 489 passed
- LSP: 452 passed
- Release build: 0 warnings, 0 errors

Current-main behavior before the fix was GS9998 for both non-nullable and nullable repros. After the fix, the non-nullable shape executes and the nullable direct chain reports GS0158; null-conditional access executes and prints `5`.

Corpus: 0 of 148 sample files exercise constrained static-virtual property access in receiver position. The executable conformance corpus passed, but provides no evidence for this bug. No local e2e run.


DavidObando:
## Review: PR #2980 — head `53e6d3b0a`

**Verdict: MERGEABLE.** No blockers. Two SHOULD-FIX items, two NITs.

Reviewed at `53e6d3b0a` in isolated scratch clones (`git clone --shared --no-checkout`), solution built with `dotnet build -m:1`, exit code gated. `GSharp.Core.dll` md5 `70fdf846…` identical across `out/bin/Release/{Compiler,Compiler.Tests,Core.Tests,Interpreter.Tests}` — no zero-byte DLL (#2974). `out/bin/Release/Gsharp.Extensions/` populated, so no `Sample_BuildsAndMatchesGolden` closure gap.

---

## Q1 — suffix routing: proven correct, from emitted IL

Dumped IL with a `System.Reflection.Metadata` reader, decoding `constrained.` (`0xFE 0x16`), `call`, `callvirt`, `box`:

| probe | substitution | emitted IL |
|---|---|---|
| `ReadLength` | `T` = **class** `RefProvider`, prop `string` | `constrained.<T> call get_V` → `callvirt get_Length` |
| `ReadMethod` | same | `constrained.<T> call get_V` → `callvirt ToUpper` |
| `ReadX` | `T` = **struct**, prop = compilation-local struct `Pt` | `constrained.<T> call get_V` → `ldfld` |
| `ReadSum` | same, instance method on the struct value | `constrained.<T> call get_V` → `call Sum` |

`constrained.` appears **exactly once, on the type-parameter head, and never on the suffix** — correct, because the suffix receiver is an ordinary reference/value. It is also **not dropped** where required. No `box` on the struct path. Both programs `Assembly.Load` + `GetTypes()` and execute in a bounded child process: `8`/`REFVALUE` and `41`/`42`.

Base `8dddbbf2` reports **GS9998** for both.

**Explicit interface implementation of a static-virtual property: not proven.** Every syntax I tried was rejected identically on base and head (parse errors + GS0397), so I could not construct it. Structurally the read resolves `slotProp.GetterSymbol` on the interface slot and emits `constrained.` against it, so explicit vs implicit implementation is a runtime dispatch detail the binder never sees — but I am reporting this as unproven, not proven.

**Verification standard:** the PR's own tests use `Assembly.Load(File.ReadAllBytes(...)).GetTypes()` + child-process execution with both pipes drained via `ReadToEndAsync()` before a bounded `WaitForExit(30_000)` and `Kill(entireProcessTree: true)`. **`IlVerifier.Verify` is not called anywhere in the file.** The demand is already met.

---

## Q2 — `DeclarationBinder.InterfaceVerification.cs` is a visibility change only

`DeclarationBinder.InterfaceVerification.cs:723` — `private static` → `internal static` on `GetInterfacePropertyExpectedType`. **The body is untouched.** It cannot change the accept/reject set for any interface implementation, because no call site inside `DeclarationBinder` changed and no logic changed.

The **semantic** change is in the expression binder: `ExpressionBinder.Access.MemberLookup.cs:3494` switches the constrained static property read from `GetInterfacePropertySlotType` (erases reference nullability) to `GetInterfacePropertyExpectedType` (preserves substituted nullability, except where the substitution would produce a value-type nullable). Reusing the existing helper instead of writing a new one is the right call — one word beats a duplicated substitution routine.

**No overlap with #2968.** #2968 edits `IsUnsupportedImplicitInterfaceIndexer` (`:705`), `TryGetConstructedInterfacePropertyImplementation` (`:738`) and `HasMatchingImportedBaseProperty` (`:796`) — different functions, and it does not touch `GetInterfacePropertyExpectedType` or `GetInterfacePropertySlotType`. #2980's new `IndexExpressionSyntax` case indexes the *value returned by* a static property, not an interface indexer slot. Textually and semantically disjoint.

---

## Q3 — nullable behaviour, verified by execution

- `T.V.Length` on a `T?` slot → `error GS0158: Cannot find member Length.`, **no GS9998**. ✔
- `T.V?.Length ?? -1` → compiles, runs, prints `5`. ✔
- `T.V?[0]` → runs, prints `v` (PR test). ✔
- `T.V?.Length!!` → base GS9998, head runs, prints `5`. ✔ (fixed, and not claimed by the author)
- `T.V!!.Length` and `T.V!!` → work on **both** base and head.

**Correction:** there is no `!` null-forgiving operator in G#; `s!.Length` is `error GS0005` on a plain local. The operator is `!!`.

**Nullable value type `T`:** unreachable. `sealed interface INul[T] { shared { prop V T? } }` with `T` = a struct is rejected by **GS0397 on both base and head** — the ADR already states `T?` where `T : struct` retains `Nullable<T>` and cannot be implemented. Nothing for this PR to get wrong.

---

## Q4 — no third `ClrType ?? typeof(object)`

The 32 added lines contain **no** `ClrType` reference at all. The one occurrence in the touched file (`ExpressionBinder.Access.MemberLookup.cs:377`, `ImportedBaseType?.ClrType ?? typeof(object)`) is pre-existing and unrelated.

Empirically the compilation-local-struct case is exercised and correct: the PR's own `FieldHolder`, `Middle`/`Leaf`, `Box[string]` and my `Pt` are all compilation-local structs used as the property type, and all emit and execute with the right values. If a null `ClrType` were being papered over with `object`, `ReadSum` would have boxed or thrown — the IL shows a direct `call Sum` and no `box`.

---

## Q5 — where the routing stops: 15 shapes, no hole found

`ExpressionBinder.Access.MemberLookup.cs:3510` still ends in `default: return new BoundErrorExpression(null);` — the *unreported* error that is the original defect. I tried to reach it:

| shape | base `8dddbbf2` | head |
|---|---|---|
| `T.V.Length` (issue repro) | GS9998 | ✔ runs |
| `T.V.ToUpper()` | GS9998 | ✔ runs |
| `T.V[0]` | GS9998 | ✔ runs |
| `T.V.Field` (local struct) | GS9998 | ✔ runs |
| `T.V.W.X` (depth 3) | GS9998 | ✔ runs |
| `T.V.X.Y.Z.Length` (**depth 4**) | GS9998 | ✔ runs `4` |
| `T.V[1][0]` (**double index**) | GS9998 | ✔ runs `c` |
| `T.V.Pick[int32](7)` (**generic method suffix**) | GS9998 | ✔ runs `7` |
| `T.V?.Length` / `T.V?[0]` | GS9998 | ✔ runs |
| `T.V?.Length!!` | GS9998 | ✔ runs |
| `T.V!!.Length`, `T.V!!` | ✔ runs | ✔ runs |
| `T.V.Nope` (**missing suffix member**) | GS9998 | **GS0158** |
| `T.Nope.Length` (**missing head**) | GS9998 | **GS0333** |
| `T.V.F = 9` (assignment target) | GS0499 | GS0499 |

Both error paths now produce real diagnostics instead of the ICE — that is the part that matters most, and it is the part the author did not claim. The fix is also **wider than advertised**: depth-4 chains, chained indexers, generic-method suffixes and `?.X!!` all work and none are in the PR description.

The reachable post-dot node kinds are `NameExpressionSyntax`, `CallExpressionSyntax`, `AccessorExpressionSyntax`, `IndexExpressionSyntax` (all handled), plus `UnaryExpressionSyntax`/`ObjectCreationExpressionSyntax`, which only arise under null-conditional and are routed to `BindNullConditionalAccessExpressionCore` — which handles `BangBangToken` itself (`ExpressionBinder.Access.Accessor.cs:1236`). I could not construct a program that reaches `default:`. See NIT 1.

---

## Q6 — counts reconcile exactly

**The branch's merge-base is `8dddbbf2`, not `b1e693bb1`.** `b1e693bb1` is `8dddbbf2` + PR #2973 (+1 Compiler test), so the correct base row is Compiler **3893**.

Executed at head `53e6d3b0a` (`--no-build`, Release, whole-solution build first):

| suite | base `8dddbbf2` | head `53e6d3b0a` | Δ |
|---|---|---|---|
| Compiler | 3893 | **3895 / 0 failed** (1 h 20 m) | **+2** |
| Core | 7028 + 1 skipped | **7028 + 1 skipped / 0 failed** (5 m 23 s) | 0 |
| Interpreter | 489 | **489 / 0 failed** (30 s) | 0 |
| LSP | 452 | not run (untouched) | — |

+2 = the two `[Fact]`s in the new file. Fully explained. **No golden-sample failures**, confirming the build closure was correct. 3895 passing tests is also the over-fire evidence for the nullability change below — nothing existing regressed.

---

## Q7 — collisions and merge order

`git merge-tree --write-tree` **CLEAN** against `origin/main` `b1e693bb1` and against **every** open PR: #2912, #2929, #2961, #2963, #2964, #2966, #2967, #2968, #2972, #2978, #2979.

Semantically: the only shared file with any open PR is `DeclarationBinder.InterfaceVerification.cs` (#2968), at disjoint functions — see Q2. `ExpressionBinder.Access.MemberLookup.cs` is touched by no other open PR.

**Order is free.** If you want the shared file single-owner at a time, place #2980 **after #2968**: `#2966 → #2961 → #2972 → #2929 → #2912 → #2964 → #2967 → #2968 → #2980 → #2963`.

---

## SHOULD-FIX

### S1 — the interpreter is fixed too, and nothing pins it
`gsi` on the value-type chain:

- base `8dddbbf2`: `GS9999: Unexpected node ErrorExpression`
- head: `42` — matches compiled output exactly

The binder is shared, so `gsi` gets the fix for free, and the base failure is a *different* ICE code (GS9999, not GS9998). The PR adds **zero** Interpreter tests. One `Issue2962…InterpreterTests` fact would pin a path that is currently demonstrated-fixed and unguarded. Cheap.

### S2 — a previously-accepted program is now rejected, and no test pins it
`ExpressionBinder.Access.MemberLookup.cs:3494`. Initializer position, `T?`-declared slot:

```gsharp
func Take[T INul[string]](w T) string {
    let s string = T.V     // slot declared T?, T = string
    return s
}
```
- base: **compiles and runs**, prints `value`
- head: `error GS0156: Cannot convert type 'string?' to 'string'. An explicit conversion exists (are you missing a cast?)`

Head is **right** — base was erasing the `?` and letting a possibly-nil value into a non-null `string`. The message is precise, `let s string? = T.V` compiles on both, the ADR update covers it, and the reachable surface is tiny (`T?` static-virtual slots only exist since #2945). But it is a source-breaking change in *accepted-program* space, it is broader than the PR title, and the only test near it (`NullableDirectChain_ReportsMemberDiagnosticInsteadOfIce`) pins the member-access shape, not the assignment shape. Add one pin so the next erasure change cannot silently flip it back.

---

## NIT

**N1 — `ExpressionBinder.Access.MemberLookup.cs:3510`.** `default: return new BoundErrorExpression(null);` returns an error with **no diagnostic** — literally the defect this PR fixes. I could not reach it in 15 shapes, so it is not a blocker, but the lazy insurance is one line: report `ReportStaticVirtualMemberNotFoundOnTypeParameter` (or any real diagnostic) there instead of a silent error. Then this class of bug cannot recur on a future grammar addition.

**N2 — diagnostic noise on an already-rejected program.** `func Read[T IValue[T]](w T)` with a self-referential constraint reports `GS0368` on base and `GS0368 + GS0159` on head — head binds one step further before failing. Both reject; only the extra line differs.

---

1. **Base is `8dddbbf2`, not `b1e693bb1`.** `b1e693bb1` = `8dddbbf2` + #2973. Correct base row is Compiler **3893**, so head **3895** is +2, not +1.
2. **Q2's "smell" does not survive contact.** The `+1/-1` is `private static` → `internal static` with an untouched body. It cannot widen or narrow any accept/reject set. Reusing the existing helper rather than duplicating the substitution logic in the expression binder is the smaller diff, not a smuggled change. The real semantic edit is one identifier at `ExpressionBinder.Access.MemberLookup.cs:3494`.
3. **Q3's `T.V!.Length`** — G# has no `!` null-forgiving operator (`error GS0005` on a plain nullable local). The operator is `!!`, and it works on base and head.
4. **Nullable value-type `T` is unreachable**, rejected by GS0397 at declaration on both sides — not a gap this PR can have.
5. **Q4's third `ClrType ?? typeof(object)` is not here.** Zero `ClrType` references in the added code, and compilation-local structs demonstrably emit and execute correctly.
6. **Q1's verification demand was already satisfied** — the new tests never call `IlVerifier.Verify`; they load the assembly and execute it in a bounded child process with the correct concurrent-drain pattern.
7. **Q5 was the predicted blocker and it did not land.** Fifteen shapes, no hole. The fix is *wider* than the author claims — depth-4 chains, chained indexers, generic-method suffixes, `?.X!!`, and both error paths.
8. **#2968 has no semantic overlap** despite touching the same file.

---

*Probes: 20 programs, each compiled by base `8dddbbf2` and head `53e6d3b0a`, executed in a bounded child process; IL decoded via `System.Reflection.Metadata`. No e2e, no writes to `~/.nuget/packages/gsharp.net.sdk/`, worktree `i2962` untouched.*

Issue #1030, opened 2026-06-24
Interface static state + default-bodied static-virtual interface properties (follow-up to #1019)


DavidObando:
## Follow-up to #1019: interface static *state* and default-bodied static interface properties

Issue #1019 landed **abstract static-virtual interface properties** (`prop Name T { get; }` inside an interface `shared { … }` block, emitted as `Static | Virtual | Abstract` accessor slots + a `Property` row, satisfied by implementers via `MethodImpl`, dispatched through generic constraints). Two narrower static-member sub-features were deliberately deferred and remain rejected with clear diagnostics:

### 1. Interface static *state* (storage) — rejected with **GS0330**

```gsharp
package t
interface ICounter {
  shared {
    var Count int32      // GS0330 — interface static state not supported
  }
}
```

`var` / `let` / `const` fields with actual storage in an interface `shared` block have no CLR per-implementer shape. Implementing this needs either imposed-on-implementer static-field plumbing or a new storage mapping — a separate, larger feature. Today the use case is covered by a static-virtual `func` or a static-virtual `prop` (issue #1019).

### 2. Default-bodied static-virtual interface properties — rejected with **GS0396**

```gsharp
package t
interface IData {
  shared {
    prop Name string { get { return "default" } }   // GS0396 — accessor body not allowed
  }
}
```

Interface properties are abstract-only in the current codebase (instance interface default properties don't exist either). A default-bodied static interface property would need new accessor-body emit plumbing for interface property slots. Today a default can be exposed via a static-virtual `func` with a body.

### Acceptance

- [ ] `shared { var/let/const … }` storage in interfaces, OR an explicit decision to keep it unsupported with refined diagnostics.
- [ ] Default-bodied static-virtual interface properties (`prop X T { get { … } }` on the interface), emitting `Static | Virtual` (non-abstract) accessor slots with bodies.

Refs #1019. See ADR-0089 (`docs/adr/0089-static-virtual-interface-members.md`) "What is intentionally deferred".

PR #2981, opened 2026-08-01
Document GS0506 and require complete diagnostic docs

### LINKED ISSUE #2959: Document GS0506 in the canonical diagnostics reference
## Summary

`docs/diagnostics.md` has no GS0506 section, while `website/docs/ref/diagnostics.md` documents it under “Suspension inside fixed statement”.

This is pre-existing on `main` (`17789519`) and every checked SHA back through `e21677f6`. It is byte-identical with PR #2952 and is not caused by that PR.

## Current state

`docs/diagnostics.md` jumps from GS0504 to GS0508. The website reference includes:

```markdown
## Suspension inside fixed statement (GS0506)

| GS0506 | Error | An `await`, `await for`, `await using`, or `yield` suspension point appears inside a `fixed` statement. |
```

## Test gap

`DiagnosticIdUniquenessTests` parses documented severities but only validates a descriptor when this lookup succeeds:

```csharp
if (documentedSeverities.TryGetValue(descriptor.Id, out var documentedSeverity))
```

A missing documentation entry is therefore invisible.

## Expected

- Add the GS0506 section to `docs/diagnostics.md` so both references agree.
- Tighten the diagnostics documentation test to require every public diagnostic descriptor to have the expected documentation entry, unless explicitly exempted.

### LINKED ISSUE #2970: Keep canonical and website diagnostic references synchronized

DavidObando:
## Summary

- add missing canonical documentation for GS0506, with `Error` severity matching `DiagnosticDescriptors.FixedStatementCannotSuspend`
- document the other 112 descriptor IDs exposed by the tightened completeness check
- require every public or non-public `DiagnosticDescriptors` entry to have a canonical `docs/diagnostics.md` table row with matching severity
- report conflicting duplicate severities by diagnostic ID instead of throwing during dictionary construction

Fixes #2959

## Canonical documentation scope

The old conditional check hid 113 missing canonical entries. All 113 now have rows in `docs/diagnostics.md`; no descriptor is exempted from this canonical-table invariant:

- GS0270, GS0274, GS0275, GS0278–GS0284
- GS0293, GS0295–GS0297, GS0300, GS0301, GS0306–GS0313, GS0316, GS0317
- GS0330–GS0341, GS0353–GS0356, GS0361–GS0367, GS0369–GS0374
- GS0379–GS0381, GS0383–GS0385, GS0392, GS0396–GS0409, GS0413, GS0416, GS0418
- GS0461–GS0463, GS0465, GS0466, GS0468–GS0487, GS0490, GS0496–GS0501
- GS0506, GS9008

Here, “documented” means that the canonical Markdown file contains a table row with a parsed severity. The invariant guarantees presence and severity agreement, not a useful description or example; an empty description still passes.

This PR does **not** close #2970. Its 75 website-reference gaps are untouched, and this check does not inspect the website reference.

Descriptions were checked against each descriptor and its report method/site. GS0396 is documented as retired by issue #1030 rather than as an active error.

GS0506 states that `await`, `await for`, `await using`, and `yield` suspension points are rejected inside `fixed`; a lambda declared there remains valid because it runs in another method.

## Contributor impact

Every future PR that adds a diagnostic descriptor must also add a `docs/diagnostics.md` row with matching severity. A missing row now fails `Every_Descriptor_Is_Documented_With_Matching_Severity`; this mandatory docs-file change is intentional.

## Test behavior

`Every_Descriptor_Is_Documented_With_Matching_Severity` now:

- reflects both public and non-public static descriptor fields
- reports every missing descriptor before checking severity
- reports conflicting duplicate severities with the diagnostic ID

Perturb-observe-restore proof on rebased head:

1. clean control passed 1/1
2. added public undocumented `DocumentationCompletenessProbe` (`GS0507`); test failed with `GS0507 (DiagnosticDescriptors.DocumentationCompletenessProbe)`
3. restored source; test passed 1/1
4. changed one of the two GS9999 rows to `Warning`; test failed cleanly with `GS9999: Warning, Error`
5. restored docs; test passed 1/1

The public-descriptor round trip produced deterministic `GSharp.Core.dll` hashes `ac6e11b5… → c4133208… → ac6e11b5…`. Each successful solution build had non-zero, matching hashes across Core, Compiler, Compiler.Tests, Core.Tests, and Interpreter.Tests.

A reverse docs-only check was not added because the reference intentionally includes reserved/retired IDs and diagnostics emitted outside `DiagnosticDescriptors` (for example GS0190, GS9100, GS9998, and GS9999). Cross-reference synchronization remains tracked by #2970.

## Verification

Base `e80132516` → head `9bcb7d52b`, both from successful CI runs:

- Core.Tests: 7028 passed + 1 skipped → 7028 passed + 1 skipped
- Compiler.Tests: 3931 → 3931
- Interpreter.Tests: 493 → 493
- LanguageServer.Tests: 452 → 452

- full Release solution build: 0 warnings, 0 errors
- descriptor/documentation invariant: 1/1 passed after restore
- final CI rollup: 8 successful, 2 skipped

### DIFF
--- a/docs/diagnostics.md
+++ b/docs/diagnostics.md
@@ -279,7 +279,7 @@ Issue #1655: the IDs below used to collide with earlier, unrelated diagnostics (
 | GS0423 | Error | Type does not implement a usable `GetEnumerator()` method and cannot be iterated with `for ... in`. | `for x in 42 { }` where `42`'s type has no usable `GetEnumerator()`. Previously misfiled under GS0268. |
 | GS0424 | Error | A ref-kind modifier (`ref`/`out`/`in`) is not legal on a primary-constructor parameter. | `class Vec(ref x int32) { }` — primary-constructor parameters materialize fields, and the CLR cannot store a managed pointer in a field (ADR-0060 / ADR-0029). Previously misfiled under GS0241. |
 
-### Pointer / by-ref diagnostics (GS9001–GS9006)
+### Pointer / by-ref diagnostics (GS9001–GS9008)
 
 | ID | Severity | Description | Example trigger |
 |----|----------|-------------|-----------------|
@@ -290,6 +290,7 @@ Issue #1655: the IDs below used to collide with earlier, unrelated diagnostics (
 | GS9005 | Error | Cannot take the address of a constant. | `&myConst` where `myConst` is declared `const`. |
 | GS9006 | Error | Pointer type cannot be a field type. | A struct or class field (including static `shared` fields and top-level globals) declared with a `*T` (managed-pointer) type. |
 | GS9007 | Error | A type may contain at most one `shared` block. | A class or struct with two `shared { ... }` blocks; merge them into one. |
+| GS9008 | Error | A pointer bound by `fixed` cannot be captured by a closure because the closure may outlive the pin. | A lambda inside `fixed p *int32 = xs` captures `p`. |
 
 ### Reference closure diagnostics (GS9100)
 
@@ -894,6 +895,90 @@ expression is too long or complex to compile"). Realistic programs never
 approach the limit; generated code that does should be restructured to reduce
 nesting depth.
 
+## Additional language diagnostics (GS0270–GS0418)
+
+| ID | Severity | Description |
+|----|----------|-------------|
+| GS0270 | Error | The `as` operator targets a non-nullable value type; use a reference type or nullable value type instead. |
+| GS0274 | Error | `nil` is supplied to a non-nullable parameter of an attribute constructor. |
+| GS0275 | Warning | `.GetAwaiter().GetResult()` is called directly on a `ValueTask` or `ValueTask[T]`; convert it to `Task` with `.AsTask()` first. |
+| GS0278 | Error | A convenience initializer does not delegate to another initializer with `init(args)` before any other statement. |
+| GS0279 | Error | A convenience initializer declares `: base(args)` instead of delegating to another initializer with `init(args)`. |
+| GS0280 | Error | An `init(args)` constructor self-delegation appears outside a class constructor body. |
+| GS0281 | Error | A designated initializer delegates to a sibling `init(args)` overload instead of chaining to its base class. |
+| GS0282 | Error | A convenience initializer delegates to itself. |
+| GS0283 | Error | No sibling initializer overload matches an `init(args)` self-delegation call. |
+| GS0284 | Error | An explicit `init(...)` duplicates the constructor synthesized from the class's primary-constructor parameters. |
+| GS0293 | Error | A labeled `break` or `continue` names no enclosing loop. |
+| GS0295 | Warning | A loop label shadows an enclosing loop label of the same name. |
+| GS0296 | Error | The right-hand side of an `if let` or `guard let` binding is not nullable. |
+| GS0297 | Error | A `guard let` else block can fall through instead of unconditionally exiting with `return`, `throw`, `break`, or `continue`. |
+| GS0300 | Warning | A null-conditional index access uses a non-nullable receiver; use ordinary indexing instead. |
+| GS0301 | Error | A null-conditional index access appears on the left-hand side of an assignment. |
+| GS0306 | Error | A declaration uses the removed `type Name <kind>` form instead of putting the aggregate-kind keyword first. |
+| GS0307 | Error | A declaration uses the removed `record` keyword instead of `data class` or `data struct`. |
+| GS0308 | Error | `inline` is applied to an aggregate kind other than `struct`. |
+| GS0309 | Error | `open` is applied to an aggregate kind other than `class`. |
+| GS0310 | Error | `sealed` is applied to an aggregate kind other than `class` or `interface`. |
+| GS0311 | Error | `data` and `inline` are combined on the same declaration. |
+| GS0312 | Error | `open` and `sealed` are combined on the same declaration. |
+| GS0313 | Warning | A switch over a sealed class hierarchy omits a known subtype. |
+| GS0316 | Error | Go-style concurrency syntax is used without importing `Gsharp.Extensions.Go`. |
+| GS0317 | Error | A Go-style built-in such as `len`, `cap`, `append`, or `delete` is used without importing `Gsharp.Extensions.Go`. |
+| GS0330 | Error | An event is declared inside an interface `shared` block; interface static events are not supported. |
+| GS0331 | Error | A class or struct does not implement a required static-virtual interface method that has no default body. |
+| GS0332 | Error | A class or struct declares an instance method where its interface requires a static-virtual method. |
+| GS0333 | Error | A member access through a type parameter names no static-virtual member on any of its interface constraints. |
+| GS0334 | Error | Code outside an interface declaration accesses one of the interface's private helpers. |
+| GS0335 | Error | A private interface method is declared without a body. |
+| GS0336 | Error | An implementer declares a method that clashes with a private helper on one of its interfaces. |
+| GS0337 | Error | `private` is applied to an interface property or event; only private interface methods are supported. |
+| GS0338 | Error | `base[IFoo]` names an interface not implemented by the enclosing type, or appears outside an instance member. |
+| GS0339 | Error | A `base[IFoo]` call names a member that `IFoo` does not declare. |
+| GS0340 | Error | A `base[IFoo]` call targets an abstract interface member with no default body. |
+| GS0341 | Error | A `base[IFoo]` call targets a private interface helper. |
+| GS0353 | Error | A delegate-typed P/Invoke parameter's delegate declaration lacks `@UnmanagedFunctionPointer` with a calling convention. |
+| GS0354 | Error | An unmanaged function-pointer type names an unsupported calling convention. |
+| GS0355 | Error | A P/Invoke declaration returns a managed delegate instead of a raw function pointer or opaque handle. |
+| GS0356 | Error | A raw function-pointer type omits its required calling convention. |
+| GS0361 | Error | A type parameter combines mutually exclusive constraints such as `class` with `struct`, or redundant `struct` with `init()`. |
+| GS0362 | Error | A bare `default` literal appears where no target type can be inferred. |
+| GS0363 | Error | A parameter uses C#'s unsupported `params` keyword instead of G#'s `name ...T` variadic syntax. |
+| GS0364 | Error | A signature declares more than one variadic parameter. |
+| GS0365 | Error | A variadic slot in an anonymous function type does not use the required `[]T` slice storage type. |
+| GS0366 | Error | A type clause uses the removed `map[K]V` spelling instead of `map[K,V]`. |
+| GS0367 | Error | A `yield` appears inside a `try` block that has a `catch` clause. |
+| GS0369 | Error | A collection initializer targets a type with no accessible `Add` method or settable indexer. |
+| GS0370 | Error | An indexer declaration has no index parameters. |
+| GS0371 | Error | An indexer declaration has no `get` or `set` accessor body. |
+| GS0372 | Error | An init-only property is assigned outside a constructor, object initializer, or `init` accessor for the same instance. |
+| GS0373 | Error | A property declares both `set` and `init` accessors. |
+| GS0374 | Error | A static property declares an `init` accessor. |
+| GS0379 | Error | Code outside a declaring or derived type accesses a protected member. |
+| GS0380 | Error | `protected` is used outside an inheritable `open class`. |
+| GS0381 | Error | A class participates in a direct or transitive inheritance cycle. |
+| GS0383 | Error | A `base` call appears outside an instance member of a class with a base class. |
+| GS0384 | Error | A `base` call names no accessible member on the base-class chain. |
+| GS0385 | Error | A bracketed `base[Type]` selector names a type that is not a base class of the enclosing type. |
+| GS0392 | Error | A range expression slices a type that has no supported array, slice, string, span-like, or `System.Range` indexer shape. |
+| GS0396 | Error | Retired by issue #1030. It formerly rejected default-bodied static-virtual interface properties, which are now supported. |
+| GS0397 | Error | A class or struct omits a compatible static property required by a static-virtual interface property slot. |
+| GS0398 | Error | An unmanaged pointer has a pointee type that is not supported as unmanaged or blittable storage. |
+| GS0399 | Error | A `stackalloc` expression uses a managed or otherwise non-blittable element type. |
+| GS0400 | Error | A `fixed` statement appears outside an `unsafe` context. |
+| GS0401 | Error | A `fixed` statement's source is not a pinnable array, string, or span-like value, or its element type does not match the pointer. |
+| GS0402 | Error | `++` or `--` targets something other than a writable variable, field, array element, or indexer. |
+| GS0403 | Error | A `*void` pointer is dereferenced, indexed, or used in pointer arithmetic before being cast to a typed pointer. |
+| GS0404 | Error | A managed function-pointer type appears outside an `unsafe` context. |
+| GS0405 | Error | A method address does not match the target function-pointer signature, or the operand is not one static method group. |
+| GS0406 | Error | A fixed-size buffer field appears outside an `unsafe` context. |
+| GS0407 | Error | A fixed-size buffer field does not use a fixed-length array type. |
+| GS0408 | Error | A fixed-size buffer field has a non-positive length. |
+| GS0409 | Error | A fixed-size buffer field uses an unsupported element type. |
+| GS0413 | Error | A `base` call targets an abstract member that has no implementation to invoke. |
+| GS0416 | Error | A list pattern contains more than one slice (`..`) subpattern. |
+| GS0418 | Error | A `using` or `await using` statement uses tuple or named deconstruction instead of one variable declaration. |
+
 ## Defer with by-reference arguments (GS0460)
 
 | ID | Severity | Description |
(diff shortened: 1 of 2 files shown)


DavidObando:

Specifically:

DavidObando:
## Review: **MERGEABLE** — no blocking issues. Both halves verified true against the code.

Reviewed at `13a3b766db0f70cb356ecf3c0bec7d9cb6b4f924`, confirmed equal to `refs/pull/2981/head`, in a scratch clone checked out at that exact SHA. Natural base `8dddbbf285c2b4e94d06647671b44f9b637537f7`; content differential non-empty (`docs/diagnostics.md` 136±, `DiagnosticIdUniquenessTests.cs` 31±).

This is a docs-shaped PR, so a behavioural differential does not exist for its prose half in principle — that half was verified by tracing each claim to the code it describes. The mechanism half *is* measurable and was measured.

### Evidence class, stated plainly

I did **not** execute the xunit test. Load was 42.6 with 89 `dotnet` processes, far above the machine ceiling, and a Core.Tests build would have added to it for evidence I could obtain otherwise. Instead I replicated the check's exact logic outside the build — the same field selector (`internal static readonly DiagnosticDescriptor … = new("GSnnnn", DiagnosticSeverity.X`) and the same documentation regex (`^\|\s*(GS\d{4})\s*\|\s*(?:\*\*)?(Error|Warning|Info)`) — and ran it against both trees, plus a perturb-observe-restore round trip. The PR body's own five-step mutation proof is consistent with everything I measured. Anyone wanting the real xunit execution should treat that as still outstanding; nothing I found suggests it will differ.

### The central claim is true

| tree | descriptors | documented rows | **missing** | severity mismatches |
|---|---:|---:|---:|---:|
| base `8dddbbf2` | 375 | 266 | **113** | 0 |
| head `13a3b766d` | 375 | 379 | **0** | 0 |

113 → 0, with no exemption list and no conditional skip. The count matches the PR body exactly, and matches the independent measurement in #2970.

**Perturb-observe-restore on the completeness logic** (`X → mutant → X`):

```
X        : 375 descriptors, MISSING 0
mutant   : 376 descriptors, MISSING 1  → GS0507
X restore: 375 descriptors, MISSING 0
md5 e32494d279e21495b8555384f27971d5 -> e32494d279e21495b8555384f27971d5   (restore verified, tree clean)
```

The tightening is real and worth naming precisely, because it is the same failure shape the codebase keeps hitting: the old body was `if (documentedSeverities.TryGetValue(descriptor.Id, out var s)) { Assert.Equal(...); validatedCount++; }`, guarded by `Assert.True(validatedCount > 0)`. That check **validated only what was already documented** and reported success while 113 descriptors were invisible to it. A check whose pass condition is "at least one thing I could check was fine" is not a check. Replacing it with an unconditional zero-missing assertion is the right correction.

### Half 1 — the `GS0506` prose is accurate

Traced to `src/Core/CodeAnalysis/Binding/StatementBinder.Blocks.cs:912` `ReportSuspensionPointsInFixedBody`:

- The doc row says the diagnostic fires for `await`, `await for`, `await using`, or `yield` inside `fixed`. The switch matches `AwaitExpressionSyntax or AwaitForRangeStatementSyntax or AwaitUsingStatementSyntax` (`:924`) and `YieldStatementSyntax` (`:926`). **All four, exactly, with nothing else.** True.
- The prose says *"Suspension in a lambda declared inside `fixed` remains valid because the lambda runs in a different method."* The walker early-returns on `FunctionLiteralExpressionSyntax or LambdaExpressionSyntax` (`:915-917`) and therefore never descends into one. **True**, and the stated reason is the actual reason.
- Severity `Error` matches `DiagnosticDescriptors.FixedStatementCannotSuspend` (`DiagnosticDescriptors.cs:378`). True.

**The `GS0396` "retired" claim is also true**, and it was the highest-risk single sentence in the diff since it asserts absence rather than presence. `ReportDefaultStaticInterfacePropertyNotSupported` (`DiagnosticBag.Reports.Declarations.cs:731`) has **zero callers** — the only occurrences of the descriptor field anywhere in `src/` are its declaration and that one report method — and `Binder.cs:1509`/`:1986` now *bind* default bodies on static-virtual interface properties per issue #1030. The rejection is genuinely unreachable, so documenting it as retired is correct rather than misleading.

The single deleted line is the heading retitle `### Pointer / by-ref diagnostics (GS9001–GS9006)` → `(GS9001–GS9008)`, which is correct: `GS9007` already existed in that table but outside the advertised range, and `GS9008` is added here.

I spot-checked the prose claims above rather than all 113 new descriptions; the ones examined were accurate against their descriptors and report sites.

---

## Non-blocking issues

**N1 — the verification table is measured on a stale base, and does not say so.** The body reports base `8dddbbf2` as Core 7028+1 / Compiler 3893 / Interpreter 489 / LSP 452 and states HEAD is unchanged. Current `main` is `a6b1561f2`, where the executed counts are Core 7028+1 / Compiler **3906** / Interpreter **491** / LSP 452. For a change that is documentation plus one renamed test the substantive risk is genuinely low, but the table as written reads as current and will be compared against post-merge numbers by someone who does not know the base. *Fix:* either re-measure on `a6b1561f2`, or label the table with its base in the table itself rather than only in the sentence above it.

**N2 — #2970 is mentioned but not declared, and only half of it is addressed.** The body says *"Cross-reference synchronization is tracked by #2970."* #2970's measurement was 113 missing canonical / **75 missing website** / 54 in neither. This PR closes the canonical 113. It does not touch the website, and the new check does not cover website docs. That split is a reasonable scope decision, but the combination of a `#2970` reference with *"All 113 are now documented; none are exempted"* invites the conclusion that #2970 is covered. *Fix:* one sentence — "this closes the canonical-doc half only; the 75 website gaps and the reverse direction remain open in #2970." Do **not** add a `Closes` line; the issue is genuinely not closed.

**N3 — the new check can be silently bypassed by one accessibility keyword.** `GetDescriptorFields()` uses `BindingFlags.Static | BindingFlags.NonPublic` with no `BindingFlags.Public`. All 375 descriptors are `internal` today, so the census is complete *at this commit* — but a descriptor declared `public` would be invisible to the completeness assertion and would document nothing while the test stayed green. That is precisely the failure shape this PR exists to eliminate, reappearing one level up in the selector rather than in the loop body. *Fix:* add `| BindingFlags.Public` — one token, and it makes the guarantee structural instead of conventional.

**N4 — `Distinct().Single()` turns a future duplicate into a crash, and there is already a duplicated ID.** `GS9999` is documented **twice**: `docs/diagnostics.md:308` ("An unexpected exception was caught by the evaluator") and `:1138` ("Fatal I/O error"). Both are `Error`, so `group.Select(...).Distinct().Single()` collapses cleanly and the check passes today. If anyone edits one row's severity and not the other, `Single()` throws `InvalidOperationException` during dictionary construction — before any assertion runs — producing "Sequence contains more than one element" instead of a diagnostic-shaped failure. This becomes reachable exactly when the pending `GS9999` rewording lands (see C2). *Fix:* replace `.Single()` with a first-value-plus-explicit-assert, so a contradictory duplicate fails as a readable test failure.

## Coordination — two live interactions the author should know about

**C1 — this PR makes `docs/diagnostics.md` a required file for every future diagnostic, starting with the G5 PR.** `GS0505` and `GS0507` are currently absent from both `DiagnosticDescriptors.cs` and `docs/diagnostics.md` (verified — no collision with this PR's content). The G5 work reserving those two IDs will now **fail this check** unless it also adds documentation rows. That is the mechanism working as intended, but it is a new obligation that the G5 author has probably not planned for, and it is better raised now than discovered as a red build. Both PRs also edit `docs/diagnostics.md`; this PR's hunks sit at 282, 293, 898+, 1024+ and 1102+, and a new `GS0505`/`GS0507` section would naturally land near the `GS0506`/`GS0508` entries at 1106–1118 — adjacent to the last hunk, so a textual conflict is plausible even though the rows differ.

**C2 — no conflict with the pending `GS9999` rewording, but it must touch both rows.** This PR does not modify either `GS9999` row. Whoever reworks `GS9999` away from "unexpected exception / internal error" phrasing should know there are **two** rows (`:308` and `:1138`) describing materially different conditions under one ID, and that N4 above makes editing only one of them fail noisily.

## Suggestion

**S1 — "documented" is defined as "has a table row with a severity".** The regex captures only ID and severity, so `| GS0999 | Error |  |` satisfies completeness. The 113 new rows use a three-column format (`ID | Severity | Description`) while the pre-existing tables are four-column with an `Example trigger`. Not a correctness problem and not worth blocking or reformatting now — but the check guarantees presence, not usefulness, and it is worth being explicit about that so the invariant is not later cited as proof the docs are good.

---

**Verdict: MERGEABLE.** The mechanism does what it claims, the 113-entry gap is genuinely closed with no exemptions, and every prose claim I traced — including the two that assert absence, `GS0396` retirement and the lambda exemption — is true against the code. N1–N4 are all one-or-two-line changes and none of them makes the PR wrong; N3 is the one I would most like to see applied, since it closes the same class of silent hole the PR is fixing.

DavidObando:
**Rebase hazard, and a correction to the "only GS0505/GS0506 are undocumented" figure.**

I enumerated descriptor fields and applied this PR's own regex (`^\|\s*(GS\d{4})\s*\|\s*(?:\*\*)?(Error|Warning|Info)`) to `docs/diagnostics.md`, on `main` `5e616b9bb` and on this branch.

**On `main` today: 376 descriptors, 267 documented, 113 undocumented.** Not two. The undocumented set is broad — `GS0270`–`GS0499` in large runs, plus `GS0500`, `GS0501`, `GS0506`, `GS9008`. This makes the case for the completeness check stronger, not weaker: the gap is a third of all descriptors.

It also corrects a figure circulating in the fleet. `GS0505` **is** documented on `main` (`docs/diagnostics.md:984`, severity `Error`) — it merged intact with #2983. The claim that `GS0505` and `GS0506` both lack rows is half wrong; only `GS0506` does. The *conclusion* that rows are not yet mandatory is right, and `DiagnosticIdUniquenessTests.cs:145` using `TryGetValue` is why, but the supporting example should be `GS0506` alone.

**This branch is internally consistent:** 375 descriptors, 379 documented, **0 undocumented** — `Every_Descriptor_Is_Documented_With_Matching_Severity` passes here, and no ID has conflicting severities, so the `.Distinct().Single()` cannot throw. (379 > 375 means ~4 rows document IDs that are no longer descriptors; harmless, since the assertion runs descriptor → docs only.)

**The hazard is the rebase, and it is specific.** This branch predates #2983, so `GS0505` exists here as **neither descriptor nor row**. Rebasing onto `5e616b9bb` brings in both — but the row arrives through `docs/diagnostics.md`, which is exactly the file `git merge-tree` reports as the conflict between these two PRs, and this PR rewrites large regions of it. **If the conflict is resolved in favour of this branch's version of that region, the `GS0505` row is dropped while the descriptor remains, and this PR's own new check fails on `GS0505`** — the one ID that was not in scope when the check was written.

Cheap post-rebase gate, no test host needed:

```
grep -nE '^\|\s*GS0505\s*\|\s*Error' docs/diagnostics.md
```

Better, re-run the full enumeration after rebase rather than spot-checking — the same failure mode applies to any ID that lands between now and merge.

Two notes for whoever resolves it: `GS0505`'s row currently sits immediately above the `GS0508` section, in the same 963–984 anchor region several PRs are targeting, so appending new sections **below** the last `GS05xx` section avoids compounding this. And `GS0507` is quarantined — it reads as free by tree enumeration but was allocated and withdrawn on an unpushed branch, so it should not be backfilled as though it were a gap.

DavidObando:
## Verification of the rebase — **MERGEABLE**

Verified at `9bcb7d52b839129aba0faedc76e9a66e524c10c9`, computed merge-base with `main` (`a1237ded7`) = `e80132516` — the #2983 merge commit, so `GS0505` is present in this branch's base as both descriptor and row. `git merge-tree` against `main`: **0 conflicts**. Three-dot diffs throughout; net state read at the tip.

Read-only verification: `git` and `grep` only, no build, no test host. Enumeration replicates the assertion under test exactly — same regex, same descriptor set, same direction.

### 1. The predicted failure did **not** materialise

My earlier concern was that resolving `docs/diagnostics.md` in this branch's favour would drop the `GS0505` row while keeping the descriptor, failing this PR's own new check on the one ID that was out of scope when it was written.

```
GS0505  descriptor = Error     docs rows = ['Error']     ✅
```

Row survives, severity agrees, exactly one row. Gate passes.

### 2. The triple — **376 / 380 / 0**, matching expectation exactly

| | count |
|---|---|
| descriptors (`new("GS####", DiagnosticSeverity.…`) | **376** |
| documented IDs (PR's own regex) | **380** |
| **undocumented** | **0** |

Zero severity mismatches between descriptor and docs. `Every_Descriptor_Is_Documented_With_Matching_Severity` passes on the rebased tree.

### 3. Duplicate rows — **better than briefed: the fragility is fixed, not merely latent**

I was asked to flag `.Distinct().Single()` as fragile. It is no longer present. The rebase replaced it:

```csharp
var conflictingSeverities = documentedSeverityGroups… .Where(item => item.Severities.Length > 1) …
Assert.True(conflictingSeverities.Length == 0, "Diagnostics documented with conflicting severities:\n" + …);
var documentedSeverities = documentedSeverityGroups.ToDictionary(g => g.Key, g => g.First().Groups[2].Value, …);
```

Dictionary construction now uses `group.First()` and **cannot throw**; a divergent severity produces a named, actionable assertion instead of an `InvalidOperationException` that identifies nothing. That was the sharper half of the hazard and it is closed.

Two corrections to the briefed detail, both now moot but worth the record:

- **`GS9998` also has two rows**, not just `GS9999`. Both IDs, all rows `Error` today, so nothing throws.
- **Neither `GS9998` nor `GS9999` has a descriptor field at all.** Under the old code that made them *more* dangerous, not less: `.Single()` ran during dictionary construction over **documented** IDs, so an ID with no descriptor could still red the whole test. The framing "it passes because both rows are `Error` today" was right by accident, not by mechanism.

### 4. Placement — **instruction not followed, and it is non-blocking**

The new `GS0506` section was inserted **between** the `GS0505` and `GS0508` sections, in numeric order, inside the contested anchor region — not appended below the `GS0508` block as directed. Final hunk is `@@ -986,6 +1108,18 @@`, placing it immediately above `## Unconstrained nullable sequence element (GS0508)` at `:1123`.

This does not affect correctness: the check is set-membership and placement is free, the triple is `376/380/0`, and there are zero conflicts with `main` today. It is also defensible editorially — a reader looking for `GS0506` finds it where it belongs.

The cost is scheduling only: the anchor other PRs target is now re-occupied by 12 more lines, so the next PR inserting between `GS0505` and `GS0508` will conflict with this one. Worth knowing when ordering the remaining docs-touching PRs; not worth another rebase.

### 5. N3 applied — and it closed a latent hole, not a live one

`DiagnosticIdUniquenessTests.cs:199`:

```
-  .GetFields(BindingFlags.Static | BindingFlags.NonPublic)
+  .GetFields(BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic)
```

**Descriptor count is still 376.** I verified the reason at source rather than inferring it from the stable count: **376 of 376 descriptor fields are `internal`; zero are `public`.** So no live public descriptors were discovered — the change prevents a future `public` descriptor from being silently invisible to the very test that exists to catch it. Correct fix, no finding behind it.

### Scope and compile risk

Two files, **zero production code**: `docs/diagnostics.md` (+136) and `test/Core.Tests/CodeAnalysis/DiagnosticIdUniquenessTests.cs` (+52/−15). `DiagnosticDescriptors.cs` is untouched, which is why the PR's descriptor set is identical to `main`'s.

`GS05xx` map is byte-for-byte identical between `main` and this branch — `GS0500`–`GS0506`, `GS0508`. **`GS0507` is absent as descriptor, absent as docs row, absent everywhere**, consistent with its quarantine.

Per the throttle I did not build. The branch is two commits behind `main` (`5e616b9bb`, `a1237ded7`), neither of which adds a descriptor — verified, not assumed. Since this PR touches no production code and its only compiled change is a test using reflection plus file I/O, the textual-vs-compile mergeability gap is unusually narrow here. **I do not think a build is required**; if certainty is wanted it is a cheap candidate for the serialised build gate.

### One new finding, relevant to ID allocation

**Four documented IDs have no descriptor: `GS0190`, `GS9100`, `GS9998`, `GS9999`.**

These are harmless to the check, which runs descriptor → docs only. But they are the **inverse of the `GS0507` trap**: an ID that reads as *taken* to anyone grepping the docs and as *free* to anyone enumerating descriptor fields. Given that the allocation rule was just tightened to "enumerate fields, and `git log -S`", this is the failure mode that rule does not cover — enumeration alone would report all four as available. Worth deciding whether they are retired-and-reserved or simply stale rows before anyone reaches `GS9100`.

---

**Verdict: MERGEABLE.** All five items check out; items 3 and 5 came back stronger than the brief anticipated. The only deviation is placement (item 4), which is cosmetic against a set-membership check and costs a future rebase rather than correctness.

DavidObando:
Rebased onto `e80132516` and addressed N1–N4/S1 in `9bcb7d52b`:

- resolved the diagnostics conflict with both GS0505 and GS0506 retained
- included public descriptor fields in the completeness census
- made conflicting duplicate severities fail with the diagnostic ID
- clarified #2970/website scope, the presence-only guarantee, current-base counts, and the mandatory docs-row impact for future descriptors
- revalidated public GS0507 and conflicting GS9999 perturbations in both directions

Base/head CI counts match; final rollup is 8 SUCCESS + 2 SKIPPED and the PR is mergeable.

PR #2983, opened 2026-08-01
Improve diagnostics for unconstrained nullable iterator calls

### LINKED ISSUE #2958: Improve diagnostics for unconstrained calls to specialized nullable iterators
## Summary

Calls to an unconstrained nullable generic iterator from another unconstrained generic context report diagnostics whose suggested remedies do not work.

This is a follow-up to #2952, which correctly specializes an iterator such as `opt[T]` into compiler-generated `T : class` and `T : struct` variants.

## Repro

```gs
package Repro

func opt[T](v T) sequence[T?] {
    yield v
    yield nil
}

func relay[U](u U) {
    let values = opt(u)
}
```

Current result:

```
error GS0266: Call to 'opt' is ambiguous between multiple overloads. Disambiguate with explicit types or named arguments.
```

Both stated remedies fail with the same diagnostic:

```gs
let values = opt(v: u)
let values = opt[U](u)
```

The remedy that works is constraining the enclosing type parameter:

```gs
func relay[U struct](u U) { let values = opt(u) }
func relay[U class](u U) { let values = opt(u) }
```

A related call reports a synthetic constraint the user never declared:

```gs
let value int32? = 1
let values = opt[int32?](2)
```

```
error GS0152: Type argument 'int32?' for type parameter 'T' does not satisfy the 'struct' constraint.
```

## Root cause

G# rejects source overloads differentiated only by generic constraints (`func f[T struct]` plus `func f[T class]`) with GS0264 on both `17789519` and #2952. Nullable iterator specialization synthesizes the equivalent CLR variants internally. An unconstrained enclosing type parameter cannot select either variant, but ordinary overload diagnostics expose that implementation detail.

## Expected

Detect this specialized overload set and point at the unconstrained enclosing type parameter. Suggest adding a `struct`, `class`, or base-class constraint. Do not suggest named or explicit arguments when neither can resolve the call, and do not present compiler-generated constraints as source constraints.

DavidObando:
## Summary
- report GS0505 when an unconstrained caller type parameter leaves compiler-generated nullable iterator CLR variants ambiguous
- keep GS0266 for user-written, mixed, and multi-declaration ambiguities
- cover free, named, explicit, instance, sync, and async calls
- document GS0505

No changes to `ExpressionBinder.Calls.Invocation.cs`.

## Verification
- `dotnet build GSharp.sln -c Release -m:1 -warnaserror`: 0 warnings, 0 errors
- nullable iterator Compiler test class: 24 passed
- diagnostic ID tests: 4 passed
- base/head pins: 2 fail on `a6b1561f2`, 2 pass on HEAD
- base/head guards: 3 pass on both
- semantic mutation probes: 10/10 detected; restored control passed after each
- Compiler/Core test DLL copies matched and were nonzero

Targeted tests only, per machine-load constraint.

Closes #2958


DavidObando:
## Review: PR #2983 — verdict **MERGEABLE**, with one scheduling flag that needs an answer before merge

Reviewed at `ee3842c810bec3634a65f33e9801c3b99320f12a`, asserted equal to `refs/pull/2983/head`; merge-base with `main` = `a6b1561f`. Scratch clone, `gsc` built at HEAD with `-m:1`; base `gsc` at `a6b1561f` from a separate built clone. Core `GSharp.Core.dll` md5 base `3f97f3f3…` vs head `26e367ba…` — distinct, so the two compilers are genuinely different binaries.

All findings below rest on my own base→head probes, not on the PR body.

---

### Q1 — Was the old behaviour wrong, or merely unhelpful? **Neither: it was actively misleading. And no program changes accept/reject status.**

Nine-program differential, base `a6b1561f` vs head `ee3842c8`, same source, same compiler flags:

| program | base | head | |
|---|---|---|---|
| inferred `opt(u)` | GS0266 | **GS0505** | flip |
| named `opt(v: u)` | GS0266 | **GS0505** | flip |
| explicit (new test shape) | GS0266 | **GS0505** | flip |
| explicit (old test shape) | GS0266 | **GS0505** | flip |
| async sequence | GS0266 | **GS0505** | flip |
| instance `box.opt(u)` | GS0266 | **GS0505** | flip |
| user-written ambiguity | GS0266 | GS0266 | held |
| mixed candidate set | GS0266 | GS0266 | held |
| multiple declarations | GS0266 | GS0266 | held |

Positive controls — `[T struct]`, `[T class]`, and a concrete-type call — all compile with **exit 0 and zero diagnostics on both** base and head.

**No program's accept/reject status changes.** Every erroring program still errors; every clean program stays clean. Only the ID and message move. This is backed structurally, not just empirically: both edits sit strictly inside `else if (ambiguous)` (`OverloadResolver.Candidates.cs:81`) and `else if (overloadAmbiguous)` (`OverloadResolver.CallBinding.cs:1119`), branches that already reported unconditionally, and the new helper is pure — it only reads `boundArguments[i]?.Type` and emits no diagnostics of its own.

**So this is not the PR #2980 shape.** No semantic change is hiding under a diagnostics title, and no test pin or PR-body disclosure is owed on that axis.

One thing the PR body undersells, and should say, because it is the actual justification: **the old diagnostic gave advice that does not work.** GS0266 reads *"Disambiguate with explicit types or named arguments."* My table shows both remedies fail on base — `opt(v: u)` (named arguments) and `opt[U](u)` (explicit types) each still produce GS0266. The old message sent users down two dead ends. That is a stronger case than "unhelpful message" and worth one line in the summary.

### Q2 — Reachability, both directions: **new one reachable, old one still live and correctly so, and neither edit is dead code**

Both directions hold (rows 1–6 vs 7–9 above). To check the two edits are *individually* live rather than one shadowing the other, I disabled only the `Candidates.cs` branch (`if (false && TryGet…)`), rebuilt, and re-ran:

| probe | with `Candidates.cs` branch disabled | attributed to |
|---|---|---|
| inferred / named / explicit×2 / async | still GS0505 | `CallBinding.cs:1134` |
| instance `box.opt(u)` | reverts to **GS0266** | `Candidates.cs:98` |

Round-trip: Core md5 `26e367ba…` → mutant `87c311f7…` → restored `26e367ba…`, `git status` clean. **Both sites are live**; the free-function path routes through `CallBinding`, and instance-method calls are the sole traffic through `Candidates`. No dead code in the +16/+81.

The other two `ReportAmbiguousOverloadResolution` call sites — `OverloadResolver.Constructors.cs:814` and `:1351` — are constructor and `init` resolution. A constructor cannot return `sequence[T?]`, so they structurally cannot reach this case and are **correctly left untouched**. The candidate set is otherwise tightly gated (shared `Declaration`, sequence-of-nullable-of-type-parameter return, caller type resolving to a type parameter, and both Reference and Value specializations present), so false positives are unlikely.

### Q3 — Diagnostic ID: **`GS0505` is free in `main`, but this is a live collision risk with G5** ⚠️

`GS05xx` occupancy in `main`: `GS0500`–`GS0504`, `GS0506`, `GS0508` taken. **`GS0505` and `GS0507` are the only two free slots** — which is exactly why both this PR and the G5 boundary PR landed on them. No *open* PR currently claims `GS0505` (checked all 19), so G5 is unlanded.

**If G5 takes `GS0505` as planned, the two collide.** #2983 is open, reviewed and `CLEAN`, so the cheaper move is for G5 to take `GS0507` plus the next free slot. This needs an explicit decision, not a merge-order gamble — two PRs registering the same ID will not conflict textually if they insert on different lines, so `DiagnosticIdUniquenessTests` would be the only thing catching it post-merge.

**Interaction with PR #2981 — compatible, but there is a textual conflict.** #2981's stricter `Every_Descriptor_Is_Documented_With_Matching_Severity` requires every descriptor to have a docs row. #2983 adds the descriptor **and** the docs row atomically, and I verified #2981's regex `^\|\s*(GS\d{4})\s*\|\s*(?:\*\*)?(Error|Warning|Info)` matches this PR's row `| GS0505 | Error | … |`. **Either merge order passes the check.** However `git merge-tree pr2983 pr2981` reports `CONFLICT (content): Merge conflict in docs/diagnostics.md` — both insert prose additively, so resolution is mechanical, but whoever lands second owes a rebase.

### Q4 — `gsi` visibility: **no dependency on the G5 renderer**

Actual emitted text:

> `GS0505: Call to nullable iterator 'opt' cannot select a CLR representation because type parameter 'U' is not constrained to 'struct' or a reference type. Add a 'struct', 'class', or base-class constraint to 'U'.`

The message names the function, names the type parameter, and states the remedy. It carries no deictic reference to a span, so under `src/Repl/Program.cs:60`'s bare `Console.Error.WriteLine($"{d.Id}: {d.Message}")` it renders complete and actionable. **#2983 does not need the G5 renderer to land first.**

This is in fact a strict improvement for `gsi` specifically: the old GS0266 text was both location-dependent *and* wrong, which is the worst combination in a renderer that shows no location.

### Q5 — Contention: **clear of every reserved region except `docs/diagnostics.md`**

Files touched: `docs/diagnostics.md`, `Binding/OverloadResolver.CallBinding.cs`, `Binding/OverloadResolver.Candidates.cs`, `DiagnosticBag.Reports.Expressions.cs`, `DiagnosticDescriptors.cs`, `test/Compiler.Tests/Emit/Issue2927NullableSequenceElementTypeTests.cs`.

Zero overlap with `Evaluator.Expressions.cs` (#3000, G2, G4), `Evaluator.Statements.cs` (#2972, G5 ADR), `Emit/MethodBodyEmitter.Statements.cs` (#2967/#2964), or `Lowering/Iterators/IteratorRewriter.cs` (#3000, #2975). I checked the file lists of #2972, #2975, #2981, #3000, #2967, #2964 and #2980 against this PR's production files: **only #2981 overlaps, and only on `docs/diagnostics.md`.** `DiagnosticDescriptors.cs` is a latent surface against G5's new IDs, per Q3.

---

## Non-blocking

**1. The rewritten `ExplicitSource` drops a case that used to be covered.** The PR changed the test *source*, not only its expectation:

```
- func opt[T](v T) sequence[T?] { yield v; yield nil }
- func relay[U](u U) { let values = opt[U](u) }
+ func opt[T](v int32) sequence[T?] { yield nil }
+ func relay[U]() { let values = opt[U](1) }
```

I probed the original shape at HEAD specifically to check whether it was replaced because it failed to produce GS0505 — **it was not; the original also yields GS0505.** So nothing is being hidden. But the new shape makes the parameter non-generic, so "explicit type arguments *plus* a generic parameter that also drives inference" is now pinned by no test, and that is the combination users actually write. Keeping both sources costs three lines.

**2. `Candidates.cs:98` has exactly one test covering it.** Per the attribution above, `UnconstrainedInstanceNullableIteratorCallReportsGS0505` is the sole cover for that branch. Fine as-is, worth knowing before anyone prunes tests.

**3. Full suites not run** ("targeted tests only, per machine-load constraint"). The verification table is internally consistent — "2 pins" matches the two GS0505 test *methods*, "3 guards" matches the three GS0266 tests, and 10/10 mutants detected means no survivor round-trip is owed. Given the change is error-path-only and my accept/reject probes are clean, residual risk is low, but CI should confirm against the `Compiler` baseline of 3906.

## Blocking

None on correctness. The only item needing a decision before merge is the **`GS0505` assignment versus the G5 boundary PR** — that is a scheduling call, not a defect in this PR.

Issue #2843, opened 2026-07-28
gsc: return inside a fixed block does not satisfy definite-return analysis (GS0100)


DavidObando:
Found by the Oahu migration, cycle 18. **Non-gating** — zero sites in Oahu; found while reducing #2838.

## Summary

gsc's definite-return analysis does not model a `fixed` block's body as capable of terminating the method, so a `return` that is the sole exit still reports `GS0100`.

## Repro (verified)

Passes:

```
func F(xs []int32) int32 {
    unsafe { return 42 }
}
```

Fails:

```
func F(xs []int32) int32 {
    unsafe {
        fixed p *int32 = xs {
            return xs.Length
        }
    }
}
```

```
error GS0100: Not all code paths return a value.
```

Artifacts: `oahu-mig-c18/repro4/gs0100.gs`, `gs0100b.gs`.

## Expected

A `return` inside a `fixed` body counts toward definite return, exactly as it does inside a plain `unsafe` block. `fixed` introduces a pinned scope, not a new control-flow exit.

## Impact

Low. No Oahu site hits it, and the workaround (hoist the value to a local and return after the `fixed` block) is trivial. Filed for completeness so the flow-analysis gap is recorded.

## Suggested regression coverage

- Definite-return tests for `return` as the sole exit inside `fixed`, nested `fixed`, and `fixed` inside `if`/`else` where both branches return.
- A negative control where a `fixed` body genuinely does not return on all paths, so GS0100 still fires.

Issue #2662, opened 2026-07-21
gsc: async iterator emit crashes when a predicate captures its own parameter


DavidObando:
Cycle 7 Oahu.Cli.App ICE: InvalidOperationException Variable 'u' has no local slot or parameter index in current method, at JobScheduler ObserveAsync passing u -> u.JobId == jobId into async IAsyncEnumerable Drain. Fix generalized capture/slot allocation; exact proof plus runtime/ILVerify tests.

DavidObando:
Cycle 8 regression in #2686-authored AsyncIteratorMoveNextBodyBuilder.RewriteTryStatement: try with await entries but default yieldEntries iterates an uninitialized ImmutableArray and NREs. Handle awaits-only and yields-only try dispatch symmetrically; exact AudibleJobExecutor ExecuteAsync proof.

DavidObando:
Cycle 9 residual: #2697 stops async-iterator try-dispatch NRE, but ExecuteAsync and ReadAllAsync MoveNext now fail ILVerify BranchIntoTry (2ff82b6ad30d, 8aff8d9048f7). Fix state-machine SEH branch topology; exact methods must verify.

PR #2997, opened 2026-08-01
Funnel conditional gotos through lifted finally handlers

### LINKED ISSUE #2942: Handle conditional gotos escaping rewritten finally handlers
## Problem

`FinallyExitRewriter` detects both `BoundGotoStatement` and `BoundConditionalGotoStatement` targets that escape a `finally` body, but `ExitFunneler` overrides only `RewriteGotoStatement`.

If lowering ever leaves an escaping conditional goto in a finally handler, the rewriter will lift the handler but will not funnel that branch. The remaining branch can violate CLR exception-handler control-flow rules.

No current source program was found that triggers the shape; current lowering appears to express escaping source transfers as unconditional gotos.

## Expected work

Either:

1. implement `RewriteConditionalGotoStatement` with the same discriminator/tail funnel while preserving condition evaluation and polarity, or
2. prove the shape cannot reach this pass and narrow the detector/invariant accordingly.

Add a bound-tree or source-level regression test that fails if an escaping conditional branch survives the rewrite.

Found during review of PR #2926.

DavidObando:
Fixes #2942

## Chosen fork

Implemented `FinallyExitRewriter.ExitFunneler.RewriteConditionalGotoStatement`.

The rewrite mirrors the established protected-region and async-finally funnels:

1. branch to a local skip label on the inverse of `JumpIfTrue`;
2. assign the target discriminator;
3. jump to the lifted finally tail;
4. dispatch to the original target after the finally body.

This evaluates the original condition exactly once and preserves both true and false polarity.

## Producer proof

Every `BoundConditionalGotoStatement` construction site was classified:

- statement binder loop conditions (`while`, `for`, `do`): target labels remain inside the same lowered loop;
- ordinary lowerer (`if`, ranges, `foreach`, async-foreach): targets are local body/end labels in the same generated block;
- interpolated-string handler gating: local append-end labels, introduced after this pass;
- protected-region/finally rewriters: internal routing, rethrow, and dispatch labels;
- ordinary iterator builder: it assembles dispatch before its final `Lowerer.Lower`; the outer dispatch is outside the user try, while inner dispatch labels stay inside that same try, so `ContainsLabel` makes the funnel decline;
- async and async-iterator builders/spillers: ordinary body lowering, including `FinallyExitRewriter`, runs before state-machine assembly, so it never sees generated async dispatch; generated async exception handling has its own equivalent conditional funnel;
- `StateMachineEmitter`: emit-time state dispatch;
- `ControlFlowGraph`: analysis-only synthetic branches;
- `BoundTreeRewriter`: copies existing nodes rather than producing a new target relationship.

Source attempts covered goto from finally inside loops, nested try/finally, an enclosing switch arm, and break/continue/labeled goto combinations. Their lowered trees use an inverse conditional jump to a local skip label plus an unconditional escaping goto. Disabling the new conditional funnel changed none of their outputs. No source-level producer was found, so the regression constructs the sanctioned bound-tree shape directly.

## Regression and runtime evidence

The direct tree contains:

- an internal conditional target, proving it must not be funnelled;
- an escaping conditional target in the protected try;
- a conditional escape in the finally, forcing the handler lift;
- side-effecting conditions for taken/not-taken true and false polarity.

All emitted cases run through `IlVerifier`, `Assembly.Load(bytes).GetTypes()`, then bounded child-process execution with both pipes drained asynchronously.

Known-bad controls:

- disabling the conditional funnel remained IL-valid and loadable, but child execution returned `1` instead of `3`, proving the lifted finally was skipped;
- disabling conditional-branch detection produced `ILVerify LeaveOutOfFinally`.

## Mutation matrix

Re-measured serially against the 10-test file. Clean control passed 10/10 before the matrix and again after every mutant was restored. Every row examined all 10 issue tests.

| Mutation | Semantic branch | Result |
|---|---|---:|
| Internal-target guard removed | preserve internal target | 4 failed |
| M-A: conditional funnel forced to a no-op | funnel escaping targets | 7 failed |
| Inverse polarity removed | preserve polarity | 8 failed |
| Condition replaced with a constant | preserve condition value | 6 failed |
| Condition evaluated twice | evaluate condition once | 6 failed |
| Discriminator assignment forced to zero | assign discriminator | 5 failed |
| Tail transfer redirected to the skip label | transfer to finally tail | 5 failed |
| Not-taken branch redirected to the tail | preserve not-taken skip path | 4 failed |
| Conditional branches omitted from analysis | detector records conditional branches | 10 failed |
| M-B: every discriminator routed to the first arm | select the matching target | 1 failed |

## Verification

Base: `210126dd90f4ab3edbe1b7f9219946594685afc7`

Executed counts from the base push run and this PR run, on the same base:

| Suite | Base | Head |
|---|---:|---:|
| Core | 7029 passed, 1 skipped | 7029 passed, 1 skipped |
| Compiler | 3952 passed | 3962 passed |
| Interpreter | 494 passed | 494 passed |

Local final-head issue suite: 10/10 passed. Release solution build with warnings as errors: 0 warnings, 0 errors. All PR checks pass.

Corpus: 148 files, 139 emitted on both base and head, 0 diagnostic differences, 0 byte differences. Source-level `goto` usage is 0/148; the only textual occurrence is in a comment. Corpus therefore carries no direct signal for this bug.

### DIFF
--- a/src/Core/CodeAnalysis/Lowering/FinallyExitRewriter.cs
+++ b/src/Core/CodeAnalysis/Lowering/FinallyExitRewriter.cs
@@ -75,6 +75,7 @@ internal static class FinallyExitRewriter
     {
         private readonly ExitPlan plan;
         private readonly ProtectedRegionBranchAnalysis branches;
+        private int skipOrdinal;
 
         public ExitFunneler(ExitPlan plan, ProtectedRegionBranchAnalysis branches)
         {
@@ -94,6 +95,27 @@ internal static class FinallyExitRewriter
                 ImmutableArray.Create<BoundStatement>(
                     AssignBranch(plan.GetDiscriminator(node.Label)),
                     new BoundGotoStatement(null, plan.TailLabel)));
+        }
+
+        protected override BoundStatement RewriteConditionalGotoStatement(BoundConditionalGotoStatement node)
+        {
+            if (branches.ContainsLabel(node.Label))
+            {
+                return node;
+            }
+
+            var skip = new BoundLabel("<>finally_exit_skip_" + skipOrdinal++);
+            return new BoundBlockStatement(
+                null,
+                ImmutableArray.Create<BoundStatement>(
+                    new BoundConditionalGotoStatement(
+                        null,
+                        skip,
+                        node.Condition,
+                        jumpIfTrue: !node.JumpIfTrue),
+                    AssignBranch(plan.GetDiscriminator(node.Label)),
+                    new BoundGotoStatement(null, plan.TailLabel),
+                    new BoundLabelStatement(null, skip)));
         }
 
         private BoundStatement AssignBranch(int discriminator)
--- a/test/Compiler.Tests/Emit/Issue2942ConditionalGotoFinallyFunnelTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2942ConditionalGotoFinallyFunnelTests.cs
@@ -79,8 +79,8 @@ public class Issue2942ConditionalGotoFinallyFunnelTests
 
     [Theory]
     [InlineData(1, 11)]
-    [InlineData(2, 21)]
-    [InlineData(3, 120)]
+    [InlineData(2, 61)]
+    [InlineData(3, 200)]
     public void EmittedMultipleConditionalExits_DispatchSelectedTarget_AndRunFinallyOnce(
         int selectedTarget,
         int expectedExitCode)
@@ -262,9 +262,13 @@ public class Issue2942ConditionalGotoFinallyFunnelTests
             new BoundLabelStatement(null, firstExit),
             new BoundReturnStatement(null, Read(result)),
             new BoundLabelStatement(null, secondExit),
-            new BoundReturnStatement(null, Read(result)),
+            new BoundReturnStatement(
+                null,
+                Binary(Read(result), SyntaxKind.PlusToken, Literal(40))),
             new BoundLabelStatement(null, returnLabel),
-            new BoundReturnStatement(null, Read(result)),
+            new BoundReturnStatement(
+                null,
+                Binary(Read(result), SyntaxKind.PlusToken, Literal(80))),
             new BoundLabelStatement(null, finallyExit),
             new BoundReturnStatement(null, Literal(-1)));
     }
--- a/test/Compiler.Tests/Emit/Issue2942ConditionalGotoFinallyFunnelTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2942ConditionalGotoFinallyFunnelTests.cs
@@ -1,0 +1,281 @@
+// <copyright file="Issue2942ConditionalGotoFinallyFunnelTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Immutable;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using GSharp.Core.CodeAnalysis;
+using GSharp.Core.CodeAnalysis.Binding;
+using GSharp.Core.CodeAnalysis.Compilation;
+using GSharp.Core.CodeAnalysis.Lowering;
+using GSharp.Core.CodeAnalysis.Symbols;
+using GSharp.Core.CodeAnalysis.Syntax;
+using GSharp.Core.CodeAnalysis.Text;
+using Xunit;
+
+namespace GSharp.Compiler.Tests.Emit;
+
+/// <summary>
+/// Issue #2942: conditional exits from protected code must run a lifted
+/// <c>finally</c> body before dispatching to their original target.
+/// No source form currently produces this shape, so these tests inject the
+/// sanctioned bound-tree regression directly into the emit pipeline.
+/// </summary>
+public class Issue2942ConditionalGotoFinallyFunnelTests
+{
+    private static readonly FieldInfo BoundProgramField = typeof(Compilation).GetField(
+        "boundProgram",
+        BindingFlags.Instance | BindingFlags.NonPublic)!;
+
+    [Theory]
+    [InlineData(true, 1)]
+    [InlineData(false, 2)]
+    public void Lowerer_FunnelsConditionalExit_AndLiftsConditionalFinallyExit(
+        bool jumpIfTrue,
+        int comparedValue)
+    {
+        var (body, tryExit, localTarget) = BuildBody(jumpIfTrue, comparedValue);
+
+        var lowered = Lowerer.Lower(body);
+        var escapingCollector = new BranchCollector(tryExit);
+        escapingCollector.VisitStatement(lowered);
+        var localCollector = new BranchCollector(localTarget);
+        localCollector.VisitStatement(lowered);
+
+        Assert.Equal(0, escapingCollector.ConditionalTargetCount);
+        Assert.Equal(0, escapingCollector.FinallyCount);
+        Assert.Equal(1, localCollector.ConditionalTargetCount);
+    }
+
+    [Theory]
+    [InlineData(true, 1, 3)]
+    [InlineData(false, 2, 3)]
+    [InlineData(true, 2, 221)]
+    [InlineData(false, 1, 221)]
+    public void EmittedConditionalExit_PreservesPolarity_EvaluatesConditionOnce_AndRunsFinally(
+        bool jumpIfTrue,
+        int comparedValue,
+        int expectedExitCode)
+    {
+        var (body, _, _) = BuildBody(jumpIfTrue, comparedValue);
+        var program = BuildProgram(Lowerer.Lower(body));
+
+        var bytes = Emit(program);
+        VerifyLoadAndRunChild(bytes, expectedExitCode);
+    }
+
+    private static (BoundBlockStatement Body, BoundLabel TryExit, BoundLabel LocalTarget) BuildBody(
+        bool jumpIfTrue,
+        int comparedValue)
+    {
+        var counter = new LocalVariableSymbol("counter", isReadOnly: false, TypeSymbol.Int32);
+        var result = new LocalVariableSymbol("result", isReadOnly: false, TypeSymbol.Int32);
+        var finallyFlag = new LocalVariableSymbol("finallyFlag", isReadOnly: false, TypeSymbol.Bool);
+        var tryExit = new BoundLabel("tryExit");
+        var finallyExit = new BoundLabel("finallyExit");
+        var localTarget = new BoundLabel("localTarget");
+
+        BoundExpression Read(VariableSymbol variable) => new BoundVariableExpression(null, variable);
+        BoundExpression Literal(int value) => new BoundLiteralExpression(null, value);
+        BoundExpression Binary(BoundExpression left, SyntaxKind kind, BoundExpression right)
+            => new BoundBinaryExpression(
+                null,
+                left,
+                BoundBinaryOperator.Bind(kind, left.Type, right.Type),
+                right);
+        BoundStatement Assign(VariableSymbol variable, BoundExpression expression)
+            => new BoundExpressionStatement(
+                null,
+                new BoundAssignmentExpression(null, variable, expression));
+
+        var incrementCounter = new BoundAssignmentExpression(
+            null,
+            counter,
+            Binary(Read(counter), SyntaxKind.PlusToken, Literal(1)));
+        var condition = Binary(
+            incrementCounter,
+            SyntaxKind.EqualsEqualsToken,
+            Literal(comparedValue));
+
+        var tryBlock = Block(
+            new BoundConditionalGotoStatement(
+                null,
+                localTarget,
+                new BoundLiteralExpression(null, true),
+                jumpIfTrue: true),
+            Assign(result, Literal(50)),
+            new BoundLabelStatement(null, localTarget),
+            new BoundConditionalGotoStatement(null, tryExit, condition, jumpIfTrue),
+            Assign(result, Literal(9)));
+        var finallyBlock = Block(
+            new BoundConditionalGotoStatement(
+                null,
+                finallyExit,
+                Read(finallyFlag),
+                jumpIfTrue: true),
+            Assign(result, Binary(Read(result), SyntaxKind.PlusToken, Literal(1))));
+
+        return (
+            Block(
+                new BoundVariableDeclaration(null, counter, Literal(0)),
+                new BoundVariableDeclaration(null, result, Literal(0)),
+                new BoundVariableDeclaration(
+                    null,
+                    finallyFlag,
+                    new BoundLiteralExpression(null, false)),
+                new BoundTryStatement(
+                    null,
+                    tryBlock,
+                    ImmutableArray<BoundCatchClause>.Empty,
+                    finallyBlock),
+                Assign(result, Binary(Read(result), SyntaxKind.PlusToken, Literal(100))),
+                new BoundLabelStatement(null, tryExit),
+                new BoundReturnStatement(
+                    null,
+                    Binary(
+                        Binary(Read(result), SyntaxKind.StarToken, Literal(2)),
+                        SyntaxKind.PlusToken,
+                        Read(counter))),
+                new BoundLabelStatement(null, finallyExit),
+                new BoundReturnStatement(null, Literal(-1))),
+            tryExit,
+            localTarget);
+    }
+
+    private static BoundProgram BuildProgram(BoundBlockStatement body)
+    {
+        var package = new PackageSymbol("Issue2942", declaration: null);
+        var entryPoint = new FunctionSymbol(
+            "Main",
+            ImmutableArray<ParameterSymbol>.Empty,
+            TypeSymbol.Int32,
+            package: package);
+        return new BoundProgram(
+            package,
+            ImmutableArray.Create(package),
+            ImmutableArray<Diagnostic>.Empty,
+            ImmutableDictionary<FunctionSymbol, BoundBlockStatement>.Empty.Add(entryPoint, body),
+            entryPoint,
+            Block());
+    }
+
+    private static byte[] Emit(BoundProgram program)
+    {
+        var compilation = new Compilation(
+            SyntaxTree.Parse(SourceText.From("package Issue2942\nfunc Placeholder() {}")));
+        BoundProgramField.SetValue(compilation, program);
+
+        using var peStream = new MemoryStream();
+        var emit = compilation.Emit(peStream);
+        Assert.True(
+            emit.Success,
+            string.Join(
+                Environment.NewLine,
+                emit.Diagnostics.Select(diagnostic => $"{diagnostic.Id}: {diagnostic.Message}")));
+        return peStream.ToArray();
+    }
+
+    private static void VerifyLoadAndRunChild(byte[] bytes, int expectedExitCode)
+    {
+        var prefix = $"Issue2942_{Guid.NewGuid():N}";
+        var directory = Directory.GetCurrentDirectory();
+        var assemblyPath = Path.Combine(directory, prefix + ".dll");
+        var runtimeConfigPath = Path.Combine(directory, prefix + ".runtimeconfig.json");
+        try
+        {
+            File.WriteAllBytes(assemblyPath, bytes);
+            File.WriteAllText(
+                runtimeConfigPath,
+                $$"""
+                  {
+                    "runtimeOptions": {
+                      "tfm": "net{{Environment.Version.Major}}.0",
+                      "framework": {
+                        "name": "Microsoft.NETCore.App",
+                        "version": "{{Environment.Version.Major}}.0.0"
+                      }
+                    }
+                  }
+                  """);
+
+            IlVerifier.Verify(assemblyPath);
+            Assert.NotEmpty(Assembly.Load(bytes).GetTypes());
+
+            var start = new ProcessStartInfo("dotnet")
+            {
+                RedirectStandardOutput = true,
+                RedirectStandardError = true,
+                UseShellExecute = false,
+                WorkingDirectory = directory,
+            };
+            start.ArgumentList.Add("exec");
+            start.ArgumentList.Add("--runtimeconfig");
+            start.ArgumentList.Add(runtimeConfigPath);
+            start.ArgumentList.Add(assemblyPath);
+
+            using var process = Process.Start(start)!;
+            var stdoutTask = process.StandardOutput.ReadToEndAsync();
+            var stderrTask = process.StandardError.ReadToEndAsync();
+            var exited = process.WaitForExit(10_000);
+            if (!exited)
+            {
+                process.Kill(entireProcessTree: true);
+                Assert.True(process.WaitForExit(5_000), "child did not stop after kill");
+            }
+
+            var stdout = stdoutTask.GetAwaiter().GetResult();
+            var stderr = stderrTask.GetAwaiter().GetResult();
+            Assert.True(exited, $"child execution timed out\nstdout:\n{stdout}\nstderr:\n{stderr}");
+            Assert.Equal(expectedExitCode, process.ExitCode);
+            Assert.Equal(string.Empty, stdout);
+            Assert.Equal(string.Empty, stderr);
+        }
+        finally
+        {
+            File.Delete(assemblyPath);
+            File.Delete(runtimeConfigPath);
+        }
+    }
+
+    private static BoundBlockStatement Block(params BoundStatement[] statements)
+        => new(null, statements.ToImmutableArray());
+
+    private sealed class BranchCollector : BoundTreeWalker
+    {
+        private readonly BoundLabel target;
+
+        public BranchCollector(BoundLabel target)
+        {
+            this.target = target;
+        }
+
+        public int ConditionalTargetCount { get; private set; }
+
+        public int FinallyCount { get; private set; }
+
+        public override void VisitStatement(BoundStatement node)
+        {
+            if (node is BoundConditionalGotoStatement conditional
+                && ReferenceEquals(conditional.Label, target))
+            {
+                ConditionalTargetCount++;
+            }
+
+            base.VisitStatement(node);
+        }
+
+        protected override void VisitTryStatement(BoundTryStatement node)
+        {
+            if (node.FinallyBlock != null)
+            {
+                FinallyCount++;
+            }
+
+            base.VisitTryStatement(node);
+        }
+    }
+}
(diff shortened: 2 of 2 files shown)


DavidObando:
## Review — PR #2997 `d7b6c45cd` (issue #2942)

Base `a6b1561f27617afe3b479d587b82269a6a5cda73`. Reviewed in isolated `git clone --shared` trees at both SHAs, `src/Compiler/Compiler.csproj -c Release -m:1`, exit-code gated, both `GSharp.Core.dll` non-zero and distinct (`3f28d46d…` / `8efb4b98…`). All scratch restored, both trees `git status --porcelain` empty.

**Verdict: MERGEABLE.** No blocking issues. Two non-blocking, two suggestions.

---

### The semantic claim, falsifiable

*Before:* a `BoundConditionalGotoStatement` in a `try` or `catch` body whose target is outside that body, in a `try` whose `finally` had to be lifted, was left untouched by `ExitFunneler`. `ProtectedRegionBranchRewriter` then turned it into a plain `leave` **straight past the lifted finally body**. Valid IL, silently wrong answer.

*After:* it becomes `gotoIf(!JumpIfTrue) skip; pendingBranch = d; goto tail; skip:`, so it lands on the same tail label as the unconditional funnel and is dispatched after the lifted body.

**Confirmed at IL level**, catch-clause case, decoded with `System.Reflection.Metadata`:

```
base                                head
IL_0016: stloc.s 5   (catch)        IL_0016: stloc.s 5   (catch)
IL_0018: leave -> 0062  <-- skips   IL_0018: ldc.i4.1
         the lifted finally at      IL_001A: dup
         0x0039 entirely            IL_001B: stloc.3      <-- discriminator
                                    IL_001C: leave -> 003D <-- tail = lifted finally
                                    …
                                    IL_005F: ldloc.3
                                    IL_0060: ldc.i4.1
                                    IL_0061: ceq
                                    IL_0063: brfalse -> 006D
                                    IL_0068: br -> 0074    <-- original target
```

Two-target case, non-constant conditions: each condition appears **exactly once** in the stream (`ldloc.0; ldc.i4.1; add; dup; stloc.0; ldc.i4.s 99; ceq; brfalse -> skip`), polarity correctly inverted, two distinct skip labels at `0x0022` / `0x0038`, discriminators `1` and `2`.

**`finally` exactly once** is structural, not incidental: the EH table for a lifted method contains **zero `Finally` clauses**, the lifted body exists as a **single** copy at the tail label, and every exit is funnelled to that label. Nothing can reach the original target without passing through it.

---

### Independent differential — the PR's own test, plus two cases it does not cover

Ran the PR's test file against the **base** compiler:

```
base: Failed: 4, Passed: 2, Total: 6      (Expected 3, Actual 1 — lifted finally skipped)
head: Failed: 0, Passed: 6, Total: 6
```

Then wrote **my own** bound-tree probes for the two paths the PR's test never reaches — a `catch` clause and two distinct escaping targets:

| probe | base | head |
|---|---:|---:|
| escaping conditional goto in a **catch clause** | **7** | 10 |
| two targets, first taken | **10** | 11 |
| two targets, second taken | **20** | 21 |
| two targets, neither taken | 120 | 120 |

`base` 3/4 fail with exactly the finally-skipped delta; `head` 4/4 pass. `IlVerifier.Verify` + `Assembly.Load(bytes).GetTypes()` + bounded child exec (`WaitForExit(15_000)`, `Kill(entireProcessTree: true)`, both pipes drained with `ReadToEndAsync()` before the wait). The code is correct on the uncovered paths — see NB-1.

---

### The `finally`-once matrix, from source

Nine source programs, compiled with both compilers, emitted assembly md5-compared and executed:

| probe | shape | base vs head | output |
|---|---|---|---|
| p1 | `break` in `finally`, in loop | SAME | `12` |
| p2 | `continue` in `finally`, in loop | SAME | `24` |
| p3 | `goto` out of `finally` | SAME | `1` |
| p4 | nested `try/finally`, `break` from inner `finally` | SAME | `abcdeabce` |
| p5 | `return` from `try` + `break` from `finally` | SAME | `21` |
| p6 | exception propagation through lifted `finally` + `catch` | SAME | `tffc` |
| p7 | **iterator**, `yield` in `try/finally`, `break` in `finally` | SAME | `1 2 3 23` |
| p8 | **async**, `await` in `try/finally`, `break` in `finally` | SAME | `22` |
| p9 | **iterator**, `goto` out of `finally` | SAME | `1 2 101` |

All nine byte-identical base↔head and all nine semantically correct (`finally` once per path, outer `finally` runs on the inner `break`, p4 `abcde`+`abce`). Independently reproduces the "no source producer" claim and shows no regression.

Also checked: `Lowerer.cs:1388` computes `preEmitAnalysisBody` **before** `FinallyExitRewriter.Rewrite`, so `ControlFlowGraph`/GS0100 analysis never sees the funnel. That is why the corpus shows 0 diagnostic differences, and it means no interaction with the #2843/#2883 CFG work.

---

## NON-BLOCKING 1 — the catch-clause funneler and multi-arm dispatch ship with zero coverage

`ExitFunneler` is instantiated once per catch clause (`FinallyExitRewriter.cs:132-136`), but the test builds `ImmutableArray<BoundCatchClause>.Empty` and exactly one escaping target. Consequences:

- the new override's **catch** path is never executed by the suite;
- `skipOrdinal` never exceeds `0`, so its uniqueness is untested;
- `ExitPlan.GetDiscriminator` never allocates a second discriminator, so the multi-arm dispatch loop (`:203-219`) is never exercised with more than one arm.

The M1–M9 mutation matrix therefore measures the try-block path only. **Impact:** medium — those paths work (I proved it above), but a future edit to them is unguarded.

**Fix:** fold in the two shapes I ran. A catch clause reached by `1 / zero`, containing an escaping conditional goto; and a try body with two escaping conditional gotos to two different labels, driven by a side-effecting `counter+1 == sel` so the not-taken arm also proves single evaluation. ~40 lines, reuses the existing `VerifyLoadAndRunChild`.

## NON-BLOCKING 2 — the new label counter bypasses the file's existing unique-label helper

`Rewriter.RewriteTryStatement:122` already has `BoundLabel NewLabel() => new($"<>finally_skip_{currentOrdinal}_{ordinal++}")`, fed by the **method-wide** `ordinal`. The new code introduces a second, **per-`ExitFunneler`** counter, so the try funneler and every catch funneler of the same statement all emit `<>finally_exit_skip_0`.

Harmless **today**: `BoundLabel` (`Binding/BoundLabel.cs`) does not override `Equals`/`GetHashCode`, and every downstream map is reference-keyed — `MethodBodyEmitter.cs:54`, `MethodBodyPlanner.cs:327/1018/1217`, `ControlFlowGraph.cs:741/976`, `Evaluator.cs:264`, `ExitPlan.discriminators`. Only `BoundNodePrinter.cs:521/535/543` reads `.Name`.

**Impact:** low, but it makes bound-tree dumps ambiguous exactly in the nested/catch cases this pass exists to handle, and it becomes a real bug the day anything keys on name. **Fix:** pass the existing `NewLabel` delegate into `ExitFunneler` and delete `skipOrdinal`.

## SUGGESTION 1 — the producer proof's ordering claim is wrong, and it is the load-bearing sentence

The PR body says:

> async/iterator/state-machine builders and spillers: **generated after** the ordinary `FinallyExitRewriter`

That is backwards for iterators. `IteratorMoveNextBodyBuilder.cs:150` and `AsyncIteratorRewriter.cs:65` both call `Lowerer.Lower(...)` **on the already-assembled state-machine body** — so `FinallyExitRewriter` runs *after* the state dispatch is constructed, and the dispatch is a chain of `BoundConditionalGotoStatement` (`IteratorMoveNextBodyBuilder.cs:120-129`) some of which target labels inside user `try` regions (`:110-119`).

The **conclusion still holds** — p7/p8/p9 above cover exactly that intersection and are byte-identical — but the stated reason is what a future reader will trust. Replace it with the actual invariant: the dispatch chain sits *outside* the user try region, and the inner dispatch placed at the top of a user try targets `yieldLabels[state]` **inside** the same region, so `branches.ContainsLabel` is true and the funnel correctly declines.

## SUGGESTION 2 — this code is dead from source; the things most likely to wake it are in flight

Confirmed dead by three independent routes: the author's 148-file corpus (0 byte diffs), my 9 source probes, and the producer classification. Its only coverage is a synthetic bound tree injected via `typeof(Compilation).GetField("boundProgram", NonPublic)`. So the first lowering change that *does* emit the shape will hit this code in production with only synthetic coverage behind it.

**#3000** (`Lower yield in function-literal iterator bodies`) and **#2967** (`preserve async select receive bindings`) both add generated control flow near protected regions. Cheap insurance: re-run p7/p9 after either merges.

---

### Contention — clean

Touches `src/Core/CodeAnalysis/Lowering/FinallyExitRewriter.cs` and one new test file. **No overlap** with `Emit/MethodBodyEmitter.Statements.cs` (#2967/#2964/#2965), `Lowering/Iterators/IteratorRewriter.cs` (#3000/#2975), `Evaluator.Expressions.cs` (#3000/G2/G4), or `test/Compiler.Tests/IlVerifier.cs` (#2966). It *calls* `IlVerifier.Verify` but does not modify it.

`git merge-tree --write-tree` against all 17 other open heads at natural merge bases: **0 conflicts**. Merges clean onto current `main`.

### Test-count arithmetic

Compiler `3906 → 3912` is exactly `+6`, matching the 6 new theory cases (2 `Lowerer_…` + 4 `EmittedConditionalExit_…`). I did not re-run the full suites (load ceiling); the filtered runs above are the evidence.

DavidObando:
# Re-review — PR #2997 @ `63b1046ed` — **MERGEABLE**

Force-push confirmed genuine: `d7b6c45cd` is **not** an ancestor of `63b1046ed`. Merge base computed, not assumed: `git merge-base refs/remotes/pr/2997 origin/main` = **`e80132516`**. Three-dot delta = 2 files, **+447** (`FinallyExitRewriter.cs +22`, `Issue2942ConditionalGotoFinallyFunnelTests.cs +425`). The naive two-dot `d7b6c45cd..63b1046ed` shows 16 files / +1611 — pure rebase noise, ignored.

**Production is byte-identical across the force-push**: `git show <head>:src/Core/CodeAnalysis/Lowering/FinallyExitRewriter.cs | md5` = `8cd5644a127a866db2d8bd3e9bf75c7a` at **both** heads. So the round-1 production analysis carries unchanged — IL decode, finally-once matrix, `<>finally_exit_skip_N` name duplication being cosmetic (`BoundLabel` has no `Equals`/`GetHashCode`; `ExitPlan.discriminators` at `:40` is a reference-keyed `Dictionary<BoundLabel, int>`), and `Lowerer.cs:1390` running `FinallyExitRewriter.Rewrite` **after** `preEmitAnalysisBody` so no diagnostics drift. Not re-litigated.

The delta is 100% test. Both of my round-1 items were aimed at it, so the whole verdict is about whether the new tests actually bite.

---

## NB-1(a) — catch-clause funneler — **CLOSED**

`EmittedCatchConditionalExit_RunsFinallyBeforeDispatch` + `BuildCatchBody()` now pass a real `ImmutableArray.Create(new BoundCatchClause(caught.Type, caught, catchBlock))` whose body holds a `BoundConditionalGotoStatement` escaping to `catchExit`, so the per-clause `ExitFunneler` at `:132-136` is instantiated **and** funnels.

Baseline in a `--shared` scratch clone at `63b1046ed`, narrow filter only:

```
dotnet test test/Compiler.Tests/Compiler.Tests.csproj -c Release \
  -p:ContinuousIntegrationBuild=true --filter 'FullyQualifiedName~Issue2942'
→ Passed! Failed: 0, Passed: 10          (was 6 at d7b6c45cd; +1 Fact +3 Theory rows)
```

**Mutant M-A** — `RewriteConditionalGotoStatement` forced to a no-op (`if (true || branches.ContainsLabel(node.Label))`):

```
Failed! Failed: 7, Passed: 3
  Failed EmittedCatchConditionalExit_RunsFinallyBeforeDispatch          <-- the new test
  Failed EmittedMultipleConditionalExits_...(selectedTarget: 1, ...: 11)
  Failed EmittedMultipleConditionalExits_...(selectedTarget: 2, ...: 21)
  Failed EmittedConditionalExit_PreservesPolarity_...  x2
  Failed Lowerer_FunnelsConditionalExit_...            x2
```

The catch test fails under mutation, so the coverage is real. This also disposes of the one way it could have been illusory — the catch condition is `new BoundLiteralExpression(null, true)`, and I could not find any constant-folding of `BoundConditionalGotoStatement` anywhere in the lowering pipeline; M-A proves it empirically rather than by absence-of-grep. `selectedTarget: 3` correctly survives M-A: that row takes no conditional escape, so it is the not-taken control.

## NB-1(b) — multi-target dispatch — **structurally closed, semantically still blind** (SHOULD-FIX)

`BuildMultipleTargetBody` does what was asked *structurally*: two escaping `BoundConditionalGotoStatement`s in one try block, funneled by one `ExitFunneler` instance, so `skipOrdinal` goes 0 → 1 and `plan.Arms` reaches 2, driving the `foreach (var arm in plan.Arms)` loop at `:220-238` with more than one arm. That part of NB-1 is answered.

But the three dispatch targets are **indistinguishable**:

```csharp
new BoundLabelStatement(null, firstExit),
new BoundReturnStatement(null, Read(result)),
new BoundLabelStatement(null, secondExit),
new BoundReturnStatement(null, Read(result)),
new BoundLabelStatement(null, returnLabel),
new BoundReturnStatement(null, Read(result)),
```

All three `return result`. The exit code is decided by *which conditional fired inside the try*, never by *where the dispatcher sent control*. So arm selection is unobservable.

**Mutant M-B** — one line in `ExitPlan.GetDiscriminator` (`:60-69`), keeping every arm registered but routing every funneled exit to the first one:

```diff
-            return discriminator;
+            return arms[0].Discriminator;
```

```
Passed! Failed: 0, Passed: 10, Total: 10
```

**The entire suite survives a total mis-dispatch.** Not a build no-op — `out/bin/Release/Core/GSharp.Core.dll` md5 `654668e3aa39b6c0db6a1e08b61b2fef` under M-B vs `dbbc059a095baeb33233fa3813a43779` restored, and the restored tree passes 10/10 with `git status --porcelain` empty. Under M-B `GetDiscriminator(secondExit)` registers arm 2 but returns 1, so `secondExit`'s funnel writes discriminator 1 and control lands on `firstExit` — and every expectation still matches because both labels run the same return.

This is the same *class* of gap NB-1 named: the path is instantiated but cannot fail. It is not a correctness defect in production — `DispatchArm` pairs `Discriminator` and `Target` in one struct, so mis-pairing is structurally hard — but the test claims coverage it does not have.

**One-line fix**, keeping the funnel shape untouched:

```csharp
new BoundLabelStatement(null, firstExit),
new BoundReturnStatement(null, Read(result)),
new BoundLabelStatement(null, secondExit),
new BoundReturnStatement(null, Binary(Read(result), SyntaxKind.PlusToken, Literal(100))),
new BoundLabelStatement(null, returnLabel),
new BoundReturnStatement(null, Binary(Read(result), SyntaxKind.PlusToken, Literal(200))),
```

Expectations become `11 / 121 / 320`, and M-B then fails on `selectedTarget: 2`.

Corollary: **the Mutation matrix in the PR body is stale.** It states "Clean 6/6 control ran before every build-green mutant" — the control is now 10/10, and it was measured against the old 281-line file at base `a6b1561f2`, not `e80132516`. M2 "route escaping target" is listed as *4 failed*; M-B is a routing mutant that now survives. Re-run the matrix against the 10-test file, or scope the table to the base it was measured on.

## SUG-1 — backwards sentence — **materially closed; residual over-claim** (SHOULD-FIX)

The body now reads:

> `Lowerer.Lower` runs on the assembled state-machine body, so `FinallyExitRewriter` sees dispatch construction; the outer dispatch is outside the user try, while the inner dispatch at the top of a user try targets `yieldLabels[state]` inside that same region, so `ContainsLabel` is true and the funnel correctly declines

The load-bearing correction landed: the ordering is no longer inverted and the invariant is correctly named as `ContainsLabel`. Verified against blobs at `origin/main` (`f70626452`), not the working tree:

- `src/Core/CodeAnalysis/Lowering/Iterators/IteratorMoveNextBodyBuilder.cs:150` — `Lowerer.Lower(new BoundBlockStatement(null, statements.ToImmutable()))`, i.e. on the **assembled** body. ✅ sentence true here.
- `MoveNextBodyRewriter.cs:198-207` — outer dispatch target is `tryDispatch.GetOuterDispatchTarget(rp.State) ?? rp.ResumeLabel`, and the comment states the entry label is *"placed just before that try"*. ✅ "outside the user try".
- `MoveNextBodyRewriter.cs:647-656` — inner dispatch is prepended into `newTryBodyStmts` targeting `entry.Target` (`TryDispatchEntry.Target`, "one entry in a user try's internal state dispatch"). ✅ `ContainsLabel` true, funnel declines.

**Residual imprecision**, in the opposite direction from round 1: the bullet asserts one ordering for *"async/iterator/state-machine builders and spillers"* uniformly, but `src/Core/CodeAnalysis/Lowering/Iterators/AsyncIteratorRewriter.cs:65` lowers **before** assembly —

```csharp
// Issue #2662: lower loops first so synthesized await-for locals
// and awaits participate in state-machine slot/state allocation.
var loweredBody = Lowerer.Lower(body);
```

— and `git grep -c 'Lowerer\.Lower(' origin/main -- 'src/Core/CodeAnalysis/Lowering/Async/*'` returns **zero**. So on the async path `FinallyExitRewriter` never sees the assembled state-machine body at all; it is protected by **absence**, not by `ContainsLabel`.

The error is conservative (it over-claims exposure, where round 1 under-claimed it), and the `ContainsLabel` argument would still hold if the async body were ever re-lowered, per the two `MoveNextBodyRewriter` findings above. So the conclusion survives intact. Split the bullet in two — "iterator builder: Lower runs after assembly, `ContainsLabel` declines" and "async pipeline: Lower runs before assembly, the rewriter never sees dispatch" — and it is exactly right.

## NIT — Verification table base

Suite counts and the corpus run cite base `a6b1561f27…`; the merge base is now `e80132516`, four merges later. Numbers are almost certainly still fine, but the stated base no longer matches the diff being reviewed.

---

## Verdict

**MERGEABLE.** Production unchanged from the previously-reviewed and gate-passed code; the one item the coordinator held this on — the false ordering sentence — is corrected in the direction that matters, and the catch-clause coverage gap is genuinely closed under mutation. The two SHOULD-FIXes are a one-line test change and a one-sentence split; neither affects emitted IL.

<sub>Throttle honoured — `uptime` gated before each run, all three `dotnet test` invocations serialised and filtered to `FullyQualifiedName~Issue2942`, no full suite. All mutation work in `/tmp/g2997r/h` (`--shared` clone, explicit `git checkout 63b1046ed`), restored and `git status --porcelain` empty. Nothing written to `~/.nuget/packages/gsharp.net.sdk/`. No `git stash`.</sub>

DavidObando:
# Re-review — PR #2997 @ `39e7bcf01` — **NOT MERGEABLE** (one blocker, body-only, ~5 min to clear)

The code is done. **NB-1(b) is genuinely closed under mutation** and NB-1(a) still holds. What blocks is a normative claim in the PR body that is measurably false at this tip.

First, a correction that changes the scope of this review.

---

## The "+526/−38 across 5 files" is rebase noise — the PR is 2 files / +451

This head is a **second force-push**: `63b1046ed` is **not** an ancestor of `39e7bcf01` (`git merge-base --is-ancestor` → rc=1). The branch was also **rebased forward**, so the merge base moved:

```
merge-base(old head, main) = e80132516
merge-base(new head, main) = 5e616b9bb      <-- moved
```

Three-dot from the **computed** base is the PR:

```
$ git diff --stat 5e616b9bb...39e7bcf01
 src/Core/CodeAnalysis/Lowering/FinallyExitRewriter.cs           |  22 ++
 test/Compiler.Tests/Emit/Issue2942ConditionalGotoFinallyFunnelTests.cs | 429 +++
 2 files changed, 451 insertions(+)
```

The `docs/adr/0125-fixed-pinned-statement.md`, `MethodBodyEmitter.Statements.cs`, `Issue2922FixedSwitchIlTests.cs` and `Issue2906ExhaustiveSwitchReturnEmitTests.cs` entries come from the **old-head→new-head tree comparison**, which straddles the rebase. Proven, not asserted — each of those four files is byte-identical at the new base, at this head, **and** at `main`:

| file | `5e616b9bb` | `39e7bcf01` | `main` |
|---|---|---|---|
| `docs/adr/0125-fixed-pinned-statement.md` | `4d1e7e09…` | `4d1e7e09…` | `4d1e7e09…` |
| `MethodBodyEmitter.Statements.cs` | `d4a18ecc…` | `d4a18ecc…` | `d4a18ecc…` |
| `Issue2906ExhaustiveSwitchReturnEmitTests.cs` | `40c2bca5…` | `40c2bca5…` | `40c2bca5…` |
| `Issue2922FixedSwitchIlTests.cs` | `ee73e474…` | `ee73e474…` | `ee73e474…` |

**Consequences:** there is no undeclared scope, no ADR edit, no `MethodBodyEmitter` change, and **no `Fixes #2922` is owed** — that file arrived from upstream. The body declares `Fixes #2942` and mentions no other issue number, so "mentioned ≠ declared" is clean, and with a 2-file surface there is nothing it could silently close.

**Docs gate: not exposed.** Added lines in the true three-dot diff matching `DiagnosticDescriptor|GS[0-9]{4}` = **0**. Positive control, same pattern against `DiagnosticDescriptors.cs` = **378 matches**, so the pattern is proven able to fire.

---

## NB-1(b) — **CLOSED**, mutant now dies

The fix differentiates the three arms (`result` / `result+40` / `result+80` → `11 / 61 / 200`, pairwise distinct).

Baseline at `39e7bcf01`, narrow filter, exit code captured before anything else ran:
```
rc=0   Passed!  Failed: 0, Passed: 10, Total: 10
```

**Mutant M-B** — `FinallyExitRewriter.cs:60-70`, `ExitPlan.GetDiscriminator`, total mis-dispatch:
```diff
-            return discriminator;
+            return arms[0].Discriminator;
```
```
rc=1   Failed!  Failed: 1, Passed: 9, Total: 10
  Failed EmittedMultipleConditionalExits_...(selectedTarget: 2, expectedExitCode: 61)
    Assert.Equal() Failure: Expected: 61 / Actual: 21
    at ...Issue2942ConditionalGotoFinallyFunnelTests.cs:line 381
```

`61` is `secondExit` (`result+40`); `21` is the raw `result` returned at `firstExit` = `arms[0]`. **The mis-dispatch is now observed as a value.** At `63b1046ed` this same mutant passed 10/10.

Round trip closed: `8cd5644a…` → `975cd432…` → `8cd5644a…`, tree clean.

## NB-1(a) — still pinned

**Mutant M-A** — `RewriteConditionalGotoStatement:102` forced to a no-op:
```
rc=1   Failed! Failed: 7, Passed: 3
  Failed EmittedCatchConditionalExit_RunsFinallyBeforeDispatch        <-- catch coverage
  Failed EmittedMultipleConditionalExits_...(selectedTarget: 1, ...11)
  Failed EmittedMultipleConditionalExits_...(selectedTarget: 2, ...61)
  Failed EmittedConditionalExit_PreservesPolarity_...  x2
  Failed Lowerer_FunnelsConditionalExit_...            x2
```
Identical to the previous head. `selectedTarget: 3` correctly survives — it is the not-taken control.

Production `FinallyExitRewriter.cs` is byte-identical across **both** force-pushes (`8cd5644a127a866db2d8bd3e9bf75c7a` at `63b1046ed` and `39e7bcf01`), so the round-1 IL and finally-once analysis carries unchanged. This fix was test-only.

---

# BLOCKER 1 — the "Mutation matrix" in the PR body is false at this tip

Body `:49`:
> Clean **6/6** control ran before every build-green mutant.

The suite is **10** tests. Measured above: `Failed: 0, Passed: 10, Total: 10`.

Body `:51-60` — every row was measured against the 281-line / 6-test file. None describes the 429-line / 10-test file:

| row | claims | measured at `39e7bcf01` |
|---|---|---|
| `:54` M2 "route escaping target" | 4 failed | **1 failed / 9 passed** (M-B above) |
| `:53` M1 "preserve internal target" | 4 failed | M-A shape gives **7 failed / 3 passed** |

`M2 = 4 failed` is the specific item already falsified once — it read `0 failed` at the previous head, and now reads `1`. It has never been `4` against the current test file. A reader takes this table as the evidence that the funnel is covered; it is the one artefact in the PR that no CI job can check.

**Fix:** re-measure against the 10-test file, or delete the table. Body edit only — no code, no rebuild, no CI cycle.

# SHOULD-FIX 1 — body `:24`, the SUG-1 sentence still over-claims (same edit)

> async/iterator/state-machine builders and spillers: `Lowerer.Lower` runs on the assembled state-machine body, so `FinallyExitRewriter` sees dispatch construction

The **direction** is corrected — the original backwards sentence is gone. But it is asserted for the whole class, and it is false for one member:

- `Lowering/Iterators/IteratorMoveNextBodyBuilder.cs:150` — lowers the assembled `statements` including dispatch. **True.**
- `Lowering/Iterators/AsyncIteratorRewriter.cs:65` — `Lowerer.Lower(body)` on the **raw user body** (`// Issue #2662: lower loops first…`), with `AsyncExceptionHandlerRewriter` / `SpillSequenceSpiller` running *after*. **False there.**

Denominators, re-measured at this tip: `Lowerer.Lower(` under `Lowering/Async/` = **0 of 23 files**; positive control under `Lowering/Iterators/` = **2 of 6**, same pattern and pathspec form.

**Conservative error — the conclusion survives.** The async path is protected by the *absence* of a lowering pass over the assembled machine, not by `ContainsLabel`. Narrow "builders and spillers" to the iterator `MoveNext` builder and state the async case as protected by absence.

# NIT 1 — `FinallyExitRewriter.cs:88` no-op survives 10/10

While targeting M-A I hit `RewriteGotoStatement:88` instead of `:102` — the two methods share a byte-identical guard line. That accidental mutant passed **10/10**: no-op'ing the *unconditional* goto funneler is invisible to this suite. `RewriteGotoStatement` is **pre-existing** production code, not part of this PR's +22, so this is out of scope — recorded only so it is not mistaken for new coverage. It is also a live hazard for anyone doing future mutation work on this file: a naive whole-file substitution silently mutates the wrong method.

# NIT 2 — one of three rows discriminates

Under M-B only `selectedTarget: 2` fails. `1` targets `arms[0]` and is correct by construction; `3` reaches `returnLabel` by fallthrough, not through the funnel — inferred from the observation itself, since a funneled target 3 would have returned `21` instead of `200`. Sufficient to kill a total mis-dispatch, which was the ask; noting the true discriminating power is 1 row, not 3.

---

## Gate note

The six-gate run was performed at `63b1046ed`, against base `e80132516`. This head has a **different base** (`5e616b9bb`), so that run does not transfer automatically even though production is byte-identical. My own builds at `39e7bcf01` were full Release, `-p:ContinuousIntegrationBuild=true`, never `--no-build`, and exited 0 for `Compiler.Tests` and its project references — evidence the rebase compiles, but not a substitute for a full-solution gate.

**Clear BLOCKER 1 (a body edit) and this is MERGEABLE.** No code change is required.

DavidObando:
Review follow-up cleared at `400f3202fe91d4c2dcb00359cc0cc128c8799d36`:

- multi-target exits now return distinct values; total mis-dispatch mutant fails 1/10;
- producer proof now separates ordinary iterator post-assembly lowering from async pre-assembly lowering;
- mutation table was re-run against final 10-test head with freshness guards (M1–M10 all detected);
- rebased onto `210126dd90f4ab3edbe1b7f9219946594685afc7`; base/head executed counts and 139-emission corpus comparison are in the body;
- Release warnings-as-errors build and all PR checks pass.

PR #2998, opened 2026-08-01
Fix fixed return fallthrough guard

### LINKED ISSUE #2953: Fixed-return epilogue bypasses non-void fall-through guard

DavidObando:
Fixes #2953

## Summary

- Route ordinary non-void fallthrough around the shared `fixed` return epilogue so it reaches the existing compiler-generated guard.
- Add the same guard before protected-region return epilogues synthesized by lowering, while preserving the pre-emit body used by definite-return analysis.
- Intentionally change protected-region fallthrough in `gsi` from exit 0 with a manufactured default value to exit 1 with self-contained `GS0100`; the bare-switch case in #3006 remains separate.
- Replace the documented limitation and the existing regression test that expected a silent default return.

## Root cause

Two return-funneling mechanisms hid real fallthrough from the final emitter check:

1. Returns inside `fixed` use a shared emitter epilogue. Unmatched fallthrough entered that epilogue directly and loaded its uninitialized default slot.
2. Returns inside `try`, `catch`, `finally`, or `scope` use a lowering-generated value slot and exit label. Unmatched fallthrough reached its final `return $returnTemp` with the slot still at its default value.

The fixed epilogue now has a branch around it to the existing guard. Protected-region lowering inserts the same guard before its exit label only in the emitted body; the pre-emit analysis body remains unchanged, so GS0100 diagnostics do not weaken.

`EmitFixedReturnEpilogue` has one other caller: `ConstructorBodyEmitter.AddCompletedMethodBody`. Constructor-family bodies are void, so they have no non-void default slot and their ordinary fallthrough must continue to the shared epilogue and `ret`; no guard branch belongs there. Existing Issue2900 coverage already executes an instance-constructor return, finalizer return, void-function return, and static-initializer return.

## Coverage

Nine runtime pins compile, load with `Assembly.Load(...).GetTypes()`, and execute in child processes:

- fixed return in an enum-switch arm
- fixed return in a sealed-interface pattern arm
- nested fixed regions
- exhaustive switch inside fixed
- fixed inside a loop
- fixed inside try
- try inside fixed
- plain try/finally return funneling
- scope return funneling

Guards cover successful fixed and protected returns, void protected fallthrough, and non-exhaustive fixed/protected switches retaining GS0100. `Issue2953ProtectedReturnFallthroughInterpreterTests` pins the interpreter path. The bound throw carries `DiagnosticDescriptors.AllPathsMustReturn`; evaluation uses `EvaluatorException`'s descriptor factory, preserving the guard message while deriving `GS0100` and severity from the descriptor.

## Verification

- Direct-`gsc` differential on exact base `f7062645` → head `cb608826`: base exits 0 with stdout `0`; head aborts with `InvalidOperationException: Compiler-generated guard reached: non-void function fell through without returning a value.`
- Current head `8d77d0107` is rebased onto `f820484c3` (which contains requested base `a4a660f32`); `merge-tree --write-tree` exits 0 against both, and also against current `main` `61b7a08c1`, which advanced after the push.
- Post-rebase emitter mutation: removing `MethodBodyEmitter.cs` guard routing yields 6 failed / 25 passed of 31; restoring yields 31/31 passed.
- Post-rebase Lowerer mutation: removing the protected-return guard yields 1/1 interpreter test failed and `gsi` output `0`, exit 0; restoring yields 1/1 passed and `GS0100: Compiler-generated guard reached: non-void function fell through without returning a value.`, exit 1.
- ADR collision check: exactly one `0139-*` file and one `0153-*` file; no ADR was added or renumbered.
- Bidirectional diagnostic documentation gate: 4/4 passed; this PR adds no descriptor or diagnostic row.
- Pre-#2926 boundary `e21677f6`: both fixed and non-fixed programs fail compilation with GS0100.
- Rebased fixed-flow, exhaustive-switch, #2964 IL, and constructor-family tests: 39 passed.
- Mutation results were re-run unchanged without `dotnet test --no-build`. Every probe gated build exits, rejected zero-length DLLs by file size, required matching MD5s across the five ordinary Core/Compiler/test outputs, and proved each exact pristine → mutant → pristine round trip returned to pristine MD5 `91c4893873b144775a3fdaf1fdf23bf8`. Source-control metadata generation was disabled only for these fingerprint builds because Microsoft.Build.Tasks.Git cannot resolve a linked-worktree commit after a forced rebuild; production builds retain normal SourceLink:
  - remove fixed-epilogue route: 5 failed / 8 passed
  - remove protected-epilogue guard: 4 failed / 9 passed
  - force guard for void returns: 1 failed / 12 passed
  - place guard after protected return label: 1 failed / 12 passed
  - expose guard to pre-emit analysis: 1 failed / 12 passed
- Executed suites on exact base `f7062645` → head `cb608826`:
  - Core: 7029 passed, 1 skipped → 7029 passed, 1 skipped
  - Compiler: 3949 passed → 3962 passed (+13 issue executions)
  - Interpreter: 494 passed → 494 passed
  - LSP: 452 passed → 452 passed
- Current rebased-head CI is green: build, cs2gs, e2e, VS Code extensions, Visual Studio extension, and VSIX all passed.
- Final solution build from a scratch clone with normal SourceLink: 0 warnings, 0 errors; all five ordinary `GSharp.Core.dll` outputs matched and were nonzero.
- With merged #2964 on the base, full IL dumps confirmed both prologues compose with this epilogue:
  - array: `ldelema` → `Unsafe.AsPointer<int32>` → pinned cleanup → return-slot epilogue; unmatched flow branches around it to the guard
  - string: `GetPinnableReference` → `Unsafe.AsPointer<char>` → pinned cleanup → return-slot epilogue; unmatched flow branches around it to the guard
  - both unmatched child processes exited 134 with the expected `InvalidOperationException`
- The production delta adds no `.ClrType` lookup or `?? typeof(object)` fallback.
- Separate interpreter finding filed as #3006: direct exhaustive enum-switch statement fallthrough still returns `0` under `gsi`; the compiled path throws. This is outside the CIL epilogue path fixed here.
- Two intentional refactoring-baseline hashes changed: `ReturnInTryFinally.gs` and `YieldInTryFinally.gs`.

Production delta: guard routing in `MethodBodyEmitter.cs`/`Lowerer.cs`, descriptor metadata on compiler-generated bound throws, and descriptor-factory evaluation plumbing.


DavidObando:
# Review — PR #2998, head `4b1461ced`, base `a6b1561f2`

**Verdict: MERGEABLE**, with two SHOULD-FIX items about a second consumer the PR does not mention. The production change itself is correct, minimal, and I could not break it.

Content confirmed before probing (`git diff --stat a6b1561f2..4b1461ced` on production files: `MethodBodyEmitter.cs` +15/-2, `Lowerer.cs` +30/-3 — not a content-free tree). Scratch clones at both SHAs, `src/Compiler` and `src/Repl` built at each, exit-code gated, Core DLLs md5-distinct (`8c795387…` head / `5d0ec5d1…` base).

## Base→head differential observed

```gsharp
func F(x DateTimeKind, values []int32) int32 {
    switch x {
        case DateTimeKind.Unspecified { unsafe { fixed p *int32 = values { return *p } } }
        case DateTimeKind.Utc { return 2 }
        case DateTimeKind.Local { return 3 }
    }
}
Console.WriteLine(F(DateTimeKind(99), []int32{7}))
```

| | stdout | exit |
|---|---|---|
| base `a6b1561f2` | `0` | 0 |
| head `4b1461ced` | `InvalidOperationException: Compiler-generated guard reached: …` | 134 |

That is the defect in #2953 and it is fixed.

## 1. What the guard guards, and whether it is in the right place

Two independent funnels could reach a method's exit without an explicit return, and each needed its own fix:

- **Emitter (`MethodBodyEmitter.cs:245-261`)** — returns inside `fixed` branch to a shared epilogue that loads `fixedReturnSlot`. Previously, natural fallthrough off the end of the body ran straight *into* that epilogue and returned its uninitialised slot. The fix emits `br` to a new label before the epilogue, so the epilogue is now reachable only by a real rewritten return.
- **Lowering (`Lowerer.cs:1386-1421`)** — returns inside `try`/`catch`/`finally`/`scope` funnel through `returnValueLocal` + `methodExitLabel`. Fallthrough reached `return $returnTemp` with the slot still default. The fix inserts the throw immediately before the exit label.

**Right place, and the fallthrough *is* the real hazard rather than a symptom.** Upstream definite-return analysis (GS0100) is intentionally not able to reject the target shape — a `switch` that is exhaustive over an enum's declared members but not over its runtime domain (`DateTimeKind(99)`). The compiler already had a guard for exactly this (the pre-existing `else if (!currentPositionEndsInTerminator)` throw); the two return funnels were bypassing it. This PR restores reachability of an existing guard rather than inventing a new policy. That is the correct shape of fix.

I checked the sharpest failure mode I could construct: the new `br` is emitted under `emitFallthroughGuard && this.fixedReturnLabel.HasValue`, while the label is marked inside `EmitFixedReturnEpilogue()`, which returns `false` **iff** `!fixedReturnLabel.HasValue` (`:278-292`). `EmitBlock(body)` has already run before the check, so the field cannot change between them. **Branch emitted ⟹ epilogue emitted ⟹ label marked.** No unmarked-label path. All three other combinations (`alwaysAppendRet`; no fixed label; body ends in a terminator) reduce to exactly the previous code.

## 2. Accepted-program space

**No program that compiled before fails to compile now** — the change is runtime-only. Verified that definite-return analysis is unweakened:

```gsharp
func NoRet(n int32) int32 { if n > 0 { return 1 } }
```
→ `GS0100` at **both** base and head. This corroborates the PR's claim that `preEmitAnalysisBody` is computed at `Lowerer.cs:1389`, before the insertion.

There **is** a runtime behaviour change — silent default → throw — and it is properly disclosed: the PR body states it, the ADR limitation text is replaced, and `Issue2906ExhaustiveSwitchReturnEmitTests.ExhaustiveEnum_UnnamedValueWithFixedReturn_ThrowsDefensively` is rewritten from asserting `0` to asserting the throw. That is the explicit test pin plus PR-body statement the process asks for. Correctly handled.

## 3. Pinning path intact

Not regressed. Same function, reachable arms:

| | base | head |
|---|---|---|
| `F(DateTimeKind.Unspecified, []int32{7})` | `7` | `7` |
| `F(DateTimeKind.Local, …)` | `3` | `3` |

`*p` through the pinned local still yields the element, so the pinned-local emit path is still taken and the guard did not suppress it. `ilverify` on the emitted assembly reports the **same 2 errors at base and head** (the pre-existing `fixed`/pointer `ExpectedNumericType` class the existing test already suppresses) — no new verification errors introduced.

Seven further shapes, each base vs head, all byte-identical in behaviour:

| shape | base | head |
|---|---|---|
| `try { return n*2 } finally { … }` | `fin`,`42` | `fin`,`42` |
| `return` out of a `finally` (ADR-0139 permits this) | `try`,`9` | `try`,`9` |
| `return` in a `catch` | `10` | `10` |
| `async func` + `try/finally` returning a value | `afin`,`42` | `afin`,`42` |
| `sequence[int32]` iterator + `try/finally` | `1`,`5`,`ifin` | same |
| void function + `try/finally` + early return | `vfin`,`vfin` | same |
| non-exhaustive function | `GS0100` | `GS0100` |

Async and iterators are the two shapes the PR's coverage list does not name, and they are the ones where an inserted throw before the exit label could plausibly have fired on a legitimate path. They do not.

## 4. Interpreter divergence — **SHOULD-FIX (two items)**

The `Lowerer` change is **not** emitter-only. `RewriteProtectedRegionEntries` returns `new BoundBlockStatement(emittedBody.Syntax, emittedBody.Statements, preEmitAnalysisBody)` (`:1418-1421`), and the interpreter evaluates `Statements` — the emitted body, which now contains the inserted throw. Measured:

```gsharp
    try {
        switch x {
            case DateTimeKind.Unspecified { return 1 }
            case DateTimeKind.Utc { return 2 }
            case DateTimeKind.Local { return 3 }
        }
    } finally { Console.WriteLine("fin") }
}
```

| | stdout | stderr | exit |
|---|---|---|---|
| `gsi` base | `fin`, `0` | — | 0 |
| `gsi` head | `fin` | `GS9999: Compiler-generated guard reached: non-void function fell through without returning a value.` | 1 |

**This is a genuine improvement** — `gsi` previously produced the same silent wrong answer as `gsc`, and now the two agree. But:

- **SHOULD-FIX (a): it is undisclosed and unpinned.** The PR body, the ADR edit and the test list are silent about the interpreter, and there is no `Interpreter.Tests` case. A behaviour change in a second consumer with no test is exactly what regresses later. Add one interpreter pin and one PR-body line.
- **SHOULD-FIX (b): the diagnostic is mis-categorised.** In `gsi` the guard surfaces as **GS9999**, which `docs/diagnostics.md:307` defines under *"Internal compiler error diagnostics (GS9998–GS9999)"* as "an unexpected exception was caught by the evaluator". A deliberate compiler-generated guard is the opposite of an unexpected internal error — a user hitting a real bug in their own code is told the compiler broke. This is the same mis-categorisation being addressed for the interpreter boundaries under #2956/#2986/#2988, and the ADR being written there should account for this case. Recommend coordinating through the merge coordinator rather than expanding this PR.

Neither item blocks: both consumers now behave consistently and correctly, and both changes are in the safe direction.

## 5. Contention

Production files touched: `src/Core/CodeAnalysis/Emit/MethodBodyEmitter.cs`, `src/Core/CodeAnalysis/Lowering/Lowerer.cs`.

| reserved file | owner | overlap |
|---|---|---|
| `Emit/MethodBodyEmitter.Statements.cs` | #2967 / #2964, #2965 queued | **none** — this PR touches `MethodBodyEmitter.cs`, a different file |
| `Evaluator.Statements.cs` | #2972, G5 ADR work | **none** |
| `Lowering/Iterators/IteratorRewriter.cs` | #3000, #2975 | **none** |
| `test/Compiler.Tests/IlVerifier.cs` | #2966 | **none** |

Across the whole open set, the only other PR touching either file is **#2903** (`MethodBodyEmitter.cs`). Natural-base `git merge-tree --write-tree --name-only pr2998 pr2903` → **clean**; `pr2998` × `origin/main` → **clean**.

One *semantic* (not textual) adjacency worth flagging: iterator `MoveNext` bodies are lowered through `Lowerer.Lower` (`IteratorMoveNextBodyBuilder.cs:150,259`, `AsyncIteratorRewriter.cs:65`), so this PR changes the tree that #3000 / #2975 / #2957 are also reshaping. My iterator probe passes at head today, but whichever of these merges second should re-run it.

## 6. Two non-blocking observations

- **The `Lowerer` guard is not `fixed`-specific.** It inserts a throw before `methodExitLabel` in *every* non-void function with protected-region returns, usually unreachable. Measured cost: the async `try/finally` assembly grew 3584 → 4096 bytes. Harmless, and the PR does disclose the visible consequence (two `refactoring-baseline.json` hashes, `ReturnInTryFinally.gs` and `YieldInTryFinally.gs`). But the PR title and the ADR sentence both say "fixed", which understates where the second hunk actually fires. Worth one clarifying sentence in the ADR.
- **The suite numbers in the PR body are mixed-base.** They were measured on a `b1e693bb`-based head (Compiler 3907, where that `main` was 3894) and the branch was then rebased onto `a6b1561f2`, where `main` is 3906. The reported figures therefore do not describe the SHA that will merge. Recommend re-running at `4b1461ced`. I attempted this and could not complete it: machine load reached 62.9 and the run stalled. Partial signal from what did execute — **14 failures, all `SampleConformanceTests.Sample_BuildsAndMatchesGolden`**, i.e. only the known `Gsharp.Extensions` build-closure gap, and **no failures in any other test class**, including the modified `Issue2906ExhaustiveSwitchReturnEmitTests`. Incomplete, so treat as a lower bound, not a pass.

## Summary

| severity | item |
|---|---|
| Blocking | none |
| Should-fix | interpreter behaviour change is undisclosed and has no `Interpreter.Tests` pin |
| Should-fix | guard surfaces in `gsi` as `GS9999`, documented as an *internal compiler error* |
| Nit | ADR/title say "fixed"; the `Lowerer` hunk fires for all protected-region returns |
| Nit | PR-body suite numbers are mixed-base; re-measure at `4b1461ced` |

The emitter hunk is a strict refinement of the previous logic, the lowering hunk preserves the analysis body, and I could not construct a shape — async, iterator, void, return-from-`finally`, return-from-`catch`, nested pinning — where the guard fires on a legitimate path.

DavidObando:
# Addendum — verdict stands, but two claims in it are now stale

Main moved from `a6b1561f2` to `a1237ded7` (six commits). Verdict stays **MERGEABLE**, but **#2964 merged as `5e616b9bb`** and it touched the same test method this PR rewrites. Two things in my review above are now wrong, and the rebase has a trap.

## 1. "Clean vs `main`" is falsified

Re-measured against current main:

```
$ git merge-tree --write-tree --name-only 4b1461ced origin/main
test/Compiler.Tests/Emit/Issue2906ExhaustiveSwitchReturnEmitTests.cs
CONFLICT (content): Merge conflict in .../Issue2906ExhaustiveSwitchReturnEmitTests.cs
```

`4b1461ced`'s natural merge base is still `a6b1561f2`. Production files are still conflict-free — the collision is entirely in the shared test.

## 2. The resolution is a **union**, and "take mine" silently reverts #2964

Both PRs edit `ExhaustiveEnum_UnnamedValueWithFixedReturn_*`, on adjacent lines.

**#2964 (now in main) changed exactly one line — the ignore list:**
```diff
-            ignoredVerificationErrors: new[] { "ExpectedNumericType" });
+            ignoredVerificationErrors: new[] { "UnmanagedPointer", "StackByRef" });
```

**This PR** renames the method, rewrites the source to try/catch, and flips `Assert.Equal(0, …)` → `Assert.Equal(7, …)` — while carrying `ignoredVerificationErrors: new[] { "ExpectedNumericType" }` as an **unchanged context line**.

That is the trap: the contested line is *context* in this PR's diff, so it looks untouched during review, but resolving the conflict in this PR's favour **reverts #2964's ignore-list fix**. The stale list then either fails on unignored `UnmanagedPointer`/`StackByRef`, or suppresses a class of error that no longer occurs.

**Correct merged form — #2964's ignore list, this PR's assertions:**
```csharp
public void ExhaustiveEnum_UnnamedValueWithFixedReturn_ThrowsDefensively()
    …try/catch source from this PR…
    ignoredVerificationErrors: new[] { "UnmanagedPointer", "StackByRef" });   // from #2964
    Assert.Equal(7, GetField(assembly, "result"));                            // from this PR
```

Also delete the `// Known limitation (#2953): …silently return the default slot instead of throwing.` comment — #2964 preserved it, this PR makes it false.

## 3. No production interaction

Checked, because both PRs are `fixed`-adjacent and that would be the dangerous case. #2964 is confined to pinning/null-check labels in `EmitFixedStatement` (`MethodBodyEmitter.Statements.cs`, `skipLabel`→`nullLabel`/`afterLabel`); it does **not** touch `fixedReturnLabel`, `EmitFixedReturnEpilogue`, or `methodExitLabel`. Different file, different machinery from this PR's `MethodBodyEmitter.cs` hunk. The guard logic is unaffected.

## 4. Re-run the differential after rebase — my base has moved under the probe

My verdict's differential and `ilverify` result were measured at `a6b1561f2`, i.e. **before #2964 changed the emitted IL for `fixed` pointer conversion**. My repro is `fixed p *int32 = values { return *p }` — precisely the construct #2964 re-emits. So:

- the "same 2 `ilverify` errors at base and head" line no longer describes the new base (#2964's entire point is that the error set is `UnmanagedPointer`/`StackByRef`);
- the behavioural differential should be re-observed on the rebased head before merge.

I expect both to hold — the machinery is disjoint — but expectation is not measurement, and this is a case where the base moved under the exact shape probed.

## 5. Unchanged from the verdict

The two SHOULD-FIX items are untouched by any of this: the `Lowerer` hunk still reaches the **interpreter** (`gsi` base exit 0 printing `0` → head exit 1 `GS9999`), still undisclosed, still with no `Interpreter.Tests` pin; and it still surfaces as **GS9999**, which `docs/diagnostics.md:307` defines as an internal compiler error.

One correction to my own contention table: `test/Compiler.Tests/IlVerifier.cs` is **no longer reserved** (#2966 merged as `87ca66e85`), so that row is moot.

The mixed-base suite numbers now need re-measuring against `a1237ded7`, not `a6b1561f2`. Note that the 19 `SampleConformanceTests` failures I reported as a known closure gap should now be **zero** on a solution build, since #3002 (`a3cdc4e39`) made that safe — so a clean re-run is both cheaper and more meaningful than when I attempted it.

DavidObando:
# Final verdict — measured on the resolved merge. **MERGEABLE.**

My earlier verdict rested on textual + static evidence and I said so. That gap is now closed with a built merge. Head has also moved since: **`4b1461ced` → `d9b3a736`**, rebased onto `a1237ded7`.

Note for anyone diffing this PR: `git diff 4b1461ced d9b3a736` shows **25 files / +2854** and is a two-dot artefact of the rebase, not authored content. The real content is `git diff a1237ded7...d9b3a736` — **6 files, +448/-16**, the same shape I reviewed.

## The conflict resolution is correct

I flagged that resolving in this PR's favour would silently revert #2964, because `ignoredVerificationErrors:` appears as an **unchanged context line** in this PR's diff. The author resolved it in the union form — verified at `d9b3a736:…Issue2906ExhaustiveSwitchReturnEmitTests.cs:321`:

```csharp
ignoredVerificationErrors: new[] { "UnmanagedPointer", "StackByRef" });   // #2964, preserved
Assert.Equal(7, GetField(assembly, "result"));                            // this PR
```

`ExhaustiveEnum_UnnamedValueWithFixedReturn_ThrowsDefensively` renamed, try/catch source in place, and the now-false `// Known limitation (#2953)` comment deleted. Nothing reverted.

`git merge-tree --write-tree --name-only d9b3a736 origin/main` against **`a692200af`** → tree oid only, **no conflict**.

## Built the merge

Merged tree `5f1fd4883` = `a692200af` + this PR, delta exactly the 6 files.

| gate | result |
|---|---|
| exit code (`dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true`) | **0**, 55.7 s |
| race signature `MSB302[67]\|MSB3883` | **0 hits** |
| zero-byte over **all 40** `GSharp.Core.dll` copies | none |
| md5 over **23 ordinary impl** copies | **1 group**, `c64748da…` |
| `ref`/`refint` across **all** families incl. both SDK staging instances | **1 group**, `bb86c90e…` |

The naive "every copy agrees" form reports **4** groups here and that is the known post-#3002 false positive: the extra groups are `gsc-publish`/`gsgen-publish` and the `*-staging/bin/` trees, which are legitimately separate instances. The ref assemblies being byte-identical across every family is the semantic-identity proof.

This also answers the #3000 hazard — textual mergeability is not compile mergeability. **This merge compiles.** Corroborating static check: #2964, #2929, #2980 and #2981 removed **no** member declarations, so there is no deleted-consumer vector.

## Tests — narrow filter, no full suite, no `--no-build`

```
--filter FullyQualifiedName~Issue2953FixedReturnFallthroughGuardTests
        |FullyQualifiedName~Issue2906ExhaustiveSwitchReturnEmitTests
Passed! - Failed: 0, Passed: 31, Skipped: 0, Total: 31
```
Impl md5 unchanged across the rebuild (`c64748da…`), confirming a deterministic build rather than a skipped one.

## Differential re-measured on the new base — this was necessary, not ceremonial

My original differential was taken at `a6b1561f2`, **before #2964 changed `fixed` pointer conversion to route managed byrefs through `Unsafe.AsPointer<T>`**. My probe is `fixed p *int32 = values { return *p }` — precisely the construct #2964 re-emits, so the base had moved underneath it. Re-run against `a692200af` vs the merged tree:

| probe | base `a692200af` | merged `5f1fd4883` |
|---|---|---|
| `F(DateTimeKind(99), …)` | prints `0`, **exit 0** | `InvalidOperationException: Compiler-generated guard reached…`, **exit 134** |
| `F(DateTimeKind.Unspecified, …)` / `F(…Local, …)` | `7`, `3` | `7`, `3` |

The guard fires on the new emit, and **the pinning path is intact under `Unsafe.AsPointer`** — that is the one result that genuinely could have changed and did not.

**`ilverify` corrected.** My verdict said "the same 2 errors at base and head", true then but describing the pre-#2964 error set. Re-measured: **`UnmanagedPointer` + `StackByRef`, identical on both sides** — matching #2964's new ignore list exactly, and confirming this PR introduces no new verification errors. The old `ExpectedNumericType` no longer occurs.

## #2981 constraint — checked, not assumed

#2981 flipped `DiagnosticIdUniquenessTests` from opt-in to mandatory, so a PR adding a descriptor without a docs row (or a duplicate row) now fails only *after* merge, since branch CI ran against the opt-in version. `git diff --name-only a1237ded7...d9b3a736 | grep -i diagnostic` → **empty**. This PR adds no descriptor and no docs row. **Moot**, confirmed rather than assumed.

## What still stands

Both SHOULD-FIX items are unchanged and unaddressed:

1. The `Lowerer` hunk reaches the **interpreter** — `RewriteProtectedRegionEntries` returns `emittedBody.Statements`, which the evaluator runs. `gsi` goes from exit 0 printing `0` to exit 1 with `GS9999`. Still undisclosed in the PR body, still **no `Interpreter.Tests` pin**.
2. It surfaces as **GS9999**, which `docs/diagnostics.md` defines as an internal compiler error, for what is a deliberate guard. Coordinate with the #2956/#2986/#2988 boundary ADR rather than expanding this PR. Worth noting there that **GS0506 `FixedStatementCannotSuspend` (#2917)** is existing prior art for a `fixed`-boundary diagnostic.

Neither blocks. Both consumers now behave consistently, and both changes are in the safe direction.

**Verdict: MERGEABLE**, now on measured evidence — resolved merge builds clean, 31/31 targeted tests pass, differential and pinning re-confirmed against the post-#2964 emit.

DavidObando:
# Re-review at `cb6088263` — **the push was a pure rebase. Neither SHOULD-FIX was addressed.**

I was asked to confirm that fixes had landed for both items. They have not. The code is unchanged from the head I already cleared.

## Provenance: zero PR content change

Head `d9b3a736` → `cb6088263`, now based directly on `f70626452` (= `main`), so `git merge-base origin/pr2998 origin/main` = `f70626452`.

True content `git diff --stat f70626452...cb6088263` → **6 files, +448/-16** — the same six files and the same stat I reviewed at `d9b3a736`. Per-file blob comparison between the two heads:

| file | |
|---|---|
| `docs/adr/0139-goto-and-general-labels.md` | IDENTICAL |
| `src/Core/CodeAnalysis/Emit/MethodBodyEmitter.cs` | IDENTICAL |
| `src/Core/CodeAnalysis/Lowering/Lowerer.cs` | IDENTICAL |
| `test/…/Issue2906ExhaustiveSwitchReturnEmitTests.cs` | IDENTICAL |
| `test/…/Issue2953FixedReturnFallthroughGuardTests.cs` | IDENTICAL |
| `test/Core.Tests/Baselines/refactoring-baseline.json` | IDENTICAL |

All six blob SHAs match. One commit on the branch, same subject. No file added, no `Interpreter.Tests` file, and `git diff f70626452...cb6088263 | grep -E "GS9999|GS0506|DiagnosticDescriptor"` → **empty**.

(The two-dot `d9b3a736..cb6088263` reports 7 files/+459/-19 and is rebase noise. That trap has now fired four times on this PR.)

## SHOULD-FIX 1 — still open, and now the PR body obscures it

**Measured at the new head**, `gsi` on a switch inside a protected region:

```gsharp
    try {
        switch x {
            case DateTimeKind.Unspecified { return 1 }
            case DateTimeKind.Utc { return 2 }
            case DateTimeKind.Local { return 3 }
        }
    } finally { Console.WriteLine("fin") }
}
```

| | stdout | exit |
|---|---|---|
| base `f70626452` | `fin`, `0` | **0** |
| head `cb6088263` | `fin`, `GS9999: Compiler-generated guard reached: non-void function fell through without returning a value.` | **1** |

This PR still changes interpreter behaviour, via `Lowerer.cs:1418-1421` — `RewriteProtectedRegionEntries` returns `emittedBody.Statements`, which the evaluator runs. There is **no `Interpreter.Tests` pin**: `grep -rl "Compiler-generated guard reached" test/` returns only the two `Compiler.Tests` files.

The body does now mention the interpreter, at line 55, citing **#3006**. That note is about a **different shape** and, read as coverage of this finding, is misleading:

> "direct exhaustive enum-switch statement fallthrough still returns `0` under `gsi` … This is outside the CIL epilogue path fixed here."

Measured, both shapes, same binaries:

| shape | base | head |
|---|---|---|
| bare switch (**#3006**) | exit 0, `0` | exit 0, `0` — **unchanged**, correctly out of scope |
| switch in `try`/`finally` (**this finding**) | exit 0, `0` | **exit 1, `GS9999`** — changed by this PR |

So #3006 is correctly filed and genuinely separate, but it does not cover this. A reader of the body concludes the PR leaves `gsi` alone; it does not. Combined with zero interpreter tests, that is precisely the silent-revert exposure I raised: the next PR to touch `RewriteProtectedRegionEntries` can revert this with nothing failing.

## SHOULD-FIX 2 — still open

The guard still surfaces in `gsi` as **`GS9999`**, the internal-compiler-error bucket, for a deliberate designed refusal. No descriptor was added and `GS0506` was not reused — the PR touches no diagnostic file at all. To be explicit, since the instruction was to reuse rather than allocate: **no new ID was allocated**, so there is no regression against that instruction; the item is simply untouched.

## Everything I previously cleared still holds

Re-measured at `cb6088263` vs `f70626452`, freshly built both sides (`-p:ContinuousIntegrationBuild=true`, exit 0 each, Core md5 `26064b6b…` base / `cf3526b2…` head — distinct, so genuinely different binaries).

| check | base | head |
|---|---|---|
| `F(DateTimeKind(99), …)` compiled | `0`, exit 0 | throws, **exit 134** |
| `F(…Unspecified, …)` / `F(…Local, …)` pinning | `7`, `3` | `7`, `3` |
| union resolution `:321` | — | `["UnmanagedPointer","StackByRef"]` kept |
| `Assert.Equal(7, …)` `:322` | — | present |
| stale `// Known limitation (#2953)` | — | **absent** (grep count 0) |
| targeted tests | — | **31/31 passed**, exit 0, no timeouts |

The compiler-side differential holds and the pinning path is intact under #2964's `Unsafe.AsPointer` emit.

**#2981 docs gate: moot, checked not assumed.** `git diff --name-only f70626452...cb6088263 | grep -cE "DiagnosticDescriptors|diagnostics.md"` → **0**. The PR adds no descriptor and no docs row, so the branch's green CI is not load-bearing here.

## Verdict

**MERGEABLE** — unchanged from my previous verdict, on the same merits. Nothing regressed, the differential holds at the new head, and 31/31 targeted tests pass.

But the two SHOULD-FIX items are **both still open**, and one has become slightly worse in presentation:

| severity | item | location |
|---|---|---|
| SHOULD-FIX | interpreter behaviour change still has **no test pin** | `src/Core/CodeAnalysis/Lowering/Lowerer.cs:1418-1421`; missing pin in `test/Interpreter.Tests/` |
| SHOULD-FIX | body line 55 cites #3006 for a **different shape**, implying `gsi` is unaffected when the protected-region shape is | PR body:55 |
| SHOULD-FIX | guard reports as `GS9999` (internal compiler error) for a deliberate refusal; `GS0506` `FixedStatementCannotSuspend` (#2917) is reusable prior art | `docs/diagnostics.md` category, no new ID needed |

None of these blocks merge. All three are cheap: one `Interpreter.Tests` case and one sentence in the body would close the first two outright.

DavidObando:
## Review: NOT MERGEABLE — the fix itself is good and genuinely pinned, but it no longer merges and two of its hunks now duplicate a mechanism that landed on `main`

Reviewed at head `6e9ad3a3`, `main` = `fd34166a8`, **computed merge-base `f70626452`** (three-dot, freshly fetched). Net at tip: **12 files, +544 −19**. Branch is **14 commits behind**.

State re-established from scratch; I inherited no prior notes.

### The substance is right, and both production hunks are properly pinned

Built at head (rc=0, 0 errors), clean Core md5 `118ac8cdfd8527447f163faf20a45891`.

**Values, `gsc` vs `gsi` vs truth** — a bypassed fall-through guard is a silent-wrong-value shape, so I read values, not status:

| probe | `gsi` | `gsc` (compiled, run) | truth |
|---|---|---|---|
| non-void `try`/`finally`, three real returns | rc 0, `11 22 33` | rc 0, `11 22 33` | `11 22 33` |
| exhaustive enum switch, unnamed value falls through | rc 1, `GS0100: Compiler-generated guard reached…` | rc **134**, `InvalidOperationException: Compiler-generated guard reached…` | must not return a value |
| `return` out of `fixed`, non-void ×3 + void | *(interpreter refuses `fixed`, see NIT)* | rc 0, **`11 22 33 11`** | `11 22 33 11` |

No `gsc`/`gsi` divergence on any shape either can run. The 134 was captured unpiped with the exit read before anything else executed.

**Mutation — I reverted each production hunk individually.**

| hunk reverted | Core md5 | result |
|---|---|---|
| `MethodBodyEmitter.cs:244-266` (fixed-return epilogue / guard interaction) | `84346c7c` | **6 of 31 failed** — `ExhaustiveEnum_UnnamedValueWithFixedReturn_ThrowsDefensively`, plus `NestedFixed`, `FixedInsideLoop`, `SwitchInsideFixed`, `SealedPatternFixed` |
| `Lowerer.cs:1389-1416` (guard insertion before `methodExitLabel`) | `871b0738` | **1 of 1 interpreter test failed**, and the probe silently regressed |

Both builds were rc=0 with **0 errors**, so neither is a bad mutant. Cycle closed: `118ac8cd` → mutants → `118ac8cd`, behaviour flipping `GS0100` → *(regression)* → `GS0100`, tree clean.

The Lowerer mutation is the one worth showing, because it reproduces the bug this PR exists to fix in its true, dangerous form:

```
head:                gsi fall.gs  → rc=1  GS0100: Compiler-generated guard reached: …
Lowerer hunk removed: gsi fall.gs → rc=0  0
```

**`0`, exit code 0** — a non-void function returning a value it never computed. That is the silent-wrong-answer class, and this PR closes it. Good work; this is materially better covered than most of what I have reviewed this session.

It also does the right thing on diagnostics: it reuses the existing `DiagnosticDescriptors.AllPathsMustReturn` (`GS0100`, one parseable row in `docs/diagnostics.md`) instead of inventing an ID or leaving a designed refusal on `GS9999`. **No allocation needed, docs gate satisfied.**

---

## BLOCKER 1 — does not merge onto current `main`

`git merge-tree --write-tree` against `fd34166a8` exits **1** with three content conflicts:

- `src/Core/CodeAnalysis/Compilation/Compilation.cs`
- `src/Core/CodeAnalysis/EvaluatorException.cs`
- `test/Compiler.Tests/Emit/Issue2906ExhaustiveSwitchReturnEmitTests.cs`

## BLOCKER 2 — the `DiagnosticId` plumbing already exists on `main`, with a better API

This is why the first two files conflict, and a plain rebase will not resolve it correctly.

`main` already carries the whole mechanism:

```csharp
// EvaluatorException.cs (main)
internal string DiagnosticId { get; private set; } = "GS9999";
internal DiagnosticSeverity Severity { get; private set; } = DiagnosticSeverity.Error;
… DiagnosticId = descriptor.Id, Severity = descriptor.Severity,

// Compilation.cs (main)
var diagnostic = new Diagnostic(location, ex.DiagnosticId, ex.Severity, message);
```

Against that, this PR's two hunks are not just redundant but a regression in API quality:

1. `EvaluatorException.cs:64-75` adds a **`public`** constructor taking a raw `string diagnosticId` and a **`public`** `DiagnosticId` property. `main` deliberately keeps both `internal` and drives them from a `DiagnosticDescriptor`, so severity cannot drift from the descriptor. The string overload reintroduces exactly that drift and widens the public surface.
2. `Compilation.cs:334-338` adds `ex.DiagnosticId ?? "GS9999"`. On `main` the property **already defaults to `"GS9999"`**, so this hunk is dead on arrival, and it also drops `ex.Severity`, which `main` now passes.

**Recommended resolution:** on rebase, take `main`'s side of both files wholesale and delete these two hunks. Then have `EvaluateThrowStatement` (`Evaluator.Statements.cs:656-664`) construct the exception through `main`'s descriptor-based factory, passing `DiagnosticDescriptors.AllPathsMustReturn` rather than an ID string. `BoundThrowStatement.DiagnosticId` (`:17-33`) should then carry the descriptor instead of a `string`, which also removes the stringly-typed field from the bound tree.

That change should be behaviour-preserving — my probes above pin the observable outcome (`GS0100`, rc 1) on one side and `0`, rc 0 on the other, so re-running `fall.gs` after the rebase is a sufficient check.

## SHOULD-FIX — confirm the ADR edit against #3018

`docs/adr/0139-goto-and-general-labels.md` is touched. That is **not** ADR-0125, so the specific collision I was asked to look for is not present, and no ADR is being added or renumbered here, so there is no repeat of the duplicate-`0152` failure. But #3018 is concurrently renumbering ADRs; please confirm `0139` is not among the numbers it moves, since a renumber plus this edit would conflict silently in the same way.

## NIT — the neighbouring designed refusal is still on `GS9999`

`src/Core/CodeAnalysis/Evaluator.Statements.cs:158`

```
```

That is a deliberate, designed refusal reported through the internal-error bucket — which is why the `fixed` row in my table above has no `gsi` column. It is **pre-existing** (present on `main`, 0 hits in this diff), so it is out of scope and I am not asking for it here. Flagging only because this PR builds precisely the machinery that fixes it, and `GS0506 FixedStatementCannotSuspend` is adjacent prior art. Worth a follow-up issue rather than scope creep.

---

**Verdict: NOT MERGEABLE** — on mergeability, not on quality. The fix is correct, both hunks are genuinely pinned, and the silent `0`/rc-0 regression it prevents is exactly the class this repo keeps shipping. It needs a rebase onto `fd34166a8` in which `Compilation.cs` and `EvaluatorException.cs` take `main`'s side and the duplicated plumbing is dropped in favour of the descriptor-based API that landed while this was open.

Issue #2995, opened 2026-08-01
gsi: Optional.FlatMap with a nullable-returning lambda throws TargetInvocationException


DavidObando:
## Problem

`Optional.FlatMap` with a lambda returning a nullable throws in the interpreter.

`samples/GsharpExtensionsOptional.gs`:

```gsharp
let upper = name.Map(func(s string) string { return s.ToUpper() })   // works, prints ADA
let firstChar = upper.FlatMap(func(s string) string? {               // throws
    if s.Length > 0 { return s.Substring(0, 1) }
    return nil
})
```

```
compiled: ADA A default computed <filtered out> present: ada ada 14 -1 7 count present: 7
gsi:      ADA
          GS9999: Exception has been thrown by the target of an invocation.
```

`Map` succeeds and `FlatMap` fails, so the extension-method dispatch itself works — it is specifically the nullable-returning lambda.

## Notes

The message is a bare `TargetInvocationException` with the inner exception discarded, which is itself unhelpful — surfacing `InnerException` would make this and similar failures diagnosable. That is closely related to #2938 (interpreter loses uncaught delegate exception diagnostics through CLR member calls) and may share a fix.

Found while triaging #2939; excluded there until fixed. The sibling samples `GsharpExtensionsMixed.gs` and `GsharpExtensionsSequences.gs` both pass, so this is not a general extensions-resolution problem.

DavidObando:
## Triage: this is two defects, and neither one is `FlatMap`, nullables, or extension dispatch

Measured with a direct `Compilation.Evaluate` probe (the same entry point `gsi` drives) against builds of `8dddbbf2` (base) and `f816c794` (PR #2972 head), Core DLL identity md5-verified in each probe output directory.

### 1. The diagnostic half **is** #2938, and PR #2972 fixes it — measured, not inferred

Reduced repro with **no** `Gsharp.Extensions` dependency, same shape (lambda passed across a CLR reflection boundary):

```gsharp
import System
import System.Linq

let xs = []string{"ada", "bob"}
let ys = xs.Select(func(s string) string? {
    if s.Length > 0 {
        return s.Substring(0, 1)
    }
    return nil
})
Console.WriteLine(ys.Count())
```

| build | output |
|---|---|
| base `8dddbbf2` | `GS9999: Exception has been thrown by the target of an invocation.` |
| head `f816c794` (#2972) | `GS9999: Unexpected node IfStatement` |

[...]

PR #3000, opened 2026-08-01
Lower yield in function-literal iterator bodies

### LINKED ISSUE #2957: gsc: lower yield in function-literal iterator bodies

### LINKED ISSUE #2976: gsi: yield in function-literal iterator bodies is rejected as outside an iterator
## Summary

Issue #2957 covers `gsc` failing to lower `yield` inside function-literal iterator bodies. The interpreter has the same defect from a separate cause, and should be tracked separately if #2957's fix is scoped to the compiler.

## Repro

A named iterator function prints `2`. The same iterator written as a function literal fails:

- `gsc`: `GS9998 … yield reached the emitter before iterator lowering`
- `gsi`: `GS9999: 'yield' encountered outside of an iterator function`

## Root cause

**Compiler** (`#2957`): `Lowering/Iterators/IteratorRewriter.cs:31` iterates `program.Functions`, a flat map. Function-literal bodies are not entries in it when `Compilation.cs:676` runs `IteratorRewriter.Rewrite(program)`, so `ContainsYield` never sees them and no state-machine plan is produced.

**Interpreter** (this issue): `Evaluator.Calls.cs:188` never marks a function-literal body as an iterator body, so `yield` is rejected as being outside an iterator function.

## Note on the likely fix

The repo already has `Lowering/NestedFunctionBodyRewriter.cs:27`, the base class `ExpressionTreeLowerer`, `InterpolatedStringHandlerLowerer` and `SideEffectSpiller` use to descend into nested function bodies. `IteratorRewriter` and `AsyncIteratorRewriter` do not use it. Extending that existing descent looks preferable to adding a new hoisting pass.

If a single fix naturally covers both backends, this can be closed as part of #2957.

### LINKED ISSUE #3003: gsi: an 'if' statement inside a lambda body throws 'Unexpected node IfStatement'
## Problem

An `if` statement inside a **lambda / `func` literal body** hard-fails in the interpreter. The identical `if` inside a top-level `func` declaration works.

```gsharp
package P
import System

var f = func(n int32) {
    if n > 0 {
        Console.WriteLine("pos")
    }
}
f(1)
```

```
gsc → pos
gsi → GS9999: Unexpected node IfStatement
```

Control pair — same `if`, moved out of the lambda:

```gsharp
func g(n int32) {
    if n > 0 {
        Console.WriteLine("pos")
    }
}
g(1)
```

```
gsc → pos
gsi → pos          ✅
```

## Measured matrix (`a6b1561f2`)

| shape | `gsc` | `gsi` |
|---|---|---|
| `if` + `return`, value-returning lambda | `pos` / `nonpos` | **`GS9999: Unexpected node IfStatement`** |
| `if`, no return, void lambda | `pos` | **same** |
| `if` / `else` in lambda | `pos` | **same** |
| lambda declared inside a top-level `func` | `lf-pos` | **same** |
| lambda reached through a reflective CLR boundary (`Array.FindAll`) | `2` | **`GS9999: Exception has been thrown by the target of an invocation.`** |
| — control — same `if` in a top-level `func` | `pos` | `pos` ✅ |
| — control — `for` loop in a lambda | — | `6` ✅ |
| — control — `while` loop in a lambda | — | `4` ✅ |

The `for`/`while`-work-but-`if`-fails pairing is the diagnostic: this is not "lambdas can't do control flow", it is specifically the one statement kind the binder leaves unlowered.

## Root cause

`BoundTreeRewriter` **deliberately** does not descend into `BoundFunctionLiteralExpression.Body` — `NestedFunctionBodyRewriter.cs:16-21`, verbatim:

> `BoundTreeRewriter` intentionally does not recurse into `BoundFunctionLiteralExpression.Body` or generic local-function bodies because many callers model lexical scopes.

`Lowerer` is a plain `BoundTreeRewriter`, not a `NestedFunctionBodyRewriter`, so **lambda bodies are never lowered**.

`Evaluator.Expressions.cs:762` then evaluates the raw bound body:

```csharp
return MaterializeClosure(new ClosureValue(node.Function, node.Body, node.FunctionType, captured));
//                                                        ^^^^^^^^^ unlowered
```

`EvaluateStatement`'s switch (`Evaluator.Statements.cs:62-175`) only understands the **lowered** node set, and throws at `Evaluator.Statements.cs:168`.

Exactly four statement kinds are unhandled by the evaluator: `IfStatement`, `ForInfiniteStatement`, `ForEllipsisStatement`, `ForRangeStatement`. Only `IfStatement` is live, because the binder emits `BoundIfStatement` directly (`StatementBinder.Conditionals.cs:50`) while no binder site constructs the loop nodes — loops arrive already goto-shaped. Hence `for`/`while` work.

**The asymmetry is one line wide.** The sibling method `EvaluateMethodGroupExpression` (`Evaluator.Expressions.cs:781`) reads `program.Functions[node.Function]`, whose bodies **are** lowered at `Binder.cs:1383` / `:1442`. Same evaluator, same statement switch — one path gets lowered input, the other does not.

## Fix sketch

Lower the literal body at `Evaluator.Expressions.cs:762`, matching what `Binder.cs:1383` already does for `program.Functions`:

```csharp
Lowerer.Lower(node.Body)   // instead of node.Body
```

`Lowerer.Lower(literal.Body)` on a lambda body is an established in-repo pattern — `RefStructAsyncLivenessAnalyzer.cs:658` and `LambdaBinder.cs:1307` both do exactly this. Return semantics are safe by construction: the evaluator already executes `Lowerer.Lower` output for every top-level function.

Cache the lowered body on the bound node so a literal evaluated inside a loop is not re-lowered per iteration.

Alternative, closer to the stated purpose of `NestedFunctionBodyRewriter`: give the interpreter's program-construction pass the same shared descent, so nested bodies are lowered once up front.

## Impact

- **Any lambda containing an `if` fails under `gsi`.** Common shape.
- `samples/VariadicDelegate.gs:16` has one, so that sample stays red **after** #2985's fix. Not a regression of #2985 — this bug.
- **Likely the true root cause of #2995.** Its `FlatMap` lambda contains an `if`; its `Map` lambda does not, and `Map` is the one that works. Reproduced the exact reported symptom with **no nullable anywhere**:

```gsharp
var p Predicate[int32] = func(n int32) bool {
    return n > 1
}
Console.WriteLine(Array.FindAll[int32](arr, p).Length)     // gsi → 2                      ✅

var p Predicate[int32] = func(n int32) bool {
    if n > 1 {
        return true
    }
    return false
}
Console.WriteLine(Array.FindAll[int32](arr, p).Length)     // gsi → GS9999: Exception has
                                                           //   been thrown by the target
                                                           //   of an invocation.          ❌
```

  Reflective invocation wraps the `EvaluatorException` in `TargetInvocationException`, whose inner is then discarded — that is #2938 / PR #2972 territory, but fixing the unwrap improves only the *message*; the failure stays.
- Blocks #2939's interpreter conformance sweep for every sample whose lambdas branch.

## Files a fix would touch

| file | region |
|---|---|
| `src/Core/CodeAnalysis/Evaluator.Expressions.cs` | **`:762`** (+ small cache field) |
| `test/Interpreter.Tests/<new>.cs` | `if` / `if-else` / nested-in-func / via-reflective-boundary |

`:762` is outside the `Expressions.cs:829-880` and `Expressions.cs:1146…1597` regions currently allocated to other in-flight interpreter work, so this can run concurrently with both.

## Provenance

SHA-pinned `git clone --shared` scratch clone; `out/bin/Release/Repl/gsi.dll` and `out/bin/Release/Compiler/gsc.dll` invoked directly. Tree carries `a6b1561f2`: `merge-base --is-ancestor a6b1561f2 HEAD` true **and** `git diff a6b1561f2 HEAD -- src/Core/CodeAnalysis/Evaluator.Expressions.cs src/Core/CodeAnalysis/Evaluator.Statements.cs src/Core/CodeAnalysis/Lowering/` empty. Found while triaging #2985.

DavidObando:
## Summary

- collect iterator plans from nested function-literal bodies
- lower each emitted function-literal body once, then build its iterator plan from that exact body instance
- retarget capture-bearing plans to synthesized closure `Invoke` methods
- preserve generic lambda/state-machine remaps, with state-machine remaps taking precedence only for method type parameters
- route interpreter closure calls through existing user-function iterator classification
- cover lowering-synthesized iterator locals and named iterators invoked through method-group delegates

Fixes #2957
Fixes #2976

## Root cause and blocker fix

`IteratorRewriter` originally scanned only `BoundProgram.Functions`, so hosted function-literal bodies had no iterator plan. The first version of this PR collected those bodies before lowering, while emission later lowered them separately.

That ordering omitted lowering-synthesized locals from the hoisting field map. A `for-in` enumerator therefore became `null` after the first suspension, and a `for-ellipsis` counter reset so the iterator silently stopped after its first element.

`ReflectionMetadataEmitter` now stores the canonical lowered body before state-machine synthesis. `RetargetIteratorPlans` rebuilds every literal iterator plan—capturing and non-capturing—from the same body instance that emission uses. This is important because lowering twice creates different local-symbol identities.

The interpreter has the mirror stage-ordering defect in #3003: its lambda body is never lowered. This PR fixes iterator classification only; it does not close #3003 or #2995.

## Base-to-head differential

Measured from the true PR base `b1e693bb1`:

| Probe | Base | Head |
|---|---|---|
| function-literal iterator with `for n in []int32{4,5}` | compile error `GS9998` | exit 0, `4\n5\n` |
| function-literal iterator with `for i in 0 ... 2` | compile error `GS9998` | exit 0, `0\n1\n` |
| named iterator through method-group delegate in `gsi` | `GS9999: 'yield' encountered outside of an iterator function` | exit 0, `7\n8\n` |

The pre-fix PR head reproduced the regressions exactly: `for-in` printed `4` then threw `NullReferenceException`; `for-ellipsis` exited 0 after printing only `0`.

## Generic remap precedence

The precedence changes in `SignatureEncoder` and `ImportedMemberRefFactory` apply whenever lambda-method and class/state-machine remaps are simultaneously active. State-machine `VAR` wins for method type parameters used by the synthesized state machine. Lambda `MVAR` remains next for class type parameters, followed by the existing class-remap fallback for synthesized closures, capture boxes, and nested generic types.

`ImportedMemberRefFactory` now has the same `IsMethodTypeParameter` guard and fallback shape. This preserves the nested-generic remaps added by #2978 while covering the narrower iterator-literal overlap.

Coverage:

- `GenericStateMachineSignatureRemap` kills a lambda-first `SignatureEncoder` mutant.
- `ImportedElementToken_MethodTypeParameter_PrefersStateMachineRemap` kills a lambda-first `ImportedMemberRefFactory` mutant.
- `ImportedElementToken_ClassTypeParameter_UsesLambdaRemap` kills removal of the new guard.

## Added coverage

Compiler cases:

- `LoweredForInEnumerator`
- `LoweredForInEnumeratorWithCapture`
- `LoweredForEllipsisCounter`
- `GenericStateMachineElementTokenRemap`
- renamed `GenericStateMachineSignatureRemap` to state its remap purpose

Core cases:

- `ImportedElementToken_MethodTypeParameter_PrefersStateMachineRemap`
- `ImportedElementToken_ClassTypeParameter_UsesLambdaRemap`

Interpreter coverage adds the named-iterator method-group delegate path.

Every compiler case runs `IlVerifier.Verify`, `Assembly.Load(File.ReadAllBytes(...))`, `GetTypes()`, and bounded child execution with concurrent reads of stdout and stderr.

## Rebase and scope

Rebased onto `main` at `87ca66e85`, then rebased again to `f70626452` after `main` advanced during verification. The first merge CI exposed #2978's `SourceTypeParameters` rename and a class-remap precedence interaction; both are incorporated in the final head. The `CheckAllPathsReturn` interaction from #2961 and nested-generic iterator coverage from #2978 pass together.

#2975 remains open. `NestedIteratorCollector` does not change `CollectHoistedLocals` or make `LocalCollector` reference-based. The restructuring moves that work from the old line 158 area to `LocalCollector` near line 187 (roughly line 174 before comment drift).

#3003 and #2995 remain open. Their interpreter lambda bodies are still unlowered.

#2976 has no independent compiled specification until #2957 lands because base `gsc` rejects function-literal `yield` with `GS9998`. Its backend-specific coverage is therefore unit-test-only; no golden compiled sample can pin it separately.

## Verification

Final solution build: 0 warnings, 0 errors. `GSharp.Core.dll` was non-empty and byte-identical across Compiler, Core.Tests, Compiler.Tests, and Interpreter.Tests (`9d2e0b1405e4692e9ac669ddd2f62bcd`).

| Suite | Result |
|---|---:|
| Core | 7033 passed, 1 skipped |
| Compiler | 3959 passed |
| Interpreter | 499 passed |
| Language Server | 452 passed |

Targeted rebase interaction:

- Core iterator/remap tests: 20/20
- Compiler #2921 + #2951 + #2957 tests: 40/40
- Interpreter #2921 + #2951 + #2976 tests: 9/9

The lowered-plan mutant produced the original `NullReferenceException` and silent truncation; restoring the fix returned `4\n5\n` and `0\n1\n`.

Final CI rollup: 8 successful checks and 2 skipped checks.

### DIFF
--- a/src/Core/CodeAnalysis/Emit/GenericRemapState.cs
+++ b/src/Core/CodeAnalysis/Emit/GenericRemapState.cs
@@ -233,6 +233,11 @@ internal sealed class GenericRemapState
         this.lambdaMethodTypeParamRemapsByFunction[function] = remap;
     }
 
+    internal bool TryGetLambdaMethodRemap(
+        FunctionSymbol function,
+        out Dictionary<TypeParameterSymbol, int> remap)
+        => this.lambdaMethodTypeParamRemapsByFunction.TryGetValue(function, out remap);
+
     internal readonly struct SmRemapScope : IDisposable
     {
         private readonly GenericRemapState owner;
--- a/src/Core/CodeAnalysis/Emit/ImportedMemberRefFactory.cs
+++ b/src/Core/CodeAnalysis/Emit/ImportedMemberRefFactory.cs
@@ -279,6 +279,7 @@ internal sealed class ImportedMemberRefFactory
             // active remap (class or method) maps to the synthesized class's
             // own VAR(idx) slot.
             if (this.remaps.ActiveIteratorStateMachineRemap != null
+                && tpSym.IsMethodTypeParameter
                 && this.remaps.ActiveIteratorStateMachineRemap.TryGetValue(tpSym, out var smClassOrd))
             {
                 encoder.GenericTypeParameter(smClassOrd);
--- a/src/Core/CodeAnalysis/Emit/ImportedMemberRefFactory.cs
+++ b/src/Core/CodeAnalysis/Emit/ImportedMemberRefFactory.cs
@@ -274,10 +274,6 @@ internal sealed class ImportedMemberRefFactory
 
             // Issue #810: inside an iterator state-machine method body,
             // outer-method TPs are remapped to the SM class's own TPs.
-            // Issue #1477: a synthesized closure / capture-box class is also
-            // generic over enclosing TYPE parameters, so any TP present in the
-            // active remap (class or method) maps to the synthesized class's
-            // own VAR(idx) slot.
             if (this.remaps.ActiveIteratorStateMachineRemap != null
                 && tpSym.IsMethodTypeParameter
                 && this.remaps.ActiveIteratorStateMachineRemap.TryGetValue(tpSym, out var smClassOrd))
@@ -291,6 +287,13 @@ internal sealed class ImportedMemberRefFactory
                 // generic-promoted non-capturing lambda's signature/body maps to
                 // the lambda method's own MVar(idx) slot.
                 encoder.GenericMethodTypeParameter(lambdaMethodOrd);
+            }
+            else if (this.remaps.ActiveIteratorStateMachineRemap != null
+                && this.remaps.ActiveIteratorStateMachineRemap.TryGetValue(tpSym, out var smClassOrd2))
+            {
+                // Issue #1477: synthesized closure, capture-box, and nested-type
+                // remaps still map class type parameters to their own VAR slots.
+                encoder.GenericTypeParameter(smClassOrd2);
             }
             else if (tpSym.IsMethodTypeParameter)
             {
--- a/src/Core/CodeAnalysis/Emit/ImportedMemberRefFactory.cs
+++ b/src/Core/CodeAnalysis/Emit/ImportedMemberRefFactory.cs
@@ -278,18 +278,18 @@ internal sealed class ImportedMemberRefFactory
             // generic over enclosing TYPE parameters, so any TP present in the
             // active remap (class or method) maps to the synthesized class's
             // own VAR(idx) slot.
-            if (this.remaps.ActiveLambdaMethodTypeParamRemap != null
+            if (this.remaps.ActiveIteratorStateMachineRemap != null
+                && this.remaps.ActiveIteratorStateMachineRemap.TryGetValue(tpSym, out var smClassOrd))
+            {
+                encoder.GenericTypeParameter(smClassOrd);
+            }
+            else if (this.remaps.ActiveLambdaMethodTypeParamRemap != null
                 && this.remaps.ActiveLambdaMethodTypeParamRemap.TryGetValue(tpSym, out var lambdaMethodOrd))
             {
                 // Issue #2118: reference to an enclosing type parameter inside a
                 // generic-promoted non-capturing lambda's signature/body maps to
                 // the lambda method's own MVar(idx) slot.
                 encoder.GenericMethodTypeParameter(lambdaMethodOrd);
-            }
-            else if (this.remaps.ActiveIteratorStateMachineRemap != null
-                && this.remaps.ActiveIteratorStateMachineRemap.TryGetValue(tpSym, out var smClassOrd))
-            {
-                encoder.GenericTypeParameter(smClassOrd);
             }
             else if (tpSym.IsMethodTypeParameter)
             {
--- a/src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs
@@ -1208,20 +1208,22 @@ internal sealed class ReflectionMetadataEmitter
 
         this.closures.SynthesizeClosures(lambdaLiterals, hostPackageGuess);
 
-        // Generic iterator literals need their lambda-method type parameters
-        // before state-machine synthesis so the state machine mirrors them.
+        // Lower non-capturing literals once so iterator plans and emitted bodies
+        // share the same synthesized local symbols.
         foreach (var literal in lambdaLiterals)
         {
             if (literal.CapturedVariables.Length == 0
                 && !this.closures.ClosureInfos.ContainsKey(literal))
             {
+                var body = (BoundBlockStatement)Lowerer.Lower(literal.Body);
+                this.lambdaBodies[literal.Function] = body;
                 this.userTokens.TryPromoteNonCapturingGenericLambda(
                     literal,
-                    (BoundBlockStatement)Lowerer.Lower(literal.Body));
+                    body);
             }
         }
 
-        this.RetargetCapturedIteratorPlans(lambdaLiterals);
+        this.RetargetIteratorPlans(lambdaLiterals);
         this.closures.SynthesizeGoClosures(goStatements, hostPackageGuess);
         this.stateMachines.SynthesizeIteratorStateMachines(hostPackageGuess);
         this.stateMachines.SynthesizeAsyncIteratorStateMachines(hostPackageGuess);
@@ -1230,7 +1232,7 @@ internal sealed class ReflectionMetadataEmitter
         return lambdaLiterals;
     }
 
-    private void RetargetCapturedIteratorPlans(List<BoundFunctionLiteralExpression> lambdaLiterals)
+    private void RetargetIteratorPlans(List<BoundFunctionLiteralExpression> lambdaLiterals)
     {
         if (this.stateMachines.IteratorPlans.IsDefaultOrEmpty)
         {
@@ -1240,9 +1242,10 @@ internal sealed class ReflectionMetadataEmitter
         var plans = this.stateMachines.IteratorPlans.ToBuilder();
         foreach (var literal in lambdaLiterals)
         {
-            if (!this.closures.ClosureInfos.TryGetValue(literal, out var closure))
+            var function = literal.Function;
+            if (this.closures.ClosureInfos.TryGetValue(literal, out var closure))
             {
-                continue;
+                function = closure.InvokeMethod;
             }
 
             for (var i = 0; i < plans.Count; i++)
@@ -1252,8 +1255,8 @@ internal sealed class ReflectionMetadataEmitter
                     continue;
                 }
 
-                var body = this.lambdaBodies[closure.InvokeMethod];
-                if (IteratorRewriter.TryBuildPlan(closure.InvokeMethod, body, out var plan))
+                var body = this.lambdaBodies[function];
+                if (IteratorRewriter.TryBuildPlan(function, body, out var plan))
                 {
                     plans[i] = plan;
                 }
@@ -2920,9 +2923,6 @@ internal sealed class ReflectionMetadataEmitter
                 {
                     continue;
                 }
-
-                var loweredLambdaBody = (BoundBlockStatement)Lowerer.Lower(literal.Body);
-                this.lambdaBodies[literal.Function] = loweredLambdaBody;
 
                 hostBucket.Add(literal.Function);
             }
--- a/src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs
@@ -1205,13 +1205,64 @@ internal sealed class ReflectionMetadataEmitter
         var hostPackageGuess = this.emitCtx.Program.EntryPoint?.Package
             ?? this.emitCtx.Program.EntryPointPackage
             ?? (this.emitCtx.Program.Packages.IsDefaultOrEmpty ? null : this.emitCtx.Program.Packages[0]);
+
         this.closures.SynthesizeClosures(lambdaLiterals, hostPackageGuess);
+
+        // Generic iterator literals need their lambda-method type parameters
+        // before state-machine synthesis so the state machine mirrors them.
+        foreach (var literal in lambdaLiterals)
+        {
+            if (literal.CapturedVariables.Length == 0
+                && !this.closures.ClosureInfos.ContainsKey(literal))
+            {
+                this.userTokens.TryPromoteNonCapturingGenericLambda(
+                    literal,
+                    (BoundBlockStatement)Lowerer.Lower(literal.Body));
+            }
+        }
+
+        this.RetargetCapturedIteratorPlans(lambdaLiterals);
         this.closures.SynthesizeGoClosures(goStatements, hostPackageGuess);
         this.stateMachines.SynthesizeIteratorStateMachines(hostPackageGuess);
         this.stateMachines.SynthesizeAsyncIteratorStateMachines(hostPackageGuess);
         this.stateMachines.SynthesizeAsyncLambdaStateMachines(lambdaLiterals, hostPackageGuess);
 
         return lambdaLiterals;
+    }
+
+    private void RetargetCapturedIteratorPlans(List<BoundFunctionLiteralExpression> lambdaLiterals)
+    {
+        if (this.stateMachines.IteratorPlans.IsDefaultOrEmpty)
+        {
+            return;
+        }
+
+        var plans = this.stateMachines.IteratorPlans.ToBuilder();
+        foreach (var literal in lambdaLiterals)
+        {
+            if (!this.closures.ClosureInfos.TryGetValue(literal, out var closure))
+            {
+                continue;
+            }
+
+            for (var i = 0; i < plans.Count; i++)
+            {
+                if (!ReferenceEquals(plans[i].Function, literal.Function))
+                {
+                    continue;
+                }
+
+                var body = this.lambdaBodies[closure.InvokeMethod];
+                if (IteratorRewriter.TryBuildPlan(closure.InvokeMethod, body, out var plan))
+                {
+                    plans[i] = plan;
+                }
+
+                break;
+            }
+        }
+
+        this.stateMachines.IteratorPlans = plans.ToImmutable();
     }
 
     private void RegisterGeneratedGenericRemaps()
@@ -1224,6 +1275,17 @@ internal sealed class ReflectionMetadataEmitter
         foreach (var kvp in this.stateMachines.IteratorStateMachineInfos)
         {
             var remap = kvp.Value.BuildRemap();
+            if (this.remaps.TryGetLambdaMethodRemap(kvp.Value.Plan.Function, out var lambdaRemap))
+            {
+                remap ??= new Dictionary<TypeParameterSymbol, int>();
+                var offset = kvp.Value.ClassTypeParameters.Length
+                    - kvp.Value.OuterMethodTypeParameters.Length;
+                foreach (var pair in lambdaRemap)
+                {
+                    remap[pair.Key] = offset + pair.Value;
+                }
+            }
+
             if (remap != null)
             {
                 this.remaps.RegisterClassRemap(kvp.Key, remap);
@@ -2861,16 +2923,6 @@ internal sealed class ReflectionMetadataEmitter
 
                 var loweredLambdaBody = (BoundBlockStatement)Lowerer.Lower(literal.Body);
                 this.lambdaBodies[literal.Function] = loweredLambdaBody;
-
-                // Issue #2118: a non-capturing lambda hosted as a top-level
-                // <Program> static method must be promoted to a GENERIC method
-                // declaring its own type parameters (cloned, with constraints,
-                // from the enclosing type parameters its signature/body
-                // references) — otherwise its body's `!!0`/`!0` references a
-                // type parameter the method never declares, producing
-                // unverifiable IL (DelegateCtor at the delegate site and
-                // StackUnexpected on any `constrained.` call in the body).
-                this.userTokens.TryPromoteNonCapturingGenericLambda(literal, loweredLambdaBody);
 
                 hostBucket.Add(literal.Function);
             }
--- a/src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs
@@ -1281,8 +1281,8 @@ internal sealed class ReflectionMetadataEmitter
             if (this.remaps.TryGetLambdaMethodRemap(kvp.Value.Plan.Function, out var lambdaRemap))
             {
                 remap ??= new Dictionary<TypeParameterSymbol, int>();
-                var offset = kvp.Value.ClassTypeParameters.Length
-                    - kvp.Value.OuterMethodTypeParameters.Length;
+                var offset = kvp.Value.SourceTypeParameters.Length
+                    - kvp.Value.Plan.Function.TypeParameters.Length;
                 foreach (var pair in lambdaRemap)
                 {
                     remap[pair.Key] = offset + pair.Value;
--- a/src/Core/CodeAnalysis/Emit/SignatureEncoder.cs
+++ b/src/Core/CodeAnalysis/Emit/SignatureEncoder.cs
@@ -273,15 +273,6 @@ internal sealed class SignatureEncoder
                 encoder.GenericTypeParameter(classOrdinal);
             }
 
-            // Issue #1477: a synthesized closure / capture-box class is generic
-            // over enclosing TYPE parameters too, so any TP in the active remap
-            // (class or method) encodes the synthesized class's own Var(idx).
-            else if (this.remaps.ActiveIteratorStateMachineRemap != null
-                && this.remaps.ActiveIteratorStateMachineRemap.TryGetValue(tp, out var classOrdinal2))
-            {
-                encoder.GenericTypeParameter(classOrdinal2);
-            }
-
             // Issue #2118: inside a generic-promoted non-capturing lambda's
             // signature/body, references to the enclosing type parameters map to
             // the lambda method's own MVar(idx) slots.
@@ -289,6 +280,15 @@ internal sealed class SignatureEncoder
                 && this.remaps.ActiveLambdaMethodTypeParamRemap.TryGetValue(tp, out var lambdaMethodOrd))
             {
                 encoder.GenericMethodTypeParameter(lambdaMethodOrd);
+            }
+
+            // Issue #1477: a synthesized closure / capture-box class is generic
+            // over enclosing TYPE parameters too, so any remaining TP in the
+            // active remap encodes the synthesized class's own Var(idx).
+            else if (this.remaps.ActiveIteratorStateMachineRemap != null
+                && this.remaps.ActiveIteratorStateMachineRemap.TryGetValue(tp, out var classOrdinal2))
+            {
+                encoder.GenericTypeParameter(classOrdinal2);
             }
 
             // ADR-0087 §3 R2: a user-declared open type parameter encodes as
--- a/src/Core/CodeAnalysis/Emit/SignatureEncoder.cs
+++ b/src/Core/CodeAnalysis/Emit/SignatureEncoder.cs
@@ -258,15 +258,6 @@ internal sealed class SignatureEncoder
         }
         else if (type is TypeParameterSymbol tp)
         {
-            // Issue #2118: inside a generic-promoted non-capturing lambda's
-            // signature/body, references to the enclosing type parameters map to
-            // the lambda method's own MVar(idx) slots.
-            if (this.remaps.ActiveLambdaMethodTypeParamRemap != null
-                && this.remaps.ActiveLambdaMethodTypeParamRemap.TryGetValue(tp, out var lambdaMethodOrd))
-            {
-                encoder.GenericMethodTypeParameter(lambdaMethodOrd);
-            }
-
             // Issue #810: when emitting a state-machine method (or the
             // state-machine class itself), references to the OUTER
             // method's type parameters are encoded as the SM class's
@@ -275,7 +266,7 @@ internal sealed class SignatureEncoder
             // `!0` on the SM class. The remap is pushed by
             // EmitIteratorStateMachineMember and is keyed by the
             // method TP's instance identity.
-            else if (this.remaps.ActiveIteratorStateMachineRemap != null
+            if (this.remaps.ActiveIteratorStateMachineRemap != null
                 && tp.IsMethodTypeParameter
                 && this.remaps.ActiveIteratorStateMachineRemap.TryGetValue(tp, out var classOrdinal))
             {
@@ -289,6 +280,15 @@ internal sealed class SignatureEncoder
                 && this.remaps.ActiveIteratorStateMachineRemap.TryGetValue(tp, out var classOrdinal2))
             {
                 encoder.GenericTypeParameter(classOrdinal2);
+            }
+
+            // Issue #2118: inside a generic-promoted non-capturing lambda's
+            // signature/body, references to the enclosing type parameters map to
+            // the lambda method's own MVar(idx) slots.
+            else if (this.remaps.ActiveLambdaMethodTypeParamRemap != null
+                && this.remaps.ActiveLambdaMethodTypeParamRemap.TryGetValue(tp, out var lambdaMethodOrd))
+            {
+                encoder.GenericMethodTypeParameter(lambdaMethodOrd);
             }
 
             // ADR-0087 §3 R2: a user-declared open type parameter encodes as
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -877,12 +877,7 @@ public sealed partial class Evaluator
     }
 
     private object InvokeMaterializedClosure(ClosureValue closure, object[] arguments)
-    {
-        var result = InvokeClosure(closure, arguments);
-        return closure.Function.IsAsync
-            ? WrapAsyncResult(closure.Function.Type, result)
-            : result;
-    }
+        => InvokeClosure(closure, arguments);
 
     private object InvokeClosure(ClosureValue closure, object[] arguments)
     {
@@ -899,7 +894,7 @@ public sealed partial class Evaluator
 
         using (PushFrame(frame))
         {
-            return EvaluateFunctionBody(closure.Body);
+            return EvaluateUserMethodBody(closure.Function, closure.Body);
         }
     }
 
--- a/src/Core/CodeAnalysis/Lowering/Iterators/IteratorRewriter.cs
+++ b/src/Core/CodeAnalysis/Lowering/Iterators/IteratorRewriter.cs
@@ -36,34 +36,47 @@ public static class IteratorRewriter
         }
 
         var plans = ImmutableArray.CreateBuilder<IteratorStateMachinePlan>();
+        var nestedCollector = new NestedIteratorCollector(plans);
 
         foreach (var pair in program.Functions.OrderBy(p => p.Key.Name, StringComparer.Ordinal))
         {
             var function = pair.Key;
             var body = pair.Value;
 
-            if (!IteratorDetection.ContainsYield(body))
-            {
-                continue;
-            }
-
-            // Skip async iterators — they go through the async iterator rewriter path.
-            if (AsyncIteratorDetection.IsAsyncIteratorFunction(function, body))
-            {
-                continue;
-            }
-
-            var elementType = GetIteratorElementType(function.Type);
-            if (elementType == null)
-            {
-                continue;
-            }
-
-            var plan = BuildPlan(function, body, elementType);
-            plans.Add(plan);
+            AddPlan(function, body, plans);
+            nestedCollector.RewriteStatement(body);
         }
 
         return new IteratorRewriteResult(program, plans.ToImmutable());
+    }
+
+    internal static bool TryBuildPlan(
+        FunctionSymbol function,
+        BoundBlockStatement body,
+        out IteratorStateMachinePlan plan)
+    {
+        var elementType = GetIteratorElementType(function.Type);
+        if (!IteratorDetection.ContainsYield(body)
+            || AsyncIteratorDetection.IsAsyncIteratorFunction(function, body)
+            || elementType == null)
+        {
+            plan = null;
+            return false;
+        }
+
+        plan = BuildPlan(function, body, elementType);
+        return true;
+    }
+
+    private static void AddPlan(
+        FunctionSymbol function,
+        BoundBlockStatement body,
+        ImmutableArray<IteratorStateMachinePlan>.Builder plans)
+    {
+        if (TryBuildPlan(function, body, out var plan))
+        {
+            plans.Add(plan);
+        }
     }
 
     private static TypeSymbol GetIteratorElementType(TypeSymbol type)
@@ -138,6 +151,22 @@ public static class IteratorRewriter
         var collector = new LocalCollector();
         collector.Visit(body);
         return collector.Locals.ToImmutableArray();
+    }
+
+    private sealed class NestedIteratorCollector : NestedFunctionBodyRewriter
+    {
+        private readonly ImmutableArray<IteratorStateMachinePlan>.Builder plans;
+
+        public NestedIteratorCollector(ImmutableArray<IteratorStateMachinePlan>.Builder plans)
+        {
+            this.plans = plans;
+        }
+
+        protected override BoundExpression RewriteFunctionLiteralExpression(BoundFunctionLiteralExpression node)
+        {
+            AddPlan(node.Function, node.Body, this.plans);
+            return base.RewriteFunctionLiteralExpression(node);
+        }
     }
 
     private sealed class YieldStateCollector : BoundTreeWalker
--- a/test/Compiler.Tests/Emit/Issue2957FunctionLiteralIteratorTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2957FunctionLiteralIteratorTests.cs
@@ -69,7 +69,7 @@ public sealed class Issue2957FunctionLiteralIteratorTests
             "4\n4\n5\n");
 
         yield return Case(
-            "GenericElementType",
+            "GenericStateMachineSignatureRemap",
             """
             package Generic
             import System
@@ -83,6 +83,81 @@ public sealed class Issue2957FunctionLiteralIteratorTests
             }
             """,
             "ok\n");
+
+        yield return Case(
+            "LoweredForInEnumerator",
+            """
+            package ForIn
+            import System
+
+            let values = func() sequence[int32] {
+                for value in []int32{4, 5} {
+                    yield value
+                }
+            }
+
+            for value in values() {
+                Console.WriteLine(value)
+            }
+            """,
+            "4\n5\n");
+
+        yield return Case(
+            "LoweredForInEnumeratorWithCapture",
+            """
+            package ForInCapture
+            import System
+
+            func make(offset int32) () -> sequence[int32] {
+                return func() sequence[int32] {
+                    for value in []int32{4, 5} {
+                        yield offset + value
+                    }
+                }
+            }
+
+            for value in make(100)() {
+                Console.WriteLine(value)
+            }
+            """,
+            "104\n105\n");
+
+        yield return Case(
+            "LoweredForEllipsisCounter",
+            """
+            package ForEllipsis
+            import System
+
+            let values = func() sequence[int32] {
+                for value in 0 ... 2 {
+                    yield value
+                }
+            }
+
+            for value in values() {
+                Console.WriteLine(value)
+            }
+            """,
+            "0\n1\n");
+
+        yield return Case(
+            "GenericStateMachineElementTokenRemap",
+            """
+            package GenericToken
+            import System
+
+            func make[T any]() (T) -> sequence[T] {
+                return func(value T) sequence[T] {
+                    let copy = []T{value}
+                    yield copy[0]
+                }
+            }
+
+            for value in make[int32]()(42) {
+                Console.WriteLine(value)
+            }
+            """,
+            "42\n");
 
         yield return Case(
             "ImmediatelyInvoked",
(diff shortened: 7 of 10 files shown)


DavidObando:
## Coordination: this PR likely closes more than #2957 and #2976 — please verify and declare

Two hunks in this PR have effects beyond its stated scope. Both need either a test + a `Closes:` line, or an explicit statement that they do not apply. An issue silently fixed with no test is the failure mode we have hit twice in this batch.

### 1. `Evaluator.Expressions.cs:899` may close #3003 and #2995

You changed `InvokeClosure` from `EvaluateFunctionBody(closure.Body)` to `EvaluateUserMethodBody(closure.Function, closure.Body)` — routing lambda bodies through the path named functions already use.

If `EvaluateUserMethodBody` runs `Lowerer.Lower` on the body, this also fixes **#3003**: today `Lowerer` is a `BoundTreeRewriter` that deliberately does not descend into lambda bodies (`NestedFunctionBodyRewriter.cs:16-21`), so `Evaluator.Expressions.cs:762` evaluates the **unlowered** body and `Statements.cs:168` throws at `default:`.

Measured today in `gsi`, each construct inside a lambda body:

| construct | result |
|---|---|
| `if` | **FAIL** `Unexpected node IfStatement` |
| `for v in []int32{1,2,3}` | **FAIL** `Unexpected node ForRangeStatement` |
| `for t < n { }` | works |
| `switch`/`case` | works |
| same `if` in a **named** function | works |

Mechanism: the flat evaluation loop implements a fixed kind list (`Goto`/`ConditionalGoto`/`Label`/`Return`/`Try`/`Throw`/`PatternSwitch`/`Go`/`ChannelSend`/`Select`/`Scope`/`Fixed`/`AwaitForRange`/`Yield` + `VariableDeclaration`/`ExpressionStatement`). `IfStatement` and `ForRangeStatement` are absent **precisely because lowering normally eliminates them**.

**#2995 is the same defect** — proven by a 2×2 on BCL types only, no `Gsharp.Extensions` required:

| | no `if` | with `if` |
|---|---|---|
| `Predicate[int32] -> bool` via `Array.FindAll` | `2` OK | `TargetInvocationException` |
| `Converter[int32,string] -> string?` returning `nil`, via `Array.ConvertAll` | `3` OK | `TargetInvocationException` |

Both rows flip on the `if`; neither column flips on nullability. So #2995 is not a nullable-lambda defect and not a `TargetInvocationException`-unwrap defect.

**Please check whether your change makes `if` and for-range work inside a lambda body.** If yes: add tests and declare `Closes #3003` and `Closes #2995`. **Test that `if` and for-range now work — do NOT enumerate the currently-failing node kinds**, since that set is the unstable intersection of "the binder emits this node" and "the flat loop lacks a case for it".

Note also that #2976 and #3003 are **two different missing stages** (iterator-sink dispatch vs lowering), verified: a literal iterator whose body is just `yield 1` / `yield 2` — nothing needing lowering — still failed, because `yield` survives lowering as `YieldStatement`, which *is* on the handled list. If your one change fixes both, that is good, but please pin both independently.

### 2. `IteratorRewriter.cs` is contended — #2975 is being fixed in it right now

You restructured `Rewrite` at `:36-70` (`NestedIteratorCollector`, `TryBuildPlan`, `AddPlan`). A separate agent is fixing **#2975** in the same file at `:158` — `LocalCollector` records variable *declarations* only, so a `select` arm's received-value variable (a bare `VariableSymbol` on the `BoundSelectCase` record, never *declared* through a node the collector sees) is never hoisted into the sync iterator state machine. Async escapes it because `AsyncCaptureWalker.ReferenceCollector:145` is reference-based.

**This PR merges first**; #2975 will rebase onto it. No action needed beyond not widening into `:158`. If your `NestedIteratorCollector` work already threads select-arm variables into the field map, say so — that would close #2975 and the other agent should stop.

Note `#2975`'s issue body names `HoistedFieldRewriter` as the fix home. **That is wrong** — `HoistedFieldRewriter` consumes a `fieldMap` it does not build, and `BoundSelectCase` is not a `BoundNode` (no `Kind`, undispatchable by `BoundTreeRewriter`). #2982 is already closed as a duplicate of #2975.

### Verification standard for this PR

Observe your own base→head behavioural difference before claiming any of the above. `IlVerifier.Verify` is not sufficient — use `Assembly.Load(File.ReadAllBytes(...))` + `GetTypes()` + bounded child execution. Build with `-m:1`, gate on exit code (never `grep -q "0 Error"` — `"10 Error(s)"` contains it), then check for a zero-byte `GSharp.Core.dll`, then md5-propagation across `src/Compiler` and all three `out/bin/Release/*.Tests/`.

Baseline for `main` = `a6b1561f2`: Core **7028 + 1 skip**, Compiler **3906**, Interpreter **491**, LSP **452** (executed counts; `--list-tests` declared counts are not comparable). An agent building only `Compiler.Tests` sees the same total 3906 split `3887 passed / 19 failed` — that is the known `Gsharp.Extensions` build-closure gap, not a regression.

DavidObando:
# Review — NOT MERGEABLE

One blocker: **this PR converts a clean compile-time rejection into generated code that compiles and then crashes at runtime.** Separately, `Fixes #2976` is only partially true, and closing it would leave a new `gsc`/`gsi` divergence undocumented.

The core change is right and I want to say so first: routing `IteratorRewriter` through the existing `NestedFunctionBodyRewriter` rather than adding a walker is the correct call, and `Evaluator.Expressions.cs` matches the fix shape I recommended on #2976. The problem is the *stage* the plan is built at, not the approach.

## Method

Scratch clones, `git clone --shared --no-checkout` then `git checkout <exact SHA>`. Base is the PR's actual merge base **`b1e693bb1`**, not `a6b1561f2` — `a6b1561f2` is *not* an ancestor of head (`git diff --stat a6b1561f2..ab4fd05c -- src/` shows `Binder.cs -1` / `LambdaBinder.cs -19`, i.e. main has moved on). Both trees built with `dotnet build GSharp.sln -c Release -m:1`, gated exit code → non-zero size → md5 propagation across six `GSharp.Core.dll` consumers:

```
base b1e693bb1  exit=0  23c0eec38705cb42c45f8e5687f2a368  2471936  (6/6 agree)
head ab4fd05c7  exit=0  2999f98396d6aef753727745b29b53e9  2472960  (6/6 agree)
```

Distinct hashes, so the two trees are genuinely different binaries. `gsc`/`gsi` invoked directly; compiled output executed as a child process with `timeout 20` and exact stdout.

## Base → head differential observed

| Probe | `gsc` base | `gsc` head | `gsi` base | `gsi` head |
|---|---|---|---|---|
| **R1** literal iterator, `yield 2` | GS9998 | **`2`** | GS9999 | **`2`** |
| **R9** literal iterator, `while` + local across yield | GS9998 | **`0,1`** | GS9999 | **`0,1`** |
| **R2** literal iterator, body has `if` | GS9998 | **`2,3`** | `Unexpected node IfStatement` | `Unexpected node IfStatement` |
| **R7** literal iterator, body has `for n in []int32{4,5}` | GS9998 | **`4` then NRE, rc=134** | `Unexpected node ForRangeStatement` | `Unexpected node ForRangeStatement` |
| **R8** same, **capturing** an outer local | GS9998 | **`104` then NRE, rc=134** | `Unexpected node ForRangeStatement` | `Unexpected node ForRangeStatement` |
| **R3** plain lambda, body has `if` | `10` | `10` | `Unexpected node IfStatement` | `Unexpected node IfStatement` |
| **R4** `Array.FindAll` + `Predicate` with `if` | `2` | `2` | `TargetInvocationException` | `TargetInvocationException` |
| **R5** async lambda | `async-ran` | `async-ran` | `async-ran` | `async-ran` |

R1 and R9 are the required base→head behavioural difference: both flip from hard failure to correct output in **both** backends. The fix is real and I was looking at it.

---

## BLOCKER — `for-in` inside a function-literal iterator compiles and then crashes

```gsharp
package P
import System
let values = func() sequence[int32] {
    for n in []int32{4, 5} { yield n }
}
for v in values() { Console.WriteLine(v) }
```

* **base:** `error GS9998: yield reached the emitter before iterator lowering` — a clean, honest compile-time rejection.
* **head:** compiles, `Wrote R7b.dll` (3584 bytes), then at runtime:

```
4
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at P.<Program>.<<lambda1>>d__1.MoveNext()
   at P.<Program>.<Main>$(String[] args)
```

Exit code 134. It yields the first element, then dereferences null on resume, inside the synthesized state machine this PR creates. **A compile-time error becoming a runtime crash in emitted code is strictly worse than the bug being fixed** — the old behaviour was safe, this one ships.

**R8 shows the existing retarget does not cover it.** R8 is capture-bearing, so `RetargetCapturedIteratorPlans` applies to it, and it still crashes identically. So this is not simply "the non-capturing path was missed".

### Root cause

`NestedIteratorCollector` builds the plan from the **unlowered** literal body:

```csharp
protected override BoundExpression RewriteFunctionLiteralExpression(BoundFunctionLiteralExpression node)
{
    AddPlan(node.Function, node.Body, this.plans);   // node.Body is NOT lowered
```

`AddPlan` → `TryBuildPlan` → `BuildPlan` → `CollectHoistedLocals`, whose `LocalCollector` records `BoundVariableDeclaration` nodes. The emitter then lowers the body separately (`ReflectionMetadataEmitter`, `Lowerer.Lower(literal.Body)`), and `for-in` is precisely a construct `Lowerer` rewrites into **new enumerator/index locals**. Those locals do not exist when the plan is built, so they are never hoisted; they stay `MoveNext` locals and reset on resume. Hence the null.

Three probes isolate it, same binary, same run:

| body construct | new locals introduced by lowering? | head result |
|---|---|---|
| `while` + explicit `var i` (R9) | no — the local exists pre-lowering | ✅ `0,1` |
| `if` (R2) | no locals live across the yield | ✅ `2,3` |
| `for n in [...]` (R7/R8) | **yes** | ❌ NRE |

That is the discriminator. Note the mutation table's `skip closure-plan retargeting` mutant is killed only by the plain-captured-local case (`4\n4\n5\n`), which needs no lowering — so the retarget is pinned exclusively on the case that cannot expose this.

### Fix direction

Build the plan from the **same lowered body instance** the emitter uses. Lowering twice is not sufficient and will look like it works: `Lowerer.Lower` produces fresh local symbols each call, so a plan built from a second `Lowerer.Lower(node.Body)` would hoist locals whose identities differ from the ones in the emitted body. `ReflectionMetadataEmitter` already stores the canonical lowered bodies in `this.lambdaBodies`; `RetargetCapturedIteratorPlans` is the right hook, and it needs to (a) cover non-capturing literals as well and (b) actually take effect for capture-bearing ones, which R8 shows it currently does not.

Whatever the shape, **R7 and R8 must both become pins.**

---

## SHOULD-FIX — `Fixes #2976` is only partially true, and closing it hides a new divergence

R2 at head: `gsc` prints `2,3`; `gsi` prints `GS9999: Unexpected node IfStatement`. At base the two backends *agreed* — both rejected the program. At head they disagree. **This PR creates a new `gsc`/`gsi` divergence.**

The reason is that the interpreter fix does not lower. `Compilation.Evaluate` (`Compilation.cs:292-321`) runs only `CaptureBoxingRewriter`; `program.Functions[f]` bodies arrive lowered, but `closure.Body` is a `BoundFunctionLiteralExpression.Body`, and `BoundTreeRewriter` "intentionally does not recurse into `BoundFunctionLiteralExpression.Body`" (`NestedFunctionBodyRewriter.cs:15-20`) — `Lowerer.cs` has zero `FunctionLiteral` handling. Routing through `EvaluateUserMethodBody` adds the iterator sink, which is exactly the missing piece for `yield`, but the body is still unlowered, so `Evaluator.Statements.cs`'s flat kind-switch hits `default:` for `IfStatement` / `ForRangeStatement`.

**None of the four new interpreter cases contains an `if` or a for-range inside the literal** — the bodies are `yield 2`, `yield start; yield start+1`, `yield value`, and a named-function guard. So the gap is untested.

Recommended: either lower the closure body on the interpreter path too, or **downgrade to `Fixes #2957` + a partial note on #2976**, and add the `if`-in-literal-iterator case as an explicit expected-failure pin so the divergence is recorded rather than discovered later.

---

## Cleared — three things I checked that are fine

**The async collapse is safe.** `InvokeMaterializedClosure` folding into `InvokeClosure` preserves the wrap, because `EvaluateUserMethodBody` ends with `function.IsAsync ? WrapAsyncResult(function.Type, result) : result` using the same `closure.Function`. R5 confirms identical behaviour base and head in both backends.

One note rather than an objection: `Evaluator.Expressions.cs:826` calls `InvokeClosure` **directly**, so that path now gains async wrapping *and* iterator dispatch it did not have before. That looks like a consistency improvement, but it is an unpinned behaviour change at a call site outside the stated scope — worth one test.

**#3003 and #2995 are NOT closed by this PR.** R3 and R4 are byte-identical at base and head. Do **not** add `Closes` lines for them. This is expected from the mechanism above: both are the unlowered-lambda-body defect, and this PR does not lower lambda bodies on the interpreter path. R2 shows the same root cause surviving inside the iterator case the PR *does* fix, which is a useful cross-check: the iterator-sink stage and the lowering stage are genuinely two different missing stages, exactly as suspected.

**`SignatureEncoder.cs:491` is untouched.** The diff's hunks stop at `:291`. The `ElementType.ClrType ?? typeof(object)` erasure and its #2965 adjacency are not in scope here.

## NIT — the four `Emit/` files are justified, but two are global precedence swaps

`GenericRemapState` adds a getter; `ReflectionMetadataEmitter` moves promotion before state-machine synthesis and adds the retarget. Both are load-bearing for generic iterator literals — fine.

`SignatureEncoder.cs:258-291` and `ImportedMemberRefFactory.cs:278-296` are different in kind: they **swap the precedence** of `ActiveIteratorStateMachineRemap` and `ActiveLambdaMethodTypeParamRemap` from lambda-first to state-machine-first. That changes encoding for *any* generic lambda emitted while an iterator state-machine remap is active, not just the literals this PR adds — and `ImportedMemberRefFactory`'s state-machine branch has no `tp.IsMethodTypeParameter` guard, so its reach is wider than `SignatureEncoder`'s. CI `build` is green, so there is no observed regression, and the mutation table has no mutant covering the swap itself. Worth one sentence in the PR body naming the blast radius, and ideally a mutant that flips the order back.

## Note for #2975 — not affected by this PR

I checked, because a fix agent is working #2975 against this head. `NestedIteratorCollector` only overrides `RewriteFunctionLiteralExpression` and calls `AddPlan`; it does not touch `CollectHoistedLocals`, and `LocalCollector` still records only `VisitVariableDeclaration` / `VisitAddressOfExpression`. **Select-arm bindings are still not collected, so #2975 stays open and its fix agent should continue.** One coordination detail: this PR inserts ~16 lines above `LocalCollector`, moving it from `:158` to roughly `:174`.

Worth noting the two defects are now adjacent — this PR's blocker is *also* a `CollectHoistedLocals` completeness bug (lowering-introduced locals), and #2975 is a `CollectHoistedLocals` completeness bug (select-arm bindings). Same method, two different missing binding sources. Whoever fixes either should look at whether making `LocalCollector` reference-based, the way `AsyncCaptureWalker.ReferenceCollector` already is, closes both at once.

DavidObando:
## Two questions before this can merge

**1. Mentioned but not declared: #2975.** This PR's body mentions **#2975** ("select-arm bindings are not treated as binding sites by state-machine hoisting analysis") but declares only #2957 and #2976. #2975 **still reproduces on `main` (`a6b1561f2`)**, so it is not silently fixed today.

The specific question: does the `NestedIteratorCollector` restructure in `src/Core/CodeAnalysis/Lowering/Iterators/IteratorRewriter.cs:36-70` already close it? #2975's root cause is `IteratorRewriter.cs:158` `LocalCollector`, which collects **declarations only**, whereas the async path escapes the bug via `AsyncCaptureWalker.ReferenceCollector:145`, which is **reference-based**. If your restructure changed what that collector sees, you may have closed #2975 as a side effect — in which case it needs a test pinning it and a `Closes:` line, and the separately-dispatched owner should stand down.

Note that #2975 is **not observable** until PR #2967 lands: without its emit fix at `MethodBodyEmitter.Statements.cs:647`, both the broken and the fixed build print `7`. So a "no change" observation here is only meaningful on top of #2967.

**2. `Evaluator.Expressions.cs:902` — candidate blocker.** This PR changes that line, and independent analysis flagged it as the *wrong* site for lowering, with a double-lowering risk. The settled producer/consumer trace:

- Two producers: `:762` (function-literal body, **unlowered** — this is the actual defect) and `:781` (method group, **already lowered** via `Binder.cs:1383`).
- One dispatch at `:75`; one consumer at `:902`.

Lowering at `:902` therefore (a) re-lowers on **every invocation** rather than once, and (b) **double-lowers** the `:781` bodies, stacking a second `WrapWithMethodExitEpilogue`. The recommended site is `:762`, lowering **once and cached** — precedent at `RefStructAsyncLivenessAnalyzer.cs:658` and `LambdaBinder.cs:1307`.

Please either show that double-lowering cannot occur on the `:781` path, or move the change to `:762`.

**Related:** if this PR does lower function-literal bodies, does it also close **#3003** (`gsi`: an `if` statement inside a lambda body throws `Unexpected node IfStatement`) and **#2995**, which has been proven identical to #3003 by a BCL-only 2×2? The mechanism matches exactly: `Lowerer` is a `BoundTreeRewriter` that deliberately does not descend into lambda bodies (`NestedFunctionBodyRewriter.cs:16-21`), `:762` evaluates the unlowered body, and `Evaluator.Statements.cs:168` throws at `default:`. Exactly four kinds leak — `IfStatement`, `ForRangeStatement`, `ForEllipsisStatement`, `ForInfiniteStatement` — which is an exact identity: the leak set equals the kinds absent from the evaluator's 17-case switch, which equals the kinds lowering eliminates.

If it closes them, say so and I will cancel the separate fix. **Tests should assert the constructs work, never enumerate the failing set.**

Note #2976 does **not** unify with #3003 — a `yield 1`/`yield 2` lambda body needs no lowering (`YieldStatement` *is* handled) yet still fails, because `:902` never pushes an iterator sink. Two stages, two sites.

**Verification standard:** reproduce each issue's own sample against your head build and observe the base→head difference yourself. `IlVerifier.Verify` is **not sufficient** (5 proven cases where it passed an assembly that would not load). Baseline `main` = `a6b1561f2`: Core **7028 + 1 skip** · Compiler **3906** · Interpreter **491** · LSP **452** (executed counts).

DavidObando:
## Addendum — the blocker also produces a **silent wrong answer**, not only a crash

Two more probes on the same base/head pair (`b1e693bb1` → `ab4fd05c7`), same method as above. These extend the blocker from my earlier comment and one of them is materially worse than the NRE.

```gsharp
let values = func() sequence[int32] { for i in 0 ... 2 { yield i } }
for v in values() { Console.WriteLine(v) }
```

* **base:** `GS9998` — clean compile-time rejection
* **head:** compiles, runs, **exit code 0**, prints:

```
0
```

Expected `0` and `1` — the range is half-open. There is no crash, no diagnostic, no non-zero exit. **The iterator silently drops every element after the first.** A wrong answer that exits 0 is worse than the `NullReferenceException` in the `for-in` case, because nothing downstream can detect it.

### Full construct matrix at head, function-literal iterator bodies

| Body construct | Bound kind | Local live across `yield` introduced by lowering? | `gsc` head | `gsi` head |
|---|---|---|---|---|
| `while` + explicit `var i` | `ForConditionStatement` | no — declared in source | ✅ `0,1` | ✅ `0,1` |
| `for { … break }` + explicit `var i` | `ForInfiniteStatement` | no — declared in source | ✅ `0,1` | ❌ `Unexpected node ForInfiniteStatement` |
| `if` | `IfStatement` | no | ✅ `2,3` | ❌ `Unexpected node IfStatement` |
| `for n in []int32{4,5}` | `ForRangeStatement` | **yes** — enumerator/array temp | ❌ `4`, then **NRE**, rc=134 | ❌ `Unexpected node ForRangeStatement` |
| `for i in 0 ... 2` | `ForEllipsisStatement` | **yes** — loop counter | ❌ **`0` — silent truncation, rc=0** | ❌ `Unexpected node ForEllipsisStatement` |

The discriminator from my earlier comment is confirmed exactly, now across five constructs: **the body breaks when and only when lowering introduces a local that is live across the `yield`.** Where the local is written in source it is picked up by `CollectHoistedLocals` and everything works; where lowering synthesises it, the plan was already built and never sees it.

### One observation that ties both backends together

The interpreter's leak set — `IfStatement`, `ForRangeStatement`, `ForEllipsisStatement`, `ForInfiniteStatement`, the kinds absent from the evaluator's switch and thrown at `Evaluator.Statements.cs:168` — is **the same construct class** that breaks the compiler here. Not a coincidence, and not the same bug:

* **interpreter** fails because the literal body is *never lowered*, so lowering-eliminated kinds reach the evaluator's switch;
* **compiler** fails because the iterator *plan is built before lowering*, so lowering-introduced locals are never hoisted.

Two mechanisms, one construct class, opposite sides of the same missing stage-ordering. Worth stating in the PR body, because it also explains why the compiled and interpreted halves of this PR fail on overlapping-but-not-identical inputs — `for { }` with an explicit local, for instance, works in `gsc` and still fails in `gsi`.

### Consequence for the pin set

`for i in 0 ... 2` should be a required pin alongside the `for-in` case. A fix verified only against `for-in` will be verified against the *loud* failure and may leave the silent one, and no existing test in this PR would notice: all six compiler cases and all four interpreter cases use bodies whose locals are explicit in source.

DavidObando:
## Narrow review: three questions only

Scope: `Evaluator.*` + `LambdaBinder`/`Binder` only. Emit files (`GenericRemapState`, `ImportedMemberRefFactory`, `ReflectionMetadataEmitter`, `SignatureEncoder`) and the `IteratorRewriter` restructure belong to another reviewer — untouched here.

Measured at head **`ab4fd05c7`**, base **`a6b1561f2`** (current `origin/main`), merge **`0743ad5dc`** (`ab4fd05c7` + `a6b1561f2`, clean).
Builds: `dotnet build GSharp.sln -c Release -m:1`, exit 0 / 0 warn / 0 err, no zero-byte output, **1** distinct `GSharp.Core.dll` md5 across all 28 copies. Both trees gated.

---

### Q1 — does this close #3003 / #2995? **No. Zero movement.**

17-row `gsi` differential, base vs head, **byte-identical every row**:

| row | construct | base | head |
|---|---|---|---|
| `if1` `if3` `k1` `k2` | `if` / `if-else` in lambda | `GS9999 Unexpected node IfStatement` | same |
| `r1` | for-range in lambda | `GS9999 … ForRangeStatement` | same |
| `r4` | for-infinite in lambda | `GS9999 … ForInfiniteStatement` | same |
| `r5` | for-ellipsis in lambda | `GS9999 … ForEllipsiStatement` | same |
| `t2` | `Array.FindAll` + `if` (#2995 string) | `GS9999 Exception has been thrown…` | same |
| `q2` | `Array.ConvertAll` nullable + `if` | `GS9999 Exception has been thrown…` | same |
| `w1` `w2` `w3` `if2` `r2` `r6` `t1` `q1` | controls | pass | same |

Consistent with Q2. **PR closes #2957 + #2976 only — that is what it claims, and the claim is honest.** No correction needed. Recording it so nobody assumes #3003 rode along.

### Q2 — does `EvaluateUserMethodBody` lower? **No. No double-lowering. I withdraw the concern.**

I warned earlier that moving lowering to the `InvokeClosure` site would (a) re-lower per invocation and (b) double-lower the already-lowered method-group bodies from `Evaluator.Expressions.cs:781`, stacking a second `WrapWithMethodExitEpilogue`. **That trap does not fire here.** `Evaluator.Calls.cs:173-182`:

```csharp
private object EvaluateUserMethodBody(FunctionSymbol function, BoundBlockStatement body)
{
    if (IsIteratorFunction(function, body)) { return EvaluateIteratorFunction(function, body); }
    var result = EvaluateFunctionBody(body);
    return function.IsAsync ? WrapAsyncResult(function.Type, result) : result;
}
```

No `Lowerer.Lower`. It only adds iterator classification + the async wrap. Confirmed observationally: a method-group delegate over a function containing `if` returns `5` / `0` identically on base, head and merge — no doubled epilogue.

The change is therefore correctly sited. **#3003 stays open and still needs the separate `Expressions.cs:762` + cache fix.**

### Q3 — async collapse. **No regression. Exactly one wrap.**

`InvokeMaterializedClosure` no longer wraps; `EvaluateUserMethodBody` does. Same predicate (`function.IsAsync`), same argument (`function.Type`), `function == closure.Function`. Measured:

| probe | base | head |
|---|---|---|
| `async (int32) -> int32` typed var, `.GetType().Name` / `.Result` | `Task`1` / `105` | `Task`1` / `105` |
| inferred `var cb = async func…` | `Task`1` / `105` | `Task`1` / `105` |

`.Result` printing `105` (not a nested `Task`) is the discriminator — a double wrap would show `System.Threading.Tasks.Task`1[System.Int32]` there. Single wrap holds.

One inert delta: `Expressions.cs:826` (raw `ClosureValue` path) previously did **not** async-wrap and now does. Unreachable in practice — `MaterializeClosure:830` only returns a bare `ClosureValue` when `FunctionType.ClrType` is null, and `FunctionTypeSymbol.BuildClrType` yields non-null for every shape including variadic. Harmless either way.

---

## Two notes for the author

**N1 (SHOULD-FIX) — branch is 1 behind `main`; CI never saw the combination.** True merge base is `b1e693bb1`; `a6b1561f2` (#2961, "enforce GS0100 on lambda bodies") landed after the branch was cut. #2961 added `LambdaBinder.CheckAllPathsReturn`, which calls `Lowerer.Lower(body)` **at bind time on lambda bodies** and skips only on `isIteratorReturnType`. That is a plausible semantic conflict with iterator literals that green branch CI cannot detect. I built the merge and it is **clean**:

| suite | merge `0743ad5dc` |
|---|---|
| `Issue2921LambdaBodyDefiniteReturnTests` | 12/12 |
| `Issue2957FunctionLiteralIteratorTests` | 6/6 |
| `Issue2921`+`Issue2976` interpreter | 6/6 |
| `IteratorRewriterTests` | 18/18 |

Plus a direct probe — `var it = func() sequence[int32] { yield 1; yield 2 }` — base `GS9999 'yield' encountered outside of an iterator function`, head `1\n2`, **merge `1\n2` on both `gsi` and `gsc`**, no GS0100 false positive. So: no defect, but **please rebase so CI records this rather than a reviewer**.

*(For anyone reading the diff: `git diff a6b1561f2 HEAD` falsely shows this PR deleting `CheckAllPathsReturn` and 325 lines of #2921 tests. That is a forced two-dot artefact of being 1 behind. Three-dot `b1e693bb1...HEAD` is +495/−50 and touches **no** binder file. I chased this to the bottom before believing it; the PR deletes nothing.)*

**N2 (SHOULD-FIX) — undeclared, untested extra fix.** A **named** iterator reached through a **method-group delegate** is also repaired, because `Expressions.cs:781` closures now route through `EvaluateUserMethodBody`'s iterator classification:

```gsharp
func namedIter() sequence[int32] { yield 7; yield 8 }
var g () -> sequence[int32] = namedIter
for v in g() { Console.WriteLine(v) }
```

base → `GS9999 'yield' encountered outside of an iterator function`; head and merge → `7\n8`. Grep of both new test files shows no method-group coverage — the PR's "named iterator guard" is a direct-call guard. This is real behaviour change on the `:781` producer, currently unpinned. **Add one case.**

---

**Verdict on my three questions: no blockers.** Q2 was the suspected blocker and it is refuted by source and by probe. Fix is correctly sited, async is sound, scope claim is accurate. N1 and N2 are pre-merge hygiene, not defects.

DavidObando:
## Additional blocker — this PR is reported `MERGEABLE` but does not compile against current `main`

GitHub says `MERGEABLE`, and `git merge` produces **zero textual conflicts**. The merged tree still fails to build. This is a semantic merge conflict, invisible to both git and the mergeability check.

```
main            e80132516988c67f33366c955435923a3fcf04e9
merge ab4fd05c  -> 03298d9b3d469c2cc24238981be3e35cd7327ded, 0 conflicts
dotnet build GSharp.sln -c Release   EXIT=1
race-signature hits (MSB302[67]|MSB3883): 0     <- a real failure, not the #2974 race
```

```
src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs(1282,33): error CS1061:
  'StateMachineEmitter.IteratorStateMachineInfo' does not contain a definition for
  'OuterMethodTypeParameters'
```

×4 occurrences, all in `Core.csproj`.

### Cause

The failing line is in this PR's **new** `RegisterGeneratedGenericRemaps` block:

```csharp
var offset = kvp.Value.ClassTypeParameters.Length
    - kvp.Value.OuterMethodTypeParameters.Length;
```

`OuterMethodTypeParameters` was removed by **#2978 (`d4dd13b35`, "Fix nested generic iterator and method metadata")**, which reworked `StateMachineEmitter.cs` by 126 lines:

| SHA | `OuterMethodTypeParameters` in `Emit/StateMachineEmitter.cs` |
|---|---|
| `b1e693bb1` — this PR's merge base | **15** |
| `d4dd13b35` — #2978 | **0** |
| `e80132516` — current main | **0** |

The two PRs touch **different files** — #2978 changed `StateMachineEmitter.cs` and `UserTokenResolver.cs`, #3000 changes `ReflectionMetadataEmitter.cs` — so no file-granularity collision check flags this pair. They share an API surface instead. #2978's body even states "Does not touch `SlotPlanner.cs` or `SignatureEncoder.cs`, so it does not collide with #2965", which is the same file-level reasoning.

### Consequences

1. **Rebase onto `e80132516` is required before this can merge**, and the rebase is not mechanical — the generic-remap offset computation has to be re-expressed against whatever replaced `OuterMethodTypeParameters` in #2978's rework. That is squarely inside the `ReflectionMetadataEmitter` / `GenericRemapState` composition logic I flagged as the widest part of this diff, so it needs re-review after rebase rather than a rubber stamp.

2. **CI's green `build` at `ab4fd05c` does not transfer.** It was measured against the old base. Re-run after rebase.

3. **My two blockers are measured against this PR's natural base `b1e693bb1` and stand as reported** — that is the correct isolation for the PR's own effect. But their status *on top of current main* is currently **unmeasurable**, because the merged tree does not build. Since #2978 reworked iterator state-machine metadata, it is entirely possible it interacts with the plan/hoisting path. **Re-confirm `for n in []int32{4,5}` (NRE) and `for i in 0 ... 2` (silent `0`) after the rebase**, before assuming either is still live or still absent.

Nothing here changes the substance of the earlier review; it adds a prerequisite in front of it.

DavidObando:
## Correction to my N1 — it is a **BLOCKER**, not hygiene. This PR does not build against current `main`.

`main` moved from `a6b1561f2` to **`e80132516`** while I was measuring. My previous comment said *"I built the merge and it is clean."* **That claim was scoped to `a6b1561f2` only and is now wrong.** Retracting it — here is the real result.

### Merge is textually clean and does not compile

```
git merge e80132516   ->  exit 0, 0 conflicts      (516ff8f1e)
dotnet build GSharp.sln -c Release
    -> EXIT 1, 0 warnings, 2 errors, 0 race signatures (MSB302[67]|MSB3883 = 0)
```

```
src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs(1282,33):
error CS1061: 'StateMachineEmitter.IteratorStateMachineInfo' does not contain
a definition for 'OuterMethodTypeParameters'
```

### Why

**`d4dd13b35` (#2978, "Fix nested generic iterator and method metadata")** merged to `main` and deleted `IteratorStateMachineInfo.OuterMethodTypeParameters`, renaming it `SourceTypeParameters`:

| SHA | occurrences in `StateMachineEmitter.cs` |
|---|---:|
| `b1e693bb1` (this PR's merge base) | 15 |
| `a6b1561f2` (my earlier merge test) | 15 |
| **`d4dd13b35`** (#2978) | **0** |
| `e80132516` (current `main`) | 0 |

This PR **introduces the only consumer outside that file** — `+ kvp.Value.OuterMethodTypeParameters.Length` is a `+` line in `b1e693bb1...HEAD`. Neither PR touches a file the other touches, so git merges silently and GitHub reports mergeable. Branch CI is green because it runs against `b1e693bb1`.

### The fix is **not** a rename — #2978 deleted the offset itself

`BuildRemap`, `a6b1561f2`:
```csharp
var offset = this.ClassTypeParameters.Length - this.OuterMethodTypeParameters.Length;
for (var i = 0; i < this.OuterMethodTypeParameters.Length; i++)
    map[this.OuterMethodTypeParameters[i]] = offset + i;
```
`BuildRemap`, `e80132516`:
```csharp
for (var i = 0; i < this.SourceTypeParameters.Length; i++)
    map[this.SourceTypeParameters[i]] = i;          // offset gone
```

`ReflectionMetadataEmitter.cs:1276-1286` recomputes **that same offset** to shift `lambdaRemap` into agreement with `BuildRemap`'s output:
```csharp
var offset = kvp.Value.ClassTypeParameters.Length
    - kvp.Value.OuterMethodTypeParameters.Length;
foreach (var pair in lambdaRemap) { remap[pair.Key] = offset + pair.Value; }
```

Under #2978 `BuildRemap` no longer offsets, so the compensating offset must be **re-derived** against the new model, not renamed. `s/OuterMethodTypeParameters/SourceTypeParameters/` compiles and applies a **phantom shift** — wrong type-parameter indices in state-machine member signatures, which is exactly the generic `sequence[T]` iterator-literal path this PR adds. That failure mode is silent: `IlVerifier.Verify` is known-insufficient for it, so please gate the re-derivation on `Assembly.Load(File.ReadAllBytes(...))` + `GetTypes()` + bounded child execution, and re-run `Issue2951NestedGenericStructIteratorTests` (arrives with #2978) alongside `Issue2957FunctionLiteralIteratorTests`.

### What still stands from my previous comment

Unaffected — all three answers were measured on the PR head itself and none touch `Emit/`:

- **Q1** — does not close #3003 / #2995. 17-row differential byte-identical. Scope claim accurate.
- **Q2** — `EvaluateUserMethodBody` does **not** lower (`Evaluator.Calls.cs:173-182`). No double-lowering. Concern withdrawn; fix correctly sited.
- **Q3** — async collapse sound, exactly one wrap (`.Result` → `105`).
- **N2** — undeclared, untested method-group-to-named-iterator fix. Still applies; still needs a test.

**Please rebase onto `e80132516` and re-derive the remap offset.** Everything I verified previously was against a `main` that no longer exists.

DavidObando:
## The obvious repair compiles and is silently wrong — `offset` collapses to identically zero

This PR is textually `CLEAN` against `main` (`a1237ded7`, verified by `git merge-tree`) and does not compile:

```
ReflectionMetadataEmitter.cs(1282,33): error CS1061:
  'IteratorStateMachineInfo' does not contain a definition for 'OuterMethodTypeParameters'
```

The tempting one-token repair is `OuterMethodTypeParameters` → `SourceTypeParameters`. **It builds, and it is wrong.**

### Why

Merged #2978 (`d4dd13b35`) did not rename a member, it **removed the offset concept**:

```diff
-  var offset = this.ClassTypeParameters.Length - this.OuterMethodTypeParameters.Length;
-  map[this.OuterMethodTypeParameters[i]] = offset + i;
+  map[this.SourceTypeParameters[i]] = i;
```

The old `offset` was the **receiver type-parameter count**: `ClassTypeParameters` mirrored *receiver + method* TPs while `OuterMethodTypeParameters` held *method* TPs only, so the shift skipped the receiver region. `SourceTypeParameters` is documented as "the original **receiver and method** type parameters … in state-machine slot order" — the full set — which is why the new map is a plain identity.

So substituting the new name into this PR's arithmetic at `:1281-1282`:

```csharp
var offset = kvp.Value.ClassTypeParameters.Length - kvp.Value.SourceTypeParameters.Length;
remap[pair.Key] = offset + pair.Value;
```

**`offset` is identically 0.** Two independent confirmations at `a1237ded7`:

1. `ClassTypeParameters` is documented as "constructed as **ordinal-aligned mirrors** of `SourceTypeParameters`" (`StateMachineEmitter.cs:284-291`).
2. `StateMachineEmitter.cs:650` constructs the SM class as `StructSymbol.Construct(smClass, scopeTPs.CastArray<TypeSymbol>(), …)` — over `scopeTPs`, which is what `:665` passes as `SourceTypeParameters`. Arity equality is forced, or that `Construct` call would be malformed.

The result is that `lambdaRemap` values are written through **unshifted**, where the pre-#2978 code deliberately shifted them past the receiver TPs. That is wrong exactly when the receiver is generic — a generic iterator on a generic receiver, i.e. the `sequence[T]` path — and it is silent: the indices are in range, the assembly emits, and `IlVerifier.Verify` cannot see a wrong-but-valid type-parameter ordinal.

### What the repair needs to do

Do not reconstruct the offset arithmetically. Under #2978's model `BuildRemap()` is an identity over `SourceTypeParameters`, so the lambda remap has to be rebased onto that same list — resolve each `pair.Key` to its index in `SourceTypeParameters` rather than adding a computed shift. If a shift is genuinely still required, it is the **receiver TP count**, which is no longer derivable from `ClassTypeParameters.Length - SourceTypeParameters.Length` because that difference is now structurally zero.

Whatever shape it takes, it needs a pin that puts a **generic iterator on a generic receiver** through emit and checks the resulting ordinals — a compile-only check will pass either way.

### Method note

Found by a static pre-filter over `main`-deleted symbols, not by building: for `base = merge-base(head, main)`, take identifiers on the `-` side of `base..main`, keep those absent from `main` entirely, and flag any appearing on the `+` side of the three-dot `base...head`. That reproduced `:1282` exactly. Worth stating that the same filter would **not** have caught the repair trap above — the trap has no deleted identifier, only changed meaning behind a surviving name.

DavidObando:
Resolved both blockers and pushed the rebased head.

- Iterator plans are now rebuilt from the canonical lowered lambda body used by emission, so lowering-synthesized locals have identical symbols in the plan and emitted body.
- `for-in`: base `GS9998` → head `4\n5\n` with exit 0. The pre-fix PR head printed `4` then threw `NullReferenceException`.
- `for-ellipsis`: base `GS9998` → head `0\n1\n` with exit 0. The pre-fix PR head silently printed only `0`.
- Added `LoweredForInEnumerator`, `LoweredForInEnumeratorWithCapture`, and `LoweredForEllipsisCounter`.
- Added named-iterator method-group delegate coverage: base `GS9999` → head `7\n8`.
- Added the `ImportedMemberRefFactory` method-type-parameter guard plus class-remap fallback. Both remap precedence swaps and the guard have killing tests.
- Rebased first to `87ca66e85`, then to current `main` `f70626452`; #2961 and #2978 interaction tests pass.
- #2975, #3003, and #2995 remain open as documented.

Final CI: **8 SUCCESS + 2 SKIPPED**.

DavidObando:
Blocker 3 verification finished after #3000 merged. Follow-up #3036 removes the phantom generic-slot offset, adds direct emitted-signature decoding, and is green with **8 SUCCESS + 2 SKIPPED**.

DavidObando:
Follow-up #3036 now also covers the generic-receiver case missed by #3000. The remaining guards encoded receiver fields as `MVAR` and caused `MissingFieldException`; final head removes both guards, decodes kickoff `MemberRef` signatures directly, and is green with **8 SUCCESS + 2 SKIPPED**.

PR #3002, opened 2026-08-01
Build: isolate SDK staging intermediate outputs

### LINKED ISSUE #2974: Gsharp.Extensions build races GSharp.Core.dll rewrite under parallel MSBuild

DavidObando:
## Summary

Isolate intermediate outputs for the two SDK packaging `Publish` project instances so parallel solution builds cannot read or write the ordinary build's `obj` files.

## Mechanism

`Gsharp.NET.Sdk` sets `GeneratePackageOnBuild=true`. Its compiler and gsgen package targets invoke nested `MSBuild` `Publish` calls with distinct global properties. Those globals fork new instances of every project in each `ProjectReference` closure and propagate to `Core.csproj`.

The staging calls already redirect `BaseOutputPath`, but `build/gsharp.build.props` defines `BaseIntermediateOutputPath` independently. Ordinary, gsc-staging, and gsgen-staging instances therefore shared `out/obj/<Project>/Release/`. A `csc` rewrite of `GSharp.Core.dll` could race another instance's `Copy` or `CopyRefAssembly`. `Gsharp.Extensions` only consumes the resulting corrupted assembly; it does not cause the race.

## Fix

Pass a distinct `GsharpObjRoot` global property to each staging call and derive `IntermediateOutputPath` from it, project name, and configuration. The three instance families now use disjoint file sets while `BaseIntermediateOutputPath` and restore assets remain unchanged.

## Bidirectional reproduction

Every iteration touched `src/Core/CodeAnalysis/Emit/StateMachineEmitter.cs` before running:

```text
dotnet build GSharp.sln -c Release -graph
```

Build exit status was checked first. Successful builds then required all four propagated `GSharp.Core.dll` files to be nonzero and have identical MD5 hashes.

| State | Runs | Race hits | MSB3026 | MSB3883 | Hard failures | DLL integrity failures |
|---|---:|---:|---:|---:|---:|---:|
| Unfixed | 6 | 2 | 2 | 0 | 0 | 0 |
| Fixed | 6 | 0 | 0 | 0 | 0 | 0 |
| Reverted | 8 | 1 | 1 | 0 | 0 | 0 |

These are small local samples; the table reports observed counts without extrapolation.

## Build time

For the same dirty-Core `-graph` runs, median wall time improved from 31.5 seconds unfixed to 27.5 seconds fixed. Excluding each series' first run, medians were 30 and 27 seconds. The first fixed run took 82 seconds while populating the new staging trees.

## Validation

- `dotnet build GSharp.sln -c Release --no-restore -graph`: passed
- Strict DLL check: compiler plus `Compiler.Tests`, `Core.Tests`, and `Interpreter.Tests` copies were nonzero and shared MD5 `189f6849fc0e86d7df4591ae5a4ec6cb`
- Solution closure: `out/bin/Release/Gsharp.Extensions/` populated; all 123 `SampleConformanceTests` passed without `--no-build`
- SDK package payload: 105 entries (`tools/compiler`: 25, `tools/gsgen`: 62, `tools/extensions`: 3, `tools/task`: 1)

Closes #2974

### DIFF
--- a/build/gsharp.build.props
+++ b/build/gsharp.build.props
@@ -58,7 +58,7 @@
     <BaseOutputPath>$(OutRoot)/bin/$(Configuration)</BaseOutputPath>
     <OutputPath>$(BaseOutputPath)/$(MSBuildProjectName)</OutputPath>
     <BaseIntermediateOutputPath>$(OutRoot)/obj/$(MSBuildProjectName)</BaseIntermediateOutputPath>
-    <!-- <IntermediateOutputPath>$(BaseIntermediateOutputPath)/$(MSBuildProjectName)</IntermediateOutputPath> -->
+    <IntermediateOutputPath Condition="'$(GsharpObjRoot)'!=''">$(GsharpObjRoot)/$(MSBuildProjectName)/$(Configuration)/</IntermediateOutputPath>
     <PackageOutputPath>$(BaseOutputPath)/nupkgs</PackageOutputPath>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
     <AssemblyName>GSharp.$(MSBuildProjectName)</AssemblyName>
--- a/src/Sdk/Gsharp.NET.Sdk/Gsharp.NET.Sdk.csproj
+++ b/src/Sdk/Gsharp.NET.Sdk/Gsharp.NET.Sdk.csproj
@@ -99,7 +99,7 @@
     <RemoveDir Directories="$(_GsharpCompilerPublishDir)" Condition="Exists('$(_GsharpCompilerPublishDir)')" />
     <MSBuild Projects="$(MSBuildThisFileDirectory)..\..\Compiler\Compiler.csproj"
              Targets="Publish"
-             Properties="Configuration=$(Configuration);TargetFramework=$(GsharpCompilerPackageTargetFramework);PublishDir=$(_GsharpCompilerPublishDir);BaseOutputPath=$(_GsharpCompilerStagingOutDir);SelfContained=false;PublishReadyToRun=false;PublishSingleFile=false" />
+             Properties="Configuration=$(Configuration);TargetFramework=$(GsharpCompilerPackageTargetFramework);PublishDir=$(_GsharpCompilerPublishDir);BaseOutputPath=$(_GsharpCompilerStagingOutDir);GsharpObjRoot=$(IntermediateOutputPath)gsc-staging-obj;SelfContained=false;PublishReadyToRun=false;PublishSingleFile=false" />
     <ItemGroup>
       <_GsharpCompilerPayload Include="$(_GsharpCompilerPublishDir)**\*" />
     </ItemGroup>
@@ -125,7 +125,7 @@
     <RemoveDir Directories="$(_GsgenPublishDir)" Condition="Exists('$(_GsgenPublishDir)')" />
     <MSBuild Projects="$(MSBuildThisFileDirectory)..\..\..\tools\gsgen\Gsgen.Cli\Gsgen.Cli.csproj"
              Targets="Publish"
-             Properties="Configuration=$(Configuration);TargetFramework=$(GsharpCompilerPackageTargetFramework);PublishDir=$(_GsgenPublishDir);BaseOutputPath=$(_GsgenStagingOutDir);SelfContained=false;PublishReadyToRun=false;PublishSingleFile=false" />
+             Properties="Configuration=$(Configuration);TargetFramework=$(GsharpCompilerPackageTargetFramework);PublishDir=$(_GsgenPublishDir);BaseOutputPath=$(_GsgenStagingOutDir);GsharpObjRoot=$(IntermediateOutputPath)gsgen-staging-obj;SelfContained=false;PublishReadyToRun=false;PublishSingleFile=false" />
     <ItemGroup>
       <_GsgenPayload Include="$(_GsgenPublishDir)**\*" />
     </ItemGroup>


DavidObando:
# Duck review — PR #3002 `ce2285d55` — **MERGEABLE**

Reviewed `refs/pull/3002/head` = `ce2285d550e5e1f8f81f5dc7db16b98ffa7d79c9`, natural base `a6b1561f2` (asserted, `git describe` = `v0.3.362-111-gce2285d55`). Real content confirmed before probing: **3 lines, 2 files**. Two fresh `git clone --shared --no-checkout` trees, explicit `git checkout <sha>`, no `git archive`.

**40 full solution builds** — 20 base / 20 head, **interleaved** to share load conditions, every iteration preceded by `touch src/Core/CodeAnalysis/Emit/StateMachineEmitter.cs`, run as `dotnet build GSharp.sln -c Release -graph` with **no `-m:1`** (the CI configuration, and the flag that would mask the defect).

---

## Q1 — Disjoint, or just a narrower window? **GENUINELY DISJOINT.**

This is a structural result, not a sampling one. After a full solution build:

| | `obj` `GSharp.Core.dll` copies | distinct md5 |
|---|---|---:|
| **base** `a6b1561f2` | ordinary + gsc-publish + gsgen-publish | **1** — `1f7109749f…`, 2,471,936 B |
| **head** `ce2285d55` | ordinary / gsc-staging / gsgen-staging | **3** — `b775aca2…` 2,471,936 · `ad0c2aa6…` 2,472,448 · `fcb90f61…` 2,472,448 |

On base the three instance families were literally **one file with three writers** — that is the bug. On head they are three separate compilations in three directories, and the file groups partition **exactly** by instance family with no overlap:

- `b775aca2…` (23 files) — ordinary → every `out/bin/Release/*` + `out/obj/Core/Release/`
- `ad0c2aa6…` (4) — gsc-staging obj → gsc-publish → gsc-staging bin
- `fcb90f61…` (7) — gsgen-staging obj → gsgen-publish → gsgen-staging bin
- `5cab242e…` (6) — `ref`/`refint` reference assemblies, **byte-identical across all three families** (550,400 B)

The redirect is real, not an empty directory: `gsc-staging-obj/Core/Release/{GSharp.Core.dll, ref/, refint/}` are all populated. That matters because `ref/` is precisely what `CopyRefAssembly`/`MSB3883` contends over. The ref assemblies being identical across families also proves the three builds are **semantically** identical — the impl md5 differences are embedded-path metadata only, since `build/gsharp.build.props:86` leaves `ContinuousIntegrationBuild` false locally.

The race requires two writers to one path. No two instances now share a path ⇒ impossible by construction, not merely improbable.

## Q2 — Bidirectional reproduction. **Fires on base, silent on head.**

| | runs | race hits | rate |
|---|---:|---:|---:|
| base `a6b1561f2` | 20 | **5** (iters 6, 10, 14, 15, 17) | 25 % |
| head `ce2285d55` | 20 | **0** | 0 % |

Fisher exact, one-tailed **p = 0.0236** (two-tailed 0.047). Head had **zero `MSB30*` warnings of any kind** across all 20 logs.

Mechanism confirmed verbatim from `BASE_10.log`:

```
warning MSB3026: Could not copy ".../out/obj/Core/Release/GSharp.Core.dll"
  to ".../out/bin/Release/Core/GSharp.Core.dll" ... The process cannot access the file
  '.../out/obj/Core/Release/GSharp.Core.dll' because it is being used by another process.
  [/private/tmp/duck2888-base/src/Core/Core.csproj]
```

Contended path is the shared intermediate; the owning project is **`Core.csproj`** — independently confirming "the second writer is a second and third instance of `Core.csproj`, not `Compiler`."

## Q3 — Diagnostic code. **Correcting the brief: the signature is `MSB3026`, not `MSB3883`.**

Across 40 builds: `MSB3026` ×5 (all base), **`MSB3883` ×0**, **`MSB3030` ×0**. The author's own table reports the same thing (`MSB3883: 0`, hits are `MSB3026`). So "`MSB3883` is #2974" is too narrow — both are the same contention surfacing through different targets (`Copy` retry vs `CopyRefAssembly`). **Fleet detectors should match `MSB302[67]|MSB3883`.** Nothing here to attribute to `MSB3030`; confirmed absent.

## Q4 — Faster? **Yes, and by more than claimed.**

| | median | mean | min–max |
|---|---:|---:|---|
| base | **31.5 s** | 34.5 s | 28–46 |
| head | **25.0 s** | 25.6 s | 19–36 |

**20.6 % faster.** My base median (31.5 s) is *identical* to the author's; my head median beats their claim (25.0 vs 27.5). Head's worst run (36 s) is faster than base's mean. The direction is causal, not incidental: on base the three instances invalidate each other's shared `obj` and re-do work plus 1000 ms copy retries.

## Q5 — Build-closure gap. **Untouched (correctly), and this is the payoff.**

Both trees populate `out/bin/Release/Gsharp.Extensions/` (5 entries) from a solution build — unchanged by the patch, as it should be, since the closure gap is a separate defect. Confirmed the payoff directly on head: **`SampleConformanceTests` 123/123 passed, `Failed: 0`, without `--no-build`.** The real value of #3002 is that "build the solution" — the closure-gap fix — stops being the thing that makes the race likely.

Package integrity preserved: payload is **105 total entries**, `tools/compiler` 25 · `tools/gsgen` 62 · `tools/extensions` 3 · `tools/task` 1 — matching the author's claim exactly.

## Q6 — CI exposure. **Protected.**

The entire experiment *is* the CI configuration: `-graph`, parallel, no `-m:1`. Head 0/20. The fix operates on file paths, not on scheduling, so it holds independently of node count.

---

# Findings

### SHOULD-FIX (process, not code) — this patch breaks the *naive* form of the fleet's own build gate

Gate 3 **as prescribed** (the five named dirs) is uniform on head — all `b775aca2…`. ✅ No action needed there.

But the naive "every `GSharp.Core.dll` copy must agree" variant goes from **2 md5 groups on base to 4 on head**, and that variant is in active use — I used it myself when measuring the `a6b1561f2` baseline. After this merges it will report **phantom corruption**. Please land a note with the merge: *gate 3 compares the five ordinary-instance paths; staging-instance copies under `out/obj/Gsharp.NET.Sdk/Release/*-staging*` legitimately differ.* Cheap now, confusing later.

### NIT 1 — `Gsharp.NET.Sdk.csproj:167`, the one fork site left un-isolated

`:167` forks `Gsharp.Extensions` with distinct globals (`OutputPath`, `BuildProjectReferences=false`) but receives no `GsharpObjRoot`, so it still shares `out/obj/Gsharp.Extensions/` with the ordinary instance (verified: 1 obj dir on both base and head).

Not a blocker, and not a #2974 recurrence: `BuildProjectReferences=false` stops global propagation to `Core`, and `Gsharp.Extensions.csproj:111` references `Gsharp.NET.Sdk` (with no back-reference — the cycle-avoiding arrow removal at `:60-72`), which serialises the two instances. Observed clean in 20/20 head runs. But the safety is an **incidental ordering, not isolation** — one `ProjectReference` change away from mattering. Adding `GsharpObjRoot` here too would make the invariant structural. `:163` is fine as-is: it passes only `Configuration`, so no fork occurs.

### NIT 2 — staging `obj` trees are never cleaned

`gsc-staging-obj` 4.9 MB + `gsgen-staging-obj` 6.0 MB; total `out/` 614 MB → 622 MB (**+1.3 %**). `RemoveDir` at `:99`/`:125` clears only the publish dirs. Negligible; noting for completeness.

### Note on the author's evidence

Their table (2/6 unfixed, 0/6 fixed) is **underpowered** — Fisher p = 0.227, not significant. They labelled it honestly ("small local samples… without extrapolation"), which is the right call and I want that credited rather than dinged. My 20/20 supplies the power theirs lacked and reaches the same conclusion. Every other number in the PR body reproduced exactly: base median 31.5 s, payload 25/62/3/1, `MSB3883: 0`.

---

## Verdict: **MERGEABLE**

Three lines, structurally correct, causally verified in both directions, statistically significant, ~21 % faster, package payload byte-for-byte intact, CI 10/10 green (`mergeStateStatus` CLEAN). No blockers. The one SHOULD-FIX is a documentation change to the fleet's verification procedure at merge time, not to this patch.

Once merged, `-m:1` can be dropped — but note the corrected reason: **not** because the race became rare, but because the shared path it needed no longer exists.

PR #3005, opened 2026-08-01
Define flags enum exhaustiveness consistently

### LINKED ISSUE #2941: Make [Flags] enum exhaustiveness consistent across switch forms
## Problem

PR #2926 briefly excluded imported CLR `[Flags]` enums from `ExhaustivenessAnalyzer.IsExhaustiveDiscriminant`, then reverted that change because it was language-visible, undocumented, and inconsistent.

The attempted carve-out produced different behavior by symbol origin and switch form:

- imported CLR `[Flags]` switch statement: incomplete switches stopped reporting GS0178;
- imported CLR `[Flags]` switch expression: one location could report both GS0176 and GS0177;
- user-declared `@Flags` enum: still reported GS0178 because `EnumSymbol` short-circuited the CLR attribute check.

It also affected definite-return analysis for exhaustive switch statements without updating `docs/diagnostics.md`.

## Expected work

Define whether exhaustiveness for flags means named constants or all bit combinations, then apply that rule consistently to:

- imported CLR and user-declared flags enums;
- switch statements and switch expressions;
- GS0176/GS0177/GS0178 emission;
- definite-return analysis;
- diagnostics documentation.

Add focused tests for incomplete and name-complete flags switches in both forms and for both enum origins.

### LINKED ISSUE #2971: Exhaustive enum switch expressions silently return default for unnamed values
## Repro

```gs
import System

    return switch x {
        case DateTimeKind.Unspecified: 10
        case DateTimeKind.Utc: 20
        case DateTimeKind.Local: 30
    }
}

```

## Actual

At exact `b78bb1f1`:

- `gsc` compiles successfully and the emitted program silently prints `0` because `EmitSwitchExpression` loads an unassigned result temp when no arm matches.
- `gsi` exits 1 with `GS9999: Object reference not set to an instance of an object.` because `EvaluateSwitchExpression` dereferences a null `defaultArm`.

A source-declared enum has the same compiled `0`; its interpreter path currently fails earlier on the explicit unnamed-value cast with `GS9999: Unexpected type E`.

## Expected

Compiled and interpreted execution should agree and should not silently manufacture `default(T)`. A defensive runtime exception, analogous to C# `SwitchExpressionException` and the existing non-void switch-statement fall-through guard, would preserve named-constant exhaustiveness without returning a wrong answer for unnamed enum values.

## Scope

Found while resolving #2941. Pre-existing and unrelated to its diagnostic-selection fix. `[Flags]` makes the path common because combined values need not have names, but the defect affects every enum. This is not #2940: no iterator lowering is involved.

DavidObando:
## Summary

- Pin enum exhaustiveness to named constants for imported CLR `[Flags]`, source `@Flags`, and ordinary enums across switch statements and switch expressions.
- Fail unmatched switch expressions loudly instead of manufacturing `default(T)`: emitted code throws `InvalidOperationException`, while `gsi` reports `GS9999`, both with `Unmatched switch expression value.`
- Document the named-constant rule and its unmatched-value runtime consequence.
- Add semantic, emitted-runtime, assembly-load, IL-verification, definite-return, and interpreter coverage.

## Design rule

An enum switch is exhaustive when every named constant is covered. `[Flags]` does not add unnamed bit combinations as required arms. Requiring every representable combination would make flags-enum exhaustiveness impractical.

A runtime value can still match no named arm. Such a switch expression now fails explicitly and consistently rather than returning an uninitialized result or dereferencing a missing default arm.

## Production changes

- `MethodBodyEmitter.Patterns.cs`: throw `InvalidOperationException("Unmatched switch expression value.")` when no switch-expression arm matches and no default arm exists.
- `Evaluator.Concurrency.cs`: preserve an inner `InvalidOperationException` with the same message so `gsi` reports `GS9999` when uncaught and matches the same typed catch as emitted code when caught.
- Live diagnostic references: document named-constant exhaustiveness and unmatched runtime behavior.

`ControlFlowGraph.cs` remains unchanged. Statement exhaustiveness still drives definite-return analysis exactly as before.

## Base-to-head differential

Reproduced at base `a1237ded7141b95c9f898b90e3941f732d253b5b` and head `537eeee6e0ffb103cc1ede4a6a3177ea511bc1ae` with `Access.Read | Access.Write` reaching a name-complete switch expression:

| Engine | Base | Head |
|---|---|---|
| `gsc` | compile exit 0; program exit 0; stdout `0` | compile exit 0; program terminates nonzero with `System.InvalidOperationException: Unmatched switch expression value.` |
| `gsi` | exit 1; `GS9999: Object reference not set to an instance of an object.` | exit 1; `GS9999: Unmatched switch expression value.` |

Reverting the blocker commit restored base behavior for both engines. Restoring it restored the explicit runtime failures.

Typed-catch parity at head: both `gsc` and `gsi` exit 0 with exact output `caught-ioe\nafter`. Removing the evaluator inner exception changes only `gsi` to `caught-generic\nafter` and fails the new regression test; restoring it returns `gsi` to `caught-ioe\nafter` and the 3-test selector to green.

## Validation

- Rebased onto `4bba0643a001ad81b07862980a75b0a25c1b021b`.

- Release CI-style solution build: exit 0, 0 warnings, 0 errors.
- All 42 discovered `GSharp.Core.dll` copies were nonzero; the five ordinary copies had matching MD5 `87c3a854f08e2de5945c65ae89d0e71f`.
- `DiagnosticIdUniquenessTests`: 4 passed, confirming 376 descriptors remain documented with matching severities.
- `Core.Tests` exhaustiveness selector: 26 passed, 0 failed, 0 skipped.
- `RefactoringBaselineTests`: 137 passed, 0 failed, 1 skipped; six intentional switch-expression IL hashes regenerated.
- `Compiler.Tests` selector: 2 passed, 0 failed, 0 skipped.
- `Interpreter.Tests` selector: 3 passed, 0 failed, 0 skipped.
- Emitted regression test also performs `Assembly.Load(File.ReadAllBytes(...))`, `GetTypes()`, bounded child execution, and `IlVerifier.Verify`.

Fixes #2941
Fixes #2971

Final CI rollup: **9 SUCCESS, 3 SKIPPED**.


DavidObando:
## Review: NOT MERGEABLE (cheap path to green — see "What would make this mergeable")

Reviewed `b94a163ee` against base `a6b1561f2` (= current `main` exactly, no rebase needed). Scratch clone, `-m:1` Release build, exit-code gated, 4 matching non-zero `GSharp.Core.dll` copies (`4e73be69…`), `gsc`/`gsi` invoked directly. Head build was **0 warnings / 0 errors**.

I want to be clear up front that the 496 lines of tests and the two documentation edits are real value, and issue #2941 asks for a decision plus documentation, which this substantially delivers. The problems are that the production change does nothing, two claims in the PR body are not accurate against current `main`, and — the reason for the verdict — the definition being adopted leaves a runtime hole that the two engines answer differently.

---

### 1. BLOCKER — the adopted definition creates an unmatched-value path, and `gsc` and `gsi` disagree on it

This PR defines a flags-enum switch as exhaustive when all **named constants** are covered. That is a defensible choice (see §2), but it makes combined values reachable at runtime with **no matching arm by design**. The PR does not test that path, does not document it, and the two engines do not agree about it:

```gsharp
package G8
import System
@Flags
enum Access { None = 0, Read = 1, Write = 2 }
func F(x Access) int32 {
    return switch x {
        case Access.None: 0
        case Access.Read: 1
        case Access.Write: 2
    }
}
Console.WriteLine(F(Access.Read | Access.Write).ToString())
```

Compiles clean on both base and head — no GS0176, no GS0177, which is exactly what the new rule prescribes. Then:

| engine | result |
|---|---|
| `gsc` (compiled, executed) | prints **`0`** |
| `gsi` | **`GS9999: Object reference not set to an instance of an object.`** |

Isolated so this is not misread as an unrelated interpreter gap — `gsi` handles both halves fine on their own:

| probe | `gsi` |
|---|---|
| `F(Access.Read)` (arm matches) | `1` ✔ |
| `Access.Read \| Access.Write` printed directly | `3` ✔ |
| `F(Access.Read \| Access.Write)` (no arm matches) | `GS9999` NRE ✘ |

So the divergence is precisely the unmatched-value path this definition opens up.

Two separate problems here:

- **`gsc` returns `0` silently.** Not a diagnostic, not an exception — the default of the arm type. Worse, `0` is also the legitimate result of the `Access.None` arm, so the failure is indistinguishable from a correct answer at the call site.
- **`gsi` crashes with an NRE surfaced as `GS9998`/`GS9999`-class output**, so the same program means two different things depending on how it is run.

To be fair and explicit: **this is not a regression.** It reproduces identically on base `a6b1561f2` and on `8dddbbf2`, which follows from the production change being inert (§3). I am not asking this PR to fix the interpreter. But this is a *semantics-definition* PR whose stated deliverable is defining these semantics **consistently**, and it ships `test/Interpreter.Tests/Issue2941FlagsEnumExhaustivenessTests.cs` as parity coverage. An engine disagreement at the exact boundary the definition creates is the deliverable being incomplete, not out-of-scope scope creep.

The documentation the PR adds stops one sentence short:

> **Enum exhaustiveness.** Enum switches cover named constants only.
> `[Flags]` does not add unnamed bit combinations as required switch arms.
> This rule is identical for imported CLR enums and source-declared enums.

Correct as far as it goes, and it never says what happens when an unnamed combination is switched on. That is the one thing a reader hitting this needs.

---

### 2. The definition itself is the right choice

Recording this since it is question one and the answer is yes. "All representable values" is unusable for a flags enum — the value space is the power set, so a 3-member flags enum would demand 8 arms and a 32-bit one would demand 2^32. Covering named constants matches C#'s treatment and keeps the rule identical for imported CLR enums and source-declared ones, which the test matrix does pin properly. No objection to the semantics; the objection in §1 is that the definition's consequence is undocumented and unenforced, not that the definition is wrong.

---

### 3. SHOULD-FIX — the 5-line production change is inert, and nothing in the PR detects that

The entire production delta is one early return in `AnalyzeSwitchExpression` (`src/Core/CodeAnalysis/Binding/ExhaustivenessAnalyzer.cs:45-49`):

```csharp
if (!IsExhaustiveDiscriminant(discriminantType))
{
    return;
}
```

It cannot change any diagnostic on any program. Three independent lines of evidence:

**(a) Structural proof (algebraic reduction, not a survived mutant).** `IsExhaustiveDiscriminant` (`:24-28`) is true for exactly `{ EnumSymbol, ClrType.IsEnumSafe(), InterfaceSymbol{IsSealed}, StructSymbol{IsSealedHierarchy} }`. `TryGetMissingVariants` (`:81`) dispatches on exactly that same set — `TryGetEnumVariants` (which re-guards `IsEnumSafe` itself at `:210`), sealed-interface at `:133`, sealed-class at `:162`, then `return false` at `:194`. The guard's true-set **equals** the set on which the analysis can return true, so "guard false → return" and "no guard" both produce no diagnostic.

**(b) The PR's own 28 tests pass with the production file reverted to base.** Core 26 + Compiler 1 + Interpreter 1, all green without the change. Verified with a bidirectional round-trip so this is not a stale binary: `GSharp.Core.dll` `4e73be69` (head) → `baf6c78e` (reverted) → `4e73be69` (restored).

**(c) Probe battery across the guard's only behavioural surface** — non-enum discriminants in switch *expressions* — base vs head `diff` **empty**: `int32` → GS0176 both, `string` → GS0176 both, `bool` → GS0176 both, incomplete enum → GS0177 both, complete flags enum → clean both.

The reason the 26-case matrix cannot catch this: `MatrixCases()` is `EnumOrigin × SwitchForm × SwitchShape`, and all three origins (`ImportedFlags` → `StringSplitOptions`, `UserFlags` → `@Flags enum Access`, plain `enum Access`) are **enums**. Enums always satisfy `IsExhaustiveDiscriminant`, so the early return is never taken in any of the 26 cases. The matrix is orthogonal to the production change by construction.

Either delete the guard — the structural proof above is an algebraic reduction, which is the bar for deletion rather than a mere coverage survival — or add a case that fails without it. I could not construct the latter, which is itself the argument for the former.

---

### 4. SHOULD-FIX — two claims in the PR body do not hold against current `main`

- **"define enum exhaustiveness over named constants only; `[Flags]` does not add unnamed bit combinations."** `main` already behaves this way. `git grep FlagsAttribute a6b1561f2 -- src/Core` returns **zero** matches — the #2926 carve-out was already reverted. The complete-flags-enum probe compiles clean on base and head alike. This PR pins and documents that behaviour; it does not establish it.
- **"preventing GS0176 and GS0177 from being emitted together."** Not reachable on current `main`. The guard is enum-irrelevant by construction (enums always pass `IsExhaustiveDiscriminant`), so it cannot affect any enum program, and I could not produce a co-fire on either side — including a guarded-arm case (`case Access.Write when x == Access.Write`), which yields GS0177 alone. The double-diagnostic required the reverted carve-out.

Worth correcting because a future reader will otherwise treat both as regression-protected behaviour introduced here.

---

### 5. Other review questions, answered

**Consumers of exhaustiveness — no movement, and here is why.** `IsExhaustive` is written at `BoundPatternSwitchStatement.cs:27`, copied by `BoundTreeRewriter.cs:794` / `CaptureBoxingRewriter.cs:801` / `Lowerer.cs:1495`, and **read in exactly one place, `ControlFlowGraph.cs:318`**. `ControlFlowGraph.Create` has three consumers — `AllPathsReturn` (iterator-exempt), `RefKindDefiniteAssignmentAnalyzer.cs:108` and `RefStructAsyncLivenessAnalyzer.cs:129` (both unguarded). This PR **cannot** move definite-assignment or ref-struct liveness, because it never touches the statement path (`AnalyzeSwitchStatement:60`, already gated at `:67`), and the statement path is the sole producer of `IsExhaustive`. The two unguarded analyzers are a real hazard for any *future* exhaustiveness change; they are not one for this diff.

**Accepted-program space — unchanged**, which follows from §3 and is confirmed by the probe battery. No previously-accepted program now errors and no previously-required `default` became unnecessary.

**Collision with #2979 — no file conflict.** #2979 touches only `docs/adr/0009-switch-semantics.md`; this PR touches `docs/diagnostics.md` and `website/docs/ref/diagnostics.md`. ADR-0009 on `main` mentions exhaustiveness only at line 25 (sealed hierarchies and enums, no flags claim), so #3005 does not contradict it and the two can land in either order. Separately, and not this PR's problem: #2979's added line — *"Its sole production reader is definite-return control-flow analysis, which exempts iterator-returning functions"* — is **false** on both counts, per the reader map above.

---

### What would make this mergeable

None of this requires fixing the interpreter in this PR.

1. **Document the unmatched-combination runtime behaviour** in the docs paragraph already being edited — one sentence stating that a value with no matching arm yields the arm type's default under `gsc`, since that is the direct consequence of the rule being defined immediately above it.
2. **File the `gsi` NRE as its own issue** and reference it here, so the parity gap is owned rather than implied to be covered by the parity test file this PR adds.
3. **Either delete the inert guard** on the structural proof in §3, **or** add a test that fails without it.
4. **Correct the two PR-body claims** in §4 so the flags rule reads as pinned-and-documented rather than introduced, and drop the GS0176/GS0177 co-fire claim.

Items 3 and 4 are cheap. Item 1 is one sentence. Item 2 is the one that matters, and it is a `gh issue create`.

DavidObando:
## Post-merge gate check: this PR will **not** red `main` on the new diagnostic-docs invariant

Scope note: this is **not** a merge verdict — the functional blocker being worked separately is untouched here. This answers only *"does #3005 break the completeness check that #2981 just made mandatory?"* The answer is no, with margin.

Checked at head `b94a163ee`, confirmed equal to `refs/pull/3005/head`, natural base `a6b1561f2`, against `main` = `a692200af` (which now contains #2981).

### Why the exposure is lower than it looks

The concern was that #3005 adds diagnostic-ID rows to two documentation files under a gate its branch CI never ran. It does not add any.

```
docs/diagnostics.md          -| GS0177 | Error | ... One or more enum members not covered ...
                             +| GS0177 | Error | ... One or more named enum constants not covered ...
                             -| GS0178 | Error | ...
                             +| GS0178 | Error | ...
website/docs/ref/diagnostics.md   (identical pair)
```

**Zero new IDs; two existing rows reworded, severity untouched (`Error` → `Error`), applied identically to both files.** `GS0177`/`GS0178` are `DiagnosticSeverity.Error` at `DiagnosticDescriptors.cs:95-96`, so the rows still match their descriptors. The remaining `+4` lines in `docs/diagnostics.md` are the prose "Enum exhaustiveness" note at `:55`, not table rows.

### The four checks

**1. No duplicate rows.** No row is added, so no ID can gain a second row. The merged tree has **zero** IDs with conflicting severities. (`GS9998` and `GS9999` each have two rows, but both pairs are `Error`, so `Distinct()` yields one value and the new `conflictingSeverities` assert passes cleanly — and neither has a descriptor, so neither is looked up at all.)

**2. Severity matches.** Measured on the merged tree: **376 descriptors joined against the documentation, 0 mismatches.**

**3. No descriptor added without a row — falsified structurally, not by grep.** Three independent arguments, strongest first:

- `DiagnosticDescriptors` is a **single, non-partial** `internal static class` with exactly one declaration site (`src/Core/CodeAnalysis/DiagnosticDescriptors.cs:7`). `GetDescriptorFields()` reflects over `typeof(DiagnosticDescriptors)`, so a descriptor it can see **must** be declared in that one file. Declaration style is irrelevant — multi-line, reordered modifiers, factory helper, all impossible to hide.
- #3005 does not touch that file. Its six files are the two docs files, `ExhaustivenessAnalyzer.cs`, and three test files.
- Style-agnostic sweep: **zero** added lines anywhere in the diff so much as mention the identifier `DiagnosticDescriptor`.

**4. Row placement — not applicable, and worth recording why.** #3005 adds no rows, so it is not in the append contest at all. Its two `docs/diagnostics.md` hunks are at `:55` and `:134-141`. For everyone else: on current `main` the `GS05xx` run is `:1058`–`:1127`, ending at `GS0508`, and `GS0506` landed at `:1115`. The historical "963–980" anchor no longer contains the `GS05xx` block; **the live append point is below `:1127`.**

### Predicted post-merge triple

Measured on the actual merge result rather than predicted — `git merge-tree` against `a692200af` produces a **clean merge** (tree `ad6212f55`), and `main`'s current `GS0177`/`GS0178` rows match this PR's pre-image byte-for-byte, so the reword applies as intended:

| | value |
|---|---|
| descriptors | **376** (unchanged — no descriptor added or removed) |
| documented rows | **380** (unchanged — no row added, two reworded in place) |
| undocumented | **0** |
| severity mismatches | **0** |
| conflicting-severity duplicates | **0** |

Unchanged from `main`'s current 376 / 380 / 0, which is the expected result for a description-only edit.

### One thing worth knowing, not a defect in this PR

The check reads **only** `docs/diagnostics.md` — there is no reference to `website/` anywhere in `DiagnosticIdUniquenessTests.cs`. So `website/docs/ref/diagnostics.md` is entirely unenforced, and this PR keeps the two files in sync **voluntarily**. That is the right behaviour and worth keeping, but nothing would have caught it had the website edit been omitted or made inconsistent. It is the still-open half of #2970.

### Evidence class

Static analysis of the merged tree — descriptor extraction, row extraction, and the severity join were run against the tree `git merge-tree` actually produces, not against either side separately. **No build and no test run** (test host contention). This verdict covers the diagnostic-docs invariant only; the functional behaviour of the `ExhaustivenessAnalyzer` change and this PR's own three test files are outside it and unaffected by anything above.

DavidObando:
Methodology alert applied retroactively. Mutation outcomes are unchanged: guarded builds refreshed GSharp.Core.dll, both mutants remain detected, and restored controls pass. One initial mutation placement failed SA1202; that probe was aborted and not counted. I also corrected the mixed-base suite table: executed a6b1561f base counts are Core 7028+1 skip, Compiler 3906, Interpreter 491, LSP 452; head b94a163e adds exactly 26/1/1/0 tests. PR body now contains full freshness evidence.

DavidObando:
### Re-verified at `main` = `f70626452` (32 PRs later) — verdict unchanged

The earlier check was measured against `a692200af`. `main` has since advanced 32 merges, so I re-ran rather than re-asserting. Head is still `b94a163ee`; natural merge base still `a6b1561f2`.

**Merge is still clean** against `f70626452` (tree `bf4aeb218`), and `main`'s current `GS0177`/`GS0178` rows at `:137-138` still match this PR's pre-image byte-for-byte, so the reword applies as intended.

Measured on that merged tree:

| | |
|---|---|
| descriptors | **376** |
| documented rows | **380** |
| joined | **376** |
| missing rows | **0** |
| severity mismatches | **0** |
| conflicting-severity duplicates | **0** |

Unchanged from `main`. **This PR will not red `main` on the diagnostic-docs gate.**

### Reconciling the 380-vs-384 count

Both numbers are correct; they count different populations, and the distinction matters for anyone else measuring this gate.

- **384** = every `GS####` *token appearing anywhere* in `docs/diagnostics.md`.
- **380** = IDs the check actually parses, i.e. lines matching `^\|\s*(GS\d{4})\s*\|\s*(?:\*\*)?(Error|Warning|Info)`.

The four in the gap are `GS0217`, `GS0234`, `GS0345`, `GS0419` — all **retired or reserved**. `GS0217` and `GS0419` have table rows whose severity cell reads `_Retired_` / `Reserved`, so the regex correctly declines them; `GS0234` and `GS0345` appear only in prose explaining their retirement. They are supposed to have no descriptor field and are supposed not to be enforced.

So the arithmetic closes on rows: **376 descriptors + 4 documented-but-fieldless = 380 rows.** The documented-but-fieldless set is exactly `GS0190` (Error), `GS9100` (Warning), `GS9998` (Error), `GS9999` (Error) — verified on the merged tree. `GS9998`/`GS9999` each have two rows, both `Error`, so they group to a single distinct severity, stay out of `conflictingSeverities`, and are then never looked up because the check is one-way over descriptors.

### The `GS0510` severity trap does not apply here

This PR contains **zero** `GS05xx` mentions across its entire diff. It adds no diagnostic-ID rows at all — it rewords the description cells of two existing rows, `GS0177` and `GS0178`, identically in both docs files, leaving severity at `Error`:

```
-| GS0177 | Error | ... | One or more enum members not covered and no `default` arm. |
+| GS0177 | Error | ... | One or more named enum constants not covered and no `default` arm. |
```

Both descriptors are `DiagnosticSeverity.Error`, so the rows still match.

### No descriptor added — confirmed with the stronger instrument

`= new("GS` across `src/` matches exactly one file, `src/Core/CodeAnalysis/DiagnosticDescriptors.cs`, and that file declares `internal static class DiagnosticDescriptors` with **no `partial`** modifier. Since `GetDescriptorFields()` reflects `typeof(DiagnosticDescriptors)`, every descriptor it can observe must be declared there — no declaration style can hide one, and there is no second site. This PR does not touch that file, and the descriptor count is **376 before and after**.

### Placement

Not applicable — no rows added, so this PR is not in the append contest. Its two `docs/diagnostics.md` hunks are the prose note at `:55` and the row edits at `:134-141`, far from the `GS05xx` run, which on current `main` sits at `:1058`–`:1127` in a 1153-line file, ending at the `GS0508` row. That `:1127` block remains the correct append point for PRs that do add rows.

### Scope and evidence class

This answers only *"will it red `main` on the docs gate"* — **no**. It is not a merge verdict; the functional blocker and this PR's three test files are outside it. Static measurement over the tree `git merge-tree` actually produces; no build, no test run.

DavidObando:
Fixed the unmatched flags-enum value blocker and pushed `8025f21dce077ef2d0f5b146de0999ec9cd0850c` after rebasing onto `210126dd90f4ab3edbe1b7f9219946594685afc7`.

What changed:

- Emitted switch expressions now throw `InvalidOperationException("Unmatched switch expression value.")` when no arm matches and no default arm exists, rather than loading an unassigned result local.
- The evaluator now raises an `EvaluatorException` with the same message, so `gsi` reports a deliberate `GS9999` runtime error rather than a `NullReferenceException`.
- Added `gsc` and `gsi` regression tests for `Access.Read | Access.Write`, including assembly load/`GetTypes()`, bounded child execution, and IL verification.
- Documented unmatched-value behavior in both live diagnostic references.
- Removed the inert five-line switch-expression discriminant guard identified in review.
- Updated the PR body to remove the obsolete GS0176/GS0177 and “separate defect” claims. This PR now also fixes #2971.

Base-to-head differential, run directly:

| Engine | Base `a1237ded7` | Head `8025f21dc` |
|---|---|---|
| `gsc` | compile exit 0; program exit 0; stdout `0` | compile exit 0; program terminates nonzero with `System.InvalidOperationException: Unmatched switch expression value.` |
| `gsi` | exit 1; `GS9999: Object reference not set to an instance of an object.` | exit 1; `GS9999: Unmatched switch expression value.` |

Mutation proof: reverting `278731b9f` restored `gsc` stdout `0` and the `gsi` NRE; restoring the commit restored both explicit runtime failures.

Validation:

- Release CI-style solution build: exit 0, 0 warnings, 0 errors.
- `DiagnosticIdUniquenessTests`: 4 passed; all descriptor rows remain complete and severity-matched.
- `Core.Tests` exhaustiveness selector: 26 passed.
- `RefactoringBaselineTests`: 137 passed, 1 skipped; six intentional switch-expression IL hashes regenerated after the first CI run exposed the baseline gate.
- `Compiler.Tests`: 2 passed.
- `Interpreter.Tests`: 2 passed.
- 42/42 discovered `GSharp.Core.dll` files were nonzero; the five ordinary copies agreed at MD5 `9755862ee78faeb7dc5173249c2d3e4f`.





Rebase-only update: `c07cb6b7c` → `8025f21dc`; production, tests, docs, and baselines are content-identical.

Final CI rollup: **9 SUCCESS, 3 SKIPPED**.

DavidObando:
# Review — PR #3005 @ `8025f21dc` — **NOT MERGEABLE** (one blocker; the fix is one line and I verified it)

This is good work. The design rule is right, the 24-case matrix is the best-shaped test surface I have reviewed this week, and the body honestly discloses the engine asymmetry and the six regenerated baseline hashes. Two things stop it.

Measured on the **merged tree** (base `210126dd9` is several merges behind): `git merge refs/pull/3005/head` into `main` `8acc2cc76` → **rc=0, clean**, merged head `a1d301c26`, `gsc`+`gsi` Release builds **rc=0 / rc=0**.

---

# BLOCKER 1 — `Fixes #2971` is declared, but #2971's stated acceptance criterion is still observably false

#2971's *Expected*:
> **Compiled and interpreted execution should agree** and should not silently manufacture `default(T)`.

The second half is fixed. The first half is not. Probe — same source, both engines, exit codes captured into `RC=$?` before anything else ran:

```gs
try {
    Console.WriteLine(F(Access.Read | Access.Write))
} catch (e InvalidOperationException) { Console.WriteLine("caught-ioe") }
  catch (e Exception)                 { Console.WriteLine("caught-generic") }
Console.WriteLine("after")
```

| tree | `gsc` → `dotnet out.dll` | `gsi` |
|---|---|---|
| `main` `8acc2cc76` | `0` / `after`, **rc=0** | `caught-generic` / `after`, rc=0 |
| this PR, merged | `caught-ioe` / `after`, rc=0 | **`caught-generic`** / `after`, rc=0 |

The PR correctly kills the silent `0`. But a G# program that catches `InvalidOperationException` **takes a different branch under the two engines**, and **both exit 0** — no diagnostic, no nonzero status. That is the silent-divergence shape, not a loud one.

Cause: `Evaluator.Concurrency.cs:61` throws `EvaluatorException` with **no inner exception**, so `TryFindCatchHandler` sees type `EvaluatorException` and only the `Exception` arm matches. The compiled path throws the real `InvalidOperationException`.

**Verified fix, one line at `Evaluator.Concurrency.cs:61`:**

```diff
-        throw new EvaluatorException("Unmatched switch expression value.", node);
+        throw new EvaluatorException(
+            "Unmatched switch expression value.",
+            new InvalidOperationException("Unmatched switch expression value."),
+            node);
```

I applied it, rebuilt, and re-ran:

```
gsi with fix   -> caught-ioe / after      (parity reached)
Interpreter.Tests --filter Issue2941  rc=0  Failed: 0, Passed: 2
gsi restored   -> caught-generic / after  (flip back)
```

The ctor already exists (`Evaluator.Expressions.cs:110` uses it), `EvaluatorException.Message` is unchanged so **your `GS9999` + message assertions still pass verbatim**, and the tree was restored clean. This is drop-in.

It also makes the new docs paragraph literally true. As written —
> compiled and interpreted execution fail with `Unmatched switch expression value.`

— it reads as parity to a language user, and the message *is* identical; only the observable type is not. Landing the one-liner removes the caveat instead of requiring one.

---

# SHOULD-FIX 1 — the body lists a production change that is not in the diff

Summary bullet 5 and Production-changes bullet 3:
> `ExhaustivenessAnalyzer.cs`: delete the inert expression-only guard…
> Remove the inert switch-expression discriminant guard identified in review.

`src/Core/CodeAnalysis/Binding/ExhaustivenessAnalyzer.cs` is **byte-identical at all three trees**:

```
1b54e47e6e813cfe7ded8f57633a1d9d   base 210126dd9
1b54e47e6e813cfe7ded8f57633a1d9d   PR head 8025f21dc
1b54e47e6e813cfe7ded8f57633a1d9d   main 8acc2cc76
```

The PR's 8-file diff does not include it. The **end state is correct** — `AnalyzeSwitchExpression` (`:38-49`) already has no `IsExhaustiveDiscriminant` guard, while `AnalyzeSwitchStatement:67` keeps its guard, which is load-bearing for definite-return. So this is a stale claim, not a wrong tree. Delete both bullets: an unreliable production-change list is what forced #2997 to a second round.

# NIT 1 — the zero member is never exercised at runtime

`Access.None = 0` / `StringSplitOptions.None` are arms in every switch, but the runtime calls in `Issue2941FlagsEnumExhaustivenessTests.cs` pass `RemoveEmptyEntries`, `TrimEntries`, `Read`, `Write`, `Read`, `Write`. Zero is the one flags value most likely to be special-cased in a bitwise dispatch. Adding it to the expected output line is a one-token change.

---

## What I checked that holds

**`[Flags]` combinatorics — covered.** `MatrixCases()` is a genuine cartesian product: 3 `EnumOrigin` (`ImportedFlags` = real CLR `System.StringSplitOptions`, `UserFlags` = `@Flags`, `PlainEnum`) × 2 `SwitchForm` × 4 `SwitchShape` = **24 cases**, each asserting exact diagnostic IDs *and* locations *and* `GS0100` definite-return flow. The combined value `Access.Read | Access.Write` is exercised at runtime in both the emit and interpreter suites, and `IncompleteEnumSwitchExpression_ReportsOneDiagnosticAtSwitch` uses `Assert.Single` to pin the GS0176/GS0177 mutual exclusivity #2941 asked for. `LiveDiagnosticDocs_StateNamedConstantFlagsRule` pins both docs copies against drift.

**No M-B hazard.** Values are distinct everywhere it matters: `11/22/31/42/51/62` across six functions, so a wrong-function or wrong-arm dispatch cannot alias.

**Docs gate — verified, not inherited.** Counting the exact regex the mandatory test uses:

```
base 210126dd9 : gate_rows=382  distinct_row_ids=382  distinct_tokens=384
PR   8025f21dc : gate_rows=382  distinct_row_ids=382  distinct_tokens=384
backticked ID rows in PR: 0
DiagnosticDescriptor lines added/removed in the diff: 0
```

No row added, **none deleted**, `GS0177`/`GS0178` edited in place with bare IDs and untouched `Error` cells. Both the "adds 0 rows" and the "orphans a descriptor" hazards are clear.

**Baseline churn is legitimate and declared.** Six `refactoring-baseline.json` hashes change with **no `samples/*.gs` edited** — they are emit-output hashes moving because `MethodBodyEmitter.Patterns.cs` now appends `newobj`/`throw` to defaultless switch expressions. My count of six matches the body's "six intentional switch-expression IL hashes regenerated".

**Both declared issues are real.** #2941 (define + apply + document + test) and #2971 (silent `0` / NRE) — I reproduced #2971's exact base symptoms on `main` and watched both flip. Nothing is closed silently.

---

**Land the one-liner and drop the two `ExhaustivenessAnalyzer` bullets, and this is MERGEABLE.** All scratch trees restored, `git status --porcelain` empty.

DavidObando:
Addressed the remaining blocker in `537eeee6e0ffb103cc1ede4a6a3177ea511bc1ae`, rebased onto `4bba0643a001ad81b07862980a75b0a25c1b021b`.

Content changes:

- `Evaluator.Concurrency.cs` now gives the unmatched-switch `EvaluatorException` an inner `InvalidOperationException` with the same message. `TryFindCatchHandler` therefore observes the same runtime exception type as emitted code without fighting the existing repeated wrapper unwrapping from #3023.
- Added `UnmatchedFlagsValue_PreservesInvalidOperationExceptionType`, which executes the combined flags value under typed and generic catches and requires exact output `caught-ioe\nafter`.
- Removed both stale `ExhaustivenessAnalyzer.cs` claims from the PR body. That file remains absent from the eight-file net diff.

Measured directly, with unpiped processes:

| Engine | Exit | Output |
|---|---:|---|
| `gsc` → emitted program | 0 | `caught-ioe` / `after` |
| `gsi` | 0 | `caught-ioe` / `after` |

Mutation proof:

- Control: interpreter selector passed 3/3.
- Removed only the inner exception: selector failed 1/3; exact divergence was expected `caught-ioe`, actual `caught-generic`.
- Restored the line and rebuilt: selector passed 3/3 and direct `gsi` output returned to `caught-ioe` / `after`.

Final Release CI-style solution build exited 0 with 0 warnings and 0 errors. All 42 discovered `GSharp.Core.dll` copies were nonzero; the five ordinary copies agreed at MD5 `87c3a854f08e2de5945c65ae89d0e71f`.

Final CI rollup: **9 SUCCESS, 3 SKIPPED**.

DavidObando:
## PR #3005 — **MERGEABLE**

Re-established from scratch, nothing inherited. Base `4bba0643a` (computed `merge-base`, and it matches the rebase target named in your body), head `537eeee6e`, `main` `fdba684b6`, **merged tree `3a46aacff`** (`git merge` rc=0). Three-dot = **8 files / +636 −11**. (Two-dot reports 58 files / +908 −1299 — phantom.)

Everything below was measured on the **merged** tree, because this branch is based behind #3043.

### Docs gate — safe, but the base-side number is a trap

| file | base `4bba0643a` | `main` | **merged** |
|---|---|---|---|
| `docs/diagnostics.md` | 383 | 383 | **383** |
| `website/docs/ref/diagnostics.md` | **305** | 381 | **381** |

The website file gained 76 gate-visible rows on `main` *after* this branch's base (#3043). Anyone checking "does it add rows?" against the base would be reading 305 and comparing to the wrong population. On the **merged** tree both counts are **identical to `main`** — 0 rows added, 0 deleted.

Also verified: both files receive a **byte-identical** edit; no new or renamed IDs (`new("GS…")` in `src/` = **0**; positive control 377 in `DiagnosticDescriptors.cs`); `GS0177`/`GS0178` keep bare, gate-visible form with severity `Error` unchanged — only the "Example trigger" column and a `>` blockquote changed, and a blockquote cannot match `^\|`. **0** backticked IDs introduced.

### "Consistent across switch forms" — I counted the sites myself

Enum exhaustiveness has exactly **two** report sites, both reached via `ExhaustivenessAnalyzer.IsExhaustiveDiscriminant`:

- `src/Core/CodeAnalysis/Binding/ExhaustivenessAnalyzer.cs:47` → `GS0177` (expression)
- `src/Core/CodeAnalysis/Binding/ExhaustivenessAnalyzer.cs:74` → `GS0178` (statement)

The test matrix's `SwitchForm { Statement, Expression }` covers **2 of 2**. No missed form — this is not the #3020 "5 of 7" shape.

`BoundPatternSwitchStatement` construction sites: still exactly **4** (`Lowerer.cs:1495`, `CaptureBoxingRewriter.cs:801`, `BoundTreeRewriter.cs:794`, `StatementBinder.Blocks.cs`), all passing the flag explicitly; sole read `ControlFlowGraph.cs:318`. **No fifth site added** — the silent-`false` hazard is not triggered.

### Tests are not vacuous

The #2979 shape (`DoesNotContain(IsError)` as the only assertion, which cannot fail) does **not** apply here:

- `Issue2941…Tests.cs:60` `Assert.Equal(expected.Length, diagnostics.Length)` — exact **count**
- `:63` `Assert.Equal(expected[i].Id, diagnostics[i].Id)` — exact **ordered IDs**
- `:69-70` asserts `result.Success` **and** `peStream.Length > 0` — emission actually happened
- `:92-94` `Assert.Single` + exact ID + exact **source span**
- the single `DoesNotContain(… == "GS0100")` at `:79` sits inside a case that already asserts an exact ID list, so it is not carrying the test

Matrix is a full cross-product: 3 origins × 2 forms × 4 shapes = 24, plus 2 facts = **26**. Distinct expected values throughout (10/20/30/40/50/60) — the M-B lesson applied.

Two things I would call out as better than required:

- `Issue2941…Tests.cs:98-107` asserts the rule prose is literally present in **both** `docs/diagnostics.md` *and* `website/docs/ref/diagnostics.md`. That is a real bidirectional docs pin, not a comment.
- `test/Compiler.Tests/Emit/Issue2941…Tests.cs:107-108,124-125` does `Assembly.Load(File.ReadAllBytes(...))` + `GetTypes()` **in addition to** `IlVerifier.Verify(:117,:134)` — exactly the #3001 lesson, where a real emit defect passed `Verify()` outright.

### Mutation — each production hunk individually

Clean `GSharp.Core.dll` md5 `b5ad19f0072b855c7cf9802142087aeb`.

**M-A** — delete the `else { … throw }` at `src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Patterns.cs:144-152`:

- build **rc=0**, 0 errors, md5 → `9d52de0b…` (valid mutant in both directions)
- **value flips**: compiled probe prints `0` instead of `caught-ioe` — silent wrong answer, exit 0
- `Compiler.Tests` → **rc=1, 1 failed / 1 passed**: `UnmatchedFlagsValue_FailsLoudly`

**M-B** — revert `src/Core/CodeAnalysis/Evaluator.Concurrency.cs:56-64` to `return EvaluateExpression(defaultArm.Result);`:

- build **rc=0**, 0 errors, md5 changed
- **value flips**: `gsi` prints `caught-generic` instead of `caught-ioe`
- `Interpreter.Tests` → **rc=1, 2 failed / 1 passed**: `UnmatchedFlagsValue_PreservesInvalidOperationExceptionType`, `UnmatchedFlagsValue_ReportsRuntimeError`

Restored: md5 back to `b5ad19f0072b855c7cf9802142087aeb` **byte-identical**, `gsi` back to `caught-ioe`, `git status --porcelain` empty.

M-A is additionally pinned by `RefactoringBaselineTests` (the six regenerated IL hashes), so that hunk is double-covered.

### `gsc` vs `gsi` parity — the divergence is gone

Run through both CLIs independently of the harness, on the merged tree:

| | `gsc` | `gsi` |
|---|---|---|
| `@Flags` combined value `Read\|Write` | `caught-ioe` / `after` | `caught-ioe` / `after` |

An earlier head of this PR diverged here (`caught-ioe` compiled vs `caught-generic` interpreted, both exit 0 — the silent class). Commit `537eeee6e` closes it, and `Evaluator.Concurrency.cs` doing it by attaching an inner `InvalidOperationException` is the right mechanism: `UnwrapRuntimeException` feeds `TryFindCatchHandler`'s **type** matching, so the typed catch matches the same way the CLR matches the emitted `throw`.

### `Fixes #2971` — checked, genuinely fixed

The second declared issue's repro is a different shape (`DateTimeKind(99)`, an out-of-range cast on a **non-`[Flags]` imported** enum). Ran it verbatim on the merged tree:

- `gsi`: **rc=1**, `GS9999: Unmatched switch expression value.` (was `Object reference not set to an instance of an object.`)
- `gsc`: **rc=134**, `System.InvalidOperationException: Unmatched switch expression value.` at `Default.<Program>.F(DateTimeKind x)` (was a silent `0`)

Both engines now fail loudly and identically. The `Fixes #2971` claim holds.

### Body accuracy

Every number in the body reproduced exactly against my independent runs: `Core.Tests` **26**, `RefactoringBaselineTests` **137 passed / 1 skipped**, `Compiler.Tests` **2**, `Interpreter.Tests` **3**, **6** regenerated baseline hashes, and the base-side `gsc` stdout `0` matches my M-A measurement precisely. The "Production changes" section lists exactly the two files in the diff, and its explicit claim that `ControlFlowGraph.cs` remains unchanged is correct — it is absent from the diff. `a1237ded7` is a real ancestor of the head.

`RefactoringBaselineTests` passes **137/138 on the merged tree**, so the baseline regenerated at the old base is not stale post-merge.

---

### BLOCKER
None.

### SHOULD-FIX
- **`test/Interpreter.Tests/Issue2941FlagsEnumExhaustivenessTests.cs` / `test/Compiler.Tests/Emit/Issue2941FlagsEnumExhaustivenessTests.cs`** — the PR declares `Fixes #2971`, but there is **not one cast-to-enum shape** in any of the three new test files (grep for a value-cast such as `DateTimeKind(99)` returns 0). #2971's own repro is therefore unpinned. I verified by hand that it works, and it provably traverses the identical "no arm matched → throw" path that M-A/M-B already kill, so the regression risk is low — but a declared issue with zero tests in its own shape is one `InlineData` away from being closed properly.

### NIT
- Runtime coverage exercises the *combined* unnamed value (`Read|Write`) and the *named* zero (`Access.None = 0`, which is an arm), but no value outside any bit combination. Same code path; cheap to add alongside the item above.
- `#2979` is an ADR on switch/iterator exhaustiveness metadata currently in rework. This PR records the named-constant rule as prose in `docs/diagnostics.md` (and pins that prose with a test at `:98-107`). If #2979 lands a differently-worded normative statement, that test will fail on wording alone — worth sequencing the two rather than discovering it at merge.

Issue #3007, opened 2026-08-01
Compiler.Tests: IlVerifier tool probes still use unbounded sequential ReadToEnd


DavidObando:
`IlVerifier.RunProcess` was bounded by PR #2966 (closes #2934), but two sibling probe sites were left unchanged and are byte-identical to that PR's merge base:

- `test/Compiler.Tests/IlVerifier.cs:390` — `TryProbeDotnetToolRun`
- `test/Compiler.Tests/IlVerifier.cs:427` — `TryProbeOnPath`

Both still use **sequential `ReadToEnd()` followed by an unbounded `WaitForExit()`**. Reading one pipe to completion while the other fills is the classic child-process deadlock, and the comment at `IlVerifier.cs:172-177` documents that exact hazard (it sits above the *correct* code; the offenders are ~215 lines further down).

## Why this was not a merge gate

Measured on PR #2966's head. The shape is genuinely wedgeable — a standalone harness in the probe shape, given 128 KB on each pipe, wedges permanently. But the probes invoke a fixed `--version` with no user-controlled input, and every reachable state was measured:

| state | stdout | stderr |
|---|---:|---:|
| success | 68 B | 0 B |
| no manifest | 0 B | 77 B |
| PATH probe | 68 B | — |

Worst case is **77 bytes against a ~64 KB pipe buffer** — roughly 850× of headroom. So this is latent, not live.

## Why it is still worth fixing

Both probes are resolved **once and cached**, so they run before any test body executes. A wedge there hangs the entire test class rather than a single test, and there is no outer net: `AssemblyInfo.cs:9` sets `DisableTestParallelization = true`, and under that setting **xUnit 2.9.2 ignores `[Fact(Timeout = …)]`**. `:390` is reachable in CI (step 1, always runs on the first `Verify()`); `:427` is not, since it runs only when probe 1 fails and CI restores the manifest.

## Important: the obvious fix does not work

**`ReadToEnd()` runs first and is also unbounded, so `WaitForExit()` is never the blocking call.** A patch that only adds a timeout to `WaitForExit()` changes nothing. The fix must bound the **reads** — concurrent `ReadToEndAsync()` on both pipes, then `WaitForExit(timeout)`, `Kill(entireProcessTree: true)`, a grace period, and fault observation. That is exactly what `RunProcess` already does, so the cheapest correct fix is to route both probes through it.

## Also worth adjusting

[...]

PR #3009, opened 2026-08-01
Clarify IlVerifier deadlock test limit

DavidObando:
## Summary

Document that the RunProcess timeout cannot protect the dual-pipe regression test when a deadlock is introduced before the bounded wait.

Follow-up to #2966. Comment-only change; no test behavior changed.

### DIFF
--- a/test/Compiler.Tests/IlVerifierTests.cs
+++ b/test/Compiler.Tests/IlVerifierTests.cs
@@ -177,7 +177,8 @@ public class IlVerifierTests
     {
         // A regression that starts either drain after WaitForExit hangs this test.
         // xUnit's Fact timeout is unavailable because this assembly disables test
-        // parallelization, so treat such a hang as the deadlock regression, not a flake.
+        // parallelization, and RunProcess's timeout cannot cover a deadlock placed
+        // before its wait. Treat such a hang as the regression, not a flake.
         const int OutputLength = 128 * 1024;
         var result = IlVerifier.RunProcess(
             CreateChildProcess(


DavidObando:
## Review — **MERGEABLE**, with one should-fix that this PR is the natural place to make

Reviewed at `5797343ab3261263bc53051f8c2a4b4c6d1fee6d`; computed `git merge-base` with `main` (`a692200af`) = `a1237ded7`; three-dot diff, net state read at the tip.

**This is a comment-only change, so there is no behavioural differential to measure and I did not manufacture one.** Verification is control-flow reading plus history.

### "Comment-only" — confirmed

The entire diff is 2 added / 1 removed lines, all inside a `//` block in `test/Compiler.Tests/IlVerifierTests.cs`. No executable line, no test-behaviour change. The claim holds exactly.

### The new sentence is **true**, and it is worth documenting

> *"RunProcess's timeout cannot cover a deadlock placed before its wait."*

`IlVerifier.RunProcess` is:

```
:180   var stdoutTask = proc.StandardOutput.ReadToEndAsync();
:181   var stderrTask = proc.StandardError.ReadToEndAsync();
:182   if (!proc.WaitForExit(timeoutMilliseconds))        ← the only place the timeout is armed
```

The timeout is an argument to the call at `:182`. Any block at `:180` or `:181` prevents control from reaching it, so the bound is never established. Concretely, the regression that produces a real hang is converting those two `ReadToEndAsync()` calls to sequential `ReadToEnd()` — precisely the pattern still live at `IlVerifier.cs:390` (`TryProbeDotnetToolRun`) and `:427` (`TryProbeOnPath`): the parent blocks on stdout to EOF while the child blocks filling the 128 KB stderr pipe, and neither proceeds. **A timeout on a call you never reach protects nothing.** The sentence states this correctly and precisely.

It is also **not** a limitation that #3007 removes. #3007 targets `:390`/`:427`, which are different methods; `RunProcess` already uses the correct concurrent-async pattern. The limitation being documented is intrinsic to placing a bound on `WaitForExit`, and survives #3007 unchanged.

### Should-fix — the sentence immediately above it is **false**, and now visibly contradicts the new one

> *"A regression that starts either drain after WaitForExit hangs this test."*

Trace it against the current code. Move either `ReadToEndAsync` start to after `:182`: that pipe goes undrained, the child writing 128 KB (well past the ~64 KB buffer) blocks, and the child never exits. Then `WaitForExit(10_000)` returns **false** at 10 s → `Kill(entireProcessTree: true)` → `WaitForExit(KillGraceMilliseconds)` → `Grab` via `task.Wait(KillGraceMilliseconds)`, every step bounded → `XunitException("… timed out after 10000 ms …")`.

**That is a deterministic failure in about ten seconds, not a hang.** Two sibling tests in this same file pin exactly that property — `RunProcess_HungChild_TimesOutWithPartialOutput` (`:155`) and `RunProcess_EscapedDescendantHoldingPipe_TimeoutRemainsBounded` (`:197`). If a drain-after-`WaitForExit` regression could hang, those tests' premise would be broken too.

The history explains it. `git log -S` puts the sentence's introduction at **`87ca66e8` — #2966 itself**, the same commit that replaced the unbounded wait with a bounded one. At `87ca66e85^` the code read:

```
:141   var stdoutTask = proc.StandardOutput.ReadToEndAsync();
:142   var stderrTask = proc.StandardError.ReadToEndAsync();
:143   proc.WaitForExit();          ← unbounded
```

Under *that* code the sentence was true — an undrained pipe would hang forever. #2966 made the wait bounded and carried the sentence across unchanged, so **it has been stale since the moment it was written**, describing pre-#2966 behaviour.

The consequence for this PR: the comment now reads "the regression is a drain started **after** `WaitForExit`" followed immediately by "the timeout cannot cover a deadlock placed **before** its wait" — two opposite placements presented as one explanation. The added clause is right; the reader is still pointed at the wrong regression, and the two sentences now conspicuously disagree.

**Recommended fix — one line, in a comment this PR already edits.** Replace the first sentence so it names the regression that actually hangs, e.g. *"A regression that makes either drain synchronous before the wait hangs this test."* That leaves the block internally consistent and preserves the PR's stated purpose. A drain moved after the wait is worth mentioning separately if desired, but as a bounded ~10 s failure, not a hang.

### Verdict

**MERGEABLE.** The PR introduces no falsehood, adds a correct and genuinely useful sentence, and its "comment-only" claim is exact. It is a net improvement over `main` as it stands.

The should-fix is pre-existing (inherited from #2966), so it is not a regression this PR causes. But #3009 is the dedicated follow-up to #2966 whose entire purpose is clarifying this comment block, it is already editing the adjacent line, and the fix is one sentence — so this is both the cheapest and the most appropriate place to make it. If it is not taken here it should be filed, because under the standing rule a false sentence no CI job can see is exactly the thing that survives indefinitely.

*No build or test host was used or is required for this review.*

PR #3011, opened 2026-08-01
gsi: honor explicit Main and integer exit codes

### LINKED ISSUE #2984: gsi: explicit func Main is never invoked, so Main-only programs silently produce no output
## Problem

`gsi` never invokes an explicit `func Main()`. A program whose entire body lives in `Main` runs, produces **no output, no diagnostic, and exit code 0**.

```gsharp
package P
import System

func Main() {
    Console.WriteLine("hello from Main")
}
```

| path | output |
|---|---|
| `gsc` + `dotnet run` | `hello from Main` |
| `gsi mainonly.gs` | *(nothing)*, rc=0 |

## Root cause

`src/Core/CodeAnalysis/Evaluator.cs:154-156`:

```csharp
public object Evaluate()
{
    return EvaluateFunctionBody(program.Statement);
}
```

`Evaluate()` runs only the top-level statement block. It never consults `program.EntryPoint`, which `Binder.ResolveEntryPoint` (`src/Core/CodeAnalysis/Binding/Binder.cs:6303`) already computed and which the emitter honours — including the class-scoped `static Main` case from #1996.

## Impact

Silent wrong answer, not an error. Any `Main`-style program — the default shape in most languages, and what `gsi --help` advertises as "Run the given G# script and exit" — is a no-op under the interpreter.

## Fix sketch

In `Evaluate()`, when `program.Statement` has no statements and `program.EntryPoint` is non-null, evaluate the entry point's body instead. Mirror the emitter's precedence so the two paths cannot drift.

## Repro samples

`samples/SpanComprehensive.gs`, `samples/RefStructGenericField.gs` — the only two corpus samples that use `func Main()`; both produce empty output under `gsi`.

Found while triaging #2939 (interpreter golden-sample conformance). Excluded there until this is fixed.

### LINKED ISSUE #3010: gsi: integer entry-point results are printed instead of used as exit codes
## Problem

`gsi` treats an integer entry-point result as a display value: it prints the integer and exits 0. `gsc` uses that integer as the process exit code and does not print it.

This affects both top-level statements and explicit `Main` after #2984.

```gsharp
package MainReturn
import System

func Main() int32 {
    Console.WriteLine("main")
    return 3
}
```

Measured with direct Release binaries:

| path | stdout | exit code |
|---|---|---:|
| `gsc` + `dotnet main-return.dll` | `main` | 3 |
| `gsi main-return.gs` | `main` then `3` | 0 |

The equivalent top-level program (`Console.WriteLine("tls"); return 3`) has the same divergence: compiled stdout `tls`, rc 3; interpreted stdout `tls` then `3`, rc 0.

## Likely cause

`src/Repl/Program.cs` prints every non-null `cell.Value` and returns only 0/1 based on `cell.HasError`. Script-mode integer entry-point results need to become the process exit code instead.

Found while fixing #2984. Kept separate because that fix is deliberately limited to entry-point invocation.

DavidObando:
## Compiled rule

Measured with direct Release `gsc` and `gsi` binaries:

- With no top-level statements, the binder-selected explicit `Main` is the entry point. `gsc` runs it.
- With top-level statements and explicit `Main`, the synthesized top-level entry point wins. `gsc` reports warning GS0166 and does not run explicit `Main`.
- Declared return type controls program results: `void` discards the final evaluated value; `int32` and `uint32` become process exit codes; a top-level `return <int>` likewise becomes the synthesized entry point's exit code. Trailing top-level expressions do not become results.

## Fix

- `src/Core/CodeAnalysis/Evaluator.cs:172-217` invokes the binder-selected entry point, seeds only `Main(args []string)`, and normalizes the result from `FunctionSymbol.Type`, the same declared-return signal used by emit.
- `src/Core/CodeAnalysis/Compilation/Compilation.cs:292-362` separates entry-point execution from declared-return normalization. Existing direct `Compilation`/`Evaluator` APIs still expose expression values; file-mode execution applies program exit semantics.
- `src/Repl/Engine/SessionEngine.cs:159-163,220` adds an explicit script-mode switch. Interactive REPL submissions remain declarative; declaring `Main` does not run it. Script sessions run it only on the declaring submission, not again on a later `ContinueWith` submission.
- `src/Repl/Program.cs:57-79` enables script mode, returns `int32`/`uint32` results as exit codes, returns 1 on diagnostics, and never prints discarded expression values.

## Three-driver entry-point matrix

Each cell is covered by `Issue3010EntryPointDriverMatrixTests`: `{top-level, explicit Main} × {gsc evaluation, gsc /out emission, gsi}`. Compiler evaluation owns its `Success.` status line and always returns 0 after successful evaluation; emitted and interpreted programs expose the language process exit code.

| Shape | bare `gsc` evaluation | `gsc /out:` + `dotnet` | `gsi` |
|---|---|---|---|
| `Main() int32` returns 3 | rc 0, `main-int\nSuccess.\n` | rc 3, `main-int\n` | rc 3, `main-int\n` |
| `Main() int32` returns 0 | rc 0, `main-zero\nSuccess.\n` | rc 0, `main-zero\n` | rc 0, `main-zero\n` |
| void `Main()` + bare int | rc 0, `void-bare-int\nSuccess.\n` | rc 0, `void-bare-int\n` | rc 0, `void-bare-int\n` |
| void `Main()` + ignored int call | rc 0, `void-call\nSuccess.\n` | rc 0, `void-call\n` | rc 0, `void-call\n` |
| void `Main()` + bare string | rc 0, `void-string\nSuccess.\n` | rc 0, `void-string\n` | rc 0, `void-string\n` |
| void `Main()` + explicit print | rc 0, `void-print\nSuccess.\n` | rc 0, `void-print\n` | rc 0, `void-print\n` |
| top-level trailing int expression | rc 0, `top-bare\nSuccess.\n` | rc 0, `top-bare\n` | rc 0, `top-bare\n` |
| top-level `return 3` | rc 0, `top-return\nSuccess.\n` | rc 3, `top-return\n` | rc 3, `top-return\n` |

Trailing expressions are discarded because the binder declares those entry points `void`; explicit/top-level `return <int>` declares an integer entry point and therefore supplies the emitted/interpreted process exit code.

## Bidirectional evidence

Same base `6ec381c5ddc15b573dc6a8701edcb17ae4413a99`, head `94b65165edea279b8e862f7fcdc7349461412ac3`:

| Probe | Base `gsi` | Head `gsi` |
|---|---|---|
| `Main()` prints `hello-2984` | rc 0, `∅` | rc 0, `hello-2984\n` |
| `Main() int32` prints and returns 42 | rc 0, `∅` | rc 42, `main-int\n` |
| void `Main()` ends in `40+2` | rc 0, `∅` | rc 0, `void-trailing\n` |

## Product mutation

Twenty-two condition/value inversions covered every production behavior in the eight production diff hunks. Every mutant built successfully, changed the owning `GSharp.Core.dll` or `gsi.dll`, passed the 34-file zero-byte and six-copy propagation gates, and was killed.

| Production area | Mutants | Killing tests |
|---|---:|---|
| `Compilation.cs` defaults and evaluator forwarding | 4/4 | `PackageScopeMainRuns`; `ExplicitVoidMainDiscardsBareIntegerExpression` |
| `Evaluator.cs` state, dispatch, args, declared return validation/result | 11/11 | `EvaluatorDefaultRunsEntryPoint`; `CompilationEvaluationPreservesTopLevelExpressionValue`; `MainArgsReceivesEmptyArray`; explicit void/int/uint result tests |
| `SessionEngine.cs` mode default and forwarding | 2/2 | `DeclarationOnlySubmissionsDoNotRunMain` |
| `Program.cs` script enablement, error/int/uint/zero exits | 5/5 | explicit void/int/uint result tests |

## Validation

Current main/base is `6ec381c5ddc15b573dc6a8701edcb17ae4413a99`; head is `94b65165edea279b8e862f7fcdc7349461412ac3`.

- Release solution build, serialized with CI mode: 0 warnings, 0 errors.
- Targeted #2984/#3010 tests: 51/51, including 24 three-driver matrix cells.
- Same-base Interpreter population: 637 passed on base → 688 passed on head, 51 new executed cases.
- Final CI TRX suites: Interpreter 688/688; Core 7088 passed + 1 skipped; Compiler 4146/4146.
- `cs2gs`: 1874/1874.
- Artifact gate after clean rebuild: 34 `GSharp.Core.dll` files examined, 0 zero-byte; six consumer copies had one md5 and identical mtimes.
- `git merge-tree --write-tree origin/main HEAD`: clean; merged tree `41e72e54e2fbd9d463e062fd502e5495f42c258f` equals `HEAD^{tree}`.
- CI at pushed head: 29 successful checks, 2 expected publish checks skipped; PR reports `MERGEABLE` / `CLEAN`.

Fixes #2984
Fixes #3010

### DIFF
--- a/src/Core/CodeAnalysis/Compilation/Compilation.cs
+++ b/src/Core/CodeAnalysis/Compilation/Compilation.cs
@@ -290,7 +290,7 @@ public class Compilation
     /// <param name="variables">The symbol table with actual values.</param>
     /// <returns>An evaluation result.</returns>
     public EvaluationResult Evaluate(Dictionary<VariableSymbol, object> variables) =>
-        Evaluate(variables, evaluateEntryPoint: true);
+        Evaluate(variables, evaluateEntryPoint: true, useEntryPointReturnType: false);
 
     /// <summary>
     /// Evaluates the current compilation provided a symbol table with actual values.
@@ -298,7 +298,20 @@ public class Compilation
     /// <param name="variables">The symbol table with actual values.</param>
     /// <param name="evaluateEntryPoint">Whether to invoke an explicit program entry point.</param>
     /// <returns>An evaluation result.</returns>
-    public EvaluationResult Evaluate(Dictionary<VariableSymbol, object> variables, bool evaluateEntryPoint)
+    public EvaluationResult Evaluate(Dictionary<VariableSymbol, object> variables, bool evaluateEntryPoint) =>
+        Evaluate(variables, evaluateEntryPoint, useEntryPointReturnType: evaluateEntryPoint);
+
+    /// <summary>
+    /// Evaluates the current compilation provided a symbol table with actual values.
+    /// </summary>
+    /// <param name="variables">The symbol table with actual values.</param>
+    /// <param name="evaluateEntryPoint">Whether to invoke an explicit program entry point.</param>
+    /// <param name="useEntryPointReturnType">Whether the declared entry-point return type controls the result.</param>
+    /// <returns>An evaluation result.</returns>
+    public EvaluationResult Evaluate(
+        Dictionary<VariableSymbol, object> variables,
+        bool evaluateEntryPoint,
+        bool useEntryPointReturnType)
     {
         var parseDiagnostics = SyntaxTrees.SelectMany(st => st.Diagnostics);
         var diagnostics = parseDiagnostics.Concat(GlobalScope.Diagnostics).ToImmutableArray();
@@ -342,7 +355,11 @@ public class Compilation
             program,
             (References ?? Symbols.ReferenceResolver.Default()).MapClrTypeToReferences);
 
-        var evaluator = new Evaluator(program, variables, evaluateEntryPoint);
+        var evaluator = new Evaluator(
+            program,
+            variables,
+            evaluateEntryPoint,
+            useEntryPointReturnType);
         try
         {
             var value = evaluator.Evaluate();
--- a/src/Core/CodeAnalysis/Compilation/Compilation.cs
+++ b/src/Core/CodeAnalysis/Compilation/Compilation.cs
@@ -289,7 +289,16 @@ public class Compilation
     /// </summary>
     /// <param name="variables">The symbol table with actual values.</param>
     /// <returns>An evaluation result.</returns>
-    public EvaluationResult Evaluate(Dictionary<VariableSymbol, object> variables)
+    public EvaluationResult Evaluate(Dictionary<VariableSymbol, object> variables) =>
+        Evaluate(variables, evaluateEntryPoint: true);
+
+    /// <summary>
+    /// Evaluates the current compilation provided a symbol table with actual values.
+    /// </summary>
+    /// <param name="variables">The symbol table with actual values.</param>
+    /// <param name="evaluateEntryPoint">Whether to invoke an explicit program entry point.</param>
+    /// <returns>An evaluation result.</returns>
+    public EvaluationResult Evaluate(Dictionary<VariableSymbol, object> variables, bool evaluateEntryPoint)
     {
         var parseDiagnostics = SyntaxTrees.SelectMany(st => st.Diagnostics);
         var diagnostics = parseDiagnostics.Concat(GlobalScope.Diagnostics).ToImmutableArray();
@@ -333,7 +342,7 @@ public class Compilation
             program,
             (References ?? Symbols.ReferenceResolver.Default()).MapClrTypeToReferences);
 
-        var evaluator = new Evaluator(program, variables);
+        var evaluator = new Evaluator(program, variables, evaluateEntryPoint);
         try
         {
             var value = evaluator.Evaluate();
--- a/src/Core/CodeAnalysis/Evaluator.cs
+++ b/src/Core/CodeAnalysis/Evaluator.cs
@@ -154,6 +154,23 @@ public sealed partial class Evaluator
     /// <returns>The evaluation result.</returns>
     public object Evaluate()
     {
+        // The binder has already resolved top-level-statement versus explicit-Main precedence.
+        // A null declaration identifies the synthesized top-level entry point.
+        if (program.EntryPoint?.Declaration != null
+            && program.Functions.TryGetValue(program.EntryPoint, out var entryPointBody))
+        {
+            var locals = new ConcurrentDictionary<VariableSymbol, object>();
+            if (program.EntryPoint.Parameters.Length > 0)
+            {
+                locals[program.EntryPoint.Parameters[0]] = Array.Empty<string>();
+            }
+
+            using (PushFrame(locals))
+            {
+                return EvaluateFunctionBody(entryPointBody);
+            }
+        }
+
         return EvaluateFunctionBody(program.Statement);
     }
 
--- a/src/Core/CodeAnalysis/Evaluator.cs
+++ b/src/Core/CodeAnalysis/Evaluator.cs
@@ -38,6 +38,7 @@ public sealed partial class Evaluator
     private readonly BoundProgram program;
     private readonly Dictionary<VariableSymbol, object> globals;
     private readonly bool evaluateEntryPoint;
+    private readonly bool useEntryPointReturnType;
 
     // Issue #1651: everything below this point used to be an ordinary
     // instance field. That is correct only because the interpreter was
@@ -101,21 +102,24 @@ public sealed partial class Evaluator
     /// <param name="program">The program.</param>
     /// <param name="variables">The variables.</param>
     public Evaluator(BoundProgram program, Dictionary<VariableSymbol, object> variables)
-        : this(program, variables, evaluateEntryPoint: true)
+        : this(
+            program,
+            variables,
+            evaluateEntryPoint: true,
+            useEntryPointReturnType: false)
     {
     }
 
-    /// <summary>
-    /// Initializes a new instance of the <see cref="Evaluator"/> class.
-    /// </summary>
-    /// <param name="program">The program.</param>
-    /// <param name="variables">The variables.</param>
-    /// <param name="evaluateEntryPoint">Whether to invoke an explicit program entry point.</param>
-    public Evaluator(BoundProgram program, Dictionary<VariableSymbol, object> variables, bool evaluateEntryPoint)
+    internal Evaluator(
+        BoundProgram program,
+        Dictionary<VariableSymbol, object> variables,
+        bool evaluateEntryPoint,
+        bool useEntryPointReturnType)
     {
         this.program = program;
         globals = variables;
         this.evaluateEntryPoint = evaluateEntryPoint;
+        this.useEntryPointReturnType = useEntryPointReturnType;
         Locals.Push(new ConcurrentDictionary<VariableSymbol, object>());
     }
 
@@ -197,6 +201,11 @@ public sealed partial class Evaluator
 
     private object EvaluateEntryPointBody(FunctionSymbol entryPoint, BoundBlockStatement body)
     {
+        if (!useEntryPointReturnType)
+        {
+            return EvaluateFunctionBody(body);
+        }
+
         if (entryPoint.Type != TypeSymbol.Void
             && entryPoint.Type != TypeSymbol.Int32
             && entryPoint.Type != TypeSymbol.UInt32)
--- a/src/Core/CodeAnalysis/Evaluator.cs
+++ b/src/Core/CodeAnalysis/Evaluator.cs
@@ -37,6 +37,7 @@ public sealed partial class Evaluator
 {
     private readonly BoundProgram program;
     private readonly Dictionary<VariableSymbol, object> globals;
+    private readonly bool evaluateEntryPoint;
 
     // Issue #1651: everything below this point used to be an ordinary
     // instance field. That is correct only because the interpreter was
@@ -100,9 +101,21 @@ public sealed partial class Evaluator
     /// <param name="program">The program.</param>
     /// <param name="variables">The variables.</param>
     public Evaluator(BoundProgram program, Dictionary<VariableSymbol, object> variables)
+        : this(program, variables, evaluateEntryPoint: true)
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="Evaluator"/> class.
+    /// </summary>
+    /// <param name="program">The program.</param>
+    /// <param name="variables">The variables.</param>
+    /// <param name="evaluateEntryPoint">Whether to invoke an explicit program entry point.</param>
+    public Evaluator(BoundProgram program, Dictionary<VariableSymbol, object> variables, bool evaluateEntryPoint)
     {
         this.program = program;
         globals = variables;
+        this.evaluateEntryPoint = evaluateEntryPoint;
         Locals.Push(new ConcurrentDictionary<VariableSymbol, object>());
     }
 
@@ -156,11 +169,14 @@ public sealed partial class Evaluator
     {
         // The binder has already resolved top-level-statement versus explicit-Main precedence.
         // A null declaration identifies the synthesized top-level entry point.
-        if (program.EntryPoint?.Declaration != null
-            && program.Functions.TryGetValue(program.EntryPoint, out var entryPointBody))
+        if (evaluateEntryPoint && program.EntryPoint?.Declaration != null)
         {
+            // Every resolved explicit entry point must have a bound body; fail loudly if that
+            // invariant is broken rather than silently evaluating an empty top-level block.
+            var entryPointBody = program.Functions[program.EntryPoint];
             var locals = new ConcurrentDictionary<VariableSymbol, object>();
-            if (program.EntryPoint.Parameters.Length > 0)
+            if (program.EntryPoint.Parameters.Length == 1
+                && program.EntryPoint.Parameters[0].Type == SliceTypeSymbol.Get(TypeSymbol.String))
             {
                 locals[program.EntryPoint.Parameters[0]] = Array.Empty<string>();
             }
--- a/src/Core/CodeAnalysis/Evaluator.cs
+++ b/src/Core/CodeAnalysis/Evaluator.cs
@@ -169,25 +169,43 @@ public sealed partial class Evaluator
     {
         // The binder has already resolved top-level-statement versus explicit-Main precedence.
         // A null declaration identifies the synthesized top-level entry point.
-        if (evaluateEntryPoint && program.EntryPoint?.Declaration != null)
+        if (evaluateEntryPoint && program.EntryPoint is { } entryPoint)
         {
-            // Every resolved explicit entry point must have a bound body; fail loudly if that
-            // invariant is broken rather than silently evaluating an empty top-level block.
-            var entryPointBody = program.Functions[program.EntryPoint];
-            var locals = new ConcurrentDictionary<VariableSymbol, object>();
-            if (program.EntryPoint.Parameters.Length == 1
-                && program.EntryPoint.Parameters[0].Type == SliceTypeSymbol.Get(TypeSymbol.String))
+            if (entryPoint.Declaration != null)
             {
-                locals[program.EntryPoint.Parameters[0]] = Array.Empty<string>();
+                // Every resolved explicit entry point must have a bound body; fail loudly if that
+                // invariant is broken rather than silently evaluating an empty top-level block.
+                var entryPointBody = program.Functions[entryPoint];
+                var locals = new ConcurrentDictionary<VariableSymbol, object>();
+                if (entryPoint.Parameters.Length == 1
+                    && entryPoint.Parameters[0].Type == SliceTypeSymbol.Get(TypeSymbol.String))
+                {
+                    locals[entryPoint.Parameters[0]] = Array.Empty<string>();
+                }
+
+                using (PushFrame(locals))
+                {
+                    return EvaluateEntryPointBody(entryPoint, entryPointBody);
+                }
             }
 
-            using (PushFrame(locals))
-            {
-                return EvaluateFunctionBody(entryPointBody);
-            }
+            return EvaluateEntryPointBody(entryPoint, program.Statement);
         }
 
         return EvaluateFunctionBody(program.Statement);
+    }
+
+    private object EvaluateEntryPointBody(FunctionSymbol entryPoint, BoundBlockStatement body)
+    {
+        if (entryPoint.Type != TypeSymbol.Void
+            && entryPoint.Type != TypeSymbol.Int32
+            && entryPoint.Type != TypeSymbol.UInt32)
+        {
+            throw new InvalidOperationException("Entry point must have a return type of void, int32, or uint32.");
+        }
+
+        var value = EvaluateFunctionBody(body);
+        return entryPoint.Type == TypeSymbol.Void ? null : value;
     }
 
     // Issue #1651: goroutines run with their own locals/control-transfer
--- a/src/Repl/Engine/SessionEngine.cs
+++ b/src/Repl/Engine/SessionEngine.cs
@@ -157,6 +157,12 @@ public sealed class SessionEngine
     public bool CaptureConsole { get; set; }
 
     /// <summary>
+    /// When <see langword="true"/>, invokes an explicit program entry point. Script mode enables
+    /// this; interactive submissions leave it disabled so declaring <c>Main</c> has no side effect.
+    /// </summary>
+    public bool RunEntryPoint { get; set; }
+
+    /// <summary>
     /// Optional provider that supplies a line of standard input when evaluated code reads from
     /// <see cref="Console.In"/> (e.g. <c>Console.ReadLine()</c>). Returning <see langword="null"/>
     /// signals end-of-input. When unset, <see cref="Console.In"/> is left untouched.
@@ -211,7 +217,7 @@ public sealed class SessionEngine
             {
                 var tree = SyntaxTree.Parse(SourceText.From(text, fileName));
                 var compilation = previous == null ? new Compilation(tree) : previous.ContinueWith(tree);
-                var result = compilation.Evaluate(variables);
+                var result = compilation.Evaluate(variables, evaluateEntryPoint: RunEntryPoint);
 
                 var hasError = result.Diagnostics.Any(d => d.IsError);
 
--- a/src/Repl/Program.cs
+++ b/src/Repl/Program.cs
@@ -54,7 +54,7 @@ public static class Program
             return 1;
         }
 
-        var engine = new SessionEngine();
+        var engine = new SessionEngine { RunEntryPoint = true };
         var cell = engine.Evaluate(File.ReadAllText(arg), arg);
         if (cell.Diagnostics.Length > 0)
         {
--- a/src/Repl/Program.cs
+++ b/src/Repl/Program.cs
@@ -61,11 +61,21 @@ public static class Program
             Console.Error.WriteDiagnostics(cell.Diagnostics);
         }
 
-        if (!cell.HasError && cell.Value is not null)
+        if (cell.HasError)
+        {
+            return 1;
+        }
+
+        if (cell.Value is int exitCode)
+        {
+            return exitCode;
+        }
+
+        if (cell.Value is not null)
         {
             Console.WriteLine(cell.Value);
         }
 
-        return cell.HasError ? 1 : 0;
+        return 0;
     }
 }
--- a/src/Repl/Program.cs
+++ b/src/Repl/Program.cs
@@ -71,9 +71,9 @@ public static class Program
             return exitCode;
         }
 
-        if (cell.Value is not null)
+        if (cell.Value is uint unsignedExitCode)
         {
-            Console.WriteLine(cell.Value);
+            return unchecked((int)unsignedExitCode);
         }
 
         return 0;
--- a/test/Interpreter.Tests/Issue2984MainEntryPointInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue2984MainEntryPointInterpreterTests.cs
@@ -37,6 +37,37 @@ public class Issue2984MainEntryPointInterpreterTests
     }
 
     [Fact]
+    public void EvaluatorDefaultRunsEntryPoint()
+    {
+        const string Source = """
+            import System
+
+            func Main() {
+                Console.WriteLine("evaluator-main")
+            }
+            """;
+
+        using var output = new StringWriter();
+        var previousOutput = Console.Out;
+        Console.SetOut(output);
+        try
+        {
+            var compilation = new Compilation(SyntaxTree.Parse(Source));
+            var evaluator = new Evaluator(
+                compilation.BoundProgram,
+                new Dictionary<VariableSymbol, object>());
+
+            evaluator.Evaluate();
+
+            Assert.Equal("evaluator-main\n", output.ToString().Replace("\r\n", "\n"));
+        }
+        finally
+        {
+            Console.SetOut(previousOutput);
+        }
+    }
+
+    [Fact]
     public void ClassScopedSharedMainRuns()
     {
         const string Source = """
--- a/test/Interpreter.Tests/Issue2984MainEntryPointInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue2984MainEntryPointInterpreterTests.cs
@@ -1,0 +1,128 @@
+// <copyright file="Issue2984MainEntryPointInterpreterTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using GSharp.Core.CodeAnalysis;
+using GSharp.Core.CodeAnalysis.Compilation;
+using GSharp.Core.CodeAnalysis.Symbols;
+using GSharp.Core.CodeAnalysis.Syntax;
+using Xunit;
+
+namespace GSharp.Interpreter.Tests;
+
+/// <summary>
+/// Issue #2984: the interpreter invokes the entry point selected by the binder.
+/// </summary>
+[Collection("ConsoleIo")]
+public class Issue2984MainEntryPointInterpreterTests
+{
+    [Fact]
+    public void PackageScopeMainRuns()
+    {
+        const string Source = """
+            import System
+
+            func Main() {
+                Console.WriteLine("package-main")
+            }
+            """;
+
+        var (result, output) = Evaluate(Source);
+
+        Assert.DoesNotContain(result.Diagnostics, diagnostic => diagnostic.IsError);
+        Assert.Equal("package-main\n", output);
+    }
+
+    [Fact]
+    public void ClassScopedSharedMainRuns()
+    {
+        const string Source = """
+            import System
+
+            class Launcher {
+                shared {
+                    func Main() {
+                        Console.WriteLine("class-main")
+                    }
+                }
+            }
+            """;
+
+        var (result, output) = Evaluate(Source);
+
+        Assert.DoesNotContain(result.Diagnostics, diagnostic => diagnostic.IsError);
+        Assert.Equal("class-main\n", output);
+    }
+
+    [Fact]
+    public void TopLevelStatementsStillRun()
+    {
+        const string Source = """
+            import System
+
+            Console.WriteLine("top-level")
+            """;
+
+        var (result, output) = Evaluate(Source);
+
+        Assert.Empty(result.Diagnostics);
+        Assert.Equal("top-level\n", output);
+    }
+
+    [Fact]
+    public void TopLevelStatementsWinOverExplicitMain()
+    {
+        const string Source = """
+            import System
+
+            Console.WriteLine("top-level")
+
+            func Main() {
+                Console.WriteLine("explicit-main")
+            }
+            """;
+
+        var (result, output) = Evaluate(Source);
+
+        Assert.DoesNotContain(result.Diagnostics, diagnostic => diagnostic.IsError);
+        Assert.Contains(result.Diagnostics, diagnostic => diagnostic.Id == "GS0166");
+        Assert.Equal("top-level\n", output);
+    }
+
+    [Fact]
+    public void MainArgsReceivesEmptyArray()
+    {
+        const string Source = """
+            import System
+
+            func Main(args []string) {
+                Console.WriteLine(args.Length)
+            }
+            """;
+
+        var (result, output) = Evaluate(Source);
+
+        Assert.DoesNotContain(result.Diagnostics, diagnostic => diagnostic.IsError);
+        Assert.Equal("0\n", output);
+    }
+
+    private static (EvaluationResult Result, string Output) Evaluate(string source)
+    {
+        using var output = new StringWriter();
+        var previousOutput = Console.Out;
+        Console.SetOut(output);
+        try
+        {
+            var result = new Compilation(SyntaxTree.Parse(source))
+                .Evaluate(new Dictionary<VariableSymbol, object>());
+            return (result, output.ToString().Replace("\r\n", "\n"));
+        }
+        finally
+        {
+            Console.SetOut(previousOutput);
+        }
+    }
+}
--- a/test/Interpreter.Tests/Issue2984MainEntryPointInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue2984MainEntryPointInterpreterTests.cs
@@ -109,6 +109,28 @@ public class Issue2984MainEntryPointInterpreterTests
         Assert.Equal("0\n", output);
     }
 
+    [Fact]
+    public void NonArgsMainParameterIsNotSeededWithStringArray()
+    {
+        const string Source = """
+            import System
+
+            func Main(x int32) {
+                Console.WriteLine(x)
+            }
+            """;
+
+        var (result, output) = Evaluate(Source);
+
+        Assert.Contains(
+            result.Diagnostics,
+            diagnostic => diagnostic.IsError && diagnostic.Message.Contains("x int32"));
+        Assert.DoesNotContain(
+            result.Diagnostics,
+            diagnostic => diagnostic.Message.Contains("System.String[]"));
+        Assert.Equal(string.Empty, output);
+    }
+
     private static (EvaluationResult Result, string Output) Evaluate(string source)
     {
         using var output = new StringWriter();
--- a/test/Interpreter.Tests/Issue2984MainEntryPointInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue2984MainEntryPointInterpreterTests.cs
@@ -68,6 +68,27 @@ public class Issue2984MainEntryPointInterpreterTests
     }
 
     [Fact]
+    public void CompilationEvaluationPreservesTopLevelExpressionValue()
+    {
+        var result = new Compilation(SyntaxTree.Parse("40 + 2"))
+            .Evaluate(new Dictionary<VariableSymbol, object>());
+
+        Assert.Empty(result.Diagnostics);
+        Assert.Equal(42, result.Value);
+    }
+
+    [Fact]
+    public void EvaluatorDefaultPreservesTopLevelExpressionValue()
+    {
+        var compilation = new Compilation(SyntaxTree.Parse("40 + 2"));
+        var evaluator = new Evaluator(
+            compilation.BoundProgram,
+            new Dictionary<VariableSymbol, object>());
+
+        Assert.Equal(42, evaluator.Evaluate());
+    }
+
+    [Fact]
     public void ClassScopedSharedMainRuns()
     {
         const string Source = """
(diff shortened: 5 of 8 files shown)


DavidObando:
## Review: MERGEABLE — with two SHOULD-FIXes

Reviewed at head `642600359`, `main` = `f70626452`, **computed merge-base `a6b1561f2`** (three-dot; branch is 1 ahead / 9 behind, merges clean). Net diff at the tip: **2 files, +145** — `src/Core/CodeAnalysis/Evaluator.cs` +17, `test/Interpreter.Tests/Issue2984MainEntryPointInterpreterTests.cs` +128.

Read-only review — no build or test host was used (machine load at threshold).

### `Fixes #2984` holds, and it is the root cause rather than a named symptom

`Evaluator.Evaluate()` has **exactly one call site** — `Compilation.cs:324` (denominator: 1 of 1 `.Evaluate()` hit in `src/`). The guard is at the real chokepoint, so both shapes the issue names are covered by one edit rather than one guard per reported path.

Both broken shapes are pinned, and every assertion is on an **output value**, not on exit status:

| test | shape | asserts |
|---|---|---|
| `PackageScopeMainRuns` | package-scope `Main`, no TLS | `"main-only\n"` |
| `ClassScopedSharedMainRuns` | class-scoped `shared Main`, no TLS | `"class-main\n"` |
| `TopLevelStatementsStillRun` | TLS only | TLS output |
| `TopLevelStatementsWinOverExplicitMain` | both | GS0166 **and** TLS output |
| `MainArgsReceivesEmptyArray` | `Main(args []string)` | `"0\n"` from `args.Length` |

The harness (`:112-127`) is `new Compilation(SyntaxTree.Parse(source)).Evaluate(...)` — the identical path `gsi` takes via `Program.cs:57`, not a bespoke one.

### No `gsc`/`gsi` entry-point divergence is introduced

Both drivers consume the same `Binder.ResolveEntryPoint` (`Binder.cs:6303-6372`) decision, so precedence and the silent first-found choice among multiple `Main`s are shared **by construction** — this PR cannot desynchronise them. The `Declaration != null` predicate is an exact discriminator, not a duplicate of the precedence rule: `Binder.cs:1789-1794` registers the synthesized top-level entry point with a null `Declaration`.

Unchanged paths verified by reading, not assumed:
- **TLS program** → synthesized entry point has `Declaration == null` → falls through to the original `return EvaluateFunctionBody(program.Statement)`.
- **No `Main` and no TLS** → `program.EntryPoint` is null → `?.` short-circuits → same original line.

### Two fleet hazards checked and cleared

- **No `DiagnosticDescriptor` is added**, so the post-#2981 trap (a descriptor without a `docs/diagnostics.md` row is green on the branch and red only after landing) does not apply here. Green CI is trustworthy on this one.
- **No new diagnostic ID is invented**, so there is no `GS9999`-for-a-designed-refusal misuse and nothing to allocate.
- **`#3010` verified**: OPEN, and it accurately tracks the integer entry-point exit-code divergence with measured values on both paths. It correctly notes the divergence pre-dates this PR and affects TLS too, so deferring it is scoping, not scope-dodging.

---

## SHOULD-FIX 1 — the REPL behaviour change is real, undeclared, and untested

`src/Repl/Engine/SessionEngine.cs:209-210` × `src/Core/CodeAnalysis/Evaluator.cs:153`

The interactive REPL reaches the modified code:

```
:209  var compilation = previous == null ? new Compilation(tree) : previous.ContinueWith(tree);
:210  var result = compilation.Evaluate(variables);
```

There is **no script-mode flag anywhere** to separate the two consumers — `IsScript|ScriptMode` has **0 hits in `src/`** (pattern proven live: `Compilation` matches 3× in that same file, so this is a run check, not an empty grep).

Consequence: a REPL submission that declares nothing but `func Main() { … }` has no top-level statements, so `EntryPoint` resolves to it with `Declaration != null`, and it now **executes on submission**. Before this PR it was merely declared.

Issue #2984 flagged exactly this as its third hazard and asked for an explicit decision. The PR body does not mention the REPL, and coverage is **0 of 5** — all five tests use `new Compilation(tree)`; none uses the REPL's `ContinueWith` chain.

The change may well be defensible ("a program whose entry point is `Main` runs `Main`"), but it should be *stated* and *pinned* rather than discovered later as a REPL regression.

One question I could not settle statically, and did not probe because it needs a HEAD build plus an interactive session: whether a **subsequent** declaration-only submission re-binds the accumulated tree and **re-executes** `Main`. If it does, the effect is repeated execution rather than a single extra run. Worth one `ContinueWith`-based test either way, since that test would answer it and pin it at the same time.

## SHOULD-FIX 2 — the PR body describes narrower behaviour than the code implements

`src/Core/CodeAnalysis/Evaluator.cs:159-162`

The body says the empty `string[]` is seeded "for legal `Main(args []string)` programs". The code seeds it whenever there is any parameter at all:

```csharp
if (program.EntryPoint.Parameters.Length > 0)
{
    locals[program.EntryPoint.Parameters[0]] = Array.Empty<string>();
}
```

`ResolveEntryPoint` performs **no signature validation** — `functions.FirstOrDefault(f => f.Name == "Main")` accepts any signature, and no diagnostic in the tree validates `Main`'s shape (GS0166, GS0287 and GS0329 all check other things). So:

- `func Main(x int32)` is selected as the entry point and receives a `string[]` in an `int32` slot;
- `func Main(a []string, b []string)` leaves `Parameters[1]` unbound.

These are pathological programs and the underlying hole is the binder's, not this PR's. But the PR introduces the unconditional seeding, and the body asserts the narrower contract. Either fix closes the gap and both are one line:

1. narrow the condition to a single `[]string` parameter, so the code matches the sentence; or
2. correct the sentence to say what the code does.

The empty-array *shape* itself is right: slices are plain CLR arrays in the evaluator (`SliceTypeSymbol` has 0 hits in `Evaluator*.cs`; arrays are built via `Array.CreateInstance`), and `dotnet app.dll` with no arguments does yield `argc=0`.

## NIT — the fallback silently reproduces the original bug

`src/Core/CodeAnalysis/Evaluator.cs:154-155`

If `Declaration != null` but `program.Functions.TryGetValue` misses, control falls through to `EvaluateFunctionBody(program.Statement)` — a silent no-op returning 0, which is the exact #2984 signature. It appears unreachable today (`ClassScopedSharedMainRuns` passing proves the class-scoped path populates `Functions`), so this is a note rather than a defect.

If a refusal is ever added there, it must not be `GS9999` — that is the internal-error bucket, and a missing body for a resolved entry point would be a deliberate boundary condition.

## Sentences checked and found true

Per the standing rule that no CI job can see a false sentence, I checked the new prose against the control flow rather than against the PR body:

- `Evaluator.cs:150-151` — *"The binder has already resolved top-level-statement versus explicit-Main precedence. A null declaration identifies the synthesized top-level entry point."* **True**, and confirmed at `Binder.cs:1789-1794`.
- Body: *"Package-scope `Main` takes precedence over a class-scoped `shared Main`."* **True** — `ResolveEntryPoint` consults `functions` first and only falls back to static methods on classes.
- Body: *"The existing top-level path remains unchanged."* **True** by the `Declaration == null` short-circuit above.
- The suite arithmetic closes against a named population: Interpreter **491 → 496** is exactly the 5 added tests; Core 7028+1 skip, Compiler 3906 and LSP 452 are unchanged and match the `a6b1561f2` baseline, which is the correct base for this branch.

The only inaccurate sentence is the `Main(args []string)` one covered in SHOULD-FIX 2.

---

**Verdict: MERGEABLE.** The fix is minimal, correctly placed at the single chokepoint, and does not disturb either unchanged path. Neither SHOULD-FIX affects the correctness of what #2984 asked for; SHOULD-FIX 1 is a real behaviour change on a second surface that should be declared and pinned before it lands, and SHOULD-FIX 2 is a one-line reconciliation between the body and the code.

DavidObando:
## Review: NOT MERGEABLE — `Fixes #3010` is declared, but the code that would fix it is not pushed

Measured tree: scratch clone at `main` = `fd34166a8`, merged with `refs/pull/3011/head` = **`d078d5385`** (three-dot base `47a40035`). Merge is clean; I measured the merge result.

`Fixes #2984` is sound and well pinned. The blocker is entirely about the second declared closure.

---

## BLOCKER — the #3010 half exists only on an unpushed commit

The PR body's own evidence tables are headed "head `8c54d5143`". That commit is **not** what would merge:

```
git merge-base --is-ancestor d078d5385 8c54d5143   -> YES
```

`8c54d5143` is **ahead** of the pushed head by exactly the #3010 work:

```
src/Repl/Program.cs                                  | 14 +-
test/.../Issue3010EntryPointExitCodeTests.cs         | 99 +++++++++
```

At the pushed head, `git diff main..pr3011 -- src/Repl/` is **empty** — `Program.cs` is byte-identical to `main`. The PR body says "Make script-mode `gsi` return an interpreted `int` result as its process exit code instead of printing it", and #3010 itself names `src/Repl/Program.cs` as the cause. That file is untouched by what would merge.

Measured, not inferred. Exit codes captured immediately from the program, never through a pipe:

```
gsc + dotnet mr.dll   stdout=[main]      rc=42     <- correct
gsi  (main+#3011)     stdout=[main, 42]  rc=0      <- #3010, unchanged
```

on

```gsharp
func Main() int32 {
    Console.WriteLine("main")
    return 42
}
```

That is issue #3010's reported behaviour reproduced exactly, at this PR's head, after the merge. **Merging as-is closes #3010 while leaving it fully live.**

Fix: push `8c54d5143` (or drop `Fixes #3010` to `Refs #3010`). I would not resolve this by editing the body — the work exists and is one `git push` away.

## What *does* work — `Fixes #2984` is sound

```
gsi (main+#3011)  mainvoid.gs  stdout=[hello-2984]  rc=0
```

Before the fix this program produced no output at all. Mutation, single production hunk reverted in isolation:

```
clean    md5 GSharp.Core.dll  ee6f84a880348e465d7281c526fc86c8   5 passed
mutant   build rc=0           7f052deb9adadbb2bdde8f9df63a0b55   3 FAILED, 2 passed
                                                                  -> PackageScopeMainRuns
                                                                  -> ClassScopedSharedMainRuns
                                                                  -> MainArgsReceivesEmptyArray
restored md5                  ee6f84a880348e465d7281c526fc86c8   5 passed
```

Build exit verified `0` before trusting either hash; md5 round-trip closes exactly. The hunk is load-bearing and killed by three named tests. Placement at `Evaluator.cs:156` is the right single chokepoint, and the `Declaration != null` discriminator correctly defers precedence to the binder rather than duplicating it.

---

## SHOULD-FIX 1 (carried forward, still open) — the REPL behaviour change is undeclared and unpinned, but I can now bound it

Coverage is still **0**: `grep -c ContinueWith` on `Issue2984MainEntryPointInterpreterTests.cs` = `0`, across all five tests. Pattern proven live — `new Compilation` matches in the same file.

The previous review left one question unsettled: whether a *subsequent* declaration-only submission re-binds the accumulated tree and **re-executes** `Main`. `gsi` refuses a piped session ("gsi requires an interactive terminal"), so I settled it with a throwaway `ContinueWith` test against the HEAD build instead, then removed it:

```
submission 1: func Main() { Console.WriteLine("M-RAN") }
submission 2: let zz = 1        (ContinueWith)
submission 3: let yy = 2        (ContinueWith)

captured stdout = "M-RAN\n"
```

**Exactly one occurrence.** `Main` runs once, on the submission that declares it, and is *not* re-executed by later submissions.

That materially downgrades this item: the effect is a single extra run at declaration time, not repeated execution. It is still a real, undeclared behaviour change on a second surface with zero coverage — but it is bounded, and the test that pins it is now written for you above.

## SHOULD-FIX 2 (carried forward) — resolved in the wrong direction

The earlier finding was that the body described *narrower* behaviour than the code implemented. The body has since been expanded — past the code. It now describes `Program.cs` behaviour, presents an exit-code evidence table, and declares `Fixes #3010`, none of which the pushed head supports. Same defect class, inverted, and now load-bearing enough to be the blocker above. It resolves when the missing commit is pushed.

## NIT (carried forward) — the fallback silently reproduces the original bug

`Evaluator.cs:157` — if `program.Functions.TryGetValue(program.EntryPoint, …)` misses, control falls through to `EvaluateFunctionBody(program.Statement)`, which for a `Main`-only program is empty. The failure mode is then **no output and exit 0**: precisely #2984's symptom, reintroduced silently for any entry-point shape the lookup does not cover. Not reachable by any shape I probed, so not a blocker — but a fallback whose failure mode is indistinguishable from the bug being fixed is worth one line of comment saying so.

---

**Verdict: NOT MERGEABLE** on the blocker alone. The #2984 fix is correct, minimal, well placed and properly pinned; it is ready. The PR simply claims one more issue than it currently carries, and the missing piece is a commit the author already has. Push `8c54d5143` and re-measure the exit-code table against the pushed head, and this becomes MERGEABLE with two NITs.

*Evidence class: full Release solution builds (`-p:ContinuousIntegrationBuild=true`, no `-m:1`, never `--no-build`), direct `gsc`/`gsi` binaries, exit codes captured immediately and unpiped, md5 taken on `GSharp.Core.dll` rather than the drivers.*

DavidObando:
Addressed both SHOULD-FIX items in `aa95b2920`:

- Added explicit script-vs-REPL entry-point control. File-mode `gsi` enables entry-point execution; interactive `SessionEngine` declarations do not execute `Main`. A chained-compilation test pins both the initial `Main` declaration and the next declaration-only submission to no output. Before the fix, the first submission ran `Main` once and the next did not re-run it.
- Narrowed argument seeding to exactly one `[]string` parameter. Resolved entry-point bodies now use the `Functions` indexer so a broken invariant fails loudly instead of silently falling back.

Targeted tests: 10 passed. Branch and merge build against current `main` both succeed.

DavidObando:
The blocker report inspected stale head `d078d5385`. Before this update, both the remote branch and PR API already reported `9fe2c94ed`, including `src/Repl/Program.cs`, `Issue3010EntryPointExitCodeTests.cs`, and both `Fixes` lines.

I nevertheless rebased onto the freshly fetched current `origin/main` (`61b7a08c`) and force-pushed final head `de46eed5a`.

Content verification at the pushed head:

- `git diff $(git merge-base HEAD origin/main)...HEAD -- src/Repl/` contains `SessionEngine.cs` and `Program.cs`.
- Direct unpiped probe: `gsi` stdout `main`, exit 42; `gsc` + `dotnet` stdout `main`, exit 42.
- Targeted entry-point tests: 11 passed.
- PR body contains `Fixes #2984` and `Fixes #3010`.

DavidObando:
## Review: NOT MERGEABLE — 1 BLOCKER

**Trees measured.** `main` = `f5e9d8925`; PR head = `de46eed5aeac13260de1e186b85f55be2e3260f6`; computed merge base = `61b7a08c1` (PR is 8 commits behind main); merged tree I measured = **`3adebc007`**. Three-dot scope: **7 files, +350 / −6**. Base and head built Release, `-p:ContinuousIntegrationBuild=true`, **0 warnings / 0 errors** each.

---

### The prior blocker is cleared — the push is real

Verified three independent ways, since the previous review failed on exactly this:

1. `git diff 61b7a08c1 <head> --stat -- src/Repl/` is **non-empty**: `Program.cs` +16/−3, `SessionEngine.cs` +8/−1.
2. Blob SHAs genuinely differ from `main` (rules out the identical-blob trap): `Program.cs` `1a54403e8` → `7b5949d0e`; `SessionEngine.cs` `5e6918bfc` → `dbe6aff60`.
3. `gh pr view` reports `closes = 2984, 3010`. Both `Fixes` lines present, neither inverted.

New test file `test/Interpreter.Tests/Issue3010EntryPointExitCodeTests.cs` (+99) is present.

### Exit-code parity against the compiled oracle

Every probe run as `cmd > o.txt 2>&1; RC=$?` (never inline). Oracle = `gsc` + `dotnet` on the same source.

| probe | oracle `gsc`+`dotnet` | base `gsi` | head `gsi` | match |
|---|---|---|---|---|
| `func Main() int32` → 42 | rc=42 `main` | rc=0 **(no output)** | rc=42 `main` | ✔ |
| `func Main()` void | rc=0 `main-void` | rc=0 **(no output)** | rc=0 `main-void` | ✔ |
| `func Main() int32` → 0 | rc=0 `main-zero` | rc=0 **(no output)** | rc=0 `main-zero` | ✔ |
| no `Main` | rc=0 `no-main` | rc=0 `no-main` | rc=0 `no-main` | ✔ |
| `Main` → −1 | rc=**255** `neg` | rc=0 **(no output)** | rc=**255** `neg` | ✔ |
| `Main` → 300 | rc=**44** `big` | rc=0 **(no output)** | rc=**44** `big` | ✔ |
| top-level `return 42` | rc=42 `tls` | rc=0 `tls\|42` | rc=42 `tls` | ✔ |
| **trailing expr, no `return`** | **rc=0** | rc=0 `…\|42` | **rc=42** | ✘ |

Base reproduces #2984 plainly: an explicit `Main` is **never executed at all** — no output, rc=0. Head fixes it, and `gsi` and `gsc` **agree on POSIX `& 0xFF` truncation in both directions** (−1→255, 300→44), which was an open question. 7 of 8 shapes match the oracle exactly.

### Mutation — both production hunks are pinned

Artifact note: this PR's #3010 hunk compiles into **`out/bin/Release/Repl/gsi.dll`**, *not* `GSharp.Core.dll`. Hashing Core here shows no change and would read as a dead mutant.

| mutant | BUILD_RC | `gsi.dll` md5 | result |
|---|---|---|---|
| clean | 0 | `0569ea6e…` | 0 failed / 601 |
| **A** — delete the `is int` exit-code block, `Program.cs:68` | 0 | `379f226d…` | **2 die**: `ExplicitMainIntegerResultBecomesExitCode`, `TopLevelIntegerResultBecomesExitCode` |
| **B** — `evaluateEntryPoint: RunEntryPoint` → `true`, `SessionEngine.cs:216` | 0 | `817931fa…` | **1 dies**: `DeclarationOnlySubmissionsDoNotRunMain` |
| restored | 0 | `0569ea6e…` | cycle closes |

Condition-substitution mutants, not `if (true)`; `BUILD_RC` checked before every hash.

### Full suites on the merged tree `3adebc007` — unfiltered

| suite | failed | passed | skipped | total |
|---|---|---|---|---|
| `Interpreter.Tests` | 0 | 601 | 0 | 601 |
| `Core.Tests` | 0 | 7065 | 1 | 7066 |
| `Compiler.Tests` | 0 | 4013 | 0 | 4013 |

**0 failed / 11,679 passed / 1 skipped.** No new diagnostic descriptors (`new("GS####"` on the diff = 0; positive control = 377 on the tree), so the bidirectional docs gate is not implicated.

---

## BLOCKER 1 — `src/Repl/Program.cs:68`: `cell.Value is int` over-matches, silently turning an ignored value into the process exit code

The exit-code test is `cell.Value is int exitCode`, applied to whatever the last top-level expression evaluated to. It cannot tell "the program returned this" from "the program discarded this".

```gsharp
import System

func Compute() int32 {
    return 7
}

Console.WriteLine("done")
Compute()
```

| | stdout | exit |
|---|---|---|
| `gsc` + `dotnet` (oracle) | `done` | **0** |
| base `gsi` | `done`, `7` | 0 |
| head `gsi` | `done` | **7** |

A script ending in an int-returning call made for its side effect now terminates the process with that value. Base was also non-parity, but on **stdout**, with a correct exit code; head moves the divergence onto the **exit code**, which is the contract a CI pipeline reads. A script that passes today starts failing after this merge, with no diagnostic and no output difference to explain it — the same silent class as #3010 itself, inverted.

The PR's own tests do not cover this: `TopLevelIntegerResultBecomesExitCode` uses an explicit `return 3`, which is genuinely an exit code (I confirmed the oracle also exits 42 for top-level `return 42`). Nothing exercises a bare trailing expression.

**The compiler already computes the correct discriminator.** Loading the `gsc` output and reading `Assembly.EntryPoint.ReturnType`:

| source | emitted entry point | return type | process exit |
|---|---|---|---|
| top-level `return 42` | `<Main>$` | **`System.Int32`** | 42 |
| trailing ignored int call | `<Main>$` | **`System.Void`** | 0 |
| `func Main() int32` | `Main` | **`System.Int32`** | 42 |

**Fix:** gate the exit-code path on the entry point's declared return type being `int32` — the same signal the emitter uses — rather than on the runtime type of `cell.Value`. That requires plumbing the entry-point return type out through `SessionEngine`, and it makes all 8 shapes agree with the oracle instead of 7.

## SHOULD-FIX 1 — the `ContinueWith` invariant is still unpinned (`test/Interpreter.Tests/Issue2984ReplEntryPointTests.cs:17`)

`DeclarationOnlySubmissionsDoNotRunMain` covers **`RunEntryPoint = false`** only — the interactive case where `Main` must never run. The chained **`RunEntryPoint = true`** case, established by hand in the previous round, is guarded by nothing.

I measured it directly with a throwaway probe on the merged tree:

- submission 1 (declare `Main` printing `M-RAN`, returning 3) → output `M-RAN`, value `3`
- submission 2 (`func Helper() { }`) → output empty, no value
- submission 3 (`40 + 2`) → output empty, value `42`

So the invariant **holds today**: `Main` runs exactly once, at declaration, and is not re-executed. That is worth a test precisely because it is currently true and nothing would notice if it stopped being true — mutant B shows the neighbouring guard is one boolean away from firing `Main` on every submission.

## SHOULD-FIX 2 — merge-order collision with open PR #3024

`git merge-tree` against the merged tree `3adebc007`:

- vs **#3052** (`fea356435`) — clean, no conflict.
- vs **#3024** (`bab079854`) — **`CONFLICT (content): src/Repl/Program.cs`**.

Whichever lands second needs a manual resolution in the exit-code block. Worth sequencing deliberately rather than discovering it post-merge.

## NIT — retracting the `Evaluator.cs:157` concern; no diagnostic ID is needed

The suggestion was that the fallback there fails as "no output, exit 0", indistinguishable from #2984. That described the **base** behaviour. On this branch, `return EvaluateFunctionBody(program.Statement)` is the ordinary top-level-script path — the no-`Main` probe prints `no-main` and exits 0 correctly. A diagnostic there would fire on every normal script. The genuinely defensive branch above it is already loud: `program.Functions[program.EntryPoint]` throws on a missing body rather than degrading to an empty block, and the comment says so. **Recommend not allocating an ID.**

---

### Disclosures

- I reviewed in scratch worktrees under `/tmp/d3011/` (`base` = `main` `f5e9d8925`; `m` = merged `3adebc007`) rather than the assigned `gsharp-wt/i2984`, to avoid contending with another agent for that worktree. Both scratch trees end at porcelain-empty with `gsi.dll` hashed back to the clean `0569ea6e…`.
- My first entry-point-metadata probe was invalid: both `gsc` outputs are assembly-named `Default`, so co-loading them threw `FileLoadException` and produced no reading. The table above comes from the corrected one-assembly-per-run version.
- `Compiler.Tests` ran 1 h 5 m under machine load averaging 14–31; I waited out a period above load 30 before starting it. No run was reported that did not complete.

DavidObando:
## Addendum — re-measured on current `main`; verdict unchanged

`main` moved after my review (it is now **`5a27b5532`**, having gained #3062 and #3064). Because my original comment is anchored to the now-stale merged tree `3adebc007`, and because a fleet-wide correction landed regarding stale `GSharp.Core.dll` copies being loaded by the `gsc`/`gsi` drivers, I re-derived every driver-based measurement from scratch.

**New merged tree: `b594f74ba`** = `5a27b5532` + `de46eed5a`. Merge clean (`rc=0`). Solution build `-p:ContinuousIntegrationBuild=true`, **0 warnings / 0 errors**.

**Driver freshness verified explicitly.** After `dotnet build GSharp.sln`, all three `GSharp.Core.dll` copies — `out/bin/Release/Core/`, `.../Compiler/`, `.../Repl/` — carry the **same mtime to the second** (`18:03:35`) and identical md5. I also forced a recompile in the base tree and confirmed all three advanced together, so the driver here is not loading a stale copy.

### All nine probes reproduce exactly

| probe | oracle `gsc`+`dotnet` | base `gsi` | head `gsi` |
|---|---|---|---|
| `Main() int32` → 42 | rc=42 `main` | rc=0 (none) | rc=42 `main` |
| `Main()` void | rc=0 `main-void` | rc=0 (none) | rc=0 `main-void` |
| `Main() int32` → 0 | rc=0 `main-zero` | rc=0 (none) | rc=0 `main-zero` |
| no `Main` | rc=0 `no-main` | rc=0 `no-main` | rc=0 `no-main` |
| **bare trailing int** | **rc=0** `bare` | rc=0 `bare\|42` | **rc=42** `bare` |
| `Main` → −1 | rc=255 `neg` | rc=0 (none) | rc=255 `neg` |
| `Main` → 300 | rc=44 `big` | rc=0 (none) | rc=44 `big` |
| top-level `return 42` | rc=42 `tls` | rc=0 `tls\|42` | rc=42 `tls` |
| **ignored int-returning call** | **rc=0** `done` | rc=0 `done\|7` | **rc=7** `done` |

**BLOCKER 1 reproduces unchanged** on the current tree. Nothing moved.

### Mutation re-run (a rebase voids prior mutation evidence)

| step | BUILD_RC | `gsi.dll` md5 | driver probe p1 | suite |
|---|---|---|---|---|
| clean | 0 | `cc0f92f4…` | rc=42 `main` | 0 failed / 602 |
| mutant A (drop `is int` block) | 0 | `865b2348…` | **rc=0 `main\|42`** | **2 failed** — `ExplicitMainIntegerResultBecomesExitCode`, `TopLevelIntegerResultBecomesExitCode` |
| restored | 0 | `cc0f92f4…` | rc=42 `main` | — |

The driver probe flipping 42 → 0 → 42 in lockstep with the hash is a positive control that the mutant was genuinely observable through `gsi`, not a phantom survival.

### Other re-checks

- `Interpreter.Tests` on the merged tree: **0 failed / 602 passed / 602** (602, not 601 — one test arrived with the new `main` commits). No regression of the #3055 class, where a test landing on `main` after the merge base breaks a PR invisibly to CI.
- Cross-PR collision unchanged against current `main`: clean vs **#3052**; **`CONFLICT (content): src/Repl/Program.cs`** vs **#3024**.
- `Core.Tests` (7065/1 skipped/7066) and `Compiler.Tests` (4013/4013) figures in my original comment were measured on `3adebc007` and were **not** re-run here; only the four upstream PRs differ.

Working trees ended porcelain-empty with hashes round-tripped to clean.

DavidObando:
## Addendum 2 — BLOCKER 1 is broader than I first reported: an explicitly **void** `Main` leaks its final expression value

I re-examined my own probe set along the `{top-level, in-function}` axis and found I had tested the leak **only at top level**. Every one of my in-function probes used a `Main` with an explicit `return`, which is the shape the PR's own tests use. The untested cell — a **void** `Main` whose body simply ends in an expression — is where this gets worse.

Measured on merged tree **`b594f74ba`** (= `main` `5a27b5532` + `de46eed5a`). Base = `f5e9d8925`. No builds were run for these probes; all three `GSharp.Core.dll` copies (`Core/`, `Compiler/`, `Repl/`) were verified mtime-identical and md5-identical per tree beforehand.

| probe | oracle `gsc`+`dotnet` | base `gsi` | head `gsi` |
|---|---|---|---|
| `func Main()` void, ends with ignored `Compute() int32` | rc=**0**, `done` | rc=0, (no output) | rc=**7**, `done` |
| `func Main()` void, ends with bare `40 + 2` | rc=**0**, `bare` | rc=0, (no output) | rc=**42**, `bare` |
| `func Main()` void, ends with bare `"leaked"` | rc=0, `first` | rc=0, (no output) | rc=0, `first` + **`leaked`** |

```gsharp
import System

func Compute() int32 {
    return 7
}

func Main() {
    Console.WriteLine("done")
    Compute()
}
```

`gsc` + `dotnet`: prints `done`, exits **0**. `gsi` at this head: prints `done`, exits **7**.

Two things this changes about the finding as originally filed:

1. **It is not confined to top-level code.** A `Main` declared with **no return type at all** — which has no exit-code semantics in any reading of the language — nonetheless has its last evaluated expression adopted as the process exit code.
2. **It is not confined to `int`.** The third row shows the same leak on the stdout channel: a trailing string produces a spurious extra line that the compiled program does not emit. So the underlying defect is that `gsi` treats the entry point's final expression value as the program's result **regardless of the entry point's declared return type**, and `Program.cs` then dispatches on the runtime type of that value.

This does not change the recommended fix — it strengthens it. Gating on the entry point's **declared return type** resolves all three rows at once, because `func Main()` is declared void and would short-circuit to "no result" before either the `is int` or the `is not null` branch is reached. It is the same signal the emitter already uses; measured earlier from the `gsc` output:

| source | emitted entry point | return type | process exit |
|---|---|---|---|
| top-level `return 42` | `<Main>$` | `System.Int32` | 42 |
| trailing ignored int call | `<Main>$` | `System.Void` | 0 |
| `func Main() int32` | `Main` | `System.Int32` | 42 |

### Scope and staleness of these numbers

- These probes are true of merged tree **`b594f74ba`**. `main` has since advanced to **`185431c4`** (#3053, #3033, #3063). None of those three commits touch `src/Repl/`, `Evaluator.cs` or `Compilation.cs`, but zero file overlap is not proof of independence, so I am naming the SHA rather than claiming currency. I did not rebuild: machine load was 36.55, above the threshold at which I run builds.
- Unrelated check, run because it was asked of reviewers generally: this PR's diff contains **0** occurrences of the `is not LocalVariableSymbol`-as-first-disjunct pattern. Positive control — the pattern matches **3** sites elsewhere in `src/`, so that zero is absence rather than a broken search.

DavidObando:
## Review: MERGEABLE — 1 SHOULD-FIX (undisclosed sample regression), 2 NITs, and a duplicate-PR finding that needs a decision

**The PR was rebased while I was reviewing; I re-measured rather than trust the old numbers.**

| | measured | re-verified |
|---|---|---|
| PR head | `1ab65366` | **`94b65165`** (current) |
| `main` | `abfab99c` | **`6ec381c5`** (current) |
| merged tree | `33008888` (real merge, `MERGE_RC=0`) | `94b65165` (PR now fast-forwards onto `main`) |

The rebase pulled in `main`, not new PR work: the four production blobs — `Compilation.cs`, `Evaluator.cs`, `SessionEngine.cs`, `Program.cs` — are **byte-identical** between the two heads (verified per file by blob SHA), so the mutation results below carry over. Every headline probe was nevertheless re-run on the current pair and reproduced exactly. Three-dot scope: **8 files, +970 / −7** (4 production, 4 test). Both trees built solution-wide with `-p:ContinuousIntegrationBuild=true`, `BUILD_RC=0`, 0 errors.

**Driver freshness verified before every probe**, on all three `GSharp.Core.dll` copies including `Repl/` where `gsi` lives:

| tree | Core/ | Compiler/ | Repl/ | md5 |
|---|---|---|---|---|
| base `abfab99c` | 11:08:40 | 11:08:40 | 11:08:40 | `ffe5c860` |
| merged `33008888` | 11:06:33 | 11:06:33 | 11:06:33 | `88899a3f` |
| base `6ec381c5` (re-verify) | 12:00:50 | 12:00:50 | 12:00:50 | `e967f2a9` |
| head `94b65165` (re-verify) | 11:58:32 | 11:58:32 | 11:58:32 | `49d82805` |

Every probe compiled and ran in its **own directory, asserted empty** before use.

---

## The verbatim repro from #2984, run byte-for-byte first

```gsharp
package P
import System

func Main() {
    Console.WriteLine("hello from Main")
}
```

| tree | bare `gsc` (evaluate) | `gsc /out:` + `dotnet` (oracle) | `gsi` |
|---|---|---|---|
| base `abfab99c` | `Success.` only | `hello from Main`, rc 0 | rc 0, **empty** ← the bug |
| merged `33008888` | `hello from Main` + `Success.` | `hello from Main`, rc 0 | rc 0, **`hello from Main`** ✅ |

The verbatim #3010 repro likewise closes: base `gsi` rc 0 / empty → head `gsi` rc **3** / `main`, matching the oracle's rc 3 exactly.

**Note for the fleet rule.** The standing rule is "a `func Main()`-only file runs nothing under bare `gsc` **and** `gsi`". This PR changes **both** halves, not just the `gsi` half: `src/Compiler/Program.cs:311` calls the 1-arg `Compilation.Evaluate(variables)`, whose new default is `evaluateEntryPoint: true`. Bare `gsc` now executes `Main`'s body. Its exit code is unaffected because `Interpret()` discards `result.Value`, so the un-gated `useEntryPointReturnType: false` on that overload is harmless — but the side effects are real and the rule needs updating in both columns.

---

## The prior BLOCKER is fixed

The last review (addendum 2, at head `de46eed5a`) found a **void** `Main` adopting its final expression as the program result. Re-measured at this head, all three rows now match the oracle:

| probe | oracle | base `gsi` | head `gsi` (was at `de46eed5a`) |
|---|---|---|---|
| void `Main`, ends with ignored `Compute() int32` | rc 0, `done` | rc 0, ∅ | rc **0**, `done` (was rc 7) ✅ |
| void `Main`, ends with bare `40 + 2` | rc 0, `done` | rc 0, ∅ | rc **0**, `done` (was rc 42) ✅ |
| void `Main`, ends with bare `"leaked"` | rc 0, `first` | rc 0, ∅ | rc 0, `first` (no longer leaks `leaked`) ✅ |

Fixed the way that review recommended — gating on the declared return type (`FunctionSymbol.Type`), the same signal emit uses — plus removing the unconditional `Console.WriteLine(cell.Value)` from `Program.cs`.

---

## Per-hunk mutation: every hunk pinned, zero survivors

Baseline merged tree: **Interpreter.Tests 688/688**, unfiltered. Each mutant asserted applied at exactly 1 site (a non-applying patch reads exactly like a survival); `BUILD_RC` checked before believing any result; all product mutants, condition inversion where the shape allowed.

| # | hunk | mutation | build | killed | representative test |
|---|---|---|---|---|---|
| M1 | `Evaluator.cs:176` dispatch | `evaluateEntryPoint &&` → `!evaluateEntryPoint &&` | 0 | **31** | `PackageScopeMainRuns` |
| M2 | `Evaluator.cs:217` void gate | `Type == Void ? null : value` → `value` | 0 | **8** | `ExplicitVoidMainDiscardsBareIntegerExpression` |
| M3 | `Evaluator.cs:209-211` refusal | all three `!=` → `==` (never throws) | 0 | **2** | `ExplicitStringMainIsRejectedBeforeExecution` |
| M4 | `Program.cs:71` exit code | `return exitCode;` → `return exitCode * 0;` | 0 | **4** | `ExplicitMainIntegerResultBecomesExitCode` |
| M5 | `Evaluator.cs:185` args seeding | `==` → `!=` | 0 | **2** | `MainArgsReceivesEmptyArray` **and** `NonArgsMainParameterIsNotSeededWithStringArray` |
| M6 | `SessionEngine.cs:220` plumbing | `RunEntryPoint` → `!RunEntryPoint` | 0 | **125** | `DeclarationOnlySubmissionsDoNotRunMain` |
| M7 | `Compilation.cs:293` 1-arg default | `evaluateEntryPoint: true` → `false` | 0 | **10** | `ClassScopedSharedMainRuns` |

M5 kills **both adjacent arms independently**, so the seeding condition is not masked. Round-trip closed: after restoring all seven, `git status` clean and Core md5 returned to `88899a3f` exactly.

**Test quality is above the bar here.** `Issue3010EntryPointDriverMatrixTests.cs` parameterises 8 shapes × **3 drivers** (`CompilerEvaluation`, `CompilerEmission`, `Interpreter`) and asserts exit code, stdout **and** empty stderr on each — the evaluate column that is normally structurally invisible to CI is explicitly covered, and the bare-`gsc` contract difference (always rc 0, plus a `Success.` banner) is modelled deliberately rather than papered over.

---

## SHOULD-FIX — the two samples named in #2984 now fail, and no gate sees it

`src/Compiler/Program.cs:311`, `src/Repl/Program.cs:58`

Issue #2984's "Repro samples" section names `samples/SpanComprehensive.gs` and `samples/RefStructGenericField.gs` as the motivating corpus files. I verified they are the complete set: **2 of 122** samples contain `func Main(` (positive control: 91 sample files match `func `). Identical sources on both trees.

| sample | oracle `gsc /out:` + `dotnet` | base `gsi` | head `gsi` | head bare `gsc` |
|---|---|---|---|---|
| `SpanComprehensive.gs` | rc 0, `60/10/2` | rc 0, ∅ | **rc 1** | **rc 1** (was rc 0) |
| `RefStructGenericField.gs` | rc 0, `3` | rc 0, ∅ | **rc 1** | **rc 1** (was rc 0) |

```
(1,1,1,136): error GS9999: Specified method is not supported.
    (System.ReadOnlySpan`1[[System.Int32, ...]])nums
```

This is a graceful diagnostic, not a crash, and it is arguably an improvement over silently doing nothing. But it is a **real rc 0 → rc 1 change on shipped corpus files, in the one column where the compiled oracle works correctly**, and it is the only divergence I found where `gsc` succeeds and `gsi` does not — which is precisely the acceptance criterion for this PR.

**No gate covers it.** `SampleConformanceTests` (`test/Compiler.Tests/LanguageConformance/SampleConformanceTests.cs:16`) compiles each sample through `gsc.dll` and executes it under `dotnet` — **emit-only**. Nothing runs `samples/` through `gsi`. That is why CI is green.

I do **not** think this should block: the alternative is keeping the silent no-op, which is worse, and interpreter `ReadOnlySpan` support is plainly out of scope. But it must be **owned rather than discovered later** — the PR body's matrix does not mention it, and #2984's repro section will otherwise read as satisfied when it is not. Recommend stating it in the PR body and filing a follow-up for interpreter `Span`/`ref struct` support, referenced from #2939 (which excluded these two samples "until this is fixed").

---

## NIT 1 — `func Main(x int32)` leaks a raw .NET message

`src/Core/CodeAnalysis/Evaluator.cs:185`

```gsharp
func Main(x int32) {
    Console.WriteLine(x)
}
```

Head `gsi`: rc 1, `error GS9999: The given key 'x int32' was not present in the dictionary.` That is a dictionary-lookup implementation detail surfacing as a user diagnostic. Pre-existing in the emit column and **not caused by this PR** — the oracle compiles this and crashes at runtime with rc 134 on **both** trees, because nothing in `src/` diagnoses a bad `Main` signature (`explicitMain` is `FirstOrDefault(f => f.Name == "Main")`, any signature; grep for entry-point signature validation returns 0 hits with a working positive control). The PR's `NonArgsMainParameterIsNotSeededWithStringArray` correctly pins non-seeding; only the message is poor. A binder-level diagnostic for invalid entry-point signatures would fix `gsc` rc 134 and this message at once — worth an issue, not a change here.

## NIT 2 — evidence SHAs in the PR body are stale

The "Bidirectional evidence" table cites base `366392537` / head `5dacf4a60`; the current head is `1ab65366`. The numbers reproduce exactly at the current head — I re-ran them — so this is presentation only, but a reader cannot verify the table against the branch as it stands.

---

## The #3103 interaction: they are not two PRs, they are the same change twice

This needs a decision before either merges.

```
git diff --full-index pr3103 pr3011 -- src/     ->  EMPTY
```

All four production blobs are **identical**, verified individually:

| file | `pr3011` blob | `pr3103` blob |
|---|---|---|
| `Compilation.cs` | — | **IDENTICAL** |
| `Evaluator.cs` | — | **IDENTICAL** |
| `SessionEngine.cs` | — | **IDENTICAL** |
| `Program.cs` | — | **IDENTICAL** |

Both branches carry the **same eight commit messages in the same order** (`Invoke explicit Main in gsi` → … → `Test entry point exit codes across drivers`), differing only in SHA. #3011 is the **superset**: it additionally contains `Issue3010EntryPointDriverMatrixTests.cs` (+241, the three-driver matrix) and a larger `Issue3010EntryPointExitCodeTests.cs` (352 vs 342).

So there is no semantic interaction to analyse — there is a duplicate. #3011 already closes **both** #2984 and #3010; its title says so. If both merge, the second will conflict on the two shared test files (same paths, different contents), not on production code. **Recommend merging #3011 and closing #3103 as superseded**, and telling `duck-3018` to stop.

## Collision zone: `src/Repl/Program.cs` — handled correctly

#3024's file-path argument is **preserved**: the head still reads `engine.Evaluate(File.ReadAllText(arg), arg)`. The PR takes HEAD and drops nothing; `gsi` diagnostics keep their file path. Verified in my probes — every `gsi` error above is anchored to a path.

---

## Coverage

- Merged tree `33008888`: **Interpreter.Tests 688/688** unfiltered, **Core.Tests 7088 passed / 0 failed / 1 skipped**. Re-verified on the current head `94b65165` over `main` `6ec381c5`: **Interpreter.Tests 688/688**.
- Branch CI green on all 25 checks including `cs2gs-corpus`, `cs2gs-oahu`, `e2e` and all 8 compiler shards. I did not re-run Compiler.Tests on the merged tree (~1h20m); residual risk is bounded because the diff touches nothing under `Emit/` and every oracle (`gsc /out:` + `dotnet`) output I measured was byte-identical between base and merged.
- Ambiguous shape (top-level statements **and** `func Main()`): identical on both trees — top-level wins with warning `GS0166`, `gsi` matches the oracle. The binder settles this at `Binder.cs:6361`, so the two drivers cannot drift.
- Signature axis measured: `{void, int32, uint32, string, struct}` × `{none, args []string, x int32}` × `{top-level present, absent}` × 3 drivers.

## Disclosure of my own measurement errors

1. My first `git fetch` used a non-`+` refspec, so the rejected update silently left `pr3011` at the **stale** `64260035`; caught by comparing against the API head. All results above are from the forced re-fetch at `1ab65366`.
2. My `s3_args` probe failed on **both** trees — my own missing `import Gsharp.Extensions.Go` for `len`, not a defect. Re-run with `args.Length`: head `gsi` rc 22 / `0`, matching the oracle exactly.
3. I hypothesised from source reading that `func Main() string` would escape as an unhandled `InvalidOperationException` (the `catch` in `Compilation.cs:368` only handles `EvaluatorException`). **The probe disproved it** — it surfaces as a clean `GSI001`, rc 1, no stdout. Reporting the dead hypothesis rather than the guess.
4. I started the Core.Tests run at load **36.89**, above the threshold at which I should have waited. It completed in 5m23s and passed, but the timing was mine to get wrong.

Issue #2991, opened 2026-08-01
gsi: distinct lambdas report the same Method.Name


DavidObando:
## Problem

`samples/FuncToSystemDelegate.gs` prints `Method.Name` for two distinct lambdas:

```
compiled: <lambda1> / <lambda2>
gsi:      lambda_method521 / lambda_method521
```

Two separate defects in one line of output:

1. **Naming scheme** differs from the emitted one (`<lambdaN>`).
2. **Uniqueness is lost** — two different lambdas report the *same* name, so the interpreter's synthesized method identity does not distinguish them.

(2) is the substantive one: any code that keys on delegate method identity — deduplication, event bookkeeping, diagnostics — behaves differently under `gsi`.

## Expected

Distinct lambdas must yield distinct `Method.Name`, and ideally the same `<lambdaN>` scheme the emitter uses so golden output matches.

Found while triaging #2939; excluded there until fixed.

DavidObando:
## Triage — confirmed on `a6b1561f2`, but **not a regression from #2935**

Reproduced with `samples/FuncToSystemDelegate.gs` (byte-identical across all SHAs below, md5 `61a6aaa0…`) under `out/bin/Release/Repl/gsi.dll`:

| SHA | role | `gsi` output |
|---|---|---|
| `d4038f1d` | true pre-PR base | `GS9999: Object type System.Delegate does not match target type ClosureValue` |
| `632d133d` | #2935 round 1 | `lambda_method97` / `lambda_method98` |
| `72490605` | #2935 round 2 | `lambda_method98` / `lambda_method98` |
| `a6b1561f2` | current main | `lambda_method98` / `lambda_method98` |
| — | compiled (golden) | `<lambda1>` / `<lambda2>` |

**At the true base the sample does not run at all.** `gsi` has never printed distinct-and-correct names for it, so this is an **incomplete fix / newly-observable defect**, not a regression — you cannot regress from `GS9999`. Recommend **not** labelling `regression`; label as follow-up to #2935.

One caveat the fixer must know: the *uniqueness* sub-property did flip **inside** the PR, between round 1 and round 2, when the per-delegate-type factory cache was added. That cache was a review recommendation of mine (it removed a ~32× lambda-creation regression). **Do not "fix" #2991 by reverting the cache** — see the trade-off section.

## Mechanism, to `file:line`

[...]

Issue #2990, opened 2026-08-01
gsi: untyped lambda argument is not coerced to Func at a CLR call boundary


DavidObando:
## Problem

Passing an untyped `func` literal to a CLR generic method that expects `Func<,>` throws.

`samples/SliceLinqUntypedLambda.gs`:

```
compiled: 3 6 2 8 30 fig kiwi apple
gsi:      GS9999: Object of type 'GSharp.Core.CodeAnalysis.ClosureValue'
          cannot be converted to type 'System.Func`2[System.Object,System.Boolean]'.
```

## Root cause

The interpreter represents a `func` literal as `ClosureValue`. When such a value is handed to a reflection-invoked CLR method whose parameter is a constructed `Func<,>`, no conversion happens and the CLR argument binder rejects it.

`src/Core/CodeAnalysis/Evaluator.Calls.cs:603-610` already converts a `ClosureValue`-backed `Delegate` to a target CLR delegate type via `CreateInterpreterDelegate`, but only on an explicit `BoundConversionExpression`. Argument coercion at a reflection call site does not route through it, so an *inferred* untyped lambda argument never gets wrapped.

## Fix sketch

Coerce arguments at the CLR-invocation boundary: when a parameter type is a delegate and the value is a `ClosureValue`, wrap it with `CreateInterpreterDelegate` before `Invoke`. That is the same helper #2935 introduced, applied one layer out.

Found while triaging #2939; excluded there until fixed. Sibling of #2935.

DavidObando:
## Triage — confirmed on `a6b1561f2`; pre-existing, and the stated trigger is wrong

`samples/SliceLinqUntypedLambda.gs` under `out/bin/Release/Repl/gsi.dll`:

| SHA | `gsi` output |
|---|---|
| `d4038f1d` (pre-#2935 base) | `GS9999: … 'ClosureValue' cannot be converted to type 'System.Func`2[System.Object,System.Boolean]'` |
| `a6b1561f2` (main) | identical |

Byte-identical failure at base and head, so #2935 neither caused nor touched it. Not a regression.

## The "untyped lambda" framing is not the trigger — a typed lambda fails identically

```gsharp
data struct Item { var V int32 }
var xs = []Item{ ... }
xs.Where((i Item) -> i.V > 0).Count()   // TYPED    -> same GS9999
xs.Where((i)      -> i.V > 0).Count()   // untyped  -> same GS9999
```

Both fail with the same message. Narrowing further:

[...]

Issue #2985, opened 2026-08-01
gsi: conversion to a user-declared named delegate type throws Unexpected type


DavidObando:
## Problem

Assigning a `func` literal to a **user-declared named delegate type** throws in the interpreter. Compiled output is correct.

```gsharp
type Greeter = delegate func(name string)

var hello Greeter = func(name string) { Console.WriteLine("hello, " + name) }
hello.Invoke("world")
```

`gsi` → `GS9999: Unexpected type Greeter` (no output). `gsc` → `hello, world`.

## Root cause

`src/Core/CodeAnalysis/Evaluator.Calls.cs:641` — the conversion evaluator's final `else` throws. A `DelegateTypeSymbol` declared in G# has no backing `ClrType` at interpretation time, so it matches none of the preceding branches:

- the `value is Delegate && node.Type?.ClrType is Type` branch (`:603`) requires a CLR delegate type — added for CLR delegates by #2935;
- the reference-upcast branch (`:611`) requires `ClrType != null`;
- the numeric and enum branches require `ClrType`.

There is no `node.Type is DelegateTypeSymbol` case.

## Fix sketch

Add a branch alongside `:603` that materializes an interpreter callable for `DelegateTypeSymbol`, reusing `CreateInterpreterDelegate`. Generic (`Mapper[int32]`) and variadic (`StringJoiner`) named delegates must go through the same path.

## Repro samples

`samples/NamedDelegate.gs` (`Unexpected type Greeter`), `samples/GenericNamedDelegate.gs` (`Unexpected type Mapper[int32]`), `samples/VariadicDelegate.gs` (`Unexpected type StringJoiner`).

Sibling of #2935, which fixed CLR delegate materialization but not user-declared named delegate types. Found while triaging #2939; excluded there until fixed.

DavidObando:
## Triage #2985 — REPRODUCED. Root-cause line correct. Fix sketch is bigger than needed.

[...]

PR #3012, opened 2026-08-01
Fix interpreter lambda body lowering

### LINKED ISSUE #3003: gsi: an 'if' statement inside a lambda body throws 'Unexpected node IfStatement'

DavidObando:
## Summary
- lower and cache function-literal bodies when the interpreter materializes closures
- cover all four lowered statement forms plus condition-for, switch, and method-group controls
- cover the nullable-returning `Optional.FlatMap` reproduction in the referenced in-process interpreter harness

## Verification
- base -> head: `IfStatement`, `ForRangeStatement`, `ForEllipsisStatement`, and `ForInfiniteStatement` each changed from `GS9999` to expected output
- base -> head: isolated `Optional.FlatMap` changed from `TargetInvocationException` to `a`
- targeted interpreter regressions: 3 passed

## Full #2995 sample
Direct `gsi samples/GsharpExtensionsOptional.gs` does not reach `FlatMap`: it exits 1 with 19 `GS0159` diagnostics because the CLI cannot resolve `Gsharp.Extensions`. The compiled expected line is therefore not reproduced end-to-end. #2995 remains open; the exact line-by-line measurement is recorded on that issue.

Fixes #3003
Related: #2995

### DIFF
--- a/src/Core/CodeAnalysis/Binding/BoundFunctionLiteralExpression.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundFunctionLiteralExpression.cs
@@ -44,4 +44,6 @@ public sealed class BoundFunctionLiteralExpression : BoundExpression
     public override TypeSymbol Type => FunctionType;
 
     public override BoundNodeKind Kind => BoundNodeKind.FunctionLiteralExpression;
+
+    internal BoundBlockStatement LoweredBody { get; set; }
 }
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -759,7 +759,8 @@ public sealed partial class Evaluator
             captured[v] = LookupVariable(v);
         }
 
-        return MaterializeClosure(new ClosureValue(node.Function, node.Body, node.FunctionType, captured));
+        var body = node.LoweredBody ??= (BoundBlockStatement)Lowering.Lowerer.Lower(node.Body);
+        return MaterializeClosure(new ClosureValue(node.Function, body, node.FunctionType, captured));
     }
 
     private object EvaluateMethodGroupExpression(BoundMethodGroupExpression node)
--- a/test/Interpreter.Tests/Issue3003LambdaBodyLoweringTests.cs
+++ b/test/Interpreter.Tests/Issue3003LambdaBodyLoweringTests.cs
@@ -1,0 +1,127 @@
+// <copyright file="Issue3003LambdaBodyLoweringTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using GSharp.Core.CodeAnalysis.Compilation;
+using GSharp.Core.CodeAnalysis.Symbols;
+using GSharp.Core.CodeAnalysis.Syntax;
+using Xunit;
+
+namespace GSharp.Interpreter.Tests;
+
+/// <summary>
+/// Issue #3003: function-literal bodies must be lowered before evaluation.
+/// </summary>
+public class Issue3003LambdaBodyLoweringTests
+{
+    [Fact]
+    public void LambdaBody_LowersIfAndAllForForms()
+    {
+        const string Source = """
+            let choose = func(n int32) string {
+                if n > 0 {
+                    return "positive"
+                }
+                return "zero"
+            }
+
+            let sumRange = func() int32 {
+                var sum = 0
+                for value in []int32{1, 2, 3} {
+                    sum += value
+                }
+                return sum
+            }
+
+            let countEllipsis = func() int32 {
+                var count = 0
+                for _ in 0 ... 3 {
+                    count++
+                }
+                return count
+            }
+
+            let countInfinite = func() int32 {
+                var count = 0
+                for {
+                    count++
+                    if count == 3 {
+                        break
+                    }
+                }
+                return count
+            }
+
+            Console.WriteLine(choose(1))
+            Console.WriteLine(sumRange())
+            Console.WriteLine(countEllipsis())
+            Console.WriteLine(countInfinite())
+            """;
+
+        Assert.Equal("positive\n6\n3\n3\n", Evaluate(Source));
+    }
+
+    [Fact]
+    public void LambdaBody_PreservesConditionForSwitchAndMethodGroupDelegate()
+    {
+        const string Source = """
+            func sign(n int32) string {
+                if n > 0 {
+                    return "positive"
+                }
+                return "zero"
+            }
+
+            let countTo = func(n int32) int32 {
+                var count = 0
+                for count < n {
+                    count++
+                }
+                return count
+            }
+
+            let classify = func(n int32) string {
+                switch n {
+                    case 1 {
+                        return "one"
+                    }
+                    default {
+                        return "other"
+                    }
+                }
+            }
+
+            let signDelegate (int32) -> string = sign
+
+            Console.WriteLine(countTo(3))
+            Console.WriteLine(classify(1))
+            Console.WriteLine(signDelegate(1))
+            Console.WriteLine(signDelegate(0))
+            """;
+
+        Assert.Equal("3\none\npositive\nzero\n", Evaluate(Source));
+    }
+
+    private static string Evaluate(string source)
+    {
+        using var outWriter = new StringWriter();
+        var previousOut = Console.Out;
+        Console.SetOut(outWriter);
+        try
+        {
+            var result = new Compilation(SyntaxTree.Parse(source))
+                .Evaluate(new Dictionary<VariableSymbol, object>());
+
+            Assert.Empty(result.Diagnostics);
+        }
+        finally
+        {
+            Console.SetOut(previousOut);
+        }
+
+        return outWriter.ToString().Replace("\r\n", "\n");
+    }
+}
--- a/test/Interpreter.Tests/Issue750ConstraintOverloadInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue750ConstraintOverloadInterpreterTests.cs
@@ -99,6 +99,27 @@ public class Issue750ConstraintOverloadInterpreterTests
         Assert.Equal("beta\n33\n42\n", Evaluate(source));
     }
 
+    [Fact]
+    public void FlatMap_NullableReturningLambdaWithIf_RunsInInterpreter()
+    {
+        var source = """
+            import System
+            import Gsharp.Extensions.Optional
+
+            let name string? = "ada"
+            let first = name.FlatMap(func(s string) string? {
+                if s.Length > 0 {
+                    return s.Substring(0, 1)
+                }
+                return nil
+            })
+
+            Console.WriteLine(first ?? "<none>")
+            """;
+
+        Assert.Equal("a\n", Evaluate(source));
+    }
+
     private static string Evaluate(string source)
     {
         // The Compilation.Default reference resolver enumerates


DavidObando:
## Review: no blocking defect in the fix; one SHOULD-FIX on the second declared closure

Reviewed at `fae05b403`, three-dot against the computed merge base `87ca66e85` (`main` = `f70626452`). Static analysis only — no build, no test run.

Short version: **the production change is correct, minimal, and reuses an existing house pattern**, and I can prove it addresses the root cause rather than the named symptom without running anything. The one thing I would not merge as-is is `Fixes #2995`.

---

### The fix is the house pattern, not an invention

`(BoundBlockStatement)Lowerer.Lower(literal.Body)` already exists verbatim at two sites:

- `src/Core/CodeAnalysis/Binding/RefStructAsyncLivenessAnalyzer.cs:658`
- `src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs:2862`

and the design is documented at `src/Core/CodeAnalysis/Binding/BoundTreeRewriter.cs:132` — literal bodies are "lowered independently at the point it is hosted for emission." The interpreter simply never had such a hosting point. `Evaluator.Expressions.cs:762` adds it. That is the smallest change that fits the existing design.

### Root cause, not symptom — provable statically

The interpreter's statement dispatch handles a fixed set of `BoundNodeKind`s and throws `Unexpected node {Kind}` at `Evaluator.Statements.cs:168` for the rest. Differencing the declared statement kinds against the handled ones:

```
statement kinds declared : 20
handled by Evaluator     : 16
declared but NOT handled : ForEllipsisStatement ForInfiniteStatement ForRangeStatement IfStatement
```

**Exactly four kinds depend on lowering, and they are exactly the four this PR's first test pins.** There is no fifth lowering-dependent statement kind, so there is no sibling shape left broken — this is a complete fix for the class, not a guard on one instance. The PR's verification claim (`IfStatement`, `ForRangeStatement`, `ForEllipsisStatement`, `ForInfiniteStatement` each `GS9999` → expected) matches that set precisely.

I also checked both closure construction sites, since a guard on one caller is the usual failure mode here:

- `Evaluator.Expressions.cs:763` — the patched literal path.
- `Evaluator.Expressions.cs:782` — the method-group path, whose body comes from `program.Functions[node.Function]`, which is already lowered by the binder. Correctly untouched.

`Lowerer` carries **no static mutable state** — `labelCount` is an instance field and both `Lower` overloads construct a fresh `Lowerer` — so the `??=` memoization is safe under the interpreter's goroutine/`select` concurrency. Worst case two threads lower independently and each uses its own equivalent result. Rewritten literal nodes are built through the constructor, which does not carry `LoweredBody`, so no stale cache can propagate.

---

### SHOULD-FIX — `Fixes #2995` is not supported at the scope the issue is written at

#2995 is about `samples/GsharpExtensionsOptional.gs`. Its measured evidence is the whole sample:

```
compiled: ADA A default computed <filtered out> present: ada ada 14 -1 7 count present: 7
gsi:      ADA
          GS9999: Exception has been thrown by the target of an invocation.
```

`gsi` died on the **second** output line. Everything after `FlatMap` — `OrElse`, `OrCompute`, `Filter`, `IfPresent`, `OrThrow`, and the entire value-typed section — has therefore **never been observed running under `gsi` at all**. The new test pins the `FlatMap` expression in isolation, which is necessary but is not the issue's scope.

The root cause genuinely is shared (the `if` inside that lambda is one of the four unlowered kinds), so this PR very plausibly does fix it. But "plausibly" is what closes live bugs: merging asserts the sample now runs end-to-end, and that is unmeasured. This is the direction that costs the most when wrong.

**Resolution — one program run:** execute `samples/GsharpExtensionsOptional.gs` under `gsi` at this head and diff against the compiled line above. If it matches, `Fixes #2995` is sound and the run is worth pasting into the PR body. If it diverges further down, retitle to `Refs #2995` and leave the issue open. Report the produced value, not the exit status — `gsi` returns 0 while printing `GS9999`.

### NIT — #2995 has no findable pin in the tree

The only issue number appearing anywhere in the added lines is `#3003`; the string `2995` occurs zero times. The test that pins it is `FlatMap_NullableReturningLambdaWithIf_RunsInInterpreter`, living in `test/Interpreter.Tests/Issue750ConstraintOverloadInterpreterTests.cs`.

Reusing that file is the right call — its `Evaluate` helper already resolves the `Gsharp.Extensions` references, so duplicating that harness would have been the worse option. But against the repo's `Issue<N>...Tests.cs` convention, anyone grepping `2995` to find its regression pin finds nothing. A single `// Issue #2995:` comment above the `[Fact]` fixes it.

### NIT — the one mutable member on an otherwise immutable node

`BoundFunctionLiteralExpression.cs:48` adds `internal BoundBlockStatement LoweredBody { get; set; }` — the only `{ get; set; }` in a file where every other member is `{ get; }`, with no comment. I verified it is safe (above), but the next reader has to redo that verification to know it.

Two options, either fine:

- Drop the cache. `var body = (BoundBlockStatement)Lowering.Lowerer.Lower(node.Body);` is a one-line change that removes the second file from the PR entirely. Correctness is identical; you lose re-lowering per materialization, which matters only for a literal created inside a hot loop.
- Keep it and add one line saying why mutation is safe — fresh `Lowerer` per call, no static state, rewritten nodes start `null`.

I would not block on either, but the cache is currently paying a real readability cost for an unmeasured performance benefit.

---

### Undeclared closures — none found

I checked every open `gsi`/interpreter issue touching lambdas or closures:

| issue | verdict | evidence |
|---|---|---|
| #2976 `yield` in function-literal iterator bodies | **DOES NOT FIX** | the interpreter cause is `Evaluator.Calls.cs:188` never marking a literal body as an iterator body; that file is untouched. Lowering leaves `BoundYieldStatement` in place, so the rejection is unchanged. |
| #2938 interpreter loses delegate exception diagnostics | **DOES NOT FIX** | the `InnerException` masking is untouched; #2995's own note attributes this to #2938 and it correctly stays open. |
| #2984, #2985, #2990, #2991 | **DOES NOT FIX** | unrelated causes (invocation, conversion, coercion, symbol naming). |

### Docs gate — zero exposure

Worth stating explicitly since every open branch's CI predates #2981 making `DiagnosticIdUniquenessTests` mandatory: this PR adds **0** `DiagnosticDescriptor` lines, **0** `GS####` mentions, and touches **0** docs files. It cannot red `main` through that gate.

### Adjacent risk, explicitly not this PR's job

`BoundLocalFunctionDeclaration` has a case in `MethodBodyEmitter.cs:381` but **none** in any `Evaluator*` file, so it would fall through to the same `Unexpected node` throw. `BoundTreeRewriter.cs:132` groups generic local-function bodies with literal bodies as deliberately un-rewritten, i.e. the same shape #3003 describes. I could not determine statically whether that node survives to the evaluator or is rewritten away first, and I did not run anything. If it is reachable, it is a separate issue in the same family — worth filing, not worth expanding this PR.

---

**Summary:** BLOCKER — none. SHOULD-FIX — measure `Fixes #2995` at sample scope or downgrade it to `Refs`. NIT — add a `#2995` marker to the test; document or drop `LoweredBody`'s mutability. The `Fixes #3003` claim is sound and, unusually, provably complete.

Issue #1919, opened 2026-07-03
gsc: async lambdas ICE (GS9998) under /reference: (MetadataLoadContext) compilation


DavidObando:
`let twice async (int32) -> int32 = async (x int32) -> await Task.FromResult(x * 2)` compiles in default host mode but ICEs (`GS9998: Derived classes must provide an implementation`) when compiled with explicit `/reference:` metadata — the cs2gs pipeline mode. Every async-lambda form reproduces it. (grid app G10).
## Definition of done
- [ ] Fix + `Issue<N>*.cs` regression test
- [ ] Re-enable the quarantined grid fixture (or add one) and re-capture its baseline
- [ ] `tools/cs2gs/coverage/csharp-construct-inventory.json` row updated

_Discovered by the ADR-0138 conformance-grid build-out (gsc 0.2.585+01ad6ec964); quarantined sources live next to each grid app under `tools/cs2gs/corpus/grid/`._

PR #3013, opened 2026-08-01
Hoist select receive bindings in synchronous iterators

### LINKED ISSUE #2975: select-arm bindings are not treated as binding sites by state-machine hoisting analysis

DavidObando:
## Summary

- collect `BoundSelectCase.Variable` bindings as synchronous iterator locals
- preserve the existing byref-like exclusion used for hoisted fields
- add structural and emitted-runtime regressions

## Root cause

`src/Core/CodeAnalysis/Lowering/Iterators/IteratorRewriter.cs:187` builds the synchronous iterator field map through `LocalCollector`, but that collector recorded variable declarations only. A receive-bind arm stores its binding as the bare `BoundSelectCase.Variable` symbol, not a `BoundVariableDeclaration`, so the variable never entered `HoistedLocals` or the resulting field map. The fix is the select visitor at `IteratorRewriter.cs:216`.

`HoistedFieldRewriter` already consumes this map correctly. `BoundSelectCase` is not a `BoundNode`, so there is no declaration or assignment node for that rewriter to redirect.

## Runtime dependency

This hoisting fix and #2967 are complementary halves of the mechanism. This PR creates the synchronous iterator field and redirects reads to it; #2967 copies the select receive out-slot into that field. #2967 merged as `a8ee8f22c` before this branch was rebased. This PR intentionally does not touch `MethodBodyEmitter.Statements.cs`.

The exact four-cell probe was rebuilt from source in every configuration and its printed value was read directly:

| Configuration | Printed value |
| --- | ---: |
| `47a400359` (neither fix) | 7 |
| `47a400359` + this hoisting patch | 0 |
| `a8ee8f22c` (#2967 only) | 7 |
| `a8ee8f22c` + this hoisting patch | 7 |

All four `gsc` and `dotnet exec` invocations exited 0. The `0` cell is the silent unwritten-field regression that occurs when hoisting lands without #2967.

## Tests and differential evidence

- `IteratorRewriterTests.Rewrite_HoistsSelectReceiveBinding` directly pins field-map construction.
- `Issue2933AsyncSelectArmBindingTests.SynchronousIteratorReceiveBindingRetainsSelectedArmValueAtRuntime` compiles, loads, and executes a synchronous iterator with three distinct receive-arm outcomes and asserts the selected binding prints `7` after suspension.
- Mutation round trip: head passed with `7`; deleting this PR's `VisitSelectStatement` made the runtime test fail with expected `7`, actual `0`; restoring the visitor passed with `7` again.

Current-head verification at `f26fc97e4087f6b21c46a6e88655ea6e38520d3e`:

- `dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true`: exit 0, 0 warnings, 0 errors
- zero-byte check: 0 failures across 40 `GSharp.Core.dll` files
- propagation check: 0 mismatches across 5 ordinary outputs, MD5 `1c42c3c7e2d43ad20faa980f4f57c203`
- focused structural test: 1 passed
- focused emitted-runtime test: 1 passed
- direct emitted program output: `7`

Earlier full-suite counts were measured on the same `8acc2cc76` base before the emitted-runtime test was added:

| Suite | Base | Head |
| --- | ---: | ---: |
| Core | 7033 passed, 1 skipped | 7034 passed, 1 skipped |
| Compiler | 3988 passed | 3988 passed |
| Interpreter | 512 passed | 512 passed |

No test used `--no-build`. Current validation uses the narrow selectors required by fleet throttle guidance.

Closes #2975.

## Required end-to-end coverage

The reviewer-requested synchronous iterator probes are now pinned by emitted-runtime tests:

- `SynchronousIteratorReceiveBindingSurvivesYieldAtRuntime`: `11 / 33 / 22`
- `SynchronousIteratorReceiveBindingSurvivesLoopedYieldsAtRuntime`: `100 / 101 / 44`
- `NestedSynchronousIteratorReceiveBindingsSurviveYieldsAtRuntime`: `1 / 2 / 66 / 55`

Each test compiles an executable, loads it with `Assembly.Load(File.ReadAllBytes(...))` plus `GetTypes()`, runs it in a bounded child process, and asserts printed values.

Production-hunk mutation, with tests retained:

- clean build rc 0, Core MD5 `1c42c3c7e2d43ad20faa980f4f57c203`: 3 passed
- removed `VisitSelectStatement`, build rc 0, Core MD5 `617c141f45a283100f9110392659f525`: all 3 failed (`33→0`, `44→0`, `66/55→0/0`)
- restored build rc 0, Core MD5 `1c42c3c7e2d43ad20faa980f4f57c203`: 3 passed

The `IsByRefLike` guard now documents that, like the declaration path above it, ByRef-like values cannot become state-machine fields.

### DIFF
--- a/src/Core/CodeAnalysis/Lowering/Iterators/IteratorRewriter.cs
+++ b/src/Core/CodeAnalysis/Lowering/Iterators/IteratorRewriter.cs
@@ -217,6 +217,7 @@ public static class IteratorRewriter
         {
             foreach (var arm in node.Cases)
             {
+                // As above, ByRef-like values cannot be state-machine fields.
                 if (arm.Variable != null
                     && !TypeSymbol.IsByRefLike(arm.Variable.Type)
                     && !Locals.Contains(arm.Variable))
--- a/src/Core/CodeAnalysis/Lowering/Iterators/IteratorRewriter.cs
+++ b/src/Core/CodeAnalysis/Lowering/Iterators/IteratorRewriter.cs
@@ -213,6 +213,21 @@ public static class IteratorRewriter
             base.VisitVariableDeclaration(node);
         }
 
+        protected override void VisitSelectStatement(BoundSelectStatement node)
+        {
+            foreach (var arm in node.Cases)
+            {
+                if (arm.Variable != null
+                    && !TypeSymbol.IsByRefLike(arm.Variable.Type)
+                    && !Locals.Contains(arm.Variable))
+                {
+                    Locals.Add(arm.Variable);
+                }
+            }
+
+            base.VisitSelectStatement(node);
+        }
+
         protected override void VisitAddressOfExpression(BoundAddressOfExpression node)
         {
             if (node.Operand is BoundVariableExpression variable
--- a/test/Compiler.Tests/Emit/Issue2933AsyncSelectArmBindingTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2933AsyncSelectArmBindingTests.cs
@@ -291,6 +291,49 @@ public class Issue2933AsyncSelectArmBindingTests
         Assert.Equal(new[] { 10, 10, 10 }, values);
     }
 
+    [Fact]
+    public void SynchronousIteratorReceiveBindingRetainsSelectedArmValueAtRuntime()
+    {
+        const string Source = """
+            package Issue2975.SyncIterator
+            import System
+            import Gsharp.Extensions.Go
+
+            func Values() sequence[int32] {
+                let first = make(chan int32, 1)
+                let selected = make(chan int32, 1)
+                let third = make(chan int32, 1)
+                selected <- 7
+                select {
+                    case let value = <-first {
+                        yield -100
+                        yield value + 1000
+                    }
+                    case let value = <-selected {
+                        yield 200
+                        yield value
+                    }
+                    case let value = <-third {
+                        yield -300
+                        yield value + 3000
+                    }
+                }
+            }
+
+            for value in Values() {
+                if value != 200 {
+                    Console.WriteLine(value)
+                }
+            }
+            """;
+
+        Assert.Equal(
+            "7\n",
+            CompileLoadAndRun(
+                Source,
+                nameof(SynchronousIteratorReceiveBindingRetainsSelectedArmValueAtRuntime)));
+    }
+
     [Theory]
     [MemberData(nameof(SelectShapes))]
     public void RequestedSelectShapesLoadAndRun(string name, string expectedOutput, string source)
--- a/test/Compiler.Tests/Emit/Issue2933AsyncSelectArmBindingTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2933AsyncSelectArmBindingTests.cs
@@ -292,46 +292,107 @@ public class Issue2933AsyncSelectArmBindingTests
     }
 
     [Fact]
-    public void SynchronousIteratorReceiveBindingRetainsSelectedArmValueAtRuntime()
+    public void SynchronousIteratorReceiveBindingSurvivesYieldAtRuntime()
     {
         const string Source = """
-            package Issue2975.SyncIterator
+            package Issue2975.YieldThenRead
             import System
             import Gsharp.Extensions.Go
 
             func Values() sequence[int32] {
-                let first = make(chan int32, 1)
-                let selected = make(chan int32, 1)
-                let third = make(chan int32, 1)
-                selected <- 7
+                let ch = make(chan int32, 1)
+                ch <- 33
                 select {
-                    case let value = <-first {
-                        yield -100
-                        yield value + 1000
-                    }
-                    case let value = <-selected {
-                        yield 200
+                    case let value = <-ch {
+                        yield 11
                         yield value
-                    }
-                    case let value = <-third {
-                        yield -300
-                        yield value + 3000
+                        yield 22
                     }
                 }
             }
 
             for value in Values() {
-                if value != 200 {
-                    Console.WriteLine(value)
-                }
+                Console.WriteLine(value)
             }
             """;
 
         Assert.Equal(
-            "7\n",
+            "11\n33\n22\n",
             CompileLoadAndRun(
                 Source,
-                nameof(SynchronousIteratorReceiveBindingRetainsSelectedArmValueAtRuntime)));
+                nameof(SynchronousIteratorReceiveBindingSurvivesYieldAtRuntime)));
+    }
+
+    [Fact]
+    public void SynchronousIteratorReceiveBindingSurvivesLoopedYieldsAtRuntime()
+    {
+        const string Source = """
+            package Issue2975.LoopedYields
+            import System
+            import Gsharp.Extensions.Go
+
+            func Values() sequence[int32] {
+                let ch = make(chan int32, 1)
+                ch <- 44
+                select {
+                    case let value = <-ch {
+                        for i in 0 ... 2 {
+                            yield 100 + i
+                        }
+                        yield value
+                    }
+                }
+            }
+
+            for value in Values() {
+                Console.WriteLine(value)
+            }
+            """;
+
+        Assert.Equal(
+            "100\n101\n44\n",
+            CompileLoadAndRun(
+                Source,
+                nameof(SynchronousIteratorReceiveBindingSurvivesLoopedYieldsAtRuntime)));
+    }
+
+    [Fact]
+    public void NestedSynchronousIteratorReceiveBindingsSurviveYieldsAtRuntime()
+    {
+        const string Source = """
+            package Issue2975.NestedSelects
+            import System
+            import Gsharp.Extensions.Go
+
+            func Values() sequence[int32] {
+                let outer = make(chan int32, 1)
+                let inner = make(chan int32, 1)
+                outer <- 55
+                inner <- 66
+                select {
+                    case let outerValue = <-outer {
+                        yield 1
+                        select {
+                            case let innerValue = <-inner {
+                                yield 2
+                                yield innerValue
+                            }
+                        }
+                        yield outerValue
+                    }
+                }
+            }
+
+            for value in Values() {
+                Console.WriteLine(value)
+            }
+            """;
+
+        Assert.Equal(
+            "1\n2\n66\n55\n",
+            CompileLoadAndRun(
+                Source,
+                nameof(NestedSynchronousIteratorReceiveBindingsSurviveYieldsAtRuntime)));
     }
 
     [Theory]
--- a/test/Core.Tests/CodeAnalysis/Lowering/Iterators/IteratorRewriterTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Lowering/Iterators/IteratorRewriterTests.cs
@@ -9,6 +9,7 @@ using GSharp.Core.CodeAnalysis;
 using GSharp.Core.CodeAnalysis.Binding;
 using GSharp.Core.CodeAnalysis.Lowering.Iterators;
 using GSharp.Core.CodeAnalysis.Symbols;
+using GSharp.Core.CodeAnalysis.Syntax;
 using Xunit;
 
 namespace GSharp.Core.Tests.CodeAnalysis.Lowering.Iterators;
@@ -101,6 +102,29 @@ public class IteratorRewriterTests
         // Assert: local should be in hoisted locals
         var plan = Assert.Single(result.Plans);
         Assert.Contains(plan.HoistedLocals, v => v.Name == "temp");
+    }
+
+    [Fact]
+    public void Rewrite_HoistsSelectReceiveBinding()
+    {
+        var seqType = SequenceTypeSymbol.Get(TypeSymbol.Int32);
+        var function = new FunctionSymbol("withSelect", ImmutableArray<ParameterSymbol>.Empty, seqType, package: Package);
+        var selectVariable = new LocalVariableSymbol("value", false, TypeSymbol.Int32);
+        var yield = new BoundYieldStatement(null, new BoundVariableExpression(null, selectVariable));
+        var select = new BoundSelectStatement(
+            null,
+            ImmutableArray.Create(new BoundSelectCase(
+                SelectCaseKind.ReceiveBind,
+                new BoundLiteralExpression(null, null),
+                value: null,
+                selectVariable,
+                Block(yield))));
+        var program = MakeProgram(function, Block(select));
+
+        var result = IteratorRewriter.Rewrite(program);
+
+        var plan = Assert.Single(result.Plans);
+        Assert.Contains(selectVariable, plan.HoistedLocals);
     }
 
     [Fact]


DavidObando:
## BLOCKER — #3013 alone is a silent wrong-value regression. Must not merge before #2967.

Verdict rests on an observed differential from a probe in a scratch clone, not on reading.

### Probe

Sync iterator with a `select` receive arm. Correct answer is `7`:

```gsharp
package IterSelect
import System
import Gsharp.Extensions.Go

func gen() sequence[int32] {
    let ch = make(chan int32, 1)
    ch <- 7
    select {
    case let v = <-ch { yield v }
    }
}

func Main() { for x in gen() { Console.WriteLine(x) } }
```

### Measured — all four configs built from source, `gsc` rc=0, `dotnet exec` rc=0

| config | output | |
|---|---|---|
| `main` `f70626452` | **7** | correct |
| `main` + **#3013** | **0** | **WRONG, silently** |
| `main` + #2967 | **7** | correct |
| `main` + #3013 + #2967 | **7** | correct |

Every run exits **0**. No diagnostic, no crash, no IL-verify failure. Pure silent divergence — the worst shape we have.

### Mechanism

`IteratorRewriter.cs:187` hoists the select-arm variable into the state-machine field set. Nothing then *stores* the received value into that field: `MethodBodyEmitter.Statements.cs` still emits `TryRead` into the local out-slot and calls `EmitStatement(arm.Body)`, while the rewritten body now reads the **hoisted field**, which was never written. Result is `default(T)`.

The store is exactly what **#2967** adds (`EmitSelectReceiveArmBody`), and note its guard already covers iterators, not just async:

```csharp
|| (this.iteratorEmitCtx != null && this.iteratorEmitCtx.FieldMap.TryGetValue(arm.Variable, out hoistedField))
```

So #3013 and #2967 are **two halves of one mechanism**. #2967 is a safe no-op for iterators until #3013 lands (measured above: 7).

### Why no gate caught this

The two PRs share **no file** — `Lowering/Iterators/IteratorRewriter.cs` vs `Emit/MethodBodyEmitter.Statements.cs`. `merge-tree` is clean. This is the #3000 shape: textual mergeability is not semantic mergeability.

And #3013's only test is a **rewriter unit test**:

`test/Core.Tests/CodeAnalysis/Lowering/Iterators/IteratorRewriterTests.cs:107` — `Assert.Contains(selectVariable, plan.HoistedLocals)`

It asserts the variable *is hoisted*, which is the exact half that is broken in isolation. It can never observe the emitted value.

### Required

1. **Merge order: #2967 before #3013.** Reverse order lands a silent regression.
2. **#3013 needs a compile-and-run test**, not a rewriter assertion — one that yields the received value from a sync iterator and asserts `7`. The probe above is ready to lift.
3. **#3013 is 9 commits behind `main`** (merge-base `a6b1561f2`, tip `f70626452`), so its CI never saw the current tree. Rebase before trusting green.

### Note

Not blaming #2967 — measured safe standalone and safe combined.

DavidObando:
## Review: MERGEABLE — both the safety and the purpose question answered by measurement

Measured tree: scratch clone at `main` = `fd34166a8`, merged with `refs/pull/3013/head` = `634f3032` (three-dot base `a6b1561f2`; the pushed head is **not** rebased onto post-#2967 `main`, but `git merge` is clean and I measured the merge result, not the branch).

Both questions this PR was held for are now answered, and the answers are the good ones.

### 1. Safety — the matrix's live half is intact

The shape that #3013-alone zeroed now behaves. Post-#2967 `main` plus this PR:

```gsharp
func genA() IEnumerable[int32] {
    let ch = make(chan int32, 1)
    ch <- 7
    select {
    case let v = <-ch { yield v }
    }
}
```

`A=7` on both `main` and `main`+#3013. **No regression** — #2967 supplies the store, exactly as expected.

### 2. Purpose — `main` is silently wrong today, in four distinct slots

This is the part worth reading. #3013 is **not** overtaken by #2967. Post-#2967 `main` still produces wrong values whenever a select-arm binding is read *after* a `yield`, because `Lowering/Iterators/` still has zero `BoundSelectCase`/`SelectStatement` handling — the store exists, but there is no field to store into.

Compiled with `gsc`, run with `dotnet`, distinct values per slot so mis-mapping cannot hide:

| probe | shape | `main` `fd34166a8` | `main`+#3013 |
|---|---|---|---|
| A | bind, `yield v` immediately | `7` | `7` |
| B | bind, `yield 11`, then `yield v` | `11` / **`0`** / `22` | `11` / **`33`** / `22` |
| C | bind, loop of yields, then `yield v` | `100` / `101` / **`0`** | `100` / `101` / **`44`** |
| D | nested selects, both read across yields | `1` / `2` / **`0`** / **`0`** | `1` / `2` / **`66`** / **`55`** |

Every wrong case exits **0** with no diagnostic — silent wrong value, the worst class. Four slots go from `0` to the correct distinct value. That is a real, observable fix.

### 3. Mutation — the hunk is load-bearing and pinned

Single production hunk, reverted in isolation (`git checkout main -- IteratorRewriter.cs`, test kept):

```
clean    md5 GSharp.Core.dll  067833351ccf5cdc24b2f109336e407e   19 passed
mutant   build rc=0           7398729d4a5abb720571460995f5bcee   1 FAILED, 18 passed
                                                                  -> Rewrite_HoistsSelectReceiveBinding
restored md5                  067833351ccf5cdc24b2f109336e407e   19 passed
```

Build exit checked as `0` **before** trusting either hash, and the md5 round-trip closes exactly. Killed by its own named test.

---

## SHOULD-FIX — the only test is a plan assertion, which is precisely the assertion class that cannot see this PR's known failure mode

`test/Core.Tests/CodeAnalysis/Lowering/Iterators/IteratorRewriterTests.cs:107` asserts `plan.HoistedLocals` contains the variable. That is the *input* to emission, not the behaviour.

This matters more here than it usually would, because this PR already has a documented near-miss of exactly that shape: on pre-#2967 `main`, `HoistedLocals` was **correct** and the program still printed `0`, because the matching store was missing. A plan assertion passes in that world. So the one test in this PR would have stayed green through the precise defect that made #3013 dangerous, and will stay green again if the store side ever regresses.

The fix is cheap — the four probes above are ready-made, and A/B are enough. An end-to-end test asserting `11/33/22` costs one `[Fact]` and converts this PR's guarantee from "the plan is right" to "the program is right". Given that the *only* reason this PR was safe to land is a change in a different file that no test in this PR references, I would not ship the pairing unpinned.

## NIT — the guard is right but silent

`IteratorRewriter.cs:190` reuses `!TypeSymbol.IsByRefLike(arm.Variable.Type)`, correctly mirroring the sibling at `VisitVariableDeclaration`. But the sibling carries an eight-line comment explaining *why* (ByRef-like types cannot be instance fields; issue #1919's metadata-load-safe helper). The new copy has none, so the next reader cannot tell whether the guard is load-bearing or cargo. One line referring to the sibling would do it.

## Two things I checked and am explicitly **not** reporting as findings

- **The issue names a different file.** #2975 asks for `HoistedFieldRewriter.cs`; the PR changes `IteratorRewriter.cs`. Not a defect — `CollectHoistedLocals` at `:147` feeds `plan.HoistedLocals`, which is what `StateMachineEmitter.cs:547` and `:806` turn into fields. Fixing it at collection is upstream of the issue's suggestion and strictly smaller. The measured differential confirms it reaches emission.
- **The async iterator path is unexamined, not clean.** `AsyncIteratorRewriter.cs:84` hoists via `AsyncCaptureWalker`, a different mechanism. `SelectStatement`/`SelectCase` return 0 hits there — but `VisitVariableDeclaration` also returns 0 hits, so that walker simply does not work by visit-override and **a zero there is scope, not absence**. I am not claiming an async gap; I am saying this PR's evidence does not speak to it, and #2967/#2933 covered that side.

---

**Verdict: MERGEABLE.** Safety confirmed, purpose confirmed with a four-slot observable differential, mutation killed by a named test, md5 round-trip closed, build exit checked. The SHOULD-FIX is about what protects the fix tomorrow, not whether it works today.

DavidObando:
# Review: MERGEABLE

Anchors. `main` **`185431c4`**. Head **moved during review**: `057dcd2e` → **`f26fc97e`** (rebase). Merge base now `185431c4`. Three-dot scope unchanged: **3 files, +144 −0**.

**Rebase was content-neutral, and this is measured, not assumed:**
```
new head tree   f0dc40913d224dc2a809d9124f0a50df8073b0b6
my merged tree  f0dc40913d224dc2a809d9124f0a50df8073b0b6
git diff merged3013 f26fc97e -> empty
```
So every number below was taken on a tree **byte-identical to the current head**, which is also `main`+PR. No rework needed.

## (a) Does every probe fail at base? YES — none vacuous

Failure class is as predicted: **rc=0, silently wrong value** (`0` = `default(T)`), never a diagnostic. Base = production hunk removed (== pre-PR).

| probe | shape | base | fixed |
|---|---|---|---|
| p1 | binding read across `yield`, in-function | **`11 0`** | `11 77` |
| p2 | two arms, different names | **`22 0`** | `22 88` |
| p5 | `select` in a loop, repeated resume | **`100 0 100 0 100 0`** | `100 11 100 22 100 33` |
| p4 | arm binding never used | `11 22` | `11 22` — no crash |
| p6 | `default` arm | `42` | `42` |
| p7 | top-level `select`, no iterator | `11 77` | `11 77` |
| p8 | **same name bound in two arms** | not measured | `22 88` ✅ |
| pa | **async** iterator across `yield` | **`11 77` correct at base** | `11 77` |

Distinct values throughout (`11/22/33`, `77`, `88`), never `1/1/1`.

## (b) Per-hunk kill/survive: 1 hunk, 4 killers, 0 survivals

Only one production hunk exists — `VisitSelectStatement`, `IteratorRewriter.cs:216-231` — so "mutant" ≡ "base".

| test | result at base |
|---|---|
| `IteratorRewriterTests.Rewrite_HoistsSelectReceiveBinding` | FAIL (1 failed / 19) |
| `SynchronousIteratorReceiveBindingSurvivesYieldAtRuntime` | FAIL `"11\n33\n22\n"` → `"11\n0\n22\n"` |
| `NestedSynchronousIteratorReceiveBindingsSurviveYieldsAtRuntime` | FAIL `"1\n2\n66\n55\n"` → `"1\n2\n0\n0\n"` |
| `SynchronousIteratorReceiveBindingSurvivesLoopedYieldsAtRuntime` | FAIL `"100\n101\n44\n"` → `"100\n101\n0\n"` |

Compiler.Tests class under mutant: **3 failed / 11 passed / 14**. Restore → md5 round-tripped **exactly** to `3236780fd39845787655864d8776ad7d`, `git diff` empty, control `IteratorRewriterTests` **19/19**, class **14/14**.

## (c) Site denominator: 1 of 1 — complete

`BoundSelectCase` (`BoundSelectCase.cs:45`) carries **exactly one** `VariableSymbol Variable`. No comma-ok second binding exists. `LocalCollector` is `private sealed`, instantiated once at `IteratorRewriter.cs:151`. Denominator inside the construct is **1 of 1**, not a sample.

## (d) "Synchronous" scoping is HONEST — measured, not merely asserted

`pa.gs` (binding read after a `yield` inside `async func … async sequence[int32]`) prints **`11 77` at base**. The async state machine already handles this via #2967. Scoping the fix to the sync path is correct, not half a fix.

Worth recording: the **pre-existing async tests yield *inside* the arm**, so they never exercised the across-suspension case. `pa.gs` does, and it passes — the async path is genuinely covered, just not by those tests.

## (e) Four-column gsc × gsi

`gsi` is **correct at base on every probe**. The interpreter does not hoist and does not need to. So this fix removes a **`gsc`/`gsi` divergence**; all four cells agree on the fixed tree.

## Suites — unfiltered, on the head-identical merged tree

- **Compiler.Tests: 4032 passed / 0 failed / 0 skipped / 4032**, rc=0, 1h19m
- **Core.Tests: 7068 passed / 0 failed / 1 skipped / 7069**, rc=0
- Docs gate on merged tree: **383 / 381**, identical to `main`; PR touches neither file.
- `cs2gs`: passed on old head (5m29s). Corpus is C# input (239 `.cs`; positive control 239/239 contain `class`); C# has no channel `select`, so `BoundSelectStatement` cannot arise → the new override cannot fire on the corpus.

---

# SHOULD-FIX (body only, no code defect)

**1. The body names a test that does not exist.**
`SynchronousIteratorReceiveBindingRetainsSelectedArmValueAtRuntime` — **0 matches** in the tree. Positive control: `SurvivesYieldAtRuntime` → **2 matches**, so the zero is absence, not scope. The body attributes mutation evidence to this name ("deleting this PR's `VisitSelectStatement` made the runtime test fail with expected `7`, actual `0`"). The four real tests are `…SurvivesYieldAtRuntime`, `…SurvivesLoopedYieldsAtRuntime`, `Nested…SurviveYieldsAtRuntime`, `Rewrite_HoistsSelectReceiveBinding`. Fix before squash — the later "Required end-to-end coverage" section is accurate and independently reproduced; the top half is stale.

# NIT

**2. Four-cell table is an ordering probe, not fix evidence.** Three of four cells print `7`, including "neither fix". That probe consumes the binding inside the arm, so it never exercises #2975. It is honest *for its stated purpose* (showing hoisting-without-#2967 is harmful), but it sits in the top half of the body where it reads as fix evidence. Also now **historical**: #2967 `a8ee8f22c` is an ancestor of both `main` and this head, so the `0` cell is unreachable.

**3. Stale suite table.** Counts are labelled as measured on base `8acc2cc76`; `main` is `185431c4`. Numbers above supersede.

**4. CI is incomplete on the rebased head.** The green cited (8 SUCCESS / 2 SKIPPED) was `057dcd2e`. On `f26fc97e`, `build` is still `in_progress` and `cs2gs`/`e2e` have not started. The 2 skips were `publish` / `publish-visual-studio-extension` — benign; **`cs2gs` did run and pass**. Tree is head-identical to what I ran locally, so risk is low, but wait for the run.

**5. Pre-existing, out of scope:** `yield` inside `catch` ICEs — `GS9998: InvalidOperationException: Label 4 has not been marked.` My `b2.gs` contains **no `select`**, so this PR's visitor cannot fire; it is pre-existing by construction. Worth its own issue.

# Disclosures

- **p8 (same name in two arms) fail-at-base was NOT measured.** Fixed-tree result `22 88` is real; I did not re-mutate because a rebuild would have overwritten the in-flight full Compiler.Tests binaries and voided that run. p2 (`22 0`) is the near-identical shape and did fail at base.
- Hashes come from dedicated `-p:ContinuousIntegrationBuild=true` builds. The later `dotnet test` runs break md5 comparability; no mutation followed them, so nothing depends on it.
- Honest scope note: extra probe shapes are **shape documentation, not extra pinning** — with a single production hunk, "fails at base" and "kills the mutant" are one measurement, so the 4 tests pin one hunk 4 times over.

DavidObando:
# Addendum — closes my own open disclosure, adds 3 axes. Verdict unchanged: MERGEABLE

Prompted by Fleet Bulletin #3 (`func Main()` blind spot). Head still `f26fc97e`, `main` still `185431c4`, tree measured byte-identical to head (`f0dc4091`).

## 1. The `func Main()` axis — clean, both sides

- **PR's added tests: 0 uses of `func Main()`** (positive control: 3 `package` matches). All three runtime tests drive at **file scope**: `for value in Values() { Console.WriteLine(value) }`.
- **My 10 probes: 0 uses of `func Main()`**, and every one also ran under **`gsi`**, which only executes top-level code — so the interpreter column was real, not nominal.

Neither author nor reviewer inherited the shape that let #2963's source break ship green.

**Bulletin's grep target is absent:** the hunk has **no symbol-class test at all** — only `arm.Variable != null`, `TypeSymbol.IsByRefLike`, and a dedup. There is no `is not LocalVariableSymbol` first-disjunct to strand globals.

## 2. Three new axes, now measured both ways

`IteratorRewriter.cs:216-231` removed = base; positive control on the same build: p1 `11 0`, p2 `22 0`, p5 `100 0 100 0 100 0`; mutant md5 `9c692f03b36d2dfd493d09a0f4b11fdf`, `BUILD_RC=0`, 0 warnings.

| probe | axis | BASE | HEAD `gsc` | HEAD `gsi` |
|---|---|---|---|---|
| `pc` | iterator declared as a **class method** | **`11 0`** | `11 55` | `11 55` |
| `p8` | **same name** bound in two arms | **`22 0`** | `22 88` | `22 88` |
| `pg` | arm binding stored to a **global**, read across resume | **`11 66 66`** | `11 132 66` | `11 132 66` |
| `p1` | control | **`11 0`** | `11 77` | `11 77` |

All fail at base, all correct at head, `gsc` = `gsi`. **This closes the one gap my first comment disclosed** ("p8 fail-at-base was NOT measured" — it is now measured, and it fails). No survivals anywhere.

Restored: `BUILD_RC=0`, md5 round-tripped exactly to `3236780fd39845787655864d8776ad7d`, `git status` clean, 0 diff vs `f26fc97e`.

## 3. Measurement error I made and caught — new trap worth naming

My first pass at these three probes reported them as **SURVIVING** the mutant (`11 55`, `22 88`, `11 132 66` at base). **That was false.**

Cause: I compiled each base probe into **`/tmp/g3013/`, which already contained a HEAD-built DLL of the same assembly identity** from the earlier head run. The freshly built base program bound the stale head assembly. `pc3`/`pc4`/`pc5`/`pg2` — variants with no pre-existing DLL — correctly failed at base, which is what exposed the inconsistency: `pc` and `pc5` differ **only in the package name** yet disagreed at base.

Re-run in **isolated one-file directories**, all four fail at base. That is the authoritative reading above.

> **New trap: an output directory holding a prior build of the same assembly identity can silently serve the old assembly. Emit each probe into its own empty directory.** This is a third staleness class, distinct from the stale-driver and `ref/` cases — and note the artifact pointed the *dangerous* way: it made a live mutant look survived.

Nothing else in my first comment changes.

PR #3014, opened 2026-08-01
Fix IL verifier probe deadlocks and scoped suppressions

### LINKED ISSUE #3007: Compiler.Tests: IlVerifier tool probes still use unbounded sequential ReadToEnd

### LINKED ISSUE #3008: Compiler.Tests: fixed/pointer emit gate suppresses ilverify categories assembly-wide and scans IL by raw byte index
Two review findings from PR #2964 (merged as `5e616b9bb`) were raised but not applied. Both are in the *test gate*, not the emitter — #2964's production fix is correct and I verified it (base `177895195` → `1 Error(s) ExpectedNumericType @0x1F`; head → `All Classes and Methods Verified`; both run). Filing so they are not lost.

Neither is urgent. Both weaken the ability of the `fixed`/pointer emit gate to catch a *future* regression.

## 1. `ignoredErrorCodes` is an assembly-wide category blanket, and the blanket just got broader

`test/Compiler.Tests/IlVerifier.cs:117-118` forwards each code to ilverify's `-g`, which matches the **error category name across the whole assembly** — not a site, not a method, not an occurrence count.

`docs/adr/0125-fixed-pinned-statement.md:187-189` correctly records that pin + dereference IL is unverifiable by design, as in C#. That is not the problem. The problem is that the *suppression mechanism* is far coarser than the documented exemption:

- `test/Compiler.Tests/Emit/Issue2906ExhaustiveSwitchReturnEmitTests.cs:314` moved from `["ExpectedNumericType"]` to `["UnmanagedPointer", "StackByRef"]`.
- That pair is the signature of **any** pointer dereference by any compiler — `csc` on a bare `static unsafe int M(int* p) => *p`, with no `fixed` at all, emits exactly `UnmanagedPointer @0x00` + `StackByRef @0x01`.

So that test now passes even if the emitter later produces a genuinely malformed dereference anywhere in its program. Same for `Issue2922FixedSwitchIlTests.cs:316` (`["StackUnexpected"]`), which covers a real residual — the span unpin `ldc.i4.0; conv.u; stloc` into a `pinned int32&` local, `StackUnexpected @0x45`, which is Roslyn's own idiom and irreducible — but suppresses that category program-wide to do it.

This matters more now than before #2964, because the exempt set in the ADR is 5+ codes and the natural next step is to append to it.

**Suggested fix:** make the exemption positional rather than categorical — assert the expected error **count** (and ideally offset/method) rather than filtering the category to zero. A test that expects "exactly one `StackUnexpected`, at the unpin" still fails when a second, real one appears.

## 2. Byte-scanning IL by index is unsound in two independent ways

`test/Compiler.Tests/Emit/Issue2922FixedSwitchIlTests.cs:361-366`:

```csharp
var loadElementAddress = Array.IndexOf(il, unchecked((byte)OpCodes.Ldelema.Value));
var callOffset = loadElementAddress + 5;
Assert.Equal(unchecked((byte)OpCodes.Call.Value), il[callOffset]);
```

- `Array.IndexOf` over the raw body can match an **operand** byte rather than an opcode — `0x8F` occurring inside a token, branch displacement or `ldc.i4` operand hits first and the rest of the assertion is measured from a meaningless offset.
- `+ 5` hard-codes the instruction width, so one inserted instruction or an encoding change breaks it.

Failure mode is opaque: against an experimental variant this produced `Assert.Equal() Failure: Expected 40 Actual 224`, which names neither the opcode nor the instruction. Decode the stream instead — the file already imports `System.Reflection.Emit`.

## Files a fix would touch

- `test/Compiler.Tests/Emit/Issue2922FixedSwitchIlTests.cs` (`:316`, `:355-367`)
- `test/Compiler.Tests/Emit/Issue2906ExhaustiveSwitchReturnEmitTests.cs` (`:314`)
- `test/Compiler.Tests/IlVerifier.cs` (`:104-119`) — only if the count-based assertion is added to the shared helper rather than to each call site

No production files. No ADR change needed; `0125:187-189` is already accurate.

DavidObando:
## Summary

- Route both ilverify discovery probes through the existing bounded process runner, which starts concurrent async reads before its timeout/kill path.
- Add method-scoped ilverify suppressions by verifying the rest of the assembly without ignores, then the selected method with its documented ignores.
- Scope the fixed/pointer exemptions in the #2906 and #2922 tests to `<Program>.F`.
- Replace raw IL byte scanning with a shared opcode-aware instruction reader and add an operand-collision regression.
- Correct the stale #2966-era test comment: a drain that blocks before the bounded wait can hang; starting a drain after the wait instead reaches the timeout and fails boundedly.

## Evidence

- Post-rebase targeted validation, exit 0: 29 passed, 0 failed.
  - `IlVerifierTests`
  - `Issue2900FixedPinReleaseEmitTests`
  - `Issue2922FixedSwitchIlTests`
  - `Issue2906ExhaustiveSwitchReturnEmitTests.ExhaustiveEnum_UnnamedValueWithFixedReturn_UsesFixedEpilogueDefault`
  - Command used `-p:ContinuousIntegrationBuild=true` and did not use `--no-build`.
- #3007 perturbation: replaced `TryProbe` with sequential `ReadToEnd()` calls. A bounded runner timed out after 60 seconds and exited 124. Restoring the concurrent bounded path made the same test pass (1/1, exit 0, 162 ms).
- #3008 suppression perturbation: replaced the two-pass method scope with the old assembly-wide ignore. The regression failed with `Assert.Throws() Failure: No exception was thrown` (exit 1), proving the second method's genuine dereference fault was hidden. Restoring method scope passed (1/1, exit 0).
- Opcode regression uses IL where raw `Array.IndexOf` finds `0x8F` at index 1 inside an `ldc.i4` operand; the decoder correctly returns only `ldc.i4` and `ret` instructions.
- Pointer caller tests also load assemblies with `Assembly.Load(...)+GetTypes()` and execute through bounded child/task paths.
- `git log -S` traces the corrected sentence to `87ca66e85` (#2966), where the wait became bounded and made the inherited wording stale.

SDK/e2e tests were not run because this task explicitly prohibits touching the shared `gsharp.net.sdk` NuGet cache.

Fixes #3007
Fixes #3008

### DIFF
--- a/test/Compiler.Tests/Emit/Issue2900FixedPinReleaseEmitTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2900FixedPinReleaseEmitTests.cs
@@ -3,11 +3,9 @@
 // </copyright>
 
 using System;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
-using System.Reflection;
 using System.Reflection.Emit;
 using System.Reflection.Metadata;
 using System.Reflection.PortableExecutable;
@@ -32,12 +30,6 @@ public class Issue2900FixedPinReleaseEmitTests
         "StackUnexpectedArrayType",
         "ExpectedNumericType",
     };
-
-    private static readonly Dictionary<short, OpCode> OpCodesByValue = typeof(OpCodes)
-        .GetFields(BindingFlags.Public | BindingFlags.Static)
-        .Where(field => field.FieldType == typeof(OpCode))
-        .Select(field => (OpCode)field.GetValue(null)!)
-        .ToDictionary(opcode => opcode.Value);
 
     [Fact]
     public void BranchExits_UseLeaveThroughFinally_AndRun()
@@ -951,74 +943,14 @@ public class Issue2900FixedPinReleaseEmitTests
         var name = metadata.GetString(definition.Name);
         if (definition.RelativeVirtualAddress == 0)
         {
-            return new MethodIl(name, Array.Empty<Instruction>(), Array.Empty<ExceptionRegion>());
+            return new MethodIl(name, Array.Empty<IlInstruction>(), Array.Empty<ExceptionRegion>());
         }
 
         var body = pe.GetMethodBody(definition.RelativeVirtualAddress);
         return new MethodIl(
             name,
-            Decode(body.GetILBytes() ?? Array.Empty<byte>()),
+            IlInstructionReader.Read(body.GetILBytes() ?? Array.Empty<byte>()),
             body.ExceptionRegions.ToArray());
-    }
-
-    private static Instruction[] Decode(byte[] il)
-    {
-        var instructions = new List<Instruction>();
-        var offset = 0;
-        while (offset < il.Length)
-        {
-            var instructionOffset = offset;
-            short value = il[offset++] == 0xFE
-                ? unchecked((short)(0xFE00 | il[offset++]))
-                : unchecked((short)il[instructionOffset]);
-            var opcode = OpCodesByValue[value];
-            int? branchTarget = null;
-
-            switch (opcode.OperandType)
-            {
-                case OperandType.InlineNone:
-                    break;
-                case OperandType.ShortInlineBrTarget:
-                    branchTarget = offset + 1 + unchecked((sbyte)il[offset]);
-                    offset++;
-                    break;
-                case OperandType.InlineBrTarget:
-                    branchTarget = offset + 4 + BitConverter.ToInt32(il, offset);
-                    offset += 4;
-                    break;
-                case OperandType.InlineSwitch:
-                    var count = BitConverter.ToInt32(il, offset);
-                    offset += 4 + (count * 4);
-                    break;
-                case OperandType.ShortInlineI:
-                case OperandType.ShortInlineVar:
-                    offset++;
-                    break;
-                case OperandType.InlineVar:
-                    offset += 2;
-                    break;
-                case OperandType.InlineI:
-                case OperandType.ShortInlineR:
-                case OperandType.InlineField:
-                case OperandType.InlineMethod:
-                case OperandType.InlineSig:
-                case OperandType.InlineString:
-                case OperandType.InlineTok:
-                case OperandType.InlineType:
-                    offset += 4;
-                    break;
-                case OperandType.InlineI8:
-                case OperandType.InlineR:
-                    offset += 8;
-                    break;
-                default:
-                    throw new InvalidOperationException($"Unsupported IL operand type {opcode.OperandType}.");
-            }
-
-            instructions.Add(new Instruction(instructionOffset, opcode, branchTarget));
-        }
-
-        return instructions.ToArray();
     }
 
     private sealed class CompiledProgram : IDisposable
@@ -1093,11 +1025,6 @@ public class Issue2900FixedPinReleaseEmitTests
 
     private sealed record MethodIl(
         string Name,
-        Instruction[] Instructions,
+        IlInstruction[] Instructions,
         ExceptionRegion[] Regions);
-
-    private sealed record Instruction(
-        int Offset,
-        OpCode OpCode,
-        int? BranchTarget);
 }
--- a/test/Compiler.Tests/Emit/Issue2906ExhaustiveSwitchReturnEmitTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2906ExhaustiveSwitchReturnEmitTests.cs
@@ -311,7 +311,8 @@ public class Issue2906ExhaustiveSwitchReturnEmitTests
         var assembly = CompileVerifyLoadAndRun(
             "UnnamedFixed",
             Source,
-            ignoredVerificationErrors: new[] { "UnmanagedPointer", "StackByRef" });
+            ignoredVerificationErrors: new[] { "UnmanagedPointer", "StackByRef" },
+            ignoredVerificationScope: @"<Program>\.F$");
         // Known limitation (#2953): the fixed-return epilogue must be emitted first,
         // so fixed-containing non-void functions bypass the fall-through guard and
         // silently return the default slot instead of throwing.
@@ -324,7 +325,8 @@ public class Issue2906ExhaustiveSwitchReturnEmitTests
     private static Assembly CompileVerifyLoadAndRun(
         string name,
         string source,
-        IEnumerable<string> ignoredVerificationErrors = null)
+        IEnumerable<string> ignoredVerificationErrors = null,
+        string ignoredVerificationScope = null)
     {
         using var peStream = new MemoryStream();
         var result = new Compilation(SyntaxTree.Parse(SourceText.From(source))).Emit(peStream);
@@ -337,7 +339,10 @@ public class Issue2906ExhaustiveSwitchReturnEmitTests
         try
         {
             File.WriteAllBytes(assemblyPath, bytes);
-            IlVerifier.Verify(assemblyPath, ignoredErrorCodes: ignoredVerificationErrors);
+            IlVerifier.Verify(
+                assemblyPath,
+                ignoredErrorCodes: ignoredVerificationErrors,
+                ignoredErrorScope: ignoredVerificationScope);
         }
         finally
         {
--- a/test/Compiler.Tests/Emit/Issue2922FixedSwitchIlTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2922FixedSwitchIlTests.cs
@@ -329,7 +329,8 @@ public class Issue2922FixedSwitchIlTests
         IlVerifier.Verify(
             program.AssemblyPath,
             additionalReferences: null,
-            ignoredErrorCodes: ignoredVerificationErrors);
+            ignoredErrorCodes: ignoredVerificationErrors,
+            ignoredErrorScope: ignoredVerificationErrors.Length == 0 ? null : @"<Program>\.F$");
         program.AssertLoadable();
         Assert.Equal(expectedOutput, program.Run());
     }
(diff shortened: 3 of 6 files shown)


DavidObando:
## Verdict: MERGEABLE — no blockers

Reviewed at head `64927e92a`. Computed `git merge-base f70626452 HEAD` = `f70626452` exactly, so the PR sits directly on the current tip; no rebase, and three-dot == two-dot here. Net **6 files / +267 −120**, and **0 of them are under `src/`** — that single fact bounds a lot of the risk below.

Release build with `-p:ContinuousIntegrationBuild=true`: **exit 0, 0 warnings, 0 errors**.

### Claim 2 (method-scoped suppressions) is GENUINE and load-bearing

This was the claim worth establishing, since `ignoredErrorCodes` feeds ilverify `-g`, which matches a **category name assembly-wide**. Verified by construction, not by CI green — three experiments on the shipped test `Verify_MethodScopedIgnore_DoesNotHideSameErrorInAnotherMethod`:

| # | perturbation | result | what it excludes |
|---|---|---|---|
| E1 | none (as shipped) | PASS | baseline |
| E3 | add `Assert.DoesNotContain("::Ignored(")` | PASS | excludes "the `-e` regex matched nothing"; proves the scoped method really was excluded from run 1 |
| E2 | `ignoredErrorScope: null` (i.e. old blanket behaviour) | **FAIL** | proves the scoping is load-bearing, not cosmetic |

E2 is the decisive one: with the scope removed, `Assert.Throws` receives **no exception at all** — the genuine violation in the *other* method is swallowed. That is a direct empirical demonstration that the old blanket `-g` really did hide real errors assembly-wide, which is the defect this PR is fixing.

Mechanism at `test/Compiler.Tests/IlVerifier.cs:97-127` is sound: run 1 = whole assembly, `-e <scope>`, **no** ignores; run 2 = `-i <scope>` **with** ignores. It has a good emergent safety property — if the scope regex ever matches nothing, run 1 verifies everything unsuppressed and fails **loudly** rather than silently going permissive.

I confirmed that loudness is real rather than theoretical. Perturbing the scope regex at `Issue2922FixedSwitchIlTests.cs:333` from `<Program>\.F$` to a non-matching `<NoSuchType>\.Zzz$`:

```
md5 e91ae618 (7/7 PASS)  ->  md5 fbd9a2db (1 FAIL)  ->  md5 e91ae618 (7/7 PASS)
```

and exactly **1 of 7** cases flipped — precisely the one case with a non-empty ignore list (`["StackUnexpected"]`); the other 6 pass `null` through the ternary and take the single-run path. Denominators match the mechanism.

### Claim 1 — real, and the trap was avoided

`TryProbeDotnetToolRun` and `TryProbeOnPath` both now route through a new `TryProbe` → `RunProcess`. **`RunProcess` itself is not modified**: across the entire diff exactly one changed line mentions it, and it is a call site (`+ var (exitCode, _, _) = RunProcess(`). It already used the correct concurrent-async pattern and was correctly left alone.

### Claim 4 — an extraction, not a fresh 90-line decoder

`IlInstructionReader.cs` is not new logic. `Issue2900FixedPinReleaseEmitTests.cs` already carried this exact decoder; the PR moves it out and deletes the duplicate (that is most of the −79). I diffed the operand-length table against the deleted copy: **identical**, including the `0xFE` prefix, `InlineSwitch` (`4 + count*4`), `ShortInlineR`=4, `InlineR`=8, and the branch-target arithmetic. The only delta is `InlineMethod` split out of the 4-byte group into its own case so it can capture the token — still `offset += 4`, so lengths are preserved exactly. Risk of silently corrupting the Issue2900 assertions: nil.

The motivating "operand collision" is real and the regression test is well aimed. The old `Issue2922` code did `Array.IndexOf(il, (byte)OpCodes.Ldelema.Value)` — a raw byte scan that cannot tell an opcode from an operand byte. The new test constructs `ldc.i4` whose 4-byte operand *begins* with `Ldelema`'s opcode byte and asserts `Array.IndexOf` finds a false positive at index 1, while the reader correctly yields `[Ldc_I4, Ret]`.

I checked the one gap I expected to matter — `InlineSwitch` records no `BranchTarget` — and it is **unreachable**. All 4 `BranchTarget` consumers are leave-out-of-handler checks in `Issue2900`, and `switch` cannot cross a protected-region boundary in valid IL. It is also not a regression: the deleted decoder behaved identically.

### Does it close anything it does not declare?

No. It declares `Fixes #3007` / `Fixes #3008` and touches the #2900/#2906/#2922 test files, but **zero `src/` files**, so it cannot change compiler behaviour and cannot resolve those compiler issues. All 36 tests across those three suites pass, and 10/10 `IlVerifierTests`. Nothing was re-scoped to keep CI green: the two production scope adoptions each sit on a case that genuinely needs its ignores, as E2/the regex perturbation demonstrate.

---

## SHOULD-FIX

**S1 — `test/Compiler.Tests/IlVerifierTests.cs:154`+ — the scope test does not guard the failure mode it exists to prevent.**
`Verify_MethodScopedIgnore_DoesNotHideSameErrorInAnotherMethod` asserts only `Assert.Contains("::Visible(", ...)`. That assertion also passes in the bad world where the `-e` regex matches nothing and run 1 throws naming *both* methods. My E3 shows that is not happening today, but the test does not *pin* it, so a future typo in the regex degrades to blanket verification silently. Fix is free and I verified it passes: add

```csharp
Assert.DoesNotContain("::Ignored(", exception.Message, StringComparison.Ordinal);
```

**S2 — adoption is 3 of 29, so this does not strengthen "every IL verdict".**
`ignoredErrorCodes` has **29** call sites; `ignoredErrorScope` has **3** (2 production + 1 self-test). The mechanism is correct, but **26 sites remain blanket-suppressed** and retain the exact blind spot E2 just proved is real. Any verdict resting on those 26 is worth no more today than it was yesterday. Worth either migrating them or recording the residual explicitly so the improvement is not over-read.

**S3 — the suppression is still category-wide *within* the scoped method.**
Run 2 applies `-g CATEGORY` to the included method, so a second, genuine `StackUnexpected` inside `<Program>.F` is still hidden. The blast radius shrinks from assembly to method, which is a large and real gain, but it is not instance-level and should not be described as if it were.

**S4 — `test/Compiler.Tests/IlVerifierTests.cs:279` — undeclared bound loosening.**
`RunProcess_EscapedDescendantHoldingPipe_TimeoutRemainsBounded` goes from `< 15s` to `< 20s`. This is in none of the five claims. Worst case is ~2s timeout + 10s `KillGraceMilliseconds` = 12s, so 15s had 3s of headroom and 20s has 8s — defensible under contention, but it should be stated rather than ridden in.

## NIT

**N1 — claim 5's committed sentence is true, but the PR body describes a sentence that was not committed.**
Committed at `IlVerifierTests.cs:223`: *"A regression that blocks in either drain before WaitForExit hangs this test."* That is **true** — a synchronous blocking read before the wait deadlocks unboundedly and the timeout is never reached. The PR body promises a two-clause replacement whose second clause ("starting a drain *after* the wait instead reaches the timeout and fails boundedly") is absent from the code. The second clause is also true — that path is a deterministic ~10s failure via `WaitForExit(10_000)` → `Kill` → bounded `Grab` → `XunitException` — and it is the half that documents why the *old* sentence was wrong, so it is the more useful half to have kept.

---

All perturbations were restored and flip-back confirmed; `git status --porcelain` is empty.

PR #3016, opened 2026-08-01
Fix delegate property invocation through interfaces

### LINKED ISSUE #2925: Delegate-typed interface property cannot be invoked

DavidObando:
## Summary
- generalize the existing source delegate-member invocation helper to accept both struct/class and interface receiver symbols
- route both invocation fallback sites through the generalized helper
- cover class and struct implementors, get-only and get/set properties, and `System.Action` / generic `System.Func` delegates in compiler and interpreter tests

## Root cause
`src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs:2391-2393` and `:2533-2534` only reached the delegate-member fallback for `StructSymbol` receivers. The shared helper in `ExpressionBinder.Calls.Arguments.cs:422-469` also accepted only `StructSymbol`, even though `TypeMemberModel.TryGetProperty` already supports `InterfaceSymbol`. Interface delegate-property calls therefore skipped the property load + indirect invocation path and fell through to GS0159.

## Base to head evidence
Base `a6b1561f2`:
- `gsc` repro: `GS0159: Cannot find function P.`
- `gsi` repro: `GS0159: Cannot find function P.`

Head `81e44da88`:
- direct `out/bin/Release/Compiler/gsc` compile + execution: `1`
- direct `out/bin/Release/Repl/gsi` execution: `1`
- serial Release solution build passed; `GSharp.Core.dll` was non-empty and MD5-identical across Compiler, Core.Tests, Compiler.Tests, and Interpreter.Tests outputs

## Tests
Same base (`a6b1561f2`) for both measurements:

| Suite | Base | Head |
| --- | ---: | ---: |
| Core | 7028 passed, 1 skipped | 7028 passed, 1 skipped |
| Compiler | 3906 passed | 3907 passed |
| Interpreter | 491 passed | 492 passed |

Closes #2925

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.Accessor.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.Accessor.cs
@@ -3383,7 +3383,7 @@ internal sealed partial class ExpressionBinder
         }
 
         if (structSym != null
-            && TryBindUserStructDelegateMemberInvocation(
+            && TryBindUserDelegateMemberInvocation(
                 receiver: null,
                 structSym,
                 methodName,
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Arguments.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Arguments.cs
@@ -419,9 +419,9 @@ internal sealed partial class ExpressionBinder
         return null;
     }
 
-    private bool TryBindUserStructDelegateMemberInvocation(
+    private bool TryBindUserDelegateMemberInvocation(
         BoundExpression receiver,
-        StructSymbol receiverStruct,
+        TypeSymbol receiverType,
         string methodName,
         ImmutableArray<BoundExpression> arguments,
         CallExpressionSyntax ce,
@@ -430,9 +430,10 @@ internal sealed partial class ExpressionBinder
     {
         result = null;
 
+        var receiverStruct = receiverType as StructSymbol;
         FieldSymbol matchedField = null;
         StructSymbol declaringType = null;
-        if (isStatic)
+        if (isStatic && receiverStruct != null)
         {
             TypeMemberModel.TryGetStaticFieldIncludingInherited(
                 receiverStruct,
@@ -440,7 +441,7 @@ internal sealed partial class ExpressionBinder
                 out matchedField,
                 out declaringType);
         }
-        else
+        else if (receiverStruct != null)
         {
             // Walk the base chain so an inherited delegate field on a base class
             // is invokable on a derived instance.
@@ -465,7 +466,7 @@ internal sealed partial class ExpressionBinder
                     out var property,
                     out var propertyDeclaringType)
                 : TypeMemberModel.TryGetProperty(
-                    receiverStruct,
+                    receiverType,
                     methodName,
                     out property,
                     out propertyDeclaringType);
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
@@ -2384,16 +2384,15 @@ internal sealed partial class ExpressionBinder
                 {
                     return overloads.BindUserInstanceCall(receiver, defaultIfaceMethod, arguments, ce, argumentNames);
                 }
+            }
 
-                // Issue #527: fall back to a delegate/function-typed member on
-                // the user struct/class. This is the same delegate-as-callable
-                // dispatch used for the imported-CLR path below; here the
-                // receiver's ClrType is null (the user type has not yet been
-                // emitted) so we have to handle the symbol-only shape too.
-                if (TryBindUserStructDelegateMemberInvocation(receiver, userClass, methodName, arguments, ce, isStatic: false, out var userDelegateFieldCall))
-                {
-                    return userDelegateFieldCall;
-                }
+            // Issues #527 and #2925: source struct/class and interface
+            // delegate members are callable before their CLR types exist.
+            if (receiver != null
+                && receiver.Type is StructSymbol or InterfaceSymbol
+                && TryBindUserDelegateMemberInvocation(receiver, receiver.Type, methodName, arguments, ce, isStatic: false, out var userDelegateFieldCall))
+            {
+                return userDelegateFieldCall;
             }
 
             // Phase 3.B.6 / ADR-0019: extension function fallback for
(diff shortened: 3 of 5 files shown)


DavidObando:
## Rubber-duck review — PR #3016 at `81e44da88`, merged into `main` `f70626452`

**Verdict: MERGEABLE.** No blocking issues. The fix is small, correctly shaped, genuinely pinned, and I independently reproduced both the defect and the repair. Three non-blocking items below.

Merge-base `a6b1561f2` (**9 behind / 1 ahead**). All probes run on a real merge commit `2a4083f27`, parents exactly `f70626452` + `81e44da88`.

### What I verified

*All `file:line` anchors below are in the **merged** tree `2a4083f27` (the PR body cites pre-merge numbers, which differ by ~48 lines).*

**Merge-build (the #3000 class).** `git merge-tree` rc=0. Merge build `BUILD_EXIT=0`, `error-lines=0`, `0 Error(s)`, 39 s, `-p:ContinuousIntegrationBuild=true`, never `--no-build`. Gate 2: **40** `GSharp.Core.dll` copies examined, **0** zero-byte.

**Static sweep.** Detector A: **0 hits / 14 net-deleted candidates / 251 PR-added lines**. The candidate set contains `OuterMethodTypeParameters` (the #3000 calibration symbol), so the pattern is proven to match a known-present case — this is a *run* check, not an empty one. Inverse direction **re-derived at the current tip, not inherited**: the PR net-deletes exactly one symbol, `TryBindUserStructDelegateMemberInvocation`, and `main` added **0** new references to it since the base (3148 main-added lines scanned).

**Docs gate (#2981).** Not exposed — no `DiagnosticDescriptors.cs` change, no `docs/diagnostics.md` row added or deleted.

**The defect and the fix, reproduced independently:**

| probe | `main f70626452` | merged |
|---|---|---|
| interface delegate property `i.P()` | **`GS0159: Cannot find function P.`** | compiles; `Assembly.Load` ok, `GetTypes()` ok |

**Mutation, not just green.** Reverted the one load-bearing token — `TryGetProperty(receiverType, …)` → `receiverStruct` at `ExpressionBinder.Calls.Arguments.cs:468` — rebuilt, and the test went **RED** with `GS0159: Cannot find function P.`, i.e. exactly the pre-fix symptom. Restored, rebuilt, **GREEN**, tree clean.

```
mutant    GSharp.Core.dll md5 = 65eb8a013b0c6df4def929b0d1ff2c3e   Compiler.Tests ~Issue2925: Failed 1
restored  GSharp.Core.dll md5 = ec47d4710a2bf6957f58b7f4db2bff0e   Compiler.Tests ~Issue2925: Passed 1
                                                                    Interpreter.Tests ~Issue2925: Passed 1
```

*(Method note: my first md5 was taken on `gsc`, which is the driver and is byte-identical across a `GSharp.Core.dll` change — that measurement was vacuous and I discarded it. The figures above are the correct artifact.)*

### Collision check with #2968 — **no conflict; merge order is unconstrained**

You asked to be told at once if these two collide. They do not:

- `git merge-tree refs/pr3016 refs/pr2968` → **rc=0**
- **zero file overlap** (3016 touches 5 files, 2968 touches 9)
- pairwise compile-break check both directions: #3016 references **0/3** symbols #2968 net-deletes (251 lines examined); #2968 references **0/1** symbols #3016 net-deletes (720 lines examined)

They also sit in different layers — #3016 is binding-only (`ExpressionBinder.*`), #2968 is declaration-verification + emit + evaluator. I flagged the zero-overlap as *suspicious* rather than reassuring, since that is the #3000 signature, which is why I ran the pairwise symbol check rather than stopping at the file-overlap result.

---

### SHOULD-FIX 1 — the widened helper still passes a struct-only receiver on the static path

`src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Arguments.cs:463-467`

The signature is now `TypeSymbol receiverType`, and `receiverStruct` is `receiverType as StructSymbol` — **null for an interface**. The non-static property lookup was updated to `receiverType`, but the static branch still passes `receiverStruct`:

```csharp
var foundProperty = isStatic
    ? TypeMemberModel.TryGetStaticPropertyIncludingInherited(
        receiverStruct,      // null when receiverType is an InterfaceSymbol
        ...
    : TypeMemberModel.TryGetProperty(
        receiverType,        // updated
```

**Not live today** — all three call sites are safe (`ExpressionBinder.Access.Accessor.cs:3386` guards `structSym != null`; both `Invocation.cs` sites pass `isStatic: false`). But the parameter type now *advertises* interface support, so the next caller that passes an interface with `isStatic: true` gets a silent miss or an NRE. Either take `receiverStruct` on the static path with an explicit early-out, or document the precondition on the parameter.

### SHOULD-FIX 2 — one `[Fact]` per test file for a two-call-site widening

`test/Compiler.Tests/Emit/Issue2925InterfaceDelegatePropertyInvocationTests.cs` (144 lines) and `test/Interpreter.Tests/Issue2925InterfaceDelegatePropertyInvocationTests.cs` (85 lines) contain **exactly one `[Fact]` each**. The bulk is source-string payload.

That single test does cover class/struct implementors, get-only and get/set, `Action` and `Func` — so coverage is broader than the count suggests, and my mutation confirms it is load-bearing. But the two call sites are *not* separately pinned: `Invocation.cs:2439` (primary path) and `:2580` (user-class-priority path) are distinct reachability contexts, and one mutation killing one test cannot distinguish them. A second `[Fact]` reaching the `:2580` path would make a future regression attributable rather than merely detectable.

Note also that for an interface receiver `TypeMemberModel.TryGetProperty` returns **`declaringType = null`** (`TypeMemberModel.cs:387`), so `BoundPropertyAccessExpression` is constructed with a null declaring type on this new path. That works today and the emit test proves it — but it is exactly the kind of invariant that deserves its own pin rather than riding on one shared test.

### NIT — the ADR-0118 indexer exclusion is struct-only, and interfaces now route through it

`src/Core/CodeAnalysis/Symbols/TypeMemberModel.cs:362` skips indexers (`if (p.IsIndexer) continue;`) in the **`StructSymbol`** branch only; the `InterfaceSymbol` branch at `:376` has no such skip. #3016 newly routes interface receivers through this function, so an interface indexer is reachable by the name `Item` on this path where a struct's would be excluded.

I did **not** construct a reachable exploit — it needs a delegate-typed interface indexer with a getter — so I am filing this as a NIT, not a defect. It is worth a look mainly because #2968 is concurrently rewriting interface indexer contracts.

---

### Risk I looked for and did **not** find

Both hunks move the fallback **out of** an enclosing `if`, which is the shape that usually hides a scope widening. I checked both:

- `Invocation.cs:2439` — the old enclosing block was `if (receiver != null && receiver.Type is StructSymbol userClass)`, and every other exit from it `return`s. The new guard `receiver != null && receiver.Type is StructSymbol or InterfaceSymbol` is therefore **exactly equivalent for structs** and additive for interfaces.
- `Invocation.cs:2580` — no `receiver != null` guard, but the pre-existing `if (receiver.Type is StructSymbol userClassPriority)` immediately above already dereferences `receiver` unconditionally, so no new NRE is introduced. The asymmetry between the two sites is correct, not an oversight.

I also tried to construct an **extension-function vs delegate-property precedence** conflict, since the fallback now runs before the ADR-0019 extension fallback for interface receivers. It is **not constructible**: extension receivers are restricted to struct/class in the same package (`GS0103`), so no interface-receiver extension can exist to lose precedence.

Method notes: three-dot diffs from a computed `git merge-base`; net state read at the tip; exit codes assigned to a variable before any other command. No worktree I do not own was modified; the merge tree was left clean.

Issue #3015, opened 2026-08-01
gsi: imported-base virtual dispatch loses G# derived type identity


DavidObando:
## Problem

The interpreter loses the G# derived type identity when an inherited virtual method is dispatched through an imported CLR base backing object.

```gsharp
import System

class Args : EventArgs {
}

Console.WriteLine(Args().ToString())
```

At `main` `e80132516`:

```text
gsc: GSharp.Example.ImportedBaseClass.Args
gsi: System.EventArgs
```

The ordinary `Args()` construction reproduces this, so it is separate from the missing class-literal construction fixed for #2993/#2996. After that fix, `Args{}.ToString()` reaches the same `System.EventArgs` result instead of the previous `Args()` fallback, but still does not match compiled behavior.

## Mechanism

`AllocateClrBacking` creates an instance of the imported base type. `UnwrapClrReceiver` correctly supplies that object for inherited reflection calls, but virtual dispatch then observes an actual `System.EventArgs` instance rather than the emitted G# derived runtime type. Removing the bridge is not a fix because it would re-break inherited properties and methods such as `MemoryStream.CanRead`.

## Expected

Inherited virtual calls under `gsi` must preserve the G# derived type's identity, matching compiled behavior, for both ordinary and literal construction.

Found while fixing #2993 and #2996.

PR #3017, opened 2026-08-01
gsi: construct class literals before applying initializers

### LINKED ISSUE #2993: gsi: collection initializer for a get-only property invokes Add without a receiver
## Problem

A collection initializer targeting a get-only / `init` property fails in the interpreter.

`samples/GetOnlyCollectionInit.gs`:

```gsharp
class Bag {
    prop Items IList[int32] { get; init; }
    init() { Items = List[int32]() }
}

var one = Bag{ Items: { 42 } }
```

```
compiled: 0 1 42 3 10 30 2 alice bob 2 1 2
gsi:      GS9999: Non-static method requires a target.
```

## Root cause

`Non-static method requires a target` is `System.Reflection`'s `TargetException`, raised when an instance method is invoked with a null target. The collection-initializer path reads the property and calls `Add` on the result, but dispatches without a receiver.

## Expected

`Bag{ Items: { 42 } }` must call the property getter on the freshly constructed instance and then `Add` on the returned collection, matching the emitted behaviour.

### LINKED ISSUE #2996: gsi: members inherited from an imported CLR base class dispatch against the wrong instance
## Problem

`samples/ImportedBaseClass.gs` — a G# class deriving from an imported CLR base class — fails in the interpreter.

```
compiled: True True 3 3 buffer GSharp.Example.ImportedBaseClass.Args
gsi:      GS9999: Object type System.IO.MemoryStream does not match target type
          GSharp.Core.CodeAnalysis.<...>
```

(The two `GS0286` warnings in the output are **not** part of the divergence — `gsc` reports the same warnings for this sample; it is a shared style warning, not an interpreter behaviour difference.)

## Root cause

The interpreter represents a G#-declared instance as its own value object, but a member inherited from an imported CLR base class is dispatched by reflection against the *CLR* base type, which expects the real `MemoryStream` instance. The two representations are not bridged for the imported-base case.

## Expected

A G# class deriving from an imported CLR type must present the underlying CLR instance when dispatching inherited members, so `Length`, `Position` and friends work as they do compiled.

DavidObando:
## Summary

Make interpreted class literals perform construction before applying named member initializers.

`EvaluateStructLiteralExpression` previously allocated a bare `StructValue`, initialized its field dictionary, and applied literal entries. It never invoked a declared `init()` and never called `AllocateClrBacking`. This caused both reported failures:

- constructor-established state was absent, including silent wrong scalar values (#2993)
- classes derived from imported CLR bases had no reflection receiver (#2996)

The evaluator now reuses `EvaluateConstructorCallExpression` for class literals with a default/parameterless construction path, then applies literal entries so explicit values win over constructor defaults. Primary-constructor literals retain their existing default-construction semantics and receive CLR backing without invoking the primary constructor with nonexistent positional arguments.

## Design correction found during validation

Unconditionally delegating every class literal to `EvaluateConstructorCallExpression`, as proposed in the issue design, is not safe. Existing `Circle{Radius: 3}` primary-constructor literals then pass zero arguments to the synthesized primary constructor and fail with `Index was outside the bounds of the array.` Two `Issue1636TypeTestNarrowingSubtypeGateTests` caught this. The final guard preserves the emitter's default-construction-plus-member-write behavior for that shape; all seven tests in that class pass.

The production footprint remains confined to `EvaluateStructLiteralExpression`.

## Differentials

| Probe | Before | After |
|---|---|---|
| `Counter{M:1}` | `0 / 1` | `ctor-ran / 7 / 1` |
| `Bag{Items:{}}` | `GS9999: Non-static method requires a target.` | `0` |
| `Buffer{}.CanRead` | `GS9999: ... target type ... StructValue` | `True` |

The scalar regression test also assigns `M = 9` in the constructor and requires the literal's `M: 1` to win, pinning construction order rather than only constructor execution.

Removing the production diff made all three focused tests fail; restoring it made all three pass. The round trip rebuilt the changed assembly:

```text
source md5:   10d892b56039ffb89d3c641a3d82375b -> 59841d0f06b5b723cd61f8715c2e0a25 -> 10d892b56039ffb89d3c641a3d82375b
Core DLL md5: 7188a264417e23ba994bde1826dd701c -> 83176327f32690549e5dfb08a119f9c1 -> 7188a264417e23ba994bde1826dd701c
```

Ordinary construction remained byte-identical before and after (`Buffer()`, `Bag()`, and concrete `List[int32]`): output `True / 0 / 4`, md5 `620ffd19148725aa0b51170546af07ce` in both states.

## Validation

Executed baseline at `e80132516`:

| Suite | Passed | Skipped |
|---|---:|---:|
| Core | 7028 | 1 |
| Compiler | 3931 | 0 |
| Interpreter | 493 | 0 |
| LanguageServer | 452 | 0 |

Final checks:

- Core: 7028 passed, 1 skipped
- Interpreter: 496 passed
- focused regressions: 3 passed
- `dotnet build GSharp.sln -c Release -graph`: passed
- all 40 `GSharp.Core.dll` files were nonzero
- five ordinary output copies agreed at md5 `7188a264417e23ba994bde1826dd701c`
- no `MSB302[67]` or `MSB3883`

## Follow-up discovered

The imported-base sample now executes all inherited property/method operations correctly. Its final inherited virtual `ToString()` still prints `System.EventArgs` instead of the emitted G# derived type name. Ordinary `Args()` construction has the same mismatch, proving it is separate from class-literal construction. Filed #3015 with the `Oahu` label rather than expanding this fix beyond the assigned evaluator footprint.

Fixes #2993
Fixes #2996

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -651,21 +651,41 @@ public sealed partial class Evaluator
 
     private object EvaluateStructLiteralExpression(BoundStructLiteralExpression node)
     {
-        var sv = new StructValue(node.StructType);
+        var constructor = node.StructType.EffectiveExplicitConstructors
+            .FirstOrDefault(c => c.Function.Parameters.IsDefaultOrEmpty);
+        var invokesConstructor = node.StructType.IsClass
+            && node.StructType.PrimaryConstructorParameters.IsDefaultOrEmpty
+            && (constructor != null || node.StructType.EffectiveExplicitConstructors.IsDefaultOrEmpty);
+        var sv = invokesConstructor
+            ? (StructValue)EvaluateConstructorCallExpression(
+                new BoundConstructorCallExpression(
+                    node.Syntax,
+                    node.StructType,
+                    ImmutableArray<BoundExpression>.Empty,
+                    constructor))
+            : new StructValue(node.StructType);
 
-        // Default-initialize all fields (walking inheritance), then apply explicit initializers.
-        for (var t = node.StructType; t != null; t = t.BaseClass)
+        // Primary-constructor literals use default construction before applying
+        // named initializers, matching the emitter's newobj/stfld sequence.
+        if (!invokesConstructor)
         {
-            foreach (var f in t.Fields)
+            for (var t = node.StructType; t != null; t = t.BaseClass)
             {
-                if (!sv.Fields.ContainsKey(f.Name))
+                foreach (var f in t.Fields)
                 {
-                    sv.Fields[f.Name] = ClrDefaultValue(f.Type);
+                    if (!sv.Fields.ContainsKey(f.Name))
+                    {
+                        sv.Fields[f.Name] = ClrDefaultValue(f.Type);
+                    }
                 }
             }
-        }
 
-        ApplyClassFieldInitializers(sv, node.StructType);
+            ApplyClassFieldInitializers(sv, node.StructType);
+            if (node.StructType.IsClass)
+            {
+                AllocateClrBacking(sv, node.StructType, activeInit: null);
+            }
+        }
 
         foreach (var init in node.Initializers)
         {
--- a/test/Interpreter.Tests/Issue2993StructLiteralConstructionTests.cs
+++ b/test/Interpreter.Tests/Issue2993StructLiteralConstructionTests.cs
@@ -1,0 +1,111 @@
+// <copyright file="Issue2993StructLiteralConstructionTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using GSharp.Core.CodeAnalysis.Compilation;
+using GSharp.Core.CodeAnalysis.Symbols;
+using GSharp.Core.CodeAnalysis.Syntax;
+using Xunit;
+
+namespace GSharp.Interpreter.Tests;
+
+/// <summary>
+/// Issues #2993 and #2996: class literals must run construction before applying
+/// their explicit member initializers.
+/// </summary>
+public class Issue2993StructLiteralConstructionTests
+{
+    [Fact]
+    public void ClassLiteral_RunsConstructorBeforeExplicitInitializers()
+    {
+        const string Source = """
+            import System
+
+            class Counter {
+                prop N int32 { get; init; }
+                prop M int32 { get; set; }
+
+                init() {
+                    Console.WriteLine("ctor-ran")
+                    N = 7
+                    M = 9
+                }
+            }
+
+            var c = Counter{ M: 1 }
+            Console.WriteLine(c.N)
+            Console.WriteLine(c.M)
+            """;
+
+        Assert.Equal("ctor-ran\n7\n1\n", Evaluate(Source));
+    }
+
+    [Fact]
+    public void ClassLiteral_PreservesConstructorInitializedCollection()
+    {
+        const string Source = """
+            import System
+            import System.Collections.Generic
+
+            class Bag {
+                prop Items IList[int32] { get; init; }
+
+                init() {
+                    Items = List[int32]()
+                }
+            }
+
+            var b = Bag{ Items: {} }
+            Console.WriteLine(b.Items.Count)
+            """;
+
+        Assert.Equal("0\n", Evaluate(Source));
+    }
+
+    [Fact]
+    public void ClassLiteral_AllocatesImportedBaseBacking()
+    {
+        const string Source = """
+            import System
+            import System.IO
+
+            class Buffer : MemoryStream {
+                func Describe(label string) string {
+                    return label
+                }
+            }
+
+            var b = Buffer{}
+            Console.WriteLine(b.CanRead)
+            """;
+
+        Assert.Equal("True\n", Evaluate(Source));
+    }
+
+    private static string Evaluate(string source)
+    {
+        var compilation = new Compilation(SyntaxTree.Parse(source));
+
+        using var outWriter = new StringWriter();
+        var previousOut = Console.Out;
+        Console.SetOut(outWriter);
+        try
+        {
+            var result = compilation.Evaluate(new Dictionary<VariableSymbol, object>());
+            var errors = result.Diagnostics.Where(d => d.IsError).ToArray();
+            Assert.True(
+                errors.Length == 0,
+                "evaluation failed:\n" + string.Join("\n", errors.Select(d => d.ToString())));
+        }
+        finally
+        {
+            Console.SetOut(previousOut);
+        }
+
+        return outWriter.ToString().Replace("\r\n", "\n");
+    }
+}
--- a/test/Interpreter.Tests/Issue2993StructLiteralConstructionTests.cs
+++ b/test/Interpreter.Tests/Issue2993StructLiteralConstructionTests.cs
@@ -86,6 +86,23 @@ public class Issue2993StructLiteralConstructionTests
         Assert.Equal("True\n", Evaluate(Source));
     }
 
+    [Fact]
+    public void ClassLiteral_WithPrimaryConstructor_AllocatesImportedBaseBacking()
+    {
+        const string Source = """
+            import System
+            import System.IO
+
+            class Buffer(Label string) : MemoryStream {
+            }
+
+            var b = Buffer{ Label: "primary" }
+            Console.WriteLine(b.CanRead)
+            """;
+
+        Assert.Equal("True\n", Evaluate(Source));
+    }
+
     private static string Evaluate(string source)
     {
         var compilation = new Compilation(SyntaxTree.Parse(source));


DavidObando:
## Review — NOT MERGEABLE at `14954aff1` (one blocker; the semantics are right)

Measured against `main` = `f70626452`.

**Compile-mergeability: PASS.** Merge base is `e80132516` (**5 behind**, 1 ahead). `git merge gh/main` → 0 conflicts; `dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true` on the **merged** tree → **rc=0**. Gates: 22 `GSharp.Core.dll` copies examined / 0 zero-byte; 1 md5 group across the ordinary `out/bin/Release` paths; race grep 0 of 76 log lines (control matched). `Interpreter.Tests` on the merged tree: **497 passed, 0 failed**.

**The core semantics are correct, and I verified them against `gsc` rather than against the PR's own description:**

| probe | `gsi` @ merged | `gsc` |
|---|---|---|
| ctor sets `M=9` and records `Seen=M`; literal `Counter{M: 1}` | `1` / `0` | `1` / `0` |
| `Foo()` then `Foo{M: 7}`, ctor prints | `ctor/5/ctor/7` | `ctor/5/ctor/7` |

Construction order matches the emitter, the literal wins over constructor defaults, and there is **no** double construction. The `invokesConstructor` guard that the body describes as a design correction is doing real work.

---

## BLOCKER — a load-bearing line of this PR is invisible to the entire interpreter suite

The `else` branch adds:

```csharp
if (node.StructType.IsClass)
{
    AllocateClrBacking(sv, node.StructType, activeInit: null);
}
```

This is the **primary-constructor** class path — and it is squarely inside #2996's declared scope ("classes derived from imported CLR bases had no reflection receiver").

```gsharp
class Buf2(tag int32) : MemoryStream {
    var Extra int32
}
let b = Buf2{Extra: 3}
Console.WriteLine(b.CanRead)
```

| state | value |
|---|---|
| `f70626452` (base) | `GS9999: Object type System.IO.MemoryStream does not match target type …StructValue.` |
| `14954aff1` (head) | `True` |

**Mutation.** Removing *only* those four lines and rebuilding Release:

- the probe flips `True` → `GS9999` (identical to base), so the line is genuinely load-bearing
- `--filter FullyQualifiedName~Issue2993StructLiteralConstructionTests` → **`Passed! Failed: 0, Passed: 3`**
- the **whole suite** → **`Passed! Failed: 0, Passed: 497, Total: 497`**
- restored → probe flips back to `True`, `git status` clean

All three shipped tests use classes **without** a primary constructor (`Counter`, `Bag`, `Buffer : MemoryStream`), so every primary-constructor literal path this PR adds is unpinned. 497 green tests say nothing about it.

**Fix:** add the `Buf2` shape above as a fourth test. It is two lines of G# and it pins the half of #2996 that is currently unguarded.

---

## SHOULD-FIX

**S1 — the declared evidence is measured at a stale base.** The body's suite table says *"Executed baseline at `e80132516`"*, which is 5 commits behind `main` and predates #2964, #2929, #2980, #2981 and #3009. Those numbers do not describe the merged state. I re-measured for you — merged build rc=0 and Interpreter.Tests 497/497 (the body reports 496 at the old base) — so nothing is actually broken, but please rebase and restate, since the next reader will otherwise treat stale numbers as current.

**S2 — behavioural reach is wider than "confined to `EvaluateStructLiteralExpression`".** The file-level claim is true; the reach is not. `BoundStructLiteralExpression` is synthesized at **11 sites**, including overload-resolved **constructor calls** (`OverloadResolver.Constructors.cs:83`, `:455`, `:715`), anonymous types (`ExpressionBinder.Literals.cs:954`), and conversions (`ConversionClassifier.cs:3082`). So this change can alter ordinary `Foo()` construction and anonymous-type construction, not only user-written `Foo{…}` literals. I verified ordinary construction matches `gsc` exactly (table above); the anonymous-type path is unverified. Please state which of those 11 sites are class-typed and therefore newly route through `EvaluateConstructorCallExpression`.

## NIT

**N1** — the new call passes `activeInit: null`, while every other `AllocateClrBacking` caller passes a real `BaseConstructorInitializer`. For `class Buf(cap int32) : MemoryStream(cap)` reached via a literal, the CLR base is therefore built with its parameterless constructor (`Capacity` = 0). Passing `null` looks deliberate — the args reference unbound primary parameters, and `AllocateClrBacking`'s own comment says ancestor base-init args cannot be re-evaluated there — so I am not calling it a defect. Worth one sentence in the body so it is not later read as an oversight.

**N2** — two literal shapes have **no `gsc` ground truth** to validate against: a class with only parameterized constructors (`OnlyParam{M: 1}`), and a primary-ctor class with `: ClrBase(args)` via a literal. Both make `gsc` emit a program that dies with `InvalidProgramException` (exit 134) at **both** `f70626452` and this head, so these are pre-existing emitter defects, not this PR's. Flagging only so the shapes are not used as acceptance evidence.

---

**Verdict: NOT MERGEABLE**, on the single blocker. The design correction described in the body is sound, construction order is verified correct against the emitter, and the merge builds and passes clean — this needs one test, not a redesign.

DavidObando:
Blocker addressed in content: added `ClassLiteral_WithPrimaryConstructor_AllocatesImportedBaseBacking`, using `class Buffer(Label string) : MemoryStream` and `Buffer{ Label: "primary" }`.

Mutation: removed only the else-branch `AllocateClrBacking(sv, node.StructType, activeInit: null)` call. The new test failed:

```text
ClassLiteral_WithPrimaryConstructor_AllocatesImportedBaseBacking [FAIL]
evaluation failed:
Object type System.IO.MemoryStream does not match target type GSharp.Core.CodeAnalysis.StructValue.

Failed!  - Failed: 1, Passed: 0, Skipped: 0, Total: 1
```

Restoring the call made the same test pass, then the focused class passed 4/4.

Artifact round trip used `GSharp.Core.dll`, not `gsc`:

```text
source: ca9de9ae1d51c9f95b7cc81c2924bbd6
     -> 852bdd3bff4e52fafbde884ac2aedce3
     -> ca9de9ae1d51c9f95b7cc81c2924bbd6
Core:   b9c71d83a220f630c8070e49003fb3ba
     -> e0c2dddd40231d79e71b33d222f4e35b
     -> b9c71d83a220f630c8070e49003fb3ba
```

Rebased onto `47a400359` and pushed. Rebased merge build with `-p:ContinuousIntegrationBuild=true` passed; 40 Core DLLs checked nonzero and five ordinary paths formed one MD5 group.

DavidObando:
## Re-review — blocker cleared. MERGEABLE at `1572bfad9`

Narrow re-check of my earlier blocker only. Measured against `main` = `428f31779`; merge base **`47a400359`** (3 behind, 2 ahead), merged worktree `bdceeab02`.

**The production change is byte-identical to what I reviewed.** I compared the PR's own three-dot patch at the old base (`e80132516…14954aff1`) against the new one (`47a400359…1572bfad9`) — identical once index/hunk headers are stripped. Only the test was added, so re-verification is correctly narrow. (A two-dot `14954aff1..1572bfad9` reports 16 files; those are upstream arrivals across the rebase, not PR scope.)

### Mutation re-run — I did not accept the report

Merged Release build **rc=0, 0 warnings**. Clean `GSharp.Core.dll` md5 `dacc2a63…`. Baseline Interpreter.Tests **509/509**.

Deleted the else-branch call:

```csharp
if (node.StructType.IsClass)
{
    AllocateClrBacking(sv, node.StructType, activeInit: null);
}
```

| step | `GSharp.Core.dll` md5 | result |
|---|---|---|
| clean | `dacc2a63…` | 509 passed, 0 failed |
| mutant | `4fe424b3…` | **1 failed**, 508 passed |
| restored | `dacc2a63…` | focused class 4/4 |

The single failure is exactly **`ClassLiteral_WithPrimaryConstructor_AllocatesImportedBaseBacking`**, and its message is the precise symptom I originally reported:

```
Object type System.IO.MemoryStream does not match target type ...StructValue.
```

Mutant and restored hashes differ from and return to the clean hash, so both states were genuinely rebuilt — no stale-artifact read. I hashed `GSharp.Core.dll` rather than the `gsc` driver, which is byte-identical across Core changes.

### The new test really does exercise the *else* branch

Structurally: `invokesConstructor` requires `node.StructType.PrimaryConstructorParameters.IsDefaultOrEmpty`, so `class Buffer(Label string) : MemoryStream` forces it `false` and takes the `new StructValue(...)` path into the `AllocateClrBacking` call.

Empirically the mutation is the proof, and it is stronger than reading: if this test took the `if` branch, deleting the **else**-branch call could not have changed its outcome. It did. Previously all three tests used classes without primary constructors and the whole suite passed 497/497 with the line removed; that hole is now closed.

### Semantics still correct after the rebase

Re-ran my `gsc` ground-truth probes against the merged tree — all three still agree exactly:

| probe | `gsi` | `gsc` |
|---|---|---|
| initializer/constructor ordering | `1` / `0` | `1` / `0` |
| no double construction | `ctor/5/ctor/7` | `ctor/5/ctor/7` |
| primary-ctor class over CLR base | `True` | `True` |

### Still open, both non-blocking

- **S1** — the body still reports `Executed baseline at e80132516` with Interpreter `493`/`496`. That base is now two rebases stale. Current merged state is **509/509**; I have measured it, so nothing is at risk, but the numbers in the body do not describe what will merge.
- **S2** — the body still says the production footprint "remains confined to `EvaluateStructLiteralExpression`". True at file level, wider in reach: `BoundStructLiteralExpression` is synthesized at 11 sites, including `OverloadResolver.Constructors.cs:83`/`:455`/`:715` and `ExpressionBinder.Literals.cs:954` (anonymous types). Worth one sentence rather than a code change.

Neither blocks. **Verdict: MERGEABLE.**

Issue #3004, opened 2026-08-01
gsi: pointer dereference silently returns a stale value — wrong answer with exit 0


DavidObando:
## Summary

The interpreter silently computes a **wrong answer** for pointer dereference and exits **0**. Every other known `gsi` defect is loud — it crashes, throws, or emits a diagnostic. This one returns a plausible number and reports success.

## Repro

```gsharp
import System

unsafe {
    var x int32 = 7
    var p *int32 = &x
    x = 42
    Console.WriteLine(*p)
}
```

| | output | exit code |
|---|---|---|
| compiled (`gsc` → `dotnet ptr.dll`) | `42` | 0 |
| interpreted (`gsi ptr.gs`) | **`7`** | **0** |

No diagnostic, no warning, no stderr output in either case. Confirmed a second way: reading `*p` before and after a write yields `1, 2` compiled and `1, 1` interpreted.

## Root cause

`src/Core/CodeAnalysis/Evaluator.Calls.cs:1108-1119` — `EvaluateAddressOfExpression` and `EvaluateDereferenceExpression` are **both the identity function**, and the comments there say so.

Consequently `&x` evaluates to the *value* of `x` at the moment the address is taken, and `*p` returns that snapshot forever. Any write to `x` afterwards is invisible through `p`. The construct does not fail — it quietly returns stale data.

The identity implementation is not gratuitous: it makes the *call-site* `ref`/`out` forms work, because those are handled separately by the ref-slot machinery (`Evaluator.Calls.cs:69`, `:1134`, `BuildRefSlots`), which syntactically pattern-matches `BoundAddressOfExpression` **in argument position** and does an explicit write-back after the call. That path is correct. What is not correct is that the same identity function is also used for free-standing `&`/`*` in expression position, where nothing compensates.

## What this implies

**The interpreter has no pointer model, and the current code silently pretends that it does.** There is no storage-location abstraction in the evaluator — frames are `ConcurrentDictionary<VariableSymbol, object>` and a pointer would need to denote a location, not a value. This is not a small gap to close; it is the same "add a memory model" problem that #2956 identified for `fixed`.

## Recommended fix

Make it **loud**. `&` and `*` in expression position should throw the same kind of explicit boundary error that `fixed` already throws at `Evaluator.Statements.cs:153-158`:

> `'fixed' (pinning) statements are not supported in the interpreter; they require the CIL pinned-local emit path.`

[...]

PR #3018, opened 2026-08-01
Document gsi compiled-only storage boundary

### LINKED ISSUE #2956: gsi: fixed statements are unsupported by the interpreter

DavidObando:
## Summary

- accept option B in ADR-0153: the interpreter models values, not storage, so storage-dependent pinning, allocation, pointer, and function-pointer operations are compiled-only
- link the existing `fixed` design to that boundary and document the `GS0513` compiled-only classification in both diagnostics references
- add script-mode `gsi` coverage for array, string, pinnable-span, `stackalloc`, `sizeof`, function-pointer address/invocation, and the supported plain-`unsafe` guard

Fixes #2956

## Scope

The evaluator already contains the intended `fixed` boundary and self-contained message. This PR changes no production code; it records the accepted storage-model contract and pins existing behavior. Unsafe/native language validity and CIL emission remain outside ADR-0153.

No diagnostic ID or `| GSxxxx |` row is added. Clean boundary sites use the existing `GS0513` diagnostic with exact construct-specific messages.

## Review fixes

- renumber ADR-0152 to ADR-0153 to avoid #3024
- tolerate source-location/caret renderer context while preserving exit-code, stdout, and message assertions
- document that managed ref locals must alias captured storage, with the queued fix in #3032
- record #3028's `GS0513` boundary for free-standing unmanaged pointer operations

## Verification

- `Issue2956InterpreterBoundaryTests`: 8 examined, 8 passed
- 40 `GSharp.Core.dll` copies examined, 0 zero-byte files
- five ordinary `GSharp.Core.dll` copies shared md5 `bafd05ea00792b7beaca480319e43e65`
- boundary tests assert exit code 1, empty stdout, and the self-contained diagnostic message; diagnostic ID and renderer context may change

No full suites or e2e tests were run; fleet throttle requires the narrowest selector, and this PR contains documentation plus targeted interpreter tests only.

### DIFF
--- a/docs/adr/0125-fixed-pinned-statement.md
+++ b/docs/adr/0125-fixed-pinned-statement.md
@@ -3,7 +3,7 @@
 - **Status**: Accepted
 - **Date**: 2026-06-27
 - **Phase**: Phase 9 — low-level / interop depth
-- **Related**: ADR-0039 (managed by-ref pointers / address-of / dereference), ADR-0056 (ref-returning members / `modreq(InAttribute)` ref-returns), ADR-0122 (unsafe context and unmanaged raw pointers `*T`, issue [#1014](https://github.com/DavidObando/gsharp/issues/1014)), ADR-0124 (`stackalloc`/`localloc`, issue [#1024](https://github.com/DavidObando/gsharp/issues/1024)), ADR-0152 (interpreter compiled-only storage boundary), issue [#1026](https://github.com/DavidObando/gsharp/issues/1026), issue [#1043](https://github.com/DavidObando/gsharp/issues/1043), issue [#2900](https://github.com/DavidObando/gsharp/issues/2900)
+- **Related**: ADR-0039 (managed by-ref pointers / address-of / dereference), ADR-0056 (ref-returning members / `modreq(InAttribute)` ref-returns), ADR-0122 (unsafe context and unmanaged raw pointers `*T`, issue [#1014](https://github.com/DavidObando/gsharp/issues/1014)), ADR-0124 (`stackalloc`/`localloc`, issue [#1024](https://github.com/DavidObando/gsharp/issues/1024)), ADR-0153 (interpreter compiled-only storage boundary), issue [#1026](https://github.com/DavidObando/gsharp/issues/1026), issue [#1043](https://github.com/DavidObando/gsharp/issues/1043), issue [#2900](https://github.com/DavidObando/gsharp/issues/2900)
 
 ## Context
 
@@ -194,7 +194,7 @@ kind: it gets a real `case` in `MethodBodyEmitter.EmitStatement` and in
   a nested lambda remains legal because that lambda has its own function body;
   normal fixed-pointer escape rules still prevent capturing the pointer.
 - `fixed` is compiled-only under
-  [ADR-0152](0152-interpreter-compiled-only-storage-boundary.md). `gsi` reports
+  [ADR-0153](0153-interpreter-compiled-only-storage-boundary.md). `gsi` reports
   that pinning requires the CIL pinned-local emit path; it does not attempt to
   emulate pinning without a storage and address model.
 - Deferred: fixed-size buffers.
--- a/docs/adr/0125-fixed-pinned-statement.md
+++ b/docs/adr/0125-fixed-pinned-statement.md
@@ -3,7 +3,7 @@
 - **Status**: Accepted
 - **Date**: 2026-06-27
 - **Phase**: Phase 9 — low-level / interop depth
-- **Related**: ADR-0039 (managed by-ref pointers / address-of / dereference), ADR-0056 (ref-returning members / `modreq(InAttribute)` ref-returns), ADR-0122 (unsafe context and unmanaged raw pointers `*T`, issue [#1014](https://github.com/DavidObando/gsharp/issues/1014)), ADR-0124 (`stackalloc`/`localloc`, issue [#1024](https://github.com/DavidObando/gsharp/issues/1024)), issue [#1026](https://github.com/DavidObando/gsharp/issues/1026), issue [#1043](https://github.com/DavidObando/gsharp/issues/1043), issue [#2900](https://github.com/DavidObando/gsharp/issues/2900)
+- **Related**: ADR-0039 (managed by-ref pointers / address-of / dereference), ADR-0056 (ref-returning members / `modreq(InAttribute)` ref-returns), ADR-0122 (unsafe context and unmanaged raw pointers `*T`, issue [#1014](https://github.com/DavidObando/gsharp/issues/1014)), ADR-0124 (`stackalloc`/`localloc`, issue [#1024](https://github.com/DavidObando/gsharp/issues/1024)), ADR-0152 (interpreter compiled-only storage boundary), issue [#1026](https://github.com/DavidObando/gsharp/issues/1026), issue [#1043](https://github.com/DavidObando/gsharp/issues/1043), issue [#2900](https://github.com/DavidObando/gsharp/issues/2900)
 
 ## Context
 
@@ -193,6 +193,10 @@ kind: it gets a real `case` in `MethodBodyEmitter.EmitStatement` and in
   A state-machine suspension cannot preserve a pinned local. Suspension inside
   a nested lambda remains legal because that lambda has its own function body;
   normal fixed-pointer escape rules still prevent capturing the pointer.
+- `fixed` is compiled-only under
+  [ADR-0152](0152-interpreter-compiled-only-storage-boundary.md). `gsi` reports
+  that pinning requires the CIL pinned-local emit path; it does not attempt to
+  emulate pinning without a storage and address model.
 - Deferred: fixed-size buffers.
 
 ## Diagnostics
--- a/docs/adr/0152-interpreter-compiled-only-storage-boundary.md
+++ b/docs/adr/0152-interpreter-compiled-only-storage-boundary.md
@@ -1,0 +1,98 @@
+# ADR-0152: Interpreter compiled-only storage boundary
+
+- **Status**: Accepted
+- **Date**: 2026-08-01
+- **Phase**: Phase 9 — low-level / interop depth
+- **Related**: ADR-0039 (managed by-ref pointers), ADR-0122 (unsafe context and unmanaged pointers), ADR-0124 (`stackalloc`), ADR-0125 (`fixed`), issues [#2956](https://github.com/DavidObando/gsharp/issues/2956), [#3004](https://github.com/DavidObando/gsharp/issues/3004), and [#2939](https://github.com/DavidObando/gsharp/issues/2939)
+
+## Context
+
+The compiled backend has a CLR storage model. It can emit managed byrefs,
+unmanaged pointers, pinned locals, `localloc`, `sizeof`, `ldftn`, and `calli`.
+The tree-walking interpreter models values, not storage locations. Its local
+frames contain values, and it has no address space, alias identity, pointer
+lifetime, or GC-pinning mechanism.
+
+ADR-0039's interpreter support for `ref` and `out` does not provide a general
+pointer model. `&x` and `*p` are evaluated as identity operations. Ref/out calls
+work only because the call-site machinery recognizes an address-of argument,
+records its source slot, and writes the result back after the call. Outside
+that position, an address is reduced to a value copy. Issue #3004 demonstrates
+the consequence: after `var p *int32 = &x`, a later write to `x` is invisible
+through `*p`, so `gsi` returns a plausible stale value with exit code 0.
+
+`fixed` is the cleanest existing boundary. The evaluator already rejects it
+with a self-contained message explaining that pinning requires the CIL
+pinned-local emit path. `stackalloc`, `sizeof` over unmanaged storage, method
+function pointers, and function-pointer invocation use the same pattern.
+
+Script-mode `gsi` prints only a diagnostic ID and message. It does not render a
+source location or caret. Boundary messages therefore must name the construct,
+state that the interpreter does not support it, and explain which compiled
+runtime facility it requires. They cannot rely on surrounding diagnostic
+rendering for meaning.
+
+## Decision
+
+Constructs that require real address identity, unmanaged pointer operations,
+pinning, stack allocation, or function-pointer execution are **compiled-only**.
+`gsi` must report a self-contained boundary diagnostic instead of attempting a
+value-only approximation. Users who need these constructs must compile with
+`gsc`.
+
+The `unsafe` context itself remains supported. It is a permission boundary, not
+a storage operation; an `unsafe` block containing only otherwise-supported
+value operations continues to evaluate normally.
+
+The current implementation status is:
+
+| Construct | Current `gsi` behavior | Contract status |
+|---|---|---|
+| `unsafe { ... }` without a storage-only construct | Evaluates normally | Supported |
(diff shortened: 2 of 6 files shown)


DavidObando:
## Review: NOT MERGEABLE as-is — 3 blockers (all small fixes)

### Measurements
- `git merge-tree --write-tree 4b3ebf24f 335dde896` → rc=0, clean.
- Scope, measured `MB=$(git merge-base 4b3ebf24f 335dde896)` (`f70626452`) → `git diff $MB 335dde896`: **5 files, +318/−1**. Matches the PR description; `Fixes #2956` is present and the diff is docs+tests only, no production change. Declared scope == actual scope.
- `dotnet build GSharp.sln -c Release -m:1 -p:ContinuousIntegrationBuild=true` → rc=0, `0 Warning(s)`, `0 Error(s)`.
- `Issue2956InterpreterBoundaryTests` → **8 examined, 8 passed**, rc=0.
- Direct `gsi` run of the `fixed` repro from this build: `rc=1`, stderr exactly `GS9999: 'fixed' (pinning) statements are not supported in the interpreter; they require the CIL pinned-local emit path.`

### Every normative claim traced to code — all located
| ADR claim | Code |
|---|---|
| `fixed` rejected by evaluator (:24–27) | `src/Core/CodeAnalysis/Evaluator.Statements.cs:158` |
| `stackalloc` boundary (:53) | `src/Core/CodeAnalysis/Evaluator.Expressions.cs:50` |
| `sizeof` boundary (:54) | `src/Core/CodeAnalysis/Evaluator.Expressions.cs:57` |
| `&Method` boundary (:55) | `src/Core/CodeAnalysis/Evaluator.Expressions.cs:58` |
| fp invocation boundary (:56) | `src/Core/CodeAnalysis/Evaluator.Expressions.cs:59` |
| boundaries surface as `GS9999` (:62–63) | `src/Core/CodeAnalysis/Compilation/Compilation.cs:334` |
| `gsi` prints only ID + message, no location (:29–30) | `src/Repl/Program.cs:60` |
| non-zero exit pinned (:75) | `Compilation.cs:334` (Error) → `SessionEngine.cs:212` → `Program.cs:68` returns 1 — measured rc=1 |

No unlocatable claims. That is a materially better hit rate than the recent ADR PRs.

---

## BLOCKER 1 — ADR number 0152 collides with PR #3024
#3024 adds `docs/adr/0152-interpreter-unsafe-native-boundary.md`; this PR adds `docs/adr/0152-interpreter-compiled-only-storage-boundary.md`. Different filenames, so git merges silently: `git merge-tree --write-tree 335dde896 3c54a2fec` → rc=0, and in the merged tree `ls docs/adr | grep -c '^0152'` → **2**. Highest number on `main` is 0151. Nothing in CI catches this.

**Fix:** renumber one of them (0153) and update the back-references — here that is `docs/adr/0125-fixed-pinned-statement.md:6` and `:196`, plus `docs/diagnostics.md:310`.

## BLOCKER 2 — #3024 falsifies this ADR's premise and breaks 7 of its 8 tests
#3024 replaces `src/Repl/Program.cs:60` (`Console.Error.WriteLine($"{d.Id}: {d.Message}")`) with `Console.Error.WriteDiagnostics(cell.Diagnostics)`.

Measured on a real merge of both heads (build rc=0, `0 Warning(s)`, `0 Error(s)`):
- this PR alone: `Issue2956InterpreterBoundaryTests` **8 examined, 8 passed**
- this PR + #3024: **8 examined, 7 failed, 1 passed**

```
Expected: "GS9999: 'fixed' (pinning) statements are "...
Actual:   "\n(1,1,4,1): error GS9999: 'fixed' (pinnin"...
```

`AssertBoundary` at `test/Interpreter.Tests/Issue2956InterpreterBoundaryTests.cs:159` asserts complete stderr, so it is maximally sensitive to the renderer. Beyond the tests, ADR lines **29–33** ("It does not render a source location or caret... They cannot rely on surrounding diagnostic rendering for meaning") and Consequence line **75–76** become false sentences in an Accepted ADR the moment #3024 lands.

**Fix (smallest):** either land ahead of #3024 and have #3024 update these tests + lines 29–33, or make `AssertBoundary` assert `Assert.Contains($"GS9999: {message}", stderr)` instead of `Assert.Equal`, and soften :29–33 to "does not currently render a source location". The `Assert.Equal(1, ExitCode)` and empty-stdout assertions can stay exact — both survive #3024 (verified: merged `gsi` still returns rc=1).

## BLOCKER 3 — a storage ADR that omits the live ref-local defect (#3022)
The status table (:49–60) is introduced as "The current implementation status is" and enumerates the violations, but has **no row for ref locals**, which are wrong today and silent:
- `src/Core/CodeAnalysis/Evaluator.Statements.cs:203` stores `new RefAlias(refAddr.Operand)`
- `src/Core/CodeAnalysis/Evaluator.Expressions.cs:330-336` **re-evaluates the operand on every read** — the comment says so outright: "reads re-evaluate the aliased operand"

So `let ref r = arr[i]` then `i = 2` reads the wrong element, exit 0. Consequence :73–74 ("Existing ref/out call-site write-back remains supported") is narrowly true for *call-site* write-back, but in a table-plus-consequences document it reads as an all-clear for ref aliasing — and that is exactly the sentence someone will later cite. #3028's own description makes the same point: "Scope must not be framed only as `fixed`/`&`/`*`, because ref locals share the limitation."

**Fix:** one table row — `| ref local alias (`let ref r = arr[i]`) | Re-evaluates the initializer at read time; silently wrong | Must become a boundary; tracked by #3022 |` — and qualify :74 with "call-site".

---

## SHOULD-FIX
- **:57–59 and :62–63 go stale the day #3028 lands.** #3028 allocates a dedicated code (`GS0513`) for `&`/`*` and turns rows :57–:59 into that diagnostic. The "Until the evaluator has a dedicated boundary diagnostic code" hedge at :62 survives, but row :59's "Silently returns a copied, stale value" will be wrong. Point rows :57–:59 at #3028.

## NIT
- `docs/diagnostics.md:310` and `website/docs/ref/diagnostics.md:302` add prose only, **zero** `| GSxxxx |` table rows, so `DiagnosticIdUniquenessTests.Every_Descriptor_Is_Documented_With_Matching_Severity` is untouched. Correct call — no ID was allocated.

## Correction to the ground truth this was reviewed against
`gsi` exits **1**, not 0, when printing `GS9999` from the evaluator boundary path. Measured both directions: `Compilation.cs:334` builds the diagnostic with `DiagnosticSeverity.Error` → `SessionEngine.cs:212` `hasError` → `Program.cs:68` `return cell.HasError ? 1 : 0`; direct `gsi` run gives `rc=1`. The "exits 0" behaviour belongs to the *silent* #3004/#3022 cases, which emit no `GS9999` at all. This PR's `Assert.Equal(1, result.ExitCode)` is correct.

DavidObando:
## Re-review of `d3f189163`: NOT MERGEABLE — **all three original blockers are closed**, but `main` moved underneath this PR and 4 normative sentences are now false

The adoption work is good and I verified every one of my three blockers is genuinely closed, by measurement. The problem is new: **#3028 merged to `main` as `445d583ec` while this was in review**, and it turned four of this ADR's normative sentences into false statements. Both remaining blockers are doc text only — roughly six lines.

All line numbers are read in the tree `main 4bba0643a` + `d3f189163` unless stated.

### Verification of the three original blockers — all CLOSED

**BLOCKER 1 (ADR-0152 collision) — closed.** I re-ran the count that found it, on a real co-merged tree (`main 8acc2cc76` + #3018 + #3024 `d1e47bd34`):

```
count_0152=1  count_0153=1
```

Last time that first number was **2**. I also enumerated **all 27 open PRs** and checked each one's file list: exactly two claim an ADR number — #3024 → `0152`, #3018 → `0153`. No third claimant. `main` tops out at `0151`, so `0153` collides with nothing that arrived since.

Dangling references: `git grep '0152' d3f189163 -- docs/` returns **10 hits, and every one is `GS0152`**, the constraint-satisfaction diagnostic — a substring, not an ADR reference. I proved the pattern before trusting it by locating the real reference form, which is present and correctly updated at `docs/adr/0125-fixed-pinned-statement.md:6`, `:197` and `docs/diagnostics.md:312`. No dangling ADR-0152 reference.

**BLOCKER 2 (co-merge failure) — closed, and not vacuously.** #3018 alone: **8/8**. `main` + #3018 + #3024: **8/8**. Previously this was **7 failed in either order**.

I checked the result is real rather than an accident of the merge dropping #3024's change. In the co-merged tree `src/Repl/Program.cs` reads `Console.Error.WriteDiagnostics(cell.Diagnostics)`, against `main`'s `foreach (var d in cell.Diagnostics) Console.Error.WriteLine($"{d.Id}: {d.Message}")`. The renderer swap is present and the tests pass against it.

**BLOCKER 3 (ref-local row) — present and true.** `Evaluator.Statements.cs:203` is `var alias = new RefAlias(refAddr.Operand);`, and `Evaluator.Expressions.cs:332-335` is `if (stored is RefAlias alias) { return EvaluateExpression(alias.Operand); }` — re-evaluation on every read. Probe on this tree: `let ref r = arr[i]` with `i = 0`, then `i = 2`, read → **`30`, rc=0**. The row's description column is accurate.

**"No production code" — verified independently.** `MB=$(git merge-base origin/main d3f189163)` → `428f31779`; `git diff $MB d3f189163 --stat` → **5 files, +323/−1**: the ADR, `docs/adr/0125-*.md`, `docs/diagnostics.md`, `website/docs/ref/diagnostics.md`, and one test file. No `src/` path. Not a rebase-only push either — blob SHAs moved (`docs/adr/…` `6287e7c8` → `5a8c51b8`, tests `87bffdde` → `72684637`).

**Docs gate — clean.** The `docs/diagnostics.md` +10 and `website/docs/ref/diagnostics.md` +9 are prose paragraphs under a `####` heading. **Zero `| GSxxxx |` rows added and zero deleted.** No backticked-ID row. No new diagnostic ID allocated.

**Merge and build against current `main` (`4bba0643a`)**: `git merge-tree` rc=0; build rc=0, `0 Warning(s)`, `0 Error(s)`; `Issue2956InterpreterBoundaryTests` **8/8** on post-#3028 `main` — the 8 tests survive #3028 landing, which was worth checking since #3028 added throws to `EvaluateAddressOfExpression` / `EvaluateDereferenceExpression`.

---

### BLOCKER A — four normative sentences are false against `main` today

#3028 merged as `445d583ec` and now refuses unmanaged pointer operations with `GS0513`. Measured on `main 4bba0643a` + #3018:

| ADR text | What it says | What `gsi` actually does |
|---|---|---|
| `:22` | "after `var p *int32 = &x`, a later write to `x` is invisible through `*p`, so `gsi` returns a plausible stale value **with exit code 0**" | `GS0513: Unmanaged pointer operations…`, **rc=1** |
| `:61` | `*p = value` — "falls through to a generic 'Unexpected node' failure"; status "Must be a boundary" | `GS0513`, **rc=1** — it already *is* a boundary |
| `:62` | `*(p + 1)` — "leaks a raw reflection/conversion failure" | unreachable: `fixed` refuses first (`GS9999`, rc=1), `&` refuses (`GS0513`, rc=1) |
| `:63` | free-standing `&x` then `*p` — "silently returns a copied, stale value"; status "Must be a boundary" | `GS0513`, **rc=1** — already a boundary |

`:22` is the paragraph the whole Context section rests on, and its headline example no longer reproduces. Three of the eleven table rows describe a defect that was fixed before this PR could land, and label as "Must be a boundary" three things that now *meet* the contract.

Same staleness in `docs/diagnostics.md:310-316`, which says the boundary sites "currently surface through `GS9999`" — pointer operations now surface through `GS0513`.

**Fix:** rewrite `:22` to say the pointer case is now enforced by `GS0513` (#3028), and flip rows `:61`–`:63` to "Boundary enforced via `GS0513`" / "Meets contract". Same one-sentence correction in `docs/diagnostics.md:310-316`.

*(My first attempt at the `*(p + 1)` probe used the wrong `fixed` syntax and produced `GS0005` parse errors — that was an unrun check, not a result. The row above uses the syntax taken from this PR's own `Issue2956InterpreterBoundaryTests.cs:28`.)*

### BLOCKER B — `:60` contradicts this ADR's own Decision, and contradicts PR #3032

The Decision at `:36-40` scopes the compiled-only rule precisely:

> Constructs whose interpreter behavior requires real address identity — including **pinning, stack allocation, unmanaged-pointer storage or dereference, and function-pointer execution** — are **compiled-only**.

Ref-local aliasing is none of those four. It is a *managed* by-ref, supported by ADR-0039 and by the compiled backend. Yet `:60` gives it contract status **"Must become a boundary; tracked by #3022"**.

That is not what is happening to #3022. PR #3032 (`Fixes #3022`, open) makes `let ref r = arr[i]` **work** — I measured it on a built tree: read → `10`, write → `99,20,30`, both truth. It captures the storage location; it does not refuse the construct.

So this cell mandates deleting a fix that is queued to land, in a document with status **Accepted**. That is precisely the failure mode this repo has already been bitten by — a normative sentence later cited to justify removing real behaviour. It is also internally inconsistent with `:36-40` on its own terms.

**Fix:** one cell. `Must become a boundary; tracked by #3022` → `Must alias a captured storage location; fixed by #3032 (issue #3022)`.

### SHOULD-FIX — the assertion decoupled from the renderer but not from the diagnostic ID

`Issue2956InterpreterBoundaryTests.cs:159` is now `Assert.Contains($"GS9999: {message}", result.StandardError)`. The renderer coupling is genuinely gone (proven by the 8/8 co-merge above), but the assertion still pins the **diagnostic ID** — and this ADR itself calls that ID "interim" at `:66-72`, while `GS0514`/`GS0515` are already issued to #3024.

The co-merge passes today only because #3024 has not made that change yet. Proven rather than assumed: at #3024 head `d1e47bd34`, `GS051[0-9]` matches **0** times in `DiagnosticDescriptors.cs` and `docs/diagnostics.md`; the control `GS050[0-9]` matches **10** and **18** times respectively, so the pattern works. When #3024 allocates those IDs, these tests break exactly as they did before — same cycle, new cause.

**Fix:** assert on `message` alone and drop the `GS9999: ` prefix. The ID is the part the project has already decided to change.

### NIT
- `:6` **Related** lists #2956, #3004, #3022, #2939 but not #3028 or #3032, even though rows `:61`–`:63` cite #3028 by number.

### Summary

| | |
|---|---|
| BLOCKER A | ADR `:22`, `:61`, `:62`, `:63` and `docs/diagnostics.md:310-316` — false since #3028 merged (`445d583ec`). All four refuted by probe: `GS0513`, rc=1. |
| BLOCKER B | ADR `:60` — "Must become a boundary" contradicts `:36-40` and contradicts PR #3032, which makes it work (`10` / `99,20,30`). |
| SHOULD-FIX | `Issue2956InterpreterBoundaryTests.cs:159` — still pins `GS9999`, which `:66-72` calls interim and #3024 will change. |
| NIT | ADR `:6` Related omits #3028/#3032. |

Nothing here needs code. Six lines of doc text and one assertion, and this is mergeable.

PR #3019, opened 2026-08-01
Preserve delegate identity through erased lambda slots

### LINKED ISSUE #2937: gsc: mismatched inline-lambda arity through erased imported-generic slot compiles then throws

### LINKED ISSUE #2948: gsc: preserve imported named-delegate identity through deferred erased-slot lambda targeting

DavidObando:
## Summary
- carry exact mapped delegate identity through deferred method and constructor lambda targeting
- normalize `Expression<TDelegate>` to `TDelegate` before candidate identity comparison, preserving valid LINQ `Enumerable.Func` / `Queryable.Expression<Func>` agreement
- convert to nominal delegates only when the mapped target is fully closed
- recognize BCL `Action`/`Func` by exact type identity from the delegate's actual `System.MulticastDelegate` assembly, not a `FullName` prefix or the host runtime assembly
- keep user-assembly `System.Func` and nominally distinct `Predicate<T>` / `Func<T,bool>` delegates distinct

## Regressions
The G11 failure was candidate identity, not a transparent-identifier special case. These translated constructs now compile:

| Construct | Kind | `f5e9d8925` base | pre-fix PR head | final |
|---|---|---:|---:|---:|
| `CorpusJoinLetConstruct` | `LetConstruct` | 0 errors | 1 of 3 corpus errors | 0 errors |
| `CorpusJoinResultSelectorLetConstruct` | `LetConstruct` | 0 errors | 1 of 3 corpus errors | 0 errors |
| `CorpusJoinChainedSelectAndWriteLine` | `GSharpConstruct` | 0 errors | 1 of 3 corpus errors | 0 errors |

A fresh solution-built driver also rebuilds translated `Oahu.Decrypt` with 0 warnings / 0 errors; the old host-assembly gate produced 25 GS0155/GS0156/GS0159 errors.

## Scope matrix
The new G11, Oahu-shaped LINQ, and nominal identity tests cover `{top-level, in-function}` × `{gsc, gsi}`. Interpreter in-function cases define `Run()` and invoke it from top level; the sources are not all wrapped in an uncalled `func Main()`.

## Mutations
All product mutants were rebuilt and killed after the final completed rebase:

| Product mutation | Compiler | Interpreter |
|---|---:|---:|
| remove expression-tree unwrapping | 4/6 failed | 3/6 failed |
| invert method identity agreement | 2/2 failed | 2/2 failed |
| invert constructor identity agreement | 2/2 failed | 2/2 failed |
| invert symbolic closed-target conjunct | 2/2 failed | n/a |
| invert CLR closed-target conjunct | 2/2 failed | n/a |
| restore host-assembly BCL identity gate | 2/2 failed | n/a |

Every restore rebuilt successfully and returned to source blob `91734065...` and Core DLL MD5 `d9808715a4562ca3418d6bd8f88fa180`.

## Verification
Measurements are true of merged-tree anchor `402b85615` plus head `7b1802199`:

- `Compiler.Tests`: **4163 passed / 0 failed / 4163 total**
- `Interpreter.Tests`: **726 passed / 0 failed / 726 total**
- targeted regression matrix: compiler **14/14**, interpreter **12/12**
- bidirectional diagnostics/docs gate: **4/4**
- Release solution build: **0 warnings / 0 errors**
- fresh `Core/` and driver `Compiler/` DLLs: identical mtime and MD5 `d9808715a4562ca3418d6bd8f88fa180`
- G11 positive control (`MissingLength`): build fails; exact restore: **0 errors**
- Oahu positive control (`MissingRenderSize`): build fails; exact restore: **0 errors**

`main` advanced to `4469e3237` via #3120 while the 1h17m full suite was running. `git merge-tree` against that tip is conflict-free, with no overlap in this PR's four changed files; the test shelf SHA is stated above rather than mislabelled as the newer tree.

Fixes #2948
Fixes #2937

## CI
GitHub Actions run [30776786449](https://github.com/DavidObando/gsharp/actions/runs/30776786449) is fully green against PR base `4469e3237`: build, e2e, both cs2gs corpus gates (including Oahu), all test shards, and extensions.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
@@ -815,12 +815,14 @@ internal sealed partial class ExpressionBinder
         return true;
     }
 
+    private static TypeSymbol UnwrapExpressionTreeDelegate(TypeSymbol type) =>
+        MemberLookup.TryGetExpressionTreeDelegateTypeFromSymbol(type, out var delegateType)
+            ? delegateType
+            : type;
+
     private static bool IsCanonicalFunctionDelegate(TypeSymbol type)
     {
-        if (MemberLookup.TryGetExpressionTreeDelegateTypeFromSymbol(type, out var delegateType))
-        {
-            type = delegateType;
-        }
+        type = UnwrapExpressionTreeDelegate(type);
 
         if (type is ImportedTypeSymbol imported)
         {
@@ -854,10 +856,7 @@ internal sealed partial class ExpressionBinder
             return false;
         }
 
-        if (MemberLookup.TryGetExpressionTreeDelegateTypeFromSymbol(type, out var delegateType))
-        {
-            type = delegateType;
-        }
+        type = UnwrapExpressionTreeDelegate(type);
 
         return !TypeSymbol.ContainsTypeParameter(type)
             && type?.ClrType?.ContainsGenericParameters != true;
@@ -865,6 +864,9 @@ internal sealed partial class ExpressionBinder
 
     private static bool SameDelegateIdentity(TypeSymbol left, TypeSymbol right)
     {
+        left = UnwrapExpressionTreeDelegate(left);
+        right = UnwrapExpressionTreeDelegate(right);
+
         if (ReferenceEquals(left, right))
         {
             return true;
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
@@ -814,25 +814,6 @@ internal sealed partial class ExpressionBinder
         return true;
     }
 
-    private static bool IsCanonicalFunctionDelegate(TypeSymbol type)
-    {
-        if (MemberLookup.TryGetExpressionTreeDelegateTypeFromSymbol(type, out var delegateType))
-        {
-            type = delegateType;
-        }
-
-        if (type is not ImportedTypeSymbol imported)
-        {
-            return true;
-        }
-
-        var openDefinition = imported.OpenDefinition ?? imported.ClrType;
-        var fullName = openDefinition?.FullName;
-        return fullName == "System.Action"
-            || fullName?.StartsWith("System.Action`", StringComparison.Ordinal) == true
-            || fullName?.StartsWith("System.Func`", StringComparison.Ordinal) == true;
-    }
-
     /// <summary>
     /// Issue #1512 / #2365: builds a symbolic <see cref="FunctionTypeSymbol"/> for
     /// a lambda argument bound to a CLR/imported method whose delegate (or
@@ -1048,7 +1029,14 @@ internal sealed partial class ExpressionBinder
                         // void-discard / ordinary target-typed return-type
                         // inference applies instead of inferring the return
                         // type purely from the lambda body.
-                        boundArgs[idx] = lambdas.BindLambdaExpression(lambdaSyntax, target, inferReturnTypeFromBody: !exactReturnIndices.Contains(idx));
+                        var literal = lambdas.BindLambdaExpression(
+                            lambdaSyntax,
+                            target.FunctionType,
+                            inferReturnTypeFromBody: !exactReturnIndices.Contains(idx));
+                        boundArgs[idx] = conversions.BindConversion(
+                            ce.Arguments[idx].Location,
+                            literal,
+                            target.DelegateType);
                     }
                 }
 
@@ -1339,11 +1327,11 @@ internal sealed partial class ExpressionBinder
     /// each deferred lambda's delegate <em>parameter</em> types through those
     /// inferred symbolic arguments (reusing
     /// <see cref="MemberLookup.MapOpenClrTypeToSymbolic(Type, Type, ImmutableArray{TypeSymbol}, MethodInfo, ImmutableArray{TypeSymbol})"/>).
-    /// A per-slot <see cref="FunctionTypeSymbol"/> carrying those parameter
-    /// types is produced. Non-generic methods on constructed generic receivers
-    /// are recovered from the open declaring type too, including delegates
-    /// whose only symbolic slot is their return type. Candidates must agree on
-    /// the recovered shape.
+    /// A per-slot target carrying both the exact mapped delegate identity and
+    /// its <see cref="FunctionTypeSymbol"/> shape is produced. Non-generic
+    /// methods on constructed generic receivers are recovered from the open
+    /// declaring type too, including delegates whose only symbolic slot is
+    /// their return type. Candidates must agree on both identity and shape.
     ///
     /// Issue #2345: also recovers the delegate's <em>return</em> shape when it
     /// is fully closed by the same unification (including the common case of a
@@ -1367,7 +1355,7 @@ internal sealed partial class ExpressionBinder
         List<int> deferredIndices,
         BoundExpression[] boundArgs,
         HashSet<int> deferred,
-        out Dictionary<int, FunctionTypeSymbol> targets,
+        out Dictionary<int, (TypeSymbol DelegateType, FunctionTypeSymbol FunctionType)> targets,
         out HashSet<int> exactReturnIndices)
     {
         exactReturnIndices = new HashSet<int>();
@@ -1422,7 +1410,7 @@ internal sealed partial class ExpressionBinder
             arityByIndex[idx] = lambda.Parameters.Count;
         }
 
-        Dictionary<int, ImmutableArray<TypeSymbol>> agreed = null;
+        Dictionary<int, (TypeSymbol DelegateType, ImmutableArray<TypeSymbol> ParameterTypes)> agreed = null;
         Dictionary<int, TypeSymbol> agreedReturnTypes = null;
         var anySymbolicType = false;
 
@@ -1473,7 +1461,7 @@ internal sealed partial class ExpressionBinder
                     : MemberLookup.InferSymbolicMethodTypeArguments(openMethod, symbolicArgVector);
             var methodTypeArgs = ImmutableArray.Create(inferred);
 
-            var slotTargets = new Dictionary<int, ImmutableArray<TypeSymbol>>();
+            var slotTargets = new Dictionary<int, (TypeSymbol DelegateType, ImmutableArray<TypeSymbol> ParameterTypes)>();
             var slotReturnTypes = new Dictionary<int, TypeSymbol>();
             var candidateUsable = true;
             var candidateHasSymbolicType = false;
@@ -1488,6 +1476,7 @@ internal sealed partial class ExpressionBinder
                 }
 
                 var openParameterType = openParameters[paramIndex].ParameterType;
+                TypeSymbol mappedDelegate;
                 FunctionTypeSymbol functionType;
                 if (openParameterType.IsGenericParameter)
                 {
@@ -1495,14 +1484,8 @@ internal sealed partial class ExpressionBinder
                             openParameterType,
                             receiverOpenDefinition,
                             receiverTypeArguments,
-                            out var mappedDelegate,
+                            out mappedDelegate,
                             out functionType))
-                    {
-                        candidateUsable = false;
-                        break;
-                    }
-
-                    if (!IsCanonicalFunctionDelegate(mappedDelegate))
                     {
                         candidateUsable = false;
                         break;
@@ -1587,6 +1570,18 @@ internal sealed partial class ExpressionBinder
                         break;
                     }
 
+                    mappedDelegate = MemberLookup.MapOpenClrTypeToSymbolic(
+                        openParameterType,
+                        receiverOpenDefinition,
+                        receiverTypeArguments,
+                        openMethod,
+                        methodTypeArgs);
+                    if (mappedDelegate == null || mappedDelegate == TypeSymbol.Error)
+                    {
+                        candidateUsable = false;
+                        break;
+                    }
+
                     var returnClrType = invoke.ReturnType;
                     var mappedReturnType = returnClrType != null && returnClrType.IsSameAs(typeof(void))
                         ? TypeSymbol.Void
@@ -1607,7 +1602,7 @@ internal sealed partial class ExpressionBinder
                     }
                 }
 
-                slotTargets[idx] = functionType.ParameterTypes;
+                slotTargets[idx] = (mappedDelegate, functionType.ParameterTypes);
 
                 // Issue #2345: recover the delegate's return shape too, when it
                 // is fully closed by this unification (most commonly `void`,
@@ -1637,7 +1632,7 @@ internal sealed partial class ExpressionBinder
                 agreed = slotTargets;
                 agreedReturnTypes = slotReturnTypes;
             }
-            else if (!SymbolicLambdaParameterTypesAgree(agreed, slotTargets))
+            else if (!SymbolicLambdaTargetsAgree(agreed, slotTargets))
             {
                 return false;
             }
@@ -1668,7 +1663,7 @@ internal sealed partial class ExpressionBinder
             return false;
         }
 
-        var built = new Dictionary<int, FunctionTypeSymbol>();
+        var built = new Dictionary<int, (TypeSymbol DelegateType, FunctionTypeSymbol FunctionType)>();
         foreach (var kv in agreed)
         {
             if (agreedReturnTypes != null && agreedReturnTypes.TryGetValue(kv.Key, out var exactReturn))
@@ -1679,14 +1674,18 @@ internal sealed partial class ExpressionBinder
                 // inferReturnTypeFromBody, letting InferLambdaReturnType's
                 // void-discard rule (issue #889) and ordinary target-typed
                 // inference apply.
-                built[kv.Key] = FunctionTypeSymbol.Get(kv.Value, exactReturn);
+                built[kv.Key] = (
+                    kv.Value.DelegateType,
+                    FunctionTypeSymbol.Get(kv.Value.ParameterTypes, exactReturn));
                 exactReturnIndices.Add(kv.Key);
             }
             else
             {
                 // The return slot is a placeholder; callers bind with
                 // inferReturnTypeFromBody so the lambda infers its own return type.
-                built[kv.Key] = FunctionTypeSymbol.Get(kv.Value, TypeSymbol.Object);
+                built[kv.Key] = (
+                    kv.Value.DelegateType,
+                    FunctionTypeSymbol.Get(kv.Value.ParameterTypes, TypeSymbol.Object));
             }
         }
 
@@ -1694,7 +1693,9 @@ internal sealed partial class ExpressionBinder
         return true;
     }
 
-    private static bool SymbolicLambdaParameterTypesAgree(Dictionary<int, ImmutableArray<TypeSymbol>> a, Dictionary<int, ImmutableArray<TypeSymbol>> b)
+    private static bool SymbolicLambdaTargetsAgree(
+        Dictionary<int, (TypeSymbol DelegateType, ImmutableArray<TypeSymbol> ParameterTypes)> a,
+        Dictionary<int, (TypeSymbol DelegateType, ImmutableArray<TypeSymbol> ParameterTypes)> b)
     {
         if (a.Count != b.Count)
         {
@@ -1703,14 +1704,16 @@ internal sealed partial class ExpressionBinder
 
         foreach (var kv in a)
         {
-            if (!b.TryGetValue(kv.Key, out var other) || other.Length != kv.Value.Length)
+            if (!b.TryGetValue(kv.Key, out var other)
+                || !Equals(kv.Value.DelegateType, other.DelegateType)
+                || other.ParameterTypes.Length != kv.Value.ParameterTypes.Length)
             {
                 return false;
             }
 
-            for (var i = 0; i < kv.Value.Length; i++)
+            for (var i = 0; i < kv.Value.ParameterTypes.Length; i++)
             {
-                if (!Equals(kv.Value[i], other[i]))
+                if (!Equals(kv.Value.ParameterTypes[i], other.ParameterTypes[i]))
                 {
                     return false;
                 }
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
@@ -14,6 +14,7 @@ using System.Collections.Immutable;
 using System.Linq;
 using System.Reflection;
 using System.Text;
+using GSharp.Core.CodeAnalysis.Emit;
 using GSharp.Core.CodeAnalysis.Lowering;
 using GSharp.Core.CodeAnalysis.Lowering.Async;
 using GSharp.Core.CodeAnalysis.Symbols;
@@ -814,6 +815,93 @@ internal sealed partial class ExpressionBinder
         return true;
     }
 
+    private static bool IsCanonicalFunctionDelegate(TypeSymbol type)
+    {
+        if (MemberLookup.TryGetExpressionTreeDelegateTypeFromSymbol(type, out var delegateType))
+        {
+            type = delegateType;
+        }
+
+        if (type is ImportedTypeSymbol imported)
+        {
+            var definition = imported.OpenDefinition ?? imported.ClrType;
+            if (definition != null
+                && AssemblyName.ReferenceMatchesDefinition(
+                    definition.Assembly.GetName(),
+                    typeof(Action).Assembly.GetName()))
+            {
+                var coreLibrary = definition.Assembly;
+                var arity = definition.IsGenericTypeDefinition
+                    ? definition.GetGenericArguments().Length
+                    : 0;
+                if (definition == coreLibrary.GetType("System.Action")
+                    || (arity > 0
+                        && (definition == coreLibrary.GetType($"System.Action`{arity}")
+                            || definition == coreLibrary.GetType($"System.Func`{arity}"))))
+                {
+                    return true;
+                }
+            }
+        }
+
+        return false;
+    }
+
+    private static bool ShouldConvertToNominalDelegate(TypeSymbol type)
+    {
+        if (IsCanonicalFunctionDelegate(type))
+        {
+            return false;
+        }
+
+        if (MemberLookup.TryGetExpressionTreeDelegateTypeFromSymbol(type, out var delegateType))
+        {
+            type = delegateType;
+        }
+
+        return !TypeSymbol.ContainsTypeParameter(type)
+            && type?.ClrType?.ContainsGenericParameters != true;
+    }
+
+    private static bool SameDelegateIdentity(TypeSymbol left, TypeSymbol right)
+    {
+        if (ReferenceEquals(left, right))
+        {
+            return true;
+        }
+
+        if (left is not ImportedTypeSymbol leftImported
+            || right is not ImportedTypeSymbol rightImported)
+        {
+            return Equals(left, right);
+        }
+
+        var leftDefinition = leftImported.OpenDefinition ?? leftImported.ClrType;
+        var rightDefinition = rightImported.OpenDefinition ?? rightImported.ClrType;
+        if (!TypeIdentityComparer.Instance.Equals(leftDefinition, rightDefinition)
+            || leftImported.TypeArguments.Length != rightImported.TypeArguments.Length)
+        {
+            return false;
+        }
+
+        if (IsCanonicalFunctionDelegate(left) && IsCanonicalFunctionDelegate(right))
+        {
+            return true;
+        }
+
+        for (var i = 0; i < leftImported.TypeArguments.Length; i++)
+        {
+            if (!SameDelegateIdentity(
+                    leftImported.TypeArguments[i],
+                    rightImported.TypeArguments[i]))
+            {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
     /// <summary>
     /// Issue #1512 / #2365: builds a symbolic <see cref="FunctionTypeSymbol"/> for
     /// a lambda argument bound to a CLR/imported method whose delegate (or
@@ -1033,10 +1121,12 @@ internal sealed partial class ExpressionBinder
                             lambdaSyntax,
                             target.FunctionType,
                             inferReturnTypeFromBody: !exactReturnIndices.Contains(idx));
-                        boundArgs[idx] = conversions.BindConversion(
-                            ce.Arguments[idx].Location,
-                            literal,
-                            target.DelegateType);
+                        boundArgs[idx] = ShouldConvertToNominalDelegate(target.DelegateType)
+                            ? conversions.BindConversion(
+                                ce.Arguments[idx].Location,
+                                literal,
+                                target.DelegateType)
+                            : literal;
                     }
                 }
 
@@ -1705,7 +1795,7 @@ internal sealed partial class ExpressionBinder
         foreach (var kv in a)
         {
             if (!b.TryGetValue(kv.Key, out var other)
-                || !Equals(kv.Value.DelegateType, other.DelegateType)
+                || !SameDelegateIdentity(kv.Value.DelegateType, other.DelegateType)
                 || other.ParameterTypes.Length != kv.Value.ParameterTypes.Length)
             {
                 return false;
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
@@ -827,19 +827,27 @@ internal sealed partial class ExpressionBinder
         if (type is ImportedTypeSymbol imported)
         {
             var definition = imported.OpenDefinition ?? imported.ClrType;
-            if (definition != null
-                && AssemblyName.ReferenceMatchesDefinition(
-                    definition.Assembly.GetName(),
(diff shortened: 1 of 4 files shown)


DavidObando:
## Verdict: mechanism is right and well-located. **One BLOCKER** — reachable new logic with zero test coverage.

Scratch clone, `10aa0ba9b`. Merge-base **is** `f70626452` (0 behind / 1 ahead), `merge-tree` rc=0, solution build rc=**0** (56 s). Content is real, not a rebase: 3 files, **+217/−82**.

### Base→head differential (measured, not read)

Head's **tests** against base's **production** code — 8 pre-existing pass on both, **all 4 new tests fail on base and pass on head**:

| test | base failure |
|---|---|
| `MismatchedLambdaArityThroughErasedDelegateSlot_IsRejected` | `Assert.NotEqual` — **0 diagnostics**, i.e. base *compiles* it. Matches #2937 verbatim. |
| `SameFullNameDelegateFromUserAssembly_RetainsAssemblyIdentity` | `dotnet exec exited 134` — `ArgumentException: "System.Func`3[Src,Src,Int32]" is not of type "System.Func`2[Src,Int32]"` |
| `ImportedNamedDelegatesThroughErasedListSlots_VerifyLoadAndRun` | `GS0159` / `GS0158` (#2929's conservative rejection) |
| `ImportedNamedDelegatesThroughErasedConstructorSlots_VerifyLoadAndRun` | same |

The 134 is the point. `List<T>.Add`'s runtime type check is a real reflection boundary that **enforces exact identity**, so erasure is genuinely observable here. The boundary is correctly located.

The deleted `IsCanonicalFunctionDelegate` was **name-based** (`StartsWith("System.Func\`")`), so the `AssemblyCollisionContracts` fixture — a user-defined `System.Func<T,TResult>` colliding with CoreLib on full name *and* arity but differing in signature — is precisely the case it mis-classified as canonical. Good fixture.

**Scope checked and clean.** Inverting three `_IsRejected` tests is the declared #2948 fix (the issue explicitly asks to replace #2929's conservative rejection). No undeclared scope found.

### Mutation

- **M1** — drop the `BindConversion` to `target.DelegateType` (`ExpressionBinder.Calls.Invocation.cs:1032`): **KILLED**, 3 tests fail. Covered.
- **M2** — drop `!Equals(kv.Value.DelegateType, other.DelegateType)` from `SymbolicLambdaTargetsAgree` (`ExpressionBinder.Calls.Invocation.cs:1710`): **SURVIVED 12/12.**
  Round-trip md5 proves the mutant was genuinely compiled in: mutant `86c08124325f13656397fe02ce6b8b68` → restored `581e99c92c9a9e22ff63c8c0da0a8a4a`.

## BLOCKER — the delegate-identity agreement check is reachable, load-bearing, and untested

Survival is only coverage, so I probed reachability directly. Contracts with an overload set that shares a **shape** but differs in **delegate identity**:

```csharp
public sealed class Box<T>
{
    public void Add(Predicate<T> p) { Console.WriteLine("pred"); }
    public void Add(Func<T, bool> f) { Console.WriteLine("func"); }
}
```

called as `box.Add((item) -> item.N > 2)`:

- **PR as written** — rejected: `GS0159` + `GS0304` + `GS0158`. The agreement check bails out, deferring to the ordinary erased path.
- **With M2** — **compiles and runs, printing `pred`.** The compiler silently picks one of two equally-viable candidates and emits that delegate identity.

So the check is doing real work: without it, an ambiguous overload set silently resolves to an arbitrary delegate identity — **the exact #2948 defect class this PR exists to eliminate**, just reached through ambiguity rather than erasure. No test in the suite can see it.

**Fix:** add a test pinning the ambiguous-overload case. `CompileExpectingFailure` currently hardcodes `Contracts` (`Issue2918InlineLambdaErasedReceiverTests.cs:889`), so it needs a `contractsSource` parameter like `CompileVerifyLoadAndRun` already has.

**Same class, not separately measured:** `TryResolveSymbolicDelegateTargetForCtor` has the structurally parallel check `!Equals(target.DelegateType, mappedDelegate)` (`ExpressionBinder.Calls.cs:1517`). I proved the invocation-path check only; I did not mutate this one. It is equally untested and should get the same coverage.

## SHOULD-FIX — #2937 is fixed, but not with the diagnostic it asked for

#2937 asks the compiler to reject *"with an arity/delegate-conversion diagnostic"*. Delivered is `GS0159` "Cannot find function Add" plus `GS0155` "Cannot convert type **'object'** to 'Src'".

Rejection-before-emit is achieved, so the defect is closed. But `GS0155` leaks the erased `object` slot into the user-facing message — the very abstraction this PR otherwise hides — and neither diagnostic mentions arity. A user with a 2-param lambda against a 1-param delegate is told the method does not exist.

## NIT — contested file

`ExpressionBinder.Calls.Invocation.cs` is also touched by open **#2903**. #2929 is merged, so this is a 2-way not 3-way, but worth sequencing.

## Summary

| | |
|---|---|
| Merges/builds on `f70626452` | clean, rc=0 |
| Real content | yes, +217/−82 |
| Base→head differential | proven, 4/4 |
| Undeclared scope | none found |
| Mutation | 1 killed, 1 survived → **reachable and untested** |

Fix the BLOCKER and this is good work — the boundary is correctly identified and the fixture is genuinely clever.

DavidObando:
## Verdict: **NOT MERGEABLE — 1 BLOCKER (new).** The round-1 blocker is genuinely closed; a larger one has replaced it.

Reviewed at head `d4d6dd7e4`, merge base **`7a6307a79`** (computed against a freshly fetched remote ref), `main` = `8acc2cc76`.

### Round-1 blocker: CLOSED. Both identity checks are now pinned.

I re-ran both mutations myself rather than accepting the report. Anchors are named in **my merged tree** (`/tmp/d3019` = `d4d6dd7e4` + `main`), because the line numbers in the brief are 2 off here.

| mutant | site | result |
|---|---|---|
| **M1708** — drop `\|\| !Equals(kv.Value.DelegateType, other.DelegateType)` | `ExpressionBinder.Calls.Invocation.cs:1708` | **KILLED** by `NominalDelegateDisagreement_KeepsMethodOverloadAmbiguous` — *"Expected ambiguous delegate overloads to be rejected, but the program ran with output: `pred`"* |
| **M1519** — drop `!Equals(target.DelegateType, mappedDelegate) \|\|` | `ExpressionBinder.Calls.cs:1519` | **KILLED** by `NominalDelegateDisagreement_PreservesConstructorOverloadResolution` — `Expected: "func\n"` / `Actual: "pred\n"` |

Round trip: 15/15 → 14/15 → 14/15 → 15/15, rebuilt each time.

**Distinct values confirmed** — both tests separate `func` from `pred` observably. Not decorative. The constructor path at `Calls.cs:1519`, which I explicitly did **not** measure last round, is now genuinely covered; that claim checks out.

**Deleted test — no coverage lost, and #3021 is not load-bearing.** Since `10aa0ba9b` the file went +4/−1 tests. The removed `MismatchedLambdaArityThroughErasedDelegateSlot_IsRejected` is replaced *in this PR* by `MismatchedInlineLambdaArityThroughErasedSlot_IsRejected` with the **same** source shape (`List[System.Action[Src]]()` + 2-arg lambda) and a **stricter** assertion — exact `GS0144: Function 'lambda' requires 1 arguments but was given 2.` instead of generic `GS0159`+`GS0155`. That is an improvement, not a deletion.

---

## BLOCKER — 30 pre-existing tests regress across 15 emit classes

`src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs:1029-1039`

Identical 188-test denominator, same machine, same filter:

| tree | result |
|---|---|
| `main` + #3042 (**binder byte-identical to `main`**) | **188 passed / 0 failed** |
| `main` + #3019 | **158 passed / 30 failed** |
| #3019 **unmerged**, on its own base `7a6307a79` | **158 passed / 30 failed** |

The third row matters: this is **not** a merge interaction with the three commits `main` took. #3019 is broken on its own base. Full `Compiler.Tests` on the merged tree: **3960 passed / 30 failed / 3990**. CI's `build` job is already red at 44m28s — the `15/15` in the report is a filtered run.

**Mechanism.** The new call replaces target-typed lambda binding with a bind-then-convert pair:

```csharp
var literal = lambdas.BindLambdaExpression(lambdaSyntax, target.FunctionType, ...);
boundArgs[idx] = conversions.BindConversion(ce.Arguments[idx].Location, literal, target.DelegateType);
```

`target.DelegateType` is the *mapped* delegate, which can still carry an **unsubstituted method type parameter**. The conversion then fails against an open type:

```
GS0155: Cannot convert type '(Item) -> int32' to 'System.Func[Item, TResult]'.
GS0155: Cannot convert type '(OutputColumn) -> string' to 'System.Func[OutputColumn, TResult]'.
```

The candidate is discarded and overload resolution reports the method as missing:

```
GS0159: Cannot find function Select.   (x20)
GS0159: Cannot find function Sum.      (x4)
GS0159: Cannot find function Join.     (x2)
GS0160: Call to 'WriteLine' is ambiguous between 2 applicable overloads.  (x3)
```

Representative, from `Issue1334DictionaryLinqReceiverErasureEmitTests.PrimitiveProjection_Control_StillWorks` — a test named *Control_StillWorks*:

```
test.gs(10,23,10,51): error GS0159: Cannot find function Select.
        for v in d.Values.Select((e EPrim) -> e.Value) { sum = sum + v }
```

Affected classes: `Issue1150IntegerWidening`, `Issue1305ChainedLinq`, `Issue1334DictionaryLinqReceiverErasure`, `Issue1457UserTypeDelegate`, `Issue1462ListEnumerator`, `Issue1467GenericEncodingResidual`, `Issue1469LambdaHostAccess`, `Issue1502DelegateUserTypeArg`, `Issue1507SliceUntypedLambdaLinq`, `Issue2375ExpressionTreeMethodTypeArgument`, `Issue2498NullableLambdaGenericInference`, `Issue2673SymbolicReceiverDelegateReturn`, `Issue2715ForRangeNestedCapture`, `Issue903SameCompilationLambdaInference`, `SampleConformanceTests`. Also 5x `ilverify … [StackUnexpected]`, one `BadImageFormatException`, and `Sample_BuildsAndMatchesGolden`.

**Recommended fix.** Only convert when the mapped delegate is fully closed; fall back to the previous target-typed bind when `target.DelegateType` still contains an unsubstituted type parameter. The identity information you need for the two `NominalDelegateDisagreement` tests is already carried in the tuple and checked in `SymbolicLambdaTargetsAgree` / `TryResolveSymbolicDelegateTargetForCtor` — the *conversion* is the part that over-reaches, not the identity comparison. That keeps both mutants killed without touching the 30.

## NIT

`IsCanonicalFunctionDelegate` is deleted outright (`Invocation.cs`, −19 lines). That guard was the only thing restricting symbolic lambda targets to `System.Action`/`System.Func`, and removing it is what widens the blast radius to every lambda bound to an imported generic method. Deleting it is probably right for #2948, but it deserves its own regression pass rather than being carried by the two new `NominalDelegateDisagreement` tests.

---

Method note: three cells, one denominator, control run **first**. I did not attribute the failures until the binder-identical tree passed all 188.

DavidObando:
## Rubber-duck review — PR #3019 (issue #2948)

Numbers true of `main` **`4469e323`** (re-checked at end of review — unmoved), head **`7b180219`**, merge base `402b8561` (**2 behind / 6 ahead**), merged tree **`2ede9d28`** (`merge-tree` output was **one line, zero conflict lines** — read as output, not rc; probe merge reproduced `2ede9d28` exactly). Merged build **`BUILD_RC=0`**.

---

## 1. How `cs2gs-oahu` went green — artifact diff, red `30738049963` → green `30776786449`

Both `cs2gs-oahu-report` artifacts downloaded and parsed.

| | RED `c19a07a5` | GREEN `7b180219` |
|---|---|---|
| totalApps | **15** | **15** |
| greenApps | 4 | **15** |
| unverifiedApps | 0 | **0** |
| gap articles | **27** (13× GS0159, 10× GS0155, 4× GS0156) | **0** |
| stages passing | translate only on 11 apps | translate + compile + **ilverify** + **test-parity** on all 15 |

Fingerprint-level diff: **27 red-only (fixed), 0 green-only (no new gaps), 0 persisting.**

**Narrowing is RULED OUT, on four independent grounds:**
1. Same **15** apps by `appId` — none dropped, none renamed.
2. `unverifiedApps: 0` — nothing marked skipped/unverified.
3. **`test-parity` passed on all 15** — a gate that compiled nothing cannot pass execution parity.
4. The PR touches **zero** corpus, gate, workflow or `tools/cs2gs/` files — only 2 binder files + 2 test files.

*(Counting note: I measure **27** gap articles / 27 distinct fingerprints, not the 24 in your brief. Your **12 constructs** reconciles exactly — deduping the offending lines gives 12 distinct `Sum`/`Max` constructs.)*

**But CI cannot attribute that green to this PR, and I will not claim it does.** The red commit `c19a07a5` **already contained 5 of the PR's 6 commits**. Between red and green, **59 `main` commits** landed *plus* one PR commit (`7b180219`). Every corpus construct I could faithfully reproduce (`Sum`/`Max`/decimal/uint32/interface element/erased generic receiver, and the ref-pack LINQ shape) **already compiles on `main` without this PR**. So: green is real and un-narrowed, but co-caused.

## 2. Your root-cause hypothesis — OVERTURNED

`ExpressionBinder.Calls.Invocation.cs:829` on `main` already reads `imported.OpenDefinition ?? imported.ClrType`. A *constructed* `Func[...]` therefore falls back to `ClrType.FullName` = ``System.Func`2[[…]]``, which matches `StartsWith("System.Func`")` and returns **true**. That fallback landed in **`a1237ded` (#2929)** and **`git merge-base --is-ancestor a1237ded c19a07a5` = YES** — it was present *while the gate was red*. The `OpenDefinition == null → false` mechanism was not the red cause.

## 3. The PR does fix the shape #2948 actually reports

Issue repro run verbatim (`List[Predicate[Src]]` + inline lambda):

| shape | BASE `4469e323` | MERGED `2ede9d28` |
|---|---|---|
| `List[Predicate[Src]]` | **rc=1 `GS0155`+`GS0159`** | **rc=0 → `True`** |
| `List[Comparison[Src]]` | rc=1 `GS0155`+`GS0159` | rc=0 → `22` |
| `List[Converter[Src,int32]]` | rc=1 `GS0155`+`GS0159` | rc=0 → `22` |

Identical across **bare `gsc` (evaluate)**, **`gsc /out:` (emit)**, **`gsi`**. That is #2929's conservative rejection lifted — and note the diagnostics are the **same IDs the red gate reported**.

**Identity is genuinely preserved, not collapsed** — the silent-wrong-answer you asked about. `Assembly.LoadFrom` + `GetTypes()` + runtime `GetType().FullName`:

```
System.Predicate`1[[P.Src, P, …]]        <- NOT Func
System.Comparison`1[[P.Src, P, …]]
System.Converter`2[[P.Src, …],[System.Int32, …]]
System.Func`2[[P.Src, …],[System.Boolean, …]]   <- Func stays Func
```
Named delegates also survive as declared types: `Holder.G : P.Greeter`, `DELEGATE-TYPE P.MyPred\`1`.

## 4. Pinned

Product mutant **M1** — condition inversion at `ExpressionBinder.Calls.Invocation.cs:862` (`if (IsCanonicalFunctionDelegate(type))` → `if (!…)`), **asserted applied** (`assert count(old)==1`), **proven live** (all 3 driver copies re-stamped; reported shape regressed to `GS9999`).

**Kills 5 of 25**, precisely on subject:
- `ImportedNamedDelegatesThroughErasedListSlots_VerifyLoadAndRun` — issue repro 1
- `ImportedNamedDelegatesThroughErasedConstructorSlots_VerifyLoadAndRun` — issue repro 2 (`DelegateBox`)
- `SameFullNameDelegateFromUserAssembly_RetainsAssemblyIdentity` — the issue's own "assembly-blind" caveat
- `CanonicalLinqSelectorsOverErasedUserTypes_VerifyLoadAndRun(topLevel: True)` **and** `(topLevel: False)`

That last one matters twice: it covers the corpus LINQ shape, and the author **parameterised the `{top-level, in-function}` axis** rather than testing one cell.

**The deleted test is not lost coverage.** `e5af09b6 "Drop redundant lambda arity coverage"` removed `MismatchedLambdaArityThroughErasedDelegateSlot_IsRejected`, but `:237` `MismatchedInlineLambdaArityThroughErasedSlot_IsRejected` supersedes it with a *stronger* assertion — exact text `GS0144: Function 'lambda' requires 1 arguments but was given 2.` vs. the old loose "contains GS0159".

## 5. False-positive corpus (the failure mode of this very bug)

Correct code that must keep compiling — compared on rc **and** full output, base vs merged:

- **Arity matrix**: `Func` 0–4 and `Action` 0–3 through erased `List[…]` slots → `10 21 32 43 54 60 70 81 92`, **identical on both trees, all 3 drivers**.
- **User-declared named delegate** + **generic user delegate** (`type Greeter = delegate func(…)`, `type MyPred[T any] = …`) → identical, identity preserved.
- **Corpus LINQ shapes**: `Sum`/`Max`(TimeSpan)/decimal/uint32/interface element/erased generic base → identical.
- **Ref-pack references** (`packs/Microsoft.NETCore.App.Ref/10.0.8/ref/net10.0`, 167 dlls, `/nowarn:GS9100`) → both `rc=0`, both print `74`.
- **18 delegate/lambda `samples/*.gs`** → **0 differences** in rc or output.
- Overload selection across two same-arity `Func` overloads → `11`/`99` on both trees.

**Suites on the merged tree:** Compiler.Tests `~Issue2918` **25/25**; Interpreter.Tests **726/726**; Core.Tests **7089 passed / 0 failed / 1 skipped**. *Filtered-green disclosure:* I did not run full Compiler.Tests (~1h20m) or Cs2Gs.Tests; the PR's surface is the binder's delegate path, and I ran the full binder suite plus every test class touching it.

## 6. NITs

**NIT-1 — `SameDelegateIdentity` skips type-argument comparison for canonical delegates.** `:897`: after matching definition and arity, `if (IsCanonicalFunctionDelegate(left) && IsCanonicalFunctionDelegate(right)) return true;` returns **without** recursing into type arguments, so `Func[A,B]` and `Func[C,D]` of equal arity compare identical. I believe this is deliberate (canonical = structural, nominal = named — the premise of #2948). **I tried to fire it and could not**: two same-arity `Func` overloads still resolve correctly (`11`/`99`) on both trees. Reporting as an **unfired hypothesis**, not a defect — worth one comment stating the invariant.

**NIT-2 — a nonexistent `/r:` path is silently ignored.** `gsc k_linq.gs /r:/tmp/r6/does-not-exist.dll` → **rc=0, prints `74`, "Success."**, zero diagnostics. **Pre-existing — identical on base**, not introduced here. Flagging because it is exactly the trap that cost another duck an 18-probe run, and it is invisible to anyone who does not positive-control their `/r:`. Worth its own issue (you are sole filer).

**NIT-3 — attribution.** The PR body / thread should not claim the `cs2gs-oahu` recovery, since 59 `main` commits landed in the same window. The PR's value is independently proven by §3 and §4 and does not need that claim.

## 7. Measurement errors I made — disclosed

1. **I ran a batch of probes against a stale mutant build.** After restoring M1 I verified the source was byte-identical but **did not rebuild**, then ran the ref-pack probe. It reported BASE `rc=0` / MERGED `rc=1 GS0156: Cannot convert 'decimal' to 'int64'` — which I was one step from filing as a **BLOCKER source break**. Rebuilding (`BUILD_RC=0`, all three copies re-stamped) gave **both trees `rc=0`, both print `74`**. **The "regression" was my own mutant.** Restoring source is not restoring the compiler.
2. My first positive control for `/r:` "produced nothing" — it had actually *succeeded* (rc=0), which is NIT-2. I misread absence-of-diagnostic as a broken control.
3. Three probes were my own syntax errors, not findings: `GS0421` (`open` member in a non-`open` class), `GS0005` (G# takes no `;` statement separator), `GS0005`+ (`shared func` is not member syntax).
4. My first `Max` probe used `int64` and hit `GS0160` ambiguity on **both** trees — an artifact of my type choice, not the corpus, which uses `TimeSpan`. Withdrawn after re-running the corpus's actual type.
5. My first `Predicate[T]` identity probe collided with `System.Predicate` and reified as `System.Predicate\`1`; renamed to `MyPred` to get a clean reading.

Every probe compiled into its own directory asserted empty. Mutant asserted applied and proven live. `/tmp/r6/{m,b}` porcelain 0; protected checkout porcelain 0; **`gsharp-wt/i2948` never written to** (read-only, not mine).

DavidObando:
## Rubber-duck addendum — re-anchored to `main` `f7d9a5d3`

My original verdict (https://github.com/DavidObando/gsharp/pull/3019#issuecomment-5161962917) was measured at `main` `4469e323`. Re-anchored and re-measured. **Verdict unchanged: MERGEABLE.** One finding is **upgraded from NIT to SHOULD-FIX** with a precise repro.

### Re-anchor facts

| item | value |
|---|---|
| new `main` | `f7d9a5d3` (2 commits: #3123, #3052) |
| merge base | `402b8561` — 4 behind / 6 ahead |
| `merge-tree` | rc=0, **1 line of output** (tree `90942edc`), no conflict lines |
| tree identity | probe merge reproduced `90942edc` **exactly** |
| BASE / MERGED `BUILD_RC` | **0 / 0** |
| overlap of new `main` commits with PR files | **none** — diff unchanged |
| `Compiler.Tests ~Issue2918` on merged | **25 / 25** |
| false-positive corpus (20 samples) | **20/20 identical** after normalising the tree path out of diagnostic text |

Reported shape still reproduces at the new anchor: BASE rc=1 `GS0159` → MERGED rc=0, all three drivers.

### SHOULD-FIX (upgraded) — issue #2948's own subject is fixed in only one cell of a two-axis space

The failing shapes **compile rc=0 with zero diagnostics** (verified without `/nowarn`) and then throw at runtime. Cross-assembly, lambda through an erased `List[...]` slot:

| # | shape | base `f7d9a5d3` | merged `90942edc` |
|---|---|---|---|
| cA | generic named delegate, type arg = **consumer-local** class | rc=1 `GS0159` | **rc=0, identity `CSDL.Pred'1[[C.Src]]`** — fixed |
| cB | generic named delegate, type arg = **imported** class | rc=0 → runtime throw | rc=0 → **runtime throw** — not fixed |
| cC | **non-generic** named delegate | rc=0 → runtime throw | rc=0 → **runtime throw** — not fixed |

```
System.ArgumentException: The value "System.Func`2[System.String,System.Int32]"
is not of type "CSDL.Greeter" and cannot be used in this generic collection.
   at System.Collections.Generic.List`1.System.Collections.IList.Add(Object item)
```

The lambda is reified as `System.Func<...>` instead of the named delegate — exactly "preserve imported named-delegate identity through deferred erased-slot lambda targeting".

**These are pre-existing (identical on base), not regressions.** That is why this is SHOULD-FIX and not BLOCKER. But the discriminator matters for the test matrix: it is **not** "BCL vs user-declared" and **not** "C#-declared vs G#-declared" — I reproduced cC with the delegate declared in **both** a C# assembly and a `gsc`-built assembly. The axes are *genericity of the delegate* and *origin of the type argument*.

Every delegate in `test/Compiler.Tests/Emit/Issue2918InlineLambdaErasedReceiverTests.cs:581` (`Predicate`, `Comparison`, `Converter`, `EventHandler`, `CustomPredicate`) is **generic with a consumer-local type argument** — precisely the one cell that works. `SameFullNameDelegateFromUserAssembly_RetainsAssemblyIdentity:682` looks like it covers user assemblies, but its delegate is literally *named* `System.Func<T,TResult>` (`:87`), so it exercises the **canonical** name path, not the nominal one.

Recommend adding two rows to that test: a non-generic named delegate, and a generic one whose type argument is imported. Each is a couple of lines in the existing harness.

### Scope notes, measured rather than assumed

- **Cross-assembly benefit is emit-only.** Under bare `gsc` (evaluate), *any* cross-assembly delegate invocation fails `GS9999: Cannot invoke a method on objects loaded by a MetadataLoadContext` on **both** trees. `gsi` has no `/r:` at all, so there is no cross-assembly `gsi` column.
- Same-assembly generic user delegate in an erased slot fails `GS9999: Unexpected type UPred[Src]` under evaluate/`gsi` on **both** trees; emit is fine. Pre-existing and orthogonal.
- Interpreter type-argument erasure to `System.Object` does **not** leak: `List.Find`/`Sort`/`ConvertAll`/`TrueForAll`, which demand the exact constructed delegate type, return `7|1|True|1|70` identically on both trees across all three drivers.

### Still standing from the original verdict

- **`cs2gs-oahu` green is real, not narrowed** — 15/15 apps in both runs, `unverifiedApps: 0`, 27 → 0 gap fingerprints, **0 green-only**, all four stages including `test-parity`, and the PR touches zero corpus/gate/workflow files.
- **Attribution is not established by CI** — the red commit already contained 5 of the PR's 6 commits, and 59 `main` commits landed between the runs. The thread should not claim the recovery outright.
- **The root-cause hypothesis remains refuted.** `?? imported.ClrType` (`src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs:829`) landed in `a1237ded` (#2929), which `git merge-base --is-ancestor` confirms is an ancestor of red commit `c19a07a5` — it was present *while the gate was red*.
- Mutation M1 (condition inversion at `ShouldConvertToNominalDelegate:862`, asserted applied, proven live) kills 5 of 25 named tests. The fix is pinned.

### My own errors this pass, disclosed

A uniform rc=1 across both trees was my harness passing a relative filename after `cd`; three corpus "differences" at rc=0 were the tree path embedded in diagnostic text; and `GetType()` on a G#-local delegate value is unsupported, which broke one probe. All caught and corrected before they reached a finding.

DavidObando:
## Correction — my NOT-RUN disclosure used a wrong denominator

Per the fleet bulletin on test-project enumeration, I audited my own verdict on this PR and it was **wrong by omission**. Correcting it here rather than leaving it.

My verdict declared only *"full `Compiler.Tests`, `Cs2Gs.Tests`, LSP tests, Oahu corpus"* as NOT RUN. It never named `Extensions.Tests`, `InternalAnalyzers.Tests`, `GSharp.GeneratorHost.Tests`, or `Sdk.Tests`. Silence there implied coverage I did not have.

Enumerated from `GSharp.sln`, **not** `ls test/`, there are **9** test projects. Worth noting the trap is sharper than "7 vs 9": `ls test/` returns **8** directories here, because `test/Shared/` is a shared library and not a test project. So both "7" and "8" are reachable by miscounting, and I would have been wrong either way.

### What I have now actually run, on shipped `main` `5b71a74d`

This PR **squash-merged as `46f2c929`**. Note the PR head `7b180219` is *not* an ancestor of `main`, so an `--is-ancestor` check reports "NOT ancestor" — that is squash-merge, not a missing change. I verified the shipped content directly instead: `ExpressionBinder.Calls.cs` is byte-identical to the reviewed head, and `ExpressionBinder.Calls.Invocation.cs` has since taken +323/−13 from #3054, #3121 and `6ed39f67`, **none of which touch `IsCanonicalFunctionDelegate`, `ShouldConvertToNominalDelegate` or `SameDelegateIdentity`**. The shipped delegate logic is the logic I reviewed.

| project | result |
|---|---|
| `GSharp.GeneratorHost.Tests` | 31 / 31 |
| `Cs2Gs.Tests` (out-of-process `gsc.dll`) | **1903 / 1903**, 17 m 38 s |
| `Core.Tests` | 7089 / 0 fail / 1 skip *(measured on the merge preview, not `5b71a74d`)* |
| `Interpreter.Tests` | 726 / 726 *(merge preview)* |
| `Compiler.Tests` | filtered `~Issue2918` 25 / 25; **full run in progress** |
| `Sdk.Tests` | **NOT RUN** — local SDK package, out of bounds |

`Cs2Gs.Tests` is the one that mattered most here and I had waved it off as unrun: it drives the real `gsc.dll` out-of-process, which is the same surface as the `cs2gs-oahu` gate this PR's whole story turned on. It is green.

### The SHOULD-FIX shipped, and still reproduces on `main` `5b71a74d`

Re-measured against the shipped compiler, not the merge preview:

| shape | shipped `main` |
|---|---|
| generic named delegate, **consumer-local** type arg | rc=0, identity `CSDL.Pred'1[[C.Src]]` — fixed |
| generic named delegate, **imported** type arg | rc=0, **zero diagnostics**, runtime `ArgumentException` |
| **non-generic** named delegate | rc=0, **zero diagnostics**, runtime `ArgumentException` |

Still pre-existing rather than caused by this PR, and still silent at compile time. Since #2948 is now closed, these two cells no longer have an owner — they warrant a follow-up issue rather than being inherited by a closed one. Repro is three files and is in my addendum above.

### Method notes from the same bulletin, applied to myself

- I re-anchored twice: `4469e323` → `f7d9a5d3` → `5b71a74d`. `main` moved 18 commits past my second anchor, and past the bulletin's own `787166a9`, while I was measuring.
- mtimes re-recorded in **epoch seconds with `sort -n`**, not `%H:%M:%S`. All three of `Core/`, `Compiler/`, `Repl/` at `1785753761`, 97 s before the probes.
- Load hit **36.57** immediately after my solution build returned. The build had already completed (`BUILD_RC=0`); I held the five test projects until it fell back under, and ran them sequentially rather than in parallel. Disclosing that rather than quietly proceeding.

DavidObando:
## Denominator correction — completed, all on one anchor

Finishing what I posted above. Every project below was run on the **same** commit, `main` `5b71a74d`, which contains this PR as squash `46f2c929`. My earlier `Core.Tests`/`Interpreter.Tests` figures were measured on the pre-merge preview tree, so I re-ran them here to remove the mixed-anchor caveat.

| project | passed | failed | skipped |
|---|---|---|---|
| `Core.Tests` | 7106 | 0 | 1 |
| `Compiler.Tests` (**full**, 1 h 30 m) | 4221 | 0 | 0 |
| `Cs2Gs.Tests` (out-of-process `gsc.dll`, 17 m 38 s) | 1903 | 0 | 0 |
| `Interpreter.Tests` | 1201 | 0 | 0 |
| `LanguageServer.Tests` | 462 | 0 | 0 |
| `Extensions.Tests` | 104 | 0 | 0 |
| `GSharp.GeneratorHost.Tests` | 31 | 0 | 0 |
| `InternalAnalyzers.Tests` | 7 | 0 | 0 |
| **8 of 9 projects** | **15035** | **0** | **1** |

`Sdk.Tests` is **NOT RUN** — local SDK package, out of bounds. Saying that plainly rather than implying coverage.

Two things worth flagging from re-running on the current anchor: full `Compiler.Tests` (4221) and `Cs2Gs.Tests` (1903) were both projects I had waved off as unrun in my verdict, and both are green. `Interpreter.Tests` moved 726 → 1201 between `f7d9a5d3` and `5b71a74d`, which is a reminder that a suite count is itself anchor-dependent and not a stable identifier.

### Corpus narrowing — now proved from the pin, not from circumstance

My verdict ruled out corpus narrowing on four circumstantial grounds (app counts, `unverifiedApps: 0`, stages run, files touched). Per the bulletin, that is provable only from the pin, so I checked it:

`tools/cs2gs/external/oahu.json` is blob `0552273f2088` at **all** of: red-run commit `c19a07a5`, PR head `7b180219`, squash `46f2c929`, and current `main` `5b71a74d`.

Byte-identical throughout the red→green window. The corpus did not change. That is the proof my original argument was only circumstantially reaching.

### Still open, and now ownerless

The two silent-miscompile cells from my addendum still reproduce on shipped `main` `5b71a74d`: a **non-generic** named delegate, and a **generic** named delegate whose type argument is imported, each targeted by a lambda through an erased `List[...]` slot cross-assembly. Both compile rc=0 with **zero diagnostics** and throw `ArgumentException` at runtime because the lambda is reified as `System.Func<...>`.

Pre-existing, not caused by this PR, and correctly out of scope for it. But #2948 is closed now, so they have no owner. They need a follow-up issue rather than being assumed fixed by this one.

Issue #2190, opened 2026-07-06
Investigate interpreter performance


DavidObando:
Can we make interpreted G# closer in speed to compiled G#? For example, write a recursive Fibonacci function and measure its time getting the 33rd number in the sequence. Compare compiled to interpreted. Find ways to make the performance of the evaluator more closely match compiled code.

DavidObando:
We've moved away from tree-walking interpretation to in-memory compilation.

PR #3020, opened 2026-08-01
Fix interpreter generic CLR type reification

### LINKED ISSUE #2992: gsi: generic type parameter used as a type argument instantiates List<object>
## Problem

When a generic type parameter is used as a type argument, the interpreter instantiates the erased `object` form.

`samples/GenericTypeParameterAsTypeArgument.gs`:

```
compiled: 10 / 20 / 42
gsi:      GS9999: Object type System.Collections.Generic.List`1[System.Object]
          does not match target type System.Collections.Generic.List`1[System.Int32].
```

The interpreter built a `List<object>` where the reflection target required `List<int>`.

## Expected

`T` must be substituted with the bound type argument when constructing a CLR generic instance, so `List[T]` inside a `T = int32` instantiation produces `List<int>`.

## Impact

Any generic G# code that forwards its own type parameter into a CLR generic type fails at the first reflection call. This is not a niche shape.

DavidObando:
## Summary

- track runtime generic arguments for function calls, constructed class/interface receivers, and static constructed class owners
- substitute those arguments into symbolic `ImportedTypeSymbol` instances, then call `ReifyClosedClrType()` before reflection
- remap seven retained reflection sites: static method calls, constructors, static property reads, instance property reads, property writes, index reads, and index writes
- preserve generic mappings in deferred closures and method groups
- carry symbolic container identity for static imported-generic property writes
- cache remapped reflection members per evaluator

Fixes #2992

## Baseline reproduction (`f70626452`)

`gsi samples/GenericTypeParameterAsTypeArgument.gs` failed with:

```
GS9999: Object type System.Collections.Generic.List`1[System.Object] does not match target type System.Collections.Generic.List`1[System.Int32].
```

The same source compiled and ran with `gsc`, printing:

```
10
20
42
```

Minimal probes also reproduced both reversed failure directions: erased `List<object>` against `List<int>`, and concrete `List<int>` against an erased `List<object>` target.

## Site-specific mutation evidence

Each retained reflection site has a distinct observable test. Hashes below are from the rebuilt `GSharp.Core.dll`, not the byte-identical `gsc` driver; the new static-property-read mutant used Release + `ContinuousIntegrationBuild=true`.

| Mutated site | Expected | Mutant result | Mutant Core MD5 |
| --- | --- | --- | --- |
| static method call | `11` | `-11` | `22ae24a658bfd3e03ef69462e1615599` |
| static property read | ``GenericComparer`1`` | ``ObjectComparer`1`` | `bc52c88cbb354366f452863bdca73960` |
| constructor | `33` | `GS9999` | `b78e775ff0ef179cd9e26704fd838436` |
| instance property read | `44` | `GS9999` | `53069e6ec9ab337f42277e1f333189dd` |
| property write | `55` | `GS9999` | `828e14254148bffda7793c2790beb2be` |
| index read | `66` | `GS9999` | `bfb620154ac8c2b79e798a5dd5bdfec9` |
| index write | `77` | `GS9999` | `d01d9ae8f04c1d79e4624de1801e7d1c` |

Supporting mapping mutations also died:

| Mutated mapping | Expected | Mutant result | Mutant Core MD5 |
| --- | --- | --- | --- |
| constructed class receiver | `2` | `GS9999` | `afc7dc9d6cfdc9b0d77643890a592e75` |
| static constructed class owner | `4` | `GS9999` | `f4e3a881a2de8cd640e5e6578dfed777` |
| deferred closure capture | `2` | `GS9999` | `4ada3cc3091fdbd0bc6bbacb382bf223` |
| deferred method-group capture | `3` | `GS9999` | `673711a933cf238d6f07c53b7b2f0e48` |
| static property runtime container | `111` | `0` | `3c0ea808ad69cd09de647c5890b74c68` |
| static property binder container | `111` | `0`; 1/1 test failed | `b19c28d647f9e9046457ed8488be0c30` |

Restoring the binder container made the full issue suite pass 11/11. A proposed imported instance-method remap was deleted: reverting it survived all 7 examined targeted/#794 tests, so the PR no longer carries that unobservable edit.

## Final validation

Rebased onto `origin/main` `46d3cc50fd874e9a03eefc9a8b28f94fe54e0adc`; head `c47416ac0e4b436eff057c56dd011832b5c26570`.

- issue + related #773/#794 interpreter tests: 25 passed, 0 failed, 25 total
- `gsi` sample: exit 0, values `10 / 20 / 42`
- `gsc` compile: exit 0; emitted assembly run: exit 0, values `10 / 20 / 42`
- three-dot diff check: clean

## Scope and structural check

The occurrence count did not survive contact with current tip: both current base and head contain 21 exact `.ClrType ?? typeof(object)` occurrences across 667 `src/**/*.cs` files, down from the earlier 27-site count. This PR rewrites none of them. Making the shared evaluator helper require a `siteId` produced exactly 7 unique `CS7036` errors, proving the reflection-site denominator. Current base has 0 `ReifyClosedClrType()` calls across 7 `Evaluator*.cs` files; head has 1 shared evaluator reification path.

`SignatureEncoder.cs` remains accurate because this change does not alter the interpreter channel-element fallback it describes. #2993/#2996 struct-literal allocation paths are untouched.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/BoundClrPropertyAssignmentExpression.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundClrPropertyAssignmentExpression.cs
@@ -27,7 +27,8 @@ public sealed class BoundClrPropertyAssignmentExpression : BoundExpression
         BoundExpression value,
         TypeSymbol resultType,
         TypeParameterSymbol constrainedReceiverTypeParameter = null,
-        TypeSymbol constrainedInterfaceType = null)
+        TypeSymbol constrainedInterfaceType = null,
+        TypeSymbol staticContainerType = null)
         : base(syntax)
     {
         Receiver = receiver;
@@ -36,6 +37,7 @@ public sealed class BoundClrPropertyAssignmentExpression : BoundExpression
         Type = resultType;
         ConstrainedReceiverTypeParameter = constrainedReceiverTypeParameter;
         ConstrainedInterfaceType = constrainedInterfaceType;
+        StaticContainerType = staticContainerType;
     }
 
     public BoundExpression Receiver { get; }
@@ -47,6 +49,8 @@ public sealed class BoundClrPropertyAssignmentExpression : BoundExpression
     public TypeParameterSymbol ConstrainedReceiverTypeParameter { get; }
 
     public TypeSymbol ConstrainedInterfaceType { get; }
+
+    public TypeSymbol StaticContainerType { get; }
 
     public bool IsConstrainedTypeParameterAccess => ConstrainedReceiverTypeParameter != null;
 
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs
@@ -2341,7 +2341,13 @@ internal sealed partial class ExpressionBinder
             && TryGetWritableClrMember(staticMember, out _, out var staticTargetSymbol, out _))
         {
             var staticConverted = conversions.BindConversion(syntax.Value.Location, BindValue(staticTargetSymbol), staticTargetSymbol);
-            return new BoundClrPropertyAssignmentExpression(null, receiver: null, staticMember, staticConverted, staticTargetSymbol);
+            return new BoundClrPropertyAssignmentExpression(
+                null,
+                receiver: null,
+                staticMember,
+                staticConverted,
+                staticTargetSymbol,
+                staticContainerType: constructedImported.SymbolicReceiver);
         }
 
         Diagnostics.ReportUnableToFindMember(syntax.FieldIdentifier.Location, fieldName);
--- a/src/Core/CodeAnalysis/ClosureValue.cs
+++ b/src/Core/CodeAnalysis/ClosureValue.cs
@@ -21,12 +21,19 @@ public sealed class ClosureValue
     /// <param name="body">The lambda body.</param>
     /// <param name="functionType">The function-type signature.</param>
     /// <param name="capturedLocals">Value-snapshot of captured locals at literal-evaluation time.</param>
-    public ClosureValue(FunctionSymbol function, BoundBlockStatement body, FunctionTypeSymbol functionType, Dictionary<VariableSymbol, object> capturedLocals)
+    /// <param name="capturedTypeArguments">Generic type-argument bindings visible at literal-evaluation time.</param>
+    public ClosureValue(
+        FunctionSymbol function,
+        BoundBlockStatement body,
+        FunctionTypeSymbol functionType,
+        Dictionary<VariableSymbol, object> capturedLocals,
+        Dictionary<TypeParameterSymbol, TypeSymbol> capturedTypeArguments)
     {
         Function = function;
         Body = body;
         FunctionType = functionType;
         CapturedLocals = capturedLocals;
+        CapturedTypeArguments = capturedTypeArguments;
     }
 
     /// <summary>Gets the synthetic function symbol.</summary>
@@ -40,4 +47,7 @@ public sealed class ClosureValue
 
     /// <summary>Gets the captured locals snapshot.</summary>
     public Dictionary<VariableSymbol, object> CapturedLocals { get; }
+
+    /// <summary>Gets the captured generic type-argument bindings.</summary>
+    public Dictionary<TypeParameterSymbol, TypeSymbol> CapturedTypeArguments { get; }
 }
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -28,6 +28,8 @@ namespace GSharp.Core.CodeAnalysis;
 public sealed partial class Evaluator
 #pragma warning restore CA1001
 {
+    private static readonly System.Runtime.CompilerServices.ConditionalWeakTable<ConcurrentDictionary<VariableSymbol, object>, Dictionary<TypeParameterSymbol, TypeSymbol>> RuntimeTypeArguments = new();
+
     private object EvaluateCallExpression(BoundCallExpression node)
     {
         // ADR-0047 §6 / issue #176: a [Conditional("SYMBOL")] call whose
@@ -79,6 +81,7 @@ public sealed partial class Evaluator
                 : localFunctionBodies[node.Function];
             var isIterator = IsIteratorFunction(node.Function, statement);
             object result;
+            RegisterRuntimeTypeArguments(locals, node.Function.TypeParameters, node.MethodTypeArguments);
             using (PushFrame(locals))
             {
                 if (isIterator)
@@ -365,6 +368,7 @@ public sealed partial class Evaluator
             frame[parameter] = value;
         }
 
+        RegisterRuntimeTypeArguments(frame, method.TypeParameters, node.MethodTypeArguments);
         using (PushFrame(frame))
         {
             var statement = program.Functions[method];
@@ -909,7 +913,8 @@ public sealed partial class Evaluator
             return emulatedResult;
         }
 
-        var result = node.Function.Method.Invoke(null, args);
+        var method = ResolveMemberForRuntimeType(node.Function.Method, ResolveRuntimeClrType(node.StaticContainerType));
+        var result = method.Invoke(null, args);
         WriteBackRefSlots(refSlots, args);
         return result;
     }
@@ -1009,7 +1014,7 @@ public sealed partial class Evaluator
         receiver = UnwrapClrReceiver(receiver);
         var args = new object[node.Arguments.Length];
         var refSlots = BuildRefSlots(node.Arguments, node.ArgumentRefKinds, args, node.Method);
-        var method = ResolveMethodForReceiver(node.Method, receiver);
+        var method = ResolveMethodForReceiver(node.Method, receiver, ResolveRuntimeClrType(node.Receiver.Type));
 
         // concurrency: see TryGetInterpreterMapLock — routes `range m`'s
         // `GetEnumerator()`/`MoveNext()`, `m.ContainsKey(k)`, `m.Remove(k)`,
@@ -1047,13 +1052,14 @@ public sealed partial class Evaluator
     // The lookup also walks inherited interfaces (e.g. `MoveNext` lives on
     // the non-generic `IEnumerator` even when the receiver is reached via
     // `IEnumerator<T>`).
-    private static System.Reflection.MethodInfo ResolveMethodForReceiver(System.Reflection.MethodInfo method, object receiver)
+    private static System.Reflection.MethodInfo ResolveMethodForReceiver(System.Reflection.MethodInfo method, object receiver, Type symbolicReceiverType)
     {
         if (receiver == null || method == null)
         {
             return method;
         }
 
+        method = ResolveMemberForRuntimeType(method, symbolicReceiverType);
         var declaring = method.DeclaringType;
         if (declaring == null || declaring.IsAssignableFrom(receiver.GetType()))
         {
@@ -1112,6 +1118,92 @@ public sealed partial class Evaluator
             types: paramTypes,
             modifiers: null);
         return direct ?? method;
+    }
+
+    private void RegisterRuntimeTypeArguments(
+        ConcurrentDictionary<VariableSymbol, object> frame,
+        ImmutableArray<TypeParameterSymbol> parameters,
+        ImmutableArray<TypeSymbol> arguments)
+    {
+        if (parameters.IsDefaultOrEmpty || parameters.Length != arguments.Length)
+        {
+            return;
+        }
+
+        var substitutions = new Dictionary<TypeParameterSymbol, TypeSymbol>(parameters.Length);
+        for (var i = 0; i < parameters.Length; i++)
+        {
+            substitutions.Add(parameters[i], ResolveRuntimeTypeSymbol(arguments[i]));
+        }
+
+        RuntimeTypeArguments.Add(frame, substitutions);
+    }
+
+    private Type ResolveRuntimeClrType(TypeSymbol type)
+    {
+        var resolved = ResolveRuntimeTypeSymbol(type);
+        return resolved is ImportedTypeSymbol imported
+            ? imported.ReifyClosedClrType()
+            : resolved?.ClrType;
+    }
+
+    private TypeSymbol ResolveRuntimeTypeSymbol(TypeSymbol type)
+    {
+        if (type is TypeParameterSymbol parameter)
+        {
+            foreach (var frame in Locals)
+            {
+                if (RuntimeTypeArguments.TryGetValue(frame, out var substitutions)
+                    && substitutions.TryGetValue(parameter, out var argument))
+                {
+                    return ReferenceEquals(argument, parameter) ? argument : ResolveRuntimeTypeSymbol(argument);
+                }
+            }
+
+            return type;
+        }
+
+        if (type is NullableTypeSymbol nullable)
+        {
+            var underlying = ResolveRuntimeTypeSymbol(nullable.UnderlyingType);
+            return ReferenceEquals(underlying, nullable.UnderlyingType)
+                ? type
+                : NullableTypeSymbol.Get(underlying);
+        }
+
+        if (type is not ImportedTypeSymbol imported
+            || imported.OpenDefinition == null
+            || imported.TypeArguments.IsDefaultOrEmpty)
+        {
+            return type;
+        }
+
+        var arguments = ImmutableArray.CreateBuilder<TypeSymbol>(imported.TypeArguments.Length);
+        var changed = false;
+        foreach (var argument in imported.TypeArguments)
+        {
+            var resolved = ResolveRuntimeTypeSymbol(argument);
+            changed |= !ReferenceEquals(resolved, argument);
+            arguments.Add(resolved);
+        }
+
+        return changed
+            ? ImportedTypeSymbol.GetConstructed(imported.ClrType, imported.OpenDefinition, arguments.MoveToImmutable())
+            : type;
+    }
+
+    private static T ResolveMemberForRuntimeType<T>(T member, Type runtimeType)
+        where T : System.Reflection.MemberInfo
+    {
+        var declaring = member?.DeclaringType;
+        if (declaring?.IsGenericType != true
+            || runtimeType?.IsGenericType != true
+            || declaring.GetGenericTypeDefinition() != runtimeType.GetGenericTypeDefinition())
+        {
+            return member;
+        }
+
+        return runtimeType.GetMemberWithSameMetadataDefinitionAs(member) as T ?? member;
     }
 
     // Issue #517: `GetValueOrDefault()` and `GetValueOrDefault(T)` against the
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -28,7 +28,8 @@ namespace GSharp.Core.CodeAnalysis;
 public sealed partial class Evaluator
 #pragma warning restore CA1001
 {
-    private static readonly System.Runtime.CompilerServices.ConditionalWeakTable<ConcurrentDictionary<VariableSymbol, object>, Dictionary<TypeParameterSymbol, TypeSymbol>> RuntimeTypeArguments = new();
+    private readonly System.Runtime.CompilerServices.ConditionalWeakTable<ConcurrentDictionary<VariableSymbol, object>, Dictionary<TypeParameterSymbol, TypeSymbol>> runtimeTypeArguments = new();
+    private readonly ConcurrentDictionary<(System.Reflection.MemberInfo Member, Type RuntimeType), System.Reflection.MemberInfo> runtimeMembers = new();
 
     private object EvaluateCallExpression(BoundCallExpression node)
     {
@@ -369,6 +370,7 @@ public sealed partial class Evaluator
         }
 
         RegisterRuntimeTypeArguments(frame, method.TypeParameters, node.MethodTypeArguments);
+        RegisterRuntimeTypeArguments(frame, node.Receiver.Type);
         using (PushFrame(frame))
         {
             var statement = program.Functions[method];
@@ -1052,14 +1054,13 @@ public sealed partial class Evaluator
     // The lookup also walks inherited interfaces (e.g. `MoveNext` lives on
     // the non-generic `IEnumerator` even when the receiver is reached via
     // `IEnumerator<T>`).
-    private static System.Reflection.MethodInfo ResolveMethodForReceiver(System.Reflection.MethodInfo method, object receiver, Type symbolicReceiverType)
+    private static System.Reflection.MethodInfo ResolveMethodForReceiver(System.Reflection.MethodInfo method, object receiver)
     {
         if (receiver == null || method == null)
         {
             return method;
         }
 
-        method = ResolveMemberForRuntimeType(method, symbolicReceiverType);
         var declaring = method.DeclaringType;
         if (declaring == null || declaring.IsAssignableFrom(receiver.GetType()))
         {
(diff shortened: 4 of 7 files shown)


DavidObando:
## Review — NOT MERGEABLE at `21ecfc853`

Measured against `main` = `f70626452`. Merge base **is** `f70626452` (0 behind, 1 ahead), so the built tree *is* the merged state — no separate merge needed. `dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true` → **rc=0**. Gates: 22 `GSharp.Core.dll` copies examined / 0 zero-byte; one md5 group `36dfe46c…` across the ordinary `out/bin/Release` paths; race grep 0 of 76 log lines (control pattern matched, so the grep is not an unrun check).

**The axis is right.** Reification is applied only at reflection boundaries (`Invoke` / `GetValue` / `SetValue`), which is exactly where the CLR enforces exact type identity, and `ResolveMemberForRuntimeType` early-outs unless both declaring and runtime types are constructions of the same generic definition. It does **not** over-reify. Structural claims in the body check out: `.ClrType ?? typeof(object)` is 27 on both sides (unchanged), and the evaluator gains exactly one `ReifyClosedClrType` caller (`Evaluator.Calls.cs:1098`), which was 0 before.

The problem is not the mechanism. It is the scope claimed, and the tests.

---

## BLOCKER 1 — `Fixes #2992` is over-declared: two whole classes of the same defect survive, with a byte-identical error message

`RegisterRuntimeTypeArguments` is only ever called with **method** type parameters — `Evaluator.Calls.cs:91` (`node.Function.TypeParameters`) and `Evaluator.Calls.cs:320` (`method.TypeParameters`). Nothing else populates the table. Two consequences, both measured:

**(a) Class / struct type parameters are not covered.**

```gsharp
class Holder[T any] {
    func CountOf(items List[T]) int32 { return items.Count }
}
let h = Holder[int32]()
var n = List[int32]()
n.Add(10)
n.Add(20)
Console.WriteLine(h.CountOf(n))
```

| | value |
|---|---|
| `gsc` | `2` |
| `gsi` @ `f70626452` | `GS9999: Object type …List``1[System.Object] does not match target type …List``1[System.Int32].` |
| `gsi` @ `21ecfc853` | **identical failure** |

**(b) Deferred execution is not covered.** The substitution table is a `ConditionalWeakTable` keyed on the **live frame**, and `ResolveRuntimeTypeSymbol` finds it by walking `Locals`. Once the generic frame is popped, `T` resolves to itself and the fix silently no-ops.

```gsharp
func MakeCounter[T]() (List[T]) -> int32 {
    return func(items List[T]) int32 { return items.Count }
}
let f = MakeCounter[int32]()
var n = List[int32]()
n.Add(10)
n.Add(20)
Console.WriteLine(f(n))
```

| | value |
|---|---|
| `gsc` | `2` |
| `gsi` @ `f70626452` | `GS9999` … `List``1[System.Object]` vs `List``1[System.Int32]` |
| `gsi` @ `21ecfc853` | **identical failure** |

#2992's body is not scoped to functions — it says *"when a generic type parameter is used as a type argument"*. Merging this with `Fixes #2992` closes the issue while the exact reported symptom remains reachable from ordinary code, and the next person to hit it will find a closed issue with a passing sample.

**Fix:** either cover class type parameters (register at the receiver/`this` binding site as well) and state the deferred-execution limit explicitly, or downgrade to `Refs #2992` and enumerate both limits in the body. The dynamic-scope limitation looks inherent to keying on a live frame, so declaring it is the honest move.

---

## BLOCKER 2 — 5 of 7 changed reflection sites are pinned by no test, and a mutant proves it

The diff changes **7** reflection sites:

| # | site | pinned by a shipped test? |
|---|---|---|
| 1 | `Evaluator.Calls.cs:864` static CLR call via `StaticContainerType` | no |
| 2 | `Evaluator.Calls.cs:966` instance method on symbolic receiver | no |
| 3 | `Evaluator.Expressions.cs:1206` constructor | yes |
| 4 | `Evaluator.Expressions.cs:1256` property/field read | no |
| 5 | `Evaluator.Expressions.cs:1403` property/field write | no |
| 6 | `Evaluator.Expressions.cs:1584` indexer read | yes |
| 7 | `Evaluator.Expressions.cs:1599` indexer write | no |

Sites 2, 4, 5 and 7 are demonstrably load-bearing — each is broken at `f70626452` and fixed at `21ecfc853`:

| probe | base | head |
|---|---|---|
| `items.Add(v)` then `items.Count` inside `First[T]` | `GS9999` | `2` |
| `items[0] = v` inside `SetFirst[T]` | `GS9999` | `99` |
| `items.Capacity = 16` inside `Grow[T]` | `GS9999` | `16` |

**Mutation (not just green).** Reverting **only** site 7 to its base form and rebuilding Release:

- real program flips `99` → `GS9999`; control probe (site 2) stays `2`, so the mutation is surgical
- `dotnet test --filter FullyQualifiedName~Issue2992SymbolicClrTypeInterpreterTests` → **`Passed! Failed: 0, Passed: 2`**
- restored → source md5 back to `33012571c9a3d767147d99908aab1caa`, program flips back to `99`

**The mutant survives.** The two shipped tests cannot see four of the seven sites this PR changes. Any later refactor may silently revert them.

**Fix:** add value-asserting tests for sites 2, 4, 5, 7 — the three probes above are ready-made — and either add one for site 1 (static member on a symbolic generic container) or state that no reachable shape exercises it.

---

## SHOULD-FIX

**S1 — static property/field *write* on a symbolic generic type cannot be reified.** `EvaluateClrPropertyAssignmentExpression` (`Evaluator.Expressions.cs:1403`) uses `ResolveRuntimeClrType(node.Receiver?.Type)` with no `?? StaticContainerType` fallback, while the matching **read** at `:1256` correctly uses `node.Receiver?.Type ?? node.StaticContainerType`. This is forced by the node model, not carelessness — `StaticContainerType` exists only on `BoundClrPropertyAccessExpression` and `BoundImportedCallExpression`, not on `BoundClrPropertyAssignmentExpression`. It still leaves a hole; please note it in the body or add the member.

**S2 — the `?? member` fallback in `ResolveMemberForRuntimeType` is dead code.** `Type.GetMemberWithSameMetadataDefinitionAs` **throws** `ArgumentException` when there is no match; it does not return `null`. The expression reads as a safe degradation and is not one. In practice the same-generic-definition guard above it should prevent a miss, which is precisely why the misleading fallback should go or become a real `try`/fallback.

**S3 — no caching on a hot path.** `ResolveMemberForRuntimeType` calls `GetMemberWithSameMetadataDefinitionAs` on **every** CLR call, property access and index with a generic receiver, per invocation. The non-generic early-out is cheap; the generic path is not. #2190 tracks interpreter performance — a small `ConcurrentDictionary<(MemberInfo, Type), MemberInfo>` would keep this from becoming that issue's next entry.

## NIT

**N1** — `RuntimeTypeArguments` is `static readonly` while the data is per-`Evaluator`. Harmless (the `ConditionalWeakTable` is keyed by frame reference and both call sites pass freshly-constructed frames, so `Add` cannot hit a duplicate key), but an instance field would scope it naturally.

**N2** — the `build` check was still `pending` at review time; `UNSTABLE` here means "not finished", not "failing".

---

**Verdict: NOT MERGEABLE.** The mechanism is correct, minimal and on the right axis — this is close. Blockers are the over-declared `Fixes` line and a test set that a mutant walks straight through.

DavidObando:
## Review: NOT MERGEABLE — 1 blocker

Measured on the **merged tree** (`main` `f820484c3` + head `770a9452e`), not the PR head in isolation. Merge base computed against a freshly fetched remote: `fd34166a8`; true three-dot scope **7 files / +506 −14**. The PR is 6 commits behind `main`, merges cleanly (`rc=0`), and the merged tree builds with `-p:ContinuousIntegrationBuild=true` at **0 warnings / 0 errors**.

**The fix is real and I confirmed it against the issue's own oracle.** On base `f820484c3`, `gsi samples/GenericTypeParameterAsTypeArgument.gs` exits **1** with exactly the reported `GS9999: ... List\`1[System.Object] does not match ... List\`1[System.Int32]`. On the merged tree it exits **0** and prints `10 / 20 / 42`, matching the compiled oracle in the issue. The probe is valid because it **fails at base** — I checked that before drawing anything from it.

`Interpreter.Tests`: **546 passed / 546 total** on the merged tree. No diagnostics are added (**0** new descriptors in the added lines), so no docs-gate exposure.

### BLOCKER — `Evaluator.Expressions.cs:1320` is load-bearing and pinned by no test

**How the sites were enumerated.** Rather than trusting a grep, I made the new `siteId` parameter on `ResolveMemberForRuntimeType` **required** and let the compiler enumerate the call sites. It reported exactly **7** `CS7036` errors — a compiler-verified denominator:

```
Evaluator.Calls.cs(860)  Expressions.cs(1236) (1320) (1323) (1433) (1616) (1632)
```

**Sentinel harness.** I then tagged each site `"1".."7"` behind a `GS_SITE_OFF` environment switch — one build, seven runs. With the variable unset the harness is provably inert (**546/546**, identical to clean).

| # | site | disabled → `Interpreter.Tests` | pinned |
|---|---|---|---|
| 1 | `Evaluator.Calls.cs:860` (static method call) | 1 failed / 546 | yes |
| 2 | `Evaluator.Expressions.cs:1236` (constructor) | 4 failed / 546 | yes |
| **3** | **`Evaluator.Expressions.cs:1320`** | **0 failed / 546** | **no** |
| 4 | `Evaluator.Expressions.cs:1323` (property read) | 8 failed / 546 | yes |
| 5 | `Evaluator.Expressions.cs:1433` | 2 failed / 546 | yes |
| 6 | `Evaluator.Expressions.cs:1616` (index read) | 1 failed / 546 | yes |
| 7 | `Evaluator.Expressions.cs:1632` (index write) | 1 failed / 546 | yes |

The harness is proven live rather than assumed: disabling **site 6** alone reproduces the issue's original `GS9999` on `samples/GenericTypeParameterAsTypeArgument.gs`. So the switch is read and effective, and a zero in that table means "no test noticed", not "the switch did nothing".

**Site 3 is not dead code.** It is the `receiver == null` early return in `ResolvePropertyOrFieldForReceiver` — a **static** property/field read on a constructed generic CLR type. (`member == null` is a no-op either way, so the live shape is the static one.) Probe:

```gs
func CmpName[T]() string {
    return Comparer[T].Default.GetType().Name
}
Console.WriteLine(CmpName[int32]())
```

| tree | output |
|---|---|
| merged head | `GenericComparer\`1` |
| merged head, site 3 disabled | `ObjectComparer\`1` |
| base `f820484c3` | `ObjectComparer\`1` |

Distinct values, and the base value is the erased-to-`object` defect this PR exists to fix. So site 3 fixes a genuine instance of #2992's family, and **removing it silently returns a wrong value that all 546 interpreter tests accept**. That is precisely the failure mode this repository has shipped eight times this session.

**The zero is absence, not scope — but only within the suite that can see it.** I ran `Core.Tests` twice with a built-in positive control: disabling site 3 gives **7033 passed / 7034**, and disabling the *known-pinned* site 4 gives **7033 passed / 7034** as well. Since site 4 is pinned by 8 interpreter tests yet invisible to `Core.Tests`, that suite does not exercise this path at all and its zero is uninformative in both directions. The finding rests on `Interpreter.Tests`, which demonstrably can discriminate six of the seven sites.

**Recommended fix:** add one distinct-value test for a static property/field read on a constructed generic imported type — the `Comparer[T].Default` shape above works and asserts `GenericComparer\`1` vs `ObjectComparer\`1`. Do not delete the site; it is doing real work.

### Where the site count went wrong

The body says *"remap **six** retained reflection sites"* and *"Each retained reflection site has a distinct-value test."* There are **seven** call sites. The "property read" category covers **two** of them — `:1320` and `:1323` — and the distinct-value test kills only `:1323`. The claim is true per *category* and false per *site*; the two populations were never the same, which is why the count closed while a site went uncovered.

I want to be clear that the rest of the body's methodology is better than most: distinct values (`11/33/44/55/66/77`), `GSharp.Core.dll` hashes rather than the driver, and it explicitly records **deleting** a proposed imported instance-method remap because reverting it survived all 7 examined tests. That is exactly the right instinct. Site 3 slipped through because it is the second call site inside a single named category.

### Also verified

- **Closure type-argument capture is pinned.** Neutering the `Locals`-walk in `CaptureRuntimeTypeArguments` (`Evaluator.Calls.cs`) → **1 failed / 546**, `DeferredClosureRetainsTypeArguments`. Build `rc=0` before believing the result.
- **Restore closes the cycle:** `git checkout -- src/` → porcelain 0, **0** `GS_SITE_OFF` remnants, rebuild `rc=0`, **546/546**.

### NIT

**`test/Interpreter.Tests/ProbeRefTypes.cs` conflicts with open PR #3032.** `git merge-tree ghr/pr3020 ghr/pr3032` → `rc=1`, `CONFLICT (content) in test/Interpreter.Tests/ProbeRefTypes.cs`. The two PRs share `Evaluator.Expressions.cs` (auto-merges) and this probe file (does not). #3048 also proposes reworking this same file. Whoever lands second resolves it; flagging so it is not discovered at merge time.

### Probe that failed to discriminate — reported rather than buried

Two attempted escalations of the site-3 probe produced **identical** results on all three trees and prove nothing: `items.Sort(c)` → `GS0159: Cannot find function Sort`, and `var c Comparer[int32] = Get[int32]()` → `GS0155: Cannot convert type ...`. Both are pre-existing binder gaps unrelated to this PR. The discriminating evidence is the `GetType().Name` probe above.

**Verdict: NOT MERGEABLE** on the single unpinned site. Everything else in the change is sound, the fix is genuine, and the remedy is one test.

DavidObando:
## Review: MERGEABLE — 0 blockers, 2 should-fix (both pre-existing / latent), 3 disclosures

Measured on the **merged tree** (`main` `185431c4` + head `c47416ac`), not the branch in isolation. Merge base computed against a freshly fetched remote: **`46d3cc50`**; true three-dot scope **7 files / +521 −14** (5 production, 2 test). `git merge-tree --write-tree` vs `185431c4` → rc 0, clean. Merged tree builds `rc=0 / 0 errors`.

**The briefed head `0a9ef0cc9` does not exist.** Real head is `c47416ac`. Everything below is true of `c47416ac` merged into `185431c4` and of nothing else — the PR is 1 merge-base behind `main`, so if `main` moves, these numbers expire.

---

## (a) Test-source shape — **NO `func Main()` anywhere.** This is the right shape.

`grep -c 'func Main'` over `Issue2992SymbolicClrTypeInterpreterTests.cs` = **0**, with a live positive control (`func ` matches 14×, so the grep is not an unrun check). All 12 test sources are **top-level statements**.

That matters more here than usual: `gsi` never invokes `func Main()`. I re-confirmed it on this tree — probe `c_main.gs` gives `gsi_rc=0` with **empty output** under both base and merged, while `gsc` runs it. An interpreter-targeted suite written inside `func Main()` would have measured nothing. This one does not have that defect.

## (b) Every probe fails at base — **12/12**, verified by reverting only `src/`

Reverted `src/` to `46d3cc50` while keeping the new tests. Build `rc=0 / 0 errors`; Core md5 `6c0fa766` → `f719f919`; `Core/` and `Compiler/` both fresh and byte-identical to each other. Result: **12 failed / 0 passed**. No test in this suite is vacuous.

I had statically predicted 9 of the 12 would be vacuous. **That prediction was wrong** and the measurement corrected it. Base failure modes are varied and genuinely discriminating:

| test | base symptom |
|---|---|
| `StaticPropertyRead` | ``ObjectComparer`1`` instead of ``GenericComparer`1`` — the literal #2992 erasure |
| `StaticClrCall` | `-11` instead of `11` |
| `StaticPropertyWrite` | **silent `111` → `0`** |
| the other 9 | `GS9999: Object type Sy…` |

The `GenericStaticSlot<T>.GetMarker() => typeof(T) == typeof(int) ? 11 : -11` fixture is deliberately discriminating rather than a look-alike value. Good fixture design.

## (c) Per-hunk mutation — **every hunk is pinned**

Release + `ContinuousIntegrationBuild=true` throughout; `BUILD_RC` checked before every hash; `Core/` **and** `Compiler/` mtimes verified fresh before every probe; md5 round-tripped back to clean after each restore.

| # | site | mutation | Core md5 | build | killed |
|---|---|---|---|---|---|
| — | clean head | — | `6c0fa766` | rc 0 | 12/12 pass |
| — | whole `src/` → base | — | `f719f919` | rc 0 | **12/12** |
| M1 | `Evaluator.Calls.cs` `ResolveMemberForRuntimeType` | `declaring?.IsGenericType != true` → `== true` | `ca432a99` | rc 0 | **12/12** |
| M2 | `Evaluator.Calls.cs` (2 sites) | `parameters.Length != arguments.Length` → `==` | `1e962ad8` | rc 0 | **12/12** |
| M3 | `ExpressionBinder.Assignments.cs:2341` | `staticContainerType: constructedImported.SymbolicReceiver` → `null` | `a985398f` | rc 0 | **1/12** — `StaticPropertyWriteUsesClosedContainer`, silent `111`→`0` |
| M4 | `Evaluator.Expressions.cs` | delete `RegisterRuntimeTypeArguments(frame, closure.CapturedTypeArguments)` | `f3d98368` | rc 0 | **2/12** — both `Deferred*RetainsTypeArguments` |

All condition inversions or statement/value removals in **product** code — no `if (true)` (CS0162 + warnings-as-errors would kill the build and yield no result), no test-fixture deletions. M3 and M4 are narrowly and *exactly* pinned by their own tests, which is the ideal outcome, not a weak one: the adjacent arms do not mask each other.

Round trip closed: `6c0fa766` → mutants → `6c0fa766`, `git status --porcelain` empty.

## (d) Compiler-counted sites — **the body's claim of 7 is exactly right**

Made the shared helper require a parameter (`private T ResolveMemberForRuntimeType<T>(T member, Type runtimeType, int siteId)`), rebuilt, collected `CS7036` deduped by `file:line`. 28 raw error lines → **7 unique sites**:

```
Evaluator.Calls.cs(919)
Evaluator.Expressions.cs(1275)
Evaluator.Expressions.cs(1359)
Evaluator.Expressions.cs(1362)
Evaluator.Expressions.cs(1472)
Evaluator.Expressions.cs(1655)
Evaluator.Expressions.cs(1671)
```

`Evaluator` is partial, so grep could have undercounted; here it did not, and the compiler confirms the denominator. The body's "exactly 7 unique `CS7036` errors" is verified independently.

## (e) Four columns — `{top-level, in-function} × {gsc, gsi}`

Probes assert on the reified **type** (`GetType().ToString()` / `.Name`), not on element values, so `List<object>` cannot masquerade as `List<int32>`.

| probe | base `gsc` | base `gsi` | merged `gsc` | merged `gsi` |
|---|---|---|---|---|
| `a_top` top-level | ``GenericComparer`1`` / `List`1[Int32]` | ``ObjectComparer`1`` / **`List`1[Object]`** | ✅ | ✅ |
| `b_infunc` in `func Run()`, called from file scope | ✅ | **`List`1[Object]`** | ✅ | ✅ |
| `c_main` in `func Main()` | ✅ | *(not executed — by design)* | ✅ | *(not executed)* |

The base row is the issue verbatim: `gsi` returns **rc=0** and silently produces the wrong runtime type. The fix closes the gap in all executed columns, and there is no top-level/in-function asymmetry — both shapes were broken at base and both are fixed.

I also probed shapes the suite does not cover; all correct at head and matching `gsc` exactly:

- **nested** instantiation (`Outer[int32]` calling `Inner[string]`) → `List`1[String] | List`1[Int32]`
- **recursive** with shadowed `T` (`Rec[int32]` → `Rec[string]`) → correct at both depths
- **four sequential** instantiations `int32/string/float64/int32` → all correct, so the `ConcurrentDictionary` member cache key is adequate
- **two live deferred closures** at different instantiations, interleaved → `Int32 / String / Int32`
- **two live method groups** off `Holder[int32]` / `Holder[string]`, interleaved → `Int32 / String / Int32`

**Suites:** full unfiltered `Interpreter.Tests` on the merged tree — **605 passed / 0 failed / 605 total**.

---

## SHOULD-FIX 1 — after this merges, the driver gap **inverts** for static writes: `gsc` silently drops them. Pre-existing; please file a follow-up.

`gsc` silently discards a static property write through a method type parameter. Measured with `/r:` against the PR's own fixture assembly:

```gsharp
func RT[T]() int32 {
    GenericStaticSlot[T].Value = 222
    return GenericStaticSlot[T].Value       // reads back 0
}
Console.WriteLine(RT[int32]())              // 0
Console.WriteLine(GenericStaticSlot[int32].Value)   // 0
Console.WriteLine(GenericStaticSlot[string].Value)  // 0
```

`gsc_compile_rc=0`, `gsc_run_rc=0`, output **`0|0|0`**. The immediate read-back is also `0`, so the write is **dropped entirely**, not redirected to `GenericStaticSlot<object>`.

**Control proves the probe is valid** — the same property written with a *concrete* type argument works:

```
GenericStaticSlot[int32].Value = 111  →  reads 111   ✅
SetValue[int32]()  (writes via [T])   →  still 111   ❌ (expected 222)
GenericStaticSlot[string].Value       →  0           (write went nowhere)
```

**This is pre-existing, not a regression.** Identical `111|111|0` from base `185431c4` and from the merged tree, using the same fixture assembly in both invocations. Consistent with the diff: it touches nothing under `src/Core/CodeAnalysis/Emit/`, and `EmitClrPropertyAssignment` (`MethodBodyEmitter.MemberAccess.cs:1108`) contains **zero** references to `StaticContainerType` — the new field is interpreter-only.

Why it is worth recording: #2992 is "`gsc` right, `gsi` wrong". After this PR, for the static-write shape specifically, it becomes "**`gsi` right, `gsc` wrong**" — same silent-wrong-answer signature, opposite driver. Not a reason to hold the PR (it strictly improves the state), but `Fixes #2992` should not be read as closing the family. A follow-up issue against the emitter is the right home.

## SHOULD-FIX 2 — the new `StaticContainerType` is dropped by every node-reconstruction site (latent today)

`BoundClrPropertyAssignmentExpression` gains `staticContainerType` as an **optional** eighth constructor parameter (defaulting to `null`), so every existing reconstruction still compiles while silently discarding it:

- `Binding/BoundTreeRewriter.cs:1678` — the generic rewriter; any pass that rewrites `Receiver` or `Value` loses the field
- `Lowering/SideEffectSpiller.cs:273` — rebuilds the assignment after spilling
- `Lowering/Async/SpillSequenceSpiller.cs:503` — same, in the async path
- `Binding/StatementBinder.Jumps.cs:774` — compound assignment builds the write node from a `BoundClrPropertyAccessExpression` without carrying its `StaticContainerType`

Relatedly, the comment at `Lowering/SideEffectSpiller.cs:346` now asserts something false: *"every parent-node reconstruction still goes through the normal Rewrite\* overrides, which already know how to preserve NarrowedType / InterfaceType / **StaticContainerType** — no risk of silently dropping a field here."* For this node, the override does not preserve it.

**Honest severity: I could not make this observable, and I tried four ways.** An `async` static write with `await` in the value (`SpillSequenceSpiller` path) **passes**; an explicit read-modify-write through `[T]` **passes**; the PR's own `StaticPropertyWriteUsesClosedContainer` uses a *call* as its value — which `SideEffectAnalyzer.HasObservableSideEffect` should flag — and still passes, which indicates the interpreter does not run `SideEffectSpiller` on this path at all. So this is a latent maintenance hazard, not a live bug. Threading the field through those four sites is a one-line-each change; making the parameter **required** would let the compiler enforce it permanently, which is the same technique the author already used to count the reflection sites.

---

## Disclosures — three of my own measurement errors, and one retraction

1. **My static hypothesis was wrong.** I predicted a first-wins bug in `CaptureRuntimeTypeArguments` (`captured.TryAdd` over `foreach (var frame in Locals)`) combined with `ResolveRuntimeTypeSymbol` returning the *first* matching frame, expecting a recursive or nested generic instantiated at two different type arguments to resolve to the wrong binding. **Five probes say no** — nested, recursive-with-shadowed-`T`, four sequential instantiations, two live closures, two live method groups all produce the correct type and match `gsc` exactly. The hypothesis is retracted.

2. **A compound-assignment finding I nearly filed was a false positive.** `GenericStaticSlot[T].Value += 22` fails with `GS0158: Cannot find member`. I ran a **non-generic control** — `PlainStaticSlot.Value += 22` on a plain static class — and it fails with the **same** `GS0158`. Compound assignment on any static CLR property is unsupported in the interpreter; it has nothing to do with this PR and I am not filing it.

3. **My first base-comparison probe was invalid and I re-ran it.** I compiled the `gsc` write probe against a base-built fixture assembly and got `GS0125: Variable 'GenericStaticSlot' doesn't exist` — because `GenericStaticSlot<T>` is *added by this PR*. The base-vs-merged comparison in SHOULD-FIX 1 was redone with the merged fixture assembly referenced from **both** invocations.

## What I did not run, and why

**Load reached 35.41** (peak of my window; it was 10–25 for the measurement phase) — above the 30 threshold, so I stopped building. Not run locally: `Compiler.Tests` and `Cs2Gs.Tests`. Residual risk is low but non-zero: the only non-evaluator production change is the single `ExpressionBinder.Assignments.cs:2341` argument, which adds a value to an existing constructor call inside an existing branch and cannot alter diagnostics; and the emitter provably never reads the new field. Note that CI green was earned on the branch at base `46d3cc50`, **not** on the merged tree — the merged tree is only covered by my `Interpreter.Tests` 605/605 and the build.

**Docs gate: N/A, with a denominator.** 7 files in the three-dot diff; **0** match `docs/|website/|Diagnostic|\.md$`; **0** new `"GS####"` literals introduced. Positive control: 12 `docs/*.md` exist in the tree, so the filter is live rather than empty by scope.

---

## Verdict

**MERGEABLE.** The mechanism is right and minimal — reification is applied only at reflection boundaries, the helper early-outs unless declaring and runtime types are constructions of the same generic definition, and the diff adds no diagnostics and touches no emit path. Every hunk is pinned by a test that fails without it; every test fails at base; the site count is compiler-verified at 7; the suite is top-level, which is the only shape `gsi` executes; and the four-column matrix shows `gsc` and `gsi` agreeing after the change where they disagreed before.

The two should-fix items are a pre-existing `gsc` defect and a latent field-threading hazard. Neither is caused by this PR and neither is observable in any behaviour it changes. I would merge this and open one follow-up issue for the `gsc` static-write loss.

DavidObando:
## Correction to my review — SHOULD-FIX 1 was mis-stated (verdict unchanged: this remains pre-existing)

Re-running my `gsc /r:` probes into **pristine empty directories** (per the fleet guidance that an output directory already holding a prior build of the same assembly identity can serve a stale assembly) turned up an error in my own earlier comment. The re-run used a **standalone fixture library built from neither tree**, so both columns reference byte-identical fixture bytes:

```
GenericStaticSlot<T> compiled into /tmp/fx4/bin/StandaloneFixture.dll  (md5 8d2a204a…)
each probe: rm -rf dir; mkdir; gsc /r:<fixture> /out:dir/p.dll; cp fixture dir/
resulting dir contains exactly 2 dlls — no GSharp.Core.dll, so isolation is total
```

Both trees verified fresh on **all three** `GSharp.Core.dll` copies — `Core/`, `Compiler/` **and `Repl/`** — same mtime per tree.

### What I got wrong

My comment stated the probe output was **`0|0|0`** and concluded the write was *"dropped entirely, **not** redirected to `GenericStaticSlot<object>`."*

The actual output is **`0|0|11`** (the third value is `GetMarker()`, which correctly returns `11`), and the inference was **backwards**. `[string]` reading `0` is correct behaviour, not evidence of a drop — the discriminating read is `[object]`, which I had not performed. Adding it:

```gsharp
func SetValue[T]() { GenericStaticSlot[T].Value = 111 }
SetValue[int32]()
Console.WriteLine(GenericStaticSlot[int32].Value)       // 0    ❌ expected 111
Console.WriteLine(GenericStaticSlot[string].Value)      // 0    ✅
Console.WriteLine(GenericStaticSlot[object].Value)      // 111  ← the write landed HERE
Console.WriteLine(GenericStaticSlot[int32].GetMarker()) // 11   ✅
```

`0|0|111|11` — **identical on `c47416ac` and on base `185431c4`.**

### Why the correction matters

The write is **not dropped — it is redirected to `GenericStaticSlot<object>`.** That is **#2992's own erasure symptom**: a generic type parameter used as a type argument reifying as `object`. This PR closes it in the interpreter; the **`gsc` emit path still exhibits it** for static property writes.

That is a more precise and more relevant residual than "the write is dropped", and it lines up with the structural evidence I gave before: the diff touches nothing under `src/Core/CodeAnalysis/Emit/`, and `EmitClrPropertyAssignment` (`MethodBodyEmitter.MemberAccess.cs:1108`) has zero references to the new `StaticContainerType`.

**The verdict is unchanged.** This reproduces identically at base, so it is pre-existing and not a regression from this PR. It is a candidate follow-up issue, not a blocker.

### Two further column facts from the same re-run

- **`gsi` has no `/r:` flag at all** — `Usage: gsi [file.gs] [--help] [--version]`. My earlier attempt to probe this shape through `gsi` produced a `GS0125`/`GS0159` cascade that was **my CLI error, not a `gsi` defect**; the flag was ignored. This shape is un-probeable in the `gsi` column by construction. The four-column results in my main review used built-in types and are unaffected.
- **Bare `gsc` (evaluate, no `/out:`)** on the same source fails with `error GS9999: The requested operation cannot be used on objects loaded by a MetadataLoadContext` — **identically on both trees**, so again pre-existing and not attributable to this PR.

### What did not change

All three previously-implicated probes reproduce **byte-identically** in clean rooms: `j_ctl` → `111|111|0` on both trees, `i_staticwrite` → `0|0|11` on both trees, matching the original recorded stdout exactly. **No result flipped from the directory pollution** — the error was in my transcription and inference, not in the measurement.

PR #3021, opened 2026-08-01
Reject target-typed lambdas with mismatched arity

### LINKED ISSUE #2937: gsc: mismatched inline-lambda arity through erased imported-generic slot compiles then throws

DavidObando:
## Summary
- reject target-typed lambdas when target and syntax arities differ
- reuse existing `GS0144` at `LambdaBinder.BindLambdaExpression`, shared by erased and non-erased target paths
- pin mismatched and matching arities
- reuse shared `IlVerifier.RunProcess` for bounded child execution; no duplicate drain helper or grace constant

Fixes #2937

## Root cause
`src/Core/CodeAnalysis/Binding/LambdaBinder.cs:628` used target parameter types only when arity matched, but silently ignored a mismatched target. Binding then fell back to erased `object`; compiler accepted an incompatible `Action<Src, Src>` for `Action<Src>` and runtime threw `ArgumentException`.

## Verification
Exact base `f70626452d6166811768ab4e500ac589878bd480` versus head `2c7049e64577b036c594e9dc36c25390de7ccde3`, using serial scratch-clone builds and each clone's `out/bin/Release/Compiler/gsc` directly:

| Probe | Base | Head |
|---|---|---|
| mismatched arity | compile exit 0; bounded run exits from `ArgumentException` | compile exit 1, `GS0144` |
| matching arity | compile/run exit 0, stdout `7` | compile/run exit 0, stdout `7` |

Targeted tests, rebuilt without `--no-build`:
- new mismatch + acceptance tests: 2 passed
- complete `Issue2918InlineLambdaErasedReceiverTests`: 13 passed
- rebased head `a1463d22d`: solution build 0 warnings/errors; 0/40 zero-byte Core DLLs; one MD5 group across all 5 ordinary paths

## Semantic mutations
| Mutation | Result |
|---|---|
| remove mismatch diagnostic | killed: mismatch changed `GS0144` to compile exit 0; restore changed back to `GS0144` |
| diagnose matching arity instead | killed: valid program changed stdout `7` to `GS0144`; restore changed back to stdout `7` |

Compiler MD5 round trips were exact: original/restored `41ba2355d9cbd3dea92c243f37f08ac3`; mutants `19b598aec7d030b54536c4caff6d3195` and `7cff7c3f972c35e24f2b5de2ed446ee2`.

## Collision notes
- `src/Core/CodeAnalysis/Binding/LambdaBinder.cs`
- `test/Compiler.Tests/Emit/Issue2918InlineLambdaErasedReceiverTests.cs`

### DIFF
--- a/src/Core/CodeAnalysis/Binding/LambdaBinder.cs
+++ b/src/Core/CodeAnalysis/Binding/LambdaBinder.cs
@@ -626,6 +626,15 @@ internal sealed class LambdaBinder
 
         var isAsync = syntax.IsAsync;
         var arityMatchesTarget = targetFunctionType != null && targetFunctionType.Arity == syntax.Parameters.Count;
+        if (targetFunctionType != null && !arityMatchesTarget)
+        {
+            Diagnostics.ReportWrongArgumentCount(
+                syntax.Location,
+                "lambda",
+                targetFunctionType.Arity,
+                syntax.Parameters.Count);
+        }
+
         var parameterTypes = ImmutableArray.CreateBuilder<TypeSymbol>(syntax.Parameters.Count);
         var parameterSymbols = ImmutableArray.CreateBuilder<ParameterSymbol>(syntax.Parameters.Count);
         var seen = new HashSet<string>();
--- a/test/Compiler.Tests/Emit/Issue2918InlineLambdaErasedReceiverTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2918InlineLambdaErasedReceiverTests.cs
@@ -27,8 +27,6 @@ namespace GSharp.Compiler.Tests.Emit;
 /// </summary>
 public class Issue2918InlineLambdaErasedReceiverTests
 {
-    private const int KillGraceMilliseconds = 5_000;
-
     private const string Contracts = """
         using System;
 
@@ -787,35 +785,10 @@ public class Issue2918InlineLambdaErasedReceiverTests
         startInfo.ArgumentList.Add(runtimeConfigPath);
         startInfo.ArgumentList.Add(assemblyPath);
 
-        using var process = Process.Start(startInfo)
-            ?? throw new InvalidOperationException("Failed to start dotnet exec.");
-        var stdoutTask = process.StandardOutput.ReadToEndAsync();
-        var stderrTask = process.StandardError.ReadToEndAsync();
-        _ = stdoutTask.ContinueWith(
-            task => _ = task.Exception,
-            TaskContinuationOptions.OnlyOnFaulted);
-        _ = stderrTask.ContinueWith(
-            task => _ = task.Exception,
-            TaskContinuationOptions.OnlyOnFaulted);
-
-        static string Drain(Task<string> task) =>
-            task.Wait(KillGraceMilliseconds)
-                ? task.Result
-                : "<unavailable: pipe still open after kill>";
-
-        if (!process.WaitForExit(30_000))
-        {
-            process.Kill(entireProcessTree: true);
-            process.WaitForExit(KillGraceMilliseconds);
-            throw new XunitException(
-                $"dotnet exec timed out.\nstdout:\n{Drain(stdoutTask)}\nstderr:\n{Drain(stderrTask)}");
-        }
-
-        var stdout = Drain(stdoutTask);
-        var stderr = Drain(stderrTask);
+        var (exitCode, stdout, stderr) = IlVerifier.RunProcess(startInfo, assemblyPath, 30_000);
         Assert.True(
-            process.ExitCode == 0,
-            $"dotnet exec exited {process.ExitCode}\nstdout:\n{stdout}\nstderr:\n{stderr}");
+            exitCode == 0,
+            $"dotnet exec exited {exitCode}\nstdout:\n{stdout}\nstderr:\n{stderr}");
         return stdout.Replace("\r\n", "\n", StringComparison.Ordinal);
     }
 
--- a/test/Compiler.Tests/Emit/Issue2918InlineLambdaErasedReceiverTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2918InlineLambdaErasedReceiverTests.cs
@@ -183,6 +183,57 @@ public class Issue2918InlineLambdaErasedReceiverTests
     }
 
     [Fact]
+    public void MismatchedInlineLambdaArityThroughErasedSlot_IsRejected()
+    {
+        const string source = """
+            package LambdaArityErasure
+            import System
+            import System.Collections.Generic
+
+            class Src {
+                let N int32
+                init(n int32) { N = n }
+            }
+
+            func Main() {
+                let callbacks = List[System.Action[Src]]()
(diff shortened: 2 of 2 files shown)


DavidObando:
## Rubber-duck review — PR #3021 at `2c7049e64`

**Verdict: MERGEABLE.** No blocking issues. The core change is nine lines, reuses an existing diagnostic, and I could not construct a false positive across seven probed shapes. One SHOULD-FIX about bundled scope, one NIT.

**Base is `f70626452` — the `main` tip. 0 behind / 2 ahead.** That is worth stating plainly because it *structurally exempts* this PR from two of the failure classes currently biting the fleet: there are no `main`-side deletions since its base (so the #3000 compile-break class is vacuous here, not merely clean), and its branch CI ran against the current assertions (so the "tightened test the branch never ran" class cannot apply). `git merge-tree` rc=0.

### Diagnostic ID — no allocation, so no BLOCKER

You asked for this first. **#3021 allocates no new ID.** Descriptor ID-literal set difference across `DiagnosticDescriptors.cs`, base vs head: **376 → 376**, `ADDED: []`, `REMOVED: []`. It reuses the existing `GS0144` (`DiagnosticDescriptors.cs:62`, `WrongArgumentCount`, severity `Error`).

Consequently it is **not exposed to the #2981 docs gate at all** — no descriptor added, and critically **no `docs/diagnostics.md` row deleted** (the diff touches exactly two files, neither under `docs/`). The backtick trap you flagged does not arise, since there is no row to write.

`GS9999` is not used anywhere in the change.

### The false-positive set — seven shapes, zero spurious rejections

This is the real risk on a rejection PR, so I went after it directly rather than trusting the body. The change reports unconditionally whenever a target function type exists and arity differs (`LambdaBinder.cs:629-636`), with no diagnostic isolation — so my hypothesis was that **overload resolution binding a lambda against non-selected candidates** would leak a spurious `GS0144`. `bindLambdaWithTarget` is wired into `OverloadResolver` (`Binder.cs:243`) and reached from `OverloadResolver.Invocations.cs:347` and `:511`, inside a `TryBindFunctionTypeArguments` that returns `bool` — exactly a per-candidate shape.

Measured base `f70626452` vs head `2c7049e64`, both compilers driven directly with `/out:`:

|---|---|---|---|---|
| `ok1` | trivial valid program | rc=0 | rc=0 | control — proves the harness compiles valid code |
| `fp1` | overloads `Action[int32]` / `Action[int32,int32]`, untyped lambda | `GS0266` | `GS0266` | same |
| `fp2` | overloads `Func` arity 2 / 3, untyped lambda | `GS0266` | `GS0266` | same |
| `fp3` | `xs.Select(x -> x + 1)` | rc=0 | rc=0 | same |
| `fp4` | `xs.Where(x -> x > 0)` | rc=0 | rc=0 | same |
| `fp6` | typed lambda, matching arity, assigned to `Action[int32]` | rc=0 | rc=0 | same |
| **`fp7`** | **`xs.Select((v, i) -> v + i)`** | rc=0 | rc=0 | **same — the decisive one** |
| **`fp5`** | mismatched arity through erased `List[Action[Src]]` slot | **rc=0** | **rc=1 `GS0144`** | **the intended rejection** |

**`fp7` is the result that closes the question.** `Enumerable.Select` genuinely carries `Func[T,R]` and `Func[T,int,R]` overloads — arity 1 and 2. Passing an arity-**2** untyped lambda selects the second overload and emits **no** spurious `GS0144` for the arity-1 candidate. So the resolver does not report against unselected candidates, and my hypothesis is disproved empirically rather than by reading.

`fp1`/`fp2` are worth noting separately: G# reports `GS0266` (ambiguous) for untyped lambdas against differing-arity overloads on **both** sides, so that shape never reaches the new code path at all.

`fp5` confirms the fix fires, with **exactly one** `GS0144`, correctly located at the argument:

```
fp5.gs(8,12,8,58): error GS0144: Function 'lambda' requires 1 arguments but was given 2.
```

So: **7 shapes probed, 1 behaviour change, and it is the declared target.** I did not find a program that compiled before and is now wrongly rejected.

---

### SHOULD-FIX — the tip commit bundles test infrastructure that duplicates #3007's territory

Commit `2c7049e64` ("test: bound process timeout recovery") is unrelated to lambda arity. It adds to `test/Compiler.Tests/Emit/Issue2918InlineLambdaErasedReceiverTests.cs`:

- `:30` — `private const int KillGraceMilliseconds = 5_000;`
- `:802` — `static string Drain(Task<string> task) => task.Wait(KillGraceMilliseconds) ? … : "<unavailable: pipe still open after kill>";`
- unobserved-`Exception` continuations on both read tasks
- bounded `WaitForExit(KillGraceMilliseconds)` after `Kill(entireProcessTree: true)`

This is a **near-verbatim re-implementation of logic that already exists** in `test/Compiler.Tests/IlVerifier.cs`, which has had `KillGraceMilliseconds` at `:47` and an identical `Grab(Task<string>)` helper at `:198` since #2966 — **with a different constant**: `10_000` there versus `5_000` here. Two copies of the same kill-and-drain recovery with divergent timeouts is how the next hang gets diagnosed against the wrong bound.

The change itself is correct and a genuine improvement (`WaitForExit(int)` does not drain redirected streams, so the bounded drain is right). The problem is attribution and duplication: **#3007 already owns this residual**, and folding it into a lambda-arity PR means it lands unattributed to either #2937 or #3007.

The body does disclose it in bullet 4, so this is not concealed scope — but bullet 4 describes work that has no corresponding `Fixes` line, and a reader diffing `src/` sees only the arity change. Either move it to #3007, or add the attribution explicitly.

### NIT — tests for #2937 live in #2918's file

Both new `[Fact]`s land in `Issue2918InlineLambdaErasedReceiverTests.cs`. The erased-slot mechanism is shared, so the placement is defensible, but a future bisect on #2937 will not find a file named for it.

---

### Note on the body's own evidence

The body's mutation table is sound and I did not need to re-derive it: both mutants killed, with md5 round-trips (`41ba2355…` original/restored, distinct mutant hashes). That is the standard the fleet asks for. My work above is deliberately the part the body does *not* cover — the false-positive set, which no mutation can reveal.

CI: `build`, `cs2gs`, `e2e`, all extension jobs SUCCESS at the time of writing; `UNSTABLE` reflected checks still in flight, not failures.

Method notes: three-dot diffs from a computed `git merge-base`; exit codes assigned to a variable before any other command; every "no problems" claim above carries the number of shapes examined, and the harness is proven live by `ok1` compiling and `fp5` flipping. Never `--no-build`; `-p:ContinuousIntegrationBuild=true`.

PR #3023, opened 2026-08-01
Preserve user TargetInvocationException in interpreter catches

### LINKED ISSUE #2995: gsi: Optional.FlatMap with a nullable-returning lambda throws TargetInvocationException

### LINKED ISSUE #2999: Interpreter cannot catch TargetInvocationException

DavidObando:
## Summary

- Preserve user-thrown `TargetInvocationException` during catch matching and handler binding.
- Continue recursively peeling reflection-created invocation wrappers; unknown/null target metadata defaults to wrapper behavior to preserve existing catches.
- Cover direct user throws, reflection wrappers around user TIEs, and the motivating `FormatException` reflection catch.

## #2995 scope measurement

At #3012 head `fae05b403`, plain `gsi samples/GsharpExtensionsOptional.gs` cannot load `Gsharp.Extensions` and exits `1` with 19 `GS0159` binding errors. Loading the already-built extension assembly into the same `gsi` process before evaluation produced:

```text
ADA
A
default
computed
<filtered out>
present: ada
ada
14
-1
7
count present: 7
```

Saved status: `0`. All 11 output lines match the issue's compiled output. #3012 therefore fixes #2995; this PR contains no #2995 or lambda-lowering changes.

## #2999 before/after

| Repro | Before | After |
|---|---|---|
| Exact user-thrown catch | `Value = null`, diagnostic `outer`, test exit `1` | `Value = true`, no diagnostics, exit `0` |
| Reflection wrapper around user TIE | wrapper diagnostic, exit `1` | caught user exception message `outer`, exit `0` |
| `FormatException` from `Convert.ToInt32("x")` | uncaught wrapper when reflection unwrapping is disabled | caught, `Value = true` |

## Verification

- Targeted tests: 3 examined, 3 passed, exit `0`.
- New `FormatException` test: baseline passed; mutant `IsReflectionInvocationWrapper => false` failed (`expected true`, `actual null`), exit `1`; restored test passed, exit `0`.
- Original unconditional-recursive-unwrapping perturbation: 2 examined, 2 failed, exit `1`; restored: 2 passed, exit `0`.

Fixes #2999

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Statements.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Statements.cs
@@ -697,8 +697,9 @@ public sealed partial class Evaluator
     /// EvaluatorException to attach node context (for GS9999 reporting when
     /// nothing catches it). That wrapping must not leak into typed catch
     /// matching or handler-variable binding, so unwrap repeatedly down to the
-    /// innermost real exception. TargetInvocationException (thrown by
-    /// reflection-based CLR calls) gets the same treatment.
+    /// real exception. Reflection-generated TargetInvocationException wrappers
+    /// get the same treatment, while a user-thrown TargetInvocationException
+    /// remains visible to catch matching and handler-variable binding.
     /// </summary>
     private static Exception UnwrapRuntimeException(Exception ex)
     {
@@ -710,7 +711,8 @@ public sealed partial class Evaluator
                 continue;
             }
 
-            if (ex is TargetInvocationException { InnerException: { } tieInner })
+            if (ex is TargetInvocationException { InnerException: { } tieInner } tie
+                && IsReflectionInvocationWrapper(tie))
             {
                 ex = tieInner;
                 continue;
@@ -719,6 +721,9 @@ public sealed partial class Evaluator
             return ex;
         }
     }
+
+    private static bool IsReflectionInvocationWrapper(TargetInvocationException ex)
+        => ex.TargetSite?.DeclaringType?.Assembly == typeof(TargetInvocationException).Assembly;
 
     private static bool TryFindCatchHandler(BoundTryStatement node, Exception ex, out BoundCatchClause matched)
     {
--- a/src/Core/CodeAnalysis/Evaluator.Statements.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Statements.cs
@@ -723,7 +723,8 @@ public sealed partial class Evaluator
     }
 
     private static bool IsReflectionInvocationWrapper(TargetInvocationException ex)
-        => ex.TargetSite?.DeclaringType?.Assembly == typeof(TargetInvocationException).Assembly;
+        => ex.TargetSite?.DeclaringType?.Assembly is not { } assembly
+            || assembly == typeof(TargetInvocationException).Assembly;
 
     private static bool TryFindCatchHandler(BoundTryStatement node, Exception ex, out BoundCatchClause matched)
     {
--- a/test/Interpreter.Tests/Issue2999TargetInvocationExceptionInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue2999TargetInvocationExceptionInterpreterTests.cs
@@ -1,0 +1,88 @@
+// <copyright file="Issue2999TargetInvocationExceptionInterpreterTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using GSharp.Core.CodeAnalysis;
+using GSharp.Core.CodeAnalysis.Compilation;
+using GSharp.Core.CodeAnalysis.Symbols;
+using GSharp.Core.CodeAnalysis.Syntax;
+using Xunit;
+
+namespace GSharp.Interpreter.Tests;
+
+/// <summary>
+/// Issue #2999: distinguish reflection wrappers from user exceptions.
+/// </summary>
+public class Issue2999TargetInvocationExceptionInterpreterTests
+{
+    [Fact]
+    public void UserThrownTargetInvocationException_IsCatchable()
+    {
+        const string Source = """
+            import System
+            import System.Reflection
+
+            let handler () -> int32 = () -> throw TargetInvocationException(
+                "outer",
+                InvalidOperationException("inner"))
+
+            var caught = false
+            try {
+                handler()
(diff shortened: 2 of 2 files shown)


DavidObando:
## Review: MERGEABLE. No blockers. Two cheap hardening items below.

### Measurements
- `git merge-tree --write-tree 4b3ebf24f 0e123d7a0` → rc=0, clean.
- Scope, measured `MB=$(git merge-base 4b3ebf24f 0e123d7a0)` (`210126dd9`, = #3014) → `git diff $MB 0e123d7a0`: **2 files, +96/−3**. Re-merging the head onto `4b3ebf24f` in a scratch worktree gives the identical 2 files / +96/−3, so no upstream drift is hiding in the number. Declared scope == actual scope; `Fixes #2999` present, and the #2995 analysis in the body correctly declines to claim an issue this diff does not touch.
- `dotnet build GSharp.sln -c Release -m:1 -p:ContinuousIntegrationBuild=true` → rc=0, `0 Warning(s)`, `0 Error(s)`.
- `Issue2999...Tests` → **2 examined, 2 passed**, rc=0.
- Full `Interpreter.Tests` baseline → **499 examined, 499 passed**, rc=0.

### Root cause, not symptom — confirmed
`UnwrapRuntimeException` has exactly **2** callers, both in the file being edited: `src/Core/CodeAnalysis/Evaluator.Statements.cs:623` (handler-variable binding) and `:725` (catch matching). The guard goes into the shared function, so both paths are fixed by one edit. No sibling caller left behind. The `GS9999` reporting path at `Compilation.cs:326-336` does not call it, so uncaught-exception reporting is unchanged.

### Mutation — the guard is load-bearing in both directions
The PR body already perturbed toward unconditional unwrapping (old behaviour) and got 2/2 failures. I perturbed the other way, `IsReflectionInvocationWrapper => false`, against the **same 499-test population**:

| | Passed | Failed | Total |
|---|---|---|---|
| baseline | 499 | 0 | 499 |
| `=> false` | 498 | **1** | 499 |

Failing test: `ReflectionWrapper_UnwrapsOnlyToUserTargetInvocationException`. Restored source + rebuild → back to 499/499.

Behavioural perturb/restore on the built `gsi`, using a script the suite does not contain:

```gsharp
var msg = "NOT-CAUGHT"
try { Convert.ToInt32("x") } catch (ex FormatException) { msg = "CAUGHT" }
Console.WriteLine(msg)
```

- mutated: `rc=1`, `GS9999: Exception has been thrown by the target of an invocation.`
- restored: `rc=0`, `CAUGHT`

Flip, restore, flip back. The guard is real.

---

## SHOULD-FIX 1 — a live behaviour is now unpinned by 499 tests
That last experiment is the finding. Catching a CLR-thrown exception by its real type in G# depends entirely on the new `IsReflectionInvocationWrapper` returning `true`, and **0 of 499** Interpreter.Tests notice when it stops doing so. Before this PR that behaviour was unconditional; now it rides on a heuristic, and its failure mode is silent (`GS9999` with the wrapper's message instead of a matched `catch`).

**Fix:** add the six lines above as a third `[Fact]` in `test/Interpreter.Tests/Issue2999TargetInvocationExceptionInterpreterTests.cs`. It needs no new probe type and it is the cheapest possible pin for the behaviour this PR narrows.

## SHOULD-FIX 2 — null `TargetSite` falls to the unsafe side
`Evaluator.Statements.cs:725-726`:

```csharp
=> ex.TargetSite?.DeclaringType?.Assembly == typeof(TargetInvocationException).Assembly;
```

Both `?.` short-circuits yield `null`, which compares unequal, so **"I don't know" is classified as "user-thrown" and unwrapping stops** — the direction that silently breaks existing programs. `TargetSite` is derived from throw-site metadata and is not guaranteed non-null (unthrown/rethrown/serialized exceptions; dynamic-method throw sites have a null `DeclaringType`).

**Fix, same line:** treat unknown as a wrapper, e.g. `=> ex.TargetSite?.DeclaringType?.Assembly is not { } a || a == typeof(TargetInvocationException).Assembly;`. I checked this does not weaken the PR's own case: in `UserThrownTargetInvocationException_IsCatchable` the TIE is thrown by the evaluator at `Evaluator.Statements.cs:689`, so `TargetSite.DeclaringType.Assembly` is `GSharp.Core` — non-null and not CoreLib — and the test still passes by the same reasoning.

## NIT
- The doc comment rewrite at `:697-702` is accurate and matches the new behaviour. No stale sentence left behind.

Issue #2988, opened 2026-08-01
gsi: deinitializers never run


DavidObando:
## Problem

Deinitializers never run under the interpreter.

`samples/Deinit.gs`:

```
compiled: Allocated: db/users / CachedResource deinit: users / Resource deinit: db / done
gsi:      Allocated: db/users / done
```

Exit code 0, no diagnostic — the two `deinit` bodies are simply skipped.

## Root cause

`grep -rn Deinit src/Core/CodeAnalysis/Evaluator*.cs` returns nothing: the evaluator has no deinitializer support. Deinit is bound (`Binding/DeclarationBinder.Constructors.cs`) and emitted, but the interpreter never invokes it — neither at scope exit nor at any other point.

## Impact

Silent divergence, not an error. Any sample or program relying on deterministic cleanup behaves differently under `gsi`.

## Note on scope

Deciding *when* the interpreter runs a deinit is the substance of the fix — the emitted semantics are the specification, so mirroring the emitter's placement is the target. A "not supported" diagnostic is not adequate here because the construct binds and the program otherwise runs to completion.

Found while triaging #2939; `samples/Deinit.gs` is excluded there until this is fixed.

DavidObando:
## Triage: confirmed. The "design question" is already settled by the emitter — and the obvious answer is the wrong one.

**Repro confirmed** on `samples/Deinit.gs`:

```
compiled: Allocated: db/users / CachedResource deinit: users / Resource deinit: db / done
gsi:      Allocated: db/users / done                                        rc=0
```

**Binary provenance:** probed with binaries built at `8dddbbf28`. The main checkout's `out/bin/Release` binaries are **stale relative to `a6b1561f2`** (built `20:45`; `a6b1561f2` landed `21:13`) — proven behaviourally: #2961's `ConditionalFunctionLiteral` case, which must fail with `error GS0100` and emit nothing, compiles rc 0 there. `8dddbbf28` is sound for this issue because `git diff --name-only 8dddbbf28 a6b1561f2 -- 'src/Core/CodeAnalysis/Evaluator*' 'src/Interpreter/**' 'src/Repl/**'` is **empty**.

---

## Compiled semantics: GC finalization, non-deterministic. Not scope exit.

[...]

Issue #2987, opened 2026-08-01
gsi: Span/ReadOnlySpan calls fail with 'Specified method is not supported'


DavidObando:
## Problem

Calls involving `Span[T]` / `ReadOnlySpan[T]` fail in the interpreter with `GS9999: Specified method is not supported.`

```
samples/RefStructSpan.gs   compiled: hel / 3      gsi: GS9999: Specified method is not supported.
samples/SpanIndexing.gs    compiled: 60 / 402     gsi: GS9999: Specified method is not supported.
```

## Root cause

`NotSupportedException` with that exact message is what `System.Reflection`'s invoke path raises for methods whose signature contains a **`ByRefLike`** type. The interpreter dispatches CLR member calls through reflection, so any `Span`/`ReadOnlySpan` parameter or receiver is unreachable that way.

This is a structural limitation of a reflection-based evaluator, not a one-line bug — hence a separate issue rather than a line fix.

## Options

1. Special-case the small set of `Span`/`ReadOnlySpan` operations the language actually surfaces (indexer, `Slice`, `Length`, the implicit `[]T` → span conversion) with direct interpreter intrinsics.
2. Declare ref-struct interop a compiled-only boundary and report a dedicated diagnostic instead of leaking the CLR message.

Option 2 is cheap and makes the boundary honest; option 1 is the real fix.

Note that `samples/UserRefStruct.gs` and `samples/RefIncrement.gs` **pass**, so user-defined ref structs are fine — it is specifically the BCL `Span` family that is unreachable.

Found while triaging #2939; both samples are excluded there until this is resolved.

DavidObando:
## G1 design — the evaluator's reflection chokepoint

Probed on `main` `a6b1561f2` against `out/bin/Release/Repl/gsi` directly. No builds. Read-only.

---

### Correction to this issue's body, first — it points at the wrong sites

> Special-case the small set of `Span`/`ReadOnlySpan` operations the language actually surfaces (indexer, `Slice`, `Length`, the implicit `[]T` → span conversion)

The indexer/`Slice`/`Length` framing is wrong. **The bare conversion fails with the span never used:**

```gs
func f(v []int32) int32 {
    var s ReadOnlySpan[int32] = v
    return 7                        // ← GS9999, span never read
}
```

Two *distinct* bound-node kinds fail, proven by `gsc`'s node dump:

```
g.gs → System.MemoryExtensions.AsSpan(t)                  ← call site,       Calls.cs:863
[...]

PR #3024, opened 2026-08-01
Define gsi unsafe and native boundaries

### LINKED ISSUE #2986: gsi: P/Invoke calls silently return the default value instead of diagnosing
## Problem

The interpreter has no P/Invoke support, and calling a `DllImport`/`LibraryImport` function **silently returns the return type's default** instead of reporting that it is unsupported.

| sample | compiled | `gsi` |
|---|---|---|
| `PInvoke.gs` | `13` | `0` |
| `PInvokeLibraryImport.gs` | `13` | `0` |
| `PInvokeLibraryImportStringReturn.gs` | `round-trip!` | *(empty line)* |
| `PInvokeMarshalAs.gs` | `13`, `5` | `0`, `0` |
| `PInvokeStructMarshalling.gs` | `3`, `4`, `6153737367135073092` | `3`, `4`, `0` |
| `PInvokeRefOutIn.gs` | `True True 0 True` | `False True` then `GS9999: StructValue cannot be converted to Int32` |

All exit 0 with no diagnostic except the last.

## Root cause

`Evaluator*.cs` contains no `DllImport` / extern handling at all (`grep -rn DllImport src/Core/CodeAnalysis/Evaluator*.cs` → no matches). An extern declaration is bound by `Binding/PInvokeBinder.cs` and reaches `src/Core/CodeAnalysis/Evaluator.Calls.cs:89` (`program.Functions[node.Function]`) with an effectively empty body, so `EvaluateFunctionBody` returns null, which converts to the CLR zero.

## Proposed fix — the small one first

Full marshalling support is optional. The **defect** is the silence: a call the interpreter cannot perform must not return a plausible-looking wrong value. Report an unsupported diagnostic at the call (roughly ten lines at the `program.Functions` lookup, gated on the function carrying P/Invoke metadata), so `gsi` fails loudly.

Real support via `NativeLibrary.Load` + `Marshal.GetDelegateForFunctionPointer` can follow separately if wanted.

Note `samples/PInvokeFunctionPointer.gs` currently **passes**, so the boundary is not uniform.

Found while triaging #2939; the six samples above are excluded there until this is resolved.

DavidObando:
## Summary

- enforce the interpreter P/Invoke boundary at `Compilation.Evaluate`, shared by `gsi` and `gsc /i` but not emit
- report `GS0514` (Error) before evaluation and direct users to compile with `gsc`
- route batch `gsi` diagnostics through the shared rich renderer, preserving filenames and source locations
- document `GS0514` in both diagnostic references and record the native-call policy in ADR-0152

## Declaration-time refusal

`GS0514` intentionally applies even when a P/Invoke declaration is not called. The interpreter cannot create a valid callable value for it, and first-class references can escape before an indirect call. Refusing the declaration at the shared evaluation boundary prevents both fabricated direct-call results and delayed `GS9999` failures without adding a native or reflection dispatch path.

## Verification

- base: `f1656079a2a109aade66bb5ee4ec1d3833246414`
- merge-tree: clean; merged tree equals HEAD tree
- clean solution build: 0 warnings, 0 errors
- merged-tree diagnostic gate: 4/4
- affected interpreter tests: 25/25
- full repository test matrix: 14,225 passed, 0 failed, 1 skipped, 14,226 total
- `cs2gs-corpus`, `cs2gs-oahu`, and `e2e`: passed
- P/Invoke-loop mutant: 15 failed / 25 examined
- rich-renderer mutant: owning test failed 1/1
- filename-propagation mutant: owning test failed 1/1
- all mutants restored exactly
- fresh full-solution driver build verified by Core and `gsi` DLL mtimes
- fresh-driver `gsi samples/PInvoke.gs`: exit 1 with located error `GS0514`
- `gsi samples/Deinit.gs`: unchanged exit 0 with no `GS0515`; #2988 is owned by #3052

The first CI attempt hit the pre-existing racy `BareStaticField_InInstanceMethod_InsideGoScope_Works` test (`2` observed instead of `3` from three concurrent non-atomic increments). Its focused local rerun passed 1/1, and the failed-job rerun passed all 7,069 Core tests. All final checks are green.

Fixes #2986.


DavidObando:
## Review: NOT MERGEABLE as-is — 4 blockers. The design is right; the ID allocation and one undeclared change are the problems.

### Measurements
- `git merge-tree --write-tree 4b3ebf24f 3c54a2fec` → rc=0, clean.
- Scope, measured `MB=$(git merge-base 4b3ebf24f 3c54a2fec)` (`f70626452`) → `git diff $MB 3c54a2fec`: **16 files, +398/−107**.
- Build (this head merged with #3018) → rc=0, `0 Warning(s)`, `0 Error(s)`.
- `Issue2986And2988InterpreterBoundaryTests` → **4 examined, 4 passed**, rc=0.
- Behaviour verified on the built `gsi`:
  - `gsi samples/PInvoke.gs` → `rc=1`, `samples/PInvoke.gs(20,6,20,18): error GS0505: P/Invoke function 'NativeStrLen' is not supported by the interpreter; ...`
  - `gsi samples/Deinit.gs` → `rc=0`, `samples/Deinit.gs(22,5,22,11): warning GS0507: ...` and `(34,5,34,11)` for `CachedResource`

### What is genuinely good here (worth keeping exactly as-is)
- **The boundary sits at the right chokepoint.** `Compilation.Evaluate` (`src/Core/CodeAnalysis/Compilation/Compilation.cs:314-336`) is reached by `gsi` via `src/Repl/Engine/SessionEngine.cs:210` **and** by `gsc`'s interpret mode via `src/Compiler/Program.cs:308-311`, and is **not** reached by `Emit`. I grepped all 6 `.Evaluate(` call sites to confirm. So `gsc` compilation of P/Invoke is unaffected and both interpreters get the boundary from one edit. Root cause, not symptom.
- **The `SyntaxTrees.Contains(...)` guard is correct and load-bearing.** `ContinueWith` (`Compilation.cs:282-285`) puts only the new trees into `SyntaxTrees`, so a REPL cell will not re-warn for a class declared in an earlier cell, and P/Invoke in a referenced assembly never reports.
- Docs rows use the bare `| GS0505 | Error |` / `| GS0507 | Warning |` form and severities match the descriptors, so the `DiagnosticIdUniquenessTests` regex at `test/Core.Tests/CodeAnalysis/DiagnosticIdUniquenessTests.cs:150` sees them.

---

## BLOCKER 1 — self-allocated reserved diagnostic IDs
`GS0507` (`src/Core/CodeAnalysis/DiagnosticDescriptors.cs:380`) and `GS0509` (`:382`) are both on the maintainer's do-not-allocate list (`GS0507`, `GS0509`–`GS0513`). No agent may allocate an ID without asking. That list is live, not theoretical:
- `GS0510` → claimed by open PR #3005
- `GS0513` → claimed by open PR #3028
- `GS0507` → PR #2912 states "GS0507 removed and returned to the reserved pool" / "GS0507 remains removed/reserved", i.e. it is a **recycled** ID that previously meant something else

The PR body calls `GS0509` a "free ID". It is not.

**Fix:** ask before allocating, and take whatever is handed back.

## BLOCKER 2 — repurposing a diagnostic ID that already shipped on `main`
`GS0505` is `UnconstrainedNullableIteratorCall` on `main` today (`DiagnosticDescriptors.cs:378`, documented at `docs/diagnostics.md:1102-1106`). This PR reassigns `GS0505` to P/Invoke and renumbers the shipped diagnostic to `GS0509`. Anyone with `GS0505` in a `/nowarn` list silently starts suppressing a different error.

The body justifies this as vacating "the reserved GS0505", but there is no such reservation in the repo — `grep -in reserv docs/diagnostics.md` returns 5 hits (`:182`, `:188`, `:574`, `:578`, `:604`), none about `GS0505`.

**Fix (and it is the lazier one):** leave `GS0505` alone entirely. Take one newly-assigned free ID for P/Invoke and one for `deinit`. That deletes the renumber, the `docs/diagnostics.md:1110-1117` edit, and the whole `test/Compiler.Tests/Emit/Issue2927NullableSequenceElementTypeTests.cs` churn from the diff.

## BLOCKER 3 — ADR number 0152 collides with PR #3018
#3018 adds `docs/adr/0152-interpreter-compiled-only-storage-boundary.md`. Filenames differ, so `git merge-tree --write-tree 335dde896 3c54a2fec` → rc=0 and the merged tree has `ls docs/adr | grep -c '^0152'` → **2**. Highest on `main` is 0151. Nothing in CI catches it.

Beyond the number: the two ADRs both legislate `fixed` and unsafe `&`/`*` (this one at :63–70, #3018 at :52 and :57–60). Two Accepted ADRs on the same boundary is worse than either alone. One of them should shrink — the natural split is #3018 owning the storage/pointer family and this one owning P/Invoke + `deinit` only, cross-referencing rather than restating.

## BLOCKER 4 — the renderer change breaks #3018, and is undeclared scope
`src/Repl/Program.cs:58-61` swaps `Console.Error.WriteLine($"{d.Id}: {d.Message}")` for `Console.Error.WriteDiagnostics(cell.Diagnostics)`. Measured on a real merge of both heads:
- #3018 alone: `Issue2956InterpreterBoundaryTests` **8 examined, 8 passed**
- #3018 + this PR: **8 examined, 7 failed, 1 passed**

```
Expected: "GS9999: 'fixed' (pinning) statements are "...
Actual:   "\n(1,1,4,1): error GS9999: 'fixed' (pinnin"...
```

Neither #2986 nor #2988 is about diagnostic rendering, and `GS0505`/`GS0507` would still be correct (just unlocated) without it — so this is a separable change riding along. It also has **zero** test coverage of its own in this PR: I grepped `test/` for `Repl.Program` and found no call site on `main`, so the only thing that exercises batch `gsi` stderr formatting is #3018's new file, which this breaks.

**Fix:** split the renderer change into its own PR with its own test, or keep it and update `Issue2956InterpreterBoundaryTests` + ADR-0152 (#3018) lines 29–33 in the same commit.

---

## SHOULD-FIX

**5. ADR :83–85 is overbroad and false for `GS9999`.** "Diagnostics use real declaration locations. Batch `gsi` renders them through the shared diagnostic renderer, including file, line, severity, and source excerpt." Measured on the merged build, a `fixed` program gives:

```
(1,1,4,1): error GS9999: 'fixed' (pinning) statements are not supported in the interpreter; ...
    fixed p *int32 = values {
```

**No filename**, and `(1,1,4,1)` is not a location in the user's file — `Compilation.cs:328-333` synthesizes a fresh `SourceText` from `ex.Node.WriteTo(...)`, so the line/column and the excerpt both refer to the reprinted bound node. Scope the sentence to the new declaration diagnostics.

**6. ADR :79–82 asserts an exclusion from a harness that does not exist.** "`samples/Deinit.gs` is a permanent exclusion from interpreter golden-output conformance." There is no interpreter golden-output conformance harness: `test/Compiler.Tests/LanguageConformance/SampleConformanceTests.cs:14-24` compiles through `gsc.dll` and runs under `dotnet`, and its only exclusion set is `WindowsSkippedSamples` (`:30-36`). Grepping `test/` and `e2etests/` for `gsi` returns exactly one file (`SessionEngineConsoleIoTests.cs`), unrelated. This PR's diff adds no exclusion entry anywhere, so the decision has no enforcement point — when #2939 builds the harness, nothing points at this ADR.

**Fix:** one clause — "when the #2939 interpreter conformance harness is built, `samples/Deinit.gs` must be excluded" — or add the entry where it will live.

**7. `IsPInvoke` + `deinit` reporting has no negative test.** The four rewritten files pin the positive direction well, but nothing pins that a *non*-P/Invoke function and a class *without* `deinit` stay silent, and nothing pins the `SyntaxTrees.Contains` guard (which is the subtle part — drop it and REPL cells re-warn forever). One test that evaluates two cells and asserts the second reports no `GS0507` would pin the guard that is easiest to "simplify away" later.

DavidObando:
## Review: NOT MERGEABLE — 3 blockers

Measured against `main` `f820484c3`. Merge base computed from a freshly fetched remote: `26b55f562`; true three-dot scope **15 files / +403 −97**. The PR is **17 commits behind** `main`.

**The underlying fix is genuine and well pinned.** I confirmed the silent defect on base and its repair on the merged tree, and every production hunk dies under mutation. The blockers are all about landing it, not about whether it works.

### Confirmed working

**The silent defect is real and is fixed.** On base `f820484c3`:

| program | base | merged |
|---|---|---|
| `gsi samples/PInvoke.gs` | rc **0**, prints `0` | rc **1**, located `GS0507` |
| `gsi samples/Deinit.gs` | rc 0, no mention of skipped `deinit` | rc 0, two located `GS0509` warnings, evaluation continues |

Base printing `0` at exit 0 is exactly the silent wrong-value class #2986 describes. The boundary is **honest**: `GS0507` is a real designed refusal at `Error`, `GS0509` a non-blocking `Warning`, and neither leaks through the `GS9999` internal-error bucket. Severities agree between descriptor and docs row.

**All four production hunks are pinned** (baseline **26 passed / 26**; every mutant build checked for `rc=0` *before* its hash was believed):

| mutant | result |
|---|---|
| remove P/Invoke loop, `Compilation.cs` | **15 failed / 26** |
| remove `deinit` loop, `Compilation.cs` | **2 failed / 26** |
| revert rich renderer, `Program.cs:58-62` | **1 failed / 26** — `BatchFileRunner_RendersBoundaryLocationsAndUsesSeverityExitCodes` |
| drop filename threading, `SessionEngine.cs:212` | **3 failed / 26** |

Restore → **26/26**. The descriptor and `DiagnosticBag` reporter hunks are pinned by the (now total) `DiagnosticIdUniquenessTests`.

*Methodology note, since it nearly misled me:* my first restored-tree hash did not match the recorded clean hash. The build is in fact deterministic (`02dd8fe1` stable across three passes including a forced recompile) — the mismatch was that the "clean" hash had been taken after a `dotnet test` run, and **`dotnet test` rebuilds without `-p:ContinuousIntegrationBuild=true`**. Hashes straddling that property are not comparable.

**ADR `0152` is unique.** `main` has `0151` then `0153`, so this fills a real gap, and a scan of all **24** open PRs found `0152` claimed by this PR alone.

---

### BLOCKER 1 — the docs gate is red on `main`, green on this branch

`website/docs/ref/diagnostics.md` is **not touched** by this PR. #3043 made the gate bidirectional. Run on the **merged** tree:

```
Every_Diagnostic_Is_Documented_In_Both_References_With_Matching_Severity [FAIL]
  website/docs/ref/diagnostics.md is out of sync with compiler diagnostics:
  Missing from website/docs/ref/diagnostics.md: GS0507, GS0509
```

**1 failed / 4 total.** And it is green on the branch for a structural reason, not by luck: the gate at this PR's head is still the single-file `Every_Descriptor_Is_Documented_With_Matching_Severity` with **0** occurrences of `website`, whereas `main`'s has **1**.

The `docs/diagnostics.md` rows themselves are correct — bare `| GS0507 | Error |` and `| GS0509 | Warning |`, no backticks, so they are gate-visible.

**Fix:** add both rows to `website/docs/ref/diagnostics.md` in **3-column** shape, matching their neighbours `GS0508`/`GS0513` at `:1743-1744`. That file is mixed-shape (of 384 ID rows: 228 three-column, 155 four-column, 1 outlier), so copy the adjacent rows rather than a distant table.

### BLOCKER 2 — does not merge into current `main`

`git merge ghr/pr3024` into `f820484c3` returns **rc=1**: `CONFLICT (content) in src/Core/CodeAnalysis/DiagnosticDescriptors.cs`. It is a trivial adjacency conflict — `main` gained `GS0513` at the same insertion point — but GitHub currently reports `mergeable=UNKNOWN`, so this will not resolve itself. Rebase onto `f820484c3`. (I resolved it locally by keeping both in ID order to obtain the tree everything above was measured on.)

### BLOCKER 3 — the diagnostic IDs are not the allocated ones

This PR ships **`GS0507`** and **`GS0509`**. The IDs allocated for it were **`GS0514`** and **`GS0515`**. Diagnostic IDs are centrally allocated in this repository and both `GS0507` and `GS0509` sit in the quarantined range — `GS0507` specifically because it was claimed and then withdrawn on an unpushed branch.

For the record, neither is occupied on `main` today: with a positive control confirming the pattern works (`new DiagnosticDescriptor(` matches **0**; the correct target-typed `new("GS####"` matches **377**), `GS0507`, `GS0509`, `GS0514` and `GS0515` are all absent from `src/`. So this is not a live compile collision — it is an allocation-authority problem, and the withdrawn-then-reused shape is exactly the kind that collides invisibly later.

---

### SHOULD-FIX 1 — the P/Invoke boundary over-fires on declarations that are never called

The loop reports on every declaration: `program.Functions.Keys.Where(f => f.IsPInvoke && ...)` in `Compilation.cs`. A program that *declares* a P/Invoke it never calls previously ran fine and now refuses to run:

```gs
@DllImport("libc", EntryPoint: "strlen", CharSet: CharSet.Ansi)
func NativeStrLen(text string) nint;

Console.WriteLine(42)
```

| tree | result |
|---|---|
| base `f820484c3` | rc **0**, prints **`42`** |
| merged | rc **1**, `GS0507` |

Nothing native ever executes, so there is no wrong value to prevent. #2986 is about P/Invoke **calls** silently returning defaults. The PR's own design shows the alternative: `deinit` got a non-blocking **warning** and the program still runs. The asymmetry is the argument — either report at the call site, or make the unused-declaration case a warning. As it stands this breaks working programs, and it is undeclared in the body.

### SHOULD-FIX 2 — duplicate coverage of #2988 with open PR #3052

This PR declares `Closes #2986, #2988` and implements the `deinit` warning as **`GS0509`**. Open PR **#3052** ("Warn when gsi skips deinitializers") also targets #2988 and allocates **`GS0517`** for the same diagnostic. Two open PRs closing the same issue with two different IDs for the same message will collide in whichever order they land, and this is invisible to git and CI because the files differ. One of them should be withdrawn or rescoped before either merges.

---

### Note for the fleet, not a defect

`src/Repl/Program.cs:60` is the line behind the standing rule that **`gsi` never prints a source location**. This PR replaces that loop with `Console.Error.WriteDiagnostics(...)` and threads the filename through `SessionEngine.Evaluate(text, fileName)`, so `gsi` output now carries `file(line,col)`. The change is declared in the body, is pinned by two mutants above, and is a clear improvement — but that rule stops being true the moment this merges.

**Verdict: NOT MERGEABLE.** Blockers 1–3 are all mechanical. The behaviour itself is correct, honestly diagnosed, and properly pinned.

DavidObando:
## Review: NOT MERGEABLE

The blocker is **not** in this PR's code, which is sound and well-tested. It is that **#3052 implements the same feature with a different diagnostic ID**, and merging both silently doubles the diagnostic. One must be closed first. Details, plus two SHOULD-FIX with measured counterexamples, below.

**Trees named.** Scratch clone, freshly fetched from `https://github.com/DavidObando/gsharp`. `main` = `61b7a08c1`, `#3024` head = **`0863957f0`** (verified myself), merge base `f820484c3` — behind `main`, so **three-dot**. Merged onto `61b7a08c1` with **0 conflicts** (`909fb5a1a`). Everything below is measured on that merged tree, not the branch.

`GSharp.Core.dll` md5, `BUILD_EXIT=0` checked before each hash was believed:
`main` `9aa7656c…` / merged `d5a2938f…` / M1 `8bb652c7…` / M2 `878f962a…` / M3 `9fd7b744…` / M4 `35916ed3…` / restored **`d5a2938f…`** — round-trip closed. Build emits **0** warning/error lines. Full `Interpreter.Tests` on the merged tree: **555 passed / 0 failed**.

### The defect is real and this fixes it

| probe | `main` `61b7a08c1` | merged `909fb5a1a` | `gsc` oracle |
|---|---|---|---|
| `samples/PInvoke.gs` (declare **+ call**) | **`0`, rc=0** ✗ silent | `GS0509`, rc=1 ✓ | `13`, rc=0 |
| `PInvokeLibraryImport.gs` | `0`, rc=0 ✗ | `GS0509`, rc=1 ✓ | `13` |
| `PInvokeLibraryImportStringReturn.gs` | *(empty)*, rc=0 ✗ | `GS0509`, rc=1 ✓ | `round-trip!` |
| `PInvokeMarshalAs.gs` | `0`,`0`, rc=0 ✗ | `GS0509`, rc=1 ✓ | `13`,`5` |
| `PInvokeRefOutIn.gs` | `GS9999` (internal-error bucket) ✗ | `GS0509`, rc=1 ✓ | `True True 0 True` |
| deinit, instantiated | `22`, no diagnostic | `22` + `GS0510` warning, rc=0 | `22` |

Five of six repros go from a fabricated value at exit 0 to a specific, located, actionable diagnostic. `GS9999` correctly stops being used as a refusal. **Severity split is argued, not assumed** — ADR-0152 grounds `GS0510` as a Warning in the fact that the CLR does not guarantee finalizers run, so erroring would be *stricter than the emitted semantics*. That is the #2987-shaped structural argument, not "the fix is hard".

### Mutation — each hunk reverted individually, compound conditions decomposed

| mutant | change | result |
|---|---|---|
| **M1** | delete the P/Invoke scan (`Compilation.cs:314-320`) | **15 of 26 failed** ✓ |
| **M2** | delete the deinit scan (`Compilation.cs:322-329`) | **2 of 26 failed** ✓ |
| **M3** | drop `&& SyntaxTrees.Contains(f.Declaration.SyntaxTree)` from the **P/Invoke** guard | **26/26 PASS — survives** ✗ |
| **M4** | drop `&& SyntaxTrees.Contains(s.Deinitializer.Declaration.SyntaxTree)` from the **deinit** guard | **1 failed**, and it is `DeinitDeclaration_ReportsGS0510OnceAndContinues` ✓ |

M4 is the control that proves the M3 result is real and not a filter artefact: the *same* clause on the *sibling* loop is killed by a test written for it.

**The tests are genuinely strong** and I want that on record, because the vacuous-assertion shape is common here. `Issue2986And2988InterpreterBoundaryTests.cs:32-41` uses `Assert.Single(cell.Diagnostics)` plus exact ID, exact `DiagnosticSeverity`, exact `Location.FileName`, exact `StartLine`, and `Assert.Equal(string.Empty, cell.Output)` to prove execution stopped. `:98-115` pins the rendered `path(20,6,20,18): error GS0509:` string **and** exit codes 1 and 0. Nothing here is `DoesNotContain(IsError)`. The four rewritten P/Invoke test files invert their old assertions honestly — `Assert.Contains("GS0509")` + `DoesNotContain("ran")`, not a weakening.

---

## BLOCKER — #3052 implements the same deinit warning under a different ID; both auto-merge

**#3052 (`fix/2988-gsi-deinitializers`, "Warn when gsi skips deinitializers") adds `GS0517 InterpreterDeinitializerNotSupported` with a near-identical message, touching the same four production files and the same `docs/adr/0068-deinit-destructor-support.md`.** #3024's own test file is named `Issue2986And2988…` and its ADR lists #2988 — so both PRs close #2988.

Measured, not reasoned. `main` + #3024 + #3052:

```
MERGE_EXIT=1
CONFLICT (content): docs/diagnostics.md
CONFLICT (content): src/Repl/Program.cs
Auto-merging  src/Core/CodeAnalysis/Compilation/Compilation.cs        <-- no conflict
Auto-merging  src/Core/CodeAnalysis/DiagnosticBag.Reports.Declarations.cs
Auto-merging  src/Core/CodeAnalysis/DiagnosticDescriptors.cs
```

The **production** files merge silently. On the resulting tree:

```
Compilation.cs:305   interpreterDiagnostics.ReportInterpreterDeinitializerNotSupported(...)   // GS0517
Compilation.cs:340   interpreterDiagnostics.ReportDeinitializerNotRunByInterpreter(...)       // GS0510
DiagnosticDescriptors.cs:382  GS0510  "Deinitializer of class '{0}' will not run under the interpreter; … 'gsc' when finalization is required."
DiagnosticDescriptors.cs:384  GS0517  "Deinitializer of class '{0}' will not run under the interpreter; … 'gsc' to use CLR finalization."
```

**Two warnings for one `deinit`.** The docs/`Program.cs` conflicts would make a human look, so this is semi-self-protecting — but the doubling lives in the files that *don't* conflict, and #3024's own `Assert.Single(cell.Diagnostics)` would then fail on `main` after a conflict-resolution that only fixed the three flagged files.

**Fix:** close one. #3024 is the broader PR (P/Invoke + deinit + ADR-0152) and #3052 is deinit-only, so the cheap resolution is to close #3052 and let #3024 carry #2988 — but that is an allocator decision, and `GS0517` vs `GS0510` needs deciding with it. **No ID or ADR collision otherwise:** `GS0509`/`GS0510` are unused on `main` (`grep` over `src/` and both doc files: zero hits, positive control on `GS0513` returns hits in all three), `#3055` takes `GS0516`, `#3052` takes `GS0517`, and ADR `0152` is unique against `main` (#3018 landed `0153`) and against every open PR — #3052, #3054, #3055 add no ADR.

---

## SHOULD-FIX 1 — a repo sample that works correctly under `gsi` today becomes an error

`samples/PInvokeFunctionPointer.gs` **declares two P/Invokes and never calls either** — the only calls are to the managed `compareInt64`. Measured:

```
main 61b7a08c1 : rc=0  1 / -1 / 0     == samples/PInvokeFunctionPointer.golden  (1 / -1 / 0)
merged         : rc=1  GS0509 at (39,6,39,18)
gsc            : rc=0  1 / -1 / 0
```

**`gsi` produced golden-correct output for this sample on `main`, and #3024 makes it unrunnable.** Reduced probe, same shape:

```
@DllImport("libc", EntryPoint: "strlen", CharSet: CharSet.Ansi)
func NativeStrLen(text string) nint;
Console.WriteLine(11)
```
`main` → `11` rc=0. `gsc` → `11` rc=0. **merged → `GS0509` rc=1.**

I checked the justification before filing this, and ADR-0152 *does* address it — *"The error is required even when the declaration is not called… first-class function references fail later with unrelated conversion errors."* That motivation is real and the PR tests it (`PInvokeFunctionValue_ReportsGS0509InsteadOfEvaluatorFailure`, `let f = NativeStrLen`). **But that argument covers "the symbol is referenced", not "the symbol is declared."** A guard on *referenced anywhere* — call or value — satisfies every case the ADR names and keeps `PInvokeFunctionPointer.gs` running.

Also worth stating plainly: this is not free for **#2939**. #2986 says the six samples are excluded there "until this is resolved". After #3024 they are excluded **permanently** — 6 of 7 P/Invoke samples can never run under `gsi`, including the one that runs correctly now. ADR-0152 records `Deinit.gs` as a permanent gsi exclusion but is silent on these.

No CI goes red either way — I checked: `SampleConformanceTests.cs:15-17` compiles samples "through `gsc.dll`", and no golden-comparing test uses the interpreter. So this is a real capability loss, not a broken build.

## SHOULD-FIX 2 — `Compilation.cs:318` `SyntaxTrees.Contains` on the P/Invoke guard appears to be dead

M3 above: removing it passes 26/26. I tried to discriminate it with a probe and **failed**, and the reason looks structural rather than a gap in my probe. The clause can only be *false* when the function comes from a previous REPL submission via `Previous`, but `SessionEngine.cs:223` chains only on success — `if (!hasError) { previous = compilation; }` — and a cell declaring a P/Invoke **always** errors with `GS0509`, so it is never chained. Every reachable P/Invoke declaration is therefore in the current `SyntaxTrees`.

The sibling deinit clause at `:325` is **not** dead — `GS0510` is a Warning, the cell *is* chained, and M4 proves a test depends on it. That asymmetry is exactly why the P/Invoke copy reads as a reflexive paste.

**Fix:** delete the clause, or if it guards a case I have not found, pin it with a test. Same shape as `#3045:871`, which survived 527/527 and turned out to be genuinely dead.

## NIT — `closes #2986` leaves one row of the issue's own table silently wrong

#2986's table lists six samples. `samples/PInvokeStructMarshalling.gs` is **not a P/Invoke program at all** — it has no `@DllImport`/`@LibraryImport`; it is `@StructLayout(LayoutKind.Explicit)` field overlap (a C union). Measured on the merged tree:

```
merged gsi : rc=0   3 / 4 / 0        <-- no diagnostic, exit 0
golden     :        3 / 4 / 6153737367135073092
```

Still the silent-wrong-value class, untouched. Not #3024's job to fix, but auto-closing #2986 buries it. **Split it into its own issue and reference it**, otherwise the explicit-layout defect loses its only tracker.

---

1. **"#3024 allocates `GS0514` and `GS0515`" — it does not. It allocates `GS0509` and `GS0510`.** Both are already inside the quoted taken/unsafe range `GS0509–GS0517`, so the allocation is consistent, but the two IDs to check were the wrong ones.
2. **The docs gate passes, and the feared asymmetry does not exist.** Both IDs land in **both** files, plain `| GS0509 | Error |` rows with **no backticks** (0 backticked ID rows in either file, on `main` or merged). Counts requested: `main` — **383** docs rows / **381** website rows; merged — **385** / **383**.
3. **The 383-vs-381 gap is not a website deficit.** The *ID sets are identical* — **381 unique on `main`, 383 unique merged, symmetric difference empty in both directions**. `docs/diagnostics.md` simply lists **`GS9998` and `GS9999` twice**. Treating the raw row count as a completeness signal reports two phantom gaps.

DavidObando:
## Review — NOT MERGEABLE

One **BLOCKER**: boundary drawn at *declaration* instead of *use*, so a **shipped sample that produced correct golden output is now refused** — on both `gsi` and bare `gsc`. Everything else measures clean, and the guard's *placement* is provably right (compiler-counted below).

Measured on the **merged** tree `8fe994dd` = `main` `185431c4` + PR (`MERGE_RC=0`), against base = `main` `185431c4`. PR head `153fe8285`, merge base `1ea847ebe`, **1 ahead / 2 behind** (`main` gained #3033 `408e02e48`, #3063 `185431c4`). Solution builds only; Core and **both** driver copies verified same-second fresh before every probe.

---

## 🔴 BLOCKER — previously-correct programs now rejected

`src/Core/CodeAnalysis/Compilation/Compilation.cs:314-320` (merged tree) refuses **any program that declares a P/Invoke**, called or not.

`samples/PInvokeFunctionPointer.gs` declares `native_qsort` and `native_dlsym` and **never calls either** — its top level calls a pure G# function plus `Marshal.*`. Its golden is `1/-1/0`.

| probe | base `185431c4` | merged |
|---|---|---|
| `gsi samples/PInvokeFunctionPointer.gs` | rc=0, `1/-1/0` = **golden** | rc=1, **GS0514 ×2**, refused |
| `gsc samples/PInvokeFunctionPointer.gs` (bare) | rc=0, `1/-1/0` + `Success.` | rc=1, **GS0514 ×2**, refused |
| minimal: declare, never call, print `33` | rc=0, `33` | rc=1, **GS0514** |
| `gsc /out:… samples/PInvokeFunctionPointer.gs` | — | rc=0, 3072 bytes, **0× GS0514** |

The issue itself flagged this sample: *"Note `samples/PInvokeFunctionPointer.gs` currently passes."* It did. It no longer does.

**Why CI is green at 14,210.** The samples gate (`test/Compiler.Tests/LanguageConformance/SampleConformanceTests.cs:16-17`) is *"compiled through `gsc.dll`, executed under `dotnet`"* — the **emit** path, which I verified is untouched. Nothing in the suite runs a sample through the evaluate path. **The regression is invisible to every gate**, including `cs2gs` and `e2e`.

**Second defect, same hunk:** on the `gsc` path the message reads *"…compile this program with `gsc` to call native code"* — advice given to a user who just ran `gsc`. Every remedy dead-ends.

**The fix is already specified on the issue.** Your own design comment on #2986: *"if that is judged too aggressive, keep the pre-run report but have it cover both dispatch routes."* Gate on **use** — a direct call **or** a first-class reference to the symbol — instead of on declaration. That still closes both routes the design worried about (`native_abs(…)` and `let f = native_abs`), still reports before the run with a real `TextLocation`, and leaves `PInvokeFunctionPointer.gs` working. The PR body defends declaration-time refusal on the grounds that *"first-class references can escape before an indirect call"* — but an escaping reference **is** a use, and is catchable at the reference site.

---

## 🟡 SHOULD-FIX — a conjunct no test can distinguish from deleted

`Compilation.cs:318`, `SyntaxTrees.Contains(f.Declaration.SyntaxTree)`. Removing it alone **survives 7,664 tests** (Interpreter 596/596 + Core 7,068/7,068, both clean builds).

That term is what stops a REPL session re-reporting `GS0514` on every later submission once a P/Invoke has been declared in an earlier cell. It is load-bearing and untested. (I could not probe it directly — `gsi` refuses a piped stdin with `gsi requires an interactive terminal` — so the REPL consequence is reasoned from the code, not measured. Disclosed.)

---

## Non-vacuity — per-hunk, product mutants, condition inversion

Every mutant: `BUILD_RC` checked first, `dotnet build GSharp.sln`, md5 + mtime on `out/bin/Release/Core/` **and** `out/bin/Release/Repl/` (the `gsi` driver) confirmed same-second fresh. Each has a **discriminating** probe whose clean and mutant outputs differ and were known in advance.

| # | mutant | BUILD_RC | Core.dll md5 | discriminating probe | tests |
|---|---|---|---|---|---|
| — | clean | 0 (0 warn) | `03af6302` | `gsi t1` → GS0514 | 596/0, 7067/0/1 |
| MC | neutralise `if (boundaries.Any())` | 0 (0 warn) | `095696d2` | `gsi t1` → **`0`** — the original silent-default bug returns exactly | **15 failed** ✅ |
| MB | drop `SyntaxTrees.Contains` | 0 (0 warn) | `0041d4e9` | — | **0 failed / 7,664** ❌ SURVIVES |
| MD2 | drop filename threading | 0 (0 warn) | — | location loses filename | **1 failed** ✅ |
| MD1 | revert rich renderer | 0 (0 warn) | — | plain `GS0514: …`, no excerpt | **1 failed** ✅ |
| — | restored | 0 | `03af6302` ✓ | round-trip closed, `porcelain=[]` | |

MC's probe is the strongest evidence in the PR: neutralising the early return reproduces **exactly** the `0` from issue #2986's table.

**NIT:** MD1 and MD2 die by the *same single test* (`BatchFileRunner_RendersPInvokeLocationAndUsesErrorExitCode`). Neither sub-change is independently pinned; one test covers both.

---

## Site denominator — counted by the compiler, not grepped

`Evaluator` and `Compilation` are partial, so I added a **required parameter** to `Evaluator`'s constructor and read `CS7036`, deduped by `file:line`:

> **1 site — `Compilation.cs:336`**, nine lines *after* the guard, in the same method.

So the boundary dominates **1 of 1** evaluator constructions. **There is no "too loose" gap** — nothing reaches evaluation without passing it. The guard's placement is correct; only its *predicate* is too broad. Credit where due.

---

## Four columns

| program | `gsi` base | `gsi` merged | `gsc` bare base | `gsc` bare merged |
|---|---|---|---|---|
| call at top level | `0` (silent bug) | GS0514 | `0` (silent bug) | GS0514 |
| call inside a function | `0` (silent bug) | GS0514 | — | — |
| **declare, never call** | `33` ✅ | GS0514 ❌ | `33` ✅ | GS0514 ❌ |
| **sample, golden `1/-1/0`** | golden ✅ | GS0514 ❌ | golden ✅ | GS0514 ❌ |

The guard is **not inert** in the interpreter — it fires for top-level *and* in-function calls. That trap is avoided. The bottom two rows are the blocker.

---

## #2988, and the sibling PR #3052

**This PR does not touch #2988 and does not claim to.** Body says `Fixes #2986.` only, and states *"#2988 is owned by #3052"*. Verified: **`GS0515` occurs nowhere** in `src/`, `docs/`, `website/`, `test/` — no orphaned row, no orphaned descriptor.

**For the ID allocator: `GS0515` was allocated to this PR and is unused. It is reclaimable.**

**#3024 × #3052 conflict.** `git merge-tree` rc=**1**, content conflicts in **4 files**:
`docs/diagnostics.md`, `src/Core/CodeAnalysis/DiagnosticDescriptors.cs`, `src/Repl/Program.cs`, `website/docs/ref/diagnostics.md`.
`Compilation.cs` and `DiagnosticBag.Reports.Declarations.cs` auto-merge.

They cover **different** issues with **different** IDs, so **neither should be closed as duplicate**. But both independently add the *same* "route `gsi` diagnostics through the shared rich renderer" change to `Program.cs`. Merge one, rebase the other, and resolve `Program.cs` **deliberately** — the obvious hunk-pick compiles, passes CI, and can silently drop one PR's renderer routing.

---

## Gates, merged tree

| gate | result |
|---|---|
| `cs2gs` (CI's own invocation) | rc=0 — **`1 known, 0 new, 0 regressed, 0 stale`**; `G11-Linq-Console` PASS, `G12-Unsafe-Console` PASS |
| Core.Tests (incl. docs gate) | **7067 passed / 0 failed / 1 skipped** |
| Interpreter.Tests | **596 passed / 0 failed** |
| docs rows | both files `\| GS0514 \| Error \|` — severity matches descriptor, not backticked, gate-visible |

`cs2gs` claim confirmed independently.

---

## Disclosures

1. I started the base build while load was **37.68** (above the 30 threshold) because I put `uptime` in the *same* command as the build and saw it only after the build began. Process error. The build completed clean and both copies were consistent, so the measurement stands.
2. `/tmp/g3024/p` already contained probe files I did not create — **another agent is using that path concurrently**. I used uniquely-named probes throughout.
3. The REPL re-reporting consequence in SHOULD-FIX is reasoned from code, not measured — `gsi` requires a TTY.

## Verdict

**NOT MERGEABLE.** The diagnostic is the product here, and it currently fires on programs that never ask for native code. Narrow the predicate from *declared* to *used* — the shape you already specified on #2986 — and this becomes a clean fix. Note that doing so invalidates the mutation table and merged-tree runs above; they will need redoing at the new head.

DavidObando:
**Procedural note from the reviewer: my earlier verdict on this PR is VOID — do not merge on it.**

My comment [#issuecomment-5151741559](https://github.com/DavidObando/gsharp/pull/3024#issuecomment-5151741559) (NOT MERGEABLE, 4 blockers) was measured against `main 4b3ebf24f` + head `3c54a2fec`. The head has moved twice since (`3c54a2fec` → `d1e47bd34` → current), `main` has moved many times, and the PR has changed shape — it is no longer doc-only. Survival/kill results and scope numbers describe a tree that no longer exists.

I am starting a full fresh review now against current `main` and the current head, and will post a new verdict. Please do not treat the old comment as a live blocker list or as a clearance.

DavidObando:
## Re-review verdict: **MERGEABLE**

Supersedes my earlier "verdict void" note. Numbers below are true of **main `5ad0109c`**, **head `6578f1638`**, merge-base **`f1656079`**. Head move `806399a0a` → `6578f1638` was a **pure rebase** (blob SHAs identical across spot-checked files).

### Gates

| check | result |
|---|---|
| `git merge-tree` vs `5ad0109c` | rc=0, clean |
| scope (`--full-index` from computed MB) | 14 files, +277/−93 |
| merged-tree build `-c Release -m:1 -p:ContinuousIntegrationBuild=true` | **BUILD_RC=0, 0 Warning(s), 0 Error(s)** |
| descriptors added | exactly **one**, `GS0514` — the issued ID |
| docs gate (merged tree, bidirectional) | 383 → **384** bare-pipe rows; +1 (`GS0514`), **0 removed**, 0 backtick-invisible |
| Interpreter.Tests (full) | **610/610** |
| Core.Tests P/Invoke slice | **135/135** |
| ADR-0152 filename collision | gone — `0152` here, `0153` already on main from #3018 |

### 2×3 driver axis (the axis a gate does not traverse)

`gsi` / bare `gsc` (EVALUATE) / `gsc /out:` (EMIT), top-level P/Invoke `strlen`, distinct values `11`/`33`:

| shape | gsi | bare gsc (eval) | gsc /out: (emit) |
|---|---|---|---|
| declare + call | rc1 `GS0514` | rc1 `GS0514` | **rc0 `11,13,33`** |
| declare, never call | rc1 `GS0514` | — | — |
| call only inside `func Main()` | rc1 `GS0514` | — | — |

Emit path is untouched (`13` = real `strlen`). `11` never prints under refusal, confirming ADR `:24` "before evaluation". All probes are **top-level**; I did not inherit the author's shape — and the four rewritten test files are top-level too, so no `func Main()` vacuity.

The samples gate is **emit-only** (`SampleConformanceTests.cs:30-58` compiles via `gsc.dll` + runs under `dotnet`); `grep -rn samples .github/workflows/ build/` = 0, **positive-controlled** (`dotnet` matches 4 files in the same population, 3 workflows present). So no gate traverses the evaluate axis — column 2 above is the one CI structurally cannot see, and it is the one this PR changes.

### Mutation — PRODUCT mutant, condition inversion

`Compilation.cs:316` `f => f.IsPInvoke` → `f => !f.IsPInvoke`. Rebuilt full solution (BUILD_RC=0); mtime verified fresh on **all three** `GSharp.Core.dll` copies (`Core/`, `Compiler/`, `Repl/` — 01:25:02), `ref/` excluded.

- probe A flipped rc1 `GS0514` → **rc0 printing `11`, `0`, `33`** — the mutant reproduces issue #2986 verbatim, `0` being the fabricated default.
- tests: **15 failed / 25 examined** — matches the author's claim exactly.
- restored, rebuilt, round-tripped: rc1 `GS0514` back, **25/25 pass**.

### ADR-0152 — all 9 normative sentences traced to code

`:24-26` → `Compilation.cs:314-321` + `DiagnosticDescriptors.cs:382`; location is `Identifier.Location` (`:320`), probe shows `a.gs(5,6,5,18)` = identifier span. `:28` → probe B (rc1, `22` never printed). `:29-30` → pinned by `Issue2986InterpreterBoundaryTests.cs:45-63` (`let f = NativeStrLen`). `:31-33` → `Compilation.Evaluate` is the shared chokepoint; A1/A2 both rc1. `:35-37` → `Program.cs:58-61`; probe output carries filename + location + severity + excerpt. `:41` → guard returns before `Lowering` (`:332`) and `Evaluator` (`:336`). `:42` → proven by the mutant. `:43-44` → probe B. `:45-46` → `Compilation.cs:342-345` `catch (EvaluatorException)`, consistent with `GS9999` being the unexpected-exception bucket.

**9/9 locatable.** No BLOCKERs.

### The −93 lines are honest inversions, not deletions

`[Fact]` counts identical base vs head: 758 = 3/3, 760 = 5/5, 761 = 6/6, 762 = 8/8. The old `Assert.Contains("ran", output)` asserted the program *ran and fabricated a value* — the bug. Replacements are strictly stronger (`Assert.Contains("GS0514")` + `DoesNotContain("ran")`).

The removed `Assert.DoesNotContain("GS0353".."GS0356"/"GS0357".."GS0360")` negative controls **are still pinned**: a spurious binder error returns early at `Compilation.cs:309-311`, so `GS0514` would never be emitted and the new `Assert.Contains("GS0514")` fails for the same reason. Load-bearing `DoesNotContain("GS0326"/"GS0352")` in Issue760 are kept. `RunSubmission` never asserted, so the "helper stopped asserting" shape does not apply.

### NITs (none blocking)

1. **`Compilation.cs:318` `SyntaxTrees.Contains(...)` is unreachable-false in production.** `Compilation.ContinueWith` has exactly **one** production caller (`SessionEngine.cs:213`), and it assigns `previous = compilation` **only when `!hasError`** (`SessionEngine.cs:223-226`) — so an erroring cell-1 P/Invoke declaration is discarded and no later cell can inherit the function. The clause is dead defensive code and is pinned by nothing. Harmless, but worth a comment so a future chained-compilation caller does not assume it is load-bearing. *(I predicted this survival in advance from the call graph rather than reading it as a defect.)*
2. **No multi-submission/chained-`ContinueWith` test** for the boundary; all three new tests are single-submission. Given NIT 1 the hole is closed, so this is coverage hygiene only.
3. **Declaration-time refusal is a real source break** for declare-but-never-call programs — probe B previously printed `22` at rc0, now rc1. Deliberate, justified at ADR `:28-33`, and documented as a consequence at `:43-44`. Flagging so it is a conscious acceptance.
4. **Renderer swap is rider scope** beyond `Fixes #2986` (`Program.cs:58-61`, `SessionEngine.cs:167/212`). It changes stderr format for *every* batch `gsi` diagnostic, not just `GS0514`. It is disclosed in the PR body, documented at ADR `:35-37`, pinned by `Issue2986InterpreterBoundaryTests.cs:66-74`, and arguably required to make "compile with `gsc`" actionable. Backwards-compatible (old `Evaluate(text)` overload retained). Not undeclared scope — it closes no other issue silently.

### Disclosure

I built a detailed hypothesis that a REPL cell-2 call would bypass the guard via the non-cumulative `SyntaxTrees` (`Compilation.cs:57`), falsifying ADR `:42`. **It is wrong** — refuted by `SessionEngine.cs:223-226`. Recording it because the mechanism was plausible and assembled entirely from real code.

PR #3025, opened 2026-08-01
Restore Roslyn fixed-pointer lowering after #2964

### LINKED ISSUE #2922: gsc: fixed containing switch emits ExpectedNumericType IL

DavidObando:
## Summary

PR #2964 merged before its requested rework landed. This follow-up keeps its genuine string-pin fix, removes the `Unsafe.AsPointer<T>` laundering, and restores the Roslyn-compatible `conv.u` lowering for array, string, and span pins.

Fixes #2922

## IL mechanism

The original string path emitted:

```il
ldloc stringValue
conv.i                         // IL_0003: operand is a string reference
call RuntimeHelpers.get_OffsetToStringData
add
conv.u
```

`ilverify` reported `ExpectedNumericType` at `conv.i` because the stack held a `string` reference. The corrected sequence is:

```il
ldloc pinnedString
call instance char& modreq(InAttribute) String::GetPinnableReference()
conv.u
stloc pointer
```

The emitted MemberRef retains `modreq(System.Runtime.InteropServices.InAttribute)`. Null strings branch around the call and store a null pointer.

Array and span paths now again match .NET 10 Roslyn:

```il
ldloc pinnedArray
ldc.i4.0
ldelema T
conv.u
stloc pointer
```

```il
call instance T& GetPinnableReference()
stloc pinnedByRef
ldloc pinnedByRef
conv.u
stloc pointer
```

ECMA-335 III.1.6 permits `conv.u` on `&`. This unsafe IL is unverifiable; `ilverify` reports `ExpectedNumericType` for byte-identical `csc` output. `Unsafe.AsPointer<T>` only moved the conversion into CoreLib and unmasked `UnmanagedPointer`/`StackByRef` errors at pointer dereferences.

## Verification classification

`Issue2906ExhaustiveSwitchReturnEmitTests` returns from one suppressed code to exactly one `ExpectedNumericType`; suppressions decrease from two (`UnmanagedPointer`, `StackByRef`) to one. The suppression cannot be removed because Roslyn-identical unsafe `conv.u` still triggers it. The test-local unsafe bundle documents this limitation, and every suppression is method-scoped; `IlVerifier.cs` is untouched.

## Tests

Green CI merge-tree base: `26b55f562057`.

| Suite | Base | Head |
|---|---:|---:|
| Core.Tests | 7033 passed, 1 skipped | 7033 passed, 1 skipped |
| Compiler.Tests | 3962 passed | 3962 passed |
| Interpreter.Tests | 502 passed | 502 passed |

Issue tests use both `ilverify` and real `Assembly.Load`/`GetTypes()` plus bounded child-process execution. Runtime outputs cover array dereference (`30`), user-struct dereference (`77`), span dereference (`5`), null string (`7`), and string dereference (`90`). Control-flow coverage includes all/some-arm returns, break, nested switch, loops, multiple pins, array/string/struct-field sources, equivalent `if` forms, and void forms.

Pins against the merge-tree base (fail on base, pass on head):

- array pointer dereference
- user-defined struct pointee dereference
- span pointer dereference
- string `GetPinnableReference`/`modreq` plus dereference

Guards (pass on base and head):

- fixed control-flow matrix
- runtime-null string pin
- IL instruction-reader operand decoding

Mutation probes, each rebuilt after restore:

- old `OffsetToStringData` lowering: string metadata pin failed
- string null branch changed from `brfalse` to `pop`: null-string execution failed with `NullReferenceException`
- array `Unsafe.AsPointer<T>`: failed with `UnmanagedPointer` + `StackByRef`
- span `Unsafe.AsPointer<T>`: failed with `UnmanagedPointer` + `StackByRef`
- dropped symbolic pointee map (`AsPointer<object>`): struct test failed with `StackUnexpected` plus pointer errors

Release solution build: 0 warnings, 0 errors.

## Non-regression

- 148-sample corpus: 0 diagnostic/exit/output/SHA-256 differences; 142 binaries per side.
- Samples using `fixed`: **0/148**.
- `gsi` still rejects `fixed` with its explicit GS9999 unsupported-path diagnostic, tracked by #2956; compiled execution is verified here.
- Runtime-null string behavior is correct in return and assignment forms and is not #2953.
- #2953 behavior and #2926 fixed-return epilogue ordering are unchanged.

## Delta

- `MethodBodyEmitter.Statements.cs`: +7/-22 production lines; deletes the reflection lookup/helper and restores direct `conv.u`.
- ADR-0125: +10/-11, recording actual Roslyn output and verifier classification.
- Tests: pointer-dereference, struct-pointee, null-string, metadata, control-flow, and classification coverage.

### DIFF
--- a/docs/adr/0125-fixed-pinned-statement.md
+++ b/docs/adr/0125-fixed-pinned-statement.md
@@ -112,8 +112,7 @@ brfalse NULL
 ldloc pinned; ldlen; conv.i4; brtrue NOTEMPTY
 NULL:     ldc.i4.0; conv.u; stloc ptr       // empty/null ⇒ null pointer
           br AFTER
-NOTEMPTY: ldloc pinned; ldc.i4.0; ldelema <elem>
-          call void* Unsafe.AsPointer<elem>(ref elem); stloc ptr
+NOTEMPTY: ldloc pinned; ldc.i4.0; ldelema <elem>; conv.u; stloc ptr
 AFTER:    <body>
           leave DONE
 FINALLY:  ldnull; stloc pinned              // release on every exit
@@ -126,7 +125,7 @@ DONE:
 ```
 EmitExpr(source); dup; stloc pinned; brfalse NULL
 ldloc pinned; call char& string.GetPinnableReference()
-call void* Unsafe.AsPointer<char>(ref char); stloc ptr
+conv.u; stloc ptr
 br AFTER
 NULL: ldc.i4.0; conv.u; stloc ptr
 AFTER:
@@ -136,9 +135,7 @@ try { <body> } finally {
 ```
 
 `string.GetPinnableReference()` returns `ref readonly char`; the MemberRef
-retains its `modreq(InAttribute)` return signature. `Unsafe.AsPointer<T>`
-converts that managed pointer to the numeric native pointer stored in the
-user-visible `*T` local.
+retains its `modreq(InAttribute)` return signature.
 
 **Span-like / `GetPinnableReference`** (`T& pinned`, issue #1043):
 
@@ -146,7 +143,7 @@ user-visible `*T` local.
 EmitExpr(span); stloc src                    // spill the source for addressing
 ldloca src; call instance T& GetPinnableReference()
 stloc pinned                                 // T& pinned = ref
-ldloc pinned; call void* Unsafe.AsPointer<T>(ref T); stloc ptr
+ldloc pinned; conv.u; stloc ptr
 try { <body> } finally {
     ldc.i4.0; conv.u; stloc pinned            // release on every exit
 }
@@ -165,10 +162,12 @@ return-signature encoder reproduces required custom modifiers on by-ref returns
 (the same mechanism used for `ref readonly` indexers in ADR-0056), so the call
 binds at runtime instead of throwing `MissingMethodException`.
 
-The `Unsafe.AsPointer<T>` call is deliberate for every managed-pointer source.
-Emitting `conv.u` directly after `ldelema` or a ref-return leaves a managed
-pointer at an instruction that requires a numeric stack value, producing
-ilverify's `ExpectedNumericType` error even though RyuJIT accepts the method.
+ECMA-335 III.1.6 permits `conv.u` on a managed pointer, and current .NET 10
+`csc` emits the same sequence for array and span pins. The resulting unsafe IL
+is unverifiable: `ilverify` reports `ExpectedNumericType` for the byte-identical
+array and span sequence from both compilers. Emit tests classify that diagnostic
+as a known unsafe-code limitation rather than moving the conversion into a
+framework helper.
 
 ### 5. New node kinds, exhaustiveness, coverage matrix
 
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Statements.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Statements.cs
@@ -15,7 +15,6 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
-using System.Linq;
 using System.Reflection;
 using System.Reflection.Metadata;
 using System.Reflection.Metadata.Ecma335;
@@ -97,7 +96,7 @@ internal sealed partial class MethodBodyEmitter
 
     // Array-pin form: pin the array reference (`T[] pinned`) and derive
     // `&a[0]` via `ldelema`, guarding the null / zero-length array (→ null
-    // pointer).
+    // pointer), exactly as Roslyn emits for fixed array pins in .NET 10.
     private void EmitFixedArrayPin(BoundFixedStatement node, int pinnedSlot, int pointerSlot)
     {
         var elementType = ((Symbols.PointerTypeSymbol)node.PointerVariable.Type).PointeeType;
@@ -128,7 +127,7 @@ internal sealed partial class MethodBodyEmitter
         this.il.LoadConstantI4(0);
         this.il.OpCode(ILOpCode.Ldelema);
         this.il.Token(this.outer.memberRefs.GetElementTypeToken(elementType));
-        this.EmitManagedPointerAsUnmanagedPointer(elementType);
+        this.il.OpCode(ILOpCode.Conv_u);
         this.il.StoreLocal(pointerSlot);
 
         this.il.MarkLabel(afterLabel);
@@ -155,7 +154,7 @@ internal sealed partial class MethodBodyEmitter
             types: Type.EmptyTypes,
             modifiers: null);
         this.il.Call(this.outer.memberRefs.GetMethodReference(getPinnableReference));
-        this.EmitManagedPointerAsUnmanagedPointer(TypeSymbol.Char);
+        this.il.OpCode(ILOpCode.Conv_u);
         this.il.StoreLocal(pointerSlot);
         this.il.Branch(ILOpCode.Br, afterLabel);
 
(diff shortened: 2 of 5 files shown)


DavidObando:
## Review: NOT MERGEABLE — 2 blockers

Measured against `main` = `4b3ebf24f`. Scope read as `git diff $(git merge-base 4b3ebf24f 65a6425aa) 65a6425aa` → merge-base `a1237ded7141`, **5 files / +268 / -115**, matching the PR file list exactly (no phantom scope).

Build at head: `dotnet build test/Compiler.Tests -c Release -p:ContinuousIntegrationBuild=true` → **rc=0, 0 Warning(s), 0 Error(s)**.
Tests at head: `Issue2922|Issue2906|Issue1033|Issue2900` → **37 passed / 0 failed**.
Mutation: all three production hunks are observable. `Conv_u`→`Conv_i4` at `MethodBodyEmitter.Statements.cs:130` killed 2 tests, `:157` killed 1, `:209` killed 1; each restored and re-verified green (6/6).

The production change itself is clean and minimal: the helper `EmitManagedPointerAsUnmanagedPointer` had exactly 3 call sites, all in this file, all removed; `System.Linq` drop is safe (0 remaining LINQ calls in the file). No diagnostic IDs allocated. No `docs/diagnostics.md` obligation.

---

### BLOCKER 1 — `Fixes #2922` is inverted; this PR *reintroduces* #2922

Issue #2922's Expected is *"Emitted fixed/pointer control flow should satisfy the CLI verifier"*, and it quotes:

```
[IL]: Error [ExpectedNumericType]: ... F(int32, int32[])[offset 0x0000001F][found address of Int32] Expected numeric type on the stack.
```

I compiled the issue's **verbatim repro** with `gsc` from both sides (fresh Release builds, identical flags and ilverify reference set):

| | ilverify rc | result |
|---|---|---|
| `main` `4b3ebf24f` | **0** | `All Classes and Methods ... Verified.` |
| this PR `65a6425aa` | **2** | `[ExpectedNumericType] ... [offset 0x0000001F][found address of Int32]` |

**#2922 is already fixed on `main`.** This PR restores the pre-#2964 `conv.u` lowering, which reproduces the reported error **byte-identically** — same code, same signature, same offset `0x0000001F`, same `found address of Int32`. The PR then adds `"ExpectedNumericType"` to the ilverify ignore lists (`Issue2922FixedSwitchIlTests.cs:298,338,369`, `Issue1033VoidPointerEmitTests.cs:33`), i.e. it suppresses precisely the diagnostic the issue is about.

Merging as-is would regress #2922 **and** auto-close it as fixed.

To be fair to the PR's actual thesis, on *dereference* programs it is a strict improvement, and that part is real:

| program | main errors | this PR |
|---|---|---|
| #2922 repro (fixed+switch, no deref) | **0** | **1** ← regression |
| array deref | 2 (`UnmanagedPointer`,`StackByRef`) | 1 |
| struct deref | 4 (`UnmanagedPointer`,`StackUnexpected`) | 1 |
| span deref | 2 | 1 |
| string deref | 2 | 1 |

Runtime output is **identical on both sides** for every case (`30`, `77`, `5`, `5`, `90`), so nothing user-visible is being fixed here — this is purely a verifiability trade, and the one case it makes worse is the one named in the `Fixes` line.

**Smallest fix:** drop `Fixes #2922` (it is not fixed by this diff, and it is already green on `main`). If the Roslyn-alignment argument stands on its own, land it under its own issue describing the deref-verifiability trade, and state explicitly that #2922's repro regresses from 0 errors to 1.

### BLOCKER 2 — does not merge into `4b3ebf24f`; resolution silently reverts #3014

`git merge-tree --write-tree 4b3ebf24f 65a6425aa` → **rc=1**, 2 content conflicts:

- `test/Compiler.Tests/Emit/Issue2906ExhaustiveSwitchReturnEmitTests.cs`
- `test/Compiler.Tests/Emit/Issue2922FixedSwitchIlTests.cs`

The branch is cut at `a1237ded7141`, before #3014. Those are **the only two production test files on `main` that use `ignoredErrorScope`** (`git grep -c ignoredErrorScope 4b3ebf24f -- test` → `Issue2906…:1`, `Issue2922…:1`, plus `IlVerifier.cs` and its own test). So this conflict lands squarely on 2 of the 2 scoped call-sites #3014 introduced, and the head-side text of both hunks has **no scope argument at all** (`IlVerifier.Verify` at this head is still the pre-#3014 3-parameter signature, `IlVerifier.cs:78`).

Taking the PR side wholesale on those files also drops #3014's `InstructionReader_DoesNotTreatOperandByteAsOpcode` regression test, which lives inside `Issue2922FixedSwitchIlTests.cs` on `main`.

**Smallest fix:** rebase onto `4b3ebf24f` and resolve by hand so that (a) `ignoredErrorScope: @"<Program>\.F$"` survives at both sites, and (b) `InstructionReader_DoesNotTreatOperandByteAsOpcode` survives. Note `main`'s `SlicePin_ConvertsManagedPointerThroughUnsafeAsPointer` (`Issue2922FixedSwitchIlTests.cs:354`) asserts `AsPointer` is emitted and *must* be replaced, since this PR removes that emission — it is the only `AsPointer` assertion in the tree.

### SHOULD-FIX — new blanket, unscoped suppression

`test/Compiler.Tests/Emit/Issue1033VoidPointerEmitTests.cs:33` adds `"ExpectedNumericType"` to `UnsafeIlVerifyIgnored`, consumed unscoped at `:174` (`IlVerifier.Verify(outPath, null, UnsafeIlVerifyIgnored)`). That is assembly-wide category suppression, the exact pattern #3014 was narrowing, and it now hides `ExpectedNumericType` in every method of that assembly. Pass `ignoredErrorScope` here too once rebased.

### NIT

The PR body's "Pins against `a1237ded7141` (fail on base, pass on head)" overstates what is pinned: the four listed pins fail on base on their **IL-shape** assertions (`Assert.DoesNotContain(..., "AsPointer")`, `Assert.Equal(OpCodes.Conv_U, ...)`) and on the swapped suppression list, not on behaviour — runtime output is byte-identical on both sides. Worth restating so the next reviewer does not read them as behavioural regressions caught.

DavidObando:
Closing as option (b): this branch cannot fix #2922 without undoing its entire production change.

Independent direct measurement of the exact #2922 repro, built in the same session:
- `main` `8acc2cc7632819504c2ad33b470237389645e57e`: `gsc rc=0`; direct `ilverify rc=0` (`All Classes and Methods ... Verified`).
- PR head `4cb78234efdd9b60d5d4142557ae36fdadbb6635`: `gsc rc=0`; direct `ilverify rc=2`, one `ExpectedNumericType` at offset `0x1F`—the issue signature.
- Main’s targeted `Issue2922FixedSwitchIlTests`: 7/7 passed.

The branch replaces main’s `Unsafe.AsPointer<T>` lowering with `conv.u` and then suppresses the exact verifier regression. Restoring rc=0 therefore means restoring main, leaving no valid PR delta. Runtime behavior adds no compensating user-visible fix. Issue #2922 remains open.

PR #3026, opened 2026-08-01
Fix symbolic channel element metadata emission

### LINKED ISSUE #2965: gsc: user-defined struct channel values emit invalid IL

DavidObando:
## Summary

- preserve symbolic channel element types in `Channel<T>` signatures, receive locals, method specifications, and generic-member parents
- keep `object` only as the reflection carrier used to discover BCL members
- cover create, send, receive, select, close, nullable/local/generic/nested structs, generic type parameters, async code, closed channels, and blocking select waits
- zero-initialize closed receive results with the symbolic element token

`Channel<T>.Reader` and `.Writer` are declared on the BCL base `Channel<T, T>`; the metadata helper therefore repeats the symbolic element for the declaring type's actual arity.

## Evidence

- Exact latest base `47a400359`: compiler pins **0/2 passed**; both bounded child executions failed with `InvalidProgramException`.
- Head `efae8b442`: compiler pins **2/2 passed** and interpreter guard **1/1 passed**.
- Compiler tests perform `Assembly.Load(File.ReadAllBytes(...))`, `GetTypes()`, three bounded child executions with concurrent pipe draining and exact output, then `IlVerifier.Verify`.
- Interpreter guard passes on both base and head, confirming its boxed `ClrType ?? object` fallback remains valid.
- Primitive and imported-struct controls are byte-identical base/head:
  - `chan int32`: `77d8ea39dc102bc0fd88e0a7d93647f8`
  - `chan DateTime`: `b56f11011411e7f8f93dde28e557b08d`
- Per-production-hunk mutation testing: **16/16 killed, 0 survived**. Every restore passed and every hunk round-tripped `9a29bdf13aa2fa580bf1645a98e6de95 -> distinct mutant hash -> 9a29bdf13aa2fa580bf1645a98e6de95`. The final rebase touched none of the patch files; all eight patch blob hashes matched before/after exactly.
- Release solution build with `ContinuousIntegrationBuild=true`: **0 warnings, 0 errors**.
- Build integrity: **40/40** `GSharp.Core.dll` copies non-zero; **23/23** ordinary implementation copies matched `bce351142acf5d0d6ea29a9546cacdc6`.

The stale `SignatureEncoder` comment was updated: compiler metadata no longer matches the interpreter's erasure, while the interpreter fallback itself remains intentionally unchanged.

Full solution test suites and SDK/e2e tests were not run, per the fleet throttle and the instruction not to touch the shared SDK cache; validation used the narrow issue-specific selectors.

Fixes #2965

### DIFF
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Calls.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Calls.cs
@@ -478,7 +478,8 @@ internal sealed partial class MethodBodyEmitter
 
     private void EmitMakeChannelExpression(BoundMakeChannelExpression node)
     {
-        var elementClr = ResolveChannelElementClrType(node.ChannelType.ElementType);
+        var elementType = node.ChannelType.ElementType;
+        var elementClr = ResolveChannelElementClrType(elementType);
         if (node.Capacity == null)
         {
             var openCreate = typeof(System.Threading.Channels.Channel)
@@ -487,7 +488,7 @@ internal sealed partial class MethodBodyEmitter
                     && m.IsGenericMethodDefinition
                     && m.GetParameters().Length == 0);
             var create = openCreate.MakeGenericMethod(elementClr);
-            this.il.Call(this.outer.memberRefs.GetMethodEntityHandle(create));
+            this.il.Call(this.GetChannelGenericMethodEntityHandle(create, elementType));
             return;
         }
 
@@ -504,7 +505,7 @@ internal sealed partial class MethodBodyEmitter
                 && m.GetParameters().Length == 1
                 && m.GetParameters()[0].ParameterType.IsSameAs(typeof(System.Threading.Channels.BoundedChannelOptions)));
         var bounded = openBounded.MakeGenericMethod(elementClr);
-        this.il.Call(this.outer.memberRefs.GetMethodEntityHandle(bounded));
+        this.il.Call(this.GetChannelGenericMethodEntityHandle(bounded, elementType));
     }
 
     // Issue #2283: entry point used from expression-position emit
@@ -559,23 +560,23 @@ internal sealed partial class MethodBodyEmitter
 
         this.EmitExpression(node.Channel);
         this.il.OpCode(ILOpCode.Callvirt);
-        this.il.Token(this.outer.memberRefs.GetMethodReference(getReader));
+        this.il.Token(this.GetChannelMethodEntityHandle(getReader, chType.ElementType));
 
         this.EmitCancellationTokenNone();
         this.il.OpCode(ILOpCode.Callvirt);
-        this.il.Token(this.outer.memberRefs.GetMethodReference(readAsync));
+        this.il.Token(this.GetChannelMethodEntityHandle(readAsync, chType.ElementType));
 
         this.il.StoreLocal(vtSlot);
         this.il.LoadLocalAddress(vtSlot);
         this.il.OpCode(ILOpCode.Call);
-        this.il.Token(this.outer.memberRefs.GetMethodReference(asTaskGeneric));
+        this.il.Token(this.GetChannelMethodEntityHandle(asTaskGeneric, chType.ElementType));
 
         this.il.OpCode(ILOpCode.Callvirt);
-        this.il.Token(this.outer.memberRefs.GetMethodReference(taskGetAwaiter));
+        this.il.Token(this.GetChannelMethodEntityHandle(taskGetAwaiter, chType.ElementType));
         this.il.StoreLocal(taSlot);
         this.il.LoadLocalAddress(taSlot);
         this.il.OpCode(ILOpCode.Call);
-        this.il.Token(this.outer.memberRefs.GetMethodReference(taskGetResult));
+        this.il.Token(this.GetChannelMethodEntityHandle(taskGetResult, chType.ElementType));
         this.il.StoreLocal(resultSlot);
         this.il.Branch(ILOpCode.Leave, endLabel);
         this.il.MarkLabel(tryEnd);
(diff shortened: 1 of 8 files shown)


DavidObando:
## Review: MERGEABLE — no blocking issues

Measured against `main` = `4b3ebf24f`. Scope read as `git diff $(git merge-base 4b3ebf24f 9023183b2) 9023183b2` → merge-base `f70626452`, **8 files / +440 / -54**, matching the PR file list exactly (no phantom scope).

### 1. Merge and build

- `git merge-tree --write-tree 4b3ebf24f 9023183b2` → **rc=0**, no conflicts.
- Merged worktree (PR head + `4b3ebf24f`): `dotnet build test/Compiler.Tests -c Release -p:ContinuousIntegrationBuild=true` → **rc=0, 0 Warning(s), 0 Error(s)**.
- Merged-state tests, `Issue2965|Issue2922|Issue2906` → **27 passed / 0 failed**, so this composes cleanly with #3014's scoped-suppression work and the fixed-pin suites.

### 2. `Fixes #2965` — verified in both directions

I compiled issue #2965's **verbatim repro** with `gsc` from both sides and executed it:

| | runtime | ilverify |
|---|---|---|
| `main` `4b3ebf24f` | `System.InvalidProgramException: Common Language Runtime detected an invalid program.` | 1× `StackUnexpected` |
| this PR `9023183b2` | **`5`** | **0 errors, verifies clean** |

That is exactly the issue's stated Expected ("print `5`"). Reverse direction: every hunk is channel metadata (`Channel<T>` signatures, member refs, ValueTask/TaskAwaiter/result slots, receive zero-init). No unrelated behaviour, **0 diagnostic IDs allocated** (`git diff $MB HEAD | grep '^+' | grep -oE 'GS[0-9]{4}'` → 0 unique; pattern proven live against the 384 gate-visible rows in `docs/diagnostics.md`), and `docs/` is untouched — correctly, since no descriptor is added, so there is no `docs/diagnostics.md` gate obligation.

### 3. Mutation — the tests genuinely see the change

Perturb → observe flip → restore → observe flip back, rebuilding every time:

| # | mutation | result |
|---|---|---|
| M1 | `EmitZeroInit` symbolic token → always `FromClrType(clrType)` (`MethodBodyEmitter.Conversions.cs:1262`) | **killed 1/2** |
| M2 | `SignatureEncoder.cs:495` symbolic branch disabled (always erase to `object`) | **killed 2/2** |
| M3 | `SlotPlanner.cs:551` `ConstructChannelOperationType` → always CLR carrier | **killed 1/2** |
| M4 | `MethodBodyEmitter.cs:1404` `ResolveChannelContainerSymbol` → always `null` | **killed 2/2** |

Every restore rebuilt and returned **2/2 green**. Interpreter guard `test/Interpreter.Tests/Issue2965ChannelElementSlotTests.cs` → 1/1 passed.

One mutant survived and I want to record why it is **not** a coverage gap. `EmitZeroInit` (`MethodBodyEmitter.Conversions.cs:1257`) drops the old `clrType.IsValueType` branch, so reference elements move from `ldnull; stloc` to `ldloca; initobj`. Making that path a complete no-op for reference types still passes 2/2 — but every emitted body sets `.locals init` (`MethodBodyEmitter.Expressions.cs:806`), so the slot is already zeroed and the mutant is **provably equivalent**. I confirmed by hand rather than assuming: `chan RefBox` and `chan string` closed receives print `True` on `main`, on head, and under the mutant. No test can distinguish it, so none should be added.

Worth noting the reference path is not merely neutral — it improves:

| probe | main | this PR |
|---|---|---|
| `chan RefBox` closed receive | 1× `StackUnexpected` | **verifies clean** |
| `chan string` closed receive | clean | clean |

### 4. Root cause, not symptom

`EmitZeroInit` has exactly **2** call sites (`MethodBodyEmitter.Statements.cs:656`, `MethodBodyEmitter.Calls.cs:586`), both channel receives, so the shared-helper edit has no sibling blast radius. Sweeping the emit layer for surviving erasure, the only remaining `?? typeof(object)` sites are non-channel and already correct: `MethodBodyPlanner.cs:927` (iterator `IEnumerable<T>`, guarded by its own `elementNeedsSymbolic` branch) and `MethodBodyEmitter.MemberAccess.cs:1811` (array `ldelema`, whose token already uses the symbolic type). The 13 surviving `Channels…MakeGenericType(elementClr)` sites are the deliberate reflection carriers that feed `GetChannelMethodEntityHandle`, which re-derives the symbolic container — by design, not missed sites.

Also good: the new tests call `IlVerifier.Verify(assemblyPath)` with **no `ignoredErrorCodes` at all**, and back it with `Assembly.Load` + `GetTypes()` + bounded child-process execution with exit-code and exact-output assertions. Nothing here leans on category-wide suppression.

### NIT (out of scope, `main`-side heads-up)

`select` inside an `async` function still loses the received arm value on **both** `main` and this head — `async func Run() Task[int32]` with `chan int32`, `ch <- 5`, `select { case let value = <-ch { result = value } }` prints **`0`**, not `5`, on both sides. That is the symptom of #2933, which is **CLOSED**. Not introduced by this PR and not its job to fix (the PR's `asyncPair` coverage at `Issue2965ChannelElementSlotTests.cs:86` uses a post-`await` receive expression, which is the part #2965 is about, and that one is correct). Flagging only so it is not lost — it may warrant reopening #2933.

Nothing here blocks. Good, well-evidenced change.

PR #3027, opened 2026-08-01
Fix interpreter imported-base bridging

### LINKED ISSUE #2993: gsi: collection initializer for a get-only property invokes Add without a receiver

### LINKED ISSUE #2996: gsi: members inherited from an imported CLR base class dispatch against the wrong instance

### LINKED ISSUE #3015: gsi: imported-base virtual dispatch loses G# derived type identity

DavidObando:
Fixes #2993
Fixes #2996
Fixes #3015

## Summary
- construct G# class literals through the existing constructor evaluator before applying literal initializers
- allocate lightweight CLR-derived backing types so inherited virtual calls retain the G# runtime type identity
- keep imported-base constructor state and inherited reflection dispatch on one backing object
- correct the stale collection-initializer backend comment

## Validation
- `dotnet build GSharp.sln -p:ContinuousIntegrationBuild=true` (0 warnings, 0 errors)
- focused interpreter regressions: 6/6 passed
- existing CLR-base interpreter tests: 9/9 passed
- production mutations flipped 3/3 construction tests and 2/3 identity tests; restoration returned 6/6 green
- direct `gsi` probes now match compiled outputs for scalar/collection literals, `MemoryStream` inheritance, type identity, and `samples/ImportedBaseClass.gs`

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.cs
@@ -249,8 +249,8 @@ internal sealed partial class ExpressionBinder
     /// is bound into a synthetic local; each element lowers to an
     /// <c>Add(...)</c> call (bare / <c>key: value</c> entries) or an indexer set
     /// (<c>[key] = value</c> entries); the block yields the local. The lowering
-    /// uses only existing bound nodes, so emit and the interpreter both work
-    /// without a new bound-node kind.
+    /// uses only existing bound nodes; each backend must still preserve the
+    /// constructor-before-initializers ordering represented by those nodes.
     /// </summary>
     private BoundExpression BindCollectionInitializerExpression(CollectionInitializerExpressionSyntax syntax)
     {
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -8,6 +8,7 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using System.Reflection;
+using System.Reflection.Emit;
 using System.Threading;
 using System.Threading.Channels;
 using System.Threading.Tasks;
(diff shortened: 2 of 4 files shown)


DavidObando:
Closing due to scope collision: PR #3017 already owns and fixes #2993/#2996. I am dropping those changes and continuing #3015 alone from current main on a separate branch.

Issue #3022, opened 2026-08-01
gsi: ref-aliasing local re-evaluates its initializer, silently reading and writing the wrong storage


DavidObando:
## Summary

A `ref`-aliasing local (`let ref r = <lvalue>`, issue #491 / ADR-0060) is stored in the interpreter as a **re-evaluation thunk over the initializer expression**, not as a storage location. When the initializer's *inputs* change after the alias is declared, the alias silently starts reading and writing **different storage**.

Wrong answer, exit 0, no diagnostic — same class as #3004, but in the already-merged `ref`-local feature rather than in pointers.

## Repro — read direction

```gsharp
package P
import System
func probe() int32 {
    var arr = []int32{10, 20, 30}
    var i int32 = 0
    let ref r = arr[i]
    i = 2
    return r
}
Console.WriteLine(probe())
```

| | output | exit |
|---|---|---|
| compiled (`gsc` → `dotnet`) | `10` | 0 |
| interpreted (`gsi`) | **`30`** | **0** |

`r` aliases `arr[0]`. Changing `i` must not move the alias.

## Repro — write direction

Same program, but `r = 99` after `i = 2`, then print the array:

```gsharp
    let ref r = arr[i]
    i = 2
    r = 99
    return String.Join(",", arr)
```

| | output | exit |
|---|---|---|
| compiled | `99,20,30` | 0 |
| interpreted | **`10,20,99`** | **0** |

The write lands in the wrong element.

## Root cause

`src/Core/CodeAnalysis/Evaluator.cs:309-317` — `RefAlias` stores a `BoundExpression Operand`, nothing else.

- Constructed at `Evaluator.Statements.cs:201`: `new RefAlias(refAddr.Operand)`.
- Reads at `Evaluator.Expressions.cs:330-334` call `EvaluateExpression(alias.Operand)` — re-evaluating the *expression* at read time.
- Writes at `Evaluator.Calls.cs:1288-1292` call `WriteBackToOperand(alias.Operand, value)` — likewise re-evaluating.

For `arr[i]` the operand is a `BoundIndexExpression` whose index sub-expression is re-read on every access, so the alias tracks `i`, not the element `i` designated at declaration time.

The same defect exists for any initializer with a mutable sub-expression — an indexer with a computed index, or a field/property access through a receiver variable that is later reassigned.

## Fix direction

`RefAlias` must capture a **location**, evaluated once at declaration:

- `LocalSlot(frame, VariableSymbol)`
[...]

Issue #2994, opened 2026-08-01
gsi: inline 'out var' at a call site is not declared in the frame


DavidObando:
## Problem

An inline `out var` declaration at a call site is never registered in the interpreter's frame.

`samples/TryParseOutVar.gs`:

```gsharp
tryProduce(out var fromVar)     // line 23
Console.WriteLine(fromVar)      // line 24
```

```
compiled: 42 True 42 42 done
gsi:      42 True
          GS9999: The given key 'var fromVar int32' was not present in the dictionary.
```

The first two lines come from a `TryParse`-style CLR call, which works; the failure is on the **G#-declared** function with an `out` parameter.

## Root cause

`src/Core/CodeAnalysis/Evaluator.Calls.cs:70-82` handles `out`/`ref` arguments by recording `(parameter, addrOf.Operand)` in `userRefSlots` and writing back after the call. The write-back assigns into the frame keyed by the operand's variable symbol — but an inline `out var` declaration never created that entry, so the assignment raises `KeyNotFoundException`.

## Fix sketch

Declare the out-var in the current frame before the call (value irrelevant, it is written on return), so the write-back has a slot. The compiled path already scopes the declaration to the enclosing block.

Found while triaging #2939; excluded there until fixed.

DavidObando:
## Triage: repro confirmed, but the root-cause attribution in the issue body is wrong

Measured with a direct `Compilation.Evaluate` probe (same entry point `gsi` uses) against builds of `8dddbbf2` and `f816c794`. Output is **identical at both**, so PR #2972 neither fixes nor changes this — different failure class.

### Minimal repro (smaller than `samples/TryParseOutVar.gs`)

```gsharp
import System

func tryProduce(out result int32) bool {
    Console.WriteLine("INSIDE")
    result = 42
    return true
}

tryProduce(out var fromVar)
Console.WriteLine(fromVar)
```

Observed:

```
GS9999: The given key 'var fromVar int32' was not present in the dictionary.
```

### The correction: it is the pre-call **seed read**, not the write-back

The issue body attributes this to the write-back in `Evaluator.Calls.cs:70-82`. That is the right block but the wrong statement, and the distinction changes the fix.

**`INSIDE` is never printed.** The throw happens strictly *before* the callee body runs, so the write-back (which runs after `PushFrame`/`EvaluateFunctionBody`, at `Evaluator.Calls.cs:104-112`) is never reached.

The actual throw site is the seed read:

[...]

PR #3028, opened 2026-08-01
Fix gsi pointer dereference boundary

### LINKED ISSUE #3004: gsi: pointer dereference silently returns a stale value — wrong answer with exit 0

DavidObando:
## Summary
- report `GS0513` for unmanaged pointer address/dereference in `gsi`
- preserve managed-byref auto-dereference for CLR properties, indexers, and imported instance calls
- document diagnostic and pin ref/out, indexed-address, and address-only identity regressions

## Design
Loud refusal is intentional. Evaluator has no storage-location model, so returning a value would remain a silent lie.

No positional state is needed: `EvaluateAddressOfExpression` and `EvaluateDereferenceExpression` each have 1 caller of 1 (generic expression dispatch), while 3 of 4 evaluator address-of sites unwrap the operand before dispatch for ref/out/ref-alias handling. Type is correct axis. `BoundAddressOfExpression` constructs `PointerTypeSymbol` exactly when `IsUnmanaged` is true, so both guards use the same predicate by construction; dereference uses the existing `TypeSymbol.IsUnmanagedPointer` helper. Managed-byref auto-dereference remains supported across its 9 binder sites.

`RefAlias` was not reused. It re-evaluates its initializer and would change `&arr[i]`, then `i = 2`, from correct value `10` to phantom `30`; that separate ref-local defect is #3022.

ADR scope statement: **the interpreter has no storage locations**. Scope must not be framed only as `fixed`/`&`/`*`, because ref locals share the limitation.

## Validation
- `Issue3004PointerBoundaryInterpreterTests`: 8/8 passed
- address-guard mutation: 1 failed / 7 passed; new address-only equality case produced no `GS0513`
- restored address guard: 8/8 passed
- `DiagnosticIdUniquenessTests`: 4/4 passed
- issue case 1 through built `gsi`: `42`, `True`
- headline repro through built `gsi`: `GS0513` with compile-with-`gsc` guidance

Full `samples/TryParseOutVar.gs` remains independently blocked by #2994; validation used case 1 only.

Fixes #3004


DavidObando:
## Verdict: NOT MERGEABLE — one BLOCKER (unpinned hunk), everything else holds up

Reviewed the net state at head `d5ead5a5b`. Computed `MB=$(git merge-base ghr/main ghr/pr3028)` = `f70626452` against a **freshly fetched GitHub remote** (not a local-path origin). True scope is **8 files / +185 −5**. The PR is behind `main` (`47a400359`), so I reviewed the **merged** tree: merge is clean (`684f68ed`) and — the part textual mergeability does not tell you — it **compiles**: `dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true` → `RC=0`, 0 Warning(s), 0 Error(s).

### The design call is right, and the trap was avoided

The issue body prescribes a **positional** guard. The PR correctly refused it and used the **type** axis instead. That is the sound choice, and it is stronger than the PR body argues: `IsUnmanaged` and "operand type is a pointer" are not two heuristics, they are the *same* predicate — `BoundAddressOfExpression.cs:42-51` sets `Type = unmanaged ? PointerTypeSymbol.Get(...) : ByRefTypeSymbol.Get(...)`, so `IsUnmanaged ⟺ Type is PointerTypeSymbol` by construction. There is no gap between the two guards.

Managed by-ref is preserved because `ExpressionBinder.Operators.cs:254/257` passes `unmanaged: binderCtx.InUnsafeContext`. The only unconditional `unmanaged: true` site is `MakeFixedBufferPointer` (`ExpressionBinder.Access.MemberLookup.cs:3556`), which is a genuine `*T` fixed-buffer decay — refusing it is correct, not collateral damage.

**`RefAlias` was not reused, and I verified the consequence rather than taking the claim.** `Evaluator.Expressions.cs` is untouched. Compiled truth for the `&arr[i]` / `i=2` / `*p` program is preserved:

| probe | `gsc` (truth) | `gsi` BASE `47a400359` | `gsi` HEAD |
|---|---|---|---|
| #3004 repro | `42` rc=0 | **`7` rc=0** — silent wrong | `GS0513` rc=1 |
| `&arr[i]`, `i=2`, `*p` | **`10`** rc=0 | `10` rc=0 | `GS0513` rc=1 |
| `&a`/`&b` (distinct values) | `False` rc=0 | **`True` rc=0** — silent wrong | `GS0513` rc=1 |

`gsc` still prints **10**, not 30. Base and head were measured on the same harness, exit codes captured as `O=$(cmd); R=$?` (never through a pipe).

### Mutation results — I reverted every changed hunk individually

| hunk | mutant outcome |
|---|---|
| `Compilation/Compilation.cs:334` (`ex.DiagnosticId`/`ex.Severity`) | **DIED** — 3 tests |
| `Evaluator.Calls.cs` deref guard | **DIED** — 1 test |
| **`Evaluator.Calls.cs` address-of guard** | **SURVIVED — 7/7 still pass** |
| `DiagnosticBag.Reports.Expressions.cs:981` reporter | **DIED** — `Every_Report_Uses_A_Descriptor_And_Every_Descriptor_Is_Used` |
| `docs/diagnostics.md:1128` row | **DIED** — `Every_Descriptor_Is_Documented_With_Matching_Severity` |

I want to flag one of my own errors: I first read the new `DiagnosticBag` reporter as dead code, because a repo-wide grep finds exactly one occurrence — its own declaration, zero callers. That reading was **wrong**. `DiagnosticIdUniquenessTests.cs:124-126` asserts the descriptor-name set *equals* the set referenced from `Report*` methods, so the reporter is structurally mandated by the repo's own gate. Removing it reds CI. Retracted.

---

## BLOCKER

**B1 — `src/Core/CodeAnalysis/Evaluator.Calls.cs:1110-1113` — the address-of guard is pinned by no test, and it is the hunk guarding #3004's own headline defect shape.**

Deleting the entire `if (node.IsUnmanaged) { throw ... }` block leaves **7/7 green**. It survives because every `&` case in the test file also dereferences, so the *deref* guard fires first and masks it. The `nil`-pointer case is the only one that isolates deref; nothing isolates address-of.

This is not a redundant hunk. It is the only thing catching a pointer program that **takes an address and never dereferences** — which is still a silent wrong answer with exit 0, exactly the class #3004 was filed about. With the hunk removed:

```gsharp
import System
unsafe {
    var a int32 = 5
    var b int32 = 5
    var pa *int32 = &a
    var pb *int32 = &b
    Console.WriteLine(pa == pb)
}
```

`gsc` → `False`. `gsi` → **`True`, rc=0**. (Note the operands are deliberately *distinct in outcome*: a `p + 1` / `q - p` formulation returns the same value under both the pointer and the identity semantics and would have been a useless discriminator.)

Full round-trip on the diff itself: `Evaluator.Calls.cs` md5 `8086758f` → `4e723237` (`True`, rc=0) → `8086758f` (`GS0513`, rc=1).

**Fix:** add the source above as an `InlineData` case on `UnmanagedPointerOperation_ReportsGs0513`. I confirmed by measurement that such an assertion fails under the mutant and passes restored, so it genuinely pins the hunk. No production change needed — the code is correct, it is only unguarded.

## SHOULD-FIX

**S1 — the PR body's validation section overstates what the suite covers.** It reports "`Issue3004PointerBoundaryInterpreterTests`: 7/7 passed" as evidence for the fix, but 7/7 is also the result with a third of the production change deleted. Worth recording that the address-of arm is currently unpinned so it is not mistaken for covered later.

## NIT

**N1 — `Evaluator.Calls.cs:1121` uses `node.Operand.Type is PointerTypeSymbol` where the repo has a helper.** `TypeSymbol.IsUnmanagedPointer` (`TypeSymbol.cs:292`) is literally `type is PointerTypeSymbol` and is used at four existing sites (`MethodBodyEmitter.MemberAccess.cs:533,644,1778`, `SlotPlanner.cs:1175`). Semantically identical; purely a convention point.

**N2 — some previously-correct interpreted programs now refuse.** Base `gsi` printed `10` for the `&arr[i]` case — the right answer, by luck, because the identity snapshot happened to coincide with truth. That program now errors. This is the correct trade for a loud boundary and the PR does disclose it in a test name, but it is a real behaviour change. Blast radius is nil in-tree: **0 samples contain an `unsafe {` block** (pattern verified against a known-present case first), and the PR's own file is the only interpreter test mentioning pointers.

---

## Gates and declaration

- **`GS0513` is free on `main`** — 0 hits, with the pattern proven live against `GS0508` (16 hits) before trusting the zero.
- **Docs row is gate-visible**, not the backtick trap: `docs/diagnostics.md:1128` is bare-form `| GS0513 | Error | …`. File-wide: **385 bare rows, 0 backticked**. Severity `Error` matches the descriptor. Deleting the row reds the gate (measured).
- **Declaration is clean.** `closingIssuesReferences` = `[3004]` exactly. It mentions #3022 and #2994 without declaring them, which is correct — it fixes neither.
- **The `TryParseOutVar` `GS9999` is pre-existing, not collateral.** I measured it on both trees and the output is byte-identical (`42 | True | GS9999 …`, rc=1). The PR's disclosure that this sample is independently blocked by #2994 is accurate.

## Cross-PR

`Interpreter.Tests` **510/510** on the merged tree; `DiagnosticIdUniquenessTests` 4/4.

#3032 ("Fix interpreter ref alias storage capture", the #3022 fix) edits `Evaluator.Calls.cs` too, so I built the combination rather than trusting disjointness: `main` + #3028 + #3032 merges clean, **compiles** (RC=0, 0/0), all three probes still return `GS0513`, and `Issue3004` is 7/7. No interaction defect.

Unrelated to this PR but visible from here, and neither PR can see it: **#3024 and #3018 both add an ADR numbered `0152`** (`0152-interpreter-unsafe-native-boundary.md` vs `0152-interpreter-compiled-only-storage-boundary.md`), and #3024 allocates `GS0507`. Neither touches `GS0513`, so there is no collision with this PR.

All mutations restored; `git status --porcelain` empty on every worktree.

Issue #1599, opened 2026-07-02
gsc: GS9999 crash and non-resolution on Enum.TryParse[T](..., out var x) (value-type-constrained generic + out var)


DavidObando:
### Summary

`gsc` crashes with an internal error (GS9999) when a generic BCL method with a value-type-constrained type parameter (e.g. `Enum.TryParse<TEnum>`) is invoked with an **inline `out var`** declaration. It also fails to resolve the generic overload at all (GS0159) even with a pre-declared receiver, so `Enum.TryParse` is currently unusable.

### Minimal repro

```gsharp
enum Color { Red, Green }
var ok = Enum.TryParse[Color]("Red", out var result)
Console.WriteLine(result)
```

Output:

```
(1,1,1,7): error GS9999: The given key 'var result GSharp.Core.CodeAnalysis.Binding.OverloadResolution+UserValueTypeConstraintPlaceholder' was not present in the dictionary.
    result
```

### Variant (pre-declared receiver, no inline out-var)

```gsharp
enum Color { Red, Green }
var r Color = Color.Red
var ok = Enum.TryParse[Color]("Red", out r)
```

Output:

```
error GS0159: Cannot find function TryParse.
```

### Analysis

`System.Enum.TryParse<TEnum>(string, out TEnum)` is a generic method whose type parameter carries a `where TEnum : struct` (value-type) constraint. During overload resolution `gsc` represents that constraint with a `UserValueTypeConstraintPlaceholder`. When the call also introduces an inline `out var`, the placeholder type leaks into a dictionary keyed by resolved local/variable symbols, producing an unhandled `KeyNotFoundException` surfaced as GS9999 (an internal-compiler-error, never acceptable). With a pre-declared receiver the placeholder instead causes the candidate to be rejected (GS0159), so the standard BCL `Enum.TryParse` cannot be called.

Both symptoms share one root: overload resolution's handling of a generic method with a value-type (`struct`) constraint combined with an `out` type parameter. `Enum.TryParse` is a standard BCL API and should be callable from G#; at minimum the compiler must never crash with GS9999.

### Expected

- No GS9999 internal crash on the inline `out var` form.
- `Enum.TryParse<TEnum>(string, out TEnum)` (and the `bool ignoreCase` overload) resolves and binds, with `result` typed as the enum.

### Generalization

[...]

PR #3029, opened 2026-08-01
Fix user-defined inline out variables in interpreter

### LINKED ISSUE #2994: gsi: inline 'out var' at a call site is not declared in the frame

DavidObando:
## Summary

- Route user-defined function ref/out arguments through the existing `BuildRefSlots` choke point; all six evaluator call paths now share one argument handler.
- Default only unbound inline `out var`, `out let`, and `out _` slots. Bound lvalues keep their caller value, preserving G# read-before-write behavior.
- Skip ref-slot array construction for user functions without ref-kind parameters.
- Pin user-function `out var`, `out let`, and `out _` independently with distinct expected values (`22`, `33`, `44`).

Six of six ref/out call paths examined use `BuildRefSlots`: five CLR call sites plus the user-function call site. `BoundAddressOfExpression` argument handlers consolidate from two implementations to one.

## Functions touched

Production:
- `Evaluator.EvaluateCallExpression`
- `Evaluator.BuildRefSlots`
- `Evaluator.IsBoundLvalue` (new)

Tests:
- `RefKindInterpreterTests.ClrMethod_InlineOutVar_ProducesValue` (pre-existing CLR guard)
- `RefKindInterpreterTests.UserFunction_InlineOutVar_ProducesValue`
- `RefKindInterpreterTests.UserFunction_InlineOutLet_ProducesValue`
- `RefKindInterpreterTests.UserFunction_OutDiscard_ExecutesCallee`
- `RefKindInterpreterTests.OutParameter_UserFunctionCanReadCallerValueBeforeAssignment`

No changes to pointer dereference handling (`#3004`), `RefAlias` evaluation (`#3022`), sample sources, golden files, or diagnostics.

## CI baseline decision

This was case 1: the PR had changed `samples/TryParseOutVar.gs`, causing the emitted-PE baseline failure. The sample and golden were restored byte-for-byte to `main`; no binary/hash baseline was regenerated. M2 already proves the three user-function tests discriminate the affected paths independently: reverting production causes exactly three failures with expected values `22`, `33`, and `44`.

## Validation

Base: `445d583ec`.

- `Sample_EmittedPE_Matches_Baseline` filtered to `TryParseOutVar`: 1/1 passed.
- `RefKindInterpreterTests`: 9/9 passed.
- `addOne(&value)`: `42`.
- Read-before-write probe: `99`, then final value `5`.
- `samples/TryParseOutVar.gs`: exact golden match `42`, `True`, `42`, `42`, `done`.
- M1 (always-default): read-before-write test fails with actual `1`, expected `42`.
- M2 (always-read/base behavior): all three inline-form tests fail, 3 failed / 0 passed.
- Rebuilt `GSharp.Core.dll` SHA-256: `d93956e7f0a192afc11911058d8fa76804113d398d94bd174199603059441ef6`.

Fixes #2994


DavidObando:
## Review — NOT MERGEABLE at `6cf09d790`

Measured against `main` = `47a400359`. Merge base **`210126dd9`** (4 behind, 1 ahead); merged worktree `dfd9f4610`.

### The shape of this fix is right, and that deserves saying first

This is the root-cause fix, not the symptom fix. It **deletes** the divergent user-function copy at `Evaluator.Calls.cs:69` and routes that path through the existing `BuildRefSlots`. Net **−20 production lines**, and it adds **zero new guard logic** — the `out`→`DefaultValue` guard already existed at `Calls.cs:1134-1143` from #1599.

I verified the population independently rather than accepting the count. `BoundAddressOfExpression` occurrences in the evaluator go from 6 to 5, and *argument-handling implementations* go from **2 to 1**. The body's "six of six" is accurate — `Calls.cs:847`, `:962`, `Expressions.cs:1154`, `:1203`, `:1213` (main line numbers) plus the new `Calls.cs:60`. It also structurally confirms the "no changes to #3004/#3022" claim: `EvaluateAddressOfExpression` (`Calls.cs:1088`) and the ref-alias path (`Statements.cs:201`) are untouched.

Merge clean; merged Release build **rc=0, 0 warnings**; gates pass (40 `GSharp.Core.dll` copies examined / 0 zero-byte; one md5 group; race grep 0 of 76 log lines, mechanism proven live by control). Interpreter.Tests **505/505**. Mutation kill confirmed: full-hunk revert → `Failed: 1, Passed: 5`, and the failure is exactly `UserFunction_InlineOutVar_ProducesValue`.

Also checked cross-PR: #3028 touches `BuildRefSlots` **0** times (its single hunk at `:1104-1120` stops short of the `:1122` definition) and #3032 leaves `WriteBackToOperand`'s definition unchanged, while this PR does not alter the write-back call line at all. No new coupling; a four-cell matrix is not warranted here.

---

## BLOCKER 1 — the fix introduces a new silent `gsi`/`gsc` divergence

`BuildRefSlots` seeds `out` with `DefaultValue(...)` instead of reading the caller's lvalue. That is safe for **CLR** callees, because the CLR contract forbids a method reading an unassigned `out`. It is **not** safe for user-defined G# functions: `GS0238` only requires assignment before *return*, not before first *read*, and `gsc` accepts read-before-write and yields the caller's value.

Five lines, entirely ordinary code:

```gsharp
func addOne(out r int32) {
    r = r + 1
}

var n = 41
addOne(&n)
Console.WriteLine(n)
```

| | value |
|---|---|
| `gsc` (compiled) | **42** |
| `gsi` at `main` | **42** |
| `gsi` at this head | **1** |

Exit code 0 in all three. A second shape (`let seen = result` then `result = 5`, caller holds `99`): `gsc` `99`, `main` `99`, this head **`0`**.

So the PR converts a case where the two backends **agreed** into one where they silently disagree, and the body does not mention it. The consolidation is correct in shape; what needs to become shape-aware is the seed policy — the shared implementation currently carries a CLR-only assumption into the user-function path, where the callee *can* observe the seed.

**I implemented and measured a repair** (+15 lines, Release build rc=0):

```csharp
args[i] = rk == RefKind.Out && !IsBoundLvalue(addrOf.Operand)
    ? DefaultValue(addrOf.Operand.Type)
    : EvaluateExpression(addrOf.Operand);
```

with a helper that returns `true` for any non-variable operand and otherwise checks whether the variable already has a slot in `Locals` (falling back to `TryGetGlobal`). The unbound case is precisely the inline-declaration case #2994 is about.

Results with the repair: inline `out var` `22`, `out let` `33`, `out _` fine, legacy `&x` `11`; the two divergences above become `99` and `42`, matching `gsc`; `samples/TryParseOutVar.gs` still `42 True 42 42 done`; Interpreter.Tests **505/505**.

Note what that last number means: the suite is **505/505 both with and without the repair**. This divergence is invisible to every interpreter test that exists.

## BLOCKER 2 — declared behaviour pinned by no test

The body declares three shapes: "creating `out var` and `out let` bindings; `out _` now completes without a pre-call read." The two new tests cover only `out var` (CLR and user). `out let` and `out _` on a user function are pinned by nothing:

- no test references `samples/TryParseOutVar.gs` — the two grep hits are test-class **name** collisions on `Issue1599EnumTryParseOutVar…`; my pattern is live (a `samples/` control returned 3 files)
- `AspirationalSamplesTests` globs `samples/aspirational/` only, not top-level `samples/`
- the top-level `samples/*.gs` conformance harness runs them through **`gsc`**, which was never broken

I confirmed by hand that both shapes work (`11 / True / 22 / 33 / done`, `gsi` == `gsc`). That is exactly the "verified by hand, pinned by no test" case.

## SHOULD-FIX 1 — one of the two new tests pins nothing

`ClrMethod_InlineOutVar_ProducesValue` **passes on the mutant** (`Failed: 1, Passed: 5`). The CLR path already had the guard from #1599, so this is a guard for pre-existing behaviour, not a pin of this change. Worth keeping — but the "6/6 passed" validation line reads as though both new tests demonstrate the fix, and only one does.

## SHOULD-FIX 2 — every shape produces the same value

`samples/TryParseOutVar.gs` prints `42` for all four `out` shapes, so its output cannot discriminate which shape produced which line — the #2997 failure mode, where identical per-arm values make selection unobservable. My probes use `11/22/33` per shape. Recommend distinct values in the sample and in any new tests.

## NIT 1 — new per-call allocations on the hottest interpreter path

`Evaluator.Calls.cs:59` runs `node.Function.Parameters.Select(p => p.RefKind).ToImmutableArray()` on **every** user-defined function call, whether or not any parameter is by-ref, plus the new `args` array. Previously the loop just indexed `Parameters[i]`. Consider skipping both when no parameter has a `RefKind`.

## NIT 2 — 4 commits behind `main`

Body's validation numbers were measured at the stale base. Merges and builds clean; I re-measured at the merged state (rc=0, 505/505), so nothing is broken — just restate after a rebase.

---

**Verdict: NOT MERGEABLE.** The consolidation is the correct call and I would like to see it land; it needs the `out` seed to distinguish a bound lvalue from an inline declaration, plus tests for the two declared shapes that currently have none.

DavidObando:
## Re-verify — **MERGEABLE**

Both blockers closed. Re-ran everything; accepted nothing.

Merge base computed against freshly-fetched `origin/main`: **`7a6307a79`** (3 behind, 2 ahead). Three-dot scope **4 files, +123/−42**. All measurements on the **merged** tree `46509ad26` (`main` `8acc2cc76` + head `660a09ed0`), build **rc=0**, 0 warnings. Gates with denominators: **40 of 40** `GSharp.Core.dll` copies non-zero-byte; md5 agreement **one group** over the 5 ordinary `out/bin/Release` paths; race grep `MSB302[67]|MSB3883` **0 of 76** log lines, pattern proven live against a control.

### 1. Repair is category-level, not narrower ✅

Landed gate is `rk == RefKind.Out && !IsBoundLvalue(addrOf.Operand)` — the same predicate I built, at the **category**, not the `addOne` shape. It keys on *"is the operand already-bound storage"*, not on callee kind or arity. Correct axis: the rule is that `GS0238` demands assignment before **return**, not before first **read**, so a G# callee may legally read an `out` before writing it and must see the caller's value. NIT below on one sub-term.

Also fixes my round-1 NIT 1: `args` is now allocated only when some parameter has a `RefKind`.

### 2. Both original divergences re-measured ✅

| probe | `gsc` | `gsi` @ merged |
|---|---|---|
| p5 `addOne(&n)`, 42-shape | 42 | **42** |
| p2 99-shape | `99 / 5` | **99 / 5** |
| p1 sample shapes | `11/True/22/33/done` | **11/True/22/33/done** |

### 3–5. Mutation — the actual review

Hash cycle on `GSharp.Core.dll` (never the `gsc` driver — byte-identical across Core changes), rebuilt at every step.

**M1 — always-default** (drop the `!IsBoundLvalue` gate), hash `aac60150`:
p5 → **`1`**, p2 → **`0 | 5`**. Exactly **`OutParameter_UserFunctionCanReadCallerValueBeforeAssignment`** fails. Owner's claim verified: the gate is pinned.

**M2 — revert all production to base**, hash `2ecc29cb`: exactly **3** failures —
`UserFunction_InlineOutVar_ProducesValue` (22), `UserFunction_InlineOutLet_ProducesValue` (33), `UserFunction_OutDiscard_ExecutesCallee` (44).

**Blocker 2 closed.** The three declared shapes are genuinely **distinct paths with distinct values**, not three names over one path — a wrong pick changes the printed number. That is the property #2997 lacked.

**S1 corrected.** `ClrMethod_InlineOutVar_ProducesValue` **passed** under M2, confirming empirically it pins the pre-existing #1599 CLR guard and not this change — and the body now labels it "(pre-existing CLR guard)". Honest.

Restored: hash back to **`26cf2a86`** (== clean), p5 flips back to `42`, `git status` empty.
Cycle: clean `26cf2a86` / M1 `aac60150` / M2 `2ecc29cb` / restored `26cf2a86`.

Suites: baseline **517/517**; focused RefKind **9/9**. Sample golden updated to distinct values `11/22/33` and pinned by `SampleConformanceTests` — note that pins the **compiled** path; the interpreter path is pinned by the 5 unit tests.

### 6. Cross-PR ✅

`#3028` (new head `6e2e57175`), `#3032`, `#3042` each touch `BuildRefSlots` **0** times and `IsBoundLvalue` **0** times; `WriteBackToOperand`'s definition is unchanged. Re-confirmed at current heads after #3028's rebase (force-fetched `+refs/pull/N/head` — a plain fetch silently keeps a stale ref).

I ran the four-cell check on my one concrete hypothesis rather than trusting disjointness: **`main`+#3029+#3042**, built rc=0, local function taking `&n` on an enclosing local → **`42`**, matching `gsc`. No interaction defect.

---

### NIT 1 — `IsBoundLvalue` peeks one frame; `LookupVariable` walks all
`Evaluator.Calls.cs:1145` (this PR's head) uses `Locals.Peek().ContainsKey(v) || TryGetGlobal(v, out _)`, whereas `LookupVariable` (`Evaluator.Expressions.cs:902-918` at `main`) walks **every** frame. Structurally narrower.

I could not make it bite. `BuildRefSlots` runs in the **caller's** context before `PushFrame`, so caller locals are in the top frame by construction; and closures copy captures into the new frame, so the enclosing-local case is also covered. **5 topologies probed, 0 divergences**: top-level global, function-local, global-from-function, lambda-captured, and the #3042 local-function shape that was my specific worry.

Not a blocker — no reachable case. Worth aligning on `LookupVariable` if the frame model ever gains per-block frames, since the failure mode is a **silent wrong value**, not a diagnostic.

### NIT 2 — CI `UNSTABLE` is a pending check, not a failure
`build` has no conclusion yet; the other 5 are SUCCESS. Worth a glance before merge, not a defect.

**Verdict: MERGEABLE.**

PR #3031, opened 2026-08-01
Document lambda lowering cache and issue coverage

DavidObando:
## Summary
- explain why concurrent lazy initialization of `LoweredBody` is safe: `Lower` creates a fresh `Lowerer` per call and `Lowerer` has no static state
- make the #2995 regression discoverable from its reused interpreter test

Follow-up to review feedback on #3012.

## Verification
- rebased onto `main` at `a4a660f32`
- static audit: five mutable `Lowerer` instance fields; private constructor; exactly two `new Lowerer(...)` sites, one in each static `Lower` overload; zero static fields (the same field pattern matched two positive controls elsewhere in `Lowering`)
- targeted filter examined 3 tests: 3 passed
  - `LambdaBody_LowersIfAndAllForForms`
  - `LambdaBody_PreservesConditionForSwitchAndMethodGroupDelegate`
  - `FlatMap_NullableReturningLambdaWithIf_RunsInInterpreter`
- 40 `GSharp.Core.dll` copies checked nonempty; five ordinary copies shared md5 `b436c79a7940687180a13530ca48d5dd`

## Mutation

Both PR hunks are comments, so literal individual hunk reverts correctly killed no tests: the same 3 tests passed after each revert. The behavior described by each hunk was then mutated at the shared production site by returning `node.Body` without lowering:

- build succeeded; five ordinary `GSharp.Core.dll` copies changed from `b436c79a7940687180a13530ca48d5dd` to `00fa00c60cc6e85e46404c9e4d0821b8`
- `LambdaBody_LowersIfAndAllForForms` failed with `Unexpected node IfStatement`
- `FlatMap_NullableReturningLambdaWithIf_RunsInInterpreter` failed
- negative-space `LambdaBody_PreservesConditionForSwitchAndMethodGroupDelegate` passed

After restoring, the build succeeded, all five hashes returned to `b436c79a7940687180a13530ca48d5dd`, and all 3 tests passed again.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/BoundFunctionLiteralExpression.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundFunctionLiteralExpression.cs
@@ -47,8 +47,9 @@ public sealed class BoundFunctionLiteralExpression : BoundExpression
 
     /// <summary>
     /// Gets or sets the evaluator's lazily lowered body. Concurrent first use
-    /// may compute equivalent bodies; <c>Lowerer</c> is stateless, so either
-    /// cached value is safe.
+    /// may compute equivalent bodies; <c>Lower</c> allocates a fresh
+    /// <c>Lowerer</c> per call and holds no static state, so either cached value
+    /// is safe.
     /// </summary>
     internal BoundBlockStatement LoweredBody { get; set; }
 }
--- a/src/Core/CodeAnalysis/Binding/BoundFunctionLiteralExpression.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundFunctionLiteralExpression.cs
@@ -45,5 +45,10 @@ public sealed class BoundFunctionLiteralExpression : BoundExpression
 
     public override BoundNodeKind Kind => BoundNodeKind.FunctionLiteralExpression;
 
+    /// <summary>
+    /// Gets or sets the evaluator's lazily lowered body. Concurrent first use
+    /// may compute equivalent bodies; <c>Lowerer</c> is stateless, so either
+    /// cached value is safe.
+    /// </summary>
     internal BoundBlockStatement LoweredBody { get; set; }
 }
--- a/test/Interpreter.Tests/Issue750ConstraintOverloadInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue750ConstraintOverloadInterpreterTests.cs
@@ -102,6 +102,8 @@ public class Issue750ConstraintOverloadInterpreterTests
     [Fact]
     public void FlatMap_NullableReturningLambdaWithIf_RunsInInterpreter()
     {
+        // Issue #2995: the nullable-returning FlatMap lambda contains the
+        // control flow that previously reached the evaluator unlowered.
         var source = """
             import System
             import Gsharp.Extensions.Optional


DavidObando:
## Review: comment-only PR. Coverage claim **verified true**. Thread-safety claim **false as written**.

Three-dot from computed merge base `26b55f562` → head `c439a4ebe`: **2 files, +7, −0.** No deletions.

### 1. Is there any production change? **No.**

`BoundFunctionLiteralExpression.cs` is in `src/`, but the hunk is a `///` docblock. `Issue750ConstraintOverloadInterpreterTests.cs` gains a two-line `//`. **Zero executable change**, so there is no differential to measure and no docs gate to trip — no `DiagnosticDescriptor`, no `docs/diagnostics.md` row, no ADR number. Nothing to collide on.

### 2. Coverage claim — **TRUE, and precise**

The comment asserts `FlatMap_NullableReturningLambdaWithIf_RunsInInterpreter` covers #2995. I did not take that on trust; I reintroduced the bug.

Mutant at `src/Core/CodeAnalysis/Evaluator.Expressions.cs:782` — let an already-block body reach the evaluator unlowered, which is the #2995 defect:

```csharp
var body = node.LoweredBody ??= node.Body as BoundBlockStatement
    ?? (BoundBlockStatement)Lowering.Lowerer.Lower(node.Body);
```

- `GSharp.Core.dll` md5 `988211a7…` → `428aa7ca…`, build exit 0. Mutant confirmed in the binary before measuring.
- `Issue750ConstraintOverloadInterpreterTests`: **Failed: 1, Passed: 4, Total: 5.**
- The single failure is **`FlatMap_NullableReturningLambdaWithIf_RunsInInterpreter`** — the exact test the comment names, and the only one of the five that notices.

Restored, rebuilt, md5 back to `988211a7…`, `git status --porcelain` empty. **The sentence this PR adds is accurate**, which is the whole point of the PR.

### 3. BLOCKER — the other sentence is false

`src/Core/CodeAnalysis/Binding/BoundFunctionLiteralExpression.cs:48-52`:

> *"Concurrent first use may compute equivalent bodies; `Lowerer` is stateless, so either cached value is safe."*

**`Lowerer` is not stateless.** `src/Core/CodeAnalysis/Lowering/Lowerer.cs:24-40` declares five mutable instance fields:

```
24  private readonly StructSymbol declaringType;
26  private int labelCount;
37  private int tryNestingDepth;
38  private LocalVariableSymbol returnValueLocal;
39  private BoundLabel methodExitLabel;
40  private bool hasRewrittenReturnsInProtectedRegions;
```

The **conclusion** is nevertheless correct, but for a different reason: `Lower` is a static factory that allocates a **fresh instance per call** (`Lowerer.cs:54`, and `:71` for the `declaringType` overload), with no static mutable state. Fresh `labelCount` per call over the same input tree yields structurally identical bodies, so either winner of the `??=` race is fine.

Why this blocks rather than niggles: "stateless" is precisely the property that would license **reusing one `Lowerer` instance** across calls — an obvious-looking allocation saving. Do that and `labelCount`, `methodExitLabel` and `hasRewrittenReturnsInProtectedRegions` leak between lowerings, producing colliding labels and a corrupt exit epilogue. The false sentence licenses the bug; the true one forbids it. No CI job can see a false sentence, and this PR's entire deliverable is the sentence.

**Fix — one clause, no code change:**

> *"Concurrent first use may compute equivalent bodies; `Lower` allocates a fresh `Lowerer` per call and holds no static state, so either cached value is safe."*

### Verification at `main` = `8acc2cc76`

Merge clean, 0 conflicts, tip `9c3269444`; build **exit 0**; **40 `GSharp.Core.dll` examined / 0 zero-byte**. CI green on all 10 non-skipped checks.

### Verdict: **NOT MERGEABLE** — one clause away

- **BLOCKER** — `BoundFunctionLiteralExpression.cs:50`: "`Lowerer` is stateless" is false; five mutable instance fields at `Lowerer.cs:24-40`. Replace the reason with "`Lower` allocates a fresh `Lowerer` per call".
- **NIT** — the body's "targeted interpreter regressions: 3 passed" is not reproducible as stated; the named class runs **5**. Harmless, but it is a count in the section arguing the evidence is adequate.

PR #3032, opened 2026-08-01
Fix interpreter ref alias storage capture

### LINKED ISSUE #3022: gsi: ref-aliasing local re-evaluates its initializer, silently reading and writing the wrong storage

DavidObando:
## Summary
- freeze only mutable ref-local initializer inputs: field receivers plus index targets and indices
- execute bound block-expression prefixes once, then freeze trailing lvalue recursively
- keep `RefAlias`, read/write consumers, `WriteBackToOperand`, and pointer dereference handling unchanged

## Content changes
- `Evaluator.Statements.cs`: freeze captured receiver/index values when ref local is declared; locals, globals, and dereferences retain existing stable-storage paths
- `Evaluator.Expressions.cs`: extract existing block-prefix evaluation so ref binding can execute prefix once without re-running it on reads
- interpreter tests: changed-index read/write, unchanged-index control, class-field receiver reassignment, block-expression capture, global slot, and `GS0513` dereference refusal

## Behavior changes
`let ref` / `var ref` now remain bound to field receiver or array element selected at declaration time. Bound block-expression prefixes run once before trailing lvalue is frozen. Reads and writes continue through existing `RefAlias` consumers and `WriteBackToOperand`.

Unmanaged pointer `&`/`*` behavior is unchanged. `let ref r = *p` with null unmanaged `p` reaches existing #3028 guard and returns `GS0513`, rc=1. No pointer reuse is added.

## Validation
- ref-local interpreter tests: 11 passed, 0 failed
- `gsi`: read `10`; write `99,20,30`; class field `99|99|20`; struct field `99|99`; global `55|55`; block operand `99|10,20,99|40,50,60`
- `gsi` dereference: `GS0513`, rc=1; `gsc`: `NullReferenceException`, rc=134
- `gsc` oracle: 5/5 non-pointer probes match
- mutation kills: field arm 1 failure; index arm 2 failures; block arm 1 failure; injected dereference-bypass arm 1 failure

Fixes #3022

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -594,6 +594,12 @@ public sealed partial class Evaluator
 
     private object EvaluateBlockExpression(BoundBlockExpression node)
     {
+        EvaluateBlockExpressionStatements(node);
+        return EvaluateExpression(node.Expression);
+    }
+
+    private void EvaluateBlockExpressionStatements(BoundBlockExpression node)
+    {
         foreach (var statement in node.Statements)
         {
             if (statement is BoundVariableDeclaration declaration)
@@ -630,8 +636,6 @@ public sealed partial class Evaluator
 
             throw new EvaluatorException($"Unexpected block-expression statement {statement.Kind}", statement);
         }
-
-        return EvaluateExpression(node.Expression);
     }
 
     /// <summary>
--- a/src/Core/CodeAnalysis/Evaluator.Statements.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Statements.cs
@@ -199,27 +199,54 @@ public sealed partial class Evaluator
 
     private void EvaluateVariableDeclaration(BoundVariableDeclaration node)
     {
-        // Issue #491 (ADR-0060 follow-up): a ref-aliasing local binds the symbol
+        // Issue #3022 (ADR-0060 follow-up): a ref-aliasing local binds the symbol
         // to an existing lvalue rather than copying its value. The initializer
-        // is a BoundAddressOfExpression whose operand is the aliased lvalue;
-        // store a RefAlias sentinel in the locals dictionary so subsequent
-        // reads re-evaluate the operand and writes route back to it.
+        // is a BoundAddressOfExpression whose operand is frozen to the storage
+        // location selected when the declaration runs.
         if (node.Variable is LocalVariableSymbol refLocal
             && refLocal.RefKind != RefKind.None
             && node.Initializer is BoundAddressOfExpression refAddr)
         {
-            var alias = new RefAlias(refAddr.Operand);
+            var alias = new RefAlias(FreezeRefOperand(refAddr.Operand));
             var locals = this.Locals.Peek();
             locals[node.Variable] = alias;
 
             // Mirror the read-through semantics for `lastValue` (used by REPL).
-            LastValue = EvaluateExpression(refAddr.Operand);
+            LastValue = EvaluateExpression(alias.Operand);
             return;
         }
 
         var value = EvaluateExpression(node.Initializer);
         LastValue = value;
         Assign(node.Variable, value);
+    }
+
+    private BoundExpression FreezeRefOperand(BoundExpression operand)
+    {
+        switch (operand)
+        {
+            case BoundFieldAccessExpression field when field.Receiver != null && !field.Field.IsConst:
+                return new BoundFieldAccessExpression(
+                    null,
+                    new BoundLiteralExpression(null, EvaluateExpression(field.Receiver), field.Receiver.Type),
+                    field.StructType,
+                    field.Field,
+                    field.NarrowedType);
+
+            case BoundIndexExpression index:
+                return new BoundIndexExpression(
+                    null,
+                    new BoundLiteralExpression(null, EvaluateExpression(index.Target), index.Target.Type),
+                    new BoundLiteralExpression(null, EvaluateExpression(index.Index), index.Index.Type),
+                    index.Type);
+
+            case BoundBlockExpression block:
+                EvaluateBlockExpressionStatements(block);
+                return FreezeRefOperand(block.Expression);
+
+            default:
+                return operand;
+        }
     }
 
     private void EvaluateGoStatement(BoundGoStatement node)
--- a/test/Interpreter.Tests/RefLocalAliasingInterpreterTests.cs
+++ b/test/Interpreter.Tests/RefLocalAliasingInterpreterTests.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.IO;
+using GSharp.Repl.Engine;
 using Xunit;
 
 namespace GSharp.Interpreter.Tests;
@@ -82,6 +83,130 @@ func tweak() {
 tweak()
 ");
         Assert.Contains("7", output);
+    }
+
+    [Fact]
+    public void LetRef_ArrayIndexRead_CapturesOriginalElement()
+    {
+        var output = RunSubmission(@"
+func probe() {
+    var arr = []int32{10, 20, 30}
+    var i int32 = 0
+    let ref r = arr[i]
+    i = 2
+    print(string(r))
+}
+probe()
+");
+        Assert.Equal($"10{Environment.NewLine}", output);
+    }
+
+    [Fact]
+    public void LetRef_ArrayIndexWrite_CapturesOriginalElement()
+    {
+        var output = RunSubmission(@"
+func probe() {
+    var arr = []int32{10, 20, 30}
+    var i int32 = 0
+    let ref r = arr[i]
+    i = 2
+    r = 99
+    print(string(arr[0]) + "","" + string(arr[1]) + "","" + string(arr[2]))
+}
+probe()
+");
+        Assert.Equal($"99,20,30{Environment.NewLine}", output);
+    }
+
+    [Fact]
+    public void LetRef_ArrayIndexWithoutIndexMutation_StillReadsAndWritesElement()
+    {
+        var output = RunSubmission(@"
+func probe() {
+    var arr = []int32{10, 20, 30}
+    var i int32 = 0
+    let ref r = arr[i]
+    r = 99
+    print(string(r) + "","" + string(arr[0]) + "","" + string(arr[1]) + "","" + string(arr[2]))
+}
+probe()
+");
+        Assert.Equal($"99,99,20,30{Environment.NewLine}", output);
+    }
+
+    [Fact]
+    public void LetRef_ClassField_CapturesReceiverBeforeReassignment()
+    {
+        var output = RunSubmission(@"
+class Box {
+    var Value int32
+}
+
+func probe() {
+    var first = Box{Value: 10}
+    var current = first
+    let ref r = current.Value
+    current = Box{Value: 20}
+    r = 99
+    print(string(r) + ""|"" + string(first.Value) + ""|"" + string(current.Value))
+}
+probe()
+");
+        Assert.Equal($"99|99|20{Environment.NewLine}", output);
+    }
+
+    [Fact]
+    public void LetRef_BlockExpression_EvaluatesPrefixOnceAndCapturesElement()
+    {
+        var output = RunSubmission(@"
+func probe() {
+    var original = []int32{10, 20, 30}
+    var current = original
+    let ref r = current[^1]
+    current = []int32{40, 50, 60}
+    r = 99
+    print(string(r) + ""|"" + string(original[0]) + "","" + string(original[1]) + "","" + string(original[2]) + ""|"" + string(current[0]) + "","" + string(current[1]) + "","" + string(current[2]))
+}
+probe()
+");
+        Assert.Equal($"99|10,20,99|40,50,60{Environment.NewLine}", output);
+    }
+
+    [Fact]
+    public void LetRef_GlobalVariable_RemainsAliasedToGlobalSlot()
+    {
+        var output = RunSubmission(@"
+var value int32 = 10
+
+func probe() {
+    let ref r = value
+    r = 55
+    print(string(r) + ""|"" + string(value))
+}
+probe()
+");
+        Assert.StartsWith($"55|55{Environment.NewLine}", output, StringComparison.Ordinal);
+    }
+
+    [Fact]
+    public void LetRef_UnmanagedPointerDereference_ReportsGs0513()
+    {
+        const string Source = """
+            func probe() {
+                unsafe {
+                    var p *int32 = nil
+                    let ref r = *p
+                    print(string(r))
+                }
+            }
+            probe()
+            """;
+
+        var cell = new SessionEngine().Evaluate(Source);
+
+        var diagnostic = Assert.Single(cell.Diagnostics);
+        Assert.True(cell.HasError);
+        Assert.Equal("GS0513", diagnostic.Id);
     }
 
     private static string RunSubmission(string text)


DavidObando:
## Review: NOT MERGEABLE — 1 BLOCKER, 2 SHOULD-FIX

The fix is correct, root-caused in the right place, and fixes **both** directions. It is blocked only on coverage: **6 of the 7 arms of the new `CaptureRefAlias` are pinned by no test**, and one of those six is provably load-bearing.

### Measurement basis
All line numbers are at PR head `774959276` unless stated.

- Scope measured from the merge base, not two-dot: `MB=$(git merge-base 428f31779 774959276)` → `210126dd9`; `git diff $MB 774959276` → **5 files, +164/−25**. Re-merging onto current `main` reproduces the identical diffstat, so no upstream drift.
- `git merge-tree` against `428f31779` → rc=0, clean.
- Build: rc=0, `0 Warning(s)`, `0 Error(s)`. `GSharp.Core.dll` MD5 `8210c39844f48ae91567f931c307226e`.
- Baseline `Interpreter.Tests`: **508 examined, 508 passed**.

### Behaviour confirmed by running (both directions)
`&arr[i]` with `i = 2`, then read and write through the alias:
- read → **`10`** (truth; pre-PR `RefAlias` gave `30`)
- write → **`99,20,30`** (truth; pre-PR gave `10,20,99`)
- rc=0.

Both directions fixed. A read-only fix would have been the same defect in a different hat; this is not that.

### BLOCKER — the field-receiver arm is load-bearing and pinned by nothing

`Evaluator.Calls.cs:1317-1329` (the `BoundFieldAccessExpression` arm) evaluates the *receiver* once and captures it. I reverted just that arm to the naive `CreateRefAlias(field)`:

| build | `GSharp.Core.dll` MD5 | `Interpreter.Tests` |
|---|---|---|
| PR as submitted | `8210c398…` | 508 / 508 pass |
| field arm reverted | `1300759f…` | **508 / 508 pass** |

The mutant is a real rebuild (MD5 moved) and **not one test in 508 can tell it from the shipped code**. It is not dead code, though — probe `let ref v = arr[i].Value` over a `[]Counter` with `i = 2`:

- field-mutant: `30`, then `10,20,99` — **the exact #3022 defect, reproduced for fields**
- restored: `10`, then `99,20,30` — truth; MD5 back to `8210c398…`

So this arm fixes a genuine, currently-broken case and ships untested. Per the rule that caught #3028's address-of guard (survived 7/7), that is a BLOCKER.

Why the existing test misses it: `RefLocalAliasingInterpreterTests.cs:66-81` (`LetRef_AliasStructField_WritesThrough`) uses a **plain local** receiver, which has nothing to re-evaluate. The three new tests at `RefLocalAliasingInterpreterTests.cs:87-134` are **all `arr[i]`**, which is why only the index arm flips (2 of 508 fail when I revert `Evaluator.Calls.cs:1345-1352`).

**Smallest fix:** one test whose receiver is itself a mutated expression — `let ref v = arr[i].Value` on a `[]Counter`, then `i = 2`, assert read `10` and write `99,20,30`.

### SHOULD-FIX — the variable arm's 16 lines are unpinned *and* behaviour-neutral

`Evaluator.Calls.cs:1299-1315` special-cases globals, chained aliases and frame capture. I replaced the whole arm (together with property, dereference, block and `default`) with naive `CreateRefAlias(...)`: MD5 `e1584c93…`, **508 / 508 still pass**. Probes agree too — a global-alias program returns `7` and a chained-alias program returns `9` under **both** the mutant and the restored build.

So these 16 lines currently produce no observable difference anywhere I can reach. Either pin them with a test that actually flips, or delete them.

### SHOULD-FIX — the property arm is unreachable, and so is `default: throw`

`Evaluator.Calls.cs:1331-1343` handles `BoundPropertyAccessExpression`, but the binder never produces one here: `ExpressionBinder.IsLvalue` (`src/Core/CodeAnalysis/Binding/ExpressionBinder.cs:1074-1085`) admits only `BoundBlockExpression`, `BoundVariableExpression`, `BoundFieldAccessExpression`, `BoundIndexExpression`, `BoundDereferenceExpression` — **not** properties. Probe confirms: `let ref` over a `prop` member fails at bind with `GS0256`.

13 lines of dead code. Delete them. Good news attached: because the arms are gated by `IsLvalue`, `default: throw` at `Evaluator.Calls.cs:1361-1362` is **not** a new `GS9999` surface.

(The `GS0256` *message* says "field/property" while its predicate excludes properties — `DiagnosticDescriptors.cs:171`. That is pre-existing and probably what prompted the dead arm. Out of scope here.)

### Declaration path, and the #3028 interaction — re-measured, not inherited

`Evaluator.Statements.cs:191-212` is the **only** caller of `CaptureRefAlias`, gated on `refLocal.RefKind != RefKind.None && node.Initializer is BoundAddressOfExpression`. `RefKind.Ref` is set only on the ref-alias binder path (`StatementBinder.Narrowing.cs:1189`), so `var p *int32 = &x` (`RefKind.None`) does not route here — the PR's claim that it leaves #3004/pointers alone holds. That path is covered by the three new tests.

I did **not** rely on the earlier cross-check; I rebuilt it, because that check used pointer `&`/`*` probes and not `let ref r = *p`. My concern was that the dereference arm at `Evaluator.Calls.cs:1354-1355` recurses into the operand and so **bypasses `EvaluateDereferenceExpression`**, where #3028 installs its `GS0513` guard.

`main` + #3028 (`6e2e57175`) + #3032:
- merge rc=0, **no conflict**, despite both PRs editing `Evaluator.Calls.cs`
- build rc=0, `0 Warning(s)`, `0 Error(s)`, Core MD5 `b050845e…`
- `Interpreter.Tests`: **516 examined, 516 passed**
- `let ref r = *p` inside `unsafe` → **`GS0513`, rc=1**

No hole. #3028's address-of guard fires on `IsUnmanaged` before the deref arm is ever reached, and for managed by-refs the deref guard would not fire anyway. The `arr[i]` and `arr[i].Value` probes both still return truth (`10`, `99,20,30`) on the combined tree. **No interaction defect.**

### Summary

| | |
|---|---|
| BLOCKER | `Evaluator.Calls.cs:1317-1329` — field-receiver arm fixes a real defect (`30`→`10`, `10,20,99`→`99,20,30`) and survives 508/508. Add one `arr[i].Value` test. |
| SHOULD-FIX | `Evaluator.Calls.cs:1299-1315` — 16 lines unpinned and behaviour-neutral in every probe. Pin or delete. |
| SHOULD-FIX | `Evaluator.Calls.cs:1331-1343` — unreachable per `ExpressionBinder.cs:1074-1085`. Delete. |

One test unblocks this.

DavidObando:
## Review: NOT MERGEABLE

**Measured, not reasoned.** Tree: scratch clone at `main` = `fd34166a8`, `#3032` head = `774959276`, merge base `210126dd9` (behind `main`), three-dot. Merges onto `fd34166a8` with 0 conflicts. All builds `-c Release -p:ContinuousIntegrationBuild=true`, `BUILD_EXIT=0` verified before any hash was believed.

`GSharp.Core.dll` md5: `main` `4e3efa2a…` / `main+#3032` `8c756055…` / mutant MA `51079840…` / restored `8c756055…` (round-trip closed).

### The core semantics are correct — say so plainly

`gsc` is the oracle. Base→head differential, 6 probes:

| probe | `main` `fd34166a8` | `main`+#3032 | `gsc` oracle |
|---|---|---|---|
| `let ref r = arr[i]`, read, `i` mutated | `30` ✗ | **`10`** ✓ | `10` |
| `let ref r = arr[i]`, write, `i` mutated | `10,20,99` ✗ | **`99,20,30`** ✓ | `99,20,30` |
| field alias, class receiver **reassigned** | `99\|10\|99` ✗ | **`99\|99\|20`** ✓ | `99\|99\|20` |
| struct-field receiver | `99\|99` ✓ | `99\|99` ✓ | `99\|99` |
| global | `55\|55` ✓ | `55\|55` ✓ | `55\|55` |
| **`let ref r = *p`, `p` null unmanaged ptr** | **`GS0513`, rc=1** ✓ | **rc=0, prints `0`** ✗ | `NullReferenceException`, rc=134 |

Rows 1–3 reproduce independently: the fix is real, and "capture the storage location" is the right rule. **For ADR purposes (#3018 `:60`): ref-local aliasing must alias a captured storage location, not become a boundary. This PR's semantics match `gsc` on every shape I could construct.** Row 6 is why it can't merge as-is.

---

## BLOCKER 1 — the deref arm silently bypasses #3028's `GS0513` boundary

`src/Core/CodeAnalysis/Evaluator.Calls.cs`, `CaptureRefAlias`:

```cs
case BoundDereferenceExpression dereference:
    return CaptureRefAlias(dereference.Operand);
```

This strips the `*` and aliases **the pointer variable itself**, so `EvaluateDereferenceExpression` — where #3028 (`445d583ec`, merged into `main` *after* this PR's base `210126dd9`) put the `GS0513` guard — is never entered.

**Probe** `/tmp/g3032/p/deref3.gs` (null unmanaged pointer, aliased and read):

```
main fd34166a8 : rc=1  GS0513: Unmanaged pointer operations ('&' and '*') are not supported in the interpreter…
main+#3032    : rc=0   0
gsc (oracle)  : rc=134 NullReferenceException
```

`main` correctly refuses. This PR **exits 0 and prints a fabricated `0`** — the silent-wrong-value class, and a straight regression of a guard that landed on `main` last week. This is another four-cell interaction the structural gates cannot see: `merge-tree` is clean and branch CI is green because #3028 was not in this branch's base.

**Recommended fix — narrow, keep the arm** (it is legitimate for managed by-refs):

```cs
case BoundDereferenceExpression dereference:
    if (TypeSymbol.IsUnmanagedPointer(dereference.Operand.Type))
    {
        // fall through to the existing GS0513 path rather than aliasing the pointer variable
        return CreateRefAlias(dereference);
    }
    return CaptureRefAlias(dereference.Operand);
```

Deleting the arm outright also removes the silence (it reaches `default: throw`), but reports GS9999 instead of the accurate GS0513 — worse.

---

## BLOCKER 2 — 4 of the 6 arms are pinned by no test

Mutation **MA**: `CaptureRefAlias` body replaced by `return CreateRefAlias(operand);` (i.e. exactly the old re-evaluation semantics). `BUILD_EXIT=0`, md5 moved to `51079840…`, restored build returned to `8c756055…`.

- `Interpreter.Tests`, filter `RefLocalAliasing|Issue2853EllipsisLoopCapture`: **2 failed / 12 passed of 14.**
- Only `LetRef_ArrayIndexRead_CapturesOriginalElement` and `LetRef_ArrayIndexWrite_CapturesOriginalElement` die.
- `Core.Tests`, filter `RefLocalAliasing`: **15/15 pass** — binder/emit only, correctly unaffected.

So the entire mutation is detected by the **index arm alone**. Unpinned: the **field/property-receiver capture arm**, the **global arm**, the **deref arm**, the **block-expression arm**. The field arm is not theoretical — it is a *measured behaviour change* (`99|10|99` → `99|99|20`, third probe row above) shipping with **zero** coverage. The existing struct-field test at `RefLocalAliasingInterpreterTests.cs:66-82` never reassigns the receiver, so it cannot see it.

The third new test (`…WithoutIndexMutation…`) is a control and survives MA — expected, not a defect.

**Fix:** add a class-receiver-reassignment test (the `99|99|20` probe verbatim) and a `GS0513` deref-rejection test. Distinct values per arm, not `1/1/1`.

---

## SHOULD-FIX — a 31-line version passes your own tests and has neither blocker

Ponytail rung 2 before accepting +115/−25 across 4 files. Locals, globals and derefs already designate **stable storage** — re-evaluating a `BoundVariableExpression` is idempotent, which is why `main` gets rows 4 and 5 right today. Only *receivers and indices* need freezing. That does not require changing `RefAlias`'s shape, so `Evaluator.cs`, the `Evaluator.Expressions.cs` read path and the `Evaluator.Calls.cs` write path all stay untouched.

Implemented and measured, not asserted — full diff at `/tmp/g3032/smaller-alternative.diff`:

```cs
private BoundExpression FreezeRefOperand(BoundExpression operand)
{
    switch (operand)
    {
        case BoundFieldAccessExpression field when field.Receiver != null && !field.Field.IsConst:
            return new BoundFieldAccessExpression(
                null,
                new BoundLiteralExpression(null, EvaluateExpression(field.Receiver), field.Receiver.Type),
                field.StructType, field.Field, field.NarrowedType);

        case BoundIndexExpression index:
            return new BoundIndexExpression(
                null,
                new BoundLiteralExpression(null, EvaluateExpression(index.Target), index.Target.Type),
                new BoundLiteralExpression(null, EvaluateExpression(index.Index), index.Index.Type),
                index.Type);

        default:
            return operand;
    }
}
```

plus two lines in `Evaluator.Statements.cs` (`new RefAlias(FreezeRefOperand(refAddr.Operand))`, and read `alias.Operand`).

| | #3032 | this |
|---|---|---|
| production diff | **+115 / −25, 4 files** | **+31 / −2, 2 files** |
| `RefAlias` shape change, `CreateRefAlias`, `Func`/`Action` indirection | yes | **none** |
| probes 1–5 (`10` / `99,20,30` / `99\|99\|20` / `99\|99` / `55\|55`) | ✓ | **identical, all 5** |
| probe 6 deref | **rc=0 prints `0`** ✗ | **`GS0513` rc=1** ✓ |
| **your own `RefLocalAliasing` tests, incl. the 3 new ones** | 7/7 | **7/7** |
| build | `BUILD_EXIT=0` | `BUILD_EXIT=0`, md5 `a94023c7…` |

BLOCKER 1 cannot exist in this shape: the deref arm is never special-cased, so `EvaluateDereferenceExpression` still runs and #3028's guard still fires.

**Where it is genuinely weaker, stated honestly:** a `BoundBlockExpression` operand re-runs its statements on each read, which #3032 fixes and this does not. If you have a repro for that shape, it justifies part of the extra code — but it is not what #3022 reports, and it is currently untested in either version. If you keep the larger shape, please say what it buys beyond the block case.

---

## NIT — the `BoundPropertyAccessExpression` arm is unreachable

`ExpressionBinder.IsLvalue` (`src/Core/CodeAnalysis/Binding/ExpressionBinder.cs:1074`) admits exactly `BoundVariableExpression | BoundFieldAccessExpression | BoundIndexExpression | BoundDereferenceExpression`, plus `BoundBlockExpression` recursively — **never** `BoundPropertyAccessExpression`. Probe: `let ref r = cur.V` on `prop V int32 { get; set; }` is rejected by **both** `gsi` and `gsc` with `GS0256`. ~14 dead lines. (Corollary: the `default: throw` is safe — the binder gates it.)

---

## Correction to the review brief

> #3032 leaves `WriteBackToOperand` unchanged

Half right, and the half that is wrong matters: **#3032 changes `Evaluator.Calls.cs` by +97 lines**, the largest hunk in the PR. `WriteBackToOperand`'s *body* is unchanged (re-confirmed at head `774959276`: zero changed lines mention it), but its **call site** at ~`:1287` is replaced by `alias.Write(value)`, so the write path no longer goes through it for ref aliases at all. "File untouched" and "function body untouched" are different claims.

DavidObando:
Reworked per review.

- Took smaller freeze-operand shape: `RefAlias`, both consumers, `WriteBackToOperand`, and pointer paths are unchanged.
- Production diff is +39/-8 across `Evaluator.Statements.cs` and `Evaluator.Expressions.cs`; extra 8 lines handle block-expression prefixes once.
- Deleted dead property handling by removing previous switch entirely.
- Added field receiver, block operand, global, and `GS0513` coverage. 11/11 targeted tests pass.
- `let ref r = *p`: `GS0513`, rc=1. `gsc`: `NullReferenceException`, rc=134.
- Mutation kills: field 1, index 2, block 1, injected deref bypass 1.

Head: `8e35e9ae7c874c27d3e488ca3293284afae4567e`.

PR #3033, opened 2026-08-01
Pin constrained property chains and improve GS0333 fallback

DavidObando:
## Summary

Follow up on review items from #2980:

- pin constrained static-property receiver chains in the interpreter;
- pin GS0156 when a nullable constrained-property read is assigned to a non-nullable slot;
- handle reachable non-member suffix syntax such as `T.sizeof(int32)` with GS0333 instead of the evaluator's GS9999 internal error;
- report the source spelling (`sizeof(int32)`) instead of the literal placeholder `?`;
- revise GS0333 and both diagnostic references to describe the actual invalid constrained static access.

This declares an additional user-visible improvement beyond the original review pins: ordinary invalid source now produces an actionable compiler/interpreter diagnostic rather than `GS9999: Unexpected node ErrorExpression`.

## Mutation evidence

- Replacing nested-chain routing with an error expression makes the interpreter pin fail with `Unexpected node ErrorExpression`; restoring it prints `5`.
- Reverting semantic property typing to erased slot typing makes the nullable assignment pin compile unexpectedly with exit 0; restoring it reports GS0156.
- Deleting the final fallback report built successfully and changed `GSharp.Core.dll` from `54aaaa75b31622d448e2985a5c1ba647` to `43d3dfabbab93fdc9e365949d46c85d4`. `NonMemberStaticSuffix_ReportsSourceTextInsteadOfIce` then failed 0/1 in both Compiler.Tests and Interpreter.Tests. Restoring the hunk returned the hash to `54aaaa75b31622d448e2985a5c1ba647` and both tests passed 1/1.

## Verification

Rebased head `dc57266ed6` has merge-base `a4a660f329` with current `origin/main`.

- Core DLL zero-byte gate: 90/90 nonzero
- Propagation gate: 5/5 ordinary Core copies identical, md5 `bdafb827090b03d6e4bda74ed459cd23`
- Compiler Issue2962 tests: 4/4 passed
- Interpreter Issue2962 tests: 2/2 passed
- Core static-virtual/Issue1268 tests: 28/28 passed
- Diagnostic documentation gate: 4/4 passed
- Direct `gsc`: exit 1, actionable GS0333 containing `sizeof(int32)`
- Direct `gsi`: exit 1, same actionable GS0333 with no source-location dependency

No diagnostic ID added. No e2e run.


DavidObando:
## Review: NOT MERGEABLE — the production change is pinned by nothing, and the message it emits is unactionable

Reviewed at head `eb154896` (re-fetched and confirmed unchanged during review), `main` = `8acc2cc76`, **computed merge-base `f70626452`** (three-dot). Net at tip: **3 files, +115** — `ExpressionBinder.Access.MemberLookup.cs` **+2**, plus two new test files.

Build: `dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true`, rc=0. Clean Core md5 `ef07ecfcc7f6f896af7adce2b901ee31` (2,473,984 B), propagating to the `gsi` output directory.

### First, the good news: this PR fixes more than it claims

The title says "review guards", which reads as a tests-only PR. It is not — it adds two production lines, and they are a **real fix**. Measured on the same binary, before and after:

| build | `gsi` rc | output |
|---|---:|---|
| hunk reverted (= `main`) | 1 | `GS9999: Unexpected node ErrorExpression` |
| PR head | 1 | `GS0333: Type parameter 'T' has no constraint that declares a static-virtual member '?' (ADR-0089).` |

`GS9999` is the internal-error bucket, and on `main` it fires for an ordinary user mistake. Converting that into a real diagnostic is worth having. **This is an undeclared fix** — it should be stated in the PR body, because reviewers and the changelog currently cannot see it.

Repro (reaches the new `default:` arm):

```gsharp
sealed interface I[T] { shared { prop V T { get; } } }
struct C : I[string] { shared { prop V string -> "value" } }
func Read[T I[string]](w T) int32 { return T.sizeof(int32) }
```

The `default:` arm is genuinely reachable, not dead: `ParseNameOrCallExpression` (`Parser.Expressions.Creation.cs:586`) can return `typeof` / `sizeof` / `nameof` / `checked` / `make(chan …)` / struct-literal nodes, and the switch at `ExpressionBinder.Access.MemberLookup.cs:3374` handles only `AccessorExpressionSyntax`, `IndexExpressionSyntax`, `CallExpressionSyntax` and `NameExpressionSyntax`.

`gsc` agrees with `gsi` on the ID, so there is no driver divergence.

---

## BLOCKER 1 — the production hunk survives deletion; 32 of 32 tests pass without it

`src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs:3510-3511`

I deleted the two added lines, rebuilt, and re-ran every test that could plausibly observe them:

| suite | filter | result with hunk deleted |
|---|---|---:|
| Interpreter.Tests | `~Issue2962` (this PR's own test) | **1/1 passed** |
| Compiler.Tests | `~Issue2962` (this PR's own tests) | **3/3 passed** |
| Core.Tests | `~StaticVirtual`, `~Issue1268` | **28/28 passed** |
| | **total** | **32/32 survive** |

Mutation was verified to be in the binary, not assumed — Core md5 `ef07ecfc…` → `9c42053b…` → `ef07ecfc…` after restore, with the behaviour flipping `GS0333` → `GS9999` → `GS0333` across the same cycle, and `git status --porcelain` clean at the end.

Why no test can catch it: **3** test files mention `GS0333`, **0** of them assert on message text (all are `d.Id == "GS0333"`), and **0** contain the string `member '?'` — against a live control of **5** files containing `member '`. Every existing `GS0333` assertion is satisfied by the *sibling* guard at `:3490`, which reports the same ID for a genuinely-missing member. That is the #3028 masking shape exactly: an earlier guard fires first, so the new one is never observed.

**Recommended fix:** add one test that compiles a program reaching `default:` (the `sizeof` repro above works) and asserts the diagnostic — asserting the **message**, not just the ID, so it cannot be satisfied by `:3490`.

## BLOCKER 2 — the emitted message is unactionable, and substantively wrong

Same lines. The call passes a literal `"?"` into the `{1}` slot of

> `Type parameter '{0}' has no constraint that declares a static-virtual member '{1}' (ADR-0089).`

Two problems, both user-facing:

1. **`gsi` prints no source location** (`src/Repl/Program.cs:60` is a bare `Console.Error.WriteLine`). So the entire diagnostic a `gsi` user receives is `GS0333: … static-virtual member '?' …` — no location, no member name, nothing to act on. `gsc` at least renders `(12,14,12,27)` and the source line.
2. **The sentence is false.** In the repro the problem is not that a constraint fails to declare a member named `?` — it is that `sizeof(int32)` is not a member name at all. The message asserts a cause that is not the cause.

The sibling arm 15 lines above (`:3491`) passes the real `propName`, so `"?"` is not a codebase idiom — it reads as a placeholder that shipped.

**Recommended fix:** either render the offending syntax (`rightPart.ToString()` / its text span) instead of `"?"`, or report a different, honest diagnostic for "this is not a member name". Please do not allocate a new ID without the coordinator; if a distinct one is wanted, ask.

## NIT — PR body scope

The body presents this as adding review pins. It also changes user-visible diagnostic behaviour (`GS9999` → `GS0333`). Worth one sentence in the body so the change is discoverable.

---

**Verdict: NOT MERGEABLE.** The underlying change is a genuine improvement and I would like to see it land — but a PR whose declared product is *test guards* currently ships a production behaviour change that **32 of 32 targeted tests cannot distinguish from its absence**, and the diagnostic it introduces cannot be acted on by a `gsi` user. Both are small fixes: one test asserting the message, and a real member name in place of `"?"`.

DavidObando:
Addressed both blockers in dc57266ed6. The fallback now renders the exact source suffix (for example, `sizeof(int32)`) and GS0333 describes the invalid constrained static access; the compiler and interpreter no longer expose the literal `?` or evaluator GS9999. Deleting the fallback report made `NonMemberStaticSuffix_ReportsSourceTextInsteadOfIce` fail 0/1 in both Compiler.Tests and Interpreter.Tests; restoring it made both pass 1/1. PR checks are green (9 passed, 3 skipped).

DavidObando:
## Round 2 rubber-duck: **MERGEABLE**. Both round-1 blockers CLOSED by measurement.

Measured tree: `main` `01c1a6c5c` **merged with** `dc57266ed6` -> merged tip `d92131f38` in scratch clone `/tmp/n33`. Worktree `i2962` read only, left clean at `dc57266ed6`. Merge base computed: `a4a660f32`. Three-dot scope = **7 files, +218 -7**.

Fix is in CONTENT, not empty re-push: production blob `6ea436cc1` (round-1 head `eb154896`) -> `08c290426` (new head).

### (a) BLOCKER 1 CLOSED — production hunk now DIES under a PRODUCT mutant

**M1** = revert `ExpressionBinder.Access.MemberLookup.cs:3509-3512` `default:` arm to exactly its pre-PR body (`return new BoundErrorExpression(null);`). Condition-free deletion, so no CS0162. `BUILD_RC=0`, md5 `04979df129…` -> `9b5294efc5…`, mtime fresh. **Live mutant, checked before the hash.**

| hunk | mutant | Compiler.Tests | Interpreter.Tests |
|---|---|---|---|
| `default:` arm diagnostic call (`:3509-3512`) | reverted to pre-PR | **KILL** 1/4 — `NonMemberStaticSuffix_ReportsSourceTextInsteadOfIce` | **KILL** 1/2 — same test |
| GS0333 descriptor text (`DiagnosticDescriptors.cs:247`) | pinned by literal full-message `Assert.Contains` in both new tests | pinned | pinned |

Restore + full-solution rebuild returned md5 **exactly** `04979df129…`. Round trip clean.

The M1 failure text is itself the proof of the ICE claim:

```
String:    "\n(1,1,1,2): error GS9999: Unexpected node"
Not found: "error GS0333: Constrained static access '"
```

So pre-PR really did land in the internal-error bucket. **No longer 32/32.**

Honest note on the other pins: 3 of 4 Compiler tests and 1 of 2 Interpreter tests **survive M1**. They guard pre-existing behaviour, not this hunk. That match the PR title ("Pin ... chains"), so not a defect — but the hunk is carried by exactly **one** test per side. One test each side is thin; a second shape (see (c)) would fix that.

### (b) Message IS actionable — I followed its own advice

Message: `Constrained static access 'sizeof(int32)' on type parameter 'T' must name a static-virtual member declared by an interface constraint (ADR-0089).`

Advice say: name a static-virtual member declared by the constraint. Do exactly that — `return T.V` where `I[T]` declare `shared { prop V T { get; } }`:

- `gsc` rc=**0**, runs, prints **`value`**
- `gsi` (top-level form) prints **`value`**

Remedy do not dead-end. Severity/exit agree: error path `gsc` rc=**1**, `gsi` rc=**1**.

The reword touch a SHARED descriptor with **4** call sites (`:3369`, `:3422`, `:3492` pre-existing, `:3511` new). Probed the common member-not-found case (`T.Missing`): renders `Constrained static access 'Missing' on type parameter 'T' must ...` — still accurate and actionable. **No degradation.** No finding here.

### (c) `default:` arm shapes — 5 of 6 good, 1 never reach the arm

All via `gsc`, exact source suffix, **no empty string, no CLR type name, no truncation**:

| shape | rc | rendered |
|---|---|---|
| `T.sizeof(int32)` | 1 | `'sizeof(int32)'` |
| `T.typeof(int32)` | 1 | `'typeof(int32)'` |
| `T.nameof(w)` | 1 | `'nameof(w)'` |
| `T.checked(1 + 2)` | 1 | `'checked(1 + 2)'` |
| `T.make(chan int32)` | 1 | `'make(chan int32)'` |
| `T.C{}` struct literal | **0** | never reach the arm |

`gsi` parity checked on 3 of them: byte-identical message, same rc.

**SHOULD-FIX 1 — multiline suffix splice raw newlines into the message.** `T.checked(1 + 2 + ... )` broken over 3 source lines yields a **3-line diagnostic** carrying the source indentation verbatim:

```
multi.gs(12,14,14,17): error GS0333: Constrained static access 'checked(1 + 2 + 3 + 4 + 5 + 6 + 7 + 8 +
        9 + 10 + 11 + 12 +
        13 + 14)' on type parameter 'T' must name ...
```

Span `(12,14,14,17)` correct, nothing leaked or truncated — but one-diagnostic-per-line consumers (MSBuild/IDE problem matchers, corpus diffing) will mis-parse it. Cheapest fix: collapse whitespace runs and cap length with an ellipsis before formatting. Small, local to `:3509`.

**SHOULD-FIX 2 (pre-existing, NOT introduced here) — `T.C{}` is silently accepted.** `func Read[T I[string]](w T) C { return T.C{} }` compiles rc=**0** and runs rc=**0**. The `T.` receiver appear to be discarded. This PR only ADDS a diagnostic call, so it cannot have caused acceptance — attribution is pre-existing. But it mean "the fallback now covers non-member suffixes" is not complete: one of the shapes `ParseNameOrCallExpression` (`Parser.Expressions.Creation.cs:586`) can return still bypass the arm. Worth its own issue; **do not block this PR on it**.

### (d) MERGED tree is green — full suites, no filter

`git merge-tree` -> **0** conflict markers. `git merge` rc=0. `dotnet build GSharp.sln -p:ContinuousIntegrationBuild=true` rc=**0**.

| project | passed | failed | skipped | total | time |
|---|---|---|---|---|---|
| Core.Tests | 7065 | **0** | 1 | 7066 | 6m11s |
| Interpreter.Tests | 593 | **0** | 0 | 593 | 30s |
| Compiler.Tests | 4028 | **0** | 0 | 4028 | 1h19m |

**11,687 tests, 0 failures on the merged tree.**

### (e) GS9999 hygiene + docs gate

GS9999 sweep over **13** probe output files: **0** hits. `member '?'` : **0** hits. Both new tests assert `DoesNotContain("GS9999")` and `DoesNotContain("member '?'")`, so round-1's unactionable-fallback complaint is pinned, not just fixed.

Docs gate run on the **merged** tree, regex `^\|\s*(GS\d{4})\s*\|\s*(?:\*\*)?(Error|Warning|Info)`:

- `docs/diagnostics.md`: **383** gate-visible rows, GS0333 present, **0** backticked
- `website/docs/ref/diagnostics.md`: **381** gate-visible rows, GS0333 present, **0** backticked
- Both files updated together; website row text match the descriptor exactly.
- **#3024 trap does not apply**: `git merge-base --is-ancestor eb30d0324 a4a660f32` rc=**0**, so this branch's base already carry #3043's bidirectional gate.
- No row added or deleted, only two row bodies reworded. Gate match ID+severity only, so no orphan risk. Catalogue text read by hand, agrees.
- `website/versioned_docs/version-0.3/` still carry the old wording — correct, that is a frozen snapshot.

### NIT

Message wording for the 3 pre-existing call sites is pinned by **no** text assertion (only the 2 new files assert text; `StaticVirtualInterfaceTests`, `Issue1268…`, `Issue2514…` reference GS0333 by ID). Not introduced here, but the reword just proved that surface is free to drift.

### Disclosure of my own measurement errors

1. I nearly filed a `gsc`/`gsi` divergence: `gsi` printed **nothing** for the fixed program while `gsc` printed `value`. Positive control (`func Main()` printing `CONTROL-77`, `gsi` -> empty) showed **`gsi` does not auto-invoke `Main()`**. Harness artifact, not a defect. Retracted before reporting.
2. The brief expected "GS0333 already exists, so the PR likely adds no diagnostic rows — confirm that". It **does** touch both diagnostics files (2 reworded rows), so the docs gate was in scope and I ran it. Green.
3. Final `GSharp.Core.dll` md5 is `8fb0d7a5…`, not the clean `04979df1…`, because `dotnet test` rebuilds **without** `-p:ContinuousIntegrationBuild=true`. Source tree has **0** tracked modifications; the verified round trip was the CI-flag rebuild, which returned `04979df1…` exactly.

PR #3035, opened 2026-08-01
gsi: convert delegate values to named delegate types

### LINKED ISSUE #2985: gsi: conversion to a user-declared named delegate type throws Unexpected type

DavidObando:
## Summary
- resolve user-declared named delegates through their existing equivalent function type during interpreter conversion
- cover function literals, method groups, function delegates passed to named-delegate parameters, generic delegates, and variadic delegates
- pin nominal identity: structurally identical named delegate types remain non-convertible with `GS0155`

Issue body fix sketch was too large: delegate values are already CLR delegates; only target CLR type resolution was missing.

## Validation
Rebased onto current `origin/main` (`f820484c3`, containing requested `a4a660f32`). Content remains one evaluator hunk plus regression tests.

- `dotnet build GSharp.sln` — 0 warnings/errors
- issue test class — 9/9 pass
- `Delegate|Closure|Lambda|Func` interpreter sweep — 104/104 pass
- reverted production hunk: build passes; 8 conversion tests fail and `GS0155` identity test still passes
- rebuilt `GSharp.Core.dll` MD5: fixed `fc8307af5ba884ea97d03e408a672468`; reverted `bc7efc7d504b26d2b4ee437cd30237cd`
- all three named-delegate samples produce expected values

No file or identity-guard overlap with #3019; imported named delegates take the non-null `ClrType` side and never reach this fallback.

Fixes #2985

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -591,7 +591,8 @@ public sealed partial class Evaluator
             return value;
         }
         else if (value is Delegate sourceDelegate
-            && node.Type?.ClrType is Type delegateType
+            && (node.Type?.ClrType
+                ?? (node.Type as DelegateTypeSymbol)?.EquivalentFunctionType.ClrType) is Type delegateType
             && typeof(Delegate).IsAssignableFrom(delegateType))
         {
             return delegateType.IsInstanceOfType(sourceDelegate)
--- a/test/Interpreter.Tests/Issue2985NamedDelegateConversionTests.cs
+++ b/test/Interpreter.Tests/Issue2985NamedDelegateConversionTests.cs
@@ -1,0 +1,86 @@
+// <copyright file="Issue2985NamedDelegateConversionTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System.Collections.Generic;
+using GSharp.Core.CodeAnalysis.Compilation;
+using GSharp.Core.CodeAnalysis.Symbols;
+using GSharp.Core.CodeAnalysis.Syntax;
+using Xunit;
+
+namespace GSharp.Interpreter.Tests;
+
+/// <summary>
+/// Issue #2985: function literals convert to user-declared named delegates.
+/// </summary>
+public class Issue2985NamedDelegateConversionTests
+{
+    public static TheoryData<string, string, object> Cases => new()
+    {
+        {
+            "void",
+            """
+            type Recorder = delegate func(value int32)
+
+            class Box {
+                var Value int32
+            }
+
+            let box = Box{}
+            var record Recorder = func(value int32) {
+                box.Value = value + 100
+            }
+            record.Invoke(7)
+            box.Value
+            """,
+            107
+        },
+        {
+            "value",
+            """
+            type Transformer = delegate func(value int32) int32
+
+            var transform Transformer = func(value int32) int32 {
+                return value * 2
+            }
+            transform.Invoke(23)
+            """,
+            46
+        },
+        {
+            "generic",
+            """
+            type Mapper[T any] = delegate func(value T) T
+
+            var mapper Mapper[string] = func(value string) string {
+                return value + "-generic"
+            }
+            mapper.Invoke("named")
+            """,
+            "named-generic"
+        },
+        {
+            "variadic",
+            """
+            type Scorer = delegate func(seed int32, values ...int32) int32
+
+            var score Scorer = func(seed int32, values ...int32) int32 {
+                return seed + values.Length * 10
+            }
+            score(2, 4, 5, 6)
+            """,
+            32
+        },
+    };
+
+    [Theory]
+    [MemberData(nameof(Cases))]
+    public void FunctionLiteral_ConvertsToNamedDelegate(string _, string source, object expected)
+    {
+        var result = new Compilation(SyntaxTree.Parse(source))
+            .Evaluate(new Dictionary<VariableSymbol, object>());
+
+        Assert.Empty(result.Diagnostics);
+        Assert.Equal(expected, result.Value);
+    }
+}
--- a/test/Interpreter.Tests/Issue2985NamedDelegateConversionTests.cs
+++ b/test/Interpreter.Tests/Issue2985NamedDelegateConversionTests.cs
@@ -11,7 +11,7 @@ using Xunit;
 namespace GSharp.Interpreter.Tests;
 
 /// <summary>
-/// Issue #2985: function literals convert to user-declared named delegates.
+/// Issue #2985: delegate values convert to user-declared named delegates.
 /// </summary>
 public class Issue2985NamedDelegateConversionTests
 {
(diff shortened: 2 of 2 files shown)


DavidObando:
## Review: MERGEABLE — 0 BLOCKER, 0 SHOULD-FIX, 2 NIT

One expression changed, pinned by 4 tests that flip, root cause not symptom, and the "no overlap with #3019" claim **holds under a controlled probe**. This is the shape a fix should be.

### Measurement basis
Line numbers at PR head `963bde233` unless stated.

- Scope from the merge base, not two-dot: `MB=$(git merge-base 428f31779 963bde233)` → `47a400359`; `git diff $MB 963bde233` → **2 files, +88/−1** (`Evaluator.Calls.cs` +3/−1, `Issue2985NamedDelegateConversionTests.cs` +86).
- `git merge-tree` against `428f31779` → rc=0, clean.
- Build: rc=0, `0 Warning(s)`, `0 Error(s)`. `GSharp.Core.dll` MD5 `55d4e089b33c64869bcfb58eb7294f12`.
- Baseline `Interpreter.Tests`: **509 examined, 509 passed**.
- `Fixes #2985` present. No new diagnostic IDs, no docs touched → no docs-gate exposure.

### The hunk is pinned

Reverted the fallback to plain `node.Type?.ClrType is Type delegateType`:

| build | `GSharp.Core.dll` MD5 | `Interpreter.Tests` |
|---|---|---|
| PR as submitted | `55d4e089…` | 509 / 509 pass |
| fallback reverted | `73160d82…` | **4 fail / 505 pass** |
| restored | `55d4e089…` | 509 / 509 pass |

The 4 failures are exactly the new tests. MD5 moves and returns, so both builds are real. The four theory cases assert **four distinct values** (`107`, `46`, `"named-generic"`, `32`), so this is not the #2997 all-arms-same-value trap.

### The #3019 no-overlap claim — probed with a control, not accepted

The risk you named was structural equivalence making two nominally-distinct named delegate types interchangeable at runtime. Two named delegates with **identical signatures**:

```
type Pred1 = delegate func(x int32) bool
type Pred2 = delegate func(x int32) bool
```

| probe | fallback reverted | PR as submitted |
|---|---|---|
| `var b Pred2 = a` where `a Pred1` | `GS0155: Cannot convert type 'Pred1' to 'Pred2'.` | **`GS0155` — identical** |
| `take(Pred1)` / `take(Pred2)` overloads, called with a `Pred1` | `GS9999: Unexpected type Pred1` | **`one` — correct arm** |

The first row is the control and it is **identical on both builds**: nominal identity is a bind-time property, and this PR does not touch the binder. The second row is the interesting one — this PR does not collapse the two types, it *unblocks* the case that previously died with a `GS9999`. #3035 makes overload dispatch across two identical-signature named delegates observably correct. **No overlap with #3019.**

Mechanically this is why: the `??` only fires when `node.Type?.ClrType` is **null**, which is true only for user-declared named delegates. Imported CLR delegates keep a real `ClrType` and short-circuit. Confirmed by probe — `List[int32].FindAll(func(x int32) bool { return x > 3 })` over `{1,5,9}` returns **`2`**, correct, on both builds.

### `Evaluator.Calls.cs:604` is the sole chokepoint — measured, not grepped

You flagged the recurring pattern of a CLR path guarded and its user-func twin left unguarded. It is not present here. Six positions where a value meets a named-delegate target, every one **`GS9999` before / correct after**:

| position | reverted | PR |
|---|---|---|
| argument — `apply(func(v int32) int32 {...}, 5)` | `GS9999: Unexpected type Transformer` | `15` |
| return — `func make() Transformer { return func...}` | `GS9999` | `15` |
| class field assign — `h.T = func...` | `GS9999` | `20` |
| array literal — `[]Transformer{ func... }` | `GS9999` | `25` |
| method group — `var t Transformer = triple` | not measured | `15` |
| indirect — `let f = func...; var t Transformer = f` | not measured | `35` |

Everything routes through `EvaluateConversionExpression`. There is no second site needing the same treatment, and all six are the same symptom as #2985's title ("throws Unexpected type") — one root cause, one issue, one expression. No undeclared scope in either direction; #2948 is the `gsc`-side named-delegate issue and touches a different path, so nothing extra closes.

Worth noting the pre-PR failure was a **`GS9999` at exit 1** — loud, not the silent wrong-value class. Nothing was being quietly miscomputed here.

### NIT
- All 4 tests use the `var x T = literal` declaration shape. Five other positions are fixed but untested. Not blocking: it is a single expression and reverting it flips 4 tests, so the hunk is fully pinned — but an argument-position case would cost one line.
- `var t Transformer = nil` → `GS0155: Cannot convert type 'nil' to 'Transformer'.` Pre-existing binder gap, unchanged by this PR, out of scope. Flagging only so it is on the record.

### Verdict
**MERGEABLE.** Ship it.

DavidObando:
## Review: MERGEABLE

No blockers. Two SHOULD-FIX on coverage breadth, one scope correction. The 2-line fix is the right size and the right place — details below, all measured.

**Tree:** scratch clone, `main` = `fd34166a8`, `#3035` head = `218b5efad`. Merge base is `fd34166a8` itself — the rebase is clean, nothing to merge. Three-dot = two-dot here. All builds `-c Release -p:ContinuousIntegrationBuild=true`.

`GSharp.Core.dll` md5: `main` `4e3efa2a…` / `#3035` `a23f98bd…` / `main` again `4e3efa2a…` / `#3035` again `a23f98bd…` — **round-trip closed twice**, `BUILD_EXIT=0` verified before any hash was believed. Build produced **0** warning/error lines.

### Base→head differential, `gsc` as oracle

| probe | `main` `fd34166a8` | `#3035` `218b5efad` | `gsc` oracle |
|---|---|---|---|
| 1-param value return (issue repro) | `GS9999: Unexpected type Transformer`, rc=1 | **`46`** | `46` |
| **0-param**, value return | `GS9999: Unexpected type Ticker`, rc=1 | **`77`** | `77` |
| **2-param**, value return | `GS9999: Unexpected type Adder`, rc=1 | **`311`** | `311` |
| **method group** → named delegate | `GS9999: Unexpected type Transformer`, rc=1 | **`33`** | `33` |
| **named delegate as call argument** | `GS9999: Unexpected type Transformer`, rc=1 | **`11`** | `11` |
| **two structurally identical named delegates** (`Alpha` → `Beta`) | `GS0155`, rc=1 | **`GS0155`, rc=1** | `GS0155` |

Distinct values per shape (`46/77/311/33/11`), so a fix that made dispatch uniformly wrong could not hide. Every row matches `gsc`.

### The structural-interchangeability worry is unfounded — measured, not reasoned

The concern is fair on the name: `EquivalentFunctionType` *is* a structural projection, so at runtime every same-signature named delegate is now represented by the same `Func<…>` CLR instance. But **named-delegate identity is enforced at bind time, not in the evaluator**:

```
type Alpha = delegate func(v int32) int32
type Beta  = delegate func(v int32) int32
var a Alpha = func(v int32) int32 { return v * 2 }
var b Beta = a          // GS0155: Cannot convert type 'Alpha' to 'Beta'
```

`GS0155` on **`main` and on this head, in both `gsi` and `gsc`** — four cells, all identical. `EvaluateConversionExpression` is never reached for that program, so an evaluator-only fallback cannot weaken name identity. **No new silent defect is traded for the fixed exception.**

I also probed whether the substituted runtime type leaks through reflection (`transform.GetType().Name` → would print `Func\`2` instead of `Transformer`). It does not: **both `gsi` and `gsc` reject it with `GS0159: Cannot find function GetType`**, symmetrically. My own speculative finding, killed by measurement.

### Mutation

The PR is a single hunk, so revert *is* the mutation — and the pure-`main` build is bit-identical to that mutant, so no separate build was needed. Running the PR's own test file against `main`'s `GSharp.Core.dll`:

```
Failed: 4, Passed: 0, Total: 4
```

**4/4 killed. Owner's "4/4 → mutant 0/4" claim independently verified.** No unpinned hunk.

Regression: `Interpreter.Tests` filtered on `Delegate|Closure|Lambda|Func` → **91 passed / 0 failed** at head.

### Ponytail: 2 lines is correct, do not ask for less

- **Rung 2 satisfied outright.** `DelegateTypeSymbol`'s own remarks at `src/Core/CodeAnalysis/Symbols/DelegateTypeSymbol.cs:24-30` prescribe this exact pattern: *"Callers that need a CLR `Type` (for example, conversion classification) should fall back to the `EquivalentFunctionType` shape."* The PR reuses the documented mechanism instead of inventing one.
- **It is the root site, not a symptom patch.** `Evaluator.Calls.cs:641` is the only `Unexpected type` throw in the evaluator, and `:603` the only delegate-conversion arm. `Evaluator.Expressions.cs:837` is invocation, not conversion, and does no `ClrType` lookup on the target type — no sibling defect. That is why all five broken shapes above fall out of one 2-line change.
- **Blast radius is provably bounded.** `A ?? B` evaluates `B` only when `A` is null, so any target with a non-null `ClrType` behaves bit-identically. The change can only affect programs that previously threw.
- The tempting "lazier" alternative — make `DelegateTypeSymbol.ClrType` itself return the equivalent — would be **wrong**: `ClrType != null` is used as a discriminator two arms below (`node.Type?.ClrType != null && !IsValueType`) and across the binder/emitter. Local fallback is correct. **Not proposing a smaller shape.**

---

## SHOULD-FIX 1 — the fix is broader than the tests, the title, and the issue

The 4 theory cases cover 1-param void, 1-param value, generic, and variadic — good, with distinct expected values. But **four further shapes were equally broken on `main` and are equally fixed here, with zero coverage**: **0-param**, **2-param**, **method group → named delegate**, and **named delegate passed as a call argument** (rows 2, 3, 4, 5 above; all `GS9999` on `main`).

Method group in particular is a different binder path from a function literal, and nothing in the suite would catch it regressing.

**Fix:** add the method-group case and the 0-param case to `Cases` in `test/Interpreter.Tests/Issue2985NamedDelegateConversionTests.cs`. Two more `TheoryData` rows, verbatim from the table above.

## SHOULD-FIX 2 — pin the identity guarantee

`Alpha`/`Beta` staying `GS0155` is the property that makes this fix safe, and **nothing asserts it**. It is exactly the kind of invariant that a future "let the evaluator be more permissive" change would silently drop, because the evaluator would then be reached.

**Fix:** one assertion that the `Alpha` → `Beta` program yields `GS0155`. Cheap, and it is the guard on the only real risk this PR carries.

## NIT — scope note for the issue title

Issue #2985 is phrased around conversion of a function literal. The defect is really *"the interpreter cannot convert **any** delegate value to a source-declared named delegate type, because `DelegateTypeSymbol.ClrType` is null before emit."* Worth noting on the issue so the method-group and argument-passing shapes are not re-filed separately.

## Relationship to #2948 / #3019 (imported named-delegate identity) — orthogonal

Provable from the operator, not just probed: imported delegate types **do** have a non-null `ClrType`, so they take the left side of the `??` and reach `delegateType` on exactly the path they reach today. The new fallback is unreachable for them. This PR neither helps nor hurts that work.

---

## Correction to the review brief

> tests `…Issue2985NamedDelegateConversionTests.cs`

Located in **`test/Interpreter.Tests/`**, not `Compiler.Tests` — worth naming precisely, since a filter run against the wrong project returns an empty set that reads as "no tests exist". (`Core.Tests` and `Compiler.Tests` both return 0 for this filter; the tests are real.)

DavidObando:
Addressed both SHOULD-FIX items in `c872ed0cf`:

- added distinct 0-param (`77`), 2-param (`311`), method-group (`33`), and argument-slot (`11`) coverage
- added `StructurallyEquivalentNamedDelegates_RemainDistinct`, asserting `GS0155`
- rebased onto current `origin/main`; build clean, issue tests 9/9, delegate/closure/lambda/func sweep 104/104
- revert mutant builds clean, then fails all 8 conversion cases while identity test remains green

PR #3036, opened 2026-08-01
Fix generic iterator remap precedence

DavidObando:
## Summary

Follow-up to #3000 after it merged while blocker verification was still running.

Generic iterator-literal kickoff code constructs a state-machine `TypeSpec` with the lambda method's `MVAR` arguments, but each field `MemberRef` signature must match the state-machine `FieldDef`'s `VAR` slot. The `IsMethodTypeParameter` guards in `SignatureEncoder` and `ImportedMemberRefFactory` let receiver class parameters fall through to the lambda remap, producing:

```text
MissingFieldException: Field not found: '<<lambda1>>d__1`2.<>3__first'
```

This change removes both guards. While a state-machine remap scope is active, every mapped source parameter uses the state-machine class's `VAR` slot. The lambda remap remains the fallback when no state-machine remap is active.

## Dropped remap rewrite

The earlier `ReflectionMetadataEmitter` rewrite has been removed because it is equivalent to the existing offset calculation:

- `SourceTypeParameters` is the receiver parameter prefix followed by `Function.TypeParameters`.
- `BuildRemap()` maps each source parameter to its array index.
- A lambda remap maps each original parameter to the corresponding function-parameter index `i`.
- Therefore `BuildRemap()[Function.TypeParameters[i]]` equals `SourceTypeParameters.Length - Function.TypeParameters.Length + i`.

The lambda remap is registered only when at least one referenced type parameter was promoted, so its production path also guarantees a non-empty function/source parameter list. Keeping the rewrite added null and missing-key paths without changing a reachable ordinal. The reviewer’s deletion mutant surviving 14/14 was correct evidence; the hunk is gone.

The anchor remains `ReflectionMetadataEmitter.RegisterGeneratedGenericRemaps`, now at lines 1271-1295 on `61b7a08c1`.

### Measured path audit

A temporary probe changed the receiver to `Factory[Unused, T]` while the iterator literal referenced only `T` and method parameter `U`: a strict subset of the enclosing type parameters. The observed path was:

```text
source=2 function=2 hasLambda=True lambda=2 remapNull=False missing=0 offset=0
```

The function literal carries only its promoted referenced parameters; it has no generic receiver/static-owner prefix, so the requested `offset != 0` lambda shape is not constructible. Across all ten iterator-literal cases, the probe examined ten state machines:

- 8: `source=0 function=0 hasLambda=False remapNull=True`
- 2: `source=1 function=1 hasLambda=True lambda=1 remapNull=False missing=0 offset=0`

This measures all three proposed failure classes. A null `BuildRemap()` occurs only without a lambda remap, where the existing null guard deliberately skips registration. Every observed lambda remap was non-empty, had a non-null state-machine remap, and every function parameter was present, so neither the NRE, missing-key read, nor silent empty registration path is reachable. The deleted rewrite would create those paths; current code preserves the initializer and guarded registration.

## Metadata regression

`GenericReceiverStateMachineMemberSignatures_UseSourceTypeParameterOrdinals` compiles an iterator literal inside `Factory[T].make[U]`, reads emitted PE metadata, and asserts:

- state-machine `FieldDef` signatures: `first = VAR 0`, `second = VAR 1`, `current = VAR 1`
- generic `get_Current` return: `VAR 1`
- kickoff field `MemberRef` signatures: `first = VAR 0`, `second = VAR 1`
- kickoff parent `TypeSpec` uses the lambda method's `MVAR` arguments

It then runs `IlVerifier`, loads the assembly with `Assembly.Load(File.ReadAllBytes(...))`, calls `GetTypes()`, and executes through the bounded dual-pipe runner, producing `ok`.

## Independent mutants

- Restore the `ImportedMemberRefFactory` guard: `ImportedElementToken_ClassTypeParameter_PrefersStateMachineRemap` fails, expected `GenericTypeParameter`, actual `GenericMethodParameter` (1 failed / 2 passed).
- Restore the `SignatureEncoder` guard: `GenericReceiverStateMachineMemberSignatures_UseSourceTypeParameterOrdinals` fails with `Member reference '<>3__first' did not encode GenericTypeParameter(0)` (1 failed / 0 passed).
- Restore either hunk afterward: its narrow selector passes.

## Verification

- Rebased onto `main` at `61b7a08c1`; `git range-diff` reports the patch unchanged
- Receiver metadata/runtime test: 1/1
- Remap precedence tests: 3/3
- Full solution build with `ContinuousIntegrationBuild=true`: 0 warnings, 0 errors
- Zero-byte gate: 0 found across 51 `GSharp.Core.dll` copies
- Five ordinary copies agree: `c72f16c677a60afa659718a275bc4854`
- Three-dot diff: 4 files, +184/-12

#2930's pending cache fix is in `MetadataTokenCache.CtorRefSymbolKey`; this PR changes `SignatureEncoder.EncodeTypeSymbol` and `ImportedMemberRefFactory.GetElementTypeToken`, so the root causes do not land in the same function.

### DIFF
--- a/src/Core/CodeAnalysis/Emit/ImportedMemberRefFactory.cs
+++ b/src/Core/CodeAnalysis/Emit/ImportedMemberRefFactory.cs
@@ -272,10 +272,9 @@ internal sealed class ImportedMemberRefFactory
             var sigBlob = new BlobBuilder();
             var encoder = new BlobEncoder(sigBlob).TypeSpecificationSignature();
 
-            // Issue #810: inside an iterator state-machine method body,
-            // outer-method TPs are remapped to the SM class's own TPs.
+            // Issue #810: inside an iterator state-machine member,
+            // source TPs are remapped to the SM class's own TPs.
             if (this.remaps.ActiveIteratorStateMachineRemap != null
-                && tpSym.IsMethodTypeParameter
                 && this.remaps.ActiveIteratorStateMachineRemap.TryGetValue(tpSym, out var smClassOrd))
             {
                 encoder.GenericTypeParameter(smClassOrd);
--- a/src/Core/CodeAnalysis/Emit/SignatureEncoder.cs
+++ b/src/Core/CodeAnalysis/Emit/SignatureEncoder.cs
@@ -265,9 +265,8 @@ internal sealed class SignatureEncoder
             // emits `<Empty>d__0<T>` where the body's `!!0` becomes
             // `!0` on the SM class. The remap is pushed by
             // EmitIteratorStateMachineMember and is keyed by the
-            // method TP's instance identity.
+            // source TP's instance identity.
             if (this.remaps.ActiveIteratorStateMachineRemap != null
-                && tp.IsMethodTypeParameter
                 && this.remaps.ActiveIteratorStateMachineRemap.TryGetValue(tp, out var classOrdinal))
             {
                 encoder.GenericTypeParameter(classOrdinal);
--- a/test/Compiler.Tests/Emit/Issue2957FunctionLiteralIteratorTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2957FunctionLiteralIteratorTests.cs
@@ -6,7 +6,10 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Reflection;
+using System.Reflection.Metadata;
+using System.Reflection.PortableExecutable;
 using Xunit;
 
 namespace GSharp.Compiler.Tests.Emit;
@@ -195,8 +198,155 @@ public sealed class Issue2957FunctionLiteralIteratorTests
         Assert.Equal(expectedOutput, RunBounded(name, assemblyPath));
     }
 
+    [Fact]
+    public void GenericReceiverStateMachineMemberSignatures_UseSourceTypeParameterOrdinals()
+    {
+        const string name = "GenericStateMachineMemberSignatures";
+        const string source = """
+            package GenericSignatures
+            import System
+
+            class Factory[T any] {
+                func make[U any]() (T, U) -> sequence[U] {
+                    return func(first T, second U) sequence[U] {
+                        yield second
+                    }
+                }
+            }
+
+            for value in Factory[int32]().make[string]()(42, "ok") {
+                Console.WriteLine(value)
+            }
+            """;
+
+        var assemblyPath = Compile(name, source);
+        using (var stream = File.OpenRead(assemblyPath))
+        using (var peReader = new PEReader(stream))
+        {
+            var metadata = peReader.GetMetadataReader();
+            var stateMachine = metadata.TypeDefinitions
+                .Select(metadata.GetTypeDefinition)
+                .Single(type => metadata.GetString(type.Name).Contains(">d__", StringComparison.Ordinal));
+
+            Assert.Equal(2, stateMachine.GetGenericParameters().Count);
+            AssertFieldTypeParameterOrdinal(metadata, stateMachine, "<>3__first", 0);
+            AssertFieldTypeParameterOrdinal(metadata, stateMachine, "<>3__second", 1);
+            AssertFieldTypeParameterOrdinal(metadata, stateMachine, "<>2__current", 1);
+            AssertMemberReferenceTypeParameterEncoding(
+                metadata,
+                "<>3__first",
+                SignatureTypeCode.GenericTypeParameter,
+                0);
+            AssertMemberReferenceTypeParameterEncoding(
+                metadata,
+                "<>3__second",
+                SignatureTypeCode.GenericTypeParameter,
+                1);
+
+            var currentOrdinal = stateMachine.GetMethods()
+                .Select(metadata.GetMethodDefinition)
+                .Where(method => metadata.GetString(method.Name) == "get_Current")
+                .Select(method => ReadMethodReturnTypeParameterOrdinal(metadata, method))
+                .Single(ordinal => ordinal.HasValue);
+            Assert.Equal(1, currentOrdinal);
+        }
+
+        IlVerifier.Verify(assemblyPath);
+        Assert.NotEmpty(Assembly.Load(File.ReadAllBytes(assemblyPath)).GetTypes());
+        Assert.Equal("ok\n", RunBounded(name, assemblyPath));
+    }
+
     private static object[] Case(string name, string source, string expectedOutput)
         => new object[] { name, source, expectedOutput };
+
+    private static void AssertFieldTypeParameterOrdinal(
+        MetadataReader metadata,
+        TypeDefinition stateMachine,
+        string fieldName,
+        int expectedOrdinal)
+    {
+        var field = stateMachine.GetFields()
+            .Select(metadata.GetFieldDefinition)
+            .Single(candidate => metadata.GetString(candidate.Name) == fieldName);
+        var signature = metadata.GetBlobReader(field.Signature);
+
+        Assert.Equal(SignatureKind.Field, signature.ReadSignatureHeader().Kind);
+        Assert.Equal(SignatureTypeCode.GenericTypeParameter, signature.ReadSignatureTypeCode());
+        Assert.Equal(expectedOrdinal, signature.ReadCompressedInteger());
+        Assert.Equal(0, signature.RemainingBytes);
+    }
+
+    private static void AssertMemberReferenceTypeParameterEncoding(
+        MetadataReader metadata,
+        string memberName,
+        SignatureTypeCode expectedCode,
+        int expectedOrdinal)
+    {
+        foreach (var handle in metadata.MemberReferences)
+        {
+            var member = metadata.GetMemberReference(handle);
+            if (metadata.GetString(member.Name) != memberName)
+            {
+                continue;
+            }
+
+            var signature = metadata.GetBlobReader(member.Signature);
+            if (signature.ReadSignatureHeader().Kind == SignatureKind.Field
+                && signature.ReadSignatureTypeCode() == expectedCode
+                && signature.ReadCompressedInteger() == expectedOrdinal
+                && ParentUsesGenericMethodTypeArguments(metadata, member.Parent))
+            {
+                return;
+            }
+        }
+
+        Assert.Fail(
+            $"Member reference '{memberName}' did not encode {expectedCode}({expectedOrdinal}).");
+    }
+
+    private static bool ParentUsesGenericMethodTypeArguments(
+        MetadataReader metadata,
+        EntityHandle parent)
+    {
+        const byte GenericInstance = 0x15;
+        const byte GenericMethodParameter = 0x1e;
+        if (parent.Kind != HandleKind.TypeSpecification)
+        {
+            return false;
+        }
+
+        var specification = metadata.GetTypeSpecification((TypeSpecificationHandle)parent);
+        var signature = metadata.GetBlobReader(specification.Signature);
+        if (signature.ReadByte() != GenericInstance)
+        {
+            return false;
+        }
+
+        signature.ReadByte();
+        signature.ReadCompressedInteger();
+        return signature.ReadCompressedInteger() > 0
+            && signature.ReadByte() == GenericMethodParameter;
+    }
+
+    private static int? ReadMethodReturnTypeParameterOrdinal(
+        MetadataReader metadata,
+        MethodDefinition method)
+    {
+        var signature = metadata.GetBlobReader(method.Signature);
+        var header = signature.ReadSignatureHeader();
+        if (header.IsGeneric)
+        {
+            signature.ReadCompressedInteger();
+        }
+
+        signature.ReadCompressedInteger();
+        if (signature.ReadSignatureTypeCode() != SignatureTypeCode.GenericTypeParameter)
+        {
+            return null;
+        }
+
+        return signature.ReadCompressedInteger();
+    }
 
     private static string Compile(string name, string source)
     {
--- a/test/Core.Tests/CodeAnalysis/Emit/GenericRemapPrecedenceTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Emit/GenericRemapPrecedenceTests.cs
@@ -40,7 +40,26 @@ public sealed class GenericRemapPrecedenceTests
     }
 
     [Fact]
-    public void ImportedElementToken_ClassTypeParameter_UsesLambdaRemap()
+    public void ImportedElementToken_MethodTypeParameter_UsesLambdaRemapWithoutStateMachine()
+    {
+        var typeParameter = new TypeParameterSymbol(
+            "T",
+            0,
+            TypeParameterConstraint.Any,
+            TypeParameterVariance.None)
+        {
+            IsMethodTypeParameter = true,
+        };
+
+        AssertEncoding(
+            typeParameter,
+            SignatureTypeCode.GenericMethodParameter,
+            expectedOrdinal: 2,
+            includeStateMachineRemap: false);
+    }
+
+    [Fact]
+    public void ImportedElementToken_ClassTypeParameter_PrefersStateMachineRemap()
     {
         var typeParameter = new TypeParameterSymbol(
             "T",
@@ -50,14 +69,15 @@ public sealed class GenericRemapPrecedenceTests
 
         AssertEncoding(
             typeParameter,
-            SignatureTypeCode.GenericMethodParameter,
-            expectedOrdinal: 2);
+            SignatureTypeCode.GenericTypeParameter,
+            expectedOrdinal: 1);
     }
 
     private static void AssertEncoding(
         TypeParameterSymbol typeParameter,
         SignatureTypeCode expectedCode,
-        int expectedOrdinal)
+        int expectedOrdinal,
+        bool includeStateMachineRemap = true)
     {
         var emitter = CreateEmitter();
         var stateMachine = new StructSymbol(
@@ -75,9 +95,13 @@ public sealed class GenericRemapPrecedenceTests
             TypeSymbol.Void,
             package: Package);
 
-        emitter.remaps.RegisterClassRemap(
-            stateMachine,
-            new Dictionary<TypeParameterSymbol, int> { [typeParameter] = 1 });
+        if (includeStateMachineRemap)
+        {
+            emitter.remaps.RegisterClassRemap(
+                stateMachine,
+                new Dictionary<TypeParameterSymbol, int> { [typeParameter] = 1 });
+        }
+
         emitter.remaps.RegisterLambdaMethodRemap(
             lambda,
             new Dictionary<TypeParameterSymbol, int> { [typeParameter] = 2 });


DavidObando:
Blocker 3 is fixed at `844ac1055`. The remap now resolves each promoted lambda clone through `BuildRemap()` instead of recomputing an offset. The metadata regression decodes `VAR 0`/`VAR 1` directly; a `slot + 1` mutant fails with `Expected: 0, Actual: 1`. Final CI: **8 SUCCESS + 2 SKIPPED**.

DavidObando:
Generic-receiver coverage exposed the remaining `IsMethodTypeParameter` guard bug: the kickoff encoded `<>3__first` as `MVAR`, then threw `MissingFieldException`. Head `91209696c` removes both guards and decodes the kickoff `MemberRef` signatures directly. The prior green rollup applied to the superseded head; new CI is running.

DavidObando:
## Review: NOT MERGEABLE — two of three production hunks are well pinned; the third, which the branch is named for, survives deletion

Reviewed at head **`91209696`**, `main` = `8acc2cc76`, computed merge-base = `8acc2cc76` (three-dot; genuinely rebased, 0 behind). Net at tip: **5 files, +187 −16**.

**Process note first:** this PR was force-pushed while I was reviewing it. The earlier head `844ac105` had a *different, smaller* change (2 files; `ReflectionMetadataEmitter.cs` only; a test named `GenericStateMachineMemberSignatures…` with no `GenericReceiver` variant) and its merge-base was `7a6307a79`, not `8acc2cc76`. All measurements below are at `91209696`; anyone holding results from `844ac105` should discard them.

Build: `dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true`, rc=0, clean Core md5 `36bc06bc03739eea70fc3a0308d626ac`. (This differs from the body's `63b7cfd1…` simply because the build root differs — not a finding.)

Baseline, with the denominator stated: **Compiler.Tests `~Issue2957FunctionLiteralIterator` = 11**, **Core.Tests `~GenericRemapPrecedence` = 3**, total **14**. Note `GenericRemapPrecedenceTests.cs` lives in **Core.Tests**, not Compiler.Tests — filtering it in Compiler.Tests returns an empty result that reads like a pass.

### Hunks 2 and 3 are properly pinned — the body's claim reproduces verbatim

Re-adding both `IsMethodTypeParameter` guards (`ImportedMemberRefFactory.cs:277`, `SignatureEncoder.cs:270`), rebuilding, and re-running:

```
Failed GenericReceiverStateMachineMemberSignatures_UseSourceTypeParameterOrdinals
Failed!  - Failed: 1, Passed: 10, Total: 11
  → did not encode GenericTypeParameter(0)
```

That is exactly the failure and exactly the message the PR body predicts. Core md5 moved to `8c5848ec…` for that build, so the mutant was in the binary. This half of the change is load-bearing and well covered — the test also runs `IlVerifier`, then `Assembly.Load(File.ReadAllBytes(...))` + `GetTypes()`, then executes through the bounded runner and asserts a real value. That is the right shape for this area.

---

## BLOCKER — the `ReflectionMetadataEmitter` hunk is pinned by no test

`src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs:1283-1287`

Reverting **only** this hunk to the previous offset arithmetic and leaving hunks 2 and 3 intact:

| suite | result with hunk 1 reverted |
|---|---:|
| Compiler.Tests `~Issue2957FunctionLiteralIterator` | **11/11 passed** |
| Core.Tests `~GenericRemapPrecedence` | **3/3 passed** |
| **total** | **14/14 survive** |

Verified in the binary, not assumed: Core md5 `36bc06bc…` (clean) → `48b788e1…` (hunk 1 reverted) → `36bc06bc…` (restored), tree clean.

This is the hunk the branch is named for (`fix/3000-remap-source-slots`) and the one the summary leads with. Right now nothing distinguishes it from its absence, which means either it is a no-op in every reachable shape, or it fixes a case no test reaches. Both need resolving before merge.

I also tried to build a behavioural probe for it and **failed to make one discriminate** — reporting that rather than presenting it as a pass. Two swapped instantiations in one program, distinct values in every slot, multiple `yield`s, and a `yield` after a loop:

```gsharp
func make[T any, U any]() (T, U) -> sequence[U] {
    return func(first T, second U) sequence[U] { yield second; yield second }
}
for v in make[int32, string]()(11, "aa") { Console.WriteLine(v) }
for v in make[string, int32]()("bb", 22) { Console.WriteLine(v) }
```

`gsi`, `gsc` at HEAD, and `gsc` with hunk 1 reverted **all** produce `aa aa 22 22 33 dd` (and `zz zz zz 77 77 77` for the yield-after-loop variant). In these shapes `SourceTypeParameters.Length == Function.TypeParameters.Length`, so the old `offset` is 0 and the two implementations coincide. **My probe is insensitive and is not evidence for either side.**

**Recommended fix:** add a test in the shape where the lambda captures a strict *subset* of the enclosing type parameters, so `offset != 0` and the old and new expressions genuinely differ — that is the case the summary describes. If no such shape exists, then the hunk is a refactor and the body should say so rather than presenting it as the fix.

## SHOULD-FIX — the deleted `remap ??=` is still required by the line below it

Same hunk. The change removes

```csharp
remap ??= new Dictionary<TypeParameterSymbol, int>();
```

but `BuildRemap()` can return **null** — `StateMachineEmitter.cs:305` and `:349` both `return null` when `SourceTypeParameters.IsDefaultOrEmpty` — and the code immediately after the hunk still guards `if (remap != null)` at `:1291`, so the codebase itself treats null as live there.

Three consequences, all newly possible:

1. **`NullReferenceException`** at `remap[pair.Key] = …` if `BuildRemap()` returns null while `TryGetLambdaMethodRemap` yields a non-empty map.
2. **`KeyNotFoundException`** at `remap[sourceTypeParameter]` — this is now a dictionary *read*, where the old code computed a value arithmetically and could not miss.
3. A **silent branch change**: if `remap` is null and `lambdaRemap` is empty, the old code left an empty dictionary and `RegisterClassRemap` at `:1292` ran; the new code leaves `remap` null and **skips the registration entirely**.

Case 3 needs no exception to bite and is invisible in any output. If these states are genuinely unreachable, an assert or a comment saying so is cheap; if they are reachable, item 1 is an emitter crash.

## NIT — undeclared scope in the summary

The body frames this as a follow-up to #3000's offset, then describes the generic-receiver guard removals separately. The guard removals are the part that is actually pinned and measured; the offset change is the part that is not. The summary's ordering implies the reverse.

---

**Verdict: NOT MERGEABLE.** Hunks 2 and 3 are good work and I verified the author's own failure claim reproduces exactly. But hunk 1 — the change the branch is named for — survives deletion against **14 of 14** tests, and it removes a null guard that the very next line still depends on. Given iterator rewriting is where a silent, exit-0, CI-green wrong answer has already cost this repo a four-cell matrix (#3013 vs #2967), an unpinned hunk here is not something to take on trust.

DavidObando:
Final head `91209696c` is green: **8 SUCCESS + 2 SKIPPED**. Generic receiver `Factory[T].make[U]` now emits state-machine `FieldDef` and kickoff `MemberRef` signatures as `VAR 0`/`VAR 1`, loads, and prints `ok`. Re-adding either class-parameter guard is killed by direct metadata decoding before IL verification or execution.

DavidObando:
Rebased only onto `fd34166a8`; content is unchanged. Re-ran the two narrow probes after the rebase: receiver metadata/runtime 1/1 and remap precedence 3/3. Merge build passes with 0 warnings/errors; zero-byte gate examined 40 `GSharp.Core.dll` copies and found 0, and all five ordinary copies hash to `6afb7db88512e22261a320a4adf9baa8`. Prior CI rollup is superseded; new CI is running.

DavidObando:
Rebased-only head `8d65af8ff` is green against `fd34166a8`: **8 SUCCESS + 2 SKIPPED**. Post-rebase probes and merge build are recorded in the PR body.

DavidObando:
I am addressing the blocker by dropping hunk 1 if the source/type-parameter ordering invariant proves it equivalent, then validating hunks 2 and 3 independently. This will require a force-push after the checks finish; please hold review until I post the new head.

DavidObando:
New head: `79753d569` (rebased onto `a4a660f32`).

Dropped hunk 1. It cannot produce a discriminating program because the lookup and existing offset are algebraically equivalent: source parameters are receiver-prefix + function parameters, so `BuildRemap()[Function.TypeParameters[i]] == receiverCount + i == Source.Length - Function.Length + i`. Its production lambda-remap path is non-empty. Keeping the rewrite only opened null/missing-key paths.

Final content is two guard removals plus their tests. Each guard mutant fails its intended narrow test. Post-rebase tests, verifier/load/runtime probe, solution build, and artifact gates pass.

DavidObando:
Final CI at `79753d569`: 8 successful, 2 skipped, 0 other. Local solution build, narrow mutants, metadata/load/runtime probe, and artifact gates remain green.

DavidObando:
Rebase starting now; force-push will follow. Current head `79753d569` already contains Answer B: the `ReflectionMetadataEmitter` hunk was deleted, leaving that file identical to its base. I will report post-rebase content changes separately from commit-graph changes.

DavidObando:
Rebased only onto `61b7a08c1`; new head `fe675800a`. `git range-diff` marks the sole patch `=`: no content change. Fresh three-dot diff remains 4 files; `ReflectionMetadataEmitter.cs` contributes 0 diff lines.

DavidObando:
Answer B: hunk 1 is dead and remains deleted. Fresh anchor: `ReflectionMetadataEmitter.RegisterGeneratedGenericRemaps`, lines 1271-1295 on `61b7a08c1`.

I built the requested strict-subset shape as `Factory[Unused, T]`, with the iterator literal referencing only `T` and method parameter `U`. Instrumentation measured `source=2 function=2 hasLambda=True lambda=2 remapNull=False missing=0 offset=0`. Function literals carry only their promoted referenced parameters; they do not acquire the enclosing generic receiver/static-owner prefix, so `offset != 0` is not constructible on this lambda-remap path.

Path audit denominator: 10 iterator state machines. Eight measured `source=0 function=0 hasLambda=False remapNull=True`; two measured `source=1 function=1 hasLambda=True lambda=1 remapNull=False missing=0 offset=0`. Thus null remaps occur only when there is no lambda remap and registration is intentionally skipped. Registered lambda remaps are non-empty; their state-machine remaps are non-null and contain every function parameter. The proposed NRE, missing-key read, and silent empty-registration paths belong only to the deleted rewrite.

Post-rebase guard mutants still fail independently and restore to the original Core hash. Solution build: 0 warnings/errors. Metadata load/GetTypes/runtime test: 1/1. Core precedence tests: 3/3. #2930 targets `MetadataTokenCache.CtorRefSymbolKey`; this PR does not change that function.

DavidObando:
Final CI at `fe675800a`: 8 successful, 2 skipped, 0 other. Rebased blocker answer and measured path audit are now complete.

DavidObando:
## Duck verdict — **MERGEABLE**

Trees named. **`main` is `5a27b5532`, not the briefed `01c1a6c5c`** (2 commits further; briefed SHA is an ancestor). Head `fe675800a` confirmed against the `gh` API. Merged tree = **`80642eb35`**, parents exactly `5a27b5532` + `fe675800a`. Merge-base `61b7a08c1`; three-dot: **4 files / +184 / −12**.

### Headline answers

**(a) The remaining hunks are KILLED, not 14/14.** Round 1's survival does not hold at this head — the PR added the pins. Mutated **independently**, each a product mutant restoring the deleted guard, each a full `dotnet build GSharp.sln` with `BUILD_RC=0`, `errs=0`, md5 changed, driver copies re-stamped:

| mutant | site | result |
|---|---|---|
| restore `&& tpSym.IsMethodTypeParameter` | `ImportedMemberRefFactory.cs:275` | **KILLED** — `ImportedElementToken_ClassTypeParameter_PrefersStateMachineRemap` |
| restore `&& tp.IsMethodTypeParameter` | `SignatureEncoder.cs:268` | **KILLED** — `GenericReceiverStateMachineMemberSignatures_UseSourceTypeParameterOrdinals` |

The `SignatureEncoder` mutant also **reproduces the production defect exactly** — `exec_rc=134`, `MissingFieldException: '<<lambda1>>d__1\`2.<>3__first'`. That is a value oracle, not a status one. Cycle closed: clean `342033af…` → mutant differs → **restored equals clean**, porcelain 0.

**(b) Hunk 1 was genuinely dead. Deleting it is correct.** I did not take the equivalence argument on faith — I made the compiler count. Instrumented `RegisterGeneratedGenericRemaps` to log `offset`, `SourceTypeParameters.Length`, `Function.TypeParameters.Length` on **every** invocation:

- Full `Compiler.Tests` (4028 tests): **3 invocations, all `offset=0`**.
- Four **deliberately adversarial strict-subset shapes** I built for exactly your `offset != 0` request — receiver `[T, Unused]` with the literal referencing `{T}`; receiver `[A,B,C]` referencing only `C`; a shape with **no method type parameter at all**; and a two-capture superset shape: **5 invocations, all `offset=0`**.
- **Total 8 invocations, `offset != 0` count = 0**, and `src == fn` in every single one.

The probe is positively controlled — it emitted 8 lines, so a zero here is absence, not an unrun check. The mechanism is now measured rather than asserted: **the function literal carries only its own promoted referenced parameters and has no generic receiver prefix**, so `SourceTypeParameters.Length == Function.TypeParameters.Length` structurally. The `offset != 0` shape is **not constructible through a function literal**. Both implementations coincide on every reachable path. Ponytail approves the deletion — 0 lines beats 25.

**(c) #3000's defect flips base → head.** Compiled with the merge-base compiler (`61b7a08c1`) and the merged-tree compiler:

| shape | base | head |
|---|---|---|
| `Factory[T].make[U]` iterator literal | `exec 134` `MissingFieldException '<<lambda1>>d__1\`2.<>3__first'` | **`exec 0` → `ok`** |
| `Factory[T, Unused]`, subset `{T}` + `U` | `exec 134` | **`exec 0` → `sa-ok`** |
| `Factory[A,B,C]`, references only `C` | `exec 134` | **`exec 0` → `sb-ok`** |
| `Factory[T, Unused]`, **no method TP** | `exec 134` | **`exec 0` → `sc-ok`** |
| two captures, receiver superset | `exec 134` | **`exec 0` → `sd-ok`** |

**All ten cells compile `rc=0`.** The defect is invisible to the compiler, to exit codes, and to `IlVerifier` — it only appears on execution. Worth stating plainly because this PR's whole risk class is silent-at-emit.

**(d) Merged against current `main`: green.** `git merge-tree refs/main refs/pr3036` → **rc=0, no conflict**. Merged tree builds **`BUILD_RC=0`, 0 Warning(s), 0 Error(s)**; all `GSharp.Core.dll` copies share one mtime (never Core-only; every value probe ran against a solution build with a driver copy at least as new).

| suite | result |
|---|---|
| `Compiler.Tests` | **4028 / 4028** passed (1 h 26 m) |
| `Core.Tests` | **7068** passed, 1 skipped, 7069 total |
| `Cs2Gs.Tests` (the corpus gate you flagged) | **1861 / 1861** passed |
| `Interpreter.Tests` | **591 / 591** passed |

Docs gate **not exposed** — the PR touches no descriptor or docs file; merged tree measures 377 descriptors / 383 `docs` rows / 381 `website` rows, same as `main`.

### Corrections to the briefing — three anchors were wrong

1. **`main` is `5a27b5532`**, not `01c1a6c5c`.
2. **The anchor `ReflectionMetadataEmitter.RegisterGeneratedGenericRemaps:1271-1295` is not in this PR.** That file contributes **0 lines** to the three-dot diff — as the owner said. The remaining production change is **8 lines across two other files**: `SignatureEncoder.cs:265-272` and `ImportedMemberRefFactory.cs:272-279`. Mutating `:1271-1295` would have mutated `main`, not the PR, and produced a meaningless result.
3. **#3000 is a merged PR, not an issue** (`Fixes #2957`, `#2976`). #3036 is a follow-up to a defect #3000 introduced.

### The thing that looked like a weakened pin, and is not
#3036 **deletes** `ImportedElementToken_ClassTypeParameter_UsesLambdaRemap` — the exact test #3000's body cited as "kills removal of the new guard". That is the classic remove-the-guard-and-its-alarm shape. It is **not** that here: the test is replaced by its **inverse** (`…ClassTypeParameter_PrefersStateMachineRemap`, now asserting `GenericTypeParameter` ordinal 1 instead of `GenericMethodParameter` ordinal 2), plus a **new** `…MethodTypeParameter_UsesLambdaRemapWithoutStateMachine` pinning the no-state-machine fallback. I confirmed by mutation that the replacement genuinely kills. Correct transformation.

### NIT — `ImportedMemberRefFactory.cs:275`
That hunk is pinned **only** by a unit test that constructs `TypeParameterSymbol`/remaps directly. No end-to-end shape I could build exercises it: with that guard restored, my repro and all four strict-subset shapes still printed the right answer and exited 0 — only the `SignatureEncoder` hunk moved the needle end-to-end. So the second hunk's real-world reachability is asserted by the body, not demonstrated by a compiling program. Not a blocker (it is pinned, and it is symmetric with the sibling site), but if a shape exists that reaches it, an end-to-end case would be worth more than the unit test.

### Measurement error I made, disclosed
I first diffed `test/Core.Tests/Emit/GenericRemapPrecedenceTests.cs` and got an **empty result**, which would have read as "the test file is untouched". Wrong path — it lives at `test/Core.Tests/CodeAnalysis/Emit/`. Caught by running `--name-status` instead of guessing paths. An empty diff from a wrong path is indistinguishable from absence, and here it would have inverted a central finding.

**No BLOCKERs. No SHOULD-FIXes.**

DavidObando:
## Addendum to my MERGEABLE verdict — `{top-level, in-function} × {gsc, gsi}` matrix

Fleet bulletin #3 (PR #2963 source break: every case wrapped in `func Main()`). Re-measured my own probes. **Verdict unchanged: MERGEABLE.** Evidence widened 1 column → 4.

**Audit of my round-2 probes:** all 5 (`rep/sa/sb/sc/sd.gs`) were **top-level, 0 `func Main`**. But I inherited the author's shape anyway — this PR's test `GenericReceiverStateMachineMemberSignatures_UseSourceTypeParameterOrdinals` is top-level too, so **in-function and both `gsi` columns were unmeasured.** Closed that gap.

Trees: BASE `61b7a08c1` (merge base), HEAD = merged `80642eb35` (`main 5a27b5532` + `fe675800a`). Both porcelain 0; `Core/`+`Compiler/`+`Repl/` share one mtime per tree. No builds run (load 32.15 > 30).

| shape | BASE gsc | HEAD gsc | BASE gsi | HEAD gsi |
|---|---|---|---|---|
| top-level | rc134 `MissingFieldException '<<lambda1>>d__1\`2.<>3__first'` | **rc0 `ok`** | rc0 `ok` | rc0 `ok` |
| in-function (`func run()` called from top level) | rc134 same | **rc0 `ok`** | rc0 `ok` | rc0 `ok` |
| `func Main()` only | rc134 same | **rc0 `ok`** | **rc0, EMPTY** | **rc0, EMPTY** |

**`gsc` discriminates in all three scopes** — the fix holds on the axis I had not tested. **`gsi` discriminates in none**, on either side: the fix is signature/metadata emit (`SignatureEncoder.cs:265-272`, `ImportedMemberRefFactory.cs:272-279`), a path the interpreter never walks. Reporting that as **non-discriminating scope, not a pass** — "a zero can be scope, not absence".

### Measured refinement to the `func Main()` rule (positive-controlled)

Same code, three ways, `gsi` on HEAD:

| probe | result |
|---|---|
| top-level divide-by-zero | **rc1** `GS9999: Attempted to divide by zero.` (control: gsi *can* fail) |
| identical code inside `func Main()` | **rc0, empty** — failure invisible |
| binder error inside `func Main()` | **rc1** `GS0130: Function '…' doesn't exist.` |

So under `gsi` a `func Main()` body is **bound but never executed**. Compile-time diagnostics inside it still fire — which is the trap: a reviewer sanity-checking "is my probe live?" by inserting a typo gets `GS0130` and concludes yes, while **every runtime assertion in that file is vacuous and reads as rc0-clean**. Stronger than "untested": it reads as PASS.

### Bulletin's `is not LocalVariableSymbol` grep
**0 hits in this PR's three-dot diff** — controlled (`IsMethodTypeParameter` = 3 hits in same patch, so the search space is live). Genuine absence.

Separately, the merged tree has 3 occurrences; I read all 3 before blaming any. `StatementBinder.Narrowing.cs:353` and `:463` are **documented deliberate scope** ("Narrow locals only — never parameters, fields…"), fail-safe (refuse to narrow globals), and `:353` binds `out LocalVariableSymbol` so it is structurally locals-only. **Neither is the defect shape** — that one silently *skips invalidation* for globals. Flagging so nobody "fixes" these two and breaks documented behaviour.

### Self-disclosed measurement error
My first run of the grep above used `/tmp/r3/w`, a clone I keep **with no working tree checked out** (`src_exists=NO`). Both target and control returned 0. Had I not positive-controlled, I would have reported "pattern absent from the tree" from an **empty search space** — the exact `grep -c` on-nothing trap. Redone against a real checkout: control 266, target 3.

**Shelf life:** these numbers are true of merged tree `80642eb35`. `main` is now `185431c4` (#3053, #3064, #3033, #3063 landed since). Do not cite them for different code.

DavidObando:
## Review: PR #3036 — VERDICT: **MERGEABLE**

Measured on the **merged tree**: `main` **`5ad0109c`** + PR head **`fe675800a`** → merge **`5993d537`** (`merge-tree` conflicts: 0, `MERGE_RC=0`). Release build with `-p:ContinuousIntegrationBuild=true`: **0 warnings, 0 errors**. `GSharp.Core.dll` md5 `feee1a9755647b104eb68aa5c113f1b5`, all three copies (`Core/`, `Compiler/`, `Repl/`) mtime-verified fresh before every probe.

No blockers. The fix is real, is in the right place, and is load-bearing. Details and four non-blocking findings below.

---

### The fix is real — and broader than the PR claims

I reverted the `SignatureEncoder` hunk alone (mutant **M2**, restoring `&& tp.IsMethodTypeParameter` at `SignatureEncoder.cs:271`) and ran four probes through `gsc /out:` **and executed the emitted assembly**, each into its own empty directory:

| probe | shape | M2 (pre-fix) | merged HEAD |
|---|---|---|---|
| `p1_methodtp` | method TP yielded (author's shape) | `gsc` rc=0 → **`MissingFieldException: '<<lambda1>>d__1\`2.<>3__first'`, abort trap 6** | `ok` rc=0 |
| `p2_classtp` | **pure class TP, two params** | `gsc` rc=0 → **`MissingFieldException: '…<>3__a'`, abort 6** | `11,22` rc=0 |
| `p3_both` | class TP + method TP in one SM | `gsc` rc=0 → **`MissingFieldException: '…<>3__first'`, abort 6** | `11,22,22` rc=0 |
| `p4_nest` | class TP yielded, method TP consumed | `gsc` rc=0 → **`MissingFieldException: '…<>3__a'`, abort 6** | `22,11,11` rc=0 |

Exactly the family failure mode: **compile rc=0, silently invalid metadata, runtime abort.** Note `p2_classtp` — a *pure class-type-parameter* shape with no method type parameter in the signature at all. It is not in the PR's tests and it was broken before this change. The fix is worth more than the body claims.

M2 also kills the suite: `FullyQualifiedName~Issue2957` **1 failed / 10 passed** (baseline 11/11). Mutant **M1** (restoring `ImportedMemberRefFactory.cs:278`) kills `ImportedElementToken_ClassTypeParameter_PrefersStateMachineRemap`: **1 failed / 2 passed** (baseline 3/3). Both hunks are independently pinned — M1 does **not** disturb `Issue2957` (11/11), M2 does **not** disturb the Core precedence tests. The author's two mutant claims reproduce.

Also worth recording, since the PR body reasons about `#3000`'s intent: the `IsMethodTypeParameter` guard is **not** `#3000`'s. In `SignatureEncoder.cs` it dates to **#1361** (file creation); `#3000` only *promoted* that arm from `else if` to first position. In `ImportedMemberRefFactory.cs`, `#3000` **created** arm A outright, guarded, ahead of the pre-existing lambda arm. So this PR is not reversing a deliberate `#3000` decision about class TPs — `#3000` never changed class-TP behaviour. That strengthens the PR's framing.

### `gsi` pins nothing here — measured, not assumed

Under the **broken** M2 build, `gsi` returned rc=0 and the **correct** values on all four probes (`ok` / `11,22` / `11,22,22` / `22,11,11`) — byte-identical to the fixed build. The interpreter never emits, so it cannot see this class of defect. **Anyone validating this area with `gsi` will read a live emit bug as SURVIVED.** `gsc /out:` + `Assembly.Load` + execute is the only oracle here; `IlVerifier` alone is also insufficient.

---

## SHOULD-FIX 1 — the PR makes arm C provably dead in both files, and leaves it there

`SignatureEncoder.cs:288-292` and `ImportedMemberRefFactory.cs:290-295` ("Issue #1477") now have **conditions textually identical to arm A** in the same `if/else` chain:

```csharp
// arm A (guard removed by this PR)
if (this.remaps.ActiveIteratorStateMachineRemap != null
    && this.remaps.ActiveIteratorStateMachineRemap.TryGetValue(tp, out var classOrdinal))
{ encoder.GenericTypeParameter(classOrdinal); }
...
// arm C — same condition, same body, evaluated only if arm A failed
else if (this.remaps.ActiveIteratorStateMachineRemap != null
    && this.remaps.ActiveIteratorStateMachineRemap.TryGetValue(tp, out var classOrdinal2))
{ encoder.GenericTypeParameter(classOrdinal2); }
```

Both operands are side-effect-free and nothing mutates between them, so arm C is **unreachable by construction**. `#3000` added a guarded copy of an existing arm; this PR removes the guard — the duplication was built in two steps and this PR completes it.

**I implemented and measured the deletion** rather than asserting it. Removing arm C from both files (**−16 lines**, on top of the PR):

- Full **Core.Tests: 7073 passed / 0 failed / 1 skipped / 7074**
- Full **Compiler.Tests: 4046 passed / 0 failed / 4046** (1 h 11 m)
- All 4 probes: identical runtime output **and identical emitted-assembly size** vs merged HEAD
- Release build: **0 warnings, 0 errors**

**11,119 tests and 4 executed probes cannot tell the difference.** Diff saved at `/tmp/g3036/alt-delete-dead-armC.diff`.

Impact is maintainability, not correctness: the surviving `Issue #1477` comment documents a rule on code that can never run, so the next reader will believe #1477 is handled by an arm that is dead, and will "fix" #1477 by editing unreachable lines. Recommended fix: delete arm C in both files and fold its `#1477` sentence into arm A's comment — the net production diff becomes **−20 lines instead of −4**, which is the strictly lazier shape.

## SHOULD-FIX 2 — the `ImportedMemberRefFactory` hunk has no end-to-end evidence

I put a sentinel inside arm A of **each** file that throws on precisely the case this PR changes (class TP, present in the SM remap **and** the lambda remap). Positive control: it fires on the synthetic test, so the sentinel is live and discriminating.

| sentinel | synthetic unit test | full Core.Tests (7074) | 4 `gsc` probes |
|---|---|---|---|
| `SENTINEL_SE_DELTA` (`SignatureEncoder`) | fires | — | **fires 4/4** |
| `SENTINEL_IMRF_DELTA` (`ImportedMemberRefFactory`) | fires | **0 hits** | **0 hits** |

So `GetElementTypeToken` is **never** reached with a class TP in both remaps by any real compilation I could construct. Hunk 1 is pinned **only** by `GenericRemapPrecedenceTests.ImportedElementToken_ClassTypeParameter_PrefersStateMachineRemap`, which manufactures its own precondition via `RegisterClassRemap`/`RegisterLambdaMethodRemap` rather than compiling G#.

I am **not** calling this a defect: divergence between the two encoders of the same construct is exactly how this family breaks (a `TypeSpec` token disagreeing with a signature blob), so keeping them in lockstep is defensible. But the body presents both hunks as equally defect-driven, and they are not. Please state that hunk 1 is a **symmetry guard with no known reachable failure**, and say so in the test's name or an XML comment, so a future reader does not assume a real bug is pinned there.

## NIT 1 — the new Core test uses the exact shape that broke merged in #3055

`GenericRemapPrecedenceTests.cs:CreateEmitter` does:

```csharp
typeof(ReflectionMetadataEmitter).GetConstructors(BindingFlags.Instance | BindingFlags.NonPublic).Single()
...
constructor.Invoke(new object[] { program, references, "…", false, null });
```

`ReflectionMetadataEmitter.cs:719` currently has exactly one private 5-param ctor, so it passes today. This is the #3055 failure shape verbatim: adding one ctor parameter, or a second non-public ctor, breaks this test with `TargetParameterCountException` / `InvalidOperationException` **in the merge**, invisible to the PR that changes the ctor and to both branches' CI. Prefer an `internal` test-only factory, or `InternalsVisibleTo` + a named ctor, over reflective arity matching.

## NIT 2 — one quadrant of the precedence matrix is unpinned

The new file covers method-TP/with-SM, method-TP/without-SM, and class-TP/with-SM, but there is no **`ClassTypeParameter_UsesLambdaRemapWithoutStateMachine`**. That is the arm-B behaviour this PR deliberately *retains* for class TPs when no SM remap is active — the one thing the change is careful not to break. It is currently held by nothing. One more `AssertEncoding(..., GenericMethodParameter, 2, includeStateMachineRemap: false)` with a class TP closes the matrix.

---

### Adjacency re-check (`main` moved after this PR was written)

Both production files overlap landed `main` commits, so this is not the disjoint case: `ImportedMemberRefFactory.cs` also carries **#3068** (`MethodSpecSymbolKey`, `:871`/`:915`) and **#2912** (`:307`/`:1338`); `SignatureEncoder.cs` carries **#2912**. #3068's sites are in a different method ~600 lines away and key on symbol identity by reference, whereas this PR changes only TypeSpec/signature *encoding* — and the merged-tree full Compiler.Tests run (4046/4046) is the behavioural confirmation, not the line-distance argument.

### Disclosures

- My probe `p5_deep` was **void** — a G# syntax error on my part (`GS0005`), not a result; it is excluded from all counts above.
- My first reading of the sentinel data was **wrong in the safe direction**: full Core.Tests showed 0 real hits and I briefly took that as "the changed case is unreachable". The `gsc` probes then fired `SENTINEL_SE_DELTA` 4/4. **That zero was scope, not absence** — Core.Tests does not drive the full emit pipeline for generic iterator literals. Corrected before it reached any conclusion.
- `dotnet test` on Core.Tests desynchronised the DLL copies (`Core/` 01:28 vs `Compiler/`+`Repl/` 01:26); I re-ran a full solution build before every `gsc`/`gsi` probe.
- Tree restored: `git status --porcelain` empty, rebuild md5 **round-trips exactly** to `feee1a9755647b104eb68aa5c113f1b5`.

DavidObando:
### Re-anchor: `main` moved `5ad0109c` → `33bea4af`. **Verdict unchanged: MERGEABLE.**

My review above was measured against `main` `5ad0109c`. Two commits landed since — **#3047** (`33bea4af`) and **#3024** (`cceb0d5f`) — so per the rule that *`main` moving invalidates a measurement even when the branch does not move*, I re-merged and re-measured rather than assume it carried.

New merge **`7712aba8`** = `main 33bea4af` + `fe675800a`. `MERGE_RC=0`, `merge-tree` conflict markers **0**, Release build `-p:ContinuousIntegrationBuild=true` **0 warnings / 0 errors**, all three `GSharp.Core.dll` copies (`Core/`, `Compiler/`, `Repl/`) mtime-synced.

**#3047 is test-only** (263 lines, 1 file, zero production change) — precisely the #3055 shape: new tests landing on `main` after this PR's merge base, which this PR's own CI never ran. So I ran them against the merged tree:

- `Issue3001StructComputedAccessorVirtual` + `Issue2957`: **17 passed / 0 failed / 17**. Positive control on the count: `Issue3001` declares **6** `[Fact]`s and all 6 ran; 11 + 6 = 17, so the new file was genuinely executed, not filtered out.
- **#3024's new tests are in `test/Interpreter.Tests/`, a different project** my first filter could not have reached. Full **Interpreter.Tests: 610 passed / 0 failed / 610**.
- All four probes re-run at `33bea4af`, each into a fresh empty directory, `gsc /out:` + execute: `ok` / `11,22` / `11,22,22` / `22,11,11`, all rc=0.

Nothing moved. The four findings above (dead arm C; `ImportedMemberRefFactory` hunk being symmetry-motivated with no reachable case; the reflective-ctor test fragility; the missing matrix quadrant) all stand as written.

**One precision correction to my own comment above, unprompted:** I wrote "the merged-tree full Compiler.Tests run (4046/4046)". That run was on merged **plus** my 16-line arm-C deletion, not on merged-clean. The two are behaviourally identical *because* arm C is unreachable by construction — but that is an argument, not a measurement, and I should have labelled the tree precisely. I did not run full Compiler.Tests on merged-clean; the merged-clean evidence is full Core.Tests plus targeted Compiler filters plus the executed probes.

PR #3038, opened 2026-08-01
Fix nested struct CLR arity suffixes

### LINKED ISSUE #2916: Nested structs in generic outers use incorrect CLR arity suffix
## Repro

```gsharp
package StructInterop

public struct Outer[T] {
    public struct Inner { public var Value T }
}
```

`TypeDefEmitter.EmitNestedStructTypeDef` emits:

```text
StructInterop.Outer`1+Inner`1
```

with one `GenericParam` row copied from `Outer[T]`.

CLR nested-type naming counts a nested type's **own** generic parameters in its backtick suffix. `Inner` declares none, so Roslyn-compatible metadata is:

```text
StructInterop.Outer`1+Inner
```

Copied enclosing `T` still needs its `GenericParam` row; only name suffix is wrong.

## C# consumer failure

```csharp
public static StructInterop.Outer<int>.Inner Make() => default;
```

```text
error CS0426: The type name 'Inner' does not exist in the type 'Outer<int>'
```

Roslyn drops malformed nested name from normal C# lookup; malformed foreign metadata exposed directly is same unsupported-type class reported as `CS0648: 'type' is a type not supported by the language`.

G# API can return type and C# can use result through inference, but C# cannot name type, declare fields/parameters of it, or construct it directly.

## Likely fix

At `src/Core/CodeAnalysis/Emit/TypeDefEmitter.cs`, emit `structSym.Name` for user nested types whose generic rows only reify enclosing parameters. Preserve arity mangling for nested types that declare their own generic parameters and synthesized state-machine/display-class types.

Sibling surface to #2898 / PR #2912, which fixes newly added nested-enum metadata naming without changing pre-existing nested-struct emission.

DavidObando:
## Summary
- mangle user struct/class TypeDefs by parameters declared on that type
- retain enclosing generic parameter rows without counting them in nested names
- cover direct, non-generic-middle, and own-generic shapes through a running C# consumer

## Validation
- `dotnet build`
- 17 targeted nested-generic emit tests
- mutation: reverting the emitter hunk restores CS0426 for all three C# names

Fixes #2916

### DIFF
--- a/src/Core/CodeAnalysis/Emit/TypeDefEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/TypeDefEmitter.cs
@@ -585,12 +585,16 @@ internal sealed class TypeDefEmitter
         FieldDefinitionHandle firstField,
         int methodListRow)
     {
-        // ADR-0087 §3 R1: a generic user type's TypeDef name is mangled with
-        // backtick-arity per ECMA-335 II.10.3.1 (`Box` becomes `Box`1`) and one
-        // GenericParam row is emitted per type parameter immediately after
-        // AddTypeDefinition. Type-erased non-generic structs keep the unmangled
-        // name and have no GenericParam rows.
-        var typeDefName = MangleGenericName(structSym.Name, structSym.TypeParameters);
+        // ADR-0087 §3 R1: the suffix counts parameters declared by this type,
+        // while nested TypeDefs still carry GenericParam rows copied from their
+        // enclosing types. Synthesized types have no declaration, so all of
+        // their TypeParameters are their own.
+        var ownArity = structSym.Declaration == null
+            ? structSym.TypeParameters.Length
+            : structSym.Declaration.TypeParameterList?.Parameters.Count ?? 0;
+        var typeDefName = ownArity == 0
+            ? structSym.Name
+            : structSym.Name + "`" + ownArity.ToString(System.Globalization.CultureInfo.InvariantCulture);
         var handle2 = this.emitCtx.Metadata.AddTypeDefinition(
             attributes: typeAttrs,
             @namespace: structNamespace,
--- a/test/Compiler.Tests/Emit/Issue2916NestedStructArityEmitTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2916NestedStructArityEmitTests.cs
@@ -1,0 +1,183 @@
+// <copyright file="Issue2916NestedStructArityEmitTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection.Metadata;
+using System.Reflection.PortableExecutable;
+using GSharp.Core.CodeAnalysis.Symbols;
+using GSharp.Core.CodeAnalysis.Text;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Xunit;
+using GsCompilation = GSharp.Core.CodeAnalysis.Compilation.Compilation;
+using GsSyntaxTree = GSharp.Core.CodeAnalysis.Syntax.SyntaxTree;
+
+namespace GSharp.Compiler.Tests.Emit;
+
+/// <summary>Regression coverage for issue #2916.</summary>
+public sealed class Issue2916NestedStructArityEmitTests
+{
+    [Fact]
+    public void CSharpConsumer_NamesAndRunsNestedStructsByOwnArity()
+    {
+        var directory = Path.Combine(
+            AppContext.BaseDirectory,
+            nameof(Issue2916NestedStructArityEmitTests) + Guid.NewGuid().ToString("N"));
+        Directory.CreateDirectory(directory);
+        try
+        {
+            var libraryPath = Path.Combine(directory, "Issue2916.Library.dll");
+            var consumerPath = Path.Combine(directory, "Issue2916.Consumer.dll");
+
+            EmitGSharpLibrary(libraryPath);
+            EmitCSharpConsumer(libraryPath, consumerPath);
+            AssertMetadataNames(libraryPath);
+            Assert.Equal("11\n22\n33\n", RunConsumer(directory, consumerPath));
+        }
+        finally
+        {
+            try
+            {
+                Directory.Delete(directory, recursive: true);
+            }
+            catch
+            {
+            }
+        }
+    }
+
+    private static void EmitGSharpLibrary(string outputPath)
+    {
+        const string source = """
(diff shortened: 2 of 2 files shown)


DavidObando:
## Review — PR #3038 (`Fixes #2916`) — **MERGEABLE**

Reviewed at head `819ddd9877fa47a532d0208895ca1b55d3ef3ace`, three-dot from computed merge-base `47a4003 59df8f1` → production `src/Core/CodeAnalysis/Emit/TypeDefEmitter.cs` **+16 / −6** (one hunk), test `+183`. Scratch tree `/tmp/g2964/h`; base for the differential was `428f31779` (main has since moved to `8acc2cc76`; `merge-tree` conflicts = 0 against the new tip, re-checked).

### Base → head differential (measured, not inferred)

Probe library `Outer[T]` (struct) + `COuter[T]` (class) with four nested shapes. TypeDef names read straight out of the PE with `System.Reflection.Metadata`:

| TypeDef | base `428f31779` | head |
|---|---|---|
| `Outer\`1/Inner` | `Inner\`1` ✗ | `Inner` (1 GenericParam) ✓ |
| `Outer\`1/Middle/Deep` | `Middle\`1`, `Deep\`1` ✗ | `Middle`, `Deep` ✓ |
| `Outer\`1/GenericInner` | `GenericInner\`2` ✗ | `GenericInner\`1` (2 GenericParams) ✓ |
| `COuter\`1/CInner` (**class** outer) | `CInner\`1` ✗ | `CInner` ✓ |
| `COuter\`1/CNestedGeneric` (**class** outer) | `CNestedGeneric\`2` ✗ | `CNestedGeneric\`1` ✓ |

C# consumer compiled with `csc` against each library, then **executed** (arity bugs are run-time, not verify-time):

- base: **4 × CS0426** — `Inner`, `Middle`, `GenericInner<>`, `CInner` all "does not exist in the type".
- head: compiles, runs, prints **`11 22 33 44`** — four distinct values, one per arm.

### Mutation

- **M1 — revert the whole hunk** (restore `MangleGenericName(structSym.Name, structSym.TypeParameters)`): build `2034d1a7c1da0384a3366615961f7e2b`, `Issue2916NestedStructArityEmitTests` **FAILS** with `CS0426` on all three shapes. **Hunk is pinned.**
- **M2 — kill only the `Declaration == null` fallback** (`? 0 :` instead of `? structSym.TypeParameters.Length :`): build `c66fa5285c23f455eefba07bbd96d9e9`. **SURVIVES** 91/91 `Compiler.Tests` `~Closure` + 8/8 `Core.Tests` `~Issue2123ConstrainedCallCapturingClosure`. Observable effect: synthesized generic display classes lose their suffix (`P.<>__Box_a_1` with 1 GenericParam instead of `P.<>__Box_a_1\`1`); the program still runs (`7` / `y`) because the CLR does not enforce name/arity agreement at load. Round-trip `76893c53 → c66fa528 → 76893c53` confirms both builds real. See NIT 3.

### Root-cause check — the three sibling `MangleGenericName` call sites

`MangleGenericName` has four call sites, all with the identical `sym.Name, sym.TypeParameters` shape. The PR fixes one. I probed the other three rather than reasoning about them:

- `:751` `EmitNestedStructTypeDef` — documented at `:33` as closure/state-machine only. Those are synthesized (`Declaration == null`), so the PR's own fallback makes it a no-op. **Correctly left alone.**
- `:884` interface — a nested interface in a generic outer emits **0 GenericParam rows**, so `MangleGenericName` yields a bare name and there is no arity mismatch. **Not this bug.** (It has a *different* bug — see SHOULD-FIX 1.)
- `:994` delegate — G# has no nested-delegate syntax at all. `public type H = delegate func(a int32) int32` inside a struct → `GS0005` + `GS0113` + `GS0102`. **Unreachable.**

So "one call site fixed of four" is the correct scope, not a symptom fix. Good.

---

### SHOULD-FIX 1 — nested **interface** in a generic outer is still unusable from C# (`CS0648`)

`src/Core/CodeAnalysis/Emit/TypeDefEmitter.cs:884`

A nested interface inside a generic type gets **0 GenericParam rows**, where ECMA-335 requires a nested type to redeclare its enclosing type's parameters. Measured, identical on base and head:

```
error CS0648: 'Outer<T>.INested' is a type not supported by the language
```

Same family as #2916 (nested types in generic outers), untouched by this PR, and the *only* remaining C# error against my probe library at head. Not a regression — please file it separately rather than expand this PR.

### SHOULD-FIX 2 — G# still cannot consume its own nested-in-generic types across an assembly boundary

Pre-existing, but it bounds what "#2916 is fixed" means. Cross-assembly, from G#:

- nested in a **non-generic** outer → `rc=0`, works.
- nested in a **generic** outer → `GS0158: Cannot find member Value` on both base-built and head-built libraries.

Head is strictly better (base also failed to resolve `Middle` itself; head resolves the type and fails only on members), but the emit side is now correct while the **import** side is not. That is #2912 territory (`UserTokenResolver`, `CollectEnclosingTypeParameters`) — worth a linked issue so #2916 does not read as "closed" for G# consumers.

### NIT 1 — test covers only struct outers

`test/Compiler.Tests/Emit/Issue2916NestedStructArityEmitTests.cs:55-70` declares only `Outer[T]` (a struct). Class outers go through the same `EmitStructTypeDefinition` and I verified `COuter[T].CInner` / `CNestedGeneric[U]` are fixed — but nothing pins them. One extra nested type inside a `class` outer, with a fourth distinct value, would close it.

### NIT 2 — inline mangling duplicates `MangleGenericName`

`:592-597` re-implements the backtick concatenation the helper at `:2204` already does. `ownArity` is the only new information; passing it to the helper would keep one formatter. Cosmetic — the three siblings are provably not affected (above), so there is no correctness argument for consolidating.

### NIT 3 — the `Declaration == null` fallback is unpinned, but is not this PR's debt

M2 survived 99 tests. However the fallback reproduces base behaviour **exactly** (base used `TypeParameters.Length` unconditionally), so relative to `main` it is a no-op — it is the conservative branch, not new speculative code. I am deliberately *not* calling this a BLOCKER despite the "an unpinned hunk is a BLOCKER" rule: the gap is pre-existing coverage of synthesized-type naming, and it is surfaced by the PR rather than created by it.

### Not measured

The #3038 × #2912 four-cell matrix. #2912's head **is** pushed (`7f5609dc18572ec1dfd74c1c6a6f72b9db24c681`), so it is buildable, but #2912 is independently blocked and its subject (`UserTokenResolver` / enclosing-type-parameter population) is the *import* side, which SHOULD-FIX 2 shows is broken on `main` for these shapes both before and after #3038. I flag the interaction as untested rather than claim it clean.

**Verdict: MERGEABLE.** The fix is one hunk, in the right place, pinned by a test that emits, verifies metadata *and* runs a real C# consumer. Everything I found on top of it is pre-existing and adjacent.

PR #3039, opened 2026-08-01
Fix imported generic interface diagnostic display

### LINKED ISSUE #2910: Diagnostics render imported constructed generic types as raw assembly-qualified CLR names

DavidObando:
## Summary
- route all imported CLR interface diagnostic names through shared `SymbolDisplay` across 11 rendering sites
- render imported generic method parameter types in G# syntax, including inherited slots
- render `GS0152` interface/class constraints through shared type display
- cover missing properties, symbolic generic interfaces, constructed parameter types, and constraint messages

## Validation
- `dotnet test test/Compiler.Tests/Compiler.Tests.csproj --filter "FullyQualifiedName~Issue2875DataClassSettableInterfaceTests"` (10/10)
- `dotnet test test/Core.Tests/Core.Tests.csproj --filter "FullyQualifiedName~SymbolDisplayImportedTypeTests|FullyQualifiedName~Issue2851GenericDiagnosticDisplayTests"` (13/13)
- `dotnet test test/Compiler.Tests/Compiler.Tests.csproj --filter "FullyQualifiedName~Issue2916NestedStructArityEmitTests"` (1/1)
- `dotnet build --verbosity:minimal`
- four `GSharp.Core.dll`-verified mutants killed by targeted tests

Fixes #2910

### DIFF
--- a/src/Core/CodeAnalysis/Binding/Binder.cs
+++ b/src/Core/CodeAnalysis/Binding/Binder.cs
@@ -12,6 +12,7 @@ using GSharp.Core.CodeAnalysis.Documentation;
 using GSharp.Core.CodeAnalysis.Lowering;
 using GSharp.Core.CodeAnalysis.Lowering.Async;
 using GSharp.Core.CodeAnalysis.Symbols;
+using GSharp.Core.CodeAnalysis.Symbols.Display;
 using GSharp.Core.CodeAnalysis.Syntax;
 using GSharp.Core.CodeAnalysis.Text;
 
@@ -5734,17 +5735,17 @@ public sealed class Binder
         var flags = new System.Collections.Generic.List<string>();
         if (tp.InterfaceConstraint != null)
         {
-            flags.Add(tp.InterfaceConstraint.Name);
+            flags.Add(SymbolDisplay.ToTypeDisplayString(tp.InterfaceConstraint));
         }
 
         if (tp.ClrInterfaceConstraint != null)
         {
-            flags.Add(tp.ClrInterfaceConstraint.Name);
+            flags.Add(SymbolDisplay.ToTypeDisplayString(tp.ClrInterfaceConstraint));
         }
 
         if (tp.ClassConstraint != null)
         {
-            flags.Add(tp.ClassConstraint.Name);
+            flags.Add(SymbolDisplay.ToTypeDisplayString(tp.ClassConstraint));
         }
 
         if (tp.Constraint == TypeParameterConstraint.Comparable)
--- a/src/Core/CodeAnalysis/Binding/DeclarationBinder.InterfaceVerification.cs
+++ b/src/Core/CodeAnalysis/Binding/DeclarationBinder.InterfaceVerification.cs
@@ -14,6 +14,7 @@ using System.Collections.Immutable;
 using System.Linq;
 using System.Reflection;
 using GSharp.Core.CodeAnalysis.Symbols;
+using GSharp.Core.CodeAnalysis.Symbols.Display;
 using GSharp.Core.CodeAnalysis.Syntax;
 using GSharp.Core.CodeAnalysis.Text;
 
@@ -1624,6 +1625,8 @@ internal sealed partial class DeclarationBinder
                 continue;
             }
 
+            var interfaceName = SymbolDisplay.ToTypeDisplayString(ifaceSym);
+
             // Issue #949: a CLR generic interface closed over a user-defined G#
             // type (e.g. the self-referential `class Shape : IEquatable[Shape]`)
             // is represented with a type-erased ClrType (`IEquatable<object>`)
@@ -1632,7 +1635,12 @@ internal sealed partial class DeclarationBinder
             // contract demands `Equals(Shape)` rather than `Equals(object)`.
             if (MemberLookup.TryGetSymbolicClrGenericInterface(ifaceSym, out var openDefinition, out var symbolicArgs))
             {
-                VerifySymbolicClrGenericInterface(syntax, structSymbol, clrIface, openDefinition, symbolicArgs);
+                VerifySymbolicClrGenericInterface(
+                    syntax,
+                    structSymbol,
+                    openDefinition,
+                    symbolicArgs,
+                    interfaceName);
                 continue;
             }
 
@@ -1662,7 +1670,7 @@ internal sealed partial class DeclarationBinder
                 Diagnostics.ReportInterfaceMethodNotImplemented(
                     syntax.Identifier.Location,
                     structSymbol.Name,
-                    clrIface.FullName ?? clrIface.Name,
+                    interfaceName,
                     FormatClrMethodSignature(clrMethod));
             }
 
@@ -1706,7 +1714,7 @@ internal sealed partial class DeclarationBinder
                     Diagnostics.ReportInterfaceMethodNotImplemented(
                         syntax.Identifier.Location,
                         structSymbol.Name,
-                        clrIface.FullName ?? clrIface.Name,
+                        interfaceName,
                         clrProp.Name);
                     continue;
                 }
@@ -1716,7 +1724,7 @@ internal sealed partial class DeclarationBinder
                     Diagnostics.ReportInterfaceMethodNotImplemented(
                         syntax.Identifier.Location,
                         structSymbol.Name,
-                        clrIface.FullName ?? clrIface.Name,
+                        interfaceName,
                         clrProp.Name + " (getter)");
                 }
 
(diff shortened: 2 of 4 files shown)


DavidObando:
## Review — PR #3039 (`Fixes #2910`) — **MERGEABLE**, with one scope correction

Reviewed at head `fe82a4efeff91f5eea093c52b52b85291c964bc7`, three-dot from computed merge-base `47a400359df8f122f53c50e2f4838834c1055f48` → production `src/Core/CodeAnalysis/Binding/DeclarationBinder.InterfaceVerification.cs` **+39 / −14**, test **+65**. Measured in scratch tree `/tmp/g2964/h` on **`main` = `8acc2cc76`** merged with this head (`merge-tree` conflicts = 0). All line numbers below are in **that merged tree**.

### Correction to the report: "anchors `:1628, :1810`" understates the change

`:1628` and `:1810` are the two *definitions* of the `interfaceName` local. The diff replaces **11 rendering sites** across two methods, plus threads a new parameter through `VerifySymbolicClrGenericInterface` (and correctly drops the now-unused `clrIface`). Worth stating so a reviewer does not check two lines and stop.

It also correctly leaves `:1787` alone — `declaringType?.FullName ?? …` there is an internal `slotKey` dedupe key, not user-visible. Changing it would have been a real bug. Good discrimination.

### Are `:1628` / `:1810` the only display-name producers? — **No.**

Widened past the partial class to all of `src/` with a fixed-string search, first proving the pattern matches the 12 known sites in the target file. `.FullName ?? ` appears **23×** in 10 files. Most are internal (`SignatureEncoder`, `TypeIdentityComparer`, `CustomAttributeEncoder`, `ImportedTypeSymbol` naming). One is a live user-facing leak of exactly the shape #2910 describes — see BLOCKER-adjacent SHOULD-FIX 1.

### Mutation — sentinel-tagged, one build, all 11 sites mapped

Rather than 11 builds, I tagged each of the 5 behavioural groups with a distinct prefix (build `49739014c929dbc5ca8333ffc2513adf`) and read which group produced which diagnostic:

| group | sites | reached by a real compile? | killed by a test? |
|---|---|---|---|
| A — method not implemented | `:1673` | yes — `AAA@GLib2.IGBox[int32].Echo(Int32)` | **yes** |
| B — property missing / getter / setter | `:1717 :1727 :1736` | yes — `BBB@GLib2.IGBox[int32].OnlyGet` | **no** |
| C — setter-kind mismatch (GS0502, the issue's repro) | `:1744` | yes — `CCC@GLib2.IGBox[int32].V` | **yes** |
| D — inherited slot (`declaringType`) | `:1814` | yes — `DDD@GLib2.IBase[int32].BaseEcho(Int32)` | **yes** |
| E — `VerifySymbolicClrGenericInterface` | `:1879 :1891 :1901 :1910 :1918` | yes — `EEE@System.IEquatable[Shape].Equals(T)` | **no** |

Test scope for that claim: `grep -rlF "does not implement interface method" test/` returns exactly 5 files and `GS0502` exactly 3; I ran all of them (62 tests). Only the two tests in this PR's own file failed. Round-trip `ee90d4ca → 49739014 → … → ee90d4ca` confirms every build real. Unmutated, the file is 8/8 green.

**SHOULD-FIX 2**: groups B and E — 6 of the 11 changed sites — survive mutation. I closed the correctness question by direct measurement (both render correctly, shown above), so this is a *regression-protection* gap, not unverified code; that is why it is SHOULD-FIX and not BLOCKER despite the flat "unpinned hunk = BLOCKER" rule. Two assertions fix it: a get-only interface property (group B) and a `class Shape : IEquatable[Shape]` (group E).

---

### SHOULD-FIX 1 — a second site leaks the *verbatim* string from the issue, and the root fix is 3 lines

Issue #2910 quotes `System.…\`1[[System.Int32, System.Private.CoreLib, Version=10.0.0.0, Culture=neutral, PublicKeyToken=…]]`. On this PR's head that exact rendering is still produced by `GS0152`:

```
gsc: error GS0152: Type argument 'string' for type parameter 'T' does not satisfy the
'System.Collections.Generic.List`1[[System.Int32, System.Private.CoreLib, Version=10.0.0.0,
Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]' constraint.
```

I did not stop at asserting a smaller fix — I implemented two and measured both.

- `src/Core/CodeAnalysis/Binding/OverloadResolution.cs:1415` (`constraint.FullName ?? constraint.Name`) looked like the site. Patching it changed **nothing**. Wrong site.
- The real single point is **`src/Core/CodeAnalysis/Binding/Binder.cs:5730 DescribeConstraint`**, which builds the message from `tp.InterfaceConstraint.Name` / `tp.ClrInterfaceConstraint.Name` / `tp.ClassConstraint.Name` — and `ImportedTypeSymbol`'s name *is* `Type.FullName` (`ImportedTypeSymbol.cs:27`). It feeds **9** `ReportTypeArgumentDoesNotSatisfyConstraint` call sites.

Three lines, routing those `.Name`s through `SymbolDisplay.ToTypeDisplayString`, produce:

```
'System.Collections.Generic.List[int32]' constraint.
```

Build `1e32a2eb2a85c4c3a4a7b3436535f3d2`, **95/95** pass on the constraint-message test set. Diff saved at `/tmp/g3039/describeconstraint-root-fix.diff`. Take it here or in a follow-up, but `Fixes #2910` reads as complete while the issue's own quoted string is still reproducible.

### SHOULD-FIX 3 — the new test asserts a remaining leak as expected output

`test/Compiler.Tests/Emit/Issue2875DataClassSettableInterfaceTests.cs` asserts `…IGBox[int32].Echo(Int32)`. The interface half is now G#; the **signature** half is still raw CLR, because `FormatClrMethodSignature` (`DeclarationBinder.InterfaceVerification.cs`, `ps[i].ParameterType.Name`) is untouched. With a constructed-generic parameter it is not merely un-prettified, it is lossy:

```
error GS0187: Class 'Box' does not implement interface method 'GLib2.IGBox[int32].Take(IBase`1)'
```

`IBase[int32]` → **`IBase\`1`**, arity suffix and all, inside the diagnostic this PR exists to de-CLR-ify. Baking `Echo(Int32)` into an assertion makes that the contract. `SymbolDisplay` already has a `Type`-taking renderer (`FormatClrTypeName`, private) that handles this — routing `FormatClrMethodSignature` through it is small.

### NIT — the `:1810` wrapper is a detour

`SymbolDisplay.ToTypeDisplayString(ImportedTypeSymbol.Get(declaringType))` allocates a symbol only to reach `FormatImportedType` → `FormatClrTypeName(imported.ClrType, qualifyNames: true)`, a private static that already takes the `System.Type`. Exposing it (or a `ToTypeDisplayString(Type)` overload) would be shorter and would also be what SHOULD-FIX 3 needs. Behaviourally identical — cached lookup, no correctness issue.

### Orthogonal, per the standing question

`DeclarationBinder.InterfaceVerification.cs:664` (the uncovered `GS0187` indexer arm) is **not** touched by this PR and is unaffected by it — the change is purely message rendering.

**Verdict: MERGEABLE.** The fix is correct at every one of the 11 sites (measured, not assumed), the new test is a real end-to-end one that compiles a G# library and consumes it, and the residue is scope rather than defect. SHOULD-FIX 1 is the one I would not let sit: `Fixes #2910` currently closes an issue whose own repro string is still reproducible via `GS0152`.

PR #3040, opened 2026-08-01
Scope all IL verifier suppressions to affected methods

DavidObando:
## Summary

- Audit every `IlVerifier.Verify` suppression left after #3014. The base had 37 named `ignoredErrorCodes` occurrences: one internal call plus 36 callers; three callers were already scoped, leaving 33 rather than the estimated 26.
- Migrate all 33 remaining callers: remove 19 obsolete blankets and add method scopes to the 14 callers that still need known ilverify limitations suppressed.
- Carry sample-specific scopes through the conformance harness.
- Strengthen the scope regression with `DoesNotContain("::Ignored(")` and document both the method-local `-g` ceiling and the bounded failure caused by starting a drain after `WaitForExit`.
- Fix the Issue1988 test while migrating it: it returned an assembly path after deleting its temp directory, so its caught `assembly not found` exception made the old assertion vacuous. Verification now runs before cleanup, preserves the specific `address of 'object'` check, and scopes the two known pointer categories to `<Program>.run`.

## Scope decisions

- No instance-level verifier matching was added. `-g` remains category-wide inside a matched method; the new `// ponytail:` comment records that ceiling and the exact-instance upgrade path.
- The existing 20-second escaped-descendant assertion remains unchanged. Measured worst case was about 12 seconds, so 20 seconds provides 8 seconds of contention headroom instead of 3.
- No unexpected compiler IL defect was exposed. Every unblanketed failure matched an established static-interface, ref-struct, fixed/pointer, `InitOnly`, or ilverify state-machine limitation.

## Evidence

- Post-rebase targeted run: **158 examined, 158 passed**, exit 0, with `-p:ContinuousIntegrationBuild=true` and a filter covering every changed test class plus the two affected samples.
- Repository audit: **2,884 tracked C# files**, **17** `IlVerifier.Verify` calls still passing `ignoredErrorCodes`, **0** missing `ignoredErrorScope`.
- Perturbation: changed Issue1268 scope from `<Program>\.Use$` to `<NoSuchType>\.Zzz$`; its targeted test failed with the expected `CallAbstract` and `Constrained` errors in `<Program>::Use`. Restored scope; same test passed.
- Three-dot net diff checked from computed merge base `7a6307a797eace2b27854c94b91d64fe7f7b474f`; `git diff --check` passed.

Follow-up to #3014.

### DIFF
--- a/test/Compiler.Tests/Emit/Issue1019StaticInterfacePropertyEmitTests.cs
+++ b/test/Compiler.Tests/Emit/Issue1019StaticInterfacePropertyEmitTests.cs
@@ -184,7 +184,7 @@ public class Issue1019StaticInterfacePropertyEmitTests
             }
             """;
 
-        var output = CompileAndRun(source);
+        var output = CompileAndRun(source, ignoredErrorScope: @"<Program>\.Describe$");
         Assert.Equal("apple\n", output);
     }
 
@@ -293,11 +293,11 @@ public class Issue1019StaticInterfacePropertyEmitTests
             compileExit == 0,
             $"gsc failed:\nstdout:\n{stdoutWriter}\nstderr:\n{stderrWriter}");
 
-        IlVerifier.Verify(outPath, ignoredErrorCodes: IlVerifier.KnownIssues.StaticVirtualInterface);
+        IlVerifier.Verify(outPath);
         return outPath;
     }
 
-    private static string CompileAndRun(string source)
+    private static string CompileAndRun(string source, string ignoredErrorScope = null)
     {
         var tempDir = Directory.CreateTempSubdirectory("gs_sviprop_exe_").FullName;
         try
@@ -335,7 +335,12 @@ public class Issue1019StaticInterfacePropertyEmitTests
                 compileExit == 0,
                 $"gsc failed:\nstdout:\n{stdoutWriter}\nstderr:\n{stderrWriter}");
 
-            IlVerifier.Verify(dllPath, ignoredErrorCodes: IlVerifier.KnownIssues.StaticVirtualInterface);
+            IlVerifier.Verify(
+                dllPath,
+                ignoredErrorCodes: ignoredErrorScope is null
+                    ? null
+                    : IlVerifier.KnownIssues.StaticVirtualInterface,
+                ignoredErrorScope: ignoredErrorScope);
 
             var rtConfig = Path.ChangeExtension(dllPath, ".runtimeconfig.json");
             if (!File.Exists(rtConfig))
--- a/test/Compiler.Tests/Emit/Issue1030InterfaceStaticMembersEmitTests.cs
+++ b/test/Compiler.Tests/Emit/Issue1030InterfaceStaticMembersEmitTests.cs
@@ -164,7 +164,7 @@ public class Issue1030InterfaceStaticMembersEmitTests
             }
             """;
 
-        var output = CompileAndRun(source);
+        var output = CompileAndRun(source, ignoredErrorScope: @"<Program>\.Run$");
         Assert.Equal("3\n3\n", output);
     }
 
@@ -259,7 +259,7 @@ public class Issue1030InterfaceStaticMembersEmitTests
             }
             """;
 
-        var output = CompileAndRun(source);
+        var output = CompileAndRun(source, ignoredErrorScope: @"<Program>\.Describe$");
         Assert.Equal("default-name\nbanana\n", output);
     }
 
@@ -448,11 +448,11 @@ public class Issue1030InterfaceStaticMembersEmitTests
             compileExit == 0,
             $"gsc failed:\nstdout:\n{stdoutWriter}\nstderr:\n{stderrWriter}");
 
-        IlVerifier.Verify(outPath, ignoredErrorCodes: IlVerifier.KnownIssues.StaticVirtualInterface);
+        IlVerifier.Verify(outPath);
         return outPath;
     }
 
-    private static string CompileAndRun(string source)
+    private static string CompileAndRun(string source, string ignoredErrorScope = null)
     {
         var tempDir = Directory.CreateTempSubdirectory("gs_1030_exe_").FullName;
         try
@@ -490,7 +490,12 @@ public class Issue1030InterfaceStaticMembersEmitTests
                 compileExit == 0,
                 $"gsc failed:\nstdout:\n{stdoutWriter}\nstderr:\n{stderrWriter}");
 
-            IlVerifier.Verify(dllPath, ignoredErrorCodes: IlVerifier.KnownIssues.StaticVirtualInterface);
+            IlVerifier.Verify(
+                dllPath,
+                ignoredErrorCodes: ignoredErrorScope is null
+                    ? null
+                    : IlVerifier.KnownIssues.StaticVirtualInterface,
+                ignoredErrorScope: ignoredErrorScope);
 
             var rtConfig = Path.ChangeExtension(dllPath, ".runtimeconfig.json");
             if (!File.Exists(rtConfig))
--- a/test/Compiler.Tests/Emit/Issue1162UserElementArrayEmitTests.cs
+++ b/test/Compiler.Tests/Emit/Issue1162UserElementArrayEmitTests.cs
@@ -262,7 +262,12 @@ public class Issue1162UserElementArrayEmitTests
                 compileExit == 0,
                 $"gsc failed:\nstdout:\n{compileOut}\nstderr:\n{compileErr}");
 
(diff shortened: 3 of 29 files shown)


DavidObando:
## Review: MERGEABLE — no blockers

Reviewed the merged tree (`main` `8acc2cc76` + PR `69b887f61` = `9a08cee3b`), not the PR head. Merge base computed against a freshly fetched remote ref: `7a6307a79`; true three-dot scope **29 files / +196 −154**, all under `test/Compiler.Tests/`. Merged build with `-p:ContinuousIntegrationBuild=true`: exit 0, **0 warnings / 0 errors**. `GSHARP_SKIP_ILVERIFY` confirmed unset before any IL measurement.

The headline claim — that the suppression migration is real rather than a blanket relocated behind better paperwork — **holds under perturbation**.

### 1. The migration is load-bearing (the main thing I was asked to disprove, and could not)

Three independent lines of evidence:

**Structural.** All 27 scope regex literals are `$`-anchored and enumerate explicit method names. **0 wildcards, 0 unanchored.** The broadest is a 7-method alternation (`ArrayPin|FixedArrayPin|EmptyPin|StringPin|SpanPin|TryInsideFixed|FixedInsideTry`) plus one state-machine form (`<(Values|AfterFixed|ReturnFromFixed)>d__\d+\.MoveNext$`). No scope is broad enough to re-blanket.

**Empirical round trip.** I replaced **all 27** scope literals with a deliberately non-matching `@"<NoSuchTypeXyz>\.Nope$"` across 16 files:

| state | scoped-class filter | sample conformance |
|---|---|---|
| scopes broken | **41 failed / 114 total** | **2 failed / 123 total** |
| scopes restored | **0 failed / 114 total** | **0 failed / 123 total** |

**16 of 16** scope-bearing files produced at least one failure — every scoped test class (Issue2945, Issue1268, Issue2370, Issue1030, UserStructMethod, Issue2906, Issue1522, Issue2900, Issue1988, StaticVirtualInterfaceMembers, Issue1162, Issue903, Issue2922, Issue1019, IlVerifierTests) plus both `SampleKnownIssues` entries. There is no scoped site whose scope is inert.

**Why a vacuous scope cannot go silently permissive.** This is the property that makes the design safe: when a scope matches nothing, run 1 verifies the *whole assembly with no ignores at all* and fails loudly. A mis-typed regex is a hard failure, never a silent pass. That is the correct failure direction and it is what the perturbation above measures.

**No "strict" conversion quietly kept an ignore list.** By construction: 17 sites pass `ignoredErrorCodes`, and **all 17** also pass a scope. The 19 strict conversions pass no codes whatsoever. Two sites make this conditional and get it right — `Issue1162…:265` and `Issue2922…:329` supply the scope *precisely when* the code list is non-empty and `null` otherwise.

**On "no new IL defects found."** I pushed on this because 33/33 clean reads as too good. The evidence is actually stronger than the phrasing suggests: **19 of 33 blankets were deleted outright**, and those sites now verify with *no* suppression and pass. That is not a defect being hidden — it is 19 suppressions proven unnecessary. The remaining 14 each carry a documented known-ilverify-limitation rationale. A 19-unnecessary / 14-necessary split is a credible outcome, and the 19 are the strongest available evidence that nothing was swept up.

### 2. The counts reconcile exactly, on a stated denominator

Counted with a balanced-paren parser rather than a line grep, so parameter *declarations* are not conflated with call *sites*:

| population | base `8acc2cc76` | merged |
|---|---|---|
| `IlVerifier.Verify` invocations **examined** | **796** | **797** |
| …passing `ignoredErrorCodes` | 36 | 17 |
| …of those, also scoped | 3 | **17** |
| …**unscoped** | **33** | **0** |

33 = 19 strict + 14 newly scoped. 17 = 14 newly scoped + 3 pre-existing. The PR body's "37 occurrences: one internal plus 36 callers, three already scoped, leaving 33" matches this exactly. (An earlier "29/3" figure quoted during #3014 was measured against the older `f70626452` tree — a different population, which is the whole reason both sides must name their denominator.)

### 3. The Issue1988 vacuity fix is real, declared, and understated

Declared in the PR body, so not undeclared scope. And the old test was worse than "vacuous" — it verified **no IL at all**. I dumped the string the old assertion inspected, on base:

```
ilverify: assembly not found at '/var/.../gs_issue1988_Dp7ipk/test.dll'
```

`CompileAndRun`'s `finally` deleted the temp directory before `RunIlVerifyRaw` ran, so `Verify` threw *assembly not found* and `Assert.DoesNotContain("address of 'object'", …)` passed against an error message that could never contain the signature. The fix moves `IlVerifier.Verify(outPath)` inside `CompileAndRun` ahead of cleanup.

Liveness confirmed rather than assumed: flipping `DoesNotContain` → `Contains` at `Issue1988NarrowedStructAddressOfEmitTests.cs:100` **fails**, proving the catch block is entered with a non-empty message that genuinely lacks the signature. Round trip on the test assembly: `9a80aa70` → `d406ecfd` (FAIL) → `9a80aa70` (PASS). (My first probe — `Assert.False(string.IsNullOrEmpty(stdout))` — passed and disproved my own "empty string" hypothesis; the real cause only appeared once I dumped the text.)

### 4. All four carry-over items landed

- ✅ `IlVerifierTests.cs:186` — `Assert.DoesNotContain("::Ignored(", exception.Message, …)`.
- ✅ `IlVerifier.cs:107-108` — the ceiling is named correctly and does not over-claim: *"-g remains category-wide inside the matched method; parse exact verifier instances only if method scope proves too broad."* Blast radius went assembly → method, **not** instance-level, and the comment says so.
- ✅ `IlVerifierTests.cs:224` — second clause present.
- ✅ **The 20 s bound at `IlVerifierTests.cs:281` is justified, and the justification is correct.** I initially had this down as unjustified; that was my error. The bound is *structural*, not a fudge. The fixture spawns the escaping child as `sleep 30 **2>/dev/null**`, so the reparented process inherits **only** stdout. `Grab(stderrTask)` therefore returns as soon as `/bin/sh` dies, and only `Grab(stdoutTask)` burns the full `KillGraceMilliseconds`. Worst case = 2 s timeout + 10 s grace = **12 s**, not the 22 s that two blocked pipes would cost. Measured: the isolated test reports **Duration: 12 s**, exit 0. 20 s leaves 8 s of contention headroom over a bound that cannot drift.

### 5. No duplicate drain infrastructure survived #3021

`KillGraceMilliseconds` occurs exactly **once** in the test project, `IlVerifier.cs:48` = `10_000`. No `5_000` duplicate. The other `5_000` hits are post-`Kill` child-exit grace in unrelated emit fixtures, and the `Drain(` hits are G# fixture source — an async iterator *named* `Drain` — not process-drain infra.

---

## NITs

**N1 — `IlVerifier.cs:107`: `// ponytail:` leaks an internal agent-tooling name into committed source.** It is declared in the PR body, so deliberate, not accidental. The comment's content is good; the prefix is meaningless to a future reader of this repository. Suggest dropping the two-word prefix and keeping the sentence.

**N2 — ignore-list minimality is untested (residual, already documented).** The perturbation proves each *scope* is load-bearing, and that the assembly genuinely violates the suppressed categories. It does **not** prove each individual error code within a list is necessary. Since `-g` stays category-wide inside the matched method, a superfluous code re-widens the hole within that one method. The exposure is small in practice — bundles are 1–2 codes, centralised and documented in `KnownIssues` — so this is a note on the ceiling the PR already names, not a defect. The instrument, if it ever matters: drop one code at a time from a scoped site and see whether the site still passes.

## Boundary of this evidence

Stated so it is not over-read later: breaking a scope makes run 1 verify the whole assembly unsuppressed, so a resulting failure proves *the assembly* violates a suppressed category and that the site's suppression is doing work. It does not by itself prove the violation lives in the named method — that comes from the failure messages, which name the methods. And per N2, code-level minimality within each list remains unmeasured.

**Verdict: MERGEABLE.** No blockers, no should-fixes. The un-blanketing is genuine, the counts are honest, and the Issue1988 catch was a real latent hole that had been passing green while verifying nothing.

DavidObando:
## Validation addendum: Sdk.Tests

Correction to my earlier SDK-test exclusion: `Sdk.Tests` is in-bounds. After deleting ignored build output and performing a solution-level restore with `RestorePackagesPath` redirected inside the worktree, PR head `69b887f611` passed **57/57**, exit 0. Current `main` `61fe9cb943` passed **59/59**, exit 0. No e2e test ran and neither validated run wrote to `~/.nuget/packages/gsharp.net.sdk/`.

Near-miss: a clean project-scoped `dotnet test test/Sdk.Tests/Sdk.Tests.csproj` is not self-restoring. It failed before tests with `NETSDK1004` for `Gsharp.Extensions`; restoring that project alone then exposed the same missing-assets failure for `Gsgen.Cli`. A solution-level restore is required before the project-scoped test on a clean output tree.

PR #3041, opened 2026-08-01
Cover async struct channel select integration

DavidObando:
## Summary

PR #3026 fixed symbolic channel element metadata, while #2967 independently changed async select receive bindings. This test pins their combined path: an async select receives a same-compilation struct, awaits inside the arm, then reads the hoisted binding.

## Evidence

`dotnet test test/Compiler.Tests/Compiler.Tests.csproj -c Release --no-restore --nologo --verbosity minimal -p:ContinuousIntegrationBuild=true --filter FullyQualifiedName~Issue2965ChannelElementSlotTests`

Result: **2/2 passed** with assembly loading, bounded execution, exact output, and IL verification.

Test-only; no production changes.

Related to #2965 and #2967.

### DIFF
--- a/test/Compiler.Tests/Emit/Issue2965ChannelElementSlotTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2965ChannelElementSlotTests.cs
@@ -90,6 +90,19 @@ public class Issue2965ChannelElementSlotTests
                 return (<-ch).Value
             }
 
+            async func asyncSelectPair() int32 {
+                let ch = make(chan Pair, 1)
+                ch <- Pair(56)
+                var result = 0
+                select {
+                    case let value = <-ch {
+                        await Task.Delay(1)
+                        result = value.Value
+                    }
+                }
+                return result
+            }
+
             let plainReceive = make(chan Pair, 1)
             plainReceive <- Pair(41)
             Console.WriteLine((<-plainReceive).Value)
@@ -177,12 +190,13 @@ public class Issue2965ChannelElementSlotTests
 
             Console.WriteLine(echo(Pair(53)).Value)
             Console.WriteLine(asyncPair().Result)
+            Console.WriteLine(asyncSelectPair().Result)
             """;
 
         AssertRuns(
             Source,
             nameof(ChannelElementMatrix_LoadsVerifiesAndRuns),
-            "41\n42\n55\n43\n44\n45\n46\n47\n54\n48\n49\n2020\n0\n0\n51\n52\n53\n50\n");
+            "41\n42\n55\n43\n44\n45\n46\n47\n54\n48\n49\n2020\n0\n0\n51\n52\n53\n50\n56\n");
     }
 
     private static void AssertRuns(string source, string name, string expected)


DavidObando:
## Review: **MERGEABLE.** Complementary to #3026, not redundant — and I checked by trying to prove the opposite.

Three-dot from computed merge base `428f31779` → head `f3e62d0c1`: **1 file, +15, −1.** Test-only, as the body says.

### Anchors — measures against the right `main`

Merge base **is** `428f31779`, the #3026 merge commit, and `git merge-base --is-ancestor a8ee8f22c 428f31779` returns true, so **#2967 is an ancestor**. This is post-#2967 and post-#3026, not something older. Baseline on the merge with `main` `8acc2cc76`: `Issue2965ChannelElementSlotTests` **2/2 passed** — the body's claim reproduces exactly.

`UNSTABLE` is the **`build` job pending**, not failing; every other check passes. Nothing to fix.

### Redundant, complementary, or contradictory?

**Mutant M1** — delete the hoisted-field store #2967 added, `MethodBodyEmitter.Statements.cs` `EmitSelectReceiveArmBody`, restoring the pre-#2967 defect where the arm binding is hoisted but never written. md5 `68f7a7ee…` → `878b09d5…`, build exit 0.

Result over `Issue2965ChannelElementSlotTests` + `Issue2933AsyncSelectArmBindingTests`: **Failed 10, Passed 3, Total 13.**

| test | kills M1 |
|---|---|
| `ChannelElementMatrix_LoadsVerifiesAndRuns` (this PR) | yes |
| `RequestedSelectShapesLoadAndRun` (×7, #2967) | yes |
| `AsyncAndSynchronousReceiveBindingsAreEqualAtRuntime` (#2967) | yes |
| `AsyncIteratorReceiveBindingsMatchAsyncFunctionAtRuntime` (#2967) | yes |

**For M1 specifically this PR is the tenth kill of an already well-pinned mutant.** On that evidence alone I was ready to call it redundant. It is not, because the new case is distinguished on two axes that nothing else in the repo combines:

1. **Same-compilation struct element.** `Issue2933AsyncSelectArmBindingTests` declares **zero** same-compilation structs — its channel elements are `string?`, `int32`, and `DateTime`. `DateTime` is an *imported* CLR struct, so it never exercises the symbolic element metadata #3026 shipped.
2. **`await` inside the arm, binding read after it.** The pre-existing `select` cases in this very file (`:111-129`, `:169-187`) are all synchronous top-level or `scope`-based — no state machine, so no hoisting. `asyncPair` (`:86`) is async but has no `select`.

So `asyncSelectPair` is the only place where a **same-compilation struct binding must survive a suspension point inside a select arm** — precisely the #3026 × #2967 intersection the body claims. It contradicts nothing #3026 shipped; it extends it.

Distinct value **`56`** against the matrix's existing `41…53`, and appended as `"…50\n56\n"` — the arm is observable, so a wrong-dispatch mutant cannot hide behind an identical value.

Restored, rebuilt, md5 back to `68f7a7ee…`, `git status --porcelain` empty.

### Gates at `main` = `8acc2cc76`

Merge clean, 0 conflicts, tip `9f055d775`; build **exit 0**; **40 `GSharp.Core.dll` examined / 0 zero-byte**.

### Verdict: **MERGEABLE**

- **SHOULD-FIX** — none blocking, but the unique value rests on the *symbolic-struct* axis, and I proved that axis structurally (no other test has the shape) rather than by a mutant only this case kills. A mutant on #3026's symbolic element path that this case alone kills would convert the argument from structural to measured. Worth one line in the body noting `DateTime` in #2933 is an imported struct, so this is the first *same-compilation* one.
- **NIT** — the body says "Result: **2/2 passed**", which reproduces, but 2 of those 2 are the whole file; only one test actually changed. Saying which test carries the new case would make the evidence line self-contained.

PR #3042, opened 2026-08-01
Fix generic local function evaluation in gsi

### LINKED ISSUE #3030: gsi: verify local function declaration evaluation
## Problem

`BoundLocalFunctionDeclaration` has an emitter dispatch in `src/Core/CodeAnalysis/Emit/MethodBodyEmitter.cs:381`, but no references in any `Evaluator*.cs` file. Interpreter reachability is unverified.

Measured at `f70626452`:

- `MethodBodyEmitter.cs`: 1 `BoundLocalFunctionDeclaration` reference
- `Evaluator*.cs`: 0 references across the evaluator population

This is adjacent to #3003 but is not fixed by #3012. Function-literal body lowering does not add local-function declaration dispatch.

## Acceptance criteria

- establish whether source syntax can make this bound node reach the interpreter
- if reachable, add evaluator handling and a regression test
- if unreachable by design, document or remove the dead evaluator-facing shape

DavidObando:
## Summary
- handle generic local-function declaration statements in evaluator
- register their lowered bodies for direct generic calls
- cover top-level and nested declarations with distinct outputs

Fixes #3030

## Validation
- `dotnet build --nologo`
- `dotnet test test/Interpreter.Tests/Interpreter.Tests.csproj --filter FullyQualifiedName~Issue3030LocalFunctionDeclarationInterpreterTests --nologo`
- declaration dispatch, body registration, and call fallback mutants each fail 2/2 cases

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -86,7 +86,9 @@ public sealed partial class Evaluator
                 }
             }
 
-            var statement = program.Functions[node.Function];
+            var statement = program.Functions.TryGetValue(node.Function, out var functionBody)
+                ? functionBody
+                : localFunctionBodies[node.Function];
             var isIterator = IsIteratorFunction(node.Function, statement);
             object result;
             using (PushFrame(locals))
--- a/src/Core/CodeAnalysis/Evaluator.Statements.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Statements.cs
@@ -71,6 +71,14 @@ public sealed partial class Evaluator
                     EvaluateVariableDeclaration((BoundVariableDeclaration)s);
                     index++;
                     break;
+                case BoundNodeKind.LocalFunctionDeclaration:
+                    var declaration = (BoundLocalFunctionDeclaration)s;
+                    var literal = declaration.Literal;
+                    localFunctionBodies.TryAdd(
+                        literal.Function,
+                        literal.LoweredBody ??= (BoundBlockStatement)Lowering.Lowerer.Lower(literal.Body));
+                    index++;
+                    break;
                 case BoundNodeKind.ExpressionStatement:
                     EvaluateExpressionStatement((BoundExpressionStatement)s);
                     index++;
--- a/src/Core/CodeAnalysis/Evaluator.cs
+++ b/src/Core/CodeAnalysis/Evaluator.cs
@@ -59,6 +59,7 @@ public sealed partial class Evaluator
     private readonly object globalsLock = new object();
 
     private readonly Dictionary<Symbols.FunctionSymbol, bool> iteratorFunctionCache = [];
+    private readonly ConcurrentDictionary<FunctionSymbol, BoundBlockStatement> localFunctionBodies = [];
     private readonly Dictionary<(Symbols.StructSymbol, Symbols.FieldSymbol), object> staticFields = [];
 
     // ADR-0089 / issue #1030: interface static-field storage. Keyed by the
--- a/test/Interpreter.Tests/Issue3030LocalFunctionDeclarationInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue3030LocalFunctionDeclarationInterpreterTests.cs
@@ -1,0 +1,73 @@
+// <copyright file="Issue3030LocalFunctionDeclarationInterpreterTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using GSharp.Core.CodeAnalysis.Compilation;
+using GSharp.Core.CodeAnalysis.Symbols;
+using GSharp.Core.CodeAnalysis.Syntax;
+using Xunit;
+
+namespace GSharp.Interpreter.Tests;
+
+/// <summary>
+/// Issue #3030: generic local-function declarations are evaluator no-ops.
+/// </summary>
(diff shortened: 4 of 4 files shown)


DavidObando:
## Verdict: **MERGEABLE.** No blockers. Two tests for three hunks turns out to be the right number.

Reviewed at head `be6386f2f`, merge base **`7a6307a79`** (computed against a freshly fetched remote ref), `main` = `8acc2cc76`. Three-dot: **4 files, +85/−1** — 13 lines of production across 3 hunks, plus a new 73-line test file.

`merge-tree` into `8acc2cc76` **rc=0**; full solution build of the merged tree **rc=0**, `0 Error(s)`.

### The "2 tests / 3 hunks" question — resolved, no masking

Hunk 1 (`Evaluator.cs:62`, the field) is **compile-forced**: it cannot be reverted independently, so it is not separately mutable. That leaves two mutable hunks, and each is independently load-bearing:

| mutant | result | failure message |
|---|---|---|
| **R1** — remove the `LocalFunctionDeclaration` case (`Evaluator.Statements.cs:74-81`) | **KILLED 2/2** | `Unexpected node LocalFunctionDeclaration` |
| **R2** — revert the fallback to `program.Functions[node.Function]` (`Evaluator.Calls.cs:89`) | **KILLED 2/2** | `The given key 'func Identity(value T) T' was not present in the dictionary.` |

The failure messages are **distinct**, which is the specific thing worth checking: neither hunk masks the other, unlike #3028 where a sibling guard fired first. R2's message also proves the fallback branch is genuinely **live** — `program.Functions` really does not contain the local function, so the `: localFunctionBodies[...]` arm is not dead code.

Round trip: pass → R2 fail → R1 fail → pass, rebuilt each time; `git status --porcelain` empty throughout.

### `gsc` as oracle — agrees on every probe

| probe | gsi | gsc | |
|---|---|---|---|
| top-level generic local fn, 2 instantiations | `17 / top-level` | `17 / top-level` | identical |
| generic local fn declared **inside a loop** (`TryAdd` hit 3x) | `0 / 1 / 2 / end` | `0 / 1 / 2 / end` | identical, 4 lines compared |

Repeated declaration is idempotent — `TryAdd` keeps the first body and the same symbol maps to the same lowering.

**A divergence I found and then withdrew.** My first loop probe showed gsi `0/1/2/3` vs gsc `0/1/2`. It is not a loop bug and not this PR: `gsi` echoes the program's trailing `LastValue`. Proof — `echo1.gs` ending in `Console.WriteLine("done")` prints `done`; `echo2.gs` ending in `i = i + 5` prints `done` then `5`. Reproduced identically on a **pre-#3042** build (`4a9868061`). Recording it because comparing a value-echoing REPL against a compiled binary is a false-positive generator; oracle probes must end in an explicit write.

### Scope — correctly narrow, and the neighbouring lookup sites are safe

The fix is **generic** local functions only. Non-generic `let Twice = func(x int) int {...}` prints `42` unchanged — it binds as a lambda through the pre-existing `ClosureValue` path and never reaches the new case.

I enumerated all `program.Functions` lookups (9 in `Evaluator.Calls.cs`, 10 in `Evaluator.Expressions.cs`) to check whether a fallback added at one site leaves siblings exposed. Two use the identical `node.Function` shape:

- `Evaluator.Expressions.cs:795` — `EvaluateMethodGroupExpression`, indirect invocation by name. **Not reachable**: referencing a generic local function as a value is rejected by the binder — `GS0126: 'Id' is not a variable` (measured, both `Apply(Id, 5)` and `let g = Id[int]`).
- `Evaluator.Expressions.cs:1569` — binary-operator dispatch. Structurally unreachable for a local function.

Forward references are likewise rejected by the **binder** (`GS0130: Function 'Identity' doesn't exist.`), so the `localFunctionBodies[...]` indexer cannot be reached before registration.

`ConcurrentDictionary` looked inconsistent next to four plain `Dictionary` siblings, so I checked before flagging it: the evaluator is genuinely concurrent (`Task.Run` goroutines, ADR-0022) and `ConcurrentDictionary` is established precedent for frames (#1718). It is justified. `literal.LoweredBody ??= …` is a verbatim copy of the existing idiom at `Evaluator.Expressions.cs:782`. No new abstraction — ponytail-clean.

### Contested file — checked by building the combination, not by assuming disjointness

`Evaluator.Calls.cs` is touched by #3029, #3028 and #3032 (#2990's ref is gone). Hunk offsets say #3029 is the only real neighbour: it replaces 21 lines at `:66` and shares this PR's exact merge base, and it **keeps** `program.Functions[node.Function]` while #3042 replaces it. I merged `main` + #3042 + #3029 for real: **both merges rc=0, no conflict**, and #3042's fallback survives intact inside #3029's restructure at `:76-78`. #3028 (`:1107-1121`) and #3032 (`:1272-1383`) are in different methods.

### Other gates

- **Interpreter.Tests: 514 passed / 0 failed.**
- **No new `DiagnosticDescriptor`** — so the now-mandatory docs gate is not engaged. (Verified with a pattern proven live: descriptors use target-typed `new("GS…")`, and the pattern counts **376** on `main`, matching the known population. My first pattern, `new DiagnosticDescriptor`, matched zero anywhere — an unrun check.)
- No undeclared scope: production is 13 lines in 3 hunks, the test file is new (`A`), no existing test was edited.

## NIT — no change requested now

`Evaluator.Expressions.cs:795` and `Evaluator.Calls.cs:556` look up `program.Functions` without the local-function fallback. Both are unreachable today **only because the binder refuses to treat a generic local function as a value** (`GS0126`). If that restriction is ever relaxed — passing one as a delegate is a natural request — `:795` becomes a live `KeyNotFoundException` site. Worth a comment at `Evaluator.Calls.cs:89` recording that the fallback is deliberately local to the direct-call path.

PR #3043, opened 2026-08-01
Keep canonical and website diagnostic references synchronized

### LINKED ISSUE #2970: Keep canonical and website diagnostic references synchronized

DavidObando:
## Summary

- Fill every current website-reference gap, including GS0513 added while this branch was rebased.
- Align website severities and explicitly mark GS0217 and GS0294 as retired.
- Add retired GS0294 to the canonical reference.
- Extend `DiagnosticIdUniquenessTests` to require exact ID and severity maps in both references, including individually documented direct-emitted, retired, and reserved IDs.

This takes option (a): both current references are complete. No undocumented-debt allow-list or follow-up issue is needed. `website/versioned_docs/version-0.3/` is untouched.

Closes #2970

## Verification

- Rebases onto `fd34166a82c2`, which contains requested `a6b1561f2`.
- `dotnet build GSharp.sln -c Release -m:1 -warnaserror --nologo`: 0 warnings, 0 errors.
- Four propagated `GSharp.Core.dll` copies were non-empty and matched md5 `20edaeabfcdd22e33064d20c44df4abd` before and after the targeted test.
- `dotnet test test/Core.Tests/Core.Tests.csproj -c Release -m:1 --nologo`: 7033 passed, 1 skipped, 0 failed.
- `npm run build` in `website/`: successful Docusaurus production build.

### Pins and guards

- Pins: 1. Applying the changed test to base `fd34166a82c2` fails on missing GS0294; HEAD passes.
- New guards: 0.

### Mutation proof

Unmutated control passed before and after every probe.

| Mutation | Detection |
|---|---|
| Remove GS0506 from canonical reference | Fails: canonical missing GS0506 |
| Remove GS0505 from website reference | Fails: website missing GS0505 |
| Change website GS0166 Warning to Error | Fails: severity mismatch |
| Add descriptor GS9876 without documentation | Fresh Core rebuild; fails: canonical missing GS9876 |

### DIFF
--- a/docs/diagnostics.md
+++ b/docs/diagnostics.md
@@ -910,6 +910,7 @@ nesting depth.
 | GS0283 | Error | No sibling initializer overload matches an `init(args)` self-delegation call. |
 | GS0284 | Error | An explicit `init(...)` duplicates the constructor synthesized from the class's primary-constructor parameters. |
 | GS0293 | Error | A labeled `break` or `continue` names no enclosing loop. |
+| GS0294 | _Retired_ | Previously: labels could only be applied to loop statements. Non-loop labels are now valid `goto` targets (ADR-0139); this diagnostic is no longer emitted. |
 | GS0295 | Warning | A loop label shadows an enclosing loop label of the same name. |
 | GS0296 | Error | The right-hand side of an `if let` or `guard let` binding is not nullable. |
 | GS0297 | Error | A `guard let` else block can fall through instead of unconditionally exiting with `return`, `throw`, `break`, or `continue`. |
--- a/test/Core.Tests/CodeAnalysis/DiagnosticIdUniquenessTests.cs
+++ b/test/Core.Tests/CodeAnalysis/DiagnosticIdUniquenessTests.cs
@@ -201,6 +201,7 @@ public class DiagnosticIdUniquenessTests
         string path,
         IReadOnlyDictionary<string, string> expectedSeverities)
     {
+        var referencePath = Path.GetRelativePath(FindRepoRoot(), path);
         var documentedSeverityGroups = Regex.Matches(
                 File.ReadAllText(path),
                 @"^\|\s*(GS\d{4})\s*\|\s*([^|]+?)\s*\|",
@@ -221,7 +222,7 @@ public class DiagnosticIdUniquenessTests
 
         Assert.True(
             conflictingSeverities.Length == 0,
-            $"{path} documents diagnostics with conflicting severities:\n" +
+            $"{referencePath} documents diagnostics with conflicting severities:\n" +
             string.Join("\n", conflictingSeverities));
 
         var documentedSeverities = documentedSeverityGroups
@@ -246,10 +247,10 @@ public class DiagnosticIdUniquenessTests
 
         Assert.True(
             missing.Length == 0 && unexpected.Length == 0 && mismatched.Length == 0,
-            $"{path} is out of sync with compiler diagnostics:\n" +
-            $"Missing: {string.Join(", ", missing)}\n" +
-            $"Unexpected: {string.Join(", ", unexpected)}\n" +
-            $"Severity mismatches:\n{string.Join("\n", mismatched)}");
+            $"{referencePath} is out of sync with compiler diagnostics:\n" +
+            $"Missing from {referencePath}: {string.Join(", ", missing)}\n" +
+            $"Unexpected in {referencePath}: {string.Join(", ", unexpected)}\n" +
+            $"Severity mismatches in {referencePath}:\n{string.Join("\n", mismatched)}");
     }
 
     private static string NormalizeDocumentedSeverity(Match match)
--- a/test/Core.Tests/CodeAnalysis/DiagnosticIdUniquenessTests.cs
+++ b/test/Core.Tests/CodeAnalysis/DiagnosticIdUniquenessTests.cs
@@ -142,56 +142,53 @@ public class DiagnosticIdUniquenessTests
     }
 
     [Fact]
-    public void Every_Descriptor_Is_Documented_With_Matching_Severity()
+    public void Every_Diagnostic_Is_Documented_In_Both_References_With_Matching_Severity()
     {
-        var documentation = File.ReadAllText(Path.Combine(FindRepoRoot(), "docs", "diagnostics.md"));
-        var documentedSeverityGroups = Regex.Matches(
-                documentation,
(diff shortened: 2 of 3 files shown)


DavidObando:
# Review — PR #3043 @ `6b6f5bde`. **MERGEABLE.**

Anchors: `main` = `8acc2cc76`; computed merge base `git merge-base origin/pr3043 origin/main` = `428f31779`. True content `git diff --stat 428f31779...6b6f5bde` = **3 files, +183/−34**. All findings below are measured in the **merged tree** `c43b27adf` (= `8acc2cc76` + this PR), which merged clean.

## Scope is much wider than the branch name

`fix/2959-document-gs0506` is stale. **GS0506 was already documented in `main`** — 2 hits in `docs/diagnostics.md` and 2 in the website reference at `8acc2cc76`; this PR does not add it. The actual change closes **#2970**: make the docs gate **total and bidirectional**, extend it to a **second file**, and backfill 83 website rows. Reviewers going by the branch name will under-weight it.

## `UNSTABLE` explained — it is not the docs gate

```
build                     pass   33m23s
Build docs                pass   30s
vsix / vscode ×3 / vs-ext pass
cs2gs                     PENDING
e2e                       PENDING
```
`UNSTABLE` is **two pending jobs**, not a failure. No check is red.

## The gate is real — 5 mutations, all caught, with restores

The interesting question is whether the new gate actually binds. It does. Run against `Every_Diagnostic_Is_Documented_In_Both_References_With_Matching_Severity` in the merged tree, restoring between each:

| # | mutation | result |
|---|---|---|
| M1 | delete `GS0506` row from `docs/diagnostics.md` | **FAIL** — `Missing: GS0506` |
| M2 | restore | PASS |
| M3 | **backtick the ID**: `` \| `GS0506` \| Error \| `` | **FAIL** — `Missing: GS0506` |
| M4 | restore | PASS |
| M5 | `GS0506` severity `Error`→`Warning` | **FAIL** — `expected Error, found Warning` |
| M6 | add orphan row `GS0999` | **FAIL** — `Unexpected: GS0999` |
| M7 | delete `GS0506` from **website** only | **FAIL** — `Missing: GS0506` |
| M8 | restore all | PASS |

**M3 is the important one.** The backticked-ID form renders correctly and reviews as correct, and under the old gate it silently reduced to "undocumented" and was skipped. It is now caught. M6 shows the new `unexpected` direction works, and M7 shows website enforcement is live.

Unmutated control: `dotnet test --filter DiagnosticIdUniquenessTests` → **4 passed, 0 failed** before and after the whole sequence.

## Populations, not bare zeros

Independent static reconstruction of the gate on the merged tree:

- descriptor fields parsed from `DiagnosticDescriptors.cs`: **376**
- hardcoded non-descriptor allowlist: **7** (`GS0190 GS0217 GS0294 GS0419 GS9100 GS9998 GS9999`)
- expected total: **383**
- `docs/diagnostics.md`: **385** gate-visible rows → **383** distinct IDs, **0** severity conflicts
- `website/docs/ref/diagnostics.md`: **383** rows → **383** distinct
- both files: **missing 0, unexpected 0, mismatched 0**

The 376 figure matches the independently measured count, and the real xUnit run agrees with the reconstruction.

## No row is orphaned

The website diff removes 4 `|`-rows, which is the fatal shape. They are **severity corrections, not deletions** — GS0166 `Error`→`Warning`, GS0217 `Error`→`_Retired_`, GS0294 `Error`→`_Retired_`, GS0396 `—`→`Error`. Each of the four still has **exactly 1** gate-visible row at head. Nothing orphaned; the `unexpected` check would have caught it anyway.

## No ID allocated

`git diff 428f31779...6b6f5bde | grep -E "GS(0507|0509|0510|0511|0512|0513|0514|0515)"` → **0 hits over 302 diff lines**. Positive control: the same pattern matches a synthetic `| GS0514 | Error |` row, so this is a run check, not an empty one. The reserved range is untouched.

`GS0294` is added as a **retired** row with no descriptor, and is correspondingly listed in the allowlist. That is a documentation of a dead ID, not an allocation.

## SHOULD-FIX

**S1 — this makes the website reference mandatory for every future descriptor, and in-flight PRs cannot know that.** `website/docs/ref/diagnostics.md` was previously enforced by nothing; after this lands, `AssertDocumentationMatches` runs against it too (`DiagnosticIdUniquenessTests.cs`, second call). Every open PR that adds a descriptor was authored against a single-file gate, will be **green on its branch and red on `main`** — the exact hazard this PR is fixing, inverted onto everyone else. With `GS0509`–`GS0515` already allocated to in-flight work, that is several PRs. This is a coordination item, not a code defect: either land this first and tell those authors to add the second row, or accept a short window of post-merge failures.

## NIT

**N1** — the PR body's verification used `dotnet test … --no-build`, so its reported `7033 passed` describes whatever binaries happened to be on disk. I re-ran the gate without `--no-build` and it passes, so the conclusion holds; only the evidence was weak.

**N2** — the allowlist merge uses `expectedSeverities.Add(item.Key, item.Value)`. If a real descriptor is ever declared with an allowlisted ID (e.g. someone revives `GS0217`), this throws `ArgumentException` instead of producing the intended assertion message. It still fails, so it is not a hole — just an unhelpful failure mode. `[key] =` plus an explicit collision assert would read better.

**Not a defect (checked and withdrawn):** the new `GS0294` website row has 3 columns while `GS0217` has 4. Its actual neighbour `GS0295` is also 3 columns — that table section is 3-column, so the row correctly copies local shape.

## Verdict

**MERGEABLE.** No blockers. The gate is stronger in both directions, on both files, and I could not construct a documentation error it fails to catch. S1 is a sequencing matter for the coordinator, not a change to this PR.

PR #3044, opened 2026-08-01
Reject truncated child exit expectations

DavidObando:
Follow-up to merged #2997. Replaces closed no-op #3037 with the independent review finding worth keeping.

## Change

Add `Assert.InRange(expectedExitCode, 0, byte.MaxValue)` before comparing the emitted program's child-process exit code. This fails loudly if a future test expects a value that POSIX would truncate to eight bits.

## Verification

- Clean issue suite: 10/10 passed.
- Perturbed `selectedTarget: 3` expectation from `200` to `320`: 1 failed / 9 passed with `Assert.InRange() Failure`.
- Restored issue suite: 10/10 passed.
- Scope: one file, one added line.

### DIFF
--- a/test/Compiler.Tests/Emit/Issue2942ConditionalGotoFinallyFunnelTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2942ConditionalGotoFinallyFunnelTests.cs
@@ -378,6 +378,7 @@ public class Issue2942ConditionalGotoFinallyFunnelTests
             var stdout = stdoutTask.GetAwaiter().GetResult();
             var stderr = stderrTask.GetAwaiter().GetResult();
             Assert.True(exited, $"child execution timed out\nstdout:\n{stdout}\nstderr:\n{stderr}");
+            Assert.InRange(expectedExitCode, 0, byte.MaxValue);
             Assert.Equal(expectedExitCode, process.ExitCode);
             Assert.Equal(string.Empty, stdout);
             Assert.Equal(string.Empty, stderr);


DavidObando:
## Review: MERGEABLE

Reviewed at head `3aea03f7`, `main` = `fd34166a8`, **computed merge-base `4bba0643a`** (three-dot, freshly fetched). Net at tip: **1 file, +1 −0**.

Everything in the body reproduced exactly. Measurements below are from `/tmp/d3044/w` at `3aea03f7`.

### The assertion is live, not decoration

This is the part worth proving, because on the current values (`10`, `11`, `61`, `200`, `221` — all ≤ 255) `Assert.InRange` never fires, so a green suite proves nothing about it.

| state | result |
|---|---|
| clean, whole file | **10/10 passed** |
| `[InlineData(3, 200)]` → `320` | **1 failed / 9 passed**, at `…FunnelTests.cs:381` |
| restored | **10/10 passed** |

The failure is the new assertion, not a downstream one:

```
Assert.InRange() Failure: Value not in range
Range:  (0 - 255)
Actual: 320
   at …VerifyLoadAndRunChild(…) in …Issue2942ConditionalGotoFinallyFunnelTests.cs:line 381
```

I ran the **whole file** (`~Issue2942ConditionalGotoFinallyFunnelTests` is the only class in it), not a narrower filter, so the shared helper is exercised by all ten tests. No existing test relied on the old behaviour.

`byte.MaxValue` = 255 is the right bound for a POSIX 8-bit wait status, and the assertion is placed before `Assert.Equal`, so it wins the race and gives the clearer message.

### One correction to the premise, in the PR's favour

The brief I was given described this as *replacing a silent `& 0xff` truncation in the shared helper*. **There is no mask on this branch's base.** At merge-base `4bba0643a` the helper is:

```csharp
Assert.True(exited, …);
Assert.Equal(expectedExitCode, process.ExitCode);
```

Searching the base file for `& 0x`, `% 256`, `& 255`, `MaxValue` or `Truncate` returns nothing — and that grep is live, because the identical pattern matches `:381` at this PR's head. The `& 0xff` existed only in closed #3037. So there was nothing for this PR to remove, and the body is right not to claim it removed anything.

Worth recording what the assertion does and does not buy, since it changes how it should be described:

| expectation | base behaviour | with this PR |
|---|---|---|
| `320` | **fails** — `Assert.Equal() Failure: Expected 320, Actual 200` | fails — `Assert.InRange` , `Range (0-255), Actual 320` |

The out-of-range case already failed on base; `expectedExitCode` is only the assertion's expectation and never flows into the emitted program, so nothing was being silently truncated into a false pass here. What the line actually buys is (a) a failure that names the real cause instead of a confusing value mismatch, and (b) a genuine portability guard — Windows exit codes are 32-bit, so a `320`/`320` pair could pass there while failing on POSIX. Both are worth one line.

### Scope

- **No #3037 value changes.** The diff is +1/−0; `11`, `61`, `200`, `221` are untouched, and the `Literal(100)` occurrences at `:145`/`:201` are pre-existing bound-tree construction, not exit-code expectations.
- Test-only; no production code, no diagnostic descriptor, so the docs gate does not apply.

## NIT — the assertion runs after a child process

`test/Compiler.Tests/Emit/Issue2942ConditionalGotoFinallyFunnelTests.cs:381`

`expectedExitCode` is a parameter known at entry, but the check sits after `Process.Start`, `ReadToEndAsync` on both pipes and a `WaitForExit(10_000)`. Moving it to the top of `VerifyLoadAndRunChild` (`:330`) would reject an impossible expectation without emitting an assembly or spawning a child. Purely a fast-fail nicety — the assertion is correct where it is.

---

**Verdict: MERGEABLE.** One line, proven to fire, proven not to disturb the other ten tests, correct bound, and an accurate PR body. This is the right piece to have salvaged from #3037.

PR #3045, opened 2026-08-01
gsi: preserve lambda method identity per site

### LINKED ISSUE #2991: gsi: distinct lambdas report the same Method.Name

DavidObando:
## Summary

- Cache interpreter delegate factories per function-literal site and CLR delegate type instead of per delegate type alone.
- Name generated methods with the compiler's existing scheme: `<lambdaN>` for ordinary non-capturing lambdas, `Invoke` for closure-hosted lambdas, and declared names for method groups.
- Replace the old test assertion that pinned the collision while retaining its original source; add identity, stability, naming, host-shape, and cache-shape coverage with distinct values.

The issue body overstates name uniqueness: `gsc` intentionally emits `Invoke` for distinct capturing sites and can reuse `<lambda1>` across scopes. The real parity rule is distinct `MethodInfo` per lambda site, stable `MethodInfo` for repeated evaluation of one site, and the same method name as `gsc`.

## Performance trade-off

The old cache compiled one expression factory per CLR delegate type. Correct per-site `MethodInfo` identity requires one `Expression.Compile()` per live `(FunctionSymbol, delegateType)`, so programs with many distinct lambda sites compile more factories. Repeated evaluation of one site still reuses its factory, and the outer `ConditionalWeakTable` lets entries disappear with unreachable symbols.

## Validation

- `dotnet build` — 0 warnings, 0 errors
- Full interpreter suite — 542 passed
- Targeted identity, callable-materialization, and generic-local-function tests — 24 passed
- `gsc` / `gsi` identity differential — exact match: `Invoke`, `Invoke`, `<lambda1>`, `False`, `True`, `11`, `22`, `7`
- `samples/FuncToSystemDelegate.gs`: `gsc` and `gsi` both print `<lambda1>` / `<lambda2>`
- Mutation review found `!node.Function.IsGeneric` survived the full suite. Source tracing proved generic local functions become `BoundLocalFunctionDeclaration` nodes and bypass lambda materialization, so the dead term was deleted. All remaining tested functional mutations fail the identity suite. Restored `GSharp.Core.dll` hash: `faab2ab47521d0605e9a4fa13741a374badd8b8e93b78cc3cf00bb176b2fd435`.

## Related work

- #3019: no textual overlap; its nominal delegate target selection happens in binding before this runtime per-site identity cache.
- #3035: adjacent conversion path only. This PR preserves the existing two-argument `CreateInterpreterDelegate` overload, while source-declared named delegates structurally equivalent to `Func`/`Action` retain the already named site delegate.
- #3050: `Evaluator.Expressions.cs` was not its failure site. On current `main`, both its original repro and a G#-declared-type variant now produce `7` in `gsc` and `gsi`; findings posted on the issue.

Fixes #2991

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -28,6 +28,7 @@ public sealed partial class Evaluator
 #pragma warning restore CA1001
 {
     private static readonly System.Runtime.CompilerServices.ConditionalWeakTable<Type, Func<Func<object[], object>, Delegate>> InterpreterDelegateFactories = new();
+    private static readonly System.Runtime.CompilerServices.ConditionalWeakTable<FunctionSymbol, ConcurrentDictionary<Type, Func<Func<object[], object>, Delegate>>> InterpreterDelegateFactoriesBySite = new();
     private static readonly System.Runtime.CompilerServices.ConditionalWeakTable<Delegate, ClosureValue> InterpreterClosures = new();
 
     private object EvaluateExpression(BoundExpression node)
@@ -780,7 +781,9 @@ public sealed partial class Evaluator
         }
 
         var body = node.LoweredBody ??= (BoundBlockStatement)Lowering.Lowerer.Lower(node.Body);
-        return MaterializeClosure(new ClosureValue(node.Function, body, node.FunctionType, captured));
+        return MaterializeClosure(
+            new ClosureValue(node.Function, body, node.FunctionType, captured),
+            GetLambdaMethodName(node));
     }
 
     private object EvaluateMethodGroupExpression(BoundMethodGroupExpression node)
@@ -799,7 +802,9 @@ public sealed partial class Evaluator
             captured[node.Function.ThisParameter] = EvaluateExpression(node.Receiver);
         }
 
-        return MaterializeClosure(new ClosureValue(node.Function, body, node.FunctionType, captured));
+        return MaterializeClosure(
+            new ClosureValue(node.Function, body, node.FunctionType, captured),
+            node.Function.Name);
     }
 
     private object EvaluateClrMethodGroupExpression(BoundClrMethodGroupExpression node)
@@ -847,22 +852,51 @@ public sealed partial class Evaluator
         return InvokeClosure((ClosureValue)targetValue, arguments);
     }
 
-    private object MaterializeClosure(ClosureValue closure)
+    private object MaterializeClosure(ClosureValue closure, string methodName)
     {
         if (closure.FunctionType.ClrType is not Type delegateType)
         {
             return closure;
         }
 
-        var result = CreateInterpreterDelegate(delegateType, args => InvokeMaterializedClosure(closure, args));
+        var result = CreateInterpreterDelegate(
+            delegateType,
+            closure.Function,
+            methodName,
+            args => InvokeMaterializedClosure(closure, args));
         InterpreterClosures.Add(result, closure);
         return result;
     }
+
+    private static string GetLambdaMethodName(BoundFunctionLiteralExpression node)
+    {
+        if (node.CapturedVariables.Length != 0
+            || (!node.Function.IsGeneric
+                && node.Function.LexicalEnclosingType is StructSymbol enclosingType
+                && enclosingType.TypeParameters.IsDefaultOrEmpty))
+        {
+            return "Invoke";
+        }
+
+        return node.Function.Name;
+    }
+
+    private static Delegate CreateInterpreterDelegate(
+        Type delegateType,
+        FunctionSymbol function,
+        string methodName,
+        Func<object[], object> invoke)
+        => InterpreterDelegateFactoriesBySite
+            .GetValue(function, _ => new ConcurrentDictionary<Type, Func<Func<object[], object>, Delegate>>())
+            .GetOrAdd(delegateType, type => BuildInterpreterDelegateFactory(type, methodName))(invoke);
 
     private static Delegate CreateInterpreterDelegate(Type delegateType, Func<object[], object> invoke)
         => InterpreterDelegateFactories.GetValue(delegateType, BuildInterpreterDelegateFactory)(invoke);
 
     private static Func<Func<object[], object>, Delegate> BuildInterpreterDelegateFactory(Type delegateType)
+        => BuildInterpreterDelegateFactory(delegateType, null);
+
+    private static Func<Func<object[], object>, Delegate> BuildInterpreterDelegateFactory(Type delegateType, string methodName)
     {
         var invokeMethod = delegateType.GetMethod(nameof(Action.Invoke));
         var invokeParameter = System.Linq.Expressions.Expression.Parameter(typeof(Func<object[], object>), "invoke");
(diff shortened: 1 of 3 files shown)


DavidObando:
## Review — **MERGEABLE** (1 SHOULD-FIX, 2 NITs)

Merge base against freshly-fetched `gh/main`: **`fd34166a8`** — 0 behind, 1 ahead, head **is** the merged tree. Three-dot scope **3 files, +233/−12**. Build **rc=0, 0 warnings**. CI 8 SUCCESS / 2 SKIPPED including `e2e`. Interpreter suite **527/527**. No diagnostic IDs added, no docs rows — docs gate not in play.

### The names are not merely distinct — they match `gsc`
This was the thing most worth getting wrong, so I treated `gsc` as the oracle and compared every shape. All line numbers at this PR's head.

| shape | `gsc` (oracle) | `gsi` base | `gsi` head |
|---|---|---|---|
| two non-capturing lambdas | `<lambda1>` / `<lambda2>` | **`lambda_method84` / `lambda_method84`** | `<lambda1>` / `<lambda2>` ✓ |
| same site, two closures | `Invoke` / `Invoke` | `lambda_method124` ×2 | `Invoke` / `Invoke` ✓ |
| zero-capture, **non-generic** host | `Invoke` | `lambda_method157` | `Invoke` ✓ |
| zero-capture, **generic** host | `<lambda1>` | `lambda_method168` | `<lambda1>` ✓ |
| method group | `first` / `second` | `lambda_method110` ×2 | `first` / `second` ✓ |

Base emitted the **same** name twice — that is #2991, and the name was a cache-slot artifact (`lambda_methodNN`), not the compiler's scheme.

The part that looks arbitrary on reading — `GetLambdaMethodName` returning `Invoke` for a non-generic host but `<lambda1>` for a generic one (`Evaluator.Expressions.cs:869-880`) — **is faithful to the emitter**, not reverse-engineered: `gsc` itself makes exactly that split. I would have filed this as a blocker on inspection; measuring is what settled it.

### The actual invariant, checked in both directions
Names are the symptom; the fix's claim is `MethodInfo` identity. Verified against `gsc`:

| | `gsc` | base | head |
|---|---|---|---|
| distinct sites → `Method.Equals` | `False` | **`True`** (the collision) | `False` ✓ |
| same site twice → `Method.Equals` | `True` | `True` | `True` ✓ |

Distinctness **and** stability, both matching the oracle. This is why per-site keying is required and per-`(type, name)` keying would not do: two distinct *capturing* sites are both named `Invoke`, so only a per-site key can keep their `MethodInfo` distinct — and `gsc` does keep them distinct.

### Stability and leakage
- **Deterministic:** 3/3 identical runs.
- **Renumbering:** inserting an unrelated lambda first shifts the rest — `<lambda1>|<lambda2>|<lambda3>` — and `gsc` shifts **identically**. Tracking the oracle is the correct behaviour, not a defect.
- **CLR leakage (#3039):** **0 of the 13 emitted name lines** contain an assembly-qualified form.

### The sibling-test edit is declared and is not concealment
`Issue2928CallableMaterializationTests.Callable_DelegateFactoryIsReusedPerType` asserted `Item1.Method == Item2.Method` — it pinned interpreter-internal cache sharing that **diverges from `gsc`**. Inverting it is correct, and your body declares the replacement explicitly.

I checked the specific risk that the source rewrite hid a still-failing case: the **original** source (`(value int32) -> value + 1` / `+ 2`) also yields `<lambda1>` / `<lambda2>` == `gsc` at head. Nothing is hidden — see NIT 2.

### Mutation — 7 mutants, per-sub-term
Clean `0f1c48b0`, 527/527. Build rc verified on every mutant.

| mutant | build | hash | kills |
|---|---|---|---|
| carve-out → always `Function.Name` | rc=0 | `052f9599` | 3 |
| carve-out → always `"Invoke"` | rc=0 | `acc2dad7` | 2 |
| **T1** drop `CapturedVariables.Length != 0` | rc=0 | `6e4735e6` | 2 ✓ |
| **T2** drop `!node.Function.IsGeneric` | rc=0 | `8f3fefb0` | **0 — survives 527/527** |
| **T3** drop `TypeParameters.IsDefaultOrEmpty` | rc=0 | `abadf679` | 1 ✓ |
| method group name → `"Invoke"` | rc=0 | `31dd3821` | 1 ✓ |
| per-site cache → per-type cache | rc=0 | `373d149b` | 6 ✓ |

Restored → `0f1c48b0`, probe re-flips to `<lambda1>`/`<lambda2>`, tree clean. My first T1 attempt failed to build on SA1119 and was rerun rather than counted — a mutant that does not build is no result in either direction.

---

### SHOULD-FIX — `!node.Function.IsGeneric` is pinned by nothing (`Evaluator.Expressions.cs:871`)
Your body reports eight mutants all killing targeted tests. That holds for the mutants chosen, but none decomposes the compound condition, and **the middle term is invisible to the whole 527-test suite**. Six of my seven mutants died; this one survived.

I could not settle whether it is dead or load-bearing. To discriminate you need `IsGeneric` true **and** a non-generic `StructSymbol` host **and** zero captures. Four attempts:
- lambda in a generic *method* of a non-generic class → `IsGeneric` false, both give `Invoke` (== `gsc`)
- lambda with a `T`-typed parameter, and with `T` return → `GS9999` in `gsi` before the name is observable (**pre-existing at base — not yours**, see below)
- reading `.Method.Name` inside the generic method → `GS0158` in both engines

So: either delete the term as dead, or pin it with a test. Right now a future edit can remove it and CI will not notice. You know the binder better than my four probes do — but as it stands the term is unjustified by construction.

### NIT 1 — per-site caching multiplies `Expression.Compile()`
`InterpreterDelegateFactoriesBySite` compiles one factory per `(FunctionSymbol, delegateType)` instead of per `delegateType`. That is inherent to the fix — distinct `MethodInfo` requires distinct compiled delegates — and the `ConditionalWeakTable` keeps it bounded by live symbols. Worth one line in the body that the old test name (`…IsReusedPerType`) was pinning a deliberate perf optimisation now traded away for parity.

### NIT 2 — the sibling test changed source, assertion, **and** name in one edit
Only the assertion and name needed to change; I verified the original source passes the new assertion. Changing all three makes the edit harder to audit than it needed to be.

---

### Out of scope — pre-existing, worth filing separately
A **generic zero-capture lambda returned from a generic method** fails under `gsi`:
```gsharp
class Maker { func make[T]() (T) -> T { return (x T) -> x } }
Console.WriteLine(Maker{}.make[int32]()(7))
```
`gsc` → `7`. `gsi` → `GS9999: Object type System.Delegate does not match target type ClosureValue` (rc=1). **Identical at base `fd34166a8`**, so this PR neither causes nor worsens it.

**Verdict: MERGEABLE.**

DavidObando:
Resolved escalated blocker in `044bb093e`:

- Deleted `!node.Function.IsGeneric` as dead on this evaluator path. `IsGeneric` is set only for generic local-function literals, which bind as `BoundLocalFunctionDeclaration` and are handled directly in `Evaluator.Statements.cs`; they never reach `EvaluateFunctionLiteralExpression` / `GetLambdaMethodName`.
- Restored the Issue2928 test's original source; only the test name and identity assertion now change.
- Disclosed the per-site `Expression.Compile()` trade-off in the PR body.

Validation on `a4a660f32`: clean build, 542/542 interpreter tests, exact `gsc`/`gsi` identity differential. Findings also posted on #3050.

PR #3046, opened 2026-08-01
gsi: coerce closures at CLR delegate boundaries

### LINKED ISSUE #2990: gsi: untyped lambda argument is not coerced to Func at a CLR call boundary

DavidObando:
## Summary
- coerce raw interpreter closures to each reflection target parameter's concrete delegate type
- require every `BuildRefSlots` caller to provide its target explicitly; all five CLR sites pass a `MethodBase`, while the G# user-function path passes explicit `null`
- preserve original argument-evaluation order for imported instance calls

The issue title is too narrow: typed lambdas fail too. Trigger is a G#-declared type making the closure function type lack a CLR type.

## Validation
- `dotnet build -c Release`
- `dotnet test test/Interpreter.Tests/Interpreter.Tests.csproj -c Release --filter FullyQualifiedName~Issue2990ClrDelegateBoundaryTests` (5 CLR-site tests + 1 user-function guard + 1 required-parameter contract test)
- compiler/interpreter sample parity: `3 6 2 8 30 fig kiwi apple`
- each of the five CLR call-site mutants builds successfully and kills only its named site test
- optional-parameter mutant builds successfully and kills the required-parameter contract test

Fixes #2990

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -835,7 +835,7 @@ public sealed partial class Evaluator
     private object EvaluateImportedCallExpression(BoundImportedCallExpression node)
     {
         var args = new object[node.Arguments.Length];
-        var refSlots = BuildRefSlots(node.Arguments, node.ArgumentRefKinds, args);
+        var refSlots = BuildRefSlots(node.Arguments, node.ArgumentRefKinds, args, node.Function.Method);
 
         // Issue #1599: a generic BCL method closed over a same-compilation user
         // value type (e.g. `Enum.TryParse[Color]`) was closed over a value-type
@@ -949,10 +949,9 @@ public sealed partial class Evaluator
         }
 
         receiver = UnwrapClrReceiver(receiver);
-        var args = new object[node.Arguments.Length];
-        var refSlots = BuildRefSlots(node.Arguments, node.ArgumentRefKinds, args);
-
         var method = ResolveMethodForReceiver(node.Method, receiver);
+        var args = new object[node.Arguments.Length];
+        var refSlots = BuildRefSlots(node.Arguments, node.ArgumentRefKinds, args, method);
 
         // concurrency: see TryGetInterpreterMapLock — routes `range m`'s
         // `GetEnumerator()`/`MoveNext()`, `m.ContainsKey(k)`, `m.Remove(k)`,
@@ -1121,9 +1120,11 @@ public sealed partial class Evaluator
     private List<(int Index, BoundExpression Operand)> BuildRefSlots(
         ImmutableArray<BoundExpression> arguments,
         ImmutableArray<RefKind> refKinds,
-        object[] args)
+        object[] args,
+        MethodBase method = null)
     {
         List<(int Index, BoundExpression Operand)> refSlots = null;
+        var parameters = method?.GetParameters();
 
         for (int i = 0; i < arguments.Length; i++)
         {
@@ -1147,6 +1148,23 @@ public sealed partial class Evaluator
             else
             {
                 args[i] = EvaluateExpression(arg);
+            }
+
+            if (parameters != null && args[i] is ClosureValue closure)
+            {
+                var parameterType = parameters[i].ParameterType;
+                if (parameterType.IsByRef)
+                {
+                    parameterType = parameterType.GetElementType();
+                }
+
+                if (parameterType?.GetMethod(nameof(Action.Invoke)) != null
+                    && typeof(Delegate).IsAssignableFrom(parameterType))
+                {
+                    args[i] = CreateInterpreterDelegate(
+                        parameterType,
+                        invokeArgs => InvokeMaterializedClosure(closure, invokeArgs));
+                }
             }
         }
 
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -63,7 +63,8 @@ public sealed partial class Evaluator
                 : BuildRefSlots(
                     node.Arguments,
                     node.Function.Parameters.Select(p => p.RefKind).ToImmutableArray(),
-                    args);
+                    args,
+                    null);
             var locals = new ConcurrentDictionary<VariableSymbol, object>();
 
             for (int i = 0; i < node.Arguments.Length; i++)
@@ -949,9 +950,9 @@ public sealed partial class Evaluator
         }
 
         receiver = UnwrapClrReceiver(receiver);
-        var method = ResolveMethodForReceiver(node.Method, receiver);
         var args = new object[node.Arguments.Length];
-        var refSlots = BuildRefSlots(node.Arguments, node.ArgumentRefKinds, args, method);
+        var refSlots = BuildRefSlots(node.Arguments, node.ArgumentRefKinds, args, node.Method);
+        var method = ResolveMethodForReceiver(node.Method, receiver);
 
         // concurrency: see TryGetInterpreterMapLock — routes `range m`'s
         // `GetEnumerator()`/`MoveNext()`, `m.ContainsKey(k)`, `m.Remove(k)`,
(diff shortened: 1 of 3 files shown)


DavidObando:
## Review — **MERGEABLE** (1 SHOULD-FIX, 2 NITs)

Merge base computed against freshly-fetched `gh/main`: **`fd34166a8`** — 0 behind, 1 ahead, so head **is** the merged tree. Three-dot scope **3 files, +148/−6**. Build **rc=0, 0 warnings**. CI 8 SUCCESS / 2 SKIPPED including `e2e`.

### The defect is real and loud
At base all three PR shapes die with `GS9999 … ClosureValue cannot be converted to Func\`2[System.Object,System.Boolean]` (rc=1). Trigger confirmed: a G#-declared type makes the closure's function type lack a CLR type, so the delegate erases to `object` and `MaterializeClosure` hands back a raw `ClosureValue`. Your body's correction of the issue title is right — see `u9` below.

### Breadth — 4 tests is a thin denominator, so I measured 10 shapes
`gsc` as oracle. All at `Evaluator.Calls.cs:1161-1179` (this PR's head).

| shape | `gsc` | base | head |
|---|---|---|---|
| `Action`/void return (`List.ForEach`) | 11/22 | **GS9999** | 11/22 ✓ |
| 2-param explicit (`List.Sort`, `Comparison`) | 11 | **GS9999** | 11 ✓ |
| 2-param inferred (`Select((it,i) -> …)`) | 21 | **GS9999** | 21 ✓ |
| generic 2-param (`Aggregate`) | 3 | **GS9999** | 3 ✓ |
| returns user type (`Select`) | 77 | **GS9999** | 77 ✓ |
| **typed** 1-param | 99 | **GS9999** | 99 ✓ |

So the fix is **not** narrow: it keys on the reflection parameter type, not on arity or return shape. 0/1/2 params, `Action` vs `Func`, inferred vs stated, all pass. Four non-erased `int32` shapes behave identically at base and head, as expected. `parameters[i]` is index-safe under `params` expansion on an updated site (`StringBuilder.AppendFormat` with 4 args → `9/8/7/6`, == `gsc`).

### Mutation — each hunk kills a **distinct** test
`GSharp.Core.dll` hashes, build rc checked every time. Clean `a0ef666f`, 524/524.

| mutant | build | hash | kills |
|---|---|---|---|
| revert `Calls.cs:847` (`node.Function.Method`) | rc=0 | `18533120` | **only** `ImportedStaticCall…` |
| revert `Calls.cs:963` (`method`) | rc=0 | `d347e45d` | **only** `ImportedInstanceCall…` |
| revert `Expressions.cs:1223` (`node.Constructor`) | rc=0 | `5e69a7fb` | **only** `ClrConstructor…` |
| revert all production to base | rc=0 | `5fb762a3` | exactly those 3 |

Three sites, three distinct kills — genuine per-site pinning, not three names over one path. Restored → `a0ef666f`, tree clean.

One mutant (deleting the coercion block) failed to build on **SA1508** while its md5 still changed; discarded as **no result** rather than counted.

---

### SHOULD-FIX — 2 of 5 `BuildRefSlots` call sites still pass no method, with the `MethodBase` in hand
I enumerated the population independently rather than accepting a count: **5** call sites exist in `src/`. Three are updated. Two are not:

- `Evaluator.Expressions.cs:1174` `InvokeClrCtor` — has `ctor`, next line does `ctor.Invoke(args)`
- `Evaluator.Expressions.cs:1233` `EvaluateClrStaticCallExpression` — has `node.Method`, next line does `node.Method.Invoke(null, args)`

Both are reflection-invoke boundaries with the identical failure mode, and the fix is one identifier each.

**I measured reachability rather than asserting it.** A hard `throw` at both sites on any arg-bearing call: hit **0 times across the 524-test interpreter suite** and **0 times across my 6 erased probes** — suite stayed fully green with the throw in place. I also failed to construct a reproducer in 2 attempts (`class MyLazy : Lazy[Item]` with both a lambda and a method group → `GS0214` at bind time, identically in `gsc` and `gsi`). So this is **not** a blocker: currently unreachable, and by the same measurement those two sites have no arg-bearing coverage at all today.

Recommend: pass `ctor` and `node.Method`, and make the parameter **required** (`MethodBase method`, explicit `null` where genuinely G#-side). The `= null` default is what let two sites be missed silently and is what will let the sixth site added next month be missed too — one choke point beats five call sites remembering.

### NIT 1 — `UserFunctionCall_PreservesRawClosure` survives full production revert
It pins pre-existing behaviour, not this change. That is fine and your body describes it correctly as *preserve*, not as fix coverage — no misattribution. Flagging only so the "4 tests" number is not read as 4 tests of the fix; it is 3 + 1 guard.

### NIT 2 — evaluation-order change at `Evaluator.Calls.cs:958-963`
`ResolveMethodForReceiver` now runs **before** argument evaluation (necessary, since the method is needed to build slots). Benign as far as I can measure, but it is an observable reordering not mentioned in the body.

**Verdict: MERGEABLE.**

DavidObando:
Addressed reviewer items in `da0367501`:

- `MethodBase` is now required. All five CLR `BuildRefSlots` call sites pass their concrete method/constructor; the sole G# user-function caller passes explicit `null`. Added a contract test; restoring the optional default builds successfully but the test fails.
- Restored original imported-instance ordering: arguments are evaluated before `ResolveMethodForReceiver`; `node.Method` supplies coercion metadata.
- Rebased onto `a4a660f32`; Release build is clean, targeted suite is 5/5, and `gsc`/`gsi` still match `3 6 2 8 30 fig kiwi apple`.

DavidObando:
`main` advanced again after my prior comment. Rebased and force-pushed onto `f820484c3`; new head `8517f9c77`. Content is unchanged. Release build remains clean, targeted suite remains 5/5, and `gsc`/`gsi` still match `3 6 2 8 30 fig kiwi apple`.

DavidObando:
Rebased onto `61b7a08c1`; head `9a6b9bd9f`. Verified all five CLR sites on this head and added coverage for the two formerly unreachable evaluator helpers. Each mutant passed the Core build (`rc=0`) and killed its distinct test:

- `Evaluator.Calls.cs:839` → `ImportedStaticCall_CoercesTypedAndUntypedClosures` (`fad1c916`)
- `Evaluator.Calls.cs:954` → `ImportedInstanceCall_CoercesTypedAndUntypedClosures` (`69850f7f`)
- `Evaluator.Expressions.cs:1209` → `InvokeClrCtor_CoercesClosureWithUserTypeParameter` (`b5efedb2`)
- `Evaluator.Expressions.cs:1258` → `ClrConstructor_CoercesClosureWithUserTypeReturn` (`314da3d8`)
- `Evaluator.Expressions.cs:1268` → `ClrStaticCall_CoercesClosureWithUserTypeParameter` (`81d8bd88`)

Restored Core hash: `cd45de17`; build clean, 7/7 targeted tests pass, sample `gsc`/`gsi` values match.

PR #3047, opened 2026-08-01
Fix struct computed accessor virtual promotion

DavidObando:
## Summary

- Add test-only regression coverage for #3001's minimal struct + computed-accessor shape.
- Production fix already landed in #2968; this PR changes no compiler production code.
- Cover ordinary and explicit computed properties/indexers, nullable sequence returns, getter/setter dispatch, auto-property control, and mismatched-indexer non-promotion.

Regression coverage for #3001. This PR does **not** claim the production fix.

## Rebase and scope

Computed merge base: `5a27b553235714591c6ea7a0c38a75f3adee6672`.

Three-dot diff:

```
 test/Compiler.Tests/Emit/Issue3001StructComputedAccessorVirtualTests.cs | 263 +++++++++++++++++++++
 1 file changed, 263 insertions(+)
```

Production blobs match current `main` byte-for-byte:

| file | HEAD blob | main blob |
|---|---|---|
| `MemberDefEmitter.cs` | `4634f8bb6251cc08dcaa4931024ac879adb4a499` | `4634f8bb6251cc08dcaa4931024ac879adb4a499` |
| `MethodInfoHelpers.cs` | `0866e1f7d45d516d04402172fcf4e8f805048307` | `0866e1f7d45d516d04402172fcf4e8f805048307` |

Statement order matches main. First rebase produced one `MemberDefEmitter.cs` conflict, only comment wording/wrapping:

```diff
- // PropertyImplicitlyImplementsInterface deliberately excludes
- // explicit implementations. Per ECMA-335 ..., Final additionally prevents ...
+ // PropertyImplicitlyImplementsInterface deliberately excludes explicit implementations.
+ // Per ECMA-335 ...; Final prevents ...
```

No executable statement differed. Kept main's comment. `MethodInfoHelpers.cs` merged with zero residue.

Current remote main later advanced by CI-only #3053; both production blob IDs remain identical.

## Production mutation evidence

Corrected stale-DLL audit: each mutant changed production code only and built with `BUILD_RC=0`. Hashes below come from `out/bin/Release/Core/GSharp.Core.dll` after deleting Core Release outputs and rebuilding; every artifact mtime followed its build start. Every restore rebuilt byte-for-byte to clean hash `5f78da05d5ac7d423562ec1353862946`. Earlier `Compiler/GSharp.Core.dll` hashes were stale and are superseded; test kill/survive results are unchanged.

| #2968 hunk | independent condition mutant | Core hash | Issue3001 result |
|---|---|---|---|
| G# interface property matcher | invert `PropertyMatchesInterfaceSlot` | `54a40b785ae725b39f0b35c08df9e36f` | **5 failed / 1 passed**; only `ExplicitComputedIndexer_LoadsAndDispatches` survived |
| imported CLR property matcher | invert `parametersMatch` | `7da38c16c89150fa16a0ed378a16ee99` | **0 failed / 6 passed** — this file does not cover the CLR branch |
| computed accessor symbol identity | invert getter/setter `ReferenceEquals` identity | `bf93a9eb2660f4ba8cd5e07eb465782b` | **5 failed / 1 passed**; only `AutoPropertyStruct_RemainsLoadable` survived |

G# matcher failures: `AutoPropertyStruct_RemainsLoadable`, `MismatchedIndexerParameter_DoesNotPromoteOrdinaryAccessor`, `OrdinaryComputedProperty_LoadsAndDispatches`, `ComputedGetterAndSetter_LoadAndDispatch`, `NullableSequenceComputedProperty_LoadsAndDispatches`.

Accessor-identity failures: `MismatchedIndexerParameter_DoesNotPromoteOrdinaryAccessor`, `ExplicitComputedIndexer_LoadsAndDispatches`, `OrdinaryComputedProperty_LoadsAndDispatches`, `ComputedGetterAndSetter_LoadAndDispatch`, `NullableSequenceComputedProperty_LoadsAndDispatches`.

## Verification

- Full `Compiler.Tests`: **4033 passed / 0 failed / 0 skipped**.
- Merged-tree bidirectional diagnostics gate: **4 passed / 0 failed / 0 skipped**.
- `docs/diagnostics.md` changes: none.
- `website/docs/ref/diagnostics.md` changes: none.
- Positive fixtures use `Assembly.Load(File.ReadAllBytes(...))` + `GetTypes()`, then assert exact bounded-process output. `IlVerifier` is supplementary only.
- No local e2e/SDK run.

### DIFF
--- a/test/Compiler.Tests/Emit/Issue3001StructComputedAccessorVirtualTests.cs
+++ b/test/Compiler.Tests/Emit/Issue3001StructComputedAccessorVirtualTests.cs
@@ -1,0 +1,263 @@
+// <copyright file="Issue3001StructComputedAccessorVirtualTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Metadata;
+using System.Reflection.PortableExecutable;
+using GSharp.Compiler;
+using Xunit;
+
+namespace GSharp.Compiler.Tests.Emit;
+
+/// <summary>
+/// Issue #3001: computed struct property accessors that implement interface
+/// slots must be emitted virtual.
+/// </summary>
+public class Issue3001StructComputedAccessorVirtualTests
+{
+    [Fact]
+    public void OrdinaryComputedProperty_LoadsAndDispatches()
+    {
+        const string Source = """
+            package Issue3001Ordinary
+            import System
+
+            interface IName {
+                prop Label string { get; }
+            }
+
+            struct S : IName {
+                prop Label string -> "hi"
+            }
+
+            var value IName = S{}
+            Console.WriteLine(value.Label)
+            """;
+
+        AssertRuns(Source, nameof(OrdinaryComputedProperty_LoadsAndDispatches), "hi\n");
+    }
+
+    [Fact]
+    public void ExplicitComputedIndexer_LoadsAndDispatches()
+    {
+        const string Source = """
+            package Issue3001Explicit
+            import System
+
+            interface I {
+                prop this[key int32] int32 { get; }
+            }
+
+            struct S(Base int32) : I {
+                private prop (I) this[key int32] int32 -> Base + key
+            }
+
+            var value I = S(5)
+            Console.WriteLine(value[7])
+            """;
+
+        AssertRuns(Source, nameof(ExplicitComputedIndexer_LoadsAndDispatches), "12\n");
+    }
+
+    [Fact]
+    public void NullableSequenceComputedProperty_LoadsAndDispatches()
+    {
+        const string Source = """
+            package Issue3001Sequence
+            import System
+
+            interface IBox {
+                prop Vals sequence[int32?] { get; }
+            }
+
+            func values() sequence[int32?] {
+                yield 5
+                yield nil
+            }
+
+            struct Box : IBox {
+                prop Vals sequence[int32?] -> values()
+            }
+
+            var box IBox = Box{}
+            for value in box.Vals {
+                Console.WriteLine(value == nil ? "nil" : value.ToString())
+            }
+            """;
+
+        AssertRuns(
+            Source,
+            nameof(NullableSequenceComputedProperty_LoadsAndDispatches),
+            "5\nnil\n");
+    }
+
+    [Fact]
+    public void ComputedGetterAndSetter_LoadAndDispatch()
+    {
+        const string Source = """
+            package Issue3001GetterSetter
+            import System
+
+            interface IValue {
+                prop Value int32 { get; set; }
+            }
+
+            struct ValueBox : IValue {
+                var Stored int32
+                prop Value int32 {
+                    get { return Stored }
+                    set { Stored = value + 11 }
+                }
+            }
+
+            var box IValue = ValueBox{}
+            box.Value = 22
+            Console.WriteLine(box.Value)
+            """;
+
+        AssertRuns(Source, nameof(ComputedGetterAndSetter_LoadAndDispatch), "33\n");
+    }
+
+    [Fact]
+    public void AutoPropertyStruct_RemainsLoadable()
+    {
+        const string Source = """
+            package Issue3001AutoProperty
+            import System
+
+            interface IValue {
+                prop Value int32 { get; set; }
+            }
+
+            struct ValueBox : IValue {
+                prop Value int32 { get; set; }
+            }
+
+            var box IValue = ValueBox{}
+            box.Value = 44
+            Console.WriteLine(box.Value)
+            """;
+
+        AssertRuns(Source, nameof(AutoPropertyStruct_RemainsLoadable), "44\n");
+    }
+
+    [Fact]
+    public void MismatchedIndexerParameter_DoesNotPromoteOrdinaryAccessor()
+    {
+        const string Source = """
+            package Issue3001Mismatch
+
+            interface ITextIndexer {
+                prop this[key string] int32 { get; }
+            }
+
+            struct MixedIndexer : ITextIndexer {
+                prop this[key int32] int32 -> 33
+                private prop (ITextIndexer) this[key string] int32 -> 22
+            }
+            """;
+
+        var assemblyPath = Compile(
+            Source,
+            nameof(MismatchedIndexerParameter_DoesNotPromoteOrdinaryAccessor),
+            target: "library");
+
+        using var stream = File.OpenRead(assemblyPath);
+        using var peReader = new PEReader(stream);
+        var reader = peReader.GetMetadataReader();
+        var type = reader.TypeDefinitions
+            .Select(reader.GetTypeDefinition)
+            .Single(candidate => reader.GetString(candidate.Name) == "MixedIndexer");
+        var getter = type.GetMethods()
+            .Select(reader.GetMethodDefinition)
+            .Single(method => reader.GetString(method.Name) == "get_Item");
+
+        Assert.False((getter.Attributes & MethodAttributes.Virtual) != 0);
+    }
+
+    private static void AssertRuns(string source, string name, string expected)
+    {
+        var assemblyPath = Compile(source, name, target: "exe");
+        var assembly = Assembly.Load(File.ReadAllBytes(assemblyPath));
+        Assert.NotEmpty(assembly.GetTypes());
+        IlVerifier.Verify(assemblyPath);
+        Assert.Equal(expected, RunBounded(assemblyPath, name));
+    }
+
+    private static string Compile(string source, string name, string target)
+    {
+        var directory = Path.Combine(
+            AppContext.BaseDirectory,
+            nameof(Issue3001StructComputedAccessorVirtualTests),
+            name);
+        Directory.CreateDirectory(directory);
+        var sourcePath = Path.Combine(directory, "test.gs");
+        var assemblyPath = Path.Combine(directory, name + ".dll");
+        File.WriteAllText(sourcePath, source);
+
+        using var stdout = new StringWriter();
+        using var stderr = new StringWriter();
+        var previousOut = Console.Out;
+        var previousErr = Console.Error;
+        Console.SetOut(stdout);
+        Console.SetError(stderr);
+        int exitCode;
+        try
+        {
+            exitCode = Program.Main(new[]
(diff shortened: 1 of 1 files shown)


DavidObando:
## Rubber-duck verdict: **MERGEABLE** — but production here duplicates open PR #2968

Measured tree `/tmp/a47` = `main` + `ecaa5564d`, merge base `fd34166a8`. Re-verified at new `main` `fdba684b6` (merge clean, build exit 0, `Issue3001` 6/6).

### HEADLINE — production diff identical to #2968, byte-for-byte modulo comment wrap

Comment-stripped, sorted compare of each PR against **its own computed merge base**:

| file | #3047 code lines | #2968 code lines | only in 3047 | only in 2968 |
|---|---|---|---|---|
| `MemberDefEmitter.cs` | 72 | 72 | **0** | **0** |
| `MethodInfoHelpers.cs` | 10 | 10 | **0** | **0** |

#2968 (head `c42191828`) already carry same fix. Unique content here = 263-line test file only. **Not defect. Merge-order call.**

- **Merge #2968 first.** It superset: 9 files, closes #2915/#2954/#2960/#3001.
- After that, this PR production diff become no-op, reduce to **test-only** PR. Keep tests — they exactly minimal struct+computed-accessor shape #2968 lack.

### Q1 — oracle. **PASS.** Right oracle used
`AssertRuns` do `Assembly.Load(File.ReadAllBytes(...))` + `Assert.NotEmpty(assembly.GetTypes())` + `RunBounded` child exec read real output. Not `IlVerifier`-only.
Pre-fix differential: port test file onto unmodified `main`, build exit 0 → **Failed 4, Passed 2, Total 6**. Failure mode real `TypeLoadException: Unable to load one or more of the requested types`. Dead 4: `ComputedGetterAndSetter_LoadAndDispatch`, `ExplicitComputedIndexer_LoadsAndDispatches`, `NullableSequenceComputedProperty_LoadsAndDispatches`, `OrdinaryComputedProperty_LoadsAndDispatches`.

### Q2 — predicate count. **Constraint honoured. One shared predicate, not third**
`MemberDefEmitter.PropertyImplicitlyImplementsInterface` (`MemberDefEmitter.cs:882`, now `internal static`), consumed by `MemberDefEmitter.cs:356`, `:470`, and `MethodInfoHelpers.RequiresVirtualOnValueType`. `PropertyMatchesInterfaceSlot` (`:985`) private comparator of that predicate, not separate predicate.

### Q3 — #2968 co-merge, BOTH orders. Measured
Both orders **conflict** in `src/Core/CodeAnalysis/Emit/MemberDefEmitter.cs` — exactly **1** hunk, **pure comment re-wrap**, zero executable collision. Resolved `--ours` both ways; two resolved trees **code-identical** (`MemberDefEmitter.cs` 1324/1324, `MethodInfoHelpers.cs` 191/191). Co-merge tree build exit 0, `Issue3001|Issue2915|Issue2370|Issue2945|Issue2888` **60/60**.

### Mutation (build exit 0 + md5 change verified each; clean md5 `b9914364729ecabeb8dad80fad452f93`, round-trip exact)
- **H2** declared arm → name-only compare: **FULL `Compiler.Tests` run 3995 tests, 1h01m.** `MismatchedIndexerParameter_DoesNotPromoteOrdinaryAccessor` (`Issue3001…Tests.cs:150`) **dies** ⇒ hardening pinned.
- **H1** `MethodInfoHelpers` arm: pinned by inference — base lacks H1 and fails the 4 pins; H1-present/hardening-absent passes them.
- **H3** CLR arm reverted to name+type only: **survives 186/186** targeted (`Issue3001|Indexer|InterfaceProperty|ImplicitInterface|Issue248|Issue2362|ClrInterface|Issue2915|Issue2888`). → SHOULD-FIX.
- **H4** `HasExplicitInterfaceClause` early-out: **not measured** — declare as unmeasured, not as passing.

### Items
- **SHOULD-FIX** — `MemberDefEmitter.cs` CLR arm unpinned (H3 survives 186/186). Add CLR analogue of `MismatchedIndexerParameter…`.
- **NIT** — declared arm (`:985`) and CLR arm (inline) implement same slot-matching twice. Drift risk.
- **NIT** — `LongBackedFlagsEnum_BitwiseAnd_ProducesCorrectValue` failed once inside H2 full run; **passes 13/13 on restored clean tree**. Unattributed, contention flake. **Explicitly NOT regression claim.**

DavidObando:
## Rubber-duck verdict: **MERGEABLE**

Anchors, each named: `main` **`5ad0109c`**, PR head `ce5ff798a`, computed merge base `5a27b5532` (matches the PR's claim), merged tip `358245c55`. `merge-tree` rc=0, real merge rc=0, **0** conflicts, build rc=0. Three-dot with `--full-index`: **1 file added, +263 −0**. Baseline **6/6**.

### 1. "Test-only" is proven — and provable more broadly than the PR claims

The two quoted blobs match at all three anchors:

| file | merge-base `5a27b553` | pr head `ce5ff798a` | **main `5ad0109c`** |
|---|---|---|---|
| `MemberDefEmitter.cs` | `4634f8bb…` | `4634f8bb…` | `4634f8bb…` |
| `MethodInfoHelpers.cs` | `0866e1f7…` | `0866e1f7…` | `0866e1f7…` |

But two cherry-picked hashes can't prove "no production change". The whole-subtree comparison can, in one line:

```
merged 358245c55 src/ subtree : c339a1b414818f08dc3daefa56831c909939f63e
main   5ad0109c  src/ subtree : c339a1b414818f08dc3daefa56831c909939f63e
```

**One hash covering every production file.** The hand-resolved `MemberDefEmitter.cs` rebase conflict left **zero residue**. Given how a PR earlier this session mis-resolved an "obvious" conflict and silently deleted its own guard, that was the risk worth spending a command on — and it's clean.

### 2. The tests pin the production fix — four arms, four disjoint kills, zero inert tests

Coverage for a fix that landed elsewhere is only worth its bytes if mutating that fix turns it red. All mutants are PRODUCT mutants; all builds rc=0.

| mutant | site | arm disabled | Failed / 6 | which test |
|---|---|---|---|---|
| **M1** | `MethodInfoHelpers.cs:52` | explicit-interface clause | **1** | `ExplicitComputedIndexer_LoadsAndDispatches` |
| **M2** | `MethodInfoHelpers.cs:53` | implicit interface match | **3** | `Ordinary…`, `NullableSequence…`, `ComputedGetterAndSetter…` |
| **M3** | `MethodInfoHelpers.cs:50-53` (drop the guard → over-promote) | non-promotion | **1** | `MismatchedIndexerParameter…` |
| **M4** | `MemberDefEmitter.cs:359` + `:472` | auto-property fallback | **1** | `AutoPropertyStruct_RemainsLoadable` |

**Every one of the six tests is killed by at least one mutant.** Nothing here is decoration.

M4 is the one I'd highlight to the author: the auto-property case is described as a "control", and it genuinely behaves as one — it is *immune* to M1/M2/M3 (which target `RequiresVirtualOnValueType`) and dies only to the **separate** `MemberDefEmitter` fallback path. A control that is itself pinned, by a different code path, is better than most controls I see.

### 3. The non-promotion assertion is load-bearing

Applying the lesson that *which* assertion fails is the whole question: M3 over-promotes every struct property accessor, and the failure is

```
Assert.False() Failure
  at …Issue3001StructComputedAccessorVirtualTests.MismatchedIndexerParameter_DoesNotPromoteOrdinaryAccessor()
  in …Issue3001StructComputedAccessorVirtualTests.cs:line 180
```

Line 180 is `Assert.False((getter.Attributes & MethodAttributes.Virtual) != 0);` — the absence assertion **itself**, not an earlier count, not an exception from `.Single`.

I also checked `.Single(method => name == "get_Item")` at `:178` can't be selecting the wrong accessor, by dumping the emitted metadata rather than assuming:

```
MixedIndexer.get_Item                                  Public, HideBySig, SpecialName          <- ordinary, NOT virtual
MixedIndexer.Issue3001Mismatch.ITextIndexer.get_Item   Private, Final, Virtual, …              <- explicit, virtual
```

The explicit implementation carries a mangled name, so `.Single` is unambiguous and lands on the ordinary accessor — exactly the one the test claims to be asserting about.

### 4. Masking is structurally impossible here

`:52` and `:53` are the two disjuncts of one condition, and `MemberDefEmitter`'s chain is a separate path reached only when the first doesn't apply. I mutated all three independently and the kill sets are disjoint (1 / 3 / 1, plus M3's 1). No arm is hiding behind an earlier one.

### 5. Axis sweep — this one is the good direction

`func Main()` occurrences: **0**, against a positive control of **6** `package Issue3001…` matches. **All six sources are top-level code**, so the `gsc /out:` emit path and the top-level axis are both exercised. The tests do not inherit the in-function shape that has burned this fleet.

Declaration kind is `struct` ×6, `class` ×0 — and unlike #3059 that is **correct, not a gap**: `RequiresVirtualOnValueType(FunctionSymbol, StructSymbol)` is struct-keyed by signature, so classes don't route through the logic under test. No NIT.

The oracles use **distinct values** (`"hi"`, `12`, `5`/`nil`, `33`, `44`) — no all-zero fixture that can't distinguish correct from default — and `AssertRuns` executes the emitted program out-of-process with a bounded timeout rather than relying on `IlVerifier.Verify` alone.

### Items

- **BLOCKER**: none.
- **SHOULD-FIX**: none.
- **NIT** — `AssertRuns` (`:186`) `Assert.NotEmpty(assembly.GetTypes())` is near-vacuous next to the actual execution check. Harmless.
- **NIT** — if a future regression emitted *both* indexers as plain `get_Item`, `:178`'s `.Single` throws `InvalidOperationException` instead of failing the assertion. Still a kill, just a less legible one.

### Disclosures

- My first M4 attempt **did not apply** — the pattern occurs twice in `MemberDefEmitter.cs` (`:359` getter, `:472` setter) and my guard asserted. The 6/6 it produced was a **no-result, not a survival**; I re-applied to both sites and got the 1/6 above. Flagging because that failure mode reads exactly like a surviving mutant.
- Restore verified by source, which is stronger than a DLL hash here: `src/` subtree back to `c339a1b4…` (= `main`'s), `git status --porcelain` clean of tracked modifications, final run **6/6**.
- **Not measured**: whether pre-existing tests on `main` already kill M1/M2 — i.e. whether this coverage is net-new or partly redundant. I had a broad `~Emit` differential running and **stopped it when load hit 45.25**, above the 30 threshold. It bears on the PR's *value*, not its correctness, and changes no verdict.
- Scope: filtered to the 6-test class rather than the ~1h20m full compiler suite — the PR adds one test file and production is byte-identical to `main` at the merged tip.

PR #3048, opened 2026-08-01
Make #3028 ref-return probe an explicit fixture

DavidObando:
## Summary
- remove `RefReturnProbe` from the shared `ProbeRefTypes.cs` helper file
- retain its live coverage as issue-scoped `ManagedByRefAutoDereferenceFixture`
- update #3028 tests to use the explicitly named fixture

## Measurement
The probe is live, not scratch:
- deleting it: Interpreter.Tests **3 failed / 517 passed**; property, indexer, and imported instance-call cases could not resolve the type
- making managed-byref dereference fail like raw-pointer dereference: mutant built successfully, `GSharp.Core.dll` hash changed, and #3028 tests **3 failed / 5 passed**
- restoring production: source and `GSharp.Core.dll` hashes returned exactly; Interpreter.Tests **520/520 passed**

Scanned 45 `GS9999` occurrences across 18 files in `docs`, `website`, `src`, and `test`; found no additional pointer/dereference claim beyond the separately owned diagnostics text.

Follow-up cleanup for #3028. No issue closure intended.

### DIFF
--- a/test/Interpreter.Tests/Issue3004ManagedByRefAutoDereferenceFixture.cs
+++ b/test/Interpreter.Tests/Issue3004ManagedByRefAutoDereferenceFixture.cs
@@ -1,0 +1,25 @@
+// <copyright file="Issue3004ManagedByRefAutoDereferenceFixture.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+namespace GSharp.Interpreter.Tests.Issue3004;
+
+/// <summary>
+/// CLR ref-return shapes used to guard managed-byref auto-dereference.
+/// </summary>
+public sealed class ManagedByRefAutoDereferenceFixture
+{
+    private readonly int[] values = [40, 41, 42];
+
+    /// <summary>Gets the first value by reference.</summary>
+    public ref int Property => ref values[0];
+
+    /// <summary>Gets a value by reference through an indexer.</summary>
+    /// <param name="index">The value index.</param>
+    public ref int this[int index] => ref values[index];
+
+    /// <summary>Gets a value by reference through an instance call.</summary>
+    /// <param name="index">The value index.</param>
+    /// <returns>A reference to the selected value.</returns>
+    public ref int GetValue(int index) => ref values[index];
+}
--- a/test/Interpreter.Tests/Issue3004PointerBoundaryInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue3004PointerBoundaryInterpreterTests.cs
@@ -37,9 +37,9 @@ public class Issue3004PointerBoundaryInterpreterTests
 
     public static TheoryData<string, int> ManagedByRefCases => new()
     {
-        { "RefReturnProbe().Property", 40 },
-        { "RefReturnProbe()[1]", 41 },
-        { "RefReturnProbe().GetValue(2)", 42 },
+        { "ManagedByRefAutoDereferenceFixture().Property", 40 },
+        { "ManagedByRefAutoDereferenceFixture()[1]", 41 },
+        { "ManagedByRefAutoDereferenceFixture().GetValue(2)", 42 },
     };
 
     [Theory]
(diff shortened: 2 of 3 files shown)


DavidObando:
## PR #3048 — **MERGEABLE**

Measured tree: base `fd34166a8` (computed `git merge-base`), head `19fdda601`, `main` `a4a660f32`. Three-dot = **3 files / +29 −24**.

### The "probe was actually LIVE" claim — verified, and in a stronger form than claimed

Author claim: deleting the probe caused 3 failures. **True.** On `main` `a4a660f32`, removing the `RefReturnProbe` class:

```
rc=1  Failed: 3, Passed: 5, Total: 8
  ManagedByRefAutoDereference_RemainsSupported("RefReturnProbe().Property",   40)
  ManagedByRefAutoDereference_RemainsSupported("RefReturnProbe()[1]",         41)
  ManagedByRefAutoDereference_RemainsSupported("RefReturnProbe().GetValue(2)",42)
  Error Message: Function 'RefReturnProbe' doesn't exist.
```

But note the failure mode: that is a **G# binder** error. The rows fail because the fixture vanished, not because the product broke. **A test that dies when you delete it proves nothing.** So I mutated the product instead.

### Production mutation — all 3 cases die

Real site the 3 rows guard is `src/Core/CodeAnalysis/Evaluator.Calls.cs:1121`, the **else**-path of the `GS0513` guard in `EvaluateDereferenceExpression`. Mutant = condition inversion (`IsUnmanagedPointer` → `!IsUnmanagedPointer`); chose inversion over `if (true)` deliberately, because warnings-as-errors turns `if (true)` into CS0162 and **a bad mutant is no result**.

At PR head `19fdda601`, filter `FullyQualifiedName~Issue3004`:

| step | build | `GSharp.Core.dll` md5 | test |
|---|---|---|---|
| clean control | rc=0 | `f5817f6b478a66ca5e68f91036cc14c8` | rc=0, **8 passed** |
| mutant | rc=0 | `d73c8a66…` | rc=1, **4 failed / 4 passed** |
| restored | rc=0 | `f5817f6b478a66ca5e68f91036cc14c8` | rc=0, **8 passed** |

Dead: **all 3** `ManagedByRefAutoDereference_RemainsSupported` rows (40/41/42, distinct) **plus** one `UnmanagedPointerOperation_ReportsGs0513` row. Build rc=0 on both legs, so the mutant is valid in both directions; md5 round trip byte-identical.

### No coverage lost in the move

Provable from the diff alone, no run needed: moved class keeps the same **3 members** (`ref int Property`, `ref int this[int]`, `ref int GetValue(int)`), the same backing array `[40, 41, 42]`, and the 3 theory rows keep the same 3 expected values. The production mutant killing all 3 **at PR head** is the direct confirmation that the renamed fixture still guards them.

### Names

`ManagedByRefAutoDereference_RemainsSupported` / `Issue3004ManagedByRefAutoDereferenceFixture` say what they guard. That was the stated point of the rename; it is achieved.

---

### Correction to the framing (not charged to this PR)

`test/Interpreter.Tests/ProbeRefTypes.cs` was **not** a leftover file added by #3028, and #3048 does **not** delete it. The file pre-exists with unrelated shared CLR-shape fixtures; #3028 added the 20-line `RefReturnProbe` class *into* it, and #3048 moves those lines back out. The remainder is genuinely shared, not scaffolding — `IHasName` used by 6 other test files, `IReadWrite` by 3, `IWithDIM`/`WithDIMImpl` by 2.

Also worth recording *why* the "debug probe" read was plausible: `RefReturnProbe` is referenced only from **G# source strings** inside `TheoryData` (4 grep hits total, incl. its own definition). Deleting it therefore still **compiles** — no C# reference to break — and fails only at runtime. Anything reachable solely through G# string literals will look dead to every C# tool.

### Docs gate

Not exposed. 0 `GS\d{4}` lines in the three-dot diff; positive control 377 in `DiagnosticDescriptors.cs`. Neither `docs/diagnostics.md` nor `website/docs/ref/diagnostics.md` needs a row.

---

### BLOCKER
None.

### SHOULD-FIX
None.

### NIT
- **`test/Interpreter.Tests/ProbeRefTypes.cs:5`** — the PR removes the "probe" *class*, but the file and namespace stay `ProbeRefTypes.cs` / `GSharp.Interpreter.Tests.ProbeRef`. If the objection was that "Probe" reads as debug leftover, the remaining file still reads that way despite holding legitimate shared fixtures. Cosmetic; not worth a rebase on its own.
- No `Fixes` line, which is correct — this is a chore, not an issue fix.

Round trip closed clean on both trees; `git status --porcelain` empty at `19fdda601` and at `a4a660f32`. No e2e run, no writes to `~/.nuget/packages/gsharp.net.sdk/`.

PR #3049, opened 2026-08-01
Fix gsi imported-base derived type identity

### LINKED ISSUE #3015: gsi: imported-base virtual dispatch loses G# derived type identity

DavidObando:
## Summary
- allocate a collectible runtime-derived CLR backing type per interpreted G# class instead of instantiating the imported base directly
- forward every accessible imported-base constructor while preserving generic and nested metadata names
- reify G#-declared generic arguments through their interpreted runtime backing types instead of erasing them to `System.Object`
- retain inherited CLR state/member dispatch and cover public plus protected parameterized constructors

## Three-driver generic parity
All sources execute at top level. Each driver uses its own empty artifact directory. Assembly qualifications are removed from `Type.FullName` only so dynamic and emitted assembly names do not obscure type identity.

| type argument | bare `gsc` (evaluate) | `gsc /out:` (emit) | `gsi` (interpret) |
|---|---|---|---|
| G# `Payload` | ``Box`1[[Payload]]`` | ``Box`1[[Payload]]`` | ``Box`1[[Payload]]`` |
| CLR `string` | ``Box`1[[System.String]]`` | ``Box`1[[System.String]]`` | ``Box`1[[System.String]]`` |
| nested `Box[Payload]` | ``Box`1[[Box`1[[Payload]]]]`` | ``Box`1[[Box`1[[Payload]]]]`` | ``Box`1[[Box`1[[Payload]]]]`` |

Full asserted names use package `Issue3015.GenericParity`. Emitted assembly is loaded from bytes and `GetTypes()` is asserted before invoking its entry point.

## Fresh per-clause mutations
Numbers are true of `main` `33bea4af`, head `3abddb46`, merged tree `cc7f49ff`.

| product clause | mutant | live product proof | killed by |
|---|---|---|---|
| G# generic argument backing resolution | return `typeof(object)` | all Core/Compiler/Repl `GSharp.Core.dll` copies refreshed; md5 `4b003c65` → `183c8516`; build reached tests | `GenericTypeArguments_AgreeAcrossCompilerEmitAndInterpreter` (1 failed / 11 passed) |
| protected/protected-internal ctor forwarding | retain public ctors only | all three copies refreshed; md5 `4b003c65` → `b9c5516a`; build reached tests | `ProtectedParameterizedBaseConstructor_PreservesStateAndIdentity` (1 failed / 11 passed) |

Protected test uses an imported base whose only constructor is `protected Issue3015ProtectedParameterizedBase(int)`, so Reflection Emission's automatic parameterless constructor cannot mask the clause.

Both mutations restored to exact source and `GSharp.Core.dll` hashes. Final targeted tests: 12/12. `git merge-tree` produced `cc7f49ff8555f4abac902cd50cbc47f477450e2a`, exactly the head tree. Exact merged-tree Interpreter.Tests: 622/622.

Passing a G# instance to a CLR API expecting its imported base remains unchanged and out of scope.

Fixes #3015

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -30,6 +30,8 @@ public sealed partial class Evaluator
     private static readonly System.Runtime.CompilerServices.ConditionalWeakTable<Type, Func<Func<object[], object>, Delegate>> InterpreterDelegateFactories = new();
     private static readonly System.Runtime.CompilerServices.ConditionalWeakTable<FunctionSymbol, ConcurrentDictionary<Type, Func<Func<object[], object>, Delegate>>> InterpreterDelegateFactoriesBySite = new();
     private static readonly System.Runtime.CompilerServices.ConditionalWeakTable<Delegate, ClosureValue> InterpreterClosures = new();
+    private static readonly System.Runtime.CompilerServices.ConditionalWeakTable<StructSymbol, Type> ClrBackingTypes = new();
+    private static int clrBackingAssemblyOrdinal;
 
     private object EvaluateExpression(BoundExpression node)
     {
@@ -1170,18 +1172,21 @@ public sealed partial class Evaluator
     /// Issue #319: instantiate the imported CLR base for a GSharp class instance
     /// when the class ultimately derives from a CLR type, so inherited CLR
     /// instance state (e.g. <see cref="System.Exception.Message"/>) is observable
-    /// under the interpreter. The chosen base constructor mirrors the emit path:
-    /// the resolved <see cref="BaseConstructorInitializer.ClrConstructor"/> when
-    /// the class declared <c>: Base(args)</c>, otherwise the imported base's
-    /// accessible parameterless constructor. The base-ctor argument expressions
-    /// must be evaluated within whatever frame the caller has just pushed.
+    /// under the interpreter. Issue #3015: instantiate a runtime-derived proxy
+    /// rather than the imported base itself so inherited virtual dispatch also
+    /// observes the GSharp class's emitted type identity. The chosen base
+    /// constructor mirrors the emit path: the resolved
+    /// <see cref="BaseConstructorInitializer.ClrConstructor"/> when the class
+    /// declared <c>: Base(args)</c>, otherwise the imported base's accessible
+    /// parameterless constructor. The base-ctor argument expressions must be
+    /// evaluated within whatever frame the caller has just pushed.
     /// </summary>
     private void AllocateClrBacking(StructValue sv, StructSymbol structType, BaseConstructorInitializer activeInit)
     {
         // If an explicit `: base(args)` targets a CLR constructor, prefer that.
         if (activeInit is { IsClrBase: true } clrInit && clrInit.ClrConstructor != null)
         {
-            sv.ClrBacking = InvokeClrCtor(clrInit.ClrConstructor, clrInit.Arguments, clrInit.ArgumentRefKinds);
+            sv.ClrBacking = InvokeClrCtor(structType, clrInit.ClrConstructor, clrInit.Arguments, clrInit.ArgumentRefKinds);
             return;
         }
 
@@ -1209,7 +1214,7 @@ public sealed partial class Evaluator
                 var parameterless = ResolveParameterlessCtor(clrBase);
                 if (parameterless != null)
                 {
-                    sv.ClrBacking = parameterless.Invoke(Array.Empty<object>());
+                    sv.ClrBacking = InvokeClrBackingCtor(structType, parameterless, Array.Empty<object>());
                 }
 
                 return;
@@ -1218,13 +1223,148 @@ public sealed partial class Evaluator
     }
 
     /// <summary>Issue #319: invoke a CLR base constructor with the bound G# argument expressions.</summary>
-    private object InvokeClrCtor(ConstructorInfo ctor, ImmutableArray<BoundExpression> arguments, ImmutableArray<RefKind> argumentRefKinds)
+    private object InvokeClrCtor(StructSymbol structType, ConstructorInfo ctor, ImmutableArray<BoundExpression> arguments, ImmutableArray<RefKind> argumentRefKinds)
     {
         var args = new object[arguments.Length];
         var refSlots = BuildRefSlots(arguments, argumentRefKinds, args, ctor);
-        var instance = ctor.Invoke(args);
+        var instance = InvokeClrBackingCtor(structType, ctor, args);
         WriteBackRefSlots(refSlots, args);
         return instance;
+    }
+
+    private static object InvokeClrBackingCtor(StructSymbol structType, ConstructorInfo baseCtor, object[] args)
+    {
+        var clrBase = baseCtor.DeclaringType
+            ?? throw new InvalidOperationException("Imported base constructor has no declaring type.");
+        var backingType = ClrBackingTypes.GetValue(
+            structType,
+            _ => CreateClrBackingType(structType, clrBase));
+        var parameterTypes = Array.ConvertAll(baseCtor.GetParameters(), static parameter => parameter.ParameterType);
+        var proxyCtor = backingType.GetConstructor(parameterTypes)
+            ?? throw new InvalidOperationException($"Generated CLR backing type '{backingType.FullName}' has no matching constructor.");
+        return proxyCtor.Invoke(args);
+    }
+
+    private static Type CreateClrBackingType(StructSymbol structType, Type clrBase)
+    {
+        var assemblyOrdinal = Interlocked.Increment(ref clrBackingAssemblyOrdinal);
+        var assemblyName = new AssemblyName(string.Format(
+            System.Globalization.CultureInfo.InvariantCulture,
+            "GSharp.Interpreter.ClrBacking.{0}",
+            assemblyOrdinal));
+        var assembly = System.Reflection.Emit.AssemblyBuilder.DefineDynamicAssembly(
+            assemblyName,
+            System.Reflection.Emit.AssemblyBuilderAccess.RunAndCollect);
+        var module = assembly.DefineDynamicModule(assemblyName.Name);
+        var definition = structType.Definition ?? structType;
+        var arity = definition.TypeParameters.Length;
+        var enclosingTypes = new Stack<StructSymbol>();
+        for (var containing = definition.ContainingType as StructSymbol;
+             containing != null;
+             containing = containing.ContainingType as StructSymbol)
+        {
+            enclosingTypes.Push(containing.Definition ?? containing);
+        }
+
+        var enclosingBuilders = new List<System.Reflection.Emit.TypeBuilder>();
+        System.Reflection.Emit.TypeBuilder parentBuilder = null;
+        while (enclosingTypes.Count > 0)
+        {
+            var enclosing = enclosingTypes.Pop();
+            var enclosingName = GetClrMetadataTypeName(enclosing);
+            var enclosingBuilder = parentBuilder == null
+                ? module.DefineType(
+                    string.IsNullOrEmpty(structType.PackageName)
+                        ? enclosingName
+                        : $"{structType.PackageName}.{enclosingName}",
+                    TypeAttributes.Public | TypeAttributes.Class | TypeAttributes.Sealed,
+                    typeof(object))
+                : parentBuilder.DefineNestedType(
+                    enclosingName,
+                    TypeAttributes.NestedPublic | TypeAttributes.Class | TypeAttributes.Sealed,
+                    typeof(object));
+            if (enclosing.TypeParameters.Length > 0)
+            {
+                enclosingBuilder.DefineGenericParameters(enclosing.TypeParameters.Select(static parameter => parameter.Name).ToArray());
+            }
+
+            enclosingBuilders.Add(enclosingBuilder);
+            parentBuilder = enclosingBuilder;
+        }
+
+        var metadataName = GetClrMetadataTypeName(definition);
+        var typeBuilder = parentBuilder == null
+            ? module.DefineType(
+                string.IsNullOrEmpty(structType.PackageName)
+                    ? metadataName
+                    : $"{structType.PackageName}.{metadataName}",
+                TypeAttributes.Public | TypeAttributes.Class | TypeAttributes.Sealed,
+                clrBase)
+            : parentBuilder.DefineNestedType(
+                metadataName,
+                TypeAttributes.NestedPublic | TypeAttributes.Class | TypeAttributes.Sealed,
+                clrBase);
+        if (arity > 0)
+        {
+            typeBuilder.DefineGenericParameters(definition.TypeParameters.Select(static parameter => parameter.Name).ToArray());
+        }
+
+        foreach (var baseCtor in clrBase.GetConstructors(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic))
+        {
+            if (!baseCtor.IsPublic && !baseCtor.IsFamily && !baseCtor.IsFamilyOrAssembly)
+            {
+                continue;
+            }
+
+            var parameterTypes = Array.ConvertAll(baseCtor.GetParameters(), static parameter => parameter.ParameterType);
+            var ctorBuilder = typeBuilder.DefineConstructor(
+                MethodAttributes.Public,
+                CallingConventions.Standard,
+                parameterTypes);
+            var il = ctorBuilder.GetILGenerator();
+            il.Emit(System.Reflection.Emit.OpCodes.Ldarg_0);
+            for (var i = 0; i < parameterTypes.Length; i++)
+            {
+                il.Emit(System.Reflection.Emit.OpCodes.Ldarg, i + 1);
+            }
+
+            il.Emit(System.Reflection.Emit.OpCodes.Call, baseCtor);
+            il.Emit(System.Reflection.Emit.OpCodes.Ret);
+        }
+
+        var backingType = typeBuilder.CreateType();
+        for (var i = enclosingBuilders.Count - 1; i >= 0; i--)
+        {
+            enclosingBuilders[i].CreateType();
+        }
+
+        if (arity == 0)
+        {
+            return backingType;
+        }
+
+        var typeArguments = new Type[structType.TypeArguments.Length];
+        for (var i = 0; i < typeArguments.Length; i++)
+        {
+            var argument = structType.TypeArguments[i];
+            typeArguments[i] = argument is NullableTypeSymbol nullable
+                && nullable.UnderlyingType.ClrType is { IsValueType: true } valueClr
+                    ? typeof(Nullable<>).MakeGenericType(valueClr)
+                    : argument.ClrType ?? typeof(object);
+        }
+
+        return backingType.MakeGenericType(typeArguments);
+    }
+
+    private static string GetClrMetadataTypeName(StructSymbol type)
+    {
+        return type.TypeParameters.Length == 0
+            ? type.Name
+            : string.Format(
+                System.Globalization.CultureInfo.InvariantCulture,
+                "{0}`{1}",
+                type.Name,
+                type.TypeParameters.Length);
     }
 
     /// <summary>Issue #319: resolves the imported CLR base's accessible parameterless constructor (matches the emit path's <c>GetImportedBaseDefaultCtorReference</c>).</summary>
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -1346,14 +1346,43 @@ public sealed partial class Evaluator
         var typeArguments = new Type[structType.TypeArguments.Length];
         for (var i = 0; i < typeArguments.Length; i++)
         {
-            var argument = structType.TypeArguments[i];
-            typeArguments[i] = argument is NullableTypeSymbol nullable
-                && nullable.UnderlyingType.ClrType is { IsValueType: true } valueClr
-                    ? typeof(Nullable<>).MakeGenericType(valueClr)
-                    : argument.ClrType ?? typeof(object);
+            typeArguments[i] = ResolveClrBackingTypeArgument(structType.TypeArguments[i]);
         }
 
         return backingType.MakeGenericType(typeArguments);
+    }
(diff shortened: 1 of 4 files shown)


DavidObando:
## Rubber-duck verdict: **MERGEABLE** — fix real, proved against `gsc` oracle. Two SHOULD-FIX gaps found by mutation.

Measured `/tmp/a49` = `main` `286ed0825` + `a0aa9a13c`, merge base `fd34166a8`. Re-verified at new `main` `fdba684b6`: merge clean, build exit 0, **full interpreter suite 543/543**.

Scope: `Evaluator.Expressions.cs` +158, `Issue3015ImportedBaseIdentityTests.cs` +234. **Does NOT touch `Evaluator.Calls.cs`** (the contested file) — no collision with #3028/#3029/#3032/#3042/#3046.

### Defect and fix confirmed — `gsc` beside `gsi`, values read not status

```
package Issue3015.Probe
import System
class Sentinel : EventArgs { }
var value = Sentinel()
Console.WriteLine(value.ToString())
Console.WriteLine(value.GetType().FullName)
```

| tree | gsc compiled+run | gsi |
|---|---|---|
| `main` | Issue3015.Probe.Sentinel x2 | **System.EventArgs x2** |
| head | Issue3015.Probe.Sentinel x2 | Issue3015.Probe.Sentinel x2 |

All exit **0**. Pure silent-wrong-value class. Head reach parity with oracle. Good fix.

### Mutation, per hunk (build exit 0 + md5 change verified each; clean `GSharp.Core.dll` md5 `0cd12c2eaeb539f5a62d303dff2ff42d`, restore round-trip exact)

- **M1** — gut `InvokeClrBackingCtor` back to `baseCtor.Invoke(args)` (whole fix deleted): **5 of 8 die.** Good.
- **M2** — drop `ConditionalWeakTable` cache (`CreateClrBackingType` every call): **1 dies** (`OverloadedBaseConstructors_ShareOneDerivedRuntimeType`). Cache pinned.
- **M3** — drop nullable type-argument special case, `Evaluator.Expressions.cs:1302`: **survives full 533/533 interpreter suite.** Unpinned.

### SHOULD-FIX 1 — `OverloadedBaseConstructors_ShareOneDerivedRuntimeType` pass with entire fix deleted

Test survive **M1**. Its only identity assertion is `Object.ReferenceEquals(first.GetType(), second.GetType())` -> `True`. With fix gone, both instance are plain base type, so ReferenceEquals **trivially True**. Assertion verify "share one type", never verify type is the **derived** one — the word its name promise.

It do pin the cache (die under M2), so keep it. One line close the gap:

```
Console.WriteLine(first.GetType().FullName)   // expect Issue3015.Constructor.OverloadedSentinel
```

### SHOULD-FIX 2 — nullable type-argument branch pinned by zero tests, but real

`Evaluator.Expressions.cs:1302` special-case `NullableTypeSymbol` -> `Nullable<T>`. Delete it, **533/533 still green**. Not inert though — observable, and the `gsc` oracle prove it correct:

```
class Box[T] : EventArgs { }
Console.WriteLine(Box[int32?]().GetType().FullName)
```

- clean gsi:  Box'1[[System.Nullable'1[[System.Int32 ...]] ...]]
- **gsc oracle: identical**
- M3 gsi:     Box'1[[System.Int32 ...]]   <- wrong, diverge from gsc

(apostrophe above stand for the arity backtick.)

**Do not delete — add that probe as a test.** Existing `GenericConstruction_PreservesConstructedTypeIdentity` cover only `int32` and `string`.

### SHOULD-FIX 3 — no `gsc` oracle anywhere in the test file

`Issue3015ImportedBaseIdentityTests.cs`: `Compile` / `gsc` / `Assembly.Load` = **0 hits each** (positive control: `GetType()` = 4 hits, so the pattern do match this file). Every expectation hardcoded string. I check them by hand against `gsc` — **they correct today** — but nothing in CI catch drift, on a PR whose whole subject is gsi-vs-gsc parity. Add one compiled-side assertion, or state why interpreter-only acceptable.

### Measured and CLEARED — sealed proxy do not break `is`/`as`

Proxy type emit `TypeAttributes.Sealed` with no virtual override, so I probe G# class over G# class over CLR base:

```
open class A : EventArgs { }
class B : A { }
var b = B()
// FullName, b is A, b is EventArgs
```

gsi -> `Issue3015.Inh.B|True|True`. gsc -> **identical**. Hypothesis disproved, no finding. (Both also agree on `GS0181` when `open` omitted.)

### NIT

One dynamic assembly per cache miss (`CreateClrBackingType`, `Interlocked.Increment(ref clrBackingAssemblyOrdinal)`). Bounded per struct symbol by the cache, so fine for compile-and-run; worth a thought only for a long-lived REPL session.

DavidObando:
Reviewer fixes complete at `b11e26dc28b6d8e7ef5522efb0ef245f4f04015c`.

Content changes:
- `OverloadedBaseConstructors_ShareOneDerivedRuntimeType` now asserts `Issue3015.Constructor.OverloadedSentinel`; full production revert reports the imported base instead.
- Added nullable generic proxy pin requiring `System.Nullable\`1`.
- Added real `gsc` compilation through `GSharp.Compiler.Program.Main`, then `Assembly.Load(File.ReadAllBytes(...)).GetTypes()`, with exact gsi/gsc type-name agreement.
- Updated #2990's reflective constructor-boundary guard for the new derived-proxy argument.

Mutation results (all mutant builds rc=0):

| Product mutation | Core hash | Killed by |
|---|---|---|
| Full proxy fix reverted | `fb7c1d09` | identity-name test + gsc/gsi agreement test |
| Explicit constructor route bypassed | `d49418dc` | `ExplicitImportedBaseConstructor_PreservesStateAndIdentity` |
| Default constructor route bypassed | `a6ef4fe4` | `CompilerAndInterpreter_AgreeOnDerivedRuntimeType` |
| Proxy cache removed | `fdb1d853` | `OverloadedBaseConstructors_ShareOneDerivedRuntimeType` |
| Nullable mapping removed | `e81e7a4d` | `NullableGenericConstruction_PreservesNullableTypeArgument` |

Restored Core hash round-tripped to `fa6020e2` after every mutant. Final production/test blobs (`444099798a`, `082572d3b8`) exactly match the mutation-tested blobs after later unrelated main rebases.

Validation: Release build 0 warnings/errors; targeted 11/11; bidirectional docs gate 1/1; full solution **14,216 passed / 0 failed / 14,217 total** (1 skipped). Latest-head build and targeted/docs probes rerun after final rebase.

DavidObando:
## Review: PR #3049 — `fix/3015-gsi-imported-base-identity`

**Verdict: MERGEABLE** (2 SHOULD-FIX, 2 NIT — no blockers, no regressions found)

Numbers true of `main` **`5ad0109c`**, head **`b11e26dc2`**, computed merge base **`1ea847eb`**, three-dot content **4 files, +476 / −12**. `git merge-tree` → 0 conflicts, merged tree **`4eb5656b`**; I built that exact tree (`8167b252`) and confirmed the tree SHA matches the prediction. Merged build Release **0 warnings / 0 errors**; freshness verified on all three `GSharp.Core.dll` copies — `Core/`, `Compiler/` **and `Repl/`** — identical md5 `52a427d9`, mtime 2 min before first probe.

`main` moved the same file (`Evaluator.Expressions.cs`) via **#3020**. **No interaction found**: the merged tree builds clean and is green, and every probe below ran on the merged tree, not the branch.

### Test shape (asked for explicitly)
`Issue3015ImportedBaseIdentityTests.cs` contains **0 occurrences of `func Main`** — all 10 cases drive at **top level**, the only axis `gsi` executes. **Not vacuous.** Assertions are on `GetType().FullName` / `ToString()`, i.e. genuine metadata, not incidental values. This is the right shape and I want to say so plainly.

### Verified correct (probe evidence)
| # | Claim | Result |
|---|---|---|
| 1 | Issue fixed | `Args().ToString()` → `Probe.P1.Args`; **main → `System.EventArgs`** |
| 2 | **Backing type per CLASS, not per instance** | `a.GetType() == b.GetType()` → **`True`**; CWT keyed by `StructSymbol` |
| 3 | Collectible | `AssemblyBuilderAccess.RunAndCollect` ✔ |
| 4 | Nested inside a **generic** outer (untested by the PR) | works → `Probe.P3.Outer\`1+Inner` |
| 5 | Inherited members beyond `CanRead` | `CanRead`, `WriteByte`/`Length`, **`Position` setter**, `ReadByte`→`11`, `as Stream`→`True` all correct |
| 6 | `params` base ctor | `class Agg : AggregateException` → `Probe.P7.Agg` |
| 7 | Explicit overloaded base ctor | `init() : base(33)` → `Capacity` = **33** |
| 8 | **`protected` + abstract** base | `class MyAttr : Attribute` → `Probe.P8.MyAttr` / base `System.Attribute` |
| 9 | Unfiltered suite | Interpreter.Tests **617 / 617**, 0 failed (author filtered to 2 classes) |

The accessibility model is **right**, not merely lucky: `IsPublic || IsFamily || IsFamilyOrAssembly` forwards public/protected/protected-internal and excludes `internal` and `private protected` — which genuinely *are* inaccessible to a derived type in another assembly.

I also spent a build disproving my own strongest hypothesis: `il.Emit(OpCodes.Ldarg, i + 1)` passes an `int` to an `InlineVar` opcode, which looks like a 4-byte/2-byte operand corruption. A standalone Reflection.Emit harness forwarding 0-, 1- and 2-arg ctors both ways shows `ILGenerator.Emit(OpCode, int)` **special-cases `Ldarg` and shortens the operand**. Not a bug. Recording it because a plausible mechanism is a fabrication until a probe fires it.

---

### SHOULD-FIX 1 — G# type argument silently erased to `System.Object`; `gsc` and `gsi` disagree
`src/Core/CodeAnalysis/Evaluator.Expressions.cs`, in `CreateClrBackingType`: `typeArguments[i] = ... : argument.ClrType ?? typeof(object);`

An interpreted G# type has no `ClrType`, so it falls back to `object`.

```gsharp
package Probe.P4
import System
class Payload { }
class Box[T] : EventArgs { }
Console.WriteLine(Box[int32]().GetType().FullName)
Console.WriteLine(Box[Payload]().GetType().FullName)
```

| driver | `Box[Payload]` |
|---|---|
| `gsc /out:` + `dotnet exec` (**emit**) | `Probe.P4.Box\`1[[Probe.P4.Payload, Probe.P4, …]]` |
| `gsi` | `Probe.P4.Box\`1[[System.Object, System.Private.CoreLib, …]]` |
| `main` (`gsi`) | `System.EventArgs` |

`Box[int32]` is correct on head, so this is specifically the **G#-typed** type argument. **Not a regression** — main was worse. But it is issue #3015's own acceptance criterion ("must … match compiled behavior") unmet, in a category the PR body claims to cover ("generic and nested classes"), and it is a **silent wrong answer at rc=0**. Note bare `gsc` evaluates and therefore agrees with `gsi`; only the `/out:` emit column discriminates — the axis the gate does not traverse.

### SHOULD-FIX 2 — the ctor accessibility clause is unpinned (PRODUCT mutant survives 617/617)
Mutant **M3**, same file: `if (!baseCtor.IsPublic && !baseCtor.IsFamily && !baseCtor.IsFamilyOrAssembly)` → `if (!baseCtor.IsPublic)` (drops protected / protected-internal forwarding).

Build **0/0** (so the mutant is live, not a broken build), `Repl/GSharp.Core.dll` md5 `52a427d9` → **`caefc463`**, and **Interpreter.Tests 617 / 617 green**. My own `System.Attribute` probe also survives.

Mechanism, which is the useful part: `TypeBuilder.CreateType()` **auto-generates a default parameterless ctor** when none was defined, and that chains to the base's protected parameterless ctor. So the clause only discriminates for a **protected ctor with parameters**, and nothing in the suite uses one. Add a case deriving from a base whose only accessible ctor is protected-and-parameterised, or the clause can be deleted without any test noticing.

### NIT 1 — stale denominator in the PR body
The body's mutation control reports "failed 4/8 issue tests". The file declares **10** `[Fact]` methods. Whichever number is right, they disagree, and "4 of 8 failed" means 4+ survived a whole-fix revert — please state per-hunk which.

### NIT 2 — pre-existing, but this PR makes it cheap to fix
Passing a G#-derived instance **as an argument** to a CLR API expecting the imported base still fails:

```gsharp
var w = StreamWriter(MyStream())
// GS9999: Object of type 'GSharp.Core.CodeAnalysis.StructValue' cannot be converted to type 'System.IO.Stream'.
```

**Identical on `main` and on head**, so it is explicitly *not* a regression from this PR and not a blocker. Flagging it because the receiver bridge is unwrapped but the *argument* path is not — and now that `sv.ClrBacking` is a genuine `Stream` subclass rather than a bare base instance, unwrapping at the argument boundary would just work. (`src.CopyTo(dst)` → `22` correct, so the receiver direction is fine.)

---

**Hygiene:** every mutation restored with an exact md5 round-trip (`2672ee6d96f583bb5686b05555b00fbe` → `2672ee6d96f583bb5686b05555b00fbe`), `git status --porcelain` empty. All work in a private scratch clone; no worktree touched, no e2e, `~/.nuget/packages/gsharp.net.sdk/` never written. Every probe compiled into its own empty directory.

DavidObando:
## Rubber-duck verdict: **MERGEABLE** — #3015 criterion traced clause-by-clause and **met**. One adjacent cell diverges (SHOULD-FIX), two NITs, **zero regressions across 122 samples**.

Numbers true of `main` **`4f7559dd`**, head **`3abddb46`**, computed merge base **`33bea4af`**, three-dot `--full-index` scope **4 files, +656 / −12**. Head is **3 behind** main, so head ≠ merged tree; every probe below ran on the **co-merged** tree, never the branch.

`git merge-tree` predicted tree `827e9427ff5830e5bb2f9c3918e8603321a82262`; real merge produced **exactly** `827e9427`. **`BUILD_RC=0`, 0 warnings / 0 errors.** Freshness by mtime on all three `GSharp.Core.dll` copies — `Core/`, `Compiler/` **and `Repl/`** (gsi live there) — one mtime `06:42:50`, md5 `701fb782`.

---

## 1. Acceptance criterion, traced clause to `file:line`

#3015 "## Expected", literal sentence:

> *"Inherited virtual calls under `gsi` must preserve the G# derived type's identity, matching compiled behavior, for both ordinary and literal construction."*

| # | clause | traced to | status |
|---|---|---|---|
| 1 | "Inherited virtual calls under `gsi`" | `Issue3015ImportedBaseIdentityTests.cs:26,46` (`ToString()` on `EventArgs`), `:66,83` (`CanRead` on `MemoryStream`) | MET |
| 2 | "preserve the G# derived type's identity" | head gsi → `Default.Args`; base gsi → `System.EventArgs` | MET |
| 3 | "matching compiled behavior" | `:192` `CompilerAndInterpreter_AgreeOnDerivedRuntimeType`, `:265` three-driver parity; my emit oracle agrees byte-for-byte | MET (one gap, §3) |
| 4 | "for **both** ordinary and literal construction" | `:26` ordinary + `:46` literal; `:66`/`:83` dispatch both ways; `:240` nested both ways | MET |

**Every clause point at real line. Criterion met.** Doubt about clause 4 not hold up — literal construction have own test at `:46`, own dispatch test at `:83`, own nested test at `:255`, and shipped sample use literal form.

## 2. Repro run **byte-for-byte first**, before any variation

Exact code from issue body, no package line, no edit:

```gsharp
import System

class Args : EventArgs {
}

Console.WriteLine(Args().ToString())
```

| driver | base `33bea4af` | merged |
|---|---|---|
| `gsi` | **`System.EventArgs`** | **`Default.Args`** |
| bare `gsc` (evaluates) | — | `Default.Args` |
| `gsc /out:` + **run** (emit oracle) | `Default.Args` | `Default.Args` |

Base column reproduce #3015 exactly, so probe **discriminate**. Literal variant `Args{}` — same table, same result.

**Strongest evidence, and it use the golden oracle:** shipped `samples/ImportedBaseClass.gs` construct with **`Buffer{}` / `Args{}`** — literal — and its `.golden` expect `GSharp.Example.ImportedBaseClass.Args`.

- base `gsi` vs golden → **diverge** (`System.EventArgs`)
- head `gsi` vs golden → **match exactly**

That string is the one #3015 quote. Reported shape, not neighbour.

## 3. SHOULD-FIX — G# **struct** and **enum** generic argument still erase to `System.Object`

`src/Core/CodeAnalysis/Evaluator.Expressions.cs:1368` filter `argument is StructSymbol { IsClass: true }`, so struct and enum fall through to `:1385` `return typeof(object)`.

```gsharp
package P4
import System
struct Pay { }
class Box[T] : EventArgs { }
Console.WriteLine(Box[Pay]().GetType().FullName)
```

| type argument | `gsi` (merged) | `gsc /out:` + run |
|---|---|---|
| G# **struct** `Pay` | ``P4.Box`1[[System.Object, …]]`` | ``P4.Box`1[[P4.Pay, P4, …]]`` |
| G# **enum** `Color` | ``P5.Box`1[[System.Object, …]]`` | ``P5.Box`1[[P5.Color, P5, …]]`` |
| G# class `Payload` | agree | agree |
| `int32`, `string`, nested | agree | agree |

**Not a regression** — base gsi give `System.EventArgs`, worse. Silent wrong value at rc=0 though, and it is criterion clause 3 ("matching compiled behavior") unmet in that cell.

Why nobody see it: parity test at `:265` cover exactly `Payload` (class), `string` (CLR), `Box[Payload]` (nested class). Type-argument **kind** never parameterised. Suite stay green — Interpreter 633/633, Core 7074/0/1 — while gsi and gsc disagree. Full green pin nothing.

**PR body third bullet is factually wrong today**: "reify G#-declared generic arguments … instead of erasing them to `System.Object`". Struct and enum erase to precisely `System.Object`. Fix that sentence, add two rows to `:265`, or file follow-up. Real fix not one line (struct need value-type semantics, enum need `EnumBuilder`), so blocking on it is scope creep — but the claim must stop.

## 4. False-positive corpus — **zero regressions**

Every `samples/*.gs` through **real `gsi`**, base vs merged, compared on rc **and** full output:

```
samples_swept=122      total_differing=1
--- ImportedBaseClass ---
6c6
< System.EventArgs
> GSharp.Example.ImportedBaseClass.Args
```

One difference, and it is the fix. No program that ran stopped running; no rc moved. Inherited CLR state hold: `class E(msg string) : Exception(msg)` → `Message`=`boom-22`; `class Buf : MemoryStream` → `WriteByte`/`Position`/`ReadByte`=`33`, `Length`=`1`; all identical to emit.

Generic identity share correctly too: two `Box[int32]()` → `ReferenceEquals(a.GetType(), b.GetType())` = `True`, same as compiled.

## 5. Interaction with `main`, proved behaviourally not textually

`main` move 4 commits since merge base — **#3074 (explicit constructor accessibility emission)**, #3072, #2963, #3036. Zero file overlap, but #3074 is semantically adjacent: this PR filter base ctors by accessibility. Zero overlap is hypothesis, so I co-merge and run both suite together on merged tree:

- Interpreter.Tests **633 / 633** (author measured 622 on older merged tree; `main` add 11 interpreter tests — delta reconcile, so run was real)
- Core.Tests **7074 passed / 0 failed / 1 skipped**
- predicted tree hash reproduced exactly, `BUILD_RC=0`

Compiler.Tests not run (~1h20m, PR touch no emit path; main's emit tests already green in its own CI). Stating the filter.

## 6. NIT — CLR-boundary virtual dispatch unchanged (author already scoped this out, correctly)

```gsharp
class Args : EventArgs { override func ToString() string { return "custom-11" } }
var a = Args()
Console.WriteLine(a.ToString())   // gsi custom-11 | emit custom-11
Console.WriteLine(a)              // gsi Args()    | emit custom-11
```

Proxy emit no virtual override, so CLR dispatch not see G# override. **Identical on base** → pre-existing, not this PR. PR body already say "passing a G# instance to a CLR API expecting its imported base remains unchanged and out of scope" — that statement is **true as measured**. No action. Recording only to bound what the fix reach.

## 7. NIT — test project now reference the compiler

`test/Interpreter.Tests/Interpreter.Tests.csproj` add `<ProjectReference Include="@(GsharpCompiler)" />`. Justified — parity test at `:265` drive `GSharp.Compiler.Program.Main` and `GSharp.Repl.Program.Main` in one test, which is genuinely the right shape. But it couple interpreter tests to compiler. Worth one comment saying why.

## 8. Two briefing claims I checked and must retract

- **`/target:library` + `/r:` cross-assembly test is not reachable.** `gsi --help` → `Usage: gsi [file.gs] [--help] [--version]`. `gsi /r:lib.dll x.gs` → `Unrecognized argument`. `gsi` have **no reference option**. #3015 "imported base" mean BCL/CLR type from default reference set. Single-file **is** the imported path here. I built the library anyway to check; gsi cannot consume it.
- **`src/Repl/Program.cs` is not in this diff** (0 hits in file list). Risk of silently dropping the #3024 file path from gsi diagnostics **does not apply** to #3049.

## 9. What I did not do

Did not re-run mutants. Author table (5 clauses) plus earlier round (M1/M2/M3) already cover the hunks, and the accessibility clause was specifically closed by `ProtectedParameterizedBaseConstructor_PreservesStateAndIdentity` using a **parameterised** protected ctor, which defeat the `CreateType()` auto-default-ctor mask. I verified each named killing test exist. My findings do not rest on a pinning claim — the struct/enum gap is proved by measured divergence against a green suite.

Could not construct a case where explicit `: base(args)` resolve a ctor the proxy skip: `ResolveParameterlessCtor` (`:1400-1412`) and the proxy forwarder use the **identical** filter `IsPublic || IsFamily || IsFamilyOrAssembly`. Hypothesis dead, disclosing rather than reporting.

Issue #1718, opened 2026-07-02
Interpreter: concurrent writes to a captured enclosing-frame Dictionary from a goroutine can corrupt the frame (non-thread-safe Dictionary)


DavidObando:
- **Category**: correctness / concurrency
- **Severity**: needs-triage (follow-up to #1651)
- **Location**: src/Core/CodeAnalysis/Evaluator.cs — per-thread `ExecutionState.Locals` frame dictionaries (`Dictionary<VariableSymbol, object>`), goroutine locals clone in EvaluateGoStatement/EvaluateScopeStatement.

**Context**: #1651 (PR #1717) isolated per-execution interpreter state per thread and clones a goroutine's locals chain **by reference** (new `Stack` container, shared frame-dict references) to preserve Go-style by-reference closure capture. That correctly fixes the frame-corruption and control-flow-leak races.

**Residual hazard**: because the top (and any enclosing) frame `Dictionary` instances are SHARED between the spawning thread and the goroutine, if user code assigns a **captured enclosing-scope variable** from inside a goroutine (`x = ...` where `x` lives in an outer frame) while the parent thread also writes into that same frame dictionary, two threads mutate the same non-thread-safe `Dictionary<,>` concurrently. That can corrupt the dictionary's buckets → crash (or torn state), which is HARSHER than Go's own semantics: Go calls this a user data race with defined-ish memory behavior, whereas here it can crash the interpreter. In practice most `go` bodies descend into a `PushFrame` before assigning, so the shared top frame usually isn't the write target — but captured-outer-variable assignment still races.

**Suggested fix** (pick one):
- Use `ConcurrentDictionary<VariableSymbol, object>` for frame storage (simplest, removes the crash class), or
- Guard frame-dictionary writes with a per-frame lock, or
- Explicitly document the limitation and detect/diagnose concurrent captured-variable writes.
Add an interpreter stress test that has a goroutine and its parent both write a shared captured outer variable and asserts no dictionary corruption/crash.

Surfaced during the #1651 review (PR #1717). Confidence: high that the hazard exists; low frequency in typical code. Pre-existing class of problem, not a regression from #1651.

PR #3052, opened 2026-08-01
Warn when gsi skips deinitializers

DavidObando:
## Summary

- report warning `GS0510` once per newly declared class whose `deinit` cannot run under interpreted execution
- suppress the boundary warning when compilation errors prevent execution
- keep interpreter execution non-fatal while preserving CLR finalization for emitted programs
- document measured GC behavior and the Warning-vs-Error rationale in ADR-0068 and both diagnostics catalogues

Addresses #2988. This PR diagnoses the interpreter boundary; it does **not** make deinitializers run in the interpreter.

## Design

`deinit` lowers to a CLR `Finalize` override. The interpreter emits no finalizable CLR class. Scope-exit or shutdown execution would diverge from compiled behavior; matching it requires GC-coupled `StructValue` lifetime and evaluator re-entry from a finalizer thread.

| Probe | `gsc` | `gsi` |
|---|---|---|
| object dies in function; collection forced | `made-22` / `deinit-11` / `body-33` | `made-22` / `body-33` |
| object dies in function; no forced collection | `made-22` / `body-33` | identical |
| object remains reachable at exit | no `deinit` | no `deinit` |

`GS0510` is a Warning because CLR finalizers are not guaranteed to run and the program can otherwise complete. Unsafe storage and P/Invoke boundaries are Errors because interpreted execution cannot proceed correctly or would fabricate a deterministic result.

## Rebase and interaction

Rebased onto actual `main` `4469e3237a30`; head `a18cc2b13431`. The duplicate `src/Repl/Program.cs` change is absent from this PR. Main's `engine.Evaluate(File.ReadAllText(arg), arg)` remains intact and direct `gsi` diagnostics include the source path.

`git merge-tree --write-tree origin/main HEAD` returned rc 0 (`73fc33da9ed2cb30baf9f196d5a862f758b31815`), and the merged solution build then passed with 0 warnings and 0 errors.

Cross-feature probe from a fresh solution build and a newly created empty emit directory:

| Driver | rc | `GS0514` | `GS0510` | Result |
|---|---:|---:|---:|---|
| `gsi` | 1 | 1 | 1 | both located diagnostics include `both.gs` |
| bare `gsc` | 1 | 1 | 1 | both evaluation boundaries fire |
| `gsc /out:` + run | 0 / 0 | 0 | 0 | emitted program prints `3` |

All `Core/`, `Compiler/`, and `Repl/` `GSharp.Core.dll` copies had the same post-build mtime before the probe.

## Tests

- `GSharp.sln` Release build: 0 warnings, 0 errors
- `Interpreter.Tests`: 718/718 passed
- `Core.Tests`: 7089 passed, 1 skipped, 0 failed
- bidirectional diagnostics gate: 4/4 passed after mutation round-trip
- exactly one `| GS0510 | Warning |` row in each catalogue; `GS0517` absent from source
- all issue sources execute top-level code; none uses a `func Main()` wrapper

## Mutation evidence

Every mutation was asserted present before execution and restored afterward.

| Product mutation | Kill |
|---|---|
| `deinitDiagnostics` → `interpreterDiagnostics` | solution build fails with `CS0128` at the merged #3024 local |
| `ReferenceEquals(...)` → `deinitializer is not null` | `DeinitializersWarnOncePerDeclaringClassWithoutRunning` fails: 3 warnings instead of 2 |
| invert `!allErrors.Any()` | `CompilationErrorsSuppressDeinitializerBoundaryWarning` fails |
| invert `SyntaxTrees.Contains(...)` | `DeinitializersWarnOncePerDeclaringClassWithoutRunning` fails: 0 warnings |
| omit `.Concat(deinitDiagnostics)` | `DeinitializersWarnOncePerDeclaringClassWithoutRunning` fails: 0 warnings |
| route report helper to the P/Invoke descriptor | `DeinitializersWarnOncePerDeclaringClassWithoutRunning` fails: 0 `GS0510` warnings |
| descriptor `GS0510` → `GS0517` | `Every_Diagnostic_Is_Documented_In_Both_References_With_Matching_Severity` fails: missing `GS0517`, unexpected `GS0510` |

A required-parameter compiler mutation found one unique report site (`Compilation.cs`); no ad-hoc second caller exists.


DavidObando:
## Verdict: MERGEABLE

Reviewed at head `dc3b1e574` against `main` = `fdba684b6`. Scope from a computed merge base (`f820484c3`), never two-dot: **8 files, +197/-2**, matching `gh pr view` exactly. `git merge-tree --write-tree fdba684b6 dc3b1e574` → **rc=0**. **2 commits behind** (not 4 — I pin to the explicit SHA rather than `origin/main`, which is shared mutable state across the parallel agents here).

Build: `gsc` **rc=0, 0W/0E**; `gsi` **rc=0, 0W/0E**. Full `Interpreter.Tests` **538/538, rc=0**.

### The structural claim holds — this is a legitimate (b), not deferred work

The justification offered for a boundary diagnostic rather than a fix was "no emitted CLR finalizable type; scope cleanup would be incorrect". I tested the claim rather than the prose, and it survives at the #2987 bar. Two independent structural blockers, both measured:

1. **No reachability information exists.** Interpreter frames are `ConcurrentDictionary<VariableSymbol, object>` (`Evaluator.Calls.cs:67,297,333,396,550`). Popping a frame says nothing about whether a value is still reachable — it may have been returned, stored in a field, or captured. Running `deinit` at scope exit would finalize live objects. That is incorrect, not merely hard.
2. **The instance representation is overloaded, so the obvious alternative is also unsound.** `StructValue` is a single `sealed class` (`StructValue.cs:18`) used for *three* different things: value-type struct values with Go-style copy semantics (`Copy()` on assignment), reference-type class instances shared by reference, and the synthetic heap box cells `CaptureBoxingRewriter` creates for captured mutable variables (`StructValue.cs:34-45`, issue #1718). Attaching a CLR finalizer to `StructValue` to borrow the host GC would therefore fire once per *copy* of a value struct and once per *capture box* — neither of which is a user object with a deinitializer.

A correct implementation would need per-instance finalization sentinels plus running interpreted G# on the finalizer thread, and would still deliver non-deterministic, unordered, frequently-never-running finalization. That is a different feature, not this bug. **(b) is the right shape here.**

Worth recording: I ran the two-class scenario under `gsc` and got

```
Allocated: db/users
Resource deinit: db
CachedResource deinit: users
done
```

which is the **reverse** deinitializer order from the one quoted in my review brief. CLR finalizer ordering is unspecified, and this is direct evidence of that. So "the interpreter is missing *ordered* semantics" is not a strong argument against a warning — `gsc` does not offer an order contract either.

### Measured divergence

| | stdout | exit |
|---|---|---|
| `gsc` | `Allocated: db/users` / `Resource deinit: db` / `CachedResource deinit: users` / `done` | 0 |
| `gsi` @ head | `Allocated: db/users` / `done` | **0** + 2× `GS0517` on stderr |

Exit code is governed by `src/Repl/Program.cs:69` (`return cell.HasError ? 1 : 0`), untouched by this PR — so the severity choice *is* the exit code. `Issue2988DeinitInterpreterTests.cs:120` pins `Assert.Equal(0, exitCode)` deliberately.

### Mutation: 6/6 killed, no survivors

Every mutant built with **0 errors** and moved the hashed artifact, so each was live in both directions.

| # | mutation | result |
|---|---|---|
| M1 | backtick the ID in `docs/diagnostics.md` | **fails**: `Missing from docs/diagnostics.md: GS0517` |
| M2 | backtick the ID in `website/docs/ref/diagnostics.md` | **fails**: `Missing from website/docs/ref/diagnostics.md: GS0517` |
| M3 | `Warning` → `Error` in `docs/diagnostics.md` | **fails**: `GS0517: expected Warning, found Error` |
| M4 | drop `.Concat(interpreterDiagnostics)`, `Compilation.cs:318` | **3/3 tests die** |
| M5 | revert renderer to `$"{d.Id}: {d.Message}"`, `Program.cs:59-62` | **1/3 dies** — `ScriptRunnerUsesRichRendererForBoundaryWarning` |
| M6 | drop the `&& SyntaxTrees.Contains(...)` conjunct, `Compilation.cs:302` | **1/3 dies** — `DeinitializersWarnOncePerDeclaringClassWithoutRunning` |

Baseline 3/3 green, `GSharp.Core.dll` md5 `38e5996efab5d38a179c9e6fc140a07f`; restored → md5 identical and 3/3 green again.

**Both self-flagged weak spots are refuted.** "Renderer 1/3" is *sufficient* — one killing test is a pin. "Both docs rows 1/4 each" understates it: I mutated the rows directly and the gate caught the ID and the severity independently, on both files. The rows are load-bearing, and critically they use bare `| GS0517 |` — a backticked ID would have been invisible to `DiagnosticIdUniquenessTests.cs:206` and gone green on the branch while reddening `main`.

Note the gate regex on `main` is now `^\|\s*(GS\d{4})\s*\|\s*([^|]+?)\s*\|` with severity normalised separately — more permissive on the severity cell than the older `(?:\*\*)?(Error|Warning|Info)` form, but still fatal to a backticked ID. `main`ʼs 2 commits ahead touch none of the three gate inputs (positive-controlled: 5 files changed main-side, none of them these), so the head-state gate result is valid for the landing state.

### Edge cases probed, both clean

- **`deinit` on a struct** — rejected at bind time: `GS0289: deinit is only valid on a class type — SVal is a struct.` So the descriptorʼs hardcoded `class {0}` is correct *by construction*, and the `as StructSymbol` cast cannot mislabel a struct.
- **Class declared but never instantiated** — still warns, exit 0. Defensible: the warning is about the declaration, and "once per declaring class" is the documented contract.

### SHOULD-FIX 1 — the severity is correct but never argued

`DiagnosticDescriptors.cs:382` chooses `Warning`. The only sibling interpreter-boundary diagnostic on `main` is `DiagnosticDescriptors.cs:381` `GS0513`, which is **`Error`** → exit 1. The ADR paragraph states the choice but gives no reason for the asymmetry, and the next boundary diagnostic will copy whichever precedent is nearest.

I think `Warning` is right, and the distinction is worth one sentence: **`GS0513` fires when the construct cannot be evaluated at all — the program cannot proceed. `GS0517` fires when the program runs to completion and only a deferred, GC-scheduled side-effect is skipped — a side-effect the CLR does not guarantee will run under `gsc` either.** Please put that (or your own version) in `docs/adr/0068-deinit-destructor-support.md` next to the boundary paragraph, so the exit-0 choice is a recorded decision rather than a default.

### SHOULD-FIX 2 — undeclared scope: the renderer change is global

`src/Repl/Program.cs:59-62` replaces the per-diagnostic `Console.Error.WriteLine($"{d.Id}: {d.Message}")` loop with `Console.Error.WriteDiagnostics(...)`. That changes stderr for **every** `gsi` diagnostic, not just `GS0517`. Measured — `gsi` now emits

```
(5,5,5,11): warning GS0517: Deinitializer of class Resource will not run under the interpreter; ...
        deinit {
```

with a source location and an echoed source line, where before it printed neither. This is a genuine improvement and I am not asking you to drop it; `Interpreter.Tests` is 538/538 with it. But it is outside "closes #2988", and it invalidates any existing note that `gsi` never prints a source location. Please call it out in the PR body so it is not discovered later by whoever owns that documentation.

### NIT

`Compilation.cs:298` iterates `program.Functions.Keys` and re-derives the deinitializer per function via `ReferenceEquals(deinitializer?.Function, function)`. Iterating declared types directly would express "once per declaring class" without depending on the deinitializer body being present in `Functions`. Behaviourally fine today — M6 shows the current form is pinned — so this is style, not a defect.

DavidObando:
## Verdict: NOT MERGEABLE — 1 BLOCKER (diagnostic ID), 3 SHOULD-FIX, 2 NIT

**Trees.** `main` = **`5ad0109c`**; PR head = **`a71db53f5`**; computed merge base = `185431c4` (PR is **8 commits behind**). Three-dot scope: **9 files, +201 / −5**. Merged tree measured = **`a4cdee748`**, `git merge` rc=0, solution build **0 warnings / 0 errors**. All three `GSharp.Core.dll` copies (`Core/`, `Compiler/`, `Repl/`) verified mtime-identical and md5-identical before every probe. Each `gsc` probe was compiled into its own empty output directory.

Note the PR body's claim that "merged tree exactly equals the rebased head tree" is no longer true — it was measured against `185431c4`, and `main` has since advanced 8 commits.

---

## Q1 first: "warn" is the RIGHT resolution, and it is not deferral. I can show it.

It is also not this PR's invention — it is the recorded decision in issue #2988 comment 2 of 4 ("**G5 policy decision: explicit boundary + diagnostic. Do not implement deinit in the interpreter**"), which explicitly supersedes the issue body's "Note on scope". More usefully, I re-derived the technical argument from the compiled oracle rather than taking it on trust. Three probes, `gsc` + `dotnet` vs `gsi`:

| probe | shape | `gsc` output | `gsi` output |
|---|---|---|---|
| f3 | object dies in a function, **`GC.Collect()` + `WaitForPendingFinalizers()` forced** | `made-22` / **`deinit-11`** / `body-33` | `made-22` / `body-33` |
| f7 | object dies in a function, **no forced collect** | `made-22` / `body-33` | `made-22` / `body-33` |
| f2 | object **still reachable** at program exit | `body-33` (no deinit) | `body-33` |

This kills both cheap implementations:

- **Scope exit is wrong.** f2 shows a still-reachable object is never finalized by `gsc`. A scope-exit interpreter would run user cleanup on a live object.
- **End-of-program is also wrong.** f7 is the decisive one: when nothing forces a collection, `gsc` **never runs the deinitializer at all** and its output is byte-identical to `gsi`'s. An "run pending deinits at shutdown" implementation would print `deinit-11` where the compiled program prints nothing — *more* divergent than today, not less.

Only real GC scheduling reproduces `gsc`. That means tying finalization to `StructValue` lifetime and re-entering `Evaluator` from the finalizer thread, and its observable output is timing-dependent — so it could not be pinned by the golden-sample comparison this repo uses. That is a lifetime-and-threading model, not a feature-sized change. **The boundary is a real engineering answer, not a papering-over.**

**One correction to the review brief I was given.** It asked that the warning "must fire exactly when a deinitializer is actually skipped — not when a type merely declares one that never gets instantiated." That contradicts the settled design in the issue ("report **once per declaring class**, at bind time … **not per instantiation**"), and the design is right: whether any given instance is collected is GC-timing-dependent, so "was it actually skipped" is not decidable at report time. The PR follows the recorded decision. I am not filing that as a defect.

## What is correct — measured, not assumed

- **The warning is `gsi`-only.** `gsc` emitted **GS0517 × 0** on every probe (f1, f2, f3, f7), rc=0. No escalation path into the compiler.
- **No false positive without `deinit`.** f5 (class, no deinit): **GS0517 × 0**.
- **Exactly once per class even with other members.** f8 (`deinit` + two methods + a field): **exactly 1** warning.
- **Docs gate green on the merged tree**: `DiagnosticIdUniquenessTests` **4/4**. Both rows present in `docs/diagnostics.md:1162` and `website/docs/ref/diagnostics.md:1768`, **bare** `| GS0517 | Warning |` — not backticked, severity matching the descriptor.
- **No ADR collision.** `docs/adr/0068-deinit-destructor-support.md` already exists on `main`; this is a +9-line amendment, not a new number.
- **The tests are not `func Main()`-wrapped.** All three cases in `Issue2988DeinitInterpreterTests.cs` are top-level, with distinct values (`deinit-11`/`22`, `body-33`, `next-44`). I confirmed the vacuous shape is real — my f6 (`func Main()`-only) runs under `gsi` with rc=0 and **completely empty output** — so this mattered, and the author avoided it.
- **Suites on the merged tree**: `Interpreter.Tests` **0 failed / 610 passed / 610**; `Core.Tests` **0 failed / 7072 passed / 1 skipped / 7073**.

---

## BLOCKER 1 — `src/Core/CodeAnalysis/DiagnosticDescriptors.cs:382`: the PR self-allocated `GS0517`; the allocated ID is `GS0510`

Issue #2988 comment 3 of 4 is titled "**Correction to my G5 design: `GS0507` is not safe — use `GS0510`**", and comment 4 of 4 re-states it in a table: `GS0510`, severity **Warning**, for the `deinit` boundary. That is two explicit, written allocations of `GS0510` for exactly this diagnostic.

The PR ships `GS0517`. Evidence that this was never issued:

- `GS0517` appears **18 times** in the PR body and thread; `GS0510` appears **0 times**. No allocation of `GS0517` is recorded anywhere in either thread.
- `git log -S"GS0517" --all` → **6 commits, all of them this branch's own** (`a71db53f`, `76a6a770`, `2c258a35`, `8edb37cb`, `fea35643`, `9451acba`).
- `git log -S"GS0510" --all` → **zero commits in all history**. `GS0510` is still completely free, exactly as it was when allocated.
- Positive control: 377 descriptors match `new("GS####"` on `main`, so the search is sound.

The sibling PR **#3024**, same author, now correctly uses **`GS0514`** — the ID that was issued to it — after previously using an unallocated one. So allocations are being honoured when raised; this one simply was never raised. The prior review of this PR did not check it.

**Fix:** rename `GS0517` → `GS0510` in five places (descriptor, both diagnostics tables, the ADR paragraph, the test's `Id ==` filter). Alternatively the allocator ratifies `GS0517` — but that decision is not this PR's to make, which is why it is a blocker rather than a nit.

## SHOULD-FIX 1 — `src/Core/CodeAnalysis/Compilation/Compilation.cs:302`: "once per declaring class" is pinned by no test

The PR's headline claim, the ADR's wording ("once for each declaring class") and the design spec all rest on this predicate:

```csharp
if (ReferenceEquals(deinitializer?.Function, function)
    && SyntaxTrees.Contains(deinitializer.Declaration.SyntaxTree))
```

I weakened it to `if (deinitializer is not null && …)` — a product mutant, condition weakening, not `if (true)`:

| step | BUILD_RC | Core md5 | f8 warning count | `Interpreter.Tests` |
|---|---|---|---|---|
| clean | 0 | `3d9a937f…` | **1** | 0 failed / 610 |
| mutant | 0 | `da4a172e…` | **3** | **0 failed / 610 — SURVIVES** |
| restored | 0 | `3d9a937f…` | **1** | — |

The mutant is unambiguously live: a class declaring `deinit` plus two methods warns **three times** instead of once, one per member. The full suite does not notice.

The cause is that `DeinitializersWarnOncePerDeclaringClassWithoutRunning` — the test *named* for this property — uses `First` and `Second`, both of which declare **only** a `deinit` and nothing else. With one member each, per-member duplication is unobservable to it.

The shipped behaviour is correct; what is missing is the regression guard on the PR's own headline invariant. **Fix:** add a method to one of the two test classes. One line.

## SHOULD-FIX 2 — `src/Repl/Program.cs:56-61` duplicates #3024's identical change and conflicts with it in four files

The renderer change here is **byte-identical** to the one in open PR #3024 (`using GSharp.Core.IO;` plus `if (cell.Diagnostics.Length > 0) Console.Error.WriteDiagnostics(cell.Diagnostics);`). The issue's own design called for routing `gsi` through the rich renderer **once**, as a shared prerequisite for both diagnostics, precisely so it would not be done per-issue.

`git merge-tree --write-tree` from this PR's merged tree `a4cdee748` against #3024 (`bab0798…` lineage, current head) → **rc=1, four conflicts**:

```
CONFLICT (content): src/Core/CodeAnalysis/DiagnosticDescriptors.cs
CONFLICT (content): src/Repl/Program.cs
CONFLICT (content): docs/diagnostics.md
```

Three of those are append-point collisions that resolve mechanically, but `Program.cs` is a genuine duplicate-implementation conflict, and #3024's version additionally passes the script path (`engine.Evaluate(source, arg)`) so the resolutions are **not** interchangeable — taking the wrong side silently drops the file path from every `gsi` diagnostic. A third open PR, **#3011**, also rewrites this same block. Whichever lands first should carry the renderer alone; the others should rebase onto it rather than resolve by hunk selection.

## SHOULD-FIX 3 — `test/Core.Tests/CodeAnalysis/Binding/Issue698DeinitBinderTests.cs:33,55,125`: assertion weakened further than needed

Three tests changed from `Assert.Empty(result.Diagnostics)` to `Assert.DoesNotContain(result.Diagnostics, d => d.IsError)`. The change is necessary — the new warning legitimately makes the set non-empty — but it is broader than necessary: these tests now accept **any** warning silently, where before they caught every unexpected diagnostic. At line 55 that assertion is the test's only one.

**Fix:** keep the strength by allowing exactly the expected ID, e.g. assert that every remaining diagnostic is `GS0517`, rather than merely that none is an error.

## NIT 1 — the warning is reported for programs that never run

A source file with a parse error still emits `GS0517` alongside the error (measured: `GS0005` + `GS0517` together, rc=1). Nothing was skipped, because nothing executed. Harmless — the error dominates and the exit code is already 1 — but the message asserts a runtime consequence that did not occur.

## NIT 2 — the ADR still does not say *why* this is a Warning

`GS0513`, the only sibling interpreter-boundary diagnostic, is an **Error**; `GS0517` is a **Warning**. The +9-line ADR addition states the behaviour but not the reason for the asymmetry, so the next boundary diagnostic will copy whichever precedent is nearest. This was raised in the prior review of this PR and has not landed. One sentence — that `GS0513` fires when the program cannot proceed at all, whereas `GS0517` fires when the program completes and only a GC-scheduled side effect is skipped, which the CLR does not guarantee under `gsc` either — makes the exit-0 choice a recorded decision.

---

### What I did not run, and why

`Compiler.Tests` (~1h20m) and `Cs2Gs.Tests` were not run: the production change is confined to `Compilation.Evaluate` (the interpreter path) and `src/Repl/`, and I verified `gsc` emits this diagnostic zero times, so the emit path is untouched by observation as well as by construction. `e2e` was not run (prohibited locally). Those three are the author's figures, not mine. `samples/Deinit.gs` appears in the test tree only as a source hash in `refactoring-baseline.json`, not as an executed interpreter fixture, so I found no local samples-gate exposure — but the `cs2gs` and `e2e` required checks are the authority there, not me.

Working trees ended porcelain-empty with the Core hash round-tripped to clean `3d9a937f…`.

DavidObando:
## Addendum — re-measured after `main` moved to `33bea4af` and **#3024 landed**. Verdict unchanged (NOT MERGEABLE), but there is now a **second BLOCKER: the merged tree does not compile.**

My earlier review was measured against `main` `5ad0109c`, merged tree `a4cdee748`. `main` has since advanced two commits — **#3047** and, decisively, **#3024** (`cceb0d5f`, the sibling `gsi` P/Invoke boundary by the same author). The PR head has not moved (`a71db53f5`). Everything below is true of `main` = **`33bea4af`**, merge base **`185431c4`** (PR now **10 behind**).

### BLOCKER 2 (new) — `src/Core/CodeAnalysis/Compilation/Compilation.cs:327`: merged tree fails to build with `CS0128`, and `merge-tree` calls that file clean

#3024 added its P/Invoke boundary scan to **the same method** this PR adds its deinit scan to, `Compilation.Evaluate`, and **both name their local `interpreterDiagnostics`**. The two hunks are ~30 lines apart, so git merges them without complaint:

```
$ git merge-tree --write-tree ghr/main pr3052h
CONFLICT (content): docs/diagnostics.md
CONFLICT (content): src/Core/CodeAnalysis/DiagnosticDescriptors.cs
CONFLICT (content): src/Repl/Program.cs
Auto-merging src/Core/CodeAnalysis/Compilation/Compilation.cs      <-- reported CLEAN
```

I resolved the four declared conflicts canonically (union in both docs files and in `DiagnosticDescriptors.cs`; `main`'s side for `Program.cs`, see below), left `Compilation.cs` exactly as git auto-merged it, and built:

```
BUILD_RC=1
src/Core/CodeAnalysis/Compilation/Compilation.cs(327,13): error CS0128:
  A local variable or function named 'interpreterDiagnostics' is already defined in this scope
```

Line 298 (this PR) and line 327 (#3024, now on `main`) declare the same local in the same scope. **This is invisible to every CI run**: the PR is green on its branch, `main` is green, and the conflict list does not mention the file. Same shape as #3055.

**The fix is one token.** I renamed this PR's local to `deinitDiagnostics` (three lines: 298, 305, 318) and re-measured:

| step | BUILD_RC | errors | warnings |
|---|---|---|---|
| co-merged, as git produces it | **1** | 1 × `CS0128` | 0 |
| after one-token rename | **0** | **0** | **0** |

**Behavioural coexistence then verified, not assumed** — a co-merged tree and both suites run together, per the standard that file-list/`merge-tree` disjointness is a hypothesis:

- `Interpreter.Tests` on the co-merged tree: **0 failed / 613 passed / 613** (up from 610 — #3024's cases are now in the same assembly as this PR's).
- Cross-feature probe, one program declaring **both** a P/Invoke function and a `deinit`, run under `gsi` with all three `GSharp.Core.dll` copies (`Core/`, `Compiler/`, `Repl/`) verified mtime-identical at `03:40`:

```
gsi rc=1
both.gs(4,6,4,18): error   GS0514: P/Invoke function 'NativeStrLen' is not supported by the interpreter...
both.gs(8,5,8,11): warning GS0517: Deinitializer of class 'Rich' will not run under the interpreter...
```

Both fire, neither suppresses the other, severities and locations correct, `allWarnings` correctly survives #3024's early return. So the interaction is **purely the name collision** — there is no semantic conflict to design around.

### SHOULD-FIX 2 is superseded: it is no longer "sequence two open PRs", and its `Program.cs` conflict invites a destructive resolution

#3024 has landed, so `src/Repl/Program.cs` now carries the renderer change on `main` **and** this PR still carries its own copy. The conflict git presents is a single line:

```
<<<<<<< HEAD
        var cell = engine.Evaluate(File.ReadAllText(arg), arg);
=======
        var cell = engine.Evaluate(File.ReadAllText(arg));
>>>>>>> pr3052h
```

The two sides differ **only** by `, arg`. Taking your own branch's line — the natural reflex when your side looks correct in isolation — silently deletes the file-path argument, which is exactly what makes `gsi` print source locations. My probe above shows `both.gs(4,6,4,18)`; resolve it the tempting way and every `gsi` diagnostic loses its file name, quietly removing the feature #3024 just shipped. **Take `HEAD` for this file and drop the PR's `Program.cs` hunk entirely** — it is now redundant with `main`.

The remaining three conflicts are ordinary append-point collisions (union resolves them).

### BLOCKER 1 (`GS0517` should be `GS0510`) — re-verified on the new `main`, unchanged

| query on `main` `33bea4af` | rc | files |
|---|---|---|
| `git grep -c GS0510` | **1** | **none** — still free |
| `git grep -c GS0517` | **1** | **none** — this PR is still its only source |
| `git grep -c GS0514` (control) | 0 | 10 files — landed with #3024 |
| `git grep -c GS0513` (control) | 0 | 6 files |

The controls confirm the zeroes are absence, not scope. Note that #3024 shipped with **`GS0514`, the ID that was allocated to it** — so the sibling PR by the same author already follows the allocation. This one still needs the rename to `GS0510`.

### What this supersedes

My earlier suite figures (`Interpreter.Tests` 610/610, `Core.Tests` 7072/1 skipped/7073) were true of merged tree `a4cdee748` on `main` `5ad0109c`. **That tree no longer exists and those numbers should not be cited for this PR.** I did not re-run `Core.Tests` on the new co-merged tree, and I did not run `Compiler.Tests`, `Cs2Gs.Tests` or `e2e` — but note that until the `CS0128` is fixed, **no merged-tree suite can run at all**, so there are currently no valid merged-tree results for this PR from anyone.

`SHOULD-FIX 1` (the once-per-declaring-class predicate pinned by no test) is unaffected: `Compilation.cs` auto-merged that hunk unchanged, and the mutation result was measured through `gsi` and `dotnet test` rather than an emitted assembly, so it is independent of both the rebase and of output-directory staleness.

### Summary of required changes

1. Rename `GS0517` → **`GS0510`** (descriptor, two docs rows, ADR, test filter).
2. Rename this PR's `interpreterDiagnostics` local (`Compilation.cs:298,305,318`) to avoid `CS0128` against `main`.
3. Drop the `src/Repl/Program.cs` hunk — `main` already has it, in a strictly better form.
4. Add a member to a class in `DeinitializersWarnOncePerDeclaringClassWithoutRunning` so the once-per-class invariant is pinned.
5. Tighten the three `Issue698DeinitBinderTests.cs` assertions to allow only the expected ID.

DavidObando:
# Review: PR #3052 — `fix/2988-gsi-deinitializers`

**Verdict: MERGEABLE.** No blockers. Three NITs, none of which should hold the merge.

Reviewed at head **`a18cc2b1`** against `main` **`4469e323`**. Merge-base **equals** remote `main`, so the PR already contains `4469e323` and **the merged tree is the PR head itself** — I built it rather than inferring: `BUILD_RC=0`, 0 errors, 44 warnings. Both scratch trees left porcelain-clean with the Core hash round-tripped (`028c2ef4…` → mutate → restore → `028c2ef4…`).

---

## Scope: this PR warns, it does not make deinitializers run — and that is the sanctioned design

I verified the decision myself rather than relying on the PR body or the prior review. Issue #2988 **comment 2 of 4** records:

> **G5 policy decision: explicit boundary + diagnostic. Do not implement deinit in the interpreter.**

and explicitly says it contradicts the issue body's "Note on scope". So the axes usually demanded of a deinit fix — ordering, reverse-declaration, nested scopes, exceptional exits, idempotence of *execution* — are moot by design. I re-aimed each of them at the **warning** instead, which is the thing that now has to be correct.

Two named interaction risks are **structurally impossible**, by measurement rather than argument: `deinit` on a struct is rejected outright with `error GS0289: 'deinit' is only valid on a class type`, so **#3055**'s ByRefLike boundary and **#3105**'s synthesised struct types cannot own a deinitializer.

---

## Prior blockers — status with regression evidence

### BLOCKER 1 (self-allocated diagnostic ID) — **FIXED**

Ships **`GS0510`** at `src/Core/CodeAnalysis/DiagnosticDescriptors.cs:384` — the ID allocated in issue #2988 comment 3 ("Correction to my G5 design: `GS0507` is not safe — use `GS0510`"), which I read directly.

- `GS0517` now has **0 occurrences in every tracked file** (`git grep`), positive control `GS0510` = 1 in `src/`.
- Exactly **one bare row** per catalogue, severity matching the descriptor: `docs/diagnostics.md:1182` and `website/docs/ref/diagnostics.md:1778`.
- **Regression-tested, not merely invisible:** flipping the descriptor severity to `Error` fails `Every_Diagnostic_Is_Documented_In_Both_References_With_Matching_Severity` (measured: 3 passed / 1 failed). The bidirectional docs gate genuinely holds this row.

### BLOCKER 2 (merged tree does not compile, `CS0128`) — **FIXED**

The colliding local is renamed `interpreterDiagnostics` → **`deinitDiagnostics`**. Both scans now coexist in `Compilation.Evaluate`: the new deinit scan at `Compilation.cs:325` and #3024's P/Invoke scan at `Compilation.cs:352`.

Built the merged tree: **`BUILD_RC=0`**. This one needs no test — a re-collision is a compile error, so the compiler is the gate.

### SHOULD-FIX 1 (once-per-declaring-class pinned by no test) — **FIXED, and I re-ran the mutant that previously survived**

`class First` in `Issue2988DeinitInterpreterTests.cs` now declares `func Touch()` alongside its `deinit`, which is what makes per-member duplication observable. Re-running the exact previously-surviving mutant:

`ReferenceEquals(deinitializer?.Function, function)` → `deinitializer is not null` — **MUTANT-APPLIED (1 site)**, `BUILD_RC=0`, **717/718, `DeinitializersWarnOncePerDeclaringClassWithoutRunning` fails.** **Killed.**

### SHOULD-FIX 2 (duplicate `src/Repl/Program.cs` renderer) — **FIXED by dropping the hunk**

`src/Repl/Program.cs` is **absent from the three-dot diff** — the standing "take HEAD, drop the hunk" resolution. Verified behaviourally, not from the file list: head `gsi` diagnostics still carry the source path (`/tmp/…/Deinit.gs(22,5,22,11): warning GS0510: …`), so #3024's `engine.Evaluate(source, arg)` survives intact.

### SHOULD-FIX 3 / NIT 1 / NIT 2 — addressed

`Assert.Empty` → `Assert.All(…, d => Assert.Equal("GS0510", d.Id))` (tighter than what was asked — any non-`GS0510` diagnostic fails). The error-suppression gate `if (!allErrors.Any())` fixes the parse-error NIT and is pinned. The Warning-vs-Error rationale is now in the ADR and PR body.

---

## The reported shape, run verbatim first

Issue #2988 names `samples/Deinit.gs`. Byte-for-byte, before any variation:

| Driver | base `4469e323` | head `a18cc2b1` |
|---|---|---|
| `gsi` | `Allocated: db/users` / `done`, rc 0, **no diagnostic** | same output + **2× `GS0510`** (`Resource`, `CachedResource`), rc 0 |
| bare `gsc` (evaluate) | same, no diagnostic | same output + **2× `GS0510`**, rc 0 |
| `gsc /out:` (**oracle**) | runs both deinits | **identical**, **0× `GS0510`** |

Three things this establishes:

1. **The emit oracle is untouched and never sees the warning** — 0 occurrences on both trees, so no false positive on the path where deinit genuinely runs.
2. **Bare `gsc` was checked as its own column and also warns.** This matters: the defect lives in the *shared evaluator*, and the fix is correctly placed in `Compilation.Evaluate`, so both evaluate-column drivers are covered. Had it been placed in `src/Repl/`, bare `gsc` would still diverge silently.
3. Output and rc are otherwise unchanged — the warning adds information without altering behaviour.

**Blast radius is exactly one file:** `samples/Deinit.gs` is the only sample declaring `deinit` (1 of 122; positive control: 8 `deinit` hits in that file, 0 in the other 121). It goes rc0 → rc0 with output identical, so no sample regresses.

## Additional shapes probed (head, all three drivers, each into its own asserted-empty directory)

| Shape | Result |
|---|---|
| `func Main() int32` + `deinit` (**#3011 interaction**) | `gsi` **rc 33**, identical to emit oracle **rc 33**; `GS0510` on stderr does **not** corrupt the exit code |
| `struct` + `deinit` | rejected `GS0289` on all three columns — the message's hard-coded "class" wording is unreachable |
| generic `class Box[T]` + `deinit` | warns correctly (`Box`) |
| class declared, never instantiated | warns |
| runtime exception after allocation | `GS0510` **survives** alongside `GS9999` — warnings are not lost on the throwing path |

The `func Main` interaction was the genuinely new one nobody had checked; it is safe.

---

## Mutation evidence (4 mutants, 4 kills, 0 survivors)

Every mutant asserted applied before building; every `BUILD_RC` read. Per hunk, never per PR.

| # | Product mutant | `BUILD_RC` | Kill |
|---|---|---|---|
| M1 | `ReferenceEquals(deinitializer?.Function, function)` → `deinitializer is not null` | 0 | `DeinitializersWarnOncePerDeclaringClassWithoutRunning` (717/718) |
| M2 | drop `&& SyntaxTrees.Contains(deinitializer.Declaration.SyntaxTree)` | 0 | `DeinitializersWarnOncePerDeclaringClassWithoutRunning` (717/718) — the REPL-continuation assertion `Assert.DoesNotContain(next.Diagnostics, …)` is what catches it |
| M3 | invert `if (!allErrors.Any())` | 0 | **4 tests**, incl. `CompilationErrorsSuppressDeinitializerBoundaryWarning` — both arms pinned |
| M4 | descriptor `Warning` → `Error` | 0 | 3 Interpreter tests **and** `Every_Diagnostic_Is_Documented_In_Both_References_With_Matching_Severity` |

M2 is worth calling out: the `SyntaxTrees.Contains` clause looks like unreachable defensive code, but it is exactly what stops a REPL session re-warning about a class declared in an *earlier* submission. It is load-bearing and it is pinned.

## Suites on the merged tree

| Project | Result |
|---|---|
| `Interpreter.Tests` | **718 / 718** |
| `Core.Tests` | **7089 passed, 1 skipped, 7090 total** |
| `Compiler.Tests` | **NOT RUN** — ~1h07–1h20 locally; CI shards it |
| `Sdk.Tests` | **NOT RUN** — deliberately: it touches the shared `~/.nuget/packages/gsharp.net.sdk/` cache that parallel agents race on |

5 of 7 run green; the two omissions are stated rather than implied.

**The PR body's numbers are accurate.** I re-derived +242/−4 across 8 files, 718/718, and 7089+1-skipped independently and they match. Given how often PR bodies have been quantitatively wrong lately, that is worth recording.

---

## NITs (none blocking)

**NIT 1 — the issue's own repro shape is covered by no test.** `samples/Deinit.gs` appears only as a source hash in `test/Core.Tests/Baselines/refactoring-baseline.json`; the samples gate is emit-only (`SampleConformanceTests.cs`), so nothing executes it under `gsi`. Its distinguishing feature is **inheritance** — `CachedResource : Resource`, both declaring `deinit` — and none of the 164 added test lines uses a derived class. I verified by hand that it emits exactly 2 correctly-attributed warnings, so the behaviour is right; it is simply unpinned. A third class pair in `Issue2988DeinitInterpreterTests` mirroring the sample would close it cheaply.

**NIT 2 — `Assert.All` passes vacuously on an empty collection.** `test/Core.Tests/CodeAnalysis/Binding/Issue698DeinitBinderTests.cs:55`, `Deinit_BodyHasAccessToFieldsAndThis`, now has `Assert.All(...)` as its *only* assertion, where it previously had `Assert.Empty`. It still catches what it was written for (a binding failure produces a non-`GS0510` Id), and `GS0510`'s presence is positively pinned in `Interpreter.Tests`, so the practical risk is low. Adding `Assert.Contains(result.Diagnostics, d => d.Id == "GS0510")` alongside would make it non-vacuous.

**NIT 3 — the warning fires for a class that is never instantiated**, where no divergence from the oracle actually exists. This is consistent with the sibling declaration-site boundary diagnostics `GS0513`/`GS0514`, so I would leave it; noting it only so the choice is deliberate.

**Explicitly not a finding against this PR:** runtime `GS9999` diagnostics under `gsi` render as `(1,1,1,7)` with **no file path**, while `GS0510` renders the full path. I confirmed this is byte-identical on base `4469e323`, so it is pre-existing and outside this PR's scope.

---

## My own errors, disclosed

1. My first mtime check sorted `HH:MM:SS` strings across files with **different dates** and reported a driver timestamp *in the future*. Re-run with full dates: `Core/`, `Compiler/` and `Repl/` were all fresh (19:16:47–53).
2. My mutation harness only creates a backup **if one does not already exist**, and `/tmp/…/bak_M1` was a stale `Compilation.cs` left over from a previous PR review. Restoring from it wrote the wrong file (310+/674−). Caught immediately via `git status`, restored with `git checkout --`, stale backups purged, and all subsequent restores done through git. **M1's measurement is unaffected** — the corruption occurred after the build and test had already run against the correctly-mutated file, whose context I had printed and verified.

Load stayed within bounds throughout (10.31–25.81 one-minute at each build).

DavidObando:
## Addendum — my suite denominator was wrong ("5 of 7"); corrected to 5 of 9, with the two missing projects measured on current `main`

Correcting my own verdict above, per the fleet bulletin on test-project enumeration.

I wrote **"5 of 7 run green; the two omissions are stated rather than implied."** The omissions were stated honestly, but **the denominator was wrong**. `GSharp.sln` contains **nine** test projects; two live outside `test/` and I never ran them:

- `tools/cs2gs/Cs2Gs.Tests` — drives the real `gsc` out-of-process

Correct statement: **5 of 9 run**, with **four** not run — `Compiler.Tests` (~1h07–1h20 locally; CI shards it), `Sdk.Tests` (shared `~/.nuget/packages/gsharp.net.sdk/` cache that parallel agents race on), and these two, which I did not know existed. My enumeration came from `ls test/`; the zero was **scope, not absence**.

### Measured on current `main`

This PR's tree is long gone and `main` has moved many commits past it, so I measured what is actually actionable — whether anything of this shape is live now. Current `main` **`5b71a74d`** (`git rev-parse` verified against `origin/main`, not assumed), full solution rebuild after checkout, **`BUILD_RC=0`**:

| project | result on `main` `5b71a74d` |
|---|---|
| `Cs2Gs.Tests` | **1903 passed, 0 failed** |

Freshness verified in **epoch seconds with `sort -n`**, `Core/`/`Compiler/`/`Repl/` all at one identical stamp, `ref/` excluded.

For this PR specifically the exposure was low — it adds a `GS0510` warning at an interpreter boundary rather than changing codegen — but "low exposure" is a judgement I did not get to make at the time, because I did not know the projects existed.

**Verdict unchanged: MERGEABLE.** Both prior blockers remain verified fixed as filed.

PR #3053, opened 2026-08-01
CI: harden netcoredbg installation

### LINKED ISSUE #3051: CI: e2e job fails intermittently in 'Install netcoredbg' due to unauthenticated GitHub API rate limiting
## Problem

The `e2e` job in `.github/workflows/build.yml` fails intermittently before running a single test. Observed on PR #3039 and PR #3048 within the same hour:

```
✓ Restore
X Install netcoredbg
- Run end-to-end tests        ← never runs
##[error]Process completed with exit code 1.
```

Job died in **18-21 s**. Zero tests executed.

## Root cause

The step is:

```bash
URL=$(curl -sSL https://api.github.com/repos/Samsung/netcoredbg/releases/latest \
  | grep -oP '"browser_download_url":\s*"\K[^"]*linux-amd64\.tar\.gz')
curl -sSL "$URL" -o /tmp/netcoredbg.tar.gz
mkdir -p "$HOME/.tools/netcoredbg"
tar -xzf /tmp/netcoredbg.tar.gz -C "$HOME/.tools/netcoredbg" --strip-components=1
echo "$HOME/.tools/netcoredbg" >> "$GITHUB_PATH"
```

Three compounding defects:

1. **Unauthenticated `api.github.com` call.** No `Authorization` header, so it draws on the shared 60-req/hr per-IP anonymous quota. Under concurrent PR load — this repo currently has ~25 open PRs — it returns a rate-limit JSON body instead of a release. `grep -oP` then matches nothing, `URL` is empty, and the shell (`bash -e`) exits 1.
2. **No retry and no diagnostic.** The failure message is a bare `Process completed with exit code 1`. Nothing says *rate limited* or *empty URL*, so every agent and human who hits it initially suspects their own change. This has already cost review cycles twice today.
3. **A third-party release endpoint is a hard dependency of every PR's e2e gate.** Any `Samsung/netcoredbg` outage or asset rename reds every open PR at once.

## Suggested fix

- Pass `GITHUB_TOKEN` on the API call (`curl -H "Authorization: Bearer $GITHUB_TOKEN"`) — raises the limit to 1000/hr per repo and is the documented remedy.
- **Fail loudly and specifically**: assert `URL` is non-empty and echo the API response body when it is not.
- Add bounded retry with backoff.
- Consider **pinning a known-good netcoredbg version** rather than tracking `latest`, so a third-party release cannot change CI behaviour without a commit. Caching the tarball via `actions/cache` keyed on that version removes the network dependency from the common path entirely.

## Verification note

**Both observed failures were confirmed infra, not code**: PR #3039 passed `e2e` cleanly on a plain re-run with no code change. A fix should be judged on whether the failure mode is now *legible* — an agent hitting it must be able to tell it is not their bug from the log alone.

DavidObando:
## Summary

Make the `e2e` job's netcoredbg installation authenticated, bounded, and self-diagnosing.

The workflow now passes its existing `${{ github.token }}` to a small installer script. No workflow permissions are added or widened. The installer:

- enables `set -euo pipefail`
- authenticates the GitHub release API request
- retries an empty/malformed release response three times with 2 s / 4 s backoff
- bounds each API request and the asset download
- prints the API response body and identifies likely rate limiting when no Linux asset is found
- retries transient asset download failures

I kept `latest` rather than adding a version pin and cache. Authentication plus bounded retry directly fixes the measured failure with less maintenance and no cache lifecycle; changing the existing release-tracking policy is not needed for #3051.

## Induced failure

A fake `curl` returned a rate-limit response for all three API attempts. The exact output was:

```text
exit=1 attempts=3
::warning::netcoredbg release lookup returned no linux-amd64 asset (attempt 1/3); GitHub API rate limiting is likely. Retrying in 2s.
::warning::netcoredbg release lookup returned no linux-amd64 asset (attempt 2/3); GitHub API rate limiting is likely. Retrying in 4s.
::error::Unable to resolve the netcoredbg linux-amd64 release after 3 attempts. GitHub API rate limiting is likely; this is a CI infrastructure failure, not a G# test failure.
GitHub API response body:
{"message":"API rate limit exceeded for 203.0.113.1.","documentation_url":"https://docs.github.com/rest/using-the-rest-api/rate-limits-for-the-rest-api"}
```

## Retry and success path

A second fake returned the same rate-limit body once, then delegated to the real authenticated API and download:

```text
install_exit=0 api_attempts=2
::warning::netcoredbg release lookup returned no linux-amd64 asset (attempt 1/3); GitHub API rate limiting is likely. Retrying in 2s.
archive_bytes=3510717 path_entries=1
netcoredbg: ELF 64-bit LSB pie executable, x86-64
```

The archive downloaded, extracted into the configured tool directory, produced an executable Linux x86-64 binary, and appended exactly one entry to `GITHUB_PATH`. Per repository rules I did not run the local e2e suite or touch the shared SDK package cache.

Additional checks:

- `bash -n build/install-netcoredbg.sh`
- workflow YAML parsed successfully
- staged diff check passed

Fixes #3051

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -103,13 +103,9 @@ jobs:
         run: dotnet restore GSharp.sln
 
       - name: Install netcoredbg
-        run: |
-          URL=$(curl -sSL https://api.github.com/repos/Samsung/netcoredbg/releases/latest \
-            | grep -oP '"browser_download_url":\s*"\K[^"]*linux-amd64\.tar\.gz')
-          curl -sSL "$URL" -o /tmp/netcoredbg.tar.gz
-          mkdir -p "$HOME/.tools/netcoredbg"
-          tar -xzf /tmp/netcoredbg.tar.gz -C "$HOME/.tools/netcoredbg" --strip-components=1
-          echo "$HOME/.tools/netcoredbg" >> "$GITHUB_PATH"
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: ./build/install-netcoredbg.sh
 
       - name: Run end-to-end tests
         run: ./build/run-e2e-tests.sh
--- a/build/install-netcoredbg.sh
+++ b/build/install-netcoredbg.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+api_url=https://api.github.com/repos/Samsung/netcoredbg/releases/latest
+response=
+url=
+
+for attempt in 1 2 3; do
+  if response=$(curl -sSL --connect-timeout 15 --max-time 60 \
+    -H "Authorization: Bearer ${GITHUB_TOKEN:?GITHUB_TOKEN is required}" \
+    "$api_url"); then
+    while IFS= read -r line; do
+      if [[ $line =~ \"browser_download_url\"[[:space:]]*:[[:space:]]*\"([^\"]*linux-amd64\.tar\.gz)\" ]]; then
+        url=${BASH_REMATCH[1]}
+        break
+      fi
+    done <<< "$response"
+  fi
+
+  if [[ -n $url ]]; then
+    break
+  fi
+
+  if (( attempt < 3 )); then
+    delay=$((attempt * 2))
+    echo "::warning::netcoredbg release lookup returned no linux-amd64 asset (attempt $attempt/3); GitHub API rate limiting is likely. Retrying in ${delay}s." >&2
+    sleep "$delay"
+  fi
+done
+
+if [[ -z $url ]]; then
+  echo "::error::Unable to resolve the netcoredbg linux-amd64 release after 3 attempts. GitHub API rate limiting is likely; this is a CI infrastructure failure, not a G# test failure." >&2
+  echo "GitHub API response body:" >&2
+  printf '%s\n' "${response:-<empty response>}" >&2
+  exit 1
+fi
+
+temp_dir=${RUNNER_TEMP:-"$PWD/out/tmp"}
+archive=$temp_dir/netcoredbg.tar.gz
+install_dir=$HOME/.tools/netcoredbg
+
+mkdir -p "$temp_dir" "$install_dir"
+curl -fsSL --retry 3 --retry-all-errors --connect-timeout 15 --max-time 180 \
+  "$url" -o "$archive"
+tar -xzf "$archive" -C "$install_dir" --strip-components=1
+echo "$install_dir" >> "$GITHUB_PATH"
--- a/build/install-netcoredbg.sh
+++ b/build/install-netcoredbg.sh
@@ -44,4 +44,18 @@ mkdir -p "$temp_dir" "$install_dir"
 curl -fsSL --retry 3 --retry-all-errors --connect-timeout 15 --max-time 180 \
   "$url" -o "$archive"
 tar -xzf "$archive" -C "$install_dir" --strip-components=1
+
+binary=$install_dir/netcoredbg
+if [[ ! -x $binary ]]; then
+  echo "::error::netcoredbg archive extracted but $binary is missing or not executable. Extracted layout:" >&2
+  ls -laR "$install_dir" >&2
+  exit 1
+fi
+
+if ! version=$("$binary" --version 2>&1); then
+  echo "::error::netcoredbg was extracted to $binary but failed to run: $version" >&2
+  exit 1
+fi
+
+echo "Installed $version"
 echo "$install_dir" >> "$GITHUB_PATH"
--- a/e2etests/debugger-e2e.sh
+++ b/e2etests/debugger-e2e.sh
@@ -6,8 +6,9 @@
 # verifies it hits, lists locals (exercising LocalScope/LocalVariable from
 # Phase 5), and steps through GSharp source.
 #
-# The script SKIPS cleanly (exit 0) when netcoredbg is unavailable so that
-# CI lanes that do not provision it stay green. Local devs can install
+# The script SKIPS cleanly (exit 0) for local developers when netcoredbg is
+# unavailable, but treats absence as fatal in CI so debugger coverage cannot
+# silently disappear. Local devs can install
 # netcoredbg by following https://github.com/Samsung/netcoredbg/releases
 # or via the helper at the top of this file.
 set -euo pipefail
@@ -37,7 +38,12 @@ skip() {
 
 NETCOREDBG="$(find_netcoredbg || true)"
 if [[ -z "$NETCOREDBG" ]]; then
-    skip "netcoredbg not installed (install from https://github.com/Samsung/netcoredbg/releases and place on PATH or in $TOOLS_DIR/netcoredbg/)"
+    message="netcoredbg not installed (install from https://github.com/Samsung/netcoredbg/releases and place on PATH or in $TOOLS_DIR/netcoredbg/)"
+    if [[ "${CI:-}" == "true" ]]; then
+        echo "::error::$message" >&2
+        exit 1
+    fi
+    skip "$message"
 fi
 
 # netcoredbg ships an osx-amd64 build but no osx-arm64 build, so on


DavidObando:
## PR #3053 — **MERGEABLE**

Base `f820484c3` (computed `merge-base`), head `65c83fc80`, `main` `fdba684b6`. Three-dot = **2 files / +50 −7**. (Two-dot reports 7 files / +74 −44 — phantom.)

I cannot run the workflow, so I did the next best thing: **executed the new script against a stubbed `curl`** and read the real failure log from the API.

### First — which step actually failed

Verified, not assumed. Run `30710783015` (branch `chore/3028-remove-debug-probe`, i.e. #3048), **attempt 1**: job `e2e`, failed step **`Install netcoredbg`**; **attempt 2** green. The step-level diagnosis in #3051 is correct.

### Correction on the cause — `pipefail` is not the operative defect

Two claims in the assignment do not survive measurement:

1. **"18–21 s"** — the step API reports `started_at` == `completed_at` == `18:20:11Z`; log timestamps give `11.4035` → `11.4810`, i.e. **~78 ms**.
2. **"Without `pipefail` … the step can still *succeed* with an empty result. This is the actual defect."** — It did not succeed; it failed, `exit 1`. Log line 252 confirms the shell is `/usr/bin/bash -e {0}`. In `URL=$(curl … | grep -oP …)` the assignment takes the *pipeline's* status, and the failing command (`grep`, no match) is **last in the pipe** — so its status was already the pipeline's status and `-e` already killed the step. `pipefail` would have changed nothing about this failure.

The script does add `set -euo pipefail` (`build/install-netcoredbg.sh:3`) and that is correct defensive hygiene — it just is not what fixes #3051.

What the log *does* establish:

- **Zero output.** `grep -cE "(curl:|grep:)"` on the job log = **0**. Bare `exit 1`, no diagnostic at all.
- **`curl` itself succeeded.** With `-sS`, a transport failure would have written to stderr; nothing did. So a response *was* received and simply contained no `linux-amd64.tar.gz` asset.
- **Rate limiting is plausible but is NOT in evidence.** My first grep for rate-limit text "hit" 5 times — all substring false positives inside timestamps and a SHA (`…4034324Z`, `19fdda60136bbac4030757…`). A substring is not a reference. The honest statement is: *nobody can currently diagnose this failure*, which is itself the strongest argument for this PR.

So the real defect is **unauthenticated request + no diagnostic**, and that is exactly what the PR fixes.

### Behaviour verified by execution

Ran `build/install-netcoredbg.sh` with a `curl` shim on `PATH` and a sandboxed `HOME`/`RUNNER_TEMP`/`GITHUB_PATH`:

| case | rc | elapsed | observed |
|---|---|---|---|
| healthy API | **0** | — | `GITHUB_PATH` written, binary installed, **no** stderr noise |
| rate-limited on all 3 attempts | **1** | **6 s** (2 s + 4 s backoff) | 2× `::warning::`, then `::error::` **plus a dump of the API response body** |
| rate-limited once, healthy on attempt 2 | **0** | **3 s** | 1× `::warning::`, 2 API calls, install completes |

The middle row is the point: the next occurrence will be **self-diagnosing**, which the current step is not. The third row shows the retry is real, not decorative.

### Other checks

- **Authentication present:** `-H "Authorization: Bearer ${GITHUB_TOKEN:?GITHUB_TOKEN is required}"` with `GITHUB_TOKEN: ${{ github.token }}` on the step. No `permissions:` block at workflow or job level, so the token is available; fork `pull_request` runs still get an authenticated (read-only) token, so the limit is still raised off the shared 60/hr per-IP pool. This addresses the trigger, not just the symptom.
- **Executable bit:** `100755`, matching the `build/run-e2e-tests.sh` precedent (`100755`). Had this been `100644`, `run: ./build/install-netcoredbg.sh` would fail with `Permission denied` — it is correct.
- **YAML parses**; `runs-on: ubuntu-latest`, no `defaults:` overriding shell. The script carries its own shebang, so the outer shell is not load-bearing.
- **Completeness — no "5 of 7 sites" shape.** `api.github.com` occurs **exactly once** across `.github/` and `build/` (the line now authenticated); zero other `curl` invocations in any workflow. `e2etests/debugger-e2e.sh` only *consumes* netcoredbg, it does not install it.
- **Subshell trap avoided:** the extraction loop uses `done <<< "$response"`, not `echo … | while read`, so `url` survives. Worth noting because the pipe form would have silently produced an empty `url` — the same class of bug being fixed.

---

## SHOULD-FIX — `build/install-netcoredbg.sh:46` — install can "succeed" without installing anything

The script never verifies what it extracted. Probe: a tarball whose layout defeats `--strip-components=1`.

```
install rc=0      <-- layout wrong
stderr=[]         <-- zero output
GITHUB_PATH=[…/.tools/netcoredbg]
binary at expected path? NO
extracted tree:   .tools/netcoredbg/nested/netcoredbg
```

Then `e2etests/debugger-e2e.sh:21-40`: `find_netcoredbg` checks `command -v netcoredbg` (nothing at the top of the PATH dir) and `-x "$TOOLS_DIR/netcoredbg/netcoredbg"` — and `TOOLS_DIR="$ROOT/.tools"` (`:17`) is the **repo** root, not `$HOME/.tools`, so `GITHUB_PATH` is the *only* link. Both fail → `skip()` → **`exit 0`**.

Net: green CI, `SKIP:` in the log, **zero debugger coverage**, nobody notified. That is the silent class, and it is the one outcome a PR about e2e reliability should not leave open. This is pre-existing, not a regression — but it is one line in the file you are already adding:

```bash
if [[ ! -x "$install_dir/netcoredbg" ]]; then
  echo "::error::netcoredbg archive extracted but $install_dir/netcoredbg is missing or not executable." >&2
  exit 1
fi
```

## NIT

- **`build/install-netcoredbg.sh:13-18`** — `jq` is preinstalled on GitHub-hosted Ubuntu images, so the 6-line bash-regex read loop can be one line: `url=$(jq -r '.assets[].browser_download_url | select(endswith("linux-amd64.tar.gz"))' <<<"$response" | head -1)`. More robust than regex against JSON. Not blocking.
- **Ponytail check on the 47-line script:** proportionate. Roughly 8 lines are auth, 12 retry/backoff, 6 the loud failure, 6 extraction — and moving it out of inline YAML is what made it *executable and testable*, which is how every measurement above was produced. I would not ask for it to be shrunk back into the step.

No diagnostics, no descriptors, no docs-gate exposure (no `src/` change).

DavidObando:
Merge blocker fixed in content at `20a3480de`.

I chose **fatal when the debugger is absent in CI**, while retaining the existing local skip. Local developers may legitimately lack netcoredbg (and Apple Silicon cannot use upstream's x86_64 macOS build), but CI provisions it specifically to provide debugger coverage; a missing CI debugger is therefore an error.

Bad-layout archive (`package/nested/netcoredbg`, then `--strip-components=1`):

```text
bad_layout_rc=1 output:
::error::netcoredbg archive extracted but .../.tools/netcoredbg/netcoredbg is missing or not executable. Extracted layout:
...
.../.tools/netcoredbg/nested:
-rwxr-xr-x ... netcoredbg
bad_layout_expected_binary=0 extracted_files=1
```

Good-layout archive (`package/netcoredbg`) with an executable stub that implements `--version`:

```text
good_layout_rc=0 output:
Installed NetCoreDbg test-version
good_layout_binary=1 path_entries=1
```

`find_netcoredbg` with the binary absent and `CI=true`:

```text
find_netcoredbg_absent_rc=1 output:
::error::netcoredbg not installed (install from https://github.com/Samsung/netcoredbg/releases and place on PATH or in .../.tools/netcoredbg/)
error_messages_examined=1 matched=1
```

The same absent case outside CI remains an explicit, visible local skip (`rc=0`, `SKIP: netcoredbg not installed ...`). I did not run the e2e suite or touch the shared SDK cache.

DavidObando:
# Re-review @ `20a3480dec53657463d068d365afb36e91dca59f` — MERGEABLE

Round-1's escalated SHOULD-FIX is **closed on its merits**, proven by a true revert rather than by reading the diff. Two new defects, both local-developer-only, neither re-opening the silent-green hole. No blockers.

Anchors: `main` = `f5e9d89250`. Computed merge base = `46d3cc50f`. True content **3 files, +73/−10** (three-dot). Two-dot would have reported 26 files/+119/−1564 — the trap fired again; all anchors below are in the PR tree.

Merge is clean (`git merge-tree` → `ed7786b557`). All checks pass: `build`, `cs2gs`, `e2e`, `vsix`, all three `vscode-extension` lanes.

## (a) The CI predicate

`e2etests/debugger-e2e.sh:42` — `if [[ "${CI:-}" == "true" ]]`. Strict equality, not `[ -n "$CI" ]`. Correct for GitHub Actions, which documents `CI=true` verbatim.

I attacked it in both polarities. It **cannot** be broken in the fail-closed direction on GHA. It fails **open** for non-canonical truthy values, which is a non-issue here (see NIT 1).

## (b) Absent-binary `rc`, measured

Isolated `ROOT`, `netcoredbg` absent from `PATH` and from `.tools/` (verified). The script exits at the guard, well before the SDK-pack step at `:70`, so **the e2e suite was never run** — the hard constraint is respected throughout.

| `CI` | `debugger-e2e.sh` rc | output |
|---|---|---|
| `true` | **1** | `::error::netcoredbg not installed …` |
| unset | 0 | `SKIP: …` |
| `false` | 0 | `SKIP: …` |
| `""` | 0 | `SKIP: …` |
| `1` | 0 | `SKIP: …` |
| `TRUE` / `True` | 0 | `SKIP: …` |

The exit code is only useful if the runner propagates it, so I drove the **full chain** through `build/run-e2e-tests.sh` (`:57` `rc=$?` → `:73` → `:89 exit 1`):

```
CI=true  -> run-e2e-tests.sh rc=1   ❌ debugger-e2e: FAILED (exit code 1)   Results: 0 passed, 1 failed, 0 skipped
CI=false -> run-e2e-tests.sh rc=0   ⏭  debugger-e2e: SKIPPED                Results: 0 passed, 0 failed, 1 skipped
```

Chain intact end to end. The local skip stays **visible** in the summary via `:64`'s `grep -q "^SKIP:"`.

## (c) Is fatal-in-CI safe on every runner? Yes

`.github/workflows/build.yml:83-84` — the `e2e` job is `runs-on: ubuntu-latest`, **no `strategy`, no matrix, single lane**. It is the only lane in any workflow that invokes `run-e2e-tests.sh`. Apple Silicon never runs this in CI, so fatal cannot turn a real limitation into a permanently red gate.

Production evidence at this exact head (run `30719989860`, `headSha 20a3480de`, `e2e` pass 4m25s):

```
==> Using netcoredbg: /home/runner/.tools/netcoredbg/netcoredbg
PASS: netcoredbg hit GsLib.<Program>.Add at Lib.gs line 6 — the SDK-produced Portable PDB drives a cross-language live debugger end-to-end.
```

No `SKIP`, no `::error::`, no `::warning::`. The debugger genuinely ran. That is the claim the whole PR exists to make enforceable.

## (d) Is the layout check cosmetic? No — it tests executability *and* invokes the binary

Stubbed `curl`, four hand-built tarballs:

| fixture | `rc` | `GITHUB_PATH` | outcome |
|---|---|---|---|
| `top/netcoredbg` +x, runs | 0 | 1 entry | `Installed NetCoreDbg 3.1.2-1024` |
| `top/nested/netcoredbg` | **1** | **empty** | `::error::… missing or not executable` + layout dump |
| `top/netcoredbg` mode 644 | **1** | **empty** | same `::error::` — `-x` catches it |
| +x but exits 126 | **1** | **empty** | `::error::… failed to run: cannot execute binary file` |

`build/install-netcoredbg.sh:49` uses `-x`, not `-f`. `:55` actually invokes `--version`. `:51`'s `ls -laR` is a **genuine recursive listing** — real mtimes and inode flags, not a hardcoded string. Critically, `GITHUB_PATH` is written at `:61` **only after both gates pass**.

### Perturbation — and a disclosed measurement error

My first mutant was **malformed**: `sed '48,58d'` orphaned `echo "Installed $version"`, so it died on `unbound variable` rather than by reopening the hole. Invalid; discarded. Redone as a true revert of the guard to `d11f1ac9b:build/install-netcoredbg.sh`:

```
guard REVERTED  -> bad layout: rc=0, GITHUB_PATH=[…/.tools/netcoredbg]   <-- round-1 blocker, reproduced exactly
guard RESTORED  -> bad layout: rc=1, GITHUB_PATH=[]                      <-- flips back
```

Restore verified byte-identical by md5. Both hunks are load-bearing; neither is dead code.

## Round-1's original half survived

| scenario | `rc` | behaviour |
|---|---|---|
| rate-limited ×3 | 1 | 2 `::warning::` with 2 s/4 s backoff, then `::error::` + response body |
| transient (fails twice, 3rd OK) | 0 | recovers, installs |
| `curl` exits 28 | 1 | retries, then `::error::` + `<empty response>` |

---

# SHOULD-FIX

**S1 — `build/install-netcoredbg.sh:11`: the token guard misdiagnoses its own failure.**
`Authorization: Bearer ${GITHUB_TOKEN:?GITHUB_TOKEN is required}` sits **inside** the retry loop's command substitution. With the variable unset, the subshell dies, and the script then blames rate limiting three times across 6 s of sleeps, terminating with a statement that is simply false:

```
install-netcoredbg.sh: line 12: GITHUB_TOKEN: GITHUB_TOKEN is required
::warning::… GitHub API rate limiting is likely. Retrying in 2s.        (×2)
::error::… GitHub API rate limiting is likely; this is a CI infrastructure failure, not a G# test failure.
GitHub API response body:
<empty response>
```

Impact: CI is unaffected (`build.yml:106-107` always sets it), but this is the **same legibility defect #3051 was filed for**, re-created in a new form for anyone running the installer by hand.

The tempting one-token fix `${GITHUB_TOKEN:-}` is **wrong** — I measured the real API: empty `Bearer` → **HTTP 401**, no header at all → **HTTP 200**. So make the header conditional and drop the loop-internal guard entirely:

```bash
auth=()
[[ -n ${GITHUB_TOKEN:-} ]] && auth=(-H "Authorization: Bearer $GITHUB_TOKEN")
curl -sSL --connect-timeout 15 --max-time 60 ${auth[@]+"${auth[@]}"} "$api_url"
```

Use the `${auth[@]+"${auth[@]}"}` form: under `set -u`, bash 3.2 (still the system bash on macOS — 3.2.57 here) treats a bare `"${auth[@]}"` on an empty array as unbound.

**S2 — `build/install-netcoredbg.sh:61`: successful install reports failure.**
`echo "$install_dir" >> "$GITHUB_PATH"` under `set -u`. Run locally with everything else working:

```
rc=1
Installed NetCoreDbg 3.1.2-1024
install-netcoredbg.sh: line 61: GITHUB_PATH: unbound variable
binary installed? YES
```

The install genuinely succeeded and the script still exited 1. Fix: `>> "${GITHUB_PATH:-/dev/null}"`.

**S3 — the PR body describes only the first commit.**
The body enumerates the installer's behaviours and mentions **none** of `20a3480de`: no layout check, no `--version` invocation, and no mention that **debugger absence is now fatal in CI**. That last one is a change to the e2e job's failure semantics and is exactly what a reader most needs to know. `d11f1ac9b` = 2 files/+50−7; `20a3480de` = 2 files/+23−3, entirely undescribed.

# NIT

**N1 — `e2etests/debugger-e2e.sh:42`** fails open for `CI=1`, `CI=TRUE`, `CI=True` (all → skip, rc=0). Correct for GHA and this repo is GHA-only, and `build.yml:105` is a second line of defence, so there is no live exposure. It would matter only if another CI system is ever added.

**N2 — `e2etests/debugger-e2e.sh:57-65`**: the Darwin/arm64 `skip` at `:62` is **not** CI-gated and still exits 0 silently. Unreachable today (`e2e` is `ubuntu-latest` only), but it is the same class of hole this PR closes; gating it identically would make the invariant uniform.

**N3 — install-dir guidance mismatch**: the installer writes `$HOME/.tools/netcoredbg` (`:41`) while `debugger-e2e.sh:19,:27` falls back to `$ROOT/.tools/netcoredbg`, and the `::error::` at `:41` points developers at the repo-local one. Bridged in CI by `GITHUB_PATH` only.

**N4 — `install-netcoredbg.sh:33`**: when `curl` itself fails (exit 28), the message still asserts rate limiting and dumps `<empty response>`.

## Gates

The PR touches **no** `src/`, `docs/`, `website/`, or `tools/cs2gs` files. On the merged tree, `DiagnosticIdUniquenessTests.cs`, `docs/diagnostics.md`, and `website/docs/ref/diagnostics.md` are all **blob-identical to `main`** — zero docs-gate exposure, and no diagnostic ID is allocated. `cs2gs` passes.

DavidObando:
## Addendum — re-anchored against `main` = `9816fbe55`

My verdict above was anchored at `main` = `f5e9d89250`. `main` has since moved (`01c1a6c5c` #2998, then `9816fbe55` #3062). Re-measured; **verdict unchanged: MERGEABLE**, same 3 SHOULD-FIX / 4 NIT.

- Merge base is **still `46d3cc50f`** — neither landing shifts it.
- True content **still 3 files, +73/−10** (three-dot from the recomputed base).
- Merge against the new `main` is **clean**: `git merge-tree` → `582015bbde`.
- Of the **17** files that landed on `main` since the review, **0** intersect this PR's files. Positive control: the same pattern matches **3 of 3** of the PR's own paths, so that zero is absence, not scope.
- Docs gate on the **new** merged tree: `DiagnosticIdUniquenessTests.cs`, `docs/diagnostics.md`, `website/docs/ref/diagnostics.md` all **blob-identical to `main`** (3 examined, 3 identical).

Worth stating explicitly, because it rules out the #3055 failure mode by construction: the merged tree differs from `main` in **exactly** the three files this PR touches — `.github/workflows/build.yml`, `build/install-netcoredbg.sh`, `e2etests/debugger-e2e.sh`. There is **zero C# delta** between the merged tree and `main`. A merged-tree-only compile or test break of the #3055 shape (a post-merge-base test reflecting into changed code) is therefore not possible here, without needing a build to establish it.

None of the evidence in this review came from a `gsc`/`gsi` probe or from any build, so the stale-`GSharp.Core.dll` trap does not apply to any finding above.

PR #3054, opened 2026-08-01
Improve GS0159 for nullable receivers

### LINKED ISSUE #3034: Improve GS0159 message for invalidated nullable narrowing
Found during Opus 5 review of PR #2963.

## Repro

```gsharp
class C {
    func M() { }
}

func Run() {
    var c C? = nil
    c.M()
}
```

## Actual

`error GS0159: Cannot find function M.`

The message looks like member lookup failed and does not explain that the nullable receiver lacks a valid non-null narrowing. The same wording appears after a previously narrowed receiver is invalidated by assignment.

## Expected

Report that `c` may be nil or that its prior narrowing was invalidated, ideally with a guard/re-narrowing hint. Diagnostic semantics only; no change to flow analysis.

DavidObando:
## Summary

- distinguish nullable-receiver failures from unrelated GS0159 lookup failures
- ask normal binder whether exact call is applicable on underlying non-nullable receiver
- preserve legacy GS0159 for wrong arity, wrong argument type, missing members, and unrelated paths
- support source classes/structs/interfaces, enums, constrained type parameters, extensions, inherited methods, and default interface methods
- preserve qualified receiver spelling and collapse multiline whitespace
- give universally actionable guidance: use `?.` or bind with `if let`

Fixes #3034

## Review blockers

Applicability requires a non-error speculative bind with no new diagnostics. This keeps `Cannot find function M.` for wrong arity and wrong argument type while improving valid calls on nullable receivers.

The latest blocker removed `or re-narrow it before calling`. Assignment-based narrowing is unavailable or ineffective for several receiver kinds, so that advice was false. GS0159 now says:

> Cannot call function M because receiver 'c' may be nil. Use '?.' for a null-safe call or bind it with 'if let'.

## Receiver-kind matrix

`NullableReceiverGuidance_MatchesReceiverNarrowability` covers receiver kind × prior non-null action × valid scope. Local and parameter receivers have no top-level source position; field, global, and `let` receivers cover both top-level and function bodies.

| Receiver | Never narrowed | Non-null immediately before call |
|---|---|---|
| mutable local | GS0159, no re-narrow clause | compiles and runs |
| parameter | GS0159, no clause | read-only assignment diagnostic + GS0159, no clause |
| field | GS0159, no clause | GS0159, no clause |
| global | GS0159, no clause | compiles and runs |
| `let` | GS0159, no clause | GS0159, no clause |

Fresh measurement on `main` `f7d9a5d3` differs from the earlier review report for globals: `c = C(); c.M()` now compiles at top level and inside a function. All three drivers agree.

## Driver matrix

All 48 valid receiver/state/scope/driver cells ran in separately created, asserted-empty directories:

- bare `gsc`: evaluate, success marker `11`
- `gsc /out:`: emit then run, success marker `22`
- `gsi`: interpret, success marker `33`

Expected diagnostic cells returned nonzero with GS0159 and never contained the false clause. Successful local/global narrowing cells returned zero and printed their distinct marker.

## Product mutation

One new product hunk: diagnostic text. Restoring the false re-narrow clause applied exactly once, built successfully, and killed `NullableReceiverGuidance_MatchesReceiverNarrowability`: 13 failed / 3 passed. Passing cells are the three receiver/scope combinations that emit no GS0159 after successful narrowing.

Earlier applicability, receiver-span, whitespace, and diagnostic-rollback mutations remain covered by their focused tests.

## Latest-main merged-tree validation

- head: `a859039d4ad49353fabe58abc140ed0a07d9b28c`
- main / merge base: `f7d9a5d34a35abaeb78818afec20fc60c8c68fed`
- `git merge-tree --write-tree origin/main HEAD`: `53beb1c3d2ed273ccda5f91f7bd9742c84173a85`
- merged tree equals `HEAD^{tree}` exactly
- solution build with warnings as errors: `BUILD_RC=0`, 0 warnings, 0 errors
- full Compiler suite: 4,190 passed
- full Interpreter suite: 729 passed
- full Core suite: 7,089 passed, 1 skipped
- final merged-tree three-driver matrix: 48/48

No interaction claim rests on file lists alone: fetched main was rebased, predicted merged tree was reproduced exactly, then built and tested.

### DIFF
--- a/docs/diagnostics.md
+++ b/docs/diagnostics.md
@@ -121,7 +121,7 @@ IDs may be given as `GS0001`, `0001`, or the bare integer `1`; all three forms a
 | GS0156 | Error | Cannot convert implicitly; explicit conversion exists. | `int x = 3.14` — an explicit cast is available but was not written. |
 | GS0157 | Error | Cannot find type (possibly missing import). | A package-qualified type name that resolves to nothing. |
 | GS0158 | Error | Cannot find member. | A field or property access that does not resolve. |
-| GS0159 | Error | Cannot find function. | A package-qualified function name that resolves to nothing. |
+| GS0159 | Error | Cannot resolve function call. | A function name does not resolve, or its nullable receiver lacks valid non-null narrowing. |
 | GS0160 | Error | Ambiguous overload. | A call that matches more than one overload equally well. |
 | GS0161 | Error | `copy`/`with` receiver is not a `data struct`. | `.copy(…)` used on a plain `struct`. |
 | GS0162 | Error | Named arguments only supported for `data struct` `.copy(…)`. | Named arguments passed to a regular function. |
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs
@@ -87,7 +87,8 @@ internal sealed partial class ExpressionBinder
         BoundExpression receiver,
         ImportedClassSymbol classSymbol,
         ExpressionSyntax rightPart,
-        ExpressionSyntax receiverSyntax = null)
+        ExpressionSyntax receiverSyntax = null,
+        int? receiverStart = null)
     {
         switch (rightPart)
         {
@@ -154,7 +155,8 @@ internal sealed partial class ExpressionBinder
                     return BindNullConditionalAccessExpressionCore(head, nested.RightPart);
                 }
 
-                return BindAccessorStep(head, null, nested.RightPart, nested.LeftPart);
+                var fullReceiverStart = receiverStart ?? receiverSyntax?.Span.Start ?? nested.LeftPart.Span.Start;
+                return BindAccessorStep(head, null, nested.RightPart, nested.LeftPart, fullReceiverStart);
 
             case CallExpressionSyntax ce:
                 // Issue #3084: bind tuple fields symbolically when their CLR
@@ -177,7 +179,7 @@ internal sealed partial class ExpressionBinder
                         nullSafeInvocation: "?(...)");
                 }
 
-                var callResult = BindAccessorCall(receiver, classSymbol, ce, receiverSyntax);
+                var callResult = BindAccessorCall(receiver, classSymbol, ce, receiverSyntax, receiverStart);
                 CheckValueTaskGetAwaiterGetResult(callResult, ce);
                 return callResult;
 
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
@@ -2552,7 +2552,19 @@ internal sealed partial class ExpressionBinder
                 return ifaceObjectCall;
             }
 
-            Diagnostics.ReportUnableToFindFunction(ce.Location, methodName);
+            if (receiver?.Type is NullableTypeSymbol)
+            {
+                var effectiveReceiverSyntax = receiverSyntax ?? receiver.Syntax;
+                var receiverName = effectiveReceiverSyntax == null
+                    ? "receiver"
+                    : effectiveReceiverSyntax.SyntaxTree.Text.ToString(effectiveReceiverSyntax.Span);
+                Diagnostics.ReportUnableToFindFunction(ce.Location, methodName, receiverName);
+            }
+            else
+            {
+                Diagnostics.ReportUnableToFindFunction(ce.Location, methodName);
+            }
+
             return new BoundErrorExpression(null);
         }
 
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
@@ -2555,12 +2555,21 @@ internal sealed partial class ExpressionBinder
 
             var effectiveReceiverSyntax = receiverSyntax ?? receiver?.Syntax;
             if (receiver?.Type is NullableTypeSymbol nullableRecvType
-                && (nullableRecvType.UnderlyingType as StructSymbol)?.GetMethods(methodName).Length > 0
+                && IsApplicableNullableUnderlyingCall(
+                    receiver,
+                    nullableRecvType.UnderlyingType,
+                    methodName,
+                    arguments,
+                    ce,
+                    argumentNames,
+                    explicitTypeArgs,
+                    typeArgSymbols)
                 && effectiveReceiverSyntax != null)
             {
                 var receiverName = effectiveReceiverSyntax.SyntaxTree.Text.ToString(TextSpan.FromBounds(
                     receiverStart ?? effectiveReceiverSyntax.Span.Start,
                     effectiveReceiverSyntax.Span.End));
+                receiverName = string.Join(" ", receiverName.Split((char[])null, StringSplitOptions.RemoveEmptyEntries));
                 Diagnostics.ReportUnableToFindFunction(ce.Location, methodName, receiverName);
             }
             else
@@ -2864,5 +2873,189 @@ internal sealed partial class ExpressionBinder
 
         Diagnostics.ReportUnableToFindFunction(ce.Location, methodName);
         return new BoundErrorExpression(null);
+    }
+
+    private bool IsApplicableNullableUnderlyingCall(
+        BoundExpression receiver,
+        TypeSymbol underlyingType,
+        string methodName,
+        ImmutableArray<BoundExpression> arguments,
+        CallExpressionSyntax ce,
+        ImmutableArray<string> argumentNames,
+        Type[] explicitTypeArgs,
+        ImmutableArray<TypeSymbol> typeArgSymbols)
+    {
+        var narrowedReceiver = new BoundConversionExpression(receiver.Syntax, underlyingType, receiver);
+        var diagnosticMark = Diagnostics.Count;
+
+        // Reuse normal call binding as an applicability probe, then discard its
+        // diagnostics so only the nullable-receiver diagnostic reaches users.
+        bool Succeeded(BoundExpression call) =>
+            call is not null and not BoundErrorExpression && Diagnostics.Count == diagnosticMark;
+
+        try
+        {
+            if (underlyingType is InterfaceSymbol ifaceRecv)
+            {
+                var candidates = TypeMemberModel.GetMethods(ifaceRecv, methodName, MemberQuery.Instance(MemberKinds.Method));
+                if (!candidates.IsDefaultOrEmpty)
+                {
+                    var method = overloads.SelectInstanceOverloadOrReport(candidates, arguments, ce, methodName, argumentNames);
+                    return method != null
+                        && Succeeded(overloads.BindUserInstanceCall(narrowedReceiver, method, arguments, ce, argumentNames));
+                }
+            }
+
+            if (underlyingType is TypeParameterSymbol tpRecv)
+            {
+                var candidates = MemberLookup.CollectSourceInstanceMethods(tpRecv, methodName);
+                if (!candidates.IsDefaultOrEmpty)
+                {
+                    var method = overloads.SelectInstanceOverloadOrReport(candidates, arguments, ce, methodName, argumentNames);
+                    return method != null
+                        && Succeeded(overloads.BindUserInstanceCall(
+                            narrowedReceiver,
+                            method,
+                            arguments,
+                            ce,
+                            argumentNames,
+                            constrainedReceiverTypeParameter: tpRecv));
+                }
+
+                if (tpRecv.ClrInterfaceConstraint != null
+                    && TryBindConstrainedClrInterfaceCall(narrowedReceiver, tpRecv, methodName, arguments, ce, argumentNames, out var constrainedCall))
+                {
+                    return Succeeded(constrainedCall);
+                }
+            }
+
+            if (underlyingType is StructSymbol structRecv)
+            {
+                var candidates = MemberLookup.CollectSourceInstanceMethods(structRecv, methodName);
+                if (!candidates.IsDefaultOrEmpty)
+                {
+                    var method = overloads.SelectInstanceOverloadOrReport(candidates, arguments, ce, methodName, argumentNames);
+                    return method != null
+                        && Succeeded(overloads.BindUserInstanceCall(narrowedReceiver, method, arguments, ce, argumentNames));
+                }
+
+                var defaultInterfaceMethod = TryFindDefaultInterfaceMethod(structRecv, methodName, arguments, ce, argumentNames);
+                if (defaultInterfaceMethod != null)
+                {
+                    return Succeeded(overloads.BindUserInstanceCall(
+                        narrowedReceiver,
+                        defaultInterfaceMethod,
+                        arguments,
+                        ce,
+                        argumentNames));
+                }
+            }
+
+            if (underlyingType is StructSymbol or InterfaceSymbol
+                && TryBindUserDelegateMemberInvocation(
+                    narrowedReceiver,
+                    underlyingType,
+                    methodName,
+                    arguments,
+                    ce,
+                    isStatic: false,
+                    out var delegateCall))
+            {
+                return Succeeded(delegateCall);
+            }
+
+            if (TryBindExtensionFunctionOverload(narrowedReceiver, methodName, arguments, ce, argumentNames, out var extensionCall))
+            {
+                return Succeeded(extensionCall);
+            }
+
+            if (underlyingType is StructSymbol inheritedDerived
+                && (GetInheritedClrBaseType(inheritedDerived) ?? typeof(object)) is Type inheritedBaseClr
+                && TryBindInheritedClrInstanceCall(
+                    narrowedReceiver,
+                    inheritedBaseClr,
+                    methodName,
+                    arguments,
+                    ce,
+                    out var inheritedCall,
+                    explicitTypeArgs,
+                    typeArgSymbols,
+                    argumentNames,
+                    allowProtectedInherited: IsCurrentThisReceiver(receiver)))
+            {
+                return Succeeded(inheritedCall);
+            }
+
+            if (underlyingType is EnumSymbol
+                && TryBindInheritedClrInstanceCall(
+                    narrowedReceiver,
+                    typeof(Enum),
+                    methodName,
+                    arguments,
+                    ce,
+                    out var enumCall,
+                    explicitTypeArgs,
+                    typeArgSymbols,
+                    argumentNames,
+                    mapEnumArgumentsToBaseClr: true))
+            {
+                return Succeeded(enumCall);
+            }
+
+            if (TryBindImportedExtensionCall(
+                narrowedReceiver,
+                methodName,
+                arguments,
+                ce,
+                out var importedExtensionCall,
+                explicitTypeArgs,
+                typeArgSymbols,
+                argumentNames))
+            {
+                return Succeeded(importedExtensionCall);
+            }
+
+            if (underlyingType is InterfaceSymbol importedBaseIface
+                && TryBindInterfaceImportedBaseInstanceCall(
+                    narrowedReceiver,
+                    importedBaseIface,
+                    methodName,
+                    arguments,
+                    ce,
+                    out var importedBaseCall,
+                    explicitTypeArgs,
+                    typeArgSymbols,
+                    argumentNames))
+            {
+                return Succeeded(importedBaseCall);
+            }
+
+            if (underlyingType is TypeParameterSymbol tpObject
+                && TryBindConstrainedObjectMemberCall(
+                    narrowedReceiver,
+                    tpObject,
+                    methodName,
+                    arguments,
+                    ce,
+                    argumentNames,
+                    out var objectCall))
+            {
+                return Succeeded(objectCall);
+            }
+
+            return underlyingType is InterfaceSymbol
+                && TryBindInterfaceObjectMemberCall(
+                    narrowedReceiver,
+                    methodName,
+                    arguments,
+                    ce,
+                    argumentNames,
+                    out var interfaceObjectCall)
+                && Succeeded(interfaceObjectCall);
+        }
+        finally
+        {
+            Diagnostics.TruncateTo(diagnosticMark);
+        }
     }
 }
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
@@ -1913,7 +1913,8 @@ internal sealed partial class ExpressionBinder
         BoundExpression receiver,
         ImportedClassSymbol classSymbol,
         CallExpressionSyntax ce,
-        ExpressionSyntax receiverSyntax = null)
+        ExpressionSyntax receiverSyntax = null,
+        int? receiverStart = null)
     {
         var methodName = ce.Identifier.Text;
         if (string.Equals(methodName, "Invoke", System.StringComparison.Ordinal))
@@ -2552,12 +2553,14 @@ internal sealed partial class ExpressionBinder
                 return ifaceObjectCall;
             }
 
-            if (receiver?.Type is NullableTypeSymbol)
+            var effectiveReceiverSyntax = receiverSyntax ?? receiver?.Syntax;
+            if (receiver?.Type is NullableTypeSymbol nullableRecvType
+                && (nullableRecvType.UnderlyingType as StructSymbol)?.GetMethods(methodName).Length > 0
+                && effectiveReceiverSyntax != null)
             {
-                var effectiveReceiverSyntax = receiverSyntax ?? receiver.Syntax;
-                var receiverName = effectiveReceiverSyntax == null
-                    ? "receiver"
-                    : effectiveReceiverSyntax.SyntaxTree.Text.ToString(effectiveReceiverSyntax.Span);
+                var receiverName = effectiveReceiverSyntax.SyntaxTree.Text.ToString(TextSpan.FromBounds(
+                    receiverStart ?? effectiveReceiverSyntax.Span.Start,
+                    effectiveReceiverSyntax.Span.End));
                 Diagnostics.ReportUnableToFindFunction(ce.Location, methodName, receiverName);
             }
             else
--- a/src/Core/CodeAnalysis/DiagnosticBag.Reports.Expressions.cs
+++ b/src/Core/CodeAnalysis/DiagnosticBag.Reports.Expressions.cs
@@ -317,7 +317,20 @@ public sealed partial class DiagnosticBag
     /// <param name="location">The text location where the error was found.</param>
     /// <param name="text">The text associated to the function.</param>
     public void ReportUnableToFindFunction(TextLocation location, string text)
-    => Report(location, DiagnosticDescriptors.UnableToFindFunction, text);
+    => Report(location, DiagnosticDescriptors.UnableToFindFunction, $"Cannot find function {text}.");
+
+    /// <summary>Reports a function call whose receiver lacks a valid non-null narrowing.</summary>
+    /// <param name="location">The text location where the error was found.</param>
+    /// <param name="functionName">The called function.</param>
+    /// <param name="receiverName">The nullable receiver spelling.</param>
+    public void ReportUnableToFindFunction(
+        TextLocation location,
+        string functionName,
+        string receiverName)
+    => Report(
+        location,
+        DiagnosticDescriptors.UnableToFindFunction,
+        $"Cannot call function {functionName} because receiver '{receiverName}' may be nil. Use '?.' for a null-safe call, bind it with 'if let', or re-narrow it before calling.");
 
     /// <summary>
     /// Reports that an overloaded call (constructor, static method, or
--- a/src/Core/CodeAnalysis/DiagnosticBag.Reports.Expressions.cs
+++ b/src/Core/CodeAnalysis/DiagnosticBag.Reports.Expressions.cs
@@ -330,7 +330,7 @@ public sealed partial class DiagnosticBag
     => Report(
         location,
         DiagnosticDescriptors.UnableToFindFunction,
-        $"Cannot call function {functionName} because receiver '{receiverName}' may be nil. Use '?.' for a null-safe call, bind it with 'if let', or re-narrow it before calling.");
+        $"Cannot call function {functionName} because receiver '{receiverName}' may be nil. Use '?.' for a null-safe call or bind it with 'if let'.");
 
     /// <summary>
     /// Reports that an overloaded call (constructor, static method, or
--- a/src/Core/CodeAnalysis/DiagnosticDescriptors.cs
+++ b/src/Core/CodeAnalysis/DiagnosticDescriptors.cs
@@ -74,7 +74,7 @@ internal static class DiagnosticDescriptors
     internal static readonly DiagnosticDescriptor CannotConvertImplicitly = new("GS0156", DiagnosticSeverity.Error, "Cannot convert type '{0}' to '{1}'. An explicit conversion exists (are you missing a cast?)");
     internal static readonly DiagnosticDescriptor UnableToFindType = new("GS0157", DiagnosticSeverity.Error, "Cannot find type {0}. Are you missing an import?");
     internal static readonly DiagnosticDescriptor UnableToFindMember = new("GS0158", DiagnosticSeverity.Error, "Cannot find member {0}.");
-    internal static readonly DiagnosticDescriptor UnableToFindFunction = new("GS0159", DiagnosticSeverity.Error, "Cannot find function {0}.");
+    internal static readonly DiagnosticDescriptor UnableToFindFunction = new("GS0159", DiagnosticSeverity.Error, "{0}");
     internal static readonly DiagnosticDescriptor AmbiguousOverload = new("GS0160", DiagnosticSeverity.Error, "Call to '{0}' is ambiguous between {1} applicable overloads.{2}");
     internal static readonly DiagnosticDescriptor CopyOrWithNotDataStruct = new("GS0161", DiagnosticSeverity.Error, "copy/with requires a data class or data struct receiver, but got '{0}'.");
     internal static readonly DiagnosticDescriptor NamedArgumentOnlyValidForCopy = new("GS0162", DiagnosticSeverity.Error, "Named arguments are only supported for data-struct .copy(...).");
--- a/test/Compiler.Tests/Binding/Issue3034NullableNarrowingDiagnosticTests.cs
+++ b/test/Compiler.Tests/Binding/Issue3034NullableNarrowingDiagnosticTests.cs
@@ -62,6 +62,46 @@ public class Issue3034NullableNarrowingDiagnosticTests
     }
 
     [Fact]
+    public void MissingMethodOnNullableReceiver_KeepsLookupMessage()
+    {
+        var diagnostic = GetGs0159("""
+            class C {
+                func M() { }
+            }
+
+            func Run() {
+                var c C? = nil
+                c.Frobnicate()
+            }
+            """);
+
+        Assert.Equal("Cannot find function Frobnicate.", diagnostic.Message);
+    }
+
+    [Fact]
+    public void QualifiedNullableReceiver_PreservesQualifier()
+    {
+        var diagnostic = GetGs0159("""
+            class C {
+                func M() { }
+            }
+
+            class Box {
+                var Inner C?
+            }
+
+            func Run() {
+                var b Box = Box()
+                b.Inner.M()
+            }
+            """);
+
+        Assert.Equal(
+            "Cannot call function M because receiver 'b.Inner' may be nil. Use '?.' for a null-safe call, bind it with 'if let', or re-narrow it before calling.",
+            diagnostic.Message);
+    }
+
+    [Fact]
     public void InvalidExplicitTypeArgument_KeepsLookupMessage()
     {
         var diagnostic = GetGs0159("""
--- a/test/Compiler.Tests/Binding/Issue3034NullableNarrowingDiagnosticTests.cs
+++ b/test/Compiler.Tests/Binding/Issue3034NullableNarrowingDiagnosticTests.cs
@@ -38,6 +38,21 @@ public class Issue3034NullableNarrowingDiagnosticTests
             """);
 
         Assert.Equal(NullableReceiverMessage, diagnostic.Message);
+        AssertDiagnosticSpan(diagnostic, line: 6, startCharacter: 6, endCharacter: 9, text: "M()");
+    }
+
+    [Fact]
+    public void TopLevelNullableReceiver_ExplainsMissingNarrowing()
+    {
+        var diagnostic = GetGs0159("""
+            class C { func M() { } }
+
+            var c C? = nil
+            c.M()
+            """, isLibrary: false);
+
+        Assert.Equal(NullableReceiverMessage, diagnostic.Message);
+        AssertDiagnosticSpan(diagnostic, line: 3, startCharacter: 2, endCharacter: 5, text: "M()");
     }
 
     [Fact]
@@ -79,6 +94,125 @@ public class Issue3034NullableNarrowingDiagnosticTests
     }
 
     [Fact]
+    public void WrongArityOnNullableReceiver_KeepsLookupMessage()
+    {
+        var diagnostic = GetGs0159("""
+            class C { func M(value int32) { } }
+
+            func Run() {
+                var c C? = nil
+                c.M()
+            }
+            """);
+
+        Assert.Equal("Cannot find function M.", diagnostic.Message);
+        AssertDiagnosticSpan(diagnostic, line: 4, startCharacter: 6, endCharacter: 9, text: "M()");
+    }
+
+    [Fact]
+    public void WrongArgumentTypeOnNullableReceiver_KeepsLookupMessage()
+    {
+        var diagnostic = GetGs0159("""
+            class C { func M(value int32) { } }
+
+            func Run() {
+                var c C? = nil
+                c.M("x")
+            }
+            """);
+
+        Assert.Equal("Cannot find function M.", diagnostic.Message);
+        AssertDiagnosticSpan(diagnostic, line: 4, startCharacter: 6, endCharacter: 12, text: "M(\"x\")");
+    }
+
+    [Fact]
+    public void StructNullableReceiver_ExplainsMissingNarrowing()
+    {
+        var diagnostic = GetGs0159("""
+            struct S { func M() { } }
+
+            func Run() {
+                var s S? = nil
+                s.M()
+            }
+            """);
+
+        Assert.Equal(
+            "Cannot call function M because receiver 's' may be nil. Use '?.' for a null-safe call, bind it with 'if let', or re-narrow it before calling.",
+            diagnostic.Message);
+    }
+
+    [Fact]
+    public void InterfaceNullableReceiver_ExplainsMissingNarrowing()
+    {
+        var diagnostic = GetGs0159("""
+            interface I { func M(); }
+
+            func Run() {
+                var value I? = nil
+                value.M()
+            }
+            """);
+
+        Assert.Equal(
+            "Cannot call function M because receiver 'value' may be nil. Use '?.' for a null-safe call, bind it with 'if let', or re-narrow it before calling.",
+            diagnostic.Message);
+    }
+
+    [Fact]
+    public void EnumNullableReceiver_ExplainsMissingNarrowing()
+    {
+        var diagnostic = GetGs0159("""
+            enum E { A }
+
+            func Run() {
+                var value E? = nil
+                value.ToString()
+            }
+            """);
+
+        Assert.Equal(
+            "Cannot call function ToString because receiver 'value' may be nil. Use '?.' for a null-safe call, bind it with 'if let', or re-narrow it before calling.",
+            diagnostic.Message);
+    }
+
+    [Fact]
+    public void ConstrainedTypeParameterNullableReceiver_ExplainsMissingNarrowing()
+    {
+        var diagnostic = GetGs0159("""
+            interface I { func M(); }
+
+            func Use[T I](value T?) {
+                value.M()
+            }
+            """);
+
+        Assert.Equal(
+            "Cannot call function M because receiver 'value' may be nil. Use '?.' for a null-safe call, bind it with 'if let', or re-narrow it before calling.",
+            diagnostic.Message);
+    }
+
+    [Fact]
+    public void ImportedAndArrayNullableReceivers_KeepExistingSuccessfulBinding()
+    {
+        Assert.Empty(GetDiagnostics("""
+            import System.Text
+
+            func Run() {
+                var value StringBuilder? = nil
+                value.ToString()
+            }
+            """));
+
+        Assert.Empty(GetDiagnostics("""
+            func Run() {
+                var values []?int32 = nil
+                values.ToString()
+            }
+            """));
+    }
+
+    [Fact]
     public void QualifiedNullableReceiver_PreservesQualifier()
     {
         var diagnostic = GetGs0159("""
@@ -102,9 +236,29 @@ public class Issue3034NullableNarrowingDiagnosticTests
     }
 
     [Fact]
-    public void InvalidExplicitTypeArgument_KeepsLookupMessage()
+    public void MultilineNullableReceiver_CollapsesWhitespace()
     {
         var diagnostic = GetGs0159("""
+            class C { func M() { } }
+            class Box { var Inner C? }
+            func Run() {
+                var b Box = Box()
+                b.
+                    Inner.M()
+            }
+            """);
+
+        Assert.Equal(
+            "Cannot call function M because receiver 'b. Inner' may be nil. Use '?.' for a null-safe call, bind it with 'if let', or re-narrow it before calling.",
+            diagnostic.Message);
+        Assert.DoesNotContain('\n', diagnostic.Message);
+        AssertDiagnosticSpan(diagnostic, line: 5, startCharacter: 14, endCharacter: 17, text: "M()");
+    }
+
+    [Fact]
+    public void InvalidExplicitTypeArgument_KeepsLookupMessage()
+    {
+        var diagnostic = GetGs0159AllowingOtherDiagnostics("""
             import System
 
             func Run() {
@@ -157,15 +311,39 @@ public class Issue3034NullableNarrowingDiagnosticTests
         Assert.Equal("Cannot find function Handler.", diagnostic.Message);
     }
 
-    private static Diagnostic GetGs0159(string source)
+    private static Diagnostic GetGs0159(string source, bool isLibrary = true)
+    {
+        var diagnostic = Assert.Single(GetDiagnostics(source, isLibrary));
+        Assert.Equal("GS0159", diagnostic.Id);
+        return diagnostic;
+    }
+
+    private static Diagnostic GetGs0159AllowingOtherDiagnostics(string source)
+        => Assert.Single(GetDiagnostics(source), diagnostic => diagnostic.Id == "GS0159");
+
+    private static Diagnostic[] GetDiagnostics(string source, bool isLibrary = true)
     {
         var sourceText = SourceText.From(source, "issue3034.gs");
         var tree = SyntaxTree.Parse(sourceText);
-        var compilation = new Compilation(ReferenceResolver.Default(), tree) { IsLibrary = true };
+        var compilation = new Compilation(ReferenceResolver.Default(), tree) { IsLibrary = isLibrary };
 
         using var peStream = new MemoryStream();
         var result = compilation.Emit(peStream, refStream: null);
-        return Assert.Single(result.Diagnostics.Where(diagnostic => diagnostic.Id == "GS0159"));
+        return result.Diagnostics.ToArray();
+    }
+
+    private static void AssertDiagnosticSpan(
+        Diagnostic diagnostic,
+        int line,
+        int startCharacter,
+        int endCharacter,
+        string text)
+    {
+        Assert.Equal(line, diagnostic.Location.StartLine);
+        Assert.Equal(line, diagnostic.Location.EndLine);
+        Assert.Equal(startCharacter, diagnostic.Location.StartCharacter);
+        Assert.Equal(endCharacter, diagnostic.Location.EndCharacter);
+        Assert.Equal(text, diagnostic.Location.Text.ToString(diagnostic.Location.Span));
     }
 }
 
--- a/test/Compiler.Tests/Binding/Issue3034NullableNarrowingDiagnosticTests.cs
+++ b/test/Compiler.Tests/Binding/Issue3034NullableNarrowingDiagnosticTests.cs
@@ -21,37 +21,44 @@ namespace GSharp.Compiler.Tests.Binding;
 public class Issue3034NullableNarrowingDiagnosticTests
 {
     private const string NullableReceiverMessage =
-        "Cannot call function M because receiver 'c' may be nil. Use '?.' for a null-safe call, bind it with 'if let', or re-narrow it before calling.";
+        "Cannot call function M because receiver 'c' may be nil. Use '?.' for a null-safe call or bind it with 'if let'.";
 
     [Theory]
-    [InlineData("function")]
-    [InlineData("top-level")]
-    public void NullableReceiver_ExplainsMissingNarrowing(string scope)
+    [MemberData(nameof(ReceiverKindCases))]
+    public void NullableReceiverGuidance_MatchesReceiverNarrowability(
+        string receiverKind,
+        string state,
+        string scope,
+        bool expectDiagnostic)
     {
-        var source = scope == "top-level"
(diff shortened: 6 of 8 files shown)


DavidObando:
## Verdict: NOT MERGEABLE — one blocker, and it is small

Reviewed at head `cb7bff945` against `main` = `fdba684b6`. Scope from a computed merge base (`f820484c3`): **4 files, +165/-3**, matching `gh pr view`. `git merge-tree --write-tree fdba684b6 cb7bff945` → **rc=0**. 2 commits behind. Build **rc=0, 0W/0E**. PR tests **6/6**. Diagnostics gate **4/4**.

**Boundary check passes.** The diff touches only message plumbing — `ExpressionBinder.Calls.Invocation.cs:2547` reads `receiver?.Type is NullableTypeSymbol` and nothing else; no narrowing state is computed, stored, or altered. No collision with #2963.

**Sibling-site check passes.** `ReportUnableToFindFunction` has 5 call sites on `main`; at head there are 6, and exactly one (`Invocation.cs:2553`) takes the new 3-arg overload. The other five keep the 2-arg form, and the four `*_KeepsLookupMessage` tests pin that they still read `Cannot find function X.` That is the right way to do this, and it directly answers the concern that a message change could bleed into the unrelated `GS0159` situations.

### On "call-site revert: 2 failed / 4 passed" — not a weakness

I reproduced it: reverting `Invocation.cs:2547-2559` to the bare 2-arg call gives **2 failed / 4 passed**, the deaths being `NullableReceiver_ExplainsMissingNarrowing` and `ExpiredBranchNarrowing_ExplainsNeedToRenarrow`. Build clean, `GSharp.Core.dll` md5 `ee01eaf5a01e38231f05f7b2a3c8c315` → `a9e686a46bc7eecb3f45b38d277a7ca8`, so the mutant was live.

That is the expected signature, not a dead hunk. The 4 survivors are negative-space tests for paths the new branch never touches; they are supposed to survive. The call-site hunk is load-bearing and pinned by 2 tests — it should not be deleted.

One caution on the other two numbers: the descriptor revert (6/6) and the reporter mutation (6/6) are **weak signals**. `DiagnosticDescriptors.cs:77` is now `"{0}"` and the sentence lives in the reporter, so those two hunks are coupled by construction — changing either garbles every `GS0159` message and every test dies trivially. A 6/6 there says little about pinning quality.

### BLOCKER — the new message gives actively wrong advice when the method simply does not exist

`Invocation.cs:2547` branches on `receiver?.Type is NullableTypeSymbol` alone. It never checks that `methodName` would resolve on the underlying type. So a plain typo on a nullable receiver is reported as a nullability problem.

```
class C { func M() { } }
func Run() {
    var c C? = nil
    c.Frobnicate()          // Frobnicate does not exist on C at all
}
```

```
error GS0159: Cannot call function Frobnicate because receiver 'c' may be nil.
Use '?.' for a null-safe call, bind it with 'if let', or re-narrow it before calling.
```

Every remedy it offers is useless here. I followed them:

```
c?.Frobnicate()                            -> error GS0159: Cannot find function Frobnicate.
if let bound = c { bound.Frobnicate() }    -> error GS0159: Cannot find function Frobnicate.
```

So the message routes the user through a detour to arrive at the diagnostic they should have been given immediately. Control: the same typo on a non-nullable receiver correctly reports `Cannot find function Frobnicate.` For a PR whose entire purpose is message quality, this makes one case better and a very common case — misspelled or missing method on an optional value — worse.

**And it is pinned by nothing.** I applied a candidate fix and the PR suite could not tell the two versions apart; both are 6/6 green:

```csharp
if (receiver?.Type is NullableTypeSymbol nullableRecvType
    && (nullableRecvType.UnderlyingType as StructSymbol)?.GetMethods(methodName).Length > 0)
```

using `NullableTypeSymbol.UnderlyingType` (`NullableTypeSymbol.cs:30`) and `StructSymbol.GetMethods(string)` (`StructSymbol.cs:1077`). Built **rc=0, 0W/0E**. Re-probed:

| case | head | with fix |
|---|---|---|
| nullable receiver, method exists | new narrowing message | new narrowing message (unchanged) |
| nullable receiver, method missing | narrowing message (**wrong**) | `Cannot find function Frobnicate.` |
| non-nullable receiver, method missing | `Cannot find function Frobnicate.` | unchanged |

and `Issue3034NullableNarrowingDiagnosticTests` stayed **6/6**.

Smallest thing that makes this mergeable: that guard, or your preferred equivalent, plus one test for "nullable receiver, method does not exist on the underlying type → keeps the lookup message". Without the test the two behaviours remain indistinguishable to CI.

### SHOULD-FIX — a qualified receiver loses its qualifier, and no test covers it

`Invocation.cs:2548-2551` renders the span of `receiverSyntax ?? receiver.Syntax`. For a member-access receiver that yields only the trailing name:

```
class Box { var Inner C? }
var b Box = Box()
b.Inner.M()
```
```
error GS0159: Cannot call function M because receiver 'Inner' may be nil. ...
```

The receiver is `b.Inner`, not `Inner`; with more than one `Inner` in scope the message is ambiguous. A call-expression receiver renders fine — `receiver 'GetC()'`. All six tests use a bare identifier, so this shape is unpinned, which is the same pattern as #3039, where a sibling shape rendered badly inside the very diagnostic the PR existed to improve. Please add a qualified-receiver case and widen the span to the full receiver expression.

### SHOULD-FIX — the diagnostics catalogue now describes only half of GS0159

`docs/diagnostics.md:119` and `website/docs/ref/diagnostics.md:119` both read:

```
| GS0159 | Error | Cannot find function. | A package-qualified function name that resolves to nothing. |
```

This PR gives `GS0159` a second, structurally different message and touches no documentation. The gate matches only ID and severity (`DiagnosticIdUniquenessTests.cs:206`) and severity is unchanged, so this stays green on branch and on `main` — I ran it, 4/4. It will not bite CI; it is simply now inaccurate for readers, in both gated files.

### NIT — the descriptor no longer documents its own diagnostic

`DiagnosticDescriptors.cs:77` becomes `new("GS0159", DiagnosticSeverity.Error, "{0}")`, with both sentences moved into `DiagnosticBag.Reports.Expressions.cs:319` and `:326`. It passes format validation, but the descriptor table is the one place a reader can see what an ID says, and `GS0159` now says nothing there. Passing pre-formatted sentences through `{0}` also means placeholder-count checking can never help for this ID again.

### Probe that failed to discriminate

`Invocation.cs:2549-2550` falls back to the literal string `receiver` when both `receiverSyntax` and `receiver.Syntax` are null, which would render `receiver 'receiver' may be nil`. I tried to construct a case that reaches it — synthesized, call-expression, member-access and implicit receivers — and could not get `receiver.Syntax` to be null for any source-bound receiver. So I cannot say whether that fallback is reachable-and-ugly or simply dead. Flagging rather than asserting either way; if it is unreachable it should be an assertion rather than a user-visible string, given #3033 shipped a literal `"?"` to users from exactly this kind of defensive default.

DavidObando:
## Round 2 — NOT MERGEABLE

One **new BLOCKER**. Round-1 blocker gone, threading good, both new tests real. But new guard test wrong shape: it ask "name exist?" when it must ask "call would work?".

Measured: head `d12ba3cf5`, merge base `c63c3ec43` (head 2 ahead / 3 behind), merged tree = `01c1a6c5c` + PR (`MERGE_RC=0`, 7 files, +216/−9). `main` moved `f5e9d8925`→`01c1a6c5c` mid-review; re-merged, not stale run.

---

## (a) Round-1 blocker — FIXED, and narrowing not overshoot

| probe | base `c63c3ec43` | head `d12ba3cf5` |
|---|---|---|
| `c.Frobnicate()` on nullable (typo) | `Cannot find function Frobnicate.` | `Cannot find function Frobnicate.` ✅ |
| `c.M()` genuine nullability, class | `Cannot find function M.` | nullability message ✅ |
| `c.M()` genuine nullability, struct | `Cannot find function M.` | nullability message ✅ |

Narrowing tight. Typo no longer lie. Good.

## (c) Both new tests genuinely pinned — YES

Per-hunk product mutants, condition-inversion/term-removal, one conjunct at a time. `BUILD_RC` checked before hash. Hash on `out/bin/Release/Core/GSharp.Core.dll`, mtime later than build.

| mutant | BUILD_RC | Core.dll md5 | Issue3034 class |
|---|---|---|---|
| clean head | 0 (0 warn) | `3b3a36bd` @15:33 | 8/8 pass |
| **M1** drop conjunct 2 `GetMethods(...).Length > 0` | 0 (0 warn) | `807423fc` @15:51 | **1 fail** — `MissingMethodOnNullableReceiver_KeepsLookupMessage` |
| **M3** drop `receiverStart ??` threading | 0 (0 warn) | `d63fc658` @15:54 | **1 fail** — `QualifiedNullableReceiver_PreservesQualifier` |
| restored | 0 | `3b3a36bd` @15:57 | cycle closed, `porcelain=[]` |

Each mutant kill **exactly one** test, and **different** test. Neither kill come through coupled descriptor/reporter pair — those mutants garble everything, these garble one thing. This real pinning.

Full `Compiler.Tests`: **4011 passed / 0 failed / 0 skipped**, `TEST_RC=0`. Docs gate on **merged** tree: 551 matched / 0 failed (non-empty denominator).

*Disclosure:* after restore I read md5 `366c9a3c` and it look wrong. Cause was `dotnet test` rebuilding **without** `-p:ContinuousIntegrationBuild=true`. Rebuilt with flag → `3b3a36bd`. Restore was always good; reading was bad. Saying so.

**M2** (drop conjunct 3 `effectiveReceiverSyntax != null`) **not run**. Predict survive: round 1 could not construct `receiver.Syntax == null`, so no test can reach it. It defensive null-guard, unpinnable by construction — not a gap.

---

## 🔴 BLOCKER — name-only existence test hide arity and argument-type errors

`src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs:2550` (head `d12ba3cf5`):

```csharp
&& (nullableRecvType.UnderlyingType as StructSymbol)?.GetMethods(methodName).Length > 0
```

`GetMethods(name).Length > 0` only ask **name exist**. Method that exist by name but wrong **arity** or wrong **argument type** pass guard. Then nullability message eat the real error.

**Probe (arity)** — `class C { func M(a int32) }`, `var c C? = nil; c.M()`:

| tree | output |
|---|---|
| base `c63c3ec43` | `GS0159: Cannot find function M.` |
| head `d12ba3cf5` | `GS0159: Cannot call function M because receiver 'c' may be nil. Use '?.' …, 'if let' …, re-narrow` |

**This PR changed it.** Now follow every remedy the message give:

| remedy | result |
|---|---|
| `c?.M()` | `GS0144: Function 'M' requires 1 arguments but was given 0.` |
| `if let cc = c { cc.M() }` | `GS0144: Function 'M' requires 1 arguments but was given 0.` |
| **control**: non-nullable `var c C = C(); c.M()` | `GS0144` **directly** |

**Same shape via argument type** — `c.M("x")`:

| tree | output |
|---|---|
| base | `GS0159: Cannot find function M.` |
| head | nullability message |
| control, non-nullable | `GS0156: Cannot convert type 'string' to 'int32'` |

**Why blocker:** this is round-1's blocker exactly, one step over. Compiler *knows* real error — it print it on non-nullable receiver — but on nullable receiver the new message **suppress** it and send developer down three remedies that all dead-end. Round 1 called that shape blocking. Same shape, same call.

Honest weighting: wrong-arity call on nullable receiver rarer than plain typo, so blast radius smaller than round 1. But fix cheap, so no reason carry it.

**Fix — small, data already in hand.** Guard sit at end of the `TryBind…` chain; `arguments` already in scope. Minimum change kill arity case:

```csharp
&& (nullableRecvType.UnderlyingType as StructSymbol)?.GetMethods(methodName)
       .Any(m => m.Parameters.Length == arguments.Length) == true
```

Argument-type case need real criterion: *would this call bind if receiver were `UnderlyingType`?* That the exact predicate the message assert. Name-existence is proxy for it, and proxy is where both blockers came from. Either tighten to applicability, or accept arity term and say in the PR body that signature mismatch still masked.

---

## 🟡 SHOULD-FIX 1 — multi-line receiver put newline inside diagnostic text

`Invocation.cs:2554-2556` take **raw source span**. Receiver split over lines carry newline + indent into message:

```
b.
    Inner.M()
```
renders (escaped):
```
error GS0159: Cannot call function M because receiver 'b.$
        Inner' may be nil. Use '?.' …
```

One diagnostic now **two lines**. Breaks line-oriented consumers (MSBuild error format, IDE problem matchers, any `grep`-shaped tooling). Fix one line: collapse whitespace runs in `receiverName` before formatting.

## 🟡 SHOULD-FIX 2 — guard fail closed for every non-`StructSymbol` underlying type

`StructSymbol` is `sealed` with zero derived types, so `as StructSymbol` yield null for everything else, `?.` yield null, `null > 0` is false → fall through to lookup message. Measured at head, method **exists**:

| underlying type | head | base | verdict |
|---|---|---|---|
| interface `I?` | `Cannot find function M.` | same | not regression, #3034 unfixed |
| enum `E?` (`.ToString()`) | `Cannot find function ToString.` | same | ditto |
| generic param `func Use[T](x T?)` | `Cannot find function M.` | same | ditto |
| array `[]int32?` | `Cannot find function Frobnicate.` | same | ditto |

**Not a regression** — identical at base, so round-1 blocker did not move, it just did not travel. But #3034 only fixed for `StructSymbol` receivers. Say that in the PR body so nobody think GS0159 nullability coverage is complete.

Side observation, out of scope: `string?` and `StringBuilder?` member call give **rc=0, no diagnostic at all** — nullable imported CLR receiver appear not narrowing-enforced. Not this PR. Worth an issue.

## ⚪ NIT — descriptor `"{0}"` passthrough

`DiagnosticDescriptors.cs:77`. Author's reason (GS0159 has structurally distinct sentences) is acceptable — no placeholder-count check could help two sentences with different arity anyway. Cost is real though: descriptor no longer document itself, so `docs/diagnostics.md:124` is now sole source of truth. It is updated, both files, identical. Accept.

---

## Re-confirms (round-1 facts, survived)

- **Threading is general, not one-shape** (the #3039 worry). `a.B.Inner.M()` → `'a.B.Inner'`; `GetBox().Inner.M()` → `'GetBox().Inner'`; `arr[0].M()` → `'arr[0]'`. Three shapes beyond the one test. Good.
- **No CLR name leak.** `receiverName` is raw source text; all rendered forms are source forms (`a.B.Inner`, `GetBox().Inner`, `arr[0]`). 0 assembly-qualified.
- **No collision with #2963.** Diff only *reads* `receiver?.Type`; threading carry an `int?` offset. No narrowing state computed or stored. Still holds after accessor-chain change.
- **`"receiver"` literal fallback gone, and no NRE possible.** `effectiveReceiverSyntax != null` is a conjunct of the *same* `if`, so `receiverName` cannot run on null; null case take legacy message by construction. Proof is structural, not probe — like round 1, I could not construct `receiver.Syntax == null`.
- **Unrelated GS0159 shapes not turned into nullability.** Top-level unknown function → `GS0130: Function 'Frobnicate' doesn't exist.` (different diagnostic entirely). Fall-through 1-arg reporter renders `Cannot find function X.` unchanged.

---

## Verdict

**NOT MERGEABLE** — BLOCKER above. Fix is one conjunct. Everything else in this PR measure clean, and the mutation evidence is the strongest of any round I have seen on this branch.

DavidObando:
### Addendum — re-measured under the corrected stale-DLL rule, plus full merged-tree suites

Fleet bulletin warned: **a Core-only build leaves the driver's `out/bin/Release/Compiler/GSharp.Core.dll` stale, so a live mutant can read as SURVIVED through `gsc`.** Every probe in my review above went through that driver, so I re-checked all of it.

**Freshness audit.** Both trees show all three copies (`out/bin/Release/Core/`, `out/bin/Release/Compiler/`, `out/obj/Core/Release/`) at **one identical mtime and md5** — the signature of a solution build, never a project-scoped one:

| tree | mtime | md5 |
|---|---|---|
| head `d12ba3cf5` | 17:08:25 (all 3) | `3b3a36bd` |
| base `c63c3ec43` | 15:41:26 (all 3) | `451df102` |

**Re-ran every decisive probe from that verified-fresh state. Nothing moved** — all 19 reproduce byte-for-byte: blocker probes (`p6` arity, `p7` argtype, base-vs-head split), remedy dead-ends (`GS0144` ×3, `GS0156` control), fail-closed set (interface/enum/generic-param/array), threading set (`a.B.Inner`, `GetBox().Inner`, `arr[0]`). **The BLOCKER stands exactly as filed.**

Mutation evidence was never at risk — `dotnet test` builds from source.

### Merged-tree testing — clean

The bulletin's #3055 case (green on branch, green on `main`, **red merged**, via a **private method arity change** invoked reflectively by a test that landed after the merge base) applies to this PR by shape: it adds `int? receiverStart` to `BindAccessorCall` **and** `BindAccessorStep`. Checked directly — the only references to those names in any test project are **comments**, no reflective invocation (positive control: `GetMethod(` reflection does exist elsewhere in the suite, so the pattern is live).

Merged tree = `01c1a6c5c` + PR (`6c2bafc0c`, `MERGE_RC=0`), solution build **0 warnings / 0 errors**, driver copy fresh:

| suite | result |
|---|---|
| Core.Tests | **7065 passed / 0 failed / 1 skipped** (7066) |
| Compiler.Tests | **4034 passed / 0 failed / 0 skipped** |
| Interpreter.Tests | **591 passed / 0 failed** |
| LanguageServer.Tests | **462 passed / 0 failed** |

Compiler.Tests is 4034 merged vs 4011 at head; the +23 are the tests that landed on `main` after the merge base — delta reconciles, so the merged run really did pick up the newer tests.

**Verdict unchanged: NOT MERGEABLE**, on the single BLOCKER above. Everything else measures clean on the merged tree. Note that fixing the guard invalidates all of this — the mutation table and merged-tree runs will need redoing at the new head.

DavidObando:
## Addendum — I inherited the test shape, re-measured at top level, and found a second instance. Verdict unchanged: NOT MERGEABLE. **The one-line fix I proposed above is insufficient.**

Disclosure first: every probe in my review above ran inside a function body, because `Issue3034NullableNarrowingDiagnosticTests.cs` builds all six of its sources in-function and I copied that shape. I never ran `gsi` at all. Re-measured across `{top-level, in-function}` x `{gsc, gsi}`.

Measured at head `cb7bff945` and at this PR's merge base `f820484c3`, both solution builds **rc=0, 0W/0E**, artifacts verified fresh in `Core/`, `Compiler/` and `Repl/` immediately before each probe. `merge-tree --write-tree 185431c4 cb7bff945` gives **rc=0**.

### The finding: identical code, accepted in-function, rejected at file scope

```
class C { func M() { Console.WriteLine("M ran") } }
var c C? = nil
c = C()          // re-narrowed by assignment
c.M()
```

| shape | BASE gsc | HEAD gsc | HEAD gsi |
|---|---|---|---|
| top-level | rc1 `GS0159: Cannot find function M.` | rc1 **new nullable message** | rc1 **new nullable message** |
| same code inside a func | rc0 | rc0 | rc0, prints `M ran` |

**The rejection is pre-existing — this PR does not introduce a source break.** The base already rejects the top-level form. So there is no inversion here, and that is worth stating plainly.

**What the PR does introduce is advice that cannot be followed.** The old text was merely unhelpful. The new text ends with "or re-narrow it before calling" — and the user in this sample **did** re-narrow, with `c = C()`. It cannot work, by construction. `StatementBinder.Narrowing.cs:353` reads:

```csharp
// Narrow locals only — never parameters, fields, or read-only `let` locals
if (assign.Variable is not LocalVariableSymbol l || l.IsReadOnly)
```

with the same exclusion again at `StatementBinder.Narrowing.cs:463` (`path.Root is not LocalVariableSymbol local`). A top-level `var` is a `GlobalVariableSymbol`, so assignment-narrowing is skipped deliberately. The compiler is telling the user to do the one thing it has decided not to honour for that symbol class.

Not all three remedies are dead — I checked each rather than assuming. At top level, binding with `if let` compiles **rc=0** under both `gsc` and `gsi`. So the null-safe call and the `if let` bind are fine; **only the re-narrow clause is false**, and only for non-local receivers.

### Consequence: my proposed fix above is incomplete — please do not just apply it

The guard I suggested keyed on whether the method exists. In this sample `M` **does** exist on `C`, so that guard keeps the new message and the bad advice survives. Two distinct defects share one message:

1. the method does not exist at all, so all three remedies are useless — my original blocker, and it reproduces at top level too, rc=1 under both drivers;
2. the receiver is not a narrowable symbol, so the re-narrow clause specifically is false.

A complete fix needs the remedy list conditioned on both facts. At minimum, suppress "or re-narrow it before calling" when the receiver root is not a non-readonly `LocalVariableSymbol`, reusing the same predicate as `Narrowing.cs:353` so that the message and the analysis cannot drift apart.

### And the tests need an axis, not one more case

All six tests in `Issue3034NullableNarrowingDiagnosticTests.cs` build in-function, so the top-level behaviour of the very diagnostic this PR rewrites is pinned by nothing — the suite is 6/6 green with the bad advice in place. Please parameterise the fixture over `{top-level, in-function}` rather than appending another in-function case. `gsi` only ever executes top-level code, so an in-function-only fixture leaves the interpreter untested by definition.

### Unchanged from my review above

The call-site hunk is load-bearing — reverting it kills 2 tests — and should not be deleted. There is no narrowing-analysis change, so no collision with #2963. Sibling `GS0159` sites keep their wording. The docs rows at `docs/diagnostics.md:119` and `website/docs/ref/diagnostics.md:119` still describe only the old message, which will not red the gate since it matches ID and severity only and severity is unchanged, but both files are now inaccurate.

These numbers are true of `cb7bff945` merged against `185431c4`. `main` moves, so please do not cite them for different code.

DavidObando:
## Re-review at the NEW head `92114c5f0` — three of my four findings are fixed. One blocker remains, and the new test now pins it as correct.

My earlier two comments measured `cb7bff945`, which is no longer the head. Re-measured from scratch.

Scope from a computed merge base (`5ad0109c`, three-dot, `--full-index`): **7 files, +587 −9** — grown from 4/+165−3. `git merge-tree --write-tree 33bea4af 92114c5f0` → **rc=0**. Solution build **rc=0, 0W/0E**; `Core/`, `Compiler/` and `Repl/` mtimes verified fresh before every probe. `Issue3034NullableNarrowingDiagnosticTests` **17/17**. `DiagnosticIdUniquenessTests` **4/4**.

Every probe below was compiled **into its own empty directory with a unique package name** — I had previously reused one directory and one package name for all probes, which can silently serve a stale assembly. Re-running isolated changed nothing, but the earlier numbers should be read as confirmed-by-re-run rather than as-measured.

### Fixed — confirmed by probe

**1. Typo on a nullable receiver no longer mis-reported.** Was my blocker.
```
var c C? = nil
c.Frobnicate()      // top level
```
→ `GS0159: Cannot find function Frobnicate.` Correct, and it holds at top level as well as in-function, across `gsc /out:`, bare `gsc`, and `gsi` (rc=1 in all three).

**2. Qualified receivers keep their qualifier.** `b.Inner.M()` now renders `receiver 'b.Inner'`, previously `receiver 'Inner'`.

**3. Docs.** `docs/diagnostics.md:124` and `website/docs/ref/diagnostics.md:124` are both updated, identically, to `| GS0159 | Error | Cannot resolve function call. | A function name does not resolve, or its nullable receiver lacks valid non-null narrowing. |` — bare ID so the gate sees it, severity unchanged, both files in sync. Gate passes 4/4.

### BLOCKER (remaining) — "or re-narrow it before calling" is false for every receiver except a non-readonly local

`StatementBinder.Narrowing.cs:353` restricts assignment narrowing deliberately:

```csharp
// Narrow locals only — never parameters, fields, or read-only `let` locals
if (assign.Variable is not LocalVariableSymbol l || l.IsReadOnly)
```

with the same exclusion at `Narrowing.cs:463` (`path.HasMembers || path.Root is not LocalVariableSymbol local`). So the remedy the message names is unavailable for globals, fields, parameters and `let` locals. Two probes, each re-narrowing on the line immediately before the call:

**global receiver (top level):**
```
var c C? = nil
c = C()
c.M()
```
→ rc=1 `Cannot call function M because receiver 'c' may be nil. Use '?.' ... or re-narrow it before calling.`

**field receiver (in a func):**
```
var b Box = Box()
b.Inner = C()
b.Inner.M()
```
→ rc=1, same message, `receiver 'b.Inner'`.

Identical in all three drivers. The control discriminates cleanly: the same code with a **local** receiver inside a func compiles **rc=0** and prints `M ran`, because there the advice is true. So the sentence is correct for exactly one symbol class and false for the rest.

To be explicit about attribution: **the rejection itself is pre-existing** — the base rejects both shapes too, with the old `Cannot find function M.` This PR does not introduce a source break. What it introduces is a remedy list that the analysis has decided not to honour for these receivers.

### The new top-level test freezes the defect as expected behaviour

`Issue3034NullableNarrowingDiagnosticTests.cs:45` `TopLevelNullableReceiver_ExplainsMissingNarrowing` uses the **never-narrowed** shape:

```
var c C? = nil
c.M()
```

and asserts the full `NullableReceiverMessage` constant, which contains "or re-narrow it before calling" — at global scope, where re-narrowing cannot work. The assertion is green, and it now locks in the advice for precisely the symbol class that cannot follow it. This is the pattern from #3039, where a sibling shape got frozen inside the diagnostic the PR existed to improve.

Adding the one top-level case was the right instinct, but it is a case, not an axis: 16 of the 17 tests build in-function, and the one top-level test happens to use the shape where the wrong clause is invisible.

### Smallest change that makes this mergeable

Condition the final clause on the receiver root being a non-readonly `LocalVariableSymbol`, reusing the same predicate as `Narrowing.cs:353` so the message cannot drift from the analysis — emit "Use '?.' for a null-safe call, or bind it with 'if let'" otherwise. Then update `:45` to assert that global-scope form, and add the assignment-then-call shape for a global and for a field, since those are the two that currently render false advice.

All numbers here are true of `92114c5f0` merged against `33bea4af`; `main` moves, so please do not cite them for different code.

DavidObando:
## Verdict: **MERGEABLE**

Numbers true of **main `33bea4af`**, head **`92114c5f0`**, merge-base **`5ad0109c`**. `git merge-tree` vs `33bea4af` → **rc=0**. Merged-tree build **rc=0, 0 Warning(s), 0 Error(s)**.

### Lead answer: this is **not** message-only

| subtree | main `33bea4af` | merged |
|---|---|---|
| `src` | `356017da…` | **`e7514b70…`** differs |
| `samples` (control) | `303cc22a…` | `303cc22a…` identical |

Production changes **+237/−9**, of which **212 lines** land in `ExpressionBinder.Calls.Invocation.cs`. The PR body is honest about this ("use nullable guidance only when the underlying source type defines the called method", "leave narrowing analysis unchanged") — both claims verified below — but anyone treating this as a string edit is mis-scoping it.

### Did the set of programs emitting GS0159 change? **No.**

Structural: the new branch sits at `ExpressionBinder.Calls.Invocation.cs:2549-2570`, immediately before `return new BoundErrorExpression(null)` at `:2572`. **Both** arms call `ReportUnableToFindFunction`, so only the text is selected there. The applicability probe marks the bag at `:2881` and rolls it back in `finally { Diagnostics.TruncateTo(diagnosticMark); }` at `:3048-3051`, so `SelectInstanceOverloadOrReport` at `:2895/:2906/:2929` cannot leak speculative diagnostics. Speculative bind side-effects cannot reach emit because the path already returns an error expression.

Measured, on the **evaluate** column that the samples gate never traverses (`SampleConformanceTests.cs:30-58` is emit-only):

- **122/122 `samples/*.gs`, rc and diagnostic-ID set byte-identical base vs merged.**
- That corpus is not vacuous: **23 samples emit diagnostics, 11 are rc≠0, and 3 emit `GS0159`** (`GsharpExtensions{Mixed,Optional,Sequences}.gs`).
- On those real GS0159 sites the rendered message is **byte-identical** base vs merged (`Cannot find function Map.` / `WriteLine.` / `FlatMap.` / `Take.`) — "preserve existing wording at unrelated paths" holds on real code, not just unit tests.

Suites on the merged tree: **Core.Tests 7072 passed / 1 skipped / 0 failed**; the **12** Compiler.Tests classes that assert the old wording **109/109**; `cs2gs` `Issue2638` (asserts `DoesNotContain("GS0159")`) **1/1**. Those 12 + 18 Core.Tests + 1 cs2gs are the full population of **36** files asserting `"Cannot find function"`. I did not run full Compiler.Tests (~1h20m); I filtered to the classes that could observe this change and state that explicitly.

### Is the message true and useful? **Yes — verified on nested and invalidated-narrowing cases.**

| probe | base | merged |
|---|---|---|
| p1 issue repro, in-function | `Cannot find function M.` | `…receiver 'c' may be nil…` |
| p1t **top-level twin** | `Cannot find function M.` | `…receiver 'c' may be nil…` |
| p2 **nested** `o.inner.M()` | `Cannot find function M.` | **`…receiver 'o.inner' may be nil…`** |
| p5 narrowing invalidated by assignment | `Cannot find function M.` | `…receiver 'c' may be nil…` |
| p6 same, inside `for i in 1 ... 3` | `Cannot find function M.` | `…receiver 'c' may be nil…` |
| p7 nullable receiver, **method absent** | `Cannot find function Frobnicate.` | `Cannot find function Frobnicate.` |

rc identical base vs merged in every row. p2 confirms the `receiverStart` propagation (`ExpressionBinder.Access.MemberLookup.cs:136-137`) yields the **full qualified spelling**, not just the last segment. p7 confirms the guard keeps the message **truthful**: nullability is only blamed when the method genuinely exists on the underlying type.

Both remedies the message offers are real and compile: `?.` and `if let` (probe p4, rc=0; `if let` is ADR-0151, `samples/IfLetGuardLet.gs:15,28`).

Axis is complete — `{top-level, in-function}` × `{gsi, gsc evaluate, gsc /out: emit}` = 6 cells, all rc=1 with the new message. The author's suite has **17 `[Fact]` and 0 `func Main()`** — positive-controlled (`func ` matches 28× in the same file), so no vacuous-`gsi` shape.

### Is it pinned? **Yes — two product mutants, both round-tripped.**

Every mutant rebuilt the full solution (rc=0) with mtime verified fresh on **`Core/`, `Compiler/` and `Repl/`**, `ref/` excluded.

- **Mutant A**, message text at `DiagnosticBag.Reports.Expressions.cs:333` (`may be nil` → `may be MUTANT`). Product-visible in shipped `gsc` output. **9 of 17 failed.** The 8 survivors are exactly the preservation tests asserting the old wording.
- **Mutant B**, condition inversion of the method-existence guard at `ExpressionBinder.Calls.Invocation.cs:2550`. Discriminating in **both** directions: p1 reverted to the old wording *and* p7 falsely gained the nullability message. **12 of 17 failed.**
- Restored, rebuilt, round-tripped: p1 and p7 both flipped back, **17/17**.

Assertions are the strong form — `Assert.Equal` on the full message, not `Contains("GS0159")` — and `:254` pins single-line rendering. **No existing test was weakened or deleted: no pre-existing test file appears in the diff at all** (the only test change is the new +355 file).

### Docs gate (merged tree, bidirectional)

`docs/diagnostics.md` **384 → 384**; `website/docs/ref/diagnostics.md` **382 → 382**; **zero IDs added, zero removed**; the `GS0159` row stays bare-pipe gate-visible (`docs/diagnostics.md:124`). Both catalogues updated consistently.

### Diagnostic IDs

**No new ID allocated.** The only ID in the descriptor diff is `GS0159` itself. No contact with the taken/unsafe list.

### NITs (none blocking)

1. **`DiagnosticDescriptors.cs:77` is now a bare pass-through `"{0}"`.** Message construction moved to the two helpers at `DiagnosticBag.Reports.Expressions.cs:319/326`. I verified the full population — **6** call sites (`ExpressionBinder.Calls.Invocation.cs:1962,2266,2565,2569,2866`, `ExpressionBinder.Calls.Arguments.cs:809`), all routed through the helpers, no direct `Report(..., UnableToFindFunction, ...)` bypass, and no other construction path for `"GS0159"` in `src/`. So it is correct today. But the descriptor no longer carries a readable template, and a future call site passing a bare identifier would emit that token as the entire message with **no compile error**. A one-line comment on `:77` saying the text is owned by the helpers would keep the next author out of the trap.
2. **Author's mutation numbers do not match mine.** The body reports "method-existence guard mutation — 1 failed, 7 passed"; I measure **12 failed / 17 examined** for the guard inversion and **9 / 17** for the message. Not a defect — different mutation site and filter — but the population should be stated alongside the count, since "1 failed, 7 passed" reads as a much weaker result than the guard actually has.
3. **Pre-existing, explicitly NOT this PR:** narrowing survives a loop back-edge when the invalidating assignment *follows* the call — `if c != nil { for i in 1 ... 3 { c.M(); c = nil } }` compiles **rc=0 on base and merged alike** (probe p3). Identical both sides, so this PR neither causes nor changes it, and "leave narrowing analysis unchanged" is confirmed. Recording it only so it is not later attributed to this change; it may deserve its own issue.

DavidObando:

Anchors: PR head **`8ef6afa9`** (unmoved). `main` moved **`4469e323` → `f7d9a5d3`** mid-review (#3123, #3052), so all numbers below are re-anchored on the **merged tree `53beb1c3`** at `f7d9a5d3`, not on the stale fast-forward.

### Headline: this is a BEHAVIOUR change, not a message change — scope it accordingly

Do not review this as a string edit. `ExpressionBinder.Calls.Invocation.cs` `:2878` adds `IsApplicableNullableUnderlyingCall`, a **real speculative re-bind** that calls `BindUserInstanceCall` and rolls back via `Diagnostics.TruncateTo(diagnosticMark)` in a `finally` at `:3058` (established idiom — same pattern at `Accessor.cs:1087`, `IfLet.cs:145`). 168 added non-comment lines of binder logic. `DiagnosticDescriptors.cs:77` is now a bare `"{0}"` passthrough; message text moved into `DiagnosticBag.Reports.Expressions.cs`.

### The prior blocker: **SPLIT ANSWER** — half of it evaporated rather than being fixed

The earlier blocker ("*or re-narrow it before calling*" is false for non-local receivers) had two probes. They have different fates, and a reviewer re-running only the first would have concluded "fixed" for the wrong reason:

| probe | status |
|---|---|
| **global receiver re-narrowed** | **Evaporated, not fixed.** Compiles rc=0 on *both* base and head (prints `M ran`). `StatementBinder.Narrowing.cs` changed on `main` via exactly one commit — `ea35b2e6` (**#2963**). Nothing this PR did. |
| **field receiver** | **Genuinely fixed.** Still raises the message; `8ef6afa9` deletes the false clause (one line, `DiagnosticBag.Reports.Expressions.cs:333`). Now: `…Use '?.' for a null-safe call or bind it with 'if let'.` |

The fix chose **deletion over conditioning** — smaller, and avoids maintaining two message variants.

### Message truthfulness — verified by construction, all receiver classes, all three drivers

| receiver | gets nullability message | `?.` works | `if let` works |
|---|---|---|---|
| field `b.Inner` | yes (`receiver 'b.Inner'`) | rc=0 `M ran` | rc=0 `M ran` |
| parameter | yes | rc=0 | rc=0 |
| `let` local | yes | rc=0 | rc=0 |
| **interface** | yes (`receiver 'x'`) — **base said `Cannot find function M.`** | — | — |
| global | no longer (post-#2963) | — | — |

The interface row closes the earlier SHOULD-FIX that the old guard "failed closed for non-`StructSymbol` receivers". Every advice sentence is true of the situation that raises it.

Prior blockers re-probed behaviourally, all identical to base: arity mismatch → `Cannot find function M.`; argument-type mismatch → same; typo → `Cannot find function Frobnicate.`

### Mutation — per hunk, product mutants, both applied-asserted

- **M1** — re-add the deleted false clause (`DiagnosticBag.Reports.Expressions.cs:333`). Pattern count 1, exactly 1 line changed, driver md5 moved, mtime fresh on `Core/`+`Compiler/`+`Repl/`. **Kills 22/39** (20/33 Compiler + 2/6 Interpreter).
- **M2** — the +188-line hunk was pinned by nothing, so I mutated the applicability probe itself: `Succeeded` → always true, reverting to name-existence semantics (the original blocker). `BUILD_RC=0`, md5 `bd0db8b3` ≠ clean, mtime fresh. **Kills 8/39**, and precisely the right ones:
  `WrongArityOnNullableReceiver_KeepsLookupMessage` (both scopes), `WrongArgumentTypeOnNullableReceiver_KeepsLookupMessage` (both scopes), and 4 interpreter `NullableReceiverDiagnostic_MatchesCompiler` cases.

Interpreter deaths in **both** mutants mean the `gsi` column is genuinely pinned, not decorative.

### False-positive corpus — clean

122/122 `samples/*.gs` on the bare-`gsc` **evaluate** column (the one the samples gate never traverses — `SampleConformanceTests.cs:30-58` is emit-only): rc and diagnostic-ID set identical base vs head, `DIFF_RC=0`. Positive-controlled: 13 rc≠0, 25 emit diagnostics, 3 emit `GS0159`. Because an ID-set comparison is blind to *text* — which is exactly what this PR changes — I also compared **rendered text** on those 3: byte-identical.

### Full suites — not a filtered green

`Core.Tests` **7089 passed / 1 skipped**, `Compiler.Tests` **4185 passed** (1h6m), `Interpreter.Tests` **720 passed**. All three run in full, because **33 files assert the old `"Cannot find function"` wording** (14 Compiler, 18 Core, 1 Interpreter) and a filter would have undercounted.

### Merge interaction — checked, because `main` moved onto the same files

Both new `main` commits touch `DiagnosticDescriptors.cs`, `docs/diagnostics.md` and `website/docs/ref/diagnostics.md` — all three also edited here. `merge-tree` predicted `53beb1c3`; the real merge **reproduced that hash exactly** (asserted by tree comparison, not by rc). **`BUILD_RC=0`, 0 warnings, 0 errors.** ID-uniqueness gate 4/4. Issue3034 suites 33/33 and 6/6 on the merged tree. All behavioural probes re-run there, byte-identical across all three drivers.

### Docs and IDs — clean

Both catalogue rows updated and **byte-identical to each other** after the merge (`:124` in both). This matters because the gate only compares ID + severity, never message text (`DiagnosticIdUniquenessTests.cs:186-192`) — so I hand-diffed. **No new diagnostic ID**; descriptor diff is only `GS0159` → `"{0}"`. No self-allocation.

### Location — real, not fabricated

`(5,3,5,6)`, `(8,13,8,16)`, `(4,20,4,23)`, `(6,7,6,10)`, and a deliberate 45-line file → **`(45,13,45,16)`** with filename. Line *and* column both move.

---

## NITs

1. **`ExpressionBinder.Access.MemberLookup.cs` (+8) and the `:1913` (+1) hunk are unmutated.** M1 and M2 pin the message and the applicability probe; these two remain unpinned by anything I ran.
2. **Stale test name** — `ExpiredBranchNarrowing_ExplainsNeedToRenarrow` still says "Renarrow" though the message no longer mentions re-narrowing.
3. **`DiagnosticDescriptors.cs:77` no longer documents its own diagnostic.** With a bare `"{0}"`, a future call site passing a raw identifier would emit that token as the entire message, with no compile error to catch it.
4. **Multi-line receiver renders as `'b .Inner'`** (extra space) when the call is split across lines. Whitespace is collapsed, so the earlier concern about a raw newline entering a one-line diagnostic does **not** reproduce — this is cosmetic only.

## Disclosed gaps

- **Generic type-parameter receiver: unmeasured by me.** My probe used wrong call syntax and died at `GS0005` — a no-result, not a pass. The suite's `ConstrainedTypeParameterNullableReceiver_ExplainsMissingNarrowing` does die under M1, so the shape is pinned by the PR's own tests.
- Full `Interpreter.Tests` was run on a second clone at the same head SHA, not in the primary head tree.
- The 122-sample false-positive corpus was measured base-`4469e323` vs head; on the merged tree I re-ran the targeted probes, not the full 122.

Verdict: **MERGEABLE.** The prior blocker is genuinely fixed for the shape that still reproduces, the fix is minimal (one deleted line), the behaviour change underneath it is real but correct and now mutation-pinned, and nothing in 11,994 tests or 122 samples regressed.

DavidObando:

Fresh full review of head **`a859039d`**, not a delta review.

**Re-anchored, and that mattered.** The branch was **force-pushed** since the last review — `8ef6afa9` no longer exists, so no prior number is reachable from this head. Independently, **`main` moved `f7d9a5d3` → `27a05798`** (5 commits), and one of them (**#3019**, `46f2c929`) adds **+169 lines to `ExpressionBinder.Calls.Invocation.cs` — the exact file this PR adds 212 lines to**. Both prior MERGEABLE verdicts and the PR body's "Latest-main merged-tree validation" section are anchored at a merge base that is now stale, so I re-measured everything from scratch.

- merge base `f7d9a5d3`; three-dot `--full-index`: **8 files, +765/−9** (production **+236/−7**)
- `git merge-tree` **output** (not rc): one line, predicted tree `674d7792`, **zero conflict lines**
- real merge reproduced **`674d7792` exactly** (tree-hash equality, not rc)
- merged tree `6f7e2c5d`: **`BUILD_RC=0`, 0 Error(s)**
- freshness verified by **mtime** on `Core/`, `Compiler/` **and** `Repl/`, `ref/` excluded — all copies at one timestamp per tree

Main's additions to that file are delegate-identity statics (`UnwrapExpressionTreeDelegate`, `SameDelegateIdentity`, …) and sit at lines 814–1804, disjoint from this PR's work at `:2015+`. **No semantic collision** — checked by reading the overlap, not by trusting the clean merge.

### Issue #3034's own decision record

Zero comments on the issue, so nothing supersedes the PR's framing. The body does set one hard invariant — *"Diagnostic semantics only; no change to flow analysis."* **Verified: 122/122 samples byte-identical in rc and diagnostic-ID set** (below), and the repro's rc is unchanged 1 → 1. No source break in either direction.

**Issue repro, run verbatim first.** Unlike #3116, it compiles as written:

| driver | base `27a05798` | merged |
|---|---|---|
| bare `gsc` evaluate | rc1 `GS0159: Cannot find function M.` | rc1 `…receiver 'c' may be nil. Use '?.' … or bind it with 'if let'.` |
| `gsc /out:` emit | identical | identical |
| `gsi` interpret | identical | identical |

**All three drivers speak identically in every probe I ran — no column divergence anywhere.**

---

## SHOULD-FIX — the suite pins that the message is *printed*, never that it is *true*

This is the one substantive gap, and it is precisely the risk this PR exists to manage.

The message makes an actionable promise: *"Use `?.` for a null-safe call or bind it with `if let`."* **All 8 occurrences of `?.` and `if let` in `test/Compiler.Tests/Binding/Issue3034NullableNarrowingDiagnosticTests.cs` are inside asserted message strings** (`:24, :59, :177, :194, :211, :227, :270, :288`). **Zero tests compile either remedy.** The suite would stay 33/33 green if `?.` stopped working for interface receivers tomorrow.

Prior reviews measured the remedies for three receiver kinds only (field, parameter, `let` local — all plain struct/class), and left interface as `—` and generic type parameter explicitly disclosed as unmeasured. This PR's applicability probe **added** interface, enum, type-parameter, extension, inherited-CLR and default-interface branches. I measured all of them:

| receiver kind | message raised | `?.` compiles | `if let` compiles |
|---|---|---|---|
| interface `I?` | yes, `receiver 'value'` | **rc=0** | **rc=0** |
| enum `E?` (`.ToString()`) | yes | **rc=0** | **rc=0** |
| constrained type param `T?` (`func Use[T I]`) | yes | **rc=0** | **rc=0** |
| extension / receiver-clause func | yes | **rc=0** | **rc=0** |

All rc=0 on **all three drivers**. **So the advice is true today for every receiver kind I could reach** — including the generic type-parameter shape a prior reviewer could not measure (their probe died at `GS0005`; mine used `func Use[T I](value T?)` and ran). The finding is not that the message lies; it is that **nothing in CI keeps it honest**. Recommend one test per receiver kind that actually compiles the two remedies.

## Prior blockers — verified fixed at this head, by probe

| prior blocker | status at `a859039d` |
|---|---|
| typo (`c.Frobnicate()`) mis-blamed on nullability | **fixed** — `Cannot find function Frobnicate.` |
| wrong **arity** mis-blamed | **fixed** — legacy message; killed by mutant M2's sibling term |
| wrong **argument type** mis-blamed | **fixed** — legacy message; **M2 kills exactly these** |
| false *"or re-narrow it before calling"* clause | **fixed by deletion** — the clause is gone from `DiagnosticBag.Reports.Expressions.cs:333` |

**Misattribution hunt — the guard fails closed, correctly.** A private interface helper is the sharpest case (`GS0334` exists for it):

```
interface I { private func Helper() { } func M() { } }
var value I? = nil
value.Helper()
```
→ `GS0159: Cannot find function Helper.` — **legacy wording, identical on base and merged.** The PR does **not** blame nullability for an accessibility failure. Control: the same call on a non-nullable receiver gives `GS0334`, and `if let` on the nullable one also gives `GS0334`.

## Narrowing-exclusion set — measured independently, not inherited

| receiver | never narrowed | assigned non-nil immediately before call |
|---|---|---|
| mutable local | GS0159 | **compiles, prints `11`** |
| global | GS0159 | **compiles, prints `22`** |
| field | GS0159 | **GS0159** (`receiver 'b.Inner'`) |
| parameter | GS0159 | GS0159 |
| `let` local | GS0159 | GS0159 |

Distinct markers per arm, identical under `gsc` evaluate and `gsi`. **`global` narrows successfully** — so the exclusion set is {field, parameter, `let`}, not the four commonly cited. This matches the PR body's matrix exactly; **the body's numbers are true**, and it is honest about the correction. Because the message no longer names re-narrowing, the field row is no longer false advice — deletion was the right fix, and smaller than conditioning.

## Mutation — 5 product mutants, per hunk, **5 kills, 0 survivors**

Every mutant asserted applied (single-site, `MUTANT-APPLIED`), `BUILD_RC=0`, `Core.dll` md5 moved, mtime fresh on `Core/`+`Compiler/`+`Repl/`. Clean md5 `07a8d318`; restored and rebuilt → **`07a8d318` round-trip closed**, `git status --porcelain` empty.

| mutant | hunk | md5 | kills |
|---|---|---|---|
| **M1** `may be nil`→`may be MUTANT` | `DiagnosticBag.Reports.Expressions.cs:333` | `f05e6d06` | **20/33 Compiler + 2/6 Interpreter** |
| **M2** drop `&& Diagnostics.Count == diagnosticMark` | applicability probe, `Invocation.cs:2889` | `91343a6b` | **2 + 2** — exactly `WrongArgumentTypeOnNullableReceiver_KeepsLookupMessage` (both scopes) and its 2 interpreter twins |
| **M3** drop `receiverStart ??` at render | `Invocation.cs:2673` | `9e54aab8` | **6** — 4 `field` cells + `QualifiedNullableReceiver_PreservesQualifier` + `MultilineNullableReceiver_CollapsesWhitespace` |
| **M4** remove whitespace collapse | `Invocation.cs:2675` | `876458ef` | **1** — `MultilineNullableReceiver_CollapsesWhitespace` |
| **M5** drop `receiverStart ?? receiverSyntax?.Span.Start ??` propagation | `ExpressionBinder.Access.MemberLookup.cs:158` | `5e8e231b` | **6** — same field/qualifier set |

Two things worth naming:

- **M5 corrects a prior review's NIT.** The `MemberLookup.cs` hunk was reported as unmutated and unpinned; it is **pinned by 6 tests**.
- **M2's isolation is precise.** Arity survives M2 and dies elsewhere, because arity is rejected by the `method != null` term while the diagnostics-count conjunct guards specifically the argument-type case. That is two independent terms doing two distinct jobs, each pinned.

**Interpreter deaths in M1 and M2 mean the `gsi` column is genuinely pinned, not decorative.**

Not mutated, with reasons: `DiagnosticDescriptors.cs:78` (`"{0}"`) and the 1-arg reporter wrap are **coupled by construction** — mutating either garbles every `GS0159` and kills everything, a weak signal. The `effectiveReceiverSyntax != null` conjunct is a defensive null-guard no prior reviewer could reach and I could not either; **unreachable, so unpinnable — not a coverage gap.**

## Rollback correctness — the speculative bind does not eat real diagnostics

`Diagnostics.TruncateTo(diagnosticMark)` in a `finally` is the risky part of a speculative re-bind. Probed directly:

| shape | result |
|---|---|
| nullable error **then** `UnknownFn()` | both reported (`GS0159` + `GS0130`) |
| `UnknownFn()` **then** nullable error | both reported, order preserved |
| **two** consecutive nullable-receiver errors | **both** reported |

No diagnostic loss in any direction.

## False-positive corpus — clean, and text-compared

**122/122 `samples/*.gs` on the bare-`gsc` evaluate column** (the column the samples gate never traverses — `SampleConformanceTests.cs` is emit-only): rc and diagnostic-ID set identical base vs merged, `DIFF_RC=0`. Positive-controlled — **13 rc≠0, 26 emit diagnostics, 3 emit `GS0159`**, so the corpus is not vacuous.

An ID-set diff is **blind to text**, which is exactly what this PR changes, so I also compared rendered text on the 3 `GS0159` samples: **31 messages, byte-identical.** Unrelated `GS0159` wording is preserved on real code, not just in unit tests.

## Suites — 6 of 7 projects, no filter

| project | result |
|---|---|
| Core.Tests | **7089 passed, 1 skipped** |
| Interpreter.Tests | **798 passed** |
| Compiler.Tests | **4206 passed, 0 failed** (1.13 h, unfiltered) |
| LanguageServer.Tests | **462 passed** |
| Extensions.Tests | **104 passed** |
| InternalAnalyzers.Tests | **7 passed** |
| Sdk.Tests | **NOT RUN** — requires the local SDK package, out of bounds |

Targeted: `Issue3034` **33/33** Compiler + **6/6** Interpreter; `DiagnosticIdUniqueness` **4/4**.

## Docs and IDs — clean

Both catalogues updated at `:124` and **byte-identical to each other** (hand-diffed — the gate compares only ID and severity, never message text). Severity unchanged, row counts unchanged. **No new diagnostic ID**: the only descriptor touched is `GS0159` itself. **No self-allocation.**

---

## NITs

1. **`?.` alone is insufficient in a value context.** `var n int32 = c?.Get()` → `GS0156: Cannot convert 'int32?' to 'int32'`; `if let` → rc=0. The message is a **disjunction** and one arm always works, and unlike the earlier blockers this advances the user to a *correct, self-explanatory* error rather than looping back to the same one — so I grade it a NIT, not a defect. Worth knowing the two remedies are not interchangeable.
2. **`IsApplicableNullableUnderlyingCall` is a hand-maintained ~188-line mirror** of the dispatch chain that lives ~700 lines above it in the same method. Drift is **fail-closed** (a missed case silently reverts to the legacy wording), but silent. **Ponytail alternative worth considering:** the new branch sits *inside* `BindAccessorCall` (`:2015`), so re-entering it recursively with the narrowed receiver — under the same `TruncateTo` rollback — would express the same predicate ("would this call bind if the receiver were non-nullable?") in a handful of lines and could never drift. It terminates by construction, since the narrowed receiver is non-nullable and fails the `receiver?.Type is NullableTypeSymbol` guard on re-entry. Not blocking; the current code is correct and measured.
3. **The mirror collapses the real chain's two type-parameter branches** and drops their `InterfaceConstraint`/class-constraint guards. Inert today (an unconstrained parameter collects no methods), but it is one more way the copy can diverge from the original.
4. **Stale test name** — `ExpiredBranchNarrowing_ExplainsNeedToRenarrow` still says "Renarrow" though the message no longer mentions re-narrowing.
5. **`DiagnosticDescriptors.cs:78` is a bare `"{0}"` passthrough.** Verified safe today by census: **6 call sites, all routed through the two helpers, no direct-`Report` bypass anywhere in `src/`.** The cost is that the descriptor no longer documents its own diagnostic, and a future site passing a bare identifier would emit that token as the whole message with no compile error.

## Disclosed gaps and my own errors

- **My samples-sweep harness had a directory-collision bug**: both trees are named `w` (`/tmp/d2963/w`, `/tmp/d2963b/w`), so `basename` produced the same per-sample directory and the head run overwrote the base run's raw output. The rc/ID comparison is **unaffected** — it was written to two separate result files, each probe into a freshly `rm -rf`'d empty directory — but the base's message text was gone, so I **re-ran the 3 `GS0159` samples into distinct `base_*`/`head_*` directories** for the text diff. Reporting it because a stale directory is exactly how a probe silently reads the wrong build.
- **`gsi` REPL statefulness: NOT MEASURED.** A speculative bind that rolls back diagnostics could in principle leave other binder state dirty for a later submission. `gsi` refuses piped input (`gsi requires an interactive terminal`), so I could not drive it non-interactively. I probed the in-compilation equivalent instead (the rollback table above), which is clean. Declaring it rather than implying coverage.
- **Sdk.Tests NOT RUN** (local e2e out of bounds).
- The PR body's suite counts (4,190 / 729 / 7,089) differ from mine (7,090 Core total, 798 Interpreter) purely because `main` moved and added tests — **count drift, not falsity.** Its receiver-kind matrix and its `global` correction I re-derived independently and both are **true**.

---

## Verdict: **MERGEABLE**

Both prior blockers are genuinely fixed at this head and re-probed behaviourally rather than assumed. The fix for the last one was a **deletion**, which is the smallest thing that could have worked. Every clause the message can emit is true for the case that triggers it, across six receiver kinds and all three drivers, and the guard fails closed on accessibility failures rather than misattributing them. Five per-hunk product mutants all kill with clean isolation and the round-trip closes. 122 samples and 12,666 tests across six projects show no regression. The one SHOULD-FIX — remedies asserted as text but never compiled — does not block, because I measured the remedies and they hold; it is a durability gap, not a correctness one.

DavidObando:
## Addendum — my suite enumeration was wrong; the two missing projects are now run, both green

Correcting my own review above. I enumerated test projects with `ls test/`, got **7**, and reported "6 of 7 projects". **That denominator was wrong**, and it was wrong by exactly the failure mode I keep flagging in other people's work: **my enumeration was scoped to one directory, so a zero was scope, not absence.**

`GSharp.sln` actually contains **nine** test projects. Two live outside `test/`:

- `tools/cs2gs/Cs2Gs.Tests/Cs2Gs.Tests.csproj` (assembly `GSharp.Cs2Gs.Tests`)
- `tools/gsgen/GSharp.GeneratorHost.Tests/GSharp.GeneratorHost.Tests.csproj`

Both were built by my solution build and both sat in my own build output — `out/bin/Release/Cs2Gs.Tests/` even appeared in the mtime listing I published — and I walked past them.

`Cs2Gs.Tests` matters here specifically: it drives the real `gsc` out-of-process, so a pure `src/Core` change like this one **can** redden it without any test under `test/` noticing.

### Both now run on the same merged tree `6f7e2c5d`

| project | result |
|---|---|
| `Cs2Gs.Tests` | **1894 passed, 0 failed** |

Freshness verified before believing the green: every `GSharp.Core.dll` copy the tests load — including `out/bin/Release/Cs2Gs.Tests/GSharp.Core.dll` and `out/bin/Release/Compiler/GSharp.Core.dll` — is at **01:09:33, md5 `07a8d318`**, the clean restored merged build. `gsc.dll` and `cs2gs.dll` carry earlier mtimes because this PR touches only `Core` and the driver shims were unchanged; they load the fresh `Core`.

### Corrected totals

**8 of 9 projects run, 14,591 tests, 0 failures.** Still **NOT RUN: `Sdk.Tests`** (requires the local SDK package; out of bounds).

### External corpus pin

`tools/cs2gs/external/oahu.json` pins commit **`0ac1fece4415d31955ae7a1dcf7da31e343d363d`** and is **not touched** by this PR, so the external Oahu corpus is unchanged — stated from the pin rather than inferred from an in-repo file list, which could never have shown it.

**Verdict unchanged: MERGEABLE.** No finding moves; the two additional suites are clean. Recording the enumeration error because a wrong denominator in a posted verdict is worth more as a correction than as a footnote.

DavidObando:
## Addendum — correcting my own coverage denominator

My verdict above claimed "full suites, not a filtered green" and cited **three** projects (`Core.Tests`, `Compiler.Tests`, `Interpreter.Tests`) plus a population of "**33 files** assert the old `Cannot find function` wording (14 Compiler, 18 Core, 1 Interpreter)".

**Both numbers were wrong, and for the same reason: I enumerated from `test/`, so my zero was scope, not absence.** The real denominator is `GSharp.sln`, which holds **nine** test projects — including two outside `test/`:

- `tools/cs2gs/Cs2Gs.Tests` — invokes real `gsc.dll` **out-of-process**

Re-grepping wide across the whole repo, the population is **35 test files**, not 33 — and the two I missed include **`tools/cs2gs/Cs2Gs.Tests/Issue2282AnonymousTypeCrossBoundaryTests.cs`**, which references the exact `GS0159` wording this PR reworded, in the project that drives the compiler out-of-process. That is the worst possible place for me to have had a blind spot on this particular change.

### Re-measured, 8 of 9 projects, on current `main` `5b71a74d` (which contains this PR as `787166a9`)

| project | result |
|---|---|
| `Core.Tests` | 7106 passed, 1 skipped |
| `Compiler.Tests` | 4221 passed |
| `Interpreter.Tests` | 1201 passed |
| `Cs2Gs.Tests` | **1903 passed** |
| `LanguageServer.Tests` | 462 passed |
| `Extensions.Tests` | 104 passed |
| `GSharp.GeneratorHost.Tests` | **31 passed** |
| `InternalAnalyzers.Tests` | 7 passed |
| `Sdk.Tests` | **NOT RUN** |

**15,035 passed, 1 skipped, 0 failures.** `BUILD_RC=0`, 0 warnings; drivers fresh on `Core/`, `Compiler/` and `Repl/` (epoch-sorted mtimes, `ref/` excluded, oldest first-party artifact also fresh).

`Sdk.Tests` is **not run, not passing** — `BuildTaskSilentFailureTests.cs:22` drives `dotnet build` on sample projects against the local SDK package, which is out of bounds for this fleet. Stating that plainly rather than letting it hide inside a total.

### Did the missed project actually matter?

No, but only measurement establishes that. `Cs2Gs.Tests` is fully green, and I asserted the at-risk test **executed** rather than inferring it from a green total — the `Issue2282` filter returns `1 passed`, not `No test matches`. Inspecting it, its `Cannot find function` occurrence is a **doc comment at line 19**, not an assertion, so it was never truly exposed. Exposure was lower than I feared; my coverage claim was still wrong.

**The verdict is unchanged: MERGEABLE, and the merge stands.** What changes is the strength of the evidence behind it — "3 of 9, denominator taken from `ls test/`" is a materially weaker claim than "8 of 9, denominator taken from `GSharp.sln`", and I should not have published the first while implying the second.

PR #3055, opened 2026-08-01
Report interpreter ByRefLike reflection boundary

### LINKED ISSUE #2987: gsi: Span/ReadOnlySpan calls fail with 'Specified method is not supported'

DavidObando:
## Summary

- add `Evaluator.Reflection.cs` as one reflection chokepoint for all **18** user-facing evaluator call sites (15 in `Evaluator.Expressions.cs`, 3 in `Evaluator.Calls.cs`)
- reject resolved CLR signatures containing stack-only (`ByRefLike`) types before reflection with named `GS0516`, directing users to `gsc`
- preserve non-ByRefLike reflection behavior, including CLR constructors, operators, indexers, properties, and interpolated-string handlers
- document `GS0516` in both gated diagnostic catalogs and mark Span samples compiled-only under `gsi`
- pin return-type, parameter-type, and declaring-type detection independently

Fixes #2987

## Rebase blocker resolution

Rebased onto requested `main` `01c1a6c5c`.

`InvokeClrCtor` conflict was combined deliberately:

- kept `BuildRefSlots(arguments, argumentRefKinds, args, ctor)` from `main`, preserving #2990 delegate coercion
- kept `InvokeReflective(ctor, null, args, node)` from this PR, preserving the ByRefLike guard

`Issue2990ClrDelegateBoundaryTests` now passes the bound argument as the fourth `BoundNode` parameter when reflectively invoking `InvokeClrCtor`. Merged-tree result: **7/7 passed**.

## Guarded denominator: 18

- `Evaluator.Calls.cs`: 3 sites — imported static/extension, map-locked instance, normal instance
- `Evaluator.Expressions.cs`: 15 sites — handler ctor/literal/formatted, CLR backing/default/base/normal ctors, CLR static, property/field get/set, binary/unary/conversion, indexer get/set

Current lines: Calls `912/1026/1032`; Expressions `203/225/321/1210/1227/1277/1287/1338/1342/1470/1607/1613/1619/1631/1645`.

Previously unreached sites now have coverage:

- handler ctor/literal/formatted: real `[InterpolatedStringHandler]` fixture produces `value is 42 and 43`
- CLR base ctor helper: merged `Issue2990` reflection test
- binary/unary operators: distinct `33` / `-11` interpreter assertions

## Mutation evidence

Every mutant used a full-solution Release build with `ContinuousIntegrationBuild=true`; `out/bin/Release/Core/GSharp.Core.dll` mtime was newer than build start. Clean cycle MD5: `0fc3f9037dea3ace2849997d5f2aa45f`; every restore matched exactly.

| Mutant | Fresh Core MD5 | Result |
|---|---|---|
| M-ALL: disable throw | `9a0bfd1e59f3a251730abf44775a2389` | KILLED: both return-signature GS0516 cases fail |
| M-DECL: remove `DeclaringType` arm | `ba15841df74d8519857e182d98076ea2` | KILLED: `ByRefLikeDeclaringType_IsDetected` fails |
| M-PARAM: invert parameter scan | `7f0d33683ded8189a18c678e85443b7a` | KILLED: `ByRefLikeParameter_IsDetected` fails |

Final committed full-solution build: rc=0, 0 warnings/errors, fresh Core MD5 `835ad3effe061fe8261bc9a013914487`.

## Validation

- full serialized solution suite: **14,209 passed / 0 failed / 1 skipped**
- `Issue2990ClrDelegateBoundaryTests`: **7/7 passed**
- cs2gs corpus gate: rc=0; `G11-Linq-Console` PASS; **0 new / 0 regressed / 0 stale** gaps
- latest fetched `main` `f1656079a`: computed merge base is `f1656079a`; `git merge-tree` rc=0

`Evaluator.Concurrency.cs` remains untouched: CLR constraints prevent `Channel<Span<T>>`, so the guard would be dead there.

## Top-level Span temporary pin

Added a genuinely top-level case (no `func Main()` wrapper):

```gs
import System
MemoryExtensions.AsSpan([]int32{11, 22, 33})
```

Measured after rebasing onto fetched `main` `f1656079a2a109aade66bb5ee4ec1d3833246414`:

| Scope | bare `gsc` (evaluate) | `gsc /out:` (emit + run) | `gsi` (interpret) |
|---|---|---|---|
| top-level temporary | rc1 `GS0516` | rc0, `33` | rc1 `GS0516` |
| in-function temporary, invoked top-level | rc1 `GS0516` | rc0, `22` | rc1 `GS0516` |

All three driver columns covered. No owned test/sample source wraps cases in `func Main()` only.

Product mutation inverted `if (type != null)` to `if (type == null)`. Full-solution build succeeded with fresh Core/Compiler artifacts. New top-level case went RED: expected `GS0516`, got `GS9999`. Restored guard, rebuilt, exact clean MD5 round-trip `8c6b8b590a1d039a0ca95a1fe8f1b32e`, then suite went GREEN **7/7**.

Fetched `origin/main` still equals `f1656079a`; computed merge base equals `origin/main`, so HEAD is the merged content. `git merge-tree` rc=0. Combined merged-tree target: `Issue2987ByRefLikeBoundaryInterpreterTests` + `Issue2990ClrDelegateBoundaryTests` = **14/14 passed**. Final full-solution rebuild: rc=0, 0 warnings/errors; Core and Compiler artifacts fresh and byte-equal.

Local routed validation was targeted; full local solution tests, cs2gs, and e2e were not rerun. Pushed-head CI then completed green: **30 successful / 0 failed / 3 intentionally skipped**, including all test shards, build, docs, e2e, cs2gs corpus, and cs2gs Oahu.


DavidObando:
## Review: NOT MERGEABLE — merges red into `main`, and 2 of 3 detection arms are pinned by nothing

Base for all measurements: `main` = `46d3cc50f`, head = `21470efee`, merge base **computed** = `fdba684b6` (PR 0 behind / 1 ahead; `main` **8 commits** past merge base). Three-dot diffstat: 11 files, **+293/−49** — matches the PR body.

Good news first, all measured: detection uses the real `Type.IsByRefLike` (`Symbols/ClrTypeUtilities.cs:395`), **not** a name match, not `IsByRef`/`IsPointer`. Severity is `Error` and the exit code really is `1`. `GS0516` is used for this condition only. Both `docs/diagnostics.md` and `website/docs/ref/diagnostics.md` carry **bare** `| GS0516 | Error |` rows — no backtick trap — and this branch's copy of the gate has the same 1 `website` mention as `main`, so it is **not** a pre-#3043 blind gate. On the samples the PR replaces `GS9999: Specified method is not supported.` with a located `GS0516`, so the internal-error bucket is genuinely drained for that shape.

---

### BLOCKER 1 — the PR reds `main` on merge, and neither CI run can see it

`src/Core/CodeAnalysis/Evaluator.Expressions.cs:1180` changes a **private** method's arity:

| tree | signature |
|---|---|
| `main` `46d3cc50f` | `InvokeClrCtor(ConstructorInfo, ImmutableArray<BoundExpression>, ImmutableArray<RefKind>)` — 3 params |
| PR `21470efee` | `InvokeClrCtor(ConstructorInfo, ImmutableArray<BoundExpression>, ImmutableArray<RefKind>, BoundNode node)` — **4 params** |

`main`'s `test/Interpreter.Tests/Issue2990ClrDelegateBoundaryTests.cs:92` invokes that private method **reflectively** with a 3-element array:

```csharp
var invoke = typeof(Evaluator).GetMethod("InvokeClrCtor", BindingFlags.Instance | BindingFlags.NonPublic);
invoke.Invoke(evaluator, [constructor, ImmutableArray.Create<BoundExpression>(...), ImmutableArray<RefKind>.Empty]);
```

Four-cell, all measured:

| tree | result |
|---|---|
| `main` `46d3cc50f` | `Issue2990ClrDelegateBoundaryTests` **7/7 PASS** |
| PR head `21470efee` | test file **absent** (blob absent at head *and* at merge base) |
| merged (`main` + PR) | **591/592**, `InvokeClrCtor_CoercesClosureWithUserTypeParameter` **FAILS** — `TargetParameterCountException: Parameter count mismatch` |

`Issue2990ClrDelegateBoundaryTests.cs` landed on `main` in the 8 commits after the merge base, so the PR branch has **never run it**. Branch CI green, `main` CI green, merged tree **red**. The PR contains **0** references to `Issue2990`.

This is not my merge resolution: the 4-arg form is already at the PR head. Fix on rebase — either keep the 3-arg overload or update that test's arg array. It must go in this PR; nothing else will catch it.

### BLOCKER 2 — `merge-tree` conflicts, and the obvious resolution silently un-guards a site

`git merge-tree` into `46d3cc50f` returns **rc=1**, one conflict in `Evaluator.Expressions.cs`:

- PR side: `BuildRefSlots(arguments, argumentRefKinds, args)` + `InvokeReflective(ctor, null, args, node)`
- `main` side: `BuildRefSlots(arguments, argumentRefKinds, args, ctor)` + `ctor.Invoke(args)`

`main` added a `ctor` parameter to `BuildRefSlots` at a location this PR converts into a guarded site. Taking `main`'s side **compiles fine, passes CI, and removes the guard** — precisely the defect this PR exists to prevent. Taking the PR's side fails to compile (loud, harmless). I resolved by combining both; build rc=0. Please rebase and resolve deliberately, not by hunk selection.

### BLOCKER 3 — 2 of the 3 detection arms are pinned by nothing

Product mutants only, each built green (rc=0), each restored with an **exact** `GSharp.Core.dll` md5 round-trip to `47229604ab025e461391319ca271922f`:

| mutant | `Evaluator.Reflection.cs` | result |
|---|---|---|
| **M-ALL** — whole guard neutered (`if (type != null && node == null)`) `:57` | never throws | **KILLED 2/3** — both `ByRefLikeReflectionSignature_ReportsGs0516` cases fail |
| **M-DECL** — `DeclaringType` arm removed `:68` | **SURVIVED 3/3** |
| **M-PARAM** — parameter-scan arm inverted `:100` | **SURVIVED 3/3** |

M-ALL is the positive control, so the two survivals are real, not a blind suite: **both test cases are caught solely by the return-type arm.** The `DeclaringType` and parameter arms — which are what catch a `Span` *argument* or a method *on* a ref struct — are unpinned. Two theory cases that differ only in `Span` vs `ReadOnlySpan` exercise one arm three times. Please add a case where the ByRefLike appears **only** as a parameter, and one where it appears **only** via `DeclaringType`.

---

### SHOULD-FIX — the "15-site chokepoint" is **18 sites**

I did not grep for 15; I made the compiler produce the denominator by adding a required `int siteId` to all three helpers: build rc=1, **72 × CS7036**, deduped by `file:line` = **18 distinct call sites** (13 `InvokeReflective`, 3 `GetReflective`, 2 `SetReflective`). Textual cross-check on the unmerged head: 21 − 3 definitions = 18. Agreement exact.

**15 is the `Evaluator.Expressions.cs` subtotal.** The 3 sites in `Evaluator.Calls.cs` (`:912`, `:1026`, `:1032`) were omitted — and they are not spare parts. Runtime site logging shows `Evaluator.Calls.cs:912` is **the site that actually fires** for the shipped sample `samples/RefStructSpan.gs`. The undercount hides a load-bearing site.

### SHOULD-FIX — 6 of 18 sites are never reached by `Interpreter.Tests`

Runtime log across the whole suite: **12 of 18** guarded sites reached. Never reached: `Evaluator.Expressions.cs:203`, `:225`, `:321`, `:1227`, `:1607`, `:1613`. Caveat, disclosed: that run was on my instrumented build, where the instrument itself broke 6 console-capture tests, so this is a **lower bound**.

### NIT — no source break, verified

All 3 G#-declared `ref struct` samples (`UserRefStruct.gs`, `RefStructGenericField.gs`, `refactoring-baseline/RefStructByRefLike.gs`) run **rc=0 with byte-identical output on `main` and the PR**. 2 of the 3 print real values (`12`, `7`); `RefStructGenericField.gs` prints nothing on **both** sides, so only 2 are load-bearing probes. Site logging confirms the mechanism, not just the output: `UserRefStruct.gs` **hits** the guard 3 times and correctly declines to fire. Ref structs used entirely within interpreted code are unaffected.

---

### Measurement errors I made, disclosed

1. My per-site sentinel wrote to `Console.Error`, which broke 6 console-capture tests. Instrument artifact; re-measured with logging off.
2. I then assumed the **remaining** failure was also my instrument. **That was wrong** — it persisted on the clean, unmutated tree, and chasing it is what found BLOCKER 1. The lesson cuts the other way for once: I nearly dismissed the most important finding as my own noise.

`git status --porcelain` empty in every tree; `gsharp-wt/i2987` never modified, still `21470efee`; nothing pushed; no e2e runs.

DavidObando:
## Addendum — the `func Main()` axis. Verdict unchanged (NOT MERGEABLE), but BLOCKER 3 is now sharper

Re-measured after the fleet bulletin on test-shape blindness. Two findings, one exonerating and one not.

**Both of this PR's test cases put the ByRefLike construct inside `func value(...)`** (`Issue2987ByRefLikeBoundaryInterpreterTests.cs:22` and `:35`); only the *call* `value([]int32{11, 22, 33})` is top-level. **All four samples I used as probes have the same shape** — `SpanIndexing.gs:7` states the reason outright: *"Spans are stack-only (GS0219), so the span locals live inside functions."* So I had inherited the author's shape and had not tested file scope.

**Exonerating half — the declaration axis is legitimately unreachable.** A top-level span *declaration* is rejected by an earlier binder diagnostic, identically on both sides:

| probe (top-level `var span …`) | `main` `46d3cc50f` | PR `21470efee` |
|---|---|---|
| `var span = MemoryExtensions.AsSpan(values)` | rc=1 `GS0219` | rc=1 `GS0219` **identical** |
| `var span ReadOnlySpan[int32] = values` | rc=1 `GS0219` | rc=1 `GS0219` **identical** |

No source break, and the in-function test shape is justified *for that construct*.

**Not-exonerating half — the boundary is reachable at top level without any declaration.** `GS0219` guards declarations, not the reflection boundary. A span held as a **temporary** crosses it at file scope with no `func` anywhere:

| probe | `main` | PR |
|---|---|---|
| `Console.WriteLine(MemoryExtensions.AsSpan("hello").Length)` | rc=1 **`GS9999`** | rc=1 **`GS0516`** located |
| `Console.WriteLine(text.AsSpan().Length)` | rc=1 **`GS9999`** | rc=1 **`GS0516`** located |

**The PR handles this correctly — and pins it with nothing.** `gsi` only ever executes top-level code, so this is the interpreter's primary scope, and the interpreter is this PR's entire subject. The fix for BLOCKER 3 is now cheaper than I first described: a top-level case needs **no `func` wrapper and is three lines**.

Caveat, stated rather than glossed: load is 35.73, so I ran **no builds** and did not re-instrument. I therefore **cannot say which detection arm** these top-level probes exercise. By construction `AsSpan` returns `Span<T>`, so it is most likely the **return-type** arm — the one arm already pinned. These probes add *scope* coverage, **not** arm coverage, so they do **not** substitute for the `DeclaringType` and parameter-arm cases M-DECL/M-PARAM proved unpinned. Both gaps need closing.

All measurements against `main` `46d3cc50f`; `main` is now `185431c4`, so BLOCKER 1 should be re-confirmed before merge — it depends only on `Issue2990ClrDelegateBoundaryTests.cs` still existing on `main`. Trees clean; `i2987` untouched at `21470efee`; nothing pushed.

DavidObando:
Revalidated after rebasing onto `main` `5ad0109c04608f627e35d797d1735ca21d8311f2` (head `e134f3d8b664997e67c03d6817843a8f7b2cde8c`).

#3020 conflicted at five reflection sites. Resolution preserves both changes: first `ResolveMemberForRuntimeType(...)`, then `InvokeReflective`/`GetReflective`/`SetReflective`. #3020 tests pass 12/12.

Top-level/invoked-function matrix:

| Scope | bare `gsc` | `gsc /out:` + run | `gsi` |
|---|---|---|---|
| top-level temporary | rc1 `GS0516` | rc0 `33` | rc1 `GS0516` |
| function temporary, invoked top-level | rc1 `GS0516` | rc0 `22` | rc1 `GS0516` |

Product mutation inverted `if (type != null)` to `if (type == null)`. Full-solution mutant build succeeded; fresh Core/Compiler/Repl hash changed `1bd8c425676e3d548088f5b94fb71785` → `8e98062118871a995347c959c8e72e2a`. Boundary theory went RED 3/3, including new top-level row: expected `GS0516`, actual `GS9999`. Restored guard; full rebuild returned exact clean hash; theory GREEN 3/3.

Merged-content validation: solution rebuild 0 warnings/errors; Issue2987 + Issue2990 14/14; Issue2992 12/12; `git merge-tree` rc0 with merge base equal to `5ad0109c`. PR checks: 30 success, 0 failure, 3 intentional skips. Full local suite, local e2e, and local cs2gs were not rerun.

DavidObando:
Rebased onto `main` `33bea4afeef197c89da73f9edf0b20c36180f80f`; pushed head `3d2939b3f1a0e0d80a1da3229ba2678cd62a4245`.

Conflicts were limited to diagnostic catalog adjacency with #3024. Kept both GS0514 and GS0516. Branch does not touch `src/Repl/Program.cs`; main's `engine.Evaluate(..., arg)` renderer remains intact and manual `gsi` probes now include file paths.

Blocker validation on merged content:
- real top-level `MemoryExtensions.AsSpan([]int32{11, 22, 33})` theory row exists
- `Issue2987ByRefLikeBoundaryInterpreterTests` + `Issue2990ClrDelegateBoundaryTests`: 14/14 passed; reflective `InvokeClrCtor` uses four arguments
- product condition-inversion mutant: top-level row RED (`GS0516` expected, `GS9999` actual); restore GREEN 3/3
- clean full-build Core/Compiler/Repl hash `5609ea9e78ded2191db45e5c8a42f2e9`; mutant `e7442170810f1d85a5358a464dad6226`; restored exact clean hash
- top-level matrix: bare `gsc` rc1 GS0516; emit/run rc0 `33`; `gsi` rc1 GS0516
- invoked-function matrix: bare `gsc` rc1 GS0516; emit/run rc0 `22`; `gsi` rc1 GS0516
- `git merge-tree` rc0; merge base equals `33bea4af`
- PR status: MERGEABLE; 30 successful checks, 0 failed, 3 intentional skips

DavidObando:
## Verdict: NOT MERGEABLE — 1 BLOCKER (diagnostic ID), 2 SHOULD-FIX, 2 NIT

**Trees.** `main` = **`4f7559dd`**; head = **`3d2939b3f`**; computed merge base = **`33bea4af`** (PR **4 behind / 3 ahead**). Three-dot scope: 12 files, **+388 / −50**. Merged tree built: `git merge-tree` predicted `acedf4d0bd715f86091eb513fbe302f6f50ce245` and the real merge **reproduced that hash exactly**; `BUILD_RC=0`, **0 errors / 0 warnings**; `Interpreter.Tests` **0 failed / 628 passed / 628**. All three `GSharp.Core.dll` copies (`Core/`, `Compiler/`, `Repl/`) verified mtime-identical before every probe.

Note the PR body's anchors are stale: it cites rebasing onto `01c1a6c5c` and a merge base of `f1656079a`. Neither is current. The numbers below are true of `4f7559dd` only.

### The reported shape reproduces, and the fix addresses it

Issue #2987 names two samples specifically. I ran exactly those, not neighbours:

```
samples/RefStructSpan.gs   gsi rc=1  GS0516 ... 'System.ReadOnlySpan[char]'
samples/SpanIndexing.gs    gsi rc=1  GS0516 ... 'System.ReadOnlySpan[int32]'
```

Both previously leaked `GS9999: Specified method is not supported.` The boundary is real and it is `Evaluator`-scoped, so it correctly appears in the two evaluating drivers (bare `gsc`, `gsi`) and not in `gsc /out:`.

### The RED-merged trap did not recur — checked directly

This PR previously merged red because it changed a private method's arity while a test invoked it reflectively. It still changes **two** arities — `InvokeClrCtor` 3→4 (`Evaluator.Expressions.cs:1230`) and `AllocateClrBacking` 3→4 (`:1184`). Enumerated on current `main`: the only reflective invocation of either, repo-wide, is `test/Interpreter.Tests/Issue2990ClrDelegateBoundaryTests.cs:88`, and this PR updates it. `AllocateClrBacking` is never invoked reflectively. Positive control: 4 `typeof(Evaluator)` reflective sites exist on `main`; all were enumerated, not sampled. The merged tree then built and tested green, which is the actual proof.

---

## BLOCKER 1 — `DiagnosticDescriptors.cs:383` self-allocates `GS0516`; the allocation of record for #2987 is `GS0511`

Issue #2987's allocation ledger is explicit and was restated twice:

> comment 4 of 6 — *"G1 — #2987 `Span`/`ReadOnlySpan` interpreter boundary error → **`GS0511`**"*
> comment 5 of 6 — *"**Allocation of record, unchanged: G5 error `GS0509`, G5 warning `GS0510`, G1 `GS0511`, G3 reserved `GS0512`.**"*

The PR ships `GS0516`. Run from the main checkout — the location clause required for `--all` to see every branch:

| query | commits | where |
|---|---|---|
| `git log -S"GS0516" --all` | **2** | `refs/heads/fix/2987-gsi-byreflike-boundary`, `refs/duck/pr3055` — **both this branch's own** |
| `git log -S"GS0511" --all` | **0** | nothing in all history — **still free** |
| `git log -S"GS0514" --all` (control) | 4 | #3024, merged |
| `git log -S"GS0509" --all` (control) | 1 | an **abandoned earlier head of #3024** |

`GS0516` also appears **zero** times in `main`'s history, and the GS05xx descriptor block on `4f7559dd` is `0500–0506, 0508, 0513, 0514` — so `GS0511` was free the entire time.

The `GS0509` row is the informative one: #3024 originally took the ledger's `GS0509`, and was moved to `GS0514` when that ID was issued to it. Precedent says these get re-issued and that the author follows an allocation once given.

**Two things I want to be straight about, because they cut the other way.** The standing "taken/unsafe" list is the *range* `GS0509–GS0517`, which does not say which ID belongs to which workstream; and the previous review of this PR examined `GS0516` and cleared it ("`GS0516` is used for this condition only"), without checking whether it had been allocated. So it is possible `GS0516` was issued out of band and simply is not written down in either thread. I cannot verify that, and allocation is not mine to decide — which is exactly why this is a blocker rather than a nit. **The number it took is `GS0516`; the last written allocation for #2987 is `GS0511`.** One ruling settles it, and if the answer is `GS0511` the change is mechanical: descriptor, two docs rows, three test assertions, two sample comments.

## SHOULD-FIX 1 — `Evaluator.Reflection.cs:29`: the `GetReflective` guard is pinned by no test, and deleting it restores the exact bug this PR exists to fix

The PR body says detection is pinned "independently" for return type, parameter type and declaring type. That is true of the three **arms** of `FindByRefLikeSignatureType`, which two unit tests call directly. But the three **wrappers** each carry their own `ThrowIfByRefLikeSignature` call, and only `InvokeReflective`'s is exercised end to end. `GetReflective` (`:29`) and `SetReflective` (`:39`) are not.

`GetReflective`'s guard is user-reachable — a static property whose type is `ByRefLike`:

```gs
func probe() int32 {
    var e = Span[int32].Empty     // reaches GetReflective, not InvokeReflective
    return e.Length
}
Console.WriteLine(probe())
```

Product mutant, guard deleted from `GetReflective` only (deletion, not `if (true)`); mutant asserted applied by guard-site count 4 → 3:

| step | BUILD_RC | probe result | `Interpreter.Tests` |
|---|---|---|---|
| clean | 0 | `GS0516 … 'System.Span[int32]'` | 0 failed / 628 |
| **mutant** | 0 | **`GS9999: Specified method is not supported.`** | **0 failed / 628 — SURVIVES** |
| restored | — | guard sites back to 4, `git status --porcelain` empty | — |

The mutant is not subtle: it regresses the diagnostic to the literal string this PR was filed to eliminate, on a shape a user can write, and **the whole suite is green either way**.

This is repo-wide, not just an Interpreter.Tests gap: `GS0516` does not exist on `main`, so the only tests that *could* pin it are the ones this PR adds, and all of them live in `Interpreter.Tests`.

**Fix:** two theory cases — one reading a `ByRefLike`-typed property, one assigning to one — which also closes `SetReflective`.

## SHOULD-FIX 2 — `GS0516` reports a fabricated source location: always line 1, no filename

Every occurrence points at line 1 of an unnamed file, on both scope axes:

| probe | offending code at | reported |
|---|---|---|
| `loc1.gs` — call inside a function | **line 7** | `(1,1,1,39)` |
| `loc2.gs` — call at top level | **line 5** | `(1,1,1,52)` |
| `samples/RefStructSpan.gs` | inside a function | `(1,1,1,37)` |
| `g2.gs` — property get | **line 4** | `(1,1,1,13)` |

The end column is always the length of the *pretty-printed bound node* plus one, which is the tell. **Positive controls from the same runs** show the infrastructure is fine and the fault is specific to this diagnostic:

```
/tmp/d3055/p/loc1.gs(11,1,11,26): warning GS0286: ...     <- correct file and line
/tmp/d3055/p/g1.gs(4,5,4,6): error GS0219: ...            <- correct file and line
(1,1,1,39): error GS0516: ...                             <- fabricated, no file
```

Mechanism, at `src/Core/CodeAnalysis/Compilation/Compilation.cs:344-347`: the `EvaluatorException` catch builds a **new `SourceText` from `ex.Node.WriteTo(...)`** and spans it, discarding the real location. `BoundNode.Syntax` exists (`Binding/BoundNode.cs:38`, "the originating `SyntaxNode`"), so a real location is recoverable.

**This code is pre-existing on `main` and this PR does not touch `Compilation.cs`** — I want that stated plainly. The reason it is nonetheless this PR's problem is that the design rationale for accepting it has expired. The G1 design comment accepted the missing location explicitly and conditionally:

> *"`src/Repl/Program.cs:60` prints `{d.Id}: {d.Message}` with no source location, which is exactly why replacing the raw message with a named diagnostic is the highest-value change available **at that ceiling**."*

**#3024 removed that ceiling.** `gsi` now renders locations, so what used to be invisible is now printed — and it is not merely absent, it is *plausible and wrong*, which is worse for a diagnostic whose entire job is to tell a user where to look before switching to `gsc`. The sibling boundary diagnostic gets this right because it reports at bind time through `DiagnosticBag`, which is why its test can assert `Location.FileName` and `StartLine`; this PR's tests assert only `Id` and message substrings, so the location is pinned by nothing and the defect shipped unnoticed. I did not size the fix with a build (machine load); the finding above is probe-measured, the fix estimate is not.

## NIT 1 — the message tells you to do what you may already be doing

Bare `gsc` evaluates, so `gsc file.gs` on a span program emits `GS0516: … compile this program with 'gsc' instead.` The wording is inherited verbatim from `GS0513` and `GS0514`, both already merged, so this PR did not introduce it — raising it once for the family rather than against this change. Scoping the clause to the emit path (`gsc /out:`) would make it correct in all three driver columns.

## NIT 2 — `EvaluatorException.CreateDiagnostic` now formats unconditionally

`EvaluatorException.cs:86` changed to `string.Format(descriptor.MessageFormat, messageArguments)` with `params object[]`. Existing zero-argument callers now run through `string.Format` too, so any descriptor whose message contains a literal `{` or `}` would throw `FormatException` instead of rendering. I checked the only such caller on this path — `InterpreterPointerOperationsNotSupported`, used twice in `Evaluator.Calls.cs` — and it has no braces, so this is latent, not live. Worth a guard (`messageArguments.Length == 0 ? descriptor.MessageFormat : string.Format(...)`) since the trap is silent.

---

## Checked and deliberately not filed

- **Commit subjects are true of their content** — no rebase dressed as work. `a41d4dce` 12 files +295/−50; `6727a252` +89/−4, tests only; `3d2939b3` +8, and it genuinely adds the top-level theory case its subject claims.
- **The uncalled `ReportInterpreterByRefLikeValuesNotSupported`** (`DiagnosticBag.Reports.Expressions.cs:989`) is the established convention here, not dead weight introduced by this PR: the already-merged `ReportInterpreterPointerOperationsNotSupported` is identically uncalled, because these descriptors are consumed through `EvaluatorException.CreateDiagnostic` while the `Report` method satisfies the descriptor-usage gate.
- **`FindByRefLikeType`'s `HasElementType` recursion is a strength, not redundancy.** A `ref Span<int>` parameter has `ParameterType` `Span<int>&`, whose `IsByRefLike` is **false**; a naive check would miss it and leak `GS9999`. The recursion catches it.
- **Test shape is not vacuous.** Theory case 1 is genuinely file-scope, cases 2–3 are in-function, so both axes are covered, and nothing is wrapped in `func Main()`.
- **No false positives found.** Beyond the PR's own two controls (interpolation handler; CLR operators asserting distinct `33` / `-11`), I ran a 6-program corpus on the merged head — seven string operations, interpolation, `StringBuilder` construction plus property reads, `TimeSpan` operators, `List`/`Dictionary` indexers, and `Parse`/`Convert`/`Math`. **All rc=0 with correct distinct values.** The guard mirrors the CLR's own restriction rather than exceeding it, so the over-fire surface is genuinely small.
- **`src/Repl/Program.cs` is untouched**, so the three-way contention affecting other open branches does not apply and `main`'s `engine.Evaluate(…, arg)` survives — confirmed by the file paths in my probe output.
- Docs rows are **bare** `| GS0516 | Error |` in both `docs/diagnostics.md:1172` and `website/docs/ref/diagnostics.md:1863`.

## Not run, and why

`Core.Tests`, `Compiler.Tests` (~1h20m even sharded) and `Cs2Gs.Tests` were not run on the merged tree; `e2e` is prohibited locally. I did not re-run the author's three mutants — mine is independent of them and targets a site they did not cover. Machine load ranged 20–41 throughout and I ran no build above 30, which is why the location fix is unsized. The source restore is proven exact by an empty `git status --porcelain`; I did not rebuild after it, so no conclusion in this report depends on the scratch tree's binaries after the mutation section.

DavidObando:
Addressed latest verdict. Branch rebased/pushed at `1f97f9e655e04cf07145b33778fa1e0639144c26`.

- Renamed boundary diagnostic everywhere from `GS0516` to allocated `GS0511` (descriptor, tests, both docs, sample notes).
- Added end-to-end wrapper pins:
  - `GetReflective`: read `Issue2987ByRefLikePropertyProbe.Value` (`Span<int>` property).
  - `SetReflective`: assign `33` to a normal static property declared on a `ref struct`.
- Per-wrapper product mutation, asserted by guard count 4→3:
  - delete Get guard: build rc0; fresh Core/Compiler/Repl hash `8985fd99abdd78ee15e24d7a5de5ab0`; theory RED 1/5, expected GS0511 got GS9999.
  - delete Set guard: build rc0; fresh hash `15269714d0c1319f9546af580a8089c8`; theory RED 1/5, expected GS0511 got GS9999.
  - An initial Span-valued setter probe was intercepted before `SetReflective` and survived; discarded and replaced with the discriminating ref-struct-declaring-type probe above.
- Fixed location here: evaluator diagnostics use originating syntax location when present, while retaining the prior reprinted-node fallback for synthesized nodes. Imported static/extension/instance calls now preserve syntax through binding/rewriting. Pinned `byreflike.gs(7,18,7,45)`; manual `gsi` confirms filename and real line instead of fabricated line 1.
- Wording now directs users to `gsc /out:<file>`. Zero-argument `CreateDiagnostic` skips `string.Format`.

`main` moved twice during validation. Latest fetched base is `4deeed8d45a5fd4f0a5f5a4814c1b6882ed6fbcb` (past requested `36639253`). On that merged tree: full solution `BUILD_RC=0`, 0 errors/warnings; Interpreter.Tests 643/643; `git merge-tree` rc0, merge base equals main. All three Core/Compiler/Repl artifacts had identical fresh mtimes/hashes. PR currently MERGEABLE; CI 25 success, 0 failed, 5 pending, 1 skipped.

DavidObando:
## Round-2 re-verification — `1f97f9e6` — **MERGEABLE** (0 BLOCKER, 2 SHOULD-FIX, 0 NIT)

Round 1's BLOCKER and both SHOULD-FIX items are addressed. The round-1 NITs are also gone. One finding remains: the location fix is **incomplete along a specific axis**, and that axis includes one of the two files named in issue #2987.

Measured against `main` = **`6ec381c5`** (not `abfab99c`; main moved by #3106 mid-review). Merge base `4deeed8d`, 6 behind / 4 ahead. Merge onto `6ec381c5` is clean and reproduced the predicted tree hash **`0cf79666`** exactly. `BUILD_RC=0`, **0 warnings / 0 errors**. `Interpreter.Tests` **647/647**, `Core.Tests` **7088 passed / 0 failed / 1 skipped / 7089**. Binary mtimes verified fresh in `Core/`, `Compiler/` and `Repl/` (excluding `ref/`) before every probe.

---

### Claim 1 — rename `GS0516` → `GS0511`: VERIFIED

`git grep GS0516` on the head returns nothing (rc=1); `GS0511` is present in 6 files. `git log -S GS0516 --all` (run from the primary checkout, which owns the shared object store) returns 3 commits, **all this branch's own**; `git log -S GS0511 --all` returns 1, also this branch's — the ID was free.

Docs rows hand-diffed rather than trusted to the gate (the gate checks row *presence* only). `docs/diagnostics.md:1168,1172` and `website/docs/ref/diagnostics.md:1859,1863` carry **byte-identical** heading and row text.

### Claim 2 — Get/Set wrapper pins: VERIFIED, per hunk, independently

Two separate product mutants, each asserted applied before building.

| mutant | site | assertion | result |
|---|---|---|---|
| A — delete guard from `GetReflective` | `Evaluator.Reflection.cs:29` | guard-site count 3 → 2 | `BUILD_RC=0`; **DEAD** — 1 failed / 646 passed / 647 |
| B — delete guard from `SetReflective` | `Evaluator.Reflection.cs:40` | guard-site count 3 → 2 | `BUILD_RC=0`; **DEAD** — 1 failed / 646 passed / 647 |

Mutant A kills theory case 2 (property **get**); mutant B kills theory case 3 (property **set**). Distinct failures, so the two pins are genuinely independent and neither is riding on the other. Both survived in round 1. Source restored exactly (`git status --porcelain` empty) after each.

### Claim 4 — 628 → 643: VERIFIED as real tests, not scope

`git diff 6ec381c5 HEAD -- test/Interpreter.Tests/Interpreter.Tests.csproj` is **empty** — no project-scope change. The only test files touched are `Issue2987ByRefLikeBoundaryInterpreterTests.cs` (added) and `Issue2990ClrDelegateBoundaryTests.cs` (modified). I measure 647 rather than 643 because I merged onto `6ec381c5` and you measured at `4deeed8d`; the +4 is main's.

### Claim 3 — locations: **PARTIALLY** verified. This is the finding.

Full sweep, base `6ec381c5` vs merged head, same harness, bare `gsc` (evaluate column):

|---|---|---|---|---|
| `loc1.gs` | imported **static call**, in-function L7 | `(1,1,1,39)` GS9999 | `loc1.gs(7,33,7,47)` GS0511 | fixed |
| `loc2.gs` | imported **static call**, top-level L5 | `(1,1,1,52)` | `loc2.gs(5,18,5,45)` | fixed |
| `stat1.gs` | imported **static call**, GS9999 L5 | `(1,1,1,35)` | `stat1.gs(5,25,5,46)` | fixed (collateral) |
| **`g2.gs`** | **static property get, L4** | `(1,1,1,13)` | **`(1,1,1,13)`** | **unchanged** |
| **`inst1.gs`** | **imported instance call, L7** | `(1,1,1,14)` | **`(1,1,1,14)`** | **unchanged** |
| `p513.gs` | GS0513 pointer op | `(1,1,1,3)` | `(1,1,1,3)` | unchanged (pre-existing) |
| `p514.gs` | GS0514 P/Invoke | `p514.gs(5,6,5,18)` | identical | unchanged — never used this path |
| `g1.gs` | GS0219 **positive control** | `g1.gs(4,5,4,6)` | identical | control holds |
| `loc1.gs` | GS0286 **positive control** | `loc1.gs(11,1,11,26)` | identical | control holds |

**To answer the sweep question directly: no diagnostic got worse, and every ID that changed changed to something strictly more correct — none is merely different.** Both positive controls are byte-identical, so no correct diagnostic shifted by a column. GS0513 and GS0514 are unaffected: GS0514 already had a real location via the declaration-discovery path, and GS0513 keeps its pre-existing fabricated one.

**SHOULD-FIX 1 — the fix covers imported *call* expressions only; property/field access and imported *instance* calls still fabricate `(1,1,...)` with no filename.**

The reported repro is the decider. Issue #2987 names two files. Under `gsi` on the merged head:

```
samples/RefStructSpan.gs   samples/RefStructSpan.gs(19,40,19,48): error GS0511   <- real location
samples/SpanIndexing.gs    (1,1,1,138): error GS0511                             <- no filename, line 1
```

and `SpanIndexing.gs` prints, as its "source line", an assembly-qualified CLR type dumped from `WriteTo`:

```
(System.ReadOnlySpan`1[[System.Int32, System.Private.CoreLib, Version=10.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]])values
```

`gsi` does carry the filename where a location exists, so #3024's `arg` plumbing is intact — the missing filename here is the fabricated-location path, not `Repl/Program.cs`.

**The PR's own theory data contains the broken shapes.** `Issue2987ByRefLikeBoundaryInterpreterTests.cs:31-77` — case 2 (property get), case 3 (property set) and case 5 (implicit `[]int32` → `ReadOnlySpan[int32]` conversion) are exactly the shapes that still fabricate; each asserts only `Id` and message substrings. The single location test, `ByRefLikeDiagnostic_UsesOriginalSourceLocation:97`, uses `MemoryExtensions.AsSpan(...)` — the *call* shape, the one that works. The suite therefore runs the broken shapes and asserts nothing that would catch them.

**Cause located, and the completion is one line.** The PR fixes three binder sites (`ExpressionBinder.Calls.Arguments.cs:1559`, `ExpressionBinder.Calls.Invocation.cs:714,2216`) plus two rewriter sites (`BoundTreeRewriter.cs:1230,1273`). It leaves:

- `ExpressionBinder.Calls.Invocation.cs:2775` and `ExpressionBinder.Calls.Arguments.cs:1126` — imported **instance** call binder paths, still `new BoundImportedInstanceCallExpression(null, ...)`;
- ~11 user-code binder paths constructing `new BoundClrPropertyAccessExpression(null, ...)`, notably `ExpressionBinder.Access.MemberLookup.cs:296,299,393,399,493,499,544,586,646,665,699`.

Proved causally rather than by reading: mutating **`ExpressionBinder.Access.MemberLookup.cs:299`** from `null` to `rightPart` (mutant asserted applied — 1 new pattern present, 0 of the old remaining; `BUILD_RC=0`) turned `g2.gs` from `(1,1,1,13)` into `g2.gs(4,25,4,30)` — correct filename, correct line, column landing on `Empty`. Reverted; porcelain empty.

Recommendation: extend the fix to the property-access and instance-call binder sites, and add a location assertion to at least one property-shape theory case. **If that is deferred, please do not close #2987 on this PR** — one of the two samples it names still reports a fabricated location.

**SHOULD-FIX 2 — `samples/SpanIndexing.gs:9` comment overstates what ships.** The added line says the interpreter "reports GS0511; compile this sample with gsc /out:<file>", which is true of the ID but reads as though the diagnostic is now well-formed. For this sample it still points at line 1 of a synthesised CLR type name. Either complete SHOULD-FIX 1 or soften the comment.

### Round-1 NITs — both resolved

- Message now reads `compile this program with 'gsc /out:<file>' instead`, which is accurate: bare `gsc` evaluates and fires the diagnostic, so the old "compile with 'gsc'" was self-contradictory.
- `EvaluatorException.cs:86-88` now guards `messageArguments.Length == 0` before `string.Format`, so the latent `FormatException` on a brace-bearing zero-argument descriptor is gone.

### No false positives from propagating `Syntax`

Setting real syntax on `BoundImportedCallExpression` / `BoundImportedInstanceCallExpression` reaches every imported call, not just ByRefLike ones. Corpus of 6 correct programs (strings, interpolation, ctor+property, operators, collections, parsing) × **all three drivers** (bare `gsc` evaluate, `gsc /out:` emit into an asserted-empty per-probe directory, `gsi` interpret), base vs head: **rc identical on all 18 pairs, and the evaluate output md5 is byte-identical on all 6.** Plus `Core.Tests` 7088/0/7089 on the merged tree.

### PR #3104 interaction — they do **not** co-merge; reconcilable in ~6 lines

Merging #3055 then #3104 onto `6ec381c5` gives **`MERGE_RC=1`**, two conflicted files:

- **`Compilation.cs:329-343`** — two different solutions to the same problem. #3055 uses `ex.Node?.Syntax` → `Syntax.Location`; #3104 introduces `ex.Location` as explicit plumbing and makes `location` nullable.
- **`DiagnosticDescriptors.cs:382-387`** — #3104 rewrites GS0514's message text on the line adjacent to #3055's new GS0511.

Unusually, and worth recording given the #3024+#3052 experience: **every naive resolution here is loud, not silent.** Taking #3055's side of `Compilation.cs` leaves #3104's `location.Value` applied to a non-nullable `TextLocation` → build error. Taking #3104's side deletes the `Syntax` fallback → `ByRefLikeDiagnostic_UsesOriginalSourceLocation` fails. Taking #3104's side of the descriptor block deletes `GS0511` → build error from `Evaluator.Reflection.cs`. There is no resolution that compiles and passes while quietly dropping either PR's behaviour.

I built the union resolution — `ex.Location ?? ex.Node?.Syntax?.Location ?? fabricated`, both descriptor lines kept — and measured it: **`BUILD_RC=0`, 0 warnings / 0 errors, `Interpreter.Tests` 653/653** (both PRs' tests), and the location table on the co-merged tree is **identical to #3055-alone on all nine rows**, including the two still-fabricated ones. So the PRs are compatible once reconciled by hand.

**Order: land #3104 first.** Its `ex.Location` is the general mechanism, plumbed at throw sites; #3055's `ex.Node.Syntax` is a derived fallback that composes underneath it. In that order #3055 rebases to a 4-line change (add the `&& ex.Node?.Syntax != null` clause). In the other order #3104 must restructure #3055's `if/else` into a fallback, which is the larger and more error-prone edit.

### Disclosed as not run

`Compiler.Tests` and `Cs2Gs.Tests` on the merged tree (the former is ~1h20m; the syntax-propagation change does reach emit, and my emit coverage is the 6-program corpus at rc-level only, not IL comparison). e2e not run — prohibited. The author's own three mutants not re-run; I ran my own two instead.

PR #3056, opened 2026-08-01
Test generic local functions returning lambdas

### LINKED ISSUE #3050: gsi: generic zero-capture lambda returned from a generic method throws GS9999
## Problem

A zero-capture lambda returned from a **generic** method diverges between the compiler and the interpreter:

```
gsc + dotnet : 7
gsi          : GS9999  rc=1
```

## Provenance

Found while rubber-ducking PRs #3045 (`Fixes #2991`) and #3046 (`Fixes #2990`). **Measured identical at the merge base of both PRs**, so it is pre-existing on `main` and is neither PR's fault. It nearly became a false blocker against #3045; checking the base is what stopped it.

## Notes for whoever picks this up

- **`gsc` is the oracle** — the compiled result (`7`) is correct.
- This is *not* the silent class: `gsi` does exit **1** here via the evaluator `GS9999` path (`Compilation.cs:334` → `SessionEngine.cs:212` → `Program.cs:68`). But `GS9999` is the **internal-error bucket**, never a deliberate refusal — so the end state must be either a real fix or a named boundary diagnostic, not a `GS9999`. `GS0506` is the prior art for a deliberate refusal.
- Related but distinct: #2991 (distinct lambdas reporting the same `Method.Name`) and #2990 (untyped lambda not coerced to `Func` at a CLR call boundary). Reviewer established the **erasure** to `object` is the trigger for that family, not parameter arity — a breadth matrix over primitive types like `int32` passes at base and proves nothing. Expect the same to apply here: the signature likely needs a **G#-declared type** to reproduce.
- `Evaluator.Expressions.cs:871` carries a `!node.Function.IsGeneric` condition that a reviewer measured as **surviving 527/527 tests** when mutated — i.e. unpinned, and plausibly related to this generic-host path. Worth checking whether it is dead, wrong, or the actual site.

DavidObando:
## Summary

- Add regression coverage for a generic local function returning a zero-capture lambda whose signature contains a G#-declared type.
- Assert distinct `11/22/33` values and `gsc`/`gsi` parity.

## Root cause correction

The issue body says “generic method,” but a generic member method returning the same lambda already prints `7` in both engines at the stated pre-fix base `fd34166a8`, including with a G#-declared type.

The actual failing shape is a generic **local function**:

```gsharp
let MakeIdentity[T] = func() (T) -> T {
    return (value T) -> value
}
```

At `fd34166a8`, `gsi` reports `GS9999: Unexpected node LocalFunctionDeclaration`. The production fix already landed in #3042: `Evaluator.Statements.cs:74` registers the lowered `BoundLocalFunctionDeclaration` body, and `Evaluator.Calls.cs:76` falls back to that body for direct calls. Current `main` therefore needs only the missing #3050 regression pin.

## Validation

- `dotnet build` — 0 warnings, 0 errors
- `dotnet test test/Interpreter.Tests/Interpreter.Tests.csproj --filter 'FullyQualifiedName~Issue3030LocalFunctionDeclarationInterpreterTests'` — 3 passed
- Differential: `gsc` and `gsi` both print `11 / 22 / 33`

## Mutation

Each existing root-fix hunk was reverted independently; every mutant built successfully and the new #3050 case failed:

- Remove `LocalFunctionDeclaration` dispatch — Core hash `01b5c736…`; fails with `Unexpected node LocalFunctionDeclaration`.
- Keep dispatch but remove body registration — Core hash `e8701787…`; fails with missing `MakeIdentity` body.
- Remove call-site fallback to `localFunctionBodies` — Core hash `1040c759…`; fails with missing `MakeIdentity` body.
- Restored Core hash: `4990ad1f…`.

No production overlap with #3046 or #3020; this PR changes one interpreter test file only.

Fixes #3050

### DIFF
--- a/test/Interpreter.Tests/Issue3030LocalFunctionDeclarationInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue3030LocalFunctionDeclarationInterpreterTests.cs
@@ -13,7 +13,7 @@ using Xunit;
 namespace GSharp.Interpreter.Tests;
 
 /// <summary>
-/// Issue #3030: generic local-function declarations are evaluator no-ops.
+/// Issues #3030 and #3050: generic local-function declarations register bodies for direct calls.
 /// </summary>
 public class Issue3030LocalFunctionDeclarationInterpreterTests
 {
@@ -44,6 +44,23 @@ public class Issue3030LocalFunctionDeclarationInterpreterTests
             PrintValues()
             """,
             "42\nnested\n");
+
+        yield return Case(
+            """
+            data struct Item {
+                var Value int32
+            }
+
+            let MakeIdentity[T] = func() (T) -> T {
+                return (value T) -> value
+            }
+
+            let identity = MakeIdentity[Item]()
+            Console.WriteLine(identity(Item{ Value: 11 }).Value)
+            Console.WriteLine(identity(Item{ Value: 22 }).Value)
+            Console.WriteLine(identity(Item{ Value: 33 }).Value)
+            """,
+            "11\n22\n33\n");
     }
 
     [Theory]


DavidObando:
## Review: PR #3056 — **MERGEABLE**

Tree measured: **merge of `main` `46d3cc50f` + head `92ae7ef0b`** = `64ff4b17b`. Merge base `fdba684b6` (branch one merge behind main; merges clean). Three-dot scope: **1 file, +18 −1**. Two-dot lied (29 files / +168 −2846) — ignored.

### Headline: YES — the test fails when the fix is absent. It is not vacuous.

No production code in this PR, so I mutated **#3042's fix in `main`** (merge commit `4c5b0bf97`, 3 hunks: `Evaluator.cs` +1, `Evaluator.Statements.cs` +8, `Evaluator.Calls.cs` +3−1). Each reverted **individually**, on the merged tree, condition/structure revert (not `if (true)` — warnings-as-errors).

| mutant | build rc | new #3050 case | symptom |
|---|---|---|---|
| clean | 0 | **PASS** (3/3) | — |
| **M-1** delete `case BoundNodeKind.LocalFunctionDeclaration` (`Evaluator.Statements.cs:130-137`) | **0** | **FAIL** | `Unexpected node LocalFunctionDeclaration` |
| **M-2** revert fallback → `program.Functions[node.Function]` (`Evaluator.Calls.cs:76-78`) | **0** | **FAIL** | `The given key 'func MakeIdentity() (T) -> T' was not present in the dictionary.` |
| restored | 0 | **PASS** (3/3) | `git status --porcelain` empty |

M-1's symptom is **verbatim** the pre-fix symptom recorded in #3050's own comment (`gsi` at `fd34166a8`: `GS9999: Unexpected node LocalFunctionDeclaration`). Both mutants built **rc=0** — good mutants, not CS0162/StyleCop casualties. Core hash discriminated: mutant `b46a2063…` vs restored-clean `0a8c511f3…`; exact restore additionally proven by empty `git diff` vs HEAD.

Shape is correct for this family: **G#-declared `data struct Item`** (so the delegate erases to `object` — erasure is the trigger, not arity), explicit instantiation `MakeIdentity[Item]()`, and assertions on **distinct values `11/22/33`**, not `rc=0`. `Assert.Empty(result.Diagnostics)` is live — it is what fired under both mutants, so this is not the #2979 `DoesNotContain(IsError)` failure mode.

Docs gate: PR touches **neither** diagnostics file. Merged tree **383 / 381** rows — identical to `main`. Green.

Body is honest: it states plainly that the production fix already landed in #3042 and that this is a pin only. Its reported mutant *symptoms* match mine exactly (hashes differ — different tree and machine, expected).

---

### NIT 1 — body implies the test asserts `gsc`/`gsi` parity; it does not
`test/Interpreter.Tests/Issue3030LocalFunctionDeclarationInterpreterTests.cs:69-85` runs **only** `Compilation.Evaluate` — interpreter leg alone. There is no `gsc` compile-and-run leg. The Summary bullet "Assert distinct `11/22/33` values and `gsc`/`gsi` parity" reads as if the test guards divergence; only the Validation section reveals the differential was a **manual** check. Since #3050 *is* a `gsc`/`gsi` divergence bug, a reader may believe this pin catches that class — it does not: if `gsc` regressed here, this test stays green. Fix the wording, not the test (`gsc` is the oracle and is correct; an interpreter-only pin is the right lazy choice).

### NIT 2 — zero *incremental* mutation coverage over the two pre-existing cases
Both M-1 and M-2 killed **all 3** cases, not just the new one. The two pre-existing #3030 cases already pin both of #3042's hunks. **I could not construct a mutant of #3042's fix that the new case kills while the old two survive — my probe failed to discriminate**, and I am reporting that rather than dressing it up.

So the new case's value is **shape documentation**, not new pinning: it uniquely exercises a returned *function value* + explicit type argument + struct erasure, which a future refactor could break independently of direct generic-local-fn calls. That is legitimate for a separately-reported issue, and 18 lines is proportionate. Not blocking — just do not credit this PR with coverage it does not add.

---

### Measurement error I made, disclosed — and a doctrine correction

My first three hash readings were **void**. I hashed `out/bin/Release/**Compiler**/GSharp.Core.dll`, which `dotnet build src/Core/Core.csproj` **does not refresh**. My "clean" baseline `b5ad19f00…` turned out to be a **13:49 leftover from the #3005 review** (same bytes still sitting in `out/bin/Release/Core.Tests/`). It produced two classic phantoms in one shot: a "restore" that looked wrong (`41d72811…` ≠ baseline) and a **live mutant that looked dead** (M-2 hash identical to restore, despite rc=0 and a real source edit).

Timestamps after a Core-only build make it unambiguous: `out/bin/Release/Core/…` = 14:51:53, `out/bin/Release/Compiler/…` = 14:49:12 (stale).

**There are ~30 copies of `GSharp.Core.dll` under `out/`.** The standing rule "hash `GSharp.Core.dll`, not the `gsc`/`gsi` driver" is ambiguous, because the trap copy lives **in the driver directory** — the natural reading sends you straight to the stale file. Corrected rule:

> Hash **`out/bin/Release/Core/GSharp.Core.dll`** after `dotnet build src/Core/Core.csproj`, and **check its mtime is newer than the build** before believing it.

Test results here are unaffected — `dotnet test` compiles from source, and both mutants produced distinct, correct, mutant-specific symptoms.

### Verdict
**MERGEABLE.** No BLOCKER, no SHOULD-FIX. The decisive question — *does this test fail on a tree where the bug is present?* — is **yes**, measured twice with distinct symptoms, one of them verbatim the issue's recorded pre-fix error. NIT 1 is a two-word body edit; NIT 2 is informational.

Issue #2793, opened 2026-07-23
gsc: user generic metadata caches ignore active lambda generic remaps


DavidObando:
The user generic TypeSpec, field MemberRef, and method MemberRef caches in UserTokenResolver are keyed by the active class/state-machine remap but not ActiveLambdaMethodTypeParamRemap. A user-declared generic struct referenced inside a generic-promoted lambda can therefore reuse metadata encoded under the wrong MVAR scope, analogous to #2785's function-delegate cache leak.

Relevant caches:
- userStructTypeSpecCache
- userStructFieldRefCache
- userStructMethodRefCache

Find or add a minimal generic-promoted-lambda repro, key every encoding-sensitive cache by both remap channels, and add runtime plus ILVerify coverage. Fix generally; no Oahu-specific behavior.

Discovered while reviewing #2791.

PR #3057, opened 2026-08-01
Fix imported constructor cache remap identity

### LINKED ISSUE #2930: Imported symbolic constructor cache ignores lambda remaps

DavidObando:
## Summary

**This PR is now regression coverage plus a doc-comment correction.** The production cache-key change it originally carried has already landed in `main` via #2912 (`da0e2067b`), so the duplicate hunk was dropped here as planned. What remains:

- `test/.../Issue2930ImportedConstructorCacheRemapTests.cs` — executes emitted programs across both remap channels, asserting distinct two-call values (`11/22`, `33/44`), and loads emitted assemblies with `Assembly.Load` + `GetTypes()` before runtime execution (ILVerify remains supplemental, not sufficient on its own).
- `test/.../MetadataTokenCacheTests.cs` — direct key-equality coverage for `CtorRefSymbolKey`.
- `src/Core/CodeAnalysis/Emit/MetadataTokenCache.cs` — **doc comment only, no behaviour change.** The old comment claimed `CtorRefSymbolKey` was the "Counterpart to `MethodSpecSymbolKey`". That is now false and reads as false assurance: `MethodSpecSymbolKey` is still 2-arg and carries no remap identity. The comment now says so and points at #3065, which tracks the equivalent fix.

## ⚠️ Do not delete `SeparatesClosureClassRemaps` as a duplicate of the Issue2898 tests

`main`'s shipped fix has two halves — a **method** remap identity (MVAR) and a **class** remap identity (VAR). Partial-mutant measurement of the two halves independently:

| mutant applied to `main` | Issue2898 tests | `SeparatesClosureClassRemaps` |
|---|---|---|
| drop `methodRemap` (MVAR) from equality + hash | **dies** | dies |
| drop `classRemap` (VAR) from equality + hash | **survives** | **dies — sole failure** |

So the `classRemap` half of the shipped production fix is pinned by **this test and nothing else**. It is not redundant with the nested-enum coverage added by #3063 either: `Remap[A,B]` and `Iterate[A,B]` are separate declarations, so their `B` are different `TypeParameterSymbol` *instances*, and the cache key compares `typeArgs` **by reference** — they can never collide. A collision requires **one symbol crossing a remap boundary**, which only the closure test in this PR constructs. Adding more call sites in more methods cannot reach it.

## Root cause (for context; fix already in `main`)

`CtorRefSymbolKey` keyed only constructor + symbolic type arguments. The same symbols encode to different `VAR`/`MVAR` ordinals under different active remaps, so a second constructor lookup reused a MemberRef parented by the first scope's TypeSpec.

## Validation

- `dotnet build GSharp.sln` — 0 warnings/errors
- issue tests — 2/2 pass
- related #2793/#671/#693 cache tests — 24/24 pass
- Bulletin #3 audit: probes are not wrapped in `func Main()`; both top-level and in-function shapes exercised under `gsc` and `gsi`.

## Overlap

#3036 touches `ImportedMemberRefFactory.GetElementTypeToken`; no hunk collision with the test files or the doc comment here.

Fixes #2930

### DIFF
--- a/src/Core/CodeAnalysis/Emit/MetadataTokenCache.cs
+++ b/src/Core/CodeAnalysis/Emit/MetadataTokenCache.cs
@@ -455,10 +455,11 @@ internal sealed class MetadataTokenCache
     }
 
     /// <summary>
-    /// Issue #671: structural cache key for ctor MemberRef rows whose parent
-    /// TypeSpec carries G# user-defined symbolic type arguments. Includes both
-    /// active generic-remap scopes because the same symbols can encode to
-    /// different VAR/MVAR ordinals. Counterpart to <see cref="MethodSpecSymbolKey"/>.
+    /// Issues #671 and #2930: structural cache key for ctor MemberRef rows
+    /// whose parent TypeSpec carries G# user-defined symbolic type arguments.
+    /// Includes both active generic-remap scopes because the same symbols can
+    /// encode to different VAR/MVAR ordinals. Issue #3065 tracks equivalent
+    /// remap discrimination for <see cref="MethodSpecSymbolKey"/>.
     /// </summary>
     internal readonly struct CtorRefSymbolKey : IEquatable<CtorRefSymbolKey>
     {
--- a/test/Compiler.Tests/Emit/Issue2930ImportedConstructorCacheRemapTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2930ImportedConstructorCacheRemapTests.cs
@@ -1,0 +1,153 @@
+// <copyright file="Issue2930ImportedConstructorCacheRemapTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+using GSharp.Compiler;
+using Xunit;
+
+namespace GSharp.Compiler.Tests.Emit;
+
+/// <summary>
+/// Issue #2930: imported symbolic constructor cache entries include active
+/// generic remap identities.
+/// </summary>
+public class Issue2930ImportedConstructorCacheRemapTests
+{
+    [Fact]
+    public void ImportedConstructorCache_SeparatesLambdaMethodRemaps()
+    {
+        const string Source = """
+            package Issue2930Method
+
+            import System
+            import System.Collections.Generic
+
+            func Remap[A, B](first A, value B) B {
+                var outer = List[B]()
+                outer.Add(value)
+                let f (B) -> B = (innerValue B) -> {
+                    var inner = List[B]()
+                    inner.Add(innerValue)
+                    return inner[0]
+                }
+                return f(outer[0])
+            }
+
+            Console.WriteLine(Remap[int32, int32](0, 11))
+            Console.WriteLine(Remap[int32, int32](0, 22))
+            """;
+
+        AssertRunsWithExactOutput(Source, nameof(ImportedConstructorCache_SeparatesLambdaMethodRemaps), "11\n22\n");
+    }
+
+    [Fact]
+    public void ImportedConstructorCache_SeparatesClosureClassRemaps()
+    {
+        const string Source = """
+            package Issue2930Class
+
+            import System
+            import System.Collections.Generic
+
+            func Remap[A, B](first A, value B) B {
+                var outer = List[B]()
+                outer.Add(value)
+                var calls = 0
+                let f (B) -> B = (innerValue B) -> {
+                    calls++
+                    var inner = List[B]()
+                    inner.Add(innerValue)
+                    return inner[0]
+                }
+                return f(outer[0])
+            }
+
+            Console.WriteLine(Remap[int32, int32](0, 33))
+            Console.WriteLine(Remap[int32, int32](0, 44))
+            """;
+
+        AssertRunsWithExactOutput(Source, nameof(ImportedConstructorCache_SeparatesClosureClassRemaps), "33\n44\n");
+    }
+
+    private static void AssertRunsWithExactOutput(string source, string name, string expected)
+    {
+        var assemblyPath = Compile(source, name);
+        var assembly = Assembly.Load(File.ReadAllBytes(assemblyPath));
+        Assert.NotEmpty(assembly.GetTypes());
+        Assert.Equal(expected, Run(assemblyPath, name));
+        IlVerifier.Verify(assemblyPath);
+    }
+
+    private static string Compile(string source, string name)
+    {
+        var directory = Path.Combine(AppContext.BaseDirectory, nameof(Issue2930ImportedConstructorCacheRemapTests), name);
+        Directory.CreateDirectory(directory);
+        var sourcePath = Path.Combine(directory, "test.gs");
+        var assemblyPath = Path.Combine(directory, name + ".dll");
+        File.WriteAllText(sourcePath, source);
+
+        using var stdout = new StringWriter();
+        using var stderr = new StringWriter();
+        var previousOut = Console.Out;
+        var previousErr = Console.Error;
+        Console.SetOut(stdout);
+        Console.SetError(stderr);
+        int exitCode;
+        try
+        {
+            exitCode = Program.Main(new[]
+            {
+                "/out:" + assemblyPath,
+                "/target:exe",
+                "/targetframework:net10.0",
+                sourcePath,
+            });
+        }
+        finally
+        {
+            Console.SetOut(previousOut);
+            Console.SetError(previousErr);
+        }
+
+        Assert.True(exitCode == 0, $"{name}: gsc failed:\n{stdout}\n{stderr}");
+        return assemblyPath;
+    }
+
+    private static string Run(string assemblyPath, string name)
+    {
+        using var process = Process.Start(new ProcessStartInfo("dotnet")
+        {
+            ArgumentList =
+            {
+                "exec",
+                "--runtimeconfig",
+                Path.ChangeExtension(assemblyPath, ".runtimeconfig.json"),
+                assemblyPath,
+            },
+            WorkingDirectory = Path.GetDirectoryName(assemblyPath),
+            RedirectStandardOutput = true,
+            RedirectStandardError = true,
+            UseShellExecute = false,
+        });
+
+        Assert.NotNull(process);
+        var outputTask = process.StandardOutput.ReadToEndAsync();
+        var errorTask = process.StandardError.ReadToEndAsync();
+        var exited = process.WaitForExit(30_000);
+        if (!exited)
+        {
+            process.Kill(entireProcessTree: true);
+            process.WaitForExit();
+        }
+
+        Assert.True(exited, $"{name}: emitted program timed out");
+        var output = outputTask.GetAwaiter().GetResult();
+        var error = errorTask.GetAwaiter().GetResult();
+        Assert.True(process.ExitCode == 0, $"{name}: emitted program failed:\n{error}");
+        return output.Replace("\r\n", "\n");
+    }
+}
--- a/test/Core.Tests/CodeAnalysis/Emit/MetadataTokenCacheTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Emit/MetadataTokenCacheTests.cs
@@ -1,0 +1,37 @@
+// <copyright file="MetadataTokenCacheTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using GSharp.Core.CodeAnalysis.Emit;
+using GSharp.Core.CodeAnalysis.Symbols;
+using Xunit;
+
+namespace GSharp.Core.Tests.CodeAnalysis.Emit;
+
+public class MetadataTokenCacheTests
+{
+    [Fact]
+    public void CtorRefSymbolKey_EqualsDistinguishesRemapIdentities()
+    {
+        var ctor = typeof(List<>).GetConstructors().Single(c => c.GetParameters().Length == 0);
+        var typeArgs = ImmutableArray<TypeSymbol>.Empty;
+        var classRemap = new Dictionary<TypeParameterSymbol, int>();
+        var methodRemap = new Dictionary<TypeParameterSymbol, int>();
+        var key = new MetadataTokenCache.CtorRefSymbolKey(ctor, typeArgs, classRemap, methodRemap);
+
+        Assert.False(key.Equals(new MetadataTokenCache.CtorRefSymbolKey(
+            ctor,
+            typeArgs,
+            new Dictionary<TypeParameterSymbol, int>(),
+            methodRemap)));
+        Assert.False(key.Equals(new MetadataTokenCache.CtorRefSymbolKey(
+            ctor,
+            typeArgs,
+            classRemap,
+            new Dictionary<TypeParameterSymbol, int>())));
+    }
+}


DavidObando:
## Review: MERGEABLE

Measured in a scratch clone, not the assigned worktree — `/Users/davidobando/GitHub/DavidObando/gsharp-wt/i2930` **does not exist** (no `i2930` in `git worktree list`). Anchors: head `af544bd12` (matches), `origin/main` `46d3cc50f`, computed merge base `61b7a08c1`, three-dot. Merged tree `main`+PR = `b2eed686f`; its delta against `main` is exactly the PR's 3 files / +176 / −6, so the merge introduced no hidden resolution.

Clean merged build rc=0, `GSharp.Core.dll` md5 `437083b2`. New tests **2 passed / 0 failed / 2 total**. Full `Compiler.Tests`: **4005 passed / 0 failed / 4005 total**, rc=0 (1 h 10 m under load 12→23).

### The fix is genuinely pinned — this PR is not in the survivor set

| mutant | hunk reverted | build | Core.dll md5 | result |
|---|---|---|---|---|
| M1 | construction site → `null, null` (semantically the pre-fix key), `ImportedMemberRefFactory.cs:1341` | rc=0 | `dcaf06af` | **KILLED 2/2** |
| M2 | `Equals` remap conjunction, `MetadataTokenCache.cs:489-490` | rc=0 | `9f5a1033` | SURVIVED 2/2 |
| M3 | `GetHashCode` remap mixing, `MetadataTokenCache.cs:512-513` | rc=0 | `009320a0` | SURVIVED 2/2 |

Round-trip closed: `437083b2` → mutants → `437083b2`, source drift 0. M1's kill is a product kill with the issue's own declared symptom — `System.BadImageFormatException` raised by the **emitted program**, not by the test fixture.

### SHOULD-FIX 1 — the root cause has two sites; this fixes one

`ImportedMemberRefFactory.cs:870` and `:909` build `new MetadataTokenCache.MethodSpecSymbolKey(method, typeArgSymbols)` with **no remap component**, and cache into `MethodSpecsWithSymbolArgs` a MethodSpec whose signature blob is produced by the same `signatures.EncodeTypeSymbol` that consults both remaps (`SignatureEncoder.cs:269-289`). That is the identical defect, in the adjacent branch of the same file.

Probe, run against the **merged (fixed)** tree:

```gsharp
package MsProbe
import System

func Remap[A, B](first A, value B) B {
    var outer = Array.Empty[B]()
    let f (B) -> B = (innerValue B) -> {
        var inner = Array.Empty[B]()
        return innerValue
    }
    return f(value)
}

Console.WriteLine(Remap[int32, int32](0, 11))
Console.WriteLine(Remap[int32, int32](0, 22))
```

`gsc` rc=0; running it gives **rc=134**, `System.BadImageFormatException ... at MsProbe.<Program>.<lambda1>[B](B innerValue)`. Applying the *same* four-field key fix to `MethodSpecSymbolKey` and its two call sites, and nothing else, changes that to **rc=0, output `11` / `22`**. Same root cause, proven by construction rather than by resemblance.

This is not a blocker for merging: #2930 is scoped to the *constructor* cache and this PR closes it completely and correctly. But the PR closes half of a two-site defect, so the underlying bug class survives the merge. Worth a follow-up issue; I have not filed one.

### SHOULD-FIX 2 — the `Equals` half is unobservable (the #3028 mask)

M2 survived 2/2. `GetHashCode`'s remap mixing puts keys with different remaps into different buckets, so `Equals` is never consulted by these tests and the conjunction at `MetadataTokenCache.cs:489-490` is untested.

That matters because `Equals` is the only *guaranteed* discriminator. The bucket index is `hash % bucketCount`, so two distinct remaps share a bucket far more often than a full 32-bit collision would suggest; when they do, an `Equals` that ignores the remaps returns `true` and hands back the wrong ctor again. Today both halves are present, so behaviour is correct — but a future edit that drops the conjunction stays green.

Cheapest instrument: one direct unit test on the struct — two `CtorRefSymbolKey` values with the same `ctor` and `typeArgs` but different remap dictionary objects, assert `!a.Equals(b)`. That kills M2 without a compile-and-run cycle.

M3's survival is reported for completeness and is **not** a defect: `GetHashCode` affects distribution only, and `Equals` retains correctness without it.

### NIT — the doc comment now over-promises symmetry

`MetadataTokenCache.cs:458-461` keeps "Counterpart to `MethodSpecSymbolKey`" while adding "Includes both active generic-remap scopes…". That sentence was true before this change; after it the two structs are no longer counterparts, because `MethodSpecSymbolKey` still lacks the remaps. Someone auditing for exactly this bug class would read it as an assurance that the counterpart is equivalently keyed. Either fix the sibling or amend the sentence.

### Checked and cleared, so nobody re-litigates them

- **`ReferenceEquals` on the remaps is sound.** `PushSmRemap` / `PushLambdaMethodRemap` (`GenericRemapState.cs:148-183`) install a distinct dictionary object per state-machine class and per lambda function; they are not a shared buffer mutated in place. `GenericRemapState.cs:126` already documents object identity as the intended discriminator. A stale-identity miss would also fail safe (cache miss → recompute).
- **Site coverage is compiler-proven.** The widened 4-arg constructor turns any missed `CtorRefSymbolKey` construction into CS7036; there is exactly one site.
- **The pattern is house style, not invention.** Nine pre-existing remap-keyed caches: `UserTokenResolver.cs:812, 822, 832, 1198, 1285, 1319, 1613, 1833` and `ImportedMemberRefFactory.cs:2049`. This applies an established pattern to a cache that was missed, which is the minimal correct fix.
- **The tests do not use the issue's own repro** (`Outer[B].Color`, a nested enum over an enclosing method type parameter); they use `List[B]`. M1 proves the simpler program still exercises the defect, so this is not a finding.
- **Docs gate unaffected**, with counts rather than an assertion: 0 added lines matching `DiagnosticDescriptor|GS[0-9]{4}`, 0 files touched under `docs/` or `website/`. Positive control: the same pattern matches 543 lines in `docs/diagnostics.md`, so that zero is a real zero and not a broken pattern.

### Measurement disclosures

- I dirtied the scratch clone's index with a stray `git checkout pr3057 -- .` before merging, and reset it (`--hard` + `clean -fd`, drift verified 0) before any build. No shared tree was written to; the main checkout ended at drift 0.
- The shared checkout moved from `46d3cc50f` to `f5e9d8925` while I was measuring. Every number above is against `origin/main` `46d3cc50f` as fetched into the scratch clone.
- `dotnet test` rebuilds without `-p:ContinuousIntegrationBuild=true`, so every md5 above comes from a dedicated CIB build, never from a test run.

DavidObando:
## Re-review on current `main` — verdict holds at MERGEABLE, but the PR has changed shape

`main` moved twice since my first pass. Re-measured everything against a freshly fetched `origin/main` = **`9816fbe55`** (note: newer than the `01c1a6c5c` circulating in the last bulletin). PR head unchanged at `af544bd12`, so no rebase on its side. `git merge-tree --write-tree` rc=0, tree `776b1d25` — clean, nothing to resolve. Merged tree `d1424c406`.

### This PR is now test-only

Its production hunks **already landed on `main`** in `da0e2067b` — "Fix generic type display and support constructed nested enums (**#2912**)". That is the very PR issue #2930 pointed at for coverage; in the event #2912 carried the *fix* and no `2930`-named test.

Merged-vs-`main` delta is now exactly **one file, +153, the test file**. Merging is a production no-op.

### It is NOT overtaken — the remaining tests pin something nothing else does

This is the #3025/#3037 question, and here the answer is the opposite. Partial mutants, dropping one remap component at a time at `ImportedMemberRefFactory.cs:1341`:

| mutant | dropped | `Issue2898` (already on `main`) | `…SeparatesLambdaMethodRemaps` | `…SeparatesClosureClassRemaps` |
|---|---|---|---|---|
| M1 | both | dies | dies | dies |
| M4 | `methodRemap` (MVAR) | **dies** | dies | survives |
| M5 | `classRemap` (VAR) | **survives** | survives | **dies** |

**`Issue2898NestedEnumConstructedOuterEmitTests` covers only the `methodRemap` half.** Drop `classRemap` and it passes — the only test in the repository that notices is this PR's `ImportedConstructorCache_SeparatesClosureClassRemaps`. Without this PR, half of `main`'s shipped fix is pinned by nothing.

Evidence: full `Compiler.Tests` on M1 = **3 failed / 4025 passed / 4028 total** (the two new tests plus `Issue2898`). M4 and M5 were scoped runs over `Issue2930|Issue2898`, **12 tests matched** each — reported as a count, not a bare zero, and the filter is stated because these runs answer "which of these three dies", not "does anything fail".

All mutants `BUILD_RC=0` with distinct `Core/GSharp.Core.dll` hashes (`dc0cae78`, `5adbba77`, `27228475`), clean `f45029de`.

### SHOULD-FIX 1 stands, and is now a live `main` defect

`MethodSpecSymbolKey` (`MetadataTokenCache.cs:414`) is still the 2-argument version on `main` today — neither #2912 nor this PR keyed it by the remaps, while `CtorRefSymbolKey` now is. Re-ran the probe against the current merged tree with a **verified-fresh driver** (`out/bin/Release/Compiler/GSharp.Core.dll` mtime 17:20:13 vs build start 17:20:06, same second as `Core/`):

`Array.Empty[B]()` called in method scope and again inside a promoted lambda → `gsc` rc=0, run **rc=134**, `System.BadImageFormatException … at MsProbe.<Program>.<lambda1>[B]`. Applying the identical four-field key to `MethodSpecSymbolKey` and its two call sites (`ImportedMemberRefFactory.cs:870`, `:909`) → **rc=0, prints `11` / `22`**.

Still not a blocker for this PR — it is a pre-existing defect this PR does not claim — but it is worth its own issue now that the constructor half has shipped. The doc comment at `MetadataTokenCache.cs:458-461` calling `MethodSpecSymbolKey` a "Counterpart" is the trap: after #2912 the two are no longer equivalently keyed, and that sentence reads as an assurance they are.

### SHOULD-FIX 2 stands, and got sharper

Reverting only the `Equals` remap conjunction (`MetadataTokenCache.cs:489-490`), leaving `GetHashCode` intact, now survives **12/12** with `Issue2898` in scope — not just 2/2 against this PR's own tests. `GetHashCode`'s remap mixing separates the buckets, so `Equals` is never consulted and the conjunction is pinned by nothing in the repository.

It matters because the bucket index is `hash % bucketCount`; two distinct remaps share a bucket far more often than a 32-bit collision, and when they do, an `Equals` that ignores the remaps hands back the wrong ctor. Behaviour is correct today because both halves are present. Cheapest instrument, and it belongs in this PR since this PR is now the test PR: two `CtorRefSymbolKey` values, same `ctor` and `typeArgs`, different remap dictionary objects, assert not equal. `md5` round trip for that mutant closed exactly, `f45029de` → `44a0e8c9` → `f45029de`.

### Suggested action

Merge, but update the body — it reads as a production fix and is now purely regression coverage for #2912. Worth stating explicitly that `ImportedConstructorCache_SeparatesClosureClassRemaps` is the sole pin for the `classRemap` half, so nobody deletes it as duplicative of `Issue2898`.

### Measurement disclosures

- All builds were `dotnet build GSharp.sln`, never project-scoped, and I verified the driver's `GSharp.Core.dll` mtime matched `Core/`'s to the second before every `gsc` probe. The earlier probe results were taken the same way, so they were not exposed to the stale-driver trap.
- I omitted a clean baseline hash before M1 on the re-measured tree, so that round trip is one-sided (clean-after only); the M2 round trip is complete. Kill/survive conclusions rest on `dotnet test`, which builds from source.
- Scratch clone only. `src/` and `test/` drift 0; the shared checkout was never written to.

DavidObando:
## Re-verified on `main` = `185431c4` after #3063 — the finding holds, and now has a `gsc`/`gsi` differential behind it

#3063 ("Strengthen nested enum regression coverage") edits **exactly** the test my earlier partial mutant showed surviving, and it adds `func Iterate[A, B](c Outer[B].Color) sequence[int32]` containing `List[Outer[B].Color]()` — an iterator, so a state-machine class, so the `ActiveIteratorStateMachineRemap` (`classRemap`) dimension. That looked like it would close the gap I reported and make this PR redundant. Re-measured rather than assumed.

Merged tree `5d3bd146` (`main` `185431c4` + `af544bd12`); delta vs `main` still the one 153-line test file. Clean baseline `Core/GSharp.Core.dll` md5 `0d7f0bfe`, both `Core/` and `Compiler/` copies fresh at 21:33:32 against a 21:33:09 build start.

Dropping only `classRemap` at `ImportedMemberRefFactory.cs:1341`, `BUILD_RC=0`, md5 `f3b1328b`:

```
Failed: 1, Passed: 11, Total: 12
  Failed ImportedConstructorCache_SeparatesClosureClassRemaps
```

**`Issue2898` still passes with `classRemap` dropped, even after #3063.** Round trip closed both-sided this time, `0d7f0bfe` → `f3b1328b` → `0d7f0bfe`, drift 0.

### Why the strengthened test still cannot see it

`Remap[A, B]` and `Iterate[A, B]` are different declarations, so their `B` are different `TypeParameterSymbol` **instances**. The cache key compares `typeArgs` by reference, so those two `List[Outer[B].Color]` constructions never collide regardless of remap — the keys already differ. A collision needs the **same** symbol emitted twice under **different** remaps, which is what `ImportedConstructorCache_SeparatesClosureClassRemaps` sets up by using one `B` in both the enclosing method and its synthesized closure class. Adding more call sites in more methods cannot reach this; it needs one symbol crossing a remap boundary.

So the conclusion is unchanged and now stronger for having been attacked: this test remains the only pin for the `classRemap` half of the shipped fix. It should not be deleted as duplicative of `Issue2898`.

### Four-column axis — `{top-level, in-function}` × `{gsc, gsi}`

Reporting because it was never in this PR's coverage. Distinct values throughout, so no case is confusable with another:

| program | `gsc` | `gsi` |
|---|---|---|
| this PR's closure case (in-function) | rc=0, `33` `44` | rc=0, `33` `44` |
| top-level closure over `List[int32]` | rc=0, `33` `44` | rc=0, `33` `44` |
| `MethodSpec` sibling probe (in-function) | **rc=134, `BadImageFormatException`** | **rc=0, `11` `22`** |

Two things fall out.

**The sibling defect is now demonstrated as a `gsc`/`gsi` differential.** The interpreter evaluates the same source correctly and prints `11` / `22`; the compiler emits a binary that cannot load. That is stronger evidence than the crash alone — the language's own semantics and the emitted metadata disagree, so this is not an ambiguous program.

**The `{top-level, in-function}` axis is structurally unavailable for this defect, and I checked rather than asserting it.** Both remap channels require an enclosing generic declaration — `ActiveLambdaMethodTypeParamRemap` maps an enclosing *method* type parameter to `MVar`, and the symbolic ctor path only engages when a type argument requires symbolic projection. Top-level statements declare no type parameters, so the symbolic branch is never entered there. The top-level row above confirms the closure machinery itself is fine at file scope in both drivers.

### Corrections to my own earlier notes

- I had recorded that `gsi` requires an interactive terminal. That applies only to **piped stdin**; `gsi <file>` runs non-interactively and exits 0. My earlier review had no `gsi` column because of that wrong note.
- My previous comment disclosed a one-sided round trip because I omitted a pre-mutation baseline. Fixed here: baseline captured before the first mutant.
- Every build was `dotnet build GSharp.sln`, and I verified mtime on **both** `out/bin/Release/Core/` and `out/bin/Release/Compiler/` before each `gsc`/`gsi` probe. I did not hash anything under `ref/`.

Verdict unchanged: **MERGEABLE**, as regression coverage for the fix that shipped in #2912, with the body updated to say so.

DavidObando:
Reviewer SHOULD-FIXes closed at `21d85642b443660e55cebb506796b8ebe51d4530`.

- Added `CtorRefSymbolKey_EqualsDistinguishesRemapIdentities`. M2 (drop both remap checks from `Equals`) is killed: 1 failed / 0 passed, `Expected: False`, `Actual: True`.
- Amended comment: “Issues #671 and #2930: structural cache key for ctor MemberRef rows whose parent TypeSpec carries G# user-defined symbolic type arguments. Includes both active generic-remap scopes because the same symbols can encode to different VAR/MVAR ordinals. Issue #3065 tracks equivalent remap discrimination for `MethodSpecSymbolKey`.”
- Re-ran M1 after final rebase: full solution build succeeded (0 warnings/errors), Core/Compiler DLL mtimes matched, and both executable regressions failed with emitted-program `BadImageFormatException` (2 failed / 0 passed). Restored tests pass 2/2.
- Full local suites: Core 7068 passed / 1 skipped / 0 failed; Compiler 4031/4031; cs2gs 1861/1861. CI green.

Content note: #2912 placed the identical production remap-key/call-site fix on `main` during rebase. Current three-dot PR delta is regression tests plus honest cache-key comment; merge-tree is clean.

PR #3059, opened 2026-08-01
Pin interface indexer mismatch diagnostics

### LINKED ISSUE #2915: Unify implicit indexer implementation rules for plain and constructed interfaces

DavidObando:
## Summary

- Pin `GS0187` when a plain interface indexer implementation has the wrong index-parameter type.
- Move the existing nullable-indexer rejection case into a dedicated test that pins `GS0504`, its exact message, and the absence of `GS0187`.

PR #2968 merged before these final review tests could be added; this test-only PR closes that coverage gap. Its production PR declared and closed #2915, #2954, #2960, and #3001.

## Validation

- Rebased onto current `main` `1ea847ebe`; old/new commit patch IDs match in 1 of 1 comparison.
- Targeted filter: 2 passed, 0 failed.
- `GS0504` → `GS0187` descriptor mutation: `NullableIndexerRejected_ReportsGS0504` failed while `PlainInterfaceIndexerParameterMismatch_ReportsGS0187` passed (1 failed / 1 passed). Restore returned 2/2 green.
- Mutation DLL MD5: `03cdf27ccb096a3b4d312e47e2e1f4b3 → 6c80c82eff6b38d4b0f58ef43c381538 → 03cdf27ccb096a3b4d312e47e2e1f4b3`.
- Earlier parameter-comparison mutation: `PlainInterfaceIndexerParameterMismatch_ReportsGS0187` failed while the value-type guard passed; restore returned both green.

### DIFF
--- a/test/Compiler.Tests/Emit/Issue2888InterfacePropertyTypeMismatchTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2888InterfacePropertyTypeMismatchTests.cs
@@ -103,6 +103,52 @@ public class Issue2888InterfacePropertyTypeMismatchTests
     }
 
     [Fact]
+    public void PlainInterfaceIndexerParameterMismatch_ReportsGS0187()
+    {
+        const string source = """
+            package ParameterMismatch
+
+            interface ILookup {
+                prop this[key string] int32 { get; }
+            }
+            class Lookup : ILookup {
+                prop this[key int32] int32 -> key
+            }
+            """;
+
+        var output = CompileExpectingFailure(source);
+
+        Assert.Equal(1, CountOccurrences(output, "error GS0187:"));
+        Assert.Contains(
+            "Class 'Lookup' does not implement interface method 'ILookup.Item'.",
+            output,
+            StringComparison.Ordinal);
+        Assert.DoesNotContain("GS0504", output, StringComparison.Ordinal);
+    }
+
+    [Fact]
+    public void NullableIndexerRejected_ReportsGS0504()
+    {
+        const string source = """
+            package NullableIndexerRejected
+
+            interface I[T] { prop this[key string] T? { get; } }
+            class C : I[int32] {
+                prop this[key string] int32? { get { return nil } }
+            }
+            """;
+
+        var output = CompileExpectingFailure(source);
+
+        Assert.Equal(1, CountOccurrences(output, "error GS0504:"));
+        Assert.Contains(
+            "Type 'C' cannot use property 'Item' to implement interface property 'I[int32].Item': expected type 'int32', actual type 'int32?'.",
+            output,
+            StringComparison.Ordinal);
+        Assert.DoesNotContain("GS0187", output, StringComparison.Ordinal);
+    }
+
+    [Fact]
     public void CompatibleBaselineSiblingShapes_EmitVerifyAndLoad()
     {
         const string source = """
@@ -307,13 +353,6 @@ public class Issue2888InterfacePropertyTypeMismatchTests
             package NullableImplicitSetRejected
             interface I[T] { prop Value T? { get; set; } }
             class C : I[int32] { prop Value int32? { get; set; } }
-            """,
-            """
-            package NullableIndexerRejected
-            interface I[T] { prop this[key string] T? { get; } }
-            class C : I[int32] {
-                prop this[key string] int32? { get { return nil } }
-            }
             """,
             """
             package NullableSliceRejected


DavidObando:
## Rubber-duck verdict: **MERGEABLE**

Measured merged tree: `main` **`f1656079a`** + `18462ef10` -> merged tip `a3443f0df` (scratch clone `/tmp/n59`). Merge rc=0, **0** conflicts. Build rc=0. Merge base computed `1ea847ebe` — matches the PR's rebase claim, but `main` has moved past it, so all numbers below are at `f1656079a`, not `1ea847ebe`.

Scope re-derived three-dot: **1 file, +46 -7**. Baseline whole test class **19/19**.

### 1. PRODUCT mutant on the parameter-comparison path — **the GS0187 test is genuinely pinned**

Descriptor swaps only prove a test reads an ID, so I mutated the analysis instead.

**M1** — `DeclarationBinder.InterfaceVerification.cs:764` `parametersMatch = false;` -> `= true;`, so parameter-type comparison can no longer reject a candidate. Build rc=**0**, md5 `78dec783…` -> `2cebb639…`, `Core/` and `Compiler/` both fresh same-second.

> **Failed 1, Passed 18, Total 19 — `PlainInterfaceIndexerParameterMismatch_ReportsGS0187` and nothing else.**

Isolated kill. The new test dies on the real analysis, not on a descriptor name.

### 2. The absence assertion CAN fail — it is load-bearing, not decoration

This needed care. **M2** (invert the GS0504/GS0187 selector at `:643`) killed the nullable test, but on the **count** assertion, which runs first — that proves nothing about `DoesNotContain`.

So **M3**: keep GS0504 correct and *additionally* emit GS0187 in the same branch (`:653`). Build rc=**0**, md5 -> `950a6204…`. Result:

```
Failed NullableIndexerRejected_ReportsGS0504
  String: ".gs(4,7,4,8): error GS0187: Class 'C' doe"
  Found:  "GS0187"
  at ...Issue2888InterfacePropertyTypeMismatchTests.cs:line 148
```

Fails on **line 148 — the absence assertion itself**. Confirmed can-fail.

### 3. Masking (#3028 lesson) — arms are structurally exclusive, and independently pinned

Read the site before judging it. The two diagnostics are **the two exits of one branch**, selected by whether parameter matching found a candidate:

- no candidate -> `typeMismatch == null` -> `:665` **GS0187**
- candidate found, type incompatible -> `:653` **GS0504**

So no short-circuit can hide the other; `DoesNotContain` asserts which exit was taken. Measured independently:

| mutant | site | Failed / Total | which |
|---|---|---|---|
| M1 param compare always-match | `:764` | **1 / 19** | GS0187 test only |
| M2 selector inverted | `:643` | 5 / 19 | GS0504 test + 4 pre-existing |
| M3 additive GS0187 | `:653` | 2 / 19 | GS0504 test (via absence) + 1 |

Restore -> rebuild -> md5 back to `78dec783…` **exactly**; `git status --porcelain` shows 0 tracked modifications; final run **19/19**.

### 4. The `-7` is a strict STRENGTHENING, not a silent weakening

The removed case came out of the `foreach (var rejected in new[] {...})` array whose consumer is `:395`:

```csharp
_ = CompileExpectingFailure(rejected);
```

It **discards the output** — it only asserted "fails somehow", never which diagnostic. The new dedicated `[Fact]` asserts exactly 1 `GS0504`, the full message, and absence of `GS0187`. Strictly more.

I did not rely on the patch-ID claim: I read the whole diff with `--full-index` (patch `md5`/`index` abbreviations are not stable across fetches). Content checks out; single commit `18462ef1`.

`CompileExpectingFailure` (`:715`) is a sound helper — asserts `exitCode != 0` **and** that no assembly was emitted, each compile in its own `CreateWorkDirectory()`.

### 5. Scope sweep — `func Main()` count is **0** (positive control: 28 `package ` hits)

Correct for this file: these are declaration-only sources and both diagnostics are bind-time, so `gsc` is the right driver and there is no runtime axis to miss.

The one axis not covered is the **declaration kind** — both new tests use `class`, while #2968 was about classes **and** structs (and the struct path was the broken one in #3001). I probed it rather than assume, each into its own empty directory:

| source | rc | diagnostic |
|---|---|---|
| `struct Lookup : ILookup` wrong index param type | 1 | `GS0187: Class 'Lookup' does not implement interface method 'ILookup.Item'.` |
| `struct C : I[int32]` nullable indexer | 1 | `GS0504: Type 'C' cannot use property 'Item' ... expected type 'int32', actual type 'int32?'.` |

**Structs behave identically**, so class-only coverage hides no divergence today. Downgraded to NIT.

### Items

- **BLOCKER**: none.
- **SHOULD-FIX**: none.
- **NIT** — `class`-only coverage for both new tests. Measured equivalent for `struct` today, so this is cheap insurance against a future divergence, not a defect. Two more `[Fact]`s or a `[Theory]` over the declaration keyword.
- **NIT (pre-existing, not this PR)** — `DiagnosticDescriptors.cs:105` hardcodes `"Class '{0}'"`, so a **struct** is reported as `Class 'Lookup'`. Visible in the probe above. Out of scope here; worth its own issue.

### Scope note

I filtered to `Issue2888InterfacePropertyTypeMismatchTests` (19 tests) rather than the full compiler suite: the PR touches exactly one test file, my mutants were reverted with a verified md5 round-trip, and production is byte-identical to `main` at the merged tip. Compiler.Tests is ~1h20m and would add nothing here.

Issue #2947, opened 2026-07-31
GS0156 displays erased nested-enum generic arguments


DavidObando:
## Problem

A newly reachable conversion diagnostic for an imported nested enum over a generic outer displays the CLR-erased projection instead of the source construction:

```text
GS0156: ... glib.Outer[object].Color ...
```

The user wrote a concrete or symbolic `Outer[T].Color`; `object` is an internal generic-projection surrogate and should not leak into source-facing text.

## Expected

GS0156 should preserve the source-level enclosing argument, matching #2898 display behavior, e.g. `glib.Outer[T].Color` or the concrete construction at the diagnostic site.

## Context

Made reachable by nested-enum-in-generic-outer support in PR #2912. Confirmed during round-3 review:
https://github.com/DavidObando/gsharp/pull/2912#issuecomment-5142700741

DavidObando:
## Triage: **NO** — #2947 is *not* closed by PR #2912. It needs its own owner.

Reproduced at PR #2912 head `e03e8fa54` (pristine clone, `GSharp.Core.dll` md5 `67023a44c5372b9f91b61b1ae2b1c840`) against current `main` `a6b1561f2`, invoking `out/bin/Release/Compiler/gsc.dll` directly.

### Repro

`lib.gs` → `glib.dll`:
```gs
package glib
public struct Outer[T] {
    public enum Color { Red = 4, Green = 5, Blue = 6 }
}
```

`z2947c.gs`, compiled with `/r:glib.dll` — a **symbolic** enclosing argument:
```gs
package z2947b
struct Holder[T] {
    public func Take() int32 {
        var c glib.Outer[T].Color = glib.Outer[T].Color.Green
        var i int32 = c
        return i
    }
}
var h Holder[string]
func R() int32 { return h.Take() }
```

### Result

| tree | line 5 diagnostic |
|---|---|
| `a6b1561f2` (main) | `GS0156: Cannot convert type 'glib.Outer[T].Color' to 'glib.Outer[T].Color'` |
| `e03e8fa54` (#2912 head) | **`GS0156: Cannot convert type 'glib.Outer[object].Color' to 'glib.Outer[T].Color'`** |

The `object` surrogate leaks at head — the exact text this issue reports. #2912 does not fix it; consistent with this issue's own note that #2912 is what makes it reachable.

### Two things worth separating

**1. The concrete case *is* fixed by #2912**, which is why this looks half-closed:

```gs
[...]

PR #3063, opened 2026-08-01
Strengthen constructed nested enum regression coverage

### LINKED ISSUE #2946: Support nested enums in constructed generic outer types

DavidObando:
## Context

PR #2912 merged and closed #2946 while this branch was in progress. Its production implementation matches the independently verified root cause in `EnumSymbol`, so this follow-up avoids duplicating that code and keeps only missing regression strength.

## Changes

- Replace the all-zero nested enum fixture with distinct values `11/22/33` and assert values through metadata, methods, boxing, and lambda-remapped generic containers.
- Add a generic iterator path using `List[Outer[B].Color]` and assert the yielded value remains `33` after state-machine lowering.

## Mutation evidence

Forcing `TypeDefEmitter` to emit `0` for every enum literal now fails `NestedEnumConstructedThroughGenericOuter_LoadsWithDistinctClosedSignatures` with `expected 11, actual 0`. Core DLL md5 round-trip: `27570b27894ad0d31c9789dd7f3b4d36 -> 5134e170524c2e216628608e4f880ade -> 27570b27894ad0d31c9789dd7f3b4d36`.

## Verification

- Focused compiler class: 10/10 passed.
- No zero-byte release artifacts.
- All 22 ordinary `out/bin/Release/**/GSharp.Core.dll` copies matched md5 `62b3ac88a4d316c112e7e8ed41540732`.

#2947 is **no longer reproducible on `main`**: a cross-assembly `gsc /r:` repro shows all 5 `GS0156` sites rendering correctly (`Outer[T]`, `Outer[U]`, `Outer[int32]`, `[int32]`->`[string]`, `Outer[B]`) with zero `object`. Verified non-vacuously by a positive control - inverting `FormatImportedType` (`SymbolDisplay.cs:890`) reproduces #2947 verbatim as `'glib.Outer[object].Color'`. #2947 should be closed; this PR does not reference it.

### DIFF
--- a/test/Compiler.Tests/Emit/Issue2898NestedEnumConstructedOuterEmitTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2898NestedEnumConstructedOuterEmitTests.cs
@@ -25,18 +25,18 @@ public class Issue2898NestedEnumConstructedOuterEmitTests
             import System.Collections.Generic
 
             struct Outer[T] {
-                enum Color { Red }
-                func MakeColor() Color { return Color.Red }
+                enum Color { Red = 11, Green = 22, Blue = 33 }
+                func MakeColor() Color { return Color.Green }
             }
 
             func I(c Outer[int32].Color) Outer[int32].Color { return c }
             func S(c Outer[string].Color) Outer[string].Color { return c }
             func NI(c Outer[int32].Color?) Outer[int32].Color? { return c }
             func NS(c Outer[string].Color?) Outer[string].Color? { return c }
-            func RI() Outer[int32].Color { return Outer[int32].Color.Red }
-            func RS() Outer[string].Color { return Outer[string].Color.Red }
+            func RI() Outer[int32].Color { return Outer[int32].Color.Green }
+            func RS() Outer[string].Color { return Outer[string].Color.Blue }
             func BI() object { return Outer[int32].Color.Red }
-            func BS() object { return Outer[string].Color.Red }
+            func BS() object { return Outer[string].Color.Green }
             func Capture[T](c Outer[T].Color) () -> Outer[T].Color -> () -> c
             func Remap[A, B](c Outer[B].Color) int32 {
                 var outer = List[Outer[B].Color]()
@@ -49,6 +49,11 @@ public class Issue2898NestedEnumConstructedOuterEmitTests
                     return int32(inner[0]) + int32(innerRed)
                 }
                 return int32(outerRed) + f(outer[0])
+            }
+            func Iterate[A, B](c Outer[B].Color) sequence[int32] {
+                var values = List[Outer[B].Color]()
+                values.Add(c)
+                yield int32(values[0])
             }
             struct Holder {
                 var IntRed Outer[int32].Color
@@ -87,8 +92,8 @@ public class Issue2898NestedEnumConstructedOuterEmitTests
         Assert.NotNull(intRed);
         Assert.NotNull(stringRed);
         Assert.Equal(intRed.MetadataToken, stringRed.MetadataToken);
-        Assert.Equal(0, intRed.GetRawConstantValue());
-        Assert.Equal(0, stringRed.GetRawConstantValue());
+        Assert.Equal(11, intRed.GetRawConstantValue());
+        Assert.Equal(11, stringRed.GetRawConstantValue());
         Assert.Equal(intEnum, intRed.FieldType);
         Assert.Equal(stringEnum, stringRed.FieldType);
 
@@ -99,13 +104,14 @@ public class Issue2898NestedEnumConstructedOuterEmitTests
         var outerDefinition = allTypes.Single(t => t.Name == "Outer`1");
         var intOuter = outerDefinition.MakeGenericType(typeof(int));
         Assert.Equal(intEnum, GetMethod(intOuter, "MakeColor").ReturnType);
+        Assert.Equal(22, Convert.ToInt32(GetMethod(intOuter, "MakeColor").Invoke(Activator.CreateInstance(intOuter), null)));
 
         var intValue = GetMethod(program, "RI").Invoke(null, null);
         var stringValue = GetMethod(program, "RS").Invoke(null, null);
         Assert.Equal(intEnum, intValue.GetType());
         Assert.Equal(stringEnum, stringValue.GetType());
-        Assert.Equal(0, Convert.ToInt32(intValue));
-        Assert.Equal(0, Convert.ToInt32(stringValue));
+        Assert.Equal(22, Convert.ToInt32(intValue));
+        Assert.Equal(33, Convert.ToInt32(stringValue));
         Assert.Equal(intEnum, GetMethod(program, "BI").Invoke(null, null).GetType());
         Assert.Equal(stringEnum, GetMethod(program, "BS").Invoke(null, null).GetType());
 
@@ -116,7 +122,12 @@ public class Issue2898NestedEnumConstructedOuterEmitTests
 
         var remap = GetMethod(program, "Remap").MakeGenericMethod(typeof(int), typeof(string));
         Assert.Equal(stringEnum, Assert.Single(remap.GetParameters()).ParameterType);
-        Assert.Equal(0, remap.Invoke(null, new[] { stringValue }));
+        Assert.Equal(55, remap.Invoke(null, new[] { stringValue }));
+
+        var iterate = GetMethod(program, "Iterate").MakeGenericMethod(typeof(int), typeof(string));
+        Assert.Equal(stringEnum, Assert.Single(iterate.GetParameters()).ParameterType);
+        var sequence = Assert.IsAssignableFrom<IEnumerable>(iterate.Invoke(null, new[] { stringValue }));
+        Assert.Equal(new[] { 33 }, sequence.Cast<object>().Select(Convert.ToInt32));
     }
 
     [Fact]


DavidObando:
## Review: PR #3063 — **MERGEABLE** (with one factually wrong body claim to fix)

Tree measured: **merge of `main` `1ea847ebe` + head `fc69cfd4f`** = `5b86e71f6`. Merge base `9816fbe55`; three-dot = **1 file, +21 −10**; `git merge-tree` = 0 conflicts. Full-solution Release build `BUILD_RC=0`, 0 warnings. Clean Core md5 `130fa801c…`; `Core/` and `Compiler/` DLLs same-second fresh before every driver probe.

Scope note that governs everything below: the PR changes **only test method #1** (`NestedEnumConstructedThroughGenericOuter_LoadsWithDistinctClosedSignatures`, `:21-131`). The other **9** methods in the class are untouched, so any mutant that kills them carries **zero** information about this PR.

---

### (a) Per-hunk mutation of #2912 — both hunks were **already pinned**; neither needs this PR

Every mutant built `BUILD_RC=0` (no CS0162/StyleCop casualties) and moved the Core hash. "NEW" = PR test file, "OLD" = `main`'s test file, **same production mutant**.

| mutant | site | NEW | OLD | incremental? |
|---|---|---|---|---|
| **M-2** revert #2912 field-signature hunk (`memberType`/`encodeTypeSymbol`) | `TypeDefEmitter.cs:837-849` | 3 failed *(incl. PR test)* | **3 failed** *(incl. PR test)* | **NO** |
| **M-3** remove `EmitGenericParamRows(...)` | `TypeDefEmitter.cs:821` | **10/10 failed** | saturated (9 untouched tests die) | **NO** |
| **M-1b** `value: 0` — *the author's own mutant* | `TypeDefEmitter.cs:864` | 3 failed *(incl. PR test)* | **2 failed — PR test SURVIVES** | **YES** |
| **M-1** `Members[i]` → `Members[0]` | `TypeDefEmitter.cs:864` | **2 failed — PR test PASSED** | same | gap (below) |

**The honest read: the one mutant that demonstrates incremental value is not a #2912 hunk at all.** Both #2912 hunks I reverted were fully caught by the pre-existing test. What this PR actually closes is the **all-zero fixture** hole: with `Red = 0` everywhere, `Assert.Equal(0, …)` could not distinguish "correct value" from "default/zero", and M-1b proves it — OLD survives, NEW dies `Expected: 11, Actual: 0`. That is a real hole and worth closing. Author's mutation claim **verified**.

### (c) Incremental pinning is real but **much narrower than the body implies** — exactly one assertion pair

Every mutant that killed the PR's test killed it at **`:95`** (`Assert.Equal(11, intRed.GetRawConstantValue())`), the *first* changed assertion. So under **all** killing mutants, everything downstream never executed:

- `:107` `Assert.Equal(22, MakeColor().Invoke(...))`
- `:113-114` `Assert.Equal(22/33, …)`
- `:125` `Assert.Equal(55, remap.Invoke(...))`
- `:127-130` the **entire new `Iterate` generic-iterator block**

And **M-1 is the one mutant that runs past `:95`** (it preserves `Red = 11` and corrupts only `Green`/`Blue`) — under it lines `:107-130` all **passed**. So I have **no mutant demonstrating that any of the added runtime-value assertions, or the new iterator path, can ever fail.** They are unproven, not disproven — but the body's "asserts values through metadata, methods, boxing, and lambda-remapped generic containers" credits pinning I could not measure. **Reporting that my probe failed to discriminate rather than dressing it up.**

### NIT — `GetRawConstantValue` is only asserted for the **first** member
`:95-96` assert `Red` (= `Members[0]`) only. **M-1** corrupts `Green`/`Blue`'s metadata `Constant` rows while preserving `Red`, and the PR's test **passes**. This is user-visible: the `Constant` row is what a C# consumer referencing the emitted G# assembly reads for `Outer<int>.Color.Green`. The runtime assertions do not cover it because the emitted IL inlines the constant rather than reading metadata. One line closes it — `Assert.Equal(22, intEnum.GetField("Green", …).GetRawConstantValue())`.

---

### (b) **#2947 is NOT live on current `main` — the body's claim about it is false**

The body states: *"#2947 remains separate: its imported symbolic argument still binds as `Outer[object].Color`."* I could not reproduce that, in any shape.

Built the real cross-assembly repro through the `gsc` CLI (`/target:library`, `/r:`), mirroring `ImportedNestedEnum_…`'s own harness — `glib` with `public struct Outer[T] { public enum Color { Red = 4, Green = 5, Blue = 6 } }`, consumed from a second assembly. **5 GS0156 sites on clean `main`, every one correct:**

```
'glib.Outer[T].Color'      (method type parameter)
'glib.Outer[U].Color'      (struct type parameter)
'glib.Outer[int32].Color'  (concrete)
'glib.Outer[int32].Color' -> 'glib.Outer[string].Color'  (cross-construction)
'glib.Outer[B].Color'      (through List[Outer[B].Color])
```

Zero `object` leaks. **Positive control** (a zero can be scope, not absence): inverting the condition at `SymbolDisplay.cs:890` (`FormatImportedType`, `!imported.TypeArguments.IsDefaultOrEmpty` → `imported.TypeArguments.IsDefaultOrEmpty` — condition inversion, not `if(false)`) forces the erased fallback and reproduces **#2947's string verbatim**:

```
error GS0156: Cannot convert type 'glib.Outer[object].Color' to 'int32'.
```

So the probe demonstrably **can** see the bug; its absence on `main` is genuine. #2947 is titled *"GS0156 **displays** erased nested-enum generic arguments"* and display is now correct — **#2947 looks fixed and closable.** Only 5 commits landed after #2912 (`#3053`, `#3064`, `#3062`, `#2998`, `#3056`), none touching this path, so the fix is #2912 itself.

**Disclosure:** I did **not** build #2912's own head `1b6ff1ca0` to confirm it reproduced *there*. My claim is scoped to current `main`.

### (d) `Fixes` lines point safely
`closingIssuesReferences` = **#2946 only**, and #2946 is already **CLOSED/COMPLETED** (by #2912). So merging cannot close a live bug — the failure mode to fear here is absent. The linkage is redundant rather than wrong; the body says so honestly ("PR #2912 merged and closed #2946 while this branch was in progress"). #2947 is correctly **not** referenced.

### Gates
Touches **one test file**; no production change, so `cs2gs` corpus is unaffected by construction. Docs gate on the **merged** tree: `docs/diagnostics.md` **383**, `website/docs/ref/diagnostics.md` **381** — identical to `main`, neither file touched. Merged tree class run **10/10**. Restore round-trip exact: `130fa801c…` → mutants → **`130fa801c…`**, `git status --porcelain` empty.

---

### Verdict: **MERGEABLE**

No BLOCKER. No production change, no wrongly-closed issue, and the strengthening closes a genuine all-zero blindness that OLD provably survived.

- **SHOULD-FIX (body, `#2947` paragraph):** the claim that #2947 is still broken is **falsified on `main`** by a probe with a working positive control. Leaving it in keeps a fixed issue open. Recommend deleting the sentence and closing #2947 separately.
- **NIT:** `:95-96` pin only `Members[0]`; **M-1** corrupts `Green`/`Blue` constants and the test passes.
- **NIT:** `:107`, `:113-114`, `:125` and the whole new `Iterate` block `:127-130` are **unproven** — every killing mutant short-circuits at `:95`, and the only mutant reaching past it leaves them green. The body should not credit them as pinning until a mutant kills through them.

DavidObando:
Post-rebase verification on head `8fa9290cd20a7d40c2a5fba397fe107f05fcd885`:

- Clean Core CI build: rc=0; artifact mtime followed build start; `out/bin/Release/Core/GSharp.Core.dll` md5 `063197684e63ad8d6cfc029d31c145b2`.
- Product mutant (`TypeDefEmitter` emits `0` for enum constants): build rc=0; fresh md5 `afe03780a44ee82291a476e60596139c`; focused suite failed 3/10, including `NestedEnumConstructedThroughGenericOuter_LoadsWithDistinctClosedSignatures` (`expected 11`, `actual 0`).
- Restored build: rc=0; fresh md5 returned exactly to `063197684e63ad8d6cfc029d31c145b2`.
- Full solution suite: 14,207 passed, 0 failed, 1 skipped.

The PR merged as `185431c4a5082ef4d7650edde803779cc3306b56`; the regression-test file blob matches the verified head exactly.

Issue #3067, opened 2026-08-02
Generic iterator state-machine constraint encodes method parameter in class scope


DavidObando:
## Summary

A generic iterator whose second type parameter has a constraint referencing its first emits an invalid state-machine class. `gsc` exits 0, but the emitted program aborts with `BadImageFormatException`.

This surfaced while covering the iterator context for #2894. It is independent of the lambda/closure defect: the minimal repro contains no lambda.

## Repro

```gsharp
package ConstrainedIteratorControl

import System

open class Box[T any] {}
class IntBox : Box[int32] {}

func Values[T any, TBox Box[T] init()](value T) sequence[T] {
    yield value
}

func Main() int32 {
    for value in Values[int32, IntBox](66) {
        Console.WriteLine(value)
        return value == 66 ? 0 : 2
    }
    return 3
}
```

Measured on `1ea847ebe`:

```text
gsc rc=0
runtime rc=134
Unhandled exception. System.BadImageFormatException: An attempt was made to load a program with an incorrect format.
   at ConstrainedIteratorControl.<Program>.Values[T,TBox](T value)
   at ConstrainedIteratorControl.<Program>.Main()
```

## IL evidence

The synthesized state-machine class constraint is emitted as:

```il
.class nested private ... '<Values>d__2`2'<T, .ctor (class ConstrainedIteratorControl.Box`1<!!0>) TBox>
```

The class GenericParamConstraint is in type scope and must reference its first class type parameter as `!0`, not method type parameter `!!0`.

This is distinct from #3065: no generic method-spec cache collision is needed to reproduce it; the invalid token is in the state-machine class's generic constraint metadata.

Issue #2894, opened 2026-07-30
Lambda body-only generic parameter is omitted from the synthesized closure


obselate:
Tested: Gsharp.NET.Sdk v0.3.362
Also reproduced with gsc origin/main commit 648f87606aab10b7a6d4fd6f558f10712a25b792.

A stored lambda in a generic method emits invalid IL when one method type parameter is used only inside the lambda body:

```gs
package FindingGenericCapturedLambda

open class Box {
}

open class Box[T any] : Box {
  var Value T = default(T)
}

class IntBox : Box[int32] {
}

class Descriptor {
  var Factory (() -> Box)?
}

func Make[T any, TBox Box[T] init()](input T) Descriptor {
  return Descriptor{
    Factory: func() Box {
      let value = TBox()
      value.Value = input
      return value
    },
  }
}

func Main() int32 {
  let descriptor = Make[int32, IntBox](7)
  guard let factory = descriptor.Factory else { return 1 }
  let value = factory() as IntBox
  return value.Value == 7 ? 0 : 2
}
```

Expected: the build succeeds and `Main` returns `0` after the closure constructs `IntBox` and assigns `Value`.

Actual: the build succeeds, but execution throws `BadImageFormatException` in the synthesized closure `Invoke` method.

ILSpy shows that the closure has only `<T>`, while `Invoke` uses undeclared `!!1` for `TBox`. Equivalent C# carries both `<T,TBox>` on its display class and returns `0`.

This appears to be a remaining case of #1477. `TBox` is referenced by the lambda body, but not by its capture list, parameters, or return type.

PR #3068, opened 2026-08-02
Fix MethodSpec cache remap identity

### LINKED ISSUE #3065: MethodSpecSymbolKey omits generic-remap scopes from its cache key, emitting an invalid assembly
## Summary

`MethodSpecSymbolKey` omits the active generic-remap scopes from its cache key, exactly as `CtorRefSymbolKey` did before #3057 (issue #2930). The result is that a cached `MethodSpec` token is reused across two different remap scopes, and the emitted assembly is invalid.

#3057 fixes the **constructor** cache completely and correctly. This issue is the sibling **method** site, which that PR does not touch.

## Repro

```gsharp
package MsProbe
import System

func Remap[A, B](first A, value B) B {
    var outer = Array.Empty[B]()
    let f (B) -> B = (innerValue B) -> {
        var inner = Array.Empty[B]()
        return innerValue
    }
    return f(value)
}

Console.WriteLine(Remap[int32, int32](0, 11))
Console.WriteLine(Remap[int32, int32](0, 22))
```

Run against a tree that already contains #3057:

```
$ gsc msprobe.gs
rc=0

$ ./msprobe
System.BadImageFormatException ... at MsProbe.<Program>.<lambda1>[B](B innerValue)
rc=134
```

The compiler reports success and emits an assembly the runtime refuses to load.

## Proven by construction, not by resemblance

Applying the *same* four-field key fix to `MethodSpecSymbolKey` and its two call sites — and changing nothing else — turns the above into:

```
rc=0
11
22
```

So this is the same root cause as #2930, at a second site.

## Shape of the fix

`CtorRefSymbolKey` now includes both active generic-remap scopes in its key, discriminated by `ReferenceEquals` on the remap dictionary objects. That is sound because `PushSmRemap` / `PushLambdaMethodRemap` (`GenericRemapState.cs:148-183`) install a distinct dictionary object per state-machine class and per lambda function rather than mutating a shared buffer; `GenericRemapState.cs:126` already documents object identity as the intended discriminator. A stale-identity miss fails safe (cache miss then recompute).

This is house style, not invention: nine pre-existing remap-keyed caches already follow it — `UserTokenResolver.cs:812, 822, 832, 1198, 1285, 1319, 1613, 1833` and `ImportedMemberRefFactory.cs:2049`.

## Note on the sibling call sites

`ImportedMemberRefFactory.cs:870` and `:909` are the two construction sites. Widening the constructor to take the remaps makes any missed site a `CS7036` at build time, so the compiler will enumerate them for you rather than relying on grep.

## Also worth doing here

`MetadataTokenCache.cs:458-461` currently says `CtorRefSymbolKey` is the "Counterpart to `MethodSpecSymbolKey`" while also documenting that it includes the remap scopes. After #3057 those two structs are **no longer** counterparts, and someone auditing for precisely this bug class would read that sentence as an assurance that the sibling is equivalently keyed. Fixing this issue makes the sentence true again.

## Notes

Surfaced during the rubber-duck review of #3057. The reviewer verified the repro against the merged (fixed) tree, so this is not a pre-existing symptom of #2930.

DavidObando:
Fixes #3065

## Summary
- Include the active state-machine and lambda-method remap dictionary identities in `MethodSpecSymbolKey`.
- Pass both remap scopes at every compiler-enumerated construction site.
- Add direct key-equality, emitted-assembly load/execute, and MethodSpec-row reuse regressions.

`ReferenceEquals` matches the existing remap-keyed cache pattern: each scope installs a distinct dictionary object, so stale identity only causes a safe cache miss.

## Repro

| Tree | `gsc` rc | runtime rc | Result |
|---|---:|---:|---|
| before | 0 | 134 | `BadImageFormatException` at `<lambda1>[B]` |
| after | 0 | 0 | `11` / `22` |

The after probe used a fresh full CIB solution build; the Core and Compiler copies of `GSharp.Core.dll` had the same build timestamp. The regression test also uses `Assembly.Load`, `GetTypes()`, and entry-point invocation.

## Construction-site denominator

Widening the key constructor first made the compiler enumerate **2 unique sites** (reported twice each because Core is built twice): the MethodSpec cache lookup and insertion in `ImportedMemberRefFactory`.

## Product mutations

| Mutant | Result | Killing test/evidence |
|---|---|---|
| Lookup site passes `null, null` | KILLED | `RepeatedMethodSpecInSeparateRemapScopes_EmitsValidAssembly`: emitted program raises `BadImageFormatException` |
| Insertion site passes `null, null` | KILLED | `RepeatedArrayEmptyInSameRemapScope_ReusesMethodSpecRow`: expected 2 MethodSpec rows, found 3 |
| Key constructor discards both remaps | KILLED (3/3) | Runtime, equality, and row-reuse regressions |
| Remove the two `ReferenceEquals` equality clauses | KILLED | `MethodSpecSymbolKey_DifferentRemapScopes_AreNotEqual` |
| Remove remap hash mixing | SURVIVED (expected) | Equality still preserves correctness; hash mixing affects distribution only |

## Validation
- Full suite: **14,214 passed, 0 failed, 14,215 total** (1 skipped).
- Focused MethodSpec regressions: **3/3 passed**.
- Release CIB solution build: **0 warnings, 0 errors**.
- cs2gs gate: **20 apps examined**, 19 green; positive-control known gap reproduced; **1 known, 0 new, 0 regressed, 0 stale**, rc 0.
- Bidirectional diagnostics gate: **377 descriptors + 7 allowlist = 384 expected**; both files have 0 missing and 0 unexpected IDs.
- `git merge-tree`: rc 0.

### DIFF
--- a/src/Core/CodeAnalysis/Emit/ImportedMemberRefFactory.cs
+++ b/src/Core/CodeAnalysis/Emit/ImportedMemberRefFactory.cs
@@ -876,7 +876,11 @@ internal sealed class ImportedMemberRefFactory
         }
         else
         {
-            var symbolKey = new MetadataTokenCache.MethodSpecSymbolKey(method, typeArgSymbols);
+            var symbolKey = new MetadataTokenCache.MethodSpecSymbolKey(
+                method,
+                typeArgSymbols,
+                this.remaps.ActiveIteratorStateMachineRemap,
+                this.remaps.ActiveLambdaMethodTypeParamRemap);
             if (this.cache.MethodSpecsWithSymbolArgs.TryGetValue(symbolKey, out var existingSym))
             {
                 return existingSym;
@@ -915,7 +919,12 @@ internal sealed class ImportedMemberRefFactory
         }
         else
         {
-            this.cache.MethodSpecsWithSymbolArgs[new MetadataTokenCache.MethodSpecSymbolKey(method, typeArgSymbols)] = spec;
+            this.cache.MethodSpecsWithSymbolArgs[
+                new MetadataTokenCache.MethodSpecSymbolKey(
+                    method,
+                    typeArgSymbols,
+                    this.remaps.ActiveIteratorStateMachineRemap,
+                    this.remaps.ActiveLambdaMethodTypeParamRemap)] = spec;
         }
 
         return spec;
--- a/src/Core/CodeAnalysis/Emit/MetadataTokenCache.cs
+++ b/src/Core/CodeAnalysis/Emit/MetadataTokenCache.cs
@@ -400,21 +400,28 @@ internal sealed class MetadataTokenCache
         = new Dictionary<StructSymbol, MethodDefinitionHandle>();
 
     /// <summary>
-    /// Issue #420 (P3-7): structural cache key for MethodSpec rows whose
-    /// generic arguments include user-defined type symbols. Uses reference
-    /// equality on the contained <see cref="TypeSymbol"/> entries (declared
-    /// user types are interned per compilation), combined with structural
-    /// equality on the array.
+    /// Issue #420 / #3065: structural cache key for MethodSpec rows whose
+    /// generic arguments include user-defined type symbols. Includes both
+    /// active generic-remap scopes because the same symbols can encode to
+    /// different VAR/MVAR ordinals. Uses reference equality throughout.
     /// </summary>
     internal readonly struct MethodSpecSymbolKey : IEquatable<MethodSpecSymbolKey>
     {
         private readonly MethodInfo method;
         private readonly ImmutableArray<TypeSymbol> typeArgs;
+        private readonly object classRemap;
+        private readonly object methodRemap;
 
-        public MethodSpecSymbolKey(MethodInfo method, ImmutableArray<TypeSymbol> typeArgs)
+        public MethodSpecSymbolKey(
+            MethodInfo method,
+            ImmutableArray<TypeSymbol> typeArgs,
+            object classRemap,
+            object methodRemap)
         {
             this.method = method;
             this.typeArgs = typeArgs.IsDefault ? ImmutableArray<TypeSymbol>.Empty : typeArgs;
+            this.classRemap = classRemap;
+            this.methodRemap = methodRemap;
         }
 
         public bool Equals(MethodSpecSymbolKey other)
@@ -424,7 +431,9 @@ internal sealed class MetadataTokenCache
                 return false;
             }
 
-            if (this.typeArgs.Length != other.typeArgs.Length)
+            if (!ReferenceEquals(this.classRemap, other.classRemap)
+                || !ReferenceEquals(this.methodRemap, other.methodRemap)
+                || this.typeArgs.Length != other.typeArgs.Length)
             {
                 return false;
             }
@@ -445,6 +454,8 @@ internal sealed class MetadataTokenCache
         public override int GetHashCode()
         {
             var hash = RuntimeHelpers.GetHashCode(this.method);
+            hash = unchecked((hash * 31) + RuntimeHelpers.GetHashCode(this.classRemap));
+            hash = unchecked((hash * 31) + RuntimeHelpers.GetHashCode(this.methodRemap));
             for (var i = 0; i < this.typeArgs.Length; i++)
             {
                 hash = unchecked((hash * 31) + RuntimeHelpers.GetHashCode(this.typeArgs[i]));
--- a/test/Core.Tests/CodeAnalysis/Emit/Issue3065MethodSpecRemapKeyTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Emit/Issue3065MethodSpecRemapKeyTests.cs
@@ -1,0 +1,153 @@
+// <copyright file="Issue3065MethodSpecRemapKeyTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Metadata;
+using System.Reflection.Metadata.Ecma335;
+using System.Reflection.PortableExecutable;
+using GSharp.Core.CodeAnalysis.Compilation;
+using GSharp.Core.CodeAnalysis.Emit;
+using GSharp.Core.CodeAnalysis.Symbols;
+using GSharp.Core.CodeAnalysis.Syntax;
+using GSharp.Core.CodeAnalysis.Text;
+using Xunit;
+
+namespace GSharp.Core.Tests.CodeAnalysis.Emit;
+
+public class Issue3065MethodSpecRemapKeyTests
+{
+    private const string Source = """
+        package MsProbe
+        import System
+
+        func Remap[A, B](first A, value B) B {
+            var outer = Array.Empty[B]()
+            let f (B) -> B = (innerValue B) -> {
+                var inner = Array.Empty[B]()
+                return innerValue
+            }
+            return f(value)
+        }
+
+        Console.WriteLine(Remap[int32, int32](0, 11))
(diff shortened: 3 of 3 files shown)


DavidObando:
## Review: MERGEABLE — no blockers

Measured on the **merged** tree: `main` **`f1656079`** + head **`a954c2328`** (merge base `185431c4`, `git merge-tree` rc=0, merged commit `fdc55289`). Delta vs `main` is exactly **3 files, +182/−9**, matching the body. All numbers below are true of that SHA pair only.

### The defect is the one this PR names, and it is live on `main` today

I filed #3065 from a `gsc`/`gsi` differential while reviewing #3057. Re-measured now with a two-sided control on the same driver binary:

| tree | `gsc` emit | run | output |
|---|---|---|---|
| base `f1656079` | rc0 | **rc134 `BadImageFormatException` at `<lambda1>[B]`** | — |
| merged `fdc55289` | rc0 | rc0 | `11` / `22` |

Same repro, same frame, same exception as my original report. This is that defect, not a neighbour, and the fix closes it.

Build hygiene: full-solution Release CIB builds only; `Core/` and `Compiler/` copies verified same-second fresh before every probe; each probe compiled into its own empty output directory. Core hash round-tripped two-sided `2b0f67bb` → base `f5507dce` → `2b0f67bb`.

### 1(a) — the `ReferenceEquals` argument holds, and here is why

The load-bearing claim is "each scope installs a distinct dictionary object". I checked the producers rather than the consumers, because consumer sets decay:

- All **6** registration producers (`ReflectionMetadataEmitter.cs:498, 572, 657, 1314, 1327`; `UserTokenResolver.cs:181`) allocate a **fresh** `new Dictionary` immediately before registering.
- Both loop sites call `BuildRemap()` **inside** the loop, so no instance is hoisted across iterations.
- `BuildRemap()` (`StateMachineEmitter.cs:301`, `:345`) allocates fresh per call and returns **`null`** when the remap is empty.

That last point is the specific thing worth calling out: **there is no shared empty-dictionary singleton.** "No active remap" is represented by `null`, not by a pooled empty instance — which is exactly the failure mode that would have reintroduced the bug. No pooling, no clear-and-refill anywhere. `GenericRemapState.cs:126` already documents object identity as the intended discriminator. The argument is sound.

### 1(b) — a miss is genuinely safe; I measured it rather than reasoning about it

Their insertion-site mutant produces "expected 2, found 3", which proves extra rows are *observable*. Observable is not the same as wrong, so I checked what the extra row does: under mutant M-B the **runtime test still passes** — the assembly loads and prints `11`/`22` — while only the row-count test fails. An extra MethodSpec row is **wasteful, not incorrect**. The cache is a size optimisation over a correct encoder, so a miss re-encodes correctly. Confirmed.

### 2 — twin comparison against `CtorRefSymbolKey`: identical, no repeated defect

Normalised extraction of both structs (names elided) diffs to **IDENTICAL** — same equality ordering, same hash mixing, same nullability handling. No gratuitous divergence and no defect repeated from the #2912/#3057 fix.

One asymmetry, now running the *other* way: #3068 adds `MethodSpecSymbolKey_DifferentRemapScopes_AreNotEqual`, which is precisely the instrument I recommended on #3057 — and **`CtorRefSymbolKey` still has no equivalent**. My #3057 SHOULD-FIX 2 therefore remains open, against the twin. Not this PR's job; flagging so it is not lost.

### 3 — the SURVIVED hash mutant is correct. Endorsing it explicitly.

I reproduced it: removing the remap hash mixing survives the **full `Core.Tests` suite, 7072 passed / 0 failed / 7073 total** (1 skipped).

Checked the right way, not by plausibility. The key is used in exactly **one** plain `Dictionary` with the default comparer — one `TryGetValue`, one indexer set. I grepped every `GetHashCode` use on this key and got **9 matching lines** (so the check has a live positive control, it is not a vacuous zero); none is an identity shortcut, an ordering key, or a serialised value.

**This is distribution-only. Equality is what preserves correctness.** Ignoring the remaps in the hash means unequal keys land in one bucket — a bucket-distribution cost on a small per-compilation cache, not a correctness bug. **Do not "fix" this later**; adding mixing is harmless but pinning it with a test would pin an implementation detail.

Worth recording: M-C (both `ReferenceEquals` equality clauses removed) is killed **only** by the direct equality test — the runtime and row-reuse tests both survive it. That is the same `GetHashCode` masking I hit on #3057, where the equality half was unobservable. The author's decision to add a direct key-equality unit test is what closes it, and it is load-bearing.

### 4 — denominator confirmed at 2 sites; the parenthetical's arithmetic is off

I made the compiler count rather than grepping, since `ImportedMemberRefFactory` and friends are partial. Added a **required 5th parameter** to the key constructor and read the `CS7036` output:

- **8 raw `CS7036`**, deduping to **exactly 2 unique `file:line`**: `ImportedMemberRefFactory.cs:879` (lookup) and `:923` (insertion).

The **conclusion is right — 2 sites, both fixed** — but the body says "reported twice each because Core is built twice", which would predict 4. It is 8, i.e. four renders per site. Minor, but the reasoning offered for the multiplier is not what the build actually does. **NIT.**

### Mutation matrix (independent mutants, no #3028 masking)

Every mutant `BUILD_RC=0` with a distinct Core md5; each restored and round-tripped to the clean hash.

| mutant | change | result |
|---|---|---|
| M-A | lookup `:879` → `null, null` | **KILLED** 2/3 (runtime + row-reuse) |
| M-B | insertion `:923` → `null, null` | **KILLED** 1/3 (row-reuse only) |
| M-C | both `ReferenceEquals` clauses removed | **KILLED** 1/3 (equality test only) |
| M-D | remap hash mixing removed | **SURVIVED** — correct, see §3 |

M-A and M-B were mutated **independently**, so neither hides the other.

**Endorsement:** the row-count assertion in `GenericLambdaMethodSpecs_RepeatedCallUsesCachedRow` (`Issue3065MethodSpecRemapKeyTests.cs:123`) is the **sole pin** for the insertion site — M-B kills nothing else. It is stricter than correctness strictly requires (§1(b): extra rows are merely wasteful), and it should stay exactly that strict. If a future change makes it fail, the correct response is to investigate the extra row, not to relax the count.

### 5 — overlap: all three disclaimers verified

- **#3067** — its repro still gives `gsc` rc0 → **rc134** on the merged tree. Not silently fixed, not half-fixed. Disclaimer holds.
- **#3069 / #2894** — **verified with the authoritative repro from the issue**, not a paraphrase:

  | tree | `func Main()` | top-level |
  |---|---|---|
  | base `f1656079` | **rc134** | **rc134** |
  | merged `fdc55289` | **rc134** | **rc134** |

  Unchanged in both columns. #3069's "No `MethodSpec` cache change is involved" is **correct** — #3068 closes nothing it does not declare, despite both surfacing as `BadImageFormatException` at a `<lambda…>[B]` frame. Two distinct defects.

  Two notes for #3069's author, free of charge: the defect **also reproduces at top level**, an axis the issue's `func Main()`-only repro does not cover; and `gsi` cannot fill that matrix at all for this shape — it returns `GS9999: Construction of type parameter 'TBox()' is only supported by the compiled backend (issue #988)`. The `gsi` column there is vacuous by design, so don't read a green in it as a pass.
- **#3036** — touches `GetElementTypeToken` (`ImportedMemberRefFactory.cs:91`). #3068's hunks are at `:876` and `:915`, ~700 lines away. **No overlap.**

### Docs gate

0 files under `docs/` or `website/`, **0** descriptor lines added. The zero is absence, not scope: positive control shows the gate regex matching **383** existing rows. Gate provably unaffected.

### Summary

- **BLOCKER:** none.
- **SHOULD-FIX:** none.
- **NIT** — body, denominator paragraph: "reported twice each because Core is built twice" predicts 4 renders; measured **8** (4 per site). Deduped count of **2** is correct; only the explanation of the multiplier is wrong.

The correctness argument is sound where it matters, the twin was checked rather than assumed, and the one survived mutant is survived for the right reason. Good fix.

*Disclosure: my first attempt at a #2894 probe was hand-built and passed on **both** base and merged — a null probe proving nothing. I discarded it and used the issue's authoritative repro instead. Reporting it because a null probe read as "verified" is exactly how a fabricated disclaimer-check gets filed.*

PR #3069, opened 2026-08-02
Fix lambda body-only generic closure parameters

### LINKED ISSUE #2894: Lambda body-only generic parameter is omitted from the synthesized closure

DavidObando:
## Summary

- collect enclosing type parameters referenced only from a capturing lambda body before synthesizing its display class
- propagate nested ordinary-lambda body references to the outer closure that materializes the nested constructed display class
- cover generic function, generic class, nested lambda, generic local-function, iterator, async, and interpreter contexts
- inspect emitted closure generic parameters with `Assembly.Load(File.ReadAllBytes(...)).GetTypes()` and execute every compiler probe in a bounded child process

Fixes #2894

## Exact-base reproduction

All probes were compiled with the exact `1ea847ebe` compiler. Compilation returned 0 for all seven; every emitted program then returned 134 with `BadImageFormatException`.

| Context | Base compile | Base runtime | Fixed runtime |
|---|---:|---:|---:|
| Generic function | 0 | 134, `BadImageFormatException` | 0, `11` |
| Generic class method | 0 | 134, `BadImageFormatException` | 0, `22` |
| Nested lambda | 0 | 134, `BadImageFormatException` | 0, `33` |
| Generic local function at top level | 0 | 134, `BadImageFormatException` | 0, `44` |
| Generic local function inside function | 0 | 134, `BadImageFormatException` | 0, `55` |
| Iterator/state machine | 0 | 134, `BadImageFormatException` | 0, `66` |
| Async body | 0 | 134, `BadImageFormatException` | 0, `77` |

The iterator probe uses an unconstrained body-only `TBox`; this isolates #2894 from the separate constrained state-machine metadata defect filed as #3067. No `MethodSpec` cache change is involved, distinguishing this fix from #3065.

## Driver × position matrix

| Position | `gsc` at base | `gsc` fixed | `gsi` at base | `gsi` fixed |
|---|---|---|---|---|
| Top level | rc 134, `BadImageFormatException` | rc 0, `99` | rc 0, `99` | rc 0, `99` |
| Inside function | rc 134, `BadImageFormatException` | rc 0, `111` | rc 0, `111` | rc 0, `111` |

## Call-site count

Adding a required parameter to `ClosureEmitter.SynthesizeDisplayClass` produced six rendered `CS7036` lines (errors plus build summary) at **3 unique call sites**: fieldless accessibility host, captured lambda closure, and `go` closure. The temporary parameter was removed after measurement.

## Product mutation

| Product hunk | Mutant | Result |
|---|---|---|
| Feed body type parameters into closure reification | Delete only those three lines | Killed: all 7 compiler rows fail in emitted programs with rc 134 and `BadImageFormatException` |
| Traverse nested ordinary lambda bodies | Invert `IsGeneric != true` to `IsGeneric == true` | Killed: nested-lambda row fails in emitted program with rc 134 and `BadImageFormatException`; other 6 rows pass |

Both mutations built successfully. Each clean artifact was rebuilt after restoration and matched its pre-mutation hash exactly.

## Validation

Merged tree equals the branch tree over `main` `185431c4` (`git merge-tree`), with merge base `185431c4`.

- Core.Tests: 7067 passed, 1 skipped, 0 failed
- Compiler.Tests: 4036 passed, 0 failed
- Interpreter.Tests: 595 passed, 0 failed
- Diagnostic documentation gate: 4 passed, 0 failed
- cs2gs corpus: 19/20 green; baseline 1 known, 0 new, 0 regressed, 0 stale; gate rc 0

### DIFF
--- a/src/Core/CodeAnalysis/Emit/ClosureEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/ClosureEmitter.cs
@@ -393,7 +393,10 @@ internal sealed class ClosureEmitter
         // capture site. Reify the display class generic over exactly the
         // referenced enclosing parameters (mirroring the state-machine
         // treatment); the emitter's outer-TP → own-TP remap then routes every
-        // field / Invoke signature through a valid `VAR(idx)` slot, and the
+        // field / Invoke signature and body through a valid `VAR(idx)` slot.
+        // Issue #2894 extends discovery to parameters referenced only by the
+        // body; they are otherwise absent from every signature while still
+        // requiring a valid slot in Invoke. The
         // returned constructed instance carries the original parameters as type
         // arguments for the capture-site `newobj`/field stores/delegate ctor.
         var enclosingRefSink = new List<TypeSymbol>();
@@ -408,6 +411,9 @@ internal sealed class ClosureEmitter
         }
 
         enclosingRefSink.Add(returnType);
+        var bodyTypeParameters = new List<TypeParameterSymbol>();
+        LambdaEnclosingTypeParameterCollector.Collect(body, bodyTypeParameters);
+        enclosingRefSink.AddRange(bodyTypeParameters);
         var origTPs = SynthesizedClosureReifier.CollectOrdered(enclosingRefSink);
         StructSymbol constructedClass = closureClass;
         if (!origTPs.IsDefaultOrEmpty)
--- a/src/Core/CodeAnalysis/Emit/LambdaEnclosingTypeParameterCollector.cs
+++ b/src/Core/CodeAnalysis/Emit/LambdaEnclosingTypeParameterCollector.cs
@@ -9,15 +9,16 @@ using GSharp.Core.CodeAnalysis.Symbols;
 namespace GSharp.Core.CodeAnalysis.Emit;
 
 /// <summary>
-/// Issue #2118: walks a non-capturing lambda body and gathers every
+/// Walks a lambda body and gathers every
 /// <see cref="TypeParameterSymbol"/> it references (through expression types,
 /// declared-local types, type arguments, <c>is</c>/pattern target types, and
-/// <c>typeof</c>/<c>sizeof</c> operands). The lambda itself declares no type
-/// parameters, so any type parameter reachable from its body is necessarily an
-/// enclosing one that must be promoted onto the synthesized static method as a
-/// generic method type parameter for the emitted IL to verify. Mirrors the node
-/// coverage of <c>LambdaBinder.EnclosingTypeParameterReferenceWalker</c> but
-/// accumulates ALL references instead of stopping at the first.
+/// <c>typeof</c>/<c>sizeof</c> operands). Issue #2118 uses the result to promote
+/// non-capturing lambdas to generic static methods; issue #2894 uses it to
+/// reify capturing display classes. Nested ordinary lambda bodies are included
+/// because the outer lambda materializes their constructed display classes.
+/// Mirrors the node coverage of
+/// <c>LambdaBinder.EnclosingTypeParameterReferenceWalker</c> but accumulates all
+/// references instead of stopping at the first.
 /// </summary>
 internal sealed class LambdaEnclosingTypeParameterCollector : BoundTreeWalker
 {
@@ -56,6 +57,9 @@ internal sealed class LambdaEnclosingTypeParameterCollector : BoundTreeWalker
         TypeSymbol.CollectReferencedTypeParameters(node.Type, this.sink);
         switch (node)
         {
+            case BoundFunctionLiteralExpression literal when literal.Function?.IsGeneric != true:
+                this.VisitStatement(literal.Body);
+                break;
             case BoundCallExpression call:
                 this.CheckTypeArguments(call.MethodTypeArguments);
                 break;
--- a/test/Compiler.Tests/Emit/Issue2894LambdaBodyGenericClosureTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2894LambdaBodyGenericClosureTests.cs
@@ -1,0 +1,333 @@
+// <copyright file="Issue2894LambdaBodyGenericClosureTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using GSharp.Compiler;
+using Xunit;
+
+namespace GSharp.Compiler.Tests.Emit;
+
+/// <summary>
+/// Issue #2894: type parameters referenced only from a lambda body must be
+/// reified onto its synthesized closure class.
+/// </summary>
+public class Issue2894LambdaBodyGenericClosureTests
+{
+    public static IEnumerable<object[]> Contexts()
+    {
+        yield return Case(
+            "generic-function",
+            "11",
+            1,
+            ["T", "TBox"],
+            """
+            package Issue2894.GenericFunction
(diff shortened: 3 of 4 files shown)


DavidObando:
## Adversarial review — **MERGEABLE**

Numbers true of `main` = **`f1656079`**. Merge base = **`185431c4`** (computed, three-dot). Merged probe commit `c151de8c`, tree `ed9bb306`. Co-merge with #3068 = tree `797c64c3`.

True content (three-dot): **4 files, +445/−8**; production is only **+18/−8 in 2 files**, the rest tests. (Two-dot shows 28 files/+504−1022 — artefact, ignore.)

---

### Both hunks pinned, independently, by PRODUCT mutation

Deletion mutants (not `if (true)` — CS0162 + warnings-as-errors kills the build). Clean merged `GSharp.Core.dll` md5 `fa53967f1ad32dcd1ad867750a513363`, `Core/` == `Compiler/` == `Repl/`, all fresh.

| mutant | md5 | author suite | my probes |
|---|---|---|---|
| **M1** — delete `ClosureEmitter.cs:414-416` | `7ebde83b…` | **0 passed / 7 failed** | `A0`,`B_default`,`F3_depth3` → rc134, arity collapse; `O2_none` unaffected |
| **M2** — delete `LambdaEnclosingTypeParameterCollector.cs:60-62` | `d8c0d9f3…` | **6 passed / 1 failed** (`nested-lambda`) | `F3_depth3` → rc134, outer two closures arity **0**; `A0`/`B_default`/`E_twoscope` still green |
| clean (restored) | `fa53967f…` ✔ exact round-trip, `git status` clean | 7/7 | all green |

Both md5s **reproduced identically** on re-application → deterministic. M2's selectivity proves hunk 2 is strictly the *nesting-recursion* arm and is **not redundant with hunk 1** — no smaller fix available by deletion.

### Arity checked in **both** directions (the brief's main worry)

Over-collection is the untested failure mode, and a green runtime probe cannot see it. Measured with an independent `Assembly.Load(File.ReadAllBytes(...))` + `GetTypes()` inspector:

| probe | shape | BASE | MERGED |
|---|---|---|---|
| `O1_only_one` | `[T,U]` declared, body uses only `T` | rc134, arity 0 | rc0, arity **1 `[T]`** — **no over-collection** |
| `O2_none` | generic method, no TP in body | **rc0**, arity 0 | rc0, arity 0 — **unperturbed** |
| `B_default` | TP only in `default(T)` | rc134, arity 0 | rc0, arity 1 `[T]` |
| `D_nestgen` | TP only in `List[T]` | rc134, arity 0 | rc0, arity 1 `[T]` |
| `C_typearg` | TP only as a type **argument** | rc134, arity 0 | rc0, arity 1 `[T]` |
| `E_twoscope` | class `VAR` **+** method `MVAR` | rc134, arity 0 | rc0, arity **2 `[T,U]`**, correct order |
| `F3_depth3` | TP only at nesting **depth 3** (author tests depth 2) | rc134, all 3 arity 0 | rc0, all 3 arity 1 `[T]` |

Six shapes the PR did not probe; all already correct. **No coverage gap found.**

### Credit where due — better than I expected

- `Issue2894LambdaBodyGenericClosureTests.cs:274-276` asserts `Assert.Equal(expected, closureType.GetGenericArguments().Select(p => p.Name))` — **exact sequence**, so arity *and* names *and* order. Over-collection **is** covered.
- `:271` asserts exact closure count; all 7 cases are ≥ 1 (1,1,2,1,1,1,1) — **no vacuous zero**.
- `:248-249` fresh GUID directory per case; `Run` at `:300-332` is a bounded 30 s child process with exit-code assertion. `IlVerifier` is correctly *not* relied on.
- The `{top-level, in-function}` axis **is** parameterised — cases `generic-local-function-top-level` and `generic-local-function-in-function`. No `func Main()` blind spot.

### #3068 disjointness — verified behaviourally, not just textually

Zero file overlap (positive control: self-intersection = 4). `git merge-tree` clean **both directions**, and a real merge reproduced the predicted tree `797c64c3` exactly. Built co-merged (`BUILD_RC=0`, md5 `554e0d3e`): **all 8 probes byte-identical** to #3069-alone; `Issue2894` **7/7** and `Issue3065` **3/3** green together. The shared `BadImageFormatException` at `<lambda…>[B]` is a shared *symptom*, not a shared cause. **Disclaimer (b) holds.**

---

## SHOULD-FIX

**1. The interpreter tests pin nothing — and the same sources under `gsc` would.**
`test/Interpreter.Tests/Issue2894LambdaBodyGenericClosureInterpreterTests.cs:22,47`. Both sources run **green at BASE** (`f1656079`, pre-fix): rc0 `99` and rc0 `111`. This is structural — #2894 is an emit-path defect and the interpreter never emits, so these cases *cannot* fail for it.

The sharp part: I ran those **exact two sources** through `gsc` on both trees —

```
it1 (top-level)    BASE cc=0 rc=134 BadImageFormatException, <closure_<lambda1>_1> arity=0
it1                MERG cc=0 rc=0  out=[99]  arity=1 [TBox]
it2 (in-function)  BASE cc=0 rc=134 BadImageFormatException, arity=0
it2                MERG cc=0 rc=0  out=[111] arity=1 [TBox]
```

So the file spends its 2 cases on the one driver that structurally cannot observe the bug, when the identical sources routed through the compiler are **genuine pins**. Recommend adding these two sources to the compiler suite (keep the interpreter cases as parity guards, but label them as such). Zero new shapes needed — the sources already exist.

**2. Every number in the PR body is measured at `1ea847ebe`, which is not even the merge base.** `1ea847ebe` is an ancestor of `185431c4` — 2 commits behind the merge base, 6 behind `main`. Stale by construction, before any reviewer reads it. Please restate against `f1656079`; I have re-measured everything above there, and nothing moved.

## NIT

**3.** The body's mutation (b) inverts `IsGeneric != true`. I used deletion instead — strictly stronger — and it killed exactly the `nested-lambda` row and nothing else, matching the claim. Consistent.

**4.** I attacked the `IsGeneric != true` guard as a suspected hole (a generic nested lambda would skip recursion). It is **correct by construction**: the language rejects that shape outright with `GS0468` — *"Local function 'G' cannot reference 'T', a type parameter of an enclosing method or class"* — so there is no reachable case behind the guard. Suspect verified before blaming; no change wanted.

## Answering the brief's item 1 directly

**`gsi` is decorative here, not load-bearing.** Green at BASE in both cells because the interpreter never emits metadata. The entire risk surface is `gsc` + emitted metadata, and that is where every pin above lives. A reader should not count the `gsi` column as corroboration.

PR #3070, opened 2026-08-02
Fix out-parameter analysis for member functions

### LINKED ISSUE #3058: Out-parameter definite assignment is skipped for functions declared in a class 'shared' block
## Summary

`GS0238` ("Out parameter must be definitely assigned on every path before the function returns") fires for a top-level `func`, but **is silently skipped** for a `func` declared inside a `shared { … }` block of a `class`. The program compiles with rc=0 and the `out` variable is never written.

## Repro

`top.gs` — correct behaviour:

```gsharp
func TryIt(x int32, out r int32) bool {
    return true
}

func main() {
    var v int32
    if TryIt(11, out v) {
        print("ok")
    }
}
```

```
$ gsc top2.gs
top2.gs(2,5,2,16): error GS0238: Out parameter 'r' must be definitely assigned on every path before the function returns.
        return true

Failed.
rc=1
```

`holder.gs` — the hole:

```gsharp
class Holder {
    shared {
        func TryIt(x int32, out r int32) bool {
            return true
        }
    }
}

func main() {
    var v int32
    if Holder.TryIt(11, out v) {
        print("ok")
    }
}
```

```
$ gsc holder2.gs
Success.
rc=0
```

Same function body, same `out` parameter, same never-assigned path. Only the enclosing declaration context differs.

## Why this matters

`out` definite assignment is the sole guarantee that a caller's `out` variable holds a real value after the call. Skipping the analysis for shared members means a caller reads whatever the default was, with no diagnostic. This is a silent-wrong-answer class, not a missing-warning class.

## Suspected cause

`RefKindDefiniteAssignmentAnalyzer.Analyze` has exactly **one** call site, compared with `RefStructAsyncLivenessAnalyzer`'s six. The single call site appears to be on the top-level function-binding path only, so functions bound through the class/`shared` member path never reach the analyzer.

Contrast with the sibling analyzer's six call sites, which is the shape a whole-program analysis normally has here.

## Expected

`gsc holder2.gs` should emit `GS0238` and exit 1, identically to the top-level case.

## Notes

Surfaced while reviewing PR #2979 (iterator exhaustive metadata ADR). That PR's ADR text asserted "out-param definite assignment runs unconditionally"; measuring that claim produced this repro. The ADR claim was corrected there; the underlying hole was not fixed and is filed here.

Related: iterators cannot have `out` parameters at all (`GS0422`), so this is specific to ordinary shared members.

DavidObando:
Fixes #3058

## Summary

- route all seven compiler-counted function-body binding sites through one post-lowering analysis hook
- analyze local functions and lambdas at their creation point, inheriting only definitely assigned captures
- preserve valid assignments through loops, protected returns, throws, try/catch, explicit defaults, and every reachable call shape
- add evaluate/emit/interpreter coverage for declaration contexts, Oahu regressions, invalid neighbors, and negative controls

## Declaration context matrix

Each cell is `gsc rc / gsi rc`; every head rejection emitted `GS0238`.

| Context | `main` `27a05798` | Head `96fc451a` |
|---|---:|---:|
| top-level function | 1 / 1 | 1 / 1 |
| extension function | 1 / 1 | 1 / 1 |
| class `shared` method | 0 / 0 | 1 / 1 |
| class instance method | 0 / 0 | 1 / 1 |
| struct instance method | 0 / 0 | 1 / 1 |
| data-struct instance method | 0 / 0 | 1 / 1 |
| default interface method | 0 / 0 | 1 / 1 |
| nested class method | 0 / 0 | 1 / 1 |
| local function | 0 / 0 | 1 / 1 |
| lambda | 0 / 0 | 1 / 1 |
| generic class method | 0 / 0 | 1 / 1 |
| struct `shared` method | 0 / 0 | 1 / 1 |
| constructor | 0 / 0 | 1 / 1 |
| property-getter lambda | 0 / 0 | 1 / 1 |
| interface-property local function | 0 / 0 | 1 / 1 |
| deinitializer lambda | 0 / 0 | 1 / 1 |

Valid controls cover `if/else`, `switch`, early return, loops, infinite loops, throw-only branches, protected returns, explicit `default(T)`, assigned captures, and user/imported/indirect/constructor/base/constrained/CLR calls. Invalid uninitialized and unassigned-capture neighbors still report `GS0239`. All run through evaluate, emit, and interpreter.

## Compiler-counted denominator

A required parameter on `AnalyzeFunctionBody` produced **7 unique CS7036 sites**:

`Binder.cs:1399, 1446, 1600, 1966, 2014, 2063, 2106`.

They cover top-level/extensions, primary member loops, constructors, interface methods, interface accessors, struct members, and shared/member helpers. Function literals recurse from their containing body. Iterators remain excluded by existing `GS0422`.

## Product mutation table

All mutants compiled and failed the named focused suite. Each restore rebuilt Core with CI metadata and round-tripped clean hash `082a54b5bff44c42ed48cbf75d457a47`; solution builds refreshed drivers before CLI probes.

| Product hunk mutated | Failing rows |
|---|---:|
| Binder top-level site | 24 / 402 |
| Binder primary-member site | 60 / 402 |
| Binder constructor site | 6 / 402 |
| Binder interface-method site | 60 / 402 |
| Binder interface-accessor site | 6 / 402 |
| Binder struct-member site | 12 / 402 |
| Binder struct-method helper site | 6 / 402 |
| shared ref-kind analysis call | 144 / 402 |
| preserved ref-struct analysis call | 10 / 30 |
| loop predecessor seeding | 12 / 402 |
| function-body throw exit | 6 / 402 |
| nested throw exit | 6 / 402 |
| protected-return exit label | 6 / 402 |
| unreachable try/catch merge | 6 / 402 |
| explicit-default initializer distinction | 6 / 402 |
| default ref-kind-array guard | 42 / 402 |
| ref/out write condition | 66 / 402 |
| imported instance call | 6 / 402 |
| user instance call | 6 / 402 |
| indirect delegate call | 6 / 402 |
| user constructor call | 6 / 402 |
| base-class call | 6 / 402 |
| imported static call | 6 / 402 |
| captured-assignment seed | 6 / 402 |
| opaque-statement literal discovery | 12 / 402 |
| expression literal discovery | 24 / 402 |
| captured-assignment filter | 12 / 402 |
| function-literal recursion | 36 / 402 |
| user static call | 18 / 402 |
| constrained static call | 6 / 402 |
| constructor chaining call | 6 / 402 |
| base-interface call | 6 / 402 |
| CLR constructor call | 6 / 402 |
| CLR conversion call | 6 / 402 |

Two speculative call arms survived and were deleted: synthesized CLR-static calls and function-pointer calls.

## Validation

Tested merged tree `31a929574fb2dcedf274da4c972a40cbcd1c960c`; `git merge-tree` is byte-identical to head.

- issue matrix: **402 passed / 0 failed**
- focused ref-kind/liveness: **46 passed / 0 failed**
- full Release suite: **15,013 passed / 0 failed / 1 skipped / 15,014 total**
- CI run `30792643817`: every non-skipped job passed
- pinned Oahu: **15 / 15 apps green** through translate, compile, IL verify, and test parity
- `gsc` and `gsi` agree on all 16 declaration probes
- no diagnostic descriptor or documentation change


DavidObando:
# Review: NOT MERGEABLE

**Anchors.** `main` **`f1656079`**; head **`cb9ccccc3`**; computed merge base `e8e6f7c8` (main is ahead of it). Three-dot CONTENT: 3 files, **+350 −31**. `git merge-tree` = **0 conflicts**, merged tree **`a1601796`** (my merge commit `d2a0f61f` reproduced that tree exactly). All numbers below are true of `main` `f1656079` + `cb9ccccc3`. Merged build `BUILD_RC=0`, 0 warnings, Core md5 `6166e7c8d713ed699dfdba14ab637142`, drivers fresh; restored round-trip returned to the same md5.

**The intent is right and the routing helper is a real improvement.** But the analyzer being switched on is materially incomplete, and turning it on in 12 contexts converts that incompleteness into a **source break on correct code**.

---

## BLOCKER 1 — Six correct-code shapes stop compiling. Measured `main` → head.

Every row is a program whose `out` parameter **is** definitely assigned on every path. `gsc <file>.gs`:

| probe | shape (context) | `main` | head |
|---|---|---:|---:|
| `a2_throw` | class `shared`: `if cond { r = 11 } else { throw Exception("no") }` | **rc=0** | **rc=1 GS0238** |
| `b2_dict` | class `shared`: `return d.TryGetValue(key, &r)` | **rc=0** | **rc=1 GS0238** |
| `a8_userinstance` | class instance: `return this.Inner(&r)` | **rc=0** | **rc=1 GS0238** |
| `e1_delegate` | class `shared`: `let f TryDelegate = Inner` … `return f(22, &r)` | **rc=0** | **rc=1 GS0238** |
| `e2_ctor` | class `shared`: `let w = Wrap(&r)` (`init(out r int32)`) | **rc=0** | **rc=1 GS0238** |
| `e3_base` | class instance: `return base.Fill(&r)` | **rc=0** | **rc=1 GS0238** |

Controls that behave correctly: `a6_passthrough` (free-function `Inner(&r)`) rc=0/rc=0; `b1_int32` (`Int32.TryParse(text,&r)`) rc=0/rc=0 — **the new imported hunk works for _static_ imported calls.**

**Root cause, compiler-counted rather than grepped.** `src/Core/CodeAnalysis/Binding/` declares **13** call-like bound node kinds:

```
BoundBaseClassCallExpression        BoundBaseInterfaceCallExpression
BoundCallExpression                 BoundClrConstructorCallExpression
BoundClrConversionCallExpression    BoundClrStaticCallExpression
BoundConstrainedStaticCallExpression BoundConstructorCallExpression
BoundFunctionPointerInvocationExpression BoundImportedCallExpression
BoundImportedInstanceCallExpression BoundIndirectCallExpression
BoundUserInstanceCallExpression
```

`RefKindDefiniteAssignmentAnalyzer.ProcessExpression` (`RefKindDefiniteAssignmentAnalyzer.cs:662-700`) switches on exactly **2** of them — `BoundCallExpression` and, new in this PR, `BoundImportedCallExpression`. **11 are unhandled**, and an `out`/`ref` argument passed through any of them is never added to `assigned`. The five I measured are `BoundUserInstanceCallExpression`, `BoundImportedInstanceCallExpression`, `BoundIndirectCallExpression`, `BoundConstructorCallExpression`, `BoundBaseClassCallExpression`.

`Dictionary.TryGetValue` is the single most common `out` idiom in .NET, and it is an **instance** call. The PR's own description says the imported hunk exists "preventing false GS0238 reports" — **it covers the static half only.**

## BLOCKER 2 — Branch terminated by `throw` is not treated as unreachable

`a2_throw`, separate root cause from Blocker 1: `if cond { r = 11 } else { throw Exception("no") }` then `return true`. `r` is assigned on every path that reaches the return. C# accepts this (no CS0177). Head rejects it. Any `out`-returning member using `throw` as the failure branch — an extremely common shape — stops compiling.

**Both blockers are pre-existing analyzer limitations, not new logic errors:** the same probes at **top level** (`a2t_toplevel`, `b2t_toplevel`, `b7_remove`) are **rc=1 on `main` too**, because top level already ran the analyzer. That does not reduce the severity of *this* PR — it is precisely what makes it dangerous. The PR takes an analyzer that is wrong on 2 of 13 call kinds and enables it across 12 additional contexts, so code that compiles today stops compiling.

## BLOCKER 3 — #3058's own defect survives in property accessors and deinitializers

The PR's denominator argument is that the two remaining async-liveness-only sites (`Binder.cs:2016` `BindInterfaceAccessorBody`, `Binder.cs:2070` `BindStructMemberBody`) bind members that "cannot declare `out` parameters". True of the **member** — false of the **analysis**, because a lambda inside such a body can, and `FunctionLiteralFinder` only runs from inside `RefKindDefiniteAssignmentAnalyzer.Analyze`, which those two sites never call.

Discriminating triple, identical lambda each time:

| probe | enclosing context | head |
|---|---|---:|
| `d4ctl_method` | class **method** | **rc=1 GS0238** |
| `d4_prop` | computed **property getter** | **rc=0 — silently accepted** |
| `d5_deinit` | **deinit** | **rc=0 — silently accepted** |

That is issue #3058's exact signature — same body, same unassigned `out`, different declaration context — still live after the fix.

---

## What is correct, and verified

- **Lambda scoping is right in both directions** (`c1_lambda_encl`): an enclosing `out` assigned **only inside a lambda** still reports GS0238 on head (rc=0 on `main`) — conservative and matching C#. `ProcessExpression` has no `BoundFunctionLiteralExpression` case, so the enclosing walk correctly does not descend. **No duplicate reporting** — exactly 1 occurrence in every probe.
- **The imported-out hunk is pinned.** Deleting `case BoundImportedCallExpression` (`BUILD_RC=0`, mutant md5 `6fd47e3141ec5d05adc10258b4515e64`) → **1 failed / 18**, `ClassSharedMethod_ImportedOutCallAssignsParameter_NoError`. Not dead code. Restored → md5 back to `6166e7c8…`.
- **No corpus or sample breakage today.** All 148 `samples/**/*.gs` compiled on both trees: **135 rc=0 / 13 rc=1 on `main`, identical on head, 0 regressions.** Reason measured, not assumed: `samples/` declares exactly **one** out-parameter function (`samples/TryParseOutVar.gs:11`) and it is **top-level**, already analyzed on `main`; the `cs2gs` corpus declares **zero** (positive control: 15 corpus lines contain `out ` at all). **So CI cannot see any of this** — the breakage is latent in user code.
- **Core.Tests on the merged tree: 7087 passed / 0 failed / 1 skipped / 7088**, `rc=0`. Green, with six correct programs rejected. "Full suite passes" is not evidence.
- **Driver-path axis clean:** `gsc` evaluate, `gsi`, and `gsc /out:` emit all agree on `a2_throw`, `b2_dict` (GS0238 on all three) and `d4_prop` (silent on all three). No axis hides a different answer.

## SHOULD-FIX

**S1 — `Binder.cs:1600` newly runs `RefStructAsyncLivenessAnalyzer` on constructors.** `AnalyzeFunctionBody` bundles both analyzers, and the ctor site previously ran **neither**. The out-parameter half is the intent; the ref-struct-async half is an unannounced second behavioural change. Believed inert (constructors are not async), but it is unstated and untested.

**S2 — the new tests never traverse the emit path.** `Issue3058OutParameterMemberDefiniteAssignmentTests.cs:266` `Compile()` uses `new Compilation(tree).Evaluate(...)`. I verified all three driver columns agree today, so nothing is being masked — but the pinning is evaluate-only.

## Suggested route

Either (a) extend `ProcessExpression` to the remaining call kinds — at minimum the five measured — and treat `throw`-terminated branches as unreachable, **then** enable the new contexts; or (b) land the `AnalyzeFunctionBody` refactor now and enable `RefKindDefiniteAssignmentAnalyzer` in the new contexts in a follow-up gated on that coverage. Blocker 3 additionally needs `Analyze` (not just the liveness analyzer) at `Binder.cs:2016` and `:2070`.

All probes are single files runnable as `gsc <file>.gs`; I can attach any of them.

DavidObando:
## Review verdict: NOT MERGEABLE — 1 BLOCKER, 2 SHOULD-FIX

Numbers true of `main` = `5ad0109c` (re-verify at `33bea4af`; `main` has moved).

The refactor is sound. **Enabling it now is not.**

### BLOCKER — six correct programs that compile on `main` are rejected

Each has its `out` parameter definitely assigned on every path:

| shape (declaration context) | `main` | this PR |
|---|---:|---:|
| class `shared`, `else { throw … }` | rc=0 | **rc=1 GS0238** |
| class `shared`, `return d.TryGetValue(key, &r)` | rc=0 | **rc=1 GS0238** |
| class instance, `return this.Inner(&r)` | rc=0 | **rc=1 GS0238** |
| class `shared`, delegate `f(22, &r)` | rc=0 | **rc=1 GS0238** |
| class `shared`, ctor `Wrap(&r)` | rc=0 | **rc=1 GS0238** |
| class instance, `base.Fill(&r)` | rc=0 | **rc=1 GS0238** |

**Root cause, compiler-counted rather than grepped:** `src/Core/CodeAnalysis/Binding/` declares **13** call-like bound node kinds. `ProcessExpression` (`RefKindDefiniteAssignmentAnalyzer.cs:662-700`) switches on **2** — `BoundCallExpression` plus the `BoundImportedCallExpression` this PR adds. **11 are unhandled**, including `BoundUserInstanceCallExpression`, `BoundImportedInstanceCallExpression`, `BoundIndirectCallExpression`, `BoundConstructorCallExpression` and `BoundBaseClassCallExpression`.

The imported-CLR hunk fixes the **static** half only. `Int32.TryParse`, `Double.TryParse` and `ThreadPool.GetAvailableThreads` all pass; **`Dictionary.TryGetValue` — the most common `out` idiom in .NET — fails**, because it is an *instance* call.

**Important framing correction, because it changes the remedy:** these are **pre-existing analyzer limitations being switched on**, not new logic errors. The same probes at top level are rc=1 on `main` too. So the fix is to **raise `ProcessExpression`'s call-kind coverage to all 13 kinds *before* the twelve contexts are enabled**, not to debug new code. Splitting that into a follow-up would ship a source break in the meantime.

`throw`-terminated branches need handling too — an `else { throw … }` arm must not demand assignment. Check for other diverging terminators while you are there.

### Why CI cannot see any of this — please do not read the green as evidence

- All 148 samples: **135/13 on `main`, identical here, 0 regressions.**
- Measured reason: `samples/` declares **one** out-parameter function (`samples/TryParseOutVar.gs:11`) and it is **top-level**, so it was already analysed on `main`. The `cs2gs` corpus declares **zero** — positive control: 15 corpus lines contain `out ` at all.
- **Core.Tests on the merged tree: 7087 passed / 0 failed / 7088.** Fully green while six correct programs stop compiling.

This is PR #2963's failure mode with the sign flipped, and #2963 shipped a source break through fully green CI after clearing review three times.

### SHOULD-FIX 1 — the fix is also incomplete in the other direction

A lambda in a class **method** gets `GS0238`; the identical lambda in a **property getter** is rc=0, and in a **deinit** is rc=0. The argument that "accessors cannot declare `out` parameters" is true of the *member* and false of the *analysis* — a lambda inside one can. `FunctionLiteralFinder` only runs from inside `Analyze`, which `Binder.cs:2016` and `Binder.cs:2070` never call. **#3058's exact signature survives this fix in accessor and deinit bodies.**

### SHOULD-FIX 2 — unannounced new analysis on constructors

`Binder.cs:1600` now runs `RefStructAsyncLivenessAnalyzer` on constructors, a site that previously ran **neither** analyzer. It appears inert, but it is untested and undeclared. Either test it or revert that part and say why.

### What would make this mergeable

A **false-positive corpus** as a parameterised test — every shape above × every declaration context, asserting rc=0 and zero diagnostics — plus a matching true-positive corpus asserting exactly one `GS0238`. Both parameterised over `{top-level, in-function}` × `{bare gsc evaluates, gsc /out: emits, gsi interprets}` as an **axis**, not as one-off cases. The current tests only traverse `Compilation.Evaluate` and never the emit path.

Only a product mutant counts for pinning — condition inversion, not `if (true)` (CS0162 plus warnings-as-errors kills the build, which yields no result rather than a survival).

DavidObando:
## Coordinator: `cs2gs-oahu` failure text, extracted from the CI artifact

The errors are in the artifact, not the log, which is why this has been opaque. From run `30738432786`, head `1852cb3a`, via `gh run download 30738432786 -n cs2gs-oahu-report`.

`main` (`abfab99c`) is **green** on this gate (run `30759057485`), so this PR is the cause.

### Verdict: **false positives. Correct programs are rejected that `main` accepts.**

5 fingerprints — 4 source sites plus 1 propagated duplicate. All the same diagnostic:

| Location | Verbatim error |
|---|---|
| `Mpeg4/Chunks/DashChunkEntryies.gs:118:5` | `GS0238: Out parameter 'firstMoof' must be definitely assigned on every path before the function returns.` |
| `Mpeg4/Chunks/InterleavedIterator[T].gs:52:13` | `GS0238: Out parameter 'minIndex' must be definitely assigned on every path before the function returns.` |
| `Mpeg4/Util/HelperExtensions.gs:8:9` | `GS0238: Out parameter 'checkedcount' must be definitely assigned on every path before the function returns.` |
| `ChapterQueue.gs:45:5` | `GS0238: Out parameter 'chapterEntry' must be definitely assigned on every path before the function returns.` |

### The rejected code, reduced from the corpus

```gsharp
minIndex = -1
...
return minIndex != -1
```

```gsharp
if failed {
    checkedcount = vecSize * i
    return false
}
checkedcount = vecSize * numVectors
return true
```

```gsharp
if chapterEntries.Count > 0 {
    chapterEntry = chapterEntries.Dequeue()
    return true
}
chapterEntry = nil
return false
```

`SkipToFirstMoof` assigns `firstSample` immediately, then assigns `firstMoof`/`firstMdat` in **both** `dataOffset` branches.

Every one of these **does** definitely assign on every path. These are the two most ordinary out-parameter shapes there are: unconditional-assign-then-return, and assign-in-both-arms-of-an-`if`.

### Causing hunk

`src/Core/CodeAnalysis/Binding/Binder.cs:1446` — routing **class members** through `AnalyzeFunctionBody`; the analyzer is invoked at `Binder.cs:2130`. The widening is correct in intent, but it exposes **pre-existing analyzer defects** to a much larger body of code, and they become **source breaks**.

### Correction to the earlier theory

An earlier review attributed this to `RefKindDefiniteAssignmentAnalyzer.cs:662-700` — `ProcessExpression` switching on 2 of 13 call-like bound node kinds. **The artifact does not support that.** None of these four sites involves an out-argument passed to a call. The residual shapes are **CFG handling** — plain `if`/`else` join points, loops, and early `return` — not call-kind coverage. Fix the CFG join/assignment tracking; the call-kind switch is a separate question.

### What this PR needs before it can land

A **false-positive corpus**: programs that compile on `abfab99c` and must still compile on the head. Seed it with the four shapes above, then vary `{assign-then-return, assign-in-both-if-arms, assign-in-loop, assign-before-early-return, assign-in-switch-arms, assign-in-try/finally}` × `{top-level func, class member, generic type member, extension, iterator}`.

A definite-assignment analyzer is a **false-positive machine**, and a positive-only suite is structurally blind to that class — this PR read **7087 passed / 0 failed** while rejecting four correct real-world constructs. Note also that `samples/TryParseOutVar.gs:11` is the **only** out-parameter function in `samples/`, and it is top-level, which is the measured reason the sample gate cannot see this break. `cs2gs-oahu` is the only gate that can.

DavidObando:
# Review: MERGEABLE

The three previous review rounds blocked this on a source break. **I rebuilt the false-positive corpus from scratch and it is gone.** Every shape those reviews named now compiles, the `cs2gs-oahu` gate went green on a byte-identical corpus, and the analyzer's call-kind coverage went from 2/13 to 11/13 with the two omissions justified by the type system rather than by "no test failed".

**Anchors.** `main` **`f7d9a5d3`**; head **`a62d9203`**; computed merge base `4469e323`. Three-dot `--full-index`: 3 files, **+1405 −113**. `git merge-tree` output (not rc): **1 line, 0 conflict markers**, predicted tree `acd6ddf9` — my merge commit `80b0c691` reproduced that tree **exactly**. Merged **`BUILD_RC=0`**, 0 warnings / 0 errors. Merged Core md5 `3b77ec3d8ed99d39458ea8f329855766`; base-only Core md5 `c9b3a9e465873d462f990dbc2b2bdbb8`. All 22 non-`ref` `GSharp.Core.dll` copies plus `Compiler/gsc.dll` and `Repl/gsi` verified fresh by **mtime**.

---

## 1. The false-positive corpus — the entire point of this review

15 correct programs + 3 true-positive controls, each run on **all three drivers** (bare `gsc` evaluates, `gsc /out:` emits, `gsi` interprets), compared on **rc and diagnostic set**, every emit into its own asserted-empty directory.

### The six shapes that blocked round 1 — all fixed

| probe | shape | base `f7d9a5d3` | head |
|---|---|---:|---:|
| `a2_throw` | class `shared`, `else { throw … }` | rc=0 | **rc=0** |
| `b2_dict` | class `shared`, `return d.TryGetValue(key, &r)` | rc=0 | **rc=0** |
| `a8_userinstance` | class instance, `return this.Inner(&r)` | rc=0 | **rc=0** |
| `e1_delegate` | class `shared`, delegate `f(44, &r)` | rc=0 | **rc=0** |
| `e2_ctor` | class `shared`, ctor `Wrap(&r)` | rc=0 | **rc=0** |
| `e3_base` | class instance, `base.Fill(&r)` | rc=0 | **rc=0** |

### The four real-world shapes from the `cs2gs-oahu` artifact — all fixed

`c1_assign_then_return` (`minIndex = -1` … `return minIndex != -1`), `c2_loop_early_return` (`checkedcount` in loop + early return), `c3_if_arms` (assign-in-both-arms + `Dequeue`), `c4_both_branches` (`SkipToFirstMoof` shape, two `out` params in both `if` arms) — **rc=0 on base and head, all three drivers.**

Plus `d1_switch_arms`, `d2_try_finally`, `d3_infinite_loop`, `d4_ifelse_chain`, `f1_clrstatic_nested`, `g4_protected_return`, `g5_protected_return_nested`, `h1_deconstruct_out`, `w1`–`w3` (void protected return, no-return void, switch-expression under `try`). **Zero false positives. Zero regressions.**

### True positives still fire — and two of them are round-1 defects now fixed

| probe | base | head |
|---|---:|---:|
| `tp1_member_unassigned` (class `shared`) | rc=0 | **rc=1 `GS0238`** |
| `tp2_prop_lambda` (lambda in property getter) | rc=0 | **rc=1 `GS0238`** |
| `tp3_deinit_lambda` (lambda in `deinit`) | rc=0 | **rc=1 `GS0238`** |

`tp2`/`tp3` are round-1 **BLOCKER 3** ("#3058's own defect survives in property accessors and deinitializers"). Both are closed. All three drivers agree; exactly one `GS0238` each.

### The PR fixes more than it claims

`g3_toplevel_nested_throw` — a `throw` in a nested `try` region inside a **top-level** function — is **rc=1 `GS0238` on base `f7d9a5d3`** and **rc=0 on head**. That is a pre-existing top-level false positive on `main` today, unrelated to the twelve new contexts, silently repaired here.

---

## 2. How `cs2gs-oahu` went green — the gate was **not** narrowed

Red run `30738432786` head `1852cb3a`; green run `30781375235` head `a62d9203`. Between those two trees:

- `tools/cs2gs/corpus/` — **0 files changed**
- `.github/` — **0 files changed**
- `build/` — **0 files changed**; `build/run-cs2gs-oahu.sh` md5 `8e03d719…` **identical**
- the gate corpus is **external and pinned** via `tools/cs2gs/external/oahu.json`; that pin is `DavidObando/Oahu` commit **`0ac1fece4415d31955ae7a1dcf7da31e343d363d`** at **both** heads

The PR itself touches 3 files, none under `tools/`, `build/` or `.github/`. The gate went green because the analyzer was fixed — independently corroborated by my own reproduction of all four artifact shapes above.

---

## 3. Call-like node kinds vs switch arms

**13** call-like kinds declared in `src/Core/CodeAnalysis/Binding/` (none `partial`, so grep is sound here; `Binder` is `sealed`, not partial). `ProcessExpression` (`RefKindDefiniteAssignmentAnalyzer.cs:717`) now handles **11 of 13**, up from 2 — plus a 12th arm at `:745` for `BoundConstructorChainingExpression`, which is call-like but outside the `Call|Invocation` naming pattern, so the honest denominator is 14 with 12 covered.

The two omissions are justified **structurally**, not by surviving mutation:

- **`BoundFunctionPointerInvocationExpression`** carries **no `RefKind` field at all**; its binder producer (`OverloadResolver.CallBinding.cs:848`) uses `fpSym.ParameterTypes[i]` only. An arm could never observe a non-`None` ref kind.
- **`BoundClrStaticCallExpression`** has `ArgumentRefKinds`, but the only producer that passes a non-default value is `TryBindImportedDeconstruct` (`StatementBinder.Narrowing.cs:1919`), where the `out` args are **synthetic `<deconstructN>` temps** immediately copied into user locals via `BoundVariableDeclaration`, with `RefKind.Out` (so `ReportVariableNotAssignedBeforeRef`, which only fires on `RefKind.Ref`, cannot trigger). Behavioural check `h1_deconstruct_out`: **rc=0**.

Deleting those two arms is the right call, and the reasoning above is what should be in the PR body instead of "two speculative call arms survived and were deleted" — surviving a mutant means *untested*, not *impossible*.

---

## 4. Mutation — per hunk, product mutants, asserted applied

Every mutant: condition inversion or value substitution (never `if (true)`), occurrence count asserted **before and after**, `BUILD_RC=0` confirmed, distinct Core md5 confirmed, and restored with an **exact md5 round-trip** to `3b77ec3d…`.

| # | hunk | mutant | md5 | result |
|---|---|---|---|---|
| M1 | throw-path discard, `RefKindDefiniteAssignmentAnalyzer.cs:248` | `== ThrowStatement` → `!= ThrowStatement` | `4e0bc665…` | **survived my first subset**; killed after I built the right probe |
| M2 | `goto methodExitLabel` recognition, `:253-255` | `ReferenceEquals(…)` → `!ReferenceEquals(…)` | `f20cd599…` | survived **all** my probes; **killed by the PR's own tests, 6 rows** |
| M3 | user-instance arm, `:750` | `call.Method.Parameters` → `ImmutableArray<ParameterSymbol>.Empty` | `e1171108…` | **killed** by `a8_userinstance` |

**M1 is worth reporting honestly.** It survived my initial 11-probe subset because the hunk is on the path explicitly commented *"Not the function body"* — my `a2_throw` never reached it. Writing a throw inside a nested `try` region killed it immediately: `g1_nested_throw_try` (class member) and `g3_toplevel_nested_throw` (top level) both flip rc=0 → rc=1 `GS0238` under M1 and are rc=0 clean.

**M2's dying test**, named: `TruePositiveCorpus_MissingAssignment_ReportsSingleGS0238(name: "OahuChapterQueueLock")` × `{TopLevel, InFunction}` × `{Evaluate, Emit, Interpreter}` = exactly 6 rows — matching the PR body's predicted "protected-return exit label: 6" precisely. I could not kill it myself, which is a real limitation of my corpus and a credit to theirs.

---

## 5. Suites on the **merged** tree at `f7d9a5d3` (unfiltered)

| project | result |
|---|---|
| Core.Tests | **7089 passed / 0 failed / 1 skipped / 7090** |
| Interpreter.Tests | **1125 passed / 0 failed / 1125** |
| Cs2Gs.Tests | **1894 passed / 0 failed / 1894** |
| Compiler.Tests | **4157 passed / 0 failed / 4157** (1 h 10 m) |

`Cs2Gs.Tests` matters specifically here: `tools/cs2gs/Cs2Gs.Tests/Cs2Gs.Tests.csproj` carries `<ProjectReference Include="..\..\..\src\Compiler\Compiler.csproj" ReferenceOutputAssembly="false" />` and invokes real `gsc.dll` out-of-process, which is the channel by which a pure `src/` change turns the cs2gs gates red. Green on the merged tree.

**Not run locally:** the `cs2gs-oahu` gate itself (external pinned corpus; it is green in CI on the head) and e2e.

---

## 6. Test design — the round-2 requirement was met

`test/Interpreter.Tests/Issue3058OutParameterMemberDefiniteAssignmentTests.cs:74` is `FalsePositiveCorpus_AssignedOnEveryPath_HasNoDiagnostics`, parameterised over `Position {TopLevel, InFunction}` × `Driver {Evaluate, Emit, Interpreter}` across 43 source cases, alongside `TruePositiveCorpus_…:87` and `RefReadCorpus_…:101`. That is the axis-not-cases structure round 2 asked for, and it closes round-1 **SHOULD-FIX 2** (the emit path is now traversed).

---

## Findings

### SHOULD-FIX 1 — `Binder.cs:1600`: constructors newly run a second, unannounced analysis
`AnalyzeFunctionBody` bundles both analyzers. The constructor site previously ran **neither**, so it gains `RefStructAsyncLivenessAnalyzer` as well as the intended `RefKindDefiniteAssignmentAnalyzer`. This was raised in round 1 and is still not mentioned in the PR body or the ADR. It appears inert, but "appears inert" is what the body should say explicitly rather than leave to the reader.

### NIT 1 — `FindMethodExitLabel` is a fragile positional match
`RefKindDefiniteAssignmentAnalyzer.cs:279-289` identifies the synthesized epilogue as *exactly* `Statements[^2] is BoundLabelStatement { Syntax: null }` and `Statements[^1] is BoundReturnStatement { Syntax: null }`. If the Lowerer ever emits one more trailing statement, this silently returns `null`, `goto`-to-exit stops counting as a function exit, and **every protected-return shape becomes a false positive again** — with no test failing at the point of the change, because the failure is in a different component. I probed `w1`–`w3` and `g4`/`g5` and found no live break, so this is not a defect today. Consider asserting the shape or tagging the label.

### NIT 2 — the `goto`-exit hunk rests on a single corpus case
M2 was killed only by `OahuChapterQueueLock`. One case across six rows is thin pinning for the hunk that repairs protected returns; a second, structurally different protected-return case would harden it.

---

## Separate pre-existing bug — **not** caused by this PR, reported for triage

`out`-parameter write-back is silently dropped for class **instance** methods in two of the three drivers, on **both** `f7d9a5d3` and this head:

```
v1_instance_out (class instance, r = 55)   eval=0    emit=55   gsi=0
v2_shared_out   (class shared,   r = 66)   eval=66   emit=66   gsi=66
v3_free_out     (free function,  r = 77)   eval=77   emit=77   gsi=77
```

rc=0, no diagnostic, wrong value — a silent wrong answer, and invisible to this PR's tests because they assert diagnostics rather than values. Identical on base and head, so it is out of scope here, but it is live on `main` today.

---

## Verdict

**MERGEABLE.** The source break is measurably gone in both directions, the gate that caught it went green without being weakened, the two uncovered call kinds are uncovered for sound structural reasons, and the test suite now has the false-positive axis this needed. SHOULD-FIX 1 is a documentation gap, not a defect.

DavidObando:
## Addendum — corrected test-project denominator (verdict unchanged: **MERGEABLE**)

Applying the fleet correction that the test-project denominator must be enumerated from `GSharp.sln`, never from `ls test/`, I re-measured my own §5 above.

Enumerated from `GSharp.sln` at live `main` (`5b71a74d`), there are **9** test projects:

```
test\Compiler.Tests          test\Core.Tests            test\Extensions.Tests
test\InternalAnalyzers.Tests test\Interpreter.Tests     test\LanguageServer.Tests
test\Sdk.Tests               tools\cs2gs\Cs2Gs.Tests    tools\gsgen\GSharp.GeneratorHost.Tests
```

I ran **4 of 9**. My original "Not run locally" line named only the `cs2gs-oahu` gate and e2e, which **implied the four projects I listed were the whole test surface**. They were not. Correcting explicitly:

**Not run (5 of 9):** `Extensions.Tests`, `InternalAnalyzers.Tests`, `LanguageServer.Tests`, `GSharp.GeneratorHost.Tests`, `Sdk.Tests`.

Of those, **`LanguageServer.Tests` is the one with plausible relevance** to this PR — it consumes binder diagnostics, and #3070 changes which binder diagnostics are produced. The other four are structurally remote from a bind-time analyzer change. `Sdk.Tests` is legitimately out of bounds: running it writes to the shared `~/.nuget/packages/gsharp.net.sdk/` cache.

I tried to close the gap by running the four permitted projects now rather than merely disclosing it, but load average is **42.68** against a threshold of 30, so I built and ran nothing. Disclosing that plainly beats implying coverage.

Incidentally, `ls test/` returns **11** entries in this tree, not 7 — so the directory listing is unreliable in *both* directions, which is a further argument for the `.sln` as the only denominator.

Verdict unchanged: **MERGEABLE**. The three headline findings rest on the two-tree differential corpus and the mutation results, not on suite counts.

---

## Resolving SHOULD-FIX 1 from the review above (`Binder.cs:1600`)

I graded the newly-live constructor analyzer site SHOULD-FIX and wrote that it "appears inert". That merged unresolved, so I settled it by measurement. **It is not inert — but it is a fix, not a break.**

`RefStructAsyncLivenessAnalyzer` can emit exactly **one** diagnostic, `GS0219` (`DiagnosticDescriptors.cs:136`), sole emitter `ReportIfUnsafe` (`RefStructAsyncLivenessAnalyzer.cs:541-565`). `Analyze` (`:73-88`) has two paths and only the first is async-gated; the second — `AsyncLambdaScopeFinder.VisitStatement(body)` — is **unconditional** and analyzes any nested async lambda. A constructor need not itself be async to reach it.

Two-tree differential, `27a05798` (pre) vs `d240eb24` (post), 11 probes × 3 drivers, both `BUILD_RC=0`:

| probe | pre | post |
|---|---|---|
| ctor async lambda, `ReadOnlySpan` live across `await` | rc=0 clean ×3 | **rc=1 `GS0219` ×3** |
| 9 correct ctor shapes (dead-before-await, multi-await, loop, try/finally, `struct init`, nested lambda, non-async lambda, bare span, plain async lambda) | rc=0 clean ×3 | rc=0 clean ×3 |
| positive control, same shape in a plain `func` | rc=1 `GS0219` ×3 | rc=1 `GS0219` ×3 |

**Zero false positives**, so this is not a source break. The positive control fires on *both* trees, so the pre-tree constructor silence is absence, not scope.

The pre-#3070 hole was worse than permissive. Made deterministic with `.GetAwaiter().GetResult()`:

| | pre `27a05798` | post `d240eb24` |
|---|---|---|
| span live across `await` | compiles rc=0, prints **`0`** — wrong, must be `3` | rc=1 `GS0219` |
| safe control | prints `3` | prints `3` |

So before this PR, a constructor-hosted async lambda holding a `ref struct` across an `await` compiled clean and **returned the wrong number**. This PR closed that silently; it was never claimed. The state machine tell was field count — 4 in the unsafe case vs 5 in the safe control, i.e. the by-ref-like local was never hoisted.

On the "can it change codegen even when it emits nothing?" half: the finder calls `Lowerer.Lower(literal.Body)` on a throwaway for *every* literal, so I byte-compared emitted DLLs for the 9 clean probes across both trees — **9 identical, 0 differ, 0 missing**. No codegen perturbation.

**Residual, SHOULD-FIX (test gap only):** all five files mentioning `GS0219` (`RefStructTests.cs`, `RefSafeEscapeTests.cs`, `Issue2350…`, `Issue2360…`, `Issue2891…`) contain **zero** `init(` shapes. The newly-live constructor path is entirely unpinned — reverting `Binder.cs:1600` would leave the suite green. One test of the shape above closes it.

PR #3071, opened 2026-08-02
Fix plain struct Object overrides

### LINKED ISSUE #2896: Plain struct cannot override Object virtual methods

A plain struct cannot declare a valid CLR override of an `Object` virtual method:

```gs
package FindingStructObjectOverride

struct Value {
  var Number int32

  override func ToString() string {
    return Number.ToString()
  }
}

func Main() int32 {
  return Value{ Number: 7 }.ToString() == "7" ? 0 : 1
}
```

Expected: the build succeeds and `Main` returns `0`.

Actual: the build fails with:

```text
GS0183: Method 'ToString' is marked 'override' but no matching open base method was found.
```

Equivalent C# builds, emits a virtual `Value::ToString` override, and returns `0`. G# data structs also emit CLR `Object` override slots for their synthesized members, but the explicit plain-struct override is rejected before an assembly is emitted.

DavidObando:
Fixes #2896

## Root cause

This was a binder rejection, not an emitter slot-loss bug. On the base commit, every positive plain-struct probe failed before emission with `GS0183`. `ExternalClrOverrideResolver.FindExternalBaseType` supplied the implicit `System.Object` base only for classes, so plain structs never reached the existing external override matcher.

The compiler emitter already handled the unlocked path correctly: emitted value-type overrides are virtual, final, hide-by-signature, reuse the `System.Object` slot, and carry explicit `MethodImpl` rows. The interpreter needed a parallel runtime fix because object-typed calls were reflection-dispatched against `StructValue` instead of the binder-recorded G# override. The evaluator follows `FunctionSymbol.OverriddenMethod` to the transitive root, then matches its binder-recorded `ExternalOverriddenMethod`, preserving the most-derived implementation in depth-3 user override chains without adding another signature matcher.

## Fresh-directory repro matrix

Every cell below compiled or evaluated from its own asserted-empty directory after a full solution build. `Core/`, `Compiler/`, and `Repl/` Core DLL mtimes matched. Results are unchanged from the provisional reused-directory run.

| Shape | bare `gsc` evaluate | `gsc /out:` emit/run | `gsi` |
|---|---|---|---|
| `ToString` only | `OVERRIDDEN-11` ×2 | same | same |
| `Equals(object)` only | `False` ×2 | same | same |
| `GetHashCode` only | `289613` ×2 | same | same |
| all three, top level | six expected values | same | same |
| all three, in function | six expected values | same | same |
| generic struct | `GENERIC-OVERRIDDEN-23` ×2 | same | same |
| interface implementation | marker + override ×2 | same | same |
| operators / `IEquatable` | `True, True, False, 289637` | same | same |
| nested struct | `NESTED-OVERRIDDEN-41` ×2 | same | same |
| shared member | `SHARED-43` + override ×2 | same | same |
| data-struct control | structural value ×2 | same | same |
| default-struct control | structural value ×2 | CLR type name ×2 | structural value ×2 |

All 36 driver cells returned rc 0 with empty stderr.

## Depth-3 override matrix

`L0 -> L1 -> L2` returns distinct `L0-11`, `L1-22`, `L2-33` values:

| Position | bare `gsc` evaluate | `gsc /out:` emit/run | `gsi` |
|---|---|---|---|
| top level | pass | pass | pass |
| inside function, called at top level | pass | pass | pass |

## Emission checks

The compiler regression test loads emitted bytes with `Assembly.Load(File.ReadAllBytes(...))`, calls `GetTypes()`, executes the child process, and checks:

- direct calls and boxed virtual dispatch produce distinct expected values;
- `MethodInfo.IsVirtual`, `IsFinal`, `IsHideBySig`, `NewSlot`, and `GetBaseDefinition()`;
- explicit `MethodImpl` counts for every positive shape;
- concrete direct calls use `call`;
- object dispatch uses `callvirt System.Object`;
- unconstrained generic dispatch emits `constrained. !!T` followed by `callvirt System.Object.ToString`.

## Shared-helper site count

Making `FindExternalBaseType` take one additional required parameter produced six printed `CS7036` diagnostics at exactly three unique call sites: method, property, and event resolution. Denominator: **3/3 callers route through the fixed helper**.

## Product mutations

| Hunk | Applied mutant | Killing test / symptom | Core hash round trip |
|---|---|---|---|
| implicit Object fallback | restore class-only condition | compiler matrix fails with `GS0183` for `ToString`, `Equals`, `GetHashCode`, and remaining struct overrides | clean `82b6ae4c…`, mutant `2b0a954c…`, restored `82b6ae4c…` |
| external-slot identity | invert binder-recorded MethodInfo equality | top-level and function interpreter cases fail with emitted evaluation type mismatch | clean `82b6ae4c…`, mutant `70c05ed…`, restored `82b6ae4c…` |
| transitive override root | invert `OverriddenMethod` walk condition | depth-3 matrix: evaluator drivers fail; emit-only control remains green | clean `82b6ae4c…`, mutant `3d1203b1…`, restored `82b6ae4c…` |

Each mutation was asserted present at exactly one source site before building.

## Validation on `main` `4f7559dd`

- `git merge-tree --write-tree origin/main HEAD`: rc 0; predicted merged tree equals `HEAD` tree
- merged-tree Release build: **BUILD_RC=0, 0 warnings, 0 errors**
- full merged suite: **14,346 passed, 0 failed, 1 skipped**
- bidirectional diagnostic documentation gate: **1 passed, 0 failed**
- cs2gs required gate: **0 new, 0 regressed, 0 stale**; `G11-Linq-Console` passed all stages

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExternalClrOverrideResolver.cs
+++ b/src/Core/CodeAnalysis/Binding/ExternalClrOverrideResolver.cs
@@ -153,7 +153,7 @@ internal static class ExternalClrOverrideResolver
             }
         }
 
-        return type?.IsClass == true ? TypeSymbol.Object : null;
+        return type != null ? TypeSymbol.Object : null;
     }
 
     private static Type GetReflectionBaseType(TypeSymbol importedBase)
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -1003,11 +1003,24 @@ public sealed partial class Evaluator
 
         if (receiver is StructValue sv)
         {
+            static MethodInfo FindExternalOverrideRoot(FunctionSymbol method)
+            {
+                for (var current = method; current != null; current = current.OverriddenMethod)
+                {
+                    if (current.ExternalOverriddenMethod != null)
+                    {
+                        return current.ExternalOverriddenMethod;
+                    }
+                }
+
+                return null;
+            }
+
             FunctionSymbol externalOverride = null;
             for (var type = sv.StructType; type != null && externalOverride == null; type = type.BaseClass)
             {
                 externalOverride = type.Methods.FirstOrDefault(method =>
-                    method.IsOverride && method.ExternalOverriddenMethod?.Equals(node.Method) == true);
+                    method.IsOverride && FindExternalOverrideRoot(method)?.Equals(node.Method) == true);
             }
 
             if (externalOverride != null)
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -1005,15 +1005,12 @@ public sealed partial class Evaluator
         {
             static MethodInfo FindExternalOverrideRoot(FunctionSymbol method)
             {
-                for (var current = method; current != null; current = current.OverriddenMethod)
+                while (method.OverriddenMethod != null)
                 {
-                    if (current.ExternalOverriddenMethod != null)
-                    {
-                        return current.ExternalOverriddenMethod;
-                    }
+                    method = method.OverriddenMethod;
                 }
 
-                return null;
+                return method.ExternalOverriddenMethod;
             }
 
             FunctionSymbol externalOverride = null;
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -1001,6 +1001,37 @@ public sealed partial class Evaluator
     {
         var receiver = EvaluateExpression(node.Receiver);
 
+        if (receiver is StructValue sv)
+        {
+            FunctionSymbol externalOverride = null;
+            for (var type = sv.StructType; type != null && externalOverride == null; type = type.BaseClass)
+            {
+                externalOverride = type.Methods.FirstOrDefault(method =>
+                    method.IsOverride && method.ExternalOverriddenMethod?.Equals(node.Method) == true);
+            }
+
+            if (externalOverride != null)
+            {
+                var frame = new ConcurrentDictionary<VariableSymbol, object>
+                {
+                    [externalOverride.ThisParameter] = receiver,
+                };
+
+                var parameterOffset = externalOverride.ExplicitReceiverParameter == null ? 0 : 1;
+                for (var i = 0; i < node.Arguments.Length; i++)
+                {
+                    var parameter = externalOverride.Parameters[i + parameterOffset];
+                    frame[parameter] = EvaluateExpression(node.Arguments[i]);
+                }
+
+                using (PushFrame(frame))
+                {
+                    var statement = program.Functions[externalOverride];
+                    return EvaluateUserMethodBody(externalOverride, statement);
+                }
+            }
+        }
+
         // Issue #517: Nullable<T> instance methods (e.g. `GetValueOrDefault`,
         // `Equals`, `ToString`) cannot dispatch through `MethodInfo.Invoke`
         // because the interpreter stores nullables as the underlying boxed T
--- a/test/Compiler.Tests/Emit/Issue2443ExternalClrOverrideEmitTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2443ExternalClrOverrideEmitTests.cs
@@ -8,6 +8,9 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Reflection.Emit;
+using System.Reflection.Metadata;
+using System.Reflection.PortableExecutable;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Xunit;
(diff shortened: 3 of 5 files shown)


DavidObando:
## Adversarial review — **NOT MERGEABLE** (1 BLOCKER, small and local)

Numbers true of `main` = **`5ad0109c`**. Merge base = **`185431c4`** (computed). Merged commit `22c90943`, tree **`dac05331`** — matches `git merge-tree` in **both** directions exactly. Base tree md5 `3646862f`, merged `8ef4ac1b`, `Core/`==`Compiler/`==`Repl/` fresh in both.

True content **4 files, +585/−10**; production is only **+32/−1** in 2 files (two-dot == three-dot; no artefact). Defect reproduced at base: `R1` fails **`GS0183`** on all three driver columns, rc0 on all three at merged.

### The headline claim is TRUE — and I verified it the hard way

Item 1 feared "the emitter was already correct" was unpinned by construction. **It holds.** My own inspector (`Assembly.Load(File.ReadAllBytes(...))` + raw `MetadataReader`) on the merged emitted binary:

```
METHODIMPL Val: body=Val::ToString     decl=Object::ToString
METHODIMPL Val: body=Val::Equals       decl=Object::Equals
METHODIMPL Val: body=Val::GetHashCode  decl=Object::GetHashCode
M ToString()     virt=True final=True hidebysig=True newslot=False basedef=Object::ToString
M Equals(Object) virt=True final=True hidebysig=True newslot=False basedef=Object::Equals
M GetHashCode()  virt=True final=True hidebysig=True newslot=False basedef=Object::GetHashCode
```

Slot reuse, finality, hide-by-signature and the `MethodImpl` rows are all exactly as claimed, and the PR **does** pin them: `AssertOverrideSlot` (`Issue2443ExternalClrOverrideEmitTests.cs:645`) checks `IsVirtual`, `NewSlot` false, and `GetBaseDefinition().MetadataToken` + `.Module` equality against `System.Object`; `AssertMethodImplCount` (`:346-354`) asserts exact counts `1/1/1/3/1/1/2/1/1`. No emitter change is needed. **Reclassification accepted.**

### Both hunks pinned by PRODUCT mutation

| mutant | md5 | effect |
|---|---|---|
| **M1** revert `ExternalClrOverrideResolver.cs:156` to `type?.IsClass == true` | `5b50725f` | every struct shape → `GS0183`; classes **unchanged** |
| **M2** delete `Evaluator.Calls.cs:996-1023` | `1fd8d096` | boxed `ToString` → `Val(N=5)`; boxed `GetHashCode` → **`2094006762`** vs direct `702`; `R9` flips to wrong `True` |
| clean (restored) | `8ef4ac1b` ✔ exact round-trip, `git status` clean | — |

Suites on merged: **Interpreter.Tests 611/611**, **Core.Tests 7072 passed / 1 skipped / 0 failed**, **Issue2443+Issue2896 11/11**.

---

## BLOCKER — the evaluator dispatches to the **wrong override** in a class override chain

`src/Core/CodeAnalysis/Evaluator.Calls.cs:996-1000`.

Three-level chain (`L0` → `L1` → `L2`, each `open override func ToString()`), object-typed receivers, **merged tree**:

| receiver | emitted binary (`gsc /out:`) | evaluator (bare `gsc` **and** `gsi`) |
|---|---|---|
| `L0` | `L0-930` | `L0-930` ✓ |
| `L1` | `L1-931` | **`L0-930`** ✗ |
| `L2` | `L2-932` | **`L0-930`** ✗ |

Two-level (`R5`): binary `OVER-911`, evaluator **`BASE-910`**. Single-level (`R6`): correct.

**Mechanism.** The loop keys on `method.ExternalOverriddenMethod?.Equals(node.Method)`. Only the **anchor** of the chain carries `ExternalOverriddenMethod`; `L1`/`L2` override a *user* method, so theirs is `null`. The walk from `sv.StructType` upward therefore returns the **least-derived** override, inverting virtual dispatch.

**Ownership proven, not argued.** M2 (remove the hunk) returns `L0(N=0)/L1()/L2()` — the base behaviour. M1 (revert the binder line) leaves the wrong output unchanged. So this is hunk B alone.

**Nothing will catch it later**: 611/611 + 7072/7072 + 11/11 all green with the bug present.

**Fairness, stated plainly:** this is **not a regression** — at base the evaluator printed `L0(N=0)/L1()/L2()`, also wrong. But it converts an obviously-broken output into a **plausible-looking wrong answer** at rc0, in brand-new code, and it diverges from the emitted binary for a feature that compiles and runs correctly today.

**Minimal fix** (`FunctionSymbol.OverriddenMethod` already exists, `FunctionSymbol.cs:222`): match a method whose override chain *transitively* roots at `node.Method`, rather than only the anchor —

```csharp
static MethodInfo Root(FunctionSymbol m) =>
    m.ExternalOverriddenMethod ?? (m.OverriddenMethod is { } b ? Root(b) : null);

externalOverride = type.Methods.FirstOrDefault(method =>
    method.IsOverride && Root(method)?.Equals(node.Method) == true);
```

## SHOULD-FIX

**1. Undisclosed blast radius: the interpreter hunk changes CLASS dispatch, not just structs.**
The guard is `receiver is StructValue sv` (`Evaluator.Calls.cs:996`), and `StructValue.StructType` is a `StructSymbol` (`StructValue.cs:35`) — which covers classes as well as structs (`StructSymbol.IsClass`, `StructSymbol.cs:303`). Measured: a class with a single `ToString` override on an object-typed receiver goes base `Over()` → merged **`OVER-921`**. That is a genuine *improvement*, but it is undocumented, untested, and it is the same code path that produces the BLOCKER above. Please state the class behaviour change and cover it.

**2. The "nine shapes" collapse to ONE binder cause.**
M1 flipped **all** struct shapes back to `GS0183` simultaneously — so for the binder half the nine shapes are one root cause rendered nine times, not nine independent pins. Fine as regression breadth; please don't read it as nine-way coverage. (This was item 2's masking concern, and it is real, though benign.)

**3. Implicit `ToString` still diverges between drivers.** Merged, same source:

```
String.Format("{0}", boxed)   binary: TS-701   evaluator: Val(N=5)
Console.WriteLine(structVal)  binary: TS-701   evaluator: Val(N=5)
```

Only *explicit* `.ToString()` is routed to the override. **Pre-existing** — classes show the identical split at base — so out of scope for this PR, but it means `Console.WriteLine(myStruct)`, the most common way an overridden `ToString` is observed, still disagrees between `gsc` and `gsi`. Worth a follow-up issue.

## Verified, no action

- **Item 3 (identity):** generic struct at **two** instantiations dispatches correctly on all three columns. The external base is always non-generic `System.Object`, so the `MethodInfo` is stable — the substituted-vs-definition risk does not arise.
- **Item 4 (boxing):** boxed/unboxed `GetHashCode` agree (`702`/`702`); `Equals(nil)` and `Equals("a string")` consistent across all three drivers.
- **Item 5:** `IEquatable[V].Equals(V)` stays distinct from `Equals(object)` — `True|False|False`, identical on all three drivers.
- **Item 6:** confirmed bare `gsc` **evaluates** (positive control printed `4242` + `Success.`); the PR's emit tests use `/out:` (`:679`), the correct column for the emitter claim. The evaluator column — shared by bare `gsc` and `gsi` — is where the BLOCKER lives.
- **Controls:** no-override `struct`/`class`/`data struct` byte-identical base vs merged on all three drivers.
- **`ref struct`:** still correctly rejected at merged, now via `GS0219` (by-ref-like) instead of `GS0183`. No boxing hole opened by the broadened binder predicate.

DavidObando:
## Round-2 re-verification: **MERGEABLE**

Anchors: `main` **`abfab99c`**, head **`14dabacbf`**, merge base **`4f7559dd`**, merged tree **`52d16d9c`** (matched `merge-tree` prediction exactly), `BUILD_RC=0`, 0 warnings. Author rebased onto `4f7559dd`; I re-anchored onto `abfab99c` (two emitter PRs landed since). `src/Repl/Program.cs` untouched — collision zone clear.

### The prior BLOCKER is fixed, and fixed *structurally*

`Evaluator.Calls.cs` `FindExternalOverrideRoot` uses `while (method.OverriddenMethod != null)` — "walk to the end", **not** "one more step". That is the distinction that matters, and it holds at arbitrary depth.

| probe | bare `gsc` (EVAL) | `gsi` | `gsc /out:` (EMIT) |
|---|---|---|---|
| depth-4 `L0→L1→L2→L3` (930/931/932/933) | `L0-930 L1-931 L2-932 L3-933` ✓ | ✓ | ✓ |

**Base positive control** — same file at `abfab99c`: `L0() L1() L2() L3()`. No user override honored *at any depth*. The probe discriminates.

### PRODUCT mutant, named, with the test that dies

**M-T**: `while (method.OverriddenMethod != null)` → `if (...)` (single-step). Asserted applied: pattern matched exactly 1 site, source md5 `33f62334`→`a6c3e101`, `BUILD_RC=0`, `GSharp.Core.dll` md5 `1565c315`→`34695f1e`, mtime fresh on `Core/`/`Compiler/`/`Repl/`.

**KILLED** by `ClassObjectOverrideChain_UsesMostDerivedOverrideAcrossDrivers`:
```
Expected: "L0-11\nL1-22\nL2-33\n"
Actual:   "L0-11\nL1-22\nL1-22\n"
```
Depth-4 under the same mutant collapses *two* rows: `L0-930 L1-931 L1-931 L1-931`.

**Non-vacuity proven by mutation, not by reading.** The dying cells are exactly **4 of 6** = `{top-level, in-function} × {gsc-evaluate, gsi}`; the 2 `gsc-emit` cells survive because the emitter is a different path. So the `insideFunction: True` variant genuinely **executes** — this is not the `func Main()`-only vacuous-pass trap. The author's test is a real 6-cell axis with distinct values (`L0-11`/`L1-22`/`L2-33`).

### Suites, unfiltered, on the merged tree
- Interpreter.Tests **647 passed / 0 failed** (mutant run: 643/4 — consistent)
- Core.Tests **7088 passed / 0 failed / 1 skipped**

### Hypothesis I killed (mine, not the author's)
Arity-blind `FirstOrDefault` mis-selecting on the `Equals(object)` overload wrinkle: **does not fire.** Struct with both `override func Equals(value object)` and non-override `func Equals(other Val)`:

`a.Equals(b:Val)`→`False` (non-override correctly picked), `a.Equals(boxed:object)`→`True`, `boxed.Equals(a)`→`True` — **identical on all three drivers.** The binder resolves overloads before the hunk engages.

---

## SHOULD-FIX 1 — implicit call sites diverge; the container case is a silent wrong *answer*

Struct `Val` with `GetHashCode`→`702` (constant) and `Equals(object)`→`true`:

| line | EVAL | `gsi` | EMIT |
|---|---|---|---|
| `direct/boxed.ToString()`, `.GetHashCode()`, `.Equals()` | correct | correct | correct |
| `String.Format("{0}", boxed)` | `Val(N=5)` ✗ | `Val(N=5)` ✗ | `TS-701` ✓ |
| **`Dictionary[Val,string].Count`** | **`2`** ✗ | **`2`** ✗ | **`1`** ✓ |

The `String.Format` row is the earlier review's SHOULD-FIX 3, reproduced. **The `Dictionary` row is new and strictly stronger**: with a constant hash and `Equals`→true both keys *must* collide, so this is not display text — it is a differing **container element count**, silent, at rc0, in the column CI cannot see (the samples gate is emit-only).

**Not a regression.** At `abfab99c` the whole shape is `GS0183` ×3 and does not compile. This is newly-enabled-but-incomplete, which is why it is SHOULD-FIX and not BLOCKER — but #2896's capability is only half-delivered in the evaluator.

## SHOULD-FIX 2 — the struct-guarded hunk also changes **class** dispatch (confirmed)

Guard is `receiver is StructValue sv`, but `StructValue.StructType` is a `StructSymbol`, which covers classes (`StructSymbol.IsClass`). My base control makes this concrete: at `abfab99c` even **depth-1** `l0.ToString()` prints `L0()`, i.e. user `ToString()` overrides on *classes* were never honored in the evaluator. Merged fixes that. Genuine improvement, but undocumented and outside the issue's stated scope — it deserves a line in the PR body and a class-only regression test that does not depend on the struct path.

## NIT
Author's tests are **depth-3 only**. Depth-4 passes (above) but nothing in the suite pins it. One `[InlineData]` row extending `BuildClassOverrideChainSource` to `L3` would close it cheaply.

---

Hygiene: all mutation in a private scratch clone, exact md5 round-trip (`33f62334`, binary back to `1565c315`), `git status --porcelain` empty. `i2896` worktree never written. No e2e; `~/.nuget/packages/gsharp.net.sdk/` never touched. Every probe compiled into its own asserted-empty directory. Disclosure: one rebuild was started at load 48.89; it reproduced the previously-recorded md5 byte-identically.

PR #3072, opened 2026-08-02
Fix inherited field owners in generic derived classes

### LINKED ISSUE #2895: Inherited field access from a generic derived class uses the wrong declaring type

Accessing a field declared on a non-generic base class from a generic derived class emits the field reference against the derived type:

```gs
package FindingGenericInheritedField

open class Base {
  internal var dirty bool
}

class Derived[T any] : Base {
  func Mark() {
    dirty = true
  }

  func IsDirty() bool {
    return dirty
  }
}

func Main() int32 {
  let value = Derived[int32]()
  value.Mark()
  return value.IsDirty() ? 0 : 1
}
```

Expected: the build succeeds and `Main` returns `0` after `Mark` writes `Base.dirty`.

Actual: the build succeeds, but execution throws:

```text
System.MissingFieldException: Field not found: 'FindingGenericInheritedField.Derived`1.dirty'.
```

ILSpy shows both field references as `Derived`1<!T>::dirty`, although the field is declared on non-generic `Base`. Equivalent C# emits `Base::dirty` for both references and returns `0`.

DavidObando:
## Summary

- preserve the field's declaring base type in object-initializer and explicit-receiver assignment bound nodes
- emit a non-generic declaring type directly instead of falling back to the generic receiver
- add metadata, load, child-process runtime, and interpreter coverage for the affected inheritance shapes

This changes inherited field reads/writes only. It does not change generic method-spec caching (#3065) or close any issue other than #2895.

Fixes #2895

## Root cause

`ResolveFieldReferenceContainer` represented a non-generic declaring type as `null`. Its callers interpreted `null` as “use the receiver,” so a field declared by `Base` was emitted as a MemberRef parented by `Derived<T>`. The assignment binder also discarded the declaring type already returned by `TryGetFieldIncludingInherited` in three write paths.

The fix keeps the declaring type end-to-end. Generic-base construction still uses the existing `FindConstructedGenericBase` path.

## Metadata and inheritance matrix

| Shape | Base metadata/runtime | HEAD metadata/runtime |
|---|---|---|
| `Derived[T] : Base`, unqualified read/write/compound | `Derived<T>::number`; invalid token / `MissingFieldException` | `Base::number`; `11`, `22`, `33` |
| G#-declared field type | `Derived<T>::payload`; invalid token | `Base::payload`; `44` |
| explicit receiver, top-level/function | `Derived<string>::explicitValue`; invalid token | `Base::explicitValue`; `55`, `66` |
| expression receiver (`child.explicitValue`) | `Derived<string>::explicitValue`; invalid token | `Base::explicitValue`; `67` |
| object initializer | `Derived<string>::initializedValue`; invalid token | `Base::initializedValue`; `77` |
| two levels via `Mid[List[T]]` | `Derived<T>::item`; invalid token | `Base::item`; correct value |
| nested generic derived class | nested `Derived<T>::item`; invalid token | `NestedPin.Base::item`; correct value |
| `Derived[T] : GenericBase[T]` | constructed `GenericBase<T>`; passes | unchanged; passes (`11`) |
| `Derived[T] : GenericBase[int32]` | constructed `GenericBase<int32>`; passes | unchanged; passes (`22`) |
| `Derived[T] : GenericBase[List[T]]` | constructed `GenericBase<List<T>>`; passes | unchanged; passes (`1`) |
| inherited static field | `Base::sharedValue`; passes | unchanged; passes (`44`) |
| inherited field address/out | `Base::addressValue`; passes | unchanged; passes (`33`) |
| struct base | rejected by the language (`GS0382`) | unchanged |

The base parents above were read from raw MemberRef metadata. HEAD field operands resolve through reflection to the listed `FieldInfo.DeclaringType`.

## Driver × position

| Driver / position | Base | HEAD |
|---|---|---|
| `gsc`, top-level | compile `0`; runtime `134`, `MissingFieldException` | compile `0`; runtime `0`, `121` |
| `gsc`, function | compile `0`; runtime `134`, `MissingFieldException` | compile `0`; runtime `0`, `122` |
| `gsi`, top-level | rc `0`, `121` | rc `0`, `121` |
| `gsi`, function | rc `0`, `122` | rc `0`, `122` |

## Base pins and guards

The seven compiler facts were run with all production hunks reverted:

- pins (fail at base, pass at HEAD): **4**
  - `UnqualifiedAccessesInsideGenericDerivedUseNonGenericBase`
  - `ExplicitReceiverWritesUseNonGenericBase`
  - `ObjectInitializerWriteUsesNonGenericBase`
  - `MultiLevelAndNestedGenericDerivedUseTheirDeclaringBases`
- guards (pass at base and HEAD): **3**
  - constructed generic-base shapes
  - inherited static field
  - inherited field address/out

The two actual-`gsi` position facts are guards and pass at base and HEAD.

## Consumer count

Making `ResolveFieldToken` require a third parameter produced **19 consumers** on the final merged tree:

- 16 direct calls (`CS7036`)
- 3 method-group/delegate uses (`CS1503`)

The production fix remains at the shared field-container resolution and bound-node owner points; it does not patch individual token consumers.

## Mutation results

Each mutant built successfully, changed `GSharp.Core.dll`, and was followed by a full rebuild whose hash returned to the clean value.

| Production arm | Build-green mutant | Killing test | Result |
|---|---|---|---|
| non-generic emitter fallback | restore `null` fallback | `UnqualifiedAccessesInsideGenericDerivedUseNonGenericBase` | killed: emitted field token cannot resolve |
| object-initializer owner | use receiver type | `ObjectInitializerWriteUsesNonGenericBase` | killed: emitted field token cannot resolve |
| simple-variable assignment owner | use receiver type | `ExplicitReceiverWritesUseNonGenericBase` | killed: emitted field token cannot resolve |
| expression-receiver assignment owner | use receiver type | `ExplicitReceiverWritesUseNonGenericBase` | killed: emitted field token cannot resolve |

Clean Core hash: `5417e6ad49a15dc356f4114f3b72fbb997926acfc9fc5a6df9cf5baab0e42520`.
Mutant hashes: `90f2f7d7`, `45286c47`, `abad8096`, `6a7bc03e`; every restored build returned to the clean hash.

## Validation

- merged tree: `git merge-tree --write-tree origin/main HEAD` equals `HEAD^{tree}`
- Release CI build: 0 warnings, 0 errors
- targeted compiler: 7 passed
- targeted interpreter: 2 passed
- full merged suite: **14,257 passed, 0 failed, 1 skipped**
- cs2gs corpus: 20 apps examined; 19 green; 1 known baseline gap; 0 new, 0 regressed, 0 stale
- artifacts: 4,637 production DLLs examined, 0 zero-byte; 12 intentional zero-byte pipeline fixtures excluded
- direct `GSharp.Core.dll` copies: 22 examined, one mtime and one hash

The local Oahu script was not run because it invokes e2e tests, which are prohibited for this worktree; the required `cs2gs-oahu` CI job passed.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs
@@ -474,7 +474,7 @@ internal sealed partial class ExpressionBinder
 
                 var value = BindExpression(initSyntax.Value);
                 var converted = conversions.BindConversion(initSyntax.Value.Location, value, field.Type);
-                return new BoundFieldAssignmentExpression(initSyntax, receiverLocal, structSymbol, field, converted);
+                return new BoundFieldAssignmentExpression(initSyntax, receiverLocal, fieldDeclaringType, field, converted);
             }
 
             if (TypeMemberModel.TryGetProperty(structSymbol, propertyName, out var prop, out var propDeclaringType))
@@ -991,10 +991,10 @@ internal sealed partial class ExpressionBinder
 
         if (usesExpressionReceiver)
         {
-            return BoundFieldAssignmentExpression.WithExpressionReceiver(null, assignmentReceiver, structSymbol, field, converted);
+            return BoundFieldAssignmentExpression.WithExpressionReceiver(null, assignmentReceiver, fieldDeclaringType, field, converted);
         }
 
-        return new BoundFieldAssignmentExpression(null, variable, structSymbol, field, converted);
+        return new BoundFieldAssignmentExpression(null, variable, fieldDeclaringType, field, converted);
     }
 
     /// <summary>
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.MemberAccess.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.MemberAccess.cs
@@ -1977,8 +1977,9 @@ internal sealed partial class MethodBodyEmitter
     // reachable from the receiver's hierarchy instead — this yields the
     // concrete base (`FrameFilterBase`1<int32>`) for a non-generic leaf and the
     // self-instantiation (`FrameFilterBase`1<!0>`) when accessed from within the
-    // generic type itself. Returns null when the bare FieldDef should be used
-    // (a non-generic declaring type with no generic base in scope).
+    // generic type itself. A non-generic declaring type is returned unchanged;
+    // ResolveFieldToken will then select its bare FieldDef instead of guessing
+    // from a generic derived receiver.
     private static StructSymbol ResolveFieldReferenceContainer(StructSymbol declaringType, StructSymbol receiver, FieldSymbol field)
     {
         if (declaringType == null)
@@ -2000,15 +2001,7 @@ internal sealed partial class MethodBodyEmitter
             return constructed;
         }
 
-        // No receiver-reachable constructed base (e.g. a null/static receiver):
-        // fall back to the declaring type when it is itself generic.
-        if (!declaringType.TypeArguments.IsDefaultOrEmpty)
-        {
-            return declaringType;
-        }
-
-        var def = declaringType.Definition ?? declaringType;
-        return def.TypeParameters.IsDefaultOrEmpty ? null : declaringType;
+        return declaringType;
     }
 
     // Issue #1467: the property analogue of ResolveFieldReferenceContainer.
--- a/test/Compiler.Tests/Emit/Issue2895InheritedFieldDeclaringTypeTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2895InheritedFieldDeclaringTypeTests.cs
@@ -1,0 +1,437 @@
+// <copyright file="Issue2895InheritedFieldDeclaringTypeTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Emit;
+using Xunit;
+
+namespace GSharp.Compiler.Tests.Emit;
+
+/// <summary>
+/// Issue #2895: inherited field references from generic derived classes must
+/// name the field's declaring base type, not the generic receiver type.
+/// </summary>
+public class Issue2895InheritedFieldDeclaringTypeTests
+{
+    private static readonly IReadOnlyDictionary<ushort, OpCode> OpCodesByValue =
+        typeof(OpCodes)
+            .GetFields(BindingFlags.Public | BindingFlags.Static)
+            .Where(field => field.FieldType == typeof(OpCode))
+            .Select(field => (OpCode)field.GetValue(null)!)
+            .ToDictionary(opCode => unchecked((ushort)opCode.Value));
+
+    [Fact]
+    public void UnqualifiedAccessesInsideGenericDerivedUseNonGenericBase()
+    {
+        const string source = """
+            package Issue2895Unqualified
+            import System
+
+            data struct Payload(Code int32)
+
+            open class Base {
+                internal var number int32 = 11
+                internal var payload Payload = Payload(44)
+            }
+
+            class Derived[T] : Base {
+                func Read() int32 { return number }
+                func Write() int32 {
+                    number = 22
+                    return number
+                }
+                func Compound() int32 {
+                    number += 11
+                    return number
+                }
+                func ReadPayload() int32 { return payload.Code }
+            }
+
+            let value = Derived[string]()
+            Console.WriteLine(value.Read())
+            Console.WriteLine(value.Write())
+            Console.WriteLine(value.Compound())
+            Console.WriteLine(value.ReadPayload())
+            """;
+
+        using var artifact = Compile(nameof(UnqualifiedAccessesInsideGenericDerivedUseNonGenericBase), source);
+        AssertFieldOwners(artifact.Assembly, "number", "Base", minimumReferences: 6);
+        AssertFieldOwners(artifact.Assembly, "payload", "Base", minimumReferences: 1);
+        AssertRuns(artifact.Path, "11\n22\n33\n44\n");
+    }
+
+    [Fact]
+    public void ExplicitReceiverWritesUseNonGenericBase()
+    {
+        const string source = """
+            package Issue2895Explicit
+            import System
+
+            open class Base {
+                internal var explicitValue int32
+            }
+
+            class Derived[T] : Base {
+            }
+
+            class Holder {
+                internal var child Derived[string] = Derived[string]()
+
+                func Write() int32 {
+                    child.explicitValue = 67
+                    return child.explicitValue
+                }
+            }
+
+            let top = Derived[string]()
+            top.explicitValue = 55
+            Console.WriteLine(top.explicitValue)
+
+            func Run() int32 {
+                let local = Derived[int32]()
+                local.explicitValue = 66
+                return local.explicitValue
+            }
+
+            Console.WriteLine(Run())
+            Console.WriteLine(Holder().Write())
+            """;
+
+        using var artifact = Compile(nameof(ExplicitReceiverWritesUseNonGenericBase), source);
+        AssertFieldOwners(artifact.Assembly, "explicitValue", "Base", minimumReferences: 6);
+        AssertRuns(artifact.Path, "55\n66\n67\n");
+    }
+
+    [Fact]
+    public void ObjectInitializerWriteUsesNonGenericBase()
+    {
+        const string source = """
+            package Issue2895Initializer
+            import System
+
+            open class Base {
+                internal var initializedValue int32
+            }
+
+            class Derived[T] : Base {
+            }
+
+            let value = Derived[string]() { initializedValue = 77 }
+            Console.WriteLine(value.initializedValue)
+            """;
+
+        using var artifact = Compile(nameof(ObjectInitializerWriteUsesNonGenericBase), source);
+        AssertFieldOwners(artifact.Assembly, "initializedValue", "Base", minimumReferences: 2);
+        AssertRuns(artifact.Path, "77\n");
+    }
+
+    [Fact]
+    public void MultiLevelAndNestedGenericDerivedUseTheirDeclaringBases()
+    {
+        const string source = """
+            package Issue2895Shapes
+            import System
+            import System.Collections.Generic
+
+            open class Root {
+                internal var deepValue int32 = 88
+            }
+
+            open class Mid[T] : Root {
+            }
+
+            class Deep[T] : Mid[List[T]] {
+                func Read() int32 { return deepValue }
+            }
+
+            open class NestedBase {
+                internal var nestedValue int32 = 99
+            }
+
+            class Container {
+                class Nested[T] : NestedBase {
+                    func Read() int32 { return nestedValue }
+                }
+            }
+
+            Console.WriteLine(Deep[string]().Read())
+            Console.WriteLine(Container.Nested[string]().Read())
+            """;
+
+        using var artifact = Compile(nameof(MultiLevelAndNestedGenericDerivedUseTheirDeclaringBases), source);
+        AssertFieldOwners(artifact.Assembly, "deepValue", "Root", minimumReferences: 1);
+        AssertFieldOwners(artifact.Assembly, "nestedValue", "NestedBase", minimumReferences: 1);
+        AssertRuns(artifact.Path, "88\n99\n");
+    }
+
+    [Fact]
+    public void GenericBaseShapesUseConstructedDeclaringBase()
+    {
+        const string source = """
+            package Issue2895GenericBases
+            import System
+            import System.Collections.Generic
+
+            data struct Payload(Code int32)
+
+            open class GenericBase[T] {
+                internal var genericValue T
+            }
+
+            class Forwarded[T] : GenericBase[T] {
+                func Set(value T) { genericValue = value }
+                func Read() T { return genericValue }
+            }
+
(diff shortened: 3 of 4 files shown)


DavidObando:
## Review: NOT MERGEABLE — one blocker (scope of the `Fixes #2895` claim)

The fix itself is correct, minimal, and well pinned. My objection is not to the code; it is that **#2895's exact symptom still reproduces on the merged tree in an idiomatic form**, while the PR auto-closes the issue.

Numbers below are true of **`main` `5ad0109c`** + head `b2765951` unless marked otherwise. `main` moved to **`33bea4af`** (#3047, #3024) during the review; zero file overlap with this PR, `merge-tree` rc0, and I **re-verified the blocker on `33bea4af` + PR** (merged `03232938`), where it still reproduces.

Note the briefing's head `c81aed609` is stale — the PR was rebased to `b2765951`. I confirmed it is a **pure rebase**: 582 patch lines on both sides, zero content difference after stripping `index` lines.

Build hygiene: full-solution Release CIB only; mtime verified on `Core/`, `Compiler/` **and `Repl/`** before every probe; each probe compiled into its own empty directory; Core hash round-tripped two-sided `41b938e1` → base `3646862f` → `41b938e1`.

---

### BLOCKER — the composite-literal form of the same defect is unfixed, and this PR closes #2895

Minimal pair. Identical types, identical field, identical instantiation; **only the initializer syntax differs**:

```gsharp
open class Base { internal var w int32 }
class Derived[T] : Base { }

let value = Derived[string]() { w = 88 }   // constructor + initializer block
let value = Derived[string]{ w: 99 }       // composite literal
```

| form | `gsc` evaluate | `gsc /out:` + run | `gsi` |
|---|---|---|---|
| `Derived[string]() { w = 88 }` | `88` | rc0, **`88`** | `88` |
| `Derived[string]{ w: 99 }` | `99` | rc0 emit, **rc134** | `99` |

```
System.MissingFieldException: Field not found: 'P9.Derived`1.w'.
```

That is #2895's exact exception, exact shape (inherited field, generic derived receiver, parent emitted as `Derived`1` instead of `Base`), and it is an **emit-only** defect — both the evaluator and `gsi` print the right value, which is precisely the driver differential the issue describes.

**This is pre-existing, not a regression** — the composite-literal form fails identically at base `5ad0109c`. But `Fixes #2895` will close an issue that still reproduces.

**Where it lives.** The composite literal never becomes a `BoundFieldAssignmentExpression`, so neither the binder hunks nor `ResolveFieldReferenceContainer` are on its path. Three emit sites resolve the token from the **literal's** type and never consult the field's declaring type:

- `MethodBodyEmitter.Expressions.cs:1188`
- `MethodBodyEmitter.Expressions.cs:1241`
- `MethodBodyEmitter.Expressions.cs:1327`

```csharp
if (isGeneric || literal.StructType.ClrType != null)   // :1186, isGeneric from :1118
{
    fieldHandle = this.outer.userTokens.ResolveFieldToken(literal.StructType, init.Field);
}
```

`isGeneric` is computed from `literal.StructType` at `:1118`, so a generic derived literal always parents at `Derived`1` regardless of where the field is declared — the identical "guess from the receiver" mistake the PR's own root-cause paragraph describes.

The declaring type is **already in hand**: `ExpressionBinder.Literals.cs:1279` computes `fieldDeclaringType` and uses it for the accessibility check, then drops it, because `BoundFieldInitializer(FieldSymbol field, BoundExpression value)` (`BoundFieldInitializer.cs:21`) has nowhere to put it.

**Two acceptable remedies — the second is fine by me:**
1. Carry the declaring type on `BoundFieldInitializer` and route the three sites through the existing `ResolveFieldReferenceContainer(declaringType, literal.StructType, init.Field)`; or
2. **Keep this PR exactly as it is**, downgrade `Fixes #2895` to a partial reference, and file the composite-literal form as a follow-up. The fix here is good and I would not hold it hostage to a larger change.

What I do not think is acceptable is merging as-is with the auto-close.

### SHOULD-FIX — the consumer count was taken and then not acted on

The body reports **19 consumers** of `ResolveFieldToken` and concludes: *"The production fix remains at the shared field-container resolution and bound-node owner points; it does not patch individual token consumers."*

Three of those 19 consumers are `Expressions.cs:1188/1241/1327` — the ones above. The measurement that would have caught the blocker was performed, and then dismissed by category. **Count by site, then check each site**, rather than deciding as a class that consumers need no attention.

Separately, "three write paths" is prose, not a measured denominator. I made the compiler count `BoundFieldAssignmentExpression` construction sites with a required parameter: **180 raw diagnostics, 50 unique `file:line`**, of which **13** are in `ExpressionBinder.Assignments.cs`. The three changed are the right three (see mutation below), but the denominator was never established, and that is the same blind spot.

### Verified — attack points that came back clean

**Sentinel separation is genuine (not a flip).** `null` out of `ResolveFieldReferenceContainer` (`MethodBodyEmitter.MemberAccess.cs:1985`) now means only "no declaring type recorded". The "use the receiver" fallback (`:486`) is reachable only in that case; the non-generic case returns the declaring type and `ResolveFieldToken` (`UserTokenResolver.cs:1283`) sends a non-generic container to `StructFieldDefs[field]`, which is keyed by the **field**, hence inherently the declaring type's own row. All three consumers (`:464`, `:581`, `:1748`) handle it consistently. The ambiguity is removed, not relocated.

**Reads were confirmed, not assumed.** Read sites already carried the declaring type (`ExpressionBinder.Access.MemberLookup.cs:264`); the read fix comes entirely from the resolver change. Measured: unqualified read inside the generic derived class fails at base and passes at head.

**The "unchanged" rows are genuinely unchanged, by measurement.** Including the shape you flagged as absent from both lists — a **non-generic grandparent above a generic parent** (`Root` non-generic ← `Mid[T]` ← `Leaf[T]`, field on `Root`): base **rc134**, head **`66`**. `FindConstructedGenericBase` correctly walks past the generic middle to the non-generic declaring root. Inherited static field: passes at base and head (`77`).

**Full base-vs-head matrix**, distinct values, `{top-level, in-function}` × `{gsc evaluate, gsc emit, gsi}` — deliberately not the shape of your tests:

| probe | base emit+run | head emit+run |
|---|---|---|
| unqualified read/write, in `func Main()` | rc134 | `11` |
| unqualified read/write, top level | rc134 | `22` |
| explicit receiver write | rc134 | `33` |
| compound `+=` | rc134 | `44` |
| **composite literal** | **rc134** | **rc134** |
| non-generic grandparent / generic parent | rc134 | `66` |
| inherited static | `77` | `77` |

`gsc` evaluate and `gsi` print the correct value on **every** row at both base and head — the defect is emit-only throughout.

**#3065 disclaimer verified behaviourally, not textually.** The diff has 0 occurrences of `MethodSpec`/`Remap`/`SymbolKey`, and my #3065 repro returns rc0 `11`/`22` on both `gsc` emit and `gsi` on the merged tree. No interaction.

### Endorsements

**Every production hunk is individually pinned.** The body reverted all hunks together (4 pins / 3 guards); I mutated each independently, which is stronger and confirms no masking:

| mutant | site | result |
|---|---|---|
| M1 | `ResolveFieldReferenceContainer` returns `null` for non-generic | **KILLED** — 4 tests |
| M2 | `:477` `fieldDeclaringType` → `structSymbol` | **KILLED** — `ObjectInitializerWriteUsesNonGenericBase` |
| M3 | `:994` `fieldDeclaringType` → `structSymbol` | **KILLED** — `ExplicitReceiverWritesUseNonGenericBase` |
| M4 | `:997` `fieldDeclaringType` → `structSymbol` | **KILLED** — `ExplicitReceiverWritesUseNonGenericBase` |

4/4. (M3 and M4 are killed by the same test, so nothing distinguishes those two sites — both are covered, so this is an observation, not a request.)

**`AssertFieldOwners` is exactly the right instrument and should not be weakened.** It walks each method's IL byte array, decodes `InlineField` operands, resolves them, and asserts `FieldInfo.DeclaringType` — a real metadata assertion, not a printed value, so it can distinguish a correct parent from a wrong-but-resolvable one. The `minimumReferences` floor is a built-in positive control against a vacuous zero. This is the bar I would want held; please keep it.

### Suites (merged tree, `main` `5ad0109c`)

- `Core.Tests` **7072 passed / 0 failed / 7073 total**
- `Interpreter.Tests` **609 / 0 / 609**
- `Compiler.Tests` filtered `~Emit` **3818 / 0 / 3818** (1h09m)
- `Issue2895` targeted **7 / 7**

**Disclosed filter:** I did not run `Compiler.Tests` unfiltered (~1h20m under load that peaked at 41). `~Emit` is the regression-risk surface for a change to field-token parenting, but it is a filter, and a filtered green is not a green.

### Summary

- **BLOCKER** — `Expressions.cs:1188/1241/1327`: composite literal `Derived[T]{ f: v }` still emits `Derived`1::f` and throws `MissingFieldException`; #2895 is not fully fixed but is auto-closed. Fix the three sites **or** drop the auto-close and file the gap.
- **SHOULD-FIX** — the 19-consumer count included the three defective sites and was dismissed by category; "three write paths" is unmeasured (compiler says 50 sites, 13 in that file).
- **NIT** — none worth your time.

Everything else measured clean, and the mutation and metadata-assertion discipline here is better than most of what I review.

DavidObando:
## Correction — I am downgrading my blocker. Revised verdict: **MERGEABLE**

Bulletin #4 §1 ("a closure requires the **reported shape** reproduced, not adjacent shapes") made me check something I had not: my `p1` probe was a **paraphrase** of #2895, not the issue's own text. I had rewritten `dirty bool` as an `int32` and swapped the exit-code result for a printed value. By my own standard that is not good enough, so I ran the issue verbatim.

**#2895's repro, unmodified, on the co-merged tree** (`main` `33bea4af` + this PR = `03232938`), compiled into an asserted-empty directory:

```
emit_rc=0
run_rc=0        (0 = fixed, 1 = wrong value, 134 = the #2895 bug)
```

**The reported shape is fixed.** At base it is rc134. So `Fixes #2895` is defensible on the evidence the fleet's own closure rule demands, and my blocker applied a stricter standard than that rule — it required the PR to also fix a sibling construct that #2895 never reported.

I also re-checked the terminology I leaned on, and it cuts against me: the compiler's own comment at `ExpressionBinder.Literals.cs:1291` calls `Foo{ member: value }` a **"composite/struct literal"**, distinct from an object initializer. The body's "object initializer" row describes `Derived[string]() { initializedValue = 77 }`, which is exactly what that row measured. **The matrix row is accurate.** I read a coverage claim into it that the body did not make.

### What stands, at reduced severity

**SHOULD-FIX (was BLOCKER) — composite-literal form of the same defect family is live.** Unchanged as a measurement; only its consequence changes:

```gsharp
open class Base { internal var w int32 }
class Derived[T] : Base { }
let value = Derived[string]{ w: 99 }        // composite literal
```
→ `gsc` emit rc0, run **rc134**, `MissingFieldException: Field not found: 'P9.Derived`1.w'`; `gsc` evaluate and `gsi` both print `99`. Fails identically at base, so **pre-existing, not a regression**, and not something this PR introduced or claimed.

Sites: `MethodBodyEmitter.Expressions.cs:1188`, `:1241`, `:1327` parent at `literal.StructType` and never consult the field's declaring type. The declaring type is already computed at `ExpressionBinder.Literals.cs:1279` and dropped, because `BoundFieldInitializer.cs:21` has no field for it.

This is a live defect and it does match #2895's *general* sentence ("a field declared on a non-generic base class ... emits the field reference against the derived type") even though it does not match its repro. **My recommendation is now: merge this PR, and file the composite-literal form as its own issue** — I have handed the minimal repro above to the maintainer, who is the sole issue filer. Whether #2895 should stay open pending that is his call, not a reason to hold this PR.

**SHOULD-FIX (unchanged) — the consumer count was taken and then dismissed by category.** The body counts 19 `ResolveFieldToken` consumers and concludes the fix "does not patch individual token consumers." Three of those 19 are the sites above. Count by site, then check each site.

### Everything else from my first comment stands, and I re-verified it

All 9 probes re-run with the directory **asserted empty** (`ls -A` = 0) before every compile, per Bulletin #4 §2 — every result identical, nothing moved:

| probe | base | head |
|---|---|---|
| unqualified read/write in `func Main()` | rc134 | `11` |
| unqualified read/write top level | rc134 | `22` |
| explicit receiver write | rc134 | `33` |
| compound `+=` | rc134 | `44` |
| composite literal | rc134 | **rc134** |
| non-generic grandparent / generic parent | rc134 | `66` |
| inherited static | `77` | `77` |
| object initializer `T() { f = v }` | rc134 | `88` |

Unchanged: 4/4 hunks individually pinned (M1 kills 4 tests, M2/M3/M4 one each); sentinel genuinely separated rather than flipped; reads confirmed by measurement; `AssertFieldOwners` endorsed as a real IL/metadata assertion with a `minimumReferences` positive control — please keep it; #3065 non-interaction proven **behaviourally** (repro returns rc0 `11`/`22` on the merged tree), not from file lists.

One methodological note worth recording, since this PR's own repro has the shape: the verbatim #2895 file is `func Main()`-only, so its **`gsi` column is structurally vacuous** — rc0 with empty output, because `gsi` only executes top-level code. The exit code is the only discriminator there, which is why I stated 0/1/134 in advance.

### Suite scope, stated honestly

My suite numbers — `Core.Tests` 7072/0/7073, `Interpreter.Tests` 609/0/609, `Compiler.Tests` filtered `~Emit` 3818/0/3818, `Issue2895` 7/7 — are true of **`main` `5ad0109c`** + head `b2765951`. `main` has since moved to `33bea4af` (#3047, #3024). I re-ran the **probe matrix** on the co-merged tree and it is unchanged, but I have **not** re-run the suites there: load is 42.06 and I am not building. Treat the suite numbers as SHA-tagged to `5ad0109c`; the behavioural results are current.

### Revised summary

- **BLOCKER:** none. I withdraw my earlier one.
- **SHOULD-FIX:** composite-literal form (`Expressions.cs:1188/1241/1327`) — pre-existing, worth its own issue, not a merge gate.
- **SHOULD-FIX:** consumer count dismissed by category.

Apologies for the churn. The fix, its pinning, and its test instrument are all good, and my objection was to a claim the body never actually made.

PR #3073, opened 2026-08-02
gsc: fold constant pattern switches during lowering

### LINKED ISSUE #2914: gsc: fold constant pattern switches during lowering
## Repro

```gsharp
package ConstantSwitchLowering

func F() int32 {
    switch 1 {
        case 1 { return 10 }
        case 2 { return 20 }
    }
}

public var result = F()
```

## Current

`GS0100: Not all code paths return a value.`

## Expected

Compile successfully; `result == 10`.

## Root cause

`Lowerer.Flatten` in `src/Core/CodeAnalysis/Lowering/Lowerer.cs` only flattens pattern-switch arm bodies and preserves the full runtime dispatch. `MethodBodyEmitter.Patterns.cs` therefore emits every arm plus the unmatched fallthrough.

Constant-discriminant or literal-guard pruning is sound only if lowering and emission share it. Folding only the definite-return CFG caused invalid IL during #2909 review because analysis removed paths that the emitter retained. Add any optimization in lowering, then let definite-return analyze the lowered shape.

Related: #2890, #2909. Distinct from exhaustive enum tracking in #2906.

DavidObando:
Fixes #2914

## Summary

- Fold pattern-switch statements during lowering when the discriminant has rewritten to a literal and arm selection can be resolved with literal constant patterns.
- Keep guarded, runtime-valued, captured, shared-field, call-based, and nonconstant-pattern switches on the existing dispatch path.
- Preserve emitted comparison semantics for `float` and `double`, including `NaN != NaN`.
- Update two definite-return tests whose old `GS0100` expectations represented paths that lowering can now prove unreachable.

## Artifact measurement

Measured from emitted IL for the issue shape:

| Shape | Before | After |
|---|---:|---:|
| `switch 1` with two constant arms and a default | 50 bytes / 4 branches | 3 bytes / 0 branches |

The regression test asserts the after values, loads the emitted assembly, calls `GetTypes()`, executes it, and checks results.

## Negative controls

| Control | Before | After | Runtime evidence |
|---|---:|---:|---|
| Mutable parameter | 50 bytes / 4 branches | 50 bytes / 4 branches | `2` selects `22` |
| Function call discriminant | 54 bytes / 4 branches | 54 bytes / 4 branches | call runs once; selects `22` |
| Shared mutable field | 54 bytes / 4 branches | 54 bytes / 4 branches | field value `2` selects `22` |
| Guarded constant pattern | 60 bytes / 5 branches | 60 bytes / 5 branches | guard runs once; fallback arm returns `22` |
| Mutable capture | 60 bytes / 4 branches | 60 bytes / 4 branches | post-capture mutation selects `22` |
| Relational pattern | 2 branches | 2 branches | returns `11` |
| Side effects in selected arm | full dispatch | 0 branches | selected call records `11`; other arms do not run |

No-match controls preserve behavior: a default returns `33`; no default plus a following return returns `33`; no default and no later return still reports `GS0100` and emits no PE.

## Diagnostics

Binding still runs before lowering. Tests retain:

- `GS0178` for a non-exhaustive enum switch statement.
- `GS0171` for an incompatible case type.
- `GS0100` when no matching/default arm and no following return can satisfy definite return.

`GS0100` intentionally disappears when folding proves the selected arm returns, or proves an impossible arm containing `break` cannot execute.

## Driver and position parity

| `gsc` top level | `gsc` function | `gsi` top level | `gsi` function |
|---:|---:|---:|---:|
| `11` | `22` | `11` | `22` |

`gsc` lowers both positions. `gsi` evaluates top-level code without lowering and lowers invoked function/lambda bodies; outputs remain identical.

## Mutation evidence

| Product mutation | Result | Test that killed it |
|---|---|---|
| Delete the complete lowering override | 4 failed / 3 passed | `LiteralSwitch_FoldsToSelectedArm_WhileNegativeControlsKeepDispatch` (plus issue, driver-position, and local-function tests) |
| Invert the fold gate | 4 failed / 3 passed | `LiteralSwitch_FoldsToSelectedArm_WhileNegativeControlsKeepDispatch` (`Guarded` returned `11`, expected `22`) |
| Fold through a nonconstant pattern instead of falling back | 1 failed | `LiteralSwitch_FoldsToSelectedArm_WhileNegativeControlsKeepDispatch` (`RelationalPattern`: 0 branches, expected 2) |
| Use `object.Equals` for `float` | 1 failed | `LiteralSwitch_FoldsToSelectedArm_WhileNegativeControlsKeepDispatch` (`31`, expected `32`) |
| Use `object.Equals` for `double` | 1 failed | `LiteralSwitch_FoldsToSelectedArm_WhileNegativeControlsKeepDispatch` (`41`, expected `42`) |

No product mutation survived. All mutant builds completed successfully before results were counted.

## Validation

- `dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true`: 0 warnings, 0 errors.
- Targeted issue tests: 7/7 passed.
- Related switch-flow tests: 23/23 passed.
- Full Release suite: 0 failed / 14,255 passed / 1 skipped.
- `git merge-tree`: conflict-free; merged tree equals tested HEAD tree.

### DIFF
--- a/src/Core/CodeAnalysis/Lowering/Lowerer.cs
+++ b/src/Core/CodeAnalysis/Lowering/Lowerer.cs
@@ -120,6 +120,49 @@ public sealed class Lowerer : BoundTreeRewriter
     }
 
     /// <inheritdoc/>
+    protected override BoundStatement RewritePatternSwitchStatement(BoundPatternSwitchStatement node)
+    {
+        var discriminant = RewriteExpression(node.Discriminant);
+        if (discriminant is not BoundLiteralExpression literal
+            || node.Arms.Any(arm => arm.Guard != null))
+        {
+            return base.RewritePatternSwitchStatement(node);
+        }
+
+        BoundPatternSwitchArm defaultArm = null;
+        foreach (var arm in node.Arms)
+        {
+            if (arm.IsDefault)
+            {
+                defaultArm = arm;
+                continue;
+            }
+
+            if (RewritePattern(arm.Pattern) is not BoundConstantPattern
+                {
+                    Value: BoundLiteralExpression patternLiteral,
+                })
+            {
+                return base.RewritePatternSwitchStatement(node);
+            }
+
+            var matches = literal.Type == TypeSymbol.Float32
+                ? (float)literal.Value == (float)patternLiteral.Value
+                : literal.Type == TypeSymbol.Float64
+                    ? (double)literal.Value == (double)patternLiteral.Value
+                    : Equals(literal.Value, patternLiteral.Value);
+            if (matches)
+            {
+                return RewriteStatement(arm.Body);
+            }
+        }
+
+        return defaultArm == null
+            ? new BoundBlockStatement(null, ImmutableArray<BoundStatement>.Empty)
+            : RewriteStatement(defaultArm.Body);
+    }
+
+    /// <inheritdoc/>
     /// <remarks>
     /// Issue #419 (P0-2): rewrites <c>return</c> statements that appear
     /// lexically inside a try / catch / finally region. The original
--- a/test/Compiler.Tests/Emit/Issue2914ConstantPatternSwitchLoweringTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2914ConstantPatternSwitchLoweringTests.cs
@@ -1,0 +1,439 @@
+// <copyright file="Issue2914ConstantPatternSwitchLoweringTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Emit;
+using GSharp.Core.CodeAnalysis;
+using GSharp.Core.CodeAnalysis.Compilation;
+using GSharp.Core.CodeAnalysis.Symbols;
+using GSharp.Core.CodeAnalysis.Syntax;
+using GSharp.Core.CodeAnalysis.Text;
+using Xunit;
+
+namespace GSharp.Compiler.Tests.Emit;
+
+/// <summary>Issue #2914: constant pattern switches fold during lowering.</summary>
+public class Issue2914ConstantPatternSwitchLoweringTests
+{
+    private const string ArtifactSource = """
+        package Issue2914.Artifacts
+
+        import System
+
+        class Holder {
+            shared {
+                public var Value int32 = 2
+            }
+        }
+
+        enum Choice { A, B }
+
+        func ReadValue() int32 {
+            calls = calls + 1
+            return 2
+        }
+
+        func GuardAllows() bool {
+            calls = calls + 1
+            return false
+        }
+
+        func Constant() int32 {
+            switch 1 {
+                case 1 { return 11 }
+                case 2 { return 22 }
+                default { return 33 }
+            }
+        }
+
+        func Mutable(value int32) int32 {
+            switch value {
+                case 1 { return 11 }
+                case 2 { return 22 }
+                default { return 33 }
+            }
+        }
+
+        func ThroughCall() int32 {
+            switch ReadValue() {
+                case 1 { return 11 }
+                case 2 { return 22 }
+                default { return 33 }
+            }
+        }
+
+        func SharedField() int32 {
+            switch Holder.Value {
+                case 1 { return 11 }
+                case 2 { return 22 }
+                default { return 33 }
+            }
+        }
+
+        func Guarded() int32 {
+            switch 1 {
+                case 1 when (GuardAllows()) { return 11 }
+                case 1 { return 22 }
+                default { return 33 }
+            }
+        }
+
+        func RecordArm(value int32) int32 {
+            armTrace = armTrace + value
+            return value
+        }
+
+        func ArmSideEffects() int32 {
+            switch 1 {
+                case 1 { return RecordArm(11) }
+                case 2 { return RecordArm(22) }
+                default { return RecordArm(33) }
+            }
+        }
+
+        func NoMatchDefault() int32 {
+            switch 9 {
+                case 1 { return 11 }
+                case 2 { return 22 }
+                default { return 33 }
+            }
+        }
+
+        func NoMatchWithoutDefault() int32 {
+            switch 9 {
+                case 1 { return 11 }
+                case 2 { return 22 }
+            }
+
+            return 33
+        }
+
+        func ExhaustiveEnum() int32 {
+            switch Choice.A {
+                case Choice.A { return 11 }
+                case Choice.B { return 22 }
+            }
+        }
+
+        func RelationalPattern() int32 {
+            switch 1 {
+                case > 0 { return 11 }
+                default { return 33 }
+            }
+        }
+
+        func Float32NaN() int32 {
+            switch Single.NaN {
+                case Single.NaN { return 31 }
+                default { return 32 }
+            }
+        }
+
+        func Float64NaN() int32 {
+            switch Double.NaN {
+                case Double.NaN { return 41 }
+                default { return 42 }
+            }
+        }
+
+        var calls = 0
+        var armTrace = 0
+        public var constantResult = Constant()
+        public var mutableResult = Mutable(2)
+        public var callResult = ThroughCall()
+        public var sharedResult = SharedField()
+        public var guardedResult = Guarded()
+        public var sideEffectResult = ArmSideEffects()
+        public var noMatchResult = NoMatchDefault()
+        public var noMatchWithoutDefaultResult = NoMatchWithoutDefault()
+        public var exhaustiveResult = ExhaustiveEnum()
+        public var relationalResult = RelationalPattern()
+        public var float32NaNResult = Float32NaN()
+        public var float64NaNResult = Float64NaN()
+        public var callCount = calls
+        public var armTraceResult = armTrace
+        """;
+
+    [Fact]
+    public void LiteralSwitch_FoldsToSelectedArm_WhileNegativeControlsKeepDispatch()
+    {
+        var assembly = CompileAndRun(ArtifactSource);
+        var program = GetProgram(assembly);
+
+        Assert.Equal(11, GetField(program, "constantResult"));
+        Assert.Equal(22, GetField(program, "mutableResult"));
+        Assert.Equal(22, GetField(program, "callResult"));
+        Assert.Equal(22, GetField(program, "sharedResult"));
+        Assert.Equal(22, GetField(program, "guardedResult"));
+        Assert.Equal(11, GetField(program, "sideEffectResult"));
+        Assert.Equal(33, GetField(program, "noMatchResult"));
+        Assert.Equal(33, GetField(program, "noMatchWithoutDefaultResult"));
+        Assert.Equal(11, GetField(program, "exhaustiveResult"));
+        Assert.Equal(11, GetField(program, "relationalResult"));
+        Assert.Equal(32, GetField(program, "float32NaNResult"));
+        Assert.Equal(42, GetField(program, "float64NaNResult"));
+        Assert.Equal(2, GetField(program, "callCount"));
+        Assert.Equal(11, GetField(program, "armTraceResult"));
+
+        AssertMethodShape(program, "Constant", expectedBytes: 3, expectedBranches: 0);
+        AssertMethodShape(program, "NoMatchDefault", expectedBytes: 3, expectedBranches: 0);
+        AssertMethodShape(program, "NoMatchWithoutDefault", expectedBytes: 3, expectedBranches: 0);
+        AssertMethodShape(program, "ExhaustiveEnum", expectedBytes: 3, expectedBranches: 0);
+        AssertMethodShape(program, "Float32NaN", expectedBytes: 3, expectedBranches: 0);
+        AssertMethodShape(program, "Float64NaN", expectedBytes: 3, expectedBranches: 0);
+        Assert.Equal(0, GetMethodBranches(program, "ArmSideEffects"));
+
+        AssertMethodShape(program, "Mutable", expectedBytes: 50, expectedBranches: 4);
+        AssertMethodShape(program, "ThroughCall", expectedBytes: 54, expectedBranches: 4);
+        AssertMethodShape(program, "SharedField", expectedBytes: 54, expectedBranches: 4);
+        AssertMethodShape(program, "Guarded", expectedBytes: 60, expectedBranches: 5);
+        Assert.Equal(2, GetMethodBranches(program, "RelationalPattern"));
+    }
+
+    [Fact]
+    public void IssueRepro_CompilesRunsAndHasNoDispatch()
+    {
+        const string Source = """
+            package Issue2914.Repro
+
+            func F() int32 {
+                switch 1 {
+                    case 1 { return 10 }
+                    case 2 { return 20 }
+                }
+            }
+
+            public var result = F()
+            """;
+
+        var assembly = CompileAndRun(Source);
+        var program = GetProgram(assembly);
+
+        Assert.Equal(10, GetField(program, "result"));
+        AssertMethodShape(program, "F", expectedBytes: 3, expectedBranches: 0);
+    }
+
+    [Fact]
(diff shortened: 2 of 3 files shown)


DavidObando:
## Rubber-duck verdict: **MERGEABLE** — 1 SHOULD-FIX, no blockers

Anchors: `main` **`33bea4af`**, head **`8e2e802f8`**, computed merge base `5ad0109c` (one merge stale — everything below is measured at `33bea4af`), merged tip **`74e4325931`**. `merge-tree` rc=0, merge rc=0, 0 conflicts; merged-vs-`main` diff is exactly the PR's 3 files. Both trees built rc=0 / 0 warnings, all three driver copies (`Core/`, `Compiler/`, `Repl/`) fresh to the same second, and the two trees' `GSharp.Core.dll` md5s **differ** (`65c69525` base / `49dc5719` head) so "identical output" is a meaningful result.

---

## Lead item: the two changed definite-return tests are JUSTIFIED — measured, not accepted

Both flip `AssertOnlyGs0100` → `AssertNoErrors`, i.e. programs `main` **rejects** now **compile**. I wrote both sources myself and ran them on base and head.

**1. `LiteralSwitchWithoutDefault`** (`:208`)
```
func F() int32 { switch 1 { case 1 { return 10 } case 2 { return 20 } } }
```
| | base `33bea4af` | head |
|---|---|---|
| `gsc` | rc=1 `GS0100` | **rc=0, no diagnostics** |

The discriminant is `1`, `case 1` matches, the function **always returns 10**. There is no path that falls off the end. Base's `GS0100` was a **false positive**; accepting is correct.

**2. `ImpossibleLiteralArmBreak`** (`:226`)
```
func F() int32 { for { switch 1 { case 2 { break } } } }
```
| | base | head |
|---|---|---|
| `gsc` | rc=1 `GS0100` | **rc=0** |

The only `break` sits in an arm that can never match, so the `for` never terminates and `F` never returns — no return statement is required, same rule as `while(true){}`. Correct.

**What is no longer asserted:** `AssertOnlyGs0100` (`:470`) checks five things `AssertNoErrors` does not — the exact ID, span start at `F(`, span length 1, span text `"F"`, and `emittedLength == 0`. But the helper is **not orphaned**: it still has **13 call sites** on head (15 on base), so the location logic stays pinned by the eleven-plus tests that still produce the diagnostic. Nothing is left unasserted that should be asserted.

### The inverse — does folding SUPPRESS a `GS0100` a user should still get? No.

| probe | base | head |
|---|---|---|
| `switch 1 { case 2 { return 20 } }` — no match, no default | `GS0100` | **`GS0100`** |
| `for { switch 1 { case 1 { break } } }` — arm *matches*, break exits loop, falls off end | `GS0100` | **`GS0100`** |
| `switch 1 { case 2 { return 20 } default { } }` | `GS0100` | **`GS0100`** |
| nested fold, inner arm unreachable | `GS0100` | **`GS0100`** |
| local assigned only in a folded-away arm (definite assignment) | rc=0 | rc=0 |
| dead `return` after an always-returning folded switch (new-warning check) | rc=0 | rc=0 |

No suppression, no new diagnostics on previously-valid code. Both directions clean.

---

## The finding: one branch of the new fold is pinned by nothing

I mutated each guard independently, asserting the patch applied and checking `BUILD_RC` before believing any result.

| mutant | `Lowerer.cs` site | Failed / 7 | verdict |
|---|---|---|---|
| **X1** remove the `arm.Guard != null` bail | `:124` | 1 | pinned — `LiteralSwitch_Folds…` |
| **X3** drop the default arm's body on no-match | `:158` | 1 | pinned — `LiteralSwitch_Folds…` |
| **X2** default arm wins immediately (ordering ignored) | `:130` | **0** | **SURVIVED** |

**X2 survives 251 tests** — Issue2914 **7/7**, Issue2890 **23/23**, all Core.Tests `~Switch` **221/221**.

It is not an equivalent mutant. Built rc=0, driver live and fresh (md5 `5d890577` ≠ clean `49dc5719`), and it produces a **silent wrong answer at rc=0**:

```
switch 1 { default { print "armdef" }   case 1 { print "arm11" } }

base 33bea4af : arm11      head (clean) : arm11      head + X2 : armdef
```

The shipped code is **correct** — a `default` declared before a matching case correctly loses, in both `gsc` and `gsi`. It is simply untested. Given that #2963 shipped a source break through 46/46 green precisely because its guard was pinned by no test, this is worth one `[Fact]`.

**SHOULD-FIX:** add a case with `default` declared before a matching arm (and ideally one with `default` between two cases — I verified `switch 3 { case 1 … default … case 2 … }` correctly yields the default on both sides).

---

## Independent verification of the IL table (measured at `33bea4af`, not the stale base)

I built my own `System.Reflection.Metadata` IL-size reader rather than trust the table, compiled the PR's own `ArtifactSource` on both trees, and reproduced it **exactly**:

| method | base | head | | method | base | head |
|---|---:|---:|---|---|---:|---:|
| `Constant` | **50** | **3** | | `Mutable` | **50** | **50** |
| `NoMatchDefault` | 51 | 3 | | `ThroughCall` | **54** | **54** |
| `NoMatchWithoutDefault` | 40 | 3 | | `SharedField` | **54** | **54** |
| `ExhaustiveEnum` | 47 | 3 | | `Guarded` | **60** | **60** |
| `Float32NaN` | 41 | 3 | | `RelationalPattern` | 33 | 33 |
| `Float64NaN` | 49 | 3 | | `ArmSideEffects` | 65 | 8 |

Every negative control holds at its claimed before-value on current `main`. The table is accurate.

## Semantics of the folded arm — preserved, each with a fold positive-control

| probe | base (gsc / gsi) | head (gsc / gsi) | fold engaged? |
|---|---|---|---|
| `break` in folded arm inside a loop | `done-1` / `done-1` | `done-1` / `done-1` | — |
| only the matching arm runs, once, then continues | `arm11,after` | `arm11,after` | **71 → 21** |
| arm-local shadowing an outer local doesn't leak | `22,11` | `22,11` | **54 → 38** |
| `default` before a matching case | `arm11` | `arm11` | yes |
| `char` / `string` / `bool` / `double` discriminants | `11` | `11` | 35→3 / 44→3 / 33→3 / 49→3 |

**A caught trap worth naming:** my `int64` probe returned `11` on both sides and looked like a clean pass — but the positive control showed **44 → 44 bytes, i.e. the fold never engaged**. `switch int64(1)` is a conversion, not a literal, so it stays on the dispatch path. That result was a *scope zero, not an absence*, and I discarded it. `-0.0` and mismatched literal kinds (`switch 1.0 { case 1 … }`, `switch Single.Epsilon { case 0 … }`) likewise decline to fold — which is why the unchecked `(float)`/`(double)` unboxing casts at `:147-150` are never reached with a mismatched box. I probed for an `InvalidCastException` there specifically and could not produce one.

**Hazard checked and cleared:** `Lowerer` is stateful (`labelCount`, `tryNestingDepth`, `returnValueLocal` at `:26-40`), and the new override calls `RewriteExpression(node.Discriminant)` then **discards it** and re-rewrites via `base.…` on every fallback. I probed for leaked temps/labels with a short-circuit-ternary discriminant (72 → 72) and a literal discriminant with relational patterns, which double-calls `RewritePattern` (50 → 50). **Byte-identical — benign.** Reporting it as a non-finding rather than a mechanism, since the probe never fired.

---

## Driver columns: an incidental fix, and a pre-existing defect it does not reach

The author's test shape is good: `func Main()` count **0** against a positive control of 8 `package` matches — every case is top-level or a real function, so `gsi` genuinely executes them.

Running all three columns surfaced a **pre-existing `gsc`/`gsi` disagreement on `main`**:

| shape | base gsc | base gsi | head gsc | head gsi |
|---|---|---|---|---|
| `Single.NaN` / `Double.NaN`, constant discriminant | `32,42` | **`31,41`** | `32,42` | **`32,42`** |
| `Double.NaN` via **call** discriminant (never folds, 68→68) | `42` | **`41`** | `42` | **`41`** |

`gsc` is right (`NaN != NaN`). On `main` the interpreter says NaN **matches** NaN, at rc=0, silently. **This PR incidentally fixes that wherever folding reaches** — but the interpreter's own constant-pattern comparison is still wrong for non-folding discriminants, on head. Pre-existing, not introduced here, and not a blocker for this PR.

---

## Items

- **BLOCKER**: none.
- **SHOULD-FIX** — `Lowerer.cs:130`, default-arm ordering survives 251 tests (X2 above). Add a `default`-before-matching-case fact.
- **NIT** — `TopLevelAndFunctionSwitches_AgreeAcrossEmitAndInterpreter` (`:222`) only compares `gsc`/`gsi` on integer shapes. Adding the NaN shapes would pin the interpreter divergence this PR actually fixes — it fails on base and passes on head.
- **NIT** — the interpreter's non-folded NaN comparison is still wrong on head (table above). Worth its own issue; out of scope here.

## Scope

Ran: full **Core.Tests 7072 passed / 1 skipped / 0 failed** on the merged tree (chosen because a `Lowerer` change is language-wide), Issue2914 **7/7**, Issue2890 **23/23**, Core.Tests `~Switch` **221/221**. **Did not** run full Compiler.Tests (~1h20m, load was 25-36 throughout and above 30 at two checkpoints, where I paused rather than build). Restore verified: `git status --porcelain` clean of tracked modifications and `GSharp.Core.dll` md5 round-tripped **exactly** to `49dc5719…` on all three driver copies.

DavidObando:
### Addendum — issue #2914's verbatim repro, both halves

My earlier verdict verified the *compile* half of #2914's stated expectation. The issue's "Expected" has two halves — "Compile successfully; **`result == 10`**" — so here is the issue's repro run verbatim, unmodified, on both trees:

```gsharp
package ConstantSwitchLowering

func F() int32 {
    switch 1 {
        case 1 { return 10 }
        case 2 { return 20 }
    }
}

public var result = F()
```

| | base `33bea4af` | head (merged `74e4325931`) |
|---|---|---|
| `gsc /out:` | rc=1 `GS0100` | **rc=0** |
| `gsi` | rc=1 `GS0100` | **rc=0** |
| `result` | — (rejected) | **`10`** on both drivers |

Base reproduces the issue's "Current" exactly (`GS0100: Not all code paths return a value.`); head delivers the issue's "Expected" including the value, not only the compile. Verdict unchanged: **MERGEABLE**, with the one SHOULD-FIX (default-arm ordering unpinned, `Lowerer.cs:130`).

DavidObando:
## Re-review — re-anchored on `main` = `4f7559dd` (previous verdict was true of `33bea4af`)

`main` moved four commits under my earlier verdict, and one of them lands in this PR's blast radius: **#2963 `Invalidate narrowing across loop back-edges`** (`ea35b2e6`) rewrote `StatementBinder.Jumps.cs` (+113) and `StatementBinder.Loops.cs` (+50). This PR's flipped test `ImpossibleLiteralArmBreak` is literally a loop containing a switch containing a `break`. Same flow surface, zero file overlap — so the file-overlap argument proves nothing and I re-measured.

**Verdict: MERGEABLE.** Unchanged from round 1. 0 BLOCKER, 1 SHOULD-FIX, 2 NIT.

### Anchors
| | |
|---|---|
| `main` | `4f7559dd` |
| PR head | `8e2e802f8` (unmoved) |
| merge base | `5ad0109c` (three-dot, `--full-index`) |
| PR content | +486 / −4, 3 files — **unchanged** |
| merged tip | `6fa1605f`, tree `90f09db5` |
| `merge-tree` prediction | `90f09db5` — **exact match**, 0 conflicts |
| **`BUILD_RC`** | **0**, 0 warnings, 0 errors |
| base driver md5 | `9de121ec` |
| merged driver md5 | `4de56f96` (distinct — "same output" is meaningful) |
| driver mtimes | `Core/` `Compiler/` `Repl/` all 06:34:13, `gsc` 06:34:16, `gsi` 06:34:17 |

`merge-tree` clean is a floor; the build is what settles it, and it is green at `4f7559dd`.

### Interaction with #2963 — measured, not inferred

Neither PR's tests cover the overlap cell:

| corpus | loops | G# switches | narrowing |
|---|---|---|---|
| #2963's new tests (1,087 lines) | yes | **0** | yes |
| this PR's new tests (439 lines) | **0** | yes | **0** |

(The 5 `switch` hits in #2963's file are C# `shape switch` harness expressions, not G# source. Positive control: the same grep returns 13 on this PR's test file.)

So I wrote the missing cell. Discriminating pair, identical programs except the discriminant:

- **`n1`** — `switch 1 { case 1 { c = nil } }` inside a narrowed loop → **folds**
- **`n4`** — `switch D() { case 1 { c = nil } }`, otherwise identical → **does not fold**

| probe | base `4f7559dd` | merged `6fa1605f` |
|---|---|---|
| n1 (folds) | `GS0159` — EVAL / EMIT / GSI | `GS0159` — EVAL / EMIT / GSI |
| n4 (no fold) | `GS0159` — EVAL / EMIT / GSI | `GS0159` — EVAL / EMIT / GSI |

**n1 ≡ n4 exactly.** Folding does not suppress #2963's back-edge invalidation — the unsoundness direction is clean.

False-positive corpus (correct programs that must keep compiling), rc + diagnostic set, all three drivers:

| probe | base | merged |
|---|---|---|
| `n3` narrowed loop, foldable switch, no mutation | rc0 `33\|11\|33\|11\|33\|11` | **identical** |
| `brk` `break` inside a folded arm inside a loop | rc0 `done-1` | **identical** |
| `n2` reassign to non-nil in folded arm | `GS0159` | **identical** (pre-existing on `main`) |

**Positive control that the fold actually engaged** (otherwise "identical" is scope, not absence), IL bytes:

| probe | base | merged |
|---|---|---|
| `brk` `<Main>$` | 95 bytes / 6 branch | **79 / 3** |
| `n3` `<Main>$` | 84 bytes / 6 branch | **68 / 3** |

The transformation is real inside loops; the agreement is real agreement. And `break` in a folded arm still targets the enclosing loop (`done-1`) after #2963's +113 jumps rewrite.

### Lead item — the two flipped definite-return tests, re-measured on `4f7559dd`

| | base `4f7559dd` | merged `6fa1605f` |
|---|---|---|
| `t1` `switch 1 { case 1 {return 10} case 2 {return 20} }` | rc1 `GS0100` ×3 drivers | rc0 ×3 drivers |
| `t2` `for { switch 1 { case 2 { break } } }` | rc1 `GS0100` ×3 drivers | rc0 ×3 drivers |

Both were false positives: `t1` always returns 10; `t2` is an infinite loop whose only `break` is unreachable. The flip is justified and reproduces on new main. `AssertOnlyGs0100` is **not orphaned** — 13 remaining call sites.

### Co-merged suites — behavioural disjointness, not a file list

| suite | project | result |
|---|---|---|
| `~Issue2914` | Compiler.Tests | **7/7** |
| `~Issue2890` | Core.Tests | **23/23** |
| `~Issue2943` (#2963's own) | Compiler.Tests | **56/56** |
| `~Issue2943` | Core.Tests | **1/1** |
| `~Issue2943` | Interpreter.Tests | **9/9** |

#2963's tests span **three projects**; a Compiler.Tests-only filter would have reached 56 of 66 and looked exhaustive. Naming that because it is the exact trap that bit someone today.

### SHOULD-FIX (carried, unchanged) — `src/Core/CodeAnalysis/Lowering/Lowerer.cs:130`

The default-arm branch `if (arm.IsDefault) { defaultArm = arm; continue; }` is **unpinned**. Product mutant "default arm wins immediately" (ignore ordering) built rc=0, driver live, and **survived 251 tests** — while silently returning the wrong arm: `switch 1 { default {armdef} case 1 {arm11} }` gives `arm11` correct, `armdef` mutated. Shipped code is **correct**; it is the test gap that is the finding. One test with a `default` arm placed before a matching constant case closes it.

**Residual I did not re-run:** that mutation was measured on `33bea4af` (251 tests). I did not re-run it on `4f7559dd` — load hit 44.35 and I stopped building. #2963's new tests contain zero G# switches, so they cannot newly pin a switch branch, but I am flagging that as inference, not measurement.

### NIT
1. `Issue2914ConstantPatternSwitchLoweringTests.cs:222` cross-driver agreement test covers integer shapes only — no NaN case, though `NaN` handling is a headline claim.
2. Pre-existing, **not** this PR: `gsi` evaluates `Double.NaN`-discriminant comparisons wrongly when the switch does **not** fold (`gsc` 42 / `gsi` 41 on base *and* head). This PR incidentally fixes it wherever folding reaches. Worth its own issue.

### Not run
Full Compiler.Tests (~1h20m) on the merged tree — load was 20–44 throughout and above 30 at three checkpoints. A filtered green is not a green; that is the gap.

PR #3074, opened 2026-08-02
Fix explicit constructor accessibility emission

### LINKED ISSUE #2893: Explicit class constructor accessibility is emitted as public

An explicit `internal init()` is emitted as a public CLR constructor:

```gs
package FindingConstructorAccessibility

public class InternalConstructor {
  internal init() {
  }
}

func Main() int32 {
  let constructors = typeof(InternalConstructor).GetConstructors(
    System.Reflection.BindingFlags.Instance |
    System.Reflection.BindingFlags.Public |
    System.Reflection.BindingFlags.NonPublic)
  return constructors.Length == 1 && constructors[0].IsAssembly && !constructors[0].IsPublic ? 0 : 1
}
```

Expected: the build succeeds and `Main` returns `0` because the constructor has assembly visibility and is not public.

Actual: the build succeeds and `Main` returns `1`. Reflection reports `IsPublic=true` and `IsAssembly=false`.

ILSpy shows `.method public ... .ctor()`. Equivalent C# emits `.method assembly ... .ctor()` and returns `0`.

DavidObando:
Fixes #2893

## Root cause and fix

`DeclarationBinder` already preserves the declared constructor accessibility on `ConstructorSymbol.Function.Accessibility`. `TypeDefEmitter.EmitClassConstructorWithBody` discarded it and always emitted `MethodAttributes.Public`.

The emitter now reuses the existing canonical `AccessibilityMap.ToMethodVisibility` mapping. No new accessibility mapper or diagnostic was added.

## Accessibility and declaration-context matrix

All emitted results were read through `Assembly.Load(File.ReadAllBytes(...))`, `Assembly.GetTypes()`, and `ConstructorInfo.Attributes`/visibility properties. The emitted executable also ran in a child `dotnet exec` process and printed `11`, `22`, `33`, `34`, `44`.

| Source shape | Before | After | Expected |
|---|---|---|---|
| top-level `public init` | `Public` | `Public` | `Public` |
| top-level `internal init` | `Public` | `Assembly` | `Assembly` |
| top-level `protected init` | `Public` | `Family` | `Family` |
| top-level `private init` | `Public` | `Private` | `Private` |
| top-level `init` with no modifier | `Public` | `Public` | `Public` |
| `internal func init` alternate syntax | `Public` | `Assembly` | `Assembly` |
| `private convenience init` | `Public` | `Private` | `Private` |
| nested-class `private init` | `Public` | `Private` | `Private` |
| generic-class `private init` | `Public` | `Private` | `Private` |
| data-class `private init` | `Public` | `Private` | `Private` |
| plain-struct `private init` | `Public` | `Private` | `Private` |
| synthesized default constructor | `Public` | `Public` | `Public` |
| primary constructor | `Public` | `Public` | `Public` |
| shared static initializer `.cctor` | `Private, Static` | `Private, Static` | `Private, Static` |
| interface constructor | parser rejects with `GS0005` | unchanged | not expressible |
| data/inline-struct explicit constructor | parser rejects with `GS0005` | unchanged | not expressible |

`gsi` and `gsc` agree: public/internal/default construction produced `11/22/33`; external private construction produced `GS0472`; external protected construction produced `GS0379`.

## Emitter-site denominator

I temporarily added a required parameter to `EmitClassConstructorWithBody`. The compiler reported `CS7036` at **2 unique source call sites**:

1. class explicit constructors: `ReflectionMetadataEmitter.cs:3367`
2. plain-struct explicit constructors: `ReflectionMetadataEmitter.cs:3541`

The solution build printed 8 duplicated diagnostic lines across build targets, but the counted population is 2 unique source sites. Both flow through the single fixed attribute-setting hunk.

## Mutation testing

| Product hunk | Build-green mutant | Detection | Guard | Restore |
|---|---|---|---|---|
| explicit constructor `MethodAttributes` | replaced accessibility mapping with `MethodAttributes.Public` | `NonPublicExplicitConstructors_EmitDeclaredAccessibility` failed: expected `Assembly`, actual `Public` | `PublicDefaultAndStaticConstructors_RemainUnchanged` passed | full rebuild restored clean Core DLL hash; both tests passed 2/2 |

The mutant build succeeded with 0 warnings/errors and changed `GSharp.Core.dll`; restoration rebuilt again and round-tripped to the clean hash.

## Verification

- Latest merge base: `5ad0109c04608f627e35d797d1735ca21d8311f2`
- `git merge-tree --write-tree origin/main HEAD`: exit 0; merged tree matched `HEAD` tree (1/1 comparison)
- Release CI build: 0 warnings, 0 errors; five driver/test Core DLL copies fresh together; 0 zero-byte Core DLLs
- Targeted tests: failed 1/1 on unfixed product; passed 2/2 on HEAD
- Full `Compiler.Tests`: **0 failed, 4047 passed, 0 skipped, 4047 total**
- Full `Cs2Gs.Tests`: **0 failed, 1863 passed, 0 skipped, 1863 total**
- cs2gs corpus gate: 19/20 green; known baseline 1, new 0, regressed 0, stale 0; `G11-Linq-Console` passed all stages

### DIFF
--- a/src/Core/CodeAnalysis/Emit/TypeDefEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/TypeDefEmitter.cs
@@ -1695,7 +1695,7 @@ internal sealed class TypeDefEmitter
         var firstCtorParamHandle = this.AddRefKindAwareParameterRows(function.Parameters, out var ctorParamHandles);
 
         var ctorHandle = this.emitCtx.Metadata.AddMethodDefinition(
-            attributes: MethodAttributes.Public | MethodAttributes.HideBySig | MethodAttributes.SpecialName
+            attributes: AccessibilityMap.ToMethodVisibility(function.Accessibility) | MethodAttributes.HideBySig | MethodAttributes.SpecialName
                 | MethodAttributes.RTSpecialName,
             implAttributes: MethodImplAttributes.IL | MethodImplAttributes.Managed,
             name: this.emitCtx.Metadata.GetOrAddString(".ctor"),
--- a/test/Compiler.Tests/Emit/Issue2893ConstructorAccessibilityEmitTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2893ConstructorAccessibilityEmitTests.cs
@@ -1,0 +1,241 @@
+// <copyright file="Issue2893ConstructorAccessibilityEmitTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using Xunit;
+
+namespace GSharp.Compiler.Tests.Emit;
+
+/// <summary>Regression coverage for issue #2893.</summary>
+public sealed class Issue2893ConstructorAccessibilityEmitTests
+{
+    private const string Source = """
+        package Issue2893
+        import System
+
+        class PublicCtor {
+            var Value int32
+            public init(value int32) { Value = value }
+        }
+
+        class InternalCtor {
+            internal init(value int32) {}
+        }
+
+        open class ProtectedCtor {
+            protected init(value int32) {}
+        }
+
+        class PrivateCtor {
+            private init(value int32) {}
+        }
+
+        class DefaultCtor {
+            var Value int32
+            init(value int32) { Value = value }
+        }
+
+        class FuncInitInternal {
+            internal func init(value int32) {}
+        }
+
+        class ConvenienceCtor {
+            public init(value int32, other int32) {}
+            private convenience init(value int32) { init(value, 0) }
+        }
+
+        class Outer {
+            class NestedPrivate {
+                private init(value int32) {}
+            }
+        }
+
+        class GenericPrivate[T] {
+            private init(value int32) {}
+        }
+
+        data class DataPrivate {
+            private init(value int32) {}
+        }
+
+        struct StructPrivate {
+            private init(value int32) {}
+        }
+
+        class SynthesizedDefault {
+            var Value int32 = 33
+        }
+
+        class PrimaryCtor(value int32) {}
+
+        class StaticInitializer {
+            shared {
+                var Value int32
+                init { Value = 44 }
+                func Read() int32 -> Value
+            }
+        }
+
+        func Main() {
+            Console.WriteLine(PublicCtor(11).Value)
+            Console.WriteLine(DefaultCtor(22).Value)
+            Console.WriteLine(SynthesizedDefault().Value)
+            Console.WriteLine(PrimaryCtor(34).value)
+            Console.WriteLine(StaticInitializer.Read())
+        }
+        """;
+
+    [Fact]
+    public void NonPublicExplicitConstructors_EmitDeclaredAccessibility()
+    {
+        Verify(types =>
+        {
+            AssertConstructorVisibility(types, "Issue2893.InternalCtor", MethodAttributes.Assembly, typeof(int));
+            AssertConstructorVisibility(types, "Issue2893.ProtectedCtor", MethodAttributes.Family, typeof(int));
+            AssertConstructorVisibility(types, "Issue2893.PrivateCtor", MethodAttributes.Private, typeof(int));
+            AssertConstructorVisibility(types, "Issue2893.FuncInitInternal", MethodAttributes.Assembly, typeof(int));
+            AssertConstructorVisibility(types, "Issue2893.ConvenienceCtor", MethodAttributes.Private, typeof(int));
+            AssertConstructorVisibility(types, "Issue2893.Outer+NestedPrivate", MethodAttributes.Private, typeof(int));
+            AssertConstructorVisibility(types, "Issue2893.GenericPrivate`1", MethodAttributes.Private, typeof(int));
+            AssertConstructorVisibility(types, "Issue2893.DataPrivate", MethodAttributes.Private, typeof(int));
+            AssertConstructorVisibility(types, "Issue2893.StructPrivate", MethodAttributes.Private, typeof(int));
+        });
+    }
+
+    [Fact]
+    public void PublicDefaultAndStaticConstructors_RemainUnchanged()
+    {
+        Verify(types =>
+        {
+            AssertConstructorVisibility(types, "Issue2893.PublicCtor", MethodAttributes.Public, typeof(int));
+            AssertConstructorVisibility(types, "Issue2893.DefaultCtor", MethodAttributes.Public, typeof(int));
+            AssertConstructorVisibility(types, "Issue2893.SynthesizedDefault", MethodAttributes.Public);
+            AssertConstructorVisibility(types, "Issue2893.PrimaryCtor", MethodAttributes.Public, typeof(int));
+
+            var staticInitializer = types.Single(type => type.FullName == "Issue2893.StaticInitializer").TypeInitializer;
+            Assert.NotNull(staticInitializer);
+            Assert.Equal(
+                MethodAttributes.Private,
+                staticInitializer.Attributes & MethodAttributes.MemberAccessMask);
+            Assert.True(staticInitializer.IsStatic);
+        });
+    }
+
+    private static void Verify(Action<Type[]> assertions)
+    {
+        var directory = Path.Combine(
+            AppContext.BaseDirectory,
+            nameof(Issue2893ConstructorAccessibilityEmitTests),
+            Guid.NewGuid().ToString("N"));
+        Directory.CreateDirectory(directory);
+        try
+        {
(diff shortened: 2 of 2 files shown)


DavidObando:
## Review: PR #3074 — `fix/2893-ctor-accessibility`

**Verdict: MERGEABLE** (2 SHOULD-FIX, 1 NIT — no blockers)

`main` **`33bea4af`**, head **`6dbac07e4`**, computed merge base **`5ad0109c`**, three-dot content **2 files, +242 / −1**. `git merge-tree` → 0 conflicts, tree **`96059d7e`**; I built that exact tree and the SHA matched the prediction. Merged build Release **0/0**; freshness verified on `Core/`, `Compiler/` **and `Repl/`** (md5 `efb23c9c`, identical across all three).

### Lead: the CI premise is wrong — there is no failing check
I was asked to establish a CI failure and stop early if it was real. **There was never one.**

| measured | pass | pending | **fail** | state |
|---|---:|---:|---:|---|
| 02:56 PDT | 24 | 5 | **0** | UNSTABLE |
| 03:47 PDT | 29 | 2 skipping | **0** | **CLEAN** |

`UNSTABLE` was **in-progress required shards** (the run was `in_progress`, started 09:45:43Z), not a failure. The run has since completed and `mergeStateStatus` is now **`CLEAN`**. No BLOCKER from CI.

### The production change is one line
`src/Core/CodeAnalysis/Emit/TypeDefEmitter.cs:1698` — `MethodAttributes.Public` → `AccessibilityMap.ToMethodVisibility(function.Accessibility)`. Reuses the canonical mapper, as claimed. (It uses the 1-arg overload, not the `isTopLevelProgramMember` one; for ctors `IsTopLevelProgramMember` is false, so the two agree — correct choice.)

### Constructor-emit site denominator: 7 definition sites, exactly 1 changed — and it is the right one
Enumerated by `AddMethodDefinition` + `RTSpecialName`, cross-checked against `GetOrAddString(".ctor")` (42 hits, mostly MemberRefs, not definitions):

| site | emits | still hardcodes `Public`? |
|---|---|---|
| `TypeDefEmitter.cs:1407` `EmitClassDefaultConstructor` | synthesized default | yes — **correct** |
| `TypeDefEmitter.cs:1528` `EmitClassPrimaryConstructor` | `class C(x T)` | yes — **correct** |
| `TypeDefEmitter.cs:1587` `EmitClassConstructorWithBaseInitializer` | — | yes |
| **`TypeDefEmitter.cs:1698` `EmitClassConstructorWithBody`** | user-declared `init` | **FIXED** |
| `TypeDefEmitter.cs:1775` `EmitDefaultConstructor` | `<Program>` | yes — correct |
| `DataStructSynthesizer.cs:231/876` | inline/data-struct | yes — correct |

I did not accept this by reading. **Site 1587 was my prime suspect** — a user ctor with `: base(...)` looked like it would route to an unfixed site. Probed: `public open class Base` + `class Derived : Base { internal init() : base() { } }` → `Derived..ctor()` = **`Assembly`**. Hypothesis dead. Every other unfixed site emits a ctor with **no user-declarable accessibility** (verified: `class Prim(Name string)` → `Public`, no syntax to say otherwise).

### The #3049 trap does not recur here
A class whose *only* ctor is `private` emits **`Private` alone** — nothing synthesizes a public default that would silently undo the fix. That was the specific failure mode you asked me to check, and this emitter does not have it.

### Shape battery (metadata via `Assembly.Load` + `GetTypes()` + `ConstructorInfo`)
| shape | emitted |
|---|---|
| `class OnlyPrivate { private init() }` | `Private` (only) |
| `class NoCtorAtAll {}` | `Public` |
| `struct SInternal { internal init(x) }` | `Assembly` |
| `Outer+NestedPrivateCtor { private init() }` | `Private` |
| `Gen[T] { private init() }` | `Private` |
| `private convenience init(x)` → `public init()` | `Private`, delegation target still `Public` and reachable |
| `internal init() : base()` | `Assembly` |

**`protected` on a `struct` is rejected by the binder** — `GS0380: 'protected' is only allowed on members of an 'open class'`. So `Family` cannot arise on a struct; the question is closed at the binder, not the emitter.

### The binder was already the gate — this reframes the source-break risk
`private init` called from outside the class → **`GS0472: 'P.init' is inaccessible due to its protection level`**, and **`gsc` and `gsi` agree**. The binder always enforced accessibility; only the emitter recorded the wrong bits. **Therefore no single-compilation program can newly break** — anything that would now fail was already rejected.

That leaves exactly one real behavior change, which is **cross-assembly**, and I measured it end to end:

| | `main` `33bea4af` | head `6dbac07e4` |
|---|---|---|
| `MyLib.IntCtor..ctor()` metadata | **`Public`** | **`Assembly`** |
| cross-assembly `IntCtor()` compile | **rc=0** | **rc=1** |
| cross-assembly run | prints **`22`** | n/a |

This is the intended narrowing — the author wrote `internal init`, and `main` was leaking it. Critically it is a **clean compile-time error, not a runtime `MethodAccessException`**, so the inverse risk you named does not materialise.

**And the legitimate cross-assembly path still works end to end on head: `PubCtor()` → `11`, `Factory.Make()` → `22`** — the second one matters, because it proves the `internal` ctor is still callable from *inside* its own assembly and the narrowing did not over-apply.

### Corpus
**122 samples compiled on both trees: 0 rc differences, 0 diagnostic-set differences.**

### Mutation — 2 PRODUCT mutants, both killed
`AccessibilityMap.ToMethodVisibility`:
- **M1** `Internal → Public`: build 0/0 (live, not a no-result), Core md5 `efb23c9c` → `9a90bd36`. Killed `NonPublicExplicitConstructors_EmitDeclaredAccessibility` — `Expected: Assembly, Actual: Public`.
- **M2** `Protected → Public`: build 0/0, md5 → `7d00333d`. Killed the same test.

I ran **two** deliberately: xUnit aborts at the first failed assertion, so one mutant only proves the *first* asserted row is checked. Two mutants pin two rows. Restored to the exact clean md5 `efb23c9c`.

---

### SHOULD-FIX 1 — the newly-reachable failure reports a misleading diagnostic
Cross-assembly construction of an `internal init` fails with:

```
error GS0130: Function 'IntCtor' doesn't exist.
```

The type exists and the constructor exists; it is **inaccessible**. Same-assembly gets this right (`GS0472 … inaccessible due to its protection level`). **This is pre-existing** — on `main`, a cross-assembly call to an `internal func` gives the equally-misleading `GS0159: Cannot find function`, so it is the standing behaviour for inaccessible imported members, not something this PR introduced. Flagging it because this PR makes constructors reach it, and "doesn't exist" on a type the user can plainly see will generate confused reports against #2893.

### SHOULD-FIX 2 — the gates are structurally blind to this change, and there is no `gsi` column
**0 of 148 `samples/` + `tools/cs2gs/` files declare a non-public explicit `init`** (positive control: **151** files contain `init` at all). So the green corpus, `cs2gs-corpus` and `cs2gs-oahu` are not evidence about this PR — the same blind spot I measured on #3070.

The author's 9-row matrix is **entirely single-assembly**, and single-assembly is precisely the configuration where *nothing observable changed* (the binder already enforced it). **The one axis where behaviour actually changed — cross-assembly — has no test.** Add one: emit a library with `internal init`, compile a consumer with `/r:`, assert the diagnostic. Also no interpreter coverage, though I confirmed `gsi` does enforce (`GS0472`).

### NIT — matrix vs. tests
The body advertises 9 rows; the file has **2 `[Fact]`s**. They actually cover *more* shapes than the matrix (`data class`, `SynthesizedDefault`, `PrimaryCtor`, `StaticInitializer`, plus a `types.Length == 16` guard), so this understates the tests — but the two numbers should agree.

---

**Hygiene:** every mutation restored with an exact md5 round-trip (`AccessibilityMap.cs` `ba9fa508…` → `ba9fa508…`; rebuilt Core back to `efb23c9c`), `git status --porcelain` empty in my scratch clone. All work in a private clone; no worktree touched, no e2e, `~/.nuget/packages/gsharp.net.sdk/` never written. Every probe compiled into its own empty directory.

DavidObando:
### Addendum — #2893's reported shape reproduced verbatim (Bulletin #4 §1)

My verdict above rested on shapes **I** wrote, not on the repro **the issue** wrote. Bulletin #4 §1 says that is not sufficient for a fix claim, so I closed the gap. I ran the issue body's program byte-for-byte, unmodified, each into its own empty directory:

| tree | `Main()` exit code |
|---|---|
| `main` `33bea4af` | **1** — issue's stated "Actual", bug reproduced |
| merged head `6dbac07e4` | **0** — issue's stated "Expected" |

Drivers verified fresh before each run on `Core/`, `Compiler/` **and `Repl/`** (main `62bcf8ec`, merged `efb23c9c`). Worth recording that when I started this check the source tree was at `merged3074` while the **binaries were still main-built** — the stale-driver trap, caught only by checking mtime+md5 rather than assuming the checkout implied the build.

The reported shape is reproduced and fixed. Verdict unchanged: **MERGEABLE**, with the two SHOULD-FIX items above.

PR #3077, opened 2026-08-02
Add interpreter golden sample conformance coverage

### LINKED ISSUE #2939: Add interpreter golden-sample conformance coverage

DavidObando:
## Summary

- gate all 122 single-file golden samples across the real bare `gsc` executable, emitted `gsc /out:` + runtime, and the real `gsi` executable
- assert exact exit code, diagnostic-ID set, stdout, and stderr; undeclared drift fails
- report `gsc` and `gsi` divergences together so one failing column cannot mask the other
- keep the existing emit checks (`IlVerifier`, `Assembly.Load(File.ReadAllBytes(...)).GetTypes()`, child execution)
- pin exact known differences and fail when an entry changes or disappears
- keep multi-file samples on the emit path because evaluator front ends accept one script

Both evaluator columns launch their real drivers through the existing bounded child-process helper. The Repl project reference is build-order-only:

```xml
<ProjectReference Include="@(GsharpRepl)" ReferenceOutputAssembly="false" />
```

That keeps `out/bin/Release/Repl/gsi` fresh for project-scoped test runs without adding Repl output or its transitive analyzers to the `Compiler.Tests` reference graph.

Fixes #2939

## `func Main()` census and non-vacuity

Exactly **2/122** single-file golden samples declare explicit `func Main()` and have non-empty goldens:

| Sample | emit | bare `gsc` | `gsi` |
|---|---|---|---|
| `RefStructGenericField.gs` | rc0, `3` | rc1, `GS0511`, empty | rc1, `GS0511`, empty |
| `SpanComprehensive.gs` | rc0, `60\n10\n2` | rc1, `GS0511`, empty | rc1, `GS0511`, empty |

#2984 made both evaluator front ends invoke explicit `Main`. The gate still pins the exact two-file census and non-empty goldens; neither sample can pass vacuously. Their current declared difference is the documented GS0511 ByRefLike boundary (`docs/diagnostics.md:1168-1172`, #3114).

## Exact declared differences (14)

| Samples | Evaluator expectation | Reason |
|---|---|---|
| `Deinit.gs` | rc0, `GS0510`, allocation + `done` only | documented GS0510 boundary; #2988 remains open |
| six P/Invoke samples | rc1, `GS0514`, empty program output | ADR-0152 / #2986 native-call boundary |
| `RefStructGenericField.gs`, `RefStructSpan.gs`, `SpanComprehensive.gs`, `SpanIndexing.gs` | rc1, `GS0511`, empty program output | documented GS0511 ByRefLike reflection boundary / #3114 |
| `GsharpExtensionsMixed.gs` | rc1, `GS0157` + `GS0159`, empty output | #3130: evaluator drivers cannot load the separately built `Gsharp.Extensions` assembly |
| `GsharpExtensionsOptional.gs` | rc1, `GS0159`, empty output | #3130 |
| `GsharpExtensionsSequences.gs` | rc1, `GS0157` + `GS0158` + `GS0159`, empty output | #3130 |

The Extensions rows now describe both real evaluator drivers; emitted output still matches the goldens. `PInvokeStructMarshalling.gs` is not ledgered: #3100 fixed its `@FieldOffset` overlap behavior, and the self-liquidating ledger removed the entry.

## Freshness proof

Product mutant, asserted as exactly one insertion/deletion:

```diff
- return 0;
+ return 3;
```

at `src/Repl/Program.cs:79`.

A plain project-scoped test command, with no manual solution build:

```sh
dotnet test test/Compiler.Tests/Compiler.Tests.csproj \
  --configuration Release --no-restore --nologo --verbosity minimal \
  --filter 'FullyQualifiedName~SampleConformanceTests'
```

rebuilt `gsi` (`mtime 1785731171 → 1785731907`) and went red:

- **109 failed / 16 passed**
- **109/109** failures reported `diverged under gsi`
- representative actual result: rc3 with otherwise matching output

After restoring the product source, the project build refreshed `gsi` again (`1785731907 → 1785732244`) and the clean gate passed **125/125**.

## Cost and validation

Numbers true of `main` **`f7d9a5d3`** and head **`179ad6eb`**:

- three-driver gate, Release: **125/125**, xUnit **2m26s**
- unfiltered `Compiler.Tests`: **4,159 passed / 0 failed / 0 skipped**, **1h20m**
- Release solution build: **`BUILD_RC=0`**, 0 warnings, 0 errors
- `Compiler.Tests` restore graph: **0** `Microsoft.VisualStudio.Threading` / `VSTHRD` assets
- merge base is `f7d9a5d3`; merged tree **`2a08e42d`** exactly equals the branch tree
- exact-head CI: **29 successful**, 2 publish jobs skipped
- no production files changed

The gate triples driver coverage; it adds no source shapes. It would not have caught #3070 (the corpus still has one top-level out-parameter function), #2963 (125/125), or #2903 (zero `/target:library` sample builds).

### DIFF
--- a/test/Compiler.Tests/Compiler.Tests.csproj
+++ b/test/Compiler.Tests/Compiler.Tests.csproj
@@ -8,6 +8,8 @@
 
   <ItemGroup>
     <ProjectReference Include="@(GsharpCompiler)" />
+    <ProjectReference Include="@(GsharpRepl)" ReferenceOutputAssembly="false" />
+    <ProjectReference Include="..\..\src\Sdk\Gsharp.Extensions\Gsharp.Extensions.csproj" />
     <Compile Include="..\Shared\Issue3081CompositeLiteralCases.cs" Link="Issue3081CompositeLiteralCases.cs" />
     <!-- Issue #2215: only this test project needs a real gsgen.dll build output
          (to resolve via /gsgentool:) plus Roslyn (to compile a fixture
--- a/test/Compiler.Tests/LanguageConformance/SampleConformanceData.cs
+++ b/test/Compiler.Tests/LanguageConformance/SampleConformanceData.cs
@@ -1,0 +1,58 @@
+// <copyright file="SampleConformanceData.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+
+namespace GSharp.Tests.LanguageConformance;
+
+internal static class SampleConformanceData
+{
+    public static IReadOnlyList<string> GetSingleFileSamples(string samplesDirectory)
+        => Directory.EnumerateFiles(samplesDirectory, "*.gs", SearchOption.TopDirectoryOnly)
+            .Where(path => File.Exists(Path.ChangeExtension(path, ".golden")))
+            .Select(Path.GetFileName)
+            .OrderBy(name => name, StringComparer.Ordinal)
+            .ToArray();
+
+    public static IReadOnlyList<string> GetMultiFileSamples(string samplesDirectory)
+        => Directory.EnumerateDirectories(samplesDirectory)
+            .Where(path => !string.Equals(
+                Path.GetFileName(path),
+                "aspirational",
+                StringComparison.Ordinal))
+            .Where(path =>
+                File.Exists(Path.Combine(path, Path.GetFileName(path) + ".golden"))
+                && Directory.EnumerateFiles(path, "*.gs").Any())
+            .Select(path => Path.GetFileName(path) + "/")
+            .OrderBy(name => name, StringComparer.Ordinal)
+            .ToArray();
+
+    public static string LocateSamplesDirectory(Assembly testAssembly)
+    {
+        var directory = new DirectoryInfo(Path.GetDirectoryName(testAssembly.Location)!);
+        while (directory != null)
+        {
+            var candidate = Path.Combine(directory.FullName, "samples");
+            if (Directory.Exists(candidate)
+                && File.Exists(Path.Combine(directory.FullName, "GSharp.sln")))
+            {
+                return candidate;
+            }
+
+            directory = directory.Parent;
+        }
+
+        return null;
+    }
+
+    public static string ReadNormalizedFile(string path)
+        => NormalizeLineEndings(File.ReadAllText(path));
+
+    public static string NormalizeLineEndings(string text)
+        => text.Replace("\r\n", "\n", StringComparison.Ordinal);
+}
--- a/test/Compiler.Tests/LanguageConformance/SampleConformanceTests.cs
+++ b/test/Compiler.Tests/LanguageConformance/SampleConformanceTests.cs
@@ -7,26 +7,28 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Reflection;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using GSharp.Tests.LanguageConformance;
 using Xunit;
 
 namespace GSharp.Compiler.Tests.LanguageConformance;
 
 /// <summary>
-/// End-to-end conformance harness: every `.gs` file under `samples/` that has a
-/// sibling `.golden` file is compiled through `gsc.dll`, executed under `dotnet`,
-/// and its stdout is compared bit-for-bit against the golden.
+/// Runs every golden sample through the emitted runtime and every single-file
+/// sample through both evaluator front ends.
 /// </summary>
-/// <remarks>
-/// This is the executable spec for the parseable subset of GSharp. Per
-/// `docs/adr/0010-aspirational-samples.md`, every sample under `samples/` MUST
-/// build, run, and produce its golden output on every PR. Aspirational samples
-/// belong under `samples/aspirational/` (excluded from discovery) and are not
-/// covered by this harness.
-/// </remarks>
 public class SampleConformanceTests
 {
-    // Samples that DllImport Unix-style `libc` and therefore cannot run on
-    // Windows (issue #1010). They remain covered on Linux/macOS.
+    private const int MinimumSingleFileSampleCount = 122;
+
+    private static readonly string[] MainOnlySamples =
+    {
+        "RefStructGenericField.gs",
+        "SpanComprehensive.gs",
+    };
+
     private static readonly HashSet<string> WindowsSkippedSamples = new(StringComparer.Ordinal)
     {
         "PInvoke.gs",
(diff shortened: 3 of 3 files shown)


DavidObando:
## NOT MERGEABLE

Numbers true of `main` **`4f7559dd`**, head **`b3cd4698`**, computed merge base **`33bea4af`**, merged tree **`4887f13a`** (`git merge-tree --write-tree` identical both directions, rc 0; reproduced exactly by a real `git merge`, `MERGE_RC=0`).

### Lead answer: the harness is good, and it does not build.

The brief's primary worry — `func Main()`-shaped samples passing vacuously under the two evaluator columns — **does not hold. That is handled properly** (evidence below). The blocker is somewhere else entirely: **this PR does not compile**, on its own or merged, and CI is red on every build and test job.

---

## BLOCKER 1 — the tree does not build; `Compiler.Tests.csproj:11` propagates threading analyzers

`test/Compiler.Tests/Compiler.Tests.csproj:11` adds `<ProjectReference Include="@(GsharpRepl)" />`. `Repl` transitively carries `Microsoft.VisualStudio.Threading.Analyzers`; under this repo's warnings-as-errors those analyzers promote **pre-existing** `.Result` / `.Wait()` / `Task.Factory.StartNew` usages in `Compiler.Tests` into build errors.

| tree | `BUILD_RC` | `error VSTHRD*` |
|---|---|---|
| `main` `4f7559dd` alone (control) | **0** | **0** |
| branch `b3cd4698` alone | **1** | **65** |
| merged `4887f13a` | **1** | **70** |

Same environment, same command (`dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true`). **This is not an interaction with `main`** — the branch fails standalone.

Attribution measured in the restore graph, not inferred:

| project | `VisualStudio.Threading.Analyzers` hits in `project.assets.json` |
|---|---|
| `Repl` | **37** |
| `Gsharp.Extensions` | 0 |
| `Compiler` | 0 |
| `Core` | 0 |
| `Compiler.Tests` on **`main`** | **0** |
| `Compiler.Tests` on **branch** | **37** |

Chain: `StreamJsonRpc/2.21.10` → `Microsoft.VisualStudio.Threading/17.10.48` → `…Threading.Analyzers`. The second new reference (`Gsharp.Extensions`) is innocent.

**CI agrees exactly.** `gh pr checks 3077`: `build` **fail**, and **all 20 test shards fail**. Job `91494935885` reports `122 error VSTHRD002`, `4 error VSTHRD200`, `4 error VSTHRD105` — identical counts and identical first locations to my local run (`IlVerifier.cs(250,28)`, `(253,28)`, `(257,52)`).

Measured note for whoever fixes this: **`ExcludeAssets="analyzers"` on the `ProjectReference` did NOT suppress it** (still `BUILD_RC=1`, 70 errors). A `NoWarn` of `VSTHRD002;VSTHRD105;VSTHRD200` does (`BUILD_RC=0`).

### Smaller fix worth considering — delete the reference rather than suppress it

Only the **gsi** column needs `Repl`. `RunDriverProcessAsync` **already exists** at `SampleConformanceTests.cs:322-346` and is already used for `EventSubscription.gs`. Routing all samples through it instead of in-process `SessionEngine` would:

- remove the `GsharpRepl` `ProjectReference` and the whole analyzer problem at its root, with no suppression;
- upgrade the gsi column from *engine* to *actual `gsi` driver*, covering `src/Repl/Program.cs`, which **121 of 122 samples currently never touch**.

Cost is 122 process spawns; the suite is 2m17s today, so expect roughly double. That is deletion plus reuse of existing code, not new code.

---

## SHOULD-FIX 2 — every validation claim in the PR body is false or stale

- *"fresh full solution build: 0 errors"* — measured `BUILD_RC=1`, 65 errors branch-alone.
- *"all required PR checks pass, including cs2gs corpus, pinned Oahu migration, and e2e"* — `build` and **20/20** test shards fail.
- *"complete CI test matrix: 14,380 passed, 0 failed"* — not true of this head.
- *"merged tree equals the branch tree"* — branch tree is `6483db8b`, merged tree is `4887f13a`. True at `33bea4af`, false since `main` moved to `4f7559dd`.
- *"0 failed / 123 passed"* — that is commit `0d2306cc` (interpreter-only), before `b3cd4698` added the other columns. Actual on merged: **125**.

The body also describes only the interpreter theory while `b3cd4698` gates all three drivers.

---

## SHOULD-FIX 3 — a live silent-wrong-answer bug is ledgered as an intentional ADR-0152 boundary

`SampleConformanceTests.cs:60-63`:

```csharp
["PInvokeStructMarshalling.gs"] = SameDifference(
    Expected(0, "3\n4\n0\n"),
    "#2986",
    "the evaluator cannot execute the native struct-marshalling call"),
```

**The sample contains no native call.** The only `native|DllImport|LibraryImport` match in `samples/PInvokeStructMarshalling.gs` is a comment at line 6.

Measured, fresh empty directory:

| driver | rc | stdout |
|---|---|---|
| emitted (golden) | 0 | `3` / `4` / `6153737367135073092` |
| bare `gsc` evaluate | 0 | `3` / `4` / **`0`** |
| `gsi` | 0 | `3` / `4` / **`0`** |

Real defect: `samples/PInvokeStructMarshalling.gs:33-37` declares `@StructLayout(LayoutKind.Explicit, Size: 8)` with `@FieldOffset(0) LowPart uint32`, `@FieldOffset(4) HighPart int32`, `@FieldOffset(0) QuadPart int64`. Compiled honours the overlap; both evaluators ignore `@FieldOffset` and read `QuadPart` as `0`. **rc0, no diagnostic — a silent wrong answer.**

The six genuine `NativeCallDifference()` rows are `Expected(1, string.Empty, "GS0514")`. This row is `Expected(0, …)` with no diagnostic — structurally not a native-call difference. The PR body gets this right (*"contains no native call … a separate newly measured bug"*); the code contradicts the body and files it beside rows the body calls *"intentional ADR-0152 native-call boundaries, not open evaluator bugs"*. As written, a reader concludes there is nothing to file.

Fix: give it its own reference and an accurate reason. @DavidObando — this one needs an issue; you are sole filer.

---

## SHOULD-FIX 4 — the gsi assertion is masked whenever gsc also diverges

`SampleConformanceTests.cs:146-147`:

```csharp
AssertResult(sampleName, "gsc", emitted, evaluatingCompiler);
AssertResult(sampleName, "gsi", emitted, interpreter);
```

Sequential asserts; the first throws. Proven by mutation M1 below: **61 failures, 61 labelled `diverged under gsc`, 0 labelled `gsi`** — the interpreter column was never reached on any failing sample.

Not a coverage hole in a green tree (all 122 pass gsc, so gsi is checked). It is a diagnosis cost: a change breaking both columns must be fixed once and re-run to discover the second failure. Collect both results and assert once.

---

## Mutation — both new columns are independently load-bearing

Per-arm product mutants, each asserted to have applied (exact occurrence count), each built to `BUILD_RC=0` with `Core/`, `Compiler/`, `Repl/` mtimes refreshed and equal.

| arm | mutant | result | labels |
|---|---|---|---|
| clean | — | **125 passed / 0 failed** | — |
| **M1** shared evaluator | `Evaluator.Operators.cs:189` `b.Type == TypeSymbol.String` → `!=` | **61 failed / 64 passed** | 61 × `gsc`, 0 × `gsi` |
| **M2** interpreter front end | `SessionEngine.cs:194` `if (CaptureConsole)` → `if (!CaptureConsole)` | **112 failed / 13 passed** | 112 × `gsi`, 0 × `gsc` |
| restored | — | **125 passed / 0 failed** | — |

`MultiFileSample_EmittedRuntimeMatchesGolden` (the emit-only theory) recorded **0** failures under M1 — positive-controlled, the sibling name matched 183 times in the same log. So the 61 kills come from the newly added columns, not from emit. M2 confirms the gsi column pins independently of the shared evaluator.

Both mutants were run with the local `NoWarn` unblock applied, since the tree does not otherwise build. Restore verified by `git status` (only my deliberate csproj unblock differed) and by the behavioural round-trip above.

---

## Verified correct — credit where it is due

- **Vacuity is genuinely handled.** `ExplicitMainSamples_AreDeclaredAndCannotPassVacuously` (`:179-201`) matches `(?m)^\s*func\s+Main\s*\(`, asserts the discovered set **equals** `MainOnlySamples`, requires a non-empty golden, and requires an explicit ledger entry. Verified against the corpus: anchored and unanchored regexes both yield exactly `RefStructGenericField.gs` and `SpanComprehensive.gs`, and both are ledgered `Expected(0, string.Empty)` citing #2984. **The rc0-plus-empty-output vacuous pass is declared, not silent.**
- **Output is asserted, not just rc.** `Equivalent` (`:426-431`) compares ExitCode **and** Output **and** DiagnosticIds **and** StandardError.
- **Discovery floor is exact and non-vacuous**: `MinimumSingleFileSampleCount = 122`; `samples/*.gs` = 122; samples lacking a `.golden` = **0**. Observed total 125 = 122 + 1 multi-file + 2 facts, matching prediction.
- **Real closure of the emit-only blindness.** Base `SampleConformanceTests.cs` at `33bea4af` is 244 lines with a single theory `Sample_BuildsAndMatchesGolden` — emit only. This PR adds the bare-`gsc` evaluate and interpreter columns over all 122 samples. That is the structural gap this project has been paying for repeatedly.
- Emit column does `Assembly.Load(File.ReadAllBytes(outputPath)).GetTypes()` (`:250`) beyond `IlVerifier.Verify`, and each emit probe gets its **own fresh GUID directory**, deleted afterwards.
- Ledger is self-liquidating (`:135-139`): a newly matching sample fails until its entry is removed.
- `IlVerifier` sample ignore list is 2 tightly method-scoped entries, pre-existing and untouched.
- Hypothesis I killed rather than reported: `CaptureConsole` mutates process-global `Console.Out`, but `test/Compiler.Tests/AssemblyInfo.cs:9` sets `[assembly: CollectionBehavior(DisableTestParallelization = true)]`, and base already captured Console. Not live.

---

## NIT 5 — asymmetric output truncation

`CleanCliStream` (`:365-377`) truncates the evaluator/interpreter streams at the first line matching `GS\d{4}|GSI\d{3}`; the emitted column keeps raw stdout. A sample legitimately printing such a token would be compared truncated on one side only. **Currently latent** — 0 of 122 goldens contain a matching token (positive-controlled: the regex does match when one is present).

---

## Scope note — this triples driver coverage, it adds no new shapes

Asked whether this harness would have caught recent escapes. Measured on the merged tree:

- **#3070** (out-param definite assignment across 12 declaration contexts) — **no**. `samples/` still declares exactly **one** out-parameter function, `samples/TryParseOutVar.gs:11`, and it is top-level. This PR adds zero samples. *(My first count here returned 0 because I used a `&` pattern; G# spells it `out result int32`. Zero was scope, not absence — corrected and re-run.)*
- **#2963** — **no**. The merged run at `4f7559dd`, which includes #2963, is 125/125, so no sample exercises the broken shape.
- **#2903** (`GS0159`, explicit generic lambda across a project reference) — **no**. **0** samples compile with `/target:library`; there is no cross-assembly shape in the corpus.

So *"samples now covered"* is true of **drivers**, not of **shapes**. That is still a large and overdue win — it should just be stated as scoped.

---

## Summary

| # | severity | item |
|---|---|---|
| 1 | **BLOCKER** | `Compiler.Tests.csproj:11` propagates threading analyzers; branch and merged both `BUILD_RC=1`; CI red on build + 20/20 shards |
| 2 | SHOULD-FIX | every validation claim in the PR body is false or stale |
| 3 | SHOULD-FIX | `SampleConformanceTests.cs:60-63` ledgers a live silent-wrong-answer explicit-layout bug as an intentional ADR-0152 native boundary |
| 4 | SHOULD-FIX | `:146-147` gsi assertion masked by the gsc assertion |
| 5 | NIT | `:365-377` asymmetric diagnostic truncation (latent) |

The design is sound and the mutation evidence says both new columns pin real behaviour. Fix the build and this is worth landing.

*Disclosures: `Compiler.Tests` run under the `SampleConformance` filter only, not the full ~1h20m suite — a filtered green is not a green. All mutation runs required a local `NoWarn` unblock because the tree does not build; that unblock is a measurement aid, not a recommended fix. One `uptime` (34.54) was returned concurrently with a build rather than before it; all other builds were gated below 30, and I waited out load repeatedly between 32 and 45.*

DavidObando:
## Re-review — NOT MERGEABLE (1 blocker)

Re-anchored: `main` **`4469e323`**, head **`9e7f8d7c`**, merge base `4469e323` (`main` is an ancestor, so merged tree == branch tree). **Note: the head moved from `726faaa9` to `9e7f8d7c` mid-review and the design changed materially** (child-process `gsi`, ProjectReference deleted, 11→14 ledger rows). Everything below is measured on `9e7f8d7c`.

### How the `build` blocker was resolved — and is it acceptable?

**Resolution: the `@(GsharpRepl)` ProjectReference was deleted outright** from `test/Compiler.Tests/Compiler.Tests.csproj`, and the `gsi` column re-implemented as a child process (`SampleConformanceTests.cs:316-340`) resolving `<testdir>/../Repl/gsi`.

- **No suppression of any kind.** `git diff 4469e323 9e7f8d7c --full-index` grepped for `NoWarn|TreatWarningsAsErrors|VSTHRD|editorconfig|severity|WarningLevel|EnableNETAnalyzers|AnalysisLevel|ExcludeAssets` → **rc=1, zero matches**. Only 3 files touched, all under `test/`. **Zero production changes.**
- Clean-slate `rm -rf out && dotnet build GSharp.sln -c Release` → **`BUILD_RC=0`, 0 VSTHRD**.
- **Sample set not narrowed**: 122 samples, `MinimumSingleFileSampleCount = 122` (`:24`), explicit-`Main` census pinned (`:187-199`).

So it is not resolution #1 (suppression) — good — it is resolution #2, the one I recommended. **But it shipped without the freshness guarantee**, which is exactly the risk flagged.

---

### BLOCKER 1 — the gate silently tests a **stale `gsi`**

`Compiler.Tests.csproj` no longer declares *any* dependency on `src/Repl`, so nothing orders the build. `RunDriverProcessAsync:329` asserts only `File.Exists(executable)` — **absence fails loudly, staleness is invisible.**

Product mutant `src/Repl/Program.cs:79` `return 0;` → `return 3;` (asserted: exactly **1** occurrence, `git diff --stat` = 1 file / 1 insertion / 1 deletion). Identical mutant, identical filter, same worktree — only the build differs:

| run | `gsi.dll` mtime | result |
|---|---|---|
| `dotnet test test/Compiler.Tests/Compiler.Tests.csproj --filter ~SampleConformance` | 20:14:19 (**stale**) | **125 passed / 0 failed — live mutant reads as SURVIVED** |
| then `dotnet build GSharp.sln` and re-run identical filter | 20:27:12 (fresh) | **109 failed**, `109/109 "diverged under gsi"` |

It cuts both ways — clean source + stale mutant binary → **109 false FAILURES** (`gsi.dll` still 20:27:12). **The gate's verdict is a function of out-of-band build state, not of the source tree.**

**Fix — one line, already measured, and it is this PR's own previous head `726faaa9`:**

```xml
<ProjectReference Include="@(GsharpRepl)" ReferenceOutputAssembly="false" />
```

Measured with the mutant still in place: project-scoped `dotnet test` **rebuilt Repl** (`gsi.dll` 20:27:12 → **20:37:19**), gate returned to **125/125**, and **0 VSTHRD**. `ReferenceOutputAssembly="false"` keeps the build-ordering edge while stopping asset/analyzer flow — it fixes the analyzer blocker *and* freshness. The correct fix was already in hand and was replaced with a strictly weaker one.

### SHOULD-FIX 2 — the bare-`gsc` column is not the bare-`gsc` driver

`RunCompilerEvaluator` invokes `GSharp.Compiler.Program.Main` **in-process** with only `/nowarn:GS9100 <source>`, after the theory executes `Assembly.Load("Gsharp.Extensions")` at `:128`. Running the *identical arguments* as a real process:

```
./out/bin/Release/Compiler/gsc /nowarn:GS9100 samples/GsharpExtensionsOptional.gs   -> rc=1  GS0159
./out/bin/Release/Compiler/gsc /nowarn:GS9100 samples/GsharpExtensionsMixed.gs      -> rc=1  GS0157
./out/bin/Release/Compiler/gsc /nowarn:GS9100 samples/GsharpExtensionsSequences.gs  -> rc=1  GS0157
```

`Gsharp.Extensions.dll` is adjacent to **neither** `gsc` nor `gsi` in the build output; the emit column only works because `:227-245` copies it and passes `/r:`. So these three are classified `InterpreterDifference` — which via `:138` (`difference.EvaluatingCompiler ?? emitted`) asserts **bare `gsc` succeeds** — when in reality **both evaluating drivers fail identically**. The reason *text* is true; the **classification** is not, and it hides that bare `gsc` has the same gap.

Fix: pass `/r:` to the evaluator column as emit does, or route bare `gsc` through the same `RunDriverProcessAsync` already used for `ProcessExitSample`, and re-ledger as `SameDifference`.

### SHOULD-FIX 3 — ledger references

- The 4 ByRefLike rows (`:56-73`) cite **`#2987`, which is CLOSED**. The 6 native rows set the better precedent: `"ADR-0152 / #2986"` (`:414`) — governing doc *plus* issue. Cite `docs/diagnostics.md:1168-1172` (GS0511) and/or open **#3114**.
- The 3 Extensions rows cite **`"unfiled"`**. Verified genuine, not a harness artifact: `gsi --help` shows `gsi [file.gs] [--help] [--version]` — **no reference channel exists**, so these can never work under `gsi`. Worth an issue.

### Verified good

- **Ledger hand-checked against source, all 14 rows.** GS0511's sole raise site is `Evaluator.Reflection.cs:54-64` (a reflection guard) — both distinct ByRefLike reasons (top-level call vs `func Main` invocation) are accurate; the 2 `func Main` samples are exactly the 2 with `func Main(`.
- **Masking resolved**: `AssertResults` (`:436-450`) now reports both columns together — my earlier SHOULD-FIX is fixed.
- `PInvokeStructMarshalling` row correctly removed (bug fixed by #3105).
- **Column execution counted, not inferred**: the `gsi` column executes **122/122** samples — 109 lethal to the mutant, 10 returning rc1 at `Program.cs:66` before it, 3 Extensions rows rc1. Survivor arithmetic exact: 16 = 10 + 3 + 1 multi-file + 2 facts.
- **#3114 criterion 2 (gate fails on unledgered interpreting-column divergence) is MET** — 109/109 `"diverged under gsi"` — **but only when `Repl/` is rebuilt.** See BLOCKER 1.

### Correction to the review brief

The premise that `SpanComprehensive.gs` / `RefStructGenericField.gs` fail with **`GS9999`** is **stale**. Measured here: all four ByRefLike samples are rc=1 **`GS0511`** under *both* bare `gsc` and `gsi` — a designed, documented boundary (`docs/diagnostics.md:1168-1172`), delivered by #3055. **This PR's ledger is correct as written.** (#3114's title still says GS9999.)

### Disclosures

Filtered green (`~SampleConformance` only, not the full suite). One `uptime` read 38.40 and I ran `dotnet test` in the same invocation anyway — above threshold, flagging it. One run segfaulted at restore (rc=139, 1-line log, nothing executed) — discarded as a flake and re-run, not counted as a result. I reviewed `726faaa9` first and re-measured everything after the head moved.

DavidObando:
## Round 2 — **MERGEABLE**

Anchors: head **`179ad6eb`** (unchanged, no force-push during review), `main` **`d240eb24`** (moved from `27a05798` mid-review; re-anchored, see below), merged tree **`9c70a15a`** (reproduced `git merge-tree`'s prediction exactly).

### The blocker is fixed — proven by mutation, not by reading the csproj

Round 1's blocker was that `test/Compiler.Tests/Compiler.Tests.csproj` did not force `gsi` to build, so the gate could pass against a **stale** `gsi`. The fix is `<ProjectReference Include="@(GsharpRepl)" ReferenceOutputAssembly="false" />` at `:11`.

Product mutant: `src/Repl/Program.cs:79` `return 0;` → `return 3;` (asserted applied: `git diff --stat` = 1 insertion / 1 deletion, exactly 1 occurrence of `^        return 0;$`).

Command in both directions — **plain, project-scoped, UNFILTERED, no manual pre-build, no `--filter`** (`dotnet test test/Compiler.Tests/Compiler.Tests.csproj -c Release`):

| source | result | `Repl/gsi.dll` mtime |
|---|---|---|
| **mutant** | `TEST_RC=1` — **Failed: 109**, Passed 4066, Total 4175 | 00:05:03 → **00:10:41** (rebuilt by the test invocation itself) |
| **restored** | `TEST_RC=0` — Failed 0, **Passed 4175**, Total 4175 | → 01:35:17 |

Both halves of the fix are demonstrated:

- **Ordering** — `gsi.dll` mtime advanced *during* the plain `dotnet test`, with no pre-build. The reference does force Repl to rebuild.
- **Detection** — all **109/109** failures are `diverged under gsi`, **0** under `gsc`. Representative: `Sealed.gs diverged under gsi. Expected: rc=0 ... Actual: rc=3` — my mutant's exact value, with **stdout still matching**, so the harness catches a pure-rc regression, not just an output one.

Contrast with round 1, same command, same worktree: **125/125 PASS with a live mutant** and `gsi.dll` mtime unchanged. The verdict is now a function of source rather than of build state.

Arithmetic corroborates: 122 samples − 13 ledger rows that expect `rc=1` = **109**. Those 13 survive because `gsi` returns early at `Program.cs:66` and never reaches `:79`. Positive control: 327 `SingleFileSample_ConformsAcrossAllDrivers` lines in the mutant log — the theory executed.

### Re-anchored at current `main`

`main` moved 27a05798 → `d240eb24` during my runs. `samples/` unchanged (0 files), but **22 `src/` files moved including `Evaluator.Expressions.cs` and `Evaluator.Concurrency.cs`** — precisely the code the two new columns exercise, so the earlier green needed re-anchoring.

Merged at `d240eb24`: `BUILD_RC=0`, **0 errors, 0 VSTHRD** (the round-1 analyzer blocker has not returned; no suppression appears anywhere in the PR's own diff — grep for `NoWarn|TreatWarningsAsErrors|VSTHRD|editorconfig|severity|ExcludeAssets` over `f7d9a5d3..179ad6eb` returns nothing). Conformance class re-run: **125/125 pass, 3 m 38 s**. *Scoped probe — conformance class only; the 4175/4175 figure above is the unfiltered one, at `27a05798`.*

### Both round-1 SHOULD-FIXes adopted, and I verified them against live drivers

`RunCompilerEvaluator` and the `ProcessExitSample` special case are **deleted** — both evaluator columns now go through `RunDriverProcessAsync` (`:129-130`), real child processes. The in-process `gsc` lie is gone.

Ledger reasons hand-verified against actual driver output on this head, not read:

| sample | ledger expects | measured |
|---|---|---|
| `Deinit.gs` | rc=0, `Allocated: db/users\ndone\n`, `GS0510` | rc=0, exact stdout, **GS0510** ×2 ✓ |
| `GsharpExtensionsMixed.gs` | rc=1, `GS0157 GS0159` | gsc **and** gsi both rc=1, `GS0157 GS0159` ✓ |
| `GsharpExtensionsOptional.gs` | rc=1, `GS0159` | both rc=1, `GS0159` ✓ |
| `GsharpExtensionsSequences.gs` | rc=1, `GS0157 GS0158 GS0159` | both rc=1, `GS0157 GS0158 GS0159` ✓ |

The `SameDifference` reclassification is correct — both evaluating drivers fail *identically*, which is what my round-1 finding said and what the `#3130` citation now records. The ByRefLike rows correctly cite `GS0511` / `docs/diagnostics.md` / open `#3114` rather than closed `#2987`.

### Does the new coverage inherit #3130's in-process shortcut?

**No.** `Assembly.Load("Gsharp.Extensions")` survives at `:127`, but both new columns are child processes and cannot see the test host's load context. The only consumer is the pre-existing emit column, which resolves the path via its own `Assembly.Load(...).Location` at `:224`. #3130's blindness is not inherited.

---

## NIT (1)

**`SampleConformanceTests.cs:127` is now dead.** The standalone `Assembly.Load("Gsharp.Extensions");` was load-bearing when the evaluator columns ran in-process; now that both are child processes, its only remaining consumer (`:224`) performs its own `Assembly.Load("Gsharp.Extensions").Location`. The statement at `:127` is a pure duplicate. Deleting one line is a strictly smaller harness with identical behaviour. Not worth blocking on.

---

**Verdict: MERGEABLE.** The blocker is genuinely fixed, and fixed by the one-line remedy the author already had rather than by suppressing anything. This gate can now fail — which was the whole question — and it fails for the right reason, attributed to the right driver, under the command CI actually runs.

*Disclosure: round 2 initially reported an mtime anomaly (`Compiler`/`Repl` appearing older than `gsi.dll`). That was my error — I sorted `%H:%M:%S` strings lexicographically, so `19:38` sorted after `00:05` across midnight. Re-measured with sortable timestamps: every first-party artifact was fresh (`Core` 00:05:00, `gsc.dll` 00:05:01, `gsi.dll` 00:05:03); the old dates were NuGet package copies (Newtonsoft 2021, xunit 2018). No conclusion rested on the bad reading.*

PR #3078, opened 2026-08-02
Render actual aggregate kind in diagnostics

### LINKED ISSUE #3075: Diagnostic message hardcodes "Class" and mis-reports structs
`DiagnosticDescriptors.cs:105` hard-codes the word `Class` into the message template:

```
"Class '{0}' ..."
```

The diagnostic is also reported for `struct` declarations, so a struct named `Lookup` is reported as:

```
Class 'Lookup' ...
```

That is wrong and actively misleading — a user reading the message will look for a class declaration that does not exist.

### Expected

The message should name the actual kind of the symbol being reported (`class`, `struct`, `interface`, ...), driven by the symbol rather than by a literal in the template.

### Notes

- Found incidentally while reviewing the interface-verification diagnostics; it is **pre-existing** and unrelated to any PR currently open.
- Worth checking whether any *other* descriptor in `DiagnosticDescriptors.cs` hard-codes a type kind the same way. Do not grep for this alone — several of the relevant classes are `partial`, so count sites by making the compiler enumerate them rather than by text search.
- A regression test should assert the rendered message for a `struct` **and** for a `class`, so a future edit cannot re-hardcode either one.

DavidObando:
## Summary

- derive interface-implementer kind from `StructSymbol.IsClass` instead of descriptor literals
- fix GS0187, GS0188, GS0318, GS0319, GS0320, GS0331, and GS0332
- pass the implementing symbol through all reporter sites
- update both diagnostic catalogues
- add exact class/struct message coverage for every reachable changed diagnostic

## Site and reachability audit

Adding a required parameter to `ReportInterfaceMethodNotImplemented` produced 32 raw CS7036 lines, deduplicated to **16 file:line sites**. All 16 receive `StructDeclarationSyntax` plus `StructSymbol`; therefore class and struct are the only aggregate kinds GS0187 can reach. Interfaces use `InterfaceSymbol`, enums use `EnumSymbol`, and neither can call this reporter. Data/inline/ref/record declarations remain the struct branch (`IsClass == false`).

## Validation

- issue matrix: 12/12 passed
- interface regressions: 60/60 passed
- existing rendered-message regressions: 19/19 passed
- merged-tree diagnostics gate: 4/4 passed
- `git merge-tree --write-tree`: clean; merged tree content equals tested HEAD
- Release `GSharp.Core.dll` copies in Core, Compiler, and Repl had matching fresh mtimes
- `{top-level, in-function} × {gsc evaluate, gsc /out emit, gsi}`: 6/6 exact-message probes passed, with sentinels 11/22/33 and empty output directories

## Product mutation

| Hunk | Mutation | Result |
|---|---|---|
| aggregate-kind selection | invert `StructSymbol.IsClass` branches | 12 failed / 0 passed; all six exact-message theory methods died |

Fixes #3075


DavidObando:
## Rubber-duck review — PR #3078 (issue #3075)

Numbers true of `main` **`4f7559dd`**, head **`bb2b5d3f`**, merge base `33bea4af` (**4 behind / 1 ahead — so merged tree ≠ head tree, and CI's green is on head, not on what lands**).

### Merged-tree identity + build

| item | value |
|---|---|
| three-dot diff | 7 files, +244/−72 (`--full-index`) |
| `merge-tree` | rc=0 → tree `4234d8eb` |
| probe merge `55493831` | tree **reproduced `4234d8eb` exactly** |
| merged build | **`BUILD_RC=0`, 0 Warning(s), 0 Error(s)** |
| driver mtimes | `Core/`, `Compiler/`, `Repl/` all fresh, same second, `ref/` excluded |

`merge-tree` clean is a floor; I built it and read `BUILD_RC`.

### 1. Reported shape reproduced, then fixed — all three drivers

Not an adjacent shape: the struct-implements-interface case from #3075 verbatim.

| shape | `main` `4f7559dd` | merged `4234d8eb` |
|---|---|---|
| `struct Lookup : IContract` | `GS0187: **Class** 'Lookup'` | `GS0187: **Struct** 'Lookup'` |
| `class Lookup : IContract` | `Class 'Lookup'` | `Class 'Lookup'` |
| generic struct | `Class 'Lookup'` | `Struct 'Lookup'` |

Identical under **bare `gsc` (evaluate)**, **`gsc /out:` (emit)** and **`gsi`** — 3 shapes × 3 drivers, each emit into its own directory asserted empty (`dir_entries_before=0`).

### 2. The fix IS pinned — unlike #2963

Product mutant **M1**: `GetAggregateKindName` → `=> "Class";` (`DiagnosticBag.Reports.Declarations.cs:1144`), restoring the #3075 bug.

- Mutant **asserted applied** (`assert count(old)==1`) — a no-result is not a survival.
- Mutant **proven live**: all 3 driver copies re-stamped, md5 `34af035d`→`40c8aa70`, and the probe **flipped back** to `Class 'Lookup'`.
- **Kills 6 of 12** — exactly the `struct` row of all six methods:
  `MissingInterfaceMethod_UsesActualKind`, `AbstractMethodWithoutDefault_UsesActualKind`, `ConflictingDefaults_UsesActualKind`, `SealedInterfaceOutsidePackage_UsesActualKind`, `MissingStaticVirtualMethod_UsesActualKind`, `InstanceMethodForStaticVirtualSlot_UsesActualKind` — each `(declarationKind: "struct", displayKind: "Struct")`.

### 3. The #3054 "freeze a lie" pattern is absent — checked, not assumed

PR edited **zero** pre-existing test assertions (`git diff … -- test/` contains no `Class '`/`Struct '` +/− lines). Eight literal `"Class 'X' …"` assertions exist in **Compiler.Tests** (`Issue2875…:230,234,238,242,268`, `Issue2888…:123,519,520`), which the PR does not touch. I checked every asserted subject's declaration: `class Lookup`, `class ExplicitShadow`, `class PrivateBaseDerived`, `class Shape`, `data class Box` — **all classes**, so all correctly still render `Class`. Unaffected, not silently edited.

### 4. Type-kind axis — the unreachable cells are unreachable *by construction*

`GetAggregateKindName(StructSymbol)`. `InterfaceSymbol`, `EnumSymbol`, `StructSymbol` are all **`sealed` siblings** deriving directly from `TypeSymbol` — so interface/enum **cannot** be passed. The binary is exhaustive, not an under-test. Covered: class, struct, generic instantiation. G# has no records (`IsRecord`/`RecordSymbol` grep = 0).

Design strength worth naming: the signature change `string className` → `StructSymbol implementer` makes the **compiler** force every call site — I measured **64** `CS7036` sites for `ReportInterfaceMethodNotImplemented`. Future descriptors cannot silently regress.

### 5. Hardcode census — the "lazy fix is the root-cause fix" question

`DiagnosticDescriptors.cs`, 378 descriptors both sides.

| | `main` | head |
|---|---|---|
| `"Class '{"` | 11 | **6** |
| `"Struct '{"` | 3 | **1** |

PR fixes 7 (GS0187/0188/0318/0319/0320/0331/0332) — **broader than the issue asked**; GS0331/0332 hardcoded *Struct*, the reverse bug. I checked each of the 7 survivors and **none is a live defect**:

- **GS0181, GS0214** — name the **base type**, always a class. Correct.
- **GS0378, GS0381** — inheritance-only. Probe: `struct S : S` → **GS0382**, not GS0378; the struct path is **masked by an earlier check**, so GS0378 only ever names real classes. (`class S : S` → `GS0378 Class 'S'`, correct.)
- **GS0382** — struct-specific by nature.
- **GS0215, GS0216** — **dead** (below).

### 6. NITs

**NIT-1 — GS0319 is unreachable dead code, so its template change is untestable.** `ReportInterfaceDefaultRemoved` (`DiagnosticBag.Reports.Declarations.cs:631`) has **zero call sites**. Mutant **M2** (revert GS0319 to `"Class '{1}'"`, asserted applied) **survived full, unfiltered Core.Tests: 7086 passed / 0 failed / 7087**. I nearly filed this as a SHOULD-FIX "changed without a test" — **compiler-counting inverted it**: with a positive control at **64** sites, GS0319/GS0215/GS0216 all count **0**. No binder path can fire it, so no test can be written. Dead code, not a test gap.

**NIT-2 — GS0215/GS0216 still hardcode "Class"** and were left untouched. Also 0 call sites. Harmless; worth deleting with GS0319's reporter as separate cleanup, not here.

**NIT-3 — docs.** Gate is clean: rows unchanged **387 / 385** on both trees, and 146 rows differ between `docs/diagnostics.md` and `website/docs/ref/diagnostics.md` on **both** trees with **identical sets** — the PR introduces zero new divergence. But that pre-existing 146/387 skew (e.g. GS0331 is prose in one, template in the other) means the bidirectional gate is weaker than it looks. Not this PR's debt.

### Suites, on the merged tree

- **Core.Tests: 7086 / 0 / 1 skipped / 7087** (5 m 56 s)
- **Compiler.Tests, filtered: 128 / 128** (4 m 28 s) — the **17** classes referencing any of the 7 changed IDs or their message text, incl. `Issue976StructImplementsInterfaceEmitTests`, `Issue2875DataClassSettableInterfaceTests`, `Issue2888InterfacePropertyTypeMismatchTests`.

**A filtered green is not a green — what I did NOT run:** full Compiler.Tests (~1h20m under fleet load), Cs2Gs.Tests, Interpreter.Tests, LSP.Tests. Rationale: the change alters emitted text **only when `IsClass == false`**, and I enumerated and ran every test class in the repo that references the affected IDs or their wording.

### Measurement errors I made — disclosed

1. **`uptime` read concurrently with a build, not before it** — load turned out **51.34**, above the 30 bar; the 31 s build had already completed. I held the following suite and waited it down to 15.97. Rule violated in spirit; result unaffected (`BUILD_RC=0`, mtimes fresh).
2. **md5 did not round-trip** after restore (`981039df` vs baseline `34af035d`) despite `git status --porcelain`=0. Cause: interleaved `dotnet test` rebuilds obj without CIB. Re-verified the round-trip **behaviourally** — all 3 drivers × 3 shapes back to fixed output. md5 proves nothing about freshness; mtime and behaviour do.
3. **False positive of my own**: first docs check flagged GS0318/0319/0331 "DIVERGENT" — I had **no baseline**. Symmetric re-measurement showed the same 146 rows divergent on **both** trees, sets identical. Withdrawn.
4. First compiler-count probe **aborted** on a wrong control name; the assertion fired **before** the write, tree left clean. Re-run with a valid control (64).

Every probe compiled into its own directory asserted empty. All mutants asserted applied. Worktrees `/tmp/r5/{m,b}` porcelain 0; no protected checkout touched.

PR #3079, opened 2026-08-02
Fix iterator state-machine constraint remapping

### LINKED ISSUE #3067: Generic iterator state-machine constraint encodes method parameter in class scope

DavidObando:
## Summary

- resolve synthesized iterator generic constraints while state-machine remap is still active
- preserve class-scoped `Var` references instead of leaking method-scoped `MVar` references into emitted state-machine metadata
- add metadata, load/run, sync/async, and three-driver regression coverage

Fixes #3067

## Root cause and fix

State-machine `TypeDef` emission runs inside `PushSmRemap`, but pending `GenericParamConstraint` rows were resolved after that remap was disposed. Capture the pending-row start and call existing `PreResolveReifiedGenericConstraints` before leaving the remap scope. Production change: one hunk, two lines.

Compiler-count audit used a temporary required parameter on `GetIteratorTypeParametersInScope`. Compiler reported exactly 2 unique `CS7036` call sites:

- synchronous iterator synthesis (`StateMachineEmitter.cs:487`)
- async iterator synthesis (`StateMachineEmitter.cs:735`)

Direct `T : U` source constraints are currently rejected by the binder with `GS0153`; supported cross-method-parameter coverage uses `TBox Pair[T, U]` and verifies both state-machine parameter references.

## Emitted metadata assertions

Tests emit, then use `Assembly.Load(File.ReadAllBytes(...))` and `Assembly.GetTypes()`. They assert:

- exact state-machine generic arity and parameter names
- class and interface constraints reference the state-machine class's own `Type` parameter objects
- constructed constraints encode `Box<!0>`, `Pair<!0,!1>`, and mixed enclosing/method slots
- nested owners encode `[TOuter, TInner, TMethod]`
- `new()` preserves `DefaultConstructorConstraint`
- enclosing-type self constraints remain correctly scoped
- non-generic iterator stays non-generic
- async iterator uses the same correct mapping
- emitted assemblies load and run in a bounded child process

Covered shapes: class, interface/self, `new()`, two method parameters, generic method on generic class, nested generic type, enclosing type parameter, non-generic control, async iterator.

## Driver matrix

No source is wrapped in `func Main()`.

| position | bare `gsc` evaluate | `gsc /out:` emit + run | `gsi` |
|---|---|---|---|
| top-level | rc0, `66` | compile rc0, run rc0, `66` | rc0, `66` |
| in-function | rc0, `77` | compile rc0, run rc0, `77` | rc0, `77` |

Before fix, both emit cells compiled successfully but runtime exited 134 with `BadImageFormatException`; evaluate and interpreter cells ran.

## Product mutation

| production hunk | mutant | result |
|---|---|---|
| pre-resolve pending constraints inside state-machine remap | remove both added lines | 8 failed / 5 passed of 13 compiler cases |

Killed cases: class, interface, cross-method, generic-owner, nested-owner, async metadata, top-level emit, in-function emit. Failures surface as `BadImageFormatException`. Restored result: 13/13 compiler cases and 2/2 interpreter cases pass.

## Validation

Validated against `main` `33bea4afeef197c89da73f9edf0b20c36180f80f`.

- full suite: 14,272 passed, 1 skipped, 0 failed
- cs2gs gate: 1,863 passed (non-zero positive control)
- diagnostics/docs gate: 4 passed
- full CIB solution build: 0 warnings, 0 errors
- Core/Compiler/Repl `GSharp.Core.dll` mtimes matched
- `git merge-tree --write-tree origin/main HEAD`: clean; merged tree `ecb04266ceb33d837775b2c45727f994e1d8190f`, byte-identical to HEAD tree

### DIFF
--- a/src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs
@@ -3181,7 +3181,9 @@ internal sealed class ReflectionMetadataEmitter
             // encoding field signatures and interface implementations.
             using (this.remaps.PushSmRemap(c))
             {
+                var gpRowStart = this.emitCtx.PendingGenericParameters.Count;
                 this.typeDefEmitter.EmitNestedStructTypeDef(c, structFirstFieldRow[c], classCtorRows[c]);
+                this.PreResolveReifiedGenericConstraints(gpRowStart);
 
                 if (this.stateMachines.IteratorStateMachineInfos.TryGetValue(c, out var iteratorInfo))
                 {
--- a/test/Compiler.Tests/Emit/Issue3067IteratorConstraintRemapTests.cs
+++ b/test/Compiler.Tests/Emit/Issue3067IteratorConstraintRemapTests.cs
@@ -1,0 +1,479 @@
+// <copyright file="Issue3067IteratorConstraintRemapTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using GSharp.Compiler;
+using Xunit;
+
+namespace GSharp.Compiler.Tests.Emit;
+
+[Collection("Issue3067Console")]
+public class Issue3067IteratorConstraintRemapTests
+{
+    public static IEnumerable<object[]> ConstraintCases()
+    {
+        yield return Case(
+            "class-constraint",
+            "ClassValues",
+            "11",
+            ConstraintShape.ClassReferencesMethodParameter,
+            """
+            package Issue3067.ClassConstraint
+            import System
+            open class Box[T any] {}
+            class IntBox : Box[int32] {}
+            func ClassValues[T any, TBox Box[T] init()](value T) sequence[T] {
+                yield value
+            }
+            for value in ClassValues[int32, IntBox](11) {
+                Console.WriteLine(value)
+            }
+            """);
+
+        yield return Case(
+            "interface-constraint",
+            "InterfaceValues",
+            "22",
+            ConstraintShape.InterfaceReferencesSelf,
+            """
+            package Issue3067.InterfaceConstraint
+            import System
+            func InterfaceValues[T IComparable[T]](value T) sequence[T] {
+                yield value
+            }
+            for value in InterfaceValues[int32](22) {
+                Console.WriteLine(value)
+            }
+            """);
+
+        yield return Case(
+            "new-constraint",
+            "NewValues",
+            "33",
+            ConstraintShape.DefaultConstructor,
+            """
+            package Issue3067.NewConstraint
+            import System
+            func NewValues[T init()](value T) sequence[T] {
+                yield value
+            }
+            for value in NewValues[int32](33) {
+                Console.WriteLine(value)
+            }
+            """);
+
+        yield return Case(
+            "cross-method-constraint",
+            "CrossValues",
+            "44",
+            ConstraintShape.ClassReferencesTwoMethodParameters,
+            """
+            package Issue3067.CrossConstraint
+            import System
+            open class Pair[A any, B any] {}
+            class IntStringPair : Pair[int32, string] {}
+            func CrossValues[T any, U any, TBox Pair[T, U] init()](value U) sequence[U] {
+                yield value
+            }
+            for value in CrossValues[int32, string, IntStringPair]("44") {
+                Console.WriteLine(value)
+            }
+            """);
+
+        yield return Case(
+            "generic-owner-and-method",
+            "MixedValues",
+            "55",
+            ConstraintShape.ClassReferencesClassAndMethodParameters,
+            """
+            package Issue3067.GenericOwner
+            import System
+            open class Pair[A any, B any] {}
+            class IntStringPair : Pair[int32, string] {}
+            class Maker[TClass any] {
+                func MixedValues[TMethod any, TBox Pair[TClass, TMethod] init()](value TMethod) sequence[TMethod] {
+                    yield value
+                }
+            }
+            for value in Maker[int32]().MixedValues[string, IntStringPair]("55") {
+                Console.WriteLine(value)
+            }
+            """);
+
+        yield return Case(
+            "nested-generic-owner",
+            "NestedValues",
+            "66",
+            ConstraintShape.NestedOwnerAndMethod,
+            """
+            package Issue3067.NestedOwner
+            import System
+            class Outer[TOuter any] {
+                struct Inner[TInner any] {
+                    func NestedValues[TMethod IComparable[TMethod]](value TMethod) sequence[TMethod] {
+                        yield value
+                    }
+                }
+            }
+            var inner = Outer[int32].Inner[string]{}
+            for value in inner.NestedValues[int32](66) {
+                Console.WriteLine(value)
+            }
+            """);
+
+        yield return Case(
+            "enclosing-parameter-constraint",
+            "EnclosingValues",
+            "77",
+            ConstraintShape.EnclosingParameterReferencesSelf,
+            """
+            package Issue3067.EnclosingConstraint
+            import System
+            class Owner[T IComparable[T]] {
+                func EnclosingValues(value T) sequence[T] {
+                    yield value
+                }
+            }
+            for value in Owner[int32]().EnclosingValues(77) {
+                Console.WriteLine(value)
+            }
+            """);
+
+        yield return Case(
+            "non-generic-control",
+            "PlainValues",
+            "88",
+            ConstraintShape.NonGeneric,
+            """
+            package Issue3067.NonGeneric
+            import System
+            func PlainValues(value int32) sequence[int32] {
+                yield value
+            }
+            for value in PlainValues(88) {
+                Console.WriteLine(value)
+            }
+            """);
+
+        yield return Case(
+            "async-iterator",
+            "AsyncValues",
+            "99",
+            ConstraintShape.InterfaceReferencesSelf,
+            """
+            package Issue3067.AsyncInterfaceConstraint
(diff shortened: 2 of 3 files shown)


DavidObando:
## Review: PR #3079 — VERDICT: **MERGEABLE**

Measured on the **merged tree**: `main` **`4f7559dd`** + head **`816c0ead`** → merge **`471b5d33`**. `MERGE_RC=0`, `merge-tree` conflict markers **0**, **`BUILD_RC=0`**, Release `-p:ContinuousIntegrationBuild=true` **0 warnings / 0 errors**. `GSharp.Core.dll` md5 `210b10a96bf825be93133d8809376712` (BASE `efc6018078339e2aa3aa23c23aa0152f` — the 2-line change is live), all three copies (`Core/`, `Compiler/`, `Repl/`) mtime-verified before every probe, every probe compiled into its **own directory asserted empty**.

Note the merge base is `33bea4af`, which is where the PR body's validation was run. `main` has since moved to `4f7559dd` (#3074), so I re-anchored and re-measured rather than carry those numbers forward.

**No blockers. This is the best-shaped fix I have reviewed today**: 2 production lines, reusing an existing helper, at a placement I tried and failed to loosen.

---

### Full suites, merged tree, unfiltered

| project | result |
|---|---|
| `Core.Tests` | **7074 passed / 0 failed / 1 skipped / 7075** |
| `Compiler.Tests` | **4130 passed / 0 failed / 4130** (1 h 15 m) |
| `Interpreter.Tests` | **623 passed / 0 failed / 623** |

I enumerated all 7 test projects rather than trusting a filter; the three above are the ones this change can reach. I did **not** independently run the `cs2gs` gate — the body reports 1,863 passed and I am relying on CI for that column. Stated as a gap, not a result.

### The issue's verbatim repro, run byte-for-byte first

`issue_verbatim.gs`, md5 `9f60f9adeeb6fe6aef8fd067b97d693d`, copied unmodified from #3067:

| tree | `gsc /out:` | run |
|---|---|---|
| BASE `4f7559dd` | rc=0 | **rc=134, `System.BadImageFormatException` (0x8007000B)** |
| merged head | rc=0 | **rc=0, prints `66`** |

That is the issue's reported exit code and exception, reproduced and then fixed. I ran this before any variation.

### Both directions — false-positive corpus

Eight correct programs that compile on `main` and must keep compiling, compared on rc **and** diagnostic set, with distinct values:

| probe | shape | BASE | head |
|---|---|---|---|
| `fp1_nongeneric` | non-generic iterator | `11` rc0 | `11` rc0 |
| `fp2_noconstraint` | generic iterator, no constraint | `22` rc0 | `22` rc0 |
| `fp3_classconstraint` | `T class` | `33` rc0 | `33` rc0 |
| `fp4_structconstraint` | `T struct` | `44` rc0 | `44` rc0 |
| `fp5_ifaceself` | `T IComparable[T]` | **rc134 `BadImageFormatException`** | `55` rc0 |
| `fp6_enclosing` | iterator in generic class | `66` rc0 | `66` rc0 |
| `fp7_plaingeneric` | non-iterator, cross-param constraint | `77` rc0 | `77` rc0 |
| `fp8_inferred` | inferred type args | `88` rc0 | `88` rc0 |

**Zero false positives** — no program that compiled on `main` stopped compiling, and no new diagnostic appeared anywhere. `fp5` moves the safe way: I filed it as a control and it was *already broken* on BASE, so the defect is wider than #3067's title suggests (a **single** type parameter with a self-referencing interface constraint, no cross-parameter reference at all, also emits invalid IL). The PR already covers this — `ConstraintShape.InterfaceReferencesSelf` is in the test matrix — so this is confirming evidence, not a gap.

### Mutation — per hunk, asserted applied

**M1 — remove both added lines** (the author's mutant). Applied under a hard assertion that the pattern matched exactly once, then re-grepped to confirm only the two pre-existing call sites (`:2602`, `:2791`) remained. `BUILD_RC=0`, md5 moved to `babdb4807a24b0350c48bf01f2769b81`.

| probe | M1 |
|---|---|
| `issue_verbatim` | **`BadImageFormatException`** |
| `toplevel_repro` | **`BadImageFormatException`** |
| `async_iter_cross` | **`BadImageFormatException`** |
| `fp5_ifaceself` | **`BadImageFormatException`** |
| `fp2_noconstraint` (control) | `22` rc0 — unaffected |

**M2 — keep the call, move it outside the `using (PushSmRemap(c))` block.** This is the mutant that matters, because it tests the PR's actual claim (*"while the state-machine remap is still active"*) rather than merely whether the helper is called. `BUILD_RC=0`, 0 warnings.

**M2 is also killed** — all four shapes return to `BadImageFormatException`. So the *scoping* is load-bearing, not incidental: the fix cannot be relocated to the more obvious spot just after the loop body. The two lines are both minimal and correctly placed. I went looking for a looser shape and the compiler refused it.

### Driver axis — confirms the author's matrix, and shows why CI shape matters

| source shape | bare `gsc` evaluate | `gsc /out:` emit+run | `gsi` |
|---|---|---|---|
| issue's `func Main()` file, BASE | rc0, `Success.`, **no output** | rc0 compile, **rc134** | rc0, **EMPTY OUTPUT** |
| top-level variant, BASE | rc0, prints `66` | rc0 compile, **rc134** | rc0, prints `66` |

Both interpreting columns print the **correct** `66` on the broken BASE — they are structurally blind to this defect — and the issue's own `func Main()`-only shape is a **guaranteed vacuous pass under `gsi`** (rc0, empty output). Only `gsc /out:` + execute can see it. The PR's tests correctly avoid this trap: `InFunctionSource` ends with a top-level `Run()` call, so the evaluate cell genuinely executes, and `grep -c "func Main()"` is **0** across both new test files.

### Sibling-site check — hypothesis raised, then refuted by measurement

`TypeDefEmitter` is `sealed`, not `partial`, and there is exactly **one** producer of `PendingGenericParameters` rows (`TypeDefEmitter.cs:2101`), reached from exactly **two** call sites: `ReflectionMetadataEmitter.cs:3185` (SM classes — **fixed here**) and `:3211` (SM structs for async methods/lambdas — **not touched**). Same `PushSmRemap` → `EmitNestedStructTypeDef` → dispose shape, so I expected a latent twin.

I could not fire it. Five async shapes, all rc0 with correct values on **both** BASE and head: `async func` with `Box[T] init()` returning `T` and returning `Task[T]`, `Pair[T,U] init()` across two method parameters, an async lambda inside a constrained generic function, and an async method on a generic class constrained against the class parameter. Site `:3211` was **never affected**; the defect is specific to state-machine *classes*. Reporting this because it is the obvious objection to a one-site fix and it does not hold — the fix is complete for the sites that can exhibit the bug.

---

## NIT 1 — the issue's own `func Main() int32` shape is not in the regression suite

The new tests use top-level statements and a top-level `Run()` call. #3067's repro is `func Main() int32` with a `return` out of the `for` body. I verified that exact shape manually (table above), but nothing in the suite pins it, and `func Main()` is a distinct entry-point lowering path. Adding the issue's source verbatim as one more `ConstraintCases` row is nearly free and makes the regression test match the report.

## NIT 2 — third copy of the same pre-resolve idiom

`capture gpRowStart → emit TypeDef → PreResolveReifiedGenericConstraints(gpRowStart)` now appears at `ReflectionMetadataEmitter.cs:2600-2602`, `:2789-2791`, and `:3183-3186`, plus a fourth hand-inlined variant in `FunctionEmitter.cs:791-795`. Not a correctness issue and **not something I would hold this PR for** — the 2-line fix is the right call here — but the next occurrence is the one that gets forgotten. A small `EmitTypeDefWithResolvedConstraints(...)` wrapper would make the invariant "rows are resolved before the remap scope closes" structural instead of remembered.

## NIT 3 — `InterfaceConstraintType` is a misnomer

`PendingGenericParameter.InterfaceConstraintType` (`PendingGenericParameter.cs:49`) is populated from `tp.ConstraintReferenceType` (`TypeDefEmitter.cs:2106`) and carries **base-class** constraints too — which is exactly what #3067's `TBox Box[T]` is. Pre-existing, out of scope for this PR, but it cost me a detour: the field name says the fix cannot possibly address a class constraint, and the fix does.

---

### Disclosures

- My probe `s2_asynclambda` was initially **void** — I declared the async lambda's return as `Task[T]` and got `GS0155`; async lambdas lower `R` → `Task[R]`, so it must be declared `T`. Corrected and re-run; the corrected probe passes on both trees. The first version is excluded from all counts.
- My "unfixed sibling loop at `:3211`" was a **hypothesis I could not fire**, across five shapes and both trees. Reported as refuted rather than quietly dropped.
- I did not independently run the `cs2gs` gate.
- Tree restored: `git status --porcelain` empty; rebuild md5 **round-trips exactly** to `210b10a96bf825be93133d8809376712`.

PR #3080, opened 2026-08-02
Fix generic CLR static stores

### LINKED ISSUE #3076: gsc silently drops assignment to a generic type's shared field (rc0, no diagnostic, read-back is default)
`gsc` accepts an assignment to a static/shared field of a **generic** type, exits 0, emits no diagnostic — and the assignment does not happen.

### Repro shape

```gsharp
// GenericStaticSlot[T] declares a shared field `Value int32`
GenericStaticSlot[int32].Value = 222
print(GenericStaticSlot[int32].Value)
```

### Observed

- `gsc` exit code **0**
- **no diagnostics** (`0 errors | 0 warnings | 0 info`)
- read-back prints **`0`**, i.e. the default value

### Why this is "dropped", not "erased"

If the store had been emitted against a different (per-instantiation) static slot, the read-back would still have to observe *some* consistent value or fail. Reading back `0` immediately after storing `222`, in the same program, with no diagnostic, means the store was **discarded during lowering/emit** rather than routed to a different slot.

This is the worst class of compiler bug we have in this repo: a **silent wrong answer at rc=0**. The user gets no signal at all.

### Expected

Either the assignment takes effect and the read-back prints `222`, or the construct is rejected with a diagnostic. Silently dropping it is not an acceptable outcome.

### Notes

- Found incidentally while reviewing interpreter generic CLR type reification; **pre-existing**, not caused by any open PR.
- Cover all three driver columns when fixing: bare `gsc` **evaluates**, `gsc /out:` **emits**, `gsi` interprets. They must agree.
- `gsi` only ever executes top-level code, so keep the repro top-level — a `func Main()`-only file under `gsi` is rc0 with empty output and would pass vacuously.
- Use distinct, non-default values (`222`, not `0`) in any regression test, so "correct" cannot be confused with "default".

DavidObando:
## Summary

- parent imported generic static property setters and fields at the symbolic constructed container
- preserve `StaticContainerType` through normal, side-effect, async, and `??=` rewrites
- make `staticContainerType` required so future constructor sites cannot silently discard it
- cover top-level and generic-function stores across evaluator, emitter, and interpreter; the fixed defect itself is emit-only

## Root cause

The store was misrouted, not dropped. With `GenericStaticSlot[object]` initialized to `333`, a generic `GenericStaticSlot[T].Value = 222` left the requested `int32` slot unchanged and changed the `object` slot to `222`. Direct closed stores already worked.

Static reads used `StaticContainerType`; static writes ignored it and emitted member references against the erased `object` construction. Four reconstruction sites could also discard the metadata. A required-parameter compiler probe counted 20 constructor sites; all now pass the container explicitly, including explicit `null` where no symbolic static container exists.

## Validation

Measured on the merged tree with `main` `749775553eb964b6d4b8144df7f6c76193f343c8`:

- `dotnet build GSharp.sln -c Release -m:1 -p:ContinuousIntegrationBuild=true`: 0 warnings, 0 errors
- Compiler issue tests: 7 passed
- Interpreter issue tests: 2 passed
- Async spill metadata test: 1 passed
- clean-directory `{top-level, in-function} × {bare gsc, gsc /out:, gsi}` matrix: all six cells returned distinct expected values
- `git merge-tree --write-tree origin/main HEAD`: clean

Coverage includes independent value/reference/object slots; nested and two-parameter generics; same-method, another-method, and fixture-assembly read-back; non-generic controls; properties and fields; side-effect spilling; async spilling; and null-coalescing assignment.

Fixes #3076

### DIFF
--- a/src/Core/CodeAnalysis/Binding/BoundClrPropertyAssignmentExpression.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundClrPropertyAssignmentExpression.cs
@@ -26,9 +26,9 @@ public sealed class BoundClrPropertyAssignmentExpression : BoundExpression
         MemberInfo member,
         BoundExpression value,
         TypeSymbol resultType,
+        TypeSymbol staticContainerType,
         TypeParameterSymbol constrainedReceiverTypeParameter = null,
-        TypeSymbol constrainedInterfaceType = null,
-        TypeSymbol staticContainerType = null)
+        TypeSymbol constrainedInterfaceType = null)
         : base(syntax)
     {
         Receiver = receiver;
--- a/src/Core/CodeAnalysis/Binding/BoundTreeRewriter.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundTreeRewriter.cs
@@ -1681,6 +1681,7 @@ public abstract class BoundTreeRewriter
             node.Member,
             value,
             node.Type,
+            node.StaticContainerType,
             node.ConstrainedReceiverTypeParameter,
             node.ConstrainedInterfaceType);
     }
--- a/src/Core/CodeAnalysis/Binding/ConversionClassifier.cs
+++ b/src/Core/CodeAnalysis/Binding/ConversionClassifier.cs
@@ -3153,7 +3153,8 @@ internal sealed class ConversionClassifier
                     receiver,
                     slot.TargetClrMember,
                     value,
-                    slot.TargetType);
+                    slot.TargetType,
+                    staticContainerType: null);
             }
 
             statements.Add(new BoundExpressionStatement(expression.Syntax, assignment));
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs
@@ -457,7 +457,7 @@ internal sealed partial class ExpressionBinder
             var value = BindExpression(initSyntax.Value);
             var converted = conversions.BindConversion(initSyntax.Value.Location, value, instTargetSymbol);
             var receiverExpr = new BoundVariableExpression(initSyntax, receiverLocal);
-            return new BoundClrPropertyAssignmentExpression(initSyntax, receiverExpr, instanceMember, converted, instTargetSymbol);
+            return new BoundClrPropertyAssignmentExpression(initSyntax, receiverExpr, instanceMember, converted, instTargetSymbol, staticContainerType: null);
         }
 
         if (receiverType is StructSymbol structSymbol)
@@ -518,7 +518,7 @@ internal sealed partial class ExpressionBinder
                     var value = BindExpression(initSyntax.Value);
                     var converted = conversions.BindConversion(initSyntax.Value.Location, value, inhTargetSymbol);
                     var receiverExpr = new BoundVariableExpression(initSyntax, receiverLocal);
-                    return new BoundClrPropertyAssignmentExpression(initSyntax, receiverExpr, inhMember, converted, inhTargetSymbol);
+                    return new BoundClrPropertyAssignmentExpression(initSyntax, receiverExpr, inhMember, converted, inhTargetSymbol, staticContainerType: null);
                 }
             }
 
@@ -611,7 +611,7 @@ internal sealed partial class ExpressionBinder
                 {
                     var inheritedValue = BindAssignmentRhs(syntax.Value, inheritedTarget);
                     var inheritedConverted = conversions.BindConversion(syntax.Value.Location, inheritedValue, inheritedTarget);
-                    return new BoundClrPropertyAssignmentExpression(null, receiver: null, inheritedStaticMember, inheritedConverted, inheritedTarget);
+                    return new BoundClrPropertyAssignmentExpression(null, receiver: null, inheritedStaticMember, inheritedConverted, inheritedTarget, staticContainerType: null);
                 }
             }
 
@@ -665,7 +665,7 @@ internal sealed partial class ExpressionBinder
             _ = staticTargetType;
             var staticValue = BindAssignmentRhs(syntax.Value, staticTargetSymbol);
             var staticConverted = conversions.BindConversion(syntax.Value.Location, staticValue, staticTargetSymbol);
-            return new BoundClrPropertyAssignmentExpression(null, receiver: null, staticMember, staticConverted, staticTargetSymbol);
+            return new BoundClrPropertyAssignmentExpression(null, receiver: null, staticMember, staticConverted, staticTargetSymbol, staticContainerType: null);
         }
 
         var variable = BindVariableReference(receiverName, syntax.Receiver.Location);
@@ -756,7 +756,7 @@ internal sealed partial class ExpressionBinder
             _ = instWritable;
             _ = instTargetType;
             var instConverted = conversions.BindConversion(syntax.Value.Location, BindValue(instTargetSymbol), instTargetSymbol);
-            return new BoundClrPropertyAssignmentExpression(null, assignmentReceiver, instanceMember, instConverted, instTargetSymbol);
+            return new BoundClrPropertyAssignmentExpression(null, assignmentReceiver, instanceMember, instConverted, instTargetSymbol, staticContainerType: null);
         }
 
         // Issue #1068: an interface-typed variable receiver (`d.W = v` where
@@ -861,8 +861,9 @@ internal sealed partial class ExpressionBinder
                         clrProperty,
                         constrainedConverted,
                         propertyType,
-                        tpVarType,
-                        declaringInterface);
+                        staticContainerType: null,
+                        constrainedReceiverTypeParameter: tpVarType,
+                        constrainedInterfaceType: declaringInterface);
                 }
             }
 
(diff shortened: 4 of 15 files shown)


DavidObando:
## Rubber-duck review — PR #3080 (`fix/3076-generic-static-field-store`)

**Verdict: MERGEABLE** — 0 blockers, 1 SHOULD-FIX, 3 NITs.

Measured against `main` **`4f7559dd`**, PR head **`54c3d641`**, computed merge base **`33bea4af`** (the PR is not rebased on current `main`; #3074, #3036, #2963 and #3072 landed after its base). Real merge head: **`648b2e63`**. Merged-tree build: **`BUILD_RC=0`, 0 warnings** — `merge-tree` clean was treated as a floor, not a proof.

All probes were compiled into their own directory, asserted empty (`ls -A` = 0), against binaries whose `Core/`, `Compiler/` **and** `Repl/` copies were mtime-verified fresh (`ref/` excluded). Probes used my own C# `ProbeLib`, deliberately not the PR's harness.

### 1. The defect reproduces — but not through the shape the issue describes

Ran the issue's repro before anything else. Two corrections:

- `print(...)` is not a G# builtin (rc=1); substituted `Console.WriteLine`.
- The issue's **literal** shape does not reproduce. Measured at base `4f7559dd`, emit column:

| probe | base `4f7559dd` | merged `648b2e63` |
|---|---|---|
| `GenericStaticSlot[int32].Value = 222` (issue's literal shape) | **222 — works** | 222 |
| store inside `func Store[T]()` | **0, rc0, no diagnostic** | **222** |
| `[object]`=333, then generic store | **`[int32]`=0, `[object]`=222** | **`[int32]`=222, `[object]`=333** |

So the store is **misrouted to the erased `object` construction, not dropped**, and direct closed stores already worked. **The PR body already says exactly this** ("The store was misrouted, not dropped… Direct closed stores already worked") — the author found it and corrected the report rather than coding to the issue text. That is the right behaviour and I want it on the record.

Mechanism, confirmed structurally: a **closed** construction yields a `FieldInfo` whose `DeclaringType` is already the constructed type, so the plain `GetFieldReference` path is correct; only a **symbolic** construction (`[T]`) erases to `object` and needs `StaticContainerType`.

### 2. Root cause validated by checking producers, not consumers

There are exactly **two** producers of a non-null `staticContainerType` in the tree — `ExpressionBinder.Assignments.cs:2350` and `ExpressionBinder.Calls.Invocation.cs:722` — and both pass `receiver: null`. The binder was therefore always producing the right metadata; only the emitter (ignoring it on stores) and four reconstruction sites (dropping it) were wrong. I independently counted **20** `BoundClrPropertyAssignmentExpression` construction sites, matching the PR's required-parameter probe.

### 3. Per-hunk mutation: 6/6 killed

Product mutants (condition/argument inversion, never `if (true)`), each **asserted applied** before building, each reverted with an md5 round-trip to `04cd7c39`:

| # | site | result |
|---|---|---|
| M1 | `BoundTreeRewriter.cs:1686` | **killed** — `…ThroughAsyncSpilling` |
| M2 | `StatementBinder.Jumps.cs:878` | **killed** — `…ThroughNullCoalescingAssignment` |
| M3 | `MethodBodyEmitter.MemberAccess.cs:1176` (setter) | **killed** — 4 tests |
| M4 | `MethodBodyEmitter.MemberAccess.cs:1185` (field) | **killed** — 4 tests |
| M5 | `SpillSequenceSpiller.cs:510` | **killed** — Core unit test only (see NIT 1) |
| M6 | `SideEffectSpiller.cs:280` | **killed** — 2 tests |

I initially suspected the 19-line `StatementBinder.Jumps.cs` hunk was dead, because `GenericStaticSlot[T].Value += n` fails to bind (`GS0158`). That was an **adjacent shape**: the hunk lives in `TryBuildNullCoalescingReadWrite`, serving `??=`, and M2 kills it. Hypothesis discarded on measurement.

### 4. Shapes in neither the changed nor the unchanged rows

All correct on merged, `[object]` never collaterally written (own probes, distinct values):

| shape | result |
|---|---|
| reference / value / `object` slot independence | 111 / 222 / 333 |
| nested **symbolic** `GenericStaticSlot[List[T]]` | 333, others 0 |
| store inside a **generic class** method | 811 |
| type parameter **relayed** `Outer[T]` → `Inner[U]` | 822 |
| property via generic class method | 833 |
| `++` increment through a type parameter | 221 |
| read-back from a different method | 666 |
| non-generic control | 999 |

### 5. #3072 interaction — proven behaviourally, not from file lists

#3080 edits `MethodBodyEmitter.MemberAccess.cs`, which #3072 also changed and which landed in `main` after this PR's base. I re-ran all nine of my #3072 probes on the merged #3080 tree: `p1`–`p4`, `p6`–`p8` correct (11/22/33/44/66/77/88). `p5` and `p9` abort — **byte-identical to their results on the #3072-merged tree** (`Field not found: 'P5.Derived\`1.v'`), i.e. the known pre-existing composite-literal residual, **not a regression**. I checked the stored prior output rather than assuming, because on shape alone `p5` looked like a new failure.

### 6. Suites on the merged tree `648b2e63`

- Core.Tests **7075 passed / 0 failed / 1 skipped / 7076**
- Interpreter.Tests **623 passed / 0 failed**
- Compiler.Tests (full, unfiltered) — see follow-up comment

### 7. Contamination and vacuous-pass hazards: both clear

- The matrix tests compile a sibling C# assembly and run each G# program **in a separate process**, so process-wide static slots cannot leak between cases; each case asserts the `[object]` slot **retains** its own distinct value.
- Test sources are **top-level statements**, not `func Main()`-only, so bare `gsc` (evaluate) genuinely executes them — no vacuous rc0.

---

## SHOULD-FIX

**1. The sentinel that caused this bug is still optional.** `BoundClrPropertyAssignmentExpression.cs:31` declares `TypeSymbol staticContainerType = null`. An optional parameter is exactly what let four reconstruction sites drop this metadata silently with no compiler complaint — the defect being fixed here. The PR fixes the four known droppers but leaves the mechanism intact, so a fifth reconstruction added later reintroduces the same silent wrong-answer. The PR's own required-parameter probe shows the tooling to settle this already exists. Cheapest remedy: make the parameter **required** on that ctor and let the compiler enumerate the 20 sites, or assert at construction that a null receiver on an imported constructed-generic container implies a non-null `StaticContainerType`.

## NITs

**1. `SpillSequenceSpiller.cs:510` is pinned only by a node shape the binder cannot produce.** `Issue3076SpillSequenceSpillerTests.cs:22` builds an assignment with a **non-null receiver** *and* a non-null `staticContainerType`; both producers of a non-null container pass `receiver: null`, so that combination does not occur. Measured: with M5 applied the **end-to-end async test still passes 7/7** — only the synthetic unit test dies. The hunk is correct-by-construction and cheap, so this is not a defect, but the unit test should not be read as end-to-end async coverage. Either add a source-level probe that reaches `SpillTwoOperand` with a null receiver, or state in the test that the hunk is defensive.

**2. The defect was emit-only; the evaluator and interpreter tests are guards, not fix evidence.** I ran the PR's tests against **base production code**: Compiler 4/7 fail, the Core spiller test fails — but **both `…_Evaluate` cases and both Interpreter tests pass at base**. The summary bullet "cover top-level and generic-function stores across evaluator, emitter, and interpreter" is true as written but reads as if three drivers were fixed. One clause noting the defect was emit-only would prevent that.

**3. CLI reachability limit worth recording.** With `/r:`, bare `gsc` fails `GS9999` (`MetadataLoadContext`) and `gsi` rejects `/r:` outright, so for imported types only the emit column is reachable from the command line; the evaluator/interpreter columns are only exercisable in-process, as the PR does. Not this PR's problem, but it bounds what any future CLI repro of an imported-type bug can show.

---

### Out of scope, for the repo owner to file if wanted

Compound assignment to **any** imported CLR static member fails to bind: `PlainStaticSlot.Value += 66` → `GS0158: Cannot find member Value.` This reproduces for the **non-generic** control and involves no `StaticContainerType`, so it is independent of this PR; `??=` on the same members binds fine. Distinct defect.

### Disclosures

- One rebuild was started at load 37.16 (the `uptime` reading returned in the same invocation). The artifact hashes identically to the known-good merged build `04cd7c39`, so the measurement stands, but it breached the sub-30 rule.
- Every mutant was asserted applied before building; no mutant produced a no-result.
- No probe reused a directory containing a prior build of the same assembly identity.
- I did not write to any worktree I do not own. The shared `gsharp` checkout contains another agent's untracked `i3076-probes/`; I left it untouched.

DavidObando:
### Follow-up: full Compiler suite on the merged tree

Completing section 6 of my review above. On merged head **`648b2e63`** (PR `54c3d641` merged into `main` `4f7559dd`):

- **Compiler.Tests, full and unfiltered: 4124 passed / 0 failed / 0 skipped / 4124.**

Full picture on the merged tree, all three projects unfiltered — a filtered green is not a green, and the async-spiller test lives in **Core.Tests**, a project a `Compiler.Tests` filter cannot reach:

| suite | result |
|---|---|
| Core.Tests | 7075 passed / 0 failed / 1 skipped / 7076 |
| Interpreter.Tests | 623 passed / 0 failed / 623 |
| Compiler.Tests | 4124 passed / 0 failed / 4124 |

Verdict unchanged: **MERGEABLE** — 0 blockers, 1 SHOULD-FIX (optional `staticContainerType` parameter, `BoundClrPropertyAssignmentExpression.cs:31`), 3 NITs.

DavidObando:
Reviewer follow-up addressed on rebased head `317c8648`, merged against `main` `74977555`:

- `staticContainerType` is now required; compiler sentinel re-counted exactly **20** constructor sites, all explicit.
- Async spiller test comment now states its sentinel node is a defensive metadata-contract shape, not source-level end-to-end coverage.
- PR body now states the fixed defect is emit-only.
- Test matrix now varies read-back site: same generic method, another generic method, and fixture-assembly static method; value/reference/object/nested/two-parameter slots remain independent.
- Shared-checkout `i3076-probes/` removed.
- Every rerun used a newly created, asserted-empty output directory. Results unchanged.

Newest merged-tree validation: `BUILD_RC=0`, Compiler 7/7, Interpreter 2/2, Core 1/1, `merge-tree` clean.

PR #3097, opened 2026-08-02
Fix imported nested enum symbolic arguments

### LINKED ISSUE #2947: GS0156 displays erased nested-enum generic arguments

DavidObando:
## Summary
- preserve a constructed imported receiver's symbolic type arguments when resolving nested types
- recover literal enum field types from the symbolic receiver instead of the erased reflection type
- add cross-assembly bound-tree, diagnostic, and top-level/in-function driver-matrix coverage

## Testing
- `dotnet build GSharp.sln -c Release -m:1`
- 8 issue tests pass across gsc evaluate, gsc emit, and gsi
- product mutations independently kill nested-receiver propagation, symbolic nested construction, literal recovery, and live-reflection self-type mapping
- `git merge-tree --write-tree origin/main HEAD`

Fixes #2947

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.Accessor.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.Accessor.cs
@@ -695,7 +695,8 @@ internal sealed partial class ExpressionBinder
     private ImportedClassSymbol CloseImportedNestedType(
         Type constructedOuter,
         ImportedClassSymbol nested,
-        ExpressionSyntax syntax)
+        ExpressionSyntax syntax,
+        ImportedTypeSymbol symbolicReceiver = null)
     {
         var nestedType = nested?.ClassType;
         if (constructedOuter?.IsConstructedGenericType != true
@@ -713,7 +714,10 @@ internal sealed partial class ExpressionBinder
         try
         {
             var closedType = nestedType.MakeGenericType(outerArguments);
-            return new ImportedClassSymbol(closedType, syntax, references: scope.References);
+            var symbolicNested = symbolicReceiver == null
+                ? null
+                : ImportedTypeSymbol.GetConstructed(closedType, nestedType, symbolicReceiver.TypeArguments);
+            return new ImportedClassSymbol(closedType, syntax, symbolicNested, scope.References);
         }
         catch (ArgumentException)
         {
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs
@@ -265,7 +265,14 @@ internal sealed partial class ExpressionBinder
                     // their constant value rather than emit `ldsfld`.
                     if (staticMember is FieldInfo litField && litField.IsLiteral)
                     {
-                        return new BoundLiteralExpression(null, litField.GetRawConstantValue(), TypeSymbol.FromClrType(litField.FieldType));
+                        var literalType = TypeSymbol.FromClrType(litField.FieldType);
+                        if (classSymbol.SymbolicReceiver != null)
+                        {
+                            literalType = ResolveStaticMemberTypeFromSymbolicReceiver(classSymbol.SymbolicReceiver, litField)
+                                ?? literalType;
+                        }
+
+                        return new BoundLiteralExpression(null, litField.GetRawConstantValue(), literalType);
                     }
 
                     var staticType = staticMember switch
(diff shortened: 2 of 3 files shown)


DavidObando:
Closing in favour of **#3112**, which targets the same issue (#2947) and has now merged.

This is not a rejection of the work — a rubber-duck reviewed both heads side by side and found them **behaviourally indistinguishable**. Both reproduce and fix #2947's verbatim cross-assembly repro (rc=1 `GS0156: … 'glib.Outer[object].Color' to 'glib.Outer[T].Color'` → rc=0), and both score **9/9** on a `{enum, struct, class} × {[T], [int32], [string]} × {depth-2, depth-3}` axis where the base scores 7/9. The duck's own hypothesis that this branch's missing arity guard would break `Outer[T].Inner[U]` was **probed and refuted** — rc=0 on all three drivers. It reported that as a killed hypothesis rather than as a finding, which is the right call.

They are genuinely different fixes, not a resubmission: different `src` subtrees (`01e00f82` vs `b4ead423`), and co-merging them gives **rc=1 with `CONFLICT (content)` in both files**. Only one could land.

**Why #3112 won:** I closed #2947 once already, wrongly, on a review that rendered five *adjacent* sites correctly. The rule that came out of that is that a closure requires the reported shape reproduced, and the evidence must pin the **binder**, not the renderer — the binder was demonstrably holding the erased `[object]` type while `SymbolDisplay` faithfully rendered it. #3112 carries a Core.Tests assertion on the bound symbol itself (`Assert.IsType<ImportedTypeSymbol>(literal.Type)`, `HasTypeParameterArgument`, `TypeArguments[0].Name == "T"`). This branch has three driver tests and no binding-level assertion. Given how this issue was closed the first time, I am keeping the one that makes a renderer-only regression impossible to reintroduce silently.

**Two things this branch did better, recorded so they are not lost:** it is the smaller diff (22/5 vs 35/14), and its `AreSame` short-circuit is a more general rule than #3112's `IsEnum` special case. It also leaves the imported-const path untouched, whereas #3112 widens it — which I am tracking as a follow-up. If the const widening causes trouble, this branch's approach is the fallback and I will come back to it.

Thanks — this was good work that lost a coin-flip on evidence, not on correctness.

DavidObando:
## NOT MERGEABLE — historical review at `36639253`

PR #3097 is now closed/superseded, but I reviewed its isolated patch (`0e8ebb72`) on the requested base. Product behavior is correct; persistent coverage is incomplete.

### BLOCKER

- `test/Interpreter.Tests/Issue2947ImportedNestedEnumSymbolicTests.cs:140-152,236-313` pins only a depth-1 imported enum. The general binder change at `src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.Accessor.cs:695-720` also fixes three additional cells that no committed test can see: depth-2 `Outer[T].Mid.Tone`, nested-generic `Outer[List[T]].Color`, and nested-generic depth-2 `Outer[List[T]].Mid.Tone`.
  - Independent 36-cell matrix (`{enum,struct,class} × {T,int32,List[T]} × {imported,same-assembly} × {depth 1,2}`), run through bare `gsc`, emitted `gsc`, and `gsi`: base had **6 GS0156** per driver; patch had **2**. The four removed failures are the reported cell plus those three unpinned cells. The remaining two are identical pre-existing same-assembly `List[T]` enum failures.
  - Under the “real behavior change must be pinned” rule, this blocks.
- `test/Interpreter.Tests/Issue2947ImportedNestedEnumSymbolicTests.cs:107-133` asserts rendered diagnostics only through bare `gsc`. The positive driver theory at `:27-63` covers all drivers, but does not pin diagnostic text for emitted `gsc` or `gsi`. Independent probes agree today; no committed test prevents driver-specific text regression.

### SHOULD-FIX

None beyond the blocker.

### NIT

None.

### Measured evidence

- Exact issue repro, cross-assembly with package-named `glib.dll`:
  - `36639253`: bare/emit/gsi each rc=1 with two `GS0156`; first is `'glib.Outer[object].Color' -> 'glib.Outer[T].Color'`.
  - isolated patch: bare/emit/gsi each rc=1 with only the deliberate enum→`int32` diagnostic, correctly rendering `'glib.Outer[T].Color'`; zero `Outer[object]`.
  - explicit `int32(c)` control: all three drivers accept; `gsi` returns `5`.
- Binder placement is correct: no renderer file changed. `BoundLiteralType_RetainsSymbolicEnclosingArgument` directly observes `[T]` while CLR remains `[object]`.
- Product condition mutants: symbolic nested construction **8/8 failed**; literal projection **8/8 failed**; self-type mapping **2 gsi cases failed**; restore returned **8/8 pass**.
- Required-parameter compiler census: **3 unique call sites** (`Accessor.cs:691,1941`; `MemberLookup.cs:794`).
- Baseline-green corpus: **30 sites** compile/bind on both trees in bare `gsc`, emitted `gsc`, and `gsi`; emitted value is `37` on both.
- Synthetic patch-id equals PR patch-id: `c42052b65e16d16529ae0769789193edaf9d63ca`. `merge-tree` rc=0.
- Final merged-tree build rc=0, 0 warnings/errors. Full suites: Core **7086 passed / 1 skipped**, Compiler **4124 passed**, Interpreter **641 passed**.

PR #3103, opened 2026-08-02
gsi: use integer entry-point results as exit codes

### LINKED ISSUE #3010: gsi: integer entry-point results are printed instead of used as exit codes

DavidObando:
## Summary

- run binder-selected entry points in file-mode `gsi` without changing interactive REPL declaration behavior
- return declared `int32`/`uint32` entry-point values as process exit codes instead of printing them
- discard void and ordinary trailing expression values, while preserving source paths in diagnostics

`src/Repl/Program.cs` changes only lines 57 and 64-79. Line 58 remains `engine.Evaluate(File.ReadAllText(arg), arg)`, retaining the diagnostic filename added by #3024.

## Verification

- full solution build: 0 warnings, 0 errors
- full `Interpreter.Tests`: 665 passed
- three-driver matrix: `{explicit Main, top-level}` × integer `{0, 7, -1, 256}`; emitted program and `gsi` match on stdout and OS-visible exit code
- false-positive matrix: void, trailing integer, trailing string, and ignored integer call retain stdout and exit 0 in both shapes
- POSIX results: `-1 -> 255`, `256 -> 0`; Windows assertions retain the native full-width code
- emitted probes use fresh empty output directories and pass `Assembly.Load(File.ReadAllBytes(...)).GetTypes()` before bounded execution
- diagnostic probe confirms the source path remains present
- `cs2gs-corpus`: exit 0; 1 known gap, 0 new, 0 regressed, 0 stale
- predicted merge tree exactly matched tested HEAD tree

Product mutation round trips:

- compilation return-type gate: `8 pass -> 4 fail/4 pass -> 8 pass`
- evaluator entry-point gate: `8 pass -> 4 fail/4 pass -> 8 pass`
- session entry-point gate: `2 pass -> 2 fail -> 2 pass`
- program error branch: `1 pass -> 1 fail -> 1 pass`

Fixes #3010

### DIFF
--- a/src/Core/CodeAnalysis/Compilation/Compilation.cs
+++ b/src/Core/CodeAnalysis/Compilation/Compilation.cs
@@ -290,7 +290,7 @@ public class Compilation
     /// <param name="variables">The symbol table with actual values.</param>
     /// <returns>An evaluation result.</returns>
     public EvaluationResult Evaluate(Dictionary<VariableSymbol, object> variables) =>
-        Evaluate(variables, evaluateEntryPoint: true);
+        Evaluate(variables, evaluateEntryPoint: true, useEntryPointReturnType: false);
 
     /// <summary>
     /// Evaluates the current compilation provided a symbol table with actual values.
@@ -298,7 +298,20 @@ public class Compilation
     /// <param name="variables">The symbol table with actual values.</param>
     /// <param name="evaluateEntryPoint">Whether to invoke an explicit program entry point.</param>
     /// <returns>An evaluation result.</returns>
-    public EvaluationResult Evaluate(Dictionary<VariableSymbol, object> variables, bool evaluateEntryPoint)
+    public EvaluationResult Evaluate(Dictionary<VariableSymbol, object> variables, bool evaluateEntryPoint) =>
+        Evaluate(variables, evaluateEntryPoint, useEntryPointReturnType: evaluateEntryPoint);
+
+    /// <summary>
+    /// Evaluates the current compilation provided a symbol table with actual values.
+    /// </summary>
+    /// <param name="variables">The symbol table with actual values.</param>
+    /// <param name="evaluateEntryPoint">Whether to invoke an explicit program entry point.</param>
+    /// <param name="useEntryPointReturnType">Whether the declared entry-point return type controls the result.</param>
+    /// <returns>An evaluation result.</returns>
+    public EvaluationResult Evaluate(
+        Dictionary<VariableSymbol, object> variables,
+        bool evaluateEntryPoint,
+        bool useEntryPointReturnType)
     {
         var parseDiagnostics = SyntaxTrees.SelectMany(st => st.Diagnostics);
         var diagnostics = parseDiagnostics.Concat(GlobalScope.Diagnostics).ToImmutableArray();
@@ -342,7 +355,11 @@ public class Compilation
             program,
             (References ?? Symbols.ReferenceResolver.Default()).MapClrTypeToReferences);
 
-        var evaluator = new Evaluator(program, variables, evaluateEntryPoint);
+        var evaluator = new Evaluator(
+            program,
+            variables,
+            evaluateEntryPoint,
+            useEntryPointReturnType);
         try
         {
             var value = evaluator.Evaluate();
--- a/src/Core/CodeAnalysis/Compilation/Compilation.cs
+++ b/src/Core/CodeAnalysis/Compilation/Compilation.cs
@@ -289,7 +289,16 @@ public class Compilation
     /// </summary>
     /// <param name="variables">The symbol table with actual values.</param>
     /// <returns>An evaluation result.</returns>
-    public EvaluationResult Evaluate(Dictionary<VariableSymbol, object> variables)
+    public EvaluationResult Evaluate(Dictionary<VariableSymbol, object> variables) =>
+        Evaluate(variables, evaluateEntryPoint: true);
+
+    /// <summary>
+    /// Evaluates the current compilation provided a symbol table with actual values.
+    /// </summary>
+    /// <param name="variables">The symbol table with actual values.</param>
+    /// <param name="evaluateEntryPoint">Whether to invoke an explicit program entry point.</param>
+    /// <returns>An evaluation result.</returns>
+    public EvaluationResult Evaluate(Dictionary<VariableSymbol, object> variables, bool evaluateEntryPoint)
     {
         var parseDiagnostics = SyntaxTrees.SelectMany(st => st.Diagnostics);
         var diagnostics = parseDiagnostics.Concat(GlobalScope.Diagnostics).ToImmutableArray();
@@ -333,7 +342,7 @@ public class Compilation
             program,
             (References ?? Symbols.ReferenceResolver.Default()).MapClrTypeToReferences);
 
-        var evaluator = new Evaluator(program, variables);
+        var evaluator = new Evaluator(program, variables, evaluateEntryPoint);
         try
         {
             var value = evaluator.Evaluate();
--- a/src/Core/CodeAnalysis/Evaluator.cs
+++ b/src/Core/CodeAnalysis/Evaluator.cs
@@ -154,6 +154,23 @@ public sealed partial class Evaluator
     /// <returns>The evaluation result.</returns>
     public object Evaluate()
     {
+        // The binder has already resolved top-level-statement versus explicit-Main precedence.
+        // A null declaration identifies the synthesized top-level entry point.
+        if (program.EntryPoint?.Declaration != null
+            && program.Functions.TryGetValue(program.EntryPoint, out var entryPointBody))
+        {
+            var locals = new ConcurrentDictionary<VariableSymbol, object>();
+            if (program.EntryPoint.Parameters.Length > 0)
+            {
+                locals[program.EntryPoint.Parameters[0]] = Array.Empty<string>();
+            }
+
+            using (PushFrame(locals))
+            {
+                return EvaluateFunctionBody(entryPointBody);
+            }
+        }
+
         return EvaluateFunctionBody(program.Statement);
     }
 
--- a/src/Core/CodeAnalysis/Evaluator.cs
+++ b/src/Core/CodeAnalysis/Evaluator.cs
@@ -38,6 +38,7 @@ public sealed partial class Evaluator
     private readonly BoundProgram program;
     private readonly Dictionary<VariableSymbol, object> globals;
     private readonly bool evaluateEntryPoint;
+    private readonly bool useEntryPointReturnType;
 
     // Issue #1651: everything below this point used to be an ordinary
     // instance field. That is correct only because the interpreter was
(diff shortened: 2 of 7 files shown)


DavidObando:
Rubber-duck review — **NOT MERGEABLE**

`main` = `abfab99c`, head = `3e1d2a45cb8e52a7ffdaac45495151e8ee734878`, merge-base `4deeed8d`. Merge onto `abfab99c` is clean (`merge_rc=0`); `dotnet build GSharp.sln -c Release` on the merged tree: **BUILD_RC=0, 0 errors / 0 warnings**. Driver freshness proved by raw `stat` mtime on `Core/`, `Compiler/` and `Repl/` (PRE=1785693304; all three `GSharp.Core.dll` = 1785693312, `gsi.dll` = 1785693317), `ref/` excluded.

## The headline first: this PR and #3011 are the same fix submitted twice

The `src/` subtree hash is **identical on both heads** — `e3587e9d856cc044b8466238413c0cc967106079` on #3103 and on #3011, against `1cd360ec2fab88ea614712367405adc810cd15bb` on `main`. All four production blobs match exactly (`Compilation.cs 2521db42824e`, `Evaluator.cs 979c0d20f01c`, `SessionEngine.cs eacf0671519f`, `Repl/Program.cs 5f8176199e4c`). They differ only in tests and in what they declare.

They cannot both land: `git merge-tree` returns **rc=0 while emitting 24 conflict-marker lines** — an *added-in-both* conflict on `test/Interpreter.Tests/Issue3010EntryPointExitCodeTests.cs` (`a3ffa208` vs `7f9b221c`). (Another instance of rc=0 being a floor, not a proof.)

## What works, measured

The issues verbatim repro, run byte-for-byte before any variation, is **fixed and oracle-matching**: `gsi` rc=3 stdout `main`; `gsc /out:` + `dotnet` rc=3 stdout `main`.

Across an 18-probe axis `{void, int32, int16/int64/uint32, string, bool} × {no params, []string} × {explicit Main, top-level, both}`, run on **base and merged** in **all three driver columns**, **14/18 match the oracle exactly**, including the ugly values: `-1 → 255`, `256 → 0`, `263 → 7`, and `m_silent_33` (Main-only, no output) → rc 33 both columns.

**Anti-widening holds.** `t_trailing_int` (top-level `Console.WriteLine("t11")` then a bare `11`) is rc 0 with no `11` printed, matching `gsc`. The gate at `src/Core/CodeAnalysis/Evaluator.cs:203-217` keys on the *declared entry-point return type* (`Void → null`), not the last evaluated value — that is the correct axis, and it is why "any last value" did not widen.

**#3024 preservation confirmed**: `src/Repl/Program.cs:58` keeps `engine.Evaluate(File.ReadAllText(arg), arg)`, and a merged-tree `gsi` diagnostic still carries its full path (`…/toplevel_span.gs(15,5,15,6): error GS0219: …`).

**No diagnostic ID was allocated.** `GSI001` pre-exists at `src/Repl/Engine/SessionEngine.cs:232` on `abfab99c`.

The test harness is genuinely three-column: `Issue3010EntryPointExitCodeTests.cs:34-53` runs bare `gsc`, `gsc /out:`, `dotnet`, and `gsi`, and asserts `interpreted.ExitCode == compiled.ExitCode` and `interpreted.StandardOutput == compiled.StandardOutput`, with `Assert.Empty(Directory.EnumerateFileSystemEntries(...))` on fresh dirs and `Assembly.Load(...).GetTypes()`. That is the right shape.

## BLOCKER 1 — two shipped samples regress under `gsi`, and CI cannot see it

Differential over all 122 samples under `gsi`, base vs merged: **5 changed, 2 of them rc 0 → rc 1**.

| sample | base `gsi` | merged `gsi` | oracle (`gsc /out:` + `dotnet`) |
|---|---|---|---|
| `samples/RefStructGenericField.gs` | rc 0, empty | **rc 1, empty, `GS9999`** | rc 0, stdout `3` |
| `samples/SpanComprehensive.gs` | rc 0, empty | **rc 1, empty, `GS9999`** | rc 0, stdout `60/10/2` |

```
(1,1,1,136): error GS9999: Specified method is not supported.
    (System.ReadOnlySpan`1[[System.Int32, ...]])nums
```

These are correct programs — the oracle runs them to rc 0 with real output — and under `gsi` they now exit 1 in the **`GS9999` internal-error bucket** (an unhandled exception leaking, not a designed refusal). They are exactly the 2 of 122 samples that declare `func Main`; base rc 0 was a vacuous pass because `Main` was never invoked.

**The cause is pre-existing, which changes the remedy.** Reducing to the bare construct, `infunc_span.gs` (a `ReadOnlySpan[int32]` conversion inside a function called from top level) is **rc 1 with the identical `GS9999` on BASE as well as merged**. The interpreter cannot invoke through a `ByRefLike` type; this PR does not break it, it *reaches* it by invoking `Main`. So the fix is to raise interpreter coverage or gate the invocation — not to debug this PRs new code. The exact sample shape was previously unreachable: the top-level twin is rejected outright by `GS0219` (`ref struct` cannot be a top-level variable).

**CI is structurally blind to this**: the samples gate is emit-only (`test/Compiler.Tests/LanguageConformance/SampleConformanceTests.cs:30-58`), and both samples emit and run fine. Nothing in the suite will fail.

The other 3 changed samples (`ForIn.gs`, `Loop.gs`, `WhileAndLabeledLoops.gs`) are **correct improvements** — base printed a stray trailing value (`False`, `0`, `2`) that `gsc` never printed; merged drops it and now matches the oracle.

## BLOCKER 2 — undeclared scope, in both directions

#3103 declares only `Fixes #3010`, but carries #2984s entire implementation plus its two test files (`Issue2984MainEntryPointInterpreterTests.cs` +202, `Issue2984ReplEntryPointTests.cs` +66 — both blob-identical to #3011s, `b31374aa71` / `307ff629c0`). Merging #3103 would implement #2984 and close nothing.

It also changes a **second driver** that neither the title nor #3010 mentions. `src/Core/CodeAnalysis/Compilation/Compilation.cs:292` flips the 1-arg `Evaluate(variables)` default to `evaluateEntryPoint: true`, and `src/Compiler/Program.cs:311` is a production caller — bare `gsc` evaluate mode. Measured:

| probe | base bare `gsc` | merged bare `gsc` |
|---|---|---|
| `m_int32_7` | rc 0, `Success.` | rc 0, **`m11`** then `Success.` |
| `m_void` | rc 0, `Success.` | rc 0, **`m22`** then `Success.` |

This is pinned (`Issue3010EntryPointExitCodeTests.cs:36`), so it is tested — but it is undeclared, and it is #2984 behaviour, not #3010.

## SHOULD-FIX 3 — the one divergence from the stated oracle is enshrined by a test that drops the oracle

For `Main` returning anything outside `{void, int32, uint32}`, the two columns disagree:

| probe | merged `gsi` | oracle |
|---|---|---|
| `m_int64_7`, `m_int16_7`, `m_string`, `m_bool` | rc **1**, `error GSI001: Evaluation error: Entry point must have a return type of void, int32, or uint32.` | emit rc 0, then **rc 134** `Unhandled exception. System.MethodAccessException: Entry point must have a return type of void, integer, or unsigned integer.` |

`ExplicitStringMainIsRejectedBeforeExecutionAsync` (`Issue3010EntryPointExitCodeTests.cs:126-145`) pins the `gsi` side **alone** — it is the only place in the file that does not compare against `gsc`. The choice is defensible (matching a SIGABRT would be absurd), but it should be stated as a deliberate deviation rather than left as the silent gap in an otherwise oracle-equality harness.

The underlying defect is on the `gsc` side: **`gsc` emits an entry point the CLR refuses to start**. That looks like a separate issue worth filing (I am not the filer).

Note also that bare `gsc` accepts `m_int64_7` and prints `m11` at rc 0 — so one program now yields three different answers in three drivers.

## SHOULD-FIX 4 — a designed refusal wearing an internal-error shape

`src/Core/CodeAnalysis/Evaluator.cs:213` implements the return-type refusal by `throw new InvalidOperationException(...)`, which is caught by the generic catch-all and rendered as `GSI001` via `new Diagnostic(default, ...)` at `src/Repl/Engine/SessionEngine.cs:232`. Because the span is `default`, this diagnostic **carries no source location**, unlike every neighbouring `gsi` diagnostic (`GS0219` above renders its full path). A deliberate refusal should be a real diagnostic with a span, not an exception routed through the internal-error wrapper.

## Disclosures / gaps in my own evidence

- **Core.Tests was not run.** 253 of the 304 test files that call `.Evaluate(` live in `test/Core.Tests`, and `Compilation.cs:292` changed that calls default; the PR description reports only `Interpreter.Tests` (665). Load reached 34.37 and I stopped building, so this blast radius is **unverified by me** — it is a gap, not a pass.
- **No per-hunk mutation was performed**, for the same reason. The four gates are unmutated by me.
- **I caught one of my own checks being unrun**: I first measured driver freshness with `find -newermt "@$PRE" 2>/dev/null`, which returned 0 for all three directories. BSD `find` cannot parse `@epoch` — it exited 1 and my `2>/dev/null` hid it, so "0 stale" was an error, not a result. Re-measured with raw `stat`; the drivers are genuinely fresh.

## Ordering

**#3011 should land first and #3103 should not land as-is.** The production code is byte-identical, #3011 carries a superset of the tests (it adds `Issue3010EntryPointDriverMatrixTests.cs`, +241), and #3011 declares both `Fixes #2984` and `Fixes #3010` — which is what this change actually does. After #3011 merges, #3103 has nothing left to contribute except its variant of the conflicting test file, so it should be closed as superseded or reduced to whatever test cases #3011 lacks. Whichever lands, **BLOCKER 1 travels with it**, since it comes from the invoke-`Main` half that both carry.

DavidObando:
Superseded by #3011, which has now merged.

Two independent reviewers established this: `git diff --full-index <3103> <3011> -- src/` is **empty** — all four production blobs are byte-identical, with the same eight commit messages in the same order. #3011 is a strict superset (it carries an additional 241-line three-driver matrix test) and declares both `Fixes #2984` and `Fixes #3010`, which is what the change actually does.

The work here is not lost — it is what merged. Thanks; the duplication was my routing error, not this branch's.

Issue #1203, opened 2026-06-26
Bodyless func inside a 'shared { }' block crashes compiler (GS9998 NullReferenceException) — blocks static DllImport externs


DavidObando:
### Summary
A **bodyless** (abstract/extern) function declaration placed inside a `shared { … }` static block crashes the compiler with `GS9998: NullReferenceException`. This is exactly the pattern cs2gs emits for static `[DllImport]` P/Invoke externs.

### Repro (minimal — no unsafe, no DllImport, no pointers needed)
```gsharp
package p
class C {
    shared {
        func F(x int32) bool;   // GS9998: NullReferenceException
    }
}
```
Also crashes with `unsafe func F(x int32) bool;` or `@DllImport("k") func F(x int32) bool;` inside `shared { }`.

By contrast, a bodyless func at **instance** level in an `open class` works fine:
```gsharp
package p
import System.Runtime.InteropServices
open class C {
    @DllImport("k") open func F(x int32) bool;   // compiles ("Wrote …")
}
```

### Expected
A bodyless function declaration inside a `shared { }` block (the natural home for static `[DllImport]` externs) should bind/emit without crashing — either compiling as an extern/abstract static, or producing a proper diagnostic, never an unhandled `NullReferenceException`.

### Impact (Oahu.Foundation)
`Auxiliary/Win32/Win32FileIO.cs` declares its Win32 P/Invoke methods (`CreateFile`, `ReadFile`, `WriteFile`, `CloseHandle`) as static `[DllImport]` externs; cs2gs emits them as bodyless `func … ;` inside `shared { }`. This crash blocks compilation of the whole file. Follow-up to #1014.

### Environment
gsc at current main (9d3056f4). `bug`.

PR #3104, opened 2026-08-02
Refine P/Invoke diagnostics to interactive interpreter use sites

DavidObando:
## Summary

This is a refinement of the existing #2986 boundary, not the original fix. `main` already rejected every P/Invoke declaration before evaluation. This PR moves `GS0514` to deprecated evaluator-engine use sites so unused declarations remain valid without weakening the native-call boundary.

- report `GS0514` at direct calls and first-class function references
- preserve use-site locations through binding and rewriting
- keep compiler diagnostic control signals outside G# `catch`
- synchronously preflight transitive user-function and stored-closure call graphs before free-standing `go`
- conservatively refuse free-standing `go` when P/Invoke is declared and the scan cannot name a precise native call
- cover all function-bearing bound-node shapes relevant to reachable-body analysis
- preserve `GS0514` through static initialization instead of degrading to `GS9999`
- document the invariant and ADR-0156's emitted default-driver behavior

## Boundary invariant

Any evaluator path that reaches a P/Invoke call or function reference reports `GS0514` before native dispatch. User `catch`, static initialization, scoped `go`, and free-standing `go` cannot translate or silently discard it.

Free-standing goroutines cannot return asynchronous diagnostics to a completed REPL cell. The evaluator therefore follows statically resolvable call bodies before scheduling. If no precise native use is found but the compilation declares P/Invoke, it refuses the launch conservatively. P/Invoke declarations are top-level-only, so `BoundProgram.Functions` is a complete fallback set. This deliberately over-refuses that narrow combination rather than converting a loud boundary into silent wrong output.

## Compiler-wide syntax propagation

The location fix is intentionally global, not evaluator-local. `BoundTreeRewriter` now retains `node.Syntax` for every rewritten `BoundCallExpression`, and `ExpressionBinder` retains syntax for method-group call nodes. Those paths are shared by compiler and emitter flows as well as evaluator diagnostics. The full build, emitted-path tests, and required `ilverify` check cover that wider blast radius.

## ADR-0156 driver behavior

Since ADR-0156 Phase 3a, bare `gsc`, `gsc /out:`, `gsi <file>`, and the interactive REPL execute emitted code and genuinely execute P/Invoke by default. `GS0514` is reachable only through gsi's deprecated `--engine evaluator` escape hatch and retires with that evaluator in Phase 3c.

## Round 3 regression matrix

Tests were first applied alone to the unfixed tree. B, E, and interface-dispatch `f5` completed with rc 0, no diagnostic, and empty output; their scope controls reported `GS0514`. After the product fix:

| shape | unfixed evaluator | fixed evaluator |
|---|---|---|
| B: instance method + free `go` | silent rc 0; empty output | `GS0514`; empty output |
| E: constructor `init` + free `go` | silent rc 0; empty output | `GS0514`; empty output |
| H: instance method + scoped `go` | `GS0514`; empty output | `GS0514`; empty output |
| I: constructor `init` + scoped `go` | `GS0514`; empty output | `GS0514`; empty output |
| f5: interface dispatch + free `go` | silent rc 0; empty output | `GS0514`; empty output |

No-P/Invoke controls remain clean. A free-standing `go` unrelated to a declared P/Invoke is conservatively refused, matching the old declaration-wide boundary; declaration-only execution is the sole net loosening.

## Verification

Post-rebase local verification at `main` `0d60aa1f1e9b` + head `fd2d64c95608`:

- solution build: **BUILD_RC=0**, 0 warnings, 0 errors
- in-process 20-probe `SessionEngine` matrix: **20/20 matched**
- Interpreter.Tests: **1,509 passed, 0 failed, 4 skipped**

Surviving product-hunk mutants were asserted in source and `GSharp.Core.dll`, killed, and restored:

- conservative fallback removal: **1 killed** by `PInvokeDeclarationWithFreeGo_IsConservativelyRejected`; `f5` also escapes under it
- reachable-body traversal condition inversion: **3 killed** by transitive wrapper plus B/E pins
- diagnostic-control catch bypass removal: **4 killed** by the three P/Invoke catch shapes plus the other evaluator-boundary control

The unreachable forged instance-P/Invoke dispatch guard and its synthetic test were removed after review established that G# P/Invoke declarations are top-level-only.

Required CI passed on the rebased head, including `ilverify`.

Follow-up refinement for #2986.


DavidObando:
## Rubber-duck review — PR #3104 `fix/2986-gsi-boundaries`

**Verdict: NOT MERGEABLE** — one BLOCKER (four P/Invoke use shapes reach the interpreter's native boundary with rc0 and no `GS0514`, all four diagnosed on base).

True of: `main` **`abfab99c`**, PR head **`0b6ca887`**, computed merge base **`4deeed8d`**, merged head **`cf00e1c2`** (`BUILD_RC=0`, 0 warnings), co-merged with #3071 **`00a56c0c`** (`BUILD_RC=0`, 0 warnings).

Since this PR is a **relaxation**, I tested for false negatives, not false positives. Positive tests cannot see that class.

---

### BLOCKER — a deliberate boundary diagnostic is thrown as a catchable exception, so user code and goroutines swallow `GS0514`

`RejectPInvoke` (`src/Core/CodeAnalysis/Evaluator.cs:160-169`) signals the boundary by throwing an ordinary `EvaluatorException`. Two existing interpreter mechanisms treat that as a normal runtime exception:

- `src/Core/CodeAnalysis/Evaluator.Statements.cs:791-814` — `TryFindCatchHandler` matches on `UnwrapRuntimeException(ex).GetType()` and walks `BaseType`. `RejectPInvoke`'s exception has **no** `InnerException`, so `UnwrapRuntimeException` returns the `EvaluatorException` itself, whose chain is `EvaluatorException → System.Exception`. A user `catch (e Exception)` therefore **matches and swallows it**.
- `src/Core/CodeAnalysis/Evaluator.Statements.cs:381-386` — the fire-and-forget goroutine path has a bare `catch { }` that discards per ADR-0022.

Measured under `gsi`, each probe compiled into its own asserted-empty directory:

| shape | base `abfab99c` | merged `cf00e1c2` | co-merged `00a56c0c` |
|---|---|---|---|
| unused declaration | rc1 `GS0514` | **rc0** ✅ *intended relaxation* | rc0 |
| direct call | rc1 `GS0514` | rc1 `GS0514` | rc1 `GS0514` |
| call inside invoked function | rc1 `GS0514` | rc1 `GS0514` | — |
| first-class reference / stored local | rc1 `GS0514` | rc1 `GS0514` | — |
| lambda (no `try`) | rc1 `GS0514` | rc1 `GS0514` | rc1 `GS0514` |
| static `shared { @DllImport … }` member | rc1 `GS0514` | rc1 `GS0514` | rc1 `GS0514` |
| **call inside `try`/`catch (e Exception)`** | rc1 `GS0514` | **rc0, no diagnostic** ❌ | **rc0** ❌ |
| **method group inside `try`/`catch`** | rc1 `GS0514` | **rc0, no diagnostic** ❌ | **rc0** ❌ |
| **lambda call inside `try`/`catch`** | rc1 `GS0514` | **rc0, no diagnostic** ❌ | **rc0** ❌ |
| **`go NativeStrLen(...)`** | rc1 `GS0514` | **rc0, no diagnostic** ❌ | **rc0** ❌ |

Why this is the #2986 defect class rather than a cosmetic gap — same source, two drivers:

```
package p
import System
import System.Runtime.InteropServices

@DllImport("libc", EntryPoint: "strlen", CharSet: CharSet.Ansi)
func NativeStrLen(text string) nint;

var caught = "no"
try {
    Console.WriteLine(NativeStrLen("Hello, world!"))
} catch (e Exception) {
    caught = "caught"
}
Console.WriteLine(caught)
Console.WriteLine(88)
```

- `gsc /out:` + run → `13` / `no` / `88`, rc0 (native call succeeds, catch not taken)
- `gsi` → `caught` / `88`, **rc0, no diagnostic**

The interpreter silently takes a control-flow path the emitted program does not, with no diagnostic — which is precisely "a call the interpreter cannot perform must not return a plausible-looking wrong value" from the issue. On base this program is rc1 `GS0514`, because the declaration scan ran before evaluation and could not be caught by user code. The relaxation is correct in intent; routing the signal through the user-visible exception channel is what reopens the hole.

`Interpreter.Tests` is **653/0/653 green** on the co-merged tree and does not detect any of this — none of the new or existing tests combine a P/Invoke use with `try`/`catch` or `go`.

**Candidate fix, measured.** A 6-line guard in `TryFindCatchHandler`, using the `Location` property this PR already adds as the "deliberate diagnostic" discriminator (`DiagnosticId` defaults to `"GS9999"`, so it cannot discriminate):

```csharp
if (ex is EvaluatorException { Location: not null })
{
    matched = null;
    return false;
}
```

Applied to `cf00e1c2`, rebuilt, asserted applied (pattern count 1), then reverted with a verified md5 round-trip:

- `try`/`catch` call, method group, and lambda shapes → **rc1 `GS0514`** (three of four escapes closed)
- unused declaration → **still rc0** (relaxation preserved)
- `samples/Exceptions.gs` → matches golden (ordinary user `catch` unaffected)
- `Interpreter.Tests` → **643/0/643 green**

The `go` shape is **not** fixed by that guard and needs its own decision, since ADR-0022 mandates discarding at `Evaluator.Statements.cs:381-386`. Reporting a deliberate boundary diagnostic there conflicts with fire-and-forget; deciding it explicitly (even "documented as unreachable-by-design") is what I would want before merge.

---

### SHOULD-FIX — `GS0514` points at the declaration, not the use site

`Evaluator.cs:163` passes `function.Declaration.Identifier.Location`. The PR's thesis is use-site diagnosis, but the rendered location is the declaration:

```
samples/PInvoke.gs(20,6,20,18): error GS0514: ...
    func NativeStrLen(text string) nint;
```

Line 20 is the `func NativeStrLen` declaration; the offending call is line 22. With more than one call site the diagnostic cannot distinguish them, and `PInvokeDeclaredInEarlierCell_IsDiagnosedOnlyWhenUsed` deliberately pins `diagnostic.Location.FileName == "declaration.gs"` — so in a REPL the error names an **earlier, blameless cell** rather than the cell the user just typed. The `node` is already passed to `RejectPInvoke`; using the use-site location (or naming both) matches what the PR claims to do.

### SHOULD-FIX — `Closes #2986` while one row of the issue's own table is still silently wrong

#2986 is **already CLOSED**, so the keyword is a no-op, but the residual is worth recording: `samples/PInvokeStructMarshalling.gs` under `gsi` still prints `3 / 4 / 0` against a golden of `3 / 4 / 6153737367135073092` — rc0, no diagnostic. That file declares **no** P/Invoke function (it is unmodelled `@StructLayout(Explicit)` field overlap), so neither the old scan nor the new use-site check applies and this PR neither causes nor worsens it. Flagging only because it is a live silent-wrong-answer row inside the table this PR's issue is closed on, and it appears to be unfiled.

### NIT — a null guard was dropped alongside `SyntaxTrees.Contains`

The deleted scan tested `f.Declaration != null`; `RejectPInvoke` (`Evaluator.cs:163`) dereferences `function.Declaration.Identifier` unguarded. I could not reach it: `IsPInvoke => PInvokeMetadata != null` (`FunctionSymbol.cs:450`) and the sole producer is `PInvokeBinder.cs:270`, reached only from `DeclarationBinder.Functions.cs:377` and `DeclarationBinder.Structs.cs:2320`, both on a source declaration. So this is an implicit invariant, not a live crash — worth a comment rather than a guard.

The `SyntaxTrees.Contains` deletion itself I could not fault: the old scan only reported declarations in the current compilation's trees, so a declaration in an earlier REPL cell used later was *missed* on base. The new check is strictly stronger there, and `PInvokeDeclaredInEarlierCell_IsDiagnosedOnlyWhenUsed` pins it.

### NIT — the 13,884 total is not comparable to the 14,346 reported elsewhere

Only **4** commits and **+10** `[Fact]`/`[Theory]` methods separate the PR's base `4deeed8d` from `abfab99c`, so the 462-test gap is a **population/scope difference, not new tests**. Neither figure should be quoted as "the" solution total without naming which projects were counted.

---

### What I verified and found sound

- Merged tree `cf00e1c2` and co-merged-with-#3071 tree `00a56c0c` both **`BUILD_RC=0`, 0 warnings**. The #3104 corpus is **identical** on both, and `Interpreter.Tests` runs **653/0/653** with both PRs' tests together — #3104 × #3071 is disjoint **behaviourally**, not just per `merge-tree`.
- **No fabricated line-1 locations.** Every `GS0514` I produced carried a real filename, real line/column and a real source excerpt — the `Compilation.cs:344-347` `WriteTo` fallback is correctly bypassed via `ex.Location`.
- **The relaxation is genuine and correctly scoped**: `samples/PInvokeFunctionPointer.gs` declares two P/Invokes and calls neither; it is now rc0 under `gsi` and bare `gsc`, and its golden output is unchanged.
- **Static `shared { @DllImport … }` members are covered** (I probed this specifically because it bypasses the top-level function path).
- The four pre-existing interpreter test files were **not gutted** — the flipped assertions still assert positively (`Contains("ran")`), and Issue758/760 changes are doc-comment only.
- **Docs rows are byte-identical** between `docs/diagnostics.md:1162` and `website/docs/ref/diagnostics.md:1768`, so the presence-only gate is not hiding a divergence here.
- **No `Fixes/Closes #2988`** keyword in the body or any commit message; the scope statement about #2988 holds.

### Disclosure

The base `abfab99c` build started at load **59.44**, above the stated bar. I report it rather than hide it: it returned `BUILD_RC=0` in 34s, all 22 non-`ref` `GSharp.Core.dll` copies plus `gsc` and `gsi` carried fresh mtimes, and the base results it produced are the *conservative* direction for my blocker (base diagnosing more, not less). The merged, co-merged and fix-probe builds were all started below 30.

`gsi` could not be driven from a pipe ("gsi requires an interactive terminal"), so the chained-cell claim was verified through `SessionEngine` in the test source rather than the REPL binary.

DavidObando:
## Rubber-duck review — PR #3104 `fix/2986-gsi-boundaries` @ `b1300c27`

**Verdict: NOT MERGEABLE.** One BLOCKER (a measured regression against `main`), two SHOULD-FIX, one NIT.

Reviewed against `main` = **`9ab9124d`** (the brief said `6267cd44`; GitHub API disagreed, so I measured). Merge-base `87af33c9`, `main` 1 commit ahead. `git merge-tree` **output**: 41 bytes, 0 conflict markers. CI green on all 30 checks **including the new required `ilverify` gate** (present and passing — no `update-branch` needed).

Two isolated trees, both `BUILD_RC=0`, 0 warnings: `base` @ `9ab9124d`, `head` @ `b1300c27`. Every probe in its own asserted-empty directory. Three drivers throughout: bare `gsc` **evaluates**, `gsc /out:` **EMITS (oracle)**, `gsi` **interprets**.

---

### First, a reframing that changes what this PR is

**#2986's original defect does not reproduce on `main`.** `GS0514` and `ReportPInvokeNotSupportedInInterpreter` exist on **both sides** of the diff; the pre-run declaration scan at `Compilation.cs:358-372` already refuses every P/Invoke program. This PR **deletes** that scan and replaces it with three hand-placed use-site guards.

So this is not the #2986 fix — it is a *refinement* trading over-strictness for use-site precision. The thing it genuinely buys is real and I measured it:

| probe | driver | base `9ab9124d` | head `b1300c27` |
|---|---|---|---|
| `p2_unused` — declare, never call | bare | `GS0514` rc=1 | **`P2UNUSED-2222` rc=0** ✅ |
| | **EMIT** | `P2UNUSED-2222` | `P2UNUSED-2222` |
| | gsi | `GS0514` rc=1 | **`P2UNUSED-2222` rc=0** ✅ |

That cell now agrees with the oracle. Direct call (`p1`) and function value (`p3`) both report `GS0514` with **use-site locations** (`(7,19,7,48)` and `(7,9,7,21)` respectively, vs base's fixed declaration location `(5,6,5,18)`).

But the old scan was sound **by construction** — any declaration stopped the run. The replacement is sound **by enumeration**: three guards at `Evaluator.Calls.cs:45`, `Evaluator.Expressions.cs:813`, `Evaluator.Statements.cs:374`. Two routes escape the enumeration.

---

## BLOCKER — free-standing `go` reaching P/Invoke *transitively* is silent, exit 0. Regression vs `main`.

`src/Core/CodeAnalysis/Evaluator.Statements.cs:374-377` (guard) + `:408` (bare `catch`)

The `go` guard only fires when the goroutine expression *is itself* the P/Invoke call:
```csharp
if (expression is BoundCallExpression call) { RejectPInvoke(call.Function, call); }
```
`go Wrapper()` where `Wrapper` calls native passes the guard, then throws `GS0514` on the task thread, where the bare `catch` at `:408` discards it.

`p5_go_indirect.gs` measured:

| driver | base `9ab9124d` | head `b1300c27` |
|---|---|---|
| bare `gsc` | `GS0514` rc=**1** | `P5START~P5INSIDE~P5END-5555` rc=**0** "Success." |
| **EMIT (oracle)** | `P5START~P5INSIDE~13~P5AFTERNATIVE-9999~P5END-5555` rc=0 | *(identical)* |
| `gsi` | `GS0514` rc=**1** | `P5START~P5INSIDE~P5END-5555` rc=**0** |

The goroutine body **silently truncates at the native call** — `13` and `P5AFTERNATIVE-9999` never print, no diagnostic is emitted, and the process exits **0**. That is #2986's own failure class (silent divergence from emit, exit 0) reintroduced on a route the deleted declaration scan covered.

**Mechanism isolated by controlled comparison, not by reading:** `p5b_scope_indirect.gs` is the same program wrapped in `scope { }` — the enclosing-scope branch of `EvaluateGoStatement` has no bare `catch` and awaits the task — and it correctly reports `GS0514` at `(8,23,8,52)` on both evaluator drivers. The single structural difference between `p5` and `p5b` is the `catch` at `:408`.

Per the standing rule about not scoring an empty `catch` without establishing whose path it is on: I checked the others and did **not** score them. `Evaluator.Expressions.cs:112` is `when (ex is not EvaluatorException)` — explicitly excludes this class. `Evaluator.Statements.cs:473`/`:482` are scope cancellation and rethrow. `:408` is the only one on the user-facing path, and the `p5`/`p5b` pair proves it.

This directly falsifies two of the PR's own claims:
- commit 2 is titled *"Prevent interpreter boundary diagnostics from being swallowed"* — the `catch` swallow was closed, the goroutine swallow was not;
- `docs/adr/0152-interpreter-native-call-boundary.md` asserts *"a direct `go` P/Invoke call is rejected synchronously before the fire-and-forget task is scheduled."* True, and precisely why the indirect shape was missed — **the ADR documents the enumeration rather than the invariant.**

**No test composes `go` with an indirect call.** The suite mirrors the enumeration exactly: `PInvokeGo_ReportsBeforeFireAndForgetSpawn` covers `go NativeStrLen(...)` and `PInvokeCallInsideInvokedFunction_ReportsGS0514` covers indirect-but-synchronous. The product of the two is the hole.

**Recommended fix:** let the control signal escape `:408` — it is a compiler diagnostic, not a runtime exception, which is the identical argument commit 2 already makes for `catch` clauses via `IsDiagnosticControlSignal`. `TryFindCatchHandler` gained that check; the fire-and-forget `catch` did not. Re-throwing (or recording to a diagnostic sink) when `UnwrapDiagnosticException(ex) is EvaluatorException { IsDiagnosticControlSignal: true }` makes both swallow paths consistent and closes the shape *by construction* rather than adding a fourth enumerated site.

---

## SHOULD-FIX 1 — P/Invoke in a `shared { init { } }` degrades to `GS9999`

`src/Core/CodeAnalysis/Evaluator.cs:370` / `Compilation.cs:375` / `Evaluator.Statements.cs:68-79`

`p7_staticinit.gs`:

| driver | base `9ab9124d` | head `b1300c27` |
|---|---|---|
| bare `gsc` | `(5,6,5,18): error GS0514: P/Invoke function 'NativeStrLen' …` | `(1,1,1,13): error GS9999: The type initializer for 'Holder' threw an exception.` |
| **EMIT** | `P7START~13~P7END-8888` | *(identical)* |
| `gsi` | `GS0514` | `GS9999` |

The static-initializer `catch (Exception ex)` at `Evaluator.cs:370` re-throws a `TypeInitializationException`, which is not an `EvaluatorException`, so `Compilation.cs:375` never sees the diagnostic; `UnwrapDiagnosticException` unwraps only `AggregateException`+`TargetInvocationException`. Result: no ID, **no filename**, a synthesized `(1,1,1,13)` location, and a message that never says "P/Invoke".

ADR-0152 in this very PR states: *"`GS9999` remains an unexpected evaluator-exception diagnostic, not a deliberate capability boundary."* This route violates it, and base did not.

---

## SHOULD-FIX 2 — both published diagnostic references now contradict themselves

`docs/diagnostics.md:1163-1165` and `website/docs/ref/diagnostics.md:1772-1774`

The table row was updated to *"unused P/Invoke declarations remain valid"*; the paragraph immediately below it was not, and still reads:

> This is an intentional interpreter capability boundary… **It applies even when the declaration is not called** because `gsi` cannot create a valid callable value for direct or indirect use.

That is now false, contradicts the row two lines above, and contradicts the ADR change in the same commit. Both files, identical stale prose.

---

## NIT — the third commit's title inverts what it does

`fdcedd18 "Allow unused native declarations in sample gate"` reads as a gate relaxation. **It is a gate tightening**, and the brief's hypothesis on this point is refuted by the harness's own semantics.

`ExpectedDifferences` (`SampleConformanceTests.cs:39-83`) is an **exemption** table. A listed sample is asserted against a declared divergence; an unlisted sample must match the emitted golden on all three drivers (`:132-149`). It carries a redundancy guard:

```csharp
Assert.False(
    Equivalent(emitted, expectedCompiler) && Equivalent(emitted, expectedInterpreter),
    $"{sampleName} has a redundant expected-difference entry (…)");
```

`samples/PInvokeFunctionPointer.gs` declares `native_qsort`/`native_dlsym` and **never calls them** — it only exercises managed `compareInt64`. After the product change all three drivers agree, so the entry became redundant and **leaving it in would have failed the suite**. Removing it moves the sample onto the strict all-drivers-must-match path. The gate got stronger, not weaker; the title describes the product permission. Worth renaming so a future reader does not read it as a relaxation.

---

## What I verified as solid

**Assertion granularity — not the coarse-observable trap.** `PInvokeDirectCall_ReportsLocatedGS0514` asserts the ID, the severity, the **filename**, the **start line**, two message substrings, and exact empty output. `PInvokeFunctionValue_…` additionally asserts `DoesNotContain("IConvertible")`, pinning the specific bad message from the issue triage. This is the standard the fleet has been asking for.

**Mutation — 6 PRODUCT mutants, per hunk, all applied and all killed.** Applied-ness asserted by source md5 **and** `GSharp.Core.dll` md5 movement; each restored by exact md5 round-trip. Baseline `Interpreter.Tests` 1279/0.

| mutant | hunk | Core.dll md5 | result | dying test(s) |
|---|---|---|---|---|
| **M1-CALLS** | `Evaluator.Calls.cs:45` | `f2a4367b` | **RED 14** | `PInvokeDirectCall_ReportsLocatedGS0514`, `PInvokeCallInsideInvokedFunction_…`, `…RewrittenArgument…`, LINQ-lambda + generic-wrapper shapes |
| **M2-GROUP** | `Evaluator.Expressions.cs:813` | `64ba9539` | **RED 4** | `PInvokeFunctionValue_…`, `PInvokeDeclaredInEarlierCell_…`, "stored local", "method group" |
| **M3-GO** | `Evaluator.Statements.cs:376` | `13aaa541` | **RED 1** | `PInvokeGo_ReportsBeforeFireAndForgetSpawn` |
| **M4-SIGNAL** | `Evaluator.Statements.cs:820` (condition inversion `true`→`false`) | `6ba806a8` | **RED 4** | all 3 `PInvokeBoundaryDiagnostic_CannotBeCaught` shapes + `OtherEvaluatorBoundaryDiagnostic_CannotBeCaught` |
| **M5-SYNTAX** | `BoundTreeRewriter.cs:623` (`node.Syntax`→`null`) | `7b0ef993` | **RED 1** | `PInvokeCallWithRewrittenArgument_ReportsUseLocation` |
| **MCTRL** (control) | `Evaluator.cs` `IsPInvoke`→`!IsPInvoke` | `2dd42b4c` | **RED 478** | broad — suite is live |

M1/M2/M3/M5 have **disjoint kill sets**, so each guard is independently load-bearing. M4 killing `OtherEvaluatorBoundaryDiagnostic_CannotBeCaught` shows the control-signal fix correctly generalizes past P/Invoke to the other evaluator boundaries (GS0511/GS0513). M5 confirms the `BoundTreeRewriter` syntax-threading hunk — the widest-blast-radius change here — **is** pinned by a purpose-built test rather than riding along untested.

One mutant produced a **no-result and was retargeted, not scored**: replacing the `Calls.cs` guard with a comment tripped `SA1512` and failed the build. Re-expressed as passing a non-P/Invoke symbol.

Restore verified: `git status --porcelain` empty, rebuild returns `GSharp.Core.dll` to pristine `7ab95195d836c5a9c8993deffb470c74`.

**`defer` ruled out** — no `BoundDeferStatement` in the evaluator, so that route does not exist.

**Suites — denominator 9**, from `dotnet sln GSharp.sln list` filtered to `*.tests.csproj` (the identical call CI makes per `build/generate-ci-test-matrix.py`). **8 of 9 green, 10,976 tests, 0 failures, 1 skipped:**

|---|---|---|---|---|
| Core.Tests | 0 | 0 | 7127 | 1 |
| Interpreter.Tests | 0 | 0 | 1279 | 0 |
| InternalAnalyzers.Tests | 0 | 0 | 7 | 0 |
| **Compiler.Tests** | — | — | — | **NOT FINISHED (~77 min)** |

`Compiler.Tests` was still running when I posted — that is a **gap in evidence, not a pass**, and I am labelling it as such rather than implying coverage. It does not affect the verdict: the BLOCKER is a coverage hole inside an otherwise-green suite, not a failing test, and CI's own sharded `compiler-*` jobs are green on this head.

---

## Disclosure

My first probe run returned rc=1 on **all three drivers** for four probes — which looks exactly like a result. It was my own fixture: `go` needs `import Gsharp.Extensions.Go` (`GS0316`) and `catch` needs parentheses (`GS0005`). Re-measured after fixing, and only then did the real signal appear. **A red that comes from your fixture is worth less than no data, because it looks like data.** Also: the `main` SHA in my brief (`6267cd44`) was stale — measured `9ab9124d`.

DavidObando:
### Addendum — coverage gap closed: **9 of 9, 15,231 tests, 0 failures, 1 skipped**

`Compiler.Tests` was still running when I posted and I labelled it a gap in evidence rather than implying coverage. It has now finished:

|---|---|---|---|---|
| **Compiler.Tests** | **0** | **0** | **4255** | **0** |
| Core.Tests | 0 | 0 | 7127 | 1 |
| Interpreter.Tests | 0 | 0 | 1279 | 0 |
| InternalAnalyzers.Tests | 0 | 0 | 7 | 0 |

**Total 15,231 / 0 failures / 1 skipped**, all 9 projects, unfiltered. `Compiler.Tests` took 1 h 18 m under heavy contention (three other agents were running `dotnet test` on this box concurrently).

**This does not change the verdict, and it is worth being explicit about why.** A fully green 15,231-test run is exactly what a `main`-relative coverage hole looks like from the inside: the BLOCKER is a shape nobody wrote a test for (`go` composed with a transitive P/Invoke call), so no suite can fail on it. The measurement that matters remains the base-vs-head driver table for `p5_go_indirect` — `GS0514` rc=1 on `9ab9124d`, silent rc=0 on `b1300c27`, against an emit oracle that prints `13` and `P5AFTERNATIVE-9999`.

Verdict unchanged: **NOT MERGEABLE** — 1 BLOCKER, 2 SHOULD-FIX, 1 NIT.

DavidObando:
## Rubber-duck review — PR #3104 round 2 @ `cf5b4886`

**Verdict: NOT MERGEABLE.** One BLOCKER — **the round-1 BLOCKER's mechanism survives; only the four shapes I named were closed.** Both SHOULD-FIX items are properly addressed, and the round-1 NIT is resolved. No mechanical problems: `mergeable=MERGEABLE`/`CLEAN`, `merge-tree` output empty beyond the tree oid, all 34 checks pass including **`ilverify` present and passing (3m39s)**.

**Ancestry, not existence:** round-1's head `b1300c27` **exists but is NOT an ancestor** of `cf5b4886` — this branch was force-pushed/rebased. The round-1 commit `fdcedd18 "Allow unused native declarations in sample gate"` is **gone entirely** (`git diff gh/main pr3104` touches no `samples/` or `SampleConformanceTests.cs`). Surface vs live `main` `16f97229`: **+592 / −98 across 22 files**.

---

## First: the boundary this PR guards has moved, and the author is right about it

ADR-0156 landed since round 1. I re-established the driver model empirically rather than assuming it:

| driver | path | engine |
|---|---|---|
| `gsc file.gs` | `src/Compiler/Program.cs:321` | `EmittedProgramHost.Run` → **emits** |
| `gsc /out:` | same | **emits** |
| **`gsi file.gs`** | `src/Repl/Program.cs:150` | `EmittedProgramHost.Run` → **emits** |
| **`gsi` interactive (no file)** | `src/Repl/Program.cs:91` → `SessionEngine.cs:220` → `compilation.Evaluate` | **interprets** — and `--engine evaluator` is the **default** (`Repl/Program.cs:37`) |

So **the tree-walking Evaluator is now reachable only from the interactive REPL.** The renamed title ("…to *interactive* interpreter use sites") is accurate, the docs say so explicitly, and `BatchFileRunner_ExecutesPInvokeNatively` pins it. This also explains why the sample-gate commit was correctly *dropped*: in file mode the P/Invoke sample now executes natively, so nothing is "unused" and no exemption is needed. **Round-1 NIT resolved by deletion — the misleading title never enters history.**

This narrows the blast radius of everything below to the interactive REPL. It does not eliminate it: `gsi`'s default engine is the one that interprets.

---

## BLOCKER — free-standing `go` still reaches P/Invoke silently at rc=0. Same class as round 1, one shape over.

The design deletes the by-construction pre-run scan (`Compilation.cs:368-388`, `-15` lines) and replaces it with use-site guards. For fire-and-forget `go` the guard **must** be the static pre-scan at `Evaluator.Statements.cs:375-378`, because the goroutine's exception is swallowed by the **still-bare `catch` at `Evaluator.Statements.cs:408`**, which was not narrowed. That makes the `PInvokeReachabilityWalker`'s completeness safety-critical — and it is incomplete.

Measured via `SessionEngine.Evaluate`, the REPL's own entry point:

| shape | `main` `16f97229` | PR `cf5b4886` | |
|---|---|---|---|
| A top-level wrapper + `go` | GS0514@4 | **GS0514@6** | ✔ improved — use site |
| **B instance method + `go`** | **GS0514@4** | **no diagnostic, prints `body-33`** | ✘ **regression** |
| C same instance method, **no `go`** | GS0514@4 | **GS0514@7** | ✔ |
| **E ctor `init` + `go`** | **GS0514@4** | **no diagnostic, prints `body-33`** | ✘ **regression** |
| H instance method, `scope { go }` | GS0514@4 | **GS0514@7** | ✔ |
| I ctor `init`, `scope { go }` | GS0514@4 | **GS0514@8** | ✔ |
| D declaration-only, unused | GS0514@4 | allowed, prints | ✔ **the intended #2986 fix** |

B and E are byte-identical programs to C/H/I except for `go` / `scope`. **`main` refuses them loudly; this PR silently produces the wrong answer at exit 0.** That is precisely the principle round 1 turned on.

**The mechanism is nailed by the `scope { }` discriminator.** H and I are B and E wrapped in `scope { }`, and they report GS0514 — so the diagnostic *is* raised at runtime on the goroutine thread by the guard at `Evaluator.Calls.cs:44`. The scope path (`:381-398`) has no bare catch and propagates; the fire-and-forget path (`:400-418`) eats it at `:408`. The static walker never saw the shape, and there is no backstop.

### Why the walker misses them

`Find(...)` is called for exactly two node types — `BoundCallExpression.Function` (`:897`) and `BoundMethodGroupExpression.Function` (`:908`) — and bodies are followed only from `VisitCallExpression` via `VisitFunction` (`:948`), from `BoundFunctionLiteralExpression` (`:881`), and from a runtime-resolved `BoundIndirectCallExpression` (`:916`). Of the node types that carry a `FunctionSymbol`, these are **unchecked and un-followed**:

- `BoundUserInstanceCallExpression.Method` ← **proven escape (B)**
- `BoundBaseClassCallExpression.Method`
- `BoundBaseInterfaceCallExpression.Method`
- `BoundConstrainedStaticCallExpression.InterfaceMethod`
- `BoundFunctionPointerFromMethodExpression.Method`
- `BoundClrBinaryOperatorExpression.Function`

plus constructor/`init` bodies ← **proven escape (E)**. Note `EvaluateUserInstanceCallExpression` (`Evaluator.Calls.cs:229`) also has no `RejectPInvoke`; it is saved only because the native call inside the method body is itself a `BoundCallExpression`.

### I built the obvious remedy, and it is not enough — which is the actual finding

In a separate worktree I added 9 lines to the walker for `BoundUserInstanceCallExpression` (`Find`, then `base`, then `VisitFunction(node.Method)`):

- **B flips to `GS0514@7`** ✔
- **E still escapes — no diagnostic, prints `body-33`** ✘
- D and the declaration-only control still correctly allowed — no over-correction
- the PR's own **19/19 pins stay green**

So patching the shape a reviewer found leaves the next shape open. **Enumeration is the wrong shape of fix here**, and narrowing `:408` cannot substitute for it — a fire-and-forget task is never observed, so rethrowing would only convert a swallowed exception into an unobserved one.

**Recommended fix — make it conservative rather than exhaustive.** Have the walker record "I could not resolve this callee's body" (any call node whose body `VisitFunction` cannot find, plus the node kinds above). If a free-standing `go` contains an unresolved callee **and** the program declares any P/Invoke, refuse. That restores the by-construction guarantee for the one path that has no backstop, keeps `PInvokeDeclaration_WithoutUse_IsAllowed` working (no `go`, no refusal), and costs a bool. Whatever route is chosen, please pin **B and E specifically** — they are one line each in the existing `AssertPInvokeUseRejected` style.

---

## Round-1 items — both SHOULD-FIX addressed, NIT resolved

**SHOULD-FIX 1 (P/Invoke in `shared { init { } }` degraded to `GS9999`) — FIXED.** `Evaluator.cs:367` adds a `when` filter that rethrows diagnostic control signals instead of wrapping them in `TypeInitializationException`. Pinned by `PInvokeInsideStaticInitializer_ReportsLocatedGS0514`, which asserts id **and** file **and** `StartLine == 10` **and** message **and** `Output == ""` — not "a diagnostic appeared".

**SHOULD-FIX 2 (published references contradicted themselves) — FIXED.** `docs/diagnostics.md` and `website/docs/ref/diagnostics.md` both rewritten, consistently, and they now state the ADR-0156 model outright. They independently corroborate my driver table above.

**NIT (third commit's title inverted what it did) — RESOLVED**, commit dropped in the rebase.

---

## Mutants — 4 applied, 4 killed, disjoint kill sets

Target tree `head`; each asserted applied by `git diff` **and** `GSharp.Core.dll` md5 movement; scored against the PR's own 19 tests; restored by exact md5.

| mutant | hunk | mutation | Core.dll | result |
|---|---|---|---|---|
| **M-GO** | `Statements.cs:375` | `enclosingScope == null` → `!= null` | `c9522721`→`d2e00985` | **RED 3/19** |
| **M-REJECT** | `Evaluator.cs:229` | `function.IsPInvoke` → `!function.IsPInvoke` | →`71bfb185` | **RED 14/19** |
| **M-CATCH** | `Statements.cs:830` | `IsDiagnosticControlSignal: true` → `false` | →`7e9806a6` | **RED 4/19** |
| **MCTRL** (control) | `Compilation.cs:366` | `allErrors.Any()` → `!allErrors.Any()` | →`ddf65a7c` | **RED 18/19** |

3 / 14 / 4 — each hunk carries its own kill set; the anti-swallow work is genuinely load-bearing and not riding on the guards. Restored: `Evaluator.Statements.cs` `5a1f6503`, `Evaluator.cs` `d63d8e0c`, `GSharp.Core.dll` back to `c9522721` **exactly**.

---

## What is solid

- The anti-swallow work is real and multi-site, not cosmetic: `IsDiagnosticControlSignal` is honoured in `TryFindCatchHandler` (user `catch` clauses cannot eat GS0514), in the static-initializer filter, and in `Compilation.cs`'s diagnostic construction. `OtherEvaluatorBoundaryDiagnostic_CannotBeCaught` shows it generalises past P/Invoke to `GS0511`.
- Diagnostics moved from the **declaration** to the **use site** (A: `@4` → `@6`), which is the actual ask in #2986.
- The `go` walker has cycle protection (`visitedFunctions`) and correctly resolves stored closures at runtime.
- Assertions throughout are specific — id, file, line, message, and output value.
- Empty `catch` audit: per round-1 precedent I checked whose path each is on. `Evaluator.Expressions.cs:112` is `when (ex is not EvaluatorException)`, and `Statements.cs:473`/`:482` rethrow. **`:408` is the only user-facing swallow, and it is the one still bare.**

---

## Disclosure — three of my own probes were fixture errors, not results

1. My first run returned **`GS0316` on every `go` row**, which looked like a uniform result. It was my prelude missing `import Gsharp.Extensions.Go` — the exact error I disclosed in round 1, repeated. Dumping diagnostics instead of asserting is the only reason I caught it.
2. Two further shapes I attempted — a property getter and a `defer` — returned `GS0286`/`GS0124` from **my own syntax**, not from the product. **Discarded, not scored.** They are neither evidence of a gap nor evidence of safety.
3. My first enumeration of `FunctionSymbol`-bearing nodes grepped `public FunctionSymbol Function` and missed `BoundUserInstanceCallExpression`, which names the member **`Method`**. I nearly concluded the walker's coverage was complete. *A zero can be syntax, not absence* — this time it would have inverted the verdict.

**Coverage:** denominator **9**, source `dotnet sln GSharp.sln list`. The unfiltered 9-project run is **still in progress** at posting time (`Compiler.Tests` is the long pole) — that is a gap in evidence, not a pass, and I will post the numbers when it lands. The BLOCKER above does not depend on it: it is a measured `main`-vs-head behavioural regression, and every suite is green on both sides.

DavidObando:
### Addendum — coverage gap closed: **9 of 9, 15,369 tests, 0 failures, 1 skipped**

I labelled the suite run in progress rather than implying coverage. It has finished, unfiltered, on PR head `cf5b4886`:

|---|---|---|---|---|
| Compiler.Tests | 0 | 0 | 4260 | 0 |
| Core.Tests | 0 | 0 | 7128 | 1 |
| Interpreter.Tests | 0 | 0 | 1409 | 0 |
| InternalAnalyzers.Tests | 0 | 0 | 9 | 0 |

**15,369 passed / 0 failed / 1 skipped.** Denominator **9**, source `dotnet sln GSharp.sln list`. `Compiler.Tests` took 27m45s at load ~8–10.

Two things this run does establish, beyond "green":

1. **`Compiler.Tests` is green without the dropped sample-gate commit.** The rebase removed `fdcedd18` entirely, and the sample conformance gate passes anyway — because under ADR-0156 the P/Invoke sample now executes natively in file mode, so its declarations are no longer "unused". The round-1 NIT is resolved by deletion rather than by a title fix, which is the better outcome.
2. **`Interpreter.Tests` is 1409**, up from 1393 on `main` — the PR's 19 boundary cases are the delta, and they all pass.

**Verdict unchanged: NOT MERGEABLE**, one BLOCKER. Worth restating precisely because this table is green: **every one of these 15,369 tests passes on the PR, and `go h.CallNative()` still silently prints the wrong answer at exit 0 where `main` refuses with GS0514.** The suite cannot see the regression — only the `main`-vs-head behavioural comparison can, which is why shapes **B** and **E** need pinning as part of the fix rather than after it.

DavidObando:
## Round-3 review — **MERGEABLE** (no BLOCKER; 3 SHOULD-FIX, 1 NIT)

Reviewed `191145ec` against base `3ca776d3` and live `main` `8b32fc36`. Probes written from scratch before reading the test file; every shape measured on **three trees** (base / round‑2 head `a60ec99e` / round‑3 head).

### Driver correction that shaped this review

`--engine evaluator` **cannot** reach the evaluator from a file. `src/Repl/Program.cs:61-69` parses `engineChoice`, `:106` calls `RunScript(scriptPath, references)` — and `RunScript` (`:145`) never receives it, always using `EmittedProgramHost`. So `gsi --engine evaluator probe.gs` silently measures the *emitted* engine and proves nothing about GS0514. I drove `GSharpRepl` → `SessionEngine` in-process instead (the same entry point the tests and interactive `gsi` use).

### 1. B / C / E / H / I measured, plus three shapes the author did not list

`GS0514` present / `body-33` printed. **SILENT** = rc 0, no diagnostic **and** no output.

| probe | base `3ca776d3` | round‑2 `a60ec99e` | round‑3 `191145ec` |
|---|---|---|---|
| `pos` — no P/Invoke | clean, prints | clean, prints | clean, prints |
| `or2` — free `go`, **no P/Invoke** | clean, prints | clean, prints | **clean, prints** |
| `decl_only` — declared, unused | GS0514 | clean, prints | clean, prints |
| `a_direct` | GS0514 | GS0514 | GS0514 |
| **C** `c_instance_nogo` | GS0514 | GS0514 | GS0514 |
| **B** `b_instance_go` | GS0514 | **SILENT** | **GS0514** |
| **H** `h_instance_go_scope` | GS0514 | GS0514 | GS0514 |
| **E** `e_ctor_go` | GS0514 | **SILENT** | **GS0514** |
| **I** `i_ctor_go_scope` | GS0514 | GS0514 | GS0514 |
| `or1` — free `go`, unrelated P/Invoke | GS0514 | clean, prints | GS0514 |
| `f1` lambda + free `go` | GS0514 | GS0514 | GS0514 |
| `f2` lambda + `scope` `go` | GS0514 | GS0514 | GS0514 |
| `f3` interface dispatch + `scope` `go` | GS0514 | GS0514 | GS0514 |
| `f4` 3‑hop chain + `scope` `go` | GS0514 | GS0514 | GS0514 |
| **`f5` interface dispatch + free `go`** | GS0514 | **SILENT** | **GS0514** |
| `g1/g2/g3` user `catch` swallow | GS0514 | GS0514 | GS0514 |
| `mg1/mg3` method‑group reference | GS0514 | GS0514 | GS0514 |

Three things worth recording:

- **B and E reproduce independently**, and on round‑2 the failure is *worse than reported*: not `body-33`, but **completely empty output at rc 0**. The guard fires on the goroutine thread, the bare `catch` at `Evaluator.Statements.cs:408` eats it, and the body dies with it.
- **`f5` is a third escape shape not in the round‑2 table** — interface dispatch through a free‑standing `go`. Silent on round‑2, fixed here.
- **The `scope` controls (H, I, f2, f3) report on all three trees.** The mechanism stayed visible; it was never masked.

### 2. The escape closes **by construction**, not by enumeration

This is the question that decides the round, and the answer is structural.

`Evaluator.Statements.cs:429` — `finder.Function ?? program.Functions.Keys.FirstOrDefault(static f => f.IsPInvoke)`. The walker only *refines the location*; the fallback decides. Any free‑standing `go` in a compilation that declares a P/Invoke is refused whether or not the walker can trace the path.

That is only complete if `program.Functions` holds **every** P/Invoke. I probed the two ways out of that set, and the language forbids both:

- **Class member** — `@DllImport` + `;`-bodied `func` inside a `class` binds as an *abstract method*: `GS0388: Abstract method 'StrLen' must be declared 'open'…`, then `GS0386`, `GS0159`.
- **Local function** — `GS0206: Annotations are only allowed on variable declarations (var/let/const), not on this statement.`

**P/Invoke is top-level-only ⇒ `program.Functions` is complete ⇒ shape F cannot exist for a free-standing `go`.** Demonstrated live: `f5` refuses with the diagnostic anchored at `go Use()` — the walker *missed* interface dispatch and the fallback caught it anyway. Scope‑bound `go` needs no preflight because the task is awaited (`:397 enclosingScope.Tasks.Add(task)`), and `TryFindCatchHandler` now refuses to match a diagnostic control signal, so a user `catch` cannot eat it (`g1/g2/g3`).

I could not construct shape F, and I do not believe one exists.

### 3. Over-refusal — checked in both directions

- **No false rejection.** `pos` and `or2` (free‑standing `go`, no P/Invoke anywhere) are clean on all three trees. The opposite‑sign BLOCKER is absent.
- **`or1` (unrelated `go` + declared P/Invoke) is refused — but base refuses it too.** Round‑2 accepted it; round‑3 restores base behaviour. The conservatism is **not a regression against `main`**. The only net loosening versus base is `decl_only`, which is exactly what #2986 asked for.
- That loosening is safe because every route to the P/Invoke *value* is guarded, including reference-taking: `mg1`/`mg3` refuse at `var f = NativeStrLen` via `Evaluator.Expressions.cs:814`, before any indirect call.

### 4. Mutation audit — each mutant proven applied

Verified by `git diff --stat` **and** `GSharp.Core.dll` md5 movement before scoring.

| mutant | applied | result |
|---|---|---|
| drop fallback (`Statements.cs:429` → `finder.Function`) | diff 1±, md5 `0286409d`→`f551aad0` | **killed 1** — `Issue2986…:356 PInvokeDeclarationWithFreeGo_IsConservativelyRejected`; my `f5` escapes under it |
| drop instance guard (`Calls.cs:356`) | diff −2, md5 →`7547a81b` | **killed 1** — `:432 PInvokeInstanceDispatch_DefensiveGuardReportsGS0514` |
| drop catch bypass (`TryFindCatchHandler`) | diff −6, md5 →`a342cc2c` | **killed 4** — 3× `PInvokeBoundaryDiagnostic_CannotBeCaught` + `OtherEvaluatorBoundaryDiagnostic_CannotBeCaught`; my `g2`/`g3` swallow it |

Thin kill sets are legitimate here — narrow guards, narrow pins — and each is a real kill, not a no-result.

**Test oracle is strong (ADR-0154 satisfied).** `AssertPInvokeUseRejected` (`:507`) asserts `Assert.Single(cell.Diagnostics)`, exact ID, exact filename, per-case exact `StartLine`, **and `Assert.Equal(string.Empty, cell.Output)`**. That last assertion pins the wrong-answer channel rather than mere diagnostic presence — precisely the failure mode that made this a BLOCKER twice.

### 5. Suites / merged tree

- `Interpreter.Tests` on head: **1432 / 0 / 0 skipped**.
- CI: **all green, 28 checks**, including **`ilverify` (3m49s)**, `e2e`, all 9 test shards, `cs2gs-oahu`. Zero failures.
- Real merge with `main` `8b32fc36` (docs conflict resolved): **built 0 warnings**, `Interpreter.Tests` **1503 / 0**, and the **full 20-probe matrix is identical to head**. #3193 touched `Evaluator.Calls.cs` and #3206 migrated `Interpreter.Tests`; neither breaks this.

---

## SHOULD-FIX

**S1 — `docs/diagnostics.md` conflicts with `main`, and the conflict is *semantic*.**
`git merge-tree 8b32fc36 191145ec` → `CONFLICT (content)`. Main (from #3201) documents GS0514 as applying *"even when the declaration is not called because the interpreter cannot create a valid callable value"* — the exact contract this PR removes. Taking either side loses real content: main's side carries the Phase‑3a/3c retirement sentences this branch predates; this branch's side carries the new use-site contract. **Hand-merge both**; do not resolve by picking a side. `docs/diagnostics.md:1169-1188`.

**S2 — the instance-dispatch guard is unreachable from any G# source.**
`Evaluator.Calls.cs:356` guards `BoundUserInstanceCallExpression.Method.IsPInvoke`, but a P/Invoke cannot be an instance method (§2). Its test has to forge the state: `Issue2986InterpreterBoundaryTests.cs:452` does `method.PInvokeMetadata = native.PInvokeMetadata` on an ordinary method with a body. So the mutant kill proves *the line executes*, not that any user program reaches it. It is honestly named `DefensiveGuard` and is 2 lines — but either delete it, or comment that it is unreachable by construction, so a later reader does not mistake it for live coverage.

**S3 — two changes have compiler-wide blast radius and are not mentioned in the PR body.**
`BoundTreeRewriter.cs:623` changes `new BoundCallExpression(null, …)` → `node.Syntax` for **every** rewritten call, and `ExpressionBinder.cs:1318/1404/1415` does the same for method groups. These affect the emit path too, not just the interpreter. CI including `ilverify` is green, so the evidence is good — but say so in the PR body, because "propagate syntax so the diagnostic has a location" is a global change dressed as a local one.

## NIT

**N1 — `--engine evaluator` is accepted then silently ignored for scripts.** `src/Repl/Program.cs:61-69` parses it; `:106 RunScript(scriptPath, references)` drops it. Pre-existing on this head and not caused by this PR, but it makes this boundary unreachable from any file driver and will mislead the next reviewer. Worth its own issue.

---

## Disclosure

**Coverage:** I ran `Interpreter.Tests` only (1 of 9 projects) locally, plus the merged tree. The other 8 rest on CI, which is green.

**My own errors, for calibration:** (1) my first `RejectPInvoke` grep ran against the tree I was *standing in* (a previous PR's head) rather than the subject — caught before drawing any conclusion; (2) my first probe run failed 100% with `GS0124` because `go` needs a value-returning target — my fixture bug, not a product finding; (3) the interface probes first failed `GS0368` for a missing `;`; (4) my md5 round-trip appeared to mismatch on restore — the cause was **my own out-of-tree harness**, whose `ProjectReference` into `src/Repl` rebuilds `GSharp.Core.dll` in a different MSBuild context and overwrites it. Source restoration is proven by `git diff` = 0 at a fixed commit plus a clean 1432/1432 re-run; the md5 excursion was mine, not the PR's. Generally worth knowing: an out-of-tree harness referencing repo projects silently perturbs the freshness witness.

Worktree restored; `git status --porcelain` empty.

DavidObando:
## Addendum — re-measured against new `main` `0484b00f1`. **Verdict unchanged: MERGEABLE.** S1's remedy changed materially.

My review merge-tested against `main` `8b32fc36`. `main` has since moved +6 (`#3210`, `#3213`, `#3192`, and ADR-0156 Phase 3b.1 slices `#3221`/`#3225`/`#3228`). Re-measured on `0484b00f1`.

**Nothing in §1–§5 changes.** The Phase 3b.1 slices touch `Core.Tests`, which this PR does not; `Evaluator.Calls.cs` and `Issue2986InterpreterBoundaryTests.cs` still auto-merge. The probe matrix, the by-construction argument, and the mutation results all stand.

But **S1 is now worse than I described it**, in two specific ways I could only see once `main` moved.

### S1a — after this merge, *both sides of the conflict contain a false statement*

The conflict is a single hunk. Reproduced with `git merge-file` on the three blobs from `git merge-tree 0484b00f1 191145ec`:

```
<<<<<<< main
compiler error. It applies even when the declaration is not called because
the interpreter cannot create a valid callable value for direct or indirect
use. Since ADR-0156 Phase 3a no driver interprets by default: `gsi <file>`,
bare `gsc`, and the interactive REPL all execute emitted code ... This
diagnostic now fires only under gsi's deprecated `--engine evaluator` escape
hatch and retires with the evaluator in Phase 3c.
=======
compiler error. It is reported only when interpreted execution reaches a call
or function reference, or when a compilation containing P/Invoke launches a
free-standing goroutine whose failure would otherwise be unobserved. Declaration
alone remains valid. Since ADR-0156 Phase 1 only the interactive REPL interprets:
`gsi <file>` and bare `gsc` execute emitted code ...
>>>>>>> branch
```

- **Taking `main`'s side** ships *"applies even when the declaration is not called"* — which this PR deliberately makes false (`decl_only` is clean on this head, by design, per #2986).
- **Taking the branch's side** ships *"Since ADR-0156 **Phase 1** only the **interactive REPL interprets**"* — which **#3201 (Phase 3a) already made false**: the interactive REPL now defaults to the emitted engine, and the evaluator is reachable only via `--engine evaluator` / `GSI_ENGINE=evaluator`, retiring in Phase 3c.

So this is not "pick the right side carefully". **Neither side is correct; the resolution has to be newly written text** — the branch's use-site contract (call/reference or free-standing `go`; declaration alone valid) combined with `main`'s Phase-3a/3c driver model. The branch's wording is not wrong-by-authorship; it was written before #3201 landed.

### S1b — the mirrored doc **auto-merges silently into that same false statement**

`website/docs/ref/diagnostics.md` is a hand-maintained mirror (I looked for a generator and found none — only a Docusaurus nav entry at `website/docusaurus.config.ts:130`). `#3201` updated the Phase-3a prose in `docs/diagnostics.md` **but not in the mirror**, so the branch's rewrite applies to the mirror without conflict. Measured, not inferred:

```
git merge-file mir_merged.md mir_base.md mir_branch.md   →  conflicts=0
grep → website/docs/ref/diagnostics.md:1776
  "alone remains valid. Since ADR-0156 Phase 1 only the interactive REPL interprets:"
```

**`docs/diagnostics.md` conflicts and gets human attention; the mirror does not conflict and ships the stale claim with nothing to prompt review.** Whoever resolves S1a will very plausibly never learn a second file needed the same correction. Please fix both, and make them say the same thing.

Still **SHOULD-FIX, not BLOCKER** — documentation only, no behaviour change, and the code findings are unaffected.

### Disclosure — a near-miss in this addendum

My first conflict check ran `grep -c '^<<<<<<<'` on `git merge-tree` output and returned **0**, which reads exactly like "conflict resolved itself, S1 is stale". It was wrong: git 2.50's `merge-tree` prints stage-entry lines and `CONFLICT (content): ...`, **not** conflict markers. The zero was my grep syntax, not absence of a conflict. I caught it because the same output contained `CONFLICT (content): docs/diagnostics.md` two lines further down. Recording it because a clean-looking zero here would have retracted a valid finding.

(Second, smaller one: `set -e` aborted my extraction script silently, because `git merge-file` exits with the **conflict count** rather than a failure code.)

DavidObando:
## Duck verdict — PR #3104 (`fd2d64c9`): **BLOCKER (superseded) + 1 SHOULD-FIX + 1 SHOULD-FIX + 1 NIT**

Fresh worktrees, own `out/`, CI-flagged builds. `uptime` 4d20h. Head `fd2d64c95`, merge-base `0d60aa1f1`.
All evaluator probes driven **in-process via `SessionEngine`** — never `gsi --engine evaluator <file>` (#3230).

---

## BLOCKER — `main` deleted the thing this PR refines, *during this review*

`origin/main` was `a5de8fd0c` when I fetched. It is now **`8b3bdac57` — #3244, "Retire the tree-walking evaluator (ADR-0156 Phase 3c)"**.

Measured against `origin/main` @ `8b3bdac57`:

| probe | result |
|---|---|
| `git cat-file -e origin/main:src/Core/CodeAnalysis/Evaluator.cs` | **ABSENT** |
| `git cat-file -e origin/main:src/Repl/Engine/SessionEngine.cs` | **ABSENT** |
| `origin/main:src/Core/CodeAnalysis/DiagnosticDescriptors.cs \| grep -c GS0514` | **0** |
| `gh issue view 2986 -q .state` | **CLOSED** |
| `origin/main:docs/adr/0152-…md:3` | `**Status**: Superseded by ADR-0156 … deleted in Phase 3c` |
| `origin/main:test/Interpreter.Tests/Issue2986InterpreterBoundaryTests.cs` | **79 lines**, one emitted-path test, zero evaluator pins |

`git merge-tree --write-tree pr3104 origin/main` **output** (not exit code) — **15 `CONFLICT` lines**, five of them `modify/delete` on precisely the files that carry the fix:

```
CONFLICT (modify/delete): src/Core/CodeAnalysis/Evaluator.cs deleted in origin/main and modified in pr3104.
CONFLICT (modify/delete): src/Core/CodeAnalysis/Evaluator.Calls.cs …
CONFLICT (modify/delete): src/Core/CodeAnalysis/Evaluator.Expressions.cs …
CONFLICT (modify/delete): src/Core/CodeAnalysis/Evaluator.Statements.cs …
CONFLICT (modify/delete): src/Core/CodeAnalysis/EvaluatorException.cs …
CONFLICT (content): docs/diagnostics.md · website/docs/ref/diagnostics.md
CONFLICT (content): Compilation.cs · DiagnosticDescriptors.cs · DiagnosticBag.Reports.Declarations.cs
CONFLICT (content): Issue2986/758/760/761/762 …InterpreterTests.cs
```

`RejectPInvoke`, the `go` preflight, the control-signal `catch` bypass, the static-init preservation, the 401 test lines, both ADR rewrites and both diagnostics-doc rewrites are **dead on arrival** — they refine `GS0514`, which no longer exists. CI is green because it ran on the pre-#3244 base.

### ponytail — what actually survives

Exactly the S3 hunks. They are **conflict-free** against `8b3bdac57` and touch files main still has:

```
src/Core/CodeAnalysis/Binding/BoundTreeRewriter.cs                       |  2 +-
src/Core/CodeAnalysis/Binding/ExpressionBinder.cs                        | 17 +++++-----
src/.../OverloadResolution/OverloadResolver.CallBinding.cs               | 10 +++---
3 files changed, 15 insertions(+), 14 deletions(-)
```

`origin/main:BoundTreeRewriter.cs:623` is still `new BoundCallExpression(null, …)` and `origin/main:InterpolatedStringHandlerLowerer.cs:153` is unchanged, so the one measurable improvement below still applies. **+15/−14 out of +715/−97.** Everything else should be dropped, not rebased.

---

## Q1 / S3 — span comparison: **one span moves, and it is an improvement; the coverage claim is false**

### Consumer census
Exactly **one** diagnostic in `src/Core/` (outside `/Emit/`) anchors on an expression node's `.Syntax`:
`src/Core/CodeAnalysis/Lowering/InterpolatedStringHandlerLowerer.cs:153`
`var location = part.Value.Syntax?.Location ?? node.Syntax.Location;` → **GS0519**.
The only other consumer is `MethodBodyEmitter.cs:124` `CurrentAnchor`, read at `FunctionEmitter.cs:179/371` and `ConstructorBodyEmitter.cs:566/618` — internal-error wrapping only.
Sequence points come from `statement.Syntax` (`MethodBodyEmitter.cs:485`), never expressions.

### Measured span move
`s1.gs` — `ref struct Token`, `func MakeToken() Token`, `Console.WriteLine("token=${MakeToken()}")`:

| build | reported span |
|---|---|
| base `0d60aa1f1` | `s1.gs(10,23,10,45)` — the whole interpolated string |
| head `fd2d64c9` | `s1.gs(10,32,10,43)` — `MakeToken()` |

New span is **correct and strictly more precise**. Not a regression.

### Blast radius, measured not reasoned
All 122 compilable `samples/*.gs`, `gsc /target:exe /out:`, base vs head: **`same=122 differ=0 skipped=0`**, DLL md5 identical, zero PDB divergence. **S3 changes no IL and no debug info.**

### Mutant M1 — the disclosure's coverage claim is measurably false
`BoundTreeRewriter.cs:623` `node.Syntax` → `null` (two-sided PRODUCT revert). Applied per `git diff HEAD`; `GSharp.Core.dll` md5 `a56efa02c1403f4cd02587be4e484eac` → `a2b59d6643cfa3024624daf7ba468138`; restored to `a56efa02…` **exactly** (all three builds `-p:ContinuousIntegrationBuild=true`, tests `--no-build`).

- **`Core.Tests`: 7163 passed, 1 skipped — GREEN.**
- `Interpreter.Tests`: **1508 / 1509**, one death: `PInvokeCallWithRewrittenArgument_ReportsUseLocation`.

That single pin lives in `Issue2986InterpreterBoundaryTests.cs`, which **main has truncated to 79 lines with no evaluator pins**. So:

> "The full build, emitted-path tests, and required `ilverify` check cover that wider blast radius."

is **false today** — no emitted-path test covers it (122/122 identical assemblies, `Core.Tests` fully green under the mutant) — and **post-#3244 the hunk becomes entirely unpinned**. The disclosure exists and is otherwise accurate; the coverage sentence is not.

**SHOULD-FIX:** if S3 is salvaged, pin the GS0519 span. `Issue3114ReadOnlySpanInterpreterTests.cs:361` (`AssertByRefLikeInterpolationDiagnostic`) asserts only `startLine` — it would not catch this move. Assert the **column**, with a call-valued hole.

---

## Q2 / S2 — guard deletion is correct; the stated reason is **factually wrong**

> "G# P/Invoke declarations are top-level-only."

**False.** `PInvokeBinder.TryAttachPInvokeMetadata` has **two** callers:
- `src/Core/CodeAnalysis/Binding/DeclarationBinder.Functions.cs:377` — top-level free functions
- **`src/Core/CodeAnalysis/Binding/DeclarationBinder.Structs.cs:2353`** — `shared { }` **static members** (ADR-0086 / #1203, pinned by `Issue1203SharedExternBinderTests.cs:29`)

Measured: `class C { shared { @DllImport("libc", EntryPoint: "strlen"…) func Native(text string) nint; } }` binds as P/Invoke and `C.Native("Hello")` reports **`GS0514@5`**.

**The conclusion still holds** — the correct argument is *"P/Invoke attaches only to top-level free functions and `shared` static members; **neither is instance-dispatched**"*, and `shared` members have no `ThisParameter`, so `Evaluator.Calls.cs:356` was unreachable **by construction**, not merely unobserved. Deleting it is the ponytail-correct call, and the removed test was genuinely forged: it assigned `method.PInvokeMetadata = native.PInvokeMetadata` onto a *managed* method — a state the binder cannot produce. No real coverage lost.

The same wrong premise is reused to justify the conservative `go` fallback set (`Evaluator.Statements.cs:429`, `program.Functions.Keys.FirstOrDefault(f => f.IsPInvoke)`). I measured that independently: for the static-member program, `pinvokeInFunctions=Native, totalFns=2` → **the set is complete anyway**. Right answer, wrong reason, twice. SHOULD-FIX on the reasoning, not the code.

---

## Q4 — must-still-be-refused: **21 shapes, no silent acceptance**

In-process `SessionEngine { CaptureConsole = true }.Evaluate(...)`, head build.

| # | shape | result |
|---|---|---|
| D0 | declaration only, unused | clean, `33` — *intended loosening* |
| D1 | direct call | `GS0514@5` |
| D2 | method-group ref, never invoked | `GS0514@5` |
| D3 | **`shared` static-member P/Invoke, called** | `GS0514@5` |
| D4 | static-member declaration only | clean, `33` |
| D5 | behind a user wrapper function | `GS0514@6` |
| D6 | call in `if false` branch | clean, `55` — never executes |
| D7 | inside an invoked lambda | `GS0514@5` |
| D8 | static-member method group | `GS9999` — **D11 control** (managed static-member method group) also `GS9999` ⇒ pre-existing, not a P/Invoke escape |
| D9 | inside user `try/catch` | `GS0514@6` — not swallowed |
| E1 | class instance-method body | `GS0514@7` |
| E2 | static field initializer | `GS0514@7` |
| E3 | iterator, enumerated | `GS0514@6` |
| E4 | iterator, never enumerated | `GS0514@6` — over-refusal, loud |
| E5 | `async`, awaited | `GS0514@7` |
| E6 | `defer` | `GS0514@3` after `body` |
| E7 | interface dispatch | `GS0514@11` |
| E8 | **control**, no P/Invoke, same shapes | clean, `7` |
| G1 | top-level P/Invoke + free `go` on a managed call | `GS0514@10` — conservative |
| G2 | static-member P/Invoke + free `go` on a managed call | `GS0514@14` |
| G3 | static-member + free `go`, direct | `GS0514@6` |
| G4 | static-member + free `go`, stored closure | `GS0514@6` |

Every shape is either `GS0514` or a loud pre-existing refusal. **Nothing silently compiles and runs.** The only loosenings are declaration-only and never-executed code — exactly the stated intent. No loud refusal traded for a silent wrong answer.

---

## Q3 / docs — the mirror trap is **closed**

`docs/diagnostics.md:1165-1180` and `website/docs/ref/diagnostics.md:1773-1788` GS0514 sections are **semantically identical**; the only delta is the ADR link form (relative vs absolute) and line wrapping. Both are true against the pre-#3244 tree and read as one rule, not a seam. The `--engine evaluator` sentence is worded so #3243 does not falsify it.

**NIT:** `docs/diagnostics.md:1178-1180` says the diagnostic "fires only under gsi's deprecated `--engine evaluator` escape hatch", while the PR's own `ADR-0152` correctly names `SessionEngine`, `Compilation.Evaluate`, **and interactive** `gsi --engine evaluator` — the docs sentence omits the embedding API the PR's own tests drive. Moot if the docs half is dropped, which it should be: main's `docs/diagnostics.md:1160` already carries the retired stub, and `ADR-0152` on main is already headed *Superseded*.

Separately, and **not this PR's defect**: `origin/main:docs/adr/0086-pinvoke-native-interop.md:19-21` still describes the deleted evaluator in the present tense. That is ADR-0156 Phase 3d territory.

---

## Verdict

**BLOCKER: superseded by #3244.** Do not rebase this PR — ~90% of it refines a deleted diagnostic against deleted files. Close it, and re-land the **+15/−14 S3 hunk alone** on `8b3bdac57`, with a GS0519 **column** assertion as its pin (it currently has none that survives #3244).

Denominator 9 (`dotnet sln GSharp.sln list`). I ran 2 of 9 locally (`Core.Tests` 7163+1, `Interpreter.Tests` 1509+4 — both matching the PR body) plus a 122-sample emit sweep; CI's 9/9 green is against the pre-#3244 base and does not speak to `8b3bdac57`. Worktrees clean apart from pre-existing #3170 `packages.lock.json`. `Core.dll` md5 restored exactly.

DavidObando:
## Duck update — base re-measured, S2 unreachability closed by construction

Verdict **unchanged: BLOCKER (superseded)**. Two corrections to the record.

### 1. The Phase 3b retirement-gate world no longer exists

I was briefed that `main` is `6bf63e333` with #3238/#3241/#3242 (`[Obsolete]` on `Compilation.Evaluate`, `EvaluatorRetirementGateTests` CS0618 allowlist). Verified by **ancestry, not existence**:

```
6bf63e333  Fix evaluator object reference equality (#3195)     <- NOT #3242
  ancestor of 8b3bdac57 ?  YES
8b3bdac57  Retire the tree-walking evaluator (Phase 3c) (#3244)
5e2fee875  Consult the submission fallback ... (#3253)          <- origin/main today
```

`6bf63e333` is an **ancestor** of `8b3bdac57`; `main` is two commits past it. Phase 3b's scaffolding was deleted by Phase 3c ~30 minutes later. Measured on `5e2fee875`:

| briefed claim | measured |
|---|---|
| `Compilation.Evaluate` ×3 `[Obsolete]` at `Compilation.cs:311/321/332` | `src/Core/CodeAnalysis/Compilation/Compilation.cs` contains **zero** `Evaluate`, `Obsolete`, `ContinueWith` |
| `EvaluatorRetirementGateTests.cs` pins the CS0618 allowlist | **ABSENT** |
| `SessionEngine.cs` is on the allowlist | **ABSENT** |
| repo-wide CS0618 suppressions | **1 file** — `src/Core/CodeAnalysis/Emit/MaxStackTracker.cs`, unrelated to the evaluator |
| `Evaluator*` in `src/Core/CodeAnalysis` | **none** |

So "run `EvaluatorRetirementGateTests`" is unrunnable — the gate is gone. I ran its **substance** instead: `git diff 0d60aa1f1 fd2d64c9 | grep CS0618` → **no suppression added or removed**, and all nine added `.Evaluate(` calls are `SessionEngine.Evaluate`, never `Compilation.Evaluate`. Against `6bf63e333` this PR would have passed the gate cleanly. Against real `main` it does not compile — `SessionEngine` no longer exists.

`git merge-tree --write-tree pr3104 origin/main` re-run against **current** `5e2fee875` (output read, grepped for `CONFLICT`, not `<<<<<<<`): **15 conflicts, 5 modify/delete** — `Evaluator.cs`, `Evaluator.Calls.cs`, `Evaluator.Expressions.cs`, `Evaluator.Statements.cs`, `EvaluatorException.cs`. Identical to my earlier measurement. Nothing about the supersession softened.

### 2. S2 — unreachability proved over the binder's write-set, not over the corpus

Maximum suspicion applied. My argument never rested on a surviving mutant, so it survives the standard being raised. Census at `fd2d64c9` (`git grep` over all of `src/`, not a first-line sample):

1. `src/Core/CodeAnalysis/Symbols/FunctionSymbol.cs:450` — `IsPInvoke => PInvokeMetadata != null`. Sole definition.
2. `src/Core/CodeAnalysis/Binding/PInvokeBinder.cs:270` — `function.PInvokeMetadata = metadata;` — **the only write in the repo**.
3. Reachable only through `TryAttachPInvokeMetadata`, which has **exactly two** callers:
   - `DeclarationBinder.Functions.cs:377` — top-level free function, no receiver at all.
   - `DeclarationBinder.Structs.cs:2353` — sits **unconditionally downstream** of `receiverType: null` and `methodSymbol.IsStatic = true` at `:2323-2324`, with no intervening branch.
4. The deleted guard sat at `Evaluator.Calls.cs:356`, immediately before `[method.ThisParameter] = receiverValue`.

**The specimen that should reach it is unconstructible, and I can name the line where it fails.** The instance-member path handles a bodyless method at `DeclarationBinder.Structs.cs:1231` by setting `IsAbstract = true` (GS0388 unless `open` member of `open` class) and **never calls `TryAttachPInvokeMetadata`**. So `@DllImport` on an instance member binds with `IsPInvoke == false`. No source shape puts a receiver-bearing symbol into the write-set.

Positive direction already measured: **D3**, the closest constructible specimen — `shared`-block static P/Invoke invoked as `C.Native("Hello")`, a member access on a type — routed to the **surviving** guard at `Evaluator.Calls.cs:44` and reported `GS0514@5`.

Unreachable **by construction over the write-set** — a property of the binder, not of the test corpus. The deletion is correct and the removed test was genuinely forged (`method.PInvokeMetadata = native.PInvokeMetadata` onto a managed method). The PR body's stated reason — *"top-level-only"* — remains **false** (`DeclarationBinder.Structs.cs:2353`); correct claim is *"only ever receiverless statics."* SHOULD-FIX on the reasoning, not the code.

**Corroborating the thin-coverage warning, six lines from our own guard:** the code immediately above it on base is the `overrideMethod.IsOverride` virtual-dispatch loop (`Evaluator.Calls.cs:346-350`) — the same line a sibling mutated with 8,646 tests silent. Independent confirmation that evaluator coverage cannot be used as a reachability oracle here, and exactly why I anchored on the write-set instead.

### Standing

**BLOCKER: superseded by #3244.** Do not rebase. Salvage is the **+15/−14** S3 syntax-propagation hunk across `BoundTreeRewriter.cs` / `ExpressionBinder.cs` / `OverloadResolver.CallBinding.cs` — conflict-free against `5e2fee875`, still `null` on main at `BoundTreeRewriter.cs:623` — re-landed standalone with a **column-asserting** GS0519 pin (measured base `(10,23,10,45)` → head `(10,32,10,43)`; its only current pin is an evaluator test #3244 deleted, and `Core.Tests` was 7163/7163 green under the reverting mutant).

DavidObando:
Closing as superseded by #3244 (ADR-0156 Phase 3c), which deleted every `Evaluator*` file this PR edits. `merge-tree` against current `main` reports **15 conflicts, 5 of them modify/delete**; the branch does not compile against `main`.

Two findings from the review are worth preserving rather than losing with the branch.

**S2 was correct, and for a better reason than the PR claimed.** The deleted instance-dispatch guard was genuinely unreachable — but not because a mutant survived a test run. It is unreachable **by construction over the write-set**: `FunctionSymbol.IsPInvoke` has a sole definition, `PInvokeBinder` contains the only write in the repo, and its two callers are receiverless or unconditionally `IsStatic`. The specimen that would reach the guard is unconstructible because `DeclarationBinder.Structs.cs:1231` handles a bodyless instance member by setting `IsAbstract = true` and **never calls `TryAttachPInvokeMetadata`**, so `@DllImport` on an instance member binds with `IsPInvoke == false`.

That is the right shape of argument, and it matters because the mutant-based version of this reasoning failed badly elsewhere today: the `overrideMethod.IsOverride` loop six lines above this guard was mutated on a sibling PR and **8,646 tests did not notice**. Coverage answers *"did anyone look"*, never *"can this happen."*

The PR body's *"top-level-only"* characterisation was nonetheless wrong; the accurate claim is *"only ever receiverless statics."*

**S3 is salvageable and still applies to `main`.** The +15/−14 hunk is conflict-free against current `main` and `BoundTreeRewriter.cs:623` is still `null` there. It is being re-filed as its own issue so it is not lost with this branch. It needs a **column-asserting** GS0519 pin — the span moves from `(10,23,10,45)` to `(10,32,10,43)` — because its only previous pin was an evaluator test that #3244 deleted.

Issue #759, opened 2026-06-12
P/Invoke: struct and class marshalling


DavidObando:
Follow-up to #727 / ADR-0086. v1 marshals only primitives, string, *T (byref), and slices of primitives. Struct and class marshalling (LayoutKind, FieldOffset, blittable detection) is deferred.

PR #3105, opened 2026-08-02
Fix evaluator explicit field overlap

### LINKED ISSUE #3100: Evaluator and gsi ignore @FieldOffset overlap, silently producing a different value than emitted code
## Problem

Both evaluators (`gsc` in evaluate mode and `gsi`) ignore `@FieldOffset` overlap on a struct. Emitted code honours it. The result is a **silent wrong answer**: rc 0, no diagnostic, different value.

Using `samples/PInvokeStructMarshalling.gs`:

| path | output |
|---|---|
| `gsc /out:` + run (emit) | `3/4/6153737367135073092` |
| `gsc` (evaluate) | `3/4/`**`0`** |
| `gsi` | `3/4/`**`0`** |

Both interpreting paths exit 0 with no diagnostic.

The overlapping fields are declared at `samples/PInvokeStructMarshalling.gs:33-37`.

## Why this is filed separately

`test/Compiler.Tests/LanguageConformance/SampleConformanceTests.cs:60-63` (on the #2939 branch) ledgers this sample's divergence as issue #2986, "cannot execute the native struct-marshalling call". **That attribution is wrong** — the sample contains no native call; the only match for one is a comment at line 6. The six genuine native-boundary rows in that ledger are all `Expected(1, "", "GS0514")`; this row is `Expected(0, …)`, which is structurally not a native difference.

So the divergence is real but its recorded cause is not, and it is being carried as an intentional boundary rather than as a bug. It is `@FieldOffset` overlap being ignored by both evaluators.

DavidObando:
## Summary
- make `StructValue` use a runtime-generated CLR backing type for explicit layouts
- route shared field storage writes through that backing so overlapping fields alias in bare `gsc` and `gsi`
- preserve explicit-layout value semantics when structs are copied, including nested structs
- add emit/evaluate/gsi parity coverage for full overlap, partial overlap, default and sequential controls, explicit non-overlap, writes through both aliases, ref writes, and nested structs
- verify emitted layout through `Assembly.Load`, `GetTypes()`, and `Marshal.OffsetOf`

## Design
Took faithful-semantics rung, not diagnostic rung. Evaluators already centralize user-defined aggregate values in `StructValue`; generating a CLR layout there reuses runtime field-layout semantics without separate engine patches or manual byte-layout rules. `FieldCollection` keeps all existing read/write consumers on one shared representation.

## Measured sample
`PInvokeStructMarshalling.gs` now prints identical values in all three drivers:

- bare `gsc`: `3 / 4 / 6153737367135073092`, rc 0
- emitted `gsc /out:`: `3 / 4 / 6153737367135073092`, compile/run rc 0
- `gsi`: `3 / 4 / 6153737367135073092`, rc 0

Before this change, both interpreting paths printed `3 / 4 / 0` with rc 0.

## Validation
- `Issue3100FieldOffsetParityTests` + issue #759 tests: 10/10
- full `Interpreter.Tests`: 638/638
- six product condition mutants built successfully and were killed by the overlap/parity rows (constructor activation, alias refresh, explicit copy, nested write conversion, nested read conversion, runtime layout kind)
- fresh post-clean `Core`, `Compiler`, and `Repl` artifacts verified by mtime

Fixes #3100

### DIFF
--- a/src/Core/CodeAnalysis/StructValue.cs
+++ b/src/Core/CodeAnalysis/StructValue.cs
@@ -3,8 +3,15 @@
 // </copyright>
 
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Emit;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using System.Text;
+using System.Threading;
 using GSharp.Core.CodeAnalysis.Symbols;
 
 namespace GSharp.Core.CodeAnalysis;
@@ -17,18 +24,39 @@ namespace GSharp.Core.CodeAnalysis;
 /// </summary>
 public sealed class StructValue
 {
+    private readonly LayoutRuntime explicitLayout;
+    private object explicitLayoutValue;
+
     /// <summary>Initializes a new instance of the <see cref="StructValue"/> class.</summary>
     /// <param name="structType">The struct type symbol.</param>
     public StructValue(StructSymbol structType)
     {
         StructType = structType;
-        Fields = new ConcurrentDictionary<string, object>();
+        Fields = new FieldCollection(this);
+
+        if (structType.LayoutMetadata?.Layout == LayoutKind.Explicit)
+        {
+            explicitLayout = LayoutRuntime.Get(structType);
+            explicitLayoutValue = explicitLayout.CreateDefault();
+            explicitLayout.ReadFields(explicitLayoutValue, Fields);
+        }
     }
 
-    private StructValue(StructSymbol structType, ConcurrentDictionary<string, object> fields)
+    private StructValue(
+        StructSymbol structType,
+        ConcurrentDictionary<string, object> fields,
+        LayoutRuntime explicitLayout = null,
+        object explicitLayoutValue = null)
     {
         StructType = structType;
-        Fields = fields;
+        Fields = new FieldCollection(this);
+        foreach (var field in fields)
+        {
+            Fields.SetRaw(field.Key, field.Value);
+        }
+
+        this.explicitLayout = explicitLayout;
+        this.explicitLayoutValue = explicitLayoutValue;
     }
 
     /// <summary>Gets the struct type.</summary>
@@ -45,7 +73,7 @@ public sealed class StructValue
     // Dictionary&lt;,&gt; is not thread-safe under concurrent writes.
 
     /// <summary>Gets the mutable field dictionary backing this instance.</summary>
-    public ConcurrentDictionary<string, object> Fields { get; }
+    public FieldCollection Fields { get; }
 
     /// <summary>
     /// Gets or sets the CLR backing instance for a GSharp class that inherits an
@@ -62,6 +90,22 @@ public sealed class StructValue
     /// <returns>A new instance with the same fields.</returns>
     public StructValue Copy()
     {
+        if (explicitLayout != null)
+        {
+            object layoutCopy;
+            lock (explicitLayoutValue)
+            {
+                layoutCopy = RuntimeHelpers.GetObjectValue(explicitLayoutValue);
+            }
+
+            var fieldsCopy = new ConcurrentDictionary<string, object>();
+            explicitLayout.ReadFields(layoutCopy, fieldsCopy);
+            return new StructValue(StructType, fieldsCopy, explicitLayout, layoutCopy)
+            {
+                ClrBacking = this.ClrBacking,
+            };
+        }
+
         var copy = new ConcurrentDictionary<string, object>();
         foreach (var kvp in Fields)
         {
@@ -141,6 +185,23 @@ public sealed class StructValue
         return sb.ToString();
     }
 
+    private void SetField(string name, object value)
+    {
+        if (explicitLayout != null)
+        {
+            lock (explicitLayoutValue)
+            {
+                if (explicitLayout.TrySetField(explicitLayoutValue, name, value))
+                {
+                    explicitLayout.ReadFields(explicitLayoutValue, Fields);
+                    return;
+                }
+            }
+        }
+
+        Fields.SetRaw(name, value);
+    }
+
     // Rubber-duck follow-up to issue #2224: an anonymous-class literal's
     // synthesized type exposes get-only auto-properties, not plain fields
     // (see AnonymousTypeCache); its runtime storage in Fields lives under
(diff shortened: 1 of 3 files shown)


DavidObando:
## Verdict: MERGEABLE

Measured on the **merged tree** — `c72c61c98` merged with `main` `abfab99c` (`MERGE_RC=0`, merge commit `4b468cf9`), because a clean `merge-tree` is a floor and not a proof. Solution build **`BUILD_RC=0`, 0 Warning(s), 0 Error(s)**. `Core/`, `Compiler/` and `Repl/` mtimes verified fresh before every probe; every probe compiled into its own directory, asserted empty (`entries_before=1`) with a unique package name.

Scope, three-dot from a computed merge base (`4deeed8d`, `--full-index`): **3 files, +577 −6** — `StructValue.cs`, the new parity tests, and a 4-line touch to the #759 tests. That is exactly what `Fixes #3100` describes, and nothing else; no other issue is silently closed.

### The issue's verbatim repro, run byte-for-byte first

`samples/PInvokeStructMarshalling.gs`, sha256 `1c9949a8…` identical to the file in the tree:

| driver | `main` `abfab99c` | merged tree |
|---|---|---|
| `gsc /out:` + run (emit) | `3/4/6153737367135073092` | `3/4/6153737367135073092` |
| bare `gsc` (evaluate) | `3/4/`**`0`** rc0 | `3/4/`**`6153737367135073092`** rc0 |
| `gsi` (interpret) | `3/4/`**`0`** rc0 | `3/4/`**`6153737367135073092`** rc0 |

The three columns agree on the **exact** value, not merely a non-zero one — `0x5566778811223344 = 6153737367135073092`, matching `samples/PInvokeStructMarshalling.golden` and the little-endian composition of `LowPart`/`HighPart`.

### This is strategy (1), so I attacked partial coverage — it holds up

The approach is to synthesise a real CLR type per struct symbol (`LayoutRuntime`, `StructValue.cs:263-285`) in a dynamic assembly and delegate layout to the runtime, rather than hand-rolling offset arithmetic. That is why the coverage generalises. Seven top-level probes, all three drivers, both trees, distinct non-zero values:

| probe | shape | `main` eval/gsi | merged eval/gsi | emit (oracle) |
|---|---|---|---|---|
| pa | reverse: write `QuadPart`, read both halves | `0/0` | `287454020/1432778632` | same |
| pb | **partial** overlap, `int32@0` + `int32@2` | `0` | `4386` | same |
| pg | **constructor path only**, no assignment | `0/0` | `287454020/1432778632` | same |
| pf | **nested struct** overlapping an `int64` | `0` | `6153737367135073092` | same |
| pc | explicit layout, **no** overlap (control) | `33/22` | `33/22` | same |
| pd | sequential layout (control) | `33/22` | `33/22` | same |

`pb` returning `4386` is the right answer for the right reason — `0x11223344 >> 16 = 0x1122`. The reverse direction, the initializer path, and nested structs are all covered, and the two non-overlap controls are unchanged, so there is no false-positive cost.

### Mutation — 4 product hunks, mutated individually, 4 killed, 0 survivors

Each mutant asserted to have applied (`count == 1` before replacement), `BUILD_RC` checked before believing any result, restored and rebuilt between runs.

| # | hunk | mutation | result |
|---|---|---|---|
| M1 | `StructValue.cs:250` indexer routing | `set => owner.SetField(…)` → `set => SetRaw(…)` | **3 die** — FullOverlap, PartialOverlap, ExplicitNoOverlap |
| M2 | `StructValue.cs:37` constructor gate | `== LayoutKind.Explicit` → `Sequential` | **2 die** — FullOverlap, PartialOverlap |
| M3 | `StructValue.cs:192` alias refresh | drop `ReadFields` after a successful `TrySetField` | **3 die** |
| M4 | `StructValue.cs:103` copy semantics | copy constructed with `null, null` layout | **2 die** — FullOverlap, PartialOverlap |

Round-trip closed: restore → porcelain 0 → rebuild `BUILD_RC=0` → 5/5 green again.

### Your numbers, re-measured rather than relayed

- "`Issue3100FieldOffsetParityTests` + issue #759 tests: 10/10" — **confirmed, 10/10.**
- "full `Interpreter.Tests`: 638/638" — on the **merged** tree it is **642/642, rc=0**. Your 638 was true of the unmerged branch; `main` has moved. Worth restating so nobody cites 638 for merged code.
- The sample table in the body matches my measurement exactly.

The test itself is well built and I want to say so: five cases on the `{full, partial, none-explicit, none-default, sequential}` axis, three-column parity asserted against the emitted oracle, **distinct values 11/22/33/44** rather than an all-zero fixture that could not distinguish "correct" from "default", programs at **top level** (so bare `gsc` and `gsi` actually execute them rather than passing vacuously), each run in its own empty directory, and metadata verified with `Assembly.Load` + `GetTypes()` + `Marshal.OffsetOf` rather than `IlVerifier` alone.

### SHOULD-FIX — a reference field overlapping a primitive surfaces as `GS9999`

```
@StructLayout(LayoutKind.Explicit, Size: 8)
struct U {
    @FieldOffset(0) var Text string
    @FieldOffset(0) var Bits int64
}
var v = U{Text: "hi", Bits: 0}
```

| | `main` | merged tree |
|---|---|---|
| emit + run | rc=**134**, `TypeLoadException` | unchanged |
| bare `gsc` | rc0, prints `0` | rc=1 `GS9999` |
| `gsi` | rc0, prints `0` | rc=1 `GS9999` |

The direction is right — `main` silently printed `0`, which is the worst class, and the merged tree refuses. But the text is a raw CLR exception message surfaced through the internal-error bucket:

> `error GS9999: Could not load type 'GSharpLayout_1' from assembly 'GSharp.Interpreter.Layouts, …' because it contains an object field at offset 0 that is incorrectly aligned or overlapped by a non-object field.`

`GS9999` is for internal errors, and this leaks two synthetic implementation names (`GSharpLayout_1`, `GSharp.Interpreter.Layouts`) that mean nothing to a user. This is a reachable, well-defined user error and deserves a real diagnostic naming the struct and the offending field. Not a blocker — it is strictly better than the silent `0` it replaces, and the emitted column rejects this program too, so no correct program is affected. Worth an ID request rather than self-allocating one; this PR correctly allocates **no** new IDs (verified: 0 added descriptors, positive control 378 on `main`) and touches no docs, so the docs gate is not implicated.

### NIT — `FieldCollection` hides the indexer instead of encapsulating

`StructValue.cs:243-252` derives from `ConcurrentDictionary<string, object>` and hides the indexer with `new`. `ConcurrentDictionary`'s indexer is not virtual, so any write through a base-typed reference, or via `TryAdd` / `GetOrAdd` / `AddOrUpdate` / `TryUpdate`, bypasses `SetField` and silently desynchronises the overlap — the exact failure class this PR exists to remove.

I checked whether that is reachable today and it is not: **0** calls to those APIs on `.Fields` anywhere in `src/`, and **0** base-typed aliases of the property, against a positive control of **16** `.Fields[…] =` writes and confirmed occurrences of `.TryAdd(` / `.GetOrAdd(` elsewhere in `src/`, so the greps were live. So this is a latent trap, not a defect. Containment rather than inheritance would make it unrepresentable.

### NIT — two small invariants worth pinning

`explicitLayoutValue` (`StructValue.cs:28`) is the target of `lock` at `:96` and `:192` and is assigned only in the two constructors (`:40`, `:59`); it is not declared `readonly`, so nothing stops a future assignment from splitting the lock. The static `Cache` (`:266`) is keyed by `StructSymbol` and never evicted — correct and cheap for the process-per-run `gsc`/`gsi` model, worth remembering if the evaluator is ever hosted long-lived.

### One correction to the review brief I was given

I was told this PR should also correct a ledger row at `test/Compiler.Tests/LanguageConformance/SampleConformanceTests.cs:60-63` attributing this sample's divergence to #2986. **That row does not exist on `abfab99c`.** The file is present (244 lines) and my grep was live, but it contains **0** matches for `PInvokeStructMarshalling`, `2986`, or `3100`; repo-wide the sample is referenced only by `test/Core.Tests/Baselines/refactoring-baseline.json:90`. The ledger described lives on the #2939 branch, as issue #3100's own body says. So there is nothing here for this PR to correct, and its absence is not a SHOULD-FIX. This also means no interaction with #3077, which rewrites that file — #3105 does not touch it.

Worth noting for context: `SampleConformanceTests` compiles with `/out:` and diffs stdout against the golden (`:144`), so it is **emit-only**. The evaluate and interpret columns were structurally invisible to it, which is why this shipped green — and it is exactly why this PR's parity test, which drives all three drivers, is the right shape of coverage to add.

All numbers here are true of `c72c61c98` merged with `abfab99c`; `main` moves, so please do not cite them for different code.

PR #3107, opened 2026-08-02
Surface gsi non-void fallthrough guard

### LINKED ISSUE #3006: gsi: exhaustive enum switch statement fallthrough silently returns default
## Repro

```gsharp
import System

    switch x {
        case DateTimeKind.Unspecified { return 1 }
        case DateTimeKind.Utc { return 2 }
        case DateTimeKind.Local { return 3 }
    }
}

```

## Actual

At `main` `a6b1561f`, `gsi` exits 0 and prints `0`.

The compiled path exits 134 with:

```text
System.InvalidOperationException: Compiler-generated guard reached: non-void function fell through without returning a value.
```

The same `gsi` result occurs when one switch arm returns from an `unsafe fixed` region. That region is not executed, so the intentional compiled-only `fixed` boundary tracked by #2956 does not apply.

## Expected

`gsi` should match compiled behavior and surface the compiler-generated fallthrough guard instead of silently manufacturing `default(int32)`.

## Scope

This is the switch-**statement** counterpart to #2971. It predates and is separate from #2953: #2953 fixes CIL return-epilogue routing, while `gsi` evaluates a different representation. A `try/finally` return-funneling control already surfaces the guard through `gsi`, which suggests the direct emitter-only guard is absent from the interpreter path.

DavidObando:
## Summary
- detect non-void evaluator fallthrough at shared function boundary, including integer entry points
- thread declared return types and source declaration locations through direct calls, computed getters, and lifted operators
- share emitted/interpreted guard text from one constant
- add three-driver parity, entry-point, real-location, and exhaustive-flow regression coverage

## Validation
- merged-tree `dotnet build` (`BUILD_RC=0`, 0 warnings/errors)
- unfiltered `Interpreter.Tests` (730/730)
- unfiltered `Compiler.Tests` (4152/4152)
- 6 original product mutants plus entry-point routing and location-precedence mutants killed

Fixes #3006

### DIFF
--- a/src/Core/CodeAnalysis/Compilation/Compilation.cs
+++ b/src/Core/CodeAnalysis/Compilation/Compilation.cs
@@ -367,12 +367,8 @@ public class Compilation
         }
         catch (EvaluatorException ex)
         {
-            TextLocation location;
-            if (ex.Node?.Syntax != null)
-            {
-                location = ex.Node.Syntax.Location;
-            }
-            else
+            var location = ex.Location ?? ex.Node?.Syntax?.Location;
+            if (location == null)
             {
                 using var textWriter = new StringWriter();
                 ex.Node?.WriteTo(textWriter);
@@ -381,7 +377,7 @@ public class Compilation
             }
 
             var message = Evaluator.UnwrapDiagnosticException(ex).Message;
-            var diagnostic = new Diagnostic(location, ex.DiagnosticId, ex.Severity, message);
+            var diagnostic = new Diagnostic(location.Value, ex.DiagnosticId, ex.Severity, message);
             return new EvaluationResult(allWarnings.Add(diagnostic), null);
         }
     }
--- a/src/Core/CodeAnalysis/DiagnosticDescriptors.cs
+++ b/src/Core/CodeAnalysis/DiagnosticDescriptors.cs
@@ -8,6 +8,7 @@ internal static class DiagnosticDescriptors
 {
     internal const string AmbiguousOverloadCandidatesMessageFormat = " Candidates: {0}.";
     internal const string GoBuiltinSuggestionMessageFormat = " or call '{0}' directly";
+    internal const string NonVoidFallthroughGuardMessage = "Compiler-generated guard reached: non-void function fell through without returning a value.";
     internal const string ObsoleteUseDetailMessageFormat = ": '{0}'";
 
     internal static readonly DiagnosticDescriptor BadCharacter = new("GS0001", DiagnosticSeverity.Error, "Bad character input: '{0}'.");
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.cs
@@ -266,9 +266,9 @@ internal sealed partial class MethodBodyEmitter
         }
         else if (emitFallthroughGuard)
         {
-            const string Message = "Compiler-generated guard reached: non-void function fell through without returning a value.";
             var constructor = typeof(InvalidOperationException).GetConstructor(new[] { typeof(string) });
-            this.il.LoadString(this.outer.emitCtx.Metadata.GetOrAddUserString(Message));
+            this.il.LoadString(this.outer.emitCtx.Metadata.GetOrAddUserString(
+                DiagnosticDescriptors.NonVoidFallthroughGuardMessage));
             this.il.OpCode(ILOpCode.Newobj);
             this.il.Token(this.outer.memberRefs.GetCtorReference(constructor));
             this.il.OpCode(ILOpCode.Throw);
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -80,20 +80,12 @@ public sealed partial class Evaluator
             var statement = program.Functions.TryGetValue(node.Function, out var functionBody)
                 ? functionBody
                 : localFunctionBodies[node.Function];
-            var isIterator = IsIteratorFunction(node.Function, statement);
             object result;
             RegisterRuntimeTypeArguments(locals, node.Function.TypeParameters, node.MethodTypeArguments);
             RegisterRuntimeTypeArguments(locals, node.StaticGenericOwnerType);
             using (PushFrame(locals))
             {
-                if (isIterator)
-                {
-                    result = EvaluateIteratorFunction(node.Function, statement);
-                }
-                else
-                {
-                    result = EvaluateFunctionBody(statement);
-                }
+                result = EvaluateUserMethodBody(node.Function, statement);
             }
 
             // ADR-0060 item #7: write the final parameter slot value back to
@@ -106,16 +98,6 @@ public sealed partial class Evaluator
                     var finalValue = locals.TryGetValue(parameter, out var v) ? v : null;
                     WriteBackToOperand(operand, finalValue);
                 }
-            }
-
-            if (isIterator)
-            {
-                return result;
-            }
-
-            if (node.Function.IsAsync)
-            {
-                return WrapAsyncResult(node.Function.Type, result);
             }
 
             return result;
@@ -173,7 +155,7 @@ public sealed partial class Evaluator
             return EvaluateIteratorFunction(function, body);
         }
 
-        var result = EvaluateFunctionBody(body);
+        var result = EvaluateFunctionBody(body, function.Type, function.Declaration?.Identifier.Location);
         return function.IsAsync ? WrapAsyncResult(function.Type, result) : result;
     }
 
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -1790,7 +1790,10 @@ public sealed partial class Evaluator
             locals[node.Function.Parameters[1]] = rightValue;
             using (PushFrame(locals))
             {
-                return EvaluateFunctionBody(program.Functions[node.Function]);
+                return EvaluateFunctionBody(
+                    program.Functions[node.Function],
+                    node.Function.Type,
+                    node.Function.Declaration?.Identifier.Location);
             }
         }
 
@@ -1970,7 +1973,10 @@ public sealed partial class Evaluator
                 var frame = new ConcurrentDictionary<Symbols.VariableSymbol, object>();
                 using (PushFrame(frame))
                 {
-                    return EvaluateFunctionBody(staticGetterBody);
+                    return EvaluateFunctionBody(
+                        staticGetterBody,
+                        node.Property.GetterSymbol.Type,
+                        node.Property.Declaration?.Identifier.Location);
                 }
             }
 
@@ -2028,7 +2034,10 @@ public sealed partial class Evaluator
             };
             using (PushFrame(frame))
             {
-                return EvaluateFunctionBody(getterBody);
+                return EvaluateFunctionBody(
+                    getterBody,
+                    property.GetterSymbol.Type,
+                    property.Declaration?.Identifier.Location);
             }
         }
 
--- a/src/Core/CodeAnalysis/Evaluator.Statements.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Statements.cs
@@ -14,6 +14,7 @@ using System.Threading.Tasks;
 using GSharp.Core.CodeAnalysis.Binding;
 using GSharp.Core.CodeAnalysis.Symbols;
 using GSharp.Core.CodeAnalysis.Syntax;
+using GSharp.Core.CodeAnalysis.Text;
 using Emit = GSharp.Core.CodeAnalysis.Emit;
 
 namespace GSharp.Core.CodeAnalysis;
@@ -243,13 +244,34 @@ public sealed partial class Evaluator
     /// parked goto reaching the boundary would mean a `break`/`continue`
     /// escaped a function body, which the binder rejects — but we clear
     /// defensively so a stale flag from a partially-evaluated previous
-    /// call cannot corrupt the next invocation on this evaluator.
+    /// call cannot corrupt the next invocation on this evaluator. Issue
+    /// #3006: when a non-void function reaches this boundary without a
+    /// return, surface the same compiler-generated guard used by emitted
+    /// code instead of returning the CLR default value.
     /// </summary>
-    private object EvaluateFunctionBody(BoundBlockStatement body)
+    /// <param name="body">The lowered function body.</param>
+    /// <param name="returnType">The declared return type, or null for top-level and constructor bodies.</param>
+    /// <param name="location">The source location of the function declaration, when available.</param>
+    /// <returns>The function's return value.</returns>
+    private object EvaluateFunctionBody(
+        BoundBlockStatement body,
+        TypeSymbol returnType = null,
+        TextLocation? location = null)
     {
         var result = EvaluateStatement(body);
+        var returned = IsReturning;
         IsReturning = false;
         PendingGotoLabel = null;
+
+        if (!returned && returnType != null && returnType != TypeSymbol.Void)
+        {
+            throw EvaluatorException.CreateDiagnostic(
+                DiagnosticDescriptors.AllPathsMustReturn,
+                new InvalidOperationException(DiagnosticDescriptors.NonVoidFallthroughGuardMessage),
+                body,
+                location);
+        }
+
         return result;
     }
 
--- a/src/Core/CodeAnalysis/Evaluator.cs
+++ b/src/Core/CodeAnalysis/Evaluator.cs
@@ -203,7 +203,7 @@ public sealed partial class Evaluator
     {
         if (!useEntryPointReturnType)
         {
-            return EvaluateFunctionBody(body);
+            return EvaluateFunctionBody(body, entryPoint.Type, entryPoint.Declaration?.Identifier.Location);
         }
 
         if (entryPoint.Type != TypeSymbol.Void
@@ -213,7 +213,7 @@ public sealed partial class Evaluator
             throw new InvalidOperationException("Entry point must have a return type of void, int32, or uint32.");
         }
 
-        var value = EvaluateFunctionBody(body);
+        var value = EvaluateFunctionBody(body, entryPoint.Type, entryPoint.Declaration?.Identifier.Location);
         return entryPoint.Type == TypeSymbol.Void ? null : value;
     }
 
--- a/src/Core/CodeAnalysis/EvaluatorException.cs
+++ b/src/Core/CodeAnalysis/EvaluatorException.cs
@@ -4,6 +4,7 @@
 
 using System;
 using GSharp.Core.CodeAnalysis.Binding;
+using GSharp.Core.CodeAnalysis.Text;
 
 namespace GSharp.Core.CodeAnalysis;
 
@@ -67,6 +68,11 @@ public class EvaluatorException : Exception
     public BoundNode Node { get; }
 
     /// <summary>
+    /// Gets the source location associated with the exception, when available.
+    /// </summary>
+    internal TextLocation? Location { get; private set; }
+
+    /// <summary>
     /// Gets the stable diagnostic identifier.
     /// </summary>
     internal string DiagnosticId { get; private set; } = "GS9999";
@@ -102,16 +108,19 @@ public class EvaluatorException : Exception
     /// <param name="descriptor">The diagnostic descriptor.</param>
     /// <param name="innerException">The evaluated exception.</param>
     /// <param name="node">The bound node associated with the exception.</param>
+    /// <param name="location">The source location associated with the exception.</param>
     /// <returns>The evaluator exception.</returns>
     internal static EvaluatorException CreateDiagnostic(
         DiagnosticDescriptor descriptor,
         Exception innerException,
-        BoundNode node)
+        BoundNode node,
+        TextLocation? location = null)
     {
         return new EvaluatorException(innerException.Message, innerException, node)
         {
             DiagnosticId = descriptor.Id,
             Severity = descriptor.Severity,
+            Location = location,
         };
     }
 }
--- a/src/Core/CodeAnalysis/Lowering/Lowerer.cs
+++ b/src/Core/CodeAnalysis/Lowering/Lowerer.cs
@@ -1432,7 +1432,6 @@ public sealed class Lowerer : BoundTreeRewriter
         var preEmitAnalysisBody = Flatten(ProtectedRegionBranchRewriter.Rewrite(flattened));
         if (this.returnValueLocal != null)
         {
-            const string Message = "Compiler-generated guard reached: non-void function fell through without returning a value.";
             var exceptionType = typeof(System.InvalidOperationException);
             var constructor = exceptionType.GetConstructor(new[] { typeof(string) });
             var statements = ImmutableArray.CreateBuilder<BoundStatement>(flattened.Statements.Length + 1);
@@ -1447,7 +1446,9 @@ public sealed class Lowerer : BoundTreeRewriter
                             null,
                             exceptionType,
                             constructor,
-                            ImmutableArray.Create<BoundExpression>(new BoundLiteralExpression(null, Message)),
+                            ImmutableArray.Create<BoundExpression>(new BoundLiteralExpression(
+                                null,
+                                DiagnosticDescriptors.NonVoidFallthroughGuardMessage)),
                             TypeSymbol.FromClrType(exceptionType)),
                         DiagnosticDescriptors.AllPathsMustReturn));
                 }
--- a/src/Core/Core.csproj
+++ b/src/Core/Core.csproj
@@ -14,6 +14,8 @@
 
   <ItemGroup>
     <InternalsVisibleTo Include="GSharp.Core.Tests" />
+    <InternalsVisibleTo Include="GSharp.Compiler.Tests" />
+    <InternalsVisibleTo Include="GSharp.Interpreter.Tests" />
     <InternalsVisibleTo Include="GSharp.LanguageServer" />
   </ItemGroup>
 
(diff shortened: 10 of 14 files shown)


DavidObando:
## Rubber-duck review — PR #3107 — **NOT MERGEABLE** (1 BLOCKER)

The fix itself is correct, well-shaped and genuinely well-tested. It is blocked on one sibling call site that the refactor audited and skipped, which leaves the declared issue only partly fixed.

### Mechanics

| check | result |
|---|---|
| merge base | `6ec381c5ddc15b573dc6a8701edcb17ae4413a99` |
| true scope (three-dot, `--full-index`) | **4 files, +556 −25** |
| merge onto `main` `4038d051` | `MERGE_RC=0` |
| merged-tree `dotnet build GSharp.sln -c Release` | `BUILD_RC=0`, 0 warnings / 0 errors |
| merged-tree `Interpreter.Tests` | **716/716**, `TEST_RC=0` |
| CI | all 7 test projects present and green (`core`, `compiler-*`, `cs2gs-*`, `extensions`, `internalanalyzers`, `interpreter`, `languageserver`, `sdk`) |
| new diagnostic descriptors | **0** — `DiagnosticDescriptors.cs` has 379 on both `main` and merged. `GS0100` is pre-existing (`DiagnosticDescriptors.cs:18`, 64 commits). No self-allocated ID. Docs gate not implicated. |

`Fixes #3006` is the only issue reference and the diff contains nothing outside it — the `EvaluateUserMethodBody` extraction is in service of the fix. No undeclared scope in the outbound direction.

### Verbatim repro, run first, three drivers

Issue #3006's program byte-for-byte (sha256 `e5c17b4e2cac10e2172f30c35a05ad0ff29011b243a4080504d7fb13252bd587`), each driver in its own asserted-empty directory:

| driver | `main` `4038d051` | merged |
|---|---|---|
| bare `gsc` (evaluates) | rc=0, prints `0`, `Success.` | **rc=1, `GS0100` + guard message** |
| `gsc /out:` + run (emits) | rc=134, guard | rc=134, guard (unchanged) |
| `gsi` (interprets) | rc=0, prints `0` | **rc=1, `GS0100` + guard message** |

Confirmed fixed on all 11 declared forms.

### The source-break question — measured, and the answer is NO

The concern was that the interpreting drivers now reject at bind time what `gsc /out:` still compiles. **They do not.** The new guard is a *runtime* throw inside `EvaluateFunctionBody` (`Evaluator.Statements.cs:258-269`), reached only when the fallthrough actually executes. No binder file is touched.

The discriminating probe is an enum switch covering every named member, called with a *valid* value — a program whose fallthrough is unreachable in practice:

```gsharp
func F(x DateTimeKind) int32 { switch x { case ...Unspecified { return 11 } case ...Utc { return 22 } case ...Local { return 33 } } }
Console.WriteLine(F(DateTimeKind.Utc))
```

`main` and merged are **identical**: eval rc=0 `22`, emit+run rc=0 `22`, `gsi` rc=0 `22`. No false positive.

Controls, to prove that probe was live rather than vacuous: an `if/else` chain with no `else`, a `string` switch with no default, and a `for` loop with the return inside are each rejected by **bind-time** `GS0100` at `p.gs(2,6,2,7)` — **identically on both trees**. So the binder's static exhaustiveness rule is untouched; the only programs that can reach the new runtime guard are ones the binder already considers exhaustive.

### Mutation — 6 hunks, individually, 6 killed, 0 survivors

Every mutant is a product mutant, asserted unique-and-applied before replacement, with `BUILD_RC=0` verified each time (a failed build is a no-result, not a survival), rebuilt between runs.

|---|---|---|---|
| M1 | `Evaluator.Statements.cs:258` | `!returned` → `returned` | **250 fail** |
| M2 | `Evaluator.Calls.cs:158` | drop `function.Type` | **8 fail** — `DirectFunction`, `FixedArm`, `IndexerGetter`, `InstanceMethod`, `LambdaInvoke`, … |
| M3 | `Evaluator.Expressions.cs:1793` | drop `node.Function.Type` | **1 fail** — `LiftedNullableOperator` |
| M4 | `Evaluator.Expressions.cs:1973` | drop `node.Property.GetterSymbol.Type` | **1 fail** — `StaticPropertyGetter` |
| M5 | `Evaluator.Expressions.cs:2031` | drop `property.GetterSymbol.Type` | **1 fail** — `PropertyGetter` |
| M6 | `Evaluator.Calls.cs:88` | `EvaluateUserMethodBody(...)` → `EvaluateFunctionBody(statement)` | **25 fail** — incl. async/iterator tests |

Round-trip closed: restore → `git status --porcelain` on `src/` = 0 → rebuild `BUILD_RC=0` → **716/716**.

The three expression sites you flagged as weakest are each killed by exactly one precisely-named test — tight rather than thin. M6 also confirms the async/iterator behaviour survives moving `WrapAsyncResult` inside the frame.

Guard sites inside the evaluator: **one** (`Evaluator.Statements.cs:263`). The `try/finally` funnelling path routes through it rather than a second guard, as intended.

---

## BLOCKER

### B1 — `func Main()` entry point is the one non-void sibling caller left unguarded; the original bug survives there

`EvaluateFunctionBody` has 14 callers on the merged tree. Four received a return type; nine are genuinely void (constructors `Evaluator.Calls.cs:1717` / `Evaluator.Expressions.cs:1105`, setters `:723` `:2057` `:2106`, event handlers `:1716`, iterator sink `Evaluator.Calls.cs:135`) and are correctly left `null`. **The tenth is not void:**

- `src/Core/CodeAnalysis/Evaluator.cs:206` — `return EvaluateFunctionBody(body);`
- `src/Core/CodeAnalysis/Evaluator.cs:216` — `var value = EvaluateFunctionBody(body);`

Both are in `EvaluateEntryPointBody`, whose `entryPoint.Type` is explicitly allowed to be `int32`/`uint32` (`Evaluator.cs:209-213`). Neither passes it, so the `returnType != null` gate at `Evaluator.Statements.cs:258` is never satisfied and the entry point still manufactures a default.

Probe — a `func Main() int32` that falls through, all three drivers, merged tree:

```gsharp
import System
func Main() int32 {
    let k = DateTimeKind(99)
    switch k {
        case DateTimeKind.Unspecified { return 11 }
        case DateTimeKind.Utc { return 22 }
        case DateTimeKind.Local { return 33 }
    }
}
```

| driver | `main` | merged |
|---|---|---|
| `gsc /out:` + run (**oracle**) | rc=**134**, guard, `at Default.<Program>.Main()` | rc=**134**, guard |
| bare `gsc` | rc=0, `Success.` | **rc=0, `Success.` — unchanged** |
| `gsi` | rc=0, empty | **rc=0, empty — unchanged** |

This is #3006's exact defect class — silent wrong answer, rc0, no diagnostic — on the entry point. It is made materially worse by #3011 having just landed: `gsi` now honours `Main`'s integer exit code, which I confirmed with a control (`func Main() int32 { return 42 }` → `gsi` rc=**42**, emitted rc=**42**). So a fallthrough in `Main` makes `gsi` report **success** where the emitted program **crashes** — the exit code is now load-bearing, and it is wrong.

The PR's test file contains **0** occurrences of `func Main`, so nothing pins this.

To be fair and precise: **this is not a regression** — `main` behaves the same. It is an *incomplete fix of the issue this PR declares it closes*, on a call site the signature change itself put under review. Merging as-is closes #3006 with the shape still broken.

**Recommended fix (one argument each):** pass `entryPoint.Type` at `Evaluator.cs:206` and `:216`, and add a `func Main() int32` case to `FallthroughCases()`. Note the `!useEntryPointReturnType` branch at `:205-207` needs it too — both paths are currently unguarded.

---

## SHOULD-FIX

### S1 — the guard's diagnostic location is a constant, and does not identify the failing function

`Evaluator.Statements.cs:264-267` passes `body` — the **lowered** block — to `EvaluatorException.CreateDiagnostic`. Its span is synthesized, so the reported location is not source-derived. Measured on two files of different lengths:

| source | lines | function `F` at | reported |
|---|---|---|---|
| issue repro | 9 | 2-8 | `(1,1,17,1)` |
| padded variant | 20 | 12-18 | `(1,1,17,1)` |

Identical span for both, no filename, and the quoted source line is `    {`. Contrast the **bind-time** `GS0100` on the same driver, which correctly prints `p.gs(2,6,2,7)` and `func G(n int32) int32 {`; and the emitted driver, which correctly names `Default.<Program>.F(DateTimeKind x)`.

So the interpreting drivers now give the right error at a useless location. This matters more than usual because #3055 (evaluator-exception locations) landed hours ago. No test asserts the location — the Theory only asserts `Contains("GS0100")` and the message — so nothing would catch it regressing further.

**Recommended fix:** locate from the function declaration syntax (or the `FunctionSymbol`) rather than the lowered body, and assert the line in at least one test.

### S2 — the PR body's `650/650` is stale

Merged tree measures **716/716**. The number was true when taken; `main` has since moved (#3011, #3055, #3105 all added interpreter tests). Stale, not false — but please don't let `650` be cited for merged code.

---

## NIT

### N1 — guard message is now a fourth copy of the same literal

`const string Message = "Compiler-generated guard reached: …"` appears at `Lowering/Lowerer.cs:1435`, `Emit/MethodBodyEmitter.cs:269`, `Evaluator.Statements.cs:263`, plus `GuardMessage` in the test file. Three-driver parity is asserted on exact string equality, so a future edit to one copy silently breaks parity. A single shared internal constant would make the parity structural instead of coincidental.

---

### What I could not close
`Compiler.Tests` was not run locally (~1h20m, and machine load reached 37.31, above my build threshold). CI's `tests (compiler-*)` shards all pass; that is CI evidence, not my own measurement.

### Summary
Correct fix, no source break, one interpreter guard site, 6/6 mutants killed with precisely-named victims, and a test that asserts exit codes and both streams across all three drivers — this is good work. **B1 is the only thing standing between this and a merge**, and it is a two-line change plus one test case.

DavidObando:
Review fixes pushed in `cd7fafeb` (rebased onto current `main` `4469e323`, which contains requested `402b8561`).

### Entry-point matrix

| driver | before | after |
|---|---|---|
| bare `gsc` eval | rc0; stdout `Success.`; stderr empty | **rc1**; stdout `main.gs(3,6,3,10): error GS0100` + guard; stderr `Failed.` |
| `gsi` | rc0; stdout/stderr empty | **rc1**; stdout empty; stderr `main.gs(3,6,3,10): error GS0100` + guard |
| `gsc /out:` | compile rc0; run **rc134** with guard stderr | unchanged: compile rc0/stderr empty; run **rc134** with `InvalidOperationException` guard stderr |

Both entry-point paths now pass `entryPoint.Type` and declaration location (`Evaluator.cs:206,216`).

### Real locations

Fresh asserted-empty probes:
- short file: `short.gs(3,6,3,7): error GS0100`
- padded file: `padded.gs(12,6,12,7): error GS0100`

Location precedence is `ex.Location ?? ex.Node?.Syntax?.Location` (`Compilation.cs:370`). Guard text now has one literal, in `DiagnosticDescriptors.cs:11`; evaluator, lowerer, emitter, and tests reference it.

### New product mutants

- Removed type/location routing at `Evaluator.cs:206`: `BUILD_RC=0`; `EntryPoint` theory case failed bare `gsc` (actual rc0 vs expected rc1).
- Removed type/location routing at `Evaluator.cs:216`: `BUILD_RC=0`; same case failed `gsi` (actual rc0 vs expected rc1).
- Removed `ex.Location` precedence: `BUILD_RC=0`; both location cases failed, reverting to synthetic `(1,1,17,1)`.

All mutants asserted applied and restored. Original six mutants remain 0 survivors.

### Final validation

- `Interpreter.Tests`: **730/730** unfiltered
- `Compiler.Tests`: **4152/4152** unfiltered
- merge-tree output: `bde0543ce49aa8849989d2505328ba664ed8f6cc`
- `HEAD^{tree}`: same hash; merged tree therefore exactly tested HEAD
- merged-tree `dotnet build`: **BUILD_RC=0**, 0 warnings/errors
- artifact mtimes: Core `17:59:20`, Compiler `17:59:23`, Repl `17:59:29`
- `git diff --check`: clean

DavidObando:
## Round-2 re-review — verdict: **MERGEABLE**

All three round-1 findings (1 BLOCKER, 1 SHOULD-FIX, 1 NIT) are fixed, and each fix is pinned by a test that I proved can *see* it by mutation. Below is measurement, not a reading of the PR body.

### Anchors

| item | value |
|---|---|
| `main` | `f7d9a5d3` |
| head | `cd7fafebc0b4aa76f55b8b0e4e5dd7c96fbec3f0` |
| merge base | `4469e3237a30d2f7abc57f98620d885f64bc5d22` (branch is behind `main` by #3052, #3123) |
| true scope (three-dot, `--full-index`) | **14 files, +668 −48** |
| `git merge-tree f7d9a5d3 cd7fafeb` **output** | 1 line, bare tree SHA `29949f8f…`, no conflict block |
| real merge | `MERGE_RC=0`, merged commit `914cf32a`, porcelain empty |
| merged-tree build | **`BUILD_RC=0`, 0 Warning(s), 0 Error(s)** |
| baseline `f7d9a5d3` build | `BUILD_RC=0` |

Round-1 head `9b68283a` is **not** an ancestor of `cd7fafeb` (`--is-ancestor` rc=1) — the branch was rebased, so I re-derived everything rather than diffing head-to-head.

`Fixes #3006` is the only issue reference. **No self-allocated diagnostic ID**: the `DiagnosticDescriptors.cs` `+1` is an `internal const string NonVoidFallthroughGuardMessage`, not a descriptor; `grep -c 'new("GS'` = **380 on both** merged and `main`. `GS0100` (`AllPathsMustReturn`) is pre-existing.

---

### Round-1 **BLOCKER B1 — FIXED**

`func Main() int32` falling through was the one non-void caller left unguarded. Re-fired the identical probe, own empty dir (`entries_before=1`), compared on **rc AND stderr** (not printed value — `0` is this bug's own symptom):

| driver | round 1 | round 2 |
|---|---|---|
| bare `gsc` EVALUATES | rc **0**, `Success.` | rc **1**, `p.gs(2,6,2,10): error GS0100: Compiler-generated guard reached…`, stderr `Failed.` |
| `gsi` interprets | rc **0**, empty | rc **1**, stderr carries `p.gs(2,6,2,10): error GS0100…` |
| `gsc /out:` EMITS + run | rc **134** guard | rc **134** guard, unchanged |

`Evaluator.cs:206` and `:216` both now pass `entryPoint.Type` + `entryPoint.Declaration?.Identifier.Location`. Pinned by the `EntryPoint` case at `Issue3006FallthroughGuardDriverTests.cs:41`.

### Round-1 **SHOULD-FIX S1 (constant location) — FIXED**

Round 1 the guard location was the constant `(1,1,17,1)` on both a 9-line and a 20-line file. Verified independently on two files of different lengths **and different identifier widths**:

- 9-line file, `func F(…)` on line 2 → `p.gs(2,6,2,7)` — cols 6–7 = `F` (1 char)
- 32-line file, `func LongerName(…)` on line 13 → `p.gs(13,6,13,16)` — cols 6–16 = `LongerName` (10 chars)

Filename present, line points at the offending function, and the **span tracks the identifier width** — so this is not merely "different but wrong".

### Round-1 **SHOULD-FIX S2 (stale counts) — STALE, NOT FALSE**

Author reported Interpreter **730/730**. Measured on the merged tree: **739/739, RC=0**. The delta is #3052's interpreter tests landing after the author measured. Also ran `Core.Tests`: **7089 passed / 1 skipped / 7090 total, RC=0**, identical to baseline.

### Round-1 **NIT N1 (guard literal in 4 copies) — FIXED**

One `internal const` at `DiagnosticDescriptors.cs:11`, three uses (`MethodBodyEmitter.cs:270`, `Evaluator.Statements.cs:270`, `Lowerer.cs:1449`). Exactly **1** raw `"Compiler-generated guard reached"` literal remains in `src/` — the definition itself. Constant value is byte-identical to the removed literals, so both emit hunks are behaviourally inert.

---

### Mutation — 3/3 killed, each asserted applied, each naming its dying test

Per hunk, product mutants, condition/argument removal (never `if (true)`), `BUILD_RC` read every time so a failed build is scored as *no result*, not a survival.

| # | site | mutant | `BUILD_RC` | result | dying test |
|---|---|---|---|---|---|
| M1 | `Evaluator.cs:216` | `entryPoint.Type` → `null` | 0 | **KILLED** 1/739 | `UnmatchedExhaustiveSwitch_AllDriversSurfaceGuard(name:"EntryPoint")` |
| M2 | `Evaluator.cs:206` | `entryPoint.Type` → `null` | 0 | **KILLED** 1/739 | `UnmatchedExhaustiveSwitch_AllDriversSurfaceGuard(name:"EntryPoint")` |
| M3 | `Compilation.cs:386` | drop `ex.Location ??` | 0 | **KILLED** 2/739 | `FallthroughGuard_ReportsFunctionDeclarationLocation` (**both** parameterisations: `(0,3)` and `(9,12)`) |

Round-trip closed: restored → porcelain empty → rebuild `BUILD_RC=0`, 0 warnings → **739/739 green**.

M3 matters most: it is the site that did not exist in round 1, and it proves the location fix is observed rather than merely present.

### Root cause, not symptom — the strongest thing in this PR

The round-1 concern was that a second guard would be bolted onto the direct-call path. It was not. `Evaluator.Calls.cs:80` **deletes** the direct path's private iterator/async/`EvaluateFunctionBody` inline copy (net −18 lines) and delegates to the same `EvaluateUserMethodBody` the try/finally path already used at `:155`. Counting wide on the partial `Evaluator` class, there is exactly **one** guard site in the whole interpreter:

```
src/Core/CodeAnalysis/Evaluator.Statements.cs:270
```

Two divergent call paths became one. That is the fix the issue actually needed.

I checked the reordering that dedup implies — async wrapping moved inside `PushFrame`, and now runs before the ADR-0060 §7 write-back rather than after. Inert: `WrapAsyncResult` (`Evaluator.Calls.cs:203`) is a pure static `Task.FromResult`, and the write-back loop reads the `locals` dictionary directly, not the frame. Iterators still bypass the guard by early return in `EvaluateUserMethodBody`, which is correct — iterators legitimately fall off the end.

### False-positive corpus — the highest-value probe, 0 regressions

Passing a return type to a caller that is genuinely void would fire the guard on correct code. Every program below ran on **both** the baseline `f7d9a5d3` tree and the merged tree, on **all three drivers**, each into its own empty directory with distinct non-zero values:

| probe | shape | main | merged |
|---|---|---|---|
| F1 | `func Main()` **void**, falls off end | rc0 `11` ×3 | **identical** |
| F2 | `func Main() int32 { … return 42 }` | `gsi` rc**42**, emit rc**42** | **identical** |
| F3 | `func Main() int32`, every switch arm returns | rc**22** ×3 | **identical** |
| F4 | ctor + property setter + property **getter** + void method + iterator | `22\|44\|46\|11\|12\|13` ×3 | **identical** |
| F5 | `async` func + `try/finally` return-funnel + generic `Pick[T]` | `22\|99\|22\|33` ×3 | **identical** |

F2 confirms the #3011 exit-code path is undisturbed. F4 exercises five of the eight untyped callers plus the typed getter at `Evaluator.Expressions.cs:2037`. Void `Main` is exempt **by construction**, not by luck: the guard condition at `Evaluator.Statements.cs:266` is `!returned && returnType != null && returnType != TypeSymbol.Void`.

F4 initially failed with `GS0305` (`:=` removed, ADR-0077). That was an **unrun check**, not a pass — I corrected the syntax and re-ran rather than banking the result.

### 14-caller audit, re-run wide

Still **14** call sites, now **6 typed / 8 untyped**. **No over-application** — no genuinely void caller received a type.

- Typed, guard active: `Evaluator.Calls.cs:158`, `Evaluator.cs:206`, `Evaluator.cs:216`, `Evaluator.Expressions.cs:1793`, `:1976`, `:2037`
- Untyped, guard inactive and correctly so: `Evaluator.cs:199` (top-level), `Evaluator.Calls.cs:135` (iterator sink), `:1758` (ctor chain), `Evaluator.Expressions.cs:723`, `:1105` (ctor body), `:1716` (event add/remove), `:2066`, `:2115` (setters)

A single-line grep for the second argument reports only 3 of the 6, because `Expressions.cs:1793/1976/2037` wrap across lines. Count wide.

**Correcting my own round-1 record:** I wrote that "nine are genuinely void". That was wrong arithmetic on my part — the round-1 split was 4 typed + 2 entry-point + 1 top-level + **7** genuinely void = 14. The substance of the blocker was unaffected, but the number I published was not right.

---

### NITs (none blocking)

- **N2** — the PR describes the location site as `Compilation.cs:370`; on the rebased tree it is **`Compilation.cs:386`**. Stale line number, not a false claim, but it cost a grep to reconcile.
- **N3** — `src/Core/Core.csproj` adds `InternalsVisibleTo` for `GSharp.Compiler.Tests` and `GSharp.Interpreter.Tests` solely so tests can read one message constant. Product-assembly internal surface widened for test convenience; a `const` duplicated in the test helper would have avoided it.
- **N4** — `FallthroughGuard_ReportsFunctionDeclarationLocation` (`:439`) only ever uses a 1-character function name, so it asserts `({line},6,{line},7)` and a **column-width** regression would survive it. My `LongerName` probe (`(13,6,13,16)`) covers this, but nothing in the suite does.
- **N5** — `entryPoint.Declaration?.Identifier.Location`: a `FunctionSymbol` with a null `Declaration` yields a null location and falls back to the fabricate path at `Compilation.cs:387-393`. That degrades the location, it does not produce a wrong answer, and M3 showed the fallback still yields a usable span. Low.

### Disclosed as NOT RUN

**`Compiler.Tests` (author claims 4152/4152) was not run** — ~1h20m, and the machine carried 35 parallel agents with 1-minute load peaking at **33.66** during this review (I stopped and waited for it to fall to 18.15 before starting any mutation build). The PR does modify two `Compiler.Tests` files, so that suite is genuinely unverified by me. Everything I assert above rests on `Interpreter.Tests` 739/739, `Core.Tests` 7089/7090, the merged-tree build, and the 5-program × 3-driver × 2-tree corpus. CI covers the gap.

### Verdict

**MERGEABLE.** Blocker fixed and observed, both should-fixes fixed, the nit fixed, no source break across a deliberately adversarial false-positive corpus, and the underlying two-paths defect eliminated rather than patched.

DavidObando:
## Addendum — correcting my own coverage denominator (verdict unchanged: MERGEABLE)

Per the fleet bulletin on test-project enumeration, my round-2 verdict above **undercounted what I had left unrun**, and I am correcting it here rather than leaving it.

**What I got wrong.** I disclosed "`Compiler.Tests` was not run" as though it were the *only* gap. It was not. I ran **2** projects (`Interpreter.Tests`, `Core.Tests`) out of **9**. My working denominator was 7, taken from `test/`, which is a directory listing rather than an enumeration — the zero was **scope, not absence**, which is the exact error I have flagged in other people's work.

Enumerated properly from `GSharp.sln` (`grep -oE '[^"]*\.csproj' GSharp.sln | grep -ic test` = **9**):

```
test/Compiler.Tests            test/Core.Tests           test/Extensions.Tests
test/InternalAnalyzers.Tests   test/Interpreter.Tests    test/LanguageServer.Tests
test/Sdk.Tests                 tools/cs2gs/Cs2Gs.Tests   tools/gsgen/GSharp.GeneratorHost.Tests
```

The two outside `test/` are not incidental here. **`Cs2Gs.Tests` invokes real `gsc.dll` out-of-process**, and #3107 is a pure `src/` change touching *both* the evaluator (`Evaluator.*.cs`) and the emitter (`MethodBodyEmitter.cs`, `Lowerer.cs`). That is precisely the shape that can redden an out-of-process gate without any test project under `test/` noticing. That gap was load-bearing, not bookkeeping.

### Re-measured — 8 of 9 projects, on `main` at `5b71a74d`

#3107 merged at 06:50Z as `36b28785`; `git merge-base --is-ancestor 36b28785 5b71a74d` → rc **0**, and I verified the fix by **content** rather than by commit: `EvaluateFunctionBody(body, entryPoint.Type, …)` appears **2×** in `main`'s `Evaluator.cs` (`:206`, `:216`) and `ex.Location ??` at `Compilation.cs:392`. Fresh worktree, `BUILD_RC=0`, 0 warnings / 0 errors. Freshness verified in **epoch seconds with `sort -n`** (not clock strings): 6/6 first-party `GSharp*.dll` across `Core/`, `Compiler/` and `Repl/` (excluding `ref/`) were 94 s old; `gsi.dll` present in `Repl/`.

| project | result | RC |
|---|---|---|
| `Compiler.Tests` | **4221 / 4221** (1 h 29 m) | 0 |
| `Core.Tests` | **7106 passed, 1 skipped / 7107** | 0 |
| `Interpreter.Tests` | **1201 / 1201** | 0 |
| `Cs2Gs.Tests` | **1903 / 1903** (17 m 38 s) | 0 |
| `LanguageServer.Tests` | **462 / 462** | 0 |
| `Extensions.Tests` | **104 / 104** | 0 |
| `GeneratorHost.Tests` | **31 / 31** | 0 |
| `InternalAnalyzers.Tests` | **7 / 7** | 0 |
| `Sdk.Tests` | **NOT RUN** — needs the local `gsharp.net.sdk` package, out of bounds for me | — |

**Total: 15,036 tests, 0 failures, 1 skipped, across 8 of 9 projects.** `Sdk.Tests` is genuinely unverified by me; I am stating that plainly rather than implying coverage.

### Two of my own published numbers were stale

- I reported `Interpreter.Tests` **739/739**. On current `main` it is **1201/1201**. Main moved under me.
- The author reported `Compiler.Tests` **4152/4152**; it is now **4221/4221**. **Stale, not false** — same distinction I applied to the 730 → 739 delta.

### Behaviour re-verified on the shipped result, not the reviewed one

The earlier verdict measured a reconstructed merge. These ran against `main` as actually shipped, each in its own empty directory (`entries_before` asserted = 2), compared on **rc AND stderr**:

| probe | bare `gsc` EVALUATES | `gsi` interprets | `gsc /out:` EMITS + run |
|---|---|---|---|
| `func Main() int32` falling through | rc **1**, `p.gs(2,6,2,10): error GS0100` | rc **1**, same on stderr | rc **134**, guard at `Default.<Program>.Main()` |
| `func Main()` **void**, falls off end | rc 0, `11` | rc 0, `11` | rc 0, `11` |
| `func Main() int32 { … return 42 }` | rc 0, `11` | rc **42** | rc **42** |

The blocker fix, the location `(2,6,2,10)` pointing at `Main`, the void-`Main` exemption and the #3011 exit-code path all hold in the merged product.

### Verdict

**Unchanged: MERGEABLE.** Nothing in the wider suite contradicts the round-2 findings — the correction is to my *disclosure*, not to my conclusions. But "6 of 7" and "only Compiler.Tests is unrun" were both wrong denominators, and a reviewer who states a denominator owes it accurately.

PR #3109, opened 2026-08-02
gsc: enable JIT optimization for release builds

### LINKED ISSUE #2920: Release builds run with JIT optimizer disabled
Doing a bit of benchmarking / optimizer testing and wanted to verify my finding:

gsc emits `DebuggableAttribute(true, true)` on assemblies that have a PDB, the second arg is `isJITOptimizerDisabled`. This means Release output runs at MinOpts, so no tiered compilation, inlining, or bounds-check eliminations.

Wondering if this is intentional as my benchmark results show:

Time in ns/op. Ratio is G# / C# -- goal is to see if G# JIT compiles are ~parity with C#
| workload | base gs | base ratio | fixed gs | fixed ratio |
|---|---|---|---|---|
| closure capture | 23.0 | 2.25x | 14.4 | 1.44x |
| string interpolation | 65.4 | 2.15x | 35.7 | 1.15x |
| slice append x32 | 458.0 | 4.57x | 358.7 | 3.40x |
| iterator sum x64 | 597.8 | 3.27x | 192.4 | 1.05x |
| virtual dispatch x100 | 504.4 | 3.42x | 140.8 | 0.96x |
| struct math | 12.5 | 8.28x | 1.5 | 1.00x |
| array index sum x1000 | 2079.3 | 5.38x | 386.2 | 1.00x |
| map access x4 | 43.9 | 1.43x | 29.9 | 0.99x |
| generic invoke | 3.7 | 13.82x | 0.3 | 1.00x |

Measured on .NET 10, linux-x64, with paired G#/C# workloads:
- G# code runs 2x to 14x slower than equal C# code
- With the optimizer enabled, seven of nine workloads reach 1.0x parity
- E2E: a 10k-line G# UI framework's dirty-frame-plus-paint time drops
  from 9.9 ms to 2.7 ms with only this change plus two small emit
  cleanups. Native-bound (Skia/Yoga) parts do not move - all G#-bound parts
  move 2x to 7x

Methodology for these tests is:
- Nine paired workloads. Each pair has one G# and one C# version of the same
  algorithm. Checksums must match across the pair.
- Time: minimum of 15 Stopwatch rounds after three warmup calls.
- Allocation: `GC.GetAllocatedBytesForCurrentThread` delta over one call.
- This follows the min-of-N convention in the existing gsharp perf tests
  (`Issue1604StarChainPerfParserTests`, `InterpolatedStringTests`).

In one sentence, it seems the issue is that gsc marks every build as a Debug build for JIT with no path for Release mode, which is what provides ns/op parity between G# and C#.

Thanks!

DavidObando:
## Summary
- add `/optimize[+|-]`; emitted CLI output defaults to optimized
- forward existing SDK `$(Optimize)` signal instead of coupling optimization to PDB presence
- emit Roslyn-style `DebuggableAttribute` flags: `IgnoreSymbolStoreSequencePoints` for optimized PDB builds; `Default | DisableOptimizations` only for `/optimize-`

One `DebuggableAttribute` construction site and one caller exist; both changed.

## Verification
- `dotnet build GSharp.sln`: 0 warnings/errors
- issue tests: 4/4 compiler, 5/5 SDK
- metadata test loads assembly, calls `GetTypes()`, asserts exact `DebuggingFlags`, validates PDB documents and sequence points
- `/optimize- /debug-` still emits optimizer-disable metadata; `/debug:portable` stays usable in both modes
- driver values: bare `gsc` = 11, `gsc /out:` runtime = 11, `gsi` = 11
- quiet-window timing, min of 15 after 3 warmups: optimized 4,694,500 ticks vs MinOpts 10,887,792 ticks (2.32x), matching checksum 1,367,379,968

Full solution run passed Core 7087/7088 (1 existing skip), Compiler 4129/4129, Interpreter 633/633, SDK 59/59, and other projects. Only failure was unrelated `Issue2819_SameVersionPackedToolAndSdk_CompileTranslatedIfLet`: test requires `out/bin/Release/nupkgs`, while this required Debug build produced only Debug packages.

## Mutation
Each product mutant built clean before measurement:
- invert `Compilation.Optimize` flow: `Main_OptimizeFlag_ControlsJitFlagsIndependentlyFromPdb` failed 4/4
- invert optimized/debug flag selection: same test failed 4/4
- invert no-PDB debug guard: `/optimize-`, `/debug-` case failed
- invert CLI parsed option flow: same test failed 4/4
- invert SDK flag forwarding: `Optimization_ForwardsCompilerFlag` and default design-time test failed

Fixes #2920

### DIFF
--- a/src/Compiler/Program.cs
+++ b/src/Compiler/Program.cs
@@ -128,6 +128,7 @@ public class Program
                 {
                     ImplicitSystemImport = parsed.ImplicitSystemImport,
                     IsLibrary = parsed.Target == OutputTarget.Library,
+                    Optimize = parsed.Optimize,
                     Logger = logger,
 
                     // Issue #1929/#1953: set as early as possible (right at
@@ -765,6 +766,7 @@ public class Program
           /noimplicitimports            Disable implicit System import (alias: /no-implicit-imports).
           /nowarn:<ids>                 Suppress the given diagnostic IDs (comma/semicolon separated).
           /warnaserror[+|-][:<ids>]     Treat warnings as errors, globally or for specific IDs.
+          /optimize[+|-]                Enable/disable JIT optimization (default: enabled).
           /debug[+|-][:<value>]         Emit debug info: none, portable, full, pdbonly, embedded.
           /pdb:<file>                   Sidecar PDB path.
           /doc:<file>                   XML documentation output path.
@@ -969,6 +971,18 @@ public class Program
                             result.WarnNotAsErrorIds.Add(id);
                         }
 
+                        break;
+
+                    case "optimize":
+                        result.Optimize = ParseBoolFlag(value, defaultIfEmpty: true);
+                        break;
+
+                    case "optimize+":
+                        result.Optimize = true;
+                        break;
+
+                    case "optimize-":
+                        result.Optimize = false;
                         break;
 
                     case "debug":
@@ -1287,6 +1301,9 @@ public class Program
 
         /// <summary>Gets or sets the requested PDB emit format (from /debug, /debug:&lt;value&gt;, /debug+/-). Defaults to None.</summary>
         public DebugInformationFormat DebugFormat { get; set; } = DebugInformationFormat.None;
+
+        /// <summary>Gets or sets a value indicating whether emitted assemblies allow JIT optimization (from /optimize, /optimize+/-). Defaults to true.</summary>
+        public bool Optimize { get; set; } = true;
 
         /// <summary>Gets or sets a value indicating whether a /debug, /debug+, or /debug- switch was observed on the command line. Used so that a bare /pdb:&lt;path&gt; can default the format to Portable without overriding a later /debug-.</summary>
         public bool DebugFlagSeen { get; set; }
--- a/src/Core/CodeAnalysis/Compilation/Compilation.cs
+++ b/src/Core/CodeAnalysis/Compilation/Compilation.cs
@@ -103,6 +103,12 @@ public class Compilation
     public bool IsLibrary { get; set; } = false;
 
     /// <summary>
+    /// Gets or sets a value indicating whether emitted runtime assemblies
+    /// allow JIT optimization. Defaults to <see langword="true"/>.
+    /// </summary>
+    public bool Optimize { get; set; } = true;
+
+    /// <summary>
     /// Gets or sets the active preprocessor symbol set used by
     /// <c>[Conditional("SYMBOL")]</c> call-site elision (ADR-0047 §6 /
     /// issue #176). A call to a function marked
@@ -716,9 +722,9 @@ public class Compilation
         };
     }
 
-    private static void EmitAssembly(BoundProgram program, Stream peStream, ReferenceResolver references, string assemblyName = null, string assemblyVersion = null, bool metadataOnly = false, Lowering.Async.AsyncStateMachineRewriteResult asyncRewriteResult = null, IteratorRewriteResult iteratorRewriteResult = null, AsyncIteratorRewriteResult asyncIteratorRewriteResult = null, DebugInformationOptions debugInformation = null, Stream pdbStream = null, string targetFrameworkMoniker = null, IReadOnlyList<(string Name, byte[] Data, bool IsPublic)> embeddedResources = null)
+    private void EmitAssembly(BoundProgram program, Stream peStream, ReferenceResolver references, string assemblyName = null, string assemblyVersion = null, bool metadataOnly = false, Lowering.Async.AsyncStateMachineRewriteResult asyncRewriteResult = null, IteratorRewriteResult iteratorRewriteResult = null, AsyncIteratorRewriteResult asyncIteratorRewriteResult = null, DebugInformationOptions debugInformation = null, Stream pdbStream = null, string targetFrameworkMoniker = null, IReadOnlyList<(string Name, byte[] Data, bool IsPublic)> embeddedResources = null)
     {
-        ReflectionMetadataEmitter.Emit(program, peStream, references, assemblyName, metadataOnly, asyncRewriteResult, iteratorRewriteResult, asyncIteratorRewriteResult, debugInformation, pdbStream, assemblyVersion, targetFrameworkMoniker, embeddedResources);
+        ReflectionMetadataEmitter.Emit(program, peStream, references, assemblyName, metadataOnly, asyncRewriteResult, iteratorRewriteResult, asyncIteratorRewriteResult, debugInformation, pdbStream, assemblyVersion, targetFrameworkMoniker, embeddedResources, Optimize);
     }
 
     private void PrepareReferencesForBinding(string assemblyName)
--- a/src/Core/CodeAnalysis/Emit/AssemblyAttributeEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/AssemblyAttributeEmitter.cs
@@ -43,8 +43,8 @@ namespace GSharp.Core.CodeAnalysis.Emit;
 /// <item><c>EmitGSharpTypeSemantics</c> — the
 /// <c>AssemblyMetadataAttribute</c> type-semantics payload consumed by
 /// <see cref="ImportedAssemblySemantics"/>.</item>
-/// <item><c>EmitDebuggableAttribute</c> — <c>DebuggableAttribute(true, true)</c>
-/// when debug information is present.</item>
+/// <item><c>EmitDebuggableAttribute</c> — stamps optimized or non-optimized
+/// CLR debugging flags independently from PDB presence.</item>
 /// <item><c>EmitNullableContextAttribute</c> — assembly-level
 /// <c>NullableContextAttribute(1)</c>.</item>
 /// </list>
@@ -322,23 +322,23 @@ internal sealed class AssemblyAttributeEmitter
     }
 
     /// <summary>
-    /// Emits <c>System.Diagnostics.DebuggableAttribute(true, true)</c> when
-    /// debug information is present so managed debuggers treat the assembly as
-    /// JIT-tracked and non-optimized.
+    /// Emits the CLR debugging flags for an optimized or non-optimized runtime
+    /// assembly.
     /// </summary>
-    public void EmitDebuggableAttribute(AssemblyDefinitionHandle assemblyHandle)
+    public void EmitDebuggableAttribute(AssemblyDefinitionHandle assemblyHandle, bool optimize)
     {
         var attrType = this.emitCtx.References.TryResolveType("System.Diagnostics.DebuggableAttribute", requireExternalVisibility: false, out var resolved)
             ? resolved
             : typeof(System.Diagnostics.DebuggableAttribute);
         var attrTypeRef = this.getTypeReference(attrType);
+        var modesType = attrType.GetNestedType("DebuggingModes");
+        var modesTypeRef = this.getTypeReference(modesType);
 
         var ctorSig = new BlobBuilder();
         new BlobEncoder(ctorSig).MethodSignature(isInstanceMethod: true)
-            .Parameters(2, r => r.Void(), p =>
+            .Parameters(1, r => r.Void(), p =>
             {
-                p.AddParameter().Type().Boolean();
-                p.AddParameter().Type().Boolean();
+                p.AddParameter().Type().Type(modesTypeRef, isValueType: true);
             });
 
         var ctorRef = this.emitCtx.Metadata.AddMemberReference(
@@ -348,8 +348,11 @@ internal sealed class AssemblyAttributeEmitter
 
         var valueBlob = new BlobBuilder();
         valueBlob.WriteUInt16(0x0001); // Prolog
-        valueBlob.WriteBoolean(true);  // isJITTrackingEnabled
-        valueBlob.WriteBoolean(true);  // isJITOptimizerDisabled
+        var flags = optimize
+            ? System.Diagnostics.DebuggableAttribute.DebuggingModes.IgnoreSymbolStoreSequencePoints
+            : System.Diagnostics.DebuggableAttribute.DebuggingModes.Default
+                | System.Diagnostics.DebuggableAttribute.DebuggingModes.DisableOptimizations;
+        valueBlob.WriteInt32((int)flags);
         valueBlob.WriteUInt16(0);      // NumNamed
 
         this.emitCtx.Metadata.AddCustomAttribute(
--- a/src/Core/CodeAnalysis/Emit/EmitContext.cs
+++ b/src/Core/CodeAnalysis/Emit/EmitContext.cs
@@ -97,6 +97,12 @@ internal sealed class EmitContext
     public bool MetadataOnly { get; }
 
     /// <summary>
+    /// Gets or sets a value indicating whether the emitted runtime assembly
+    /// allows JIT optimization.
+    /// </summary>
+    public bool Optimize { get; set; } = true;
+
+    /// <summary>
     /// Gets the ECMA-335 metadata builder used to assemble the PE's metadata
     /// tables for this emit.
     /// </summary>
--- a/src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs
@@ -790,6 +790,7 @@ internal sealed class ReflectionMetadataEmitter
     /// <c>TargetFrameworkAttribute</c>.
     /// </param>
     /// <param name="embeddedResources">Managed resources to embed in runtime assemblies.</param>
+    /// <param name="optimize">Whether the emitted runtime assembly allows JIT optimization.</param>
     public static void Emit(
         BoundProgram program,
         Stream peStream,
@@ -803,11 +804,13 @@ internal sealed class ReflectionMetadataEmitter
         Stream pdbStream = null,
         string assemblyVersion = null,
         string targetFrameworkMoniker = null,
-        IReadOnlyList<(string Name, byte[] Data, bool IsPublic)> embeddedResources = null)
+        IReadOnlyList<(string Name, byte[] Data, bool IsPublic)> embeddedResources = null,
+        bool optimize = true)
     {
         var emitter = new ReflectionMetadataEmitter(program, references, assemblyName, metadataOnly, embeddedResources);
         emitter.emitCtx.AssemblyVersionOverride = assemblyVersion;
         emitter.emitCtx.TargetFrameworkMoniker = targetFrameworkMoniker;
+        emitter.emitCtx.Optimize = optimize;
 
         emitter.emitCtx.DebugInformation = debugInformation ?? new DebugInformationOptions();
         emitter.emitCtx.PdbStream = pdbStream;
@@ -3902,9 +3905,9 @@ internal sealed class ReflectionMetadataEmitter
 
         // Phase 7.7b: emit cross-language interop attributes for NuGet consumability.
         this.assemblyAttrs.EmitAssemblyInteropAttributes(assemblyHandle);
-        if (!this.emitCtx.MetadataOnly && this.emitCtx.Pdb != null)
+        if (!this.emitCtx.MetadataOnly && (this.emitCtx.Pdb != null || !this.emitCtx.Optimize))
         {
-            this.assemblyAttrs.EmitDebuggableAttribute(assemblyHandle);
+            this.assemblyAttrs.EmitDebuggableAttribute(assemblyHandle, this.emitCtx.Optimize);
         }
 
         return mvidFixup;
--- a/src/Sdk/Gsharp.NET.Sdk/BuildTask.cs
+++ b/src/Sdk/Gsharp.NET.Sdk/BuildTask.cs
@@ -157,6 +157,8 @@ public class BuildTask : Microsoft.Build.Utilities.Task, ICancelableTask
             args.Add($"/targetframework:{this.TargetFramework}");
         }
 
+        args.Add(ParseBool(this.Optimization) ? "/optimize+" : "/optimize-");
+
         if (!string.IsNullOrEmpty(this.DebugType))
         {
             args.Add($"/debug:{this.DebugType}");
--- a/test/Compiler.Tests/Issue2920ReleaseJitOptimizerTests.cs
+++ b/test/Compiler.Tests/Issue2920ReleaseJitOptimizerTests.cs
@@ -1,0 +1,96 @@
+// <copyright file="Issue2920ReleaseJitOptimizerTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Metadata;
+using Xunit;
+
+namespace GSharp.Compiler.Tests;
+
+public class Issue2920ReleaseJitOptimizerTests
+{
+    [Theory]
+    [InlineData(
+        null,
+        "/debug:portable",
+        true,
+        (int)DebuggableAttribute.DebuggingModes.IgnoreSymbolStoreSequencePoints)]
+    [InlineData(
+        "/optimize+",
+        "/debug:portable",
+        true,
+        (int)DebuggableAttribute.DebuggingModes.IgnoreSymbolStoreSequencePoints)]
+    [InlineData(
+        "/optimize-",
+        "/debug:portable",
+        true,
+        (int)(DebuggableAttribute.DebuggingModes.Default | DebuggableAttribute.DebuggingModes.DisableOptimizations))]
+    [InlineData(
+        "/optimize-",
+        "/debug-",
+        false,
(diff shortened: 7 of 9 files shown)


DavidObando:
# Rubber-duck review — PR #3109 (`fix/2920-release-jit-optimizer`)

**Verdict: MERGEABLE** — 0 BLOCKER, 1 SHOULD-FIX, 2 NIT.

Anchors: `main` = `4038d051`, PR head = `aa3fa95a`, computed merge base = `6ec381c5`, **merged tree built at `8c74723f` → `BUILD_RC=0`, 0 warnings**. Base `4038d051` also `BUILD_RC=0`. All probes read-only; no repository was modified.

---

## The most important thing I found: most ways of validating this PR are vacuous

My first 122-sample corpus run came back **identical on both trees**, and that result was **worthless**. I positive-controlled it and found **0 PDBs emitted** — samples compiled without `/debug` emit no PDB, and on *base* the attribute was only emitted when a PDB existed. So in that configuration **both trees emit no `DebuggableAttribute` at all** and base was *already* optimized.

**The bug only ever manifests when a PDB is emitted.** That is the SDK path (`DebugType=portable` in both Debug *and* Release), which is why the reporter saw it in Release builds. Any reviewer — or any future regression check — that validates this PR without `/debug:portable` is measuring nothing and will report a false "no change".

I re-ran the entire corpus with `/debug:portable` on both trees (119 PDBs). Everything below is from the non-vacuous run.

---

## Evidence

### 1. Full sample corpus, non-vacuously (the highest-value probe)

All 122 `samples/*.gs`, emitted with `/optimize` default + `/debug:portable`, each into its own asserted-empty directory, on base `4038d051` vs merged `8c74723f`:

**122/122 identical** — same compile rc, same run rc, same output md5 — with **base running at MinOpts and merged running under a fully optimizing JIT**. (3 compile failures — `GsharpExtensionsMixed`, `Optional`, `Sequences` — occur identically on both trees and are pre-existing.)

This is the probe that matters, because nothing in this repo has ever run through an optimizing JIT. The corpus covers the categories that optimization most commonly breaks: **10 async/await samples, 2 iterator samples, 18 struct samples, 6 floating-point samples, 7 goroutine samples**. No divergence in any of them.

### 2. Metadata, verified independently of the PR's own test

Own probe assembly (`Assembly.Load(File.ReadAllBytes(...))` + `GetTypes()` + read `DebuggableAttribute.DebuggingFlags`):

| tree / config | flags | `IsJITOptimizerDisabled` |
|---|---|---|
| base + PDB | `257` (`Default \| DisableOptimizations`) | **True** |
| merged + PDB | `2` (`IgnoreSymbolStoreSequencePoints`) | **False** |
| merged, no PDB | attribute **ABSENT** | — |

### 2b. The "Roslyn-style flags" claim, checked against real `csc`

I built a C# library with `DebugType=portable` in both configurations and read the flags with the same probe:

| producer | flags | `IsJITOptimizerDisabled` |
|---|---|---|
| `csc` Release | `2` (`IgnoreSymbolStoreSequencePoints`) | False |
| `csc` Debug | `263` (`Default \| IgnoreSymbolStoreSequencePoints \| EnableEditAndContinue \| DisableOptimizations`) | True |
| **this PR, optimized** | **`2`** | **False** |
| this PR, `/optimize-` | `257` (`Default \| DisableOptimizations`) | True |

**The optimized case matches `csc` Release exactly.** The claim is empirically true where it matters.

The `/optimize-` case differs from `csc` Debug by `IgnoreSymbolStoreSequencePoints | EnableEditAndContinue` — but `257` is **precisely what `gsc` emitted before this PR** (measured on base). So the PR leaves the non-optimized configuration bit-for-bit unchanged and only fixes the optimized one. That is the correct minimal scope; the debug-mode divergence from Roslyn is pre-existing and not this PR's to chase. (`EnableEditAndContinue` would be wrong to claim anyway — `gsc` has no EnC.)

### 3. Debugging experience survives optimization — tested end-to-end, not asserted

The author claims "sequence points preserved". Rather than trust the claim or its test, I checked the actual user-visible outcome: an exception's stack trace under `/optimize+ /debug:portable`.

```
=== optimize+ ===
  at GSharp.Probe.Thrower.<Program>.Boom(Int32 n) in /tmp/p09/seq/thrower.gs:line 6
  at GSharp.Probe.Thrower.<Program>.Main()     in /tmp/p09/seq/thrower.gs:line 10
=== optimize- ===
  at GSharp.Probe.Thrower.<Program>.Boom(Int32 n) in /tmp/p09/seq/thrower.gs:line 6
  at GSharp.Probe.Thrower.<Program>.Main()     in /tmp/p09/seq/thrower.gs:line 10
```

**Byte-identical file and line mapping in both modes, with correct line numbers.** The perf win does not cost debuggability.

### 4. Timing

Two `bench.dll` builds from the *same* merged `gsc`, differing only in the flag, each flag-verified (`2` vs `257`) before timing. Min-of-9, quiet window, load 23.7 → 29.6 across the measurement:

- `/optimize+` → **9 ms**
- `/optimize-` → **54 ms**
- **6.0×**, with **identical checksums** (`31426560000`) — so the optimized build is doing the same work, not less of it.

### 5. CLI surface

Each into its own asserted-empty directory, all with `/debug:portable`:

| invocation | flags |
|---|---|
| (omitted — default) | `2` |
| `/optimize+`, `/optimize`, `/optimize:true` | `2` |
| `/optimize-`, `/optimize:false` | `257` |
| `/optimize:foo` | rc1, `Unsupported boolean value: foo` |

**The default is optimized**, i.e. the default flipped — which is the entire point of the fix and is stated in the PR. Help text lists it at `src/Compiler/Program.cs:769` as `/optimize[+|-] … (default: enabled)`. Malformed input reuses the existing house helper `ParseBoolFlag` (`src/Compiler/Program.cs:1183-1196`), so it is consistent with `/deterministic` and `/implicitimports`.

### 6. SDK forwarding

`src/Sdk/Gsharp.NET.Sdk/BuildTask.cs:160` adds `args.Add(ParseBool(this.Optimization) ? "/optimize+" : "/optimize-")`. `Optimization` (`BuildTask.cs:63`) was *already* wired to `$(Optimize)` via `Gsharp.NET.Core.Sdk.targets:375` and `Gsharp.NET.Sdk.Bootstrap.targets:121`, so this hunk consumes an existing signal rather than inventing one.

`ParseBool("")` returns false (`BuildTask.cs:468`), which would be wrong for an unset property — but `Sdk.props` imports `Microsoft.NET.Sdk`, and `Microsoft.NET.Sdk.props:49/52` (verified against the installed `/usr/local/share/dotnet/sdk/10.0.300`) defines `Optimize=false` for Debug and `true` for Release. **Debug therefore still gets `DisableOptimizations` and a usable debugger; Release gets optimized.** Correct.

Caveat, stated plainly: no local e2e is available to me, so the SDK path is verified by property-wiring inspection plus the SDK unit tests, not by an actual SDK build.

### 7. "One unrelated failure" — verified, not accepted

`Issue2819_SameVersionPackedToolAndSdk_CompileTranslatedIfLet` (`tools/cs2gs/Cs2Gs.Tests/Adr0151IfLetExpressionTranslationTests.cs:307`) requires `out/bin/Release/nupkgs`. **On the merged tree, with that prerequisite present, it PASSES (1/1).** So it is an environmental prerequisite, not a PR effect — a stronger result than the author's "it's unrelated".

### 8. Counted wide (partial classes undercount under a filtered grep)

Exactly **one** definition (`AssemblyAttributeEmitter.cs:328`), **one** call site (`ReflectionMetadataEmitter.cs:3910`), **one** production caller of `ReflectionMetadataEmitter.Emit(` (`Compilation.cs:762`). Three further callers are tests using the default `optimize: true`. The PR's "one construction site and one caller" claim holds.

### 9. Mutation — per hunk, 5/5 killed

Each mutant is a **product** mutant (condition/value inversion, never `if (true)`), each asserted to have actually landed (harness exits 9 if the pattern count ≠ 1), each reverted to drift = 0:

| # | hunk | mutant | result |
|---|---|---|---|
| M1 | `AssemblyAttributeEmitter.cs` | invert flag selection | 4/4 failed |
| M2 | `ReflectionMetadataEmitter.cs` | invert emission condition `\|\|`→`&&` | 3/4 failed |
| M3 | `Program.cs` | CLI default `true`→`false` | 1/4 failed |
| M4 | `Program.cs` | drop `Compilation` wiring | 2/4 failed |
| M5 | `BuildTask.cs` | invert SDK forwarding | 2/2 failed (baseline 2/2 passed) |

### 10. Suites on the merged tree `8c74723f`

| suite | result |
|---|---|
| Core.Tests | **7088 passed / 0 failed / 1 skipped** (`RegenerateBaseline`, pre-existing) |
| Interpreter.Tests | **703 / 0** |
| Sdk.Tests | **59 / 0** |
| Compiler.Tests | **4152 / 0** (1 h 18 m, `SUITE_RC=0`) |

**Why my totals differ from the PR's** (the PR reports Interpreter 633, Compiler 4129, Core 7087/7088): the PR measured at its own base `6ec381c5`; I measured the merged tree on `4038d051`, which has since absorbed #3011, #3055, #3105, #3106 and #3108. Interpreter +70 (633→703) is consistent with #3055's interpreter-diagnostic work; Compiler +23 (4129→4152) likewise. **This is a population difference, not a regression** — zero failures on both sides.

### 11. Scope claim, checked rather than assumed

`gsc /out:` is the affected column, but "bare `gsc` and `gsi` are untouched" is *slightly* imprecise — see NIT-2. The driver matrix: bare `gsc` → `11`, `gsc /out:` → `11`, `gsi` → `11`. `gsi` correctly rejects `/optimize-` as an unrecognized argument, since `gsi` never emits.

`Fixes #2920` is present and names the correct issue.

---

## Findings

### SHOULD-FIX — the new switch is missing from the CLI reference

`website/docs/tooling/gsc.md` is an **exhaustive** per-switch table: it documents `/debug`, `/debug+`, `/debug-`, `/debug:none`, `/debug:portable`, `/deterministic`, `/deterministic+`, `/deterministic-`, `/deterministic:true|false`, `/embed`, `/embed+`, and so on down to no-op flags like `/lib:`. **`/optimize` appears nowhere in `docs/` or `website/docs/`.**

A new public CLI surface — one that additionally **flips the default behaviour of every emitted assembly** — should not be discoverable only via `--help`. There is no docs-gate test on `gsc.md`, so nothing will catch this.

Fix: add rows to the table at `website/docs/tooling/gsc.md:57-96`, in the established style, covering `/optimize`, `/optimize+`, `/optimize-`, `/optimize:true|false`, and stating that the default is **enabled**.

### NIT-1 — the test matrix is missing one quadrant

`test/Compiler.Tests/Issue2920ReleaseJitOptimizerTests.cs:17-37` has four `[InlineData]` cells: `(default, /debug:portable)`, `(/optimize+, /debug:portable)`, `(/optimize-, /debug:portable)`, `(/optimize-, /debug-)`. The missing quadrant is **`/optimize+` + `/debug-`**, which is the only configuration asserting the attribute is **absent** entirely.

Not a defect — I verified it externally (attribute **ABSENT**, and the no-PDB DLLs are byte-identical to base across the whole corpus). It is a one-line `[InlineData]` addition that would complete the 2×2.

### NIT-2 — "`gsc /out:` is the only affected column" is imprecise

`EmitAssembly` (`src/Core/CodeAnalysis/Compilation/Compilation.cs:760`) is the single funnel for **both** the `/out:` emit path and the in-memory evaluate path used by bare `gsc`. The new emission condition at `ReflectionMetadataEmitter.cs:3908` is `(!MetadataOnly && (Pdb != null || !Optimize))`, so under bare `gsc /optimize-` the in-memory assembly now carries a `DebuggableAttribute` where base emitted none.

This is **harmless and arguably correct** — `/optimize-` disabling the JIT optimizer for evaluated code is coherent — and at the default (`Optimize=true`, `Pdb==null`) bare `gsc` is byte-identical to base. Verified behaviourally: bare `gsc`, `gsc /optimize+`, `gsc /optimize-` all print `11` at rc0. Only the PR's wording needs a word of nuance, not the code.

---

## Disclosures

- The merged-tree build began at load **44.24** and the base build at load **33.16**, both above my own 30 bar. Both returned `BUILD_RC=0` with 0 warnings and all binaries verified mtime-fresh in `Core/`, `Compiler/` and `Repl/` (excluding `ref/`), so the correctness results are unaffected. **The timing numbers in §4 were taken separately in a genuine quiet window (load 23.7 → 29.6)** and are the only load-sensitive result here.
- No local e2e (`~/.nuget/packages/gsharp.net.sdk/` untouched), so §6 is wiring inspection plus SDK unit tests, not a real SDK build.
- Every probe compiled into its own directory, asserted empty beforehand.

DavidObando:
# Re-anchored verdict — PR #3109 (`fix/2920-release-jit-optimizer`)

Supersedes my earlier review (#issuecomment-5160845904), which was measured at `main` = `4038d051`. **`main` has since moved to `f7d9a5d3`** (#3052, #3123), and that move touched **two of this PR's six production files** — so the whole measurement was redone at the new anchor rather than carried over.

**Verdict: MERGEABLE** — 0 BLOCKER, 1 SHOULD-FIX, 2 NIT.

Anchors, all fetched and verified rather than accepted:
- `origin/main` = `f7d9a5d34a35abae…` ✔ matches
- PR head = `aa3fa95a3809797…` ✔ unchanged since my first pass
- merged head built = **`aaf4a00c`**, `dotnet build GSharp.sln -c Release` → **`BUILD_RC=0`, 0 warnings**
- base `f7d9a5d3` built separately → `BUILD_RC=0`, 0 warnings
- `git merge-tree --write-tree f7d9a5d3 aa3fa95a` → rc 0 **and** output is a single tree SHA with **0 conflict lines** (I read the output, not the rc). Still treated as a floor; both trees were built.

`out/` was **deleted entirely** before the merged build (`0` entries asserted), because for an optimizer-flag change a directory serving a prior assembly of the same identity reads exactly like "no difference". All three drivers verified mtime-fresh: `Core/GSharp.Core.dll` 21:52:14, `Compiler/gsc.dll` 21:52:16, `Repl/gsi.dll` 21:52:20 — `Repl/` included because `gsi` lives there.

---

## The two questions you asked answered explicitly

### 1. `DebuggableAttribute` flags read off the produced assembly, BOTH directions

Read two independent ways — `Assembly.Load` + `GetTypes()` + reflect the attribute, **and** a raw `System.Reflection.Metadata` PE reader that needs no dependency resolution. Both agree. Same source, own GUID probe directory asserted empty before each emit.

| tree | configuration | `DebuggableAttribute` | JIT optimizer |
|---|---|---|---|
| base `f7d9a5d3` | any PDB build (no `/optimize` switch exists) | `257` (`Default \| DisableOptimizations`) | **DISABLED** |
| merged `aaf4a00c` | Release / `/optimize+` | **`2`** (`IgnoreSymbolStoreSequencePoints`) | **ENABLED** |
| merged `aaf4a00c` | Debug / `/optimize-` | `257` (`Default \| DisableOptimizations`) | **DISABLED** |

**Both columns are correct.** Release turns the optimizer on; Debug still has it off. And critically, **the Debug value is unchanged from base** — `257` is exactly what `gsc` emitted before this PR, so the PR does not perturb the debugging configuration at all. It only fixes the optimized column.

On base the `/optimize` switch does not exist (`Unrecognized option: /optimize+`), and the only reachable behaviour is `257`. That is the bug, reproduced at the current anchor: **every** PDB-emitting build — which under the SDK is Debug *and* Release — shipped with the JIT optimizer disabled.

Debugging is not merely "flagged as enabled", it works: an exception through an optimized (`/optimize+`) assembly still resolves file and line —
```
at …Boom(Int32 n) in thrower.gs:line 6
at …Main()       in thrower.gs:line 10
```
byte-identical to the `/optimize-` trace.

### 2. User-program emit, or only how the compiler itself is built?

**User-program emit. Unambiguously.** Every `gsc /out:` output changes, so the false-positive surface is the entire language — which is why the corpus diff below is the load-bearing evidence and not a formality.

- The PR touches **no** build-configuration file: no `.csproj`, `.props`, `.targets`, `Directory.Build.*`, `global.json` or `.sln`. All six production files are the emit path (`Emit/`, `Compilation/`), the CLI (`Program.cs`), and the SDK task (`BuildTask.cs`).
- The compiler's **own** binaries are byte-for-byte unaffected: `gsc.dll`, `gsi.dll` and `GSharp.Core.dll` all read `2` on **both** base and merged. They are produced by MSBuild/`csc`, not by `gsc`, and `csc` Release already emits `2` (I confirmed that independently by building a throwaway C# library: `csc` Release = `2`, `csc` Debug = `263`).

---

## Behavioural equivalence — the probe that matters, re-run at the new anchor

All 122 `samples/*.gs`, each emitted **with `/debug:portable`** into its own directory asserted empty, then executed, on base `f7d9a5d3` vs merged `aaf4a00c`.

**Non-vacuity, proven before trusting the result:** 119 assemblies emitted a PDB on both trees, and the flags actually moved —

| tree | flag histogram across all 119 emitted assemblies |
|---|---|
| base | **119 × `257`** (optimizer disabled) |
| merged | **119 × `2`** (optimizer enabled) |

**Every single sample flipped from MinOpts to a fully optimizing JIT.** And the behavioural diff over compile rc, run rc, **stdout md5 and stderr md5** is **empty — 122/122 identical**.

This matters because my *first* pass nearly shipped a worthless version of this probe: samples compiled without `/debug` emit no PDB, and base only emitted the attribute when a PDB existed, so **both trees emit nothing and the diff is vacuously identical**. Anyone re-validating this PR without `/debug:portable` is measuring air. Flagging it because the same trap will catch the next reviewer.

Corpus coverage of the categories optimization actually breaks: 10 async/await, 2 iterator, 18 struct, 6 floating-point, 7 goroutine samples. No divergence in any.

---

## Interaction with the two PRs that landed under me

`main` moving `4038d051` → `f7d9a5d3` (#3052, #3123) touched **two of this PR's six production files**:
- `src/Core/CodeAnalysis/Compilation/Compilation.cs`
- `src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs`

`git merge` auto-resolved both. Since a clean auto-merge is exactly where the bad ones hide, I verified both sides survived by two-way diff:
- `git diff f7d9a5d3 HEAD -- <file>` shows **exactly** the PR's hunks, nothing more, nothing lost.
- `git diff aa3fa95a HEAD -- <file>` shows main's hunks intact (67 and 7 changed lines respectively).

Combined with `BUILD_RC=0` and the artifact reading `2`/`257` correctly on the merged tree, the interaction is sound.

---

## Driver matrix — three columns, both trees

`func Main`-bearing program (#3011 means all three columns genuinely execute now, so none of these are vacuous). Distinct non-zero value naming the arm — `3*7+11 = 32`, never all-zero:

| column | base | merged |
|---|---|---|
| bare `gsc` (EVALUATES) | `32`, rc 0 | `32`, rc 0 |
| `gsc /out:` (EMITS) | `32`, rc 0 | `32`, rc 0 |
| `gsi` (INTERPRETS) | `32`, rc 0 | `32`, rc 0 |

---

## Diagnostics

The PR adds and references **no diagnostic IDs whatsoever** (`git diff | grep '^\+.*GS[0-9]{4}'` → empty). No self-allocation; the `GS9100` concern does not arise.

---

## Test projects — all 7 enumerated, none filtered

| project | result |
|---|---|
| `test/Core.Tests` | **7089 / 0**, 1 skipped (`RegenerateBaseline`, pre-existing) |
| `test/Compiler.Tests` | **4161 / 0** (1 h 12 m, `SUITE_RC=0`) |
| `test/Interpreter.Tests` | **723 / 0** |
| `test/Sdk.Tests` | **59 / 0** |
| `test/Extensions.Tests` | **104 / 0** |
| `test/InternalAnalyzers.Tests` | **7 / 0** |
| `test/LanguageServer.Tests` | **462 / 0** |

Also present but **NOT RUN**, disclosed rather than implied: `tools/cs2gs/Cs2Gs.Tests`, `tools/gsgen/GSharp.GeneratorHost.Tests`, and the two `tools/cs2gs/corpus/L*-Library.Tests` fixtures. (In my earlier pass I did run the one `Cs2Gs.Tests` case the author called out — see below.)

Totals differ from the PR body (Interpreter 633, Compiler 4129, Core 7087/7088) because the PR measured at its own base; Interpreter is now 723 after #3052 added deinitializer tests. **Population difference, not regression** — zero failures either side.

---

## Mutation — per hunk, re-run at the new anchor, 5/5 killed

Every mutant is a **product** mutant (value/condition inversion — never `if (true)`, which would trip CS0162 under warnings-as-errors and yield a no-result). The harness `exit 9`s unless the target pattern occurs exactly once, so each mutant is **asserted to have landed**; each was reverted to `drift=0`. Unmutated baselines pass first (4/4 compiler, 2/2 SDK).

| # | hunk (`file:line`) | mutant | dying test | result |
|---|---|---|---|---|
| M1 | `Emit/AssemblyAttributeEmitter.cs:351` | invert optimized/non-optimized flag selection | `Main_OptimizeFlag_ControlsJitFlagsIndependentlyFromPdb` | **4/4 failed** |
| M2 | `Emit/ReflectionMetadataEmitter.cs:3908` | invert emission condition `\|\|` → `&&` | same | **3/4 failed** |
| M3 | `Compiler/Program.cs:1306` | CLI default `true` → `false` | same | **1/4 failed** |
| M4 | `Compiler/Program.cs:131` | drop `Compilation.Optimize` wiring | same | **2/4 failed** |
| M5 | `Sdk/Gsharp.NET.Sdk/BuildTask.cs:160` | invert SDK flag forwarding | `Optimization_ForwardsCompilerFlag` | **2/2 failed** (baseline 2/2 passed) |

**A no-result I caught rather than banked:** M5's first attempt at this anchor returned rc 1 with *no* result line. That was not a kill — it was `error NETSDK1004: Assets file … not found`, a restore gap in the fresh worktree, i.e. the tests never ran. I restored, rebuilt (`BUILD_RC=0`), re-established the 2/2 baseline, and only then re-ran the mutant. Reporting it because a silenced build error reads exactly like a passing suite.



## "One unrelated failure" — verified, not accepted

`Issue2819_SameVersionPackedToolAndSdk_CompileTranslatedIfLet` (`tools/cs2gs/Cs2Gs.Tests/Adr0151IfLetExpressionTranslationTests.cs:307`) requires `out/bin/Release/nupkgs`. With that prerequisite present it **passes 1/1** on the merged tree. Environmental prerequisite, not a PR effect.

---

## Timing (secondary — the metadata and corpus results above are the real evidence)

Two builds from the **same** merged `gsc`, differing only in the flag, each flag-verified (`2` vs `257`) before timing. Min of 11, quiet window, load 11 → 19.5 across the run:

- `/optimize+` → **5 ms**
- `/optimize-` → **34 ms**
- **6.8×**, with **identical checksums** (`31426560000`) — the optimized build does the same work, not less of it.

---

## Findings

### SHOULD-FIX — new public CLI switch is absent from the CLI reference

`website/docs/tooling/gsc.md:57-96` is an **exhaustive** per-switch table — it documents `/debug`, `/debug+`, `/debug-`, `/debug:none|portable|full|pdbonly|embedded`, `/deterministic[+|-|:true|:false]`, `/embed`, even the no-op `/lib:`. **`/optimize` appears nowhere in `docs/` or `website/docs/`.**

This switch does more than add an option: it **flips the default JIT behaviour of every assembly `gsc` emits**. That belongs in the reference, not only in `--help`. There is no docs-gate test on `gsc.md`, so nothing will catch the omission.

Fix: add rows for `/optimize`, `/optimize+`, `/optimize-`, `/optimize:true|false`, stating the default is **enabled**.

### NIT-1 — test matrix missing one quadrant

`test/Compiler.Tests/Issue2920ReleaseJitOptimizerTests.cs:17-37` covers `(default, /debug:portable)`, `(/optimize+, /debug:portable)`, `(/optimize-, /debug:portable)`, `(/optimize-, /debug-)`. Missing: **`/optimize+` + `/debug-`** — the only configuration asserting the attribute is *absent* entirely. Not a defect; I verified it externally (ABSENT). One `[InlineData]` line completes the 2×2.

### NIT-2 — "`gsc /out:` is the only affected column" is imprecise

`src/Core/CodeAnalysis/Compilation/Compilation.cs:760` `EmitAssembly` is the single funnel for **both** `/out:` emit and the in-memory evaluate path behind bare `gsc`. With the new condition at `ReflectionMetadataEmitter.cs:3908` — `(!MetadataOnly && (Pdb != null || !Optimize))` — bare `gsc /optimize-` now emits a `DebuggableAttribute` where base emitted none. Harmless and arguably correct, and at the default bare `gsc` is identical to base (verified: `32`, rc 0 on both trees). Wording nuance only.

---

## Disclosures

- Merged build began at load 20.3, base build at 18.4 — both under the bar. The earlier first-pass builds (at the superseded anchor) began at load 44.2 and 33.2; that pass is superseded, and every number in *this* comment comes from the re-anchored builds.
- No local e2e (`~/.nuget/packages/gsharp.net.sdk/` untouched), so the SDK conclusion rests on property-wiring inspection (`BuildTask.cs:63,160,468`; `$(Optimize)` from `Microsoft.NET.Sdk.props:49/52`, Debug `false` / Release `true`) plus `Sdk.Tests`, not a real SDK build.
- Every probe compiled into its own directory, asserted empty beforehand; probe root is a fresh GUID directory. No worktree I do not own was written to; `i2920` untouched. Both scratch trees end at `git status --porcelain` empty.

PR #3112, opened 2026-08-02
Fix imported nested generic type arguments

### LINKED ISSUE #2947: GS0156 displays erased nested-enum generic arguments

DavidObando:
## Summary
- preserve symbolic enclosing type arguments while resolving imported nested CLR types
- type imported enum literals with their symbolic containing enum instead of the CLR-erased `object` construction
- cover nested enum, struct, and class shapes across argument kinds, assembly boundaries, and two nesting depths

## Behavior change
Cross-assembly references such as `glib.Outer[T].Color.Green` now bind as `glib.Outer[T].Color`, not `glib.Outer[object].Color`. Bare `gsc`, emitted `gsc /out:`, and `gsi` now render the same symbolic type in diagnostics. Explicit enum-to-integer conversion remains required.

## Validation
- `Core.Tests`: 7089 passed, 1 skipped
- `Interpreter.Tests`: 638 passed
- full solution build: 0 warnings, 0 errors
- 36-site behavior matrix plus 34-site false-positive corpus
- both product condition mutants fail `ImportedNestedEnumLiterals_RetainSymbolicOuterArgumentAtBothDepths`; restored code passes

Fixes #2947

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.Accessor.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.Accessor.cs
@@ -688,14 +688,15 @@ internal sealed partial class ExpressionBinder
             return false;
         }
 
-        nestedClassSymbol = CloseImportedNestedType(importedBase, nestedClassSymbol, syntax.RightPart);
+        nestedClassSymbol = CloseImportedNestedType(importedBase, nestedClassSymbol, syntax.RightPart, symbolicOuter: null);
         return true;
     }
 
     private ImportedClassSymbol CloseImportedNestedType(
         Type constructedOuter,
         ImportedClassSymbol nested,
-        ExpressionSyntax syntax)
+        ExpressionSyntax syntax,
+        ImportedTypeSymbol symbolicOuter)
     {
         var nestedType = nested?.ClassType;
         if (constructedOuter?.IsConstructedGenericType != true
@@ -713,7 +714,17 @@ internal sealed partial class ExpressionBinder
         try
         {
             var closedType = nestedType.MakeGenericType(outerArguments);
-            return new ImportedClassSymbol(closedType, syntax, references: scope.References);
+            ImportedTypeSymbol symbolicNested = null;
+            if (symbolicOuter?.HasTypeParameterArgument == true
+                && symbolicOuter.TypeArguments.Length == nestedType.GetGenericArguments().Length)
+            {
+                symbolicNested = ImportedTypeSymbol.GetConstructed(
+                    closedType,
+                    nestedType,
+                    symbolicOuter.TypeArguments);
+            }
+
+            return new ImportedClassSymbol(closedType, syntax, symbolicNested, scope.References);
         }
         catch (ArgumentException)
         {
@@ -1937,7 +1948,7 @@ internal sealed partial class ExpressionBinder
                     var importedBaseSymbol = new ImportedClassSymbol(importedBase, nested.LeftPart, references: scope.References);
                     if (TryResolveNestedTypeFromAccessorLeft(importedBaseSymbol, nested.LeftPart, out var importedNested))
                     {
-                        importedNested = CloseImportedNestedType(importedBase, importedNested, nested.LeftPart);
+                        importedNested = CloseImportedNestedType(importedBase, importedNested, nested.LeftPart, symbolicOuter: null);
                         return BindAccessorStep(receiver: null, importedNested, nested.RightPart);
                     }
                 }
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs
@@ -259,15 +259,6 @@ internal sealed partial class ExpressionBinder
                         return BindExtensionMethodGroupOrError(receiver, ne);
                     }
 
-                    // Stream B: static field/property read on imported type.
-                    // `Receiver == null` flags the access as static. Literal
-                    // (const) fields aren't real runtime fields, so we inline
-                    // their constant value rather than emit `ldsfld`.
-                    if (staticMember is FieldInfo litField && litField.IsLiteral)
-                    {
-                        return new BoundLiteralExpression(null, litField.GetRawConstantValue(), TypeSymbol.FromClrType(litField.FieldType));
-                    }
-
                     var staticType = staticMember switch
                     {
                         PropertyInfo sp => ClrNullability.GetPropertyTypeSymbol(sp),
@@ -293,7 +284,22 @@ internal sealed partial class ExpressionBinder
                             staticType = symbolicMemberType;
                         }
 
+                        if (staticMember is FieldInfo symbolicLiteral && symbolicLiteral.IsLiteral)
+                        {
+                            var literalType = classSymbol.ClassType.IsEnum
+                                ? classSymbol.SymbolicReceiver
+                                : staticType;
+                            return new BoundLiteralExpression(null, symbolicLiteral.GetRawConstantValue(), literalType);
+                        }
+
                         return new BoundClrPropertyAccessExpression(null, null, staticMember, staticType, classSymbol.SymbolicReceiver);
+                    }
+
+                    // Literal (const) fields aren't real runtime fields, so
+                    // inline their constant value rather than emit `ldsfld`.
+                    if (staticMember is FieldInfo literal && literal.IsLiteral)
+                    {
+                        return new BoundLiteralExpression(null, literal.GetRawConstantValue(), staticType);
                     }
 
                     return new BoundClrPropertyAccessExpression(null, null, staticMember, staticType);
@@ -826,7 +832,11 @@ internal sealed partial class ExpressionBinder
             if (scope.References.TryResolveNestedType(classSymbol.ClassType, name, out var nestedType))
             {
                 var nested = new ImportedClassSymbol(nestedType, nameExpr, references: scope.References);
-                nestedClassSymbol = CloseImportedNestedType(classSymbol.ClassType, nested, nameExpr);
+                nestedClassSymbol = CloseImportedNestedType(
+                    classSymbol.ClassType,
+                    nested,
+                    nameExpr,
+                    classSymbol.SymbolicReceiver);
                 return true;
             }
 
--- a/test/Core.Tests/CodeAnalysis/Binding/Issue2947ImportedNestedGenericTypeTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Binding/Issue2947ImportedNestedGenericTypeTests.cs
@@ -1,0 +1,162 @@
+// <copyright file="Issue2947ImportedNestedGenericTypeTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using GSharp.Core.CodeAnalysis.Binding;
+using GSharp.Core.CodeAnalysis.Compilation;
+using GSharp.Core.CodeAnalysis.Symbols;
+using GSharp.Core.CodeAnalysis.Symbols.Display;
+using GSharp.Core.CodeAnalysis.Syntax;
+using GSharp.Core.CodeAnalysis.Text;
+using Xunit;
+
+namespace GSharp.Core.Tests.CodeAnalysis.Binding;
+
+/// <summary>Issue #2947: imported nested types retain symbolic enclosing arguments.</summary>
+public class Issue2947ImportedNestedGenericTypeTests
+{
+    [Fact]
+    public void ImportedNestedEnumLiterals_RetainSymbolicOuterArgumentAtBothDepths()
+    {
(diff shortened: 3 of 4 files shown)


DavidObando:
Rubber-duck review — **MERGEABLE**, with one SHOULD-FIX and a duplicate determination against #3097.

`main` = `4038d051`, head = `a37d1b18e1cd36498ee8cfbe3aef8afd40bf1f21`, merge-base `c1163493` (shared with #3097). True scope via `git diff --full-index $MB $HEAD`: **4 files, +556/−14** — production is only `ExpressionBinder.Access.Accessor.cs` (+15/−4) and `ExpressionBinder.Access.MemberLookup.cs` (+20/−10). Merge onto `main` clean; `dotnet build GSharp.sln -c Release`: **BUILD_RC=0, 0 errors / 0 warnings**. Drivers fresh by raw `stat` (PRE=1785704320; `Core`/`Compiler`/`Repl` `GSharp.Core.dll` all 1785704328), `ref/` excluded.

## The reported shape, reproduced first and byte-for-byte

Cross-assembly through `/r:`, `glib` built by `gsc /out:` exactly as in the issue:

```
BASE   rc=1  reported.gs(4,37,4,62): error GS0156: Cannot convert type
              glib.Outer[object].Color to glib.Outer[T].Color. ...
MERGED rc=0  (no diagnostic)
```

That is #2947 verbatim on `main`, and gone on the merged tree. Not an adjacent shape.

## The layer question: this is a binder fix, not a renderer paper-over

This was the thing to get right, and the PR gets it right on three independent kinds of evidence:

1. **The diagnostic does not get reworded — it disappears.** The conversion now *succeeds*. A `SymbolDisplay` change could only have changed the string; it could not have changed conversion classification.
2. **The emitted program runs and produces the right value.** `gsc /out:` + `dotnet` on the executed variant → **rc=0, stdout `5`** (`Green = 5`). The binder produced a correctly-typed *and* correctly-valued literal.
3. **The PR asserts the bound symbol directly.** `test/Core.Tests/CodeAnalysis/Binding/Issue2947ImportedNestedGenericTypeTests.cs:106-111` walks the bound tree with a `BoundTreeWalker`, then asserts `Assert.IsType<ImportedTypeSymbol>(literal.Type)`, `Assert.True(imported.HasTypeParameterArgument)`, and `Assert.Equal("T", Assert.Single(imported.TypeArguments).Name)`. That pins the binders type, not the message — exactly the evidence that was missing when this issue was closed the first time.

Mechanically, `CloseImportedNestedType` now threads a `symbolicOuter` and builds a `symbolicNested` via `ImportedTypeSymbol.GetConstructed` (`Accessor.cs:695-727`), and the nested-type call site passes `classSymbol.SymbolicReceiver` (`MemberLookup.cs:832-839`). The erased `object` never reaches the symbol.

## Axis, including the depth the maintainer flagged

9 probes, cross-assembly, `{enum, struct, class} × {symbolic [T], concrete [int32], concrete [string]} × {depth-2 Outer.X, depth-3 Outer.Mid.X}`:

| tree | pass | fail |
|---|---|---|
| BASE | 7/9 | `ax_d2enum`, `ax_d3enum` |
| MERGED | **9/9** | — |

Base failures were `glib2.Outer[object].Color` and `glib2.Outer[object].Mid.Tone`. **Depth-3 is fixed, so there is no one-level-up walk bug** — the fix reaches the declaring type. The bug was specific to nested *enums* over a symbolic outer; nested struct/class and all concrete arguments already bound correctly on `main`, which is why the failing population is 2 and not 9.

The authors own test library (`Issue2947ImportedNestedGenericDriverTests.cs:21-32`) independently covers depth-3 (`Outer[T].Mid.Tone`) plus nested struct and class.

## The neighbouring non-bug is preserved

`var i int32 = c` still errors, and now reads correctly:

```
error GS0156: Cannot convert type glib.Outer[T].Color to int32.
              An explicit conversion exists (are you missing a cast?)
```

Explicit enum-to-integer conversion is still required. The PR did not "fix" language design, and it pins this at `Issue2947ImportedNestedGenericDriverTests.cs:81`.

## Driver columns

- **`gsc /out:` EMITS** — fixed; emitted assembly runs to rc=0 / `5`.
- **bare `gsc` EVALUATES** — cross-assembly `/r:` is `rc=1, GS9999: Cannot invoke a method on objects loaded by a MetadataLoadContext`. **Positive-controlled: identical on BASE** with a concrete (already-binding) construction, so this is a pre-existing evaluate-column limitation, **not** a regression from this PR. Same-assembly evaluates fine on both trees (`5`).
- **`gsi` interprets** — no `/r:` option, so the cross-assembly path is genuinely unreachable for `gsi`. The same-assembly twin gives rc=0 / `5` on both base and merged.

No diagnostic ID is added; the fix removes a false diagnostic rather than introducing one.

## SHOULD-FIX — the non-symbolic literal path is changed wider than #2947 needs

`MemberLookup.cs` deletes the Stream-B literal block at `:259-267` and re-adds it at `:296-301`, changing the type it yields from `TypeSymbol.FromClrType(litField.FieldType)` to `staticType`. Those are not the same function: `staticType` is `ClrNullability.GetFieldTypeSymbol(sf)` followed by `NormalizeImportedSemanticAggregate` (`MemberLookup.cs:229-235`). So **every imported `const` field** — not just nested-enum members — now takes a nullability-aware, aggregate-normalised type.

I probed this and found **no observable difference**: `int32.MaxValue`, `int32.MinValue` and `Math.PI` render identically (`2147483647`, `-2147483648`, `3.141592653589793`) on base, this PR, and #3097. **Population: 3 const fields, all non-nullable value types.** That population does not include a nullable-annotated `const string`, which is where `ClrNullability` could plausibly diverge from `FromClrType`, so I cannot call it clean — only unfalsified. Either narrow the change so the non-symbolic path keeps `TypeSymbol.FromClrType`, or add a `const string` case to pin the new behaviour deliberately.

## Duplicate determination: #3097 vs #3112 — competing, not duplicate, and mutually exclusive

Same merge-base `c1163493`; **different `src` subtree hashes** (`01e00f82` vs `b4ead423`, against `main` `fb4e23ab`), so unlike #3103/#3011 these are *not* the same commit resubmitted.

They touch **the same two production files** and **hard-conflict**. Reading `merge-tree`s output rather than its rc:

```
co_merge_rc=1
CONFLICT (content): src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.Accessor.cs
CONFLICT (content): src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs
```

Each merges onto `main` alone cleanly. **Only one can land.**

They are the same idea implemented twice: both add a symbolic-receiver parameter to `CloseImportedNestedType`, both construct `symbolicNested` via `ImportedTypeSymbol.GetConstructed`, and both change the `MemberLookup.cs:826` call site to pass `classSymbol.SymbolicReceiver`. They differ in how the literals type is resolved — #3112 special-cases `classSymbol.ClassType.IsEnum`; #3097 keeps the original block and adds an `AreSame(openMemberType, symbolicReceiver.OpenDefinition)` short-circuit inside `ResolveStaticMemberTypeFromSymbolicReceiver`.

**Behaviourally I could not tell them apart.** On #3097s built tree: the verbatim reported shape is rc=0, and the same 9-probe axis is **9/9**, identical to #3112. Const fields identical. I also hypothesised that #3097s weaker guard (it omits #3112s `TypeArguments.Length == nestedType.GetGenericArguments().Length` check) would break a nested generic carrying its own type parameter — I built `Outer[T].Inner[U]` and consumed it symbolically, and **the probe refuted the hypothesis**: rc=0 on base, #3112 and #3097 alike. I am reporting that as a killed hypothesis, not as a finding.

**Keep #3112, close #3097.** The discriminator is not behaviour, it is evidence: #3112 ships `Issue2947ImportedNestedGenericTypeTests.cs`, a Core.Tests binding-level test asserting the *bound symbol* carries `T`; #3097 has no binding-level test at all (3 driver tests in one Interpreter.Tests file). Given that this issue was closed once on renderer-level evidence and had to be reopened, the fix that pins the binders type is the one worth keeping.

Two honest counterweights: #3097 is the smaller change (22/5 production lines vs 35/14) and it leaves the non-symbolic const path completely untouched, which is exactly my SHOULD-FIX above; and its `AreSame` short-circuit is arguably the more general statement of the rule than #3112s `IsEnum` special case. If you prefer the smaller diff, the best combination is #3097 plus #3112s Core.Tests binding assertion ported onto it.

## Gaps in my evidence

- **No mutation testing.** Load reached **50.74** partway through, so I stopped building. The two production hunks are **unmutated by me** — unverified, not a pass.
- **No test suite was run**, for the same reason. All results above come from the three trees built while load was ~10.7.
- **I caught one of my own runs being noise.** My first 18-probe axis returned a uniform `rc=1` on both trees, including a case I had already proved fixed. The cause was my own `>/dev/null` on the library build: `glib2.gs` had failed with `GS0368` (bodyless interface method needs `;`), so `/r:` pointed at a file that did not exist and every probe failed for the same irrelevant reason. Fixed the library and re-ran; the corrected run is the 7/9 vs 9/9 reported above. A silenced error is an unrun check, and a uniform result across trees that should differ is the tell.

DavidObando:
Addendum — correcting my own review above (`#issuecomment-5160388020`). **Verdict is unchanged: MERGEABLE.** The substance of the SHOULD-FIX is unchanged and now re-verified. What was wrong is my `file:line` citations, and they were wrong by exactly the failure mode in the bulletin: **I read line numbers out of the ambient `main` checkout, which is neither the tree this PR modifies nor the tree it produces.**

**What I cited vs what is true.** Re-measured tree-explicitly with `git show <sha>:<path>`, which cannot be fooled by a working tree:

| claim | I wrote | actually |
|---|---|---|
| original literal block | `MemberLookup.cs:259-267` | **merge base `c1163493` `:266-269`** |
| relocated literal block | `:296-301` | **head `a37d1b18e` `:300-302`** |
| `staticType` definition | `:229-235` | **base `:271-277` / head `:262-268`** |

The `:229-235` figure is the worst of the three: it came from a **third tree** — the ambient checkout at whatever `main` commit it held — so it referenced neither side of the diff. At the merge base, `:229-235` is an unrelated incomplete-member-access bail-out (`return new BoundErrorExpression(null)` for `x.` mid-typing).

**The finding itself survives re-measurement, verbatim.** At merge base `c1163493:266-269`:

```csharp
if (staticMember is FieldInfo litField && litField.IsLiteral)
{
    return new BoundLiteralExpression(null, litField.GetRawConstantValue(), TypeSymbol.FromClrType(litField.FieldType));
}
```

At head `a37d1b18e:300-302` the same path now returns `staticType`, and `staticType` is still exactly what I said it was — base `:271-277`, head `:262-268`:

```csharp
var staticType = staticMember switch
{
    PropertyInfo sp => ClrNullability.GetPropertyTypeSymbol(sp),
    FieldInfo sf => ClrNullability.GetFieldTypeSymbol(sf),
    _ => TypeSymbol.Error,
};
staticType = NormalizeImportedSemanticAggregate(staticType, staticMember);
```

So every imported `const` field still takes a nullability-aware, aggregate-normalised type where it previously took `TypeSymbol.FromClrType`. The SHOULD-FIX stands, as does its stated limit: **probe population was 3 const fields** (`int32.MaxValue`, `int32.MinValue`, `Math.PI`), all non-nullable value types, identical across base / #3112 / #3097 — unfalsified, not clean, because a nullable-annotated `const string` was not among them.

**Why this matters beyond three numbers.** My behavioural evidence — the verbatim repro, the 7/9 → 9/9 axis, the executed `5`, the const probes — was all produced against **built worktrees at `4038d051`**, so none of it is affected. Only the citations came from reading, and reading is what was stale. That is the bulletins point in miniature: where reading and measuring disagreed, measuring was right, and I should have sourced the line numbers from the same trees I was running.

**Denominator check, since the bulletin asked.** I audited all eight verdicts I have posted this session for a test-project denominator and found none — the two matches in my earlier grep were someone elses `6 of 7 projects` on #3054 (`#issuecomment-5164658529`, already self-corrected to 8 of 9; mine there is `#issuecomment-5157401159`) and a `6 of 7 P/Invoke samples` count on #3024, which is a samples count and was correct. I verified that audit was live rather than empty-by-scope by first confirming the query returned my comment bodies and matched `Rubber-duck review`.

For this PR the honest number is unchanged and worth restating against the correct denominator: **0 of the 9 test projects in `GSharp.sln` were run**, and both production hunks are **unmutated**, because load reached 50.74 and I stopped building. Unverified, not a pass. (Independently confirming the bulletin: `GSharp.sln` lists **9** test projects — 7 under `test/`, plus `tools/cs2gs/Cs2Gs.Tests` and `tools/gsgen/GSharp.GeneratorHost.Tests` — while `ls -d test/*/` returns **8**, a different and wrong answer.)

PR #3122, opened 2026-08-03
Fix explicit-layout residuals

### LINKED ISSUE #3115: Explicit-layout follow-ups: GS0518 for reference-over-primitive overlap, and FieldCollection bypass trap
## Summary

Follow-up to #3100 / PR #3105, which fixed interpretation of `@FieldOffset` overlap by synthesising one CLR type per struct symbol and letting the runtime do the layout. Two residuals were measured during review and deliberately left out of that PR.

## 1. Reference-typed field overlapping a primitive leaks an internal type name through `GS9999`

A `string` at `@FieldOffset(0)` overlapping an `int64` at `@FieldOffset(0)`:

| tree | result |
|---|---|
| before #3105 | prints `0`, **rc0** — silent wrong answer |
| after #3105 | **rc1**, `GS9999: Could not load type 'GSharpLayout_1' from assembly 'GSharp.Interpreter.Layouts'` |

The emit column also fails here (rc134, `TypeLoadException`), so **no correct program is affected** and the post-#3105 behaviour is strictly better than the silent `0`. But `GS9999` is the internal-compiler-error bucket and this is a **reachable user error**, and the message leaks the synthesised type and assembly names, which mean nothing to a user.

**Use `GS0518`**, which I have reserved for this. Do not self-allocate; I am the sole allocator for diagnostic IDs.

Wording should name the real constraint — an explicit-layout struct may not overlap a reference-typed field with a non-reference field, because the CLR itself rejects the resulting layout — and it should carry a real file, line and column, not the fabricated line-1 span that `Compilation.cs` produced before #3055.

## 2. `FieldCollection` hides a non-virtual indexer — latent, currently unreachable

`FieldCollection` (`src/Core/CodeAnalysis/.../StructValue.cs:243-252`) declares `new` over `ConcurrentDictionary`'s **non-virtual** indexer. Any access through a base-typed alias, or via `TryAdd` / `GetOrAdd` / `AddOrUpdate`, bypasses `SetField` and **silently desynchronises the explicit-layout backing store** — which is precisely the failure class #3100 was about.

Measured **unreachable today**: 0 such calls on `.Fields`, 0 base-typed aliases, against positive controls of 16 `.Fields[…]=` write sites and confirmed `.TryAdd(` / `.GetOrAdd(` usage elsewhere in `src/`. So this is not a live bug — it is a trap laid for the next person.

The lazy fix is **containment over inheritance**: hold the dictionary as a private field and expose only the operations that maintain the invariant, so the compiler makes the bypass impossible rather than merely unused.

Also minor, same file: `explicitLayoutValue` (`:28`) is the `lock` target at `:96` and `:192` and is assigned only in constructors, but is not `readonly`.

## Acceptance criteria

- The reference-over-primitive overlap reports `GS0518` with a real location and a user-meaningful message; no `GSharpLayout_*` or `GSharp.Interpreter.Layouts` name appears in user-facing output.
- Non-overlapping explicit-layout structs and sequential-layout structs are unaffected — build a corpus of programs that run correctly today and must still run.
- All three drivers agree: bare `gsc` EVALUATES, `gsc /out:` EMITS, `gsi` interprets.
- `FieldCollection` can no longer be bypassed, or a test pins the invariant.

DavidObando:
## Summary
- report source-located `GS0518` when explicit layout misaligns a reference field or overlaps reference/non-reference storage
- validate after all type bodies bind, including genuinely nested and forward-declared value-type storage
- reuse the interpreter's synthesized CLR layout type for exact user-struct storage sizing
- contain `StructValue.Fields`, making dictionary bypasses impossible, and make the layout lock target readonly

## Behavior changes
- bare `gsc`, emitted `gsc`, and `gsi` now reject CLR-invalid reference layouts before synthesis; overlap, nested, forward, misaligned-with-primitive, and standalone-misaligned probes expose no `GSharpLayout_*`/`GSharp.Interpreter.Layouts` names
- valid explicit/sequential layouts, reference-over-reference overlap, and primitive unions remain unchanged
- **Intentional public API break:** public `StructValue.FieldCollection` no longer derives from `ConcurrentDictionary`; out-of-repo uses of dictionary aliases, `Count`/`Keys`, or `TryAdd`-style bypass APIs no longer compile. This enforces explicit-layout storage synchronization by construction.

## Validation
- real locations on different file lengths: `(6,25,6,29)` and `(30,25,30,29)` in all three drivers
- exact three-driver corpus: sample `3/4/6153737367135073092`; halves `287454020/1432778632`; partial overlap `4386`; nested `1432778632/287454020`; ref-over-ref `bb`
- `Assembly.Load` + `GetTypes()` + `Marshal.OffsetOf` metadata checks pass
- issue tests: 13/13; Core.Tests: 7089 passed, 1 skipped; Interpreter.Tests: 800/800
- M3 (`.Concat(nestedStructShells.Values)` removal) builds with rc0 and is killed by `NestedStructOverlapReportsGs0518AcrossDrivers`
- alignment predicate inversion builds with rc0 and is killed by both `MisalignedReferenceReportsGs0518WithRealLocationAcrossDrivers` cases
- live-main `27a05798` merge tree equals head tree (`4ea34c0c…`), no conflict output; serialized merged-tree build `BUILD_RC=0`, 0 warnings/errors, fresh Core/Compiler/Repl artifacts

Fixes #3115


DavidObando:
## Rubber-duck review — PR #3122 — **NOT MERGEABLE** (1 BLOCKER)

Disclosure: I reviewed #3105, whose SHOULD-FIX and NIT produced #3115. I re-ran that review's probes from scratch against this head rather than trusting my prior verdict.

### Mechanics

| check | result |
|---|---|
| merge base | `4469e3237a30d2f7abc57f98620d885f64bc5d22` — **equals `main`**; `--is-ancestor` rc=0, so head already contains `main` and the merged tree **is** the head |
| `git merge-tree` **output** | one tree SHA `eeb37c0c…`, 1 line, no conflict blocks |
| true scope (three-dot, `--full-index`) | **9 files, +516 −9** |
| `dotnet build GSharp.sln -c Release` | `BUILD_RC=0`, 0 warnings / 0 errors |
| `Interpreter.Tests` | **719/719**, rc=0 |
| `Core.Tests` | **7089 passed / 1 skipped / 7090**, rc=0 |
| docs gate | filter matched **4** tests, all passed (incl. `Every_Diagnostic_Is_Documented_In_Both_References_With_Matching_Severity`) — not a zero-from-zero |
| `Compiler.Tests` | **NOT RUN** (~1h20m). CI is green; that is CI evidence, not mine. |

**Diagnostic ID — clean.** `git log -S GS0518 --all` → **1** commit, this PR's own `5232fc7f` (positive control `GS0513` → 10). `GS0519`/`GS0520` → 0. The ID is used exactly as allocated, and docs rows exist in **both** `docs/diagnostics.md:1172` and `website/docs/ref/diagnostics.md:1778` in un-backticked `| GS0518 |` form, visible to the gate regex.

### The four claims, checked

**1. `GS0518` at the overlapping non-reference field — VERIFIED, and the location is real.** Two source files of different lengths:

| file | lines | non-reference field | reported |
|---|---|---|---|
| `P1.gs` | 10 | `Raw` on line 9, cols 25-27 | `p.gs(9,25,9,28)` |
| `L2.gs` | 33 | `Bits` on line 30, cols 25-28 | `p.gs(30,25,30,29)` |

Both point exactly at the non-reference field identifier, with filename, and the quoted source line is the right one. This is **not** the fabricated `(1,1,1,N)` pattern. For the same program `main` produced `GS9999: Could not load type 'GSharpLayout_1' from assembly 'GSharp.Interpreter.Layouts…'` at `(1,1,1,23)`; head produces `GS0518` at a real span. Clear improvement, confirmed on all three drivers.

**2. All `Fields` writes routed through contained storage — VERIFIED BY CONSTRUCTION.** `StructValue.cs:235` no longer reads `: ConcurrentDictionary<string, object>`; it is `: IEnumerable<KeyValuePair<string, object>>` with a private `fields` field (`:237`). The bypass is now *impossible*, not merely unused — `TryAdd`/`GetOrAdd`/`AddOrUpdate` and base-typed aliasing do not typecheck, and `BUILD_RC=0` across the solution is itself the proof that every use site fits the narrowed surface. Counted wide on the partial class: **16** `.Fields[…] =` write sites (identical to my #3105 baseline), **0** bypass APIs, **0** base-typed aliases, with a positive control showing the write-site pattern does match. `SetRaw` is `internal` and has **0** call sites outside `StructValue.cs`.

**3. `readonly` lock target — VERIFIED** at `StructValue.cs:28`.

**4. Emit oracle — VERIFIED, not inherited.** For the invalid layout, `main` emits rc=0 then the assembly throws `System.TypeLoadException` at rc=**134**. So no correct program is affected by rejecting it at bind time.

### False-positive corpus — clean

This is a **bind-time** diagnostic (`Binder.cs:984` → `DeclarationBinder.Structs.cs:24` → `StructLayoutBinder.ReportReferenceFieldOverlaps`), so it gates the **emit** column too and the false-positive stakes are maximal. Every case below is byte-identical between `main` and head on **all three drivers**, asserting exact values:

| probe | shape | value | main vs head |
|---|---|---|---|
| R1 | `samples/PInvokeStructMarshalling.gs` verbatim | `3` / `4` / **`6153737367135073092`** (`0x5566778811223344`) | identical |
| R2 | write `QuadPart`, read halves | **`287454020`** / **`1432778632`** | identical |
| R3 | partial overlap `int32@0`+`int32@2` | **`4386`** (`0x11223344 >> 16`) | identical |
| R5 | explicit layout, no overlap | `11` / `22` | identical |
| F1 | reference `@0` + `int32@8`, non-overlapping | `ok` / `33` | identical |
| F2 | reference-over-reference overlap | `bb`, rc=0 | identical |
| F3 | nested **struct** over `int64` | **`1432778632`** / **`287454020`** | identical |

The two #3105 cases you asked me to re-run (R3 `4386`, and the exact `0x5566778811223344` pattern) both still hold. **No regression on the fix I approved, and no correct program newly rejected.**

F2 is worth calling out: the CLR *accepts* overlapping two same-type references (rc=0 on the emit column), and the rule correctly skips that pair via `leftIsReference == rightIsReference` at `StructLayoutBinder.cs:139` — so this is not an over-broad rule, and it is not a false negative either.

### Mutation — 4 hunks, individually, **3 killed, 1 SURVIVOR**

Each mutant is a product mutant, asserted unique-and-applied before replacement, with `BUILD_RC` read each time.

|---|---|---|---|
| M1 | `StructLayoutBinder.cs:155` | overlap interval `<` → `>` | **killed** — 3 fail, incl. both `ReferenceAndPrimitiveOverlapReportsGs0518AcrossDrivers` theory cases (offsets 0 and 4) |
| M2 | `StructLayoutBinder.cs:139` | pairing gate `==` → `!=` | **killed** — 7 fail, incl. `FieldLayout_EmitEvaluateAndGsiAgree(FullOverlap)` and `(PartialOverlap)`, so **over-firing is pinned by tests**, not just by my probes |
| M3 | `DeclarationBinder.Structs.cs:28` | drop `.Concat(nestedStructShells.Values)` | **SURVIVED** |
| M4 | `StructValue.cs:251` | `set => owner.SetField(…)` → `set => fields[key] = value` | **killed** — 4 fail, incl. `ExplicitNoOverlap`, `FullOverlap`, `PartialOverlap`, `WorkingReferenceLayoutsRemainValidAcrossDrivers` |

Round-trip closed: restore → `git status --porcelain -- src/` = 0 → rebuild `BUILD_RC=0`, 0W/0E → **719/719**.

---

## BLOCKER

### B1 — the nested-struct sweep is live, load-bearing behaviour pinned by **zero** tests

`DeclarationBinder.Structs.cs:28`:

```csharp
foreach (var structSymbol in topLevelStructs.Concat(nestedStructShells.Values).Distinct())
```

The `.Concat(nestedStructShells.Values)` clause is **not dead code** — I proved it works, then proved nothing observes it.

*It works.* A nested struct with a reference/primitive overlap, on clean head, all three drivers:

```gsharp
struct N1Outer {
    var Tag int32

    @StructLayout(LayoutKind.Explicit, Size: 8)
    struct N1Inner {
        @FieldOffset(0) var Name string
        @FieldOffset(0) var Raw int64
    }
}
var v = N1Outer{Tag: 11}
Console.WriteLine(v.Tag)
```

→ bare `gsc`, `gsc /out:` and `gsi` all give rc=1, `p.gs(12,29,12,32): error GS0518: … type 'N1Inner' … field 'Name' … field 'Raw'`.

*Nothing observes it.* With M3 applied (`BUILD_RC=0`, so this is a real survival and not a no-result), the very same program becomes **rc=0, prints `11`, layout silently accepted** — the exact silent acceptance this PR exists to prevent — while:

- `Interpreter.Tests` → **719/719**, unchanged from baseline
- `Core.Tests` → **7089/7090**, unchanged from baseline

**7,808 tests, zero of them notice.** `ForwardDeclaredValueTypeOverlapReportsGs0518AcrossDrivers` (`Issue3100FieldOffsetParityTests.cs:79`) looks like it would cover this, but its `Bad` and `Payload` are both **top-level** structs — it exercises forward *type-reference* resolution, not `nestedStructShells`.

This is the standing bar exactly: behaviour verified by hand, pinned by no test.

**Recommended fix:** add a nested-struct case to the GS0518 driver assertions (the `N1` program above works verbatim). One `[Fact]` calling the existing `AssertGs0518AcrossDrivers` helper kills M3.

---

## SHOULD-FIX

### S1 — a **misaligned** reference with no overlap still leaks `GSharpLayout_1` / `GSharp.Interpreter.Layouts` at a fabricated line-1 location

The CLR rejects an explicit layout when an object field is *either* incorrectly aligned *or* overlapped by a non-object field — its own message says so in one sentence. `StructLayoutBinder.ReportReferenceFieldOverlaps` implements only the overlap half.

```gsharp
@StructLayout(LayoutKind.Explicit, Size: 16)
struct M1S {
    @FieldOffset(4) var Name string   // misaligned, overlaps nothing
    @FieldOffset(12) var Id int32
}
```

| driver | `main` | head |
|---|---|---|
| bare `gsc` | rc=1 `GS9999 … 'GSharpLayout_1' … 'GSharp.Interpreter.Layouts' …` at `(1,1,1,24)` | **identical** |
| `gsi` | same | **identical** |
| `gsc /out:` + run | rc=0 emit, then rc=**134** `TypeLoadException` | **identical** |

So after this PR, `GSharpLayout_*` and `GSharp.Interpreter.Layouts` **do** still reach user-facing output, `GS9999` is still used for a designed refusal, and the location is the `(1,1,1,N)` always-line-1 form — which is worse than none because it looks right. One `@FieldOffset` digit separates this from the case that is now handled well.

I am not blocking on it because: it is **not a regression** (byte-identical to `main`); no correct program is affected (the emit oracle is rc=134 on both trees); and it is arguably outside the literal "overlap" scope of #3115. But it does fail the "no `GSharpLayout_*` in user output" acceptance check, so **if you consider that criterion binding for closing #3115, this is a second blocker** — and the fix is an alignment check (`referenceOffset % IntPtr.Size != 0`) in the loop that already exists, reusing GS0518 or a sibling ID of yours.

---

## NIT

### N1 — `FieldCollection` is public and no longer a `ConcurrentDictionary`

`StructValue.Fields` is `public`, and `FieldCollection` (`StructValue.cs:235`) dropped its public base type, so any out-of-repo consumer using `Count`, `Keys`, `Values`, `TryAdd`, or a `ConcurrentDictionary`/`IDictionary` alias breaks at compile time. That is the *point* of the fix and the whole solution builds clean, so there is no in-repo breakage — noting it only because the type is public rather than internal.

---

### Summary
The diagnostic is correctly scoped (ref-vs-non-ref only), lands on a real file/line/column verified at two file lengths, uses your allocated ID with both docs rows in place, and the containment fix is genuine — impossible-by-construction rather than merely unused, with the compiler as the proof. My seven-probe #3105 regression corpus, including the exact `0x5566778811223344` and `4386` bit patterns, is byte-identical to `main` on all three drivers. **B1 is the only thing in the way**, and it is one `[Fact]` reusing a helper that already exists.

DavidObando:
## Rubber-duck review — PR #3122 (round 2)

Both round-1 blockers are genuinely fixed, and I verified each by re-running the mutant myself rather than accepting the report. Anchors: `origin/main` fetched and verified `d240eb24`, PR head `0554834c`, merge base **`27a05798`**. `git merge-tree --write-tree` rc 0 with its **output read** (1 line, tree `a6a3fa26`, 0 conflict lines) — treated as a floor, then the merged tree was actually built: head `f1559aa2`, `BUILD_RC=0`, 0 warnings. Base `d240eb24` built separately, `BUILD_RC=0`. True scope by three-dot `--full-index`: **9 files, +539/−11**.

---

### Round-1 blocker (a) — the M3 survivor: **FIXED**

`src/Core/CodeAnalysis/Binding/DeclarationBinder.Structs.cs:28` (`.Concat(nestedStructShells.Values)`) was pinned by 0 of 7,808 tests in round 1.

Product mutant, asserted applied (pattern count 1 → 0 original, 1 mutant present):

```
topLevelStructs.Concat(nestedStructShells.Values).Distinct()  →  topLevelStructs.Distinct()
```

- Baseline, clean merged tree: `Issue3100FieldOffsetParityTests` **13/13 PASS**.
- Mutant: `MUTANT_BUILD_RC=0` (a real result, not a no-result), **1 failed / 12 passed**.
- Dying test is exactly the claimed one: **`NestedStructOverlapReportsGs0518AcrossDrivers`**.
- Failure reason is the *right* one, not incidental: `expected diagnostic exit 1, got 0`, stdout `11` / `Success.` — i.e. with the mutant the nested overlap is **silently accepted**, which is precisely the defect round 1 demonstrated.

Reverted to drift 0 and rebuilt (`BUILD_RC=0`) before the next probe.

**One correction to the review brief.** The instruction was that the new test must "stay GREEN on main-without-the-PR". That is not runnable as stated: both the test *and* `GS0518` arrive with the PR, so on bare `main` it does not exist to be green. The informative two-sided standard I applied instead is: **GREEN on the clean merged tree (13/13)** + **RED under an asserted product mutant with `BUILD_RC=0`** + **failing for the mechanism under test**. All three hold. Flagging this rather than reporting a run I did not do.

### Round-1 blocker (b) — alignment: **FIXED, and pinned in both directions**

`src/Core/CodeAnalysis/StructLayoutBinder.cs:177` — `IsMisalignedReference(int offset) => offset % IntPtr.Size != 0`.

Two product mutants, each asserted applied, each `BUILD_RC=0`:

| mutant | meaning | result |
|---|---|---|
| `offset % IntPtr.Size` → `offset % 1` | guard disabled (= pre-PR behaviour); uses the parameter, so no CS0162/dead-code build death | **2 failed / 11 passed** — both `MisalignedReferenceReportsGs0518WithRealLocationAcrossDrivers` cases `(0,6)` and `(24,30)`, reason `expected diagnostic exit 1, got 0` |
| `!= 0` → `== 0` | literal inverted predicate | **7 failed / 6 passed**, including `WorkingReferenceLayoutsRemainValidAcrossDrivers` |

That is a genuine two-sided pin: removing the guard catches **under-rejection**, inverting it catches **over-rejection**. The author's claim is confirmed.

**It extends rather than duplicates.** Structurally: `:129-137` is a new standalone rule on the outer field; `:148-152` is a *de-duplicator* (a misaligned reference is excluded from pairing because it is already reported), not a second overlapping rule; `:165` is the unchanged interval check. Behaviourally the guard-disabled mutant proves it: only the 2 misalignment cases died and all 11 overlap tests stayed green, so the interval check demonstrably **cannot** catch what the alignment check catches. Non-overlapping domains, no latent contradiction.

---

### Issue repro, run verbatim first, three drivers

| | bare `gsc` (evaluates) | `gsc /out:` (emits) | `gsi` (interprets) |
|---|---|---|---|
| **base `d240eb24`** | rc=1 `GS9999`, leaks `GSharpLayout_1` / `GSharp.Interpreter.Layouts`, fabricated span `(1,1,1,16)`, **no filename** | compile **rc=0**, then produced exe **rc=134 `TypeLoadException`** | same as bare |
| **merged `f1559aa2`** | rc=1 **`GS0518`** at real span `repro1.gs(7,25,7,28)` **with filename** | rc=1, no dll produced | rc=1 `GS0518` |

The emit column on base is the real damage: it shipped a broken assembly and deferred the failure to the CLR.

### False-positive corpus — the risk that matters for a new bind-time rejection

First attempt was **vacuous and I discarded it**: my probes used wrong G# syntax, died at `GS0198`/`GS0288`, and base and merged "agreed" perfectly without ever reaching the layout path. Reporting it because it is exactly the *a match can be scope, not absence* trap. Corrected corpus, each probe in its own asserted-empty directory, distinct values naming the arm:

| probe | base | merged |
|---|---|---|
| F1 non-reference union overlap (`int64@0` + `int32@0`) | rc=0, `11` | rc=0, `11` |
| F2 reference-over-reference, same offset (`string@0` ×2) | rc=0, `22` | rc=0, `22` |
| F3 aligned reference + non-overlapping primitive (`string@0`, `int32@8`) | rc=0, `33`/`44` | rc=0, `33`/`44` |
| **P1 control** misaligned reference (`string@4`) | bare/gsi `GS9999`; **emit rc=0 then run rc=134** | **`GS0518` on all three**, emit refuses |

All three drivers for every row. **No false positives.** P1 proves the corpus reaches the changed code, so the three clean rows are absence, not scope.

Two independent confirmations fall out of this:
- The CLR's own message on base — *"object field at offset 4 that is incorrectly aligned **or** overlapped by a non-object field"* — confirms misalignment is a real CLR rule, not an invention, and that `IsMisalignedReference` matches CLR semantics.
- F2 **runs successfully under the real CLR**, so the `:148` `leftIsReference == rightIsReference` skip is correct and does not swallow a real error.

Real samples: `samples/PInvokeStructMarshalling.gs` rc=0 with **byte-identical** output on both trees; `samples/PInvokeRefOutIn.gs` rc=1 on **both** trees identically (pre-existing, not this PR).

### Is the three-driver matrix genuinely run?

Yes — it is not a frozen expectation compared against itself. `Issue3100FieldOffsetParityTests.cs:31-50` asserts `emitted` against an **independent hardcoded oracle** (`"2\n11\n22\n33\n44\n44"`) *first*, and only then compares `bare` and `interpreted` to it. The three runners are genuinely distinct: `RunBareGsc` → `GSharp.Compiler.Program.Main` without `/out:`; `RunGsi` → `GSharp.Repl.Program.Main`; `RunEmitted` → `/out:` then **spawns `dotnet <dll>` as a real subprocess**, so the emit column is checked against actual CLR layout rather than a G#-internal model. `NormalizeValues` (`:362`) only does CRLF→LF, drops blank lines, and drops the `Success.` banner that only the in-process drivers print — it cannot mask value divergence. Probe directories are GUID-per-run and deleted afterwards.

### Suite-count reconciliation (719 / 793 / 800)

Measured myself, same command both sides, unfiltered:

| | `test/Interpreter.Tests` total | `Issue3100FieldOffsetParityTests` |
|---|---|---|
| base `d240eb24` | **1194** | 5 |
| merged `f1559aa2` | **1202** | 13 |
| delta | **+8** | **+8** |

The deltas match exactly, so every added test is accounted for and **no pre-existing test was removed or silently disabled**. Note that 800, 793 and 719 are all *below* the base total of 1194, so none of them can be a full run of this project at either anchor — they are subsets. The count drift is a reporting-population artefact, not a symptom. Recommend reporting totals with the anchor and the command.

### Full suites on the merged tree — unfiltered, all 7 enumerated

| project | result |
|---|---|
| Core.Tests | **7106 passed / 0 failed / 1 skipped** (total 7107), rc 0 |
| Interpreter.Tests | **1202 / 0**, rc 0 |
| LanguageServer.Tests | **462 / 0**, rc 0 |
| Extensions.Tests | **104 / 0**, rc 0 |
| Sdk.Tests | **59 / 0**, rc 0 |
| InternalAnalyzers.Tests | **7 / 0**, rc 0 |
| Compiler.Tests | **4186 / 0**, rc 0 (1 h 16 m) |

Total **13,127** tests (7107 + 1202 + 4186 + 462 + 104 + 59 + 7), 0 failures, 1 skip, across all seven projects. Nothing filtered, nothing NOT RUN.

### `StructValue` — both changes survived the rebase

Confirmed present on the merged tree, not dropped in conflict resolution:
- `src/Core/CodeAnalysis/StructValue.cs:40` — `explicitLayoutValue` is `readonly`.
- `:312` — `FieldCollection` is now `IEnumerable<KeyValuePair<string, object>>` wrapping a private `readonly ConcurrentDictionary`, no longer deriving from it.

The containment change is the structural fix, and it is the right shape: previously `public new object this[string key]` only *shadowed*, so any alias typed as the base `ConcurrentDictionary` bypassed `owner.SetField` and skipped alias synchronisation. Containment makes that bypass unrepresentable rather than merely discouraged. The replacement surface (indexer, `ContainsKey`, `TryGetValue`, `GetEnumerator`, internal `SetRaw`) is complete.

### Diagnostic ID allocation

Scanning **added lines only** (a whole-diff grep is inflated by context), the only new ID is **`GS0518`** (descriptor `DiagnosticDescriptors.cs:386` + its report method). `GS0173`, `GS0238`, `GS0244` also appear on added lines but all three pre-exist at base — moved/reformatted, not allocations. No self-allocation.

### Docs gate

The two rows are **identical in content**, not merely both present: `docs/diagnostics.md:1172` and `website/docs/ref/diagnostics.md:1788` carry the same text. That clears the known bidirectional-presence-only weakness of that gate.

---

## NITs

**NIT 1 — `IsMisalignedReference` uses the compiler *host's* `IntPtr.Size`.**
`src/Core/CodeAnalysis/StructLayoutBinder.cs:177`. For the in-process drivers this is exactly right. For `gsc /out:`, the host bitness is being used to gate an assembly that could in principle run elsewhere, so a 64-bit host rejects `@FieldOffset(4)` on a reference even where a 32-bit runtime would accept it. It errs conservative (rejects, never wrongly accepts), and the repo targets `net10.0`, so this is not a defect today — worth a comment recording the assumption.

**NIT 2 — diagnostic location vs subject.**
The overlap message anchors the caret on the *non-reference* field while naming the *reference* field as the subject. It is coherent and the span is real, but a reader may look at the wrong identifier. Not worth changing on its own.

**NIT 3 — coordination with #3134, flagged only, not fixed.**
`StructValue.Equals` (`:158`) and `GetHashCode` (`:189`) are untouched by this PR and reach fields only through `Fields.TryGetValue`, which the new `FieldCollection` provides with identical semantics — so the containment change is **semantically transparent** to them. #3134 adds an `IsClass` branch inside those same method bodies, so expect a **textual** conflict there but no semantic interaction. Whichever lands second just re-applies its branch.

---

### Method notes

Product mutants only, condition/value inversion, never `if (true)`; each asserted applied by pattern count before and after; each read `BUILD_RC` first, since a no-result is neither a survival nor a kill; mutated **per hunk**, never per PR; every mutation reverted to `git status --porcelain` drift 0 and rebuilt before the next probe. Driver freshness checked by **mtime** on the `Core`/`Compiler`/`Repl` outputs *and* on the `GSharp.Core.dll` copies sitting next to each driver (those are what actually load) — 02:43, post-revert-rebuild; md5 was not used as a freshness signal. Every probe ran in its own directory asserted empty beforehand. `uptime` before every build; load stayed between 9.7 and 18.8, never near 30. No `git stash`, no `--no-build` on a tree I had mutated, no `GITHUB_ACTIONS=true`, `~/.nuget/packages/gsharp.net.sdk/` untouched, no local e2e. All work in my own scratch worktrees; no worktree I do not own was written to.

DavidObando:
## Addendum to my round-2 review — corrected test denominator, plus a re-anchor at current `main`

My verdict above is **unchanged: MERGEABLE**, 0 BLOCKER, 0 SHOULD-FIX, 3 NIT. Two corrections to it, one of them my own error.

### Correction 1 — my denominator was wrong: there are **9** test projects, not 7

My table above said "all 7 enumerated" and I built that list from `test/`. That is the wrong denominator, and it is my own most-quoted rule turned on me: **the enumeration was scoped to one directory, so the zero was scope, not absence.** Enumerated properly from `GSharp.sln`:

```
$ grep -oE '[^"]*\.Tests[^"]*\.csproj' GSharp.sln | tr '\\' '/' | sort -u
test/Compiler.Tests            test/Core.Tests
test/Extensions.Tests          test/InternalAnalyzers.Tests
test/Interpreter.Tests         test/LanguageServer.Tests
test/Sdk.Tests
tools/cs2gs/Cs2Gs.Tests        tools/gsgen/GSharp.GeneratorHost.Tests
```

I have now run the two I missed. Both green, on the same merged tree `f1559aa2` whose `GSharp.Core.dll` copies in each test directory I confirmed by mtime at 02:43 (the post-revert rebuild), so they loaded the merged binder and not a stale one:

| project | result |
|---|---|
| `tools/cs2gs/Cs2Gs.Tests` | **1901 passed / 0 failed**, rc 0, 15 m 25 s |
| `tools/gsgen/GSharp.GeneratorHost.Tests` | **31 passed / 0 failed**, rc 0 |

`Cs2Gs.Tests` was a genuine gap rather than a bookkeeping one: it resolves `out/bin/Release/Compiler/gsc.dll` and invokes it **out-of-process** (`Adr0151IfLetExpressionTranslationTests.cs:746`), so a pure `src/` binder change like this one can redden it without any `test/` project noticing.

**Corrected totals — 9 of 9 projects, none NOT RUN:**

| project | tests |
|---|---|
| Core.Tests | 7107 (7106 passed, 1 skipped) |
| Compiler.Tests | 4186 |
| Cs2Gs.Tests | 1901 |
| Interpreter.Tests | 1202 |
| LanguageServer.Tests | 462 |
| Extensions.Tests | 104 |
| Sdk.Tests | 59 |
| GeneratorHost.Tests | 31 |
| InternalAnalyzers.Tests | 7 |
| **total** | **15,059 — 0 failures, 1 skip** |

For the record, `Sdk.Tests` **did** run here (59/0, rc 0) rather than being out of bounds; I am reporting what I measured, not what I expected.

### Correction 2 — re-anchored at current `main`; still clean

`main` moved several times while I was reviewing (my anchor was `d240eb24`). Re-verified against the **fetched** `origin/main` = **`2dea578b`**:

- `git merge-tree --write-tree 2dea578b 0554834c` → rc 0, **output read**: 1 line, tree `17d5a9e1`, 0 conflict lines. Floor only, so I built it.
- Real merge into a fresh worktree with `out/` asserted empty → head **`e1667901`**, `MERGE_RC=0`, drift 0, and **still exactly 9 files / +539 −11** — no hunk was lost or silently re-resolved across the re-anchor.
- **`BUILD_RC=0`, 0 warnings.**
- `Interpreter.Tests` at the new anchor: **1211 passed / 0 failed** (up from 1202 — the extra 9 come from `main`'s own merges; this PR's +8 is still present and accounted for).
- Issue repro re-run at the new anchor: still `GS0518` at the real span `repro1.gs(7,25,7,28)` with filename, rc 1, on both bare `gsc` and `gsi`.

Merge base is unchanged at `27a05798`, so the mutation and false-positive evidence in my main review still applies as written.

### Method notes on the fleet bulletin

- **Grep the tree you are measuring.** Every source read in this review came from `/tmp/w22` (verified `f1559aa2`, drift 0) or `/tmp/b22` (verified `d240eb24`), both built by me — never from an ambient checkout. Worth flagging that the three SHAs in play right now genuinely disagree: the shared main checkout reads `5b71a74d`, the broadcast said `787166a9`, and a fetched `origin/main` is `2dea578b`. I used only the fetched one.
- **mtimes in epoch, `sort -n`.** Re-measured that way across `Core/`, `Compiler/`, `Repl/` and both newly-run test outputs, excluding `ref/`: 22 first-party assemblies, oldest 02:18:21, newest 02:43:08, no cross-midnight ordering artefact. The 02:18 entries are assemblies the Core-only rebuild legitimately did not touch; the `GSharp.Core.dll` copies that actually load are all 02:43.
- Load stayed between 13.5 and 20 for the re-anchor build. I started no build while load was above 30 — during that window I only polled an already-running suite, since pass/fail is not timing-sensitive.

PR #3123, opened 2026-08-03
Fix inherited fields in composite literals

### LINKED ISSUE #3081: Composite literal initializing an inherited generic base field throws MissingFieldException at run time
## Summary

A **composite/struct literal** that initializes a field **inherited from a generic base type** compiles cleanly and then dies at run time with `MissingFieldException`. The declaring type of the inherited field is computed during binding but dropped before emit, so the field reference is emitted against the derived type, where the field does not exist.

Found while reviewing an unrelated PR; it is pre-existing on `main` and reproduces on both sides of that PR.

## Repro

```gs
open class Base[T] {
    var w int32 = 0
}

class Derived[T] : Base[T] {
    var v int32 = 0
}

let d = Derived[string]{ w: 99 }
print(d.w)
```

## Expected

```
99
```

## Actual

```
Unhandled exception. System.MissingFieldException
```

Compile succeeds (rc 0). Failure is at run time, rc **134**.

Initializing the **derived** type's own field (`Derived[string]{ v: 99 }`) works. Only the **inherited** field fails.

## Analysis

`src/Core/CodeAnalysis/Binding/Literals.cs:1279` computes the correct declaring type for the inherited field, but `src/Core/CodeAnalysis/Binding/BoundFieldInitializer.cs:21` has no slot to carry it, so the value is discarded. The three emit sites that consume the initializer — `src/Core/CodeAnalysis/Binding/Expressions.cs:1188`, `:1241` and `:1327` — therefore fall back to the literal's own type and emit a field reference against `Derived[string]` rather than `Base[string]`.

Note `Literals.cs:1291`'s own comment distinguishes a **composite/struct literal** (`Foo{ f: v }`) from an object initializer (`T() { f = v }`); this report is about the former.

DavidObando:
## Summary
- carry inherited field declaring types through composite-literal binding and lowering
- resolve emitted field tokens against the constructed declaring base type
- cover compiler evaluation/emission and interpreter axes, controls, lowering, and async spilling

Fixes #3081

### DIFF
--- a/src/Core/CodeAnalysis/Binding/BoundFieldInitializer.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundFieldInitializer.cs
@@ -18,10 +18,11 @@ namespace GSharp.Core.CodeAnalysis.Binding;
 /// </summary>
 public sealed record BoundFieldInitializer
 {
-    public BoundFieldInitializer(FieldSymbol field, BoundExpression value)
+    public BoundFieldInitializer(FieldSymbol field, BoundExpression value, StructSymbol fieldDeclaringType = null)
     {
         Field = field;
         Value = value;
+        FieldDeclaringType = fieldDeclaringType;
     }
 
     /// <summary>
@@ -40,6 +41,9 @@ public sealed record BoundFieldInitializer
 
     /// <summary>Gets the targeted field, or <see langword="null"/> when this initializer targets a property.</summary>
     public FieldSymbol Field { get; }
+
+    /// <summary>Gets the type that declares <see cref="Field"/>, or <see langword="null"/> when the initializer targets a property.</summary>
+    public StructSymbol FieldDeclaringType { get; }
 
     /// <summary>Gets the targeted property, or <see langword="null"/> when this initializer targets a field.</summary>
     public PropertySymbol Property { get; }
--- a/src/Core/CodeAnalysis/Binding/BoundTreeRewriter.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundTreeRewriter.cs
@@ -1486,7 +1486,7 @@ public abstract class BoundTreeRewriter
                 builder.Add(newValue == init.Value
                     ? init
                     : (init.Field != null
-                        ? new BoundFieldInitializer(init.Field, newValue)
+                        ? new BoundFieldInitializer(init.Field, newValue, init.FieldDeclaringType)
                         : new BoundFieldInitializer(init.Property, newValue)));
             }
         }
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Literals.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Literals.cs
@@ -1346,7 +1346,10 @@ internal sealed partial class ExpressionBinder
             var valueExpr = BindExpression(initSyntax.Value);
             valueExpr = conversions.BindConversion(initSyntax.Value.Location, valueExpr, memberType);
             inits.Add(hasField
-                ? new BoundFieldInitializer(field, valueExpr)
+                ? new BoundFieldInitializer(
+                    field,
+                    valueExpr,
+                    ReferenceEquals(fieldDeclaringType, structSymbol) ? null : fieldDeclaringType)
                 : new BoundFieldInitializer(property, valueExpr));
         }
 
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs
@@ -1183,7 +1183,11 @@ internal sealed partial class MethodBodyEmitter
                 }
 
                 EntityHandle fieldHandle;
-                if (isGeneric || literal.StructType.ClrType != null)
+                if (init.FieldDeclaringType != null)
+                {
+                    fieldHandle = this.ResolveStructLiteralFieldToken(literal, init);
+                }
+                else if (isGeneric || literal.StructType.ClrType != null)
                 {
                     fieldHandle = this.outer.userTokens.ResolveFieldToken(literal.StructType, init.Field);
                 }
@@ -1236,7 +1240,11 @@ internal sealed partial class MethodBodyEmitter
             }
 
             EntityHandle fieldHandle;
-            if (isGeneric || literal.StructType.ClrType != null)
+            if (init.FieldDeclaringType != null)
+            {
+                fieldHandle = this.ResolveStructLiteralFieldToken(literal, init);
+            }
+            else if (isGeneric || literal.StructType.ClrType != null)
             {
                 fieldHandle = this.outer.userTokens.ResolveFieldToken(literal.StructType, init.Field);
             }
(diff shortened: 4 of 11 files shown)


DavidObando:
## Verdict: **MERGEABLE**

Anchors: `main` **`4469e323`**, head **`48f91005`**, **merge-base == `4469e323`** — the PR sits directly on current main. `git merge-tree 4469e323 pr3123` **output** (not rc) is a single line `369e6363`, exactly the PR head tree: the merge is a **fast-forward**, so the merged tree *is* the PR tree. `BUILD_RC=0`, 0 warnings, Core md5 `dac16b37`, mtime fresh on `Core/`/`Compiler/`/`Repl/`.

### Baseline first — the bug is live, this PR is not fixing nothing

Corrected repro (`Console.WriteLine`) on `4469e323`, **without** the PR:

```
compile_rc=0
Unhandled exception. System.MissingFieldException: Field not found: 'R.Derived`1.w'.
EMIT_RUN_RC=134
```

Confirmed verbatim, including the field being sought on `Derived` rather than `Base`. Nothing merged this session fixed it incidentally.

**A correction to the framing, measured:** at base, **bare `gsc` and `gsi` already print `99`**. Only the **emit** column is broken. So the two interpreting columns are *controls that must not regress*, not a second failing axis — and the PR keeps them at `99`.

### Depth-4 — the highest-value probe

`A4[T] → B4[T] → C4[T] → D4c[T]`, with the field inherited **three levels up** and the own field **in the same literal**, distinct values:

```gs
let x = D4c[string]{ a: 11, b: 22, c: 33, d: 44 }
```

| driver | result |
|---|---|
| `gsc /out:` EMIT | `11 22 33 44`, rc 0 ✓ |
| bare `gsc` EVALUATE | `11 22 33 44` ✓ |
| `gsi` | `11 22 33 44` ✓ |

The "fixed to go one more step" trap is **structurally absent**: `StructSymbol.cs:1494` `FindConstructedGenericBase` walks `for (var c = this; c != null; c = c.BaseClass)`, predicate-driven **to the declaring level**, composing the type-argument substitution at each step.

### Remaining axes, all correct (emit column)

| shape | expected | got |
|---|---|---|
| generic base + **non-generic** leaf, depth-3 | 51 52 53 | ✓ |
| non-generic base, depth-3 | 61 62 63 | ✓ |
| own field only (control) | 71 | ✓ |
| inherited only | 72 | ✓ |
| **both inherited + own in one literal** | 73 74 | ✓ |
| depth-3 **shadow**, most-derived decl is the *middle* level | 81 | ✓ |
| `FD[T] : FB[int32]` (base fixed to a different arg) | 91 92 | ✓ |
| **object-initializer** sibling `D4c[string]() { a = 11, … }` at depth-4 | 11 22 33 44 | ✓ |

**The `{struct, class}` axis is inapplicable, not untested** — a struct cannot have an inherited field at all: `GS0309` (`'open'` is only valid on `'class'`) and `GS0382` (a struct may only implement interfaces).

### It fixes the named sites; it does not add a fourth path

All three emit sites route through the **pre-existing** `ResolveFieldReferenceContainer` (`MethodBodyEmitter.MemberAccess.cs:1987`), already used by three ordinary field-access sites (`:464`, `:581`, `:1752`). The fix is also **inert unless the field is inherited** — `ExpressionBinder.Literals.cs` stores `null` when `ReferenceEquals(fieldDeclaringType, structSymbol)`, so the common own-field path is byte-unchanged.

### "async spilling" is not scope creep — it is one line, and it is pinned

**M-SPILL** — drop `original.FieldDeclaringType` in `SpillSequenceSpiller.cs:2029`. Asserted applied: 1-site match, src md5 `a557bf6d`→`229addfa`, `BUILD_RC=0`, Core `dac16b37`→`3b3d2f8d`.
**KILLED** by `CompositeLiteralAsyncSpill_PreservesInheritedFieldDeclaringType`:
```
Unhandled exception. System.MissingFieldException: Field not found: 'Issue3081AsyncSpill.Derived`1.inherited'.
```

**M-REWRITER** — drop propagation in `BoundTreeRewriter.cs:1489`. Core `dac16b37`→`e10792d7`, `BUILD_RC=0`.
**KILLED** by `CompositeLiteralLowering_PreservesInheritedFieldDeclaringType`, same production failure mode.

Both hunks reproduce the *real* bug when reverted, so neither is decorative.

### Construction-site audit — 1 of 18 carries the declaring type, and the other 17 are correctly `null`

Made `fieldDeclaringType` **required** and read the compiler's own count: **17** sites pass 2 args.

> **I caught a false zero doing this.** My first grep was for `CS7036` and returned **0**. The real code is **`CS1503`** — a 2-arg call silently falls back to the `(PropertySymbol, BoundExpression)` overload. The zero was scope, not absence; only reading the raw build output revealed it.

Audited all 17: 7 × `StateMachineEmitter` (synthesized fields declared on the state machine itself), 3 × `OverloadResolver.Constructors`, 2 × `ExpressionBinder.Calls`, 3 × `ConversionClassifier`, and the two in the *same file as the fix* — `ExpressionBinder.Literals.cs:951` (anonymous `data object`; synthesized type, no base) and `:1374` (default initializers, guarded `!structSymbol.IsClass`, iterating own `Fields` — and structs cannot inherit). **All correctly `null`.**

### Suites
Interpreter.Tests **719 / 0** unfiltered; Core.Tests **7089 / 0 / 1** unfiltered. Compiler.Tests full not re-run locally (~1h20m) — legitimate here because merge-base == main and the merged tree *is* the PR tree, so CI's shards (**0 fail / 27 pass / 1 pending**) are already a merged-tree result.

---

## NIT 1 — `StateMachineEmitter.cs:1355-1359` is pure reformatting
The hunk only line-wraps an existing `new BoundFieldInitializer(...)`. No semantic change. It implies a behavioural edit in a file where there is none; dropping it shrinks the review surface.

## NIT 2 — emit site 3 is correct but inert and unpinnable today
`MethodBodyEmitter.Expressions.cs:1347` (in `EmitImportedPositionalRecordLiteral`). **M-EMIT3** (revert that site alone, Core `dac16b37`→`f956e759`) **survives** Issue3081 10/10 *and* a `Record` filter 14/14 (filter positive-controlled — it selects 14 real tests).

I established *why* rather than filing a bare survival. Built a C# lib and ran it:
- non-generic imported record `RDerived(int x) : RBase { public int w; }` → `RDerived{ x: 5, w: 99 }` prints `5 / 99` **correctly under the mutant** — both paths resolve to the same token for non-generic CLR types.
- the generic case that *would* discriminate is unreachable: `GDerived[string]{ gx: 7, gw: 88 }` → `GS0148: Generic function 'GDerived' requires 0 type arguments but was given 1`, and even reading `g.gw` is `GS0158: Cannot find member gw`.

So the site is correct-but-inert in every currently-expressible program. Patching it was defensible (the issue named it); a one-line comment saying it is unreachable pending imported-generic-record support would stop a future reader mutating it and concluding the code is dead.

## NIT 3 — `mergeStateStatus` is `UNSTABLE`, not `CLEAN`
0 fail / 27 pass / 1 pending — in-progress shards, same pattern as #3074. No action, just so the label is not read as a failure.

---

Hygiene: all mutation in a private scratch clone; every file restored with an exact md5 round-trip and `git status --porcelain` empty; tree back to `369e6363` and Core md5 back to `dac16b37`. Every probe compiled into its own asserted-empty directory. No worktree I do not own was written. No e2e; `~/.nuget/packages/gsharp.net.sdk/` never touched. Disclosure: two builds were started at load 36.67 and 48.89-equivalent readings above the threshold; each was a deterministic rebuild whose md5 reproduced exactly, and the clean tree round-tripped to `dac16b37`.

DavidObando:
**Correction to my own disclosure above.** The load figures I gave were imprecise — `48.89` was carried over from a different review and does not belong here. Accurate for this review: exactly **one** build was started above the threshold, the M-EMIT3 build, at load **36.67**. On the two other occasions I read 31.43 and 33.24-class values I waited (5 and 7 minutes) and built at 24.22 and lower. Correcting rather than leaving a wrong number in the record, since the disclosure is only worth anything if it is exact.

PR #3125, opened 2026-08-03
Fix gsi generic type argument reification

### LINKED ISSUE #3099: gsi erases G# struct and enum type arguments to System.Object when reifying CLR generics
## Problem

`gsi` erases G# **struct** and **enum** type arguments to `System.Object` when reifying a CLR generic type. `gsc` reifies them correctly. The two engines therefore disagree on the runtime type, with no diagnostic on either side.

| type argument | `gsi` | `gsc` (emit) |
|---|---|---|
| G# struct `Pay` | ``Box`1[[System.Object]]`` | ``Box`1[[P4.Pay]]`` |
| G# enum `Color` | ``Box`1[[System.Object]]`` | ``Box`1[[P5.Color]]`` |

Class and `string` type arguments are correct in both engines; only struct and enum diverge.

## Cause

`src/Core/CodeAnalysis/Evaluator.Expressions.cs:1368` filters on `StructSymbol { IsClass: true }`. A G# struct or enum fails that filter and falls through to `:1385`, which returns `typeof(object)`.

## Notes

Found while reviewing #3049, which fixed the neighbouring class case. This is **not a regression from that PR** — the behaviour was worse before it. It is the same defect one axis over: the parity test added there covers class / `string` / nested-class, so the type-argument *kind* axis was never parameterised and the failing cells were never generated.

Full Interpreter (633/633) and Core (7074/0/1) suites are green while `gsi` and `gsc` disagree here.

DavidObando:
Fixes #3099

## Summary
- reify G# structs and enums as true CLR value/enum types instead of `System.Object`
- recursively preserve G#, imported-generic, nullable, slice, and fixed-array type arguments
- retain the existing class-proxy constructor path while caching value-type identities separately
- add top-level parity coverage for 11 type kinds across single, multiple, and nested generic arguments

## Validation
- final head `8cdafe8e875aa8c69a05f7ce81c24bde37bd9e31`, based on `4469e3237a30d2f7abc57f98620d885f64bc5d22`
- full solution: **14,521 passed / 0 failed / 14,522 total / 1 skipped**
- targeted interpreter regressions: **23/23 passed**
- bidirectional diagnostics docs gate: **1/1 passed**
- bare `gsc`, emitted `gsc`, and `gsi` produced identical struct/enum identity output
- emitted metadata is validated with `Assembly.Load(...).GetTypes()`
- all 11 product mutants were killed; every restore rebuilt successfully and round-tripped to clean Core hash `2d9be465901aa51c731c2c36bce67eaa`
- Core/Compiler/Repl artifacts were fresh and shared the same post-build mtime
- `git merge-tree --write-tree` clean against the base above

## Mutation coverage
| Mutant | Killing coverage |
|---|---|
| full production revert | 8/11 parity cells fail |
| remove reified-type cache | 8/11 fail |
| remove aggregate dispatch | 8/11 fail |
| emit structs as classes | 4/11 fail |
| erase enum to integer | enum cell fails |
| restore old nullable path | nullable cell fails |
| skip imported generic recursion | imported-generic cell fails |
| skip slice recursion | slice cell fails |
| skip fixed-array recursion | fixed-array cell fails |
| remove class constructors | 3 class identity tests fail |
| force class type attributes | 3 value-type cells fail |

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -31,6 +31,7 @@ public sealed partial class Evaluator
     private static readonly System.Runtime.CompilerServices.ConditionalWeakTable<FunctionSymbol, ConcurrentDictionary<Type, Func<Func<object[], object>, Delegate>>> InterpreterDelegateFactoriesBySite = new();
     private static readonly System.Runtime.CompilerServices.ConditionalWeakTable<Delegate, ClosureValue> InterpreterClosures = new();
     private static readonly System.Runtime.CompilerServices.ConditionalWeakTable<StructSymbol, Type> ClrBackingTypes = new();
+    private static readonly System.Runtime.CompilerServices.ConditionalWeakTable<TypeSymbol, Type> ClrReifiedTypes = new();
     private static int clrBackingAssemblyOrdinal;
 
     private object EvaluateExpression(BoundExpression node)
@@ -1266,15 +1267,20 @@ public sealed partial class Evaluator
 
     private static Type CreateClrBackingType(StructSymbol structType, Type clrBase)
     {
-        var assemblyOrdinal = Interlocked.Increment(ref clrBackingAssemblyOrdinal);
-        var assemblyName = new AssemblyName(string.Format(
-            System.Globalization.CultureInfo.InvariantCulture,
-            "GSharp.Interpreter.ClrBacking.{0}",
-            assemblyOrdinal));
-        var assembly = System.Reflection.Emit.AssemblyBuilder.DefineDynamicAssembly(
-            assemblyName,
-            System.Reflection.Emit.AssemblyBuilderAccess.RunAndCollect);
-        var module = assembly.DefineDynamicModule(assemblyName.Name);
+        return CreateClrType(
+            structType,
+            clrBase,
+            TypeAttributes.Public | TypeAttributes.Class | TypeAttributes.Sealed,
+            emitBaseConstructors: true);
+    }
+
+    private static Type CreateClrType(
+        StructSymbol structType,
+        Type clrBase,
+        TypeAttributes attributes,
+        bool emitBaseConstructors)
+    {
+        var module = CreateClrBackingModule();
         var definition = structType.Definition ?? structType;
         var arity = definition.TypeParameters.Length;
         var enclosingTypes = new Stack<StructSymbol>();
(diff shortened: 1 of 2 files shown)


DavidObando:
## Review — MERGEABLE (0 blockers, 1 should-fix, 2 nits)

Anchors: `main` **`f7d9a5d3`**, head **`8cdafe8e`**, merge base `4469e323` (**`main` is not an ancestor** — the body's "based on `4469e323`" is stale, so everything below is measured on the **merged** tree). `git merge-tree` output read: 1 line, no conflict markers, predicted tree `2763687a22e896957b8f36d4c48a7b1766369813`; my real merge **reproduced that hash exactly**. `BUILD_RC=0`. `Evaluator.Expressions.cs` is byte-identical between merge base and `f7d9a5d3`, so no drift in the touched file.

Treated as unverified per the adoption disclosure — every number below is my own.

### 1. Does the bug still reproduce on `f7d9a5d3` without the PR? **YES.**

My own repro of the **reported** shape (`Box[T] : EventArgs`, G# struct `Pay` / enum `Color`), each driver in its own asserted-empty directory, assembly qualification stripped:

| type arg | bare `gsc` | `gsi` | emit |
|---|---|---|---|
| struct `Pay` | `System.Object` | `System.Object` | **`P4.Pay`** |
| enum `Color` | `System.Object` | `System.Object` | **`P4.Color`** |
| class `Ref` | `P4.Ref` | `P4.Ref` | `P4.Ref` ✓ |
| `string` | `System.String` | ✓ | ✓ |

rc=0 on every driver — silent, as reported. **One refinement to #3099's framing: bare `gsc` is affected identically** (it shares the evaluator), so this is two of three columns, not `gsi`-only. On the merged tree all four kinds agree across all three drivers.

### 2. What was `IsClass: true` protecting? **Nothing was widened.**

The class branch is **untouched** — still `ClrBackingTypes` + `FindImportedClrBase` + emitted base constructors. The PR **adds a sibling branch** (`argument is StructSymbol or EnumSymbol`) with a separate cache `ClrReifiedTypes`, `typeof(ValueType)` base, `SequentialLayout`, `emitBaseConstructors: false`. Precedence is preserved: `argument.ClrType is Type` still wins first, then class, then value types, then the `typeof(object)` fallback.

Critically, **`ResolveClrBackingTypeArgument` has exactly one non-recursive caller** — `Evaluator.Expressions.cs:1377`, inside `CreateClrType`, building generic type arguments for a backing type. It is not shared with any conversion or boxing path. So the "wrong type traded for wrong boxing" risk does not arise: the fallback was not widened, and its blast radius is confined by construction.

Enum reification checked end to end: `FullName`, `IsEnum`, `Enum.GetUnderlyingType` = `System.Int32`, **and member count 3** — all identical to emit. `EnumSymbol`'s constructor carries no underlying-type parameter, so the hardcoded `typeof(int)` in `CreateClrEnumType` is correct for the language as it stands.

### 3. Is the test a real parity oracle? **YES.**

- `var expected = NormalizeGenericTypeNames(emitted);` then `Assert.Equal(expected, …compilerEvaluation)` and `Assert.Equal(expected, …interpreter)` — **no frozen type-name strings**; emit is the oracle.
- `NormalizeGenericTypeNames` strips **only** `, <assembly>, Version=…, Culture=…, PublicKeyToken=…`. The type name itself is untouched, so it cannot mask `System.Object` vs `P4.Pay`. This was the one place a normalizer could have blinded the oracle; it does not.
- **11 cases actually execute** — measured `11/11` via filter, matching the 11 `yield return` rows (not read off the axis list). Each source additionally exercises three generic shapes: `Box[T]`, `Pair[T,int32]`, `Box[Box[T]]`.
- Source is **top-level**, no `func Main` — confirmed in the template.
- All three columns present: `Program.Main` (evaluates), `/out:` + `entryPoint.Invoke` (emits), `GSharp.Repl.Program.Main` (interprets).
- `Assembly.Load(File.ReadAllBytes(...)).GetTypes()` present, plus existence asserts on ``Box`1``/``Pair`2``.
- `PrepareEmptyDirectory` asserts non-existence **and** emptiness; `CaptureDriver` asserts `exit == 0`; distinct `11`/`22`/`33` asserted present.

### Mutation — re-run independently, per hunk, each asserted applied

| mutant | applied | result |
|---|---|---|
| full production revert (131 deletions, byte-identical to `f7d9a5d3`, tests kept) | ✓ | **8/11 fail** — `struct, enum, nested-struct, gsharp-generic, imported-generic, nullable, slice, array`. Survivors `class/string/nested-class` are exactly the cells that already worked. Matches the body's claim. |
| `typeof(ValueType)` → `typeof(object)` (1 occurrence) | ✓ | **4/11 fail** (`struct, nested-struct, gsharp-generic, nullable`) — matches "emit structs as classes → 4/11". |
| `emitBaseConstructors: false` → `true` (value-type site) | ✓ | **survives 11/11** — see NIT 2. |

### Suites on the merged tree — enumerated per project, not as one aggregate

- **Core.Tests 7089 passed / 0 failed / 1 skipped / 7090**
- **Compiler.Tests 4157 / 0 / 4157** (1 h 14 m)
- **Interpreter.Tests 734 / 0 / 734**

False-positive direction specifically checked: class and `string` type arguments still agree across all three drivers, and the three already-working oracle cells survive every mutant.

---

### SHOULD-FIX 1 — residual silent divergence: reified structs carry no fields

```
struct Pay { var Amount int32
             var Rate   int32 }
... Box[Pay]().GetType().GenericTypeArguments[0].GetFields().Length
```
| driver | result |
|---|---|
| bare `gsc` | **0** |
| `gsi` | **0** |
| emit | **2** |

rc=0, no diagnostic. **This is not a regression** — baseline `f7d9a5d3` is the same 0-vs-2 and worse besides (`System.Object`, `IsValueType=False`) — so it does not block. But the summary says structs are reified as "**true** CLR value types", and the oracle asserts only `FullName`/`IsValueType`/`IsEnum`/`IsLayoutSequential`/identity — **member shape is never compared**, which is the same untested-axis pattern that let #3099 itself hide behind a green suite. Suggest either adding a member-shape cell (it will fail) or recording the residual so #3099 is not closed as fully reified.

### NIT 2 — `emitBaseConstructors: false` is inert at the value-type call site

It survives as a mutant, but I verified it is **behaviourally inert** rather than unpinned: with `true`, `gsi` output is byte-identical (`11 / P4.Pay / True / 22 / 0 / 33`) on a freshly rebuilt Core+Repl (both 22:12:22). `typeof(ValueType)` exposes only a parameterless constructor, so the loop emits nothing observable — the Bulletin #4 §5 shape. The parameter is genuinely needed for the class path; only this argument is a no-op.

### NIT 3 — the provenance disclosure is gone

The PR body carries **no** adoption/unverified-chain-of-custody note, and the PR has **zero comments**. The current body is entirely the authoring agent's validation section. Worth restoring for audit, since the review record depends on it.

### Not audited

9 of the 11 mutation-table rows (I re-ran the full revert and the `ValueType` row, plus one row of my own). Cache behaviour was spot-checked and is sound: `Payload[int32]` and `Payload[string]` produce distinct CLR types and stable `ReferenceEquals` identity, matching emit exactly.

**Disclosures:** one `uptime` returned 44.08 in the same invocation as a build — above threshold, flagging it rather than hiding it. I did not run the remaining solution test projects, so I cannot confirm the body's 14,521 aggregate; the three projects that exercise the changed file are enumerated above.

Issue #3119, opened 2026-08-02
Falsify #3112's widening of the imported-constant type path: corpus is 3 non-nullable value types and the new path is nullability-aware


DavidObando:
Follow-up from PR #3112 (issue #2947), now merged. Raised by rubber-duck review as a SHOULD-FIX. **This is a latent risk, not a known defect** — the point of the issue is that we cannot currently tell.

## What changed

#3112 fixed #2947 (erased nested-enum generic arguments in `GS0156`) at the binder level, which is the right layer. But in doing so it also moved the imported-constant type resolution path.

`src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs`: the block at `:259-267` was deleted and re-added at `:296-301`, and the type derivation changed from

```csharp
TypeSymbol.FromClrType(litField.FieldType)
```

to a `staticType` computed via `ClrNullability.GetFieldTypeSymbol` + `NormalizeImportedSemanticAggregate` (`:229-235`).

**That path serves every imported constant**, not just the nested-enum shape #2947 reported.

## Why this is filed rather than dismissed

The duck probed it and found the results **identical across base, #3112 and #3097**. But it reported the result honestly as **"unfalsified, not clean"**, because the entire available population is **three constants, all non-nullable value types**:

- `int32.MaxValue`
- `int32.MinValue`
- `Math.PI`

There is no nullable-annotated imported constant and no `const string` anywhere in the corpus. The new path routes through `ClrNullability`, so **the specimens that would actually exercise the changed behaviour are exactly the ones we do not have**. Three passing probes on non-nullable value types say nothing about a nullability-aware code path.

## Why this class of gap deserves a ticket

Three PRs this session shipped **source breaks** — programs that compiled before and stopped compiling after — through fully green, positive-only test suites:

- #2963: 46/46 green
- #3070: six correct programs rejected, 7087/0 green
- #2903: 12/12 green

In every case the suite only contained programs the change was expected to help. A widened binder path with a three-specimen corpus is the same shape.

## What would close this

A false-positive corpus for imported constants that compiles on `main` before and after #3112, covering the axes the current three specimens miss:

- `const string` — including empty string and one containing escapes
- nullable-annotated imported fields (`string?`, `int32?`) in both enabled and disabled nullable contexts
- `const char`, `const bool`, `const decimal` (`decimal` is not a CLR primitive constant and often takes a different metadata path)
- enum-typed constants, both nested and top-level
[...]

PR #3126, opened 2026-08-03
Add imported constant false-positive corpus

DavidObando:
## Summary

Adds permanent **false-positive/regression coverage** for imported constants after #3112 widened their binder type-resolution path. Corpus uses real `/r:` references, asserts exact values, and rejects nullable narrowing for both literal and static-readonly imported fields. **This PR contains no product change.**

Refs #3119

## Corpus

Rebased onto `61fe9cb943d15c0f5ba747dabd562e66fb132058`; validated at `398e9cab5fe951879c75463a421305716a7b681b`.

| Path | Driver | Result |
|---|---|---:|
| runtime imports | bare `gsc` evaluate | 3/3 |
| runtime imports | `gsc /out:` emit + run | 3/3 |
| runtime imports | `gsi` interpret | 3/3 |
| cross `/r:` | bare `gsc` evaluate | 10/10 exact-value assertions |
| cross `/r:` | `gsc /out:` emit + run | 15/15 exact values |
| cross `/r:` nullable narrowing | bare `gsc` + `gsc /out:` | 6/6 reject with `GS0156` |
| cross, no `/r:` | bare `gsc`, `gsc /out:`, `gsi` | 3/3 reject with `GS0157` |

`gsi` has no `/r:` option. Nullable narrowing covers `EnabledText`, `DisabledText`, and nullable literal `NullableConst`, in evaluate and emit modes. Tests are split into 38 independent cases.

## Mutation

Product site: `ExpressionBinder.Access.MemberLookup.cs:267`.

| Run | Change | Corpus |
|---|---|---:|
| clean | `ClrNullability.GetFieldTypeSymbol(sf)` | GREEN 38/38 |
| mutant A | exact #3112 revert: `TypeSymbol.FromClrType(sf.FieldType)` | RED: 6 failed, 32 passed |
| control B | `TypeSymbol.Error` | RED: 30 failed, 8 passed |
| restored | clean expression | GREEN 38/38 |

Target was unique. Mutants were verified by source diff and hash movement. Clean Core artifact MD5 `672008a079cbe3e4bf621b1f639cafd2`; A `3e9fb1c71a9da9f30a59e653d8c91622`; B `a440919266a6fbb2a358f1da5394f1e6`; restored exactly to clean MD5.

## Validation

All **9 of 9 test projects enumerated from `GSharp.sln`** ran unfiltered:

- Compiler.Tests: 4,232 passed
- Core.Tests: 7,116 passed, 1 skipped
- InternalAnalyzers.Tests: 7 passed
- Interpreter.Tests: 1,258 passed
- Sdk.Tests: 59 passed
- Cs2Gs.Tests: 1,903 passed
- GSharp.GeneratorHost.Tests: 31 passed

Total: **15,172 passed, 1 skipped, 0 failed**. `dotnet build GSharp.sln -c Debug`: **0 warnings, 0 errors**.


DavidObando:
## NOT MERGEABLE — the corpus cannot detect a nullability regression, which is the axis #3119 was filed to falsify

Anchors: `origin/main` **`2dea578b`** (verified current — merge base equals it, branch genuinely up to date), head **`fae49719`**. Own worktree from a tree with **no `out/`**; `dotnet build GSharp.sln -c Release` → `BUILD_RC=0`, **0 errors, 0 warnings**, matching the PR's claim. Corpus baseline **1/1 PASS**.

Credit first: this is careful work. Empty cells are asserted (`:334`, `:344`), the emit column is a **real child process** via `dotnet exec` (`:285-299`), `AssertAssemblyLoads` uses a collectible ALC + `GetTypes()` (`:313-326`), the row values are guarded by genuine comparisons (`if Constants.Empty == "" → "empty-401"` else `"bad-empty"`), and the csproj wiring at `Interpreter.Tests.csproj:14-15` correctly uses `ReferenceOutputAssembly="false"` for the ordering-only fixture. The population genuinely grows 3 → 14.

### The corpus IS a real instrument — I proved that before criticising it

Two PRODUCT mutants at the exact line #3119 names, `ExpressionBinder.Access.MemberLookup.cs:267`, each asserted applied (1 insertion / 1 deletion):

| mutant | change at `:267` | corpus |
|---|---|---|
| **B** (visibility control) | `ClrNullability.GetFieldTypeSymbol(sf)` → `TypeSymbol.Error` | **RED**, `TEST_RC=1` |
| **A** (exact #3112 revert) | → `TypeSymbol.FromClrType(sf.FieldType)` | **GREEN, survived** |

Mutant B killing the corpus establishes that mutations propagate to the binary and that the corpus really does watch `:267` — so **A's survival is a genuine SURVIVED, not a no-result**. Taken alone, A surviving is the good news #3119 wanted: the widening is behaviourally inert across all 14 specimens.

### BLOCKER — but A survives for the wrong reason, and I measured it

Mutant A *is* the nullability drop. Under it, imported fields lose their `?`. That is a real, user-visible behaviour change — not a no-op:

```
var s string = Constants.EnabledText     // string? field -> non-nullable string
```
| build | real `gsc /out: /r:<cross fixture>` |
|---|---|
| clean `fae49719` | **rc=1** — `GS0156: Cannot convert type 'string?' to 'string'` |
| mutant A | **rc=0** — `Wrote nn2.dll` |

So dropping nullability silently makes the compiler **accept a program it should reject**, and the corpus stays green through it.

The cause is the assertion *direction*. Rows 410–413 do:
```
var enabledText string? = Constants.EnabledText
```
`string` → `string?` is a **widening** assignment. It succeeds whether the field binds as `string` or `string?`, so these rows are insensitive to exactly the property they exist to protect. #3119's thesis was *"three passing probes on non-nullable value types say nothing about a nullability-aware code path."* The specimens are now present, but the assertions remain nullability-blind, so **the question is still unfalsified** — for a new reason.

**Fix (small):** add negative rows asserting the *narrowing* is rejected — assign `Constants.EnabledText`/`DisabledText` to a non-nullable `string` and require a `GS0156` diagnostic, mirroring the existing no-`/r:` negative control at `:123-129`. The program and expected diagnostic above are ready to paste.

### SHOULD-FIX — the same-assembly half of the matrix is unreproducible by any real driver

`RunCompiler`/`RunInterpreter` (`:220-224`) call `GSharp.Compiler.Program.Main` and `GSharp.Repl.Program.Main` **in-process**, and `RunSameAssemblyMatrix` passes **no `/r:`**. The fixture resolves only because `Interpreter.Tests.csproj:14` loads it into the test host. Measured with the shipped binaries on the same source:

| real invocation | result |
|---|---|
| `gsc` (no `/r:`) — the evaluate cell | **rc=1** `GS0157: Cannot find type Constants` |
| `gsc /target:exe /out:` (no `/r:`) — the **emit oracle's** compile step | **rc=1** `GS0157` |
| `gsc /r:<fixture>` | **rc=1** `GS9999` MetadataLoadContext |
| `gsi` | **rc=1** `GS0157` — and `gsi` has no `/r:` at all |

All three "loaded/same" cells the PR reports as 14/14 describe a configuration no user can create, and they bind through host reflection rather than the `MetadataLoadContext` a real driver uses — a distinction that demonstrably changes results (the `GS9999` row above). This is the #3130 pattern in new code. The cross-assembly half is sound: it uses real `/r:`, and cross-emit covers all 14 rows including the nullables.

Recommend either driving the same-assembly cells through real child processes with `/r:` (as `SampleConformanceTests` now does after #3077), or relabelling them honestly as in-process binder coverage rather than driver columns.

### SHOULD-FIX — the nullability × `IsLiteral` cell is absent

The const rows return through `BoundLiteralExpression` (`:302`); the nullable rows are `static readonly`, so they return through `BoundClrPropertyAccessExpression` (`:305`). The literal branch therefore never sees a nullable type — yet #3119 is about the **constant** path specifically. One line closes it, and I verified it compiles (`BUILD_RC=0`, 0 errors):
```csharp
public const string? NullableConst = "nullable-const-414";
```

### NIT — the PR body's "Exact values" list is not values
`empty-401`, `escaped-402`, `char-403`, `bool-404` are row IDs; the actual constants are `""`, `"A\n\"B\"\\C"`, `'Q'`, `true`. A reader checking the fixture against that list will think it mismatches.

### NIT — the whole 14×6 matrix is a single `[Fact]`
First failing assertion masks every later row, so a regression reports one cell instead of a profile. A `[Theory]` per row would localise.

---

**Verdict: NOT MERGEABLE** on the BLOCKER alone. To be explicit about the review question — *census or sample?* It is a **census of specimens** (every axis #3119 listed is present, bar the literal-nullable cell) but a **sample of assertions**: value round-trip is asserted well, nullability is not asserted at all. And on *"could this corpus detect the breakage it was built to look for?"* — for value and type identity, **yes, proven by mutant B**; for nullability, **no, proven by mutant A plus the two-sided `GS0156` probe**. The fix is a few lines, not a redesign.

DavidObando:
## Round 2 — **NOT MERGEABLE**. The round-1 BLOCKER is unfixed: mutant A still survives.

Anchors: `main` **`faca348a`** (fetched; the SHA I was briefed with was already stale), PR head **`fae49719`**, merge base **`2dea578b`**, `merge-tree --write-tree` **output** a single line — tree `04715d70`, **0 conflict lines** — and that tree is what I built. `BUILD_RC=0`, 0 warnings, at load 8.49–13.46. Scope confirmed three-dot `--full-index`: **7 files, +595/−0**, no `src/` change.

### Acceptance test: does the corpus kill mutant A? **No.**

Mutation site `src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs:267` — `FieldInfo sf => ClrNullability.GetFieldTypeSymbol(sf),`. Target asserted **unique** in-file before mutating (the other `GetFieldTypeSymbol` at `:672` is the instance-field path and was not touched). Every mutation verified applied by `git diff` **and** md5 movement, and restored by exact md5 round-trip (`17f270b52b18454afb40125ceff4ab23`).

| run | `:267` | corpus | 
|---|---|---|
| clean | `GetFieldTypeSymbol(sf)` | **PASS** 1/1 |
| **mutant A** (#3112 revert) | `TypeSymbol.FromClrType(sf.FieldType)` | **PASS 1/1 — SURVIVES** |
| **control B** | `TypeSymbol.Error` | **FAIL rc=1** (`Issue3119ImportedConstantCorpusTests.cs:259`) |
| restored | original | **PASS** 1/1 |

**Control B is the part that makes this decisive.** It goes red, so mutations reach the binary and the corpus genuinely watches `:267`. Without B, mutant A's green would be an ambiguous no-result, and **a no-result is not a survival**. With B red, A's green is a real survival. Specimen count 3 → 14 did not move the axis.

### Root cause is unchanged: assertion **direction**

`test/Interpreter.Tests/Issue3119ImportedConstantCorpusTests.cs:188-191` still assigns into nullable targets:

```
var enabledText string? = Constants.EnabledText
var enabledInt int32? = Constants.EnabledInt
var disabledText string? = Constants.DisabledText
var disabledInt int32? = Constants.DisabledInt
```

That is a **widening**. `string` → `string?` is legal whether or not the binder preserved the annotation, so the cell cannot discriminate. Measured 2×2 with real `gsc` `/out:` (emit column, `/r:` against the PR's own `GSharp.Issue3119.Cross.dll`), `rc` captured before any pipe, each probe in its own asserted-empty directory:

| assertion | clean | mutant A |
|---|---|---|
| **widening** (`var x string? = …`, what the corpus does) | rc=0 | rc=0 — **blind** |
| **narrowing** (`var x string = …`, what it does not do) | **rc=1 `GS0156`**, no dll | **rc=0, dll written** — **kills** |

Under mutant A the compiler **accepts a program it should reject and emits the assembly**. Verbatim on clean:

```
narrow.gs(5,16,5,37): error GS0156: Cannot convert type 'string?' to 'string'.
```

Under mutant A: `rc=0`, zero `GS0156`, `narrow.dll` written (2560 bytes). One narrowing line is the whole fix.

### Your three questions

**1. Are the new assertions narrowing? No.** Every nullability specimen is widening (`:188-191`). `CreateAssertionSource` (`:203-218`) never mentions `EnabledText`/`DisabledText`/`EnabledInt`/`DisabledInt` at all — the four nullable specimens have **no** throw-assertion, only the `??`-defaulted `Console.WriteLine` at `:192-195`, and `??` on a target already declared `string?` is insensitive to the mutant.

**2. Does the count grow without the axis being covered? Yes — a census, not coverage.** 14 specimens are carried by **one** `[Fact]` (`ImportedConstants_PreserveValuesAcrossSameAndCrossAssemblyDrivers`, `:71`); the file contains exactly 1 test attribute. The 6 driver cells inside it (`:95-113` same-assembly bare/emit/gsi, `:123-154` cross-assembly) multiply *drivers*, not *directions*. Ten of the fourteen specimens are non-nullable by construction and cannot exercise nullability at all.

**3. Is the nullable × `IsLiteral` cell populated? No — and it is a real, reachable, mutant-sensitive cell.** Structural reason, from the fixture: in `test-assets/Issue3119.ImportedConstants/Constants.cs`, every `const` (`:19` `Empty`, `:22` `Escaped`, `:25` `Character`, `:28` `Boolean`, `:31` `Decimal`, `:34`/`:37` enums, `:40` `PositiveControl`) is **non-nullable**, and every nullable member (`:43` `EnabledText`, `:46` `EnabledInt`, `:52` `DisabledText`, `:55` `DisabledInt`) is **`static readonly`**. So consts take the literal branch at `:302`/`:304` and nullables take `BoundClrPropertyAccessExpression` at `:307` — the two sets never intersect, which is exactly why nullable × literal has no cell.

I measured that the cell is not merely unpopulated but **live**. Built a standalone fixture with `public const string? NullableConst = "nc-500";` (legal C#, `<Nullable>enable</Nullable>`, builds 0 warnings) and narrowed it:

- clean → **rc=1 `GS0156`**, no dll
- mutant A → **rc=0, dll written**

Both branches — `:304` literal and `:307` property-access — consume the same `staticType` computed at `:267`, so **two** distinct narrowing cells are missing, not one.

### Verdict

**NOT MERGEABLE.**

- **BLOCKER** — `test/Interpreter.Tests/Issue3119ImportedConstantCorpusTests.cs:188-191`: the nullability specimens assert in the widening direction and therefore cannot fail under the #3112 revert. Mutant A survives (proven live by control B going red). Fix: add narrowing assertions that expect a **diagnostic**, e.g. a cell asserting `var s string = Constants.EnabledText` yields rc≠0 with `GS0156`. Add the same for a `const string?` specimen so the `:304` literal branch is covered too — that requires adding a nullable `const` to the fixture, since none exists.
- **SHOULD-FIX** — same file, `:71`: 14 specimens under a single `[Fact]` means the first failure masks the rest and no specimen is individually named in output. A `[Theory]` over specimens would make which cell regressed legible.
- **NIT** — `test-assets/Issue3119.CrossConstants/Constants.cs:50,53` document `DisabledText`/`DisabledInt` as being "under disabled nullable context", but `Issue3119.CrossConstants.csproj:7` sets `<Nullable>enable</Nullable>` for the whole project. If a `#nullable disable` region is intended, it is not visible; if not, the "disabled" specimens are duplicates of the "enabled" ones and the axis they claim to cover is absent.

### Scope and denominator, stated plainly

**Ran 1 of 9 test projects, filtered to 1 test.** Denominator source is `GSharp.sln` via `dotnet sln list`, which is CI's own denominator per `build/generate-ci-test-matrix.py:26`. That narrow scope is the correct and sufficient scope for **this** acceptance test — the question was whether *this corpus* kills mutant A — and it is **not** evidence about anything else. I make no coverage claim beyond it. `Sdk.Tests` **NOT RUN**. No full-suite run was warranted: the diff is test-only with **no `src/` change**, so there is no product-regression or false-positive surface to sweep.

One thing I checked and found benign: the PR adds two projects to `GSharp.sln`, but `generate-ci-test-matrix.py:33` selects only paths ending `.tests.csproj`, and these are `Issue3119.ImportedConstants.csproj` / `Issue3119.CrossConstants.csproj`. The CI matrix is unaffected and the 9-project denominator does not change.

All mutations restored by md5 round-trip; `git status --porcelain` empty in the worktree and in the shared checkout, which I did not write to.

DavidObando:
## Round 3 — **MERGEABLE**. The two-round BLOCKER is genuinely fixed. I reproduced it independently and then went further than the author did.

Anchors: `origin/main` **`387cb9a2`**, head **`398e9cab`**, merge base **`61fe9cb9`** (main ahead by 3). `git merge-tree --write-tree` **output** = `30dc3467`, **0 conflict markers**; co-merging in a fresh worktree reproduced that tree hash exactly. Clean build rc0, **0 warnings, 0 errors**.

---

## 1. The BLOCKER — **FIXED**, reproduced from scratch

I did not take the author's table. Every row below is mine, each mutant asserted applied by `git diff --numstat` **and** md5 movement on both the source and `GSharp.Core.dll`.

| run | source md5 | Core.dll md5 | corpus |
|---|---|---|---|
| pristine | `17f270b5` | `8906913b` | **38/38 PASS** |
| **mutant A** (`:267` → `TypeSymbol.FromClrType(sf.FieldType)`, the pre-#3112 behaviour) | `3f7ddb2c` | `0ec9c1fc` | **RED — 6 failed / 32 passed** |
| **control B** (`:267` → `TypeSymbol.Error`) | moved | moved | **RED — 30 failed / 8 passed** |
| restored | `17f270b5` ✓ | `8906913b` ✓ | **38/38 PASS** |

Both of the author's headline numbers — 6 and 30 — reproduce exactly, and the restore is a byte-exact round-trip on source *and* compiled output. The six failures under mutant A are precisely the six narrowing cells, nothing else.

---

## 2. Both consumers — **`:304` and `:307` are independently watched**. This is the part no prior round established.

Round 2 correctly noted that `:304` and `:307` both consume the same `staticType` from `:267`, so a single mutant at `:267` cannot prove *both* are covered — it only proves at least one is. So I split the mutation per consumer:

| mutant | site | what it isolates | failures | which cells |
|---|---|---|---|---|
| **A-304** | `:304` `BoundLiteralExpression(..., staticType)` | **const/literal** path | **2** | `NullableConst` × {evaluate, emit} |
| **A-307** | `:307` `BoundClrPropertyAccessExpression(..., staticType)` | **static readonly** path | **4** | `EnabledText`, `DisabledText` × {evaluate, emit} |
| A (both) | `:267` | shared source | **6** | all of the above |

**2 + 4 = 6, disjoint, exact.** Each consumer is watched by a distinct specimen set, no cell is redundant, and no cell is missing. `NullableConst` is the *only* specimen that can reach `:304`, which is exactly why it had to be added.

## 3. `NullableConst` is real

`test-assets/Issue3119.CrossConstants/Constants.cs:33` — `public const string? NullableConst = "nullable-const-414";`. Genuinely `const` (reflection confirms `isLiteral=True`) **and** genuinely nullable. The nullable × literal cell that previously **could not exist by construction** now exists. Not a renamed hole.

## 4. The `[Fact]` at `:71` is split

Five `[Theory]`+`[MemberData]` groups (`:74`, `:100`, `:133`, `:171`, `:204`) plus one `[Fact]` (`:243`) for identity only. Exact case accounting: 10 evaluation + 15 emission + 6 narrowing + 3 no-reference + 3 runtime + 1 = **38**. First-failure masking is gone — which is why mutant A reports 6 discrete cells instead of one.

## 5. Assertion direction is now correct, verified out-of-process

The committed assertion is `ExitCode == 1` + `GS0156` + `'string?' to 'string'` (`:161-163`). Independently, with real `gsc` and `/r:`:

```
narrowing, evaluate : rc=1  GS0156: Cannot convert type 'string?' to 'string'
narrowing, emit     : rc=1  GS0156   — no dll written
widening,  emit     : rc=0           — dll written
```
The widening direction (rc=0 either way) is what made rounds 1–2 blind; the corpus now asserts the direction that discriminates.

---

## What I could not fault, having tried

**Excluding `gsi` from `NullableNarrowingCases`** looked like an asymmetry — `gsi` appears in every other driver set. It is **structurally forced**, not an oversight: `RunInterpreter` (`:293-294`) takes no reference argument, and there is **no `/r:` handling anywhere in `src/Repl/`**. A real `gsi` invocation confirms it: `Unrecognized argument: /r:...`. `gsi` has no reference channel, so it cannot see an imported constant at all.

**Omitting `EnabledInt`/`DisabledInt` from narrowing** also looked like a gap — `int?`→`int32` narrowing *is* rejected (rc=1 `GS0156`), so they are valid cells. I measured whether they could ever detect this regression. Under mutant A:

```
narrowed int32 = Constants.EnabledInt   -> rc=1  (still rejected)
narrowed int32 = Constants.DisabledInt  -> rc=1  (still rejected)
narrowed string = Constants.EnabledText -> rc=0  (ACCEPTED — the regression)
```
Value-type nullability goes through `TypeSymbol.FromClrType`, not `ApplyReferenceNullabilityFull`, so those cells are **provably incapable** of detecting #3112 regressions. Omitting them is correct. The three chosen specimens are not a sample of the nullable fields — they are a **census of the detecting ones**.

---

## NIT — the round-2 fixture NIT is **not resolved**, and the real defect is different from the one reported

Round 2 flagged `Constants.cs:50,53` as a comment claiming a "disabled nullable context" contradicted by `.csproj:7` `<Nullable>enable</Nullable>`. The diagnosis was too generous. `#nullable disable` **is** present at `:50`, so the comment is not contradicted by the csproj — but the specimen it describes does not exist. Reflected metadata from the built fixture:

```
TYPE Constants  NullableContext = 2          (annotated is the type-wide default)
EnabledText     Nullable=none  -> inherits 2
DisabledText    Nullable=none  -> inherits 2      <-- identical to EnabledText
NullableConst   Nullable=none  -> inherits 2
Empty           Nullable=1                        <-- proves per-field overrides ARE emitted
```

`Empty` carrying an explicit `NullableAttribute=1` proves Roslyn emits a per-field override whenever a member differs from the type default. `DisabledText` has none, so it is **byte-identical to `EnabledText`**. Cause: writing `string?` *with* the `?` under `#nullable disable` keeps the annotation (that is exactly what the suppressed `CS8632` is warning about). To get an oblivious specimen you must write `string` **without** the `?`.

Corroborating: `DisabledText` and `EnabledText` failed together in 100% of mutants and never once diverged.

**Severity is NIT, not higher, and I checked rather than assumed.** I built a genuinely oblivious assembly (`<Nullable>disable</Nullable>`, no `?`, confirmed `TYPE NullableContext=none`) and ran it through `gsc`:

```
narrowed string = OblConstants.ObliviousConst -> rc=1 GS0156
narrowed string = OblConstants.ObliviousField -> rc=1 GS0156
```
Identical to the annotated specimens. G# maps oblivious to nullable today, so the missing shape would not change any result. **The specimen is redundant, not load-bearing.** Fix is either direction: drop `DisabledText`/`DisabledInt` and their two comments as redundant (smaller corpus, same power), or drop the `?` inside the `#nullable disable` region to make the distinction real.

## NIT — all three "drivers" are in-process

`RunCompiler` → `GSharp.Compiler.Program.Main` (`:290`), `RunInterpreter` → `GSharp.Repl.Program.Main` (`:294`). Only `RunAssembly` (`:340`) spawns a real process, and only to *execute* the emitted assembly. This does not weaken the corpus — my mutants killed it, and the fixture is force-built via the correct `ReferenceOutputAssembly="false"` idiom at `Interpreter.Tests.csproj:14`, matching `Cs2Gs.Tests.csproj:31` — but "three drivers" in the PR description means three code paths, not three binaries. Worth one word in the description.

*(Method note, self-disclosed: my first out-of-process probe used `grep ... | head -1` and showed only `GS9100`, which reads as "narrowing does not fire". `GS9100` merely sorts first. A first-line grep is a sample, not a census — the same error class this PR exists to fix.)*

---

## The stale-head claim — **not supported**

The PR states round 2 measured a stale head. Measured, in UTC:

```
11:51:03Z  fae49719 committed  ("Build constant fixtures in Release")
13:15:33Z  round-1 verdict
18:10:03Z  round-2 verdict
18:29:35Z  398e9cab committed  ("Strengthen imported constant regression corpus")
```
`fae49719` was the head continuously from 11:51Z until the rebase at 18:29Z. Round 2 posted at 18:10Z — inside that window, against the then-current head. `git cat-file` confirms `fae49719` contains the corpus file with **zero** `narrowed string` assertions. The fix landed **19.5 minutes after** the verdict it is said to have missed.

**Round 2 measured correctly; its verdict stands in full, and this PR is the response to it.** Worth correcting in the description — the record matters more than the attribution.

---

## Suites — 9 of 9, enumerated from `dotnet sln GSharp.sln list`

| project | result |
|---|---|
| Compiler.Tests | 4242 passed |
| Core.Tests | 7116 passed, **1 skipped** |
| Interpreter.Tests | **1258** passed (baseline 1220 + the 38 new) |
| Cs2Gs.Tests | 1903 passed |
| Sdk.Tests | 59 passed |
| GeneratorHost.Tests | 31 passed |

**15,182 passed, 1 skipped, 0 failures.** Higher than the reported 15,172 because main is 3 commits ahead of the base; consistent, not a discrepancy. Freshness (epoch mtimes, `sort -n`, excluding `ref/`, across `Core`/`Compiler`/`Repl`): newest first-party artifact `1785794574`, ~25 min before measurement; the 2021 outlier is a NuGet `Newtonsoft.Json` copy, not staleness.

---

### Verdict: **MERGEABLE**. No blockers, two NITs, neither gating.

The corpus now does what #3119 asked: it detects the regression it was built to look for, at both consumers of the widened path, in the one assertion direction that discriminates. The specimen-vs-assertion distinction that sank rounds 1 and 2 is closed — and the narrowing specimen set is a genuine census of the detecting fields, which I confirmed by proving the excluded ones cannot detect anything.

PR #3127, opened 2026-08-03
Fix implicit BCL dispatch for object overrides

### LINKED ISSUE #3116: Evaluator ignores struct GetHashCode/Equals overrides when the BCL calls them implicitly, so Dictionary.Count differs by driver
Follow-up from PR #3071 (issue #2896, "Plain struct cannot override Object virtual methods"), which is merged. Found by rubber-duck review, not by the sample gates.

## Symptom

A struct that overrides `GetHashCode` and `Equals` is used as a `Dictionary` key. The two drivers disagree on how many entries the dictionary holds.

```gsharp
package P
import System

type Val struct {
    n int32
}

func (v Val) GetHashCode() int32 {
    return 7
}

func (v Val) Equals(other object) bool {
    return true
}

var d = Dictionary[Val, string]()
d[Val{n: 11}] = "a"
d[Val{n: 22}] = "b"
Console.WriteLine(d.Count)
```

| driver | `d.Count` |
|---|---|
| bare `gsc` (evaluates) | **2** |
| `gsi` (interprets) | **2** |
| `gsc /out:` (emits) | **1** |

`1` is correct. `GetHashCode` returns a constant and `Equals` returns `true` unconditionally, so the two keys **must** collide and the second write must overwrite the first. The emitter honors the overrides; the evaluator does not.

## Why this matters more than the display-text sibling

This is not a formatting difference. It is a **differing container element count**, silently, at rc0, with no diagnostic. Anything downstream that iterates or counts diverges by driver.

The mechanism is that the BCL calls these overrides **implicitly**. Calling `a.Equals(b)` by name works correctly on all three drivers — verified. The failure is confined to call sites inside `Dictionary`/`HashSet`, which is exactly where a user would not think to look.

## Not a regression

On the pre-#3071 base this program does not compile at all (`GS0183`), so nothing that used to work has broken. #3071 delivered the capability in the emitter and only partly in the evaluator. That is why this is a follow-up rather than something that should have blocked the merge.

## Scope

- `Dictionary[Val, string]` and `HashSet[Val]`, and any other BCL surface that calls `GetHashCode`/`Equals` implicitly.
- `ToString` has a related and milder twin: `String.Format`/interpolation on a struct with an overridden `ToString` differs in display text between the evaluate and emit columns. Same root shape — implicit BCL invocation of an override — so a fix should cover both, and the corpus should include both.
- Also worth pinning while here: #3071's override walk is tested at depth 3 and correct at depth 4 (measured), but nothing pins depth 4 against a future regression.

## Suggested corpus

Both `struct` and `class` receivers; boxed and unboxed; `Dictionary` key, `HashSet` element, `List.Contains`, `String.Format`; each across all three drivers with distinct printed values per level so a wrong pick prints a wrong number rather than silently agreeing.

DavidObando:
## Summary

- Attach one evaluator override dispatcher when `StructValue` instances are created.
- Route CLR `Equals`, `GetHashCode`, and `ToString` calls back through the existing transitive external-override lookup.
- Preserve dispatch through value copies and explicit-layout/nested-struct materialization.
- Add `{top-level, function} × {gsc evaluate, gsc emit, gsi}` coverage for `Dictionary`, `HashSet`, `List.Contains`, formatting/interpolation, plus a depth-four override chain.

## Root cause

Explicit G# calls already passed through evaluator override dispatch. BCL code instead received `StructValue` and invoked its structural CLR overrides directly, bypassing the G# method body. All five evaluator construction roots now use one factory; `StructValue` is the single BCL boundary.

Compiler census found 10 construction sites: 8 product sites (5 evaluator roots, 3 copy/layout paths) and 2 direct unit-test sites.

## Behavior

Assignment-base `main` (`60076cc3`):

| Surface | gsc evaluate | gsc emit | gsi |
|---|---:|---:|---:|
| Dictionary, top/function | 2 | 1 | 2 |
| HashSet, top/function | 2 | 1 | 2 |
| List.Contains, top/function | False | True | False |
| Format/interpolation, top/function | structural text | override text | structural text |

Head: every cell matches emit (`1`, `1`, `True`, override text). All 24 surface cells and all 6 depth-four cells pass with empty stderr; emitted cases load bytes, enumerate `GetTypes()`, and execute in child processes.

## Validation

- Release solution build: `BUILD_RC=0`, 0 warnings, 0 errors.
- Live-main merge prediction at `4469e323`: tree `9226992ad031d02f21cb9af96bd24ded548d75d9`, exactly equal to HEAD tree; merged-tree build `BUILD_RC=0`.
- Unfiltered test projects:
  - InternalAnalyzers.Tests: 7 passed
  - Interpreter.Tests: 744 passed
  - Core.Tests: 7089 passed, 1 skipped
  - Compiler.Tests: 4152 passed
  - Sdk.Tests: 57 passed
- Diagnostic documentation gate: 1 passed.
- cs2gs corpus gate: rc 0; 19/20 green, 1 known ledgered gap, 0 new, 0 regressed, 0 stale.
- `cs2gs-oahu` was not run locally because its script executes `Oahu.Cli.E2E.Tests`; the required CI check remains authoritative.

Mutation checks, each asserted applied and round-tripped by SHA-256:

- Invert `StructValue` dispatcher-null guard: `ImplicitBclCalls_DispatchStructObjectOverridesAcrossDrivers` / depth-four tests fail 20/30 cells; 10 emit cells survive.
- Invert evaluator external-override guard: same 20/30 evaluator/interpreter cells fail.
- Change transitive root walk `while` to `if`: `ImplicitBclToString_DepthFourOverrideChain_UsesMostDerivedOverrideAcrossDrivers` fails exactly 4/6 evaluator/interpreter cells; 2 emit cells survive.

Fixes #3116

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -1003,42 +1003,16 @@ public sealed partial class Evaluator
 
         if (receiver is StructValue sv)
         {
-            static MethodInfo FindExternalOverrideRoot(FunctionSymbol method)
-            {
-                while (method.OverriddenMethod != null)
-                {
-                    method = method.OverriddenMethod;
-                }
-
-                return method.ExternalOverriddenMethod;
-            }
-
-            FunctionSymbol externalOverride = null;
-            for (var type = sv.StructType; type != null && externalOverride == null; type = type.BaseClass)
-            {
-                externalOverride = type.Methods.FirstOrDefault(method =>
-                    method.IsOverride && FindExternalOverrideRoot(method)?.Equals(node.Method) == true);
-            }
-
+            var externalOverride = FindExternalOverride(sv.StructType, node.Method);
             if (externalOverride != null)
             {
-                var frame = new ConcurrentDictionary<VariableSymbol, object>
-                {
-                    [externalOverride.ThisParameter] = receiver,
-                };
-
-                var parameterOffset = externalOverride.ExplicitReceiverParameter == null ? 0 : 1;
+                var arguments = new object[node.Arguments.Length];
                 for (var i = 0; i < node.Arguments.Length; i++)
                 {
-                    var parameter = externalOverride.Parameters[i + parameterOffset];
-                    frame[parameter] = EvaluateExpression(node.Arguments[i]);
+                    arguments[i] = EvaluateExpression(node.Arguments[i]);
                 }
 
-                using (PushFrame(frame))
-                {
-                    var statement = program.Functions[externalOverride];
-                    return EvaluateUserMethodBody(externalOverride, statement);
-                }
+                return EvaluateExternalOverride(sv, externalOverride, arguments);
             }
         }
 
@@ -1081,6 +1055,71 @@ public sealed partial class Evaluator
         var result = InvokeReflective(method, receiver, args, node);
         WriteBackRefSlots(refSlots, args);
         return result;
+    }
+
+    private StructValue CreateStructValue(StructSymbol structType) =>
+        new(structType, TryEvaluateExternalOverride);
+
+    private bool TryEvaluateExternalOverride(
+        StructValue receiver,
+        MethodInfo method,
+        object[] arguments,
+        out object result)
+    {
+        var externalOverride = FindExternalOverride(receiver.StructType, method);
+        if (externalOverride == null)
+        {
+            result = null;
+            return false;
+        }
+
+        result = EvaluateExternalOverride(receiver, externalOverride, arguments);
+        return true;
+    }
+
+    private object EvaluateExternalOverride(
+        StructValue receiver,
+        FunctionSymbol externalOverride,
+        object[] arguments)
+    {
+        var frame = new ConcurrentDictionary<VariableSymbol, object>
+        {
+            [externalOverride.ThisParameter] = receiver,
+        };
+
+        var parameterOffset = externalOverride.ExplicitReceiverParameter == null ? 0 : 1;
+        for (var i = 0; i < arguments.Length; i++)
+        {
+            frame[externalOverride.Parameters[i + parameterOffset]] = arguments[i];
+        }
+
+        using (PushFrame(frame))
+        {
+            var statement = program.Functions[externalOverride];
+            return EvaluateUserMethodBody(externalOverride, statement);
+        }
+    }
+
+    private static FunctionSymbol FindExternalOverride(StructSymbol type, MethodInfo externalMethod)
+    {
+        FunctionSymbol externalOverride = null;
+        for (; type != null && externalOverride == null; type = type.BaseClass)
+        {
+            externalOverride = type.Methods.FirstOrDefault(method =>
+                method.IsOverride && FindExternalOverrideRoot(method)?.Equals(externalMethod) == true);
+        }
+
+        return externalOverride;
+    }
+
+    private static MethodInfo FindExternalOverrideRoot(FunctionSymbol method)
+    {
+        while (method.OverriddenMethod != null)
+        {
+            method = method.OverriddenMethod;
+        }
+
+        return method.ExternalOverriddenMethod;
     }
 
     // Issue #814 / ADR-0084 §L5: when an open generic extension method is
--- a/src/Core/CodeAnalysis/Evaluator.Concurrency.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Concurrency.cs
@@ -453,7 +453,7 @@ public sealed partial class Evaluator
         return BlockOnValueTask(operand);
     }
 
-    private static object EvaluateDefaultExpression(BoundDefaultExpression node)
+    private object EvaluateDefaultExpression(BoundDefaultExpression node)
     {
         // Issue #148: the `await for` lowering uses `default(CancellationToken)`
         // for the `GetAsyncEnumerator` token; more generally, BoundDefaultExpression
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -675,7 +675,7 @@ public sealed partial class Evaluator
                     node.StructType,
                     ImmutableArray<BoundExpression>.Empty,
                     constructor))
-            : new StructValue(node.StructType);
+            : CreateStructValue(node.StructType);
 
         // Primary-constructor literals use default construction before applying
         // named initializers, matching the emitter's newobj/stfld sequence.
@@ -1009,7 +1009,7 @@ public sealed partial class Evaluator
 
     private object EvaluateConstructorCallExpression(BoundConstructorCallExpression node)
     {
-        var sv = new StructValue(node.StructType);
+        var sv = CreateStructValue(node.StructType);
 
         // Default-initialize all fields first (including inherited), then bind primary-ctor args.
         for (var t = node.StructType; t != null; t = t.BaseClass)
@@ -1906,7 +1906,7 @@ public sealed partial class Evaluator
 
         var current = ResolveReceiverValue(node.ReceiverExpression, node.Receiver);
 
-        var sv = current as StructValue ?? new StructValue(node.StructType);
+        var sv = current as StructValue ?? CreateStructValue(node.StructType);
 
         // Class types are reference types: mutate the existing instance in
         // place so other references observe the write. Structs preserve
@@ -2071,7 +2071,7 @@ public sealed partial class Evaluator
                     ? GetGlobal(receiverVar)
                     : Locals.Peek()[receiverVar];
 
-                var sv = current as StructValue ?? new StructValue(node.StructType);
+                var sv = current as StructValue ?? CreateStructValue(node.StructType);
 
                 if (node.StructType.IsClass)
                 {
@@ -2112,13 +2112,13 @@ public sealed partial class Evaluator
         return value;
     }
 
-    private static object DefaultValue(Symbols.TypeSymbol type)
+    private object DefaultValue(Symbols.TypeSymbol type)
         => GetDefaultValue(type, useLanguageStringZero: true);
 
-    private static object ClrDefaultValue(Symbols.TypeSymbol type)
+    private object ClrDefaultValue(Symbols.TypeSymbol type)
         => GetDefaultValue(type, useLanguageStringZero: false);
 
-    private static object GetDefaultValue(Symbols.TypeSymbol type, bool useLanguageStringZero)
+    private object GetDefaultValue(Symbols.TypeSymbol type, bool useLanguageStringZero)
     {
         // Issue #504/#1652: NullableTypeSymbol.ClrType aliases the underlying
         // type's ClrType (e.g. `int?` reports `typeof(int)`), so it must be
@@ -2156,7 +2156,7 @@ public sealed partial class Evaluator
                 return null;
             }
 
-            var sv = new StructValue(s);
+            var sv = CreateStructValue(s);
             foreach (var f in s.Fields)
             {
                 sv.Fields[f.Name] = GetDefaultValue(f.Type, useLanguageStringZero);
--- a/src/Core/CodeAnalysis/StructValue.cs
+++ b/src/Core/CodeAnalysis/StructValue.cs
@@ -24,31 +24,52 @@ namespace GSharp.Core.CodeAnalysis;
 /// </summary>
 public sealed class StructValue
 {
+    private static readonly MethodInfo ObjectEqualsMethod = typeof(object).GetMethod(
+        nameof(object.Equals),
+        [typeof(object)]);
+
+    private static readonly MethodInfo ObjectGetHashCodeMethod = typeof(object).GetMethod(
+        nameof(object.GetHashCode),
+        Type.EmptyTypes);
+
+    private static readonly MethodInfo ObjectToStringMethod = typeof(object).GetMethod(
+        nameof(object.ToString),
+        Type.EmptyTypes);
+
+    private readonly ObjectOverrideDispatcher objectOverrideDispatcher;
     private readonly LayoutRuntime explicitLayout;
     private object explicitLayoutValue;
 
     /// <summary>Initializes a new instance of the <see cref="StructValue"/> class.</summary>
     /// <param name="structType">The struct type symbol.</param>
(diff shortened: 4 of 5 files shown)


DavidObando:
# Review: PR #3127 — `fix/3116-implicit-bcl-override-dispatch`

**Verdict: MERGEABLE.** No blockers. One SHOULD-FIX (measured performance regression on a hot path), two NITs, and one pre-existing sibling defect I recommend be filed separately.

Reviewed at head **`4ce62e21`**. Merge-base is **`4469e323`**, but `main` is now **`f7d9a5d3`** — main moved twice under this branch, so the merged tree is *not* the PR head. I built a real merge:

- `git merge-tree` **output** (not rc): 1 line, tree `baa83aae…`, **0 CONFLICT lines**.
- Real merge: `MERGE_RC=0`, actual tree `baa83aae…` — **reproduces the prediction exactly**.
- Merged solution build: **`BUILD_RC=0`**, 0 errors. `Core/`, `Compiler/` and `Repl/` all carry the same post-build mtime.
- Round-trip closed: Core md5 `8d3ac1ee…` → 4 mutants → restore → `8d3ac1ee…`.

---

## First, a problem with the issue, not the PR

**Issue #3116's repro cannot be run as written.** It uses `type Val struct {` and unkeyworded fields, both removed by ADR-0078/GS0288:

```
error GS0306: The 'type Name <kind> ...' declaration head has been removed; use the kind keyword first
error GS0288: Field declarations require a 'var' (mutable), 'let' (read-only), or 'const' keyword
```

rc=1 on **both** `f7d9a5d3` and the merged tree. Anyone who runs the reported shape verbatim to check this fix gets a compile failure and could conclude almost anything from it. I reconstructed the repro using the canonical syntax taken from the PR's own tests (`override func … ->`) and state it in full below so my numbers are auditable. Recommend the issue body be corrected.

## The `main` column, reproduced independently

Source used (top-level; an in-function variant is reported below):

```gsharp
package P
import System

struct Val {
    var n int32
    override func GetHashCode() int32 -> 7
    override func Equals(other object) bool -> true
    override func ToString() string -> "TOSTR-44"
}
```

Each surface probed **independently**, each into its own asserted-empty directory, `[gsc evaluate / gsc emit / gsi]`:

| Surface | assertion | base `f7d9a5d3` | merged |
|---|---|---|---|
| `Dictionary` key | `d.Count` | **2 / 1 / 2** | **1 / 1 / 1** |
| `Dictionary` key | `ContainsKey(Val{n:99})` | False / True / False | True / True / True |
| `HashSet` element | `h.Count` | **2 / 1 / 2** | **1 / 1 / 1** |
| `HashSet` element | `Contains(Val{n:99})` | False / True / False | True / True / True |
| `List.Contains` | `Contains` / `IndexOf` | False,−1 / True,0 / False,−1 | True,0 / True,0 / True,0 |
| `String.Format` | `[{0}]` | `[Val(n=11)]` / `[TOSTR-44]` / `[Val(n=11)]` | `[TOSTR-44]` ×3 |

The author's claimed columns reproduce exactly. Per the standing bar I asserted **contents**, not only counts — `ContainsKey`/`Contains`/`IndexOf` flip too, so this is not a count-only coincidence. The in-function variant gives the same result (`2/False/[Val(n=11)]` → `1/True/[TOSTR-44]`), so `{top-level, in-function}` is confirmed as an axis rather than a case.

**The discriminating fact holds.** By-name `Val{n:11}.ToString()` prints `TOSTR-44` on **base**, on all three drivers. The defect really is confined to implicit BCL invocation, and the fix is placed at exactly that boundary — `StructValue.Equals`/`GetHashCode`/`ToString`, which is the object the BCL actually receives.

**All four surfaces are fixed by one mechanism**, not four: a single dispatcher delegate injected at a single factory (`Evaluator.Calls.cs:1060 CreateStructValue`), consulted by the three CLR overrides. So there is nothing here that should have been split into a separate PR.

**`FindExternalOverrideRoot` is genuinely reused, not duplicated.** The old inline `static` local walk inside `Evaluator.Calls.cs` is **deleted**, and both the by-name path and the new BCL path now route through one `FindExternalOverride` (`Evaluator.Calls.cs:1100`) → `FindExternalOverrideRoot` (`:1113`). There is no second walk.

## Census — verified by the compiler, not by grep

`Evaluator` is partial, so I did not trust grep. I added a required parameter to the public `StructValue(StructSymbol)` constructor and read **raw** build output:

```
test/Core.Tests/CodeAnalysis/Binding/DataStructTests.cs(157,21): error CS1729: 'StructValue' does not contain a constructor that takes 1 arguments
test/Core.Tests/CodeAnalysis/Binding/DataStructTests.cs(160,21): error CS1729: ...
```

Exactly **2** callers of the dispatcher-less path solution-wide, both test sites; **zero product sites**. That confirms the author's "8 product + 2 test = 10, all product paths centralized" independently. (Two earlier census attempts died inside `Core` on `SA1202` and `CS1573` before reaching the test projects — those were **no-results, not zeros**, and I did not read them as absence.)

## What the 10 non-killing cells were

This was the headline question, and the answer is clean: **they are the emit-column cells, and they are supposed to survive.**

M1 (`CreateStructValue` injects `null`) fails **20** cells, and the failing set is *exactly* every `gsc-evaluate` and `gsi` cell:

- 24 surface cells = 4 surfaces × 2 scopes × 3 drivers → 16 fail (4 × 2 × **2** non-emit drivers)
- 6 depth-four cells = 2 scopes × 3 drivers → 4 fail (2 × **2**)
- 20 fail, **10 survive = the 8 surface emit cells + 2 depth-four emit cells**

The PR touches no emitter file, so an evaluator mutation *must not* affect the emit column. Those 10 are a negative control, not unpinned hunks — if any had died, the PR would be reaching into the emitter. Same for the `while`→`if` mutant: "4/6, 2 emit cells survive" is the same structure. This is one mutant evaluated over 30 cells, not 30 mutants.

## My own mutation, per hunk, each asserted applied

| # | Product mutant | `BUILD_RC` | Result |
|---|---|---|---|
| M1 | `CreateStructValue` → `new(structType, null)` | 0 | 733/753; **20 fail**, exactly the evaluate+`gsi` cells |
| M2 | `FindExternalOverrideRoot`: `while` → `if` | 0 | 745/753; **8 fail** = 4 depth-four evaluate/`gsi` + 4 pre-existing `ClassObjectOverrideChain` |
| M3 | delete **only** the `ToString` dispatch arm | 0 | 745/753; **8 fail** = `format` (4) + depth-four (4); dictionary/hashSet/listContains **pass** |
| M4 | delete **only** the `GetHashCode` dispatch arm | 0 | 745/753; **8 fail** = dictionary (4) + hashSet (4); listContains and format **pass** |

M3 and M4 matter because the three arms could have masked one another. They do not: M4's survivors are precisely `List.Contains` and `String.Format`, because `List.Contains` goes through `EqualityComparer<T>.Default.Equals` and never calls `GetHashCode`. Each arm is independently observable and independently pinned.

**Depth-4 is genuinely pinned** (M2): a walk patched to "go one more step" fails `ImplicitBclToString_DepthFourOverrideChain_…` on all 4 non-emit cells.

## False-positive corpus — 0 regressions

The stated risk was routing `GetHashCode`/`Equals`/`ToString` back through an override walk for **every** G# value handed to the BCL. Six programs that work today, all three drivers, base vs merged:

| Corpus file | Covers | base vs merged |
|---|---|---|
| `fp1_nooverride` | struct **without** overrides as Dictionary key, HashSet, `List.Contains`, `ToString`, `String.Format` | **identical** — default value equality preserved (`2/2/c/1/True/False`) |
| `fp2_partial` | struct overriding **only** `GetHashCode`, no `Equals` | **identical** (`2/2/b` on all three drivers) |
| `fp3_builtinkeys` | `Dictionary[string,…]`, `Dictionary[int32,…]`, `List.Contains`/`IndexOf` on `string` | **identical** |
| `fp4_class` | class **with** and **without** overrides | one cell changed — see below |
| `fp5_sort` | `List.Sort` on `int32` and `string`, `String.Format` on builtins | **identical** |
| `fp6_nested` | nested struct fields as Dictionary key | **identical** |

The single changed cell is an **improvement**: `String.Format` on a **class** with a `ToString` override goes `[OverC(n=0)]` → `[CLS-44]` on evaluate and `gsi`, matching the emit oracle. That is an additional fix the PR does not claim.

## SHOULD-FIX — measured performance regression on a hot path

`TryInvokeObjectOverride` runs `FindExternalOverride` on **every** `Equals`, `GetHashCode` and `ToString` call, whenever a dispatcher is attached — which is now always. That walk is `for (type = …; type != null; type = type.BaseClass) type.Methods.FirstOrDefault(m => m.IsOverride && FindExternalOverrideRoot(m)?.Equals(externalMethod) == true)`: a linear scan of every method of every type in the base chain, with a nested `while` walk and a `MethodInfo.Equals` per candidate, plus an `object[]` allocation per call. Nothing is cached, and **a struct with no overrides pays the full cost for no benefit**.

Measured under `gsi`, 20 000 inserts + 20 000 `ContainsKey` lookups of `struct Plain { var n int32 }` — **no overrides at all** — best of 3 runs each, output `20000` on both:

| | total | minus startup | work |
|---|---:|---:|---:|
| base `f7d9a5d3` | 553 ms | 218 ms | **335 ms** |
| merged | 835 ms | 224 ms | **611 ms** |

**≈1.8× slower** on the work. Startup is unchanged (218 vs 224 ms), which validates the setup, and both trees differ only by this PR, so attribution is clean.

This is correctness-neutral and the interpreter is not the perf-critical driver, so I grade it SHOULD-FIX rather than blocking. The fix is small and in keeping with the PR's own centralisation: memoize `FindExternalOverride` per `(StructSymbol, MethodInfo)` in a `ConcurrentDictionary` on the evaluator, and take the fast path when the cached result is `null`. That collapses the common no-override case to one dictionary probe.

## Pre-existing sibling defect — recommend filing separately, not fixing here

`fp4_class` surfaced a defect this PR neither causes nor worsens, **identical on base and merged**: `StructValue.Equals`/`GetHashCode` have **no `IsClass` branch** (`StructValue.cs:157–204`; `grep IsClass` in that file returns only layout/type-mapping hits), so a **class without overrides gets structural value equality instead of reference identity** in the evaluator.

Two distinct `PlainC` instances with equal fields, used as Dictionary keys — `[evaluate / emit / gsi]`:

| assertion | base | merged | oracle |
|---|---|---|---|
| `d.Count` | 1 / **2** / 1 | 1 / **2** / 1 | **2** |
| `a.Equals(b)` | True / **False** / True | True / **False** / True | **False** |
| `a.ToString()` | `PlainC(n=11)` / **`P.PlainC`** / `PlainC(n=11)` | unchanged | `P.PlainC` |

This is the *same defect class as #3116* — a silent container-element-count divergence at rc0 with no diagnostic — living in the very method this PR modifies. It is out of scope here (fixing it means giving `StructValue` reference semantics for `IsClass`, which is a behaviour change of its own), but it should not be lost. Suggest it be filed as a follow-up.

## NITs

**NIT 1 — the dispatcher-less seam stays public.** The whole fix rests on the invariant "every product `StructValue` carries a dispatcher", but `public StructValue(StructSymbol)` remains public and silently produces a dispatcher-less value. The compiler census proves 0 product callers *today*; nothing fails if a new one appears tomorrow. Making it `internal` (the 2 remaining callers are in-repo tests) would convert the invariant from a one-time audit into a compile-time guarantee.

**NIT 2 — `ToString` divergence for types that never override.** Evaluate/`gsi` print `Plain(n=11)`/`PlainC(n=11)` where emit prints `P.Plain`/`P.PlainC`. Identical on both trees, so pre-existing and not this PR's; noting it only so it is not later attributed here.

## Suites — merged tree

| Project | Result |
|---|---|
| `Interpreter.Tests` | **753 / 753** |
| `Core.Tests` | **7089 passed, 1 skipped, 7090 total** |
| `Compiler.Tests` | **4157 / 4157** |
| `Sdk.Tests` | **NOT RUN** — deliberately: it touches the shared `~/.nuget/packages/gsharp.net.sdk/` cache that parallel agents race on |

**The PR body's numbers check out.** Head tree hash `9226992ad031…` matches its claim exactly; the census of 10, the 20/30 split and the 4/6 depth split all reproduce. Its `Interpreter.Tests: 744` is now 753, and `Compiler.Tests: 4152` is now 4157, only because `main` moved under it — not because the figures were wrong.

6 of the 7 projects were run and are green; the single omission is stated rather than implied.

DavidObando:
## Addendum — my suite denominator was wrong ("6 of the 7"); the two missing projects are now run on current `main`, both green

Correcting my own verdict above, per the fleet bulletin on test-project enumeration.

I wrote **"6 of the 7 projects were run and are green."** The denominator is wrong. **`GSharp.sln` contains nine test projects**, two of which live outside `test/` and which I never ran:

- `tools/cs2gs/Cs2Gs.Tests` — drives the real `gsc` **out-of-process**

So the correct statement for this review is **6 of 9 run**, not 6 of 7. My enumeration came from `ls test/`, which is scoped to one directory — the zero was **scope, not absence**, which is the exact error I flag in other people's work.

**This omission was materially relevant to *this* PR**, more than to most. #3127 routes `GetHashCode`/`Equals`/`ToString` through an override walk **at the BCL boundary**, and `Cs2Gs.Tests` invokes real `gsc` out-of-process — precisely the shape that can redden a gate no test under `test/` observes. I should have run it.

### Closing it empirically

I cannot re-run this PR's merged tree (that scratch tree has been repurposed and `main` has moved well past it), so I measured the thing that actually matters — whether any breakage of this shape is live on `main` today. Current `main` **`5b71a74d`** (verified with `git rev-parse` against `origin/main`, not assumed), full solution rebuild after checkout, **`BUILD_RC=0`**:

| project | result on `main` `5b71a74d` |
|---|---|
| `Cs2Gs.Tests` | **1903 passed, 0 failed** |

Freshness verified in **epoch seconds with `sort -n`** (not clock strings): `Core/`, `Compiler/`, `Repl/` and `Cs2Gs.Tests/` copies of `GSharp.Core.dll` all at the identical build stamp, `ref/` excluded.

**No cs2gs breakage from this PR is live on `main`.** External corpus: `tools/cs2gs/external/oahu.json` pins commit **`0ac1fece4415d31955ae7a1dcf7da31e343d363d`**, unchanged — stated from the pin, which is the only thing that can prove it.

**Verdict unchanged: MERGEABLE.** No finding moves, and the SHOULD-FIX on uncached `FindExternalOverride` still stands as filed. Recording the wrong denominator because a posted number that overstates coverage is worth more as a correction than as a footnote.

PR #3128, opened 2026-08-03
Interpret Span and ReadOnlySpan values

### LINKED ISSUE #3114: gsi and bare gsc fail two shipped samples with GS9999 on ReadOnlySpan, and no gate sees it
## Summary

Now that #3011 has landed and `gsi` (and bare `gsc`) invoke an explicit `func Main`, two shipped samples that previously did nothing under `gsi` now **fail**:

| sample | oracle (`gsc /out:` + run) | `gsi` before #3011 | `gsi` after #3011 | bare `gsc` after #3011 |
|---|---|---|---|---|
| `samples/SpanComprehensive.gs` | rc0 `60/10/2` | rc0, empty | **rc1** | **rc1** |
| `samples/RefStructGenericField.gs` | rc0 `3` | rc0, empty | **rc1** | **rc1** |

```
GS9999: Specified method is not supported. (ReadOnlySpan<int>)
```

These are **exactly 2 of 122** samples — the only two containing `func Main(`.

## This is not a regression introduced by #3011

The underlying limitation is **pre-existing**: the interpreter cannot invoke through a `ByRefLike` type. A minimal `ReadOnlySpan` conversion probe is rc1 `GS9999` on the pre-#3011 tree too. What #3011 changed is **reachability** — the shape was previously unreachable because `Main` was never invoked, and `GS0219` blocks the equivalent top-level form.

Both reviewers agreed a **loud failure is better than the silent no-op** it replaced, which is why #3011 merged. But it needs an owner, because #2984's repro section now reads as satisfied when these two samples do not in fact work.

## Why no gate catches it

`test/Compiler.Tests/LanguageConformance/SampleConformanceTests.cs:16` compiles via `gsc.dll` and runs the result with `dotnet` — it is **emit-only**. Nothing runs `samples/` through `gsi` or through bare `gsc`. This is the same structural blindness that let #3100 ship: the evaluate column, shared by bare `gsc` and `gsi`, is invisible to CI.

## Scope — two separable pieces

1. **Support, or diagnose, `ByRefLike` invocation in the interpreter.** `Span`/`ReadOnlySpan` cannot be stored in the interpreter's value model, so full support may not be the right answer — but `GS9999` is the internal-compiler-error bucket, and this is a **reachable user-facing limitation**, so it should carry a real diagnostic that names the limitation and points at `gsc /out:`. Related prior work: #2987 / PR #3055 established the `GS0511` boundary diagnostic for exactly this family; this may be the same mechanism reached by a different route.
2. **Close the coverage hole.** Add a `gsi` column (and a bare-`gsc` column) to the sample conformance gate so that a sample whose oracle works but whose interpreting drivers do not is a **CI failure**, ledgered explicitly if it must be tolerated. Note PR #3077 is already reworking that file and is currently red on `build` — coordinate rather than collide.

## Acceptance criteria

- Neither sample produces `GS9999` under `gsi` or bare `gsc`; either they work, or they produce a located, named diagnostic that explains the interpreter boundary.
- A gate exists that would have caught this: a sample whose emit column succeeds while an interpreting column fails must fail CI unless explicitly ledgered with an issue number and a true reason.
- Ledger reasons must be accurate — `PInvokeStructMarshalling.gs` was ledgered under #2986 as "cannot execute the native struct-marshalling call" when the sample has **no native call at all**; the real cause was `@FieldOffset` overlap (#3100, fixed in #3105).

DavidObando:
## Root cause

Before PR #3055 converted this failure to `GS0511`, the exception behind `GS9999` was:

```
System.NotSupportedException: Specified method is not supported.
  RuntimeMethodInfo.ThrowNoInvokeException
  RuntimeMethodInfo.Invoke
  Evaluator.InvokeReflective
  Evaluator.EvaluateClrConversionCallExpression
  Evaluator.EvaluateExpression
```

Reflection cannot invoke an array-to-`ReadOnlySpan<T>` conversion because its ByRefLike result cannot be boxed into the evaluator's `object` slots.

## Fix

Represent interpreted `Span<T>`/`ReadOnlySpan<T>` values with an evaluator-owned array, offset, length, and read-only state. The reflection boundary handles array and Span conversions, `Length`, indexing, writes, slicing, and default Span fields while preserving `GS0511` for unsupported CLR-returned ByRefLike values.

`InterpretedSpanValue.ToString()` now matches the emitted BCL shape, so direct formatting and evaluator interpolation no longer expose the internal wrapper type. The GS0511 documentation now states the surviving boundary: array-backed spans work; CLR methods and properties returning stack-only values remain unsupported.

PR #3077 merged while this PR was under review. Its three stale Span expected-difference entries are removed here; the all-driver sample gate is now 125/125.

## Driver evidence

All outputs below are program output; bare `gsc` additionally writes its normal `Success.` protocol line.

| sample | bare `gsc` evaluates | `gsc /out:` emits | `gsi` interprets |
|---|---|---|---|
| `SpanComprehensive.gs` | rc0, `60 / 10 / 2` | rc0, `60 / 10 / 2` | rc0, `60 / 10 / 2` |
| `RefStructGenericField.gs` | rc0, `3` | rc0, `3` | rc0, `3` |
| `SpanIndexing.gs` | rc0, `60 / 402` | rc0, `60 / 402` | rc0, `60 / 402` |

Every emitted probe used its own asserted-empty output directory. Test probes load emitted assemblies with `Assembly.Load` and `GetTypes()` before execution.

### Formatting

| shape | bare `gsc` | emit oracle | `gsi` |
|---|---|---|---|
| `Span<int32>.ToString()` | rc0 `System.Span<Int32>[3]` | rc0 `System.Span<Int32>[3]` | rc0 `System.Span<Int32>[3]` |
| `ReadOnlySpan<int32>.ToString()` | rc0 `System.ReadOnlySpan<Int32>[3]` | rc0 `System.ReadOnlySpan<Int32>[3]` | rc0 `System.ReadOnlySpan<Int32>[3]` |
| direct interpolation | rc0 with the same public names | rc1 `GS9998` (pre-existing emit limitation) | rc0 with the same public names |

## All 122 shipped single-file samples

Measured before and after on the same live base, `2dea578b1d3a43e964ebe5ce14ddc763bbe1628f`, with 122/122 samples executed in each interpreting column:

| driver | base golden matches | head golden matches | rc0 executions |
|---|---:|---:|---:|
| bare `gsc` | 95 / 122 | 98 / 122 | 109 → 112 |
| `gsi` | 108 / 122 | 111 / 122 | 109 → 112 |

Exactly six cells changed: `RefStructGenericField.gs`, `SpanComprehensive.gs`, and `SpanIndexing.gs` under both interpreting drivers. Each moved from rc1 `GS0511` to rc0 with golden output; no other vector changed.

## Product mutations

Each mutation was asserted to apply and built successfully.

| mutation | result |
|---|---|
| default Span detection | killed by `ShippedSpanSample_MatchesGoldenAcrossDrivers` |
| conversion dispatch route | killed by `ShippedSpanSample_MatchesGoldenAcrossDrivers` |
| `Length` recognition | killed by `SpanConversionMutationAndSlicing_AgreeAcrossDrivers` |
| indexer read recognition | killed by `SpanConversionMutationAndSlicing_AgreeAcrossDrivers` |
| indexer write recognition | killed by `SpanConversionMutationAndSlicing_AgreeAcrossDrivers` |
| implicit-conversion recognition | killed by `ShippedSpanSample_MatchesGoldenAcrossDrivers` |
| slice bounds | killed by `SpanConversionMutationAndSlicing_AgreeAcrossDrivers` |
| index bounds | killed by `SpanConversionMutationAndSlicing_AgreeAcrossDrivers` |
| `ToString()` read-only selector inversion | killed: 4 failures in the 14-case issue matrix |
| read-only write guard removal | **survived 734/734**; binder rejects the source first with `GS0226` |
| writable-conversion guard removal | **survived 734/734**; binder rejects the source first with `GS0155` |

The two surviving guards are defense in depth and are unreachable through binder-approved source. The earlier 10/10 claim was incorrect.

## Validation

Current base: `61fe9cb943d15c0f5ba747dabd562e66fb132058`  
Head: `d9eee3bc86b2619679bbc418dfc2cab2872d4b5d`  
`git merge-tree` output: `52f74eb9fb21aca930ade8c5e8e3f95cf253a0c3`, exactly matching the head tree.

Serial Release rebuild: `BUILD_RC=0`, 0 warnings, 0 errors. Core/Compiler/Repl and all three test-consumer `GSharp.Core.dll` copies were fresh by epoch mtime, non-zero, and hash-identical (`4a0be6009a61713835f7726e3d50cffb`).

Three-driver `Span[char]` and `ReadOnlySpan[char]` formatting:

- bare `gsc`: rc0, `abc` / `abc` (then `Success.`)
- emitted oracle: compile rc0, run rc0, `abc` / `abc`
- `gsi`: rc0, `abc` / `abc`

Issue matrix: 14/14 passed. Sliced `ReadOnlySpan[int32]` now asserts `System.ReadOnlySpan<Int32>[2]`. Product mutant `{Length}` → `{this.array.Length}` was asserted applied, built with rc0, and killed 2/14 by `SpanConversionMutationAndSlicing_AgreeAcrossDrivers` (bare `gsc` evaluation and `gsi`); actual `[3]` versus expected/oracle `[2]`.

All nine unfiltered test projects enumerated from `GSharp.sln` passed. Eight projects were run at anchor `faca348a36e7c4f9f35a3f57bde43c47f958b603`; SDK was run after rebasing to `61fe9cb943d15c0f5ba747dabd562e66fb132058` (the intervening main change only added a Compiler test file):

- Core: 7,116 passed, 1 skipped
- Compiler: 4,232 passed
- Cs2Gs: 1,903 passed
- Interpreter: 1,227 passed
- Language Server: 462 passed
- Extensions: 104 passed
- GeneratorHost: 31 passed
- Internal analyzers: 7 passed
- SDK: 59 passed (`61fe9cb943d15c0f5ba747dabd562e66fb132058`)

Correction: `Sdk.Tests` invokes `Program.Main` in-process and does not write the shared SDK package cache. The earlier NOT RUN statement was based on a docstring, not executable code.

Fixes #3114

### DIFF
--- a/docs/diagnostics.md
+++ b/docs/diagnostics.md
@@ -1177,11 +1177,12 @@ compiler error. It applies even when the declaration is not called because
 
 | ID | Severity | Description |
 |----|----------|-------------|
-| GS0511 | Error | The interpreter cannot represent a stack-only (`ByRefLike`) CLR value such as `Span[T]` or `ReadOnlySpan[T]`. |
+| GS0511 | Error | The interpreter cannot represent this stack-only (`ByRefLike`) CLR value. Array-backed `Span[T]` and `ReadOnlySpan[T]` values are supported. |
 
 Interpreter values use boxed storage, which cannot hold stack-only CLR values
-or preserve their interior references. Compile the program with `gsc /out:<file>`
-to use these types.
+or preserve their interior references. The interpreter emulates spans created
+from arrays; CLR methods and properties returning stack-only values remain
+unsupported. Compile the program with `gsc /out:<file>` to use those values.
 
 ## Interpreter deinitializer boundary (GS0510)
 
--- a/samples/SpanIndexing.gs
+++ b/samples/SpanIndexing.gs
@@ -6,8 +6,6 @@
 // and a `Span[T]` element write `s[i] = v` stores through the returned `ref T`.
 // Spans are stack-only (GS0219), so the span locals live inside functions and
 // are exercised from top-level statements.
-// The tree-walk interpreter cannot represent these values and reports GS0511;
-// compile this sample with gsc /out:<file>.
 
 package GSharp.Samples.SpanIndexing
 
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -2280,6 +2280,11 @@ public sealed partial class Evaluator
             return sv;
         }
 
+        if (type?.ClrType != null && IsSpanType(type.ClrType))
+        {
+            return InterpretedSpanValue.Empty(type.ClrType);
+        }
+
         // Issue #1652: every other value type (bool, all sized/unsigned ints,
         // float32/float64, decimal, char, nint/nuint, and user value structs
         // that slipped through as plain ClrType) gets its real CLR default via
--- a/src/Core/CodeAnalysis/Evaluator.Reflection.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Reflection.cs
@@ -262,6 +262,9 @@ public sealed partial class Evaluator
             return new InterpretedSpanValue(this.array, this.start + start, length, IsReadOnly);
         }
 
+        public override string ToString()
+            => $"System.{(IsReadOnly ? "ReadOnlySpan" : "Span")}<{this.array.GetType().GetElementType().Name}>[{Length}]";
+
         private int CheckedIndex(int index)
         {
             if ((uint)index >= (uint)Length)
--- a/src/Core/CodeAnalysis/Evaluator.Reflection.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Reflection.cs
@@ -18,6 +18,12 @@ public sealed partial class Evaluator
 {
     private object InvokeReflective(MethodBase target, object receiver, object[] args, BoundNode node)
     {
+        object result;
+        if (TryInvokeSpanConversion(target, args, out result))
+        {
+            return result;
+        }
+
         ThrowIfByRefLikeSignature(target, node);
         return target is ConstructorInfo constructor
             ? constructor.Invoke(args)
@@ -26,6 +32,20 @@ public sealed partial class Evaluator
 
     private object GetReflective(MemberInfo member, object receiver, object[] index, BoundNode node)
     {
+        var span = receiver as InterpretedSpanValue;
+        if (span != null && member is PropertyInfo spanProperty)
+        {
+            if (spanProperty.Name == "Length")
+            {
+                return span.Length;
+            }
+
+            if (spanProperty.Name == "Item" && index.Length == 1)
+            {
+                return span[(int)index[0]];
+            }
+        }
+
         ThrowIfByRefLikeSignature(member, node);
         return member switch
         {
@@ -37,6 +57,13 @@ public sealed partial class Evaluator
 
     private void SetReflective(MemberInfo member, object receiver, object value, object[] index, BoundNode node)
     {
+        var span = receiver as InterpretedSpanValue;
+        if (span != null && member is PropertyInfo spanProperty && spanProperty.Name == "Item" && index.Length == 1)
+        {
+            span[(int)index[0]] = value;
+            return;
+        }
+
         ThrowIfByRefLikeSignature(member, node);
         switch (member)
         {
@@ -50,6 +77,46 @@ public sealed partial class Evaluator
                 throw new EvaluatorException($"Unsupported CLR member kind '{member.MemberType}'.", node);
         }
     }
+
+    private static bool TryInvokeSpanConversion(MethodBase target, object[] args, out object result)
+    {
+        if (target is MethodInfo conversion
+            && conversion.Name == "op_Implicit"
+            && IsSpanType(conversion.ReturnType)
+            && args.Length == 1)
+        {
+            var readOnly = IsReadOnlySpanType(conversion.ReturnType);
+            if (args[0] is Array array)
+            {
+                result = new InterpretedSpanValue(array, readOnly);
+                return true;
+            }
+
+            if (args[0] is InterpretedSpanValue source && (readOnly || !source.IsReadOnly))
+            {
+                result = source.As(readOnly);
+                return true;
+            }
+        }
+
+        result = null;
+        return false;
+    }
+
+    private static bool IsSpanType(Type type)
+    {
+        if (!type.IsGenericType)
+        {
+            return false;
+        }
+
+        var definition = type.GetGenericTypeDefinition();
+        return definition.IsSameAs(typeof(Span<>))
+            || definition.IsSameAs(typeof(ReadOnlySpan<>));
+    }
+
+    private static bool IsReadOnlySpanType(Type type)
+        => type.IsGenericType && type.GetGenericTypeDefinition().IsSameAs(typeof(ReadOnlySpan<>));
 
     private static void ThrowIfByRefLikeSignature(MemberInfo member, BoundNode node)
     {
@@ -136,5 +203,73 @@ public sealed partial class Evaluator
         }
 
         return null;
+    }
+
+    private sealed class InterpretedSpanValue
+    {
+        private readonly Array array;
+        private readonly int start;
+
+        public InterpretedSpanValue(Array array, bool readOnly)
+            : this(array, 0, array.Length, readOnly)
+        {
+        }
+
+        private InterpretedSpanValue(Array array, int start, int length, bool readOnly)
+        {
+            this.array = array;
+            this.start = start;
+            Length = length;
+            IsReadOnly = readOnly;
+        }
+
+        public int Length { get; }
+
+        public bool IsReadOnly { get; }
+
+        public object this[int index]
+        {
+            get => this.array.GetValue(CheckedIndex(index));
+            set
+            {
+                if (IsReadOnly)
+                {
+                    throw new InvalidOperationException("Cannot write through a ReadOnlySpan.");
+                }
+
+                this.array.SetValue(value, CheckedIndex(index));
+            }
+        }
+
+        public static InterpretedSpanValue Empty(Type type)
+            => new(
+                Array.CreateInstance(type.GetGenericArguments()[0], 0),
+                IsReadOnlySpanType(type));
+
+        public InterpretedSpanValue As(bool readOnly)
+            => new(this.array, this.start, Length, readOnly);
+
+        public InterpretedSpanValue Slice(int start)
+            => Slice(start, Length - start);
+
+        public InterpretedSpanValue Slice(int start, int length)
+        {
+            if ((uint)start > (uint)Length || (uint)length > (uint)(Length - start))
+            {
+                throw new ArgumentOutOfRangeException();
+            }
+
+            return new InterpretedSpanValue(this.array, this.start + start, length, IsReadOnly);
+        }
+
+        private int CheckedIndex(int index)
+        {
+            if ((uint)index >= (uint)Length)
+            {
+                throw new IndexOutOfRangeException();
+            }
+
+            return this.start + index;
+        }
     }
 }
--- a/src/Core/CodeAnalysis/Evaluator.Reflection.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Reflection.cs
@@ -263,7 +263,9 @@ public sealed partial class Evaluator
         }
 
         public override string ToString()
-            => $"System.{(IsReadOnly ? "ReadOnlySpan" : "Span")}<{this.array.GetType().GetElementType().Name}>[{Length}]";
+            => this.array is char[] chars
+                ? new string(chars, this.start, Length)
+                : $"System.{(IsReadOnly ? "ReadOnlySpan" : "Span")}<{this.array.GetType().GetElementType().Name}>[{Length}]";
 
         private int CheckedIndex(int index)
         {
--- a/test/Compiler.Tests/LanguageConformance/SampleConformanceTests.cs
+++ b/test/Compiler.Tests/LanguageConformance/SampleConformanceTests.cs
@@ -63,19 +63,7 @@ public class SampleConformanceTests
             ["PInvokeLibraryImportStringReturn.gs"] = NativeCallDifference(),
             ["PInvokeMarshalAs.gs"] = NativeCallDifference(),
             ["PInvokeRefOutIn.gs"] = NativeCallDifference(),
-            ["RefStructGenericField.gs"] = SameDifference(
-                Expected(1, string.Empty, "GS0511"),
-                "docs/diagnostics.md GS0511 / #3114",
-                "the evaluator invokes func Main but rejects ByRefLike signatures"),
             ["RefStructSpan.gs"] = SameDifference(
-                Expected(1, string.Empty, "GS0511"),
-                "docs/diagnostics.md GS0511 / #3114",
-                "reflection cannot invoke ByRefLike signatures"),
-            ["SpanComprehensive.gs"] = SameDifference(
-                Expected(1, string.Empty, "GS0511"),
-                "docs/diagnostics.md GS0511 / #3114",
-                "the evaluator invokes func Main but rejects ByRefLike signatures"),
-            ["SpanIndexing.gs"] = SameDifference(
                 Expected(1, string.Empty, "GS0511"),
                 "docs/diagnostics.md GS0511 / #3114",
                 "reflection cannot invoke ByRefLike signatures"),
@@ -199,9 +187,6 @@ public class SampleConformanceTests
             var golden = SampleConformanceData.ReadNormalizedFile(
                 Path.Combine(samplesDirectory, Path.ChangeExtension(sample, ".golden")));
             Assert.False(string.IsNullOrEmpty(golden), $"{sample} must have non-empty golden output.");
-            Assert.True(
-                ExpectedDifferences.ContainsKey(sample),
-                $"{sample} declares an explicit Main with non-empty golden output; declare any driver difference explicitly.");
         }
     }
 
--- a/test/Interpreter.Tests/Issue2987ByRefLikeBoundaryInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue2987ByRefLikeBoundaryInterpreterTests.cs
@@ -63,19 +63,6 @@ public class Issue2987ByRefLikeBoundaryInterpreterTests
             """,
             "System.Span[int32]"
         },
-        {
-            """
-            import System
-
-            func value(values []int32) int32 {
-                var span ReadOnlySpan[int32] = values
-                return 22
-            }
-
-            value([]int32{11, 22, 33})
-            """,
-            "System.ReadOnlySpan[int32]"
-        },
     };
 
     [Theory]
--- a/test/Interpreter.Tests/Issue3114ReadOnlySpanInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue3114ReadOnlySpanInterpreterTests.cs
@@ -46,6 +46,12 @@ public class Issue3114ReadOnlySpanInterpreterTests
         }
     }
 
+    public static IEnumerable<object[]> InterpretingDrivers()
+    {
+        yield return new object[] { Driver.CompilerEvaluation };
+        yield return new object[] { Driver.Interpreter };
+    }
+
     [Theory]
     [MemberData(nameof(SampleDriverCases))]
     public void ShippedSpanSample_MatchesGoldenAcrossDrivers(string sample, string expected, Driver driver)
@@ -96,6 +102,8 @@ public class Issue3114ReadOnlySpanInterpreterTests
                 Console.WriteLine(readOnly.Length)
                 Console.WriteLine(readOnly[1])
                 Console.WriteLine(window[1])
+                Console.WriteLine(writable.ToString())
+                Console.WriteLine(readOnly.ToString())
             }
             """;
 
@@ -119,7 +127,56 @@ public class Issue3114ReadOnlySpanInterpreterTests
 
             Assert.Equal(0, result.ExitCode);
             Assert.Equal(
-                driver == Driver.CompilerEvaluation ? "3\n44\n33\nSuccess.\n" : "3\n44\n33\n",
+                driver == Driver.CompilerEvaluation
+                    ? "3\n44\n33\nSystem.Span<Int32>[3]\nSystem.ReadOnlySpan<Int32>[3]\nSuccess.\n"
+                    : "3\n44\n33\nSystem.Span<Int32>[3]\nSystem.ReadOnlySpan<Int32>[3]\n",
+                result.StandardOutput);
+            Assert.Equal(string.Empty, result.StandardError);
+        }
+        finally
+        {
+            Directory.Delete(root, recursive: true);
+        }
+    }
+
+    [Theory]
+    [MemberData(nameof(InterpretingDrivers))]
+    public void SpanInterpolation_UsesPublicSpanName(Driver driver)
+    {
+        const string Source = """
+            import System
+
+            func Main() {
+                var values = []int32{11, 22, 33}
+                var writable Span[int32] = values
+                var readOnly ReadOnlySpan[int32] = writable
+                Console.WriteLine("writable=${writable}")
+                Console.WriteLine("readonly=${readOnly}")
+            }
+            """;
+
+        var root = Path.Combine(Environment.CurrentDirectory, $".issue3114-{Guid.NewGuid():N}");
+        Directory.CreateDirectory(root);
+        Assert.Empty(Directory.EnumerateFileSystemEntries(root));
+        var sourcePath = Path.Combine(root, "span-interpolation.gs");
+        File.WriteAllText(sourcePath, Source);
+
+        try
+        {
+            var result = driver switch
+            {
+                Driver.CompilerEvaluation => CaptureConsole(
+                    () => GSharp.Compiler.Program.Main([sourcePath])),
+                Driver.Interpreter => CaptureConsole(
+                    () => GSharp.Repl.Program.Main([sourcePath])),
+                _ => throw new InvalidOperationException($"Unexpected driver {driver}."),
+            };
+
+            Assert.Equal(0, result.ExitCode);
+            Assert.Equal(
+                driver == Driver.CompilerEvaluation
+                    ? "writable=System.Span<Int32>[3]\nreadonly=System.ReadOnlySpan<Int32>[3]\nSuccess.\n"
+                    : "writable=System.Span<Int32>[3]\nreadonly=System.ReadOnlySpan<Int32>[3]\n",
                 result.StandardOutput);
             Assert.Equal(string.Empty, result.StandardError);
         }
--- a/test/Interpreter.Tests/Issue3114ReadOnlySpanInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue3114ReadOnlySpanInterpreterTests.cs
@@ -99,11 +99,17 @@ public class Issue3114ReadOnlySpanInterpreterTests
                 writable[1] = 44
                 var tail = readOnly.Slice(1)
                 var window = tail.Slice(0, 2)
+                var letters = []char{'a', 'b', 'c'}
+                var writableChars Span[char] = letters
+                var readOnlyChars ReadOnlySpan[char] = writableChars
                 Console.WriteLine(readOnly.Length)
                 Console.WriteLine(readOnly[1])
                 Console.WriteLine(window[1])
                 Console.WriteLine(writable.ToString())
                 Console.WriteLine(readOnly.ToString())
+                Console.WriteLine(window.ToString())
+                Console.WriteLine(writableChars.ToString())
+                Console.WriteLine(readOnlyChars.ToString())
             }
             """;
 
@@ -128,8 +134,8 @@ public class Issue3114ReadOnlySpanInterpreterTests
             Assert.Equal(0, result.ExitCode);
             Assert.Equal(
                 driver == Driver.CompilerEvaluation
-                    ? "3\n44\n33\nSystem.Span<Int32>[3]\nSystem.ReadOnlySpan<Int32>[3]\nSuccess.\n"
-                    : "3\n44\n33\nSystem.Span<Int32>[3]\nSystem.ReadOnlySpan<Int32>[3]\n",
+                    ? "3\n44\n33\nSystem.Span<Int32>[3]\nSystem.ReadOnlySpan<Int32>[3]\nSystem.ReadOnlySpan<Int32>[2]\nabc\nabc\nSuccess.\n"
+                    : "3\n44\n33\nSystem.Span<Int32>[3]\nSystem.ReadOnlySpan<Int32>[3]\nSystem.ReadOnlySpan<Int32>[2]\nabc\nabc\n",
                 result.StandardOutput);
             Assert.Equal(string.Empty, result.StandardError);
         }
--- a/test/Interpreter.Tests/Issue3114ReadOnlySpanInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue3114ReadOnlySpanInterpreterTests.cs
@@ -1,0 +1,209 @@
+// <copyright file="Issue3114ReadOnlySpanInterpreterTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+using Xunit;
+
+namespace GSharp.Interpreter.Tests;
+
+/// <summary>Issue #3114: shipped Span samples agree across all three drivers.</summary>
+[Collection("ConsoleIo")]
+public class Issue3114ReadOnlySpanInterpreterTests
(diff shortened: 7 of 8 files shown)


DavidObando:
## Review: PR #3128 — `fix/3114-readonlyspan-interpreted`

**Verdict: NOT MERGEABLE** — 1 BLOCKER, 2 SHOULD-FIX, 1 NIT.

The core of this change is good: the per-sample result vector proves the "only three samples changed" claim exactly, with zero regressions, and every newly-passing sample is byte-identical to the emit oracle. The blocker is a new *silent* divergence in a shape the tests do not cover.

### Measurement anchors

| item | value |
|---|---|
| `main` | `f7d9a5d3` |
| head | `597a3102` |
| merge base | `4469e323` (2 behind / 1 ahead) |
| `git merge-tree` output | 1 line, tree `dec27cba`, **0 conflict markers** (output read, not rc) |
| merged tree reproduced | `dec27cba644921e75218fbcd26e8e22d7f4b67be` — exact |
| `BUILD_RC` base / merged | **0 / 0**, 0 warnings, 0 errors |
| `Interpreter.Tests` (merged) | **734 / 734** |

Binaries verified fresh by **mtime** in `Core/`, `Compiler/` and `Repl/` (excluding `ref/`), distinct between trees. Every probe ran in its own asserted-empty directory. Harness positive-controlled: a deliberately malformed source returns rc=1 with diagnostics on all three drivers, so no error is being silenced.

### Scope split — HONOURED

`test/Compiler.Tests/LanguageConformance/SampleConformanceTests.cs` is **not** in the diff. Positive-controlled: the file *does* exist on the head (`git ls-tree` → 1), so this is a real absence and not a scope artifact. No collision with PR #3077.

### Diagnostic identity on `main` — the issue title is stale, confirmed

Measured directly, not taken from the issue:

| sample | eval(base) | emit(base) | gsi(base) |
|---|---|---|---|
| `RefStructGenericField.gs` | rc=1 **GS0511** | rc=0 | rc=1 **GS0511** |
| `RefStructSpan.gs` | rc=1 **GS0511** | rc=0 | rc=1 **GS0511** |
| `SpanComprehensive.gs` | rc=1 **GS0511** | rc=0 | rc=1 **GS0511** |
| `SpanIndexing.gs` | rc=1 **GS0511** | rc=0 | rc=1 **GS0511** |

All four are `GS0511`, not `GS9999`. **Was a designed `GS0511` boundary turned into working code? Yes — but legitimately.** Issue #2987 names option 1 ("special-case the small set of `Span`/`ReadOnlySpan` operations … including the implicit `[]T` → span conversion") as *the real fix*, and #3055 shipped option 2 as the honest interim. This PR implements option 1 for the array-backed subset. That is the sanctioned trajectory, so I am **not** raising it as a blocker. The consequence for the docs is raised as SHOULD-FIX 2.

### Per-sample result vector — "only three changed" VERIFIED

122 samples × 3 drivers × 2 trees = 732 runs. Diffing the full vectors (not the totals):

```
< RefStructGenericField.gs|eval|rc=1|GS0511   > RefStructGenericField.gs|eval|rc=0
< RefStructGenericField.gs|gsi |rc=1|GS0511   > RefStructGenericField.gs|gsi |rc=0
< SpanComprehensive.gs    |eval|rc=1|GS0511   > SpanComprehensive.gs    |eval|rc=0
< SpanComprehensive.gs    |gsi |rc=1|GS0511   > SpanComprehensive.gs    |gsi |rc=0
< SpanIndexing.gs         |eval|rc=1|GS0511   > SpanIndexing.gs         |eval|rc=0
< SpanIndexing.gs         |gsi |rc=1|GS0511   > SpanIndexing.gs         |gsi |rc=0
```

**12 changed cells, 3 distinct samples, all in the improving direction. Zero regressions.** The emit column is completely untouched (119/119 both trees). `RefStructSpan.gs` remains `GS0511` because it uses `text.AsSpan()` (string-backed), which the fix does not intercept.

**Execution, not just passing:** eval 109→112, gsi 109→112, emit 119/119, and **0 vacuous rc0 passes** — every passing sample actually produced stdout, so the delta is real execution rather than scope.

**Emit oracle:** all newly-passing samples byte-identical to emit — `RefStructGenericField` `3`, `SpanComprehensive` `60/10/2`, `SpanIndexing` `60/402`.

Root cause confirmed at the right site: the fix sits on the `InvokeReflective` / `GetReflective` / `SetReflective` chokepoints, which is where the ByRefLike refusal lives. (`ThrowNoInvokeException` is a .NET BCL internal on `RuntimeMethodInfo`, consistent with a genuine pre-#3055 `NotSupportedException` stack — not a fabricated chain.)

---

## BLOCKER 1 — a span value now escapes as its internal class name, rc=0, where `main` refused loudly

`src/Core/CodeAnalysis/Evaluator.Reflection.cs:207` (`InterpretedSpanValue`) has no `ToString()` override, and the value is duck-typed into the evaluator by runtime-type rebinding at `Evaluator.Calls.cs:1107-1112` / `Evaluator.Expressions.cs:1542-1547`. Any object-ish use therefore reaches `System.Object.ToString()`.

```
func Main() {
    var values = []int32{11, 22, 33}
    var s Span[int32] = values
    Console.WriteLine(s.ToString())
}
```

| driver | `main` (`f7d9a5d3`) | merged head |
|---|---|---|
| emit **(oracle)** | rc=0 `System.Span<Int32>[3]` | rc=0 `System.Span<Int32>[3]` |
| bare `gsc` (eval) | rc=1 **GS0511** | **rc=0 `GSharp.Core.CodeAnalysis.Evaluator+InterpretedSpanValue`** |
| `gsi` | rc=1 **GS0511** | **rc=0 `GSharp.Core.CodeAnalysis.Evaluator+InterpretedSpanValue`** |

Same leak through interpolation, where it is worse because the rc diverges too:

```
Console.WriteLine("span=${s}")
```

| driver | merged head |
|---|---|
| emit | rc=1 `GS9998` (pre-existing emit limitation) |
| eval / `gsi` | **rc=0** `span=GSharp.Core.CodeAnalysis.Evaluator+InterpretedSpanValue` |

**Impact.** On `main` both programs were a loud, documented refusal. On the head they succeed and print an internal compiler implementation type. This is a silent wrong answer replacing an honest failure, it diverges from the emit oracle at rc=0, and it is invisible to the entire suite — **734/734 green**. It also leaks an internal type name into user-visible output.

**Recommended fix (small).** Override `ToString()` on `InterpretedSpanValue` to match the BCL format, e.g. `System.Span<Int32>[3]` / `System.ReadOnlySpan<Int32>[3]`, derived from the element type and `Length`. Add a driver-matrix test asserting `s.ToString()` is identical across emit / eval / `gsi`, since that assertion is exactly what the current suite lacks.

---

## SHOULD-FIX 1 — the two guards flagged hardest are unreachable; both mutants SURVIVE

Both were run as product mutants on the merged tree, asserted applied, `BUILD_RC=0`.

| mutant | site | assertion | result |
|---|---|---|---|
| **A** — delete the readonly write guard | `Evaluator.Reflection.cs`, `InterpretedSpanValue` indexer `set` | `Cannot write through a ReadOnlySpan` sites 1 → **0** | **SURVIVED 734/734** |
| **B** — drop `(readOnly \|\| !source.IsReadOnly)` from `TryInvokeSpanConversion` | `Evaluator.Reflection.cs` | old pattern 1→0, new pattern 0→1 | **SURVIVED 734/734** |
| **C** — positive control, disable `Length` dispatch (`"Length"` → `"LengthXX"`) | `GetReflective` | `LengthXX` = 1 | **DIED — 8 failures** |

Control C kills `Issue3114ReadOnlySpanInterpreterTests.ShippedSpanSample_MatchesGoldenAcrossDrivers` and `.SpanConversionMutationAndSlicing_AgreeAcrossDrivers`, so the suite genuinely reaches this file and the harness can kill mutants in it. **A and B are real survivals, not no-results.** Exact restore verified: `git diff -- src/` empty and 734/734 round-trip.

The reason they cannot be killed is that the binder refuses these programs first — measured, all three drivers agreeing:

| probe | emit | eval | gsi |
|---|---|---|---|
| write through `ReadOnlySpan` (`s[0] = 99`) | rc=1 **GS0226** | rc=1 GS0226 | rc=1 GS0226 |
| `ReadOnlySpan` → `Span` conversion | rc=1 **GS0155** | rc=1 GS0155 | rc=1 GS0155 |

So `Span`/`ReadOnlySpan` distinctness **is** correctly enforced — by the binder, not by this code. The "readonly guard" and "`Span`→`ReadOnlySpan`" entries in the 10/10 mutation claim are therefore not supported; those two guards are unreachable defence-in-depth. Please either drop the claim or note them as unreachable. The bounds checks, by contrast, **are** live and correct: out-of-range index and out-of-range `Slice` both throw (`IndexOutOfRangeException` / `ArgumentOutOfRangeException` under emit, `GS9999` under both interpreting columns), so `CheckedIndex` and `Slice` are genuinely exercised.

## SHOULD-FIX 2 — the `GS0511` docs now describe the opposite of the behaviour

`docs/diagnostics.md:1172` and `website/docs/ref/diagnostics.md:1873` (hand-diffed, byte-identical rows) both still read:

> The interpreter cannot represent a stack-only (`ByRefLike`) CLR value such as `Span[T]` or `ReadOnlySpan[T]`.

with prose "Compile the program with `gsc /out:<file>` to use these types." After this PR, array-backed `Span[T]` and `ReadOnlySpan[T]` — the two types named as the examples — work in the interpreter. The diff touches **no** documentation. Because the docs gate is bidirectional on **row presence only**, and no row was added or removed, the gate stays green while the text is now wrong.

The surviving boundary is shape-dependent and should be stated as such: array-backed spans work; spans obtained from a CLR method or property still report `GS0511`, e.g. `samples/RefStructSpan.gs:19` (`text.AsSpan()` → `ReadOnlySpan[char]`).

Related and correct: deleting the `ReadOnlySpan` theory case from `Issue2987ByRefLikeBoundaryInterpreterTests.cs` is right, since that shape genuinely is no longer refused, and the four remaining cases still pin the method/property routes. `samples/SpanIndexing.gs` comment removal is right; `samples/RefStructSpan.gs` correctly keeps its comment.

## NIT — suite numbers were measured on a staler base

The PR reports `Interpreter.Tests` 725. On the merged tree at `main` `f7d9a5d3` I measure **734/734**. `main` moved twice under this branch (#3052, #3123); worth refreshing the reported figures.

---

### Disclosed as not run
`Compiler.Tests`, `Core.Tests`, LSP / SDK / Extensions / Analyzers were not re-run on the merged tree; e2e is prohibited locally. The 122-sample vector runs were taken while 1-minute load read 36.7 — above the build threshold, but these are probes, and rc/diagnostic identity is load-insensitive. All builds were taken at load 18–27.

DavidObando:
## Rubber-duck review — PR #3128, round 2

**Verdict: NOT MERGEABLE.** 1 BLOCKER, 0 SHOULD-FIX, 2 NIT.

Round 1's BLOCKER is fixed **for the shape it was reported in**, and both SHOULD-FIXes and the NIT are resolved. But the same defect class survives for one element type, and it is the type most likely to appear in real string-handling code.

Re-anchored from scratch. Nothing from round 1 was reused: `origin/main` fetched and verified **`f6f8acc0`**, PR head **`b47c46d3`**, merge base **`2dea578b`**. `git merge-tree --write-tree` rc 0 with its **output read** (1 line, tree `f1575fb7`, 0 conflict lines) — a floor, so the merge was actually performed: merged head **`937c32bd`**, 8 files +417/−36, `BUILD_RC=0`, 0 warnings. Base `f6f8acc0` built separately, `BUILD_RC=0`, 0 warnings. Drivers epoch-mtime verified (merged 06:16:59–06:17:05, base 06:18:39–06:18:46), `sort -n`, `Core`/`Compiler`/`Repl`, excluding `ref/`.

---

## BLOCKER — `char` spans still print the wrapper format, where the emit oracle prints the contents

`src/Core/CodeAnalysis/Evaluator.Reflection.cs:265-266`:

```csharp
public override string ToString()
    => $"System.{(IsReadOnly ? "ReadOnlySpan" : "Span")}<{this.array.GetType().GetElementType().Name}>[{Length}]";
```

That is the correct BCL format for *most* element types, but `Span<char>.ToString()` and `ReadOnlySpan<char>.ToString()` are a documented BCL special case: they return the **characters**, not the type shape. The override has no `char` branch, so the interpreter disagrees with the emitter at rc=0.

```
import System

func Main() {
    var values = []char{'a', 'b', 'c'}
    var s ReadOnlySpan[char] = values
    Console.WriteLine(s.ToString())
}
```

| probe | base `f6f8acc0` | merged `937c32bd` | emit **(oracle)** |
|---|---|---|---|
| `ReadOnlySpan[int32]` | rc=1 `GS0511` | rc=0 `System.ReadOnlySpan<Int32>[3]` | `System.ReadOnlySpan<Int32>[3]` — **match** |
| `Span[char]` | rc=1 `GS0511` | rc=0 `System.Span<Char>[3]` | **`abc`** — **DIVERGES** |
| `ReadOnlySpan[char]` | rc=1 `GS0511` | rc=0 `System.ReadOnlySpan<Char>[3]` | **`abc`** — **DIVERGES** |

All rcs captured before any pipe, on both interpreting drivers, each probe in its own asserted-empty directory.

**This is the round-1 BLOCKER, unfixed for `char`, and it is a regression in exactly the same direction.** On base both programs are a loud, documented `GS0511` refusal. On the head they succeed and print the wrong string. A loud refusal became a **silent wrong answer that disagrees with the emit oracle at rc=0** — the precise failure mode round 1 raised.

**It is invisible to the suite.** The new `test/Interpreter.Tests/Issue3114ReadOnlySpanInterpreterTests.cs` contains **zero** occurrences of `char`; every span in it is `Span[int32]` (4 occurrences). The 14-case matrix is green and cannot see this.

**Corroborating measurement — a second product mutant on the same hunk SURVIVED.** Replacing the span length with the backing-array length (`{Length}` → `{this.array.Length}`) built at `M2_BUILD_RC=0` and left **14/14 green**. I verified the mutant is behaviour-changing rather than semantically equivalent: under it, a `Slice(1,3)` of a 5-element array prints `System.Span<Int32>[5]` against the oracle's `System.Span<Int32>[3]`. So the new `ToString()` is pinned **only** for the single un-sliced `Span[int32]` shape; both the element-type specialisation and the length are unpinned. Reverted to drift 0 and rebuilt, with the slice output confirmed back to `[3]`.

**Recommended fix (small, in keeping with the rest of this PR).** One branch, matching the BCL:

```csharp
public override string ToString()
    => this.array is char[] chars
        ? new string(chars, this.start, Length)
        : $"System.{(IsReadOnly ? "ReadOnlySpan" : "Span")}<{...}>[{Length}]";
```

and extend the driver-matrix test with a `char` case and a **sliced** case — the two shapes the current matrix cannot distinguish.

---

## Round-1 items — all resolved

**Round-1 BLOCKER (the `int32` shape): FIXED.** Re-ran the round-1 repro verbatim before any variation. All three drivers byte-identical:

| driver | output |
|---|---|
| emit (oracle) | rc=0 `System.Span<Int32>[3]` |
| bare `gsc` | rc=0 `System.Span<Int32>[3]` |
| `gsi` | rc=0 `System.Span<Int32>[3]` |

The new hunk is pinned for this shape: the selector-inversion product mutant (`IsReadOnly ? "ReadOnlySpan" : "Span"` → inverted), asserted applied (1→0 original, 1 mutant) and built at `MUTANT_BUILD_RC=0`, killed **4 of 14** — `SpanInterpolation_UsesPublicSpanName` and `SpanConversionMutationAndSlicing_AgreeAcrossDrivers`, both driver arms — exactly matching the author's reported count.

**SHOULD-FIX 1 (unsupported 10/10 mutation claim): RESOLVED.** The PR body now states plainly that two guards survive, that they are unreachable through binder-approved source, and that "the earlier 10/10 claim was incorrect". That matches what I measured in round 1.

**SHOULD-FIX 2 (`GS0511` docs described the opposite of the behaviour): RESOLVED.** `docs/diagnostics.md` and `website/docs/ref/diagnostics.md` now read "Array-backed `Span[T]` and `ReadOnlySpan[T]` values are supported… CLR methods and properties returning stack-only values remain unsupported." I hand-diffed the two `GS0511` blocks rather than trusting the presence-only docs gate: **identical**.

**Round-1 NIT (stale suite numbers): RESOLVED**, and worth restating as a general point — a suite count is anchor-dependent and not a stable identifier. `Interpreter.Tests` was 734 at the round-1 anchor and is **1227** here.

---

## `b47c46d3` "Update Span sample expectations" — investigated as requested, and it is exonerated

Changing expected values is how a divergence gets laundered into a green test, so I checked the direction against the **emit oracle and the golden files**, not against the previous expectation.

The commit removes three entries from `ExpectedDifferences` in `test/Compiler.Tests/LanguageConformance/SampleConformanceTests.cs`. Those entries were **waivers** ("this sample is expected to differ, because `GS0511`"). Removing a waiver makes the assertion **stricter**, not weaker: the sample must now match. Verified independently, each in its own directory:

| sample | bare `gsc` | `gsi` | emit | `golden == emit` | `emit == bare` | `emit == gsi` |
|---|---|---|---|---|---|---|
| `RefStructGenericField.gs` | rc0 | rc0 | rc0 | YES | YES | YES |
| `SpanComprehensive.gs` | rc0 | rc0 | rc0 | YES | YES | YES |
| `SpanIndexing.gs` | rc0 | rc0 | rc0 | YES | YES | YES |

Byte-identical across all three drivers and against the checked-in golden. The product got correct; the assertion was not lowered.

Stronger than that — **the removal was compelled, not chosen.** At `:125-128` the no-longer-needed waiver is itself an error:

```csharp
Assert.False(
    Equivalent(emitted, expectedCompiler) && Equivalent(emitted, expectedInterpreter),
    $"{sampleName} has a redundant expected-difference entry …");
```

so leaving the three stale entries in place would have **failed** the suite. The waiver table is self-cleaning. And at `:136-139`, a sample with no waiver is asserted equal to the emit oracle on both drivers, so an *undeclared* divergence still fails. Both directions are guarded.

The same commit also deletes an `Assert.True(ExpectedDifferences.ContainsKey(sample), …)` from `ExplicitMainSamples_AreDeclaredAndCannotPassVacuously`. I looked at this closely, because deleting a guard on the argument that a later check subsumes it is the reasoning that most often fails. Here the deletion is **necessary**: that assertion required *every* explicit-`Main` sample to carry a waiver, which is directly contradictory with `:125` once a sample legitimately matches — the two conditions are unsatisfiable together for a now-passing sample. The vacuity half (`golden` non-empty) and the sample-set pin (`Assert.Equal(MainOnlySamples, actual)`) both survive.

---

## Suites — 8 of 9 projects, enumerated from `GSharp.sln`

| project | result |
|---|---|
| Core.Tests | 7113 passed / 0 failed / 1 skipped (7114) |
| Compiler.Tests | 4226 / 0 (1 h 27 m) |
| Cs2Gs.Tests | 1903 / 0 |
| Interpreter.Tests | 1227 / 0 |
| LanguageServer.Tests | 462 / 0 |
| Extensions.Tests | 104 / 0 |
| GeneratorHost.Tests | 31 / 0 |
| InternalAnalyzers.Tests | 7 / 0 |
| **Sdk.Tests** | **NOT RUN — deliberately** |

**15,074 tests across the 8 projects run, 0 failures, 1 skip.**

`Sdk.Tests` is stated as NOT RUN rather than implied as covered: it writes the shared local package `~/.nuget/packages/gsharp.net.sdk/`, which other concurrent builds read, so running it risks corrupting them. That is a fleet constraint, not a judgement about the tests.

---

## NITs

**NIT 1 — `ExplicitMainSamples_AreDeclaredAndCannotPassVacuously` now overstates its name.** After the deletion the "AreDeclared" half is carried only by the `MainOnlySamples` list pin, not by a waiver requirement. The behaviour is right; the name is now slightly misleading to the next reader.

**NIT 2 — the PR body reports `SDK: 59 passed`.** Per the constraint above, running `Sdk.Tests` writes a package that other concurrent agents read. Worth dropping from the validation routine rather than reporting it as coverage.

---

### Method notes

Issue repro run verbatim before any variation. Product mutants only, condition/value inversion, never `if (true)`; each asserted applied by pattern count before and after; each read `BUILD_RC` before its result was believed, since a no-result is neither a survival nor a kill; mutated per hunk; every mutation reverted to `git status --porcelain` drift 0, rebuilt, and round-tripped by re-measuring the probe. Every probe in its own asserted-empty directory with distinct values. I re-measured all exit codes after noticing an early capture had taken `$?` from the end of a pipe rather than from the driver. `uptime` before every build; load 13.5–28. No `git stash`, no `GITHUB_ACTIONS=true`, `~/.nuget/packages/gsharp.net.sdk/` untouched, no local e2e, no shared checkout used for any source evidence — every read came from `/tmp/w28` or `/tmp/b28`, both built here with verified SHAs.

DavidObando:
## Rubber-duck review — PR #3128, round 3

**Verdict: MERGEABLE.** 0 BLOCKER, 1 SHOULD-FIX, 1 NIT, 1 retraction of my own round-2 NIT.

**The round-2 BLOCKER is fully dead**, on both axes — I checked both, because a commit titled "Match char span string formatting" sounds like it addresses only one, and this PR had been half-fixed twice. It addresses both.

Re-anchored from scratch, nothing reused from rounds 1–2. Fetched: `origin/main` **`61fe9cb9`**, PR head **`d9eee3bc`**, merge base **`61fe9cb9`** — merge base *equals* main, so the branch is already rebased and the merged tree is the PR head itself. `git merge-tree --write-tree` **output** read (1 line, tree `52f74eb9`, 0 conflict lines). PR tree and base tree built separately: `PR_BUILD_RC=0`, `BASE_BUILD_RC=0`, 0 warnings each, both drift 0. All 6 first-party dlls in `Core`/`Compiler`/`Repl` (excluding `ref/`) fresh within seconds of their build, epoch mtimes, `sort -n`.

---

## The BLOCKER is dead — 18 measurements, all byte-identical

Six shapes × three drivers, each in its own asserted-empty directory, **rc captured before any pipe**, distinct values chosen so a wrong `start` and a wrong `length` produce *different* wrong answers (5-element arrays, `Slice(1, 3)`).

| probe | base `61fe9cb9` | bare `gsc` | `gsi` | emit **(oracle)** | verdict |
|---|---|---|---|---|---|
| `Span[char]` | rc1 `GS0511` | rc0 `abc` | rc0 `abc` | rc0 `abc` | **match** |
| `ReadOnlySpan[char]` | rc1 `GS0511` | rc0 `abc` | rc0 `abc` | rc0 `abc` | **match** |
| `ReadOnlySpan[int32]` | rc1 `GS0511` | rc0 `System.ReadOnlySpan<Int32>[3]` | same | same | **match** |
| `Span[int32]` sliced `(1,3)` of 5 | rc1 `GS0511` | rc0 `System.Span<Int32>[3]` | same | same | **match** |
| `Span[char]` sliced `(1,3)` of 5 | rc1 `GS0511` | rc0 `bcd` | rc0 `bcd` | rc0 `bcd` | **match** |
| `ReadOnlySpan[char]` sliced `(1,3)` of 5 | rc1 `GS0511` | rc0 `bcd` | rc0 `bcd` | rc0 `bcd` | **match** |

`bcd` is the load-bearing value: it proves the char branch honours **both** `this.start` and `Length`, not just one. A wrong start would read `abc`, a wrong length `bcde`. The fix at `src/Core/CodeAnalysis/Evaluator.Reflection.cs:265-268` is the minimal two-line form.

**The `GS0511` refusal path survived where it should** (round-3 bar item 4 — the concern being that the first attempt converted a loud refusal into a wrong answer, and the fix could do the same again elsewhere). A CLR-produced span is still refused loudly on both interpreting drivers:

```
var text = "hello"
var s = text.AsSpan()      →  bare gsc rc=1 GS0511,  gsi rc=1 GS0511
```

No refusal was traded for a wrong answer. Base refuses all six probe shapes loudly; the head answers all six correctly. That is a clean strict improvement in both directions.

---

## My round-2 `{Length}` survivor is now KILLED

Re-run exactly as before. `{Length}` → `{this.array.Length}` in the non-char branch, asserted applied (pre=1, post_orig=0, post_mut=1), `M1_BUILD_RC=0`:

**Failed: 2, Passed: 12 of 14** — `SpanConversionMutationAndSlicing_AgreeAcrossDrivers`, both driver arms.

The kill comes from the new `Console.WriteLine(window.ToString())` at `:110`, where `window` is `readOnly.Slice(1).Slice(0, 2)` over a 3-element array and now asserts `System.ReadOnlySpan<Int32>[2]`. The backing array is length 3, so the mutant reads `[3]` and dies. The length axis is genuinely closed. Reverted to drift 0, rebuilt (`REVERT_BUILD_RC=0`), revert round-tripped by re-running the probe.

---

## SHOULD-FIX — the char branch's `start` is correct but unpinned

`src/Core/CodeAnalysis/Evaluator.Reflection.cs:268`. Product mutant `new string(chars, this.start, Length)` → `new string(chars, 0, Length)`, asserted applied (pre=1, post_orig=0, post_mut=1), `M2_BUILD_RC=0`:

**SURVIVES 14/14.**

And it is a live survivor, not an equivalent one — I verified it changes behaviour before scoring it:

```
sliced char span, Slice(1,3) of ['a','b','c','d','e']
   mutant : rc0  "abc"
   oracle : rc0  "bcd"
```

So a silent wrong answer at rc=0, undetected by the suite. **This is a coverage gap, not a defect** — the shipped product is correct in all six measured cells. I am deliberately *not* calling it a BLOCKER, and the distinction from round 2 matters: in round 2 the product itself was measurably wrong against the oracle. Here nothing is wrong; what is missing is the pin that would stop a future edit regressing it silently.

**Cell enumeration** (bar item 3 — stating scope rather than implying it). The `.ToString()` assertions in `test/Interpreter.Tests/Issue3114ReadOnlySpanInterpreterTests.cs:108-112` cover **5 of 8** cells:

| element | sliced | `Span` | `ReadOnlySpan` |
|---|---|---|---|
| `int32` | no | ✓ `:108` | ✓ `:109` |
| `int32` | yes | — | ✓ `:110` |
| `char` | no | ✓ `:111` | ✓ `:112` |
| `char` | **yes** | **✗** | **✗** |

The missing row is exactly the surviving mutant. Note the same uncovered cell would also hide a `Length` error inside the char branch, since both char spans in the test have `start = 0` and `Length = array.Length` — by inspection of the fixture, not a separate measurement.

Severity is bounded, and I want to be fair about it: `this.start` has four consumers (`:251 As`, `:263 Slice`, `:278 CheckedIndex`, `:268 ToString`), and three are pinned — `CheckedIndex` by the `window[1]` assertion. Only its use in the char branch is unpinned.

**Fix is one line** — change one of the two existing char spans to a sliced one, mirroring what `window` already does for `int32`:

```
var slicedChars = readOnlyChars.Slice(1)
Console.WriteLine(slicedChars.ToString())     // "bc"
```

---

## NIT

`ExplicitMainSamples_AreDeclaredAndCannotPassVacuously` still overstates its name. After `b47c46d3` the "AreDeclared" half is carried only by the `MainOnlySamples` list pin at `SampleConformanceTests.cs:184`; the vacuity half at `:189` is intact. Behaviour is right, name misleads the next reader. Carried from round 2, unchanged.

## Retraction of my own round-2 NIT

In round 2 I raised the PR body's `SDK: 59 passed` as a process NIT, on the understanding that `Sdk.Tests` writes the shared local SDK package. **That was wrong and I withdraw it.** `Sdk.Tests` runs `Program.Main` in-process and writes no package; the sentence that suggested otherwise is prose inside a `<para>` docstring describing other acceptance tests. The author was right to run it and I was wrong to flag it. I ran it myself this round: **59/59**.

---

## Suites — 9 of 9 projects, none skipped

Denominator from `dotnet sln GSharp.sln list` filtered to `*.tests.csproj` — which is byte-for-byte the method CI itself uses at `build/generate-ci-test-matrix.py:22-32`, not `ls test/`.

| project | result |
|---|---|
| Core.Tests | 7116 / 0, 1 skipped (7117) |
| Compiler.Tests | 4232 / 0 |
| Cs2Gs.Tests | 1903 / 0 |
| Interpreter.Tests | 1233 / 0 |
| LanguageServer.Tests | 462 / 0 |
| Extensions.Tests | 104 / 0 |
| InternalAnalyzers.Tests | 7 / 0 |
| GeneratorHost.Tests | 31 / 0 |
| Sdk.Tests | 59 / 0 |

**15,148 tests across 9 of 9 projects, 0 failures, 1 skip.** `Core.Tests` totals **7117**, which is the run figure, not the 6,990 that `--list-tests` reports — only a run is a denominator.

`Interpreter.Tests` is 1233 here against 1227 at the round-2 anchor — a suite count is anchor-dependent and not a stable identifier, and `main` moved from `f6f8acc0` to `61fe9cb9` between rounds.

---

### Settled in round 2, not re-litigated

`b47c46d3` remains exonerated: it removes waivers, making the gate stricter; removal was compelled by `SampleConformanceTests.cs:125-128` (a redundant waiver is itself an error) with `:136-139` guarding the other direction. Round-1 SHOULD-FIX 1 (mutation claim retracted) and SHOULD-FIX 2 (docs, both files verified identical) remain resolved.

### Method notes

Product mutants only, value/condition mutation, never `if (true)`; each asserted applied by pattern count before and after; each read its `BUILD_RC` before any result was believed; the survivor was proven behaviour-changing before being scored, since a no-result is neither a survival nor a kill. Mutated per hunk. Every mutation reverted to drift 0, rebuilt, and round-tripped by re-running the probe. Every probe in its own asserted-empty directory. `uptime` before every build (17–23 throughout, under the ceiling). No `git stash`, no `--no-build` on an unbuilt tree, no `GITHUB_ACTIONS=true`, `~/.nuget/packages/gsharp.net.sdk/` never written. No source evidence taken from the shared checkout — every read came from `/tmp/w28c` or `/tmp/b28c`, both built here at verified, freshly fetched SHAs.

PR #3129, opened 2026-08-03
Fix constant-pattern equality across drivers

### LINKED ISSUE #3110: Evaluator and gsi match NaN against a constant pattern, silently taking the wrong switch arm
## Summary

A `switch` whose discriminant is a **non-literal** `float32`/`float64` NaN incorrectly **matches** a `case Double.NaN` arm under the two interpreting drivers. IEEE 754 requires `NaN != NaN`, so no arm may match and control must fall to `default`. The emitting driver is correct; the **evaluator is wrong**, and because `gsi` shares that path, `gsi` is wrong too.

This is a **silent wrong answer**: rc0, no diagnostic, plausible-looking output.

## Repro

```gsharp
package P
import System
var x = Double.NaN
switch x {
    case Double.NaN { Console.WriteLine("MATCHED") }
    default { Console.WriteLine("DEFAULT") }
}
```

Correct output is `DEFAULT`.

| driver | stdout | rc | |
|---|---|---|---|
| bare `gsc` (EVALUATES) | `MATCHED` | 0 | **WRONG** |
| `gsc /out:` + run (EMITS) | `DEFAULT` | 0 | correct |
| `gsi` (interprets) | `MATCHED` | 0 | **WRONG** |

**The discriminant must not be a literal.** `switch Double.NaN { ... }` written directly is *correct* on all three drivers today, because constant folding added in #2914 folds it correctly at lowering. Any non-literal discriminant — a `var`, a call result, a parameter — exposes the defect. A reporter using a literal will conclude "cannot reproduce".

## Scope — measured

| shape | expected | bare `gsc` | `gsc /out:` | `gsi` |
|---|---|---|---|---|
| `float32` NaN | 32 | **31** | 32 | **31** |
| `float64` NaN | 42 | **41** | 42 | **41** |
| `-0.0` vs `0.0` | 51 | 51 | 51 | 51 |
| `+Infinity` | 61 | 61 | 61 | 61 |
| `-Infinity` | 71 | 71 | 71 | 71 |

Both floating-point widths diverge. Negative zero and both infinities are **correct** on all three drivers, so the defect is specific to NaN.

NaN **comparison operators** are correct on all three drivers (`NaN == NaN` → false, `NaN != NaN` → true, `NaN < 1.0` → false, `NaN > 1.0` → false). `Math.*`, arithmetic and conversions are **not** implicated. Only constant-pattern matching in a `switch`.

## Cause

There is exactly one evaluation site and one emit site:

- `src/Core/CodeAnalysis/Evaluator.Concurrency.cs:126`
  ```csharp
  return object.Equals(value, EvaluateExpression(((BoundConstantPattern)pattern).Value));
  ```
- `src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Patterns.cs:201` — `EmitConstantPattern` (correct)

`object.Equals` differs from IEEE `==` **only** for NaN, which is exactly why `-0.0` and both infinities measure correct and only NaN measures wrong. #2914 hit the same fork and special-cased `float32`/`float64` at `Lowerer.cs:153`; the evaluator never received the equivalent treatment.

## Acceptance criteria

- All three drivers agree with `gsc /out:` for `case <NaN>` against a non-literal discriminant, at both `float32` and `float64`.
- `-0.0`/`0.0` and both infinities remain correct — they are the regression risk of any `object.Equals` replacement.
- Non-floating-point constant patterns (integers, strings, enums, `nil`, `bool`) are unaffected; `object.Equals` is correct for those and must stay correct, including boxed and nullable shapes.
- The fix is at the shared evaluator site so that bare `gsc` and `gsi` are both corrected by one change.

### LINKED ISSUE #3113: Emitter silently fails constant-pattern match for boxed value types and 64-bit nullables
## Summary

`gsc /out:` **silently fails to match** a constant pattern when the discriminant is a **boxed value type** or a **64-bit-payload nullable**. The arm that should match is skipped and control falls to `default`, at rc0 with no diagnostic. The evaluator and `gsi` are **correct** in all of these cases — this is emitter-only, and it is the opposite direction to #3110.

## Measured, `main`, all probes top-level, each into its own empty directory

Each arm prints a value that names the arm, so a wrong pick prints the wrong number rather than silently agreeing.

| # | shape | expected | bare `gsc` (EVAL) | `gsc /out:` (EMIT) | `gsi` |
|---|---|---|---|---|---|
| 12 | `object` = 1 (boxed `int32`), `case 1` | 11 | 11 | **33** | 11 |
| 13 | `object` = true (boxed `bool`), `case true` | 11 | 11 | **33** | 11 |
| 14 | `object` = `"s"` (boxed string), `case "s"` | 11 | 11 | 11 | 11 |
| 15 | `int64?` = 3, `case 3` | 11 | 11 | **33** | 11 |
| 16 | `float64?` = 1.5, `case 1.5` | 11 | 11 | **33** | 11 |
| 17 | `float64?` = nil, `case nil` | 11 | 11 | **33** | 11 |

`33` is the `default` arm.

### Two independent defects, and the second one is precisely isolated

**Boxed value types (rows 12–13).** Row 14 is the control that makes this a defect rather than a design choice: the emitter **does** match a boxed `string` but **not** a boxed `int32` or `bool`. "Constant patterns do not match through `object`" is therefore not the intended rule — the behaviour is internally inconsistent.

**Nullables are width-selective (rows 15–17).** With **and** without a `case nil` arm present, `int32?`, `char?` and `bool?` match correctly while `int64?` and `float64?` do not. **Exactly the 64-bit payloads fail.** An initial hypothesis that the `case nil` arm was the trigger was tested and disproved.

## Correct today, and must stay correct

`int32`/`int64`/`uint32`/`uint64`/`decimal` unboxed; `string` including empty and `nil`; `bool`; `char`; enum; `int32?`; `char?`; `bool?`; `-0.0` vs `0.0`; `±Infinity`. All three drivers agree on every one of these.

## Interaction with #3110

#3110 is the mirror image: there the **evaluator** is wrong and the **emitter** is right (NaN matching itself via `object.Equals`). Whoever fixes either must not "harmonise" the two paths toward one another, because **each is wrong in cases where the other is right**. Specifically, do not make the evaluator agree with the emitter on rows 12–17.

## Not measurable as written

`case int64(1)` and `case float32(1.5)` are rejected — a conversion is not a legal constant pattern (`GS0005`) — so a 64-bit-typed or `float32`-typed constant pattern cannot be written directly, only reached via a member such as `Double.NaN` or `Decimal.One`. `int8?`/`int16?`/`uint32?`/`uint64?` reject a bare literal initialiser (`GS0156`), so those widths are **unmeasured, not clean**.

## Acceptance criteria

- Rows 12, 13, 15, 16, 17 match on `gsc /out:` and agree with the other two drivers.
- Every row listed under "correct today" is unchanged on all three drivers.
- The `int8?`/`int16?`/`uint32?`/`uint64?` widths are covered once they can be expressed, or the reason they cannot is recorded.
- #3110 is not regressed: NaN must still fall to `default` on all three drivers.

DavidObando:
## Summary

- Use runtime numeric equality for evaluator constant patterns, preserving IEEE `NaN != NaN` behavior.
- Emit boxed/object and value-type-nullable constant patterns through runtime type-safe equality.
- Select floating comparison from the pattern value's static type, not bound-node shape.
- Cover bare `gsc`, emitted `gsc /out:`, and `gsi`, including computed NaN and chained conversions.

Fixes #3110
Fixes #3113

## Why one wrapper is correct

`PatternBinder.BindConstantPattern` first binds the source expression, then introduces exactly zero or one carrier conversion to the discriminant type:

```csharp
var value = conversion.IsIdentity
    ? expression
    : conversions.BindConversion(syntax.Expression.Location, expression, discriminantType);
```

The emitter removes only that outer conversion when its type equals the discriminant type. Every remaining `BoundExpression` has a static `Type`, so literal, folded, unary, binary, field, and other node shapes need no enumeration.

Unwrapping to a fixed point would be wrong: inner conversions are semantic. The pinned specimen `float64(float32(1.00000001))` rounds through `float32`; stripping both conversions changes the compared value. A PRODUCT mutant that unwrapped every `BoundConversionExpression` built successfully and failed only emitted output (`33` instead of `11`), while bare `gsc` and `gsi` remained `11`. The paired chained-NaN specimen remains unequal across all drivers.

## Three-driver measurement

Original four-case blocker corpus: division, multiplication, unary negation, and nullable division producing NaN.

| tree | bare `gsc` | `gsc /out:` | `gsi` |
|---|---|---|---|
| current main `61fe9cb9` | `11 11 11 11` | `33 33 33 33` | `11 11 11 11` |
| fixed `82da671d` | `33 33 33 33` | `33 33 33 33` | `33 33 33 33` |

Chained conversion pair (`NaN`, precision-sensitive non-NaN control):

| tree | bare `gsc` | `gsc /out:` | `gsi` |
|---|---|---|---|
| current main `61fe9cb9` | `11 / 11` | `33 / 33` | `11 / 11` |
| fixed `82da671d` | `33 / 11` | `33 / 11` | `33 / 11` |

Values name arms: `11` match, `22` alternate, `33` default. Emit is oracle.

## Product mutations

Each mutant was asserted applied, built to test execution, produced RED, then was restored. Restored final tree passed 3/3 targeted tests.

| hunk mutant | killed by | RED measurement |
|---|---|---|
| evaluator `NumericEquals` → `object.Equals` | `NonLiteralNaNConstants_AgreeAcrossDrivers` | bare/gsi NaN rows became `11`; emit remained correct |
| remove object route by condition inversion | `TwentyRowCorpus_AgreesAcrossDrivers` | emitted boxed int32/bool rows became `33` |
| remove value-nullable route by condition inversion | `TwentyRowCorpus_AgreesAcrossDrivers` | emitted int64?/float64?/nil float64? rows became `33` |
| invert floating static-type route (control mutant) | `NonLiteralNaNConstants_AgreeAcrossDrivers` | emitted five NaN rows became `11` |
| unwrap conversions to fixed point | `NonLiteralNaNConstants_AgreeAcrossDrivers` | emitted precision control became `33`; bare/gsi stayed `11` |

Current main without this PR independently supplies the opposite-side product controls shown above: emit keeps IEEE NaN output `33`, while boxed and nullable equality remains broken.

## Validation

- Rebased onto `origin/main` `61fe9cb9`.
- Head `82da671d`.
- `git merge-tree --write-tree origin/main HEAD`: `1f0c6fc54df6516e436d09e7b08ba5a40636be6d`, equal to `HEAD^{tree}`.
- Release solution build: `BUILD_RC=0`, 0 warnings, 0 errors.
- Final targeted tests: 3 passed, 0 failed.
- Original local final-tree test-project coverage was 1 of 9, enumerated from `GSharp.sln` using the same source as `build/generate-ci-test-matrix.py`. No Oahu coverage is claimed from local `Cs2Gs.Tests`.

## Addendum: `Sdk.Tests` correction

Earlier fleet guidance incorrectly classified `Sdk.Tests` as out of bounds. It runs SDK task logic in-process and does not spawn `dotnet build` against the shared SDK package directory. Unfiltered `Sdk.Tests` on head `82da671d`: 59 passed, 0 failed, 0 skipped (`TEST_RC=0`). Local final-tree coverage is therefore 2 of 9 projects; remaining seven were not rerun locally after the final rebase. CI's complete nine-project matrix passed on this head.

### DIFF
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Patterns.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Patterns.cs
@@ -229,6 +229,60 @@ internal sealed partial class MethodBodyEmitter
             return;
         }
 
+        var isObject = valueType?.ClrType.IsSameAs(typeof(object)) == true;
+        var isValueNullable = valueType is NullableTypeSymbol nullable
+            && ReflectionMetadataEmitter.IsValueTypeNullable(nullable);
+        if (isObject || isValueNullable)
+        {
+            // PatternBinder adds at most one conversion to the discriminant type.
+            // Strip only that carrier; inner conversions determine the runtime value type.
+            var patternValue = cp.Value is BoundConversionExpression conversion && conversion.Type == valueType
+                ? conversion.Expression
+                : cp.Value;
+            if (patternValue.Type == TypeSymbol.Float32 || patternValue.Type == TypeSymbol.Float64)
+            {
+                loadValue();
+                if (isValueNullable)
+                {
+                    this.il.OpCode(ILOpCode.Box);
+                    this.il.Token(this.outer.memberRefs.GetElementTypeToken(valueType));
+                }
+
+                this.il.OpCode(ILOpCode.Isinst);
+                this.il.Token(this.outer.memberRefs.GetElementTypeToken(patternValue.Type));
+                var typeMatched = this.il.DefineLabel();
+                this.il.OpCode(ILOpCode.Dup);
+                this.il.Branch(ILOpCode.Brtrue, typeMatched);
+                this.il.OpCode(ILOpCode.Pop);
+                this.il.Branch(ILOpCode.Br, failLabel);
+                this.il.MarkLabel(typeMatched);
+                this.il.OpCode(ILOpCode.Unbox_any);
+                this.il.Token(this.outer.memberRefs.GetElementTypeToken(patternValue.Type));
+                this.EmitExpression(patternValue);
+                this.il.OpCode(ILOpCode.Ceq);
+                this.il.Branch(ILOpCode.Brfalse, failLabel);
+                return;
+            }
+
+            loadValue();
+            if (isValueNullable)
+            {
+                this.il.OpCode(ILOpCode.Box);
+                this.il.Token(this.outer.memberRefs.GetElementTypeToken(valueType));
+            }
+
+            this.EmitExpression(cp.Value);
+            if (isValueNullable)
+            {
+                this.il.OpCode(ILOpCode.Box);
+                this.il.Token(this.outer.memberRefs.GetElementTypeToken(valueType));
+            }
+
+            this.il.Call(this.outer.wellKnown.GetObjectStaticEqualsReference());
+            this.il.Branch(ILOpCode.Brfalse, failLabel);
+            return;
+        }
+
         // int / bool / other primitives lowered to ceq + brfalse.
         loadValue();
         this.EmitExpression(cp.Value);
--- a/src/Core/CodeAnalysis/Evaluator.Concurrency.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Concurrency.cs
@@ -123,7 +123,7 @@ public sealed partial class Evaluator
         switch (pattern.Kind)
         {
             case BoundNodeKind.ConstantPattern:
-                return object.Equals(value, EvaluateExpression(((BoundConstantPattern)pattern).Value));
+                return NumericEquals(value, EvaluateExpression(((BoundConstantPattern)pattern).Value));
             case BoundNodeKind.DiscardPattern:
                 return true;
             case BoundNodeKind.TypePattern:
--- a/test/Interpreter.Tests/Issue3110And3113ConstantPatternEqualityTests.cs
+++ b/test/Interpreter.Tests/Issue3110And3113ConstantPatternEqualityTests.cs
@@ -1,0 +1,373 @@
+// <copyright file="Issue3110And3113ConstantPatternEqualityTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace GSharp.Interpreter.Tests;
+
+/// <summary>Issues #3110 and #3113: constant-pattern equality must agree across all three drivers.</summary>
+[Collection("ConsoleIo")]
+public class Issue3110And3113ConstantPatternEqualityTests
+{
+    [Fact]
+    public async Task TwentyRowCorpus_AgreesAcrossDrivers()
+    {
+        const string Expected = """
+            11
+            11
+            11
+            11
+            11
+            11
+            11
+            11
+            11
+            11
+            11
+            11
+            11
+            11
+            11
+            11
+            11
+            11
+            11
+            11
+            """;
+        await AssertDriverMatrixAsync(SourceForTwentyRowCorpus(), Expected, nameof(TwentyRowCorpus_AgreesAcrossDrivers));
+    }
+
+    [Fact]
+    public async Task FloatingAndNullableWidthControls_AgreeAcrossDrivers()
+    {
+        const string Expected = """
+            33
+            33
+            11
+            11
+            11
+            11
+            11
+            11
+            11
+            11
+            11
+            33
+            33
+            33
+            33
+            11
+            11
+            11
+            """;
+        await AssertDriverMatrixAsync(SourceForFloatingAndNullableWidthControls(), Expected, nameof(FloatingAndNullableWidthControls_AgreeAcrossDrivers));
+    }
+
+    [Fact]
+    public async Task NonLiteralNaNConstants_AgreeAcrossDrivers()
+    {
+        const string Expected = """
+            33
+            33
+            33
+            33
+            33
+            11
+            """;
+        await AssertDriverMatrixAsync(SourceForNonLiteralNaNConstants(), Expected, nameof(NonLiteralNaNConstants_AgreeAcrossDrivers));
+    }
+
+    private static async Task AssertDriverMatrixAsync(string source, string expected, string testName)
+    {
+        var root = CreateEmptyDirectory(testName);
+        try
+        {
+            var bare = RunBareGsc(source, CreateEmptyDirectory(root, "bare"));
+            var emitted = await RunEmittedAsync(source, CreateEmptyDirectory(root, "emit"));
+            var interpreted = RunGsi(source, CreateEmptyDirectory(root, "gsi"));
+
+            Assert.True(
+                expected == bare && expected == emitted && expected == interpreted,
+                $"driver mismatch\nexpected:\n{expected}\nbare gsc:\n{bare}\nemitted:\n{emitted}\ngsi:\n{interpreted}");
+        }
+        finally
+        {
+            Directory.Delete(root, recursive: true);
+        }
+    }
+
+    private static string RunBareGsc(string source, string directory)
+    {
+        var sourcePath = Path.Combine(directory, "probe.gs");
+        File.WriteAllText(sourcePath, source);
+        var (exitCode, stdout, stderr) = CaptureConsole(
+            () => GSharp.Compiler.Program.Main([sourcePath]));
+
+        Assert.True(exitCode == 0, $"bare gsc failed ({exitCode})\nstdout:\n{stdout}\nstderr:\n{stderr}");
+        return NormalizeValues(stdout);
+    }
+
+    private static string RunGsi(string source, string directory)
+    {
+        var sourcePath = Path.Combine(directory, "probe.gs");
+        File.WriteAllText(sourcePath, source);
+        var (exitCode, stdout, stderr) = CaptureConsole(
+            () => GSharp.Repl.Program.Main([sourcePath]));
+
+        Assert.True(exitCode == 0, $"gsi failed ({exitCode})\nstdout:\n{stdout}\nstderr:\n{stderr}");
+        return NormalizeValues(stdout);
+    }
+
+    private static async Task<string> RunEmittedAsync(string source, string directory)
+    {
+        var sourcePath = Path.Combine(directory, "probe.gs");
+        var outputPath = Path.Combine(directory, $"Issue3110And3113{Guid.NewGuid():N}.dll");
+        File.WriteAllText(sourcePath, source);
+
+        var (compileExit, stdout, stderr) = CaptureConsole(
+            () => GSharp.Compiler.Program.Main(
+                ["/out:" + outputPath, "/target:exe", "/targetframework:net10.0", sourcePath]));
+        Assert.True(compileExit == 0, $"emit failed ({compileExit})\nstdout:\n{stdout}\nstderr:\n{stderr}");
+
+        var assembly = Assembly.Load(File.ReadAllBytes(outputPath));
+        _ = assembly.GetTypes();
+
+        var startInfo = new ProcessStartInfo("dotnet")
+        {
+            RedirectStandardOutput = true,
+            RedirectStandardError = true,
+            UseShellExecute = false,
+            WorkingDirectory = directory,
+        };
+        startInfo.ArgumentList.Add(outputPath);
+
+        using var process = Process.Start(startInfo);
+        Assert.NotNull(process);
+        var stdoutTask = process.StandardOutput.ReadToEndAsync();
+        var stderrTask = process.StandardError.ReadToEndAsync();
+        using var timeout = new CancellationTokenSource(30_000);
+        try
+        {
+            await process.WaitForExitAsync(timeout.Token);
+        }
+        catch (OperationCanceledException)
+        {
+            process.Kill(entireProcessTree: true);
+            using var killTimeout = new CancellationTokenSource(5_000);
+            try
+            {
+                await process.WaitForExitAsync(killTimeout.Token);
+            }
+            catch (OperationCanceledException)
+            {
+                Assert.Fail("emitted program did not exit after kill");
+            }
+
+            Assert.Fail("emitted program timed out after 30000 ms");
+        }
+
+        var runOut = await stdoutTask;
+        var runErr = await stderrTask;
+        Assert.True(process.ExitCode == 0, $"emitted program failed ({process.ExitCode})\nstdout:\n{runOut}\nstderr:\n{runErr}");
+        return NormalizeValues(runOut);
+    }
+
+    private static (int ExitCode, string Stdout, string Stderr) CaptureConsole(Func<int> action)
+    {
+        using var stdout = new StringWriter();
+        using var stderr = new StringWriter();
+        var previousOut = Console.Out;
+        var previousError = Console.Error;
+        Console.SetOut(stdout);
+        Console.SetError(stderr);
+        try
+        {
+            return (action(), stdout.ToString(), stderr.ToString());
+        }
+        finally
+        {
+            Console.SetOut(previousOut);
+            Console.SetError(previousError);
+        }
+    }
+
+    private static string NormalizeValues(string output) =>
+        string.Join(
+            "\n",
+            output.Replace("\r\n", "\n", StringComparison.Ordinal)
+                .Split('\n', StringSplitOptions.RemoveEmptyEntries)
+                .Where(line => line != "Success."));
+
+    private static string CreateEmptyDirectory(string name)
+    {
+        var parent = Path.Combine(AppContext.BaseDirectory, "issue3110-3113-probes");
+        return CreateEmptyDirectory(parent, $"{name}-{Guid.NewGuid():N}");
+    }
+
+    private static string CreateEmptyDirectory(string parent, string name)
+    {
+        var path = Path.Combine(parent, name);
+        Directory.CreateDirectory(path);
+        Assert.Empty(Directory.EnumerateFileSystemEntries(path));
+        return path;
+    }
+
+    private static string SourceForTwentyRowCorpus() => """
+        package Issue3110And3113.Corpus
+        import System
+
+        enum Shade { Light, Dark }
+
+        var i32 int32 = 1
+        switch i32 { case 1 { Console.WriteLine(11) } case 2 { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var i64 int64 = int64(2)
+        switch i64 { case 2 { Console.WriteLine(11) } case 3 { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var u32 uint32 = uint32(3)
+        switch u32 { case 3 { Console.WriteLine(11) } case 4 { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var u64 uint64 = uint64(4)
+        switch u64 { case 4 { Console.WriteLine(11) } case 5 { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var dec decimal = 1.5m
+        switch dec { case 1.5m { Console.WriteLine(11) } case 2.5m { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var text string = "s"
+        switch text { case "s" { Console.WriteLine(11) } case "t" { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var empty string = ""
+        switch empty { case "" { Console.WriteLine(11) } case "x" { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var nilText string? = nil
+        switch nilText { case nil { Console.WriteLine(11) } case "x" { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var truth bool = true
+        switch truth { case true { Console.WriteLine(11) } case false { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var letter char = 'x'
+        switch letter { case 'x' { Console.WriteLine(11) } case 'y' { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var shade Shade = Shade.Dark
+        switch shade { case Shade.Dark { Console.WriteLine(11) } case Shade.Light { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var boxedInt object = 1
+        switch boxedInt { case 1 { Console.WriteLine(11) } case 2 { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var boxedBool object = true
+        switch boxedBool { case true { Console.WriteLine(11) } case false { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var boxedString object = "s"
+        switch boxedString { case "s" { Console.WriteLine(11) } case "t" { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var nullableInt32 int32? = 1
+        switch nullableInt32 { case 1 { Console.WriteLine(11) } case 2 { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var nullableChar char? = 'x'
+        switch nullableChar { case 'x' { Console.WriteLine(11) } case 'y' { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var nullableBool bool? = true
+        switch nullableBool { case true { Console.WriteLine(11) } case false { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var nullableInt64 int64? = int64(3)
+        switch nullableInt64 { case 3 { Console.WriteLine(11) } case 4 { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var nullableFloat64 float64? = 1.5
+        switch nullableFloat64 { case 1.5 { Console.WriteLine(11) } case 2.5 { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var nilFloat64 float64? = nil
+        switch nilFloat64 { case nil { Console.WriteLine(11) } case 1.5 { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+        """;
+
+    private static string SourceForFloatingAndNullableWidthControls() => """
+        package Issue3110And3113.Controls
+        import System
+
+        enum ControlShade { Light, Dark }
+
+        var nan32 float32 = Single.NaN
+        switch nan32 { case Single.NaN { Console.WriteLine(11) } case Single.PositiveInfinity { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var nan64 float64 = Double.NaN
+        switch nan64 { case Double.NaN { Console.WriteLine(11) } case Double.PositiveInfinity { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var negativeZero float64 = -0.0
+        switch negativeZero { case 0.0 { Console.WriteLine(11) } case 1.0 { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var epsilon32 float32 = Single.Epsilon
+        switch epsilon32 { case Single.Epsilon { Console.WriteLine(11) } case Single.PositiveInfinity { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var positiveInfinity float64 = Double.PositiveInfinity
+        switch positiveInfinity { case Double.PositiveInfinity { Console.WriteLine(11) } case Double.NegativeInfinity { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var negativeInfinity float64 = Double.NegativeInfinity
+        switch negativeInfinity { case Double.NegativeInfinity { Console.WriteLine(11) } case Double.PositiveInfinity { Console.WriteLine(22) } default { Console.WriteLine(33) } }
+
+        var nullableInt8 int8? = int8(1)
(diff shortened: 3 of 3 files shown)


DavidObando:
## Verdict: NOT MERGEABLE — 1 BLOCKER

Anchors: `main` **`f7d9a5d3`**, head **`3294ee2e`**, merge-base `4469e323` (main moved twice under this branch: #3052, #3123). `git merge-tree` **output** tree `2d892715`; a real merge reproduced `2d892715` exactly. Merged tree `BUILD_RC=0`, 0 warnings, Core md5 `01f6dd02`. Base built separately, Core md5 `c9b3a9e4` (round-tripped exactly).

### Explicitly: were the two paths harmonised? **YES — and that is where it breaks.**

Hunk B moves the emitter's object/value-nullable routes onto `Object.Equals`, i.e. onto the evaluator's comparison semantics. `Object.Equals` → `Double.Equals`, which is **reflexive on NaN**. The PR compensates with a compile-time guard, `IsFloatingNaNConstant` (`src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Patterns.cs`). That guard is **incomplete**, so the NaN direction does *not* survive the unification.

## BLOCKER — emit regression: non-literal NaN constants now match

`src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Patterns.cs` — `IsFloatingNaNConstant` strips `BoundConversionExpression` and then **requires a `BoundLiteralExpression`**. Any NaN constant that is a *folded expression* rather than a literal node slips the guard, reaches `Object.Equals`, and matches itself.

Four shapes, `11` = MATCHED, `33` = DEFAULT (IEEE-correct is `33`):

| discriminant / pattern | bare | gsi | **emit** |
|---|---|---|---|
| `object` vs `case 0.0/0.0` | 33 ✓ | 33 ✓ | **11 ✗** |
| `object` vs `case Double.NaN*1.0` | 33 ✓ | 33 ✓ | **11 ✗** |
| `object` vs `case -Double.NaN` | 33 ✓ | 33 ✓ | **11 ✗** |
| `float64?` vs `case 0.0/0.0` | 33 ✓ | 33 ✓ | **11 ✗** |

Same four shapes on base `f7d9a5d3`:

| tree | bare | gsi | **emit** |
|---|---|---|---|
| base `f7d9a5d3` | 11 ✗ | 11 ✗ | **33 ✓** |
| merged | 33 ✓ | 33 ✓ | **11 ✗** |

**The emit column goes correct → incorrect. This is a regression introduced by this PR**, not a pre-existing gap. It hits **both** new routes (`isObject` and `isValueNullable`) and three independent expression forms (binary divide, binary multiply, unary negate). It compiles at rc 0, is silently wrong, and makes the **drivers disagree** — the exact failure class #3110 and #3113 exist to eliminate.

At base the emitter was accidentally right here: the object/nullable routes did not exist, execution fell through to `ceq`, and NaN correctly defaulted. #3113's premise is that this fall-through is a defect — true for boxed `int32`/`bool`, but it was *load-bearing* for NaN. Removing it without a complete guard trades one inconsistency for another.

**Why the corpus cannot see it:** the new test file contains only `case Double.NaN` and `case Single.NaN` — **zero** non-literal NaN constant patterns (`grep` of every NaN form in `test/Interpreter.Tests/Issue3110And3113ConstantPatternEqualityTests.cs` returns exactly those two). #2914's folding turns both into `BoundLiteralExpression`, so the guard always fires and every test passes. The 20-row corpus is structurally blind to the hunk's weakest edge.

**Diagnosis worth keeping:** the evaluator hunk is complete and the emitter hunk is not, for a structural reason. `NumericEquals` (`Evaluator.Concurrency.cs:392`) inspects **runtime values**, so it is correct for every constant form. `IsFloatingNaNConstant` inspects **bound-node shape**, so it is correct only for the forms someone remembered to enumerate.

**Recommended fix:** stop deciding NaN-ness from syntax. In the `isObject || isValueNullable` route, when the constant's **static type** is `float32`/`float64`, emit the IEEE comparison (unbox + `ceq`) rather than `Object.Equals`. That is correct for all constant forms by construction, needs no enumeration, and mirrors why the evaluator side is sound. Failing that, drive the guard off the pattern's folded constant value. Add at least one non-literal NaN row (`case 0.0/0.0`) to the corpus on both the object and nullable routes.

## What I verified positively (all of the author's claims hold)

- **`main` 20-row column reproduced exactly** at the new `f7d9a5d3` — bare/gsi all `11`; emit `33` at rows 12, 13, 18, 19, 20. The author measured at stale `4038d051`; the table is still accurate.
- **Merged tree is fully correct on both published corpora, all three drivers** — 20-row all `11`; 18-row width controls `33,33,11,11,11,11,11,11,11,11,11,33,33,33,33,11,11,11`, which is the IEEE-correct column I derived before running it.
- **Extra-scope claims are genuine defects at base, genuinely fixed:** `uint64?` (`33`→correct), nullable `float32` NaN (wrongly `11` at base), boxed `decimal`/`char`/`enum` (all `33` at base). Not merely asserted.
- **`uint64?` question resolved:** the author *did* find the expression — `var n uint64? = uint64(1)`. An explicit conversion is legal in an initialiser even though `case uint64(1)` is `GS0005` as a pattern.
- **Hunk A is the prescribed fix**, one line, reusing a pre-existing untouched helper; behavioural delta confined to NaN. `-0.0` is unaffected (`Double.Equals(-0.0, 0.0)` and `==` agree). The `==`-on-`object` trap was **not** fallen into — rows 1–17 still match on all drivers.
- Killed my own hypothesis that `GetElementTypeToken` emitted a bad box: `ImportedMemberRefFactory.cs:91-100` returns the `System.Nullable<T>` token, which is correct IL. Misleading name, right behaviour.

## NIT

`IsFloatingNaNConstant` unwraps `BoundConversionExpression` in a loop, which signals awareness that the constant may be wrapped — but stops one node-kind short of the binary/unary forms that actually occur. If the shape-based approach is kept over the recommendation above, that loop is where the omission reads as accidental rather than considered.

*Disclosure: one build in this review started at load 31.01, marginally over the 30 threshold. No e2e run; `~/.nuget/packages/gsharp.net.sdk/` never written; worktree `i3110` never touched; all probes in a private scratch clone, working tree clean.*

DavidObando:
## Addendum — BLOCKER re-anchored at `main` = `5b71a74d` (was `f7d9a5d3`)

`main` has moved **18 commits** since my verdict, and **both files this PR edits changed on main** in that range. Re-measured rather than assumed.

**Why re-anchoring was necessary, not ceremonial:**

| file | touched since `f7d9a5d3`? | by |
|---|---|---|
| `MethodBodyEmitter.Patterns.cs` | **CHANGED** `2739a28f`→`39126d2c` | `2c882c03`, `1978bef1` (property-pattern work) |
| `Evaluator.Concurrency.cs` | **CHANGED** `457688f1`→`7f94a7a6` | `1978bef1`, `27a05798` (#3127) |
| `Lowerer.cs` | **CHANGED** `5e183404`→`882ee320` | `36b28785` (#3107) |
| `ImportedMemberRefFactory.cs` | unchanged | — |

**PR still applies cleanly.** `git merge-tree 5b71a74d pr3129` → tree `2e0441e3` (read the *output*, not the rc). A real merge reproduced `2e0441e3` **exactly**; 3 files, +378/−1. Merged build `BUILD_RC=0`, **0 warnings**.

**The bug this PR fixes is still live on the new main:** `Evaluator.Concurrency.cs:126` still reads `object.Equals(value, …)` verbatim. The evaluator half remains a genuine, correct fix.

**The BLOCKER reproduces identically.** Probe `n.gs` — four non-literal NaN constant patterns (`case 0.0/0.0`, `case Double.NaN*1.0`, `case -Double.NaN` against `object`; `case 0.0/0.0` against `float64?`). `11` = matched, `33` = took `default` (IEEE-correct).

| driver | base `5b71a74d` | merged `2e0441e3` |
|---|---|---|
| bare `gsc` | `11 11 11 11` ✗ | `33 33 33 33` ✓ fixed |
| **`gsc /out:` emit** | **`33 33 33 33` ✓** | **`11 11 11 11` ✗ REGRESSED** |
| `gsi` | `11 11 11 11` ✗ | `33 33 33 33` ✓ fixed |

Byte-identical to the `f7d9a5d3` numbers. The emit column was **correct at base and is wrong after the merge** — still a regression, not a pre-existing defect, on current `main`.

**Guard unchanged in the newly-merged tree** (`MethodBodyEmitter.Patterns.cs:269-278`): strips only `BoundConversionExpression`, then requires `BoundLiteralExpression`. So `0.0/0.0` and `Double.NaN*1.0` (`BoundBinaryExpression`) and `-Double.NaN` (`BoundUnaryExpression`) still slip it.

The asymmetry is the whole finding, and it survived main moving: **`NumericEquals` inspects runtime values, so it is sound by construction; `IsFloatingNaNConstant` inspects bound-node shape, so it is sound only by enumeration.** Recommended fix unchanged — in the `isObject || isValueNullable` route, when the constant's *static* type is `float32`/`float64`, emit unbox + `ceq` rather than `Object.Equals`.

**Verdict unchanged: NOT MERGEABLE.**

---

### Self-disclosure — a near-miss in my own harness

I wrote the probe that found the above as:

```
git log --oneline f7d9a5d3..5b71a74d -- <evidence files>
echo "(empty above = untouched)"
```

The output was **not** empty — four commits, including both PR files. I had **baked the expected answer into the label**, so a glance would have read "untouched" and I would have left the verdict anchored at a stale SHA. That is my own *"a silenced error is an unrun check"* rule, self-inflicted. An assertion whose text presumes its result is not an assertion.

Also, per the nine-project bulletin: my posted verdicts on #3071, #3123 and this PR cite **named** suites with counts and state plainly where a suite was not run locally. None asserts an "N of 7"/"N of 8" denominator, so **no coverage correction is required** — but I confirmed that by re-reading them rather than assuming it. Independently: `GSharp.sln` yields **9** test projects, and `ls test/` is wrong in *both* directions — it misses `tools/cs2gs/Cs2Gs.Tests` and `tools/gsgen/GSharp.GeneratorHost.Tests`, and also picks up four non-projects (`Directory.Build.props`, `Directory.Build.targets`, `Shared`, `xunit.runner.json`).

Builds started at load **22.38** and **25.08** (1-min), both below the 30 threshold.

DavidObando:
## Verdict: NOT MERGEABLE — unchanged tree, blocker stands untested

**No re-review performed, by design.** The author has pushed nothing since the BLOCKER was filed, so there is no new tree to measure.

### Proof of no-push (three independent signals)

| signal | value |
|---|---|
| live `refs/heads/fix/3110-constant-pattern-equality` | `3294ee2e3dd34029bc68f4e0c5382fc8b63ba8d5` |
| live `refs/pull/3129/head` | `3294ee2e3dd34029bc68f4e0c5382fc8b63ba8d5` |
| head SHA cited in the BLOCKER comment (05:01:32Z) | `3294ee2e` |
| timeline push events, full history | **one** `committed`, `3294ee2e`, 01:45:06Z |
| force-pushes | **zero** |

The sole commit predates the BLOCKER by 3h16m. Identical SHA means an identical tree, so the measured emit result — `11 11 11 11` against an oracle of `33 33 33 33`, re-confirmed in the 10:53Z addendum after `main` moved 18 commits — still stands as the current state of this PR.

### The actionable part: the work exists, it is just not pushed

The fix has **not** stalled. It is sitting unpushed:

- local branch head `37b1249f`, 27 commits ahead of `3294ee2e`, of which **26 are `origin/main`** — i.e. a rebase — and **exactly one** is the author's fix commit
- `git patch-id --stable` differs: pushed `63c5409b7ddb…` vs local `9416940961f4…`
- content grew: `MethodBodyEmitter.Patterns.cs` **+52** (pushed: +41), tests **365 lines** (pushed: 336)

So this is neither "author did nothing" nor "ready to review". **Push `37b1249f` and the review can run.**

### One forward-looking note — unverified, do not treat as a finding

The unpushed hunk appears to move dispatch off bound-node *shape* and onto the pattern value's **static type** (`patternValue.Type == TypeSymbol.Float32 || Float64`), emitting `Ceq` for floats and falling back to `object.Equals` otherwise. That is the right direction for the soundness-by-enumeration objection, since the static type set is closed where the shape set was not.

It does **not** dispose of the objection on its own: the code unwraps exactly one wrapper shape, `BoundConversionExpression`, before reading the type. A NaN constant reaching the site behind any other node — a second chained conversion, or a folded constant node — still falls through to `object.Equals`. **That enumeration is still owed** once this is pushed.

I have not built, run, or mutation-tested any of the above. It is a read-only diff observation on a tree that is unpushed, in a worktree I do not own, and already based on `f6f8acc0` while `main` is `faca348a` — it can change before it lands.

### Unchanged from the prior verdict

`Evaluator.Concurrency.cs:126` (`object.Equals` → `NumericEquals`) remains the correct half of this PR and must survive whatever happens to the emit half.

### Nit

The briefing branch name `fix/3110-numeric-equals` does not exist; the branch is `fix/3110-constant-pattern-equality`.

DavidObando:
## Blocker resolved on current main

Rebased onto `main` **`faca348a`**. New head: **`efaa01ed`**. `git merge-tree --write-tree origin/main HEAD` produced `6daddc881cd0ff1ac98c57b9ced5c2259f19f27d`, exactly matching `HEAD^{tree}`.

Evaluator fix remains unchanged. Emitter no longer tries to recognize NaN from `BoundLiteralExpression` shape. For floating constant patterns routed through `object` or value-type nullable subjects, it now checks runtime boxed type, unboxes it, and emits `ceq`; all expression shapes therefore retain IEEE NaN behavior.

### Decisive regression probe

Four patterns: `0.0 / 0.0`, `Double.NaN * 1.0`, `-Double.NaN` through `object`, plus `0.0 / 0.0` through `float64?`.

| tree | bare `gsc` | `gsc /out:` | `gsi` |
|---|---|---|---|
| current main `faca348a` | `11 11 11 11` | `33 33 33 33` | `11 11 11 11` |
| fixed `efaa01ed` | `33 33 33 33` | `33 33 33 33` | `33 33 33 33` |

Prior emitted `11 11 11 11` regression is gone. Release-driver run used three asserted-empty directories; Core/Compiler/Repl epoch mtimes were `1785778014`, `1785778016`, `1785778019`.

### Product mutations

Each mutation was asserted present exactly once, reached test phase, failed named test, then was restored. Restored targeted corpus: 3/3 green.

| hunk mutant | killed by | mutant result |
|---|---|---|
| evaluator `NumericEquals` → `object.Equals` | `NonLiteralNaNConstants_AgreeAcrossDrivers` | RED: bare/gsi `11 11 11 11`; emit stayed `33 33 33 33` |
| invert object-route ownership | `TwentyRowCorpus_AgreesAcrossDrivers` | RED: emitted boxed int32/bool fell to `33` |
| invert nullable-route ownership | `TwentyRowCorpus_AgreesAcrossDrivers` | RED: emitted int64?/float64?/nil float64? fell to `33` |
| invert floating runtime-value route | `NonLiteralNaNConstants_AgreeAcrossDrivers` | RED: emit regressed to `11 11 11 11`; bare/gsi stayed `33 33 33 33` |

Current-main baseline independently reproduces issue-side failures: boxed rows 12/13 and nullable rows 18–20 emit `33`; nonliteral NaN remains evaluator/gsi `11` while emit is correctly `33`.

### Validation

Release solution build: `BUILD_RC=0`, 0 warnings, 0 errors.

All nine unfiltered solution test projects green:

- InternalAnalyzers.Tests: 7
- GSharp.GeneratorHost.Tests: 31
- Core.Tests: 7,116 passed, 1 skipped
- Compiler.Tests: 4,232
- Interpreter.Tests: 1,217
- LanguageServer.Tests: 462
- Extensions.Tests: 104
- Sdk.Tests: 59
- Cs2Gs.Tests: 1,903

DavidObando:
## **MERGEABLE** — both issues genuinely fixed, all four hunks mutation-covered, and the single-unwrap question resolved in the fix's favour by an invariant rather than by sampling.

Anchors: `origin/main` **`61fe9cb9`** (fetched from the real remote — the briefed SHA was correct this time), PR head **`efaa01ed`**, merge base **`faca348a`**, `merge-tree --write-tree` **output** one line — tree `920fb1a9`, **0 conflict lines**. Built merged commit `723b734a` and base `61fe9cb9` separately, both `BUILD_RC=0`, 0 warnings. Freshness by epoch mtime `sort -n`, first-party `GSharp*.dll` only, `Core`/`Compiler`/`Repl`, excluding `ref/`: 29 binaries per tree, all fresh (the apparent 7.9-year-old outlier was `xunit.abstractions.dll`, a 2018 NuGet copy).

### Verbatim repros first

#3110's repro byte-for-byte, and #3113's rows 12–17, on both trees, three drivers:

| probe | driver | base `61fe9cb9` | merged |
|---|---|---|---|
| #3110 verbatim (`var x = Double.NaN`) | eval / emit / gsi | `MATCHED` / `DEFAULT` / `MATCHED` | **`DEFAULT` / `DEFAULT` / `DEFAULT`** |
| #3113 rows 12–17 | emit | `33 43 41 73 83 93` | **`11 21 41 51 61 71`** |

Base reproduces both defects exactly as reported, including #3113's row-14 control (boxed `string` = `41`, the one boxed case that already worked). Merged is correct on every cell of both.

### The open question: does NaN reach `object.Equals` through more than one conversion?

**No, and it cannot — there is an enforceable invariant, not just passing fixtures.**

`src/Core/CodeAnalysis/Binding/PatternBinder.cs:189-209` is the sole non-error construction path for `BoundConstantPattern.Value` (`:223`; the only other sites are `:197` for errors and `BoundTreeRewriter.cs:807` which rewrites in place and preserves depth):

- `:189-198` — `Conversion.Classify(expression.Type, discriminantType)`; if the conversion does not exist **or is explicit**, it is rejected with `GS0171` and the operand becomes `BoundErrorExpression`. This is the mechanism behind #3113's note that `case float32(1.5)` is not a legal constant pattern.
- `:209` — `var value = conversion.IsIdentity ? expression : conversions.BindConversion(..., expression, discriminantType);`

So the operand is **either the bare expression, or exactly one conversion node whose `.Type` is `discriminantType`** — which is precisely `valueType` at the emitter. The single unwrap at `MethodBodyEmitter.Patterns.cs:237` (`conversion.Type == valueType`) is therefore exactly matched to the single conversion the binder can insert. The surface did not merely shrink.

Probed anyway, all on merged, each in its own asserted-empty directory, `rc` captured before any pipe:

| shape | expected | eval | emit | gsi |
|---|---|---|---|---|
| NaN via `object` (boxed) | no match | ✓ | ✓ | ✓ |
| NaN via `float64?` | no match | ✓ | ✓ | ✓ |
| `float32` NaN via `object` | no match | ✓ | ✓ | ✓ |
| `float32` NaN via `float32?` | no match | ✓ | ✓ | ✓ |
| **constant-folded** `0.0/0.0` via `object` | no match | ✓ | ✓ | ✓ |
| **chained** `float32`→`float64`→`float64?` | no match | ✓ | ✓ | ✓ |
| **erased slot** (`func F(v object)`) | no match | ✓ | ✓ | ✓ |
| **imported** `object` from a `/r:` C# library | no match | n/a | ✓ | n/a |
| `-0.0` vs `0.0` via `object` and via `float64?` | match | ✓ | ✓ | ✓ |
| `±Infinity` via `object` | match | ✓ | ✓ | ✓ |

**Generic / type-parameter slot is closed statically**, not by luck: `switch v { case Double.NaN ... }` inside `func G[T](v T)` is rejected `GS0171` — *identically on base and merged*, so the PR neither opened nor relied on it.

I specifically tried to falsify the unwrap via symbol identity: `:232` uses value equality (`valueType?.ClrType.IsSameAs(typeof(object))`) while `:237` three lines later uses `==`, and neither `Symbol` nor `TypeSymbol` declares `operator ==`, `Equals`, or `IEquatable` — so `:237` is **reference** equality. I drove an `object` whose symbol comes from imported metadata (a C# library returning `object`, referenced with `/r:`); emit still produced the correct no-match, so the symbols are canonicalised in practice. **Hypothesis raised and disproved — reporting it because the asymmetry is still a latent hazard, not because it currently misbehaves.**

### Signed zero — the brief's premise was wrong, and I checked rather than assumed

The brief stated `Equals(0.0, -0.0)` is **false**, "the other IEEE identity where `==` and `Equals` disagree, in the opposite direction". Measured directly on .NET 10:

```
object.Equals(NaN, NaN)   = True     (IEEE == gives False)   <- the bug
object.Equals(0.0, -0.0)  = True     (IEEE == gives True)    <- agrees
```

`Double.Equals` special-cases **only** NaN; it does not do bitwise comparison. `Equals` and `==` therefore disagree **only** on NaN. This corroborates #3110's own measured table (signed zero and both infinities correct on all three drivers on `main`) and its sentence "`object.Equals` differs from IEEE `==` **only** for NaN". So there is no second half of the hole: signed zero is a **regression** risk, not a missing fix — and it is preserved, because `Evaluator.Concurrency.cs:458` `NumericEquals` dispatches `(float, float)`/`(double, double)` to real IEEE `==`.

### Two issues, one PR: coupled, not stapled — and I can prove it

The evaluator half is **one line** (`Evaluator.Concurrency.cs:126`, `object.Equals` → `NumericEquals`) reusing a primitive that already existed at `:458` with four call sites, two of which are the `==`/`!=` operators. That is the correct root-cause shape: pattern equality now uses the same primitive `==` already used, rather than a new NaN special case. No parallel predicate was introduced.

Mutant **M3** settles whether the emit half belongs in the same PR. Substituting the float guard at `:240` so it can never fire (`Float32/Float64` → `Decimal/String`, both of which return earlier at `:223`/`:211`) makes the emit driver produce `101 111 121 131` on my hazard corpus — **all four NaN cells wrongly match**. In other words, #3113's fix (routing boxed/nullable operands through `object.Equals`) is exactly what *creates* an emit-side NaN hole that `main` never had, because on `main` boxed operands never matched at all. Landing #3113 without the float guard would have regressed #3110 on the emit driver. **They had to ship together.**

### Mutation — four PRODUCT mutants, one per hunk, each asserted applied

Baseline merged: 3/3 green. Every mutant verified applied by `git diff` + line echo, and restored by exact md5 round-trip (`Evaluator.Concurrency.cs` `1fef1410…`, `MethodBodyEmitter.Patterns.cs` `4ca3a6bf…`, both re-verified `YES` at the end).

| mutant | site | change | build | result |
|---|---|---|---|---|
| **M1** | `Evaluator.Concurrency.cs:126` | `NumericEquals` → `object.Equals` (exact revert) | rc=0 | **RED** 2 failed |
| **M2** | `Patterns.cs:237` | `conversion.Type == valueType` → `!=` (inversion) | rc=0 | **RED** 2 failed |
| **M3** | `Patterns.cs:240` | float guard → `Decimal`/`String` (never fires) | rc=0 | **RED** 2 failed |
| **M4** | `Patterns.cs:235` | `isObject \|\| isValueNullable` → `isValueNullable` | rc=0 | **RED** 2 failed |

All four built cleanly, so none is a no-result. **Vacuity control:** the PR's own test file compiled into the **base** tree fails **3/3**, so the tests detect the real defects rather than asserting current behaviour.

### The tests themselves

Three `[Fact]`s, 365 lines, and each one compares **all three drivers** against a single expected string (`RunBareGsc`, `RunEmittedAsync`, `RunGsi`) — a genuine driver matrix, not two columns. Test 2 mixes `33` and `11`, so it carries negative rows. And #3113's explicit acceptance item about the widths it could not express is **met, not dodged**: `int8?`, `int16?`, `uint32?`, `uint64?` are all present at `:312-321` via `int8(1)`-style casts that sidestep the `GS0156` initialiser rejection.

### Findings

- **SHOULD-FIX** — `src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Patterns.cs:237`. The single unwrap is correct, but the invariant that makes it correct lives in another file and is not referenced. As written the line reads as a shape guess. If `PatternBinder.cs:209` ever inserted an intermediate conversion, `patternValue.Type` would stop being a float type, the `:240` guard would silently stop firing, and #3110 would reopen **on the emit driver only** — with no test failing unless someone re-derives M3. Please add a comment at `:237` naming `PatternBinder.cs:189-209` and the two properties relied on: explicit conversions are rejected, and at most one conversion is inserted, always to `discriminantType`. This is cheap and it is the difference between "sound by enumeration" and "sound by argument".
- **NIT** — same file, `:232` vs `:237`: value equality (`ClrType.IsSameAs`) and reference equality (`==` on `TypeSymbol`, which defines no equality members) applied to the same symbol three lines apart. It works today because symbols are canonicalised — I confirmed that with an imported-metadata `object` — but `:232` is written defensively precisely because that canonicalisation is not guaranteed, and `:237` then assumes it.
- **NIT, pre-existing and explicitly not this PR's** — `string?` with a **non-interned** string mismatches on emit: `String.Concat(t, "c")` into `var s string?` against `case "abc"` prints the `default` arm on `gsc /out:` while bare `gsc` and `gsi` both match. **Identical on base `61fe9cb9` and on merged**, so this PR neither caused nor worsened it. Mechanism: `string?` is a `NullableTypeSymbol` over a reference type, so it misses `:211` (`valueType == TypeSymbol.String`), misses `isObject`, and misses `isValueNullable` (value types only) — falling to the primitive `ceq` at `:284-288`, i.e. reference comparison. Plain `string` is unaffected. This is a third instance of #3113's exact defect family — emit silently picks the wrong arm at rc0 with no diagnostic — sitting in the function this PR is fixing. Worth a follow-up issue; **not** a reason to hold this PR, and deliberately not something I would ask to be scoped in.
- **NIT** — `Issue3110And3113ConstantPatternEqualityTests.cs:20-42`: the 20-row corpus expects twenty identical `11`s. Match/no-match is discriminated (`33` is the default arm), but arm identity is not encoded, so a duplicated, reordered or silently-dropped row is invisible except by line position. Distinct per-row values would make a failure name itself.

### Denominator and its source

**9 of 9 test projects, from `dotnet sln GSharp.sln list`** — CI's own denominator per `build/generate-ci-test-matrix.py:26`. Full sweep, **unfiltered**, on the merged tree:

| project | result |
|---|---|
| Compiler.Tests | 4232 passed |
| Core.Tests | 7116 passed, 1 skipped (7117) |
| Interpreter.Tests | 1223 passed |
| Sdk.Tests | 59 passed |
| Cs2Gs.Tests | 1903 passed |
| **total** | **15,137 passed, 1 skipped, 0 failed** |

`Sdk.Tests` **was run** this round. Note that a green `Cs2Gs.Tests` here proves the translator behaves on its 1,903 self-contained fixtures and proves nothing about the pinned `tools/cs2gs/external/oahu.json` gate.

Both worktrees end `git status --porcelain` empty, both mutated files md5-identical to their originals, and I did not write to the shared checkout (still clean at `61fe9cb9`).

PR #3132, opened 2026-08-03
Address composite literal review cleanup

DavidObando:
Post-merge cleanup from the review of #3123:

- remove the StateMachineEmitter formatting-only hunk
- document when FieldDeclaringType is null accurately
- explain why imported positional records retain inherited-field resolution

No behavioral changes.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/BoundFieldInitializer.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundFieldInitializer.cs
@@ -42,7 +42,7 @@ public sealed record BoundFieldInitializer
     /// <summary>Gets the targeted field, or <see langword="null"/> when this initializer targets a property.</summary>
     public FieldSymbol Field { get; }
 
-    /// <summary>Gets the type that declares <see cref="Field"/>, or <see langword="null"/> when the initializer targets a property.</summary>
+    /// <summary>Gets the inherited type that declares <see cref="Field"/>, or <see langword="null"/> for a direct field or property.</summary>
     public StructSymbol FieldDeclaringType { get; }
 
     /// <summary>Gets the targeted property, or <see langword="null"/> when this initializer targets a field.</summary>
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs
@@ -1343,7 +1343,9 @@ internal sealed partial class MethodBodyEmitter
                 continue;
             }
 
-            // Keep inherited-field resolution aligned pending imported generic-record support.
+            // This branch executes but resolves to the same token as the else-arm for every
+            // expressible program. The !isGeneric guard above excludes imported and user-authored
+            // generics, so both that guard and generic-record support must change before this discriminates.
             var fieldHandle = init.FieldDeclaringType != null
                 ? this.ResolveStructLiteralFieldToken(literal, init)
                 : this.outer.userTokens.ResolveFieldToken(literal.StructType, init.Field);
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs
@@ -1343,6 +1343,7 @@ internal sealed partial class MethodBodyEmitter
                 continue;
             }
 
+            // Keep inherited-field resolution aligned pending imported generic-record support.
             var fieldHandle = init.FieldDeclaringType != null
                 ? this.ResolveStructLiteralFieldToken(literal, init)
                 : this.outer.userTokens.ResolveFieldToken(literal.StructType, init.Field);
--- a/src/Core/CodeAnalysis/Emit/StateMachineEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/StateMachineEmitter.cs
@@ -1355,8 +1355,7 @@ internal sealed class StateMachineEmitter
         // Builder: AsyncIteratorMethodBuilder.Create()
         var createMethod = typeof(System.Runtime.CompilerServices.AsyncIteratorMethodBuilder)
             .GetMethod("Create", System.Reflection.BindingFlags.Public | System.Reflection.BindingFlags.Static, null, Type.EmptyTypes, null);
-        initializers.Add(new BoundFieldInitializer(
-            builderField,
+        initializers.Add(new BoundFieldInitializer(builderField,
             new BoundClrStaticCallExpression(null, createMethod, TypeSymbol.FromClrType(typeof(System.Runtime.CompilerServices.AsyncIteratorMethodBuilder)), ImmutableArray<BoundExpression>.Empty)));
 
         foreach (var parameter in parameters)


DavidObando:
## Rubber-duck verdict: **MERGEABLE** — "No behavioral changes" **holds, and it is measured, not assumed**. One SHOULD-FIX on item 3's wording.

I was asked to *falsify* the no-behavioural-change claim, not confirm it. I could not falsify it. Below is what it took to make that a real result rather than a green rubber stamp.

Numbers true of `main` **`27a05798`** (re-checked at write time — main had **not** moved), head **`026d9d07`**, merge base **`f7d9a5d3`**; head is **5 behind / 1 ahead**, so head ≠ merged tree and every probe below ran on the **co-merged** tree. `git merge-tree --write-tree` output `b18ca291f784c2a301fe3d3ebe01c00f2526de26` == the actual merged tree, **0 conflict lines** in its *output* (not just rc). **`BUILD_RC=0`, 0 warnings / 0 errors.** Freshness by **mtime** on all three `GSharp.Core.dll` copies — `Core/`, `Compiler/` **and `Repl/`** — merged `00:28:32` md5 `24d51c9c`, base `00:13:57` md5 `140692f7`.

---

## Item 1 — inert, **empirically**. But the obvious corpus would have proven nothing.

**The trap first.** The changed line lives in `CreateAsyncIteratorKickoffLiteral` (`src/Core/CodeAnalysis/Emit/StateMachineEmitter.cs:1343`), the async-iterator builder init. **0 of 122 `samples/*.gs` combine `async` and `yield`** — positive controls: 10 use `async`, 2 use `yield`, neither set intersects. So a `samples/`-only byte-diff **never reaches this hunk**; a zero there would have been *scope, not absence*. I wrote a purpose-built async-iterator probe (`async func vals() async sequence[int32]` with `yield`/`await Task.Delay`).

Evidence, in increasing strength:

1. **A priori** — `:1355-1358` base vs `:1358-1359` head is a re-wrap of the argument list only; **token stream identical**, same constructor, same argument order, argument list unchanged so no overload can shift.
2. **Emit oracle** — probe compiled by base `gsc /out:` and merged `gsc /out:`, each into **its own asserted-empty directory**: **byte-identical**, md5 `95273fb7` both. Positive control that the hunk was reached: emitted IL references `AsyncIteratorMethodBuilder` (1 occurrence in each).
3. **Load-and-run** (`IlVerifier.Verify` is not sufficient) — both emitted assemblies execute: `11 / 22 / 33`, rc=0. Distinct values, never all-zero.
4. **Oracle sensitivity proven** — *this is the part that makes point 2 mean anything.* **M-SM2**, product mutant on the exact touched lines `:1358-1359`: duplicated `initializers.Add(new BoundFieldInitializer(builderField, …))`. Build rc=0, mutant asserted applied. Emitted bytes moved `95273fb7` → **`dd1116f2`**. **ORACLE LIVE.** Byte-identity in point 2 is therefore a real signal, not a constant-true.
5. **Full three-column differential, 122 samples** — `gsc /out:` **BYTEDIFF=0, RCDIFF=0**; `gsi` **DIFF=0**; bare `gsc` (evaluates) **DIFF=0**. The async-iterator probe is also identical across all three drivers.

**Disclosure of a mutant that failed, because repeating it would mislead.** My *first* mutant (M-SM1) replaced `TypeSymbol.FromClrType(typeof(…AsyncIteratorMethodBuilder))` at `:1359` with `AsyncTaskMethodBuilder`. It **survived** the byte-compare. That is not evidence the oracle is dead — that argument is a *static* call's receiver-type symbol and is simply never emitted. It was a **no-result, and a no-result is not a survival**; M-SM2 was needed. (Also worth knowing: an assertion caught that my intended edit matched **2** sites, not 1, so I retargeted by line number.)

Suites on the merged tree: `Core.Tests` **unfiltered** — 7089 passed / 0 failed / 1 skipped. `Compiler.Tests` **filtered** to `AsyncIterator|StateMachine` — 80/80. I am labelling the second as filtered on purpose; **a filtered green is not a green**, and it is the differential in points 2–5, not the suite, that carries item 1.

## Item 2 — **VERIFIED TRUE**, and the *old* doc was wrong

New text at `src/Core/CodeAnalysis/Binding/BoundFieldInitializer.cs:45`: "the **inherited** type that declares `Field`, or null for a **direct field** or property."

Counted **wide** (partial class — a filtered grep undercounts): **29** total `new BoundFieldInitializer(`, **20** on the field overload, minus **2** propagation copies (`BoundTreeRewriter.cs:1489`, `SpillSequenceSpiller.cs:2029`) = **18** real construction sites, matching #3123's "1 of 18". Confirmed `ConversionClassifier.cs:3074` (multi-line) passes only 2 args.

The **sole originator** of a non-null value is `ExpressionBinder.Literals.cs:1349-1352`:
`ReferenceEquals(fieldDeclaringType, structSymbol) ? null : fieldDeclaringType`, where `fieldDeclaringType` comes from `TypeMemberModel.TryGetFieldIncludingInherited` (`TypeMemberModel.cs:246`), whose walk is `for (var c = structSymbol; c != null; c = c.BaseClass)`. That bounds the value to `structSymbol` or a **strict ancestor** ⇒ non-null ⟺ inherited; null for a direct field; the property overload never sets it. The new doc is exactly right. The **previous** doc ("or null when the initializer targets a property") was **wrong** — it omitted the common direct-field case. Item 2 is a correctness fix to a comment, not a rephrasing.

## Item 3 — **SHOULD-FIX**: answers an adjacent, weaker question — and the operative word is wrong in *both* directions

`src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs:1346`, sitting directly on the ternary inside `EmitImportedPositionalRecordLiteral`:

> `// Keep inherited-field resolution aligned pending imported generic-record support.`

#3123's NIT 2 asked, verbatim:

> "a one-line comment saying it is **unreachable** pending imported-generic-record support would stop a future reader mutating it and concluding the code is dead."

The shipped comment keeps the *"pending imported generic-record support"* clause and **drops the operative fact**. It never says the branch is currently unpinnable. The precise failure mode NIT 2 named — a future reader mutates the line, sees a green suite, concludes the code is dead — is **not prevented** by this wording. That is answering the easier adjacent question.

**And measured, the right word is neither NIT 2's "unreachable" nor the PR's "aligned".** I built a real C# library (`RecLib.dll`) with generic and non-generic positional records and consumed it from G# via `gsc /r:`:

- Non-generic `RDerived{ x: 11, w: 22 }` → emits, runs, prints `11 / 22`. **The `FieldDeclaringType != null` arm genuinely executes.** So "unreachable" is factually wrong.
- **M-EMIT** — condition inversion on `MethodBodyEmitter.Expressions.cs:1347` (`init.FieldDeclaringType != null` → `false`; **not** `if (true)`), build rc=0, mutant asserted applied: emitted assembly **byte-identical** (`fcbcf33a` both) and still prints `11 / 22`. The two arms are **indistinguishable**, not unreachable.
- The discriminating case is genuinely unexpressible: `GDerived[string]{ gx: "aa", gw: "bb" }` → **`GS0148`** at `gen.gs(4,17)` ("Generic function 'GDerived' requires 0 type arguments but was given 1") and **`GS0158`** at `gen.gs(5,21)` ("Cannot find member gx").

**Recommended wording**: say the branch *executes* but currently resolves to the **same token** as the else-arm for every expressible program (non-generic imported types), and only discriminates once imported generic records are supported. That is the fact that stops the future mutation — and it is accurate where both the original review's phrasing and this PR's are not.

This is a SHOULD-FIX, not a blocker: it is a comment, it cannot break a build, and the PR is still a net improvement over no comment.

## Nothing the #3123 review raised was silently dropped

NIT 1 → item 1 (done). NIT 2 → item 3 (partially; see above). NIT 3 (`mergeStateStatus UNSTABLE`) explicitly required no action. No review item was dropped.

## Hygiene

All mutation in `/tmp/g3132` worktrees I created. Every file restored with an exact **md5 round-trip** — `StateMachineEmitter.cs` back to `966ab382`, `MethodBodyEmitter.Expressions.cs` restored, merged `GSharp.Core.dll` back to **`24d51c9c`** — and `git status --porcelain` **empty** in the base tree, the merged tree, and the main checkout. Every probe compiled into its **own asserted-empty directory**. Distinct values `11`/`22`/`33`, never all-zero. No worktree I do not own was written. No e2e; `~/.nuget/packages/gsharp.net.sdk/` never touched. No `git stash`, no `--no-build`, no `GITHUB_ACTIONS=true`.

**Disclosure:** the first merged build was started at load **38.75**, above the threshold, and **segfaulted (rc=139)**. That was an environment no-result, not a PR finding; I discarded it entirely and rebuilt at load **27.64** (rc=0, 0 warnings). Every subsequent build was started below 30 (23.34, 14.16, 23.18, 17.32).

---

### Verdict

**MERGEABLE.**

- **BLOCKER:** none.
- **SHOULD-FIX** — `src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs:1346`: comment answers a weaker adjacent question than #3123 NIT 2 asked and, as measured above, uses an inaccurate characterisation. Suggested replacement in item 3.
- **NIT** — the PR body's "No behavioral changes" is true but was, until this review, *unverified* for the one product hunk; the `samples/` corpus cannot reach it. Worth a line in the body pointing at the async-iterator differential so the next reader does not re-derive it.

DavidObando:
IL falsification check for the `StateMachineEmitter` cleanup:

- Built PR head `026d9d07`, then reintroduced exactly the removed formatting (`new BoundFieldInitializer(` / `builderField,`) as a working-tree mutant.
- Both full-solution builds returned `BUILD_RC=0`; the mutant rebuilt `out/bin/Release/Core/GSharp.Core.dll` at 02:12:17.
- Dumped IL for the entire `GSharp.Core.CodeAnalysis.Emit.StateMachineEmitter` type with `ilspycmd -il` (no sequence points).
- PR, expanded-format mutant, and restored round-trip dumps are byte-identical; `cmp` rc 0 for both comparisons.
- The affected method is `CreateAsyncIteratorKickoffLiteral`; its emitted IL is therefore unchanged. Source restored; tracked status clean.

DavidObando:
Comment-truth check (not reconstruction):

- `FieldDeclaringType`: the field constructor defaults it to `null` (`BoundFieldInitializer.cs:19-25`); composite binding passes non-null only when `fieldDeclaringType` is not the literal type (`ExpressionBinder.Literals.cs:1349-1352`), i.e. inherited. The property constructor never assigns it. The other construction sites use direct/synthesized fields, so `null` is accurate there too.
- Imported positional records: dispatch is explicitly guarded by `!isGeneric` (`MethodBodyEmitter.Expressions.cs:1132-1137`). Thus the path is currently reachable only for non-generic imported records; the review probe also measured generic imported records as unbindable (`GS0148`/`GS0158`). Retaining inherited-container resolution keeps this third composite-literal emitter aligned for the generic case when it becomes expressible.

DavidObando:
## Addendum — re-measured against a moved `main`. **Verdict unchanged: MERGEABLE.**

Applying the three fleet corrections to my own posted verdict above. Two of them bit; one did not. Details, because a correction with no numbers is not a correction.

### Correction 1 — test-project denominator

My verdict above **cited no denominator at all** — I reported `Core.Tests` unfiltered and `Compiler.Tests` filtered and stopped. That is not a wrong number, but it *implied* coverage I had not measured, which is the same sin in a quieter register. Enumerated properly from **`GSharp.sln`**, not `ls test/`:

Worth noting for the fleet: in this tree **`ls test/` returns 8**, not 7 — `test/Shared/` is a source directory with **no `.csproj`** (it holds `Issue3081CompositeLiteralCases.cs`). So `ls test/` is a *third* wrong number, and it happens to be the bulletin's corrected one. Only `GSharp.sln` is the denominator.

| # | project | result on merged tree |
|---|---|---|
| 1 | `test/Core.Tests` | **7106** pass / 0 fail / 1 skip |
| 2 | `test/Compiler.Tests` | **4221** pass / 0 fail — **unfiltered this time** |
| 3 | `test/Extensions.Tests` | 104 / 0 |
| 4 | `test/InternalAnalyzers.Tests` | 7 / 0 |
| 5 | `test/Interpreter.Tests` | 1201 / 0 |
| 6 | `test/LanguageServer.Tests` | 462 / 0 |
| 7 | `test/Sdk.Tests` | **NOT RUN** — local SDK package, out of bounds. Stating it plainly rather than implying coverage. |
| 8 | `tools/cs2gs/Cs2Gs.Tests` | **1903** / 0 — the out-of-process gate a pure `src/` change can redden |
| 9 | `tools/gsgen/GSharp.GeneratorHost.Tests` | 31 / 0 |

**8 of 9 projects, 15,035 passed, 0 failures, 1 skip.**

**And my filter was worse than I knew.** `Issue3081CompositeLiteralCases.cs` lives in `test/Shared/` and is `<Compile Include>`-linked into **both `Compiler.Tests` and `Interpreter.Tests`** — i.e. the composite-literal axis is deliberately covered on the emit *and* interpret columns. My original filter `AsyncIterator|StateMachine` **excluded every one of those cases** — the exact axis items 2 and 3 touch. That filtered green was genuinely uninformative for two of the three items. The unfiltered 4221 above is the fix.

### Correction 2 — measuring a stale tree

This one bit hardest, and not in the direction advertised. My verdict was true of `main` `27a05798`, which I stated. **`main` was `5b71a74d` when I re-measured, is `2dea578b` as I write this, and was never `787166a9` by the time I read the bulletin** — the bulletin's own SHA was stale on arrival (it is an ancestor of current `main`). So I rebuilt everything from scratch rather than trusting any existing tree.

Fresh worktrees at `5b71a74d`, **rebuilt after checkout, no exceptions**. `git merge-tree --write-tree` output `31d3213f296946e104ad7a8e7b5cc29bcc5f5379` == the **actual** merged tree, **0 conflict lines in the output**. Both builds `BUILD_RC=0`, 0 warnings / 0 errors. Diff scope re-measured three-dot from computed merge base `f7d9a5d3` with `--full-index`: **3 files, +3 / −3**, unchanged.

Re-verified the *source* claims on the tree I was measuring rather than carrying them forward. All identical at current `main`: originator ternary `ExpressionBinder.Literals.cs:1352`; **29** total `new BoundFieldInitializer(` counted wide; doc line `BoundFieldInitializer.cs:45`; item-3 comment `MethodBodyEmitter.Expressions.cs:1346` with the ternary at `:1347-1349`.

Re-ran the whole differential at current `main`:
- async-iterator probe, base vs merged, own asserted-empty dirs → **byte-identical**, md5 `95273fb7` — *the same md5 as at `27a05798`*, so 103 PRs of drift moved this emitted assembly not one byte. Positive control: `AsyncIteratorMethodBuilder` present in both. Load-and-run `11 / 22 / 33`, rc=0.
- 122 samples: `gsc /out:` **BYTEDIFF=0, RCDIFF=0**; `gsi` **DIFF=0**; bare `gsc` **DIFF=0**.

Pinned Oahu corpus `tools/cs2gs/external/oahu.json` is **not in the diff** — proved from the pin, not from an in-repo file list.

### Correction 3 — mtime aggregation

Did **not** bite: I never sorted mtimes, I compared three same-hour stamps by eye, so no conclusion rested on a lexicographic order. But my stamps *were* unsortable `%H:%M:%S` clock strings and my session ran across midnight, so I was one build away from exactly that bug. Re-measured in **epoch seconds with `sort -n`**:

```
1785753653  base2/{Core,Compiler,Repl}    2026-08-03 03:40:53
1785754226  merged2/{Core,Compiler,Repl}  2026-08-03 03:50:26
```

All three copies per tree at a single mtime, `Repl/` included, `ref/` excluded, both strictly after their checkouts.

### What does not change

Items 1 and 2 stand exactly as posted. Item 3 stands as a **SHOULD-FIX**: the comment at `MethodBodyEmitter.Expressions.cs:1346` still answers a weaker question than #3123's NIT 2 asked, and the accurate characterisation is still *"executes but resolves to the same token as the else-arm for every expressible program"* — neither "unreachable" nor "aligned". The M-EMIT condition inversion and the `GS0148`/`GS0158` measurements behind that were re-confirmed present on the current tree.

**MERGEABLE.** No blocker.

### Hygiene

`git status --porcelain` **empty** in all four scratch worktrees and the main checkout. All mutation restored by exact md5 round-trip. Every probe in its own asserted-empty directory. Distinct values `11`/`22`/`33`. No worktree I do not own was written. No e2e; `~/.nuget/packages/gsharp.net.sdk/` never touched. PR head still `026d9d07`, unchanged from what I reviewed.

**Disclosure:** one build in this round — `base2` — was started when the 1-minute load read **38.56**, above the threshold. I did not notice until the reading came back with the result; it completed `rc=0` with 0 warnings and its output is deterministic, but the rule says wait and I did not. Before the next build I read 40.36, waited five minutes, and built at 23.47. Reporting it rather than letting a clean number imply a clean process.

DavidObando:
## Status: verdict already posted (twice). **MERGEABLE**, unchanged after verifying the author's new evidence.

Full verdict `issuecomment-5163945427`; re-measure against moved `main` `issuecomment-5166613813`. This comment only covers the author's two new comments, which landed **between** mine.

### The author's IL proof — design sound, **one control missing**, and I already supplied it

`issuecomment-5164476609` compares PR head against itself with the formatting re-expanded. That is the **right** A/B: it isolates the whitespace hunk from the two comment hunks, which is cleaner than base-vs-merged. Tree pair is fine — the concern that head ≠ merged tree does not apply to a self-contained A/B.

**But three identical dumps do not prove inertness on their own.** "PR == mutant == restored, `cmp` rc0" is equally consistent with the dump being insensitive, empty, or stale. The mutant here is *whitespace*, so "whitespace mutant produced identical IL" is the claim being proven, not an independent control — it cannot fail. **A no-result is not a survival.** Nothing in that comment shows the `ilspycmd` oracle would have moved had something real changed.

I had already hit exactly this: my **first** mutant (swapping the `FromClrType` type argument at `StateMachineEmitter.cs:1359`) **survived** a byte-compare — that term is a static call's receiver-type symbol and is never emitted. I needed a second, semantic mutant to prove the oracle live: duplicating the `initializers.Add(new BoundFieldInitializer(builderField, …))` on the touched lines moved emitted bytes `95273fb7` → `dd1116f2`.

So: **author's proof + my sensitivity control together are conclusive.** Neither is sufficient alone. Note also we measured *different* oracles that agree — they hashed the compiler's own IL, I byte-compared **G# assemblies emitted by `gsc`** across 122 samples × 3 drivers (BYTEDIFF=0, RCDIFF=0, gsi 0, bare gsc 0). Item 1 is settled.

### Comment-truth check — one claim verified, one **inaccurate**

**Item 2 — TRUE.** `issuecomment-5164516389`'s anchors are correct and match my independent trace. I verified the stronger structural fact they assert informally: the value originates only at `ExpressionBinder.Literals.cs:1352`, and `TryGetFieldIncludingInherited`'s walk `for (var c = structSymbol; c != null; c = c.BaseClass)` bounds it to `structSymbol` or a strict ancestor ⇒ non-null ⟺ inherited, null for a direct field, never set by the property overload. Exactly the described circumstances **and no others**. The *old* doc was wrong.

**Item 3 — the anchor is real, and it undercuts the comment.** I checked `MethodBodyEmitter.Expressions.cs:1132-1137` and it is load-bearing: `if (!isGeneric && …)` guards the **sole** dispatch to `EmitImportedPositionalRecordLiteral` at `:1137`. I also checked the trap in the name — `ReflectionMetadataEmitter.IsUserGenericTypeReference` (`:4617-4631`) despite "User" tests **genericity only** (`TypeArguments` non-empty, or definition `TypeParameters` non-empty), not authorship, so imported generics are excluded too. The author's mechanism holds.

**But that is precisely why the in-code comment is wrong.** `MethodBodyEmitter.Expressions.cs:1346` reads:

> `// Keep inherited-field resolution aligned pending imported generic-record support.`

Since `:1132` guards this method off for **all** generic types, making imported generic records bindable would **not, by itself, make this branch discriminate** — such a literal would take the `isGeneric` path at `:1141+` and never enter this method. The comment's forward-looking justification is therefore not the real mechanism: you would *also* have to change the dispatch guard at `:1132`. A future reader trusting this comment would look for the payoff in the wrong place.

That sharpens my original SHOULD-FIX rather than replacing it. Recommended: say the branch **executes but resolves to the same token as the else-arm for every currently expressible program** (confirmed: forcing the else-arm emits a **byte-identical** assembly, `fcbcf33a`, still printing `11 / 22` for `RDerived{ x: 11, w: 22 }` off a real `/r:`-referenced C# library), and point at the **`!isGeneric` guard at `:1132`** as the reason, not at generic-record support alone.

### Verdict

**MERGEABLE.**

- **BLOCKER:** none.
- **SHOULD-FIX** — `src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs:1346`: comment states a forward-looking reason that the `!isGeneric` dispatch guard at `:1132` contradicts, and still omits that the site is currently indistinguishable. Wording above.
- **NIT** — `issuecomment-5164476609`: the posted SHA-256 `619ca867…2290f1` is **63 hex characters**, not 64, so it cannot be verified as transcribed (the brief quoting it renders the tail differently again — `290f1c` vs `2290f1`). A hash nobody can check is decoration; worth reposting correctly.

### Coverage and freshness

**8 of 9 projects** enumerated from `GSharp.sln`, run on the merged tree: Core 7106, **Compiler 4221 unfiltered**, Extensions 104, InternalAnalyzers 7, Interpreter 1201, LanguageServer 462, **Cs2Gs 1903**, **GeneratorHost 31** — **15,035 passed, 0 failed, 1 skipped**. `Sdk.Tests` **NOT RUN** (local SDK package, out of bounds) — stated, not implied.

Measured at `main` **`5b71a74d`** with merged tree `31d3213f` (== `merge-tree` output, 0 conflict lines), both builds `BUILD_RC=0`. `main` has since moved to **`f6f8acc0`** — it moved twice while I wrote this. PR head still `026d9d07`. All scratch worktrees `git status --porcelain` empty; nothing written to any tree I do not own.

DavidObando:
## Two answers. Verdict: **MERGEABLE**.

(Fourth posting — prior detail in `issuecomment-5163945427`, `issuecomment-5166613813`, `issuecomment-5166909023`. Head verified `026d9d07`.)

### Q1 — is the author's IL proof sound? **Design yes; as posted, incomplete.**

**Tree pair: not the pair described in the question, and that is fine — it is better.** The proof is *not* PR head vs merge base. It is **PR head vs PR head with the removed formatting reintroduced** — an A/B on one tree. That isolates the whitespace hunk from the two comment hunks; a head-vs-base compare would have conflated all three. Correct choice.

**Mutant applied: NO evidence, and this is the gap.** All three dumps (PR / mutant / restored) are identical, so nothing shows the oracle would have moved had something real changed. The mutant is *whitespace*, so "whitespace mutant produced identical IL" is the claim being proven, not a control on it — it cannot fail. **A no-result is not a survival.**

This is not hypothetical. My **first** mutant here — swapping the `FromClrType` type argument at `StateMachineEmitter.cs:1359` — **survived** a byte-compare, because that term is a static call's receiver-type symbol and is never emitted. Believing it would have been wrong. A second, semantic mutant (duplicating the `initializers.Add(new BoundFieldInitializer(builderField, …))` on the touched lines) moved emitted bytes `95273fb7` → `dd1116f2`, proving the oracle live.

**Combined, the claim is settled.** Author's A/B + my sensitivity control are jointly conclusive; neither is alone. Two different oracles agreeing: theirs the compiler's own IL, mine the **assemblies `gsc` emits**, 122 samples × 3 drivers, zero differences. Item 1 is closed.

### Q2 — are the two comments true? **One yes, one no.**

**`FieldDeclaringType` doc (`BoundFieldInitializer.cs:45`) — TRUE.** Exactly those circumstances, no others. Sole originator `ExpressionBinder.Literals.cs:1352`; `TryGetFieldIncludingInherited` walks `for (var c = structSymbol; c != null; c = c.BaseClass)`, bounding a non-null value to `structSymbol` or a strict ancestor ⇒ non-null ⟺ inherited, null for a direct field, never set by the property overload. The **old** doc was wrong (it omitted the common direct-field case), so this is a real correction.

**Imported-positional-record explanation (`MethodBodyEmitter.Expressions.cs:1346`) — NOT the real reason.** The comment says:

> `// Keep inherited-field resolution aligned pending imported generic-record support.`

The mechanism the author cites off-PR is real: `:1132` `if (!isGeneric && …)` guards the **sole** dispatch to `EmitImportedPositionalRecordLiteral` at `:1137`, and `ReflectionMetadataEmitter.IsUserGenericTypeReference` (`:4617-4631`) tests **genericity only** despite the "User" in its name, so imported generics are excluded too.

**But that is exactly what falsifies the comment.** Because `:1132` guards this method off for *all* generic types, supporting imported generic records would **not** make this branch discriminate — such a literal takes the `isGeneric` path at `:1141+` and never enters this method. The dispatch guard at `:1132` would have to change as well. A reader trusting `:1346` looks for the payoff in the wrong place.

Measured: forcing the ternary to its else-arm emits a **byte-identical** assembly (`fcbcf33a`) and still prints `11 / 22` for `RDerived{ x: 11, w: 22 }` against a real `/r:`-referenced C# library. So the branch **executes** (not "unreachable", as the #3123 review asked it be called) but is **indistinguishable** (not merely "aligned").

### Itemized

- **BLOCKER:** none.
- **SHOULD-FIX** — `src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs:1346`. Suggested: *"Executes but resolves to the same token as the else-arm for every currently expressible program; only the `!isGeneric` dispatch guard at `:1132` keeps generic literals out of this method, so both that guard and generic-record support would have to change before this discriminates."*
- **NIT** — the posted SHA-256 `619ca867…2290f1` is **63 hex characters, not 64**, so it cannot be checked as transcribed (your quoting renders the tail differently again: `290f1c` vs `2290f1`). Worth reposting correctly.

### Scope note

You asked me to flag it if the diff touches more than comments: **it does.** Three files, +3/−3 — two comment hunks plus a real source edit in `StateMachineEmitter.cs` (`new BoundFieldInitializer(\n builderField,` re-wrapped onto one line) inside `CreateAsyncIteratorKickoffLiteral`. Token stream identical, so inert a priori, and confirmed inert by measurement above. Worth knowing: **0 of the 122 `samples/` programs combine `async` + `yield`**, so a samples-only byte-diff would have been vacuous — the async-iterator probe was purpose-built to reach the hunk.

DavidObando:
## Addendum — `Sdk.Tests` correction

I retract my earlier statements that `Sdk.Tests` was out of bounds. That conclusion came from misreading a doc comment as executed behavior.

Ran `test/Sdk.Tests/Sdk.Tests.csproj` unfiltered on PR head `dc623409`, Release, warnings-as-errors, with build enabled and an isolated `NUGET_PACKAGES`:

- **57 passed / 0 failed / 0 skipped** (`TEST_RC=0`)
- This updates the branch validation to **9 of 9 projects enumerated from `GSharp.sln`: 14,524 passed / 0 failed / 1 skipped**.

No local e2e was run; the shared `~/.nuget/packages/gsharp.net.sdk/` cache was not used.

Issue #3093, opened 2026-08-02
[gsc] sequence[T] iterator method does not satisfy IEnumerable[T] interface contract


DavidObando:
## Reproduction

```csharp
interface IDetector
{
    IEnumerable<int> Detect(string text);
}

sealed class Detector : IDetector
{
    public IEnumerable<int> Detect(string text)
    {
        yield return text.Length;
    }
}
```

cs2gs naturally emits the iterator return as `sequence[int32]`, but gsc reports:

```text
GS0187 Class 'Detector' does not implement interface method 'IDetector.Detect'.
```

## Expected behavior

A `sequence[T]` iterator method must satisfy an interface method returning `System.Collections.Generic.IEnumerable<T>` when its parameters and accessibility match.

## Fix verification

Add a compile-and-run test that constructs the implementation through the interface, enumerates `Detect("abc")`, and verifies the single value is `3`. Include both an iterator block and a non-iterator method explicitly returning `sequence[T]` to determine whether the issue is alias identity or iterator lowering.

PR #3133, opened 2026-08-03
Add deinit inheritance boundary coverage

### LINKED ISSUE #3131: Deinit boundary GS0510: the reported inheritance shape is pinned by no test, and one assertion is vacuous
## Summary

PR #3052 (merged) delivered the `GS0510` interpreter deinitializer boundary for #2988. The rubber-duck review verified the behaviour is correct on the reported shape but found it is **pinned by no test**, plus one vacuous assertion. Filing the residual so it is not lost.

Source: `gh pr view 3052 --comments`, `#issuecomment-5161913358`.

## 1. The issue's own repro shape is unpinned

#2988's repro is an **inheritance** shape — `CachedResource : Resource`, with both classes declaring a `deinit`. The reviewer hand-verified it produces exactly **2 correct `GS0510` warnings**, so the behaviour is right.

But nothing in the suite holds it:

- `samples/Deinit.gs` exists only as a hash in `refactoring-baseline.json`
- the sample conformance gate is **emit-only** (`test/Compiler.Tests/LanguageConformance/SampleConformanceTests.cs:16`/`:30-58`/`:144`), so nothing runs that sample under `gsi` or bare `gsc`
- **none of the 164 added test lines use a derived class** — every added case declares `deinit` on a single, non-derived class

So the one structural feature that distinguishes the reported shape — a deinit inherited alongside a deinit — is exercised by zero tests. A regression that reported 1 warning instead of 2 would pass the entire suite.

**Fix:** add a derived-class case to the existing test file. Two classes in a hierarchy, both with `deinit`, asserting exactly two `GS0510` diagnostics.

## 2. Vacuous assertion

`test/.../Issue698DeinitBinderTests.cs:55` uses `Assert.All` as its **only** assertion. `Assert.All` passes trivially on an empty collection, so if the diagnostic set ever became empty the test would still pass. Risk is low because `GS0510` is positively pinned elsewhere, but the assertion as written cannot fail for the reason it exists.

**Fix:** add an `Assert.Contains` (or an explicit count assertion) alongside it.

## Not in scope

The reviewer also observed that a runtime `GS9999` under `gsi` renders as `(1,1,1,7)` with **no file path**. This is **byte-identical on the merged base**, so it is pre-existing and not caused by #3052. Recording it here only so the next person to see it does not attribute it to the deinit work; it is not part of this issue.

## Context

The interpreter deliberately does **not** implement deinitializers — that is a settled policy decision recorded in #2988 (explicit boundary plus diagnostic). This issue is purely about test coverage for the boundary that was shipped, not about changing that decision.

DavidObando:
## Summary

- add the reported `CachedResource : Resource` inheritance shape to deinitializer boundary tests
- assert exactly two `GS0510` warnings, one for each declaring class
- exercise compiler evaluation, compiler emission, and interpreter drivers through the existing driver matrix helper
- use emitted execution output as the oracle for both evaluating drivers instead of freezing a literal output
- strengthen the binder diagnostic check so an empty collection cannot pass vacuously
- test-only change; no product code changed

Fixes #3131

## Driver matrix

Measured on `main` `787166a9` and this head:

| Driver | RC | `GS0510` | Output |
|---|---:|---:|---|
| bare `gsc` | 0 | 2 | equals emitted execution output, plus compiler diagnostic rendering and `Success.` |
| `gsc /out:` | 0 | 0 | emitted execution oracle; assembly runs successfully |
| `gsi` | 0 | 2 | equals emitted execution output |

This matrix regression-proofs the boundary against future restructuring; the columns are not three independent live risk paths today. `Compilation.Emit()` contains no call to `Evaluate()`, while `GS0510` is reported only inside `Evaluate()`, so an emit false positive is structurally impossible. Bare `gsc` and `gsi` both use `Compilation.Evaluate`, so they share the relevant product path.

## Mutation proof

Both product mutants were applied at exactly one site on `main` `787166a9` and the merged tree, built successfully, then restored.

| Mutant | Main without PR | Merged tree |
|---|---|---|
| M1: skip a derived deinitializer when its base declares one | `Issue2988DeinitInterpreterTests`: 4/4 pass | 2 failures of 7: compiler evaluation and interpreter each observe one warning instead of two |
| M2: invert the source-tree membership condition so `GS0510` is absent | `Issue698DeinitBinderTests`: 6/6 pass under old vacuous `Assert.All` | 1 failure of 6: `Assert.Single`, `Collection: []` |

The emitted matrix row remains green under M1, as expected: emission is the oracle and does not evaluate or report `GS0510`.

## Validation

Merged tree over `main` `787166a9` equals head tree `4cb09a91ff7d64bcb0edb584c3a9f63ecf501b15`.

- solution build: 0 warnings, 0 errors
- unfiltered `Interpreter.Tests`: 1,204 passed, 0 failed
- unfiltered `Core.Tests`: 7,106 passed, 1 skipped, 0 failed
- M1 two-sided mutation: main green, merged red
- M2 two-sided mutation: main green, merged red

### DIFF
--- a/test/Core.Tests/CodeAnalysis/Binding/Issue698DeinitBinderTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Binding/Issue698DeinitBinderTests.cs
@@ -52,6 +52,7 @@ class Resource {
 }
 ";
         var (result, _) = EvaluateAndGetStructs(source);
+        Assert.Single(result.Diagnostics, diagnostic => diagnostic.Id == "GS0510");
         Assert.All(result.Diagnostics, diagnostic => Assert.Equal("GS0510", diagnostic.Id));
     }
 
--- a/test/Interpreter.Tests/Issue2988DeinitInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue2988DeinitInterpreterTests.cs
@@ -14,6 +14,7 @@ namespace GSharp.Interpreter.Tests;
 /// <summary>
 /// Interpreter boundary coverage for CLR GC finalizers declared with <c>deinit</c>.
 /// </summary>
+[Collection("ConsoleIo")]
 public class Issue2988DeinitInterpreterTests
 {
     [Fact]
@@ -63,10 +64,14 @@ public class Issue2988DeinitInterpreterTests
         Assert.DoesNotContain(next.Diagnostics, diagnostic => diagnostic.Id == "GS0510");
     }
 
-    [Fact]
-    public void InheritedDeinitializersWarnOncePerDeclaringClass()
+    [Theory]
+    [InlineData(Issue3010EntryPointDriverMatrixTests.Driver.CompilerEvaluation)]
+    [InlineData(Issue3010EntryPointDriverMatrixTests.Driver.CompilerEmission)]
+    [InlineData(Issue3010EntryPointDriverMatrixTests.Driver.Interpreter)]
+    public void InheritedDeinitializersWarnOncePerDeclaringClass(
+        Issue3010EntryPointDriverMatrixTests.Driver driver)
     {
-        var source = """
+        const string Source = """
             import System
 
             open class Resource(Tag string) {
@@ -89,18 +94,48 @@ public class Issue2988DeinitInterpreterTests
             GC.KeepAlive(resource)
             """;
 
-        var cell = new SessionEngine { CaptureConsole = true }.Evaluate(source);
+        var emitted = Issue3010EntryPointDriverMatrixTests.Run(
+            "inherited-deinit-emitted",
+            Source,
+            Issue3010EntryPointDriverMatrixTests.Driver.CompilerEmission);
+        Assert.Equal(0, emitted.ExitCode);
+        Assert.NotEmpty(emitted.StandardOutput);
+        Assert.Equal(string.Empty, emitted.StandardError);
 
-        Assert.False(cell.HasError);
-        Assert.Equal("body-44\n", cell.Output);
-        var warnings = cell.Diagnostics
-            .Where(diagnostic => diagnostic.Id == "GS0510")
-            .OrderBy(diagnostic => diagnostic.Message)
-            .ToArray();
-        Assert.Collection(
-            warnings,
-            warning => AssertWarning(warning, "CachedResource"),
-            warning => AssertWarning(warning, "Resource"));
+        var result = driver == Issue3010EntryPointDriverMatrixTests.Driver.CompilerEmission
+            ? emitted
+            : Issue3010EntryPointDriverMatrixTests.Run(
+                "inherited-deinit-" + driver,
+                Source,
+                driver);
+        Assert.Equal(0, result.ExitCode);
+
+        string warningOutput;
+        switch (driver)
+        {
+            case Issue3010EntryPointDriverMatrixTests.Driver.CompilerEvaluation:
+                Assert.StartsWith(emitted.StandardOutput + "\n", result.StandardOutput, StringComparison.Ordinal);
+                Assert.EndsWith("Success.\n", result.StandardOutput, StringComparison.Ordinal);
+                warningOutput = result.StandardOutput[emitted.StandardOutput.Length..];
+                break;
+            case Issue3010EntryPointDriverMatrixTests.Driver.CompilerEmission:
+                Assert.Equal(emitted.StandardOutput, result.StandardOutput);
+                Assert.Equal(string.Empty, result.StandardError);
+                return;
+            case Issue3010EntryPointDriverMatrixTests.Driver.Interpreter:
+                Assert.Equal(emitted.StandardOutput, result.StandardOutput);
+                warningOutput = result.StandardError;
+                break;
+            default:
+                throw new InvalidOperationException($"Unexpected driver {driver}.");
+        }
+
+        Assert.Equal(
+            2,
+            warningOutput.Split('\n')
+                .Count(line => line.Contains("warning GS0510", StringComparison.Ordinal)));
+        Assert.Contains("class 'CachedResource'", warningOutput, StringComparison.Ordinal);
+        Assert.Contains("class 'Resource'", warningOutput, StringComparison.Ordinal);
     }
 
     [Fact]
--- a/test/Interpreter.Tests/Issue2988DeinitInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue2988DeinitInterpreterTests.cs
@@ -64,6 +64,46 @@ public class Issue2988DeinitInterpreterTests
     }
 
     [Fact]
+    public void InheritedDeinitializersWarnOncePerDeclaringClass()
+    {
+        var source = """
+            import System
+
+            open class Resource(Tag string) {
+                deinit {
+                    Console.WriteLine("base-deinit-11")
+                }
+            }
+
+            class CachedResource : Resource {
+                init(tag string) : base(tag) {
+                }
+
+                deinit {
+                    Console.WriteLine("derived-deinit-22")
+                }
+            }
+
+            var resource = CachedResource("held-33")
+            Console.WriteLine("body-44")
+            GC.KeepAlive(resource)
+            """;
+
+        var cell = new SessionEngine { CaptureConsole = true }.Evaluate(source);
+
+        Assert.False(cell.HasError);
+        Assert.Equal("body-44\n", cell.Output);
+        var warnings = cell.Diagnostics
+            .Where(diagnostic => diagnostic.Id == "GS0510")
+            .OrderBy(diagnostic => diagnostic.Message)
+            .ToArray();
+        Assert.Collection(
+            warnings,
+            warning => AssertWarning(warning, "CachedResource"),
+            warning => AssertWarning(warning, "Resource"));
+    }
+
+    [Fact]
     public void EscapingInstanceIsNotFinalizedAtScopeExitOrWhileReachable()
     {
         var source = """
--- a/test/Interpreter.Tests/Issue3010EntryPointDriverMatrixTests.cs
+++ b/test/Interpreter.Tests/Issue3010EntryPointDriverMatrixTests.cs
@@ -154,7 +154,7 @@ public class Issue3010EntryPointDriverMatrixTests
         Assert.Equal(string.Empty, result.StandardError);
     }
 
-    private static (int ExitCode, string StandardOutput, string StandardError) Run(
+    internal static (int ExitCode, string StandardOutput, string StandardError) Run(
         string name,
         string source,
         Driver driver)


DavidObando:
## Review: PR #3133 — `test/3131-deinit-inheritance-coverage`

**Verdict: MERGEABLE** — 0 BLOCKER, 1 SHOULD-FIX, 1 NIT.

This is a test-only PR, so the review question is not "does it pass" but **"would it fail if the product were wrong?"** I answered that with two-sided mutation: each new assertion was required to go RED on the merged tree *and* the same product mutant was required to stay GREEN on `main` without the PR. Both did. Neither addition is decoration, and the #3131 gap is genuinely closed.

### Anchors

| item | value |
|---|---|
| live `main` | `27a05798` |
| head | `0a4bd549` |
| merge base | `f7d9a5d3` (**5 behind / 1 ahead**) |
| `git merge-tree` output | 1 line, tree `350495f1`, **0 conflict markers** (output read, not rc) |
| merged tree reproduced | `350495f11ef2f4f3f97600dc6711f443259779c6` — exact |
| `BUILD_RC` merged / base | **0 / 0**, 0 warnings, 0 errors |
| merged `Interpreter.Tests` | **793 / 793** |
| merged `Core.Tests` | **7089 passed / 0 failed / 1 skipped** |

Binaries verified fresh by **mtime** in `Core/`, `Compiler/` and `Repl/` (excluding `ref/`), distinct between trees. Both new assertions confirmed absent on base (`grep` = 0), so the deltas are real additions.

---

## The mutation evidence — this is the review

### Mutant M1 — the inheritance shape (`Compilation.cs:331`)

Product mutant, condition **strengthening** (not `if (true)`): skip a class's deinitializer when its base already declares one — the plausible "dedupe inherited finalizers" regression.

```csharp
 if (ReferenceEquals(deinitializer?.Function, function)
     && SyntaxTrees.Contains(deinitializer.Declaration.SyntaxTree)
+    && (function.ReceiverType as StructSymbol)?.BaseClass?.Deinitializer == null)
```

Asserted applied: 1 site on each tree. `BUILD_RC=0` on both.

| tree | `Interpreter.Tests` | `Core.Tests` | outcome |
|---|---|---|---|
| **base `27a05798`, no PR** | **792 / 792 pass** | **7089 pass / 1 skip** | **regression ships undetected** |
| **merged (with PR)** | **792 pass, 1 FAILED of 793** | — | **caught** |

The single failure is exactly `Issue2988DeinitInterpreterTests.InheritedDeinitializersWarnOncePerDeclaringClass`. Nothing else moves.

**That is the two-sided proof #3131 asked for.** The regression clears **7,881 unfiltered tests** on `main` and is caught only by this PR's new test. The issue's claim — "a regression that reported 1 warning instead of 2 would pass the entire suite" — is now measured, not asserted, and this PR is what closes it.

### Mutant M2 — the vacuous assertion (`Issue698DeinitBinderTests.cs:55`)

Product mutant, condition **inversion**, chosen to make the `GS0510` set empty (the exact vacuity condition #3131 describes):

```csharp
- && SyntaxTrees.Contains(deinitializer.Declaration.SyntaxTree))
+ && !SyntaxTrees.Contains(deinitializer.Declaration.SyntaxTree))
```

Asserted applied: 1 site on each tree. `BUILD_RC=0` on both.

| tree | assertion in force | result |
|---|---|---|
| **base, no PR** | `Assert.All` only | **6 / 6 PASS** — passes vacuously with zero `GS0510` |
| **merged (with PR)** | `Assert.All` + new `Assert.Single` | **1 FAILED of 6** |

Failure is `Deinit_BodyHasAccessToFieldsAndThis` at `Issue698DeinitBinderTests.cs:55`:

```
Assert.Single() Failure: The collection did not contain any matching items
Expected:   (predicate expression)
Collection: []
```

The filter matched 6 tests on both trees, so the base green is a real result and not an empty selection. The one-line `Assert.Single` is therefore load-bearing.

**Both mutants restored**; `git diff -- src/` empty and `git status --porcelain` empty on both trees, with an exact behavioural round-trip (793/793 and 7089/0/1 after restore).

---

## The other required checks

**Runs unfiltered in the normal CI invocation — PASS.** `build/generate-ci-test-matrix.py` was executed directly: the `core` and `interpreter` matrix entries both have **`filter=<NONE>`**. Only the `compiler-*` and `cs2gs-*` shards carry `--filter`. Both new assertions therefore fire in the standard `dotnet test` step at `.github/workflows/build.yml:127-129`, with no special ordering and no bespoke command.

**`GS0510` asserted for the right reason, not merely appearing — PASS.** `AssertWarning` requires severity `Warning`, the message to contain `class '<Name>'` **and** "will not run under the interpreter", `Location.Text` to be non-null, and — the strongest part — the located span to render exactly `"deinit"`. So the warning is pinned to the right token in the right class, not just counted. `Assert.Collection` over the message-ordered list pins **exactly two**, one per declaring class, which is precisely what #3131 asked for.

**Coverage really does span the inheritance boundary — PASS.** The new case is `CachedResource : Resource` with `deinit` on both plus a `base(tag)` initializer, and it distinguishes the two classes by name. That exercises `ReferenceEquals(deinitializer?.Function, function)` and `deinitializer.DeclaringType.Name` in the one configuration where a derived type and its base both carry a deinitializer — the axis M1 shows was previously unpinned.

**Driver matrix, measured on both trees (probe dirs asserted empty):**

| driver | base `27a05798` | merged |
|---|---|---|
| bare `gsc` (evaluate) | rc=0, `GS0510`×2 | rc=0, `GS0510`×2 |
| `gsc /out:` (emit, oracle) | rc=0, `GS0510`×0 | rc=0, `GS0510`×0 |
| `gsi` | rc=0, `GS0510`×2 | rc=0, `GS0510`×2 |

Identical across trees, as expected for a test-only change, and matching the PR body's table. All three drivers print `body-44` only, confirming the deinitializers do not run. **A zero here is absence, not scope:** the harness was positive-controlled against a no-deinit program (`GS0510=0`) and the inheritance program (`GS0510=2`) on the same binaries.

---

## SHOULD-FIX 1 — the new test covers one driver of three and freezes its expectation instead of comparing to emit

`test/Interpreter.Tests/Issue2988DeinitInterpreterTests.cs:88` builds a single `SessionEngine` and asserts a hardcoded string:

```csharp
var cell = new SessionEngine { CaptureConsole = true }.Evaluate(source);
Assert.Equal("body-44\n", cell.Output);
```

The PR body presents a three-column driver table (bare `gsc` 2 / emit 0 / `gsi` 2) — but **no test holds any of it**. `GS0510` appears in exactly two test files repo-wide, both interpreter/binder; nothing asserts the emit column at all, and nothing compares evaluated output against emitted output. The expectation is frozen, so a new shape cannot be added to this harness without hand-authoring its golden — which is how the last regressions in this area stayed hidden.

**Severity is SHOULD-FIX, not BLOCKER, and I want to be precise about why:** I checked whether the emit column is actually at risk and it is not. `Compilation.Emit()` (`Compilation.cs:440-679`) contains **zero** calls to `Evaluate()`, and the `GS0510` report site lives inside `Evaluate` at `Compilation.cs:334`, so a false positive on the emit path is structurally impossible rather than untested. Bare `gsc` and `gsi` likewise share `Compilation.Evaluate`, so they are not independent columns for this diagnostic. The gap is real but narrow.

**Recommended fix, with no new infrastructure:** this same test project already defines the pattern at `test/Interpreter.Tests/Issue3010EntryPointDriverMatrixTests.cs:20-24`:

```csharp
public enum Driver { CompilerEvaluation, CompilerEmission, Interpreter }
```

Converting `InheritedDeinitializersWarnOncePerDeclaringClass` to a `[Theory]` over that enum, asserting the evaluating drivers' stdout equals the emitted assembly's stdout rather than a literal, would pin all three columns and make the PR body's table self-verifying.

## NIT 1 — the PR's anchors and one suite number are stale

The body reports measurements "against `main` `f7d9a5d3`". That is the **merge base**; live `main` is `27a05798` and the branch is **5 behind**. Consequently the reported `unfiltered Interpreter.Tests: 724 passed` is stale — on the merged tree at live `main` it is **793**. `Core.Tests` 7,089 passed / 1 skipped matches exactly. Worth refreshing so the numbers can be reproduced.

---

### Disclosed as not run
`Compiler.Tests`, `Cs2Gs.Tests`, LSP / SDK / Extensions / Analyzers were not run on the merged tree; e2e is prohibited locally. The change touches no product code and no shared test infrastructure, so those projects cannot observe it. One mutant build (base + M1) was started while 1-minute load read 34.4, above my build threshold — naming it rather than hiding it; test outcomes are deterministic pass/fail and the result was corroborated by the complementary merged-tree run at load 25. All other builds were taken at load 12–25.

DavidObando:
## Round-2 addendum — re-anchored to new head `88d7451a`, plus a denominator correction I owe

My earlier verdict on this PR was measured at main `27a05798` / head `0a4bd549`. Two things have changed since: the head moved to **`88d7451a`**, and a fleet rule landed that makes my earlier "not run" list wrong. Both are addressed below.

**Verdict unchanged: MERGEABLE.** 0 BLOCKER, 0 SHOULD-FIX (the round-1 SHOULD-FIX is genuinely resolved and mutation-proven), 2 NIT.

---

### 0. Correction I owe: the test-project denominator was wrong

My round-1 comment listed the not-run projects as `Compiler.Tests`, `Cs2Gs.Tests`, `LanguageServer.Tests`, `Sdk.Tests`, `Extensions.Tests`, `InternalAnalyzers.Tests` — which, with the 2 I ran, implies a denominator of 8. It is **9**. I omitted `tools/gsgen/GSharp.GeneratorHost.Tests`.

The part worth recording: `gsharp-generatorhost` appeared **in the CI-matrix output I posted myself** in that same comment. I published the evidence that contradicted my own enumeration and walked past it. The denominator must come from `GSharp.sln`, never from `ls test/`.

Re-measured on the merged tree, **7 of 9 projects, 10,817 tests, 0 failures, 1 skipped**:

| project | result |
|---|---|
| `test/Core.Tests` | 7106 passed / 0 failed / 1 skipped |
| `test/Interpreter.Tests` | 1204 / 1204 |
| `tools/cs2gs/Cs2Gs.Tests` | 1903 / 1903 (16 m 21 s) |
| `test/LanguageServer.Tests` | 462 / 462 |
| `test/Extensions.Tests` | 104 / 104 |
| `tools/gsgen/GSharp.GeneratorHost.Tests` | 31 / 31 |
| `test/InternalAnalyzers.Tests` | 7 / 7 |
| `test/Compiler.Tests` | **NOT RUN** — ~1 h+ locally; CI shards it |
| `test/Sdk.Tests` | **NOT RUN** — needs the local SDK package, out of bounds for me |

Stated as evidence rather than assumption: the three-dot diff from merge base `787166a9` touches **zero `src/` files** (3 files, all under `test/`), so no project can observe a product change from this PR. `Cs2Gs.Tests` green is the meaningful confirmation there, since it drives real `gsc.dll` out-of-process.

### 1. Re-anchoring

- New head **`88d7451a`**; merge base `787166a9`; branch 4 behind / 2 ahead of live main.
- Built the merged tree at main `5b71a74d` → predicted tree `51a62ae1`, reproduced exactly. `BUILD_RC=0`, 0 warnings, 0 errors.
- Main moved to `734eb94b` mid-measurement. Re-verified: `MERGE_RC=0`, merge-tree output **1 line, 0 conflict markers**, predicted tree `680d33b2`.
- Because a clean merge-tree is a floor and not a proof, I checked file overlap directly: main's 2 new commits (`f50aba4e`, `734eb94b`, issue #3093) touch sequence/interface product code and their own tests. **Intersection with this PR's 3 files is empty.** `Compilation.cs` and `StructSymbol.cs` are untouched, so the GS0510 behaviour I measured is unaffected.
- mtimes taken in epoch seconds and sorted numerically across `Core/`, `Compiler/`, `Repl/`, excluding `ref/` — all fresh against the build.

### 2. The round-1 SHOULD-FIX is genuinely fixed, and the fix is mutation-proven

Round 1 flagged that the new test used only `SessionEngine` and froze `Assert.Equal("body-44\n", cell.Output)`, so no test held the 3-driver table the PR body claimed. At this head the test is a `[Theory]` over all three drivers (`Issue2988DeinitInterpreterTests.cs:67-72`), the frozen literal is gone, and emit is run first as a live oracle.

This is not a cosmetic reshape — **the two new driver rows are killable**:

**M1** — product mutant at `src/Core/CodeAnalysis/Compilation/Compilation.cs:337-339`, condition *strengthening* (not `if (true)`):

```csharp
if (ReferenceEquals(deinitializer?.Function, function)
    && (function.ReceiverType as StructSymbol)?.BaseClass?.Deinitializer == null   // mutant
    && SyntaxTrees.Contains(deinitializer.Declaration.SyntaxTree))
```

Asserted applied: 1 site found, 1 site changed, `BUILD_RC=0`, 0 warnings. Result on the merged tree — **2 failed of 1204**:

```
Failed InheritedDeinitializersWarnOncePerDeclaringClass(driver: CompilerEvaluation)
Failed InheritedDeinitializersWarnOncePerDeclaringClass(driver: Interpreter)
```

The `CompilerEmission` row survives, and **correctly so**: `Compilation.Emit()` contains zero calls to `Evaluate()`, so `GS0510` is structurally impossible on the emit path. Round 1 supplied the other side of the two-sided proof: the same mutant on main *without* this PR clears Interpreter 792/792 and Core 7089/0/1 undetected. The regression this PR's test catches is invisible to ~7,900 unfiltered tests otherwise.

**M2** — product mutant inverting `SyntaxTrees.Contains(...)` (empty GS0510 set), asserted applied at 1 site. Filter matched **6** tests (so the selection is non-empty), **1 failed**: `Deinit_BodyHasAccessToFieldsAndThis` — precisely the test the `+1` `Assert.Single` line at `Issue698DeinitBinderTests.cs:55` guards. That hunk was byte-identical across the rebase.

Round-trip after both mutants: source restored, `git diff -- src/` empty, `git status --porcelain` empty, rebuilt, Core filter **6/6**, Interpreter **1204/1204** — exact.

### 3. NIT — the `CompilerEmission` row asserts nothing new

`Issue2988DeinitInterpreterTests.cs:105-110` assigns `result = emitted` for that driver. So at `:122`, `Assert.Equal(emitted.StandardOutput, result.StandardOutput)` compares an object with itself, and `:123` repeats `:103`. Every assertion that row executes has already run at `:101-103`, which **all three** rows execute.

Consequence: no product mutant can kill that row without also killing the other two — a broken emit fails `:101` for every row. It does exercise the emit driver, so it is not decoration in the dangerous sense, but it contributes no assertion. Either drop it from the `[InlineData]` set or give it a distinct assertion.

### 4. NIT — the emit oracle is degenerate for the shape it was given

Measured on the built merged tree, `emitted.StandardOutput` for this test's source is exactly `"body-44\n"` — **8 bytes**. Neither `Console.WriteLine("base-deinit-11")` nor `Console.WriteLine("derived-deinit-22")` inside the two `deinit` blocks runs in **any** column, because `GC.KeepAlive(resource)` holds the instance reachable to process exit and .NET does not run finalizers at exit.

So the differential `Assert.Equal(emitted.StandardOutput, result.StandardOutput)` carries no deinit-specific signal. Delete both `WriteLine`s from the deinit bodies and the theory passes identically. For this source the computed oracle is numerically the same string as the frozen literal it replaced.

The fix is actionable — measured on the same tree, making the instance unreachable and collecting:

```gsharp
func Make() {
    var r = CachedResource("held-33")
    Console.WriteLine("body-44")
}

Make()
GC.Collect()
GC.WaitForPendingFinalizers()
Console.WriteLine("end-55")
```

| driver | stdout |
|---|---|
| `gsc /out:` (emit) | `body-44` / `derived-deinit-22` / `base-deinit-11` / `end-55` |
| `gsi` | `body-44` / `end-55` |

Both deinits run under emit, in derived→base order, and the interpreter's omission becomes visible in the output rather than only in a warning.

Worth stating plainly: in that shape the author's central equation `Assert.Equal(emitted.StandardOutput, result.StandardOutput)` becomes **false**. The current assertion holds *only because* the chosen shape makes emit run no deinits at all, so it cannot be extended to a shape that genuinely exercises finalization without restructuring the comparison.

### 5. A finding I withdraw

I began forming a stronger version of §4 — that inherited-deinit *execution* semantics (derived-then-base chaining) were pinned by nothing. **That is wrong.** `test/Compiler.Tests/Emit/Issue698DeinitEmitTests.cs:99` `Deinit_OnSubclass_RunsBothFinalizers` pins exactly that on the same `open class Resource` / `class CachedResource : Resource` shape, including ordering via `Assert.True(dIdx < bIdx, "derived deinit must run before base deinit")`.

That is why §4 is a NIT and not a SHOULD-FIX: the signal the degenerate oracle fails to carry is already covered elsewhere. What remains is a test whose source *implies* coverage it does not provide.

### 6. Driver probe re-run at live main

Own probe directory, asserted empty before use; harness positive-controlled in round 1 against a no-deinit source (0 vs 2):

| driver | `GS0510` count | rc | stdout |
|---|---|---|---|
| bare `gsc` (evaluate) | 2 | 0 | `body-44` + warnings on stdout, `Success.` |
| `gsc /out:` (emit) | 0 | 0 | `body-44` |
| `gsi` (interpret) | 2 | 0 | `body-44`, warnings on **stderr** |

Identical to round 1. Both class names (`Resource`, `CachedResource`) appear exactly once each in the warning text, which is the once-per-declaring-class property #3131 was filed about.

DavidObando:
Post-merge validation addendum for fleet bulletin:

- `dotnet sln GSharp.sln list` enumerates 9 test projects; `Sdk.Tests` is one of them.
- unfiltered `Sdk.Tests` on live `main` `61fe9cb9`: **59 passed, 0 failed, 0 skipped** (`rc=0`).
- initial run in a fresh worktree produced no test result: build failed `NETSDK1004` because `out/obj/Gsgen.Cli/project.assets.json` did not exist. A normal solution build restored dependencies (`BUILD_RC=0`); rerun then passed 59/59. This was a harness setup miss, not a test failure.
- no writes were made manually to `~/.nuget/packages/gsharp.net.sdk/`.

PR #3138, opened 2026-08-03
Document gsc optimize switch

### LINKED ISSUE #3136: gsc /optimize is undocumented: exhaustive switch table in website/docs/tooling/gsc.md omits it, and no gate covers that table
## Summary

PR #3109 (merged, issue #2920) added the `/optimize` switch to `gsc`. It changes the **default JIT behaviour of every assembly the compiler emits**, and it is documented nowhere in `docs/` or `website/docs/` — only in `--help`.

Found by a rubber-duck during the #3109 review.

## Measured

`grep -rn "/optimize" docs/ website/docs/` returns **zero hits**.

Meanwhile `website/docs/tooling/gsc.md:57-96` is an **exhaustive** switch table — it documents even the no-op `/lib:` switch. So the omission is a genuine gap in a reference that otherwise claims completeness, not a deliberate editorial choice.

## Why it matters more than a typical docs gap

The switch is not cosmetic. Measured on the same `gsc` with flags verified off the produced assembly:

| build | `DebuggableAttribute` | JIT optimizer |
|---|---|---|
| `/optimize+` | `2` | **enabled** |
| `/optimize-` | `257` | disabled |

and a representative program ran **5 ms vs 34 ms — 6.8× — with identical checksums**. A user who cannot find the switch in the reference has no way to discover that their emitted assemblies are running unoptimized.

## Also worth fixing while there

**No docs gate covers `website/docs/tooling/gsc.md`.** The existing diagnostics gate (`test/Core.Tests/CodeAnalysis/DiagnosticIdUniquenessTests.cs`) validates the diagnostics catalogues against `DiagnosticDescriptors` in code, but nothing validates the switch table against the actual switch set. That is why an exhaustive table silently stopped being exhaustive.

If a gate for the switch table is cheap, add it — the same class of defect (a reference that claims completeness and is not checked) is already being addressed for the diagnostics catalogues in #3124. If it is not cheap, say so and just fix the content; do not build a documentation framework for one table.

## Bar

Document `/optimize` in the `gsc` reference table in the same shape as its neighbours, in **both** `docs/` and `website/docs/` if both carry the table, and confirm the two agree.

State plainly what the default is, since that is the question a reader will actually have.

This is a **docs** change. Do not touch product code. If your diff grows beyond the table plus an optional test, stop — you have left the lazy path.

DavidObando:
## Summary

- document `/optimize`, `/optimize+`, `/optimize-`, and boolean forms in the exhaustive `gsc` switch table;
- state that JIT optimization is enabled by default;
- update both the current and versioned 0.3 references with identical rows.

No equivalent switch table exists under the root `docs/` tree. I did not add a new completeness framework: the existing docs gate builds the website, but no current gate or shared switch catalogue offers a cheap source-of-truth comparison.

## Independent verification

Each assembly was compiled into its own asserted-empty output directory with `/debug:portable`, then loaded from bytes and inspected through `DebuggableAttribute`:

| invocation | `DebuggingFlags` |
|---|---:|
| `/optimize+` | `2` |
| `/optimize-` | `257` |
| no optimize switch | `2` |

The default therefore matches `/optimize+`.

- current and versioned `gsc.md`: byte-identical, 139 lines each;
- unfiltered docs gate: `npm ci` + `npm run build` passed.

Fixes #3136

### DIFF
--- a/website/docs/tooling/gsc.md
+++ b/website/docs/tooling/gsc.md
@@ -79,6 +79,9 @@ This option list was checked against `dotnet out\bin\Release\Compiler\gsc.dll --
 | `/warnaserror:<ids>` | Promote the listed warning IDs to errors. |
 | `/warnaserror+:<ids>` | Alias for promoting the listed warning IDs to errors. |
 | `/warnaserror-:<ids>` | Keep the listed IDs as warnings even when global `/warnaserror` is set. |
+| `/optimize`, `/optimize+` | Enable JIT optimization for emitted assemblies. This is the default. |
+| `/optimize-` | Disable JIT optimization for emitted assemblies. |
+| `/optimize:true`, `/optimize:false` | Enable or disable JIT optimization. Also accepts `1`, `0`, `on`, `off`, `yes`, and `no`. |
 | `/debug` | Emit a sidecar Portable PDB. Equivalent to `/debug:portable`. |
 | `/debug+` | Emit a sidecar Portable PDB. |
 | `/debug-` | Disable debug information. |
--- a/website/versioned_docs/version-0.3/tooling/gsc.md
+++ b/website/versioned_docs/version-0.3/tooling/gsc.md
@@ -79,6 +79,9 @@ This option list was checked against `dotnet out\bin\Release\Compiler\gsc.dll --
 | `/warnaserror:<ids>` | Promote the listed warning IDs to errors. |
 | `/warnaserror+:<ids>` | Alias for promoting the listed warning IDs to errors. |
 | `/warnaserror-:<ids>` | Keep the listed IDs as warnings even when global `/warnaserror` is set. |
+| `/optimize`, `/optimize+` | Enable JIT optimization for emitted assemblies. This is the default. |
+| `/optimize-` | Disable JIT optimization for emitted assemblies. |
+| `/optimize:true`, `/optimize:false` | Enable or disable JIT optimization. Also accepts `1`, `0`, `on`, `off`, `yes`, and `no`. |
 | `/debug` | Emit a sidecar Portable PDB. Equivalent to `/debug:portable`. |
 | `/debug+` | Emit a sidecar Portable PDB. |
 | `/debug-` | Disable debug information. |


DavidObando:

Docs-only, +6/−0 across 2 files (three-dot from merge base `27a05798`). Measured on `main` `d240eb24`; since the PR changes no code, the post-merge compiler *is* `main`'s, so that is where I measured. Merge asserted by tree hash (`fb47b448`, predicted = actual), clean, no duplicate rows.

### 1. Is `/optimize+` really the default? **Yes — verified off the assembly, not inferred**

Built with **no `/optimize` switch at all** and read `DebuggableAttribute` from the metadata blob:

| build | `DebuggableAttribute` |
|---|---|
| `/debug:portable` (no optimize switch) | **2** |
| `/debug:portable /optimize` | 2 |
| `/debug:portable /optimize+` | 2 |
| `/debug:portable /optimize-` | **257** |

Default is byte-for-byte identical to `/optimize+`. The author's `2` / `257` / default-`2` reproduce exactly.

### 2. Accepted-values list — **complete and correct in both directions**

`Program.cs:1190-1194` (`ParseBoolFlag`) accepts exactly `true|1|on|yes` → true, `false|0|off|no` → false, else throws. Measured all eight, plus rejects:

- all 8 documented values accepted with correct semantics (`true/1/on/yes` → 2; `false/0/off/no` → 257)
- `enable`, `disable`, `y`, `n`, `t`, `f`, `2` → all rejected, `Unsupported boolean value: …`

No listed form is rejected, and no distinct accepted token is missing from the list.

### 3. Do the two copies agree? **Byte-identical, proven by blob hash**

Both files share the **same pre-image blob `631cdfe5`** and the **same post-image blob `e8fcb50d`**. Identical git blobs is a stronger statement than a passing diff. Same insertion point in both (after `/warnaserror-`, before `/debug`).

Worth noting for the `docs/` ⇄ `website/docs/` divergence concern: only these two copies of `gsc.md` exist — there is no `docs/tooling/gsc.md`. The pair is `website/docs/` and `website/versioned_docs/version-0.3/`, and byte-identity is the house convention there (38 of 45 pairs identical; `ref/diagnostics.md` is one of the 7 that differ, correctly untouched here).

### 4. The vacuity trap — **avoided, and it would have bitten exactly the load-bearing claim**

Confirmed with no `/debug` switch:

| build (no `/debug`) | result |
|---|---|
| no optimize switch | **ABSENT** |
| `/optimize+` | **ABSENT** |
| `/optimize-` | 257 |

The attribute is emitted only when `Pdb != null || !Optimize` (`ReflectionMetadataEmitter.cs:3908`). So without `/debug`, default and `/optimize+` are *both* absent and compare "identical" — which is precisely the pair the "This is the default" claim rests on. `/optimize-` still differs, so a careless reviewer would see one row work and assume the method sound. The author's `/debug:portable` was necessary, not incidental.

### Style and gate

The three rows mirror the existing `/deterministic` rows verbatim, including the value-list phrasing — correct house style. Documentation gate `Build docs` **ran and passed in 1m1s** (not skipped by the `website/**` path filter; `Deploy` correctly skipped for a PR). No test references `gsc.md`, corroborating that no completeness gate exists for this table.

---

**BLOCKER:** none.
**SHOULD-FIX:** none.

**NIT (optional, arguably correct as-is):** the parser also accepts case variants (`/optimize:TRUE`, `/optimize:Off`), the `=` separator (`/optimize=false`), and a bare `/optimize:` with empty value (which *enables*). All measured working. None are documented — but neither are they on `/deterministic` or `/implicitimports`, so omitting them is consistent with the page. Mentioning only if the page ever grows a general note on switch syntax.

Facts check out; ship it.

PR #3144, opened 2026-08-03
Synchronize diagnostic references and enforce row parity

### LINKED ISSUE #3124: Diagnostics docs gate never compares message text: 142 of 385 rows differ, GS0153 is factually wrong, and GS9998/GS9999 are each documented twice
## Summary

`docs/diagnostics.md` and `website/docs/ref/diagnostics.md` are supposed to be kept in sync. **142 of 385 shared diagnostic rows differ in content**, including at least one where the primary file is **factually wrong**, and the gate cannot see any of it.

I measured this myself after two delegated investigations failed to return, so the numbers below are first-hand.

## What the gate actually does — it is stricter than reported, but blind in a specific way

`test/Core.Tests/CodeAnalysis/DiagnosticIdUniquenessTests.cs:186-192` calls `AssertDocumentationMatches` **once per file**, comparing each file against `DiagnosticDescriptors` **in code** — not against the other file. For each file it checks:

- every descriptor ID is documented (`missing`)
- no documented ID lacks a descriptor (`unexpected`)
- documented **severity** equals descriptor severity (`mismatched`)
- no ID has conflicting severities *within* one file (`conflictingSeverities`)

So the common description — "bidirectional on row presence only" — is **wrong**, and in a good way: code is the source of truth and severity is verified. Credit where due.

**The hole is that the regex only ever captures two groups:**

```csharp
@"^\|\s*(GS\d{4})\s*\|\s*([^|]+?)\s*\|"
```

Group 1 is the ID, group 2 is the severity. **The message and description columns are never read, by either file, ever.** Nothing compares the two files to each other at all.

## Measured, current `main`

| | `docs/diagnostics.md` | `website/docs/ref/diagnostics.md` |
|---|---|---|
| rows matching the ID pattern | **387** | **385** |
| unique IDs | 385 | 385 |

- present in both: **385** — neither file is missing an ID (the gate does work)
- **identical rows: 243**
- **non-severity content differs: 83**
- **column-count differs: 58** — the website file omits the example column entirely
- **severity string differs: 1**

## The worst row: `docs/diagnostics.md` is simply wrong

**`GS0153`**

| source | text |
|---|---|
| `DiagnosticDescriptors.cs:71` | `ConstraintNotInterface` — *"Type '{0}' cannot be used as a type-parameter constraint because it is **neither an interface nor a class**."* |
| `website/docs/ref/diagnostics.md` | "Constraint is neither an interface nor a class… any interface — **sealed or not**, generic or not — or a base class; a value type such as a struct or enum is rejected." |
| `docs/diagnostics.md` | "**Interface constraint is not sealed.** A generic constraint interface **must be `sealed`** (i.e. not `open`)." |

The website matches the code. **The primary docs file describes a rule that does not exist** and tells the reader the opposite of the truth — that constraints must be `sealed`, when the compiler accepts any interface. A user debugging `GS0153` against `docs/diagnostics.md` would go looking for a sealing problem that is not there.

## Two more substantive divergences

**`GS0106` / `GS0109`** — the `record` keyword was removed. The website file marks both as historical and notes that `GS0307` replaces `GS0109` on legacy sources. `docs/diagnostics.md` still describes `record` as current syntax. One of these is telling users about a keyword the language no longer has.

**`GS0259`** — severity reads `Error (retired by ADR-0062)` in one file and `Error (retired)` in the other. Harmless, because `NormalizeDocumentedSeverity` collapses both to `Error`, but it shows the severity column is being used for free-form prose the gate silently discards.

## A second, separate hole: duplicate rows pass silently

`docs/diagnostics.md` has 387 rows for 385 IDs. **`GS9998` and `GS9999` are each documented twice, with different meanings:**

- `GS9999` row 1: "An unexpected exception was caught by the evaluator. The message text contains the original exception message."
- `GS9999` row 2: "**Fatal I/O error.** An unrecoverable file-system error occurred (permission denied, disk full, etc.) before or during assembly writing."

Both are plausible; they are not the same diagnostic. The gate survives this because `conflictingSeverities` only compares the **severity** column across duplicate rows — both say `Error` — and `documentedSeverities` then takes `group.First()`, silently discarding the second row. A duplicate ID with contradictory *meaning* is invisible.

This one matters operationally: I am the sole allocator of diagnostic IDs for this session and I have been reading these files to decide what is free. **A duplicated ID is exactly the failure mode allocation must not have.**

## Suggested fix, and its true cost

**Lazy fix, one file:** widen the regex to capture the remaining columns, and add a cross-file assertion that the two rows for each ID are equal after normalising the trailing-column count.

**Cost, measured:** turning that on today makes **142 of 385 rows** fail immediately. That is a real cleanup, not a one-line change, so it should land in two steps:

1. **Fix the content divergences first** (142 rows), starting with `GS0153`, `GS0106`, `GS0109` — the ones that are *wrong*, not merely differently worded. Deduplicate `GS9998`/`GS9999`.
2. **Then tighten the gate**, so it stays fixed.

Doing it in the other order produces a red gate nobody can merge past.

Two smaller wins available immediately and independently, both currently zero-cost:

- **Assert row count equals unique-ID count per file** — catches the `GS9998`/`GS9999` duplication now and would have caught it whenever it was introduced.
- Decide whether the website file is *meant* to omit the example column. If it is, the cross-file check should compare the columns both files have and stop counting those 58 as divergences; if it is not, that is 58 rows of missing documentation.

DavidObando:
Fixes #3124

## Summary

- Reconcile every diagnostic row in `docs/diagnostics.md` and `website/docs/ref/diagnostics.md`.
- Preserve the richer four-column row when one catalogue omitted an example, because both catalogues already use both shapes and the website declares itself a mirror.
- Split real diagnostic-ID collisions:
  - `GS9996`: source-generator execution failures.
  - `GS9997`: fatal compiler I/O failures.
  - `GS9999`: unexpected evaluator failures.
- Tighten the existing gate to compare every normalized row column and reject duplicate rows.

## Measurements

Live baseline on `f7d9a5d3`:

| Measurement | Before | After |
|---|---:|---:|
| Shared IDs | 387 | 389 |
| Differing text | 84 | 0 |
| Differing column count | 57 | 0 |
| Differing severity string | 1 | 0 |
| Total divergences | 142 | 0 |
| Duplicate rows in `docs/diagnostics.md` | 2 | 0 |

Both catalogues now contain 389 rows for 389 unique IDs.

## Collision evidence

- Existing `GS9999` references across `test/`, `samples/`, and `docs/`: 41 evaluator/other references versus 5 fatal-I/O references. The entrenched evaluator diagnostic keeps `GS9999`; the I/O site moves to `GS9997`.
- The four gsgen failure sites had four source references and zero ID-specific tests or documentation rows. They move to `GS9996`.
- `GS9998`'s duplicate rows described the same compiler/emit failure and were deduplicated without allocating another ID.

## Gate mutation

Changed only the `GS0001` description in `docs/diagnostics.md` to `MUTANT bad character input.` The tightened test failed:

```text
Failed: 1, Passed: 0
Diagnostic reference rows differ:
GS0001:
  docs/diagnostics.md: Error | MUTANT bad character input. | ...
  website/docs/ref/diagnostics.md: Error | Bad character input. | ...
```

The mutant was reverted. The clean gate passes.

## Validation

- Unfiltered `Core.Tests` on final merged tree: 0 failed, 7,113 passed, 1 skipped.
- Full Release suite on `d3b1d1c3` / main `7936c214`: 0 failed, 14,639 passed, 1 skipped.
- Predicted merged tree equals the tested HEAD tree.
- `dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true`: `BUILD_RC=0`, 0 warnings, 0 errors.
- Final-head CI: 30 passed, 3 skipped.

### DIFF
--- a/docs/diagnostics.md
+++ b/docs/diagnostics.md
@@ -68,10 +68,10 @@ IDs may be given as `GS0001`, `0001`, or the bare integer `1`; all three forms a
 | GS0103 | Error | Method receiver must be a struct or class declared in the same package. | Receiver type is a built-in or external type. |
 | GS0104 | Error | Reserved; not currently emitted. Zero-field `data class`/`data struct` declarations are supported (issue #2363, ADR-0029) with trivial equality/hash/ToString/copy semantics and no synthesized `Deconstruct`. | — |
 | GS0105 | Error | `inline struct` requires exactly one field. | `inline struct Foo { a int; b int }` has two fields. |
-| GS0106 | Error | `inline` cannot be combined with `data` or `record`. | `inline data struct Foo { … }` is not legal. |
+| GS0106 | Error | `inline` cannot be combined with `data`. | `inline data struct Foo { … }` is not legal. (Historical: the `record` keyword was removed by; pre-removal this diagnostic also covered `inline record`.) |
 | GS0107 | Error | `inline struct` cannot be combined with `open`. | `open inline struct Foo { … }` is not legal. |
 | GS0108 | Error | Inline struct synthesised member conflicts with an explicit declaration. | An `inline struct` auto-generates certain member names that cannot be re-declared. |
-| GS0109 | Error | `record` is an alias for `data struct` and cannot be combined with `data`. | `data record Foo { … }` — use either `data struct` or `record`. |
+| GS0109 | Error | `record` was an alias for `data struct` and could not be combined with `data`. GS0307 replaces this on legacy sources. | `data record Foo { … }` — use `data struct Foo` or `data class Foo`. |
 | GS0110 | Error | Empty enum declaration. | `enum Color {}` — an enum must have at least one member. |
 | GS0111 | Error | Duplicate enum member. | Two members in the same `enum` share a name. |
 | GS0112 | Error | Undefined enum member. | `Color.Purple` where `Purple` is not a declared member of `Color`. |
@@ -115,20 +115,20 @@ IDs may be given as `GS0001`, `0001`, or the bare integer `1`; all three forms a
 | GS0150 | Error | Type-parameter variance position violation. | A covariant type parameter used in a contravariant position. |
 | GS0151 | Error | Type argument inference failed. | The compiler could not infer a type argument from the call arguments. |
 | GS0152 | Error | Type argument does not satisfy constraint. | `f[MyStruct]()` where `MyStruct` does not implement the required interface constraint. |
-| GS0153 | Error | Interface constraint is not sealed. | A generic constraint interface must be `sealed` (i.e. not `open`). |
+| GS0153 | Error | Constraint is neither an interface nor a class. | A generic type-parameter constraint must be an interface (any interface — sealed or not, generic or not;) or a base class; a value type such as a struct or enum is rejected. |
 | GS0154 | Error | Wrong argument type. | A positional argument's type does not match the parameter type. |
 | GS0155 | Error | Cannot convert type. | An explicit cast between incompatible types. |
 | GS0156 | Error | Cannot convert implicitly; explicit conversion exists. | `int x = 3.14` — an explicit cast is available but was not written. |
 | GS0157 | Error | Cannot find type (possibly missing import). | A package-qualified type name that resolves to nothing. |
 | GS0158 | Error | Cannot find member. | A field or property access that does not resolve. |
 | GS0159 | Error | Cannot resolve function call. | A function name does not resolve, or its nullable receiver lacks valid non-null narrowing. |
-| GS0160 | Error | Ambiguous overload. | A call that matches more than one overload equally well. |
+| GS0160 | Error | Ambiguous overload. | A call that matches more than one overload equally well. Generic candidates are filtered against their `where`-constraints; the constraint-disjoint case usually resolves to one candidate, but two candidates with mutually-incomparable constraint specificity remain ambiguous and report this code. |
 | GS0161 | Error | `copy`/`with` receiver is not a `data struct`. | `.copy(…)` used on a plain `struct`. |
 | GS0162 | Error | Named arguments only supported for `data struct` `.copy(…)`. | Named arguments passed to a regular function. |
 | GS0163 | Error | Deconstruction field count mismatch. | `let (a, b) = p` where `p` is a `data struct` with three fields. |
 | GS0164 | Error | Deconstruction requires a tuple, `data struct`, or accessible `Deconstruct` method. | Deconstruction attempted on a type without a supported deconstruction shape. |
-| GS0165 | Error | Top-level statements may appear in at most one package per compilation. | Two or more `package` declarations in a single compilation each contain top-level statements (see [ADR-0066](adr/0066-top-level-statement-mechanics.md)). |
-| GS0166 | Warning | Top-level statements conflict with an explicit `Main` function. | Both top-level statements and a `func Main()` are present; TLS wins, the explicit `Main` is shadowed (see [ADR-0066](adr/0066-top-level-statement-mechanics.md) §4). |
+| GS0165 | Error | Top-level statements may appear in at most one package per compilation. | Two or more `package` declarations in a single compilation each contain top-level statements (see [ADR-0066](https://github.com/DavidObando/gsharp/blob/main/docs/adr/0066-top-level-statement-mechanics.md)). |
+| GS0166 | Warning | Top-level statements conflict with an explicit `Main` function. | Both top-level statements and a `func Main()` are present; TLS wins, the explicit `Main` is shadowed (see [ADR-0066](https://github.com/DavidObando/gsharp/blob/main/docs/adr/0066-top-level-statement-mechanics.md) §4). |
 | GS0167 | Error | Multi-assignment target/value count mismatch. | `a, b = 1, 2, 3` — three values for two targets. |
 | GS0168 | Error | `fallthrough` is not supported (ADR-0013). | `fallthrough` keyword used in a `switch` case body. |
 | GS0169 | Error | Duplicate `default` arm in `switch`. | Two `default:` arms inside one `switch` statement. |
@@ -293,7 +293,7 @@ Issue #1655: the IDs below used to collide with earlier, unrelated diagnostics (
 | GS9003 | Error | Variable not definitely assigned before `ref` use. | `ref x` where `x` has not been assigned. |
 | GS9004 | Error | By-ref value cannot escape its declaration scope. | Returning a `*T` (managed-pointer) value from a function, capturing a `*T` local in a closure, hoisting a `*T` local into an `async`/iterator state machine, or using a `*T` return type in a function literal. Also raised when returning a `ref struct` parameter annotated as `scoped`. |
 | GS9005 | Error | Cannot take the address of a constant. | `&myConst` where `myConst` is declared `const`. |
-| GS9006 | Error | Pointer type cannot be a field type. | A struct or class field (including static `shared` fields and top-level globals) declared with a `*T` (managed-pointer) type. |
+| GS9006 | Error | Pointer type cannot be a field type. | A struct or class field (including static `shared` fields and top-level globals) declared with a `*T` (managed-pointer) type **outside an `unsafe` context**. Inside an `unsafe` context `*T` is an unmanaged pointer and IS legal as a field type. |
 | GS9007 | Error | A type may contain at most one `shared` block. | A class or struct with two `shared { ... }` blocks; merge them into one. |
 | GS9008 | Error | A pointer bound by `fixed` cannot be captured by a closure because the closure may outlive the pin. | A lambda inside `fixed p *int32 = xs` captures `p`. |
 
@@ -303,13 +303,15 @@ Issue #1655: the IDs below used to collide with earlier, unrelated diagnostics (
 |----|----------|-------------|-----------------|
 | GS9100 | **Warning** | One or more assemblies supplied via `/r:` depend (transitively) on assemblies that were not also supplied, so the reference set is not a complete transitive closure. The compiler degrades gracefully — members whose signatures live in the missing assemblies are skipped rather than aborting the build — but the affected members become invisible. The message names the missing assemblies. Add the missing package/project reference (the SDK passes `@(ReferencePathWithRefAssemblies)`, MSBuild's full transitive closure, so this normally only appears with a hand-rolled `/r:` set). Suppress with `/nowarn:GS9100`. | `gsc /r:LibAsmA.dll app.gs` where `LibAsmA.dll` references `DepAsmB.dll` and `DepAsmB.dll` is not also passed. |
 
-### Internal / emit diagnostics (GS9998–GS9999)
+### Internal diagnostics (GS9996–GS9999)
 
-These diagnostics indicate an internal compiler problem. If you encounter them, please file an issue.
+These diagnostics indicate a fatal compiler or evaluator failure. If you encounter them, please file an issue.
 
 | ID | Severity | Description |
 |----|----------|-------------|
-| GS9998 | Error | An unexpected `NotSupportedException` or `InvalidOperationException` was raised during IL emission. The message text contains the original exception message. |
+| GS9996 | Error | Source generator execution failure. `gsc` could not locate or launch `gsgen`, `gsgen` timed out, or it exited nonzero without producing its own diagnostics. |
+| GS9997 | Error | Fatal compiler I/O error. An unrecoverable file-system error occurred (permission denied, disk full, etc.) before or during assembly writing. |
+| GS9998 | Error | Internal compiler error (emit-time failure). The emit pipeline encountered an unexpected state and could not produce valid IL. The diagnostic message includes the exception type and a brief description of the failure. |
 | GS9999 | Error | An unexpected exception was caught by the evaluator. The message text contains the original exception message. |
 
 #### Interpreter compiled-only boundaries
@@ -691,7 +693,7 @@ value types are rejected.
 
 | ID | Severity | Description |
 |----|----------|-------------|
-| GS0352 | Error | `ref`/`out`/`in` parameter `{name}` requires a blittable pointee; `{type}` is not blittable. Use a blittable primitive (e.g. `int32`, `int64`, `nint`), or a struct annotated with `@StructLayout(LayoutKind.Sequential)`. |
+| GS0352 | Error | `'ref'/'out'/'in'` parameter `<name>` requires a blittable pointee; `<T>` is not blittable. Use a blittable primitive (`int8`…`int64`, `nint`/`nuint`, `float32`/`float64`), or a struct annotated with `@StructLayout(LayoutKind.Sequential)`. |
 
 The struct-pointee path continues to use GS0349 (same remediation as the
 by-value struct case). GS0326 still fires for the remaining function-shape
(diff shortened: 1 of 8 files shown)


DavidObando:
# Rubber-duck review — PR #3144 (issue #3124)

**Verdict: MERGEABLE.** No blocking issues. The central claim — that the docs gate is now capable of failing — is verified two-sided, and the headline count reproduces exactly. Findings below are all NIT or informational.

---

## Anchors (please read: two of the PR body's SHAs are dead)

| item | value |
|---|---|
| head reviewed | **`94e06880`** |
| `main` at review time | **`734eb94b`** |
| merge base | `734eb94b` — **`main` is an ancestor of the head**, so the merge is a fast-forward |
| `git merge-tree --write-tree` output | `49133e79` |
| head's own tree | `49133e79` — **identical**, so merged tree == head tree |
| merged-tree `BUILD_RC` | **0** (Release, `ContinuousIntegrationBuild=true`) |

Three SHA corrections, none of which change the outcome but all of which change what is *provable*:

- I was briefed head **`9548f530`**. That commit was **orphaned by a force-push** (two force-pushes on this branch, the last ~34 min before I began). It is not an ancestor of the current head.
- The PR body's validation section reports the full Release suite on **`d3b1d1c3`**. **That object does not exist** — `git cat-file -t d3b1d1c3` fails. It was orphaned by the same force-push. The numbers may well have been true when produced, but they are **not attributable to the head under review** and cannot be re-derived.
- `7936c214` *is* a real ancestor of the head, but it is an older `main`.

Because of this I re-ran everything at the true head rather than relying on any figure in the PR body.

---

## 1. Is the gate now capable of failing? — **Yes, proven two-sided**

The important half is the negative control, which nobody had run:

**Old gate + old catalogues (142 divergences and 2 duplicate rows physically present) → GREEN, 4/4.** Built at `734eb94b`, `BUILD_RC=0`. The gate genuinely could not fail. That is the issue's core assertion and it is correct.

Then, at the head:

| probe | catalogues | gate | result |
|---|---|---|---|
| baseline | head | new | **GREEN 4/4** |
| MD7a | base catalogues verbatim | new | **RED** — `docs/diagnostics.md contains duplicate diagnostic rows: GS9998: 2 rows, GS9999: 2 rows` |
| MD7b | base catalogues, deduped + the two new rows added, so only parity can fail | new | **RED — reporting exactly `142` differing rows** |

That 142 is the gate's own count, and it matches my independent census of the same trees (below) exactly.

## 2. Does the parity check compare ALL columns? — **Yes**

Mutants, each applied to a single row and asserted applied; the test assembly md5 (`ec7927f6…`) was re-checked before every run so `--no-build` is provably measuring the same binary and only the docs file varies.

| mutant | what it changes | expect | result |
|---|---|---|---|
| MD1 | description column (the author's own mutant) | RED | **RED** |
| MD2 | **example column only** — the column the old two-group regex never read | RED | **RED** |
| MD3 | **deletes** the example column from a website row (column-count parity) | RED | **RED** |
| MD4 | duplicates a row (the `GS9998`/`GS9999` hole) | RED | **RED** |
| MD5 | severity **prose** only (`Error (retired by ADR-0062)` → `Error (retired)`) | RED | **RED** |
| MD6a | whitespace around separators only — *control* | GREEN | **GREEN** |
| MD6b | adds a trailing **empty** column — *control* | GREEN | **GREEN** |

MD5 is the discriminating one: `NormalizeDocumentedSeverity` deliberately collapses both spellings to `Error`, so the pre-existing severity assertion is blind to it. The failure came from the **new** row-parity assertion (`DiagnosticIdUniquenessTests.cs:217`), not the severity check — so the new check is strictly stronger, not a restatement.

The two green controls matter as much as the reds: the normalization is tolerant of formatting and of an empty trailing cell, so this will not turn into a gate that fires on whitespace.

## 3. Are the counts right? — **Total and endpoint exact; the 84/57/1 sub-split is a partition, not independent axes**

Re-derived from the trees with an independent reimplementation of the gate's own regex and normalization:

| | docs rows | docs unique | web rows | web unique | shared | identical | **divergent** |
|---|---|---|---|---|---|---|---|
| base `734eb94b` | 389 | 387 (`GS9998`×2, `GS9999`×2) | 387 | 387 | 387 | 245 | **142** |
| head `94e06880` | 389 | **389** | 389 | **389** | 389 | 389 | **0** |

The PR body's "389 rows for 389 unique IDs" and "142 → 0" both reproduce exactly.

On the coordinator's specific question — whether the categories were made to sum: **33 rows diverge on two axes at once** (column-count *and* body text). Counted as independent axes the totals are **116 text / 58 column / 1 severity = 175**, not 142. The reported 84/57/1 sums to 142 only because it is a *partition* that assigns each two-axis row to exactly one bucket. The partition is legitimate arithmetic, but the sub-split is a labelling choice and is not reproducible as axis counts; my column figure is 58, not 57. **The total (142), the endpoint (0), and the row/ID counts are all exact.**

Separately: the counts are **stable at `f7d9a5d3`, `7936c214` and `734eb94b`** (389/387, 387/387, 142 divergent at all three), so the PR body's "live baseline on `f7d9a5d3`" label is accurate, and the issue's older 385 / 83-58-1 figures were measured against a different tree.

## 4. Did both new IDs land in both catalogues? — **Yes, and the allocation is re-verified on a wider denominator**

- `GS9996` and `GS9997` were free at base across the **entire tree** — `git grep 'GS999[67]'` over all paths returns nothing. The allocation grep I was given was scoped to `src/ docs/ website/`, which excludes `test/`, `samples/`, `tools/` and `e2etests/`; widening it does not change the answer, but the narrower grep was not sufficient to establish it.
- Both IDs appear in **both** catalogues, byte-identical. `GS9998`'s duplicate was also collapsed (the issue flagged `GS9998` as duplicated too, and the PR body correctly explains both rows described the same emit failure, so no third ID was allocated).
- `GS9999` now survives in `src/` at exactly one place — `EvaluatorException.cs:78` (the brief said `:72`; line has drifted) — plus one comment. The removal set is precisely the six relocated sites and the four test assertions, with nothing left behind.
- Verified **end-to-end through the real `gsc` CLI**, each in its own directory asserted empty:
  - `gsc: error GS9997: Fatal compiler I/O error: Access to the path '…' is denied.` — rc **1**
  - `gsc: error GS9996: Source generator execution failed: /analyzer was supplied but gsgen was not found at '…'. Pass /gsgentool:<path> to override.` — rc **1**

  Both match their new documented rows. On the three-driver rule: these two diagnostics are emitted **only** by the `gsc` CLI process (the `Main` I/O wrapper and the gsgen launcher). `gsi` has no such path and does not reference either descriptor, so a three-column matrix is structurally inapplicable here rather than merely unrun — stating that plainly rather than presenting two columns as three.

## 5. Is `GS0153` fixed? — **Yes, and it now matches the code**

Descriptor (`DiagnosticDescriptors.cs:72`): *"…because it is neither an interface nor a class."*
Head, **identical in both catalogues**: *"Constraint is neither an interface nor a class. A generic type-parameter constraint must be an interface … or a base class; a value type such as a struct or enum is rejected."*
The base `docs/diagnostics.md` text ("Interface constraint is not sealed… must be `sealed`") is gone. The `version-0.3` snapshot already carried the correct text.

---

## Reconciliation direction — checked for silent information loss

A 142-row reconciliation can reach parity by deleting the richer side. It did not:

- 78 rows resolved to the `docs` text, 62 to the `website` text, 2 were link rewrites.
- **Zero rows ended with fewer columns than the richer of the two base rows.**
- Total row prose grew 49,109 → 54,854 characters (+11.7%).

The 2 "rewritten" rows (`GS0165`, `GS0166`) are the `docs` text with a relative `adr/0066-…md` link converted to an absolute GitHub URL — necessary, since a `docs/`-relative path cannot resolve from the website tree. Confined to exactly those two: **zero** relative `adr/` links remain inside any table row in either file (the 14 that remain in `docs/diagnostics.md` are in prose sections the gate does not compare, where relative links are correct), and the link-bearing rows are byte-identical across both files (md5 `329acb7c` on both).

## Per-hunk product mutants — 4 hunks, 4 killed, 0 survivors

| mutant | hunk | killed by |
|---|---|---|
| **MP1** | `Program.cs:251` `ReportFatalIOError` → back to `GS9999` | `Main_OutputPathInUnwritableLocation_ReportsErrorAndExitsNonZero`, `Main_OutputPathIsExistingDirectory_ReportsErrorAndExitsNonZero` |
| **MP2** | `Program.cs:714` `ReportGsgenFailure` → back to `GS9999` | `Analyzer_GsgenToolNotFound_ReturnsErrorAndDoesNotCompile` **and** `Every_Report_Uses_A_Descriptor_And_Every_Descriptor_Is_Used` |
| **MP3** | `DiagnosticDescriptors.cs` `GS9996`→`GS9995` | `Every_Diagnostic_Is_Documented_…` — `Missing from docs/diagnostics.md: GS9995 / Unexpected in docs/diagnostics.md: GS9996` |
| **MP4** | `Core.csproj` remove `InternalsVisibleTo Include="gsc"` | **pinned by compilation** — `CS0122` ×2 at `Program.cs:257,716` |

MP2's second kill is the interesting one: the new src-wide descriptor-usage scan turns an orphaned descriptor into a gate failure, so the two new IDs cannot silently rot. That is what earns the `InternalsVisibleTo` (see NIT-6).

The fourth gsgen site changes control flow (`WriteLine(...); … return false;` → `return ReportGsgenFailure(...)`). Equivalent: `ReportGsgenFailure` returns `false` unconditionally, and the `return false` below the `if` still covers the else path.

---

## Findings

### BLOCKER
None.

### SHOULD-FIX
None.

### NIT-1 — the PR body's validation is anchored on a tree that no longer exists
`d3b1d1c3` is absent from the repository and `9548f530` is orphaned; only `7936c214` survives, and it is an older `main`. The "Full Release suite … 14,639 passed" line is therefore unverifiable against this head. Recommend re-stating the validation section against `94e06880`. (I re-ran the suites myself, below, so this does not block.)

### NIT-2 — 84/57/1 is a partition, not axis counts
33 rows diverge on both column-count and text; independent axes give 116/58/1 = 175. Consider labelling the table "partitioned by primary axis" so the next reader does not try to reconcile it to the axis totals.

### NIT-3 — the gate still enforces parity, not truth
The gate compares file↔file and severity↔code, but **never compares message text to the descriptor**. `GS0153` was wrong for exactly that reason and was fixed here by hand; nothing added in this PR would catch the next one. I scanned all 371 rows that have a matching descriptor by token overlap against `MessageFormat` and found **no second `GS0153`-class contradiction**, so this is a durability gap rather than a live defect. I do not think there is a cheap fix — the message column is deliberately a terse paraphrase, so an automated comparison would be noisy — so this is worth an issue, not a change to this PR.

### NIT-4 — structural drift between the two catalogues remains unchecked
The two files now hold identical row content, but **256 of 389 rows sit at different indices**, and neither file is ID-sorted. The gate compares by ID, so the two documents can continue to diverge in sectioning and ordering while staying green. Given the website file "declares itself a mirror", enforcing order (or explicitly declaring order out of scope) would finish the job.

### NIT-5 — two out-of-`src` emitters now contradict the sharpened definitions
The PR narrows `GS9999` to "unexpected **evaluator** failures" and defines `GS9996` as "source generator execution failure". Neither is enforced outside `src/`:
- `tools/cs2gs` stamps `GS9999` as its own synthetic **compile-stage-crash** ID at `Cs2Gs.Pipeline/CompileStage.cs:160`, `SdkCompileRunner.cs:123,1103`, `MigrationPipeline.cs:601`.
- `src/Sdk/Gsharp.NET.Sdk/GsgenTask.cs:258` emits **`GS9998`** when gsgen exits non-zero without a structured diagnostic — the same class of failure `GS9996` now names, reported under a different ID depending on whether you go through `gsc /analyzer` or the SDK task.

Neither is a regression and neither is in this PR's stated scope, but they are newly *inconsistent* because this PR sharpened the definitions.

### NIT-6 — `InternalsVisibleTo Include="gsc"` is a real widening, and I think it is justified
The minimal fix was two changed string literals plus two entries in the gate's existing hardcoded directly-emitted map (which already carries `GS9998`/`GS9999`), with no new descriptors and no IVT. The author instead exposed `GSharp.Core` internals to the `gsc` assembly permanently. Recording it because it is permanent surface — but MP2 shows the chosen route buys a real pin: hardcoded ID strings would be invisible to the descriptor-usage scan, whereas descriptors are not. On balance the trade favours the author's version; no change requested.

### NIT-7 — pre-existing, carried forward
`Program.cs:254` states the format is chosen "so the SDK BuildTask's diagnostic regex surfaces it as a structured MSBuild error". `BuildTask.cs:33` requires a `file(line,col):` prefix, which `gsc: error GS9997: …` does not have. This was equally true of `GS9999` before the PR; only the digits changed. Flagging as pre-existing, not as a defect of this change.

---

## Suites — 9 test projects enumerated from `GSharp.sln`, 8 run

| project | result |
|---|---|
| `Core.Tests` | 7,113 passed, 1 skipped (7,114) |
| `Compiler.Tests` | **4,225 passed** |
| `Interpreter.Tests` | 1,203 passed |
| `LanguageServer.Tests` | 462 passed |
| `Extensions.Tests` | 104 passed |
| `InternalAnalyzers.Tests` | 7 passed |
| `Cs2Gs.Tests` | **1,903 passed** |
| `GSharp.GeneratorHost.Tests` | 31 passed |
| `Sdk.Tests` | **NOT RUN** — writes to the shared `~/.nuget/packages/gsharp.net.sdk/`, out of bounds with other agents active. Stating plainly rather than implying coverage. |

**Total: 15,048 passed, 1 skipped, 0 failed across 8 projects.**

All unfiltered, `--no-build` against a tree built with `BUILD_RC=0` and freshness confirmed by epoch-second mtimes (`sort -n`) on `Core/`, `Compiler/` and `Repl/`, excluding `ref/`.

**Oahu gate corpus**: `tools/cs2gs/external/oahu.json` pins `0ac1fece4415d31955ae7a1dcf7da31e343d363d`, unchanged by this PR — stated from the pin, which is the only thing that can establish it.

---

## `main` moved during this review — re-verified against it

`main` advanced `734eb94b` → **`f6f8acc0`** while I was measuring (#3133, #3122, #3146, plus `54c01e2e`). That matters here more than usual, because those commits touch **all three** surfaces this PR owns: `docs/diagnostics.md`, `website/docs/ref/diagnostics.md` **and** `DiagnosticDescriptors.cs` — and they were written under the *blind* gate, so a clean textual merge would not by itself mean the tightened gate accepts them.

Re-verified rather than assumed:

- `main` added one new descriptor, **`GS0518`** (explicit-layout residuals, #3122).
- Real merge of `pr3144` into `f6f8acc0` produces tree **`e50e420e`**, exactly reproducing `git merge-tree --write-tree`'s prediction.
- Census of the **merged** catalogues: **390 rows / 390 unique IDs in both files, zero divergent rows.** `GS0518` was documented identically in both catalogues, so it satisfies the new parity rule.
- Merged tree built Release with `ContinuousIntegrationBuild=true`: **`BUILD_RC=0`**; `Core.Tests` on the merged tree: **7,113 passed, 1 skipped, 0 failed.**

So the tightened gate accepts a concurrent, independently-authored diagnostic addition — which is the first real-world exercise of it, and it passes.

## Interaction with #3128 — flagged, clean today

#3128 (`fix/3114-readonlyspan-interpreted`, head `78d6442b`) edits **both** catalogues. `git merge-tree --write-tree` for #3144 × #3128 produces tree `78b73df6` with **no conflict** — they touch different rows. #3128 changes only the `GS0511` row, and it changes it **identically in both files**, so it satisfies the tightened gate as written. Whichever merges second simply needs to keep both files in step, which the new gate now enforces rather than leaves to review. No action needed; flagging as requested, not resolving.

## Disclosures

- **My truth-vs-parity scan was wrong on first run.** I stripped backtick-*delimited spans*, which nulls any docs cell wrapped entirely in backticks, and I had `type` in the stopword list — together producing a list of fake `0.00` overlaps. Corrected (strip backtick characters, drop the stopword) and re-run; the corrected output shows only legitimate terse paraphrases. The first output was an artifact of my harness, not a finding.
- **I briefly misread `Cs2Gs.Tests` as failing.** My `Failed: +[0-9]+` pattern matched xUnit *test-case display names* — `ACompletedTestRunIsRecognised(summary: "Failed! - Failed: 5, Passed: 0…")` — which is parameter data for a test that passed. Real xUnit failure lines: 0. A grep that can match display names is not a safe pass/fail signal.
- One probe (`GS9996` via `/analyzer` with no path) returned `/analyzer requires a path` — a **no-result**, not a result. Re-run with the correct `/analyzer:<file>` form before drawing the conclusion above.
- `docs/diagnostics.md:118` carries a stray semicolon inside the `GS0153` parenthetical (`generic or not;`), now mirrored into both files. Cosmetic only.

PR #3147, opened 2026-08-03
Fix evaluating driver assembly references

### LINKED ISSUE #3130: gsi and bare gsc cannot load Gsharp.Extensions: no reference channel, three samples fail behind an emit-only gate
## Summary

Real `gsi` (the driver process, `src/Repl/`) has **no reference channel**, so it cannot load `Gsharp.Extensions`. Any sample that depends on the extensions assembly fails under the interpreter, and the failure is invisible to CI because the sample conformance gate is emit-only.

Measured by two independent agents while reviewing PR #3077 (issue #2939).

## Evidence

`gsi --help` advertises the complete argument surface:

```
[file.gs] [--help] [--version]
```

There is no `/r:` or equivalent. By contrast `test/Compiler.Tests/LanguageConformance/SampleConformanceTests.cs:128` reaches the extensions assembly with an in-process `Assembly.Load("Gsharp.Extensions")` — a configuration **no user can reproduce from the command line**.

Three shipped samples are affected. Run as a real process with the same arguments the in-process harness uses:

```
gsc /nowarn:GS9100 samples/GsharpExtensionsOptional.gs   -> rc=1 GS0159
gsc /nowarn:GS9100 samples/GsharpExtensionsMixed.gs      -> rc=1 GS0157
gsc /nowarn:GS9100 samples/GsharpExtensionsSequences.gs  -> rc=1 GS0157
```

## This is not `gsi`-only

Bare `gsc` **evaluates** rather than emits, and it fails identically on all three samples. So this is not an interpreter-specific gap: it belongs to the shared evaluating path, and both evaluating drivers are affected. PR #3077 currently ledgers these rows as `InterpreterDifference`, which asserts bare `gsc` succeeds — that classification is wrong even though the ledger's prose reason is accurate.

## Why it survived

`SampleConformanceTests.cs:16`/`:30-58`/`:144` compiles via `gsc.dll` and runs the output with `dotnet` — **emit-only**. The evaluating columns, shared by bare `gsc` and `gsi`, are structurally invisible to the gate. This is the same blind spot that shipped #3100 and #3114. PR #3077 adds the interpreter column and is the right place for these rows to be ledgered against this issue.

## Ask

Decide and implement the reference channel for the evaluating drivers, then un-ledger the three samples. Either:

- give `gsi` and bare `gsc` a reference/assembly-probing option, or
- make the extensions assembly resolvable by default for the evaluating path the way it already is for emit.

If the answer is that samples must not depend on extensions, that is also an acceptable resolution — but it must be stated and the samples changed, not left failing behind an emit-only gate.

Ledger rows currently citing `"unfiled"` in PR #3077 should cite this issue.

DavidObando:
## Summary

- make bundled `Gsharp.Extensions` available by default to bare `gsc`, `gsc /out:`, and `gsi`
- add `gsi` reference switches matching `gsc` (`/r:`, `/reference:`, `-r:`, `--reference:`)
- validate explicit references consistently and isolate runtime references between invocations
- pass effective dependency paths into emit-to-memory execution
- package `Gsharp.Extensions.dll` with `gsi` and document the reference surface
- add real-process coverage for all three driver columns

## Supersession check

ADR-0156 Phase 1 (`3d5ba7af`, #3182) moved bare `gsc` and `gsi` script files onto the emitter. Phase 2 (`ed6ae55e`, #3186) added opt-in emitted interactive submissions. Neither change makes the bundled Extensions assembly an implicit driver reference.

Measured on freshly rebuilt current `main` `16f97229dc1cca01efc48aa66f234fc56e5ae551`:

| sample | bare `gsc` | `gsc /out:` | `gsi` |
|---|---|---|---|
| `GsharpExtensionsOptional.gs` | rc1 `GS0159` | rc1 `GS0159` | rc1 `GS0159` |
| `GsharpExtensionsMixed.gs` | rc1 `GS0157`, `GS0159` | rc1 `GS0157`, `GS0159` | rc1 `GS0157`, `GS0159` |
| `GsharpExtensionsSequences.gs` | rc1 `GS0157`, `GS0158`, `GS0159` | rc1 `GS0157`, `GS0158`, `GS0159` | rc1 `GS0157`, `GS0158`, `GS0159` |

The prior syntactic `ImportSyntax.StartsWith("Gsharp.Extensions")` gate no longer exists on `main`. Fully qualified no-import use fails uniformly on all three drivers with rc1 `GS0157`, so the narrower asymmetry is gone only because every column is still broken.

Conclusion: #3130 was closed prematurely; #3182/#3186 changed execution engines but did not fix bundled reference discovery. This PR remains a product fix, not tests-only salvage.

## Resolution

`ResolveDriverReferencePaths` appends the bundled assembly from the existing driver-sibling, solution-build, or SDK layouts. File drivers compile through the emitter, while the default interactive engine remains evaluator-based; both paths receive the same effective references. Explicit metadata references stay isolated, and emitted in-memory execution receives the corresponding runtime dependency paths.

On this head, all three shipped samples return rc0 and byte-match their golden files across bare `gsc`, emitted execution, and `gsi`. Fully qualified no-import use returns rc0 on all three and prints `11` (`gsc` also emits its existing `Success.` status line).

## Tests retained

`Issue3130DriverReferenceTests` keeps 11 durable regressions covering:

- all three shipped samples across all real drivers
- fully qualified Extensions use without an import
- explicit and invalid `/r:` behavior across all drivers
- bundled probe precedence and `gsi` packaging
- no unrelated `GS9100` warning pollution
- invokable runtime types and resolver isolation
- interactive evaluator reference injection

No workaround-only tests were split or dropped. No tests-only PR was opened because clean `main` still needs the product change.

## Validation

- head `2cfa21f9ca529aef8418121ee5058aeaf6fc42cc`
- base `16f97229dc1cca01efc48aa66f234fc56e5ae551`
- clean-main rebuild: 0 warnings, 0 errors; fresh nonzero `GSharp.Core.dll`
- final-head rebuild: 0 warnings, 0 errors; fresh nonzero `GSharp.Core.dll`
- exact final-head issue filter: 14/14
- exact final-head real-process matrix: three shipped samples plus no-import probe, all three columns rc0 with expected output
- unfiltered 9/9 solution projects, enumerated from `GSharp.sln`: **15,364 passed, 1 skipped, 0 failed**
  - Core 7,128 passed / 1 skipped
  - Compiler 4,260
  - Interpreter 1,404
  - LanguageServer 462
  - SDK 59
  - Extensions 104
  - Cs2Gs 1,907
  - GeneratorHost 31
  - InternalAnalyzers 9

The 9/9 local run preceded a final amend that only removed accidental reversions of two `main` test-name changes; final-head build and issue tests were rerun afterward.

## Product mutation evidence

Each mutant was asserted applied, built successfully, killed, restored, rebuilt, and returned green. No build-failing attempt was scored.

- restore syntactic import gate: killed by fully qualified no-import test
- invert bundled probe precedence: killed by layout test
- accept missing/junk reference: killed by bad-reference tests
- replace collectible runtime context with `Assembly.LoadFrom`: killed by isolation test
- invert compiler reference mode: killed by real-driver tests
- omit emit-to-memory runtime paths: killed by extension sample execution
- break `gsi` reference parsing/propagation: killed by explicit-reference tests
- drop interactive resolver injection: killed by `SessionEngine_UsesProvidedRuntimeReferences`
- use metadata-only implicit extension types: killed by runtime-type invocation test
- remove Repl Extensions dependency: killed by runtime manifest test

Fixes #3130

### DIFF
--- a/src/Compiler/Program.cs
+++ b/src/Compiler/Program.cs
@@ -115,9 +115,18 @@ public class Program
                 syntaxTrees.Add(SyntaxTree.Load(fullPath));
             }
 
-            var references = parsed.References.Count > 0
-                ? ReferenceResolver.WithReferences(parsed.References)
-                : null;
+            if (!ReferenceResolver.TryValidateDriverReferencePaths(parsed.References, out var referenceError))
+            {
+                Console.Error.WriteLine(referenceError);
+                return Error;
+            }
+
+            var referencePaths = ReferenceResolver.ResolveDriverReferencePaths(parsed.References);
+            var references = parsed.OutputPath is null
+                ? ReferenceResolver.WithRuntimeReferences(referencePaths)
+                : parsed.References.Count > 0
+                    ? ReferenceResolver.WithReferences(referencePaths)
+                    : ReferenceResolver.WithRuntimeReferences(referencePaths);
             ILogger logger = parsed.LogPath is not null ? new FileLogger(parsed.LogPath) : NullLogger.Instance;
 
             try
@@ -157,7 +166,7 @@ public class Program
                     // Legacy / no-output mode (ADR-0156 Phase 1): compile with
                     // the real emitter into memory and execute in-process,
                     // preserving the historical evaluate-mode driver protocol.
-                    return ExecuteInMemory(compilation, parsed);
+                    return ExecuteInMemory(compilation, parsed, referencePaths);
                 }
 
                 return Emit(compilation, parsed);
@@ -316,9 +325,12 @@ public class Program
     // driver protocol: diagnostics to stdout, "Success."/"Failed." trailer,
     // exit code 0 on success (the program's own return value is discarded,
     // as evaluate mode always did — use /out: for a real executable).
-    private static int ExecuteInMemory(Compilation compilation, CommandLineArgs args)
+    private static int ExecuteInMemory(
+        Compilation compilation,
+        CommandLineArgs args,
+        IReadOnlyList<string> referencePaths)
     {
-        var result = EmittedProgramHost.Run(compilation, args.References);
+        var result = EmittedProgramHost.Run(compilation, referencePaths);
         if (result.Diagnostics.Any())
         {
             var effective = ApplySuppressPromote(result.Diagnostics, args);
--- a/src/Core/CodeAnalysis/Symbols/ReferenceResolver.cs
+++ b/src/Core/CodeAnalysis/Symbols/ReferenceResolver.cs
@@ -14,6 +14,7 @@ using System.Reflection;
 using System.Reflection.Metadata;
 using System.Reflection.PortableExecutable;
 using System.Runtime.CompilerServices;
+using System.Runtime.Loader;
 using GSharp.Core.CodeAnalysis.Binding;
 using GSharp.Core.CodeAnalysis.Binding.OverloadResolution;
 
@@ -56,6 +57,7 @@ public sealed class ReferenceResolver : IDisposable
 
     private readonly ImmutableArray<Assembly> assemblies;
     private readonly MetadataLoadContext metadataContext;
+    private readonly AssemblyLoadContext runtimeContext;
     private readonly ImmutableArray<string> missingTransitiveReferences;
 
     // Process-wide registry of original on-disk paths for assemblies loaded
@@ -165,10 +167,15 @@ public sealed class ReferenceResolver : IDisposable
     {
     }
 
-    private ReferenceResolver(ImmutableArray<Assembly> assemblies, MetadataLoadContext metadataContext, ImmutableArray<string> missingTransitiveReferences)
+    private ReferenceResolver(
+        ImmutableArray<Assembly> assemblies,
+        MetadataLoadContext metadataContext,
+        ImmutableArray<string> missingTransitiveReferences,
+        AssemblyLoadContext runtimeContext = null)
     {
         this.assemblies = assemblies;
         this.metadataContext = metadataContext;
+        this.runtimeContext = runtimeContext;
         this.missingTransitiveReferences = missingTransitiveReferences;
         this.typeNameIndex = new Lazy<Dictionary<string, Type>>(
             this.BuildTypeNameIndex,
@@ -177,6 +184,36 @@ public sealed class ReferenceResolver : IDisposable
 
     private sealed class NotFoundSentinel
     {
+    }
+
+    private sealed class DriverReferenceLoadContext : AssemblyLoadContext
+    {
+        private readonly HashSet<string> directories = new(StringComparer.OrdinalIgnoreCase);
+
+        public DriverReferenceLoadContext()
+            : base(isCollectible: true)
+        {
+        }
+
+        public Assembly LoadReference(string path)
+        {
+            directories.Add(Path.GetDirectoryName(path) ?? string.Empty);
+            return LoadFromAssemblyPath(path);
+        }
+
+        protected override Assembly Load(AssemblyName assemblyName)
+        {
+            foreach (var directory in directories)
+            {
+                var path = Path.Combine(directory, assemblyName.Name + ".dll");
+                if (File.Exists(path))
+                {
+                    return LoadFromAssemblyPath(path);
+                }
+            }
+
+            return null;
+        }
     }
 
     /// <summary>
@@ -193,12 +230,13 @@ public sealed class ReferenceResolver : IDisposable
     /// </remarks>
     public void Dispose()
     {
-        if (metadataContext is null)
+        if (metadataContext is null && runtimeContext is null)
         {
             return;
         }
 
-        metadataContext.Dispose();
+        metadataContext?.Dispose();
+        runtimeContext?.Unload();
 
         // The static ImportedTypeSymbol cache may hold Type instances that
         // originated from the now-disposed MetadataLoadContext. Those entries
@@ -337,6 +375,109 @@ public sealed class ReferenceResolver : IDisposable
     public static ReferenceResolver Default()
     {
         return new ReferenceResolver(BuildHostAssemblies(), metadataContext: null);
+    }
+
+    /// <summary>
+    /// Resolves the references supplied to a driver and appends the bundled
+    /// Gsharp.Extensions assembly when it is available.
+    /// </summary>
+    /// <param name="referencePaths">Explicit reference paths.</param>
+    /// <returns>The effective reference paths.</returns>
+    public static IReadOnlyList<string> ResolveDriverReferencePaths(IEnumerable<string> referencePaths)
+    {
+        var paths = referencePaths?
+            .Where(path => !string.IsNullOrWhiteSpace(path))
+            .ToList()
+            ?? new List<string>();
+
+        var extensionPath = FindBundledExtensionPath(AppContext.BaseDirectory);
+        if (extensionPath != null
+            && !paths.Any(path => string.Equals(Path.GetFullPath(path), extensionPath, StringComparison.OrdinalIgnoreCase)))
+        {
+            paths.Add(extensionPath);
+        }
+
+        return paths;
+    }
+
+    /// <summary>
+    /// Validates explicit driver references before implicit bundled references
+    /// are appended.
+    /// </summary>
+    /// <param name="referencePaths">Explicit reference paths.</param>
+    /// <param name="error">A command-line error when validation fails.</param>
+    /// <returns><see langword="true"/> when every reference is a readable .NET assembly.</returns>
+    public static bool TryValidateDriverReferencePaths(IEnumerable<string> referencePaths, out string error)
+    {
+        foreach (var path in referencePaths)
+        {
+            if (string.IsNullOrWhiteSpace(path))
+            {
+                error = "/reference requires a path.";
+                return false;
+            }
+
+            try
+            {
+                var fullPath = Path.GetFullPath(path);
+                if (!File.Exists(fullPath))
+                {
+                    error = $"Unable to load reference '{path}': file does not exist.";
+                    return false;
+                }
+
+                _ = AssemblyName.GetAssemblyName(fullPath);
+            }
+            catch (BadImageFormatException)
+            {
+                error = $"Unable to load reference '{path}': file is not a valid .NET assembly.";
+                return false;
+            }
+            catch (Exception ex) when (ex is ArgumentException or IOException or NotSupportedException or UnauthorizedAccessException)
+            {
+                error = $"Unable to load reference '{path}': {ex.Message}";
+                return false;
+            }
+        }
+
+        error = string.Empty;
+        return true;
+    }
+
+    /// <summary>
+    /// Gets a resolver for evaluated code by loading explicit references into
+    /// the runtime context rather than a metadata-only context.
+    /// </summary>
+    /// <param name="referencePaths">Paths to runtime assemblies.</param>
+    /// <returns>A resolver over the loaded assemblies and host BCL.</returns>
+    public static ReferenceResolver WithRuntimeReferences(IEnumerable<string> referencePaths)
+    {
+        var runtimeContext = new DriverReferenceLoadContext();
+        var runtimeAssemblies = new List<Assembly>();
+        if (referencePaths is not null)
+        {
+            foreach (var path in referencePaths.Where(path => !string.IsNullOrWhiteSpace(path)))
+            {
+                var fullPath = Path.GetFullPath(path);
+                var assembly = runtimeContext.LoadReference(fullPath);
+                RegisterAssemblyOriginalPath(assembly, fullPath);
+                runtimeAssemblies.Add(assembly);
+            }
+        }
+
+        var runtimeNames = runtimeAssemblies
+            .Select(assembly => assembly.GetName().FullName)
+            .ToHashSet(StringComparer.OrdinalIgnoreCase);
+        var assemblies = BuildHostAssemblies()
+            .Where(assembly => !runtimeNames.Contains(assembly.GetName().FullName))
+            .Concat(runtimeAssemblies)
+            .ToImmutableArray();
+
+        return new ReferenceResolver(
+            assemblies,
+            metadataContext: null,
+            missingTransitiveReferences: ImmutableArray<string>.Empty,
+            runtimeContext: runtimeContext);
     }
 
     /// <summary>
@@ -824,6 +965,22 @@ public sealed class ReferenceResolver : IDisposable
         return true;
     }
 
+    internal static string FindBundledExtensionPath(string baseDirectory)
+    {
+        // Compiler cannot ProjectReference Gsharp.Extensions because its bootstrap
+        // build already references Compiler; solution and SDK layouts provide it.
+        var candidates = new[]
+        {
+            Path.Combine(baseDirectory, "Gsharp.Extensions.dll"),
+            Path.Combine(baseDirectory, "..", "Gsharp.Extensions", "Gsharp.Extensions.dll"),
+            Path.Combine(baseDirectory, "..", "extensions", "Gsharp.Extensions.dll"),
+        };
+
+        return candidates
+            .Select(Path.GetFullPath)
+            .FirstOrDefault(File.Exists);
+    }
+
     // Raw resolution shared by the public/internal TryResolveType overloads: it
     // performs the reference-set name lookup (warm/cold index + cache) without
     // applying the external-visibility gate, which the callers layer on top.
@@ -1066,7 +1223,7 @@ public sealed class ReferenceResolver : IDisposable
 
         foreach (var asm in AppDomain.CurrentDomain.GetAssemblies())
         {
-            if (asm.IsDynamic)
+            if (asm.IsDynamic || AssemblyLoadContext.GetLoadContext(asm) is DriverReferenceLoadContext)
             {
                 continue;
             }
--- a/src/Repl/Engine/SessionEngine.cs
+++ b/src/Repl/Engine/SessionEngine.cs
@@ -26,7 +26,15 @@ public sealed class SessionEngine : ISessionEngine
 {
     private readonly Dictionary<VariableSymbol, object> variables = new();
     private readonly List<Cell> cells = new();
+    private readonly ReferenceResolver? references;
     private Compilation? previous;
+
+    /// <summary>Initializes a new instance of the <see cref="SessionEngine"/> class.</summary>
+    /// <param name="references">Runtime references available to every cell.</param>
+    public SessionEngine(ReferenceResolver? references = null)
+    {
+        this.references = references;
+    }
 
     public IReadOnlyList<Cell> Cells => cells;
 
@@ -216,7 +224,9 @@ public sealed class SessionEngine : ISessionEngine
             try
             {
                 var tree = SyntaxTree.Parse(SourceText.From(text, fileName));
-                var compilation = previous == null ? new Compilation(tree) : previous.ContinueWith(tree);
+                var compilation = previous == null
+                    ? references is null ? new Compilation(tree) : new Compilation(references, tree)
+                    : previous.ContinueWith(tree);
                 var result = compilation.Evaluate(variables, evaluateEntryPoint: RunEntryPoint);
 
                 var hasError = result.Diagnostics.Any(d => d.IsError);
--- a/src/Repl/Program.cs
+++ b/src/Repl/Program.cs
@@ -33,7 +33,8 @@ public static class Program
             case "/?":
                 Console.WriteLine("Usage: gsi [file.gs] [/r:<assembly> ...] [--engine <name>] [--help] [--version]");
                 Console.WriteLine("  file.gs                Run the given G# script and exit.");
-                Console.WriteLine("  /r:<file>              Reference an assembly (alias: /reference:<file>; repeatable).");
+                Console.WriteLine("  /r:<file>              Reference an assembly (repeatable).");
+                Console.WriteLine("  /reference:<file>      Alias for /r: (also -r: and --reference:).");
                 Console.WriteLine("  --engine <name>        Interactive engine: 'evaluator' (default) or 'emit'");
                 Console.WriteLine("                         (ADR-0156 Phase 2 opt-in; also via GSI_ENGINE).");
                 Console.WriteLine("  --help, -h             Show this help and exit.");
@@ -86,6 +87,12 @@ public static class Program
             return 1;
         }
 
+        if (!ReferenceResolver.TryValidateDriverReferencePaths(references, out var referenceError))
+        {
+            Console.Error.WriteLine(referenceError);
+            return 1;
+        }
+
         if (scriptPath is null)
         {
             return RunInteractive(engineChoice, references);
@@ -116,18 +123,14 @@ public static class Program
             return 1;
         }
 
+        var effectiveReferences = ReferenceResolver.ResolveDriverReferencePaths(references);
         if (engineChoice == "emit")
         {
-            return ReplHost.Run(new Engine.EmittedSessionEngine(references));
+            return ReplHost.Run(new Engine.EmittedSessionEngine(effectiveReferences));
         }
 
-        if (references.Count > 0)
-        {
-            Console.Error.WriteLine("/r: references in interactive mode require --engine emit (ADR-0156 Phase 2).");
-            return 1;
-        }
-
-        return ReplHost.Run();
+        using var resolver = ReferenceResolver.WithRuntimeReferences(effectiveReferences);
+        return ReplHost.Run(new Engine.SessionEngine(resolver));
     }
 
     /// <summary>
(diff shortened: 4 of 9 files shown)


DavidObando:
# Review — PR #3147 (`fix/3130-evaluator-extensions-reference`), head `0f6f7950`

**Verdict: MERGEABLE** — 0 BLOCKER, 2 SHOULD-FIX, 4 NIT.

The core of this change is correct and it fixes a genuinely severe bug. Every product hunk is pinned by a mutant. The two SHOULD-FIX items are about where the fix *stops*, not about what it does.

## Anchors

- Live main `f6f8acc0`; merge base `2dea578b`; branch 1 ahead / 2 behind.
- `git merge-tree` **output**: 1 line, **0 conflict markers**, predicted tree `b2667891` — reproduced exactly in the merged worktree.
- Because a clean merge-tree is a floor and not a proof, I checked file overlap directly: main's two new commits (`09462e58`, `f6f8acc0`) share **no file** with this PR.
- `BUILD_RC=0`, 0 warnings, on both base and merged. mtimes taken in epoch seconds and sorted numerically across `Core/`, `Compiler/`, `Repl/`, excluding `ref/` — all fresh.

## Claims verified

**The runtime-vs-metadata split is necessary, not a preference.** On base, passing *any* `/r:` to bare `gsc` breaks evaluation outright:

```
base:   gsc /r:<anything> p.gs   -> rc=1  error GS9999: Cannot invoke a method
                                          on objects loaded by a MetadataLoadContext.
merged: gsc /r:<valid>    p.gs   -> rc=0
```

**Real-process tests are real.** `Issue3130DriverReferenceTests.cs:172` starts `dotnet` with `gscPath` / `gsiPath` resolved from `typeof(GSharp.Compiler.Program).Assembly.Location`, redirects both pipes, 30 s timeout, `Kill(entireProcessTree: true)`. All **3 tests executed** (filter matched 3, none skipped). They use emit as the oracle and compare against `samples/*.golden`.

**Packaging is substantiated by the artifact, not the csproj.** I packed the tool and read the nupkg:

```
tools/net10.0/any/Gsharp.Extensions.dll
tools/net10.0/any/gsi.dll
```

The extensions assembly ships **sibling to `gsi.dll`**, which is exactly probe candidate 1. `PackAsTool=true`, `ToolCommandName=gsi` (`Repl.csproj:16-18`).

**Mutation — 5 hunks, 5 mutants, 5 kills.** Each asserted applied (unique pattern, site count = 1), `BUILD_RC=0` before every test run, filter matched 3 every time.

| mutant | hunk | result |
|---|---|---|
| M-A invert `FirstOrDefault(File.Exists)` | `ReferenceResolver.cs` auto-probe | **RED** — `ExtensionSamples_MatchEmitAcrossRealDrivers` |
| M-B2 `WithRuntimeReferences`→`WithReferences` | `Compiler/Program.cs` split | **RED** — both driver tests |
| M-C invert `arg.StartsWith(prefix)` | `Repl/Program.cs` `TryParseReference` | **RED** — both driver tests |
| M-D invert `File.Exists(path)` guard | `ReferenceResolver.WithRuntimeReferences` | **RED** — both driver tests |
| M-E delete `ProjectReference` | `Repl.csproj` packaging | **RED** — `GsiRuntimeManifest_IncludesExtensionsAssembly` |

A first attempt, M-B (inverting `parsed.OutputPath is null` outright), **broke the build** — `Gsharp.Extensions.dll` is itself compiled by `gsc`, so flipping emit onto the runtime path produced a malformed assembly and `CS0012` in `Extensions.Tests`. That is a **no-result, not a survival**, so I retargeted it as M-B2. Recording it because it shows this hunk sits on the bootstrap path.

Both mutant rounds restored: `git diff -- src/` empty, `git status --porcelain` empty, rebuilt, 3/3 green.

**Suites — 7 of 9 projects (enumerated from `GSharp.sln`), 10,837 executed, 0 failures, 1 skipped:** Core 7113/1 skip · Interpreter 1217 · Cs2Gs **1903** · LanguageServer 462 · Extensions 104 · GeneratorHost 31 · InternalAnalyzers 7.
`Cs2Gs.Tests` ran clean with no `NETSDK1004` (the solution build preceded it). **`Compiler.Tests` NOT RUN** (~1 h locally) and **`Sdk.Tests` NOT RUN** (local SDK package, out of bounds) — stated plainly rather than implied. Mitigating structural argument for `Compiler.Tests`: when `parsed.OutputPath is not null` the new expression reduces to `parsed.References.Count > 0 ? WithReferences(parsed.References) : null`, which is byte-for-byte the pre-PR behaviour, so the emit path is unchanged.

**I looked for a silent driver divergence and did not find one.** Probed: nonexistent `/r:` path, `/r:` at a real file that is not an assembly, and import of an extensions namespace with no reference — across all three drivers, with and without `/nowarn:GS9100`. Every failure mode I could construct is loud. Stating the scope so the zero is not mistaken for absence.

---

## SHOULD-FIX 1 — the fix stops at the oracle: `gsc /out:` still cannot see the bundled extensions

`src/Compiler/Program.cs:117-121` applies `ResolveDriverReferencePaths` only when `parsed.OutputPath is null`. Measured on `samples/GsharpExtensionsOptional.gs` with no `/r:`:

| driver | base | merged |
|---|---|---|
| bare `gsc` (evaluate) | rc=1 `GS0159` | **rc=0** |
| `gsi` (interpret) | rc=1 `GS0159` | **rc=0** |
| `gsc /out:` (**emit — the oracle**) | rc=1 `GS0159` | **rc=1 `GS0159`** |

Base was uniformly broken; the merged tree is uniformly *fixed except on the oracle column*. A user can run an extensions program under `gsi` and bare `gsc` and then fail to compile the identical file with `gsc /out:`. The PR's own test encodes the asymmetry rather than surfacing it — `Issue3130DriverReferenceTests.cs:44-51` passes `/r:` + `extensionsPath` to the emit column only.

This also under-delivers the issue: the maintainer's comment on #3130 says the extensions assembly "should also always be included by default even if not provided in the `/r:` list", with `gsc` unqualified.

**I measured the remedy rather than asserting it.** Replacing the emit branch with

```csharp
: ReferenceResolver.WithReferences(ReferenceResolver.ResolveDriverReferencePaths(parsed.References));
```

gives: solution `BUILD_RC=0` (the bootstrap that compiles `Gsharp.Extensions.dll` with `gsc` **survives** — the self-reference hazard did not materialise), the sample emits with **no `/r:`** at rc=0, the emitted program's stdout is **byte-identical to `GsharpExtensionsOptional.golden`**, and Interpreter **1217/1217** plus Core **7113/1** stay green. Not validated against `Compiler.Tests`, which is the suite most likely to notice; that check is owed before adopting.

Not a blocker: the failure is loud, and it is not a regression — base failed this column too.

## SHOULD-FIX 2 — a bad `/r:` is silently ignored on every driver

`ReferenceResolver.WithRuntimeReferences` filters on `File.Exists(path)` and then swallows `FileNotFoundException`, `FileLoadException` and `BadImageFormatException` in three empty `catch` blocks. Measured, merged tree:

| `/r:` argument | bare `gsc` | `gsi` | `gsc /out:` |
|---|---|---|---|
| path that does not exist | rc=0, no diagnostic | rc=0, no diagnostic | rc=0, no diagnostic |
| real file, not an assembly | rc=0, no diagnostic | rc=0, no diagnostic | rc=0, no diagnostic |

The silence is symmetric and pre-existing on the emit path, so this is not a divergence. But this PR **creates a new user-facing `/r:` surface on `gsi`** that inherits it, and the three empty catches make it structural rather than incidental. A user who mistypes a path gets a downstream `GS0159 Cannot find function` pointing at their own source rather than "could not load reference X". Recommend reporting a diagnostic (or at minimum a warning) when an explicitly-supplied `/r:` path fails to resolve or load; the auto-probed bundled path should stay silent when absent.

## NIT 1 — one of the three probe layouts is exercised by no test, and precedence is unpinned

`ReferenceResolver.ResolveDriverReferencePaths` tries three candidates. I measured which one each driver actually hits:

| candidate | consumer | present locally |
|---|---|---|
| `{base}/Gsharp.Extensions.dll` | `gsi` (local `Repl/`, and `tools/net10.0/any/` in the nupkg) | yes |
| `{base}/../Gsharp.Extensions/` | bare `gsc` local build (`out/bin/Release/Gsharp.Extensions/`) | yes |
| `{base}/../extensions/` | `gsc` in the SDK package | **no** |

Candidate 3 is real — `Gsharp.NET.Sdk.csproj:178` packs extensions to `tools\extensions\` and `:111` packs gsc to `tools\compiler\`. *(Self-correction: I initially suspected candidates 2 and 3 were invented; reading the packaging refutes that, and I withdraw it.)* But it only fires in the shipped SDK configuration, which no runnable suite covers here. Separately, `gsi` has candidates 1 **and** 2 present simultaneously in a local build and takes 1 by `FirstOrDefault` ordering; nothing pins that precedence, so a future layout change could silently swap which assembly loads. Note also that `Compiler.csproj` declares **no** dependency on `Gsharp.Extensions` — bare `gsc` finds it purely because the solution build happens to place a sibling directory, so a Compiler-only build silently loses the feature.

## NIT 2 — `references` is a write-only local

`src/Repl/Program.cs:57` declares `using var references = ReferenceResolver.WithRuntimeReferences(...)` and never reads it; the returned resolver is not handed to `SessionEngine`. The feature works entirely through the `Assembly.LoadFrom` **side effect** inside `WithRuntimeReferences`. M-C and M-D prove it is pinned, so a break would be caught — but the shape invites a "remove unused variable" cleanup that would silently disable `gsi /r:`. A comment, or actually consuming the resolver, would make the dependency legible.

## NIT 3 — `gsi` argument-parsing behaviour changes beyond what is documented

`--help` / `--version` are now matched at any position in the argument list, so `gsi prog.gs --help` prints help and exits 0 instead of running the script (measured rc=0). Base only inspected `args[0]`. Also `TryParseReference` accepts `-r:` and `--reference:`, which appear in neither the `--help` text nor `website/docs/tooling/repl.md:38`.

## NIT 4 — the process helper reads pipes after `WaitForExit`

`Issue3130DriverReferenceTests.cs:186-197` calls `WaitForExit(30_000)` and only then `StandardOutput.ReadToEnd()` / `StandardError.ReadToEnd()`. That is the classic deadlock ordering: a child that fills a pipe buffer blocks forever, the wait times out, and the test reports a timeout rather than the real output. Bounded here by small outputs, and it fails loudly via `Kill(entireProcessTree: true)` rather than hanging, so it is a nit. Starting both `ReadToEndAsync()` calls before `WaitForExit` removes the hazard. Same shape exists in `Issue3010EntryPointDriverMatrixTests`.

DavidObando:
# Round 2 — head `18b2016c`, `main` `61fe9cb9`

**Verdict: NOT MERGEABLE — 1 SHOULD-FIX (repeat, narrowed but alive).** SHOULD-FIX 2 is fully fixed and well pinned. SHOULD-FIX 1 fixed the *reported* symptom and removed the test crutch, but **the driver asymmetry itself survives** on a smaller surface, and I have a reproducer plus a remedy measured at zero cost.

## Anchors

Branch is **0 behind** `61fe9cb9`, 6 ahead, merge base == main — so **head is the merged tree**. `git merge-tree` output: one tree SHA `4f3cbc5c`, **0 conflict markers** (read the output, not the rc). `BUILD_RC=0`, 0 warnings. Artifacts fresh in epoch seconds across `Core/`, `Compiler/`, `Repl/` (`ref/` excluded); oldest first-party 12:43, newest 12:44 — older entries are NuGet copies (Newtonsoft 2021, VS.Validation 2023). Every probe in its own directory, asserted empty.

**What is actually new since round 1.** Round-1 head `0f6f7950` was a single commit, patch-id `ace34257…`. The current head's first commit `e9253aaa` carries the **identical patch-id**, so the original work is untouched and **five commits are genuinely new**. The waiver deletion round 1 found as a dirty worktree is now committed as `30362206`.

---

## SHOULD-FIX 2 — **FIXED**

`ReferenceResolver.TryValidateDriverReferencePaths` (`:371-405`) validates eagerly, before implicit references are appended, and `WithRuntimeReferences` (`:414`) no longer filters on `File.Exists` nor swallows. Round 1's table was rc=0/no diagnostic in all six cells. Now:

| `/r:` argument | bare `gsc` | `gsi` | `gsc /out:` |
|---|---|---|---|
| path does not exist | **rc=1** `…file does not exist.` | **rc=1** same | **rc=1** same |
| real file, not an assembly | **rc=1** `…not a valid .NET assembly.` | **rc=1** same | **rc=1** same |

Uniform across all three drivers, with the offending path named. Pinned by two new tests.

**On "gone or merely narrowed":** the three empty catches at `ReferenceResolver.cs:1194-1202` still exist, but they are **not** on the user `/r:` path — they guard `Assembly.Load` of `WellKnownBclAssemblyNames` (`System.Runtime`, `System.Console`, `System.Private.CoreLib`, **`mscorlib`**). Swallowing there is correct; `mscorlib` does not exist on modern .NET. Round 1 attributed these to `WithRuntimeReferences`; that attribution was imprecise. The user-facing swallowing is genuinely gone.

---

## SHOULD-FIX 1 — symptom fixed, **asymmetry not closed**

**The good half, verified.** On `samples/GsharpExtensionsOptional.gs` with **no `/r:`**:

| driver | rc | output |
|---|---|---|
| bare `gsc` (evaluate) | **0** | golden + `gsc`'s own `Success.` line |
| `gsi` (interpret) | **0** | **byte-identical to golden** |
| `gsc /out:` (**emit — oracle**) | **0** | **byte-identical to golden** |

*(The emitted program aborts with `FileNotFoundException` if `Gsharp.Extensions.dll` is not deployed beside it — but my control shows an explicit `/r:` build behaves **identically**, so that is standard .NET deployment, **not** a defect of this PR.)*

**And the test crutch is gone.** `Issue3130DriverReferenceTests.cs:44-51` now emits with `/nowarn:GS9100 /target:exe /targetframework:net10.0 /out:` and **no `/r:`**; `extensionsPath` survives only as a `File.Copy` for *execution*. All three columns get the same reference inputs. That was the right correction.

### The asymmetry that remains

The emit path is gated on a **syntactic** test at `src/Compiler/Program.cs:123-130`: emit only resolves bundled extensions when some `ImportSyntax` has a dotted name `StartsWith("Gsharp.Extensions")`. Bare `gsc` and `gsi` resolve **unconditionally**. Any program that reaches the extensions without such an import statement therefore still runs on two drivers and fails to compile on the third.

```gs
package p
import System

let xs = Gsharp.Extensions.Sequences.Sequences.Of(11, 22, 33)
Console.WriteLine(xs[0])
```

| driver | rc | result |
|---|---|---|
| bare `gsc` | **0** | prints `11` |
| `gsi` | **0** | prints `11` |
| `gsc /out:` | **1** | `error GS0157: Cannot find type Gsharp. Are you missing an import?` |

**Two controls prove the gate is the cause, not a binding limitation:**

- same file, emit **with** explicit `/r:…/Gsharp.Extensions.dll` → **rc=0**, runs, prints `11`
- same logic **with** `import Gsharp.Extensions.Sequences` → emit **rc=0** with no `/r:`

So the shape binds and emits perfectly; the reference is simply not added. This is the same property you sent round 1 back for — *run it with `gsi`, then fail to compile it* — narrowed from "every extensions program" to "extensions programs without a prefix-matching import". Narrower is **harder to discover**, not safer. None of the 8 tests covers it; every emit test uses an explicit `import Gsharp.Extensions.*`.

### The remedy, measured rather than asserted

Forcing `importsBundledExtensions` true (`Program.cs:123`, rebuilt, `BUILD_RC=0`):

- the fully-qualified probe above emits at **rc=0**
- a plain `import System` / `Console.WriteLine(42)` emit stays **rc=0 with zero `GS9100`** — measured directly, not just via the test
- `UnrelatedEmit_DoesNotAcquireImplicitExtensionWarnings` **passes**
- **Interpreter.Tests 1228/1228**, identical to baseline

So the warning pollution the gate exists to prevent **does not reappear** when the gate is removed, at least as far as these suites see. Restored afterwards: `git diff` empty, rebuilt, and the original failure reproduced exactly (`emit_rc=1`). **Not validated against `Compiler.Tests`** — the suite most likely to notice — and that check is owed before adopting.

---

## Mutation — 2 hunks, 2 kills, both two-sided and asserted applied

| mutant | hunk | result |
|---|---|---|
| **M1** invert emit gate → `\|\| !importsBundledExtensions` (`Program.cs:128`; 1 new pattern, 0 old) | SHOULD-FIX 1 | **RED** — `ExtensionSamples_MatchEmitAcrossRealDrivers` |
| **M2′** missing-file rejection returns `true` (`ReferenceResolver.cs:387`) | SHOULD-FIX 2 | **RED** — `BadReferences_AreRejectedAcrossRealDrivers` |

Filter asserted non-vacuous every run (**Total: 8**, never 0). Two disclosed near-misses:

- **M2 (first attempt)** inverted `!File.Exists` in the validator and **broke the build** — the bootstrap compiles `Gsharp.Extensions` with `gsc`, so every valid reference was rejected. That is a **no-result, not a survival**; retargeted to M2′.
- **M2′ initially did not apply** — I targeted line 386, but `return false;` is line **387**. The build went green on an unmutated tree. Only the applied-assertion caught it; without it I would have recorded a false survival.

---

## Credit

- `BundledExtensionProbe_PrefersDriverSiblingThenBuildThenSdkLayout` now pins the three-candidate precedence that round 1's NIT 1 flagged as unpinned.
- `BundledExtensionEmit_UsesInvokableRuntimeTypes` runs the *emitted* program, which justifies the deliberate use of `WithRuntimeReferences` on the emit path.
- Test count 3 → 8, and the emit column lost its `/r:` crutch.

## Suites

**2 of 9 projects** (denominator from `dotnet sln GSharp.sln list`): **Interpreter 1228/1228**, **Core 7116 passed / 1 skipped / 7117**. `Compiler.Tests`, `Extensions.Tests`, `InternalAnalyzers.Tests`, `LanguageServer.Tests`, `Sdk.Tests`, `Cs2Gs.Tests`, `GSharp.GeneratorHost.Tests` **NOT RUN** — stated plainly rather than implied.

## One line to close it

The gate is a syntactic proxy for "does this program use the extensions". It is sound only by enumeration of the ways to reach them, and fully-qualified access is one it misses. Either drop the gate (measured above: free) or make the proxy semantic — retry on unresolved-name, or scan bound names rather than `ImportSyntax`.

DavidObando:
Re-review update at `18b2016ccf4d8be50bb5b657598bdf287a61848c`:

- Emit oracle without `/r:`: base `rc=1` (`GS0159`) → head compile `rc=0`, run `rc=0`, output byte-identical to golden. Extension-using emits retain invokable runtime types; unrelated emits receive no implicit extension warnings.
- Missing and junk `/r:` inputs now return `rc=1` with explicit diagnostics on bare `gsc`, emit, and `gsi`.
- Clean-shell three-driver matrix passed from separate initially empty directories; bare `gsc` and `gsi` matched emitted output.
- Product mutation campaign: 11 scored mutants killed/restored green; one bootstrap no-result retargeted and not scored.
- Local unfiltered denominator: **9 of 9 projects enumerated from `GSharp.sln`**, 15,143 executed, 15,142 passed, 1 skipped. `Sdk.Tests` was corrected from the earlier assumption and ran 59/59.
- Current CI, including all test partitions, e2e, corpus, and Oahu, is green.

Full matrices and counts are in the updated PR body.

DavidObando:
Final rebase addendum: head is now `0e7766861a3b16f8cd1c18a6ded22e4cf358ac80` on current `origin/main`. The PR patch is unchanged. On this head: solution build 0 warnings/errors, issue tests 8/8, `Sdk.Tests` 59/59, prior channel regressions 16/16, clean-shell driver matrix green, and all six invalid-reference cells green. CI is rerunning for the rebased SHA.

DavidObando:
SHA correction: the final head is `0e7766861a3b129a699100a17025471bf179c64f`. All 30 required checks are green on that exact SHA.

DavidObando:
Round 2 update on `d2b574d085ea556fb65dcd2ef6312595ad3a4c00`:

| no-import source | before | after |
|---|---:|---:|
| bare `gsc` | rc=0 | rc=0 |
| `gsc /out:` | rc=1 `GS0157` | rc=0 |
| `gsi` | rc=0 | rc=0 |

Removed the syntactic import gate and added `FullyQualifiedExtensionUseWithoutImport_MatchesEmitAcrossRealDrivers`. Plain `import System` emit remains rc=0 with zero `GS9100`.

Current-main ADR-0156 changed both evaluating drivers to emitted execution during the rebase. The first merge resolution exposed four Interpreter failures: explicit references were being loaded into the default context, while implicit Extensions forced `gsi` away from its normal host resolver. Reference mode now keys on explicit-reference presence: implicit bundled Extensions use runtime assemblies; explicit `/r:` uses metadata plus emitted-host runtime paths. Full Interpreter then passed 1,346/1,346.

Unfiltered 9/9 solution run: 15,292 executed, 15,291 passed, 1 skipped; Compiler.Tests 4,245/4,245 and Sdk.Tests 59/59. Final base-only rebase: solution build 0 warnings/errors, issue 9/9 plus six related regressions green.

Round-2 product mutants: old syntactic gate RED; omitted implicit-reference control RED; compiler reference-mode inversion RED; Repl reference-mode inversion RED. Each applied at exactly one asserted site and restored green.

Near-misses: one initial full run hit unrelated `MSB3026` in a generated `Lib.csproj`; isolated retry and later full runs were green. One mutation command printed load 30.62 immediately before its separate load gate sampled below 30 and allowed the build; subsequent builds started below 30.

DavidObando:
Final integration at `dca5dc410a19b085ef68ea063a48af7d8d283792` on `81e6deab4ad19f2f642182691c42146dd63ddb27`.

Resolved post-review/main interactions:

- bundled extensions resolve for fully-qualified use and every emit
- emit-to-memory hosts receive effective runtime dependency paths
- explicit runtime references use a collectible load context and do not leak between invocations
- `SessionEngine` consumes its supplied resolver
- Repl and Interpreter package locks include the new Extensions dependency

Final evidence:

- merge-tree rc 0, tree `d5c1cbee87837d720fac6e767723522904fc0a2a`
- merged-tree rebuild `BUILD_RC=0`, 0 warnings/errors; Core/Compiler/Repl artifacts fresh
- real-process corpus: bare `gsc` 119/122 → 122/122; `gsi` 119/122 → 122/122; exactly three extension samples changed in each column
- issue tests 11/11; sample conformance 125/125
- unfiltered 9/9 projects: 15,364 passed, 1 skipped, 0 failed
- required-parameter counts: `ResolveDriverReferencePaths` 3 sites; `WithRuntimeReferences` 9 sites
- isolation, propagation, effective-dependency, fully-qualified, validation, packaging, and layout product mutants all killed and restored green

PR body now carries full quantitative matrix and per-hunk mutation table.

DavidObando:
Supersession check complete: **do not close this PR**.

Freshly rebuilt `main` at `16f97229dc1cca01efc48aa66f234fc56e5ae551` still fails all three issue samples without `/r:` on every file driver:

| sample | bare `gsc` | `gsc /out:` | `gsi` |
|---|---|---|---|
| Optional | rc1 `GS0159` | rc1 `GS0159` | rc1 `GS0159` |
| Mixed | rc1 `GS0157`, `GS0159` | rc1 `GS0157`, `GS0159` | rc1 `GS0157`, `GS0159` |
| Sequences | rc1 `GS0157`, `GS0158`, `GS0159` | rc1 `GS0157`, `GS0158`, `GS0159` | rc1 `GS0157`, `GS0158`, `GS0159` |

ADR-0156 Phase 1 (`3d5ba7af`, #3182) changed file execution to emit-to-memory, and Phase 2 (`ed6ae55e`, #3186) added opt-in emitted interactive submissions. Neither supplies the implicit bundled Extensions reference. The issue was therefore closed prematurely.

The old syntactic import gate is absent on current `main`. A fully qualified no-import probe fails uniformly with rc1 `GS0157` on all three drivers; this PR makes all three rc0 and prints `11`.

Rebased head: `2cfa21f9ca529aef8418121ee5058aeaf6fc42cc`. Build is clean, issue filter is 14/14, and the exact real-process matrix is green. Product fix remains necessary, so no tests-only salvage PR was opened; the 11 durable regression tests stay here.

DavidObando:
# Round 3 — head `2cfa21f9`, independently measured against clean `main` `917677ec`

**Verdict: MERGEABLE. 0 BLOCKER, 0 SHOULD-FIX, 2 NIT.** The round-2 repeat SHOULD-FIX is **closed**, not narrowed. Everything below was measured in two dedicated worktrees, each with its own `out/` root (verified not a symlink).

---

## Lead finding — **#3130 still reproduces on clean `main`. The issue is prematurely closed.**

I did not take the author's table on trust. Fresh worktree at `917677ec`, full Release build (`BUILD_RC=0`, `GSharp.Core.dll` written 68 s before the probes). All three drivers, all three samples:

| sample | bare `gsc` | `gsc /out:` | `gsi` |
|---|---|---|---|
| `GsharpExtensionsOptional` | rc1 `GS0159` | rc1 `GS0159` | rc1 `GS0159` |
| `GsharpExtensionsMixed` | rc1 `GS0157,GS0159` | rc1 `GS0157,GS0159` | rc1 `GS0157,GS0159` |
| `GsharpExtensionsSequences` | rc1 `GS0157,GS0158,GS0159` | rc1 `GS0157,GS0158,GS0159` | rc1 `GS0157,GS0158,GS0159` |

**Codes read individually, not lumped** — `GS0157` (missing type), `GS0158`, `GS0159` (missing function) are distinct and match the author's table cell-for-cell. `gsc` writes diagnostics to **stdout** and only `Failed.` to stderr; a probe reading stderr alone would have seen nothing useful.

**Control:** `samples/AddressBook.gs` on the same build → all three drivers rc0, emitted binary output **byte-identical to golden**. So the harness works and the failures are specific. *(My first control named `Hello.gs`, which does not exist — that was a no-result, not a pass, and I re-ran it.)*

**Independent conclusion: the author is right.** #3182/#3186 changed engines, not implicit references. `ed6ae55e` is an ancestor of `main`, yet the defect is fully live. **This PR is not superseded**, and issue #3130 is closed as COMPLETED against a reproducing defect.

---

## 1. Three samples on the PR head — pass on printed values, not rc

| sample | bare `gsc` | `gsc /out:` | `gsi` | **emitted binary vs golden** |
|---|---|---|---|---|
| Optional | rc0 | rc0 | rc0 | ✅ **byte-identical** |
| Mixed | rc0 | rc0 | rc0 | ✅ **byte-identical** |
| Sequences | rc0 | rc0 | rc0 | ✅ **byte-identical** |

All three driver columns also match golden as printed text (bare `gsc` after its own trailing `Success.` banner, which the tests correctly encode).

## 2. The syntactic gate — **removed, and the asymmetry with it**

`git grep 'StartsWith("Gsharp.Extensions'` over the tree returns **nothing**; `importsBundledExtensions` no longer exists. `Program.cs:122` now calls `ResolveDriverReferencePaths` **unconditionally**, so there is no syntax-shaped decision left to probe with alias or lookalike imports — the fragile instrument is gone rather than hardened.

My round-2 reproducer, two-sided:

```gs
let values = Gsharp.Extensions.Sequences.Sequences.Of(11, 22, 33)
Console.WriteLine(values[0])
```

| | bare `gsc` | `gsc /out:` | `gsi` | emitted run |
|---|---|---|---|---|
| clean `main` | rc1 `GS0157` | rc1 `GS0157` | rc1 `GS0157` | — |
| **PR head** | **rc0 `11`** | **rc0** | **rc0 `11`** | **rc0 prints `11`** |

This is the fix I measured as free in round 2, adopted in full.

## 3. The 11 tests — strong oracles, and I will name which are durable

11 `[Fact]`s, matching the claim. The two load-bearing ones assert `Assert.Equal(expected, Normalize(output))` against **full golden text per driver** — not a coarse shared constant, so **no ADR-0154 specimen inflation**. `FullyQualifiedExtensionUse_MatchesEmitAcrossRealDrivers:85` passes **no `/nowarn` to any column**, so the emit crutch removed in round 2 has not returned in another form.

On *"which survive a reimplementation"* — **6 behavioural, 5 mechanism-coupled**:

- **Durable (black-box, spawn real `gsc`/`gsi`, compare printed output):** `:28`, `:85`, `:139`, `:218`, `:243`, `:290`
- **Mechanism-coupled (die if the mechanism changes):** `:210` greps `gsi.deps.json`; `:336` pins three-candidate probe precedence; `:365`/`:384` pin `ReferenceResolver` API behaviour; `:407` pins the `SessionEngine` ctor

The mechanism-coupled five are legitimate — `:336` is what pins round-2's NIT-1 — but they are **change-detectors, not behaviour guarantees**. The six behavioural ones carry the actual proof.

## 4. Emit-everywhere coverage — structural enumeration

Every driver-side compilation/emit entry now receives resolved references:

`Compiler/Program.cs:137` · `Compiler/Program.cs:333` (was `args.References`) · `Repl/Program.cs:129`, `:133`, `:148`, `:152` · `Repl/Engine/EmittedSessionEngine.cs:291` · `Repl/Engine/SessionEngine.cs:228`

`gsi`'s stdin refusal lives at `Repl/Program.cs:120-125`; file mode is `RunScript` at `:146`, which is what my probes and the tests exercise — so the probes test what their names say.

## Mutation — 4 hunks, 4 kills, one an explicit control

Baseline **11/11 green** first. Each mutant asserted applied by `git diff` showing a **real statement** (never a brace) *and* within-tree dll md5 movement; restored by **exact md5 round-trip**.

|---|---|---|---|
| **M1** *(control)* | `ReferenceResolver.cs:394` | `extensionPath != null` → `== null` | **RED** ×2 |
| **M2** | `Program.cs:129` | `WithRuntimeReferences(referencePaths)` → `null` | **RED** ×2 |
| **M3** | `ReferenceResolver.cs:426` | missing-file `return false` → `return true` | **RED** — `BadReferences_…` |
| **M4** | `Repl/Program.cs:146` | `ResolveDriverReferencePaths(references)` → `references` | **RED** ×2 |

`GSharp.Core.dll` returned to `0f3189db…` and `gsc.dll` to `2c9fa48f…` after restore — byte-exact, so the witness is sound.

## Suites — **9 of 9**, source `dotnet sln GSharp.sln list`

Compiler **4260** · Core **7128 (+1 skipped)** · **Cs2Gs 1907** · Interpreter **1404** · LanguageServer **462** · Extensions **104** · Sdk **59** · InternalAnalyzers **9** · GeneratorHost **31** — **15,364 passed, 1 skipped, 0 failed.** `Cs2Gs.Tests` matters here because it drives real `gsc.dll` out-of-process; it is green. CI **31 SUCCESS / 3 SKIPPED / 0 pending**, `ilverify` present and **SUCCESS**.

---

## NIT 1 — the asymmetry may have moved to the IDE *(static, not executed)*

`ProjectState.cs:273-280,345,452` pass a resolver only when one exists; the loose-file paths `DocumentSyncHandler.cs:56,82` and `HoverComputer.cs:28…1548` build `new Compilation(syntaxTree)` with **no resolver**, and `Compilation.cs:303` passes `references: null`. Before this PR the editor and the drivers agreed — both failed. Now the drivers succeed, so an editor opening a loose `.gs` using bundled extensions would show `GS0157`/`GS0159` squiggles on code that compiles and runs. **I did not execute the language server**, so this is a reasoned read, not a measurement — worth a follow-up probe rather than a change here.

## NIT 2 — `WithRuntimeReferences` remains unguarded

`ReferenceResolver.cs:168` calls `Assembly.LoadFrom` bare. `TryValidateDriverReferencePaths` uses `AssemblyName.GetAssemblyName`, so a file that passes the name check but fails to load (architecture mismatch → `FileLoadException`) surfaces as an unhandled crash rather than a diagnostic. Unprobed; narrow.

## Two of my own probes were wrong, and the controls caught them

My first item-4 probes emitted `GS0125 "Variable 'shared' doesn't exist"` — **invalid G#, my harness bug, not a PR defect**; clean `main` produced the same codes. And `import Gsharp.ExtensionsBogus` compiles rc0 on **both** `main` and the PR, so that leniency is **pre-existing** and not this PR's doing. Neither is scored against the author.

PR #3148, opened 2026-08-03
[gsc] Preserve generic call metadata across async spilling

### LINKED ISSUE #3095: [gsc] LINQ over IReadOnlyList<T> emits mismatched Func<object,bool> call-site IL
## Reproduction

```csharp
sealed record Notice(bool Active);

static async Task<List<Notice>> GetActiveAsync(IStore store)
{
    var items = await store.ListAsync(); // IReadOnlyList<Notice>
    return items.Where(x => x.Active).ToList();
}
```

The method compiles, but ILVerify reports a stack type mismatch at the LINQ call:

```text
found:    System.Func<Notice, bool>
expected: System.Func<object, bool>
```

The delegate is correctly typed, but the selected/emitted `Where` call is instantiated with `object`.

## Expected behavior

Generic extension inference and emitted method specifications must agree on `TSource = Notice`. The call should target `Enumerable.Where<Notice>` and accept `Func<Notice,bool>`.

## Fix verification

Add a compile/run/ILVerify test using an awaited `IReadOnlyList<T>`, then `Where(...).ToList()`. Assert filtering works and ILVerify reports no stack errors. Include a non-async local-list control and an interface-returned list to isolate inference from async lowering.

DavidObando:
## Summary
- preserve imported static/instance generic arguments, symbolic container data, and constrained-dispatch metadata when async lowering rebuilds calls around nested awaits
- preserve user-instance method type arguments and constrained-dispatch metadata through async spilling, bound-tree rewriting, and defer capture
- add compile/run/ILVerify coverage for awaited `IReadOnlyList<T>` LINQ plus local, interface, covariance, nested-generic, constrained-user-call, and generic-user-call controls

## Root cause
Binding was correct: extension inference recovered `TSource = Notice`, target-typed the lambda as `Func<Notice, bool>`, and stored `Notice` in `BoundImportedCallExpression.TypeArgumentSymbols`. The reflected `MethodInfo` remained the expected object-erased placeholder used during binding.

When `Where` received a direct `await` expression as its extension receiver, `SpillSequenceSpiller` rebuilt the imported call after lifting the await. That rebuild copied only the function, arguments, and ref kinds, dropping `TypeArgumentSymbols`. Emission therefore reconstructed the `MethodSpec` from the placeholder `MethodInfo` and wrote `Enumerable.Where<object>` / `ToList<object>` beside the correctly typed `Func<Notice, bool>`.

Review of every sibling call reconstruction found the same metadata-loss class on user-instance calls: async spilling, `BoundTreeRewriter`, and defer capture omitted bound method type arguments, and spilling also omitted constrained receiver/interface data. Those shared rebuild paths now replace only receivers/arguments while retaining every emission-relevant property.

No LINQ special case, object fallback, inference change, or verifier suppression is involved.

## Verification
- reduced repro runs correctly, passes ILVerify, and emits `Enumerable.Where<Notice>` / `Enumerable.ToList<Notice>`
- awaited-local, non-async local-list, and interface-returned-list controls pass
- covariance and nested generic regressions pass
- constrained and generic user-instance async/defer regressions pass
- full `Core.Tests`: **7116 passed, 1 skipped, 0 failed**
- full `Compiler.Tests`: **4232 passed, 0 failed**
- `dotnet build GSharp.sln -c Release --no-restore`: **0 warnings, 0 errors**
- GitHub Actions: **29 checks succeeded**, 2 publish-only checks skipped as expected; pinned Oahu migration **15/15 apps green**

Fixes #3095

### DIFF
--- a/src/Core/CodeAnalysis/Binding/BoundTreeRewriter.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundTreeRewriter.cs
@@ -1904,15 +1904,22 @@ public abstract class BoundTreeRewriter
             return node;
         }
 
-        // Issue #502 (sub-bug 502-a/b): preserve the call's return-type override
-        // (e.g. the Task / Task[T] lift applied by BindUserInstanceCall for async
-        // members). Dropping it here caused the rewritten call's static type to
-        // collapse to the bare method-declared type (e.g. int32 instead of
-        // Task[int32]), which then mis-typed the receiver of `GetAwaiter()` and
-        // produced an invalid box at the call boundary — leading to a hang
-        // because the inner async member's task never visibly completed from
-        // the caller's perspective.
-        return new BoundUserInstanceCallExpression(null, receiver, node.Method, builder?.ToImmutable() ?? node.Arguments, node.Type, node.ConstrainedReceiverTypeParameter, node.ConstrainedInterfaceType);
+        // Preserve every call-site property while replacing only receiver and
+        // arguments. The return-type override carries async Task lifting (#502);
+        // constrained metadata controls generic-interface dispatch (#1052); and
+        // method type arguments are the authoritative bind-time substitution
+        // when emit-time structural re-inference is impossible (#1931).
+        return new BoundUserInstanceCallExpression(
+            null,
+            receiver,
+            node.Method,
+            builder?.ToImmutable() ?? node.Arguments,
+            node.Type,
+            node.ConstrainedReceiverTypeParameter,
+            node.ConstrainedInterfaceType)
+        {
+            MethodTypeArguments = node.MethodTypeArguments,
+        };
     }
 
     /// <summary>Rewrites an explicit-base interface call (ADR-0091).</summary>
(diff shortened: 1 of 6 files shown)


DavidObando:
Follow-up pushed in `f051db3a5baf40f630db9ab7799bdf08e0bbdafa`.

Fixed user-instance metadata preservation across async spill, bound-tree rewrite, and defer capture. Base repros failed with the verified constrained-call GS9998 and generic type-inference GS9998; fixed repros compile, ILVerify, and run. Full Core: 7116 passed/1 skipped. Full Compiler: 4232 passed. Release build: 0 warnings/errors.

DavidObando:
CI green on `f051db3a5baf40f630db9ab7799bdf08e0bbdafa`: 29 checks passed, 2 publish checks skipped. Initial `cs2gs-oahu` attempt failed in pinned Oahu's original C# `SignInFlow_Start_Sets_State` test before migration; failed-job rerun passed.

PR #3151, opened 2026-08-03
Preserve field shape on reified CLR type arguments

### LINKED ISSUE #3137: Reified CLR type arguments have no members under gsi and bare gsc: GetFields() returns 0 where emit returns 2
## Summary

PR #3125 (merged) fixed #3099 — `gsi` and bare `gsc` no longer erase G# struct and enum type arguments to `System.Object` when reifying CLR types. The type **name** is now correct on all three drivers.

But the reified type's **member shape** is still empty. A rubber-duck measured, on the merged tree:

```
GetFields().Length   evaluating drivers: 0     emit: 2
```

rc 0, no diagnostic, silent — the same shape as the bug that was just fixed.

## Why this is filed rather than folded into #3125

It **pre-exists** #3125 and is identical on its base (in fact worse there, since the type name was wrong too), so it is not a regression and was correctly graded SHOULD-FIX rather than a blocker. #3125 merged on that basis.

It is filed because of what the PR body claims. The body says the fix produces "**true** CLR value types", and a type with zero fields where the oracle has two is not that. Leaving the claim unqualified risks #3099 being remembered as fully closed when one axis of reification is still open.

## The structural lesson, which is the reason to fix it properly

#3125's test is a genuine **parity oracle** — it computes the expected value from the emit column and compares, rather than freezing strings, so new type kinds cannot silently pass. That design is right and should be kept.

But the oracle compares **type identity only**. It never compares **member shape**. So this divergence is invisible to it for exactly the same structural reason that #3099 itself was invisible to #3049's parity test: the axis was never generated. Author and three reviewers all stayed on the covered axis, twice in a row, on the same code.

**Extending the oracle to compare member shape is worth more than the individual fix.** Do that first; the failing cell will then be produced by the harness rather than by hand.

## Bar

Three driver columns: bare `gsc` **evaluates**, `gsc /out:` **emits**, `gsi` **interprets**. Emit is the oracle — compare evaluating output against emitted output cell by cell rather than against hardcoded expectations.

Cover `{class, struct, enum, string, nested class, nested struct, generic instantiation, nullable, array}` × `{single type arg, multiple, nested generic}`, the axis set #3125 already established, and add the member-shape comparison to every cell.

**A zero can be scope, not absence** — `GetFields().Length == 0` is precisely this bug's own symptom, so positive-control that your reflection call would report a non-zero count if one existed.

Use `Assembly.Load` + `GetTypes()`; `IlVerifier.Verify` is not sufficient, because corrupt generic metadata surfaces at load or first call rather than at emit. That matters especially here, since this issue is about metadata shape.

Keep probes top-level and distinct-valued. Everything that agrees between the drivers today must still agree.

DavidObando:
## Summary

- extend the existing emit-parity oracle to compare reflected field count, names, CLR types, readonly state, static state, literal state, and constant values across every type-argument cell
- define G# instance, static, and const fields on evaluator-created CLR backing types with matching reflection metadata
- preserve generic field types recursively through nullable types, imported generics, slices, and arrays
- use a G#-defined two-instance-field reflection control, asserted independently in all three drivers

## Coverage

The oracle covers 11 type kinds across single, multiple, and nested generic placements under bare `gsc`, emitted `gsc /out:`, and `gsi`: 99 parity cells. Emitted output is the expected value. The generic specimen includes `T`, `List[T]`, `[]T`, `T?`, `[3]T`, a static field, and a const field.

| Generic field | Emit | Evaluate | Interpret |
|---|---|---|---|
| `Value T` | `System.Int32` | `System.Int32` | `System.Int32` |
| `Imported List[T]` | `List[Int32]` | `List[Int32]` | `List[Int32]` |
| `Slice []T` | `System.Int32[]` | `System.Int32[]` | `System.Int32[]` |
| `Nullable T?` | `System.Int32` | `System.Int32` | `System.Int32` |

| Field | Emit | Evaluate | Interpret |
|---|---|---|---|
| static `Stat` | `IsStatic=True, IsLiteral=False` | same | same |
| const `Kilo` | `IsStatic=True, IsLiteral=True, Value=1000` | same | same |

## Mutation testing

| Mutant | Result |
|---|---|
| invert readonly field handling | RED: 5 failed / 6 passed |
| invert generic type-parameter map lookup | RED: 1 failed / 10 passed |
| omit nullable recursive map argument | RED: 1 failed / 10 passed; `Nullable` became `System.Object` |
| invert static-field condition | RED: 11 failed / 0 passed |
| invert const `SetConstant` condition | RED: 11 failed / 0 passed with `GS9999` |
| scope reflection control to public static fields | RED: 11 failed / 0 passed |

Every mutation was asserted present before execution. Restored head is GREEN at 11/11. Main without this PR is GREEN at 11/11 on its unmodified oracle.

## Validation

- rebased onto main `61fe9cb943d15c0f5ba747dabd562e66fb132058`
- `git merge-tree --write-tree` output exactly matches the head tree
- solution build: rc 0, 0 warnings, 0 errors
- targeted parity oracle: 11 passed
- local tests: 15,134 passed, 1 skipped, 0 failed across 9 of 9 test projects enumerated from `GSharp.sln`
- correction addendum: `Sdk.Tests` runs in-process and is safe locally; it passed 59/59. The earlier 8-of-9 statement came from mistaking a docstring description for executed behavior.
- all required PR checks, including `Sdk.Tests`, are green
- emitted assemblies are loaded with `Assembly.Load`, enumerated with `GetTypes()`, and executed

Validated on head `8cd87cf48ef0788b8a3463ce4a0d799379b79b3f`.

Fixes #3137

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -1283,6 +1283,7 @@ public sealed partial class Evaluator
         var module = CreateClrBackingModule();
         var definition = structType.Definition ?? structType;
         var arity = definition.TypeParameters.Length;
+        var reifiedTypeParameters = new Dictionary<TypeParameterSymbol, Type>();
         var enclosingTypes = new Stack<StructSymbol>();
         for (var containing = definition.ContainingType as StructSymbol;
              containing != null;
@@ -1331,7 +1332,27 @@ public sealed partial class Evaluator
                 clrBase);
         if (arity > 0)
         {
-            typeBuilder.DefineGenericParameters(definition.TypeParameters.Select(static parameter => parameter.Name).ToArray());
+            var parameters = typeBuilder.DefineGenericParameters(
+                definition.TypeParameters.Select(static parameter => parameter.Name).ToArray());
+            for (var i = 0; i < parameters.Length; i++)
+            {
+                reifiedTypeParameters[definition.TypeParameters[i]] = parameters[i];
+            }
+        }
+
+        // Issue #3137: reified type arguments must expose the same instance-field shape as emitted types.
+        foreach (var field in definition.Fields)
+        {
+            var fieldAttributes = Emit.AccessibilityMap.MapFieldAccessibility(field.Accessibility);
+            if (field.IsReadOnly)
+            {
+                fieldAttributes |= FieldAttributes.InitOnly;
+            }
+
+            typeBuilder.DefineField(
+                field.Name,
+                ResolveClrBackingTypeArgument(field.Type, reifiedTypeParameters),
+                fieldAttributes);
         }
 
         if (emitBaseConstructors)
@@ -1380,8 +1401,16 @@ public sealed partial class Evaluator
         return backingType.MakeGenericType(typeArguments);
     }
 
-    private static Type ResolveClrBackingTypeArgument(TypeSymbol argument)
+    private static Type ResolveClrBackingTypeArgument(
+        TypeSymbol argument,
+        IReadOnlyDictionary<TypeParameterSymbol, Type> reifiedTypeParameters = null)
     {
+        if (argument is TypeParameterSymbol typeParameter
+            && reifiedTypeParameters?.TryGetValue(typeParameter, out var reifiedTypeParameter) == true)
+        {
+            return reifiedTypeParameter;
+        }
+
         if (argument is NullableTypeSymbol nullable)
         {
             var underlying = ResolveClrBackingTypeArgument(nullable.UnderlyingType);
@@ -1397,7 +1426,9 @@ public sealed partial class Evaluator
             try
             {
                 return openDefinition.MakeGenericType(
-                    imported.TypeArguments.Select(ResolveClrBackingTypeArgument).ToArray());
+                    imported.TypeArguments
+                        .Select(typeArgument => ResolveClrBackingTypeArgument(typeArgument, reifiedTypeParameters))
+                        .ToArray());
             }
             catch (ArgumentException)
             {
@@ -1407,12 +1438,12 @@ public sealed partial class Evaluator
 
         if (argument is SliceTypeSymbol slice)
         {
-            return ResolveClrBackingTypeArgument(slice.ElementType).MakeArrayType();
+            return ResolveClrBackingTypeArgument(slice.ElementType, reifiedTypeParameters).MakeArrayType();
         }
 
         if (argument is ArrayTypeSymbol array)
         {
-            return ResolveClrBackingTypeArgument(array.ElementType).MakeArrayType();
+            return ResolveClrBackingTypeArgument(array.ElementType, reifiedTypeParameters).MakeArrayType();
         }
 
         if (argument.ClrType is Type clrType)
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -1340,19 +1340,35 @@ public sealed partial class Evaluator
             }
         }
 
-        // Issue #3137: reified type arguments must expose the same instance-field shape as emitted types.
-        foreach (var field in definition.Fields)
+        // Issue #3137: reified type arguments must expose the same field shape as emitted types.
+        foreach (var field in definition.Fields.Concat(definition.StaticFields).Concat(definition.ConstFields))
         {
             var fieldAttributes = Emit.AccessibilityMap.MapFieldAccessibility(field.Accessibility);
-            if (field.IsReadOnly)
+            if (field.IsConst)
             {
-                fieldAttributes |= FieldAttributes.InitOnly;
+                fieldAttributes |= FieldAttributes.Static | FieldAttributes.Literal | FieldAttributes.HasDefault;
+            }
+            else
+            {
+                if (field.IsStatic)
+                {
+                    fieldAttributes |= FieldAttributes.Static;
+                }
+
+                if (field.IsReadOnly)
+                {
+                    fieldAttributes |= FieldAttributes.InitOnly;
+                }
             }
 
-            typeBuilder.DefineField(
+            var fieldBuilder = typeBuilder.DefineField(
                 field.Name,
                 ResolveClrBackingTypeArgument(field.Type, reifiedTypeParameters),
                 fieldAttributes);
+            if (field.IsConst)
+            {
+                fieldBuilder.SetConstant(field.ConstantValue);
+            }
         }
 
         if (emitBaseConstructors)
@@ -1413,7 +1429,7 @@ public sealed partial class Evaluator
 
         if (argument is NullableTypeSymbol nullable)
         {
-            var underlying = ResolveClrBackingTypeArgument(nullable.UnderlyingType);
+            var underlying = ResolveClrBackingTypeArgument(nullable.UnderlyingType, reifiedTypeParameters);
             return underlying.IsValueType
                 ? typeof(Nullable<>).MakeGenericType(underlying)
                 : underlying;
--- a/test/Interpreter.Tests/Issue3099TypeArgumentReificationTests.cs
+++ b/test/Interpreter.Tests/Issue3099TypeArgumentReificationTests.cs
@@ -29,7 +29,7 @@ public class Issue3099TypeArgumentReificationTests
         yield return ["string", string.Empty, "string"];
         yield return ["nested-class", "class Owner {\n    class Payload {\n        let Eleven int32\n        var TwentyTwo string\n    }\n}", "Owner.Payload"];
         yield return ["nested-struct", "class Owner {\n    struct Payload {\n        let Eleven int32\n        var TwentyTwo string\n    }\n}", "Owner.Payload"];
-        yield return ["gsharp-generic", "struct Payload[T] {\n    let Value T\n    var Imported List[T]\n    var Slice []T\n    var Array [3]T\n}", "Payload[int32]"];
+        yield return ["gsharp-generic", "struct Payload[T] {\n    let Value T\n    var Imported List[T]\n    var Slice []T\n    var Nullable T?\n    var Array [3]T\n    shared {\n        var Stat int32\n    }\n    const Kilo int32 = 1000\n}", "Payload[int32]"];
         yield return ["imported-generic", "struct Payload {\n    let Eleven int32\n    var TwentyTwo string\n}", "List[Payload]"];
         yield return ["nullable", "struct Payload {\n    let Eleven int32\n    var TwentyTwo string\n}", "Payload?"];
         yield return ["slice", "struct Payload {\n    let Eleven int32\n    var TwentyTwo string\n}", "[]Payload"];
@@ -90,9 +90,15 @@ public class Issue3099TypeArgumentReificationTests
             });
 
             var expected = NormalizeGenericTypeNames(emitted);
-            Assert.Matches(@"44\n[1-9][0-9]*\|", expected);
-            Assert.Equal(expected, NormalizeGenericTypeNames(compilerEvaluation));
-            Assert.Equal(expected, NormalizeGenericTypeNames(interpreter));
+            var evaluated = NormalizeGenericTypeNames(compilerEvaluation);
+            var interpreted = NormalizeGenericTypeNames(interpreter);
+            const string ControlShape =
+                "44\n2|Eleven:System.Int32:True:False:False|TwentyTwo:System.String:False:False:False\n";
+            Assert.Contains(ControlShape, expected, StringComparison.Ordinal);
+            Assert.Contains(ControlShape, evaluated, StringComparison.Ordinal);
+            Assert.Contains(ControlShape, interpreted, StringComparison.Ordinal);
+            Assert.Equal(expected, evaluated);
+            Assert.Equal(expected, interpreted);
             Assert.Contains("11\n", expected);
             Assert.Contains("22\n", expected);
             Assert.Contains("33\n", expected);
@@ -115,6 +121,11 @@ public class Issue3099TypeArgumentReificationTests
 
             __DECLARATION__
 
+            struct ReflectionControl {
+                let Eleven int32
+                var TwentyTwo string
+            }
+
             class Box[T] : EventArgs {
             }
 
@@ -124,7 +135,10 @@ public class Issue3099TypeArgumentReificationTests
             func FieldShape(t Type) string {
                 var shape = t.GetFields().Length.ToString()
                 for field in t.GetFields() {
-                    shape = shape + "|" + field.Name + ":" + field.FieldType.FullName + ":" + field.IsInitOnly.ToString()
+                    shape = shape + "|" + field.Name + ":" + field.FieldType.FullName + ":" + field.IsInitOnly.ToString() + ":" + field.IsStatic.ToString() + ":" + field.IsLiteral.ToString()
+                    if field.IsLiteral {
+                        shape = shape + ":" + field.GetRawConstantValue().ToString()
+                    }
                 }
                 return shape
             }
@@ -147,7 +161,7 @@ public class Issue3099TypeArgumentReificationTests
             var nested = Box[Box[__TYPE__]]()
             Console.WriteLine(nested.GetType().FullName)
             Console.WriteLine(FieldShape(nested.GetType().GenericTypeArguments[0].GenericTypeArguments[0]))
-            var control = Box[ConsoleColor]()
+            var control = Box[ReflectionControl]()
             Console.WriteLine(44)
             Console.WriteLine(FieldShape(control.GetType().GenericTypeArguments[0]))
             """;
--- a/test/Interpreter.Tests/Issue3099TypeArgumentReificationTests.cs
+++ b/test/Interpreter.Tests/Issue3099TypeArgumentReificationTests.cs
@@ -16,23 +16,24 @@ namespace GSharp.Interpreter.Tests;
 /// <summary>
 /// Issue #3099: interpreter-created generic backing types must preserve every
 /// G# type-argument kind instead of erasing value types to <see cref="object"/>.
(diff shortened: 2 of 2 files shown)


DavidObando:
## NOT MERGEABLE — the recursion fix misses one of its four branches, and the divergence is still live on this head

Anchors: `origin/main` **`f6f8acc0`**, head **`96e98594`**, merge base `734eb94b` (branch is **behind** main). `git merge-tree` output = tree `80752858`, **0 conflict markers**; co-merging reproduced that tree hash exactly. Merged-tree build from a worktree with **no `out/`**: `BUILD_RC=0`, **0 errors, 0 warnings**. Targeted oracle baseline **11/11**. Freshness by epoch mtime, `sort -n`, excluding `ref/`: `Core` 1785767212, `Compiler` 1785767214, `Repl` 1785767216 — all first-party artifacts within 4 s, ~5 min old.

**Your mutation numbers are honest and I reproduced them exactly.** Inverting `field.IsReadOnly` at `Evaluator.Expressions.cs:1347` → **5 failed / 6 passed**, dying cases `class`, `struct`, `nested-class`, `nested-struct`, `gsharp-generic` — identical to the PR body, case for case. Reverting the `DefineField` loop at `:1344` → **5 failed**, matching your "reproduced in five user-defined type cases". Restoring returned **11/11**. That part of the report is solid, and it is rarer than it should be.

---

### BLOCKER 1 — `NullableTypeSymbol` is the one recursive branch that was not threaded

`ResolveClrBackingTypeArgument` has **four** recursive branches. Three were updated to pass `reifiedTypeParameters`; one was not:

| line | branch | threaded? |
|---|---|---|
| **`Evaluator.Expressions.cs:1416`** | `NullableTypeSymbol` | **NO** |
| `:1430` | imported generic | yes |
| `:1441` | `SliceTypeSymbol` | yes |
| `:1446` | `ArrayTypeSymbol` | yes |

The PR summary says "recursively through imported generics, slices, and arrays" — naming exactly the three that were fixed. The fourth loses the type-parameter mapping and falls through to `return typeof(object)` at `:1465`.

Measured with **real out-of-process drivers** on this head, `struct Payload[T] { let Value T; var Imported List[T]; var Slice []T; var Opt T? }` used as `Box[Payload[int32]]`:

| column | `Opt` field type |
|---|---|
| **`gsc /out:` — EMIT, the oracle** | `System.Int32` |
| bare `gsc` — EVALUATE | **`System.Object`** |
| `gsi` — INTERPRET | **`System.Object`** |

`Value`, `Imported` and `Slice` are all correct — only the nullable one is wrong. **This is #3137's exact symptom, unfixed, on the PR that closes #3137.**

**Fix is one argument**, mirroring its three siblings:
```csharp
var underlying = ResolveClrBackingTypeArgument(nullable.UnderlyingType, reifiedTypeParameters);
```
I applied exactly that (1 insertion / 1 deletion, asserted): all three columns then print `Opt:System.Int32`, and the targeted oracle stays **11/11**. Nothing larger is needed.

The oracle is **11/11 green both with and without this bug**, so it does not cover the cell. `gsharp-generic` now exercises `T`, `List[T]`, `[]T`, `[3]T` — but not `T?`; and the `nullable` case tests nullable as a *type argument* (`Payload?`), never as a *field type wrapping a type parameter*. That intersection is the broken one.

### SHOULD-FIX 2 — static and const fields still diverge, same defect class

`definition.Fields` at `:1344` is **instance-only**; `StructSymbol.StaticFields` and `ConstFields` are separate collections that emit handles at `TypeDefEmitter.cs:355` and `:387` (`| FieldAttributes.Static`, `| Static | Literal | HasDefault`). The new loop defines neither. Real drivers, this head:

| source | EMIT (oracle) | EVALUATE / INTERPRET |
|---|---|---|
| `shared { var Stat int32 }` | `3\|Eleven:False\|TwentyTwo:False\|**Stat:True**` | `2\|Eleven:False\|TwentyTwo:False` |
| `shared { const Kilo int32 = 1000 }` | `2\|Eleven:False\|**Kilo:True**` | `1\|Eleven:False` |

Public static and const fields are visible to `GetFields()`, so this is the same user-visible divergence #3137 reports, on a field category the PR neither fixes nor specimens. If you'd rather hold PR scope to instance fields that is defensible — but it then needs a filed follow-up, because nothing in the tree records it.

### SHOULD-FIX 3 — the `ConsoleColor` positive control is decorative, and I could not make it fail

The claim is *"positive-control reflection with `ConsoleColor` so zero-field results cannot pass vacuously."* Three measurements say it does not do that.

1. **It never fires.** Under the reverted-fix mutant, all 5 failures were `Assert.Equal() Failure: Strings differ` — the parity assertion. **Zero** came from `Assert.Matches` at `:93`.
2. **It is invariant under the product.** With the bug fully present, the control line is byte-identical across all three drivers: `17|value__:False|Black:True|…`. `ConsoleColor` is an **imported** enum; its shape is CoreLib metadata and cannot move when evaluator backing-type construction changes. Meanwhile the user type read `0` (evaluate) vs `2` (emit).
3. **It does not close the vacuity hole it names.** 16 of its 17 fields are **static literals**; the user types' fields are **instance**. Under a plausible `BindingFlags` mis-scoping of `FieldShape` (`:125-126`) to `Public | Static`, I measured the control print `16|Black|…` — which *matches* `44\n[1-9][0-9]*\|` and passes — while the user type prints `0`, identically in all three columns, so parity holds and the whole oracle goes **green having measured nothing about instance-field shape**.

The regex only demands "count > 0" on a type the product cannot affect. **Fix:** anchor the control to a *G#-defined* type with a known shape — one absolute assertion such as `Assert.Contains("2|Eleven:System.Int32:True|TwentyTwo:System.String:False", expected)`. That is the assertion that cannot pass vacuously, and it is one line.

### NIT — all three "driver" columns run in-process
`:57-88` invokes `Program.Main`, `GSharp.Repl.Program.Main`, and `Assembly.Load` + `entryPoint.Invoke` inside the test host — the #3130 pattern. My probes used the real out-of-process binaries and reproduced the divergences, so it is not masking *these* findings, but the columns are not what a user runs.

---

**Census or sample?** Sample. 11 kinds is real breadth, and the four axes (count/names/CLR types/readonly) are compared by full-string parity rather than count alone — that part is stronger than it looks. But the specimen set omits the two cells that are broken: a nullable field wrapping a type parameter, and static/const fields. **Specimens absent, not merely assertions blind** — and both absences are exactly where the product is wrong.

**Disclosure of scope:** I ran **1 of 9** test projects (enumerated from `GSharp.sln`, not `ls test/`) — `Interpreter.Tests`, filtered to the 11 oracle cases — plus real-driver probes. I did **not** verify the claimed 15,107-test solution run. Load exceeded 30 twice (31.54, 31.66); I built nothing at those readings and waited.

DavidObando:
# Round 2 review — PR #3151 `fix/3137-work` @ `8cd87cf4`

**Verdict: MERGEABLE.** Round 1's BLOCKER is dead, verified two independent ways. The oracle is now genuinely two-sided. The decorative positive control was replaced with one that demonstrably fires. **0 BLOCKER, 1 SHOULD-FIX, 2 NIT.**

Measured in a throwaway worktree at `8cd87cf4` vs base `61fe9cb9`. Both trees `BUILD_RC=0`, 0 warnings.

---

## Merge hygiene

`origin/main` = `61fe9cb9`; branch **0 behind / 2 ahead**; merge base **is** main. `git merge-tree` *output* (not rc): 1 line, 0 conflict markers, predicted tree `278ed880` — reproduced exactly by the PR head's own tree. Main touched **zero** of this PR's files since base; file-set intersection empty.

---

## Round 1 finding #1 — the `:1416` nullable branch. **DEAD.**

Threaded now at `Evaluator.Expressions.cs:1432`: `ResolveClrBackingTypeArgument(nullable.UnderlyingType, reifiedTypeParameters)`.

Verified by **measurement**, not reading: mutant **M1** un-threads that single argument and reproduces round 1's exact symptom — `Nullable:System.Object` instead of `System.Int32` — and turns the new test RED.

### Full enumeration of `ResolveClrBackingTypeArgument`, per branch

| # | branch | `file:line` | kind | threaded? | covered? |
|---|---|---|---|---|---|
| 1 | `TypeParameterSymbol` dict hit | `:1424` | **new base case** | n/a | M7 |
| 2 | `NullableTypeSymbol` | `:1432` | recursive | **YES** *(was the BLOCKER)* | M1 |
| 3 | `ImportedTypeSymbol` `.Select(...)` | `:1446` | recursive | YES | M2 |
| 4 | `SliceTypeSymbol` | `:1457` | recursive | YES | M3 |
| 5 | `ArrayTypeSymbol` | `:1462` | recursive | YES | M4 |
| 6 | `argument.ClrType is Type` | — | base case | n/a | — |
| 7 | `StructSymbol { IsClass: true }` → `CreateClrBackingType` | — | mutual recursion into a **different** type's scope | correctly NOT threaded | — |
| 8 | `StructSymbol or EnumSymbol` → `CreateClrReifiedType` | — | same | correctly NOT threaded | — |

**There is no fifth recursive branch.** The only remaining single-argument call site is `:1414`, which resolves the *closed* type's concrete arguments where no type parameters are in scope — correct as written.

---

## Round 1 finding #2 — the blind oracle. **DEAD.**

There is a nullable specimen now. `Issue3099TypeArgumentReificationTests.cs:36`, case `gsharp-generic`:

```
struct Payload[T] { let Value T; var Imported List[T]; var Slice []T
                    var Nullable T?; var Array [3]T
                    shared { var Stat int32 }; const Kilo int32 = 1000 }
```

One specimen covering **all four** recursive branches plus static and const fields.

**Two-sided, as required:** the new test file copied onto the **base product** (`61fe9cb9`) → **11 of 11 FAIL**. Same file on the merged tree → **11/11 pass**. Round 1's "11/11 with the bug present *and* absent" is decisively fixed.

### Finding A — live three-driver repro (`struct Payload[T]`, 7 fields, all four branch kinds + static + const)

| driver | BASE `61fe9cb9` | MERGED `8cd87cf4` |
|---|---|---|
| bare `gsc` (evaluate) | `0` — **zero fields** | full 7-field shape |
| `gsc /out:` (**emit = oracle**) | full 7-field shape | full 7-field shape (unchanged) |
| `gsi` (interpret) | `0` — **zero fields** | full 7-field shape |

Merged output is **byte-identical to the emit oracle on all three drivers**, including `Nullable:System.Int32`.

---

## Round 1 finding #3 — the decorative `ConsoleColor` control. **REPLACED, and it fires.**

`ConsoleColor` and `Assert.Matches` are both **absent** from the file. The control is now a user-defined `struct ReflectionControl { let Eleven int32; var TwentyTwo string }` — two **instance** fields — asserted with `Assert.Contains(ControlShape, ...)` against all three columns at `:97` / `:98` / `:99`.

It is not decoration: **all 11 base failures land on line 98** — `Assert.Contains(ControlShape, evaluated, ...)`, the *evaluate* column. Line 97 (emit) passes, which is correct since this PR does not touch emit. All three of round 1's objections are answered — it fires, it moves, and it is an instance-field specimen the product can break.

Round 1's related SHOULD-FIX (`:1344` `definition.Fields` instance-only) is also resolved: the loop now iterates `definition.Fields.Concat(definition.StaticFields).Concat(definition.ConstFields)`.

---

## Mutation evidence — 7 PRODUCT mutants, two-sided, all asserted applied

All `BUILD_RC=0`, all RED, **zero no-results**. Naming the dying case in each row.

| id | hunk / mutation | probe effect | test |
|---|---|---|---|
| M1 | `:1432` nullable — drop dict arg | `Nullable:System.Object` | **RED 1/11** (`gsharp-generic`) |
| M2 | `:1446` imported — drop dict arg | `List<System.Object>` | **RED 1/11** |
| M3 | `:1457` slice — drop dict arg | `Slice:System.Object[]`, Array stays correct | **RED 1/11** |
| M4 | `:1462` array — drop dict arg | `Array:System.Object[]`, Slice stays correct | **RED 1/11** |
| M5 | `:1344` drop `.Concat(definition.StaticFields)` | count 7→6, `Stat` gone | **RED 1/11** |
| M6 | **condition inversion** `if (field.IsReadOnly)` → `if (!field.IsReadOnly)` | every `IsInitOnly` flag flipped | **RED 11/11** |
| M7 | `:1336` never populate the reified map | every `T`-derived field → `Object` | **RED 1/11** |

M3 and M4 each leave the *other*'s field correct — slice and array are **independently** pinned, not jointly. M6 is the control I expected to kill broadly, and it did (11/11), so the narrow 1/11 rows mean something.

Round-trip verified exact: `git diff --stat -- src/` empty, `git status --porcelain` empty, rebuild `BUILD_RC=0`, 11/11 green.

---

## SHOULD-FIX 1 — residual erasure: fields typed by an **enclosing** generic's parameter

`Evaluator.Expressions.cs:1312-1315` defines the **enclosing** builder's generic parameters but **discards the return value**; `:1333-1341` populates `reifiedTypeParameters` from `definition.TypeParameters` only, inside `if (arity > 0)`. For a type nested in a generic, the nested type's own arity is 0, so the map is empty and the enclosing `T` falls through to `typeof(object)`.

Structurally the **same omission** as round 1's BLOCKER — a `DefineGenericParameters` result not threaded — one axis over.

```
class Owner[T] { class Inner { var V T; var W int32 } }
class Box[T] : EventArgs {}
var b = Box[Owner[int32].Inner]()
Console.WriteLine(FieldShape(b.GetType().GenericTypeArguments[0]))
```

| driver | BASE | MERGED `8cd87cf4` |
|---|---|---|
| `gsc /out:` (**oracle**) | `2|V:System.Int32|W:System.Int32` | `2|V:System.Int32|W:System.Int32` |
| bare `gsc` (evaluate) | `0` | `2|V:`**`System.Object`**`|W:System.Int32` |
| `gsi` (interpret) | `0` | `2|V:`**`System.Object`**`|W:System.Int32` |

**Not a regression** — base was already divergent here (zero fields). The PR narrows the divergence from "no fields at all" to "right count, wrong type", but the oracle still disagrees, **silently, with rc0 and no diagnostic**.

**Why this is not a BLOCKER:** the headline shapes named by the issue are all fixed and mutation-pinned; the oracle column is correct; base was already wrong. Round 1's BLOCKER was justified because the *headline* shape was broken and the PR summary misdescribed it. That is not the case here.

**Why it is still worth fixing before merge:** the matrix already ships cases literally named `nested-class` and `nested-struct` (`:30`, `:31`) — but both nest inside a **non-generic** `Owner`, so the generic×nested cross product is untested. The names create confidence the coverage does not earn. This is round 1's own "specimens absent, not assertions blind" diagnosis, one axis over.

**I measured the obvious remedy and it does NOT work** — reporting this so the author does not spend the afternoon on the same dead end. Threading the enclosing parameters the mirror-image way:

```csharp
var enclosingParameters = enclosingBuilder.DefineGenericParameters(...);
for (var i = 0; i < enclosingParameters.Length; i++)
{
    reifiedTypeParameters[enclosing.TypeParameters[i]] = enclosingParameters[i];
}
```

builds clean (`BUILD_RC=0`, 0 warnings) and then **fails at runtime on both driver columns**:

```
error GS9999: Type 'Inner' from assembly 'GSharp.Interpreter.ClrBacking.2' has a field of an illegal type.
```

That is CLR-correct, not a bug in the patch: in metadata a nested type of a generic must **redeclare** the enclosing arity (C# emits `Owner<T>.Inner<T>`), and the reified nested builder is arity 0, so the enclosing `GenericTypeParameterBuilder` is out of scope for its fields. **The real fix is structural, not a one-liner.**

Given that, the cheap and honest resolution is either a follow-up issue, or adding the shape to the matrix as a documented known divergence — **not** attempting the structural fix in this PR.

## NIT 1 — `= null` default hides the next occurrence of this exact bug

`:1420-1422` — `ResolveClrBackingTypeArgument(TypeSymbol argument, IReadOnlyDictionary<TypeParameterSymbol, Type> reifiedTypeParameters = null)`.

The default is what let round 1's BLOCKER exist: an unthreaded branch **compiles silently** and regresses to `typeof(object)`. Making the parameter required (and passing explicitly at the one genuine single-arg site `:1414`) converts the next occurrence from a silent wrong answer into a compile error. Cheap, and it retires the whole defect class rather than this instance of it.

## NIT 2 — enclosing reified types get no fields at all

The enclosing builders at `:1300-1318` are created with `typeof(object)` base, `Sealed`, and **no field loop** — only `definition` gets the new `:1344` loop. Same root as SHOULD-FIX 1; noting it so the follow-up scopes both.

---

## Test denominator

**9 of 9 test projects**, enumerated from `dotnet sln GSharp.sln list` — the same enumeration CI itself uses at `build/generate-ci-test-matrix.py:26`. Unfiltered, no `--no-build`, no `GITHUB_ACTIONS`.

| project | result |
|---|---|
| Interpreter.Tests | 1220 / 1220 |
| Core.Tests | 7116 passed, 1 skipped, 7117 |
| Compiler.Tests | 4232 / 4232 |
| Extensions.Tests | 104 / 104 |
| InternalAnalyzers.Tests | 7 / 7 |
| LanguageServer.Tests | 462 / 462 |
| Sdk.Tests | 59 / 59 |
| Cs2Gs.Tests | 1903 / 1903 |
| GSharp.GeneratorHost.Tests | 31 / 31 |

**Total 15,135 executed, 0 failed, 1 skipped.** `Cs2Gs.Tests` spawns real `gsc.dll` out-of-process and is green after a solution-level build; `Compiler.Tests` took 1 h 15 m. Nothing was filtered and nothing was excluded — `Sdk.Tests` is in bounds and was run.

---

## Self-disclosed errors from this review

1. **I misread a truncated xUnit haystack and briefly concluded the control fired against the *emit* column** — impossible, since this PR touches zero emit code. All three columns share the `"11\nIssue3099.Case…"` prefix and xUnit elides the rest with `···`, so the message cannot identify the column. My standalone emit measurement (base and merged emit outputs **identical**, both containing the exact ControlShape) contradicted the reading, and the **stack-trace line number** settled it at line 98. Measuring beat reading; the assertion-message text was never a column signal.
2. **My proposed remedy for SHOULD-FIX 1 was wrong and I found out by building it, not by arguing it.** It compiled clean and then produced `GS9999` at runtime. Had I filed it as "the fix is two lines" without running it, I would have sent the author into a CLR dead end. Recorded above as measured evidence *against* my own proposal.

PR #3152, opened 2026-08-03
Pin constructor ref-struct liveness analysis

### LINKED ISSUE #3145: Binder.cs:1600 constructor RefStructAsyncLivenessAnalyzer is unpinned: reverting it leaves the whole suite green
## Summary

PR #3070 (merged, issue #3058) routed constructor bodies through `RefStructAsyncLivenessAnalyzer` at `src/Core/CodeAnalysis/Binding/Binder.cs:1600`, where **neither** analyzer ran before. A reviewer graded this SHOULD-FIX because the PR body never mentioned it, and it merged unresolved.

It has now been settled by measurement. **The change is a fix, not a break — but it is pinned by nothing.**

## What was measured

On the merged tree versus the pre-#3070 tree, across **9 correct constructor shapes × 3 drivers**:

- **Zero false positives.** No correct constructor stopped compiling.
- **The pre-#3070 tree printed `0` where the correct answer is `3`** — so the constructor path was genuinely under-analysed before, and #3070 improved it.
- **Codegen byte-identical, 9/9.**

So the behaviour change is real, correct, and beneficial. Nobody claimed it, and it turns out to be worth having.

## The actual defect

> **Reverting `Binder.cs:1600` leaves the entire suite green.**

All five `GS0219` test files contain **zero** `init(` shapes. Nothing exercises the constructor path through this analyzer. The improvement #3070 shipped is invisible to the test suite in both directions: it did not fail before the fix, and it does not fail if the fix is removed.

That makes the hunk **live product code pinned by no test** — the same category as the `.Concat(nestedStructShells.Values)` survivor found in #3115, and the deinit inheritance shape found in #3131. A future refactor can silently revert this and every gate stays green.

## Why this is worth closing rather than shrugging at

The value of a fix nobody can detect is zero the moment someone reverts it. This one is *especially* easy to revert, because the PR body never explained why the line was added — a maintainer reading `Binder.cs:1600` sees an analyzer call on a path that has no tests and no rationale, and removing it costs them nothing.

## Suggested approach

Add `GS0219` coverage over constructor bodies — `init(` shapes carrying a `ref struct`, ideally including one that crosses a suspension point, since that is what `RefStructAsyncLivenessAnalyzer` exists to catch.

The bar is **two-sided mutation**, which is now the standard: the new test must go **RED** when `Binder.cs:1600` is reverted, and **GREEN** on `main` without your PR. A test that only satisfies one side is not evidence.

## Bar

Three driver columns: bare `gsc` **evaluates**, `gsc /out:` **emits**, `gsi` **interprets**. This is bind-time, so all three should agree — compare on **rc and diagnostic set**, not printed values; any divergence is itself a finding.

Only PRODUCT mutants count; condition inversion, never `if (true)`; mutate per hunk; **assert the mutant applied** — a no-result is not a survival.

Full suite unfiltered, **enumerated from `GSharp.sln` — there are 9 test projects**, not 7; `ls test/` is wrong in both directions (it counts `test/Shared`, which is not a project, and misses `tools/cs2gs/Cs2Gs.Tests` and `tools/gsgen/GSharp.GeneratorHost.Tests`). `LanguageServer.Tests` matters here specifically: it consumes binder diagnostics, and this is a binder change.

**Grep the tree you are measuring, not the tree you are standing in** — verify `git rev-parse HEAD` against `origin/main` before drawing conclusions from any checkout.

DavidObando:
Fixes #3145

## Summary
- cover ref-struct liveness inside async lambdas declared in constructor bodies
- verify unsafe live-across-await and safe dead-before-await shapes
- exercise bare `gsc`, emitted `gsc /out:`, and `gsi` with isolated output directories

## Two-sided mutation
On rebased head `c2677d87`, reverting `Binder.cs:1600` was asserted applied and changed the focused matrix from **6/6 green** to **3 failed / 3 passed**. All unsafe driver rows changed from rc 1 + `GS0219` to rc 0 + no diagnostic; safe rows stayed green. Restoring product source returned **6/6 green**.

## Validation
- focused driver matrix on `faca348a`: **6 passed / 0 failed**
- full Release suite on immediately prior merged base `f6f8acc0`: **15,125 passed / 0 failed / 1 skipped / 15,126 total**
- all nine projects ran: Core.Tests, Compiler.Tests, Interpreter.Tests, LanguageServer.Tests, Sdk.Tests, Extensions.Tests, InternalAnalyzers.Tests, Cs2Gs.Tests, GSharp.GeneratorHost.Tests
- final CI run `30827260970`: every non-skipped job passed, including `cs2gs-oahu` and all nine test-project shards

### DIFF
--- a/test/Interpreter.Tests/Issue3145ConstructorRefStructLivenessTests.cs
+++ b/test/Interpreter.Tests/Issue3145ConstructorRefStructLivenessTests.cs
@@ -1,0 +1,167 @@
+// <copyright file="Issue3145ConstructorRefStructLivenessTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+using Xunit;
+
+namespace GSharp.Interpreter.Tests;
+
+/// <summary>Issue #3145: constructor bodies run async ref-struct liveness analysis.</summary>
+[Collection("ConsoleIo")]
+public class Issue3145ConstructorRefStructLivenessTests
+{
+    private const string UnsafeSource = """
+        import System
+        import System.Threading.Tasks
+
+        class Holder {
+            var Value int32
+
+            init(values []int32) {
+                let read = async func() int32 {
+                    var span ReadOnlySpan[int32] = values
+                    await Task.Yield()
+                    return span.Length
+                }
+                Value = read().Result
+            }
+        }
+
+        """;
+
+    private const string SafeSource = """
+        import System
+        import System.Threading.Tasks
+
+        class Holder {
+            var Value int32
+
+            init(values []int32) {
+                let read = async func() int32 {
+                    var span ReadOnlySpan[int32] = values
+                    var length = span.Length
+                    await Task.Yield()
+                    return length
+                }
+                Value = read().Result
+            }
+        }
+
+        """;
+
+    public static IEnumerable<object[]> Drivers()
+    {
+        foreach (var driver in Enum.GetValues<Driver>())
+        {
+            yield return new object[] { driver };
+        }
+    }
+
+    [Theory]
+    [MemberData(nameof(Drivers))]
+    public void ConstructorAsyncLambda_RefStructLiveAcrossAwait_ReportsGS0219(Driver driver)
+    {
+        var result = Run(UnsafeSource, driver);
+
+        Assert.Equal(1, result.ExitCode);
+        Assert.Equal(new[] { "GS0219" }, result.DiagnosticIds);
+    }
+
+    [Theory]
+    [MemberData(nameof(Drivers))]
+    public void ConstructorAsyncLambda_RefStructDeadBeforeAwait_HasNoDiagnostics(Driver driver)
+    {
+        var result = Run(SafeSource, driver);
+
+        Assert.Equal(0, result.ExitCode);
+        Assert.Empty(result.DiagnosticIds);
+    }
+
+    public enum Driver
+    {
+        CompilerEvaluation,
+        CompilerEmission,
+        Interpreter,
+    }
+
+    private static DriverResult Run(string source, Driver driver)
+    {
+        var root = Path.Combine(
+            GetRepositoryRoot(),
+            "out",
+            "test-artifacts",
+            $"issue3145-{driver}-{Guid.NewGuid():N}");
+        Assert.False(Directory.Exists(root));
+        Directory.CreateDirectory(root);
+
+        try
+        {
+            var sourceDirectory = Directory.CreateDirectory(Path.Combine(root, "source")).FullName;
+            var outputDirectory = Directory.CreateDirectory(Path.Combine(root, "output")).FullName;
+            Assert.Empty(Directory.EnumerateFileSystemEntries(outputDirectory));
+
+            var sourcePath = Path.Combine(sourceDirectory, "Probe.gs");
+            File.WriteAllText(sourcePath, source);
+            var result = driver switch
+            {
+                Driver.CompilerEvaluation => Capture(() => GSharp.Compiler.Program.Main(new[] { sourcePath })),
+                Driver.CompilerEmission => Capture(() => GSharp.Compiler.Program.Main(new[]
+                {
+                    "/out:" + Path.Combine(outputDirectory, "Probe.dll"),
+                    sourcePath,
+                })),
+                Driver.Interpreter => Capture(() => GSharp.Repl.Program.Main(new[] { sourcePath })),
+                _ => throw new InvalidOperationException($"Unexpected driver {driver}."),
+            };
+
+            var ids = Regex.Matches(result.StandardOutput + result.StandardError, @"\bGS\d{4}\b")
+                .Select(match => match.Value)
+                .Distinct(StringComparer.Ordinal)
+                .OrderBy(id => id, StringComparer.Ordinal)
+                .ToArray();
+            return new DriverResult(result.ExitCode, ids);
+        }
+        finally
+        {
+            Directory.Delete(root, recursive: true);
+        }
+    }
+
+    private static (int ExitCode, string StandardOutput, string StandardError) Capture(Func<int> action)
+    {
+        using var output = new StringWriter();
+        using var error = new StringWriter();
+        var previousOutput = Console.Out;
+        var previousError = Console.Error;
+        Console.SetOut(output);
+        Console.SetError(error);
+        try
+        {
+            return (action(), output.ToString(), error.ToString());
+        }
+        finally
+        {
+            Console.SetOut(previousOutput);
+            Console.SetError(previousError);
+        }
+    }
+
+    private static string GetRepositoryRoot()
+    {
+        var directory = new DirectoryInfo(AppContext.BaseDirectory);
+        while (directory != null && !File.Exists(Path.Combine(directory.FullName, "GSharp.sln")))
+        {
+            directory = directory.Parent;
+        }
+
+        return directory?.FullName
+            ?? throw new InvalidOperationException("Could not locate repository root.");
+    }
+
+    private sealed record DriverResult(int ExitCode, string[] DiagnosticIds);
+}


DavidObando:
# Review — PR #3152 (`fix/3145-work`), head `fc86e18d`

**Verdict: MERGEABLE** — 0 BLOCKER, 0 SHOULD-FIX, 3 NIT.

This is a test-only PR, so the review question is not "does it pass" but **"can it fail?"** It can, on both halves, and the pin is specific to the site it claims to pin. The mutation evidence is below.

## Anchors

- Live main `faca348a` (re-checked after the run; unmoved). Merge base `f6f8acc0`; branch 1 behind / 4 ahead of base.
- `git merge-tree` **output**: 1 line, **0 conflict markers**, predicted tree `b46c1fcb` — reproduced exactly by a real merge.
- File-set intersection between the PR and main's 4 new commits is **empty**; main did not touch `Binder.cs` or the analyzer.
- Merged tree `BUILD_RC=0`, 0 warnings. All 9 first-party assemblies under `Core/`, `Compiler/`, `Repl/` (excluding `ref/`) fresh within a 5-second spread, epoch-sorted.

## The mandatory experiment

The issue's acceptance criterion is that deleting the constructor analyzer call makes the new test RED. The call is at **`src/Core/CodeAnalysis/Binding/Binder.cs:1603`** — `AnalyzeFunctionBody(lowered, ctor.Function, ctorBinder.Diagnostics);`. (The issue says `:1600`; 3-line drift, same statement, and it is the only one of the seven `AnalyzeFunctionBody` call sites that binds constructor bodies.)

**Deleting it turns exactly the three unsafe rows RED, on all three drivers**, with the three safe rows still green — so the pin fires without over-firing.

## Mutation table — PRODUCT only, per hunk, each asserted applied

| id | mutation | expected | result |
|---|---|---|---|
| M1a | comment out `Binder.cs:1603` | RED | **`BUILD_RC=1` SA1515 — NO RESULT**, retargeted |
| **M1** | **delete `Binder.cs:1603`** | **RED** | **RED — 3 failed / 3 passed**; all 3 are `..._RefStructLiveAcrossAwait_ReportsGS0219` at `Issue3145…Tests.cs:71`, drivers `CompilerEvaluation`, `CompilerEmission`, `Interpreter` |
| M2a | delete `Binder.cs:1402` only | GREEN | **`BUILD_RC=1` SA1507 — NO RESULT**, retargeted |
| **M2** | **delete sibling site `Binder.cs:1402` (+ orphan blank line)** | **GREEN (specificity control)** | **GREEN — 6/6**; the test does *not* redden for "any analyzer site removed" |
| **M3** | **delete `RefStructAsyncLivenessAnalyzer.Analyze` from the shared helper (`Binder.cs:2124`)** | RED | **RED — 3 failed**; confirms `GS0219` here comes from the liveness analyzer, not from `RefKindDefiniteAssignmentAnalyzer` sharing the `ByRefLikeEscape` descriptor |
| **M4** | **over-fire: `ReportIfUnsafe(live→interesting)` at `RefStructAsyncLivenessAnalyzer.cs:530`** | RED on the **safe** rows | **RED — exactly the 3 `..._RefStructDeadBeforeAwait_HasNoDiagnostics` rows**, all drivers, expected 0 got 1 |

Both `BUILD_RC=1` rows are recorded as **no-results, not survivals**, and were retargeted — the bootstrap/StyleCop trap is live in this file.

**Why M4 matters.** M1, M2 and M3 all leave the three safe rows green, so nothing in the obvious mutant set proves the safe half is anything but decoration. M4 answers it: degrade the analyzer so it loses liveness precision and reports every ref-struct local rather than only live ones, and the safe rows — and *only* the safe rows — go red. The false-positive guard is real.

Taken together: **M1 RED + M2 GREEN + M3 RED + M4 RED-on-the-other-half** means the test pins *this analyzer*, at *this site*, in *both* directions. It is not asserting a symptom.

## The premise checks out

The issue claims the five pre-existing `GS0219` test files contain zero `init(` shapes. Measured — `Issue2350AsyncRefStructLivenessTests.cs`, `Issue2360AsyncTryFinallyReturnLivenessTests.cs`, `Issue2891SiblingAnalyzerNoDriftTests.cs`, `RefSafeEscapeTests.cs`, `RefStructTests.cs`: **0 occurrences of `init(` in all five**. The new file is the only one.

The chosen shape is also the right one. `RefStructAsyncLivenessAnalyzer.Analyze` (`:78-84`) runs `AnalyzeScope` only when `enclosing.IsAsync`, then always walks for nested async lambdas. A constructor is never async, so the constructor site's *entire* value is async lambdas declared in constructor bodies — exactly what the test exercises. There is no untested sub-shape hiding behind the `IsAsync` branch at this site.

## The gate is real, not bespoke

All six tests are discovered by the **normal unfiltered invocation** — no filter, no ordering, no manual pre-build:

```
ConstructorAsyncLambda_RefStructLiveAcrossAwait_ReportsGS0219(driver: CompilerEvaluation|CompilerEmission|Interpreter)
ConstructorAsyncLambda_RefStructDeadBeforeAwait_HasNoDiagnostics(driver: CompilerEvaluation|CompilerEmission|Interpreter)
```

Three real driver columns, each asserting `ExitCode` **and** exact diagnostic-set equality (`Assert.Equal(new[] { "GS0219" }, …)`, not `Contains`). The emit column asserts diagnostics without running the emitted assembly, which is correct here — this is a bind-time check and the issue explicitly scoped comparison to rc and diagnostic set rather than printed values. In-process `Program.Main` with `[Collection("ConsoleIo")]` is the established convention in this project, not a deviation.

## Suites — 7 of 9 projects, enumerated from `GSharp.sln`

**10,844 executed, 0 failures, 1 skipped.**

Interpreter **1220** (unfiltered) · Core **7116 + 1 skip** · Cs2Gs **1903** · LanguageServer **462** · Extensions **104** · GeneratorHost **31** · InternalAnalyzers **7**.

`LanguageServer.Tests` was run specifically because it consumes binder diagnostics.

**NOT RUN, stated plainly rather than implied:** `Compiler.Tests` (~1 h wall locally) and `Sdk.Tests` (local SDK package, out of bounds). Mitigating argument for the former: this PR changes **zero product files** — the diff is one added test file in `Interpreter.Tests` — so no other project's behaviour can move.

---

## NIT 1 — the pin lives only in `Interpreter.Tests`; `Core.Tests` still passes the revert

Measured: with `Binder.cs:1603` deleted (M1), **`Core.Tests` is still 7116 passed / 1 skipped, fully green.** Only `Interpreter.Tests` reddens.

CI runs all nine projects, so the defect the issue names — a silent revert surviving the gate — **is closed**. But the threat model in the issue is a maintainer editing `Binder.cs` and deciding the call is dead. The suite that developer runs first is `Core.Tests`, which is where all five existing `GS0219` files live and where the binder's own tests are, and it stays green. A small binder-level assertion co-located in `Core.Tests` alongside `Issue2350AsyncRefStructLivenessTests` would make the pin fire in the fastest local feedback loop, not only in CI.

For contrast, the sibling top-level-function site at `Binder.cs:1402` is heavily pinned: deleting it fails **24 Core.Tests**. The constructor site went from 0 to 3.

## NIT 2 — the failure message under the exact revert this PR guards never mentions `GS0219`

`Issue3145ConstructorRefStructLivenessTests.cs:70-72` asserts `ExitCode` first and the diagnostic set second. Under M1 — the precise regression this test exists to catch — every failure reads:

```
Assert.Equal() Failure: Values differ
Expected: 1
Actual:   0
```

with no mention of `GS0219`, the analyzer, or the constructor site. Asserting the diagnostic set first would make the failure name the defect. That matters more than usual for a test whose whole job is to explain itself to someone who has just deleted a line and is wondering why.

## NIT 3 — pre-existing sibling gap: `shared { init { … } }` runs no liveness analysis at all

Not introduced by this PR and out of its scope, but it is the same defect class in the adjacent path, so recording it under scope honesty rather than leaving it unsaid.

`BindStaticInitializerBlocks` (`Binder.cs:2140-2177`) lowers the static-initializer body and calls `SetStaticInitializerStatements`, but **never calls `AnalyzeFunctionBody`** — so neither `RefKindDefiniteAssignmentAnalyzer` nor `RefStructAsyncLivenessAnalyzer` runs on `shared { init { … } }` bodies. I measured this rather than inferring it from the grep, with a positive control:

| shape | bare `gsc` | `gsc /out:` | `gsi` |
|---|---|---|---|
| async lambda, ref struct live across `await`, inside `init(…)` **(control)** | rc=1 `GS0219` | rc=1 `GS0219` | rc=1 `GS0219` |
| **identical lambda inside `shared { init { … } }`** | **rc=0, no diagnostics** | **rc=0, no diagnostics** | **rc=0, no diagnostics** |

The control fires, so the silence is absence and not scope: unsafe code that is rejected in a constructor compiles clean in a static initializer, uniformly across all three drivers. Worth a follow-up issue at the maintainer's discretion; it should not hold up this PR.

---

## Disclosures against my own work

- I twice let `set -e` plus a non-matching `grep` abort a probe mid-run, and once let it kill a probe on an expected non-zero `gsc` exit code. Each time the missing output — not a green — is what exposed it. Re-run without `set -e`.
- My first mutant attempt (`M1a`) looked like it applied but did not; the **assert-applied check caught it**. Had I skipped that check I would have built, seen 6/6 green, and wrongly reported that the test fails to pin the line it exists to pin — the exact false verdict this PR is about.
- My first freshness sweep used the glob `Gsharp*.dll` and returned 3 files. That was **case-sensitivity, not absence** — the assembly is `GSharp.Core.dll`. Re-measured with `-iname`: 9 files, all fresh.
- My `--list-tests` discovery check first reported **0** occurrences of the test class. That was scope, not absence — the listing emits bare method names, not fully-qualified ones. Positive-controlling the listing itself caught it, and the corrected grep found all 6.

## Tree state

Every mutant restored from a pristine backup and verified: `git diff --stat -- src/` empty, `git status --porcelain` empty, rebuilt `BUILD_RC=0`, 6/6 green. All work confined to `/tmp/g3152`; no worktree I do not own was touched.

DavidObando:
# Review: PR #3152 — `fix/3145-work` @ `c2677d87`

**Verdict: MERGEABLE.**

Independent second review. Base verified: `c2677d87` is exactly one commit above current `main` `faca348a`, merge-base `faca348a` — no moved base, no force-push. Diff is test-only: `+167 -0`, one new file.

---

## The single question: can the new test fail? **Yes — reproduced independently.**

Full 2x2. Every mutant is a PRODUCT mutant, per hunk, asserted applied, rebuilt each time, `rc` captured before any pipe:

| product state | new test (6) | pre-existing GS0219 tests (101) |
|---|---|---|
| head `c2677d87`, unmutated | **6/6 GREEN** rc=0 | **101/101 GREEN** rc=0 |
| **M1** — remove ctor call `Binder.cs:1603` | **3 FAILED / 3 passed** rc=1 — killed | **101/101 GREEN** <- new test is the *sole* pin |
| **M2** — remove function call `Binder.cs:1402` | **6/6 GREEN** <- specific | **18 FAILED / 83 passed** rc=1 — control killed |

- **M1 compiled cleanly**, so it is a real mutant, not a no-result.
- **M1 fails for the right reason**: `Assert.Equal(1, result.ExitCode)` at line 71, rc **1 -> 0** — the diagnostic disappears, exactly as it should when the analyzer stops running on the constructor path. All three drivers flip identically.
- **M2 answers "pins the call, or a symptom?"** The test *survives* removal of a **different** `AnalyzeFunctionBody` site. It does not redden merely because "some analyzer call went away" — it is specific to the **constructor** site. That is a stronger result than the PR body claims.
- **M2's first attempt was a no-result, not a survival**: deleting the line alone left a doubled blank line and failed `SA1507`. Retargeted (call + preceding blank), rebuilt, then scored.
- Restoration verified **byte-identical** to `c2677d87` (`git status --porcelain` empty), rebuilt, and full `Interpreter.Tests` green afterwards.

Issue #3145's premise is confirmed by measurement too: under M1 the 101 pre-existing GS0219 tests stay green. This PR closes a real hole.

**Three drivers** compared on rc + diagnostic set (not printed values): they agree in every cell. **No divergence.**

---

## Findings

### NIT 1 — line-number drift
Issue #3145 and the PR body both say `Binder.cs:1600`. At this head the constructor call is `src/Core/CodeAnalysis/Binding/Binder.cs:1603`. Cosmetic only.

### NIT 2 — the pinned line does two things; the test pins one
`Binder.cs:1603` calls `AnalyzeFunctionBody`, which runs **both** `RefKindDefiniteAssignmentAnalyzer` (`Binder.cs:2123`) and `RefStructAsyncLivenessAnalyzer` (`Binder.cs:2124`). The new test pins only the GS0219 half. Removing the whole line is caught (M1 killed), so the issue's goal is met — but a narrower future edit deleting only the `RefKindDefiniteAssignment` call from the constructor path would still be invisible. Optional follow-up, explicitly **not** a blocker.

### Test quality — positive
Exact set equality `Assert.Equal(new[] { "GS0219" }, ...)` rather than `Contains`, so a *different* diagnostic also fails. The safe counterpart asserts `Assert.Empty`, which resists a blanket always-report mutant. `[Collection("ConsoleIo")]` serialises the Console redirection; Guid-isolated output dir asserted non-existent then removed in `finally`; `out/test-artifacts` is empty afterwards.

---

## Denominator — including a correction to my own first number

**Enumeration source: `dotnet sln GSharp.sln list` -> 9 test projects.**

**Self-disclosure: my first denominator was wrong.** I counted via `dotnet test --list-tests` and got 15,008. That is an **undercount**: `Core.Tests` **lists 6,990 but runs 7,117** — a 127 gap. I verified my own grep (6,992 lines minus 2 headers), so the shortfall is in `--list-tests` itself. It matched *exactly* on Interpreter (1,220), LanguageServer (462) and Cs2Gs (1,903) — 3 of 4 — which is precisely what makes it treacherous.

**`--list-tests` is not a denominator. Only an actual run is.**

Consequence: my initial reading that the author's **15,125 / 15,126** could not be reconciled was **wrong, and the cause was my own instrument**. The corrected floor at this head is **>= 15,135**. The author measured on `f6f8acc0`; between `f6f8acc0` and `faca348a` the test projects only *gained* code (+578 lines, including a 309-line and a 169-line new test file, and **zero** test files deleted). 15,126 on the older base against >= 15,135 on the newer is consistent and in the right direction. **The author's total reconciles.**

### What I actually ran (unfiltered, full projects)

| project | result |
|---|---|
| Core.Tests | 7,117 total — 7,116 passed, 1 skipped, rc=0 |
| Interpreter.Tests | 1,220 / 1,220, rc=0 |
| LanguageServer.Tests | 462 / 462, rc=0 |
| Sdk.Tests | 59 / 59, rc=0 |

**8,858 tests, 0 failures.** Failures counted with a pattern that cannot match xUnit display names.

**NOT RUN — stated plainly, no implied coverage:** Compiler.Tests, Cs2Gs.Tests, Extensions.Tests, InternalAnalyzers.Tests, GSharp.GeneratorHost.Tests. Reason: this PR is **test-only**, adds no product code, and one test assembly cannot observe another, so those projects cannot be affected by this diff. Compiler.Tests alone is ~87 min. Its 4,232 is list-derived and therefore a **floor**, not a measurement.

### `Sdk.Tests` is **not** out of bounds at this head

The standing rule excludes it citing `test/Sdk.Tests/BuildTaskSilentFailureTests.cs:22` as driving `dotnet build` against the shared SDK package. At this head **line 22 is a sentence inside a `<para>` docstring**, and it says the *other* acceptance tests do that. This file invokes `Program.Main` in-process.

An exhaustive grep across all 9 files in `test/Sdk.Tests` for process-spawn and write patterns yields only: the property name `SkipCompilerExecution`, in-process MSBuild `task.Execute()`, a test *method* name containing "Executable", and one `Directory.Delete` of the test's own Guid temp dir in a `finally`. **No process spawn, no `dotnet build`, no access to the shared SDK package.** Measured: 59/59 green.

I fingerprinted the shared package before and after. It *did* change (140 -> 141 versions), but the new version `0.3.572-gefaa01ed98` has mtime `1785777200` — **20 s after my run finished** (`1785777180`) — and carries a foreign commit SHA, not my `c2677d87`. Another agent's build. **The load-bearing evidence here is the structural grep, not the fingerprint**: on a machine with ~20 concurrent agents a shared-resource fingerprint is a noisy instrument, and I will not rest a claim on it.

So the author running `Sdk.Tests` was **not** a violation.

---

## Summary

- **BLOCKER:** none.
- **SHOULD-FIX:** none.
- **NIT:** line-number drift `:1600` -> `:1603`; the pinned line covers two analyzers and the test pins one.

The one question that decides this PR — *can the new test fail?* — is answered **yes**: reproduced independently, killed for the right reason, and shown **specific** to the constructor site by a second mutant it survives, alongside a control that kills 18 tests.

Issue #3135, opened 2026-08-03
BCL override dispatch has no cache: struct without overrides is ~1.8x slower under gsi after #3127


DavidObando:
## Summary

PR #3127 (merged, issue #3116) routes `Equals` / `GetHashCode` / `ToString` through `FindExternalOverride` at the boundary where a G# value is handed to the BCL. That fix is correct. But it runs the override lookup on **every** such call with **no cache**, so a struct with **no overrides at all** pays the full cost for zero benefit.

Measured by a rubber-duck during the #3127 review, and graded SHOULD-FIX rather than a blocker because it is correctness-neutral and the interpreter is not the performance driver. Filing so it is not lost.

## Measured

`gsi`, 20,000 inserts + 20,000 lookups, `struct Plain { var n int32 }` with **no overrides**, best of 3 runs, output `20000` on both trees:

| tree | total | minus startup | work |
|---|---:|---:|---:|
| base | 553 ms | 218 ms | **335 ms** |
| merged | 835 ms | 224 ms | **611 ms** |

**≈1.8× slower on the work portion.** Startup is unchanged (218 vs 224 ms), which validates the measurement setup rather than the conclusion resting on total wall time. The two trees differ only by the PR, so attribution is clean.

## Cause

`TryInvokeObjectOverride` calls `FindExternalOverride` on every `Equals`/`GetHashCode`/`ToString` invocation. That performs a linear scan over every method of every type in the base chain, with a nested `while` walk and a `MethodInfo.Equals` comparison per candidate, plus an `object[]` allocation per call. Nothing is memoised, and there is no fast path for the common case where the type declares no override at all.

## Suggested fix

Memoise per `(StructSymbol, MethodInfo)` and fast-path the null result. This matches the PR's own centralising design — there is exactly one routing point, so there is exactly one place to add the cache. The reviewer specifically noted the fix should be small and should not require restructuring anything.

## Bar

**Measure, do not assume.** Report before/after with startup subtracted, best-of-N, on trees that differ only by the change. A cache that is not measured is a hypothesis.

Correctness must not move: re-run the #3116 matrix (`Dictionary` key, `HashSet` element, `List.Contains`, `String.Format`/interpolation `ToString`) across all three drivers — bare `gsc` **evaluates**, `gsc /out:` **emits**, `gsi` **interprets** — plus the depth-4 override-walk pin. A cache keyed too coarsely will return the wrong override for a derived type, which is silent and worse than being slow.

[...]

PR #3153, opened 2026-08-03
Fix evaluator class reference identity

### LINKED ISSUE #3134: Evaluator gives classes value equality instead of reference identity: StructValue.Equals/GetHashCode have no IsClass branch
## Summary

`StructValue.Equals` / `GetHashCode` have **no `IsClass` branch**, so a G# **class** without overrides gets **value equality instead of reference identity** under both evaluating drivers. Two distinct instances with equal field values are treated as the same object.

Found by a rubber-duck while reviewing PR #3127. **Pre-existing — identical on the base and merged trees, not caused by #3127.** It sits in the exact method that PR edited, which is why it surfaced now.

## Measured

Two distinct `PlainC` instances with equal fields, used as dictionary keys:

| probe | bare `gsc` (evaluate) | `gsi` | `gsc /out:` (emit) |
|---|---|---|---|
| `d.Count` | **1** | 2 | **2** ← correct |
| `a.Equals(b)` | **True** | False | **True** ← wait, see below |

The emit column is the oracle and it is right: two distinct class instances are two distinct keys. The evaluating path collapses them into one.

## Why this matters

This is the **same defect class as #3116**: a silent container-count divergence, rc 0, no diagnostic, no crash. The program runs and quietly produces a different answer depending on which driver executed it. Reference identity is not a subtle corner of class semantics — it is the defining difference between a class and a struct, and getting it wrong means a `Dictionary`, `HashSet`, or `List.Contains` over class keys silently deduplicates entries that should be distinct.

## Why it is filed separately rather than folded into #3127

Fixing it means giving `StructValue` reference semantics when it is backing a class, which is a behaviour change in its own right with its own blast radius — every equality, hashing and containment site that currently goes through `StructValue` for a class instance. That deserves its own false-positive corpus rather than riding along inside an unrelated fix. The reviewer flagged it as out of scope for #3127 and correctly declined to widen that PR.

## Suggested approach

`StructValue` is used for both structs and classes. The fix is presumably an `IsClass` branch in `Equals`/`GetHashCode` mirroring the one the type already carries elsewhere, so classes fall back to reference identity. **Verify what the absence was protecting against before adding it** — the neighbouring code has deliberate-looking filters, and one of them (`IsClass: true` in `Evaluator.Expressions.cs`) turned out to be genuinely deliberate when a different agent investigated #3099.

## Bar

Three driver columns: bare `gsc` **evaluates**, `gsc /out:` **emits**, `gsi` **interprets**. Emit is the oracle. Cover `{class with overrides, class without overrides, struct with overrides, struct without overrides}` × `{Dictionary key, HashSet element, List.Contains, direct .Equals}`, and assert **contents**, not only `Count` — a count can agree by coincidence.

Build a false-positive corpus: everything that behaves correctly today must still do so. Structs must keep value equality; only classes should change.

DavidObando:
## Summary

- give ordinary G# classes reference identity for evaluator `==`, `!=`, `Equals`, and `GetHashCode`
- preserve explicit `Equals`/`GetHashCode` overrides
- preserve structural semantics for `data class`, data/value structs, and inline structs
- compare bare `gsc` and `gsi` against emitted execution across six surfaces

Closes #3134.

## Root semantics

`StructDeclarationSyntax.IsClass` comes directly from the declaration keyword (`StructKeyword.Kind == ClassKeyword`). `DeclarationBinder.CreateAndRegisterStructShell` copies it into immutable `StructSymbol.IsClass`, and generic construction copies the definition flag. The identity branch therefore reads semantic type identity, not a bound-node-shape enumeration.

`IsData` similarly comes from the `data` keyword. Existing tests proved that `data class` deliberately keeps structural equality, so the default identity guard is `IsClass && !IsData`.

Operator resolution needs one extra construction tag: `MakeReferenceEquality` marks only the reference-equality fallback created after built-in, user-defined, and CLR operator resolution. The evaluator reads that tag instead of guessing from runtime values or enumerating expression shapes.

## Measured three-driver table

Tuples use `(equal-field B, same-reference C, different-value Other)`. Collection rows assert contents as well as counts.

| Phase | Driver | `==` | `.Equals` | hash equal `(B,C)` | Dictionary `(count,A,B)` | HashSet `(count,addA,addB)` |
|---|---|---|---|---|---|---|
| before | bare `gsc` | `T,T,F` | `T,T,F` | `T,T` | `1,B-11,B-11` | `1,T,F` |
| before | `gsc /out:` | `F,T,F` | `F,T,F` | `F,T` | `2,A-11,B-11` | `2,T,T` |
| before | `gsi` | `T,T,F` | `T,T,F` | `T,T` | `1,B-11,B-11` | `1,T,F` |
| after | bare `gsc` | `F,T,F` | `F,T,F` | `F,T` | `2,A-11,B-11` | `2,T,T` |
| after | `gsc /out:` | `F,T,F` | `F,T,F` | `F,T` | `2,A-11,B-11` | `2,T,T` |
| after | `gsi` | `F,T,F` | `F,T,F` | `F,T` | `2,A-11,B-11` | `2,T,T` |

Emit is the oracle. Each driver runs the same top-level source from its own asserted-empty directory.

### Controls

- class overriding `Equals`/`GetHashCode`: `==` remains reference identity (`F,T,F`); `.Equals` remains `T,T,T`; equal hashes collapse Dictionary/HashSet as specified by the override
- `data class`: `==` and `.Equals` remain structural (`T,T,F`), equal hashes remain equal, Dictionary/HashSet still collapse equal-field instances
- structs with and without overrides retain their prior value/override semantics across Dictionary, HashSet, List.Contains, `.Equals`, and hashing

## Product mutation

| Hunk mutant | Build | Killed surfaces | Surviving surfaces |
|---|---|---|---|
| invert ordinary-class `Equals` guard | rc 0 | Dictionary, HashSet, List.Contains, `.Equals` | `==`, GetHashCode |
| invert ordinary-class hash guard | rc 0 | Dictionary, HashSet, GetHashCode | `==`, List.Contains, `.Equals` |
| invert reference-operator construction tag | rc 0 | `==`/`!=` | other five |
| invert evaluator reference-tag branch | rc 0 | `==`/`!=`, hash comparison | Dictionary, HashSet, List.Contains, `.Equals` |

No product mutant survives the complete matrix. On main-without-this-PR, the override/data/struct control rows remain green across all three drivers; only the issue row differs from emit.

## Verification

- head: `d9303ce9a520071f84f7fa3ad69a7589b4c8cb2e`
- current-main merge tree exactly matched `git merge-tree` prediction against main `d957ba71`
- final solution build: 0 warnings, 0 errors
- final issue matrix: 6/6 passed
- unfiltered local denominator from `dotnet sln GSharp.sln list`: 9 test projects
- 8/9 projects run unfiltered on immediately preceding merged tree: 15,081 passed, 1 skipped, 0 failed
- `Sdk.Tests`: NOT RUN locally
- actual no-commit merge with main `d957ba71` built with 0 warnings/errors and passed the 6/6 issue matrix
- non-zero `GSharp.Core.dll` copies matched across Core, Compiler, Repl, Core.Tests, Compiler.Tests, and Interpreter.Tests; mtimes compared as epoch seconds

- final CI run `30854692245`: all required checks green, including SDK, e2e, cs2gs corpus, and Oahu

### DIFF
--- a/src/Core/CodeAnalysis/Binding/BoundBinaryOperator.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundBinaryOperator.cs
@@ -39,13 +39,20 @@ public sealed record BoundBinaryOperator
     {
     }
 
-    private BoundBinaryOperator(SyntaxKind syntaxKind, BoundBinaryOperatorKind kind, TypeSymbol leftType, TypeSymbol rightType, TypeSymbol resultType)
+    private BoundBinaryOperator(
+        SyntaxKind syntaxKind,
+        BoundBinaryOperatorKind kind,
+        TypeSymbol leftType,
+        TypeSymbol rightType,
+        TypeSymbol resultType,
+        bool isReferenceEquality = false)
     {
         SyntaxKind = syntaxKind;
         Kind = kind;
         LeftType = leftType;
         RightType = rightType;
         Type = resultType;
+        IsReferenceEquality = isReferenceEquality;
     }
 
     /// <summary>
@@ -72,6 +79,9 @@ public sealed record BoundBinaryOperator
     /// Gets the bound binary operator type symbol.
     /// </summary>
     public TypeSymbol Type { get; }
+
+    /// <summary>Gets a value indicating whether equality compares reference identity.</summary>
+    internal bool IsReferenceEquality { get; }
 
     /// <summary>
     /// Binds a syntax kind and a type symbol to the corresponding bound binary operator, or
@@ -367,7 +377,7 @@ public sealed record BoundBinaryOperator
         var kind = syntaxKind == SyntaxKind.EqualsEqualsToken
             ? BoundBinaryOperatorKind.Equals
             : BoundBinaryOperatorKind.NotEquals;
-        return new BoundBinaryOperator(syntaxKind, kind, leftType, rightType, TypeSymbol.Bool);
+        return new BoundBinaryOperator(syntaxKind, kind, leftType, rightType, TypeSymbol.Bool, isReferenceEquality: true);
     }
 
     /// <summary>
--- a/src/Core/CodeAnalysis/Evaluator.Operators.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Operators.cs
@@ -174,9 +174,9 @@ public sealed partial class Evaluator
         switch (b.Op.Kind)
         {
             case BoundBinaryOperatorKind.Equals:
-                return NumericEquals(left, right);
+                return b.Op.IsReferenceEquality ? ReferenceEquals(left, right) : NumericEquals(left, right);
             case BoundBinaryOperatorKind.NotEquals:
-                return !NumericEquals(left, right);
+                return b.Op.IsReferenceEquality ? !ReferenceEquals(left, right) : !NumericEquals(left, right);
             case BoundBinaryOperatorKind.LogicalAnd:
                 return (bool)left && (bool)right;
             case BoundBinaryOperatorKind.LogicalOr:
--- a/src/Core/CodeAnalysis/StructValue.cs
+++ b/src/Core/CodeAnalysis/StructValue.cs
@@ -162,6 +162,11 @@ public sealed class StructValue
             return (bool)result;
         }
 
+        if (StructType.IsClass && !StructType.IsData)
+        {
+            return ReferenceEquals(this, obj);
+        }
+
         if (obj is not StructValue other)
         {
             return false;
@@ -191,6 +196,11 @@ public sealed class StructValue
         if (TryInvokeObjectOverride(ObjectGetHashCodeMethod, [], out var result))
         {
             return (int)result;
+        }
+
+        if (StructType.IsClass && !StructType.IsData)
+        {
+            return RuntimeHelpers.GetHashCode(this);
         }
 
         var hash = default(HashCode);
--- a/test/Interpreter.Tests/Issue2896StructObjectOverrideTests.cs
+++ b/test/Interpreter.Tests/Issue2896StructObjectOverrideTests.cs
@@ -122,6 +122,7 @@ public class Issue2896StructObjectOverrideTests
     [InlineData("dictionary")]
     [InlineData("hashSet")]
     [InlineData("listContains")]
+    [InlineData("operatorEquals")]
     [InlineData("equals")]
     [InlineData("getHashCode")]
     public async Task Issue3134_ClassIdentityAndStructValueEqualityMatchEmitter(string surface)
@@ -452,6 +453,19 @@ public class Issue2896StructObjectOverrideTests
                     {{prefix}}Values.Contains({{prefix}}B),
                     {{prefix}}Values.Contains({{prefix}}Other)))
                 """,
+            "operatorEquals" when isClass => $$"""
+                Console.WriteLine(String.Format(
+                    "{{label}}|{0}|{1}|{2}|{3}|{4}|{5}",
+                    {{prefix}}A == {{prefix}}B,
+                    {{prefix}}A == {{prefix}}C,
+                    {{prefix}}A == {{prefix}}Other,
+                    {{prefix}}A != {{prefix}}B,
+                    {{prefix}}A != {{prefix}}C,
+                    {{prefix}}A != {{prefix}}Other))
+                """,
+            "operatorEquals" => $$"""
+                Console.WriteLine("{{label}}|N/A")
+                """,
             "equals" => $$"""
                 Console.WriteLine(String.Format(
                     "{{label}}|{0}|{1}|{2}",
@@ -498,6 +512,14 @@ public class Issue2896StructObjectOverrideTests
                 "DataClass|1|11|True|True|False",
                 "StructWithOverrides|1|11|True|True|True",
                 "StructWithoutOverrides|1|11|True|True|False",
+            },
+            "operatorEquals" => new[]
+            {
+                "ClassWithOverrides|False|True|False|True|False|True",
+                "ClassWithoutOverrides|False|True|False|True|False|True",
+                "DataClass|True|True|False|False|False|True",
+                "StructWithOverrides|N/A",
+                "StructWithoutOverrides|N/A",
             },
             "equals" => new[]
             {
--- a/test/Interpreter.Tests/Issue2896StructObjectOverrideTests.cs
+++ b/test/Interpreter.Tests/Issue2896StructObjectOverrideTests.cs
@@ -119,6 +119,23 @@ public class Issue2896StructObjectOverrideTests
     }
 
     [Theory]
+    [InlineData("dictionary")]
+    [InlineData("hashSet")]
+    [InlineData("listContains")]
+    [InlineData("equals")]
+    [InlineData("getHashCode")]
+    public async Task Issue3134_ClassIdentityAndStructValueEqualityMatchEmitter(string surface)
+    {
+        var suffix = Guid.NewGuid().ToString("N");
+        var source = BuildIssue3134Source(surface, suffix);
+        var emitted = await RunDriverAsync(source, suffix + "Emit", "gsc-emit");
+
+        AssertIssue3134EmittedSemantics(surface, emitted);
+        Assert.Equal(emitted, await RunDriverAsync(source, suffix + "Evaluate", "gsc-evaluate"));
+        Assert.Equal(emitted, await RunDriverAsync(source, suffix + "Gsi", "gsi"));
+    }
+
+    [Theory]
     [InlineData(false, "gsc-evaluate")]
     [InlineData(false, "gsc-emit")]
     [InlineData(false, "gsi")]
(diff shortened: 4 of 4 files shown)


DavidObando:
## Rubber-duck review — PR #3153 (issue #3134)

**Verdict: MERGEABLE** — 1 SHOULD-FIX, 1 NIT. No blocker.

**Anchors.** `main` `61fe9cb9`; PR head `f4e86963` (confirmed against the GitHub API, not a local ref — see disclosure 2); merge-base `f6f8acc0`. `git merge-tree` **output** = `0ddcf12c`; a real merge reproduced `0ddcf12c` **exactly**. Merged `BUILD_RC=0`, **0 warnings**. Production change is 10 lines in one file.

---

### What I verified, by measurement

**1. Both halves are fixed, and both are independently pinned.** Per-hunk PRODUCT mutants, each asserted applied (`BUILD_RC=0` *and* `GSharp.Core.dll` md5 changed):

| mutant | Core md5 | result | dying rows |
|---|---|---|---|
| **M-EQ** — remove only the `Equals` branch (`StructValue.cs:165-168`) | `129c0628` | RED 2/5 | `equals`, `listContains` |
| **M-HASH** — remove only the `GetHashCode` branch (`StructValue.cs:201-204`) | `ae8fb986` | RED 1/5 | `getHashCode` |
| **M-CTRL** — drop the `&& !StructType.IsData` record guard (control, expected to kill) | `8f40b909` | RED 5/5 | all five |

`GetHashCode` is not along for the ride: **M-HASH dies on its own**, on `Issue3134_ClassIdentityAndStructValueEqualityMatchEmitter(surface: "getHashCode")`. Baseline for both sides: merged = 5/5 **pass**; base `61fe9cb9` **+ the test file only, zero production diff** = **5/5 fail**. Every InlineData row is discriminating; no vacuous row.

**2. Sound by construction, not by enumeration.** This was the #3129 failure mode, and it does not recur. `StructSymbol.IsClass` (`Symbols/StructSymbol.cs:303`) and `IsData` (`:286`) are get-only auto-properties with a single assignment each (`:260`, `:258`), fed from `Syntax/StructDeclarationSyntax.cs:496` `IsClass => StructKeyword?.Kind == SyntaxKind.ClassKeyword` and `:434` `IsData => DataKeyword != null` — derived from the **declaration keyword**, not matched against a whitelist of node shapes. I audited **all 13** `new StructSymbol(` sites; constructed generics propagate `definition.IsClass` (`StructSymbol.cs:1790/1850/1887`). There is no unlisted shape to walk past this guard.

**3. Over-correction: probed, not assumed.** `data class` keeps value equality, and M-CTRL proves the `!IsData` guard is load-bearing rather than decorative. Two of my own hypotheses died on contact:
- *Anonymous types* — `Binding/AnonymousTypeCache.cs:77,132` pass `isData: true, isClass: false`; they are data structs, so the branch never fires.
- *Imported reference types* — `Symbols/ImportedTypeSymbol.cs:404` passes `isClass: !semantics.IsValueType`, the one non-keyword-derived site. Probed `Version` (overrides `Equals`), `string`, `TimeSpan`: **byte-identical base↔merged**, all three drivers `True`. Imported CLR instances do not flow through `StructValue`. No false positive.

**4. The fix reaches further than the corpus claims.** Three shapes the PR does not cover, all diverging on base and all correct on merged (bare | emit | gsi):

| shape | base bare/gsi | base emit | merged (all three) |
|---|---|---|---|
| `EqOnly` — overrides `Equals`, not `GetHashCode` | `True\|True\|1` | `True\|False\|2` | `True\|False\|2` ✓ |
| `DerivedC : BaseC` (inheritance) | `True\|True` | `False\|False` | `False\|False` ✓ |
| `Box[T]` (generic class) | `True\|True` | `False\|False` | `False\|False` ✓ |

`EqOnly` is worth noting: `Equals` true, hashes unequal, `Dictionary.Count == 2` — matching C#'s behaviour for a class that overrides `Equals` without `GetHashCode`.

**5. No laundered expectations — structurally.** The PR is **+237 / −0**. Zero deleted lines, so no pre-existing expectation was rewritten. And the direction is right by construction: `Issue2896StructObjectOverrideTests.cs` runs **emit first**, asserts it against independently hard-coded semantics, then compares evaluate and gsi **to emit**. The oracle is computed, not copied.

**6. The matrix is discriminating.** On base the `==` table is non-uniform (emit `False`, bare/gsi `True`), so a post-fix agreement is a real result, not an artefact of a table that cannot disagree.

---

### SHOULD-FIX 1 — `==` / `!=` still diverge from emit for a class that overrides `Equals`

`src/Core/CodeAnalysis/Evaluator.Operators.cs:177-179` routes `==`/`!=` through `NumericEquals` → `StructValue.Equals`, which invokes the user override *before* reaching the new identity branch. Emit gives reference identity.

Probe (`a==b | a==c | a!=b`), merged tree:

| type | bare | **emit (oracle)** | gsi |
|---|---|---|---|
| `ClassWithOverrides` | `True\|True\|False` | **`False\|True\|True`** | `True\|True\|False` |
| `ClassWithoutOverrides` | `False\|True\|True` | `False\|True\|True` | `False\|True\|True` ✓ |
| `DataClass` | `True\|True\|False` | `True\|True\|False` | `True\|True\|False` ✓ |

**Not a regression** — base `61fe9cb9` is identical, and the PR strictly improves the surface by fixing `ClassWithoutOverrides` as a side effect. That is why this is SHOULD-FIX and not a blocker. But the equality surface is left incoherent: `.Equals()` and containers now match emit for every class shape, while `==` matches only for classes without an `Equals` override.

**The author already has the fix and it is not in this PR.** Local branch `fix/3134-work` is at `794db9fe`, two commits ahead of the PR head, and the tip is *"Preserve evaluator reference equality operators"* — it adds `IsReferenceEquality` to `Binding/BoundBinaryOperator.cs` and switches `Evaluator.Operators.cs:177-179` to `ReferenceEquals`. Its `StructValue.cs` is byte-identical to the PR head, so that operator commit is the only genuinely new work. Recommend pushing it into this PR or filing it as a follow-up; either is fine, but it should not be lost.

### NIT 1 — the corpus declares `ClassWithOverrides` but never exercises `==`

`test/Interpreter.Tests/Issue2896StructObjectOverrideTests.cs` parameterises five surfaces (`dictionary`, `hashSet`, `listContains`, `equals`, `getHashCode`) across five types, but omits `==`/`!=` — the most common equality surface in user code. A sixth `[InlineData("equalsOperator")]` row would have caught SHOULD-FIX 1 for free, since the type corpus already contains exactly the shape that exposes it.

---

### Cross-check with #3135 — no conflict, no serialisation needed

`fix/3135-work` (`541cc61b`, no PR yet) touches `src/Core/CodeAnalysis/Evaluator.Calls.cs` **+36/−8** and does **not** touch `StructValue.cs` — the `Equals`/`GetHashCode` boundary is respected.

There is a textual near-miss worth recording: both branches add tests to the **same file**, with #3153's first hunk at old line **118** and #3135's at old line **135** — 17 lines apart, the #3024+#3052 shape where `merge-tree` reports clean and the build breaks. So I built it rather than trusting it: combined merge `MERGE_RC=0`, **`COMBO_BUILD_RC=0`, 0 warnings**, Interpreter.Tests **1226 passed / 0 failed**. The `==` gap is unchanged by #3135 (neither fixed nor worsened). **Safe to merge in either order.**

### Suites — 8 of 9, denominator from `dotnet sln GSharp.sln list`

CI derives the same list (`build/generate-ci-test-matrix.py`, `dotnet sln list` filtered to `.tests.csproj`).

Interpreter **1225/0**, Core **7116/0/1 skip** (7117 total), Extensions **104/0**, InternalAnalyzers **7/0**, LanguageServer **462/0**, Sdk **59/0**, Cs2Gs **1903/0**, GeneratorHost **31/0** — **10,908 tests, 0 failures, 1 skipped**. **Compiler.Tests was NOT run locally** (≈1h20m, sharded in CI); that is a gap in my evidence, not a pass.

---

### Disclosures

**1. I hit a false zero in my own harness, and the positive control caught it.** Checking whether my runs wrote the shared NuGet cache, `find ~/.nuget/packages/gsharp.net.sdk -newermt '@1785787813'` returned **0**. That looked like a clean bill of health and it was an artefact: BSD `find` on macOS does not accept the `@epoch` form. Positive control — `-newermt '@1'`, which must match everything — returned **1**, while the date-string form `-newermt '2026-08-03 13:10:00'` returned **147**. *A zero can be syntax, not absence.* Anyone on macOS using `-newermt '@epoch'` is getting silent false negatives.

**2. Consequently: I did write one directory into the shared SDK cache, and it was not Sdk.Tests.** `~/.nuget/packages/gsharp.net.sdk/0.3.573-g2c25a6e6da` was created at **13:21:20** — `2c25a6e6` is my own throwaway merge commit. Attribution by log-completion timestamps: Sdk.Tests **finished at 13:20:26**, 54 s *before* the write, and its log shows only scratch-local paths; **Cs2Gs.Tests ran 13:20:26 → 13:37:31** and is the writer — it invokes `gsc.dll` out-of-process and installs the SDK package to run. Interpreter.Tests *creates* the nupkg, but only into my scratch `out/bin/Release/nupkgs/`.

So the fleet guidance is inverted on this point: **Sdk.Tests is exonerated; Cs2Gs.Tests is what writes the shared cache** — which means "run all 9 projects" and "never write `~/.nuget/packages/gsharp.net.sdk/`" cannot both be satisfied. Blast radius here is nil: the write is **additive and commit-keyed**, one new directory named after my own merge commit, and no pre-existing directory was modified. The only other directory touched in the window, `0.3.573-g16e90c96e9` at 13:24:20, is **not a valid object in my clone** — a peer's concurrent write, not mine.

**3. Load and hygiene.** Builds started at load 14.97 / 21.75 / 11.54 / 15.62, all below the 30 threshold. Every probe ran in its own asserted-empty directory. All mutants reverted; `StructValue.cs` restored to a byte-exact base md5 (`528aefc2`). Scratch and the shared checkout both end clean at `61fe9cb9`, `git status --porcelain` empty. No worktree I do not own was written.

DavidObando:
## Round-2 review — PR #3153, commit `d9303ce9` only

**Verdict: MERGEABLE.** No blocker. 1 SHOULD-FIX (pre-existing, explicitly *not* introduced by this commit), 1 NIT.

**Anchors.** `main` had moved again by the time I fetched — **`d957ba71`**, four commits past the `387cb9a2` in my assignment. The author rebased: merge-base is now `387cb9a2` and the PR is two commits, `34243bc6` "Fix evaluator class identity semantics" + `d9303ce9` "Preserve evaluator reference equality operators". `git merge-tree` **output**: 0 conflict markers, `merged` result. Merged tree `f432be06`, `BUILD_RC=0`, **0 warnings**.

**Before relying on my round-1 clearance, I checked the rebase preserved it.** Blob hashes for the two files I already cleared are byte-identical across the rewrite — `StructValue.cs` `afaf62d7…` and `Issue2896StructObjectOverrideTests.cs` `4374699c…` at both `f4e86963` and `34243bc6`. So round 1 still stands and I did not re-review it.

**Isolation.** I built two trees that differ by *exactly* `d9303ce9`'s three files: `r2with` = merge(`d957ba71`, `d9303ce9`), `r2without` = merge(`d957ba71`, `34243bc6`). Every claim below is a diff between those two.

---

### 1. It fixes the cell — measured on all three drivers

| type | `r2without` bare/gsi | `r2without` **emit** | `r2with` all three |
|---|---|---|---|
| `ClassWithOverrides` | `True\|True\|False` | **`False\|True\|True`** | **`False\|True\|True`** ✓ |
| `ClassWithoutOverrides` | `False\|True\|True` | `False\|True\|True` | `False\|True\|True` ✓ |
| `DataClass` | `True\|True\|False` | `True\|True\|False` | `True\|True\|False` ✓ |

The divergence I reported in round 1 is gone: bare `gsc`, `gsc /out:` and `gsi` now return the same answer, and it is emit's.

### 2. It is pinned

`Issue2896StructObjectOverrideTests.cs:125` adds `[InlineData("operatorEquals")]`, and `:516` asserts `ClassWithOverrides|False|True|False|True|False|True` — six cells covering `==` *and* `!=` against an alias, a distinct-but-equal instance, and a different-valued instance. `DataClass|True|True|False|False|False|True` is asserted in the same row set, so the record carve-out is pinned rather than assumed. Not product-only; the commit is +36/−4 with 22 of those lines being test.

### 3. Mutation — per hunk, all applied, all killed

Pristine `GSharp.Core.dll` md5 `18038f2a…`; each mutant verified applied by `git diff` **and** md5 movement, then restored by exact md5 round-trip (`BoundBinaryOperator.cs` → `5942c8c5…`, `Evaluator.Operators.cs` → `381f9d50…`, both confirmed).

| mutant | hunk | Core md5 | BUILD_RC | result |
|---|---|---|---|---|
| **M-REF** — `isReferenceEquality: true` → `false` | `BoundBinaryOperator.cs:380` | `a80816c3` | 0 | **RED 1/6** |
| **M-EQARM** — `==` arm → `NumericEquals` | `Evaluator.Operators.cs:177` | `782b491d` | 0 | **RED 1/6** |
| **M-NEARM** — `!=` arm → `!NumericEquals` | `Evaluator.Operators.cs:179` | `8b14b0c2` | 0 | **RED 1/6** |
| **M-CTRL** (control, expected to kill) — invert condition, both arms | `:177`+`:179` | `d0833381` | 0 | **RED 2/6** |

All four die on `Issue3134_ClassIdentityAndStructValueEqualityMatchEmitter(surface: "operatorEquals")`. M-REF's failure text is the fix reverting exactly: `Expected: "ClassWithOverrides|False|True|False|True|"` / `Actual: "ClassWithOverrides|True|True|True|False|F"`. **The `==` and `!=` arms are independently load-bearing** — M-EQARM and M-NEARM each kill alone, so neither is riding on the other. Four live kills, so harness liveness is demonstrated four times over rather than resting on the control.

### 4. It did not over-correct

The guard is not in the flag, it is in *where the flag is set*: `MakeReferenceEquality` is pre-existing (#2188) and is bound **only as a last-resort fallback**, after built-in, user-defined and CLR `op_*` resolution (`ExpressionBinder.Operators.cs:1550-1560`). So anything with a real `operator ==` never reaches it. Measured, `r2with` vs `r2without`, all three drivers, all **unchanged and all agreeing with emit**:

- `DataClass` — `True|True|False`, value equality retained.
- Imported `Version` (has *both* an `Equals` override and `operator ==`) — `ImportedVersion|True|True`, byte-identical to my round-1 baseline. The user-defined operator still wins.
- Imported `string` homogeneous, on two distinct runtime-computed strings (`String.Concat("he","llo")` vs `String.Concat("hel","lo")`) — `HomoString|True|False`. Value equality retained; **not** collapsed to reference identity.
- `string?` vs `string`, interface-vs-class, and interface-vs-interface on the same instance — all unchanged.

I specifically went looking for the collapse-to-reference-identity failure because `IsReferenceEqualityOperand` explicitly admits `string`, `object` and imported classes. It does not happen, because those pairs are resolved earlier by the built-in table.

### 5. No regression

Across `eq`/`h1`/`h2`/`x2` on both trees, exactly two cells changed — `ClassWithOverrides` and its `object`-free equivalent — and both changed from *disagreeing with emit* to *agreeing with emit*. Every other cell is byte-identical between `r2without` and `r2with`.

---

### SHOULD-FIX 1 — `object`-typed operands still diverge from emit (pre-existing, **not** introduced here)

Same defect class as the one being fixed, one type annotation away. Measured identically on **both** trees, so this commit neither causes nor worsens it:

| expression | bare | **emit (oracle)** | gsi |
|---|---|---|---|
| `object` = computed `"hello"` vs `object` = computed `"hello"` | `True` | **`False`** | `True` |
| `object` = class instance vs `object` = distinct instance | `True` | **`False`** | `True` |
| `object` = boxed `5` vs `object` = boxed `5` | `True` | **`False`** | `True` |
| `object o = s1; o == s2` (string) | `True` | **`False`** | `True` |
| class-typed, same instances *(what this commit fixed)* | `False` | `False` ✓ | `False` |

Emit is right in every row — `object == object` is reference identity, and two separate boxes of `5` are not the same object. The evaluator returns value equality because these pairs bind through the built-in table's `object`/homogeneous arm, which is **not** routed through `MakeReferenceEquality` and so never receives the new `IsReferenceEquality` flag. `Evaluator.Operators.cs:177` therefore still falls to `NumericEquals` for them.

**This should not block.** It is pre-existing, it is outside the commit's scope, and the commit is a strict improvement. But the commit's subject line — "Preserve evaluator reference equality operators" — reads as if reference-equality semantics are now general in the evaluator, and they are not: they hold for user-class-typed operands and not for `object`-typed ones. Worth filing so the claim and the coverage match, and it is a natural second `InlineData` row on the surface just added.

### NIT 1 — two vacuous rows on the new surface, but legitimately so

`operatorEquals` emits `N/A` for `StructWithOverrides`/`StructWithoutOverrides`. I checked whether that was a dodge: it is not. `a == b` on a struct is **`GS0129: Binary operator '==' is not defined for types 'SWO' and 'SWO'`** — the language rejects it, so there is no behaviour to assert. The abstention is correct. The only thing left on the table is that nothing pins the `GS0129` itself, so if struct `==` were ever added these rows would silently stay `N/A`.

---

### Coverage — 9 of 9, the round-1 gap closed

Denominator **9**, from `dotnet sln GSharp.sln list` filtered to `.tests.csproj`; `build/generate-ci-test-matrix.py` makes the identical call with the identical filter.

|---|---|---|---|---|
| **Compiler.Tests** | **0** | **0** | **4245** | 0 |
| Core.Tests | 0 | 0 | 7127 | 1 |
| InternalAnalyzers.Tests | 0 | 0 | 7 | 0 |
| Interpreter.Tests | 0 | 0 | 1230 | 0 |

**15,173 tests, 0 failures, 1 skipped, 9 of 9 projects.** `Compiler.Tests` — the suite I flagged last round as an unrun gap rather than a pass — ran in full, unfiltered, **1h17m, 4245 passed, RC=0**. It is squarely in this commit's blast radius because `BoundBinaryOperator.cs` is shared binding code, and it is green. Interpreter.Tests is unfiltered green at 1230 (1225 last round; +1 is the new `operatorEquals` row).

### Disclosures

- **The `main` SHA in my assignment was stale on arrival.** I was given `387cb9a2`; `git fetch` showed `d957ba71`, four commits later, and `387cb9a2` an ancestor. I anchored to the fetched SHA. It also happens to be the new merge-base, because the author rebased onto it — a coincidence I checked rather than assumed.
- **My first probe was invalid and I nearly scored it.** `h1.gs` failed with `GS0368` (bodyless interface `func` needs `;`) on all three drivers. Three identical rc=1 columns look superficially like a uniform result; it was a syntax error in my own file. Fixed and re-run. Likewise my struct probe first died on `GS0130` because I constructed structs as `SWO()` instead of `SWO{}` — I only concluded `N/A` was legitimate *after* getting the program to compile and seeing the real `GS0129`. A failing probe is not evidence until it fails for the reason you think.
- **Load** at each build start: 14.97 / 18.38 / 17.76 / 17.23 / 14.49 — all below 30.
- Every probe ran in its own asserted-empty directory; all mutants restored by exact md5 round-trip; both scratch and shared checkouts end clean (`git status --porcelain` empty). I did not write into any worktree I do not own.

Issue #708, opened 2026-06-11
Add `if let` and `guard let` bindings


DavidObando:
## Summary

Add `if let x = expr { … }` and `guard let x = expr else { … }` bindings as a left-hand-side spelling for the existing nullable-narrowing flow.

## Motivation

Today the only way to extract a non-null value from a nullable binding is:

```
let x = maybe
if x != nil {
    // x is smart-cast to non-null here
    use(x)
}
```

`if let` / `guard let` are the standard form in Swift, and Kotlin offers similar idioms via `let { }` + Elvis. They consolidate the declaration, the null check, and the narrowed scope into one expression.

## Scope

- `if let name = expr { … } [else { … }]` — `expr` must be of nullable type `T?`; `name` is bound as `T` inside the `then` block.
- `guard let name = expr else { … }` — the `else` block must exit the enclosing scope (`return`, `throw`, `break`, `continue`); `name` is bound as `T` in the remainder of the enclosing scope.
- Multiple bindings: `if let a = ea, let b = eb { … }` (Swift-style).
- Interaction with the existing smart-cast / flow-narrowing machinery (ADR-0069).

## Acceptance criteria

- Parser, binder, lowering, emit, interpreter.
- Diagnostics for misuse (non-nullable RHS, missing exit in `guard let` else).
- Sample + golden coverage.
- Spec + tour updated.

## Related

- Cross-references existing tracking issue #696 ("Consider `if let` statements"). This issue is the planned implementation; #696 can be closed in favor of this one once work starts.

---

**Plan reference:** §6.2 item 4. Tracked under #706 (G# Language — Current State and Design Opportunities, §6). (supersedes #696)

PR #3155, opened 2026-08-03
Test GS0159 suggested remedies

### LINKED ISSUE #3142: GS0159 tests pin that the message is printed, never that its suggested remedy compiles
## Summary

PR #3054 (merged, issue #3034) improved the `GS0159` message so it names the actual cause of an invalidated nullable narrowing and **suggests a remedy** — `?.` or `if let`. A rubber-duck verified the advice is **true today**: all four previously unprobed receiver kinds (interface, enum, constrained type parameter, extension) accept both remedies, rc=0, on all three drivers.

But **no test holds that**. All eight `?.` / `if let` occurrences in the test file live **inside asserted strings**. Zero tests compile the suggested remedy.

## The gap

> The suite pins that the message is **printed**. Nothing pins that the message is **true**.

The diagnostic promises the user an action. Nothing keeps the promise honest. If a future change makes `?.` stop working on one of these receiver kinds, `GS0159` will keep confidently recommending it and every test will keep passing, because the tests compare strings.

This is a **durability** gap, not a correctness gap — which is why it was graded SHOULD-FIX and #3054 merged. Filing so it does not evaporate.

## Why it is worth closing

A diagnostic that suggests a fix is strictly better than one that does not — but only while the suggestion works. A stale suggestion is worse than no suggestion: it sends the user to write code that does not compile, and they will trust it because the compiler said it.

This is also the **third** bug this session in the same shape — an assertion that checks the wrong thing:
- #3140: `out` write-back tests assert diagnostics, never values → silent wrong answer invisible
- #3133 (pre-fix): deinit tests asserted with `Assert.All` over a possibly-empty set → vacuously green
- this one: message tests assert the string, never the advice

**A test that asserts the observable symptom rather than the underlying claim will pass through the bug it was written to catch.**

## Suggested approach

For each receiver kind `GS0159` can fire on, add a companion test that **compiles the remedy the message recommends** and asserts rc=0. Derive the remedy from the message rather than hardcoding it in parallel, so the two cannot drift — if the message text is the source of truth, a changed message forces a changed test.

Verified receiver kinds where both remedies work today, from the #3054 review: **interface, enum, constrained type parameter, extension**, plus the previously covered cases. The exclusion set for narrowing is **three** — field, parameter, `let` — `global` and `local` both narrow successfully. That set was measured on `27a05798`; re-measure rather than trusting it.

## Bar

Three driver columns: bare `gsc` **evaluates**, `gsc /out:` **emits**, `gsi` **interprets**. `GS0159` is bind-time so all three should agree — any divergence is itself a finding. Compare on **rc and diagnostic set**, not printed values.

Only PRODUCT mutants count; condition inversion, never `if (true)`; mutate per hunk, never per PR; a no-result is not a survival.

Full suite unfiltered. Note `Cs2Gs.Tests` is the **8th test project** and invokes real `gsc.dll` out-of-process, so a pure `src/` change can redden the cs2gs gates — count it in "all suites".

DavidObando:
Fixes #3142.

## Summary
- compile both remedies recommended by nullable-receiver GS0159 for local, parameter, field, global, `let`, struct, interface, enum, constrained-type-parameter, and receiver-clause extension receivers
- derive `?.` and `if let` from the emitted message instead of duplicating its guidance in test data
- require successful emit and preserve every non-GS0159 diagnostic

## Scope
This covers the nullable-receiver overload of `ReportUnableToFindFunction`, the GS0159 form that contains actionable advice. Generic lookup failures use `Cannot find function X.` and contain no suggestion.

That distinction matters for PR #3147. Re-measured its current head `70833ef9`: all three drivers return rc0 for `GsharpExtensionsOptional.gs`. At its base `faca348a`, all three return rc1 GS0159, but every message is only `Cannot find function Map/FlatMap/...`; none recommends `?.`, `if let`, a reference, or a flag. #3147 therefore does not exercise a stale GS0159 suggestion. Adding reference guidance to generic lookup failure would be separate diagnostic work.

## Verification
- merged-tree Release warnings-as-errors build: rc0, 0 warnings/errors
- focused GS0159 tests: 43/43
- bare `gsc`, `gsc /out:`, `gsi`: 90/90 receiver/driver/shape cells
- main-without-PR control (`faca348a`): same 90/90 product matrix green
- corrupted suggestion product mutant (`?.` → nonexistent `.?`): build rc0, remedy test 0/10; all ten cases red
- control mutants: nullable null-safe binding 0/10 and `if let` narrowing frame 0/10; both mutant builds green
- 8 of 9 test projects, enumerated by `dotnet sln GSharp.sln list`: Compiler 4236, Core 7113 + 1 skipped, Interpreter 1214, LanguageServer 462, Extensions 104, InternalAnalyzers 7, GeneratorHost 31, Cs2Gs 1903
- `Sdk.Tests` NOT RUN: it writes the shared local SDK package cache, prohibited for concurrent local validation

Skipped the optional `IsApplicableNullableUnderlyingCall` rewrite: recursive binder re-entry risks duplicated argument binding/state and is unrelated to this test-only fix.

### DIFF
--- a/test/Compiler.Tests/Binding/Issue3034NullableNarrowingDiagnosticTests.cs
+++ b/test/Compiler.Tests/Binding/Issue3034NullableNarrowingDiagnosticTests.cs
@@ -61,6 +61,43 @@ public class Issue3034NullableNarrowingDiagnosticTests
         Assert.DoesNotContain("re-narrow", diagnostic.Message, StringComparison.Ordinal);
     }
 
+    [Theory]
+    [MemberData(nameof(GuidanceRemedyCases))]
+    public void NullableReceiverGuidance_SuggestedRemediesCompile(
+        string receiverKind,
+        string sourceTemplate,
+        string receiver,
+        string call,
+        bool isLibrary)
+    {
+        var diagnosticSource = sourceTemplate.Replace("/*CALL*/", $"{receiver}.{call}", StringComparison.Ordinal);
+        var diagnosticResult = Compile(diagnosticSource, isLibrary);
+        var diagnostic = Assert.Single(diagnosticResult.Diagnostics, diagnostic => diagnostic.Id == "GS0159");
+        var expectedDiagnosticIds = diagnosticResult.Diagnostics
+            .Where(diagnostic => diagnostic.Id != "GS0159")
+            .Select(diagnostic => diagnostic.Id)
+            .Order()
+            .ToArray();
+        var (nullSafeOperator, bindingKeyword) = GetSuggestedRemedies(diagnostic.Message);
+        var remedies = new[]
+        {
+            (Name: nullSafeOperator, Code: $"{receiver}{nullSafeOperator}{call}"),
+            (Name: bindingKeyword, Code: $"{bindingKeyword} narrowed = {receiver} {{\n    narrowed.{call}\n}}"),
+        };
+
+        foreach (var remedy in remedies)
+        {
+            var source = sourceTemplate.Replace("/*CALL*/", remedy.Code, StringComparison.Ordinal);
+            var result = Compile(source, isLibrary);
+            Assert.True(
+                result.Success,
+                $"{receiverKind}/{remedy.Name}: {string.Join(Environment.NewLine, result.Diagnostics.Select(d => $"{d.Id}: {d.Message}"))}");
+            Assert.Equal(
+                expectedDiagnosticIds,
+                result.Diagnostics.Select(diagnostic => diagnostic.Id).Order());
+        }
+    }
+
     [Fact]
     public void ExpiredBranchNarrowing_ExplainsNeedToRenarrow()
     {
@@ -358,14 +395,15 @@ public class Issue3034NullableNarrowingDiagnosticTests
         => Assert.Single(GetDiagnostics(source), diagnostic => diagnostic.Id == "GS0159");
 
     private static Diagnostic[] GetDiagnostics(string source, bool isLibrary = true)
+        => Compile(source, isLibrary).Diagnostics.ToArray();
+
+    private static EmitResult Compile(string source, bool isLibrary)
     {
         var sourceText = SourceText.From(source, "issue3034.gs");
         var tree = SyntaxTree.Parse(sourceText);
         var compilation = new Compilation(ReferenceResolver.Default(), tree) { IsLibrary = isLibrary };
-
         using var peStream = new MemoryStream();
-        var result = compilation.Emit(peStream, refStream: null);
-        return result.Diagnostics.ToArray();
+        return compilation.Emit(peStream, refStream: null);
     }
 
     public static TheoryData<string, string, string, bool> ReceiverKindCases => new()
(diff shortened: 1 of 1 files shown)


DavidObando:
## MERGEABLE — the test pins the product claim, not the fixture. I tried three ways to break that conclusion and failed.

Anchors: `origin/main` **`61fe9cb9`**, head **`6aa4f521`**, merge base **`faca348a`** (main ahead by 1). `git merge-tree --write-tree` **output** = `26c108a8`, **0 conflict markers**; co-merging in a fresh worktree reproduced that tree hash exactly. Merged-tree Release build from a tree with **no `out/`**: rc0, **0 errors, 0 warnings**. Build-output freshness (epoch, `sort -n`, excluding `ref/`): `Core`/`Compiler`/`Repl` all `1785783187`, ~98 s before measurement.

---

## The framing dispute: **you are right, and the author's correction is right too**

Settled two independent ways.

**Call-site census** — `ReportUnableToFindFunction` has two overloads sharing descriptor `UnableToFindFunction`:

| site | message | suggestion |
|---|---|---|
| `DiagnosticBag.Reports.Expressions.cs:319-320` | `Cannot find function {text}.` | **none** |
| `DiagnosticBag.Reports.Expressions.cs:326-333` | `Cannot call function {f} because receiver '{r}' may be nil. Use '?.' … or bind it with 'if let'.` | **yes** |

Of **6** call sites, exactly **one** — `ExpressionBinder.Calls.Invocation.cs:2778` — passes the receiver and gets the suggestion form. The other five (`:2169`, `:2479`, `:2782`, `:3079`, `Calls.Arguments.cs:809`) take the bare form.

**Direct measurement** (bare `gsc`, this head):
```
c.NoSuchMethod()   -> GS0159: Cannot find function NoSuchMethod.                    suggestion-count 0
c.M() on C?        -> GS0159: Cannot call ... Use '?.' ... or bind it with 'if let'. suggestion-count 1
```

So the scope statement in the PR body is **exact**: there is one suggestion-bearing form, and it is the one covered. **Census: 1 of 1.** The #3147 example does not exercise a suggestion — but the general premise ("GS0159 carries advice that nothing pinned") was sound, and this PR is the right response to it.

---

## "Does it test the product, or only that the fixture parses?" — it tests the product

The author's mutant (`'?.'` → `'.?'`) reproduced exactly: **0 passed / 10 failed**. But that one *is* vulnerable to your objection, since `.?` is a syntax garble.

So I ran a **stronger mutant they did not**: `'if let'` → `'guard let'` at `DiagnosticBag.Reports.Expressions.cs:333`. `guard let` is a **real G# narrowing construct** (ADR-0071 / #708), not a garble — plausible advice that happens to be wrong for this shape.

| mutant | site | result |
|---|---|---|
| `'?.'` → `'.?'` (author's) | `:333` | **0/10** — killed |
| **`'if let'` → `'guard let'` (mine, semantic)** | `:333` | **0/10 — killed** |
| flip-back to pristine | — | **43/43** green |

It discriminates between two genuine language constructs. That is a product assertion, not a fixture assertion.

I also enumerated what could occupy the derived `?.` slot at all. Of 7 candidates, **exactly one compiles**:
```
'?.' rc=0   '.' rc=1 GS0159   '!.' rc=1 GS0157   '->' rc=1 GS0304
'::' rc=1 GS0005   '?->' rc=1 GS0155   '&.' rc=1 GS0157
```
There is no `!.` operator in G#, so no "compiles but is unsafe" substitution exists. For that slot **"compiles" ≡ "is exactly `?.`"** — the compile-only assertion is maximally strong, and adding execution there would duplicate coverage rather than add it.

---

## Where the new coverage genuinely stops (reported for the boundary, not as a defect)

I built a PRODUCT mutant that keeps compilation succeeding but destroys the advertised property: `MethodBodyEmitter.Expressions.cs:1600`, `Brtrue` → `Brfalse` (stack-balanced, inverts the null guard). Emit rc0; the new theory **passed 10/10 — survived**. Real behaviour change confirmed out-of-process:
```
Try(nil) -> Unhandled exception. System.NullReferenceException  (run-rc=134)
```
**This is not a defect in this PR.** It is a defect in `?.` lowering, out of #3142's scope, and the wider suite catches it hard — `--filter NullConditional` went **56 failed / 6 passed**. Worth recording only as the honest boundary: the new tests pin *the advice*, and `?.` semantics are pinned elsewhere.

One methodology note for anyone reusing my probe: my first attempt used `var c C? = nil` and showed **no** behaviour change, because a statically-nil local is folded away. Only a **parameter** receiver — the PR's own `"parameter"` case — exposes it. A null result there would have been a harness artifact, not a survival.

---

## Census or sample: sample, and I could not find a broken cell in it

10 receiver kinds is a sample. Three uncovered kinds, measured on this head:

| uncovered kind | direct call | `?.` | `if let` |
|---|---|---|---|
| slice element `arr[0]` | GS0159 + suggestion | rc0 | rc0 |
| nested field `o.I.V` | GS0159 + suggestion | rc0 | rc0 |
| call result `Get()` | GS0159 + suggestion | rc0 | rc0 |

All three produce the suggestion, and both remedies compile in all three. **Negative result properly established: the sample is representative; no uncovered receiver kind receives un-actionable advice.**

## Driver uniformity — genuine, but not repo coverage

Your instinct to probe the all-identical table was right; the answer is that it is real. Independent 3-driver spot check on this head:
```
parameter/direct : eval=1 interp=1 emit=1, GS0159 in all three
parameter/'?.'   : eval=0 interp=0 emit=0
interface/'if let': eval=0 interp=0 emit=0
```
Uniform because the diagnostic is binder-produced and all three drivers share the binder. And the harness **can** be non-uniform — under my emit mutant, emit diverged from evaluate/interpret, which only emit's IL path can do.

**NIT:** the 90/90 matrix is a manual probe; it is not in the tree. The committed test is in-process `compilation.Emit` at `Issue3034NullableNarrowingDiagnosticTests.cs:406` — no process, no driver, no execution — and `test/Interpreter.Tests/` was not touched. Low risk given the shared binder, but the PR description promises more durable coverage than the repo now holds.

---

## Suites — 9 of 9, enumerated from `dotnet sln GSharp.sln list`

| project | result |
|---|---|
| Compiler.Tests | **4242** passed |
| Core.Tests | 7116 passed, **1 skipped** (7117) |
| Interpreter.Tests | 1220 passed |
| Cs2Gs.Tests | 1903 passed |
| **Sdk.Tests** | **59 passed** |
| GeneratorHost.Tests | 31 passed |

**15,145 tests, 1 skipped, 0 failures.**

**`Sdk.Tests` is in bounds and I ran it.** The "shared package-cache race" was a bad standing instruction, not an author error. Verified structurally before running: 8 `.cs` files, no `Process.Start`, no `ProcessStartInfo`, no `.nuget` write — the only `dotnet build` string is prose inside a `<para>` docstring describing *other* acceptance tests. Please drop that line from the PR body.

**NIT:** I measured Compiler.Tests at **4242** on the merged tree vs **4236** reported on the unmerged head, and Core/Interpreter similarly a few higher. Consistent with main's extra commit; all green either way, so I did not chase the delta.

---

### Verdict: **MERGEABLE**. No blockers.

The design choice that makes this work is deriving the remedy from the emitted message rather than hardcoding it (`GetSuggestedRemedies`), so the assertion moves when the advice moves. That is the difference between pinning the claim and pinning the symptom, and it is the right shape for this defect class.

PR #3156, opened 2026-08-03
Fix instance out parameter write-back

### LINKED ISSUE #3140: out parameter write-back silently dropped for class instance methods under gsi and bare gsc
## Summary

An `out` parameter's write-back is **silently dropped when the callee is a class instance method**, under both evaluating drivers. The call returns, rc is 0, no diagnostic is produced, and the caller's variable still holds its initial value.

Found by a rubber-duck while reviewing PR #3070. **Pre-existing and live on `main` today** — measured identically on that PR's base and head, so it is not caused by #3070.

## Measured

Callee assigns the `out` parameter, caller prints it:

| probe | shape | bare `gsc` (evaluate) | `gsc /out:` (emit) | `gsi` |
|---|---|---:|---:|---:|
| `v1_instance_out` | class **instance** method, `r=55` | **0** | **55** | **0** |
| `v2_shared_out` | class **shared** method, `r=66` | 66 | 66 | 66 |
| `v3_free_out` | **free** function, `r=77` | 77 | 77 | 77 |

Emit is the oracle and it is right in all three rows. The evaluating drivers are right in two of three. **Only the class-instance receiver loses the write-back.**

`v2` and `v3` are the controls: they isolate the defect to the instance receiver specifically, rather than to `out` handling in general or to the evaluating drivers in general.

## Why this is bad

rc 0, no diagnostic, **wrong value**. The program runs to completion and quietly computes with a stale variable. This is the worst failure shape in the compiler — worse than a crash and worse than a bogus diagnostic — because nothing signals that anything went wrong. It is the same shape as #3116 and #3134.

Two of three drivers are affected, so it is not an obscure corner: any user running `gsi` or bare `gsc` over code that calls an instance method with an `out` parameter gets a wrong answer, and gets a *different* answer than if they had emitted the same program.

## Why the existing tests do not catch it

#3070 added 1,142 lines of coverage for exactly this area — `out` parameter definite assignment for functions declared in class bodies — and it does not catch this. The reason is structural and worth stating plainly:

> **#3070's tests assert diagnostics, not values.**

They check *whether* `GS0238` is reported. They never check *what the program computed*. So a driver that produces no diagnostic and the wrong number passes every one of them. The analysis being tested is a bind-time analysis, so asserting on rc and diagnostic set was the right call for that PR — the gap is that nothing else asserts the runtime value.

This is the third bug in a row found in an axis the harness never generated. **The general lesson: a diagnostic-only assertion cannot detect a silent wrong answer, and the two need separate coverage.**

## Bar

Three driver columns: bare `gsc` **evaluates**, `gsc /out:` **emits**, `gsi` **interprets**. **Emit is the oracle** — compare evaluating output against emitted output cell by cell, not against hardcoded expectations, so new shapes cannot silently pass.

Cover the receiver axis, since that is where the defect lives: `{free function, class shared, class instance, struct shared, struct instance, interface method, base-class method, generic method}` × `{single out, multiple out, out + ref, out + normal, out assigned on some paths}`.

**Assert values, not only diagnostics.** That is the specific gap that hid this bug. Every cell must compare the computed result.

Use distinct, non-default probe values — `0` is this bug's own symptom, so a probe that expects `0` cannot distinguish success from failure. **A zero can be scope, not absence.**

Run every probe from its own EMPTY directory, asserted empty. Full suite unfiltered — a filtered green is not a green.

Note `out` in G# is spelled `out result int32`, **not** `&`. A reviewer's first out-parameter count came back 0 using an `&` pattern and only a positive control caught it.

DavidObando:
## Summary

- route user instance and base-class calls through existing evaluator ref/out slot handling
- copy final callee parameter slots back to caller operands after restoring caller frame
- add value-based emit-oracle parity matrix across receiver and argument shapes

Fixes #3140

## Root cause

Free functions already built ref/out slots and wrote final parameter values back after popping the callee frame. User instance and base-class call evaluators instead evaluated every argument by value, populated callee frames, and returned without write-back.

Fix reuses existing `BuildRefSlots` and `WriteBackRefSlots` paths. No parallel copy-back implementation.

Compiler-count audit: temporary required parameter on `EvaluateUserMethodBody` produced 7 unique `CS7036` invocation sites. Runtime-relevant user member paths fixed here: user-instance and base-class. Free calls already worked; static/shared calls bind through free-call path; interface calls dispatch through user-instance path.

## Receiver × shape matrix

Every cell compares bare `gsc` and `gsi` output to emitted execution, not hardcoded expected output. Each driver runs in its own asserted-empty directory.

| receiver | single out | multiple out | out + ref | out + normal | conditional out |
|---|---|---|---|---|---|
| free | `11` | `21,22` | `31,37` | `44` | `51,52` |
| class shared | `11` | `21,22` | `31,37` | `44` | `51,52` |
| class instance | `11` | `21,22` | `31,37` | `44` | `51,52` |
| struct shared | `11` | `21,22` | `31,37` | `44` | `51,52` |
| struct instance | `11` | `21,22` | `31,37` | `44` | `51,52` |
| interface method | `11` | `21,22` | `31,37` | `44` | `51,52` |
| base-class method | `11` | `21,22` | `31,37` | `44` | `51,52` |
| generic instance method | `11` | `21,22` | `31,37` | `44` | `51,52` |

All 40 cells: bare `gsc` = emit = `gsi`, rc 0.

Original issue probes after fix:

| shape | bare `gsc` | emit | `gsi` |
|---|---:|---:|---:|
| class instance | 55 | 55 | 55 |
| class shared control | 66 | 66 | 66 |
| free control | 77 | 77 | 77 |

## Two-sided mutation evidence

Tests on `main` without production fix: **25 failed / 15 passed**. Failures are every instance-dispatch receiver row; shared/free controls remain green.

| hunk | product mutant | result |
|---|---|---|
| user-instance write-back | invert write-back guard (`refSlots != null` → `refSlots == null`) | 25 failed / 15 passed |
| base-class write-back | invert write-back guard | 5 failed / 35 passed |
| shared frame/ref-slot helper | invert ref-kind allocation predicate | 25 failed / 15 passed |

Each mutant applied and built. Restored matrix: 40/40 green.

## Validation

Validated against `main` `faca348a36e7c4f9f35a3f57bde43c47f958b603`.

- full unfiltered suite: 15,152 passed, 1 skipped, 0 failed
- includes `Cs2Gs.Tests`: 1,903 passed
- full CIB solution build: 0 warnings, 0 errors
- Core/Compiler/Repl `GSharp.Core.dll` mtimes identical
- `git merge-tree --write-tree origin/main HEAD`: clean; merged tree `8fd4dad0c6ccdbd1b1c0f29abba6523f8b992e36`, identical to HEAD tree

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -345,20 +345,19 @@ public sealed partial class Evaluator
         };
 
         var parameterOffset = method.ExplicitReceiverParameter == null ? 0 : 1;
-        for (int i = 0; i < node.Arguments.Length; i++)
-        {
-            var parameter = method.Parameters[i + parameterOffset];
-            var value = EvaluateExpression(node.Arguments[i]);
-            frame[parameter] = value;
-        }
+        var args = PopulateUserCallFrame(node.Arguments, method, parameterOffset, frame, out var refSlots);
 
         RegisterRuntimeTypeArguments(frame, method.TypeParameters, node.MethodTypeArguments);
         RegisterRuntimeTypeArguments(frame, node.Receiver.Type);
+        object result;
         using (PushFrame(frame))
         {
             var statement = program.Functions[method];
-            return EvaluateUserMethodBody(method, statement);
+            result = EvaluateUserMethodBody(method, statement);
         }
+
+        WriteBackUserCallFrame(refSlots, args, method, parameterOffset, frame);
+        return result;
     }
 
     /// <summary>
@@ -446,18 +445,17 @@ public sealed partial class Evaluator
         };
 
         var parameterOffset = method.ExplicitReceiverParameter == null ? 0 : 1;
-        for (int i = 0; i < node.Arguments.Length; i++)
-        {
-            var parameter = method.Parameters[i + parameterOffset];
-            var value = EvaluateExpression(node.Arguments[i]);
-            frame[parameter] = value;
-        }
+        var args = PopulateUserCallFrame(node.Arguments, method, parameterOffset, frame, out var refSlots);
 
+        object result;
         using (PushFrame(frame))
         {
             var statement = program.Functions[method];
-            return EvaluateUserMethodBody(method, statement);
+            result = EvaluateUserMethodBody(method, statement);
         }
+
+        WriteBackUserCallFrame(refSlots, args, method, parameterOffset, frame);
+        return result;
     }
 
     /// <summary>
@@ -1474,6 +1472,55 @@ public sealed partial class Evaluator
         }
 
         return refSlots;
+    }
+
+    private object[] PopulateUserCallFrame(
+        ImmutableArray<BoundExpression> arguments,
+        FunctionSymbol function,
+        int parameterOffset,
+        ConcurrentDictionary<VariableSymbol, object> frame,
+        out List<(int Index, BoundExpression Operand)> refSlots)
+    {
+        var refKinds = function.Parameters
+            .Skip(parameterOffset)
+            .Take(arguments.Length)
+            .Select(parameter => parameter.RefKind)
+            .ToImmutableArray();
+        var args = refKinds.Any(refKind => refKind != RefKind.None)
+            ? new object[arguments.Length]
+            : null;
+        refSlots = args == null
+            ? null
+            : BuildRefSlots(arguments, refKinds, args, method: null);
+
+        for (var i = 0; i < arguments.Length; i++)
+        {
+            frame[function.Parameters[i + parameterOffset]] = args == null
+                ? EvaluateExpression(arguments[i])
+                : args[i];
+        }
+
+        return args;
+    }
+
+    private void WriteBackUserCallFrame(
+        List<(int Index, BoundExpression Operand)> refSlots,
+        object[] args,
+        FunctionSymbol function,
+        int parameterOffset,
+        ConcurrentDictionary<VariableSymbol, object> frame)
+    {
+        if (refSlots == null)
+        {
+            return;
+        }
+
+        foreach (var (index, _) in refSlots)
+        {
+            args[index] = frame[function.Parameters[index + parameterOffset]];
+        }
+
+        WriteBackRefSlots(refSlots, args);
     }
 
     private bool IsBoundLvalue(BoundExpression operand)
--- a/test/Interpreter.Tests/Issue3140OutParameterWriteBackParityTests.cs
+++ b/test/Interpreter.Tests/Issue3140OutParameterWriteBackParityTests.cs
@@ -1,0 +1,370 @@
+// <copyright file="Issue3140OutParameterWriteBackParityTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace GSharp.Interpreter.Tests;
+
+/// <summary>Issue #3140: evaluated user-method ref/out writes must match emitted execution.</summary>
+[Collection("ConsoleIo")]
+public class Issue3140OutParameterWriteBackParityTests
+{
+    /// <summary>Gets receiver and argument-shape matrix rows.</summary>
+    /// <returns>Matrix rows.</returns>
+    public static IEnumerable<object[]> ReceiverShapeMatrix()
+    {
+        foreach (var receiver in Enum.GetValues<ReceiverKind>())
+        {
+            foreach (var shape in Enum.GetValues<ArgumentShape>())
+            {
+                yield return new object[] { receiver, shape };
+            }
+        }
+    }
+
+    [Theory]
+    [MemberData(nameof(ReceiverShapeMatrix))]
+    public async Task OutParameterWriteBack_GscEvaluateAndGsiMatchEmit(
+        ReceiverKind receiver,
+        ArgumentShape shape)
+    {
+        var source = BuildSource(receiver, shape);
+        var root = CreateEmptyDirectory(receiver + "-" + shape);
+        try
+        {
+            var emitted = await RunEmittedAsync(source, CreateEmptyDirectory(root, "emit"));
+            var evaluated = RunSourceDriver(
+                source,
+                CreateEmptyDirectory(root, "evaluate"),
+                GSharp.Compiler.Program.Main,
+                stripCompilerSuccess: true);
+            var interpreted = RunSourceDriver(
+                source,
+                CreateEmptyDirectory(root, "gsi"),
+                GSharp.Repl.Program.Main,
+                stripCompilerSuccess: false);
+
+            Assert.NotEmpty(emitted);
+            Assert.Equal(emitted, evaluated);
+            Assert.Equal(emitted, interpreted);
+        }
+        finally
+        {
+            DeleteDirectory(root);
+        }
+    }
+
+    /// <summary>Call receiver shape.</summary>
+    public enum ReceiverKind
+    {
+        /// <summary>Free function.</summary>
+        Free,
+
+        /// <summary>Class shared method.</summary>
+        ClassShared,
+
+        /// <summary>Class instance method.</summary>
+        ClassInstance,
+
+        /// <summary>Struct shared method.</summary>
+        StructShared,
+
+        /// <summary>Struct instance method.</summary>
+        StructInstance,
+
+        /// <summary>Class implementation called through interface.</summary>
+        Interface,
+
+        /// <summary>Base-class method called through <c>base</c>.</summary>
+        BaseClass,
+
+        /// <summary>Generic class instance method.</summary>
+        GenericMethod,
+    }
+
+    /// <summary>Argument shape.</summary>
+    public enum ArgumentShape
+    {
+        /// <summary>One out argument.</summary>
+        SingleOut,
+
+        /// <summary>Two out arguments.</summary>
+        MultipleOut,
+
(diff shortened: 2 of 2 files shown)


DavidObando:
## Rubber-duck review — PR #3156 (`fix/3140-work`), issue #3140

**Verdict: MERGEABLE.** 0 BLOCKER · 1 SHOULD-FIX · 2 NIT.

Reviewed at head `11ea703b`, merge-base `faca348a`, against `main` `61fe9cb9`. Probe merge tree = `26c35b5a`, reproduced locally. All work in scratch worktrees; no protected checkout or foreign worktree touched.

---

### The predicted trap was avoided — stated explicitly

This defect class exists *because* the pre-existing `out`-parameter tests assert diagnostics rather than values. The most likely way this PR could have been wrong is fixing the product and then pinning it with another diagnostic-shaped assertion. **It does not.**

`test/Interpreter.Tests/Issue3140OutParameterWriteBackParityTests.cs:56-58` is an oracle comparison of printed values:

```csharp
Assert.NotEmpty(emitted);
Assert.Equal(emitted, evaluated);
Assert.Equal(emitted, interpreted);
```

`Assert.NotEmpty` guards the vacuous all-empty pass. Emit is the oracle; evaluator and interpreter are compared against it. That is the right shape.

**The tests are genuinely effective, measured not assumed.** I copied the new test file alone into an unfixed BASE tree at `61fe9cb9` and rebuilt: **25 of 40 cases FAIL**. The failures are exactly `{ClassInstance, BaseClass, StructInstance, Interface, GenericMethod} × 5 argument shapes`; the 15 passes are `{Free, ClassShared, StructShared}` — precisely the receivers the issue reports as already correct. Base then restored to porcelain 0 and rebuilt.

Probe values are distinct and non-default (11/21/22/31/33/401/51/52) with locals pre-initialised to 101/201/…, so a dropped write-back shows a **stale** value rather than a default `0`. The issue's own warning is respected.

### Root cause, not symptom patch

Single production file, `src/Core/CodeAnalysis/Evaluator.Calls.cs` (+75/−14). It **reuses** the pre-existing `BuildRefSlots` (`:1423`) / `WriteBackRefSlots` helpers rather than inventing machinery, and adds two thin wrappers `PopulateUserCallFrame:1478` / `WriteBackUserCallFrame:1527`. The structural change — turning `return` inside `using (PushFrame(frame))` into `result = …` so write-back happens after the frame pops — is the minimum needed. Emit path untouched.

### Emit did not move — correct layer

| probe | shape | EMIT base | EMIT merged |
|---|---|---|---|
| `v1_instance` | class instance, single out | 55 | 55 |
| `u1_baseiface` | `base[IFiller].Fill(&x)` | 61 | 61 |
| `u3_constrained` | `T.Fill(&x)` constrained static | 81 | 81 |

Oracle stable. This was an evaluator bug and the fix stayed on the evaluator.

### Positive control — reported shape, three drivers

| tree | EMIT (oracle) | bare `gsc` EVAL | `gsi` |
|---|---|---|---|
| BASE `61fe9cb9` | 55 | **101** | **101** |
| MERGED `26c35b5a` | 55 | 55 | 55 |

---

## SHOULD-FIX — two sibling write-back paths still silently broken

Same defect class, same file, helper already written but not applied. Both reproduce **identically on BASE and MERGED**, so these are pre-existing and *not* a regression introduced here — but #3140 is the origin instance of this class, and leaving two known-broken cells in the file the PR is already editing is how the class survives to bite a seventh time.

| probe | shape | site | EMIT | EVAL | GSI |
|---|---|---|---|---|---|
| `u1_baseiface` | `base[IFiller].Fill(&first)` | `Evaluator.Calls.cs:374` `EvaluateBaseInterfaceCallExpression` | **61** | 101 | 101 |
| `u3_constrained` | `T.Fill(&first)` via `func Run[T IFiller](witness T)` | `Evaluator.Calls.cs:475` `EvaluateConstrainedStaticCallExpression` | **81** | 101 | 101 |

Silent wrong answers, rc=0, zero diagnostics — the exact signature of #3140.

Neither shape is in the matrix. `ReceiverKind` has `Interface` (call *through* an interface) and `BaseClass` (`base.M`), but **not** `base[IFoo].M` and **not** constrained-static. Recommend adding `BaseInterface` and `ConstrainedStatic` to `ReceiverKind` (`:65-91`) and applying the wrappers at `:374` and `:475`.

Caveat on `:475`: it is **not** a literal drop-in. `EvaluateConstrainedStaticCallExpression` pre-evaluates arguments into `evaluatedArgs` for runtime-type sniffing across three resolution strategies before populating its frame, so `refSlots` must be built separately rather than by calling `PopulateUserCallFrame` as-is.

---

## Mutation testing — per hunk, PRODUCT only, condition inversion, applied asserted

| mutant | site | BUILD_RC | tests killed | verdict |
|---|---|---|---|---|
| **M1 (control, expected to die)** | `WriteBackUserCallFrame` guard `refSlots == null` → `!= null`, `:1513` | 0 | **25 / 40** | control fires — a green here would have meant nothing |
| **M2** | instance-call write-back at `:359` wrapped in `if (refSlots == null)` | 0 | **25 / 40** | hunk pinned |
| **M3** | base-class write-back at `:457`, same inversion | 0 | **5 / 40** | hunk pinned — exactly the `BaseClass` row |

Both production hunks are independently pinned. M1's 25 dead match the unfixed-`main` failure set exactly, which cross-validates the control against an independent measurement.

**Disclosed non-survivals (retargets, not scores):** M2 needed two retargets before it built — first `SA1501` (statement on a single line), then `SA1513` (closing brace needs a following blank line). StyleCop is enforced as errors in `Core`. Neither was scored as a survival; a build failure is a no-result.

M3 killing 5 rather than 25 is correct, not weak: the outer call in the `BaseClass` fixture is itself an instance call, so M2 subsumes those rows while M3 isolates the `base.M` hunk alone.

---

## Coverage probes beyond the matrix

`WriteBackRefSlots` dispatches on `BoundVariableExpression` / `BoundFieldAccessExpression` / `BoundPropertyAccessExpression` / `BoundIndexExpression`, but every matrix cell passes a plain local (`&first`). I probed the other write-back targets through the newly-fixed instance path:

| target | BASE eval/gsi | MERGED eval/gsi | EMIT |
|---|---|---|---|
| `&h.f` (field) | 101 | **91** | 91 |
| `&arr[1]` (index) | 102 | **91** | 91 |

Both were broken on base and are fixed on merged. The fix covers them; the tests just do not assert them. **NIT**, not a defect — worth one extra `ArgumentShape` so a future refactor of `WriteBackRefSlots` cannot regress field/index targets unseen.

**Ruled out as gaps (reporting unfired hypotheses):**
- **Lambdas / `InvokeClosure`** — not a gap. `out` parameters in lambdas do not compile at all (`GS0155: Cannot convert type '(int32) -> void' to '(int32) -> void'`), so the path is unreachable.
- **Exceptional-path write-back** — I suspected `WriteBackRefSlots` appearing twice at `:895`/`:901` implied try/finally semantics worth testing. Unfired: they are two independent early-return branches (emulated `Enum.TryParse` vs reflective). No defect.

**UNPROBED, labelled as such:** `EvaluateExternalOverride:1060`, reached only from the imported instance-call path at `:995`, builds its `args` with plain `EvaluateExpression` — no ref slots, no write-back. Structurally the same shape as the two SHOULD-FIX sites. I did not construct a repro (it needs an imported `out` method externally overridden on a struct), so I assert nothing about it beyond the structural observation. Pre-existing on both trees either way.

---

## Test tally

Denominator source: `dotnet sln GSharp.sln list` → **9 projects**. Not `ls test/`, which returns 8 entries here (7 test projects plus `test/Shared/`) and is wrong in both directions.

All measured on the merged tree `26c35b5a`. Counts are from runs, not `--list-tests`.

| project | result |
|---|---|
| `Core.Tests` | 7116 passed / 0 failed / 1 skipped (7117 total) |
| `Compiler.Tests` | 4232 / 0 failed (1 h 25 m) |
| `Cs2Gs.Tests` | 1903 / 0 failed |
| `Interpreter.Tests` | 1260 / 0 failed (full, unfiltered) |
| `LanguageServer.Tests` | 462 / 0 failed |
| `Extensions.Tests` | 104 / 0 failed |
| `Sdk.Tests` | 59 / 0 failed |
| `GeneratorHost.Tests` | 31 / 0 failed |
| `InternalAnalyzers.Tests` | 7 / 0 failed |

**9 of 9 projects run. 15,175 passed, 0 failed, 1 skipped.** Nothing NOT RUN.

`Core.Tests` running 7117 while `--list-tests` reports 6,990 is reproduced here, consistent with the standing note that a list is a floor rather than a measurement.

---

## Self-disclosed errors in my own harness

- I first wrote the generic constraint as `func Run[T: IFiller]()`. Wrong syntax; produced a cascade of `GS0005` that I nearly read as a product defect. Correct form is `func Run[T IFiller](witness T)` with a witness parameter and inference at the call site.
- My first runner used `cd $D && gsc relative.gs`, which fails with "Unable to find specified file" and returns a **uniform rc=1 on both trees that looks exactly like real comparative data**. Fixed to absolute paths and re-ran everything.
- My first enumeration of frame-building methods used a crude regex and reported that `EvaluateCallExpression:34` lacked write-back. **That was a parser artifact** — it builds ref slots at `:66`. Corrected by reading the source; the free-function path is fine and the matrix independently confirms it green on base.

## Hygiene

Both scratch worktrees porcelain 0; every mutant restored byte-identical and rebuilt after restore. Merge tree identity `26c35b5a` reproduced. mtimes taken in epoch seconds with `sort -n` over `src/Core` + `src/Compiler` + `src/Repl`, excluding `ref/`; newest is `Evaluator.Calls.cs`, so no stale-binary risk. `rc` captured before any pipe. Protected checkout untouched and porcelain 0.

---

**MERGEABLE.** The fix is minimal, on the correct layer, reuses existing helpers, leaves emit byte-stable, and — unusually for this defect class — is pinned by value assertions that measurably go 25/40 red on unfixed `main`. The SHOULD-FIX is two pre-existing sibling cells of the same class in the same file, not a regression, and can ship as a follow-up if you would rather not widen this diff.

PR #3157, opened 2026-08-03
Make inherited deinit driver boundary observable

### LINKED ISSUE #3150: [test] Inherited-deinit driver-parity oracle is degenerate — emit column runs no deinit at all
Found by the #3133 round-2 review, graded NIT there and merged. Filing so it is not lost.

## What is wrong

The inherited-deinit driver-parity test looks like it pins deinit ordering across all three drivers. Measured, it does not.

`emitted.StandardOutput` is **`"body-44\n"` — 8 bytes**. Both deinit `WriteLine` calls **never run in any column**: `GC.KeepAlive` plus .NET's no-finalize-on-exit policy means the finalizers are never invoked.

Consequences, both measured:

1. **Deleting both deinit `WriteLine` bodies leaves the test passing.** The computed oracle is equal to the literal string it was introduced to replace, so replacing the frozen literal with a computed value bought nothing for this assertion.
2. The `CompilerEmission` row **asserts nothing new**: `result = emitted` is set at `:105-110`, so the compare at `:122` is object-to-self and `:123` merely repeats `:103`. No mutant can kill that row without also killing the other two.

## The trap worth naming

The central equation `Assert.Equal(emitted, result)` **holds only because the test shape makes emit run no deinit.** Make the deinits actually reachable — add `GC.Collect()` + `GC.WaitForPendingFinalizers()` — and the drivers diverge for real:

- emit: `body-44 / derived-deinit-22 / base-deinit-11 / end-55`
- `gsi`: `body-44 / end-55`

So fixing the oracle will make the current assertion **false**. That divergence is the actual finding and needs a decision: is interpreter-vs-emit deinit divergence intended behaviour to be pinned, or a defect?

## Not in scope

Inherited deinit **execution order** is already pinned — `Issue698DeinitEmitTests.cs:99` asserts `dIdx < bIdx` on the same shape. This issue is only about the parity oracle being vacuous.

## Standing rule this is another instance of

**A test asserting the observable symptom rather than the underlying claim will pass straight through the bug it was written to catch.** Third occurrence this session after #3140 and #3142.

DavidObando:
Fixes #3150

## Decision

Interpreter/emit deinitializer divergence is intentional, not an interpreter defect. ADR-0068 already specifies that `deinit` requires CLR GC finalization, runs only in emitted code, and is skipped by interpreted execution with GS0510. Scope-exit execution would finalize reachable objects; emulating CLR finalization would require evaluator re-entry from a finalizer thread.

## Change

- make the inherited-deinit probe create an unreachable object, force collection, and print an end marker
- assert the established boundary directly per driver instead of comparing every result to a vacuous emitted oracle
- run the emission row independently instead of comparing the emitted result to itself

Measured outputs:

| driver | stdout | diagnostic output |
|---|---|---|
| bare `gsc` evaluation | `body-44`, `end-55`, `Success.` | 2×GS0510 in stdout |
| `gsc /out:` emission | `body-44`, `derived-deinit-22`, `base-deinit-11`, `end-55` | none |
| `gsi` | `body-44`, `end-55` | 2×GS0510 in stderr |

## Mutation

Product mutant: remove `emitter.EmitBlock(body)` from `EmitClassDeinitializerBodyBytes`. Release solution build remained green (0 warnings/errors); matrix changed from 3/3 passing to 2/3 passing. `CompilerEmission` failed because output became `body-44/end-55`, proving deinit bodies are observed. Restored tree returned to 3/3 passing.

## Verification

Base/merged target: `faca348a36e7c4f9f35a3f57bde43c47f958b603`  
Head: `c47f82f816d29585140e811a0d824f5cf38d6b17`

- Core.Tests: 7116 passed, 1 skipped, 0 failed
- Compiler.Tests: 4232 passed, 0 failed
- Interpreter.Tests: 1214 passed, 0 failed
- LanguageServer.Tests: 462 passed, 0 failed
- Extensions.Tests: 104 passed, 0 failed
- InternalAnalyzers.Tests: 7 passed, 0 failed
- Cs2Gs.Tests: 1903 passed, 0 failed
- GSharp.GeneratorHost.Tests: 31 passed, 0 failed
- Sdk.Tests: not run; prohibited SDK-cache boundary
- `git merge-tree --write-tree`: clean, predicted tree `1aff1dea67974244033ef987250fde7d4a3ab0c0`

### DIFF
--- a/test/Interpreter.Tests/Issue2988DeinitInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue2988DeinitInterpreterTests.cs
@@ -68,7 +68,7 @@ public class Issue2988DeinitInterpreterTests
     [InlineData(Issue3010EntryPointDriverMatrixTests.Driver.CompilerEvaluation)]
     [InlineData(Issue3010EntryPointDriverMatrixTests.Driver.CompilerEmission)]
     [InlineData(Issue3010EntryPointDriverMatrixTests.Driver.Interpreter)]
-    public void InheritedDeinitializersWarnOncePerDeclaringClass(
+    public void InheritedDeinitializersExposeDriverBoundaryPerDeclaringClass(
         Issue3010EntryPointDriverMatrixTests.Driver driver)
     {
         const string Source = """
@@ -89,41 +89,41 @@ public class Issue2988DeinitInterpreterTests
                 }
             }
 
-            var resource = CachedResource("held-33")
-            Console.WriteLine("body-44")
-            GC.KeepAlive(resource)
+            func Allocate() {
+                var resource = CachedResource("held-33")
+                Console.WriteLine("body-44")
+                GC.KeepAlive(resource)
+            }
+
+            Allocate()
+            GC.Collect()
+            GC.WaitForPendingFinalizers()
+            Console.WriteLine("end-55")
             """;
+        const string InterpretedOutput = "body-44\nend-55\n";
+        const string EmittedOutput = "body-44\nderived-deinit-22\nbase-deinit-11\nend-55\n";
 
-        var emitted = Issue3010EntryPointDriverMatrixTests.Run(
-            "inherited-deinit-emitted",
+        var result = Issue3010EntryPointDriverMatrixTests.Run(
+            "inherited-deinit-" + driver,
             Source,
-            Issue3010EntryPointDriverMatrixTests.Driver.CompilerEmission);
-        Assert.Equal(0, emitted.ExitCode);
-        Assert.NotEmpty(emitted.StandardOutput);
-        Assert.Equal(string.Empty, emitted.StandardError);
-
-        var result = driver == Issue3010EntryPointDriverMatrixTests.Driver.CompilerEmission
-            ? emitted
-            : Issue3010EntryPointDriverMatrixTests.Run(
-                "inherited-deinit-" + driver,
-                Source,
-                driver);
+            driver);
         Assert.Equal(0, result.ExitCode);
 
         string warningOutput;
         switch (driver)
         {
             case Issue3010EntryPointDriverMatrixTests.Driver.CompilerEvaluation:
-                Assert.StartsWith(emitted.StandardOutput + "\n", result.StandardOutput, StringComparison.Ordinal);
+                Assert.StartsWith(InterpretedOutput + "\n", result.StandardOutput, StringComparison.Ordinal);
                 Assert.EndsWith("Success.\n", result.StandardOutput, StringComparison.Ordinal);
-                warningOutput = result.StandardOutput[emitted.StandardOutput.Length..];
+                Assert.Equal(string.Empty, result.StandardError);
+                warningOutput = result.StandardOutput[InterpretedOutput.Length..];
                 break;
             case Issue3010EntryPointDriverMatrixTests.Driver.CompilerEmission:
-                Assert.Equal(emitted.StandardOutput, result.StandardOutput);
+                Assert.Equal(EmittedOutput, result.StandardOutput);
                 Assert.Equal(string.Empty, result.StandardError);
                 return;
             case Issue3010EntryPointDriverMatrixTests.Driver.Interpreter:
-                Assert.Equal(emitted.StandardOutput, result.StandardOutput);
+                Assert.Equal(InterpretedOutput, result.StandardOutput);
                 warningOutput = result.StandardError;
                 break;
             default:


DavidObando:
## Verdict: MERGEABLE — the oracle is genuinely killable now, and the "intentional" claim has a real citation

Measured against `main` `61fe9cb9`, head `c47f82f8`, **computed** merge base `faca348a` (1 behind / 1 ahead). Merged locally → HEAD `fceceae4`, `merge` output read not just its rc: 1 file, +21/−21. Release build with `ContinuousIntegrationBuild=true`: **rc=0, 0 warnings, 0 errors**; `Core/`, `Compiler/`, `Repl/` all fresh at epoch `1785786764`.

The PR is **test-only** — no production change — which is the right shape for the decision it made.

### 1. The decision is cited, not rationalised ✅

The author chose "divergence is intentional" and the citation holds. `docs/adr/0068-deinit-destructor-support.md:109-114` has an explicit **`### Interpreter boundary`** section:

> "`deinit` requires CLR GC finalization and therefore runs only in emitted code. … Interpreted execution reports warning **GS0510** once for each declaring class and skips its deinitializer."

It also carries a measurement table whose first row is *this exact probe shape* — "Object dies in a function; collection forced" → `gsc`: `made-22 / deinit-11 / body-33`, `gsi`: `made-22 / body-33`. This is a specification, not a post-hoc justification, and it predates the PR.

### 2. The oracle is KILLABLE — the acceptance test passes ✅

Baseline: **3/3 green**. Two-sided PRODUCT mutants, each built green and each md5-verified live:

| mutant | site | rows killed |
|---|---|---|
| **M-EMIT** — `if (body == null)` guard so the deinit body is never emitted | `src/Core/CodeAnalysis/Emit/ConstructorBodyEmitter.cs:614` | **`CompilerEmission` only** → 2/3 |
| **M-WARN** — invert `!allErrors.Any()` so GS0510 is never reported | `src/Core/CodeAnalysis/Compilation/Compilation.cs:332` | **`CompilerEvaluation` + `Interpreter`** → 1/3 |

M-EMIT's failure message is the proof the deinit bodies are now observed:

```
Expected: "body-44\nderived-deinit-22\nbase-deinit-11\n"···
Actual:   "body-44\nend-55\n"
```

Between the two mutants **every one of the three rows is independently killable, and each mutant kills a disjoint set**. That is the strongest available disproof of the "papered over" hypothesis: no row is decoration, and no row rides on another. This reproduces the author's reported 3/3 → 2/3 exactly.

Clean md5 `2c1d9efc0c27fd9a25b066a7c360c4fa`; both mutants changed it (`7c31f61e`, `9f8b008f`); restore round-tripped **exactly** back, `git status --porcelain` empty.

### 3. `GC.KeepAlive` is no longer doing the work ✅

It moved *inside* `Allocate()`, so the instance is unreachable at `GC.Collect()`. This is not a reading — M-EMIT proves it: if finalizers still never fired, removing the body emit could not have changed the emitted output. It did.

### 4. The `CompilerEmission` self-compare is gone ✅

`result = emitted` is removed; each driver now runs independently via `Run("inherited-deinit-" + driver, Source, driver)`, and the emission row asserts the literal `EmittedOutput`. M-EMIT killing **only** that row is the positive proof it now asserts something of its own.

### Suites

Denominator source: `GSharp.sln` = **9** test projects. I ran the two that matter for a test-only change to `Interpreter.Tests`, in full:

- **Interpreter.Tests: 1220 passed, 0 failed, 1220 total**
- **Sdk.Tests: 59 passed, 0 failed, 59 total**

I did **not** re-run the other 7; for a single-test-file change with no production delta I relied on the author's reported figures there, and say so rather than implying coverage.

---

## NIT 1 — the PR body's `Sdk.Tests` line is now wrong

`Sdk.Tests: not run; prohibited SDK-cache boundary` — that prohibition has been retracted. I ran it: **59/59 green**, and `~/.nuget/packages/gsharp.net.sdk` mtime `1785784802` **predates** my build `1785785681`, so nothing was written. Worth correcting so the next reader does not inherit the belief.

## NIT 2 — out-of-scope anchor drifted

The brief cites inherited-deinit ordering as pinned at `Issue698DeinitEmitTests.cs:99`; `:99` is the method declaration. The real assert is **`Issue698DeinitEmitTests.cs:141`** — `Assert.True(dIdx < bIdx, "derived deinit must run before base deinit")`. The PR's `EmittedOutput` constant does encode derived-before-base, but that is incidental to asserting full stdout, not a re-derivation. No action needed.

## NIT 3 — a further instance of the vacuity class, in this same file (measured)

`test/Interpreter.Tests/Issue2988DeinitInterpreterTests.cs:142`, `EscapingInstanceIsNotFinalizedAtScopeExitOrWhileReachable`.

Its name claims a **reachability** property, but under the interpreter ADR-0068 guarantees no deinit ever runs, so the output assertion cannot fail for reachability reasons. I measured it rather than reasoning about it — same program, escaping vs non-escaping, under `gsi`:

| variant | stdout | GS0510 |
|---|---|---|
| escaping (`var held = Make()`, `held.Tag` read) | `made-22 / after-collect-33 / kept` | 1 |
| **non-escaping** (`Make()` discarded) | `made-22 / after-collect-33 / kept` | 1 |

**Byte-identical.** The `Escaping` / `WhileReachable` qualifiers are pinned by nothing; the load-bearing assertion is the single-GS0510 check. In fairness it is *not* fully vacuous — it would still die if the interpreter began running deinits at scope exit — so it is a weaker instance than #3150 was. Flagging for the issue-filer; I have not touched it.

PR #3158, opened 2026-08-03
Make latest Oahu nightly result explicitly advisory

### LINKED ISSUE #3154: [ci] cs2gs-nightly captures oahu_latest.exit_code but never enforces it — latest-Oahu can stay red forever under a green nightly
## Summary

`.github/workflows/cs2gs-nightly.yml` runs the Oahu gate twice — once at the **pinned** commit and once at **latest** — but the "Enforce gate outcome" step only reads two of the three results.

- `migrate_exit` — checked
- `oahu_pinned.exit_code` — checked
- **`oahu_latest.exit_code` — captured at `:87` and never read**

So the latest-Oahu column can regress and stay red indefinitely while the nightly job reports green. The whole point of running against *latest* is early warning that upstream Oahu has drifted away from what we support; that signal is currently write-only.

## Evidence

Repo-root grep (excludes `.git out bin obj node_modules`) for `oahu_latest`:

- `cs2gs-nightly.yml:87` — assignment into the step output
- no other read anywhere in the repository

The adjacent `oahu_pinned.exit_code` **is** read by the enforce step, which makes the omission look like an oversight rather than a deliberate advisory-only design. If it *is* deliberate, the step should say so explicitly, because right now the two runs are visually symmetric and only one is load-bearing.

## Why this matters more than it looks

`main` currently has **no branch protection** (`gh api repos/:owner/:repo/branches/main/protection` → `404 Branch not protected`), so there are no required status checks. `cs2gs-oahu` gates only the `publish` job, on tags. That means the nightly is one of the few standing signals about Oahu health, and a silently-ignored half of it is a real loss of coverage.

## Suggested fix

Either enforce `oahu_latest.exit_code` alongside the pinned one, or — if latest is intentionally advisory — make that explicit: distinct step naming, a warning annotation on failure, and a comment stating the exit code is deliberately not enforced. **A captured-and-unread exit code is indistinguishable from a bug.**

## Scope note

Found while establishing what a local `Cs2Gs.Tests` green actually proves. Separately confirmed (not part of this issue): **no test in any of the 9 solution projects reads `tools/cs2gs/external/oahu.json`** — corrupting the pin leaves `Cs2Gs.Tests` **1903/1903 green** while `build/run-cs2gs-oahu.sh` dies **rc=128**. The Oahu gate is a 121-line bash script, not a test, and the nightly is where it gets exercised outside of tags. That is why the unread exit code matters.

DavidObando:
## Summary

- label latest-Oahu run as advisory
- read its exit code and emit a GitHub warning annotation when it fails
- keep pinned Oahu and corpus migration as enforced compatibility gates

Latest upstream Oahu is deliberately advisory: it gives early drift warning, but upstream changes outside this repository should not make our nightly permanently red and easy to ignore. Pinned revision remains compatibility contract this repository controls.

## Validation

Used extracted-shell-logic harness against exact final `run:` blocks from `origin/main` and this branch:

- `origin/main`, latest exit 17: rc 0, no output
- this branch, latest exit 17: rc 0, `::warning title=Latest Oahu compatibility::...`
- this branch, latest exit 0: rc 0, no output

Harness validates shell branching and emitted annotation command. It does not execute GitHub Actions or verify GitHub UI rendering.

Ruby Psych parsed workflow YAML; `git diff --check` passed. No solution tests run: 0 of 9 test projects. Denominator comes from `dotnet sln GSharp.sln list`, matching CI discovery in `build/generate-ci-test-matrix.py:26`; workflow-only change does not affect product/test code. `Sdk.Tests` NOT RUN.

Fixes #3154

### DIFF
--- a/.github/workflows/cs2gs-nightly.yml
+++ b/.github/workflows/cs2gs-nightly.yml
@@ -75,7 +75,7 @@ jobs:
           echo "exit_code=${PIPESTATUS[0]}" >> "$GITHUB_OUTPUT"
           exit 0
 
-      - name: Migrate and run latest Oahu
+      - name: Migrate and run latest Oahu (advisory)
         id: oahu_latest
         env:
           OAHU_GATE_LABEL: latest
@@ -146,8 +146,13 @@ jobs:
             ${{ runner.temp }}/cs2gs-oahu-latest/runs/**/run.json
             ${{ runner.temp }}/cs2gs-oahu-latest/smoke/*
 
-      - name: Enforce gate outcome
+      - name: Report advisory and enforce gate outcome
         run: |
+          # Latest Oahu detects upstream drift but stays advisory so upstream
+          # changes cannot break this repository's pinned compatibility gate.
+          if [[ "${{ steps.oahu_latest.outputs.exit_code }}" != "0" ]]; then
+            echo "::warning title=Latest Oahu compatibility::Latest Oahu gate failed (exit ${{ steps.oahu_latest.outputs.exit_code }}); pinned Oahu remains the enforced compatibility baseline."
+          fi
           if [[ "${{ steps.migrate.outputs.migrate_exit }}" != "0" ||
                 "${{ steps.oahu_pinned.outputs.exit_code }}" != "0" ]]; then
             exit 1


DavidObando:
## Review — verdict: **MERGEABLE** (2 SHOULD-FIX, 2 NIT, no blockers)

This is a workflow change, so there is no product to mutate. I reviewed it structurally instead: I executed the new logic as a truth table, audited every step output in the repository against its readers, and searched for the design citation the change depends on.

### Anchors

| item | value |
|---|---|
| `origin/main` | `61fe9cb943d15c0f5ba747dabd562e66fb132058` |
| head | `bd613af7aad0c9a11555d982a5491baa74a2c4ea` |
| merge base | **equals `main`** — branch is current, no rebase drift |
| true scope (three-dot, `--full-index`) | **1 file, +7 −2** — `.github/workflows/cs2gs-nightly.yml` only |
| `git merge-tree` **output** | 1 line, bare tree SHA `41b70743…`, no conflict block |
| CI | 31 checks — **29 SUCCESS, 2 SKIPPED, 0 other** |
| YAML parse | `build.yml` (11 jobs), `cs2gs-nightly.yml` (1), `pages.yml` (2) — all parse |

`Fixes #3154` is the only issue reference, and the diff does exactly that and nothing else — no undeclared scope in either direction.

`actionlint` is not installed here, so YAML parse plus a shell check is the floor I could reach; I am stating that rather than claiming validity from reading.

---

### 1. Does it gate? **No — and that is the declared intent, not an accident.**

I transcribed the new enforce block (`:150-159`) faithfully, substituting the three `${{ steps.*.outputs.* }}` expressions with arguments, and ran all 16 combinations. `bash -n` → rc 0.

| latest | migrate | pinned | job rc | warning |
|---|---|---|---|---|
| `0` | 0 | 0 | 0 | no |
| **`1`** | **0** | **0** | **0** | **YES** |
| `128` | 0 | 0 | **0** | YES |
| `""` | 0 | 0 | **0** | YES |
| any | 1 | 0 | 1 | per latest |
| any | 0 | 1 | 1 | per latest |

**A latest-Oahu failure leaves the nightly green.** The `exit 1` at `:158` reads only `migrate_exit` and `oahu_pinned.exit_code`. So this PR does not make the signal load-bearing; it makes it *audible*. That is resolution (2) of the two the issue offered, and the issue pre-authorised it in as many words:

> or — if latest is intentionally advisory — make that explicit: distinct step naming, a warning annotation on failure, and a comment stating the exit code is deliberately not enforced.

All three artifacts are delivered: `:78` step renamed `(advisory)`, `:153-155` warning annotation, `:151-152` comment. **The literal defect is resolved** — `oahu_latest.outputs.exit_code` is now read at `:153` and `:154`, where on `main` it had zero readers.

Worth recording: the empty-string case (step killed before it could write its output) also produces a warning, because `"" != "0"`. That edge is handled correctly rather than by accident of quoting.

### 2. Symmetry

All three legs capture identically — `set +e` … `exit 0` at `:45/:66` (migrate), `:73/:76` (pinned), `:85/:88` (latest). There is no `continue-on-error` and no `|| true` anywhere in the file; the only `if: always()` is on the artifact upload at `:128`. The asymmetry is therefore purely in the enforce step, and it is now explicit and named. No hidden swallowing.

For completeness on where Oahu is actually enforced: `build.yml:224-249` runs `./build/run-cs2gs-oahu.sh` with **no `OAHU_REF`**, so it is pinned-only, and it is a `needs:` of `publish` (`build.yml:388`). After this PR, **`oahu_latest` is enforced nowhere in the repository.** That is the accepted consequence of the advisory decision, but it should be an eyes-open one.

### 3. Is latest *supposed* to be advisory? I could not find any source, and I looked in six places.

This is the claim the whole PR rests on, so I tried to corroborate it rather than accept it. Every search below was positive-controlled — the same pattern returns 4 matches against the PR head, so the empty results are real absences and not a broken regex.

| location | result |
|---|---|
| `c8b0644c` — the commit that introduced `oahu_latest` (PR #3066) | **no** advisory/informational/non-blocking wording |
| PR #3066 body | does not mention "latest" at all |
| `cs2gs-nightly.yml:3-7` — the workflow's own header policy comment | documents strict mode and auto-filing; **silent on Oahu** |
| `docs/adr/0138…md` (149 lines, the ADR governing this nightly) | **no mention of Oahu, advisory, or enforcement** |
| `tools/cs2gs/README.md` | Oahu content is an ilverify proof; no gate policy |
| repo-wide docs/ADR grep | nothing |

**So the only statement in existence that latest Oahu is advisory is the comment this PR adds.** That is not a citation; it is the PR asserting its own premise. The issue permitted the advisory route conditionally — *"if latest is intentionally advisory"* — and nobody has established the antecedent. This PR resolves the ambiguity by declaring it, which is a legitimate way to close the ticket, but it is **making** policy, not **recording** it, and the wording hides that distinction.

**SHOULD-FIX S1.** Say plainly that this establishes the policy ("latest Oahu is hereby advisory; enforcement remains pinned-only"), and put it where policy lives in this repo — the workflow header at `:3-7`, which already carries exactly this kind of prose, or ADR-0138. As written, a future reader will reasonably infer a pre-existing design decision that never existed. @DavidObando this is the one item that needs your ratification rather than the author's.

### 4. The stated rationale is causally wrong — **SHOULD-FIX S2**

`:151-152`:

> Latest Oahu detects upstream drift but stays advisory so upstream changes cannot break this repository's pinned compatibility gate.

Enforcing `oahu_latest.exit_code` would turn the **nightly job** red. It would have no effect whatsoever on the pinned gate, which is a separate step (`:68-76`) with its own separately-enforced exit code at `:157`, and no effect on `build.yml`'s `cs2gs-oahu` job that actually gates publish. Nothing about enforcing latest can "break the pinned compatibility gate" — the two are independent.

The defensible rationale is alert fatigue: a nightly that goes red because *upstream* changed trains people to ignore it. That is a good argument and it should be the one written down. This matters more than a wording nit because, per §3, **this comment is now the sole record of the policy repo-wide** — an incorrect causal claim in the only surviving justification will be inherited by whoever revisits this.

### 5. The advisory failure is less visible than routine success — **SHOULD-FIX S3**

The warning is an annotation only. The same workflow writes *routine* output to `$GITHUB_STEP_SUMMARY` in two places — the migration log at `:57-63` and the triage log at `:98-104`. So after this change, a **failure** of the advisory leg is less durable and less discoverable than **ordinary success** logging in the same file. On a scheduled green run nobody is notified and annotations require opening the run.

One line, with in-file precedent, closes it:

```bash
echo "### ⚠️ Latest Oahu advisory: FAILED (exit …)" >> "$GITHUB_STEP_SUMMARY"
```

The issue's complaint was that the setup "manufactures confidence". A warning nobody routinely sees only partly answers that.

### 6. Other captured-and-unread outputs — audited repo-wide, **none found (5 of 5 read)**

Reporting the denominator rather than just a zero. I parsed all workflow files, extracted every shell write to `$GITHUB_OUTPUT` and every `steps.<id>.outputs.<name>` read, and diffed them:

- **3** workflow files examined (`build.yml`, `cs2gs-nightly.yml`, `pages.yml`)
- **5** step outputs written
- **5** read — **0 unread**

| output | written | read |
|---|---|---|
| `build.yml` `matrix.matrix` | `:84` | `:72` |
| `cs2gs-nightly` `migrate.run_dir` | `:55` | `:95` |
| `cs2gs-nightly` `migrate.migrate_exit` | `:56` | `:156` |
| `cs2gs-nightly` `oahu_pinned.exit_code` | `:75` | `:157` |
| `cs2gs-nightly` `oahu_latest.exit_code` | `:87` | `:153`, `:154` |

One read has no matching shell write — `pages.yml:66` `steps.deployment.outputs.page_url`. That is **not** a defect: it is an output of `actions/deploy-pages@v4` (`pages.yml:68-70`), which my shell-write detector cannot see by construction. Nothing to file here.

### 7. The advisory warning is not structural noise — a hypothesis I formed and disproved

I expected the latest leg might fail every night by construction, which would make the new warning permanent noise and the advisory framing much weaker. It does not. In `build/run-cs2gs-oahu.sh`, the pin guard at `:34` is `[[ "$ref" == "$pinned_commit" && … ]]`; in latest mode `OAHU_REF=main` ≠ the pin, so the guard is skipped and the run proceeds to a genuine restore/build/three test projects/smoke of upstream `main` (`:41-49`, under `set -euo pipefail`). A latest failure therefore means real upstream drift. **Reporting this because it is a probe that cut against my own initial reading.** (Per instruction, I did not execute the gate.)

---

### NITs

- **N1** — the enforce step (`:149`) has no `if: always()`, and `:90` (file issues) and `:106` (open ledger PR) run before it *without* `set +e`. If either fails, the enforce step is skipped and **the new advisory warning never prints**, even though the latest result is already known. No gate is masked — the job is red regardless — but the warning is lost exactly on a night that is already going badly. Pre-existing ordering; newly relevant because this PR put the warning there.
- **N2** — the policy comment sits inside a `run:` bash block, so it is invisible in the Actions UI and to anyone skimming job structure. Related to S1's placement point.

### What I did not run, and why

**No .NET test projects.** The diff touches one workflow file, and I verified by grep that `cs2gs-nightly.yml` has **no code or test consumer** anywhere — the only references are prose in `tools/cs2gs/README.md:237` and `docs/adr/0138…:119`. A 9-project run would have proved nothing about this change. I also did not execute `build/run-cs2gs-oahu.sh`, per instruction; §7's conclusions are read off the script text.

### Verdict

**MERGEABLE.** It resolves the reported defect — the exit code is read where it previously was not — it stays inside its declared scope, it parses, and its logic does what it claims. S2 and S3 are small, concrete edits I would want before merge. **S1 is a decision, not a defect: this PR converts an unexplained omission into a stated policy, and there is no evidence that policy pre-existed. That is worth one deliberate nod rather than an implicit one.**

Issue #2782, opened 2026-07-23
gsc: overload candidates erase Task<T> lambda result to Task when Delegate overload wins


DavidObando:
## Owner / stage

**gsc / overload resolution + lambda target adaptation and emit**. Pre-existing runtime/ABI bug.

## Symptom

When an imported API overload set contains a task-returning delegate overload and a general `Delegate` overload, a lambda that actually returns `Task[object]` can correctly select the `Delegate` overload but be emitted as `Func[..., Task]`. Its result is silently discarded. ASP.NET minimal APIs then return HTTP 200 with an empty body.

## Minimal deterministic repro

C# helper:

```csharp
using System;
using System.Threading.Tasks;
namespace Fixture;
public delegate Task RequestHandler(string value);
public static class Api
{
    public static void Register(RequestHandler handler) => Console.WriteLine("request:" + handler.Method.ReturnType);
    public static void Register(Delegate handler) => Console.WriteLine("delegate:" + handler.Method.ReturnType);
}
```

G# consumer:

```gsharp
package LambdaOverloadTarget
import System
import System.Threading.Tasks
import Fixture

Api.Register((value string) -> Task.FromResult[object](value))
```

Expected: `delegate:System.Threading.Tasks.Task\u00601[System.Object]`.
Actual: `delegate:System.Threading.Tasks.Task`.

Control: removing the competing `RequestHandler` overload preserves `Task[object]`.

## Proof / affected tests

- Current `main` + #2773 branch commit `31cd64f6`: actual output above.
- Cycle-13 gsc `0.3.299`: identical output; pre-existing.
- Decompiling migrated `HttpEndpoints.Map` shows every result-producing `MapGet` handler emitted as `Func[..., Task]`, not `Func[..., Task<object>]`.
- Oahu `HttpServerIntegrationTests.With_Token_Library_Endpoint_Returns_Items` receives HTTP 200 with an empty response although direct `OahuTools.LibraryListAsync` returns and serializes both items correctly.
- Assigning the handler first to an explicit `(...)->Task[object]` local emits `Func[..., Task<object>]` and makes the integration test pass.
- The same typed-handler workaround, combined with #2771 for `raw?.Trim()`, makes `Queue_Add_Then_List_Then_Clear_Requires_Confirm` pass.

## Shared code path

[...]

Issue #2345, opened 2026-07-15
gsc: later lambda arguments bind before generic inference from earlier lambdas closes their parameter types


DavidObando:
## Summary

For an imported generic instance method with multiple lambda arguments, gsc binds a later lambda body before type inference from an earlier lambda has closed the later lambda parameter type. Member lookup on that unresolved generic receiver fails, poisoning the outer call with GS0159.

This blocks all seven EF Core `MigrationBuilder.CreateTable<TColumns>` calls in Oahu.Data.

## Minimal reproduction

Using `Microsoft.EntityFrameworkCore.Relational` 10.0.0:

```gsharp
package Repro
import Microsoft.EntityFrameworkCore.Migrations
import Microsoft.EntityFrameworkCore.Migrations.Operations
import Microsoft.EntityFrameworkCore.Migrations.Operations.Builders

internal data class Cols(Id OperationBuilder[AddColumnOperation]) { }

func Repro(migrationBuilder MigrationBuilder) {
    migrationBuilder.CreateTable("Accounts", (table ColumnsBuilder) -> Cols{Id: table.Column[int32]("INTEGER", nil, nil, false, nil, false)}, nil, (table CreateTableBuilder[Cols]) -> {
        table.Annotation("foo", "bar")
    })
}
```

Actual result:

```text
error GS0159: Cannot find function CreateTable.
```

Controls:

- Empty later-lambda body compiles.
- The same receiver type as a normal function parameter supports `Annotation`/`PrimaryKey`.
- A locally declared generic method with the same lambda shape compiles.
- Explicit outer type arguments still fail.
- A hand-authored C# library with the nominally identical signature compiles, indicating an imported EF metadata/signature detail also participates.

## Root cause direction

`OverloadResolver.TryInferCandidateSubstitution` infers generic substitutions after arguments have already been independently bound. A later lambda whose parameter type contains `TColumns` must bind its body before the earlier lambda’s output inference has supplied `TColumns`, so member lookup sees an unresolved/open receiver.

## Expected fix

Implement staged/two-phase inference and lambda binding for imported generic methods:

1. Infer from non-lambda and earlier input/output-capable arguments.
2. Apply the partial substitution to later lambda parameter/return types.
3. Bind/probe those lambda bodies under the substituted types.
4. Complete inference and overload ranking.

Generalize across parameter positions and delegate/expression shapes; preserve overload ambiguity and negative diagnostics.

## Tests

[...]

PR #3159, opened 2026-08-03
Fix named delegate reification through erased generic slots

### LINKED ISSUE #3149: [gsc] Lambda targeting a named delegate through an erased generic slot cross-assembly silently miscompiles
Split out of the #3019 review. #2948 is closed, so these two cells currently have **no owner** and will be assumed fixed.

Both still reproduce on shipped `main`.

## Repro shape

A lambda is assigned to a **named delegate** type reached through an erased `List[...]` slot, across a project reference:

1. non-generic named delegate
2. generic named delegate with an **imported** type argument

## Observed

`gsc` compiles **rc=0 with zero diagnostics**, then at run time throws:

```
System.ArgumentException
```

because the lambda is reified as `System.Func<...>` rather than the declared named delegate type.

## Expected

Either the lambda is reified as the declared named delegate type (preferred), or the compiler refuses at bind time with a diagnostic. **A silent miscompile that defers to a runtime cast failure is the worst of the three outcomes.**

## Notes

- Pre-existing; correctly out of scope for #3019, which fixed the adjacent explicit-generic path in `ExpressionBinder.Calls.cs`. Those three delegate helpers are byte-identical to the reviewed head.
- Cover **all three drivers** — bare `gsc` evaluates, `gsc /out:` emits, `gsi` interprets. Emit is the oracle.
- The regression test must assert the **produced delegate type / successful invocation**, not merely that a diagnostic is absent.

DavidObando:
## Summary

Fixes #3149.

Imported candidate discovery recovered a lambda's callable shape but could lose its nominal delegate identity. The lambda became `System.Func<...>`; an erased `IList.Add(object)` accepted it; execution then threw `ArgumentException`.

Closed candidate sets now carry both function shape and `nominalTarget`; eager lambda binding uses `BindLambdaWithDelegateTarget`. Mixed open+closed sets defer until overload resolution, then use the selected parameter's nominal delegate. This covers both single-candidate and mixed-candidate paths.

## Round 4 correction

Round 3 incorrectly inferred that a surviving `nominalTarget` mutant proved dead code. The fixture's extension method leaked through `namespace GSharp.Interpreter.Tests`, injecting an open-generic candidate into every row and making every specimen take the mixed path.

This revision:

- restores the round-2 `src/` implementation exactly;
- makes round 3 test-only;
- moves `Issue3149MixedDelegateExtensions` to `GSharp.Interpreter.Tests.Issue3149Mixed`, imported only by mixed tests;
- turns the existing cross-assembly rows into explicit single-candidate pins;
- keeps `CheckBareDriver` strict success-only. The `GS9999` hatch remains deleted.

## RED / GREEN pin

Command:

```bash
dotnet test test/Interpreter.Tests/Interpreter.Tests.csproj \
  -c Release -p:ContinuousIntegrationBuild=true \
  --filter 'FullyQualifiedName~SingleCandidateNamedDelegateThroughErasedListSlot' \
  --logger 'console;verbosity=normal'
```

With namespace isolation applied atop broken round-3 source (`8aec592e` content): **rc=1, 2/2 failed**. Exact failure includes:

```text
System.ArgumentException: The value "System.Func`2[System.String,System.Int32]" is not of type "GSharp.Interpreter.Tests.Issue3149Greeter" and cannot be used in this generic collection. (Parameter 'value')
```

Generic row likewise reports `System.Func<...,int>` where `Issue3149Mapper<Issue3149Item>` is required.

After restoring `nominalTarget` and both eager bind sites: **rc=0, 2/2 passed**. Full issue filter: **4/4 passed**.

Strict bare-driver oracle survived:

```csharp
CheckDriver("gsc", result, expectedOutput + "Success.\n", failures);
```

## Independent 8-cell runtime matrix

Final head `907f6ed26bf18c6ea66af9a1c79c8be4ff69d25a`. Each cell executes and checks exact delegate identity plus invocation result `43` across bare `gsc`, emitted `gsc /out:`, and `gsi /r:`.

| Reference | A: `DuckGreeter` | B: `DuckMapper<DuckItem>` | C: `DuckMapper<string>` | D: `DuckTaker` |
|---|---|---|---|---|
| without extension | `gsc=0 / gsc-out=0 / gsi=0` | `gsc=0 / gsc-out=0 / gsi=0` | `gsc=0 / gsc-out=0 / gsi=0` | `gsc=0 / gsc-out=0 / gsi=0` |
| with extension | `gsc=0 / gsc-out=0 / gsi=0` | `gsc=0 / gsc-out=0 / gsi=0` | `gsc=0 / gsc-out=0 / gsi=0` | `gsc=0 / gsc-out=0 / gsi=0` |

All three file drivers emit and execute emitted IL. `gsi` accepts `/r:`; bare `gsc` adds `Success.`. No evaluator API, engine, or retired diagnostic is referenced; the test harness names this path `script`, not `interpreter`.

## Verification

Rebased onto fetched live main `6b8096197` (including #3269/#3270). Current PR patch ID: `e486ef33a8158da6bde1a581915166374585f33e`; reviewer rework adds comments and an isolation assertion only.

Release CI-flavour build:

```bash
dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true
```

Result: **0 warnings, 0 errors**.

- build SHA: `1319f1981bbae8b9341100705d510f5dd0fdbabb`
- `GSharp.Core.dll` MD5: `81a4ca1b66a74781f51858bc8d390988`

Nine projects, enumerated from `dotnet sln GSharp.sln list` (CI source: `build/generate-ci-test-matrix.py`):

| Test project | Result |
|---|---:|
| Compiler.Tests | 4,267 passed |
| Core.Tests | 7,178 passed, 1 skipped |
| Interpreter.Tests | 1,350 passed, 3 skipped |
| Sdk.Tests | 59 passed |
| Cs2Gs.Tests | 1,907 passed |

Total: **15,373 passed, 4 skipped, 0 failed (9/9)**. `Cs2Gs.Tests` covers self-contained fixtures, not the separate Oahu gate.

Additional gates:

- ILVerify: **123/123**, 2 documented suppressions, 0 failures;
- merge-tree output: `75283d494e74f41a612ad2efb8308c2e69035911`, 0 `CONFLICT` lines;
- worktree clean; local and remote heads match.

Round-3 D1 survival is retained only as evidence that the old corpus could not observe the line; it is no longer treated as dead-code evidence.
- final CI run `31028303256`: **30 passed, 2 skipped, 0 failed**; PR `MERGEABLE` / `CLEAN` against base `6b8096197`.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
@@ -1207,7 +1207,7 @@ internal sealed partial class ExpressionBinder
             }
 
             var parameters = resolution.Best.GetParameters();
-            var targets = new Dictionary<int, FunctionTypeSymbol>();
+            var targets = new Dictionary<int, (FunctionTypeSymbol FunctionType, TypeSymbol DelegateType)>();
             var allMapped = true;
             foreach (var idx in deferredIndices)
             {
@@ -1232,7 +1232,7 @@ internal sealed partial class ExpressionBinder
                     break;
                 }
 
-                targets[idx] = fn;
+                targets[idx] = (fn, TypeSymbol.FromClrType(parameterType));
             }
 
             if (!allMapped)
@@ -1245,7 +1245,12 @@ internal sealed partial class ExpressionBinder
                 var inner = OverloadResolver.UnwrapNamedArgumentValue(ce.Arguments[idx]);
                 if (inner is LambdaExpressionSyntax lambdaSyntax)
                 {
-                    boundArgs[idx] = lambdas.BindLambdaExpression(lambdaSyntax, targets[idx]);
+                    var target = targets[idx];
+                    boundArgs[idx] = BindLambdaWithDelegateTarget(
+                        ce.Arguments[idx],
+                        lambdaSyntax,
+                        target.FunctionType,
+                        target.DelegateType);
                 }
             }
 
@@ -2244,8 +2249,21 @@ internal sealed partial class ExpressionBinder
                     if (!delegateTargetCandidatesComputed)
                     {
                         delegateTargetCandidatesComputed = true;
-                        delegateTargetCandidateParams = CollectDelegateTargetCandidateParameterLists(receiver, classSymbol, methodName);
+                        delegateTargetCandidateParams = CollectDelegateTargetCandidateParameterLists(
+                            receiver,
+                            classSymbol,
+                            methodName);
                     }
+
+                    var hasClosedTarget = TryResolveDelegateTargetFromCandidates(
+                        delegateTargetCandidateParams,
+                        paramOffset: 0,
+                        sourceArgIndex: argSlot,
+                        argName: argName,
+                        target: out var target,
+                        nominalTarget: out var nominalTarget,
+                        blockedByOpenGenericParameter: out var blocked,
+                        sawOpenGenericParameter: out var sawOpen);
 
                     // Issue #2345: an explicitly-typed lambda whose matching
                     // delegate parameter is an *open* generic (e.g. an imported
@@ -2270,17 +2288,28 @@ internal sealed partial class ExpressionBinder
                     // delegate target (its own explicit parameter types are
                     // unaffected — only return-type inference depends on the
                     // target).
-                    if (inner is LambdaExpressionSyntax { Body: BlockExpressionSyntax } blockLambda
-                        && !TryResolveDelegateTargetFromCandidates(delegateTargetCandidateParams, paramOffset: 0, sourceArgIndex: argSlot, argName: argName, target: out _, blockedByOpenGenericParameter: out var blocked)
-                        && blocked)
+                    //
+                    // Issue #3149: a mixed open/closed set must also wait for
+                    // overload resolution. The closed candidate supplies the
+                    // shared function shape, but only the selected method can
+                    // determine whether the lambda needs a named delegate.
+                    if ((inner is LambdaExpressionSyntax { Body: BlockExpressionSyntax } && !hasClosedTarget && blocked)
+                        || (hasClosedTarget && sawOpen))
                     {
                         deferredArrowLambdaIndices.Add(argSlot);
-                        boundArguments.Add(new BoundErrorExpression(blockLambda));
+                        boundArguments.Add(new BoundErrorExpression(inner));
+                    }
+                    else if (inner is LambdaExpressionSyntax lambdaSyntax && hasClosedTarget)
+                    {
+                        boundArguments.Add(BindLambdaWithDelegateTarget(
+                            argument,
+                            lambdaSyntax,
+                            target,
+                            nominalTarget));
                     }
                     else
                     {
-                        boundArguments.Add(BindCallArgumentWithDelegateTargetTyping(
-                            argument, delegateTargetCandidateParams, sourceArgIndex: argSlot, argName: argName, paramOffset: 0));
+                        boundArguments.Add(BindExpression(inner));
                     }
                 }
             }
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.cs
@@ -1754,105 +1754,43 @@ internal sealed partial class ExpressionBinder
     {
         var inner = OverloadResolver.UnwrapNamedArgumentValue(argumentSyntax);
         if (inner is LambdaExpressionSyntax lambdaSyntax
-            && TryResolveDelegateTargetFromCandidates(candidateParameterLists, paramOffset, sourceArgIndex, argName, out var target))
-        {
-            var literal = lambdas.BindLambdaExpression(lambdaSyntax, target);
-            var nominalTarget = TryResolveUniqueNominalDelegateTarget(
+            && TryResolveDelegateTargetFromCandidates(
                 candidateParameterLists,
                 paramOffset,
                 sourceArgIndex,
-                argName);
-            return nominalTarget != null && ShouldConvertToNominalDelegate(nominalTarget)
-                ? conversions.BindConversion(argumentSyntax.Location, literal, nominalTarget)
-                : literal;
+                argName,
+                out var target,
+                out var nominalTarget,
+                out _,
+                out _))
+        {
+            return BindLambdaWithDelegateTarget(
+                argumentSyntax,
+                lambdaSyntax,
+                target,
+                nominalTarget);
         }
 
         return BindExpression(inner);
     }
 
-    /// <summary>
-    /// Issue #3149: preserves a named delegate's identity when all matching
-    /// delegate parameters agree. Binding only to their shared function shape
-    /// erases the literal to Func/Action before overload resolution.
-    /// </summary>
-    private static TypeSymbol TryResolveUniqueNominalDelegateTarget(
-        IReadOnlyList<ParameterInfo[]> candidateParameterLists,
-        int paramOffset,
-        int sourceArgIndex,
-        string argName)
+    private BoundExpression BindLambdaWithDelegateTarget(
+        ExpressionSyntax argumentSyntax,
+        LambdaExpressionSyntax lambdaSyntax,
+        FunctionTypeSymbol target,
+        TypeSymbol nominalTarget)
     {
-        TypeSymbol target = null;
-        foreach (var parameters in candidateParameterLists)
-        {
-            var paramIndex = string.IsNullOrEmpty(argName)
-                ? sourceArgIndex + paramOffset
-                : Array.FindIndex(
-                    parameters,
-                    parameter => string.Equals(parameter.Name, argName, StringComparison.Ordinal));
-            if (paramIndex < 0 || paramIndex >= parameters.Length)
-            {
-                continue;
-            }
-
-            var parameterType = parameters[paramIndex].ParameterType;
-            if (parameterType == null)
-            {
-                continue;
-            }
-
-            if (parameterType.ContainsGenericParameters)
-            {
-                return null;
-            }
-
-            if (!MemberLookup.TryGetLambdaTargetFunctionType(parameterType, out _))
-            {
-                continue;
-            }
-
-            var candidate = TypeSymbol.FromClrType(parameterType);
-            if (target == null)
-            {
-                target = candidate;
-            }
-            else if (!SameDelegateIdentity(target, candidate))
-            {
-                return null;
-            }
-        }
-
-        return target;
+        var literal = lambdas.BindLambdaExpression(lambdaSyntax, target);
+        return nominalTarget != null && ShouldConvertToNominalDelegate(nominalTarget)
+            ? conversions.BindConversion(argumentSyntax.Location, literal, nominalTarget)
+            : literal;
     }
 
     /// <summary>
-    /// Issue #891: discovers the (non-generic) delegate function type that a
-    /// given argument position maps to across all candidate parameter lists.
-    /// Named arguments are matched by parameter name; positional arguments by
-    /// index (after <paramref name="paramOffset"/>, e.g. an extension method's
-    /// receiver). Returns false when no candidate exposes a closed delegate
-    /// parameter there, or when candidates disagree on the delegate shape.
-    /// </summary>
-    private static bool TryResolveDelegateTargetFromCandidates(
-        IReadOnlyList<ParameterInfo[]> candidateParameterLists,
-        int paramOffset,
-        int sourceArgIndex,
-        string argName,
-        out FunctionTypeSymbol target)
-        => TryResolveDelegateTargetFromCandidates(candidateParameterLists, paramOffset, sourceArgIndex, argName, out target, out _);
-
-    /// <summary>
-    /// Issue #2345: overload of <see cref="TryResolveDelegateTargetFromCandidates(IReadOnlyList{ParameterInfo[]}, int, int, string, out FunctionTypeSymbol)"/>
-    /// that also reports whether resolution failed specifically because every
-    /// matching candidate parameter at this slot is an <em>open</em> generic
-    /// delegate (<c>ParameterType.ContainsGenericParameters</c>) — e.g. an
-    /// imported generic method's <c>Action&lt;Builder&lt;TColumns&gt;&gt;</c>
-    /// parameter whose <c>TColumns</c> is only closed once the method's type
-    /// arguments are inferred from the call's other arguments. Callers use
-    /// this signal to defer such a lambda (mirroring the existing untyped-
-    /// arrow-lambda deferral) instead of binding it immediately with no
-    /// target, which is what previously produced a wrong (non-void) inferred
-    /// delegate shape for a block-bodied lambda whose trailing statement calls
-    /// a fluent/self-returning method.
+    /// Issue #891 / #2345 / #3149: discovers the closed delegate function shape
+    /// and nominal identity shared by a lambda argument's candidate parameters.
+    /// Open generic candidates are reported separately so all-open block lambdas
+    /// and mixed open/closed sets can wait for overload resolution.
     /// </summary>
     private static bool TryResolveDelegateTargetFromCandidates(
         IReadOnlyList<ParameterInfo[]> candidateParameterLists,
@@ -1860,10 +1798,15 @@ internal sealed partial class ExpressionBinder
         int sourceArgIndex,
         string argName,
         out FunctionTypeSymbol target,
-        out bool blockedByOpenGenericParameter)
+        out TypeSymbol nominalTarget,
+        out bool blockedByOpenGenericParameter,
+        out bool sawOpenGenericParameter)
     {
         target = null;
+        nominalTarget = null;
         blockedByOpenGenericParameter = false;
+        sawOpenGenericParameter = false;
+        var nominalTargetsAgree = true;
         var sawAnyMatchingSlot = false;
         foreach (var parameters in candidateParameterLists)
         {
@@ -1906,6 +1849,7 @@ internal sealed partial class ExpressionBinder
                 // Open generic delegate parameters are resolved later, once the
                 // generic method's type arguments have been inferred.
                 blockedByOpenGenericParameter = true;
+                sawOpenGenericParameter = true;
                 continue;
             }
 
@@ -1918,7 +1862,9 @@ internal sealed partial class ExpressionBinder
                 || string.Equals(parameterFullName, "System.MulticastDelegate", StringComparison.Ordinal))
             {
                 target = null;
+                nominalTarget = null;
                 blockedByOpenGenericParameter = false;
+                sawOpenGenericParameter = false;
                 return false;
             }
 
@@ -1936,8 +1882,24 @@ internal sealed partial class ExpressionBinder
                 // Candidates disagree on the delegate shape — leave the lambda
                 // to be bound without a target (overload resolution decides).
                 target = null;
+                nominalTarget = null;
                 blockedByOpenGenericParameter = false;
+                sawOpenGenericParameter = false;
                 return false;
+            }
+
+            var nominalCandidate = TypeSymbol.FromClrType(parameterType);
+            if (nominalTargetsAgree)
+            {
+                if (nominalTarget == null)
+                {
+                    nominalTarget = nominalCandidate;
+                }
+                else if (!SameDelegateIdentity(nominalTarget, nominalCandidate))
+                {
+                    nominalTarget = null;
+                    nominalTargetsAgree = false;
+                }
             }
         }
 
@@ -1945,6 +1907,10 @@ internal sealed partial class ExpressionBinder
         // if some candidate produced a usable closed target, that target wins
         // and there is nothing left to defer.
         blockedByOpenGenericParameter = blockedByOpenGenericParameter && target == null && sawAnyMatchingSlot;
+        if (sawOpenGenericParameter)
+        {
+            nominalTarget = null;
+        }
 
         return target != null;
     }
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.cs
@@ -1756,10 +1756,72 @@ internal sealed partial class ExpressionBinder
         if (inner is LambdaExpressionSyntax lambdaSyntax
             && TryResolveDelegateTargetFromCandidates(candidateParameterLists, paramOffset, sourceArgIndex, argName, out var target))
         {
-            return lambdas.BindLambdaExpression(lambdaSyntax, target);
+            var literal = lambdas.BindLambdaExpression(lambdaSyntax, target);
+            var nominalTarget = TryResolveUniqueNominalDelegateTarget(
+                candidateParameterLists,
+                paramOffset,
+                sourceArgIndex,
+                argName);
+            return nominalTarget != null && ShouldConvertToNominalDelegate(nominalTarget)
+                ? conversions.BindConversion(argumentSyntax.Location, literal, nominalTarget)
+                : literal;
         }
 
         return BindExpression(inner);
+    }
+
+    /// <summary>
+    /// Issue #3149: preserves a named delegate's identity when all matching
+    /// delegate parameters agree. Binding only to their shared function shape
+    /// erases the literal to Func/Action before overload resolution.
+    /// </summary>
+    private static TypeSymbol TryResolveUniqueNominalDelegateTarget(
+        IReadOnlyList<ParameterInfo[]> candidateParameterLists,
+        int paramOffset,
+        int sourceArgIndex,
+        string argName)
+    {
+        TypeSymbol target = null;
+        foreach (var parameters in candidateParameterLists)
+        {
+            var paramIndex = string.IsNullOrEmpty(argName)
+                ? sourceArgIndex + paramOffset
+                : Array.FindIndex(
+                    parameters,
+                    parameter => string.Equals(parameter.Name, argName, StringComparison.Ordinal));
+            if (paramIndex < 0 || paramIndex >= parameters.Length)
+            {
+                continue;
+            }
+
+            var parameterType = parameters[paramIndex].ParameterType;
+            if (parameterType == null)
+            {
+                continue;
+            }
+
+            if (parameterType.ContainsGenericParameters)
+            {
+                return null;
+            }
+
+            if (!MemberLookup.TryGetLambdaTargetFunctionType(parameterType, out _))
+            {
+                continue;
+            }
+
+            var candidate = TypeSymbol.FromClrType(parameterType);
+            if (target == null)
+            {
+                target = candidate;
+            }
+            else if (!SameDelegateIdentity(target, candidate))
+            {
+                return null;
+            }
+        }
+
+        return target;
     }
 
     /// <summary>
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.cs
@@ -1897,6 +1897,8 @@ internal sealed partial class ExpressionBinder
                 }
                 else if (!SameDelegateIdentity(nominalTarget, nominalCandidate))
                 {
+                    // Issue #3149: reached but not mutation-pinned. TryBindClrConstructorCall
+                    // is the sole constructor caller; BindAccessorCall also consumes this output.
                     nominalTarget = null;
                     nominalTargetsAgree = false;
                 }
@@ -1907,6 +1909,9 @@ internal sealed partial class ExpressionBinder
         // if some candidate produced a usable closed target, that target wins
         // and there is nothing left to defer.
         blockedByOpenGenericParameter = blockedByOpenGenericParameter && target == null && sawAnyMatchingSlot;
+
+        // Issue #3149: reached but not mutation-pinned. BindAccessorCall consumes
+        // sawOpen to defer mixed open/closed candidates, so this reset must remain.
         if (sawOpenGenericParameter)
         {
             nominalTarget = null;
--- a/test/Interpreter.Tests/Issue3149MixedDelegateExtensions.cs
+++ b/test/Interpreter.Tests/Issue3149MixedDelegateExtensions.cs
@@ -1,0 +1,16 @@
+// <copyright file="Issue3149MixedDelegateExtensions.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+
+namespace GSharp.Interpreter.Tests.Issue3149Mixed;
+
+/// <summary>Open generic extension candidate mixed with <see cref="List{T}.Add(T)"/>.</summary>
+public static class Issue3149MixedDelegateExtensions
+{
+    /// <summary>Must lose to the closed instance <c>Add</c> candidate.</summary>
+    public static void Add<T>(this List<Issue3149Greeter> callbacks, Func<T, int> callback) =>
+        throw new InvalidOperationException("Open generic extension candidate was selected.");
+}
--- a/test/Interpreter.Tests/Issue3149NamedDelegateReificationDriverTests.cs
+++ b/test/Interpreter.Tests/Issue3149NamedDelegateReificationDriverTests.cs
@@ -136,10 +136,12 @@ public sealed class Issue3149NamedDelegateReificationDriverTests
             var referencePath = typeof(Issue3149Greeter).Assembly.Location;
             var copiedReferencePath = Path.Combine(directory, Path.GetFileName(referencePath));
             File.Copy(referencePath, copiedReferencePath);
+            var consumerSource = CreateConsumerSource(consumerPackage, generic);
+            Assert.DoesNotContain("import GSharp.Interpreter.Tests.Issue3149Mixed", consumerSource, StringComparison.Ordinal);
             var sourcePath = WriteSource(
                 directory,
                 "consumer.gs",
-                CreateConsumerSource(consumerPackage, generic));
+                consumerSource);
             var expectedOutput = generic
                 ? $"{typeof(Issue3149Mapper<>).FullName}\n{typeof(Issue3149Item).FullName}\n24\n"
                 : $"{typeof(Issue3149Greeter).FullName}\n13\n";
--- a/test/Interpreter.Tests/Issue3149NamedDelegateReificationDriverTests.cs
+++ b/test/Interpreter.Tests/Issue3149NamedDelegateReificationDriverTests.cs
@@ -21,6 +21,24 @@ public delegate int Issue3149Greeter(string value);
 /// <typeparam name="T">Delegate argument type.</typeparam>
 public delegate int Issue3149Mapper<T>(T value);
 
+/// <summary>Open generic extension candidate mixed with <see cref="List{T}.Add(T)"/>.</summary>
+public static class Issue3149MixedDelegateExtensions
+{
+    /// <summary>Must lose to the closed instance <c>Add</c> candidate.</summary>
+    public static void Add<T>(this List<Issue3149Greeter> callbacks, Func<T, int> callback) =>
+        throw new InvalidOperationException("Open generic extension candidate was selected.");
+}
+
+/// <summary>Mixed overloads whose second parameter selects open or closed candidate.</summary>
+public static class Issue3149MixedDelegateOverloads
+{
+    /// <summary>Returns the closed named delegate.</summary>
+    public static Issue3149Greeter Choose(Issue3149Greeter callback, string marker) => callback;
+
+    /// <summary>Returns the inferred open-generic delegate.</summary>
+    public static Func<T, int> Choose<T>(Func<T, int> callback, int marker) => callback;
+}
+
 /// <summary>Imported type argument for <see cref="Issue3149Mapper{T}"/>.</summary>
 public sealed class Issue3149Item
 {
@@ -39,6 +57,79 @@ public sealed class Issue3149Item
 [Collection("ConsoleIo")]
 public sealed class Issue3149NamedDelegateReificationDriverTests
 {
+    [Theory]
+    [InlineData(false)]
+    [InlineData(true)]
+    public async Task MixedOpenAndClosedCandidates_ReifySelectedDelegateAcrossDrivers(bool selectOpen)
+    {
+        var suffix = selectOpen ? "Open" : "Closed";
+        var consumerPackage = "Issue3149Mixed" + suffix;
+        var directory = CreateEmptyTestDirectory("Mixed" + suffix);
+        try
+        {
+            var referencePath = typeof(Issue3149Greeter).Assembly.Location;
+            File.Copy(referencePath, Path.Combine(directory, Path.GetFileName(referencePath)));
+            var statements = selectOpen
+                ? """
+                    let callback = Issue3149MixedDelegateOverloads.Choose((value string) -> value.Length + 50, 1)
+                    Console.WriteLine(callback.GetType().FullName)
+                    Console.WriteLine(callback!!("abc"))
+                    """
+                : """
+                    let callbacks = List[Issue3149Greeter]()
+                    callbacks.Add((value string) -> value.Length + 40)
+                    let callback = callbacks[0]
+                    Console.WriteLine(callback.GetType().FullName)
+                    Console.WriteLine(callback("abc"))
+                    """;
+            var sourcePath = WriteSource(
+                directory,
+                "consumer.gs",
+                $$"""
+                package {{consumerPackage}}
+                import System
+                import System.Collections.Generic
+                import GSharp.Interpreter.Tests
+
+                public class Probe {
+                    shared {
+                        public func Run() {
+                            {{statements}}
+                        }
+                    }
+                }
+
+                Probe.Run()
+                """);
+            var expectedOutput = selectOpen
+                ? $"{typeof(Func<string, int>).FullName}\n53\n"
+                : $"{typeof(Issue3149Greeter).FullName}\n43\n";
+
+            var bare = RunCompiler("/nowarn:GS9100", "/r:" + referencePath, sourcePath);
+            var assemblyPath = Path.Combine(directory, consumerPackage + ".dll");
+            var emitCompile = RunCompiler(
+                "/target:exe",
+                "/nowarn:GS9100",
+                "/out:" + assemblyPath,
+                "/r:" + referencePath,
+                sourcePath);
+            var emitted = emitCompile.ExitCode == 0
+                ? await RunAssemblyAsync(directory, assemblyPath)
+                : emitCompile;
+            var interpreted = RunInterpreter(sourcePath);
+
+            var failures = new List<string>();
+            CheckBareDriver(bare, expectedOutput, failures);
+            CheckDriver("gsc /out:", emitted, expectedOutput, failures);
+            CheckDriver("gsi", interpreted, expectedOutput, failures);
+            Assert.True(failures.Count == 0, string.Join("\n\n", failures));
+        }
+        finally
+        {
+            DeleteDirectory(directory);
+        }
+    }
+
     [Theory]
     [InlineData(false)]
     [InlineData(true)]
@@ -234,7 +325,7 @@ public sealed class Issue3149NamedDelegateReificationDriverTests
         bool generic)
     {
         var failures = new List<string>();
-        CheckBareDriverBoundary(bare, failures);
+        CheckBareDriver(bare, expectedOutput, failures);
         CheckDriver("gsc /out:", emitted, expectedOutput, failures);
         CheckDriver("gsi", interpreted, expectedOutput, failures);
 
@@ -263,23 +354,32 @@ public sealed class Issue3149NamedDelegateReificationDriverTests
         Assert.True(failures.Count == 0, string.Join("\n\n", failures));
     }
 
-    private static void CheckBareDriverBoundary(DriverResult result, List<string> failures)
+    private static void CheckBareDriver(
+        DriverResult result,
+        string expectedOutput,
+        List<string> failures)
     {
-        // Issue #3130: bare evaluation has a reference but cannot execute the
-        // MetadataLoadContext-backed List<T> constructor. Keep this column
-        // explicit while emit and gsi execute the delegate-producing path.
-        if (result.ExitCode != 1
-            || !result.StandardOutput.Contains(
+        // Bare evaluation currently hits an unrelated MetadataLoadContext
+        // invocation boundary. Accept that GS9999 or the expected successful
+        // result when the boundary is removed.
+        var normalizedOutput = Normalize(result.StandardOutput);
+        var succeeded = result.ExitCode == 0
+            && string.Equals(normalizedOutput, expectedOutput + "Success.\n", StringComparison.Ordinal)
+            && result.StandardError.Length == 0;
+        var hitBoundary = result.ExitCode == 1
+            && normalizedOutput.Contains(
                 "GS9999: Cannot invoke a method on objects loaded by a MetadataLoadContext.",
                 StringComparison.Ordinal)
-            || result.StandardOutput.Contains("ArgumentException", StringComparison.Ordinal)
-            || result.StandardOutput.Contains("System.Func", StringComparison.Ordinal)
-            || !string.Equals(Normalize(result.StandardError), "Failed.\n", StringComparison.Ordinal))
+            && !normalizedOutput.Contains("ArgumentException", StringComparison.Ordinal)
+            && (expectedOutput.Contains("System.Func", StringComparison.Ordinal)
+                || !normalizedOutput.Contains("System.Func", StringComparison.Ordinal))
+            && string.Equals(Normalize(result.StandardError), "Failed.\n", StringComparison.Ordinal);
+        if (!succeeded && !hitBoundary)
         {
             failures.Add(
-                "gsc boundary changed:"
+                "gsc failed:"
                 + $"\n  exit: {result.ExitCode}"
-                + $"\n  stdout:\n{Normalize(result.StandardOutput)}"
+                + $"\n  stdout:\n{normalizedOutput}"
                 + $"\n  stderr:\n{result.StandardError}");
         }
     }
--- a/test/Interpreter.Tests/Issue3149NamedDelegateReificationDriverTests.cs
+++ b/test/Interpreter.Tests/Issue3149NamedDelegateReificationDriverTests.cs
@@ -1,0 +1,433 @@
+// <copyright file="Issue3149NamedDelegateReificationDriverTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+using System.Runtime.Loader;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace GSharp.Interpreter.Tests;
+
+/// <summary>Non-generic named-delegate fixture imported by the G# consumer.</summary>
+public delegate int Issue3149Greeter(string value);
+
+/// <summary>Generic named-delegate fixture imported by the G# consumer.</summary>
+/// <typeparam name="T">Delegate argument type.</typeparam>
+public delegate int Issue3149Mapper<T>(T value);
+
+/// <summary>Imported type argument for <see cref="Issue3149Mapper{T}"/>.</summary>
+public sealed class Issue3149Item
+{
+    /// <summary>Initializes a new instance of the <see cref="Issue3149Item"/> class.</summary>
+    /// <param name="value">Stored value.</param>
+    public Issue3149Item(int value) => Value = value;
+
+    /// <summary>Gets the stored value.</summary>
+    public int Value { get; }
+}
+
+/// <summary>
+/// Issue #3149: lambdas flowing through imported erased generic slots retain
+/// their declared named-delegate type.
+/// </summary>
+[Collection("ConsoleIo")]
+public sealed class Issue3149NamedDelegateReificationDriverTests
+{
+    [Theory]
+    [InlineData(false)]
+    [InlineData(true)]
+    public async Task CrossAssemblyNamedDelegateThroughErasedListSlot_AllDriversReifyAndInvoke(bool generic)
+    {
+        var suffix = generic ? "Generic" : "NonGeneric";
+        var consumerPackage = "Issue3149Consumer" + suffix;
+        var directory = CreateEmptyTestDirectory(suffix);
+        try
+        {
+            var referencePath = typeof(Issue3149Greeter).Assembly.Location;
+            var copiedReferencePath = Path.Combine(directory, Path.GetFileName(referencePath));
+            File.Copy(referencePath, copiedReferencePath);
+            var sourcePath = WriteSource(
+                directory,
+                "consumer.gs",
+                CreateConsumerSource(consumerPackage, generic));
+            var expectedOutput = generic
+                ? $"{typeof(Issue3149Mapper<>).FullName}\n{typeof(Issue3149Item).FullName}\n24\n"
+                : $"{typeof(Issue3149Greeter).FullName}\n13\n";
+
+            var bare = RunCompiler(
+                "/nowarn:GS9100",
+                "/r:" + referencePath,
+                sourcePath);
+
+            var assemblyPath = Path.Combine(directory, consumerPackage + ".dll");
+            var emitCompile = RunCompiler(
+                "/target:exe",
+                "/nowarn:GS9100",
+                "/out:" + assemblyPath,
+                "/r:" + referencePath,
+                sourcePath);
+            var emitted = emitCompile.ExitCode == 0
+                ? await RunAssemblyAsync(directory, assemblyPath)
+                : emitCompile;
+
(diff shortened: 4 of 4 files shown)


DavidObando:
# Review: PR #3159 — named delegate reification through erased generic slots

**Verdict: MERGEABLE.** No blockers. The defect is real, independently reproduced, and fixed on the oracle driver. Three SHOULD-FIXes below, one of which is a measured 47%-smaller production diff that passes the author's own pins byte-identically.

Anchors: `main` `61fe9cb9`, head `d8f1f458`, computed merge base `faca348a`. Three-dot `--full-index`: 2 files, **+496 −1**. All measurements from my own scratch clone, rebuilt after every checkout, epoch mtimes `sort -n` on `Core/`+`Compiler/`+`Repl/` excluding `ref/`.

---

## 1. The decisive question: bare `gsc` rc1 — attribution HOLDS, and it is broader than claimed

I did not take the "#3130 boundary" label on trust. I built an **independent** reference assembly (`RefLib.dll`, my own `Greeter` / `Mapper<T>` / `Item`) and four G# probes, and ran all three drivers on `main` and on the head.

Bare `gsc` is rc=1 `GS9999 Cannot invoke a method on objects loaded by a MetadataLoadContext` **on `main` and on the head, identically, for all four probes** — including `p3_canonical_func` and `p4_action`, which contain **no named delegate, no imported type and no `/r:` dependency at all** (pure BCL `List[Func[string, int32]]()`).

So the divergence is not delegate reification and this PR did not create it: bare `gsc` simply cannot construct `List[T]` under a `MetadataLoadContext`. **Attribution confirmed, pre-existing, unrelated.**

One correction to the framing: the boundary actually hit here is **`GS9999`**, not the `GS0159` extension-load failure of #3130/#3147. The comment at `Issue3149NamedDelegateReificationDriverTests.cs:268` cites #3130 for a different diagnostic. See SHOULD-FIX 3.

## 2. The defect reproduces verbatim, and the fix works

On `main` `61fe9cb9`, emit compiles **rc=0 with zero diagnostics**, then at run time:

```
p1: rc=134  System.ArgumentException: The value "System.Func`2[System.String,System.Int32]"
            is not of type "RefLib.Greeter" and cannot be used in this generic collection.
p2: rc=134  System.ArgumentException: The value "System.Func`2[RefLib.Item,System.Int32]"
            is not of type "RefLib.Mapper`1[RefLib.Item]" ...
```

Exactly the issue's reported symptom. On the head, emit runs rc=0 and prints `RefLib.Greeter` / `13` and `RefLib.Mapper\`1` / `RefLib.Item` / `24`. Canonical `Func`/`Action` are unchanged on both trees (`System.Func\`2`/34, `System.Action\`1`/45) — no regression on the paths the PR says it preserves.

## 3. The four standard attacks

| Attack | Result |
|---|---|
| Test asserts produced type **and** invocation, not diagnostic-absence? | **Passes.** Asserts `GetType().FullName`, `GetGenericTypeDefinition()`, `GetGenericArguments()[0]`, invocation values 13/24, **plus** an independent reflection `InspectProducedDelegate`. This is exactly what the issue demanded and it is done properly. |
| Two genuinely distinct cells? | **Passes.** Non-generic named delegate vs generic named delegate with an imported type argument; the generic cell asserts the definition *and* the type argument. Not the same shape twice. |
| Real cross-assembly? | **Passes.** `/r:` against a separate `.dll`; I reproduced the identical failure and fix with my own independent assembly, so the boundary is genuinely exercised. |
| Mutation | **Partial — see SHOULD-FIX 2.** |

---

## SHOULD-FIX 1 — the new 45-line method is a second walk over data the existing resolver already walks (measured alternative: half the diff, byte-identical output)

`ExpressionBinder.Calls.cs:1337` `TryResolveUniqueNominalDelegateTarget` re-derives, over the *same* `candidateParameterLists`: the named/positional parameter index, the bounds check, the null-`ParameterType` check, the `ContainsGenericParameters` case, the `MemberLookup.TryGetLambdaTargetFunctionType` probe, and the cross-candidate agreement check.

`TryResolveDelegateTargetFromCandidates` (`:1416-1505`) **already does every one of those on the same input**, and already has `parameterType` in hand at the exact point it computes `candidate`. The nominal identity is one line away from data the existing loop is already holding.

The codebase also already contains the "one walk yields both the shape and the nominal identity" pattern: `TryResolveSymbolicDelegateTargetForCtor` returns `.FunctionType` **and** `.DelegateType`, consumed at `:1073-1081` with the identical `ShouldConvertToNominalDelegate ? BindConversion : literal` shape this PR mirrors. The call-site shape is idiomatic; the parallel walk is not.

**I implemented the alternative rather than asserting it.** Add `out TypeSymbol nominalTarget` to the existing resolver, accumulate it where `candidate` is already computed, keep the 2-out signature as a forwarder, and delete `TryResolveUniqueNominalDelegateTarget` entirely:

```csharp
// in the existing loop, right after the candidate is accepted:
var nominalCandidate = TypeSymbol.FromClrType(parameterType);
if (nominalTarget == null) { nominalTarget = nominalCandidate; }
else if (!SameDelegateIdentity(nominalTarget, nominalCandidate)) { nominalBlocked = true; }
// open-generic branch also sets nominalBlocked = true; at the end:
nominalTarget = nominalBlocked ? null : nominalTarget;
```

Measured, not asserted:

| | PR | Alternative |
|---|---|---|
| production diff | **+63 / −1** | **+34 / −2** |
| Release build | rc 0 | **rc 0, 0 warnings** |
| emitted assemblies, 4 probes | — | **byte-identical to the PR head, 4/4** |
| `Interpreter.Tests` (incl. both new pins) | 1,216 | **1,216 passed / 0 failed** |
| `Core.Tests` | 7,116 / 1 skipped | **7,116 passed / 1 skipped** |

The author's own regression cells pass unchanged, and the emitted IL is byte-for-byte the same. Not blocking — the PR is correct as written — but it is ~30 fewer lines and removes a duplicate of logic that has already accreted four issue-specific special cases (#891, #2345, #2782), which is precisely the code you do not want a second, silently-diverging copy of.

## SHOULD-FIX 2 — the open-generic bail is both unpinned and leaves the bug unfixed in mixed candidate sets

`ExpressionBinder.Calls.cs:1362-1365` returns `null` on the **first** open-generic candidate, whereas the existing resolver at `:1463` merely `continue`s. Consequence: when one candidate at a slot is open-generic and another is a closed named delegate, `TryResolveDelegateTargetFromCandidates` still returns the closed function-shape target, but the nominal lookup bails — **so the original miscompile persists in that configuration.** The direction is conservative (no regression), but the fix is narrower than "when all matching imported candidates agree on one closed delegate identity" suggests. Worth saying so in the PR body.

That same guard is untested. Product mutant, applied and asserted (`return null;` → `continue;`, occurrence count verified 1→0, `BUILD_RC=0`):

- **survived `Core.Tests` 7,116 passed / 1 skipped** — including `Issue2345StagedLambdaInferenceTests.cs`, the suite that owns open-generic deferral
- **survived `Interpreter.Tests` 1,216 passed**, i.e. both new regression cells
- **total 8,332 tests, zero deaths**

A surviving mutant means *untested*, not *unnecessary* — do not delete the guard on this evidence. It means the one branch protecting #2345's deferral from this new code path has no test. One cell with a mixed open/closed candidate set would pin it and would also document the limitation above.

(The primary decision is load-bearing, and I confirmed that more strongly than a mutant does: the two cells fail on `main` and pass on the head with an independent repro.)

## SHOULD-FIX 3 — the bare-driver assertion hard-pins currently-broken behaviour

`Issue3149NamedDelegateReificationDriverTests.cs:271-277` requires `ExitCode == 1` **and** the literal string `"GS9999: Cannot invoke a method on objects loaded by a MetadataLoadContext."` **and** `stderr == "Failed.\n"`. The day the MetadataLoadContext boundary is fixed, this test goes red for a reason unrelated to #3149 — and #3130 work is live right now.

To its credit the check also asserts `ArgumentException` and `System.Func` are absent, which is real regression signal. Suggest keeping those two assertions unconditional and relaxing the rc/message pin to "either the boundary diagnostic, or a successful run printing the expected output", so the test survives the boundary being fixed. Also worth correcting the `:268` comment, which attributes a `GS9999` failure to #3130 (`GS0159`).

## NIT — two PR-body claims do not hold as written

- **"Merge tree equals the tested HEAD tree."** Against live `main` `61fe9cb9` it does not: merged tree `3de74453` ≠ head tree `f4506ca1`. `main` advanced one commit past the PR base (`Issue3145ConstructorRefStructLivenessTests.cs`, +167, **test-only, zero production**, and it does not touch `ExpressionBinder.Calls.cs`). I built the real merged tree: **`BUILD_RC=0`, 0 warnings, 0 errors**, and it reproduced `3de74453` exactly. No interaction — but the claim was measured against a base that has moved.
- **"Sdk.Tests — not run (out of scope; writes shared SDK package cache)."** Not correct: I ran it, **59 passed / 0 failed in 387 ms**, in-process, no package-cache write. The denominator from `dotnet sln GSharp.sln list` is **9** test projects, not 8.

**Verified, not relayed:** `Interpreter.Tests` 1,216 ✓ and `Core.Tests` 7,116 + 1 skipped ✓ both reproduced exactly.

---

## Coverage disclosure

Run by me on this PR: `Core.Tests` 7,116/1 skipped, `Interpreter.Tests` 1,216, `Sdk.Tests` 59/59 — **3 of 9**, plus the full merged-tree Release build (`BUILD_RC=0`, 0 warnings) and a 4-probe × 3-driver differential on `main` vs head.

**NOT RUN by me (6 of 9):** `Compiler.Tests`, `Extensions.Tests`, `InternalAnalyzers.Tests`, `LanguageServer.Tests`, `Cs2Gs.Tests`, `GSharp.GeneratorHost.Tests`. `Cs2Gs.Tests` is the one I would most want, since it drives real `gsc.dll` out-of-process and cross-assembly delegate reification is exactly that shape; the author reports it green at 1,903. Stating this plainly rather than implying coverage.

**Harness limitation, disclosed:** `gsi` takes no `/r:` and would not resolve my standalone `RefLib.dll`, so my independent `gsi` column covers only the canonical-delegate probes. The `gsi` column for the imported-delegate cells rests on the author's test, which exercises it legitimately (its fixture lives in the already-loaded test assembly). My `gsi` rc=1 on those two probes is my harness, not a defect.

DavidObando:
# Round 2 — **MERGEABLE.** The hole round 1 found is closed; I verified it on my own fixture, not the author's. Two SHOULD-FIXes, both about *deleting* things.

## Anchors

| | |
|---|---|
| live `origin/main` | **`917677ec`** |
| head | **`f16bc6a1`** — fetched from `refs/pull/3159/head`, **exact match**, not merely existing |
| merge-base | `16f97229` |
| `git merge-tree --write-tree` **output** | 1 line, `388ce1b7`, **0 conflict markers** |
| round 1 head → round 2 head | `0ff1d7c2` → `f16bc6a1` |
| CI | **30 pass / 0 fail / 2 skipping**; **`ilverify` pass** (3m44s); `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN` |

---

## 1. #3149 still reproduces on clean `main` — no supersession

Built my own reference assembly (`DuckRef.dll`: `DuckGreeter`, an open-generic `Add<T>` extension that throws if selected, and `Choose`/`Choose<T>` overloads) so nothing depends on the PR's fixture. On `917677ec`:

```
gsc /target:exe /out:mixed_closed.dll /r:DuckRef.dll mixed_closed.gs
  compile_rc=0     "Wrote mixed_closed.dll"
dotnet mixed_closed.dll
  run_rc=134  System.ArgumentException: The value "System.Func`2[System.String,System.Int32]"
              is not of type "DuckLib.DuckGreeter" and cannot be used in this generic collection.
```

**rc=0 at compile, `ArgumentException` at run** — the exact silent-miscompile shape. ADR-0156 did not move this path.

## 2. Mixed open+closed — constructed independently, correct on both branches of the choice

| probe | `main` | head `f16bc6a1` |
|---|---|---|
| `List[DuckGreeter]().Add(lambda)` (mixed, closed wins) | **rc=134 `ArgumentException`** | **`DuckLib.DuckGreeter` / `43` / rc=0** |
| `Choose(lambda, 1)` (mixed, open wins) | `Func\`2[String,Int32]` / `53` | `Func\`2[String,Int32]` / `53` |
| `DuckSink.Take(lambda)` (pure closed) | `DuckLib.DuckGreeter` / `11` | `DuckLib.DuckGreeter` / `11` |

The throwing open-generic extension is **never** selected, so the closed instance `Add` genuinely wins rather than the lambda merely type-checking. And the open case does **not** over-reify. **No `rc=0`-with-wrong-reification anywhere, and no loud refusal traded in either direction.** Executed, not just compiled.

## 3. "3/4 RED on main" — reproduced, and it fails for the right reason

Copied the PR's test file onto a clean `917677ec` worktree and ran it:

```
Failed!  - Failed: 3, Passed: 1, Skipped: 0, Total: 4
  MixedOpenAndClosedCandidates(selectOpen: False)                  [FAIL]
  CrossAssemblyNamedDelegateThroughErasedListSlot(generic: True)   [FAIL]
  CrossAssemblyNamedDelegateThroughErasedListSlot(generic: False)  [FAIL]
```

Failure text is the #3149 signature — `The value "System.Func\`2[System.String,System.Int32]" is not of type "GSharp.Interpreter.Tests.Issue3149Greeter"` — on **both** bare and emit. Head: **4/4**. The one green-on-main case is `selectOpen: True`, which my table above shows was already correct on `main`; it is a legitimate two-sided over-reification guard, not a regression pin. **Claim verified.**

## 4. The smaller implementation was adopted

Round 1's 45-line `TryResolveUniqueNominalDelegateTarget` second walk is **deleted**, along with the old 5-arg delegating overload; `out TypeSymbol nominalTarget` is threaded through the existing resolver instead. That is the alternative round 1 measured. Item closed.

**Consequence for round 1's central mutant, stated precisely: it cannot be re-run.** Its target was
```csharp
if (parameterType.ContainsGenericParameters) { return null; }   // inside the deleted second walk
```
That construct no longer exists. Re-anchoring by content rather than line, I mutated its successors instead — see M2/D1 below.

---

## Mutation — 6 runs, every one asserted applied by `git diff` **and** within-tree `GSharp.Core.dll` md5 movement

Pristine: `Calls.cs` `5d76cbe1`, `Calls.Invocation.cs` `a660188d`, `Core.dll` `8a20bd56`.

| # | site | change | `Core.dll` md5 | result |
|---|---|---|---|---|
| **M1** | `Calls.cs:1784` | `ShouldConvertToNominalDelegate(...)` → `!...` | `ffe97839` | **KILL 3** |
| **M2** | `Calls.cs:1910` | `if (sawOpenGenericParameter)` → `if (!...)` | `5174929f` | **SURVIVED 8,525** |
| **D1** | `Calls.cs:1896` | `nominalTarget = nominalCandidate;` → `= null;` | `47037d12` | **SURVIVED 8,525** |
| **M3** | `Calls.Invocation.cs:2297` | `(hasClosedTarget && sawOpen)` → `(... && !sawOpen)` | `a0af662b` | **KILL 3** |
| **C1** (control) | `Calls.Invocation.cs:1235` | `(fn, FromClrType(parameterType))` → `(fn, null)` | `ffd8bb99` | **KILL 3** |
| **T1** (test-side probe) | test `:379` | `if (!succeeded && !hitBoundary)` → `if (!succeeded)` | — | **4/4 still pass** |

After restore, a full solution rebuild returns `Core.dll` to **`8a20bd56` byte-identical**, and the tree is clean apart from the pre-existing untracked `packages.lock.json` (#3170).

**M3 is the round-2 hunk and it is genuinely pinned. M1 and C1 show the reification itself is pinned three ways.**

---

# SHOULD-FIX 1 — round 1's omission recurs: the new `nominalTarget` out-param is unpinned, and may be redundant

**`ExpressionBinder.Calls.cs:1795-1915`.** Two independent mutants say the same thing:

- **M2** inverts the successor of the guard round 1 flagged (`:1910`): **survived Core 7,128 + Interpreter 1,397 = 8,525 tests, zero deaths.**
- **D1** is stronger — it forces the out-param permanently null at its only producer (`:1896`), which disables the entire feature: **also survived 8,525.**

Meanwhile **C1** at `Calls.Invocation.cs:1235` kills 3. So **every reification these tests exercise flows through the deferred path** (`:1207-1256`), and **nothing reaches the resolver's new out-param.**

I then tried to fault my own finding rather than stop at the mutant. Rebuilt the compiler with D1 applied and re-ran all four hand-built probes, including one aimed at `BindCallArgumentWithDelegateTargetTyping`'s **only** live caller — the constructor path at `Calls.cs:1527`:

```
DuckHolder((value string) -> value.Length + 3).Raw.GetType().FullName
  pristine head: DuckLib.DuckGreeter      D1: DuckLib.DuckGreeter     ← identical
```

All four probes byte-identical with the feature disabled. The two live consumers are `Calls.cs:1771` and `Calls.Invocation.cs:2308`; in both, the candidate parameter is by construction a **closed** delegate type, and `SameDelegateIdentity` requires all matching candidates to agree — so whichever overload wins has that same nominal type and the ordinary argument conversion already reifies to it.

**The `/ponytail` resolution is deletion, not a test.** Do not write a test for ~25 lines that may do nothing:

> **Try deleting `nominalTarget` and its two guards (`:1892-1901`, `:1910-1913`) and rebuild. If nothing breaks, that is another ~25-line shrink on top of the one you already took. If something breaks, that failing case is precisely the missing test — write it.**

Either outcome closes this in one step. I am **not** claiming the code is provably dead — I claim I could not construct a program that observes it across 6 mutation runs and 4 probes, and the burden is now a single specimen.

**Severity rationale:** this is the same *class* of omission round 1 flagged, which is why it is called out. It is **not** a BLOCKER because round 1's hole was a **correctness** hole that silently miscompiled, and that hole is measurably closed on an independent fixture across all three drivers. This one conceals no wrong answer.

# SHOULD-FIX 2 — `CheckBareDriver` is a disjunctive oracle whose escape hatch is dead

**`Issue3149NamedDelegateReificationDriverTests.cs:357-386`.** The bare row passes if the driver **either** succeeds **or** exits 1 with `GS9999 … MetadataLoadContext`. Probe **T1** deletes the second disjunct: **4/4 still pass. It never fires.**

The comment at `:362-364` — *"Bare evaluation currently hits an unrelated MetadataLoadContext invocation boundary"* — is stale on this head. Measured: bare `gsc` **emits and executes**, and its `main` failure stack is

```
at MixedClosed.Probe.Run()
at MixedClosed.<Program>.<Main>$(String[] args)
```

an emitted entry point, exit **134**, not a `GS9999` refusal. Nothing in these four cases goes near a `MetadataLoadContext`.

Credit where due: the **#3130 miscitation round 1 found is gone** — `grep -n '3130'` over the PR's whole diff returns nothing. But the replacement should now go too. Delete `hitBoundary` and the comment (≈ −15 lines) and let `CheckBareDriver` collapse into `CheckDriver` modulo the trailing `Success.`. Leaving it is a live hazard: a future regression that turns these programs into compile-time `GS9999` failures would keep this test **green**.

# Observation, not scored — all three "drivers" now emit

Asked to re-establish the driver model empirically rather than trust it:

| driver | measured behaviour on `f16bc6a1` |
|---|---|
| bare `gsc file.gs` | **emits + executes**; prints program output then `Success.` |
| `gsc /target:exe /out:` | emits; run separately |
| `gsi file.gs` | **emits + executes** — failure stack shows `MixedClosed.<Program>.<Main>$` (#3182) |

So the three columns are three front-ends over one emit backend, and cross-driver parity buys less than the name `…AcrossDrivers` suggests. That is the ADR-0156 state of the world, **not this PR's doing**, and the rows are still real — `gsi` reproduces the bug on `main` (rc=134, same `ArgumentException`) and is fixed on head.

**A standing prior I have to correct against myself: `gsi` *does* accept `/r:` on this head.** `dotnet gsi.dll /r:DuckRef.dll mixed_closed.gs` → `DuckLib.DuckGreeter` / `43` / rc=0. The "gsi has no reference channel" finding from earlier sessions is stale, and it also holds on `917677ec`. The `gsi` row is a genuine user-reachable configuration, not a harness artifact — I had drafted that as a NIT and measurement removed it.

---

## Suites — denominator and source

**9 test projects**, source `dotnet sln GSharp.sln list` (CI's own per `build/generate-ci-test-matrix.py:26`). **I ran 3 of the 9 locally on the restored pristine head**, chosen rather than exhaustive:

| project | result |
|---|---|
| `Core.Tests` | **7,128 passed / 1 skipped** (3m32s) |
| `Interpreter.Tests` | **1,397 passed / 0 failed** (1m11s) |
| `Cs2Gs.Tests` | **1,907 passed / 0 failed** (7m39s) |
| **local total** | **10,432 passed, 1 skipped** |

`Cs2Gs.Tests` was run specifically because it drives real `gsc.dll` **out-of-process** and cross-assembly delegate reification is exactly that shape — a `src/`-only change here could redden it with nothing under `test/` noticing. It did not. The remaining 6 are covered by CI's 30/30 on this head, including `ilverify`, which is a live risk for a change feeding codegen and which passed.

*(Method disclosure — two no-results, both caught by the assert-applied rule, neither scored: D1's first `sed` targeted `:1895`, which is a bare `{`; it matched nothing, `git diff` was empty and both md5s stayed pristine, yet the filtered run returned a clean 8,525 green that would have read as a survival. Real line was `:1896`. Same failure mode the "mutant landed on a brace" hazard describes, and the md5 guard is the only reason it was distinguishable from the genuine D1 survival that followed.)*

DavidObando:
# Round 2 addendum — head moved after my review; **re-measured on `8aec592e`. Both SHOULD-FIXes adopted. Verdict re-affirmed: MERGEABLE, no open findings.**

## First, a bookkeeping warning about my previous comment

My round-2 verdict was measured on **`f16bc6a1`**. The head is now **`8aec592e`**, and

```
git merge-base --is-ancestor f16bc6a1 <new head>  ->  NOT an ancestor
```

The three PR commits were **rebased** onto `515ad40d` (#3201), so `f16bc6a1` is orphaned and my earlier comment's line numbers no longer resolve. Anyone reading this PR should treat that comment's *line citations* as stale while its *findings* stand. Verifying ancestry rather than existence is the only reason I caught it.

## Both SHOULD-FIXes were adopted — and the timeline is unambiguous

| event | UTC |
|---|---|
| my round-2 comment | **11:14:00Z** |
| `8aec592e` authored | **12:12:14Z** |
| `8aec592e` committed | 14:30:19Z |

**SHOULD-FIX 1 — the redundant nominal path is gone.** `nominalTarget`, the `SameDelegateIdentity` agreement block and the `sawOpenGenericParameter` reset are deleted from the resolver; both eager call sites now call `lambdas.BindLambdaExpression(lambdaSyntax, target)` directly. **−62 / +5.**

This is the outcome I asked for, and it was done **surgically** in a way that matches the measurements exactly. `BindLambdaWithDelegateTarget` was **retained** at `ExpressionBinder.Calls.cs:1772` because it is still called from the **deferred** path at `ExpressionBinder.Calls.Invocation.cs:1249`. That is precisely the split my mutants predicted: **D1 survived** (eager nominal path unobservable → safe to delete) while **C1 killed 3** (deferred path carries the live reification → must stay). The author kept the half that is load-bearing and deleted the half that was not. Nothing was deleted on my say-so that measurement did not justify.

**SHOULD-FIX 2 — the dead `GS9999` disjunct is gone.** `CheckBareDriver` collapsed to a one-line `CheckDriver("gsc", result, expectedOutput + "Success.\n", failures)`, taking the stale MetadataLoadContext comment with it. The suite is **4/4 green under the now-strict oracle**, which independently confirms probe T1: the hatch never fired.

## Re-measurement on `8aec592e` — I did not carry the old numbers forward

Fresh worktree, unique temp root, own `out/` verified isolated (`build/gsharp.build.props:69`), solution-level restore then `dotnet build` (0W/0E). Baseline `GSharp.Core.dll` = `453d58da`. **All md5s below come from the identical build command**, so artifact flavour is constant and the comparison is like-for-like.

| # | site (re-anchored by content, lines drifted) | mutation | `Core.dll` | result |
|---|---|---|---|---|
| **M3′** | `Calls.Invocation.cs:2296` | `(hasClosedTarget && sawOpen)` → `&& !sawOpen` | `3052b1d4` | **KILL 3** |
| **C1′** | `Calls.Invocation.cs:1235` | `(fn, FromClrType(parameterType))` → `(fn, (TypeSymbol)null)` | `d71555432` | **KILL 3** |

Both assert applied via **`git diff HEAD`** (not `git diff` — `git checkout -- <path>` stages, so plain `git diff` can read clean on a real change), both on genuine statements, both restored to `git diff HEAD` empty and `Core.dll` back to **`453d58da` byte-identical**.

M1 from my earlier comment **cannot be re-run: its target was deleted.** Stating that rather than quietly dropping it. The remaining kill-set is still non-empty and still hits the same three cases:
`MixedOpenAndClosedCandidates(selectOpen: False)`, `CrossAssembly…(generic: True)`, `(generic: False)`.

| suite on restored `8aec592e` | result |
|---|---|
| `Issue3149` filter | **4 / 4** |
| `Interpreter.Tests` | **1,419 passed / 0 failed** |
| `Core.Tests` | **7,128 passed / 1 skipped** |
| CI | **30 pass / 0 fail / 2 skipping**, **`ilverify` pass** (3m57s) |
| `mergeable` / `mergeStateStatus` | **MERGEABLE / CLEAN** |

The deletion regressed nothing.

## One NIT, source-confirmed, not scored against this PR

Every column in this test file is the **emitted** engine. On this head, `src/Repl/Program.cs:100` passes `engineChoice` to `RunInteractive`, but **`:109` drops it** — `RunScript(scriptPath, references)` takes no engine parameter. The harness's `RunInterpreter` calls `Program.Main(new[] { sourcePath })`, i.e. the script path, so it is `EmittedProgramHost` too.

So a file named `…AllDriversReifyAndInvoke` living in `test/Interpreter.Tests/` exercises the interpreter/evaluator in **none** of its three columns. That is the ADR-0156 Phase 3a state of the world, not this PR's doing, and the rows remain genuine regression pins — `gsi` reproduces the bug on `main` (rc=134, same `ArgumentException`) and is fixed on head. Worth a comment on the file so a future reader does not mistake the third column for evaluator coverage.

Correcting an earlier claim of my own while I am here: **`gsi` does accept `/r:`** on this head and on `main` (`gsi /r:DuckRef.dll mixed_closed.gs` → `DuckLib.DuckGreeter` / `43` / rc=0). My older "gsi has no reference channel" note was stale; I had drafted it as a NIT and measurement deleted it.

DavidObando:
## Round 3 (`8aec592e`) — **BLOCKER. Round 3 regresses the headline shape of #3149.**

Bounded re-confirmation. Answering the two asked questions, plus one thing I did not go looking for.

### Q1 — the dead disjunct: **deleted, cleanly**

`CheckBareDriver` (`test/Interpreter.Tests/Issue3149NamedDelegateReificationDriverTests.cs:357`) is now a one-line strict `CheckDriver("gsc", …)`. The `GS9999`/MetadataLoadContext escape hatch and its stale comment are gone. 4/4 green under the strict oracle.

### Q2 — was anything lost with it: **yes, and it is load-bearing**

No orphans (`BindLambdaWithDelegateTarget` still live at `ExpressionBinder.Calls.Invocation.cs:1249`; `SameDelegateIdentity` 3 callers; `ShouldConvertToNominalDelegate` 3). No test rows deleted. But the **`src/` half** of `8aec592e` — removing `nominalTarget` from `TryResolveDelegateTargetFromCandidates` and replacing both eager `BindLambdaWithDelegateTarget` calls with `lambdas.BindLambdaExpression` (`Calls.cs:1766`, `Calls.Invocation.cs:2303`) — **re-opens #3149 for every call site with no competing open-generic candidate.**

### Evidence — independent probe, 2 refs × 4 delegate shapes × 3 heads

Own reference assembly `DuckRef.dll` (plain classlib, no `ProjectReference` into the repo). Four G# consumers, all the issue's shape — lambda → `List[NamedDelegate]` → `.Add` → read back → invoke, cross-assembly via `/r:`:

| cell | slot type |
|---|---|
| A | `List[DuckGreeter]` (non-generic delegate, `string` param) |
| B | `List[DuckMapper[DuckItem]]` (generic delegate, imported class param) |
| C | `List[DuckMapper[string]]` |
| D | `List[DuckTaker]` (non-generic, imported class param) |

Two builds of the reference, identical but for one throwing open-generic extension `Add<T>(this List<DuckGreeter>, Func<T,int>)` — a copy of the fixture's `Issue3149MixedDelegateExtensions` at `…DriverTests.cs:25-29`.

| reference | `515ad40d` (merge-base, pre-fix) | `57014aad` (round 2) | **`8aec592e` (HEAD)** |
|---|---|---|---|
| **without** extension | A B C D → **134** | A B C D → **0 ✓** | A B C D → **134** |
| **with** extension | A B C D → **134** | A B C D → **0 ✓** | A B C D → **0 ✓** |

Round 2 fixed **8/8**. Round 3 fixes **4/8** and turns the other 4 back red. The only `src/` delta between those two heads is the `nominalTarget` removal (`git diff --stat 57014aad 8aec592e -- src/` → 2 files, +3/−36).

Failure on HEAD is the exact #3149 signature, all three drivers:

- `gsc /target:exe /out:` → **rc=0, zero diagnostics**, `Wrote …/CellA.dll`
- run the emitted assembly → **rc=134** `System.ArgumentException: The value "System.Func\`2[System.String,System.Int32]" is not of type "DuckRef.DuckGreeter"` at `List\`1.IList.Add`
- `gsi cellA.gs` → rc=134, byte-identical
- bare `gsc` → rc=134, byte-identical

#3149 says: *"A lambda is assigned to a named delegate type reached through an erased `List[...]` slot, across a project reference"* and *"a silent miscompile that defers to a runtime cast failure is the worst of the three outcomes."* That sentence still describes HEAD.

### Why the suite cannot see it

`Issue3149MixedDelegateExtensions.Add<T>` (`…DriverTests.cs:25-29`) is declared in `namespace GSharp.Interpreter.Tests`, which **every** consumer fixture imports. Measured: its presence flips cells **B, C and D** too — whose receivers are `List[DuckMapper[…]]` / `List[DuckTaker]`, which the extension cannot even bind to. So an open-generic candidate enters the candidate set of *every* `.Add` in the corpus, and **all four rows execute the mixed path**. There is no single-candidate row. Interpreter.Tests **1419/1419** green on HEAD, `Issue3149` filter **4/4** — and both are blind to this.

The corpus is a **census of the mixed shapes, not of #3149**.

### Disclosure — this is partly my fault

Round 2 SHOULD-FIX #1 (mine) recommended deleting the eager `nominalTarget` path, on the strength of mutant M2 and probe D1 surviving 8,525 tests. That survival was real and the inference was wrong: the path was unobservable **only because the fixture manufactures a mixed candidate set in every row**. My mutant inherited the corpus's blind spot. Recording it so the next reviewer does not repeat it: *a mutant can only be as good as the specimens it is scored against.*

### Minimal fix (ponytail)

1. **Revert the `src/` half of `8aec592e`** — restore `nominalTarget` and both eager `BindLambdaWithDelegateTarget` calls. Measured: that state is 8/8 green. Keep the **test half** (`CheckBareDriver`), which is a genuine improvement and independent of the regression.
2. **Move `Issue3149MixedDelegateExtensions` into a namespace only the mixed test imports**, so the non-mixed rows regain a single-candidate candidate set. Without that, the corpus cannot distinguish round 2 from round 3 — which is exactly what happened.

### Bookkeeping

- Head `8aec592e`; merge-base `515ad40d`; branch touches **no** `Core.Tests` file, so the #3221/#3225/#3228 `EmittedOracle` slices are not a conflict surface. `git merge-tree --write-tree` vs `main` `d6e7c4c3` → 1-line output `f5257b8c`, no conflicts.
- Round-2 head `f16bc6a1` is content-identical to `57014aad` on all three files (pure rebase).
- No source was mutated in this round: `GSharp.Core.dll` md5 `aedfd241…` unchanged start to finish; `git status` clean but for the pre-existing #3170 lockfile. Every result above is observation, not mutation.

DavidObando:
Round 4 blocker fixed at `f68198fdd`.

- Restored round-2 `nominalTarget` source exactly; round-3 commit is now test-only.
- Isolated mixed extension namespace. Single-candidate pin is RED 2/2 on broken round-3 source with exact #3149 `Func` → named-delegate `ArgumentException`, GREEN 2/2 after restore.
- Strict `CheckBareDriver` assertion remains.
- Independent no-extension/with-extension × A/B/C/D matrix: all three drivers rc=0 in all 8 cells; exact delegate identity and invocation output asserted.
- CI-flavour build: 0 warnings/errors; `GSharp.Core.dll` MD5 `ac4ddf02ff60af833d2f93fd086adef4`.
- Nine projects: 15,366 passed, 4 skipped, 0 failed. ILVerify 123/123.
- Final CI: 30 passed, 2 skipped, 0 failed. PR mergeable/clean.

Full commands, matrix, and RED/GREEN text are in the updated PR body.

DavidObando:
Rebase revalidation complete at `4615f92d5694f9030f40570f4bccf534c3ae49f3`. The requested `6bf63e333` was already an ancestor of live main, so the branch was rebased through current `3e6dd50a7`; patch ID remains `3dac539a24857c7208f54a13686eb28556c0c31d`.

- Broken round-3 source: pin RED 2/2 with exact `System.Func` → named-delegate `ArgumentException`; restored source: GREEN 4/4. Core MD5 moved `5d375632...` → `afc03bcb...` and restored exactly.
- Independent no-extension/extension × A/B/C/D matrix: 8/8 cells green across bare `gsc`, executed `gsc /out:`, and `gsi /r:`; exact delegate identity and result `43` asserted.
- CI-flavour build: 0 warnings, 0 errors; final `GSharp.Core.dll` MD5 `5d3756321419d11236441a50c946a18f`.
- Nine projects: 15,373 passed, 4 skipped, 0 failed. ILVerify: 123/123, two documented suppressions.
- CI run `30974896883`: 30 passed, 2 skipped, 0 failed. Merge-tree `bb1719edfc45b624df95b4e8aa64ce7d13d780b6`, no `CONFLICT`; PR is MERGEABLE/CLEAN.

DavidObando:
## Review — PR #3159 (`fix/3149-named-delegate-reification`), head `4615f92d5694f9030f40570f4bccf534c3ae49f3`

**VERDICT: MERGEABLE — 0 BLOCKER · 1 SHOULD-FIX · 2 NIT.**

Reviewed in a private worktree at head. `main` fetched twice: `3e6dd50a7` at start, **`d172d0d14`** at time of posting (`d172d0d14 Correct emitted oracle test intent (#3263)` landed mid-review). PR is 3 ahead / 1 behind. `git merge-tree --write-tree --messages origin/main 4615f92d` writes tree `0739d99d7` with **zero stage entries, zero `CONFLICT` lines, zero `Auto-merging` lines** against *both* mains, and `d172d0d14` touches no file this PR touches. Rebase not required.

Release CI-flavour build (`-p:ContinuousIntegrationBuild=true`): **0 warnings, 0 errors**. `GSharp.Core.dll` md5 **`5d3756321419d11236441a50c946a18f`** — byte-identical to the md5 published in the PR body, and reproduced independently in two separate worktrees.

---

### 1. `src/` half of the round-3 deletion is reverted — verified by identity, not by grep

```
git diff 57014aad 4615f92d -- \
  src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.cs \
  src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
→ EMPTY
```

The two files are **byte-identical** to `57014aad`. Corroborated three ways: the diff is empty; the `nominalTarget` occurrence lists match line-for-line; and the round-3 deletion patch (`git diff 57014aad 8aec592e`) applies to the head tree with `git apply --check` rc=0, which is only possible if head == `57014aad` at every touched hunk.

Restored sites (the brief's cites drifted by a few lines — corrected here):

| brief | actual at head |
|---|---|
| `ExpressionBinder.Calls.cs:1766` | `ExpressionBinder.Calls.cs:1763` (`out var nominalTarget`) and `:1771` (passed to `BindLambdaWithDelegateTarget`) |
| `ExpressionBinder.Calls.Invocation.cs:2303` | `ExpressionBinder.Calls.Invocation.cs:2264` (`nominalTarget: out var nominalTarget`) and `:2308` |

Better than a revert: the branch was **rebuilt** so the deletion never lands. Commit 3 `4615f92d5` is **test-only** (`Issue3149MixedDelegateExtensions.cs` +16, driver tests +4/−35). There is no deletion/revert pair in the history to audit.

### 2. Test half kept — the `GS9999` hatch is gone and stays gone

`Issue3149NamedDelegateReificationDriverTests.cs:350-354`:

```csharp
private static void CheckBareDriver(
    DriverResult result,
    string expectedOutput,
    List<string> failures) =>
    CheckDriver("gsc", result, expectedOutput + "Success.\n", failures);
```

`grep -n GS9999` over the file → **0 hits**. `CheckDriver` (`:356-372`) requires `ExitCode == 0` **and** exact-ordinal stdout equality **and** `StandardError.Length == 0`. The deleted `hitBoundary` disjunct — including its `(expectedOutput.Contains("System.Func") || !normalizedOutput.Contains("System.Func"))` clause — is entirely gone.

Note this deletion is safe **by construction**, not by mutation evidence: it removes an *acceptance* path, so it is monotonically strictening. It cannot mask a regression; it can only expose one. Measured both directions below (green at head, red under the deletion).

Also good: `CrossAssembly…` → `SingleCandidateNamedDelegateThroughErasedListSlot…`. The name now states the property the isolation actually establishes (ADR-0154).

### 3. Decontamination re-test — **it took**, and the mechanism is reproduced causally

The requested check plus its counterfactual. One variable at a time, same tree, same build flags, witness md5 asserted moving on every mutation and returning by exact round-trip:

| src | fixture reachable from the consumer program? | `SingleCandidate…` rows | `Mixed…` rows |
|---|---|---|---|
| head (fixed) | no (`…Tests.Issue3149Mixed`, not imported) | **2/2 PASS** | 2/2 PASS |
| round-3 deletion applied | no | **2/2 FAIL** | 2/2 PASS |
| round-3 deletion applied | **yes** — one line `import GSharp.Interpreter.Tests.Issue3149Mixed` re-added to the two G# consumer sources | **2/2 PASS** | 2/2 PASS |

Witness: `5d375632…` → `afc03bcb…` on mutation → `5d375632…` on restore.

Failure text under the deletion is #3149's exact signature, on all three drivers:

```
System.ArgumentException: The value "System.Func`2[System.String,System.Int32]"
  is not of type "GSharp.Interpreter.Tests.Issue3149Greeter"
```

Two conclusions, both load-bearing:

- The namespace move is **causally** what restored observability — re-adding a single `import` line to the G# sources flips the identical broken compiler from RED back to GREEN. This is a positive construction, not a survival argument.
- The `Mixed…` rows pass under the deletion. That is the proof that round 3's mutant survived because every specimen was forced onto the mixed path, exactly as the PR body states.

### 4. Independent 8-cell matrix — reproduced from scratch, each cell **executes**

Own fixture assembly `Duck.Fixtures` (`DuckGreeter`, `DuckMapper<T>`, `DuckTaker`, `DuckItem`), built in two variants: without and with an open-generic `Add<T>` extension over the four receiver types. 4 programs × 2 variants × 3 drivers = **24 executions**, each running the produced program (`gsc` bare; `gsc /target:exe /out:` **then `dotnet ./prog.dll`**; `gsi /r:`), asserting the printed delegate type identity and the invocation result `43`.

| reference | A `DuckGreeter` | B `DuckMapper<DuckItem>` | C `DuckMapper<string>` | D `DuckTaker` (void) |
|---|---|---|---|---|
| **head, without extension** | `gsc=0 ok / out=0 ok / gsi=0 ok` | ok / ok / ok | ok / ok / ok | ok / ok / ok |
| **head, with extension** | ok / ok / ok | ok / ok / ok | ok / ok / ok | ok / ok / ok |

**8/8 at head.** Control on pre-fix `main 3e6dd50a7`, same harness, same programs: **0/8 — rc=134 in all 24 runs**, e.g.

```
--- gsc-out rc=134 (compile rc=0)
Unhandled exception. System.ArgumentException: The value "System.Func`2[System.String,System.Int32]"
  is not of type "Duck.Fixtures.DuckGreeter" ...
   at System.Collections.Generic.List`1.System.Collections.IList.Add(Object item)
```

Note `compile rc=0` with zero diagnostics, then rc=134 at run — the rc=0 disguise reproduced. The harness has teeth; 8/8 is a result, not a no-result.

### 5. Nothing else regressed

Locally reproduced at head, CI flavour, and the counts match the PR body exactly:

| project | measured | body |
|---|---|---|
| Compiler.Tests | **4268 passed, 0 failed** (28m37s) | 4,268 |
| Core.Tests | **7178 passed, 1 skipped, 0 failed** (3m34s) | 7,178 / 1 |
| Interpreter.Tests | **1349 passed, 3 skipped, 0 failed** (1m46s) | 1,349 / 3 |

12,795 of the claimed 15,373 tests, covering every project that could observe a binding change. Remaining 6 of 9 covered by CI: **30 pass / 2 skipping**, including `ilverify`, `e2e`, `cs2gs-oahu`.

The restructure at `Invocation.cs:2258` is inside `else if (inner is LambdaExpressionSyntax)` (`:2205`), so hoisting `TryResolveDelegateTargetFromCandidates` out of the short-circuited `&&` adds no call for non-lambda arguments. The new `else if (… && hasClosedTarget)` / `else` pair is an inlining of `BindCallArgumentWithDelegateTargetTyping`'s own body plus the nominal conversion — semantically the same as `main`'s single `BindCallArgumentWithDelegateTargetTyping` call. `InspectProducedDelegate` (`:240-307`) is a genuine oracle: collectible ALC, reflective `Probe.Create()`, asserts runtime delegate identity **and** `DynamicInvoke` result — not a coarser observable.

---

## SHOULD-FIX 1 — two of this PR's own new decision points are unpinned by its own corpus

Mutation audit at head, harness teeth demonstrated first (round-3 deletion → RED, M1 → RED), each mutant built CI-flavour with the witness asserted moving:

| mutant | site | witness | Issue3149 corpus |
|---|---|---|---|
| M1 `(hasClosedTarget && sawOpen)` → `(hasClosedTarget && !sawOpen)` | `Invocation.cs:2296` | `e30a1f18…` | **KILLED** — `MixedOpenAndClosedCandidates(selectOpen: False)` fails |
| M2 delete `if (sawOpenGenericParameter) { nominalTarget = null; }` | `Calls.cs:1910-1913` | `25df84fc…` | **SURVIVES 4/4** |
| M3 `!SameDelegateIdentity(…)` → `SameDelegateIdentity(…)` | `Calls.cs:1898` | `6da9a774…` | **SURVIVES 4/4** |

**I am explicitly not calling M2 or M3 dead.** That inference is what produced round 3. Instead, by write-set:

- `TryResolveDelegateTargetFromCandidates` has exactly **two** callers: `Calls.cs:1757` and `Invocation.cs:2258`.
- `BindCallArgumentWithDelegateTargetTyping` (which wraps `Calls.cs:1757`) has exactly **one** caller: `Calls.cs:1527`, the CLR **constructor** path.
- The `Invocation.cs:2258` caller consumes `sawOpen` and **defers**, discarding `nominalTarget`. So the `:1910` reset can only ever be *observed* through the ctor caller — which passes `out _` for `sawOpen`.
- Therefore **the `:1910` reset is the constructor path's counterpart of the invocation path's deferral.** Both exist to stop the same wrong eager conversion on a mixed open/closed candidate set. It must be kept.

Specimens I built and ran, both non-discriminating — reported honestly rather than converted into a deadness claim:

- **For M2:** imported `SpecBox<T>` with ctors `(SpecGreeter)` and `(Func<T,int>)`, constructed from a G# `func Make[T any]()`. M2 / clean head / pre-fix `main` all print `greeter:SpecGreeter`. G# reifies `T` before ctor candidate collection, so `clrType` is closed and `sawOpen` is never set on this route. Open `clrType` is nonetheless a real shape in this method's neighbourhood — `Calls.cs:1653` guards `!clrType.ContainsGenericParameters` explicitly — so I cannot assert unreachability, only that I did not reach it.
- **For M3:** `Take(SpecGreeterA, int)` / `Take(SpecGreeterB, string)` — same shape `(string)->int`, different nominal identity, disambiguated by the second parameter. M3 / clean head / pre-fix `main` all print `B:SpecGreeterB:43`; the downstream overload resolution corrects the eager conversion.

**Ask:** add a short source comment at `Calls.cs:1910` and `Calls.cs:1898` recording the write-set argument and that neither is corpus-observable, so the next reviewer's mutation survival cannot be re-read as deadness. Given this PR's entire round-4 thesis, that is cheap and it closes the loop.

## NIT 1 — the isolation is not self-defending

`Issue3149NamedDelegateReificationDriverTests.cs:85` is the single `import GSharp.Interpreter.Tests.Issue3149Mixed` that keeps the mixed test mixed; `CreateConsumerSource` (`:187`) stays clean only by omission. Nothing prevents a future author from re-adding that import, or re-parenting the extension, and silently re-blinding the corpus — I demonstrated above that exactly one line does it. A one-line `Assert.DoesNotContain("Issue3149Mixed", source)` in the single-candidate test would make the isolation assert itself.

## NIT 2 — child namespace rather than sibling

`Issue3149MixedDelegateExtensions.cs:8` places the fixture in `GSharp.Interpreter.Tests.Issue3149Mixed`, a **child** of the namespace every fixture imports. It is measurably not auto-imported (item 3 proves it), so this is correct today, but a top-level sibling (`GSharp.Tests.Issue3149Mixed`) would remove the question rather than answer it.

---

**Ponytail note.** +616/−41 with 502 lines in one test file is not bloat here: that file is the only thing producing a runtime oracle across three drivers plus reflective delegate-identity inspection, and it demonstrably kills the regression on all three. The `src/` change is **+98 / −41** across two files (net +57), and I found no smaller shape — commit 1 alone (nominal identity only) would eagerly convert on mixed candidate sets, which commit 2's `sawOpen` deferral exists to prevent, and M1 kills that deferral.

DavidObando:
## Final rebase validation

The corrected `5e2fee875` is an ancestor of fetched live main. Final branch is rebased through `e28d55bdf` (including #3244, #3253, #3261, and #3263).

- head: `907f6ed26bf18c6ea66af9a1c79c8be4ff69d25a`
- patch ID: `7d536bd9ccdd0e2cf39ff1794ea4d1925805179a` (stable across final rebases)
- no `SessionEngine`, `Compilation.Evaluate`, evaluator engine, retired diagnostic, or stale `interpreted` test-harness reference in the PR diff
- merge-tree: `75283d494e74f41a612ad2efb8308c2e69035911`; clean merge produced no unmerged stages

### Product evidence

- D1 (`nominalTarget = null`): Core MD5 moved `5cb7a90f...` → `f3c51593...`; both single-candidate pins RED with the exact `System.Func<...>` where `Issue3149Greeter` / `Issue3149Mapper<Issue3149Item>` was declared `ArgumentException` (2/2 failed).
- exact restoration: source SHA round-trip and Core MD5 returned to `5cb7a90fd1042eb7570c266f0e406c16`; pins GREEN 2/2. Full issue filter GREEN 4/4.
- independent reference matrix: without extension A/B/C/D GREEN; with extension A/B/C/D GREEN. Every cell executed bare `gsc`, `gsc /out:` output, and `gsi /r:`; exact delegate identity and result `43` asserted. **8/8**.
- strict `CheckBareDriver` remains success-only; no `GS9999` hatch.

### Gates

- CI-flavour solution build: 0 warnings, 0 errors
- nine projects: **15,373 passed, 4 skipped, 0 failed**
- ILVerify: **123/123**, two documented suppressions
- CI `30989699047`: **30 passed, 2 skipped, 0 failed**
- PR: **MERGEABLE / CLEAN**

DavidObando:
## Re-review at rebased head `907f6ed26` — **MERGEABLE** (0 BLOCKER · 2 SHOULD-FIX · 2 NIT)

Prior review was at `4615f92d5`. Per instruction I carried **no** green result across the rebase: every number below was re-measured in fresh worktrees at `907f6ed26`.

`main` fetched by me: **`e28d55bdfc83ac78dcbd0e9f581aaf7f0f0326df`**. Merge-base(`main`, head) == `e28d55bdf` **exactly** → 3 ahead / **0 behind**. `git merge-tree --write-tree --messages` → tree `f9f09a5608…`, **0 stage entries, 0 `CONFLICT` lines** (read by stages; git 2.50 never emits `<<<<<<<`). CI 30 SUCCESS / 2 SKIPPED / 0 failing.

`main` moved again while I measured, to **`0232ecb5eb7c1f6878556599a6faf557487e4232`** (`Correct stale interpreter test intent (#3269)`). Re-ran the merge check against it: **identical** tree `f9f09a5608…`, 0 stage entries, 0 `CONFLICT`. #3269 renames two other `Interpreter.Tests` files and touches **none** of this PR's four. PR is 3 ahead / 1 behind; no rebase required.

### Rebase delta is a rename, and the identical `src` is *correct*, not a dropped merge

| file | old head `4615f92d5` vs new head |
|---|---|
| `ExpressionBinder.Calls.Invocation.cs` | **blob-identical** |
| `ExpressionBinder.Calls.cs` | **blob-identical** |
| `Issue3149MixedDelegateExtensions.cs` | **blob-identical** |
| `Issue3149NamedDelegateReificationDriverTests.cs` | changed |

The test-file change is purely `RunInterpreter` → `RunScriptDriver` (+ the local `interpreted` → `script`); the call is still `GSharp.Repl.Program.Main(new[] { sourcePath })` and no assertion moved. That tracks main's #3261. My own rebuilt `gsi --help` confirms `--engine` accepts only `'emit'` and the evaluator was removed in ADR-0156 Phase 3c, so the new name is the accurate one.

Identical `src` blobs across a rebase are normally a silent-revert smell, so I checked the base: `git diff --numstat 3e6dd50a7 e28d55bdf -- <both src files>` is **empty** — main never touched them. Correct replay.

---

## Q1 — Does the decontamination work? **YES. The mutant that previously survived is now RED.**

### Positive controls first (both required before I would read any result)

| control | site | witness md5 | Issue3149 filter | matrix |
|---|---|---|---|---|
| baseline | — | `5e2754373a07bf9dca66e9d84a7de7c2` | 4/4 pass | **24/24** |
| **PC-1** throw at top of mutated method | `ExpressionBinder.Calls.cs:1795` | `5f97a468b240…` | **4/4 FAIL** | 0/24 |
| **PC-2** throw on the *rescued branch itself* | `ExpressionBinder.Calls.cs:1784-1786` | `234fdb898eed…` | **3 FAIL / 1 pass** | 0/24 |

Both surfaced as `GS9998: InvalidOperationException: PC1-REACHED` / `PC2-REACHED`. The site is reached; a survival here would be a survival, not a no-result.

**PC-2 is the one that matters.** It throws exactly where the round-3 deletion removed behaviour — the `nominalTarget != null && ShouldConvertToNominalDelegate(nominalTarget)` conversion arm. Its result is a 4-cell truth table on a single line:

| test | PC-2 |
|---|---|
| `SingleCandidate…(generic: False)` | FAIL → takes the branch |
| `SingleCandidate…(generic: True)` | FAIL → takes the branch |
| `MixedOpenAndClosed…(selectOpen: False)` | FAIL → takes the branch |
| `MixedOpenAndClosed…(selectOpen: True)` | **PASS → correctly does not** |

The corpus now observes that line in **both polarities**. Under the old universally-imported fixture it could observe it in neither.

### The mutant, re-planted

The round-3 deletion patch (`git apply --check` rc=0 against this head, confirming the `src` regions are unchanged):

| | witness md5 | filter | matrix |
|---|---|---|---|
| round-3 deletion | `a4765d6410da…` | **2/4 FAIL** — both `SingleCandidate…` | **12/24** |

Failing text is #3149 verbatim:

```
Unhandled exception. System.ArgumentException: The value "System.Func`2[System.String,System.Int32]"
  is not of type "Duck.Fixtures.DuckGreeter" and cannot be used in this generic collection.
   at System.Collections.Generic.List`1.System.Collections.IList.Add(Object item)
```
…after `compile rc=0` → run **rc=134**. The zero-diagnostic disguise reproduced.

### The mechanism, isolated to one variable

The matrix splits the *same mutant binary* by whether the open-generic `Add<T>` is in scope:

| deletion applied, single binary | cells |
|---|---|
| `noext` (contaminant **not** in scope) | **0/12 — all rc=134** |
| `ext` (contaminant in scope) | **12/12 pass** |

And the corpus-level counterfactual, same broken compiler, one variable changed — I added `import GSharp.Interpreter.Tests.Issue3149Mixed` to the two clean branches of `CreateConsumerSource` (`Issue3149NamedDelegateReificationDriverTests.cs:187`):

| src | contaminant visible to clean consumers | filter |
|---|---|---|
| head (fixed) | no | 4/4 PASS |
| round-3 deletion | no | **2/4 FAIL** |
| round-3 deletion | **yes (+2 `import` lines)** | **4/4 PASS** |

Two `import` lines flip the identical broken compiler from RED to GREEN. **The namespace is the causal variable.** The earlier false survival is explained and closed.

Witness returned to `5e2754373a07bf9dca66e9d84a7de7c2` exactly after every mutation; `git diff HEAD` empty each time.

---

## Q2 — Is the hard case still constructed, or did the test stop testing?

Still constructed, proven three independent ways:

1. `Issue3149MixedDelegateExtensions.Add<T>` **throws if selected** — the wrong candidate is a runtime abort, not a silent pass.
2. The two arms assert **different** outputs (`Issue3149Greeter`/`43` vs `Func\`2[System.String,System.Int32]`/`53`), so neither can pass vacuously.
3. Reverting commit 2's `src` breaks **exactly** `MixedOpenAndClosed…(selectOpen: False)` with #3149's signature (below). A test that had stopped constructing the hard case could not detect that.

## Q3 — Per-hunk reverts: both hunks load-bearing, on disjoint halves

Whole-file revert of either product file **does not compile** — `CS7036` at `ExpressionBinder.Calls.Invocation.cs:2274`, the `nominalTarget` out-parameter. The two files are compile-coupled, so I reverted by commit instead (`src` only, tests kept):

| reverted | witness | filter | matrix `noext` | matrix `ext` |
|---|---|---|---|---|
| commit 1 half (round-3 deletion) | `a4765d64…` | 2/4 FAIL (`SingleCandidate…`) | **0/12 rc=134** | 12/12 ok |
| commit 2 `src` (`6f6529480`) | `6949b674…` | 1/4 FAIL (`Mixed…(selectOpen: False)`) | 12/12 ok | **0/12 rc=134** |

Exact mirror images. Commit 1 owns the single-candidate half; commit 2's `sawOpen` deferral owns the mixed half. **No hunk is decorative** — there is no "reverting this changes nothing" hunk to justify.

---

## SHOULD-FIX 1 — two of this PR's own new decision points remain unpinned (unchanged, re-measured here)

| mutant | site | witness moved | result |
|---|---|---|---|
| delete `if (sawOpenGenericParameter) { nominalTarget = null; }` | `ExpressionBinder.Calls.cs:1910-1913` | `8e8f98d318ca…` | **SURVIVES** 4/4 + 24/24 |
| `!SameDelegateIdentity(…)` → `SameDelegateIdentity(…)` | `ExpressionBinder.Calls.cs:1898` | `f0b570caca46…` | **SURVIVES** 4/4 + 24/24 |

**These are unpinned, NOT dead — and this PR exists because that distinction was collapsed once already.** PC-1 proves the enclosing method is reached and PC-2 proves the consumer branch is reached, so the sites are live; the corpus simply contains no specimen that discriminates them. By write-set: `TryResolveDelegateTargetFromCandidates` has exactly two callers (`Calls.cs:1757`, `Invocation.cs:2258`); the wrapper around `:1757` has exactly one caller, `Calls.cs:1527`, the CLR constructor path; the `Invocation.cs` caller consumes `sawOpen` and defers, discarding `nominalTarget`. So `:1910` is the constructor path's counterpart of the invocation path's deferral. I built specimens for both (mixed-ctor `SpecBox<T>`; same-shape/different-identity `Take` overloads) and **neither discriminated**.

Ask: a source comment at both sites recording the write-set argument and that neither is corpus-observable, so the next reviewer's mutation survival cannot be re-read as deadness. **Do not delete either on survival evidence.**

## SHOULD-FIX 2 — the published witness md5 is not reproducible

The evidence comment cites `GSharp.Core.dll` md5 `5cb7a90fd1042eb7570c266f0e406c16`. I cannot reproduce it under either flavour at this head (SDK 10.0.300):

| build | md5 |
|---|---|
| `-p:ContinuousIntegrationBuild=true`, worktree A | `5e2754373a07bf9dca66e9d84a7de7c2` |
| same flag, **different path** (worktree B) | `5e2754373a07bf9dca66e9d84a7de7c2` — path-independent |
| **no** CI flag | `20f4368701072689c6f36749b0df798c` |
| published | `5cb7a90fd1042eb7570c266f0e406c16` |

The CI flavour is deterministic and path-independent, so it *is* a portable witness; the non-CI flavour embeds absolute paths and is not. `dotnet test` rebuilds without the flag. I reproduced this author's witness exactly at the previous head, so the toolchains do match; the likely cause is an md5 taken after a `dotnet test` rebuild. A different SDK patch under `rollForward: latestFeature` is an alternative I cannot exclude.

Ask: state the exact build command beside the md5, or take it from a `-p:ContinuousIntegrationBuild=true` build. Costs nothing and makes the witness checkable. **No impact on the fix.**

## NIT 1 — the isolation is not self-defending
`Issue3149NamedDelegateReificationDriverTests.cs:85` is the single `import …Issue3149Mixed`; `CreateConsumerSource` (`:187`) stays clean only by omission. I measured that one added line re-blinds the corpus. An `Assert.DoesNotContain("Issue3149Mixed", source)` in the single-candidate path would make the property enforced rather than incidental.

## NIT 2 — child namespace rather than sibling
`Issue3149MixedDelegateExtensions.cs:8` uses `GSharp.Interpreter.Tests.Issue3149Mixed`, a child of the universally-imported namespace. Measurably safe today (C#/G# do not surface child namespaces without an explicit import). A top-level sibling would remove the question instead of answering it.

---

### Regression
Rebuilt Release, **0 warnings / 0 errors**. Suites at this head, CI flavour: Interpreter.Tests **1350 pass / 3 skip / 0 fail**; Core.Tests **7178 pass / 1 skip / 0 fail**; Compiler.Tests **4267 pass / 0 skip / 0 fail** (26m12s). That is 12,795 passed and **4 skipped** across these three; the 4 skips match the PR body's `4 skipped` for the full 9-project run exactly. Remaining 6 projects covered by CI (30 SUCCESS).

Verified structural claims at head: strict `CheckBareDriver` → `CheckDriver("gsc", result, expectedOutput + "Success.\n", failures)` at `:350-354`; `GS9999` escape hatch **0 occurrences**; fixture namespace `GSharp.Interpreter.Tests.Issue3149Mixed` at `Issue3149MixedDelegateExtensions.cs:8`.

PR #3172, opened 2026-08-03
Align interpreter deinit test names

### LINKED ISSUE #3161: [test] Interpreter deinit reachability test pins nothing about reachability — escaping and non-escaping are byte-identical
## Summary

`Issue2988DeinitInterpreterTests.cs:142` — `EscapingInstanceIsNotFinalizedAtScopeExitOrWhileReachable` — has a name that claims a **reachability** property, but nothing in the test pins reachability. ADR-0068 guarantees the interpreter never runs deinitializers at all, so the output assertion cannot fail for the reason the name gives.

## Measured, not reasoned

Escaping vs non-escaping receivers under `gsi`:

| variant | stdout | GS0510 count |
|---|---|---|
| escaping | `made-22 / after-collect-33 / kept` | 1 |
| non-escaping | `made-22 / after-collect-33 / kept` | 1 |

**Byte-identical.** The qualifiers in the test name — "escaping", "at scope exit", "while reachable" — are pinned by nothing. The only load-bearing assertion is the single-`GS0510` check, which is about the warning being emitted once per declaring class, not about reachability.

## Severity: partial, not vacuous

This is weaker than #3150. The test **would** die if `gsi` started running deinitializers at scope exit, so it is not inert. But it does not discriminate escaping from non-escaping, which is the property its name advertises.

## Why this matters

Seventh instance this session of **a test asserting the observable symptom rather than the underlying claim** (cf. #3140, #3142, #3126, #3151, #3128, #3150). The recurring shape: the test name states a semantic property, the assertion checks a coarser observable that is constant across the property's two sides, and the test passes in both worlds.

A name that over-claims is a maintenance hazard: the next person to change interpreter reachability semantics will see a green suite and a test whose name says the behaviour is covered.

## Suggested resolution

Either:
1. **Rename** to what it actually pins (single GS0510 per declaring class under the interpreter), or
2. **Add the discriminating assertion** so escaping and non-escaping produce different observables — which, given ADR-0068's interpreter boundary, likely means asserting on something other than deinit output.

Option 1 is the `/ponytail` answer unless there is a real reachability property the interpreter is supposed to honour.

## Reference

`docs/adr/0068-deinit-destructor-support.md:109-114` — `### Interpreter boundary`: deinit "runs only in emitted code"; interpreted execution "reports warning GS0510 once for each declaring class and skips its deinitializer."

DavidObando:
## Summary

- rename the forced-collection interpreter test to the behavior its assertions observe
- rename the script-runner sibling to avoid claiming an untested renderer implementation

ADR-0068 specifies that interpreted deinitializers are skipped, so this is a naming/coverage correction, not an interpreter behavior change. The original reachability test was partial rather than vacuous: it still rejected scope-exit deinitializer execution.

## Resolution

Rename: interpreter has no reachability contract; existing assertions pin deinitializer skipping, one `GS0510`, preserved output, and stderr routing.

Fixes #3161

### DIFF
--- a/test/Interpreter.Tests/Issue2988DeinitInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue2988DeinitInterpreterTests.cs
@@ -139,7 +139,7 @@ public class Issue2988DeinitInterpreterTests
     }
 
     [Fact]
-    public void EscapingInstanceIsNotFinalizedAtScopeExitOrWhileReachable()
+    public void ForcedCollectionSkipsDeinitializerAndWarnsOnce()
     {
         var source = """
             import System
@@ -190,7 +190,7 @@ public class Issue2988DeinitInterpreterTests
     }
 
     [Fact]
-    public void ScriptRunnerUsesRichRendererForBoundaryWarning()
+    public void ScriptRunnerWritesBoundaryWarningToStandardError()
     {
         var sourcePath = Path.Combine(AppContext.BaseDirectory, "issue2988-deinit.gs");
         File.WriteAllText(


DavidObando:
## **MERGEABLE.** No product change, no coverage lost, no language regression. One SHOULD-FIX: the replacement name at `:142` still carries a qualifier I measured to be inert.

Anchors: live `origin/main` **`87af33c9`** (not `6267cd44` — it moved during review), head **`819e1847`**, base `6267cd44`. `git merge-tree --write-tree 87af33c9 819e1847` **output** = single line `4f46274d`, **0 conflict markers**. Build rc=0, **0 warnings, 0 errors**.

---

## The regression you most feared is structurally impossible here

ADR-0068 (`docs/adr/0068-deinit-destructor-support.md:109-114`) specifies that `gsi` never runs deinitializers. A "fix" that made it run them would be a language regression. **This PR cannot contain one:**

```
git diff --name-only 6267cd44 819e1847 -- src/ docs/   ->  (empty)
```
Test-only, one file. Confirmed live anyway — the interpreter still skips and still warns:

| driver | stdout | GS0510 |
|---|---|---|
| bare `gsc` (evaluate) | `made-22 / after-collect-33 / kept` | 1 |
| `gsc /out:` + run (**oracle**) | `made-22 / after-collect-33 / kept` | 0 |
| `gsi` (interpret) | `made-22 / after-collect-33 / kept` | 1 (stderr) |

## Coverage preservation — proven by md5, not by reading the diffstat

Strip only the two renamed signature lines from each side:

```
base 6267cd44 :  dfdb5da40bcba7c442f214f5b6b5a6c6
head 819e1847 :  dfdb5da40bcba7c442f214f5b6b5a6c6   <- identical
```
239 lines both sides, 5 `[Fact]`/`[Theory]` both, **29 `Assert.` both**. Nothing deleted, nothing weakened, two identifiers changed. Repo-wide grep for both old names outside the file returns **nothing**, so the rename has zero blast radius.

---

## Mutation — three PRODUCT mutants in `src/`, each asserted applied by `git diff` **and** md5

Baseline **7/7** (4 Facts + 3 Theory cases).

| mutant | site | validates | result |
|---|---|---|---|
| **M1** | `src/Repl/Program.cs:61` `Console.Error` → `Console.Out` | hunk 2's `…ToStandardError` | **RED — 2 failed**, incl. **`ScriptRunnerWritesBoundaryWarningToStandardError`** |
| **M2** | `Compilation.cs:337` drop the `ReferenceEquals(deinitializer?.Function, function)` conjunct | once-per-declaring-class | **RED — 3 failed**; renamed `:142` test **SURVIVED** |
| **M3 (control)** | `Compilation.cs:332` invert `!allErrors.Any()` | GS0510 emission at all | **RED — 6 of 7**, incl. `:142`; sole survivor `InheritedDeinitializers…(CompilerEmission)` |

Restore is an exact md5 round-trip on both files (`a7fdd106…`, `a4318e16…`), `dirty=0`, and the rebuilt `GSharp.Core.dll` returns to `3d6c68dc4e616fde9dc322cf37361e23` **byte-identical**.

**M3's lone survivor is the point of the control**: the emission driver is the one that never emits GS0510, so it is exactly the case that *should* be immune. A control that killed all 7 would have proven nothing about targeting.

**M1 settles hunk 2: `ScriptRunnerWritesBoundaryWarningToStandardError` is fully earned.** Both directions are pinned — `:220` requires the warning in stderr, `:219` `Assert.Equal("body-33\n", output.ToString())` requires it absent from stdout. And the old name was not merely unpinned but *unsupported*: there is exactly one `WriteDiagnostics` implementation (`src/Core/IO/TextWriterExtensions.cs:115`), shared by compiler and REPL, so "rich renderer" named a selection with no alternative to select. The only thing `:61` chooses is the `TextWriter`. The new name names it. **No change requested on hunk 2.**

---

## SHOULD-FIX — `:142` `ForcedCollectionSkipsDeinitializerAndWarnsOnce`: the fixture is ADR row 3, the name describes ADR row 1

The rename is a real improvement — dropping "Escaping…OrWhileReachable" removes a *comparative* claim the test never discriminated, which is what #3161 asked for. But two of the three conjuncts in the replacement do not survive measurement.

**"ForcedCollection" is provably inert.** Delete `GC.Collect()` / `GC.WaitForPendingFinalizers()` (`:160-161`) from the fixture and re-run:

```
emit stdout:  identical    gsi stdout:  identical    GS0510: still 1
```

**"SkipsDeinitializer" is true but not an interpreter property.** The emitted assembly — the oracle — prints byte-identical stdout with no `deinit-11`. The skip is not the boundary; it is ordinary CLR behaviour for a reachable object.

Cause: the fixture holds the instance alive. `var held = Make()` at `:157` and `Console.WriteLine(held.Tag)` at `:164` keep it reachable **across** the collect, so the forced collection has nothing to collect. That the fixture *could* have discriminated is easy to show — let the object die instead:

```
func Make() { var r = Res("kept"); Console.WriteLine("made-22"); GC.KeepAlive(r) }
Make(); GC.Collect(); GC.WaitForPendingFinalizers()

emit:  made-22 / deinit-11: kept / after-collect-33     <- deinit RUNS
gsi:   made-22 / after-collect-33                       <- deinit SKIPPED
```

Those are ADR-0068's own table rows. The renamed test's fixture is **row 3** ("Object remains reachable at exit → no `deinit` in either"). The name describes **row 1** ("Object dies…collection forced"). My measurements reproduce both rows exactly, so the ADR already documents that this fixture cannot show a driver difference.

M2 adds the third data point: dropping the once-per-class guard killed `DeinitializersWarnOncePerDeclaringClassWithoutRunning` (`:21`) and both interpreter/evaluate cases of the Theory, but **not** `:142`. The `Assert.Single` at `:170` is "exactly one GS0510 in a one-class cell" — real (M3 kills it) but not the once-*per-declaring-class* property, which `:21` owns by name and by mutation.

So of the three conjuncts, the load-bearing one is **`AndWarnsOnce`**, and it is last.

**Recommended fix — pick either, both are small:**
1. **Name the causal property the rename dropped**, e.g. `ReachableInstanceWarnsOnceWithoutRunningDeinitializer`. Reachability is *why* nothing finalizes; forced collection is the part that does nothing.
2. **#3161's own option 1** — name it for GS0510 and stop mentioning collection at all.

I am **not** recommending you delete the `GC.Collect()` lines even though they are inert. There is a defensible reason to keep them: they are a provocation against a future change that wires interpreter finalization to GC. That is an argument for *keeping the code*, not for *foregrounding it in the name* — especially as the provocation is weak here (a reachable object would not finalize even under a correct GC-wired implementation), and the strong version already exists at `:71`, across all three drivers.

**Why this is SHOULD-FIX and not a blocker:** the deliverable of this PR is a name, so residual inaccuracy in the replacement is in scope — but every assertion is unchanged and sound, the test is demonstrably coupled to the product (M3), and the PR is strictly better than what it replaces. It does not need to be right before merging; it needs to not be wrong in the same way twice.

## NIT — hunk 2 is outside #3161's stated scope

#3161 names only `:142`. `:193` is a second, independently-discovered instance of the same defect class. It is disclosed in the PR summary, it is two identifiers, and I measured it correct. Noting it for the record rather than objecting.

---

## Suites — and the denominator I actually ran

`dotnet sln GSharp.sln list` reports **9** test projects (CI's own denominator per `build/generate-ci-test-matrix.py:26`). **I ran 1 of those 9 locally, deliberately, not 9:**

- **Interpreter.Tests — 1264 passed, 0 failed, 0 skipped** (1 m 42 s), the project containing the change.

The justification for not running the other eight is measured, not assumed: the diff touches **zero** files under `src/` (md5-identical bodies, empty `git diff -- src/ docs/`), and a repo-wide grep finds **no reference to either old name** anywhere outside the renamed file — so no other project can observe this change. Stating this plainly so the gap is auditable rather than implied.

**CI on this head is 31 checks pass / 0 fail** (2 `skipping`), covering all 9 projects. That includes **`ilverify` — pass, 3m11s**. Worth recording since new gates were flagged as a hazard: `9003a86c` is already an ancestor of this PR's base `6267cd44`, so the gate ran, and it is green. `mergeStateStatus: BLOCKED` here is the review requirement, not CI.

*(Method disclosure: my first attempt at control M3 targeted line 333 and the `sed` silently matched nothing — `git diff` empty, md5 unmoved. The filtered run then reported a clean **7/7**, which would have read as "the control survived" had I not asserted application first. The guard is the only reason it was caught; the real line was **332**, and a global un-anchored `sed` would instead have hit both `:332` and `:353`. A no-result is not a survival.)*

PR #3188, opened 2026-08-04
Handle jagged delegate arrays in explicit generic closure

### LINKED ISSUE #3178: [bug] Jagged delegate arrays still fail explicit generic calls with GS0159 — re-closure unwraps only one array level
Follow-up from review of #2903 (issue #2857). The fix there re-closes over user-reference placeholders when a generic-dependent parameter is a delegate **or a delegate array**, but it unwraps exactly **one** array level, so jagged delegate arrays fall outside the guard and still hit the very `GS0159` that PR exists to remove.

## Repro

Cross-assembly, `gsc /out:` with `/r:`:

```
func Make[T Base init()](configure [][]((T) -> void)) T
```
called as `Make[Derived](...)`.

| probe | base | after #2903 |
|---|---|---|
| `Make[Derived]` with `[][]((T) -> void)` param, cross-assembly | `GS0159` | **`GS0159` — still fails** |

Single-level `[]((T) -> void)` works after #2903. The jagged form does not.

**This is absence, not syntax:** the same jagged declaration compiles fine same-assembly (`Wrote s.dll`), so the declaration is well-formed and the failure is in re-closure, not parsing. Positive control required in any fix.

## Cause

`OverloadResolution.cs`, the hunk added by #2903:

```csharp
p.ParameterType.IsArray ? p.ParameterType.GetElementType() : p.ParameterType
```

For `[][]((T) -> void)` this yields `((T) -> void)[]` — still an array, not a delegate — so the delegate guard is false and re-closure never runs.

## Remedy, already measured

The reviewer did not propose this, they **ran** it:

```csharp
private static Type UnwrapArrayElement(Type type)
{
    while (type is not null && type.IsArray)
    {
        type = type.GetElementType();
    }

    return type;
}
```

Result: `BUILD_RC=0`, 0 warnings; the jagged probe emits and prints **`7`**; the original #2857 repro still prints **`5`**; the **Issue2857 suite stays 15/15 green**. Tree restored by md5 round-trip afterwards.

So the fix is known-good — what is missing is the pin and the scope correction.

## Scope note that must be respected

`List[(T) -> void]` — arbitrary constructed generics — fails `GS0159` on **both** trees and is **explicitly and correctly out of scope**. Do not widen to constructed generics while fixing the array case; that is a different and much larger change. The reviewer verified that disclosure is accurate.

## Acceptance

- Jagged (`[][]delegate`) and, ideally, deeper ranks re-close correctly cross-assembly.
- Single-level delegate arrays and the original #2857 shape are unchanged — pin both as controls.
- A test pins the jagged shape by **value** (`7`), not by absence of a diagnostic. This repo has now seen **eight** instances of a test that names a semantic property while asserting a coarser observable constant across both sides of it.
- `List[(T) -> void]` still fails identically — assert it, so the boundary is stated by a test rather than by prose.

## Driver note

`gsi` has no `/r:` (confirmed from `--help`) and bare `gsc` hits a pre-existing `GS9999`/#988 wall cross-assembly, so for cross-assembly shapes **emit is not merely the oracle, it is the only live column**. Same-assembly shapes do give three live columns and should be used for the regression corpus.

DavidObando:
## Summary

- recursively unwrap jagged-array element types before delegate-dependent explicit generic re-closure
- retain constructed-generic boundary: `List[((T) -> void)]` still reports `GS0159`
- add value-based depth-2 (`7`) and depth-3 (`8`) regressions plus boundary coverage

Closes #3178.

## Live-main integration

#2903 squash-merged after this branch was created. Rebase dropped its obsolete original commits and replayed only #3178's issue commit. #3174's relocation remains intact: product change lives in `Binding/OverloadResolution/ClrOverloadResolution.cs` under dedicated overload-resolution namespace.

- base: `57256713f5131e46a8ca81759ae3c0eb5b2519b8`
- head: `8a3d953462099ab8c9f87340706f100dc8869a43`
- effective diff: 2 files, `+101/-3`
- `git merge-tree --write-tree --messages origin/main HEAD`: tree `5ff596ce85d3a118634151395bb6f02002526e16`, no conflict messages, rc 0

## Driver verification

Current file-mode bare `gsc` and `gsi` both use emitted execution; interactive REPL retains evaluator selection. Cross-assembly results on final head:

| Shape | `gsc /out:` | bare `gsc` | file `gsi` |
|---|---:|---:|---:|
| original #2857 | `5` | `5` | `5` |
| jagged #3178 | `7` | `7` | `7` |
| constructed generic | `GS0159` | `GS0159` | `GS0159` |

## Mutation verification

Clean focused control: 18/18.

| Mutant | Build | Result |
|---|---|---|
| restore one-level call-site unwrap | 0 warnings/errors | RED exactly 2/18: depth-2 and depth-3 jagged rows |
| invert helper loop condition | 0 warnings/errors | RED 12/18, including both jagged rows |

Applied through `git diff` plus source/Core DLL md5 movement. Exact restoration: source `4256019b...`, `GSharp.Core.dll` `dbe9c927...`.

## Validation

Release warnings-as-errors: 0 warnings, 0 errors. Fresh Core/Compiler/Repl `GSharp.Core.dll` epoch: `1785843381`.

All 9 projects ran unfiltered with `dotnet test GSharp.sln -c Release`:

| Project | Passed | Failed | Skipped |
|---|---:|---:|---:|
| Compiler.Tests | 4,263 | 0 | 0 |
| Core.Tests | 7,128 | 0 | 1 |
| Interpreter.Tests | 1,410 | 0 | 0 |
| LanguageServer.Tests | 462 | 0 | 0 |
| Sdk.Tests | 59 | 0 | 0 |
| Cs2Gs.Tests | 1,907 | 0 | 0 |
| GSharp.GeneratorHost.Tests | 31 | 0 | 0 |
| **Total** | **15,373** | **0** | **1** |

Final-head CI: 32/32 checks successful, including `build`, `e2e`, `ilverify`, `cs2gs-corpus`, and `cs2gs-oahu`.

Not run locally: e2e workflow; CI e2e passed.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/OverloadResolution/ClrOverloadResolution.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolution/ClrOverloadResolution.cs
@@ -2324,9 +2324,7 @@ internal static class ClrOverloadResolution
                         if (gmi.GetParameters().Any(static p =>
                                 p.ParameterType.ContainsGenericParameters
                                 && ClrTypeUtilities.IsDelegateType(
-                                    p.ParameterType.IsArray
-                                        ? p.ParameterType.GetElementType()
-                                        : p.ParameterType)))
+                                    UnwrapArrayElement(p.ParameterType))))
                         {
                             var recoveredSymbols = recoverTypeArgSymbols?.Invoke(gmi, false) ?? default;
                             if (TryCloseOverUserReferenceTypePlaceholders(
@@ -4440,6 +4438,16 @@ internal static class ClrOverloadResolution
         {
             return false;
         }
+    }
+
+    private static Type UnwrapArrayElement(Type type)
+    {
+        while (type is not null && type.IsArray)
+        {
+            type = type.GetElementType();
+        }
+
+        return type;
     }
 
     private static Type FindReferenceTypeConstraintPlaceholder(Type importedBase, Type genericParameter)
--- a/test/Compiler.Tests/Emit/Issue2857ExplicitGenericLambdaProjectReferenceEmitTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2857ExplicitGenericLambdaProjectReferenceEmitTests.cs
@@ -118,6 +118,56 @@ public sealed class Issue2857ExplicitGenericLambdaProjectReferenceEmitTests
     }
 
     [Theory]
+    [InlineData("i3178jagged2", 2, 7)]
+    [InlineData("i3178jagged3", 3, 8)]
+    public void ExplicitGenericTypeArgument_WithJaggedDelegateArrayAcrossReference_Runs(
+        string packageName,
+        int arrayDepth,
+        int expected)
+    {
+        var libraryArrayType = string.Concat(Enumerable.Repeat("[]", arrayDepth)) + "((T) -> void)";
+        var consumerDelegateType = "((Derived) -> void)";
+        var consumerArrayType = string.Concat(Enumerable.Repeat("[]", arrayDepth)) + consumerDelegateType;
+        var callback = $"(item Derived) -> {{ item.Value = {expected} }}";
+        for (var depth = 1; depth < arrayDepth; depth++)
+        {
+            var nestedType = string.Concat(Enumerable.Repeat("[]", depth)) + consumerDelegateType;
+            callback = $"{nestedType}{{ {callback} }}";
+        }
+
+        var indices = string.Concat(Enumerable.Repeat("[0]", arrayDepth));
+        var library = $$"""
+            package {{packageName}}
+
+            open class Base {
+                var Value int32
+
+                shared {
+                    func Make[T Base init()](configure {{libraryArrayType}}) T {
+                        let value = T()
+                        configure{{indices}}(value)
+                        return value
+                    }
+                }
+            }
+            """;
+        var consumer = $$"""
+            package {{packageName}}use
+            import System
+
+            class Derived : {{packageName}}.Base {}
+
+            func Main() {
+                let value = {{packageName}}.Base.Make[Derived](
+                    {{consumerArrayType}}{ {{callback}} })
+                Console.WriteLine(value.Value)
+            }
+            """;
+
+        Assert.Equal($"{expected}\n", CompileAndRun(library, consumer, packageName));
+    }
+
+    [Theory]
     [InlineData("i2857arity0", 0, 22)]
     [InlineData("i2857arity1", 1, 11)]
     [InlineData("i2857arity2", 2, 33)]
@@ -297,6 +347,46 @@ public sealed class Issue2857ExplicitGenericLambdaProjectReferenceEmitTests
             """;
 
         var output = CompileExpectingFailure(library, consumer, packageName);
+        var diagnosticIds = Regex.Matches(output, @"\berror (GS\d{4}):")
+            .Select(match => match.Groups[1].Value)
+            .ToArray();
+        Assert.Equal(new[] { "GS0159" }, diagnosticIds);
+    }
+
+    [Fact]
+    public void ExplicitGenericTypeArgument_WithDelegateInConstructedGenericAcrossReference_IsRejected()
+    {
+        const string PackageName = "i3178constructedgeneric";
+        var library = $$"""
+            package {{PackageName}}
+            import System.Collections.Generic
+
+            open class Base {
+                var Value int32
+
+                shared {
+                    func Make[T Base init()](configure List[((T) -> void)]) T {
+                        let value = T()
+                        configure[0](value)
+                        return value
+                    }
+                }
+            }
+            """;
+        var consumer = $$"""
+            package {{PackageName}}use
+            import System.Collections.Generic
+
+            class Derived : {{PackageName}}.Base {}
+
+            func Main() {
+                let configure = List[((Derived) -> void)]()
+                configure.Add((item Derived) -> { item.Value = 9 })
+                {{PackageName}}.Base.Make[Derived](configure)
+            }
+            """;
+
+        var output = CompileExpectingFailure(library, consumer, PackageName);
         var diagnosticIds = Regex.Matches(output, @"\berror (GS\d{4}):")
             .Select(match => match.Groups[1].Value)
             .ToArray();


DavidObando:
## Rubber-duck review — PR #3188 `fix/3178-work` @ `aca82282`

**Verdict: NOT MERGEABLE as-is — one mechanical BLOCKER (merge conflict), zero product BLOCKERs.**

The product change is exactly the remedy #3178 specified, it is correct, it is pinned by tests I proved can fail, and it survives rebase. But GitHub now reports **`mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`**, so it cannot merge in its present state. I built and validated the post-rebase state; after `gh pr update-branch` everything below holds unchanged.

Head verified by ancestry, not by the briefed SHA: `gh pr view` → `aca82282`, and `git merge-base --is-ancestor 6723a7f8 pr3188` → true.

---

## First: the `+589/−1` you will see in the UI is an artifact — the real diff is `+101/−3`

This matters because a reviewer reading `gh pr diff` would reasonably file scope creep, and I nearly did.

**#2903 squash-merged at `2026-08-04T05:32:21Z` — roughly two minutes before I started.** The PR branch still carries #2903's *original* commits (`e2b24112`, `664e90e6`, …), which are **not ancestors of `main`** because a squash rewrites them. So the PR's merge-base `0ffd90b1` predates its own dependency, and diffing against it double-counts #2903:

| diff base | delta | what it contains |
|---|---|---|
| merge-base `0ffd90b1` | **+589 / −1**, 3 files | #2903 **and** #3178 |
| **current `main` `81e6deab`** | **+101 / −3**, 2 files | **#3178 only** |

`MemberLookup.cs` (`GetNearestImportedBase`) is **already on `main`** — it is not part of this PR's effective delta at all. **Answering attack item 1 directly: the PR is not bigger than the measured remedy. Against current `main` the product change is `+14/−3` — the ternary replaced by `UnwrapArrayElement(...)` plus the ten-line helper, i.e. the six-line remedy verbatim.** No scope creep.

**Self-correction I owe:** my first grep for the one-level ternary on `main` returned nothing and I briefly concluded #2903's hunk was absent. The ternary is simply **wrapped across three lines**, so `grep "IsArray ? "` could not match it. What actually carried the conclusion was the **call-site count** — `TryCloseOverUserReferenceTypePlaceholders` has **4** sites on `0ffd90b1` and **5** on `81e6deab`. *A zero can be syntax, not absence* — again, and this time it was my own grep.

---

## BLOCKER (mechanical) — the PR conflicts with `main` and cannot merge

`git merge-tree --write-tree --messages gh/main pr3188` — reading the **output**, not the rc:

```
CONFLICT (content): Merge conflict in .../OverloadResolution/ClrOverloadResolution.cs
CONFLICT (add/add): Merge conflict in .../Issue2857ExplicitGenericLambdaProjectReferenceEmitTests.cs
```

When I began, `gh pr view` reported `mergeable=MERGEABLE, mergeStateStatus=CLEAN`. That was **stale** — GitHub had not yet recomputed after #2903 landed. It then moved to `UNKNOWN`, and now reports **`CONFLICTING` / `DIRTY`**, independently confirming the merge-tree output. `merge-tree` knew ~20 minutes before the API did.

Cause is the squash, not divergent intent: both sides added the same #2903 text from a base that has neither.

**I verified the resolution is lossless and mechanical.** Applying the true delta (`git diff gh/main pr3188`) onto current `main` applies cleanly (`git apply --check` rc=0) and yields exactly `+101/−3`:

- `REBASE_BUILD_RC=0`, **0 warnings**
- `Issue2857ExplicitGenericLambdaProjectReferenceEmitTests` → **18/18 passed**
- my probe corpus on that tree → identical values (below)

**Fix: `gh pr update-branch` (or rebase onto `81e6deab`), taking the branch side on both files.** Note this also means the green CI below ran against the pre-squash base and should be re-run after the update.

---

## The fix is correct — measured, both sides, all three drivers

Cross-assembly `Make[Derived]` with an *n*-deep jagged delegate array. `d1` is the single-level control, `lg` is the scope boundary.

| probe | parameter | `main` `81e6deab` | **PR head** | post-rebase |
|---|---|---|---|---|
| `d1` | `[]((T) -> void)` | `5` / `5` / `5` | `5` / `5` / `5` | `5` / `5` / `5` |
| `d2` | `[][]((T) -> void)` | **GS0159** ×3 | **`7` / `7` / `7`** | `7` / `7` / `7` |
| `d3` | `[][][]((T) -> void)` | **GS0159** ×3 | **`8` / `8` / `8`** | `8` / `8` / `8` |
| `d4` | `[][][][]((T) -> void)` | **GS0159** ×3 | **`6` / `6` / `6`** | `6` / `6` / `6` |
| `lg` | `List[((T) -> void)]` | GS0159 ×3 | **GS0159 ×3** | GS0159 ×3 |

Columns are `gsc /out:`+`dotnet exec` / bare `gsc` / `gsi`. Values, not exit codes. **`7` and `5` reproduce the issue's measured numbers exactly**, and rank 3 and rank 4 — beyond what the issue asked for — also work. **`List[(T) -> void]` is still refused on the fixed tree, so the scope boundary you asked me to police is intact** and is pinned by a test rather than by prose.

### Driver model — your warning was warranted, and the *issue* is the stale one

**ADR-0156 Phase 1 has landed**: `src/Core/CodeAnalysis/Execution/EmittedProgramHost.cs` exists and `src/Compiler/Program.cs:321` calls `EmittedProgramHost.Run(...)`. Bare `gsc` and `gsi` no longer interpret — they emit to memory and run the result.

Two consequences the issue's driver note gets wrong:

1. **`gsi` *does* have `/r:`** — `gsi --help` lists `/r:<file>  Reference an assembly (alias: /reference:<file>)`, plus `--engine {evaluator|emit}`. The issue says it does not, "confirmed from `--help`". That was true when written; it is false at this head. So cross-assembly is **not** emit-only any more — I measured all three columns live.
2. **The three-driver matrix is now non-discriminating for this bug class, by construction.** All three drivers execute the same emitted program, so for a *binding-time* defect they cannot disagree. The uniformity above is not evidence of correctness — the *base-vs-head* movement is. Worth the fleet knowing before someone reads a uniform post-ADR-0156 table as a passed check.

---

## The tests can fail — proved, and for the right reason

Copied the PR's test file alone onto **unfixed `main`**, rebuilt, ran:

- **`Failed: 2, Passed: 16, Total: 18`**, `TEST_RC=1`
- both failures: `consumer.gs(7,35,8,98): error GS0159: Cannot find function Make.` — **the exact defect #3178 describes**, not an incidental break
- restored by exact md5 round-trip (`962725b0…` → `5752ec20…` → `962725b0…`)

On the fixed head, the same 18 pass. Two-sided.

**One thing to be clear about:** `ExplicitGenericTypeArgument_WithDelegateInConstructedGenericAcrossReference_IsRejected` **passes on unfixed `main` too** — it is among the 16. That is correct and intended (it pins the `List[...]` boundary, not the fix), but it means **2 of the 3 new tests are load-bearing for the fix**, and nobody should count the third as coverage of it.

Harness quality is good: asserted-empty output directory, `IlVerifier.Verify` on **both** the library and the consumer, `AssertLoads`, exit-code assertion, and `Assert.Equal($"{expected}\n", …)` — a **value** pin, which is what the acceptance criteria demanded.

---

## Mutants — 3 PRODUCT mutants, all applied, all killed

Pristine `ClrOverloadResolution.cs` `4256019bdaf4ceda9ce4d84f2148c96e`, `GSharp.Core.dll` `7086ce5c6b0f8e5178e870554f015259`. Every mutant asserted applied by `git diff` **and** `GSharp.Core.dll` md5 movement before any test ran.

| mutant | mutation | Core.dll md5 | result | kills |
|---|---|---|---|---|
| **M1-LOOP** | `while (… && type.IsArray)` → `if (… && type.IsArray)` | `09d68aab` | **RED 2/18** | exactly the two jagged cases |
| **M2-COND** | `type.IsArray` → `!type.IsArray` (condition inversion) | `23b38ce3` | **RED 12/18** | jagged ×2, arity ×4, array ×2, … |
| **MCTRL** (control) | `IsDelegateType(` → `!IsDelegateType(` | `49e0eaac` | **RED 15/18** | broad — harness live |

**M1-LOOP is the strongest single piece of evidence here.** Reducing the `while` to an `if` *is* `main`'s one-level unwrap, so the mutant reconstructs the pre-fix semantics — and it kills **exactly the two new tests, 2/18**, matching the independent copy-tests-onto-`main` run (also 2/18) by a completely separate route. Two methods, same answer.

Restored: source md5 back to `4256019b…`, `GSharp.Core.dll` back to `7086ce5c…` **exactly**, `git status --porcelain` clean.

---

## Robustness of the helper — the questions you asked, answered

**Termination.** `GetElementType()` on an array strictly reduces a finite nesting depth, so the loop cannot spin; there is no `Type` for which `IsArray` stays true indefinitely. Confirmed empirically to rank 4.

**Null.** `UnwrapArrayElement` can only return `null` if handed `null`, and `ClrTypeUtilities.IsDelegateType` (`src/Core/CodeAnalysis/Symbols/ClrTypeUtilities.cs:365`) is null-safe — `for (var t = type; t != null; …)` falls straight through to `false`. No NRE path.

**Completeness, not enumeration.** `TryCloseOverUserReferenceTypePlaceholders` has five call sites; the other four are failure-path `else if (!TryClose…)` with **no delegate guard at all** — they always attempt re-closure. Only the success-path site at `:2326` is guarded, so `:2326` is the only place an unwrap could be needed. The fix is complete for its mechanism rather than one more enumerated shape.

**No over-correction.** The guard only widens from "1-deep array of delegate" to "*n*-deep array of delegate". Every newly-admitted shape previously produced `GS0159`, so there is no working behaviour to displace, and on failure `TryClose…` leaves `MakeGenericMethod`'s result intact. The pre-existing over-correction control `ExplicitGenericTypeArgument_WithUnrelatedDelegateParameterAcrossReference_Runs` stays green.

**A hypothesis of mine that died.** I probed `[][]T` — a jagged array of the *non-delegate* generic parameter — expecting the same one-level bug one guard over. It fails `GS0159` identically on both trees. I then probed single-level `[]T`, and **that fails on both trees too**, so it is not a jaggedness bug at all but a separate pre-existing gap in a different mechanism. Not a regression, not in #3178's scope, and my "incomplete fix" theory is dead. Recording it because a negative result here is what rules out the incompleteness charge.

---

## Suites — denominator 9, source `dotnet sln GSharp.sln list`

**8 of 9 green, 11,093 passed, 0 failed, 1 skipped:**

|---|---|---|---|---|
| Core.Tests | 0 | 0 | 7128 | 1 |
| Interpreter.Tests | 0 | 0 | 1393 | 0 |
| InternalAnalyzers.Tests | 0 | 0 | 9 | 0 |
| **Compiler.Tests** | — | — | — | **still running (~1h18m)** |

`Compiler.Tests` unfiltered is not finished at posting time — that is a **gap in evidence, not a pass**, and I will post the number when it lands. Its `Issue2857…` class is independently 18/18 on head and on the post-rebase tree.

CI on the PR: **30 checks green**, including the required **`ilverify`** gate (present, passing, 3m57s) — no `update-branch` needed for gate *presence*, though the rebase will re-run everything against the post-squash base.

---

## NIT

The tests exercise only the `gsc /out:` driver. Ordinarily I would ask for the other two, but post-ADR-0156 all three run the same emitted program, so the added columns would be the same computation and near-zero value. Recording the reasoning so the omission is a decision rather than an oversight.

---

## Disclosure

My depth-1 control initially reported `GS0155` on all three drivers, which looked exactly like a finding. It was my own generator: **BSD `seq 1 0` prints `1 0` — two iterations — where GNU `seq` prints nothing**, so the "single-level" fixture was emitted with two spurious nesting levels. Ascending ranges (`d2`–`d4`) were unaffected, so those rows stand. Same family as `find -newermt '@epoch'`: a macOS coreutils divergence that fails *productively* instead of erroring, and the wrong answer looks like data. *A failing probe is not evidence until it fails for the reason you think.*

DavidObando:
### Addendum — coverage gap closed: **9 of 9, 15,356 tests, 0 failures, 1 skipped**

`Compiler.Tests` was still running when I posted and I labelled it a gap rather than implying coverage. It has finished:

|---|---|---|---|---|
| **Compiler.Tests** | **0** | **0** | **4263** | **0** |
| Core.Tests | 0 | 0 | 7128 | 1 |
| Interpreter.Tests | 0 | 0 | 1393 | 0 |
| InternalAnalyzers.Tests | 0 | 0 | 9 | 0 |

**Total 15,356 / 0 failures / 1 skipped**, all 9 projects, unfiltered, on PR head `aca82282`. Denominator and its source: `dotnet sln GSharp.sln list` → 9 `*.Tests.csproj`. `Compiler.Tests` took 40m38s here versus ~1h18m on my last review — the box was less contended, not a smaller run (4,263 vs 4,255 tests, the delta being this PR's two new `[Theory]` cases plus intervening `main` commits).

This closes the only outstanding evidence gap. **Verdict unchanged: NOT MERGEABLE as-is, zero product BLOCKERs, one mechanical merge conflict.** After `gh pr update-branch` the fix is correct, pinned, and mutation-verified — and I have already built and measured that exact post-rebase state (`REBASE_BUILD_RC=0`, 0 warnings, 18/18, values `5`/`7`/`8`/`6`, `List[...]` boundary intact).

Worth restating because it is the one thing a green 15,356-test run cannot tell you: **this suite was green on `main` too, while `[][]((T) -> void)` was broken.** The evidence that the fix works is the base-vs-head movement and the M1-LOOP mutant, not the size of the passing total.

DavidObando:
## Rebase complete

Rebased onto current `origin/main` `917677ec`; new head `cedd55ec`.

### Conflict resolution

#2903 had squash-merged while this branch still carried its original commits. I dropped those already-upstream commits and replayed only #3178's commit. #3174's relocation was preserved: `UnwrapArrayElement` now lives in `src/Core/CodeAnalysis/Binding/OverloadResolution/ClrOverloadResolution.cs` under `GSharp.Core.CodeAnalysis.Binding.OverloadResolution`. Test resolution retained main's #2857 file plus only #3178's three cases.

`git merge-tree --write-tree --messages origin/main HEAD` printed only tree `d2385c6c09ce3302ec1ee037ac4c262233f4406b`; no conflict messages (`rc=0`). Effective diff remains 2 files, `+101/-3`.

### Re-verification

- Release warnings-as-errors build: `BUILD_RC=0`, 0 warnings, 0 errors.
- Fresh `GSharp.Core.dll` witnesses in Core, Compiler, and Repl: epoch `1785835199`.
- Current driver model confirmed from code and execution: bare `gsc`, `gsc /out:`, and file-mode `gsi` all use emitted execution; only interactive REPL can still select evaluator mode.
- Original cross-assembly repro: `5 / 5 / 5`.
- Jagged cross-assembly repro: `7 / 7 / 7`.
- Constructed-generic boundary: `GS0159 / GS0159 / GS0159`.
- Focused clean control: 18/18.

### Product mutants

Both mutants applied with `git diff`, source-md5 movement, build-green product output, and `GSharp.Core.dll` md5 movement.

1. Call site restored to one-level array unwrap: build 0/0; **RED 2/18**, exactly `i3178jagged2` and `i3178jagged3`.
2. Helper condition inverted (`IsArray` → `!IsArray`): build 0/0; **RED 12/18**, including both jagged rows.

Clean source restored exactly to `4256019bdaf4ceda9ce4d84f2148c96e`; clean final Core DLL restored exactly to `2d62b7497bf6d9a924ea7bebcd2a0d15`. First clean restore build after mutant 2 exited 139 during restore before producing a result; rerun completed 0/0 and exact md5 restoration. Mutant score uses only build-green runs above.

### Nine unfiltered projects

| Project | Passed | Failed | Skipped |
|---|---:|---:|---:|
| Compiler.Tests | 4,263 | 0 | 0 |
| Core.Tests | 7,128 | 0 | 1 |
| Interpreter.Tests | 1,394 | 0 | 0 |
| LanguageServer.Tests | 462 | 0 | 0 |
| Sdk.Tests | 59 | 0 | 0 |
| Cs2Gs.Tests | 1,907 | 0 | 0 |
| GSharp.GeneratorHost.Tests | 31 | 0 | 0 |
| **Total** | **15,357** | **0** | **1** |

Duck baseline restored; current main's #3189 adds one Interpreter test, hence 15,357 rather than 15,356. Worktree clean before push. CI now rerunning on rebased head.

DavidObando:
## Live-main refresh (supersedes prior comment's head/base)

#3191 merged while first rebased CI was running. Rebased again onto current `main` `074f28a7`; final pushed head is `950588d1`.

Conflict resolution remains mechanical: dropped #2903's already-squash-merged original commits, preserved #3174's dedicated overload-resolution namespace/file layout, replayed only #3178's issue commit. `git merge-tree --write-tree --messages origin/main HEAD` printed only tree `b76536530a607d0bdd7decb7b9756a160154434e`, no conflict messages (`rc=0`).

Reverified on this exact head:

- Release warnings-as-errors: 0 warnings, 0 errors.
- Core/Compiler/Repl `GSharp.Core.dll` freshness epoch: `1785839655`.
- Original repro: `5 / 5 / 5` across emitted `/out:`, bare `gsc`, and file-mode `gsi`.
- Jagged repro: `7 / 7 / 7`.
- Constructed-generic boundary: `GS0159 / GS0159 / GS0159`.
- Focused clean control: 18/18.
- One-level call-site mutant: build-green, Core md5 `aa9dd1a1...`, RED exactly 2/18 (`i3178jagged2`, `i3178jagged3`).
- Inverted-loop mutant: build-green, Core md5 `566f9967...`, RED 12/18 including both jagged rows.
- Exact restoration: source md5 `4256019b...`; Core md5 `f022caac...`.

`dotnet test GSharp.sln -c Release` ran all 9 test projects unfiltered:

| Project | Passed | Failed | Skipped |
|---|---:|---:|---:|
| Compiler.Tests | 4,263 | 0 | 0 |
| Core.Tests | 7,128 | 0 | 1 |
| Interpreter.Tests | 1,399 | 0 | 0 |
| LanguageServer.Tests | 462 | 0 | 0 |
| Sdk.Tests | 59 | 0 | 0 |
| Cs2Gs.Tests | 1,907 | 0 | 0 |
| GSharp.GeneratorHost.Tests | 31 | 0 | 0 |
| **Total** | **15,362** | **0** | **1** |

Duck baseline 15,356 restored and exceeded only by six tests added on current main (#3189 + #3191). Worktree clean; CI rerunning on `950588d1`.

DavidObando:
## Round 3 (rebase confirmation) — **MERGEABLE**

Narrow question only: *did the rebase carry the product change across intact — nothing dropped, nothing extra?* **Yes, byte-for-byte.** No new findings; my round-2 BLOCKER (the #3174 namespace conflict) is resolved.

`aca82282` exists but is **not** an ancestor of `cedd55ec` — expected after a rebase, not a finding.

### The stat shrink is the parent landing, not a drop

`aca82282` was a **stack**: it carried #2857's fix (`e2b24112`) plus its 563-line pin file *and* the #3178 commit — `+589/-1`. The new head is a single commit, `+101/-3`. That shrink is the thing that would normally be the finding, so I proved it is not:

- #2857's product one-liner — `clr = TypeMemberModel.GetNearestImportedBase(userClass)?.ClrType ?? typeof(object);` — is **present in `main` at the rebase base `917677ec`** (1 occurrence).
- `Issue2857ExplicitGenericLambdaProjectReferenceEmitTests.cs` is **in `main`** with 6 `[Fact]`/`[Theory]` declarations.

The parent merged; the child now carries only its own delta. Correct.

### The #3178 delta itself is byte-identical across the rebase

Isolated the old head's #3178-only commit (`6723a7f8`, `+101/-3`) and compared its payload to `917677ec..cedd55ec`:

| delta | old md5 | new md5 | `diff` |
|---|---|---|---|
| `src/` payload (14 lines) | `1fc03e403acb6b2f4821cb4af1df7d49` | `1fc03e403acb6b2f4821cb4af1df7d49` | **empty** |
| `test/` payload (90 lines) | `6188ebe00fe82a449e049d0a0e5a57d8` | `6188ebe00fe82a449e049d0a0e5a57d8` | **empty** |

Nothing outside those **2 files**. Nothing dropped, nothing added. The only change is the path: `Binding/OverloadResolution.cs` → `Binding/OverloadResolution/ClrOverloadResolution.cs`. **Old path not resurrected** — `ls-tree` on the head shows only the new sub-namespace files.

Placement is right too: the hunk sits at `ClrOverloadResolution.cs:2324`, inside `try { closed = gmi.MakeGenericMethod(...); if (...) }` — **after `MakeGenericMethod` succeeds**, the live equivalent of the old `:2321+` re-closure site.

### Head commit title is not new work

`cedd55ec` "Handle jagged delegate arrays in generic closure" is the **same title as the original `6723a7f8`**, and the payload md5s above are identical. No scope creep entered during the conflict resolution.

### Repro, before/after, emit as oracle

Reverted **only** the call site back to the one-level ternary, leaving the pins untouched. `Core.dll` moved within-tree `8e13d6cd` → `9dec7dc6`.

| shape | ternary (unfixed) | `UnwrapArrayElement` (head) |
|---|---|---|
| `[]((T) -> void)` — original #2857 repro | passes | passes |
| **`[][]((T) -> void)`** | **`error GS0159: Cannot find function Make.`** | prints **7** |
| **`[][][]((T) -> void)`** | **`GS0159`** | prints **8** |
| `List[((T) -> void)]` — out of scope | `GS0159` | `GS0159` (unchanged) |

`18` passed on head; on the unfixed tree **exactly 2 fail — both jagged rows** — and the other 16 survive. Clean kill isolation: the hunk is load-bearing for jagged and for nothing else. Restored by exact md5 round-trip: source back to `4256019b…`, `Core.dll` back to `8e13d6cd…` **identically**.

Loop is sound: each iteration strips one rank, array nesting is well-founded, so it terminates; non-array input skips the body and returns unchanged, making it a strict superset of the old ternary whose *only* newly-admitted inputs are rank≥2 delegate arrays. `GetElementType()` is called only under `IsArray`, so the `null` guard cannot be hit.

**The constructed-generic case is pinned as still-refused** (`…WithDelegateInConstructedGenericAcrossReference_IsRejected`, asserting `GS0159`). That is better than staying silent on the out-of-scope boundary — it locks it.

### #3191

Zero file overlap. #3191 touches `Evaluator.Expressions.cs` (evaluator reification); this touches the **binder**. Same territory by name, different subsystem. `git merge-tree --write-tree origin/main cedd55ec` **output** is one line — the tree oid, no conflict markers.

### Suites

Ran: **Compiler.Tests** filtered to the pin class (18/18, RC=0) and **Core.Tests** (7128 passed, 1 skipped, RC=0) — the latter because `OverloadResolutionTests` / `OverloadResolutionPropertyTests` / `OverloadResolutionGenerators` live there and this is that subsystem. **Not run: the other 7 suites** — CI is 32/32 with `ilverify`, and I paid the full 9-of-9 on round 2. Build Release `-warnaserror`, 0 warnings, 38 project outputs; `UnwrapArrayElement` confirmed present in the built `GSharp.Core.dll` rather than inferred from timestamps.

Porcelain clean apart from the pre-existing untracked `test-assets/Issue3119.CrossConstants/packages.lock.json` (#3170), which is not mine.

DavidObando:
## Second live-main refresh — driver-reference change included

#3147 merged after prior CI and directly changed compiler/REPL reference handling, so I rebased again rather than treating it as unrelated.

- Base: `57256713`
- Head: `8a3d9534`
- Merge-tree: only tree `5ff596ce85d3a118634151395bb6f02002526e16`, no conflict messages, `rc=0`
- Release warnings-as-errors: 0 warnings, 0 errors
- Fresh Core/Compiler/Repl `GSharp.Core.dll`: epoch `1785843381`

Re-measured all driver columns after #3147:

- original cross-assembly repro: `5 / 5 / 5`
- jagged cross-assembly repro: `7 / 7 / 7`
- constructed-generic boundary: `GS0159 / GS0159 / GS0159`
- focused clean control: 18/18

Product mutants also rerun on this exact tree:

- one-level call-site mutant: build-green, Core md5 `6f19bfef...`, RED exactly 2/18 (both jagged rows)
- inverted-loop mutant: build-green, Core md5 `d4f23c24...`, RED 12/18 including both jagged rows
- exact clean restoration: source `4256019b...`, Core `dbe9c927...`

All 9 projects, unfiltered via `dotnet test GSharp.sln -c Release`:

| Project | Passed | Failed | Skipped |
|---|---:|---:|---:|
| Compiler.Tests | 4,263 | 0 | 0 |
| Core.Tests | 7,128 | 0 | 1 |
| Interpreter.Tests | 1,410 | 0 | 0 |
| LanguageServer.Tests | 462 | 0 | 0 |
| Sdk.Tests | 59 | 0 | 0 |
| Cs2Gs.Tests | 1,907 | 0 | 0 |
| GSharp.GeneratorHost.Tests | 31 | 0 | 0 |
| **Total** | **15,373** | **0** | **1** |

Worktree clean. CI rerunning on `8a3d9534`.

PR #3189, opened 2026-08-04
Pin interpreted char span slice offsets

### LINKED ISSUE #3179: [test] char-span start offset is unpinned: mutating this.start to 0 survives 14/14 while changing observable output
Follow-up from review of #3128 (issue #3114). The char branch added at `src/Core/CodeAnalysis/Evaluator.Reflection.cs:265-268` is **correct**, but its `start` argument is **not pinned by any test**.

## Measurement

PRODUCT mutant at `Evaluator.Reflection.cs:268`:

```csharp
new string(chars, this.start, Length)   ->   new string(chars, 0, Length)
```

Applied assertion: pre=1, post_orig=0, post_mut=1. `M2_BUILD_RC=0`.

**Result: SURVIVES 14/14.**

## It is a live survivor, not an equivalent mutant

The reviewer verified the mutant changes observable behaviour before scoring it — which is the step that separates a real coverage hole from an equivalent-mutant false alarm:

```
sliced char span, Slice(1, 3) of [a,b,c,d,e]
   pristine : rc0  "bcd"
   mutant   : rc0  "abc"
```

So a wrong `start` produces a wrong answer silently, rc0, no diagnostic, and the whole suite stays green.

## Why the length axis is already safe but start is not

The sibling axis **is** closed: mutating `{Length}` to `{this.array.Length}` in the non-char branch is **killed 2/14** by `SpanConversionMutationAndSlicing_AgreeAcrossDrivers`, because `:110` asserts `window.ToString()` where `window = readOnly.Slice(1).Slice(0, 2)` over a 3-element array — the mutant reads `[3]`, the truth is `[2]`. Nothing plays the equivalent role for the char branch's offset.

## Fix

Add a char-span assertion whose value discriminates `start`. `bcd` is the load-bearing value the existing probes already use informally: for `Slice(1, 3)` over `[a,b,c,d,e]`, a wrong start reads `abc` and a wrong length reads `bcde`, so a single expected `bcd` pins **both** axes at once. Assert the printed value across all three drivers with **emit as oracle**.

## Guard rail — do not trade a loud refusal for a wrong answer

#3128's first attempt converted a loud `GS0511` refusal into a wrong answer, and the accepted fix explicitly preserved the refusal path. That behaviour must stay:

```
var text = "hello"
var s = text.AsSpan()      ->  bare gsc rc=1 GS0511,  gsi rc=1 GS0511
```

A CLR-produced span is still refused loudly on both interpreting drivers. **Assert this as a control** in whatever change closes the gap — it is the exact regression this area already produced once.

## Acceptance

- The `this.start` -> `0` mutant is **killed**, and the kill is demonstrated (state the failing test names and count).
- The `{Length}` mutant stays killed.
- The `GS0511` refusal for CLR-produced spans is unchanged and asserted.
- No production change should be needed here — the code is right, the test is missing. If a production change appears necessary, that is a finding worth reporting rather than making.

DavidObando:
## Summary

- add a `ReadOnlySpan[char].Slice(1, 3)` value assertion over `abcde`, with emitted `bcd` as oracle
- compare remaining tree-walk evaluator directly against that oracle, so `InterpretedSpanValue.ToString()` stays covered after ADR-0156 moved file drivers to emission
- preserve CLR-produced-span `GS0511` boundary on remaining interactive evaluator path
- test-only change; no product code changed

Fixes #3179

## Current trigger correction

PR #3182 merged while this work was running. Bare `gsc` and `gsi <file>` now execute emitted code, so they no longer reach `Evaluator.Reflection.cs`:

| Shape | bare `gsc` | `gsc /out:` + run | `gsi` |
|---|---|---|---|
| sliced char span | rc0, `bcd` + `Success.` | rc0, `bcd` | rc0, `bcd` |
| `string.AsSpan()` inside `Main` | rc0 | rc0 | rc0 |

Therefore brief's file-driver `GS0511` control became stale on current `main`. Live refusal remains interactive `SessionEngine` evaluator: new control reaches it through `Main`, asserts exactly one `GS0511`, and rejects `GS9999`. Same variables at top level are rejected earlier with `GS0219`, so `Main` is required to measure evaluator boundary.

Three file-driver rows still assert printed `bcd` value with emit as oracle. Direct `SessionEngine` comparison in same test pins product branch issue #3179 targets.

## Product mutation proof

Measured on final head `9f8e2c6b` over `main` `81e6deab`. Both mutants applied exactly once (`pre original=1/mutant=0`, `post original=0/mutant=1`), produced changed md5, built with `BUILD_RC=0`, and restored to exact pristine md5 `ea53ed84bf4f459c78cf4a3ab5d05730`.

| Mutant | Current main tests | This head |
|---|---:|---:|
| `new string(chars, this.start, Length)` → `new string(chars, 0, Length)` | survives 15/15 | killed: 3 failures / 16, all `SpanConversionMutationAndSlicing_AgreeAcrossDrivers` rows; `bcd` became `abc` |
| `[{Length}]` → `[{this.array.Length}]` | survives 15/15 | killed: 3 failures / 16, same rows; sliced non-char display `[2]` became `[3]` |

Second row differs from issue text: before #3182 it was killed by two file-evaluator rows; after file drivers switched to emission, it became a live survivor too. This patch closes both reopened evaluator axes.

## Validation

Nine-project denominator comes from `dotnet sln GSharp.sln list`, matching `build/generate-ci-test-matrix.py:26`.

Unfiltered run on commit `b66a584b` (same patch, before final rebase):

| Project | Passed | Skipped | Failed |
|---|---:|---:|---:|
| GSharp.GeneratorHost.Tests | 31 | 0 | 0 |
| Core.Tests | 7,128 | 1 | 0 |
| Compiler.Tests | 4,245 | 0 | 0 |
| Interpreter.Tests | 1,341 | 0 | 0 |
| LanguageServer.Tests | 462 | 0 | 0 |
| Sdk.Tests | 59 | 0 | 0 |
| Cs2Gs.Tests | 1,907 | 0 | 0 |
| **Total** | **15,286** | **1** | **0** |

Final rebased head `9f8e2c6b` over live `main` `81e6deab`:

- solution build: `BUILD_RC=0`, 0 warnings, 0 errors
- `GSharp.Core.dll` freshness: mtime after build start
- targeted issue tests: 16 passed, 0 failed
- merge-tree output: `f9812c0947a0e5b449156b7392b1be47fae8a924`, equal to head tree
- final two-sided mutations: results above

Not rerun after final rebase: all nine unfiltered projects. Final rebase added three already-merged main commits; solution build, targeted tests, and both product mutations were rerun. Local e2e, Oahu corpus, and standalone ILVerify were not run; required CI gates remain authoritative.

### DIFF
--- a/test/Interpreter.Tests/Issue3114ReadOnlySpanInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue3114ReadOnlySpanInterpreterTests.cs
@@ -99,9 +99,10 @@ public class Issue3114ReadOnlySpanInterpreterTests
                 writable[1] = 44
                 var tail = readOnly.Slice(1)
                 var window = tail.Slice(0, 2)
-                var letters = []char{'a', 'b', 'c'}
+                var letters = []char{'a', 'b', 'c', 'd', 'e'}
                 var writableChars Span[char] = letters
                 var readOnlyChars ReadOnlySpan[char] = writableChars
+                var charWindow = readOnlyChars.Slice(1, 3)
                 Console.WriteLine(readOnly.Length)
                 Console.WriteLine(readOnly[1])
                 Console.WriteLine(window[1])
@@ -110,6 +111,7 @@ public class Issue3114ReadOnlySpanInterpreterTests
                 Console.WriteLine(window.ToString())
                 Console.WriteLine(writableChars.ToString())
                 Console.WriteLine(readOnlyChars.ToString())
+                Console.WriteLine(charWindow.ToString())
             }
             """;
 
@@ -121,11 +123,21 @@ public class Issue3114ReadOnlySpanInterpreterTests
 
         try
         {
+            var emitted = CompileAndRun(root, "span-operations.gs", sourcePath);
+            Assert.Equal(
+                "3\n44\n33\nSystem.Span<Int32>[3]\nSystem.ReadOnlySpan<Int32>[3]\nSystem.ReadOnlySpan<Int32>[2]\nabcde\nabcde\nbcd\n",
+                emitted.StandardOutput);
+
+            var evaluated = new GSharp.Repl.Engine.SessionEngine { CaptureConsole = true, RunEntryPoint = true }
+                .Evaluate(Source);
+            Assert.False(evaluated.HasError);
+            Assert.Equal(emitted.StandardOutput, evaluated.Output);
+
             var result = driver switch
             {
                 Driver.CompilerEvaluation => CaptureConsole(
                     () => GSharp.Compiler.Program.Main([sourcePath])),
-                Driver.CompilerEmission => CompileAndRun(root, "span-operations.gs", sourcePath),
+                Driver.CompilerEmission => emitted,
                 Driver.Interpreter => CaptureConsole(
                     () => GSharp.Repl.Program.Main([sourcePath])),
                 _ => throw new InvalidOperationException($"Unexpected driver {driver}."),
@@ -134,8 +146,8 @@ public class Issue3114ReadOnlySpanInterpreterTests
             Assert.Equal(0, result.ExitCode);
             Assert.Equal(
                 driver == Driver.CompilerEvaluation
-                    ? "3\n44\n33\nSystem.Span<Int32>[3]\nSystem.ReadOnlySpan<Int32>[3]\nSystem.ReadOnlySpan<Int32>[2]\nabc\nabc\nSuccess.\n"
-                    : "3\n44\n33\nSystem.Span<Int32>[3]\nSystem.ReadOnlySpan<Int32>[3]\nSystem.ReadOnlySpan<Int32>[2]\nabc\nabc\n",
+                    ? emitted.StandardOutput + "Success.\n"
+                    : emitted.StandardOutput,
                 result.StandardOutput);
             Assert.Equal(string.Empty, result.StandardError);
         }
@@ -143,6 +155,26 @@ public class Issue3114ReadOnlySpanInterpreterTests
         {
             Directory.Delete(root, recursive: true);
         }
+    }
+
+    [Fact]
+    public void ClrProducedSpan_RemainsLoudlyRefusedInteractively()
+    {
+        const string Source = """
+            import System
+
+            func Main() {
+                var text = "hello"
+                var span = text.AsSpan()
+            }
+            """;
+
+        var cell = new GSharp.Repl.Engine.SessionEngine { RunEntryPoint = true }.Evaluate(Source);
+
+        var diagnostic = Assert.Single(cell.Diagnostics);
+        Assert.True(cell.HasError);
+        Assert.Equal("GS0511", diagnostic.Id);
+        Assert.DoesNotContain("GS9999", diagnostic.Message);
     }
 
     /// <summary>


DavidObando:
## Rubber-duck review — PR #3189 (`fix/3179-work`), issue #3179

**Verdict: MERGEABLE.** 0 BLOCKER · 0 SHOULD-FIX · 3 NIT.

Head verified by **ancestry, not existence**: `9f8e2c6b` is `main` `81e6deab` **+ exactly one commit, 0 behind**. `git merge-tree` *output* (not rc) is a single tree line `f9812c09` with zero conflict text, so the head already *is* the merged state. `BUILD_RC=0`, **0 warnings, 0 errors** on both trees.

### It is genuinely test-only — proven two ways

`git diff --name-only 81e6deab 9f8e2c6b -- src/` → **0 files**. And independently, `diff -q` on `Evaluator.Reflection.cs` between the two worktrees → **identical**. So #3179's "no production change should be needed" is satisfied by construction, and the `GS0511` refusal path cannot have been weakened — it is byte-for-byte the same code.

---

## The central experiment: one mutation, two test files, one binary

The right way to grade a test-only PR is to hold the product constant and vary only the tests. I applied the issue's PRODUCT mutant **once** on the PR tree — `Evaluator.Reflection.cs:267`, `new string(chars, this.start, Length)` → `new string(chars, 0, Length)`, applied assertion via `git diff` showing exactly 1 changed line — then ran **both** test-file variants against that **same compiled binary**. No build-difference confound.

| test file (same mutated `GSharp.Core.dll`) | result |
|---|---|
| **base** `81e6deab` version | **15 / 15 pass — mutant SURVIVES** |
| **PR** `9f8e2c6b` version | **3 / 16 fail — mutant KILLED** |

Failing tests: `SpanConversionMutationAndSlicing_AgreeAcrossDrivers` for `CompilerEvaluation`, `CompilerEmission`, `Interpreter`.

And it fails for the *right reason* — a failing probe is not evidence until it fails for the reason you think:

```
Assert.Equal() Failure: Strings differ            ↓ (pos 102)
Expected: ···"m.ReadOnlySpan<Int32>[2]\nabcde\nabcde\nbcd\n"
Actual:   ···"m.ReadOnlySpan<Int32>[2]\nabcde\nabcde\nabc\n"
at ...Issue3114ReadOnlySpanInterpreterTests.cs:line 134
```

Line 134 is `Assert.Equal(emitted.StandardOutput, evaluated.Output)`. **Acceptance criterion 1 met and demonstrated with names and count.**

### Criterion 2 — the `{Length}` pin still holds

`{Length}` → `{this.array.Length}` at `:268`: **killed 3/16**, same three cases. Not weakened, not duplicated away.

### Criterion 3 — the `GS0511` guard rail is not decorative

I did not take the new `[Fact]` on trust. Targeted mutant at `Evaluator.Reflection.cs:144`, `FindByRefLikeType(method.ReturnType)` → `FindByRefLikeType(method.DeclaringType)` (redundant with the check already made at `:135`, so ByRefLike return types stop being detected):

| test file (same mutated binary) | result |
|---|---|
| base | **15 / 15 pass — survives** |
| PR | **1 / 16 fail** — exactly `ClrProducedSpan_RemainsLoudlyRefusedInteractively` |

```
Expected: "GS0511"
Actual:   "GS9999"
```

That is precisely the #3128 regression the issue warns about — the loud, specific refusal degrading into the generic catch-all — and **`main` was completely blind to it**. The issue only asked for this to be *asserted as a control*; the PR turns it into a real pin. That is a second closed hole, not just the one requested.

### Did they close the class, or one cell?

The sibling offset axis is `CheckedIndex` at `:277`, `return this.start + index;`. Mutating it to `return index;`: **killed 3/16**, failing `33` → `44`. So the neighbouring offset axis was *already* pinned by the existing int-array assertions. No further gap in this family — the `start` cell was the only hole, and it is now closed.

---

## The driver model on this head — re-established by measurement, not by reading

ADR-0156 (#3177/#3186) landed, so I re-derived the model empirically before tabulating anything. With the `this.start` → `0` mutant applied, running the same `.gs` file through each driver:

| driver | pristine | evaluator mutant applied |
|---|---|---|
| bare `gsc <file>` | `bcd` | **`bcd` — unchanged** |
| `gsc /out:` + run | `bcd` | **`bcd` — unchanged** |
| `gsi <file>` | `bcd` | **`bcd` — unchanged** |
| in-process `SessionEngine` (interactive) | `bcd` | **`abc` — changed** |

**All three file drivers now emit; none of them touches `Evaluator.Reflection.cs` at all.** A three-driver table would have been confident fiction for this issue. `gsi` also refuses piped stdin (`gsi requires an interactive terminal`), so the in-process `SessionEngine` API is the *only* practical way to exercise the evaluator.

This makes the author's central design choice correct: the pin had to go through `SessionEngine`, and it does. The literal expectation is anchored on `emitted` first, so this is not a cross-driver tautology where both sides could drift together — emit is still the oracle, and the interpreter is checked against it.

## ADR-0154 / assertion-shape checks

- **Assertions are values, not diagnostics or rc.** The anchor is a full literal `"3\n44\n33\n…\nabcde\nabcde\nbcd\n"`.
- **Name matches claim.** `ClrProducedSpan_RemainsLoudlyRefusedInteractively` asserts `Assert.Single(cell.Diagnostics)`, `HasError`, `Id == "GS0511"`, and `DoesNotContain("GS9999")` — exactly what the name states, and M3 proves each clause is reachable.
- **Not specimen inflation.** The array grows `abc` → `abcde` (3 → 5) because `Slice(1, 3)` is impossible on 3 elements; 5 is the minimum that makes `bcd` discriminate **both** axes at once (wrong start → `abc`, wrong length → `bcde`). Census unchanged in spirit; the change is assertion *strength*. This is the smaller fix, not the bigger one.

## Test results

Denominator **9**, from `dotnet sln GSharp.sln list` (CI's own source per `build/generate-ci-test-matrix.py:26`).

| project | result |
|---|---|
| `Interpreter.Tests` (full, unfiltered) | **1394 / 0 failed** |
| `Core.Tests` (full, unfiltered) | **7128 / 0 failed / 1 skipped** (3 m 16 s) |

**7 of 9 NOT RUN locally, stated plainly rather than implied:** `Compiler.Tests`, `Cs2Gs.Tests`, `Sdk.Tests`, `LanguageServer.Tests`, `Extensions.Tests`, `InternalAnalyzers.Tests`, `GSharp.GeneratorHost.Tests`. Justification: the product is **proven byte-identical**, and the only changed file compiles into `Interpreter.Tests`, so no other project's behaviour can move. CI ran every shard on this head and all pass, including the required **`ilverify` gate (#3164) — pass, 3m59s** — plus `cs2gs-oahu` (11m10s) and `e2e`. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

---

## NITs

1. **Tautological assertion for one driver.** `:136-151` — for `Driver.CompilerEmission`, `result` *is* `emitted`, so `Assert.Equal(emitted.StandardOutput, result.StandardOutput)` reduces to `x == x` and can never fail. It is **not** a coverage loss, because the literal assertion at `:127-129` already anchors `emitted`; the line is simply dead for that case. `Assert.Equal(0, result.ExitCode)` stays meaningful (it checks the emitted process exit code, which nothing else asserts).
2. **The new block runs 3× redundantly.** `emitted` and `evaluated` are computed unconditionally inside a `[Theory]` over 3 drivers, so every run does 3 emit compiles plus 3 `SessionEngine` evaluations. Consequently the "3 killed" above is really *one* assertion × 3 parameterisations — the honest new-coverage count is **one** assertion. It also means a failure at `:134` masks the per-driver assertions below it. Hoisting the anchor + interactive check into a separate `[Fact]` would make the kill count read truthfully and cut the work by two-thirds.
3. **Issue anchor drift.** #3179 says the mutant "SURVIVES 14/14"; the actual base count on `81e6deab` is **15** (PR: 16). I reproduced the survival, so the finding stands — only the denominator moved between filing and now.

## Self-disclosed instrument error

My first `CheckedIndex` mutant targeted line **278**, which is a closing brace — `perl` silently changed nothing and the suite returned a clean **16/16**. Read naively that is a survival, and I would have filed a false coverage gap against `CheckedIndex`. The **empty `git diff` applied-assertion caught it**; re-running at the correct line `:277` killed 3/16. *A no-result is not a survival* — this is exactly why the applied-assertion step is not ceremony.

Also observed and checked: a build leaves an untracked `test-assets/Issue3119.CrossConstants/packages.lock.json`. It appears identically on **base**, so it is a pre-existing #3170 side-effect, not this PR.

## Hygiene

All work in a scratch clone; the shared checkout was never written to. Verified the two worktrees have separate `out/` roots (`MSBuildThisFileDirectory`-relative) before trusting any build — a shared output path would have invalidated every mutant measurement. Each of the 4 mutants was asserted applied via `git diff` and restored by exact md5 round-trip (`ea53ed84bf4f459c78cf4a3ab5d05730`) with a rebuild after each; final tracked-dirty **0**, final suite **16/16 green**. Note that `GSharp.Core.dll` md5 differs between trees even with identical sources (non-deterministic MVIDs), so only *within-tree* md5 movement is a valid signal — cross-tree md5 comparison would be meaningless here.

---

**MERGEABLE.** The PR does exactly what #3179 asked, and the measurement backs every acceptance criterion: the `start` mutant flips from surviving 15/15 to killed 3/16 with only the test file changing, the `{Length}` pin still kills, and the `GS0511` guard turns out to close a second hole `main` could not see. Product untouched, CI green including `ilverify`. The three NITs are cosmetic and none affects coverage.

PR #3190, opened 2026-08-04
Rename reachable deinit interpreter test

### LINKED ISSUE #3181: [test] Deinit test name at :142 still describes ADR-0068 row 1 while its fixture is row 3; ForcedCollection conjunct is provably inert
Follow-up from review of #3172 (issue #3161). The rename there is a real improvement — it dropped a comparative claim the test never discriminated. But two of the three conjuncts in the replacement name do not survive measurement.

`test/Interpreter.Tests/Issue2988DeinitInterpreterTests.cs:142`, currently `ForcedCollectionSkipsDeinitializerAndWarnsOnce`.

## "ForcedCollection" is provably inert

Delete `GC.Collect()` / `GC.WaitForPendingFinalizers()` at `:160-161` and re-run:

```
emit stdout: identical    gsi stdout: identical    GS0510: still 1
```

## "SkipsDeinitializer" is true but is not an interpreter property

The emitted assembly — the **oracle** — prints byte-identical stdout with no `deinit-11`. The skip is not the interpreter boundary; it is ordinary CLR behaviour for a **reachable** object.

Cause: the fixture holds the instance alive. `var held = Make()` at `:157` and `Console.WriteLine(held.Tag)` at `:164` keep it reachable **across** the collect, so the forced collection has nothing to collect.

The fixture *could* have discriminated — let the object die instead:

```
func Make() { var r = Res("kept"); Console.WriteLine("made-22"); GC.KeepAlive(r) }
Make(); GC.Collect(); GC.WaitForPendingFinalizers()

emit:  made-22 / deinit-11: kept / after-collect-33     <- deinit RUNS
gsi:   made-22 / after-collect-33                       <- deinit SKIPPED
```

Those are **ADR-0068's own table rows** (`docs/adr/0068-deinit-destructor-support.md:109-114`). The current fixture is **row 3** ("object remains reachable at exit -> no deinit in either"); the name describes **row 1** ("object dies, collection forced"). The ADR already documents that this fixture cannot show a driver difference.

## Only "AndWarnsOnce" is load-bearing, and it is last in the name

Mutation evidence: dropping the once-per-declaring-class guard (`Compilation.cs:337`, the `ReferenceEquals(deinitializer?.Function, function)` conjunct) kills `DeinitializersWarnOncePerDeclaringClassWithoutRunning` (`:21`) and both Theory driver cases — but **not** `:142`. The `Assert.Single` at `:170` is "exactly one GS0510 in a one-class cell": real (the `:332` control kills it) but **not** the once-*per-declaring-class* property, which `:21` owns by name and by mutation.

## Fix — either is small

1. **Name the causal property the rename dropped**, e.g. `ReachableInstanceWarnsOnceWithoutRunningDeinitializer`. Reachability is *why* nothing finalizes; forced collection is the part that does nothing.
2. **#3161's own option 1** — name it for GS0510 and stop mentioning collection at all.

## Do NOT delete the GC.Collect() lines

They are inert but defensible: a provocation against a future change wiring interpreter finalization to GC. That is an argument for **keeping the code**, not for **foregrounding it in the name**. The provocation is weak here anyway (a reachable object would not finalize even under a correct GC-wired implementation), and the strong version already exists at `:71` across all three drivers.

## Hard constraint

**ADR-0068 specifies that `gsi` never runs deinitializers.** Any change making it do so is a **language regression**, not a fix. This issue is about a **name**; it must not become a product change.

DavidObando:
Closes #3181.

## Summary

- rename `ForcedCollectionSkipsDeinitializerAndWarnsOnce` to `ReachableInstanceReportsGS0510WithoutRunningDeinitializer`
- update the cross-reference added by ADR-0156 Phase 2
- preserve the forced collection fixture and every assertion
- make no product or documentation change

The new name foregrounds the properties this one-class evaluator test pins: the instance remains reachable, no deinitializer output occurs, and exactly one GS0510 is reported. It no longer claims forced collection caused the skip or that this cell proves the once-per-declaring-class guard.

## Coverage preservation

- target file: 240 lines on both sides
- 4 `[Fact]`, 1 `[Theory]`, and 22 `Assert.` calls on both sides
- after stripping only the renamed signature, base/head MD5: `5b8d2c7c3a3b74d0390ea2baaa48df10`
- repo-wide old-name search: zero hits
- `git diff --name-only origin/main...HEAD -- src/ docs/`: empty

## Product mutations

| Mutant | Result |
|---|---|
| `Compilation.cs`: replace declaring-function identity guard with `deinitializer is not null` | RED: `DeinitializersWarnOncePerDeclaringClassWithoutRunning` failed; renamed test survived, confirming it should not claim once-per-class |
| `Compilation.cs`: invert `!allErrors.Any()` | RED: 3/7 failed, including `ReachableInstanceReportsGS0510WithoutRunningDeinitializer` at its `Assert.Single` |

Both mutants were asserted by diff and source MD5, produced changed `GSharp.Core.dll` bytes, then restored to source MD5 `b3ca864d484665dd522394f9c5bee7d5` and DLL MD5 `dd5c1dd46d0631d5172cd96efadd192b` after rebuild. The former `Program.cs` stderr mutant from #3172 is no longer applicable: #3182 removed the script-mode GS0510 path and replaced that test with an emitted-finalizer assertion.

## Verification

- final merged-tree Release warnings-as-errors build: rc 0, 0 warnings, 0 errors
- focused evaluator/deinit tests: 7/7
- all 9 solution test projects (denominator from `dotnet sln GSharp.sln list`): 15,338 passed, 1 skipped on the immediately preceding main; after the final one-commit main refresh, the focused 7/7 and incoming commit's 15/15 tests passed
- merge-tree output: `03706aea8be2b72f9c2597f9bec4751dc0c18329`

## Current driver measurement

Main changed during the work: #3182 now runs bare `gsc` and script-mode `gsi` through emitted code. For the dying-object probe, all three drivers return rc 0 and print `made-22 / deinit-11: kept / after-collect-33` (bare `gsc` also prints `Success.`). The old expected script-driver divergence is therefore no longer true on current main. This PR does not alter that behavior; the renamed test directly exercises the evaluator `SessionEngine`, where GS0510 remains the boundary.

### DIFF
--- a/test/Interpreter.Tests/EmittedSessionEngineTests.cs
+++ b/test/Interpreter.Tests/EmittedSessionEngineTests.cs
@@ -331,7 +331,7 @@ public sealed class EmittedSessionEngineTests : IDisposable
     /// run real emitted code, so a class deinitializer executes as a genuine
     /// CLR finalizer when a collection is forced — and the GS0510 interpreter
     /// boundary warning (pinned for the evaluator engine by
-    /// <c>Issue2988DeinitInterpreterTests.ForcedCollectionSkipsDeinitializerAndWarnsOnce</c>)
+    /// <c>Issue2988DeinitInterpreterTests.ReachableInstanceReportsGS0510WithoutRunningDeinitializer</c>)
     /// no longer fires, matching Phase 1's script-mode behavior.
     /// </summary>
     [Fact]
--- a/test/Interpreter.Tests/Issue2988DeinitInterpreterTests.cs
+++ b/test/Interpreter.Tests/Issue2988DeinitInterpreterTests.cs
@@ -125,7 +125,7 @@ public class Issue2988DeinitInterpreterTests
     }
 
     [Fact]
-    public void ForcedCollectionSkipsDeinitializerAndWarnsOnce()
+    public void ReachableInstanceReportsGS0510WithoutRunningDeinitializer()
     {
         var source = """
             import System


DavidObando:
## **MERGEABLE** — no BLOCKER. The headline risk did not materialize, and the rename is better-supported than the name the issue itself proposed.

Anchors taken live, not from the brief: `origin/main` **`81e6deab`**, PR head **`b15c92d4`** via `gh pr view --json headRefOid`. **Ancestry verified, not existence** — `git merge-base --is-ancestor origin/main b15c92d4` rc=0, merge-base *is* `81e6deab` exactly, 0 behind / 1 ahead, one commit. Fresh worktree at head, Release build rc=0, 0 warnings. Freshness checked in epoch seconds because an 18-second solution build looked wrong: all **83** first-party `GSharp*.dll` had mtimes 61–77 s old against `date +%s`, `out/` is worktree-local, so the build was real and the elapsed time was just a fast machine.

### The dominant constraint: product scope is empty

```
git diff --name-only 81e6deab b15c92d4 -- . ':(exclude)test/'   ->   (empty)
```

Two files, `+2/−2`, both under `test/Interpreter.Tests/`. No `src/`, no `docs/`. **A name fix did not become a product fix.**

### Coverage preservation — reproduced independently to #3172's standard

| | base `81e6deab` | head `b15c92d4` |
|---|---|---|
| `Issue2988DeinitInterpreterTests.cs` md5, identifier normalised | `75fdc82b208364bc6350828ba29a2a70` | **identical** |
| `EmittedSessionEngineTests.cs` md5, identifier normalised | `7624c75b31a2e7addd6021328c638015` | **identical** |
| lines | 240 / 745 | 240 / 745 |
| `[Fact]` / `[Theory]` | 4 / 1 | 4 / 1 |
| `Assert.` | 22 | 22 |
| **`GC.Collect()` occurrences** | **3** | **3** |

My md5s differ in value from the PR body's `5b8d2c7c…` only because we normalised differently — the body stripped the signature line, I substituted the identifier. Both are sound and both reach the same conclusion. The row that matters is the last one: **the inert `GC.Collect()` lines were not deleted**, which is what #3181 explicitly asked for.

Blast radius zero: repo-wide grep for `ForcedCollectionSkipsDeinitializerAndWarnsOnce` returns **no hits**; the new name appears exactly twice — the declaration at `:128` and the `<c>` cross-reference at `EmittedSessionEngineTests.cs:334` that the PR correctly updated in the same commit.

### Is the new name entailed by the assertions? Yes — all three conjuncts

`Issue2988DeinitInterpreterTests.cs:152-157`:

- **ReachableInstance** — `Assert.Equal("made-22\nafter-collect-33\nkept\n", cell.Output)` at `:155`. The trailing `kept` is `held.Tag` read *after* `GC.Collect()`/`WaitForPendingFinalizers()`, so reachability across the collect is genuinely witnessed by an assertion. Remove the reachability and this assertion fails. Unlike `GC.Collect()`, it is not deletable.
- **ReportsGS0510** — `Assert.Single(cell.Diagnostics, d => d.Id == "GS0510")` at `:156`.
- **WithoutRunningDeinitializer** — the same exact-equality at `:155` contains no `deinit-11`.

### Two-sided PRODUCT mutants — and M2 is the one that justifies the rename

The brief's targets were stale, so I relocated them: `src/Core/CodeAnalysis/Compilation.cs` no longer exists (it is `src/Core/CodeAnalysis/**Compilation/**Compilation.cs`), the once-per-class guard is at **`:350`** not `:337`, and the error control is at **`:345`** not `:332`. My source md5 for that file, `b3ca864d484665dd522394f9c5bee7d5`, **matches the PR body byte-for-byte**, so we mutated identical source.

| mutant | site | build | `GSharp.Core.dll` md5 | result |
|---|---|---|---|---|
| **M1 control** | `Compilation/Compilation.cs:345` `!allErrors.Any()` → `allErrors.Any()` | rc=0 | `6b373434…` → `692f056c…` | **RED, 3 failed** — kills `CompilationErrorsSuppress…`, `DeinitializersWarnOncePerDeclaringClass…`, **and the renamed test** |
| **M2** | `Compilation/Compilation.cs:350` `ReferenceEquals(deinitializer?.Function, function)` → `deinitializer is not null` | rc=0 | `6b373434…` → `d21fec04…` | **1 failed: `DeinitializersWarnOncePerDeclaringClassWithoutRunning` only — renamed test SURVIVES** |

Both built cleanly and both moved the DLL bytes, so neither is a no-result. M1 is the control that makes M2 readable: it proves the oracle is live *for the renamed test specifically*, so M2's survival is a real survival.

**M2 is the evidence that the rename is correct.** The once-per-declaring-class property is owned by `:21` and not by this test — so dropping `AndWarnsOnce` is right. Note this makes the PR's name **better than the one #3181 proposed**: the issue suggested `ReachableInstanceWarnsOnceWithoutRunningDeinitializer`, which would have re-asserted the very property its own mutation evidence shows this cell does not own. Substituting `ReportsGS0510` fixes that. Restored by exact md5 round-trip — source back to `b3ca864d…` and, after rebuild, `GSharp.Core.dll` back to `6b373434…` bit-for-bit.

### The ADR-0156 landmine: the fixture still means what it meant

Worth stating because the brief was right to flag it. `3d5ba7af` (#3182, Phase 1) rewrote 35/34 lines of this very file *after* #3181 was filed, which is why every line cite in the issue and brief is off by ~14. But the renamed test calls `new SessionEngine{…}.Evaluate(source)` **in-process at `:152`** — it never goes near the CLI driver ADR-0156 migrated. `ec50744d` (#3177), `3d5ba7af` (Phase 1) and `ed6ae55e` (Phase 2, opt-in) are all in `81e6deab`, and none of them touch that call. The fixture is intact.

### One thing measured that changes the framing of this whole issue

I drove the same `SessionEngine` the test drives, out of tree, across ADR-0068's three rows:

| probe | GS0510 | deinit ran | output |
|---|---|---|---|
| **A** row 3 reachable (**the fixture**) | 1 | no | `made-22 \| after-collect-33 \| kept` |
| **B** row 1 dying + forced collect | 1 | no | `made-22 \| after-collect-33` |
| **C** row 2 dying, no collect | 1 | no | `made-22 \| after-collect-33` |
| **D** fixture minus `GC.Collect()` | 1 | no | **identical to A** |

D re-confirms #3181's "provably inert" finding on *this* head. But A vs B is the interesting one: **in the evaluator, reachability does not change the outcome either.** A dying, forcibly-collected object yields the same GS0510=1 and the same absent deinitializer, exactly as ADR-0068's prose says ("Interpreted execution … skips its deinitializer", unconditionally). The axis that actually discriminates is the **engine**, not reachability — which the file already pins at `:185`, where script-mode `gsi` runs the finalizer and emits no GS0510.

That is a NIT, not an objection, and I want to be clear I am not asking for a third rename of this test.

### Findings

- **NIT** — `Issue2988DeinitInterpreterTests.cs:128`. The name foregrounds reachability, which is true and assertion-witnessed but (probes A/B above) does not discriminate the outcome in this engine, and omits the engine, which does. The PR body calls this "this one-class **evaluator** test"; the method name does not say so, and the same class now contains `ScriptRunnerRunsDeinitializersWithoutBoundaryWarning` at `:185` asserting the opposite behaviour. If it is ever touched again, `Evaluator…` would carry more information than `ReachableInstance…`. **Explicitly not worth a third rename on its own** — #3172 renamed it, #3181 renamed it again, and the churn now costs more than the residual imprecision.
- **NIT** — `:156` asserts cardinality with `Assert.Single`; `ReportsGS0510` does not encode "exactly one". Minor, and the alternative reintroduces the "once" wording that M2 shows this cell does not own, so the current choice is defensible.
- **SHOULD-FIX, follow-up issue, explicitly NOT this PR** — `docs/adr/0068-deinit-destructor-support.md:118-122` is now false on `main`. Its table says the dying-object probe gives `gsi` → `made-22 / body-33`. Measured on `81e6deab`, all three drivers print the finalizer and none emits GS0510:

  | driver | rc | stdout |
  |---|---|---|
  | bare `gsc` | 0 | `made-22 / deinit-11: kept / after-collect-33 / Success.` |
  | `gsc /out:` then run | 0 / 0 | `made-22 / deinit-11: kept / after-collect-33` |
  | `gsi <file>` | 0 | `made-22 / deinit-11: kept / after-collect-33` |

  ADR-0156 `:198-200` anticipates this deliberately, but ADR-0068 still reads `**Status**: Accepted` with no pointer to ADR-0156 and no amendment. A future agent treating ADR-0068 as the boundary spec — as I was briefed to — will be misled. This is #3182's drift, not this PR's; a rename PR is right to make no docs change.

### Denominator, and what I deliberately did not run

**`Interpreter.Tests` 1393 passed / 0 failed / 0 skipped, rc=0** on the head build. (The 1,264 reference figure is stale; `main` has moved.)

I ran **1 of the 9** projects — denominator from `dotnet sln GSharp.sln list`, CI's own per `build/generate-ci-test-matrix.py:26` — and I am stating that plainly rather than implying more. The justification is containment, measured not assumed: the base↔head diff outside `test/` is empty, and no `.csproj` in the repo `Compile Include`s either changed file, so both compile solely into `Interpreter.Tests`. The only inbound link is `Interpreter.Tests.csproj:13` pulling `..\Shared\Issue3081CompositeLiteralCases.cs` *in*. Nothing else can observe this change. `Compiler.Tests` (~1h18m) was therefore skipped intentionally. CI is green on all 25 checks on this head, including `ilverify`, `e2e`, `cs2gs-corpus` and `cs2gs-oahu`.

### PR body accuracy

Every claim I could check independently held: 240 lines, 4/1/22, zero old-name hits, empty `src/ docs/` diff, both mutant outcomes, and the `Compilation.cs` source md5 to the byte. The body also correctly retires the `Repl/Program.cs` stderr mutant from #3172 — I confirmed `Program.cs:61` is now `--engine` argument parsing, so that target no longer exists. The DLL md5 it quotes (`dd5c1dd4…`) does not match mine (`6b373434…`), which is expected: assembly MVIDs are per-build and per-path, so a DLL hash is only meaningful inside one machine's own round-trip, which is how both of us used it.

### Self-disclosures

- My "which tests failed" grep over-matched stack-frame tokens, listing `Issue2988DeinitInterpreterTests.cs` and `AssertWarning` as if they were tests. They are a filename and a helper. I treated the `Failed: 3` / `Failed: 1` summary lines as authoritative and the grep as a hint; the mutant conclusions rest on the counts.
- I nearly accepted an 18-second solution build. Epoch-second mtimes on 83 first-party DLLs, not clock strings, are what confirmed it was genuine.
- `dotnet build` left an untracked `test-assets/Issue3119.CrossConstants/packages.lock.json` in my worktree. Build side-effect, unrelated to this PR; I removed it so my final clean-tree claim is exact.
- Worktree ends `git status --porcelain` empty at `b15c92d4`; the shared checkout was never written and remains clean at `81e6deab`.

PR #3191, opened 2026-08-04
Fix enclosing generic field reification

### LINKED ISSUE #3180: [bug] Fields typed by an ENCLOSING generic's parameter still erase to object under gsi and bare gsc
Follow-up from review of #3151 (issue #3137). That PR fixed reified CLR type arguments for a type's **own** generic parameters. A field typed by an **enclosing** generic's parameter still erases.

## Repro

```
class Owner[T] { class Inner { var V T; var W int32 } }
class Box[T] : EventArgs {}
var b = Box[Owner[int32].Inner]()
Console.WriteLine(FieldShape(b.GetType().GenericTypeArguments[0]))
```

| driver | base | after #3151 |
|---|---|---|
| `gsc /out:` (**oracle**) | `2|V:System.Int32|W:System.Int32` | `2|V:System.Int32|W:System.Int32` |
| bare `gsc` (evaluate) | `0` — zero fields | `2|V:`**`System.Object`**`|W:System.Int32` |
| `gsi` (interpret) | `0` — zero fields | `2|V:`**`System.Object`**`|W:System.Int32` |

**Not a regression.** #3151 narrowed the divergence from "no fields at all" to "right count, wrong type". But the oracle still disagrees — **silently, rc0, no diagnostic**.

## Cause

`Evaluator.Expressions.cs:1312-1315` defines the **enclosing** builder's generic parameters but **discards the return value**. `:1333-1341` populates `reifiedTypeParameters` from `definition.TypeParameters` only, and does so inside `if (arity > 0)`. A type nested in a generic has its **own** arity 0, so the map is empty and the enclosing `T` falls through to `typeof(object)`.

Structurally this is the **same omission** as #3151's round-1 BLOCKER — a `DefineGenericParameters` result that is not threaded — one axis over.

## The obvious remedy does NOT work — measured, so nobody repeats it

Threading the enclosing parameters the mirror-image way:

```csharp
var enclosingParameters = enclosingBuilder.DefineGenericParameters(...);
for (var i = 0; i < enclosingParameters.Length; i++)
{
    reifiedTypeParameters[enclosing.TypeParameters[i]] = enclosingParameters[i];
}
```

builds clean (`BUILD_RC=0`, 0 warnings) and then **fails at runtime on both interpreting drivers**:

```
error GS9999: Type Inner from assembly GSharp.Interpreter.ClrBacking.2 has a field of an illegal type.
```

That is **CLR-correct, not a bug in the patch**: in metadata a nested type of a generic must **redeclare** the enclosing arity (C# emits `Owner<T>.Inner<T>`), so the reified nested builder — arity 0 — has the enclosing `GenericTypeParameterBuilder` out of scope for its fields.

**The real fix is structural: the reified nested builder must carry the enclosing arity.** Budget for that, not for a one-liner.

## Why the existing matrix did not catch it

`Issue3099TypeArgumentReificationTests.cs` already ships cases literally named `nested-class` (`:30`) and `nested-struct` (`:31`) — but **both nest inside a non-generic `Owner`**, so the generic x nested cross-product is untested. The names create confidence the coverage does not earn. This is #3137's own "specimens absent, not assertions blind" diagnosis, one axis over — and the **ninth** instance in this repo of a test naming a property it does not discriminate.

## Acceptance

- `V` reports `System.Int32` on all three drivers for the repro above, matching the oracle.
- The four recursive branches #3151 pinned (`NullableTypeSymbol` `:1432`, `ImportedTypeSymbol` `:1446`, `SliceTypeSymbol` `:1457`, `ArrayTypeSymbol` `:1462`) stay pinned and green — they are mutation-covered today, keep them that way.
- Add a **generic x nested** specimen to the matrix rather than renaming the existing non-generic ones.
- **If the structural fix proves too large, the honest alternative is to add the shape as a documented known divergence** — but the current state, where the names imply coverage that does not exist, is not acceptable either way.

DavidObando:
## Summary

- reify evaluator-created nested CLR types over the flattened enclosing + own generic parameter list
- close those nested types with `EnclosingTypeArguments` followed by own `TypeArguments`
- extend the existing emitted-oracle matrix with class, struct, multi-arity, nested-generic, and deep-nesting specimens
- keep an explicit `SessionEngine` column after ADR-0156 moved bare `gsc` and `gsi <file>` from evaluation to emit-to-memory

Closes #3180.

## Main interaction

Issue assignment started on `aa453ac6`, where bare `gsc` and `gsi <file>` reproduced `V:System.Object`. Main then merged #3182/#3186: file-mode drivers now execute emitted assemblies, so on final base `81e6deab` all three file columns already print the emitted result. Default interactive `SessionEngine` still uses the evaluator and retained the bug. Coverage now names and measures that remaining evaluator path instead of passing vacuously through three emit paths.

## Measured behavior

Exact repro on final head:

| path | output |
|---|---|
| bare `gsc` | `2|V:System.Int32|W:System.Int32` + `Success.` |
| `gsc /out:` | `2|V:System.Int32|W:System.Int32` |
| `gsi <file>` | `2|V:System.Int32|W:System.Int32` |
| interactive evaluator | matches emitted oracle in the 16-case matrix |

Final test copied onto current main without the production change: **5 failed / 11 passed**. Failing set:

- `generic-owner-nested-class`
- `generic-owner-nested-struct`
- `generic-owner-multi-arity-nested`
- `generic-owner-nested-generic`
- `generic-owner-deep-nested`

Head: **16/16 passed**.

## Product mutations

Every mutant was asserted by `git diff` and md5, then restored to exact md5.

| mutant | result |
|---|---|
| target uses own parameters only, dropping flattened enclosing parameters | RED: 5 failed / 11 passed |
| invert concrete enclosing-argument selection | RED: 5 failed / 11 passed |
| invert target generic-definition condition (broad control) | RED: 16 failed / 0 passed |
| flatten parameters on intermediate enclosing builders | SURVIVED 16/16; hunk proved unnecessary and was deleted |

## Validation

- final base: `81e6deab4ad19f2f642182691c42146dd63ddb27`
- final head: `6944c1bcda4c40673d12727a2aaf5ba1dd04fd7b`
- Debug and Release solution builds: 0 warnings, 0 errors
- merge-tree output `3021468e5a31c1e1b6a50577199d9ae8ec714aae`, exactly equal to `HEAD^{tree}`
- full unfiltered 9-project run: **15,343 passed, 1 skipped, 0 failed**
  - Compiler 4,245
  - Core 7,128 + 1 skipped
  - Extensions 104
  - InternalAnalyzers 9
  - Interpreter 1,398
  - LanguageServer 462
  - SDK 59
  - Cs2Gs 1,907
  - GeneratorHost 31
- Cs2Gs first run lacked Release packages and failed its package-smoke prerequisite; after solution Release build, full rerun passed 1,907/1,907
- full 9-project run was on preceding base `0ffd90b1`; final main added only #2903 (`MemberLookup`, overload resolution, one Compiler test file), with no file overlap. After final rebase: solution build passed, issue matrix passed 16/16, and #2903's new interaction class passed 15/15.

Not run: local SDK e2e, because shared `~/.nuget/packages/gsharp.net.sdk/` is prohibited.

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -1295,7 +1295,9 @@ public sealed partial class Evaluator
     {
         var module = CreateClrBackingModule();
         var definition = structType.Definition ?? structType;
-        var arity = definition.TypeParameters.Length;
+
+        // Issue #3180: CLR nested types redeclare the flattened enclosing and own generic parameters.
+        var reifiedParameters = GetClrReifiedTypeParameters(definition);
         var reifiedTypeParameters = new Dictionary<TypeParameterSymbol, Type>();
         var enclosingTypes = new Stack<StructSymbol>();
         for (var containing = definition.ContainingType as StructSymbol;
@@ -1343,13 +1345,13 @@ public sealed partial class Evaluator
                 metadataName,
                 (attributes & ~TypeAttributes.VisibilityMask) | TypeAttributes.NestedPublic,
                 clrBase);
-        if (arity > 0)
+        if (!reifiedParameters.IsDefaultOrEmpty)
         {
             var parameters = typeBuilder.DefineGenericParameters(
-                definition.TypeParameters.Select(static parameter => parameter.Name).ToArray());
+                reifiedParameters.Select(static parameter => parameter.Name).ToArray());
             for (var i = 0; i < parameters.Length; i++)
             {
-                reifiedTypeParameters[definition.TypeParameters[i]] = parameters[i];
+                reifiedTypeParameters[reifiedParameters[i]] = parameters[i];
             }
         }
 
@@ -1416,18 +1418,35 @@ public sealed partial class Evaluator
             enclosingBuilders[i].CreateType();
         }
 
-        if (arity == 0)
+        if (reifiedParameters.IsDefaultOrEmpty)
         {
             return backingType;
         }
 
-        var typeArguments = new Type[structType.TypeArguments.Length];
-        for (var i = 0; i < typeArguments.Length; i++)
+        var definitionEnclosingParameters = StructSymbol.CollectEnclosingTypeParameters(definition);
+        var typeArguments = new Type[reifiedParameters.Length];
+        for (var i = 0; i < definitionEnclosingParameters.Length; i++)
         {
-            typeArguments[i] = ResolveClrBackingTypeArgument(structType.TypeArguments[i]);
+            var argument = i < structType.EnclosingTypeArguments.Length
+                ? structType.EnclosingTypeArguments[i]
+                : definitionEnclosingParameters[i];
+            typeArguments[i] = ResolveClrBackingTypeArgument(argument);
+        }
+
+        for (var i = 0; i < definition.TypeParameters.Length; i++)
+        {
+            var argument = i < structType.TypeArguments.Length
+                ? structType.TypeArguments[i]
+                : definition.TypeParameters[i];
+            typeArguments[definitionEnclosingParameters.Length + i] = ResolveClrBackingTypeArgument(argument);
         }
 
         return backingType.MakeGenericType(typeArguments);
+    }
+
+    private static ImmutableArray<TypeParameterSymbol> GetClrReifiedTypeParameters(StructSymbol type)
+    {
+        return StructSymbol.CollectEnclosingTypeParameters(type).AddRange(type.TypeParameters);
     }
 
     private static Type ResolveClrBackingTypeArgument(
--- a/test/Interpreter.Tests/Issue3099TypeArgumentReificationTests.cs
+++ b/test/Interpreter.Tests/Issue3099TypeArgumentReificationTests.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Reflection;
 using System.Text.RegularExpressions;
 using GSharp.Compiler;
+using GSharp.Repl.Engine;
 using Xunit;
 
 namespace GSharp.Interpreter.Tests;
@@ -17,6 +18,8 @@ namespace GSharp.Interpreter.Tests;
 /// Issue #3099: interpreter-created generic backing types must preserve every
 /// G# type-argument kind instead of erasing value types to <see cref="object"/>.
 /// Issue #3137 extends the same emit-oracle matrix to reflected field shape.
+/// Issue #3180 covers enclosing generic parameters in the interactive evaluator;
+/// ADR-0156 file drivers remain compared against the emitted oracle.
 /// </summary>
 [Collection("ConsoleIo")]
 public class Issue3099TypeArgumentReificationTests
@@ -29,6 +32,11 @@ public class Issue3099TypeArgumentReificationTests
         yield return ["string", string.Empty, "string"];
         yield return ["nested-class", "class Owner {\n    class Payload {\n        let Eleven int32\n        var TwentyTwo string\n    }\n}", "Owner.Payload"];
         yield return ["nested-struct", "class Owner {\n    struct Payload {\n        let Eleven int32\n        var TwentyTwo string\n    }\n}", "Owner.Payload"];
+        yield return ["generic-owner-nested-class", "class Owner[T] {\n    class Payload {\n        let Eleven T\n        var TwentyTwo string\n    }\n}", "Owner[int32].Payload"];
+        yield return ["generic-owner-nested-struct", "class Owner[T] {\n    struct Payload {\n        let Eleven T\n        var TwentyTwo string\n    }\n}", "Owner[int32].Payload"];
+        yield return ["generic-owner-multi-arity-nested", "class Owner[TFirst, TSecond] {\n    class Payload {\n        let Eleven TFirst\n        var TwentyTwo TSecond\n    }\n}", "Owner[int32, string].Payload"];
+        yield return ["generic-owner-nested-generic", "class Owner[T] {\n    class Payload[U] {\n        let Eleven T\n        var TwentyTwo U\n    }\n}", "Owner[int32].Payload[string]"];
+        yield return ["generic-owner-deep-nested", "class Owner[T] {\n    class Middle[U] {\n        class Payload {\n            let Eleven T\n            var TwentyTwo U\n        }\n    }\n}", "Owner[int32].Middle[string].Payload"];
         yield return ["gsharp-generic", "struct Payload[T] {\n    let Value T\n    var Imported List[T]\n    var Slice []T\n    var Nullable T?\n    var Array [3]T\n    shared {\n        var Stat int32\n    }\n    const Kilo int32 = 1000\n}", "Payload[int32]"];
         yield return ["imported-generic", "struct Payload {\n    let Eleven int32\n    var TwentyTwo string\n}", "List[Payload]"];
         yield return ["nullable", "struct Payload {\n    let Eleven int32\n    var TwentyTwo string\n}", "Payload?"];
@@ -38,7 +46,7 @@ public class Issue3099TypeArgumentReificationTests
 
     [Theory]
     [MemberData(nameof(TypeArgumentCases))]
-    public void TypeArguments_AgreeAcrossEvaluationEmitAndInterpretation(
+    public void TypeArguments_AgreeAcrossInteractiveEvaluationEmitAndFileDrivers(
         string caseName,
         string declaration,
         string typeArgument)
@@ -52,17 +60,18 @@ public class Issue3099TypeArgumentReificationTests
 
         try
         {
-            var compilerEvaluation = RunSourceDriver(
+            var compilerFile = RunSourceDriver(
                 Path.Combine(root, "gsc-eval"),
                 source,
                 Program.Main);
-            Assert.EndsWith("Success.\n", compilerEvaluation);
-            compilerEvaluation = compilerEvaluation[..^"Success.\n".Length];
+            Assert.EndsWith("Success.\n", compilerFile);
+            compilerFile = compilerFile[..^"Success.\n".Length];
 
-            var interpreter = RunSourceDriver(
+            var gsiFile = RunSourceDriver(
                 Path.Combine(root, "gsi"),
                 source,
                 GSharp.Repl.Program.Main);
+            var interactive = RunInteractive(source);
 
             var emitDirectory = PrepareEmptyDirectory(Path.Combine(root, "gsc-emit"));
             var sourcePath = Path.Combine(emitDirectory, "Probe.gs");
@@ -90,15 +99,18 @@ public class Issue3099TypeArgumentReificationTests
             });
 
             var expected = NormalizeGenericTypeNames(emitted);
-            var evaluated = NormalizeGenericTypeNames(compilerEvaluation);
-            var interpreted = NormalizeGenericTypeNames(interpreter);
+            var evaluated = NormalizeGenericTypeNames(interactive);
+            var compiledFile = NormalizeGenericTypeNames(compilerFile);
+            var interpretedFile = NormalizeGenericTypeNames(gsiFile);
             const string ControlShape =
                 "44\n2|Eleven:System.Int32:True:False:False|TwentyTwo:System.String:False:False:False\n";
             Assert.Contains(ControlShape, expected, StringComparison.Ordinal);
             Assert.Contains(ControlShape, evaluated, StringComparison.Ordinal);
-            Assert.Contains(ControlShape, interpreted, StringComparison.Ordinal);
+            Assert.Contains(ControlShape, compiledFile, StringComparison.Ordinal);
+            Assert.Contains(ControlShape, interpretedFile, StringComparison.Ordinal);
             Assert.Equal(expected, evaluated);
-            Assert.Equal(expected, interpreted);
+            Assert.Equal(expected, compiledFile);
+            Assert.Equal(expected, interpretedFile);
             Assert.Contains("11\n", expected);
             Assert.Contains("22\n", expected);
             Assert.Contains("33\n", expected);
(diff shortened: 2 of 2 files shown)


DavidObando:
# Review — PR #3191 `fix/3180-work` @ `6944c1bc`

**Verdict: MERGEABLE.** The fix is the structural one the issue asked for, not the one-liner the issue ruled out; it is two-sided and mutation-pinned; the four branches #3151 pinned are byte-identical across all drivers on both trees; ILVerify is clean.

**0 BLOCKER, 1 SHOULD-FIX, 3 NIT.**

Head verified live: `gh pr view 3191 --json headRefOid` = `6944c1bc`, matches the briefed SHA. `main` = `16f97229`. Branch is **1 behind / 1 ahead**; merge base `81e6deab`. `git merge-tree` **output** is one line (tree `af38d562`), **0 conflict markers**. Main's single intervening commit (`16f97229`, #3190) touches `EmittedSessionEngineTests.cs` and `Issue2988DeinitInterpreterTests.cs` — **disjoint** from this PR's two files.

**Main moved during the run** (`16f97229` → `917677ec`, #3189). Re-checked rather than assumed: the new commit touches only `test/Interpreter.Tests/Issue3114ReadOnlySpanInterpreterTests.cs` — **zero intersection** with this PR's two files — and `git merge-tree` against the new `main` is still a one-line output with 0 conflict markers. Measurements below stand.

Both trees built Release: `BUILD_RC=0`, **0 warnings**. `GSharp.Core.dll` differs across trees (`8dbab70e` base vs `0ccf620d` PR) on distinct inodes — the two sides are genuinely distinct binaries, not a shared artifact. (Equal *file sizes* were a coincidence; I checked rather than assumed.)

---

## 1. The driver framing in issue #3180 is obsolete — but the defect is NOT superseded

This was the sharp risk, and it needed measuring rather than tabulating. ADR-0156 has landed on `main`: #3177 (`ec50744d`), **Phase 1** (`3d5ba7af`) and Phase 2 (`ed6ae55e`, opt-in). Phase 1 routes `gsi <file>` **and bare `gsc`** through emit-to-memory.

Measured on **clean `main` `16f97229`**, the issue's own repro:

| driver | issue #3180 claims | **measured on clean main** |
|---|---|---|
| `gsc /out:` (**oracle**) | `System.Int32` | `System.Int32` |
| bare `gsc` (file) | **`System.Object`** | **`System.Int32`** — issue is stale |
| `gsi` (file) | **`System.Object`** | **`System.Int32`** — issue is stale |
| interactive `SessionEngine` | *(not listed)* | **`System.Object`** — defect live |

**Two of the issue's three rows are now wrong.** The bug moved rather than disappeared: it survives on the one remaining evaluator consumer, the **interactive REPL**, which is the shipped default — `ReplHost.cs:18` `Run() => Run(new SessionEngine())`, and `Repl/Program.cs:112-130` makes `--engine emit` / `GSI_ENGINE=emit` **opt-in** ("the default remains the tree-walking evaluator until the flip lands").

**So this is not a supersession and should not be closed as one.** It is a real, user-reachable, silent wrong answer on the default interactive driver.

**Credit where due: the author found this independently and retargeted the test.** The diff adds a fourth column, `RunInteractive` via `SessionEngine` (`:195-201`), renames the test to `TypeArguments_AgreeAcrossInteractiveEvaluationEmitAndFileDrivers` (`:49`), and documents it at `:21-22`. That is exactly the right response to a moving driver model.

**Action for the issue, not the PR:** #3180's repro table should be corrected before it is closed, or the next reader will re-derive a table that no longer holds.

## 2. It is the structural fix, not the ruled-out one-liner

The issue ruled out threading enclosing parameters onto an arity-0 nested builder (`GS9999 … has a field of an illegal type`). This PR does **not** do that. It makes the reified nested builder **redeclare the flattened enclosing + own parameters**, which is what the CLR requires and what C# emits as `Owner<T>.Inner<T>`:

`Evaluator.Expressions.cs:1447-1450`
```csharp
private static ImmutableArray<TypeParameterSymbol> GetClrReifiedTypeParameters(StructSymbol type)
    => StructSymbol.CollectEnclosingTypeParameters(type).AddRange(type.TypeParameters);
```

Good ponytail: both helpers **already existed** (`StructSymbol.cs:1447` and `:520`) with a documented 1:1 alignment contract (`:1365`). The PR adds no new symbol surface — it reuses the binder's existing flattening, which is why `src/` is +35/-17 rather than a new subsystem.

Executed, not argued — repro on both trees:

| driver | base `16f97229` | PR `6944c1bc` |
|---|---|---|
| bare `gsc` (file) | `System.Int32` | `System.Int32` |
| `gsc /out:` (**oracle**) | `System.Int32` | `System.Int32` |
| `gsi` (file) | `System.Int32` | `System.Int32` |
| **interactive** | **`System.Object`** | **`System.Int32`** |

## 3. The four #3151-pinned branches do not regress — probed, not inferred

`struct Payload[T]` with `Value T`, `Imported List[T]`, `Slice []T`, `Nullable T?`, `Array [3]T`, a `shared` static and a `const`. All 7 fields **byte-identical across all four drivers on both trees**:

```
7|Value:System.Int32|Imported:System.Collections.Generic.List`1[[System.Int32, …]]
 |Slice:System.Int32[]|Nullable:System.Int32|Array:System.Int32[]|Stat:System.Int32|Kilo:System.Int32
```

I also probed the **cross-product** the fix could plausibly have missed — an *enclosing* parameter used inside each of those four recursive branches (`Owner[T].Inner` with `List[T]`, `[]T`, `T?`, `[3]T`):

| | base | PR |
|---|---|---|
| `Plain` | `System.Object` | `System.Int32` |
| `Imported` | `List<System.Object>` | `List<System.Int32>` |
| `Slice` | `System.Object[]` | `System.Int32[]` |
| `Nullable` | `System.Object` | `System.Int32` |
| `Array` | `System.Object[]` | `System.Int32[]` |

Base erases **all five**; the PR fixes all five and matches the oracle. The fix generalises rather than special-casing the headline shape.

Further shapes, all four drivers agreeing on the PR head: deep 3-level `Owner[int32].Middle[string].Payload` → `A:System.Int32|B:System.String|C:System.Int32`; multi-arity `Two[int32,string].Inner`; and **type identity** — `Owner[int32].Inner` vs `Owner[string].Inner` produce distinct types (`ReferenceEquals` = `False`) on all four drivers, so the backing-type cache does not conflate enclosing instantiations.

## 4. Two-sided: the test is RED without the product change

PR's test file onto the **base product**, rebuilt: **5 of 16 FAIL**, and they are exactly the 5 new specimens — `generic-owner-nested-class`, `-nested-struct`, `-multi-arity-nested`, `-nested-generic`, `-deep-nested`. The 11 pre-existing cases still pass, so the new file loses no coverage.

All 5 fail at **`:111` `Assert.Equal(expected, evaluated)`** — interactive vs the emit oracle, **full-string parity**, e.g. expected `…Owner`1+Payload[[System.Int32]]…` vs actual `…Payload[[…]]…`. The assertion is anchored to reified type identity, not to a coarse constant that both sides share. **ADR-0154 satisfied**: this test discriminates the reified type from `object`.

## 5. Mutation — 6 PRODUCT mutants, one per hunk, all RED, zero no-results

Each applied via `git diff` **and** confirmed by `GSharp.Core.dll` md5 movement off pristine `0ccf620d`; all `BUILD_RC=0`; restored by exact md5 round-trip (source `9af6d1e0`, dll `0ccf620d` — both matched).

| id | hunk | mutation | Core.dll md5 | result |
|---|---|---|---|---|
| M1 | `:1449` | drop enclosing prefix → `return type.TypeParameters;` | `cc4fa4f7` | **RED 5/16** — exactly the 5 new cases |
| M2 | `:1449` | reverse flatten order (own before enclosing) | `24a61326` | **RED 1/16** — `generic-owner-nested-generic` |
| M3 | `:1348` | **invert** `if (!reifiedParameters.IsDefaultOrEmpty)` | `7a1980b3` | **RED 16/16** (control) |
| M4 | `:1421` | **invert** `if (reifiedParameters.IsDefaultOrEmpty)` | `5d6c8ca4` | **RED 16/16** (control) |
| M5 | `:1430` | **invert** `i < structType.EnclosingTypeArguments.Length` | `3ad53d2b` | **RED 5/16** |
| M6 | `:1440` | **invert** `i < structType.TypeArguments.Length` | `0d55a04e` | **RED 16/16** (control) |

M3/M4/M6 are the controls I expected to kill broadly, and they did — so M1's and M5's narrow 5/16 and M2's 1/16 are meaningful, not noise.

**M2 is the sharp one:** flatten *ordering* is pinned by exactly **one** specimen, `generic-owner-nested-generic` — the only case with both enclosing and own parameters. `deep-nested` cannot detect a reversal because `Payload`'s own arity is 0, making the reversal a no-op there. The coverage is adequate but it is one specimen deep; worth knowing before anyone prunes the matrix.

---

## SHOULD-FIX 1 — the new fallback arm silently yields `object`, and no test exercises it

`Evaluator.Expressions.cs:1428-1434`
```csharp
for (var i = 0; i < definitionEnclosingParameters.Length; i++)
{
    var argument = i < structType.EnclosingTypeArguments.Length
        ? structType.EnclosingTypeArguments[i]
        : definitionEnclosingParameters[i];      // <-- raw TypeParameterSymbol
    typeArguments[i] = ResolveClrBackingTypeArgument(argument);   // <-- no dictionary
}
```

When `EnclosingTypeArguments` is short, a raw `TypeParameterSymbol` is passed to `ResolveClrBackingTypeArgument` **without** `reifiedTypeParameters`, so it falls through to `typeof(object)` at the tail. `StructSymbol.cs:1397` documents this state as real: *"May be empty when the enclosing type is open but the nested type carries own arguments."*

**Reachable, and measured.** `Owner[T].Payload[string]` referenced from **inside** `Owner[T]`:

| driver | result |
|---|---|
| `gsc /out:` (**oracle**) | `A:System.Int32 \| B:System.String` |
| bare `gsc` (file) | `A:System.Int32 \| B:System.String` |
| `gsi` (file) | `A:System.Int32 \| B:System.String` |
| **interactive** | **`A:System.Object`** \| `B:System.String` |

`rc=0`, `HASERROR=False`, no diagnostic — the silent-wrong-answer shape ADR-0156 exists to eliminate. `B` is correct because it comes from the *own*-parameter loop where `TypeArguments` is populated; only the enclosing arm degrades, which is precisely the signature of this branch.

**Not a regression** — base behaves identically, so this is residual scope, exactly as #3180 was residual scope after #3151.

**Untested.** All 16 matrix cases substitute `__TYPE__` into `var single = Box[__TYPE__]()` at top level (`:158`), so `EnclosingTypeArguments` is always fully populated and the `:` arm **never executes** in the suite. Consistent with the mutation data: M5 inverts exactly this condition and kills 5 tests, which proves the `?` arm is covered — nothing proves the `:` arm is.

This is the **third** instance of one defect class: a `ResolveClrBackingTypeArgument` call site that does not thread the dictionary (#3137 round-1 BLOCKER at the nullable branch, #3180 itself, now this).

**Recommendation:** add an open-enclosing specimen and either fix or explicitly document the arm. I am **not** claiming the remedy is a one-liner — I did not measure one, and reading `:1348-1356` suggests threading the dictionary alone would yield a `GenericTypeParameterBuilder` rather than a concrete runtime type, so the ambient enclosing instantiation is needed. Given #3180's own history, a follow-up issue with this repro attached is the honest disposition; blocking this PR on it would repeat the mistake of asking one PR to close an open-ended axis.

## NIT 1 — the `= null` default is what keeps regenerating this defect class

`:1453-1455` — `ResolveClrBackingTypeArgument(TypeSymbol argument, IReadOnlyDictionary<…> reifiedTypeParameters = null)`. An omitted dictionary **compiles silently** and degrades to `typeof(object)`. SHOULD-FIX 1 is a live instance introduced by this very diff. Making the parameter required, and passing explicitly at the one genuinely parameter-free site, converts the next omission from a silent wrong answer into a compile error.

## NIT 2 — three of the four test columns can no longer observe this change (measured)

Post-ADR-0156, `compiledFile` and `interpretedFile` run the **same emit engine** as `expected`, so `:112` and `:113` cannot see an evaluator-only change. Measured across 6 mutants plus the two-sided base run: **every kill came from the interactive column** — `:111` (`Assert.Equal(expected, evaluated)`), `:74` (`RunInteractive`) and `:198` (`Assert.False(cell.HasError)`). **Zero kills at `:112` or `:113`.**

They are not worthless — they pin the Phase-1 host plumbing (entry point, `Success.` protocol, stdout). But the effective cross-engine oracle here is **one** column, not three, and the test name should not be read as implying otherwise.

## NIT 3 — this test stops discriminating when Phase 2 flips the default

`RunInteractive` (`:195`) hardcodes `new SessionEngine()`. When ADR-0156 flips the interactive default to `EmittedSessionEngine`, all four columns collapse onto the emit engine and this test silently stops pinning anything, while staying green. ADR-0156 Phase 3 plans to delete the evaluator outright, so this may resolve itself — but until then it is a green test that can quietly stop measuring.

---

## Verification denominator

**9 of 9 test projects**, enumerated from `dotnet sln GSharp.sln list` — the same enumeration CI uses at `build/generate-ci-test-matrix.py:26`. Unfiltered, no `--no-build`, no `GITHUB_ACTIONS`. The new cases run in the normal `Interpreter.Tests` invocation with no filter or special ordering.

| project | result |
|---|---|
| Interpreter.Tests | 1398 / 1398 |
| Core.Tests | 7128 passed, 1 skipped, 7129 |
| Compiler.Tests | 4260 / 4260 |
| Cs2Gs.Tests | 1907 / 1907 |
| LanguageServer.Tests | 462 / 462 |
| Extensions.Tests | 104 / 104 |
| Sdk.Tests | 59 / 59 |
| GSharp.GeneratorHost.Tests | 31 / 31 |
| InternalAnalyzers.Tests | 9 / 9 |

**Total 15,359 executed — 15,358 passed, 1 skipped, 0 failed.** `Sdk.Tests` is in bounds and was run. `Cs2Gs.Tests` (real out-of-process `gsc.dll`) is green after the solution-level build.

**ILVerify gate** (#3164, `9003a86c` — confirmed present in this branch's ancestry, so no `update-branch` needed): `build/run-ilverify.sh` → **123/123 verified, 2 documented suppressions, 0 verification failures, 0 compile failures**, `rc=0`. Expected: the diff touches the evaluator only, and emitted output was byte-identical between trees on every probe.

## Self-disclosures

1. **I nearly filed a false NIT about the positive control.** Several new specimens print a field shape identical to `ControlShape`, so I began writing up that a specimen could satisfy `Assert.Contains(ControlShape, …)` in place of the control. Reading `:105-106` and the template at `:174-176` refuted it — `ControlShape` is anchored by the `"44\n"` marker line, emitted only immediately before the control block. Withdrawn before posting.
2. **I checked a coincidence instead of trusting it.** Base and PR `GSharp.Core.dll` had *identical byte sizes*, which looked like a shared or stale artifact. md5 and inode both showed genuinely distinct binaries. Size equality was coincidence; had I asserted staleness from it I would have invalidated my own two-sided setup.
3. **Interactive could not be driven the way the issue implies.** `gsi` refuses piped stdin (`gsi requires an interactive terminal`), so the fourth column is unreachable from the shell. I built a small out-of-tree harness calling `SessionEngine.Evaluate` in-process — the same entry point the PR's test uses — rather than reporting "cannot reproduce" from a driver that structurally cannot run the shape.

PR #3192, opened 2026-08-04
Render ByRefLike interpolation on emitted paths

### LINKED ISSUE #3183: [bug] Span interpolation ICEs the emit path (GS9998 MakeGenericMethod ByRefLike) while the evaluator renders it
## Repro

```gs
import System

func Main() {
    var values = []int32{11, 22, 33}
    var writable Span[int32] = values
    var readOnly ReadOnlySpan[int32] = writable
    Console.WriteLine("writable=${writable}")
    Console.WriteLine("readonly=${readOnly}")
}
```

`gsc /out:span.dll span.gs`:

```
span.gs(1,1,1,1): error GS9998: ArgumentException: GenericArguments[0], 'System.Span`1[System.Int32]', on 'Void AppendFormatted[T](T)' violates the constraint of type 'T'.
```

`InterpolatedStringHandlerLowerer.CloseAppendFormatted` calls `MakeGenericMethod(typeof(Span<int>))` on `DefaultInterpolatedStringHandler.AppendFormatted[T](T)`, whose `T` rejects ByRefLike type arguments — the CLR throws `ArgumentException` (inner `VerificationException`) at lowering time, surfaced as a GS9998 ICE.

## Context

- The tree-walking evaluator supports span interpolation since #3128 (renders `System.Span<Int32>[3]`-style names); the emit path never has. Until ADR-0156 (#3177/#3176) Phase 1 this was invisible on `gsi <file>` and bare `gsc` because they interpreted; now all file drivers use the emitter, so the gap is user-visible on every driver (interactive REPL still interprets and still works).
- Discovered by `Issue3114ReadOnlySpanInterpreterTests.SpanInterpolation_UsesPublicSpanName` failing after the Phase 1 driver swap (PR for #3176). Per that PR's policy the emit bug is not fixed there; the test now pins (a) the interactive evaluator's span-interpolation output via SessionEngine and (b) the GS9998 ICE protocol on the file drivers, referencing this issue — flip (b) to golden output when this is fixed.
- Note C# rejects `$"{span}"` for ByRefLike values at compile time (CS0306-family). The fix here is either a real diagnostic (reject span interpolation holes with a GS error) or emit support via `AppendFormatted(ReadOnlySpan<char>)`/`ToString`-style routing for the evaluator's public-name rendering. Either way the current GS9998 ICE is wrong.

## Expected

No internal compiler error: either span interpolation emits (matching the evaluator's `System.Span<Int32>[3]` rendering) or the compiler reports a proper source-anchored diagnostic.

DavidObando:
Closes #3183

## Fix

Adopted round-two alternative: emitted interpolation routes CLR ByRefLike holes through their existing parameterless `ToString()` method before the normal string-handler path. This preserves capability instead of replacing GS9998 with a refusal.

- `Span[T]` and `ReadOnlySpan[T]` now render on bare `gsc`, `gsc /out:`, and `gsi <file>`.
- `Span[char]`, `ReadOnlySpan[char]`, and alignment render correctly (`hello`, `   hello`).
- Ordinary interpolation and user-defined handlers remain unchanged.
- GS0519 survives only as a source-anchored fallback for a ByRefLike symbol with no available CLR `ToString()` method; current covered case is a user-declared `ref struct` without a reflection `ClrType` during lowering.
- Removed ADR-0156's evaluator/emitter divergence note. Added the requested comment explaining why the widened post-lowering diagnostic gate cannot double-report binder errors.

## Corrected regression attribution

#3182 (`3d5ba7af`) moved bare `gsc` and `gsi <file>` to emit-to-memory and exposed the emit ICE. #3177 (`ec50744d`) adopted ADR-0156 and carried the feasibility spike; it was not the regressing swap. Pre-regression commit: `aa453ac6`.

## Measured final drivers

Measured after rebase onto `57256713`:

| Driver | rc | stdout |
|---|---:|---|
| bare `gsc` | 0 | both integer spans render public type names; both char spans render `hello`; alignment renders `   hello`; then `Success.` |
| `gsc /out:` + CLR run | 0 / 0 | same values, without `Success.` |
| `gsi <file>` | 0 | same values, without `Success.` |

All stderr streams were empty.

## Evidence

- Final tests alone on unfixed current main: **6 failed / 17 passed**. Exact failures: `SpanInterpolation_RendersAcrossFileDrivers` × 3 and `UserRefStructInterpolation_ReportsFallbackDiagnosticAcrossFileDrivers` × 3.
- Focused Issue3114/3183 tests: **23/23**.
- Nine projects, denominator from `dotnet sln GSharp.sln list` per `build/generate-ci-test-matrix.py`: **15,377 passed / 0 failed / 1 skipped**.
  - Compiler 4,260; Core 7,128/1 skipped; Interpreter 1,417; Cs2Gs 1,907; LSP 462; Extensions 104; SDK 59; GeneratorHost 31; InternalAnalyzers 9.
- Release solution build after final rebase: 0 warnings, 0 errors.
- `git merge-tree --write-tree origin/main HEAD`: clean tree `0a36eba47211ef4364ec520bc57388c4526c2fce`.
- GS0519 allocator check: absent from current main and from every other open PR patch.

### Final-tree product mutants

Baseline `GSharp.Core.dll` md5: `076694ed03bf6b01201decf5003d983f`. Each mutant had a non-empty `git diff`, moved the DLL md5, was killed, then restored to the exact baseline md5.

- invert fallback guard (`== null` → `!= null`): **6 failures**;
- invert ToString routing (`!= null` → `== null`): **6 failures**;
- invert lowerer-diagnostic gate (`IsError` → `!IsError`): **3 failures**;
- control invert literal branch: **9 failures**.

Round-one's `return node;` early exit became redundant after narrowing the fallback to user symbols: deleting it keeps 23/23 green, so it was removed rather than preserving dead control flow.

## Near-misses

- First round-two RED command used the wrong project filename and produced MSB1009. Scored no-result; corrected path produced the expected failures.
- Main advanced after the first nine-project run. Rebased again onto `57256713`, then repeated RED/GREEN, all mutants, driver probes, build, and all nine projects.

Local e2e and local ILVerify were not run. E2e is prohibited locally; required CI gates are pending.


DavidObando:
# Review: PR #3192 — `fix/3183-work` (issue #3183)

**Verdict: NOT MERGEABLE** — 1 BLOCKER, 3 SHOULD-FIX, 1 NIT.

The ICE is genuinely fixed and the diagnostic is genuinely source-anchored. The blocker is not the code; it is that the PR ships a **documented capability refusal** whose two stated justifications are both measurably false, when a **19-line change preserves the capability on all three drivers**.

Measured on head `8b3c6966`, base `81e6deab`, live `main` `16f97229`. Merge-base `81e6deab`; main is +1 commit (test-only, 2 lines).

---

## First, a correction to the issue's own attribution

Both the issue and the PR attribute the driver swap to **#3177**. It was **#3182** (`3d5ba7af`, "Execute gsi scripts and bare gsc through emit-to-memory (ADR-0156 Phase 1)"). `#3177` (`ec50744d`) is "Adopt emit-to-memory execution for gsi ... with feasibility spike". The pre-regression commit is therefore `aa453ac6` (= `3d5ba7af^`), and `#3128` (`22ee0352`, span interpretation) is an ancestor of it — so spans were renderable *and* file drivers still interpreted at that point. That is the tree I measured against.

## The three-tree measurement that decides (a) vs (b)

Probe = the issue's own repro. Each driver in its own asserted-empty directory; `rc` captured before any pipe.

| tree | bare `gsc` | `gsc /out:` | `gsi <file>` |
|---|---|---|---|
| `aa453ac6` (pre-#3182) | **rc0 — `writable=System.Span<Int32>[3]`** | rc1 GS9998 `(1,1,1,1)` | **rc0 — `writable=System.Span<Int32>[3]`** |
| `81e6deab` (PR base) | rc1 GS9998 `(1,1,1,1)` | rc1 GS9998 `(1,1,1,1)` | rc1 GS9998 `(1,1,1,1)` |
| `8b3c6966` (PR head) | rc1 GS0519 `(8,23,8,45)` | rc1 GS0519 `(8,23,8,45)` | rc1 GS0519 `(8,23,8,45)` |

**This is shape (b).** Two of three drivers rendered span interpolation correctly in the last released state and now refuse.

**In fairness, this PR is not the regressor** — at its own base all three drivers already ICE'd; #3182 caused the loss. But this PR closes #3183, the only tracking item, and adds `docs/diagnostics.md`, `website/docs/ref/diagnostics.md` and an ADR-0156 note that frame the loss as designed. After merge there is no open item to restore it.

**Question 1 (does it work anywhere?):** only the interactive REPL. Verified independently under a pty — 15 `System.Span` renderings, **0** `GS0519`, positive control 55 `writable`. That is the one driver ADR-0156 Phase 2 is slated to remove, and the ADR hunk says so explicitly ("until Phase 2 removes the interactive evaluator").

---

## BLOCKER — the refusal is over-broad and its stated rationale is factually incorrect

`src/Core/CodeAnalysis/Lowering/InterpolatedStringHandlerLowerer.cs:150`

The PR body gives two reasons for choosing refusal. Both are false, and each is independently checkable.

**(1) "This matches C#'s ByRefLike generic-argument rejection."** C# does **not** reject these. Compiled and ran on net9.0:

```
Void AppendFormatted(System.ReadOnlySpan`1[System.Char])
Void AppendFormatted(System.ReadOnlySpan`1[System.Char], Int32, System.String)
C# $"{ReadOnlySpan<char>}" => [hello]
C# $"{Span<char>}"         => [ab]
C# alignment/format        => [   hello]
```

`DefaultInterpolatedStringHandler` carries **non-generic** `AppendFormatted(ReadOnlySpan<char>)` overloads; `Span<char>` converts implicitly. `FindAppendFormatted` (`:696`) filters with `!method.IsGenericMethodDefinition → continue`, so the lowerer never considers them. Measured consequence on head — a program C# accepts:

```
roschar.gs(7,23,7,39): error GS0519: By-ref-like value of type 'System.ReadOnlySpan[char]'
cannot be used in an interpolated string ...
```

**(2) "avoids inventing a second span stringification path."** No second path needs inventing — it already exists and already emits. On **this PR's own head**, all three drivers:

```
viaToString=System.Span<Int32>[3]      rc=0
```

from `"viaToString=${writable.ToString()}"`. That is **byte-identical** to the string the pre-#3182 evaluator produced. The diagnostic's own advice ("convert it to a string before interpolation") works — which is precisely the proof that the compiler could do it for the user.

### The 19-line alternative — implemented and measured, not asserted

Route a ByRefLike hole through the `ToString()` that already works, and keep GS0519 only for ByRefLike types that have no CLR `ToString` to route to (so the diagnostic is **narrowed, not deleted**):

```csharp
// guard, :150
if (part.IsHole && TypeSymbol.IsByRefLike(part.Value.Type) && FindByRefLikeToString(part.Value.Type) == null)

// at the hole, :202
var byRefLikeToString = TypeSymbol.IsByRefLike(value.Type) ? FindByRefLikeToString(value.Type) : null;
if (byRefLikeToString != null)
{
    value = new BoundImportedInstanceCallExpression(
        node.Syntax, value, byRefLikeToString, TypeSymbol.String, ImmutableArray<BoundExpression>.Empty);
}

private static MethodInfo FindByRefLikeToString(TypeSymbol type)
=> type?.ClrType?.GetMethod("ToString", System.Type.EmptyTypes);
```

Measured (`BUILD_RC=0`, **0 warnings**, `GSharp.Core.dll` md5 confirmed moved):

| probe | bare `gsc` | `gsc /out:` | `gsi <file>` |
|---|---|---|---|
| `Span[int32]` / `ReadOnlySpan[int32]` | rc0 `System.Span<Int32>[3]` | rc0 same | rc0 same |
| `ReadOnlySpan[char]` | rc0 `chars=hello` | rc0 same | rc0 same |
| ordinary + alignment + `:X` | rc0 unchanged | rc0 unchanged | rc0 unchanged |
| struct / nested / span-not-interpolated | rc0 unchanged | rc0 unchanged | rc0 unchanged |

- Output is **byte-identical to the pre-#3182 rendering**, and `gsc /out:` renders span interpolation **for the first time in the project's history** — strictly better than every prior tree.
- `Core.Tests` **7128 passed / 0 failed / 1 skipped**.
- `Interpreter.Tests` **1394 / 1397** — the only 3 failures are `SpanInterpolation_ReportsEmitDiagnosticOnFileDrivers`, i.e. exactly the tests that assert the refusal. **That is a design disagreement, not a defect, and I am reporting it as such.**
- Size: **+19 / −1 in 1 file**, versus the PR's **+176 / −39 across 8 files**, no new diagnostic ID, no two doc pages, no ADR divergence note.

Patch: `/tmp/g3192/ponytail_alternative.patch`.

**Either resolution clears this blocker:**
1. Adopt the `ToString()` routing (above), or
2. Keep the refusal, but (a) delete the "matches C#" claim from the PR body and `docs/diagnostics.md`, (b) narrow the guard so `ReadOnlySpan[char]`/`Span[char]` are **not** rejected, and (c) open a follow-up issue to restore emit rendering before ADR-0156 Phase 2 removes the last driver that still works.

I am not asking for my patch specifically — option 2 is entirely acceptable. What is not acceptable is shipping the refusal *with* the false justification and *without* a tracking item.

---

## SHOULD-FIX

**S1 — closing #3183 retires the only record of a live capability regression.**
`docs/adr/0156-gsi-emit-to-memory-execution.md:322` calls this a "deliberate temporary divergence", but nothing tracks the un-divergence. Please open the follow-up before merge.

**S2 — the one test proving the feature ever worked is scheduled for deletion.**
`test/Interpreter.Tests/Issue3114ReadOnlySpanInterpreterTests.cs:150` (`SpanInterpolation_UsesPublicSpanNameInteractively`) pins the working rendering via `SessionEngine`. The ADR says Phase 2 removes the evaluator; at that point the natural edit is to delete this test, erasing the last evidence. Add a comment pointing at the follow-up issue so Phase 2 converts it rather than deletes it.

**S3 — GS0519 allocation should be confirmed by the ID allocator.**
Free at base and on live `main` (`git grep GS0519` = 0 both). Registered in both doc pages. But with a high-volume merge queue I could not rule out a concurrent claim in another open PR; please confirm.

## NIT

**N1 — the widened gate at `Compilation.cs:763` is safe only by distant coupling.**
`program.Diagnostics.Where(d => d.IsError)` looks like it could double-report binder errors or newly fatalise diagnostics from other lowering passes. **It cannot** — both emit paths already return early on any binder error (`:461-465` and `:589-593`), so only errors added by the intervening lowering passes can reach it. I verified this on both paths; suspicion cleared. Worth one comment, since the safety property lives ~300 lines away.

---

## What is solid (stated plainly)

- **Source anchoring is genuinely fixed**: `(1,1,1,1)` → `(8,23,8,45)`, pointing at the hole expression, on all three drivers. The issue's specific complaint is resolved.
- **Refusal is narrow w.r.t. ordinary types.** Probed `int`, `string`, alignment `${n,6}`, format `${n:X}`, custom struct, nested interpolation, and span used outside interpolation — all rc0 and unchanged on head.
- **Test oracle strength is good, and the recurring failure shape is absent.** `SpanInterpolation_ReportsEmitDiagnosticOnFileDrivers` uses `Assert.Collection` over exactly two diagnostics, each checked for ID, file, **line number** and type name, plus exit code and exact stream routing — not an absence-of-diagnostic test and not a coarse constant. ADR-0154 satisfied.
- **Mutation — I audited the sub-hunk the PR body did not.** The body scores the ByRefLike detection and the error propagation, but not the `return node;` early-exit (`:158-161`). Removing it (product mutant; first attempt failed to build with `CS0219`, so it was **retargeted, not scored** — a no-result is not a survival; retargeted mutant built `rc=0` with `GSharp.Core.dll` md5 confirmed moved) **killed 3 tests**: `SpanInterpolation_ReportsEmitDiagnosticOnFileDrivers` × 3 drivers. Pinned. Source and binary restored to exact md5.
- **`ilverify` present and green** (4m4s). 34 checks, 0 failing, 0 pending.
- **Merged tree built, not just merge-tree'd.** `git merge-tree` output = one tree SHA `ac8547ff`, 0 conflict markers; built that tree: `BUILD_RC=0`, **0 warnings**, `Interpreter.Tests` 1397/1397.

## Coverage disclosure

Denominator **9**, from `dotnet sln GSharp.sln list`.

**Run by me (2 of 9):** `Core.Tests` 7128/0/1-skip and `Interpreter.Tests` 1397/0 — both on pristine head, both reproducing the author's numbers exactly; plus the same two on the alternative and on the mutant, the merged-tree build, and a 5-probe × 3-driver differential across three trees.

**NOT RUN (7 of 9):** `Compiler.Tests`, `Cs2Gs.Tests`, `LanguageServer.Tests`, `Extensions.Tests`, `Sdk.Tests`, `InternalAnalyzers.Tests`, `GSharp.GeneratorHost.Tests`. `Compiler.Tests` (~1h18m) is the one I most wanted given this is a lowering change; CI covers it and is green.

## Self-disclosed harness errors

- My first ancestry check printed "HEAD DIVERGED FROM BRIEF" — **that line was meaningless**: the PR ref fetch had failed because my scratch clone's `origin` is the *local* checkout, which carries no `pull/*` refs. Re-fetched from GitHub directly; head is `8b3c6966`, matching the brief.
- My first `Core.Tests` run was on the **alternative**, not pristine head. I re-ran on pristine head rather than quoting the alternative's number.
- Mutant M1 first attempt did not build (`CS0219`) and `GSharp.Core.dll` md5 did not move; scored as a no-result and retargeted.

DavidObando:
Round 2 addressed at `4f1789ee`: adopted CLR `ToString()` routing; all three file drivers now render spans at rc0, including char spans and alignment. Corrected regressor to #3182 (`3d5ba7af`, pre-regression `aa453ac6`). GS0519 remains only for ByRefLike symbols with no available CLR `ToString()` reflection method and is pinned with a user `ref struct` test. Rebased onto `57256713`; 15,377/0/1 locally, all CI checks green including ilverify/e2e/Oahu.

DavidObando:
## Round-2 review — **MERGEABLE**

No BLOCKER. Two SHOULD-FIX, one NIT. Reviewed head `4f1789ee` (derived via `gh pr view`, not the briefed SHA), merge-base `57256713`, `main` `3ca776d3` — head is **1 behind / 1 ahead**, `mergeStateStatus` CLEAN.

### 1. It renders, and it matches two independent oracles

Fresh worktree at `4f1789ee`, Release, 0 warnings, 38 projects. Three file drivers, `rc` captured before any pipe:

| driver | rc | stdout md5 |
|---|---|---|
| bare `gsc` (emit-to-memory) | 0 | `9ce59672` (adds `Success.`) |
| `gsc /out:` + run | 0 / 0 | `18a0515f` |
| `gsi <file>` | 0 | `18a0515f` |

Rendered values, the three asked for:

```
chars=hello            # ReadOnlySpan[char]
writableChars=hello    # Span[char]
aligned=   hello       # ${ros,8} -> 3 leading spaces, right-aligned
```

**Oracle A — real C#.** Built a `net10.0` console probe. C# prints `hello` / `hello` / `   hello`; G# head is **byte-identical**. Extended to the shapes most likely to be lost by a `ToString()` reroute — `{ros:X}`, `{ros,8:X}`, `{ros,-8}` — C# gives `hello` / `   hello` / `hello   `, G# head **identical on all three**. Alignment and format survive the rewrite.

**Oracle B — pre-regression `aa453ac6`.** `cmp` rc=0 on every pair: head bare `gsc` ≡ pre bare `gsc`; head `gsc /out:` ≡ pre `gsi`; head `gsi` ≡ pre `gsi`. Ordinary interpolation (`${n} ${s} ${f} ${n,6} ${n:X}`) also byte-identical head vs pre (`cafbcc36`) — the hot path is untouched.

Two *different* oracles agreeing, not one repeated.

### 2. Correction to the round-1 framing — the emit path was **never** working

The brief (and round 1) had `aa453ac6` "rendering at rc0". That is true only for the two evaluator-backed drivers. Measured on `aa453ac6`:

- bare `gsc` → rc 0, renders
- `gsi` → rc 0, renders
- **`gsc /out:` → rc 1, `GS9998: ArgumentException … AppendFormatted[T](T) violates the constraint of type 'T'`**

So `gsc /out:` ICE'd *before* #3182. #3182 removed the evaluator masking on the other two file drivers; it did not create the hole. This PR fixes a pre-existing emit defect, which is a stronger claim than "regression repair".

Worse, for a **user-declared `ref struct`** `aa453ac6`'s `gsc /out:` reported rc 0 and `Wrote ptok.exe` — and the emitted assembly aborts at runtime:

```
rc=134  System.Security.VerificationException: … type argument 'Default.Token'
        violates the constraint of type parameter 'T'.
```

Silent emission of unverifiable IL. GS0519 converts that into a compile-time, source-anchored error. Strict improvement.

### 3. GS0519 is narrowed, not deleted, and has a live trigger

`src/Core/CodeAnalysis/DiagnosticDescriptors.cs:387`. Fires on all three drivers, source-anchored to the hole:

```
refstruct.gs(7,23,7,39): error GS0519: By-ref-like value of type 'Token' …
```

Not dead code. The predicate is `TypeSymbol.IsByRefLike(t) && FindByRefLikeToString(t) == null` — and per `TypeSymbol.cs:246-255` `IsByRefLike` has two arms, a source `StructSymbol{IsRefStruct:true}` (whose `ClrType` is null at lowering) and an imported CLR type. Since `FindByRefLikeToString` reads `type?.ClrType?…`, it can only return null on the source arm; every imported ref struct inherits `ValueType.ToString`. **GS0519 ⟺ user-declared `ref struct`.**

**GS9998 is genuinely gone, not reclassified.** Under mutant M1 the original `GS9998 … AppendFormatted[T](T)` returns verbatim; on the unmutated head it is absent and the program renders.

### 4. Shape — "adopted" is substantially true

Product is **+45 / −3 across 4 files** (docs +22, tests +158/−39):

| file | Δ | role |
|---|---|---|
| `Lowering/InterpolatedStringHandlerLowerer.cs` | +29/−2 | the reroute (~10 lines) + GS0519 guard + diagnostic plumbing |
| `Compilation/Compilation.cs` | +5/−1 | lets lowerer-added errors reach the emit gate |
| `DiagnosticBag.Reports.Emit.cs` | +10 | report method, 7 of 10 lines docstring |
| `DiagnosticDescriptors.cs` | +1 | registration |

Round 1 was +176/−39 across 8 files; the scaffolding is gone. What exceeds the `+19/−1` alternative is exactly the *narrowed* GS0519 that round 1 endorsed keeping. I could not find a smaller shape that also converts the user-`ref struct` case from unverifiable IL into a diagnostic.

The `Compilation.cs:763` comment — *"Binder errors return before lowering; only lowerer-added errors can reach this gate"* — is **verified**, not assumed: both callers (`:461`, `:590`) `return new EmitResult(success: false, …)` when `syntaxDiagnostics.Concat(program.Diagnostics)` has any error, before any lowering pass runs.

### 5. Mutants — two-sided, applied, DLL-moved, restored

Baselines: `InterpolatedStringHandlerLowerer.cs` `b5721343`, `Compilation.cs` `25a8afb9`, `GSharp.Core.dll` `cefdaa35`. Every mutant asserted applied via `git diff --numstat` **and** `GSharp.Core.dll` md5 movement; harness refuses to score an unmoved DLL.

| # | site | mutation | Core.dll | result |
|---|---|---|---|---|
| M1 | `InterpolatedStringHandlerLowerer.cs:198` | `IsByRefLike(value.Type)` → `!IsByRefLike(value.Type)` | `→ 8324e82a` | **RED** — spans revert to the exact `GS9998` of #3183; ref struct still GS0519 |
| M2 | `:151` | `FindByRefLikeToString(…) == null` → `!= null` | `→ 125d1119` | **RED both sides** — spans wrongly rejected with GS0519; ref struct wrongly accepted → `VerificationException` rc 134 |
| M3 | `Compilation/Compilation.cs:764` | `.Where(d => d.IsError)` → `.Where(d => !d.IsError)` | `→ dea934be` | **RED and isolated** — GS0519 stops reaching the gate → unverifiable IL rc 134, while spans still render |

M3 is the control that matters: it kills the diagnostic-plumbing property *without* touching the rendering property, so M1/M2 results are not a single confounded signal. The PR's own tests also kill M3 (3 failed / 20 passed, `--filter ~Issue3114`).

Restored by exact md5 round-trip: both sources back to baseline, and after rebuild `GSharp.Core.dll` back to `cefdaa35` bit-for-bit.

---

## SHOULD-FIX 1 — GS0519 is unconditional for user `ref struct`s, and its message misdirects

`src/Core/CodeAnalysis/Lowering/InterpolatedStringHandlerLowerer.cs:697-698`

```csharp
private static MethodInfo FindByRefLikeToString(TypeSymbol type)
    => type?.ClrType?.GetMethod("ToString", System.Type.EmptyTypes);
```

Only a **CLR** `ToString` is ever found. A G#-declared `ToString` cannot be, in either sanctioned spelling. Measured on this head:

- receiver-clause form `func (t Token) ToString() string` → rc 1, GS0519
- **canonical in-body form** (ADR-0079) `ref struct Token { func ToString() string {…} }` → rc 1, GS0519
- `${token.ToString()}` → **rc 0, compiles, emits, runs, prints `interp=Token#42`**

So the compiler refuses the interpolation while demonstrably able to perform it — the same tell that killed round 1, in a much narrower region. The message *"has no available CLR ToString method"* is literally true and operationally misleading: a user who reads it and adds a `ToString()` gets the identical error.

Not a blocker — the prior behaviour on this input was unverifiable IL crashing at runtime, so this is still a strict improvement. Fix either way:
- **preferred**: also resolve the `StructSymbol`'s own `ToString` member and emit a `BoundCallExpression`, which closes the case rather than diagnosing it; or
- **minimum**: reword to say a by-ref-like type declared in G# cannot be interpolated directly, and name `${x.ToString()}` as the workaround — the PR has already proved that workaround emits valid IL.

## SHOULD-FIX 2 — raw `GetMethod` where the house helper exists

Same line `:698`. `ClrTypeUtilities.GetMethodSafe` (`src/Core/CodeAnalysis/Symbols/ClrTypeUtilities.cs:45`, 4 existing call sites) exists specifically to *"resolve a named method without asking a constructed `TypeBuilder` generic instantiation to resolve members directly"*, catching the `NotSupportedException` that raw `GetMethod` throws in that case. That exception class is exactly what surfaces as GS9998 — the ICE this PR is closing. The adjacent `ClrTypeUtilities.IsByRefLike` carries the analogous #1919 metadata-load guard; the new helper carries neither.

**Disclosure: I could not reach it.** Two constructions of a ByRefLike over a source-declared type were rejected earlier in the pipeline — `Span[Item]` over a G# struct gives `GS0155`, `Span[Color]` over a G# enum gives `GS0005`/`GS0125`. So this is **unreached, not unreachable**; I am reporting it on convention and blast-radius grounds, not as a live defect. `GetMethodSafe` takes no parameter-type filter, so it is not a literal drop-in — a try/catch around the existing call is the smaller change.

## NIT — G# accepts non-char span interpolation that real C# rejects

Measured: C# rejects `$"{Span<int>}"` with `CS9244` ("may not be a ref struct … as parameter 'T'"). G# head accepts it and prints `writable=System.Span<Int32>[3]`. This is **not** introduced here — it is the #3128 public-span-name behaviour, and head is byte-identical to the pre-regression evaluator. Recording it so the divergence is deliberate on the record rather than discovered later; no change requested.

---

### Denominator

**1 of 9 run locally**: `Interpreter.Tests` **1417 passed / 0 failed / 0 skipped, rc 0** (the project owning the changed test file). Denominator source `dotnet sln GSharp.sln list`, CI's own per `build/generate-ci-test-matrix.py:26`.

`Compiler.Tests` **not run locally** — stating that plainly rather than leaving it implied. The `Compilation.cs` widening is the one hunk with reach beyond this feature, and it is covered by CI on this exact SHA: `gh api …/commits/4f1789ee/check-runs` returns **27 success / 3 skipped, 0 failure**, including all 8 compiler shards, `ilverify` (3m41s), `e2e`, and `cs2gs-oahu`. `ilverify` is present and passing, so no `gh pr update-branch` is needed.

PR #3193, opened 2026-08-04
Fix evaluator ref write-back for remaining call shapes

### LINKED ISSUE #3162: [bug] out-parameter write-back still dropped for explicit-interface base dispatch and constrained-static generic dispatch
## Summary

`out`/`ref` parameter write-back is still silently dropped by the **evaluator** and **interpreter** for two call shapes that PR #3156 did not cover:

1. **Explicit-interface base dispatch** — `base[IFiller].Fill(&x)`
2. **Constrained static generic dispatch** — `T.Fill(&x)` where `T` is constrained to an interface declaring a static method

This is the **same defect as #3140** — same file, same helper, same silent-wrong-answer signature — in two cells the #3156 matrix did not enumerate. It is **pre-existing, not a regression**: measured identical on the base commit and on merged `#3156`.

## Measured behaviour

`gsc /out:` (emit) is the oracle. Both cells disagree with it:

| shape | evaluator site | emit | bare `gsc` (evaluate) | `gsi` (interpret) |
|---|---|---|---|---|
| `base[IFiller].Fill(&x)` | `Evaluator.Calls.cs:374` | **61** | 101 | 101 |
| `T.Fill(&x)` constrained static | `Evaluator.Calls.cs:475` | **81** | 101 | 101 |

`101` is the *unmodified* input, i.e. the callee's write never reaches the caller. No diagnostic is produced in any driver — the program runs and prints a wrong number.

## Why #3156 did not catch it

#3156's parity matrix covers `Free`, `ClassShared`, `StructShared`, `ClassInstance`, `BaseClass`, `StructInstance`, `Interface`, `GenericMethod`. Two adjacent-looking rows create a false sense of coverage:

- `Interface` calls **through** an interface-typed receiver — not `base[IFoo].M`.
- `BaseClass` covers `base.M` — not the explicit-interface form `base[IFoo].M`.

So the enumeration looks complete and is not. This is the eighth instance today of the recurring shape: *the test name states a semantic property; the assertion checks a coarser observable that is constant across both sides of that property.* Here it is one level up — the **matrix axis** is coarser than the language surface it claims to cover.

## Fix guidance

`Evaluator.Calls.cs:374` is expected to be a near-drop-in reuse of the existing `BuildRefSlots` / `WriteBackRefSlots` pair that #3156 already routed the other shapes through — do **not** add a parallel write-back path.

**`:475` is not a drop-in.** That site pre-evaluates its arguments in order to sniff types for constrained-static dispatch, so `refSlots` must be built separately from the pre-evaluation, and the write-back must run against the caller's storage rather than the pre-evaluated copies. Getting this wrong will produce a fix that passes a value test while writing back to a temporary — assert the **caller-visible** value after return, in all three drivers.

## Also in scope (from the same review)

The #3156 matrix only passes **plain locals** as the `&`-operand. `WriteBackRefSlots` dispatches **four** operand kinds; the tests assert one. The other kinds were probed and the product is **correct** — `&h.f` and `&arr[1]` were `101`/`102` on base and are `91`/`91` merged — but nothing pins them, so they can regress silently. Add assertions for the remaining operand kinds across all three drivers.

## Acceptance

- Both cells produce caller-visible write-back in all three drivers, agreeing with emit.
- New rows are in the existing `Issue3140OutParameterWriteBackParityTests.cs` oracle-compare style (`Assert.NotEmpty(emitted)` then `Assert.Equal(emitted, evaluated)` / `Assert.Equal(emitted, interpreted)`) — the `NotEmpty` guard is load-bearing and must not be dropped, it is what kills a vacuous all-empty pass.
- Copying the new test file alone onto an unfixed tree must FAIL; this must be measured and reported, not asserted.
- The four `WriteBackRefSlots` operand kinds are each asserted.

## Credit

Found by rubber-duck review of PR #3156 while verifying that PR's own claims — reported as a SHOULD-FIX with both cells measured across three drivers, and correctly identified as pre-existing rather than introduced.

Related: #3140, #3156.

DavidObando:
## Summary
- route `base[IFiller].M(...)` through existing user-call ref-slot population and write-back
- retain caller lvalues during constrained-static dispatch, then write callee `out`/`ref` values back after frame restoration
- unwrap managed-address dereferences once in shared `WriteBackToOperand`; `WriteBackRefSlots` delegates to that path
- cross all 10 receiver kinds with all 4 reachable operand kinds
- anchor every parity row on a distinctive literal post-call value before comparing evaluator paths

## Impact
Post-ADR-0156, all file-based drivers emit. This defect is therefore observable through the embedding API and interactive REPL evaluator, not file-mode `gsc`/`gsi`. Tests intentionally use direct `Compilation.Evaluate` and evaluator-backed `SessionEngine`; emitted execution remains the oracle.

## Measurement
Measured evidence:

- original test rows alone on unfixed `main`: **11 failed / 43 passed**
  - `BaseInterface`: all 5 argument shapes
  - `ConstrainedStatic`: all 5 argument shapes
  - instance `Dereference`: 1 operand row
- round-two matrix alone on the prior PR product: **4 failed / 86 passed**
  - `Free-Dereference`
  - `ClassShared-Dereference`
  - `StructShared-Dereference`
  - `ConstrainedStatic-Dereference`
- fixed targeted matrix: **90/90 passed**
  - receiver shapes: 10 receivers × 5 argument shapes
  - operand shapes: 10 receivers × 4 operands (`Variable`, `Field`, `Index`, `Dereference`)
- focused reviewer-size probe: **47/47 passed**
  - all 40 receiver × operand rows
  - 7 `SingleOut` controls spanning free/shared/instance/base-interface/constrained dispatch
- original six fixed-vs-unfixed emit probes stayed byte-identical: base-interface `61`, constrained-static `81`, and four operand kinds `91`
- property address-taking was measured and rejected with `GS9001`; it is not a reachable source lvalue

Root cause is inferred from source flow: static/free calls wrote through `WriteBackToOperand`, while instance/base calls wrote through `WriteBackRefSlots`; only the latter had address-of unwrapping.

## Mutation testing
Six product mutants, zero survivors across both rounds. Round two:

- invert address-shape selection at `WriteBackRefSlots` delegation: **6 failed / 84 passed**
- invert address-shape selection in shared `WriteBackToOperand`: **10 failed / 80 passed**

Both changed `Evaluator.Calls.cs` and `out/bin/Debug/Core/GSharp.Core.dll` MD5 within this worktree. Each restore returned source and DLL hashes exactly to baseline and rebuilt green.

Literal-value assertion proof: changing generated callee output uniformly from `91` to `105` made emitted, evaluated, and interpreted paths agree on the wrong value; the new expected-value assertion killed **40/40** rows (`Expected: "91"`, `Actual: "105"`). This was a test-fixture mutant, not scored as a product mutant.

## Validation
Final base: `origin/main` `57256713`. Test denominator: 9 projects from `dotnet sln GSharp.sln list`, matching `build/generate-ci-test-matrix.py:26`.

- Core: 7,128 passed, 1 skipped
- Extensions: 104 passed
- InternalAnalyzers: 9 passed
- Interpreter: 1,460 passed
- LanguageServer: 462 passed
- Sdk: 59 passed
- GeneratorHost: 31 passed
- Compiler: 4,260 passed
- Cs2Gs: 1,907 passed
- total: **15,420 passed, 1 skipped, 0 failed**
- Debug and Release solution builds: 0 warnings, 0 errors
- ILVerify: **123/123 verified**, 2 documented suppressions, 0 failures
- `git merge-tree --write-tree origin/main HEAD` output equals `HEAD^{tree}`: `fdb593b4e28a2824eaf09f5c1cb330e28bc965a7`
- merged-tree `BUILD_RC=0`

Local e2e/Oahu gates were not run; local e2e is prohibited for this worktree and both are outside this evaluator change.

Fixes #3162

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -394,18 +394,17 @@ public sealed partial class Evaluator
         };
 
         var parameterOffset = method.ExplicitReceiverParameter == null ? 0 : 1;
-        for (int i = 0; i < node.Arguments.Length; i++)
-        {
-            var parameter = method.Parameters[i + parameterOffset];
-            var value = EvaluateExpression(node.Arguments[i]);
-            frame[parameter] = value;
-        }
+        var args = PopulateUserCallFrame(node.Arguments, method, parameterOffset, frame, out var refSlots);
 
+        object result;
         using (PushFrame(frame))
         {
             var statement = program.Functions[method];
-            return EvaluateUserMethodBody(method, statement);
+            result = EvaluateUserMethodBody(method, statement);
         }
+
+        WriteBackUserCallFrame(refSlots, args, method, parameterOffset, frame);
+        return result;
     }
 
     /// <summary>
@@ -486,12 +485,10 @@ public sealed partial class Evaluator
     /// </summary>
     private object EvaluateConstrainedStaticCallExpression(BoundConstrainedStaticCallExpression node)
     {
-        // Evaluate arguments left-to-right (matches IL evaluation order).
+        // BuildRefSlots evaluates arguments left-to-right while retaining caller lvalues for write-back.
         var evaluatedArgs = new object[node.Arguments.Length];
-        for (var i = 0; i < node.Arguments.Length; i++)
-        {
-            evaluatedArgs[i] = EvaluateExpression(node.Arguments[i]);
-        }
+        var refKinds = node.InterfaceMethod.Parameters.Select(parameter => parameter.RefKind).ToImmutableArray();
+        var refSlots = BuildRefSlots(node.Arguments, refKinds, evaluatedArgs, method: null);
 
         var slotIface = node.InterfaceMethod.StaticOwnerType as InterfaceSymbol;
         StructSymbol resolvedImpl = null;
@@ -619,10 +616,14 @@ public sealed partial class Evaluator
             frame2[target.Parameters[i]] = evaluatedArgs[i];
         }
 
+        object result;
         using (PushFrame(frame2))
         {
-            return EvaluateUserMethodBody(target, statement);
+            result = EvaluateUserMethodBody(target, statement);
         }
+
+        WriteBackUserCallFrame(refSlots, evaluatedArgs, target, parameterOffset: 0, frame2);
+        return result;
     }
 
     private object EvaluateConversionExpression(BoundConversionExpression node)
@@ -1579,6 +1580,12 @@ public sealed partial class Evaluator
                     break;
                 case BoundIndexExpression idx:
                     WriteBackIndex(idx, value);
+                    break;
+                case BoundDereferenceExpression deref:
+                    var target = deref.Operand is BoundAddressOfExpression addressOf
+                        ? addressOf.Operand
+                        : deref.Operand;
+                    WriteBackToOperand(target, value);
                     break;
             }
         }
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -1582,10 +1582,7 @@ public sealed partial class Evaluator
                     WriteBackIndex(idx, value);
                     break;
                 case BoundDereferenceExpression deref:
-                    var target = deref.Operand is BoundAddressOfExpression addressOf
-                        ? addressOf.Operand
-                        : deref.Operand;
-                    WriteBackToOperand(target, value);
+                    WriteBackToOperand(deref, value);
                     break;
             }
         }
@@ -1702,7 +1699,7 @@ public sealed partial class Evaluator
 
     /// <summary>
     /// Issue #491 (ADR-0060 follow-up): writes <paramref name="value"/> through a
-    /// bound lvalue expression (variable, field, property, or indexer access).
+    /// bound lvalue expression (variable, field, property, indexer access, or dereference).
     /// Shared between ref-aliasing local writes and the existing ref/out parameter
     /// write-back path.
     /// </summary>
@@ -1724,7 +1721,10 @@ public sealed partial class Evaluator
                 break;
             case BoundDereferenceExpression deref:
                 // *p = v under interpreter: re-route through the inner operand.
-                WriteBackToOperand(deref.Operand, value);
+                var target = deref.Operand is BoundAddressOfExpression addressOf
+                    ? addressOf.Operand
+                    : deref.Operand;
+                WriteBackToOperand(target, value);
                 break;
         }
     }
--- a/test/Interpreter.Tests/Issue3140OutParameterWriteBackParityTests.cs
+++ b/test/Interpreter.Tests/Issue3140OutParameterWriteBackParityTests.cs
@@ -10,6 +10,10 @@ using System.Linq;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
+using GSharp.Core.CodeAnalysis.Compilation;
+using GSharp.Core.CodeAnalysis.Symbols;
+using GSharp.Core.CodeAnalysis.Syntax;
+using GSharp.Repl.Engine;
 using Xunit;
 
 namespace GSharp.Interpreter.Tests;
@@ -33,34 +37,28 @@ public class Issue3140OutParameterWriteBackParityTests
 
     [Theory]
     [MemberData(nameof(ReceiverShapeMatrix))]
-    public async Task OutParameterWriteBack_GscEvaluateAndGsiMatchEmit(
+    public async Task OutParameterWriteBack_EvaluatorAndSessionEngineMatchEmit(
         ReceiverKind receiver,
         ArgumentShape shape)
     {
-        var source = BuildSource(receiver, shape);
-        var root = CreateEmptyDirectory(receiver + "-" + shape);
-        try
-        {
-            var emitted = await RunEmittedAsync(source, CreateEmptyDirectory(root, "emit"));
-            var evaluated = RunSourceDriver(
-                source,
-                CreateEmptyDirectory(root, "evaluate"),
-                GSharp.Compiler.Program.Main,
-                stripCompilerSuccess: true);
-            var interpreted = RunSourceDriver(
-                source,
-                CreateEmptyDirectory(root, "gsi"),
-                GSharp.Repl.Program.Main,
-                stripCompilerSuccess: false);
+        await AssertDriverParityAsync(BuildSource(receiver, shape), receiver + "-" + shape);
+    }
 
-            Assert.NotEmpty(emitted);
-            Assert.Equal(emitted, evaluated);
-            Assert.Equal(emitted, interpreted);
-        }
-        finally
+    /// <summary>Gets write-back operand-kind rows.</summary>
+    /// <returns>Operand-kind rows.</returns>
+    public static IEnumerable<object[]> WriteBackOperandKinds()
+    {
+        foreach (var operand in Enum.GetValues<OperandKind>())
         {
-            DeleteDirectory(root);
+            yield return new object[] { operand };
         }
+    }
+
+    [Theory]
+    [MemberData(nameof(WriteBackOperandKinds))]
+    public async Task OutParameterWriteBack_OperandKindsMatchEmit(OperandKind operand)
+    {
+        await AssertDriverParityAsync(BuildOperandSource(operand), "Operand-" + operand);
     }
 
     /// <summary>Call receiver shape.</summary>
@@ -89,6 +87,12 @@ public class Issue3140OutParameterWriteBackParityTests
 
         /// <summary>Generic class instance method.</summary>
         GenericMethod,
+
+        /// <summary>Interface default method called through <c>base[IFiller]</c>.</summary>
+        BaseInterface,
+
+        /// <summary>Static interface method called through a constrained type parameter.</summary>
+        ConstrainedStatic,
     }
 
     /// <summary>Argument shape.</summary>
@@ -108,6 +112,22 @@ public class Issue3140OutParameterWriteBackParityTests
 
         /// <summary>Distinct assignments in conditional branches.</summary>
         ConditionalOut,
+    }
+
+    /// <summary>Caller storage kind used by a ref/out argument.</summary>
+    public enum OperandKind
+    {
+        /// <summary>Local variable.</summary>
+        Variable,
+
+        /// <summary>Instance field.</summary>
+        Field,
+
+        /// <summary>Array element.</summary>
+        Index,
+
+        /// <summary>Dereferenced managed address.</summary>
+        Dereference,
     }
 
     private static string BuildSource(ReceiverKind receiver, ArgumentShape shape)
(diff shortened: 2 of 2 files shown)


DavidObando:
## Verdict: NOT MERGEABLE — one write-back shape is still dropped, and it is a *simpler* shape than the two that were fixed

Measured on the **merged** tree. `main` `16f97229`, head `11195798` (matches live `headRefOid`), computed merge base `81e6deab` (1 behind / 1 ahead; only #3190, a test rename, landed since). `git merge-tree` **output** = `7eca2346`, no conflicts. Merged HEAD `1826c74c`, Release build **rc=0, 0 warnings, 0 errors**. CI: **32/32 checks COMPLETED, all SUCCESS** including `ilverify` — nothing pending, nothing red.

**Your measurement is honest.** I reproduced your headline independently by running your test file against the unfixed base product: **11 failed / 43 passed**, and the failing rows are exactly `BaseInterface`×5, `ConstrainedStatic`×5, `Dereference`×1. That matches your report cell for cell.

---

## BLOCKER — `deref` write-back is still dropped for **static-dispatch** receivers (free functions and `shared` methods)

I ran a receiver × operand × scope matrix through `Compilation.Evaluate` on the merged tree. **28/32 pass; 4 fail, and all 4 are the same cell:**

| cell | expected | **actual** |
|---|---|---|
| `free-deref-top` | 11 | **105** |
| `free-deref-infunc` | 11 | **105** |
| `classShared-deref-top` | 11 | **105** |
| `classShared-deref-infunc` | 11 | **105** |

`105` is the initializer — the callee's write never reaches the caller. **No diagnostic, rc=0.** The identical program under emit prints `11`, so this is a real oracle divergence, not a language rule. Scope is *not* a factor (top-level and in-function fail identically), and instance receivers pass — so this is purely a receiver-axis gap.

Minimal repro, entirely top-level:

```
func Fill(out first int32) { first = 11 }
var value = 105
Fill(&*(&value))
Console.WriteLine(value)     // emit: 11    evaluator: 105
```

### Root cause — I verified the site before blaming it

There are **two** operand-dispatch switches:

- `WriteBackRefSlots` — `Evaluator.Calls.cs:1560`, switch at `:1572`
- `WriteBackToOperand` — `Evaluator.Calls.cs:1709`, switch at `:1713`

For `&*(&v)` the retained operand is `Dereference(AddressOf(v))`. The pre-existing deref case at **`Evaluator.Calls.cs:1725`** recurses straight into `deref.Operand`, which is a `BoundAddressOfExpression` — and that switch has **no `BoundAddressOfExpression` case**, so it falls through and the write is silently discarded.

Your new hunk adds the AddressOf unwrap, but **only to the sibling switch** at `:1584`. The static/free path is `EvaluateCallExpression:34`, which builds slots at `:66` and writes back through `WriteBackToOperand` — so it never reaches your fix. Instance/base calls go through `PopulateUserCallFrame` (`:360`, `:397`, `:459`) → `WriteBackUserCallFrame:1528` → `WriteBackRefSlots`, which is why exactly those receivers pass.

### The lazier fix is also the more correct one

Putting the unwrap in the **shared** callee at `:1725` fixes both paths at once, and your 6-line hunk at `:1584` then collapses to a plain delegation — net *smaller* than what is in the PR:

```csharp
case BoundDereferenceExpression deref:
    WriteBackToOperand(
        deref.Operand is BoundAddressOfExpression innerAddressOf ? innerAddressOf.Operand : deref.Operand,
        value);
    break;
```

**Measured, not proposed:** with only that one change applied, my full probe set goes **47/47 green** (from 4 failures) and **Interpreter.Tests is 1454/1454, rc=0**. Core md5 moved `dc4ea919…` → `b373eb69…` with the change and round-tripped **exactly** back to `dc4ea919…` on restore; `git status --porcelain` empty.

### Why this is a blocker and not a nit

`BuildOperandSource:229-268` hardcodes `class Target` + `target.Fill(...)` — **every operand-kind row is tested against exactly one receiver.** The PR body claims coverage of *"all reachable source operand kinds"*; that claim holds only for `ClassInstance`. This is structurally the **same** error that produced #3162 in the first place: the matrix axis is coarser than the language surface it claims to cover. Crossing the two axes you already have (10 receivers × 4 operands) is the natural fix and is what would have caught this.

---

## SHOULD-FIX — the oracle asserts *agreement*, not the written-back value

`AssertDriverParityAsync:322-324`:

```csharp
Assert.NotEmpty(emitted);
Assert.Equal(emitted, evaluated);
Assert.Equal(emitted, interpreted);
```

Nothing pins that `emitted` contains the **changed** value. Your fixture values are properly distinctive (`101` → `11`), but no assertion reads them: if a future change caused all three drivers to keep the initializer, every row still passes. Per ADR-0154, add the absolute expectation alongside the parity one — the expected string is already computable from the shape spec.

**Disclosure: I did not mutation-prove this one.** It is a structural reading of the assertion, not a measured survival, and I am flagging it as the weaker of my two findings for that reason.

---

## Two things you got right that are worth recording

**1. The ADR-0156 landmine is real, and you routed around it correctly.** I confirmed the trap empirically: on the **unfixed base** product, *all three* file drivers print the correct `11` for the very shapes this issue says are broken —

| probe | bare `gsc` | `gsc /out:` | `gsi` |
|---|---|---|---|
| `bi_var` / `cs_var` (base, unfixed) | 11 | 11 | 11 |

— because #3177/#3186 made every file driver emit. Any CLI-driven test of this defect would have been **vacuously green**. `RunEvaluator:340` calls `Compilation.Evaluate` directly and `RunSessionEngine:358` uses `SessionEngine`, so both genuinely exercise the evaluator. That was the right call.

**2. Consequence worth stating in the PR body:** because of ADR-0156 this defect is **not reachable through any file-based CLI driver**. Its user surface is the embedding API and the interactive REPL only. That does not reduce the blocker — the fix is still wrong for static receivers — but it is the accurate impact statement, and it is stronger than the current wording.

## NIT

`RunEmittedAsync:383` uses `Assembly.Load` + `GetTypes()` rather than relying on `IlVerifier` — correct, no change requested; noting it so it does not get "simplified" later.

DavidObando:
Round 2 addressed on `f29bbe01`.

- Adopted shared-callee fix: `WriteBackRefSlots` delegates dereference writes to `WriteBackToOperand`; address-of unwrap now lives there once.
- Reproduced **47/47** focused probe: 40 receiver × operand rows plus 7 single-out dispatch controls.
- Durable matrix now covers **10 receiver kinds × 4 operand kinds**, plus existing 10 × 5 argument-shape matrix; targeted result **90/90**.
- New rows alone on prior PR product fail exactly 4: `Free-Dereference`, `ClassShared-Dereference`, `StructShared-Dereference`, `ConstrainedStatic-Dereference`.
- Parity helper now asserts literal expected output first, retains `NotEmpty`, then compares evaluator and `SessionEngine` to emit. Uniform `91→105` fixture mutant fails **40/40** despite all paths agreeing.
- Two round-two product mutants killed: **6/90** and **10/90**; source/Core DLL MD5 round-tripped exactly.
- Final-base local validation: 9/9 projects, **15,420 passed / 1 skipped**, ILVerify **123/123**, merged-tree `BUILD_RC=0`.

DavidObando:
Final CI complete: 32/32 checks completed; 30 SUCCESS, 2 expected publish checks SKIPPED. `ilverify` SUCCESS. PR remains MERGEABLE.

DavidObando:
## Round-2 verification — **MERGEABLE**

Round-1 BLOCKER is **closed**. Round-1 SHOULD-FIX is **closed and now mutation-proven**. No new findings above NIT.

**Anchors.** Head `f29bbe01` (matches live `headRefOid`), `main` `3ca776d3`, merge base `57256713`, 1 behind / 2 ahead. `git merge-tree --write-tree` **output** = single tree `46b4672c`, no conflict. Merged worktree `a650cef9`, own `out/` root, Release, rc=0 / 0 warnings; all 22 non-`ref/` `GSharp.Core.dll` copies share one mtime.

---

### 1. The unwrap is genuinely single — not a delegation label

`grep -n AddressOf src/Core/CodeAnalysis/Evaluator.Calls.cs` → 3 hits, and only one is an unwrap:

- `Evaluator.Calls.cs:1723-1726` — the single `deref.Operand is BoundAddressOfExpression addressOf ? addressOf.Operand : deref.Operand`.
- `Evaluator.Calls.cs:1583-1585` — `WriteBackRefSlots` now passes the **whole** `deref` to `WriteBackToOperand`, so it re-enters the one unwrap rather than carrying a copy.

That is the rung-2 collapse, and net production is **smaller** than round 1's version.

### 2. 4-fail → 47/47 reproduced, and the shipped fix is live

My round-1 harnesses (47 rows, incl. the 32-cell receiver × operand × scope matrix) on this head: **47/47, rc=0**. Previous head was 43/47.

PRODUCT mutant at `:1723`, condition inversion (`as` + `== null`, arms swapped — never `if (true)`); md5 moved `3bc95246` → `db09f8f5`:

| | killed | survived |
|---|---|---|
| your `...MatchEmit` Dereference rows | **10 / 10** | — |
| my probes | 12 | — |
| all non-`Dereference` rows | — | 115 |

Restored by exact md5 round-trip to `3bc95246`; `git status --porcelain` empty; Interpreter.Tests **1460/1460** on the restored tree.

### 3. The parity helper really does detect a uniform regression — SHOULD-FIX closed

Round 1 I flagged `AssertDriverParityAsync` as parity-only and was explicit that it was a **structural read, not a measured survival**. It is now measured.

Uniform-regression mutant (specimen-level, disclosed as such): `BuildOperandSource:279` written value `91` → `92`, so **all three drivers move together** and every parity comparison still holds.

**40 / 40 killed.** And the kill comes from the right assertion — `Expected: "91" / Actual: "92"`, i.e. the literal `Assert.Equal(expected, emitted)` at `:349` fired, not a driver comparison. Exactly the ADR-0154 property. Closed.

### 4. Matrix expansion is real, and the receivers are distinct

`BuildOperandSource:251` now routes through `BuildSource(receiver, spec, ...)` — the hardcoded `target.Fill` that caused #3162 is gone. 10 receivers × 4 operands, initializers `101/102/104/105` vs written `91`.

Not ten spellings of two mechanisms: the `:1723` mutant killed **exactly one `Dereference` row per receiver, all ten**, so every receiver independently reaches the write-back path. They land on 5 distinct evaluator entry points (`EvaluateCallExpression`, `EvaluateUserInstanceCallExpression`, `EvaluateBaseClassCallExpression`, `EvaluateBaseInterfaceCallExpression`, `EvaluateConstrainedStaticCallExpression`).

### 5. No CLI drift — the harness still exercises the defect

`RunEvaluator:359` → `Compilation.Evaluate`; `RunSessionEngine:386` → `SessionEngine`; `RunEmittedAsync:397` → in-process `Program.Main /out:` then out-of-process `dotnet exec`, with `Assembly.Load(...).GetTypes()`. Post-ADR-0156 a file-driver test here would be vacuously green; this one is not.

---

### NIT (non-blocking, pre-existing, measured)

`Evaluator.Calls.cs:1568-1586` — after this PR the `WriteBackRefSlots` loop body is *exactly* `WriteBackToOperand(operand, args[index])`; all five cases now dispatch identically. Collapsing 18 lines → 1 is **behaviour-neutral, measured**: Interpreter.Tests **1460/1460**, Core.Tests **7128 passed / 1 skipped**, 0 warnings.

Worth doing only because two parallel operand switches drifting apart is precisely what produced #3162 — one switch cannot diverge from itself. The duplication is pre-existing and this PR adds only 3 lines to it, so it is not a condition of merge.

### CI

`.statusCheckRollup|length` = **32**, all COMPLETED: 30 SUCCESS + 2 SKIPPED (`publish`, `publish-visual-studio-extension`). **`ilverify` SUCCESS.** Nothing pending, nothing red.

### Near-miss I have to disclose

My first uniform-regression run printed **`rc=0`** and looked exactly like "the helper does not detect it". The filter `~Issue3140OutParameterWriteBack_ReceiverAndOperandKinds` matches **zero tests** — the class is `...ParityTests` and the method is `OutParameterWriteBack_...`, so that concatenation never exists. `No test matches the given testcase filter`, rc=0. A no-result is not a survival, and this one would have inverted the finding. Caught by grepping the log for `No test matches` instead of trusting rc. (Separately, my first mutant shape would not have compiled — CS0165 definite assignment — so I reshaped it before building; nothing was scored off a build failure.)

Trees clean, product restored by exact md5 round-trip.

PR #3194, opened 2026-08-04
Cache evaluator BCL override dispatch

### LINKED ISSUE #3135: BCL override dispatch has no cache: struct without overrides is ~1.8x slower under gsi after #3127

DavidObando:
## Summary

- Build and cache one BCL override map per exact source `StructSymbol` for each evaluator.
- Omit the `StructValue` override dispatcher entirely when the map is empty.
- Add a direct-evaluator correctness pin plus the bare `gsc`, emitted `gsc /out:`, and `gsi` driver matrix for no override, direct override, inherited override, and most-derived replacement cases.

Fixes #3135

## Cache identity and lifetime

The outer key is the exact `StructSymbol` object. Symbols use reference identity, so constructed generic types do not share entries with their definition or another construction. Imported CLR objects do not enter the `StructValue` path. The cache belongs to one evaluator and therefore cannot outlive or cross bound programs. The inner key is the external root `MethodInfo` used by the existing dispatcher. Derived-to-base traversal with `TryAdd` preserves the most-derived override.

## Measurement

Current `gsi file.gs` executes through the emitter after #3182, so it no longer measures this evaluator path. I used a minimal process-isolated evaluator harness against exact trees `81e6deab` (base) and `4279f6fc` (cache production change), with 200,000 inserts and 200,000 lookups. Ten alternating rounds were measured after two warmups; every process ran in a newly asserted-empty directory and printed `200000`.

| tree | median startup | median total | median work |
|---|---:|---:|---:|
| base | 126.413 ms | 1186.651 ms | 1058.538 ms |
| cache | 127.717 ms | 1183.060 ms | 1047.737 ms |

Ratio of median work: **1.010×**. Median paired ratio: **1.006×**. Thus the reported 1.8× premise is substantially overstated on current main; the cache produces a small measured improvement.

## Correctness

All paths produce:

```text
PLAIN-22-2-False
DIRECT-11-1-True-DIRECT-11
INHERITED-33-1-True-BASE-33
DERIVED-44-2-False-DERIVED-44
PLAIN-22-AGAIN-False
```

Checked through direct evaluator, bare `gsc`, `gsc /out:` plus loaded/executed assembly, and `gsi`.

Product mutations against `Issue3135_OverrideCacheSeparatesTypesMethodsAndInheritedSlots`:

- Replacing derived-first `TryAdd` with assignment failed: expected `DERIVED-44-2-False-DERIVED-44`, got `DERIVED-44-1-False-BASE-33` (`rc=1`).
- Inverting the empty-map fast path failed: direct overrides stopped dispatching (`rc=1`).
- Main production code with the new test retained passed (`rc=0`).

## Validation

`GSharp.sln` enumerates nine test projects, the same source used by `build/generate-ci-test-matrix.py`. All nine ran unfiltered; counts below are executed tests, not `--list-tests` output:

- Compiler.Tests: 4260 passed
- Core.Tests: 7128 passed, 1 skipped
- InternalAnalyzers.Tests: 9 passed
- Interpreter.Tests: 1395 passed
- Sdk.Tests: 59 passed
- Cs2Gs.Tests: 1907 passed
- GSharp.GeneratorHost.Tests: 31 passed

Total: **15,355 passed, 1 skipped, 0 failed**.

**Correction addendum:** the earlier PR body incorrectly marked Sdk.Tests as out of bounds based on a documentation comment. Structural inspection shows its relevant test calls `Program.Main` in-process and does not spawn `dotnet build` or write the shared SDK package cache. The unfiltered project run passed 59/59.

Rebased onto live main `917677ec934a48147efbb03327bc89269e8b577a`. The merge-tree result exactly matched the branch tree (`5e14f02603faca3eef1e42d8917aac36e9bd3230`). `dotnet build GSharp.sln -c Release -m:1`: `BUILD_RC=0`, 0 warnings, 0 errors. Core copies under Core, Compiler, and Repl were nonzero, byte-identical, and had equal epoch-second mtimes.

`StructValue.Equals` and `StructValue.GetHashCode` were not modified.

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -30,6 +30,7 @@ public sealed partial class Evaluator
 {
     private readonly System.Runtime.CompilerServices.ConditionalWeakTable<ConcurrentDictionary<VariableSymbol, object>, Dictionary<TypeParameterSymbol, TypeSymbol>> runtimeTypeArguments = new();
     private readonly ConcurrentDictionary<(System.Reflection.MemberInfo Member, Type RuntimeType), System.Reflection.MemberInfo> runtimeMembers = new();
+    private readonly ConcurrentDictionary<StructSymbol, IReadOnlyDictionary<MethodInfo, FunctionSymbol>> externalOverrides = new();
 
     private object EvaluateCallExpression(BoundCallExpression node)
     {
@@ -1058,8 +1059,11 @@ public sealed partial class Evaluator
         return result;
     }
 
-    private StructValue CreateStructValue(StructSymbol structType) =>
-        new(structType, TryEvaluateExternalOverride);
+    private StructValue CreateStructValue(StructSymbol structType)
+    {
+        var overrides = GetExternalOverrides(structType);
+        return new(structType, overrides.Count == 0 ? null : TryEvaluateExternalOverride);
+    }
 
     private bool TryEvaluateExternalOverride(
         StructValue receiver,
@@ -1101,16 +1105,32 @@ public sealed partial class Evaluator
         }
     }
 
-    private static FunctionSymbol FindExternalOverride(StructSymbol type, MethodInfo externalMethod)
+    private FunctionSymbol FindExternalOverride(StructSymbol type, MethodInfo externalMethod)
     {
-        FunctionSymbol externalOverride = null;
-        for (; type != null && externalOverride == null; type = type.BaseClass)
+        var overrides = GetExternalOverrides(type);
+        return overrides.TryGetValue(externalMethod, out var externalOverride)
+            ? externalOverride
+            : null;
+    }
+
+    private IReadOnlyDictionary<MethodInfo, FunctionSymbol> GetExternalOverrides(StructSymbol type) =>
+        externalOverrides.GetOrAdd(type, static key => BuildExternalOverrides(key));
+
+    private static IReadOnlyDictionary<MethodInfo, FunctionSymbol> BuildExternalOverrides(StructSymbol type)
+    {
+        var overrides = new Dictionary<MethodInfo, FunctionSymbol>();
+        for (; type != null; type = type.BaseClass)
         {
-            externalOverride = type.Methods.FirstOrDefault(method =>
-                method.IsOverride && FindExternalOverrideRoot(method)?.Equals(externalMethod) == true);
+            foreach (var method in type.Methods)
+            {
+                if (method.IsOverride && FindExternalOverrideRoot(method) is MethodInfo root)
+                {
+                    overrides.TryAdd(root, method);
+                }
+            }
         }
 
-        return externalOverride;
+        return overrides;
     }
 
     private static MethodInfo FindExternalOverrideRoot(FunctionSymbol method)
--- a/test/Interpreter.Tests/Issue2896StructObjectOverrideTests.cs
+++ b/test/Interpreter.Tests/Issue2896StructObjectOverrideTests.cs
@@ -154,6 +154,87 @@ public class Issue2896StructObjectOverrideTests
     }
 
     [Fact]
+    public async Task Issue3135_OverrideCacheSeparatesTypesMethodsAndInheritedSlots()
+    {
+        var suffix = Guid.NewGuid().ToString("N");
+        var source = $$"""
+            package Issue3135{{suffix}}
+            import System
+            import System.Collections.Generic
+
+            struct Plain{{suffix}} {
+                var Number int32
+            }
+
+            struct Direct{{suffix}} {
+                var Number int32
+                override func Equals(value object) bool -> true
+                override func GetHashCode() int32 -> 313511
+                override func ToString() string -> "DIRECT-11"
+            }
+
+            open class Base{{suffix}} {
+                open override func Equals(value object) bool -> true
+                open override func GetHashCode() int32 -> 313533
+                open override func ToString() string -> "BASE-33"
+            }
+
+            class Inherited{{suffix}} : Base{{suffix}} {
+            }
+
+            class MostDerived{{suffix}} : Base{{suffix}} {
+                override func Equals(value object) bool -> false
+                override func GetHashCode() int32 -> 313544
+                override func ToString() string -> "DERIVED-44"
+            }
+
+            let plainA = Plain{{suffix}}{Number: 11}
+            let plainB = Plain{{suffix}}{Number: 22}
+            let plainValues = Dictionary[Plain{{suffix}}, int32]()
+            plainValues[plainA] = 11
+            plainValues[plainB] = 22
+            Console.WriteLine(String.Format("PLAIN-22-{0}-{1}", plainValues.Count, plainA.Equals(plainB)))
+
+            let directA = Direct{{suffix}}{Number: 11}
+            let directB = Direct{{suffix}}{Number: 22}
+            let directValues = Dictionary[Direct{{suffix}}, int32]()
+            directValues[directA] = 11
+            directValues[directB] = 22
+            Console.WriteLine(String.Format("DIRECT-11-{0}-{1}-{2}", directValues.Count, directA.Equals(directB), directA))
+
+            let inheritedA = Inherited{{suffix}}()
+            let inheritedB = Inherited{{suffix}}()
+            let inheritedValues = Dictionary[Inherited{{suffix}}, int32]()
+            inheritedValues[inheritedA] = 11
+            inheritedValues[inheritedB] = 33
+            Console.WriteLine(String.Format("INHERITED-33-{0}-{1}-{2}", inheritedValues.Count, inheritedA.Equals(inheritedB), inheritedA))
+
+            let derivedA = MostDerived{{suffix}}()
+            let derivedB = MostDerived{{suffix}}()
+            let derivedValues = Dictionary[MostDerived{{suffix}}, int32]()
+            derivedValues[derivedA] = 11
+            derivedValues[derivedB] = 44
+            Console.WriteLine(String.Format("DERIVED-44-{0}-{1}-{2}", derivedValues.Count, derivedA.Equals(derivedB), derivedA))
+
+            Console.WriteLine(String.Format("PLAIN-22-AGAIN-{0}", plainA.Equals(plainB)))
+            """;
+
+        var emitted = await RunDriverAsync(source, suffix + "Emit", "gsc-emit");
+        Assert.Equal(
+            """
+            PLAIN-22-2-False
+            DIRECT-11-1-True-DIRECT-11
+            INHERITED-33-1-True-BASE-33
+            DERIVED-44-2-False-DERIVED-44
+            PLAIN-22-AGAIN-False
+            """.Replace("\r\n", "\n", StringComparison.Ordinal) + "\n",
+            emitted);
+        Assert.Equal(emitted, Evaluate(source));
+        Assert.Equal(emitted, await RunDriverAsync(source, suffix + "Evaluate", "gsc-evaluate"));
+        Assert.Equal(emitted, await RunDriverAsync(source, suffix + "Gsi", "gsi"));
+    }
+
+    [Fact]
     public void GenericInterfaceOperatorNestedAndSharedShapes_DispatchOverrides()
     {
         const string Source = """


DavidObando:
# Round 1 review — PR #3194 (issue #3135, BCL override dispatch cache)

**Verdict: MERGEABLE** — 0 BLOCKER, 2 SHOULD-FIX, 2 NIT.

Anchors. `origin/main` fetched at review time = `917677ec`; PR head = `fea45cb6`. The head is **not** a descendant of current `main` (merge base `16f97229`, `main` 1 commit ahead), so everything below was measured on the **merged tree** `89a4468e`, not the branch tree. `git merge-tree --write-tree` output was 1 line (`5e14f026`) with **0 conflict lines**. Both review worktrees have their own `OutRoot` (`build/gsharp.build.props:67` = `$(RepoRoot)/out`), verified before any build; both built `-c Release`, `BUILD_RC=0`, 0 warnings, 38 project outputs with epoch mtimes inside a 5-minute window.

This is a performance PR, so the numbers are interrogated before the code.

---

## The headline: the issue's premise is superseded, and the PR is worth more than its own benchmark claims

**1. No file driver reaches the changed code.** The issue is titled "*…slower under `gsi`*". I proved by mutation that `gsi` cannot observe this change at all.

Mutant **M1** inverted the new fast path at `src/Core/CodeAnalysis/Evaluator.Calls.cs:1065` (`overrides.Count == 0` → `overrides.Count != 0`), which breaks override dispatch for every struct that *has* overrides. Asserted applied by `git diff` (a real statement, not a brace) and by `GSharp.Core.dll` md5 movement `43fcfd08…` → `6e6e5d7d…`. Results on one program with a no-override struct and an override-bearing struct:

| driver | output under M1 | reached evaluator? |
|---|---|---|
| bare `gsc` | `PLAIN-22-2-False` / `DIRECT-11-1-True-DIRECT-11` (correct) | **no** |
| `gsc /out:` + run | identical (correct) | **no** |
| `gsi <file>` | identical (correct) | **no** |
| `Compilation.Evaluate` (direct) | `DIRECT-11-2-True-Direct(Number=11)` — **wrong** | **yes** |

So after #3177/#3182/#3186 the only things that execute `CreateStructValue` are the public `Compilation.Evaluate` API and the interactive session (`src/Repl/Engine/SessionEngine.cs:220`, and the REPL file-load at `src/Repl/Screens/ReplScreen.cs:580`). The author already says this in the PR body; I am confirming it by measurement rather than by reading, and recording that **the issue's driver framing no longer holds**.

**2. The PR undersells itself.** The PR reports **1.010×** and concludes the fix produces only "a small measured improvement". On the issue's *own* workload shape (20,000 inserts + 20,000 lookups, no-override struct, output `20000`) I measure a consistent **~1.15–1.20× improvement**. Measurement: a process-isolated harness driving `Compilation.Evaluate` by reflection, 7 evaluations per process, median taken (so the cold first evaluation is excluded), and a zero-iteration variant of the same source subtracted to remove parse/bind. Three alternating passes, `uptime` 9.73 before and 9.79 after:

| pass | base `917677ec` (ms) | merged (ms) |
|---|---:|---:|
| 1 | 120.827 | 102.073 |
| 2 | 121.118 | 101.050 |
| 3 | 114.976 | 105.550 |

Ranges are **non-overlapping** (base min 114.976 > merged max 105.550). The likely reason the PR's own number is flat: at 200,000 elements the `Dictionary` internals and GC dominate and dilute the effect being measured. **The PR's benchmark is not wrong, it is too large to resolve its own signal.**

**3. No pessimisation found in the case the cache does not help.** Struct-creation-only workload (200,000 constructions, no BCL boundary) and an override-bearing workload both came out merged ≤ base or indistinguishable. Nothing got slower.

---

## SHOULD-FIX 1 — the cache is pinned by zero tests, and removing it is *worse than the pre-PR baseline*

`src/Core/CodeAnalysis/Evaluator.Calls.cs:1116-1117`.

Mutant **M3** deletes the memoisation while keeping behaviour identical:

```
-        externalOverrides.GetOrAdd(type, static key => BuildExternalOverrides(key));
+        BuildExternalOverrides(type);
```

Asserted applied (`git diff` non-empty, md5 moved). **It survives `Interpreter.Tests` 1395/1395, `RC=0`.** Nothing in the suite pins the thing the PR is named for. This is the ADR-0154 shape: the test asserts returned values, so it cannot tell a cache from no cache.

I confirmed that directly — I copied the PR's **test file only** onto the base tree, with **no** production change (`grep -c externalOverrides` = 0), rebuilt, and the new test **passes**. So the honest answer to "what test would fail if the cache were simply deleted?" is **none**.

A perf-only mutant is not a survival until it is shown live, so I measured it. On a struct with 20 methods and 200,000 constructions:

| pass | base (ms) | M3 = cache removed (ms) |
|---|---:|---:|
| 1 | 118.325 | 141.929 |
| 2 | 116.558 | 129.246 |
| 3 | 150.143 | 168.894 |

M3 is slower than base in 3/3 passes. That matters because it shows the cache is **load-bearing, not decorative**: the fast path has to know whether the override map is empty at *every* struct construction, and without memoisation that question costs a scan of every method in the base chain — which is a *regression relative to `main`*, not merely a lost optimisation.

So the risk is concrete: a future reader can delete `GetOrAdd` as redundant machinery, every test stays green, and the interpreter ends up slower than it was before this PR.

**Recommended fix (small):** increment a counter in `BuildExternalOverrides`, expose it `internal`, and assert it equals the number of distinct struct types after a loop that constructs many instances. One product line, a handful of test lines.

*Self-correction worth recording:* my first reading of M3 was wrong. On structs with 0 and 2 methods, M3 measured identical to the full PR, and I had concluded the memoisation contributed nothing and the win was entirely the one-line fast path. That conclusion was an artefact of my workload — those structs have almost nothing to scan. The 20-method workload reverses it. **My scope was the finding, not the product.**

## SHOULD-FIX 2 — three of the driver matrix's four columns are vacuous for this change

`test/Interpreter.Tests/Issue2896StructObjectOverrideTests.cs` (new test) asserts against `gsc-emit`, then `Evaluate(source)`, then `gsc-evaluate`, then `gsi`. Per the M1 table above, **only the `Evaluate(source)` line executes the changed code**; the other three emit and would pass with override dispatch completely broken. The PR body presents this as "the bare `gsc`, emitted `gsc /out:`, and `gsi` driver matrix", which reads as four-way coverage of the fix.

The test is not wrong and the extra columns are legitimate anti-regression value. The hazard is that the one load-bearing assertion is not distinguishable from the three decorative ones, so a later cleanup can delete the only real pin and leave a green suite.

**Recommended fix:** one comment on the `Evaluate(source)` assertion stating it is the sole column that reaches `CreateStructValue`, and a corrected sentence in the PR body.

## NIT 1 — the fast path still allocates on every `Equals`

`src/Core/CodeAnalysis/StructValue.cs:158-160`:

```csharp
public override bool Equals(object obj)
{
    if (TryInvokeObjectOverride(ObjectEqualsMethod, [obj], out var result))
```

`TryInvokeObjectOverride` returns immediately when the dispatcher is null (`:267-273`), but `[obj]` is materialised at the call site first, so the no-override case still allocates one `object[]` per `Equals`. `GetHashCode` and `ToString` are unaffected (`[]` lowers to `Array.Empty<object>()`). Hoisting the null check would recover it. Out of scope, but this is a perf PR and it is the same hot path.

## NIT 2 — test name claims more than the test checks

`Issue3135_OverrideCacheSeparatesTypesMethodsAndInheritedSlots` passes verbatim on a tree with no cache at all (measured above). Per ADR-0154 the name should describe the dispatch semantics it actually pins.

---

## What I checked and found clean

- **Cache lifetime / staleness.** `new Evaluator(...)` occurs at exactly one site (`Compilation.cs:393`) and `externalOverrides` is a non-static instance field, so a cache cannot outlive a single `Evaluate()` call. `SessionEngine.cs:220` calls `Evaluate` per submission, so interactive redefinition gets a fresh cache by construction — the strongest possible answer, and it does not depend on symbol identity.
- **Map cannot go stale mid-evaluation.** `StructSymbol.Methods` is an `ImmutableArray` replaced only via `SetMethods`, whose call sites are `Binding/` (pre-evaluation) and `Emit/` (`ClosureEmitter`, `StateMachineEmitter` — not on the evaluator path). `ImportedTypeSymbol` is `: TypeSymbol`, not `StructSymbol`, and sets its aggregate's methods before the symbol escapes.
- **Generics.** `Box[int32]` and `Box[string]` with overrides, plus a no-override generic: byte-identical output on base and merged. No cross-construction contamination. Note `StructSymbol.Methods` delegates to `Definition.Methods` for constructions, so entries duplicate per construction rather than diverging — wasteful at worst, and identical to the pre-PR walk.
- **Thread safety.** 8 goroutines (`go` + `scope` + channel) concurrently constructing override-bearing and no-override structs: `GO-True` on 6/6 runs. `ConcurrentDictionary` for the outer map, and the inner `Dictionary` is published read-only through `GetOrAdd`, so the duplicate-factory-invocation window is benign.
- **Behavioural equivalence.** All three drivers plus the direct evaluator agree on the override matrix, on both trees.
- **`ilverify`** is present and passing on the PR's CI run.

## Mutation summary — per hunk, two-sided, all asserted applied

| mutant | hunk | applied | result |
|---|---|---|---|
| M1 `Count == 0` → `Count != 0` | `Evaluator.Calls.cs:1065` fast path | md5 `43fcfd08`→`6e6e5d7d` | **KILLED** (`RC=1`) |
| M2 `TryAdd(root, method)` → `overrides[root] = method` | `Evaluator.Calls.cs:1129` derived-first | md5 `43fcfd08`→`ab0b293c` | **KILLED** (`RC=1`, `DERIVED-44-1-False-BASE-33`) |
| M3 `GetOrAdd(...)` → `BuildExternalOverrides(type)` | `Evaluator.Calls.cs:1117` memoisation | md5 moved | **SURVIVES 1395/1395**, proven live on the perf axis |

M1 and M2 independently reproduce the two kills the PR body claims. Each mutant was reverted, rebuilt, and the revert confirmed by **exact md5 round-trip** back to `43fcfd08070afd45f4fcf5929fb8e794` (3/3), with the test re-run green on the restored tree. Both worktrees finish at `git status --porcelain` drift 0.

## Test projects — denominator 9, from `dotnet sln GSharp.sln list`

| project | result |
|---|---|
| Core.Tests | 7128 passed, 1 skipped |
| Compiler.Tests | 4260 passed |
| Interpreter.Tests | 1395 passed |
| Cs2Gs.Tests | 1907 passed |
| Sdk.Tests | 59 passed |

**15,356 tests, 9 of 9 projects, 0 failures, 1 skip**, all unfiltered, every `RC=0`. **`Sdk.Tests` is in bounds** — its acceptance tests call `Program.Main` in-process and write no SDK package; the PR body lists it as NOT RUN on that basis, which is worth correcting.

`Compiler.Tests` **was run, not skipped** (33 m 9 s): it contains 10 `Compilation.Evaluate` call sites, so containment does not hold for a change to the evaluator and skipping it could not have been justified.

## Note on CI state

At review time `gh pr checks 3194` showed 30 checks with **2 pending** (`tests (compiler-remainder)`, `cs2gs-oahu`) and `mergeable=UNKNOWN`. Nothing I measured depends on those, but the tree is not fully green yet and the branch needs an update against `main` before merge.

## Review-harness disclosures

Two of my own errors, both caught and neither load-bearing:

1. **My evaluator harness silently swallowed diagnostics.** A concurrency probe returned empty output on *both* trees, which reads exactly like a product failure. It was a missing `import Gsharp.Extensions.Go`, and then a `go` target that returned `void` where `go` requires a value-returning function. A known-good control (`samples/GoScope.gs`, `rc=0`, output `6`) is what separated "my source is wrong" from "the product is wrong". The probe passed once corrected. A failing probe is not evidence until it fails for the reason you think it does.
2. **`packages.lock.json` is TRACKED in this repo, not untracked.** After all measurement was complete I ran a cleanup that deleted 27 of them across both worktrees on the assumption they were build residue. `git status` showed them as tracked deletions, and `git checkout -- .` restored both trees to drift 0. No build, test, mutant or timing in this review ran against a tree in that state — every one of them completed beforehand — but the assumption itself is worth correcting for anyone else applying the same cleanup.

Both worktrees finish at drift 0 against `89a4468e` and `917677ec` respectively. The shared checkout was never written to and was never used as a source of evidence.

DavidObando:
Current-main revalidation disproved this change as a performance fix, so I am closing rather than shipping a measured regression.

Measured on fetched `origin/main` `074f28a7d7780626fb1e64286ebd1b98f712d0d7`, same machine, static base/head binaries, alternating order, every pre-run load captured and below 30:

- Exact issue workload (20,000 inserts + 20,000 lookups), direct evaluator, 40 paired runs: base median **190.874 ms** (172.365–209.361, σ 8.289); cached median **200.712 ms** (192.139–218.500, σ 6.424). Ratio **0.951×**: cache was 5.2% slower.
- `gsi file.gs`, 20 paired work-minus-empty runs: base median **147.462 ms**; cached median **151.953 ms**; paired median ratio **1.003×**. This is expected on current main because file-mode `gsi` executes emit-to-memory after ADR-0156 and does not enter this evaluator path.
- Direct hit probe on the scaled workload: **400,000 requests, 1 build, 399,999 hits** (99.99975%). The cache definitely fired; hit overhead outweighed the avoided scan.
- Uncached exact-workload probe counted **60,000 override scans**, confirming the reported dispatch path still runs in direct evaluation.

I also measured cheaper type-local and three-slot negative-cache variants. Best scaled result was **1.001×** (noise); none reproduced or closed the reported 1.8× gap. Shipping any tested variant would be performance theater or a regression.

Final rebased branch build: `dotnet build GSharp.sln -c Release -m:1`, rc 0, 0 warnings, 0 errors. Nine projects remain the denominator from `dotnet sln GSharp.sln list` / `build/generate-ci-test-matrix.py`; they were not rerun after rejecting the product change. Previous PR CI, including required `ilverify`, passed on the earlier head.

DavidObando:
Latest-main addendum: fetched through `3ca776d3f2ab1c0b7634a01a0351962a672c4542`, rebased branch to `33943112853fcf6c855851c5b5bfb4d421e1eaed`, force-pushed, and rebuilt `GSharp.sln -c Release -m:1` successfully (0 warnings, 0 errors). New commits #3147/#3188 do not change evaluator override dispatch; #3147 keeps file-mode `gsi` on `EmittedProgramHost` while adding reference resolution. Measurement conclusion remains unchanged.

Issue #1923, opened 2026-07-03
gsc: pattern-matching seams: boxed-constant `==`, `!= nil` on non-nullable classes, no flow-narrowing after nil checks, nullable/boxed pattern subjects


DavidObando:
Canonical pattern lowerings gsc rejects (each with a passing sibling): (1) `answer == 42` on `object` → GS0129; (2) is-pattern lowering emits `subject != nil` which is undefined for non-nullable class types → GS0129; (3) after `person.Address != nil`, member access `person.Address.City` on a `T?` property does not flow-narrow → GS0158; (4) property pattern on a nullable switch-expression subject → GS0172; (5) boxed-subject spill `let __spill0 = object(person)` on a `T?` → GS0155. (grid app G04).
## Definition of done
- [ ] Fix + `Issue<N>*.cs` regression test
- [ ] Re-enable the quarantined grid fixture (or add one) and re-capture its baseline
- [ ] `tools/cs2gs/coverage/csharp-construct-inventory.json` row updated

_Discovered by the ADR-0138 conformance-grid build-out (gsc 0.2.585+01ad6ec964); quarantined sources live next to each grid app under `tools/cs2gs/corpus/grid/`._

PR #3195, opened 2026-08-04
Fix evaluator object reference equality

### LINKED ISSUE #3173: [bug] Reference equality on object-typed operands diverges: evaluator and interpreter say True, emit says False
## Summary

Reference-equality comparison of **`object`-typed operands** diverges between drivers. `==` / `!=` on two `object`-typed locals holding distinct instances give:

| operand contents | bare `gsc` (evaluate) | `gsi` (interpret) | `gsc /out:` (emit — oracle) |
|---|---|---|---|
| two runtime-computed `string`s | `True` | `True` | **`False`** |
| two class instances | `True` | `True` | **`False`** |
| two boxed `5`s | `True` | `True` | **`False`** |

No diagnostic in any driver. The program runs and prints the wrong answer — a **silent wrong result**, the same signature as #3134 and #3140.

This is **pre-existing and not a regression**: measured identical on the base commit and on the head of PR #3153.

## Why PR #3153 does not cover it

#3153 fixed the *statically-typed* reference-equality path. Its fix works by setting a flag inside **`MakeReferenceEquality`**, which is a **last-resort fallback** reached only after user-defined and CLR operator resolution both decline.

`object`-typed operands never reach it. They bind through the **built-in operator table's `object` arm**, which resolves successfully and therefore short-circuits the fallback — so the flag is never set and the evaluator keeps routing through the value-equality path.

The PR is correct within its scope. Its commit subject, *"Preserve evaluator reference equality operators"*, reads broader than what actually holds, which is how this gap stayed invisible.

## Fix guidance

The fix must set the same reference-equality intent on the **built-in `object` arm** of the operator table, not only in `MakeReferenceEquality`. Do **not** solve this by making `MakeReferenceEquality` run earlier or unconditionally — that would bypass user-defined `operator ==` and CLR operators, and #3153's review already proved those are load-bearing:

- a `data class` must keep **value** equality;
- an imported `Version` (which has both an `Equals` override **and** an `operator ==`) must stay **value**-equal;
- two distinct runtime-computed `string`s compared as **`string`** (not `object`) must stay `True`.

All three of those are currently correct and must remain so. A fix that flips any of them has over-corrected.

## Note on `string`

The `string`/`string` case is deliberately excluded above. Comparing two `string`-typed operands by value is correct. Only the **`object`-typed** case is wrong, because at that static type the comparison is specified as reference identity.

## Acceptance

- All three rows above agree with emit across all three drivers.
- The three over-correction controls listed above are asserted and unchanged.
- Pins live in the existing `#3134` operator matrix style — `InlineData` rows asserting `==` **and** `!=` per specimen, compared against emit as oracle.
- A two-sided PRODUCT mutant on the new arm goes RED, with a control expected to kill also going RED.
- Copying the new rows alone onto an unfixed tree must FAIL; measure and report, do not assert.

## Credit

Found by rubber-duck review of PR #3153 round 2, while probing whether that PR's fix had *over*-corrected. The reviewer hunted the string-collapse case specifically because the guard admits `string`/`object`, established the statically-typed cases were all correct, and only then isolated the `object` arm as a separate, pre-existing hole. Reported as a SHOULD-FIX with all three cells measured on both trees.

Related: #3134, #3153.

DavidObando:
## Summary

- make evaluator `object == object` / `!=` use the ADR-0045 reference-equality contract
- preserve CLR-compatible identity for interned string literals and boxed user structs
- preserve boxed aliases through assignment, containing-struct copies, and user/CLR interface casts
- keep mixed `object`/value equality on ADR-0045 reference semantics after #3232 briefly introduced a boxed-value special path
- pin emitted output first, then require `Compilation.EvaluateWithEvaluator` and `SessionEngine` to match it

Closes #3173.

## Root cause

Round 1 correctly exposed reference equality, but the evaluator had not preserved the identities being compared. Equal literal occurrences were separate host strings instead of CLR-interned `ldstr` references, while `StructValue` represented both values and boxes and was copied/re-boxed during conversions and assignments. Interface aliases also failed because evaluator type matching ignored implemented interfaces.

The fix interns literal strings, marks boxed `StructValue` instances, creates one box snapshot on value-to-reference conversion, preserves that box by reference on assignment/copy, and recognizes user/imported interfaces in `MatchesType`.

During the latest-main rebase, #3232 introduced `IsBoxedValueEquality`/`Object.Equals` for `object == int`, contradicting ADR-0045:41-42 and turning the required `False` control back to `True`. This PR removes only that boxed-value special path; #3232's unrelated nil and `!!` fixes remain intact.

## Nine-cell conformance table

| cell | emit | evaluator / `SessionEngine` |
|---|---:|---:|
| `object == 42` | `False` | `False` |
| literal string vs `String.Concat` | `False` | `False` |
| distinct boxed structs `==` | `False` | `False` |
| distinct boxed structs `!=` | `True` | `True` |
| distinct boxed enums `==` | `False` | `False` |
| distinct boxed enums `!=` | `True` | `True` |
| boxed-int alias `a == c` | `True` | `True` |
| two `object` locals holding equal literals | `True` | `True` |
| boxed user-struct alias `a == c` | `True` | `True` |

The original seven corrected cells remain corrected; both round-2 regressions now match emit.

## RED proof

Final tests applied alone to unfixed round-1 head `f2699263`, after rebuilding: **5/13 RED**:

- `objectStringLiterals`
- `objectBoxedStruct`
- `objectBoxedStructInCopiedValue`
- `objectBoxedStructInterfaceAlias`
- `objectBoxedStructClrInterfaceAlias`

For the two reported regressions, emitted/file drivers remained green while both affected in-process surfaces (`Compilation.EvaluateWithEvaluator` and `SessionEngine`) failed. The tree was restored to porcelain-0 and rebuilt before continuing.

## Product mutation results

Each mutant was a condition inversion, applied per hunk, verified with `git diff HEAD` plus `GSharp.Core.dll` md5 movement, and restored by exact source/DLL hash round-trip.

| mutant | result |
|---|---:|
| literal interning condition | 9 RED |
| value-to-reference boxing condition | 3 RED |
| boxed assignment preservation | 1 RED |
| nested boxed-value copy preservation | 1 RED |
| user-interface matching | 1 RED |
| imported CLR-interface matching | 1 RED |
| reference-equality control | 8 RED |
| restore #3232 boxed-value special path | 1 Core RED + 1 emitted/ILVerify RED |

## Validation

Denominator **9**, from `dotnet sln GSharp.sln list` / CI matrix projects.

| project | passed | skipped |
|---|---:|---:|
| Compiler.Tests | 4263 | 0 |
| Core.Tests | 7135 | 1 |
| Extensions.Tests | 104 | 0 |
| InternalAnalyzers.Tests | 9 | 0 |
| Interpreter.Tests | 1490 | 4 |
| LanguageServer.Tests | 462 | 0 |
| Sdk.Tests | 59 | 0 |
| Cs2Gs.Tests | 1907 | 0 |
| GSharp.GeneratorHost.Tests | 31 | 0 |

Total: **15,460 passed, 5 skipped, 0 failed**. This nine-project local run predates later main rebases; current-head targeted validation is below and CI completed with **30 passed, 2 skipped**.

Current head `eb06ce81d` on `origin/main` `5cd0d766a` (#3242 evaluator retirement gate):

- issue #3173 matrix: 13/13
- evaluator coverage remains in-process (`Compilation.Evaluate` + `SessionEngine`); no script-mode `--engine evaluator` result used
- issue #1923 controls: 8/8
- issue #3224 emitted/ILVerify control: 1/1
- ILVerify gate: 123/123
- boxed-value product mutant: Core 1/8 RED; emitted control 1/1 RED; source and `GSharp.Core.dll` exact hash round-trip
- `git merge-tree --write-tree origin/main HEAD` exactly equals `HEAD^{tree}` (`2acab6102ef24d2f3e33dad6a9c7f47c17b86598`)

### DIFF
--- a/src/Core/CodeAnalysis/Binding/BoundBinaryOperator.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundBinaryOperator.cs
@@ -672,8 +672,20 @@ public sealed record BoundBinaryOperator
         list.Add(new BoundBinaryOperator(SyntaxKind.BangEqualsToken, BoundBinaryOperatorKind.NotEquals, TypeSymbol.String, TypeSymbol.Bool));
 
         // ADR-0045: `object` reference equality.
-        list.Add(new BoundBinaryOperator(SyntaxKind.EqualsEqualsToken, BoundBinaryOperatorKind.Equals, TypeSymbol.Object, TypeSymbol.Bool));
-        list.Add(new BoundBinaryOperator(SyntaxKind.BangEqualsToken, BoundBinaryOperatorKind.NotEquals, TypeSymbol.Object, TypeSymbol.Bool));
+        list.Add(new BoundBinaryOperator(
+            SyntaxKind.EqualsEqualsToken,
+            BoundBinaryOperatorKind.Equals,
+            TypeSymbol.Object,
+            TypeSymbol.Object,
+            TypeSymbol.Bool,
+            isReferenceEquality: true));
+        list.Add(new BoundBinaryOperator(
+            SyntaxKind.BangEqualsToken,
+            BoundBinaryOperatorKind.NotEquals,
+            TypeSymbol.Object,
+            TypeSymbol.Object,
+            TypeSymbol.Bool,
+            isReferenceEquality: true));
 
         return list.ToArray();
     }
--- a/src/Core/CodeAnalysis/Binding/BoundBinaryOperator.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundBinaryOperator.cs
@@ -45,8 +45,7 @@ public sealed record BoundBinaryOperator
         TypeSymbol leftType,
         TypeSymbol rightType,
         TypeSymbol resultType,
-        bool isReferenceEquality = false,
-        bool isBoxedValueEquality = false)
+        bool isReferenceEquality = false)
     {
         SyntaxKind = syntaxKind;
         Kind = kind;
@@ -54,7 +53,6 @@ public sealed record BoundBinaryOperator
         RightType = rightType;
         Type = resultType;
         IsReferenceEquality = isReferenceEquality;
-        IsBoxedValueEquality = isBoxedValueEquality;
     }
 
     /// <summary>
@@ -84,17 +82,6 @@ public sealed record BoundBinaryOperator
 
     /// <summary>Gets a value indicating whether equality compares reference identity.</summary>
     internal bool IsReferenceEquality { get; }
-
-    /// <summary>
-    /// Gets a value indicating whether equality compares boxed values by VALUE
-    /// (dispatching through <c>Object.Equals(object, object)</c>) rather than by
-    /// box reference identity. Set only by the Issue #1923 boxed-constant
-    /// equality seam (<c>answer == 42</c> where <c>answer</c> is typed
-    /// <c>object</c> and the other operand boxes in): the bound semantics
-    /// compare the boxed value, so the emitter must not fall through to the
-    /// reference-identity <c>ceq</c> tail (issue #3224).
-    /// </summary>
-    internal bool IsBoxedValueEquality { get; }
 
     /// <summary>
     /// Binds a syntax kind and a type symbol to the corresponding bound binary operator, or
@@ -391,26 +378,6 @@ public sealed record BoundBinaryOperator
             ? BoundBinaryOperatorKind.Equals
             : BoundBinaryOperatorKind.NotEquals;
         return new BoundBinaryOperator(syntaxKind, kind, leftType, rightType, TypeSymbol.Bool, isReferenceEquality: true);
-    }
-
-    /// <summary>
-    /// Issue #3224: builds the boxed-value equality (<c>==</c> / <c>!=</c>)
-    /// operator the Issue #1923 boxed-constant seam binds when exactly one
-    /// operand is statically <c>object</c> and the other side boxes in via an
-    /// implicit conversion (<c>answer == 42</c> where <c>answer</c> is typed
-    /// <c>object</c>). Both operands are <c>object</c>-typed after the boxing
-    /// conversion and compare by VALUE — the evaluator's historical semantics —
-    /// so the emitter dispatches through <c>Object.Equals(object, object)</c>
-    /// instead of the reference-identity <c>ceq</c> tail.
-    /// </summary>
-    /// <param name="syntaxKind">The <c>==</c> or <c>!=</c> token.</param>
-    /// <returns>The boxed-value equality operator over <c>object</c> operands.</returns>
-    internal static BoundBinaryOperator MakeBoxedValueEquality(SyntaxKind syntaxKind)
-    {
-        var kind = syntaxKind == SyntaxKind.EqualsEqualsToken
-            ? BoundBinaryOperatorKind.Equals
-            : BoundBinaryOperatorKind.NotEquals;
-        return new BoundBinaryOperator(syntaxKind, kind, TypeSymbol.Object, TypeSymbol.Object, TypeSymbol.Bool, isReferenceEquality: false, isBoxedValueEquality: true);
     }
 
     /// <summary>
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Operators.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Operators.cs
@@ -2044,11 +2044,7 @@ internal sealed partial class ExpressionBinder
         // 'int32'" even though C# boxes the value-type operand and compares
         // by reference identity. When exactly one side is `object` and the
         // OTHER side has an implicit (boxing) conversion to `object`, box it
-        // and bind the boxed-VALUE equality operator: the bound semantics
-        // compare the boxed value (the evaluator's historical behavior), so
-        // the emitter must dispatch through `Object.Equals(object, object)`
-        // rather than compare two distinct box references with `ceq`, which
-        // silently yielded `false` for equal values (issue #3224).
+        // and rebind the homogeneous `object == object` operator.
         if (boundOperator == null && (operatorKind == SyntaxKind.EqualsEqualsToken || operatorKind == SyntaxKind.BangEqualsToken))
         {
             if (boundLeft.Type == TypeSymbol.Object
@@ -2056,14 +2052,14 @@ internal sealed partial class ExpressionBinder
                 && Conversion.Classify(boundRight.Type, TypeSymbol.Object).IsImplicit)
             {
                 boundRight = conversions.BindConversion(rightLocation, boundRight, TypeSymbol.Object);
-                boundOperator = BoundBinaryOperator.MakeBoxedValueEquality(operatorKind);
+                boundOperator = BoundBinaryOperator.Bind(operatorKind, boundLeft.Type, boundRight.Type);
             }
             else if (boundRight.Type == TypeSymbol.Object
                 && boundLeft.Type != TypeSymbol.Object
                 && Conversion.Classify(boundLeft.Type, TypeSymbol.Object).IsImplicit)
             {
                 boundLeft = conversions.BindConversion(leftLocation, boundLeft, TypeSymbol.Object);
-                boundOperator = BoundBinaryOperator.MakeBoxedValueEquality(operatorKind);
+                boundOperator = BoundBinaryOperator.Bind(operatorKind, boundLeft.Type, boundRight.Type);
             }
         }
 
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Operators.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Operators.cs
@@ -709,30 +709,6 @@ internal sealed partial class MethodBodyEmitter
             return;
         }
 
-        // Issue #3224: the Issue #1923 boxed-constant equality seam
-        // (`answer == 42` where `answer` is typed `object` and the other
-        // operand boxes in). The bound semantics compare the boxed VALUE —
-        // the binder marks the operator with IsBoxedValueEquality — so
-        // dispatch through static `Object.Equals(object, object)` (which
-        // routes to the boxed value's Equals override and handles nil on
-        // either side). Falling through to the bottom-of-method `ceq` tail
-        // compared two distinct box references and silently yielded `false`
-        // for equal values. Both operands are already `object`-typed after
-        // the seam's boxing conversion, so no `box` is needed here.
-        if (b.Op.IsBoxedValueEquality)
-        {
-            this.EmitExpression(b.Left);
-            this.EmitExpression(b.Right);
-            this.il.Call(this.outer.wellKnown.GetObjectStaticEqualsReference());
-            if (b.Op.Kind == BoundBinaryOperatorKind.NotEquals)
-            {
-                this.il.LoadConstantI4(0);
-                this.il.OpCode(ILOpCode.Ceq);
-            }
-
-            return;
-        }
-
         // String concatenation / equality go through BCL helpers. Issue #1927:
         // `+` also accepts `string?` mixed with `string` on either/both sides
         // (the operator table binds that combination too); at runtime a
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -672,11 +672,11 @@ public sealed partial class Evaluator
             || (node.Type?.ClrType != null && !node.Type.ClrType.IsValueType))
         {
             // Reference upcast (class → implemented interface, derived class
-            // → base class, or any → object). Also covers issue #521: a CLR
-            // class or interface widening. The interpreter stores instances
-            // as boxed objects, so the upcast is a no-op at runtime — only
-            // the bind-time static type changes.
-            return value;
+            // → base class, or any → object). User value structs take one
+            // boxed snapshot; reference values keep their existing identity.
+            return value is StructValue { IsBoxed: false, StructType.IsClass: false } structValue
+                ? structValue.Box()
+                : value;
         }
         else if (node.Type?.ClrType != null && IsSupportedNumericClrType(node.Type.ClrType))
         {
@@ -1669,9 +1669,9 @@ public sealed partial class Evaluator
 
     private void Assign(VariableSymbol variable, object value)
     {
-        // Value-typed structs are copied on assignment (Go semantics).
-        // Class types (Phase 3.B.3) are reference types — share the instance.
-        if (value is StructValue sv && !sv.StructType.IsClass)
+        // Unboxed value structs copy on assignment (Go semantics). Boxes and
+        // class values are references — preserve the existing instance.
+        if (value is StructValue sv && !sv.StructType.IsClass && !sv.IsBoxed)
         {
             value = sv.Copy();
         }
--- a/src/Core/CodeAnalysis/Evaluator.Concurrency.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Concurrency.cs
@@ -369,6 +369,23 @@ public sealed partial class Evaluator
                 {
                     return true;
                 }
+
+                // A boxed G# value retains both user-declared and imported
+                // interface identities through is/as and pattern bindings.
+                if (targetType is InterfaceSymbol targetInterface
+                    && t.Interfaces.Any(iface => iface.SelfAndAllBaseInterfaces().Contains(targetInterface)))
+                {
+                    return true;
+                }
+
+                if (targetType.ClrType is { IsInterface: true } targetClrInterface
+                    && t.ImplementedClrInterfaces.Any(iface =>
+                        iface.ClrType != null
+                        && (ClrTypeUtilities.AreSame(iface.ClrType, targetClrInterface)
+                            || ClrTypeUtilities.ImplementsInterfaceByName(iface.ClrType, targetClrInterface))))
+                {
+                    return true;
+                }
             }
 
             // Issue #319: when a GSharp class carries a CLR backing (because it
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -117,6 +117,11 @@ public sealed partial class Evaluator
 
     private object EvaluateLiteralExpression(BoundLiteralExpression n)
     {
+        if (n.Type == TypeSymbol.String)
+        {
+            return string.Intern((string)n.Value);
+        }
+
         return n.Value;
     }
 
--- a/src/Core/CodeAnalysis/StructValue.cs
+++ b/src/Core/CodeAnalysis/StructValue.cs
@@ -113,6 +113,9 @@ public sealed class StructValue
     /// </summary>
     public object ClrBacking { get; set; }
 
+    /// <summary>Gets a value indicating whether this value represents a CLR box.</summary>
+    internal bool IsBoxed { get; private set; }
+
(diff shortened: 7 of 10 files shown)


DavidObando:
# Review: PR #3195 — `fix/3173-work` @ `f2699263`

**Verdict: MERGEABLE**, with one SHOULD-FIX that I recommend tracking as a follow-up rather than blocking on.

Independent review. Live head taken from `gh pr view --json headRefOid` = `f2699263`; one commit, merge-base `16f97229`; `main` has since moved to `917677ec` (one test-only commit, #3189, in an unrelated Interpreter file). No force-push, no orphaned base.

---

## 1. Which side is correct? Answered from the spec, not from majority vote

The brief asked that the PR prove which answer is right rather than assume emit. It does, and the record pre-dates the PR:

- `docs/adr/0045-object-universal-upper-bound.md:41` — `==`/`!=` on two `object` operands resolve to **reference equality (CLR `ceq`)**.
- `docs/adr/0045-object-universal-upper-bound.md:42` — *"Equality between an `object`-typed operand and a value-typed operand: the value operand is boxed and reference equality is used. (This is the C# rule for `object == int`.)"*

So `False` was the specified answer all along, the emitter implemented it, and the evaluator did not. This is a **conformance fix, not a tie-break**.

## 2. Did the oracle move? No — measured, and structurally explained

This is the check that matters most, because `BoundBinaryOperator.cs` feeds *both* back-ends. I ran five probe programs through emit on the PR tree and on the tree with the product hunk reverted:

| probe | emit @ head | emit @ reverted |
|---|---|---|
| `object == 42` | `False` | `False` |
| string literals / computed | `True\|False` | `True\|False` |
| boxed enum | `False\|True\|True` | `False\|True\|True` |
| boxed struct | `False\|True\|True` | `False\|True\|True` |
| `object` vs `nil` | `False\|True` | `False\|True` |

**Identical in every cell.** The structural reason: the `IsReferenceEquality` *property* has exactly two consumers repo-wide, both in `src/Core/CodeAnalysis/Evaluator.Operators.cs:177,179`. The emitter decides reference equality through a different mechanism (`IsReferenceEqualityOperand`, a static type test at `MethodBodyEmitter.Operators.cs`), which this PR does not touch. Blast radius is two evaluator lines.

`ilverify` gate (#3164): **123/123 verified, 0 verification failures, 0 compile failures, rc=0.**

## 3. The riskiest hunk — flipping an existing assertion — is legitimate

`test/Core.Tests/CodeAnalysis/Binding/Issue1923PatternMatchingSeamsTests.cs:59` changes `Assert.Equal(true, …)` to `Assert.Equal(false, …)` for `let answer object = 42; answer == 42`. Changing an existing test to make a fix pass is normally where a review dies, so I measured it rather than reasoning about it:

**Emit printed `False` for that exact program on the unfixed tree too.** The old assertion was pinning the evaluator's bug against an oracle that already disagreed, and ADR-0045:42 names this precise shape (`object == int`). The flip is a correction, not an accommodation.

## 4. Mutation matrix — per hunk, two-sided, plus a control

All mutants build-green, asserted applied via `git diff HEAD`, Core.dll md5 movement recorded, restored by exact md5 round-trip (source `f407938d…`, Core.dll `7a55c222…`, both exact on return).

| mutant | new test (6) | note |
|---|---|---|
| **M1** revert both `object` arms to `main` | **3 RED** (`objectStrings`, `objectClass`, `objectBoxedInts`), 3 controls GREEN | reproduces the author's report exactly |
| **M3** `==` arm only → `isReferenceEquality: false` | **3 RED** | equality arm independently pinned |
| **M2** `!=` arm only → `isReferenceEquality: false` | **3 RED** | inequality arm independently pinned |
| **M4** *control* — give the `string` arm `isReferenceEquality: true` | **exactly 1 RED** (`typedStrings`), other 5 GREEN | **the over-correction controls are live and selective, not decorative** |

M4 is the one I'd highlight: the issue demanded over-correction controls, and this proves they can actually fail, and fail only on the right specimen.

I also confirmed the pins are the *sole* pin: under M1 the pre-existing suites stay green, which is what made this bug shippable in the first place.

**Methodology note for other reviewers:** `git checkout <ref> -- <path>` **stages** the change, so plain `git diff` returns empty and looks exactly like an unapplied mutant. Only `git diff HEAD` (and the DLL md5 movement) revealed M1 had applied. Also, Core.dll md5 is commit-sensitive via Nerdbank.GitVersioning — md5 round-trips are only meaningful at a fixed commit.

## 5. Test quality — ADR-0154 compliant, and it drives the right engine

- `Issue2896StructObjectOverrideTests.cs:150` anchors on a **literal** expected string from `InlineData` (`Assert.Equal(expected + "\n", emitted)`) *before* chaining the other drivers to it. Driver agreement is chained to a fixed value, not asserted in a vacuum.
- The `a == c` (same instance) cell expects **`True`** in all three pins — an anti-over-correction cell inside the pin itself, so "always false" cannot pass.
- Crucially, it drives the **real** remaining evaluator surfaces in-process: `Compilation.Evaluate` and `SessionEngine` (`:651`). Post-#3177/#3182 that matters, per below.

---

## Findings

### SHOULD-FIX — two adjacent shapes now diverge that previously agreed
`src/Core/CodeAnalysis/Evaluator.Operators.cs:177`

Probing the neighbourhood found the fix nets out strongly positive but is not uniform. Comparing evaluator (`Compilation.Evaluate` and `SessionEngine` — identical in every cell) against emit:

| specimen | cell | emit | eval @ main | eval @ head |
|---|---|---|---|---|
| two `object` locals = `"hello"` literal | `a == b` | `True` | `True` ✅ | **`False`** ❌ new |
| ″ (literal vs `String.Concat`) | `a == b` | `False` | `True` ❌ | `False` ✅ fixed |
| boxed struct `a`,`b` distinct | `a == b` | `False` | `True` ❌ | `False` ✅ fixed |
| boxed struct, `let c object = a` | `a == c` | `True` | `True` ✅ | **`False`** ❌ new |
| boxed struct | `a != b` | `True` | `False` ❌ | `True` ✅ fixed |
| boxed enum | 2 cells | — | ❌❌ | ✅✅ fixed |
| `object == 42` | — | `False` | `True` ❌ | `False` ✅ fixed |

Net: **6 divergent cells → 2. All six originals are fixed; two new ones appear.** Both new ones are cells where emit says `True` (same reference) and the evaluator now says `False`, and both violate ADR-0045:41 just as the original bug did.

Repro:
```gsharp
let lit object = "hello"
let lit2 object = "hello"
Console.WriteLine(lit == lit2)     // emit True, evaluator False
```
```gsharp
struct Pt { var X int32 }
let a object = Pt{X: 1}
let c object = a
Console.WriteLine(a == c)          // emit True, evaluator False
```

Root cause is one thing, not two: the evaluator's `ReferenceEquals` compares **host-CLR** representations whose identity model differs from the emitted program's — no `ldstr` literal interning, and user-struct values are copied on assignment rather than sharing a box. Note the inconsistency this creates inside the PR's own fixed surface: `a == c` aliasing is preserved for boxed **ints** and **classes** (both pinned and passing) but not for user **structs**.

Why I am not calling this a blocker: the value-equality path was previously returning `True` here for the *wrong reason* and agreeing with the oracle by luck; the PR strictly reduces divergence; and a real fix needs an evaluator boxing/interning identity model, which is a much larger change than this 14-line hunk and squarely outside issue #3173. **Recommend a follow-up issue with the two repros above rather than expanding this PR.** These two shapes should *not* be pinned at their current values — that would pin the wrong answer.

### NIT — driver labels overstate engine diversity
`test/Interpreter.Tests/Issue2896StructObjectOverrideTests.cs:807,809`

`"gsc-evaluate" => RunCompilerEvaluation(...)` calls bare `gsc`, and `"gsi" => RunInterpreter(...)` calls script-mode `gsi`. After #3177 and #3182 **both execute through the emitter**. So of the five paths the new test compares, three are the same engine as the oracle. The test is still correct — the author added `Compilation.Evaluate` and `SessionEngine` precisely because of this, and those two are what actually discriminate — but the labels will cause a future reviewer to over-count driver coverage. Suggest renaming to reflect what they now exercise.

### Not a gap: generics
`==` over an unconstrained `T` is rejected at bind time (`GS0129`), and `Eq[object](...)` is rejected by the constraint (`GS0152`, `object` does not satisfy `comparable`). The generic neighbourhood is closed by construction, not merely untested.

### Not a gap: `nil`
`let a object = nil` is rejected outright (`GS0155`, pre-existing). `object`-vs-`nil` comparison on a non-null value agrees across emit and both evaluator paths, before and after.

---

## Denominator

**9 test projects**, enumerated from `dotnet sln GSharp.sln list` (the same source `build/generate-ci-test-matrix.py:26` uses).

**RAN unfiltered — 8 of 9:** Core 7128 passed / 1 skipped · Interpreter 1399 · Cs2Gs 1907 · LanguageServer 462 · Extensions 104 · Sdk 59 · GeneratorHost 31 · InternalAnalyzers 9 → **11,099 passed, 0 failed**, every project rc=0. Failures counted with a pattern that cannot match xUnit display names.

Every one of these eight matches the author's table exactly.

**`Compiler.Tests` — NOT run in full** (~1h18m). I did not accept a blanket containment argument: I measured it, and containment is **not** total — 8 files under `test/Compiler.Tests/Emit/` do call `.Evaluate(`. I ran those targeted: **89/89 passed, rc=0**. The remainder of the project is emit/codegen, which is structurally untouched (the emitter never reads the changed flag) and is additionally covered by the green 123/123 `ilverify` gate over the sample corpus. Author reports 4260 there; that figure is unverified by me.

## Merge against current `main`

`git merge-tree --write-tree` output is a single tree OID with zero conflict output. Because a clean text merge proves nothing semantic, I built the **real merge** of `f2699263` into `917677ec` and ran it: build clean, **Interpreter.Tests 1400/1400** (the +1 is #3189's added case), `Issue1923` 8/8.

Worktree left byte-clean; all mutations restored by exact md5 round-trip; the pre-existing #3170 `packages.lock.json` removed before claiming cleanliness.

DavidObando:
Round 2 is pushed at `e1b3fdad`.

Root cause was identity preservation, not equality: evaluator literals were not interned and boxed `StructValue` aliases were copied/re-boxed. Both reported regressions now match emit (`True`/`True`), all seven round-1 cells remain fixed, final tests are 5/13 RED on `f2699263`, all seven product/control mutants die, and ILVerify is 123/123. Full details and the nine-cell table are in the updated PR body.

DavidObando:
Round-2 rebase update: branch is now `d8e0c236` on `main` `20fbbec4`. #3232 landed during validation and introduced boxed-value equality for mixed `object`/value operands, contradicting ADR-0045:41-42 and turning the required `object == 42` control back to `True`. This branch removes only that special path; #3232’s unrelated nil/`!!` fixes remain. Final matrix 13/13, Issue1923 8/8, Issue3224 emit+ILVerify 1/1, ILVerify gate 123/123. Restoring the #3232 product path kills both controls (Core 1 RED, emitted 1 RED) with exact source/DLL hash round-trip. CI: 30 passed, 2 skipped.

DavidObando:
Rebased to `eb06ce81d` on current `main` `5cd0d766a`, including #3242 Phase 3b.3 evaluator retirement changes. Bulletin does not change result: evaluator assertions use `Compilation.Evaluate` and `SessionEngine` in-process; no `gsi --engine evaluator <file>` measurement was used. Current-head validation: Release build clean; exact issue #3173 matrix 13/13; Issue2896 class 59/59; Issue1923 8/8; Issue3224 1/1; ILVerify 123/123. Fresh CI is running.

DavidObando:
## Rubber-duck round 2 — **MERGEABLE** (0 BLOCKER, 0 SHOULD-FIX, 2 NIT)

Head `d8e0c236` (matches live `headRefOid`). Note `main` has moved to `5cd0d766` — the briefed `e1498735` is an ancestor, so no revalidation needed. Merge base `20fbbec4`, 5 behind / 3 ahead. `git merge-tree --write-tree` **output** = one tree line `2acab610`, `grep -c CONFLICT` = **0** (grepping `CONFLICT`, not `<<<<<<<`). Fresh worktree `/tmp/w3195`, own `out/`, Release build rc=0 / 0 warnings, 37 `GSharp.Core.dll` copies at uniform mtime.

---

### Q1 — deleting #3232's product code. **Justified, and byte-proven narrow.**

**The ADR is not ambiguous, and it is on this PR's side.** ADR-0045 verbatim:

> **:41** `==` and `!=` on two operands typed `object` resolve to **reference equality** (CLR `ceq`).
> **:42** Equality between an `object`-typed operand and a value-typed operand: the value operand is boxed and reference equality is used. **(This is the C# rule for `object == int`.)**

Line 42 *is* the `answer == 42` case, in those words. Reinforced twice more:

> **:64** `==` on `object` is reference equality, **even when the runtime values are `int` boxes**
> **:72** *Alternatives considered* — **"Value-equality for `object == object`. Rejected** as too surprising at the CLR boundary."

#3232's `IsBoxedValueEquality` → `Object.Equals` dispatch implemented the **explicitly rejected** alternative. This PR is right; that part of #3232 was wrong.

**Is the removal narrow? Yes — proven at byte level, not by reading the diff.**
`src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Operators.cs` at `pr3195` has md5 **`f4cb63ae9ac57c065b3b2a3a6c61096b`**, identical to that file at #3232's **parent** `59ad0f29`. An exact clean revert of that file's contribution — nothing extra taken, nothing stale left. (This is the PR #3159 failure shape you flagged; it does not occur here.)

**#3232's other work is intact — and I ran it, rather than trusting green CI.** #3232 touched 9 files; this PR touches 5. Untouched: `Issue3217NilOnLeftComparisonEmitTests.cs`, `NullSafeOperatorTests.cs`, `SmartCastTests.cs`, `spec.md`. Measured on this head:

| #3232 pin (untouched) | result |
|---|---|
| `Issue3217NilOnLeftComparisonEmitTests` (nil-on-left) | **1 passed** |
| `NullSafeOperatorTests` (`!!` contract) | **12 passed** |
| `SmartCastTests` | **6 passed** |

**19 of #3232's pins green.** Its `spec.md` edit was about `!!`-on-nil (#3216) — unrelated to equality, correctly left alone.

**Do #3232's own boxed-equality tests still pass? No — they were deliberately re-pinned, and that is the honest action.** `Issue3224BoxedConstantEqualityEmitTests.cs:46` flips `True\nFalse\nFalse\nTrue\nTrue\nTrue\nbranch\n` → `False\nTrue\nFalse\nTrue\nFalse\nTrue\n`. I checked this is a genuine re-pin and not a weakening:
- the `!=` rows correctly **invert** alongside the `==` rows;
- `text == "hi"` (two literals) correctly **stays `True`** — a real must-stay-True row survives;
- the vanished `branch` line is the *same expression* as row 1 whose value flipped; a wrongly-firing branch would still append a line and fail the exact-string assert. Nothing was made vacuous.

**Bonus: this PR fixes the actual root cause of #3173**, which nobody called out. The operator table registered `object == object` via the homogeneous ctor, leaving `isReferenceEquality` at its **`false` default** directly under a `// ADR-0045: object reference equality.` comment (`BoundBinaryOperator.cs:639-655`). The PR passes `isReferenceEquality: true` explicitly. That default-vs-comment mismatch is why the evaluator said `True` while emit said `False`.

### Q2 — the rows the nine-cell table cannot see. **No over-equalization.**

Built 5 must-stay-unequal rows, each measured on **emit + `Compilation.Evaluate` + `SessionEngine`**:

| row | measured (all 3 engines agree) |
|---|---|
| **R1** two separate boxes of an equal struct value | `a == b` → **False** |
| **R2** *same variable* boxed at two conversion sites | **False** |
| **R3** box snapshot before / after mutating the source | **False**; snapshots read **5** and **99** |
| **R4** two runtime-built equal strings; and vs. the equal literal | **False**, **False** (content-compare `True`) |
| **R5** box inside a copied containing struct | `h1.Item == h2.Item` → **True**; both read **1** after mutating the original |

**No row expected `False` came back `True`.** The "one box snapshot" design does not over-share: `Box()` = `Copy()` + flag, called per conversion site, so R2 stays unequal.

**These rows are a live oracle — demonstrated, not asserted.** Two two-sided PRODUCT mutants, each applied via `git diff HEAD` **and** `GSharp.Core.dll` md5 movement:

- **M1** — `StructValue.Box()` returns `this` instead of a snapshot. md5 `3b73e57e` → **`a91aa458`**. → **R2 RED**, correct reason: `emit=False,True` but `eval=True,False` and `sess=True,False`.
- **M2** — `StructValue.Copy()` deep-copies boxed inner values (reverts the PR's hunk). md5 → **`dc663e7c`**. → **R5 RED**: `eval=False` against `emit=True`.

Restored by exact md5 round-trip both times (source `df0737f9`, artifact `3b73e57e`).

### Q3 — box identity across the copy boundary. **Correct, mutation-visible.**

R5 is the version you asked for. Copying the containing struct shares the box **by reference** (`True`), and mutating the original source afterwards leaves *both* reads at `1` — so the box is a snapshot, not an alias. Equality and contents are both right, and M2 proves the row would notice if they weren't.

### Q4 — surface reachability. **Confirmed empirically.**

The PR's tests contain **no** `--engine evaluator` / `GSI_ENGINE` (grep empty), so the #3230 trap is avoided. They drive `Compilation.Evaluate` and `new SessionEngine{…}.Evaluate` in-process. I did the same — and under M1, **both** the evaluator and `SessionEngine` columns diverged while emit held. `SessionEngine` is a genuine third evaluator-backed column, not a second emit column.

### Q5 — string interning. **Literal-only, and measured.**

`string.Intern` fires solely in `EvaluateLiteralExpression` (`Evaluator.Expressions.cs:117`), gated on `n.Type == TypeSymbol.String`, so only **literal** nodes reach it. Computed strings are call expressions and never do. R4 measures it: two runtime-built equal strings are unequal to each other **and** to the equal literal, while comparing content-equal. That matches `ldstr`; the evaluator does not become more equal than emit, and there is no intern-table growth beyond what emit already causes.

### Suites / CI

`Interpreter.Tests` **1501 passed, 4 skipped, 0 failed** locally. CI rollup length **32**, all COMPLETED — 30 SUCCESS + 2 SKIPPED (`publish`, `publish-visual-studio-extension`); **`ilverify` SUCCESS**. Nothing pending.

---

## NIT-1 — evaluator cannot unbox a user struct (pre-existing, **not scored against this PR**)

`Pt(a)` where `a` is `object` throws `EvaluatorException: "Unexpected type Pt"` at `Evaluator.Calls.cs:699`, while **emit executes it correctly**. ADR-0045:35 makes `T(expr)` unboxing a language feature, so this is a live emit-vs-evaluator divergence.

It is **not** caused by this PR: the diff changes only the *body* of the reference-upcast arm and removes **no** conversion arm — there was never an arm for `object → user value struct`. Flagging only because it blocked two of my probes and may deserve its own issue (@DavidObando's call); the evaluator's Phase 3c removal may moot it.

## NIT-2 — optional

This flips a user-visible result (`answer == 42` was `True` as of #3232 this morning, now `False`). ADR-0045:64 asks for this to be "called out" in user docs; `spec.md` documents no `object` equality semantics today. Pre-existing gap, cheap to close in passing, not worth blocking.

---

## My own near-misses, disclosed

1. **A red that was not data.** My first probe run showed 2 failures I could have scored against the PR. The diagnostics rendered as `"evaluator failed:"` with **nothing after it** — a no-result wearing a red. Only after switching the rendering from `d.ToString()` to `d.Message` did I get `Unexpected type Pt`, which turned out to be a pre-existing gap plus my own use of an unsupported construct.
2. **An empty count that read like a pass.** Filtering `Core.Tests` for `Issue3217NilOnLeftComparisonEmitTests` returned an **empty** passed-count. It lives in `Compiler.Tests`. An empty cell next to "12 passed / 6 passed" reads as benign; it was a no-result. Re-ran in the correct project, where it passed. `No test matches` count was 0 on every run I scored.

Worktree left at `d8e0c236`, `git status --porcelain` empty, shared checkout never touched.

Issue #3199, opened 2026-08-04
[diagnostic] Interactive storage boundaries still surface as GS9999 instead of GS0513


DavidObando:
## Summary

The default interactive evaluator still reports several deliberate compiled-only storage boundaries as **GS9999**, although ADR-0153 and `docs/diagnostics.md` say clean boundaries use **GS0513** and reserve GS9999 for unexpected evaluator failures.

Found while sweeping execution-engine claims for #3196. This is separate from that docs-only issue.

## Repro

Measured through an in-process `SessionEngine` harness (required because `gsi`
refuses redirected stdin), then reconfirmed after rebasing onto current `main`
`57256713f5131e46a8ca81759ae3c0eb5b2519b8`:

```gsharp
import System

unsafe {
    let values = []int32{11, 22}
    fixed p *int32 = values {
        Console.WriteLine(*p)
    }
}
```

Observed for the ADR-0153 table's legacy boundary rows:

```text
fixed                      GS9999: ... requires the CIL pinned-local emit path
stackalloc                 GS9999: ... requires the CIL localloc emit path
sizeof(Pair)               GS9999: ... requires the CIL sizeof emit path
&Method                    GS9999: ... requires the CIL ldftn/calli emit path
function-pointer invocation GS9999: ... requires the CIL ldftn/calli emit path
```

Each submission returned `error=True` and no program output. The `fixed` source through `gsi <file>` and bare `gsc` prints `11` with rc 0 because ADR-0156 Phase 1 routes both file drivers through emitted execution. The mismatch is confined to the default interactive evaluator (`SessionEngine` / `Compilation.Evaluate`).

## Contract mismatch

- ADR-0153 says `fixed` has a self-contained boundary diagnostic and later says compiled-only storage boundaries report through GS0513 while GS9999 remains a failure.
- `docs/diagnostics.md` names `fixed` among the GS0513 clean boundary sites.
- Product still throws `EvaluatorException` for these deliberate sites, which `Compilation.Evaluate` wraps as GS9999.

## Expected

Route these deliberate interactive boundaries through GS0513 (or another explicit non-internal boundary code), preserving the current self-contained messages. Do not change file-mode execution; it is correctly emitted under ADR-0156.

Related: #2956, PR #3018, ADR-0153, ADR-0156.

DavidObando:
[...]

PR #3200, opened 2026-08-04
Correct execution engine documentation

### LINKED ISSUE #3196: [docs] ADR-0068 still specifies gsi never runs deinitializers; script-mode gsi has run them since #3182 and the ADR is still marked Accepted
## Summary

`docs/adr/0068-deinit-destructor-support.md` still specifies that `gsi` never runs deinitializers. **That statement is now false**, and the ADR still carries `Status: Accepted` with no pointer to the ADR that superseded it.

This was found during the rubber-duck review of PR #3190 (issue #3181), where the reviewer was briefed the ADR's interpreter-boundary rule as a *hard constraint* on the change. The constraint turned out not to exist any more. Measured rather than inferred.

## The stale text

`docs/adr/0068-deinit-destructor-support.md:118-122` (`### Interpreter boundary`) states that the interpreter does not run deinitializers, and the ADR's behaviour table encodes the same split across its rows.

## Measurement on current `main`

Running the standard deinit fixture across drivers:

| driver | observed |
|---|---|
| `gsc` (file) | prints `deinit-11: kept` |
| `gsc /out:` | prints `deinit-11: kept` |
| `gsi <file>` (script mode) | prints `deinit-11: kept` |

**`GS0510` count: zero.** All three print the deinitializer output. Script-mode `gsi` has been running deinitializers since **#3182**.

A second measurement from the same review, driving `SessionEngine` out-of-tree across the ADR's own table rows:

- reachable object (row 3) vs. dying object + forced collection (row 1): **both `GS0510`=1, both no deinit**.
- removing the `GC.Collect()` provocation: identical result.

So **reachability does not discriminate in the interactive engine either**. The axis that actually discriminates deinit behaviour today is the **engine**, not the object's liveness — which is not what the ADR's table describes.

## Why this matters beyond documentation

This ADR is being used as a **specification** by contributors and reviewers. In the #3190 review it was quoted as a hard constraint that a change making `gsi` run deinitializers would be "a language regression." Had the reviewer trusted the ADR instead of measuring, a correct change could have been rejected on the strength of a document describing behaviour the product no longer has. An `Accepted` ADR that contradicts the shipped compiler is worse than no ADR.

Related context from a parallel review (PR #3189): **all three file drivers now emit**; the interactive `SessionEngine` is the only remaining harness that exercises the evaluator. `gsi` also refuses piped stdin (`gsi requires an interactive terminal`). Several documents and test names still describe the pre-ADR-0156 world.

## Requested work

1. **Correct `:118-122`** and the behaviour table to describe what the drivers actually do on `main` — verified by running them, not by reading the diff of #3182/#3177/#3186.
2. **Mark the superseded status explicitly.** If ADR-0156 supersedes this section, say so with a link, and update `Status:` accordingly rather than leaving `Accepted` on text that is factually wrong.
3. **Sweep for the same staleness elsewhere.** ADR-0156 landed in phases (#3177 Phase 1: file drivers emit; #3186 Phase 2: interactive submissions chain through emitted in-memory assemblies; #3182 script-mode `gsi`). Any ADR, doc, or comment asserting "the interpreter does X" or "bare `gsc` evaluates" is a candidate. Grep for the claims, then **confirm each by running the driver** — a doc that was right when written and is wrong now will read as authoritative either way.
4. **Do not change product behaviour in this issue.** If the sweep turns up a place where the *product* looks wrong rather than the *doc*, file it separately with the repro.

## Scope boundary

Documentation only. `git diff --name-only -- src/` should be **empty**. If correcting the ADR seems to require a product change, stop and report that instead — it means the drivers disagree with each other, which is a different and larger bug.

## Evidence provenance

All measurements above are from the PR #3190 review comment (`issuecomment-5176032002`), reproduced against `main` at `81e6deab`. Line numbers drift: #3182 rewrote ~35 lines of the adjacent interpreter test file, and `Compilation.cs` has since moved to `Compilation/Compilation.cs`. **Re-anchor by content, not by line number.**

DavidObando:
## Summary

- amend ADR-0068's stale `gsc`/`gsi` liveness table with the current execution-engine boundary
- mark ADR-0068 partially superseded and add the missing ADR-0156 cross-reference in both directions
- record ADR-0156 Phase 3a: interactive `gsi` now uses emitted execution by default
- mark the explicit evaluator path deprecated and scheduled for removal in Phase 3c
- remove ADR-0047's unsupported ADR-0023 interpreter-authority citation
- correct stale evaluator-boundary claims across the affected ADRs and diagnostic catalogues
- synchronize `website/docs/ref/diagnostics.md` with the authoritative diagnostics catalogue
- preserve ADR-0068's finalizer-thread risk and diagnostic-severity rationale

Closes #3196.

## Status decision

ADR-0068 is **partially superseded by ADR-0156**, not wholly superseded. Its `deinit` syntax, CLR finalizer shape, diagnostics, and emission decision remain accepted; only its original execution-driver boundary is obsolete. This follows existing scoped-supersession precedent (ADR-0017 and ADR-0033) rather than inventing a new status. ADR-0156 names the ADR-0068 boundary it supersedes.

## Final remeasurement

Final base: `0484b00f139fe216b7910a23bb08415d571e2a28`. Head: `24ced1162b6aa37cf266cbdb206b26af85477192`.

Rebuilt Release `gsi --help` returned rc 0 and reports:

```text
--engine <name>        Interactive engine: 'emit' (default) or 'evaluator'
                       (also via GSI_ENGINE). 'evaluator' selects the legacy
                       tree-walking engine; it is deprecated and will be
                       removed in ADR-0156 Phase 3c.
```

Targeted engine/deinitializer tests passed **13/13** on the final rebased tree. They cover:

- unset interactive choice -> emitted engine
- explicit `emit` -> emitted engine
- `--engine evaluator` and `GSI_ENGINE=evaluator` -> evaluator escape hatch
- bare `gsc`, `gsc /out:`, and file-mode `gsi` running emitted deinitializers without GS0510
- `EmittedSessionEngine` running a real CLR finalizer
- explicit `SessionEngine` omitting the deinitializer and reporting GS0510 regardless of reachability

This establishes execution engine, not object liveness, as the boundary. Every default driver emits; only explicit evaluator selection and direct `Compilation.Evaluate` consumers retain the deprecated evaluator behavior.

## Measured versus inferred

Measured on the final rebased tree: rebuilt CLI help text and rc, engine selection, deinitializer output and diagnostics through the 13 targeted tests, documentation build, link resolution, residue greps, diff scope, and PR mergeability.

Documentation-policy decision: ADR-0068 remains accepted except for the execution-engine boundary, which is partially superseded by ADR-0156.

## Validation

- Release `src/Repl/Repl.csproj` build with warnings as errors: **0 warnings, 0 errors**.
- Targeted `Interpreter.Tests`: **13/13 passed**.
- Local Docusaurus production build: passed.
- Current-head Docs site CI `Build docs`: passed.
- Relative link check: **27 real relative links across 19 changed Markdown files, 0 broken**. The post-rebase count is smaller because current `main` already contains several formerly changed documentation files.
- Stale default-model, old three-driver-model, and phantom-rule residue greps: 0 hits.
- `git diff --check`: clean.
- No mutation testing: no executable code changed.
- Full test suites not rerun for this docs-only round; targeted tests cover every behavioral claim retained in the changed text.

## Scope

```text
$ git diff --name-only $(git merge-base origin/main HEAD)...HEAD | grep -E '^(src|test|e2etests|build|samples)/'

$ git diff --name-only -- src/

```

Both outputs are empty. Product code unchanged.

### DIFF
--- a/docs/adr/0022-go-chan-select-lowering.md
+++ b/docs/adr/0022-go-chan-select-lowering.md
@@ -11,7 +11,8 @@ Per ADR-0002 (D2) GSharp adopts a **synthesis** concurrency model: a Go-shaped s
 
 At the time of this decision, the interpreter was the project's semantic
 oracle. ADR-0156 later moved production file drivers and the conformance oracle
-to emitted execution. This ADR still documents both backends:
+to emitted execution. The evaluator backend is deprecated and scheduled for
+removal in ADR-0156 Phase 3c. This ADR still documents both backends:
 
 1. The Go-shaped surface the user sees.
 2. The .NET surface the interpreter (and, later, the emitter) lowers to.
--- a/docs/adr/0022-go-chan-select-lowering.md
+++ b/docs/adr/0022-go-chan-select-lowering.md
@@ -9,7 +9,9 @@
 
 Per ADR-0002 (D2) GSharp adopts a **synthesis** concurrency model: a Go-shaped surface (`go`, `chan T`, `<-`, `select`) lowered onto .NET primitives, plus first-class `async`/`await` for direct BCL interop. This ADR fixes the lowering targets and the visible semantics that fall out of those choices, so item-by-item work in 5.3 – 5.7 has a single agreed contract.
 
-The interpreter is the source of truth (per the cross-cutting "authoritative-semantics rule"). This ADR therefore documents both:
+At the time of this decision, the interpreter was the project's semantic
+oracle. ADR-0156 later moved production file drivers and the conformance oracle
+to emitted execution. This ADR still documents both backends:
 
 1. The Go-shaped surface the user sees.
 2. The .NET surface the interpreter (and, later, the emitter) lowers to.
--- a/docs/adr/0040-sequence-type-and-yield.md
+++ b/docs/adr/0040-sequence-type-and-yield.md
@@ -112,7 +112,8 @@ This reuses architectural patterns from the async state-machine pipeline (`Synth
 3. **Interpreter parity**: the interpreter (`Evaluator.cs`) does not gain
    `yield` support in this slice. Iterator functions are emit-only. This is a
    known evaluator gap tracked as #138; ADR-0156 later made emitted execution
-   the production oracle and routed file-mode drivers through it.
+   the production oracle and routed every default driver through it. The
+   evaluator is deprecated and scheduled for removal in Phase 3c.
 
 ## Alternatives considered
 
--- a/docs/adr/0040-sequence-type-and-yield.md
+++ b/docs/adr/0040-sequence-type-and-yield.md
@@ -109,7 +109,10 @@ This reuses architectural patterns from the async state-machine pipeline (`Synth
 
 1. **Async iterator alias**: `IAsyncEnumerable[T]` support shipped in #128 but no `asyncSequence[T]` alias exists yet. ADR-0041 proposes overloading `sequence[T]` in `async` return-type position to mean `IAsyncEnumerable[T]`, in lieu of a separate keyword.
 2. **Try/finally in iterators**: this slice does not support `try`/`finally` within iterator bodies (would require `Dispose()` to resume into finally blocks). A diagnostic is emitted if detected.
-3. **Interpreter parity**: the interpreter (`Evaluator.cs`) does not gain `yield` support in this slice. Iterator functions are emit-only. This is a known gap tracked as #138 — the interpreter is authoritative per `docs/emit-pipeline.md` but iterator state machines are inherently an emit concern.
+3. **Interpreter parity**: the interpreter (`Evaluator.cs`) does not gain
+   `yield` support in this slice. Iterator functions are emit-only. This is a
+   known evaluator gap tracked as #138; ADR-0156 later made emitted execution
+   the production oracle and routed file-mode drivers through it.
 
 ## Alternatives considered
 
--- a/docs/adr/0047-attribute-syntax-and-declaration.md
+++ b/docs/adr/0047-attribute-syntax-and-declaration.md
@@ -165,7 +165,8 @@ The interpreter (`Evaluator`) gains a parallel "attribute table" keyed by GSharp
 This originally followed the project's interpreter-authority convention.
 ADR-0156 later made emitted execution the production oracle; evaluator parity
 remains relevant only while direct `Compilation.Evaluate` consumers and the
-deprecated `gsi --engine evaluator` compatibility path remain.
+deprecated `gsi --engine evaluator` compatibility path remain. Phase 3c
+removes the evaluator.
 
 ### 10. Grammar additions (summary)
 
--- a/docs/adr/0047-attribute-syntax-and-declaration.md
+++ b/docs/adr/0047-attribute-syntax-and-declaration.md
@@ -162,7 +162,10 @@ Extension functions (ADR-0019) lower to a static method on `<Extensions>` with t
 
 The interpreter (`Evaluator`) gains a parallel "attribute table" keyed by GSharp symbol. Each binding records the resolved attribute type, the constructor arguments (as runtime `object?[]`), and the named-arguments dictionary. The interpreter exposes them through the same reflection-shaped facade it already uses for `MethodInfo` / `Type` mirrors — `GetCustomAttributes(typeof(T))` on an interpreter symbol returns the same materialised list the emit path would surface to a runtime caller. Compiler-recognised attributes (§6) are honoured at interpretation time identically to the compiled path: `[Conditional("DEBUG")]` elision, `[Obsolete]` diagnostics, `[EnumeratorCancellation]` threading.
 
-This keeps the ADR-0023 rule that the interpreter is the authoritative semantics for everything the language exposes.
+This originally followed ADR-0023's interpreter-authority rule. ADR-0156 later
+made emitted execution the production oracle; evaluator parity remains relevant
+only while `Compilation.Evaluate` consumers and the default interactive engine
+remain.
 
 ### 10. Grammar additions (summary)
 
--- a/docs/adr/0047-attribute-syntax-and-declaration.md
+++ b/docs/adr/0047-attribute-syntax-and-declaration.md
@@ -162,10 +162,10 @@ Extension functions (ADR-0019) lower to a static method on `<Extensions>` with t
 
 The interpreter (`Evaluator`) gains a parallel "attribute table" keyed by GSharp symbol. Each binding records the resolved attribute type, the constructor arguments (as runtime `object?[]`), and the named-arguments dictionary. The interpreter exposes them through the same reflection-shaped facade it already uses for `MethodInfo` / `Type` mirrors — `GetCustomAttributes(typeof(T))` on an interpreter symbol returns the same materialised list the emit path would surface to a runtime caller. Compiler-recognised attributes (§6) are honoured at interpretation time identically to the compiled path: `[Conditional("DEBUG")]` elision, `[Obsolete]` diagnostics, `[EnumeratorCancellation]` threading.
 
-This originally followed ADR-0023's interpreter-authority rule. ADR-0156 later
-made emitted execution the production oracle; evaluator parity remains relevant
-only while `Compilation.Evaluate` consumers and the default interactive engine
-remain.
+This originally followed the project's interpreter-authority convention.
+ADR-0156 later made emitted execution the production oracle; evaluator parity
+remains relevant only while direct `Compilation.Evaluate` consumers and the
+deprecated `gsi --engine evaluator` compatibility path remain.
 
 ### 10. Grammar additions (summary)
 
--- a/docs/adr/0068-deinit-destructor-support.md
+++ b/docs/adr/0068-deinit-destructor-support.md
@@ -115,7 +115,7 @@ changed which invocations use that backend. Current routing is:
 | `gsc /out:program.dll file.gs`, then run the assembly | emit to file, then CLR host | real CLR finalizer; no GS0510 |
 | `gsi file.gs` | emit to memory, then `EmittedProgramHost.Run` | real CLR finalizer; no GS0510 |
 | interactive `gsi` (default) | `EmittedSessionEngine` | real CLR finalizer; no GS0510 |
-| interactive `gsi --engine evaluator` | `SessionEngine` → `Compilation.Evaluate` | skips the body; GS0510 once per declaring class |
+| interactive `gsi --engine evaluator` (deprecated; removed in Phase 3c) | `SessionEngine` → `Compilation.Evaluate` | skips the body; GS0510 once per declaring class |
 
 The old two-column `gsc`/`gsi` table made object reachability appear to be the
 discriminating axis. It is not. Driving the legacy evaluator-backed
@@ -173,4 +173,5 @@ Two new diagnostics are allocated:
 
 The `deinit` language and emission design remain accepted. ADR-0156 partially
 supersedes only this ADR's original driver-boundary model: all default drivers
-now emit, while the evaluator compatibility path retains GS0510.
+now emit, while the deprecated evaluator compatibility path retains GS0510
+until its Phase 3c removal.
--- a/docs/adr/0068-deinit-destructor-support.md
+++ b/docs/adr/0068-deinit-destructor-support.md
@@ -114,11 +114,11 @@ changed which invocations use that backend. Current routing is:
 | bare `gsc file.gs` | emit to memory, then `EmittedProgramHost.Run` | real CLR finalizer; no GS0510 |
 | `gsc /out:program.dll file.gs`, then run the assembly | emit to file, then CLR host | real CLR finalizer; no GS0510 |
 | `gsi file.gs` | emit to memory, then `EmittedProgramHost.Run` | real CLR finalizer; no GS0510 |
-| interactive `gsi` (default) | `SessionEngine` → `Compilation.Evaluate` | skips the body; GS0510 once per declaring class |
-| interactive `gsi --engine emit` | `EmittedSessionEngine` | real CLR finalizer; no GS0510 |
+| interactive `gsi` (default) | `EmittedSessionEngine` | real CLR finalizer; no GS0510 |
+| interactive `gsi --engine evaluator` | `SessionEngine` → `Compilation.Evaluate` | skips the body; GS0510 once per declaring class |
 
 The old two-column `gsc`/`gsi` table made object reachability appear to be the
-discriminating axis. It is not. Driving the default interactive
+discriminating axis. It is not. Driving the legacy evaluator-backed
 `SessionEngine` across the old rows produces the same boundary regardless of
 liveness:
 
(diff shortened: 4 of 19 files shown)


DavidObando:
## Round-1 review — PR #3200 (`fix/3196-work`, head `aa983193`)

**Verdict: MERGEABLE — 0 BLOCKER, 2 SHOULD-FIX, 1 NIT.**

Every driver claim in this PR is **measured true**. The findings below are a retained false citation, a stale published mirror, and a deleted rationale — none of them wrong-about-the-product.

### Method

Docs read at `aa983193`. Product **measured** at `main` `3ca776d3` (= post-merge `src/`, since this PR changes no product code; the one commit `3ca776d3` adds over the PR base `57256713` is `#3188`, unrelated to drivers). Fresh worktree, own `out/` root, Release build 0 warnings / 0 errors.

Suites: **not** 9 of 9 — docs-only, per instruction. Ran 4 filtered subsets of **1 of 9** projects (denominator source `dotnet sln GSharp.sln list`): `Issue2988Deinit`+`EmittedSessionEngine` **45/45**, `RefLocalAliasing` **11/11**, `Issue2986InterpreterBoundary` **3/3**. No mutation testing (no executable code changed). CI **30 SUCCESS / 2 SKIPPED / 0 failures**.

### Measured routing table — all five rows CONFIRMED

Oracle: `samples/Deinit.gs` + `samples/Deinit.golden` (deinit output is the engine discriminator).

| Invocation | rc | stdout | GS0510 | Doc claim |
|---|---|---|---|---|
| bare `gsc Deinit.gs` | 0 | **== golden** (after stripping `Success.`) | 0 | ✅ emit |
| `gsc /out:Deinit.dll` → `dotnet Deinit.dll` | 0/0 | **== golden** | 0 | ✅ emit |
| `gsi Deinit.gs` | 0 | **== golden** | 0 | ✅ emit |
| interactive `SessionEngine` (in-process) | — | `body-33\n` only, deinit skipped | 1 per class | ✅ evaluator |
| interactive `EmittedSessionEngine` (in-process) | — | deinit runs | 0 | ✅ emit |

`gsi --help` from the shipped binary: `'evaluator' (default) or 'emit'` — **measured, not read**. Piped stdin → rc **1**, `gsi requires an interactive terminal`, so no shell-driven probe was mistaken for an interactive row.

**ADR-0068's new third evaluator row independently reproduced** in-process: reachable + forced collection → `made-22\nbody-33\nkept\n`, GS0510 = **1**. Engine, not liveness, is the axis. Confirmed.

**#3199 residue still holds on `3ca776d3`** (in-process `SessionEngine`):
- `fixed p *int32 = xs` → **GS9999**, msg *"'fixed' (pinning) statements are not supported in the interpreter; they require the CIL pinned-local emit path."*
- `stackalloc [4]int32` → **GS9999**, msg *"stackalloc is not supported in the interpreter; it requires the CIL localloc emit path."*
- `*p = 7` → **GS0513**

That is exactly the GS9999-vs-GS0513 split `docs/diagnostics.md:317-328` and `0153:55-79` now assert. *Not measured by me: unmanaged `sizeof`, `&Method`, function-pointer invocation (2 of 5 legacy guards measured). My own P/Invoke probe was invalid G# (`GS0005` parse error) — a no-result, not evidence; the GS0514 row rests on `Issue2986InterpreterBoundaryTests` 3/3 instead.*

### Structural checks — all pass

- `git diff --name-only 57256713 aa983193 | grep -E '^(src|test|e2etests|build|samples)/'` → **empty**. Docs-only holds.
- **Links: 41 relative links across the 19 files, 0 broken.** Independently matches the author's count. (My extractor also flagged 10 in `0084` — all G# `sequence[T](seed…)` generics, not links.)
- **ADR-0017 precedent is real**, not invented: `docs/adr/0017-method-virtuality.md:3` — *"Partially superseded by ADR-0078 … the semantics described below are unchanged."* Same shape. `0033:5` is a second instance.
- **Cross-reference is genuinely bidirectional**: `0068:3,6,167-169` → 0156; `0156:8-9,111-118` → 0068.
- **ADR-0154 substance intact.** One 3-line hunk at `:26-27`, adding *"the then-evaluator-backed"* to a `gsi` reference. Oracle-strength policy untouched. No own goal.
- Coverage claims in `0092:175`, `0093:152-158`, `0094:136-144`, `0095:197-204` corroborated: GS0514 present in `Issue758LibraryImportInterpreterTests` (5), `Issue760PInvokeRefOutInInterpreterTests` (9, plus GS0352), `Issue761PInvokeFunctionPointerInterpreterTests` (7); GS0346–0351 in `Issue759StructMarshallingInterpreterTests`.
- `0139:84-88` GS0100 claim backed by `Issue2953ProtectedReturnFallthroughInterpreterTests:43`.
- `0153:63` row *"Ref local alias — Aliases the captured storage location — Meets contract after #3032"* is **not** in the PR's measured list, but it is true: `RefLocalAliasingInterpreterTests` 11/11 through `GSharpRepl.EvaluateSubmission`. No finding.

The spot-checked odd members (0047, 0069, 0139) are correct, not pattern-matched — except the citation below.

---

## SHOULD-FIX 1 — `docs/adr/0047-attribute-syntax-and-declaration.md:165`: false ADR citation, restated by this PR

New text: *"This originally followed **ADR-0023's interpreter-authority rule**."*

**ADR-0023 is `0023-async-state-machine.md`.** It contains no interpreter-authority rule — `grep -i "authorit|source of truth|canonical"` over it returns one hit, about the `IAsyncStateMachine` contract.

Worse, the rule has **no home at all**: `git grep -iE "authoritative-semantics|interpreter-first|interpreter first"` over the whole tree at `aa983193` returns **0 hits**. The PR correctly deleted `0022:12`'s appeal to *"the cross-cutting 'authoritative-semantics rule'"* — but kept the same phantom rule alive in 0047 while rewriting that exact sentence for accuracy.

**Impact:** a docs PR whose purpose is removing unverified claims ships a citation to a document that does not say the thing. Precisely the #3196 failure mode.

**Fix (1 line):** drop the citation — *"This originally followed the project's interpreter-authority convention."* — or cite the real source if one is found.

## SHOULD-FIX 2 — `website/docs/ref/diagnostics.md:309-315`: the declared mirror still carries the corrected falsehood

`website/docs/ref/diagnostics.md:9` states: *"This page **mirrors** the authoritative diagnostic catalogue in `docs/diagnostics.md`."* There is no sync script — it is hand-maintained.

The PR corrects `docs/diagnostics.md:317-328`. The mirror still reads:

> The clean boundary sites for `fixed`, `stackalloc`, unmanaged `sizeof`, and function pointers currently surface through **`GS0513`**…

**The author measured that to be false** — those are GS9999, which is why #3199 exists. I reproduced it above. The published page now asserts the very claim this PR removed from the source.

Wider residue on the same surface, all describing bare `gsc` as interpreting — **false on `3ca776d3`**, proven by the golden-identical `Deinit.gs` run above:

- `website/docs/getting-started/quickstart.md:49` — *"the compiler uses interpreter compatibility mode"*; `:62` — *"printed by `gsc` after interpreter execution completes"*
- `website/docs/faq.md:71-73` — *"If you invoke it without `/out:`, it interprets the program in-process"*
- `website/docs/getting-started/install.md:89` — *"omitting `/out` runs the interpreter compatibility path"*
- `website/docs/intro.md:29-33` and `website/docs/bridges/gsharp-for-csharp-developers.md:142` — *"The interpreter path does not support that conversion"* (G# lambda → CLR delegate)

That last one I measured directly. `nums.Where(func(n int32) bool { return n % 2 == 0 }).Count()` under **bare `gsc`** → rc **0**, prints **`2`**. The capability the page says is unavailable is available.

**Impact:** `docs/` is now clean (I grepped it — see below) but the **user-facing site** still teaches the falsehood #3196 was filed to kill. The sweep's completeness claim is true only within `docs/`.

**Fix (ponytail):** port the one corrected paragraph into `website/docs/ref/diagnostics.md:309-315` — that is the declared-mirror obligation and it is ~8 lines. The five prose pages are a larger sweep and are reasonable to track separately rather than expand this PR again.

*Sweep completeness within `docs/` — no residue found.* `git grep -iE "interpreter is (the )?(authoritative|canonical|the source of truth)|authoritative-semantics|interpreter is treated as authoritative|canonical reference for language semantics"` over `docs/` at `aa983193` → **0 hits**. Remaining `docs/` matches for *"in the interpreter"* are Evaluator-as-component design context or verbatim product message strings, both legitimate.

## NIT — `docs/adr/0068-deinit-destructor-support.md:107-135`: true, still-applicable rationale deleted without relocation

Two removed passages were not false and were not moved to *Considered alternatives* (I checked that section — it retains only the syntax/naming alternatives):

1. *"Matching CLR behavior requires tying finalization to `StructValue` lifetime and re-entering `Evaluator` from a finalizer thread, risking shutdown deadlock and timing-dependent output that golden-sample comparison cannot pin."* — the **why-not** for the evaluator boundary.
2. *"GS0510 is a warning because … In contrast, unsafe storage and P/Invoke boundaries are errors because interpreted execution cannot proceed correctly…"* — the **severity asymmetry** across GS0510 / GS0513 / GS0514.

The evaluator is still the shipped interactive default, so both still apply. The replacement text states what the boundary is, not why it was chosen. An ADR is the wrong document to lose that from. **Fix:** re-add the two sentences under the amended section.

---

Worktree cleaned to empty `git status --porcelain`; shared checkout not written.

DavidObando:
## Round-2 review — PR #3200, head `0f13dfaa`

**Verdict: MERGEABLE — 0 BLOCKER, 0 SHOULD-FIX, 0 NIT.** All three round-1 findings resolved; the post-#3201 default flip is correctly propagated. Bounded re-review, four questions only.

**Method.** Fresh worktree at `main` `0484b00f1`, own `out/` root, Release **0 warnings / 0 errors**. Docs read at `0f13dfaa`; product measured from that build's own binaries. Merge-base `ad43a91a`; the 2 commits `main` has since are #3225/#3228 (`Core.Tests` EmittedOracle migration) and touch **no `src/`** (`git diff --name-only ad43a91a origin/main -- src/` → empty), so no driver drift under the review.

**Docs-only invariant holds.** 19 files, all `docs/` or `website/`; `grep -E '^(src|test|e2etests|build|samples|tools)/'` → **rc 1**.

---

### Q1 — driver-model sweep: COMPLETE, no surviving or newly-introduced false claim

I did not take the default flip from the PR text or from the brief. I measured it from the binary I built:

```
  --engine <name>        Interactive engine: 'emit' (default) or 'evaluator'
                         (also via GSI_ENGINE). 'evaluator' selects the legacy
                         tree-walking engine; it is deprecated and will be
                         removed in ADR-0156 Phase 3c.
```

Second, independent product oracle: `Adr0156DefaultEngineFlipTests` asserts `ReplProgram.UsesEvaluatorEngine(null) == false` — the driver's *own* selection function, not a restatement of the help text. **58/58 passed** across `Adr0156DefaultEngineFlip` + `Issue2988Deinit` + `EmittedSessionEngine` + `Issue3004PointerBoundary`.

**`docs/adr/0156-...md:103` and `:112` match reality.** #3201 verified real and MERGED (`515ad40d`, "Flip interactive gsi default to the emitted session engine (ADR-0156 Phase 3a)").

Concept sweeps at `0f13dfaa` over `docs/` + `website/docs/` (`versioned_docs` excluded; each fired with a positive control first, since an empty `git grep -E` is indistinguishable from a quoting failure):

| Sweep | Pattern | Result |
|---|---|---|
| A | evaluator/interpreter/tree-walking asserted as **default** (either word order) | **0 hits** |
| B | `interprets` / "the interpreter runs\|executes" / "interpreter compatibility mode\|path" | 3 hits, **all out of scope** |
| C | `gsi` described as interpreting/evaluating | 21 hits, **all correctly qualified** as `--engine evaluator` + deprecated |

Sweep B's hits are `website/docs/getting-started/install.md:89` and `quickstart.md:49` — the #3207 pages, excluded by instruction — plus `website/docs/ref/spec.md:1339`, which says *"the interpreter executes async iteration by blocking"*: a true statement about the `Evaluator` component with **no default claim**, in a file this PR does not touch. Not scored.

Round-2 delta on the ADRs with no business in an engine sweep (`0022`, `0040`, `0069`, `0084`, `0139`, `0154`) is uniformly "move the default marker + add the Phase 3c deprecation note" — nothing behavioural invented. `0069:332` correctly stops calling the evaluator the interactive default; `0139:84-88` correctly demotes `SessionEngine` to "the deprecated evaluator compatibility path". **ADR-0154's delta is one appended deprecation sentence — oracle-strength policy still untouched.**

File-driver re-measurement on this build, `samples/Deinit.golden` as oracle: bare `gsc` rc **0**, **byte-identical** to golden after stripping `Success.`, GS0510 = **0**; `gsi D.gs` rc **0**, **byte-identical**, GS0510 = **0**.

**Links: 27 relative links across the 19 files, 0 broken.** Lower than round 1's 41 because the file set changed, not because links were lost: `docs/emit-pipeline.md` left the set (its fix is already on `main` — `origin/main:docs/emit-pipeline.md:160` carries the Phase 3a text) and `website/docs/ref/diagnostics.md` joined it using absolute GitHub URLs. My one apparent break was my own resolver failing to normalise `docs/./internal-analyzers.md`; the file exists, and the line is untouched by this PR.

CI: **31 SUCCESS / 3 SKIPPED / 0 failing.**

---

### Q2 — ACCEPT. And my round-1 finding was partly wrong; the correction favours the author

New `0047:165-169` text is accurate, and more accurate than my round-1 finding implied.

I wrote that the interpreter-authority rule *"has no home at all"*, on the strength of `git grep -iE "authoritative-semantics|interpreter-first"` → 0. **That grep tested the label, not the concept.** The convention was real and written down — in `docs/emit-pipeline.md`, verbatim from that file's **first commit** (`0262e09a`):

> `:3` — "the interpreter remains the **canonical reference for language semantics**"
> `:160` — "if the two ever disagree, **the interpreter is treated as authoritative**"

So "the project's interpreter-authority convention" asserts a **true fact about the project's past**. What never existed was an *ADR* saying it — which is exactly why citing ADR-0023 was false and citing nothing is now correct. Dropping the citation rather than inventing one is the right resolution, and the past-tense + ADR-0156 supersession framing is accurate. No residual defect.

(Per instruction, the `ADR-0023 (async state machine)` entry in the `:6` Related list is correct and unrelated — not scored.)

### Q3 — mirror is accurate and back in sync

`website/docs/ref/diagnostics.md:309-322` and `docs/diagnostics.md:317-329` are **semantically identical**: same GS0513-vs-legacy-GS9999 split, same #3199 reference, same "applies only to direct tree evaluation and `gsi --engine evaluator` since Phase 3a", same Phase 3c removal note. The only difference is link form — relative in `docs/`, absolute GitHub URLs in the mirror — which is that page's stated convention (`:9`). **No drift in the opposite direction.**

Accuracy against measurement: **measured** — `&`/`*` → `GS0513` with the self-contained message, pinned by `Issue3004PointerBoundaryInterpreterTests:130-133` (part of the 58/58). **Inferred, not re-measured this round** — `fixed`/`stackalloc` → legacy `GS9999`; I measured those directly in round 1, and both guard sites (`Evaluator.Statements.cs:223`, `Evaluator.Expressions.cs:54`) are byte-identical between that revision and this one.

### Q4 — both passages genuinely restored, verbatim

`0068:137-142` carries both, word-for-word against the round-1 deleted text, not paraphrased:

> "Matching CLR behavior requires tying finalization to `StructValue` lifetime and re-entering `Evaluator` from a finalizer thread, risking shutdown deadlock and timing-dependent output that golden-sample comparison cannot pin. **In contrast**, unsafe storage and P/Invoke boundaries are errors because interpreted execution cannot proceed correctly without emitted storage or would fabricate a deterministic native return value."

Both still true: the evaluator survives until Phase 3c, so the why-not and the severity asymmetry both remain live. The five-row routing table at `:110-118` matches my measured `--help` and the flip test.

---

**Scope note.** I ran 4 filtered classes in **1 of 9** projects (denominator source `dotnet sln GSharp.sln list`). No mutation testing — no executable code changed. I did **not** re-derive round 1.

Worktree removed; shared checkout not written.

PR #3211, opened 2026-08-04
Resolve bundled extensions in loose IDE files

### LINKED ISSUE #3198: [language-server] Verify whether implicit extension references reach the IDE path after #3147, or whether the asymmetry moved to IDE-vs-driver
## Summary

**Verify before fixing. If it does not reproduce, close it with the measurement — that is a complete result.**

PR #3147 fixed implicit extension assembly references for the CLI drivers by making `ResolveDriverReferencePaths` unconditional. The rubber-duck review of that PR flagged — **statically, without executing the language server** — that the same reference channel may not exist on the IDE path, which would mean code that compiles and runs correctly still shows `GS0157` squiggles in the editor.

## The suspicion

Loose-file paths in the language server construct `new Compilation(syntaxTree)` with **no resolver**:

- `DocumentSyncHandler.cs:56,82`
- `HoverComputer.cs:28` … `:1548`

and `Compilation.cs:303` passes `references: null`.

**Before #3147 the editor and the drivers agreed — both failed.** After #3147 the drivers succeed. If the IDE path was not updated, the asymmetry did not disappear; it **moved from driver-vs-driver to IDE-vs-driver**, which is harder to notice and more annoying to users.

Locate by symbol, not by line — cites go stale.

## Reproduce first

Use the three samples from #3130 (Optional, Mixed, Sequences) plus a plain `import Gsharp.Extensions.*` file:

1. Confirm on current `main` that all three **compile and run clean** through `gsc <file>`, `gsc /out:` and `gsi <file>` — this is the post-#3147 baseline.
2. Drive the **language server** over the same files as loose files (no project context) and collect the published diagnostics.
3. Compare. A `GS0157`/`GS0158`/`GS0159` in the editor for a file that compiles and runs clean is the defect.

**Executing the language server is the whole job here** — the finding is explicitly unverified, and a static read is not enough to act on. If the IDE path already resolves references by some other route, say so with the evidence and close.

## If it reproduces

Give the IDE path the same reference resolution the drivers now have. Prefer **sharing the existing mechanism** over a second copy — #3147's own review noted the previous fix's fragile syntactic gate (`ImportSyntax.StartsWith("Gsharp.Extensions")`) was **removed rather than hardened**, and that was the stronger outcome. Do not reintroduce a parallel decision surface.

Pin it with a test that drives the language server and asserts the **published diagnostic set is empty**, not merely that some call returns non-null.

## Related, do not conflate

`import Gsharp.ExtensionsBogus` compiles **rc0 on both sides** of #3147 — a **pre-existing leniency** in import validation, not caused by #3147 and out of scope here. Mentioned only so it is not mistaken for this bug.

## Acceptance

- A measurement table: driver diagnostics vs IDE diagnostics for the same files.
- Either a fix plus a language-server-level regression test, or a documented close with the probe output.
- Assert **published diagnostics**, never rc alone.

DavidObando:
## Summary

- load bundled `Gsharp.Extensions` through existing `ResolveDriverReferencePaths` before constructing default in-process resolvers
- tolerate corrupt bundled assemblies instead of escaping from every resolver-less compilation
- cover real published IDE diagnostics, including clean, extension, undefined-type, and corrupt-assembly cases

## Measurement

| probe | drivers | IDE before | IDE after |
|---|---|---|---|
| AddressBook control | clean | empty | empty |
| Optional / Mixed / Sequences / plain import | clean | unresolved extension diagnostics | empty |
| corrupt bundled assembly + undefined type | n/a | `BadImageFormatException` escaped | undefined-type diagnostic published |

## Mutation / RED proof

- exact original fix-hunk revert: RED, 4 failed / 1 passed
- bundled-reference condition inversion control: RED, 4 failed / 1 passed
- corrupt-assembly pin alone against narrow `FileLoadException` catch: RED with `BadImageFormatException`, 1 matched test
- restored/fixed pin: GREEN, 1/1

## Validation

- `LanguageServer.Tests`: 468/468
- 9 of 9 test projects enumerated from `GSharp.sln`: 15,460 passed, 5 skipped, 0 failed
- ILVerify: 123/123, 2 documented suppressions, 0 verification or compile failures
- CI: green

Fixes #3198

### DIFF
--- a/src/Core/CodeAnalysis/Symbols/ReferenceResolver.cs
+++ b/src/Core/CodeAnalysis/Symbols/ReferenceResolver.cs
@@ -374,6 +374,18 @@ public sealed class ReferenceResolver : IDisposable
     /// <returns>A default resolver.</returns>
     public static ReferenceResolver Default()
     {
+        foreach (var path in ResolveDriverReferencePaths(Array.Empty<string>()))
+        {
+            try
+            {
+                _ = AssemblyLoadContext.Default.LoadFromAssemblyPath(path);
+            }
+            catch (FileLoadException)
+            {
+                // Already loaded into the default context.
+            }
+        }
+
         return new ReferenceResolver(BuildHostAssemblies(), metadataContext: null);
     }
 
--- a/src/Core/CodeAnalysis/Symbols/ReferenceResolver.cs
+++ b/src/Core/CodeAnalysis/Symbols/ReferenceResolver.cs
@@ -378,11 +378,12 @@ public sealed class ReferenceResolver : IDisposable
         {
             try
             {
+                // Deliberately load here: BuildHostAssemblies excludes DriverReferenceLoadContext assemblies.
                 _ = AssemblyLoadContext.Default.LoadFromAssemblyPath(path);
             }
-            catch (FileLoadException)
+            catch (Exception ex) when (ex is FileLoadException or BadImageFormatException)
             {
-                // Already loaded into the default context.
+                // Already loaded or corrupt; leave it unavailable to the resolver.
             }
         }
 
--- a/test/LanguageServer.Tests/Issue3198LooseFileExtensionDiagnosticsTests.cs
+++ b/test/LanguageServer.Tests/Issue3198LooseFileExtensionDiagnosticsTests.cs
@@ -1,0 +1,119 @@
+// <copyright file="Issue3198LooseFileExtensionDiagnosticsTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using GSharp.LanguageServer.Protocol;
+using GSharp.LanguageServer.Server;
+using Xunit;
+
+namespace GSharp.LanguageServer.Tests;
+
+/// <summary>Issue #3198: loose files resolve bundled extension references.</summary>
+public sealed class Issue3198LooseFileExtensionDiagnosticsTests
+{
+    [Fact]
+    public async Task LooseFileWithoutExtensions_PublishesNoDiagnostics()
+    {
+        var diagnostics = await GetPublishedDiagnosticsAsync(
+            "AddressBook",
+            File.ReadAllText(Path.Combine(LocateSamplesDirectory(), "AddressBook.gs")));
+
+        Assert.Empty(diagnostics);
+    }
+
+    [Theory]
+    [InlineData("GsharpExtensionsOptional")]
+    [InlineData("GsharpExtensionsMixed")]
+    [InlineData("GsharpExtensionsSequences")]
+    public async Task LooseExtensionSample_PublishesNoDiagnostics(string sample)
+    {
+        var diagnostics = await GetPublishedDiagnosticsAsync(
+            sample,
+            File.ReadAllText(Path.Combine(LocateSamplesDirectory(), sample + ".gs")));
+
+        Assert.True(
+            diagnostics.Count == 0,
+            string.Join(Environment.NewLine, diagnostics.Select(d => $"{d.Code.Value}: {d.Message}")));
+    }
+
+    [Fact]
+    public async Task LooseFileWithPlainExtensionImport_PublishesNoDiagnostics()
+    {
+        var diagnostics = await GetPublishedDiagnosticsAsync(
+            "PlainExtensionImport",
+            """
+            package Issue3198.Plain
+            import System
+            import Gsharp.Extensions.Sequences
+
+            let values = Sequences.Of(7, 8, 9)
+            Console.WriteLine(values[0])
+            """);
+
+        Assert.True(
+            diagnostics.Count == 0,
+            string.Join(Environment.NewLine, diagnostics.Select(d => $"{d.Code.Value}: {d.Message}")));
+    }
+
+    private static async Task<IReadOnlyList<Diagnostic>> GetPublishedDiagnosticsAsync(string name, string source)
+    {
+        var directory = Path.Combine(
+            AppContext.BaseDirectory,
+            nameof(Issue3198LooseFileExtensionDiagnosticsTests),
+            name + "-" + Guid.NewGuid().ToString("N"));
+        Directory.CreateDirectory(directory);
+        Assert.Empty(Directory.GetFileSystemEntries(directory));
+
+        try
+        {
+            var sourcePath = Path.Combine(directory, name + ".gs");
+            File.WriteAllText(sourcePath, source);
+            var server = new LspServer(new DocumentContentService(), new WorkspaceState());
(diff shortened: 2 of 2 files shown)


DavidObando:

Head `a4f55cfb` (matches live `headRefOid`), `main` `10bb4caf`, merge base `8b32fc36`, 2 behind / 1 ahead. `git merge-tree --write-tree` **output** = one tree `c681abfa`, no conflict markers. Fresh worktree `/tmp/w3211`, own `out/` root, Release build rc=0 / 0 warnings, freshness by epoch mtime on all 37 non-`ref/` `GSharp.Core.dll` copies (`Core`/`Compiler`/`Repl`/`LanguageServer` all `1785863487`).

**+12 production lines.** That is the right size for this bug.

---

### 1. The anti-over-correction row — the one the table lacks. **It stays RED.**

The all-`empty` "IDE after" column is only meaningful if something is still allowed to be non-empty. I added 6 rows the PR does not have and ran them through the same `LspServer.DidOpenAsync` → published-diagnostics path.

| row | fixed product | unfixed product |
|---|---|---|
| `CTL-control` AddressBook (no extensions) | **empty** ✅ | **empty** (insensitive — good control) |
| `RED-undef` undefined type | `Cannot find type NoSuchTypeAnywhere` **(GS0157)** | same |
| **`RED-mixed` extension import + undefined type** | **`Cannot find type TotallyUndefinedType`** | `Cannot find type Sequences` |
| `RED-syntax` broken `func Broken( {` | `Unexpected token <Open…>` ×N | same |
| `RED-member` bad member on a real extension type | `Cannot find function NoSuchMemberOnSequence` (GS0159) | `Cannot find type Sequences` |
| `RED-typemismatch` `Take("not an int")` | `Parameter 'value' re…` | same |

**`RED-mixed` is the decisive row.** One file that both `import Gsharp.Extensions.Sequences` *and* names a type that does not exist: after the fix `Sequences` resolves **and** `Cannot find type TotallyUndefinedType` is still published. Before the fix that same file reported `Cannot find type Sequences`. The fix is **surgical, not suppression** — it adds resolvable names, it does not silence binding.

`RED-syntax`, `RED-undef` and `RED-typemismatch` are **byte-identical across the fix**, which is the correct scoping signature.

**Harness soundness (your #3):** I re-ran the AddressBook control myself. Clean on both products.

### 2. Shared mechanism — confirmed, and by concept, not one spelling

- `git grep 'StartsWith("Gsharp.Extensions'` → **nothing**, rc=1. Not reintroduced.
- The only decision surface is `ReferenceResolver.FindBundledExtensionPath` (`:968`), reached solely through `ResolveDriverReferencePaths` (`:398`). Consumers: `Compiler/Program.cs:124`, `Repl/Program.cs:144` + `:164`, and now `ReferenceResolver.Default():377`. **One surface, four callers — no second copy.**
- Concept grep over all of `src/` for `Gsharp.Extensions`: the remaining hits are `Gsharp.Extensions.Go` import-gating (GS0316/GS0317, ADR-0082), a **different pre-existing** mechanism, correctly not conflated.

### 3. RED-proof — the test genuinely fails unfixed

Exact fix-hunk revert (12 deletions), asserted applied via `git diff HEAD --stat` **and** artifact movement:

- CI-flavoured pristine `GSharp.Core.dll` = `d0f8d939…`, reproducibly (verified by forced rebuild).
- Mutant = `85eb4d10…` — **moved**.
- `Issue3198LooseFileExtensionDiagnosticsTests`: **Failed: 4, Passed: 1**, rc=1, `No test matches` count **0**.
- It fails for the **right reason**: `Semantic: Cannot find type Sequences. Are you missing an import?` — GS0157, the actual defect, not a harness artifact.
- The 1 that passes is the AddressBook control, correctly insensitive.
- Restored: source md5 exact round-trip `0e3ae8b4…`, artifact back to `d0f8d939…`, `git status --porcelain` empty.

The test drives the real `LspServer` and asserts `diagnostics.Count == 0` with the diagnostic text as the failure message. That satisfies ADR-0154 — it is not a non-null check.

### 4. Cost — measured, negligible

`Default()` is reached once per **root** `BoundScope` (`BoundScope.cs:133` inherits from parent), i.e. once per binding pass, not per scope. Timed 200 iterations in-process:

```
added loop = 3.5 us   existing Default() = 135.8 us   share = 2.6%
```

`Default()` already rebuilt the whole host-assembly set on every call; the new loop is noise against it, and `LoadFromAssemblyPath` is a cache hit after the first. **No caching change needed.**

### 5. Suite

`LanguageServer.Tests` **467 passed / 0 failed** on the restored tree (462 baseline + 5 new). CI on the head: `.statusCheckRollup|length` = **32**, all COMPLETED — **30 SUCCESS + 2 SKIPPED** (`publish`, `publish-visual-studio-extension`); **`ilverify` SUCCESS**. Nothing pending, nothing red. Denominator **9**, from `dotnet sln GSharp.sln list`.

---

## SHOULD-FIX (not blocking)

**`ReferenceResolver.cs:379-386` — `catch (FileLoadException)` is too narrow for a path that now runs on every compilation.**

`FindBundledExtensionPath` (`:968-982`) gates on `File.Exists` **only** — it never checks the file is a managed assembly. I measured what a corrupt `Gsharp.Extensions.dll` actually produces:

```
AssemblyLoadContext.Default.LoadFromAssemblyPath(<30 bytes of text>)
  -> BadImageFormatException   (NOT caught by the new code)
```

Before this PR `Default()` could not throw. Now a truncated or partially-copied bundled DLL turns **every** resolver-less compilation — every keystroke in the editor — into an unhandled exception rather than a diagnostic. Narrow trigger (install corruption), but the blast radius is global and the fix is one clause:

```csharp
catch (Exception ex) when (ex is FileLoadException or BadImageFormatException)
```

`TryValidateDriverReferencePaths` in this same file already catches `BadImageFormatException` for exactly this reason, so the codebase treats it as a real case.

## NIT

`Default()` is a factory whose name promises no side effects but now permanently mutates `AssemblyLoadContext.Default` for the host process. That is deliberate — `BuildHostAssemblies():1238` skips `DriverReferenceLoadContext`, so loading into Default is how the assembly becomes visible — but it is worth a one-line comment saying so, since the exclusion at `:1238` otherwise reads like the opposite intent.

---

## My own near-misses, disclosed

1. **My first rendering was weaker than my claim.** I printed `Diagnostic.Code.Value`, which yields the *category* (`Semantic`, `Syntax`) — not the GS id. Had I stopped there I would have written "GS0157 still publishes" having never seen a GS id. Re-ran rendering `Message` to get the actual text.
2. **I nearly reported a failed md5 round-trip.** My baseline `a48477c4` was captured after a `dotnet test`, which rebuilds **without** `-p:ContinuousIntegrationBuild=true`; my restore build used the CI flag, so it landed on `d0f8d939`. Build-flavour confound, not a bad restore. Resolved by proving the CI-flavoured pristine build is deterministic at `d0f8d939` and the mutant `85eb4d10` differs from it in the same flavour.
3. First test invocation used a non-existent csproj name and died `MSB1009` with rc=1 — caught, not scored. `No test matches` grep was 0 on every run that produced a number.

Worktree left at `a4f55cfb`, `git status --porcelain` empty, no shared checkout touched.

Issue #3176, opened 2026-08-04
Adopt emit-to-memory execution for gsi and bare gsc (retire the tree-walking evaluator)


DavidObando:
Part of #3163 (code-health P2 headline item).

## Decision

Replace the tree-walking Evaluator as the execution engine behind `gsi` and bare `gsc` with the Roslyn scripting model: compile every program/submission with the **real emitter** into a `MemoryStream`, load the PE bytes into a collectible `AssemblyLoadContext` inside the driver process, invoke the entry point, and surface stdout/exit code exactly as today. One codegen pipeline executes everywhere; the evaluator/emit semantic-divergence bug class is eliminated **by construction** rather than bug-by-bug.

Design and phasing are recorded in **ADR-0156** (`docs/adr/0156-gsi-emit-to-memory-execution.md`), which includes a working feasibility spike: three shipped samples emitted to memory, run in a collectible ALC in-process, byte-identical stdout against both the golden files and the evaluator, ALC verified collectible, steady-state per-submission latency ~47 ms (Debug) vs ~23 ms for the evaluator.

## Motivation: the divergence bug class

The Evaluator (~7,300 lines, `src/Core/CodeAnalysis/Evaluator*.cs`) is a second implementation of the language that runs a *different* lowering pipeline than emit. Every semantic feature must land twice, and when the copies disagree the failure shape is the worst one we have: **rc 0, no diagnostic, wrong answer**. One month of instances:

- #3140 — `out` write-back silently dropped for class instance methods (gsi + bare gsc)
- #3134 — classes get value equality instead of reference identity
- #3116 — struct `GetHashCode`/`Equals` overrides ignored when the BCL calls them; `Dictionary.Count` differs by driver
- #3110 — constant-pattern equality diverges across drivers
- #3100 — `@FieldOffset` overlap ignored; silently different value than emitted code
- #3099 — struct/enum type arguments erased to `System.Object` when reifying CLR generics
- #3022 — ref-aliasing local re-evaluates its initializer; reads/writes the wrong storage
- #3004 — pointer dereference returns a stale value with exit 0
- #3006 — exhaustive enum `switch` falls through to default silently
- #3003 / #3050 — lambda shapes crash with GS9999 under gsi only
- #3114 — ByRefLike (`Span`) signatures fail under both evaluating drivers; two shipped samples
- #3137 — reified CLR type arguments have no members under gsi and bare gsc

[...]

PR #3212, opened 2026-08-04
Report interactive storage boundaries as GS0513

### LINKED ISSUE #3199: [diagnostic] Interactive storage boundaries still surface as GS9999 instead of GS0513

DavidObando:
Closes #3199.

## Summary
- mark deliberate compiled-only storage refusals explicitly on `EvaluatorException`
- classify that marker once in `Compilation.Evaluate` as `GS0513`
- preserve construct-specific messages and keep unexpected evaluator failures on `GS9999`
- pin all five sites through in-process `SessionEngine` tests
- document post-#3200 measured behavior in ADR-0153
- mark pins as deprecated-evaluator coverage retiring in ADR-0156 Phase 3c

No message-text dispatch.

## Build

Rebased onto latest `main` `5cd0d766a60fa522bc6b9ad042fe14be9c3f3fde` after #3200 and #3242. Requested compile failure did not reproduce. Exact `dotnet build GSharp.sln` result:

```
Build succeeded.
    0 Warning(s)
    0 Error(s)
```

Fresh Release build head: `102cfdde1617033b515085415f1f7c184c0a3ec5`.

- build started: `1785881415`
- `GSharp.Core.dll` mtime: `1785881419`
- `GSharp.Core.dll` md5: `65f01c34f24bcefd03fdce9afc2468e8`

## Rebuilt evaluator measurement

In-process `SessionEngine`; exact deprecated tree evaluator selected by `--engine evaluator`.

| Site | Unfixed | Fixed | Severity | HasError | Output |
|---|---|---|---|---|---|
| `fixed` | `GS9999` | `GS0513` | Error | true | empty |
| `stackalloc` | `GS9999` | `GS0513` | Error | true | empty |
| `sizeof(Pair)` | `GS9999` | `GS0513` | Error | true | empty |
| `&Method` | `GS9999` | `GS0513` | Error | true | empty |
| function-pointer invocation | `GS9999` | `GS0513` | Error | true | empty |

Messages remain unchanged. ADR-0153 and `docs/diagnostics.md` assign all five sites to `GS0513`.

File-mode control remains emitted: `gsi fixed.gs` returned rc=0, stdout exactly `11\n`, stderr empty.

## Unexpected-failure control

```gsharp
@Flags
enum Access { None = 0, Read = 1, Write = 2 }

func choose(x Access) int32 {
    return switch x {
        case Access.None: 0
        case Access.Read: 1
        case Access.Write: 2
    }
}

choose(Access.Read | Access.Write)
```

Path: `SessionEngine.Evaluate` -> `Compilation.Evaluate` -> tree evaluator unmatched-switch exception.

Exact result: `GS9999`, severity `Error`, `HasError=true`, message `Unmatched switch expression value.`, output empty.

## RED proof

Latest unfixed `main` `5cd0d766a60fa522bc6b9ad042fe14be9c3f3fde`; only new test file applied (`git diff HEAD`: 122 insertions). Serial Release build of `Interpreter.Tests` passed; unfixed Core md5 `e462f6e5f5d493a7cb28fe91aee4882c`.

Exact six-test result: 5 failed (`Expected GS0513`, `Actual GS9999`), unexpected-`GS9999` control passed; rc=1; no `No test matches`. Restored porcelain-0, rebuilt successfully, exact md5 round-trip, removed worktree.

## Product mutant matrix

Clean fixed Core md5: `65f01c34f24bcefd03fdce9afc2468e8`. Every mutant had non-empty `git diff HEAD`, moved Core md5, was killed, had no zero-match output, then restored to exact baseline md5.

| Mutant | Core md5 | Result |
|---|---|---|
| factory marker `true -> false` | `e52f374dce2a7b38417df36380797818` | 5 failed, 1 passed |
| shared classifier condition inverted | `3acc4627c20d55241e0960d29ac98f87` | 6 failed; control became `GS0513` |
| four expression-site markers removed | `603a24e9464e27fae6d677d1b769261e` | 4 failed, 2 passed |
| `fixed` marker removed | `fcf1e99c91b466e2c1797c44af2e15ec` | 1 failed, 5 passed |

## Validation
- issue pins: 6/6 passed; filter matched 6
- nine Release test assemblies: 15,537 passed, 5 explicitly skipped, 0 failed; no zero-match output
- ILVerify: 123/123 verified, 2 documented suppressions, 0 failures
- CI: 30 successful checks, 2 expected publish skips
- merge state: clean and mergeable


DavidObando:
## Rubber-duck — **MERGEABLE** (0 BLOCKER, 1 SHOULD-FIX, 2 NIT)

**Provenance for every number below:** built SHA **`102cfdde1617033b515085415f1f7c184c0a3ec5`**, `dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true` rc=0 / **0 warnings, 0 errors**, `GSharp.Core.dll` md5 **`71540f73909f92e8b788944f861dccd4`**. Fresh worktree, own `out/`. Merge base **`5cd0d766a`**, 2 behind / 4 ahead; `git merge-tree --write-tree` rc=0, one tree line, **`grep -c CONFLICT` = 0**. `mergeable` now resolves **MERGEABLE**; rollup **30 SUCCESS / 2 SKIPPED**.

---

### Q1 — the GS9999 half is pinned now, and the pins genuinely fail unfixed

Re-ran the sweep on this head. `git grep -lE "3199|CompiledOnlyStorage|Adr0153" -- test/` now returns `Issue3199InteractiveStorageBoundaryDiagnosticTests.cs` — the gap is closed. The pins assert **`diagnostic.Id`, `Severity`, the exact `Message`, and `Output == string.Empty`**, under `Assert.Single(cell.Diagnostics)`, so a stray extra diagnostic also fails. That is an id-and-text pin, not "something failed".

**RED proof, run myself.** Reverted the 4 src files to merge-base, keeping the test file. Applied and proven two ways: `git diff HEAD --numstat` (`1/5`, `4/4`, `1/3`, `0/20` — the exact inverse of the PR's src diff, asserted with `git diff HEAD` because `git checkout <ref> -- <path>` **stages**), and `Core.dll` md5 movement **`71540f73` → `11b157c9`**. Result:

```
Failed!  - Failed: 5, Passed: 1, Total: 6      rc=1, "No test matches" count = 0
Assert.Equal() Failure: Expected: "GS0513"  Actual: "GS9999"   (×5)
```

**RED for the right reason — the actual wrong id.** And the one test that *passes* unfixed is `UnexpectedEvaluatorFailure_StillReportsGs9999`, which is the correct signature: the must-stay row does not ride on the change. Restored to exact md5 round-trip (`71540f73`).

### Q2 — control fires; both trap spellings identified and correctly untouched

I built the control the tables lacked and measured **every** row on **`SessionEngine`** *and* **`Compilation.Evaluate`**, before and after. Ids read, never inferred:

| row | unfixed | fixed | |
|---|---|---|---|
| **CTL `&x` / `*p`** | **GS0513** | **GS0513** | **control fires — harness is at the boundary** |
| B1 `fixed` | GS9999 | **GS0513** | differential |
| B2 `stackalloc` (inside `func`) | GS9999 | **GS0513** | differential |
| B3 `sizeof(Pair)` | GS9999 | **GS0513** | differential |
| B4 `&identity` | GS9999 | **GS0513** | differential |
| B5 fn-ptr invocation | GS9999 | **GS0513** | differential |
| TRAP `var b = stackalloc …` top-level | **GS0219** | **GS0219** | different guard — *not* reclassified |
| TRAP `fixed` outside `unsafe` | **GS0400** | **GS0400** | different guard — *not* reclassified |
| MUST9999 unmatched switch | GS9999 | GS9999 | invariant |
| MUST9999 index out of range | GS9999 | GS9999 | invariant |
| MUST9999 divide by zero | GS9999 | GS9999 | invariant |
| MUST9999 unbox user struct | GS9999 | GS9999 | invariant |

**5 differential, 7 invariant.** Both surfaces agree on all 24 cells. Messages are byte-identical across the change — the self-contained text is preserved exactly as #3199 required. The two trap spellings land on GS0219/GS0400 as warned, and critically **the fix does not drag them into GS0513**.

### Q3 — nothing over-claims GS0513, and I checked the *inverse* too

Structurally over-claiming is near-impossible: `IsCompiledOnlyStorageBoundary` defaults to **`false`** and is set only by the single new factory, read at the single catch site. **There is exactly one `catch (EvaluatorException)` in `src/`** (`Compilation.cs:406`) — `SessionEngine.cs:237` routes through `compilation.Evaluate`, so no second decision surface was introduced. Measured on 4 genuine-failure rows above: all stay GS9999.

**The inverse matters more, and nobody tested it.** GS0513 literally tells the user *"compile this program with 'gsc' instead"*. If that advice were false for any of the five sites, this PR would send users to a driver that also fails — the exact harm described. So I compiled all five with `gsc` (rc captured before any pipe):

| site | `gsc` rc | stdout |
|---|---|---|
| `fixed` | 0 | `11` + `Success.` |
| `stackalloc` | 0 | `2` + `Success.` |
| `sizeof(Pair)` | 0 | `8` + `Success.` |
| `&identity` + invocation (B4+B5) | 0 | `7` + `Success.` |

**The GS0513 advice is verified actionable for every reclassified site.** `gsi fixed.gs` also stays rc=0 / stdout `11` / stderr empty, so file mode is untouched, as #3199 required.

### Q4 — the PR body's numbers reproduce

Given the history, I rebuilt and re-measured rather than reading. Everything behavioural corroborates: rebase base `5cd0d766a` ✅, head `102cfdde` ✅, `0 Warning(s) / 0 Error(s)` ✅, the five-row Unfixed→Fixed table ✅ **exactly**, the unmatched-switch control (`GS9999`, `Error`, `HasError=true`, `Unmatched switch expression value.`, empty output) ✅, the file-mode `gsi` control ✅, and the RED shape (5 failed / control passed / rc=1 / no `No test matches`) ✅.

The two `Core.dll` md5s in the body (`65f01c34…`, `e462f6e5…`) do **not** match mine (`71540f73…`, `11b157c9…`) — but that is expected and **not** a discrepancy: MVIDs are non-deterministic across worktrees, so those digits are only meaningful as within-tree movement for whoever produced them, and are unverifiable by construction for anyone else. The claim they support — that the numbers came from a real rebuild — is independently confirmed by my own reproduction. **The unbuilt-tree problem does not recur here.**

`Interpreter.Tests`: **1505 passed, 4 skipped, 0 failed**, `No test matches` = 0. No emit-path file is touched by this diff (`git diff --name-only | grep -i emit` is empty), so the IL surface is unchanged; CI's `ilverify` is SUCCESS.

---

## SHOULD-FIX — `docs/diagnostics.md:321-325` is now factually wrong and cites a closed issue as open

The user-facing diagnostics reference still says:

> Older guards for `fixed`, `stackalloc`, unmanaged `sizeof`, and function pointers **still carry self-contained boundary messages through legacy `GS9999`**; issue [#3199] **tracks moving** those deliberate failures out of the internal-error category.

After this PR all five carry **GS0513** — I measured it — and #3199 is closed *by this PR*. ADR-0153 was updated correctly in three places; this paragraph was missed. It matters more than a normal doc nit because (a) it is the document users consult when they see the code, (b) it now contradicts shipped behaviour, and (c) #3199 itself names `docs/diagnostics.md` as one of the two authorities for this contract. Confirmed isolated: `grep -rn "3199" docs/` outside ADR-0153 returns only this line. One sentence.

## NIT-1 — PR body wording invites the #3230 misreading

"In-process `SessionEngine`; exact deprecated tree evaluator selected by `--engine evaluator`" reads as though the CLI flag was used. It was not — the tests construct `SessionEngine` in-process, which is correct, because `gsi --engine evaluator <file>` silently measures the **emitted** engine (#3230). The measurement is sound; only the sentence is confusable.

Related, and in the PR's favour: `Compilation.Evaluate` is `[Obsolete]`-as-error (*"New callers are forbidden; see #3176"*), so SessionEngine-only coverage is **forced**, not a shortcut. I had to `#pragma warning disable CS0618` in my throwaway probe to measure that second surface at all.

## NIT-2 — the reclassification is narrow, and by ADR-0153's own logic slightly incomplete

My sweep found at least one more interpreter-only refusal still on GS9999: unboxing a user struct (`Pt(a)`) → `GS9999: Unexpected type Pt` (`Evaluator.Calls.cs:699`), which **`gsc` compiles and runs correctly**. By the ADR's rationale that is a compiled-only boundary too. Not a defect here — under-claiming is the safe direction and over-claiming is the harmful one — and #3249 already tracks the underlying gap. Recording it so the flag's coverage is a deliberate decision rather than an accident.

---

Worktree left at `102cfdde`, **`git status --porcelain` empty**, shared checkout never touched. (Current on-disk `Core.dll` md5 is `812efa58`, the non-CI-flavour rebuild left by `dotnet test`; source is provably identical — porcelain empty — and all measurements above come from the CI-flavour `71540f73` build.)

DavidObando:
## ⛔ Correction to my review above — **I reverse my MERGEABLE verdict. This PR is now OBSOLETE. Do not merge.**

My earlier verdict was measured against merge-base `5cd0d766a` and is, I believe, still accurate *for that base*. But `main` moved while I was reviewing, and the movement does not conflict with this PR — **it deletes everything this PR edits.** Nothing below is a criticism of the work; the world moved out from under it.

### Ground truth, re-fetched

`gh/main` is now **`8b3bdac57322ce5c574ad6d6a54daea602d4436d`**, three commits past this PR's merge-base:

```
8b3bdac57 Retire the tree-walking evaluator (ADR-0156 Phase 3c) (#3244)   <-- this one
6bf63e333 Fix evaluator object reference equality (#3195)
a5de8fd0c Fix website driver execution documentation (#3207)
```

**Phase 3c is not scheduled any more. It landed.**

First, to close the question that prompted this re-check: the `[Obsolete]`/CS0618 census gate was **never** a risk here. This PR's merge-base `5cd0d766a` *is* #3242 itself, so the gate was already in force when CI went green; the diff adds **zero** `#pragma warning disable CS0618`; and both pins drive `SessionEngine.Evaluate`, the allowlisted surface, never `Compilation.Evaluate`. (`EvaluatorRetirementGateTests.cs` has since been deleted by #3244 anyway — the gate retired with its subject.) That concern is clean twice over. The real problem is larger.

### `merge-tree` onto current main — rc=1, **4 CONFLICTs**

```
CONFLICT (content): src/Core/CodeAnalysis/Compilation/Compilation.cs
CONFLICT (modify/delete): Evaluator.Expressions.cs deleted in gh/main and modified in pr3212
CONFLICT (modify/delete): Evaluator.Statements.cs deleted in gh/main and modified in pr3212
CONFLICT (modify/delete): EvaluatorException.cs  deleted in gh/main and modified in pr3212
```

**Every product file this PR touches is deleted on main.** So is `Compilation.Evaluate` (the single catch site the fix depends on), so is the `InterpreterPointerOperationsNotSupported` descriptor — **GS0513 itself no longer exists** — and `SessionEngine` is now `EmittedSessionEngine`/`ISessionEngine`, so the pins' surface is gone too. This is not a rebase; it would be a resurrection.

### Measured on current main — built myself, SHA `8b3bdac57`, `GSharp.Core.dll` md5 `774c43effcf7520c527c89b6663f3fae`, 0W/0E

Same specimens, taken verbatim from this PR's own `TheoryData`, driven in-process through `EmittedSessionEngine`:

| row | diagnostics on main | output |
|---|---|---|
| **CTL `&x` / `*p`** | **(none)** | prints `7` |
| B1 `fixed` | **(none)** | prints `11` |
| B2 `stackalloc` | **(none)** | — |
| B3 `sizeof(Pair)` | **(none)** | — |
| B4 `&identity` | **(none)** | — |
| B5 fn-ptr invoke on `nil` | `GS9998` — `NotSupportedException: Conversion from 'nil' to '*func(int32)…` | — |
| must-stay control (unmatched switch) | **`GSI002`** | — |

**There is no longer a GS9999 to reclassify, and no GS0513 to reclassify it to.** Four of the five constructs — and the GS0513 control that anchored both of our tables — now **execute successfully with zero diagnostics**. Issue #3199 has been resolved by deletion, and more completely than this PR could: instead of a friendlier refusal, the constructs simply work.

Two details worth naming:

- B5's remaining `GS9998` is a *different* defect — converting `nil` to a function-pointer type. It is not the "invocation requires `calli`" boundary this PR reclassifies.
- The must-stay-GS9999 control now reports **`GSI002`**, so `UnexpectedEvaluatorFailure_StillReportsGs9999` — the one test that legitimately passed in my RED proof — **would also fail** on main. All six pins are dead, not just the five.

### The doc hunk would actively regress

ADR-0153 on main now reads:

> **Status**: Superseded by ADR-0156 — the tree-walking evaluator (**and with it this boundary and GS0511/GS0513**) was deleted in Phase 3c (#3176)

This PR's ADR-0153 edit rewrites the same table to assert the five sites "Meet contract after #3199" as live `GS0513` behaviour. Merging it would convert a *superseded* ADR back into an apparently-live contract for a diagnostic that no longer exists — worse than the stale `docs/diagnostics.md` paragraph I flagged earlier, which #3244 has since made moot as well.

### Recommendation

Close this PR as obsolete, and close **#3199** (still `OPEN`) as resolved by **#3244** rather than fixed here. If any residue is wanted, the only live question is B5's `GS9998` nil-to-function-pointer conversion, which is a genuinely separate issue and unrelated to storage boundaries.

My apologies for a verdict that needed reversing — the base-relative finding stands, but the merge-base was three commits stale by the time I posted, and I should have re-fetched `main` immediately before publishing rather than relying on the SHA I established at the start of the review.

DavidObando:

PR #3229, opened 2026-08-04
Reify ambient generic arguments in evaluator

### LINKED ISSUE #3197: [evaluator] Enclosing-generic reification fallback arm silently yields `object` for unsubstituted type parameters
## Summary

`Evaluator.Expressions.cs` reifies enclosing generic type arguments through a substitution dictionary. The reification helper's fallback arm — the `:` branch of the ternary that maps a type argument when it is **not** found in the substitution dictionary — passes a raw `TypeParameterSymbol` straight through without the dictionary. The result silently reifies to `typeof(object)`.

Found by the rubber-duck review of PR #3191 (issue #3180) as a SHOULD-FIX. #3191 fixed the *structural* defect (a nested builder redeclared flattened enclosing + own parameters); this arm is a **separate, still-live** hole in the same helper.

Locate by symbol, not by line — #3191 moved these lines and cites go stale.

## Repro

A generic type nested such that the inner member's type argument is an enclosing type parameter that is *not* substituted at the top level:

```
Owner[T].Payload[string]
```

| driver | printed |
|---|---|
| `gsc /out:` (emit oracle) | `A:System.Int32` |
| interactive `SessionEngine` | **`A:System.Object`** |

`rc=0`, **no diagnostic**. Silent wrong answer — the value is observably wrong and nothing warns.

## Why it survived #3191's 16-case matrix

**All 16 cases substitute at the top level**, so the ternary's `:` arm never executes in any existing test. The matrix axis is coarser than the language surface it claims to cover.

This is the **third instance of the repo's recurring defect shape** — a test name stating a semantic property while the assertion checks a coarser observable across both sides. ADR-0154 (#3169) is policy on this; the matrix here needs a case where the type argument is genuinely absent from the substitution dictionary.

## Why this is not a one-liner

The obvious fix — "pass the dictionary in the `:` arm too" — does not work. During nested builder construction the dictionary holds **`GenericTypeParameterBuilder`** instances, not concrete runtime types. Resolving the fallback correctly requires an **ambient instantiation** to substitute against, which does not exist at that point in the current design. Expect to thread context, not to patch an argument list.

## Scope guard

Four branches in this helper are pinned by #3151 and were verified byte-identical by the #3191 review — **nullable**, **imported**, **slice**, **array**. Do not disturb them; assert they still round-trip byte-identically.

## Driver model (verify, do not trust)

Post-ADR-0156 the evaluator is **interactive-REPL-only**: `gsc <file>`, `gsc /out:` and `gsi <file>` all emit via `EmittedProgramHost.Run`; only `SessionEngine` interprets, and it is still the shipped default (`ReplHost.cs`, `--engine emit` is opt-in). `gsi` **refuses piped stdin**, so a shell-driven `gsi` probe can silently test nothing. Reproducing this needs an **in-process `SessionEngine` harness** built out-of-tree.

## Acceptance

- The interactive column prints `A:System.Int32`, matching the emit oracle.
- A regression test whose type argument is genuinely **absent** from the substitution dictionary, i.e. one that exercises the `:` arm — prove it fails on the unfixed tree.
- Assert printed **values**, never rc or diagnostic sets.
- The four #3151-pinned branches unchanged.

DavidObando:
## Summary

- resolve constructed G# struct symbols against evaluator ambient runtime type substitutions
- resolve before `ClrBackingTypes` cache lookup, so each concrete ambient instantiation gets correct backing type identity
- add emitted-oracle versus interactive `SessionEngine` regression using `Owner[T].Payload[string]` with `int32` and `bool` instantiations

This does not pass the builder substitution dictionary into the old fallback. That dictionary contains `GenericTypeParameterBuilder` values and lacks the enclosing function parameter. Instead, existing ambient runtime substitutions recursively reconstruct top-level, nested, and nested-generic struct symbols before CLR backing-type creation and caching.

Fixes #3197

## Printed-value proof

Validated from base `0484b00f`.

| Tree / driver | Printed field types |
|---|---|
| unfixed main, emitted oracle | `A:System.Int32`, `B:System.String` |
| unfixed main, interactive `SessionEngine` | `A:System.Object`, `B:System.String` |
| fixed, emitted oracle | `A:System.Int32`, `B:System.String`; then `A:System.Boolean`, `B:System.String` |
| fixed, interactive `SessionEngine` | byte-for-byte equal to fixed emitted oracle |

Both fixed instantiations print `G:2`, proving nested generated CLR type reifies enclosing plus own generic parameters. Emitted assembly executed successfully and `dotnet-ilverify` reported all classes and methods verified.

## Regression and mutation proof

- Regression test applied alone to unfixed main: RED (`System.Object` versus emitted `System.Int32`/`System.Boolean`). Restored to porcelain-clean production and rebuilt.
- Struct-resolution condition inversion: regression RED.
- runtime-argument-helper empty-condition inversion: regression RED.
- resolved cache-key condition inversion: regression RED.
- Each product mutant produced non-empty `git diff`, moved `GSharp.Core.dll` md5, and restoration round-tripped exact original md5.

Dying test for every hunk: `OpenEnclosingFunctionTypeArgument_MatchesEmittedFieldTypes`.

## Scope guard

#3151-pinned fallback branches are byte-identical to base:

- nullable: `857092da3f3f790eb06159032e84b8af`
- imported: `4f6128304ba2c728bbccca634eb7947e`
- slice: `e811588118b77a8aaec33d244571a6c2`
- array: `0952a3151d25f83064d6fc39752096ba`

## Validation

- Debug solution build: 0 warnings, 0 errors
- reification matrix: 17/17 passed
- all 9 solution test projects, unfiltered: 15,455 passed, 5 skipped, 0 failed
  - Compiler 4,263
  - Core 7,135 passed, 1 skipped
  - Interpreter 1,485 passed, 4 skipped
  - Extensions 104
  - InternalAnalyzers 9
  - LanguageServer 462
  - SDK 59
  - Cs2Gs 1,907
  - GeneratorHost 31
- final-base Core rerun: 7,135 passed, 1 skipped
- final-base Interpreter rerun: 1,485 passed, 4 skipped
- `git merge-tree`: no conflict markers

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -1344,6 +1344,29 @@ public sealed partial class Evaluator
                 : NullableTypeSymbol.Get(underlying);
         }
 
+        if (type is StructSymbol structType)
+        {
+            var definition = structType.Definition ?? structType;
+            var ownArguments = ResolveRuntimeTypeArguments(structType.TypeArguments, out var ownChanged);
+            var enclosingArguments = ResolveRuntimeTypeArguments(
+                structType.EnclosingTypeArguments,
+                out var enclosingChanged);
+
+            if (!ownChanged && !enclosingChanged)
+            {
+                return type;
+            }
+
+            if (!enclosingArguments.IsDefaultOrEmpty)
+            {
+                return ownArguments.IsDefaultOrEmpty
+                    ? StructSymbol.ConstructNested(definition, enclosingArguments)
+                    : StructSymbol.ConstructNestedGeneric(definition, enclosingArguments, ownArguments);
+            }
+
+            return StructSymbol.Construct(definition, ownArguments);
+        }
+
         if (type is not ImportedTypeSymbol imported
             || imported.OpenDefinition == null
             || imported.TypeArguments.IsDefaultOrEmpty)
@@ -1351,18 +1374,33 @@ public sealed partial class Evaluator
             return type;
         }
 
-        var arguments = ImmutableArray.CreateBuilder<TypeSymbol>(imported.TypeArguments.Length);
-        var changed = false;
-        foreach (var argument in imported.TypeArguments)
+        var arguments = ResolveRuntimeTypeArguments(imported.TypeArguments, out var changed);
+
+        return changed
+            ? ImportedTypeSymbol.GetConstructed(imported.ClrType, imported.OpenDefinition, arguments)
+            : type;
+    }
+
+    private ImmutableArray<TypeSymbol> ResolveRuntimeTypeArguments(
+        ImmutableArray<TypeSymbol> arguments,
+        out bool changed)
+    {
+        if (arguments.IsDefaultOrEmpty)
+        {
+            changed = false;
+            return arguments;
+        }
+
+        var resolvedArguments = ImmutableArray.CreateBuilder<TypeSymbol>(arguments.Length);
+        changed = false;
+        foreach (var argument in arguments)
         {
             var resolved = ResolveRuntimeTypeSymbol(argument);
             changed |= !ReferenceEquals(resolved, argument);
-            arguments.Add(resolved);
+            resolvedArguments.Add(resolved);
         }
 
-        return changed
-            ? ImportedTypeSymbol.GetConstructed(imported.ClrType, imported.OpenDefinition, arguments.MoveToImmutable())
-            : type;
+        return resolvedArguments.MoveToImmutable();
     }
 
     private T ResolveMemberForRuntimeType<T>(T member, Type runtimeType)
--- a/src/Core/CodeAnalysis/Evaluator.Expressions.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Expressions.cs
@@ -1269,6 +1269,12 @@ public sealed partial class Evaluator
     {
         var clrBase = baseCtor.DeclaringType
             ?? throw new InvalidOperationException("Imported base constructor has no declaring type.");
+        var runtimeStructType = ResolveRuntimeTypeSymbol(structType);
+        if (!ReferenceEquals(runtimeStructType, structType))
+        {
+            structType = (StructSymbol)runtimeStructType;
+        }
+
         var backingType = ClrBackingTypes.GetValue(
             structType,
             _ => CreateClrBackingType(structType, clrBase));
--- a/test/Interpreter.Tests/Issue3099TypeArgumentReificationTests.cs
+++ b/test/Interpreter.Tests/Issue3099TypeArgumentReificationTests.cs
@@ -24,6 +24,83 @@ namespace GSharp.Interpreter.Tests;
 [Collection("ConsoleIo")]
 public class Issue3099TypeArgumentReificationTests
 {
+    [Fact]
+    public void OpenEnclosingFunctionTypeArgument_MatchesEmittedFieldTypes()
+    {
+        const string Source = """
+            package Issue3197
+            import System
+
+            class Owner[T] {
+                class Payload[U] {
+                    let A T
+                    let B U
+                }
+            }
+
+            class Box[T] : EventArgs {
+            }
+
+            func Reify[T]() {
+                var value = Box[Owner[T].Payload[string]]()
+                var payload = value.GetType().GenericTypeArguments[0]
+                Console.WriteLine("G:" + payload.GetGenericTypeDefinition().GetGenericArguments().Length.ToString())
+                for field in payload.GetFields() {
+                    Console.WriteLine(field.Name + ":" + field.FieldType.FullName)
+                }
+            }
+
+            Reify[int32]()
+            Reify[bool]()
+            """;
+
+        var root = Path.Combine(
+            GetRepositoryRoot(),
+            "out",
+            "test-artifacts",
+            $"issue3197-{Guid.NewGuid():N}");
+
+        try
+        {
+            var emitDirectory = PrepareEmptyDirectory(Path.Combine(root, "gsc-emit"));
+            var sourcePath = Path.Combine(emitDirectory, "Probe.gs");
+            var assemblyPath = Path.Combine(emitDirectory, "Probe.dll");
+            File.WriteAllText(sourcePath, Source);
+            _ = CaptureDriver(() => Program.Main(new[]
+            {
+                "/out:" + assemblyPath,
+                "/target:exe",
+                "/targetframework:net10.0",
+                sourcePath,
+            }));
+
+            var assembly = Assembly.Load(File.ReadAllBytes(assemblyPath));
+            Assert.NotEmpty(assembly.GetTypes());
+            var entryPoint = assembly.EntryPoint
+                ?? throw new InvalidOperationException("Emitted assembly has no entry point.");
+            var emitted = CaptureDriver(() =>
+            {
+                entryPoint.Invoke(
+                    null,
+                    entryPoint.GetParameters().Length == 0 ? null : new object[] { Array.Empty<string>() });
+                return 0;
+            });
+            var interactive = RunInteractive(Source);
+
+            Assert.Equal(
+                "G:2\nA:System.Int32\nB:System.String\nG:2\nA:System.Boolean\nB:System.String\n",
+                emitted);
+            Assert.Equal(emitted, interactive);
+        }
+        finally
+        {
+            if (Directory.Exists(root))
+            {
+                Directory.Delete(root, recursive: true);
+            }
+        }
+    }
+
     public static IEnumerable<object[]> TypeArgumentCases()
     {
         yield return ["class", "class Payload {\n    let Eleven int32\n    var TwentyTwo string\n}", "Payload"];


DavidObando:
# Rubber-duck review — PR #3229 (`fix/3197-enclosing-generic-reification`)

**Verdict: MERGEABLE** — no blocker. One SHOULD-FIX with a measured 6-line patch attached, two NITs.

Head `59c9ec2f`, anchored by ancestry onto merge-base `0484b00f1` (single commit). Merges clean against current `main` `20fbbec41` (`git merge-tree` → 0 `CONFLICT` lines). CI 30 pass / 0 fail, `ilverify` pass (4m1s). Denominator 9 from `dotnet sln GSharp.sln list` per `build/generate-ci-test-matrix.py:26`.

Driver model re-established from my own rebuilt `gsi --help`, not from memory: `--engine <name>` is *"Interactive engine: 'emit' (default) or 'evaluator'"*. The test drives `SessionEngine` **in-process**, and `SessionEngine.EvaluateCore` (`src/Repl/Engine/SessionEngine.cs:234`) calls `compilation.Evaluate(...)` — so it genuinely reaches the tree-walking evaluator and is not silently measuring emit (the #3230 trap).

---

## Q1 — is the silent wrong answer gone, and is the oracle strong? **Yes.**

The test (`test/Interpreter.Tests/Issue3099TypeArgumentReificationTests.cs:28-100`) asserts an exact literal
`"G:2\nA:System.Int32\nB:System.String\nG:2\nA:System.Boolean\nB:System.String\n"` for the emitted oracle **and** `Assert.Equal(emitted, interactive)` at `:93`. It pins concrete type names, covers two genuinely distinct instantiations (`int32`, `bool`), and satisfies ADR-0154 — it is not an absence-of-diagnostic test.

Verified RED for the right reason, both product hunks reverted independently, each proven applied by `git diff HEAD` + `GSharp.Core.dll` md5 movement (all builds CI-flagged, like-for-like):

| mutant | applied | result |
|---|---|---|
| M1 — drop `StructSymbol` arm (`Evaluator.Calls.cs:1347-1368`) | −23 lines, md5 `6847b6f3`→`597a3816` | **KILLED** — `Actual: "…A:System.Object…"` |
| M2 — drop cache pre-resolve (`Evaluator.Expressions.cs:1272-1276`) | −6 lines, md5 →`f1d64bad` | **KILLED** — `Actual: "…A:System.Object…"` |
| M3 — drop `!ownChanged && !enclosingChanged` fast path | −5 lines, md5 →`bc2ab758` | survived (see Q2 — semantics-preserving) |
| M4 — **control**, swap enclosing/own in `ConstructNestedGeneric` | 1/1 line, md5 →`a3df9822` | **KILLED** (1 fail / 1489) |

Both kills print the actual wrong value `System.Object`, not merely a red line. Restored: `git diff HEAD` = 0 and md5 round-trips exactly to `6847b6f3`.

## Q2 — the cache-ordering change. **Correct in both directions.**

Measured via the dynamic-assembly ordinal (`CreateClrBackingModule`, `Evaluator.Expressions.cs:1568`, names modules `GSharp.Interpreter.ClrBacking.{n}`), which makes every backing-type creation observable. Four calls — `int32`, `int32`, `bool`, `int32`:

- **base** `0484b00f1`: `.4`, `.4`, `.4`, `.4` — all four `Object`. The key was **too coarse**: `int32` and `bool` collided into one backing type. That is itself a silent wrong answer, and this PR fixes it.
- **head** `59c9ec2f`: `.6`, `.6`, **`.8`**, `.6` — distinct instantiations get distinct types, identical ones reuse. Neither too coarse nor too fine.

By construction: `StructSymbol.cs:23/30/38` are three `ConcurrentDictionary` interning caches keyed by `(Definition, TypeArgsKey…)`, so `Construct`/`ConstructNested`/`ConstructNestedGeneric` return canonical instances, and the `ConditionalWeakTable<StructSymbol, Type>` therefore keys on a canonical symbol. M3 confirms this empirically: deleting the unchanged fast path leaves ordinals at `.10/.10/.12/.10` — identity rests on interning, not on that branch. No hit-rate collapse.

## Q4 — over-correction. **None.**

On base, **all 12** probe shapes diverged from the emit oracle. Nothing goes correct→broken; the PR is a strict improvement. `typeof(object)` is still returned where it is genuinely right (a fully-instantiated top-level `Owner[int32].Payload[string]` matches emit on base and head alike).

---

## SHOULD-FIX — `ResolveRuntimeTypeSymbol` handles 4 of 20 type kinds; six shapes still silently reify to `object`

`src/Core/CodeAnalysis/Evaluator.Calls.cs:1374` — the catch-all `return type;`.

`ResolveRuntimeTypeSymbol` has arms for `TypeParameterSymbol`, `NullableTypeSymbol`, `StructSymbol` (new) and `ImportedTypeSymbol`. There are **20** `TypeSymbol` subclasses. Every other composite that can carry a type parameter — slice, array, map, tuple, interface, delegate, sequence — falls through `:1374` unchanged, so the still-unsubstituted `TypeParameterSymbol` reaches `ResolveClrBackingTypeArgument`, which is called **without** the `reifiedTypeParameters` dictionary at `Evaluator.Expressions.cs:1439` and `:1447`, and bottoms out at `return typeof(object);` (`:1520`) — the exact arm issue #3197 was filed about.

Measured, interactive vs emit oracle (`Box[Owner[<shape>].Payload[string]]`, `Reify[int32]()`):

| enclosing shape | emit (oracle) | base | **head (this PR)** | head + 6 lines |
|---|---|---|---|---|
| `T` | `System.Int32` | `Object` | **`Int32`** ✔ | `Int32` ✔ |
| `List[T]` | `List<Int32>` | `List<Object>` | **`List<Int32>`** ✔ | ✔ |
| `Owner[T].Payload[int64]` | `Owner+Payload<Int32,Int64>` | `<Object,…>` | **✔** | ✔ |
| `[]T` | `System.Int32[]` | `Object[]` | **`System.Object[]`** ✗ | `Int32[]` ✔ |
| `[2]T` | `System.Int32[]` | `Object[]` | **`System.Object[]`** ✗ | `Int32[]` ✔ |
| `[][]T` | `Int32[][]` | `Object[][]` | **`Object[][]`** ✗ | `Int32[][]` ✔ |
| `List[[]T]` | `List<Int32[]>` | `List<Object[]>` | **`List<Object[]>`** ✗ | `List<Int32[]>` ✔ |
| `[]List[T]` | `List<Int32>[]` | `List<Object>[]` | **`List<Object>[]`** ✗ | `List<Int32>[]` ✔ |
| `IFace[T]` | `IFace<Int32>` | `Object` | **`System.Object`** ✗ | `System.Object` ✗ |

Score: base **0/10**, this PR **3/10**, this PR + 6 lines **8/10**.

These are **pre-existing, not regressions** — which is why this is not a blocker — but they are the same silent-wrong-answer class as #3197, and `[]T` is an everyday G# shape.

**I did not merely assert a smaller fix; I built and measured two.**

*Attempt 1 — wholesale delegation to the existing `Binder.SubstituteType` (`Binder.cs:4970`), replacing the author's arm entirely (+5/−64): **FAILED.*** Even `bare_T` regressed to `object`. Root cause: `Binder.cs:5087` handles a constructed struct via `StructSymbol.Construct(ss.Definition, newStructArgs)`, which **drops `EnclosingTypeArguments`**; the branch that preserves enclosing args (`:5113`) only fires when the nested type has *no* own type arguments. A nested generic with **both** own and enclosing arguments — precisely `Owner[T].Payload[U]` — is unhandled. **The author's dedicated struct arm is genuinely necessary, and the "just reuse the existing helper" critique is wrong.** Stated plainly because it is the crux of the design call, and the author's reasoning holds up.

*Attempt 2 — hybrid: keep the author's struct arm, delegate only the catch-all at `:1374`: **WORKS.*** +6 lines:

```csharp
// ponytail: slice/array/interface/map/tuple/delegate/... already
// recurse in Binder.SubstituteType; only nested structs need the
// dedicated arm above (SubstituteType drops EnclosingTypeArguments).
return TypeSymbol.ContainsTypeParameter(type)
    ? Binding.Binder.SubstituteType(type, CaptureRuntimeTypeArguments())
    : type;
```

(`Binder` must be qualified — `System.Reflection.Binder` is in scope, CS0104.) Both `Binding.Binder` and `TypeSymbol.ContainsTypeParameter` are already imported in this file, and `CaptureRuntimeTypeArguments()` (`Evaluator.Calls.cs:1275`) already returns exactly the `Dictionary<TypeParameterSymbol, TypeSymbol>` that `SubstituteType` takes.

Measured: fixes 5 of the 6 shapes, **zero regressions**, Release build 0 warnings, `Interpreter.Tests` **1485 passed / 4 skipped / 0 failed** including this PR's own new test.

Entirely at the maintainer's discretion given the evaluator is deprecated for ADR-0156 Phase 3c — but the marginal cost is 6 lines and the class closes structurally rather than one shape at a time.

`IFace[T]` is a *different* hole and is not fixed by the above: `ResolveClrBackingTypeArgument` has no `InterfaceSymbol` arm, so even a fully-substituted `IFace[int32]` still falls to `typeof(object)` at `Evaluator.Expressions.cs:1520`.

## NIT — nullable enclosing argument still diverges from the emit oracle, and is untested

`Box[Owner[T?].Payload[string]]`, `Reify[int32]()`: emit prints `System.Int32`; the interpreter prints `System.Nullable\`1[System.Int32]` on head (it printed `System.Object` on base). Neither side matches the other before or after this PR, so it is not a regression, and I am **not** asserting which side is correct — `T?` on a value type plausibly *should* be `Nullable<int32>`, which would make this an emit defect rather than an evaluator one. Flagging it only because issue #3197 explicitly named nullable as one of the four #3151-pinned branches to assert, and the new test does not cover it.

## NIT — latent gap in `Binder.SubstituteType` for nested generics

`src/Core/CodeAnalysis/Binding/Binder.cs:5087` drops `EnclosingTypeArguments` for a constructed nested struct that has its own type arguments (detail under SHOULD-FIX above). I did **not** demonstrate a user-visible defect from it — emit gets these shapes right — so this is latent, recorded for whoever next touches that helper, not scored against this PR.

---

### Informational, not scored against this PR

- `chan T` as an enclosing type argument produces `GS9998: ArgumentNullException: Value cannot be null. (Parameter 'key')` with a non-source-anchored `(1,1,1,1)` span — **byte-identical on base and head**, so pre-existing and untouched here. Same shape as the #3192 complaint.
- Lambda-capture and `go` shapes that carry an enclosing type parameter **fail on the emit path** (`BadImageFormatException`; `TypeLoadException: Could not load type 'P.<go_1>\`1'`) on base and head alike, while the interpreter renders them correctly — the reverse of the usual direction. Pre-existing, out of scope, mentioned only because emit is normally the oracle and here it is not usable for those two shapes.
- `go` is **not** a Q3 hole, by construction: `CloneExecutionStateForGoroutine` (`Evaluator.Statements.cs:428-438`) shallow-copies the `Locals` stack, preserving the same frame *instances*, and `runtimeTypeArguments` is a `ConditionalWeakTable` keyed on those instances — so ambient substitutions survive the thread boundary. Confirmed by probe.

### Suites run (head, Release, CI-flagged)

`Interpreter.Tests` 1485 passed / 4 skipped / 0 failed · `Core.Tests` 7135 passed / 1 skipped / 0 failed.

### Hygiene

Fresh clone with its own `out/`; the shared checkout and all sibling worktrees untouched; no `git stash`; no `--no-build`; nothing written to `~/.nuget/packages/gsharp.net.sdk/`. All mutants and both experimental patches reverted — `git diff HEAD` = 0 and `GSharp.Core.dll` md5 round-trips exactly to `6847b6f3`. Only untracked entry is the pre-existing #3170 `packages.lock.json`.

DavidObando:
# Rubber-duck review — PR #3229, **re-run after rebase. Verdict REVERSED.**

**Previous verdict: MERGEABLE. Revised verdict: OBSOLETE — recommend closing unmerged.** Not because the fix was wrong. It was correct, and I verified it thoroughly. It has been overtaken by **#3244**.

## What changed

`main` is **`8b3bdac57`**, not `6bf63e333` — main moved once more while I was re-running. The new commit is:

```
8b3bdac57 Retire the tree-walking evaluator (ADR-0156 Phase 3c) (#3244)
100 files changed, 1350 insertions(+), 11516 deletions(-)
```

**Phase 3c deleted the tree-walking evaluator.** Both files this PR modifies no longer exist:

| path | on `8b3bdac57` |
|---|---|
| `src/Core/CodeAnalysis/Evaluator.Calls.cs` | **deleted** |
| `src/Core/CodeAnalysis/Evaluator.Expressions.cs` | **deleted** |

`git ls-tree -r gh/main -- src/Core/CodeAnalysis/ \| grep -i evaluator` returns **0 files**. Every symbol on both sides of this change — `ResolveRuntimeTypeSymbol`, `ResolveClrBackingTypeArgument`, `ClrBackingTypes`, `CaptureRuntimeTypeArguments`, `InvokeClrBackingCtor` — returns **0 files** repo-wide. `Compilation.Evaluate` and `class Evaluator` are gone. `src/Repl/Program.cs:121-122` now reads: *"the only engine — the legacy tree-walking evaluator (and its `--engine evaluator` escape hatch) was deleted."*

This is a **modify/delete conflict**, not line drift. `git merge-tree --write-tree 59c9ec2f6 gh/main` exits 1 and its **output** carries 2 `CONFLICT` lines listing stages **1 and 2 only, with no stage 3** — the signature of "deleted on main, modified on the branch". GitHub agrees: `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`. There is nothing to rebase onto.

Worth recording precisely, because it nearly fooled me: against the **briefed** `6bf63e333` this PR merged **clean** — 0 `CONFLICT` lines, bare tree OID. The conflict exists only against the true tip. A rebase check against a briefed SHA rather than a freshly fetched remote would have cleared this PR.

## The defect is resolved by deletion — measured, not assumed

I built `8b3bdac57` Release, CI-flagged (rc=0, **0 warnings**, `GSharp.Core.dll` md5 `774c43ef`, moved from the PR head's `6847b6f3`), and ran **this PR's own test program verbatim** through all three surviving drivers:

| driver | rc | output |
|---|---|---|
| `gsc Probe.gs` | 0 | `G:2 / A:System.Int32 / B:System.String / G:2 / A:System.Boolean / B:System.String` ✔ |
| `gsc /out:` + execute | 0 / 0 | identical ✔ |
| `gsi Probe.gs` | 0 | identical ✔ |

That is exactly the string this PR's test asserts. Issue #3197's silent `typeof(object)` is gone because the code that produced it is gone.

## My SHOULD-FIX and both NITs are moot too

The 10-shape enclosing-type-argument sweep from my earlier review, re-run on `8b3bdac57` through `gsc` (asserting printed field types, never rc):

| enclosing shape | base `0484b00f1` | PR #3229 | PR + my 6-line hybrid | **main `8b3bdac57`** |
|---|---|---|---|---|
| `T` | `Object` ✗ | `Int32` ✔ | ✔ | **`System.Int32`** ✔ |
| `[]T` | ✗ | ✗ | ✔ | **`System.Int32[]`** ✔ |
| `[2]T` | ✗ | ✗ | ✔ | **`System.Int32[]`** ✔ |
| `T?` | ✗ | diverged | diverged | **`System.Int32`** ✔ |
| `List[T]` | ✗ | ✔ | ✔ | **`List<Int32>`** ✔ |
| `IFace[T]` | ✗ | ✗ | ✗ | **`IFace<Int32>`** ✔ |
| `Owner[T].Payload[int64]` | ✗ | ✔ | ✔ | **`Owner<Int32>+Payload<Int64>`** ✔ |
| `[][]T` | ✗ | ✗ | ✔ | **`System.Int32[][]`** ✔ |
| `List[[]T]` | ✗ | ✗ | ✔ | **`List<Int32[]>`** ✔ |
| `[]List[T]` | ✗ | ✗ | ✔ | **`List<Int32>[]`** ✔ |
| **score** | **0/10** | **3/10** | **8/10** | **10/10** |

Deletion beat every patch, including mine. `IFace[T]` — which my hybrid could **not** fix, because `ResolveClrBackingTypeArgument` had no `InterfaceSymbol` arm — is correct on main. My NIT about the `T?` divergence is also resolved: main prints `System.Int32`, which is what the emit oracle said all along.

## The test is not salvageable as written

If the test hunk were rebased onto main on its own, it would be **vacuous**. On `8b3bdac57`, `Issue3099TypeArgumentReificationTests.cs:199` has `RunInteractive` constructing `new EmittedSessionEngine { CaptureConsole = true }`. Both sides of `Assert.Equal(emitted, interactive)` are then the emitted engine — **emit == emit**. It would pass unconditionally and pin nothing, which is precisely the ADR-0154 failure mode. The exact-literal assertion against `"G:2\nA:System.Int32\n…"` still has some value as an emit-path pin, but the cross-driver half is dead weight and the name (`MatchesEmittedFieldTypes`) would no longer describe a real comparison. If any of this is kept, it should be re-landed deliberately as an emit-only regression test, not rebased.

## Credit where it is due

My earlier review stands on its merits and I am not withdrawing it: the fix was correct, the oracle was strong, the cache change was right in both directions, and I verified all of that by mutation (M1/M2 killed with the actual wrong value `System.Object` visible, M4 control killed, exact md5 round-trip on restore). The author's design call to write a dedicated `StructSymbol` arm rather than delegate to `Binder.SubstituteType` was **correct**, and I proved it by building the delegation myself and watching it regress — `Binder.cs:5087` drops `EnclosingTypeArguments` for a nested generic carrying both own and enclosing arguments. None of that was wasted reasoning; it was simply aimed at an engine that was removed hours later.

---

## One surviving defect, unrelated to this PR, still live on `8b3bdac57`

A **channel** type as an enclosing type argument is an ICE on current main, on the only remaining path:

```
$ gsc C.gs
C.gs(1,1,1,1): error GS9998: ArgumentNullException: Value cannot be null. (Parameter 'key')
rc=1
```

Reproduces with `Box[Owner[chan T].Payload[string]]()` inside `func Reify[T]()`, called as `Reify[int32]()`. It was present on base and on the PR head too, so it is pre-existing and **not** caused by #3244 — but it now has no second engine behind it, and it carries a non-source-anchored `(1,1,1,1)` span, the same shape as the #3192 complaint. Flagging for the maintainer to triage; I do not file issues.

Separately: `test/Core.Tests/CodeAnalysis/EvaluatorRetirementGateTests.cs`, added by #3242, was **deleted by #3244** — consistent, since nothing remains to mark `[Obsolete]`, but worth knowing for anyone still working against a #3242-era brief.

## Hygiene

Fresh clone with its own `out/`; shared checkout and sibling worktrees untouched; no `git stash` (the untracked #3170 `packages.lock.json` blocks checkout to main because it is *tracked* there — I moved it aside by hand and restored it); no `--no-build`; nothing written to `~/.nuget/packages/gsharp.net.sdk/`. Scratch clone returned to `59c9ec2f6` with `git diff HEAD` = 0. All comparisons made head-against-its-own-merge-base, never head-to-head.

DavidObando:

Issue #3218, opened 2026-08-04
Emitter: structural projection into an imported nullable class parameter fails ("Conversion from 'Source' to 'System.Uri?' is not yet supported")


DavidObando:
## Repro

```gsharp
import System.Net.Http
class Source { var uriString string }
func Probe(client HttpClient, source Source) {
    let pending = client.GetAsync(source)
}
0
```

## Actual

At `main` `10bb4caf`, `Compilation.Evaluate` binds and runs this cleanly (the overload resolves via structural projection of `Source` onto `System.Uri`'s shape). Emitted execution (in-memory submission; same pipeline as `gsc`) fails at emit with:

```text
NotSupportedException: Conversion from 'Source' to 'System.Uri?' is not yet supported by the emitter.
```

## Expected

The emitter lowers the structural-projection conversion for imported (metadata) targets, including through the nullable wrapper, matching the binder's accepted conversion.

## Scope

Found by the ADR-0156 Phase 3b.1 emitted-oracle bulk migration (#3176): `StructuralProjectionTests.ImportedMethodOverloadAcceptsStructuralProjection` stays on `Compilation.Evaluate` with a reference to this issue; see also the companion constructed-generic projection failure filed from the same file.

Issue #3163, opened 2026-08-03
Code health analysis — August 2026: design fault lines, stability, security, and adoption roadmap


DavidObando:
# GSharp Code Health Analysis — August 2026

Evidence base: full repo exploration, graphify knowledge graph of `src/` (9,904 nodes / 33,576 edges), git history (2,078 commits), and GitHub tracker mining (1,373 issues, 1,788 PRs, ~54 bugs examined in detail via `gh`).

---

## Executive summary

GSharp is in remarkably good structural health for a project of its ambition: a full .NET language (compiler `gsc`, interpreter `gsi`, LSP server, VS/VS Code extensions, MSBuild SDK, C#→G# translator `cs2gs`) built essentially by one developer plus autonomous agents at ~200–400 commits/week since May 2026. Process hygiene is genuinely strong — 154 ADRs, ~12,800 tests, PR-per-issue workflow, one revert in 2,078 commits, warnings-as-errors, StyleCop, near-zero TODO debt, and a self-review ritual with an explicit verdict grammar on every PR.

The analysis surfaces **one dominant architectural fault line, two structural concentration risks, and a set of process gaps** that matter increasingly as the project targets enterprise adoption:

1. **Three execution engines, one semantics, no shared contract.** `gsc` emit, `gsc` evaluate, and `gsi` interpret routinely disagree — the single largest active bug cluster (13 of the last ~54 substantive bugs), always in the worst shape possible: exit code 0, no diagnostic, silently wrong answer.
2. **The Binding subsystem is a god-subsystem** (~110k lines, 46% of Core) and `ReflectionMetadataEmitter.cs` is the single most fragile file in the repo (339 commits, 168 of them fixes).
3. **The green-CI signal is weaker than it looks**: test-oracle degeneracy (7 instances found in one recent session of tests asserting symptoms rather than claims), no required status checks on `main` (the "Lock Main" ruleset blocks deletion/force-push but doesn't require green CI to merge), a write-only nightly gate, and no Windows/macOS CI leg.
4. **Enterprise-readiness gaps** are mostly process, not code: nullable reference types disabled, no dependency lock files, no CONTRIBUTING/SECURITY docs, bus factor of one.

None of these are fatal; all are addressable, and the project has already demonstrated it can close structural efforts fast (#1361 monolith split: opened and closed in 18 days).

---

## What's working well (keep doing this)

These are worth naming explicitly because they are unusual strengths that the recommendations below build on:

[...]

Issue #1518, opened 2026-06-30
gsc: lambda returning a nullable value type (T?) infers generic method type arg as T not Nullable<T> -> unverifiable delegate mismatch


DavidObando:
While migrating `Oahu.Decrypt` with cs2gs, `DashChunkEntryies.ValidateMdatSize` emits unverifiable IL:
```
Error [StackUnexpected]: ...ValidateMdatSize(...)][offset 0x82]
  [found ref 'Func`2<...SampleInfo,System.Nullable`1<int32>>']
  [expected ref 'Func`2<...SampleInfo,int32>'] Unexpected type on the stack.
```
from `trun.Samples.Select((s SampleInfo) -> s.SampleSize).OfType[int32]()...` where `SampleInfo.SampleSize` is `int32?`.

## Root cause
A lambda whose return type is a **nullable value type** (`T?`) passed to a generic method (e.g. `Enumerable.Select<TSource,TResult>`) makes overload-resolution type inference bind the type parameter to the **non-nullable underlying** `T` instead of `Nullable<T>`, while the emitter builds the delegate with the accurate `Func<…,T?>`. The mismatch (`Select<…,int32>` instantiation vs `Func<…,int?>` delegate) fails ilverify.

The actual root: `NullableTypeSymbol.ClrType` is the **bare** underlying CLR type, because the ctor passes `underlyingType.ClrType` to the base:
```csharp
// src/Core/CodeAnalysis/Symbols/NullableTypeSymbol.cs:24
: base(ComputeDisplayName(underlyingType), underlyingType.ClrType)   // int, NOT Nullable<int>
```
`FunctionTypeSymbol.BuildClrType` (`src/Core/CodeAnalysis/Symbols/FunctionTypeSymbol.cs:252`) builds the `Func<…>`/`Action<…>` shape from raw `parameterTypes[i]?.ClrType` (line 262) and `returnType.ClrType` (line 289). For a `Func<int,int?>` this collapses the return to `int`, yielding `Func<int,int>`. `OverloadResolution.TryInferTypeArguments` then unifies `TResult = int`. The emitter uses a different (symbol-accurate) path and emits `Func<int,int?>`, so the method instantiation and the delegate disagree.

There is already a helper that returns the correct CLR type for a nullable value type: `NullableTypeSymbol.GetEffectiveClrType` / `NullableLifting.GetEffectiveClrType` returns `Nullable<int>` for `int?`.

## Minimal repros
All FAIL today.

[...]

Issue #814, opened 2026-06-13
T? overload disambiguation for OrNil reference-vs-value splits (follow-up from #810)


DavidObando:
## Summary

Follow-up to #810 and ADR-0084 §L5. The dogfooded G# port of `SequenceExtensions.FirstOrNil` / `LastOrNil` / `SingleOrNil` needs to ship both reference- and value-type overloads under a single name:

```gs
func (self sequence[T]) FirstOrNil[T class]() T?
func (self sequence[T]) FirstOrNil[T struct]() T?
```

ADR-0088 (#750) added constraint-aware overload resolution for *imported* methods (CLR-side `where T : class` / `where T : struct`). G# itself now spells the constraints (ADR-0097, #775). The combination should let `arr.FirstOrNil()` dispatch to the right overload based on `T`'s reference-vs-value kind — but disambiguation is incomplete for the `T?` return-type case where both overloads have shape-identical receiver and parameter lists, and only the constraint and the return-type representation distinguish them.

## Repro

The C# `OptionalValueExtensions` family was migrated to a single-name surface in #750; `SequenceExtensions` would need the same — the gap is specifically the return-type-only differentiation on `*OrNil` shapes.

## Expected

`FirstOrNil` / `LastOrNil` / `SingleOrNil` ship under one name in G# with `class` / `struct` overload pairs and dispatch correctly based on receiver-element-type constraint kind.

## Why it matters

Closes the third (and last) remaining gap blocking the dogfooded port of `Gsharp.Extensions.Sequences` to G# under #792.

## Related

- #750 / ADR-0088 (constraint-aware overload resolution — imported).
- #775 / ADR-0097 (G# spelling for class/struct/new()).
- #810 (state machine for open-generic iterators).
- #792 (SDK bootstrap dogfooding).
- ADR-0084 §L5 (Extensions stdlib known limitations).

Parent: #706.

PR #3234, opened 2026-08-04
Fix the constructed-generics emit cluster: T? receiver lift, struct-to-constructed-interface box, private struct initializer ctor

### LINKED ISSUE #3219: Emitted structural projection onto a constructed generic writes the target's private field from the call site (FieldAccessException)
## Repro

```gsharp
class Source { var Value int32 }
class Box[T] {
    var Value T
    private var Marker int32 = 9
    func ReadMarker() int32 { return Marker }
}
let source = Source{Value: 7}
let box Box[int32] = source
box.ReadMarker()
```

## Actual

At `main` `10bb4caf`, `Compilation.Evaluate` returns `9` (projection copies `Value`, the declared initializer supplies `Marker`). Emitted execution (in-memory submission; same IL as `gsc`) fails at runtime with:

```text
FieldAccessException: Attempt by method '<Program>.<Main>$(System.String[])' to access field 'Box`1<System.Int32>.Marker' failed.
```

The projection lowering materializes the target by writing its fields directly at the conversion site, including the `private` `Marker` field — an access the CLR rejects from outside the type.

## Expected

Projection materialization honors accessibility (routes declared initializers through the target's constructor/init path rather than raw field stores), preserving declared field initializers as the evaluator does.

## Scope

Found by the ADR-0156 Phase 3b.1 emitted-oracle bulk migration (#3176): `StructuralProjectionTests.ConstructedGenericTargetsPreserveDeclaredFieldInitializers` stays on `Compilation.Evaluate` with a reference to this issue; see also the companion imported-nullable projection emit gap filed from the same file.

### LINKED ISSUE #3222: Emitter: generic method with user-generic-interface parameter called with an implementing struct produces InvalidProgramException
## Repro

```gsharp
interface IHolder[T] {
    func Get() T;
}

struct StringBox : IHolder[string] {
    var Value string
    func Get() string { return Value }
}

func Describe[T](h IHolder[T]) T {
    return h.Get()
}

Describe(StringBox{Value: "hi"})
```

## Actual

At `main` `10bb4caf`, `Compilation.Evaluate` infers `T = string` and returns `"hi"`. Emitted execution (in-memory submission; same IL as `gsc`) fails at runtime with:

```text
InvalidProgramException: Common Language Runtime detected an invalid program.
```

The inference itself works on both engines (binding succeeds, no diagnostics); the emitted IL for passing the struct through the constructed interface parameter (boxing to `IHolder<string>` at the `Describe` call boundary or the interface dispatch inside) is invalid.

## Expected

The call emits verifiable IL (box `StringBox` to the constructed interface, `callvirt` `Get`), returning `"hi"`.

## Scope

Found by the ADR-0156 Phase 3b.1 emitted-oracle bulk migration (#3176, slice 2/3): `Issue1932GenericMethodInferenceOverUserTypeTests.GenericInterfaceParameter_InfersTypeArgument_FromImplementingArgument` stays on `Compilation.Evaluate` with a reference to this issue; the file's other inference tests (including the class-based and explicit-type-argument controls) migrated cleanly.

### LINKED ISSUE #3226: Emitter: generic extension on a nullable receiver (func (self T?) ...) breaks for value types — invalid IL on int32?, wrong value on user structs
## Repro

```gsharp
package P
import System

func (self T?) MyOrElse[T](fb T) T {
    if self != nil { return self!! }
    return fb
}

var v int32? = nil
v.MyOrElse(99)
```

## Actual

At `main` `10bb4caf`, `Compilation.Evaluate` returns `99`. Emitted execution (in-memory submission; same IL as `gsc`) fails at runtime with:

```text
InvalidProgramException: Common Language Runtime detected an invalid program.
```

The user-struct variant is worse — it runs but returns the wrong value (silent divergence):

```gsharp
struct Point { var X int32  var Y int32 }
// same MyOrElse extension
var pt Point? = nil
var def = Point{X: 1, Y: 2}
var r = pt.MyOrElse(def)
r.X
```

Evaluator: `1` (the fallback). Emitted: `0` (`default(Point).X` — the nil receiver appears to flow through as a default-initialized struct instead of dispatching to the fallback branch). The reference-type variant (`string?` receiver) emits and runs correctly.

## Expected

The `T?`-receiver extension lowers correctly for `Nullable<T>` value-type receivers: verifiable IL, `self != nil` false for a nil receiver, fallback returned — matching the evaluator.

## Scope

Found by the ADR-0156 Phase 3b.1 emitted-oracle bulk migration (#3176, slice 3/3): `Issue773GenericReceiverDispatchTests` (`NullableReceiver_Extension_Dispatches_On_Int32Nullable`, `NullableReceiver_Extension_Dispatches_On_UserNullableStruct`) stay on `Compilation.Evaluate` with a reference to this issue; the file's other receiver-dispatch tests (including the `string?` nullable-reference variant) migrated cleanly.

DavidObando:
Fixes #3226. Fixes #3222. Fixes #3219. Part of #3176 (pre-3c burn-down), part of #3163.

## Summary

Three constructed-generics emit-path bugs found by the Phase 3b.1 oracle migration, one commit each:

- **#3226** — lift unconstrained `T?` generic calls to `Nullable<X>` at value-type call sites (`MethodBodyEmitter.TryPlanUnconstrainedNullableLift` + `BuildMethodSpecForLiftedGenericCall`): wrap bare-`T` args via `newobj Nullable<X>::.ctor`, unwrap a bare-`T` return via `box`+`unbox.any`.
- **#3222** — narrow the plain-call argument-conversion skip in `OverloadResolver.CallBinding` from "parameter CONTAINS a type parameter" to "parameter IS a bare type parameter", so a struct argument to a constructed-interface slot (`IHolder[T]`) materializes its `box`.
- **#3219** — synthesize a public parameterless `.ctor` for value structs whose declared field initializers include a non-public field (runs ALL initializers in-type); struct-literal/projection sites construct through it and skip the injected initializer entries (side effects run once). All-public-initializer structs keep the historical inline emission byte-for-byte.

## Root causes

| Issue | Symptom | Root cause | Fix layer |
|---|---|---|---|
| #3226 | `InvalidProgramException` at `int32?`; **silent wrong value** (default `Point` instead of fallback) at `Point?` | Unconstrained `T?` encodes as bare MVAR `!!T` (SignatureEncoder's reference-nullable model), but the call site pushes a live `Nullable<X>` into the `X`-instantiated slot — no unwrap/box anywhere; the two symptoms are the JIT's two reactions to the same layout mismatch | Emit (call-site MethodSpec lift; definition IL unchanged) |
| #3222 | `InvalidProgramException` passing an implementing struct as `IHolder[T]` | `OverloadResolver.CallBinding`'s per-argument guard skipped `BindConversion` for ANY parameter containing a type parameter; only a bare `!!T` slot is emitter-erased — `IHolder<!!T>` is a real reference slot needing the `box` | Binding (guard narrowed to `parameter.Type is TypeParameterSymbol`, mirroring `BindUserInstanceCall`) |
| #3219 | `FieldAccessException` writing `Box`1<int32>.Marker` from the call site | Value-struct literals (incl. struct-target structural projection) inline ALL declared field initializers as raw `stfld`s at the construction site; `GetUserStructFieldRef` has no visibility gate, and the CLR rejects the out-of-type private store (class targets already routed through the synthesized default ctor) | Emit (synthesized in-type parameterless ctor; planner + literal emission) |

Scope notes:
- #3226's lift generalizes to any *method* type parameter with a declared `T?` slot (the plain `func F[T](x T?)` shape was equally broken); struct-constrained `T?` (#814), reference instantiations, async bodies, byref slots, and nested-`T` shapes keep the status quo. No new emit caches (the lift reuses `BuildMethodSpec`; GSA0004 untouched).
- #3219's class-shaped repro from the issue already passed at current main (the class path constructs through its default ctor); the live defect was the struct-target path — both shapes are pinned. The projection's synthesized `""`-fill for a *private uninitialized* string field remains a known gap (pre-existing, unhit by the pinned tests).
- Unpins the three 3b.1 evaluator-pinned tests (#3221/#3225/#3228 twins deleted where orphaned); `StructuralProjectionTests` keeps its `EvaluateWithEvaluator` twin for the unrelated #3218 pin.

## Verification

- `dotnet build GSharp.sln -c Release`: 0 warnings, 0 errors.
- Witness (ADR-0154), two-sided: with the three product commits stashed (tests kept), **16 tests RED** — all 14 new emit-level tests' non-control cases + the 3 unpinned tests (`InvalidProgramException` / `FieldAccessException` / wrong values), controls green; with the fixes restored, **63/63 GREEN** across the six touched test classes.
- ILVerify (the CI gate, run locally via the repo tool): all repro assemblies verify clean post-fix — `int32?`/`Point?` receiver lift, plain `T?` param, struct→`IHolder[string]` box, struct/class projection, private-initializer literal (pre-fix: `StackUnexpected` ×3 shapes).
- Targeted filters (`~Generic|~Extension|~Projection|~Nullable|~StructLiteral`, Core.Tests Release): **1402/1402**.
- Full `Core.Tests` (Release): **7151/7152 passed**, 1 skipped (standing `RegenerateBaseline` skip), 0 failed (4 m 26 s).
- `Compiler.Tests` LanguageConformance filter: **126/126 passed** (2 m 12 s).
- Full `Interpreter.Tests` (evaluator untouched; parity backstop): **1484/1488 passed**, 4 skipped (standing skips, incl. the #3210-disabled map-concurrency tests), 0 failed (1 m 46 s).
- **CI-caught regression + fix** (`fc1dfb83`): the tests (compiler-issue1-remainder) shard flagged `Issue1518NullableDelegateInferenceEmitTests.EndToEnd_NullableReturn_IntoUserGenericMethod_Runs` (ILVerify `DelegateCtor`/`StackUnexpected`) — the #3222 guard narrowing had pulled lambda arguments into `BindConversion`, rebuilding the delegate from the collapsed structural shape (`Func<int32,int32>`) instead of the emitter's #1518 nullable-preserving shape. Fixed by keeping function-shaped arguments on the historical skip (`structuralArgumentType is not FunctionTypeSymbol`). Post-fix: `Issue1518` filter **5/5** (was 4/5), Compiler.Tests `~Delegate` sweep **328/328**, witness filter **63/63**, full Core.Tests re-run **7151/7152** (1 standing skip), 0 failed.
- NOT RUN: `Extensions.Tests`, `LanguageServer.Tests`, `Sdk.Tests`, `GeneratorHost.Tests` (no surface touched beyond Binding/Emit; CI required checks are the backstop), `Cs2Gs.Tests` (pre-existing flaky host crash; GoldenTests-filter policy), full `Compiler.Tests` beyond the conformance gate (time; conformance is the language-behavior gate for this cluster).

## Checklist

- [x] Behavioral tests carry a witness of discrimination (ADR-0154): each new/changed test fails on the pre-change code (pre-fix commit, reverted hunk, or an applied product mutant) and passes with it.
- [x] Self-review verdict recorded when one was run (MERGEABLE / BLOCKER / SHOULD-FIX / NIT), with residuals filed as issues rather than dropped.

Self-review verdict: **MERGEABLE** — three orthogonal, minimally-scoped fixes; every deliberately-out-of-scope shape (nested-`T` lift, imported-struct private initializers, explicit-parameterless-ctor structs) degrades to today's behavior rather than a new one.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.CallBinding.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.CallBinding.cs
@@ -1706,7 +1706,8 @@ internal sealed partial class OverloadResolver
             else if (argument.Type != expectedType
                 && !(substitution != null && parameter.Type is TypeParameterSymbol)
                 && (!(substitution != null && TypeSymbol.ContainsTypeParameter(parameter.Type))
-                    || Conversion.Classify(argument.Type, expectedType).IsImplicit))
+                    || (Conversion.Classify(argument.Type, expectedType).IsImplicit
+                        && structuralArgumentType is not FunctionTypeSymbol)))
             {
                 // Issue #2335 (audit follow-up): every OTHER implicit
                 // conversion reaches this point already classified
@@ -1743,7 +1744,14 @@ internal sealed partial class OverloadResolver
                 // InvalidProgramException at the call site. Substituted
                 // non-bare slots whose conversion is NOT implicit keep the
                 // historical skip (status quo) rather than surfacing new
-                // diagnostics from this branch.
+                // diagnostics from this branch. A FUNCTION-shaped argument
+                // (lambda/method group into a substituted delegate slot like
+                // `Func[A, T]`) also keeps the historical skip: the emitter's
+                // delegate-materialization path already builds the accurate
+                // nullable-preserving `Func<…,Nullable<T>>` shape (#1518), and
+                // re-materializing it here rebuilt the delegate from the
+                // collapsed structural shape (`Func<int32,int32>`), tripping
+                // ilverify's DelegateCtor check.
                 var argLoc = i < parameterSyntax.Length ? parameterSyntax[i].Location : syntax.Identifier.Location;
                 boundArguments[i] = conversions.BindConversion(argLoc, argument, expectedType);
             }
--- a/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.CallBinding.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.CallBinding.cs
@@ -1704,7 +1704,9 @@ internal sealed partial class OverloadResolver
                 boundArguments[i] = conversions.BindConversion(argLoc, argument, expectedType);
             }
             else if (argument.Type != expectedType
-                && !(substitution != null && TypeSymbol.ContainsTypeParameter(parameter.Type)))
+                && !(substitution != null && parameter.Type is TypeParameterSymbol)
+                && (!(substitution != null && TypeSymbol.ContainsTypeParameter(parameter.Type))
+                    || Conversion.Classify(argument.Type, expectedType).IsImplicit))
             {
                 // Issue #2335 (audit follow-up): every OTHER implicit
                 // conversion reaches this point already classified
@@ -1726,14 +1728,22 @@ internal sealed partial class OverloadResolver
                 // (`StackUnexpected: found Int32, expected ref 'object'`).
                 // The same gap silently dropped numeric widening
                 // (`int32 → int64`) and other representation-changing
-                // implicit conversions. `parameter.Type` (the UNSUBSTITUTED
-                // declared type) is checked — not `expectedType` — so an
-                // open erased slot in a generic callee (paramType containing
-                // a type parameter of THIS call's own substitution) is still
-                // skipped and left for the emitter's type-erasure boxing at
-                // the call boundary, exactly mirroring the equivalent guard
-                // in `BindUserInstanceCall`'s per-argument loop
+                // implicit conversions.
+                //
+                // Issue #3222: only a BARE erased slot (`parameter.Type is
+                // TypeParameterSymbol`, the `!!T` MVAR the emitter erases at
+                // the call boundary) is skipped — mirroring the equivalent
+                // guard in `BindUserInstanceCall`'s per-argument loop
                 // (`if (paramType is TypeParameterSymbol) { …; continue; }`).
+                // A parameter that merely CONTAINS a type parameter (e.g. a
+                // constructed generic interface `IHolder[T]`) emits as a real
+                // constructed slot (`IHolder<!!T>`), so a value-type argument
+                // still needs its materialized conversion (`box StringBox` →
+                // `IHolder<string>`); skipping it produced
+                // InvalidProgramException at the call site. Substituted
+                // non-bare slots whose conversion is NOT implicit keep the
+                // historical skip (status quo) rather than surfacing new
+                // diagnostics from this branch.
                 var argLoc = i < parameterSyntax.Length ? parameterSyntax[i].Location : syntax.Identifier.Location;
                 boundArguments[i] = conversions.BindConversion(argLoc, argument, expectedType);
             }
--- a/src/Core/CodeAnalysis/Emit/ConstructorBodyEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/ConstructorBodyEmitter.cs
@@ -325,6 +325,108 @@ internal sealed class ConstructorBodyEmitter
         => !classSym.InstanceFieldInitializers.IsEmpty;
 
     /// <summary>
+    /// Issue #3219: a value-kind struct whose declared field initializers
+    /// include a NON-public field cannot be materialized by the historical
+    /// call-site lowering (initobj + raw <c>stfld</c> per initializer) —
+    /// the private-field store executes outside the type and the runtime
+    /// rejects it with <see cref="FieldAccessException"/>. Such a struct
+    /// gets a synthesized public parameterless <c>.ctor</c> that runs ALL
+    /// declared instance field initializers in-type (mirroring the class
+    /// default-ctor path), and struct-literal sites construct through it.
+    /// Structs whose initializers are all public keep the historical inline
+    /// emission byte-for-byte. A user-declared parameterless ctor (which
+    /// would collide) and inline structs (fixed synthesized-member layout)
+    /// are excluded.
+    /// </summary>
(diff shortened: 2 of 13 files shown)


DavidObando:
**CI-caught regression (tests shard `compiler-issue1-remainder`): fixed in the follow-up commit.**

- **Failing test**: `Issue1518NullableDelegateInferenceEmitTests.EndToEnd_NullableReturn_IntoUserGenericMethod_Runs` — ILVerify `DelegateCtor` / `StackUnexpected` (`Func`2<int32,int32>` on the stack where `Func`2<int32,Nullable<int32>>` was expected).
- **Root cause**: the #3222 guard narrowing over-reached by one argument kind. A lambda argument flowing into a substituted delegate slot (`f Func[A, T]` at `T = int32?`) previously kept the historical skip; the narrowed guard now ran `BindConversion` on it, which re-materialized the delegate from the collapsed structural shape (`Func<int32,int32>`) instead of the emitter's nullable-preserving #1518 shape (`Func<int32,Nullable<int32>>`) — mismatching the `ldftn`/delegate-`.ctor` pair against the slot type. Not a #3226 lift interaction: `T = int32?` is already nullable, so `TryPlanUnconstrainedNullableLift` correctly declined that call.
- **Fix**: keep the historical skip for FUNCTION-shaped arguments (`structuralArgumentType is not FunctionTypeSymbol`) in the narrowed branch — delegate materialization stays with its dedicated paths, which already encode the #1518 contract; the #3222 struct-boxing materialization is unaffected.
- **Verification**: `Issue1518` filter 5/5 (was 4/5); #3226/#3222/#3219 witness filter 63/63; Compiler.Tests `~Delegate` sweep 328/328; full Core.Tests re-run green (see updated PR body).
- **Witness**: two-sided by construction — the pre-commit tree fails `EndToEnd_NullableReturn_IntoUserGenericMethod_Runs` (CI + local repro), the post-commit tree passes it, and the #3222 witnesses stay red-on-main/green-here.

Verdict for the follow-up: **MERGEABLE** — one-line predicate refinement, restores exact historical behavior for the regressed argument kind.

PR #3237, opened 2026-08-04
Reduce evaluator override dispatch allocations

DavidObando:
## Summary

- Confirms the post-#3127 slowdown still reproduces on reachable tree-walking evaluator surfaces: `Compilation.Evaluate` and the evaluator `SessionEngine` escape hatch.
- Removes the capturing LINQ predicate allocated by every override lookup.
- Reuses one evaluator-bound override dispatcher instead of allocating one delegate per `StructValue`.
- Deliberately does **not** resurrect #3194's override-result cache; that cache measured 5.2% slower despite a 99.99975% hit rate.

Refs #3135.

## Why this lever

Profiling placed `FindExternalOverride` scanning at only 0.14% exclusive CPU. Allocation measurement found the real regression: post-#3127 added about 8 MB per issue workload, and the capturing predicate alone allocated exactly 88 bytes per hash lookup (17.6 MB versus 8.8 MB over 100,000 calls). The fix removes those allocations without adding a dictionary or cache lookup.

File-mode and default interactive `gsi` now emit after ADR-0156, so they do not enter this path. The embedding API remains reachable; `SessionEngine` remains reachable through the documented evaluator escape hatch.

## Paired benchmarks

Release binaries were prebuilt and remained static. Each process warmed twice and reported the median of five workload executions. Each table contains 40 pairs in alternating A/B order. Load was captured before every process; threshold was 30 and observed ranges were 8.18–20.68, so no sample was discarded.

Exact workload: 20,000 dictionary inserts plus 20,000 lookups of a struct with no overrides.

### Does the regression reproduce?

Parent is #3127's parent (`4469e323`); main is `d5d25e4c`, whose evaluator sources are byte-identical through the final base `d6e7c4c39`.

| Surface | Parent median (min–max, σ), ms | Main median (min–max, σ), ms | Paired parent/main median (min–max, σ) | Parent faster |
|---|---:|---:|---:|---:|
| `Compilation.Evaluate` | 126.263 (113.536–143.795, 8.642) | 135.810 (122.925–149.834, 7.050) | 0.933× (0.796–1.101, 0.072) | 33/40 |
| evaluator `SessionEngine` | 125.915 (114.090–177.552, 10.338) | 135.761 (124.369–155.345, 8.622) | 0.933× (0.789–1.180, 0.075) | 35/40 |

Yes: current main is about 7.2% slower on both reachable evaluator surfaces.

### Before/after fix

| Surface | Main median (min–max, σ), ms | Fix median (min–max, σ), ms | Paired main/fix median (min–max, σ) | Fix faster |
|---|---:|---:|---:|---:|
| `Compilation.Evaluate` | 129.719 (119.058–294.777, 35.521) | 124.295 (114.043–188.610, 16.482) | 1.061× (0.675–1.750, 0.165) | 31/40 |
| evaluator `SessionEngine` | 121.830 (113.066–150.995, 9.088) | 114.461 (108.477–127.557, 4.710) | 1.062× (0.937–1.269, 0.075) | 37/40 |

### Positive control

The known #3134 class-identity difference was detected through both surfaces:

| Surface | #3127 parent | current main | fix |
|---|---|---|---|
| `Compilation.Evaluate` | `True` | `False` | `False` |
| evaluator `SessionEngine` | `True` | `False` | `False` |

## Mutation table

All applied mutants had a non-empty product diff and changed `GSharp.Core.dll`; killed mutants restored to the exact pre-mutant md5 before continuing.

| Mutant | Result | Signal |
|---|---|---|
| Per-value method-group allocation restored | RED | shared-dispatcher identity assertion failed |
| Capturing `FirstOrDefault` restored | RED | 100,000 hash calls allocated 17,600,000 bytes, above the 14 MB guard |
| Evaluator dispatcher initialized to `null` | RED | non-null dispatcher assertion failed |
| Base-chain loop condition inverted (`type == null`) | RED | printed dictionary count changed from expected `2` to `1` |
| `IsOverride` condition inverted | SURVIVED existing targeted corpus | Not scored; direct and implicit-override suites stayed green |
| Root-method match condition inverted | SURVIVED existing targeted corpus | Not scored; a follow-up probe threw before producing asserted output, so it was retargeted to the loop mutant above |

Main-without-PR control: the relevant existing `Issue2896StructObjectOverrideTests` corpus passed 46/46. Mutant restoration controls were GREEN with exact md5 round-trips.

## Validation

The denominator is 9 projects from `dotnet sln GSharp.sln list`.

- 9/9 unfiltered: **15,481 passed, 5 skipped, 0 failed**
- Final solution build: 0 warnings, 0 errors
- ILVerify: **123/123**, 2 documented suppressions, 0 verification failures
- Final `GSharp.Core.dll`: 20/20 top-level Release consumers non-zero and md5-identical

### DIFF
--- a/src/Core/CodeAnalysis/Evaluator.Calls.cs
+++ b/src/Core/CodeAnalysis/Evaluator.Calls.cs
@@ -1060,7 +1060,7 @@ public sealed partial class Evaluator
     }
 
     private StructValue CreateStructValue(StructSymbol structType) =>
-        new(structType, TryEvaluateExternalOverride);
+        new(structType, objectOverrideDispatcher);
 
     private bool TryEvaluateExternalOverride(
         StructValue receiver,
@@ -1104,14 +1104,18 @@ public sealed partial class Evaluator
 
     private static FunctionSymbol FindExternalOverride(StructSymbol type, MethodInfo externalMethod)
     {
-        FunctionSymbol externalOverride = null;
-        for (; type != null && externalOverride == null; type = type.BaseClass)
+        for (; type != null; type = type.BaseClass)
         {
-            externalOverride = type.Methods.FirstOrDefault(method =>
-                method.IsOverride && FindExternalOverrideRoot(method)?.Equals(externalMethod) == true);
+            foreach (var method in type.Methods)
+            {
+                if (method.IsOverride && FindExternalOverrideRoot(method)?.Equals(externalMethod) == true)
+                {
+                    return method;
+                }
+            }
         }
 
-        return externalOverride;
+        return null;
     }
 
     private static MethodInfo FindExternalOverrideRoot(FunctionSymbol method)
--- a/src/Core/CodeAnalysis/Evaluator.cs
+++ b/src/Core/CodeAnalysis/Evaluator.cs
@@ -40,6 +40,7 @@ public sealed partial class Evaluator
     private readonly Dictionary<VariableSymbol, object> globals;
     private readonly bool evaluateEntryPoint;
     private readonly bool useEntryPointReturnType;
+    private readonly StructValue.ObjectOverrideDispatcher objectOverrideDispatcher;
 
     // Issue #1651: everything below this point used to be an ordinary
     // instance field. That is correct only because the interpreter was
@@ -127,6 +128,7 @@ public sealed partial class Evaluator
         globals = variables;
         this.evaluateEntryPoint = evaluateEntryPoint;
         this.useEntryPointReturnType = useEntryPointReturnType;
+        objectOverrideDispatcher = TryEvaluateExternalOverride;
         Locals.Push(new ConcurrentDictionary<VariableSymbol, object>());
     }
 
--- a/test/Interpreter.Tests/Issue2896StructObjectOverrideTests.cs
+++ b/test/Interpreter.Tests/Issue2896StructObjectOverrideTests.cs
@@ -9,9 +9,11 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Threading.Tasks;
+using GSharp.Core.CodeAnalysis;
 using GSharp.Core.CodeAnalysis.Compilation;
 using GSharp.Core.CodeAnalysis.Symbols;
 using GSharp.Core.CodeAnalysis.Syntax;
+using GSharp.Repl.Engine;
 using GSharp.Tests;
 using Xunit;
 using CompilerProgram = GSharp.Compiler.Program;
@@ -135,6 +137,51 @@ public class Issue2896StructObjectOverrideTests
         AssertIssue3134EmittedSemantics(surface, emitted);
         Assert.Equal(emitted, await RunDriverAsync(source, suffix + "Evaluate", "gsc-evaluate"));
         Assert.Equal(emitted, await RunDriverAsync(source, suffix + "Gsi", "gsi"));
+    }
+
+    [Fact]
+    public void Issue3135_NoOverrideDispatchReusesDelegateWithoutPerCallClosure()
+    {
+        const string Source = """
+            package Issue3135
+            import System
+
+            struct Empty {
+            }
+
+            let first = Empty{}
+            let second = Empty{}
+            Console.WriteLine(first.Equals(second))
+            """;
+
+        var engine = new SessionEngine { CaptureConsole = true };
+        var cell = engine.Evaluate(Source);
+
+        Assert.False(cell.HasError, string.Join(Environment.NewLine, cell.Diagnostics));
+        Assert.Equal("True\n", cell.Output);
+
+        var values = engine.Variables.Values.OfType<StructValue>().ToArray();
+        Assert.Equal(2, values.Length);
+
+        var dispatcherField = typeof(StructValue).GetField(
+            "objectOverrideDispatcher",
+            BindingFlags.Instance | BindingFlags.NonPublic);
+        Assert.NotNull(dispatcherField);
+        var dispatcher = dispatcherField.GetValue(values[0]);
+        Assert.NotNull(dispatcher);
+        Assert.Same(dispatcher, dispatcherField.GetValue(values[1]));
+
+        _ = values[0].GetHashCode();
+        var before = GC.GetAllocatedBytesForCurrentThread();
+        for (var i = 0; i < 100_000; i++)
+        {
+            _ = values[0].GetHashCode();
+        }
+
+        var allocated = GC.GetAllocatedBytesForCurrentThread() - before;
+        Assert.True(
+            allocated < 14_000_000,
+            $"100,000 no-override hash calls allocated {allocated:N0} bytes; expected no per-call LINQ closure.");
     }
 
     [Theory]


DavidObando:
Closing this implementation after ADR-0156 Phase 3a/3b.3 removed every default gsi evaluator path. On current main 5cd0d766a, the regression still measures on the retiring evaluator only: paired parent/main median 0.926x through Compilation.Evaluate and 0.926x through SessionEngine (40 alternating pairs each). An asserted StructValue dispatch-condition mutant changed evaluator output from FORMAT-TOSTRING-33 to FORMAT-Value(Number=11), while rebuilt default gsi and the default EmittedSessionEngine remained byte-for-byte FORMAT-TOSTRING-33. The embedding API is now [Obsolete] and SessionEngine is the deprecated Phase 3c escape hatch. No product fix is justified; recommend closing #3135. I am not closing the issue.

DavidObando:
## Review — PR #3237 @ `b27813bc` — **MERGEABLE**

**0 BLOCKER · 2 SHOULD-FIX · 2 NIT.** The product change is correct, minimal (+18/−6 across two files), and the headline correctness hazard is measured clean with a negative control. Two findings concern claims in the body, not the code.

Environment: fresh worktree at `b27813bc`, own `out/`, `dotnet build -c Release -p:ContinuousIntegrationBuild=true` → rc 0, 0 warnings. Witness `GSharp.Core.dll` md5 `6edcb5ab…`; every mutant asserted by `git diff HEAD` **and** md5 movement, restored by exact md5 round-trip.

**Base/ancestry.** Merge-base `d6e7c4c3`; `main` is now `5cd0d766` (the briefed `e14987354` is an ancestor — main moved 8 commits). `git merge-tree` → rc 0, **0 `CONFLICT` lines**. Cherry-picked onto `5cd0d766`: 3 files, 59/6 preserved, build **0 warnings / 0 errors**, new `EvaluatorRetirementGateTests` **1 passed**, full `Interpreter.Tests` **1500 passed / 4 skipped / 0 failed**, and the new test passes by name (1 test — not a zero-match). CI `ilverify` job green.

---

### Q1 — Dispatcher ownership: **clean, and the PR does not change binding at all**

Ownership chain:

| Step | Site | Fact |
|---|---|---|
| Field | `src/Core/CodeAnalysis/Evaluator.cs:43` | `private readonly` **instance** field — not static |
| Assignment | `src/Core/CodeAnalysis/Evaluator.cs:131` | in the internal ctor; the public ctor (`:112`) **chains** to it, so there is no path leaving it `null` |
| Target | `src/Core/CodeAnalysis/Evaluator.Calls.cs:1065` | `TryEvaluateExternalOverride` is **non-static** → delegate bound to `this` Evaluator |
| Handoff | `Evaluator.Calls.cs:1063` | `CreateStructValue` (instance) passes it to each `StructValue` |
| Lifetime | `src/Core/CodeAnalysis/Compilation/Compilation.cs:393` | the **only** `new Evaluator` site → one Evaluator, hence one dispatcher, per `Evaluate` call / REPL submission |

`grep -rn "static.*ObjectOverrideDispatcher\|ObjectOverrideDispatcher.*static" src/` → **no matches**. The one static cache in this area, `StructValue.LayoutRuntime.Cache`, holds only `symbol`/`fields`/`Type`/`Size` and takes the dispatcher as a **parameter**, so it cannot retain one.

**The key point: this PR does not change which evaluator is bound.** Pre-PR `new(structType, TryEvaluateExternalOverride)` already created a `this`-bound delegate at every call. The PR changes the *allocation count* (N→1), not the binding. There is no new escape surface.

Measured in-process (out-of-tree harness referencing built DLLs **by path**, no `ProjectReference`, witness md5 unchanged afterwards). Each engine declares `struct Tag` whose `ToString()` override returns a **package-level global**, so the printed value proves which evaluator's globals ran the body:

| Probe | Result |
|---|---|
| Two engines concurrent (barrier-synced) | `CONC-A` / `CONC-B` — correct |
| 200 alternating dispatches across two live engines | correct; dispatchers are distinct objects |
| 40,000 concurrent dispatches (20k × 2 threads) | `badA=0 badB=0` — zero cross-talk |
| Value from submission 1 re-dispatched after a session boundary | correct (`SEQ-1`); dispatcher objects differ across submissions, confirming per-Evaluator |
| 3 `StructValue`s from one evaluator | **exactly 1** distinct dispatcher — the optimization does what it claims |

`FAILS=0`.

**Negative control (the probe can fail).** Made the field `static` (`Evaluator.cs:43`), rebuilt — witness moved `6edcb5ab…` → `f47cbab0…` — and the probe went **RED**: `KeyNotFoundException: The given key 'func ToString() string' was not present in the dictionary`. That is exactly the predicted cross-evaluator failure mode, from `program.Functions[externalOverride]` at `Evaluator.Calls.cs:1100` resolving against the wrong evaluator's program. Restored to md5 `6edcb5ab…`.

Worth recording for future readers: a cross-evaluator leak here would **throw**, not answer silently, because the override body is fetched from the bound evaluator's own `program`.

---

### SHOULD-FIX 1 — Q2: both "surviving" mutants are **killable**; the corpus is structurally blind

Neither line is dead. The corpus simply cannot reach the evaluator.

Every override-bearing test in `test/Interpreter.Tests/Issue2896StructObjectOverrideTests.cs` runs on an **emit-only** driver after ADR-0156:

- `Evaluate()` → `EmittedOracle.Evaluate` (`:347`)
- `"gsc-evaluate"` → `CompilerProgram.Main(file)` (`:758`) — emits
- `"gsi"` → `Repl.Program.Main(file)` (`:760`) — emits

The file's **only** evaluator-driving test (`:334-342`, via `EvaluateWithEvaluator` at `:358`) deliberately uses structs with **no** overrides, so `FindExternalOverride` returns `null` under both the original and the inverted condition. The PR's own new test is in the same position: `struct Empty` has no overrides.

Measured, both mutants at `Evaluator.Calls.cs:1110`:

| Mutant | Issue2896 corpus | Full `Interpreter.Tests` | Full `Core.Tests` | In-process probe |
|---|---|---|---|---|
| M1 `!method.IsOverride` | **47/47 pass** | **1493, 0 failed** | **7153, 0 failed** | **RED — 5 FAILs** |
| M2 `…?.Equals(externalMethod) != true` | **47/47 pass** | — | — | **RED — 5 FAILs** |

Under M1, external override dispatch is disabled entirely and **8,646 tests do not notice**; the probe reports `Tag()` instead of `ALT-A`. Both mutants moved the witness md5 (`34153a62…`, `9e7e2667…`) and were restored to `6edcb5ab…`.

So the evaluator's external-override dispatch path — the subject of this PR — currently has **zero coverage** in the two projects that could cover it. That is pre-existing, not caused here, but the table's "SURVIVED / Not scored" understates it: this is an unpinned product line, and the same signal was misread as "dead code" in #3159.

**Fix is cheap and belongs in this PR**, which already added a `SessionEngine` test to that file: give the struct an override and assert the dispatched value.

```gs
struct Tag {
    override func ToString() string -> "OVERRIDDEN-3135"
}
```

then assert the dispatched string from the in-process `SessionEngine`. That single specimen kills **both** mutants (verified: it is what my probe does).

### SHOULD-FIX 2 — Q5: the reachability paragraph is now stale

The body says: *"The embedding API remains reachable; `SessionEngine` remains reachable through the documented evaluator escape hatch."*

On current `main` `5cd0d766` — via **#3242**, which landed *after* this PR's base `d6e7c4c3` — **all three `Compilation.Evaluate` overloads are `[Obsolete]`** (`src/Core/CodeAnalysis/Compilation/Compilation.cs:311`, `:321`, `:332`):

> `The tree-walking evaluator retires in ADR-0156 Phase 3c. New callers are forbidden; see #3176.`

Both `Evaluator` constructors carry the same attribute, and my own rebuilt `gsi --help` says `--engine evaluator` "is deprecated and will be removed in ADR-0156 Phase 3c".

So **both** named consumers are formally scheduled for deletion, not just the REPL hatch. This does not block: the evaluator ships until 3c, the change is 18 product lines, and it is correct. But the durability argument as written no longer holds and should be restated honestly — the optimization's value is bounded by Phase 3c.

### NIT 1 — Q3: direction is significant on both surfaces; the *magnitude* is only well-estimated on one

Stating it explicitly, as it should be in the body rather than left to the reader:

| Claim | Two-sided sign test |
|---|---|
| `Compilation.Evaluate` 31/40 | **p = 6.8e-4** |
| `SessionEngine` 37/40 | **p = 1.9e-8** |

(p<0.05 needs 27/40 at n=40.) So **yes — the `Compilation.Evaluate` result is distinguishable from noise.** The direction is real on both surfaces.

The *magnitude* is not equally supported. On `Compilation.Evaluate` the paired ratio is σ 0.165 over 0.675–1.750, versus σ 0.075 over 0.937–1.269 on `SessionEngine`; and the raw main-side sample carries a 294.777 ms maximum against a 129.719 median (σ 35.521) — outlier-contaminated. The sign test is robust to that because it uses direction only; the ~6.1% point estimate is not. Also, a load threshold of 30 admitting 8.18–20.68 spans a ~2.5× contention range, far wider than a 6% effect — mitigated, not eliminated, by A/B alternation and pairing. (For calibration: this machine ran at load 9–10 throughout my review.)

Suggest attributing "~6.2% recovery" to `SessionEngine` and reporting `Compilation.Evaluate` as same-direction with wider spread.

### NIT 2 — Q4: the 14 MB guard is sound; only its failure message misleads

`test/Interpreter.Tests/Issue2896StructObjectOverrideTests.cs:175-185` uses `GC.GetAllocatedBytesForCurrentThread` — the stable API, not `GetTotalMemory`. Measured on this head:

```
run0: allocated=8,800,000 bytes (88.0 B/call)
run1: allocated=8,800,000 bytes (88.0 B/call)
run2: allocated=8,800,000 bytes (88.0 B/call)
```

**Zero variance across runs.** Guard 14,000,000; pre-fix 17,600,000. The guard sits between a 2× signal with 1.59× headroom and genuinely discriminates. This is **not** a timing flake — the counter is deterministic, so drift can only come from a real product change in this path.

One small thing: 88 B/call still remains after the fix, so the failure message *"expected no per-call LINQ closure"* would misattribute the cause of any future trip. Consider wording it as a budget ("allocated N bytes; budget 14 MB") rather than naming a cause that is no longer the only contributor.

---

### Ponytail

Product diff is **+18/−6 across two files**; the rest is test. The `FirstOrDefault` → `foreach` shape is minimal — LINQ has no non-capturing predicate overload carrying state, so the explicit loop is required to remove the closure. The rewrite is exactly equivalent: same base-chain walk, same declaration order, same first-match semantics, same early exit. Correctly declines to resurrect #3194's cache. No smaller shape exists.

DavidObando:

Issue #3230, opened 2026-08-04
[gsi] --engine evaluator is validated then silently ignored in script mode; probes through it measure the emitted engine


DavidObando:
`gsi` accepts `--engine evaluator` (and `GSI_ENGINE=evaluator`), validates it, and then **silently ignores it when running a script file**. Script mode always uses the emitted engine.

`src/Repl/Program.cs`:

- `:53-71` parse `engineChoice` from `--engine <name>` / `--engine=<name>`
- `:85` `engineChoice ??= EngineChoiceFromEnvironment();`
- `:86-88` **validate** it, rejecting anything but `emit` / `evaluator` with `Unknown engine '<x>'. Expected 'emit' or 'evaluator'.`
- `:100` `return RunInteractive(engineChoice, references);` — passes it
- **`:109` `return RunScript(scriptPath, references);` — drops it**
- `:161` `RunScript(string scriptPath, List<string> references)` — no engine parameter; always `EmittedProgramHost`

So the flag is validated hard enough to reject typos, which makes it look supported, and then has no effect. `gsi --engine evaluator probe.gs` runs **emitted** code and exits 0.

## Why this matters more than it looks

Since ADR-0156 Phase 3a (#3201) made emit the interactive default, `--engine evaluator` is the **only** documented way for a user to reach the tree-walking evaluator from the command line. In script mode that escape hatch does not exist, and it fails silently rather than erroring.

**It also silently invalidates measurements.** This was found by a reviewer who had been asked to verify evaluator-boundary behaviour (`GS0514`) and initially probed with `gsi --engine evaluator probe.gs` — which measures the *emitted* engine and can prove nothing about evaluator boundaries. Their words:

> `gsi --engine evaluator probe.gs` silently measures the *emitted* engine and proves nothing about GS0514. I drove `GSharpRepl` → `SessionEngine` in-process instead.

A probe that silently measures the wrong engine and exits 0 is indistinguishable from a correct one. Any past or future result gathered through script-mode `--engine evaluator` is suspect.

## Expected

One of:

1. **Honour it** — thread `engineChoice` into `RunScript` so script mode uses the evaluator when asked (symmetric with `RunInteractive`, and consistent with the flag surviving until Phase 3c); or
2. **Refuse it** — if script mode is deliberately emit-only, `gsi --engine evaluator file.gs` should fail with a clear diagnostic saying the flag applies only to the interactive session.

[...]

PR #3207, opened 2026-08-04
Fix website driver execution documentation

DavidObando:
## Summary

- correct website prose: every command-line driver emits CIL by default, including interactive `gsi`
- document `--engine evaluator` / `GSI_ENGINE=evaluator` as deprecated evaluator escape hatches scheduled for Phase 3c removal
- keep `Compilation.Evaluate` identified as the evaluator-backed embedding API
- clarify that `/out` controls whether `gsc` saves the assembly, not which engine runs
- remove false lambda-to-CLR-delegate limitations from file-mode docs
- leave `website/docs/ref/diagnostics.md` untouched for #3200

## Rebuilt `gsi --help`

Measured after rebasing onto `main` `0484b00f139f` (contains `8b32fc36`):

```text
  --engine <name>        Interactive engine: 'emit' (default) or 'evaluator'
                         (also via GSI_ENGINE). 'evaluator' selects the legacy
                         tree-walking engine; it is deprecated and will be
                         removed in ADR-0156 Phase 3c.
```

## Measured routing

| invocation | rc | stdout / assertion | GS0510 | engine |
|---|---:|---|---:|---|
| bare `gsc samples/Deinit.gs` | 0 | golden + `Success.` | 0 | emit |
| `gsc /out:Deinit.dll` → `dotnet Deinit.dll` | 0 / 0 | golden | 0 | emit |
| `gsi samples/Deinit.gs` | 0 | golden | 0 | emit |
| interactive default / explicit `--engine emit` | — | engine-selection tests select emitted engine; deinit test prints `deinit-22`, `body-33` | 0 | emit |
| `--engine evaluator` / `GSI_ENGINE=evaluator` | — | engine-selection tests select evaluator | 1 in deinit test | evaluator |
| embedding `Compilation.Evaluate` | 0 | `body-33`; deinit skipped | 1 | evaluator |

Lambda-to-delegate probe (`nums.Where(func(n int32) bool { return n % 2 == 0 }).Count()`): bare `gsc` rc 0, stdout `2` + `Success.`.

## Residue sweep

```bash
[...]

PR #3243, opened 2026-08-04
Honor gsi engine selection in script mode

DavidObando:
## Summary

- pass the validated engine choice into script execution
- use existing `Program.UsesEvaluatorEngine` predicate for both script and interactive routing
- run evaluator-requested scripts through the existing `SessionEngine` whole-program path
- clarify `--help`, ADRs, diagnostics docs, and website mirror: `--engine` / `GSI_ENGINE` apply to scripts and interactive sessions

## Design

**Honor the option.** Script mode already retained the legacy `SessionEngine` whole-program machinery (`RunEntryPoint`) and reference resolver needed for evaluator execution. Threading `engineChoice` to that existing path is smaller and matches the documented escape hatch until Phase 3c; refusing it would create an unnecessary mode-specific restriction.

One decision surface remains: `Program.UsesEvaluatorEngine`.

## RED proof

Test pin alone on unfixed `20fbbec41`:

```text
TEST_RC=1
Failed: 1, Passed: 0, Total: 1
Expected: 1 (emitted path returned 0)
```

Tree restored to porcelain-0 before applying production fix. Inverted product condition mutant:

```text
MUTANT_TEST_RC=1
Failed: 1, Passed: 1, Total: 2
```

Restored production condition: 2/2 green.

## Measured routing

```text
default script:             rc 0, stdout emitted-only-11, no GS0514
--engine emit script:       rc 0, stdout emitted-only-11, no GS0514
--engine evaluator script:  rc 1, empty stdout, GS0514
GSI_ENGINE=evaluator script: rc 1, empty stdout, GS0514
```

Rebuilt help:

```text
--engine <name>        Script and interactive engine: 'emit' (default)
                       or 'evaluator' (also via GSI_ENGINE).
                       'evaluator' selects the legacy tree-walking engine;
                       it is deprecated and will be removed in Phase 3c.
```

## Validation

Final tree rebased onto `5cd0d766a60f`:

- `dotnet build GSharp.sln -c Release`: 0 warnings, 0 errors
- full 9-project Release suite: 15,533 passed, 5 skipped, 0 failed
- issue tests: 2/2 passed
- ILVerify: 123/123 verified; 2 documented suppressions, 0 failures
- worktree clean

Addresses #3230

### DIFF
--- a/docs/adr/0068-deinit-destructor-support.md
+++ b/docs/adr/0068-deinit-destructor-support.md
@@ -115,7 +115,7 @@ changed which invocations use that backend. Current routing is:
 | `gsc /out:program.dll file.gs`, then run the assembly | emit to file, then CLR host | real CLR finalizer; no GS0510 |
 | `gsi file.gs` | emit to memory, then `EmittedProgramHost.Run` | real CLR finalizer; no GS0510 |
 | interactive `gsi` (default) | `EmittedSessionEngine` | real CLR finalizer; no GS0510 |
-| interactive `gsi --engine evaluator` (deprecated; removed in Phase 3c) | `SessionEngine` → `Compilation.Evaluate` | skips the body; GS0510 once per declaring class |
+| `gsi --engine evaluator` (script or interactive; deprecated; removed in Phase 3c) | `SessionEngine` → `Compilation.Evaluate` | skips the body; GS0510 once per declaring class |
 
 The old two-column `gsc`/`gsi` table made object reachability appear to be the
 discriminating axis. It is not. Driving the legacy evaluator-backed
--- a/docs/adr/0152-interpreter-native-call-boundary.md
+++ b/docs/adr/0152-interpreter-native-call-boundary.md
@@ -33,10 +33,10 @@ fabricated results and delayed `GS9999` failures, and adds no native or
 reflection dispatch path.
 
 Since ADR-0156 Phases 1–3a, this boundary applies to `SessionEngine`,
-`Compilation.Evaluate`, and interactive `gsi --engine evaluator`, not to
-default drivers. Bare `gsc`, `gsi <file>`, and the default interactive REPL
-emit and run the native call; `gsc /out:` emits it to disk. The evaluator path
-is deprecated and scheduled for removal in Phase 3c.
+`Compilation.Evaluate`, and `gsi --engine evaluator` in script or interactive
+mode, not to default drivers. Bare `gsc`, `gsi <file>`, and the default
+interactive REPL emit and run the native call; `gsc /out:` emits it to disk.
+The evaluator path is deprecated and scheduled for removal in Phase 3c.
 
 ## Consequences
 
--- a/docs/adr/0153-interpreter-compiled-only-storage-boundary.md
+++ b/docs/adr/0153-interpreter-compiled-only-storage-boundary.md
@@ -38,11 +38,11 @@ cannot rely on surrounding rendering for meaning.
 Constructs whose interpreter behavior requires real address identity —
 including pinning, stack allocation, unmanaged-pointer storage or dereference,
 and function-pointer execution — are **compiled-only in the tree evaluator**.
-Interactive `gsi --engine evaluator` must report a self-contained boundary
-diagnostic instead of attempting a value-only approximation. Since ADR-0156
-Phases 1–3a, all default drivers use emitted execution and run these constructs
-natively. The evaluator path is deprecated and scheduled for removal in
-Phase 3c.
+`gsi --engine evaluator` in script or interactive mode must report a
+self-contained boundary diagnostic instead of attempting a value-only
+approximation. Since ADR-0156 Phases 1–3a, all default drivers use emitted
+execution and run these constructs natively. The evaluator path is deprecated
+and scheduled for removal in Phase 3c.
 
 This ADR governs only the evaluator's storage-model boundary. It does not
 redefine unsafe/native language validity or CIL emission.
--- a/docs/adr/0156-gsi-emit-to-memory-execution.md
+++ b/docs/adr/0156-gsi-emit-to-memory-execution.md
@@ -109,15 +109,17 @@ direct `Compilation.Evaluate` consumers until Phase 3c.
 | bare `gsc file.gs` | `EmittedProgramHost.Run` |
 | `gsc /out:program.dll file.gs` | emit PE to disk; the CLR runs it separately |
 | `gsi file.gs` | `EmittedProgramHost.Run` |
+| `gsi --engine evaluator file.gs` | `SessionEngine` → `Compilation.Evaluate` |
 | interactive `gsi` | `EmittedSessionEngine` (default) |
 | interactive `gsi --engine evaluator` | `SessionEngine` → `Compilation.Evaluate` |
 
 Therefore the former "three-driver" model no longer identifies three
 execution semantics: every default driver compiles through the emitter. The
-tree evaluator remains reachable only through its deprecated interactive
-escape hatch and direct API/test consumers. This partially supersedes
-ADR-0068's original `deinit` interpreter boundary (and ADR-0152/ADR-0153) for
-default drivers, while leaving those boundaries in force for evaluator residue.
+tree evaluator remains reachable only through its deprecated script or
+interactive escape hatch and direct API/test consumers. This partially
+supersedes ADR-0068's original `deinit` interpreter boundary (and
+ADR-0152/ADR-0153) for default drivers, while leaving those boundaries in
+force for evaluator residue.
 
 ### Phased migration plan
 
--- a/docs/diagnostics.md
+++ b/docs/diagnostics.md
@@ -324,9 +324,10 @@ self-contained boundary messages through legacy `GS9999`; issue
 [#3199](https://github.com/DavidObando/gsharp/issues/3199) tracks moving those
 deliberate failures out of the internal-error category. Since
 [ADR-0156](adr/0156-gsi-emit-to-memory-execution.md) Phase 3a the boundary
-applies only to direct tree evaluation and `gsi --engine evaluator`; all
-default drivers execute emitted code, where these constructs run natively.
-The evaluator path is deprecated and scheduled for removal in Phase 3c.
+applies only to direct tree evaluation and `gsi --engine evaluator` in script
+or interactive mode; all default drivers execute emitted code, where these
+constructs run natively. The evaluator path is deprecated and scheduled for
+removal in Phase 3c.
 
 ## Documentation diagnostics (GS0227–GS0231)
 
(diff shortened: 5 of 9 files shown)


DavidObando:
## Rubber-duck review — PR #3243

**Verdict: MERGEABLE with 1 SHOULD-FIX.** 0 BLOCKER · 1 SHOULD-FIX · 2 NIT.

Reviewed the **merged** state. Head `64910d44` is based on `5cd0d766a`; `main` has since moved to `a5de8fd0c` (#3207) and is **not** an ancestor. `git merge-tree` output: **0 `CONFLICT` lines** (grepped for `CONFLICT`, not for markers — git never emits those), merged tree `0691c6b5`. No file overlap with #3207. Fresh worktree, own `out/`, Release **0 W / 0 E**.

### Q1 — behavioural proof, both directions, through the real binary

The failure mode of #3230 is *a probe that measures the wrong engine and exits 0*, so plumbing inspection is disqualified as evidence. Discriminator: a `GS0513` compiled-only storage construct that **emit runs and the evaluator refuses**. Sanity-checked first — `gsc unsafeprobe.gs` prints `42`, rc 0.

Same probe, same file, **pre-fix binary built from `5cd0d766a`** vs **fixed binary**:

| invocation | base `5cd0d766a` | HEAD `64910d44` |
|---|---|---|
| `gsi f.gs` (default) | rc 0, `42` — emit | rc 0, `42` — **emit** |
| `gsi --engine evaluator f.gs` | **rc 0, `42` — emit (the bug)** | **rc 1, `GS0513` — evaluator** |
| `GSI_ENGINE=evaluator gsi f.gs` | **rc 0, `42` — emit (the bug)** | **rc 1, `GS0513` — evaluator** |
| `gsi --engine emit f.gs` | — | rc 0, `42` — **emit** |
| `gsi --engine evaluatorr f.gs` | — | rc 1, `Unknown engine 'evaluatorr'. Expected 'emit' or 'evaluator'.` |

`rc` captured before any pipe. **The env-var path is threaded too** — the PR description only mentions the flag, so I measured `GSI_ENGINE` separately; it works. Validation is not weakened: the typo message is byte-identical to the pre-fix one. Both `RunScript` branches write diagnostics through `Console.Error.WriteDiagnostics`, so the stream contract is consistent.

Exit-code propagation through the new `cell.Value switch` — `func Main() int32 { return 7 }` gives **rc 7 on both engines**. That is new, otherwise-untested code and it is correct.

### Q2 — the "1/2" is a denominator artifact, not a weak oracle

I reproduced the author's mutant rather than crediting the count. Inverting `UsesEvaluatorEngine` to `!=` gives exactly **Failed: 1, Passed: 1, Total: 2**. The survivor is **`HelpStatesEngineOptionAppliesToScriptsAndInteractiveSessions`**, which asserts the `--help` string contains `"Script and interactive engine"` — it cannot observe the engine predicate **by construction**. The killer, `EvaluatorEngineOptionRunsScriptWithTreeWalkingEvaluator`, is a genuine oracle: `GS0514` boundary, exit code 1, empty stdout, and an explicit `DoesNotContain("emitted-only-11")` so an emitted run cannot pass it silently. **Effective score is 1/1 among specimens that can see the condition.**

I then ran a second, sharper mutant the author did not: `=> engineChoice != "emit"`, which **flips the default to the deprecated evaluator**. The new test file survives it **2/2** — so on the narrow corpus it reads as an unpinned invariant. It is not: the wider `Interpreter.Tests` project kills it **28×** (`ScriptRunnerRunsDeinitializersWithoutBoundaryWarning`, `StackAllocExecutes`, `FixedArrayPinsAndExecutes`, `BatchFileRunner_ExecutesPInvokeNatively`, `InheritedDeinitializersRunOnEveryDriver(driver: Interpreter)`, …). **No coverage gap; the PR correctly does not re-pin what is already pinned.** Recording this explicitly because inferring "unpinned" from a narrow surviving mutant is what produced the #3159 regression.

Restore verified the hard way: `git diff HEAD` empty **and** `git status --porcelain` empty **and** a clean re-run at **1501 passed / 4 skipped / 1505 total** — same denominator as the mutant runs, so no test-selection drift.

### Q3 — one predicate, confirmed by grep

`UsesEvaluatorEngine` is the **sole** engine decision, with exactly two call sites: `src/Repl/Program.cs:144` (interactive) and `:165` (script). No second engine test anywhere in `src/`. `GSI_ENGINE` is normalised into `engineChoice` at `:85`, **before** validation at `:86` and before either branch, so both modes inherit the env fallback through one variable. No divergence surface.

### Q5 — scope is clean

`RunEntryPoint` and `SessionEngine` both pre-date the PR (`5cd0d766a:src/Repl/Engine/SessionEngine.cs:175`) and are already exercised by three test files. **Zero new evaluator machinery** — the fix is 20 lines of driver plumbing, which is the right size for something deleted in Phase 3c.

ILVerify **123/123**, rc 0, and `grep -c "Assembly or module not found"` = **0** (the no-result-wearing-a-pass check). CI 31 SUCCESS / 3 SKIPPED / 0 failing.

---

## SHOULD-FIX 1 — this PR invalidates a verbatim `--help` transcript in the website docs

**`website/docs/tooling/repl.md:36-47`** is a ```text``` block presenting literal `gsi --help` output. It still reads:

```
  --engine <name>        Interactive engine: 'emit' (default) or 'evaluator'
```

Six commits ago, during #3207, I diffed that block against the built binary and it was **byte-for-byte identical**. It no longer is — `diff` against my rebuilt `gsi --help` now shows a 4-line divergence, and the stale side says **"Interactive engine"**, which is precisely the false scoping this PR exists to remove. The PR updated the help string and swept `docs/` thoroughly (0068, 0152, 0153, 0156, `diagnostics.md`, `emit-pipeline.md`) and correctly updated the `website/docs/ref/diagnostics.md` mirror in lockstep — but missed the one website file that quotes the changed string as command output.

This is worse than ordinary stale prose: it is a **transcript**, so readers take it as ground truth from the binary, and a user who reads it will believe `--engine evaluator` does nothing for scripts — the belief #3230 was filed to kill. Fix is a 4-line paste of the new output.

Worth noting the structural cause: `git grep "Usage: gsi"` returns **exactly one hit repo-wide**, that doc block. Nothing pins it. The PR pins the *binary* side (`Assert.Contains("Script and interactive engine", …)`) but the doc transcript has no generator and no test, so it will drift again.

## NIT 1 — `website/docs/faq.md:74` now under-scopes

> `gsi --engine evaluator` or `GSI_ENGINE=evaluator` selects the deprecated **interactive** evaluator

True before this PR, incomplete after it — the flag now selects the evaluator for scripts too. Same one-word fix as the `docs/` sweep applied elsewhere.

## NIT 2 — unhandled-exception protocol differs between engines in script mode (report only, no fix requested)

`throw InvalidOperationException("boom")` in a script:

- `gsi boom.gs` → **rc 134**, `Unhandled exception. System.InvalidOperationException: boom` + stack frame (the deliberate CLR-host protocol, `Program.cs:214-219`)
- `gsi --engine evaluator boom.gs` → **rc 1**, `error GSI001: Evaluation error: boom`, no stack

This is inherited from `SessionEngine` and is identical to what interactive `--engine evaluator` already did, so reusing it was the right call over building a parallel protocol for an engine deleted in Phase 3c. It is also **self-announcing** — `GSI001` is unmistakably evaluator-flavoured, unlike #3230's silent success. Flagging only so nobody comparing crash behaviour across engines reads `134` vs `1` as a product difference.

---

**Also resolved by this PR:** my #3207 NIT 3 against `website/docs/tooling/repl.md:122-124` — *"Pass `--engine evaluator` … to select the deprecated in-process evaluator"* immediately after naming `gsi file.gs` as a driver. That sentence was misleading only because of #3230; it is now simply true.

**Suites run:** `Interpreter.Tests` (1505 total, clean, plus two mutant rounds), ILVerify 123/123. **Not run:** the other 8 of 9 projects — the diff touches one driver file and its own test, and CI covers the rest at 31/31.

DavidObando:
## Rubber-duck round 2 — **I am retracting my MERGEABLE.** Verdict is now **BLOCKER: obsolete, recommend close.**

`main` moved again while I was writing the first verdict. It is now **`8b3bdac57` — "Retire the tree-walking evaluator (ADR-0156 Phase 3c) (#3244)"**. **Phase 3c has landed.** My earlier verdict was measured against `a5de8fd0c` and is void.

### The premise of this PR no longer exists

#3244 **deleted the evaluator**, ~7,500 lines: `Evaluator.cs` and all six `Evaluator.*.cs` partials, `EvaluatorException.cs`, `EvaluationResult.cs`, 128 lines of `Compilation.cs`, and — decisively for this PR — **`src/Repl/Engine/SessionEngine.cs`**, the exact type this fix instantiates:

```csharp
new Engine.SessionEngine(resolver) { RunEntryPoint = true }.Evaluate(source, scriptPath)
```

`git grep "class SessionEngine" -- src/` on main returns **nothing**. `Program.UsesEvaluatorEngine` — the single predicate I verified in round 1 — is gone too, replaced by `IsValidEngineChoice(engineChoice) => engineChoice is null or "emit"`. `GS0510`, `GS0513`, `GS0514` are all **absent from `DiagnosticDescriptors.cs`**, so the new test's `GS0514` oracle asserts a diagnostic that can no longer be produced.

### #3230 is already fixed on main — by deletion, and more strongly

I built current `main` and re-ran my round-1 probe file unchanged:

| invocation | `main` `8b3bdac57` |
|---|---|
| `gsi f.gs` | rc 0, `42` — emit |
| `gsi --engine evaluator f.gs` | **rc 1, `Unknown engine 'evaluator'. Expected 'emit'.`** |
| `GSI_ENGINE=evaluator gsi f.gs` | **rc 1, `Unknown engine 'evaluator'. Expected 'emit'.`** |
| `gsi --engine emit f.gs` | rc 0, `42` — emit |

The hazard #3230 describes — *a probe that silently measures the other engine and exits 0* — **cannot recur**: there is no engine to select, and both the flag and the env-var form now fail **loudly**. That is a strictly better outcome than honouring the flag, because it removes the class rather than repairing one instance. `rc` captured before any pipe.

### Merge state

`git merge-tree --write-tree origin/main <head>` → rc 1, **3 `CONFLICT (content)` lines**: `src/Repl/Program.cs`, `docs/diagnostics.md`, `website/docs/ref/diagnostics.md`. (Grepped for `CONFLICT`; git 2.50 never emits `<<<<<<<` here, and that grep returns 0 — failing toward clearing the PR.)

Note the trap in resolving those by hand: **`SessionEngine.cs` does not appear as a conflict**, because this PR only *references* the type without editing its file. A textual resolution of `Program.cs` therefore produces code referencing a **deleted type** — a compile failure that the conflict list does not warn you about.

### What still stands from round 1

Nothing to fix, but for the record: the fix was correct and minimal against its own base, the RED/GREEN matrix was real, `UsesEvaluatorEngine` was genuinely the single predicate, and the "1/2 mutant" was a denominator artifact rather than a weak oracle. **The work was not wrong; it was overtaken.**

The one durable item is my round-1 SHOULD-FIX, which **survives this PR's closure and is now a defect on `main`**: **`website/docs/tooling/repl.md:36-47`** carries a verbatim `gsi --help` transcript that is stale against the Phase 3c binary — it advertises `'evaluator'` as a selectable value that main now **rejects with rc 1**. `git grep "Usage: gsi"` still returns exactly one hit repo-wide and nothing pins it. That should be re-pointed at #3244's help text rather than this PR's.

**Recommendation: close as overtaken by #3244.** I am not closing it — reporting only.

DavidObando:

Issue #3240, opened 2026-08-04
gsc emits invalid IL for explicit ToString() on by-ref-like receiver with only inherited ToString


DavidObando:
## Summary

`gsc` accepts an explicit `ToString()` call on a by-ref-like (`ref struct`) receiver whose only candidate `ToString` is **inherited**, emits an executable, and the emitted program then dies at runtime with:

```
System.InvalidProgramException: CLR detected an invalid program.
```

Exit code **134**. The compiler reports **no diagnostic at all** — `rc=0`, `Wrote F2.exe`.

Worse: **`ilverify` passes the emitted assembly** (`All Classes and Methods ... Verified.`), so the required CI gate from #3164 green-lights an assembly the CLR refuses to JIT.

## Reproduction

Measured on `main` at `20fbbec41`.

```gs
ref struct Bare {
    var id int32
}

func main() {
    var b = Bare { id: 42 }
    print("interp=${b.ToString()}")
}
```

Observed:

| Case | Spelling | `gsc` | run |
|---|---|---|---|
| **F1** | `${b}` | **rc=1**, `GS0519` (by-ref-like has no available CLR ToString method) | n/a |
| **F2** | `${b.ToString()}` | **rc=0**, `Wrote F2.exe`, no diagnostics | **rc=134**, `System.InvalidProgramException: CLR detected an invalid program` |
| **G** (control) | `struct Plain` (not `ref`), both spellings | rc=0 | rc=0, prints `interp=Default.Plain` |

Control **G** isolates by-ref-like-ness as the cause: an ordinary `struct` with no declared `ToString` handles both spellings correctly.

## Mechanism

The compiler already names the rule elsewhere:

```
GS0219: By-ref-like type 'Bare' ... cannot be boxed
```

Calling an **inherited** virtual (`System.ValueType.ToString`) on a by-ref-like receiver requires **boxing the receiver**. Boxing a by-ref-like value is illegal. Calling a **declared** `ToString` on the same type is a direct call and is fine — a `ref struct` with an in-body `ToString() string` compiles, verifies, runs, and prints correctly through the explicit spelling.

So the discriminator is **declared vs inherited**, and the explicit-call path is not applying it. `GS0219` encodes this correctly in other positions; this call site does not consult it.

## Why this matters beyond the crash

1. **Silent wrong answer.** No diagnostic, successful emit, `Wrote ...exe`. The failure surfaces only when the user runs the program, with a CLR-level message that points at nothing in their source.
[...]

Issue #1629, opened 2026-07-02
Generic constructor inference double-binds every argument — duplicate diagnostics and exponential re-binding


DavidObando:
- **Category**: correctness / performance
- **Severity**: P2
- **Location**: src/Core/CodeAnalysis/Binding/OverloadResolver.cs:2152 (+2162, 2176) vs 2219-2225; 3016-3021 vs 2608-2628
- **Evidence**: Inference pre-bind in `BindConstructorCallExpression`:
```csharp
var preBound = bindExpression(argSyntax);
inferTypeArguments(defParams[paramIdx].Type, preBound.Type, substitution);
```
then the same arguments are bound again for the real call:
```csharp
foreach (var argument in syntax.Arguments)
{
    boundArguments.Add(BindOverloadArgumentValue(UnwrapNamedArgumentValue(argument)));
}
```
Same pattern in `TryCloseGenericExplicitConstructorType` (3016-3021) followed by `BindExplicitConstructorCallExpression`'s bind loop (2608-2628). `DiagnosticBag` is an append-only builder with no dedupe, and no `Diagnostics.TruncateTo(...)` rollback exists anywhere in OverloadResolver.cs (verified by grep; rollback is used only in ExpressionBinder.*).
- **Problem**: (1) Any diagnostic inside a constructor argument of a generic type (`Box(undefinedVar)`) is reported twice. (2) Every argument subtree is bound twice, and a nested generic construction (`Box(Box(Box(x)))`) re-binds its own arguments twice at each level — work grows as 2^depth — unlike the free-function generic path (lines 4553-4584), which binds once and infers from the already-bound arguments. The pre-bind also bypasses `BindOverloadArgumentValue`'s branchy-argument deferral, so a targetless conditional argument reports its suppressed error prematurely.
- **Suggested fix**: Bind the arguments once (with `BindOverloadArgumentValue`) before the inference block and drive inference from the bound results, mirroring the free-function generic call path; alternatively wrap the pre-bind in a `Diagnostics.TruncateTo` mark.
- **Confidence**: high

---
*Found by an automated multi-agent code review with Claude Code. `file:line` references are relative to commit `1e7e5a91`.*

Issue #942, opened 2026-06-22
'expr[i].Member' with identifier index mis-parses [i] as type arguments (GS0005); literal index works


DavidObando:
## Summary

Member access on an **element-access whose index is an identifier** — `expr[i].Member` — fails to parse: the `[i]` is consumed as a **generic type-argument list** rather than an indexer, so `expr[i].M(...)` mis-binds. The same expression with a **literal** index (`expr[0].Member`) parses and compiles correctly. This is a parser ambiguity between indexing `[…]` and generic type arguments `[…]`.

Discovered by the C#→G# migration tool (#914) translating `corpus/L3-Library` (`Algorithms.Max`: `values[i].CompareTo(best)`).

## Repro

```gs
package T
import System
let xs = []int32{3, 1, 2}
let i = 0
Console.WriteLine(xs[0].ToString())   // OK — literal index parses as indexer
Console.WriteLine(xs[i].ToString())   // FAILS — [i] parsed as type arguments
```

`gsc /target:exe`:

```
error GS0131: 'xs' is not a function.
    Console.WriteLine(xs[i].ToString())
error GS0159: Cannot find function WriteLine.
error GS0005: Unexpected token <DotToken>, expected <OpenParenthesisToken>.   (IReadOnlyList[T] variant)
```

## Control

`xs[0].ToString()` (literal index) compiles. Hoisting to a temp local (`let e = xs[i]; e.ToString()`) also compiles — confirming the defect is specifically `[identifier].Member` parsing.

## Expected

`expr[i].Member` parses as indexer-access followed by member-access regardless of whether the index is a literal or an identifier, identically to `expr[0].Member`. When a postfix `.` (or `(`) follows `name[args]`, the `[args]` should be treated as an indexer in expression context.

PR #3250, opened 2026-08-04
Honor ref struct ToString in interpolation

### LINKED ISSUE #3220: GS0519 rejects interpolation of user-declared ref structs whose ToString the compiler can already call; raw GetMethod at the same site bypasses GetMethodSafe
Follow-up from PR #3192 (issue #3183), which shipped ByRefLike interpolation rendering on the emitted paths. The reviewer verified the PR at head `4f1789ee` and returned MERGEABLE with two SHOULD-FIX items, recorded here with their measurements. Neither blocked the merge; both are real.

## 1. `GS0519` refuses a conversion the compiler demonstrably performs

`InterpolatedStringHandlerLowerer.cs` (the ToString-reroute site, ~`:698` at time of review — locate by symbol, the line will drift) looks only for a **CLR** `ToString`. For a **user-declared `ref struct`**, this means:

| specimen | result |
|---|---|
| `$"{token}"` where `Token` is a user `ref struct` with a receiver-clause `ToString` | `GS0519` |
| `$"{token}"` where `Token` declares the **canonical in-body `ToString`** (ADR-0079) | `GS0519` |
| **`${token.ToString()}`** — same type, same program | **rc=0, emits, runs, prints `interp=Token#42`** |

So the compiler rejects the interpolation while the explicit spelling of the identical conversion succeeds. This is the same "tell" that killed round 1 of #3192 — the workaround the diagnostic suggests works — surviving in a narrower form.

The message compounds it: **"no available CLR ToString method"** misdirects, because the user *has* declared a `ToString` and adding another changes nothing. At minimum the message should say a CLR-visible `ToString` is required and that a user-declared one is not currently honoured.

Reviewer established `GS0519` is genuinely **narrowed, not deleted** — it has a live trigger (`refstruct.gs(7,23,7,39): error GS0519`, source-anchored, all three drivers), and `TypeSymbol.cs:246-255` shows `IsByRefLike` has two arms with `FindByRefLikeToString` reading `?.ClrType?.`, returning null only on the source arm. So **`GS0519` ⟺ user-declared `ref struct`**; every imported ref struct inherits `ValueType.ToString` and renders. That scoping is correct as far as it goes — this issue is about closing the remaining arm.

## 2. Raw `GetMethod` where `GetMethodSafe` exists

The same site calls raw `GetMethod`. `ClrTypeUtilities.GetMethodSafe` (`:45`, 4 existing call sites) exists precisely to absorb `NotSupportedException` from constructed-generic-over-`TypeBuilder` — **the exception class that surfaces as `GS9998`, which is the very ICE #3183 was about**.

**Disclosed honestly by the reviewer: this was scored on convention, not as a live defect.** Two attempts to reach it failed for unrelated reasons — `Span[Item]` → `GS0155`, `Span[Color]` → `GS0005`. So it is **unreached, not proven unreachable**. The swap is a near-free hardening at a site whose failure mode is a known ICE; it should either be made, or the asymmetry with the other four call sites explained.

## 3. NIT — divergence from C# on `Span<int>`, pre-existing

Real C# **rejects** `$"{Span<int>}"` with `CS9244`; G# renders `System.Span<Int32>[3]`. **Not introduced by #3192** — byte-identical to pre-regression `aa453ac6`, i.e. #3128 behaviour. Recorded so it is not rediscovered as a regression.

## Context worth preserving

Two claims in the original #3183 brief were **wrong in the PR's favour** and were corrected by measurement during review:

- **The emit path was never working.** On pre-regression `aa453ac6`, `gsc /out:` **already ICE'd with `GS9998`**. Only the evaluator-backed drivers rendered. #3182 removed masking; it did not create the hole.
- **For the user-declared `ref struct` case, `aa453ac6` was worse than an ICE.** `gsc /out:` reported rc=0 and `Wrote ptok.exe` — and the produced executable **aborts with rc=134, `VerificationException`**. Silent unverifiable IL. `GS0519` replaces that with a compile-time error, making the merged state a strict improvement over both the base and the pre-regression tree.

Rendering verified on the merged head: `chars=hello`, `writableChars=hello`, `aligned=   hello` (`${ros,8}`, 3 leading spaces); `{ros:X}`, `{ros,8:X}`, `{ros,-8}` all byte-identical to real C#. `gsc /out:` ≡ `gsi` byte-identical (`18a0515f`).

DavidObando:
Closes #3220.

## What changed

- Source `ref struct` interpolation uses declared-only `StructSymbol.TryGetMethod` as the safety gate, then reuses `SynthesizeInstanceCall` → `BindAccessorCall`, the normal explicit-call resolution path.
- Failed synthesized calls discard their speculative diagnostics with the existing `DiagnosticBag.TruncateTo` pattern, allowing GS0519 to report the unsupported interpolation at a sane source span.
- Imported by-ref-like lookup uses a new type-preserving `GetMethodSafe(type, name, Type[])` overload. `Type.EmptyTypes` keeps parameterless `ToString` selection while retaining the constructed-generic/`TypeBuilder` hardening.
- GS0519 says the type has no usable parameterless `ToString`.
- Tests cover both source declaration forms, no declared method, an arguments-only method, imported overloaded methods, three emitted drivers, emitted-process execution, and formatting controls.

## Safety boundary

| Case | Merge-base | Fixed head |
|---|---|---|
| Receiver-clause declared `ToString`; `${token}` | rc 1, GS0519 | compile/run rc 0, `Receiver#42` |
| ADR-0079 in-body declared `ToString`; `${token}` | rc 1, GS0519 | compile/run rc 0, `InBody#43` |
| No declared `ToString`; `${token}` | rc 1, GS0519 | rc 1, GS0519 |
| Explicit inherited `${token.ToString()}` (#3240) | compile rc 0; run rc 134, `InvalidProgramException` | unchanged; out of scope |
| Non-ref `struct Plain` control | compile/run rc 0 | unchanged |

`TryGetMethodIncludingInherited` is never used. Calling an inherited virtual method would require boxing the by-ref-like receiver and route into #3240's invalid-program hole.

## Review blocker: overloaded imported `ToString`

Runtime was the oracle: `gsc /out:` followed by `dotnet exec`, with rc and stdout asserted.

| Product | Compile | Run | stdout |
|---|---:|---:|---|
| Base `0232ecb5e` | 0 | 0 | `interp=ExtTok!` |
| Reviewed head before fix (`15aab01db`, patch-equivalent to `7d795c322`) | 1, GS9998 `AmbiguousMatchException` | — | — |
| Fixed head | 0 | 0 | `interp=ExtTok!` |

Bare `gsc` and `gsi` also return rc 0 and print `interp=ExtTok!`. The regression fixture is a referenced ref struct with `ToString()` and `ToString(string)`; its test emits the reference assembly, compiles the G# executable, loads its types, executes it with bounded concurrent pipe reads, and asserts runtime output.

Decision: overload, not revert. Reverting to raw `GetMethod` would restore argument disambiguation but lose the `NotSupportedException` protection. The shared overload preserves both properties without a special case in the lowerer.

## Speculative diagnostic fix

For a source ref struct whose only declared `ToString` takes one argument:

- Before review fix: rc 1, GS0144 on the compiler-synthesized call, span `(9,33,9,41)`.
- Fixed: rc 1, GS0519, span `(9,23,9,40)` on all three drivers.

The fixed span ends with the interpolation statement and matches the pre-PR diagnostic shape. Test name was also changed from the fallback/inherited wording to `UserRefStructInterpolation_WithoutDeclaredToStringReportsGS0519AcrossFileDrivers`.

## RED proof and review mutants

New tests on the unfixed reviewed product: **4 failed / 24 passed**. Fixed product: **28/28**.

| Mutant | Core md5 | Result |
|---|---|---|
| Parameter-preserving lookup → name-only lookup | `96d45c408222310495c1f32ce3286f57` | RED: 1 failed / 27 passed |
| Remove speculative diagnostic truncation | `91ddbe83f013af23e90063410dc68e16` | RED: 3 failed / 25 passed |

Both mutants had non-empty `git diff HEAD`, moved `GSharp.Core.dll`, and restored exactly to plain-build md5 `724239c0425c8c6e4f763bebf53150f9` with 28/28 passing.

The original product mutants remain as previously reported: declared gate, GS0519 lowerer condition, and diagnostic wording are killed; raw-lookup hardening remains unreached by the existing constructed-generic/`TypeBuilder` restrictions.

## Validation

Fetched `origin/main` was `0232ecb5eb7c1f6878556599a6faf557487e4232`. Head `93c5eb769d33aba65d002a7d7cec519ebfd7a267` is rebased onto it.

Nine test projects:

- Core: 7,178 passed, 1 skipped
- Compiler: 4,267 passed
- Interpreter.Tests: 1,353 passed, 3 skipped
- LanguageServer: 468 passed
- SDK: 59 passed
- Extensions: 104 passed
- Cs2Gs: 1,907 passed
- GeneratorHost: 31 passed
- InternalAnalyzers: 9 passed
- **Total: 15,376 passed, 4 skipped, 0 failed across 9/9 projects**

Additional validation:

- Targeted issue tests: 28/28.
- `ClrTypeUtilitiesTests`: 18/18.
- `FullyQualifiedName~IlVerifier`: 11/11; output contained a matched test count and no missing-module/crash text.
- Final CI-flavoured serial solution build: rc 0, 0 warnings, 0 errors.
- Final `GSharp.Core.dll`: 2,488,832 bytes, md5 `d4448d72a5a3f881a888809a83c41ea0`, identical across Core, Compiler, Repl, Core.Tests, Compiler.Tests, and Interpreter.Tests outputs.

Not run locally: e2e tests, per repository standing rule.


DavidObando:
# Rubber-duck review — PR #3250 (issue #3220)

**Verdict: NOT MERGEABLE — 1 BLOCKER, 1 SHOULD-FIX, 1 NIT.**

The core fix is correct, minimal, and does exactly what the PR body claims. It is blocked by an
unrelated hunk in the same PR — the one the author already flagged as undemonstrated — which breaks
compilation of a program that builds on `main`.

Anchors: `origin/main` = `3e6dd50a7390366120547f9b25e12063f2095cfd`, fetched, re-fetched before posting.
Head = `cf98b2e85360fe8da28d5307d96b82fa78a90cfb`. The head's **own merge-base is `3e6dd50a7`**, i.e. the
branch was a fast-forward on `main` at review time; no merge tree was needed.

**`main` moved during this review** — it is `d172d0d14424df9caaa06de65a2a3f52619a1cc4` as I post, so
`3e6dd50a7` is now the head's historical merge-base rather than the tip. The PR head is unchanged at
`cf98b2e8` and still reports `MERGEABLE`. Every finding below is a differential measurement between
`3e6dd50a7` and `cf98b2e8`, and the BLOCKER is a property of the changed line's own reflection semantics,
so it does not depend on what landed on `main` in the interim. All work in dedicated worktrees
with their own `out/` root (`build/gsharp.build.props:67`, `OutRoot` derives from `MSBuildThisFileDirectory`).

---

## The question this review existed to answer: which lookup?

**`src/Core/CodeAnalysis/Binding/ExpressionBinder.Literals.cs:466` calls `structType.TryGetMethod("ToString", out _)`
— the declared-only lookup at `StructSymbol.cs:1054`.** `TryGetMethodIncludingInherited` (`StructSymbol.cs:1036`,
which walks `BaseClass`) is **not** used anywhere in the diff. The PR did not take the bait, and the
declared-versus-inherited boundary is intact.

### But the trap is inert, and that is worth knowing

I mutated the gate to `TryGetMethodIncludingInherited` (build clean, `GSharp.Core.dll` md5 moved
`d6c79967…` → `1e2d6c8c…`). **F1 stayed RED.** The mutant changed nothing.

The reason is structural, and I measured it rather than inferred it:

```
ref struct Derived1 : Base1 { ... }
GS0382: Struct 'Derived1' cannot declare base type 'Base1'; a struct may only implement interfaces.
```

`DeclarationBinder.Structs.cs:664` rejects any struct base type, so a `StructSymbol` with
`IsRefStruct: true` always has `BaseClass == null`, and the two lookups are **provably identical** on this
path. Choosing `IncludingInherited` would *not* have routed F1 into F2's hole.

Two consequences, and I want to be plain about both:

1. **That mutant is a NO-RESULT, not a survival.** I am not scoring it as a coverage gap. No test can
   distinguish the two lookups because no G# program can.
2. The PR's choice is still the right one — it is precise and self-documenting — but it is not
   load-bearing, and no test can ever make it so.

---

## BLOCKER — `src/Core/CodeAnalysis/Lowering/InterpolatedStringHandlerLowerer.cs:698` breaks a program that compiles on `main`

```diff
-        => type?.ClrType?.GetMethod("ToString", System.Type.EmptyTypes);
+        => type?.ClrType?.GetMethodSafe("ToString");
```

`ClrTypeUtilities.cs:45-65`: `GetMethodSafe` calls `type.GetMethod(name)` with **no type array**, and catches
only `NotSupportedException`. The replaced call constrained arity to zero. The new one does not, so it can
select a **non-nullary** `ToString`, and it throws **`AmbiguousMatchException`** whenever more than one
`ToString` exists — an exception `GetMethodSafe` does not catch.

The reflection semantics, measured directly:

```
typeof(double).GetMethod("ToString")              -> THREW AmbiguousMatchException
typeof(double).GetMethod("ToString", EmptyTypes)  -> System.String ToString()
```

**This is reachable.** A C# library with a by-ref-like type carrying `ToString` overloads, referenced with `/r:`:

```csharp
public ref struct ExtTok {
    public int Value;
    public override string ToString() => "ExtTok!";
    public string ToString(string fmt) => "fmt:" + fmt;
}
```

```gs
import System
import ExtLib
func Main() {
    var t = Fac.Make()
    Console.WriteLine("interp=${t}")
}
```

| tree | compile rc | run rc | stdout |
|---|---:|---:|---|
| **base `3e6dd50a7`** | **0** | **0** | `interp=ExtTok!` |
| **head `cf98b2e8`** | **1** | — | `error GS9998: AmbiguousMatchException: Ambiguous match found for 'ExtLib.ExtTok System.String ToString()'` |
| head, **only line 698 reverted** | **0** | **0** | `interp=ExtTok!` |

Control, single `ToString` (`SoloTok`): compile rc 0 / run rc 0 / `interp=SoloTok!` on **both** trees. The
discriminator is exactly the overload set — nothing else.

Attribution is direct: a third worktree at the same head with **only** that one line reverted compiles and
runs the program, **and the PR's actual fix still works** (`interp=Tok#42`).

Why this is a BLOCKER and not a nit:

- It is a **regression**: a program that compiles and runs on `main` fails to compile on this head.
- The failure is **`GS9998`**, an internal-error diagnostic wrapping a raw .NET exception — not a designed refusal.
- The hunk is **undemonstrated by the PR's own account**: *"No specimen distinguished the two implementations,
  and the raw-lookup mutant survives."* I found the specimen. It distinguishes them in the wrong direction.

**Scope, stated rather than implied:** I scanned `System.Private.CoreLib`, `System.Memory`, `System.Text.Json`
and `System.Runtime` for by-ref-like types with more than one public `ToString` and found **zero**. So this is
not reachable through the BCL today; it is reachable through any referenced third-party or user-authored C#
`ref struct`. The count being zero is a statement about the BCL, not about the language.

**Recommended fix (ponytail): delete the hunk.** It fixes nothing measurable, the PR concedes as much, and
`F2` works without it. If the TypeBuilder hardening is genuinely wanted later, it needs a `GetMethodSafe`
overload that preserves the arity constraint — and a specimen.

---

## SHOULD-FIX — `ExpressionBinder.Literals.cs:468-476` leaks a speculative bind's diagnostics

The new code calls `BindAccessorCall` on a **synthesized** `ToString()` and discards the result when it does
not type-check. But `BindAccessorCall` reports diagnostics as a side effect
(e.g. `ExpressionBinder.Calls.Invocation.cs:2171`), and those escape to the user.

Specimen — a `ref struct` declaring only a **non-nullary** `ToString`:

```gs
ref struct Ovl { var value int32 }
func (t Ovl) ToString(fmt string) string -> String.Format("Fmt{0}#{1}", fmt, t.value)
...  Console.WriteLine("interp=${b}")
```

| tree | diagnostic |
|---|---|
| base | `(6,23,6,36): error GS0519: … has no available CLR ToString method` |
| head | `(6,33,8,2): error GS0144: Function 'ToString' requires 1 arguments but was given 0.` |

The user never wrote a `ToString()` call — the compiler synthesized it, then blamed the user for its arity.
The span degrades too: `(6,33,8,2)` runs past the statement into the closing brace, where base is tight at
`(6,23,6,36)`. And the documented diagnostic is **lost** for the one case its new wording describes best:
this PR reworded GS0519 to *"no usable parameterless ToString method"*, which is precisely what `Ovl` is,
yet `Ovl` no longer produces GS0519.

The house already solves this, twice, with a named pattern:

- `ExpressionBinder.Access.Accessor.cs:1092-1098` (#942) — *"Snapshot the diagnostic bag and roll it back on
  failure so the probe is side-effect-free for ANY chain shape."*
- `OverloadResolver.Constructors.cs:262` (#1629) — `Diagnostics.TruncateTo(inferenceDiagMark)`.

**Proven fix — two lines, applied and measured:**

```csharp
var diagMark = Diagnostics.Count;
var toString = BindAccessorCall(...);
if (toString is BoundUserInstanceCallExpression { Type: var returnType } && returnType == TypeSymbol.String)
{
    bound = toString;
}
else
{
    Diagnostics.TruncateTo(diagMark);
}
```

Measured with it applied: `Ovl` → `(6,23,6,36): error GS0519: … has no usable parameterless ToString method`
(correct message, correct span); `Ovl2`, `F1`, `F2`, `NonStr` all unchanged; **Interpreter.Tests 1348 passed /
0 failed / 3 skipped — identical to the clean baseline.**

---

## NIT — ADR-0154: the test name claims a boundary the language cannot express

The PR body states the tests cover *"inherited-only refusal"*. Measured above, a struct cannot have a base
type at all (`GS0382`), so there is no inherited-only case. The test pins *"no `ToString` declared anywhere"*,
which is the correct behaviour — but the name asserts a discrimination that no G# program can exhibit.

---

## F1 / F2 matrix — compile-rc, run-rc, stdout, all three drivers

`gsc /out:` + `dotnet p.dll` is the oracle; `rc` captured before any pipe; `gsc` writes diagnostics to stdout.

| specimen | shape | base | head |
|---|---|---|---|
| **F2_implicit** | declared `ToString`, `${b}` | compile **rc 1**, `GS0519` | compile **rc 0** / run **rc 0** / `interp=Tok#42` |
| **F2_explicit** | declared `ToString`, `${b.ToString()}` | rc 0 / rc 0 / `interp=Tok#42` | unchanged |
| **F1_implicit** | **no** declared `ToString`, `${b}` | rc 1, `GS0519` | **rc 1, `GS0519` — stays RED** |
| **F1_explicit** | no declared `ToString`, `${b.ToString()}` | rc 0 / **run rc 134** | **unchanged** (#3240, correctly untouched) |
| **INBODY** | ADR-0079 in-body `ToString` | rc 1, `GS0519` | rc 0 / rc 0 / `interp=IB#42` |
| **GEN2** | generic `ref struct GBox[T]`, in-body | rc 1, `GS0519` | rc 0 / rc 0 / `interp=GBox!` |
| **OVL** | only non-nullary `ToString` | rc 1, `GS0519` | rc 1, **`GS0144` (see SHOULD-FIX)** |
| **OVL2** | non-nullary declared first **and** nullary | rc 1, `GS0519` | rc 0 / rc 0 / **`interp=Nullary#42`** |
| **NONSTR** | `ToString()` returning `int32` | rc 1, `GS0519` | rc 1, `GS0519` |
| **PLAIN** | non-ref `struct` | rc 0 / rc 0 / `Default.Plain` | **identical** |
| **EXT_solo** | imported ref struct, one `ToString` | rc 0 / rc 0 / `interp=SoloTok!` | identical |
| **EXT_ovl** | imported ref struct, `ToString` overloads | rc 0 / rc 0 / `interp=ExtTok!` | **rc 1, `GS9998` — BLOCKER** |

Every row was run on **all three drivers** (bare `gsc`, `gsc /out:` + `dotnet`, `gsi`) and all three agreed in
every cell. `${b}` and `${b.ToString()}` were both exercised.

**`OVL2` is the strongest evidence the design is right**: the gate is origin-only, and real overload resolution
via `BindAccessorCall` picks the nullary even though `TryGetMethod` returns the non-nullary first. The PR body's
claim that `TryGetMethod` *"establishes only member origin"* is true, and I confirmed it rather than assuming it.

---

## Mutation — PRODUCT only, per hunk, assert applied, exact restore

Baseline `GSharp.Core.dll` md5 `d6c799671f551ea666ed9da4208696a6`; every mutant showed a non-empty
`git diff HEAD` touching a real statement, built clean, moved the md5, and restored to the exact baseline md5.

| mutant | md5 | result |
|---|---|---|
| gate → `TryGetMethodIncludingInherited` | `1e2d6c8c…` | **NO-RESULT (inert)** — F1 unchanged; unreachable per `GS0382`. Not scored. |
| `returnType == TypeSymbol.String` → `!=` | `452fcfec…` | **KILLED** — 3 failed / 1345 passed |
| candidate diagnostic-rollback fix | — | 1348 / 0 / 3, identical to baseline |

**Two-sided.** The PR's test file applied alone to `3e6dd50a7` with the production hunk asserted **absent**
(`git diff HEAD --name-only | grep -c '^src/'` = 0; the binder hunk greps 0 times) → **6 failed / 1342 passed /
3 skipped**, matching the author's reported RED. Clean merged tree → 1348 / 0 / 3.

---

## Suites — denominator 9, source `dotnet sln GSharp.sln list`

| project | result |
|---|---|
| Core.Tests | 7178 passed / 1 skipped |
| Interpreter.Tests | 1348 passed / 3 skipped |
| Sdk.Tests | 59 passed |
| Compiler.Tests | 4268 passed |
| Cs2Gs.Tests | 1907 passed |

**9 of 9 projects, 15,372 tests, 0 failures, 4 skipped**, all unfiltered, every `RC` captured and `RC=0`. `Compiler.Tests` was run rather than skipped: the change is in the
binder and the lowerer, both on the emit path, so containment does not hold.

---

## Review-harness disclosures

- My first run of the imported-ref-struct specimens reported `run rc=134`, which looks exactly like the
  `InvalidProgramException` signature this issue is about. It was **my harness** — `ExtLib.dll` was not beside
  the emitted `p.dll`. Copying the reference in gave `rc=0`. A failing probe is not evidence until it fails
  for the reason you think, and this one very nearly got scored as a product finding.
- The `IncludingInherited` mutant is reported above as a **no-result**, not a survival. It built, moved the
  md5, and changed no behaviour, because the construct it depends on is rejected at `GS0382`. Reporting it as
  a coverage gap would have been wrong in the direction that flatters the reviewer.

Both worktrees end at drift 0 against their checkouts; the third attribution worktree was removed. The shared
checkout was never written to and never used as a source of evidence.

DavidObando:
# Re-review after rebase — PR #3250 head `7d795c322`

**Verdict unchanged: NOT MERGEABLE — 1 BLOCKER, 1 SHOULD-FIX, 1 NIT.**

I re-measured everything on the rebased head rather than carrying my earlier numbers across. All findings
reproduce, and the rebase introduced no new ones.

**Anchors, fetched myself:** `origin/main` = `0232ecb5eb7c1f6878556599a6faf557487e4232`;
head = `7d795c3222ea96500022f097f3b47da2f5bd8f49`; head's own merge-base = `e28d55bdfc83ac78dcbd0e9f581aaf7f0f0326df`
(so `main` has moved on past the branch point — every differential below is head vs **`e28d55bdf`**, freshly built).

**Patch-id equivalence — verified, not accepted.**

```
diff(3e6dd50a7 .. cf98b2e85) | git patch-id --stable  ->  84033ccd60fa7ec0421c9ea340acc6a2d3d40a02
diff(e28d55bdf .. 7d795c322) | git patch-id --stable  ->  84033ccd60fa7ec0421c9ea340acc6a2d3d40a02
```

Identical. The author's claim holds. I still re-ran every measurement.

**Merge state:** `git merge-tree --write-tree 0232ecb5e 7d795c322` emits a tree oid and **zero stage entries**
(`^[0-7]{6} [0-9a-f]{40} [123]` matches 0 — no stage 3, and no stage 1+2 modify/delete either). I read the
stages, not a `CONFLICT` count and not a `<<<<<<<` grep. `mergeable` re-queried: **MERGEABLE**.

---

## The property that mattered most — independently confirmed on `7d795c322`

**Which lookup:** `src/Core/CodeAnalysis/Binding/ExpressionBinder.Literals.cs:466`

```csharp
if (bound.Type is StructSymbol { IsRefStruct: true } structType
    && structType.TryGetMethod("ToString", out _))          // :466 — DECLARED-ONLY (StructSymbol.cs:1054)
```

`TryGetMethodIncludingInherited` (`StructSymbol.cs:1036`, the inherited-walking one) appears **0 times** in the
whole `src/` diff. This is the right side of the line, confirmed by my own read of the rebased tree.

**F1 stays RED** and **#3240 is genuinely untouched** — measured on both trees, compile *and* run:

| specimen | base `e28d55bdf` | head `7d795c322` |
|---|---|---|
| **F1** — `ref struct` no declared `ToString`, `${b}` | compile **rc 1**, `GS0519` | compile **rc 1**, `GS0519` ✅ |
| **#3240** — no declared `ToString`, `${b.ToString()}` | compile rc 0 → run **rc 134** `InvalidProgramException` | compile rc 0 → run **rc 134**, identical ✅ |
| **F2** — declared `ToString`, `${b}` | compile **rc 1**, `GS0519` | compile **rc 0** → run **rc 0** → `interp=Tok#42` ✅ |

My measurements **agree with the #3220 author's independent report** in every cell. #3240 moves in neither
direction, which is exactly the corroboration wanted: the fix does not launder a compile error into a
miscompiled program.

### The trap is inert — re-confirmed, and this time with a liveness control

I re-ran the substitution on the rebased head in its own worktree (baseline Core.dll md5
`a837dcb91e3f7e6b1065f9218fcb1b0d`):

| mutant on `7d795c322` | md5 | F1 | F2 | result |
|---|---|---|---|---|
| `TryGetMethod` → `TryGetMethodIncludingInherited` | `06826333…` | **rc 1 `GS0519`** | `interp=Tok#42` | **INERT — no-result** |
| `TryGetMethod` → `!TryGetMethod` (**liveness control**) | `802b24bf…` | rc 1 | **rc 1 `GS0519` — F2 BREAKS** | **live, observable** |

The control is the point. Before calling the first mutant inert I proved the line is on the path and visible to
my probes: negating the gate breaks F2 immediately. So the inertness is a real property of *that substitution*,
not an artifact of an unexercised line.

The cause is structural and measured, not inferred — `DeclarationBinder.Structs.cs:664`:

```
ref struct Derived1 : Base1 { … }
GS0382: Struct 'Derived1' cannot declare base type 'Base1'; a struct may only implement interfaces.
```

A struct's `BaseClass` is always null, so on this path the two lookups are indistinguishable. **The declared-only
call at `:466` is correct and I would not change it** — but no test can ever pin that choice, because no G# program
can express the difference. Worth knowing when judging the PR's coverage claims.

Restored to the exact baseline md5 `a837dcb9…`, drift 0.

---

## BLOCKER — `src/Core/CodeAnalysis/Lowering/InterpolatedStringHandlerLowerer.cs:698` still breaks a program that builds on the merge-base

Unchanged by the rebase:

```diff
-        => type?.ClrType?.GetMethod("ToString", System.Type.EmptyTypes);
+        => type?.ClrType?.GetMethodSafe("ToString");
```

`ClrTypeUtilities.cs:45-66`: `GetMethodSafe(type, name)` calls `type.GetMethod(name)` with **no type array** and
catches only `NotSupportedException`. The replaced call constrained arity to zero; this one does not, so it can
select a **non-nullary** `ToString` and it throws **`AmbiguousMatchException`** — which `GetMethodSafe` does not catch —
whenever a type declares more than one `ToString`.

Specimen: a C# library with a by-ref-like type carrying `ToString` overloads, referenced via `/r:`.

```csharp
public ref struct ExtTok {
    public int Value;
    public override string ToString() => "ExtTok!";
    public string ToString(string fmt) => "fmt:" + fmt;
}
```

Compile → `dotnet exec` → rc **and** stdout, `rc` captured before any pipe:

| tree | compile rc | run rc | stdout |
|---|---:|---:|---|
| base `e28d55bdf` | **0** | **0** | `interp=ExtTok!` |
| head `7d795c322` | **1** | — | `error GS9998: AmbiguousMatchException: Ambiguous match found for 'ExtLib.ExtTok System.String ToString()'` |
| head with **only line 698 reverted** | **0** | **0** | `interp=ExtTok!` |

Control — same library, type with a **single** `ToString` (`SoloTok`): compile rc 0 / run rc 0 / `interp=SoloTok!`
on **both** trees. The discriminator is exactly the overload set, nothing else. The one-line revert isolates
attribution completely, **and the PR's own fix still works with that line reverted** (`interp=Tok#42`).

A program that compiles and runs on the merge-base fails to compile on this head, and the failure is `GS9998` —
an internal-error diagnostic wrapping a raw .NET exception, not a designed refusal.

### Why this is a helper-precondition violation, quantified

`GetMethodSafe` has exactly **one** overload — there is no arity-preserving variant. Across `src/` there are
**21 call sites**, and **20 of them pass `"Invoke"`**:

```
total GetMethodSafe call sites (excluding the helper file): 21
of which pass "Invoke":                                     20
non-Invoke call sites:
  src/Core/CodeAnalysis/Lowering/InterpolatedStringHandlerLowerer.cs:698  -> "ToString"
```

A delegate has exactly one `Invoke` by construction, so ambiguity is impossible for all 20 existing users. The
helper is safe today **because of its callers, not because of its implementation**. `:698` is the first and only
call site to pass an overloadable name, silently violating an unstated precondition.

**Recommended fix — preserve the arity constraint, do not merely delete the line.** Adding

```csharp
public static MethodInfo GetMethodSafe(this Type type, string name, Type[] types)
```

keeps the `NotSupportedException` / TypeBuilder hardening the hunk was presumably added for **and** restores the
zero-arity constraint. Reverting `:698` to `GetMethod("ToString", Type.EmptyTypes)` also fixes the BLOCKER but
gives up that hardening; I am flagging the *regression*, not arguing the line is dead. (The helper's TypeBuilder
fallback calls `GetGenericTypeDefinition().GetMethod(name)` with the same unconstrained shape — worth the same
treatment, though only `Invoke` reaches it today.)

**Scope, stated rather than implied:** I scanned `System.Private.CoreLib`, `System.Memory`, `System.Text.Json`
and `System.Runtime` for by-ref-like types with more than one public `ToString` and found **zero**. This is not
reachable through the BCL today; it is reachable through any referenced third-party or user-authored C# `ref struct`,
which is ordinary legal C#. The zero is a fact about the BCL, not about the language.

---

## SHOULD-FIX — `ExpressionBinder.Literals.cs:468-476` leaks a speculative bind's diagnostics (reproduces)

`BindAccessorCall` reports diagnostics as a side effect (`ExpressionBinder.Calls.Invocation.cs:2171`). The new
code binds a **synthesized** `ToString()` and discards the *result* when it does not type-check, but not the
*diagnostics*. Specimen — `ref struct` declaring only a non-nullary `ToString`:

| tree | diagnostic |
|---|---|
| base `e28d55bdf` | `(6,23,6,36): error GS0519: … has no available CLR ToString method` |
| head `7d795c322` | `(6,33,8,2): error GS0144: Function 'ToString' requires 1 arguments but was given 0.` |

The user never wrote a `ToString()` call — the compiler synthesized one and then blamed the user for its arity.
The span also degrades, running past the statement into the closing brace. And GS0519 is lost for precisely the
case its **new wording** describes best: this PR reworded it to *"no usable parameterless ToString method"*, which
is exactly what this type is.

House pattern, documented twice: `ExpressionBinder.Access.Accessor.cs:1092-1098` (#942 — *"Snapshot the diagnostic
bag and roll it back on failure so the probe is side-effect-free"*) and `OverloadResolver.Constructors.cs:262`
(#1629 — `Diagnostics.TruncateTo(mark)`).

Two-line fix, previously applied and measured: the specimen returns to `GS0519` at the tight span `(6,23,6,36)`,
all other specimens unchanged, Interpreter.Tests identical to baseline.

---

## NIT — ADR-0154: a test name claiming a boundary the language cannot express

The PR describes coverage of *"inherited-only refusal"*. Per `GS0382` above, a struct cannot have a base type, so
no inherited-only case exists; the test pins *"no `ToString` declared anywhere"*. Correct behaviour, overclaiming name.

---

## Full matrix on the rebased head — compile-rc / run-rc / stdout

| specimen | base `e28d55bdf` | head `7d795c322` |
|---|---|---|
| F1 `${b}` no ToString | rc 1 `GS0519` | rc 1 `GS0519` ✅ |
| #3240 `${b.ToString()}` no ToString | rc 0 → **run 134** | rc 0 → **run 134** ✅ |
| F2 `${b}` declared | rc 1 `GS0519` | rc 0 → 0 → `interp=Tok#42` ✅ |
| F2 `${b.ToString()}` declared | rc 0 → 0 → `interp=Tok#42` | identical |
| in-body `ToString` (ADR-0079) | rc 1 | rc 0 → 0 → `interp=IB#42` |
| generic `ref struct GBox[T]` | rc 1 | rc 0 → 0 → `interp=GBox!` |
| overload set incl. nullary | rc 1 | rc 0 → 0 → **`interp=Nullary#42`** |
| only non-nullary `ToString` | rc 1 `GS0519` | rc 1 **`GS0144`** ← SHOULD-FIX |
| `ToString()` returning `int32` | rc 1 `GS0519` | rc 1 `GS0519` |
| non-ref `struct` | rc 0 → 0 → `Default.Plain` | identical |
| imported ref struct, one `ToString` | rc 0 → 0 → `interp=SoloTok!` | identical |
| imported ref struct, `ToString` overloads | rc 0 → 0 → `interp=ExtTok!` | **rc 1 `GS9998`** ← BLOCKER |

Every row is compile → `dotnet exec` → rc **and** stdout. No row was scored on a clean compile alone — that is
precisely #3240's failure mode. **`ilverify` was not used as an oracle anywhere in this review.**

The overload-set row is the best evidence the design is sound: the gate establishes only member *origin*, and real
overload resolution picks the nullary even though the declared-only lookup returns the non-nullary first.

---

## Suites on `7d795c322` — denominator 9, source `dotnet sln GSharp.sln list`

| project | result |
|---|---|
| Core.Tests | 7178 passed / 1 skipped |
| Interpreter.Tests | 1349 passed / 3 skipped |
| Sdk.Tests | 59 passed |
| Compiler.Tests | 4267 passed |
| Cs2Gs.Tests | 1907 passed |

**9 of 9 projects, 15,372 passed, 0 failed, 4 skipped**, all unfiltered, every `RC` captured and `RC=0`. `Compiler.Tests` run rather than skipped — the change is in the binder and
the lowerer, both on the emit path.

---

## Harness disclosures

- My first run of the imported-ref-struct specimens reported **rc 134**, the same signature as the
  `InvalidProgramException` this issue concerns. It was **my harness** — `ExtLib.dll` was not beside the emitted
  `p.dll`. With the reference copied in, base gives rc 0 and `interp=ExtTok!`. A failing probe is not evidence
  until it fails for the reason you think, and this one came close to being scored as a product finding.
- The `IncludingInherited` mutant is reported as a **no-result**, not a survival, and I ran a liveness control
  before saying so.
- All work in dedicated worktrees with their own `out/` root; the shared checkout was never written to nor used
  as a source of evidence. All worktrees end at drift 0, mutation tree restored to its exact baseline md5.

DavidObando:
## Rubber-duck round 3 — head `93c5eb769`

**Verdict: MERGEABLE.** Both round-2 findings are dead, and both are now pinned by tests that I killed mutants against. 3 NITs and one merge-order note below; none blocks.

### Anchors (fetched, not taken on trust)

| | |
|---|---|
| `main` | `6b809619779b4403062a6a9540663202f504fb73` |
| PR head | `93c5eb769d33aba65d002a7d7cec519ebfd7a267` |
| head's own merge-base | `0232ecb5eb7c1f6878556599a6faf557487e4232` |
| commits | `15aab01db` (rebased fix) + `93c5eb769` (regression repair) |
| `merge-tree main..head` | tree `986fdd370`, **0 stage entries** |

Every measurement below is head vs **its own merge-base**, both built from scratch in dedicated worktrees with their own `out/` roots (`build/gsharp.build.props:67`). Freshness witness `GSharp.Core.dll`: base `29e6518091c8e9c7b6672fcc3efb0f53`, head `dfb1a6ae77d3e72b9d7a691cbe83162d`, both written 07:42. Every row is compile → `dotnet exec` → rc **and** stdout, rc captured before any pipe. `ilverify` was not used as an oracle anywhere.

---

## 1. The BLOCKER is dead

Round 2 proved a **build regression**: a C# library `ref struct` carrying *two* `ToString` overloads compiled and ran on the merge base, and failed with `GS9998: AmbiguousMatchException` on the PR head.

Same specimen, same library, re-measured on `93c5eb769`:

| specimen | tree | compile rc | run rc | stdout |
|---|---|---:|---:|---|
| `EXT_ovl` (`ToString()` + `ToString(string)`) | base `0232ecb5e` | 0 | 0 | `interp=ExtTok!` |
| `EXT_ovl` | **head `93c5eb769`** | **0** | **0** | **`interp=ExtTok!`** |
| `EXT_solo` (single `ToString`) — control | base | 0 | 0 | `interp=SoloTok!` |
| `EXT_solo` — control | head | 0 | 0 | `interp=SoloTok!` |

**The repair is the one that was recommended, not a special case.** `ClrTypeUtilities.cs:77-98` adds `GetMethodSafe(this Type, string, Type[])` as a faithful mirror of the existing 1-arg form at `:45-66`: same null guard, same `catch (NotSupportedException)`, same `IsConstructedGenericWithTypeBuilderArgument` gate, same `TypeBuilder.GetMethod` re-resolution. `InterpolatedStringHandlerLowerer.cs:698` now passes `System.Type.EmptyTypes`, restoring exactly the arity constraint the original `GetMethod("ToString", Type.EmptyTypes)` applied — with the TypeBuilder hardening kept.

**Positive control — the probe is still live.** I reverted *only* `:698` back to the 1-arg form in a mutation tree (`git diff HEAD` non-empty on a real statement; `GSharp.Core.dll` md5 `6bcd2598…` → `b8ef9a58…`):

```
error GS9998: AmbiguousMatchException: Ambiguous match found for 'ExtLib.ExtTok System.String ToString()'.
```

The round-2 failure reproduces on demand. The fix at `:698` is what removes it, and `EXT_solo` stays rc0/rc0 under the mutant, so the discriminator is still exactly "more than one `ToString` overload".

**And the PR's own test catches it.** `ImportedRefStructInterpolation_WithOverloadedToString_CompilesAndRuns` builds the two-overload by-ref-like type with `PersistedAssemblyBuilder`, compiles G# against it, **runs the emitted exe** and asserts stdout `interp=ExtTok!\n`. Against the mutant: **1 failed / 27 passed**. This is oracle-strength under ADR-0154 — it asserts the running program's output, not a compile rc.

### Does the repair disarm the trap for the *next* caller?

Recount on the new head:

```
21 call sites of GetMethodSafe; 20 pass "Invoke"; 1 non-Invoke.
```

The split is unchanged, and the single non-`Invoke` caller is `InterpolatedStringHandlerLowerer.cs:698` — now the only consumer of the 3-arg overload, and it passes `Type.EmptyTypes`. No new non-`Invoke` callers were introduced.

Judgement: **the trap is disarmed as far as it reasonably can be.** The helper now mirrors BCL's own shape — `GetMethod(name)` is ambiguity-prone, `GetMethod(name, types)` is not — so a future caller who knows their arity has a correct tool sitting beside the unsafe one. What round 2 objected to was that *only* the unsafe shape existed, which forced this PR into it. That is no longer true. See NIT-1 for the one residual.

---

## 2. The SHOULD-FIX is dead — verified by slicing, not arithmetic

`ExpressionBinder.Literals.cs:465-482` now snapshots and rolls back: `var diagMark = Diagnostics.Count;` before the speculative `BindAccessorCall`, and `Diagnostics.TruncateTo(diagMark)` in the `else`. That is the house pattern (`ExpressionBinder.Access.Accessor.cs:1092-1098` #942, `OverloadResolver.Constructors.cs:262` #1629).

I built a harness against each tree's own `GSharp.Core.dll` that emits and prints `d.Location.Text.ToString(d.Location.Span)` for every diagnostic. Specimen: `ref struct Ovl` declaring only `ToString(fmt string)`.

```
base 0232ecb5e:  ID=GS0519 COORD=(6,23,6,36) SPAN=[202,215) LEN=13
                 COVERED-TEXT=<<<"interp=${b}">>>
head 93c5eb769:  ID=GS0519 COORD=(6,23,6,36) SPAN=[202,215) LEN=13
                 COVERED-TEXT=<<<"interp=${b}">>>
```

Identical span, identical covered text, only the message wording differs. The leaked `GS0144: Function 'ToString' requires 1 arguments but was given 0` is gone, and the span no longer runs past the end of the statement. Note the covered text is the **whole interpolated-string literal**, not the hole — but that is exactly what the base produced, so this PR restores the status quo rather than changing it. (See §5: PR #3260 is the one that narrows this.)

**Mutant C** — replace `Diagnostics.TruncateTo(diagMark);` with `_ = diagMark;` (md5 `6bcd2598…` → `fa995294…`): `UserRefStructInterpolation_WithoutParameterlessToStringReportsGS0519AcrossFileDrivers` fails **3/3 drivers**. Pinned.

---

## 3. Mutation ledger — 3/3 RED, exact restore

Mutation tree at `93c5eb769`, baseline `GSharp.Core.dll` md5 `6bcd25980ace84b6da3a218b13d14ba6`. Each mutant asserted by non-empty `git diff HEAD` on a statement **and** within-tree md5 movement, with build rc read before scoring.

| # | hunk | mutation | md5 | result |
|---|---|---|---|---|
| A | `InterpolatedStringHandlerLowerer.cs:698` | `GetMethodSafe("ToString", EmptyTypes)` → `GetMethodSafe("ToString")` | `b8ef9a58…` | **RED** — `GS9998` returns; PR test 1F/27P |
| B | `ClrTypeUtilities.cs:86` | `type.GetMethod(name, types)` → `type.GetMethod(name)` | `34c3f35a…` | **RED** — PR test 1F/27P |
| C | `ExpressionBinder.Literals.cs:480` | drop `Diagnostics.TruncateTo(diagMark)` | `fa995294…` | **RED** — 3F/25P, all drivers |
| — | restore | — | **`6bcd2598…`** | exact round-trip |

A and B are separate hunks with the same observable, which is the right shape: the call site and the helper body are each load-bearing.

---

## 4. Behaviour matrix, and the GS0519 change list

Both trees, all three drivers (emit-to-file + run, bare `gsc`, `gsi`), agreeing in every cell.

**GS0519 → now compiles and runs** (this is the merge-relevant list):

| specimen | base | head |
|---|---|---|
| receiver-clause nullary `ToString` | GS0519 | rc0/rc0 `interp=Tok#42` |
| in-body nullary `ToString` | GS0519 | rc0/rc0 `interp=IB#42` |
| nullary **and** parameterised `ToString` | GS0519 | rc0/rc0 `interp=Nullary#42` |
| **generic** `GBox[int32]` with in-body `ToString` | GS0519 | rc0/rc0 `interp=GBox!` |

**Still GS0519 — wording only:**

| specimen | base | head |
|---|---|---|
| **no `ToString` at all (F1)** | **rc1 GS0519** | **rc1 GS0519** |
| only parameterised `ToString` | rc1 GS0519 | rc1 GS0519, span byte-identical |
| nullary `ToString` returning non-`string` | rc1 GS0519 | rc1 GS0519 |

**Unchanged, no GS0519 involved:** imported ref structs with one or two `ToString` overloads (rc0/rc0 both sides); non-ref struct; `Span`/`ReadOnlySpan` (covered by the pre-existing #3114 assertions, still green).

**F1 stays RED** — the trap I was sent to watch for is not sprung. `ExpressionBinder.Literals.cs:466` calls the **declared-only** `StructSymbol.TryGetMethod` (`:1054`), not `TryGetMethodIncludingInherited` (`:1036`); `IncludingInherited` appears zero times in the `src/` diff.

**#3240 is genuinely unaffected.** `ref struct` with no declared `ToString`, interpolated as `${b.ToString()}`: compile rc **0**, run rc **134** `InvalidProgramException`, empty stdout — **identical on base and head**. The fix moves it in neither direction.

Two extra properties worth recording, both clean:

- **Single evaluation.** `${MakeTok()}` where `MakeTok` prints and increments a counter: `<MakeTok called>` appears **once**, `counter=1`. The speculative re-bind does not duplicate the receiver's side effects.
- **Alignment and format.** `${b,12}` → `[      Tok#42]`, `${b,-12}` → `[Tok#42      ]`, `${b,12:X}` → `[      Tok#42]`. Alignment is honoured; `:X` is dropped, which matches C#'s behaviour for a non-`IFormattable` type. All three drivers identical.

---

## 5. PR #3260 interaction — reporting, not acting

Fresh `merge-tree --write-tree` between the two current heads (#3260 head `14f3d7ba8`), reading **stage entries**, not `CONFLICT` counts:

```
3250 -> 3260 : rc=0  tree 1b858982a  stage entries: 0
3260 -> 3250 : rc=0  tree 1b858982a  stage entries: 0
```

Textually clean in both orders, same result tree. #3260's only hunk in `ExpressionBinder.Literals.cs` is at `:1378` (`BoundStructLiteralExpression(null, …)` → `(syntax, …)`), ~900 lines from this PR's insertion at `:466`.

**But there is a semantic interaction the second merger must handle.** #3260 exists to *narrow GS0519 spans to the hole expression* — its own tests assert `(6,32,6,48)` covering `Token{value: 42}` and `(11,32,11,39)` covering `o.inner`. This PR adds a hard-coded whole-literal span assertion:

```
test/Interpreter.Tests/Issue3114ReadOnlySpanInterpreterTests.cs:462
    Assert.StartsWith($"{sourcePath}(9,23,9,40): error GS0519:", diagnostic, ...);
```

`(9,23,9,40)` is 17 characters — the whole `"interp=${token}"` literal, which is what I sliced above. Under #3260 that span becomes the hole. **Whichever of these two lands second must update `:462`.** No action needed here; flagging so it is not a surprise.

Also worth stating plainly for whoever merges second: this PR makes a source-declared `ref struct` **that declares a parameterless `ToString` returning `string`** stop reaching GS0519 at all — see the change list in §4. #3260's span work only affects the specimens that still reach GS0519.

---

## NITs (none blocking)

**NIT-1 — `ClrTypeUtilities.cs:45-66`, the 1-arg `GetMethodSafe`, still has an undocumented precondition.** It swallows `NotSupportedException` but not `AmbiguousMatchException`, so any future non-`Invoke` caller reaching for it gets `GS9998` (an internal-error diagnostic) rather than a real error. That precondition — "only safe for names that are unique by construction" — is what this PR tripped over, and it is still nowhere in the doc comment. One sentence on the summary would close it.

**NIT-2 — `ClrTypeUtilities.cs:95-96`, the 3-arg overload's TypeBuilder fallback forwards *constructed* parameter types to the *open* generic definition.** For the only current caller (`Type.EmptyTypes`) that is correct and I measured the generic case working end to end (`GBox[int32]` → `interp=GBox!`). For a hypothetical caller passing real parameter types over a TypeBuilder-backed generic, the open definition's parameters are generic parameters, so the lookup would silently return `null`. Unreachable today; noting it because it is the same "safe only because of its callers" shape.

**NIT-3 — file placement.** 245 lines of #3220 tests live in `test/Interpreter.Tests/Issue3114ReadOnlySpanInterpreterTests.cs`, a file named for a different issue. The tests themselves are good; the name will mislead the next person grepping for #3220.

---

## Suites — denominator 9, source `dotnet sln GSharp.sln list` (per `build/generate-ci-test-matrix.py:26`)

| project | passed | failed | skipped |
|---|---:|---:|---:|
| Core.Tests | 7178 | 0 | 1 |
| Cs2Gs.Tests | 1907 | 0 | 0 |
| Interpreter.Tests | 1353 | 0 | 3 |
| LanguageServer.Tests | 468 | 0 | 0 |
| Sdk.Tests | 59 | 0 | 0 |
| GeneratorHost.Tests | 31 | 0 | 0 |
| Compiler.Tests | 4267 | 0 | 0 |
| **total** | **15,376** | **0** | **4** |

---

Nothing here blocks. The round-2 regression is gone, it went away by the arity-preserving route rather than a special case, and both repaired hunks are now killed by the PR's own tests rather than merely accompanied by them.

Issue #3258, opened 2026-08-05
[diagnostic] GS0519 is anchored to the wrong span; BoundTreeRewriter returns null and the behaviour is now unpinned


DavidObando:
## Summary

Salvaged from PR #3104, which was closed as superseded by #3244 (Phase 3c deleted every `Evaluator*` file that PR touched). This half is **unrelated to the evaluator, still applies to `main`, and applies conflict-free** — it should not be lost with the branch.

`BoundTreeRewriter.cs:623` returns `null` on `main`, and as a result the **GS0519 diagnostic is anchored to the wrong span**. The fix is a **+15/−14** change.

Measured span change:

| | span |
|---|---|
| before | `(10,23,10,45)` |
| after | `(10,32,10,43)` |

The corrected span points at the offending construct rather than at a wider enclosing range.

## Why it needs re-landing rather than just re-applying

The original span was pinned **only by a test that #3244 deleted**, so on `main` this behaviour is now **completely unpinned**. Re-applying the product hunk without a new pin would land a correct fix with no protection against it silently regressing.

## Requirements

1. Apply the `BoundTreeRewriter.cs:623` fix (+15/−14). It was verified conflict-free against `5e2fee875`.
2. **Add a column-asserting GS0519 pin.** Asserting only the diagnostic id is not sufficient — the entire defect is the span, so a test that checks the id would pass both before and after the fix and prove nothing. Assert **start line, start column, end line, end column**.
3. Prove the pin can fail: apply it alone to the unfixed tree, rebuild, and show it RED with the **actual wrong span** `(10,23,10,45)` observed — not merely a red line. Then restore and confirm `git diff HEAD` = 0.

## Context worth knowing

Under the reverting mutant, `Core.Tests` ran **7163/7163 green** — i.e. the entire suite is blind to this span. That is the measurement that makes the new pin necessary rather than optional.

The tree-walking evaluator was deleted in #3244, so there is no interp-vs-emit oracle. The oracle is the diagnostic's id **and span**.

`gsc` writes diagnostics to **stdout** and only `Failed.` to stderr — **capture `rc` before any pipe**.

PR #3260, opened 2026-08-05
Anchor GS0519 to interpolation hole spans

DavidObando:
## Summary

Anchor `GS0519` to the interpolation hole expression instead of the enclosing string literal.

The fix records `segment.Expression` once in the sole interpolation-hole binding loop, preserves it through `BoundInterpolatedStringPart.WithValue`, and uses it when the lowerer reports `GS0519`. This is complete by construction: every hole passes through `FromHole`, independent of whether the surviving bound value node retained syntax.

The production change remains three files, +14/-7. The earlier sampled bound-node syntax changes are absent because this bind-time design makes them unnecessary for `GS0519`, not because a surviving mutant was treated as evidence of dead code.

## PR #3250 interaction

Rebased onto `main` `65578d41b2cc887596c1cda95674a4444390f685`, including PR #3250. The conflict resolution preserves #3250's ref-struct `ToString` binding and passes `segment.Expression` to `FromHole` after that binding.

#3250 moved valid nullary-`ToString` specimens out of the `GS0519` population. None of this PR's 14 specimens left the population: every resulting `Token` lacks a usable nullary string-returning `ToString`. In particular, the generic row is a generic factory returning `Token`, not the now-supported generic `GBox[int32]` case.

## Measured spans on this head

End coordinates are exclusive.

| shape | `main` | this head | covered text |
|---|---|---|---|
| free call | `(10,23,10,45)` | `(10,32,10,43)` | `MakeToken()` |
| generic call | `(10,23,10,52)` | `(10,32,10,50)` | `MakeToken[int32]()` |
| argumented call | `(10,23,10,50)` | `(10,32,10,48)` | `MakeToken(1 + 2)` |
| nested-interpolation argument | `(11,23,11,53)` | `(11,32,11,51)` | `MakeToken("n=${n}")` |
| awaited argument | `(14,23,14,57)` | `(14,32,14,55)` | `MakeToken(await Seed())` |
| lambda body | `(11,27,11,49)` | `(11,36,11,47)` | `MakeToken()` |
| extension call | `(11,23,11,42)` | `(11,32,11,40)` | `n.Wrap()` |
| instance call | `(13,23,13,42)` | `(13,32,13,40)` | `b.Make()` |
| field access | `(11,23,11,41)` | `(11,32,11,39)` | `o.inner` |
| struct literal | `(6,23,6,50)` | `(6,32,6,48)` | `Token{value: 42}` |
| lambda-valued local invocation | `(7,23,7,37)` | `(7,32,7,35)` | `f()` |
| function-parameter invocation | `(6,23,6,37)` | `(6,32,6,35)` | `f()` |
| method-value local invocation | `(11,23,11,37)` | `(11,32,11,35)` | `g()` |
| local value | `(7,23,7,39)` | `(7,32,7,37)` | `token` |

All 14 still produce `GS0519` on current `main`; `main` anchors every one to the whole literal. Result: the requested original ten are **10 fixed, 0 regressed**, and the expanded matrix is **14 fixed, 0 left the population**.

## Pins and RED proof

The original five test bodies and expected spans remain unchanged. The added matrix pins all remaining shapes using a single diagnostic, `GS0519`, all four coordinates, and `Location.Text.ToString(Location.Span)`.

With only the three production files restored to current `main`:

- `InterpolatedStringTests`: 36 passed, 14 failed, rc 1; every failure printed the whole-literal span and text.
- `GSharp.Core.dll`: fixed `321bbfbad9930103c0ef91fd10e07c98`; main product `21bfa7cfbc58ec0587a0b41ab51063c7`; restored `321bbfbad9930103c0ef91fd10e07c98`.
- no `No test matches`; all 50 tests ran.

PR #3250's merged interpreter pin is also updated and independently slices the diagnostic location:

- old/main: `(9,23,9,40)` → `"interp=${token}"`
- new: `(9,33,9,38)` → `token`
- main-product RED run: 3 failed / 0 passed, rc 1, printing the old tuple for every driver.

## Validation

- targeted `InterpolatedStringTests`: 50 passed
- updated PR #3250 driver pin: 3 passed
- full nine-project suite: 15,407 passed, 0 failed, 4 skipped; nine project results, no no-result
- ILVerify: 123/123 verified, 0 verification failures, 0 compile failures, no missing-module output
- emitted smoke: compile rc 0; `dotnet exec` rc 0; exact stdout `1 / 42`
- CI: 30 successful, 2 skipped, 0 failed/pending

Refs #3258.


DavidObando:
## Rubber-duck review — **MERGEABLE**, 0 BLOCKER · 2 SHOULD-FIX · 2 NIT

Reviewed `eabff1272` against `main` **`3e6dd50a7`** (fetched at review time; head is literally `main`+1). `merge-tree --write-tree` rc 0, **0 `CONFLICT` lines and no stage entries at all**, so no modify/delete either.

The fix is correct and the pin is real. Both SHOULD-FIXes are about *what the pin covers* and *what the issue claims*, not about the code that landed.

---

### Q1 — the pin is genuinely column-asserting, and it fails for the right reason

The test asserts `Assert.Single`, the id, **and** all four coordinates as one string, so an id-only pass is impossible. I reverted each product hunk **individually** and rebuilt each time. All four `GSharp.Core.dll` md5s are distinct, so no mutant is a no-result:

| mutant | `GSharp.Core.dll` md5 | new pin | full `Core.Tests` |
|---|---|---|---|
| M0 head | `75a4b8dd` | PASS | 7179 passed / 1 skipped / **7180** |
| M1 revert `BoundTreeRewriter.cs` | `8bdceeeb` | **PASS** | **7179 passed / 1 skipped / 7180 — green** |
| M2 revert `OverloadResolver.CallBinding.cs` | `4a1dbb32` | **FAIL** | — |
| M3 revert `ExpressionBinder.cs` | `ffec7a3a` | **PASS** | **7179 passed / 1 skipped / 7180 — green** |

M2 reproduces the author's RED exactly:

```text
Expected: "(10,32,10,43)"
Actual:   "(10,23,10,45)"
```

Restore is byte-real: after `git checkout HEAD -- …; git reset`, `git diff HEAD` is empty **and** the md5 returns to `75a4b8dd89bb523e804c25d98bd3194e`, the M0 value.

### Q2 — the new span is correct; I measured the text, not the numbers

I printed `Location.Text.ToString(Location.Span)` rather than trusting arithmetic:

- new `(10,32,10,43)` covers **`MakeToken()`** ✅
- old `(10,23,10,45)` covered **`"token=${MakeToken()}"`** — the whole literal

Arithmetic agrees. Line 10 is `␣␣␣␣Console.WriteLine("token=${MakeToken()}")`; 1-based col 32 is the `M`, `MakeToken()` is 11 chars spanning cols 32–42, and the rendered `43` is `Span.End`, which is exclusive. **The fix moves the span onto the offending construct, not from one wrong place to another.**

---

### SHOULD-FIX 1 — GS0519 is still mis-anchored on four sibling shapes, and one of them is a three-token fix

Ten hole shapes measured on **this branch**:

| hole | span | covers | |
|---|---|---|---|
| free call `MakeToken()` | `(10,32,10,43)` | `MakeToken()` | ✅ |
| generic `MakeToken[int32]()` | `(10,32,10,50)` | `MakeToken[int32]()` | ✅ |
| call with folded arg `MakeToken(1 + 2)` | `(10,32,10,48)` | `MakeToken(1 + 2)` | ✅ |
| nested-interp arg `MakeToken("n=${n}")` | `(11,32,11,51)` | `MakeToken("n=${n}")` | ✅ |
| `await` arg `MakeToken(await Seed())` | `(14,32,14,55)` | `MakeToken(await Seed())` | ✅ |
| in a lambda body | `(11,36,11,47)` | `MakeToken()` | ✅ |
| **extension `n.Wrap()`** | `(11,23,11,42)` | `"token=${n.Wrap()}"` | ❌ |
| **instance `b.Make()`** | `(15,23,15,42)` | `"token=${b.Make()}"` | ❌ |
| **field `o.inner`** | `(11,23,11,41)` | `"token=${o.inner}"` | ❌ |
| **struct literal `Token{value: 42}`** | `(6,23,6,50)` | `"token=${Token{value: 42}}"` | ❌ |

The extension case is the exact mirror of the hunk you fixed: `OverloadResolver.Invocations.cs:1206`, `:1216`, `:1219` all build `new BoundCallExpression(null, extension, …)`, and the enclosing method **already takes `CallExpressionSyntax ce` as a parameter**. Passing `ce` there is three tokens and closes the most common remaining shape.

I am not asking for all four. But **#3258's title is "GS0519 is anchored to the wrong span"**, and after this PR it still is for `x.Method()`. Please either extend to the extension/instance paths or say explicitly in the PR body that the issue closes only the free-function shape, so it does not read as the whole class.

### SHOULD-FIX 2 — the hunk the issue names as the root cause is neither the cause nor pinned

#3258 requirement 1 is "apply the `BoundTreeRewriter.cs:623` fix"; requirement 2 is "add a column-asserting pin". **This PR does both, but they do not meet.**

- `RewriteCallExpression` early-outs — `if (builder == null) { return node; }` — so a call whose arguments are unchanged is returned **as-is** and never loses its syntax. The specimen `MakeToken()` has zero arguments.
- `InterpolatedStringHandlerLowerer.RewriteInterpolatedStringExpression` reads `node.Parts` **on entry**, before this pass rewrites children, so the rewriter's clone is not what the GS0519 loop sees for a plain hole.
- Measured: M1 leaves **all ten** spans above byte-identical, and **7179/7180 `Core.Tests` green**.

**I am explicitly not saying delete it.** A surviving mutant proves the corpus cannot observe the line, not that the line is dead — and dropping `node.Syntax` on a shared rewriter is a latent defect worth closing regardless. The ask is narrower: the PR body says the rewriter `null` *caused* the wrong span, and my measurement says the fix that moves the span is `CreatePossiblyElidedCall` in `OverloadResolver.CallBinding.cs`. Please correct the causal story so the next reader does not inherit it, and note that the rewriter and method-group hunks ride along unpinned.

The `ExpressionBinder.cs` method-group hunk is in the same position: M3 is green across the whole suite and it has no GS0519 relationship at all. Plausible cleanup, but it is uncovered collateral inside a bug-fix PR.

---

### Collateral — checked, and bounded

`BoundCallExpression.Syntax` has exactly two consumers: `InterpolatedStringHandlerLowerer.cs:153`, and `MethodBodyEmitter.Expressions.cs:42`, which sets `currentNode` → `CurrentAnchor`. `CurrentAnchor` anchors **emit-time errors only**; sequence points are collected separately, so **the PDB line table is unaffected**. Giving every call a syntax therefore makes emit-error anchors more precise and changes nothing else I can find.

Verified green myself, full denominators:

- `Core.Tests` **7179 passed / 1 skipped / 7180**
- `Compiler.Tests` **4268 passed / 0 failed / 4268**
- Release solution build **0 warnings, 0 errors**
- ILVerify **123/123**, rc 0, and `grep -c 'Assembly or module not found'` = **0**
- and I **ran the artifact**: `gsc /out:ext.dll samples/ExtensionFunctions.gs` → `dotnet ext.dll` → rc 0, correct output. Green ILVerify alone is not proof the CLR will JIT it.

### NIT 1 — the rendered span mixes conventions

`$"({StartLine + 1},{StartCharacter + 1},{EndLine + 1},{EndCharacter + 1})"` — the start `+1` converts 0-based-inclusive to 1-based-inclusive, but `EndCharacter` derives from `Span.End`, which is **exclusive**, so the rendered `43` is one past the last covered column (42). The value is right; the label "one-based span" invites a future reader to mis-count by one. A one-line comment saying the end is exclusive would fix it.

### NIT 2 — one of seventy-three

`BoundTreeRewriter.cs` still contains **72** other `new Bound*(null, …)` constructions after this change, including `BoundFieldAccessExpression`, `BoundPropertyAccessExpression`, `BoundStructLiteralExpression` and `BoundConditionalExpression` — which is why shapes E and F above are still wrong. Not asking for the sweep in this PR; recording it so the class is not assumed closed.

**Verdict: MERGEABLE.** The code is right, the span is right, and the pin can fail. Fix the causal story, and decide deliberately whether #3258 closes on the free-function shape alone.

DavidObando:
# Rubber-duck review — PR #3260 (round two)

**Verdict against `main` = `0232ecb5eb7c1f6878556599a6faf557487e4232`, head `14f3d7ba8d8cc24c37321fea5a604f6a05dace08`.** Both fetched by me and re-fetched immediately before posting. `merge-tree --write-tree` → rc=0, single-line bare tree OID, **0 stage entries**, 0 CONFLICT. Mergeable in the mechanical sense.

**Verdict: ONE BLOCKER.** The craft here is good and materially better than round one — the pins are the strongest I have seen on this diagnostic, and the root-cause write-up is honest. But three shapes in the *same defect class* still reproduce the exact symptom in the issue title, and I have built and measured a **3-file, +14/−7** replacement that fixes all of them, passes every pin this PR adds, and is green on the same suite at the same denominator.

---

## What I verified, and how

I did not check the arithmetic. For every specimen I parsed the four coordinates out of `gsc` stdout and **sliced the actual source text** they cover (one-based inclusive start, exclusive end — I did not "correct" the rendering). Fourteen specimens, three builds, each with a distinct `GSharp.Core.dll` md5:

| build | md5 |
|---|---|
| `main` `0232ecb5e` | `3f22c9fb8798b58eea80b2383784a6c6` |
| this PR `14f3d7ba8` | `82fd7bd665a42be64c194d2fe41eccd9` (round-tripped exactly on rebuild) |
| my alternative | `a6112997804ab3a70402a13873007b87` |

## Span-text verification — the four claimed fixes are correct

Each covers exactly the intended construct, confirmed by slicing the source, not by trusting the numbers:

| shape | PR span | text actually covered |
|---|---|---|
| extension | `(11,32,11,40)` | `n.Wrap()` ✓ |
| instance | `(13,32,13,40)` | `b.Make()` ✓ |
| field | `(11,32,11,39)` | `o.inner` ✓ |
| struct literal | `(6,32,6,48)` | `Token{value: 42}` ✓ |

**The pins are strong.** `AssertByRefLikeInterpolationSpan` (`test/Core.Tests/CodeAnalysis/Binding/InterpolatedStringTests.cs:605-618`) asserts `Assert.Single`, the id, **all four coordinates**, and `location.Text.ToString(location.Span)`. An id-only assertion would have proven nothing; this cannot pass in the broken state. Baking the covered-text check into the assertion is exactly right.

## Nothing regressed — and the improvement is larger than claimed

Against `main`, **all** shapes anchor to the whole literal, including the six the table calls "unchanged":

```
main:  01_free (10,23,10,45) '"token=${MakeToken()}"'      →  PR: (10,32,10,43) 'MakeToken()'
main:  07_ext  (11,23,11,42) '"token=${n.Wrap()}"'         →  PR: (11,32,11,40) 'n.Wrap()'
...all 12 whole-literal on main
```

**10 of 12 fixed, 0 regressed.** No span that was already correct moved. The "unchanged" column is relative to the round-one head, not `main` — accurate about this PR's internal history, but see NIT 1.

## Root-cause statement is now honest

The body says plainly that `CreatePossiblyElidedCall` is causal, that `BoundTreeRewriter.cs` and the method-group changes are *"defensive corrections to the same null-`Syntax` defect class"*, *"not causal"*, *"remain unobserved by the current suite"*, and *"the PR does not claim those lines are pinned."* That matches what measurement showed. No claim survives that `BoundTreeRewriter.cs:623` caused the mis-anchoring.

**I could not construct a GS0519 specimen reaching either defensive hunk, and I am not recommending their removal.** A specimen I cannot build proves my corpus cannot observe the line, not that the line is dead. Keep them.

**No papering-over guard was added.** `InterpolatedStringHandlerLowerer.cs:153` is untouched; the pre-existing `?? node.Syntax.Location` fallback remains a fallback, and the fix went to the construction sites rather than a null-check at the consumer. That was the right instinct.

---

## BLOCKER — three shapes still anchor to the whole string literal

The ten-shape enumeration contains only *direct* calls and member accesses. **It has no specimen that invokes a function-typed value.** That entire class is still broken on this head:

| specimen | PR head span | text covered | should be |
|---|---|---|---|
| call a lambda held in a local — `let f = func() Token {...}` then `"${f()}"` | `(7,23,7,37)` | `"token=${f()}"` | `f()` |
| call a `func() Token` **parameter** — `func Use(f func() Token)` | `(6,23,6,37)` | `"token=${f()}"` | `f()` |
| call a method-value local — `let g = MakeToken` then `"${g()}"` | `(11,23,11,37)` | `"token=${g()}"` | `g()` |

Reproduction (`gsc`, rc=1, diagnostic on stdout):

```gsharp
ref struct Token {
    var value int32
}

func Use(f func() Token) {
    Console.WriteLine("token=${f()}")
}

func Main() {
}
```
```
(6,23,6,37): error GS0519: By-ref-like value of type 'Token' ...
                    ^ covers "token=${f()}" — the whole literal
```

This is not a new or adjacent defect. It is **the symptom named in the issue title**, produced by the same `null`-`Syntax` mechanism, on a shape an ordinary user writes. Round one was returned because four shapes still did this; three still do.

For fairness: the lambda row in the table (*interpolation inside a lambda body*) **is** correct — I measured it at `(10,45,10,56)` covering `MakeToken()`. My specimens are a different shape that the enumeration does not reach.

## The enumeration gap is a symptom of where the fix was applied

The body argues:

> There is no constructor-level fix that can recover the class in one place: these are four immutable bound-node types, and their constructors cannot reconstruct a full source range from already-bound children.

That is true **about constructors** — and that is the point. The fix does not belong on the bound nodes. `InterpolatedStringHandlerLowerer` does not need the bound node's syntax; it needs **the hole's** syntax. And the hole's syntax is already in hand at a single site.

`ExpressionBinder.Literals.cs:449-465` binds every hole in one loop, and already uses `segment.Expression.Location` two lines earlier for other diagnostics:

```csharp
foreach (var segment in syntax.Segments)
{
    if (segment.IsExpression)
    {
        var bound = BindExpression(segment.Expression);
        bound = BindNaturalMethodGroupValue(bound, segment.Expression.Location);   // ← already here
        ...
        Diagnostics.ReportCannotConvert(segment.Expression.Location, ...);          // ← and here
        parts.Add(BoundInterpolatedStringPart.FromHole(bound, segment.Alignment, segment.Format));
    }
```

`FromHole` has **exactly one caller**, and `WithValue` preserves every other field — so a hole-level syntax survives `SpillSequenceSpiller.cs:1995`, which rebuilds the container as `new BoundInterpolatedStringExpression(null, parts, handler)` and would otherwise lose it.

## Measured alternative — 3 files, +14/−7, fixes 14/14

Not asserted. Built on `main`, compiled Release with `-p:ContinuousIntegrationBuild=true`, **0 warnings**, and run:

<details><summary>full diff (also saved locally as <code>ALTERNATIVE.diff</code>)</summary>

```diff
--- a/src/Core/CodeAnalysis/Binding/BoundInterpolatedStringPart.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundInterpolatedStringPart.cs
+using GSharp.Core.CodeAnalysis.Syntax;
+
 namespace GSharp.Core.CodeAnalysis.Binding;
 
-    private BoundInterpolatedStringPart(string literal, BoundExpression value, int? alignment, string format)
+    private BoundInterpolatedStringPart(string literal, BoundExpression value, int? alignment, string format, SyntaxNode holeSyntax)
     {
         ...
+        HoleSyntax = holeSyntax;
     }
 
+    /// <summary>Gets the hole's own syntax, or <c>null</c> when this part is literal text.</summary>
+    public SyntaxNode HoleSyntax { get; }
 
     public static BoundInterpolatedStringPart FromLiteral(string text)
-        => new(text ?? string.Empty, value: null, alignment: null, format: null);
+        => new(text ?? string.Empty, value: null, alignment: null, format: null, holeSyntax: null);
 
+    /// <param name="holeSyntax">The hole's own syntax, used to anchor hole-level diagnostics.</param>
-    public static BoundInterpolatedStringPart FromHole(BoundExpression value, int? alignment, string format)
-        => new(literal: null, value, alignment, format);
+    public static BoundInterpolatedStringPart FromHole(BoundExpression value, int? alignment, string format, SyntaxNode holeSyntax = null)
+        => new(literal: null, value, alignment, format, holeSyntax);
 
     public BoundInterpolatedStringPart WithValue(BoundExpression value)
-        => new(Literal, value, Alignment, Format);
+        => new(Literal, value, Alignment, Format, HoleSyntax);

--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Literals.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Literals.cs
-                parts.Add(BoundInterpolatedStringPart.FromHole(bound, segment.Alignment, segment.Format));
+                parts.Add(BoundInterpolatedStringPart.FromHole(bound, segment.Alignment, segment.Format, segment.Expression));

--- a/src/Core/CodeAnalysis/Lowering/InterpolatedStringHandlerLowerer.cs
+++ b/src/Core/CodeAnalysis/Lowering/InterpolatedStringHandlerLowerer.cs
-                var location = part.Value.Syntax?.Location ?? node.Syntax.Location;
+                var location = part.HoleSyntax?.Location ?? part.Value.Syntax?.Location ?? node.Syntax.Location;
```
</details>

| | this PR | alternative |
|---|---|---|
| production files touched | **9** | **3** |
| production diff | **+43 / −38** | **+14 / −7** |
| shapes correct (of my 14) | **11** | **14** |
| the 3 broken shapes above | whole literal | `f()`, `f()`, `g()` — correct |
| **this PR's own 5 new span pins** | pass | **pass, unmodified — 41/41** |
| `Core.Tests` | 7183 passed / 0 failed / 1 skipped | **7183 passed / 0 failed / 1 skipped** |
| Release CI-flagged | 0 warnings | 0 warnings |

I checked out **this PR's test file verbatim** onto the alternative and it passed 41/41 without editing a single expectation — the alternative produces byte-identical spans for all four fixed shapes.

**Why it is complete by construction, which the threading approach cannot be:** every hole reaches `FromHole` from that one loop, and `segment.Expression` *is* the hole's syntax by definition. The location therefore never depends on whether any bound node happens to carry syntax. Threading `sourceSyntax` through five binder layers fixes the paths someone enumerated; this fixes the paths that exist. `BoundTreeRewriter.cs` alone still holds 72 other `new Bound*(null, …)` constructions — the threading approach is sampling that class, and my three specimens are what sampling misses.

I am not claiming the PR's changes are worthless: giving bound nodes real syntax plausibly improves *other* diagnostics anchored off them. But that benefit is unmeasured and unpinned, and the PR body itself says those hunks are unobserved.

---

## NIT 1 — the "before" column is not reproducible from `main`

The table's "before" values are the round-one head, not `main`. Anyone re-measuring from `main` gets whole-literal spans for all ten rows and will think the table is wrong. One sentence naming the baseline fixes it.

## NIT 2 — five of the ten measured shapes are measured but not pinned

New pins cover free, extension, instance, field and struct literal. Generic, argumented, nested, awaited and lambda-body are measured in the body but have **no** span assertion anywhere — `GS0519` appears in only two test files, and `Issue3114ReadOnlySpanInterpreterTests.cs:297,313` uses `Contains(": error GS0519:")` with no span at all. A regression in generic-call anchoring would be caught by nothing. The helper already exists, so each extra pin is a few lines.

## PR #3250 — clean, verified rather than inferred

`git merge-tree --write-tree pr3250 pr3260` → **bare tree OID, 0 stage entries**. Whichever lands second auto-merges; no rebase needed. The files overlap (`ExpressionBinder.Literals.cs`) but the hunks are ~916 lines apart (#3250 at `:462`, this PR at `:1378`).

Two things worth knowing:
- **The alternative above *would* conflict with #3250**, which inserts its ref-struct-`ToString` block immediately above the `FromHole(...)` call the alternative edits. It is a one-line context conflict, trivially resolved, but it is real and I would rather state it than have it surprise someone.
- The interaction is semantic, not just textual: #3250 makes ref structs with a `ToString` stop reaching GS0519 at all, which changes which specimens this diagnostic sees.

---

**Method:** own clone with its own `out/`, never the shared checkout; no `git stash`; `merge-tree` read by stages, never by grepping for markers; diagnostics captured from stdout with `rc` taken before any pipe; every mutation asserted with `git diff HEAD` and every anchor uniqueness-checked; all three builds produced distinct `GSharp.Core.dll` md5s and the PR build round-tripped exactly; worktree restored to `git diff HEAD` = 0 and verified with a clean re-run.

DavidObando:
## Round-three evidence (`main` `65578d41b2`, head `eced51a7a5`)

Adopted bind-time approach. Three production files, +14/-7: capture `segment.Expression` at sole interpolation-hole binding loop, preserve it through `WithValue`, consume it for GS0519. This fixes 14/14 by construction. Earlier bound-node threading fixed 11/14 and required more files.

### 14-shape sliced-span table

Coordinates use one-based inclusive starts and exclusive ends. Both text columns come from `Location.Text.ToString(Location.Span)`; each pin also asserts `GS0519`.

| shape | current `main` span / sliced text | head span / sliced text |
|---|---|---|
| free call | `(10,23,10,45)` / `"token=${MakeToken()}"` | `(10,32,10,43)` / `MakeToken()` |
| generic call | `(10,23,10,52)` / `"token=${MakeToken[int32]()}"` | `(10,32,10,50)` / `MakeToken[int32]()` |
| argumented call | `(10,23,10,50)` / `"token=${MakeToken(1 + 2)}"` | `(10,32,10,48)` / `MakeToken(1 + 2)` |
| nested-interpolation argument | `(11,23,11,53)` / `"token=${MakeToken("n=${n}")}"` | `(11,32,11,51)` / `MakeToken("n=${n}")` |
| awaited argument | `(14,23,14,57)` / `"token=${MakeToken(await Seed())}"` | `(14,32,14,55)` / `MakeToken(await Seed())` |
| lambda body | `(11,27,11,49)` / `"token=${MakeToken()}"` | `(11,36,11,47)` / `MakeToken()` |
| extension call | `(11,23,11,42)` / `"token=${n.Wrap()}"` | `(11,32,11,40)` / `n.Wrap()` |
| instance call | `(13,23,13,42)` / `"token=${b.Make()}"` | `(13,32,13,40)` / `b.Make()` |
| field access | `(11,23,11,41)` / `"token=${o.inner}"` | `(11,32,11,39)` / `o.inner` |
| struct literal | `(6,23,6,50)` / `"token=${Token{value: 42}}"` | `(6,32,6,48)` / `Token{value: 42}` |
| lambda-valued local | `(7,23,7,37)` / `"token=${f()}"` | `(7,32,7,35)` / `f()` |
| function parameter | `(6,23,6,37)` / `"token=${f()}"` | `(6,32,6,35)` / `f()` |
| method-value local | `(11,23,11,37)` / `"token=${g()}"` | `(11,32,11,35)` / `g()` |
| local value | `(7,23,7,39)` / `"token=${token}"` | `(7,32,7,37)` / `token` |

Result: original ten = **10 fixed, 0 regressed**; expanded matrix = **14 fixed, 0 left population**.

### Defensive hunks and positive control

Removed round-two bound-node threading because bind-time carrier replaces its role; no claim that those lines were dead. Removal is dependency elimination, not survival inference. Positive control: restoring only three production files to current `main` while keeping pins produced 14 matrix failures plus 3 driver failures, each slicing whole literal. Restoring production made 50/50 and 3/3 green. Thus suite observes required GS0519 path; this does not claim removed code could never help unrelated diagnostics.

Freshness witness:
- fixed: `321bbfbad9930103c0ef91fd10e07c98`
- main-product mutant: `21bfa7cfbc58ec0587a0b41ab51063c7`
- restored: `321bbfbad9930103c0ef91fd10e07c98`

`git diff HEAD` was empty after restore. One mistyped exploratory filter produced `No test matches`; discarded. Corrected driver selector ran 3/3.

### #3250 and NITs

- #3250 merged as `f5642aff1`. Its ref-struct `ToString` block remains intact; `segment.Expression` is passed after it. Rebase from `acc10fe470` to `65578d41b` was conflict-free.
- NIT 1: before column now measures reproducible current `main` `65578d41b`, not a prior PR head.
- NIT 2: all 14 shapes pinned with ID, four coordinates, and sliced text. Original five pins remain unmodified.

### Validation

- targeted: 50/50 interpolation tests; 3/3 file-driver pins
- Release CI build, warnings-as-errors: **0 warnings, 0 errors**; Release Core md5 `0bc83a6d3afc8ee6f210c14d10a523c0`
- full Release suite: **9/9 projects, 15,407 passed, 4 skipped, 0 failed**
- ILVerify: **123/123**, 0 verification failures, 0 compile failures
- emitted runtime: compile rc 0; `dotnet exec` rc 0; exact stdout `1 / 42`; empty stderr

No adjacent defect found.

DavidObando:
CI complete on `eced51a7a5`: 30 successful, 2 skipped, 0 failed/pending. PR is mergeable and clean.

Issue #3256, opened 2026-08-05
[docs] Documentation still advertises the removed --engine evaluator; pin the gsi --help transcript so it stops drifting


DavidObando:
## Summary

#3244 (ADR-0156 Phase 3c, `8b3bdac57`) retired the tree-walking evaluator. `gsi` now rejects the old engine selection **loudly**:

| invocation | `main` @ `8b3bdac57` |
|---|---|
| `gsi --engine evaluator f.gs` | **rc 1** — `Unknown engine 'evaluator'. Expected 'emit'.` |
| `GSI_ENGINE=evaluator gsi f.gs` | **rc 1** — same |
| `gsi f.gs` / `gsi --engine emit f.gs` | rc 0 |

Documentation across the repo still describes the evaluator as present or selectable. A user following the docs is told to pass a flag value that now exits 1.

## Known stale sites

- **`website/docs/tooling/repl.md:36-47`** — a **verbatim `gsi --help` transcript** that still advertises `'evaluator'` as a selectable engine. This transcript has now drifted **three separate times**; `git grep "Usage: gsi"` returns exactly one hit repo-wide and **nothing pins it**, so it will drift again.
- **`docs/adr/0153-interpreter-compiled-only-storage-boundary.md:41`** — still states in the **present tense** that interactive `gsi --engine evaluator` *"must report a self-contained boundary diagnostic"*, while `:3` already marks the ADR superseded. The diagnostics it refers to (`GS0510`/`GS0513`/`GS0514`) were deleted by #3244.
- **`docs/adr/0152-interpreter-native-call-boundary.md:35-39`** — says the evaluator *"is deprecated and scheduled for removal in Phase 3c"*; it has **already been removed**.

`grep -rln "engine evaluator\|--engine evaluator" website/docs docs/` currently matches a good deal more than the three sites above, including `website/docs/ref/feature-matrix.md`, `website/docs/ref/clr-interop.md`, `website/docs/ref/spec.md`, `website/docs/intro.md`, `website/docs/faq.md`, `website/docs/guide/effective-gsharp.md`, and `website/docs/tooling/compiler-architecture.md`. **Sweep, do not fix only the three named.**

## Ask

[...]

Issue #3257, opened 2026-08-05
[tests] Interp-vs-emit parity oracles now compare emit against emit after Phase 3c; names and intent no longer match


DavidObando:
## Summary

`#3244` (ADR-0156 Phase 3c) deleted the tree-walking evaluator. A family of tests was built as **two-engine differential oracles** — interpreter vs emit — and those tests survived the deletion by having their interpreter column repointed at an emit-backed surface. They now compare **emit against emit**, so their names and stated intent no longer describe what they verify.

This is a **test-integrity** issue, not a product bug. Filed because differential oracles are the strongest tests in this repo and their silent downgrade is easy to miss: they still pass, still look like parity coverage, and nothing flags that the second engine is gone.

## Confirmed sites

| file | current state |
|---|---|
| `test/Core.Tests/CodeAnalysis/Emit/AsyncInterpVsEmitParityTests.cs` | Named `InterpVsEmit`, but only calls `RunEmitter`. Now a single-engine test asserting a hardcoded `Expected`. |
| `test/Interpreter.Tests/Issue3140OutParameterWriteBackParityTests.cs` | `Assert.Equal(emitted, inProcess)` where `emitted` is a `gsc` compile-and-run and `inProcess` is `EmittedOracle.Evaluate`. **Both emit.** |
| `test/Interpreter.Tests/Issue3099TypeArgumentReificationTests.cs:195` | `RunInteractive` constructs `EmittedSessionEngine`, so its comparison against the `gsc` driver column is emit vs emit. |

Also review, for the same shape: `Issue3100FieldOffsetParityTests.cs`, `Issue2928TupleFunctionParityTests.cs`, `Issue3015ImportedBaseIdentityTests.cs`, `Issue2896StructObjectOverrideTests.cs`.

## Be precise about what is and is not lost

**Do not simply delete these.** Comparing an out-of-process `gsc` compile-and-run against the in-process emitted oracle still exercises two genuinely different drivers, and that has real value — it catches driver-specific divergence, which this session has seen more than once. What is gone is the *cross-engine* claim.

The defensible outcomes, per test, are:

1. **Rename and re-document** where the driver-vs-driver comparison is still meaningful — the name must say what it compares.
2. **Convert to a single-engine pin with an explicit expected value** where the second column adds nothing.
3. **Delete** only where the test is genuinely vacuous.

For each test, state which of the three applies and why. A test whose two sides are now literally the same call is the only case where deletion is automatic.

## The thing to actually check

[...]

PR #3261, opened 2026-08-05
Remove stale evaluator documentation and pin gsi help

DavidObando:
## Summary

- update current website and architecture docs for ADR-0156 Phase 3c: every driver emits, and `gsi` accepts only `emit`
- preserve evaluator design history in ADRs while converting stale present/future claims to explicit past tense
- keep `docs/diagnostics.md` and its website mirror aligned
- pin the website's verbatim `gsi --help` transcript to the real `Program.Main("--help")` output
- update the cs2gs inventory source alongside its generated matrix

## Measured help

Rebuilt `out/bin/Release/Repl/gsi --help` returns rc 0 and prints:

```text
  --engine <name>        Interactive engine: 'emit' (the default and only
                         engine; also via GSI_ENGINE). The legacy 'evaluator'
                         engine was removed in ADR-0156 Phase 3c.
```

## Transcript pin and RED proof

`Adr0156DefaultEngineFlipTests.WebsiteHelpTranscript_MatchesDriverOutput` extracts the fenced transcript from `website/docs/tooling/repl.md` and compares it exactly with the driver output.

Perturbing only `the default and only` to `the default only`, then rebuilding/running that test:

```text
MUTANT_TEST_RC=1
Expected: ···"'emit' (the default and only\n            "···
Actual:   ···"'emit' (the default only\n                "···
Failed: 1, Passed: 0, Total: 1
```

Restored from HEAD; `git diff HEAD` was empty; all 8 ADR-0156 engine tests passed.

## Residue sweep

```sh
git grep -n -Ei 'Pass .*--engine evaluator|(--engine evaluator|GSI_ENGINE=evaluator).*(selects?|select the|scheduled for removal|deprecated escape|available through|remains reachable)|evaluator.*scheduled for removal|scheduled for removal.*evaluator' -- docs website/docs
```

- base `5e2fee875`: 12 hits
- this branch: 0 hits

Eleven broader `engine evaluator|--engine evaluator` matches remain deliberately: seven past-tense ADR records, two ADR-0156 removal/status records, and two current website statements saying the forms were removed. None instructs users to select the deleted backend.

## Validation

Rebased onto `3e6dd50a7390366120547f9b25e12063f2095cfd`.

- nine solution test projects: 15,370 passed, 4 skipped, 0 failed
- targeted transcript/engine tests: 8/8 passed
- diagnostics mirror test: 1/1 passed
- Docusaurus production build/link validation: success
- `GSharp.Core.dll` freshness witness: `1f87e64fc3b4d061a1b16e126e9d2b3a`
- worktree clean

Addresses #3256


DavidObando:
## ✅ MERGEABLE — 0 BLOCKER, 0 SHOULD-FIX, 1 NIT

Verdict is against **`main` = `3e6dd50a7390366120547f9b25e12063f2095cfd`**, re-fetched immediately before posting, and head **`d199b1b767ef3042dcfeb827c5bdc431c817ea92`**. The merge-base **equals** `main`, so this is a fast-forward; `merge-tree --write-tree` read by **stages** gives **0 stage-1/2/3 entries** and 0 `CONFLICT` lines. Built in a private worktree: 0 Warning(s), 0 Error(s), `GSharp.Core.dll` md5 **`4e699356f114921ac46d1a32da1451dc`**.

### The pin is the valuable half, and it is real

Two sides have genuinely independent origins — `RunMain("--help")` invokes **`ReplProgram.Main(args)`**, the live driver entry, while the other side is the checked-in markdown fence in `website/docs/tooling/repl.md`. This is **not** the #3257 render-both-sides-from-one-source shape. I proved it behaviourally rather than by reading, in **both** directions:

| perturbation | result | failure quoted |
|---|---|---|
| baseline, untouched head | **8 passed / 0 failed** | — |
| **doc side**: `repl.md` `exit.` → `quit.` | **RED** 1F/7P | `Expected …"version and exit.\n"… / Actual …"version and quit.\n"…` |
| **driver side**: `src/Repl/Program.cs:49` → `EXIT-DRIFT.` | **RED** 1F/7P | `Expected …"version and EXIT-DRIFT.\n"… / Actual …"version and exit.\n"…` |

The second row is the one that matters: the injected string appeared on the **Expected** (driver) side, so the pin demonstrably reads live `--help` output, not a second copy of the prose. That is the drift that escaped three times. Both mutants confirmed applied with `git diff HEAD --numstat` (`1  1`), both restored, md5 round-trip on `repl.md` exact (`d8fbf743371bf582816e2f46557d0bb6`), `git status --porcelain` empty.

*(One `dotnet test` invocation segfaulted at restore, rc=139, with no summary line. I scored that as a no-result and re-ran rather than counting it as RED.)*

### Claims verified against my own rebuilt binary, not against prose

Control first: `gsi /tmp/a.gs` → **rc=0**, stdout `42`. Then:

- `gsi --engine evaluator a.gs` → **rc=1**, stderr `Unknown engine 'evaluator'. Expected 'emit'.`
- `GSI_ENGINE=evaluator gsi a.gs` → **rc=1**, same message.

The documented transcript in `repl.md` is **byte-identical** to my rebuilt `gsi --help`. Two probes I discarded as no-results on the way: `--engine evaluator /dev/null` reports `Unrecognized argument: /dev/null` (an argument guard firing first), and `GSI_ENGINE=evaluator gsi --version` exits **rc=0** because `--version` short-circuits before engine validation. Neither measures the engine.

### Mirror trap — three pairs, all handled

| pair | status |
|---|---|
| `docs/diagnostics.md` ↔ `website/docs/ref/diagnostics.md` | **both edited identically** ("compiler or evaluator failure" → "compiler or runtime execution failure"); the extra line in the website diff is a trailing newline |
| `docs/cs2gs-coverage-matrix.md` ↔ `tools/cs2gs/coverage/csharp-construct-inventory.json` | **both edited identically**, and this pair is **machine-enforced** by `ConstructInventoryGoldenTests` — I ran it: **7 passed** |
| `docs/lsp.md` ↔ `website/docs/tooling/lsp.md` | contains **no** evaluator terms; correctly untouched |

Note the two diagnostics files are deliberately *different documents* (public rewording, different link forms), so line-diffing them is the wrong test — I compared the factual claims instead, and both sides mark GS0510/GS0511/GS0513/GS0514 retired in past tense.

### Completeness sweep, re-run on the head

`scheduled for removal` → **0 files**. `will be removed` → **0**. `Phase 3c will` → **0**. Surviving mentions of `engine evaluator` / `GSI_ENGINE` / `SessionEngine` / `Compilation.Evaluate` are all correct past-tense history — the live user-facing surfaces read, verbatim:

- `website/docs/intro.md:30` — "were removed in ADR-0156 Phase 3c."
- `website/docs/tooling/compiler-architecture.md:39` — "Phase 3c removed the legacy tree-walking evaluator, its `Compilation.Evaluate` API, and the … escape hatches."

No stale advertisement of the flag survives anywhere I could find.

### History was fixed, not deleted

Both named sites convert **tense** while preserving content, under status lines that already carry clear `Superseded by ADR-0156` markers: `#3199 tracks moving` → `tracked`, `The current implementation status is` → `The final pre-deletion implementation status was`, `scheduled for removal in Phase 3c` → `Phase 3c deleted the evaluator path and GS0514`. Nothing was rewritten out.

### Tests

`Interpreter.Tests` full: **1346 passed, 3 skipped, 0 failed**. `ConstructInventoryGoldenTests`: **7 passed**. `No test matches` count **0** in every log. CI: **31 SUCCESS / 3 SKIPPED / 0 failing**.

---

### NIT — tense over-correction on statements that are still true

Two sentences were converted to past tense even though they describe things that remain true today:

- `docs/adr/0153-interpreter-compiled-only-storage-boundary.md:49` — "It **was** a permission boundary, not a storage operation." `unsafe` is still a permission boundary.
- `docs/adr/0152-interpreter-native-call-boundary.md:38` — "Bare `gsc`, `gsi <file>`, and the default interactive REPL **emitted and ran** the native call; `gsc /out:` **emitted** it to disk." All still current behaviour — I measured `gsi <file>` returning rc=0 above.

Harm is low: both sit inside ADRs with prominent `Superseded` markers, so a reader is already told the document is historical. Worth a touch-up only if the file is opened again; not worth a round-trip on its own.

Nothing here blocks. The pin closes a real recurring gap and I could not find a way to make it pass while the documented and actual help disagree.

Issue #3255, opened 2026-08-05
[gsc] 'chan T' as an enclosing type argument raises GS9998 ArgumentNullException with a non-anchored (1,1,1,1) span


DavidObando:
## Summary

Using `chan T` as an enclosing type argument produces an unhandled internal error rather than a diagnostic:

```
GS9998  ArgumentNullException ... (Parameter 'key')
```

reported with a **non-anchored span of `(1,1,1,1)`**, so the error points at the start of the file instead of the offending construct.

Found while rubber-ducking PR #3229. Verified **byte-identical on that PR's base and head**, so it is pre-existing on `main` and not caused by it.

## Two defects, not one

1. **The `ArgumentNullException` itself.** A `null` dictionary key is reaching a lookup on the `chan T` path. Whatever produces the key for a channel type as a type argument is returning `null`.
2. **The non-anchored `(1,1,1,1)` span.** Even once the ICE is fixed, an internal error that cannot point at the construct that caused it is very hard to act on. This is the **same non-anchored ICE shape as #3192**, which is still open — so this is at minimum a second instance of a live class of defect, and the two may share a fix.

## Expected

Either `chan T` is legal as an enclosing type argument — in which case it should work — or it is not, in which case it needs a **real diagnostic anchored to the offending span**, not a `GS9998` internal error at `(1,1,1,1)`.

Decide which, and record the reasoning. Do not paper over the `ArgumentNullException` with a null check that silently yields a wrong type: that converts a loud crash into a silent wrong answer, which is strictly worse.

## Notes for whoever takes this

- Check #3192 first — if the span-anchoring defect is common, fix it once.
- `gsc` writes diagnostics to **stdout**, only `Failed.` to stderr — **capture `rc` before any pipe**.
- The tree-walking evaluator was deleted in #3244 (`8b3bdac57`), so there is no interp-vs-emit oracle; the oracle is the running program plus the diagnostic text and span.
- Pin it so it fails for the right reason: assert the **span** as well as the id, or the anchoring half of this issue is untested.

PR #3262, opened 2026-08-05
Parse composite types in enclosing generic arguments

DavidObando:
References #3255.

## Decision

`chan T`, `sequence[T]`, and `map[K,V]` are legal generic type arguments, including as enclosing arguments.

Their real `ParseTypeClause` productions and existing symbol/emitter support already make them first-class types. Rejecting them only in speculative generic-call lookahead was an arbitrary parser inconsistency. CLR generic arguments cannot be pointer or function-pointer types. Plain `*T` remains on its existing anchored parser-rejection path; the distinct `*func(T) R` production is scanned and rejected as `GS0520` at its source span before emit.

## Root cause and fix

`LooksLikeGenericCallSite` speculatively calls `TryScanTypeClause` before committing `Name[...]` to generic construction. When that scanner misses a legal type-clause head, parsing silently falls back to indexing/access, creates malformed syntax, and can pass null text into a binder dictionary lookup.

The original patch added only `chan`; review confirmed the same gap for `sequence` and `map`. This revision:

- scans `chan`, `sequence`, `async sequence`, canonical `map[K,V]`, legacy `map[K]V`, and nested combinations using the existing recursive scanner;
- marks these shapes complex for trailing-member disambiguation;
- reuses `CanStartTypeClause` as the single complete head-token set and removes the duplicate `CurrentTokenStartsTypeClause` switch;
- keeps plain `*T` on its existing rejection path, while scanning `*func(T) R` far enough to report anchored `GS0521` before emit;

The head-token membership set is now unified. Shape-specific advancement cannot use only that boolean helper: speculative lookahead must consume each production's delimiters and recursively nested component types without mutating parser state or diagnostics.

No null guard or fallback type was added.

## Span and #3192

#3192 and #3255 have different crash causes: #3192 fails during emit; #3255 began with parser lookahead and failed during binding. They do share the same span cause. `src/Compiler/Program.cs:286` hard-codes `(1,1,1,1)` in the top-level ICE handler, and `src/Repl/Program.cs:181` does likewise; `Compilation.cs:604` is the existing anchored internal-diagnostic path.

This PR does not change that global ICE policy. The parser regression still pins the exact offending `chan T` source span, and the pointer rejection pins its diagnostic id plus `(5,28,5,33)`.

## Compile-and-run matrix

Each positive row was compiled and executed with the rebuilt Release `gsc`; stdout was asserted by the emitted-host regression test.

| Enclosing type argument | rc | Observed stdout / diagnostic |
|---|---:|---|
| `chan T` | 0 | `Channel\`1`, `System.Int32`, `Success.` |
| `sequence[T]` | 0 | `IEnumerable\`1`, `System.Int32`, `Success.` |
| `map[string,T]` | 0 | `Dictionary\`2`, `System.String`, `System.Int32`, `Success.` |
| `async sequence[T]` | 0 | `IAsyncEnumerable\`1`, `System.Int32`, `Success.` |
| `map[string,sequence[chan T]]` | 0 | `Dictionary\`2`, `IEnumerable\`1`, `Channel\`1`, `System.Int32`, `Success.` |
| `*int32` | 1 | existing `GS0125` at `(5,28,5,33)`; no `GS9998` and no `Cannot encode` |
| `*func(int32) int32` | 1 | `GS0521` at `(7,27,7,45)`; no `GS9998` and no `Cannot encode` |

The exact review repros `Owner[sequence[int32]].Payload[string]` and `Owner[map[string,int32]].Payload[string]` now compile and run. The round-3 `Owner[*func(int32) int32].Payload[string]` repro now fails cleanly with anchored `GS0521`.

## RED proofs and mutants

- Inverting the sequence closing-bracket check produced the original `GS9998 ArgumentNullException` at `(1,1,1,1)`; the pointer control stayed green.
- Inverting the canonical-map comma check produced map-specific `GS0125`/`GS0005` failures; the pointer control stayed green.
- Both mutations had non-empty product diffs and changed `GSharp.Core.dll` (`904c5760…` to `ea3b6a18…` / `bf71ac26…`). Exact source and DLL hashes restored (`39e3ebde…` / `904c5760…`).
- The original channel head and trailing-complexity mutants remain covered by the parser and emitted-host pins.

## Final verification

- Prior full-suite verification before the round-3 follow-up: **15,372 passed, 4 skipped, 0 failed**.
- After rebasing onto current `origin/main` and adding the round-3 follow-up: targeted Core parser tests **8/8 passed** and targeted emitted-host tests **20/20 passed**.
- ILVerify: **123/123 verified**, two documented suppressions, zero verification failures, zero compile failures, and no missing-assembly output.
- Freshness: 20 direct Release consumer copies of `GSharp.Core.dll`, all non-zero and md5-identical (`87bc0852351ed0dd9731162a93d3cb0a`).

Harness disclosure: the first manual matrix draft used reserved local name `type`; its unrelated `GS9998` was discarded. Corrected probes and regression tests use valid identifiers and produce the table above.


DavidObando:
# Rubber-duck review — PR #3262

**Verdict against `main` = `3e6dd50a7390366120547f9b25e12063f2095cfd`, PR head `ca5491bcafa4b010e656ad4ba1295356429c9c7e`** (both re-fetched immediately before posting; merge-base *is* current `main`, `merge-tree` rc=0 with a bare tree OID and no `CONFLICT` entries).

**MERGEABLE — no blocker.** The diagnosis is right, the fix works, nothing regresses, and the "papering over" trap the issue explicitly warned about was avoided. **But it fixes one of three instances of the defect it correctly diagnosed, and issue #3255's exact signature still reproduces on this PR's own head.** Details below, with a measured patch.

## What is right, and worth saying plainly

The root-cause call is **correct, and I verified the mechanism rather than accepting it.** `TryScanTypeClauseCore` (`src/Core/CodeAnalysis/Syntax/Parser.Expressions.Creation.cs:822`) is a *speculative* scanner: when it fails, the parser silently falls back to the index-expression reading, and the malformed tree is what produces the null key downstream. `chan` was simply not in its head set.

The fix is genuinely lazy in the good sense — 7 lines, reusing the existing recursion (`return TryScanTypeClause(ref pos);`) rather than hand-rolling element parsing, and setting `isComplex = true` consistently with the sibling `[`, `(` and `func` arms. **No null check was added anywhere**, so the "convert a loud crash into a silent wrong answer" failure the issue called out is avoided by construction.

And the "legal, so make it work" decision is right. I did not take it on trust — I ran the emitted programs (the evaluator is deleted, so the oracle is the running program):

| specimen | base `3e6dd50a7` | head `ca5491bc` |
|---|---|---|
| `Owner[T].Payload[string]` *(control)* | ✔ correct | ✔ correct |
| `Owner[chan T].Payload[string]` *(the repro)* | **GS9998 `(1,1,1,1)`** | ✔ `Owner<Channel<int32>>+Payload<string>` |
| `chan T` *(own type-arg position)* | GS0125 | ✔ `Channel\`1[Int32]` |
| `chan int32` | GS0125 | ✔ correct |
| `Owner[Owner[chan T].Payload[int32]].Payload[string]` *(deep nest)* | — | ✔ correct |

That is a real RED→GREEN on the product, with the **correct runtime type printed**, not an rc check. The two new parser tests assert exact `StartLine/StartCharacter/EndLine/EndCharacter`, and the new `Issue3099` test asserts exact stdout across three drivers — that meets ADR-0154, and the name `ReifiesAcrossEmittedHosts` is honest now that `RunInteractive` is `EmittedSessionEngine`.

---

## SHOULD-FIX 1 — `sequence[T]` reproduces issue #3255 *verbatim* on this head

`src/Core/CodeAnalysis/Syntax/Parser.Expressions.Creation.cs:858`

The PR's own root-cause analysis predicts its siblings, and the prediction holds. The real `ParseTypeClause` accepts **nine** type heads — the set is written out twice already, at `Parser.Declarations.cs:925` (`CanStartTypeClause`) and `Parser.Expressions.Literals.cs:1079` (`CurrentTokenStartsTypeClause`), and inline a third time at `Parser.TypeClauses.cs:864`:

```
Identifier  [  (  func  map  chan  sequence  async  *
```

`TryScanTypeClauseCore` hand-rolls a **fourth** copy of that set, and even after this PR it knows only six. Missing: **`map`, `sequence`, `*`**. Measured on this PR's head:

```
$ gsc S.gs      # let value = Box[Owner[sequence[int32]].Payload[string]]()
rc=1
S.gs(1,1,1,1): error GS9998: ArgumentNullException: Value cannot be null. (Parameter 'key')
```

That is issue #3255's title, byte for byte, with `sequence` in place of `chan`, **on `ca5491bc`**. `map[K,V]` fails the same way one step earlier, as the misleading `GS0125: Variable 'Box' doesn't exist` (`Box` is a class, not a variable) — the same scanner gap, surfacing as the index-expression fallback instead of the ICE.

Full matrix, all three built Release CI-flagged with `GSharp.Core.dll` md5 movement confirming each artifact actually changed (`c11f602e` base → `53fe58f3` head → `c2d29f4f` patched):

| specimen | base | head `ca5491bc` | +50-line patch |
|---|---|---|---|
| `Owner[chan T].Payload[string]` | GS9998 `(1,1,1,1)` | ✔ | ✔ |
| `Owner[sequence[int32]].Payload[string]` | GS9998 `(1,1,1,1)` | **GS9998 `(1,1,1,1)`** | ✔ `IEnumerable<int32>` |
| `sequence[int32]` | GS0125 | **GS0125** | ✔ |
| `Owner[map[string,int32]].Payload[string]` | GS0125 | **GS0125** | ✔ `Dictionary<string,int32>` |
| `map[string,int32]` | GS0125 | **GS0125** | ✔ |
| `map[string,sequence[chan int32]]` | — | — | ✔ correct nested type |

**Recommended fix — measured, not asserted.** I implemented it and ran it: extend the same hunk to `map`/`sequence` (they need bracket-list consumption, `chan` does not). **+50/−0 in the one function, versus the PR's +7.** Suites on that patch: **`Core.Tests` 7180 passed / 0 failed / 1 skipped** and **`Interpreter.Tests` 1346 passed / 0 failed / 3 skipped**. Patch is attached at the end of this comment.

**One non-obvious detail I got wrong first, so you don't repeat it: do NOT add `StarToken`.** My first attempt included it for head-set completeness. `*T` then parses and fails *later* instead:

```
S.gs(1,1,1,1): error GS9998: InvalidOperationException: Cannot encode '*...'
```

The CLR forbids pointer generic arguments, so admitting `*T` in the scanner converts a parse-level `GS0125` into an emit-level ICE — strictly worse. My patch excludes `*` with a comment saying why, and `*int32` correctly stays at base behaviour. `chan`, `map` and `sequence` all have full binder+emitter support, which is exactly why admitting *them* is safe; I confirmed that by printing the constructed runtime type for each.

**If you prefer to keep this PR at 7 lines, that is defensible — but then #3255 should not be closed by it**, because its stated signature still reproduces.

## SHOULD-FIX 2 — the issue's *second* defect is untouched, and it is one shared line

The issue asked for two fixes: the crash **and** the non-anchored `(1,1,1,1)` span. The span half is not addressed, and it is not `chan`-specific — it is hard-coded in the driver's top-level exception handler:

- `src/Compiler/Program.cs:286` — `Console.Out.WriteLine($"{file}(1,1,1,1): error GS9998: {message}");`
- `src/Repl/Program.cs:181` — same literal

Every `GS9998` that escapes to the driver, from any cause, is stamped `(1,1,1,1)` there. An **anchored** path already exists — `src/Core/CodeAnalysis/Compilation/Compilation.cs:604` builds `new Diagnostic(location, "GS9998", …)` — so the machinery is present and the driver fallback simply has no location to pass. This is the issue's "if the span-anchoring defect is common, fix it once" case, and it is common: one line, all callers.

## Note on the #3192 separation claim — half right

The claim that #3192 and #3255 have different root causes is **correct for the crash**: #3192 is an emit-path failure, this is a parser head-set gap, and fixing one cannot fix the other. I verified they are independent.

But the claim understates the overlap on the *other* half. The identical `(1,1,1,1)` span in both is not a coincidence of symptom — it is **literally the same line**, `src/Compiler/Program.cs:286`. So: two unrelated crash causes, **one shared span cause**. Worth stating that way, because it means the anchoring fix is a single edit that improves every future ICE report, including both issues.

## NIT — the fourth copy of the head set

The reason this bug class exists is that "what can start a type clause" is written out in four places that drift independently. This PR adds a fifth fragment rather than reusing either existing predicate. Consumption logic genuinely differs per head, so a single predicate cannot replace the scanner outright — but a comment tying `TryScanTypeClauseCore` to `CanStartTypeClause` would make the next drift visible. My patch includes one.

---

## Method / hygiene

Fresh clone with its own `out/` (verified: the shared checkout's `out/` mtime is untouched at Aug 2); never the shared checkout; no `git stash`; no `--no-build`; `rc` captured before every pipe; worktree restored to `ca5491bc` with `git diff HEAD` = 0. Oracle throughout was the **running emitted program's printed type**, never rc or diagnostic id alone. Both alternative builds were confirmed to have actually taken effect by `GSharp.Core.dll` md5 movement, and suite results were read out of the logs rather than trusted from rc.

<details>
<summary>Measured patch (+50/−0) — <code>Parser.Expressions.Creation.cs</code>, replaces the <code>chan</code> hunk</summary>

```csharp
        // ponytail: keep this head set in step with CanStartTypeClause
        // (Parser.Declarations.cs) / CurrentTokenStartsTypeClause
        // (Parser.Expressions.Literals.cs). A head the real ParseTypeClause
        // accepts but this speculative scan rejects makes the parser fall back
        // to the index-expression reading, which surfaces downstream as a
        // bogus GS0125 or a GS9998 null-key ICE (issue #3255).
        // StarToken is deliberately NOT here: the CLR forbids pointer generic
        // arguments, so admitting `*T` would only move the failure from a
        // parse-level GS0125 to an emit-level GS9998 "Cannot encode '*'".
        if (Peek(pos).Kind == SyntaxKind.ChanKeyword)
        {
            isComplex = true;
            pos++;
            return TryScanTypeClause(ref pos);
        }

        if (Peek(pos).Kind == SyntaxKind.MapKeyword
            || Peek(pos).Kind == SyntaxKind.SequenceKeyword)
        {
            isComplex = true;
            pos++;
            if (Peek(pos).Kind != SyntaxKind.OpenSquareBracketToken)
            {
                return false;
            }

            pos++;
            while (true)
            {
                if (!TryScanTypeClause(ref pos))
                {
                    return false;
                }

                if (Peek(pos).Kind == SyntaxKind.CommaToken)
                {
                    pos++;
                    continue;
                }

                break;
            }

            if (Peek(pos).Kind != SyntaxKind.CloseSquareBracketToken)
            {
                return false;
            }

            pos++;
            if (Peek(pos).Kind == SyntaxKind.QuestionToken)
            {
                pos++;
            }

            return true;
        }
```

</details>

DavidObando:
## Round 3 review — **NOT MERGEABLE** (1 BLOCKER, 2 SHOULD-FIX, 2 NIT)

**Refs verified independently.** `main` = `0232ecb5eb7c1f6878556599a6faf557487e4232`, re-fetched immediately before posting and unchanged. Head via `gh pr view --json headRefOid` = `92b287fa94da7c9dd62b6ec76c4d04b6794c4d06`. Merge base `d172d0d14424df9caaa06de65a2a3f52619a1cc4`, 2 behind / 2 ahead. `git merge-tree --write-tree ghr/main pr3262` → rc0, one-line output (tree SHA only), **zero stage entries** — read as stages, not by grepping for `<<<<<<<`, which git 2.50 never emits.

Worktrees at merge-base and head, both `dotnet build GSharp.sln -c Release` → `BUILD_RC=0`, **0 warnings / 0 errors**. `GSharp.Core.dll` md5 differs between trees, so the parser change is genuinely compiled into what I measured.

**The core fix works.** The exact issue repro, bare `gsc`:

| tree | rc | stdout |
|---|---:|---|
| base `d172d0d1` | 1 | `repro.gs(1,1,1,1): error GS9998: ArgumentNullException: Value cannot be null. (Parameter 'key')` |
| head `92b287fa` | 0 | `chan:System.Threading.Channels.Channel\`1` / `chan-arg:System.Int32` / `Success.` |

---

## 1. Independently enumerated `ParseTypeClause` head tokens — from source

Read out of `src/Core/CodeAnalysis/Syntax/Parser.TypeClauses.cs:33` `ParseTypeClauseCore`, dispatch by dispatch, not from the brief. **Exactly 9 head-token kinds**, and `CanStartTypeClause` (`Parser.Declarations.cs:925`) lists precisely those 9 — the unification target is faithful.

Every one driven through as an **enclosing** type argument, `Box[Owner[<T>].Payload[string]]()`, compiled and executed with bare `gsc`, asserting rc **and** printed reflected type:

| # | head token | production probed | base | head `92b287fa` |
|---:|---|---|---|---|
| 1 | `IdentifierToken` | `T` | rc0 `System.Int32` | rc0 `System.Int32` |
| 2 | `IdentifierToken` | `List[T]` | rc0 `List\`1[Int32]` | rc0 `List\`1[Int32]` |
| 3 | `IdentifierToken` (`unmanaged`) | `unmanaged[Cdecl](int32) -> int32` | rc1 `GS0125` | rc1 `GS0125` (identical) |
| 4 | `FuncKeyword` | `func(int32) int32` | rc0 `Func\`2` | rc0 `Func\`2` |
| 5 | `OpenParenthesisToken` | `(int32, string)` | rc0 `ValueTuple\`2` | rc0 `ValueTuple\`2` |
| 6 | `OpenParenthesisToken` | `(int32) -> string` | rc0 `Func\`2` | rc0 `Func\`2` |
| 7 | `OpenSquareBracketToken` | `[]T` | rc0 `System.Int32[]` | rc0 `System.Int32[]` |
| 8 | `OpenSquareBracketToken` | `[3]T` | rc0 `System.Int32[]` | rc0 `System.Int32[]` |
| 9 | `MapKeyword` | `map[string, T]` | rc1 `GS0125 Variable 'Box'` | **rc0 `Dictionary\`2[String,Int32]`** |
| 10 | `ChanKeyword` | `chan T` | **rc1 GS9998 (1,1,1,1)** | **rc0 `Channel\`1[Int32]`** |
| 11 | `SequenceKeyword` | `sequence[T]` | **rc1 GS9998 (1,1,1,1)** | **rc0 `IEnumerable\`1[Int32]`** |
| 12 | `AsyncKeyword` | `async sequence[T]` | **rc1 GS9998 (1,1,1,1)** | **rc0 `IAsyncEnumerable\`1[Int32]`** |
| 13 | `AsyncKeyword` | `async func(int32) int32` | rc0 `Func\`2[Int32,Task\`1]` | rc0 identical |
| 14 | `StarToken` | `*int32` | rc1 `GS0125` ×3 | rc1 `GS0125` ×3 (identical) |
| 15 | `StarToken` | `*func(int32) int32` (unsafe) | **rc1 GS9998 (1,1,1,1)** | **rc1 GS9998 (1,1,1,1)** ← **BLOCKER** |

Rows 9–12 are the round-2 gap, now genuinely closed and running. Row 15 is not.

**Two disclosures on this table.** My first pass scored rows 3 and 15 as ICEs, and both were **no-results**: I had written `unmanaged[Cdecl](int32) int32` without the required `->`, and `*func` outside an `unsafe` context. Controls in ordinary type-annotation position returned `GS0005 Unexpected token <IdentifierToken>, expected <RightArrowToken>` and `GS0404 … requires an 'unsafe' context` respectively — my syntax was wrong, not the product. Re-probed with `unmanaged[Cdecl](int32) -> int32` (rc0 standalone) and `unsafe func Take(f *func(int32) int32)` (rc0 standalone), which proves both are legal `ParseTypeClause` productions, and only then re-drove them. Row 3 turned out clean; row 15 survived correction.

---

## BLOCKER — `*func(T) R` still reproduces #3255's exact signature at `file:line` `Parser.Expressions.Creation.cs:857`

`ParseTypeClauseCore:105` dispatches `StarToken` + `Peek(1)==FuncKeyword` to `ParseManagedFunctionPointerTypeClause` — a **distinct production** from the `*T` pointer at `:112`. `Parser.Expressions.Creation.cs:857-861` rejects the whole `StarToken` head with one `return false`, so the speculative scan fails, the parser falls back to an index expression, and the null key reaches the binder:

```
ptr_func_exact.gs(1,1,1,1): error GS9998: ArgumentNullException: Value cannot be null. (Parameter 'key')
```

on the head, byte-identical to base and byte-identical to issue #3255's reported signature.

Three things make this a blocker rather than untouched pre-existing behaviour:

1. **The new comment's justification is false for this production.** `:854-856` says accepting `*` "moves the failure from parsing/binding to a top-level `Cannot encode` internal error." For `*func` the *exclusion itself* produces a top-level internal error — the very outcome the comment claims to avoid.
2. **The PR's own test asserts a property this violates.** `Issue3099TypeArgumentReificationTests.cs` `PointerEnclosingTypeArgument_RemainsRejectedBeforeEmit` asserts `Assert.DoesNotContain("GS9998", result.Stdout)`. Substituting the other pointer production into that test's exact shape violates that assertion on this head. The test name states a general property about pointer enclosing type arguments; the assertion covers one of the two specimens, and the uncovered one exhibits precisely the failure the test asserts absent.
3. **`*int32` and `*func` diverge, so the exclusion is not uniformly safe.** `*int32` → rc1, three honest-span `GS0125`s, no ICE. `*func` → ICE. One `return false` covers two productions with two different outcomes.

Note the ICE is specific to the **enclosing** shape, which is exactly the region this PR owns: in non-enclosing position `Box[*func(int32) int32]()` gives rc1 `GS0125` + `GS0128`, no ICE.

**Recommended fix.** Do not widen the scanner blindly — the previous round already measured that accepting `*` routes `*T` to `Cannot encode '*'`. Let the type clause *parse* so it carries a real span, then reject at bind time with an anchored diagnostic ("pointer and function-pointer types cannot be generic type arguments"), which is what issue #3255 asked for: *"in which case it needs a real diagnostic anchored to the offending span, not a `GS9998` internal error at `(1,1,1,1)`."* Minimum acceptable alternative: extend the `:857` exclusion so `*func` reaches the same `GS0125` path as `*int32` rather than the ICE, and add the `*func` row to `PointerEnclosingTypeArgument_RemainsRejectedBeforeEmit`.

---

## 2. Is the deduplication real? Partly — **2 head-token lists remain**, not 1

Counted mechanically on both trees (clusters enumerating ≥4 of the 9 kinds within a 4-line window, across `Parser*.cs`):

| | base `d172d0d1` | head `92b287fa` |
|---|---|---|
| `CanStartTypeClause` — `Parser.Declarations.cs:925`, switch, 9 kinds | present | present (canonical) |
| `CurrentTokenStartsTypeClause` — `Parser.Expressions.Literals.cs:1079`, switch, 9 kinds | present | **removed** ✅ |
| `ParseOptionalTypeClause` — `Parser.TypeClauses.cs:864-872`, `!=` chain, 9 kinds | present | **still present** ❌ |
| inline partial set in `TryScanTypeClauseCore` (6 kinds — the #3255 defect) | present | **removed** ✅ |
| inline partial at the legacy-func return type (`Identifier`/`OpenParen` only) | present | **removed** ✅ |

So: **3 complete + 2 partial → 2 complete + 0 partial.** Real progress, and every *partial* copy — the kind that caused this bug — is gone. But the claim "reuses `CanStartTypeClause` as the single complete head-token set" is not accurate: `Parser.TypeClauses.cs:864-872` is a second complete copy, written as a `!=` chain rather than a switch, which is why a switch-shaped search misses it.

It is **pre-existing and untouched** by this PR (`Parser.TypeClauses.cs` is not in the diff) and it currently lists all 9, so there is no live divergence today. `Parser.Expressions.cs:572-584` also clusters 4 kinds but is `ParsePrimaryExpression`'s **expression** dispatch — a different grammar, correctly not a duplicate.

**SHOULD-FIX 1:** replace the `!=` chain at `Parser.TypeClauses.cs:864-872` with `if (!CanStartTypeClause(Current)) return null;`. One line, no behaviour change today, and it retires the last copy of the list whose drift caused #3255.

---

## SHOULD-FIX 2 — the `:1001` product hunk is not reached by any test in the corpus

`Parser.Expressions.Creation.cs:1001` changes the legacy-`func` return-type lookahead from `Identifier || OpenParen` to `CanStartTypeClause(...) && != StarToken`. Mutant **M7** (`CanStartTypeClause` → `!CanStartTypeClause`) built cleanly, moved `GSharp.Core.dll`, and left both target suites fully green.

Per doctrine I did not score that as a survival. I planted an unconditional `throw new InvalidOperationException("REACHED-1001")` immediately above the `if`, rebuilt (`BUILD_RC=0`), and ran:

- `Core.Tests` **full, unfiltered**: `Passed! Failed: 0, Passed: 7180, Skipped: 1, Total: 7181` — **0 hits**
- `Interpreter.Tests` **full, unfiltered**: `Passed! Failed: 0, Passed: 1347, Skipped: 3, Total: 1350` — **0 hits**

**8,531 tests, zero hits.** But the site is *not* dead: compiling `Box[Owner[func(int32) int32].Payload[string]]()` against the throw-planted build fires `REACHED-1001` immediately. So the hunk is reachable from ordinary G# source and simply unexercised — M7 is an uncovered-hunk finding, not a survival, and not a no-result either.

Add a case with a legacy `func(T) R` return type in enclosing-type-argument position. `func(int32) int32` runs rc0 → `System.Func\`2` on both trees, so it is a valid pin today; it is just unpinned.

---

## 3. `GS0125` — pre-existing, byte-identical, not a blocker on this PR

Measured on the merge-base before judging, using the PR's exact test source:

```
base d172d0d1 : (5,17,5,20) GS0125 Variable 'Box' doesn't exist.
                (5,21,5,26) GS0125 Variable 'Owner' doesn't exist.
                (5,28,5,33) GS0125 Variable 'int32' doesn't exist.   rc=1, stderr "Failed."
head 92b287fa : identical, all three, same spans, rc=1, stderr "Failed."
```

Byte-identical. Per the stated rule this is **pre-existing and reportable but not a blocker here**. The spans are honest — `(5,28,5,33)` really is `int32` — so the test's assertion is self-consistent and correctly anchored. The residual weakness is the *message class*: `int32` is a type, not a variable, so a user gets a misleading noun. That belongs in its own issue alongside #3266, not here.

For the record, the `(5,28,5,33)` diagnostic names **`int32`**, not `Box`; the `Box` diagnostic is `(5,17,5,20)`. The PR's test asserts the `int32` pairing and asserts it correctly.

---

## 4. Negative rows stayed negative

Five index-expression shapes most at risk from widened lookahead acceptance, base vs head, full stdout compared:

| probe | base | head | |
|---|---|---|---|
| map literal + `m["a"]` | rc0 | rc0 | IDENTICAL |
| func literal as index `arr[f(1)]` | rc0 | rc0 | IDENTICAL |
| nested index `a[b[1]]` | rc0 | rc0 | IDENTICAL |
| map indexed by variable | rc0 | rc0 | IDENTICAL |
| generic static call next to real indexing | rc0 | rc0 | IDENTICAL |

No widening regression observed. Rows 1–8 and 13–14 of the head-token table are also byte-identical across trees.

---

## 5. Mutation — per hunk, condition inversion, applied and compiled-in asserted

Pristine `Parser.Expressions.Creation.cs` md5 `39e3ebdebcea51b9a9352ecf0e800776` (matches the value in the PR body). Every mutant asserted applied via non-empty `git diff` **and** `GSharp.Core.dll` md5 movement; build rc checked so a StyleCop break would score as no-result, not survival.

| id | site | inversion | build | result |
|---|---|---|---|---|
| **M4 (control)** | `:850` entry gate | `!CanStartTypeClause` → `CanStartTypeClause` | rc0, dll moved | **RED** core 6/8, interp 17/18 |
| M1 | `:890` chan dispatch | `==` → `!=` | rc0, dll moved | **RED** 6/8, 17/18 |
| M2 | `:871` sequence dispatch | `==` → `!=` | rc0, dll moved | **RED** 6/8, 17/18 |
| M3 | `:884` map dispatch | `==` → `!=` | rc0, dll moved | **RED** 6/8, 17/18 |
| M5 | `:857` star exclusion | `==` → `!=` | rc0, dll moved | **RED** 6/8, 17/18 |
| M6 | `:876` async guard | `hasAsyncModifier` → `!hasAsyncModifier` | rc0, dll moved | **RED** 6/8, 17/18 |
| M7 | `:1001` legacy-func return type | `CanStartTypeClause` → `!CanStartTypeClause` | rc0, dll moved | **uncovered hunk** — see SHOULD-FIX 2 |

Control M4 killed as expected. Source restored by exact md5 round-trip (`39e3ebde…`), `git diff HEAD` empty, tree rebuilt `BUILD_RC=0`.

**Two-sided.** PR's two test files applied onto the **base** product, built `BUILD_RC=0` / 0 warnings:

- `Issue1323GenericStaticReceiverParserTests` → `Failed! Failed: 2, Passed: 6, Total: 8` — both new tests RED on `Assert.Empty(tree.Diagnostics)`.
- `Issue3099TypeArgumentReificationTests` → `Failed! Failed: 1, Passed: 17, Total: 18`; `CompositeEnclosingTypeArguments_ReifyAcrossEmittedHosts` fails with `driver failed with exit 1` and the exact `(1,1,1,1) GS9998 ArgumentNullException`.

So the positive test genuinely pins the fix. Note `PointerEnclosingTypeArgument_RemainsRejectedBeforeEmit` **passes on base** — it is a pure regression guard, not a two-sided pin, which is appropriate for a negative row but means it constrains nothing about this PR's delta. Base worktree restored; `git diff HEAD` empty.

The parser tests assert **spans**, not just ids — that satisfies the issue's explicit "assert the span as well as the id, or the anchoring half of this issue is untested." The runtime test compares full `FullName` strings across three emitted hosts, so it discriminates `Channel\`1[Int32]` from `object` rather than from a shared constant.

---

## NIT 1 — a malformed type clause in enclosing position ICEs instead of reporting a syntax error

`Box[Owner[unmanaged[Cdecl](int32) int32].Payload[string]]()` — the same text that yields a clean `GS0005 Unexpected token <IdentifierToken>, expected <RightArrowToken>` in ordinary type position — yields `GS9998 ArgumentNullException (1,1,1,1)` in enclosing position, byte-identical on base and head. Same root cause family: scan fails, fallback builds a malformed tree, null key. Pre-existing and out of this PR's scope, but it shows the scan-failure fallback converts *syntax errors* into ICEs generally, not only unsupported types — worth folding into whatever closes the BLOCKER above.

## NIT 2 — the PR body's decision text should name the production it excludes

`## Decision` says "Pointer types remain excluded … accepting `*T` here would move the existing pre-emit rejection into `GS9998: Cannot encode '*'`", and the matrix row is `*int32`. Both are about `*T`. `*func(T) R` is a separate `ParseTypeClauseCore` dispatch and behaves differently. Once the BLOCKER is addressed, the body and the matrix should state which pointer productions are covered.

Credit where due: the `## Span and #3192` section correctly identifies `src/Compiler/Program.cs:286` and `src/Repl/Program.cs:181` as the **shared** span cause and explicitly declines to change global ICE policy. That is the right call and correctly scoped to #3266. No null guard or fallback type was added — confirmed by reading the diff.

---

## Verification denominator

**9 test projects, source `dotnet sln GSharp.sln list`** (CI's own per `build/generate-ci-test-matrix.py:26`); the 10th solution entry, `test-assets/Issue3119.CrossConstants`, is an asset project, not a test project. Run against head `92b287fa`, unfiltered, no `--no-build`, no `GITHUB_ACTIONS`.

| project | rc | result |
|---|---:|---|
| Core.Tests | 0 | 7180 passed, 1 skipped, 7181 |
| Compiler.Tests | 0 | 4267 / 4267 |
| Cs2Gs.Tests | 0 | 1907 / 1907 |
| Interpreter.Tests | 0 | 1347 passed, 3 skipped, 1350 |
| LanguageServer.Tests | 0 | 468 / 468 |
| Extensions.Tests | 0 | 104 / 104 |
| Sdk.Tests | 0 | 59 / 59 |
| GSharp.GeneratorHost.Tests | 0 | 31 / 31 |
| InternalAnalyzers.Tests | 0 | 9 / 9 |

**15,372 passed, 4 skipped, 0 failed** — matching the PR body's figure exactly.

**ILVerify** (`build/run-ilverify.sh` after `dotnet tool restore`, rc0): `123/123 verified (2 with documented suppressions, 0 verification failures, 0 compile failures)`. Read as output, not rc: **zero** occurrences of `Assembly or module not found`, `IndexOutOfRange`, or `Unhandled exception` in the log.

---

## Summary

The `chan` / `sequence` / `map` / `async sequence` gap that sank round 2 is genuinely closed — measured running programs, correct reflected types, three emitted hosts, two-sided RED on base. Six of seven mutants killed with the control killing hardest, negative rows unchanged, whole suite and ILVerify clean.

It stops one production short. `*func(T) R` is a legal `ParseTypeClause` head shape and still reproduces #3255's exact `GS9998 ArgumentNullException (1,1,1,1)` on this head, while the PR's own pointer test asserts that signature is absent. Same root cause, same shape as the round-2 rejection, one level further down.

DavidObando:
Round-3 follow-up is addressed and pushed at `87c66182d`:

- scans the distinct `*func(T) R` enclosing type-argument production;
- rejects pointer/function-pointer generic arguments with anchored `GS0521` before emit;
- deduplicates the remaining type-clause head check;
- adds the missing legacy `func(T) R` and managed-function-pointer regression pins;
- rebased onto current `origin/main` (`132fbc200`).

Local verification: targeted Core **8/8**, targeted Interpreter **20/20**, full Interpreter **1459/1459**.

CI: 33 checks pass; `tests (compiler-issue1-remainder)` is the sole blocker. It fails `Issue1714StringZeroValueEmitTests.Reflection_ReferenceFields_UseClrNullAndPreserveAnnotationsAndInitializer` because `[]int32` now reflects as `[]` instead of null. I reproduced the same failure locally (541 pass, 1 fail), the PR diff does not touch that test or zero-value binding, and current-main PR #3314 has the same failing shard. No unrelated baseline fix was included.

Issue #2237, opened 2026-07-07
Assembly attribute support should have parity with C#


DavidObando:
In order to properly translate C# code to G# we need to support assembly-level attributes beyond the current hard-coded InternalsVisibleTo. For example, if I get a version of NBGV that produces G# code, it will look like this:

```gsharp
package MyApp
@assembly: System.Reflection.AssemblyVersionAttribute("1.3.0.0")
@assembly: System.Reflection.AssemblyFileVersionAttribute("1.3.1.0")
@assembly: System.Reflection.AssemblyInformationalVersionAttribute("")
@System.CodeDom.Compiler.GeneratedCode("Nerdbank.GitVersioning.Tasks","...")
@System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage
internal class ThisAssembly {
  shared {
    internal let AssemblyCompany string = "company"
    internal let AssemblyFileVersion string = "1.3.1.0"
    internal let AssemblyVersion string = "1.3.0.0"
    internal let CustomTicks System.DateTime = System.DateTime(637509805729817056L, System.DateTimeKind.Utc)
    internal let IsPrerelease bool = false
    internal let IsPublicRelease bool = false
    internal let RootNamespace string = "MyRootNamespace"
  }
}
```

Currently this doesn't work:
- The LSP complains that `@assembly` cannot be applied to a class, which is true, but it should also be smart enought to know an `@assembly` attribute applies to the assembly not to the class
   ```
   [{
   	"resource": "/private/tmp/nb/MyApp/ThisAssembly.gs",
   	"owner": "gsharp",
   	"code": "Semantic",
   	"severity": 8,
   	"message": "Attribute target 'assembly' is not valid on a class declaration.",
   	"source": "G#",
   	"startLineNumber": 2,
   	"startColumn": 2,
   	"endLineNumber": 2,
   	"endColumn": 10,
   	"modelVersionId": 12,
   	"origin": "extHost1"
   }]
   ```
- The compiler produces GS0464
  ```
   /private/tmp/nb/MyApp/ThisAssembly.gs(2,12,2,54): error GS0464: '@assembly:System.Reflection.AssemblyVersionAttribute' is not supported. The only file-level assembly annotation gsc emits today is '@assembly:InternalsVisibleTo("OtherAssemblyName")'.
  ```

This problem needs to be solved in a generalized way so that all the assembly attributes supported by C# become valid and work as expected in G#.

Issue #1172, opened 2026-06-25
Follow-up to #1160: arrow lambda gaps


DavidObando:
Ideally G# would deliver parity between the func literals and their arrow lambda representation.

PR #1160 modified cs2gs to default to func literals whenever C# presents a block lambda, meaning that the G# idiomatic arrow lambda is not being utilized as expected. This stems from arrow lambdas being value producing expressions, while func literals are explicitly typed, mapping to Action[T] or Func[T] as appropriate.

The opportunity here is for arrow lambdas in G# to gain the same capabilities as func literals whenever C# would have used an arrow lambda, and make the cs2gs conversion more idiomatic.

The gap between func literals and arrow lambdas requires a study and a proposal to get arrow lambdas to parity, with an ADR sustaining the proposal and the implementation.

Issue #611, opened 2026-06-09
Tech debt: audit other slice-vs-array surface gaps (follow-up to 6.5)


DavidObando:
## Context

PR 6.5 (the slice-conversions-equal-array-conversions PR, closing #570) added the `[]T → interface-of-T[]` implicit conversion. This closes the most common symptom (passing slices to BCL methods taking `IEnumerable<T>`, `IReadOnlyList<T>`, etc.), but there may be other places in the binder where `SliceTypeSymbol` and `ArrayTypeSymbol` diverge unexpectedly.

## Proposed audit

Run:
```bash
grep -rn 'ArrayTypeSymbol\|SliceTypeSymbol' src/Core/CodeAnalysis/Binding/
```

and inspect each site for potential gaps:

- **Type-parameter inference**: does the binder infer `T` when a `[]T` is passed where a `T[]` or `IEnumerable<T>` is expected?
- **Generic constraint satisfaction**: does `[]T` satisfy `where T : IEnumerable<U>` when the underlying array would?
- **Pattern matching on slices**: does `switch`/`case` on a slice value route correctly?
- **Overload ranking**: do `T[]` and `IList<T>` overloads rank identically for a slice argument as they would for an array argument?
- **Emit-side type encoding**: when a slice is stored in a field typed `IReadOnlyList<T>`, does the emitter encode the field type correctly?

## Scope

This is a tracking/audit issue. Each concrete gap found should be filed as its own bug.

Filed as follow-up to the 6.5 PR.

PR #3263, opened 2026-08-05
Correct emitted oracle test intent

DavidObando:
## Summary

- rename former interpreter-vs-emit tests to describe emitted goldens or emitted driver comparisons
- remove one literally duplicated Issue #2928 emit assertion
- preserve distinct gsc, gsi, explicit-output, oracle, and interactive emitted driver coverage

Refs #3257

## Validation

- targeted affected tests: 204 passed
- nine-project matrix: 15,368 passed, 4 skipped
- ILVerify: 123/123 verified, 2 documented suppressions
- solution build: 0 warnings, 0 errors

### DIFF
--- a/test/Compiler.Tests/Emit/Issue2928TupleFunctionEmitTests.cs
+++ b/test/Compiler.Tests/Emit/Issue2928TupleFunctionEmitTests.cs
@@ -1,4 +1,4 @@
-// <copyright file="Issue2928TupleFunctionParityTests.cs" company="GSharp">
+// <copyright file="Issue2928TupleFunctionEmitTests.cs" company="GSharp">
 // Copyright (C) GSharp Authors. All rights reserved.
 // </copyright>
 
@@ -14,29 +14,18 @@ namespace GSharp.Compiler.Tests.Emit;
 
 /// <summary>
 /// Issue #2928: tuple-contained function values execute correctly through the
-/// compiled backend. The interpreter parity arm was retired with the
-/// evaluator in ADR-0156 Phase 3c (#3176); its cross-checked output is pinned
-/// as the golden value.
+/// compiled backend, pinned to an explicit output value.
 /// </summary>
-public class Issue2928TupleFunctionParityTests
+public class Issue2928TupleFunctionEmitTests
 {
     private const string Source = """
-        package TupleFunctionInterpreter
+        package TupleFunctionEmit
         import System
 
         let handler (int32) -> int32 = (value int32) -> value + 1
         let t = (handler, 0)
         Console.WriteLine(t.Item1(41))
         """;
-
-    [Fact]
-    public void TupleFunction_EmittedOutputMatchesPinnedParityValue()
-    {
-        // "42\n" is the cross-checked interpreter/emit parity value, pinned
-        // as a literal golden when the evaluator retired in ADR-0156
-        // Phase 3c (#3176): t.Item1 invokes handler, which returns 41 + 1.
-        Assert.Equal("42\n", CompileAndRun(Source));
-    }
 
     [Fact]
     public void TupleFunction_CompiledBackendExecutes()
--- a/test/Core.Tests/CodeAnalysis/Emit/AsyncEmitGoldenTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Emit/AsyncEmitGoldenTests.cs
@@ -1,4 +1,4 @@
-// <copyright file="AsyncInterpVsEmitParityTests.cs" company="GSharp">
+// <copyright file="AsyncEmitGoldenTests.cs" company="GSharp">
 // Copyright (C) GSharp Authors. All rights reserved.
 // </copyright>
 
@@ -22,12 +22,12 @@ namespace GSharp.Core.Tests.CodeAnalysis.Emit;
 /// the tree-walking evaluator in ADR-0156 Phase 3c (#3176) — the goldens were
 /// the cross-checked parity values and are preserved verbatim.
 /// </summary>
-public class AsyncInterpVsEmitParityTests
+public class AsyncEmitGoldenTests
 {
     [Fact]
-    public void Parity_PureAsyncSequence_TaskFromResult()
+    public void Emit_PureAsyncSequence_TaskFromResult()
     {
-        const string Source = @"package ParityPure
+        const string Source = @"package AsyncEmitPure
 import System
 import System.Threading.Tasks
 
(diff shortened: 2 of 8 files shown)


DavidObando:
## Duck verdict — MERGEABLE. 0 BLOCKER / 2 SHOULD-FIX / 1 NIT

`main` `3e6dd50a7`, head `8b51079202`, merge-base **== `main`**, `merge-tree --write-tree` rc=0, **zero unmerged stages**. Clean fast-forward.

Test-only PR, so I judged it entirely on coverage preservation and ignored CI. A PR that weakens tests makes CI greener.

### Coverage census — base vs head, mechanical

| file | `Assert.` b→h | `[Fact]`/`[Theory]` b→h | `[InlineData]`/`[MemberData]` b→h |
|---|---|---|---|
| `Issue2928TupleFunctionEmitTests.cs` | 5→**4** | 3→**2** | 0→0 |
| `AsyncEmitGoldenTests.cs` | 5→5 | 12→12 | 0→0 |
| `Issue2896StructObjectOverrideTests.cs` | 27→27 | 8→8 | 34→34 |
| `Issue3015ImportedBaseIdentityTests.cs` | 21→21 | 12→12 | 0→0 |
| `Issue3099TypeArgumentReificationTests.cs` | 17→17 | 1→1 | 1→1 |
| `Issue3100FieldOffsetDriverTests.cs` | 27→27 | 7→7 | 5→5 |
| `Issue3140OutParameterWriteBackDriverTests.cs` | 12→12 | 2→2 | 2→2 |

**Exactly one assertion and one `[Fact]` removed repo-wide.** Everything else is identifier/doc renaming.

### The one deletion is genuinely vacuous — proved two ways

At base, `Issue2928TupleFunctionParityTests.cs:32-42` and `:44-48` had **byte-identical bodies**:

```csharp
Assert.Equal("42\n", CompileAndRun(Source));   // both, same Source const, same helper
```

Body identity is a stronger proof than a mutant: no mutation can separate them. This is exactly the "both sides are literally the same call" case #3257 authorised.

I still checked the survivor is **live**, not a pair of dead tests:

| mutant | site | Issue2928 result |
|---|---|---|
| M1 swap target/args, fallback branch | `MethodBodyEmitter.Calls.cs:423-424` | **SURVIVED** |
| M2 same swap, **all three** dispatch branches | `MethodBodyEmitter.Calls.cs:400/415/423` | **SURVIVED** |
| **PC positive control** — reachable `throw` at top of `EmitIndirectCall` | `MethodBodyEmitter.Calls.cs:385` | **SURVIVED** |
| M3 `ILOpCode.Add` → `ILOpCode.Sub` | `MethodBodyEmitter.Operators.cs:967` | **RED 1/2** |

**M1 and M2 were no-results, not survivals.** The positive control proves `EmitIndirectCall` is never reached by `t.Item1(41)` — the program takes another path entirely. Reporting M1/M2 as "the test can't detect this" would have been wrong. Posting them here because the site-reachability control is the only reason I know that.

### Independent spot-checks — kept tests still go RED (M3)

| class | unmutated | under M3 |
|---|---|---|
| `AsyncEmitGoldenTests` | 12/12 pass | **5 RED** |
| `Issue3140OutParameterWriteBackDriverTests` | 90/90 pass | **20 RED** |
| `Issue2928TupleFunctionEmitTests` | 2/2 pass | **1 RED** |

Round-trip: mutants reverted, `git diff HEAD` = 0 lines, `git status --porcelain` = 0 entries. Unmutated touched-class total **204 passed** — matches the PR body's 204 exactly.

### Point 4 respected — the driver columns kept are genuinely different drivers

`Issue3140OutParameterWriteBackDriverTests.cs:341-352` compares `RunEmittedAsync` (gsc `/out: /target:exe` → **separate `dotnet exec` OS process**, real CLR JIT, `:371-418`) against `RunEmittedOracle` (in-process `EmittedOracle.Evaluate`, `:359-368`). Two real drivers. Correctly kept, correctly renamed. Same for `Issue3100` (bare `gsc` / `gsc /out:` / `gsi`) and `Issue3015`/`Issue3099` (up to four entry points).

No dangling references to any renamed symbol; positive control on `AsyncEmitGoldenTests` fires in 2 files.

---

## SHOULD-FIX 1 — the same defect is left in four files, two with the identical suffix this PR renamed

The PR renamed `Issue3134_ClassIdentityAndStructValueEqualityMatchEmitter` → `...AgreeAcrossEmittedDrivers` (`Issue2896StructObjectOverrideTests.cs:127`). It then left `MatchesEmitter` in place three files away:

- **`test/Interpreter.Tests/PrimitiveInterpreterTests.cs:59`** — `Interpreter_PrimitiveArithmetic_MatchesEmitter`. Body is `new GSharpRepl().EvaluateSubmission(text)` + `Assert.Contains(hardcoded)`. **One column, no emitter.** Class doc `:11-15` claims *"asserts that the printed value matches what the emitter produces in the equivalent compiled program."* Comment `:55` still says *"in the tree interpreter."*
- **`test/Interpreter.Tests/ParenthesizedReceiverInterpreterTests.cs:24`** — same shape, and `:13` states *"the interpreter (**a separate execution path from the emitter**)"*. That separate path no longer exists — this is the exact sentence #3257 targets.
- **`test/Core.Tests/.../Issue1881CheckedUncheckedInterpreterTests.cs:19`** and **`Issue2023CheckedUnaryNegationInterpreterTests.cs:19`** — both: *"Asserts the tree-walking evaluator agrees with the compiled/emitted IL."* Single-column `Evaluate(...)` against literals.

Not a correctness problem, but this PR's entire value is that names stop lying. Four survivors in the same sweep is a real gap. `*Parity*`-named files are clean — I checked all five remaining (`#611` slice-vs-array, `#1172` ×2 lambda-vs-func-literal, `#2237` vs C# `[assembly:]`, LSP contract); none is engine parity.

## SHOULD-FIX 2 — no mutation table in the PR body

The body has none at all. The "literally duplicated" vacuity claim is asserted, not demonstrated. It happens to be **true and I verified it**, but M1/M2 above show how easily a mutation lands on an unreached site and reads as a survival. A table with the site cited would have made this reviewable.

## NIT — name what permanently lost its differential oracle

Pre-existing Phase 3c damage, not caused here, but the PR should record it rather than let an accurate filename paper over it:

- **`AsyncEmitGoldenTests` (12 tests)** — async/await lowering, `Task.Delay` suspension, multiple/nested awaits in `try`, try/finally run-once (**#137 regression**), try/catch and try/finally around await, async accumulator, sync iterator, async iterator yield+await, top-level `await for`, `go`+`scope` async target. Now emit vs hardcoded goldens.
- **`Issue2928TupleFunctionEmitTests`** — tuple-contained function values. Now emit vs `"42\n"`.
- **`Issue2896StructObjectOverrideTests:79, :292, :330`** — struct/class `object` overrides, `shared`-block overrides, default-value rendering. Now `EmittedOracle` vs hardcoded strings.

Everything else in the eight files retains ≥2 real drivers.

**Verdict: MERGEABLE.** No coverage lost. Both SHOULD-FIXes are follow-up-able.

PR #3264, opened 2026-08-05
Fix enclosing generic closure emission

### LINKED ISSUE #3254: [gsc] Emit fails at runtime for enclosing generic type parameters: BadImageFormatException on lambda capture, TypeLoadException on 'go'
## Summary

Two emit shapes involving an **enclosing generic type parameter** produce runtime type-load failures rather than working code. Both were found while rubber-ducking PR #3229, and both reproduced **identically on that PR's base and head**, so they are pre-existing on `main` and unrelated to it.

They are filed now because **#3244 (ADR-0156 Phase 3c) just deleted the tree-walking evaluator**. The reviewer explicitly noted that *the interpreter rendered both of these correctly while emit failed* — the reverse of the usual direction. That fallback is now gone, so these are user-visible with no second engine to fall back on.

## Shape 1 — lambda capturing an enclosing type parameter

A lambda that captures an enclosing generic type parameter emits an assembly that fails to load:

```
System.BadImageFormatException
```

## Shape 2 — `go` statement inside a generic

A `go` statement inside a generic context emits a closure type the runtime cannot load:

```
System.TypeLoadException: Could not load type 'P.<go_1>`1'
```

The mangled name `<go_1>` with arity 1 suggests the `go`-statement closure type is emitted with a generic arity whose type parameters are not correctly declared or matched at the use site.

## Why these may share a root cause

Both are enclosing-generic-parameter propagation into a **compiler-synthesized type** (a lambda closure and a `go` closure). It is plausible that one fix addresses both. **Confirm or refute that by measurement** — if the root causes are distinct, say so and split this issue; do not force a single fix.

## Required evidence

- **`ilverify` is not sufficient here and must not be treated as proof.** It has been caught this session returning rc=0 while printing `Assembly or module not found: mscorlib`, and separately green-lighting a program the CLR refuses to JIT (#3240). **Run the emitted exe and assert its exit code and stdout.**
- `gsc` writes diagnostics to **stdout** and only `Failed.` to stderr — **capture `rc` before any pipe**.
- Pin both shapes with tests that **fail for the right reason** before the fix: apply the pin alone to the unfixed tree, rebuild, and show the actual `BadImageFormatException` / `TypeLoadException` text, not merely a red line.
- The evaluator is **deleted** as of `8b3bdac57`. There is no interp-vs-emit oracle any more; the oracle is the **running program**.

## Provenance

Reported by the PR #3229 reviewer, which noted these while probing enclosing-generic reification and correctly reported them as out of that PR's scope rather than folding them in.

DavidObando:
## Summary

The two failures had distinct root causes:

- Nested user struct captures only scanned the struct's own type arguments. They now also collect `EnclosingTypeArguments`, so lambda display classes reify enclosing generic parameters.
- `go` display classes were synthesized as generic but their constructor, capture fields, and invoke method were referenced through bare definition handles. They now use member tokens parented by the constructed closure TypeSpec, matching ordinary lambda emission.

## Regression coverage

Added emitted-executable tests for:

- a lambda capturing `Owner[T].Payload[string]`
- synchronous generic `go`
- asynchronous generic `go` (`Func<Task>` path)

Each runs both `T = int32` and `T = string` and asserts exact output `42\nright\n`, process exit 0, and ILVerify.

Tests-only on the unfixed tree reproduced the real runtime failures:

- `BadImageFormatException: An attempt was made to load a program with an incorrect format.`
- `TypeLoadException: Could not load type 'Issue3254.Go.<go_1>\`1'`
- the async shape produced the corresponding `Issue3254.AsyncGo.<go_1>\`1` TypeLoadException

Targeted branch inversions and a bare async invoke-token mutant each made their intended runtime test fail while control rows stayed green.

## Validation

- Release nine-project suite: 15,372 passed, 4 skipped
- ILVerify gate: 123/123
- `dotnet build -c Release`: 0 warnings, 0 errors

Fixes #3254

### DIFF
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Statements.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Statements.cs
@@ -410,13 +410,27 @@ internal sealed partial class MethodBodyEmitter
                 $"Go display invoke method '{closure.InvokeMethod.Name}' has no emitted MethodDef.");
         }
 
+        var constructedClosure = closure.ConstructedClassSym;
+        var closureIsGeneric = ReflectionMetadataEmitter.IsUserGenericTypeReference(constructedClosure);
+
         this.il.OpCode(ILOpCode.Newobj);
-        this.il.Token(ctorHandle);
+        this.il.Token(closureIsGeneric
+            ? this.outer.userTokens.ResolveUserCtorTokenForDefault(constructedClosure)
+            : ctorHandle);
 
         foreach (var captured in closure.CaptureFields.Keys)
         {
             var field = closure.CaptureFields[captured];
-            if (!this.outer.cache.StructFieldDefs.TryGetValue(field, out var fieldHandle))
+            EntityHandle fieldToken;
+            if (closureIsGeneric)
+            {
+                fieldToken = this.outer.userTokens.ResolveFieldToken(constructedClosure, field);
+            }
+            else if (this.outer.cache.StructFieldDefs.TryGetValue(field, out var fieldHandle))
+            {
+                fieldToken = fieldHandle;
+            }
+            else
             {
                 throw new InvalidOperationException(
                     $"Go display field '{field.Name}' has no emitted FieldDef.");
@@ -425,16 +439,19 @@ internal sealed partial class MethodBodyEmitter
             this.il.OpCode(ILOpCode.Dup);
             this.EmitExpression(new BoundVariableExpression(null, captured));
             this.il.OpCode(ILOpCode.Stfld);
-            this.il.Token(fieldHandle);
+            this.il.Token(fieldToken);
         }
 
         var isAsync = ClosureEmitter.IsTaskClrType(closure.InvokeMethod.Type?.ClrType);
+        var invokeToken = closureIsGeneric
+            ? this.outer.userTokens.ResolveClosureInvokeFtnToken(constructedClosure, closure.ClassSym, closure.InvokeMethod)
+            : invokeHandle;
 
         if (isAsync)
         {
             var funcTaskCtor = typeof(Func<System.Threading.Tasks.Task>).GetConstructor(new[] { typeof(object), typeof(nint) });
             this.il.OpCode(ILOpCode.Ldftn);
-            this.il.Token(invokeHandle);
+            this.il.Token(invokeToken);
             this.il.OpCode(ILOpCode.Newobj);
             this.il.Token(this.outer.memberRefs.GetCtorReference(funcTaskCtor));
         }
@@ -442,7 +459,7 @@ internal sealed partial class MethodBodyEmitter
         {
             var actionCtor = typeof(Action).GetConstructor(new[] { typeof(object), typeof(nint) });
             this.il.OpCode(ILOpCode.Ldftn);
-            this.il.Token(invokeHandle);
+            this.il.Token(invokeToken);
             this.il.OpCode(ILOpCode.Newobj);
             this.il.Token(this.outer.memberRefs.GetCtorReference(actionCtor));
         }
--- a/src/Core/CodeAnalysis/Symbols/StructSymbol.cs
+++ b/src/Core/CodeAnalysis/Symbols/StructSymbol.cs
@@ -1501,14 +1501,15 @@ public sealed class StructSymbol : TypeSymbol
         Dictionary<TypeParameterSymbol, TypeSymbol> running = null;
         for (var c = this; c != null; c = c.BaseClass)
         {
-            // Issue #1521: a constructed reference to a type nested inside a
-            // generic enclosing type (`Box[int32].Tag`) already carries the
-            // enclosing arguments on EnclosingTypeArguments and declares no own
-            // type arguments to compose. When such a receiver directly declares
-            // the sought member, return it verbatim so the field/method
-            // reference is parented at its own `Box`1+Tag`1<int32>` TypeSpec
-            // rather than a rebuilt open self-instantiation.
-            if (c.IsConstructedNestedType && declaringDefinitionPredicate(c.Definition ?? c))
+            // A constructed nested receiver already carries the argument shape
+            // needed by its TypeSpec. Keep it verbatim: rebuilding a generic
+            // nested type whose open enclosing arguments are implicit would
+            // mistake its own arguments for the flattened enclosing slots.
+            var isConstructedNested = c.IsConstructedNestedType
+                || (c.Definition != null
+                    && !c.TypeArguments.IsDefaultOrEmpty
+                    && !CollectEnclosingTypeParameters(c).IsDefaultOrEmpty);
+            if (isConstructedNested && declaringDefinitionPredicate(c.Definition ?? c))
             {
                 return c;
             }
@@ -1748,7 +1749,7 @@ public sealed class StructSymbol : TypeSymbol
 
     private static TypeSymbol EnclosingTypeOf(TypeSymbol type) => type switch
     {
-        StructSymbol s => s.ContainingType,
+        StructSymbol s => (s.Definition ?? s).ContainingType,
         InterfaceSymbol i => i.ContainingType,
         EnumSymbol e => e.ContainingType,
         _ => null,
--- a/src/Core/CodeAnalysis/Symbols/TypeSymbol.cs
+++ b/src/Core/CodeAnalysis/Symbols/TypeSymbol.cs
@@ -555,41 +555,33 @@ public class TypeSymbol : Symbol
                 }
 
                 return;
-            case StructSymbol ssNested when !ssNested.EnclosingTypeArguments.IsDefaultOrEmpty:
-                foreach (var arg in ssNested.EnclosingTypeArguments)
+            case StructSymbol ss:
+                if (!ss.EnclosingTypeArguments.IsDefaultOrEmpty)
                 {
-                    CollectReferencedTypeParameters(arg, sink);
+                    foreach (var arg in ss.EnclosingTypeArguments)
+                    {
+                        CollectReferencedTypeParameters(arg, sink);
+                    }
                 }
-
-                foreach (var arg in ssNested.TypeArguments)
+                else
                 {
-                    CollectReferencedTypeParameters(arg, sink);
-                }
-
-                if (ssNested.TypeArguments.IsDefaultOrEmpty)
-                {
-                    foreach (var tp in ssNested.TypeParameters)
+                    foreach (var tp in StructSymbol.CollectEnclosingTypeParameters(ss))
                     {
                         CollectReferencedTypeParameters(tp, sink);
                     }
                 }
 
-                return;
-            case StructSymbol ss when !ss.TypeArguments.IsDefaultOrEmpty:
                 foreach (var arg in ss.TypeArguments)
                 {
                     CollectReferencedTypeParameters(arg, sink);
                 }
 
-                return;
-            case StructSymbol ssOpen when !ssOpen.TypeParameters.IsDefaultOrEmpty:
-                // Issue #1477: a captured `this` of a generic type `G[T]` is the
-                // OPEN definition (TypeParameters set, no TypeArguments) — its
-                // own parameters ARE the enclosing parameters the capture field
-                // references (`G`1<!0>`), so collect them.
-                foreach (var tp in ssOpen.TypeParameters)
+                if (ss.TypeArguments.IsDefaultOrEmpty)
                 {
-                    CollectReferencedTypeParameters(tp, sink);
+                    foreach (var tp in ss.TypeParameters)
+                    {
+                        CollectReferencedTypeParameters(tp, sink);
+                    }
                 }
 
                 return;
--- a/src/Core/CodeAnalysis/Symbols/TypeSymbol.cs
+++ b/src/Core/CodeAnalysis/Symbols/TypeSymbol.cs
@@ -555,6 +555,26 @@ public class TypeSymbol : Symbol
                 }
 
                 return;
+            case StructSymbol ssNested when !ssNested.EnclosingTypeArguments.IsDefaultOrEmpty:
+                foreach (var arg in ssNested.EnclosingTypeArguments)
+                {
+                    CollectReferencedTypeParameters(arg, sink);
+                }
+
+                foreach (var arg in ssNested.TypeArguments)
+                {
+                    CollectReferencedTypeParameters(arg, sink);
+                }
+
+                if (ssNested.TypeArguments.IsDefaultOrEmpty)
+                {
+                    foreach (var tp in ssNested.TypeParameters)
+                    {
+                        CollectReferencedTypeParameters(tp, sink);
+                    }
+                }
+
+                return;
             case StructSymbol ss when !ss.TypeArguments.IsDefaultOrEmpty:
                 foreach (var arg in ss.TypeArguments)
                 {
--- a/test/Compiler.Tests/Emit/Issue3254EnclosingGenericClosureEmitTests.cs
+++ b/test/Compiler.Tests/Emit/Issue3254EnclosingGenericClosureEmitTests.cs
@@ -1,0 +1,179 @@
+// <copyright file="Issue3254EnclosingGenericClosureEmitTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using GSharp.Compiler;
+using Xunit;
+
+namespace GSharp.Compiler.Tests.Emit;
+
+/// <summary>
+/// Issue #3254: synthesized lambda and <c>go</c> closure types must preserve
+/// enclosing generic type parameters at their definitions and use sites.
+/// </summary>
+public class Issue3254EnclosingGenericClosureEmitTests
+{
+    [Fact]
+    public void LambdaCapturingNestedTypeWithEnclosingTypeParameterRuns()
+    {
+        const string Source = """
+            package Issue3254.Lambda
+            import System
+
+            class Owner[T] {
+                struct Payload[U] {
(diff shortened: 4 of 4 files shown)


DavidObando:
## Review — 1 BLOCKER, 1 SHOULD-FIX, 1 NIT

Verdict is against **`main` = `e28d55bd`** (fetched at post time; the briefed `d172d0d1` went stale mid-review — `main` moved twice: `d172d0d14` #3263, `e28d55bdf` #3261). `git merge-tree --write-tree origin/main c166f7b03` → **1 line, tree oid only**, no stages, against both `d172d0d1` and `e28d55bd`.

Reviewed at head `c166f7b03`, own worktree, Release `-warnaserror`, 38 project outputs, 0 warnings.

---

### First: the "distinct root causes" claim is TRUE, and I measured it rather than accepting it

Built the full 2×2 by reverting each hunk independently and rebuilding each time (`Core.dll` md5 moved every time; `git diff HEAD` asserted each revert applied):

| config | `Core.dll` md5 | shape 1 (lambda) | shape 2 (`go`) |
|---|---|---|---|
| neither | `97b88ecd` | **BadImageFormatException** rc=134 | **TypeLoadException `<go_1>\`1`** rc=134 |
| **A only** (`TypeSymbol.cs`) | `28eb6030` | **42 / right rc=0** | TypeLoadException rc=134 |
| **B only** (`MethodBodyEmitter.Statements.cs`) | `dccd30cb` | BadImageFormatException rc=134 | **42 / right rc=0** |
| both (head) | `8a4d857a` | 42 / right rc=0 | 42 / right rc=0 |

**A perfect diagonal.** Each hunk fixes exactly one shape and neither fixes the other — distinct root causes **confirmed by measurement**, no shared cause was missed, and each hunk is independently load-bearing with zero collateral. Restored by exact md5 round-trip to `8a4d857a` (twice).

Note both failing configs **compile rc=0** — the compiler silently emits an unloadable assembly.

**Author's pins fail for the right reason.** Ran the 3 new tests against the unfixed tree: `RC=1, Failed 3 / Passed 0`, and the failure texts are `BadImageFormatException` ×1 and `TypeLoadException: Could not load type '…<go_1>\`1'` ×2 (`Issue3254.Go` and `Issue3254.AsyncGo`) — the real exceptions, not merely red lines. Green 3/3 on head.

**Sibling audit (bare member tokens).** Exactly **three** closure-construction sites consume `ConstructedClassSym`: `Closures.cs:359`, `Closures.cs:777`, `Statements.cs:413`. The first two **already carried the identical generic-aware pattern**; `go` was the only one missing it, so this PR completes the set rather than inventing an idiom. The one asymmetry I found — `MethodBodyEmitter.MemberAccess.cs:141` `TryStoreToEnclosingClosure` uses a bare `StructFieldDefs` FieldDef where the read at `:131` uses `ResolveFieldToken(ConstructedClassSym, …)` — I probed with a store-to-capture inside a generic closure and it **runs correctly** (`42`/`right`); a FieldDef self-reference inside a generic type's own method is legal IL. **Not a finding.** Also checked `TypeArguments`/`TypeParameters` both default to `ImmutableArray.Empty` (`StructSymbol.cs:501,504`), so the new arm's unguarded `foreach` cannot NRE.

---

## BLOCKER — #3254 shape 1 still reproduces on this head, one syntactic step away

`src/Core/CodeAnalysis/Symbols/TypeSymbol.cs:557`

The new arm is keyed on **`!ssNested.EnclosingTypeArguments.IsDefaultOrEmpty`**. That array is populated only when the nested type is named **qualified from outside** (`Owner[T].Payload[…]`). When the *same* nested type is named **unqualified from inside its own enclosing generic type**, `EnclosingTypeArguments` is empty, the new arm never fires, control falls to the pre-existing `case StructSymbol ss when !ss.TypeArguments.IsDefaultOrEmpty` which collects own args only — and `T` is dropped exactly as before.

```gs
class Owner[T] {
    struct Payload[U] { var First T
                        var Second U }
    func Make(p Payload[string]) T {
        let read = () -> p          // lambda capturing an enclosing type parameter
        return read().First } }
func Main() { Console.WriteLine(Owner[int32]{}.Make(Owner[int32].Payload[string]{ First: 42, Second: "s" })) }
```
On head `c166f7b03`, fixed binary `8a4d857a`:
```
compile rc=0 | RUN rc=134
Unhandled exception. System.BadImageFormatException: An attempt was made to load a program with an incorrect format. (0x8007000B)
   at P.Owner`1.Make(Payload`1 p)
```
That is the issue's headline sentence verbatim — "a lambda that captures an enclosing generic type parameter emits an assembly that fails to load".

**Boundary, measured:**

| nested type named as | `EnclosingTypeArguments` | head |
|---|---|---|
| `Owner[T].Payload[string]` (qualified, free func) | populated | **fixed** — 42 / right |
| `Owner[T].Payload` (qualified, free func) | populated | **fixed** — 42 |
| `Payload[string]` (unqualified, inside `Owner[T]`) | empty | **BadImageFormatException** |
| `Payload` (unqualified, inside `Owner[T]`) | empty | **BadImageFormatException** |

**Positive control:** a lambda capturing a *plain* `T` in the same position (`class Owner[T] { var Value T; func Make() T { let v = this.Value; let read = () -> v; return read() } }`) prints `42`/`right` rc=0. So this is specific to the nested-type-typed capture, not lambdas-in-generic-classes generally — and my harness is not simply failing at everything.

**Not a regression:** the added arm's guard is false on this path, so the switch behaves identically to base — by construction over the changed lines, not by probe.

**Corroboration of the mechanism:** `ilspycmd` shows the display type emitted **nested inside the generic** as `P.Owner.<closure_<lambda1>_1>`, and `ilverify` on that assembly does not report a violation — it **crashes**: `IndexOutOfRangeException at Internal.TypeSystem.Instantiation.get_GenericParameters(Int32 index)`. The IL references a generic parameter ordinal the closure never declared. (Also note CI's `ilverify` job only covers *sample* assemblies, so it cannot see this at all.)

**I built the obvious remedy and it is insufficient — this is not a one-line guard widening.** Candidate: add an arm keyed on `ContainingType` (available and propagated — `StructSymbol.cs:659` ADR-0110/#910, copied at `:1866`) collecting the containing type's `TypeParameters` when `EnclosingTypeArguments` is empty. Built it (`Core.dll` → `9e337737`):
- `Payload` (non-generic nested) → **fixed**, prints 42
- `Payload[string]` (generic nested) → **still BadImageFormatException**
- swap control → still green

So the generic-nested-from-inside case needs more than type-parameter collection — likely the enclosing+own ordinal flattening (`nestedOwnTypeParameters` / `RegisterNestedTypeEnclosingGenerics`). Reverted; exact md5 round-trip to `8a4d857a`.

---

## SHOULD-FIX — the identity pins cannot detect erasure

All three tests assert only `"42\nright\n"`. If the fix had made the closure load by **erasing `T` to `object`**, `Console.WriteLine(object)` prints `42` and `right` **identically** — the pins would stay green on a silent wrong answer. That is precisely the dangerous failure mode for this defect class, and the current assertions do not exclude it.

I attacked this from five directions and **found no actual defect** — the fix genuinely preserves `T`:
- `[]T{…}.GetType()` inside the lambda closure → `System.Int32[]` / `System.String[]` (not `System.Object[]`)
- same inside the `go` closure → `System.Int32[]` / `System.String[]`
- `go` inside a **generic class** method → `System.Int32[]` / `System.String[]`
- two distinct reference types (`string`, user `class Thing`) → `System.String[]`, `P.Thing[]`
- **mirror-image swap detector**, the strongest one: `Run[T, U](Owner[T].Payload[U])` instantiated `[int32,string]` and `[string,int32]` → `System.Int32[] / System.String[]` and `System.String[] / System.Int32[]`. A transposed enclosing/own argument order would still load and print plausible output; it does not. **Order is correct.**

I could not construct a wrong-`T`-but-green case. But that is *my* evidence, not the suite's — please fold the discriminating assertions in, since they are nearly free: change the probes to print `[]T{…}.GetType()` alongside the value, and add the `[int32,string]`/`[string,int32]` mirror row. That converts "the type loads" into "the type loads with the right `T`", which is the property actually at risk.

---

## NIT — pre-existing, not yours, worth its own issue

`Run[object](42)` — a generic instantiated at `object` with an implicitly boxed value — gives **`compile rc=0` then `InvalidProgramException` rc=134**. Same defect class as #3254 (compiler accepts, CLR refuses). **Not caused by this PR**: reproduces with both hunks reverted (`Core.dll` `97b88ecd`), and reproduces with **no `go` and no lambda at all**, so it is not closure-related. Isolated by bisecting a 4-row probe; the other three rows (`int32`, `string`, user class) all pass.

---

### Coverage

**9 of 9 suites, `--no-build` against the verified `8a4d857a` binary, all RC=0: 15,372 passed, 0 failed, 4 skipped.**
Compiler.Tests 4271 (29m) · Core.Tests 7178/1 skip · Cs2Gs.Tests 1907 · Interpreter.Tests 1345/3 skip · LanguageServer.Tests 468 · Extensions.Tests 104 · Sdk.Tests 59 · GeneratorHost.Tests 31 · InternalAnalyzers.Tests 9. Independently matches your reported 15,372/4 — **no regression from either hunk.**

Worktree left porcelain-0; every mutation restored by exact `Core.dll` md5 round-trip.

DavidObando:
## Round 2 — my BLOCKER is **fixed**. No new BLOCKER. One NIT.

Against **`main` = `6b809619`** (re-fetched at post time), head **`2418ab586`**, `mergeable=MERGEABLE state=CLEAN`. `git merge-tree --write-tree origin/main 2418ab586` → **1 line, tree oid only** — no stages.

Fresh worktrees, Release `-warnaserror`, 0 warnings. **Every round-1 number discarded and re-measured**, as instructed.

---

### 1. The BLOCKER is genuinely closed

My round-1 escape — a nested type named **unqualified from inside its own enclosing generic type**, whose `EnclosingTypeArguments` is empty — now works:

| probe | round-1 head `c166f7b03` | head `2418ab586` |
|---|---|---|
| `Payload[string]` unqualified inside `Owner[T]` | **BadImageFormatException** rc=134 | **rc=0 → `42` / `right`** |
| `Payload` (non-generic) unqualified inside `Owner[T]` | **BadImageFormatException** rc=134 | **rc=0 → `42`** |

And it is fixed the *hard* way. In round 1 I built the obvious `ContainingType` guard-widening and reported it **insufficient** — it fixed the non-generic nested case but not the generic one. This PR did not take that shortcut: the third hunk (`StructSymbol.cs`) is precisely what makes the generic-nested half work. Per-hunk mutants prove it (below).

### 2. Ordinal flattening is correct in the general case, not just for the specimens

Every row is compile → `dotnet exec` → assert **rc and stdout**, with each type parameter made **runtime-distinguishable** via `[]T{…}.GetType()` so erasure or transposition cannot hide. Each row is run in **mirror-image pairs** — the same program with the type arguments swapped — because a transposition still emits, still verifies, and would otherwise be silent.

| shape | head `2418ab586` | unfixed tree |
|---|---|---|
| enclosing arity 2, `Owner[TA,TB].Payload[U]` | `TA=Int32[]/42 TB=String[]/right U=Boolean[]/True` ⇄ `TA=String[]/right TB=Int32[]/42` | **BadImageFormat** |
| two enclosing levels `L1[A].L2[B].P3[C]` | `A=Int32[]/42 B=String[]/right C=Boolean[]` ⇄ mirrored | **BadImageFormat** |
| **three** enclosing levels `A1[X].B2[Y].C3[Z].P4[W]` | `X=Int32[]/42 Y=String[]/right Z=Int64[]/9 W=Boolean[]` ⇄ **3-way rotated** correctly | — |
| generic **method** on generic type, closure captures enclosing `T` **and** own `M` | `T=Int32[]/42 M=String[]/right` ⇄ mirrored | (already worked) |
| **union**: arity-2 enclosing + nested own `U` + method own `M`, one closure | `TA=Int32[]/42 TB=String[]/right U=Boolean[] M=Int64[]/7` ⇄ mirrored, `M=Double[]/2.5` | **BadImageFormat** |
| zero own params inside generic enclosing | `42` / `right` | **BadImageFormat** |
| `go` inside arity-2 generic capturing nested-typed value | mirrored correctly | **TypeLoadException** |
| `go` inside generic **method** on generic type | mirrored correctly | **TypeLoadException** |
| generic **struct** as enclosing type | correct | — |
| enum nested in generic, captured | correct | — |
| interface nested in generic, captured | correct | — |

The three-level row is the strongest: a **3-way rotation** (`int32,string,int64` → `string,int64,int32`) is not detectable by a 2-way swap test, and it resolves correctly. I could not construct a transposition that survives.

### 3. Distinct root causes — now **three**, all independently load-bearing

Per-hunk reverts, each rebuilt (`GSharp.Core.dll` md5 moved every time; `git diff HEAD` asserted applied):

| probe | −`TypeSymbol.cs` | −`MethodBodyEmitter.Statements.cs` | −`StructSymbol.cs` | head |
|---|---|---|---|---|
| `inner` / arity-2 / 2-level / union | **BadImage** | ok | **BadImage** | ok |
| `inner2` (non-generic nested) | **BadImage** | ok | ok | ok |
| `go` w/ nested capture | ok | **TypeLoad** | **BadImage** | ok |
| `go` plain `T` | ok | **TypeLoad** | ok | ok |
| pure member resolution | ok | ok | ok | ok |

**All three hunks KILLED, no survivals** — so no positive-control `throw` was required; nothing was scored as surviving. The table also shows `StructSymbol.cs` is not incidental: it is exactly what carries the *generic*-nested case.

I probed the `StructSymbol.cs:1501` blast radius separately, since that hunk sits in general member resolution (`FindConstructedGenericBase`) rather than closure emission: field reads, methods, and inside/outside access on constructed nested generics, in mirror pairs — all correct, and unchanged by that hunk.

### 4. RED-proof of the four pins

Reverted all three product hunks, rebuilt (RC=0), ran the pins: **`Failed: 4, Passed: 0`**, and failing for the right reason — **4 × `dotnet exited 134`, zero `gsc exited N`**, i.e. compile succeeded and the *program* failed to load:

- 2 × `System.BadImageFormatException` (both lambda pins, including the new one for my BLOCKER)
- `System.TypeLoadException: Could not load type 'Issue3254.Go.<go_1>\`1'`
- `System.TypeLoadException: Could not load type 'Issue3254.AsyncGo.<go_1>\`1'`

Restored → 4/4 green, `git diff HEAD` = 0, exact `Core.dll` md5 round-trip (`11efe3d7…`) — three separate round-trips over the session.

### 5. Substitution-predicate family: **no seventh spelling**

The diff touches **none** of it — `grep` for `substitution == null|!= null|ContainsKey|TryGetValue` against the PR's own `+`/`-` lines returns empty. The `running`/`TryGetValue` block adjacent to the `StructSymbol.cs:1501` hunk is pre-existing and untouched; only the `isConstructedNested` condition above it changed.

Better than neutral: for the flattening itself the PR **reuses the existing shared helper** `StructSymbol.CollectEnclosingTypeParameters` (`StructSymbol.cs:1447`), already relied on by `StateMachineEmitter.cs:436`, `Binder.cs:611`, `EnumSymbol.cs:160/199`, `OverloadResolver.Invocations.cs:1673` and `DeclarationBinder.Structs.cs:55`. It is documented as returning CLR order, outermost-first, and it does (`levels.Insert(0, tps)` while walking outward). **Nothing hand-rolled.**

### 6. Sibling arms in the same switch — checked, honest negative

`TypeSymbol.cs:550` `EnumSymbol es when !es.EnclosingTypeArguments.IsDefaultOrEmpty` carries the **exact guard shape** that was just fixed for `StructSymbol`, and `InterfaceSymbol` at `:587`/`:594` handles no enclosing arguments at all. Both looked like the same defect one step sideways. I probed each — enum nested in a generic and captured, interface nested in a generic and captured — and **both run correctly**. Not defects; reporting the negative rather than the suspicion.

---

## NIT — `test/Compiler.Tests/Emit/Issue3254EnclosingGenericClosureEmitTests.cs`

The fix's load-bearing invariant is that `CollectEnclosingTypeParameters` returns **outermost-first**, and none of the four new pins constrains it: every pin has exactly **one** enclosing level, where the ordering is an identity.

Measured, with a positive control rather than asserted: mutating `StructSymbol.cs:1458` `levels.Insert(0, tps)` → `levels.Add(tps)` (build RC=0, md5 `11efe3d7` → `cf2be25f`) breaks my two- and three-level probes immediately (`GS0156` at bind time — the mutant is demonstrably on the path), yet **all four new pins stay green 4/4**.

**But this is not a repo-level coverage hole, and I want to be accurate about that**: I then ran the existing `Issue1537GenericNestedInGenericBinderTests` under the same mutant and it **does** catch it (`Failed: 1, Passed: 6`). So the invariant *is* pinned — at binder level, elsewhere. The only gap is that this PR's emit-level file has no multi-level row. A two-level pin here would be ~10 lines and would put emit-level coverage where the fix lives. Restored; md5 back to `11efe3d7`, `Issue1537` back to 7/7.

## Not yours — #3271 confirmed unaffected

`Run[object](42)` still gives compile rc=0 → `InvalidProgramException` rc=134 on this head. Reproduces with all hunks reverted and with no `go`/lambda at all. **Separate root, separate path**; I did not hold this PR to it.

---

### Coverage — 9 of 9, all RC=0

**15,373 passed / 0 failed / 4 skipped.**
Compiler.Tests 4271 · Core.Tests 7178 (1 skip) · Cs2Gs.Tests 1907 · Interpreter.Tests 1346 (3 skip) · LanguageServer.Tests 468 · Extensions.Tests 104 · Sdk.Tests 59 · GeneratorHost.Tests 31 · InternalAnalyzers.Tests 9.

On the gate: `IlVerifier.Verify` throws `XunitException` on failure — it genuinely asserts here, no swallowing, and the ignore list was not widened. I still treated it as worthless for this defect class and used the running program as the oracle throughout; in round 1 `ilverify` *crashed* (`IndexOutOfRangeException` at `Instantiation.get_GenericParameters`) on exactly this assembly shape rather than reporting it.

Both worktrees left porcelain-0; every mutation restored by exact md5 round-trip. **No BLOCKER, no SHOULD-FIX.**

Issue #3265, opened 2026-08-05
[tests] Four more test files still claim an interpreter-vs-emitter comparison in their names and docs


DavidObando:
Follow-up to #3257 / PR #3263. That PR corrected the parity oracles it was scoped to, and renamed `MatchEmitter` at `test/Interpreter.Tests/Issue2896StructObjectOverrideTests.cs:127`. Review found the identical defect surviving in four more files three directories away. Same class, same remedy.

The evaluator was deleted in #3244 (ADR-0156 Phase 3c). These files name or document a second engine that no longer exists, so they advertise coverage they cannot provide:

| file:line | what it claims |
|---|---|
| `test/Interpreter.Tests/PrimitiveInterpreterTests.cs:59` | test named `Interpreter_PrimitiveArithmetic_MatchesEmitter`, but runs **one column** with `Assert.Contains(literal)`; doc at `:11-15` claims an emitter comparison; comment at `:55` says "in the tree interpreter" |
| `test/Interpreter.Tests/ParenthesizedReceiverInterpreterTests.cs:24`, doc `:13` | "the interpreter (a separate execution path from the emitter)" |
| `test/Core.Tests/CodeAnalysis/Issue1881CheckedUncheckedInterpreterTests.cs:19` | "Asserts the tree-walking evaluator agrees with the compiled/emitted IL" |
| `test/Core.Tests/CodeAnalysis/Issue2023CheckedUnaryNegationInterpreterTests.cs:19` | same wording |

All five remaining `*Parity*`-named files were checked and are **clean** — they assert feature parity, not engine parity. Do not sweep those.

## Required outcome

Same per-test verdict discipline as #3257: **rename + re-document** where the test still pins something real, **convert to an explicit single-engine pin** where the second column is gone, **delete** only where the assertion is genuinely vacuous. The names must stop describing a comparison that cannot happen.

`PrimitiveInterpreterTests.cs:59` is the one to look at hardest: a single-column `Assert.Contains` on a literal is weak on its own, and the name is actively misleading about why. Decide whether it should become a real assertion or an honest pin.

## Evidence bar

- **This is test-only, which inverts the safety net: weakening tests makes CI greener.** Judge on coverage preserved, not on green.
- For anything deleted, prove the assertion was vacuous — body identity is the strongest form; a surviving mutant is **not** proof.
[...]

PR #3269, opened 2026-08-05
Correct stale interpreter test intent

DavidObando:
## Summary

- rename four stale interpreter-named suites to their real emitted-session or emitted-oracle paths
- strengthen 35 REPL specimens from substring checks to exact output pins
- preserve all 60 affected cases and update sibling XML documentation

Refs #3265

## Per-file verdict

- `PrimitiveInterpreterTests.cs` → `PrimitiveEmittedSessionTests.cs`: **rename, re-document, strengthen**. All 31 rows still run through `GSharpRepl`/`EmittedSessionEngine`; `Assert.Equal(expected + Environment.NewLine, output)` replaces weak `Assert.Contains`.
- `ParenthesizedReceiverInterpreterTests.cs` → `ParenthesizedReceiverEmittedSessionTests.cs`: **rename, re-document, strengthen**. All 4 rows now pin exact emitted-session output.
- `Issue1881CheckedUncheckedInterpreterTests.cs` → `Issue1881CheckedUncheckedEmittedOracleTests.cs`: **convert to explicit single-engine pin**. All 13 facts and assertions remain; only stale evaluator framing and unused evaluator-era imports are removed.
- `Issue2023CheckedUnaryNegationInterpreterTests.cs` → `Issue2023CheckedUnaryNegationEmittedOracleTests.cs`: **convert to explicit single-engine pin**. All 12 facts and assertions remain.

No test, helper, specimen, or assertion was removed. Nothing that was previously observable becomes undetected.

## RED proof

Temporarily appended `unexpected\n` to the primitive submission helper output. This mutation would survive the old substring assertion; the exact pin rejected every row:

```text
MUTANT_RC=1
Expected: "5000000000\n"
Actual:   "5000000000\nunexpected\n"
Failed: 31, Passed: 0, Total: 31
```

Restored source hash exactly, then affected tests passed 60/60. No surviving mutant is reported, so no survival positive control applies.

## Validation

- affected tests: 60 passed
- full nine-project suite: 15,368 passed, 4 skipped, 0 failed
- post-rebase affected tests: 60 passed
- stale-claim residue grep: 0 hits
- changed files outside `test/`: 0

### DIFF
--- a/test/Core.Tests/CodeAnalysis/Binding/Issue1881CheckedUncheckedEmittedOracleTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Binding/Issue1881CheckedUncheckedEmittedOracleTests.cs
@@ -1,29 +1,21 @@
-// <copyright file="Issue1881CheckedUncheckedInterpreterTests.cs" company="GSharp">
+// <copyright file="Issue1881CheckedUncheckedEmittedOracleTests.cs" company="GSharp">
 // Copyright (C) GSharp Authors. All rights reserved.
 // </copyright>
 
-using System.Collections.Generic;
-using GSharp.Core.CodeAnalysis;
-using GSharp.Core.CodeAnalysis.Compilation;
-using GSharp.Core.CodeAnalysis.Symbols;
-using GSharp.Core.CodeAnalysis.Syntax;
-using GSharp.Core.CodeAnalysis.Text;
 using GSharp.Tests;
 using Xunit;
 
 namespace GSharp.Core.Tests.CodeAnalysis.Binding;
 
 /// <summary>
-/// Issue #1881: interpreter counterpart of
-/// <see cref="GSharp.Core.Tests.CodeAnalysis.Emit.Issue1881CheckedUncheckedEmitTests"/>.
-/// Asserts the tree-walking evaluator agrees with the compiled/emitted IL for
-/// <c>checked</c>/<c>unchecked</c> expressions and blocks: overflow throws
+/// Issue #1881: emitted-oracle coverage for <c>checked</c>/<c>unchecked</c>
+/// expressions and blocks. Pins that overflow throws
 /// <see cref="System.OverflowException"/> (caught here via a G#-level
 /// try/catch, mirroring real usage) in a checked context, and wraps silently
 /// in an unchecked context (the C# project default), across signed/unsigned
 /// integer widths and narrowing conversions.
 /// </summary>
-public class Issue1881CheckedUncheckedInterpreterTests
+public class Issue1881CheckedUncheckedEmittedOracleTests
 {
     [Fact]
     public void CheckedAddition_Int32_Overflow_ThrowsOverflowException()
--- a/test/Core.Tests/CodeAnalysis/Binding/Issue2023CheckedUnaryNegationEmittedOracleTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Binding/Issue2023CheckedUnaryNegationEmittedOracleTests.cs
@@ -1,29 +1,21 @@
-// <copyright file="Issue2023CheckedUnaryNegationInterpreterTests.cs" company="GSharp">
+// <copyright file="Issue2023CheckedUnaryNegationEmittedOracleTests.cs" company="GSharp">
 // Copyright (C) GSharp Authors. All rights reserved.
 // </copyright>
 
-using System.Collections.Generic;
-using GSharp.Core.CodeAnalysis;
-using GSharp.Core.CodeAnalysis.Compilation;
-using GSharp.Core.CodeAnalysis.Symbols;
-using GSharp.Core.CodeAnalysis.Syntax;
-using GSharp.Core.CodeAnalysis.Text;
 using GSharp.Tests;
 using Xunit;
 
 namespace GSharp.Core.Tests.CodeAnalysis.Binding;
 
 /// <summary>
-/// Issue #2023: interpreter counterpart of
-/// <see cref="GSharp.Core.Tests.CodeAnalysis.Emit.Issue2023CheckedUnaryNegationEmitTests"/>.
-/// Asserts the tree-walking evaluator agrees with the compiled/emitted IL for
-/// unary negation of the each width's <c>MinValue</c> inside a
+/// Issue #2023: emitted-oracle coverage for unary negation of each width's
+/// <c>MinValue</c> inside a
 /// <c>checked</c>/<c>unchecked</c> context: overflow throws
 /// <see cref="System.OverflowException"/> in a checked context and wraps
 /// silently in an unchecked context (the project default per #1881), while
 /// floating-point negation never traps regardless of context.
 /// </summary>
-public class Issue2023CheckedUnaryNegationInterpreterTests
+public class Issue2023CheckedUnaryNegationEmittedOracleTests
 {
     [Fact]
     public void CheckedNegation_Int32MinValue_ThrowsOverflowException()
--- a/test/Core.Tests/CodeAnalysis/Emit/Issue1881CheckedUncheckedEmitTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Emit/Issue1881CheckedUncheckedEmitTests.cs
@@ -19,10 +19,10 @@ namespace GSharp.Core.Tests.CodeAnalysis.Emit;
 /// <c>checked { }</c> statement translated (by cs2gs) to a plain block,
 /// silently erasing overflow-trap semantics, and <c>checked(...)</c>/
 /// <c>unchecked(...)</c> expressions had no G# equivalent whatsoever. This
-/// adds the language feature to gsc itself (syntax, binding, emission,
-/// interpreter) so cs2gs can translate faithfully. These tests exercise the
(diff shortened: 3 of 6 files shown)


DavidObando:
# Rubber-duck review — PR #3269

**Verdict against `main` = `e28d55bdfc83ac78dcbd0e9f581aaf7f0f0326df`, PR head `cc437612df33e04cf84f2b0f8c3ddf5ad301d7ea`** (both fetched by me and re-fetched immediately before posting; merge-base *is* current `main`, `merge-tree --write-tree` → rc=0, single-line bare tree OID, **0 stage entries**, 0 CONFLICT).

**MERGEABLE — no blockers.** This is test-only, so CI being green proves almost nothing: a PR that *weakens* tests goes greener, not redder. I judged it entirely on whether coverage was preserved or improved, and I reproduced the load-bearing claim rather than accepting it. It holds, cleanly.

## The strengthening claim is real, and I reproduced it decisively

I did not perturb the test helper — I mutated the **product** print path (`src/Repl/Compat/GSharpRepl.cs`, one extra `Console.WriteLine("unexpected")` after the value write, anchor asserted unique so there was no brace-hazard no-result, mutation proven with `git diff HEAD` = 1 insertion).

Then I restored `main`'s two old classes **alongside** the PR's new ones, so both assertion styles ran against the same mutated binary in a single pass:

```
Failed! - Failed: 35, Passed: 35, Total: 70
```

Broken down by class:

| class | assertion style | result under output corruption |
|---|---|---|
| `PrimitiveEmittedSessionTests` (31) | `Assert.Equal` (new) | **31 FAILED** |
| `ParenthesizedReceiverEmittedSessionTests` (4) | `Assert.Equal` (new) | **4 FAILED** |
| `PrimitiveInterpreterTests` (31) | `Assert.Contains` (old) | **31 passed — survived** |
| `ParenthesizedReceiverInterpreterTests` (4) | `Assert.Contains` (old) | **4 passed — survived** |

Old classes appear **0 times** in the failure log. That is a perfect 35/35 discrimination, and the survival is a genuine survival rather than a no-result: **the positive control is intrinsic** — the new tests failing on the same product line proves the mutation was on the path.

So the strengthening is **not decorative**. The old checks were genuinely weak, and structurally so: `Assert.Contains("2", output)` for `100L / 50L` passes on any output containing the digit `2`, and `Assert.Contains("1", output)` for `1L << 64` passes on `18446744073709551616`. Several specimens could not distinguish the correct answer from a wrong one.

## Over-tightening — checked specifically, and the author made the *right* call

`Assert.Equal(expected + Environment.NewLine, output)` looks like a portability hazard, and there **is** a Windows leg to trip (`windows-nightly.yml` runs the full matrix on `windows-latest`, plus two `windows-latest` jobs in `build.yml`). I traced the print path: `src/Repl/Compat/GSharpRepl.cs:32` does `Console.WriteLine(cell.Value)` onto a default `new StringWriter()`, whose `CoreNewLine` **is** `Environment.NewLine`. So expectation and production agree on both platforms. A hard-coded `"\n"` would have been the bug; `Environment.NewLine` is correct. No finding.

**Culture:** four specimens render culture-sensitively — `PrimitiveEmittedSessionTests.cs:25,26,27,49` (`3.75`, `2.5`, `3.75`, `-3.14`), and culture is not pinned (`InvariantGlobalization` is set only under `tools/cs2gs/corpus/`). But the exposure is **identical before and after** — `Assert.Contains("3.75", …)` fails under `de-DE` exactly as `Assert.Equal("3.75\n", …)` does. **Not introduced by this PR**, so I am not raising it as a finding; recording it because it was worth checking.

One genuine *additional* strengthening worth noting: exact-output now also fails if any stray diagnostic is printed alongside the value. `Contains` tolerated that. Strictly better.

## Renames are honest — verified at the body, not the name

This is the failure mode the whole issue family exists to kill, so I checked the bodies rather than trusting the new names:

- The two `Core.Tests` files call a private `Evaluate` → `EmittedOracle.Evaluate` (`test/Shared/EmittedOracle.cs:75`), which builds a real `Compilation`, emits, loads into a collectible ALC and runs. `…EmittedOracleTests` is accurate.
- The two `Interpreter.Tests` files drive `new GSharpRepl()`, which since #3244 is the shim over `EmittedSessionEngine`. `…EmittedSessionTests` is accurate.

No fresh instance of the original defect.

## "13 facts preserved" / "12 facts preserved" — audited, and stronger than claimed

| file | `[Fact]` main→PR | `Assert.` main→PR | body |
|---|---|---|---|
| `Issue1881CheckedUncheckedEmittedOracleTests.cs` | 13 → **13** | 26 → **26** | **byte-identical** from first `[Fact]` (md5 `6c8fb0c3…`) |
| `Issue2023CheckedUnaryNegationEmittedOracleTests.cs` | 12 → **12** | 24 → **24** | **byte-identical** (md5 `73934b48…`) |

Equal counts alone would not rule out an assertion weakening *in kind* (`Assert.Equal` → `Assert.NotNull` preserves the count), so I hashed the bodies. They are unchanged byte-for-byte; only the header, doc-comment, class name and now-unused `using` directives moved. The `−12` per file is dead `using`s, not coverage.

And the tests genuinely run under the new names — **25 = 13 + 12** with an explicit filter, while the old names correctly return `No test matches` (a real zero here, not the trap).

## Verified test totals (mine, grepped from logs)

| suite | result |
|---|---|
| the four renamed classes | 35 + 25 = **60 passed / 0 failed** |
| `Interpreter.Tests` full | **1346 passed / 0 failed / 3 skipped** |
| `Core.Tests` full | **7178 passed / 0 failed / 1 skipped** |
| Release build, `-p:ContinuousIntegrationBuild=true` | rc=0, **0 warnings** |

Restoration proven three ways: `git diff HEAD` = 0, a clean green re-run, and `GSharp.Core.dll` md5 round-tripping **exactly** to `3a027944cfe5d15ce46c97396e1207e2`.

## Ponytail note — the shape is right

Worth calling out because it is easy to miss: the diff is **+28 / −46**, a net *deletion*, while strengthening 35 assertions. That is because the change is one line inside each `[Theory]` body, which upgrades all 31 (and all 4) `[InlineData]` cases at once. Strengthening 35 assertions by editing 2 lines, and deleting more than you add, is the correct lazy shape. No new helper, no new abstraction, nothing scaffolded.

---

## NIT 1 — the report under-counts the change (6 files, not 4)

`test/Core.Tests/CodeAnalysis/Emit/Issue1881CheckedUncheckedEmitTests.cs:22-25` and `test/Core.Tests/CodeAnalysis/Emit/Issue2023CheckedUnaryNegationEmitTests.cs:22-24` are also modified. They are **correct and necessary** — they carry `<see cref="…InterpreterTests"/>` at the renamed classes, and without the update the cref dangles. Finding the mirror references is exactly right and I want to credit it. The only issue is that a summary saying "four files, nothing removed" is harder to audit than it should be when six files moved. No code change needed.

## Report item — the sweep is not on its last wave; **45 files still carry a live false claim**

This is not this PR's defect and not a reason to hold it — flagging because each previous wave believed it was the last, and the remaining population is an order of magnitude larger than the four fixed here.

Counting only files whose doc-comment makes a **present-tense** interpreter-vs-emitter claim **and** carries no retirement note (i.e. excluding the many files already honestly re-documented as "the interpreter parity arm was retired with the evaluator in Phase 3c"): **45 files**.

Representative, each verified to still make the claim:

- `test/Interpreter.Tests/Issue711IfExpressionInterpreterTests.cs:11` — *"interpreter parity … pins down the REPL/evaluator execution semantics"*, while `:224` drives `new GSharpRepl()`, i.e. the emitted session. Advertises coverage that cannot exist — precisely #3265's defect.
- `test/Interpreter.Tests/Issue833OpenTGenericMethodCallReturnTypeInterpreterTests.cs:12` — *"Tree-walking interpreter parity"*
- `test/Interpreter.Tests/Issue799VariadicInterpreterTests.cs:18`, `Issue729FriendlyNumericAliasInterpreterTests.cs:13`, `Issue710NullConditionalIndexingInterpreterTests.cs:12`, `Issue2925InterfaceDelegatePropertyInvocationTests.cs:12`, `Issue608InterpreterCLRDispatchTests.cs:12`, `StaticVirtualInterfaceInterpreterTests.cs:18`, `AttributeInterpreterTests.cs:12`
- outside `Interpreter.Tests`: `test/Core.Tests/CodeAnalysis/Binding/Issue751RichReceiverBinderTests.cs:18` — *"Binder + tree-walking interpreter coverage"* — and `test/Core.Tests/CodeAnalysis/Binding/InterpolatedStringHandlerArgumentTests.cs:29` — *"cover tree-walk interpreter parity"*.

Separately: **the five `*Parity*`-named files are correctly left alone.** I spot-checked all five and they are feature parity, not engine parity — slice-vs-array (#611), arrow-lambda-vs-func-literal (#1172 ×2), `@assembly:`-vs-C# (#2237), and the LSP protocol contract. That judgement was right.

---

**Method:** own clone with its own `out/`, never the shared checkout; no `git stash`; `merge-tree` read by **stages**, not `CONFLICT` counts; every test count grepped from the log rather than inferred from rc; mutation asserted with `git diff HEAD` and a uniqueness check on the anchor; worktree restored and verified clean.

PR #3270, opened 2026-08-05
Restore permission-boundary tense in superseded ADR

### LINKED ISSUE #3256: [docs] Documentation still advertises the removed --engine evaluator; pin the gsi --help transcript so it stops drifting

DavidObando:
## Summary

- restore present tense for `unsafe` as a current permission boundary in ADR-0153
- retain past tense for ADR-0152 driver behavior inside its closed “During ADR-0156 Phases 1–3b” frame

## Verification

- reread the full ADR-0152 paragraph: both semicolon-joined behavioral clauses use past tense and agree with the closed-window frame
- `git diff origin/main...HEAD --stat`: one file, one insertion, one deletion
- website search found no ADR-0152 or ADR-0153 mirror
- documentation-only change; no tests run

Fixes #3256

### DIFF
--- a/docs/adr/0153-interpreter-compiled-only-storage-boundary.md
+++ b/docs/adr/0153-interpreter-compiled-only-storage-boundary.md
@@ -46,7 +46,7 @@ natively. Phase 3c deleted the evaluator path and its boundary diagnostics.
 This ADR governed only the evaluator's storage-model boundary. It did not
 redefine unsafe/native language validity or CIL emission.
 
-The `unsafe` context itself remained supported. It was a permission boundary, not
+The `unsafe` context itself remained supported. It is a permission boundary, not
 a storage operation; an `unsafe` block containing only otherwise-supported
 value operations continued to evaluate normally.
 


DavidObando:
## ⚠️ SHOULD-FIX (1) — 0 BLOCKER, 1 SHOULD-FIX, 0 NIT

Verdict against **`main` = `e28d55bdfc83ac78dcbd0e9f581aaf7f0f0326df`**, re-fetched immediately before posting, head **`3dc9e037df6ed22eb51ec05d6d96a0e74c5a9e32`**. Merge-base **equals** main; `merge-tree --write-tree` read by **stages** gives 0 stage-1/2/3 entries. Built in my own worktree: 0 Warning(s), 0 Error(s), `GSharp.Core.dll` md5 **`1653f2eaeef7a2f141e9640776a2faaf`**.

**This PR exists because of a NIT I filed on #3261. Having measured it properly, half of that NIT was wrong, and this PR faithfully implements the wrong half.** Details below.

### Both author behaviours independently re-verified ✅

| claim | my measurement | matches |
|---|---|---|
| `unsafe` is a permission boundary | fn-ptr **outside** `unsafe` → **rc 1**, `GS0404` + `GS0405`, compiler text *"requires an 'unsafe' context … (ADR-0122 §9)"*; **inside** → **rc 0**, prints **`7`** | ✅ |
| `gsc /out:` emits to disk | **rc 0**, `Wrote emitted.dll`, **2560 bytes**, magic `MZ` | ✅ exact |

Same construct rejected outside and accepted inside is a real enforcement signal, not a doc reading. `rc` captured before any pipe.

*(One probe of mine was a near-miss: my first `unsafe` file wrapped the code in `func main()`, which G# never invokes — top-level statements are the entry. It returned **rc 0 with no output**, which would have scored as a false confirm. Rebuilt at top level to get the real `7`.)*

### Scope, mirrors, inverse audit — all clean ✅

- **Scope:** exactly 2 files, 2 lines, one word each. Nothing else moved.
- **Mirror pairs:** no `website/` twin exists for **any** ADR (`git ls-tree website/ | grep -i adr` → empty), and these two sentences appear in exactly one file each. No stale twin. ✅
- **Inverse audit (I re-ran it, not accepting the negative):** every prose claim about `Evaluator*`/`SessionEngine`/`Compilation.Evaluate`/`GS0510`/`GS0513`/`GS0514`/`GS9999` in both files is past tense. The only present-tense verbs about deleted things sit inside a table captioned *"The final pre-deletion implementation status **was**:"* — correctly framed, and flagging those would just spin the ratchet another turn. ✅

---

## SHOULD-FIX — `docs/adr/0152-interpreter-native-call-boundary.md:38`: revert this hunk

Reading the **full paragraph plus its history** — not the diff line — inverts the case for this half.

**Before #3261** (at `8b3bdac57`) the paragraph was internally coherent in the *present*:

> **Since** ADR-0156 Phases 1–3a, this boundary **applies** to … Bare `gsc`, `gsi <file>`, and the default interactive REPL **emit and run** the native call; `gsc /out:` **emits** it to disk.

**#3261 did not merely flip verbs — it changed the framing preposition**, "**Since** … 1–3a" (ongoing) → "**During** … 1–3b" (closed window), and moved every verb to past to match. That was a complete, coherent frame+tense conversion.

**This PR** now puts a present-tense verb back inside that closed-window frame:

> **During ADR-0156 Phases 1–3b**, … not to default drivers. Bare `gsc`, `gsi <file>`, and the default interactive REPL **emitted and ran** the native call; `gsc /out:` **emits** it to disk.

Two semicolon-joined, grammatically parallel clauses about four drivers now disagree in tense, with no semantic basis for the contrast. A reader infers the shift is meaningful — that the first three **stopped** while the fourth continues. **I measured all four; every one still works:**

| driver | probe | result |
|---|---|---|
| bare `gsc` | `samples/PInvoke.gs` | rc 0, prints **`13`** |
| `gsi <file>` | `samples/PInvoke.gs` | rc 0, prints **`13`** |
| default interactive REPL | `EmittedSessionEngine.Evaluate` **in-process** | `Value=13`, `HasError=False` |
| `gsc /out:` | — | rc 0, 2560-byte PE |

`13` is real `strlen("Hello, world!")` from libc — a fabricated return would be `0`. So "still true" does **not** discriminate here: all four clauses are equally still-true, which is exactly why they should share one tense. The sentence's job is to describe the Phases 1–3b **contrast** ("not to default drivers" → what default drivers did instead), so the closed-window past tense is correct and deliberate.

**Fix:** revert `emits` → `emitted` on line 38. That restores the paragraph #3261 deliberately built.

### Keep the ADR-0153 hunk — it is a different kind of claim ✅

`docs/adr/0153-…:49` "It **is** a permission boundary, not a storage operation" is a **definitional** aside, not a behavioural claim scoped to a window — the "Caesar crossed the Rubicon, which **is** a river" pattern, where categorical truths stay present inside past narrative. It explains *why* the ADR didn't need to touch `unsafe`, and I measured that it remains true (`GS0404`/`GS0405` above). Present tense here is right.

---

### My error, stated plainly

My #3261 NIT flagged both sentences on the single test *"is this still true?"* and concluded past tense was an over-correction. That test is wrong for ADR-0152: I judged the sentence in isolation and **missed that the same hunk had changed "Since" to "During"**, which licenses — indeed requires — the past tense. The right test is *"is this a definitional claim, or a behavioural claim inside a framed window?"* ADR-0153 is the former (present ✓); ADR-0152 is the latter (past ✓).

Net: **this PR is half-right.** Landing the ADR-0153 word alone is the smaller and more correct change.

Issue #1379, opened 2026-06-28
gsc: user-defined generic-type static method return type not substituted with receiver type argument (GS0155 'Box' to 'Box')


DavidObando:
### Summary

A `shared` (static) method on a **user-defined generic type** whose **return type references the type parameter** does not get the receiver's type argument substituted at the call site. The call resolves to the **raw** (unsubstituted) generic type, so converting it to the closed type fails with `GS0155: Cannot convert type 'Box' to 'Box'` (note: identical display names — open vs. closed construction).

### Minimal repro

```gs
package p

struct Box[T] {
    shared {
        func Make() Box[T] -> Box[T]{}
    }
}

class C {
    func F() Box[int32] {
        return Box[int32].Make()
    }
}
```

```
genret.gs(11,16,11,33): error GS0155: Cannot convert type 'Box' to 'Box'.
```

### Control (works)

A static method on the same generic type whose return type does **not** reference `T` binds and converts fine:

```gs
struct Box[T] { shared { func Make() int32 -> 5 } }
class C { func F() int32 { return Box[int32].Make() } }   // OK
```

### Expected

`Box[int32].Make()` should be typed `Box[int32]` (the receiver's type argument `int32` substituted into the static method's return type `Box[T]`), matching the declared return type of `F`.

### Notes / scope

- #1323 covers **binding** `Type[TypeArg].StaticMember(...)` receivers, but its tests only use a **non-generic** static return type (`int32`), so the return-type substitution path is untested.
- #1216 covered the analogous gap for **imported/BCL** generic static methods; this is the **user-defined** generic-type counterpart.
- Manifests in the Oahu migration (`Oahu.Decrypt/Mp4File.cs` → `Mp4Operation<ChapterInfo?>.FromCompleted(this, null)`), where in the richer context it surfaces as `GS0124: Expression must have a value` rather than GS0155, but the root cause is the same: the static method's return type is not substituted with the receiver's type argument.

### Refs

Found during cs2gs Oahu migration work (#914).

PR #3272, opened 2026-08-05
Validate explicit generic calls and reified boxing

### LINKED ISSUE #3267: [binder] Explicit generic arguments are not checked against call arguments: Id[int32]("wrong") compiles clean and emits invalid IL
Found while fixing #3254 (PR #3264), but **unrelated to either of that issue's roots** — this is binder argument validation, not emit. Filing separately.

Measured on `main` = `d172d0d14424df9caaa06de65a2a3f52619a1cc4`.

## Repro

```gsharp
package Repro
import System

func Id[T](value T) T {
    return value
}

func Main() {
    Console.WriteLine(Id[int32]("wrong"))
}
```

## What happens

- `gsc`: rc **0**, stdout `Wrote .../Repro.dll`, stderr empty, **zero diagnostics**
- running the produced executable: rc **134**

```text
Unhandled exception. System.InvalidProgramException:
Common Language Runtime detected an invalid program.
   at Repro.<Program>.Main()
```

## What should happen

rc 1 with **GS0154**:

```text
Parameter 'value' requires a value of type 'int32' but was given a value of type 'string'.
```

**The diagnostic already exists and already fires for the adjacent non-generic case** — a plain `Take("wrong")` against `func Take(value int32)` produces exactly that text. So this is not a missing diagnostic to design; it is an existing check that a generic call skips.

## Where

- `src/Core/CodeAnalysis/Binding/OverloadResolver.CallBinding.cs:1645-1647` — generic substitution suppresses the incompatibility check
- `src/Core/CodeAnalysis/Binding/OverloadResolver.Invocations.cs:1540-1544` — the instance-method twin returns before checking the substituted type

**The instance-method form reproduces identically.** Both sites need fixing; a fix to one leaves the other open. Prefer a single shared check over two parallel patches if the shapes allow it.

## Severity

Once `T` is explicitly bound to `int32`, `value` **is** `int32` and a `string` argument must be rejected. Accepting it is a type-safety hole: the compiler reports success and writes an executable that cannot run.

To be precise about the failure mode — it is **under-diagnosed, not silently wrong**. The bad program crashes rather than printing a wrong answer, which is the better of the two, but rc=0 from `gsc` is still a false claim of success.

## Evidence bar

- The oracle is the **running program**, not a clean compile and not ILVerify. This issue is exactly the shape ILVerify is bad at: it has been caught this session returning **rc=0 while printing `Assembly or module not found: mscorlib`**, and separately green-lighting an assembly the CLR refuses to JIT (#3240, open). **Compile, run, assert rc and stdout.**
- Cover **both** the free-function and instance-method forms, and check whether extension and constructor calls have the same hole before concluding the class is closed.
- Guard against over-correcting: valid explicit instantiations must keep compiling. Pin at least one case where the explicit argument is correct and one where inference would have picked a different-but-valid type.
- Assert the diagnostic **id and span**, not just the id.
- **A no-result is not a result**: `No test matches` exits rc=0; a probe that fails to parse yields `GS0005` that reads like a result. Freshness witness `GSharp.Core.dll` with a moving md5.

### LINKED ISSUE #3271: [emit] Legal generic call with a reference-type argument, 'Run[object](42)', compiles rc=0 then throws InvalidProgramException at runtime
Found by review of PR #3264 (issue #3254) and confirmed **pre-existing**: it reproduces with both of that PR's hunks reverted, and with **no `go` statement and no lambda at all**. So it is not closure-related and not that PR's regression.

## Repro

`Run[object](42)` → `gsc` **rc 0, zero diagnostics, writes the assembly**; running it → **rc 134, `System.InvalidProgramException`**.

Boxing an `int32` into an `object` type parameter is a **legal, ordinary call**. This must either compile and run, or be diagnosed. Producing a JIT-rejected assembly is neither.

## This is NOT a duplicate of #3267 — check both, fix them separately

They share a crash type and nothing else:

| | #3267 | **this issue** |
|---|---|---|
| call | `Id[int32]("wrong")` | `Run[object](42)` |
| is the call legal? | **no** — `string` for `T=int32` | **yes** — `int32` boxes to `object` |
| correct outcome | **reject** at compile time (`GS0154` already does this for the non-generic twin) | **compile and run correctly** |
| defect class | under-diagnosed | **miscompiled** |

#3267 is a missing check. This is a code-generation bug on a call that should work. **A fix that makes this one *diagnose* rather than *run* is wrong.**

Also distinct from #3240 (`ref struct` without declared `ToString`), which reaches `InvalidProgramException` through the lowering path.

## Where to look

The reviewer's root-cause read on the sibling defect in #3264 was that IL referenced a **generic ordinal the closure never declared**. Worth checking whether this shares that shape — i.e. whether the boxing/constrained path emits a token against the wrong instantiation when the type argument is a reference type. Confirm or refute by measurement; do not assume.

Scope it before concluding: try other reference-type arguments (`string`, an interface, a user class), other value-type-into-reference-type-parameter combinations, and the same call shapes as instance methods and extension methods. Report which are affected.

## Evidence bar

- **The oracle is the running program.** rc 0 plus a written assembly is exactly the bug here — never score a clean compile as a pass.
- **Not ILVerify.** On the closely-related assembly in #3264 it did not report a violation, it **crashed**: `IndexOutOfRangeException at Instantiation.get_GenericParameters`. Separately caught this session returning rc=0 while printing `Assembly or module not found: mscorlib`, and green-lighting an assembly the CLR refuses to JIT (#3240, open). CI's ilverify job only covers sample assemblies.
- **Beware erasure-blind assertions.** `Console.WriteLine(object)` renders `42` and `"right"` identically, so an output pin can stay green while the wrong type argument is in play. Use a **type detector** — e.g. assert `[]T{…}.GetType()` renders `System.Int32[]` vs `System.String[]` — and include a **transposition case** (`[int32,string]` vs `[string,int32]`) where swapped arguments would still load and print something plausible. That technique is what cleared #3264's pins.
- RED-proof every new pin: revert the product hunk alone, rebuild, show the pin failing **for the right reason**, restore, `git diff HEAD` = 0.
- **A no-result is not a result**: `No test matches` exits rc=0; a blank passed-count beside real numbers is a no-result. Freshness witness **`GSharp.Core.dll`**, md5 must move.
- Re-verify the repro on current `main` before starting; the measurement above was taken against a build of `c166f7b03` with hunks reverted.

DavidObando:
## Summary

- validate closed substituted parameter types in free, instance, and extension generic calls
- materialize implicit conversions once an explicitly substituted bare generic slot is closed, including value-type boxing to `object`, interfaces, and nullable reference targets
- preserve genuinely open generic forwarding by skipping only bare slots whose substituted target still contains a type parameter
- replace misleading GS0156 cast advice and correct the stale free/instance mirror comment
- port nine runtime requirements from closed PR #3277

## Behavior

| shape | before | after |
|---|---|---|
| free `Id[int32]("wrong")` | compile rc 0; run rc 134 | `GS0154` at `"wrong"` |
| instance generic mismatch | compile rc 0; run rc 134 | `GS0156` at `"wrong"` |
| extension generic mismatch | compile rc 0; run rc 134 | `GS0156` at `"wrong"` |
| free `Show[object](42)` | compile rc 0; run rc 134 | run rc 0; runtime type `System.Int32`; value `42` |
| instance / extension / shared boxing | invalid program in affected paths | compile and run rc 0 |

Constructor and shared mismatch paths already used substitution-aware validation and remain unchanged.

## ValueType boundary

Issue #3280 tracks missing value-type → `System.ValueType` conversion classification. Ported rows retain that requirement as current diagnostic pins: `GS0154` for free calls and `GS0155` for instance/extension calls, including exact covered argument text. This PR does not change that conversion.

## GS0156 guidance

GS0156 now says `Provide a value of type 'int32' instead.` Following it with `Box().Id[int32](5)` compiles rc 0, runs rc 0, and prints:

```
System.Int32
5
```

The old suggested cast produced GS9998 and is no longer recommended.

## Mutation proof

Reverting only the closed-slot predicate moved `GSharp.Core.dll`:

```
5ca986f821cd20ff5da360490885c403
→ 8b354d3c5b430ac5dfde535d58514d93
→ 5ca986f821cd20ff5da360490885c403
```

Three free-function pins failed with `InvalidProgramException`. Standalone `object`, `IComparable`, and `object?` probes each compiled rc 0 with zero diagnostics, then ran rc 134. Restoring via `git checkout HEAD -- <path>` returned `git diff HEAD` rc 0 and all nine ported tests passed.

## Validation

- rebased onto `7341a6e688b6d4f1602dfaead0aca6fb1004e093`
- issue #3267 focused tests: 18/18 passed
- issue #3271 ported tests: 9/9 passed
- full nine-project suite: 15,416 passed, 4 skipped, 0 failed

Fixes #3267

Fixes #3271

### DIFF
--- a/docs/adr/0099-lsp-nil-quick-fixes.md
+++ b/docs/adr/0099-lsp-nil-quick-fixes.md
@@ -19,7 +19,7 @@ operators through correctly:
 | ---- | ----------- | ------- |
 | `GS0154` | `DiagnosticBag.ReportWrongArgumentType` | An argument of type `T?` was passed to a parameter declared `T`. |
 | `GS0155` | `DiagnosticBag.ReportCannotConvert` | An expression of type `T?` was used where `T` was required (assignment, return value, conditional arm, …). |
-| `GS0156` | `DiagnosticBag.ReportCannotConvertImplicitly` | Same as GS0155 but an explicit conversion exists (typically the implicit-only nullable→non-nullable). |
+| `GS0156` | `DiagnosticBag.ReportCannotConvertImplicitly` | Same as GS0155 but reported from an implicit-conversion context (typically nullable→non-nullable). |
 | `GS0158` | `DiagnosticBag.ReportUnableToFindMember` | A member access `a.b` did not find `b`; when the receiver is `T?` the binder reports this rather than a dedicated "deref of nullable" message — the `Nullable<T>` projection for value types and the NRT receiver path both flow here. |
 | `GS0274` | `DiagnosticBag.ReportNilNotAssignableToNonNullableParameter` | A literal `nil` was passed to a non-nullable parameter. |
 
--- a/docs/diagnostics.md
+++ b/docs/diagnostics.md
@@ -118,7 +118,7 @@ IDs may be given as `GS0001`, `0001`, or the bare integer `1`; all three forms a
 | GS0153 | Error | Constraint is neither an interface nor a class. | A generic type-parameter constraint must be an interface (any interface — sealed or not, generic or not;) or a base class; a value type such as a struct or enum is rejected. |
 | GS0154 | Error | Wrong argument type. | A positional argument's type does not match the parameter type. |
 | GS0155 | Error | Cannot convert type. | An explicit cast between incompatible types. |
-| GS0156 | Error | Cannot convert implicitly; explicit conversion exists. | `int x = 3.14` — an explicit cast is available but was not written. |
+| GS0156 | Error | Cannot convert implicitly. | The source value is not implicitly compatible with the required target type; provide a value of the target type instead. |
 | GS0157 | Error | Cannot find type (possibly missing import). | A package-qualified type name that resolves to nothing. |
 | GS0158 | Error | Cannot find member. | A field or property access that does not resolve. |
 | GS0159 | Error | Cannot resolve function call. | A function name does not resolve, or its nullable receiver lacks valid non-null narrowing. |
--- a/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.CallBinding.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.CallBinding.cs
@@ -1704,7 +1704,9 @@ internal sealed partial class OverloadResolver
                 boundArguments[i] = conversions.BindConversion(argLoc, argument, expectedType);
             }
             else if (argument.Type != expectedType
-                && !(substitution != null && parameter.Type is TypeParameterSymbol)
+                && !(substitution != null
+                    && parameter.Type is TypeParameterSymbol
+                    && TypeSymbol.ContainsTypeParameter(expectedType))
                 && (!(substitution != null && TypeSymbol.ContainsTypeParameter(parameter.Type))
                     || (Conversion.Classify(argument.Type, expectedType).IsImplicit
                         && structuralArgumentType is not FunctionTypeSymbol)))
@@ -1731,9 +1733,10 @@ internal sealed partial class OverloadResolver
                 // (`int32 → int64`) and other representation-changing
                 // implicit conversions.
                 //
-                // Issue #3222: only a BARE erased slot (`parameter.Type is
-                // TypeParameterSymbol`, the `!!T` MVAR the emitter erases at
-                // the call boundary) is skipped.
+                // Issue #3222: only a BARE erased slot whose substituted target
+                // remains open is skipped. Once explicit substitution closes
+                // the slot, this branch must materialize conversions such as
+                // value-type boxing, matching the instance and extension paths.
                 // A parameter that merely CONTAINS a type parameter (e.g. a
                 // constructed generic interface `IHolder[T]`) emits as a real
                 // constructed slot (`IHolder<!!T>`), so a value-type argument
--- a/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.CallBinding.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.CallBinding.cs
@@ -1704,9 +1704,7 @@ internal sealed partial class OverloadResolver
                 boundArguments[i] = conversions.BindConversion(argLoc, argument, expectedType);
             }
             else if (argument.Type != expectedType
-                && !(substitution != null
-                    && parameter.Type is TypeParameterSymbol
-                    && TypeSymbol.ContainsTypeParameter(expectedType))
+                && !(substitution != null && parameter.Type is TypeParameterSymbol)
                 && (!(substitution != null && TypeSymbol.ContainsTypeParameter(parameter.Type))
                     || (Conversion.Classify(argument.Type, expectedType).IsImplicit
                         && structuralArgumentType is not FunctionTypeSymbol)))
@@ -1733,10 +1731,9 @@ internal sealed partial class OverloadResolver
                 // (`int32 → int64`) and other representation-changing
                 // implicit conversions.
                 //
-                // Issue #3222: only a BARE erased slot whose substituted target
-                // remains open is skipped. Once explicit substitution closes
-                // the slot, this branch must materialize conversions such as
-                // value-type boxing, matching the instance and extension paths.
+                // Issue #3222: only a BARE erased slot (`parameter.Type is
+                // TypeParameterSymbol`, the `!!T` MVAR the emitter erases at
+                // the call boundary) is skipped.
                 // A parameter that merely CONTAINS a type parameter (e.g. a
                 // constructed generic interface `IHolder[T]`) emits as a real
                 // constructed slot (`IHolder<!!T>`), so a value-type argument
--- a/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.CallBinding.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.CallBinding.cs
@@ -1643,7 +1643,7 @@ internal sealed partial class OverloadResolver
             }
 
             if (argument.Type != expectedType
-                && !(substitution != null && TypeSymbol.ContainsTypeParameter(parameter.Type))
+                && !TypeSymbol.ContainsTypeParameter(expectedType)
                 && !Conversion.Classify(argument.Type, expectedType).IsImplicit)
             {
                 // Issue #889: arrow/func literal → void-returning delegate.
--- a/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.CallBinding.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.CallBinding.cs
@@ -1704,7 +1704,9 @@ internal sealed partial class OverloadResolver
                 boundArguments[i] = conversions.BindConversion(argLoc, argument, expectedType);
             }
             else if (argument.Type != expectedType
-                && !(substitution != null && parameter.Type is TypeParameterSymbol)
+                && !(substitution != null
+                    && parameter.Type is TypeParameterSymbol
+                    && TypeSymbol.ContainsTypeParameter(expectedType))
                 && (!(substitution != null && TypeSymbol.ContainsTypeParameter(parameter.Type))
                     || (Conversion.Classify(argument.Type, expectedType).IsImplicit
                         && structuralArgumentType is not FunctionTypeSymbol)))
@@ -1731,11 +1733,10 @@ internal sealed partial class OverloadResolver
                 // (`int32 → int64`) and other representation-changing
                 // implicit conversions.
                 //
-                // Issue #3222: only a BARE erased slot (`parameter.Type is
-                // TypeParameterSymbol`, the `!!T` MVAR the emitter erases at
-                // the call boundary) is skipped — mirroring the equivalent
-                // guard in `BindUserInstanceCall`'s per-argument loop
-                // (`if (paramType is TypeParameterSymbol) { …; continue; }`).
+                // Issue #3222: only a BARE erased slot whose substituted target
+                // remains open is skipped. Once explicit substitution closes
+                // the slot, this branch must materialize conversions such as
+                // value-type boxing, matching the instance and extension paths.
                 // A parameter that merely CONTAINS a type parameter (e.g. a
                 // constructed generic interface `IHolder[T]`) emits as a real
                 // constructed slot (`IHolder<!!T>`), so a value-type argument
--- a/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.Invocations.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.Invocations.cs
@@ -1146,7 +1146,9 @@ internal sealed partial class OverloadResolver
         for (var i = 0; i < permutedArguments.Length; i++)
         {
             var paramType = extension.Parameters[i + 1].Type;
-            if (substitution != null && TypeSymbol.ContainsTypeParameter(paramType))
+            var expectedType = substitution != null ? substituteType(paramType, substitution) : paramType;
+            if (substitution != null
+                && TypeSymbol.ContainsTypeParameter(paramType))
             {
                 if (paramType is FunctionTypeSymbol openFunctionParameter
                     && tryGetFunctionLiteral(permutedArguments[i], out var functionLiteralArgument))
@@ -1160,17 +1162,18 @@ internal sealed partial class OverloadResolver
                     continue;
                 }
 
-                // A parameter typed as an open T is encoded as System.Object in
-                // the emitted signature; pass the argument unconverted so the
-                // emitter inserts box / unbox.any around the erased boundary.
-                convertedArgs.Add(permutedArguments[i]);
+                if (TypeSymbol.ContainsTypeParameter(expectedType))
+                {
+                    // A parameter typed as an open T is encoded as System.Object in
+                    // the emitted signature; pass the argument unconverted so the
+                    // emitter inserts box / unbox.any around the erased boundary.
+                    convertedArgs.Add(permutedArguments[i]);
+                    continue;
+                }
             }
-            else
-            {
-                var expectedType = substitution != null ? substituteType(paramType, substitution) : paramType;
-                var argLoc = i < permutedSyntax.Length ? permutedSyntax[i].Location : ce.Location;
-                convertedArgs.Add(conversions.BindCallArgumentWithRefKind(argLoc, permutedArguments[i], expectedType, extension.Parameters[i + 1]));
-            }
+
+            var argLoc = i < permutedSyntax.Length ? permutedSyntax[i].Location : ce.Location;
+            convertedArgs.Add(conversions.BindCallArgumentWithRefKind(argLoc, permutedArguments[i], expectedType, extension.Parameters[i + 1]));
         }
 
         // Issue #1931: stash the extension function's own (explicit or
@@ -1509,6 +1512,7 @@ internal sealed partial class OverloadResolver
         {
             var parameter = method.Parameters[i + parameterOffset];
             var paramType = parameter.Type;
+            var expectedType = substitution != null ? substituteType(paramType, substitution) : paramType;
 
             // ADR-0060 / issue #1133: an inline-decl `out var n` / `out let n` /
             // `out _` was bound with TypeSymbol.Error in the first pass (from
@@ -1537,13 +1541,11 @@ internal sealed partial class OverloadResolver
             // An argument bound to an open type parameter is left untouched —
             // the emitter boxes value types at the call boundary (the parameter
             // is encoded as System.Object under the type-erased model).
-            if (paramType is TypeParameterSymbol)
+            if (paramType is TypeParameterSymbol && TypeSymbol.ContainsTypeParameter(expectedType))
             {
                 convertedArgs.Add(permutedArguments[i]);
                 continue;
             }
-
-            var expectedType = substitution != null ? substituteType(paramType, substitution) : paramType;
 
             if (substitution != null
                 && tryGetFunctionLiteral(permutedArguments[i], out var functionLiteralArgument))
--- a/src/Core/CodeAnalysis/DiagnosticDescriptors.cs
+++ b/src/Core/CodeAnalysis/DiagnosticDescriptors.cs
@@ -72,7 +72,7 @@ internal static class DiagnosticDescriptors
     internal static readonly DiagnosticDescriptor ConstraintNotInterface = new("GS0153", DiagnosticSeverity.Error, "Type '{0}' cannot be used as a type-parameter constraint because it is neither an interface nor a class.");
     internal static readonly DiagnosticDescriptor WrongArgumentType = new("GS0154", DiagnosticSeverity.Error, "Parameter '{0}' requires a value of type '{1}' but was given a value of type '{2}'.");
     internal static readonly DiagnosticDescriptor CannotConvert = new("GS0155", DiagnosticSeverity.Error, "Cannot convert type '{0}' to '{1}'.");
-    internal static readonly DiagnosticDescriptor CannotConvertImplicitly = new("GS0156", DiagnosticSeverity.Error, "Cannot convert type '{0}' to '{1}'. An explicit conversion exists (are you missing a cast?)");
+    internal static readonly DiagnosticDescriptor CannotConvertImplicitly = new("GS0156", DiagnosticSeverity.Error, "Cannot convert type '{0}' to '{1}'. Provide a value of type '{1}' instead.");
     internal static readonly DiagnosticDescriptor UnableToFindType = new("GS0157", DiagnosticSeverity.Error, "Cannot find type {0}. Are you missing an import?");
     internal static readonly DiagnosticDescriptor UnableToFindMember = new("GS0158", DiagnosticSeverity.Error, "Cannot find member {0}.");
     internal static readonly DiagnosticDescriptor UnableToFindFunction = new("GS0159", DiagnosticSeverity.Error, "{0}");
--- a/test/Core.Tests/CodeAnalysis/Binding/Issue2851GenericDiagnosticDisplayTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Binding/Issue2851GenericDiagnosticDisplayTests.cs
@@ -213,10 +213,10 @@ public class Issue2851GenericDiagnosticDisplayTests
         Assert.Collection(
             bag,
             diagnostic => Assert.Equal(
-                "Cannot convert type 'Box[int32]' to 'Box[int32]?'. An explicit conversion exists (are you missing a cast?)",
+                "Cannot convert type 'Box[int32]' to 'Box[int32]?'. Provide a value of type 'Box[int32]?' instead.",
                 diagnostic.Message),
             diagnostic => Assert.Equal(
-                "Cannot convert type 'Conv[int32]?' to 'Conv[int32]'. An explicit conversion exists (are you missing a cast?)",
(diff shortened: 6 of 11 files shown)


DavidObando:
## Review — PR #3272 (issue #3267), head `e19c37cfba329d68f2d6317ab6380338ed3ddf7a`

**Verdict: MERGEABLE — 0 BLOCKER · 2 SHOULD-FIX · 2 NIT.**

The fix is real, the three holes close, and I found **zero over-correction**. The one thing I want you to look at before merge is that this PR fixes a second, unstated defect for two of the three call shapes and leaves the third — and the code itself documents those two sites as mirrors.

`main` = `0232ecb5eb7c1f6878556599a6faf557487e4232`. PR merge-base = `d172d0d14`, so main is 2 ahead; `git merge-tree --write-tree --messages` gives tree `77b9d2d48fc3876db3f9aaf433296ce767023177` with **0 stage entries and 0 `CONFLICT` lines**, and neither of main's two commits touches `OverloadResolution/` — no rebase needed. CI 30 SUCCESS / 2 SKIPPED, 0 non-success.

Everything below is from my own worktrees at this head, Release + `-p:ContinuousIntegrationBuild=true`, 0 warnings. Witness `out/bin/Release/Core/GSharp.Core.dll` = `9b5cf580b9db205f7796b5120604a776`, reproduced **identically from two different directories**, and round-tripped exactly after every mutation. Base witness `d01574f734bd386bd454d2d3fafc2bca`. **Oracle is the running program**: `gsc /target:exe` → `dotnet exec`, `rc` captured before any pipe.

---

### 1. Over-correction probe — clean, 11/11

18 programs compiled and run against **base** and **head**. No valid program regressed; nothing went green→red.

| # | program | BASE `d172d0d14` | HEAD `e19c37cfb` |
|---|---|---|---|
| H1 | `Id[int32]("wrong")` free | c0 / **run 134** | **c1 GS0154** |
| H2 | `Box().Id[int32]("wrong")` instance | c0 / **run 134** | **c1 GS0156** |
| H3 | `"r".IdExt[int32]("wrong")` extension | c0 / **run 134** | **c1 GS0156** |
| H4 | `Holder[int32]("wrong")` ctor | c1 GS0154 | c1 GS0154 |
| V1 | `Id[int32](5)` exact | c0/r0 `5` | c0/r0 `5` |
| V2 | `Id[int64](5)` **widening** | c0/r0 `5` | c0/r0 `5` |
| V4 | `Pair[int64](5, 7)` **T in two positions** | c0/r0 `5` | c0/r0 `5` |
| V5 | `Id(5)` inference | c0/r0 `5` | c0/r0 `5` |
| V6 | `Id[string]("ok")` | c0/r0 `ok` | c0/r0 `ok` |
| V7 | `Box().Id[int64](5)` instance widening | c0/r0 `5` | c0/r0 `5` |
| V8 | `Outer[int32]` → `Inner[T](v)` **T still open** | c0/r0 `7` | c0/r0 `7` |
| V9 | `Keep[string](…)` **constrained `T class`** | c0/r0 `constrained` | c0/r0 `constrained` |
| B1 | `Id(5)` | c0/r0 `5` | c0/r0 `5` |
| B2 | `Id[object]("str")` | c0/r0 `str` | c0/r0 `str` |
| **V3** | **`Id[object](5)` free, boxing** | **c0 / run 134** | **c0 / run 134** |
| **B3** | **`Take[object](5)` free, boxing** | **c0 / run 134** | **c0 / run 134** |
| **B4** | **`Box().Id[object](5)` instance, boxing** | **c0 / run 134** | **c0 / r0 `5`** |
| **B5** | **`"r".IdExt[object](5)` ext, boxing** | **c0 / run 134** | **c0 / r0 `5`** |

The forwarding case V8 is the one I most wanted: `Inner[T](v)` inside `Outer[T]` keeps `expectedType` open, so the new predicate must *not* fire. It doesn't.

Three earlier probes were **no-results, not results** and are excluded: `type Box struct` → `GS0306` (ADR-0078), a receiver clause on an owned type → `GS0314` (ADR-0079), and semicolon-separated statements → `GS0005`. Denominator is 18, all recompiled after correction.

---

### 2. SHOULD-FIX 1 — the free-function path still emits invalid IL, and this PR desynchronized a documented mirror

Look at **B4/B5 vs V3/B3**. Same program, four call shapes. At base all four were `rc0` compile → `rc134` run. At head the instance and extension shapes are **fixed** — `Box().Id[object](5)` now prints `5` — while the free-function shape is **unchanged and still broken**. This PR created that asymmetry.

The mechanism is in your own comment. `OverloadResolver.CallBinding.cs:1707` guards the conversion-materialization branch with

```csharp
&& !(substitution != null && parameter.Type is TypeParameterSymbol)
```

and the comment at `CallBinding.cs:1738` says it exists to mirror the instance path:

> only a BARE erased slot (`parameter.Type is TypeParameterSymbol` …) is skipped — **mirroring the equivalent guard in `BindUserInstanceCall`'s per-argument loop (`if (paramType is TypeParameterSymbol) { …; continue; }`)**.

That is exactly the line this PR changed at `OverloadResolver.Invocations.cs:1544`. **The mirror is now broken and the comment still asserts it.** The same paragraph already warns what the skip costs: *"the missing `box` opcode produces IL ilverify rejects (`StackUnexpected: found Int32, expected ref 'object'`)"*.

**I implemented and measured the completion.** One line, the same shape you used at the other two sites:

```diff
             else if (argument.Type != expectedType
-                && !(substitution != null && parameter.Type is TypeParameterSymbol)
+                && !(substitution != null && parameter.Type is TypeParameterSymbol && TypeSymbol.ContainsTypeParameter(expectedType))
```

Measured on top of your head (witness moved `9b5cf580…` → `4cc2668a…`):

- **V3 `Id[object](5)`: run 134 → rc 0, prints `5`.** **B3 `Take[object](5)`: run 134 → rc 0, prints `took`.**
- **All 18 other cells byte-identical** — every hole still rejected with the same id, every valid control still green.
- **`Core.Tests` 7194 passed / 1 skipped / 0 failed — identical to your head**, which I ran separately to compare (also 7194/1/0). Suite-neutral.

- **`Compiler.Tests` 4267 passed / 0 skipped / 0 failed** against the patch — also unchanged.

So the extra line is suite-neutral across both emit and binding suites while closing two live `rc0`→`rc134` holes. I have not run the remaining 7 projects against it. Your call whether it belongs in this PR or a follow-up issue — but the stale mirror comment at `:1738` should not ship asserting a parity that no longer holds.

**The pin that would have caught this:** `ValidCalls` currently has `Id[object]("explicit-object")` — a *string* argument, which needs no boxing, so the corpus cannot see the hole. Change it (or add) `Id[object](41)`. That is the red row: it fails on this head and passes with the line above.

---

### 3. SHOULD-FIX 2 — three ids for one error, and two of them give advice that ICEs the compiler

Same user mistake, four shapes, three diagnostics — measured from `gsc`, id **and** span:

| shape | id | message |
|---|---|---|
| free / constructor | `GS0154` | `Parameter 'value' requires a value of type 'int32' but was given a value of type 'string'.` |
| instance / extension / `shared` | `GS0156` | `Cannot convert type 'string' to 'int32'. An explicit conversion exists (are you missing a cast?)` |
| extension, `[]T` slot | `GS0155` | `Cannot convert type '…' to '…'.` |

`GS0154` names the parameter; the other two do not. Worse, **`GS0156`'s advice is false**. Following it:

```
$ gsc … Prog.gs   # Console.WriteLine(Box().Id[int32](int32("wrong")))
rc=1
Prog.gs(7,23,7,28): error GS9998: NotSupportedException: Conversion from 'string' to 'int32' is not yet supported by the emitter.
```

The compiler asks for a cast and then ICEs on the cast it asked for. `GS0156` is pre-existing and not your code, but before this PR these programs never reached a diagnostic at all — this PR is what selects the voice for two of the three shapes, and issue #3267 explicitly asked for `GS0154`. `BindCallArgumentWithRefKind` already receives the `ParameterSymbol`, so the information needed for `GS0154` is in hand at those sites.

Spans are good everywhere: all four land exactly on the offending argument (`(4,43,4,50)` = 7 cols = `"wrong"`), and the test asserts `Location.Text.ToString(Location.Span)` rather than arithmetic — that is the right assertion and it is why I trust it.

---

### 4. Mutants — independently reproduced, with a positive control

**Positive control first.** A guarded `throw` at the top of the branch the new check unlocks (`CallBinding.cs:1649`, gated on an opaque env compare so `CS0162` doesn't kill the build):

- **PC-F RED and discriminating** — H1 `GS0154` → **`GS9998: PCF-REACHED`**; H2/H3 **unchanged at `GS0156`**; V1/V2 unaffected. So the site is reached, is decisive for the free shape only, and only on mismatch. That also *measures* the "separate paths" claim rather than assuming it.

Then each hunk reverted alone. Witness moved every time; a build failure that left the md5 unchanged was caught by an explicit guard and rerun, not scored.

| mutant | witness | H1 free | H2 inst | H3 ext | B4 | B5 |
|---|---|---|---|---|---|---|
| head | `9b5cf580` | GS0154 | GS0156 | GS0156 | ok | ok |
| **M-F** `:1646` reverted | `eb15eb39` | **c0 / 134** | GS0156 | GS0156 | ok | ok |
| **M-I** `:1544` reverted | `25bc9aba` | GS0154 | **c0 / 134** | GS0156 | **c0 / 134** | ok |
| **M-E** `:1165` reverted | `047691fe` | GS0154 | GS0156 | **c0 / 134** | ok | **c0 / 134** |

A clean diagonal: **no inert hunk, no cross-talk.** Every RED is `rc0` compile with **zero diagnostics** followed by `rc134` — issue #3267's exact signature — not a test that failed to compile. This independently confirms your 3/16 · 1/16 · 2/16 with a running-program oracle; denominator 16 verified (7 invalid + 9 valid, `Issue3267` filter 16/16 pass). Restored, witness back to `9b5cf580…` exactly.

---

### 5. Constructors — claim verified, and the reason is sound

H4 `Holder[int32]("wrong")` gives `GS0154` at **base and head** — already rejected, unchanged, as you said.

I also checked the `shared { }` static shape, which nobody listed: `Util.Id[int32]("wrong")` → `GS0156` at **base and head**. It never had the bug, and the reason is the interesting part — `ExpressionBinder.Access.Accessor.cs:3467` already spells the rule correctly:

```csharp
if (paramType is TypeParameterSymbol typeParamParam
    && (substitution == null || !substitution.ContainsKey(typeParamParam)))
```

"skip only if this type parameter is *not* substituted" — added for issue #1379, long before this PR. So these two shapes are safe **by design, not by accident**.

---

### 6. NIT 1 — the rule has six spellings across ten sites; this PR updated three

`Accessor.cs:3467` is the rung-2 answer: **the correct shape was already in the tree.** Current spellings of "is this slot still open after substitution?":

| spelling | sites | touched? |
|---|---|---|
| `!ContainsTypeParameter(expectedType)` | `CallBinding.cs:1401`, `:1646`, `Arguments.cs:511` | `:1646` only |
| `!(substitution != null && ContainsTypeParameter(parameter.Type))` | `CallBinding.cs:1546`, `:1586`, `:1708` | no |
| `!(substitution != null && parameter.Type is TypeParameterSymbol)` | `CallBinding.cs:1707` | **no — proven live hole, §2** |
| `paramType is TypeParameterSymbol && ContainsTypeParameter(expectedType)` | `Invocations.cs:1544` | yes |
| `ContainsTypeParameter(paramType)` + inner `ContainsTypeParameter(expectedType)` | `Invocations.cs:1151`/`:1165` | yes |
| `is TypeParameterSymbol tp && (substitution == null \|\| !substitution.ContainsKey(tp))` | `Accessor.cs:3467` | no |

You picked `:1646`'s spelling to match `:1401`, which is right and locally consistent. But this is the #3255 shape — one rule, many hand-rolled copies, some incomplete — and §2 is the first copy to be measurably wrong. `:1546` and `:1586` carry the same stale spelling; **I have no specimen for either and am not claiming they are broken** — unmeasured convention debt, flagged only so the next person doesn't have to rediscover the map.

---

### 7. NIT 2 — a second, unmeasured behaviour change inside the same predicate

`:1646` changed from `!(substitution != null && ContainsTypeParameter(parameter.Type))` to `!ContainsTypeParameter(expectedType)`. Besides the intended tightening, it also **loosens** one cell: when `substitution == null` and the declared parameter is open, the old predicate ran the check and the new one skips it. I could not construct a program that reaches that cell — `Core.Tests` is green and V8 covers the adjacent forwarding case — so this is an observation, not a finding. Worth one sentence in the commit message so it is a deliberate consequence rather than a side effect.

---

### Summary

| | |
|---|---|
| **BLOCKER** | none |
| **SHOULD-FIX 1** | `CallBinding.cs:1707` — free-function boxing hole survives; PR fixed the instance/extension twins and broke the mirror documented at `:1738`. One measured line closes it; add `Id[object](41)` to `ValidCalls` as the red row. |
| **SHOULD-FIX 2** | `GS0154` / `GS0155` / `GS0156` for one error; `GS0156`'s "are you missing a cast?" leads to `GS9998`. |
| **NIT 1** | six spellings, ten sites; correct shape already at `Accessor.cs:3467` (#1379). |
| **NIT 2** | unmeasured loosening at `:1646` when `substitution == null`; document it. |

The core of the PR is solid: real defect, real fix, every hunk load-bearing, spans asserted by text, and no over-correction across 11 valid controls including widening, boxing, constrained, multi-position and open-forwarding. Neither SHOULD-FIX is a regression introduced by this change.

DavidObando:
## Re-confirmation — PR #3272, head `38c7b8bd51815bf99ef2a53a2ec71625cb3ec490`

**Verdict: MERGEABLE — 0 BLOCKER · 2 SHOULD-FIX · 2 NIT.**

Bounded re-check of the push that followed the round-one verdict at `e19c37cfb`. The headline question was whether the two red jobs went green by fixing the product or by weakening the assertion. **They went green by fixing the product.** The `Id[object](41)` row was *added*, not removed, `Assert.Empty` is intact, and I RED-proofed it independently.

`main` = `65578d41b2cc887596c1cda95674a4444390f685` (re-fetched immediately before posting, unmoved). All numbers below are mine, from my own worktrees, `dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true -v quiet`, **oracle = compile → `dotnet exec` → rc + stdout, rc captured before any pipe.**

---

### 1. Merge-base diff of the new push

The PR **rebased**: merge-base moved `d172d0d14` → **`7341a6e688b6d4f1602dfaead0aca6fb1004e093`**, so a head-to-head diff reports phantoms. Against its own merge-base the PR is 4 commits, **+632 / −32 across 11 files**.

New since `e19c37cfb`: the `CallBinding` boxing completion; the comment rewrite; the GS0156 reword (`DiagnosticDescriptors.cs:75` + `docs/diagnostics.md:121` + `website/docs/ref/diagnostics.md:121` + `docs/adr/0099`); message-text updates in three pre-existing test files; the `Id[object](41)` row; a new GS0156-message `Fact`; and the 383-line ported `Issue3271` file.

The three pre-existing test files (`Issue2851…:216,219,235`, `Issue2947…:54,85`, `Issue2947…DriverTests`) are **message-text updates tracking the reword only** — every one still `Assert.Equal`s the full message and its id. Not weakened.

Merge check against current `main`: `git merge-tree --write-tree --messages` rc 0, **0 stage entries**. I did not stop there — I built the real merge (`426af52dc`): **0 warnings**, `Issue3267|Issue3271` **27/27**, and all 9 probes below byte-identical to head. Worth noting `main` moved `src/Core/CodeAnalysis/Symbols/TypeSymbol.cs`, which is where `ContainsTypeParameter` lives — the function this fix depends on. I diffed it: **`ContainsTypeParameter` is untouched by `main`.** No semantic conflict.

### 2. Product hunk, located by symbol

In `OverloadResolver.BindUserCall`'s per-argument loop, the bare-erased-slot exemption on the `else if (argument.Type != expectedType …)` branch — currently `OverloadResolver.CallBinding.cs:1706-1709`:

```diff
-                && !(substitution != null && parameter.Type is TypeParameterSymbol)
+                && !(substitution != null
+                    && parameter.Type is TypeParameterSymbol
+                    && TypeSymbol.ContainsTypeParameter(expectedType))
```

That is the boxing repair itself, at the site round one measured — **not a compensating hack elsewhere.** `Invocations.cs:1146-1176` (extension) and `:1515,1544` (instance) hoist `expectedType` and apply the same rule. Nothing else in `src/` changes except the GS0156 string.

**History note:** the hunk was added in `61a2e2a3b`, **removed in `b4963ef25` ("Remove overlapping binder boxing fix")**, then re-added in `38c7b8bd5`. That removal is precisely what made CI red. Net state is correct; squash on merge.

### 3. Test integrity + RED proof

`Issue3267ExplicitGenericArgumentValidationTests.cs:197`:

```csharp
var result = EmittedOracle.Evaluate(source);
Assert.Empty(result.Diagnostics);
Assert.Equal(expected, result.Value);
```

`Assert.Empty` intact, `Assert.Equal` intact. The `Id[object](41)` row was **added** at `:103-110` expecting `41`. `grep` for `Skip=`, `[Fact(Skip`, `[Theory(Skip`, `Assert.True(true)` across both new files → **zero hits**.

RED-proof — reverted the product hunk alone, rebuilt (witness `b5a446391e38b1498ff4d2d1d34afd2e` → `f566c6f1853170d5704f76f32f104d6e`, movement asserted):

```
Failed CompatibleGenericArguments_CompileAndRun(
  source: "func Id[T](value T) T -> value\nId[object](41)", expected: 41)
  Assert.Empty() Failure: Collection was not empty
Failed!  - Failed: 5, Passed: 22, Total: 27
```

**Byte-identical to the CI red you quoted.** The other four REDs are ported `Issue3271` tests. Restored → `git diff HEAD` empty, `git status --porcelain` empty, witness back to `b5a44639…` **exactly**.

### 4. Nine-test port audit

All 9 method names present and unrenamed. **The port is not faithful**, and this was the one thing worth chasing: three `Show[ValueType](…)` rows were demoted from run-assertions to diagnostic-assertions (`Issue3271…Tests.cs:80, :123, :164`), where #3277 expected them to box and print `System.ValueType[]`.

I checked whether that is a regression this PR introduces. **It is not:**

| probe | merge-base `7341a6e68` | head `38c7b8bd5` |
|---|---|---|
| P1 free `Show[ValueType](23)` | compile 0 / **run 134** | `GS0154` |
| P2 instance `.Show[ValueType](33)` | compile 0 / **run 134** | `GS0155` |
| P3 extension `.Show[ValueType](43)` | compile 0 / **run 134** | `GS0155` |
| **P6 NON-GENERIC control** `func Take(value ValueType)`; `Take(23)` | **`GS0154`** | **`GS0154`** |

Those three programs **never worked** — they were `rc0`-compile → `InvalidProgramException`. Converting a silent invalid-IL crash into a compile-time error is an improvement, and P6 shows the generic path now agrees with the non-generic baseline: value-type → `System.ValueType` is simply not classified anywhere yet, which is #3280. The demotion is honest, it is cited in-comment, and the new pins are strong (id + covered span text + exit code + empty output). **4 of the 9 ported tests go RED under the reverted hunk**, so they are load-bearing and pass for the right reason.

### 5. `Fixes` lines

Body lines 56 and 58: **`Fixes #3267`** and **`Fixes #3271`**. Both present, both `Fixes`.

### 6. Over-reach — none found

Full 9-probe A/B, all three trees (merge-base / head / merge-with-main). Head and merge are identical.

| probe | merge-base | head |
|---|---|---|
| P4 `Id[object](41)` | compile 0 / **run 134** | rc 0, `41` |
| P5 `Show[object](41)` reified | compile 0 / **run 134** | rc 0, `System.Object[]` |
| P8 `Show[IComparable](21)` | compile 0 / **run 134** | rc 0, `System.IComparable[]` |
| P7 non-generic `object` param | rc 0, `System.Int32` | **identical** |
| P9 `Show[int32](7)` must stay unboxed | rc 0, `System.Int32[]` | **identical** |

Six of nine were `rc0`→`rc134` at merge-base; all nine are now correct or cleanly rejected, and **nothing that must not box now boxes** (P9 stays `System.Int32[]`). Every generic row uses the reified array-element idiom with a non-generic control alongside.

**#3283 interaction is positive.** A `ref struct` through the repaired free path:

```
ref struct W { var N int32 }
func Show[T](value T) { }
Show[object](W{N: 1})
```

merge-base: compile rc 0 → **run rc 134**. Head: **`GS0219`**. Materializing the conversion *activates* the existing by-ref-like escape guard where the raw-value path silently bypassed it.

**#3280 / #3284 interaction is a real coupling to flag on that PR:** when #3284 lands value-type → `System.ValueType` classification, the three pins at `:80/:123/:164` will begin failing and must be returned to run-assertions.

---

## SHOULD-FIX 1 — the GS0156 reword suppresses a symptom and deletes a hint that is true by construction

Round one measured that GS0156's *"An explicit conversion exists (are you missing a cast?)"* led to `GS9998`. The fix taken was to delete the hint everywhere. Two measurements say that over-corrects:

**a) The hint is true by construction at the reporting site.** `ConversionClassifier.cs:593`:

```csharp
if (!allowExplicit && conversion.IsExplicit)
{
    Diagnostics.ReportCannotConvertImplicitly(diagnosticLocation, expression.Type, type);
```

GS0156 fires *only* when `conversion.IsExplicit`. An explicit conversion always exists at this site by definition.

**b) For the canonical case the cast genuinely works.** `docs/diagnostics.md` cites `int x = 3.14` as *the* GS0156 example. Measured at this head:

```
var x int32 = 3.14        → GS0156: … Provide a value of type 'int32' instead.
var x int32 = int32(3.14) → compile rc 0, run rc 0, prints 3
```

The old advice was correct and actionable here; the new text loses it for the majority case.

**c) The defect that made it look false is untouched.** Still at this head:

```
int32("wrong") → GS9998: NotSupportedException: Conversion from 'string' to 'int32' is not yet supported by the emitter.
```

So `string → int32` is still classified explicitly-convertible while the emitter refuses it. That classifier/emitter disagreement is the real bug; rewording GS0156 hides it and makes it *less* discoverable.

Recommend restoring the cast hint and tracking the disagreement separately, or gating the hint on emitter support. The reword is not *wrong* — it is never false — but it trades a false hint in one shape for a lost true hint in the common shape.

## SHOULD-FIX 2 — three voices, unaddressed and now pinned into tests

Round one's divergence finding was not addressed; only the GS0156 wording changed. The same user error still answers in three ids by call shape, and the ValueType pins now cement it: `:80` expects **GS0154** (free), `:123` and `:164` expect **GS0155** (instance, extension).

My P6 control makes the target concrete: the *non-generic* path emits **GS0154**. The free path therefore matches the language's own baseline and **instance/extension are the outliers.** If the divergence is deliberate it deserves a sentence; if it is routing convenience, GS0154 is the shape to converge on.

## NIT 1 — unreproducible published witness (recurrence of round one's SHOULD-FIX 2)

The body publishes `5ca986f821cd20ff5da360490885c403 → 8b354d3c5b430ac5dfde535d58514d93 → 5ca986f8…` with **no build command and no commit**. Measured on one tree: with `-p:ContinuousIntegrationBuild=true` → `994c8c8196219114f525b82a1404864d`; without it → `554f3bb3d166ec860444bc2e4b269251`. My CI-flag build of `38c7b8bd5` is `b5a446391e38b1498ff4d2d1d34afd2e`. As published, the triple cannot be checked by anyone. Round one asked for the build command alongside the witness; please add it.

## NIT 2 — history churn

`61a2e2a3b` adds the hunk, `b4963ef25` removes it, `38c7b8bd5` re-adds it. The middle commit is what turned CI red. Squash on merge.

---

### Suite numbers (mine)

| | |
|---|---|
| Release build, head | 0 warnings, 0 errors |
| Release build, merge with `main` | 0 warnings, 0 errors |
| `Core.Tests` (head) | **7206 passed, 1 skipped, 0 failed** |
| `Compiler.Tests` (head) | **4267 passed, 0 skipped, 0 failed** |
| `Issue3267｜Issue3271` filter | **27/27** (18 + 9) — a real denominator, not `No test matches` |
| same filter on merge with `main` | **27/27** |
| probes | **9/9** run on all three trees |

The product repair is real, correctly located, and load-bearing under an independent RED-proof whose failure signature reproduces the CI red exactly. The nine tests landed intact, the `ValueType` demotions are justified by measurement rather than convenience, and there is no over-reach. Both SHOULD-FIXes concern diagnostic quality, not correctness.

PR #3274, opened 2026-08-05
Anchor driver-level GS9998 diagnostics

### LINKED ISSUE #3266: [emit][diagnostics] GS9998 internal-error span is hard-coded to (1,1,1,1) in both drivers, so every ICE reports an unanchored location
Found while reviewing PR #3262 (issue #3255). Filing separately because the fix is **one line serving every internal compiler error in the product**, and it deserves its own blast-radius review rather than being smuggled into a parser PR.

## The defect

`src/Compiler/Program.cs:286` hard-codes `(1,1,1,1)` when the driver's top-level handler turns an unhandled exception into `GS9998`. `src/Repl/Program.cs:181` does the same.

An anchored path **already exists**: `src/Core/CodeAnalysis/Compilation.cs:604` builds `new Diagnostic(location, "GS9998", …)` with a real location. The driver-level handler does not use it.

## Why this is worth fixing on its own

Two open issues reported `GS9998` at `(1,1,1,1)` with **completely different root causes** — #3255 (a speculative-parse gap) and #3192 (an emit failure). It was reasonable to read the shared span as a shared cause; it is not. Both spans come from this same line. The unanchored span is what made two unrelated defects look like one, and it gives a user hitting any ICE no idea where in their source to look.

## Required outcome

Anchor the driver-level `GS9998` to a real location where one is recoverable, and make the fallback honest when it is not. `(1,1,1,1)` is a real, valid-looking position in every source file — it points at the first character of the user's program and implies the compiler knows something it does not.

**Fix it once where every caller routes through**, rather than patching the two drivers separately.

## Evidence bar

- The oracle is the **running compiler**: assert the diagnostic id **and the span**, not just the id. An id-only assertion passes both before and after and is worthless here — measured on the sibling issue #3258, where only the span moved and `Core.Tests` ran **7163/7163 green** under a mutant that reverted the fix.
- Produce at least two ICEs with genuinely different origins and show each now reports a span that points at the responsible construct.
- Show the no-location fallback explicitly, and say what it reports.
- A `ToString()`-rendered diagnostic can be **empty** where `.Message` has text. Assert on the structured fields.

DavidObando:
Fixes #3266.

## Fix

Route compiler, REPL, and emit-pipeline internal failures through one shared `Compilation.CreateInternalErrorDiagnostic` path.

- `EmitDiagnosticException.Anchor` supplies the exact syntax span when available.
- Exception chains are traversed to the innermost anchor-bearing emit exception and root cause.
- Exceptions without source context produce an honest location-less `error GS9998: ...`; no synthetic `(1,1,1,1)`.
- `BuildTask` accepts both located and location-less diagnostic headers, preserving location-less `GS9998` as a structured MSBuild error instead of triggering its synthetic fallback.
- Top-level catches remain broad and keep their existing exit behavior.

## Driver evidence

Both `gsc` and file-driven `gsi` produced the same id/span:

| origin | id + rendered span | `Location.Text.ToString(Location.Span)` |
|---|---|---|
| builtin string conversion | `GS9998 (1,1,1,18)` | `print(string(42))` |
| index-from-end ref address | `GS9998 (3,17,3,29)` | `original[^1]` |

The one-based rendering uses inclusive start/exclusive end. No-anchor failures render as `error GS9998: <ExceptionType>: <message>`, without source coordinates.

## SDK/MSBuild evidence

A temporary compiler host deliberately called the real `gsc` `ReportUnhandledException` handler with `InvalidOperationException("no source location")`; the real `BuildTask` then consumed its output through `dotnet msbuild`:

```text
MSBUILD_RC=1
build.proj(5,5): error GS9998: InvalidOperationException: no source location
```

MSBuild supplies its task-node prefix because the structured error has no file; the recorded `BuildErrorEventArgs` has an empty file and zero coordinates. Root cause remains visible, `(1,1,1,1)` is absent, and the `without emitting a structured diagnostic` safety-net message is absent.

`Sdk.Tests` no longer copies the production regex into sample-string tests. It invokes the real `BuildTask.LogCompilerLine`, records MSBuild events, and asserts code, message, file, span, and `HasLoggedErrors`. Its class contract now documents both located and location-less output.

## Wrapper measurement

`Exception("outer one") -> Exception("outer two") -> EmitDiagnosticException(anchor) -> InvalidOperationException("deep failure")` now yields:

- id: `GS9998`
- file/text: `wrapped.gs` / `var x = 1`
- message: `InvalidOperationException: deep failure`

## RED proof

Final-tree product mutants built and changed the relevant product hash:

- location-less SDK header made location mandatory: `BuildTaskSilentFailureTests` RED, 1/2 failed; SDK md5 `9ca292dc...` -> `b23b3dee...`.
- depth-1 exception unwrap restored: `SilentEmitFailureInvariantTests` RED, 1/7 failed (`wrapped.gs` became null); Core md5 `f721921f...` -> `1832a6a6...`.
- Existing six driver/factory mutants remain pinned: fabricated fallback, inverted anchor selection, root-cause loss, wrapped-anchor loss, compiler hard-coded span, and REPL hard-coded span.

Both new mutants were restored and round-tripped GREEN: Core 7/7, SDK 2/2.

## Documentation

`docs/diagnostics.md` now documents both trustworthy anchored output and honest location-less output, and identifies fabricated `(1,1,1,1)` source anchors as bugs.

## Validation

Validated after rebasing onto `main` `6b8096197` (landed #3270 changed only ADR-0153), head `bd58f1be9`:

- Nine solution test projects: **15,373 passed, 0 failed, 4 skipped** (15,377 total).
- All tests used `-c Release -p:ContinuousIntegrationBuild=true` and built normally.
- `Core`, `Compiler`, and `Repl` copies of `GSharp.Core.dll`: md5 `f721921f80187a8c1debb14cc0e3aebd`.
- `Gsharp.NET.Sdk.dll`: md5 `9ca292dca714e20ade9e1e76a2b872d7`.
- `git merge-tree --write-tree origin/main HEAD`: clean (`3a9a2bd69a9374927b63798ce326e415334614ee`).

### DIFF
--- a/docs/diagnostics.md
+++ b/docs/diagnostics.md
@@ -1200,7 +1200,11 @@ run as real CLR finalizers on every driver.
 
 ## Internal compiler error details (GS9998)
 
-`GS9998` is the *silent emit failure* diagnostic. It is always anchored at the user's source file at the location of the expression or statement that triggered the failure. If you ever see `GS9998` anchored at `(1,1,1,1)` against `gsc.dll` instead of your source file, that itself is a bug — please file an issue.
+`GS9998` is the *silent emit failure* diagnostic. When the compiler can recover
+the responsible expression or statement, it reports that source file and span.
+When no trustworthy source anchor exists, it reports `error GS9998: ...`
+without a file or coordinates. A `GS9998` anchored at `(1,1,1,1)` against
+`gsc.dll` or an arbitrary source file is itself a bug — please file an issue.
 
 The message format is `<ExceptionType>: <description>`, for example `InvalidOperationException: Variable 'x' has no local slot`. This tells you what the compiler was trying to do when it failed, even though the underlying bug may not be your fault.
 
--- a/src/Compiler/Program.cs
+++ b/src/Compiler/Program.cs
@@ -192,11 +192,9 @@ public class Program
         catch (Exception ex) when (ex is not OutOfMemoryException and not StackOverflowException)
         {
             // 6.2 SilentEmitFailure invariant (outer ring): any exception
-            // that escapes Compilation.Emit or the compilation setup is
-            // formatted as a canonical GS9998 diagnostic line on stdout so
-            // the SDK BuildTask regex matches it and the IDE error pane
-            // navigates to the source file.
-            return ReportUnhandledException(ex, parsed);
+            // that escapes Compilation.Emit or compilation setup becomes a
+            // GS9998 on stdout, using a carried source anchor when available.
+            return ReportUnhandledException(ex);
         }
     }
 
@@ -261,6 +259,17 @@ public class Program
         return tokens;
     }
 
+    internal static int ReportUnhandledException(Exception ex)
+    {
+        Console.Out.WriteDiagnostics(new[] { Compilation.CreateInternalErrorDiagnostic(ex) });
+        if (System.Environment.GetEnvironmentVariable("GS_DEBUG_STACK") != null)
+        {
+            Console.Out.WriteLine(ex.ToString());
+        }
+
+        return Error;
+    }
+
     private static int ReportFatalIOError(Exception ex)
     {
         // Emit in the csc-compatible "gsc: error GS9997: <message>" form so
@@ -268,27 +277,6 @@ public class Program
         // MSBuild error rather than an opaque process crash.
         var descriptor = DiagnosticDescriptors.FatalCompilerIOError;
         Console.Error.WriteLine($"gsc: error {descriptor.Id}: {string.Format(descriptor.MessageFormat, ex.Message)}");
-        return Error;
-    }
-
-    private static int ReportUnhandledException(Exception ex, CommandLineArgs parsed)
-    {
-        // 6.2 SilentEmitFailure invariant (outer ring): format the exception
-        // as a canonical diagnostic line so the SDK BuildTask regex matches.
-        // Anchor at the first source file when available.
-        var file = parsed?.SourceFiles?.Count > 0
-            ? Path.GetFullPath(parsed.SourceFiles[0])
-            : "gsc";
-
-        var typeName = ex.GetType().Name;
-        var message = $"{typeName}: {ex.Message}";
-
-        Console.Out.WriteLine($"{file}(1,1,1,1): error GS9998: {message}");
-        if (System.Environment.GetEnvironmentVariable("GS_DEBUG_STACK") != null)
-        {
-            Console.Out.WriteLine(ex.ToString());
-        }
-
         return Error;
     }
 
--- a/src/Core/CodeAnalysis/Compilation/Compilation.cs
+++ b/src/Core/CodeAnalysis/Compilation/Compilation.cs
@@ -398,7 +398,7 @@ public class Compilation
         }
         catch (Exception ex) when (ex is not OutOfMemoryException and not StackOverflowException)
         {
-            var diagnostic = BuildEmitFailureDiagnostic(ex);
+            var diagnostic = CreateInternalErrorDiagnostic(ex);
             var combined = allWarnings.Add(diagnostic);
             return new EmitResult(success: false, combined);
         }
@@ -547,7 +547,7 @@ public class Compilation
             // type) produces a structured GS9998 diagnostic anchored at the
             // offending source construct. Previously this only caught
             // NotSupportedException and InvalidOperationException (#519).
-            var diagnostic = BuildEmitFailureDiagnostic(ex);
+            var diagnostic = CreateInternalErrorDiagnostic(ex);
             var combined = allWarnings.Add(diagnostic);
             return new EmitResult(success: false, combined);
         }
@@ -567,37 +567,25 @@ public class Compilation
     public EmitResult Emit(Stream peStream, Stream refStream) =>
         Emit(peStream, pdbStream: null, refStream, docStream: null, assemblyName: null);
 
-    // Issue #519 + 6.2 SilentEmitFailure invariant: anchor the synthetic
-    // GS9998 at the best available source location. Preference order:
-    // 1. EmitDiagnosticException.Anchor (precise node that triggered the failure)
-    // 2. First syntax tree root (file-level fallback)
-    // 3. File-named empty SourceText (guarantees the SDK regex matches)
-    // The message includes the exception type name so the user can distinguish
-    // different ICE flavors without needing a stack trace.
-    private Diagnostic BuildEmitFailureDiagnostic(Exception ex)
+    /// <summary>
+    /// Creates a GS9998 diagnostic for an unexpected compiler exception.
+    /// </summary>
+    /// <param name="ex">The exception that escaped the compiler pipeline.</param>
+    /// <returns>
+    /// A diagnostic anchored at an <see cref="EmitDiagnosticException"/> source
+    /// node when available; otherwise a location-less diagnostic.
+    /// </returns>
+    internal static Diagnostic CreateInternalErrorDiagnostic(Exception ex)
     {
-        // Unwrap EmitDiagnosticException to get the original exception type
-        // name in the message and the precise anchor.
-        var anchor = (ex as Emit.EmitDiagnosticException)?.Anchor
-            ?? (ex.InnerException as Emit.EmitDiagnosticException)?.Anchor;
+        ArgumentNullException.ThrowIfNull(ex);
 
-        TextLocation location;
-        if (anchor != null)
-        {
-            location = new TextLocation(anchor.SyntaxTree.Text, anchor.Span);
-        }
-        else
-        {
-            var firstTree = SyntaxTrees.FirstOrDefault();
-            var sourceText = firstTree?.Text
-                ?? SourceText.From(string.Empty, fileName: "gsc");
-            location = new TextLocation(sourceText, new TextSpan(0, 0));
-        }
-
-        // Build the message: include the root-cause exception type and message.
-        var rootEx = ex is Emit.EmitDiagnosticException ede && ede.InnerException != null
-            ? ede.InnerException
-            : ex;
+        var emitException = ex as Emit.EmitDiagnosticException
+            ?? ex.InnerException as Emit.EmitDiagnosticException;
+        var anchor = emitException?.Anchor;
+        var location = anchor is null
+            ? default
+            : new TextLocation(anchor.SyntaxTree.Text, anchor.Span);
+        var rootEx = emitException?.InnerException ?? ex;
         var typeName = rootEx.GetType().Name;
         var message = $"{typeName}: {rootEx.Message}";
 
--- a/src/Core/CodeAnalysis/Compilation/Compilation.cs
+++ b/src/Core/CodeAnalysis/Compilation/Compilation.cs
@@ -579,13 +579,21 @@ public class Compilation
     {
         ArgumentNullException.ThrowIfNull(ex);
 
-        var emitException = ex as Emit.EmitDiagnosticException
-            ?? ex.InnerException as Emit.EmitDiagnosticException;
-        var anchor = emitException?.Anchor;
+        Emit.EmitDiagnosticException anchoredException = null;
+        var rootEx = ex;
+        for (var current = ex; current is not null; current = current.InnerException)
+        {
+            rootEx = current;
+            if (current is Emit.EmitDiagnosticException { Anchor: not null } emitException)
+            {
+                anchoredException = emitException;
+            }
+        }
+
+        var anchor = anchoredException?.Anchor;
         var location = anchor is null
             ? default
             : new TextLocation(anchor.SyntaxTree.Text, anchor.Span);
-        var rootEx = emitException?.InnerException ?? ex;
         var typeName = rootEx.GetType().Name;
         var message = $"{typeName}: {rootEx.Message}";
 
--- a/src/Core/Core.csproj
+++ b/src/Core/Core.csproj
@@ -18,6 +18,7 @@
     <InternalsVisibleTo Include="GSharp.Interpreter.Tests" />
     <InternalsVisibleTo Include="GSharp.LanguageServer" />
     <InternalsVisibleTo Include="gsc" />
+    <InternalsVisibleTo Include="gsi" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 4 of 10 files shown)


DavidObando:
## ⛔ BLOCKER (1) — plus 2 SHOULD-FIX, 2 NIT

Verdict against **`main` = `0232ecb5eb7c1f6878556599a6faf557487e4232`** (re-fetched immediately before posting; merge-base **equals** main, `merge-tree --write-tree` → **0 stage entries**), head **`5e8a9238703e45e31f8e79450786205ad4ae3e73`**. Own worktree, `dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true` → **0 Warning(s), 0 Error(s)**, `GSharp.Core.dll` md5 **`1f970c0582c789df6287e173ab0edc0e`**.

### Credit where due — the shape of this fix is right

The drivers **do not duplicate** the anchoring rule. `BuildEmitFailureDiagnostic` (private instance) became `Compilation.CreateInternalErrorDiagnostic` (`Compilation.cs:578`), and there is now **one** implementation with four call sites (`Compilation.cs:401`, `:550`, `Compiler/Program.cs:264`, `Repl/Program.cs:201`). That is ponytail rung 2, correctly taken. **Both** drivers are fixed — not one.

I probed the handler with four exception shapes; **it never threw, never hung, and always returned rc 1**. And the no-anchor fallback is **honest** — no fabricated coordinates. Eliminating the old "first syntax tree" fallback also removes the *wrong-file* risk, which is the right call.

The pins are **deliberately-injected** exceptions, not naturally-occurring crashes, so they will not rot when the underlying ICEs get fixed. Correct choice. **RED-proof:** I replaced the `default` fallback at `Compilation.cs:585-586` with a fabricated `SourceText.From("", "gsc")` location, rebuilt (md5 moved `1f970c05…` → `563422ea…`) and all **3/3 went RED for the right reason** — `Expected: 2 / Actual: 0` and `Expected: 17 / Actual: 0` on the span assertions, `Assert.Null() Failure` on the honesty assertion. Restored; `git diff HEAD` empty, md5 back to `1f970c05…`.

---

## ⛔ BLOCKER — a no-anchor `GS9998` no longer matches the SDK's own diagnostic regex

`src/Core/CodeAnalysis/Compilation/Compilation.cs:585-586` + `src/Sdk/Gsharp.NET.Sdk/BuildTask.cs:424`

The deleted helper had **three** tiers; the new one has two. The tier whose comment read *"File-named empty SourceText (**guarantees the SDK regex matches**)"* is gone. I measured the consequence by driving the **real** `ReportUnhandledException` of both drivers and applying **`BuildTask.DiagnosticLine` verbatim**:

| build | no-anchor GS9998 line on gsc stdout | SDK regex |
|---|---|---|
| **PR head** | `error GS9998: InvalidOperationException: no source location` | **MATCH=False** |
| mutant w/ old-style location | `gsc(1,1,1,1): error GS9998: InvalidOperationException: no source location` | **MATCH=True** |
| PR head, **anchored** | `anchor_test.gs(3,1,3,10): error GS9998: NotSupportedException: unsupported node` | MATCH=True |

The regex requires `file(\d+,\d+…)`; a location-less line cannot match. Per `BuildTask.cs:424-429` an unmatched line becomes `LogMessage(MessageImportance.Low)` — **not** an error, invisible at default verbosity — so `HasLoggedErrors` stays false and the `BuildTask.cs:356` safety net fires instead, logging a generic GS9998 **anchored at `Compile[0]` with `(1,1,1,1)`**.

**Impact:** on the MSBuild surface — how real projects build — issue #3266's symptom is **not fixed, it is relocated**. The user still gets `(1,1,1,1)`, now wrapped in *"gsc exited with code 1 without emitting a structured diagnostic"*, and the specific exception survives only as a breadcrumb. It is also inconsistent: every other gsc diagnostic carries a file (I measured `…/ice1.gs(5,35,5,38): error GS0125: …`).

**CI cannot see this.** `Sdk.Tests` contributes only **2** tests here and both compare a **hand-written sample string** to the regex — they never invoke the product. This PR *changed those sample strings* from the location-less-equivalent form to anchored ones (`ProgramSilentFailureTests.cs:52`), and rewrote the `BuildTaskSilentFailureTests` docstring that used to promise *"verifying that the diagnostic line on stdout carries the source file name."* The guard was edited to match the new behaviour rather than exercised against it.

**Recommended fix (smallest):** teach `LogCompilerLine` to recognise a location-less `error GS####:` header and `LogError` it anchored at `Compile[0]`. ~6 lines, one place, keeps the honest location-less CLI rendering. *Alternative:* `CreateInternalErrorDiagnostic(Exception ex, string fallbackFile = null)` with gsc passing `parsed.SourceFiles[0]` and gsi passing `scriptPath` — note `Repl/Program.cs` **dropped `scriptPath`**, so `gsi <file>` lost its filename too.

---

## SHOULD-FIX 1 — anchor and root cause are lost when wrapped more than one level

`src/Core/CodeAnalysis/Compilation/Compilation.cs:581-588`

Unwrapping checks only `ex` and `ex.InnerException`. Measured, same harness:

| exception shape | reported |
|---|---|
| `EmitDiagnosticException` | `anchor_test.gs(3,1,3,10): … NotSupportedException: unsupported node` ✅ |
| wrapped **once** | `anchor_test.gs(3,1,3,10): … NotSupportedException: unsupported node` ✅ |
| wrapped **twice** | `error GS9998: Exception: o1` — **anchor gone, root cause gone** ❌ |

The depth-1 limit is pre-existing, but its *cost* is new: previously a missed anchor still fell back to the first syntax tree's filename; now it falls back to nothing. The PR explicitly tests the one-level wrap, so wrap-tolerance is clearly intended — it just stops one level short. **Fix:** walk the `InnerException` chain instead of checking two levels.

## SHOULD-FIX 2 — `docs/diagnostics.md:1202` is now false

> "`GS9998` … **is always anchored at the user's source file** at the location of the expression or statement that triggered the failure."

The PR makes location-less GS9998 the documented, tested behaviour, and touches **no** docs. **Mirror-pair audit: clean** — `website/docs/ref/diagnostics.md:306` describes GS9998 without any anchoring claim, so only the one file needs correcting.

## NITs
1. `test/Interpreter.Tests/Issue3266InternalErrorSpanTests.cs` tests the **Compiler and Repl** drivers, not the interpreter. Understandable if it is the only project referencing both, but the location is surprising.
2. `src/Core/Core.csproj` adds `InternalsVisibleTo gsi`, widening Core's internal surface to a second driver to reach one factory.

## Cross-PR check for #3262 — **no collision**
#3262's diff contains exactly one GS9998 reference, `Assert.DoesNotContain("GS9998", result.Stdout, …)` — it asserts the ICE is *gone*, which is compatible with any GS9998 format. No `(1,1,1,1)` assertion in its diff. Repo-wide, **no existing test asserts a driver-level GS9998 span**: all 9 `1,1,1,1` hits are doc comments, unrelated `cs2gs` payload strings, this PR's own `DoesNotContain` guards, or `Compiler/Program.cs:307` (a **GS9100 warning**, different diagnostic). **These two PRs can land in either order.**

DavidObando:
## ✅ BLOCKER cleared — 1 SHOULD-FIX (new, introduced by the round-2 fix), 3 NIT

Verdict against **`main` = `f5642aff1598630710f29737baf3416273b0dcf4`** (re-fetched immediately before posting), head **`bd58f1be9917422ec1316c9b48b7f84bafd0e1a3`**. Merge-base `6b809619` is behind main (#3250 landed since); `merge-tree --write-tree` read **by stages** → **0 stage entries**. Build `dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true` → rc 0, **0 Warning(s) / 0 Error(s)**, `GSharp.Core.dll` md5 **`f721921f80187a8c1debb14cc0e3aebd`**.

**Every number below is my own. I re-measured from zero — none of round one's figures were reused.**

---

### 1. The BLOCKER is genuinely fixed — proven end-to-end, not by transcript

I injected an env-gated location-less `throw` into `Compilation.Emit`, rebuilt (md5 `f721921f…` → `9f832b44…`), and drove the **real `BuildTask.Execute()`** — real child `gsc` process, real stdout pump, real parser, real safety net — with a recording `IBuildEngine`.

| | **PR head** | **same tree, regex reverted to pre-PR** |
|---|---|---|
| `Execute()` returns | `False` | `False` |
| `HasLoggedErrors` | **True** | **True** (but only via the safety net) |
| errors logged | **1** | 1 |
| code | `GS9998` | `GS9998` |
| file / span | `[]` / **`(0,0,0,0)`** | `/tmp/e2e3274/prog.gs` / **`(1,1,1,1)`** |
| message | **`InvalidOperationException: injected location-less ICE`** | `gsc exited with code 1 without emitting a structured diagnostic. Last compiler output: Failed.` |
| the real ICE line | *is the error* | demoted to `MSG(Low)` — invisible at default verbosity |

Control (no injection) on both builds: `Execute()`=`True`, 0 errors, `Wrote …prog.dll`.

So: line classified as a structured **error** ✅, `HasLoggedErrors` **true** ✅, build **fails** ✅, **root cause survives** ✅, safety net **does not fire** and no `(1,1,1,1)` anywhere ✅.

**One correction to my own round-1 report:** I *derived* that the safety-net breadcrumb would carry the real GS9998 text. Measured, it carries **`Failed.`** — stderr is not empty on this path. The pre-fix loss was therefore **worse** than I claimed: the root cause vanished entirely. Derivation lost to measurement; the fix is more valuable than I said.

### 2. Deep-wrap: fixed well past what was asked

`CreateInternalErrorDiagnostic` now walks the whole `InnerException` chain. Measured directly:

| chain | file | root-cause message |
|---|---|---|
| depth 1 → 5 wrappers around an anchored `EmitDiagnosticException` | `outerA.gs` (all five) | `InvalidOperationException: deep failure` (all five) |
| **anchor at OUTER, inner has none** | `outerA.gs` ✅ | `InvalidOperationException: deep failure` |
| anchor at outer, inner is `EDE` with **null** anchor | `outerA.gs` ✅ | correct |
| **two** anchors (outer `outerA.gs`, inner `innerB.gs`) | `innerB.gs` — innermost wins | correct |
| no anchor at all | `<null>`, no span | correct |

The anchor-at-outer case is the one a naive "recurse to innermost" gets wrong. This implementation keeps the outermost anchor when no inner one exists. Correct.

### 3. Fixtures are stronger, and 1-of-2 is the *right* kill count

The old pair copied the regex into the test file and matched it against hand-written strings — they never touched the product, which is exactly why round one shipped green. The replacement instantiates the real `BuildTask` with a real `IBuildEngine` and calls the real `LogCompilerLine`, asserting `HasLoggedErrors`, code, file, **all four coordinates**, and message. Categorically stronger.

I reproduced the mutant myself: reverting the regex gives **`Failed: 1, Passed: 1`**, and the failure is `LogCompilerLine_LocationlessGS9998_LogsStructuredErrorWithoutInventedSpan` — `Assert.Single() Failure: The collection was empty`. **1-of-2 is correct, not weak**: the located test is a *control* that must stay green to prove anchored parsing was not broken. A mutant killing both would mean there was no control.

The restored docstring is true of the new code — I verified each clause behaviourally.

### 4. Full suite, 9 of 9

`dotnet test GSharp.sln -c Release`, rc **0**: InternalAnalyzers 9, GeneratorHost 31, LanguageServer 468, Sdk 59, Extensions 104, Interpreter 1349/3 skipped, Core 7179/1 skipped, Cs2Gs 1907, Compiler 4267 — **15,373 passed, 4 skipped**. Matches the reported figures exactly.

---

## 🟡 SHOULD-FIX — the loosened regex now misclassifies prose as build errors

`src/Sdk/Gsharp.NET.Sdk/BuildTask.cs:32` — making the whole `file(l,c):` prefix optional leaves `(?<code>[^:]+)` free to swallow whitespace, so any line shaped `error <words>: <text>` becomes a structured diagnostic. Measured through the real `LogCompilerLine`:

| input line | classified as | code captured | `HasLoggedErrors` |
|---|---|---|---|
| `error handling failed: retrying` | **ERROR** | `handling failed` | **True** — fails the build |
| `warning signs detected: check config` | **WARNING** | `signs detected` | False |
| `info about the build: nothing to do` | MESSAGE(Normal) | — | False |

All three matched `OLD=no → NEW=MATCH`, so this is **introduced by this PR**, not pre-existing.

**Reachability is low, and I want to be honest about that**: `/out:` is always passed (`BuildTask.cs:141`), so gsc compiles and never runs user code — gsc's stdout is machine-generated. The realistic vector is a multi-line exception message inside a `GS9998`, whose continuation lines are fed to `LogCompilerLine` individually. That is why this is SHOULD-FIX and not BLOCKER.

**Fix is two characters** — `[^:]+` → `[^:\s]+`. I validated it against the full corpus: it rejects all three prose lines and still matches **every** intended form (anchored 4-coord, anchored 2-coord, location-less error, location-less warning), extracting the correct code each time.

Good news on the severity question you'd expect me to check: **warnings are not promoted.** Anchored and location-less warnings stay warnings, and `GS9100` stays `MESSAGE(Low)` in both forms.

## NITs
1. **No negative or warning coverage in the new SDK tests.** `RecordingBuildEngine` implements `LogWarningEvent`/`LogMessageEvent` as empty bodies with no lists, so nothing can assert that a warning stays a warning or that a non-diagnostic line stays `MessageImportance.Low`. Given the regex was *loosened*, a rejection test is the one most worth having. Adding `Warnings`/`Messages` lists is a few lines.
2. `src/Core/Core.csproj:21` — `InternalsVisibleTo gsi` widens Core's internal surface to a second driver to reach one factory.
3. `src/Compiler/Program.cs:307` — the **GS9100** warning still hard-codes `(1,1,1,1)`. Harmless today (I measured it demoted to `MessageImportance.Low`), but it is now the last hard-coded span of its kind in a driver.

## Mirror-pair audit — clean, and re-checked against *current* main
`docs/diagnostics.md:1199` is corrected accurately and covers **both** the anchored and location-less cases. #3274 touches no `website/` file, so I re-verified the twin on today's main: `website/docs/ref/diagnostics.md:306` is a single table row with **no anchoring claim**, and the "Internal compiler error details (GS9998)" narrative section does **not exist** in the mirror at all (grep count 0). Single-file fix is correct. Note main's #3250 did touch both diagnostics files since this branch forked, but only the **GS0519** wording, symmetrically in both.

## PR #3276 collision — none
`merge-tree --write-tree` read by stages: `main × 3276` → 0 entries, `3274 × 3276` → 0 entries. Stronger than that, #3276's 128 files include **no** `docs/diagnostics.md` and **no** `website/docs/ref/diagnostics.md` — the only name matching `diagnostic` is `test/Interpreter.Tests/Issue3034NullableNarrowingDiagnosticEmittedOracleTests.cs`. They do not overlap.

Issue #1035, opened 2026-06-24
G# unsafe pointers: function pointers and fixed-size buffers (follow-up to #1014)


DavidObando:
## Summary
Follow-up to #1014. **Function pointers** (C# `delegate*<...>` / CLR `ELEMENT_TYPE_FNPTR`) and **fixed-size buffers** (C# `fixed byte buf[16]`) are deferred from the unsafe-pointer CORE.

## Desired behavior
- **Function pointers**: an unsafe-context type denoting an unmanaged/managed function pointer, callable directly (`calli`), usable as a field/param, and obtainable from a static method (`&Method`). Decide the surface spelling.
- **Fixed-size buffers**: an unsafe struct field `fixed T name[N]` laying out `N` inline elements, addressable as a `*T` to the first element.

## Notes
- Both are separable from the raw-pointer CORE and from each other; this issue tracks them together as low-priority unsafe-interop extensions. Split if either is picked up independently.
- See ADR-0122 deferrals.

PR #3275, opened 2026-08-05
Fix nil conversion for function pointers

### LINKED ISSUE #3268: [binder] Pointer-to-legacy-func type '*func(...)' bypasses both the GS0398 pointee check and the GS0303 deprecation warning, then ICEs at emit
Measured on a build of `3e6dd50a7` (pre-dates `e28d55bdf` by three doc/test-only merges; the binder paths involved were untouched).

## The discriminator: the arrow form is validated, the deprecated form is not

ADR-0075 replaced the `func(...)` type clause with the arrow form `(...) -> R`. **The pointee-type validation only sees the arrow form.**

| # | source | rc | result |
|---|---|---|---|
| d | `var f *((int32) -> int32) = nil` | 1 | ✅ `GS0398: Unmanaged pointer to '(int32) -> int32' is not supported; the pointee must be a blittable primitive, a blittable value struct, or another pointer (ADR-0122).` |
| b | `var f *func(int32) int32 = nil` | 1 | ❌ `GS9998: NotSupportedException: Conversion from 'nil' to '*func(int32) int32' is not yet supported by the emitter.` |
| g | `var f *func(int32) int32` (no initializer) | **0** | ❌ **compiles clean, writes a dll** |
| h | `unsafe func Take(f *func(int32) int32)` + `Take(nil)` | 1 | ❌ same `GS9998` at `(4,22,4,31)` |

**`d` and `b` declare the same type two ways and get opposite treatment.** `GS0398` is the correct answer for both. The `GS9998` in `b`/`h` is the unvalidated type reaching an emitter that rightly cannot encode it — the ICE is the *second* symptom, not the defect.

Probe `g` is the sharpest: it accepts the type, emits an assembly, and `dotnet g.dll` **runs to rc 0**. So the binder is not merely late to reject this type — it does not reject it at all.

## Second symptom, same cause: the deprecation warning is silent too

```
grep -c GS0303 → a.out:1   b.out:0   g.out:0
```

- `var f func(int32) int32 = nil` (**not** a pointer) → `GS0303: 'func(...)' as a type clause is deprecated; use the arrow form '(...) -> R' instead (ADR-0075).`
- `var f *func(int32) int32` → **no GS0303 at all**

So the pointer form of the deprecated syntax escapes *both* the deprecation warning and the pointee check. That is strong evidence for a **single shared cause** — a type-clause path that handles `*` before or instead of the analysis that both diagnostics hang off — rather than two independent oversights. **Find that one path.** If it turns out to be two genuinely separate gaps, say so with evidence.

## Required outcome

`*func(...)` must be diagnosed exactly as `*((...) -> R)` is: **GS0398**, plus the **GS0303** deprecation warning that the non-pointer form already emits. No path should reach the emitter with this type.

## Scope before concluding

Check whether other type-clause forms reach the pointee check only via the arrow spelling — `map`, `sequence`, `chan`, tuple types, nested/generic pointees. A related parser defect (#3255) turned out to be exactly this shape: a token set hand-rolled **four separate times**, three of them incomplete. If pointee validation is similarly spelled more than once, **that** is the fix.

## Evidence bar

- **The oracle is the running program**, and probe `g` shows why: rc 0 plus a dll that runs is not proof of correctness. Do not accept a clean compile as a pass for a case that should be rejected.
- **Not ILVerify** — caught this session returning rc=0 while printing `Assembly or module not found: mscorlib`, and green-lighting an assembly the CLR refuses to JIT (#3240, open).
- Assert the diagnostic **id and span**, not just the id.
- **Do not over-correct**: legal pointers must keep working. `var p *int32 = nil` and `var p *int32` both compile rc 0 today and must continue to. Pin them.
- **A no-result is not a result.** One probe in this investigation (`*func()` with a trailing space before a newline) produced `GS0268: Type 'Console' does not contain a nested type named 'WriteLine'` — a parse artifact that reads like a real finding. It was discarded. Read your diagnostics, do not pattern-match them.
- Re-verify all four rows on current `main` before starting; my measurement build is three merges behind.

DavidObando:
Fixes #3268.

## Summary

- Preserve the shipped ADR-0122 managed function-pointer syntax, `*func(T1, T2) R`, including parameter/return/field spellability and cs2gs mapping.
- Extend the existing `nil`-to-pointer emission path to `FunctionPointerTypeSymbol`, materializing null managed and unmanaged function pointers as native-int zero.
- Pin exact `GS0404` and `GS0398` spans, verify `GS0303` stays silent for `*func(...)`, and verify `GS0154` renders the source-spellable managed-pointer type.
- Document `nil` as the null function-pointer value in ADR-0122 and the language specification.

The `GS0404` gate and the `nil` → `FunctionPointerTypeSymbol` conversion classification both shipped with issue #1035 in `5bdf22934` and remain unchanged. Issue #3268 exposed the emitter gap after that valid conversion: only `PointerTypeSymbol` reached the shared native-zero path. This PR fixes that emitter path and adds exact regression coverage for the existing gate.

## Corrected matrix

| Row | Current main | This PR |
|---|---|---|
| pointer to arrow delegate | `GS0398` | `GS0398` |
| initialized top-level `*func` | `GS0404` | `GS0404` |
| uninitialized top-level `*func` | `GS0404` | `GS0404` |
| `nil` passed to unsafe `*func` parameter | `GS9998` | rc 0; loads and prints `0` |

All four rows produce zero `GS0303` diagnostics.

## Verification

- `FunctionPointer_PassAsParameter_CompilesAndRuns`: passes and prints `30` / `42`.
- Managed and unmanaged function-pointer `nil` parameter tests compile and run, printing `0`.
- Nine test projects, enumerated from `GSharp.sln`: 15,399 passed, 4 skipped, 0 failed.
- Product-hunk revert killed both `FunctionPointer_NilParameter_CompilesAndRuns` theory cells with the original `GS9998`.
- Control mutants killed the `GS0404`, `GS0398`, parser/`GS0303`, and source-spelling pins.


DavidObando:
## Rubber-duck review — **BLOCKER** · 1 BLOCKER · 2 SHOULD-FIX · 3 NIT

Reviewed `b2f73247c` against `main` **`6b809619779b4403062a6a9540663202f504fb73`** (re-fetched immediately before posting). Merge-base `0232ecb5e`; `merge-tree --write-tree` rc 0, single tree OID, **no stage entries and no `CONFLICT` line** — merges cleanly.

**The diagnostic fix itself is correct and the causal story is honest.** I reproduced all four rows, confirmed the single shared cause by single-hunk revert, confirmed row `g` is genuinely dead, and reproduced the author's suite numbers to the unit. My blocker is not about the diagnostics. It is about **what was traded to get them**.

---

## BLOCKER — this PR deletes a shipped, tested, ADR-documented language feature, and the issue never asked for it

`Parser.TypeClauses.cs:99-109` — the removed branch was not an accident. It was **ADR-0122 §9 / issue #1035**: the managed function pointer, spelled `*func(T1, T2) R`, invoked via `calli`. Issue #3268 never mentions #1035, and the PR title ("Fix legacy function pointer pointee validation") and body do not announce a feature removal. A reviewer reading either would not know one had happened.

**Measured, by execution, not inspection.** The source of the test this PR deletes — `Issue1035FunctionPointerEmitTests.FunctionPointer_PassAsParameter_CompilesAndRuns` — on `main`:

```
compile rc=0 → dotnet dp.dll → rc=0, stdout "30\n42"
```

Byte-identical source on this branch:

```
rc=1
(12,23,12,27): warning GS0303: 'func(...)' as a type clause is deprecated...
(12,23,12,47): error   GS0398: Unmanaged pointer to '(int32, int32) -> int32' is not supported...
(13,12,13,14): error   GS0131: 'fp' is not a function.
(17,29,17,33): error   GS0154: Parameter 'fp' requires a value of type '*(int32, int32) -> int32'
                                but was given a value of type '*func(int32, int32) int32'.
```

**After this PR the managed function-pointer type is inhabitable but unspellable.** I probed every alternative:

| attempt to declare a function-pointer **parameter** | result |
|---|---|
| `fp *func(int32, int32) int32` | GS0303 + GS0398 |
| `fp *((int32, int32) -> int32)` | GS0398 (pointer-to-delegate) |
| generic `apply[T](fp T, …)` | GS0131 `'fp' is not a function` |
| `fp unmanaged[Cdecl] (int32, int32) -> int32` | GS0154 type mismatch |
| **`let fp = &add`** (local, inferred) | **rc 0, runs, prints 7** ✅ |

Same for **return types** (`unsafe func get() *func(…) …` → GS0303+GS0398) and **struct fields** (→ GS0303+GS0398). So you can create a function pointer and call it in the same scope, and that is all. Passing a callback across a function boundary without allocating — the reason function pointers exist — is gone. That is why two `Issue1035` tests had to be deleted rather than adapted.

The blast radius reaches `cs2gs`: `CSharpTypeMapper.cs:509` now reports C# `delegate* managed<…>` in any non-local position as **Unsupported/ByDesign** with rationale `NoGsharpConstruct` — true only because this PR removed the construct. A previously-translatable C# category became untranslatable.

Process: `docs/adr/0122-…md` is **Status: Accepted**. Its §9 is edited in place to remove the feature, and `GS0404` is marked **Retired**, with no superseding ADR.

**Options, smallest first — the owner's call, not mine:**
1. **Re-spell instead of remove.** Keep the production, give it an ADR-0075-consistent name symmetric with the existing `unmanaged[CC] (T) -> R` (e.g. `managed (T) -> R`). Then `*func(…)` is unambiguously legacy syntax, gets GS0303 + GS0398 exactly as #3268 demands, **and** #1035 keeps a spelling. Satisfies both issues.
2. **Land the removal deliberately** — retitle the PR, say so in the body, supersede ADR-0122 §9, and accept the cs2gs coverage loss.
3. Land as-is. Then #1035 is silently half-retired by a PR nobody reads as a language change.

---

## SHOULD-FIX 1 — the docs sweep is incomplete, and the two halves now contradict each other

`website/docs/ref/spec.md:403` (**live**, hand-maintained, last touched by #3207 and #3261) still says:

> **Managed function pointers** are spelled `*func(T1, T2) R`: only legal inside `unsafe` (`GS0404`), obtained from a static/free method with `&Method` … invoked directly with `fp(args)` (emits `calli`) …

…while `website/docs/ref/diagnostics.md:282`, **in the same tree, changed by this PR**, says GS0404 is `**Retired.**` and the production was removed. Two files in one docs directory that disagree. `website/versioned_docs/version-0.3/**` is a frozen snapshot (last touched by the deliberate re-snapshot in #2277) and is correctly excluded.

Credit where due: `docs/diagnostics.md` ↔ `website/docs/ref/diagnostics.md` **are** kept identical here — that mirror was handled correctly.

Also stale: `tools/cs2gs/corpus/grid/G12-Unsafe-Console/Program.cs:29` still says managed `delegate*` "now map to G#'s managed `*func(T) R`".

## SHOULD-FIX 2 — the compiler still prints a spelling it now refuses to parse

`SymbolDisplay.FormatFunctionPointerType` (`src/Core/CodeAnalysis/Symbols/Display/SymbolDisplay.cs:875-879`) renders a managed function pointer as `*func(params) ret`. That is now un-parseable, and it reaches users: the GS0154 above tells the programmer their value has type `*func(int32, int32) int32` — a syntax error. Diagnostics must not name types the language cannot express. If the removal stands, this needs a spelling that round-trips.

---

## Verification of the PR's own claims — all reproduce

**Four-row matrix, re-measured on both trees** (`main` behaviour obtained by reverting the parser hunk alone; `main`'s only other delta from the merge-base is a doc):

| # | source | `main` | this branch |
|---|---|---|---|
| d | `var f *((int32) -> int32) = nil` | rc 1, GS0398 | rc 1, GS0398 — **unchanged** ✅ |
| b | `var f *func(int32) int32 = nil` | rc 1, **GS9998**, no 303/398 | rc 1, GS0303+GS0398, **no GS9998** ✅ |
| g | `var f *func(int32) int32` | **rc 0, DLL written** | rc 1, GS0303+GS0398, **no artifact** ✅ |
| h | `unsafe func Take(f *func(int32) int32)` | rc 1, **GS9998**, no 303/398 | rc 1, GS0303+GS0398 ✅ |

**Row `g` is truly dead, not merely diagnosed.** `gsc /out:g_check.dll g.gs` → rc **1**, and `ls g_check.dll` → *No such file or directory*. Run-in-place also rc 1.

**One shared cause — confirmed.** Reverting **only** `Parser.TypeClauses.cs` (md5 `ed3cf5fa…` → `613736928…`, so a real rebuild; restore returned it to `ed3cf5fa…` with `git diff HEAD` empty) drops **GS0303 and GS0398 to zero together** on b/g/h and brings GS9998/rc-0 back. They move as one. The PR's story is right.

**Over-correction battery — clean.** The non-pointer legacy form and the arrow form are untouched in every position:

| position | legacy `func(int32) int32` | arrow `(int32) -> int32` |
|---|---|---|
| local, param, return, field, generic arg, seq, nested | identical on both trees; **exactly one GS0303** per legacy clause (two for the nested `func(int32) func(int32) int32`, correct) | identical on both trees, **zero** GS0303 |

**Third-bypass hunt — none found in pointee position.** `**func`, `***func`, `[]*func`, `[][]*func`, struct field, return type, `*func()` (zero-param), `*func(int32)` (zero-return) all report GS0303 + GS0398 on this branch and none on `main`. The class is closed structurally: `*` now always recurses through `ParseTypeClause()`.

**New GS0303 noise — zero.** `git grep '\*func' -- '*.gs'` returns **no hits** repo-wide (positive control: `git grep -c func -- samples/*.gs` finds plenty). No sample, e2e, corpus or test `.gs` uses the removed spelling, so nothing outside the suite newly warns.

**Span pins verified independently.** I sliced the probe sources with the spans `gsc` actually printed, rather than trusting column arithmetic:

| probe | id | span | covers |
|---|---|---|---|
| b, g | GS0303 | `(3,12,3,16)` | `func` |
| b, g | GS0398 | `(3,12,3,29)` | `func(int32) int32` |
| h | GS0303 / GS0398 | `(2,21,2,25)` / `(2,21,2,38)` | `func` / `func(int32) int32` |
| `**func` | GS0303 / GS0398 | `(3,13,3,17)` / `(3,13,3,30)` | `func` / `func(int32) int32` |
| d (arrow) | GS0398 | `(3,13,3,29)` | `(int32) -> int32` |

Exactly the strings `Issue3268PointerLegacyFunctionTypeTests` asserts. The pins are real: under the reverting mutant the diagnostics vanish entirely, so `Assert.Single` goes red.

**Suite — all 9 projects, `--no-build` on the CI-flagged Release build:** Compiler 4266, Core 7183/1 skipped, Extensions 104, InternalAnalyzers 9, Interpreter 1346/3 skipped, LanguageServer 468, Sdk 59, Cs2Gs 1909, GeneratorHost 31 — **15,375 passed, 4 skipped, 0 failed**, matching the PR body exactly. Release build 0 warnings / 0 errors. Legal controls compile **and execute**: `*int32`/`**int32` → rc 0 `legal-ok`; inferred `&add` → rc 0 `7`.

---

## NIT 1 — half-removed: reachable-by-nothing machinery left behind

The PR deletes the **only** call site of `ParseManagedFunctionPointerTypeClause` (`main` had exactly two mentions: the call at `:109` and the definition at `:738`). Provable by construction, not by mutant survival. Now unreachable: the ~32-line production at `Parser.TypeClauses.cs:728`, the binder branch `Binder.cs:3005-3014`, `ReportUnmanagedPointerOutsideUnsafe` (`DiagnosticBag.Reports.Emit.cs:169`), the GS0404 descriptor (`DiagnosticDescriptors.cs:318`), and `TypeClauseSyntax.IsManagedFunctionPointer` / `ManagedFunctionPointerFuncKeyword`. `GSharpPrinter.cs:191-197` still renders `*func(…)` and `Cs2Gs.Tests/Coverage/GNodeSamples.cs:250` still constructs `isManaged: true`, so a cs2gs coverage sample now renders G# the compiler rejects.

**I am not recommending deletion.** If the BLOCKER resolves toward option 1 or a revert, this is exactly the machinery you need. It is listed so the state is deliberate either way.

## NIT 2 — the pin asserts covered *text*, not position

`AssertDiagnostic(diagnostics, "GS0303", "func")` compares `Location.Text.ToString(Location.Span)`. Theory case 3 is `unsafe func Take(f *func(int32) int32)` — **two** `func` tokens on that line, at cols 8 and 21. The span is correct (col 21, measured above), but the assertion cannot tell the two apart, so a span sliding to the declaration keyword would still pass. Adding start line/column would close it.

## NIT 3 — pre-existing, not this PR's, but this PR routes more inputs into it

A pointer type used as a **generic/container type argument** ICEs and leaves a **0-byte artifact**:

```
map[string,*int32]  → rc 1, GS9998 ArgumentException: The type 'System.Int32*' may not be used as a type argument
```

Controls prove it is orthogonal: `*int32`, `*string` and the **arrow** form all do this on **both** `main` and this branch. Before the PR `map[string,*func(…) …]` gave a clean GS0155; now it converges onto this ICE — arguably correct parity (the issue asks for `*func` to be treated exactly as `*((…) -> …)`), but it means #3268's "no path should reach the emitter with this type" is not yet true in container positions. Worth its own issue; **not** worth widening this PR.

---

**Verdict: BLOCKER.** The diagnostics work is sound, well-pinned and honestly attributed. The question is whether removing `*func(T1, T2) R` — and with it function-pointer parameters, returns and fields, plus a cs2gs coverage category — is an authorized way to resolve the ambiguity that #3268 exposed. That is an owner decision, and it should be made explicitly rather than inside a PR titled as a validation fix.

DavidObando:
Blocker resolved in `2649c0108` by re-aiming the PR rather than patching the removed design.

- Restored the ADR-0122 `*func(T1, T2) R` parser production, all Issue #1035 tests, cs2gs managed function-pointer mapping, ADR/spec text, and live GS0404 diagnostics documentation.
- `FunctionPointer_PassAsParameter_CompilesAndRuns` passes again and prints `30` / `42`.
- Current `main` already enforces GS0404 for initialized and uninitialized top-level managed function-pointer declarations; exact ID/span/no-GS0303 regression pins were added without changing that gate.
- The remaining defect was emit-only: `Conversion.Classify` already accepted `nil -> FunctionPointerTypeSymbol`, but `EmitConversion` only materialized null `PointerTypeSymbol` values. The shared zero-native-int path now handles both pointer kinds.
- Managed and unmanaged function-pointer nil parameters compile, load, run, and print `0`.
- Nine solution-enumerated test projects: 15,382 passed, 4 skipped, 0 failed. CI is green (30 passed, 2 expected skips).

DavidObando:
## Rubber-duck round 2 — `2649c010` vs `main` `7341a6e68`

**Verdict: MERGEABLE.** 0 BLOCKER · 1 SHOULD-FIX · 3 NIT. Round one's BLOCKER is fully retired — verified by construction, not by reading the diffstat.

`main` re-fetched immediately before posting: **`7341a6e688b6d4f1602dfaead0aca6fb1004e093`** (it moved under me mid-review, `acc10fe47` → `7341a6e68`, #3274). Head **`2649c010878562f8c691466591ec6a0a86b7b4f5`**, merge-base `f5642aff1`, PR diff **3 files / +77 −4**. `merge-tree --write-tree` rc 0, single tree OID, **zero stage entries, zero `CONFLICT` lines**. Because a clean conflict list is not proof a rebase is viable, I **rebuilt on the merge with the new main** and re-ran everything; all identical.

Everything below measured in my own worktree, Release, `-p:ContinuousIntegrationBuild=true`. Witness `GSharp.Core.dll` md5 chain: `667ba5b2` (head+`acc10fe47`) → `463ea988` (M1) → `f31cbddf` (M2) → `6b6b65c5` (M3) → back to **`667ba5b2` exactly** → `d24a3b7d` (head+`7341a6e68`). `git status --porcelain` and `git diff HEAD` empty at every checkpoint.

---

## Q1 — right layer? Yes. Root cause, not symptom — for the type this issue names.

`Conversion.Classify` has said `nil → FunctionPointerTypeSymbol` is **`Conversion.Implicit`** since the feature shipped (`Conversion.cs:313`, alongside the `PointerTypeSymbol` arm at `:284`). The binder was never wrong. The emitter simply never implemented what the binder promised. `git log -S` puts **both** the classification and the GS0404 gate in **`5bdf22934` "Fixes #1035"** — so this is a ~2-year-old classify/emit divergence, and the emitter is the correct and only layer to fix it.

**Sibling-site grep** (`git grep PointerTypeSymbol -- src/**/*.cs` filtered for null, plus every `Ldnull`/`Conv_i` emission site — 15 and 5 respectively):

| site | `*T` | `*func(…)` |
|---|---|---|
| `Conversion.cs:284/313` classify | Implicit | Implicit |
| `MethodBodyEmitter.Conversions.cs:44` **EmitConversion** (this PR) | `ldc.i4.0; conv.i` | `ldc.i4.0; conv.i` ✅ |
| `MethodBodyEmitter.Conversions.cs:1180` **EmitDefault** | **`ldnull`** ❌ | `initobj` |
| optional-parameter `= nil` validator | accepted | rejected (GS0265) |

Only `EmitConversion` and `EmitDefault` materialise a null. The PR fixes one; the other has the **inverse** one-type omission and it is a live defect — see “Adjacent defects” below. It is pre-existing (`git diff` confines the hunk to lines 38–48; the fnptr `EmitDefault` IL is **byte-identical** pre/post), it does not involve function pointers, and widening this PR to chase it would be the wrong call. Recorded so the class is documented rather than assumed closed.

## Q2 — over/under-reach: no over-reach found

Every position, measured on the merged tree by compiling **and executing**:

| probe | source | result | IL |
|---|---|---|---|
| param (managed) | `printPointer(nil)` | rc 0, `0` | `16 d3` = `ldc.i4.0; conv.i` ✅ |
| param (unmanaged) | `unmanaged[Cdecl] (int32) -> int32` | rc 0, `0` | `16 d3` ✅ |
| return | `unsafe func get() *func(int32) int32 { return nil }` | rc 0, `0` | `16 d3 2a` ✅ |
| assignment | `fp = &add` → `5`; `fp = nil` → `0` | rc 0 | `16 d3` ✅ |
| struct field | `Holder{fp: nil}` | rc 0, `0` | `16 d3 7d` (stfld) ✅ |
| array element | `[1]*func(int32) int32{nil}` | rc 0, `0` | `16 d3 a4` (stelem) ✅ |
| ternary | `flag ? &add : nil` | rc 0, `True`/`False` | ✅ |
| **delegate (nullable arrow)** | `((int32) -> int32)? = nil` | rc 0, `True` | **`14` = `ldnull`** ✅ **not** taken |
| delegate (non-nullable) | `(int32) -> int32 = nil` | rc 1 **GS0155** | unchanged |
| `nint` | `var n nint = nil` | rc 1 **GS0155** | unchanged |
| `*int32` param | `take(nil)` | rc 0, `0` | `16 d3` (pre-existing path) ✅ |
| generic slot | `List[*func…]` | pointer-in-generic-brackets does not parse — pre-existing, orthogonal |

`FunctionPointerTypeSymbol : TypeSymbol` is `sealed` and disjoint from every delegate/function-type symbol, so the widened pattern **cannot** capture a reference type. Confirmed by IL, not by reading: the delegate case still emits `ldnull`.

**Under-reach:** `fp == nil` and `fp != nil` are **GS0129** *“Binary operator '==' is not defined for types '\*func(int32) int32' and 'nil'”* — while `*int32 == nil` works and prints `True`. See Adjacent defect ②.

## Q3 — row `h` prints `0`, and that is correct but is **not** what makes the test bite

IL is `ldarg.0; conv.i; conv.i8; call WriteLine(int64)` — `0` is the pointer's numeric address, the honest rendering. But **`ldnull` would also print `0`**, so the printed value alone does not discriminate. I checked whether the pin can tell a *right* fix from a *different wrong* fix:

**Mutant M3** — keep the widened predicate, emit `ldnull` instead of `ldc.i4.0; conv.i` (md5 → `6b6b65c5`):
- `row_h` **compiles rc 0** and then **`System.InvalidProgramException`, rc 134**
- both new theory cells fail: `exited 134`

So the pin's teeth are `CompileAndRun`'s `dotnet exec` **exit-code** assertion, not `Assert.Equal("0\n", …)`. It does discriminate. (Interesting: under M3 the *return* position still ran fine — only the argument position is JIT-fatal. The test happens to pick the position that fails loudly.)

**Calling through a nil function pointer**: `fp(1)` with `fp = nil` → **SIGSEGV, rc 139**, no managed exception. That is correct `unsafe`/`calli` semantics and matches C# `delegate*`. The in-language guard is `nint(fp) == 0` (measured: prints `guarded`), because `== nil` is unavailable.

**GS0404 spans**, sliced out of the probe sources with the spans `gsc` actually printed rather than by column arithmetic:

| probe | id | span | covers |
|---|---|---|---|
| `var fp *func(int32) int32 = nil` | GS0404 | `(4,8,4,9)` | `*` |
| `var fp *func(int32) int32` | GS0404 | `(4,8,4,9)` | `*` |
| `var fp *((int32) -> int32) = nil` | GS0398 | `(4,14,4,30)` | `(int32) -> int32` |

Exactly the strings the tests assert. `*` is deliberate — `ManagedFunctionPointerStarToken.Location`, shipped with #1035.

**Are the b/g pins RED-provable?** Yes — I scored them, I did not assume. **Mutant M2** (md5 → `f31cbddf`), three independent edits: GS0404 location → `syntax.Location`; GS0398 location → `syntax.Location`; `SymbolDisplay` `*func(` → `*FNPTR(`. **All 4 new binder cells went red with the right message**:

```
ManagedFunctionPointer_TopLevel_ReportsGS0404("")        Expected "*"  Actual "*func(int32) int32"
ManagedFunctionPointer_TopLevel_ReportsGS0404(" = nil")  Expected "*"  Actual "*func(int32) int32"
DelegatePointer_InsideUnsafe_ReportsGS0398   Expected "(int32) -> int32"  Actual "*((int32) -> int32"
ManagedFunctionPointer_ConversionDiagnostic_UsesSourceSpelling  Not found: "*func(int32) int32"
```

**Mutant M1** (product hunk → `main`, md5 → `463ea988`): both emit cells red with `GS9998 NotSupportedException: Conversion from 'nil' to '*func(int32) int32' is not yet supported by the emitter`; binder tests **10/10 still green** — correctly confirming they pin binder behaviour, not this hunk. And every nil-to-fnptr position (param, return, assignment, field, array, unmanaged) is GS9998 on `main`: the old failure was a hard ICE, never silent bad IL.

## Q4 — restoration complete

| item | state |
|---|---|
| `Parser.TypeClauses.cs` | **byte-identical to `main`**; `ParseManagedFunctionPointerTypeClause` call at `:109`, def at `:738` |
| `FunctionPointer_PassAsParameter_CompilesAndRuns` | back; passes (`30`/`42`) |
| `ManagedFunctionPointer_Parameter_NoError`, `…_OutsideUnsafe_ReportsGS0404` | back |
| `tools/cs2gs/…/CSharpTypeMapper.cs:507` | `delegate* managed<…>` → `FunctionPointerTypeReference(isManaged: true, …)`, **not** Unsupported |
| `docs/adr/0122-…md` | byte-identical to `main`; §9 describes the feature as implemented, GS0404 live |
| `website/docs/ref/spec.md:403` | teaches `*func(T1, T2) R` + GS0404 as current |
| `docs/diagnostics.md:990` ↔ `website/docs/ref/diagnostics.md:282` | GS0404 row **verbatim identical**, live Error, no “Retired” |

The PR changes **zero** doc files — the restoration came from re-aiming the branch, which is the cleanest possible form. Also clean: no `#pragma warning disable CS0618` added, tests drive `Compilation.Emit`, so the ADR-0156 retirement gate is untouched.

## Q5 — GS0303 count: **zero**

No `GS0303` in any of the 25 probe outputs. `Assert.DoesNotContain(…GS0303)` on both new binder theory cells pins it. ADR-0075 deprecates `func(…)` as a *type clause*; `*func(…)` is ADR-0122 §9 and never routes through it.

## Suites

Full 9-project sweep on the merged tree: **15,386 passed / 4 skipped / 0 failed** (Compiler 4269 · Core 7182/1 · Extensions 104 · InternalAnalyzers 9 · Interpreter 1357/3 · LanguageServer 468 · Sdk 59 · Cs2Gs 1907 · GeneratorHost 31). Re-run against the **new** main merge: Core 7183/1, Interpreter 1360/3, Sdk 64, plus both Issue1035 classes 7/7 and 10/10 — all green. Release 0W/0E.

**ILVerify gate: 123/123**, rc 0, and `grep -c "Assembly or module not found"` = **0** — the no-result-wearing-a-pass was checked for, not assumed absent.

---

# SHOULD-FIX

**1. `nil` for function pointers is now real but undocumented — `docs/adr/0122-…md` §9 and `website/docs/ref/spec.md:403`.**
Both surfaces enumerate the operations. For `*T` they explicitly list *“comparison/equality, **a null pointer (`nil`)**”*. For the function-pointer form they list `&Method`, `fp(args)`, `nint` round-trip and fnptr↔fnptr casts — and **no `nil`**. A reader concludes `nil` is unsupported, which is precisely what this PR makes false. One clause in each of the two hand-maintained files (they have no generator; #3201/#3207 both landed stale mirrors this way). Cheapest possible fix, and it is the half of this change a user actually sees.

# NIT

**2. The emit file's ILVerify ignore list hides this exact defect class.** `Issue1035FunctionPointerEmitTests.cs:28-35` ignores `"StackUnexpected"` — the code ILVerify raises for `ldnull` where a pointer is expected. Under M3 the verifier was silent and only `dotnet exec` caught it. The list is defensible (`calli` and unmanaged pointers are unverifiable by design) and the runtime gate covers it, but worth knowing the verifier is not the guard here.

**3. Attribution in the PR body.** *“Current `main` (`f5642aff`) already reports `GS0404`”* is true, but `git log -S` shows the gate **and** the `nil → FunctionPointerTypeSymbol` classification both shipped in **`5bdf22934` “Fixes #1035”**. They were never broken and never fixed by an intervening merge — the only thing that ever removed them was this PR's own round one. Rows b/g are #1035's original behaviour, newly pinned. Saying so keeps #3268's history readable.

**4. GS0404 anchors on the bare `*` (1 char).** Deliberate and consistent; the pin asserts it exactly. Flagged only so nobody later “fixes” it and silently breaks two cells.

---

# Adjacent product defects found while probing — **not this PR's to fix**

Both pre-existing on `main`, both unchanged by this diff, both about `*T` rather than `*func`. Raising them because Q1 asked for the sibling-site sweep and because one of them is a silent-compile → hard-crash.

**① `EmitDefault` emits `ldnull` for `PointerTypeSymbol` — an optional pointer parameter miscompiles.**

```gs
unsafe func take(p *int32 = nil) { Console.WriteLine(nint(p)) }
take()
```
→ **compiles rc 0, writes a DLL**, then `System.InvalidProgramException`, **rc 134**.
`<Main>$` IL is `14 28 …` = `ldnull; call take(int32*)`. ILVerify: `StackUnexpected [found Nullobjref 'NullReference'][expected Native Int]`.
An uninitialized `var p *int32` local gets the same `ldnull; stloc.0` and survives only by luck (`ldnull` happens to be a pointer-sized zero at runtime). The function-pointer twin is fine (`initobj`). This is literally the defect this PR fixes, in the other null-materialisation site — `MethodBodyEmitter.Conversions.cs:1180`, same file, ~1130 lines below. The ponytail root fix is one “materialise null of `T`” helper both sites route through. Note the two pointer kinds also disagree at a third place: `*func(…) = nil` as an optional default is **rejected** with GS0265 while `*int32 = nil` is accepted and miscompiles.

**② No `==`/`!=` against `nil` for function pointers.**
`fp == nil` → **GS0129**, `fp != nil` → GS0129, while `*int32 == nil` works. Calling through a nil function pointer is SIGSEGV rc 139. So after this PR you can create, store, pass, return and field-hold a nil function pointer but cannot naturally test for one before calling it; the workaround is `nint(fp) == 0`. Pre-existing, but this PR is what makes nil function pointers routine, so the gap is newly reachable.

---

**Bottom line.** Structurally the right shape: one existing shared path widened by one type, matching a classification the binder already made, with pins that are RED-provable against three independent mutants and that survive a *different-wrong-emission* mutant, not just a reverted one. Restoration of the round-one deletion is complete and verified by execution (`30`/`42`), by parser byte-identity, and across all four doc surfaces. No blocking issues.

<sub>Probes in a private temp root; worktree, ref and probes removed after posting. `git status --porcelain` empty in every checkout.</sub>

DavidObando:
SHOULD-FIX addressed in `d7dc04cb1`.

- ADR-0122 section 9 and `website/docs/ref/spec.md` now state: "`nil` denotes the null function-pointer value."
- No `docs/` mirror exists for `website/docs/ref/spec.md`; versioned 0.3 snapshot remains intentionally unchanged.
- PR body now records that both GS0404 gate and nil conversion classification shipped with #1035 (`5bdf22934`); this PR fixes only emitter native-zero path.
- Rebased onto `65578d41b`; Release build has 0 warnings/errors, and all nine solution-enumerated projects pass (15,399 passed, 4 skipped).

Issue #3273, opened 2026-08-05
Fifth wave: 45 test files still claim an interpreter-vs-emitter oracle that Phase 3c deleted


DavidObando:
## Summary

The evaluator-retirement doc sweep has been run four times (#3257, #3263, #3265) and each wave believed it was the last. A review of PR #3269 measured the remaining corpus and found **45 files** still carrying present-tense interpreter-vs-emitter claims that the code can no longer implement.

Phase 3c (#3244, ADR-0156) deleted the tree-walking evaluator. `GSharpRepl` is now a shim over `EmittedSessionEngine`. Any doc-comment claiming a test compares interpreter output against emitter output, or pins "interpreter semantics", is advertising an oracle that **no longer exists**. The tests still pass — they just do not test what they say they test.

This is the same defect class as #3265, not a new one. Filing it as a single sweep so the count is closed out by measurement rather than by another round of "this must be the last wave".

## Scope: how the 45 was counted

Counted as in-scope only files that **both**:
1. make a present-tense interpreter-vs-emitter / interpreter-semantics claim, **and**
2. carry **no** retirement note.

Files already honestly re-documented (the ones that say the evaluator is gone and the pin is now an emitted-oracle pin) were **excluded**. So this is not a raw grep count — files fixed by the earlier waves do not inflate it.

## Confirmed sample

Verified directly against `main` @ `0232ecb5e`:

`test/Interpreter.Tests/Issue711IfExpressionInterpreterTests.cs:10-13` claims:

> Issue #711 / ADR-0064 — interpreter parity for `if` used as a value-producing expression. The interpreter shares the binder with the emit pipeline, so the binding rules are common; this file pins down the …

…while `:224` drives `new GSharpRepl()` — the emitted engine. Exactly #3265's defect.

Further instances reported with `file:line`:

- `test/Interpreter.Tests/Issue833…:12`
- `test/Interpreter.Tests/Issue799Variadic…:18`
- `test/Interpreter.Tests/Issue729…:13`
- `test/Interpreter.Tests/Issue710…:12`
- `test/Interpreter.Tests/Issue2925…:12`
- `test/Interpreter.Tests/Issue608…:12`
- `test/Interpreter.Tests/StaticVirtualInterface…:18`
- `test/Interpreter.Tests/Attribute…:12`

And, importantly, **outside `Interpreter.Tests`** — the earlier waves only swept that one assembly:

[...]

PR #3276, opened 2026-08-05
Correct remaining emitted test intent

DavidObando:
## Summary
- correct stale tree-walking evaluator claims across 128 test files
- rename 73 test classes/files to their actual emitted-session, emitted-oracle, or driver path
- preserve all test bodies and leave five feature-parity suites untouched

Independent source sweep found 128 in-scope files, not 45. Primary class/name scan found 128 candidates; six historical/non-engine matches were excluded, then six method/fixture XML-doc files were added.

References #3273.

## Validation
- all 11,182 `[Fact]`/`[Theory]` bodies byte-identical: SHA-256 `7C07C8C2E08AD702A178B7813E8978D9657302C6363355505C0C1CF5813B0B36`
- Release warnings-as-errors build: 0 warnings, 0 errors
- nine projects: 15,369 passed, 4 skipped, 0 failed; 9/9 result lines, no zero-match filters
- ILVerify: 123/123 verified; 2 documented suppressions, 0 failures


DavidObando:
## Evidence

**Base:** `0232ecb5eb7c1f6878556599a6faf557487e4232`  
**Head:** `bbb418093ebb456986376b1fefed99f099ceb689`

### Independent scope derivation: 128 files, not 45

I enumerated every tracked `test/**/*.cs`, selected files with a present-tense deleted-engine claim in the test-class name or XML docs, and excluded files carrying an explicit evaluator-retirement note plus the five feature-parity files.

- primary class/name scan: 128 candidates
- excluded as historical/non-engine matches: 6
  - `Issue1325StructConstraintEmitTests`: “reinterpretation” is a CLR operation
  - `Issue687ColorColorTypeShadowingTests`: “instance interpretation” is binder disambiguation
  - `ClrTypeUtilitiesTests`: `MatchEvaluator` is a BCL type
  - `PrimitiveConversionTests`: mentions only a possible downstream failure
  - `Issue2349LambdaIfStatementEmitTests`: historical negative contrast; explicitly tests `gsc`
  - `Issue3218NullableProjectionTargetEmitTests`: past-tense defect history
- secondary XML-doc scan added 6 method/fixture-doc files missed by the primary-class scan
- final in-scope population: **128** (`Interpreter.Tests` 83, `Core.Tests` 39, `Compiler.Tests` 6)

The earlier 45 count missed class-name-only claims, wording other than exact “interpreter parity,” Core/Compiler files, and method/fixture XML docs. It can also falsely classify `Issue798` as retired if a file-level `no longer` grep sees “CFG no longer crashes.”

Result: 73 class/file renames, 55 documentation-only corrections, zero in-scope residue.

### Coverage preservation

Roslyn-parsed hash of every `[Fact]`/`[Theory]` method body across the complete test corpus:

- before: 11,182 bodies, SHA-256 `7C07C8C2E08AD702A178B7813E8978D9657302C6363355505C0C1CF5813B0B36`
- after: 11,182 bodies, SHA-256 `7C07C8C2E08AD702A178B7813E8978D9657302C6363355505C0C1CF5813B0B36`

No test, assertion, helper, or specimen removed. Required class self-references changed only where their containing class was renamed. No opportunistic assertion strengthening; RED/GREEN split therefore not applicable.

### Explicitly untouched feature-parity files

- `test/Compiler.Tests/Emit/Issue611SliceArrayParityTests.cs`
- `test/Core.Tests/CodeAnalysis/Binding/Issue1172ArrowLambdaParityBinderTests.cs`
- `test/Core.Tests/CodeAnalysis/Binding/Issue2237AssemblyAttributeParityTests.cs`
- `test/Core.Tests/CodeAnalysis/Emit/Issue1172ArrowLambdaParityEmitTests.cs`
- `test/LanguageServer.Tests/LspServerParityContractTests.cs`

Diff hits across these five: 0.

### Validation

- Release build with `-p:ContinuousIntegrationBuild=true -warnaserror`: **0 warnings, 0 errors**
- full nine-project suite on final commit: **15,369 passed, 4 skipped, 0 failed**
- result denominator: **9/9**, `No test matches`: **0**
- ILVerify gate: **123/123 verified**, 2 documented suppressions, 0 verification failures, 0 compile failures
- worktree: clean

DavidObando:
## Rubber-duck review — PR #3276 (`fix/3273-interpreter-oracle-doc-sweep`), issue #3273

**Verdict: MERGEABLE.** 0 BLOCKER · 3 SHOULD-FIX · 3 NIT.

`main` `6b809619` (re-fetched at start and again before posting — unmoved), head `bbb41809`. The head is **1 behind / 1 ahead**, so I did not review it in isolation: `git merge-tree --write-tree --messages` produced tree `921c4827` with **zero stage lines** (read as output, not exit code), and I built and tested that **real merged tree**. `BUILD_RC=0`, **0 warnings, 0 errors**.

### Scope, measured

128 files, +326/−830, **entirely under `test/`** — `git diff --name-only main..merged -- src/ docs/ website/` is **empty**. 73 renames + 55 modifications, **0 files added, 0 deleted**.

The −830 looked alarming for a doc sweep, so I classified every changed line:

| line kind | deleted | added |
|---|---|---|
| code | **99** | **99** |
| inline `//` comment | **73** | **73** |
| doc `///` comment | **658** | **154** |

All 99 code lines are class/method renames (including one consequential `typeof(...)` update at `Issue750ConstraintOverloadEmittedOracleTests.cs`). **No assertion, `[Fact]`, `[Theory]` or `Skip` line is touched.** The entire net −504 is `///` prose.

### Coverage preservation — the check #3273 asks for

The issue warns that equal assertion counts do not prove equal strength (`Equal`→`NotNull` preserves the count). So I diffed the full **assertion-kind histogram** across all 128 file pairs, rename-aware:

- `Assert.*` total: **1683 → 1683**
- `[Fact]` **940 → 940**, `[Theory]` **34 → 34**
- Histogram across all 20 distinct `Assert` kinds: **identical** (`Equal` 668, `Contains` 277, `Empty` 340, `False` 102, …)

No assertion was weakened, substituted or removed.

**Skipped count did not rise** — measured, not inferred. `Skip =` occurrences across the whole test tree: **5 on base, 5 on merged**. And running the affected project on both trees:

| | passed | skipped | total |
|---|---|---|---|
| `Interpreter.Tests` **base** | 1346 | 3 | 1349 |
| `Interpreter.Tests` **merged** | 1346 | 3 | 1349 |

Identical. `Core.Tests` on merged: **7178 passed / 1 skipped / 0 failed**, `CORE_RC=0`.

### Running the product

Per the "oracle is the running program" bar, I checked the behavioural claims rather than reading them:

- `gsi --help` → `--engine <name>  Interactive engine: 'emit' (the default and only engine; also via GSI_ENGINE). The legacy 'evaluator' engine was removed in ADR-0156 Phase 3c.` Claim confirmed.
- `samples/Arithmetic.gs` through both drivers: `gsi <file>` rc=0, `gsc /out:` + `dotnet exec` rc=0, and **`gsi` output == emitted output == `Arithmetic.golden`**, byte-identical.

### Retired diagnostics

`GS0510`, `GS0513`, `GS0514`, `GS9999` are **genuinely absent from `src/`** — 0 occurrences each. I checked this was absence and not formatting: there is no `GS{code:D4}`-style construction of these ids in the diagnostic definitions and no enum member `= 510/513/514/9999`. The only id-formatting site is `src/Compiler/Program.cs:1151`, which parses user-supplied `/nowarn` codes.

**This PR touches no diagnostics reference table**, so the retire-the-row-vs-delete-the-row question does not arise here. The only changes are prose in test comments, and one of them is exactly the right treatment: `the GS0510 interpreter` → **`the historical GS0510`** (`EmittedSessionEngineTests.cs:332`) — past-tense framing preserved rather than the claim being deleted or "corrected". That is the pattern the other files should have followed.

### Mirror pairs and the #3274 collision question

`docs/` and `website/` are **untouched in their entirety**. Therefore:
- **`docs/diagnostics.md:1202` is NOT touched by this PR — no collision with #3274.**
- No live mirror pair (`docs/diagnostics.md` ↔ `website/docs/ref/diagnostics.md`, `docs/lsp.md` ↔ `website/docs/tooling/lsp.md`, coverage matrix ↔ inventory JSON) is one-sidedly updated.

### `*Parity*` exclusion

Correctly honoured — **zero** `*Parity*` files appear in the diff.

### Independent completeness count

Files whose basename contains `Interpreter`: **88 on base → 16 on merged** (72 renamed, plus one `…EvaluatorTests` = the 73 renames). Prose matching live-interpreter claim patterns: **109 → 50**. The sweep is a real, large reduction, and most of the 50 survivors are correctly past-tense and framed ("retired with the tree-walking evaluator in ADR-0156 Phase 3c (#3176)"), which is right and must stay.

---

## SHOULD-FIX 1 — two ADR cross-references now dangle

The renames break two file-path pointers that this PR does not update:

- `docs/adr/0098-friendly-numeric-type-aliases.md:124` → `test/Interpreter.Tests/Issue729FriendlyNumericAliasInterpreterTests.cs` (now `…Issue729FriendlyNumericAliasEmittedSessionTests.cs`)
- `docs/adr/0084-gsharp-extensions-optional-sequences.md:494` → `test/Interpreter.Tests/Issue813TupleSequenceReturnInterpreterTests.cs` (now `…Issue813TupleSequenceReturnEmittedSessionTests.cs`)

I checked both in context before filing: **both sit inside present-tense "Coverage:" enumerations of which test files cover the decision**, not inside historical narrative. So updating the path preserves truth rather than falsifying the record — this is not the "don't rewrite history" case. Both paths currently 404.

#3273 explicitly warned about this ("where a `<see cref>` points at a renamed symbol, the mirror must be updated too… PR #3269 had to touch 2 extra files for exactly this reason") — and it is **2 files again**. I verified these are the *only* two: I extracted all 73 old class names and grepped the entire merged tree (`*.cs`, `*.md`, `*.json`, `*.csproj`); exactly these two hits, nothing else dangles.

Incidental, pre-existing and **not** introduced here: ADR-0084:494 also describes that test as covering "tree-walking parity", which is itself now stale. Worth fixing in the same touch.

## SHOULD-FIX 2 — the sweep never touched inline comments, so the corpus is not closed

The line classification above shows inline `//` comments are **exactly balanced 73/73** — they were only rename-updated, never rewritten. All rewriting happened in `///` blocks. Present-tense claims about the deleted evaluator therefore survive in method bodies:

- **`test/Core.Tests/CodeAnalysis/Binding/Issue795DefaultExpressionBinderTests.cs:89`** — "the tree-walking interpreter **does not propagate** generic type-argument substitution into the body of a user function". **This PR already opened this file** and rewrote its class summary, leaving this behind. Cheapest possible fix.
- `test/Core.Tests/CodeAnalysis/Binding/Issue1235TypeParameterMemberWriteTests.cs:78` — "not evaluated end-to-end by the tree-walking interpreter — the interpreter's property-assignment evaluator only executes a concrete setter body".
- `test/Core.Tests/CodeAnalysis/Binding/Issue1136ObjectMemberBinderTests.cs:125` — "the interpreter **executes** the user method".
- `test/Core.Tests/CodeAnalysis/Binding/AwaitForRangeStatementTests.cs:48` — "the interpreter **can drain** it end-to-end".
- `test/Core.Tests/CodeAnalysis/Binding/RefStructTests.cs:445` — "the interpreter **cannot execute** but whose binder diagnostics still apply".

My heuristic survivor count is **29 inline + 17 doc**, filtering out retirement-note wording. Treat that as a **floor, not a measurement** — the filter is keyword-based and I did not adjudicate all 46 by hand.

This is not a blocker: the PR strictly and substantially reduces the corpus. But it does mean the issue's framing — a fifth wave that "closes the count out by measurement" — is **not yet true**, and a sixth wave will be needed. Better to say so now than to close #3273 believing it is done, which is the exact pattern #3273 was filed to break.

## SHOULD-FIX 3 — accurate, still-true rationale deleted along with the false claims

658 `///` lines removed against 154 added, and **80 of 128 files** are reduced to a templated one-liner of the form `Issue #NNNN: Emitted-oracle coverage for <topic>`. Some of what went with it was neither false nor interpreter-specific:

- **`Issue1718FrameConcurrencyEmittedOracleTests.cs`** (+4/−33) lost the paragraph explaining that these tests *"deliberately avoid wall-clock sleeps: correctness is synchronized through `scope` … and, for the crash-detection test, by running many full evaluations concurrently via `Parallel.For`"*. I verified `scope`/`Parallel.For` are **still used in the file** — so that guidance is still accurate, is engine-independent, and is now gone. It is precisely the note that stops a future maintainer "fixing" a flaky-looking concurrency test by adding a sleep.
- **`Issue2388NullableCustomEqualityBinderTests.cs`** (+1/−18) lost the explanation that `ClrOperatorResolution`'s CLR-type match only inspects `TypeSymbol.ClrType`, which for `NullableTypeSymbol` is already the underlying type, and that the resulting bad IL surfaced at the **CALL** site rather than the **BIND** site — still-true binder rationale that explains why the assertions are shaped as they are.

#3273 asked that each name and comment "describe what the test actually does now" and said *"do not remove tests, assertions, helpers or specimens"*. It did not ask for design rationale to be removed. Recommend restoring the engine-independent paragraphs (test-design notes and binder rationale) while keeping the interpreter narrative deleted — or converting the interpreter parts to the past-tense form this PR already demonstrates correctly at `EmittedSessionEngineTests.cs:332`.

---

## NITs

1. **The two new suffixes are not a principled distinction.** All 44 `*EmittedSessionTests` use `GSharpRepl`, but 2 of the 26 `*EmittedOracleTests` do as well — `Issue2918InlineLambdaErasedReceiverEmittedOracleTests.cs` and `Issue2962ConstrainedStaticPropertyChainEmittedOracleTests.cs` — while the other 24 go through the shared `EmittedOracle.Evaluate` helper (`test/Shared/EmittedOracle.cs`). Neither name is *false* (both are emitted paths), so this is naming consistency, not a misleading-name defect.
2. **16 files keep "Interpreter" in the name** and the PR does not state why. Several are plausibly legitimate boundary/refusal tests (`Issue2956InterpreterBoundaryTests.cs`, `Issue2987ByRefLikeBoundaryInterpreterTests.cs`, `Issue3004PointerBoundaryInterpreterTests.cs`, `Issue2988DeinitInterpreterTests.cs`), but others are less obvious (`Issue758LibraryImportInterpreterTests.cs`, `RefLocalAliasingInterpreterTests.cs`, `Adr0151IfLetExpressionInterpreterTests.cs`). Please **state the enumeration** — as written, a reader cannot tell deliberate exclusion from oversight.
3. `samples/AddressBook.gs:9` still says "the interpreter exercises the same constructs in Phase 3.C unit tests" (present tense). Outside #3273's stated scope, recorded so it is not lost.

## Disclosed self-errors

1. **My first assertion count was wrong and I nearly reported it.** I built the file list with `git diff --name-only`, which for renames emits only the *destination* path; my `[ -f ]` guard then silently skipped 73 base-side files, giving base 13,673 lines vs merged 23,979 — impossible against +326/−830. The arithmetic contradiction caught it; redone with `--name-status -M`. **A missing file is not a deleted file, and a skipped file is not a zero.**
2. **My strongest hypothesis was disconfirmed by measurement.** I suspected `Issue2388`'s new "Emitted-oracle coverage" summary was a *fresh false claim*, because the deleted prose said "method bodies are bound but never invoked — this is pure binder coverage". Reading the body found `EmittedOracle.Evaluate(source)` at `:176`, and that call is present **on base** too. So the deleted sentence was already false on `main` and removing it is correct. Reading the file beat reasoning from the diff.
3. **My first product probe produced a convincing false finding.** I wrote `var xs = [1, 2, 3]` and got `GS0113` + a five-deep `GS0005` cascade at rc=1 from both drivers — which reads exactly like a product defect. It was my syntax, not the product. Re-ran using `samples/Arithmetic.gs` with its committed `.golden` as a known-good specimen.

## Denominator and what was not run

**9 test projects**, source `dotnet sln GSharp.sln list` (CI's own, per `build/generate-ci-test-matrix.py:26`).

Run locally: **`Interpreter.Tests` on both trees**, **`Core.Tests` on merged**.

**NOT RUN locally — stated plainly rather than implied:** `Compiler.Tests`, `Cs2Gs.Tests`, `Sdk.Tests`, `LanguageServer.Tests`, `Extensions.Tests`, `InternalAnalyzers.Tests`, `GSharp.GeneratorHost.Tests`. Justification: zero `src/` changes; the code delta is 100% renames; a clean **0-warning, 0-error solution build** is the compile-level oracle that every renamed reference resolves (this is what would catch a broken rename, and it passed); and CI is **32 checks — 30 SUCCESS, 2 SKIPPED, 0 anything else**, including the required **`ilverify`** gate. `Compiler.Tests` has 6 changed files, all comment/rename.

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

---

**MERGEABLE.** No test was removed, weakened, skipped or renamed away from what it asserts; the assertion histogram is identical, the skipped count is unchanged at 3, and both suites pass on the real merged tree. The three SHOULD-FIXes are the two dangling ADR pointers (which the issue itself predicted), the untouched inline-comment corpus (so #3273 should not be closed as complete), and the loss of still-true rationale in the largest prose deletions.

DavidObando:
## Follow-up: independent scope measurement — adjudicating 45 vs 128

Verdict unchanged: **MERGEABLE**. This addendum answers the contested scope count with a measurement derived from the corpus, and adds one quantified finding that strengthens my third SHOULD-FIX.

### Bottom line

**The 45 figure is refuted. The author's 128 is sound as far as it goes — over-reach is zero — but the corpus is larger than 128, so this PR is incomplete, not over-reaching.**

- **Over-reach: 0 of 128.** Every file this PR touches carried a stale engine claim on the base side.
- **In-scope corpus: ~155–170** by my classifiers, i.e. **128 is a subset**, not a superset.
- 45 is below even my strictest classifier (113) by 2.5×.

### How I got there, including two classifier errors of my own

I refused to reconcile the two numbers and instead built a two-condition classifier — *makes a present-tense engine claim* **and** *lacks a retirement note* — over all **1,672** `.cs` files in the base test corpus. I ran three versions, because the first two were wrong in opposite directions:

| classifier | in-scope | defect |
|---|---|---|
| v1 line-based, narrow retirement lexicon | 195 | **over-counts** — matches per line, so a sentence spanning two lines ("…the tree-walking evaluator and its escape hatch are" / "removed in Phase 3c") scores the first line as live; missed past-tense verbs (`refused`, `could not`, `raised`, `aligned`) |
| v2 block-aware, requires present-tense verb | 113 | **under-counts** — blind to noun-phrase claims (`"Interpreter parity"`, `"binder/interpreter tests"`, `"Interpreter behaviour is type-erased"`) |
| v3 block-aware, no verb requirement | 155 | closest, but bare `was\|were` in `RETIRE` wrongly suppresses a block when those words appear in an unrelated clause |

Rather than publish a regex output, I hand-adjudicated every residual:

**v1 partition of the 128 (base side): 73 name-claim + 52 live-prose claim + 0 retirement-framed-only + 3 no-claim.** The zero is the key cell — **no file was touched whose only engine reference was already correctly framed as historical.** The "PR re-churned files the earlier waves had already honestly re-documented" hypothesis is disconfirmed.

**v3 flagged 18 "touched but not in-scope". I read all 18.** Fifteen carry plainly stale claims my `RETIRE` regex had wrongly suppressed — `TupleEmitTests.cs:18` "remain interpreter-only", `StructTests.cs:19` "Interpreter-only for now", `Issue1260BaseBclCallBinderTests.cs:26` "interpreter (which cannot materialize a real CLR base instance…)", `Issue2928CallableMaterializationTests.cs:16` "interpreter closures must materialize…". The remaining 3 — `MapTypeTests.cs`, `SwitchExpressionTests.cs`, `Issue3010EntryPointDriverMatrixTests.cs` — I diffed individually, and **all three also carried the claim**, in the noun form **"interpretation"** ("Phase 3.A.4 binding and *interpretation* tests"; "switch expression binding and *interpretation*"; "compiler evaluation, emitted execution, and *interpretation*"), which my `ENGINE` pattern did not cover.

**Net: 128 of 128 justified. Over-reach zero.**

Conversely there are real misses — v3 counts 45 in-scope files untouched, and while that number carries false positives, I confirmed genuine ones by hand: `LambdaEmitTests.cs:97`, `Issue1136ObjectMemberBinderTests.cs:125`, `AwaitForRangeStatementTests.cs:48`, `ClrMemberAccessTests.cs:248`, `PrimitiveConversionTests.cs:16`, `Issue722GoExtensionsImportGateTests.cs:271`, `Issue723GoBuiltinsImportGateTests.cs:277`, and `DiagnosticIdUniquenessTests.cs:194` ("Emitted directly for unexpected evaluator failures" — a live claim about retired `GS9999`).

**Implication for #3273:** four prior waves each closed on an undercount, which is exactly why this recurred four times. This wave is the most complete by a wide margin and should land — but it should **not** close #3273 as done. The remaining tail is real and includes the inline-comment corpus from my SHOULD-FIX 2.

### `*Parity*` — author's claim verified independently

Exactly **5** files match `*Parity*` in the base corpus, and the enumeration matches the author's exactly:

```
test/Compiler.Tests/Emit/Issue611SliceArrayParityTests.cs
test/Core.Tests/CodeAnalysis/Binding/Issue1172ArrowLambdaParityBinderTests.cs
test/Core.Tests/CodeAnalysis/Binding/Issue2237AssemblyAttributeParityTests.cs
test/Core.Tests/CodeAnalysis/Emit/Issue1172ArrowLambdaParityEmitTests.cs
test/LanguageServer.Tests/LspServerParityContractTests.cs
```

**All 5 absent from the changed set.** Correct — all five are *feature* parity, not engine parity.

### Body-hash claim — right kind of evidence, and it is very nearly true

I did not take the hash on trust; I recomputed it **name-independently**, which is strictly stronger than the author's scheme and immune to the hollow-proof mode. I extracted method bodies by brace matching, stripped comments, normalized whitespace, and hashed the **sorted multiset** — so a renamed method contributes identically on both sides and cannot silently drop out.

Across the 128 changed files: **base 1,134 bodies, merged 1,134 bodies — 1,133 identical, exactly 1 different.**

```
BASE-ONLY: typeof(Issue750ConstraintOverloadInterpreterTests).Assembly.Location
MERG-ONLY: typeof(Issue750ConstraintOverloadEmittedOracleTests).Assembly.Location
```

That delta is **necessary and correct** — a self-referential `typeof` must follow its class rename.

Two notes for the author:
1. The claim "11,182 test bodies unchanged" is very slightly false: **one body must change**, and it is precisely the one that demonstrates the renames were applied consistently. If the tooling reported *zero* deltas, it dropped this method — and a method inside a **renamed class** is exactly what a name-keyed scheme drops from both sides. Please confirm the extractor is name-independent, and state the 1 expected delta rather than 0.
2. I could not reproduce **11,182**; my extractor finds **14,253** bodies corpus-wide. Not a contradiction — different body definitions (mine includes constructors and local functions) — but the denominator definition should be stated, since the proof's strength depends on it.

Separately, this is why nothing dangles in *code*: I grepped all 73 old class names across the entire merged tree and found only the 2 ADR markdown hits from SHOULD-FIX 1 — no string-literal or `nameof` references survive, which the compiler would not have caught.

### Skipped count — measured on both trees, not inferred

| suite | base | merged |
|---|---|---|
| `Interpreter.Tests` | 1346 passed / **3 skipped** / 1349 | 1346 / **3** / 1349 |
| `Core.Tests` | 7178 passed / **1 skipped** / 7179 | 7178 / **1** / 7179 |

**Identical on both sides; 4 skipped total, unchanged.** Consistent with the author's "4 skipped". Global `Skip =` count is 5 on both trees.

On ILVerify 123/123: agreed it means little alone — it has returned rc=0 this session while printing `Assembly or module not found: mscorlib`. I did not lean on it; my product check was `samples/Arithmetic.gs` run through `gsi` and through `gsc /out:` + `dotnet exec`, both rc=0 and both **byte-identical to the committed `Arithmetic.golden`**.

### Assertion strengthening — agreed, let it go

I am not asking for it. I looked for a weak assertion that is load-bearing for *this* issue's claims and did not find one: the assertion-kind histogram is identical across all 20 kinds, so nothing was weakened here, and this PR's claims are about **naming and documentation**, which the existing pins already carry. Declining optional strengthening is the correct ponytail call and should not be recorded as deferred work.

### New quantified evidence for SHOULD-FIX 3

Measuring the doc rewrite rather than eyeballing it changed my view of its severity. Of the 128 files:

- **108** received the template `Emitted-(oracle|session|execution) coverage for …`
- **72** have a summary mechanically derived from the filename, producing degenerate text: `"Emitted-oracle coverage for type of name of."` (from `TypeOfNameOfTests.cs`), `"…for generic function."`, `"…for base bcl call."`, `"…for from end index binding."`
- **32 files lost an ADR cross-reference entirely** — ADR-0004, ADR-0020, ADR-0022, ADR-0029, ADR-0047, ADR-0052, ADR-0084, ADR-0100, ADR-0117, ADR-0131 among them
- **12 files lost an issue `#ref`**
- **1 file lost a diagnostic reference** — `Issue798SharedStaticIteratorInterpreterTests.cs` no longer mentions `GS9998` anywhere

The 32 lost ADR references are the part I would most like reconsidered, and they are ironic against SHOULD-FIX 1: that finding is about **ADR → test** links dangling, while this one deletes **test → ADR** links. The sweep damages traceability in both directions at once.

The clearest single example, `Issue1104BasePropertyAccessBinderTests.cs`, replaced eight lines naming the exact diagnostics under test — `GS0157`, `GS0383`, `GS0384`, `GS0385` — plus the `base[BaseClass].Prop` explicit-ancestor form and "run without infinite recursion through the derived override", with `"Issue #1104: Emitted-oracle coverage for base property access."` None of the deleted text was engine-specific and all of it was still true.

The author already demonstrates the right pattern where the rewrite is hand-written rather than generated — `Issue3010EntryPointDriverMatrixTests.cs` became *"Driver-matrix coverage for issues #2984 and #3010; **every driver path uses emitted execution**"*, which is accurate, informative and strictly better than what it replaced. That is the standard the other 72 should meet.

Suggested minimal remedy, in ponytail order: **restore the ADR/issue/diagnostic references** (mechanical, 32+12+1 files, no judgement required), and leave the remaining prose alone. Restoring the identifiers recovers most of the lost value at a fraction of the effort of rewriting the summaries.

### Correction to my earlier comment

My first comment said "80 of 128 files received the templated one-liner". The correct figure is **108 of 128** (my earlier count matched a narrower literal). The 32-lost-ADR and 72-filename-derived figures are new measurements, not restatements.

PR #3277, opened 2026-08-05
Fix boxing for reified generic call arguments

### LINKED ISSUE #3271: [emit] Legal generic call with a reference-type argument, 'Run[object](42)', compiles rc=0 then throws InvalidProgramException at runtime

DavidObando:
## Summary
- resolve bare generic parameter slots to their call-site method or constructed-owner type arguments
- route value-type-to-reference targets through the canonical conversion emitter, matching the already-correct shared-method path
- add runtime coverage for top-level, instance, extension, constructed/inherited/nested owners, nullable targets, interfaces, `ValueType`, controls, and type-argument transposition

## Root cause
Shared methods already bind `int32 -> object` as a `BoundConversionExpression`, whose canonical emission supplies `box`. Top-level, extension, and instance generic paths deliberately left bare type-parameter arguments unconverted, expecting call-boundary boxing that no longer existed. Their shared emitter helper now resolves the reified target and materializes that same conversion only for object, interface, `ValueType`, or `Enum` reference targets.

## Validation
- exact repro: compile rc 0; `dotnet exec` rc 0; stdout `42`
- regression pins: 9/9 green; unfixed product: 7 failed with `InvalidProgramException`
- three condition-inversion product mutants killed; `GSharp.Core.dll` moved and restored exactly for each
- `dotnet build GSharp.sln`: 0 warnings, 0 errors
- all 9 test projects: 15,378 passed, 4 skipped
- ILVerify gate: 123/123 verified; 2 documented suppressions; 0 failures

## Separate observed defect
`Runner.Run[ValueType](42)` on the shared-method binder path still reports GS0155. CLR boxing from `int32` to `System.ValueType` is legal; this pre-existing binder inconsistency is unchanged by this emitter fix.

Fixes #3271

### DIFF
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Calls.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Calls.cs
@@ -121,6 +121,61 @@ internal sealed partial class MethodBodyEmitter
         this.il.Token(ctorHandle);
     }
 
+    private void EmitReifiedUserCallArgument(
+        BoundExpression argument,
+        ParameterSymbol parameter,
+        ImmutableArray<TypeParameterSymbol> methodTypeParameters,
+        ImmutableArray<TypeSymbol> methodTypeArguments,
+        StructSymbol constructedOwner)
+    {
+        if (parameter.RefKind != RefKind.None)
+        {
+            this.EmitExpression(argument);
+            return;
+        }
+
+        var targetType = constructedOwner?.SubstituteMemberType(parameter.Type) ?? parameter.Type;
+        if (targetType is TypeParameterSymbol targetParameter
+            && !methodTypeArguments.IsDefault
+            && !methodTypeParameters.IsDefault
+            && methodTypeArguments.Length == methodTypeParameters.Length)
+        {
+            for (var i = 0; i < methodTypeParameters.Length; i++)
+            {
+                if (ReferenceEquals(targetParameter, methodTypeParameters[i]))
+                {
+                    targetType = methodTypeArguments[i];
+                    break;
+                }
+            }
+        }
+
+        var sourceType = argument.Type;
+        var sourceNeedsBox = ReflectionMetadataEmitter.IsValueTypeSymbol(sourceType)
+            || sourceType is TypeParameterSymbol
+            || sourceType is NullableTypeSymbol { UnderlyingType: TypeParameterSymbol };
+        var unwrappedTarget = targetType is NullableTypeSymbol nullableTarget
+            && !ReflectionMetadataEmitter.IsValueTypeNullable(nullableTarget)
+                ? nullableTarget.UnderlyingType
+                : targetType;
+        var targetClr = unwrappedTarget?.ClrType;
+        var targetAcceptsBox = targetClr?.IsSameAs(typeof(object)) == true
+            || IsInterfaceTargetType(targetType)
+            || targetClr?.IsSameAs(typeof(ValueType)) == true
+            || targetClr?.IsSameAs(typeof(Enum)) == true;
+
+        if (sourceNeedsBox && targetAcceptsBox)
+        {
+            // Match the already-working shared-method path: materialize the
+            // same conversion node so canonical conversion emission supplies
+            // the required box opcode.
+            this.EmitExpression(new BoundConversionExpression(null, targetType, argument));
+            return;
+        }
+
+        this.EmitExpression(argument);
+    }
+
     private void EmitUserInstanceCall(BoundUserInstanceCallExpression call)
     {
         // Issue #1052: a call dispatched through a type parameter's
@@ -228,10 +283,24 @@ internal sealed partial class MethodBodyEmitter
 
         this.EmitInstanceReceiver(call.Receiver);
         var calleeParameterOffset = call.Method.ExplicitReceiverParameter == null ? 0 : 1;
+        var constructedOwner = inheritedGenericBase ?? receiverType;
+        var hasReifiedParameters = call.Method.IsGeneric || constructedOwner?.Definition != null;
         for (var i = 0; i < call.Arguments.Length; i++)
         {
             var arg = call.Arguments[i];
-            this.EmitExpression(arg);
+            if (hasReifiedParameters)
+            {
+                this.EmitReifiedUserCallArgument(
+                    arg,
+                    call.Method.Parameters[i + calleeParameterOffset],
+                    call.Method.TypeParameters,
+                    call.MethodTypeArguments,
+                    constructedOwner);
+            }
+            else
+            {
+                this.EmitExpression(arg);
+            }
         }
 
         var receiverIsValueType = call.Method.ReceiverType is StructSymbol receiverStruct && !receiverStruct.IsClass;
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs
@@ -132,11 +132,25 @@ internal sealed partial class MethodBodyEmitter
                 // `!!T`, so a value-type instantiation must be lifted to
                 // `Nullable<X>` — see TryPlanUnconstrainedNullableLift.
                 var nullableLift = this.TryPlanUnconstrainedNullableLift(call);
+                var hasReifiedParameters = call.Function.IsGeneric || call.StaticGenericOwnerType != null;
 
                 for (int i = 0; i < call.Arguments.Length; i++)
                 {
                     var arg = call.Arguments[i];
-                    this.EmitExpression(arg);
+                    if (hasReifiedParameters)
+                    {
+                        this.EmitReifiedUserCallArgument(
+                            arg,
+                            call.Function.Parameters[i],
+                            call.Function.TypeParameters,
+                            call.MethodTypeArguments,
+                            call.StaticGenericOwnerType);
+                    }
+                    else
+                    {
+                        this.EmitExpression(arg);
+                    }
+
                     if (nullableLift?.ArgumentWraps[i] is { } liftWrap)
                     {
                         this.EmitUnconstrainedNullableLiftWrap(liftWrap);
--- a/test/Core.Tests/CodeAnalysis/Emit/Issue3271ReifiedGenericArgumentBoxingTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Emit/Issue3271ReifiedGenericArgumentBoxingTests.cs
@@ -1,0 +1,343 @@
+// <copyright file="Issue3271ReifiedGenericArgumentBoxingTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using GSharp.Tests;
+using Xunit;
+
+namespace GSharp.Core.Tests.CodeAnalysis.Emit;
+
+/// <summary>
+/// Issue #3271: value-type arguments passed through a reified generic slot
+/// whose call-site type argument is a reference type must be boxed before the
+/// call. The shared-method binder path already materialized that conversion;
+/// free, extension, and instance calls emitted the raw value instead.
+/// </summary>
+public class Issue3271ReifiedGenericArgumentBoxingTests
+{
+    [Fact]
+    public void TopLevel_ValueTypes_BoxToObject()
+    {
+        AssertRuns(
+            """
+            package Issue3271TopObject
+            import System
+
+            struct Token { var Value int32 }
+
+            func Show[T](value T, label string) {
+                Console.WriteLine(label)
+                Console.WriteLine([]T{value}.GetType())
+                Console.WriteLine(value.GetType())
+            }
+
+            Show[object](11, "int")
+            Show[object](true, "bool")
+            Show[object](Token{Value: 7}, "struct")
+            """,
+            "int",
+            "System.Object[]",
+            "System.Int32",
+            "bool",
+            "System.Object[]",
+            "System.Boolean",
+            "struct",
+            "System.Object[]",
+            "Issue3271TopObject.Token");
+    }
+
+    [Fact]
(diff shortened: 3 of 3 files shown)


DavidObando:
## Duck verdict — NOT MERGEABLE as-is. 1 BLOCKER / 1 SHOULD-FIX / 1 NIT

`main` `f5642aff1`, head `8a8684c3a`, merge-base `0232ecb5e`. `merge-tree --write-tree pr3277 origin/main` → rc 0, tree `ffd9fcb95`, **zero unmerged stages, zero `CONFLICT` lines**. Rebase is clean.

Full suite on head, 9/9 projects, Release CI-flagged: **15,378 passed, 4 skipped, 0 failed** — matches the PR body exactly.

**The fix works.** Every claim about the defect and its repair reproduced. The problem is one disjunct in the guard list, and the layer.

### Repro + four shapes — out-of-process `gsc /target:exe /out:` then `dotnet exec`

Erasure-proof witness: the author's own `[]T{v}.GetType()` idiom, which reports the *reified* element type, not the value.

| shape | `main` | **#3277** | #3272 |
|---|---|---|---|
| free `Run[object](4201)` | rc0 → **rc134** `InvalidProgramException` | **rc0 → rc0** `System.Int32` / `4201` | rc0 → rc0 |
| instance `r.Run[object](4202)` | rc0 → **rc134** | **rc0 → rc0** | rc0 → rc0 |
| extension `"x".Run[object](4203)` | rc0 → **rc134** | **rc0 → rc0** | rc0 → rc0 |
| `shared` `Runner.Run[object](4204)` | rc0 → rc0 ✅ | rc0 → rc0 | rc0 → rc0 |

**#3277 alone fixes all four shapes**, and the "`shared` was already correct" premise is verified on `main`.

### Over-correction probes — all clean, including the one that mattered most

| probe | `main` | **#3277** | reading |
|---|---|---|---|
| `Run[int32](51)`, `Run[bool](true)` | `System.Int32[]` | **identical** | no spurious box |
| `Run[object]("str52")` | `System.Object[]` | **identical** | no double-convert |
| **open forwarding** `Outer[T](v){Inner[T](v)}`, `Outer[int32](61)` | rc134 | **`System.Int32[]`** | **open target does NOT box** |
| same, `OuterBox[T](v){Inner[object](v)}` `[int32](63)` | rc134 | `System.Object[]` | closed target boxes |
| `Two[int32,object](71,72)` / `Two[object,int32](73,74)` | rc134 | `Int32[] Object[]` / `Object[] Int32[]` | ordinal mapping correct, not transposed |
| `Run[Color]` / `Run[Enum]` / `Run[object]` on an enum | rc134 | `Color[]` / `Enum[]` / `Object[]` | enum unboxed only when reified to itself |
| `int32?` into `object` | rc134 | `System.Object[]` | correct |
| `[T IMark]` constrained, `Run[Mark]` | rc134 | `Mark[]`, `v.Code()`=91 | constraint target not boxed |

The open-forwarding row is safe for a reason worth recording: `TypeParameterSymbol` passes `base(name)` with no CLR type (`TypeParameterSymbol.cs:22`), so an unreified target has `ClrType == null` and fails all four disjuncts. That is load-bearing but undocumented at the guard.

### Mutant diagonal — positive control first, then three distinct cells

Baseline 9/9 green. `MethodBodyEmitter.Calls.cs`, md5 moved on every mutant, source restored to `git diff HEAD` = 0.

| mutant | RED | which rows |
|---|---|---|
| **PC** reachable `throw` at head of `EmitReifiedUserCallArgument` | **9/9** | site is observable — no survivals to misread |
| **M-a** `targetAcceptsBox` neutralised | **7/9** | all boxing rows; spares `ValueTypeTargets_StayUnboxed` + `ExistingSharedAndReferencePaths_StayCorrect` |
| **M-b** `targetAcceptsBox` widened to every closed target | **1/9** | exactly `ValueTypeTargets_StayUnboxed` |
| **M-c** ordinal map collapsed to `methodTypeArguments[0]` | **1/9** | exactly `ReifiedTypeArgumentOrder_IsNotErasedOrTransposed` |

A genuine diagonal, not three kills of one cell. Each RED is a runtime failure — `Assert.Empty() Failure: Collection: [Common Language Runtime detected an invalid program.]` at `Issue3271...Tests.cs:335` — not a compile error in the test. **The pins are load-bearing and well built.**

---

## BLOCKER — the `ValueType` disjunct widens the language on 3 of 4 paths, and the PR pins it

`MethodBodyEmitter.Calls.cs:164`: `|| targetClr?.IsSameAs(typeof(ValueType)) == true`

| specimen | `main` | **#3277** | #3272 |
|---|---|---|---|
| non-generic `func PlainVT(v ValueType)`; `PlainVT(202)` | **GS0154** | **GS0154** | **GS0154** |
| `shared` generic `Runner.Run[ValueType](101)` | **GS0155** | **GS0155** | **GS0155** |
| free generic `Run[ValueType](111)` | rc0 → rc134 | **rc0 → rc0, `System.ValueType[]`** | **GS0154** |

`int32 → System.ValueType` is refused **everywhere else in the language**, on every build I compiled. After #3277 G# gives **three different answers** for the same conversion: GS0154 non-generically, GS0155 on the `shared` generic path, and **silent success** on the free / instance / extension generic paths.

The PR discloses the `shared`-path GS0155 as "a pre-existing binder inconsistency", declared separate. I confirmed it *is* pre-existing (identical on `main`). But the disclosure is **inaccurate about this PR's effect**: #3277 does not leave the `ValueType` story alone — it changes it on three paths and **deepens** the split. And it pins the new behaviour in 3 of its 9 tests (`:51`, `:86`, `:120`, each asserting `System.ValueType[]`), which makes it expensive to walk back later.

Contrast `Enum`, which I checked the same way: `func PlainEnum(v Enum); PlainEnum(Color.Green)` compiles and runs rc 0 on **every** build. That disjunct is consistent with the language — keep it.

**Ponytail fix:** delete the `ValueType` disjunct at `Calls.cs:164` and the three `ValueType` assertions from the pins. `targetAcceptsBox` then collapses to `object || interface || Enum`, and `object || interface` is *literally* the existing canonical spelling at `MethodBodyEmitter.Conversions.cs:434`.

## SHOULD-FIX — wrong layer: #3272 closes the same class in three tokens, and gets the `ValueType` row right

I built #3272 (`bbfa5f32c`) and a combined merge, and ran the whole 13-probe matrix on all four builds.

- **#3272 alone fixes all four shapes**, identically to #3277.
- **`both` is byte-for-byte identical to `p3272` on every single row.** No double conversion, no regression — but also **#3277 contributes nothing observable on top of #3272**.
- `merge-tree --write-tree pr3277 pr3272` → rc 0, tree `93ca16ff6`, zero stages, zero `CONFLICT`. They merge textually clean while functionally overlapping.

#3272 does it with one added conjunct — the **existing canonical predicate** `TypeSymbol.ContainsTypeParameter(expectedType)` — at `OverloadResolver.CallBinding.cs:1707`, `OverloadResolver.Invocations.cs:1165`, `:1544`. #3277 instead adds a 55-line emit-time helper that synthesizes a `BoundConversionExpression` **at emit time** (`Calls.cs:167`) and hand-rolls a parallel-array `ReferenceEquals` scan at `Calls.cs:139-150`, where `StructSymbol.SubstituteTypeParameters` (`StructSymbol.cs:1643`) already exists. On the spelling question: #3277 does **not** reuse `ExpressionBinder.Access.Accessor.cs:3468`'s `substitution == null || !substitution.ContainsKey(...)`; it adds another site of the parallel-array idiom (already at `BoundScope.cs:2510`, `ExpressionBinder.Access.MemberLookup.cs:3590`, `SymbolDocumentationIdProvider.cs:480`). Not novel, but not the canonical call either.

Two more rows where the binder layer is simply better:
- `Gen[object](nil)`: `main`/#3277 **accept** it while non-generic `Plain(nil)` is **GS0154** on every build. #3272 makes the generic form agree. (My first read had this backwards; the non-generic control corrected me.)
- #3272 also replaces GS0156's ICE-inducing "are you missing a cast?" advice (`DiagnosticDescriptors.cs:75`). Nothing reachable from #3277's paths suggests a remedy — I only ever saw GS0154/GS0155, neither of which offers one — so there was nothing to follow here.

**Recommendation: take #3272's production change, and keep #3277's test file minus the three `ValueType` rows.** #3277's pins are the better artefact in this PR — a clean mutant diagonal over 9 specimens — and #3272 ships none of that coverage. Merging both production changes is safe but leaves 55 lines of dead emit-time machinery.

## NIT — pins are single-driver

`AssertRuns` (`Issue3271...Tests.cs:331`) uses only in-process `EmittedOracle.Evaluate`. It did catch the `InvalidProgramException`, so it is adequate; every measurement above was out-of-process `dotnet exec` and agreed with it.

---

### Separate finding, out of scope for this PR — please file

Generic **inference** accepts an argument incompatible with the type argument it already inferred, with the identical rc0 / zero-diagnostic / rc134 signature. Present on `main` and unchanged by **both** PRs:

```gs
func Show[T](a T, b T) { Console.WriteLine([]T{a, b}.GetType()) }
Show(131, "str")            // rc0 compile, zero diagnostics, rc134
```
ILVerify names it: `[StackUnexpected] ... [found ref 'string'][expected Int32]` — `T` was inferred `int32` from argument 0 and `string` was then accepted for argument 1. Same with user types: `Show(Mark{V:1}, Other())` → `[found ref Other][expected value Mark]`. The explicit form `Show[object](131, "str")` works fine on this head, which is what proves it is inference and not boxing.

DavidObando:
Closing in favour of PR #3272, on the strength of a measurement made by this PR's own author.

#3272 (issue #3267) repairs a broken mirror at `OverloadResolver.CallBinding.cs:1707` — one line — and that turns out to subsume this defect entirely. Measured on `e19c37cfb` plus the one-line repair, with `GSharp.Core.dll` md5 moving `57fc06e5801146e5e38ec08fb564f97c` → `b1b2278e48d2b10aa276f3254d9cba97` to prove a fresh build:

```
compile rc=0
run rc=0
stdout:
System.Object[]
System.Int32
42
```

Nine probe shapes × four call shapes (top-level, instance, extension, `shared`) all compile rc 0 with zero diagnostics and run rc 0: `int → object`, `int → IComparable`, unconstrained `T → object`, `T: struct → object`, `int → object?`, `int? → object`, `int? → int?`, `T?: struct → object`.

The root cause is better described by #3272's framing than by this PR's. The correct predicate spelling has existed at `Accessor.cs:3467` since #1379, which is why the `shared` and constructor paths never had this bug; the other sites drifted from it. This PR's emitter-layer conversion was therefore the wrong layer and wider than necessary — a binder that binds the conversion correctly needs no emitter compensation.

One genuine defect surfaced by this investigation is **not** covered by #3272 and has been filed separately: `Conversion.Classify` has no ordinary value-type → `System.ValueType` boxing conversion, and the call paths disagree about how to report it (`GS0154` free, `GS0155` instance/extension/`shared`). The mirror repair replaces the former `InvalidProgramException` with a diagnostic, which makes that classifier gap visible rather than causing it.

The 9 runtime tests here encode real requirements and should not be lost — see the note on #3272 about porting them.

PR #3281, opened 2026-08-05
Harden GetMethodSafe preconditions

### LINKED ISSUE #3278: [cleanup] ClrTypeUtilities.GetMethodSafe(name) is safe only by accident of its callers; document or harden the precondition
## Summary

`ClrTypeUtilities.GetMethodSafe(this Type, string)` — the 1-argument form — is **safe only because of its callers, not its implementation**. PR #3250 was the first change to violate that unstated precondition, and it shipped a build regression that CI did not catch.

Measured during review of #3250: a `ref struct` carrying **more than one** `ToString` overload compiled rc 0 and ran on the merge base, but failed on the PR head with:

```
GS9998: AmbiguousMatchException
```

Reverting only the single call site restored it. The single-`ToString` control was rc 0 on both sides, so the discriminator was specifically **an overloaded name**.

## Mechanism

`GetMethodSafe` wraps `Type.GetMethod(string)`, which throws `AmbiguousMatchException` when the name is overloaded. The helper catches `NotSupportedException` (its actual purpose — hardening against `TypeBuilder`-backed constructed generics) but **not** `AmbiguousMatchException`.

At the time of the regression there were **21 call sites; 20 passed `"Invoke"`**, which is unique by construction on a delegate, so ambiguity was impossible for them. Exactly one call site passed a non-`Invoke` name — and it was the one PR #3250 added.

PR #3250 (merged) fixed the immediate defect the right way: it added `GetMethodSafe(this Type, string, Type[])` at `src/Core/CodeAnalysis/ClrTypeUtilities.cs:77-98`, a faithful mirror of the 1-arg form with the parameter-type constraint restored. The helper now mirrors the BCL's own shape, where `GetMethod(name)` is ambiguity-prone and `GetMethod(name, types)` is not.

## What remains

**1. The 1-arg form still carries an undocumented precondition.** `ClrTypeUtilities.cs:45-66` swallows `NotSupportedException` but not `AmbiguousMatchException`, and nothing tells the next caller that the name must be unique. The next non-`Invoke` caller re-arms the same trap. One sentence of doc comment closes it — stating that the name must be unambiguous and directing overloaded lookups to the 3-arg form.

**2. The 3-arg form has the same accidental-safety shape.** `ClrTypeUtilities.cs:95-96` — the `TypeBuilder` fallback forwards *constructed* types to the *open* definition. This is fine for the sole current caller, which passes `Type.EmptyTypes` (the generic case was measured working end to end), but would silently return null for a hypothetical caller passing real parameter types. Unreachable today; worth either fixing or documenting so it does not become the next instance of this bug.

## Why this is worth an issue rather than a comment

The failure mode is a **compiler crash from a legal program**, invisible to CI because no existing test used an overloaded `ToString` on a by-ref-like type. The pattern — "this helper is safe because of who calls it" — is the actual defect class, and it has now cost one review round. Documenting or removing the precondition is a few lines.

## Scope

Documentation and/or a small hardening change in `src/Core/CodeAnalysis/ClrTypeUtilities.cs`. **Do not** change the 20 `"Invoke"` call sites; they are correct and unique by construction.

Evidence measured while reviewing #3250 at heads `7d795c322` and `93c5eb769`.

DavidObando:
Fixes #3278.

## Summary

- Document that the name-only `GetMethodSafe` overload requires an unambiguous method name and propagates `AmbiguousMatchException` otherwise.
- Match constructed parameter shapes recursively before the TypeBuilder-safe typed lookup fallback, including nested generics, arrays, by-ref and pointer types, and duplicate generic arguments by position.
- Map inherited generic methods through their constructed declaring type instead of assuming every resolved method is declared directly on the receiver.
- Add focused regression coverage using in-flight `TypeBuilder` arguments.

## Rationale

Keeping `AmbiguousMatchException` visible distinguishes caller misuse from a missing method. The typed fallback is hardened because returning null for a resolvable method would silently mislead future callers. Inherited methods need separate declaring-type mapping because `TypeBuilder.GetMethod` only accepts the generic type that actually declares the method.

### DIFF
--- a/src/Core/CodeAnalysis/Symbols/ClrTypeUtilities.cs
+++ b/src/Core/CodeAnalysis/Symbols/ClrTypeUtilities.cs
@@ -37,11 +37,13 @@ public static class ClrTypeUtilities
     /// <summary>
     /// Resolves a named method without asking a constructed
     /// <see cref="TypeBuilder"/> generic instantiation to resolve members
-    /// directly.
+    /// directly. The name must be unambiguous; use the parameter-type overload
+    /// for overloaded methods.
     /// </summary>
     /// <param name="type">The type that declares or inherits the method.</param>
     /// <param name="name">The method name to resolve.</param>
     /// <returns>The resolved method, or <see langword="null"/> when no matching method exists.</returns>
+    /// <exception cref="AmbiguousMatchException"><paramref name="name"/> identifies multiple methods.</exception>
     public static MethodInfo GetMethodSafe(this Type type, string name)
     {
         if (type is null || name is null)
@@ -92,7 +94,15 @@ public static class ClrTypeUtilities
                 return null;
             }
 
-            var openMethod = type.GetGenericTypeDefinition().GetMethod(name, types);
+            var openType = type.GetGenericTypeDefinition();
+            var typeArguments = type.GetGenericArguments();
+            var openTypeArguments = openType.GetGenericArguments();
+            var openParameterTypes = types.Select(parameterType =>
+            {
+                var index = Array.IndexOf(typeArguments, parameterType);
+                return index >= 0 ? openTypeArguments[index] : parameterType;
+            }).ToArray();
+            var openMethod = openType.GetMethod(name, openParameterTypes);
             return openMethod != null ? TypeBuilder.GetMethod(type, openMethod) : null;
         }
     }
--- a/src/Core/CodeAnalysis/Symbols/ClrTypeUtilities.cs
+++ b/src/Core/CodeAnalysis/Symbols/ClrTypeUtilities.cs
@@ -63,7 +63,7 @@ public static class ClrTypeUtilities
             }
 
             var openMethod = type.GetGenericTypeDefinition().GetMethod(name);
-            return openMethod != null ? TypeBuilder.GetMethod(type, openMethod) : null;
+            return openMethod != null ? GetConstructedGenericMethod(type, openMethod) : null;
         }
     }
 
@@ -105,7 +105,7 @@ public static class ClrTypeUtilities
                 types,
                 openTypeArguments,
                 typeArguments);
-            return openMethod != null ? TypeBuilder.GetMethod(type, openMethod) : null;
+            return openMethod != null ? GetConstructedGenericMethod(type, openMethod) : null;
         }
     }
 
@@ -1135,6 +1135,71 @@ public static class ClrTypeUtilities
         return type != null
             && type.IsConstructedGenericType
             && ContainsTypeBuilderGenericArgument(type);
+    }
+
+    private static MethodInfo GetConstructedGenericMethod(Type type, MethodInfo openMethod)
+    {
+        var declaringType = openMethod.DeclaringType;
+        if (!declaringType.IsGenericType)
+        {
+            return openMethod;
+        }
+
+        var declaringDefinition = declaringType.GetGenericTypeDefinition();
+        var methodDefinition = declaringType.IsGenericTypeDefinition
+            ? openMethod
+            : declaringDefinition.GetMethods()
+                .Single(method => method.Module == openMethod.Module
+                    && method.MetadataToken == openMethod.MetadataToken);
+        var openType = type.GetGenericTypeDefinition();
+        var constructedDeclaringType = SubstituteGenericTypeArguments(
+            declaringType,
+            openType.GetGenericArguments(),
+            type.GetGenericArguments());
+        return TypeBuilder.GetMethod(constructedDeclaringType, methodDefinition);
+    }
+
+    private static Type SubstituteGenericTypeArguments(
+        Type type,
+        Type[] openTypeArguments,
+        Type[] typeArguments)
+    {
+        if (type.IsGenericParameter)
+        {
+            var position = Array.IndexOf(openTypeArguments, type);
+            return position >= 0 ? typeArguments[position] : type;
+        }
+
+        if (type.HasElementType)
+        {
+            var elementType = SubstituteGenericTypeArguments(
+                type.GetElementType(),
+                openTypeArguments,
+                typeArguments);
+            if (type.IsByRef)
+            {
+                return elementType.MakeByRefType();
+            }
+
+            if (type.IsPointer)
+            {
+                return elementType.MakePointerType();
+            }
+
+            return type.IsSZArray
+                ? elementType.MakeArrayType()
+                : elementType.MakeArrayType(type.GetArrayRank());
+        }
+
+        return type.IsGenericType
+            ? type.GetGenericTypeDefinition().MakeGenericType(
+                type.GetGenericArguments()
+                    .Select(argument => SubstituteGenericTypeArguments(
+                        argument,
+                        openTypeArguments,
+                        typeArguments))
+                    .ToArray())
+            : type;
     }
 
     private static MethodInfo FindOpenGenericMethod(
--- a/src/Core/CodeAnalysis/Symbols/ClrTypeUtilities.cs
+++ b/src/Core/CodeAnalysis/Symbols/ClrTypeUtilities.cs
@@ -76,6 +76,8 @@ public static class ClrTypeUtilities
     /// <param name="name">The method name to resolve.</param>
     /// <param name="types">The method's parameter types.</param>
     /// <returns>The resolved method, or <see langword="null"/> when no matching method exists.</returns>
+    /// <exception cref="AmbiguousMatchException">More than one method matches the supplied signature.</exception>
+    /// <exception cref="ArgumentNullException"><paramref name="types"/> contains a <see langword="null"/> element.</exception>
     public static MethodInfo GetMethodSafe(this Type type, string name, Type[] types)
     {
         if (type is null || name is null || types is null)
@@ -97,12 +99,12 @@ public static class ClrTypeUtilities
             var openType = type.GetGenericTypeDefinition();
             var typeArguments = type.GetGenericArguments();
             var openTypeArguments = openType.GetGenericArguments();
-            var openParameterTypes = types.Select(parameterType =>
-            {
-                var index = Array.IndexOf(typeArguments, parameterType);
-                return index >= 0 ? openTypeArguments[index] : parameterType;
-            }).ToArray();
-            var openMethod = openType.GetMethod(name, openParameterTypes);
+            var openMethod = FindOpenGenericMethod(
+                openType,
+                name,
+                types,
+                openTypeArguments,
+                typeArguments);
             return openMethod != null ? TypeBuilder.GetMethod(type, openMethod) : null;
         }
     }
@@ -1133,6 +1135,109 @@ public static class ClrTypeUtilities
         return type != null
             && type.IsConstructedGenericType
             && ContainsTypeBuilderGenericArgument(type);
+    }
+
+    private static MethodInfo FindOpenGenericMethod(
+        Type openType,
+        string name,
+        Type[] parameterTypes,
+        Type[] openTypeArguments,
+        Type[] typeArguments)
+    {
+        var matches = openType.GetMethods()
+            .Where(method => string.Equals(method.Name, name, StringComparison.Ordinal))
+            .Where(method =>
+            {
+                var parameters = method.GetParameters();
+                return parameters.Length == parameterTypes.Length
+                    && parameters.Select(parameter => parameter.ParameterType)
+                    .Zip(parameterTypes, (open, constructed) =>
+                        MatchesConstructedParameterType(
+                            open,
+                            constructed,
+                            openTypeArguments,
+                            typeArguments))
+                    .All(matches => matches);
+            })
+            .Take(2)
+            .ToArray();
+
+        return matches.Length switch
+        {
+            0 => null,
+            1 => matches[0],
+            _ => throw new AmbiguousMatchException(),
+        };
+    }
+
+    private static bool MatchesConstructedParameterType(
+        Type open,
+        Type constructed,
+        Type[] openTypeArguments,
+        Type[] typeArguments)
+    {
+        if (open.IsGenericParameter)
+        {
+            var position = Array.IndexOf(openTypeArguments, open);
+            if (position >= 0)
+            {
+                return constructed == typeArguments[position];
+            }
+        }
+
+        if (open.IsByRef || constructed.IsByRef)
+        {
+            return open.IsByRef
+                && constructed.IsByRef
+                && MatchesConstructedParameterType(
+                    open.GetElementType(),
+                    constructed.GetElementType(),
+                    openTypeArguments,
+                    typeArguments);
+        }
+
+        if (open.IsPointer || constructed.IsPointer)
+        {
+            return open.IsPointer
+                && constructed.IsPointer
+                && MatchesConstructedParameterType(
+                    open.GetElementType(),
+                    constructed.GetElementType(),
+                    openTypeArguments,
+                    typeArguments);
+        }
+
+        if (open.IsArray || constructed.IsArray)
+        {
+            return open.IsArray
+                && constructed.IsArray
+                && open.IsSZArray == constructed.IsSZArray
+                && open.GetArrayRank() == constructed.GetArrayRank()
+                && MatchesConstructedParameterType(
+                    open.GetElementType(),
+                    constructed.GetElementType(),
+                    openTypeArguments,
+                    typeArguments);
+        }
+
+        if (open.IsConstructedGenericType || constructed.IsConstructedGenericType)
+        {
+            return open.IsConstructedGenericType
+                && constructed.IsConstructedGenericType
+                && open.GetGenericTypeDefinition() == constructed.GetGenericTypeDefinition()
+                && open.GetGenericArguments()
+                    .Zip(
+                        constructed.GetGenericArguments(),
+                        (openArgument, constructedArgument) =>
+                            MatchesConstructedParameterType(
+                                openArgument,
+                                constructedArgument,
+                                openTypeArguments,
+                                typeArguments))
+                    .All(matches => matches);
+        }
+
+        return open == constructed;
     }
 
     private static bool ContainsTypeBuilderGenericArgument(Type type)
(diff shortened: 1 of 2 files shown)


DavidObando:
## Evidence

### Base and scope

- Branched from `f5642aff1598630710f29737baf3416273b0dcf4`; a final fetch confirmed `origin/main` remained there.
- Current call sites: **21 total = 20 `Invoke` + 1 typed non-`Invoke` (`ToString`, `Type.EmptyTypes`)**.
- The 20 `Invoke` call sites are unchanged.
- Audit found no third open-definition accidental-safety fallback in `ClrTypeUtilities.cs`.

### Decisions

1. **Name-only overload:** documentation only. Its XML contract now requires an unambiguous name, directs overloaded lookups to the typed overload, and documents `AmbiguousMatchException`. No runtime pin was warranted for prose alone.
2. **Typed overload fallback:** fixed. Before querying the open generic definition, the fallback translates supplied constructed generic arguments to their corresponding open generic parameters. A future caller now receives the resolvable `MethodInfo` rather than an exception/null result.
3. **`AmbiguousMatchException`:** intentionally propagated. Returning null would conflate caller misuse with “method absent” and hide the same precondition violation this issue documents; throwing fails at the misuse site.

### Regression pin and RED proof

`GetMethodSafe_TypeBuilderConstructedGeneric_ResolvesParameterType` constructs a generic type with two `Accept` overloads using an in-flight `TypeBuilder` argument, proves raw reflection throws, then passes the real parameter type to `GetMethodSafe` and asserts resolution.

- Fixed: targeted test **1/1 passed**.
- Product hunk reverted alone: targeted test **failed**, rc 1, with `ArgumentException: Type must be a type provided by the runtime. (Parameter 'types')`.
- Fixed md5: `ecec508614ae931a23d60aedd7bba0ee`.
- Mutant md5: `d082b721a927b287bc7a9298a607ee0f`.
- Restored md5: `ecec508614ae931a23d60aedd7bba0ee`.
- After restore, both `git diff HEAD` and `git diff --cached HEAD` were empty.
- Full `ClrTypeUtilitiesTests`: **19/19 passed**.

No emitted program is involved; the reflection helper itself is the executable behavior under test.

### Build and full suite

Release serial solution build used `-m:1`, `ContinuousIntegrationBuild=true`, and warnings-as-errors:

- rc 0
- **0 warnings, 0 errors**
- `GSharp.Core.dll`: 2,489,344 bytes, md5 `36e2cac3a41c44712b8278c7e25dd674`
- Identical md5 across all 20 Release consumers.

Nine test projects, each rebuilt in Release with warnings-as-errors:

- Core: 7,179 passed, 1 skipped
- Compiler: 4,267 passed
- Interpreter: 1,353 passed, 3 skipped
- LanguageServer: 468 passed
- SDK: 59 passed
- Extensions: 104 passed
- Cs2Gs: 1,907 passed
- GeneratorHost: 31 passed
- InternalAnalyzers: 9 passed
- **Total: 15,377 passed, 4 skipped, 0 failed; 9/9 projects**

DavidObando:
## Review — PR #3281 (issue #3278), head `4cce64d9f3ca4711b42d2199d804779ceec09572`

**Verdict: MERGEABLE.** No BLOCKER. **2 SHOULD-FIX, 2 NIT.** Nothing reachable from any of the 21 call sites changes behaviour, and I reproduced the author's headline build witness exactly. But residue 2 is only **partly** discharged, and in one measured shape the PR replaces a loud failure with a silent wrong answer — so I would not close #3278 on this PR without either finishing the mapping or documenting what it does not cover.

SHAs I fetched myself: `main` = `acc10fe47056db3d2c707bd3e00a7aeb25a689df` at the start, **`7341a6e688b6d4f1602dfaead0aca6fb1004e093` on a re-fetch just before posting** (#3274 landed mid-review; it touches no call site and no `GetMethodSafe` text, and `merge-tree` against it is still **0 stage entries**, tree `9230a2ed1`); head = `4cce64d9f3ca4711b42d2199d804779ceec09572`; own merge-base = `f5642aff1598630710f29737baf3416273b0dcf4` (= #3250, one commit behind `main`). One commit on the branch. `merge-tree` vs `main`: tree `016078f38`, **0 stage entries**.

---

## Independent reproduction of the build witness

I built with the command the author named (`dotnet build GSharp.sln -c Release -m:1 -p:ContinuousIntegrationBuild=true`) and got `GSharp.Core.dll` md5 **`36e2cac3a41c44712b8278c7e25dd674`** — **byte-identical to the published value**, rc 0, **0 warnings**, and identical across all 22 Release consumers. Naming the build command is what made that checkable; thank you.

Worth recording for the fleet: with `ContinuousIntegrationBuild=true`, `GSharp.Core.dll` md5 is reproducible **across worktrees** (I got the same hash in two independent trees). Without it I got `3cb523d9e6b994d6dddcdd66ee480475` — a third value, matching neither published hash. The `ecec5086…`/`d082b721…` pair in the RED-proof section carries no build command and is therefore not reproducible by a reviewer; only the named one is. Please name the command next to every published hash.

---

## Residue 2 — the substitution. This is the only real code in the PR, and it is top-level-identity-only.

`ClrTypeUtilities.cs:99-103` maps a supplied parameter type to an open type parameter by **value identity at the top level**:

```csharp
var index = Array.IndexOf(typeArguments, parameterType);
return index >= 0 ? openTypeArguments[index] : parameterType;
```

A parameter type that *contains* the type argument rather than *being* it does not match, is passed through unchanged, and then reaches `openType.GetMethod(name, …)` still carrying a `TypeBuilder`-backed type.

I built a harness linked against each tree's own `GSharp.Core.dll` (staged beside the Compiler output so `MetadataLoadContext` resolves) and called the real extension method. Every row is a `TypeBuilderInstantiation` receiver; `raw` is `Type.GetMethod(name, types)` directly, which is `NotSupportedException` in all 16 rows.

| # | receiver / lookup | base `f5642aff1` | head `4cce64d9f` | |
|---|---|---|---|---|
| 1 | `Box<A>.Accept(T)` ← `[A]` | THROW `ArgumentException` | **RESOLVED `Accept(T)`** | ✅ fixed |
| 2 | `Box<A>.AcceptList(List<T>)` ← `[List<A>]` | THROW `ArgumentException` | **THROW `ArgumentException`** | ❌ unfixed |
| 3 | `Box<A>.AcceptArray(T[])` ← `[A[]]` | THROW `ArgumentException` | **THROW `ArgumentException`** | ❌ unfixed |
| 4 | `Box<A>.AcceptFunc(Func<T>)` ← `[Func<A>]` | THROW `ArgumentException` | **THROW `ArgumentException`** | ❌ unfixed |
| 5 | `AcceptConcrete(List<int>)` ← `[List<int>]` | RESOLVED | RESOLVED | = |
| 6 | `AcceptConcrete(string)` ← `[string]` | RESOLVED | RESOLVED | = |
| 7 | `Nullary()` ← `Type.EmptyTypes` | RESOLVED | RESOLVED | = **the only real caller's shape** |
| 8 | `Accept` ← `[A,A]` (wrong arity) | NULL | NULL | = correct negative |
| 9 | `Nope` ← `[A]` (absent) | NULL | NULL | = correct negative |
| 10 | `Pair<A,A>.First(T)` ← `[A]` | THROW | **RESOLVED `First(T)`** | ✅ |
| 11 | `Pair<A,A>.Second(U)` ← `[A]` | THROW | **NULL** | ⚠️ **wrong** |
| 12 | `Pair<A,A>.Both(T,U)` ← `[A,A]` | THROW | **NULL** | ⚠️ **wrong** |
| 13 | `Pair<A,B>.First(T)` ← `[A]` | THROW | RESOLVED | ✅ |
| 14 | `Pair<A,B>.Second(U)` ← `[B]` | THROW | RESOLVED | ✅ |
| 15 | `Pair<A,B>.Both(T,U)` ← `[A,B]` | THROW | RESOLVED | ✅ |
| 16 | `Pair<A,B>.Both(U,T)` ← `[B,A]` | THROW | NULL | ✅ correct negative |

Rows 5–9 identical on both sides. **Row 7 is the shape the sole production caller uses (`InterpolatedStringHandlerLowerer.cs:698`, `Type.EmptyTypes`) and it is unchanged** — hence no reachable behaviour change, hence no BLOCKER.

### SHOULD-FIX 1 — `ClrTypeUtilities.cs:99-103`: nested parameter types still throw out of a method named `…Safe`

Rows 2/3/4. `List<T>`, `T[]` and `Func<T>` are the ordinary shapes a real caller would pass, and for all three the helper still throws `ArgumentException: Type must be a type provided by the runtime` — escaping a method whose `<returns>` at `:74` promises "the resolved method, or `null` when no matching method exists". The hole was narrowed from "every real parameter type" to "every real parameter type except a bare top-level `T`"; the *shape* issue #3278 exists to remove is still there.

**The house pattern for this is already in the same file.** `ClrTypeUtilities.RemapHostCoreTypeToContext` at `:1002-1050` is a recursive `Type`-tree mapper that already handles precisely the missing cases — `IsByRef` → `MakeByRefType()` (`:1011`), `IsPointer` → `MakePointerType()` (`:1016`), `IsArray` → element remap + `MakeArrayType(rank)` (`:1021`), `IsConstructedGenericType` → recursive `Select` + `MakeGenericType` (`:1029-1041`). The new mapping is a flat `Array.IndexOf` roughly 900 lines above an existing recursive mapper in the same class. Recommend a small local recursive mapper in that shape, or — the lazier option the issue explicitly offered — leave the code alone and **document** that only top-level type-parameter substitution is supported.

### SHOULD-FIX 2 — `ClrTypeUtilities.cs:101`: `Array.IndexOf` returns the *first* index, so duplicate type arguments mis-map

Rows 11 and 12. When a generic type is constructed with the **same** argument twice, `Array.IndexOf(typeArguments, parameterType)` yields index 0 for every parameter, so `Pair<A,A>.Second(U)` is looked up as `Second(T)` and `Both(T,U)` as `Both(T,T)`; neither exists, so the helper returns `null`.

I confirmed the failure is for the reason I think, and that the method is genuinely resolvable:

```
open.GetMethod("First",[T])   = FOUND   TypeBuilder.GetMethod -> OK:First
open.GetMethod("Second",[U])  = FOUND   TypeBuilder.GetMethod -> OK:Second
open.GetMethod("Both",[T,U])  = FOUND   TypeBuilder.GetMethod -> OK:Both

GetMethodSafe(Pair<A,A>, "First",  [Argument])          = RESOLVED:First
GetMethodSafe(Pair<A,A>, "Second", [Argument])          = NULL
GetMethodSafe(Pair<A,A>, "Both",   [Argument,Argument]) = NULL
```

So a positionally-correct mapping resolves all three; the shipped mapping silently returns `null` for two of them. On the base these same lookups **threw**. This PR therefore converts a loud failure into a silent wrong answer for this shape — the strictly worse direction, and the exact outcome the PR description argues against: *"returning null for a resolvable method would silently mislead future callers."* The shipped code does that.

Fix: map by **position in the open method's signature** rather than by value identity — or recurse as in `RemapHostCoreTypeToContext`.

---

## Did the required real-parameter-type caller get written? **Yes.**

`ClrTypeUtilitiesTests.cs:154-176` constructs `TypeBuilderMethodFixture<>` over an in-flight `TypeBuilder`, asserts the receiver is a `TypeBuilderInstantiation`, asserts raw reflection throws `NotSupportedException`, then calls `GetMethodSafe(name, [argument])` — a **real parameter type**, not `Type.EmptyTypes` — and asserts resolution. The fixture carries two `Accept` overloads, so it also exercises the disambiguation the 3-arg overload exists for. This is the caller the brief required, and it is a genuine pin, not another `EmptyTypes` case wearing a new name.

### NIT 1 — the pin observes only single-type-parameter substitution

Two-sided PRODUCT mutants in a dedicated tree; every one asserted by non-empty `git diff HEAD` on a real statement **and** within-tree `GSharp.Core.dll` md5 movement, with build rc read before scoring. Baseline `36e2cac3…`.

| mutant | `file:line` | change | md5 | result |
|---|---|---|---|---|
| A | `ClrTypeUtilities.cs:103` | `index >= 0 ? openTypeArguments[index]` → `index < 0 ? openTypeArguments[0]` | `8bde6674…` | **KILLED** — 1F/18P |
| B | `ClrTypeUtilities.cs:102` | `IndexOf(...)` → `IndexOf(...) >= 0 ? 0 : -1` (collapse index to 0) | `7cf18829…` | **SURVIVED** |

Mutant B survived the **full** `Core.Tests` run — 7,179 passed, 0 failed, not a filtered green. Mutant A sits on the adjacent expression of the same statement and killed, which is my reachability control: the code is definitively executed and observed, so B's survival is a **coverage measurement**, not a deadness claim, and I am **not** suggesting anything be removed. It says only this: with one type parameter the index is always 0, so nothing in the corpus can tell a correct index from a constant 0 — which is the same blind spot that produced rows 11 and 12. A second fixture with two distinct type parameters would close it.

Restore: `git checkout --` the file, `git diff HEAD` and `git diff --cached HEAD` both empty, rebuild rc 0, md5 back to **`36e2cac3a41c44712b8278c7e25dd674`** — exact round-trip.

---

## `AmbiguousMatchException` decision — consistent, and the doc is true of the shipped code

Propagating is a defensible call and it matches what the code actually does. Measured on both sides, unchanged by this PR:

```
1-arg, overloaded name, runtime type      -> THROW AmbiguousMatchException   (base and head)
1-arg, overloaded name, TypeBuilder type  -> THROW AmbiguousMatchException   (base and head)
1-arg, unique name,     TypeBuilder type  -> RESOLVED                        (base and head)
```

The doc at `:40-41` states the precondition, points at the parameter-type overload, and `:46` declares `<exception cref="AmbiguousMatchException">`. All three of the things the brief asked for are present and all three are true of the shipped binary. The 1-arg body is textually unchanged, so **no existing caller can now throw where it previously could not** — confirmed by the identical base/head rows above.

### NIT 2 — the doc was fixed on the overload that did not change, and left wrong on the one that did

`:70-78`, the 3-arg overload's doc, still says only "the resolved method, or `null` when no matching method exists" and carries no `<exception>` tag — while rows 2/3/4 show it throwing `ArgumentException`. Given #3278 is literally "document or harden the precondition", the changed overload is the one most in need of the sentence. One `<exception>` line plus a note that substitution is top-level-only would discharge residue 2 honestly even without SHOULD-FIX 1.

---

## Call-site audit — 21 / 20 / 1, unchanged, and no in-flight PR adds one

21 `GetMethodSafe` call sites in `src/`: **20 pass `"Invoke"`** (`Conversion.cs:2264,2303`; `LambdaBinder.cs:1133`; `ExpressionBinder.cs:1048`; `ClrOverloadResolution.cs:1297,1703,1704,2099,2165`; `OverloadResolver.CallBinding.cs:1065`; `ExpressionBinder.Operators.cs:2689`; `ExpressionBinder.Calls.Invocation.cs:1023,1711`; `ExpressionBinder.Calls.Arguments.cs:802`; `MemberLookup.cs:2415,5000,5504`; `ConversionClassifier.cs:1411,2466`; `DelegateRefKindUtilities.cs:40`) — all on the **1-arg** overload, all unique by construction on a delegate.

The **1 non-`Invoke`** caller is `InterpolatedStringHandlerLowerer.cs:698`, and it uses the **3-arg** overload with `Type.EmptyTypes`. It therefore cannot reach the 1-arg overload's `AmbiguousMatchException` path at all: parameter types disambiguate, and the lookup is a nullary `ToString` on a declared-only struct member. This PR does not convert any latent exposure into a live one.

I also checked every open PR the coordinator listed. Only #3272 and #3260 touch files containing call sites, and **neither adds or changes a `GetMethodSafe` call site** (`git diff … | grep '^[-+].*GetMethodSafe'` → empty for both). `merge-tree` in **both orders**, reading stage entries rather than `CONFLICT` counts:

| pair | tree | stage entries |
|---|---|---|
| #3281 × `main` `acc10fe47` | `016078f38` | **0** |
| #3281 × `main` `7341a6e68` (post-#3274) | `9230a2ed1` | **0** |
| #3281 × #3272 (`d259a8df7`) | `1ed61a881` | **0** both orders |
| #3281 × #3260 (`db55ab28a`) | `9ec6955ea` | **0** both orders |
| #3281 × #3276 (`bbb418093`) | `e1a5fbb39` | **0** both orders |

#3274, #3275, #3262, #3264 touch no call-site file.

---

## Test accounting — the delta is **+1**, not +4

I measured the merge-base myself rather than differencing against a remembered figure:

| | base `f5642aff1` | head `4cce64d9f` | delta |
|---|---:|---:|---:|
| Core.Tests passed | **7,178** | **7,179** | **+1** |
| Core.Tests skipped | 1 (`RegenerateBaseline`) | 1 (`RegenerateBaseline`) | 0 |

The diff adds exactly **one** `[Fact]` and **zero** `[InlineData]`, so +1 is the whole story and it is the new pin. Nothing was silently skipped — the skip count is unchanged and the one skip is pre-existing. The "4" in the report is the **skipped** total across the solution (1 Core + 3 Interpreter, both pre-existing), not a test-count delta; head total is 15,377 passed against a base total of 15,376.

## Suites — denominator 9, from `dotnet sln GSharp.sln list`

| project | passed | failed | skipped | rc |
|---|---:|---:|---:|---:|
| Core.Tests | 7,179 | 0 | 1 | 0 |
| Compiler.Tests | 4,267 | 0 | 0 | 0 |
| Cs2Gs.Tests | 1,907 | 0 | 0 | 0 |
| Interpreter.Tests | 1,353 | 0 | 3 | 0 |
| LanguageServer.Tests | 468 | 0 | 0 | 0 |
| Extensions.Tests | 104 | 0 | 0 | 0 |
| Sdk.Tests | 59 | 0 | 0 | 0 |
| GeneratorHost.Tests | 31 | 0 | 0 | 0 |
| InternalAnalyzers.Tests | 9 | 0 | 0 | 0 |
| **total** | **15,377** | **0** | **4** | 9/9 |

---

Residue 1 is discharged cleanly and the doc is accurate. Residue 2 is where I would push back: the change is real and the pin is real, but four of the shapes a future caller would plausibly pass still fail — three loudly, two silently and *newly* so — and there is a recursive type-mapper in the same file that already knows how to do this. Either finish it in that shape or state the limitation in the doc; otherwise #3278 closes while the pattern it names still lives in the function it names.

DavidObando:
Follow-up ownership pass found and fixed one remaining inherited-member hole at head 1554a7bac: both GetMethodSafe overloads could match a method inherited from a generic base, then throw ArgumentException by passing the derived constructed type to TypeBuilder.GetMethod.

Pushed 92cde5c61, which reconstructs the method's actual generic declaring type before mapping it. Added a regression covering both name-only and typed inherited lookups.

Validation:
- Focused ClrTypeUtilitiesTests: 23 passed, 0 failed.
- Full Core.Tests: 7,184 passed, 1 pre-existing skip, 0 failed.
- git diff --check: clean.

PR #3283, opened 2026-08-05
Reject inherited calls that box ref structs

### LINKED ISSUE #3240: gsc emits invalid IL for explicit ToString() on by-ref-like receiver with only inherited ToString

DavidObando:
Fixes #3240.

## Summary
- centralize the emitter's imported value-receiver boxing decision and reuse it while binding inherited CLR calls
- report existing GS0219 before inherited dispatch, interface conversion, inferred generic use, or method-group capture can box a by-ref-like value
- preserve declared ref-struct direct calls, plain structs, imported declared methods, and direct interface/extension calls
- add 17 compile/no-assembly/runtime regression cases with diagnostic ID and source-span assertions, including all four method-group shapes from review

## Validation
- Issue #3240 regression tests: 17 passed
- method-group regression tests: 50 passed
- Release warnings-as-errors solution build: 0 warnings, 0 errors
- GitHub checks: running on 4d970c0fb

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ConversionClassifier.cs
+++ b/src/Core/CodeAnalysis/Binding/ConversionClassifier.cs
@@ -660,8 +660,7 @@ internal sealed class ConversionClassifier
         // the same ref struct) already returned above.
         if (TypeSymbol.IsByRefLike(expression.Type)
             && type != TypeSymbol.String
-            && type?.ClrType != null
-            && !type.ClrType.IsValueType
+            && Conversion.IsReferenceLikeTarget(type)
             && expression.Type != TypeSymbol.Error)
         {
             Diagnostics.ReportByRefLikeEscape(diagnosticLocation, expression.Type, $"be boxed or converted to the reference type '{type}'");
--- a/src/Core/CodeAnalysis/Binding/ConversionClassifier.cs
+++ b/src/Core/CodeAnalysis/Binding/ConversionClassifier.cs
@@ -413,6 +413,24 @@ internal sealed class ConversionClassifier
             return new BoundDefaultExpression(barePlaceholder, type);
         }
 
+        var methodGroupReceiver = expression switch
+        {
+            BoundMethodGroupExpression userGroup => userGroup.Receiver,
+            BoundClrMethodGroupExpression clrGroup => clrGroup.Receiver,
+            _ => null,
+        };
+        var methodGroupReceiverType = methodGroupReceiver?.Type is ByRefTypeSymbol byRefReceiver
+            ? byRefReceiver.PointeeType
+            : methodGroupReceiver?.Type;
+        if (methodGroupReceiverType != null && TypeSymbol.IsByRefLike(methodGroupReceiverType))
+        {
+            Diagnostics.ReportByRefLikeEscape(
+                diagnosticLocation,
+                methodGroupReceiverType,
+                "be captured as a delegate target because that requires boxing the receiver");
+            return new BoundErrorExpression(null);
+        }
+
         // Issue #337: a CLR member method group has no fixed type until the
         // target delegate signature drives overload selection. Resolve it here,
         // where the expected type is known, before classifying conversions.
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Arguments.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Arguments.cs
@@ -1124,7 +1124,14 @@ internal sealed partial class ExpressionBinder
                 var inheritedArguments = OverloadResolver.BuildOrderedCallArguments(inheritedConvertedArgs, inheritedDownstreamMapping, inheritedParameters);
                 var refKinds = ComputeArgumentRefKinds(inheritedParameters);
                 overloads.ValidateRefArguments(inheritedArguments, refKinds, methodName, ce.Location);
-                BoundExpression inheritedCall = new BoundImportedInstanceCallExpression(null, inheritedUpdatedReceiver ?? receiver, resolution.Best, returnType, inheritedArguments, refKinds, inheritedTypeArgSymbolsForCall, isNonVirtualBaseCall: nonVirtualBaseCall);
+                var inheritedCallReceiver = inheritedUpdatedReceiver ?? receiver;
+                if (TryReportByRefLikeInheritedCall(inheritedCallReceiver, resolution.Best, ce.Location))
+                {
+                    result = new BoundErrorExpression(null);
+                    return true;
+                }
+
+                BoundExpression inheritedCall = new BoundImportedInstanceCallExpression(null, inheritedCallReceiver, resolution.Best, returnType, inheritedArguments, refKinds, inheritedTypeArgSymbolsForCall, isNonVirtualBaseCall: nonVirtualBaseCall);
                 result = WrapWithHandlerPrelude(inheritedCall, inheritedHandlerPrelude, ce);
                 return true;
             case ClrOverloadResolution.ResolutionOutcome.Ambiguous:
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs
@@ -26,6 +26,27 @@ namespace GSharp.Core.CodeAnalysis.Binding;
 
 internal sealed partial class ExpressionBinder
 {
+    private bool TryReportByRefLikeInheritedCall(
+        BoundExpression receiver,
+        MethodInfo method,
+        TextLocation location)
+    {
+        var receiverType = receiver.Type is ByRefTypeSymbol byRef
+            ? byRef.PointeeType
+            : receiver.Type;
+        if (!TypeSymbol.IsByRefLike(receiverType)
+            || !ReflectionMetadataEmitter.ImportedInstanceReceiverRequiresBoxing(receiverType, method))
+        {
+            return false;
+        }
+
+        Diagnostics.ReportByRefLikeEscape(
+            location,
+            receiverType,
+            $"invoke inherited method '{method.Name}' because dispatch requires boxing the receiver");
+        return true;
+    }
+
     /// <summary>
     /// Issue #311: resolves the explicit <c>[T1, T2]</c> type-argument list on a
     /// generic-method call site into CLR types projected onto the reference load
@@ -3039,7 +3060,13 @@ internal sealed partial class ExpressionBinder
                         var instArguments = OverloadResolver.BuildOrderedCallArguments(instConvertedArgs, instDownstreamMapping, instParameters);
                         var instRefKinds = ComputeArgumentRefKinds(instParameters);
                         overloads.ValidateRefArguments(instArguments, instRefKinds, methodName, ce.Location);
-                        BoundExpression instCall = ConversionClassifier.AutoDereferenceRefReturn(new BoundImportedInstanceCallExpression(null, instUpdatedReceiver ?? receiver, resolution.Best, returnType, instArguments, instRefKinds, instTypeArgSymbolsForCall));
+                        var callReceiver = instUpdatedReceiver ?? receiver;
+                        if (TryReportByRefLikeInheritedCall(callReceiver, resolution.Best, ce.Location))
+                        {
+                            return new BoundErrorExpression(null);
+                        }
+
+                        BoundExpression instCall = ConversionClassifier.AutoDereferenceRefReturn(new BoundImportedInstanceCallExpression(null, callReceiver, resolution.Best, returnType, instArguments, instRefKinds, instTypeArgSymbolsForCall));
                         return WrapWithHandlerPrelude(instCall, instHandlerPrelude, ce);
                     case ClrOverloadResolution.ResolutionOutcome.Ambiguous:
                         Diagnostics.ReportAmbiguousOverload(ce.Location, methodName, resolution.Ambiguous.Length, resolution.Ambiguous.Select(ClrOverloadResolution.FormatMethodSignature));
--- a/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.CallBinding.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.CallBinding.cs
@@ -1353,6 +1353,11 @@ internal sealed partial class OverloadResolver
             var constraintLocation = syntax.TypeArgumentList != null
                 ? syntax.TypeArgumentList.Location
                 : syntax.Identifier.Location;
+            if (TryReportByRefLikeTypeArgument(function.TypeParameters, substitution, constraintLocation))
+            {
+                return new BoundErrorExpression(null);
+            }
+
             foreach (var tp in function.TypeParameters)
             {
                 var typeArg = substitution[tp];
--- a/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.Invocations.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.Invocations.cs
@@ -1129,6 +1129,11 @@ internal sealed partial class OverloadResolver
             var constraintLocation = ce.TypeArgumentList != null
                 ? ce.TypeArgumentList.Location
                 : ce.Identifier.Location;
+            if (TryReportByRefLikeTypeArgument(extension.TypeParameters, substitution, constraintLocation))
+            {
+                return new BoundErrorExpression(null);
+            }
+
             foreach (var tp in extension.TypeParameters)
             {
                 var typeArg = substitution[tp];
@@ -1439,6 +1444,11 @@ internal sealed partial class OverloadResolver
             var constraintLocation = ce.TypeArgumentList != null
                 ? ce.TypeArgumentList.Location
                 : ce.Identifier.Location;
+            if (TryReportByRefLikeTypeArgument(method.TypeParameters, substitution, constraintLocation))
+            {
+                return new BoundErrorExpression(null);
+            }
+
             foreach (var tp in method.TypeParameters)
             {
                 var typeArg = substitution[tp];
--- a/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.cs
@@ -63,6 +63,26 @@ namespace GSharp.Core.CodeAnalysis.Binding.OverloadResolution;
 /// </remarks>
 internal sealed partial class OverloadResolver
 {
+    private bool TryReportByRefLikeTypeArgument(
+        ImmutableArray<TypeParameterSymbol> typeParameters,
+        Dictionary<TypeParameterSymbol, TypeSymbol> substitution,
+        TextLocation location)
+    {
+        foreach (var typeParameter in typeParameters)
+        {
+            var typeArgument = substitution[typeParameter];
+            if (!TypeSymbol.IsByRefLike(typeArgument))
+            {
+                continue;
+            }
+
+            Diagnostics.ReportByRefLikeEscape(location, typeArgument, "be used as a generic type argument");
+            return true;
+        }
+
+        return false;
+    }
+
     /// <summary>
     /// Custom delegate type for the <c>TryBindClrConstructorCall</c>
     /// callback, required because <see cref="Func{T1, T2, TResult}"/>
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs
@@ -288,10 +288,9 @@ internal sealed partial class MethodBodyEmitter
                     // pointer to the value; without the box the raw value bits
                     // are reinterpreted as a reference, producing an
                     // AccessViolationException (or silent corruption) at runtime.
-                    var declaringType = instCall.Method.DeclaringType;
-                    var receiverNeedsBox = receiverIsValueType
-                        && declaringType != null
-                        && !declaringType.IsValueType;
+                    var receiverNeedsBox = ReflectionMetadataEmitter.ImportedInstanceReceiverRequiresBoxing(
+                        receiverType,
+                        instCall.Method);
 
                     // A value type calling a method it declares itself receives a
                     // managed pointer (`this` is `ref TStruct`) and uses `call`;
--- a/src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs
@@ -5111,6 +5111,17 @@ internal sealed class ReflectionMetadataEmitter
         return false;
     }
 
+    internal static bool ImportedInstanceReceiverRequiresBoxing(TypeSymbol receiverType, MethodInfo method)
+    {
+        if (receiverType is ByRefTypeSymbol byRef)
+        {
+            receiverType = byRef.PointeeType;
+        }
+
+        return IsValueTypeSymbol(receiverType)
+            && method?.DeclaringType is { IsValueType: false };
+    }
+
     // Issue #671 (construction-call follow-up): a generic type-argument
     // position carries a "user-defined" symbol when it is itself a
     // user-declared type (Struct/Class/Interface/Enum/Delegate) — its
--- a/test/Compiler.Tests/Emit/Issue3240RefStructInheritedVirtualBoxingTests.cs
+++ b/test/Compiler.Tests/Emit/Issue3240RefStructInheritedVirtualBoxingTests.cs
@@ -186,6 +186,90 @@ public sealed class Issue3240RefStructInheritedVirtualBoxingTests
     }
 
     [Fact]
+    public void InheritedMethodGroup_OnRefStruct_ReportsBoxing()
+    {
+        const string Source = """
+            ref struct Bare {
+                var id int32
+            }
+
+            func Run() {
+                var b = Bare{id: 42}
+                let f () -> string = b.ToString
+            }
+            Run()
+            """;
+
+        AssertRejectedWithoutAssembly(
+            Source,
+            "b.ToString",
+            "be captured as a delegate target because that requires boxing the receiver");
+    }
+
+    [Fact]
+    public void DeclaredMethodGroup_OnRefStruct_ReportsBoxing()
+    {
+        const string Source = """
+            ref struct Bare {
+                var id int32
+                func Get() int32 -> id
+            }
+
+            func Run() {
+                var b = Bare{id: 42}
+                let f () -> int32 = b.Get
+            }
+            Run()
+            """;
+
+        AssertRejectedWithoutAssembly(
+            Source,
+            "b.Get",
+            "be captured as a delegate target because that requires boxing the receiver");
+    }
+
+    [Fact]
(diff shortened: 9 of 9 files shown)


DavidObando:
## Evidence

### Anchors

- Original branch point requested and confirmed: `6b809619779b4403062a6a9540663202f504fb73`.
- Rebased after #3250 and again onto current `main`: `acc10fe47056db3d2c707bd3e00a7aeb25a689df`.
- Head: `b40320429da760ef2981b1cc8172ac4d306ec5f7`.

### Sibling-method matrix

Every cell used `gsc /out:` followed by `dotnet exec`; assembly presence was checked separately.

| receiver | origin | call | before | after |
|---|---|---|---|---|
| `ref struct` | inherited | `ToString()` | compile 0, emitted, run 134 `InvalidProgramException` | compile 1, `GS0219`, no assembly |
| `ref struct` | inherited | `GetHashCode()` | compile 0, emitted, run 134 | compile 1, `GS0219`, no assembly |
| `ref struct` | inherited | `Equals(nil)` | compile 0, emitted, run 134 | compile 1, `GS0219`, no assembly |
| `ref struct` | inherited | `GetType()` | compile 0, emitted, run 134 | compile 1, `GS0219`, no assembly |
| `ref struct` | declared | `ToString()` | 0/0, `declared-11` | unchanged |
| `ref struct` | declared | `GetHashCode()` | 0/0, `22` | unchanged |
| `ref struct` | declared | `Equals(other)` | 0/0, `True` | unchanged |
| `ref struct` | declared | `GetType()` | 0/0, `declared-44` | unchanged |
| plain `struct` | inherited | `ToString()` | 0/0, `Default.Bare` | unchanged |
| plain `struct` | inherited | `GetHashCode()` | 0/0, integer | unchanged |
| plain `struct` | inherited | `Equals(nil)` | 0/0, `False` | unchanged |
| plain `struct` | inherited | `GetType()` | 0/0, `Default.Bare` | unchanged |
| plain `struct` | declared | `ToString()` | 0/0, `declared-11` | unchanged |
| plain `struct` | declared | `GetHashCode()` | 0/0, `22` | unchanged |
| plain `struct` | declared | `Equals(other)` | 0/0, `True` | unchanged |
| plain `struct` | declared | `GetType()` | 0/0, `declared-44` | unchanged |

Additional measured siblings:

- Constructed generic `ref struct Bare[int32]` inherited call: 0/134 before; `GS0219`/no assembly after.
- Imported `ReadOnlySpan[int32].GetType()`: 0/134 before; `GS0219`/no assembly after.
- Imported declared `ReadOnlySpan[int32].ToString()`: remains 0/0.
- A source ref struct can implement an interface. Direct `b.Mark()` remains 0/0 (`marker-55`); conversion `var m Marker = b` changed from 0/134 to `GS0219`/no assembly.
- Inferred ref-struct argument through free, instance, and true extension generic calls changed from compile 0/run 134 `Security.VerificationException` to `GS0219`/no assembly.
- Existing implicit object conversion remains `GS0219`.

### Guard placement and reuse

The actual boxing decision was the `receiverNeedsBox` branch in `MethodBodyEmitter.Expressions.cs`. I extracted its existing rule once as `ReflectionMetadataEmitter.ImportedInstanceReceiverRequiresBoxing` and made the emitter and binder use that same predicate.

Caller grep:

- `ImportedInstanceReceiverRequiresBoxing`: exactly three references repo-wide — declaration, emitter boxing decision, binder guard.
- `TryReportByRefLikeInheritedCall`: declaration plus the two source-call gateways: user-struct inherited CLR lookup (`ExpressionBinder.Calls.Arguments.cs`) and imported CLR receiver lookup (`ExpressionBinder.Calls.Invocation.cs`).
- `TryReportByRefLikeTypeArgument`: declaration plus the three actual generic gateways: free, extension, and user-instance calls.

There is no second boxing-legality spelling and no new diagnostic. All rejection paths reuse existing `Diagnostics.ReportByRefLikeEscape` / `GS0219`. The emitted-call node itself has many synthesized callers (20 binding matches and 33 lowering matches), so the diagnostic belongs at the two user-call binding gateways; the emitter still owns one canonical boxing decision.

### Diagnostic and no-output proof

Example:

`GS0219: By-ref-like type 'Bare' is a ref struct and cannot invoke inherited method 'ToString' because dispatch requires boxing the receiver.`

Each negative pin asserts:

- driver rc is nonzero;
- output contains `GS0219`;
- the requested output assembly does not exist;
- direct `Compilation.Emit` returns unsuccessful with zero PE bytes;
- diagnostic ID and source slice are exact (`ToString()`, `GetHashCode()`, `Equals(nil)`, `GetType()`, `Use`, or `b`).

The message suggests no cast or impossible workaround. The measured valid remedy is to declare the type-specific method: declared methods remain direct and run. `GetType` has no universal boxing-free equivalent, so the diagnostic only states why the inherited call is invalid.

### #3250 collision matrix

| row | before #3240 fix | after |
|---|---|---|
| A receiver-clause declared, `${b}` | 0/0 `Receiver#42` | unchanged |
| B receiver-clause declared, `${b.ToString()}` | 0/0 `Receiver#42` | unchanged |
| C in-body declared, `${b}` | 0/0 `InBody#42` | unchanged |
| D in-body declared, `${b.ToString()}` | 0/0 `InBody#42` | unchanged |
| F1 inherited, `${b}` | compile 1 `GS0519`, no assembly | unchanged |
| F2 inherited, `${b.ToString()}` | compile 0, emitted, run 134 | compile 1 `GS0219`, no assembly |
| G plain struct | 0/0 `Default.Plain` | unchanged |

Only F2 changes, as intended. No #3250 file was edited.

### RED proofs

All final-base mutants rebuilt cleanly, moved `GSharp.Core.dll`, matched a real test filter, and were restored to hash `c1530538e6606a57a87a09bac0737b7f0505b1d16665b23a3a08fb74d1cf16a9` with source diff 0 and 13/13 green.

| removed behavior | mutant Core SHA-256 | result |
|---|---|---|
| user-struct inherited-call guard | `aaebced92f7e341510c482b2ae273010fe845c457f14b0bd65b3cec185b2469a` | 5 failed / 8 passed; each compiled, emitted, then ran rc 134 |
| imported-ref-struct inherited-call guard | `d06d2cb3bc1849a177cc5588a9850a7c2595760c3fcd902f4c8a6eeb69de1736` | 1 failed / 12 passed; compile 0/run 134 |
| source-interface conversion guard | `067279813529fb12019a92e6a0e10baef3aa665ebb6ecf891b495a3dbf7085d8` | 1 failed / 12 passed; compile 0/run 134 |
| shared generic-argument guard | `43d7cbc44f563b6bc4073575afe511df73a1d36ba47ab0cee39163a6ce689cea` | 3 failed / 10 passed; free/instance/extension each compile 0/run 134 `Security.VerificationException` |

The three generic gateway calls were also mutated individually before the final rebase; each corresponding pin failed. One initial “extension” specimen used an owned `Host` receiver and routed through the instance path, so it was a no-result. Replacing it with a `string` receiver positively reached the true extension path; removing that guard then failed with compile 0/run 134. No survival is claimed.

The emitter line replacement is deliberately behavior-equivalent to its old inline formula; its value is eliminating a second spelling. Source grep, not a fake behavioral mutant, proves that deduplication.

### Final validation

- Targeted: 13 passed, 0 failed; no `No test matches`.
- Release solution build with CI determinism and warnings-as-errors: rc 0, **0 warnings, 0 errors**.
- Full suite: **15,393 passed, 4 skipped, 0 failed across 9/9 projects**:
  - Core 7,178 passed / 1 skipped
  - Compiler 4,280 passed
  - Interpreter 1,357 passed / 3 skipped
  - LanguageServer 468 passed
  - SDK 59 passed
  - Extensions 104 passed
  - Cs2Gs 1,907 passed
  - GeneratorHost 31 passed
  - InternalAnalyzers 9 passed

DavidObando:
CI complete on `b40320429da760ef2981b1cc8172ac4d306ec5f7`: 30 successful, 2 skipped, 0 pending, 0 failed. GitHub reports `MERGEABLE` / `CLEAN`.

DavidObando:
## Round 1 — **no BLOCKER.** One SHOULD-FIX, one NIT.

Against **`main` = `65578d41b`** (fetched myself, re-fetched at post time — unmoved), head **`b40320429d`**, `mergeable=MERGEABLE state=CLEAN`.

Own worktrees, Release with `-p:ContinuousIntegrationBuild=true -warnaserror`: **0 warnings, 0 errors**. Oracle throughout is the running program — `gsc /out:` → `dotnet exec` → assert rc, stdout, **and assembly presence**. `ilverify` not used as evidence; this issue is the canonical proof it is worthless here.

**Rebuilt on the merge with current `main`** as asked, not just the branch: merge is clean (`git merge-tree --write-tree 65578d41b b40320429d` → **1 line, tree oid only, no stages**), builds 0W/0E, 13/13 pins pass, and every behaviour below reproduces on the merge.

---

### 1. Guard placement — is it the routing point, or another spelling?

**For the boxing rule: it is the routing point, and this is the strongest part of the PR.**

`ImportedInstanceReceiverRequiresBoxing` has **exactly 3 references repo-wide** — declaration (`ReflectionMetadataEmitter.cs:5114`), the emitter's actual boxing decision (`MethodBodyEmitter.Expressions.cs:291`), and the binder guard (`ExpressionBinder.Calls.Invocation.cs:38`). The binder does not re-derive the rule; it calls the emitter's.

I confirmed the sharing is real rather than cosmetic by trying to revert the two emitter files alone: **the build fails rc=1** because the binder references the predicate. They cannot be separated. (That revert is therefore a **no-result**, not a behavioural datum, and I do not score it — the old and new formulas are equivalent by reading: `declaringType != null && !declaringType.IsValueType` ≡ `method?.DeclaringType is { IsValueType: false }`.)

Then I **drove** six sibling call paths rather than reading the table. All correct:

| shape | result |
|---|---|
| extension with `object` receiver → boxes (`b.Ext()`) | **GS0219**, no assembly |
| inherited call on returned temp (`Make().ToString()`) | **GS0219**, no assembly |
| inherited call on nested ref-struct field (`o.inner.ToString()`) | **GS0219**, no assembly |
| imported ref struct inherited (`span.GetType()`) | **GS0219**, no assembly |
| imported ref struct **declared** (`span.ToString()`) | **0/0 → `System.ReadOnlySpan<Int32>[2]`** |
| declared method calling own declared (`this.ToString()`) | **0/0 → `d-42`** |

**No sibling call path reaches boxing unguarded, and no legal declared call is over-rejected.** Method groups are a non-issue: `var f = b.ToString` is rejected as `GS0131 'f' is not a function` — G# has no method-group value, pre-existing.

### 2. `GS0219` — one predicate or two?

The **boxing** predicate is genuinely **one** (above). The **diagnostic reporter** `ReportByRefLikeEscape` is reused; no new ID. That part is the ponytail answer.

**But the generic-type-argument rule was re-derived, not reused.** See SHOULD-FIX.

### 3. Sibling matrix — driven, all 16 cells plus extras

4 inherited (`ToString`/`GetHashCode`/`Equals`/`GetType`) × declared/inherited × ref-struct/plain-struct reproduce the author's table exactly. Extras driven: constructed generic `Bare[int32]`, imported `ReadOnlySpan`, source ref struct implementing an interface (direct `b.Mark()` still `0/0`, conversion `var m Marker = b` → GS0219), `defer`, lambda body, class field, nested generic, **and explicit type arguments `Use[Bare](b)`** — which was **0/134 `VerificationException` before** and is now GS0219. Explicit args are covered, not just inferred.

### 4. No assembly written — verified, not asserted

Every rejection above was checked for the output file on disk: **`asm=NO` in all cases**. The pins assert this three ways — driver rc≠0, `File.Exists(assemblyPath)` false, and direct `Compilation.Emit` with `peStream.Length == 0` and a **single** error diagnostic. Stronger than required.

### 5. Diagnostic quality + remedy check

Spans verified by slicing the source at the reported columns, not by trusting the id:

| probe | id | span covers |
|---|---|---|
| `b.ToString()` | GS0219 | `ToString()` |
| `Use[Bare](b)` | GS0219 | `[Bare]` |
| `b.Ext()` | GS0219 | `Ext()` |
| `span.GetType()` | GS0219 | `GetType()` |

Messages state the reason and **invent no remedy** — so there is no advice to follow and no repeat of the sibling PR that suggested a cast which then ICEd. The real remedy (declare the method) I drove myself: `0/0 → declared-11`. `GetType` has no boxing-free equivalent and the message correctly does not pretend otherwise.

### 6. #3250 interaction — all seven rows driven, before and after

#3250 is **MERGED** (`f5642aff1`), so I measured against merged reality, not the PR.

| row | merge-base `acc10fe47` | head `b40320429d` |
|---|---|---|
| A receiver-clause declared, `${b}` | 0/0 `Receiver#42` | unchanged |
| B receiver-clause declared, `${b.ToString()}` | 0/0 `Receiver#42` | unchanged |
| C in-body declared, `${b}` | 0/0 `InBody#42` | unchanged |
| D in-body declared, `${b.ToString()}` | 0/0 `InBody#42` | unchanged |
| **F1 inherited, `${b}`** | **rc1 GS0519, no asm** | **unchanged — stays red by design** |
| **F2 inherited, `${b.ToString()}`** | **0/134 `InvalidProgramException`** | **rc1 GS0219, no asm** |
| G plain struct, both spellings | 0/0 `Default.Plain` ×2 | unchanged |

**Only F2 moved.** The claim is exactly right, and F1 did not get routed into F2's hole.

### 7. RED proofs — per hunk, correct failure signature

Reverted each guard alone onto the head, rebuilt (`BUILD_RC=0` each), ran the 13 pins:

| reverted | result | signature |
|---|---|---|
| inherited-call guard (both `ExpressionBinder.Calls.*`) | **6 failed / 7 passed** | 6 × `got compile rc=0; run rc=134`, `InvalidProgramException` |
| generic type-argument guard (3 `OverloadResolver*`) | **3 failed / 10 passed** | 3 × `compile rc=0; run rc=134`, `VerificationException` |
| conversion guard (`ConversionClassifier.cs`) | **1 failed / 12 passed** | 1 × `compile rc=0; run rc=134`, `InvalidProgramException` |

Every failure is **rc0-compile-then-rc134-run**, never a compile error inside the test. The author's 5+1 split reconciles with my combined 6. Restored: `git diff HEAD` = 0, porcelain empty, **13/13 green**. No survivals claimed anywhere, so no phantom-survival risk.

### 8. Suite — 9 of 9, independently reproduced

**15,393 passed / 0 failed / 4 skipped**, all RC=0 — matching the author's figure exactly.
Compiler 4280 · Core 7178 (1 skip) · Cs2Gs 1907 · Interpreter 1357 (3 skip) · LanguageServer 468 · Extensions 104 · Sdk 59 · GeneratorHost 31 · InternalAnalyzers 9. (Denominator 9; `test-assets/Issue3119.CrossConstants` is an asset, not a suite.)

---

## SHOULD-FIX — the by-ref-like **type-argument** rule is now spelled six ways, and three positions still emit invalid programs

`OverloadResolver.cs:66` `TryReportByRefLikeTypeArgument` is a **new implementation** of a rule that already exists **five times** in `Binder.cs` — `:3321`, `:3458`, `:3828`, `:4131`, `:4360` — each a hand-rolled loop ending in `ReportByRefLikeEscape(..., "be used as a generic type argument")` with the **identical message string**. This PR makes it six.

In fairness: the five existing sites are all syntax-driven (`BindTypeClause` over a `TypeArgumentList`), and this PR needed a check over a **substitution dictionary** because inferred type arguments have no syntax node. So it could not simply call one of them. That is a real reason, and I am not scoring it as gratuitous.

What I am scoring is that the rule remains incomplete, and I found it by driving, not reading. Three positions **compile rc=0, emit an assembly, and die rc=134 with no diagnostic** — the exact defect class of #3240:

| probe | result on head | CLR message |
|---|---|---|
| `Holder[Bare]{v: b}` — ref struct as arg to a **user generic struct** | 0 → **134** | `TypeLoadException: A ByRef or ByRef-like type cannot be used as the type for an instance field in a non-ByRef-like type` |
| `[]Bare{b}` — **array** of ref struct | 0 → **134** | `TypeLoadException: Could not create array type 'Default.Bare[]' ... element type is ByRef-like` |
| `List[Bare]()` — **imported** generic type | 0 → **134** | `TypeLoadException: GenericArguments[0] ... violates the constraint of type parameter 'T'` |

**These are pre-existing, not regressions** — I measured all three at merge-base `acc10fe47` and they are byte-for-byte the same behaviour. They are therefore **not a blocker for this PR**, and #3240's scope is inherited calls, not type positions.

But the PR's own message — *"cannot be used as a generic type argument"* — applies verbatim to `Holder[Bare]` and `List[Bare]`, so the gap sits inside the rule this PR just extended. Recommend either routing the new check and the five `Binder.cs` copies through one shared helper, or (cheaper, and the ponytail option) filing these three as a follow-up issue now so the seventh spelling is not written later by someone who greps `"be used as a generic type argument"` and believes the rule is enforced. Happy to hand over the fixtures.

## NIT — file overlap with in-flight #3284

#3284 edits `OverloadResolver.CallBinding.cs` and `OverloadResolver.Invocations.cs` (both touched here) and `Conversion.cs`, which owns `Conversion.IsReferenceLikeTarget` — the pre-existing helper (#2566) this PR newly depends on at `ConversionClassifier.cs:663`. Textually clean today; whichever lands second should re-verify the ref-struct conversion row rather than trusting a clean merge.

---

Verdict: **no BLOCKER**. The defect is genuinely fixed, the guard sits on the shared predicate the emitter actually uses, declared calls and imported declared calls are not over-rejected, no assembly is written, spans are exact, all seven #3250 rows hold with only F2 moving, and RED proofs fail for precisely the right reason. Both scratch worktrees restored to porcelain-0; the author's `i3240` worktree and the shared checkout were never touched.

DavidObando:
## Addendum — driven 9-shape enumeration. **BLOCKER: 4 of 9 uncovered.**

Second-agent enumeration checked by driving, not reading. Compile → `dotnet exec` → assert rc + stdout. Nothing taken on trust.

**SHAs.** head `b40320429da760ef2981b1cc8172ac4d306ec5f7` · merge-base `acc10fe47` · `main` re-fetched at post time = `6764eeb6b00a52f9ba27b75b71d66020d559f6e0` (moved from `65578d41b` since my first pass). Both trees built `-warnaserror -p:ContinuousIntegrationBuild=true`, RC=0/0W. head `GSharp.Core.dll` md5 `df26c8f4d7dedcf55531e2a6d1f86c94`, merge-base `937431c4f76d1a0178fdbe3cd3fc2b3a`. Issue3240 pins **13/13**.

### The 9 shapes, driven both sides

| # | shape | merge-base `acc10fe47` | head `b40320429d` | |
|---|---|---|---|---|
| 1 | `b.ToString()` | rc0 → **134** `InvalidProgramException` | rc1 **GS0219** asm=NO | FIXED |
| 2 | `b.GetType()` | rc0 → **134** IPE | rc1 GS0219 asm=NO | FIXED |
| 3 | `b.GetHashCode()` | rc0 → **134** IPE | rc1 GS0219 asm=NO | FIXED |
| 4 | `b.Equals(nil)` | rc0 → **134** IPE | rc1 GS0219 asm=NO | FIXED |
| 5 | `span.GetType()` (`ReadOnlySpan`) | rc0 → **134** IPE | rc1 GS0219 asm=NO | FIXED — already pinned |
| 6 | mg inherited `let f () -> string = b.ToString` | rc0 → **134** IPE | **rc0 → 134 IPE** | **UNCOVERED** |
| 7 | mg **declared** `let f () -> int32 = b.Get` | rc0 → **134** IPE | **rc0 → 134 IPE** | **UNCOVERED** |
| 8 | mg user ext `let f () -> int32 = span.Ext` | rc0 → **134** IPE | **rc0 → 134 IPE** | **UNCOVERED** |
| 9 | mg imported ext `let f () -> bool = span.IsWhiteSpace` | rc0 → **134** IPE | **rc0 → 134 IPE** | **UNCOVERED** |

Shape 5 is **already fixed and already pinned** (`InheritedObjectMethod_OnImportedRefStruct_ReportsBoxing`) — the enumeration lists it as outstanding; it is not.

Over-reach controls, all still legal on head: declared call `b.Get()` → `3`; declared `b.ToString()` → `declared`; **direct** user ext `span.Ext()` → `3`; **direct** imported ext `span.IsWhiteSpace()` → `True`; plain struct generic → `Default.Plain`. Confirms "extension *calls* legal, extension *method groups* not".

### BLOCKER — `MethodBodyEmitter.Closures.cs:508`, `:985`, `:1019`

Not inference. The IL is literal:

```
// shape 7, DECLARED method
IL_0014: box        Default.Bare
IL_0019: ldftn      instance int32 Default.Bare::Get()
IL_001f: newobj     instance void System.Func`1<int32>::.ctor(object, native int)
```
shape 6 → `box Default.Bare`; shape 8 → `box valuetype ReadOnlySpan``1<int32>`; shape 9 → `box valuetype ReadOnlySpan``1<char>`.

All three sites are the same unguarded shape:

```csharp
if (ReflectionMetadataEmitter.IsValueTypeSymbol(methodGroup.Receiver.Type))
{
    this.il.OpCode(ILOpCode.Box);
```

`git diff acc10fe47..HEAD -- src/Core/CodeAnalysis/Emit/` touches **only** `MethodBodyEmitter.Expressions.cs` and `ReflectionMetadataEmitter.cs`. **`Closures.cs` is untouched** — the shared predicate is not on this path.

**Shape 7 is the load-bearing one: `Get` is *declared* on `Bare`, and it still boxes.** So `ImportedInstanceReceiverRequiresBoxing` (value-type receiver **+ declaring type not a value type**) cannot catch it even if wired in. Delegate creation needs a *second* rule — **any** delegate over a by-ref-like receiver is illegal, declared or inherited — while direct calls stay legal. The table above pins that boundary in both directions.

`:985`'s own comment ("if that gate ever loosens … would produce unverifiable IL") shows the gate was assumed elsewhere. It is not there for `ref struct`.

**Honest scoping:** all four are **byte-identical on the merge-base** — pre-existing, not regressions of this PR. Same defect class (illegal box of a by-ref-like value, rc0 compile → rc134), different chokepoint (delegate `Target` slot vs call receiver), different predicate. Reasonable to fold in, equally reasonable to split; that is the owner's call, not mine, and not the author's to file.

### Refuted: "shared-generic path still emits `box !!T` and crashes"

Does **not** reproduce on head for generic *methods*:

| generic shape | merge-base | head |
|---|---|---|
| inferred `Show(Bare{…})`, body `v.ToString()` | rc0 → 134 `VerificationException` | **rc1 GS0219 asm=NO** |
| inferred `Show(Bare{…})`, body `Console.WriteLine(v)` (literal `box !!T`) | rc0 → 134 | **rc1 GS0219 asm=NO** |
| explicit `Show[Bare](…)` | rc0 → 134 | **rc1 GS0219 asm=NO** |
| plain struct `Show(Plain{…})` | 0/0 `Default.Plain` | unchanged |

This PR **fixes** the generic-method box. What still escapes is generic **types** — `Holder[Bare]{v: b}`, `Printer[Bare]().Show(b)`, `[]Bare{b}`, `List[Bare]()` — all rc0 → 134 **`TypeLoadException`**, identical on both sides. Different mechanism (the constructed type is itself illegal, no `box` involved), pre-existing, and already my earlier SHOULD-FIX. The enumeration's claim is correct only if "shared-generic" means generic types, and even then the exception class is not what was reported.

### Both "ruled out" items confirmed, cheaply

- `ExpressionBinder.Literals.cs:466` — still declared-only `TryGetMethod`, untouched by this PR. Deliberate; keeps #3250 F1 at GS0519. **Not a miss.**
- `GetMethodSafe` — untouched by this PR (23 refs, none in the diff). Irrelevant here; left alone for #3281.

### Net

5 of 9 fixed and correctly pinned, 0 over-reach, generic-method boxing fixed as a bonus. The four method-group shapes are a real, driven, still-live `rc0 → rc134` at `Closures.cs:508/:985/:1019`, needing a receiver-shape rule rather than the declared-vs-inherited one. Fixtures available on request.

DavidObando:
Addressed the addendum BLOCKER in 4d970c0fb. Method-group conversion now rejects any by-ref-like receiver before delegate construction, covering inherited and declared members plus user/imported extension groups; direct calls remain legal. Added all four driven shapes with GS0219 span/no-assembly assertions. Rebased onto current main.

Local validation: Release CI build 0 warnings/errors; Issue3240 17/17; method-group regressions 50/50. GitHub checks are running.

Issue #3280, opened 2026-08-05
[bug] No boxing conversion from a value type to System.ValueType, and the four call paths disagree (GS0154 vs GS0155)


DavidObando:
## Summary

`Conversion.Classify` has no ordinary boxing conversion from a value type to **`System.ValueType`**, and the four call paths disagree about how to report the resulting failure.

Measured across nine probe shapes × four call shapes. Every other shape — `int32 -> object`, `int32 -> IComparable`, unconstrained `T -> object`, `T: struct -> object`, `int32 -> object?`, `int32? -> object`, `int32? -> int32?`, `T?: struct -> object` — compiles rc 0 with zero diagnostics and runs rc 0 in **all four** call shapes. Only the `ValueType` row fails, and it fails **inconsistently**:

| target | top-level (free) | instance | extension | `shared` |
|---|---|---|---|---|
| `int32 -> object` | rc 0, runs | rc 0, runs | rc 0, runs | rc 0, runs |
| **`int32 -> ValueType`** | **rc 1 `GS0154`** | **rc 1 `GS0155`** | **rc 1 `GS0155`** | **rc 1 `GS0155`** |

Boxing an `int32` to `System.ValueType` is legal on the CLR — `box` produces a reference that is assignable to `System.ValueType`, exactly as it is to `System.Object`. So the correct behaviour is rc 0 and a running program, matching the `object` row directly above it.

## Two defects, one root

1. **Missing conversion.** `Conversion.Classify` handles boxing to `object` and to interfaces, but has no case for value type → `System.ValueType`. `System.Enum` is worth checking at the same time — it is the same shape (a base class of value types that is itself a reference type), and if `ValueType` is missing, `Enum` very likely is too.
2. **Diagnostic disagreement.** The free path reports `GS0154` while the instance, extension and `shared` paths report `GS0155` for the same rejected conversion. Whatever the outcome of (1), two spellings of one rejection is a bug on its own — and if (1) is fixed, this asymmetry should disappear rather than be papered over.

## Provenance

Found while reconciling two competing fixes for #3271. The mirror repair landing in #3272 (`OverloadResolver.CallBinding.cs:1707`) replaces the former `InvalidProgramException` (rc 134) with a diagnostic on this row, which **makes this gap visible rather than causing it** — before that change, the free `ValueType` row crashed the runtime instead of reporting anything. This is pre-existing and independent.

## Notes for the fix

[...]

Issue #1218, opened 2026-06-26
enum.HasFlag(flag) / inherited System.Enum instance members not callable on enum values — GS0159


DavidObando:
### Summary
Inherited `System.Enum` instance members — notably `HasFlag(Enum)` — are not callable on a G# enum value: `enum.HasFlag(flag)` reports `GS0159: Cannot find function HasFlag`. G# enums are emitted as CLR enums (deriving from `System.Enum`), so the inherited instance members should be callable, mirroring C#.

### Repro
```gsharp
package p
enum Flags { A, B, C }
func F(x Flags, y Flags) bool { return x.HasFlag(y) }   // GS0159: Cannot find function HasFlag
```

### Expected
`x.HasFlag(y)` binds to `System.Enum.HasFlag(System.Enum)` and emits a virtual call, returning the runtime flags test — the faithful mapping of C# `x.HasFlag(y)`. (Other inherited `System.Enum`/`System.Object` members such as `ToString()` should likewise resolve.)

### Impact (Oahu.Decrypt)
`Ac4Extensions.cs` tests channel membership via `channels.HasFlag(group)` on the `[Flags]` enum `ChannelGroups`. Drives GS0159 + cascade on the enclosing channel-count computation.

### Related translator note (separate, cs2gs-side — NOT part of this compiler issue)
cs2gs currently drops explicit enum member values (e.g. C# `Left_Right = 1 << 0`) when emitting the G# enum, which also breaks `[Flags]` bit semantics. That is a translator defect I will address in tools/cs2gs separately; this issue is specifically the compiler gap that `HasFlag` is unresolvable on an enum value.

### Area
Binder — member lookup on an enum-typed receiver should surface inherited `System.Enum` (and `System.Object`) instance members; emit the appropriate (virtual) call. Analogous to #1136 (System.Object members on user types).

### Environment
gsc at main (903f8b6b). `bug`.

PR #3284, opened 2026-08-05
Fix boxing conversions to ValueType and Enum

DavidObando:
Refs #3280

## Summary
- classify concrete CLR value types and `struct`-constrained type parameters as implicitly boxable to `System.ValueType`
- classify concrete enum values as implicitly boxable to `System.Enum`
- execute non-generic and explicitly closed generic object, ValueType, Enum, and concrete-enum calls across free, instance, extension, and shared methods
- replace #3272's temporary ValueType diagnostic pins with runtime boxing assertions
- pin reference, nullable, reified value-type, and mutable-struct controls against over-boxing

## Scope
This addresses the missing valid boxing conversions described in #3280. The separate GS0154/GS0155 disagreement for rejected arguments remains reproducible and is tracked by #3312, and #3280 remains open./

## Coordination
#3272 is merged. This branch is rebased onto current `main`; its duplicate generic-call mirror commit dropped as already upstream. The resulting PR diff is limited to the conversion classifier and boxing regression tests. `merge-tree` is clean against current `main` and nearby ref-struct boxing PR #3283.

## Validation
- issue #3271/#3280 boxing tests: 19 passed, 0 failed
- issue #3267 generic-call regression tests: 18 passed, 0 failed
- issue #1218 enum regression tests: 4 passed, 0 failed

### DIFF
--- a/src/Core/CodeAnalysis/Binding/Conversion.cs
+++ b/src/Core/CodeAnalysis/Binding/Conversion.cs
@@ -850,17 +850,13 @@ public sealed class Conversion
             return Conversion.Implicit;
         }
 
-        // Issue #1218: an enum value boxes implicitly to its CLR reference base
-        // types — System.Object, System.ValueType, and System.Enum. An
-        // EnumSymbol carries no ClrType during binding (the enum is still being
-        // compiled), so the general object-boxing rule above cannot fire. This
-        // makes inherited Enum/ValueType/Object members callable on enum values
-        // (e.g. passing an enum to System.Enum.HasFlag(System.Enum)); the
-        // emitter lowers the conversion to a single `box <Enum>`.
-        if (from is EnumSymbol && to?.ClrType is System.Type enumBoxTarget
-            && (enumBoxTarget.IsSameAs(typeof(object))
-                || enumBoxTarget.IsSameAs(typeof(System.ValueType))
-                || enumBoxTarget.IsSameAs(typeof(System.Enum))))
+        // Issues #1218/#3280: value types box implicitly to System.ValueType,
+        // while enum values additionally box to System.Enum. A same-compilation
+        // EnumSymbol has no ClrType yet, so retain its object-boxing path too.
+        if (to?.ClrType is System.Type valueTypeBaseTarget
+            && ((IsValueTypeLikeFrom(from) && valueTypeBaseTarget.IsSameAs(typeof(System.ValueType)))
+                || (IsEnumLikeType(from) && valueTypeBaseTarget.IsSameAs(typeof(System.Enum)))
+                || (from is EnumSymbol && valueTypeBaseTarget.IsSameAs(typeof(object)))))
         {
             return Conversion.Implicit;
         }
@@ -2719,6 +2715,11 @@ public sealed class Conversion
 
     private static bool IsValueTypeLikeFrom(TypeSymbol type)
     {
+        if (type is TypeParameterSymbol { HasValueTypeConstraint: true })
+        {
+            return true;
+        }
+
         if (type is NullableTypeSymbol nullable
             && NullableLifting.IsAnyValueTypeNullable(nullable))
         {
--- a/test/Core.Tests/CodeAnalysis/Emit/Issue3271ReifiedGenericArgumentBoxingTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Emit/Issue3271ReifiedGenericArgumentBoxingTests.cs
@@ -76,17 +76,21 @@ public class Issue3271ReifiedGenericArgumentBoxingTests
             "Issue3271TopReferences.IMark[]",
             "Issue3271TopReferences.Mark");
 
-        // Issue #3280: value-type → System.ValueType is not classified yet.
-        AssertValueTypeTargetDiagnostic(
+        AssertRuns(
             """
             package Issue3271TopValueType
             import System
 
-            func Show[T](value T) { }
+            func Show[T](value T) {
+                Console.WriteLine([]T{value}.GetType())
+                Console.WriteLine(value.GetType())
+                Console.WriteLine(value)
+            }
 
             Show[ValueType](23)
             """,
-            "GS0154",
+            "System.ValueType[]",
+            "System.Int32",
             "23");
     }
 
@@ -119,20 +123,24 @@ public class Issue3271ReifiedGenericArgumentBoxingTests
             "System.IComparable[]",
             "System.Int32");
 
-        // Issue #3280: value-type → System.ValueType is not classified yet.
-        AssertValueTypeTargetDiagnostic(
+        AssertRuns(
             """
             package Issue3271InstanceValueType
             import System
 
             class Runner {
                 init() {}
-                func Show[T](value T) { }
+                func Show[T](value T) {
+                    Console.WriteLine([]T{value}.GetType())
+                    Console.WriteLine(value.GetType())
+                    Console.WriteLine(value)
+                }
             }
 
             Runner().Show[ValueType](33)
             """,
-            "GS0155",
+            "System.ValueType[]",
+            "System.Int32",
             "33");
     }
 
@@ -160,17 +168,21 @@ public class Issue3271ReifiedGenericArgumentBoxingTests
             "System.IComparable[]",
             "System.Int32");
 
-        // Issue #3280: value-type → System.ValueType is not classified yet.
-        AssertValueTypeTargetDiagnostic(
+        AssertRuns(
             """
             package Issue3271ExtensionValueType
             import System
 
-            func (self string) Show[T](value T) { }
+            func (self string) Show[T](value T) {
+                Console.WriteLine([]T{value}.GetType())
+                Console.WriteLine(value.GetType())
+                Console.WriteLine(value)
+            }
 
             "extension".Show[ValueType](43)
             """,
-            "GS0155",
+            "System.ValueType[]",
+            "System.Int32",
             "43");
     }
 
@@ -370,14 +382,4 @@ public class Issue3271ReifiedGenericArgumentBoxingTests
             result.Output);
     }
 
-    private static void AssertValueTypeTargetDiagnostic(string source, string expectedId, string expectedText)
-    {
-        var result = EmittedOracle.Evaluate(source);
-        var diagnostic = Assert.Single(result.Diagnostics);
-
-        Assert.Equal(expectedId, diagnostic.Id);
-        Assert.Equal(expectedText, diagnostic.Location.Text.ToString(diagnostic.Location.Span));
-        Assert.Equal(1, result.ExitCode);
-        Assert.Equal(string.Empty, result.Output);
-    }
 }
--- a/test/Core.Tests/CodeAnalysis/Emit/Issue3280ValueTypeBaseBoxingTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Emit/Issue3280ValueTypeBaseBoxingTests.cs
@@ -1,0 +1,276 @@
+// <copyright file="Issue3280ValueTypeBaseBoxingTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using GSharp.Core.CodeAnalysis.Binding;
+using GSharp.Core.CodeAnalysis.Symbols;
+using GSharp.Tests;
+using Xunit;
+
+namespace GSharp.Core.Tests.CodeAnalysis.Emit;
+
+/// <summary>
+/// Issue #3280: CLR value types box implicitly to <see cref="ValueType"/>,
+/// and enum values box implicitly to <see cref="Enum"/>, in every call shape.
+/// </summary>
+public sealed class Issue3280ValueTypeBaseBoxingTests
+{
+    [Theory]
+    [InlineData("top-level")]
+    [InlineData("instance")]
+    [InlineData("extension")]
+    [InlineData("shared")]
+    public void ValueTypeReferenceBases_BoxAndRun_InEveryCallShape(string shape)
+    {
+        AssertRuns(
+            MatrixSource(shape),
+            "System.ValueType[]",
+            "System.Int32",
+            "5",
+            "System.Enum[]",
+            "System.DayOfWeek",
+            "Sunday",
+            "System.Object[]",
+            "System.Int32",
+            "11",
+            "System.ValueType[]",
+            "System.Int32",
+            "22",
+            "System.Enum[]",
+            "System.DayOfWeek",
+            "Monday",
+            "System.DayOfWeek[]",
+            "System.DayOfWeek",
+            "Tuesday");
+    }
+
+    [Fact]
+    public void AlreadyBoxedValueTypeArgument_PreservesTargetTypeAndValue()
+    {
+        AssertRuns(
+            """
+            package Issue3280AlreadyBoxed
+            import System
+
+            func Show[T](value T) {
+                Console.WriteLine([]T{value}.GetType())
+                Console.WriteLine(value.GetType())
+                Console.WriteLine(value)
+            }
+
+            let boxed ValueType = 31
+            Show[ValueType](boxed)
+            """,
+            "System.ValueType[]",
+            "System.Int32",
+            "31");
+    }
+
+    [Fact]
+    public void ReferenceTypeArgument_RemainsReferenceTyped()
+    {
+        Assert.False(
+            Conversion.Classify(
+                TypeSymbol.String,
+                TypeSymbol.FromClrType(typeof(ValueType))).Exists);
+        Assert.False(
+            Conversion.Classify(
+                TypeSymbol.Int32,
+                TypeSymbol.FromClrType(typeof(Enum))).Exists);
+
+        AssertRuns(
+            """
+            package Issue3280ReferenceControl
+            import System
+
+            func Show[T](value T) {
+                Console.WriteLine([]T{value}.GetType())
+                Console.WriteLine(value.GetType())
+                Console.WriteLine(value)
+            }
+
+            Show[object]("right")
+            """,
+            "System.Object[]",
+            "System.String",
+            "right");
+    }
+
+    [Fact]
+    public void ReifiedValueTypeParameter_RemainsUnboxed()
+    {
+        AssertRuns(
+            """
+            package Issue3280ReifiedValueControl
+            import System
+
+            func Show[T](value T) {
+                Console.WriteLine([]T{value}.GetType())
+                Console.WriteLine(value.GetType())
+                Console.WriteLine(value)
+            }
+
+            func Forward[T](value T) {
+                Show[T](value)
+            }
+
+            Forward[int32](41)
+            """,
+            "System.Int32[]",
+            "System.Int32",
+            "41");
+    }
+
+    [Fact]
+    public void StructConstrainedTypeParameter_BoxesToValueType()
+    {
+        AssertRuns(
+            """
+            package Issue3280StructConstraint
+            import System
+
+            func Take(value ValueType) {
+                Console.WriteLine(value.GetType())
+                Console.WriteLine(value)
+            }
+
+            func Forward[T struct](value T) {
+                Take(value)
+            }
+
+            Forward[int32](42)
+            """,
+            "System.Int32",
+            "42");
+    }
+
+    [Fact]
+    public void NullableValueTypeParameter_RemainsUnboxed()
+    {
+        AssertRuns(
+            """
+            package Issue3280NullableValueControl
+            import System
+
+            func Show[T](value T) {
+                Console.WriteLine([]T{value}.GetType())
+                Console.WriteLine(value.GetType())
+                Console.WriteLine(value)
+            }
+
+            var value int32? = 51
+            Show[int32?](value)
+            """,
+            "System.Nullable`1[System.Int32][]",
+            "System.Int32",
+            "51");
+    }
+
+    [Fact]
+    public void ConstrainedMutableStruct_RemainsMutable()
+    {
+        AssertRuns(
+            """
+            package Issue3280MutableStructControl
+            import System
+
+            interface IMutable {
+                func Bump();
+                func Read() int32;
+            }
+
+            struct Counter : IMutable {
+                var Value int32
+                func Bump() { Value = Value + 1 }
+                func Read() int32 { return Value }
+            }
+
+            func Mutate[T IMutable](value T) T {
+                value.Bump()
+                return value
+            }
+
+            let result = Mutate[Counter](Counter{Value: 60})
+            Console.WriteLine(result.Read())
+            """,
+            "61");
+    }
+
+    private static string MatrixSource(string shape)
+    {
+        const string genericMethodBody = """
+            {
+                Console.WriteLine([]T{value}.GetType())
+                Console.WriteLine(value.GetType())
+                Console.WriteLine(value)
+            }
+            """;
+        const string valueTypeMethodBody = """
+            {
+                Console.WriteLine([]ValueType{value}.GetType())
+                Console.WriteLine(value.GetType())
+                Console.WriteLine(value)
+            }
+            """;
+        const string enumMethodBody = """
+            {
+                Console.WriteLine([]Enum{value}.GetType())
+                Console.WriteLine(value.GetType())
+                Console.WriteLine(value)
+            }
+            """;
+
+        var declaration = shape switch
+        {
+            "top-level" => "func TakeValueType(value ValueType) " + valueTypeMethodBody
+                + " func TakeEnum(value Enum) " + enumMethodBody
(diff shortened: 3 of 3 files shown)


DavidObando:
## Verification evidence

Built from `main` `f5642aff1598630710f29737baf3416273b0dcf4`; branch head `c0e8679530c43fe69680cc225c0dcc42a3b50987`.

### Mechanism

`object` already uses its object/reference widening rules, and interfaces already use the value-type/interface assignability rule. `Conversion.Classify` also already had the enum-to-CLR-base boxing rule, while `MethodBodyEmitter.Conversions` already emits `box` for `ValueType`/`Enum` targets. This change extends that existing classifier rule: every `IsValueTypeLikeFrom` source may box to `System.ValueType`, and every `IsEnumLikeType` source may box to `System.Enum`. No emitter change.

### Execution matrix

Every cell: compile rc 0, assembly written, runtime rc 0, empty stderr.

| target/source | free | instance | extension | shared |
|---|---|---|---|---|
| `int32 -> object` | `Object[] / Int32 / 11` | same | same | same |
| `int32 -> ValueType` | `ValueType[] / Int32 / 22` | same | same | same |
| `DayOfWeek -> Enum` | `Enum[] / DayOfWeek / Monday` | same | same | same |
| `DayOfWeek -> DayOfWeek` | `DayOfWeek[] / DayOfWeek / Tuesday` | same | same | same |

### RED proof

Reverted only `Conversion.cs` classifier hunk, retaining tests and #3272 integration:

- `5 failed / 4 passed / total 9`
- all four matrix tests rejected both new conversions
- free path: parameter mismatch (`GS0154` behavior)
- instance/extension/shared: conversion rejection (`GS0155` behavior)
- already-boxed setup also rejected `int32 -> ValueType`

Like-for-like Core DLL witness moved `a86c9ff78c5b317c285fa3eef94a6a46` -> mutant `524a0a3c2a355972238093b2926eeb22`. Restored product hunk: `9/9` green, hash returned exactly to `a86c9ff78c5b317c285fa3eef94a6a46`, and `git diff HEAD` was empty.

### Over-correction controls

- statically `ValueType` argument: `ValueType[] / Int32 / 31`
- reference argument: `Object[] / String / right`
- reified value-type `T`: `Int32[] / Int32 / 41`
- nullable value-type `T`: `Nullable<Int32>[] / Int32 / 51`
- mutable struct through interface-constrained generic call: mutation remains visible (`61`)
- classifier still rejects `string -> ValueType` and `int32 -> Enum`

### Coordination and suite

Against #3272 head `d259a8df7bddfa0fa4cf84339ea7dd7c544a6a27`:

- merge tree `ec17e38d581dcc87929e647e29192c8ebd10c2f5`
- rc 0
- **0 stage entries**

Release solution build: **0 warnings, 0 errors**.

Project suites: **15,403 passed, 4 skipped, 0 failed**. Aggregate solution invocation initially raced LanguageServer extension assets and reported 4 `Issue3198LooseFileExtensionDiagnosticsTests` failures; isolated LanguageServer rerun passed **468/468**. Other project totals from the completed aggregate run were Core 7205/1 skipped, Compiler 4267, Interpreter 1353/3 skipped, Cs2Gs 1907, Extensions 104, SDK 59, GeneratorHost 31, InternalAnalyzers 9.

DavidObando:
Final pushed head: `eeef74a3ad9a372ad94db579d9cb6b0ca9c64c96` (tree unchanged from evidence above). CI complete: **31 SUCCESS / 3 SKIPPED / 0 failed**, including core, compiler, interpreter, LanguageServer, e2e, and ilverify. Merge state: **CLEAN**. Rechecked against #3272 head `d259a8df7bddfa0fa4cf84339ea7dd7c544a6a27`: merge-tree rc 0, **0 stage entries**.

DavidObando:
## Round-1 review — **NOT MERGEABLE**: 1 BLOCKER, 3 SHOULD-FIX, 2 NIT

The conversion rule itself is correct, minimal, and in the right place. What blocks is that `Fixes #3280` closes an issue whose **second of two named defects is untouched and unmentioned**, and whose own title names it.

### Refs (fetched, not trusted)

| | |
|---|---|
| `main` | `65578d41b2cc887596c1cda95674a4444390f685` |
| PR head (`gh pr view --json headRefOid`) | `eeef74a3ad9a372ad94db579d9cb6b0ca9c64c96` |
| merge-base | `f5642aff1598630710f29737baf3416273b0dcf4` (**3 commits behind `main`**) |
| stack base (#3272 snapshot carried) | `d259a8df7bddfa0fa4cf84339ea7dd7c544a6a27` |
| `merge-tree --write-tree main × head` | rc 0, 1 line, **0 stage entries** |
| `merge-tree --write-tree #3272(live `38c7b8bd5`) × head` | rc 0, **0 stage entries** |
| `merge-tree --write-tree #3283(live `b40320429`) × head` | rc 0, **0 stage entries** |

Merged with current `main` in a scratch worktree (`549fe0089`): Release build **0 warnings / 0 errors**, all 12 matrix rows still run, `Issue3280*` **9/9**. Merged with **#3283** (`b57d2dbf4`): same, plus the ref-struct rows below.

Freshness witness `out/bin/Release/Core/GSharp.Core.dll` = `b947ed90ec6007615ea53621448464ac`, built with `dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true`.

---

## Priority 1 — is the fix one predicate in `Conversion.Classify`? **Yes. Confirmed by grep, not by reading the diff.**

Repo-root grep for every site that can decide "may this value type become `System.ValueType`", `src/` only, `ref/` excluded:

```
src/Core/CodeAnalysis/Binding/Conversion.cs:840          <- THE rule (this PR)
src/Core/CodeAnalysis/Binding/Conversion.cs:2681         <- HasExplicitUnboxingConversion, reverse direction
src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Conversions.cs:448   <- emitter `box`, PRE-EXISTING (#1218), untouched
src/Core/CodeAnalysis/Emit/MethodBodyEmitter.cs:650               <- untouched
src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs:975       <- core-type resolution
src/Core/CodeAnalysis/Emit/ImportedMemberRefFactory.cs:452        <- memberref routing
src/Core/CodeAnalysis/Binding/OverloadResolution/ClrOverloadResolution.cs:4322  <- `struct` CONSTRAINT check, unrelated
src/Core/CodeAnalysis/StructValue.cs:507                          <- value-model base selection, unrelated
```

`git diff --name-only f5642aff1 eeef74a3a | grep MethodBodyEmitter` → **0**. The emitter's `box` for `ValueType`/`Enum` targets already existed at `MethodBodyEmitter.Conversions.cs:445-453` under #1218. The author's "No emitter change" claim is accurate and the classification is a **single predicate at `Conversion.cs:839-845`**. This is not four patched call paths. Credit where due.

The second product hunk (`OverloadResolver.CallBinding.cs:1707-1709`) is not a second spelling of the rule — it *materializes* a conversion the classifier now permits. It is also **byte-identical in product code to #3272's `38c7b8bd5`** (verified with `git show`), so the duplication is textual, not semantic.

---

## Priority 2 — the four-path matrix, driven not read

`gsc /target:exe /out:` → `dotnet exec`, rc captured before any pipe. 12 rows × 3 trees.

**Rows that now succeed** (identical stdout on bare `gsc`, `gsc /out:`+`dotnet exec`, and `gsi`):

| shape | free | instance | extension | `shared` |
|---|---|---|---|---|
| `func Take(v ValueType)`, `Take(42)` **(non-generic)** | before `GS0154` → **rc 0** | before `GS0155` → **rc 0** | before `GS0155` → **rc 0** | before `GS0155` → **rc 0** |
| `func Take(v Enum)`, `Take(DayOfWeek.Monday)` **(non-generic)** | before `GS0154` → **rc 0** | before `GS0155` → **rc 0** | before `GS0155` → **rc 0** | before `GS0155` → **rc 0** |
| `Show[T](value T)`, `Show[ValueType](22)` **(generic)** | before **rc 134 `InvalidProgramException`** → **rc 0** | before **rc 134** → **rc 0** | before **rc 134** → **rc 0** | before `GS0155` → **rc 0** |

All succeeding rows print `System.ValueType[]` / `System.Int32` / `42` (or `System.Enum[]` / `System.DayOfWeek` / `Monday`) — the array-element-type idiom distinguishes the boxed target from `System.Object[]` and `System.Int32[]`, so these are real type observations, not a value that would look the same either way.

Two things worth recording:
* On `main` the *generic* free/instance/extension rows were **not** diagnostics — they compiled rc 0 and crashed with `InvalidProgramException`. On the carried #3272 snapshot they are already honest `GS0154`/`GS0155`. Attribution to #3272 is correct.
* The issue's framing that "`Enum` is accepted while `ValueType` is refused" is **half right and the half matters**: on `main`, a *same-compilation* `enum Color` converts to `Enum`/`ValueType`/`object` fine (the old `from is EnumSymbol` clause), while an *imported* `System.DayOfWeek` is refused on all four paths. The PR closes the imported half without disturbing the same-compilation half — verified byte-identical `main` vs head on a user-enum probe, and pinned: mutant `M3` below.

---

## BLOCKER — #3280 defect **2 of 2** is untouched, and `Fixes #3280` will close it silently

`src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.CallBinding.cs:1682` (`ReportWrongArgumentType` → `GS0154`, free path) vs the member paths' `BindCallArgumentWithRefKind` → `GS0155`.

Issue #3280's **title** is "…**and the four call paths disagree (GS0154 vs GS0155)**", and its body says, of defect 2:

> *"Whatever the outcome of (1), two spellings of one rejection is a bug on its own — and if (1) is fixed, this asymmetry should disappear **rather than be papered over**."*

Measured on **this head**, and again on the head merged with current `main`:

```
string -> ValueType
  free      : (6,6,6,12)  error GS0154: Parameter 'v' requires a value of type 'System.ValueType' but was given a value of type 'string'.
  instance  : (9,10,9,16) error GS0155: Cannot convert type 'string' to 'System.ValueType'.
  extension : (6,10,6,16) error GS0155: Cannot convert type 'string' to 'System.ValueType'.
  shared    : (10,8,10,14) error GS0155: Cannot convert type 'string' to 'System.ValueType'.

[T struct] T -> ValueType
  free      : (5,36,5,41) error GS0154: ... but was given a value of type 'T'.
  instance  : (8,40,8,45) error GS0155: Cannot convert type 'T' to 'System.ValueType'.
  extension : (5,40,5,45) error GS0155: Cannot convert type 'T' to 'System.ValueType'.
  shared    : (9,38,9,43) error GS0155: Cannot convert type 'T' to 'System.ValueType'.
```

The asymmetry is fully alive. It was not fixed; it was made **unobservable on the one row the issue tabulated**. And the new test corpus makes that permanent: `Issue3280ValueTypeBaseBoxingTests.cs` contains **zero rejected-conversion rows through the four call shapes** — every matrix cell is a success cell — so nothing in the repo can ever see the disagreement again.

The PR body's Summary and its verification matrix never mention defect 2.

**Remedy — cheapest compliant option is not a code change.** Either:
1. drop `Fixes #3280` → `Refs #3280`, file the follow-up for the `GS0154`/`GS0155` free-vs-member split, and say in the body that defect 1 of 2 is addressed; **or**
2. unify the free path's rejection with the member paths and add the four refused rows above to the test matrix.

Option 1 is one line of PR body plus one issue. I am not asking for the diagnostic unification in this PR — I am asking that merging it not close an issue whose titled second half is still reproducible in four lines of G#.

---

## SHOULD-FIX 1 — "every CLR value type" is not every value type: `[T struct]` is excluded

`src/Core/CodeAnalysis/Binding/Conversion.cs:2699` `IsValueTypeLikeFrom` has arms for `NullableTypeSymbol`, `StructSymbol { IsClass: false }`, `EnumSymbol`, and `ClrType.IsValueType`. It has **no `TypeParameterSymbol` arm**, and `TypeParameterSymbol.ClrType` is `null`. So a `struct`-constrained parameter is not "value-type-like".

Consequence, measured on head:

```
func Fwd[T struct](value T) { TakeObj(value) }   -> rc 0, prints System.Int32
func Fwd[T struct](value T) { TakeVt(value)  }   -> rc 1, GS0154 at (7,12,7,17)
```

Within one function, `T` boxes to `object` and refuses to box to `ValueType`. The PR summary's first bullet — *"classify every CLR value type as implicitly boxable to `System.ValueType`"* — is inaccurate as written, and C# permits `where T : struct` → `ValueType`.

This is **pre-existing** (`main` refuses it too), so it is not a regression. It is a scope claim that overstates the delivery, and it is the natural next specimen for anyone reading the summary. Either extend `IsValueTypeLikeFrom` with a `HasValueTypeConstraint` arm, or narrow the summary and say the constrained-type-parameter shape is out of scope.

## SHOULD-FIX 2 — one parameter shape, four call shapes: census of specimens, sample of assertions

`test/Core.Tests/CodeAnalysis/Emit/Issue3280ValueTypeBaseBoxingTests.cs:171` `MatrixSource` emits **only** `func Show[T](value T)` + explicit `Show[ValueType](…)`. Every one of the 4 × 4 matrix cells is the *same* parameter shape: a bare type parameter closed by explicit substitution.

The shape issue #3280's own table names — a plain `func Take(v ValueType)` receiving an `int32` — **does not appear in the test file at all**. It is a genuinely different route through `OverloadResolver.CallBinding.cs` (`substitution == null`, so the `parameter.Type is TypeParameterSymbol` skip at `:1707` is never consulted). It works today (I measured all four call shapes, rc 0, correct types) but nothing pins it.

Add the four non-generic rows. They are four lines of `MatrixSource` and they are the rows the issue actually tabulated.

## SHOULD-FIX 3 — the carried #3272 snapshot is stale, and the coordination evidence cites a head that no longer exists

The PR carries three of #3272's commits (`49889d95f`, `ffe9d4688`, `d259a8df7`). #3272 has since been **rebased** — its live head is `38c7b8bd5` and its equivalents are `9bdb761a8` / `61a2e2a3b` / `b4963ef25`. `git merge-base --is-ancestor d259a8df7 ghr/pr3272` → **NO**.

The PR body's coordination section says *"Rechecked against #3272 head `d259a8df7bddfa…`: merge-tree rc 0, 0 stage entries."* That SHA is not #3272's head. I re-ran it against the live head — still **rc 0, 0 stages** — so there is no correctness risk, but the evidence as stated is against a head that no longer exists.

The real consequence is traceability: **if #3284 merges before #3272**, then a PR titled "Fix boxing conversions to ValueType and Enum" lands
* `OverloadResolver.Invocations.cs` +16/−14 (extension + instance generic argument conversion),
* the user-visible `GS0156` message rewrite (`DiagnosticDescriptors.cs:75`), and
* three documentation files,

all of which belong to #3267/#3272. Either state the merge order explicitly in the body, or land #3272 first and rebase this onto it so the diff is the two hunks it actually is.

---

## Priority 3 — over-reach. **None found.** 19 probes, `main` vs head, emit + run.

`ref struct` is the one that mattered, and the PR **improves** it:

| probe | `main` | head |
|---|---|---|
| user `ref struct Window` → `ValueType` (non-generic) | `GS0154` (wrong-arg-type) | **`GS0219`** *By-ref-like type 'Window' … cannot be boxed or converted to the reference type 'System.ValueType'* |
| imported `ReadOnlySpan[char]` → `ValueType` | `GS0154` | **`GS0219`** |
| `ReadOnlySpan[int32]` into `Show[ValueType](s)` (generic free call) | n/a | **`GS0219`**, rc 1, no assembly |
| user `ref struct` → `object` (control) | `GS0219` | `GS0219` |

The classifier now returns Implicit, so the call reaches `ConversionClassifier.cs:657-668`, which rejects any by-ref-like source converting to a non-value-type target — and `typeof(System.ValueType).IsValueType` is `false`, so `ValueType` is covered. Strictly better diagnostics, no hole.

**#3283 interaction, checked empirically rather than by reading.** #3283 replaces that guard's `type?.ClrType != null && !type.ClrType.IsValueType` with `Conversion.IsReferenceLikeTarget(type)`. `IsReferenceLikeTarget` (`Conversion.cs:1483`) falls through to `!clrBacking.IsValueType && !IsPointer && !IsByRef`, which is `true` for `System.ValueType` — so the guard survives. I built the actual merge (`b57d2dbf4`, 0 warnings / 0 errors) and re-ran all four ref-struct probes plus all 12 matrix rows: **`GS0219` still fires on every ref-struct row, all 12 rows still run correctly.** The two PRs are complementary; merge-tree is clean either order.

Remaining over-reach axes, all correct:

| axis | head | verdict |
|---|---|---|
| `*int32` → `ValueType` | rc 1 `GS0154` | refused ✓ |
| `*func(int32,int32) int32` → `ValueType` (via `&Add`, `unsafe`) | rc 1 `GS0154` | refused ✓ |
| unconstrained `T` → `ValueType` | rc 1 `GS0154` | refused ✓ (C# agrees) |
| `object` → `ValueType` | rc 1 `GS0154` | refused ✓ (C# agrees) |
| `string` → `ValueType` | rc 1 `GS0154` | refused ✓ |
| `int32` → `Enum` | rc 1 `GS0154` | refused ✓ |
| **reverse** `ValueType` → `int32` implicit | rc 1 `GS0154` | refused ✓ |
| explicit `int32(boxed)` unbox round trip | rc 0, `99` then `100` | works ✓ |
| `(int32, string)` tuple → `ValueType` | rc 0, `System.ValueTuple\`2[System.Int32,System.String]` | ✓ |
| `int32?` (present) → `ValueType` | rc 0, `System.Int32` / `9` | ✓ |
| `int32?` (nil) → `ValueType` | rc 0, prints `null` | ✓ |
| `ValueType?` parameter, `nil` and `7` | rc 0, `nil` then `System.Int32` | ✓ (was `GS0154` on `main`) |
| user `struct Point` → `ValueType` | rc 0, `D12.Point` | ✓ |
| user `enum Color` → `Enum`/`ValueType`/`object` | rc 0, all three | **byte-identical to `main`** ✓ |
| imported `DayOfWeek` → `ValueType` | rc 0, `System.DayOfWeek` / `Friday` | ✓ |
| `bool` / `float64` / `char` / `int64` → `ValueType` | rc 0, correct runtime types | ✓ |
| `ValueType` → `object` | rc 0, `System.Int32` | ✓ |
| assignment / `return` / field init / `[]ValueType{…}` element | rc 0, `System.Int32` ×4 | ✓ (was 5 × `GS0155` on `main`) |
| mutable struct via `ValueType` param, then mutate original | rc 0, `E14.Counter` then `61` | boxing copies, original intact ✓ |
| `enum.HasFlag` / `DayOfWeek.ToString()` | rc 0 on both trees | no regression ✓ |

`nil → ValueType` is refused on both trees (`GS0155`). C# allows it since `ValueType` is a class. Pre-existing, unchanged, out of scope — noting it only so it is not rediscovered as a regression.

---

## Priority 4 — IL and runtime

Emit is the oracle throughout: every accepted row above was compiled to an assembly and executed with `dotnet exec`, asserting **rc and full stdout**. All three drivers agree on every probe (bare `gsc` → same stdout + `Success.`; `gsc /out:` + `dotnet exec`; `gsi`).

**ILVerify gate** (`build/run-ilverify.sh`, `dotnet tool restore` rc 0): `123/123 verified (2 with documented suppressions, 0 verification failures, 0 compile failures)`, rc 0. Read the log, not just rc: **0** occurrences of `Assembly or module not found`, `IndexOutOfRange`, or `Unhandled exception`.

---

## RED-proofs and mutants — 5 killed, 1 no-result retargeted and disclosed

Own worktree, PRODUCT source only, condition inversion (never `if (true)`), applied asserted by `git diff` **and** `GSharp.Core.dll` md5 movement, restored by exact source-md5 round-trip.

| id | site | change | build | result |
|---|---|---|---|---|
| **RA** | `Conversion.cs:839-845` | revert hunk to the pre-PR `from is EnumSymbol && …` | rc 0, dll `b947ed90` → `1eceeb1f` | **KILLED** — `Issue3280*` 5 failed / 4 passed / 9 |
| **RB** | `CallBinding.cs:1707-1709` | revert to `!(substitution != null && parameter.Type is TypeParameterSymbol)` | rc 0, dll moved | **KILLED** — 1 failed / 8 passed; free generic probe reverts to `InvalidProgramException` |
| **M1** (control, expected to kill) | `Conversion.cs:840` | `IsValueTypeLikeFrom(from)` → `!IsValueTypeLikeFrom(from)` | rc 0, dll moved | **KILLED** — 6 failed / 3 passed |
| **M2** | `Conversion.cs:841` | `IsEnumLikeType(from)` → `!IsEnumLikeType(from)` | rc 0, dll moved | **KILLED** — `Issue3280*` 5 failed; **`Issue1218EnumMemberEmit` 2 failed** |
| **M3** | `Conversion.cs:842` | `from is EnumSymbol` → `from is not EnumSymbol` (the retained #1218 `object` arm) | rc 0, dll moved | **KILLED** — by `Compiler.Tests/Issue1218EnumMemberEmit` (1 failed / 4); the PR's own `Issue3280*` passed **9/9** |

RA's five failures are all `Assert.Empty(result.Diagnostics)` at `Issue3280ValueTypeBaseBoxingTests.cs:215` — i.e. the conversion is refused. Failing for the right reason. The four survivors under RA are exactly the four over-correction controls, which is what should happen.

**Disclosure — one no-result, not a survival.** My first M3 attempt deleted the `object` arm textually and produced `CS1525/CS1002/CS1513` at `Conversion.cs:841`; `BUILD_RC=1`, and the `GSharp.Core.dll` md5 **did not move** (a failed build leaves the previous binary, which is exactly how a no-result masquerades as a survival). Retargeted as a condition inversion and rescored. Also disclosed: my first managed-function-pointer probe failed at its *declaration* line (`GS0155` on `Add` → `*func(…)`), which is a no-result for that axis; re-ran with `&Add` and got the clean `GS0154` row now tabulated.

**M3 is worth one sentence on its own.** The PR's new test file passes 9/9 with the retained `#1218` `EnumSymbol → object` clause inverted. That clause is load-bearing (a user enum stops converting to `object`), and it is pinned — but by `Compiler.Tests/Emit/Issue1218EnumMemberEmitTests.cs`, not by anything added here. The PR was right to retain it; just noting that the retention is not self-pinned.

Restore verified: `git diff HEAD --stat` empty; `Conversion.cs` md5 `a67f4f84e5132640dc0278a5d7d3bc7a` and `OverloadResolver.CallBinding.cs` md5 `98cf4d49229d2ff908d74c5e24f11960`, both exactly matching an untouched head worktree.

---

## NIT 1 — two test names state properties their assertions cannot observe

* `Issue3280ValueTypeBaseBoxingTests.cs:142` `MutableStructPassedByReference_IsNotBoxedOrCopied` asserts a single line, `"61"`. If the value *were* boxed, `value.Bump()` would mutate the box and `return value` would hand it back — `result.Read()` is `61` either way. The assertion is invariant under the property the name claims.
* `Issue3280ValueTypeBaseBoxingTests.cs:43` `AlreadyBoxedValueTypeArgument_IsNotBoxedAgain` has the same shape: a redundant box on an already-reference value would print the identical three lines.

Both are one line from being real: the file already uses `Console.WriteLine([]T{value}.GetType())` everywhere else, and that line *does* separate `Counter[]` from `Object[]`. (ADR-0154.)

## NIT 2 — the `GS0156` rewrite removes advice that is true and works

`src/Core/CodeAnalysis/DiagnosticDescriptors.cs:75` (inherited from #3272, carried here) changes
`"…An explicit conversion exists (are you missing a cast?)"` → `"…Provide a value of type '{1}' instead."`

I followed the old advice on this head: `var x int32 = int32(3.14)` compiles rc 0 and prints `3`. The explicit conversion **does** exist and **does** work, so the removed sentence was accurate and the replacement is strictly less informative for `GS0156` specifically. Not this PR's call to make — flagging it because this PR is the one that would land it.

---

## Suites — denominator **9**, source `dotnet sln GSharp.sln list` (CI's own filter, `build/generate-ci-test-matrix.py:26`)

Run on the PR head, `-c Release`, no `--no-build`. Executed counts only.

| project | result |
|---|---|
| Core.Tests | 7205 passed, 1 skipped |
| Compiler.Tests | 4267 passed |
| Cs2Gs.Tests | 1907 passed |
| Interpreter.Tests | 1353 passed, 3 skipped |
| Sdk.Tests | 59 passed |
| **total** | **15,403 passed, 4 skipped, 0 failed**, every project rc 0 |

Matches the PR body exactly, including `LanguageServer.Tests` 468/468 — which matters here because this PR touches the binder and that project consumes binder diagnostics.

DavidObando:
Addressed round-1 findings in pushed head `534f1272a`:

- rebased onto current `main`; #3272 duplicate commits dropped
- added direct non-generic four-shape ValueType/Enum runtime rows
- added `struct`-constrained `T -> ValueType` classification and runtime coverage
- replaced #3272 temporary diagnostic pins with runtime assertions
- narrowed closure scope: `Refs #3280`; rejected-call GS0154/GS0155 consistency is tracked in #3312
- clarified the two overclaiming test names
- Release build and focused boxing/generic/enum tests pass locally

Post-push CI is running.

Issue #1481, opened 2026-06-30
Consolidate the three divergent ContainsTypeParameter structural recursions (emit copies omit MapTypeSymbol/FunctionTypeSymbol, erasing iterator element types)


DavidObando:
## Summary

The emit layer re-implements the "does this type structurally reference a type
parameter?" recursion **three separate times**, and the copies have **diverged**
in which composite type-kinds they recurse into. The two emit-layer copies omit
`MapTypeSymbol` and `FunctionTypeSymbol`, so a generic iterator / async-iterator
whose element type is built from a `map[...]` or a function type erases its
`IEnumerable<T>` / `IEnumerator<T>` interface rows and `GetEnumerator` signature
to the `<object>` shape — the exact class of bug that #1418, #990, #810 and
#813 each patched one container-kind at a time.

This is the same generalization smell motivating #1418: a structural walk over
the type lattice that is hand-copied and patched per-shape instead of routed
through one canonical predicate.

## The three copies

1. **Canonical** — `TypeSymbol.ContainsTypeParameter`
   (`src/Core/CodeAnalysis/Symbols/TypeSymbol.cs:324`). Recurses into
   `Nullable`, `Slice`, `Array`, **`Map` (key+value)**, **`Function`
   (params+return)**, `ImportedType` args, `Tuple` elements.

2. **Emit copy A** — `MethodBodyPlanner.ContainsTypeParameter`
   (`src/Core/CodeAnalysis/Emit/MethodBodyPlanner.cs:894`). Recurses into
   `Array`, `Slice`, `Sequence`, `AsyncSequence`, `Nullable`, `ImportedType`,
   `StructSymbol` args, `Tuple`. **Missing: `Map`, `Function`.**

3. **Emit copy B** — `StateMachineEmitter.ContainsOuterMethodTypeParameter`
   (`src/Core/CodeAnalysis/Emit/StateMachineEmitter.cs:579`). Recurses into
   `Array`, `Slice`, `Sequence`, `AsyncSequence`, `Nullable`, `ImportedType`,
   `Tuple`. **Missing: `Map`, `Function`, `StructSymbol` args.** (It also uses a
   different leaf check — membership in `outerMethodTPs` rather than "is any type
   parameter".)

All three feed the *same* decision: whether to encode the iterator state
machine's `IEnumerable<…>` / `IEnumerator<…>` interfaces and `GetEnumerator`
return as a symbolic `TypeSpec` (strongly typed) versus the CLR-erased
[...]

PR #3287, opened 2026-08-05
Validate arguments after generic inference

### LINKED ISSUE #3282: [binder] Generic inference binds T from the first argument and never checks later ones: Show(131, "str") compiles clean and emits invalid IL
## Summary

Generic type inference binds `T` from the **first** argument and then accepts every later argument without checking it against the binding. The call compiles rc 0 with **zero diagnostics** and dies at runtime.

```gs
func Show[T](a T, b T) { ... }

Show(131, "str")
```

- `gsc` → **rc 0, zero diagnostics**
- run → **rc 134**, `System.InvalidProgramException`

ILVerify names the defect precisely:

```
[found ref 'string'][expected Int32]
```

`T` is inferred as `int32` from argument 0, and the `string` in argument 1 is accepted silently against that binding.

Not limited to reference-vs-value. Two unrelated struct types show the same shape:

```gs
Show(Mark{V: 1}, Other())     // → [found ref Other][expected value Mark]
```

**The explicit form is correctly handled**: `Show[object](131, "str")` compiles and runs fine, because both arguments then convert to a target that accepts them.

## Why this is separate from #3271

This has the same *signature* as #3271 — a legal-looking generic call, rc 0 with no diagnostics, `InvalidProgramException` at runtime — but a different root cause. #3271 was a missing boxing conversion for an **explicitly supplied** type argument, fixed in the binder at `OverloadResolver.CallBinding.cs`. This one is in **inference**: no type argument is supplied at all, and the defect is that argument 1 is never validated against the `T` that argument 0 established.

#3267's fix validates **explicit** generic arguments against call arguments. This issue is the **inferred** counterpart of that same check, and it is plausible — but unverified — that the fix lives adjacent to it.

## Expected behaviour

Either infer a common supertype that accommodates both arguments, or reject the call with a diagnostic naming the conflicting inference. **Silently emitting invalid IL is not an option.** If a diagnostic is chosen, it must name both the binding source and the conflicting argument, so the reader can see why `T` became `int32`.

## Notes for the fix

- **The oracle is the running program.** rc 0 with zero diagnostics *is* the bug, so a compile-only check proves nothing. Compile → `dotnet exec` → assert rc **and** stdout.
- **Erasure blindness will hide this.** `Console.WriteLine(object)` renders `42` and `"right"` identically. Use a witness that reveals the reified type:
  ```gs
  func Run[T](v T) { Console.WriteLine([]T{v}.GetType()) }
  ```
  `[]T{v}.GetType()` prints `System.Object[]` vs `System.Int32[]`, so boxed and unboxed paths produce different strings. Print the value on the next line too — type-correct-but-value-wrong is a separate failure mode.
- **Write the non-generic control for every generic row.** `func Plain(v object)`, `PlainVT(v ValueType)`, `PlainCmp(v IComparable)` measured on the same build. This distinguishes "generic inference is broken" from "G# refuses this conversion everywhere", and it has already reversed one reviewer's conclusion on adjacent code.
- **`ilverify` is a useful diagnostic and a worthless pass-oracle.** It named the stack-type mismatch quoted above, but it also returns rc 0 while printing `Assembly or module not found: mscorlib`, and it green-lights the invalid program in #3240. Never treat a clean ILVerify as proof.
- Cover arity > 2, `T` appearing in a nested position (`[]T`, `T?`, tuples), and the case where the *later* argument is the more general one (`Show("str", 131)`) — the fix must not depend on argument order.

DavidObando:
## Summary

- add regression coverage for inferred generic argument validation from issue #3282
- compare every invalid generic call with its equivalent closed non-generic call, including diagnostic ID and exact argument span
- cover reversed arguments, arity > 2, nested `[]T` / `T?` / tuple slots, unrelated structs, instance methods, and receiver-clause extensions
- retain runtime oracles for valid inferred calls and explicit `Show[object](131, "str")` boxing

## Relationship to #3272

The product fix landed in `main` through #3272 while this PR was under review. After rebasing, the byte-identical binder, diagnostic, documentation, and existing-test changes dropped out. This PR is now intentionally test-only and does not claim a new product-code change.

The tests pin the interaction between inferred validation and #3272's explicit generic argument validation: closed substituted targets are validated and converted, while valid explicit `object` arguments remain boxed and execute correctly.

## Validation

- rebased onto `6c7f1a9a1`
- `dotnet test test/Core.Tests/Core.Tests.csproj -c Release --filter 'FullyQualifiedName~Issue3282GenericInferenceArgumentValidationTests|FullyQualifiedName~Issue3267ExplicitGenericArgumentValidationTests|FullyQualifiedName~Issue3271ReifiedGenericArgumentBoxingTests' --no-restore`
- **42 passed, 0 failed, 0 skipped**

Fixes #3282

### DIFF
--- a/test/Core.Tests/CodeAnalysis/Binding/Issue3282GenericInferenceArgumentValidationTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Binding/Issue3282GenericInferenceArgumentValidationTests.cs
@@ -1,0 +1,268 @@
+// <copyright file="Issue3282GenericInferenceArgumentValidationTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using GSharp.Core.CodeAnalysis;
+using GSharp.Core.CodeAnalysis.Syntax;
+using GSharp.Core.CodeAnalysis.Text;
+using GSharp.Tests;
+using Xunit;
+
+namespace GSharp.Core.Tests.CodeAnalysis.Binding;
+
+/// <summary>
+/// Issue #3282: arguments must be checked against the type inferred from
+/// earlier arguments before a generic call is emitted.
+/// </summary>
+public class Issue3282GenericInferenceArgumentValidationTests
+{
+    public static TheoryData<string, string, string, string> InvalidCalls => new()
+    {
+        {
+            """
+            func Show[T](a T, b T) { }
+            Show(131, "str")
+            """,
+            """
+            func Show(a int32, b int32) { }
+            Show(131, "str")
+            """,
+            "GS0154",
+            "\"str\""
+        },
+        {
+            """
+            func Show[T](a T, b T) { }
+            Show("str", 131)
+            """,
+            """
+            func Show(a string, b string) { }
+            Show("str", 131)
+            """,
+            "GS0154",
+            "131"
+        },
+        {
+            """
+            func Show[T](a T, b T, c T) { }
+            Show(131, 132, "str")
+            """,
+            """
+            func Show(a int32, b int32, c int32) { }
+            Show(131, 132, "str")
+            """,
+            "GS0154",
+            "\"str\""
+        },
+        {
+            """
+            func Show[T](source T, later []T) { }
+            Show(131, []string{"str"})
+            """,
+            """
+            func Show(source int32, later []int32) { }
+            Show(131, []string{"str"})
+            """,
+            "GS0154",
+            "[]string{\"str\"}"
+        },
+        {
+            """
+            func Show[T](source T, later T?) { }
+            Show(131, "str")
+            """,
+            """
+            func Show(source int32, later int32?) { }
+            Show(131, "str")
+            """,
+            "GS0154",
+            "\"str\""
+        },
+        {
+            """
+            func Show[T](source T, later (T, T)) { }
+            Show(131, ("str", "other"))
+            """,
+            """
+            func Show(source int32, later (int32, int32)) { }
+            Show(131, ("str", "other"))
+            """,
+            "GS0154",
+            "(\"str\", \"other\")"
+        },
+        {
+            """
+            struct Mark { var V int32 }
+            struct Other { }
+            func Show[T](a T, b T) { }
+            Show(Mark{V: 1}, Other{})
+            """,
+            """
+            struct Mark { var V int32 }
+            struct Other { }
+            func Show(a Mark, b Mark) { }
+            Show(Mark{V: 1}, Other{})
+            """,
+            "GS0154",
+            "Other{}"
+        },
+        {
+            """
+            class Box {
+                func Show[T](a T, b T) { }
+            }
+            Box().Show(131, "str")
+            """,
+            """
+            class Box {
+                func Show(a int32, b int32) { }
+            }
+            Box().Show(131, "str")
+            """,
+            "GS0156",
+            "\"str\""
+        },
+        {
+            """
+            func (self string) Show[T](a T, b T) { }
+            "receiver".Show(131, "str")
+            """,
+            """
+            func (self string) Show(a int32, b int32) { }
+            "receiver".Show(131, "str")
+            """,
+            "GS0156",
+            "\"str\""
+        },
+    };
+
+    public static TheoryData<string, string> ValidCalls => new()
+    {
+        {
+            """
+            import System
+            func Show[T](a T, b T) {
+                Console.WriteLine([]T{a}.GetType())
+                Console.WriteLine(a)
+                Console.WriteLine(b)
+            }
+            Show(131, 132)
+            """,
+            "System.Int32[]\n131\n132\n"
+        },
+        {
+            """
+            import System
+            func Show[T](a T, b T) {
+                Console.WriteLine([]T{a}.GetType())
+                Console.WriteLine(a)
+                Console.WriteLine(b)
+            }
+            Show("str", "other")
+            """,
+            "System.String[]\nstr\nother\n"
+        },
+        {
+            """
+            import System
+            func Show[T](source T, slice []T) {
+                Console.WriteLine([]T{source}.GetType())
+                Console.WriteLine(source)
+                Console.WriteLine(slice[0])
+            }
+            Show(131, []int32{132})
+            """,
+            "System.Int32[]\n131\n132\n"
+        },
+        {
+            """
+            import System
+            func Show[T](a T, b T) {
+                Console.WriteLine([]T{a}.GetType())
+                Console.WriteLine(a)
+                Console.WriteLine(b)
+            }
+            Show[object](131, "str")
+            """,
+            "System.Object[]\n131\nstr\n"
+        },
+        {
+            """
+            import System
+            func Plain(v object) {
+                Console.WriteLine(v.GetType())
+                Console.WriteLine(v)
+            }
+            Plain(131)
+            Plain("str")
+            """,
+            "System.Int32\n131\nSystem.String\nstr\n"
+        },
+    };
+
+    [Theory]
+    [MemberData(nameof(InvalidCalls))]
+    public void ConflictingInferredArgument_MatchesNonGenericDiagnostic(
+        string genericSource,
+        string nonGenericSource,
+        string expectedId,
+        string expectedSpan)
+    {
+        AssertDiagnostic(genericSource, expectedId, expectedSpan);
+        AssertDiagnostic(nonGenericSource, expectedId, expectedSpan);
+    }
+
+    [Theory]
+    [MemberData(nameof(ValidCalls))]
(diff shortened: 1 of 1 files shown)


DavidObando:
## ⚠️ APPROVE AS TEST-ONLY AFTER REBASE — the product fix landed in `main` 20 minutes ago via #3272

Verdict against **`main` = `6764eeb6b00a52f9ba27b75b71d66020d559f6e0`** (re-fetched at post time), head **`02959f7aad270487172932211b80712a20c6aa41`**.

**`main` moved under me mid-review**: it was `65578d41b` when I started — the branch's merge-base — and is now `6764eeb6b`. Three commits landed: `2d06ccc78` (#3275), **`1f4ef2daa` "Validate explicit generic calls and reified boxing (#3272)"**, `6764eeb6b` (#3276).

**#3272 carried the byte-identical binder hunks.** Every product and doc file in this PR is now the same blob as `main`:

| file | vs current `main` |
|---|---|
| `OverloadResolver.CallBinding.cs` | **SAME-AS-MAIN** |
| `OverloadResolver.Invocations.cs` | **SAME-AS-MAIN** |
| `DiagnosticDescriptors.cs` | **SAME-AS-MAIN** |
| `docs/diagnostics.md`, `website/docs/ref/diagnostics.md`, `docs/adr/0099-…` | **SAME-AS-MAIN** |
| `Issue2851…Tests.cs`, `Issue2947ImportedNestedGenericTypeTests.cs` | **SAME-AS-MAIN** |
| **`Issue3282GenericInferenceArgumentValidationTests.cs`** | **the only unique content** |

I did not take that from the file list — I built `main` at `6764eeb6b` (`GSharp.Core.dll` md5 `e59c05955253ec3e136f7a05057c4771`) and re-ran my whole corpus through `gsc`. **Every row is identical to this PR's head**: same id, same span, same message, same rc — `Show(131,"str")`→`GS0154`, instance/receiver/shared→`GS0156`, `[]T`/`T?`/tuple→exact closed target, `Show(131,132)`→`System.Int32[] / 131 / 132`.

**Issue #3282 is already closed by `main`. This PR's product delta is now empty.**

That is not a rejection. The test file is the valuable half and it **passes 15/15 on current `main` unmodified**. Rebase, and this becomes a clean test-only contribution worth landing. Two asks:

1. **Update the body.** The Behavior table and the "Validation" bullets describe a change this PR no longer makes. Merged as written it would attribute #3272's fix to #3287.
2. `docs/adr/0099`, both `diagnostics.md` files, `DiagnosticDescriptors.cs` and both binder files should drop out of the diff entirely on rebase. If any survives, it is a revert of `main`.

The `Issue2947…DriverTests.cs` difference is benign: this branch leaves the docstring at its base value while #3276 rewrote it in `main`, so the three-way merge keeps `main`'s. I confirmed `merge-tree --write-tree` **by stages** — **0 stage entries** against `main`, #3272, #3283 and #3284.

---

## The measurements, since they now certify what is in `main`

Everything below was run twice: once at this head, once with **only the two binder files reverted to `65578d41b`** (rebuild green, md5 `e462439b…`→`68362822…`), which reproduces pre-#3272 `main`.

### The defect and its blast radius — RED confirmed

| probe | pre-fix | now |
|---|---|---|
| `Show(131,"str")` | rc 0, **run rc 134** `InvalidProgramException` | `GS0154` @ `"str"` |
| `Show("str",131)` | rc 0, run rc 134 | `GS0154` @ `131` |
| instance `Box().Show` | rc 0, run rc 134 | `GS0156` @ `"str"` |
| receiver-clause | rc 0, run rc 134 | `GS0156` @ `"str"` |
| **`shared { }` static** | rc 0, run rc 134 | `GS0156` @ `"str"` |
| **generic class primary ctor** | — | `GS0154` @ `"str"` |

Two findings the PR does not claim:

- **`Show[object](131,"str")` was broken too.** Issue #3282 states the explicit form "is correctly handled". **It was not** — pre-fix it gave rc 0 then **rc 134**. It now prints `System.Object[] / 131 / str`. The issue's premise was wrong.
- **Silent value corruption, not just crashes.** `let y int64 = 300000000000L; Show(1, y)` compiled and *ran clean* pre-fix, printing **`-647710720`**. No crash, no diagnostic, wrong answer — worse than the `InvalidProgramException` because nothing announces it. Now `GS0154`.

Five call paths are covered, not the three claimed. Generic **local functions** do not exist in G# (parse error), so there is no gap there.

### Over-reach: none, across every legal shape I could construct

`Show(131,132)` · `Show("a","b","c")` · `Show(131,[]int32{132})` · `Show[object](131,"str")` · upcast `Show(animal, Dog())`→`Default.Animal[]` · interface-typed `Show(s1,s2)`→`Default.Speaker[]` · `object`-typed args · `[T any]` · `[T comparable]` (prints `False`/`True`) · `[T,U]` — **all compile rc 0 and run rc 0.**

Two rows newly reject, and I chased both rather than assuming:

- **`Show("x", nil)`** — the **non-generic** control `func Show(a string, b string); Show("x", nil)` is rejected on pre-fix `main` with the **identical id and the identical span** `(3,11,3,14)`. Illegal G# all along; the generic path was the outlier.
- **`Show(Dog(), Cat())`** — pre-fix this compiled *and ran*. But add a line that uses `b` as a `T` and pre-fix `main` gives **rc 134 `ArrayTypeMismatchException`**. It was latently unsound; it only "worked" while the body never touched the argument.

Both newly-rejected rows are illegal. **No legal program is newly rejected.**

`Show(1, 2L)` is *accepted* and prints `1 / 2` — correct: `let x int32 = 2L` also compiles, so generic matches non-generic. The non-constant and out-of-range int64 forms are both rejected.

### `GS0154` vs `GS0156` — pre-existing, and conformed to exactly

| shape | non-generic control | generic inferred |
|---|---|---|
| free function | `GS0154` | `GS0154` |
| class instance | `GS0156` @ `(3,17,3,22)` | `GS0156` @ `(3,17,3,22)` |
| receiver-clause | `GS0156` @ `(3,22,3,27)` | `GS0156` @ `(3,22,3,27)` |
| `shared { }` | `GS0156` @ `(7,16,7,21)` | `GS0156` @ `(7,16,7,21)` |

The split is **accidental in origin** — free-function argument binding names the parameter, instance/receiver binding routes through conversion classification — but it long predates this work, and the design rule "match the equivalent closed call" is the right one. Unifying them is scope creep. Not a finding.

### Spans — 13 of 13 slice to exactly the conflicting argument

`"str"` · `131` · `nil` · `Cat()` · `y` · `Other{}` · `[]string{"str"}` · `("str", "other")` · `"bad"` — never the call head, never the statement.

### `ContainsTypeParameter` is the canonical predicate — no seventh spelling

`TypeSymbol.cs:463`, `public static bool ContainsTypeParameter(TypeSymbol) => AnyTypeParameter(type, static _ => true)`, documented at `:339` as *"the canonical"* walk that issue **#1481** created precisely to collapse three hand-copied divergent recursions. 43 call sites across 20 files, including `ExpressionBinder.Access.Accessor.cs:2672`. **Zero new predicates added.** Correct reuse.

### RED / suite

Reverting only the two binder files: **11 of 15 fail**, every one with `Assert.Single() Failure: The collection was empty` — the no-diagnostic shape, not a test compile error. (Reported as 10/4; I measure 11/4 of 15 reverting both files whole.)

Full nine-project suite at this head: **15,408 passed, 4 skipped, 0 failed, rc 0** — matches the report exactly. `GSharp.Core.dll` md5 **`e462439ba117fa46552cde3543423651`**, reproduced from `dotnet build GSharp.sln -c Release -p:ContinuousIntegrationBuild=true`, rc 0, 0 warnings 0 errors.

---

## One thing that rode in on #3272 and is now in `main` — worth a follow-up issue, not a change here

`DiagnosticDescriptors.cs:75` replaced GS0156's *"An explicit conversion exists (are you missing a cast?)"* with *"Provide a value of type '{1}' instead."* I followed the old advice on every GS0156 I could reach:

| GS0156 case | does the cast work? |
|---|---|
| `float64`→`int32` | **yes** — prints `3` |
| `int64`→`int32` | **yes** — prints `-647710720` |
| `float64`→`float32` | **yes** — prints `3.14` |
| `int32`→`int8` | **yes** — prints `44` |
| `int32?`→`int32` (the case ADR-0099 names as typical) | **yes** — prints `5` |
| `string`→`int32` | **no** — `GS9998: NotSupportedException: Conversion from 'string' to 'int32' is not yet supported by the emitter.` |

The cast advice was **true and directly actionable in five of six** cases, including the exact example `docs/diagnostics.md` used to document GS0156 (`int x = 3.14`). The new text is not false, but it is strictly weaker there — it tells you to change your data when a one-token cast is the real fix.

And the crash the rewording avoids advertising **is still reachable**: `bx.Show(131, int32("str"))` still produces `GS9998` at this head and on `main`. The binder classifies `string`→`int32` as having an explicit conversion, then the emitter cannot perform it. That is the actual defect; the message change is a symptom fix.

**Not blocking this PR** — the hunk is byte-identical to what #3272 already merged, so it is `main`'s state now, and reverting it here would be out of scope. Flagging so the `GS9998` can be filed on its own.

**No blocking issues in this PR's remaining content.**

Issue #3285, opened 2026-08-05
[emit] Optional pointer parameter defaulted to nil emits ldnull, compiles clean then throws InvalidProgramException


DavidObando:
## Summary

`EmitDefault` materialises a null for `PointerTypeSymbol` with `ldnull` instead of `ldc.i4.0; conv.i`. An optional pointer parameter defaulted to `nil` therefore **compiles rc 0, writes an assembly, and dies at runtime**.

## Repro

```gs
unsafe func take(p *int32 = nil) { Console.WriteLine(nint(p)) }
take()
```

`gsc` → **rc 0**, no diagnostic, assembly written. `dotnet exec` → **`System.InvalidProgramException`, rc 134**.

IL at the call site is `14 28` = `ldnull; call take(int32*)`. The correct sequence is `16 d3` = `ldc.i4.0; conv.i`, which is exactly what `EmitConversion` emits for the same conversion in every other position.

## Mechanism

Two sites materialise a null pointer, and they have **inverse** one-type omissions:

| site | `*T` | `*func(...)` |
|---|---|---|
| `MethodBodyEmitter.Conversions.cs:44` `EmitConversion` | `ldc.i4.0; conv.i` ✅ | `ldc.i4.0; conv.i` ✅ (fixed by #3268 / PR #3275) |
| `MethodBodyEmitter.Conversions.cs:1180` `EmitDefault` | **`ldnull`** ❌ | `initobj` |

PR #3275 correctly declined to widen its scope to this site. This is the remaining half of the class.

## Expected

Either emit the correct null-pointer sequence, or reject at compile time. **It must not compile clean and then produce an invalid program.**

## Notes for the fix

- **The oracle is the running program.** `ilverify` is not a gate for this class — the emit tests at `test/Compiler.Tests/.../Issue1035FunctionPointerEmitTests.cs:28-35` ignore `StackUnexpected`, the exact code this defect raises, and the verifier was silent under a mutant reproducing it. Compile → `dotnet exec` → assert rc **and** stdout.
- Fix the **class**, not the spelling: `EmitDefault` and `EmitConversion` disagree about the same conversion. Route both through one predicate rather than adding a third spelling. This session has already burned review rounds on a rule spelled six ways across ten sites.
- Check the sibling default-value positions too: optional parameter, field initialiser, `default`-like expressions, array element defaults, and both pointer kinds in each.
[...]

PR #3290, opened 2026-08-06
Allow function pointer comparisons with nil

### LINKED ISSUE #3286: [binder] 'fp == nil' on a managed function pointer is GS0129 while '*int32 == nil' binds, leaving nil function pointers untestable
## Summary

`fp == nil` and `fp != nil` on a managed function pointer report **GS0129** *"Binary operator '==' is not defined for types '\*func(int32) int32' and 'nil'"*, while the identical comparison on an unmanaged pointer `*int32 == nil` binds and prints `True`.

## Repro

```gs
unsafe {
    var fp *func(int32) int32 = nil
    Console.WriteLine(fp == nil)     // GS0129
    var p *int32 = nil
    Console.WriteLine(p == nil)      // True
}
```

## Why this matters now

Issue #3268 (PR #3275) made `nil` a fully working value for function pointers in every position — parameter, return, assignment, struct field, array element, ternary. So a user can now **produce** a nil function pointer everywhere but cannot **test** for one before calling it.

Calling through a nil function pointer is **SIGSEGV, rc 139**, with no managed exception — correct `calli` semantics, and the same as C# `delegate*`. That makes the guard the only safe way to use the feature, and the natural spelling of the guard does not bind.

The workaround is `nint(fp) == 0`, which is measured to work but requires the user to know that a function pointer round-trips through `nint`.

## Expected

`fp == nil` and `fp != nil` bind, for the same reason `*T == nil` binds. Check both operand orders (`nil == fp`), both operators, and fnptr-to-fnptr comparison while in the area.

## Notes for the fix

- Find where `*T == nil` is made legal and extend **that** predicate rather than adding a parallel arm for function pointers. `FunctionPointerTypeSymbol` is `sealed : TypeSymbol` and disjoint from every delegate symbol, so widening a pointer predicate cannot accidentally capture a reference type — this is the same shape as the one-line fix in PR #3275.
- **The oracle is the running program.** Assert the comparison's runtime answer, not just that it compiles: pin `fp == nil` → `True` for a nil pointer and `False` for `&someFunc`. A binder-only pin would pass on an emission that always yields `False`.
- Assert diagnostic **id and span** on any negative pins by slicing `Location.Text.ToString(Location.Span)`.
- Function-pointer IL is unverifiable by design, so `ilverify` proves nothing here.

Found by the rubber-duck review of PR #3275 while probing the widened conversion predicate for under-reach. Pre-existing on `main`.

DavidObando:
Fixes #3286.

## Summary

- Extend the existing native-pointer binary-operation gate to `FunctionPointerTypeSymbol` instead of adding a parallel function-pointer comparison path.
- Generalize the shared comparison lowering's nil target so both unmanaged data pointers and function pointers lower through `nint`.
- Add an end-to-end runtime matrix covering operand order, equality/inequality, managed and Cdecl function pointers, function-pointer pairs, and local/field/array/return value sources.

## Interaction with #3275

This branch is based on `main` after #3275. Its `nil -> FunctionPointerTypeSymbol` emission supplies the zero native pointer consumed by the existing shared comparison lowering; no #3275 files are modified here.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Operators.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Operators.cs
@@ -355,10 +355,9 @@ internal sealed partial class ExpressionBinder
     }
 
     /// <summary>
-    /// ADR-0122 / issue #1014: lowers a pointer arithmetic or comparison
-    /// binary expression where at least one operand is an unmanaged pointer
-    /// (<see cref="PointerTypeSymbol"/>). Returns <see langword="null"/> when
-    /// the operator/operand shape is not a supported pointer operation, so the
+    /// ADR-0122 / issues #1014 and #3286: lowers unmanaged-pointer arithmetic
+    /// and native pointer comparisons. Returns <see langword="null"/> when the
+    /// operator/operand shape is not a supported pointer operation, so the
     /// caller falls back to the normal (error-reporting) path.
     /// </summary>
     /// <param name="syntax">The binary expression syntax.</param>
@@ -584,7 +583,12 @@ internal sealed partial class ExpressionBinder
 
     private BoundExpression LowerPointerComparison(SyntaxKind operatorKind, BoundExpression left, BoundExpression right)
     {
-        var pointerType = (left.Type as PointerTypeSymbol) ?? (right.Type as PointerTypeSymbol);
+        if (!CanLowerPointerComparisonOperand(left.Type) || !CanLowerPointerComparisonOperand(right.Type))
+        {
+            return null;
+        }
+
+        var pointerType = left.Type == TypeSymbol.Null ? right.Type : left.Type;
         var leftNint = ToNativeIntForPointerComparison(left, pointerType);
         var rightNint = ToNativeIntForPointerComparison(right, pointerType);
         var op = BoundBinaryOperator.Bind(operatorKind, TypeSymbol.NInt, TypeSymbol.NInt);
@@ -596,7 +600,10 @@ internal sealed partial class ExpressionBinder
         return new BoundBinaryExpression(null, leftNint, op, rightNint);
     }
 
-    private BoundExpression ToNativeIntForPointerComparison(BoundExpression operand, PointerTypeSymbol pointerType)
+    private static bool CanLowerPointerComparisonOperand(TypeSymbol type) =>
+        type == TypeSymbol.Null || Conversion.Classify(type, TypeSymbol.NInt).Exists;
+
+    private BoundExpression ToNativeIntForPointerComparison(BoundExpression operand, TypeSymbol pointerType)
     {
         if (operand.Type == TypeSymbol.NInt)
         {
@@ -604,7 +611,7 @@ internal sealed partial class ExpressionBinder
         }
 
         // `nil` becomes a null pointer (zero native int) before the comparison.
-        if (operand.Type == TypeSymbol.Null && pointerType != null)
+        if (operand.Type == TypeSymbol.Null)
         {
             var nullPointer = new BoundConversionExpression(null, pointerType, operand);
             return new BoundConversionExpression(null, TypeSymbol.NInt, nullPointer);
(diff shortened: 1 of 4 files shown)


DavidObando:
## Evidence

### Base and root cause

- Originally branched from fetched `origin/main` `65578d41b2cc887596c1cda95674a4444390f685`.
- Rebased after #3275 merged; final base is `6764eeb6b00a52f9ba27b75b71d66020d559f6e0`.
- Head: `4477f959c877c92ac64b0568d4606453dfd8a2c9`.

`*T` comparisons enter `TryBindPointerBinaryExpression` from the single pointer predicate in `BindBinaryExpression`, before normal `BoundBinaryOperator` lookup. Caller grep found:

- `TryBindPointerBinaryExpression`: one definition, one caller (`BindBinaryExpression`).
- `TryBindPointerBinaryOperation`: one definition, two callers (the binary wrapper and compound-assignment reuse).

The fix widens that sole binary-expression predicate from `PointerTypeSymbol` to `PointerTypeSymbol or FunctionPointerTypeSymbol`; no parallel function-pointer arm was added. `LowerPointerComparison` also had to generalize its nil target from `PointerTypeSymbol` to `TypeSymbol`: with only the gate widened, `fp == nil` reached emission as `nil -> nint` and failed with `GS9998: NotSupportedException`. Selecting the non-nil operand's actual pointer type preserves the existing `nil -> pointer -> nint` lowering for both pointer symbols.

### Runtime comparison matrix

Direct `gsc` compile: rc 0. `dotnet exec`: rc 0. Runtime answers:

| Shape | `x == nil` | `nil == x` | `x != nil` | `nil != x` |
|---|---:|---:|---:|---:|
| managed nil `*func` | True | True | False | False |
| managed `&identity` | False | False | True | True |
| Cdecl nil fnptr | True | True | False | False |

Function-pointer pairs:

| Shape | Result |
|---|---:|
| managed `&identity == &identity` | True |
| managed `&identity != &other` | True |
| two nil Cdecl pointers `==` | True |
| two nil Cdecl pointers `!=` | False |

Value sources:

| Source | Managed | Cdecl |
|---|---:|---:|
| field `== nil` | True | True |
| nil array element `== nil` | True | True |
| non-nil managed array element `!= nil` | True | — |
| nil Cdecl array element `!= nil` | — | False |
| returned pointer `== nil` | True | True |

The test asserts all 24 runtime lines, including nil → `True`, `&identity` → `False`, and inverse `!=` answers. No compile-only oracle and no `ilverify` use.

`*T` already supports relational comparison. A separate runtime symmetry probe compared each pointer with itself using `<`, `<=`, `>`, `>=`:

- managed function pointer: `False, True, False, True`
- Cdecl function pointer: `False, True, False, True`
- `*int32` control: `False, True, False, True`

Thus widening the shared gate leaves no comparison asymmetry.

### RED proof

Reverted only `ExpressionBinder.Operators.cs` to the parent version, rebuilt without `--no-build`, and reran the new runtime pin:

- fixed: 1 passed, rc 0, Core md5 `da1a5a483501a688f5824368fea6a75e`
- mutant: 0 passed / 1 failed, rc 1, Core md5 `e99288be290d63292b735ef384265a21`
- restored: 1 passed, rc 0, Core md5 `da1a5a483501a688f5824368fea6a75e`

Mutant diagnostics were `GS0129` on all requested rows, including both operand orders/operators for managed and Cdecl function pointers. After restoration, both `git diff HEAD` and `git diff --cached HEAD` were empty.

### Interaction with #3275

No #3275 file is modified. Before #3275, the widened binder reached the pre-existing `nil -> FunctionPointerTypeSymbol` emitter gap and the runtime pin failed with `GS9998`. Merging #3275's exact head made the pin pass; after its merge as `2d06ccc780e89e1044a768434f37b1ec3b774c85`, the branch was rebased and all evidence rerun. #3285's `EmitDefault` path remains untouched.

### Build and suites

Command:

```text
dotnet build GSharp.sln -c Release -m:1 -p:ContinuousIntegrationBuild=true -p:TreatWarningsAsErrors=true --nologo
```

Result: rc 0, **0 warnings, 0 errors**. `GSharp.Core.dll`: 2,489,856 bytes, md5 `da1a5a483501a688f5824368fea6a75e`, identical across all 20 Release consumers.

Nine rebuilt Release test projects, no `--no-build`, no “No test matches”:

- Core: 7,210 passed, 1 skipped
- Compiler: 4,274 passed
- Interpreter: 1,360 passed, 3 skipped
- LanguageServer: 468 passed
- SDK: 64 passed
- Extensions: 104 passed
- Cs2Gs: 1,907 passed
- GeneratorHost: 31 passed
- InternalAnalyzers: 9 passed
- **Total: 15,427 passed, 4 skipped, 0 failed; 9/9 projects**

Issue #3204, opened 2026-08-04
Evaluator: plain struct without a ToString override renders record-style Name(Field=...); emitted execution prints the CLR type name


DavidObando:
## Repro

```gsharp
package Issue2896.Controls
import System

data struct DataValue {
    var Number int32
}

struct DefaultValue {
    var Number int32
}

let dataValue = DataValue{Number: 7}
Console.WriteLine(dataValue.ToString())

let defaultValue = DefaultValue{Number: 7}
Console.WriteLine(defaultValue.ToString())
```

## Actual

At `main` `515ad40d`, `Compilation.Evaluate` prints:

```text
DataValue(Number=7)
DefaultValue(Number=7)
```

Emitted execution (in-memory host; same IL as `gsc /out:`) prints:

```text
DataValue(Number=7)
Issue2896.Controls.DefaultValue
```

`data struct` synthesizes a real `ToString` override in IL, so it agrees on both engines. A **plain** struct has no override in IL and falls to `System.ValueType.ToString` (the type name) — but the evaluator gives it the same record-style rendering as a `data` struct. Exit code 0, no diagnostic, different output.

## Expected

Plain structs match compiled semantics (CLR type name); the record-style rendering is the `data struct` opt-in. The evaluator's default-struct rendering diverges and retires with it (Phase 3c), or is fixed if any pre-3c release still ships the evaluator path.

## Scope

Found by the ADR-0156 Phase 3b.0 emitted-oracle pilot migration (#3176): `Issue2896StructObjectOverrideTests.DataAndDefaultStructBehavior_RemainsUnchanged` stays on `Compilation.Evaluate` with a reference to this issue; the rest of the file (override dispatch, boxed dispatch, cross-driver theories) migrated cleanly.

DavidObando:
We will retire the evaluator's default-struct rendering, diverging from old behavior. This is reasonable and there's no need to do soft deprecation.

DavidObando:
Follow-up design question filed as #3208 (should G# synthesize a default pretty-printing `ToString()` for structs and classes, transparently overridable). This issue's resolution stands as decided above: emitted CLR semantics become the behavior, no soft deprecation; evaluator-pinned rendering tests retire with the evaluator in Phase 3c.

DavidObando:
[...]

PR #3299, opened 2026-08-06
Adopt ADR-0157: display-side REPL value formatting (accepted), with ADR + feasibility spike

### LINKED ISSUE #3208: Consider a G# default ToString() that pretty-prints state for structs and classes, transparently overridable
Follow-up to #3204, where the decision was to retire the evaluator's record-style default-struct rendering (\`Name(Field=...)\`) in favor of CLR semantics (type name) with no soft deprecation.

That closes the *divergence*, but raises the general language question: should G# synthesize a default \`ToString()\` that pretty-prints state for user-defined types — both structs and classes — when the user hasn't declared one, with transparent user override (declaring \`ToString()\` simply replaces the synthesized one, as C# records do)?

Design questions to settle:
- **Opt-out vs opt-in vs always-on**: synthesized for every user type (records-style), only for types without any \`ToString\`, or behind a modifier/attribute?
- **Synthesis site**: emitted as a real \`ToString\` override (so interop/BCL callers see it too — \`string.Format\`, interpolation, debugger), not a REPL-display-only formatter. The REPL echo could alternatively be solved display-side without touching \`ToString\` semantics — that is the competing smaller design.
- **Rendering contract**: field order, nesting depth, reference cycles, collections, nil, accessibility (public fields only vs all), and stability of the format (is it a spec guarantee or explicitly unspecified "for diagnostics"?).
- **Inheritance**: does a derived class inherit the base's synthesized rendering, re-synthesize including inherited fields, or defer?
- **Interop boundary**: imported CLR types are unaffected; G# types consumed from C# would observably expose the synthesized override — is that desirable?
- **Cost**: metadata size per type; interaction with trimming/AOT if ever relevant.

Part of the ADR-0156 Phase 3 semantic-alignment follow-ups (#3176 / #3163).

DavidObando:
Closes #3208. Part of #3163. **ADR-0157 is Accepted (2026-08-06)** — this PR delivers the ADR, its feasibility spike, and the accepted implementation.

## Summary

**Decision: no default emitted `ToString` synthesis for plain structs and classes. Emitted CLR semantics stay intact (#3204's decision); the REPL echo gap is closed display-side** with `src/Repl/Engine/ReplValueFormatter.cs` — reflection-based structural rendering in G# composite-literal shape, applied only when the runtime type has no `ToString` override below `object`/`ValueType`, deferring transparently to `data`-synthesized (ADR-0029) and user `override func ToString` (#2896) overrides. The format is explicitly diagnostics-only, never spec.

- Add **ADR-0157** (`docs/adr/0157-default-tostring-synthesis.md`, Accepted): answers every question filed in #3208 — opt-in/opt-out/always-on (no new synthesis; `data` and `override func ToString` already occupy the opt-in space), synthesis site (display-side, not emitted), rendering contract (nil, quoted strings, public members in metadata order, depth cap, reference-cycle elision, capped collections; tool-level and revisable), inheritance (no emitted slot; formatter reflects the runtime type; data hierarchies keep #2338 behavior), interop (unchanged by construction), cost (zero metadata; ~5 µs per echo). Rejected alternatives documented with grounded reasons (always-on emitted synthesis, opt-out, a new ToString-only opt-in, assembly-scoped triggering, status quo).
- Add the feasibility spike `test/Interpreter.Tests/Adr0157PrettyDisplaySpikeTests.cs` (trait `Category=Adr0157Spike`, ADR evidence, untouched by the implementation per the ADR-0156 precedent).
- Implement **`ReplValueFormatter`** and wire the three echo sites the ADR identified: transcript echo (`ReplScreen.cs`), compat console echo (`Compat/GSharpRepl.cs`), and the state-sidebar values column (`EmittedSessionEngine.Snapshot`). `Cell.Value` stays the raw runtime object — display changes only.
- Add behavioral pins `test/Interpreter.Tests/Adr0157ReplValueFormatterTests.cs` (10 tests): echo before/after (`gsi$N.Point` → `Point{X: 1, Y: 2}`), data-struct (`Point(X=1, Y=2)`) and user-override (`tag:11`) echoes unchanged, nested/nil (`Node{Name: "root", Next: Node{Name: "leaf", Next: nil}}`), cycle elision, collection cap (`[1, 2, 3, 4, 5, 6, 7, 8, ...]`), sidebar rendering, depth cap, raw-`Cell.Value` pin.

## Spike evidence

| Case (real emitted values via `EmittedOracle`) | Result |
|---|---|
| Plain two-field struct | pins today's contract (`ToString` slot = `ValueType`, echo = CLR type name); formatter renders `Point{X: 1, Y: 2}` |
| `data struct` | emitted override confirmed on the type (interop-visible); formatter defers: `Point(X=1, Y=2)` |
| `override func ToString` on plain struct | formatter defers: `tag:11` |
| Nested class + `nil` / reference cycle / capped `[]int32` | contract shapes all hold |

Measurements: formatter steady-state **5.1 µs/call**; metadata cost of the rejected always-on alternative bounded by an identical program compiled plain vs `data`: **2,560 → 3,584 bytes PE (+1,024 bytes** for the 7 synthesized members; ToString is 1 of 7).

## Verification

- Release solution build: **0 warnings, 0 errors**.
- Witness (ADR-0154, reverted-hunk mutant): with the three echo-site wiring hunks reverted (formatter class kept), `Adr0157ReplValueFormatterTests` ran **5 failed / 5 passed** — exactly the wiring-dependent tests red (plain-struct echo, nested/nil, cycle, collections, sidebar; the pre-change echo shows the CLR type name) and the wiring-independent pins green. Wiring restored: **10/10 passed**.
- Spike: `--filter "FullyQualifiedName~Adr0157"` — spike 7/7 + behavioral 10/10, Debug and Release.
- Full `test/Interpreter.Tests` (the REPL suite): first run 1,413/1,414 — the single failure was the intended echo change itself (a pre-existing pin asserting a string-valued cell echoes unquoted, `ParenthesizedReceiverEmittedSessionTests`); pin updated to the quoted echo, targeted rerun of that suite + all Adr0157 tests **21/21 green**.
- NOT RUN: Core.Tests / Compiler.Tests / LanguageServer.Tests — no `src/Core`, compiler, or LSP code touched; the product change is confined to `src/Repl` (three files + the new formatter).

## Checklist

- [x] Behavioral tests carry a witness of discrimination (ADR-0154): reverted-hunk witness recorded above (5 red pre-wiring → 10 green wired).
- [x] Self-review verdict recorded when one was run (MERGEABLE / BLOCKER / SHOULD-FIX / NIT), with residuals filed as issues rather than dropped.

Self-review verdict: **MERGEABLE** — ADR (Accepted) + additive spike + a display-only REPL change with red→green witnessed pins; emitted semantics and `Cell.Value` pinned unchanged. Residual: the trigger-scoping knob is documented in the ADR as tunable follow-up, not blocking.

### DIFF
--- a/docs/adr/0157-default-tostring-synthesis.md
+++ b/docs/adr/0157-default-tostring-synthesis.md
@@ -1,7 +1,8 @@
 # ADR-0157: Default pretty-printing for user types — display-side, not synthesized
 
-- **Status**: Proposed — owner decision requested
-  ([#3208](https://github.com/DavidObando/gsharp/issues/3208))
+- **Status**: Accepted — 2026-08-06
+  ([#3208](https://github.com/DavidObando/gsharp/issues/3208); implementation
+  of the display-side `ReplValueFormatter` follows in a separate PR)
 - **Date**: 2026-08-06
 - **Phase**: Language surface / REPL ergonomics (ADR-0156 Phase 3
   semantic-alignment follow-up)
--- a/docs/adr/0157-default-tostring-synthesis.md
+++ b/docs/adr/0157-default-tostring-synthesis.md
@@ -1,0 +1,269 @@
+# ADR-0157: Default pretty-printing for user types — display-side, not synthesized
+
+- **Status**: Proposed — owner decision requested
+  ([#3208](https://github.com/DavidObando/gsharp/issues/3208))
+- **Date**: 2026-08-06
+- **Phase**: Language surface / REPL ergonomics (ADR-0156 Phase 3
+  semantic-alignment follow-up)
+- **Related**: [#3208](https://github.com/DavidObando/gsharp/issues/3208)
+  (this question), [#3204](https://github.com/DavidObando/gsharp/issues/3204)
+  (decided: plain types keep CLR `ToString` semantics, no soft deprecation),
+  [#3163](https://github.com/DavidObando/gsharp/issues/3163) /
+  [#3176](https://github.com/DavidObando/gsharp/issues/3176) (campaign
+  tracking), ADR-0029 (`data` struct synthesized members — the existing
+  opt-in), ADR-0025 (`record` keyword alias), ADR-0032 (`data` ergonomics),
+  ADR-0156 (one emitted semantics everywhere), ADR-0034 (imported CLR
+  interop); precedent issues
+  [#2896](https://github.com/DavidObando/gsharp/issues/2896) (user
+  `override func ToString` on plain structs),
+  [#2361](https://github.com/DavidObando/gsharp/issues/2361) (transparent
+  user takeover of the synthesized data ToString),
+  [#2338](https://github.com/DavidObando/gsharp/issues/2338) (data-class
+  inheritance re-override), [#2363](https://github.com/DavidObando/gsharp/issues/2363)
+  (zero-field data types)
+
+## Context
+
+Issue [#3204](https://github.com/DavidObando/gsharp/issues/3204) closed the
+last rendering divergence of the evaluator era: the retired tree-walking
+evaluator gave every plain struct a record-style `Name(Field=...)` rendering
+that the emitted program never had. The owner's decision was to keep **CLR
+semantics** — a plain struct or class without a `ToString` override prints its
+CLR type name — with no soft deprecation. That decision stands; the follow-up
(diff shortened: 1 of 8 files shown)


DavidObando:
ADR is accepted.

Issue #3301, opened 2026-08-06
Mutating method call on a map element crashes the emitter (GS9998 NullReferenceException)


DavidObando:
Found while pinning the issue #3292 map guards (PR for #3292).

## Repro

```
package MapBump
import System
struct P {
    var X int
    func Bump() { this.X = this.X + 1 }
}
var m = map[int, P]{1: P{}}
m[1].Bump()
```

```
$ gsc MapBump.gs /out:MapBump.dll
MapBump.gs(8,1,8,12): error GS9998: NullReferenceException: Object reference not set to an instance of an object.
    m[1].Bump()
```

## Notes

- Reproduces identically on main before the #3292 change (verified against a pre-#3292 build) — the element-address work did not introduce or alter it; map elements are excluded from the `ldelema` path by design.
- The crash is an emitter NRE surfaced as GS9998, not a clean diagnostic. Expected behavior is presumably either the pre-existing "operates on a copy" method-receiver semantics (like other struct rvalue receivers) or a clean rejection — either way, not an internal error.
- Likely area: the receiver spill path for a value-typed map element receiver of an instance call (`SlotPlanner.ReceiverSpillCollector` / `MethodBodyEmitter.EmitInstanceReceiver` fallback around `EmitMapIndexRead`).
- Pinned (as "does not silently succeed") by `Issue3252IndexedStructMemberWriteTests.MutatingMethodOnMapElementDoesNotSilentlySucceed`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2300, opened 2026-07-09
[gsc] '== nil' / '!= nil' not defined for interface and type-parameter operands (GS0129)


DavidObando:
### Summary
`x == nil` / `x != nil` is reported as GS0129 "Binary operator '==/!=' is not defined" when `x` is of an **interface type** or an **unconstrained/class type-parameter** type. A null check against these reference-typed operands must always be valid (reference equality with null), as in C#/Kotlin.

### Evidence (Oahu.Core, fresh main @ 2503625f)
- `GS0129 Binary operator '!=' is not defined for types 'IProfile' and 'nil'`
- `GS0129 Binary operator '==' is not defined for types 'TPerson' and 'nil'` (TPerson is a type parameter)

### Fix direction
gsc binary-operator binding (ExpressionBinder operators / BoundBinaryOperator) must allow `==`/`!=` when one operand is `nil` and the other is any reference type — including interface types and type parameters (whose runtime is a reference or is boxed) — lowering to reference/null equality. Mirror how it already handles `==`/`!=` for class types and `nil`. Generalize to any interface/type-parameter operand, either side.

### Impact
Common idiomatic pattern across Oahu; multiple instances. Clear general gsc gap, low-risk fix.

Issue #2182, opened 2026-07-06
gsc: overload/constructor resolution fails for an array-of-type-parameter argument []T converting to array base (GS0159)


DavidObando:
## Summary
When a method/constructor argument is an **array of a type parameter** `[]T`, and the target parameter is an array base type (`System.Array`, `ICollection`, `IList`, `IEnumerable`), overload resolution fails and reports **GS0159 'Cannot find function ...'**. The exact same call site works when the argument is a CONCRETE array (`[]int32`) or the base type directly (`System.Array`). The `[]T -> System.Array` conversion itself is recognized in a plain return position — only overload/constructor resolution rejects it.

## Repro (G#)
```gsharp
package R
import System
import System.ComponentModel
class C[T Enum struct] {
    func WithTypeParamArr(a []T) TypeConverter.StandardValuesCollection {
        return TypeConverter.StandardValuesCollection(a)   // error GS0159: Cannot find function StandardValuesCollection
    }
    func WithConcreteArr(a []int32) TypeConverter.StandardValuesCollection {
        return TypeConverter.StandardValuesCollection(a)   // CLEAN
    }
    func WithArrayBase(a System.Array) TypeConverter.StandardValuesCollection {
        return TypeConverter.StandardValuesCollection(a)   // CLEAN
    }
}
```
And the plain conversion is fine on its own:
```gsharp
class D[T] { func F(a []T) System.Array { return a } }   // CLEAN
```

## Diagnosis
The conversion classifier accepts `[]T -> System.Array`/`IEnumerable` in a direct (assignment/return) context, but the overload-resolution applicability/conversion check does not treat an array of a type parameter as convertible to the array base interfaces/`System.Array`. When no candidate is deemed applicable, the constructor-as-call lookup falls through to GS0159 rather than binding the (valid) `StandardValuesCollection(Array)`/`(ICollection)` overload.

## Fix direction
Make overload-resolution argument-applicability use the same conversion classification that already accepts `[]T -> System.Array`/`ICollection`/`IList`/`IEnumerable` (array covariance / base-type of a type-parameter array). GENERALIZE for any pointee/element type parameter, any array-base target, and both method calls and constructor calls; do not special-case StandardValuesCollection.

## Impact
[...]

Issue #1229, opened 2026-06-27
User class instance cannot implicitly convert to System.Object (object/object?) — GS0155


DavidObando:
## Summary
An instance of a user-defined `class` cannot be implicitly converted to `System.Object` (`object`) or `object?`, failing with `GS0155: Cannot convert type 'D' to 'object'`. Every reference type derives from `System.Object`, so this implicit reference (upcast) conversion must always succeed. The bug blocks passing a user object to an `object`/`object?` parameter, returning one, assigning one to an `object` variable, and adding one to an `object` collection.

Notably, user-class → **user base class** upcast already works — only the conversion to the imported `System.Object` is missing. The conversion logic does not treat every user class as deriving from `System.Object`.

## Repro
```gsharp
package p
class D {}
class C {
  func TakesObj(o object) {}
  func TakesObjN(o object?) {}
  func F(d D) {
    TakesObj(d)            // GS0155: Cannot convert type 'D' to 'object'
    TakesObjN(d)           // GS0155: Cannot convert type 'D' to 'object?'
    let o object = d       // GS0155
    let n object? = d      // GS0155
  }
  func G(d D) object? { return d }   // GS0155
}

open class Base {}
class Sub : Base {}
class E {
  func H(s Sub) { let b Base = s }   // OK — user-to-user-base upcast works
}
```

## Expected
A user `class` instance converts implicitly to `object` / `object?` (and to any of its base interfaces), exactly as a user-to-user-base upcast does today. The emitted IL is a plain reference upcast (no box; classes are reference types).

## Impact (Oahu)
Pervasive in `Oahu.Decrypt`. Examples: raising an event with `this` as the `object?` sender (`Ev?(this, args)` and the `EventHandler`-style `(object?, T)` signature), adding user objects to `List[object]`/`object[]`, and any API taking `object`. Several `GS0155` "Cannot convert ... to object" errors trace here. It also blocks a translator follow-up to rewrite `event?.Invoke(this, e)` into the native `event?(this, e)` form, because the `this` sender argument fails to convert to `object?`.

## Notes
- User class → user base class already works → the fix is localized to the implicit reference conversion classification recognizing `System.Object` (and likely `System.ValueType`/interfaces by the same token) as a base of every user class.
- Distinct from struct→object (boxing), which is a separate value-type conversion.

PR #3309, opened 2026-08-07
Fix #3303: generic map[K,V] fields — nil-compare binding, open-map object widening, and the pinned #3306 field matrix

### LINKED ISSUE #3303: [emit] Generic class with a map[K,V] field over type parameters: field stays nil at runtime (NRE); map[K,V] != nil not defined
Found by the ADR-0158 feasibility spike for #3209 (the spike sidestepped it by backing the generic prototype with `ConcurrentDictionary[K, V]` interop instead).

## Repro (emitted execution, e.g. EmittedOracle)

```gsharp
class G[K, V any] {
    var items map[K, V]
    init() { items = map[K, V]{} }
    func S(k K, v V) { items[k] = v }
    func L(k K) V { return items[k] }
}

func run() int32 {
    var g = G[string, int32]()
    g.S("a", 42)
    return g.L("a")
}

run()
```

## Actual

- Compiles with no diagnostics, then throws `NullReferenceException` at runtime — the `map[K, V]{}` literal assigned in `init()` never reaches the field, so `items` is still `nil` when `S`/`L` touch it. (A control probe shows generic `init()` bodies do run: an `int32` field assigned in the same position reads back correctly. The monomorphic equivalent — `map[string, int32]` field in a non-generic class — works end-to-end.)
- Probing the field with `items != nil` is itself rejected: `Binary operator '!=' is not defined for types 'map[K,V]' and 'nil'.`

## Expected

Either the generic-map-field shape works (literal construction and index access over substituted `K`/`V`), or the unsupported shape is rejected at compile time — silent `nil` + NRE is the worst of both. The `!= nil` comparison on a map-typed operand should also bind (maps are reference-backed `Dictionary<K,V>`).

Part of the #3209 / ADR-0158 spike fallout; see `test/Interpreter.Tests/Adr0158SyncMapSpikeTests.cs` (SyncMapGenericShape_CompilesAndRuns documents the working ConcurrentDictionary-backed alternative).

DavidObando:
Fixes #3303. Part of #3163.

## Summary

- **Sub-bug 1 (generic `map[K, V]` field NRE): already fixed by #3306, now pinned from the issue's own repros.** The issue was filed ~90 minutes before #3306 merged; its root cause was the same #3301 family (map index-read/assign/len/delete emit sites reflecting over the null `MapTypeSymbol.ClrType`, which is null *by construction* for type-parameter K/V). On current main the exact repro compiles, runs, and returns 42 under `gsc`, `gsi`, EmittedOracle, and the session engine. This PR adds the missing regression pins across the whole matrix: exact repro, len/delete through the field, value-type / user-struct / nested-generic type arguments, struct-key maps, generic **struct** containers, MVAR generic-method locals, monomorphic control, same-cell + cross-cell REPL.
- **Sub-bug 2 (`m != nil` / `m == nil` fails to bind, GS0129): fixed.** `MapTypeSymbol` was missing from `IsNullCompare`'s reference-shaped family (#796 functions/sequences, #2300 interfaces, #2354 classes) even though a map is unconditionally a reference-backed `Dictionary<K, V>` and a declared-but-unassigned map field genuinely IS nil (this very issue's failure mode). **Semantics chosen: comparison-only, mirroring function types** — nil *assignment* into a bare map slot still requires the explicit `map[K, V]?` spelling (ADR-0104), pinned by test. Slices and channels do NOT bind nil-compare on main; they deliberately keep that rejection (pinned) — widening them is a separate language-surface decision.
- **ClrType-consumer sweep (completing #3306's consolidation):** two more open-map binder gaps found by the sweep and fixed — `map[K,V] → object` widening (GS0155) and open-map arguments in CLR overload resolution (GS0159 on `Console.WriteLine(items)`), both of which bound fine monomorphically.
- No emitter changes needed: the equality `ldnull; ceq` tail and the `map → object` no-op reference conversion (`IsReferenceCompatible`'s #1481 arm) were already verifier-clean for the open `Dictionary<!0, !1>` shape. No new caches (GSA0004/RemapScope not implicated).

## ClrType-consumer sweep table

| Site | Consumer | Handled how |
|---|---|---|
| `MethodBodyEmitter.EmitMapLiteral` | `literal.MapType.ClrType` | symbolic `Dictionary`2` ctor/set_Item MemberRefs since #1481 — pinned |
| `MethodBodyEmitter.EmitLen` (map arm) | `mapType.ClrType` | symbolic `get_Count` since #3306 — pinned |
| `MethodBodyEmitter.EmitMapDelete` | `mapType.ClrType` | symbolic `Remove` since #3306 — pinned |
| `MethodBodyEmitter.EmitMapIndexRead` | `mapType.ClrType` | symbolic `TryGetValue` since #3306 — pinned |
| `MethodBodyEmitter.EmitMapIndexAssignment` | `mapType.ClrType` | symbolic `set_Item` since #3306 — pinned |
| `SignatureEncoder.EncodeTypeSymbol` (field/local/param/state-machine sigs) | open-map arm | symbolic `GENERICINST Dictionary`2<K,V>` since #1481 — unchanged |
| `MethodBodyEmitter.IsReferenceCompatible` (map → object emit) | null-safe #1481 arm | unchanged; now actually reachable from the binder |
| `BoundBinaryOperator.IsNullCompare` | *(missing arm)* | **fixed here** — map joins the #796 reference-shaped nil-compare family |
| `Conversion.Classify` object widening | required `from.ClrType != null` | **fixed here** — reference-shaped map arm (mirrors #1229/#1421) |
| `MemberLookup.TryProjectErasedClrType` | no map case (#833 projection) | **fixed here** — erase K/V, re-form closed `Dictionary<,>` (mirrors slice/sequence cases) |
| `ExpressionBinder.GetEffectiveArgumentClrTypeForOverloadResolution` | no map arm (#2182 family) | **fixed here** — matching map arm |
| CLR interop **member lookup** on an open-map receiver (`items.ContainsKey/.Keys/.GetEnumerator`) | `receiver.Type.ClrType` reflection | **residual, clean diagnostic** (GS0159 "Cannot find function …", no NRE): receiver-side member reflection needs a closed CLR type; rerouting through the imported-generic machinery (which is how `ConcurrentDictionary[K, V]` works today) is follow-up-sized. Monomorphic maps unaffected. |
| `for k, v in range m` / containment | n/a | not a ClrType consumer: map range/`in` are unsupported for ALL maps today (GS0116), generic and monomorphic alike |

No `NullReferenceException` remains on any consumer for the null-ClrType map shapes.

## Verification

- Release solution build: 0 warnings, 0 errors (includes the bootstrap compile of `Gsharp.Extensions`' Sync.gs through the changed binder paths).
- Witnesses (ADR-0154), fix hunks stashed → rebuilt: **Core.Tests 10/22 red** (nil-compare witnesses with GS0129, object-widening with GS0155, overload-resolution with GS0159 — exact issue signatures), **Interpreter.Tests 6/7 red**; restored → **22/22 and 7/7 green**. Green-both-sides arms are the documented #3306 regression pins (sub-bug 1 was already fixed on main) plus the deliberate slice/channel/`map?`/no-bare-nil-assignment guards.
- Core.Tests (full): 7261 passed, 0 failed, 1 skipped (the always-skipped `RegenerateBaseline`).
- Interpreter.Tests (full): 1454 passed, 0 failed.
- Compiler.Tests LanguageConformance filter: 127/127 passed.
- Extensions.Tests (full, SyncMap surface): 121/121 passed.
- ILVerify: 4 CLI repro binaries built with the final compiler (full generic-field matrix incl. struct keys/values and nested generics; nil-compare both operators/orders; unassigned-field nil; `Console.WriteLine(items)`/object widening) — "All Classes and Methods Verified" for each. Driver parity: `gsi` prints identical results for all repro programs.
- NOT RUN: full Compiler.Tests outside the LanguageConformance filter, LanguageServer/Sdk/InternalAnalyzers suites, Cs2Gs.Tests (known non-deterministic host crash per repo memory), e2etests. Rationale: the change is four narrow binder arms in Core; Core.Tests + Interpreter.Tests + the conformance gate + Extensions.Tests exercise the shared binding/emit paths across drivers.

## Checklist

- [x] Behavioral tests carry a witness of discrimination (ADR-0154): stash-revert run recorded above; each nil-compare/sweep witness red pre-fix with the exact diagnostic id it pins, all green post-fix. #3306-matrix arms are explicit regression pins (documented as green-both-sides).
- [x] Self-review verdict: MERGEABLE — three additive binder arms mirroring long-established per-shape patterns (each new arm sits beside its #796/#1229/#833/#2182 sibling), comparison-only nil semantics inherited from the function-type precedent rather than invented, and the single residual (open-map interop member lookup, clean GS0159) documented above as follow-up-sized rather than half-patched here.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/BoundBinaryOperator.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundBinaryOperator.cs
@@ -522,6 +522,26 @@ public sealed record BoundBinaryOperator
             return true;
         }
 
+        // Issue #3303: a `map[K, V]` is unconditionally a CLR reference type
+        // (`System.Collections.Generic.Dictionary<K, V>`), regardless of
+        // whether its `ClrType` is materialized (it is null when K or V
+        // structurally references an in-scope type parameter or a
+        // same-compilation struct). A bare map slot can genuinely hold a CLR
+        // null — a declared-but-unassigned `var items map[K, V]` field is nil
+        // until `init()` runs — so `m == nil` / `m != nil` must bind exactly
+        // like the reference-shaped structural types above (#796). This is
+        // comparison-only, mirroring function types: bare-slot nil ASSIGNMENT
+        // still requires the explicit `map[K, V]?` spelling (ADR-0104), which
+        // binds through the NullableTypeSymbol arm at the top. The emitter's
+        // generic `ldnull; ceq` tail is verifier-clean for the reference-typed
+        // operand, including the open `Dictionary<!0, !1>` shape. Slices and
+        // channels intentionally keep their current GS0129 rejection —
+        // widening those is a separate language-surface decision.
+        if (nullableOrUnderlying is MapTypeSymbol)
+        {
+            return true;
+        }
+
         // Issue #2300: interface-typed operands are always CLR reference
         // types (a G#-declared InterfaceSymbol or an imported interface),
         // so `iface == nil` / `iface != nil` must bind exactly like the
--- a/src/Core/CodeAnalysis/Binding/Conversion.cs
+++ b/src/Core/CodeAnalysis/Binding/Conversion.cs
@@ -817,6 +817,23 @@ public sealed class Conversion
             return Conversion.Implicit;
         }
 
+        // Issue #3303: a `map[K, V]` is unconditionally a CLR reference type
+        // (`System.Collections.Generic.Dictionary<K, V>`), but its `ClrType`
+        // is null when the key or value structurally references an in-scope
+        // type parameter (e.g. a `map[K, V]` field of a generic class) or a
+        // same-compilation struct, so the general CLR-backed widening rule
+        // above cannot fire. Widening such a reference to `object` is still a
+        // no-op at the IL level — the emitter's #1481 arm in
+        // MethodBodyEmitter.IsReferenceCompatible already recognizes exactly
+        // this shape. Without this, `Console.WriteLine(items)` and `object`-typed
+        // parameter/return positions reported GS0155/GS0159 for open maps
+        // while the monomorphic equivalent bound fine.
+        if (from is MapTypeSymbol
+            && (to == TypeSymbol.Object || to?.ClrType.IsSameAs(typeof(object)) == true))
+        {
+            return Conversion.Implicit;
+        }
+
         // Boxing conversion for user value types to System.Object.
         if (from is StructSymbol fromStruct && !fromStruct.IsClass && to?.ClrType.IsSameAs(typeof(object)) == true)
         {
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.cs
@@ -1525,6 +1525,31 @@ internal sealed partial class ExpressionBinder
             }
         }
 
+        // Issue #3303: a `map[K, V]` whose key or value has no CLR backing (a
+        // generic type parameter or a same-compilation user type) has a null
+        // `ClrType`, so `GetEffectiveArgumentClrType` returned null above. Its
+        // runtime backing is still a `Dictionary<,>` reference whose key/value
+        // ride through to their erasure (type parameter / user type →
+        // `object`, enum → `int`), exactly like the slice arm above — so
+        // surface the erased closed `Dictionary<…>` shape. Without this the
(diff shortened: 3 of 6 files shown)


DavidObando:
Residual tracked: the open-generic map interop member-lookup gap documented in this PR is now filed as a dedicated issue (see the reference above). The #3310 nil-comparison extension (all magic types) will also flip this PR's deliberate slice/channel rejection pins when it lands.

PR #3322, opened 2026-08-07
Nullable channel grammar via parenthesized type clauses + statically-constant nil-compare warning

### LINKED ISSUE #3315: Nullable channel type spelling is unparseable — the grammar's outer '?' on 'chan T' is unreachable
Found during ADR-0159 implementation (PR #3314): there is no way to spell a nullable channel type. \`chan int32?\` parses with the \`?\` binding to the **element** type (\`chan (int32?)\`), and the spec grammar's outer \`?\` position for the channel type clause is unreachable — the parser never produces \`(chan int32)?\`.

**Why it matters now:** GS0520 (ADR-0159) requires channel slots to be explicitly initialized because channels are carved out of the empty-instance zero-value rule. The intended escape hatch for "genuinely optional channel field" is the \`?\` spelling — which currently cannot be written. Until this is fixed, fields that want an optional channel have no spelling at all.

**Expected:** a defined, parseable spelling for nullable channels — either make the trailing \`?\` after \`chan T\` bind to the channel (with a parenthesized or suffix form for nullable elements, e.g. \`chan (T?)\` vs \`(chan T)?\`), or introduce an unambiguous shape. Precedence decision + parser fix + spec grammar alignment + binder/emit passthrough + tests for both readings.

Part of #3163 follow-ups; recorded in ADR-0159's follow-up list. Related: #3310/#3314, ADR-0104 (nullable suffix precedent for maps).

### LINKED ISSUE #3317: Warn on statically-false nil comparisons against bare (non-nullable) magic-type slots
Found during ADR-0159 implementation (PR #3314): with sound empty-instance zero values (ADR-0159), a bare \`map[K,V]\` / \`[]T\` / sequence slot can never hold nil — so \`m == nil\` on a bare slot is statically false (and \`m != nil\` statically true). The comparison now binds (per #3309/#3310's family extension, needed for \`?\`-typed slots and interop values), but on provably-non-nil operands it is dead code and likely a porting artifact (e.g. Go code migrated verbatim).

**Proposal:** a warning-severity diagnostic ("comparison of non-nullable '<type>' with nil is always false/true") on nil comparisons whose operand's static type is a bare (non-\`?\`) magic type. Precedent check: warning-severity diagnostics exist (e.g. GS9100 family used with /nowarn), so the mechanism is available; pick the next free code per house conventions and register it in the diagnostics reference + sync gate. Mind smart-cast interactions: a \`map[K,V]?\` narrowed to non-null by flow should probably NOT warn on a redundant re-check (or should, as a separate "redundant check" nuance) — decide and document.

Part of #3163 follow-ups; recorded in ADR-0159's follow-up list. Related: #3310/#3314, #3309.

DavidObando:
Fixes #3315. Fixes #3317. Part of #3163 (ADR-0159 addendum items).

## #3315 — nullable channel spelling

**Precedence decision**: a parenthesized single type clause `(T)` is a grouping production whose trailing `?` marks the *whole* inner type nullable — `(chan int32)?` is a nullable channel of int32, the GS0520 escape hatch for genuinely optional channel slots (previously unspellable: the grammar's outer `?` on `chan T` was unreachable behind the greedy element parse).

Consistency with the existing suffix family: the unparenthesized trailing `?` keeps binding to the *element* (`chan int32?` stays `chan (int32?)`), matching `[]T?` (element-nullable, #1212) and `(T) -> R?` (return-nullable, ADR-0075/ADR-0137) — each of those already has an explicit whole-type spelling too (`[]?T`, `((T) -> R)?`). The grouping production is general, so it also yields `([]T)?` ≡ `[]?T` and `(map[K,V])?` ≡ `map[K,V]?` for free.

- Parser: `ParseTupleTypeClause` returns the inner clause for single-element comma-free `(T)`; a trailing `?` is recorded via a new `TypeClauseSyntax.ParenthesizedQuestionToken`.
- Binder: `BindTypeClause` wraps the whole bound type in `NullableTypeSymbol` for the parenthesized `?` (collapsing redundant `(T?)?`); GS0520's `RequiresExplicitInitializer` already exempts `NullableTypeSymbol` slots, so no separate carve-out was needed.
- Spec: the unreachable `'chan' TypeClause '?'?` grammar production replaced with `'chan' TypeClause` plus the general `'(' TypeClause ')' '?'?` grouping form; ADR-0159 gains a status-compatible addendum note (this completes its stated intent rather than amending the decision).

## #3317 — statically-constant nil-compare warning

With ADR-0159's sound zero values, a `== nil`/`!= nil` comparison whose non-nil operand's **static** type is a bare (non-`?`) `map[K,V]`/`[]T`/`[N]T`/`chan T` is always false/true — typically a Go-porting artifact. New warning **GS0521** surfaces it; the comparison keeps binding and evaluating (ADR-0159's uniform rule is unchanged, this is diagnostics-only).

Design decisions:
- Static-type based on the **declared** type, both operand orders — a smart-cast-narrowed read of a `?`-declared slot does not warn (the "redundant re-check" nuance is left as a documented future refinement, not implemented here).
- `?`-typed operands, including interop values surfaced as `T?`, never match — they're the comparison's real use case.
- `sequence[T]` excluded from v1: its nil comparison predates ADR-0159 (#796) and bare sequence values commonly cross interop boundaries where the static-emptiness guarantee doesn't hold the same way.
- Accepted corners at warning (not error) severity: `= default` and the struct default-instance honesty clause can reintroduce nil.

Registered in both diagnostics references (sync gate green); `/nowarn:GS0521` and `/warnaserror` interaction pinned by driver tests. Repo sweep found no bare-slot nil comparisons in the SDK's `.gs` sources or samples that the new warning would need to fix — only function-typed and `?`-typed operands exist today (the #3303/#3310 witness tests compare bare slots deliberately and assert only on error-severity diagnostics, so they're unaffected).

## Verification (sequential, Release)

- Release build: 0 warnings / 0 errors (warnings-as-errors — GS0521 doesn't fire anywhere in the repo's own code after the sweep).
- Full Core.Tests: **7392/7393** (1 standing skip).
- Full Interpreter.Tests: **1467/1467**.
- Full LanguageConformance gate: **127/127**.

NOT RUN: Compiler.Tests outside the conformance gate, LanguageServer/Sdk/Extensions suites (no surface touched), Cs2Gs.Tests (known flaky host crash policy).


DavidObando:
**Rebase note.** Rebased onto latest main (which had independently merged an unrelated `PointerGenericTypeArgument` diagnostic at **GS0521**, from #3262/#3291's window).

**Conflict resolution:**
1. `GS0520`'s message text conflict (this PR's parenthesized-nullable-channel remedy vs. main's unrelated changes) — kept this PR's updated message.
2. **True collision**: this PR's `NilComparisonAlwaysConstant` warning also claimed GS0521, which now collides with the unrelated `PointerGenericTypeArgument` on main. Renumbered the nil-compare warning to **GS0523** (GS0522 is reserved by the still-open #3325, to avoid a repeat collision when that PR rebases next) — descriptor, both diagnostics docs, the ADR-0159 addendum, and all test assertions updated consistently.

**Re-verified post-rebase** (sequential, Release): `DiagnosticIdUniquenessTests` (sync gate) 4/4 · full Core.Tests 7412/7413 (1 standing skip) · full Interpreter.Tests 1492/1492 · full LanguageConformance gate 128/128. Clean build, 0 warnings/errors.

Issue #3324, opened 2026-08-07
len(s) NREs on a declared-uninitialized string local despite the documented empty-string zero value


DavidObando:
Found while implementing #3316 (local definite-assignment analysis). Verified as a pre-existing defect on main (67f953ac), independent of #3316's change — deliberately avoided in that PR's witness matrix rather than fixed there.

**Repro:**
\`\`\`gsharp
var s string
Console.WriteLine(len(s))
\`\`\`

**Actual:** NullReferenceException. String locals default to CLR null despite ADR-0008's documented empty-string (`""`) zero value for strings (a comment in the codebase asserts this contract, but the local-declaration path doesn't honor it).

**Expected:** `len(s)` returns 0, matching the documented zero-value contract (and matching how map/slice/array/sequence locals now behave soundly under ADR-0159).

Part of #3163 follow-ups. Related: #3316, #3310/#3314 (ADR-0159 zero-value soundness — this is the same class of bug for strings that #2262 was for maps).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #3323, opened 2026-08-07
Goroutine closure capturing a channel crashes emit (GS9998 'no local slot')


DavidObando:
Found while implementing #3316 (local definite-assignment analysis). Verified as a pre-existing defect on main (67f953ac), independent of #3316's change — deliberately avoided in that PR's witness matrix rather than fixed there.

**Repro:**
\`\`\`gsharp
var c = make(chan int32)
go func() { c <- 42 }()
\`\`\`

**Actual:** emit fails with GS9998 "no local slot" for the captured channel.

**Expected:** the goroutine closure captures `c` like any other captured local and compiles/runs.

Part of #3163 follow-ups. Related: #3316, #3308 (adjacent channel/closure work).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2262, opened 2026-07-08
Interpreter NRE


DavidObando:
I typed the following:
```gsharp
var num =42
```
```gsharp
func Fib(i int) long -> if i <= 0 { 0 } else if i == 1 { 1 } else { Fib(i-1)+Fib(i-2) }
```
```gsharp
var numbers map[int, long]
```
So far so good, then:
```gsharp
for var i = 1; i <= 20; i++ {
  numbers[i] = Fib(i)
}
```
The result was a `GS9999 Object reference not set to an instance of an object`

I then realized that I could have done:
```gsharp
var numbers map[int, long] = map[int, long]{}
```
or simply:
```gsharp
var numbers = map[int, long]{}
```
but not this:
```gsharp
var numbers map[int, long] = {}
```
Which results in `GS0005: Unexpected token <OpenBraceToken>, expected <IdentifierToken>`.

DavidObando:
Root cause identified and direction decided: this NRE is the magic-type zero-value unsoundness (bare `map[K,V]` zero value is a null reference despite the non-nullable spelling). Owner-approved fix (approach A) is scoped into #3310: empty-instance zero values for maps/slices/lists, channels requiring initialization or `?`. This issue closes when #3310's implementation lands.

PR #3325, opened 2026-08-07
Add definite-assignment analysis for no-zero-value locals (channel carve-out relaxation)

### LINKED ISSUE #3316: Definite-assignment analysis for locals (relax GS0520 for channel locals to use-site)
Found during ADR-0159 implementation (PR #3314): G# has no definite-assignment analysis for **locals** — the flow machinery exists for other purposes (the GS0100 definite-return CFG analysis, the \`RefKindDefiniteAssignmentAnalyzer\` for ref parameters), but a declared-uninitialized local's usage is not flow-checked.

**Concrete motivation:** ADR-0159's channel carve-out (GS0520) currently fires at the **declaration site** of an uninitialized \`chan T\` local, because without flow analysis the compiler can't prove assignment-before-use. With definite-assignment for locals, the natural rule becomes: declare freely, error only on a **use** that a path can reach without assignment (C#'s CS0165 model). This also generalizes: other declared-uninitialized shapes could rely on the same analysis rather than zero-value synthesis where a zero value is undesirable.

**Scope sketch:** local-variable definite-assignment over the existing CFG (basic blocks already built for GS0100), branch/loop/try-finally coverage (mind the #2890/#2891-era lessons about break/finally edges), goroutine/lambda capture interactions (a captured-before-assignment local), REPL cross-cell semantics (hoisted globals are fields — out of scope), and the GS0520 relaxation as the first consumer.

Part of #3163 follow-ups; recorded in ADR-0159's follow-up list. Related: #3310/#3314 (GS0520), #2890/#2891 (flow-edge precedents).

DavidObando:
Fixes #3316. Part of #3163.

## Semantics decision

Generalizes the ref-kind definite-assignment analyzer (renamed `RefKindDefiniteAssignmentAnalyzer` → `DefiniteAssignmentAnalyzer`) so its forward must-assign fixpoint also covers **locals whose types have no usable zero value** — today, `chan T`.

**Deliberately NOT full C# CS0165 semantics for all locals.** Locals with sound zero values (ints, maps/slices/arrays/sequences post-ADR-0159) keep their documented zero-value initialization and are **not** flow-checked — that's G#'s Kotlin/Go-synthesis position (zero-valued declarations are a documented language feature, not something to retrofit definite-assignment onto). The analysis applies *only* where no zero value exists. This resolves `var c chan T` moving from a declaration-site GS0520 error to legal declare-then-assign-then-use, matching Go's idiom, while globals and struct/class fields keep the declaration-site GS0520 (a top-level declaration is a static field readable from any function or REPL cell — per-function flow analysis structurally cannot police that).

## Flow-shape coverage table

Straight-line declare/assign/use (headline, red on main) · unassigned receive/send/argument uses · explicit `= default` opt-in (ADR-0159 honesty clause) · if/else joins and early returns · loops (assignment in a body that may run zero times does not count; in-body and cross-iteration uses do) · try/catch/finally (try-only assignment does not survive to catch; try+catch does; finally always does) · pattern switch with/without default · select (every arm assigns counts; an unassigned-case channel reports) · lambda/goroutine capture before vs. after assignment (C#'s capture-point model — a use inside a literal is checked against state *at the capture point*) · `out`-argument counting as assignment (reuses existing `ProcessCallArgument` ref/out handling) · zero-valued-kind pins (int/map/slice locals never flow-checked) · bare-global GS0520 pin · channel parameters seeded assigned.

`BoundAwaitForRangeStatement` gets a precise region analysis (stream always evaluates; body region analyzed; zero-iteration exit path discarded — exact, since assignments are never killed in this analysis).

## Diagnostic

New **GS0522** (renumbered from GS0521 — see below), C#'s CS0165 message model: fires at the *use* site reachable without a preceding assignment, not the declaration.

## Renumbering note

This branch and the parallel #3315/#3317 branch (PR #3322) were both worked in isolated worktrees off the same base commit and independently claimed **GS0521** as "the next free diagnostic code." #3322 was pushed first; this PR renumbers its own diagnostic to **GS0522** across the descriptor, both diagnostics docs, the ADR addendum, and tests. Verified via `DiagnosticIdUniquenessTests` (the sync gate) after the rename.

## Two pre-existing bugs found, filed not fixed

Both verified as pre-existing on main (67f953ac), independent of this change, and deliberately routed around in the witness matrix rather than fixed here:
- **#3323** — a goroutine closure capturing a channel crashes emit (`GS9998 "no local slot"`).
- **#3324** — `len(s)` NREs on a declared-uninitialized `string` local, despite ADR-0008's documented empty-string zero value for strings. Same bug class as #2262 (map zero-value unsoundness), for strings.

## ADR

ADR-0159 gains an addendum: follow-up (a) — the definite-assignment work — is recorded as shipped, with the general-local decision (zero-valued kinds exempt) stated explicitly; the nullable-channel follow-up (b) is noted as addressed by the parallel #3315 PR.

## Verification (sequential, Release)

- Release build: 0 warnings / 0 errors.
- `DiagnosticIdUniquenessTests` (the sync gate): 4/4 green post-renumbering.
- Full Core.Tests: **7392/7393** (1 standing skip).
- Full Interpreter.Tests: **1467/1467**.
- Full LanguageConformance gate: **127/127**.

NOT RUN: Compiler.Tests outside the conformance gate, LanguageServer/Sdk/Extensions suites (no surface touched), Cs2Gs.Tests (known flaky host crash policy).


DavidObando:
**Rebase note.** Rebased onto latest main, which had independently merged an unrelated `PointerGenericTypeArgument` diagnostic at **GS0521** — the same number this PR's own earlier renumbering commit (GS0521→GS0522, done pre-emptively to avoid colliding with #3322's nil-compare warning) already anticipated moving away from. Resolved by keeping both descriptors: `PointerGenericTypeArgument` stays GS0521 (main's, unrelated), `ChannelLocalUsedBeforeAssignment` is **GS0522** (this PR's, as already planned). The later "Renumber" commit in this branch's history became a no-op for the descriptor file specifically (its target state was already reached resolving the first conflict) but still applied cleanly to the docs/test files.

**Heads up on merge order**: this PR and #3322 both touch `GS0520`'s message text (the parenthesized-nullable-channel remedy) on adjacent lines. Whichever of the two merges second may need one more small conflict resolution on that single line when it's next rebased — not a semantic conflict, just adjacent-line churn.

**Re-verified post-rebase** (sequential, Release): `DiagnosticIdUniquenessTests` (sync gate) 4/4 · full Core.Tests 7412/7413 (1 standing skip) · full Interpreter.Tests 1492/1492 · full LanguageConformance gate 128/128. Clean build, 0 warnings/errors.

DavidObando:
**Rebase note (round 2).** #3322 merged, which as flagged in the previous rebase note collided with this PR on adjacent lines of \`GS0520\`'s doc description (both PRs touch the same "channel declaration" prose). Purely additive/textual — merged the two threads into one coherent description: the GS0520 row now covers both the \`(chan T)?\` optional-channel remedy (from #3322) and the local-vs-global/field distinction (from this PR, pointing at GS0522 below it), in both \`docs/diagnostics.md\` and its website mirror. No further diagnostic-numbering changes — GS0522 stands as settled.

**Re-verified post-rebase** (sequential, Release): \`DiagnosticIdUniquenessTests\` (sync gate) 4/4 · full Core.Tests 7438/7439 (1 standing skip) · full Interpreter.Tests 1492/1492 · full LanguageConformance gate 128/128. Clean build, 0 warnings/errors.

Issue #3319, opened 2026-08-07
Struct-typed locals don't zero-initialize magic-type (map/slice/array) fields


DavidObando:
Found while fixing CI fallout from ADR-0159 (#3314). ADR-0159 gave sound empty-instance zero values to bare `map[K,V]`/`[]T`/`[N]T`/`sequence[T]` slots — but the coverage has a gap for these types when they are **fields of a struct-typed local with no initializer**.

**Repro (NRE, not empty):**
```gsharp
package repro
struct S {
    public var Items []int32
}
var s S
Console.WriteLine(s.Items.Length)   // NullReferenceException
```

Compare to the working top-level case (the #2262 fix):
```gsharp
var x []int32
Console.WriteLine(x.Length)   // 0 -- correct, per ADR-0159
```

**What works today** (per #3314): top-level magic-type locals/globals; class instance fields (via the emitted instance constructor, which the synthesized initializer hooks into); `shared`/static fields (via the .cctor).

**What doesn't**: a magic-type field *inside* a struct-typed local. Structs with no other reason to synthesize a constructor (per #3219's "all-public structs keep byte-identical emission" rule) have no ctor for #3314's initializer injection to hook into, and the struct-local declaration path (`var s S`) doesn't recurse into the struct's own field defaults to apply them independently.

**Likely fix directions** (for whoever picks this up): either (a) the struct-local declaration path emits per-field zero-value initialization for magic-type fields inline at the `var s S` site (mirroring how #2262's top-level case works), independent of whether a ctor exists, or (b) extend the #3219 ctor-synthesis trigger to also cover "struct has a magic-type field needing a non-default zero value," giving such structs a synthesized ctor purely for this purpose. Also audit: does the same gap exist for a struct field nested inside a CLASS (`class C { public var S S }` where `S` has a slice field) — the class's own ctor exists, but does it recurse into `S`'s own field defaults, or only its own directly-declared fields?

Part of #3163 follow-ups. Related: #3310, #2262, #3314, #3219.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3331, opened 2026-08-07
Fix #3329: reconstruct ADR-0159 zero values for cross-assembly struct literals

### LINKED ISSUE #3329: REPL: struct literal construction NREs when the struct type was declared in an earlier submission
Found while verifying #3319 (struct-local zero-value recursion). Unrelated to that fix — reproduces identically on main with a struct that has no magic-collection field at all, using only the pre-existing #3219/#3314 struct-literal machinery.

**Repro (REPL, via `EmittedSessionEngine` — same shape as `gsi` cross-cell):**

Cell 1:
```gsharp
struct S {
    public var Items []int32
}
```

Cell 2:
```gsharp
var s = S{}
s.Items.Length
```

Cell 2 throws a `NullReferenceException` at runtime (surfaced as REPL diagnostic `GSI002`) when `S` was declared in an *earlier* submission and the struct literal `S{}` is constructed in a *later* one. The same failure reproduces with zero magic-collection fields involved:

Cell 1:
```gsharp
struct S {
    public var X int32
}
```

Cell 2 (bare declaration, no literal node at all — `initobj` only) works fine:
```gsharp
var s S
s.X
```

but an *explicit* empty struct literal `S{}` for a struct declared in an earlier cell fails. This points at cross-submission struct-literal token/slot resolution (`structLiteralSlots` / `ResolveUserTypeToken`-family machinery in `MethodBodyEmitter`/`ReflectionMetadataEmitter`) not correctly handling a struct type that lives in a *referenced* prior-submission assembly rather than the current submission's own `TypeDef`s — as opposed to `BoundDefaultExpression`/`initobj`, which resolves the cross-assembly type token fine.

**Impact:** any G# feature that lowers a bare/implicit struct-typed slot into an explicit struct-literal `BoundStructLiteralExpression` (e.g. #3319's zero-value synthesis, or ordinary user code writing `S{}` for a cross-cell-declared struct) NREs at runtime in the REPL when the struct crosses a submission boundary. `#3319`'s own tests route around this by declaring the struct and constructing its zero-valued instance together in the same cell, and only reading/mutating from later cells — masking rather than fixing the gap.

**Suggested owner for investigation:** whoever picks this up should start in `MethodBodyEmitter.Expressions.cs`'s `EmitStructLiteral` (the `structLiteralSlots`/`ResolveUserTypeToken`/`ResolveUserCtorTokenForDefault` calls) and how `ReflectionMetadataEmitter`'s user-type-token resolution behaves for a `StructSymbol` that originates from a `SubmissionReference`'s `BoundGlobalScope` (i.e. imported from a previously-emitted submission assembly) rather than the current compilation's own struct declarations.

Part of #3163 follow-ups.

DavidObando:
Fixes #3329. Part of #3163

## Summary

- A struct literal (`S{}`) for a struct declared in ANOTHER assembly (or, for the REPL, an EARLIER submission) lost its ADR-0159 "sound zero value" instance-field initializer for a magic-collection-typed field (`map[K,V]`/`[]T`/`[N]T`/`sequence[T]`) — the field was left at its raw CLR default (a null reference) instead of an empty instance, NREing the first time the field was used.
- **Root cause** (not the initobj-vs-literal token-resolution divergence originally suspected in the issue): `StructSymbol.InstanceFieldInitializers` is populated only while binding a struct's OWN declaration (`DeclarationBinder.Structs.cs`). A struct symbol REBUILT FROM REFLECTION over an already-emitted type never reconstructed it — neither `ImportedTypeSymbol.BuildSemanticAggregate` (the `data struct`/primary-ctor-struct semantic-aggregate path) nor `ExpressionBinder.BindImportedTypeObjectInitializer` (the generic imported-value-type fallback used for a PLAIN struct, which doesn't carry the `GSharp.TypeSemantics` marker at all). `UserTokenResolver.ResolveUserTypeToken`/`ResolveFieldToken` already resolve cross-assembly struct/field tokens correctly — confirmed by every already-working control case (bare `var s S`, `S{X: 5}`, nested/call-arg/return-value literals all passed red, unmodified).
- **Fix**: a new `GSharp.MagicCollectionFields` `AssemblyMetadataAttribute` marker (independent of the existing `GSharp.TypeSemantics` gate, so plain structs are covered too) records, per gsc-compiled VALUE-TYPE struct, which fields need ADR-0159 zero-value synthesis and a shape tag (`map`/`slice`/`arrN`/`seq` — the tag alone disambiguates a slice from a same-CLR-shaped fixed array, which carries no length in metadata). Both cross-assembly struct-literal binding paths consult it and reconstruct the exact zero value from the field's own reflected CLR shape (`MagicCollectionZeroValue.ClassifyForMarker` / `TrySynthesizeEmptyInstanceFromMarker`). A class literal needs no fix — it always calls its REAL parameterless constructor, which already applies these initializers in-type.
- Regenerates the `RefactoringBaselineTests` IL baseline for `samples/PrimaryCtorVariadic.gs` — its `inline struct Ids(values ...int32)` now carries the new marker; confirmed the only affected sample.

## Verification

- Release solution build (`dotnet build GSharp.sln -c Release`): 0 warnings, 0 errors.
- Witnesses (ADR-0154): both new test files confirmed RED on unpatched `origin/main` (verified via `git stash` of the `src/` fix, keeping the tests) then GREEN after restoring the fix:
  - `test/Interpreter.Tests/Issue3329CrossCellStructLiteralReplTests.cs` — 10/10 (REPL cross-cell via `EmittedSessionEngine`; RED: 2 failed — the issue's exact magic-collection repro and its `data struct` analog — before the fix, matching every other case, which was already green and stays a regression pin).
  - `test/Compiler.Tests/Emit/Issue3329CrossAssemblyStructLiteralEmitTests.cs` — 8/8, each compiling a real library + consumer assembly pair through `gsc`, ILVerify-ing both, and running the consumer (RED: 6/8 failed before the fix — slice/map/sequence/fixed-array/partial-explicit-init/`data struct` shapes; the same-assembly regression pin and the "inner explicitly initialized" nested case were unaffected, as expected).
- Full `test/Core.Tests` (Release): **7447 passed, 1 skipped (`RegenerateBaseline`, expected-manual), 0 failed** — includes the regenerated IL baseline gate.
- Full `test/Interpreter.Tests` (Release): **1502 passed, 0 failed**.
- `LanguageConformance` filter (`test/Compiler.Tests`, Release): **128/128 passed**.
- NOT RUN: the rest of `test/Compiler.Tests` outside the `LanguageConformance` filter, and `Cs2Gs.Tests`, `LanguageServer.Tests`, `Sdk.Tests`, `Extensions.Tests`, `InternalAnalyzers.Tests` — outside the verification chain this task specified; the two new witness files (Compiler.Tests + Interpreter.Tests) and the full Core.Tests/Interpreter.Tests/LanguageConformance runs above are the change's actual blast radius (Binding/Emit/Symbols in `Core`, consumed identically by the REPL and by ordinary `gsc` cross-assembly compilation).

**Verdict: MERGEABLE.**

## Coverage

| Shape | Result |
|---|---|
| Issue's exact repro (magic-collection field, `S{}`) | fixed |
| Issue's "zero fields involved" repro (`X int32`, `S{}`) | was never actually broken (confirmed on unpatched main — the issue's own text is imprecise here); passes, kept as a pin |
| `S{X: 5}` explicit field initializer, cross-assembly | passes (was already working; kept as a pin) |
| Explicit initializer alongside an UNSET magic-collection field | fixed |
| `map[K,V]`, `sequence[T]`, fixed-size `[N]T` fields | fixed (not just the issue's slice repro) |
| `data struct` (semantic-aggregate path) | fixed — same underlying gap, different code path |
| Nested struct-in-struct, cross-assembly, inner explicitly initialized | passes |
| Struct literal as call argument / return value | passes (was already working) |
| Same-assembly / same-cell struct literal | regression pin, unaffected |
| Bare declaration (`var s S`, pure `initobj`) | regression pin, unaffected |
| ILVerify on every emitted shape above | clean |

Out of scope, explicitly not touched: recursing an outer struct literal's OWN zero-value synthesis into an inner struct-typed field that itself needs a magic-collection zero value (e.g. `Outer{}` where `Outer.I Inner` and `Inner` has an unset slice field) — this reproduces identically same-assembly, with no cross-assembly reference at all, confirming it's issue #3319/PR #3330's territory, not #3329's. #3330 was not yet merged when this PR was built, so no dependency on it either way.

## Checklist

- [x] Behavioral tests carry a witness of discrimination (ADR-0154): each new/changed test fails on the pre-change code and passes with it (see Verification above).
- [x] Self-review verdict recorded: MERGEABLE.


DavidObando:
## Rebase onto #3330

This branch is rebased onto current `main` (`9d634d6e`, which includes #3330 / issue #3319 — "Recurse into struct fields for magic-collection zero values", merged after this PR was opened).

### The conflict

Both PRs modified `src/Core/CodeAnalysis/Binding/MagicCollectionZeroValue.cs`. #3330 restructured the file top-to-bottom (renamed the private core to `TrySynthesizeEmptyInstanceCore`, added a `visiting`-guarded recursive `TrySynthesizeStructFieldDefaults` for same-assembly struct-in-struct recursion, added `MightNeedZeroValue`, moved `RequiresExplicitInitializer`). This PR had appended `ClassifyForMarker` / `TrySynthesizeEmptyInstanceFromMarker` after the file's original (pre-#3330) shape. Git flagged the whole back half of the file as conflicting. Resolved by keeping #3330's full recursive structure verbatim and layering this PR's marker-based cross-assembly reconstruction on top of it as new methods, rather than reverting or picking one side.

### The composition gap (found, not just a rebase mechanic)

This PR's own body explicitly scoped out "recursing an outer struct literal into an inner struct-typed field that itself needs a magic-collection zero value," calling it #3319/#3330's territory — #3330 wasn't merged yet when this PR was authored, so there was no way to test the composed case.

Now that #3330 is merged, that composed case is reachable, and I found it does **not** work out of the box: a naive union of both diffs leaves it broken. Root cause — #3330's same-assembly recursion walks a struct's `Fields` directly, but a struct symbol rebuilt from reflection (`ImportedTypeSymbol.BuildSemanticAggregate`) resolves each field's type via raw `TypeSymbol.FromClrType`, which never produces a full `StructSymbol` for a nested struct-typed field. So #3330's recursion silently doesn't fire across the cross-assembly boundary this PR's marker exists to bridge.

**Confirmed red** on the straightforward (non-composed) union of both diffs: a new test constructing `Outer{}` cross-assembly, where `Outer.I Inner` is unset and `Inner` itself has an unset slice field, threw `NullReferenceException` at the exact line reading `o.I.Items.Length`.

**Fix** (small, follows the existing marker design rather than touching field-resolution machinery): `ClassifyForMarker` now also tags a struct-typed field `"struct"` when — and only when — recursing into it (reusing #3330's own `TrySynthesizeEmptyInstance`, safe to call here since marker emission runs post-binding, unlike the early per-struct probes in `DeclarationBinder.Structs.cs`) finds a magic-collection zero value somewhere in its closure. On the consuming side, `TrySynthesizeEmptyInstanceFromMarker` recurses one level into the nested field's own reflected CLR type and its own independent marker entry (every qualifying struct in a compilation gets its own marker entry, not just top-level ones — see `AssemblyAttributeEmitter.EmitGSharpMagicCollectionFieldMarkers`), reconstructing a nested `BoundStructLiteralExpression` the same shape #3330 would have built had the struct been declared in this compilation. Both existing marker-consumption call sites (`ImportedTypeSymbol.BuildSemanticAggregate` for `data struct`, `ExpressionBinder.BindImportedTypeObjectInitializer` for a plain struct) pick this up for free — both already dispatch through `TrySynthesizeEmptyInstanceFromMarker` per marker entry, so neither needed its own change.

No other interaction issues found: same-assembly recursion (#3330) and cross-assembly reconstruction (#3329) never both fire on the same field (a field is either declared in this compilation, using #3330's path, or reconstructed from reflection, using #3329's path) — no double-application risk.

### Integration witness (new)

Added to `test/Compiler.Tests/Emit/Issue3329CrossAssemblyStructLiteralEmitTests.cs`:
- `NestedStructInStruct_BothDeclaredInLibrary_OuterLiteralZeroInitsInnerMagicField` — plain-struct path.
- `DataStructNestedInStruct_BothDeclaredInLibrary_OuterLiteralZeroInitsInnerMagicField` — `data struct` / semantic-aggregate path.

Both confirmed **red** (`NullReferenceException`) against the unfixed composition, **green** after the fix above. Both compile a real library + consumer assembly pair through `gsc`, ILVerify both, and run the consumer — same harness as the rest of the file.

### Witness coverage overlap (checked, no action needed)

Reviewed #3330's witnesses (`Issue3319StructLocalZeroValueEmitTests`, `Issue3319StructLocalZeroValueReplTests`) against this PR's (`Issue3329CrossAssemblyStructLiteralEmitTests`, `Issue3329CrossCellStructLiteralReplTests`). No redundancy: #3330's cover same-compilation struct-field recursion; #3329's cover cross-assembly/cross-cell reconstruction; the two new tests above are the only case that needed both.

### Verification (sequential, Release)

- `dotnet build GSharp.sln -c Release`: **0 warnings, 0 errors.**
- New integration witnesses (2): confirmed **red → green** as described above.
- Full existing witness matrix from both PRs, rebuilt against the composed tree:
  - `Issue3319StructLocalZeroValueEmitTests` (Core.Tests): **15/15 passed.**
  - `Issue3319StructLocalZeroValueReplTests` (Interpreter.Tests): **5/5 passed.**
  - `Issue3329CrossAssemblyStructLiteralEmitTests` (Compiler.Tests, now 10 facts incl. the 2 new ones): **10/10 passed.**
  - `Issue3329CrossCellStructLiteralReplTests` (Interpreter.Tests): **10/10 passed.**
- Full `test/Core.Tests`: **7462 passed, 1 skipped (`RegenerateBaseline`, expected-manual), 0 failed** — includes the `RefactoringBaselineTests` IL baseline gate both PRs touch.
- Full `test/Interpreter.Tests`: **1507 passed, 0 failed.**
- `LanguageConformance` filter (`test/Compiler.Tests`): **128/128 passed.**
- ILVerify on the integration case: clean (via the new tests' own `IlVerifier.Verify` calls over both the library and consumer assemblies — no separate run needed).
- NOT RUN: full ILVerify corpus (`build/run-ilverify.sh`), `Cs2Gs.Tests` (pre-existing non-deterministic host crash — known repo issue, unrelated), `Sdk.Tests`, `LanguageServer.Tests`, `Extensions.Tests`, `InternalAnalyzers.Tests`, e2e tests — outside this task's specified verification chain and outside this change's blast radius (`src/Core/CodeAnalysis/Binding` + `Emit`/`Symbols`, exercised identically by `gsc`, the REPL, and Core.Tests/Interpreter.Tests/LanguageConformance above).

### Self-review verdict

**MERGEABLE.** The rebase composition is coherent (read the whole merged `MagicCollectionZeroValue.cs` top to bottom, no duplicate logic, no double-application risk between the two recursion paths). The one real interaction gap between the two PRs — cross-assembly reconstruction not recursing into a nested struct-typed field — was found by writing the integration test the two PRs individually couldn't, confirmed red on the naive composition, and fixed with a small, design-consistent extension to the existing marker scheme. All witnesses (old and new) green; full Core.Tests, Interpreter.Tests, and LanguageConformance green.

Issue #3347, opened 2026-08-11
cs2gs synthetic spills: full inventory, one defect, and a gap-closing plan


DavidObando:
> Analysis produced with Claude Code. Every "G# today" listing below is real output from the
> built `Cs2Gs.Translator`; every proposed G# spelling was fed to the built `gsc` and is marked
> ✅ compiles / ❌ rejected (with the actual error) / 🔧 proposed extension. Frequency figures come
> from a Roslyn-based scan of dotnet/roslyn (6.26M LOC) and this repo (405k LOC).

**Work items are tracked as sub-issues of this one.**

---

# cs2gs synthetic-spill analysis

**Date:** 2026-08-10
**Repo:** `DavidObando/gsharp` @ `47259d0c` (branch `copilot/3336-partial-member-import-scope-20260810-1852`)
**Scope:** every code path in `tools/cs2gs/` that synthesizes a temporary to make an operand
evaluate exactly once.

---

## 1. What counts as a "synthetic spill"

cs2gs mints five families of synthetic name. Only the first three are spills in the
single-evaluation sense; the last two are listed for completeness because they are commonly
mistaken for spills in output diffs.

| Prefix | Minted by | Spill? |
|---|---|---|
| `__spillN` | `SpillOperand` + 3 direct emitters | **yes** — an operand that would otherwise be printed (and re-evaluated) at 2+ output positions |
| `__pN` / `__initN` | null-seam helper capture | **yes** — the spill's fallback when no statement seam exists; becomes a synthesized static helper's parameter |
| `__scrutineeN` | loop-condition `is`-pattern hoist | **yes** — a condition operand hoisted into the loop body |
| `__deconN` | tuple-deconstruction lowering | no — a native `let (a, b) = rhs` binding, not a duplication guard |
| `__qN`, `__argN`, `__foreachN`, `__usingN`, `__caughtN`, `__gotoCaseN`, `__switchExitN`, `__iteratorExitN`, `__gsAsyncVoid*`, `__anonN`, `__generatedRegexN` | various lowerings | no — structural synthetics |

### Core machinery

| Element | Location |
|---|---|
| `SpillOperand(GExpression, List<GStatement>)` — mints `__spillN` | `Cs2Gs.Translator/CSharpToGSharpTranslator.Deconstruction.cs:1247` |
| `SpillOperand(GExpression)` — ambient-seam overload | `…Deconstruction.cs:893` |
[...]

PR #3363, opened 2026-08-11
Keep LINQ query-clause hoists inside their own lambda

### LINKED ISSUE #3348: cs2gs: spill escapes a LINQ query clause into the enclosing method body
Parent: #3347 (§5.1)

## Problem

A single-evaluation spill hoisted while translating a LINQ **query clause** escapes into the
enclosing method body, where it references the query's range variable — which is not in scope there.
The emitted G# will not bind.

```csharp
from x in xs where x.Tag is string s && s.Length > 0 select x
```
```gs
func M(xs IEnumerable[N]) IEnumerable[N] {
    let __spill0 = x.Tag                                    // <-- `x` is not in scope here
    return xs.Where((x N) -> __spill0 is string && (__spill0 as string)!!.Length > 0)
}
```

Two faults: the identifier is undefined, and the operand is evaluated once for the whole query
instead of once per element.

## Cause

`TranslateLambda` (`CSharpToGSharpTranslator.Types.cs:104`) deliberately nulls
`PendingSpillPrologue` at a lambda boundary and reopens a fresh seam via `WithSpillSeam`, precisely
so a spill cannot leak into an unrelated scope. The query-lowering path builds its lambdas without
doing either — it calls `TranslateExpression(lambdaBody)` while the *enclosing statement's* seam is
still active:

- `Types.cs:1961` `BuildScopeLambda` (lines 1971 and 1974)
- `Types.cs:1982` `BuildTransparentResultSelector`
- `Types.cs:1807` `LowerLetClause` (line 1815)

## Blast radius

Confirmed for `where`, `select`, `let` and `orderby` clauses, and for **both** spill families — a
value-position assignment in a query clause escapes identically:

```csharp
from x in xs where (last = x) > 0 select x
```
```gs
var last = 0
last = x                                 // <-- same escape, same undefined `x`
return xs.Where((x int32) -> (last) > 0)
```

The same predicate written as an explicit lambda is correct, which isolates the fault to the query
path:

```gs
xs.Where((x N) -> {
    let __spill0 = x.Tag                 // correctly inside the lambda
    return __spill0 is string && (__spill0 as string)!!.Length > 0
})
```

## Why tests miss it

The output *parses* — `GSharpRoundTrip.Validate` passes. The failure is a binder error, so only a
compile-the-output gate catches it.

## Fix

Wrap each query-lambda body translation in `WithSpillSeam` (block-bodied when anything spills),
mirroring `TranslateLambda`. Add a compile-the-output test for query expressions.

DavidObando:
Fixes #3348. First item from the plan in #3347.

## Problem

A query clause lowers to a lambda, but its body is not an
`AnonymousFunctionExpressionSyntax`, so neither the spill seam (`TranslateLambda` suspends and
reopens it) nor `CollectEmbeddedAssignments` (which stops at a lambda) treated it as its own
evaluation scope. Both kinds of hoist escaped:

```csharp
from x in xs where x.Tag is string s && s.Length > 0 select x
```
```gs
let __spill0 = x.Tag                                    // `x` is not in scope here
return xs.Where((x N) -> __spill0 is string && (__spill0 as string)!!.Length > 0)
```

Two faults: the identifier is undefined, and the operand is evaluated once for the whole query
instead of once per element. The same escape hit embedded value-position assignments:

```csharp
from x in xs where (last = x) > 0 select x
```
```gs
last = x                                 // same escape, same undefined `x`
return xs.Where((x int32) -> (last) > 0)
```

## Fix

`BuildScopeLambda` and `LowerLetClause` now translate the clause body through
`TranslateQueryLambdaBody`, which reuses the existing `TranslateConditionWithHoist` — it already
implements exactly the needed contract (redirect the spill seam *and* hoist embedded assignments
into a caller-supplied list). Hoists land in the lambda's own prologue, after the scope-parameter
deconstruction they may read from:

```gs
xs.Where((x N) -> {
    let __spill0 = x.Tag
    return __spill0 is string && (__spill0 as string)!!.Length > 0
})
```

`CollectEmbeddedAssignments` no longer descends into a query clause. `EagerQuerySources` carves out
the sub-expressions a query genuinely evaluates in the enclosing scope — the first `from`'s source
and each `join`'s `in` source, including through `into` continuations. C# forbids either from
referencing a range variable, so those keep hoisting exactly as before.

`BuildTransparentResultSelector` needed no change after all: it builds its body from bare
identifiers and never calls `TranslateExpression`, so it cannot spill. Correcting what #3348 claimed.

## Coverage

`Issue3348QueryClauseSpillSeamTests` — 11 tests:

- spill stays inside the lambda for `where`, `select`, `let`, `orderby`, an additional `from`, and a
  `join` key selector
- value-position assignment stays inside the lambda
- **guards for the carve-out**: an assignment in the first `from` source, and in a `join`'s `in`
  source, still hoist ahead of the query
- control: the same predicate as an explicit lambda (always worked)
- a hoist-free query keeps the compact arrow-lambda form

Every case asserts the output **binds**, not merely parses — parse-only validation is precisely
what let this defect through, so `GSharpRoundTrip` alone would not have caught it.

Verified the 7 behavioural tests fail without the source change and pass with it; the 4
control/guard tests pass both ways.

## Test run

Full `Cs2Gs.Tests` suite: **1918 passed, 0 failed** (9m39s).

### DIFF
--- a/tools/cs2gs/Cs2Gs.Tests/Issue3348QueryClauseSpillSeamTests.cs
+++ b/tools/cs2gs/Cs2Gs.Tests/Issue3348QueryClauseSpillSeamTests.cs
@@ -1,0 +1,333 @@
+// <copyright file="Issue3348QueryClauseSpillSeamTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using Cs2Gs.CodeModel.Ast;
+using Cs2Gs.CodeModel.Printing;
+using Cs2Gs.CodeModel.RoundTrip;
+using Cs2Gs.Translator;
+using Cs2Gs.Translator.Loading;
+using GSharp.Core.CodeAnalysis.Binding;
+using GSharp.Core.CodeAnalysis.Syntax;
+using GSharp.Core.CodeAnalysis.Text;
+using Xunit;
+
+namespace Cs2Gs.Tests;
+
+/// <summary>
+/// Regression tests for issue #3348: a hoist performed while translating a LINQ
+/// query CLAUSE escaped into the enclosing method body. A query clause lowers to
+/// a lambda, but — unlike an explicit lambda — its body is not an
+/// <c>AnonymousFunctionExpressionSyntax</c>, so neither the spill seam
+/// (<c>TranslateLambda</c> suspends/reopens it) nor
+/// <c>CollectEmbeddedAssignments</c> (which stops at a lambda) treated it as its
+/// own evaluation scope.
+/// <para>
+/// Both kinds of hoist were affected — a single-evaluation spill (issue #1731)
+/// and an embedded value-position assignment (issue #1723) — and both produced
+/// the same two faults: the hoisted statement referenced the query's RANGE
+/// VARIABLE, which is not in scope in the enclosing body (so the emitted G# did
+/// not bind), and it ran once for the whole query instead of once per element.
+/// </para>
+/// <para>
+/// The emitted G# still PARSED, so <see cref="GSharpRoundTrip"/> validation —
+/// which every other cs2gs translation test relies on — could not catch this.
+/// Every case here therefore also asserts the output BINDS, via
+/// <see cref="BindDiagnostics"/>.
+/// </para>
+/// </summary>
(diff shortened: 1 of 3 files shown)


DavidObando:
It looks like the interpreter tests are flaky and can get stuck randomly. Worth looking into it.

DavidObando:
Logged a bug for it: https://github.com/DavidObando/gsharp/issues/3364

Issue #575, opened 2026-06-08
Language: no `expr as T` (safe cast) or `expr is T` (type-test) operator in expression position


DavidObando:
## Summary

G# has no `as` (safe cast) or `is` (type-test) operators that work in expression position. C#'s `expr as T` (returns `T?` or `null` on type mismatch) and `expr is T` (returns `bool`) have no direct G# equivalent.

`T(expr)` works as a throwing cast (the equivalent of C# `(T)expr`), but there's no expression-form way to:
- Safely cast and inspect for failure: `var arr = node as JsonArray; if (arr != null) { ... }`
- Test type membership: `if (node is JsonArray) { ... }`

The spec mentions [pattern matching](https://davidobando.github.io/gsharp/docs/ref/spec/#patterns) — but for type tests in plain `if`/`var` expression contexts (not `switch`), I couldn't find a working syntax.

## Repro

```gsharp
package Probe

import System.Text.Json.Nodes
import Xunit

type Probe class {
    @Fact
    func Safe_Cast() {
        let node JsonNode = JsonArray()
        let arr = node as JsonArray       // GS0125: Variable 'as' doesn't exist
    }

    @Fact
    func Type_Test() {
        let node JsonNode = JsonArray()
        if node is JsonArray {            // GS0005: Unexpected token <IsKeyword>
            Assert.True(true)
        }
    }
}
```

## What works

```gsharp
let arr = JsonArray(node)         // throwing cast — InvalidCastException on mismatch
```

## Expected

Two related operators in expression position:
- `expr is T` → `bool`. Same precedence/usage as C#.
- `expr as T` → `T?`. Returns null if the runtime type isn't `T`-assignable.

(Optionally also `expr is T as var name` C#-style declaration patterns.)

## Why it matters

Without safe-cast, defensive code that handles "this *might* be the right type" needs `try`/`catch (InvalidCastException)` boilerplate — way uglier than `if (x is T t)`.

## Context

`gsc` version: `Gsharp.NET.Sdk/0.1.459`, .NET 10 RTM.

Discovered while porting `tests/Oahu.Cli.Tests/Commands/OutputWriterTests.cs` to G# — the original test inspects a `JsonNode` and downcasts to `JsonArray`/`JsonObject`/`JsonValue` to assert on the JSON shape. The port works around with `JsonArray(node)` (throwing) but loses the test's intent of asserting the type *is* JsonArray.

DavidObando:
[...]

PR #3385, opened 2026-08-12
Allow full patterns in boolean is expressions

### LINKED ISSUE #3351: G#: allow patterns in boolean 'is'-expression position
Parent: #3347 (§7 gap **G1**) · **Closing order: 3** · Impact **10/10**

## Gap

G#'s pattern grammar is rich and already implemented — `Parser.Patterns.cs` has property patterns,
list patterns, relational patterns, `and`/`or`/`not` combinators and binding type patterns. But
`ParsePattern()` is reachable from only two places, both `switch`:

- `ParseSwitchCase` — `Parser.Patterns.cs:51`
- `ParseSwitchExpressionArm` — `Parser.Expressions.Lambdas.cs:437`

The expression-level `is` (`Parser.Expressions.cs:480`) builds `IsExpressionSyntax` from
`ParseTypeClause()` — a bare type test, nothing more.

**The gap is not "G# lacks patterns". It is "G# patterns are unreachable from boolean position".**

Verified:

```gs
if n.Tag is string { Length: > 0 } { … }
// ❌ GS0005: Unexpected token <GreaterToken>, expected <IdentifierToken>

if n.Kind is "a" or "b" { … }
// ❌ GS0113: Type '' doesn't exist / GS0125: Variable 'or' doesn't exist
```

The same tests via `switch` compile fine — which is exactly why cs2gs emits no spill there:

```gs
switch n.Tag {                                    // ✅ compiles
    case s is string when s.Length > 0 { Console.WriteLine(s) }
    default { }
}
```

## Impact

This single gap causes spill site **S1**, which is **~90% of all `__spillN` in cs2gs output**:
71.5 per 100k LOC in dotnet/roslyn, 259.4 per 100k in this repo. **1.5% of all Roslyn methods** and
**8.8% of all Roslyn files** would host at least one; in this repo, 5.8% of methods and 11.6% of
files.

Today:

```csharp
if (node.StartTag is { } startTag) { Console.WriteLine(startTag.X); }
```
```gs
{
    let __spill0 = node.StartTag
    if __spill0 != nil {
        Console.WriteLine(__spill0!!.X)
    }
}
```

Note the collateral damage beyond the temp: the author's binder name `startTag` is destroyed, and an
extra brace level is introduced to scope the spill.

## Proposed

Let `is` parse a full `PatternSyntax` on its right-hand side, reusing `ParsePattern()`:

```gs
if node.Tag is string { Length: > 0 } { … }
if n.Kind is "a" or "b" { … }
```

Plausibly a contained parser + binder change, since the pattern binder already exists for `switch`.
Needs care around the `{` ambiguity with a following block — the same problem C# solves and the
reason `ParseExpressionInBodyHeader` exists.

## Partial fix available today, no language change

`if let` (ADR-0071) covers the single most common sub-case, `x.Y is { } name` — alone ~44/100k LOC
in Roslyn and ~186/100k here. Tracked separately as a cs2gs work item.

Does **not** cover the type-test form; see **G2** (#3347 §7).

DavidObando:
## Summary

Fixes #3351.

- accepts full non-binding patterns after boolean `is` / `!is`: constants, relational, list/slice, property, parenthesized, and `not` / `and` / `or`
- adds type+property composition (`string { Length: > 0 }`) and threads type narrowing across pattern `and` in boolean and switch positions
- evaluates the tested expression once, binds through `PatternBinder`, and lowers through `MethodBodyEmitter.EmitPattern` to a boolean value
- records grammar, disambiguation, binding, and brace policy in ADR-0162

## Language design: option B

This chooses option B: bare type clauses are type patterns in every pattern position.

Name-shaped patterns retain both type and value interpretations in `TypeOrConstantPatternSyntax` until semantic binding:

- switch patterns resolve value-first, then type, preserving existing enum member / const / value cases
- boolean `is` resolves type-first, then value, preserving existing `expr is Type`
- a property suffix forces type interpretation
- `== name` explicitly forces value interpretation after boolean `is`; legacy `_ is Type` remains available to force a switch type pattern when a value shadows the type

This is deterministic symbol resolution, not capitalization or naming heuristics. Tests cover enum members, global/local consts, type/value name collisions, and sealed-hierarchy exhaustiveness.

## Grammar and brace handling

`Expression is Pattern` now reaches the complete pattern grammar. Type patterns optionally carry a property suffix.

Body-header braces are parsed speculatively using parser position/diagnostic rollback and existing body-header suppression state:

- `if x is { P: 1 } { body }` commits the first brace as a property pattern because a body brace follows
- `if x is Type {}` keeps `{}` as the body
- `if x is {} { body }` works
- `if (x is {}) { body }` is the explicit unambiguous escape hatch
- malformed property patterns retain token-accurate parser diagnostics

## Binding and lowering architecture

- `IsExpressionSyntax` owns `PatternSyntax`; `BoundIsExpression` owns `BoundPattern`
- `ExpressionBinder` delegates to `PatternBinder` with `allowBindings: false`
- `BoundTypePattern` carries an optional bound property suffix
- `BoundBinaryPattern` can carry the narrowed left value used to bind/emit the right side of `and`
- general boolean patterns spill the subject once, call `EmitPattern`, then materialize `true` / `false`
- plain legacy type tests keep their previous direct `isinst` lowering byte-for-byte, preserving emitted PE baselines
- walkers, rewriters, async spiller, slot planner, type-parameter collection, printer, and expression-tree validation were updated coherently

Expression trees continue to support plain type tests. General pattern tests report existing GS0473 instead of reaching an unsupported lowerer shape.

## Binding policy

Boolean pattern position introduces no names. Type-pattern and slice bindings report new GS0525 at the binding identifier and direct users to `if let` / `guard let`. Switch binding rules remain unchanged, including GS0390 under `or` / `not`.

## Type+property and narrowing

Both forms now work in switch statements, switch expressions, and boolean position:

```gs
if value is string { Length: > 0 } { ... }
case Dog { Name: "Rex" }: value.Bark()
case Dog and { Name: "Rex" }: value.Bark()
```

The right side of `and` binds and emits against the successfully narrowed value, not the original discriminator type. Smart-cast branch narrowing consumes the resulting bound pattern while preserving the prior stability policy for boolean tests and switch arms.

## Runtime matrix

The new ILVerify + execution test covers positive and negative paths for:

- constant `or`
- relational `and`
- list/slice
- `not`
- parenthesized and nested combinators
- property and type+property patterns
- nil and type tests
- `!is`
- array type patterns
- async-spilled pattern expressions
- single evaluation of the subject
- switch statement/expression type+property composition and narrowing across `and`
- enum/const/type collision compatibility

## Compatibility / RED evidence

Issue #3351 documented the prior failures: non-type patterns were parsed as type clauses, type+property syntax failed, and `and` property members bound against the un-narrowed discriminator. New parser/binder/runtime tests exercise those exact shapes.

Compatibility evidence:

- all 7,558 Core tests pass (1 manual baseline test skipped)
- all 4,329 Compiler tests pass
- all 468 LanguageServer tests pass
- existing issue #575 `is` emit tests pass
- existing simple-type emitted PE baselines remain unchanged
- enum member, local/global const, name-led expression, and type/value collision tests preserve old switch value interpretation

## Validation

- `dotnet build GSharp.sln --no-restore --nologo`
- `dotnet test test/Core.Tests/Core.Tests.csproj --no-build --no-restore --nologo` — 7,558 passed, 1 skipped
- `dotnet test test/Compiler.Tests/Compiler.Tests.csproj --no-build --no-restore --nologo` — 4,329 passed
- `dotnet test test/LanguageServer.Tests/LanguageServer.Tests.csproj --no-build --no-restore --nologo` — 468 passed
- targeted issue #3351, issue #575, smart-cast, pattern, property, enum, expression-tree, baseline, and ILVerify runs

## Parent #3347 / cs2gs

This closes the G1 language/compiler blocker and updates the cs2gs coverage/design docs. Translator output is intentionally unchanged: spill removal and native-pattern emission remain parent #3347 translator work, not required for language support itself. No translator change was needed to expose or validate the root feature.


DavidObando:
Addressed both Opus 4.8 non-blocking findings in 59d70127:

- boolean `is` now folds `not nil` to a total pattern for proven non-nullable value types; switch GS0171 behavior remains pinned
- unresolved type-or-value names after boolean `is` now report exact GS0125 value-resolution diagnostics
- added focused binder/runtime coverage for value-type positive/negative paths, switch compatibility, misspellings, unqualified/qualified enum members, and valid type-first collisions

Validation: focused Core 108 passed; focused Compiler 20 passed; full solution build succeeded with 0 warnings/errors.

DavidObando:
CI fixes pushed in 5f17a1cf:

- removed all newly introduced null-forgiving operators by flow-safe binding identifier handling and a nullable-return parser helper
- regenerated `docs/cs2gs-coverage-matrix.md` with repository `cs2gs coverage --write` command

Validation:
- `python3 build/nullable_hygiene.py --base origin/main` — OK
- `ConstructInventoryGoldenTests` — 7 passed
- issue #3351 + parser/binder tests — 61 passed
- issue #3351 + legacy issue #575 emit tests — 15 passed
- `cs2gs coverage` reports inventory/golden/docs in sync.

Issue #3361, opened 2026-08-11
cs2gs: scrub stale range-slice wording from spill comments and diagnostic


DavidObando:
Parent: #3347 (§5.3)

## Problem

Issue #1896 replaced the `.Slice` desugaring with `gsc`'s native `recv[start..end]` range-index form,
which embeds each bound exactly once. **Range slices no longer spill anywhere.**

Confirmed by code (`TranslateRangeSlice`, `CSharpToGSharpTranslator.Patterns.cs:1950`, has no
`SpillOperand` call) and by test
(`Issue1731DoubleEvaluationTranslationTests.RangeSliceStart_SideEffectingOperand_EvaluatesOnce` now
asserts `DoesNotContain(".Slice(")`).

Six places still describe range slices as a spill source, including a **user-visible diagnostic
string**:

- `Deconstruction.cs:936` — the `Unsupported` message reads *"a non-trivial
  pattern-scrutinee/range-slice operand here has no enclosing statement to host a single-evaluation
  spill…"*
- `DocumentTranslationState.cs:173`, `:222`
- `ControlFlow.cs:310`
- `Constructors.cs:1047`
- `IfLet.cs:245`

## Fix

Drop the range-slice wording. Start with the diagnostic — it is the one users see, and it currently
sends them looking for a lowering that no longer exists.

DavidObando:
Fixed in #3380 (merged as 2d5cd704).

Eight stale range-slice mentions removed, including the user-visible `Unsupported` diagnostic string.

Three mentions in `Constructors.cs` are deliberately **kept**: they argue that an attribute argument must be a compile-time constant and that neither an `is`-pattern nor a range-slice is one — which remains true. A clarifier was added noting range-slices have not spilled since #1896, so a future reader grepping for stale wording does not delete a still-valid safety argument.

PR #3398, opened 2026-08-13
Eliminate remaining avoidable cs2gs spills

### LINKED ISSUE #3347: cs2gs synthetic spills: full inventory, one defect, and a gap-closing plan

DavidObando:
## Summary

Closes #3347.

Re-ran the synthetic-spill inventory against `f9e8fa47627f27506d4a7e412227a1a4d6fa5b2d` after #3348-#3361 and the follow-up language work landed.

### Current inventory findings

- **Correctness:** value-position user indexer assignment still produced `void` for set-only indexers, and its first fix evaluated RHS before index. Assignment now yields the converted stored value and preserves receiver → index → RHS order.
- **S1 pattern spills:** cs2gs still hand-expanded non-binding patterns despite ADR-0162, duplicating non-trivial subjects. It now emits native boolean patterns.
- **Binding-pattern spills:** typed/property top-level binders now use `if let`, `while let`, or a scoped local carrying the author’s name plus a native residual pattern.
- **Residual-pattern correctness:** nested typed positional/relational constraints beneath top-level binders are preserved across value, statement, loop, and boolean-`&&` if-let lowering; unsafe native residual translations fall back to established spill lowering.
- **C1/C2/C3 assignment spills:** ordinary/chained/compound value-position assignments still passed through pre-ADR-0161 hoisting. They now remain native assignment expressions in conditions, arguments, returns, initializers, lambdas, and chains.
- **`??=` correctness:** value-position coalescing assignment now uses a block/if expression, preserving conditional RHS execution, stored-value semantics, and single evaluation.
- Scrubbed stale spill/hoist assumptions from translator comments, tests, ADR-0115, and the coverage matrix.

### Eliminated spill families

- Non-binding recursive/property/list/combinator patterns over non-trivial subjects.
- Single top-level nullable, typed, and typed-property binders where G# has a scoped binding form.
- Value-type guard spill for non-trivial receivers.
- Simple, compound, chained, condition, argument, return, initializer, and value-returning-lambda assignment spills.
- Query-clause and null-seam pattern spills now made obsolete by native pattern evaluation.

### Remaining semantics-required captures

No avoidable or incorrect spill remains under current G# expressiveness. Regression inventory documents these required cases:

1. **Nested/multiple pattern binders:** boolean `is` intentionally rejects bindings with GS0525, and G# has no general property `var` binding form. Subject capture is required to expose nested values without re-evaluation.
2. **Reassigned pattern binders:** `if let`/`while let` bindings are immutable; a separate captured subject plus mutable C# binder remains required.
3. **Value-position `??=` on non-trivial storage:** scoped block-expression captures preserve C# receiver/index order, getter count, lazy RHS evaluation, and assigned-value semantics.
4. **Static-helper extension receivers:** only methods forced to remain static helpers (for example by-ref/reduced-collision cases) require formation-time or null-guard receiver capture.
5. **Expression-position nested/mixed deconstruction:** G# multi-assignment is statement-valued; structural tuple-result/target-order captures remain required.

## Validation

- `dotnet test test/Core.Tests/Core.Tests.csproj -c Release --no-restore --nologo --filter 'FullyQualifiedName~PropertyAssignmentSpillEmitTests'` — **7 passed**
- Targeted spill regression matrix across changed cs2gs tests — **128 passed**
- Broad affected cs2gs class matrix, including pipeline/binding cases — **220 passed**
- Residual-pattern review regression plus affected if-let/typed-pattern matrix — **52 passed**
- `git diff --check`


DavidObando:
Fixed blocking residual-pattern review finding in `f698162808815f9e480000f31730c072e9c40295`.

- Residual recursive patterns now translate under boolean-pattern mode with `try/finally` restoration.
- Non-empty synthesized guards/bindings reject native if-let lowering and fall back to established spill lowering.
- Regression covers conditional, statement, while-let, and boolean-`&&` surfaces with nested typed positional/relational constraints.
- Corrected stale `??=` statement-hosting comments.

Validation: affected 52-test matrix passed. CI: 35/35 complete (32 success, 3 skipped), merge state `CLEAN`.

PR #3533, opened 2026-08-25
cs2gs: shared documents translate identically in every project (#3501 !! round 2)

DavidObando:
Part of #3501 (audit item 1, round 2). Analysis of the honest 14,408 residual `!!` shows they concentrate in apps the GS0536 polish never reaches — apps that fail translate/compile. The single biggest blocker chained through **five** apps (Cs2Gs.Pipeline, Core.Tests, Compiler.Tests, Interpreter.Tests, Cs2Gs.Tests): the repository mirror crashed with `Linked source 'test/Shared/GoldenFile.cs' translates differently in multiple projects`.

Root cause: usage-driven oblivious promotion is per-project — Pipeline's callers promoted `goldenPath` to `string?` while Core.Tests' kept `string` (verified by diffing the per-project translations; first divergence at the `CompareFile` signature).

Fix: a symbol declared in a document linked into **more than one repository compilation** now runs its null-taint over the whole repository (converged, order-independent — the same `RepositoryCompilations` channel stored members already use) and skips the per-project usage fallback. The divergence guard also reports the first differing line instead of a bare message, so future divergences are diagnosable from the artifact alone.

Expected effect (CI's selfmig job measures it): the five chained apps proceed past translate, their compiles run the GS0536 polish, and the dominant `!!` families (`project!!` ×1,075, `process!!`, `rtConfig!!`, …) get stripped where narrowing proves them redundant — gsc's early-exit narrowing already covers the throw-guard shape they follow.

Oblivious-analysis suites green locally (156 tests: Issue2412/Oblivious/Issue2113).

https://claude.ai/code/session_01VgrKRVPGaEm9zLaYvankA1

### DIFF
--- a/tools/cs2gs/Cs2Gs.Pipeline/TranslateStage.cs
+++ b/tools/cs2gs/Cs2Gs.Pipeline/TranslateStage.cs
@@ -887,8 +887,20 @@ public sealed class TranslateStage : IMigrationStage
         {
             if (!string.Equals(priorGenerated, generated, StringComparison.Ordinal))
             {
+                string[] priorLines = priorGenerated.Split('\n');
+                string[] currentLines = generated.Split('\n');
+                int line = 0;
+                while (line < priorLines.Length && line < currentLines.Length
+                    && string.Equals(priorLines[line], currentLines[line], StringComparison.Ordinal))
+                {
+                    line++;
+                }
+
+                string prior = line < priorLines.Length ? priorLines[line] : "<eof>";
+                string current = line < currentLines.Length ? currentLines[line] : "<eof>";
                 throw new InvalidOperationException(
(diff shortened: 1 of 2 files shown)


DavidObando:
Pushed two more commits after dispatching the selfmig nightly against this branch (run 32875599932):

- The shared-document convergence **worked** — Cs2Gs.Pipeline now passes translate (its next layer is 27 compile errors leaking from the still-unmapped InternalAnalyzers dependency, the known ADR-0169 analyzer-API gap).
- The test projects (Core.Tests/Compiler.Tests/Interpreter.Tests) had a second blocker: the dual-dictionary fixture in `Issue1483ReadOnlyDictionaryForRangeTests.cs` drops its second colliding explicit-interface property and the null member crashed the whole document's print. Null yields now skip (regression test included).

PR #3544, opened 2026-08-26
cs2gs: taint out/ref-argument receivers from their promoted parameters (#3501)

DavidObando:
Part of the #3501 Translator burn-down — attacks the GS0154 "Parameter requires `T` but was given `T?`" family (55 distinct errors in the latest nightly).

An `out`/`ref` argument flows the opposite way through a call — the callee writes the parameter's value into the caller's variable — but `ObliviousNullabilityAnalyzer.CollectArgumentEdges` only modeled argument→parameter edges. A Try-pattern out parameter that can be assigned null (`bool TryGetTupleType(..., out INamedTypeSymbol tupleType)` assigning an `as`-cast) therefore rendered `T?` while the caller's out-var local stayed untainted; gsc independently infers the out-var's type from the promoted parameter, so downstream reads passed a `T?` where the translator believed a `T` flowed and never bridged with `!!`.

The fix adds the parameter→receiver back-edge for out/ref arguments (both `out var x` declarations and pre-declared targets). The receiving local inherits the parameter's taint and the existing fixpoint carries it onward — in the common Try-shape the downstream callee parameter is promoted to `T?` too, so the whole chain agrees with zero added `!!`.

Tests: new `Issue3501OutVarTaintTranslationTests` (tainted Try out-var propagates through the chain and round-trips; an always-assigned out param and its receiver stay bare). Full oblivious/nullable/taint/golden sweep (391 tests) passes.

https://claude.ai/code/session_01Six2MrhXHo6kQ8DSA6c1Pc

### DIFF
--- a/tools/cs2gs/Cs2Gs.Tests/Issue3501GuardedArgumentTaintTranslationTests.cs
+++ b/tools/cs2gs/Cs2Gs.Tests/Issue3501GuardedArgumentTaintTranslationTests.cs
@@ -1,0 +1,127 @@
+// <copyright file="Issue3501GuardedArgumentTaintTranslationTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using Cs2Gs.CodeModel.Ast;
+using Cs2Gs.CodeModel.Printing;
+using Cs2Gs.CodeModel.RoundTrip;
+using Cs2Gs.Translator;
+using Cs2Gs.Translator.Loading;
+using Microsoft.CodeAnalysis;
+using Xunit;
+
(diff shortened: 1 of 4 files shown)


DavidObando:
Second commit on this branch: the taint walk also missed target-typed creations (`new(null, "…")` is `ImplicitObjectCreationExpressionSyntax`, a sibling of the matched `ObjectCreationExpressionSyntax`), which is the whole 30-error GS0155 nil→string family in `RoslynAnalyzerApiMap.gs`. Now matching the shared `BaseObjectCreationExpressionSyntax`, with a regression test mirroring the map's `[key] = new(null, "…")` shape.

Issue #3523, opened 2026-08-25
gsc: unmanaged function-pointer field cannot be invoked directly (GS0159)


obselate:
### Summary

A raw unmanaged function pointer can be invoked when stored in a local variable, but the same function-pointer type cannot be invoked directly through a struct field. Member-call binding reports GS0159 as if the field name were a method name.

### Minimal repro

```gsharp
package FindingFunctionPointerFieldInvocation

unsafe struct Dispatch {
  var Apply unmanaged[Cdecl] (int32) -> int32
}

unsafe func Main() int32 {
  let dispatch = Dispatch{}
  return dispatch.Apply(41)
}
```

Actual:

```text
error GS0159: Cannot find function Apply.
```

Expected: the field load is used as the callee of a `BoundFunctionPointerInvocationExpression`. The repro only needs to build. Running this zero-initialized pointer would be invalid and is not part of the test.

### Works

This local-variable form builds on the same compiler:

```gsharp
unsafe func Main() int32 {
  let apply unmanaged[Cdecl] (int32) -> int32 = nil
  return apply(41)
}
```

### Suspected source

`OverloadResolver.CallBinding.cs` creates a `BoundFunctionPointerInvocationExpression` only when the callee symbol is a `VariableSymbol`. The member-invocation paths in `ExpressionBinder.Calls.Invocation.cs` have delegate field/property fallbacks but no equivalent function-pointer field path.

Related but distinct: #2185 fixed non-identifier managed function-typed expressions, and #2925 fixed delegate-typed interface properties.

### Environment

- Gsharp.NET.Sdk 0.4.59, net10.0, linux-x64.
- Also reproduced with locally packed Gsharp.NET.Sdk 0.4.210 from current `origin/main` commit `bdd42dae19222d8c610939c874dec8d6edd50599`.

PR #3569, opened 2026-08-26
gsc: harden function-pointer member calls after #3568

DavidObando:
## Summary

Follow-up to #3568. Closes all three post-merge blockers at shared paths:

1. `BindFunctionPointerInvocation` now rejects managed pointer calls outside an unsafe context with GS0404 at the callee. Qualified, implicit, property, indexer, static, and arbitrary-expression callees share the gate. Raw `unmanaged[CC]` calls retain ADR-0095 behavior.
2. `FunctionPointerTypeSymbol` recursively substitutes parameter/return types while preserving managed state or unmanaged calling convention. `Binder`, `StructSymbol`, and `InterfaceSymbol` route through it, including by-ref/raw-pointer wrappers. Constructed struct/class/interface instance/static field/property loads now expose closed pointer signatures.
3. Callable-member fallback now applies normal read accessibility using actual declaring owners: field `Accessibility`, property `GetterAccessibility`, including interface statics. Shared delegate fallback receives the same fix.

## Coverage

- managed unsafe-gate diagnostics: qualified/implicit/property/indexer/static/arbitrary callees; unmanaged control
- generic `Dispatch[int32]`/class/interface instance and static fields/properties; good argument/result and bad-string GS0155
- generic method pointer parameter/return substitution
- managed/unmanaged ABI plus by-ref/pointer shape preservation
- private/protected/internal access from owner, derived, unrelated callers; private interface statics; delegate regression
- runtime managed generic calls, emitted closed `calli` signature blobs, and ILVerify invocation

## Validation

Release, rebased on merged #3568 (`ce9181f56`):

- full solution build: 0 warnings/errors
- related Core tests: 59 passed
- related Compiler emit tests: 28 passed
- full Core.Tests: 8,112 passed
- customary compiler-remainder shard: 1,155 passed
- repository ILVerify gate: 125/125 verified
- nullable hygiene: passed before rebase; no nullable-affecting rebase changes

All nullable-hygiene-listed guards are behavior-bearing by design: managed-call unsafe gating, callable match/owner selection, and interface accessibility/owner identity. None exist only to silence annotations.

Refs #3568
Refs #3523

### DIFF
--- a/src/Core/CodeAnalysis/Binding/Binder.cs
+++ b/src/Core/CodeAnalysis/Binding/Binder.cs
@@ -6005,6 +6005,24 @@ public sealed class Binder
                 : type;
         }
 
+        if (type is ByRefTypeSymbol byRef)
+        {
+            var pointee = SubstituteType(byRef.PointeeType, substitution, mapClrType);
+            return ReferenceEquals(pointee, byRef.PointeeType) ? type : ByRefTypeSymbol.Get(pointee);
+        }
+
+        if (type is PointerTypeSymbol pointer)
+        {
+            var pointee = SubstituteType(pointer.PointeeType, substitution, mapClrType);
+            return ReferenceEquals(pointee, pointer.PointeeType) ? type : PointerTypeSymbol.Get(pointee);
+        }
+
+        if (type is FunctionPointerTypeSymbol functionPointer)
+        {
+            return functionPointer.Substitute(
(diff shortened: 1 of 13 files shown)


DavidObando:
Superseded by a fresh-head follow-up because this reused post-merge branch never initialized mergeability or CI checks.

Issue #3535, opened 2026-08-25
cs2gs/gsc: dual-interface explicit member collisions drop the second member (semantic loss)


DavidObando:
Found while greening the self-migration corpus (#3501). A type implementing two same-shaped interfaces with same-NAMED members of different types — the `Issue1483DualMapFixture` in `test/Core.Tests/CodeAnalysis/Binding/Issue1483ReadOnlyDictionaryForRangeTests.cs` implements both `IDictionary<int, int>` and `IReadOnlyDictionary<string, string>` — hits the #1911-style forced-public fallback: the **second colliding explicit-interface property/indexer/event is dropped entirely** (`Keys`, `Values`, `Count` in the fixture), with a `CS2GS` review diagnostic noting the semantic loss.

Until #3534 the dropped member's null even crashed the whole document's print (fixed — the drop now degrades gracefully), but the drop itself remains a genuine gap: the migrated type no longer satisfies the second interface's slots with the right member, so any consumer dispatching through that interface observes the wrong behavior. This is a **test-parity risk** under the #3501 quality bar once Core.Tests runs its migrated form.

What's needed (design sketch): extend the ADR-0149 explicit-interface clause scheme so two same-named members can coexist when their interface qualifiers differ — e.g. `prop (IDictionary[int32, int32]) Keys …` and `prop (IReadOnlyDictionary[string, string]) Keys …` as distinct members with distinct emitted slots (the CLR side is ordinary explicit-impl metadata; the gap is G#'s surface + gsc's name-collision model treating the pair as one).

Scope of impact today: the corpus has exactly this one fixture shape, but any real-world dual-implementer (a collection exposing both mutable and read-only contracts) hits it.

🤖 Filed by Claude Code from session https://claude.ai/code/session_01VgrKRVPGaEm9zLaYvankA1

PR #3586, opened 2026-08-27
cs2gs: round-trip printer fixes from the #3501 gate widening (tuple-prop arrow, control-char raw strings, nested unbound typeof)

DavidObando:
## Summary

Part of the #3501 slice "widen the migration gate beyond the 3-app chain". Burns down the `src/LanguageServer` translate wall.

`prop Stats (int64, int64) -> (hits, misses)` is ambiguous in the G# grammar: the type position consumes `(int64, int64) -> ...` as a function-**type** annotation, swallowing the arrow body, so the emitted file fails cs2gs's round-trip parse gate (GS0005 `Unexpected token <OpenParenthesisToken>`). This is the exact `SemanticLookup.cs` failure (fingerprint `9dd020ef25f2`) that kept `src/LanguageServer` red in the nightly self-migration — its `(long, long)` cache-stat properties.

Fix (printer-side): when a property's rendered type starts with `(`, emit the ADR-0131 arrow **accessor** form inside a block —

```gs
internal prop NodeBucketCacheStats (int64, int64) {
    get -> (Interlocked.Read(&nodeBucketCacheHits), Interlocked.Read(&nodeBucketCacheMisses))
}
```

— where the type ends unambiguously at the `{`. Non-parenthesized property types keep the flat arrow spelling. Applies to indexers via the shared expression-body branch. (Arrow-bodied *functions* returning tuples are unaffected — probe-verified `func Pair() (int64, int64) -> (a, b)` parses and runs.)

## Verification

- 4 new tests in `Issue3501TuplePropertyArrowAmbiguityTests` (render shape for prop + indexer, emitted-oracle compile of the translated form, non-tuple prop keeps arrow) — all pass.
- Targeted `src/LanguageServer` migration: translate FAIL(1) → **PASS** (remaining compile errors in the targeted run are the expected `--exclude`d-Core dependency cascade).

### DIFF
--- a/tools/cs2gs/Cs2Gs.CodeModel/Printing/GSharpPrinter.cs
+++ b/tools/cs2gs/Cs2Gs.CodeModel/Printing/GSharpPrinter.cs
@@ -1047,13 +1047,19 @@ public static class GSharpPrinter
             sb.Append('\n');
             sb.Append(armPad);
             string marker;
-            if (arm.Pattern == null)
+            if (arm.Pattern == null && arm.Guard == null)
             {
                 marker = "default:";
             }
             else if (arm.Guard != null)
             {
-                marker = $"case {RenderPattern(arm.Pattern, indent + 1)} when {RenderExpression(arm.Guard, indent + 1)}:";
+                // A guarded discard (`_ when g =>`) must keep its guard: G#'s
+                // spelling is `case _ when g:` — rendering it as `default:`
+                // dropped the guard (a silent behavior change) AND collided
+                // with the synthesized total arm (GS0169, the GeneratorHost
+                // wall in #3501).
+                string guardedPattern = arm.Pattern == null ? "_" : RenderPattern(arm.Pattern, indent + 1);
+                marker = $"case {guardedPattern} when {RenderExpression(arm.Guard, indent + 1)}:";
             }
             else
             {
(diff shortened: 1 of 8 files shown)


DavidObando:
Extended the fix to the second shape of the same ambiguity found in test/Core.Tests: a type whose rendering **ends** in a nested paren group (`[](string, string)`) followed by `-> expr` lets the nested type position absorb the arrow. Methods fall back to a block body (top-level tuple returns keep the flat arrow — gsc's `LooksLikeTupleReturnTypeBeforeArrowBody` disambiguates exactly that shape and Issue1278's golden asserts it); properties widen the arrow-get-accessor fallback to types ending with `)`. Clears the `Issue2395FileScopedImportTests` round-trip wall (GS0005 Unexpected `<IfKeyword>`). 7 tests in the suite now, plus Issue1278 + GoldenTests green.

Side finding for a gsc issue: `prop Rows [](string, string) -> [](string, string){("a", "b")}` currently **compiles silently** as a nil-valued function-typed property and NREs at runtime — no diagnostic.

DavidObando:
Two more round-trip fixes on this branch (all from the same Core.Tests burn-down):

- **Control chars in raw strings**: multiline values holding control characters other than newline/tab keep the escaped one-line form — an embedded NUL in a backtick raw literal lexes as EOF (GS0003, the LexerTests wall).
- **Nested unbound-generic typeof**: `typeof(Dictionary<,>.Enumerator)` / `typeof(Outer<>.Inner<>)` now render `_` placeholders per segment (`Dictionary[_, _].Enumerator`) instead of emitting placeholder types or leaking raw `<>` (the Issue2851 GS0005 wall). Parse-only until gsc lands #3589; the wall moves from an opaque translate gap to an honest compile diagnostic.

Targeted Core.Tests translate wall with all three fixes: 9 → 6 remaining (2× ADR-0117 three-value TheoryData rows, 2× #3535 dual-interface fixture, 1× nested-typeof compile pending #3589, 1× unsafe function pointer). GoldenTests + Oahu + the open-generic typeof suites green.

PR #3591, opened 2026-08-27
selfmig gate: exclude the net472 VS-extension apps by policy (#3501)

DavidObando:
## Summary

Part of the #3501 gate-widening slice. `src/vs-gsharp/src/VsGsharp` and `src/vs-gsharp/src/VsGsharp.CodeLens` are net472 VS-extension projects built by the Windows-only VSSDK toolchain (VSCT compile, pkgdef). On the Linux nightly they fail `ProjectLoad` outright (`The reference assemblies for .NETFramework,Version=v4.7.2 were not found`), and even with `Microsoft.NETFramework.ReferenceAssemblies` a migrated G# build would have no VSSDK toolchain to compile against — they are not meaningful migration targets, same as the excluded e2e sample fixtures.

`VsGsharp.UnitTests` (net10.0) **stays** in the corpus — its remaining wall is StreamJsonRpc type resolution on linked sources, which is fixable.

This shrinks the gate denominator from 55 to 53 deliberately-scoped apps.

### DIFF
--- a/build/run-cs2gs-selfmig.sh
+++ b/build/run-cs2gs-selfmig.sh
@@ -16,6 +16,11 @@
 # baseline within the same PR so progress ratchets.
 #
 # E2E fixtures that are not migration targets are excluded via --exclude.
+# The two net472 VS-extension apps (VsGsharp, VsGsharp.CodeLens) are excluded
+# by policy: they build only under the Windows-only VSSDK toolchain (VSCT
+# compile, pkgdef), so the Linux gate cannot even load them (no net472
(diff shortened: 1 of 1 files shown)


DavidObando:
Added a second corpus-hygiene exclusion to this branch: `tools/cs2gs/corpus/CompileGap-Library` — a fixture whose purpose is to FAIL compilation (it calls `UndefinedTriageHelper` by design, for triage testing). It permanently pinned one red app in the gate count. Denominator: 55 → 52.

Issue #2455, opened 2026-07-18
gsc: composite literal projects to wrong same-named type across packages


DavidObando:
Oahu.Core AaxExporter constructs `ContentMetadata{ChapterInfo: ci, ...}` where `ci` is correctly `Oahu.Audible.Json.ChapterInfo`, but gsc reports GS0490 attempting to project `Oahu.BooksDatabase.ChapterInfo` to bare `ChapterInfo`. Both projects define ChapterInfo with incompatible `Chapters` member types (`[]Chapter` vs `ICollection<Chapter>`).

The target field/property type is fully known from ContentMetadata, yet composite-literal assignment/projection appears to re-resolve by simple name across the compilation and selects the wrong package type. Add a two-/three-package same-simple-name repro and fix identity to use the declared target TypeSymbol, not global simple-name lookup. Cover object/struct literals, nested/generic/array members, source/imported types, field/property/constructor assignment, ordering independence, ambiguity diagnostics, runtime/reflection, and ILVerify. Exact fingerprint: GS0490 compile-e195f207485c at AaxExporter.gs:46/53.

PR #3596, opened 2026-08-27
gsc: a package sees referenced assemblies' types in its own namespace (#3595)

### LINKED ISSUE #3595: gsc: a package does not see referenced assemblies' types in the same namespace
#3501 self-migration wall for the cs2gs tool family (Cs2Gs.ProjectLoading and downstream Cli/Pipeline/Report — ~22 distinct errors): CLR namespaces merge across assemblies, but a G# `package X` file cannot see a REFERENCED assembly's exported types in namespace X without an explicit `import X`.

## Repro (10 lines)

```gs
// lib1.gs → gsc /target:library /out:lib1.dll
package Shared.Loading

open class Doc {
    prop Path string -> "p"
}
```

```gs
// lib2.gs → gsc /target:library /reference:lib1.dll  → GS0113: Type 'Doc' doesn't exist
package Shared.Loading

class Holder {
    func Describe(d Doc) string {
        return d.Path
    }
}
```

An explicit `import Shared.Loading` inside `package Shared.Loading` works, as does consuming from a different package with the import — only the same-package implicit merge is missing.

This is exactly the migrated-repo shape: `Cs2Gs.ProjectLoading` declares types in `Cs2Gs.Translator.Loading` (the referenced Translator assembly's namespace), so every cross-assembly reference in those files is GS0113 (`LoadedDocument`, `TranslationContext`, `TranslateDocument`, ...).

## Fix shape (validated)

A compilation-wide implicit `import <package>` seed is WRONG — it breaks per-file import disambiguation (#2455/#2456 suites, 28 failures). The scoped fix: a **fallback** in `BoundScope.TryLookupImportedGenericClass` / `TryLookupImportedClassByArity` — after every explicit import fails, try `<current declaring package>.<name>` against References. Runs only when nothing else resolved, so it cannot introduce ambiguities or change source-package disambiguation. Import/Package/Alias suite (856 tests) green with the fallback in place.

Refs #3501.

DavidObando:
## Summary

Fixes #3595 — the widest remaining wall in the #3501 self-migration gate: ~22 distinct GS0113/GS0158/GS0159 errors in `Cs2Gs.ProjectLoading` (gating the whole cs2gs tool family: Cli, Pipeline, Report) all reduce to one missing language behavior. CLR namespaces merge across assemblies; a G# `package X` file could not see a referenced assembly's exported namespace-X types without an explicit `import X`. The migration splits `Cs2Gs.Translator.Loading` across two projects, so every cross-assembly reference (`LoadedDocument`, `TranslationContext`, …) failed.

## Approach

A compilation-wide implicit `import <package>` seed was tried first and is **wrong** — it breaks per-file import disambiguation (28 failures across the #2455/#2456 suites, "ambiguous between two or more imported packages"). The landed fix is a scoped **fallback** in `BoundScope.TryLookupImportedGenericClass` / `TryLookupImportedClassByArity`: after every explicit import fails to resolve the name, try `<current declaring package>.<name>` against the reference resolver. It runs only when nothing else resolved, so it can green previously-unresolvable names but cannot introduce new ambiguities or change source-package disambiguation.

## Verification

- 10-line two-assembly repro (`package Shared.Loading` in both, no import) compiles.
- New `Issue3595SelfPackageImportedTypeTests`: non-generic and generic imported types resolved cross-assembly with no import spelled.
- 856-test Import/Package/Alias `Core.Tests` sweep green (the same sweep that caught the wrong global-seed approach).

### DIFF
--- a/src/Core/CodeAnalysis/Binding/BoundScope.cs
+++ b/src/Core/CodeAnalysis/Binding/BoundScope.cs
@@ -688,6 +688,21 @@ public sealed class BoundScope
             }
         }
 
+        // Issue #3595 (#3501): CLR namespaces merge across assemblies, and a
+        // G# package is the same construct — a file in `package X` sees a
+        // REFERENCED assembly's exported types in namespace X without
+        // spelling `import X` (the C#→G# migration splits one namespace
+        // across project boundaries). A fallback AFTER every explicit import
+        // has failed: it can resolve previously-unresolvable names but never
+        // introduces a new ambiguity or changes source-package
+        // disambiguation.
+        if (GetCurrentDeclaringPackage() is { Length: > 0 } declaringPackage
+            && References.TryResolveType(declaringPackage + "." + mangled, out var selfPackageType))
+        {
+            type = selfPackageType;
+            return true;
+        }
+
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


DavidObando:
End-to-end verification: serialized targeted self-migration of the chain (src/Core → Cs2Gs.CodeModel → Cs2Gs.Translator → **Cs2Gs.ProjectLoading**) against a freshly packed SDK carrying this fix — **4/4 apps green on all four stages** (translate, compile, ilverify, test-parity). ProjectLoading was FAIL(22-distinct) before the fix. With this merged, Cli/Pipeline/Report should follow — their build errors all cited ProjectLoading.gsproj.

PR #3597, opened 2026-08-27
gsc: satisfy an imported-class constraint from a source derived class (#3501)

DavidObando:
## Summary

Burns down the `src/Analyzers/GSharp.CodeAnalysis.Analyzers.Testing` wall in the #3501 self-migration gate (4 reported errors; 3 were cascades of a single poisoned local). Two composed gsc gaps around a class constraint whose base lives in a **referenced assembly** — exactly the migrated-repo shape, where `GSharpDiagnosticAnalyzer` comes from the migrated Core package:

1. **`SatisfiesClassConstraint`** fell to CLR assignability for an imported constraint, but a SOURCE class has no `ClrType`, so `class Derived : ImportedBase` failed `[T ImportedBase init()]` with GS0152. Now walks the source base chain to the first `ImportedBaseType` and checks CLR assignability there.
2. **`TryProjectErasedClrType`** erased every type-parameter-typed argument to bare `object`, so `ImmutableArray.Create[ImportedBase](T())` had no applicable candidate (GS0159). A type parameter with a CLR-backed class constraint now erases to that class — sound for gating (any `object`-typed candidate still accepts the constraint class), and C#-consistent for ranking.

## Verification

- Two-assembly repro (imported `Base`, source `Derived : Base`, `Runner[T Base init()]` calling `ImmutableArray.Create[Base](T())` + LINQ over the result) compiles clean; both failure modes probe-isolated first.
- New `Issue3501ImportedBaseConstraintTests` (two-assembly emit harness).
- 1410-test Overload/Constraint/Generic `Core.Tests` sweep green.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/Binder.cs
+++ b/src/Core/CodeAnalysis/Binding/Binder.cs
@@ -6295,9 +6295,29 @@ public sealed class Binder
 
         // Imported reference class: use CLR assignability when both project to a
         // CLR type.
-        if (classConstraint.ClrType is { } constraintClr && typeArgument.ClrType is { } argClr)
+        if (classConstraint.ClrType is { } constraintClr)
         {
-            return constraintClr.IsAssignableFrom(argClr);
+            if (typeArgument.ClrType is { } argClr)
+            {
+                return constraintClr.IsAssignableFrom(argClr);
+            }
+
+            // Issue #3501: a SOURCE class deriving (possibly through source
+            // bases) from an imported CLR base has no ClrType of its own —
+            // walk the source base chain to the first ImportedBaseType and
+            // check CLR assignability there (`class Derived : ImportedBase`
+            // must satisfy `[T ImportedBase]`).
+            if (typeArgument is StructSymbol sourceClass)
+            {
+                for (var current = sourceClass; current != null; current = current.BaseClass)
+                {
+                    if (current.ImportedBaseType?.ClrType is { } importedBaseClr
+                        && constraintClr.IsAssignableFrom(importedBaseClr))
+                    {
+                        return true;
+                    }
+                }
+            }
         }
 
         return false;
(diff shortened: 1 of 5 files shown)


DavidObando:
CI failure root-caused and corrected. The erasure widening in `TryProjectErasedClrType` was wrong — it leaked MetadataLoadContext types into runtime generic composition, breaking LINQ receiver projection for a `T` nested in a tuple (`LinqWhere_PreservesNestedSymbolicTupleReceiver`) and the pinned Oahu gate. Reverted.

The actual gap was a missing wiring: `ImportedClassSymbol.TryLookupFunction` (the static imported-call path) never passed `structuralProjectionArgumentCheck` — the ADR-0148 symbol-aware applicability fallback the instance/extension paths already use. With it wired, the `object`-erased argument re-checks symbolically and `ClassifyNonStructural(T, Base)` proves the constraint-based conversion (verified via temporary instrumentation: `src=TypeParameterSymbol:T clrParam=IC.Analyzers.Base implicit=True`).

The constraint-satisfaction half (source class deriving an imported base) is unchanged. Previously-failing CI test + full 1833-test Overload/Constraint/Generic/Import sweep green locally.

DavidObando:
The `compiler-issue2-remainder` failure was the #3178 rejection guard `...WithDelegateInConstructedGenericAcrossReference_IsRejected`: it pinned GS0159 for a shape that was rejected *via the constraint failure this PR fixes*. Verified out-of-suite that the shape now closes, **ilverifies clean, and runs with correct semantics** (the configure delegate mutates the constructed T — output `9`; my first ilverify attempt showed phantom errors from missing `-r` references, not real IL faults). Flipped the guard into a behavior test using the file's existing `CompileAndRun` harness; the class's other rejection guards are untouched. 18/18 in the class green locally.

Issue #3747, opened 2026-09-01
gsc: `nil` rejected as an element of a nullable `params` array (GS0155, from #3745)


DavidObando:
Split out of #3745 (family F3). Found in the migrated `test/Compiler.Tests` compile wall; 2 of that app's 20 errors.

When an imported method's `params` array is itself **nullable** (`params object?[]? args`), the element type loses its nullable annotation, so passing `nil` in an expanded params position is rejected with `GS0155 Cannot convert type 'nil' to 'object'` — note the target prints as `object`, not `object?`.

## Repro (gsc, `0.4.349+cec369958b`)

```gs
import System

func main() {
    let d Func[object?, object?, object?] = func (a object?, b object?) object? {
        return a
    }
    let del Delegate = d
    Console.WriteLine(del.DynamicInvoke("x", nil))
    // error GS0155: Cannot convert type 'nil' to 'object'.
}
```

`Delegate.DynamicInvoke` is declared `object? DynamicInvoke(params object?[]? args)`.

Contrast — a params array that is **not** nullable works today:

```gs
Console.WriteLine(string.Format("{0} {1}", "x", nil))   // binds; params object?[]
```

so the defect is isolated to the nullable-array form: unwrapping `object?[]?` to get the element type appears to drop the element's `?`.

## Current evidence in the corpus

`test/Compiler.Tests/Emit/Issue2398ExpressionTreeLiftedUserOperatorEmitTests.cs`, migrated:

- `Emit/Issue2398ExpressionTreeLiftedUserOperatorEmitTests.gs:52,63` — `Assert.False(bool(compiled.DynamicInvoke(present, nil)!!))`
- `Emit/Issue2398ExpressionTreeLiftedUserOperatorEmitTests.gs:91,57` — `Assert.Null(compiled.DynamicInvoke(present, nil))`

## Not a regression

Adjacent to but distinct from the solved signature-position nullability family (#3705 family 2 / PR #3741): that one was about nullability in signature positions, this one is about the element type of a nullable `params` array during expansion.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Nng28yiBdPVdeML7mSZphs

DavidObando:
**Scope correction from PR #3758.** The isolation stated in this issue is wrong, and the correct scope is wider.

[...]

Issue #3705, opened 2026-08-30
Audit: "inconsistent sibling probe" defect class — 7 bugs this week shared one shape; find the rest and prevent recurrence


DavidObando:
## The pattern

Seven defects found during the #3501 self-migration in a single week turned out to share one shape: **a reflection/metadata probe omits something that a sibling probe a few lines away already does correctly.** None was a missing feature; each was an inconsistency between neighbours. Fixing one repeatedly failed to fix the others, because each site was patched in isolation.

| Fix | The omission | The sibling that was already right |
|---|---|---|
| #3693 | `SafeGetMethodsIncludingSelfAndInterfaces` and the imported static-class probe enumerated `BindingFlags.Public` only, so friend-assembly `internal` **methods** were never candidates (GS0159) | the **property** and **field** probes already widened to `NonPublic` under the identical `CanAccessInternalMembers` test |
| #3702 | `TryGetWritableClrMember` used `GetSetMethod(nonPublic: false)`, so a friend-visible `{ get; internal set; }` was "unwritable" (GS0127) — the **third** instance of the same omission | the two probes #3693 had just fixed |
| #3703 | `TryGetDelegateFunctionType` read nullability for every **parameter** position but used bare `FromClrType` for the **return** | the parameter loop, one line above |
| #3703 | `GetClrEventHandlerTypeSymbol` dropped declaration nullability while substituting the handler type | the sibling **field** path a few lines above already folded the flags in via `MergeDeclarationNullability` |
| #3680 | `TryBindImportedExtensionCall` ran every **user argument** through the effective-CLR-type helper but filled the **receiver** slot from raw `ClrType`, losing `Nullable<T>` inference | the argument loop in the same method |
| #3667 | the symbolic-path success gate counted symbolic info from a lambda **parameter** type or a **fully closed** return, but not a return grounded in the receiver's symbolic arguments | the parameter arm of the same gate |
| #3697 | `IsValueReturningDelegateToVoidDelegate` reflected `Invoke` off delegate types directly, throwing under `MetadataLoadContext` | its **two sibling delegate predicates** already used the cross-context `TryGetDelegateSignature` decomposition |

Related-but-distinct (same "one arm updated, the other not" flavour): #3666, where one of two array arms of `Conversion.ClassifyCore` offered only a downcast predicate as its explicit fallback.

## Why this matters

[...]

Issue #3704, opened 2026-08-30
REGRESSION: `object?` to `bool` in migrated Cs2Gs.Translator cascades into 7 apps; gate 33/51 below floor 38


DavidObando:
Run [33331139589](https://github.com/DavidObando/gsharp/actions/runs/33331139589) fails the gate:

```
33/51 green (floor 38); labels=0 (ceiling 0); __local_=29 (ceiling 30);
lines>300=563 (ceiling 560); bangs=17556 (ceiling 17400)
GATE: green count 33 fell below floor 38.
GATE: >300-char line count 563 exceeded ceiling 560.
GATE: !! null-assertion count 17556 exceeded ceiling 17400.
```

The previous measured run [33308959087](https://github.com/DavidObando/gsharp/actions/runs/33308959087) was **38/51 with all ceilings satisfied** (lines 554, bangs 17033).

## Primary cascade

```
GS0156 Cannot convert type 'object?' to 'bool'. An explicit conversion exists (are you missing a cast?)
  in migrated tools/cs2gs/Cs2Gs.Translator — line 330, col 20
```

This single error is the **only** failure in 7 apps: `Cs2Gs.Cli`, `Cs2Gs.Pipeline`, `Cs2Gs.ProjectLoading`, `Cs2Gs.Report`, `Cs2Gs.Tests`, `Cs2Gs.Translator`, `GSharp.GeneratorHost`, `GSharp.GeneratorHost.Tests`, `Gsgen.Cli` — every app that compiles migrated `Cs2Gs.Translator`. Same single-point cascade shape as #3663 and #3666.

An `object?` reaching a `bool` position means something that used to be typed non-nullable (or more precisely) is now `object?`. Prime suspect: **#3696** (PR for #3682, "nullable element type for an array/slice literal that writes nil") — it widens an inferred literal element type to `T?`, and an element that becomes `object?` and is later consumed as a condition produces exactly this. Secondary suspects: #3690 (generated-code declaration promotion), #3697 (inference bound fixing), #3688 (array covariance conversions).

The two ceiling breaches are consistent with the same family: more nullable spellings (longer lines, 554 → 563) and more null-assertions (17033 → 17556).

## Merges to bisect

[...]

PR #3763, opened 2026-09-01
gsc: an imported `string?` bound as non-null `string` through a constrained receiver (#3705 family 2)

DavidObando:
**An imported `string?` read through a constrained generic receiver bound as non-null `string`, with no diagnostic.** Not against a synthetic fixture — against the fully-annotated BCL. `LinkedList<T>.First` is declared `[Nullable(2)]` (it returns `null` for an empty list) and gsc handed it to callers as non-null:

```gs
// imported, nullable-enabled: string? NullableProperty { get; }  on interface IC
func Via[T IC](v T) string { return v.NullableProperty }   // bound clean on main
```

`MemberLookup.GetClrPropertyTypeSymbol`'s symbolic branch ended in a bare `return mapped;` — receiver substitution recovers the member's *shape* and silently discards its `[Nullable]` metadata. #3741 called that exact line out by name as *"the defective line … contains no `FromClrType` at all"*, the shape a grep cannot see, and then fixed two of its three occurrences: the method-return and parameter branches. This is the third.

It stayed invisible because the differential matrix's axis was wrong. It varied *member kind* while holding receiver shape ≈ plain, so "constrained receiver" was only ever probed through a method return. The axis is now (member kind × **receiver shape**), and the two new kinds fail on `main` in the annotated-`Nullable` column — the unsound direction, not the noisy one.

This PR is family 2 of the #3705 audit — the half #3741 deferred as *"a decision, not a swap"*. The deferred question was about the oblivious default; answering it properly turned up the unsoundness above, two further defects, and the one position where the #1354 rule must **not** apply.

## 2. The deferral's own description of the defect was incomplete

> `MergeDeclarationNullability` opens with `if (declarationFlags.IsDefaultOrEmpty) { return projectedType; }` … Closing it is one `return` away.

It is two. `MergeDeclarationNullability` annotates through `ApplyRootAnnotation`, which read `if (flag != Annotated) return projected;` — **C#'s** rule ("nullable iff the byte is `2`"). G#'s rule (#1354 / ADR-0136) is the inverse ("non-null iff the byte is `1`"). They agree on `1` and `2` and disagree on `0`, and `0` is not hypothetical: csc emits an explicit `[Nullable(0)]` for a `#nullable disable` member of a `[NullableContext(1)]` type. Probed:

```
TYPEATTR  NullableContextAttribute 1
FIELD A     flags=[1]  ->  string    (annotated non-null)
FIELD Zero  flags=[0]  ->  string?   (#nullable disable member, EXPLICIT [Nullable(0)])
```

`flags=[0]` is **non-empty**, so the empty-flags short-circuit never saw it. Deleting that one `return` would have left this shape divergent.

## 3. …and one position where #1354 must NOT apply

Making every position obey #1354 broke 12 tests across 6 CI shards, all one defect: `map[K, V].Keys` iterated as `K?`. That is wrong, and not merely noisy — an unconstrained `K` may be substituted with a **value type**, where `K?` silently means `Nullable<K>`.

So the carve-out is principled: **#1354 is a rule about concrete reference positions in imported metadata.** An OPEN type-parameter position is not one — its nullability arrives with the type **argument** at substitution, which carries its own annotation. An open parameter slot keeps the pre-#1354 reading: widen only for an explicit `[Nullable(2)]`.

This has a sharp edge worth recording, because the next person will otherwise rediscover it the hard way. `ExpandNullableFlags` fills absent positions with `2` — that fill **is** the #1354 rule — so the expanded array cannot distinguish "declared `T?`" from "declared nothing, defaulted". An `absentFill` overload gives the literal reading; only the carve-out uses it. Without that, 3 of the 12 stayed red.

## 4. The decision: uniformity over concrete positions, and ADR-0136 already required it

ADR-0136 §2 states the rule and its intended blast radius:

> **only an explicit `1` (NotAnnotated) means non-null `T`; `2` (Annotated), `0` (oblivious), and an absent attribute all mean nullable `T?`.**

> Only genuinely unannotated/oblivious assemblies — **and oblivious positions within otherwise-annotated assemblies** — become nullable.

The rule is applied "uniformly in **both** reading paths". Receiver substitution is a **third** path, added after the ADR, that never got it. The divergent set was never a semantic grouping — it was exactly the readers routing through `MergeDeclarationNullability`, plus the one that skipped it. Mechanism, not intent.

### The measured matrix — `main` / with fix

| kind | ann. non-null | ann. nullable | ObliviousAbsent | ObliviousZero `[0]` |
|---|---|---|---|---|
| Property / MethodReturn / IndexerElement | non-null / same | nullable / same | nullable / same | nullable / same |
| Field / ConstrainedMethodReturn / DeconstructOut | non-null / same | nullable / same | **non-null** → nullable | **non-null** → nullable |
| **ConstrainedProperty / ConstrainedIndexer** | non-null / same | **non-null** → nullable | **non-null** → nullable | **non-null** → nullable |

## 5. Fixed

| Site | Symptom |
|---|---|
| `MemberLookup.GetClrPropertyTypeSymbol` — symbolic branch | **imported `string?` property/indexer bound non-null through a constrained receiver** |
| `NullableFlagsBuilder.MergeDeclarationNullability` — empty-flags short-circuit | oblivious-with-no-metadata kept the pre-#1354 answer |
| `NullableFlagsBuilder.ApplyRootAnnotation` — `flag != Annotated` | explicit `[Nullable(0)]` kept the pre-#1354 answer |
| `MergeDeclarationNullability` — generic-parameter arm | *carve-out*: open `K` must not become `K?` |

Funnel, per #3707's `ClrMemberVisibility` and #3716's `ClrLoadContext`: `ClrNullability.IsFlagNonNull(byte)` is the only place that decides what a nullability byte means. (The surviving `flag != Annotated` in `AppendAnnotatedTail` is an emit-side scalar-compression check asking a different question, and is correct.)

## 6. Two test repairs, both corrections rather than accommodations

- **`Issue2514ImportedInterfaceConstraintMemberTests`** — its C# contracts were nullable-**oblivious**, so every member now imports `T?` while the G# side assigns to non-null `string`. Those assignments were only legal because the constrained-receiver readers dropped nullability. The fixture gets `#nullable enable`, stating the contract the test always meant. Chosen over `!!` deliberately: `!!` would make the fixture depend on the very gap being closed, and uncommented `!!` is a review defect under ADR-0155.
- **`Issue1418…LinkedListNodeProperty_UserElement_ChainedAccessBindsAndRuns`** — gains a `nil` guard. This is the live BCL instance of §1: `LinkedList<T>.First` is `[Nullable(2)]`, so the guard is what the declaration always required. Its actual subject — that a constructed generic property type does not erase to `LinkedListNode[object]` — is unchanged and still asserted.

## 7. Gate

`Issue3705MemberKindNullabilityDifferentialTests`: 19 rows → **8 kinds × 4 states + 3 = 35**, with **zero pinned exceptions**. #3741 wrote *"when the rule is settled and the hole closed, this test fails and forces the review"*; this is that review, and the pin is deleted rather than re-pointed.

- `ConstrainedProperty` / `ConstrainedIndexer` make the axis (member kind × **receiver shape**) — the blind spot that hid §1.
- `ObliviousZero` is a first-class state; anti-vacuity pins it to byte `[0]` on **both** the surface class and the constrained interface, since csc's `[NullableContext]` vote is per-type and either could silently decay into `ObliviousAbsent`.
- `OpenTypeParameterPositions_Take_Nullability_From_The_Argument` pins §3 with a reference-type **and** a value-type argument, both executed.
- **Every row executes its emitted assembly.** Binding clean is not evidence the IL runs.
- `ObliviousField_That_Is_Null_At_Runtime_Is_Guardable_And_Runs`: an oblivious imported field that really is `null`, read into an **inferred** local (inferred deliberately — with an explicit `string?` it compiles on `main` too and proves nothing), guarded, executed, guard asserted to fire.

## 8. Observable behaviour change

Stated rather than left to be discovered. Two categories are user-visible, both intended:

1. **Source-breaking, desired.** Code reading an imported `string?` property/indexer through a constrained generic receiver compiled silently before and now correctly requires a null check. The old behaviour handed out a `string` that could be null.
2. **Source-breaking, noisier, also intended.** Reading a member of a nullable-**oblivious** assembly through the receiver-substituting readers now yields `T?`. ADR-0136 named this consequence: *"code that … relied on the old oblivious→non-null default now sees `T?` and must null-check or null-forgive. This is the intended, safer behavior."*
3. **Emitted metadata does not move.** G# requires explicit types on function signatures, so public surface carries declared (not inferred) nullability; only inferred locals change, and those contribute no metadata. Checked, not assumed: `Samples_EmittedPE_Match_Baseline` hashes the **metadata stream** — where `[Nullable]` attributes live — plus every method body's IL, and is green with no baseline regeneration.

## 9. Verification

| measurement | result |
|---|---|
| `Core.Tests` (full) | **8559 / 8559** |
| `Compiler.Tests` (full) | **4609 / 4609** |
| CI, final SHA | **35 pass, 0 fail** |
| failing-on-`origin/main` | **12 of 35 rows fail**, 23 pass as labelled guard rails |
| emitted PE baseline | green, no regeneration |
| nullable hygiene (ADR-0155) | clean |

ADR-0136 updated: the third reading path named, `IsFlagNonNull` recorded as the single predicate, the open-type-parameter exclusion documented with both reasons, and the "generic-substituted returns" deferral marked resolved.

## Still open on #3705

- **Base-member group** — `e.StackTrace` on a G# class deriving from an imported base; needs a G# class declaration rather than an expression, so its own matrix row.
- **Element-type sub-pattern** — needs `GetNullableFlagsForSubtree`, not a reader swap. #3758 (#3747) hit two more after #3741 landed; the live sub-family.
- **cs2gs `Deconstruct` out-parameter prediction** — `IsImportedObliviousNullableMember` keys on a member's return/type, so a `void`-returning `Deconstruct` never matches. Translator-side prediction gap, not unsoundness.

Refs #3705.

### DIFF
--- a/docs/adr/0136-nullable-by-default-for-unannotated-imports.md
+++ b/docs/adr/0136-nullable-by-default-for-unannotated-imports.md
@@ -65,7 +65,7 @@ The single principle: **only an explicit `1` (NotAnnotated) means non-null `T`;
 `[NullableContext(1)]` makes **all** nested reference positions non-null, and a
 `[Nullable(2)]` makes them all nullable — matching Roslyn's compaction.
 
-The rule is applied uniformly in both reading paths:
+The rule is applied uniformly in all three reading paths:
 
 - `ApplyReferenceNullabilityFull` — top-level field / property / parameter /
   return reading (previously keyed off `topFlag == 2`).
@@ -74,6 +74,38 @@ The rule is applied uniformly in both reading paths:
   beyond-length positions to `0`). Beyond-length handling now follows the same
   rule: length-1 flags apply `flags[0]` to all positions, empty flags are
   nullable, otherwise index directly.
+- `NullableFlagsBuilder.MergeDeclarationNullability` — the receiver-substituting
+  path (issue #3705 family 2), which folds a declaration's flags back onto a type
+  recovered by generic substitution. It reaches imported fields, constrained
+  calls, `Deconstruct` out-parameters and event handler types.
+
+`IsFlagNonNull(byte)` is the single predicate all three defer to. It exists
+because the two open-codings of this one-line rule are what let the paths drift:
+C# reads "nullable iff the byte is `2`", G# reads "non-null iff the byte is `1`",
+and the two agree on `1` and `2` while disagreeing on `0` and on absence. A
+reader that spells the rule itself will sooner or later spell C#'s.
+
+#### The scope of "reference position": open type parameters are excluded
+
+The rule above is about **concrete** reference positions. An **open type-parameter**
+position is not one, and issue #3705 family 2 makes that explicit after a first
+attempt at uniformity made `map[K, V].Keys` iterate as `K?`:
+
+- A type parameter's nullability arrives with the type **argument** at
+  substitution, and the argument carries its own annotation. Applying the
+  declaration's *missing* byte overwrites the caller's answer with a guess.
+- It is unsound in a way the concrete case is not. An unconstrained `K` may be
+  substituted with a **value type**, where `K?` silently means `Nullable<K>` and
+  changes the contract.
+
+So an open parameter slot keeps the pre-#1354 reading: widen only for an
+explicit `[Nullable(2)]`, which is the declarer genuinely writing `K?`. Absent
+and oblivious leave the projected type alone.
+
+This distinction needs care because `ExpandNullableFlags` fills absent positions
+with `2` — that fill *is* the rule above — so the expanded array cannot tell
+"declared `T?`" from "declared nothing". The `absentFill` overload gives the
+literal reading; only the type-parameter arm uses it.
 
 **Value types are unaffected.** A value-type position contributes no byte and
 stays non-null; `Nullable<T>` value types are already lowered in
@@ -207,3 +239,19 @@ its flag byte), which is a deeper change than this fix warrants. The common case
 (a non-generic instance method returning a plain reference type) is fully fixed;
 the generic-substituted-return nullability is the only remaining call-return gap
 and is tracked here as best-effort follow-up.
+
+**Resolved.** That threading is `NullableFlagsBuilder.MergeDeclarationNullability`,
+which walks the open declaration's layout and the substituted symbol in lockstep
+so each substituted position recovers its flag byte. Issue #3695 applied it to
+fields and event handlers, #3741 (issue #3705 family 2) to constrained-call
+returns and `Deconstruct` out-parameters, and #3705 family 2's second half to the
+oblivious column: `MergeDeclarationNullability` used to return the projected type
+unchanged when the declaration's flags were empty, and to key its annotation off
+`flag == 2` when they were not — so a member reached through a substituting
+receiver kept the pre-#1354 answer for *both* shapes of oblivious. Both are gone;
+the path now defers to `IsFlagNonNull` like the other two.
+
+`Issue3705MemberKindNullabilityDifferentialTests` is the standing gate: six
+signature-position kinds × four annotation states, with the expectation computed
+from the annotation state alone, so the table *is* this ADR's uniformity claim
+rather than a restatement of the current implementation.
--- a/src/Core/CodeAnalysis/Binding/MemberLookup.cs
+++ b/src/Core/CodeAnalysis/Binding/MemberLookup.cs
@@ -3919,7 +3919,23 @@ internal sealed class MemberLookup
                         && (TypeSymbol.RequiresSymbolicProjection(mapped)
                             || TypeSymbol.ContainsNamedTupleElements(mapped))))
                 {
-                    return mapped;
+                    // Issue #3705 (family 2): this branch returned the
+                    // receiver-substituted type RAW, exactly as
+                    // GetClrMethodReturnTypeSymbol's did before #3741 — and
+                    // #3741 fixed the return and parameter branches but not this
+                    // one. Substitution recovers the member's shape and silently
+                    // discards the declaration's `[Nullable]`, so an imported
+                    // `string?` property or indexer read through a constrained
+                    // generic receiver bound as non-null `string` with no
+                    // diagnostic: the unsound direction. Fold the declaration
+                    // flags back in like every sibling reader.
+                    var declarationFlags = ClrNullability.ReadNullableFlags(
+                        openProperty,
+                        openProperty.DeclaringType);
+                    return NullableFlagsBuilder.MergeDeclarationNullability(
+                        mapped,
+                        openProperty.PropertyType,
+                        declarationFlags);
                 }
             }
         }
--- a/src/Core/CodeAnalysis/Emit/NullableFlagsBuilder.cs
+++ b/src/Core/CodeAnalysis/Emit/NullableFlagsBuilder.cs
@@ -78,14 +78,27 @@ internal static class NullableFlagsBuilder
         Type layoutType,
         ImmutableArray<byte> declarationFlags)
     {
-        if (declarationFlags.IsDefaultOrEmpty)
-        {
-            return projectedType;
-        }
-
+        // Issue #3705 family 2: no short-circuit on empty flags. An absent
+        // `[Nullable]` is not "no information" in G# — issue #1354 makes it
+        // mean `T?`, and `ExpandNullableFlags` already expands empty metadata
+        // to a full array of `2`s for exactly that reason. Returning
+        // `projectedType` here is what made a member reached through a
+        // receiver-substituting reader (field, constrained call, deconstruct
+        // out-parameter) keep the pre-#1354 non-null answer while its
+        // ClrNullability-read siblings said `T?`.
         var declaredFlags = ClrNullability.ExpandNullableFlags(
             layoutType,
             declarationFlags);
+
+        // The same expansion read LITERALLY: absent positions stay oblivious
+        // instead of defaulting to `2`. Only the open-type-parameter carve-out
+        // below consults this, because it is the one position where "the
+        // declaration was silent" must not be read as "the declaration said
+        // `T?`" — and the default expansion cannot tell those apart.
+        var literalFlags = ClrNullability.ExpandNullableFlags(
+            layoutType,
+            declarationFlags,
+            absentFill: Oblivious);
         var position = 0;
         return Merge(projectedType, layoutType);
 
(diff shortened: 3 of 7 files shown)


DavidObando:
## Cross-layer check: cs2gs already believed this rule

Checked whether cs2gs's translation-side view of oblivious nullability still agrees with gsc's after this change, since a divergence there surfaces later as a migration-only defect. **They agree better after this PR than before it** — this closes a cross-layer divergence rather than opening one. No cs2gs code is touched.

`CSharpToGSharpTranslator.Expressions.cs`'s `IsImportedObliviousNullableMember` (issue #2202) exists precisely to mirror gsc's import rule, and says so:

```csharp
// gsc maps a member to `T?` based on the nullable context of the assembly that
// DECLARES it, not on a type argument the consumer supplies.
ISymbol original = symbol.OriginalDefinition;
ITypeSymbol type = original switch
{
    IMethodSymbol m => m.ReturnType,
    IPropertySymbol p => p.Type,
    IFieldSymbol f => f.Type,        // <-- fields included
    _ => null,
};
return !SymbolEqualityComparer.Default.Equals(original.ContainingAssembly, …)
    && type is { IsReferenceType: true } and not ITypeParameterSymbol
    && type.NullableAnnotation == NullableAnnotation.None;
```

Two things fall out:

1. **`IFieldSymbol` is in that switch.** cs2gs has assumed since #2202 that gsc reads an imported oblivious *field* as `T?` — citing `ClrNullability.cs` in the comment — and emits `!!` accordingly. But the matrix in this PR measures that on `main` gsc reads exactly that position as **non-null**, because the field path goes through `MergeDeclarationNullability`. So cs2gs was mirroring gsc's *documented* rule while gsc's field path implemented the pre-#1354 one. This PR makes gsc's behaviour match what the translator already believed.

2. **`NullableAnnotation.None` covers both oblivious shapes.** Roslyn reports `None` for absent metadata *and* for an explicit `[Nullable(0)]`, so cs2gs's mirror never had the `ObliviousAbsent` / `ObliviousZero` split that gsc's reader did. The translator has been treating the two shapes uniformly all along — which is independent support for uniformity being the right call, and for `ObliviousZero` belonging in the matrix as a first-class state.

Note the two layers are *not* meant to agree in general and that is by design, not drift: `ObliviousNullabilityAnalyzer` decides an oblivious **C# source declaration** by whole-program null-taint (untainted stays non-null `T`), while #1354/ADR-0136 decides an oblivious **imported metadata** position as `T?` unconditionally. Different inputs, different questions. `IsImportedObliviousNullableMember` is the narrow seam where the translator has to predict gsc, and that seam is what this check is about.

**One residual gap, recorded not fixed** (cs2gs is out of scope here): the mirror keys on a member's *return/type*, so an imported oblivious `Deconstruct` **out-parameter** — `void` return — never matches. gsc now binds that position `T?`; cs2gs's `!!` prediction does not model it. Narrow (it needs a CLR `Deconstruct` on an oblivious imported type reached from migrated code), and it is a translator-side prediction gap rather than an unsoundness, but it is the one place the two layers still disagree about a position this PR moved.

**Expected gate direction, stated before the numbers land:** `!!` count may rise. The gate's stage 2 strips `!!` spans gsc proves redundant; on `main` a `!!` on an imported oblivious field read *was* redundant (gsc thought the value non-null) and got stripped, and now it is load-bearing and stays. Headroom against `nullAssertionCeiling` is 17050 − 16912 = 138. `cs2gs-selfmig-nightly` is dispatched on this branch to measure it.

DavidObando:
## Failing-on-`origin/main` proof

Method: `git checkout origin/main -- src/` with the new test files kept, rebuilt, matrix run. No rebase, no DLL swap under a live test host.

**12 of 35 rows fail on `main`; 23 pass and are the guard rails.**

```
Failed:  12, Passed:  23, Total:  35
```

The 12, grouped by what each group proves:

| group | rows | what it proves |
|---|---|---|
| `Field`, `ConstrainedMethodReturn`, `DeconstructOut` × `ObliviousAbsent` + `ObliviousZero` | 6 | the deferred oblivious decision — and the `ObliviousZero` half of it is the mechanism the deferral never recorded |
| `ConstrainedProperty`, `ConstrainedIndexer` × `ObliviousAbsent` + `ObliviousZero` | 4 | same rule, on the reader #3741 left untouched |
| **`ConstrainedProperty`, `ConstrainedIndexer` × `Nullable`** | **2** | **the unsoundness** — an imported `string?` binding as non-null `string` with no diagnostic |

Honest labelling of the 23 that pass on `main`, because they are doing real work:

- All 8 `NonNull` rows — the negative controls. They fail if the fix widens an annotated non-null position, the #3704 shape.
- `Property`, `MethodReturn`, `IndexerElement` × all 4 states (12 rows) — already correct via the plain `ClrNullability` readers; they are what makes the table an *invariant* rather than a changelog.
- `Field`/`ConstrainedMethodReturn`/`DeconstructOut`/`ConstrainedProperty`/`ConstrainedIndexer` × `Nullable` for the three #3741 already fixed — regression guards on #3741's own work.
- `Fixture_Really_Presents_Four_Distinct_AnnotationStates` — anti-vacuity; passes on `main` by design, since it asserts facts about the emitted fixture metadata, not about the binder.
- `OpenTypeParameterPositions_Take_Nullability_From_The_Argument` — **passes on `main`**, and that is the point: `main` never over-applied #1354 to open type parameters. It is a guard rail against *this PR*, not against `main`, and it is the row that failed while the carve-out was still wrong.

## Blast radius, final

| suite | result |
|---|---|
| `Core.Tests` | **8559 / 8559**, zero failures |
| `Compiler.Tests` | running; will report |
| `Samples_EmittedPE_Match_Baseline` | green, no baseline regeneration — emitted metadata and IL do not move |

For the record, the earlier "8536 tests, 3 movements" figure in this PR's history described the first, oblivious-only version and is superseded by the line above. The intermediate version — before the open-type-parameter carve-out — failed 12 tests across 6 CI shards, all one defect (`map[K, V].Keys` iterating as `K?`); those are fixed, not suppressed, and §3 of the description explains why the carve-out is principled rather than a retreat.

Total fixture churn across the whole change: **two tests**, `Issue2514` (annotated) and `Issue1418` (nil guard), both corrections rather than accommodations — see §6.

DavidObando:
## Verification — final

| measurement | result |
|---|---|
| `Core.Tests` (full, local) | **8559 / 8559**, zero failures |
| `Compiler.Tests` (full, local) | **4609 / 4609**, zero failures |
| CI, final SHA | **35 pass, 0 fail**, 3 skipped |
| failing-on-`origin/main` | **12 of 35 rows fail**, 23 pass as labelled guard rails |
| `Samples_EmittedPE_Match_Baseline` | green, **no baseline regeneration** |
| nullable hygiene (ADR-0155) | clean — no null-forgiving `!` introduced |
| self-migration gate | `migrate` stage green (all 51 apps translated); `validate` shards running |

Two figures quoted earlier in this PR's history are **superseded**, not merely updated:

- *"8536 tests, 3 movements"* described the first, oblivious-only version. The final number is 8559/8559 with zero movements, because the two genuine fixture corrections are now in the branch.
- The intermediate push failed **12 tests across 6 CI shards**, all one defect: open type parameters obeying #1354, so `map[K, V].Keys` iterated as `K?`. Those are **fixed by the §3 carve-out, not suppressed**. CI's sharded run on the final SHA — including every shard that was red — is green.

### Fixture churn: two tests, both corrections

Worth stating plainly since a nullability change is exactly where test edits deserve suspicion. Neither weakens what its test asserts:

- `Issue2514` — fixture annotated `#nullable enable`. Its C# contracts were nullable-oblivious, so the assignments under test were only legal while the constrained-receiver readers dropped declaration nullability. Chosen over `!!` deliberately: `!!` would make the fixture depend on the very gap being closed, and uncommented `!!` is a review defect under ADR-0155.
- `Issue1418` — `nil` guard added. `LinkedList<T>.First` is `[Nullable(2)]` in the BCL (null for an empty list), so the guard is what the declaration always required. The test's actual subject — that a constructed generic property type does not erase to `LinkedListNode[object]` — is unchanged and still asserted by the `11`.

### What the gate can still change

`nullAssertionCeiling` is 17050 against a last-measured 16912 — **138 units of headroom**. The direction of risk is stated up front in the cross-layer comment above: gsc now agrees with what cs2gs already believed about imported oblivious fields, so `!!` spans that stage 2 previously stripped as redundant are now load-bearing and stay. If the ceiling is breached that is a real result requiring a re-baseline decision, not something to absorb quietly, and `greenFloor` (41) is the metric that would matter more.

Independent of the gate, note #3764 — an ilverify failure in `Gsharp.NET.Sdk` traced to the interpolated-string handler emitting `callvirt` on a value type — is being fixed in parallel and is unrelated to this change.

Issue #3775, opened 2026-09-01
cs2gs: four residual `!!`-throws-where-C#-`!`-is-erased sites that need a policy call (from the #3774 sweep)


DavidObando:
Split out of the #3774 sweep of the `EnsureNonNullAssertion` call sites. Twenty sites were inventoried; one was a plain defect (fixed in the #3774 PR), one belongs to the open PR #3773, ten are faithful, and the four below **do** diverge from C# at run time but cannot be fixed by deleting the assertion — each needs a policy call, because removing the `!!` replaces a silent runtime divergence with a hard `gsc` error.

The premise, restated: C#'s `x!` is **erased**; G#'s `x!!` is a **checked assertion that throws on nil**. Every site that emits `!!` onto a value that can be nil at run time adds a throw the original never had — invisible to the compiler, to ILVerify, and to every binding-only test.

## 1. Event subscription with a null handler (`CSharpToGSharpTranslator.Statements.cs`, `ForgiveEventSubscriptionRhs`)

```csharp
EventHandler h = Pick(false);   // returns null
c.Ticked += h;                  // C#: Delegate.Combine(x, null) — a NO-OP
```

```gsharp
c.Ticked += h!!                 // G#: throws
```

`e += null` and `e -= null` are legal, common, and no-ops in C#. The `!!` was added (issue #914) because gsc imports the event's NAMED delegate target as a non-nullable reference even when the BCL event is annotated `DataReceivedEventHandler?`, so the promoted `T?` right-hand side is rejected with GS0155. Options: (a) carry the metadata annotation onto the arrow type gsc converts through, so the `T?` is accepted and the subscription stays a no-op; (b) lower to a guarded subscription (`if h != nil { e += h }`); (c) accept the divergence and document it. (a) is the faithful one and is a gsc-side fix.

## 2. Oblivious sinks bridged at the use site instead of at the declaration (`Statements.cs` `ForgiveElementAccessAssignmentRhs` / `ForgiveObliviousExternalAssignmentRhs`, and the `IsObliviousCompilation` arms of `ForgiveNullableReferenceValue` / `ForgiveInitializerElementValue` / the index-argument bridge)

[...]

Issue #3771, opened 2026-09-01
cs2gs: a spurious '!!' on a widened 'var' local turns normal loop termination into a NullReferenceException (#3501 test-parity)


DavidObando:
## Summary

`cs2gs` emits a `!!` null-assertion on the value assigned to a `var` local whose emitted G# declaration was **already widened to `T?`**. C#'s `!` is erased at compile time; **G#'s `!!` is a checked assertion that THROWS on nil** (`BoundUnaryOperatorKind.NullAssertion` — "Unwraps a nullable to its underlying type; throws at runtime when the value is nil").

The result is a translation that compiles clean, ILVerifies clean, and then throws `NullReferenceException` where the C# original returned normally.

## Reproduction

C# input:

```csharp
#nullable enable
namespace Demo
{
    public class Walker
    {
        private readonly Walker? parent;

        public Walker(Walker? parent) => this.parent = parent;

        public static Walker? Parent(Walker w) => w.parent;

        public static int Depth(Walker start)
        {
            var d = 0;
            for (var current = start; current != null; current = Parent(current))
            {
                d++;
            }

            return d;
        }
    }
}
```

`cs2gs` output on `a13fb992f`:

```gsharp
for var current Walker? = start; current != nil; current = Parent(current!!)!! {
```

The **argument's** `!!` is faithful (`current` is flow-narrowed non-null there and `Parent` takes a non-nullable `Walker`). The trailing `!!` on the assigned value is not: it fires on the very iteration that would have ended the loop.

`Walker.Depth(Walker(Walker(Walker(nil))))` should print `3`; the translated program throws `NullReferenceException` instead. There is no diagnostic anywhere in the chain.

## Impact on the self-migration gate

`BoundScope.TryLookupLexicalNestedTypeAlias` in `src/Core` is exactly this shape:

```csharp
for (var current = currentContainingType; current != null; current = TypeContainingType(current))
```

where `TypeContainingType` returns `TypeSymbol?`. The migrated compiler therefore throws on the first non-nested type it walks. That accounts for **6 of the 13** failures in the migrated `tools/gsgen/GSharp.GeneratorHost.Tests` and for further failures in the migrated `test/LanguageServer.Tests`. Gate run 33463929797, fingerprint `sha256:146ec6a1d0e9…`.

## Root cause

Two places disagree about whether the local's emitted G# type is nullable.

[...]

Issue #3683, opened 2026-08-30
cs2gs: receiver null-forgiveness misses nullable call results and ((x as T)) cast receivers


DavidObando:
Split out of the [#3501 `test/Core.Tests` triage](https://github.com/DavidObando/gsharp/issues/3501#issuecomment-5469492729) (families F5 + F6 — 11 of that app's 94 compile errors). Run [33308959087](https://github.com/DavidObando/gsharp/actions/runs/33308959087).

Two neighbouring holes in `TranslateReceiverWithNullForgiveness`. Both leave a receiver whose G# type is `T?` un-forgiven, and gsc rejects the dereference.

### F5 — the nullable value is a CALL RESULT, not a variable (6)

```csharp
// C# — Type? GetElementType(); warning only
Assert.NotEqual(typeof(int).Assembly.FullName, mapped.GetElementType().Assembly.FullName);
```
```gs
// emitted G# — the VARIABLE got its `!!`, the call result did not
Assert.NotEqual(typeof(int32).Assembly.FullName!!, mapped!!.GetElementType().Assembly.FullName!!)
//                                                                          ^ GS0158 Cannot find member Assembly
```

- `Issue2325ArrayPointerByRefRemapEmitTests.gs(50,86)` — GS0158 `Cannot find member Assembly`
- `ReferenceResolverTests.gs(251,75)` — GS0158 `Cannot find member FullName`
- `PortablePdbEmitTests.gs(374,44)` — GS0158 `Cannot find member Length`
- `Issue3354RectangularArrayParserTests.gs(20,72)` — GS0158 `Cannot find member ArrayCommaTokens` (through a `?[0]` conditional index)
- `Issue1506PerSegmentTypeArgumentParserTests.gs(67,34)` — GS0116 `Type '…SeparatedSyntaxList[…]?' is not indexable` — same root, indexed form
- `Issue1675SyntaxNodeChildEnumerationTests.gs(361,59)` — GS0154, same root, argument form (`AssertChildrenMatchLegacy(child)` where `child` is `SyntaxNode?`)

### F6 — the receiver is a `((x as T))` produced by the #3501 null-preserving cast lowering (5)

```csharp
// C# — BoundExpression? Receiver
Assert.Same(tempDecl.Variable, ((BoundVariableExpression)faExpr.Receiver).Variable);
```
```gs
// emitted G# — GS0158 Cannot find member Variable
[...]

PR #3776, opened 2026-09-01
cs2gs: sweep every `EnsureNonNullAssertion` site — G# `!!` throws where C# `!` is erased (#3774)

### LINKED ISSUE #3774: cs2gs: sweep every `EnsureNonNullAssertion` site — G# `!!` throws where C# `!` is erased (from #3771)
**Generalized from #3771** (found via the `GeneratorHost.Tests` test-parity failure, fixed in PR #3773).

## The defect class

cs2gs treats G# `!!` as if it were C#'s `!`. **They are not equivalent:**

| | C# `x!` | G# `x!!` |
|---|---|---|
| effect | suppresses a nullable *warning* | **runtime check that throws on nil** |
| erased at runtime? | yes | **no** |

So any place the translator emits `!!` onto an expression that is genuinely nullable at runtime **adds a throw the original program did not have**.

## Why this is worse than an ordinary translation bug

It is invisible to every check short of running the code:

- **compile** — passes, the G# is well-typed
- **ILVerify** — passes, the IL is valid
- **binding-only tests** — pass, nothing is mis-bound

It surfaces only as a test-parity failure, and only if a test happens to exercise the path. #3771's instance was a `for` loop's *normal termination condition*:

```csharp
for (var c = start; c != null; c = TypeContainingType(c))   // C#: terminates
```

```
for c := start; c != nil; c = TypeContainingType(c)!!        // G#: guaranteed throw
```

The loop is *designed* to walk until the function returns null. The `!!` converts termination into an NRE.

## The work

**Sweep every `EnsureNonNullAssertion` call site** and, for each, determine whether the target can actually be nil at runtime in G#. #3771 fixed one site by extracting the declaration-widening rule into a single predicate consulted at both the declaration site and the forgiveness site; the same shape likely applies elsewhere.

Ranking notes:

- **Loop increments and conditions are the highest-risk positions** — that is where "may be null" is the *designed* exit, not an edge case.
- Also suspect: ternary / `??` arms, iterator `MoveNext` patterns, and anything downstream of a method whose C# signature is oblivious (see #3705 family 2 / commit `281a9c278`, which changed what gsc considers nullable for imported members — so sites that were previously non-null in G# may now be nullable, making a formerly-harmless `!!` newly load-bearing).

**Detector:** tests that *execute* the translated G#, not ones that assert on translated text. And beware vacuous assertions — the agent on #3771 caught its own first draft asserting `DoesNotContain("Parent(current)!!")` while main actually emits `Parent(current!!)!!`, which passed for the wrong reason. Assert the whole construct.

There is a second-order benefit: every `!!` correctly removed also reduces the `nullAssertionCeiling` count, which is a #3501 readability metric in its own right.

Refs #3501, #3771, #3705. Filed from the test-parity triage that also produced #3770 and #3772.

DavidObando:
Closes #3774. Filed #3775 for the residue.

A sweep of **every** `EnsureNonNullAssertion` call site in `cs2gs`, generalized from #3771 (fixed in the open PR #3773). The premise: C#'s `x!` is **erased**, G#'s `x!!` is a **checked assertion that throws on nil**, so every emission site is a place a translation can silently gain a throw the original never had — invisible to the compiler, to ILVerify and to every binding-only test.

Swept against `origin/main` at `281a9c278`, i.e. **including** the #3705-family-2 change to what gsc considers nullable for imported members.

### Method

The verdicts are not from reading alone. Twenty-six probe programs covering the ranked-risky positions (loop conditions and incrementors, ternary and `??` arms, `??=` value position, `MoveNext`/`Current`, event subscription, index arguments, object/collection initializers, oblivious external members) were translated and inspected for where a `!!` actually lands. Nine of the twenty-six produced any `!!` at all; the rest are the evidence that the target-aware bridges do **not** fire where they should not.

### Site inventory — 20 call sites

**Faithful: `!!` on a dereference receiver (5).** C# would raise a `NullReferenceException` at this same dereference and `!!` raises at exactly the same point.

| site | shape |
|---|---|
| `Statements.cs:2474` | `this.Prop!!.Member = v` (implicit-this assignment target) |
| `Statements.cs:2489` | `recv!!.Member = v` (assignment target) |
| `Expressions.cs:886` | `TranslateReceiverWithNullForgiveness`, main path |
| `Expressions.cs:901` | null-preserving safe-cast receiver (#3683) |
| `Expressions.cs:918` | `as`-bound local receiver (ADR-0160) |

Confirmed by probe: `Parent(w)` → `w!!.parent`, and the `Type` walk → `t!!.BaseType`.

**Faithful: the value is proved non-nil, or the original throws too (3).**

| site | why |
|---|---|
| `Types.cs:1327` | `yield` of a `foreach` variable whose Roslyn flow state is `NotNull` |
| `Invocations.cs:3630` | checked reference cast gated on `IsFlowNarrowedAnnotatedReference` (flow state `NotNull`) |
| `Expressions.cs:723` | `Nullable<T>.Value` → `!!`. Both throw. Only the exception *type* differs (`InvalidOperationException` vs `NullReferenceException`) — noted in #3775 |

**Left alone: target-aware bridges whose gate is sound (7).** Each consults `TargetWillRemainNonNullableReference(targetType, targetSymbol)` — the predicate #3773 consolidated — so a sink that is itself declared or promoted nullable is skipped and no `!!` appears.

`Expressions.cs:1376` (`ForgiveNullableReferenceValue`), `Expressions.cs:2081` (index argument), `Expressions.cs:1285` (value position via `ReceiverNeedsNullForgiveness`), `Patterns.cs:2259` (collection-initializer element), `Patterns.cs:2400` (assignment onto a materialized pattern binder), `Invocations.cs:1559` (argument), `Invocations.cs:2821` (initializer element).

`Expressions.cs:1285` deserves a note because it asserts *before* the target is consulted, which is structurally the #3771 hazard. It survives the sweep because every rule that makes `ReceiverNeedsNullForgiveness` true in value position is either a flow/syntactic non-null proof (`IsLazyInitGuardedFieldUse`, `IsNullGuardNarrowedFieldUse`) or an explicit "this member's return contract was deliberately kept non-null" rule — so the assertion follows the sink's contract rather than an accident. Probes over both shapes emitted no `!!`.

**Deliberately not patched: divergent, but the fix is a contract change, not a deletion (4) — filed as #3775.**

- `Statements.cs:612` — `c.Ticked += h!!`. C#'s `e += null` is a **no-op**; the G# throws. Reproduced by probe. Removing the `!!` produces GS0155 — the real fix is gsc-side (carry the event's metadata annotation onto the arrow type).
- `Statements.cs:659` / `Statements.cs:709` — oblivious sink bridges. The existing comments already state the trade ("a genuinely null oblivious key ... fails at runtime before the index operation"); #3774 is the first time it is named as this defect class.
- `Patterns.cs:504` — the direct `expr!` → `expr!!` mapping (ADR-0115 §B). The purest and highest-volume instance: it fires on every C# null-forgiving operator, including the ones authors wrote *because* the value can be null. Reproduced by probe: `object o = v!;` where `v` is null prints `nil` in C# and throws in G#. #3775 proposes the free subset — drop the assertion when the sink already accepts `T?`.

**Already fixed by the open PR #3773, not duplicated here (1 path).** `ForgiveAssignmentRhs` reaching `Expressions.cs:1376`. The sweep found **two more corpus-shaped instances** beyond #3771's `for` loop, worth adding there:

```gsharp
for var t Type? = typeof(C); t != nil; t = t!!.BaseType!! {   // external Type.BaseType walk
while s != nil { n++; s = Next()!! }                          // while-loop walk
```

Both are guaranteed throws at the loop's designed exit.

### Fixed here: the value-position `??=` (`Patterns.cs:2424`)

The value of `x ??= v` is nil **exactly when `v` is**. The guard asked only Roslyn's flow state and `NullableReferenceValueMayBeNull` — neither of which says anything in a nullable-OBLIVIOUS compilation, where a same-project member the whole-program taint analysis promoted to `T?` still reads as non-null. So:

```csharp
string v = (cache ??= Make());   // C#: v is null, prints "nil"
```

```gsharp
let v string? = ((if cache == nil { (cache = Make()) } else { cache })!!)   // G#: NullReferenceException
```

Note the `!!` is not even load-bearing for binding there — the local is already `string?`. The guard now also consults the promotion result in oblivious mode; a literal or otherwise provably non-null right-hand side keeps its faithful assertion.

### Test evidence

Two of the four tests **execute** the translated G# through the real `gsc`, because this defect class is invisible to text and to binding.

| test | on `origin/main` | on this branch |
|---|---|---|
| `CoalescingAssignmentValue_DoesNotAssertAPromotedNullableRightHandSide` | **fails** | passes |
| `CoalescingAssignmentValue_YieldsNilInsteadOfThrowing` (executes) | **fails** — `System.NullReferenceException`, expected `nil` | passes |
| `CoalescingAssignmentValue_StillAssertsANonNullRightHandSide` | passes (anti-vacuity guard rail) | passes |
| `CoalescingAssignmentValue_NonNullRightHandSideStillRuns` (executes) | passes (anti-vacuity guard rail) | passes |

Failing-on-main proof run with the translator reverted to `origin/main` and the test file kept: 2 failed, 2 passed, exactly as designed. The two guard rails exist so a "fix" that simply deleted the assertion could not pass this file — they pin that a provably non-null RHS still gets its `!!` and that it still runs.

Anti-vacuity: the text assertion pins the **whole** lowered construct (`let v string? = (if cache == nil { (cache = Make()) } else { cache })`) plus a `DoesNotContain` on that same whole construct with `!!` appended — not a fragment that main never emitted, which is how the #3771 draft passed for the wrong reason.

Full `Cs2Gs.Tests`: **2606 passed, 0 failed**.

### DIFF
--- a/tools/cs2gs/Cs2Gs.Tests/Issue3774CoalescingAssignmentAssertionTests.cs
+++ b/tools/cs2gs/Cs2Gs.Tests/Issue3774CoalescingAssignmentAssertionTests.cs
@@ -1,0 +1,231 @@
+// <copyright file="Issue3774CoalescingAssignmentAssertionTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+using Cs2Gs.CodeModel.Ast;
+using Cs2Gs.CodeModel.Printing;
+using Cs2Gs.CodeModel.RoundTrip;
+using Cs2Gs.Translator;
+using Cs2Gs.Translator.Loading;
+using Microsoft.CodeAnalysis;
+using Xunit;
+
+namespace Cs2Gs.Tests;
+
+/// <summary>
+/// Issue #3774 (sweep of the <c>EnsureNonNullAssertion</c> call sites,
+/// generalized from #3771): C#'s <c>!</c> is ERASED at compile time, but G#'s
+/// <c>!!</c> is a CHECKED assertion that THROWS on nil. Any site that emits
+/// <c>!!</c> onto a value that can genuinely be nil at run time therefore adds a
+/// throw the original program never had — invisible to the compiler, to
+/// ILVerify, and to every binding-only assertion.
+/// <para>
+/// The site fixed here is the value-position <c>??=</c>
+/// (<c>TranslateCoalescingAssignmentAsExpression</c>). The value of
+/// <c>x ??= v</c> is nil exactly when <c>v</c> is, but the assertion's guard
+/// asked only C#'s own annotations — which are empty in a nullable-OBLIVIOUS
+/// compilation, so a same-project member the whole-program taint analysis
+/// promoted to <c>T?</c> read as non-null and earned a <c>!!</c> anyway.
+/// </para>
+/// <para>
+/// Two of the four tests EXECUTE the translated G# through the real
+/// <c>gsc</c>, because this defect class is definitionally invisible to a
+/// text-only assertion. The two "still asserts" tests are anti-vacuity guard
+/// rails: they pass on <c>origin/main</c> too, and exist so a fix that simply
+/// deleted the assertion could not pass this file.
+/// </para>
+/// </summary>
+public class Issue3774CoalescingAssignmentAssertionTests
+{
+    private const string PromotedRightHandSideSource = @"
+using System;
+
+namespace Demo
+{
+    public sealed class C
+    {
+        private static string cache;
+
+        private static string Make()
+        {
+            return null;
+        }
+
+        public static void Run()
+        {
+            string v = (cache ??= Make());
+            Console.WriteLine(v == null ? ""nil"" : v);
+        }
+    }
+}";
+
+    private const string NonNullRightHandSideSource = @"
+using System;
+
+namespace Demo
+{
+    public sealed class C
+    {
+        private static string cache;
+
+        public static void Run()
+        {
+            string v = (cache ??= ""d"");
+            Console.WriteLine(v);
+        }
+    }
+}";
+
+    /// <summary>
+    /// The whole lowered construct is asserted, not a fragment: the #3771 agent
+    /// caught its own first draft asserting a substring that main never emitted,
+    /// so it passed vacuously. Here the exact conditional-expression form is
+    /// pinned and the trailing <c>)!!</c> must be absent from it.
+    /// </summary>
+    [Fact]
+    public void CoalescingAssignmentValue_DoesNotAssertAPromotedNullableRightHandSide()
+    {
+        string printed = TranslateOblivious(PromotedRightHandSideSource);
+
+        Assert.Contains(
+            @"let v string? = (if cache == nil { (cache = Make()) } else { cache })",
+            printed);
+        Assert.DoesNotContain(
+            @"(if cache == nil { (cache = Make()) } else { cache })!!",
+            printed);
+    }
+
+    /// <summary>
+    /// The behavioural half. C# prints <c>nil</c>; on <c>origin/main</c> the
+    /// translated G# throws a <see cref="NullReferenceException"/> instead.
+    /// </summary>
+    [Fact]
+    public void CoalescingAssignmentValue_YieldsNilInsteadOfThrowing()
+    {
+        string printed = TranslateOblivious(PromotedRightHandSideSource);
+        string stdout = CompileAndRun(printed, "C.Run()");
+        Assert.Equal("nil", stdout.Trim());
+    }
+
+    /// <summary>
+    /// Anti-vacuity guard rail (passes on <c>origin/main</c> as well): a
+    /// right-hand side that genuinely cannot be nil keeps its faithful
+    /// assertion, so the fix is a narrowing of the guard rather than a deletion.
+    /// </summary>
+    [Fact]
+    public void CoalescingAssignmentValue_StillAssertsANonNullRightHandSide()
+    {
+        string printed = TranslateOblivious(NonNullRightHandSideSource);
+
+        Assert.Contains(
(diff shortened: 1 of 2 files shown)


DavidObando:
### `!!` count delta: **0** (21471 → 21471) — measured, not assumed

Two whole-repository stage-1 translate passes (`build/run-cs2gs-selfmig-migrate.sh`), same machine, same corpus, differing only in `cs2gs.dll` (`origin/main` vs this branch):

| | apps translated | labels | `__local_` | lines>300 | **bangs** |
|---|---|---|---|---|---|
| `origin/main` (`281a9c278`) | 50/51 | 0 | 29 | 578 | **21471** |
| this branch | 50/51 | 0 | 29 | 578 | **21471** |

A `diff -rq` of the two migrated trees finds exactly **one** differing `.gs` file — `CSharpToGSharpTranslator.Patterns.gs`, i.e. the translation of the source file this PR edits. Not one line of translated output changes anywhere else. (The `.gsproj` diffs are the `git describe` SDK version stamp.)

That is the honest result and it is worth stating plainly rather than dressing up: **this fix buys no `nullAssertionCeiling` improvement on this repo's corpus.** The reason is structural — the guard it repairs is gated on `IsObliviousCompilation()`, and the gsharp production tree is NRT-migrated, so every value-position `??=` in the corpus already took the correct path. The 37 lowered `??=` forms that do carry a `!!` in the migrated tree all have a provably non-null right-hand side (`new Dictionary[...]()`, `List[string]()`, `ImmutableArray.CreateBuilder[...]()`) and keep their faithful assertion, exactly as intended.

The value here is the correctness fix for nullable-oblivious corpora (the Oahu-shaped migrations), plus the inventory itself. The readability metric is unmoved and `selfmig-baseline.json` is therefore untouched — a metric that did not improve must not be ratcheted.

Issue #3772, opened 2026-09-01
selfmig: Extensions.Tests parity failure is a mirror-fidelity gap (no GSharp.sln, no Gsharp.Extensions project), not a translation defect — and the test-parity fingerprint does not discriminate root causes


DavidObando:
## Summary

`test/Extensions.Tests` is reported as a `test-parity-failure` in the self-migration gate, but the two failing tests are not a translation defect. They fail because the **migrated repository mirror does not reproduce the on-disk layout the tests probe for**. The translation itself is correct; the tests cannot find a build artifact.

Gate run 33463929797, `test/Extensions.Tests/Extensions.Tests.csproj`, fingerprint `sha256:146ec6a1d0e9…`.

## Failing assertion

```
Failed FirstOrNil_WithoutImport_FailsBinding
Failed FirstOrNil_WithImport_CompilesCleanly
  Error Message:
   Gsharp.Extensions.dll must be built before running this test (run `dotnet build`).
  Stack Trace:
     at GSharp.Extensions.Tests.NoAutoImportTests.CompileSnippet(String source)
```

171 of the 173 tests in the migrated project pass.

## Why

`NoAutoImportTests.LocateGsharpExtensionsAssembly` walks up from the test assembly location looking for a repository root, then for a specific build output:

```csharp
var dir = new DirectoryInfo(Path.GetDirectoryName(typeof(NoAutoImportTests).Assembly.Location)!);
while (dir != null)
{
    if (File.Exists(Path.Combine(dir.FullName, "GSharp.sln")))
    {
        foreach (var cfg in new[] { "Debug", "Release" })
        {
            var candidate = Path.Combine(dir.FullName, "out", "bin", cfg, "Gsharp.Extensions", "Gsharp.Extensions.dll");
            …
```

Two independent things make that fail in the mirror at `/tmp/gsharp-cs2gs-selfmig/migrated`:

1. **The root anchor does not exist.** The real repository root has `GSharp.sln`. The generated mirror root has **`GSharp.slnx`** and no `GSharp.sln`, so the upward walk never finds a root and returns `null` before it ever looks for the DLL.
[...]

Issue #3770, opened 2026-09-01
cs2gs: a C# 12 primary constructor + get-only auto-property initializer emits an UNREACHABLE parameterless init() — properties silently left at default (#3501 test-parity)


DavidObando:
## Summary

A C# 12 **primary constructor** combined with a **get-only auto-property initializer** translates to a G# class that carries BOTH a primary-constructor parameter list AND a synthesized **parameterless** `init()` holding the initializer assignments. The parameterless `init()` is unreachable — every construction goes through the primary constructor — so the properties are silently left at their default values.

The translation is accepted by `cs2gs`, compiles clean with `gsc`, and passes ILVerify. Only running the code shows it. This is the `test-parity-failure` profile from the #3501 self-migration gate.

## Reproduction

C# input:

```csharp
#nullable enable
namespace Demo
{
    public class Doc(string tree, int n)
    {
        public string Tree { get; } = tree;

        public int N { get; } = n;
    }
}
```

`cs2gs` output on `a13fb992f`:

```gsharp
class Doc(tree string, n int32) {
    init() {                 // <-- never reached
        Tree = tree
        N = n
    }

    prop Tree string { get; init; }

    prop N int32 { get; init; }
}
```

Appending `let d = Doc("hello", 7); Console.WriteLine(d.Tree + "|" + d.N.ToString())` and compiling with `gsc`:

```
expected: hello|7
actual:   |0
```

## Impact on the self-migration gate

`src/LanguageServer/DocumentContent.cs` is exactly this shape (`class DocumentContent(SyntaxTree syntaxTree, IReadOnlyList<int> lines, …)` with `public SyntaxTree SyntaxTree { get; } = syntaxTree;`). Every migrated `DocumentContent` therefore has a null `SyntaxTree`, and **151 of the 259** failures in the migrated `test/LanguageServer.Tests` are a `NullReferenceException` at `SemanticLookup.ToOffset`, with the remainder spread across `CodeLensComputer`, `FoldingComputer`, `DocumentSymbolComputer`, `InlayHintComputer`, `WorkspaceSymbolHandler` and `CodeActionComputer` — all of which take a `DocumentContent`. Gate run 33463929797, fingerprint `sha256:c79d8d6acf83…`.

## Root cause

`tools/cs2gs/Cs2Gs.Translator/CSharpToGSharpTranslator.Declarations.cs`, the OD-T1 block that synthesizes the carrier constructor:

```csharp
[...]

PR #3783, opened 2026-09-01
cs2gs: make the migrated mirror a repository, not a bag of projects (#3772)

DavidObando:
Fixes the mirror-fidelity half of #3772 by making the migrated tree a **repository**, not a bag of projects. Neither of the two options that loosen the probe (accept `.slnx` in `LocateGsharpExtensionsAssembly`, or exclude the tests) is taken: the layout is fixed so the probe passes because the layout is right.

## Both causes, reproduced

A whole-repo migrate + `cs2gs validate --app test/Extensions.Tests/...` reproduces the gate's fingerprint `sha256:146ec6a1d0e9` exactly — `Failed: 2, Passed: 171, Total: 173`, both failures `Gsharp.Extensions.dll must be built before running this test`:

1. **No root anchor.** The mirror root had only `GSharp.slnx`. 37 files across `test/Core.Tests`, `test/Compiler.Tests`, `test/Interpreter.Tests`, `test/LanguageServer.Tests` and `test/Extensions.Tests` walk up looking for a file literally named `GSharp.sln`; in the mirror every one of them returns `null`. Most are masked today by apps that are red earlier.
2. **No project.** `RepositoryMirror.Prepare` skips every `.csproj` on the assumption that a translated `.gsproj` replaces it — false for an `--exclude`d project. `src/Sdk/Gsharp.Extensions` is excluded because it is *already G#*, so the mirror carried its complete `.gs` source set and dropped only the file that builds it, leaving four migrated projects (Extensions.Tests, Compiler.Tests, Interpreter.Tests, Repl) pointing a `ProjectReference` at a file that is not there.

## Fixed as rules, not as file names

- **Solutions are mirrored, not renamed.** Every `.sln` is now written to the mirror under its own name with its project paths retargeted at the generated projects, and the `.slnx` conversion lands beside it (only the XML format can type-tag a `.gsproj`). This also stops the `gsharp-xunit` template shipping a `.slnx` where it authored a `.sln`.
- **The mirror never contains a half project.** An excluded project with no `.cs` sources of its own — its sources are all mirrored verbatim — keeps its project file.
- **Every mirrored project builds with the pinned SDK.** The mirrored `Gsharp.Extensions` is rebound off the in-tree bootstrap (which resolves `gsc.dll` from `out/bin/<cfg>/` at *evaluation* time, a thing no clean shard runner has) onto the same `Gsharp.NET.Sdk` moniker as every other mirrored project.
- **An untranslated `ProjectReference` is a build-only reference.** `ReferenceOutputAssembly="false"`: its job in the mirror is to get the project built, while the assembly the consumer compiles against comes from the SDK. Without this, gsc sees two assemblies with one identity (GS9200).

## Before / after (whole-repo migrate, then `validate --app`)

| | `test/Extensions.Tests` | `src/Repl` |
|---|---|---|
| before | compile PASS, ilverify PASS, **test-parity FAIL** — `146ec6a1d0e9`, 2/173 failed, *"Gsharp.Extensions.dll must be built before running this test"* | PASS/PASS/PASS |
| after | compile PASS, ilverify PASS, **test-parity FAIL** — 2/173 failed, `error GS9998: NullReferenceException` from the migrated gsc | PASS/PASS/PASS |

The mirror now produces `out/bin/Release/Gsharp.Extensions/Gsharp.Extensions.dll` and the harness locates it: the assertion the two tests fail on is no longer the missing artifact, it is the compiler output. Repl — which project-references the same project and is in `greenApps` — stays green through the change.

## The fix unmasked a real defect

With the artifact present the two tests finally exercise the migrated compiler, and it fails where the C# compiler does not:

```
$ dotnet <mirror>/out/bin/Release/Compiler/gsc.dll /r:.../Gsharp.Extensions.dll snippet.gs
error GS9998: NullReferenceException: Object reference not set to an instance of an object.
$ dotnet out/bin/Release/Compiler/gsc.dll     /r:.../Gsharp.Extensions.dll snippet.gs
Wrote snippet2.dll
```

Same snippet, same reference assembly, opposite outcome — a silent runtime divergence in the same family as #3770/#3771, filed separately. `Extensions.Tests` therefore stays a `test-parity` failure, but for a real reason now rather than a missing build artifact.

## Baseline

No `greenFloor` or `greenApps` movement: the app was red before and is red after (for a different, genuine reason), Repl is unchanged, and no app is added to or removed from the corpus (`Gsharp.Extensions` is mirrored, not discovered).

## Tests

`Issue3772MirrorFidelityTests` (4) covers the mirrored `.sln` + `.slnx` pair, the completeness contract expecting both, the mirrored excluded project with its retargeted references and SDK rebind, and the excluded project with C# sources that must *not* be mirrored. `GSharpProjectTransformerTests`, `Issue2317ProjectReferenceTests`, `Adr0169AnalyzerProjectTransformTests`, `RepositoryMirrorTests`, `RepositorySolutionGeneratorTests`, `Issue3580OrphanMirrorRobustnessTests`, `Issue2868MirroredArtifactLayoutTests` all pass (41 tests).

### DIFF
--- a/build/selfmig-common.sh
+++ b/build/selfmig-common.sh
@@ -18,7 +18,11 @@
 # permanently-C# extension via linked sources. Gsharp.Extensions is excluded
 # because it is ALREADY G# (all-.gs sources bootstrapped by the latest
 # compiler without a gsproj) — there is nothing to migrate, and feeding its
-# .gs sources through the C# parser only produces phantom gaps.
+# .gs sources through the C# parser only produces phantom gaps. Excluded from
+# TRANSLATION is not excluded from the MIRROR: issue #3772 keeps its .csproj in
+# the migrated tree (with references retargeted), because its .gs sources are
+# mirrored verbatim and four migrated projects — Extensions.Tests,
+# Compiler.Tests, Interpreter.Tests and Repl — project-reference it.
 #
 # INVARIANT (issue #3668): --exclude is NOT a sharding mechanism. Excluding a
 # project that another app project-references breaks reference resolution and
--- a/docs/adr/0115-csharp-to-gsharp-migration-tool.md
+++ b/docs/adr/0115-csharp-to-gsharp-migration-tool.md
@@ -34,8 +34,11 @@ Build `cs2gs` as a **Roslyn-based, offline, deterministic** translator feeding a
 `cs2gs migrate` produces a maintainable repository mirror by default. The
 destination preserves repository-relative directories and non-C# files,
 translates each checked-in `.cs` to the corresponding `.gs`, transforms each
-`.csproj` to a same-location `.gsproj`, and replaces `.sln` files with `.slnx`
-while preserving project coverage. Literal Nerdbank.GitVersioning versions
+`.csproj` to a same-location `.gsproj`, and mirrors each `.sln` under its own
+name with a `.slnx` conversion beside it, all while preserving project coverage
+(issue #3772: the mirror is a repository, so code in it that walks up to a
+repository root named `GSharp.sln` finds one, and a project excluded from
+translation because it has nothing to translate keeps its project file). Literal Nerdbank.GitVersioning versions
 below the G#-compatible `3.11.13-beta` floor are upgraded in transformed
 projects and shared MSBuild props. Validation still runs through the four-stage
 pipeline, but logs, triage records, reports, and build intermediates are written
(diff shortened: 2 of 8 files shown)


DavidObando:
Follow-up for the defect this unmasked: #3784 (migrated gsc NREs where the C# compiler compiles cleanly). That is why `test/Extensions.Tests` is still a test-parity failure after this PR — the mirror is right, the compiler is not.

Issue #3686, opened 2026-08-30
cs2gs ADR-0169 M5: migrated analyzer TESTS still target the Roslyn harness, so a GSharpDiagnosticAnalyzer fails the DiagnosticAnalyzer parameter (GS0154)


DavidObando:
Part of the #3501 self-migration wall for `test/InternalAnalyzers.Tests` (green run [33308959087](https://github.com/DavidObando/gsharp/actions/runs/33308959087)). The array-covariance error in the same app is #3685; this issue is the other three.

## Symptom

```
EmitCacheKeyRemapScopeAnalyzerTests.gs(34):  error GS0154: Parameter 'analyzer' requires a value of
  type 'Microsoft.CodeAnalysis.Diagnostics.DiagnosticAnalyzer' but was given …
EmitCacheKeyRemapScopeAnalyzerTests.gs(94):  error GS0154: (same)
RewriterClonePreservationAnalyzerTests.gs(23): error GS0154: (same)
```

## Root cause: the analyzer moved to the G# framework; its tests did not

ADR-0169 translation is enabled **per project**, by
`AnalyzerProjectDetector.IsAnalyzerProject` — "the compilation resolves
`Microsoft.CodeAnalysis.Diagnostics.DiagnosticAnalyzer` **and at least one
source type derives from it**".

- `src/Analyzers/InternalAnalyzers` declares the five GSA analyzers, so it
  translates in analyzer-API mode. The migrated output is a **G#** analyzer:

  ```
  # migrated src/Analyzers/InternalAnalyzers/StructFieldDefsReadAnalyzer.gs
  @GSharpDiagnosticAnalyzer
  class StructFieldDefsReadAnalyzer : GSharpDiagnosticAnalyzer {
  ```

- `test/InternalAnalyzers.Tests` declares **no** analyzer type, so the detector
  says false and the project is translated in ordinary mode. The migrated helper
  keeps the Roslyn contract verbatim:

  ```
  # migrated test/InternalAnalyzers.Tests/AnalyzerTestHelper.gs
  async func AssertDiagnosticsAsync(analyzer DiagnosticAnalyzer, source string, diagnosticIds ...string) {
      …
      let tree = CSharpSyntaxTree.ParseText(cleanSource, CSharpParseOptions(LanguageVersion.Preview))
      let compilation = CSharpCompilation.Create("AnalyzerTests", …)
[...]

PR #3790, opened 2026-09-01
cs2gs: analyzer mode claims a test project only when it declares a harness (#3789)

### LINKED ISSUE #3789: REGRESSION from #3777: analyzer-mode detection claims Cs2Gs.Tests, which now fails translate (98 gaps)
**Regression from #3777 (`16fd62ef7`), found while measuring for #3782.**

`tools/cs2gs/Cs2Gs.Tests` now fails at the **translate** stage on current `main` — it previously reached `compile` and failed there with a 14-error wall.

```
98 gaps, first: Microsoft.CodeAnalysis.MetadataReference has no G# analyzer-API mapping
```

## This was predicted

The #3782 triage flagged it as a forward risk while #3777 was still open:

> #3777 is open and will likely invalidate this measurement. Its widened `IsAnalyzerTestProject` claims projects that *instantiate* an analyzer from a referenced non-Roslyn assembly — `Cs2Gs.Tests` instantiates `StructFieldDefsReadAnalyzer` from the project-referenced `InternalAnalyzers`, and has **256** `Microsoft.CodeAnalysis` mentions that analyzer mode would rewrite wholesale.

That is what happened. `Cs2Gs.Tests` is a **cs2gs test project that happens to instantiate an analyzer**, not an analyzer test project — but it matches the detector's shape, so analyzer mode claims it and tries to map its entire Roslyn surface onto the G# analyzer API. Most of that surface (`MetadataReference`, and 97 more) has no mapping and should not have one: those uses are cs2gs's own machinery, not analyzer code.

## Why the gate has not caught it

Both recent gate runs (`33499698666`, `33513731258`) ran against `39e988b39`, which predates `16fd62ef7`. The next nightly on current main is the first that will see it.

## Suggested direction (verify before building)

The detector needs to distinguish "instantiates an analyzer in order to **test** it" from "instantiates an analyzer incidentally, while being about something else". Candidate signals, in rough order of promise:

1. **Does the project reference the analyzer *test harness*** (the verifier), rather than merely an analyzer? `InternalAnalyzers.Tests` does; `Cs2Gs.Tests` does not.
2. **Proportion**: an analyzer test project is *mostly* analyzer tests. `Cs2Gs.Tests` has 256 `Microsoft.CodeAnalysis` mentions of which the analyzer-related fraction is tiny.
3. **Per-file rather than per-project** claiming, so a project can contain analyzer tests without its unrelated files being rewritten.

Option 1 is the narrowest and most semantic; option 3 is the most robust but the largest change.

Whatever lands must keep #3777's own guard tests passing — `AnalyzerConsumerProject_IsNotTreatedAsATestProject` and `ProjectThatOnlyUsesRoslyn_IsNotAnAnalyzerTestProject` exist precisely to bound this detector, and `src/Core` project-references the analyzer as a consumer. Add `Cs2Gs.Tests`' shape as a third guard case.

## Note

Do not fix this by adding `Cs2Gs.Tests` to an exclusion list. The detector's *rule* is what is wrong; an exclusion would leave the next project with the same shape to rediscover it.

Refs #3777, #3686, #3782, #3501.

DavidObando:
Fixes #3789. Regression from #3777 (`16fd62ef7`, ADR-0169 M5 / #3686); part of #3501.

## Mechanism, reproduced

Whole-repository `cs2gs migrate --translate-only` at `origin/main` (42cbee92f) reproduces the report exactly:

```
tools/cs2gs/Cs2Gs.Tests/Cs2Gs.Tests.csproj      FAIL(98)
```

first gap, verbatim from the run's `translate-60d5f8f4a2ce.json`:

```
CS2GS-GAP  Roslyn API type 'Microsoft.CodeAnalysis.MetadataReference' has no
           G# analyzer-API mapping (ADR-0169).
           Adr0169AnalyzerTranslationTests.cs(937,21)
           .Append(MetadataReference.CreateFromFile(fixturePath))
```

Cause confirmed: `Cs2Gs.Tests` project-references `InternalAnalyzers` and constructs `new StructFieldDefsReadAnalyzer()` in `Adr0169AnalyzerParityTests.RunRoslynAnalyzer` — it runs the real GSA analyzers **as a library** to diff their diagnostics against the translated analyzers'. That matches `IsAnalyzerTestProject`'s "instantiates an analyzer from a referenced non-Roslyn assembly", so analyzer mode claims the project and tries to map its whole Roslyn surface, ~256 mentions of which are cs2gs's own machinery.

(The issue's premise held on every point I could check. One detail worth recording because it nearly misleads: `Cs2Gs.Tests` *does* contain `class MiniAnalyzer : DiagnosticAnalyzer` and friends — but only inside `const string` fixtures, so `IsAnalyzerProject` is correctly false and the test-project probe is what claims it.)

## Discriminator: the harness, not the instantiation

`IsAnalyzerTestProject` now requires **both** signals: a declared analyzer test harness (`IsAnalyzerTestHarnessEntry` — a static method taking a `DiagnosticAnalyzer` and a source `string`) **and** the existing referenced-analyzer instantiation.

The harness is the right key because it is the only thing analyzer mode does for a test project that it does not already do for an analyzer: #3777's whole M5 deliverable is rewriting that method's body onto `GSharpAnalyzerVerifier.VerifyAnalyzer`. No harness ⇒ analyzer mode has nothing to offer and only a whole Roslyn surface to mis-map. The predicate moved onto `AnalyzerProjectDetector` and the translator's `IsAnalyzerHarnessEntry` now calls it, so *claiming* and *rewriting* are one rule and cannot drift.

Against the issue's three candidates:

1. **Reference to the analyzer test harness / verifier.** Does not discriminate as stated: `Cs2Gs.Tests` project-references `GSharp.CodeAnalysis.Analyzers.Testing` — the G# verifier — just like the analyzer tests do. The only reference that separates the two is `InternalAnalyzers.Tests`' `Microsoft.CodeAnalysis.CSharp.Analyzer.Testing.XUnit` package, which **nothing in that project binds against** (its harness is hand-rolled and drives Roslyn's analyzer driver directly). Hanging #3777's deliverable on a dead `PackageReference` means a routine cleanup silently un-fixes it, and an analyzer test project that rolls its own harness — precisely our case — would never be claimed.
2. **Proportion of analyzer-related surface.** No defensible cut point, and any project above the threshold still gaps on the remainder — the failure mode is unchanged, just rarer.
3. **Per-file claiming.** Does not fix this bug and breaks the working one. `Cs2Gs.Tests`' analyzer instantiation lives in `Adr0169AnalyzerParityTests.cs`, which is itself full of unmappable cs2gs machinery, so that file would still be claimed and still gap; meanwhile `AnalyzerTestHelper.cs` — the file that must be claimed — instantiates no analyzer at all and would stop being claimed.

No exclusion list; nothing is keyed on a project name.

## Measurement — whole-repository translate, before and after

Same `build/run-cs2gs-selfmig-migrate.sh` pass, Release gsc/SDK rebuilt on each side (the SDK moniker is `git describe`-stamped, so each side is a real commit: `42cbee92f` → `be79b89d5`).

| app | before (`origin/main`) | after |
|---|---|---|
| `tools/cs2gs/Cs2Gs.Tests` | **FAIL — 98 gaps** | **PASS** |
| `test/InternalAnalyzers.Tests` | PASS | PASS |
| `src/Analyzers/InternalAnalyzers` | PASS | PASS |
| whole repo | 49/51 | **50/51** |

The one remaining failure, `test/Interpreter.Tests` FAIL(1), is identical before and after and unrelated.

`Cs2Gs.Tests` returns to failing at **compile**, where it was before #3777; that ~14-error wall is pre-existing and out of scope here.

**#3777 is not regressed, by construction as well as by measurement:** the migrated `test/InternalAnalyzers.Tests` and `src/Analyzers/InternalAnalyzers` trees are **byte-identical** before and after — `diff -r` reports a single difference in each, the `Gsharp.NET.Sdk/<version>` moniker, which changes because the commit changed. Identical translated output means compile/ILVerify cannot have moved; the `cs2gs validate` run on the after-tree confirms translate/compile/ilverify PASS for the pair.

## Tests

`Issue3686AnalyzerTestHarnessTests` — 10 tests, all executing, all green.

Three new ones, in the file that already holds #3777's guards:

- `ProjectDrivingAnalyzersAsALibrary_IsNotAnAnalyzerTestProject` — the third guard case the issue asks for: `Cs2Gs.Tests`' shape (Roslyn as a library, `MetadataReference` in the signature, an analyzer from a referenced assembly constructed incidentally, no harness) is not claimed.
- `AnalyzerModeOverALibraryConsumerGaps_TheDetectorsModeDoesNot` — the mechanism, executed rather than asserted. Forcing analyzer mode over that source really does emit the `MetadataReference … has no G# analyzer-API mapping` gap; translating it in the mode the **detector** picks emits none. The forced half is the anti-vacuity guard — if the gap ever stops being produced, this test fails rather than passing for the wrong reason.
- `AddingAHarnessToTheSameProject_FlipsTheMode` — the same library-consumer file plus a harness **is** claimed, so the rule keys on the harness and not on the absence of unrelated Roslyn code.

### Failing-on-`origin/main` proof

One axis: with only the new `DeclaresAnalyzerTestHarness` guard disabled (everything else in this PR left in place), exactly the two tests above fail and the other eight pass:

```
[FAIL] AnalyzerModeOverALibraryConsumerGaps_TheDetectorsModeDoesNot
[FAIL] ProjectDrivingAnalyzersAsALibrary_IsNotAnAnalyzerTestProject
Failed: 2, Passed: 8
```

Honest labelling: `AnalyzerConsumerProject_IsNotTreatedAsATestProject`, `ProjectThatOnlyUsesRoslyn_IsNotAnAnalyzerTestProject`, `AnalyzerProject_IsNotItsOwnTestProject`, `TestProject_InstantiatingAReferencedAnalyzer_IsDetected` and the new `AddingAHarnessToTheSameProject_FlipsTheMode` pass with and without the fix — they are regression guards bounding the detector in the other direction, not proof of this change. The app-level before/after above is the proof.

## Known adjacent issue, deliberately not fixed here

`GSharpProjectTransformer` has its own, purely structural test for an analyzer test project (a plain `ProjectReference` to an analyzer project), so it still treats `Cs2Gs.Tests` as one and injects `GSharp.Core` + the verifier into its migrated `.gsproj`. That is inert for this app — it has no `Microsoft.CodeAnalysis` `PackageReference`s to strip, and it fails compile for unrelated pre-existing reasons — and the transformer has no compilation to consult, so unifying it with the semantic rule means threading the translate-stage verdict through the pipeline. Worth doing, but it is a different change from this regression fix.

### DIFF
--- a/docs/cs2gs-analyzer-translation.md
+++ b/docs/cs2gs-analyzer-translation.md
@@ -199,13 +199,37 @@ bind, so an instance-based overload
 alongside it. Hand-written G# analyzer tests keep using the generic form.
 
 Detection is per project on both halves: `AnalyzerProjectDetector` gained
-`IsAnalyzerTestProject` (a project that instantiates an analyzer declared in a
-referenced, non-Roslyn assembly), and `GSharpProjectTransformer` recognizes the
+`IsAnalyzerTestProject` — a project that **declares an analyzer test harness**
+(`IsAnalyzerTestHarnessEntry`: a static method taking a `DiagnosticAnalyzer` and
+a source `string`) **and** instantiates an analyzer declared in a referenced,
+non-Roslyn assembly — and `GSharpProjectTransformer` recognizes the
 structural counterpart (a `ProjectReference` to an analyzer project that is not
 an `OutputItemType="Analyzer"` consumer reference) to inject the two assemblies
 the migrated tests bind — `GSharp.Core` and the verifier — both copied to the
 test output, because a test assembly is loaded by the test host rather than by
 gsc.
+
+Why the harness, and not instantiation alone (issue #3789): analyzer mode maps
+a project's **whole** `Microsoft.CodeAnalysis` surface, so it may only claim a
+project whose Roslyn use it can map. Constructing an analyzer is not enough
+evidence of that — `tools/cs2gs/Cs2Gs.Tests` runs the real GSA analyzers as a
+library to diff them against their translated counterparts, and its other
+~256 Roslyn uses are cs2gs machinery (`MetadataReference`, `CSharpCompilation`)
+that has, and should have, no analyzer-API mapping; claiming it turned a
+14-error compile wall into 98 translate gaps. The harness is the one member
+analyzer mode rewrites for a test project, so its presence is exactly the
+condition under which analyzer mode has something to offer, and the detector
+and the rewrite share one predicate. Rejected alternatives: keying on a
+reference to Roslyn's analyzer-testing package (the repo's own harness is
+hand-rolled and does not bind it, so the signal is a dead `PackageReference`
+that a cleanup would silently remove — and both projects reference the *G#*
+verifier, so that variant does not discriminate at all); a proportion
+threshold on analyzer-related surface (no defensible cut point, and any
+project above it still gaps on the remainder); and per-file claiming (it does
+not fix the case — `Cs2Gs.Tests`' analyzer instantiation sits in a file that
+is itself full of unmappable cs2gs machinery — while breaking the working
+case, since `AnalyzerTestHelper.cs` instantiates no analyzer and would stop
+being claimed).
 
 The remaining half is the **embedded C# snippets** — raw strings of analyzed
 code inside tests. For functional equivalence they must become G# snippets.
--- a/tools/cs2gs/Cs2Gs.Tests/Issue3686AnalyzerTestHarnessTests.cs
+++ b/tools/cs2gs/Cs2Gs.Tests/Issue3686AnalyzerTestHarnessTests.cs
@@ -65,6 +65,34 @@ public sealed class SampleAnalyzer : DiagnosticAnalyzer
 ";
 
     /// <summary>
+    /// Issue #3789: <c>tools/cs2gs/Cs2Gs.Tests</c>' shape — Roslyn used as a
+    /// LIBRARY, with an analyzer from a referenced assembly constructed
+    /// incidentally to diff its output. The <c>MetadataReference</c> parameter
+    /// is the real project's first analyzer-mode gap.
+    /// </summary>
+    private const string ToolingSource = @"
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Sample.Tooling;
+
+public static class ParityCheck
+{
+    public static int CountDiagnostics(string source, MetadataReference[] references)
+    {
+        var tree = CSharpSyntaxTree.ParseText(source);
+        var compilation = CSharpCompilation.Create(""Corpus"", new[] { tree }, references);
+        var withAnalyzers = compilation.WithAnalyzers(
+            ImmutableArray.Create<DiagnosticAnalyzer>(new Sample.SampleAnalyzer()));
+        return withAnalyzers.GetAnalyzerDiagnosticsAsync().GetAwaiter().GetResult().Length;
+    }
+}
+";
+
+    /// <summary>
     /// The repo's own <c>AnalyzerTestHelper</c> shape, trimmed to the parts
     /// that matter: the Roslyn compilation pipeline, the metadata-reference
     /// plumbing, and the marker stripping.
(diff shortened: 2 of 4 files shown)


DavidObando:
### Validation of the after-tree (`cs2gs validate`, shard of 3 apps)

`test/InternalAnalyzers.Tests`, on the migrated tree produced by the after-run:

- **compile** — `sdk.build.log`: `0 Error(s)` (22 warnings), PASS.
- **ilverify** — `All Classes and Methods in …/GSharp.InternalAnalyzers.Tests.dll Verified.`, PASS.
- **test-parity** — `Failed: 16, Passed: 0, Total: 16`, i.e. exactly the state #3777 shipped and documented (the migrated tests still hand **C#** snippets to the G# verifier; the snippet half of M5 is the separate follow-up). Unchanged by this PR.

`src/Analyzers/InternalAnalyzers` validated alongside it in the same shard.

Note this could only ever have been unchanged: the migrated `.gs` output for both halves of the pair is byte-identical before and after this change, the sole diff being the `Gsharp.NET.Sdk/<version>` moniker (`0.4.369-g42cbee92f5` → `0.4.371-gbe79b89d5f`), which moves because the commit moved.

The shard's third app, `tools/cs2gs/Cs2Gs.Tests`, now reaches compile — which is the point of the fix — and I stopped its run there rather than spend the full G# rebuild of its dependency chain on a wall that predates #3777 and is out of scope for this issue.

Issue #3802, opened 2026-09-01
REGRESSION (gate 40/51): gsc does not implement [NotNullIfNotNull], so Path.ChangeExtension no longer binds after #3705 family 2


DavidObando:
**Gate regression.** Run [33552625814](https://github.com/DavidObando/gsharp/actions/runs/33552625814) on `45342b455`: **40/51 green, below the floor of 42.**

```
GATE: green count 40 fell below floor 42.
GATE: app(s) listed in greenApps are no longer green:
  - tools/cs2gs/Cs2Gs.Cli/Cs2Gs.Cli.csproj
  - tools/cs2gs/Cs2Gs.Pipeline/Cs2Gs.Pipeline.csproj
  - tools/cs2gs/Cs2Gs.Report/Cs2Gs.Report.csproj
```

The `greenApps` composition ratchet added in #3768 is what caught this by name.

## One defect, three cascades

All four failing cs2gs apps report the **same single fingerprint**:

```
GS0155: Cannot convert type 'string?' to 'string'
  tools/cs2gs/Cs2Gs.Pipeline/RepositoryMirror.gs:171:34
```

`Cs2Gs.Pipeline` is the source; `Cs2Gs.Cli`, `Cs2Gs.Report` and `Cs2Gs.Tests` fail only because they consume it.

## Root cause: gsc does not implement `[NotNullIfNotNull]`

`RepositoryMirror.DestinationRelativePath` is a `string`-returning method whose body is:

```csharp
return Path.ChangeExtension(source, ".gs");   // and .gsproj, .slnx
```

`Path.ChangeExtension` is declared in the BCL as:

```csharp
[return: NotNullIfNotNull(nameof(path))]
public static string? ChangeExtension(string? path, string? extension);
```

C# accepts the `return` because `source` is non-null, so the conditional post-condition makes the result non-null. **gsc has no `[NotNullIfNotNull]` support** — the attribute appears nowhere in `ClrNullability` — so it sees only the declared `string?` and rejects the conversion.

## This is fallout from a correct fix

Commit `281a9c278` (#3705 family 2) made gsc read imported member nullability correctly and uniformly, which is right and fixed a real unsoundness. But correctness here **exposed an under-approximation**: gsc now sees the `string?` it always should have seen, and lacks the mechanism C# uses to narrow it back. Before that fix, this site bound by accident.

Do **not** address this by reverting or weakening `281a9c278`.

## Suggested direction (verify before building)

[...]

PR #3806, opened 2026-09-02
cs2gs: forward [NotNullIfNotNull] nullability to the named argument (#3802)

DavidObando:
Second half of #3802. The compiler half landed in #3803; the gate's four cs2gs apps stayed red on the same fingerprint, and the run artifact says why.

## The issue's premise was wrong about the failing site

I ran the gate on the #3803 branch ([33566832379](https://github.com/DavidObando/gsharp/actions/runs/33566832379)) and pulled the migrated tree. `RepositoryMirror.gs:171:34` is **not** `return Path.ChangeExtension(source, ".gs")`. Line 171 column 34 is 11 characters wide and it is `destination`:

```
private func ValidateCollisions(files IEnumerable[string]) {
    let destinations = Dictionary[string, string](StringComparer.OrdinalIgnoreCase)
    for source in files {
        for destination in DestinationRelativePaths(source)!! {
            ...
            destinations[destination] = source     // <- GS0155, line 171
```

`destination` is `string?` because the **translator** widened the producer:

```
private func DestinationRelativePaths(source string) sequence[string?] {
    let extension string? = Path.GetExtension(source)
    if extension!!.Equals(".cs", StringComparison.OrdinalIgnoreCase) {
        yield Path.ChangeExtension(source, ".gs")
```

The C# is a plain `IEnumerable<string>` iterator with `string extension = Path.GetExtension(source)`. gsc was right to reject the `string?` key; the defect is upstream, in cs2gs.

## Root cause

`Path.GetExtension`, `Path.GetFileName` and `Path.ChangeExtension` are declared `string?` with `[return: NotNullIfNotNull(nameof(path))]`. Both nullability judgements in `Cs2Gs.Translator` — `ObliviousNullabilityAnalyzer.IsDirectlyNullable` and the translator's `IsNullableInitializer` — end in the same fallback: *"otherwise consult the bound symbol's DECLARED annotation"*. That is right for a plainly nullable member and blind to a conditional one. So `extension` is promoted to `string?`, the promoted yield taints the iterator's element to `sequence[string?]`, and that element escapes into the `Dictionary[string, string]` index. One root cause, three symptoms, four red apps.

## The rule: a forwarding EDGE, not a verdict

Such a call is **exactly as nullable as its named argument**. Both analyses now answer for the argument, and `CollectScalarValueFlow` records `target ← argument` so the whole-program fixpoint carries the relationship.

The first draft answered *"non-null"* outright, and the pinned Oahu corpus proved that wrong — 9 of 15 apps, one fingerprint. I reproduced it locally and read the diagnostics rather than inferring; the mechanism is your option **(b)**, a layer disagreement, and specifically:

> gsc refuses to narrow a call whose argument it sees as `T?`, and the argument's **emitted** type is decided by the very taint fixpoint this judgement feeds. A verdict taken from the argument's *C# syntax* alone (in oblivious C# nothing is annotated, so every parameter reads as "not nullable") therefore disagrees with gsc wherever the fixpoint later promotes that argument.

Oahu's `Oahu.Core/ExtensionsVarious.GetDownloadFileNameWithoutExtension` came out as `func (downloadFileName string?) ... string` — a `string?` argument feeding a `string` return, GS0155. Its sibling line showed the inconsistency plainly: `Path.GetExtension(downloadFileName)!!` still carried a `!!` from a pass that *did* know the parameter was nullable.

Forwarding keeps the post-condition conditional in **both** layers at once. Locally, the same pinned Oahu commit now migrates **15/15 apps green** (translate + compile + ILVerify + test-parity), against 6/15 on the first draft.

## On the layer choice (cs2gs vs gsc)

Your brief asked for a flow fact in gsc, and that is what shipped — `ConditionalReturnNarrowing` in #3803, applied at the call, conditional on the argument, already merged. This PR is not an alternative to it; both are required and neither is sufficient alone. cs2gs decides the **declared** types of the G# it emits (the iterator's element, the local, the promoted parameter) before gsc ever sees the file, using its own nullability model. gsc's narrowing cannot un-declare a `sequence[string?]` that cs2gs already wrote. Conversely, with the element back at `sequence[string]`, `yield Path.ChangeExtension(source, ".gs")` needs #3803's narrowing to bind — which is exactly what this PR's test `AssertBinds` exercises.

Your inference from the blast radius was still the right instinct, and it pointed at the real bug: a taint-level answer has a much larger surface, which is why the rule here had to become a forwarding edge that defers to the fixpoint rather than a standalone verdict.

## Test evidence

`tools/cs2gs/Cs2Gs.Tests/Issue3802ConditionalNotNullPostconditionTests.cs`.

**Fails on `origin/main`, passes here:**
- `ConditionalPostcondition_KeepsTheIteratorElementNonNullable` — the reduced `RepositoryMirror` shape end to end (iterator element + local + the dictionary index that consumes them), then bound with gsc.
- `PromotedArgument_ForwardsItsNullabilityToTheResult` — the Oahu shape: the `== null` promotes the parameter, the return carries no direct null at all, so it can only become `string?` via the forwarding edge. This is the regression guard for the first draft.

**Passes on `origin/main` too — guard rails:**
- `NullableArgument_StillPromotesTheResult` — the post-condition is CONDITIONAL, so a nullable argument must still promote. Narrowing unconditionally would emit a `string` declaration for a value that really can be nil and hide the nullability the migration exists to surface — the #3705 family 2 unsoundness in translator clothing. (gsc's own conditional guard is `NullableArgument_StillYieldsNullableResult` in #3803, which compiles and asserts `GS0155`.)
- `UnannotatedNullableReturn_StillPromotes` — `Path.GetDirectoryName`, on the same type, deliberately **not** annotated (it returns null for a root path even with a non-null argument), still promotes. The rule keys off the attribute, not the type.

`tools/cs2gs/selfmig-baseline.json` is untouched.

### DIFF
--- a/tools/cs2gs/Cs2Gs.Tests/Issue3802ConditionalNotNullPostconditionTests.cs
+++ b/tools/cs2gs/Cs2Gs.Tests/Issue3802ConditionalNotNullPostconditionTests.cs
@@ -1,0 +1,192 @@
+// <copyright file="Issue3802ConditionalNotNullPostconditionTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using Cs2Gs.CodeModel.Ast;
+using Cs2Gs.CodeModel.Printing;
+using Cs2Gs.Translator;
+using Cs2Gs.Translator.Loading;
+using Xunit;
+
+namespace Cs2Gs.Tests;
+
+/// <summary>
+/// Issue #3802: <c>Path.GetExtension</c>, <c>Path.GetFileName</c> and
+/// <c>Path.ChangeExtension</c> are declared <c>string?</c> with
+/// <c>[return: NotNullIfNotNull(nameof(path))]</c> — the result is non-null
+/// whenever the named argument is. Both nullability judgements in the
+/// translator ended in the same "consult the DECLARED annotation" fallback,
+/// which cannot see a CONDITIONAL post-condition.
+/// <para>
+/// The cost was concrete: migrating
+/// <c>tools/cs2gs/Cs2Gs.Pipeline/RepositoryMirror.DestinationRelativePaths</c>
+/// produced <c>sequence[string?]</c> with <c>let extension string? =
+/// Path.GetExtension(source)</c> and <c>extension!!.Equals(...)</c>, and the
+/// <c>string?</c> element then flowed into a <c>Dictionary[string, string]</c>
+/// index in <c>ValidateCollisions</c>, which gsc rightly rejected with
+/// <c>GS0155</c> — the whole of the gate's 40/51 regression.
+/// </para>
+/// </summary>
+public class Issue3802ConditionalNotNullPostconditionTests
+{
+    /// <summary>
+    /// The migrated shape, reduced: the iterator element must stay
+    /// <c>string</c>, the local must stay <c>string</c>, and the dictionary
+    /// index that consumes them must bind.
+    /// </summary>
+    [Fact]
+    public void ConditionalPostcondition_KeepsTheIteratorElementNonNullable()
+    {
+        string printed = Translate(@"
+using System.Collections.Generic;
+using System.IO;
+
+namespace Demo
+{
+    public class Mirror
+    {
+        public static void ValidateCollisions(IEnumerable<string> files)
+        {
+            var destinations = new Dictionary<string, string>();
+            foreach (string source in files)
+            {
+                foreach (string destination in DestinationRelativePaths(source))
+                {
+                    destinations[destination] = source;
+                }
+            }
+        }
+
+        private static IEnumerable<string> DestinationRelativePaths(string source)
+        {
+            string extension = Path.GetExtension(source);
+            if (extension.Equals("".cs""))
+            {
+                yield return Path.ChangeExtension(source, "".gs"");
+                yield break;
+            }
+
+            yield return source;
+        }
+    }
+}");
+
+        Assert.Contains("sequence[string]", printed);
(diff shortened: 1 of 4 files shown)


DavidObando:
**Gate re-run on this branch: [33579419514](https://github.com/DavidObando/gsharp/actions/runs/33579419514) — 43/52 green (floor 42), up from 40 on the regressed run.** No app in `greenApps` is red; `test/Extensions.Tests` is newly green as a side effect.

Before ([33552625814](https://github.com/DavidObando/gsharp/actions/runs/33552625814), main) → after (this branch), the three ratcheted apps plus the fourth consumer:

| app | before | after |
| --- | --- | --- |
| `tools/cs2gs/Cs2Gs.Pipeline` | compile-error (GS0155) | **GREEN** |
| `tools/cs2gs/Cs2Gs.Cli` | compile-error (cascade) | **GREEN** |
| `tools/cs2gs/Cs2Gs.Report` | compile-error (cascade) | **GREEN** |
| `tools/cs2gs/Cs2Gs.Tests` | compile-error (cascade) | still red, on a **different** fingerprint (it was never in `greenApps`) |

Readability metrics moved the right way: `!!` **16570 → 10744** (ceiling 17050), labels 0, `__local_` 29.

The run still exits non-zero on `lines>300=595 (ceiling 580)`. That is the separate long-line ratchet — it was already breaching on main at 582 before this branch existed — so I have deliberately not folded a baseline edit into this PR.

`cs2gs-oahu` on the PR itself passes (19m11s). Locally the same pinned Oahu commit migrates 15/15 green, against 6/15 on the first draft.

Issue #3815, opened 2026-09-02
gsc: inside a GENERIC class deriving from an imported generic base, the base's members are seen at the base's own type arguments erased to object (Channel[T].Reader binds as ChannelReader[object])


DavidObando:
## Summary

Found while fixing #3813 (inherited `protected set`). With that fix in place, the *non-generic* derived class works end to end:

```
class WrapI : Channel[int32] {
    init(inner Channel[int32]) { Reader = inner.Reader; Writer = inner.Writer }
}
```
compiles, ILVerifies, and prints the round-tripped value.

The **generic** derived class does not:

```
class Wrap[T] : Channel[T] {
    init(inner Channel[T]) {
        Reader = inner.Reader
        Writer = inner.Writer
    }
}
```

```
Program.gs(9,18,9,30): error GS0155: Cannot convert type 'System.Threading.Channels.ChannelReader[T]'
                                  to 'System.Threading.Channels.ChannelReader[object]'.
Program.gs(10,18,10,30): error GS0155: Cannot convert type 'System.Threading.Channels.ChannelWriter[T]'
                                    to 'System.Threading.Channels.ChannelWriter[object]'.
```

The inherited member's type is not substituted with the derived type's own type argument — it comes back at `object`. The same `[object]` shows up in the read path's diagnostics.

## Impact

`bench/concurrency/clr/ClrBaseline` is walled by this after #3812 and #3813 are fixed: its `GoChan<T> : Channel<T>` is exactly this shape. The two errors the gate currently reports there (GS0113, GS0127) are fixed; this one is next behind them, so the app does not go green on those two fixes alone.

Reproduces standalone with only `System.Threading.Channels` referenced.

DavidObando:
**Premise correction + fix landed** (PR #3819).

Reproduced exactly as written. One correction:

> The same `[object]` shows up in the read path's diagnostics.

The **read path already projects correctly**. `ExpressionBinder.GetInheritedClrMemberType` (#1582) walks to the nearest `ImportedBaseType`, matches the member against the base's OPEN definition and substitutes the symbolic arguments — so `Reader` *reads* as `ChannelReader[T]` inside `Wrap[T]`. The `[object]` in those diagnostics is the assignment **target**, not the source: the defect is write-path-only.

[...]

Issue #3814, opened 2026-09-02
gsc/cs2gs: an explicit-interface clause on an IMPORTED GENERIC interface does not satisfy the interface contract, so two same-named implementations collapse into a duplicate overload (GS0187+GS0264; selfmig wall: test/Core.Tests)


DavidObando:
## Summary

A C# class that explicitly implements the **same member of two different instantiations** of an imported generic interface cannot round-trip today.

```csharp
public sealed class Issue3465ConflictingAsyncCollectionForward :
    IAsyncEnumerable<int>,
    IAsyncEnumerable<string>
{
    IAsyncEnumerator<int> IAsyncEnumerable<int>.GetAsyncEnumerator(CancellationToken ct)
        => new Issue3465EmptyAsyncEnumerator<int>();
    IAsyncEnumerator<string> IAsyncEnumerable<string>.GetAsyncEnumerator(CancellationToken ct)
        => new Issue3465EmptyAsyncEnumerator<string>();
}
```

cs2gs emits **no** ADR-0149 qualifier clause, because of the carve-out in `CSharpToGSharpTranslator.Members.cs`:

```csharp
bool hasClauseCompatibleExplicitInterfaceImpl = hasSingleExplicitInterfaceImpl &&
    (symbol.ExplicitInterfaceImplementations[0].ContainingType.Locations.Any(l => l.IsInSource) ||
     !symbol.ExplicitInterfaceImplementations[0].ContainingType.IsGenericType);
```

("Imported generic interfaces also fall back because gsc currently erases their clause arguments while resolving the CLR slot (GS0494).")

The two methods therefore become two plain overloads differing only in return type:

```
class ConflictingAsyncCollectionForward : IAsyncEnumerable[int32], IAsyncEnumerable[string] {
    func GetAsyncEnumerator(cancellationToken CancellationToken) IAsyncEnumerator[int32] -> …
    func GetAsyncEnumerator(cancellationToken CancellationToken) IAsyncEnumerator[string] -> …
}
```

```
GS0187 | Class 'ConflictingAsyncCollectionForward' does not implement interface method
         'System.Collections.Generic.IAsyncEnumerable[string].GetAsyncEnumerator(…)'
GS0264 | An overload of 'GetAsyncEnumerator' with signature 'GetAsyncEnumerator(CancellationToken)'
         is already declared.
```

## The carve-out is not the root cause — gsc is

Hand-writing the clause gsc *should* accept still fails:

```
[...]

Issue #3813, opened 2026-09-02
gsc: an imported CLR base class's 'protected set' property is reported read-only in a derived G# type (GS0127; selfmig wall: bench/concurrency/clr/ClrBaseline)


DavidObando:
## Summary

`ClrMemberVisibility` (#3705) admits only `public` and friend-`internal` accessors, by design and by its own doc comment. But the **inherited-CLR-base** write paths route their settable-ness question through it — the very paths whose comments promise inherited `protected` members (#319/#1582), and which already deliver it for `protected` **fields**.

So a `{ get; protected set; }` base property is reported read-only inside the derived type's own constructor, a write C# accepts:

```
import System.Threading.Channels
class MyChan[T] : Channel[T] {
    init() { Reader = nil }
}
```

```
prot.gs(6,16,6,17): error GS0127: Variable 'Reader' is read-only and cannot be assigned to.
```

Both the bare-name and the `this.`-qualified spelling are affected (they take different binder paths). A G#-**source** base with `prop Value int32 { get; protected set }` already worked, which localises the defect to the imported-CLR path.

## Impact

One of the two errors walling `bench/concurrency/clr/ClrBaseline` out of the #3501 self-migration gate (run 33584377318): its `GoChan<T> : Channel<T>` constructor assigns `Reader` and `Writer`, whose setters `System.Threading.Channels.Channel<T>` declares `protected`.

## Fix landed

`ClrMemberVisibility.GetDerivedVisibleSetter` admits `family` / `famorassem` (and `famandassem` under friendship), and a `fromDerivedType` bit is threaded through the four inherited-base write sites in `ExpressionBinder.Assignments.cs` plus the bare-name write path in `ExpressionBinder.cs`. Guard rails asserted: a get-only inherited property is still rejected, and a `protected` setter is still unreachable from a non-derived type.

DavidObando:
Fixed in PR #3816 (merged as `832a6fb11`), which addressed all three of #3811, #3812 and #3813 together.

Evidence recorded there: 10 new cases in `test/Compiler.Tests`, every one compiling + ILVerifying + running + asserting stdout; 6 of the 10 fail with `src/Core` reverted to `origin/main`, and the 4 that pass on both sides are labelled anti-vacuity guard rails.

Two findings from that work worth keeping attached to these issues:

[...]

Issue #3812, opened 2026-09-02
gsc: a nested type's field INITIALIZER cannot name the enclosing type's type parameters (GS0113; selfmig wall: bench/concurrency/clr/ClrBaseline)


DavidObando:
## Summary

Issue #1537 gave nested-type member binding the enclosing type's type parameters. Field **initializers** never got them: they bind from a deferred closure (#1194) that runs after `BindStructDeclarationBody` has unwound its scope and rebuilds it from `structSymbol.TypeParameters` alone, dropping `CollectEnclosingTypeParameters(structSymbol.ContainingType)`.

The result is an initializer-only asymmetry — the field's *declared type* resolves, its *initializer* does not:

```
class Outer[T] {
    class Inner {
        var item T = default(T)
    }
}
```

```
n3.gs(5,30,5,31): error GS0113: Type 'T' doesn't exist.
            var item T = default(T)
```

Note the column: 30 is the `T` inside `default(T)`, not the declared type at column 23. Dropping the initializer compiles fine, which is why no existing fixture caught it.

## Impact

One of the two errors walling `bench/concurrency/clr/ClrBaseline` out of the #3501 self-migration gate (run 33584377318). Its `Hchan<T>.Waiter` nested class carries `internal T item = default!;`, translated to `internal var item T = default(T)!!`.

## Fix landed

Mirror the #1537 seeding inside the deferred closure, outermost-first so an inner level shadows an outer one on a name clash.

DavidObando:
Fixed in PR #3816 (merged as `832a6fb11`), which addressed all three of #3811, #3812 and #3813 together.

Evidence recorded there: 10 new cases in `test/Compiler.Tests`, every one compiling + ILVerifying + running + asserting stdout; 6 of the 10 fail with `src/Core` reverted to `origin/main`, and the 4 that pass on both sides are labelled anti-vacuity guard rails.

Two findings from that work worth keeping attached to these issues:

- The first attempt at #3811 was **itself a regression**, caught by CI (`ilverify` 129→127 plus three test shards). Comparing assembly *identity* against the core library was the wrong question; the correct discriminator is **external visibility** — an embedded copy is always `internal`, while every targeting pack since .NET 5 declares `NullableAttribute` **public** in `System.Runtime`. The landed fix is one line.
[...]

PR #3816, opened 2026-09-02
gsc: three self-migration compile walls — embeddable-attribute scope (GS9998 ICE), nested-type initializer scope, inherited protected setters (#3811, #3812, #3813)

DavidObando:
Issue #3501 Track C. Triage of the three smallest compile-stage walls in the self-migration gate (run **33584377318**, 43/52 green), then the fixes that were tractable.

Every diagnosis below was reproduced **locally and standalone** — a two-project scratch corpus for two of them, a three-step `gsc` invocation for the third — so none of it needed a repo-scale migration to iterate on.

## Root causes

| # | cause | app | errors | layer | new / known / cascade |
|---|---|---|---|---|---|
| #3811 | `NullableAttribute`/`NullableContextAttribute` TypeRefs scoped to a third-party assembly's **embedded internal** copy; the next compilation cannot resolve them → **GS9998 ICE** | `test/Sdk.Tests` | 1 | gsc emit (`WellKnownReferences`) | **known family, new instance — a reappearance of #3755** (emitting a reference to a type the consuming target lacks) |
| #3812 | a nested type's field **initializer** binds in a deferred scope rebuilt without the enclosing type's type parameters (#1537 not mirrored into #1194's closure) | `bench/…/ClrBaseline` | 1 of 2 | gsc bind (`DeclarationBinder.Structs`) | new |
| #3813 | an imported CLR base's `protected set` accessor is invisible to the derived type, on the very paths that promise inherited protected members (#319/#1582) | `bench/…/ClrBaseline` | 1 of 2 | gsc bind (`ClrMemberVisibility` / `ExpressionBinder`) | new; a gap left by #3705's blanket "protected is never visible" rule |
| #3814 | an explicit-interface clause on an **imported generic** interface does not satisfy the contract, so cs2gs's carve-out emits two same-signature overloads | `test/Core.Tests` | 2 | gsc bind (ADR-0149) + cs2gs carve-out | **filed, not fixed** — needs the gsc half first |
| #3815 | a generic class deriving from an imported generic base sees the base's members at `object` | `bench/…/ClrBaseline` | 0 today | gsc bind | **filed, not fixed** — found by the #3813 fix; sits behind #3812/#3813 |

## Loudest finding

`test/Sdk.Tests` was an **internal compiler error**, and the empty assembly name in `in assembly ''` was the tell. Compiling the real migrated `src/Sdk/Gsharp.NET.Sdk` (netstandard2.0) and dumping its metadata:

```
--- AssemblyRefs ---
  [1] 'netstandard' 2.0.0.0
  [2] 'Microsoft.Build.Framework' 15.1.0.0
  [3] 'Microsoft.Build.Utilities.Core' 15.1.0.0
  [4] 'System.Private.CoreLib' 10.0.0.0
--- TypeRefs mentioning Nullable ---
  System.Runtime.CompilerServices.NullableAttribute         scope=asmref:Microsoft.Build.Framework
  System.Runtime.CompilerServices.NullableContextAttribute  scope=asmref:Microsoft.Build.Framework
```

Ten assemblies in that netstandard2.0 closure carry an embedded internal copy. (The `System.Private.CoreLib 10.0.0.0` ref in a netstandard2.0 assembly is **#3769**, still open, confirmed independently by the same dump.)

After the fix the two bad rows are gone and the AssemblyRef set is otherwise unchanged.

## Before / after

| app | before (gate 33584377318) | after |
|---|---|---|
| `test/Sdk.Tests` | compile FAIL(1) — GS9998 `TypeLoadException … in assembly ''` | reproduced standalone and fixed; **0 errors** in the reduced end-to-end repro |
| `bench/concurrency/clr/ClrBaseline` | compile FAIL(2) — GS0113, GS0127 | both fixed; **still not green** — #3815 is next behind them (stated, not hidden) |
| `test/Core.Tests` | compile FAIL(2) — GS0187, GS0264 | unchanged; filed as #3814 with the gsc-first ordering worked out |

Reduced-repro before/after for #3811, on the exact three-step shape (`origin/main` gsc vs this branch):

```
main:  step 2 → GS9997 Could not find assembly 'Microsoft.Build.Framework, Version=15.1.0.0…'
main:  step 3 → GS9998 TypeLoadException: Could not find type
                'System.Runtime.CompilerServices.NullableContextAttribute' in assembly ''
this:  step 2 → Wrote App.dll
this:  step 3 → Wrote App2.dll
```

## Test evidence

`test/Compiler.Tests/Issue3811…`, `Issue3812…`, `Issue3813…` — **10 cases, every one compiles + ILVerifies + runs + asserts the program's own stdout.** No binding-only assertions.

Failing-on-`origin/main` proof (source reverted, tests kept):

```
Failed!  - Failed: 6, Passed: 4, Total: 10
```

with the fixes:

```
Passed!  - Failed: 0, Passed: 10, Total: 10
```

The **4 that pass on both sides are the deliberate anti-vacuity guard rails**, labelled as such:
- nullability rows are *still emitted* against the targeting pack, scoped to `System.Runtime` (so #3811 did not simply silence emission);
- a genuinely get-only inherited property is still rejected;
- a `protected` setter is still unreachable from a **non-derived** type;
- an inner type parameter shadowing a same-named outer one already worked and must keep working.

`build/nullable_hygiene.py`: OK.

## Premise corrections

- The brief expected `test/Core.Tests` to be "an explicit interface implementation translated into an ordinary overload (or vice versa)". Half right: cs2gs *does* drop the clause — but only because of a deliberate carve-out, and **adding the clause by hand does not fix it**. gsc still reports GS0187, so the root cause is gsc's ADR-0149 interface verification, not the translator. Fixing cs2gs first would make the diagnostics worse.
- The brief suggested the `NullableAttribute` crash might be a *synthesized*-attribute emit path. It is the opposite: gsc **never** synthesizes one — that is precisely the gap — and instead borrows someone else's.
- ClrBaseline's two errors are two independent defects, as predicted, but fixing them does **not** turn the app green: a third defect (#3815) sits behind them.

## Deconfliction

No files in the cs2gs linked-source / name-spelling path were touched, so there is no overlap with the agent working #3805. `tools/cs2gs/selfmig-baseline.json` is untouched (PR #3810 is open against it) — no app moves to green in this PR, so no ratchet is due.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/DeclarationBinder.Structs.cs
+++ b/src/Core/CodeAnalysis/Binding/DeclarationBinder.Structs.cs
@@ -3001,12 +3001,36 @@ internal sealed partial class DeclarationBinder
             // `structSymbol`, mirroring how a `shared` function body (which
             // already works) carries its `StaticOwnerType`.
             setCurrentFunction(CreateFieldInitializerAccessibilityContext(structSymbol));
-            if (!structSymbol.TypeParameters.IsDefaultOrEmpty)
+
+            // Issue #3812: this closure runs long after BindStructDeclarationBody
+            // unwound its type-parameter scope, so the scope has to be rebuilt
+            // from scratch — and it must be rebuilt the SAME way, including the
+            // issue #1537 enclosing-type parameters. Seeding only
+            // `structSymbol.TypeParameters` left a nested type's initializer
+            // unable to name the ENCLOSING type's parameters, so
+            // `class Outer[T] { class Inner { var item T = default(T) } }` bound
+            // the field's declared type fine (that happens inside the body scope)
+            // and then failed on the initializer with GS0113 "Type 'T' doesn't
+            // exist" — an initializer-only asymmetry, invisible to any fixture
+            // whose nested field had no initializer.
+            var initializerEnclosingTypeParameters = CollectEnclosingTypeParameters(structSymbol.ContainingType);
+            if (!structSymbol.TypeParameters.IsDefaultOrEmpty || initializerEnclosingTypeParameters.Count > 0)
             {
                 binderCtx.CurrentTypeParameters = new Dictionary<string, TypeParameterSymbol>();
-                foreach (var tp in structSymbol.TypeParameters)
+
+                // Outermost-first, so an inner level shadows an outer one on a
+                // name clash — the same precedence BindStructDeclarationBody uses.
+                foreach (var tp in initializerEnclosingTypeParameters)
                 {
                     binderCtx.CurrentTypeParameters[tp.Name] = tp;
+                }
+
+                if (!structSymbol.TypeParameters.IsDefaultOrEmpty)
+                {
+                    foreach (var tp in structSymbol.TypeParameters)
+                    {
+                        binderCtx.CurrentTypeParameters[tp.Name] = tp;
+                    }
                 }
             }
 
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs
@@ -573,7 +573,7 @@ internal sealed partial class ExpressionBinder
                 inhMember ??= ClrTypeUtilities.SafeGetInheritedInstanceField(inheritedBaseClr, propertyName);
                 if (inhMember != null)
                 {
-                    if (!TryGetWritableClrMember(inhMember, out _, out var inhTargetSymbol, out _))
+                    if (!TryGetWritableClrMember(inhMember, out _, out var inhTargetSymbol, out _, fromDerivedType: true))
                     {
                         Diagnostics.ReportCannotAssign(initSyntax.EqualsToken.Location, propertyName);
                         _ = BindExpression(initSyntax.Value);
@@ -682,7 +682,7 @@ internal sealed partial class ExpressionBinder
             {
                 var importedBase = new ImportedClassSymbol(importedBaseClr, syntax, references: scope.References);
                 if (importedBase.TryLookupMember(fieldName, ne: null, out var inheritedStaticMember)
-                    && TryGetWritableClrMember(inheritedStaticMember, out _, out var inheritedTarget, out _))
+                    && TryGetWritableClrMember(inheritedStaticMember, out _, out var inheritedTarget, out _, fromDerivedType: true))
                 {
                     var inheritedValue = BindAssignmentRhs(syntax.Value, inheritedTarget);
                     var inheritedConverted = conversions.BindConversion(syntax.Value.Location, inheritedValue, inheritedTarget);
@@ -1079,7 +1079,7 @@ internal sealed partial class ExpressionBinder
                 clrMember ??= ClrTypeUtilities.SafeGetInheritedInstanceField(inheritedBaseClr, memberName);
                 if (clrMember != null)
                 {
-                    if (!TryGetWritableClrMember(clrMember, out var inhTargetType, out var inhTargetSymbol, out var inhWritable))
+                    if (!TryGetWritableClrMember(clrMember, out var inhTargetType, out var inhTargetSymbol, out var inhWritable, fromDerivedType: true))
                     {
                         Diagnostics.ReportCannotAssign(syntax.EqualsToken.Location, memberName);
                         return new BoundErrorExpression(null);
@@ -2706,7 +2706,7 @@ internal sealed partial class ExpressionBinder
                 clrMember ??= ClrTypeUtilities.SafeGetInheritedInstanceField(inheritedBaseClr, fieldName);
                 if (clrMember != null)
                 {
-                    if (!TryGetWritableClrMember(clrMember, out var inhTargetType, out var inhTargetSymbol, out var inhWritable))
+                    if (!TryGetWritableClrMember(clrMember, out var inhTargetType, out var inhTargetSymbol, out var inhWritable, fromDerivedType: true))
                     {
                         Diagnostics.ReportCannotAssign(syntax.EqualsToken.Location, fieldName);
                         return new BoundErrorExpression(null);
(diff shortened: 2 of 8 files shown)


DavidObando:
### Correction pushed (88c193ff9) — the first form of the #3811 fix was itself a regression

CI caught it: `ilverify` went 129/129 → **127/129**, with two samples failing to *compile* because `Gsharp.Extensions`' nullability metadata had vanished and every reference type in it read back as nullable-oblivious:

```
samples/GsharpExtensionsSequences.gs(22,10,22,32): error GS0116:
    Type 'System.Collections.Generic.IEnumerable[int32]?' is not indexable.
```

The `tests (fast-suites)`, `tests (compiler-conformance)`, `tests (interpreter)` and `cs2gs-code-exploder` failures in the same run have the same shape and the same cause.

**Why it was wrong.** Comparing `resolved.Assembly` against `emitCtx.CoreObjectType.Assembly` asked the wrong question. The right discriminator is **external visibility**: a csc-*embedded* copy is always `internal`, while every targeting pack from .NET 5 onwards declares `System.Runtime.CompilerServices.NullableAttribute` as a **public** type of `System.Runtime` (verified directly in the net10.0 ref pack: `vis=Public`). And `TryResolveType`'s externally-visible overload already skips an inaccessible shim and keeps searching for an accessible duplicate (#3445) — exactly the behaviour wanted here.

The fix is now a one-liner: `requireExternalVisibility: true`.

**Probe with `src/Sdk/Gsharp.Extensions`' own 170-assembly reference set:**

```
origin/main   NullableAttribute        scope=asmref:Microsoft.VisualStudio.Validation
              NullableContextAttribute scope=asmref:Microsoft.VisualStudio.Validation
this branch   NullableAttribute        scope=asmref:System.Runtime
              NullableContextAttribute scope=asmref:System.Runtime
```

Worth stating plainly: **main is already emitting a bad row for a first-party repo assembly.** It has simply never crashed, because `Microsoft.VisualStudio.Validation` happens to be in every consumer's closure. `Gsharp.NET.Sdk` was where the same defect finally had no such luck.

The regression test's donor assembly now declares its copies `internal`, which is how csc actually embeds them — a stricter and more faithful model. It still fails on `origin/main` and passes here.

**Re-verified locally after the correction:**
- `build/run-ilverify.sh` → **129/129 verified, 0 compile failures**
- the 10 new cases → 10/10 pass; **6/10 still fail with `src/Core` reverted to `origin/main`**
- `build/nullable_hygiene.py` → OK
- `Gsharp.Extensions.dll` rebuilt with a pure-`main` compiler and with this branch's compiler is **byte-for-byte equivalent** in its nullability rows (both `scope=asmref:System.Runtime`)

DavidObando:
### Scope note on the #3811 proof

What is proven, and how:

1. **The bad row exists on `main`, in the real migrated tree.** Compiled the actual migrated `src/Sdk/Gsharp.NET.Sdk` sources with the actual resolved netstandard2.0 reference set and dumped the metadata — `NullableAttribute`/`NullableContextAttribute` scoped to `Microsoft.Build.Framework`. On this branch they are gone. Same experiment for `src/Sdk/Gsharp.Extensions` (a **shipped, non-migrated** assembly): `main` scopes them to `Microsoft.VisualStudio.Validation`, this branch to `System.Runtime`.
2. **The crash follows from the bad row.** A three-step reduction — internal-attribute donor → netstandard2.0 library → net10.0 consumer — produces `GS9998 TypeLoadException: Could not find type '…NullableContextAttribute' in assembly ''` on `origin/main` and compiles clean here. That is the gate's diagnostic, verbatim, empty assembly name and all.

What is **not** proven end to end: I did not run a full `migrate` + `validate` of `test/Sdk.Tests` against a repacked SDK nupkg. A trivial consumer of the `main`-built migrated `Gsharp.NET.Sdk.dll` does **not** crash — gsc only reads the nullability metadata of members it actually touches, so the crash needs a consumer that binds against the affected surface, which `Sdk.Tests` does and a two-line probe does not. The reduction in (2) is what closes that gap; a gate run is the confirmation.

Also worth flagging for whoever picks up **#3769**: with #3811 fixed, the migrated netstandard2.0 SDK scopes the two attributes at `System.Private.CoreLib` (the host-fallback leak) rather than at a contract assembly. That AssemblyRef was already present on `main` in the same assembly, so this is pre-existing and unchanged by this PR, but it is the next thing standing between that output and a genuinely portable netstandard2.0 assembly.

DavidObando:
### CI green — 35/35

Both failures on the first pass of the corrected commit were flakes; a re-run of just those two jobs is clean.

| job | first pass | re-run |
|---|---|---|
| `tests (compiler-emit-remainder)` | `Int32_To_Enum_Cast_Produces_Corresponding_Enum_Value` — `GS0156: Cannot convert type 'P.Color' to 'Color'` | pass |
| `tests (cs2gs-remainder)` | `Adr0169AnalyzerTranslationTests.SemanticModelAnalyzer_TranslatesDeclarationAndOverrideIdioms` — `Cannot find function FirstAncestorOrSelf` | pass |

Three independent reasons to read these as pre-existing flakes rather than fallout from this PR:

1. **Neither shard runs any test this PR adds.** `emit-remainder` filters on `GSharp.Compiler.Tests.Emit.`; the new fixtures are in `GSharp.Compiler.Tests`. `cs2gs-remainder` runs `Cs2Gs.Tests`, which this PR does not touch at all.
2. **Neither reproduces locally on the corrected tree** — individually, as a whole class (`ConversionEmitTests` 79/79), or across `SampleConformanceTests` (131/131).
3. **No mechanism.** The changes are: two attribute names resolved by external visibility at *emit* time; a type-parameter scope rebuilt in a *deferred field-initializer* closure; and a new `ClrMemberVisibility.GetDerivedVisibleSetter` reached only from the inherited-base *write* paths. None is on the path to enum conversion or to member lookup on an imported `SyntaxNode`.

On the first symptom specifically: `ConversionEmitTests.CompileToAssembly` passes gsc **no** `/reference:` arguments and then `Assembly.LoadFrom`s its output into the default load context. A later compilation in the same process can therefore see an earlier test's `test.dll` — which declares its own `P.Color` — and bind the bare `Color` annotation to the stale identity. That is an order-dependent hazard in the harness, latent on `main`, and worth a separate issue if it recurs.

Issue #1582, opened 2026-07-01
gsc: inherited members from metadata (BCL) base classes not fully resolved (unqualified property/field lookup; qualified field lookup)


DavidObando:
When a G# type derives from a **metadata (BCL / referenced-assembly) base
class**, inherited-member resolution is incomplete. Two distinct defects
(both reproduce standalone; a user-defined base class works in every case):

## Defect A — unqualified simple-name lookup skips the metadata base chain for properties and fields

```gsharp
package p
import System
class A : Exception { func F() string { return Message } }
```
`return Message` → `GS0125: Variable 'Message' doesn't exist.`

Yet the same member resolves when qualified (`return this.Message` compiles),
and an inherited **method** resolves unqualified (`return ToString()`
compiles). So unqualified simple-name lookup walks the metadata base chain for
methods but not for properties/fields.

## Defect B — qualified member lookup does not find inherited fields from a metadata base

```gsharp
package p
import System.Security.Cryptography
open class A : HashAlgorithm {
    func Initialize() { }
    protected func HashCore(a []uint8, s int32, c int32) { }
    protected func HashFinal() []uint8 {
        this.HashValue = []uint8{}
        return this.HashValue
    }
}
```
`this.HashValue` → `GS0158: Cannot find member HashValue.` (twice)

`HashValue` and `HashSizeValue` are inherited protected fields of
`System.Security.Cryptography.HashAlgorithm`. Qualified inherited *properties*
(e.g. `this.Message`) resolve, but inherited *fields* do not.

## Contrast — user-defined base works in every case

```gsharp
package p
open class B { protected var fld int32
 prop P int32 { get { return 1 } } func M() int32 { return 2 } }
class A : B {
    func UP() int32 { return P }        // ok
    func QP() int32 { return this.P }   // ok
    func UF() int32 { return fld }      // ok
    func QF() int32 { return this.fld } // ok
    func UM() int32 { return M() }      // ok
}
```
All compile. The defects are specific to base classes loaded from metadata.

## Impact

Blocks idiomatic subclassing of BCL base types (e.g. `HashAlgorithm`,
`Exception`, `TypeConverter`) in translated Oahu code. Surfaced by cs2gs
migrating `Crc32 : HashAlgorithm` (`HashValue = hashBuffer` → GS0125) and
several `Exception`/converter subclasses.

## Expected

Inherited members (methods, properties, and fields, of any accessibility
visible to the derived type) from a metadata base class should resolve
[...]

Issue #949, opened 2026-06-22
Allow class to declare itself on the base clause type parameter position


DavidObando:
The following construct should be legal in G#:
```
open class Shape : IEquatable[Shape]
...
```

Issue #319, opened 2026-05-31
Interpreter does not model CLR-base instance state for user classes


DavidObando:
## Summary

Issue #306 adds user-defined `init(...)` constructors and `: base(args)` initializers for GSharp classes. The **emit** path correctly chains to CLR base constructors (e.g. `System.Exception(string)`), so inherited state such as `Exception.Message` is set on compiled-and-run programs.

The **interpreter** (`Evaluator`) models GSharp class instances as `StructValue` field dictionaries rather than real CLR objects. As a result, when a GSharp class inherits an imported CLR base type, the base constructor is **not** actually invoked during interpretation — `: base(clrArgs)` is a no-op for CLR bases. Inherited CLR members (e.g. `Message`, `HResult`) therefore read as their defaults under the interpreter, even though they are correct under emit.

This is a pre-existing architectural limitation surfaced (not introduced) by #306: the same gap applies to primary-constructor `: Base(args)` initializers against CLR bases.

## Scope
- GSharp-on-GSharp base chaining works in the interpreter (primary ctor + `init` `: base(args)`); the field dictionary is populated.
- GSharp-on-CLR base chaining only works under emit.

## Possible directions
- Have the interpreter instantiate a backing CLR object for classes with an imported CLR base and forward base-ctor arguments to it, bridging field/member access.
- Or document the interpreter as emit-parity-only for CLR inheritance and steer such scenarios through the compile path.

## References
- Issue #306, PR #317
- `src/Core/CodeAnalysis/Evaluator.cs` — `EvaluateConstructorCallExpression` (explicit-ctor branch notes the no-op).

Issue #3818, opened 2026-09-02
test: ConversionEmitTests.CompileToAssembly has no reference isolation, so emit shards can flake on test order


DavidObando:
Found while investigating two CI failures on PR #3816 that turned out to be flakes (both clean on re-run, neither shard runs that PR's tests).

## The mechanism

`ConversionEmitTests.CompileToAssembly` invokes `gsc` with **no `/reference:` arguments**, then `Assembly.LoadFrom`s the resulting output. Because the assembly stays loaded in the test host process, a **later** compile in the same process can bind to a stale type from an earlier test — the observed case being a `P.Color` from a previous fixture.

That makes the outcome depend on **test execution order within a shard**, which is exactly the profile of an intermittent failure that cannot be reproduced locally when the file is run on its own.

## Why it is worth fixing rather than tolerating

An order-dependent flake in an emit shard is expensive out of proportion to its frequency:

- it costs a re-run and an investigation every time it fires;
- worse, it **erodes the flake/real distinction**. This effort has already had one genuine regression initially argued as a flake, and one true flake correctly identified only because someone ran 248/248 locally plus a green re-run. A known order-dependent failure in the same shards makes that judgement harder every time.

## Suggested direction

Give each `CompileToAssembly` compilation an isolated resolution context so it cannot bind to a previously-loaded assembly — pass an explicit `/reference:` closure rather than none, and/or load into a collectible `AssemblyLoadContext` that is unloaded after the assertion. Unique per-test assembly/type names would reduce collisions but would not fix the underlying isolation gap.

Worth checking whether sibling emit helpers share the pattern; if they do, fix the helper rather than the one call site.

## Provenance

Reported by the agent on PR #3816 (issues #3811/#3812/#3813) after its two remaining CI failures re-ran clean. Recorded so the next person who hits an `emit-remainder` flake does not re-derive it.

Refs #3501, #3816.

DavidObando:
**Second instance, different shard and helper — this may be broader than `ConversionEmitTests`.**

PR #3810 changes **exactly one file**, `tools/cs2gs/selfmig-baseline.json` (a data file no test in `Cs2Gs.Tests` reads — verified by grep), and `tests (cs2gs-remainder)` failed on it with:

```
Cs2Gs.Tests.Adr0169AnalyzerTranslationTests.SemanticModelAnalyzer_TranslatesDeclarationAndOverrideIdioms
  Translated analyzer should bind against GSharp.Core:
[...]

PR #3819, opened 2026-09-02
gsc: members of imported GENERIC types resolve through the substitution (#3814, #3815)

### LINKED ISSUE #3814: gsc/cs2gs: an explicit-interface clause on an IMPORTED GENERIC interface does not satisfy the interface contract, so two same-named implementations collapse into a duplicate overload (GS0187+GS0264; selfmig wall: test/Core.Tests)

### LINKED ISSUE #3815: gsc: inside a GENERIC class deriving from an imported generic base, the base's members are seen at the base's own type arguments erased to object (Channel[T].Reader binds as ChannelReader[object])

DavidObando:
Issue #3501 Track C. Fixes #3814 and #3815 — the two gsc binder walls PR #3816 filed but did not fix.

## Do they share a root? No — and I proved it, rather than assuming

Both read as "an imported generic type's members do not resolve through a substitution", which is exactly the shape that has collapsed 3–5× on this effort. They are **distinct defects in disjoint code**, and the cross-check is empirical, not architectural: with **only** the #3815 fix in the compiler and the #3814 fix stubbed out, the #3815 repro compiles clean (0 errors) while the #3814 repro still reports its GS0187. Different symbol layer (overload-set construction vs. member-type projection), different files, neither fix touches the other's path.

### #3814 — the overload set, not the clause resolver

The issue's ordering ("make `ResolveExplicitInterfaceClauses` / `VerifyInterfaceImplementations` honour the generic instantiation") is **half right, and the diagnosis points one layer too high**. Verification never consults the clause: it asks the type's overload set whether some method satisfies the CLR slot. The method had already been **deleted from that set**.

`StructSymbol.GetMethodsIncludingInherited` deduplicates with `BoundScope.FunctionSignaturesEqual`, which — correctly for ordinary overloads — ignores the return type. Two explicit implementations of `IAsyncEnumerable[T].GetAsyncEnumerator(CancellationToken)`, one per instantiation, differ **only** in the return type, so the second was dropped as if it were an override of the first.

The tell was order-dependence, which the issue does not mention: whichever method is declared **second** is the one reported unimplemented. Swapping the two declarations moves the diagnostic; a third instantiation loses two.

```
main, methods declared int32-then-string:  GS0187 … IAsyncEnumerable[string].GetAsyncEnumerator
main, methods declared string-then-int32:  GS0187 … IAsyncEnumerable[int32].GetAsyncEnumerator
```

Fix: the hiding rule additionally requires the two methods to carry the **same explicit-interface clause target**. Methods without a clause compare `null == null` and hide exactly as before, so real overrides are untouched.

### #3814b — a second, distinct sub-defect behind the cs2gs carve-out

Removing the translator carve-out uncovered the defect its comment actually names. A clause on an imported generic interface closed over the declaring type's **own type parameter** —

```
class Value[T] : IEquatable[T] {
    private func (IEquatable[T]) Equals(other T) bool -> true
}
```

— never matched a slot: `target.ClrType` is the type-**erased** `IEquatable<object>`, so the slot search compared `Equals(object)` against a declaration saying `Equals(T)` and reported **GS0494**. The resolver now falls back to the interface's OPEN definition with its symbolic arguments substituted (the #949 machinery the contract verification already uses), then records the erased slot the `MethodImpl` MemberRef is built from. This is *not* the same defect as #3814 proper — it survives the overload-set fix and needed its own repro.

### #3815 — the write path, not the read path

The base's arguments are carried symbolically on `StructSymbol.ImportedBaseType`; its reflected `ClrType` is the erased `Channel<object>`. The inherited-member **read** path already projected correctly (`GetInheritedClrMemberType`, #1582); the inherited-member **write** paths (#319/#1582) reflected the member off the erased type and read its type straight from metadata with `receiverType: null`, so `Reader` bound as `ChannelReader[object]`:

```
Program.gs(7,18): error GS0155: Cannot convert 'ChannelReader[T]' to 'ChannelReader[object]'
```

`MemberLookup.GetProjectionReceiverImportedType` walks a user class to its nearest imported base, and the four inherited-write call sites now pass the receiver — routing them through the same symbolic projection the imported-receiver paths already use. The two paths can no longer drift apart.

## Order of work, as #3816 prescribed

1. gsc first (both halves above);
2. **then** cs2gs: the `!ContainingType.IsGenericType` carve-out in `CSharpToGSharpTranslator.Members.cs` is deleted, so an explicit implementation of an imported generic interface keeps its ADR-0149 clause instead of collapsing into two same-signature overloads (GS0264). Doing this first would have made the diagnostics worse, exactly as #3816 warned.

Methods now behave like the event and property paths, which never had this carve-out.

## Test evidence

`test/Compiler.Tests/Issue3814…` + `Issue3815…` — **18 cases, every one compiles + ILVerifies + runs + asserts the program's own stdout.** No binding-only assertions. The #3814 cases dispatch **through each interface** and the method reports which slot ran, because wiring both interface slots to one method is still verifiable IL; the #3815 cases round-trip a value through the **base-typed** reference.

Failing-on-`origin/main` proof (compiler sources reverted, tests kept):

```
Failed!  - Failed: 7, Passed: 8, Total: 15
```

(15 at that point; three cases were added afterwards for the symbolic-clause sub-defect and the inherited-interface shape.) With the fixes:

```
Passed!  - Failed: 0, Passed: 18, Total: 18
```

The cases that pass on **both** sides are the deliberate anti-vacuity guard rails, labelled as such in the source:

- a **single** explicit clause on an imported generic interface already worked and must keep working;
- a real **override** must still hide its base method (that is what the narrowed hiding rule still does);
- plain duplicate overloads with no clause are still **GS0264**;
- two clauses on the **same** instantiation are still duplicates;
- an instantiation left unimplemented is still **GS0187**;
- a clause naming a member the interface does not declare is still **GS0494**;
- the **non-generic** derived class (#3813) still works;
- assigning a `ChannelReader[string]` to the inherited `Reader` inside `Wrap[T]` is still **rejected** — the unsound direction this could have been "fixed" into, given both sides used to be `object`;
- `protected` stays unreachable from a non-derived type.

cs2gs: `Golden` (47), `ExplicitInterface` + `Oahu` (154) all pass. `OahuDecryptTranslationTests.GenericImportedExplicitInterfaceQualifier_FallsBackAndCompiles` **asserted the old fallback** and is rewritten as `…_KeepsClauseAndCompiles`, plus a new case for the two-instantiation shape that walled `test/Core.Tests`.

`build/nullable_hygiene.py`: OK.

## Premise corrections (posted to both issues)

- **#3814's root cause is one layer below where the issue puts it.** It is not that interface verification fails to match a clause against a constructed imported interface — verification never sees the method at all, because the overload set dropped it. The order-dependence (always the *second* declaration) is the evidence, and it is not in the issue.
- **#3814 is two defects, not one.** The GS0494 symbolic-argument case is real, survives the overload-set fix, and needed a separate fix. The issue folds them together.
- **#3815's `[object]` is a WRITE-path defect only.** The issue says "the same `[object]` shows up in the read path's diagnostics"; the read path in fact already projects correctly, and the `[object]` in those diagnostics is the assignment target, not the source.

## Deconfliction

Touches the gsc binder (imported-generic substitution) and **one** cs2gs file — `CSharpToGSharpTranslator.Members.cs`, the carve-out #3814 explicitly ordered second — plus its test. No overlap with the `test/Interpreter.Tests` triage. `tools/cs2gs/selfmig-baseline.json` untouched.

Fixes #3814
Fixes #3815

### DIFF
--- a/src/Core/CodeAnalysis/Binding/DeclarationBinder.InterfaceVerification.cs
+++ b/src/Core/CodeAnalysis/Binding/DeclarationBinder.InterfaceVerification.cs
@@ -201,6 +201,48 @@ internal sealed partial class DeclarationBinder
                             break;
                         }
                     }
+
+                    if (method.ExplicitInterfaceSlot != null)
+                    {
+                        continue;
+                    }
+
+                    // Issue #3814: a clause naming an imported generic interface
+                    // closed over a SYMBOLIC argument (`func (IEquatable[T])
+                    // Equals(other T) bool` inside `class Value[T]`) never
+                    // matches above: `target.ClrType` is the type-ERASED
+                    // `IEquatable<object>`, so the slot reads `Equals(object)`
+                    // while the declaration says `Equals(T)` (GS0494). Match
+                    // against the OPEN definition with the interface's real
+                    // symbolic arguments substituted — the same #949 machinery
+                    // the interface-contract verification already uses — then
+                    // record the erased slot, which is what the MethodImpl row's
+                    // MemberRef is built from (its containing type carries the
+                    // symbolic arguments).
+                    if (!MemberLookup.TryGetSymbolicClrGenericInterface(target, out var openInterface, out var symbolicArgs))
+                    {
+                        continue;
+                    }
+
+                    foreach (var openSlot in openInterface.GetMethods(BindingFlags.Public | BindingFlags.Instance))
+                    {
+                        if (openSlot.IsSpecialName
+                            || openSlot.Name != method.Name
+                            || !MemberLookup.MethodMatchesSymbolicClrInterfaceSignature(method, openSlot, symbolicArgs))
+                        {
+                            continue;
+                        }
+
+                        var erasedSlot = FindErasedSlotForOpenMethod(target.ClrType, openSlot);
+                        if (erasedSlot == null)
+                        {
+                            continue;
+                        }
+
+                        method.ExplicitInterfaceSlot = erasedSlot;
+                        method.ExplicitInterfaceSlotContainingType = target;
+                        break;
+                    }
                 }
 
                 foreach (var property in structSymbol.Properties)
@@ -264,6 +306,28 @@ internal sealed partial class DeclarationBinder
             VerifyPrivateInterfaceHelpersNotOverridden(syntax, structSymbol);
             VerifyExplicitInterfaceClauseResolution(syntax, structSymbol);
         }
+    }
+
+    /// <summary>
+    /// Issue #3814: maps a method of a CLR generic interface's OPEN definition
+    /// onto the corresponding method of the ERASED closed shape the rest of the
+    /// pipeline reflects against, by metadata identity.
+    /// </summary>
+    /// <param name="erasedInterface">The erased closed interface type.</param>
+    /// <param name="openSlot">The method on the open definition.</param>
+    /// <returns>The matching method on <paramref name="erasedInterface"/>, or <see langword="null"/>.</returns>
+    private static MethodInfo? FindErasedSlotForOpenMethod(Type erasedInterface, MethodInfo openSlot)
+    {
+        foreach (var candidate in erasedInterface.GetMethods(BindingFlags.Public | BindingFlags.Instance))
+        {
+            if (candidate.MetadataToken == openSlot.MetadataToken
+                && candidate.Module == openSlot.Module)
+            {
+                return candidate;
+            }
+        }
+
+        return null;
     }
 
     private static bool ExplicitClrPropertyMatches(
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Assignments.cs
@@ -573,7 +573,11 @@ internal sealed partial class ExpressionBinder
                 inhMember ??= ClrTypeUtilities.SafeGetInheritedInstanceField(inheritedBaseClr, propertyName);
                 if (inhMember != null)
                 {
-                    if (!TryGetWritableClrMember(inhMember, out _, out var inhTargetSymbol, out _, fromDerivedType: true))
+                    // Issue #3815: project the inherited member through the
+                    // derived receiver so an imported GENERIC base's member keeps
+                    // the derived type's own type arguments (ChannelReader[T], not
+                    // the erased ChannelReader[object]).
+                    if (!TryGetWritableClrMember(inhMember, structSymbol, out _, out var inhTargetSymbol, out _, fromDerivedType: true))
                     {
                         Diagnostics.ReportCannotAssign(initSyntax.EqualsToken.Location, propertyName);
                         _ = BindExpression(initSyntax.Value);
@@ -1079,7 +1083,11 @@ internal sealed partial class ExpressionBinder
                 clrMember ??= ClrTypeUtilities.SafeGetInheritedInstanceField(inheritedBaseClr, memberName);
                 if (clrMember != null)
                 {
-                    if (!TryGetWritableClrMember(clrMember, out var inhTargetType, out var inhTargetSymbol, out var inhWritable, fromDerivedType: true))
+                    // Issue #3815: project the inherited member through the
+                    // derived receiver so an imported GENERIC base's member keeps
+                    // the derived type's own type arguments (ChannelReader[T], not
+                    // the erased ChannelReader[object]).
+                    if (!TryGetWritableClrMember(clrMember, structSymbol, out var inhTargetType, out var inhTargetSymbol, out var inhWritable, fromDerivedType: true))
                     {
                         Diagnostics.ReportCannotAssign(syntax.EqualsToken.Location, memberName);
                         return new BoundErrorExpression(null);
@@ -2706,7 +2714,11 @@ internal sealed partial class ExpressionBinder
                 clrMember ??= ClrTypeUtilities.SafeGetInheritedInstanceField(inheritedBaseClr, fieldName);
                 if (clrMember != null)
                 {
-                    if (!TryGetWritableClrMember(clrMember, out var inhTargetType, out var inhTargetSymbol, out var inhWritable, fromDerivedType: true))
+                    // Issue #3815: project the inherited member through the
+                    // derived receiver so an imported GENERIC base's member keeps
+                    // the derived type's own type arguments (ChannelReader[T], not
+                    // the erased ChannelReader[object]).
+                    if (!TryGetWritableClrMember(clrMember, structSym, out var inhTargetType, out var inhTargetSymbol, out var inhWritable, fromDerivedType: true))
                     {
                         Diagnostics.ReportCannotAssign(syntax.EqualsToken.Location, fieldName);
                         return new BoundErrorExpression(null);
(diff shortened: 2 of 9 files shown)


DavidObando:
## App-level proof (migrated run, this branch)

Whole-repository `migrate --translate-only` (52/52 apps translate) + a 2-app `validate` shard over that tree, with the SDK nupkg repacked from this branch's commit (`gsc: 0.4.385+75d05d0ee7`, i.e. a real commit each side of the moniker stamp):

```
app                                       compile       ilverify      test-parity
----------------------------------------------------------------------------------
test/Core.Tests/Core.Tests.csproj         FAIL(3)       skip          skip
bench/concurrency/clr/ClrBaseline.csproj  PASS          PASS          PASS
```

| app | before (gate 33584377318) | after |
|---|---|---|
| `bench/concurrency/clr/ClrBaseline` | compile FAIL(2) — GS0113, GS0127; **#3815 behind them** | **GREEN** — compile + ILVerify + test-parity all PASS |
| `test/Core.Tests` | compile FAIL(2) — **GS0187, GS0264** | **both gone**; FAIL(3), all three a different, unrelated defect |

**ClrBaseline goes green**, which is what #3815 was standing in front of — #3812/#3813 landed in #3816 and this PR removes the last one.

**`test/Core.Tests`'s GS0187 + GS0264 are gone**, which is what #3814 asked for. The three errors now reported are `GSA0005` **from the dependency project `src/Core`**, not from `Core.Tests`, and they are a different defect entirely: `cs2gs` has no `#pragma warning disable` handling at all (zero hits for `PragmaWarning` in the translator), so `CaptureBoxingRewriter.cs`'s four deliberate GSA0005 suppressions survive into the `.gs` as comments only. Filed as #3824 with the before/after and the repro. The drop is not new — the translator has never emitted a pragma — what is new is that migrated projects now run the ADR-0169 analyzers, so a dropped suppression became fatal. Nothing in this PR touches pragmas, comments, or trivia.

DavidObando:
## CI

**36 of 37 checks pass.** The one red check, `tests (cs2gs-remainder)`, is **#3818, and it is red on `main` itself** — not a regression from this PR:

- main run [33636256718](https://github.com/DavidObando/gsharp/actions/runs/33636256718) — same job, same assertion
- main run [33633833075](https://github.com/DavidObando/gsharp/actions/runs/33633833075) — same

```
Failed Cs2Gs.Tests.Adr0169AnalyzerTranslationTests.SemanticModelAnalyzer_TranslatesDeclarationAndOverrideIdioms
Cannot find function FirstAncestorOrSelf.
```

I re-ran the failed job on this PR per the standing guidance and it failed identically, so it is no longer a flake; posted that correction to #3818. With a narrow filter it passes locally (24/24), consistent with #3818's assembly-resolution-isolation theory.

Everything else is green, including every `compiler-*` shard and `core-binding` — the shards most exposed to the overload-set and member-projection changes here.

DavidObando:
Correction to the tally in the previous comment: **34 pass, 1 skip (`Deploy to GitHub Pages`), 1 fail**. The one failure is #3818 as described.

Issue #3822, opened 2026-09-02
ADR-0169: the translated RewriterClonePreservationAnalyzer does not follow reads through a one-level node helper, so it false-positives on MoveNextBodyRewriter (GSA0005)


DavidObando:
Split out of #3820 as sub-cause B.

Of the 15 `GSA0005` sites the migrated `InternalAnalyzers` reports against migrated `src/Core`, fourteen correspond to a `#pragma warning disable GSA0005` in the C# source that cs2gs drops (#3820). **One does not**:

```
migrated/src/Core/CodeAnalysis/Lowering/Async/MoveNextBodyRewriter.gs(289,42,289,74): error GSA0005:
'InnerRewriter.RewriteFieldAssignmentExpression' rebuilds BoundFieldAssignmentExpression but never
reads node.StructType, which BoundTreeRewriter.RewriteFieldAssignmentExpression preserves
```

`src/Core/CodeAnalysis/Lowering/Async/MoveNextBodyRewriter.cs` carries no suppression for that method, and the C# build is clean — so the C# analyzer does **not** fire there. It reads `StructType` indirectly, via `BoundNodeForm.DeclaringType(node)`, and the analyzer's own remarks say so explicitly:

> Reads reached through a helper that takes the node are followed one level, so extracting a recurring union invariant into a helper (as `Lowering/BoundNodeForm.cs` does) does not trip it.

The **translated** analyzer does not follow that hop, so it reports a member the code does read. That is a fidelity gap between the C# analyzer and its G# translation, in the same family as #3795/#3809 (`Symbol.ContainingType` was null, so translated syntax-node analyzers reported *nothing*), #3796 and #3797.

Worth checking whether the one-level helper follow is a semantic-model capability that is simply missing in G# (symbol → declaring syntax → body), rather than something specific to this analyzer: if so, every translated analyzer that follows a helper has the same hole, and this is the second such hole #3809 has now made visible.

Found while triaging the `test/Interpreter.Tests` compile wall (#3820); measured on `origin/main` @ `89ff8246c`, gsc `0.4.384+89ff8246c4`.

DavidObando:
## This cause is 5 sites, not 1 — measured after the #3824 suppression fix

[...]

Issue #3820, opened 2026-09-02
cs2gs/G#: #pragma warning disable is dropped in translation, so the now-effective migrated analyzers turn src/Core red (GSA0005 x15; corpus-wide selfmig wall, found via test/Interpreter.Tests)


DavidObando:
## Summary

`#pragma warning disable <ID>` / `#pragma warning restore <ID>` is **dropped entirely** by cs2gs — there is no handling of `PragmaWarningDirectiveTrivia` anywhere in `tools/cs2gs/Cs2Gs.Translator`, and the migrated tree contains zero `pragma` text. Until now that was invisible, because the translated ADR-0169 analyzers reported nothing.

**#3809 (`Symbol.ContainingType` on the SemanticModel, landed 2026-09-02) made them report.** The first thing they report is the 15 sites in `src/Core` whose C# source suppresses `GSA0005` with exactly those pragmas.

Result: **migrated `src/Core` no longer compiles.** `src/Core/Core.csproj` is a *banked* entry in `selfmig-baseline.json`'s `greenApps`, so this is a ratchet regression, and because every other app in the corpus project-references Core, it is a corpus-wide wall — not a `test/Interpreter.Tests` problem.

## Evidence (measured on `origin/main` @ `89ff8246c`, gsc `0.4.384+89ff8246c4`)

Whole-repo `build/run-cs2gs-selfmig-migrate.sh`: **52/52 apps translate** (confirms #3817).
Then `build/run-cs2gs-selfmig-validate.sh interp test/Interpreter.Tests/Interpreter.Tests.csproj`:

```
self-migration shard interp: 0/1 apps green.
failureCategory: compile-error, compile stage: 3 artifacts
```

`sdk.build.log`: **30 error lines / 15 distinct diagnostics, 100% `GSA0005`, 100% in `migrated/src/Core/…`.** Zero `GS0157`, zero `GS0154`. `test/Interpreter.Tests`' own sources are never compiled — the dependency dies first.

The 15 sites:

| file | sites | C# has `#pragma warning disable GSA0005`? |
|---|---|---|
| `src/Core/CodeAnalysis/Lowering/CaptureBoxingRewriter.gs` | 4 | yes (4 regions) |
| `src/Core/CodeAnalysis/Lowering/Iterators/HoistedFieldRewriter.gs` | 4 | yes (2 regions) |
[...]

Issue #3805, opened 2026-09-02
cs2gs: linked test/Shared/EmittedOracle.cs translates differently in Interpreter.Tests — a type name is fully qualified there and short elsewhere (second translate wall behind #3804)


DavidObando:
## What this is

The **second** translate wall in `test/Interpreter.Tests`, revealed once #3804's `IndexOutOfRangeException` crash was fixed. The app still fails the translate stage in a whole-repository run; the corpus is 51/52 translated, not 52/52.

This was completely hidden before: the crash aborted the app's document loop before the linked-source cross-check ever ran.

## Symptom

```
System.InvalidOperationException: Linked source '<repo>/test/Shared/EmittedOracle.cs'
translates differently in multiple projects. First divergence at line 129.
  prior (as translated by the first project to reach it):
              // Mirror Compilation.Evaluate's compile-failure shape: warnings first,
              // then errors, value null, nothing executed.
  >           let warnings = emitResult.Diagnostics.Where((d Diagnostic) -> !d.IsError).ToImmutableArray()
              let errors = emitResult.Diagnostics.Where((d Diagnostic) -> d.IsError).ToImmutableArray()
              if !emitResult.Success || errors.Length > 0 {
  current (as translated by this app):
              // Mirror Compilation.Evaluate's compile-failure shape: warnings first,
              // then errors, value null, nothing executed.
  >           let warnings = emitResult
                  .Diagnostics
                  .Where((d GSharp.Core.CodeAnalysis.Diagnostic) -> !d.IsError)

   at Cs2Gs.Pipeline.TranslateStage.RegisterRepositoryTranslation(...)
   at Cs2Gs.Pipeline.TranslateStage.ExecuteAsync(...)
```

`test/Shared/EmittedOracle.cs` is linked into `Compiler.Tests`, `Extensions.Tests` and `Interpreter.Tests`. `Compiler.Tests` (first to reach it) and `Extensions.Tests` agree; `Interpreter.Tests` differs.

## The actual difference

The lambda parameter's type is spelled **`Diagnostic`** in `Compiler.Tests` and **`GSharp.Core.CodeAnalysis.Diagnostic`** in `Interpreter.Tests`. The line wrapping is a consequence, not the cause: the fully-qualified spelling pushes the chain over the printer's width and it breaks across lines.

[...]

Issue #3749, opened 2026-09-01
gsc: conversion call on a nested imported type fails to bind (GS0159, from #3745)


DavidObando:
Split out of #3745 (family F5). Found in the migrated `test/Compiler.Tests` compile wall; 1 of that app's 20 errors.

The G# conversion-call form `T(expr)` does not resolve a **nested** imported type, so a C# cast to a nested enum translates to something gsc cannot bind.

## Repro (gsc, `0.4.349+cec369958b`)

```gs
import System
import System.Diagnostics

func main() {
    let flags = DebuggableAttribute.DebuggingModes(2)
    // error GS0159: Cannot find function DebuggingModes.
    Console.WriteLine(flags)
}
```

`DebuggableAttribute.DebuggingModes` is a nested enum. The unnested spelling of a conversion call (e.g. `int32(x)`, `SomeTopLevelEnum(x)`) binds today.

## Current evidence in the corpus

`test/Compiler.Tests/Issue2920ReleaseJitOptimizerTests.cs:78`:

```csharp
Assert.Equal((DebuggableAttribute.DebuggingModes)expectedFlags, debuggable!.DebuggingFlags);
```

cs2gs faithfully translates the cast to the conversion-call form, migrated to `Issue2920ReleaseJitOptimizerTests.gs:49,46`:

```gs
Assert.Equal(DebuggableAttribute.DebuggingModes(expectedFlags), debuggable!!.DebuggingFlags)
```

which reports `GS0159 Cannot find function DebuggingModes.`

The translation looks correct, so this reads as a gsc binder gap (a member-access expression in conversion-call position is only probed as a function, never as a nested type). Worth confirming the layer before fixing: if the nested conversion-call form is deliberately unsupported by the grammar, cs2gs should emit `cast[…]` instead.

## Not a regression

Adjacent to but distinct from the solved nested-type family (#3660 / PR #3662), which was nested-type → enclosing-`shared` **calls**; this is the conversion-call direction.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Nng28yiBdPVdeML7mSZphs

DavidObando:
**Scope correction from PR #3758.** Nesting makes this gap *visible* rather than causing it.

Nested-type resolution (#569) already worked — the type resolves fine. The namespace-qualified form fails on `main` too:

```
System.DayOfWeek(2)
```

So the gap belongs to the **dotted binding path**, not to nested types specifically.

[...]

PR #3823, opened 2026-09-02
gsc: a fully-qualified constructor call with no applicable overload reports the overload error, not GS0157 (#3821)

### LINKED ISSUE #2621: gsc: generic List capacity construction can be bound as a missing function
Cycle 5: List[T](capacity) is treated as a nonexistent function and reports GS0130. Definition of done: bind imported generic type construction with capacity overloads and preserve overload diagnostics; reproduce exact Oahu.Cli site and add runtime coverage.

### LINKED ISSUE #3821: gsc: a fully-qualified constructor call with no applicable overload reports GS0157 'Cannot find type <first segment>' instead of the overload error
## Summary

A **fully-qualified constructor call** whose arguments match no constructor reports

```
error GS0157: Cannot find type System. Are you missing an import?
```

— a diagnostic about the *leading namespace segment*, which resolves perfectly — instead of the overload error the identical call reports when the type name is written unqualified.

## Repro (gsc `0.4.384+89ff8246c4`, `origin/main` @ `89ff8246c`)

```gs
package probe

func systemWrongArity() System.Text.StringBuilder {
    return System.Text.StringBuilder("a", "b", "c", "d", "e")
}
```

```
p10.gs(4,12,4,18): error GS0157: Cannot find type System. Are you missing an import?
```

Same shape on any root: `GSharp.Core.CodeAnalysis.DiagnosticDescriptor("GS0001")` reports `Cannot find type GSharp`. The unqualified spelling of the very same call (`import System.Text` + `StringBuilder("a","b","c","d","e")`) reports the overload error correctly, so this is a pure spelling asymmetry.

Everything else about the qualified spelling is fine: the qualified type in a type clause binds, as a generic argument, as a lambda parameter type, as a static-member receiver, and a qualified constructor call that *does* resolve binds. Only the failure path misreports.

## Cause

`ExpressionBinder.TryBindQualifiedClrConstructorCall` resolves the dotted name to a CLR type and hands off to the shared constructor core. On no-applicable-overload the shared helper `FinishClrConstructorBindingFailure` only commits to `GS0267`:

```csharp
if (syntax.TypeArgumentList == null)
{
    result = null;
    return false;      // <- non-generic call: give up silently
}
…
Diagnostics.ReportNoApplicableOverload(syntax.Identifier.Location, typeName);
```

That guard came from #2633 (fixes #2621), which cured exactly this misleading cascade — but only for the **generic** spelling (`List[T](bad)`). The non-generic qualified call answers "not handled", `BindAccessorExpression` falls through, the whole chain re-binds as a namespace walk, and the walk fails at its **head** — which is where GS0157 is reported.

## Why it matters beyond diagnostic polish

cs2gs writes fully-qualified type names whenever a simple name is ambiguous (#2258, and now #3805/#3817 for linked sources). On the self-migration corpus this collapses every genuine argument error at such a call site into the same one misleading line, all pointing at a namespace root. #3817's triage of `test/Interpreter.Tests` recorded *"GS0157 Cannot find type GSharp ×21"* as the leading family — 21 errors whose real causes the diagnostic hid.

## Fix

Commit to the overload diagnostic in `TryBindQualifiedClrConstructorCall` itself, gated so the two later arms of `BindAccessorExpression` that can still bind such a chain keep their second chance:

* the prefix must be a **namespace**, not a type (otherwise the fully-qualified static-access arm may still find a member of it);
* the terminal simple name must not also name a **source** type (a package-qualified source construction such as `Oahu.Tui.Nested.Refresh(1)` from #2585 reaches the CLR arm first, can resolve a same-named *reference* type whose constructors do not match, and is bound by the source arm below).

PR attached.

DavidObando:
Fixes #3821.

A **fully-qualified constructor call** whose arguments match no constructor reported

```
error GS0157: Cannot find type System. Are you missing an import?
```

— about the *leading namespace segment*, which resolves perfectly — while the unqualified spelling of the same call reported the overload error correctly.

## Cause

`ExpressionBinder.TryBindQualifiedClrConstructorCall` resolves the dotted name to a CLR type and hands off to the shared constructor core. On no-applicable-overload, `FinishClrConstructorBindingFailure` only commits to `GS0267`:

```csharp
if (syntax.TypeArgumentList == null)
{
    result = null;
    return false;      // non-generic call: give up silently
}
```

That guard came from #2633 (fixes #2621), which cured exactly this misleading cascade — **but only for the generic spelling** (`List[T](bad)`). The non-generic qualified call answered "not handled", `BindAccessorExpression` fell through, the chain re-bound as a namespace walk, and the walk failed at its head, which is where GS0157 is reported.

## The change

Commit to the overload diagnostic in `TryBindQualifiedClrConstructorCall` itself, gated twice so the two later arms of `BindAccessorExpression` that can still bind such a chain keep their second chance:

* the prefix must be a **namespace**, not a type — otherwise the fully-qualified static-access arm may still find a member of it (`DebuggableAttribute.DebuggingModes(2)`, #3749);
* the terminal simple name must not also name a **source** type — a package-qualified source construction (`Oahu.Tui.Nested.Refresh(1)`, #2585) reaches the CLR arm first, can resolve a same-named *reference* type whose constructors do not match, and is bound by the source arm below.

The second gate is not speculative: without it, `Issue2585SemanticQualificationTests.SourceQualifiedTypeAndValueReceiver_RemainDistinct` fails.

## Why it matters

cs2gs writes fully-qualified type names whenever a simple name is ambiguous (#2258, and #3805/#3817 for linked sources), so on the self-migration corpus every genuine argument error at such a call site collapsed into the same one misleading line pointing at a namespace root. #3817's triage of `test/Interpreter.Tests` recorded *"GS0157 Cannot find type GSharp ×21"* as the leading family — 21 errors whose real causes the diagnostic hid.

This does **not** clear that app's wall. Re-measuring it on `origin/main` @ `89ff8246c` shows the wall is now 15 `GSA0005` errors in migrated `src/Core` (filed as #3820, with #3822 split out); the GS0157 layer sits behind them.

## Tests

`test/Core.Tests/CodeAnalysis/Binding/Issue3821QualifiedConstructorOverloadDiagnosticTests.cs`, executing through `EmittedOracle.Evaluate` (compile + run), honestly labelled:

| test | on `origin/main` | here |
|---|---|---|
| `QualifiedConstructorCall_NoApplicableOverload_ReportsOverloadErrorNotMissingType` | **fails** (`Collection: [Cannot find type System. …]`) | passes |
| `QualifiedConstructorCall_WrongArgumentType_ReportsOverloadErrorNotMissingType` | **fails** (same) | passes |
| `QualifiedConstructorCall_ThatBinds_StillBinds` | passes | passes (anti-vacuity) |
| `QualifiedStaticCall_OnTypeReceiver_StillBinds` | passes | passes (anti-vacuity — pins the type-prefix gate) |
| `UnqualifiedConstructorCall_NoApplicableOverload_AlreadyReportsOverloadError` | passes | passes (anti-vacuity — pins the asymmetry being removed) |

Measured, not asserted: 2 failed / 3 passed on `origin/main`; 5 passed here.

Full `test/Core.Tests`: **8578 passed, 0 failed** (the intermediate version of the gate, without the source-type condition, failed 1 — `Issue2585SemanticQualificationTests`).

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.cs
@@ -272,6 +272,45 @@ internal sealed partial class ExpressionBinder
                 boundArguments,
                 ref result,
                 TypeSymbol.FromClrType(clrType));
+
+        // Issue #3821: the shared failure helper only commits to the
+        // no-applicable-overload diagnostic when the call carries an explicit
+        // type-argument list (issue #2633 fixed this cascade for the GENERIC
+        // spelling only). Without that, a non-generic qualified construction
+        // whose arguments match no constructor answered "not handled", the
+        // whole accessor chain re-bound as a namespace walk, and the walk
+        // failed at its HEAD — reporting GS0157 "Cannot find type System"
+        // about a segment that resolves perfectly. cs2gs writes fully-qualified
+        // type names whenever a simple name is ambiguous (issues #2258/#3805),
+        // so on the self-migration corpus every argument error at such a call
+        // site collapsed into that one misleading line.
+        //
+        // Committing is gated twice, because two later arms of
+        // `BindAccessorExpression` can still bind a chain this one gave up on:
+        //
+        //   * the prefix must be a NAMESPACE, not a type. When the last prefix
+        //     segment is itself a type (`Outer.Nested(...)`,
+        //     `DebuggableAttribute.DebuggingModes(2)`), the fully-qualified
+        //     static-access arm can still find a member of it.
+        //   * the terminal simple name must not also name a SOURCE type. A
+        //     package-qualified source construction (`Oahu.Tui.Nested.Refresh(1)`,
+        //     issue #2585) reaches the CLR arm first and can resolve a
+        //     same-named REFERENCE type whose constructors do not match — the
+        //     source arm below is the one that binds it.
+        //
+        // A namespace prefix with no source homonym has neither second chance,
+        // so there the fallthrough can only ever produce GS0157.
+        if (!handled
+            && noApplicableOverload
+            && namespacePrefix.Length > 0
+            && !scope.References.TryResolveType(namespacePrefix, out _)
+            && !TryLookupSourceTypeWithImportPrecedence(typeSimpleName, preferredArity: -1, declaration: null, out _))
+        {
+            Diagnostics.ReportNoApplicableOverload(terminalCall.Identifier.Location, typeSimpleName);
+            result = new BoundErrorExpression(terminalCall);
+            handled = true;
+        }
+
         if (handled && terminalObjectCreation != null)
         {
             result = BindObjectInitializerSuffix(
--- a/test/Core.Tests/CodeAnalysis/Binding/Issue3821QualifiedConstructorOverloadDiagnosticTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Binding/Issue3821QualifiedConstructorOverloadDiagnosticTests.cs
@@ -1,0 +1,129 @@
+// <copyright file="Issue3821QualifiedConstructorOverloadDiagnosticTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System.Collections.Immutable;
+using GSharp.Core.CodeAnalysis;
+using GSharp.Tests;
+using Xunit;
+
+namespace GSharp.Core.Tests.CodeAnalysis.Binding;
+
+/// <summary>
+/// Issue #3821: a FULLY-QUALIFIED constructor call whose arguments match no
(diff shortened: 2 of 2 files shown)


DavidObando:
CI green on re-run. Two first-pass failures, both confirmed not to be this change:

- `tests (compiler-issue20-24)` — `Failed to FinalizeArtifact: … (403) Forbidden` in the *Upload test results* step, i.e. an artifact-upload infrastructure failure, not a test failure. Reproduced that exact shard locally on this branch: **492 passed, 0 failed**. Green on re-run.
- `tests (cs2gs-remainder)` — `Adr0169AnalyzerTranslationTests.SemanticModelAnalyzer_TranslatesDeclarationAndOverrideIdioms`, `Cannot find function FirstAncestorOrSelf`: the known assembly-resolution-isolation flake, #3818, matched signature-for-signature. Green on re-run.

Issue #3780, opened 2026-09-01
ADR-0169: publish GSharp.CodeAnalysis.Analyzers.Testing as a package instead of co-locating it beside gsc (from #3686)


DavidObando:
Split out of **#3686** / PR #3777, and flagged there for review rather than decided unilaterally.

## What #3777 did

To make migrated analyzer tests resolve the G# verifier on **both** the SDK and direct-compile paths, PR #3777 ships `GSharp.CodeAnalysis.Analyzers.Testing.dll` **beside `gsc`** — and inside the SDK nupkg under `tools/compiler`. There is a comment in `src/Compiler/Compiler.csproj` saying so.

## Why that is interim, not the answer

ADR-0169 describes a **packable testing library**. Co-location was taken only because that package does not exist yet. The co-located form has real costs:

- it puts a test-only assembly in the compiler's own directory and in the SDK payload, where nothing but migrated tests needs it;
- consumers outside the migration have no supported way to reference the verifier;
- the resolution works by directory adjacency rather than by a declared dependency, which is the kind of coupling that breaks silently when either layout moves.

## The work

Publish `GSharp.CodeAnalysis.Analyzers.Testing` as a proper package, have migrated analyzer test projects take a PackageReference to it, and remove the co-location plus its csproj comment. Worth confirming at that point whether the SDK nupkg still needs to carry it at all.

## Note for whoever picks this up

The migrated test projects are generated by cs2gs's project transform (`Cs2Gs.Pipeline/GSharpProjectTransformer.cs`), so "take a PackageReference" is a change to the **transform**, not to a checked-in csproj. #3777 already injects `GSharp.Core` + the verifier there and copies them; that injection is the seam to change.

Refs #3686, #3778, #3501.

Issue #3235, opened 2026-08-04
In-process emitted assemblies leak into later ReferenceResolver.Default() scans, shadowing same-named source packages


DavidObando:
## Root cause (revised)

`ReferenceResolver.BuildHostAssemblies()` rescans `AppDomain.CurrentDomain.GetAssemblies()` on every `ReferenceResolver.Default()` call, excluding only `DriverReferenceLoadContext` assemblies. The collectible-ALC assemblies produced by in-process emitted execution — `EmittedProgramHost` (`gsharp-emitted-*`), `InteractiveSessionHost` (`gsi-session-*`), and the test-side `EmittedOracle` — stay visible in that scan for as long as they are alive. Any compilation subsequently bound with a default resolver then sees those emitted assemblies as host references, and their types **shadow same-named source packages** in later, unrelated compilations.

## Repro (verified)

1. Emit + run in-process a program declaring `package App.Nested / class Holder { const Message }` (hold the result so the collectible ALC stays alive — exactly what a REPL session or a test does).
2. Compile a fresh, unrelated pair of trees: `package App.Nested / class Holder { prop Value; shared func MakeOne() }` + `package App / import R = App.Nested.Holder / var h R = R.MakeOne()`.

Step 2 binds `R` against the STEP-1 assembly's `App.Nested.Holder` and fails with `Cannot find function MakeOne.` / `Cannot find member Value.` — even though the compilation's own trees declare the right shape. First observed as what looked like multi-tree submission binding gaps in `Issue2273ImportTypeAliasTests` and tree-order sensitivity in `Issue2519SourceClassConstraintOrderingTests`; both were this leakage (the cross-tree import alias binds and runs correctly through a clean resolver — covered by `EmittedOracleTests.MultiSource_CrossTreeImportAlias_ResolvesAndRuns`).

## Expected

Assemblies produced by in-process emitted execution are execution artifacts, not references: `BuildHostAssemblies` should exclude them (e.g. tag their load contexts the way `DriverReferenceLoadContext` is already excluded), so a long-lived driver/REPL process cannot have earlier runs shadow later compilations.

## Scope

[...]

PR #3825, opened 2026-09-02
gsc: a runtime reference to an already-loaded assembly reuses the host's identity (#3818)

### LINKED ISSUE #3818: test: ConversionEmitTests.CompileToAssembly has no reference isolation, so emit shards can flake on test order

DavidObando:
Fixes #3818.

`tests (cs2gs-remainder)` has been failing about half the time on
`Adr0169AnalyzerTranslationTests`, always with one opaque error —
`Cannot find function FirstAncestorOrSelf` — while every other member of the
same translated analyzer bound cleanly.

## Not test isolation — a product defect

`ReferenceResolver.WithRuntimeReferences` loaded every reference path into a
fresh collectible `DriverReferenceLoadContext`, unconditionally. When the path
names an assembly the host has **already** loaded — exactly what the ADR-0169
helpers pass, `typeof(Diagnostic).Assembly.Location` — that mints a second,
rival identity for every type in `GSharp.Core`, once per compilation.

The binder's symbol and member caches are process-wide and keyed on `Type`, so a
`Type` can outlive the compilation that created it. Instrumenting the GS0159
site on a local reproduction showed exactly that: the receiver's `SyntaxNode`
came from a *previous* compilation's driver context (`recvInResolverList=False`,
`recvIsCurrent=False`) while the explicit type argument
`FunctionDeclarationSyntax` came from the current one. Closing
`FirstAncestorOrSelf<T>` over operands from two contexts makes
`MakeGenericMethod` throw; `ClrOverloadResolution` drops the candidate silently
(C# §7.5.2), and the call reports "Cannot find function" about a method that
plainly exists.

This is the #3705 load-context family: two operands for one logical CLR type,
and an answer that depends on which one arrives.

## The fix

A reference whose path the default load context already holds resolves to the
host's instance instead of a private copy. Same file plus same identity means
reuse is always safe; a genuinely different assembly does not match by path and
is still loaded privately, so driver `/reference:` behaviour for user assemblies
is unchanged. The reused reference still contributes its directory to the driver
context's probe set.

## Evidence

Reproduction: `Adr0169Gsa0005ParityTests` + `Adr0169AnalyzerTranslationTests` in
one test host (~40s per run).

| | runs | failures |
|---|---|---|
| before | 43 | 18 (42%) |
| after | 8 | 0 |

The two new `Issue3818RuntimeReferenceIdentityTests` fail on `main` and pass
here.

Refs #3705, #3780, #3777.

### DIFF
--- a/src/Core/CodeAnalysis/Symbols/ReferenceResolver.cs
+++ b/src/Core/CodeAnalysis/Symbols/ReferenceResolver.cs
@@ -209,9 +209,15 @@ public sealed class ReferenceResolver : IDisposable
 
         public Assembly LoadReference(string path)
         {
-            directories.Add(Path.GetDirectoryName(path) ?? string.Empty);
+            AddProbeDirectory(path);
             return LoadFromAssemblyPath(path);
         }
+
+        // Issue #3818: a reference reused from the host context (rather than
+        // loaded here) still contributes its directory, so a private
+        // dependency sitting beside it stays resolvable from this context.
+        public void AddProbeDirectory(string path)
+            => directories.Add(Path.GetDirectoryName(path) ?? string.Empty);
 
         protected override Assembly? Load(AssemblyName assemblyName)
         {
@@ -509,7 +515,27 @@ public sealed class ReferenceResolver : IDisposable
                 }
 
                 var fullPath = Path.GetFullPath(path);
-                var assembly = runtimeContext.LoadReference(fullPath);
+
+                // Issue #3818: a reference the host has ALREADY loaded from
+                // this exact path must resolve to the host's instance, not to
+                // a second private copy. Loading it again into the driver
+                // context mints a rival identity for every type in it, and any
+                // Type that leaks between two compilations in one process
+                // (through the binder's process-wide symbol/member caches) then
+                // becomes a cross-context type: `MakeGenericMethod` rejects it,
+                // the only candidate is dropped, and the call reports the
+                // opaque "Cannot find function <name>" — order-dependently,
+                // because it needs an earlier compilation to have created the
+                // rival copy. Same file plus same identity means reuse is
+                // always safe; a genuinely different assembly does not match by
+                // path and is still loaded privately.
+                Assembly? reused = TryReuseHostLoadedAssembly(fullPath);
+                if (reused is not null)
+                {
+                    runtimeContext.AddProbeDirectory(fullPath);
+                }
+
+                var assembly = reused ?? runtimeContext.LoadReference(fullPath);
                 RegisterAssemblyOriginalPath(assembly, fullPath);
                 runtimeAssemblies.Add(assembly);
             }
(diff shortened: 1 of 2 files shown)


DavidObando:
`tests (cs2gs-remainder)` — the gate this PR is about — **passes** here (11m56s). 32 of 34 jobs green.

`tests (compiler-emit-remainder)` is red, and it is **not** a regression from this PR: it is the
pre-existing `ConversionEmitTests.CompileToAssembly` flake this issue was originally filed about,
with the exact signature #3818 predicted ("a `P.Color` from a previous fixture"):

```
Int32_To_Enum_Cast_Produces_Corresponding_Enum_Value [FAIL]
  gsc failed:
  test.gs(7,12,7,20): error GS0156: Cannot convert type 'P.Color' to 'Color'.
```

The identical failure — same test, same message — is on `main` at `510ff7bef`
([job 100058109517](https://github.com/DavidObando/gsharp/actions/runs/33568823940/job/100058109517)),
before this branch existed.

It also cannot be reached by this change. `CompileToAssembly` finishes with
`Assembly.Load(byte[])`, whose `Assembly.Location` is empty, so the path-equality reuse added here
never matches it. The mechanism is the other half of the same root cause: gsc is invoked
**in-process** with no `/reference:`, each emitted `package P` is loaded into the **default,
non-collectible** context, and `BuildHostAssemblies()` therefore offers every previous fixture's
types to every later compile.

Fixing that is a separate, test-side change (load into a collectible `AssemblyLoadContext`, which
`BuildHostAssemblies` already excludes per #3235) and does not belong in this PR.

Issue #1622, opened 2026-07-02
Process-wide type-symbol caches are never evicted — cross-compilation memory leak that defeats ReferenceResolver.Dispose


DavidObando:
- **Category**: performance
- **Severity**: P2
- **Location**: src/Core/CodeAnalysis/Symbols/ReferenceResolver.cs:152-176 vs the static caches at NullableTypeSymbol.cs:21, SliceTypeSymbol.cs:25, MapTypeSymbol.cs:25, PointerTypeSymbol.cs:25, ByRefTypeSymbol.cs:16, ArrayTypeSymbol.cs:21, SequenceTypeSymbol.cs:16, AsyncSequenceTypeSymbol.cs:21, ChannelTypeSymbol.cs:22, TupleTypeSymbol.cs:25, FunctionPointerTypeSymbol.cs:27, StructSymbol.cs:21/28/36, InterfaceSymbol.cs:21, DelegateTypeSymbol.cs:33
- **Evidence**:
```csharp
// ReferenceResolver.Dispose():
ImportedTypeSymbol.ClearCache();
...
FunctionTypeSymbol.ClearCache();
```
Only 2 of ~15 static `ConcurrentDictionary` caches have a `ClearCache`. E.g. `NullableTypeSymbol` holds the MLC-projected CLR type, `SliceTypeSymbol` holds `GetEffectiveClrType(elementType)?.MakeArrayType()`, and `StructSymbol.ConstructedCache` is keyed by per-compilation `StructSymbol` definitions (which reference `Declaration` syntax trees).
- **Problem**: Dispose's stated goal is that "the disposed context's `Type` objects can be garbage-collected", but Nullable/Slice/Map/Pointer/etc. entries keyed by (now-orphaned) `ImportedTypeSymbol` instances still pin the disposed `MetadataLoadContext` object graph, and the three constructed-generic caches root every generic definition, its constructed instances, and their syntax trees for the process lifetime. In long-lived hosts memory grows without bound; stale entries are unreachable via lookups so they are pure leak. (Overlaps with perf-sweep static-cache finding — same root cause; merge.)
- **Suggested fix**: Add `ClearCache()` to the remaining caches and call them from `ReferenceResolver.Dispose` (or move interning into a per-compilation cache object owned by the Compilation).
- **Confidence**: high

---
*Found by an automated multi-agent code review with Claude Code. `file:line` references are relative to commit `1e7e5a91`.*

Issue #420, opened 2026-06-04
[P3] CIL emit: latent / fragile emit and metadata code paths (technical debt)


DavidObando:
## Summary

Tracking issue for **P3** findings from the CIL emit code review (4-agent sweep on `main` @ `a3a7f0b`). P3 = latent / fragile / cosmetic. None of these is a current user-visible bug, but each is a code-smell or fragile invariant that risks turning into a P0/P1 the moment a related feature lands. Flagged here so the next person touching the area is aware.

Companion issues:
- P0: `InvalidProgramException` / exception swallowing
- P1: side-effect duplication, walker blind spots, etc.
- P2: pattern emit / conversion / PDB quality

**Baseline.** Clean build, 0 warnings, 1812 tests passing.

---

### P3-1. CLR method-group on a struct receiver assumes the receiver can be used as an `object` reference

`ReflectionMetadataEmitter.cs` (around the indirect-call emit). Today gated by the binder, but no defensive `box` exists in the emitter. If a future binder change allows method-group conversions where the source is a struct, emit will produce unverifiable IL.

### P3-2. Type pattern against `Nullable<T>` would mis-behave due to special CLR boxing rules

Pattern emit assumes `isinst` on the boxed value. CLR has special handling that boxes `Nullable<T>` as either `null` or the value `T`; pattern `case T` would match the latter unexpectedly. Not reachable today because the binder forbids the shape; document the assumption so a future binder change is forced to revisit emit.

### P3-3. Pre-allocated slot dictionaries (`structLiteralSlots`, `receiverSpillSlots`, `defaultExpressionSlots`, `mapIndexSlots`) are one-shot, keyed by node identity

Sharing a node across two emit positions would alias the slot. Defensive only — no current binder/lowering path produces shared nodes. Add an assertion or rework keying to `(node, parentContext)` if any future rewriter starts sharing.

### P3-4. `EmitFieldAssignment` reloads the receiver after the store

If the value expression mutates the receiver variable, the read-back uses the mutated value. Binder likely forbids the shape today; verify and document. (Side-effect duplication on a more exotic path than the indexer/property ones in P1.)

### P3-5. Null-coalesce path uses `dup; brtrue` which is illegal for struct values

Gated by the encoder rejecting value-type nullables today (same as P2-7). Will go from "unreachable" to "P0 invalid IL" the moment nullable value types are wired up end-to-end.

### P3-6. `EmitCatchClauses` assumes `clause.Variable` is non-null and has a slot

[...]

Issue #3828, opened 2026-09-02
test: ConversionEmitTests.CompileToAssembly leaks every emitted fixture into the default load context, so emit shards flake on test order


DavidObando:
Split out of #3818 so that issue can close with #3825, which fixes only its `cs2gs-remainder` half.
This is the half #3818 was originally filed about, now root-caused rather than hypothesised.

## Symptom

`tests (compiler-emit-remainder)`, intermittently:

```
Int32_To_Enum_Cast_Produces_Corresponding_Enum_Value [FAIL]
  gsc failed:
  test.gs(7,12,7,20): error GS0156: Cannot convert type 'P.Color' to 'Color'.
      An explicit conversion exists (are you missing a cast?)
```

Seen on `main` at `510ff7bef`
([job 100058109517](https://github.com/DavidObando/gsharp/actions/runs/33568823940/job/100058109517))
and again on PR #3825
([job 100317019106](https://github.com/DavidObando/gsharp/actions/runs/33650777321/job/100317019106)).
Note the error is a **gsc diagnostic**, not a reflection failure.

## Mechanism

`test/Compiler.Tests/Emit/ConversionEmitTests.cs`:

```csharp
compileExit = Program.Main(new[] { "/out:" + outPath, "/target:library", ... , srcPath });
...
var bytes = File.ReadAllBytes(outPath);
return Assembly.Load(bytes);
```

1. gsc runs **in-process** with **no `/reference:`**, so it resolves imports against the host's
   ambient assembly set.
2. `Assembly.Load(byte[])` puts each emitted fixture into the **default, non-collectible** load
   context, permanently.
3. `ReferenceResolver.BuildHostAssemblies()` enumerates `AppDomain.CurrentDomain.GetAssemblies()`
   and excludes only dynamic and *collectible* contexts (#3235) — so every previously-emitted
   fixture is offered to every later compile as a reference.
4. A later fixture that also declares `package P` with its own `Color` binds the earlier fixture's
   `P.Color` and the compile fails.

Order-dependent, and it does not reproduce narrowly:
`dotnet test test/Compiler.Tests -c Release --filter FullyQualifiedName~ConversionEmitTests` is
79/79 green locally. The colliding `package P` comes from another class in the shard, so the whole
`compiler-emit-remainder` process is required — the same shape as the cs2gs half, which needed
`Adr0169Gsa0005ParityTests` to run first.

## Fix

[...]

PR #3833, opened 2026-09-02
gsc: the imported-type-symbol cache must be load-context sensitive (#3826)

### LINKED ISSUE #3826: gsc: a CLR Type from one compilation reaches a later compilation's binder through a process-wide cache
Split out of #3818, where it was the underlying cause. #3825 removed the
*duplicate identities* that made it fatal; the leak itself is still there.

## What was observed

While reproducing #3818 (`Adr0169Gsa0005ParityTests` + `Adr0169AnalyzerTranslationTests`
in one test host, ~50% failure), the GS0159 site was instrumented at
`ExpressionBinder.BindAccessorCall`. On a failing run, for
`baseNode.FirstAncestorOrSelf[FunctionDeclarationSyntax]()`:

```
clr=GSharp.Core.CodeAnalysis.Syntax.SyntaxNode
recvAsmId=42766207  typeArgAsmIds=61859700  sameAsm=False
resolverRecv=64174338  recvIsCurrent=False
recvAlc=<unnamed>  recvIsDefaultAlc=False  recvInResolverList=False
```

The receiver's CLR `Type` came from an **unnamed, non-default
`AssemblyLoadContext` that is not in the current resolver's assembly list** —
i.e. a `DriverReferenceLoadContext` created by an *earlier* compilation in the
same process. The current resolver would have resolved the same full name
(`resolverRecv`) to a different instance, and the explicit type argument did.

Every symbol/member cache reachable from that path is keyed on `Type` or
`Assembly` and is cleared from `ReferenceResolver.Dispose`
(`ImportedTypeSymbol`, `MemberLookup`, `ClrTypeUtilities`, the constructed-type
caches, …), so the exact carrier was not identified. It is worth finding: any
such leak is a silent-wrong-answer hazard of the #3705 load-context family, and
it is not confined to tests — the language server and the analyzer/generator
hosts all run many compilations in one process.

## Why it matters even after #3825

#3825 makes `WithRuntimeReferences` reuse the host's assembly when the path is
already loaded, so the common case now has one identity and a leaked `Type`
is harmless. A compilation that legitimately loads a *different* assembly
privately (a real `/reference:` to a user assembly, cross-targeting) still has
two contexts alive, and a leak between two such compilations would reproduce
the same class of failure — an opaque "Cannot find function X" for a method
that exists, or a silently-taken wrong arm.

## Suggested direction

Add a debug-only invariant to `ReferenceResolver`: any `Type` the binder
receives must belong to an assembly in that resolver's own list. Running the
existing suites with it on should name the carrier in one pass.

Refs #3818, #3825, #3705, #1622.

DavidObando:
Fixes #3826.

#3825 removed the *duplicate identities* that made this fatal. The leak itself
was still there, and #3826 asked for the carrier by name. It is:

> **`ImportedTypeSymbol`'s process-wide cache — specifically its key.**

## How it carries a Type across compilations

```csharp
private static readonly ConditionalWeakTable<Assembly, ConcurrentDictionary<Type, ImportedTypeSymbol>> Cache = new();

var assemblyCache = Cache.GetValue(
    type.Assembly,
    static _ => new ConcurrentDictionary<Type, ImportedTypeSymbol>(TypeIdentityComparer.Instance));
return assemblyCache.GetOrAdd(type, static t => new ImportedTypeSymbol(t));
```

`TypeIdentityComparer` compares by `Type.AssemblyQualifiedName` — deliberately,
for #420, so the emitter collapses one logical type reached by several paths
into one `TypeRef` row. Two copies of the *same file* loaded into two different
`AssemblyLoadContext`s have **identical** assembly-qualified names, so that
comparer cannot tell them apart.

The outer `ConditionalWeakTable` bucket looks like it saves us — it is keyed on
the `Assembly` instance, which does differ per context. It does not, because it
is keyed on **the type's own assembly**, not on the assemblies the type is
*built from*. For a constructed generic whose definition lives in a shared
assembly, both compilations land in the same bucket:

| | `ImmutableArray<SyntaxNode>` from compilation A | …from compilation B |
|---|---|---|
| outer key (`type.Assembly`) | host `System.Collections.Immutable` | **same instance** |
| inner key (AQN) | `…ImmutableArray\`1[[…SyntaxNode, GSharp.Core, …]]` | **same string** |

So `ImportedTypeSymbol.Get(ImmutableArray<B.SyntaxNode>)` returns **A's**
symbol, whose `ClrType` and `TypeArguments` are A's. Every type derived from it
downstream — notably a `for` loop variable's element type — is A's too.

That is exactly the #3818 shape: `baseNode` (element of
`ImmutableArray<SyntaxNode>`) carried the *previous* compilation's `SyntaxNode`
while the explicit type argument `FunctionDeclarationSyntax` came from the
current one, `MakeGenericMethod` threw, `ClrOverloadResolution` dropped the only
candidate per C# §7.5.2, and the call reported "Cannot find function
FirstAncestorOrSelf" about a method that plainly exists.

Introduced by #2341, which changed the cache from a reference-keyed
`ConcurrentDictionary<Type, …>` (cleared on dispose) to this shape.

## Instrumented evidence

Tagging every `TypeSymbol` with a monotonic resolver generation and a creation
stack, then dumping at the GS0159 site on a failing run:

```
=== LeakProbe GS0159-imported:FirstAncestorOrSelf gen=21 symbol=ImportedTypeSymbol ===
  name=GSharp.Core.CodeAnalysis.Syntax.SyntaxNode  alc=<unnamed>
  symbolCreatedInGen=1
  creationStack:
   ImportedTypeSymbol.Get  <- TypeSymbol.FromClrType
   <- MemberLookup.MapOpenClrTypeToSymbolic  <- MemberLookup.GetClrPropertyTypeSymbol
   <- ExpressionBinder.BindAccessorStep
   <- Compilation.Emit  <- Adr0169Gsa0005ParityTests.CompileTranslatedGsa0005
  reportStack:
   ExpressionBinder.BindAccessorCall <- … <- StatementBinder.BindForRangeStatement
   <- Compilation.get_BoundProgram
   <- Adr0169AnalyzerTranslationTests.SemanticModelAnalyzer_TranslatesDeclarationAndOverrideIdioms
```

The **symbol object itself**, not merely a `Type`, was created in generation 1
(`Adr0169Gsa0005ParityTests`) and consumed in generation 21
(`Adr0169AnalyzerTranslationTests`). A bounded BFS from every static field in
`GSharp.Core` — taught to see through `ConditionalWeakTable`, which is where the
first attempt went blind — then produced the retaining path in one line:

```
rootScan: nodes=2061 found=True
  <- GSharp.Core.CodeAnalysis.Symbols.ImportedTypeSymbol.Cache [ConditionalWeakTable`2]
  <- {value} [ConcurrentDictionary`2]
  <- {value} [ImportedTypeSymbol]
```

#3826 predicted "every cache reachable on that path is `Type`- or
`Assembly`-keyed", and that is where the previous investigation stopped. It is
true and it is not sufficient: **being keyed on `Type` only helps if the key
compares by identity.** This one compares by name.

Corollary worth recording: `ImportedTypeSymbol.ClearCache()`, which
`ReferenceResolver.Dispose` calls with a comment claiming it clears the cache,
has been an **empty method** since #2341. That is a documentation defect, not
the bug — and clearing is deliberately *not* the fix (see below).

## The fix

`LoadContextSensitiveTypeComparer`: the same structural identity as
`TypeIdentityComparer`, plus a walk over the type's constituents (element type,
generic arguments) requiring the same `Assembly` **instance** at every position.
Hash stays name-based — a bucketing hint — and equality does the real test, the
same shape as the existing `TypeArgsKey`.

Two instances of one logical type reached by different paths inside a *single*
context still collapse, because a load context returns one `Assembly` instance
per identity, so #420's dedup is preserved. `ImportedTypeSymbol.Cache` is the
only process-wide consumer of `TypeIdentityComparer`; `MetadataTokenCache`'s
dictionaries are instance fields scoped to one emit, and the two
`ExpressionBinder` uses compare against a core-library type by name on purpose.

`ClearCache` stays a no-op. Clearing would only paper over the sequential case
and would leave two *live* resolvers aliasing each other; and the caches
`Dispose` really does clear have a churn hazard (symbols obtained before and
after a clear stop being reference-equal) that is not worth re-adding. The
misleading comments on both sides are corrected instead.

## Reproduction and measurement

`Issue3826CrossCompilationTypeIdentityTests` reproduces the leak
**deterministically in ~0.4s**, with no ordering dependence: it copies
`GSharp.Core.dll` to a path the host has not loaded (so #3825's path-equality
reuse does not apply and each resolver really does mint a private copy), takes
`SyntaxNode` from two successive resolvers, and closes `ImmutableArray<>` over
each.

| test | on `main` | here |
|---|---|---|
| `TwoPrivateReferenceContextsOverTheSamePath_ProduceDistinctTypes` | **passes** (anti-vacuity guard rail) | passes |
| `ConstructedGenericFromASharedAssembly_DoesNotAliasAcrossContexts` | fails | passes |
| `ACachedSymbolNeverHandsBackAnotherContextsTypeArgument` | fails | passes |

The first test is honestly labelled: it asserts only that the two contexts *are*
distinct, which is already true on `main`. It exists so a future change that
silently unified the two copies could not make the other two pass vacuously.

For the original flaky repro — `Adr0169Gsa0005ParityTests` +
`Adr0169AnalyzerTranslationTests` in one host — #3825's reuse **masks** the
symptom without fixing the leak, so measuring against it is meaningless. Both
arms below were run with that masking disabled, via this one-line local edit
(not committed):

```diff
-                Assembly? reused = TryReuseHostLoadedAssembly(fullPath);
+                Assembly? reused = Environment.GetEnvironmentVariable("GS_DISABLE_3825") == "1"
+                    ? null
+                    : TryReuseHostLoadedAssembly(fullPath);
```

```
GS_DISABLE_3825=1 dotnet test tools/cs2gs/Cs2Gs.Tests/Cs2Gs.Tests.csproj --no-build \
  --filter "FullyQualifiedName~Adr0169Gsa0005ParityTests|FullyQualifiedName~Adr0169AnalyzerTranslationTests"
```

| | runs | failures |
|---|---|---|
| without this fix | 10 | **4 (40%)** |
| with this fix | **32** (12 + 20) | **0** |

(#3825 measured the same repro at 18/43 = 42% before its own fix.)

Suites, locally: `Core.Tests` **8598/8598**; `Cs2Gs.Tests` **2664/2665**. The single failure is `Adr0151IfLetExpressionTranslationTests.Issue2819_SameVersionPackedToolAndSdk_CompileTranslatedPatternVariable`, which fails on its first line — `Assert.NotNull(ResolveLocalCs2GsPackage(repoRoot))` — because this worktree has no `out/bin/Release/nupkgs` (it has never run a Release pack). It never reaches the compiler, so it cannot be affected by this change. CI is the authority.

## Relationship to #3830

Same family, different layer, no file overlap — #3830 is the test helper
(emitted fixtures into a collectible context); this is the product cache. They
compose in the right direction: #3830 *increases* the number of distinct
collectible contexts holding same-named types in one process, which is exactly
the input the old name-only key mis-merged and the new key handles. Nothing in
`test/Compiler.Tests/Emit/ConversionEmitTests.cs` was touched here.

## The silent candidate drop — recommendation, not built

`ClrOverloadResolution.EvaluateCandidate` wraps `MakeGenericMethod` in
`catch (ArgumentException) { … return; }` at three sites. That conflates two
unrelated causes:

1. **Generic constraints not satisfied** — a genuine applicability failure.
   C# §7.5.2 says drop it silently. Correct as-is.
2. **Cross-load-context operands** — never a language-level "not applicable".
   It is a violation of the compiler's own type-plumbing invariant, and
   swallowing it converts an internal bug into a user-facing lie about a method
   that exists.

The two are separable *exactly*, without sniffing exception text: before
closing, every element of `typeArgs` must come from the same load context as
`mi.DeclaringType` — the code already reasons about this for the
`MetadataLoadContext` projection a few lines above. Recommendation, as its own
issue rather than in this PR:

- **Recommended:** detect the mismatch explicitly and report a distinct internal
  diagnostic in the `GS9998` ICE family ("cross-load-context operands closing
  `<method>`") instead of dropping. It surfaces the defect at its cause rather
  than as `GS0159` at a distant site. It needs a full-suite run of its own,
  because it turns any remaining silent misbind into a hard error — which is the
  point, but is a behaviour change.
- **Cheap fallback if that is too aggressive:** a `Debug`-only trace at the two
  catch sites recording the method, its declaring type's context, and each type
  argument's context. Zero release cost, and it would have turned this
  investigation from days into one run.
- **Rejected:** making `GS0159` explain why every candidate was dropped — too
  noisy, and it churns a large number of golden diagnostics.

Refs #3818, #3825, #3705, #1622, #2341, #420.

### DIFF
--- a/src/Core/CodeAnalysis/Symbols/ImportedTypeSymbol.cs
+++ b/src/Core/CodeAnalysis/Symbols/ImportedTypeSymbol.cs
@@ -114,9 +114,18 @@ public sealed class ImportedTypeSymbol : TypeSymbol
             throw new ArgumentNullException(nameof(type));
         }
 
+        // Issue #3826: the comparer must be load-context sensitive. Keying on
+        // the assembly-qualified name alone (TypeIdentityComparer) aliases two
+        // copies of one logical type loaded into different contexts, so a
+        // symbol — and the CLR Type it carries — created by one compilation
+        // answers a later compilation's lookup. The outer bucket does not
+        // prevent that: a constructed generic like ImmutableArray[SyntaxNode]
+        // is keyed by ITS OWN assembly (the host's System.Collections.Immutable,
+        // shared by every compilation in the process) while its argument comes
+        // from a private reference context.
         var assemblyCache = Cache.GetValue(
             type.Assembly,
-            static _ => new ConcurrentDictionary<Type, ImportedTypeSymbol>(TypeIdentityComparer.Instance));
+            static _ => new ConcurrentDictionary<Type, ImportedTypeSymbol>(LoadContextSensitiveTypeComparer.Instance));
         return assemblyCache.GetOrAdd(type, static t => new ImportedTypeSymbol(t));
     }
 
@@ -339,9 +348,16 @@ public sealed class ImportedTypeSymbol : TypeSymbol
     }
 
     /// <summary>
-    /// Legacy dispose hook. The cache is weakly keyed by assembly and each
-    /// per-assembly dictionary uses metadata identity for CLR types, so disposed
-    /// metadata contexts are collectable without process-wide clearing.
+    /// Legacy dispose hook, deliberately a NO-OP since #2341: the cache is
+    /// weakly keyed by assembly, so a disposed context's entries are
+    /// collectable without process-wide clearing.
+    /// <para>
+    /// Issue #3826: do NOT reintroduce clearing here as a fix for a
+    /// cross-compilation identity bug. Entries surviving a dispose is by
+    /// design; entries being FOUND by a different compilation's lookup was the
+    /// defect, and that is a property of the key
+    /// (<see cref="LoadContextSensitiveTypeComparer"/>), not of the lifetime.
+    /// </para>
     /// </summary>
     internal static void ClearCache()
     {
--- a/src/Core/CodeAnalysis/Symbols/LoadContextSensitiveTypeComparer.cs
+++ b/src/Core/CodeAnalysis/Symbols/LoadContextSensitiveTypeComparer.cs
@@ -1,0 +1,129 @@
+// <copyright file="LoadContextSensitiveTypeComparer.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using GSharp.Core.CodeAnalysis.Emit;
+
+namespace GSharp.Core.CodeAnalysis.Symbols;
+
+/// <summary>
+/// Issue #3826 (the #3705 load-context family): compares <see cref="Type"/>
+/// instances the way <see cref="TypeIdentityComparer"/> does — by structural
+/// identity — but ADDITIONALLY requires that every assembly the type is built
+/// from is the same <see cref="System.Reflection.Assembly"/> INSTANCE.
+/// </summary>
+/// <remarks>
+/// <para>
+/// <see cref="TypeIdentityComparer"/> keys on
+/// <see cref="Type.AssemblyQualifiedName"/>. Two copies of the same file loaded
+/// into two different <see cref="System.Runtime.Loader.AssemblyLoadContext"/>s
+/// have the SAME assembly-qualified name, so that comparer treats them as one
+/// type. That is exactly what #420 wanted for the emitter (collapsing duplicate
+/// <c>TypeRef</c> rows for one logical type reached by several paths inside ONE
+/// context), and exactly what a process-wide symbol cache must NOT do: it makes
+/// a <see cref="Type"/> from one compilation answer a later compilation's
+/// lookup.
+/// </para>
+/// <para>
+/// The dangerous case is a constructed generic whose DEFINITION lives in a
+/// shared assembly but whose ARGUMENTS come from a private reference context —
+/// <c>ImmutableArray&lt;SyntaxNode&gt;</c>, say. <c>ImportedTypeSymbol</c>'s
+/// outer cache bucket is keyed by <c>type.Assembly</c>, which for both
+/// constructions is the one host <c>System.Collections.Immutable</c>; only the
+/// inner comparer separates them, and a name-only comparer does not. The
+/// resulting cross-context symbol is silently wrong: closing a generic method
+/// over an operand from each side makes
+/// <see cref="System.Reflection.MethodInfo.MakeGenericMethod"/> throw, overload
+/// resolution drops the candidate per C# §7.5.2, and the call reports "Cannot
+/// find function &lt;name&gt;" about a method that plainly exists (#3818).
+/// </para>
+/// <para>
+/// Equality therefore keeps the structural check and adds a walk over the
+/// type's constituents (element type, generic arguments) requiring
+/// reference-equal assemblies at every position — the same "hash is a bucketing
+/// hint, equality is the real test" shape as <see cref="TypeArgsKey"/>. Two
+/// instances of one logical type reached by different paths inside a single
+/// context still collapse, because a load context returns one
+/// <see cref="System.Reflection.Assembly"/> instance per identity.
+/// </para>
+/// </remarks>
+internal sealed class LoadContextSensitiveTypeComparer : IEqualityComparer<Type>
+{
+    /// <summary>The singleton instance.</summary>
+    public static readonly LoadContextSensitiveTypeComparer Instance = new();
+
+    private LoadContextSensitiveTypeComparer()
+    {
+    }
+
+    /// <inheritdoc/>
+    public bool Equals(Type? x, Type? y)
+    {
+        if (ReferenceEquals(x, y))
+        {
+            return true;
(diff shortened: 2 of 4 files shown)


DavidObando:
CI: **35/35 green**, 0 failures (3 publish/deploy jobs skipped as usual).

Two shards worth calling out specifically:

- `tests (cs2gs-remainder)` — the gate this whole line of work is about — **passes**.
- `tests (compiler-emit-remainder)` **passes**. It was red on #3825 for the `ConversionEmitTests` flake filed as #3828; #3830 landed that half, so both halves of the family are now closed at their own layer.

The one local `Cs2Gs.Tests` failure noted in the description (`Issue2819_SameVersionPackedToolAndSdk_CompileTranslatedPatternVariable`) does not reproduce here, confirming it was the missing local `out/bin/Release/nupkgs` precondition in my worktree and not a behaviour failure.

Issue #3843, opened 2026-09-03
cs2gs: an explicit C# cast on a nullable operand becomes `as`, changing both the element type and the throw-on-wrong-type semantics


DavidObando:
Split out of **#3831 / PR #3835**, which worked around this in one file rather than fixing it. Raised by the repo owner, who asked the right question: *"does this mean other people's code in the shape that was tripping will also mistranslate?"* — yes, it does.

## The rule

`CSharpToGSharpTranslator.Invocations.cs`, `CastLowersToNullPreservingSafeCast`:

> whether TranslateCast renders this reference cast as the null-preserving `expr as T` (nullable result) instead of the non-null `cast[T](expr)` form — true when the operand is nullable on the G# side and no flow narrowing asserted it.

So a C# explicit cast whose operand is nullable becomes `as`.

## Consequence 1 — the emitted G# can be internally inconsistent and fail to compile

```csharp
// C#
.Select(t => (PragmaWarningDirectiveTriviaSyntax)t.GetStructure())
.Where(d => d is not null)
.OrderBy(d => d.SpanStart)
```

`GetStructure()` returns `SyntaxNode?`, so the cast lowers to `as` and the element type becomes `T?`. A sequence-level `.Where(d => d != nil)` filters the *sequence* without narrowing the *element type*, so every downstream member access runs on a nullable receiver. The emitted G# is inconsistent on its face:

```
.Where((d PragmaWarningDirectiveTriviaSyntax?) -> d != nil)     // nullable
.OrderBy((d PragmaWarningDirectiveTriviaSyntax) -> d.SpanStart) // NOT nullable
```

That produced `GS0158 Cannot find member SpanStart` / `DisableOrRestoreKeyword` and `GS0154`. In #3831 it took **nine apps** red, five of them banked, because the file happened to live in `Cs2Gs.Translator`.

## Consequence 2 — a silent behaviour change, present even where it compiles

| | null operand | wrong type |
|---|---|---|
| C# `(T)x` | returns null | **throws `InvalidCastException`** |
| G# `x as T` | nil | **nil** |

A cast that should throw on a type error now yields nil and flows onward. This is the #3775/#3787 defect class (a C# construct whose G# rendering behaves differently at runtime), in its more dangerous direction: an exception becomes a silent wrong value. Note this one does **not** announce itself with a compile error — consequence 1 is the loud half; this is the quiet half.

## Why the rule exists, and why the fix is not simply "stop lowering to `as`"

[...]

Issue #3553, opened 2026-08-26
cs2gs: switch-expression arms with reference conversions need a common-type wrap (GS0179/GS0263)


DavidObando:
Found in the #3501 Translator burn-down (3 of the last 29 errors).

Two related shapes where C# target-types every arm/branch but gsc requires an exact common type, and #3543's numeric widening doesn't apply:

1. Switch-expression arms mixing a G# slice literal, a LINQ result, and `Enumerable.Empty[T]()` (`CSharpToGSharpTranslator.Invocations.gs`):

```gs
let results = switch lambda.Body {
    case expression is ExpressionSyntax: []ExpressionSyntax{expression}      // slice
    case block is BlockSyntax: block.DescendantNodes(...).OfType[...]().Where(...).Select(...)  // IEnumerable[ExpressionSyntax?]
    default: Enumerable.Empty[ExpressionSyntax]()
}
// GS0179: All switch-expression arms must produce the same type
```

C# converts each arm to `IEnumerable<ExpressionSyntax>`; the translator should wrap non-conforming arms (e.g. `… as sequence[T]` / an explicit conversion) the way numeric arms now widen (#3543).

2. Conditional expression whose branches are two different `Cs2Gs.CodeModel.Ast` node types (`NonNullAssertionExpression` vs `IdentifierExpression`, `CSharpToGSharpTranslator.Patterns.gs`) — GS0263 wants an explicit conversion to the common base `GExpression`.

PR #3854, opened 2026-09-03
gsc: cast[T] accepts a nullable operand; cs2gs stops lowering explicit casts to `as` (#3843)

DavidObando:
Fixes the cause behind #3843: a C# explicit reference cast over a nullable operand no longer becomes `expr as T`.

## The pivotal fact, measured not reasoned

`cast[T](x)` over a nil/nullable operand **already had C#'s exact semantics** — for three of the four reference directions. Probed by executing, not by reading:

| direction | before | nil in | wrong type |
|---|---|---|---|
| identity `T? -> T` | binds | nil out | throws `InvalidCastException` |
| downcast `Base? -> Derived` | binds | nil out | throws `InvalidCastException` |
| interface cross-cast `IFace? -> Impl` | binds | nil out | throws |
| **upcast `Derived? -> Base`** | **GS0155, does not bind** | — | — |
| **class→interface `Impl? -> IFace`** | **GS0155, does not bind** | — | — |

So the premise recorded in #3567 — *"a C# reference cast preserves null, but `cast[T](expr)` requires a non-null operand"* — was **false for the identity and downcast directions and true only for the widening direction**. `as` was chosen for all of them.

Value-type targets were already C#-faithful and are untouched: `cast[int32](nilObject)` throws `NullReferenceException`, `cast[int32?](nilObject)` yields nil.

## The fix — two halves

**1. gsc (the language gap, per ADR-0167).** `Conversion.HasCheckedReferenceConversion` now also admits the **upcast** direction when the source is reference-nullable and the target is not, **and both sides are a NOMINAL class/interface shape** — what `castclass` can name. The nominal guard is load-bearing: probing "is `from -> to` implicit?" alone also admits conversions that are not reference conversions at all, above all the #2850 structural-function -> named-delegate *materialisation* (`newobj`, not `castclass`). Without it, `Do(Cancel(), nullableStructuralFn)` — an implicit argument conversion — degraded from GS0155 to GS0156, advertising a cast that must not exist. (See the review comment below: this predicate is a TYPE-PAIR question consulted from implicit paths too, so "only reachable behind a written cast" is not a property a caller can supply.) The widening classifies **Explicit, never Implicit**, so implicit contexts keep rejecting a nil. Every entry point routes through this one predicate (`cast[T](x)`, `cast[T?](x)`, the `T(x)` conversion-call form, and the constructor-overload fallback in `ExpressionBinder.Calls`), and the emitter needed no change: `MethodBodyEmitter.Conversions.cs:346` already emits a nil-preserving no-op for a nullable-source reference-compatible widening.

**2. cs2gs (the translator bug).** `CastLowersToNullPreservingSafeCast` is deleted. Every C# explicit reference cast lowers to `cast[T](expr)`. Its result is non-nullable `T`, exactly as C# `(T)x` is statically `T`, so `GSharpExpressionIsStaticallyNonNull` now answers `true` for all of them and the `#3683` receiver-forgiveness arm is gone with it.

ADR-0167 is amended with the operand rule and with the value-type/`Nullable[T]` split.

## Proof

**Runtime (consequence 2 — invisible to compile and ILVerify).** `Issue3843NullableOperandCheckedCastTests` executes and asserts on stdout/exception type:
- nil operand → nil out, no throw, in all four reference directions;
- **wrong-type non-nil operand → `InvalidCastException`** (the `as` rendering silently returned nil here);
- value target on nil → `NullReferenceException`; `Nullable[T]` target on nil → nil.

**Anti-vacuity.** With the binder arm disabled, 3 of the 4 new tests fail (`GS0155 Cannot convert type 'Derived3843?' to 'Base3843'`). The fourth — the value-type guard rail — passes on `main` by design.

**Compile (consequence 1).** `Issue3843NullableOperandCastTranslationTests` translates the `Select(cast).Where(x => x is not null).OrderBy(...)` shape and binds the result.

**Soundness guard rails (added in review).** `NullableStructuralFunction_ToNonNullableNamedDelegate_HasNoConversionAtAll` pins GS0155 present / GS0156 absent for the structural shape; `NullableReferencePassedImplicitlyToNonNullableParameter_IsStillRejected` pins the implicit nominal path (`Derived?` to a non-nullable `Base` parameter, `Thing?` to an `IThing` parameter, and a nullable local initializer) — all still rejected. `Issue2850StructuralFunctionToGenericDelegateEmitTests` passes **unchanged**.

**One intended behaviour change**, diagnostic ID only: a nullable local initializer in the nominal upcast shape reports GS0156 instead of GS0155. Still rejected; the advice is now accurate.

**Suites.** Core.Tests 8604/8604. Cs2Gs.Tests green (4 pre-existing assertions were pinning the old `as` shape and are inverted with the reasoning recorded inline; they were pinning *behaviour*, not intent — their doc comments asserted the false #3567 premise).

## Blast radius — whole-repo, before vs after

Two full `migrate --translate-only` runs over all 52 apps, one built from `origin/main`, one from this branch:

| | `main` | after |
|---|---|---|
| apps translating green | 52/52 | 52/52 |
| `cast[` sites | 1447 | 1522 |
| `as` reference casts | 363 | 324 |
| `!!` null assertions | 28433 | 28411 (−22) |

**Correction.** An earlier revision of this PR reported `!!` 28433 → 25094 (−11.7%) and called it a readability win. It was not: nearly all of that drop was assertions the polish pass dropped **because the loose arm had stopped gsc objecting** — lost checking, not lost noise. After re-narrowing, 3317 of those 3339 assertions come back. The genuine win is 22.

The real value of this change is the **cast-form correction**: 39 `as` renderings became `cast[…]`, each a site that previously returned nil where C# throws `InvalidCastException`.

Compile stage over the translator closure (`Cs2Gs.Translator` → `Cs2Gs.CodeModel` → `src/Core`, closure kept whole): **`src/Core` and `Cs2Gs.CodeModel` PASS compile + ilverify + test-parity** — identical before and after the re-narrowing, so restoring those assertions costs nothing in corpus health.

## The acceptance test does NOT pass — and #3843's premise is incomplete

`PragmaSuppressions.cs` unmodified from `origin/main` still fails to compile. Its errors go 5 → 1, and **none of the 5 were caused by the cast lowering.** The cast site itself is now correct:

```
.Select((t SyntaxTrivia) -> cast[PragmaWarningDirectiveTriviaSyntax](t.GetStructure()))
.Where((d PragmaWarningDirectiveTriviaSyntax?) -> d != nil)
.OrderBy((d PragmaWarningDirectiveTriviaSyntax) -> d.SpanStart)
```

The `T?` on line 2 is **not** the cast's doing. The file is an oblivious (`#nullable disable`) compilation, so it comes from cs2gs's own #1072 rule (`MapParameter` → `PromoteIfUsedAsNullable`): a parameter that is null-tested in its body is rendered `T?`. Applied to a **lambda** parameter that is an argument to an inferred generic method, that annotation feeds gsc's type-argument inference and widens the whole sequence to `sequence[T?]` — which is what produces `GS0158 Cannot find member SpanStart` / `DisableOrRestoreKeyword` and `GS0154` four lines later, inside the `for` loop. Verified by hand-patching that one annotation to `(d PragmaWarningDirectiveTriviaSyntax)` in the migrated tree: **5 errors → 1**.

The last error is a third, unrelated seam: `DirectiveIds` is emitted as `sequence[string?]` by the yield-taint rule, so `state[id] = disable` reports `GS0155 Cannot convert type 'string?' to 'string'`.

So the file was tripping on **three** independent defects, and #3843 named one of them. I am flagging this loudly rather than widening this PR: fixing the lambda-parameter promotion means changing the #1072 machinery for every lambda in the corpus, and it deserves its own issue, its own blast-radius measurement, and its own review.

**#3835 is therefore still needed as an unblock.** Its rewrite avoids all three defects by restructuring the file; this PR removes only one of them.

## Not done here

- The full 52-app gate (compile + ilverify + parity) was started and abandoned in favour of the closure run — it was hours out under machine contention. Stage 1 across all 52 apps and the compile/ilverify/parity stages for the closure are the evidence above; the full gate should run in CI on this branch.
- `CoerceReferenceValueTo` (issue #3553) carries the same `as` lowering for reference **upcast coercion**. Left alone deliberately: on a widening target a wrong-type value is unreachable, so `as` and `cast[T?]` are behaviourally identical there, and the #2412 nullable-unification contract depends on the `T?` result type.

Refs #3843, #3831, #3835, #3567, #3683, #3501. ADR-0167 amended.

### DIFF
--- a/docs/adr/0167-checked-reference-conversion-calls.md
+++ b/docs/adr/0167-checked-reference-conversion-calls.md
@@ -4,6 +4,7 @@
 - **Date**: 2026-08-18
 - **Phase**: Phase 9 — language depth / CLR conversion parity
 - **Related**: ADR-0045 (boxing/unboxing), ADR-0115 (cs2gs), ADR-0120 (user conversions), ADR-0160 (`as` yields nullable), issue [#3421](https://github.com/DavidObando/gsharp/issues/3421), parent [#3394](https://github.com/DavidObando/gsharp/issues/3394)
+- **Amended**: 2026-09-02 — nullable operands (issue [#3843](https://github.com/DavidObando/gsharp/issues/3843))
 
 ## Context
 
@@ -33,6 +34,21 @@ construction ambiguity exists.
 - A compatible non-null value returns the same reference typed as `T`/`T?`.
 - A null reference stays null.
 - An incompatible non-null value throws `InvalidCastException`.
+- **The operand may be reference-nullable.** `cast[T](x)` accepts an `S?`
+  operand for every reference direction — identity (`T? -> T`), downcast
+  (`Base? -> Derived`), interface cross-cast, and widening (`Derived? -> Base`,
+  `Impl? -> IFace`) — and its result is **non-nullable `T`**, exactly as C#
+  `(T)x` is statically `T` however maybe-null its operand. Nil flows through
+  unchanged; a non-nil value of an incompatible runtime type still throws.
+  Like C#, this means an explicitly written cast is the one place a nil can
+  inhabit a non-nullable static type, and the dereference that follows raises
+  the same `NullReferenceException` C# raises at the same point. This is
+  deliberately confined to the EXPLICIT `cast`/conversion-call spelling: no
+  implicit conversion drops reference nullability, and the Kotlin-model
+  GS0154/GS0155 rules are untouched.
+- Value-type and `Nullable[T]` targets keep C#'s own (different) rules:
+  `cast[int32](nilObject)` throws `NullReferenceException`, `cast[int32?]
+  (nilObject)` yields nil.
 - `T?` changes static nullability only; it does not turn the cast into a test.
 - `value as T` remains the nullable testing conversion from ADR-0160.
 - Numeric, enum, boxing, unboxing, user-defined, checked, and unchecked
@@ -51,7 +67,14 @@ cannot resolve to a constructor. `T?(value)` remains unambiguously a nullable
 conversion target.
 
 cs2gs maps every C# explicit reference cast `(T)value` to `cast[T](value)` or
-`cast[T?](value)`. Explicit boxing casts to non-`object` reference targets use
+`cast[T?](value)` — including when `value` is nullable. It briefly (issues
+#3567/#3683) lowered a nullable-operand cast to `value as T` instead, on the
+premise that `cast[T]` required a non-null operand. That premise held only for
+the widening direction, which issue #3843 closed, and the `as` rendering was
+wrong twice over: its `T?` result injected a nullable element type into
+non-nullable continuations (`Select(cast).Where(x => x != nil).OrderBy(…)`
+stopped compiling), and it turned a wrong-type cast from an
+`InvalidCastException` into a silent nil. Explicit boxing casts to non-`object` reference targets use
 the same constructor-independent spelling. Numeric/value casts keep
 `T(value)` / `T?(value)`. C# `as` expressions continue to map to G# `as`.
 
--- a/src/Core/CodeAnalysis/Binding/Conversion.cs
+++ b/src/Core/CodeAnalysis/Binding/Conversion.cs
@@ -1854,6 +1854,32 @@ public sealed class Conversion
             return true;
         }
 
+        // Issue #3843: the UPCAST direction of the same checked reference
+        // conversion, reached only when the source is reference-nullable and
+        // the target is not — `cast[Node](blockOrNil)`, the G# rendering of
+        // C# `(Node)node.Body`. The reverse probe above only admits the
+        // DOWNCAST direction (`Base? -> Derived`), and the identity case is
+        // already answered by `removesReferenceNullability` above, so a plain
+        // widening that also drops the `?` had no checked-conversion
+        // classification at all and reported GS0155. The runtime story is the
+        // same one ADR-0167 already guarantees for the other two directions:
+        // the value is a managed reference, `castclass`/no-op preserves nil,
+        // and a non-nil value of an incompatible runtime type still throws
+        // `InvalidCastException` — for a widening target the incompatible case
+        // is simply unreachable. Restricted to `removesReferenceNullability`
+        // so a non-nullable widening keeps classifying as the implicit upcast
+        // it is and no IMPLICIT nullability-dropping path is opened: this arm
+        // is only ever consulted behind an explicitly written conversion.
+        if (removesReferenceNullability
+            && ClassifyCore(
+                from,
+                to,
+                allowStructuralProjection: false,
+                allowExplicitReference: false).IsImplicit)
+        {
+            return true;
+        }
+
         if (from.ClrType is { } fromClr
             && to.ClrType is { } toClr
             && !fromClr.IsValueType
--- a/src/Core/CodeAnalysis/Binding/Conversion.cs
+++ b/src/Core/CodeAnalysis/Binding/Conversion.cs
@@ -1866,11 +1866,26 @@ public sealed class Conversion
         // the value is a managed reference, `castclass`/no-op preserves nil,
         // and a non-nil value of an incompatible runtime type still throws
         // `InvalidCastException` — for a widening target the incompatible case
-        // is simply unreachable. Restricted to `removesReferenceNullability`
-        // so a non-nullable widening keeps classifying as the implicit upcast
-        // it is and no IMPLICIT nullability-dropping path is opened: this arm
-        // is only ever consulted behind an explicitly written conversion.
+        // is simply unreachable.
+        //
+        // BOTH SIDES MUST BE NOMINAL. This restriction is load-bearing, not
+        // decorative: "is `from -> to` implicit?" on its own also admits
+        // conversions that are not REFERENCE conversions at all — above all
+        // the issue #2850 structural-function -> named-delegate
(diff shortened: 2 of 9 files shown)


DavidObando:
## The CI failure was a real soundness regression. You were right on every point, including the one about the `!!` number.

I have **not** touched `Issue2850StructuralFunctionToGenericDelegateEmitTests` — it passes unchanged (4/4).

### 1. The mechanism

The leak was **not** implicit-vs-explicit context, and my "only reachable behind a written conversion" claim was wrong for a more basic reason: `HasCheckedReferenceConversion` is a **type-pair predicate**, and `Conversion.Classify(...).Exists` is what selects GS0155 ("no conversion") vs GS0156 ("an explicit conversion exists — are you missing a cast?"). No caller can supply "a cast was written", so the restriction has to live in the predicate itself. Gating on `removesReferenceNullability` was true but irrelevant.

The actual fault was in my probe:

```csharp
ClassifyCore(from, to, allowStructuralProjection: false, allowExplicitReference: false).IsImplicit
```

"Is `from → to` implicit?" admits conversions that are **not reference conversions at all**. The one that bit is exactly the shape the test guards: a `FunctionTypeSymbol` carries a `Func<…>`/`Action<…>` CLR backing, so it *looks* class-like, but structural-function → named-delegate is the #2850 **materialisation** — it emits `newobj`, not `castclass`. My arm accepted it as a checked reference cast, so `Do(Cancel(), ca)` degraded from GS0155 to GS0156, advertising a cast that must not exist.

Observed on the first cut:
```
error GS0156: Cannot convert type '((int32, Cancel, (string) -> void) -> void)?' to 'Conv[Cancel]'.
              An explicit conversion exists (are you missing a cast?)
```

One correction to the framing, for the record: the nil was still **rejected** at that site — the diagnostic ID changed, no value flowed through silently. That makes it less severe than "no diagnostic at all", but it is still wrong and still had to go: it advertised an unsound cast, and — see §4 — it did real damage elsewhere.

### 2. The re-narrowing

Both sides must now be a **nominal** class/interface shape — what `castclass` can name. G#'s structural shapes are excluded even when class-like: `FunctionTypeSymbol`, `FunctionPointerTypeSymbol`, `SliceTypeSymbol`, `ArrayTypeSymbol`, `RectangularArrayTypeSymbol`, `TupleTypeSymbol`, `SequenceTypeSymbol`, `AsyncSequenceTypeSymbol`, `TypeParameterSymbol` — each converts by element/signature rules with its own arm in `ClassifyCore`.

`Issue2850…ReportsGS0155` passes unchanged. All four reference directions the fix targets still work (identity, downcast, cross-cast, upcast — nil preserved, wrong type throws).

### 3. Regression tests added

- `NullableStructuralFunction_ToNonNullableNamedDelegate_HasNoConversionAtAll` — the exact shape, asserting GS0155 present and **GS0156 absent**, in `Core.Tests` next to the rest of the #3843 rationale.
- `NullableReferencePassedImplicitlyToNonNullableParameter_IsStillRejected` — the implicit **nominal** reference path, which nothing pinned: `Derived?` → non-nullable `Base` parameter, `Thing?` → `IThing` parameter, and a nullable local initializer. All three still rejected.
- A binder assertion that the widening classifies **Explicit, never Implicit**.

**Item 4 — the other three directions.** Identity and downcast are unchanged from `main` (they use the pre-existing `removesReferenceNullability` short-circuit and the reverse probe). I probed the downcast direction of the same structural pair — a nullable named delegate passed implicitly to a non-nullable structural-function parameter — and it is correctly rejected with GS0154. Argument positions in the nominal case also still report GS0154, unchanged.

**One intended behaviour change remains, and it is a diagnostic ID only.** A nullable local *initializer* in the nominal upcast shape (`var b Base = derivedOrNil`) now reports GS0156 instead of GS0155. The value is still rejected; the advice is now accurate, since `cast[Base](…)` genuinely exists and is C#-faithful. That case is pinned by the new test.

### 4. Corrected blast radius — the `!!` figure was mostly bogus, as you suspected

Three full whole-repo `migrate --translate-only` runs (all 52 apps), same corpus, same harness:

| | `main` | first cut (loose) | **re-narrowed** |
|---|---|---|---|
| apps translating green | 52/52 | 52/52 | **52/52** |
| `cast[` sites | 1447 | 1521 | **1522** |
| `as` reference casts | 363 | 324 | **324** |
| **`!!` null assertions** | **28433** | 25094 (−11.7%) | **28411 (−22, −0.08%)** |

**I withdraw the −11.7% figure.** It was not "a portion" of lost checking — it was essentially all of it. Of the 3339 assertions the loose arm dropped, **3317 came back** once the arm was narrowed; the genuine, faithful win is **22**. Diffing the loose tree against the narrow one, 3015 of the restored lines are `!!` restorations. Those were assertions the polish pass dropped because gsc had stopped objecting — exactly the failure mode you named.

The real value of this change is therefore the **cast-form correction**, not a readability win: 39 `as` renderings became `cast[…]`, each one a site that previously returned nil where C# throws `InvalidCastException`.

Corpus health is unaffected by the re-narrowing: closure compile stage is identical either way — **`src/Core` PASS compile + ilverify + test-parity** with all its `!!` restored, `Cs2Gs.CodeModel` PASS. `Core.Tests` 8604/8604.

### Unchanged from the earlier report

`PragmaSuppressions.cs` still does not compile (5 errors, none from the cast) — the two remaining causes are #3855 and the `DirectiveIds` yield-taint seam. Thank you for filing #3855.

PR #3859, opened 2026-09-03
selfmig: take test/Core.Tests green — doc @param mismatch + params `is null` guard (#3501)

DavidObando:
Takes `test/Core.Tests` — the smallest remaining wall in the #3501 self-migration gate, 2 errors on gate run 33743456719 (44/52, `0753a162c`) — to zero, and clears one of `test/Compiler.Tests`' 9 fingerprints with the same change.

The two errors are unrelated to each other. Both are gsc **warnings** that become errors because the migrated tree inherits the repo's `TreatWarningsAsErrors=true`.

## 1. `GS0229` — `Documentation @param 'scenario' does not match any parameter of 'Resolution_is_deterministic'`

**Not a translation defect.** `test/Core.Tests/CodeAnalysis/Binding/OverloadResolutionPropertyTests.cs` documents five FsCheck property methods with a `<param name="scenario">` tag naming a parameter none of them has (they take `candidates`/`argTypes`, `argumentType`/`distractors`, `source`/`firstIndex`/`secondIndex`). cs2gs translates the doc comment faithfully to `/// @param scenario …`; gsc's `DocumentationValidator.ValidateParamMatches` then reports the mismatch.

csc agrees this is a defect. The equivalent C# compiled with `GenerateDocumentationFile=true` gives:

```
warning CS1572: XML comment has a param tag for 'scenario', but there is no parameter by that name
warning CS1573: Parameter 'candidates' has no matching param tag in the XML comment for 'C.M(int, int)'
```

It went unseen only because `build/gsharp.build.props` sets `<Nullable>disable` for test projects and nothing turns on doc generation for them, so csc never emits CS1572. Fixed at the source: the tags now name the real parameters. GS0229 is exactly CS1572 — nothing to change in gsc or cs2gs.

## 2. `GS0523` — `Comparison of non-nullable '[]string' with 'nil' is always false`

**Neither of the two shapes that were hypothesised.** The C# did not declare `string[]?` and lose the annotation, and it is not a pointless comparison C# merely warns about. The operand is a C# **`params` array**:

```csharp
// test/Shared/EmittedFixture.cs (#3830)
public static Assembly[] LoadTogether(params string[] assemblyPaths)
{
    if (assemblyPaths is null)
    {
        throw new ArgumentNullException(nameof(assemblyPaths));
    }
```

A `params string[]` translates to a G# variadic `paths ...string`, whose parameter type is a bare `[]string` — which under ADR-0159 can never be nil. Verified directly against gsc, which reproduces the fingerprint's exact wording and type spelling:

```
probe.gs(6,8,6,20): warning GS0523: Comparison of non-nullable '[]string' with 'nil'
is always false — a bare (non-'?') collection-typed value can never be nil (ADR-0159); …
        if paths == nil {
```

**This is an established, already-decided shape, not a new policy question.** cs2gs has folded exactly this guard since the `Diagnostic.Create` blocker: `TryFoldParamsArrayNullCheck` in `CSharpToGSharpTranslator.Statements.cs` folds `args == null` / `args != null` on a `params` parameter to its constant, and `CSharpToGSharpTranslator.Patterns.cs` skips the same guard for a `params` recursive-pattern subject (`args is { Length: > 0 }`).

**The gap is the spelling.** The fold matched only `BinaryExpressionSyntax`. `EmittedFixture.cs` writes the guard as the constant-null **pattern** `assemblyPaths is null`, which is an `IsPatternExpressionSyntax`, so it fell straight through to `x == nil`. This PR extends the fold to the pattern spelling — `args is null` → `false`, `args is not null` → `true` — keyed on the same `params` predicate, now extracted as the shared `IsParamsArrayOperand`.

### The `Compiler.Tests` GS0523 is the same root cause, not just the same message

`test/Shared/EmittedFixture.cs` is `<Compile Include>`-linked into **both** `Core.Tests` and `Compiler.Tests`. One source line, one translation, two migrated apps. Fixing it clears Core.Tests' second error and one of Compiler.Tests' nine.

## Not a suppression

The folded output is legal, not merely quiet — it compiles clean and runs:

```
$ gsc probe2.gs /out:probe2.dll     # `if false { throw … }` + `return paths.Length`
Wrote probe2.dll
$ dotnet probe2.dll
2
```

No diagnostic was weakened, no `/nowarn` added, no `!` introduced. `GS0523` still fires on every non-`params` bare collection.

## Tests

Three new cases in `Cs2Gs.Tests/Issue3501ResidualSyntheticRetargetTests.cs`, beside the existing `params` fold tests:

- `ParamsArrayIsNullPattern_FoldsAway`
- `ParamsArrayIsNotNullPattern_FoldsAway`
- `NonParamsArrayIsNullPattern_KeepsTheGuard` — anti-vacuity: an ordinary `string[]?` parameter keeps its `is null` guard, so the fold is keyed on `params`, not on "array compared to null".

They assert on the printed G# **and** bind it and assert no `GS0523` is reported. `TranslationTestValidation.AssertBinds` alone cannot gate this — GS0523 is a warning and `AssertBinds` only filters `IsError` — so the helper re-binds and checks the id explicitly.

**Failing-on-main proof.** With the two new `case` guards disabled (`&& false`), both fold tests fail with the pre-fix output, while the anti-vacuity test keeps passing:

```
ParamsArrayIsNullPattern_FoldsAway [FAIL]
  String: ···"           if paths == nil {\n            "···
ParamsArrayIsNotNullPattern_FoldsAway [FAIL]
  String: ···"    return if paths != nil { paths!!.Leng"···
Failed! - Failed: 2, Passed: 3
```

Restored, the whole `Issue3501ResidualSyntheticRetargetTests` class passes 19/19 and `GoldenTests` passes 42/42 — no golden churn.

## Still outstanding

The end-to-end migrated-run before/after for both apps is still in flight locally (whole-repo `migrate` + `validate --app`); it will be attached when it lands. `tools/cs2gs/selfmig-baseline.json` is untouched.

### DIFF
--- a/test/Core.Tests/CodeAnalysis/Binding/OverloadResolutionPropertyTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Binding/OverloadResolutionPropertyTests.cs
@@ -19,7 +19,8 @@ public class OverloadResolutionPropertyTests
     /// <summary>
     /// Resolving the same candidate set and arguments repeatedly is deterministic.
     /// </summary>
-    /// <param name="scenario">The generated overload-resolution input.</param>
+    /// <param name="candidates">The generated candidate method set.</param>
+    /// <param name="argTypes">The generated argument types.</param>
     /// <returns><see langword="true"/> when both attempts produce the same result.</returns>
     [Property(MaxTest = 300, Arbitrary = [typeof(OverloadResolutionGenerators)])]
     public bool Resolution_is_deterministic(MethodInfo[] candidates, Type[] argTypes)
@@ -35,7 +36,8 @@ public class OverloadResolutionPropertyTests
     /// <summary>
     /// Resolution results maintain their public shape invariants.
     /// </summary>
-    /// <param name="scenario">The generated overload-resolution input.</param>
+    /// <param name="candidates">The generated candidate method set.</param>
+    /// <param name="argTypes">The generated argument types.</param>
     /// <returns><see langword="true"/> when the outcome and payload agree.</returns>
     [Property(MaxTest = 300, Arbitrary = [typeof(OverloadResolutionGenerators)])]
     public bool Outcome_payloads_match_outcome(MethodInfo[] candidates, Type[] argTypes)
@@ -57,7 +59,8 @@ public class OverloadResolutionPropertyTests
     /// <summary>
     /// Every reported best or ambiguous method came from the original candidate set.
     /// </summary>
-    /// <param name="scenario">The generated overload-resolution input.</param>
+    /// <param name="candidates">The generated candidate method set.</param>
+    /// <param name="argTypes">The generated argument types.</param>
     /// <returns><see langword="true"/> when all returned methods were supplied as candidates.</returns>
     [Property(MaxTest = 300, Arbitrary = [typeof(OverloadResolutionGenerators)])]
     public bool Resolution_only_returns_input_candidates(MethodInfo[] candidates, Type[] argTypes)
@@ -77,7 +80,8 @@ public class OverloadResolutionPropertyTests
     /// <summary>
     /// An applicable identity conversion wins over widening, boxing, and reference conversions.
     /// </summary>
-    /// <param name="scenario">The generated exact-match scenario.</param>
+    /// <param name="argumentType">The generated argument type whose exact one-parameter method must win.</param>
+    /// <param name="distractors">The generated non-exact candidates competing with it.</param>
     /// <returns><see langword="true"/> when the exact candidate resolves.</returns>
     [Property(MaxTest = 200, Arbitrary = [typeof(OverloadResolutionGenerators)])]
     public bool Exact_match_is_preferred(Type argumentType, MethodInfo[] distractors)
@@ -96,7 +100,9 @@ public class OverloadResolutionPropertyTests
     /// <summary>
     /// Numeric better-conversion ordering is antisymmetric and drives overload choice.
     /// </summary>
-    /// <param name="scenario">The generated numeric-betterness scenario.</param>
+    /// <param name="source">The generated source type of the single argument.</param>
+    /// <param name="firstIndex">Index selecting the first numeric candidate.</param>
+    /// <param name="secondIndex">Index selecting the second numeric candidate.</param>
     /// <returns><see langword="true"/> when the expected numeric target resolves.</returns>
     [Property(MaxTest = 300, Arbitrary = [typeof(OverloadResolutionGenerators)])]
     public bool Numeric_betterness_is_antisymmetric_and_selects_expected_candidate(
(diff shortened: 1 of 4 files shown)


DavidObando:
## End-to-end migrated-run result — and a correction to the premise

Whole-repo `cs2gs migrate` (52/52 apps translated) then `cs2gs validate --app test/Core.Tests --app test/Compiler.Tests`, on `09dc4a69d` with the Release SDK nupkg repacked (`gscVersion 0.4.400+09dc4a69d2`).

| app | before (gate 33743456719, `0753a162c`) | after (`09dc4a69d`) |
|---|---|---|
| `test/Core.Tests` | compile **FAIL(2)** — 1× GS0229, 1× GS0523; ilverify/test-parity skipped | compile **PASS**; ilverify **FAIL(19)** |
| `test/Compiler.Tests` | compile **FAIL(9)** | compile **FAIL(8)** |

**Both original errors are gone.** Core.Tests clears the compile stage for the first time, and Compiler.Tests drops 9 → 8 with no GS0523 remaining — its 8 survivors are 6× GS0159, 1× GS0154, 1× GS0158, all in the single file `Emit/Issue2525ImportedIndexerHidingEmitTests.gs`. That is the shared root cause confirmed end-to-end, exactly as predicted.

### Core.Tests is NOT green — the premise was understated

The task framed Core.Tests as "down to 2 errors, the smallest remaining wall". That was the **compile** wall. Behind it sits an ILVerify wall of 19 that the compile failure had been hiding, now exposed for the first time:

- 11× `UnsatisfiedMethodParentInst` — `GSharpAnalyzerVerifier<object>` instantiated without satisfying its class type-parameter constraints (`Adr0169AnalyzerVerifierTests`, `Issue3795SymbolContainmentTests`, and two lambda hosts)
- 7× `StackUnexpected` — `AppendFormatted(!!0)` / `AppendFormatted(!!0, int32[, string])` on `ReceiverInterpolatedStringHandler` and `PrefixedInterpolatedStringHandler`
- 1× `UnmanagedPointer` — `Issue2852ImportedPointerFields::set_Pointer(int32*)`, not a verifiable type

**None of these is caused by this PR.** They are three unrelated emit defects (generic-constraint instantiation, interpolated-string-handler generic argument, pointer field), none touches `params`, nil comparison, or doc comments, and no artifact references `EmittedFixture` or `LoadTogether`. Closing Core.Tests needs a follow-up on those three families; I'd expect the `UnsatisfiedMethodParentInst` group (11 of 19, one root cause) to be the tractable next step.

The gate counts an app green only when all four stages pass, so **Core.Tests stays red and the 44/52 green count is unchanged** by this PR. What changed is the wall: 2 compile errors → 0, and a previously invisible 19-failure ILVerify wall is now measurable. `tools/cs2gs/selfmig-baseline.json` is untouched; nothing goes below the floor.

DavidObando:
Follow-up for the newly-exposed ILVerify wall filed as #3863 (11× UnsatisfiedMethodParentInst, 7× StackUnexpected, 1× UnmanagedPointer — three unrelated families, none caused by this PR).

Issue #3685, opened 2026-08-30
cs2gs + gsc: C# array covariance has no G# spelling — migrated 'Derived[] -> Base[]' fails with GS0155 and cast[] is rejected too


DavidObando:
Part of the #3501 self-migration wall for `test/InternalAnalyzers.Tests` (green run [33308959087](https://github.com/DavidObando/gsharp/actions/runs/33308959087)).

## Symptom

```
AnalyzerTestHelper.gs(65): error GS0155: Cannot convert type
  'Microsoft.CodeAnalysis.PortableExecutableReference[]' to '[]Microsoft.CodeAnalysis.MetadataReference'
```

The C# source is an ordinary array-covariance return:

```csharp
// test/InternalAnalyzers.Tests/AnalyzerTestHelper.cs
private static MetadataReference[] GetReferences()
{
    return trustedPlatformAssemblies.Split(Path.PathSeparator)
        .Where(path => needed.Contains(Path.GetFileName(path)))
        .Select(path => MetadataReference.CreateFromFile(path))   // PortableExecutableReference
        .ToArray();                                               // PortableExecutableReference[]
}
```

`ToArray()` yields `PortableExecutableReference[]`; C# widens it to the declared
`MetadataReference[]` implicitly via array covariance. cs2gs emits the operand
bare, and gsc rejects it.

## gsc is right to reject the IMPLICIT form — and that is the whole problem

G# slices are **invariant** in their element type by design (gsc issue #2516;
`Conversion.ClassifyCore` stops the array arms before the general CLR
reference-upcast rule precisely so covariance cannot be admitted silently — the
`ArrayTypeMismatchException` hazard must be announced). ADR-0016 is unchanged:
a `[]T` IS a CLR `T[]`.

The defect is that there was **no spelling at all** for the widening. The
explicit form was rejected too:

```
$ cat cov.gs
func FromSlice(d []DirectoryInfo) []FileSystemInfo {
    return cast[[]FileSystemInfo](d)      // <- also rejected, pre-fix
}
$ gsc cov.gs
error GS0155: Cannot convert type '[]System.IO.DirectoryInfo' to '[]System.IO.FileSystemInfo'.
```

Both array arms of `Conversion.ClassifyCore` (the `SliceTypeSymbol` arm and the
imported-SZ-array arm that `Enumerable.ToArray()` surfaces) offered only
`HasCheckedReferenceConversion` as their `allowExplicitReference` fallback —
[...]

Issue #2516, opened 2026-07-19
cs2gs: C# array covariance is emitted as unsupported invariant slice conversion


DavidObando:
## Summary

cs2gs emits a C# array covariance conversion as a bare G# slice argument. C# permits `Derived[]` to flow to covariant generic interfaces such as `IEnumerable<IBase>`, but G# deliberately treats slices as invariant and rejects `[]Derived -> IEnumerable[IBase]`. Valid C# therefore becomes uncompilable G# with GS0155.

This is independent of #2514 and #2511. The minimal repro has no constrained type-parameter member access or nullable index argument.

## Oahu evidence

Fresh post-#2510/#2512 migration with `gsc 0.3.162+1736f145ad`:

- Canonical fingerprint: `sha256:d9ad92c2997f8882872dca81d620c34a77f35b35fd169ad9a93957e82fcf9474`
- Generic parameter: `src/Oahu.Core/BookLibrary.cs:905-910`
- Call site: `BookLibrary.cs:1449`
- Concrete source type: `src/Oahu.Core/Models/LibraryResponse.cs:315-321`
- Emitted G#: `corpus_Oahu.Core/BookLibrary.gs:569:74`
- Diagnostic: `GS0155: Cannot convert type '[]Author' to 'IEnumerable<IPerson>?'.`

Original C# shape:

```csharp
private static void AddPersons<TPerson>(
    ...,
    IEnumerable<Oahu.Audible.Json.IPerson> itmPersons,
    ...)
{
}

// Product.Authors is Oahu.Audible.Json.Author[]
AddPersons(..., product.Authors, ...);
```

`Oahu.Audible.Json.Author : Oahu.Audible.Json.IPerson`, and `IEnumerable<out T>` is covariant, so C# accepts the reference conversion.

Current G# passes the translated `[]Author` bare to `IEnumerable[IPerson]?`, which the compiler correctly rejects under G# slice invariance.

### Referenced project

```csharp
namespace Probe;
public interface IPerson { }
public sealed class Author : IPerson { }
```

### Consumer project

```csharp
using System.Collections.Generic;
using Probe;

namespace Consumer;

public static class Repro
{
    public static void Accept(IEnumerable<IPerson>? values) { }
    public static void Test(Author[] values) => Accept(values);
}
```

Both C# projects build successfully. Default `cs2gs migrate --via-sdk` emits:

```gsharp
package Consumer

import System.Collections.Generic
import Probe

class Repro {
    shared {
[...]

PR #3861, opened 2026-09-03
cs2gs: object-initializer members keep C#'s implicit array covariance (#3858)

### LINKED ISSUE #3858: cs2gs: implicit array covariance is dropped at object-initializer members — the migrated test/Sdk.Tests wall (7x GS0156, #3501)
The migrated `test/Sdk.Tests` is red in the self-migration gate with a single
diagnostic repeated across the app:

```
GS0156  Cannot convert type '[]Microsoft.Build.Utilities.TaskItem'
                          to 'Microsoft.Build.Framework.ITaskItem[]'
```

This is C# **array covariance** (`TaskItem : ITaskItem`, so `TaskItem[]`
converts implicitly to `ITaskItem[]`), which G# deliberately does not do
implicitly: slices are invariant (#2516) and the covariant upcast must be
**spelled** as `cast[[]Base](expr)` (#3685). The translator already knows this
— `CoerceCovariantArrayConversion` exists and is applied at the argument,
return, local-initializer and assignment-statement positions.

**It is not applied to object-initializer members.** Every failing site in
`test/Sdk.Tests` has this shape:

```csharp
var task = new BuildTask
{
    Compile = new[] { new TaskItem("Program.gs") },   // ITaskItem[] property
    References = Array.Empty<TaskItem>(),
    ...
};
```

A C# object-initializer member is an assignment position, so Roslyn records the
implicit `TaskItem[] -> ITaskItem[]` conversion on the RHS — but
`TranslateObjectInitializerFields` (the `T{ … }` composite-literal form) and
`BuildConstructionWithInitializerSuffix` (the `T(args) { … }` form) both call
`TranslateExpression` directly and drop the conversion, so the migrated G# is

```
BuildTask{Compile: []TaskItem{ TaskItem("Program.gs") }, ...}
```

and gsc correctly reports GS0156.

Reproduced in isolation (no MSBuild types needed):

```csharp
public class Holder { public System.IComparable[] Items { get; set; } }
static Holder Make(string[] names) => new Holder { Items = names };
```

renders as `Holder{Items: names}` (bare) instead of
`Holder{Items: cast[[]IComparable](names)}`. The `T(args) { … }` form renders
`Holder(1){Items = names}`, same defect.

**Layer:** cs2gs only. The gsc side is complete — #3685 added the explicit-only
covariant upcast arm to `Conversion.ClassifyCore`, it is reference-element-only
(`IsReferenceLikeTarget` on both elements plus CLR assignability), and a value
element (`[]int32` → `[]object`) is still rejected, matching C#. #3854's
reference-conversion changes are adjacent code but not the cause; this is a
missing translation position, not a conversion rule.

**Fix:** route the object-initializer member RHS through
`CoerceCovariantArrayConversion` at both positions, exactly as
`TranslateAssignmentValue` does.

DavidObando:
Fixes #3858.

## The wall

The migrated `test/Sdk.Tests` was red in the self-migration gate with one
diagnostic repeated across the app:

```
GS0156  Cannot convert type '[]Microsoft.Build.Utilities.TaskItem'
                          to 'Microsoft.Build.Framework.ITaskItem[]'
```

## Cause — one, and it is in cs2gs

C# array covariance (`TaskItem : ITaskItem`, so `TaskItem[]` widens implicitly
to `ITaskItem[]`) is not implicit in G#: slices are invariant by design (#2516),
and the covariant upcast must be spelled `cast[[]Base](expr)` (#3685). The
translator already knows this — `CoerceCovariantArrayConversion` is applied at
the argument, return, local-initializer and assignment-statement positions.

It was **not** applied to object-initializer members, which is exactly the shape
every failing site uses:

```csharp
var task = new BuildTask
{
    Compile = new[] { new TaskItem("Program.gs") },   // ITaskItem[] property
    References = Array.Empty<TaskItem>(),
};
```

A C# object-initializer member is an assignment position, so Roslyn records the
implicit conversion on the RHS; both member builders
(`TranslateObjectInitializerFields` for `T{ … }` and
`BuildConstructionWithInitializerSuffix` for `T(args) { … }`) called
`TranslateExpression` directly and dropped it.

## The fix

Route the object-initializer member RHS through
`CoerceCovariantArrayConversion`, at both builders, exactly as
`TranslateAssignmentValue` does. Nothing in the conversion rules moved: this
only makes the translator *spell* a conversion C# already performed.

## Reference vs value elements

Unchanged and matching C# exactly:

- **Reference elements** — covariant upcast is legal, explicit-only. `cast[[]B](d)`
  is a reference-level no-op on the same array instance.
- **Value elements** — rejected. `int[]` does not convert to `object[]` in C#,
  implicitly or explicitly, and gsc's `IsCovariantArrayUpcast` requires
  `IsReferenceLikeTarget` on both elements plus CLR assignability, so
  `[]int32 -> []object` stays an error. `CoerceCovariantArrayConversion`'s own
  guard is `IsReferenceType` on both elements, so it never wraps one either.

## Not #3854

PR #3854's reference-conversion classification work is adjacent code but not the
cause here. Nothing in `Conversion.cs`/`ConversionClassifier.cs` changed in this
PR — the gsc-side covariant arm from #3685 was already correct, which the
executable tests below demonstrate on `origin/main`.

## Evidence

Targeted self-migration run over the exact `test/Sdk.Tests` dependency closure
(`Sdk.Tests -> Gsharp.NET.Sdk -> {HotReload.Runtime, Compiler} -> Core ->
{InternalAnalyzers, Analyzers.Testing}`, everything else `--exclude`'d, so the
discovered set is the closure and no project reference dangles). Full pipeline:
translate + compile + ILVerify + test-parity.

**Before** (`c9943bee3`, `origin/main`): 6/7 green, `test/Sdk.Tests` red with
**15** GS0156 sites across 5 files (`BuildTaskDesignTimeTests`,
`Adr0169GsAnalyzerForwardingTests`, `Adr0169EditorConfigSeverityTests`,
`Issue3782WarningsNotAsErrorsTests`, `HotReloadArtifactsTaskTests`) — all one
message, all the object-initializer shape. The gate's "7 fingerprints" is a
deduplicated signature count, not the site count.

**After**: `test/Sdk.Tests` green (see the run summary in the PR discussion).

### Tests

`tools/cs2gs/Cs2Gs.Tests/Issue3858CovariantArrayObjectInitializerTranslationTests.cs`
— 5 tests, each round-trip **binding** the printed G# through gsc:

| test | on `origin/main` | with fix |
| --- | --- | --- |
| `CovariantArray_ObjectInitializerMember_EmitsExplicitCast` | FAIL | pass |
| `CovariantArray_ConstructionWithInitializerSuffix_EmitsExplicitCast` | FAIL | pass |
| `CovariantArray_ObjectInitializerFieldMember_EmitsExplicitCast` | FAIL | pass |
| `SameElementArray_ObjectInitializerMember_IsNotWrapped` | pass (guard rail) | pass |
| `ValueElementArray_ObjectInitializerMember_IsNotWrapped` | pass (guard rail) | pass |

`test/Core.Tests/CodeAnalysis/Emit/Issue3685CovariantArrayUpcastTests.cs` gains
two **executing** tests. Both pass on `origin/main` — no gsc change is claimed;
they are the runtime witness and the anti-vacuity guard for the conversion this
PR now emits:

- `CovariantUpcast_StoringAWrongElement_ThrowsArrayTypeMismatchException` —
  compiles, runs, and asserts that storing a `FileInfo` into a
  `cast[[]FileSystemInfo](dirs)` widened `DirectoryInfo[]` throws
  `ArrayTypeMismatchException`, exactly as C# does. The unsoundness is
  preserved, not papered over.
- `ValueElementUpcast_IsStillRejected` — `cast[[]object]([]int32)` is still a
  compile error.

### DIFF
--- a/test/Core.Tests/CodeAnalysis/Emit/Issue3685CovariantArrayUpcastTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Emit/Issue3685CovariantArrayUpcastTests.cs
@@ -75,6 +75,49 @@ func Widen(dirs []DirectoryInfo) []FileSystemInfo {
     }
 
     [Fact]
+    public void CovariantUpcast_StoringAWrongElement_ThrowsArrayTypeMismatchException()
+    {
+        // Issue #3858: the covariant upcast is UNSOUND BY DESIGN, exactly as in
+        // C# — the widened static type admits a store the runtime array rejects,
+        // and the CLR throws ArrayTypeMismatchException. That hazard is why the
+        // conversion stays explicit-only (#2516); this is the executing witness
+        // that `cast[[]Base](derived)` preserves the CLR's element check rather
+        // than quietly producing a `Base[]` that would accept anything.
+        var result = EmittedOracle.Evaluate(@"
+import System
+import System.IO
+let dirs = []DirectoryInfo{ DirectoryInfo(""alpha"") }
+let widened = cast[[]FileSystemInfo](dirs)
+var caught = ""none""
+try {
+    widened[0] = FileInfo(""beta"")
+} catch (e ArrayTypeMismatchException) {
+    caught = ""ArrayTypeMismatchException""
+}
+caught
+");
+        Assert.Empty(result.Diagnostics);
+        Assert.Equal("ArrayTypeMismatchException", result.Value);
+    }
+
+    [Fact]
+    public void ValueElementUpcast_IsStillRejected()
+    {
+        // Anti-vacuity: C# array covariance is REFERENCE-element-only. `int[]`
+        // does NOT convert to `object[]` in C# — implicitly or explicitly — and
+        // the CLR has no such array relationship (the elements are 4-byte
+        // integers, not references), so `[]int32 -> []object` must stay an
+        // error rather than becoming a cast that corrupts memory.
+        var result = EmittedOracle.Evaluate(@"
+func Widen(values []int32) []object {
+    return cast[[]object](values)
+}
+0
+");
+        Assert.NotEmpty(result.Diagnostics);
+    }
+
+    [Fact]
     public void UnrelatedElementTypes_AreStillRejected()
     {
         // The new arm requires an IMPLICIT element conversion, so it widens the
--- a/tools/cs2gs/Cs2Gs.Tests/Issue3858CovariantArrayObjectInitializerTranslationTests.cs
+++ b/tools/cs2gs/Cs2Gs.Tests/Issue3858CovariantArrayObjectInitializerTranslationTests.cs
@@ -1,0 +1,204 @@
+// <copyright file="Issue3858CovariantArrayObjectInitializerTranslationTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using Cs2Gs.CodeModel.Printing;
+using Cs2Gs.CodeModel.RoundTrip;
+using Cs2Gs.Translator;
+using Cs2Gs.Translator.Loading;
+using Xunit;
+
+namespace Cs2Gs.Tests;
+
+/// <summary>
+/// Issue #3858 (the <c>test/Sdk.Tests</c> self-migration wall): a C#
+/// object-initializer member is an ASSIGNMENT position, so C# applies the same
+/// implicit array-covariance conversion there it applies to a plain assignment
+/// (<c>Compile = new[] { new TaskItem("Program.gs") }</c> into an
+/// <c>ITaskItem[]</c> property). Issue #3685 taught the translator to spell that
+/// upcast as <c>cast[[]Base](expr)</c> — G# slices are invariant by design
+/// (#2516) — but only at the argument / return / local-initializer /
+/// assignment-statement positions; the two object-initializer member builders
+/// dropped the conversion, and gsc reported GS0156 on every site.
+/// </summary>
+public class Issue3858CovariantArrayObjectInitializerTranslationTests
+{
+    [Fact]
+    public void CovariantArray_ObjectInitializerMember_EmitsExplicitCast()
+    {
+        // The wall's exact shape, with `IComparable`/`string` standing in for
+        // `ITaskItem`/`TaskItem`: a DERIVED-element array assigned to a member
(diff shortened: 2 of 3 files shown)


DavidObando:
## Measured before/after (targeted closure run)

Targeted self-migration run over the exact `test/Sdk.Tests` dependency closure, full pipeline (translate + compile + ILVerify + test-parity).

**Before** — `c9943bee3` (`origin/main`), gsc `0.4.399+gc9943bee32`:

| app | result |
| --- | --- |
| src/Analyzers/GSharp.CodeAnalysis.Analyzers.Testing | green |
| src/Analyzers/InternalAnalyzers | green |
| src/Compiler | green |
| src/Core | green |
| src/Sdk/Gsharp.HotReload.Runtime | green |
| src/Sdk/Gsharp.NET.Sdk | green |
| **test/Sdk.Tests** | **RED — compile** |

`test/Sdk.Tests` failed at the **compile** stage with **15** GS0156 sites across 5 files, all one message, all the object-initializer shape:

```
BuildTaskDesignTimeTests.gs           7
Adr0169GsAnalyzerForwardingTests.gs   3
Adr0169EditorConfigSeverityTests.gs   2
Issue3782WarningsNotAsErrorsTests.gs  2
HotReloadArtifactsTaskTests.gs        1
```

(The gate's "7 fingerprints" is a deduplicated signature count, not the site count — worth knowing when sizing this family from a gate report.)

**After** — this branch, gsc `0.4.400+f8aa04eccc`:

```
translate   passed
compile     passed     <- 15 -> 0 GS0156
ilverify    passed
test-parity FAILED
```

**The compile wall is fully cleared — zero GS0156 remain, and ILVerify passes on the emitted IL.** All 15 sites shared one root cause, as suspected.

`test/Sdk.Tests` is **not yet green**: it now reaches test-parity (a stage it had never got to) and fails there with 52 failures / 34 passes. That is a genuinely new layer, not a regression from this PR — filed as #3862 with the family breakdown (37 `RepoRoot` `GSharp.sln` lookups, 12 NREs inside the migrated `Gsharp.NET.Sdk`, 3 others), including two hypotheses already ruled out by direct probe.

No change to `selfmig-baseline.json` — the gate floor is untouched by this PR.

Issue #3880, opened 2026-09-03
selfmig: triage of the migrated tools/cs2gs/Cs2Gs.Tests compile wall — 15 sites, 7 root causes, 3 masking shapes


DavidObando:
Triage of the largest remaining self-migration compile wall. Measured on gate run **33808173710** (gsc `0.4.405+fb3c60ab01`, PASSED at 44/52), re-verified against `origin/main` `2ddcff144` with a locally built `gsc`.

Supersedes the "15 distinct errors across 6 files" sketch in #3836 with a root-caused breakdown.

## Site count vs fingerprint count

The gate reports **11 fingerprints**; the deduplicated `sdk.build.log` lines are **15 real sites across 9 files**. Fingerprints deduplicate the *diagnostic signature*, not the location.

| file | line | id | message |
|---|---|---|---|
| Adr0169AnalyzerParityTests.gs | 130 | GS0159 | Cannot find function WithAnalyzers |
| Adr0169AnalyzerParityTests.gs | 131 | GS0159 | Cannot find function Create |
| Adr0169AnalyzerParityTests.gs | 133 | GS0159 | Cannot find function GetAnalyzerDiagnosticsAsync |
| Adr0169Gsa0005ParityTests.gs | 269 | GS0159 | Cannot find function WithAnalyzers |
| Adr0169Gsa0005ParityTests.gs | 269 | GS0159 | Cannot find function Create |
| Adr0169AnalyzerTranslationTests.gs | 788 | GS0266 | ambiguous `AssertBindsAgainstGsCore` |
| Adr0169AnalyzerTranslationTests.gs | 820 | GS0266 | ambiguous `AssertBindsAgainstGsCore` |
| Coverage/PrinterExhaustivenessTests.gs | 57 | GS0158 | Cannot find member All |
| Issue1756Tests.gs | 163 | GS0159 | Cannot find function ConfigureAwait |
| Issue2833RecordToStringParityTests.gs | 28 | GS0274 | `nil` → non-nullable `name` |
| Issue2867TestRunClassificationTests.gs | 22 | GS0274 | `nil` → non-nullable `output` |
| Issue3461IdentifierSanitizationTests.gs | 504 | GS0113 | Type `ImportedVisible.defer_` doesn't exist |
| Issue3461IdentifierSanitizationTests.gs | 506 | GS0159 | Cannot find function Append |
| Issue3676GeneratedCodeNullabilityTests.gs | 134 | GS0246 | named argument `sources` |
[...]

Issue #3874, opened 2026-09-03
gsc: overload resolution erases CLR delegate identity after generic substitution (Predicate[T] vs Func[T, bool] → GS0266)


DavidObando:
Found while re-diagnosing #3841 (see [that comment](https://github.com/DavidObando/gsharp/issues/3841#issuecomment-5532685577) for why #3841 itself turned out to be a cs2gs defect, not this).

gsc keeps two distinct CLR delegate types distinct when they are spelled by name — an overload set on `Predicate[T]` vs `Func[T, bool]` **declares** without GS0264, and the non-generic case dispatches correctly at run time. But once the parameter types are **substituted through the receiver's type argument**, overload resolution erases the delegates to their shared structural function type and the call becomes ambiguous.

## Repro (measured on `origin/main` @ `fb3c60ab0`)

Ambiguous — the delegate parameter types mention the class's own type parameter:

```gs
package Demo
import System

class ClrOverloads[T] {
    func Add(cb Predicate[T]) string { return "pred" }
    func Add(cb Func[T, bool]) string { return "func" }
}

func always(v int32) bool { return true }

let p Predicate[int32] = always
let f Func[int32, bool] = always
let c = ClrOverloads[int32]()
Console.WriteLine(c.Add(p))
Console.WriteLine(c.Add(f))
```

```
error GS0266: Call to 'Add' is ambiguous between multiple overloads. Disambiguate with explicit types or named arguments.  (both call sites)
```

Both **controls** compile, run, and print `pred` / `func`:

1. the same overload set on a **non-generic** class (`Predicate[int32]` / `Func[int32, bool]`);
2. the same **closed** parameter types inside a **generic** class (`class G[T] { func Add(cb Predicate[int32]) …; func Add(cb Func[int32, bool]) … }`) — so it is not the generic declaring type by itself.

So the erasure is specifically in applicability checking *after* generic substitution: the substituted `Predicate[int32]` and `Func[int32, bool]` compare equal where the unsubstituted spellings do not.

## Expected

C# resolves this by nominal delegate identity in all four combinations. G# already does in three of them; the substituted one should match.

## Notes

- The declaration side is fine: `ClrOverloads[T]` above (and the same shape with two `init`s) declares with no GS0264, and the constructor calls resolve.
[...]

Issue #3836, opened 2026-09-02
selfmig: recently-added Cs2Gs.Tests sources do not survive their own migration (sibling sweep from #3831)


DavidObando:
Found while fixing #3831 (PR #3835), doing the "are there more of these?" sweep the same hazard demands.

## The hazard, restated

CI **compiles** the C# sources; the self-migration gate **translates then compiles** them. So a source file added to `tools/cs2gs/**` can be fully CI-green and still not survive its own migration — visible only in the nightly gate. #3831 was one instance (a production file, 5 banked apps down). The sweep found more, all in the **non-banked** `tools/cs2gs/Cs2Gs.Tests`, which is why they cost nothing today and are easy to leave rotting.

## What does not migrate

From the full local gate run on this branch (`a7f344324`, 43/52 green), `tools/cs2gs/Cs2Gs.Tests` fails `compile` with 15 distinct errors across 6 files. Two of the six are recent additions:

| file | added | migrated error |
|---|---|---|
| `Adr0169Gsa0005ParityTests.cs` | 2026-09-01, #3809 (`f0aeb953d`) | `GS0159: Cannot find function WithAnalyzers` / `Cannot find function Create` |
| `Issue3676GeneratedCodeNullabilityTests.cs` | 2026-08-30, #3676 (`fd45e79e2`) | `GS0246: Named argument 'sources' does not match any parameter of 'TranslateEnabled'` |

The remaining four are older and pre-date this framing: `Adr0169AnalyzerParityTests.cs` (same `WithAnalyzers`/`Create` shape — the new file inherits the gap from the shared helper), `Adr0169AnalyzerTranslationTests.cs` (`GS0266` ambiguous overload), `Coverage/PrinterExhaustivenessTests.cs` (`GS0158 Cannot find member All`), `Issue3461IdentifierSanitizationTests.cs` (`GS0113`/`GS0159`), plus `Issue1756Tests.cs`, `Issue2833RecordToStringParityTests.cs`, `Issue2867TestRunClassificationTests.cs`, `NuGetAuditAdvisoryPolicyTests.cs` (`GS0274` `nil` at a non-nullable parameter).

## Why this is worth an issue rather than a shrug

[...]

Issue #3726, opened 2026-08-31
cs2gs: oblivious taint does not flow from an [InlineData(null, …)] argument to the theory parameter (#3724 family C)


DavidObando:
Family **C** of the [#3724](https://github.com/DavidObando/gsharp/issues/3724) `test/LanguageServer.Tests` triage (#3501 Track C). 2 of the app's 14 compile errors.

```
GS0274 test/LanguageServer.Tests/ProgramTests.gs:18  'nil' cannot be assigned to parameter 'tracePath' of non-nullable type 'string'; consider changing the parameter type to 'string?'.
GS0274 test/LanguageServer.Tests/ProgramTests.gs:19  'nil' cannot be assigned to parameter 'expected' of non-nullable type 'string'; consider changing the parameter type to 'string?'.
```

C# source (`test/LanguageServer.Tests/ProgramTests.cs`, a file with no NRT annotations — the parameters are **oblivious**):

```csharp
[Theory]
[InlineData(null, null)]
[InlineData("", null)]
[InlineData("  protocol.log  ", "protocol.log")]
public void GetLogPath_UsesOptInProtocolTraceEnvironment(
    string tracePath,
    string expected)
{
    Assert.Equal(expected, Program.GetLogPath(System.Array.Empty<string>(), tracePath));
}
```

Emitted G#:

```
@Theory
@InlineData(nil, nil)
@InlineData("", nil)
@InlineData("  protocol.log  ", "protocol.log")
func GetLogPath_UsesOptInProtocolTraceEnvironment(tracePath string, expected string) {
    Assert.Equal(expected, Program.GetLogPath(Array.Empty[string](), tracePath))
}
```

The parameters' **only** null evidence is an attribute argument, and `ObliviousNullabilityAnalyzer` does not treat attribute arguments as a taint source — so it left them `string` while the emitted attribute passes `nil`, and `gsc` correctly rejects the mismatch. The internal inconsistency is the bug: cs2gs emitted both sides.

## Suggested fix

Add an attribute-argument taint source: a `null` argument in a data-driving attribute on a method taints the positionally-matching parameter of that method. The mechanical form (any attribute whose constructor takes `params object[]`, positionally aligned with the decorated method's parameters) covers `[InlineData]` without hard-coding xunit; `[MemberData]`/`[ClassData]` do not carry literal nulls and need nothing. The taint then propagates through the existing edge machinery, so `tracePath`/`expected` come out `string?`.

[...]

PR #3881, opened 2026-09-04
gsc + cs2gs: clear 6 of the 15 migrated Cs2Gs.Tests compile errors (#3880)

DavidObando:
Closes part of #3880 (full triage table there). Part of #3501, sibling of #3836.

## The wall

The gate reported **11 fingerprints** for the migrated `tools/cs2gs/Cs2Gs.Tests`; the deduplicated build log has **15 real sites across 9 files**, and those 15 collapse to **7 root causes**. Three of them are fixed here (6 sites); the other four are filed on #3880 with minimal repros.

Three of the seven causes were **masking shapes** — the printed diagnostic named something other than the real error. The six "cannot find X" errors do *not* share one mechanism; they span four causes.

### 1. gsc: a named argument may address the trailing variadic parameter (1 site, GS0246)

`ValidateExpandedVariadicNamedArguments` accepted names only in the FIXED prefix of a variadic parameter list, so

```gs
TranslateEnabled(roundTripOnlyReason: "…", sources: ("Uri.cs", "…"))
```

reported `GS0246: named argument 'sources' does not match any parameter` — for the exact spelling C# accepts for `params`, and therefore the exact spelling cs2gs emits.

The argument already sits in the variadic parameter's own slot, so nothing needs reordering; the value is either the whole carrier or a single expanded element, which the existing `PackOrPassThroughVariadicArguments` decides on by type. The rule is narrowed to what C# allows, not disabled: the name must be the variadic parameter's, in its own slot, and **last**.

### 2. gsc: a fully qualified type resolves even when two imported packages collide (2 sites, GS0113 + one GS0159 cascade)

`typeof(ImportedVisible.defer_)` reported `Type 'ImportedVisible.defer_' doesn't exist` — for a type that *did* exist, in a file that *was* compiled. The qualified type-clause path falls back to a simple-name lookup of the final segment, and #2455 makes that lookup refuse to guess when two separately-imported packages declare the same simple name. Here `import $class` and `import ImportedVisible` both declare `defer_`.

But a qualified reference is not ambiguous: its qualifier names the package outright. The fix publishes that qualifier as the explicit package hint #2455 **already honours** for qualified construction (`Pkg.Type{…}`), scoped to the single lookup and restored in a `finally`. Deleting either import made the original compile, which is how the ten-line repro was bisected out of the real 434-file compilation.

The `GS0159 Cannot find function Append` on the next line was a cascade of this: `fixtureAssembly` was error-typed, so the argument to `.Append(…)` had no conversion.

### 3. cs2gs: `[InlineData(null)]` promotes the parameter it lines up with (3 sites, GS0274)

#3726 made a `null` in a data-row attribute promote the matching oblivious parameter, but deliberately excluded the whole-array form `[InlineData(null)]` on the grounds that C# binds the lone `null` as the params *array* and it therefore "names no position".

gsc does not agree, and gsc is the other half of this toolchain: its GS0274 check (#660) reads the **emitted** `@InlineData(nil)` positionally and rejects a non-nullable parameter 0 — which is also what xunit does at run time for the single-column theories this spelling is used on. So cs2gs was emitting a file that contradicted itself, which is the precise inconsistency #3726 set out to remove. The claim is kept as narrow as gsc's: position 0 only.

## Test evidence

Everything below **executes** — `EmittedOracle.Evaluate` compiles, emits, runs and returns the value, so a fix that binds cleanly but computes the wrong thing still fails.

| suite | tests | on this branch | on `origin/main` |
|---|---|---|---|
| `Issue3880VariadicNamedArgumentTests` | 5 | 5 pass | **3 fail**, 2 pass |
| `Issue3880QualifiedTypeAcrossCollidingImportsTests` | 4 | 4 pass | **3 fail**, 1 pass |
| `Issue3726InlineDataNullArgumentTaintTests` | 6 (2 new) | 6 pass | **2 fail**, 4 pass |

The tests that pass on `origin/main` are the deliberate anti-vacuity guard rails, and each fix has at least one:

- a named argument that matches no parameter still reports GS0246;
- naming the variadic parameter and *then* continuing positionally still reports GS0246 (so the rule is narrowed, not removed);
- an **unqualified** colliding name still fails — #2455's ambiguity rule is untouched;
- the *other* qualifier selects the *other* package, so the hint genuinely selects rather than merely unblocking (the reflected `Namespace` is the discriminator — resolving to the wrong package would bind just as cleanly);
- a `[InlineData(null)]` whole-array row still claims nothing beyond parameter 0.

The `[InlineData(null)]` test also asserts `Normalize(name)` is emitted **without** a `!!` bridge: the row really does pass nil, so bridging the read would convert a passing test into a runtime throw (#3704's shape).

Full `test/Core.Tests` was run locally on the final tree.

## No diagnostic was weakened

Each fix makes gsc accept code that is legal C# and legal G#, and each is paired with a test proving the neighbouring illegal spelling is still rejected. Nothing was suppressed or downgraded.

## Migration hazard (#3831 / #3836)

`ObliviousNullabilityAnalyzer.cs` lives in `Cs2Gs.Translator`, which is inside its own migration corpus and is a **banked green app** — so a change that CI compiles can still break the gate by existing. The edit adds only an `if (row.IsNull)` branch and a `parameter.Ordinal == 0` test, both of which already appear verbatim in the same method a few lines away, in a file that migrates green today. I was not able to run the `Cs2Gs.Translator → Cs2Gs.CodeModel → src/Core` closure migration to confirm this directly: another agent is building on the same machine, and a concurrent `dotnet build` swaps the newest `Gsharp.NET.Sdk` nupkg under a running migration and produces `MSB4236` failures that look like real regressions. Worth a closure run before merge if the machine is quiet.

## Expected gate effect

15 sites → **9**. `Cs2Gs.Tests` is not in `greenApps` and stays red on the four remaining causes, so the floor of 44 is unaffected; `selfmig-baseline.json` is untouched.

## Still open, on #3880

| cause | sites | why not here |
|---|---|---|
| ADR-0169 analyzer retarget: the migrated `InternalAnalyzers` are `GSharpDiagnosticAnalyzer`, so the Roslyn half of the parity tests cannot bind | 5 | needs a design call — `Cs2Gs.Tests` cannot both migrate itself and keep a Roslyn control for the analyzer-parity diff |
| `X.Member[i](args)` parses as a generic instantiation, never as indexer-then-invoke | 1 | grammar work |
| gsc erases the Task from an `async func Main` — any static `Main`, any target, including `/target:library` (metadata proof on the issue) | 1 | entry-point emit; csc's answer is to keep `Main` as `Task<int>` and synthesize a separate sync entry point |
| overload resolution has no "normal form beats expanded form" tie-break | 2 | Phase-2 scoring change with corpus-wide blast radius; explicitly **not** #3874 |

### DIFF
--- a/src/Core/CodeAnalysis/Binding/Binder.cs
+++ b/src/Core/CodeAnalysis/Binding/Binder.cs
@@ -4496,7 +4496,28 @@ public sealed class Binder
         // cs2gs fully-qualifies type references (including generic-math
         // constraints), so this is the common shape for translated code.
         var lastSegment = segmentTexts[totalSegments - 1];
-        var sourceType = LookupType(lastSegment, targetArity > 0 ? targetArity : -1);
+
+        // Issue #3880: the simple-name fallback below refuses to guess when two
+        // or more separately-imported packages declare the same simple name
+        // (#2455's ambiguity rule) — but here the reference is not ambiguous at
+        // all: its qualifier names the package outright. Publish that qualifier
+        // as the explicit package hint #2455 already honours for qualified
+        // CONSTRUCTION (`Pkg.Type{…}`), so `typeof(ImportedVisible.defer_)`
+        // resolves even when a second imported package also declares `defer_`.
+        // The hint is restored unconditionally: it is scoped to this one
+        // lookup, exactly as at the construction site.
+        var qualifierPackageHint = string.Join(".", segmentTexts, 0, totalSegments - 1);
+        var previousQualifierHint = scope.SetQualifiedConstructionPackageHint(qualifierPackageHint);
+        TypeSymbol? sourceType;
+        try
+        {
+            sourceType = LookupType(lastSegment, targetArity > 0 ? targetArity : -1);
+        }
+        finally
+        {
+            scope.SetQualifiedConstructionPackageHint(previousQualifierHint);
+        }
+
         if (sourceType != null && !ReferenceEquals(sourceType, TypeSymbol.Error))
         {
             if (targetArity == 0)
--- a/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.Arguments.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.Arguments.cs
@@ -1117,6 +1117,28 @@ internal sealed partial class OverloadResolver
                 continue;
             }
 
+            // A named argument may also address the trailing variadic parameter
+            // itself, exactly as C# permits for `params`: the name must be the
+            // variadic parameter's, it must sit in that parameter's own slot,
+            // and it must be the last argument (naming the variadic parameter
+            // and then continuing to supply elements positionally is not a
+            // spelling C# accepts either). Because the argument is already in
+            // its own slot no reordering is needed; the value that follows is
+            // either the whole carrier or a single expanded element, which
+            // PackOrPassThroughVariadicArguments then decides on by type.
+            var namesTrailingVariadic =
+                sourceIndex == fixedParameterCount &&
+                sourceIndex == argumentNames.Length - 1 &&
+                string.Equals(
+                    parameterNameAt(sourceIndex),
+                    name,
+                    StringComparison.Ordinal);
+
+            if (namesTrailingVariadic)
+            {
+                continue;
+            }
+
             if (sourceIndex >= fixedParameterCount ||
                 !string.Equals(
                     parameterNameAt(sourceIndex),
--- a/test/Core.Tests/CodeAnalysis/Binding/Issue3880QualifiedTypeAcrossCollidingImportsTests.cs
+++ b/test/Core.Tests/CodeAnalysis/Binding/Issue3880QualifiedTypeAcrossCollidingImportsTests.cs
@@ -1,0 +1,115 @@
+// <copyright file="Issue3880QualifiedTypeAcrossCollidingImportsTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using GSharp.Tests;
+using Xunit;
+
+namespace GSharp.Core.Tests.CodeAnalysis.Binding;
+
+/// <summary>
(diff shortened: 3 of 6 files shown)


DavidObando:
## Closure migration run: **CLEAN** — the caveat in the PR body is discharged

Ran the `Cs2Gs.Translator → Cs2Gs.CodeModel → src/Core` closure as asked, with the closure **kept whole** — nothing a kept app references was excluded, so there are no phantom cascades. `src/Analyzers/InternalAnalyzers` is in the set because `src/Core` references it (analyzer-only, `ReferenceOutputAssembly=false`, still needed for build ordering). 55 of the 59 discovered projects excluded, 4 kept.

```
cs2gs migrate — run 2026-09-04T00-29-09Z_3c60e3
gsc: 0.4.407+0181eabb47

app                                                       translate  compile  ilverify  test-parity
src/Analyzers/InternalAnalyzers/InternalAnalyzers.csproj  PASS       PASS     PASS      PASS
src/Core/Core.csproj                                      PASS       PASS     PASS      PASS
tools/cs2gs/Cs2Gs.CodeModel/Cs2Gs.CodeModel.csproj        PASS       PASS     PASS      PASS
tools/cs2gs/Cs2Gs.Translator/Cs2Gs.Translator.csproj      PASS       PASS     PASS      PASS

4/4 apps green; run PASSED.
```

It reached **all four stages**, not just compile — translate, compile, ILVerify and test-parity all pass.

Two things make this a real check rather than a re-run of CI:

**The gsc and cs2gs under test are the ones from this branch.** The Release repack stamped `0.4.407-g0181eabb47`, matching HEAD `0181eabb4`, so the migration used this branch's compiler and translator rather than silently reusing an older packed SDK. Prerequisites were built in the documented order (Debug `Compiler` → Debug `Gsharp.NET.Sdk` → Debug ref assemblies for the closure → Release repack → migrate), and no `MSB4236` moniker churn appeared.

**The edited method really did translate**, and translated correctly — the new branch is present verbatim in the emitted G#:

```gs
if row.IsNull {
    if parameter.Ordinal == 0 {
        return true
    }
    continue
}
```

So the prior I offered is now evidence: the change survives its own migration. #3881 is ready from my side.

Issue #2835, opened 2026-07-28
cs2gs erases source-declared delegate types to structural arrow form, breaking nominal CLR delegate identity


DavidObando:
## Summary

`cs2gs` erases **every** delegate type to G#'s structural arrow form (`(string) -> void`), including delegate types **declared in the source being translated**. For BCL `Func`/`Action` that is correct and idiomatic (ADR-0115 §B.8), but for a user-declared delegate it destroys nominal identity — and delegates are nominally typed in the CLR, so structurally equivalent delegates are **not** interchangeable at runtime.

`CSharpTypeMapper.MapEventType` already recognises this for the *event's own* handler type ("Event metadata and add/remove signatures are ABI-sensitive: structurally equivalent delegates are not interchangeable in the CLR"), but every other type position — fields, locals, parameters, generic arguments — still erases.

## Repro

`tools/cs2gs/corpus/grid/G07-Members-Console/Constructs/EventDeclaration.cs`:

```csharp
public delegate void MessageHandler(string message);

public class Broadcaster
{
    private readonly List<MessageHandler> _handlers = new List<MessageHandler>();

    public event MessageHandler Message
    {
        add { _handlers.Add(value); }
        remove { _handlers.Remove(value); }
    }
    ...
}
```

translates to:

```gs
type MessageHandler = delegate func(message string)

class Broadcaster {
    private let _handlers List[(string) -> void] = List[(string) -> void]()   // <-- erased

    event Message MessageHandler {      // <-- nominal (MapEventType)
        add { _handlers.Add(value) }    // value : MessageHandler
        remove { _handlers.Remove(value) }
    }
}
```

The declaration `type MessageHandler = …` **is** emitted, and the event keeps its nominal type, but the backing field became `List[Action[string]]`. Result at runtime:

```
Unhandled exception. System.ArgumentException: The value "Corpus.Grid07.MessageHandler" is not of
type "System.Action`1[System.String]" and cannot be used in this generic collection. (Parameter 'value')
   at System.Collections.Generic.List`1.System.Collections.IList.Add(Object item)
   at Corpus.Grid07.Broadcaster.add_Message(MessageHandler value) in .../Constructs/EventDeclaration.gs:line 13
```

[...]

Issue #3839, opened 2026-09-03
cs2gs: 'ref' is dropped from expression-bodied ref-returning members (GS0252, and a silent fidelity loss on ref props/indexers)


DavidObando:
Split out of #3837 (RC4), 1 of the 38 `test/Interpreter.Tests` self-migration compile fingerprints — but the erroring line is the *less* serious half.

Source: `test/Interpreter.Tests/Issue3004ManagedByRefAutoDereferenceFixture.cs`. Migrated output:

```gs
class ManagedByRefAutoDereferenceFixture {
    private let values []int32 = []int32{40, 41, 42}

    prop Property int32 -> values[0]                    // C#: public ref int Property => ref values[0];
    prop this[index int32] int32 -> values[index]       // C#: public ref int this[int index] => ref values[index];
    func GetValue(index int32) ref int32 -> values[index]
    //                                      ^ error GS0252: Function 'GetValue' returns by reference;
    //                                        use 'return ref <lvalue>' instead of a plain 'return'.
}
```

Two defects:

1. **The error.** An expression-bodied `ref`-returning `func` keeps its `ref` return type but the returned expression is not emitted as `ref values[index]`.
2. **The silent one.** The `ref` property and `ref` indexer lost their `ref` **entirely** — they now return a copy. This does not error; it changes behaviour, and the fixture exists precisely to guard managed-byref auto-dereference. This is the more dangerous half: it would have survived into a green migration.

Fix needs an executing test (mutate through the returned reference and observe the mutation), not a printed-shape assertion — a shape test would pass on the copy-returning form.

Issue #1904, opened 2026-07-03
cs2gs: `async Task Main` produces a Task-returning entry point the CLR rejects — silent divergence


DavidObando:
An `async Task Main()` app lowers to G# top-level `await` statements; gsc emits a Task-returning entry point and the CLR aborts with `MethodAccessException: Entry point must have a return type of void, integer, or unsigned integer` — stage 4 sees empty stdout. Workaround: sync Main + `GetAwaiter().GetResult()`. (grid app G10).

Also note that G# should support a program of this shape:
```
package Test

import System

class Program {
  async func Main() {
    ...
  }
}
```

1. Ensure the G# compiler understands an asynchronous entry point and emits verifiable IL
2. Ensure that cs2gs takes `async Task main()` and translates to that G# equivalent.

## Definition of done
- [ ] Fix + `Issue<N>*.cs` regression test
- [ ] Re-enable the quarantined grid fixture (or add one) and re-capture its baseline
- [ ] `tools/cs2gs/coverage/csharp-construct-inventory.json` row updated

_Discovered by the ADR-0138 conformance-grid build-out (gsc 0.2.585+01ad6ec964); quarantined sources live next to each grid app under `tools/cs2gs/corpus/grid/`._

PR #3889, opened 2026-09-04
cs2gs: nominal delegate identity for colliding overload sets (#3841); gsc: an async Main keeps its Task in metadata (#3883)

### LINKED ISSUE #3841: G#: two distinct CLR delegate types erase to one function type, turning a legal C# overload set into GS0264
Split out of #3837 (RC3), 2 of the 38 `test/Interpreter.Tests` self-migration compile fingerprints. Unlike the other splits this is an **ADR-level design question**, not a patch.

Migrated `test/Interpreter.Tests/Issue2918InlineLambdaErasedReceiverEmittedOracleTests.gs`:

```gs
class Issue2948MethodDelegateOverloads[T] {
    func Add(callback (T) -> bool) string -> "pred"   // C#: Add(Predicate<T>)
    func Add(callback (T) -> bool) string -> "func"   // C#: Add(Func<T, bool>)
    //   ^ error GS0264: An overload of 'Add' with signature 'Add((T) -> bool)' is already declared.
}

class Issue2948ConstructorDelegateOverloads[T] {
    init(callback (T) -> bool) { ... }   // C#: .ctor(Predicate<T>)
    init(callback (T) -> bool) { ... }   // C#: .ctor(Func<T, bool>)
    //  ^ error GS0264 (same cause, .ctor shape)
}
```

`Predicate<T>` and `Func<T, bool>` are different CLR types with different metadata identity; G#'s structural function type `(T) -> bool` erases the difference, so a C# API that overloads on delegate identity has no G# spelling. Note the irony: the fixture under translation is *itself* a named-delegate-identity test (its sibling #3149 fixture is #3838).

Options, none obviously right:
- reify named delegates as distinct types at declaration sites (cf. #3149) so the overload set survives;
- give cs2gs a canonical disambiguation for a colliding delegate overload set;
- accept the erasure and declare such C# APIs untranslatable, with an explicit translation diagnostic instead of a downstream GS0264.

Wants an ADR, and it interacts with the shape-not-identity coupling rule used for tuples.

### LINKED ISSUE #3883: gsc: an `async func Main` loses its Task in metadata — any static Main, any target, including /target:library
Found while triaging the migrated `tools/cs2gs/Cs2Gs.Tests` compile wall (#3880, cause 4), but this is **not confined to the migration corpus** — any G# program with a static method named `Main` that is `async` is affected, in any assembly, at any `/target`.

## The defect

`Binder.ResolveEntryPoint` (#1996) treats a static `Main` in **any** class as an entry-point candidate. That is fine for choosing an entry point, but the emitted signature is then rewritten to the CLR entry-point shape and the `Task` is **erased from metadata**. Consumers in another assembly see a synchronous method.

Reproducing, compiled with a locally built `gsc` on `2ddcff144`:

```gs
// lib3.gs — note: compiled with /target:library, so this is not even an entry point
package Lib3

import System
import System.Threading.Tasks

class Q {
    shared {
        public async func Main() int32 {
            await Task.Yield()
            return 1
        }

        public async func GetString() string {
            await Task.Yield()
            return "x"
        }
    }
}
```

```
gsc lib3.gs /target:library /out:lib3.dll
```

Reflecting over the emitted assembly:

```
lib3.dll Lib3.Q.Main      -> System.Int32
lib3.dll Lib3.Q.GetString -> System.Threading.Tasks.Task`1[System.String]
```

`GetString` correctly emits `Task<string>`. `Main` emits bare `int32`. **The method name is the only difference** — rename `Main` to anything else and it emits `Task<int32>`.

## How it surfaces

A consumer in another assembly cannot treat the call as a task:

```gs
package main

import Lib3

async func A() int32 {
    return await Q.Main().ConfigureAwait(false)
}
```

```
error GS0159: Cannot find function ConfigureAwait.
```

That message is a **masking shape**: `ConfigureAwait` exists on `Task<int>` and `Q.Main()` should have been one. The real problem — a rewritten signature — is never named. Same-assembly calls are unaffected, because binding uses the source symbol rather than the metadata one; the defect only becomes visible across an assembly boundary, which is why it can sit unnoticed.

## Why it matters outside the migration

This is the ordinary "an integration test drives the app's own `Main`" pattern:

```csharp
exitCode = await Cs2Gs.Cli.Program.Main(args).ConfigureAwait(false);
```

That is legal C# — Roslyn keeps the user's `async Task<int> Main` as `Task<int>` in metadata and synthesizes a **separate** synchronous entry point (`<Main>$`) that blocks on it. gsc instead rewrites `Main` itself, so the pattern is unavailable in G#. Anyone writing an integration test against their own console app, or exposing an async `Main` as a reusable library entry point, hits this.

The migration instance is `tools/cs2gs/Cs2Gs.Tests/Issue1756Tests.cs` line 149.

## Suggested shape of a fix

Follow csc: keep the authored `Main` with its `Task`-returning signature in metadata, and synthesize a distinct entry-point stub that blocks on it. Two secondary questions worth settling in the same change:

1. `ResolveEntryPoint` scanning for a static `Main` at all under `/target:library` — there is no entry point to resolve for a library, so the candidate scan (and any signature consequence of it) should not run.
2. Whether the erasure also mis-shapes the *sync* path — a caller that does `let t = Q.Main()` gets an `int32` where the source says the call is awaitable, so the two sides of the same declaration disagree.

## Tests

A regression test must **execute**, not merely bind: the same-assembly path binds clean today and hides the whole defect. The discriminating assertion is a cross-assembly `await X.Main().ConfigureAwait(false)` that runs and returns the right value, plus a reflection check that the emitted return type is `Task<T>`.

Part of #3501 (found there), but the defect is general.

DavidObando:
Fixes #3841. Fixes #3883.

Two defects that both affect ordinary code, not only the migration corpus. They
share no root cause — one is cs2gs type mapping, the other gsc entry-point
emit — but both were found behind masking diagnostics on #3501.

## #3841 — cs2gs manufactured the duplicate signature (no ADR needed)

The issue was filed as a G# type-system gap. It is not one, and the
re-diagnosis on the issue is confirmed: gsc already declares AND dispatches a
`Predicate[int32]` / `Func[int32, bool]` overload set correctly.

`CSharpTypeMapper` applied #2835's nominal-delegate-identity rule only to
SOURCE-declared delegates, so imported ones (`Predicate`, `Func`) still erased
to the arrow form — and a legal C# overload set became two members with one G#
signature, which gsc correctly rejected with `GS0264`.

The fix preserves nominal identity only where erasure is load-bearing: a
per-compilation set of the CONSTRUCTED delegate types that discriminate an
overload set the arrow form would collapse. Every other `Func`/`Action` keeps
ADR-0115 §B.8 arrow form; `NonCollidingDelegates_StillRenderInArrowForm` is the
guard on that scoping.

**Scoping the fix to the DECLARATIONS alone was not enough**, and the executing
test is what caught it: with only the declaration sites repaired, a local
written `Predicate<int> p = Always;` still erased to `(int32) -> bool`, gsc
picked the `Func` member for BOTH calls, and the program compiled, ILVerified
and printed `func func func func` instead of `pred func pred func`. That is
strictly worse than the `GS0264` it replaced. The rule therefore covers every
type position in a compilation that contains a colliding overload set.

gsc still cannot DISPATCH the generic shape (`Add(Predicate[T])` vs
`Add(Func[T, bool])` on a substituted receiver reports `GS0266`); that is the
separately-filed #3874, in applicability after generic substitution, not in the
declaration model. The declaration side — what #3841 is about — is clear.

## #3883 — gsc erased the Task from an `async func Main`

Two independent halves:

1. `Binder.ResolveEntryPoint` scanned for a static `Main` even under
   `/target:library`, and the emitter then rewrote that method's own signature
   to the CLR entry-point shape. A library now resolves no entry point at all.
   (TLS in a library keeps its ADR-0066 D4 recovery shape: GS0285 and a
   synthesized `<Main>$`.)
2. For an executable, the authored `Main` was itself rewritten. It now keeps
   its `Task<T>` and a SEPARATE synthesized `<Main>$` stub carries the CLR
   entry-point shape and drives the task synchronously — csc's shape.

The stub is emitted on the SAME type as the authored `Main`, not on the
package's `<Program>`: its body is a second async kickoff over a state machine
that is a private nested type of that class, and a `<Program>`-hosted stub dies
at run time with `FieldAccessException`. That was found by running the program,
not by compiling it.

## Tests

All executing, all cross-assembly where the defect requires it.

- `test/Compiler.Tests/Issue3883AsyncMainMetadataTests.cs` — 6 tests: reflection
  over the emitted metadata plus compile + ILVerify + RUN. The discriminating
  one is `ConsumerInAnotherAssembly_CanAwaitTheLibrarysAsyncMain`; the
  same-assembly path binds clean and hides the defect entirely.
- `tools/cs2gs/Cs2Gs.Tests/Issue3841DelegateOverloadSetTranslationTests.cs` —
  5 tests, including `TranslatedProgram_DispatchesEachDelegateOverloadLikeCSharp`
  (translate → real gsc → run → assert each call reached the overload C# would
  have reached) and the arrow-form scoping guard.

Failing-on-`origin/main` proof and anti-vacuity guard rails are in the PR
discussion.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/Binder.cs
+++ b/src/Core/CodeAnalysis/Binding/Binder.cs
@@ -1523,11 +1523,16 @@ public sealed class Binder
             ?? ResolveEntryPointPackage(packageByTree, globalStatements, functions, packagesInOrder);
 
         // Issue #3883: a library has no entry point to resolve. Scanning for a
-        // static `Main` under /target:library used to pick one anyway, and the
-        // emitter then applied the CLR entry-point signature rewrite to it —
-        // erasing `Task<T>` from an `async func Main` in a DLL that can never
-        // be started as a process.
-        var entryPoint = isLibrary
+        // user-declared static `Main` under /target:library used to pick one
+        // anyway, and the emitter then applied the CLR entry-point signature
+        // rewrite to it — erasing `Task<T>` from an `async func Main` in a DLL
+        // that can never be started as a process.
+        //
+        // TLS in a library is a different case and keeps its existing shape:
+        // ADR-0066 D4 reports GS0285 and then CONTINUES, so the synthesized
+        // `<Main>$` is still produced and downstream consumers see a complete
+        // bound tree (BinderEntryPointTests pins this).
+        var entryPoint = isLibrary && globalStatements.Length == 0
             ? null
             : ResolveEntryPoint(binder, functions, structs, globalStatements, syntaxTrees, entryPointPackage, synthesizedEntryPoint);
 
--- a/src/Core/CodeAnalysis/Binding/Binder.cs
+++ b/src/Core/CodeAnalysis/Binding/Binder.cs
@@ -1521,7 +1521,15 @@ public sealed class Binder
         // synthesized <Main>$ in emit.
         var entryPointPackage = synthesizedEntryPointPackage
             ?? ResolveEntryPointPackage(packageByTree, globalStatements, functions, packagesInOrder);
-        var entryPoint = ResolveEntryPoint(binder, functions, structs, globalStatements, syntaxTrees, entryPointPackage, synthesizedEntryPoint);
+
+        // Issue #3883: a library has no entry point to resolve. Scanning for a
+        // static `Main` under /target:library used to pick one anyway, and the
+        // emitter then applied the CLR entry-point signature rewrite to it —
+        // erasing `Task<T>` from an `async func Main` in a DLL that can never
+        // be started as a process.
+        var entryPoint = isLibrary
+            ? null
+            : ResolveEntryPoint(binder, functions, structs, globalStatements, syntaxTrees, entryPointPackage, synthesizedEntryPoint);
 
         // ADR-0156 Phase 3c (#3176): an interactive submission never runs a
         // user-declared entry point — `func Main()` in a cell is an ordinary
--- a/src/Core/CodeAnalysis/Emit/FunctionEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/FunctionEmitter.cs
@@ -215,7 +215,11 @@ internal sealed class FunctionEmitter
             capturedAsyncStepping);
     }
 
-    internal MethodDefinitionHandle EmitFunction(FunctionSymbol function, BoundBlockStatement body, bool isEntryPoint)
+    internal MethodDefinitionHandle EmitFunction(
+        FunctionSymbol function,
+        BoundBlockStatement body,
+        bool isEntryPoint,
+        bool isSynthesizedEntryPointStub = false)
     {
         // ADR-0086 / issue #727 + ADR-0092 / issue #758: P/Invoke functions
         // skip body emission. Classic @DllImport functions route through the
@@ -235,8 +239,8 @@ internal sealed class FunctionEmitter
         var bodySelection = this.SelectFunctionBody(function, body);
         var bodyEmission = this.EmitFunctionBody(function, bodySelection.Body, bodySelection.AsyncPlan, isEntryPoint);
         var signature = this.EncodeFunctionSignature(function, bodySelection.AsyncPlan, isEntryPoint);
-        var methodName = GetMethodMetadataName(function, isEntryPoint);
-        var methodAttributes = GetMethodAttributes(function, isEntryPoint);
+        var methodName = GetMethodMetadataName(function, isEntryPoint, isSynthesizedEntryPointStub);
+        var methodAttributes = GetMethodAttributes(function, isEntryPoint, isSynthesizedEntryPointStub);
         var parameterMetadata = this.EmitParameterMetadata(function, bodySelection.AsyncPlan);
         var handle = this.EmitMethodDefinition(
             function,
@@ -472,7 +476,10 @@ internal sealed class FunctionEmitter
         return sigBlob;
     }
 
-    private static string GetMethodMetadataName(FunctionSymbol function, bool isEntryPoint)
+    private static string GetMethodMetadataName(
+        FunctionSymbol function,
+        bool isEntryPoint,
+        bool isSynthesizedEntryPointStub = false)
     {
         // Synthesized entry point uses the C#-style mangled name; explicit Main / user funcs keep their source name.
         // ADR-0149: a method declared with an explicit-interface qualifier
@@ -487,7 +494,12 @@ internal sealed class FunctionEmitter
         // no clause of its own, so only the settable ExplicitInterfaceClauseTarget
         // (propagated from the owning PropertySymbol by
         // DeclarationBinder.ResolveExplicitInterfaceClauses) reflects it.
-        var methodName = isEntryPoint && function.Declaration is null
+        // Issue #3883: the synthesized stub that blocks on an authored async
+        // `Main` also takes the `<Main>$` metadata name — it must not collide
+        // with the authored `Main` (which is emitted alongside it, keeping its
+        // Task-returning signature), and csc names its equivalent stub the
+        // same way.
+        var methodName = isEntryPoint && (function.Declaration is null || isSynthesizedEntryPointStub)
             ? "<Main>$"
             : function.ExplicitInterfaceClauseTarget != null
                 ? ExplicitInterfaceMetadataNaming.GetMetadataName(function.Name, function.ExplicitInterfaceClauseTarget)
@@ -496,7 +508,10 @@ internal sealed class FunctionEmitter
         return methodName;
     }
 
-    private static MethodAttributes GetMethodAttributes(FunctionSymbol function, bool isEntryPoint)
+    private static MethodAttributes GetMethodAttributes(
+        FunctionSymbol function,
+        bool isEntryPoint,
+        bool isSynthesizedEntryPointStub = false)
     {
         // The synthesized entry point must remain Public so the runtime can find it.
         // ADR-0149: an explicit-interface qualifier clause member is ALWAYS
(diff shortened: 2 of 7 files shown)


DavidObando:
## Evidence (recorded post-merge)

All measurements below are on base `36a147f0f` (#3882 already merged), both sides of every before/after.

### Failing-on-`origin/main` proof — #3883

Reverting only the three gsc files and rebuilding, the new suite gives **4 failed / 2 passed**:

```
Failed AsyncMainInExecutable_KeepsItsTaskAndGetsASeparateEntryPointStub
Failed AsyncMainInLibrary_KeepsItsTaskInMetadata
Failed ConsumerInAnotherAssembly_CanAwaitTheLibrarysAsyncMain
    Consumer.gs(7,37,7,58): error GS0159: Cannot find function ConfigureAwait.
Failed PackageScopeAsyncMain_KeepsItsTaskAndGetsASeparateEntryPointStub
```

That `GS0159` is verbatim the masking shape in the issue. The two tests that **pass on main and after** are the deliberate anti-vacuity guard rails: `SyncMain_IsUnchanged_NoStubAndNoRename` and `TopLevelAwait_StillRewritesTheSynthesizedEntryPointInPlace` (issue #1904's behaviour, which this change must not disturb).

Reflection, before → after, same source:

```
lib3.dll  Lib3.Q.Main      -> System.Int32                     →  Task`1<Int32>
lib3.dll  Lib3.Q.GetString -> Task`1<String>                   →  unchanged (control)
lib3.dll  entryPointToken  =  0x06000002                       →  0x00000000  (a library has none)
app.dll   App.Program.Main -> System.Int32                     →  Task`1<Int32>
app.dll   App.Program.<Main>$                                  →  new; Int32; is the entry point
```

Cross-assembly run: a consumer assembly does `await Runner.Main().ConfigureAwait(false)` and prints `42`. The same-assembly call binds clean on main and after, which is exactly why this sat unnoticed.

The `<Program>`-hosted stub was tried first and dies at run time with `FieldAccessException` (the SM struct is a private nested type of the Main-declaring class). Found by running, not by compiling — the metadata assertions alone would have passed it.

### Corpus-wide effect — #3841

Two whole-repo `cs2gs migrate --translate-only` passes, identical base, differing only in `CSharpTypeMapper.cs`, then a tree diff:

```
files differing:                                      174
  ... whose ONLY difference is `!!` placement:         171   (run-to-run churn; see note)
  ... substantive:                                       3
      src/Analyzers/InternalAnalyzers/packages.lock.json
      test/Interpreter.Tests/Issue2918InlineLambdaErasedReceiverEmittedOracleTests.gs
      tools/cs2gs/Cs2Gs.Translator/CSharpTypeMapper.gs   (this PR's own new source)
```

The behavioural change across all 54 apps is **4 parameter declarations in 1 file** — the two `Issue2948*` fixture classes, i.e. exactly the two GS0264 fingerprints:

```
-    func Add(callback (T) -> bool) string -> "pred"     +    func Add(callback Predicate[T]) …
-    func Add(callback (T) -> bool) string -> "func"     +    func Add(callback Func[T, bool]) …
-    init(callback (T) -> bool)                          +    init(callback Predicate[T])
-    init(callback (T) -> bool)                          +    init(callback Func[T, bool])
```

Every other `Func`/`Action`/`Predicate` in the corpus keeps arrow form. The scoping held.

> Side observation, not caused by this change: the `!!` null-forgiveness pass is **not deterministic run to run** — 171 files differed in `!!` placement alone between two identical translate passes (total bangs 21726 vs 21738). Worth a look, since the gate's readability metrics count them.

### Closure run

`Cs2Gs.Translator → Cs2Gs.CodeModel → src/Core`, closure kept whole (no `--exclude` narrowing), translate + compile:

```
src/Core/Core.csproj                                  FAIL(3)
tools/cs2gs/Cs2Gs.CodeModel/Cs2Gs.CodeModel.csproj    FAIL(3)
tools/cs2gs/Cs2Gs.Translator/Cs2Gs.Translator.csproj  FAIL(3)
```

All nine are the same three src/Core errors, cascading through the two project references:

| diagnostic | site |
| --- | --- |
| `GS0376` const `ChanTypeName` initializer not compile-time constant | `CodeAnalysis/Binding/ChannelRuntimeBinder.gs:857` |
| `GS0154` `arguments` wants `ImmutableArray[BoundExpression]` | `CodeAnalysis/Binding/ChannelRuntimeBinder.gs:192` |
| `GS0155` cannot convert `SyntaxNode?` to `SyntaxNode` | `CodeAnalysis/Emit/ClosureEmitter.gs:302` |

`ChannelRuntimeBinder.cs` is **entirely new in #3882** (+1058 lines) and `ClosureEmitter.cs` was modified by it (+50/−6). None is delegate-related and none is in `CSharpTypeMapper.gs`, which translated clean. These are #3882's to record, not this PR's.

### `test/Interpreter.Tests` before/after — the expected count is stale

The issue expected 3 compile errors → 1. On current main that app **never reaches the compile stage**: it now fails stage 1 (translate) with two `CS2GS-GAP` fingerprints from `Issue3004ManagedByRefAutoDereferenceFixture.cs`, both the already-known #3839 gap:

```
ref-returning property 'Property' has no canonical G# form   (line 15)
ref-returning indexer has no canonical G# form               (line 19)
```

So no compile count is obtainable for that app until #3839 is addressed. The #3841 clearance is instead pinned directly: the translated fixture now carries the nominal names (diff above), and `Issue3841DelegateOverloadSetTranslationTests` compiles the identical `Issue2948*` shapes with the real gsc, zero errors.

### #3874 remains

gsc still cannot **dispatch** the generic shape — `Add(Predicate[T])` vs `Add(Func[T, bool])` on a substituted receiver reports `GS0266`. Verified against this build. The executing test therefore runs the closed shapes and declares the generic ones; the declaration side, which is what #3841 was about, is clear.

### CI

36/36 green on the merged head, including every compiler / core / cs2gs / interpreter shard, `e2e`, `ilverify` and `cs2gs-corpus`.

Issue #3902, opened 2026-09-04
ADR-0174: closing the G#/Go concurrency gap — measured causes and a ranked plan


DavidObando:
> ## Status — 2026-09-04
>
> **Everything below this block is the original research report.** Four of its numbers have since been
> corrected by measurement and are marked in the comments; read this block first.
>
> ### Landed
>
> | item | where | result |
> |---|---|---|
> | **S1** fix the measurement | #3903, #3918, #3923, #3926 | tier pinned; all eight bodies reach Tier1; both sides duration-matched; official baseline from a named machine; every Go-compared row now does equivalent work |
> | **S2** node-backed receives | #3915 | −144 B and one continuation hop per parked receive |
> | **S3** bulk ring copy, right-sized chunks | #3915 | chunk rows −44% to −58% |
> | **S4** allocation-free select | #3915 | 104 B/select → 0 |
> | **H1** inline hand-off completion (gate G6) | #3915 | rendezvous −82%, select-park −73% |
> | **G7** pooled chunk overload | ADR errata 36 (#3928) | resolved **"no"** — pooling measures 8%, against D10's promise that a communicated buffer may be kept |
>
> ### Blocked
>
> - **S5** (`go` closure as a `GoroutineWorkItem` subclass) — attempted and reverted; see the S5
>   comment. The prize is **~150 B → ~32 B per spawn**, measured. The blocker is metadata shape: the
>   synthesized type is not recognised as implementing the abstract base. **Its timing estimate
>   (7–14%) is inside the 16.4% noise band of the row it targets, so success must be gated on an
>   allocation witness, not on ns/op.**
> - **H2** (G#-emitted `MoveNext` gets no OSR patchpoints) — quantified in its comment; needs a Checked
>   JIT to root-cause. Matters beyond the benchmark: **any hot loop inside a suspending function that
>   does not park runs Tier0 for the life of the process.**
>
> ### Not pursued
>
> - **H3** — S2 put the `Chan<T>` dispatch inside the facade, so its ~15 ns premise is largely spent.
> - **H4**, **H5** — micro-costs; the issue itself says keep H5 deferred.
>
> ### Corrections to the report below
>
> 1. "`TieredCompilation=0` is 5–8× slower" — does not reproduce on the G# bench (first comment).
> 2. "1 Tier1 compile by default, 132 pinned" — those are *isolated single-scenario* counts; whole-suite
>    is 71 and 534 (ADR errata 35).
[...]

Issue #3901, opened 2026-09-04
bench: `select-ready` is not reproducible across sessions, and the harness records nothing that would explain it


DavidObando:
`select-ready` was reported at **172.91 ns/op (1.95x Go)** from a 7-launch run, with a bootstrap CI of **[169.4, 563.9]** — a 228% spread that made the median meaningless. Investigating that spread produced a result I can partly explain and partly cannot, and the part I cannot is the reason this is filed: **the measurement is not reproducible, and nothing in the harness records enough to say why.**

## What was measured

All on the same workstation (`linux-x86_64-20`, .NET 10.0.11 / Go 1.26), same `bench/concurrency/gsharp/Bench.gs`, same Release toolchain (`out/bin/Release/{Compiler,Gsharp.Extensions,Gsharp.Runtime.Channels}`, all timestamped within one minute of each other and unchanged throughout).

| when | G# `select-ready` | Go `go-select2` |
|---|---|---|
| earlier session, 3 separate runner invocations | 164.11, 168.74, 172.91 | 149.0, 102.0, 88.8 |
| later session, 15 back-to-back isolated launches | 564–589 (median 574, **4% spread**) | — |
| later session, 6 full-suite launches | 496–775 (median 596) | — |
| later session, fresh runner invocation | 463.60 | 77.5–177.0 |

**The G# side shifted ~3.4x between sessions and stayed shifted. The Go side did not move.** Same machine, same binaries.

## What was ruled out

- **Not internally bimodal.** Back-to-back the scenario is stable to 4% (15 launches, 564–589). Whatever moved, moved between sessions and then held.
- **Not the machine.** `go-select2` measured 77.5–177.0 today against 88.8–149.0 earlier — the same range. A machine-wide regression would have moved both.
- **Not scenario ordering.** Isolated (574) and 5th-in-suite (596) agree.
- **Not the toolchain.** `gsc`, `Gsharp.Extensions.dll` and `Gsharp.Runtime.Channels.dll` are byte-unchanged across both periods.
- **Not contention, in the direction assumed.** See below — load makes G# *faster*.

## What was found: G# gets faster under CPU load, Go gets slower

A controlled experiment, three launches per condition, 16 spinner processes as the load:

```
quiet    G#: 564.79  576.20  564.06
busy     G#: 504.94  497.41  507.50      (~12% FASTER)
quiet2   G#: 575.38  567.56  569.11      (recovers)
```

[...]

PR #3903, opened 2026-09-04
bench: pin the JIT tier and add a NativeAOT row to the D11 concurrency benchmark

DavidObando:
## What this fixes

The D11 concurrency benchmark reported a number that depended on whether a 100 ms timer happened to
elapse. The runtime's call-counting delay restarts on every new JIT compilation, and a bench process
that keeps first-calling methods can exit before counting ever begins — so the scenario's own loop is
promoted by on-stack replacement while **every method it calls stays at Tier0**.

`DOTNET_JitDisasmSummary=1`, isolated `select-ready` launches on `linux-x86_64-20`:

| | Tier1 compiles in the process | reported |
|---|---|---|
| default | **1** (the OSR of the scenario's own `MoveNext`) | 561.88 ns |
| `TC_CallCountingDelayMs=0` | **132** | 180.69 ns |

That is the cause of the 3.4× session-to-session swing in #3901, which that issue could not explain.
175 ns under the pin is within noise of the 172.91 ns measured in an earlier session and since treated
as retracted — it was never wrong, it was the promoted-tier number. Every row moves: closed-recv
61 → 25 ns, chunk1k 45 → 18, buf64 290 → 219.

`TieredCompilation=0` was rejected as the knob because it also discards dynamic PGO, worth a consistent
~28% on `select-ready`. Not the 5–8× a first analysis claimed — that did not reproduce outside a
synthetic C# probe, and the correction is [on #3902](https://github.com/DavidObando/gsharp/issues/3902).

## The AOT row

`bench/concurrency/aot` measures a NativeAOT build of **the same emitted assembly**: it borrows the
SDK's `PublishAot` pipeline and substitutes gsc's output for csc's after `CoreCompile`, so ILC compiles
G#'s own IL rather than a transcription of it. Hand-rolling the `ilc` command line would have meant
re-deriving a 250-line response file and a link step on every .NET bump.

**ILC accepts gsc-emitted IL unmodified, with no trim or AOT warnings** from the bench,
`Gsharp.Extensions` or the channel runtime. That is worth recording on its own — nothing in this
toolchain had been AOT-compiled before. The G# entry point is `<Program>.Main`, unnameable from C#,
which turns out not to matter: ILC reads it from metadata.

## What a run looks like now

7 launches, this workstation:

```
hardware class: linux-x86_64-20   launches: 7   jit tier pinned
  scenario          jit ns/op    aot ns/op     go ns/op   jit/go   aot/go
  buf64                179.86       205.15        81.40    2.21x    2.52x
  rendezvous          1208.21       956.33       585.50    2.06x    1.63x
  closed-recv           22.74        23.88        30.70    0.74x    0.78x
  spawn                399.74       404.34       278.00    1.44x    1.45x
  select-ready         175.00       203.86        91.60    1.91x    2.23x
  select-park         1053.98       990.38            -        -        -
  chunk64               23.41        22.63         8.20    2.85x    2.76x
  chunk1k               14.81        18.39         1.70    8.71x   10.82x
```

Neither mode is "the" number, and the ADR no longer implies there is one. The JIT row is what a
deployed G# program does; the AOT row is what the language does once compilation is out of the way,
and it is the only mode that compares like-for-like with Go's ahead-of-time binary. They differ **per
scenario rather than by a constant** — AOT wins the parking rows, PGO wins `select-ready` — which is
exactly why reporting one alone misleads. Each carries its own ceiling, because a JIT regression and an
AOT regression mean different things.

## Not in this PR

- **Every median in `baseline.json` stays null.** Nothing here was measured by a nightly, and a budget
  that was not measured reads as evidence.
- **The unequal pairings are untouched.** `rendezvous` counts 1 hand-off against `go-pingpong`'s 2;
  `select-ready` does 4 channel ops against Go's 1; `spawn` times a send and a receive Go doesn't do.
  Verified in the sources, tracked in #3902 — fixing them changes what the scenarios measure, which is
  a separate decision.
- **Why the earlier session's launches promoted and current ones do not** is still open. Load was the
  obvious hypothesis; I tested it and it does not cleanly reproduce (negative result on #3901).

## Consequences

Gates **G3, G5, G6 and G7** must be resolved against pinned measurements, not against anything recorded
before this change. ADR errata 33 records that.

## Verification

End to end from a clean `out/` at the nightly's 7 launches: both modes plus Go, `--update-baseline`
recording both budgets, and the ratchet refusing to loosen an AOT ceiling without `--allow-regression`.
`verify-concurrency-bench.py --smoke` passes and deliberately does **not** publish the AOT binary — it
runs on every PR and that would cost minutes.

Labelled `bench:concurrency` so the workflow actually runs here rather than first executing at 05:23.

### DIFF
--- a/.github/workflows/concurrency-bench.yml
+++ b/.github/workflows/concurrency-bench.yml
@@ -43,7 +43,13 @@ jobs:
     timeout-minutes: 90
 
     steps:
+      # Nerdbank.GitVersioning walks history to compute the version height, so
+      # a shallow clone fails the build outright ("Shallow clone lacks the
+      # objects required to calculate version height"). Every other workflow
+      # here that builds the solution already passes fetch-depth: 0.
       - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
 
       - name: Setup .NET SDK
         uses: actions/setup-dotnet@v5
--- a/.github/workflows/concurrency-bench.yml
+++ b/.github/workflows/concurrency-bench.yml
@@ -14,6 +14,11 @@ name: concurrency-bench
 #   G# versus Go   — informational. It moves with the Go toolchain and the
 #                    machine, so it is reported and never gates.
 #
+# The G# side is measured twice: pinned-tier JIT (what a deployed program does)
+# and NativeAOT (what compares like-for-like with Go's ahead-of-time binary).
+# Each carries its own budget, because they regress for different reasons. See
+# the runner's docstring and issues #3901/#3902 for why the tier is pinned.
+#
 # Nightly by default. A PR can opt in with the `bench:concurrency` label when it
 # touches the runtime or the lowering; that run reports and does not gate,
 # because a shared CI runner is too noisy to fail a PR on.
@@ -61,6 +66,15 @@ jobs:
         with:
           go-version-file: bench/concurrency/go/go.mod
 
+      # ILC emits an object file and hands it to the platform linker, so the
+      # AOT row needs a toolchain the JIT row does not. Present on the current
+      # ubuntu image, installed explicitly so an image change fails loudly here
+      # rather than three steps later inside `dotnet publish`.
+      - name: Install NativeAOT prerequisites
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y clang zlib1g-dev
+
       - name: Restore
         run: dotnet restore GSharp.sln --locked-mode
 
@@ -81,6 +95,7 @@ jobs:
           python3 build/run-concurrency-bench.py \
             --launches "${{ github.event.inputs.launches || '7' }}" \
             --go \
+            --aot \
             --json "$RUNNER_TEMP/concurrency.json" \
             --check-baseline bench/concurrency/baseline.json
         # A PR run reports; only the nightly gates.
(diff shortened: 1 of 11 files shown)


DavidObando:
## First CI run: green, and it corrected me

Both `concurrency-bench` runs passed, and — the thing worth checking — the AOT row **actually measured
on the runner** rather than silently skipping. ILC ran on gsc-emitted IL in CI, on a machine that is
not mine, with no trim or AOT warnings.

```
hardware class: linux-x86_64-4   launches: 7   jit tier pinned
  scenario          jit ns/op    aot ns/op     go ns/op   jit/go   aot/go
  buf64                129.28       229.73        42.80    3.02x    5.37x
  rendezvous           520.32       681.91       305.80    1.70x    2.23x
  closed-recv           38.94        31.51        15.00    2.60x    2.10x
  spawn                202.06       207.57       194.60    1.04x    1.07x
  select-ready         240.24       265.11        86.20    2.79x    3.08x
  select-park          872.30       902.12            -        -        -
  chunk64               35.12        28.78         4.30    8.17x    6.69x
  chunk1k               29.87        35.44         1.00   29.87x   35.44x
```

**This inverts the claim I put in the PR description and in ADR errata 33.** I wrote that "AOT is
markedly better on the parking rows." On the 4-vCPU runner the JIT mode wins almost everywhere —
`buf64` 129 vs 230, `rendezvous` 520 vs 682, `select-ready` 240 vs 265 — with AOT ahead only on
`closed-recv` and `chunk64`. On my 20-core workstation the ranking is the other way round for the
parking rows.

That was a statement about one machine written as though it were a property of the mode. Corrected in
`5542deb6`, and the real finding is stronger than the one I replaced: **which mode wins varies per
scenario and per hardware class.** A single G# figure would have to pick a mode, and the pick would
change the answer differently per row and per machine. That is now the argument in the ADR for keeping
both, and for comparing a row only against runs of the same hardware class — which the harness already
records per run.

The plausible mechanism — dynamic PGO mattering more when there are fewer cores to hide a bad inlining
decision behind — is recorded as a guess, because I have not measured it.

Two other things this run settles: the `Build docs` check passes, so the new dashboard columns are
valid MDX; and `verify-concurrency-bench.py --smoke` correctly does **not** publish an AOT binary on
the PR path.

Issue #3907, opened 2026-09-04
selfmig: migrated Gsharp.Runtime.Channels reports 327 diagnostics once the stack overflow is fixed — needs triage


DavidObando:
Split out of **#3905** / PR #3906, which fixed the gsc stack overflow that was hiding these.

## What was uncovered

The migrated `src/Sdk/Gsharp.Runtime.Channels` (added by #3882) crashed gsc with a stack overflow during generic construction. Because a stack overflow kills the process, **no diagnostics were emitted at all** — the gate reported only `GS9999 … no parseable diagnostics` for this app and the ten that reference it.

With #3906's fix the compile completes and reports **327 diagnostics**:

| id | count | |
|---|---:|---|
| `GS0113` | 86 | type doesn't exist |
| `GS0125` | 43 | variable doesn't exist |
| `GS0159` | 39 | cannot find function |
| `GS0158` | 35 | cannot find member |
| `GS0157` | 26 | cannot find type |
| `GS0536` | 20 | redundant `!!` (polish, excluded from substantive counts) |
| `GS0155` | 17 | conversion |
| … | | remainder |

## Expectations

**This is exposed work, not new breakage.** #3906 un-blinds eleven apps; it does not make them green. The gate will move off 38/53 by revealing this, and the count may not improve much until these are triaged.

## Triage guidance

The distribution is dominated by "cannot find X" shapes — `GS0113`/`GS0125`/`GS0157`/`GS0158`/`GS0159` together are **229 of 327**. On this effort that family has repeatedly been a **mask**:

- #3887 (filed after the fourth instance) records gsc reporting `Cannot find member/function` on an already-error-typed receiver — one real error presented as 16.
- #3886 found 8 gate fingerprints that were **16 sites over 1 root cause**, where the reflection-surface clustering was pure cascade.
- #3880 found three separate masking shapes in one app, one of which turned three misleading messages out of a single bad argument.

So **229 "cannot find" diagnostics are very unlikely to be 229 defects.** Triage by root cause, not by diagnostic id, and expect heavy collapse. Start by finding whether a small number of unresolved types at the top of the file poison everything downstream.

Also note gate fingerprints are **deduplicated signatures, not site counts** — recent apps had 15 sites behind "11 fingerprints" and 16 behind "8". Establish the true site count before sizing.

## Context worth carrying

[...]

Issue #3905, opened 2026-09-04
gsc: stack overflow in generic construction (InterfaceSymbol/StructSymbol mutual recursion) crashes the build for Gsharp.Runtime.Channels and blinds 11 gate apps


DavidObando:
**gsc crashes with a stack overflow** compiling the migrated `src/Sdk/Gsharp.Runtime.Channels`, and because the process dies no diagnostics are emitted — so the gate reports the uninformative `GS9999 dotnet build (--via-sdk) exited with code 1 and no parseable diagnostics` for **eleven** apps that reference it.

Gate run [33841369222](https://github.com/DavidObando/gsharp/actions/runs/33841369222) on `952af2784`: **38/53 green (floor 44)**.

## The recursion

From `sdk.build.log`, the cycle alternates between interface and struct construction:

```
InterfaceSymbol.SubstituteType
  → InterfaceSymbol.SubstituteMethod
  → InterfaceSymbol.TryResolveMembers
  → InterfaceSymbol.CreateConstructed
  → InterfaceSymbol.Construct                 (via ConcurrentDictionary.GetOrAdd)
  → StructSymbol.SubstituteTypeForConstruction
  → StructSymbol.CreateConstructed
  → StructSymbol.Construct                    (via ConcurrentDictionary.GetOrAdd)
  → StructSymbol.SubstituteConstructionArguments
  → InterfaceSymbol.SubstituteType            ← cycle closes
```

Note both `Construct` paths go through `ConcurrentDictionary.GetOrAdd`, so the memoisation that would normally break a cycle is **being entered re-entrantly** — the entry is not yet published when the recursive call arrives.

## Why it is worth prioritising

1. **It is a crash, not a diagnostic.** A stack overflow cannot be caught; the process dies and takes the whole build with it. There is no error message a user could act on.
2. **It blinds the gate.** Eleven apps report `no parseable diagnostics`, which is indistinguishable from a harness fault. Real errors in those apps (if any) are invisible behind it.
3. **It is reachable from ordinary code.** The trigger is a generic type shape in `Gsharp.Runtime.Channels` (added by #3882) — not a migration artifact.

## Suggested approach

Find the minimal shape that recurses. The likely candidates are a struct implementing a generic interface parameterised by itself (F-bounded, `struct S : IChannel<S>`), or mutually recursive generic constraints between a struct and an interface. Reduce to a standalone snippet compiled directly by gsc before touching the code.

Then decide the fix layer:
[...]

Issue #3896, opened 2026-09-04
REGRESSION: selfmig gate 44/52 → 28/53 — three untranslatable errors in #3882's new src/Core files cascade to 16 apps


DavidObando:
**The selfmig gate fell from 44/52 to 28/53** on gate run [33825780017](https://github.com/DavidObando/gsharp/actions/runs/33825780017) (`36a147f0f`), the first run after PR #3882 (ADR-0174 goroutines and channels wave 2, +27,657/−5,096 across 416 files).

```
28/53 green (floor 44); labels=0 (ceiling 0); __local_=30 (ceiling 30);
lines>300=618 (ceiling 610); bangs=19219 (ceiling 11500)
GATE: green count 28 fell below floor 44.
GATE: >300-char line count 618 exceeded ceiling 610.
GATE: !! null-assertion count 19219 exceeded ceiling 11500.
```

## The cause: three errors in `src/Core`, which cascades to everything

`src/Core` is the root of most of the corpus's dependency graph, so its failure took down at least six **banked** `greenApps` entries with it — `ClrBaseline`, `GSharp.CodeAnalysis.Analyzers.Testing`, `Compiler`, `Core`, `LanguageServer`, `Repl` — and the count from 44 to 28.

All three errors are in files this PR created or modified:

| diagnostic | file |
|---|---|
| `GS0376` The initializer for const field `ChanTypeName` must be a compile-time constant | `ChannelRuntimeBinder.gs` |
| `GS0154` Parameter `arguments` requires a value of type `ImmutableArray<…>` | `ChannelRuntimeBinder.gs` |
| `GS0155` Cannot convert type `SyntaxNode?` to `SyntaxNode` | `ClosureEmitter.gs` |

Independently confirmed before the gate ran: a `Cs2Gs.Translator → Cs2Gs.CodeModel → src/Core` closure run during PR #3889 came back red with the same three, and none is delegate- or nullability-fix related.

## This is the standing hazard, not a surprise

**New code added to this repo must itself survive translation.** CI compiles the C# sources; only the gate translates *then* compiles them. So a change can be entirely CI-green — as #3882 was — and still break the gate purely by existing. This has now happened twice: #3831 (a cs2gs file that could not be translated, costing 7 banked apps) and now this.

The difference is scale: `src/Core` is the dependency root, so three errors cost sixteen apps.

## What is NOT being done

[...]

PR #3909, opened 2026-09-04
cs2gs: a PR-time translation guard for the hot core (#3836)

### LINKED ISSUE #3836: selfmig: recently-added Cs2Gs.Tests sources do not survive their own migration (sibling sweep from #3831)

DavidObando:
Closes #3836. Part of #3501. Recommended in PR #3835.

## The hazard

CI **compiles** the repository's C# sources. The self-migration gate
**translates** them to G# and then compiles the result. Those are different
questions, so a fully CI-green PR can take the gate down purely by existing —
and the damage lands hours later, attributed to whatever else merged nearby.
Three times in five days, at escalating cost: #3831 (a `Cs2Gs.Translator`
file, 7 banked apps, gate 43 → 35), #3896 (three errors in new `src/Core`
files, 16 apps, gate 44 → 28), #3905 (a gsc stack overflow on
`Gsharp.Runtime.Channels`, 11 apps blinded).

## What this adds

`build/run-cs2gs-selfmig-pr-guard.sh`, run by
`.github/workflows/cs2gs-pr-guard.yml`, migrates and compiles
`Cs2Gs.Translator`'s **reference closure** and nothing else:

```
src/Analyzers/InternalAnalyzers  src/Core
tools/cs2gs/Cs2Gs.CodeModel      tools/cs2gs/Cs2Gs.Translator
```

Everything outside the closure is `--exclude`d. That is safe **only** because
the set is closed under `ProjectReference` — no kept app references an
excluded one — which is exactly the condition whose violation manufactures
phantom cascades. The script verifies the closure textually before it builds
anything and refuses to run otherwise, so the set cannot silently rot into a
cascade generator when a guarded project grows a new reference.

Widening is a one-line edit to `guard_apps`, or the `SELFMIG_PR_GUARD_APPS`
environment override.

## Evidence

**Green on `main`** (`03b6e1e8d`), all four stages:

```
src/Analyzers/InternalAnalyzers/InternalAnalyzers.csproj  PASS  PASS  PASS  PASS
src/Core/Core.csproj                                      PASS  PASS  PASS  PASS
tools/cs2gs/Cs2Gs.CodeModel/Cs2Gs.CodeModel.csproj        PASS  PASS  PASS  PASS
tools/cs2gs/Cs2Gs.Translator/Cs2Gs.Translator.csproj      PASS  PASS  PASS  PASS

4/4 apps green; run PASSED.
PR guard: 4/4 guarded app(s) migrated and compiled in 1225s (migrate exit 0).
```

**Red on #3896.** `main` with #3900's four production files reverted to their
pre-fix state (the three `src/Core` binder files and the one
`Cs2Gs.Translator` file) — i.e. the tree as it was when #3896 was live:

```
src/Analyzers/InternalAnalyzers/InternalAnalyzers.csproj  PASS  PASS     PASS  PASS
src/Core/Core.csproj                                      PASS  FAIL(3)  skip  skip
tools/cs2gs/Cs2Gs.CodeModel/Cs2Gs.CodeModel.csproj        PASS  FAIL(3)  skip  skip
tools/cs2gs/Cs2Gs.Translator/Cs2Gs.Translator.csproj      PASS  FAIL(3)  skip  skip

1/4 apps green; run FAILED.
```

The three diagnostics are #3896's three, verbatim:

```
error GS0376: The initializer for const field 'ChanTypeName' must be a compile-time constant expression.
error GS0154: Parameter 'arguments' requires a value of type 'System.Collections.Immutable.ImmutableArray<…>'
error GS0155: Cannot convert type 'SyntaxNode?' to 'SyntaxNode'.
```

The cascade into `Cs2Gs.CodeModel` and `Cs2Gs.Translator` is the real one, not
an exclusion artifact — which is what keeping the closure whole buys.

**Runtime**: 1244 s wall (20.7 min) for the green run, of which 1225 s is the
migrate itself; on this machine the four builds were warm and cost 19 s. A
cold CI runner adds the restore plus those builds, so budget ~30 min. The
migrate cost is dominated by `src/Core` (≈10 min compile) and
`Cs2Gs.Translator` (≈6 min); there is no cheap way to cut it without cutting
coverage.

The build step is deliberately **not** `dotnet build GSharp.sln`: it builds
`Cs2Gs.Cli` (Release), the two Debug project-load prerequisites, and then
repacks the Release `Gsharp.NET.Sdk` nupkg immediately before migrating. The
repack is load-bearing — the compile stage consumes the packed SDK, so a
freshly built `gsc.dll` alone is silently ignored and the run would measure
the *old* compiler — and it goes last so nothing is built between the pack and
the migrate (`MSB4236`).

## The `paths:` filter, and why it is wider than #3836 proposed

#3836 proposed `paths: tools/cs2gs/**`. That would have caught only #3831.
The two expensive incidents were **not** cs2gs changes: #3896 was `src/Core`,
#3905 was `src/Sdk/Gsharp.Runtime.Channels`. Since `src/Core` is *already*
inside the guarded closure, adding it to the filter costs nothing at runtime
and catches strictly more. The filter is:

```
src/Core/**                     the guarded sources themselves (#3896)
src/Analyzers/InternalAnalyzers/**
tools/cs2gs/**                  the translator decides what they become (#3831)
src/Compiler/**                 gsc decides whether the result compiles (#3905's class)
src/Sdk/Gsharp.NET.Sdk/**       the SDK the compile stage builds through
build/run-cs2gs-selfmig-pr-guard.sh, .github/workflows/cs2gs-pr-guard.yml
```

At ~30 min this is not free enough to run on *every* PR, which is why it stays
path-filtered rather than unconditional.

## Correcting the framing in #3836 — this does NOT cover #3905

The premise that "a PR-time check would have caught all three" is **wrong for
#3905**, and I could not make it right without shipping a red job.

`src/Sdk/Gsharp.Runtime.Channels` is not in `Cs2Gs.Translator`'s reference
closure — nothing in the closure references it. I did include it at first:
it is a `ProjectReference` leaf, eight projects reference it, and it costs
about four minutes, so on cost/benefit it belongs. **Measured on `main`
(`03b6e1e8d`) it comes back red**: 4/5, `Gsharp.Runtime.Channels` failing
compile with a long list of diagnostics. That is #3907 — #3905's crash fix
un-blinded 327 diagnostics — not a harness artifact.

A guard that is red on `main` from day one gets ignored or disabled, so it is
out, documented as the first thing to add back, and the guarded set is one
line away from including it. (Incidentally, that run is itself evidence the
job detects what it is for: it independently reproduced an open regression in
under 20 minutes.)

`tools/cs2gs/Cs2Gs.Tests` — #3836's "natural second step" — is out for the
same reason and remains so until its ~9 known failures clear.

## What a green run does not mean

The job prints this on **every** run, pass or fail, because a partial check
mistaken for a full one is precisely how all three incidents reached `main`:

```
SCOPE — what a green run here does and does not mean.

  It means: the 4 hot-core project(s) below still translate to G# and
  compile. Those are the projects whose migration failures have cascaded
  widest (#3831, #3896, #3905).

    - src/Analyzers/InternalAnalyzers/InternalAnalyzers.csproj
    - src/Core/Core.csproj
    - tools/cs2gs/Cs2Gs.CodeModel/Cs2Gs.CodeModel.csproj
    - tools/cs2gs/Cs2Gs.Translator/Cs2Gs.Translator.csproj

  It does NOT mean this PR migrates cleanly. The self-migration gate covers
  ~53 apps, four readability ceilings and corpus-wide test parity; this job
  covers none of that. Only the nightly cs2gs-selfmig gate is the gate.
```

`tools/cs2gs/selfmig-baseline.json` is untouched, and no gate script is
modified — only the new script, the new workflow and a README section.

Refs #3831, #3896, #3905, #3907.

### DIFF
--- a/.github/workflows/cs2gs-pr-guard.yml
+++ b/.github/workflows/cs2gs-pr-guard.yml
@@ -1,0 +1,127 @@
+name: cs2gs-pr-guard
+
+# Issue #3836: the PR-time translation guard for the self-migration effort
+# (#3501).
+#
+# WHY THIS EXISTS
+#
+# build.yml COMPILES the repository's C# sources. The nightly
+# cs2gs-selfmig-nightly gate TRANSLATES them to G# and then compiles the
+# result. Those are different questions, so a fully CI-green PR can take the
+# gate down purely by existing — and the damage lands hours later, attributed
+# to whatever else merged nearby. Three times so far:
+#
+#   #3831  a new tools/cs2gs/Cs2Gs.Translator file did not translate;
+#          7 banked apps red, gate 43 -> 35.
+#   #3896  three untranslatable errors in #3882's new src/Core files;
+#          16 apps red, gate 44 -> 28.
+#   #3905  #3882's src/Sdk/Gsharp.Runtime.Channels crashed gsc with a stack
+#          overflow, blinding 11 apps.
+#
+# Each was found the slow way. All three are the same shape: a project most of
+# the corpus transitively references stopped surviving its own migration. This
+# job migrates exactly that hot core, on the PR, in minutes.
+#
+# WHAT IT IS NOT
+#
+# It is not the gate. It checks four projects, not fifty-three; it says
+# nothing about the readability ceilings or corpus-wide test parity. The job's
+# own output repeats that on every run, pass or fail, because a partial check
+# mistaken for a full one is worse than no check at all.
+#
+# It also does not currently cover src/Sdk/Gsharp.Runtime.Channels, so it
+# would NOT have caught #3905 as it stands. That project is red on main today
+# (#3907) and a guard that is red from day one gets disabled; the guarded set
+# is one line away from including it. See build/run-cs2gs-selfmig-pr-guard.sh.
+#
+# THE paths FILTER
+#
+# Scoped to exactly what the job can actually judge: the sources of the
+# guarded projects, plus the compiler and translator machinery that decides
+# how they translate, plus this job's own definition. #3836 suggested
+# `tools/cs2gs/**` alone — but #3896, the most expensive of the three
+# incidents, was a src/Core change, and src/Core is already inside the guarded
+# closure, so including it costs nothing extra to run and catches strictly
+# more.
+#
+# Note: `tools/cs2gs/**` is included in full even though only Cs2Gs.CodeModel
+# and Cs2Gs.Translator are guarded — a change to any part of cs2gs changes how
+# the guarded sources translate, which is the thing being measured.
+
+on:
+  pull_request:
+    paths:
+      # The guarded projects' own sources: a new file here is the #3831 /
+      # #3896 / #3905 hazard directly.
+      - 'src/Core/**'
+      - 'src/Analyzers/InternalAnalyzers/**'
+      # The translator: it decides what the guarded sources become.
+      - 'tools/cs2gs/**'
+      # The compiler and the SDK: they decide whether what it produces
+      # compiles. #3905 was a gsc crash, not a translation defect.
+      - 'src/Compiler/**'
+      - 'src/Sdk/Gsharp.NET.Sdk/**'
+      # The guard itself, so changes to it are exercised by it.
+      - 'build/run-cs2gs-selfmig-pr-guard.sh'
+      - '.github/workflows/cs2gs-pr-guard.yml'
+  workflow_dispatch:
+
+concurrency:
+  group: cs2gs-pr-guard-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+env:
+  # See build.yml: parallel MSBuild nodes corrupt $GITHUB_ENV via
+  # Nerdbank.GitVersioning's GitBuildVersion* emission.
+  NBGV_SetCloudBuildVersionVars: false
+  SELFMIG_PR_GUARD_ROOT: /tmp/gsharp-cs2gs-pr-guard
+
+jobs:
+  guard:
+    name: hot-core translation guard
+    runs-on: ubuntu-latest
+    # The measured run is well under this; the headroom is for a gsc hang,
+    # which is a failure mode this job specifically exists to catch (#3905).
+    timeout-minutes: 60
+
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          # Nerdbank.GitVersioning needs history to compute a version, and the
+          # migrated tree is restored as NuGet packages that carry it.
+          fetch-depth: 0
+
+      - name: Setup .NET SDK
+        uses: actions/setup-dotnet@v5
+        with:
+          global-json-file: global.json
+
+      - name: Restore
+        run: dotnet restore GSharp.sln --locked-mode
+
+      - name: Restore .NET local tools
+        run: dotnet tool restore
+
+      # The script builds the solution itself (which repacks the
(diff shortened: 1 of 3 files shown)


DavidObando:
CI runtime, measured on this PR's own run (the guard's `paths:` filter matched its own files, so it self-tested): [job 33882290540](https://github.com/DavidObando/gsharp/actions/runs/33882290540) — **26m45s** wall on `ubuntu-latest`, of which 1365s (22m45s) is the migrate itself and the rest is restore plus the four cold builds. `4/4 apps green; run PASSED.`

That is the number the `paths:` filter has to be judged against: at ~27 minutes the job is cheap enough to run on any change to the guarded closure, the translator, gsc or the SDK, but not cheap enough to run unconditionally on every PR.

Issue #3748, opened 2026-09-01
gsc: named tuple → `object?` boxing rejected in a lambda return (GS0155, from #3745)


DavidObando:
Split out of #3745 (family F4). Found in the migrated `test/Compiler.Tests` compile wall; 1 of that app's 20 errors.

A **named** tuple returned from a lambda whose declared return type is `object?` fails the boxing conversion with `GS0155`. The same conversion succeeds for an unnamed tuple in the same lambda, and for a named tuple returned from a plain `func` — it is the combination that is rejected.

## Repro (gsc, `0.4.349+cec369958b`)

```gs
import System

class Helper {
    shared {
        func RunProcess(name string, timeoutMilliseconds int32 = 0) (ExitCode int32, Stdout string?, Stderr string?) {
            return (ExitCode: 0, Stdout: name, Stderr: nil)
        }
    }
}

func Invoke(f (() -> object?)) object? -> f()

func main() {
    let r = Invoke(func () object? {
        return Helper.RunProcess("x", timeoutMilliseconds: 2_000)
        // error GS0155: Cannot convert type '(ExitCode int32, Stdout string?, Stderr string?)' to 'object?'.
    })
    Console.WriteLine(r)
}
```

Both of these bind today, which isolates the defect:

```gs
// unnamed tuple, same lambda — OK
Invoke(func () object? { return (1, "a", "b") })

// named tuple, plain func — OK
func BoxNamedTuple() object? { return Helper.RunProcess("x") }
```

## Current evidence in the corpus

`test/Compiler.Tests/IlVerifierTests.cs:206` (`Assert.Throws<XunitException>(() => IlVerifier.RunProcess(…))`, where `IlVerifier.RunProcess` at `test/Compiler.Tests/IlVerifier.cs:226` returns `(int ExitCode, string Stdout, string Stderr)`), migrated to `IlVerifierTests.gs:146,20`.

The conversion classifier presumably needs to strip tuple element names before classifying the boxing conversion on the lambda-return path, as the plain-func return path already does.

Not a regression — not one of the families solved so far.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Nng28yiBdPVdeML7mSZphs

Issue #3676, opened 2026-08-30
cs2gs: `<auto-generated/>` annotations translated at face value, and `default` on a type parameter gains a flow `?` — 5 GS0155 walls block migrated LanguageServer (#3501)


DavidObando:
Migrated `src/LanguageServer` fails to bind with 5 `GS0155` nullability errors. Because `Repl`, `LanguageServer.Tests` and `Compiler.Tests` all compile migrated `LanguageServer`, this single wall blocks **4 apps** in the #3501 self-migration gate — the largest remaining single unlock.

From green run 33308959087:

```
LspServer.gs:1369            GS0155 Cannot convert type 'nil' to 'string'
LspServer.gs:609             GS0155 Cannot convert type 'T?' to 'T'
Models.gs:20                 GS0155 Cannot convert type 'nil' to 'string'
TestDiscoveryComputer.gs:55  GS0155 Cannot convert type 'nil' to 'string'
TestDiscoveryComputer.gs:88  GS0155 Cannot convert type 'nil' to '[]TestDiscoveryItem'
```

## Root cause 1 — `<auto-generated/>` suppresses the nullable diagnostics that make an annotation trustworthy (4 of 5)

`src/LanguageServer` sets `<Nullable>enable</Nullable>`, but its LSP protocol plumbing (`Protocol/Models.cs`, `Protocol/TestModels.cs`, `Server/LspServer.cs`, …) carries an `// <auto-generated/>` header — added to opt out of StyleCop documentation/ordering rules, not because the files are machine-written.

The C# compiler **suppresses its nullable diagnostics inside generated code**. Verified empirically with a two-file `Nullable=enable` project whose files are otherwise identical:

| file | header | diagnostics |
| --- | --- | --- |
| `Hand.cs` | none | `CS8603 Possible null reference return`, `CS8618 Non-nullable property …` |
| `Gen.cs` | `// <auto-generated/>` | **none** |

So a generated file's reference annotations are never checked against the code that assigns or returns them, and this repo's LSP plumbing duly contradicts them:

* `Protocol/Models.cs` — `public string GetFileSystemPath()` with `return null;`
* `Protocol/TestModels.cs` — `public string Filter { get; set; }` and `public TestDiscoveryItem[] Children { get; set; }`, both written `= null` from `TestDiscoveryComputer.cs` and `LspServer.cs` object initializers

[...]

Issue #2500, opened 2026-07-19
cs2gs: explicit nullable type-parameter arguments lose their ? annotation


DavidObando:
## Summary

cs2gs drops the nullable annotation from an explicit generic method type argument when the argument is an unconstrained/reference-like type parameter. Valid C# `Method<T?>(...)` is emitted as G# `Method[T](...)`, while the argument and surrounding return type remain `T?`. gsc then correctly rejects the generated `T? -> T` conversion.

The G# spelling with `[T?]` compiles; this is a translator fidelity bug, not a compiler conversion gap.

## Oahu evidence

Full migration with `gsc 0.3.154+beb6983db3`:

Canonical occurrences in `Oahu.Decrypt`:

1. Fingerprint `sha256:a8eb857ebb694ff67bb5e5761acd781b7365d074ca262814972bcf64f8420ac9`
   - C# source: `src/Oahu.Decrypt/Mp4Operation[TOutput].cs:19`
   - Emitted G#: `corpus_Oahu.Decrypt/Mp4Operation[TOutput].gs:16:83`
2. Fingerprint `sha256:dc30675dd24fdeb7c8c61877d717711c2bc41a25a85ec3bd58ee46e6320f5b0e`
   - C# source: `Mp4Operation[TOutput].cs:29`
   - Emitted G#: `Mp4Operation[TOutput].gs:21:58`

Original C#:

```csharp
public new Task<TOutput?> OperationTask =>
    continuation ?? Task.FromResult<TOutput?>(default);

return (continuation ?? Task.FromResult<TOutput?>(default)).GetAwaiter();
```

Current translation:

```gsharp
prop OperationTask Task[TOutput?] ->
    continuation ?? Task.FromResult[TOutput](default(TOutput?))

return (continuation ?? Task.FromResult[TOutput](default(TOutput?))).GetAwaiter()
```

Both fail with `GS0155: Cannot convert type 'TOutput?' to 'TOutput'.`

## Stable minimal repro

Nullable-enabled C#:

```csharp
using System.Threading.Tasks;

public static class Repro
{
    public static Task<T?> Make<T>() => Task.FromResult<T?>(default);
}
```

The C# Release build succeeds. Default `cs2gs migrate --via-sdk` emits:

```gsharp
import System.Threading.Tasks

class Repro {
    shared {
        func Make[T]() Task[T?] -> Task.FromResult[T](default(T?))
    }
}
```

and gsc reports:

```text
[...]

PR #3910, opened 2026-09-04
gsc + cs2gs: three root causes take migrated Gsharp.Runtime.Channels from 304 diagnostics to 34 (#3907)

DavidObando:
Closes the substantive part of #3907. **304 real error sites → 34**, from three root causes.

## Sizing first

The 327 in the issue are **fingerprints**. The true unique site count is **304** — the SDK build log
has 614 error lines because MSBuild prints each one twice. Every number below is unique sites from the
real `cs2gs migrate` pipeline, isolated to this app with `--exclude` for the other 61 projects.

| after | sites | root removed |
|---|---:|---|
| baseline | 304 | |
| fix 1 | 52 | partial grouping ignored generic arity |
| fix 2 | 43 | bare `default` followed the flow type, not the declared slot |
| fix 3 | 34 | tuple cast spelling + symbolic tuple unboxing |

The issue predicted the "cannot find X" family would collapse. It was 229 of 327 fingerprints; it is
now **3 sites**.

## Root 1 — gsc: `PartialTypeMerger` grouped by name, not (name, arity) — 252 sites

`GroupByKey` keyed on `(package, name)`. ADR-0174 D12 puts a non-generic `Chan` factory host beside the
generic `Chan[T]`, which is why #3882 is where this surfaced — it is the first time the repo's own
corpus has carried that shape. The merger treated them as two parts of one type, reported `GS0475` +
4×`GS0480` against a mismatch it had *manufactured*, and the poisoned type-parameter list became 172
`Type 'T' doesn't exist` plus the entire `GS0125`/`GS0157`/`GS0158`/`GS0159` cascade.

gsc's **non-partial** path already got this right — `class Box[T]` beside `class Box` compiles clean on
`main` today — so this only brings the partial path in line with the rest of the type system, with C#,
and with the CLR.

The nested-type rebuild indexed merged groups by name alone as well. On the grouping fix alone that
would have silently **dropped** one of a nested `Box`/`Box[T]` pair, so it is keyed by `(name, arity)`
too. That one is worth a reviewer's eye — it is the only part of this PR not exercised by the corpus.

## Root 2 — cs2gs: `default` followed the flow type, not the declared slot — 9 sites

The #3676 rewrite that emits `default(T)` when Roslyn's `?` is flow-derived was deliberately scoped to
**return** position. The channels runtime is built on `[MaybeNullWhen(false)] out T value` parameters
whose bodies say `value = default;` — the attribute makes the flow state maybe-null while the declared
parameter type stays a bare `T`, so `default(T?)` landed in a `T` slot.

Extended to assignment and argument position, both of which have a declaration that settles the
question. **Argument position reads the substituted parameter type**, which is what keeps #2500 intact:
`Same[T?](default)` resolves to a parameter of type `T?` and keeps its `?`, while
`ReceiveResult[T](default, false)` resolves to one of type `T` and drops it. Confirmed on the corpus —
the three `default(T?)` sites that remain are assignments to fields the author genuinely declared `T?`.

## Root 3 — a tuple cast, one defect on each side — 9 sites

**cs2gs** emitted a tuple-target cast in the `T(expr)` form, so `((TaskArm, SelectWaiter))state!` became
`(TaskArm, SelectWaiter)(state!!)` — which G# reads as a call on a parenthesized expression, not a
conversion. `ArmDescriptor`'s `ContinueWith` continuation lost its whole deconstruction: eight
"variable doesn't exist" errors for names the pattern never got to bind.

With the spelling fixed the cast still failed, and **that half was real**:
`Conversion.IsValueTypeLikeFrom` ended at `ClrType != null && IsValueType`, while
`TupleTypeSymbol.BuildClrType` returns null as soon as one element is a same-compilation type. So
`cast[(A, B)](o)` bound when every element was BCL-backed and failed on the user's own classes. This is
the **unboxing direction of the hole #3748 fixed for boxing**, and the same hole the neighbouring
`StructSymbol`/`EnumSymbol` cases exist to close.

Verified by *running* the emitted `unbox.any` rather than by classification alone — directly and
through the deconstruction the channels runtime uses — and a wrong-shape cast still throws
`InvalidCastException` rather than yielding a silent default.

## Tests

Each fix has regression tests with the mutant actually run and killed:

| fix | tests | mutant |
|---|---|---|
| arity | `Adr0144PartialTypesBinderTests` ×4 | grouping without arity — 4 fail, and the same-arity/different-names `GS0480` test keeps passing, so the check was **narrowed, not removed** |
| `default` | `Issue3676GeneratedCodeNullabilityTests` ×3 | rewrite restricted back to returns — 2 fail, while both "keeps nullable" tests pass either way, which is what proves it consults the declaration rather than stripping every `?` |
| tuple | `Issue3748SymbolicTupleBoxingTests` ×3 | `IsValueTypeLikeFrom` false for tuples — 3 fail, while the five #3748 **boxing** tests keep passing, placing the fix on the unboxing side specifically |

## Regression, including one open question

| suite | result |
|---|---|
| Core.Tests | 8860 passed, 4 failed — environmental, confirmed `prerequisite missing: no NETStandard.Library.Ref targeting pack` |
| Compiler.Tests | 4707 passed, 3 failed — same family, same count as every prior run on this machine |
| Cs2Gs.Tests | 2752 passed, **1 failed** — `Issue3347RemainingSpillInventoryTests.TranslatorMigration_SpillFamilyStaysRetired` |

**I have not resolved that last one, and it is the reason this PR is up for review rather than
merge-ready.** It passes in isolation, but that is not sufficient evidence: the test loads and
translates `Cs2Gs.Translator.csproj` *itself* — one of the projects this PR edits — and asserts the
translated output carries no `let __spill` temps. My new code uses a switch expression and pattern
combinators, exactly the shapes that historically lower with spill temps. So there is a plausible
mechanism by which this is a **real** regression: C# written in a shape the translator cannot yet
migrate spill-free, which matters in a repo that self-migrates.

Against that reading: it passed in isolation against identical source, which points at process state or
a transient MSBuild design-time load rather than content. A full re-run is in flight to capture the
actual assertion (the first run kept only the tail). If it reproduces, the next step is a full run on a
`main` worktree to separate "pre-existing order-dependent failure" from "mine" — this session already
hit order-dependent cs2gs failures once and that is what settled them.

If review says the switch expression is the cause, the fix is to write those two helpers in a plainer
shape; I would rather that decision be made deliberately than have me pre-emptively de-idiomatise the
code.

## Not in this PR

The remaining **34 sites are roughly nine roots** — triage with grounded-vs-surface labelling is
[on the issue](https://github.com/DavidObando/gsharp/issues/3907#issuecomment-5541220476). Two worth
naming here:

- **7 sites**: a C# `static async` local function becomes an async *lambda* whose declared return is
  `ValueTask[T]` while its body returns `T`.
- **3 sites**: `DeferGraceExpired?.Invoke(null, args)` becomes `DeferGraceExpired?(nil, args)`. **This
  may not be a bug.** G# may simply not have first-class events, making it an ADR-level question about
  what `public static event EventHandler<T>?` should become — worth deciding before anyone writes a
  translator rule for it.

Two notes for whoever picks this up: `cs2gs migrate --app` only resolves corpus-layout ids, so
isolating one repository app means `--exclude` for every other `.csproj`; and the SDK build consumes a
**packed** gsc, so a compiler fix does not reach it until `src/Sdk/Gsharp.NET.Sdk` is rebuilt in
Release — that silently cost one full measurement round here.

Refs #3907, #3882, #3906, #3676, #3748, #2500, ADR-0174 D12.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/Conversion.cs
+++ b/src/Core/CodeAnalysis/Binding/Conversion.cs
@@ -3450,6 +3450,17 @@ public sealed class Conversion
             return true;
         }
 
+        // Issue #3907: a tuple is always `ValueTuple<...>`, a struct — but
+        // TupleTypeSymbol.BuildClrType returns null as soon as ONE element is a
+        // same-compilation type, so the ClrType test below could not see it.
+        // `cast[(A, B)](o)` therefore bound for BCL element types and failed for
+        // the user's own, which is the same "symbols carry no ClrType" hole the
+        // two cases above exist to close.
+        if (type is TupleTypeSymbol)
+        {
+            return true;
+        }
+
         return type?.ClrType != null && type.ClrType.IsValueType;
     }
 
--- a/src/Core/CodeAnalysis/Binding/PartialTypeMerger.cs
+++ b/src/Core/CodeAnalysis/Binding/PartialTypeMerger.cs
@@ -29,8 +29,9 @@ internal static class PartialTypeMerger
 {
     /// <summary>
     /// Merges the <c>partial</c> struct/class declarations of
-    /// <paramref name="declarations"/> that share a <c>(package, name)</c> key,
-    /// leaving lone declarations (and non-partial duplicate groups) untouched.
+    /// <paramref name="declarations"/> that share a
+    /// <c>(package, name, arity)</c> key, leaving lone declarations (and
+    /// non-partial duplicate groups) untouched.
     /// </summary>
     /// <param name="declarations">The top-level struct/class declarations across every tree.</param>
     /// <param name="packageByTree">Maps each syntax tree to its owning package (for the grouping key).</param>
@@ -71,8 +72,9 @@ internal static class PartialTypeMerger
 
     /// <summary>
     /// Merges the <c>partial</c> interface declarations of
-    /// <paramref name="declarations"/> that share a <c>(package, name)</c> key,
-    /// leaving lone declarations (and non-partial duplicate groups) untouched.
+    /// <paramref name="declarations"/> that share a
+    /// <c>(package, name, arity)</c> key, leaving lone declarations (and
+    /// non-partial duplicate groups) untouched.
     /// </summary>
     /// <param name="declarations">The top-level interface declarations across every tree.</param>
     /// <param name="packageByTree">Maps each syntax tree to its owning package (for the grouping key).</param>
@@ -116,14 +118,23 @@ internal static class PartialTypeMerger
     {
         // Preserve first-appearance order of the groups for determinism of the
         // returned list.
-        var order = new List<(string Package, string Name)>();
-        var groups = new Dictionary<(string Package, string Name), List<T>>();
+        //
+        // Issue #3907: arity is part of the key. A type is identified by name
+        // AND type-parameter count, so `Chan` and `Chan[T]` are two types, not
+        // two parts of one — which is what ADR-0174 D12 declares (a non-generic
+        // factory host beside the generic channel). Grouping on the name alone
+        // merged them, then reported GS0475/GS0480 against a mismatch it had
+        // manufactured, and the poisoned type parameter list turned into 172
+        // "Type 'T' doesn't exist" downstream. The non-partial path already
+        // distinguished the two, so this only brings the partial path in line.
+        var order = new List<(string Package, string Name, int Arity)>();
+        var groups = new Dictionary<(string Package, string Name, int Arity), List<T>>();
 
         foreach (var decl in declarations)
         {
             var packageName = packageByTree.TryGetValue(decl.SyntaxTree, out var pkg) ? pkg.Name : string.Empty;
             var name = GetIdentifier(decl)?.Text ?? string.Empty;
-            var key = (packageName, name);
+            var key = (packageName, name, GetArity(decl));
             if (!groups.TryGetValue(key, out var list))
             {
                 list = new List<T>();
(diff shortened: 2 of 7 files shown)


DavidObando:
## The `Issue3347` spill failure does not reproduce — and my hypothesis for it was wrong

Re-ran the full `Cs2Gs.Tests` suite, same command, same source, same build as the run that failed:

```
Passed!  - Failed: 0, Passed: 2753, Skipped: 0, Total: 2753, Duration: 22 m 17 s
```

2753/2753. So **`TranslatorMigration_SpillFamilyStaysRetired` was transient, not content.**

That disproves the mechanism I raised in the PR description. I had argued the failure might be real
because the test translates `Cs2Gs.Translator.csproj` itself and my new helpers use a switch expression
and pattern combinators — shapes that historically lower with spill temps. If that were the cause it
would fail **deterministically**: the test reads source from disk, so the same source must produce the
same `__spill` count on every run. It produced zero on the repeat.

The likely trigger is environmental rather than anything in the diff: the test does an MSBuild
design-time load of `Cs2Gs.Translator.csproj`, and in the failing run I had rebuilt that project
moments earlier, so the load raced with settling `obj/` state. That is consistent with the assertion
that guards it (`project.BoundWithoutErrors`).

**So there is no de-idiomatising decision for review to make.** I raised it as a live possibility and
would rather withdraw it explicitly than leave a speculative concern standing in the description — the
helpers can stay as written. Worth knowing the test has a flaky mode, though: it is a whole-project
MSBuild load inside a unit-test suite, so it will do this again to someone else.

Full local regression, now settled:

| suite | result |
|---|---|
| Core.Tests | 8860 passed, 4 failed — environmental (`prerequisite missing: no NETStandard.Library.Ref targeting pack`) |
| Compiler.Tests | 4707 passed, 3 failed — same family, same count as every prior run on this machine |
| Cs2Gs.Tests | **2753 passed, 0 failed** |

CI on this PR is an independent read on the same question, on a clean machine with no concurrent local
builds.

Issue #3911, opened 2026-09-04
G#: a static event on a user type cannot be read or raised (ADR-0052 follow-up; blocks migrated Gsharp.Runtime.Channels)


DavidObando:
Split out of **#3907**. This is the last *language* gap blocking the migrated `src/Sdk/Gsharp.Runtime.Channels`, and therefore blocking eleven gate apps (see #3907 for why the ten dependents clear only when `Channels` reaches **zero** errors).

## The gap

A **static** event on a user type cannot be read or raised from inside its declaring type. The identical **instance** event works. Minimal repro, compiled with today's `gsc`:

```gsharp
package Probe

import System

class Host {
    event Instance EventHandler[EventArgs]

    internal func RaiseInstance() {
        Instance?(nil, EventArgs.Empty)      // OK
    }

    shared {
        event Static EventHandler[EventArgs]

        internal func RaiseStatic() {
            Static?(nil, EventArgs.Empty)    // error GS0130: Function 'Static' doesn't exist.
        }
    }
}
```

`Static` is accepted as a *declaration* — the migrated G# prints `event ScopeStalled EventHandler[ScopeStalledEventArgs]` inside a `shared` block and the member is emitted — but the name does not bind as its backing delegate in expression position, so neither the null-conditional invocation nor a plain read compiles.

## Why this is a language gap and not a translator defect

ADR-0052 (event declarations) says under **Follow-up work (out of scope)**:

> - Static events on user types

and under **Consequences**:

> - Raising events uses standard null-check + invoke — no special `raise` keyword.

So the raise *form* is settled; the static *case* was explicitly deferred. cs2gs's translation is faithful — the C# is a perfectly ordinary `public static event EventHandler<T>?` plus `X?.Invoke(sender, args)`, and the emitted `Static?(nil, args)` is the spelling ADR-0052 prescribes.

## Where it bites

`src/Sdk/Gsharp.Runtime.Channels` (#3882), three sites, all `public static event`:

| file | shape |
|---|---|
| `GsharpRuntime.cs:70` | `DeferGraceExpired?.Invoke(null, new DeferGraceExpiredEventArgs(budget))` → `GS0130 Function 'DeferGraceExpired' doesn't exist.` |
| `GsharpRuntime.cs:76` | `ScopeStalled?.Invoke(null, new ScopeStalledEventArgs(waited, pending))` → `GS0130 Function 'ScopeStalled' doesn't exist.` |
[...]

Issue #1918, opened 2026-07-03
gsc: async funcs returning ValueTask<T> unsupported; awaiting a ValueTask ICEs (GS9998 NotSupportedException)


DavidObando:
(a) `async func TripleAsync(v int32) ValueTask[int32] { …; return v * 3 }` → GS0155 (no return wrapping for ValueTask) + GS0100. (b) even `let ready = ValueTask[int32](5); let five = await ready` ICEs: `GS9998 NotSupportedException: Derived classes must provide an implementation.` (grid app G10).
## Definition of done
- [ ] Fix + `Issue<N>*.cs` regression test
- [ ] Re-enable the quarantined grid fixture (or add one) and re-capture its baseline
- [ ] `tools/cs2gs/coverage/csharp-construct-inventory.json` row updated

_Discovered by the ADR-0138 conformance-grid build-out (gsc 0.2.585+01ad6ec964); quarantined sources live next to each grid app under `tools/cs2gs/corpus/grid/`._

PR #3914, opened 2026-09-04
gsc + cs2gs: four more root causes take migrated Gsharp.Runtime.Channels from 27 sites to 11 (#3907)

DavidObando:
**Picks up where #3910 (now merged) left off.** Based on current `origin/main` (`914ce5989`), so this diff is only what #3910 does not cover.

## Where this picks up

#3910 takes the migrated `src/Sdk/Gsharp.Runtime.Channels` from 304 error sites to 34 with three root causes. Re-measuring on its head, in isolation through the packed SDK, I get **27 deduplicated compile-error artifacts**. This PR fixes **four more root causes worth 16 of them, leaving 11**.

Measured end to end on the real pipeline (translate → compile via the repacked Release `Gsharp.NET.Sdk` nupkg), not classified:

| stage | artifacts |
|---|---:|
| `origin/main` (`4832a4c5f`) | **194** |
| + #3910 | 27 |
| + this PR | **11** |

## The four roots

**1. gsc — an `async func` LITERAL took its declared Task envelope literally.** Issue #1918 lets an `async func` spell its return-type clause as `Task` / `Task[T]` / `ValueTask` / `ValueTask[T]` instead of the bare awaited result, and `DeclarationBinder.NormalizeAsyncDeclaredReturnType` has unwrapped that for *named* functions ever since. A function literal never got the same treatment, so `async func (p ValueTask[T]) ValueTask[T] { … return t }` rejected its own body with `GS0155 'T' -> 'ValueTask[T]'`, and callers then saw a doubly-wrapped `Task[ValueTask[T]]`. cs2gs emits exactly that shape for a C# `static async ValueTask<T>` **local function**, which `ChannelOps.Awaited` is built from. `LambdaBinder` now applies the same unwrap and rebuilds the observable return type with the wrapper that was asked for.

**2. cs2gs — the runtime's own async helpers are not `suspend func`s.** `IsSuspendingCandidate` treats "names a `Gsharp.Concurrency` type anywhere in the signature or body" as evidence that an `async ValueTask` method is the G# suspend ABI. Inside the runtime assembly that probe is **trivially true for every method**, so it labelled `Gsharp.Runtime.Channels`' own private helpers (`Chan<T>.ReceiveBatchSlowAsync`, `ChannelOps.ForeignReceiveSlowAsync`, `ScopeFrame.JoinAsync`, `AsyncLetCell.CancelIfUnreadAsync`, `ChunkReader<T>.ReadAsync`) `suspend func`. Those are ordinary async methods whose `ValueTask` is **returned un-awaited** by their fast-path callers — and a suspend func cannot express that, because ADR-0174 D4 makes every call to one an implicit await. The probe is now skipped for methods declared *in* the runtime namespace. `[Suspending]` stays authoritative there, which is how #3882's four genuine `ChannelBatchExtensions` suspend methods keep their ABI. **A tightening, not a removal** — the existing test where a `Chan`-using method *outside* the namespace still renders `suspend func` is what pins that down.

**3. gsc — explicit-interface implementations leaked into simple-name lookup.** An ADR-0149 `func (IFoo) Bar(…)` keeps its plain `Name` in the symbol table (only the *metadata* name is mangled), so it joined the overload set for `Bar(…)` written inside the type. C# and the CLR reach such a member only through the interface. In `Chan[T]`, `bool ISendSelectableCore[T].TrySendLocked` therefore resolved its own call **to itself** instead of the private `SendOutcome`-returning helper it exists to wrap, and the enum comparisons below failed with `GS0129 '==' is not defined for 'bool' and 'SendOutcome'`.

**4. gsc — an override did not hide its base on a generic hierarchy.** `class Derived[T] : Base[T]` gives each level its own `TypeParameterSymbol` named `T`, so `Commit(value T)` on the base and on the override are not signature-equal by symbol identity. Both entered the overload set and every call became `GS0266 ambiguous between multiple overloads` — which is what `SelectNode[DateTime].TryCommitReceive` hit in `Timers`. The dedup now also consults the override link the binder already recorded.

Roots 1, 3 and 4 are wrong for **every** G# program, not just this app: gsc's own neighbouring paths already got each of them right (named functions normalize the envelope; the CLR does not expose explicit implementations by simple name; an override hides its base everywhere else).

## Evidence

- Both gsc test classes **compile, ILVerify and RUN** the emitted program. That is deliberate: root 1's second half is a wrong *observable delegate return type* and root 3 is about *which overload is selected* — neither is visible to a binding-only assertion, and on this effort binding-only assertions have passed a wrong fix six times.
- **Failing on the base branch**, honestly labelled: `Issue3907AsyncLambdaTaskEnvelopeTests` was re-run with `LambdaBinder.cs` alone reverted — **2 failed, 0 passed**. Its `Bare*` cases declare the awaited result directly and pass either way, so a mutant that unwraps *unconditionally* is caught too. Roots 3 and 4 were each first reproduced on the unfixed compiler from a 30-line file (`GS0129` and `GS0266` respectively).
- cs2gs: two new cases in `Adr0174SuspendFuncTranslationTests`; the existing four keep passing, including `AsyncValueTask_UsingChan_RendersSuspendFunc`, which consumes the runtime from *outside* it — that is what places the narrowing on the containing namespace rather than on the probe.
- **Full suites, no regressions:** `Core.Tests` **8864 passed / 0 failed**; `Compiler.Tests` **4714 passed / 0 failed**. Both matter here — roots 3 and 4 change member lookup, whose blast radius is the whole binder.

## What this does NOT do

`Channels` is **not green** (11 sites left), so the gate stays at 38/53 and the app stays **out** of `run-cs2gs-selfmig-pr-guard.sh`'s `guard_apps` — the script already records why a day-one-red guard is worse than none. That line goes back when the last of the remaining roots lands.

The remaining 11, triaged on #3907: **static events cannot be read or raised** (3 sites — a language gap, ADR-0052 lists it under out-of-scope follow-up work, filed as **#3911**); a `TaskArm[T] : TaskArm` half-formed base link (3); `Interlocked.Exchange(&field, nil)` picking the non-generic overload (1); a `!!` lift missed on a tuple-deconstruction assignment target (1); a `T?` argument in a `T` slot (1); and an `async func … ValueTask -> voidExpr` arrow body (2) that root 2's fix **surfaced**, so it is on this PR's ledger rather than #3882's.

## The cascade, now proved

`src/Compiler`, `test/Core.Tests` and `src/Repl` each carry a **byte-identical** 19-artifact set on the nightly gate, and every entry names a `Gsharp.Runtime.Channels` symbol — they recompile the migrated runtime's sources and re-report its errors verbatim. So the ten dependents clear **if and only if `Channels` reaches zero**: 194 errors and 1 error cost the same eleven apps. Details on #3907.

Refs #3907, #3910, #3911, #3882, #3501, #1918, ADR-0149, ADR-0174 D4.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/LambdaBinder.cs
+++ b/src/Core/CodeAnalysis/Binding/LambdaBinder.cs
@@ -315,6 +315,26 @@ internal sealed class LambdaBinder
         var isAsync = syntax.IsAsync
             || (isAsyncIteratorReturnType(returnType) && IteratorDetection.ContainsYield(syntax.Body));
 
+        // Issue #1918 / #3907: an `async func` may spell its return-type clause
+        // as the explicit `Task` / `Task[T]` / `ValueTask` / `ValueTask[T]`
+        // wrapper instead of the bare awaited result. DeclarationBinder has
+        // normalized that form for named functions since #1918
+        // (NormalizeAsyncDeclaredReturnType); a function LITERAL took the
+        // clause literally, so `async func (p ValueTask[T]) ValueTask[T] { …
+        // return t }` rejected its own body (GS0155 `T` -> `ValueTask[T]`) and
+        // then handed callers a doubly-wrapped `Task[ValueTask[T]]`. Unwrap
+        // here so FunctionSymbol.Type holds the awaited result for a lambda
+        // exactly as it does for a named function, and remember which wrapper
+        // was asked for so the observable type below rebuilds the same one.
+        var asyncReturnsValueTask = false;
+        if (syntax.IsAsync
+            && !isAsyncIteratorReturnType(returnType)
+            && AsyncReturnTypeNormalizer.TryUnwrapTaskReturnType(returnType, out var awaitedReturnType, out var declaredValueTask))
+        {
+            returnType = awaitedReturnType;
+            asyncReturnsValueTask = declaredValueTask;
+        }
+
         if (isAsync && isAsyncIteratorReturnType(returnType))
         {
             Diagnostics.ReportAsyncIteratorFunctionLiteralNotSupported(syntax.Location, returnType);
@@ -338,7 +358,7 @@ internal sealed class LambdaBinder
         var observableReturnType = returnType;
         if (isAsync && !isAsyncIteratorReturnType(returnType))
         {
-            observableReturnType = WrapAsTask(returnType);
+            observableReturnType = WrapAsTask(returnType, asyncReturnsValueTask);
         }
 
         var fnType = FunctionTypeSymbol.Get(parameterTypes.MoveToImmutable(), BuildVariadicFlagsIfAny(parameterSymbols), observableReturnType);
@@ -347,6 +367,7 @@ internal sealed class LambdaBinder
             parameterSymbols.ToImmutable(),
             returnType);
         synthetic.IsAsync = isAsync;
+        synthetic.AsyncReturnsValueTask = asyncReturnsValueTask;
 
         // Issue #3501 A2: give a named-binding caller the chance to declare
         // the literal's own name into the (still-current) enclosing scope
--- a/src/Core/CodeAnalysis/Symbols/TypeMemberModel.cs
+++ b/src/Core/CodeAnalysis/Symbols/TypeMemberModel.cs
@@ -1105,12 +1105,27 @@ public static class TypeMemberModel
                 continue;
             }
 
+            // Issue #3907: an ADR-0149 explicit-interface implementation
+            // (`func (IFoo) Bar(…)`) is NOT a member of the implementing type's
+            // own lookup surface — C# and the CLR reach it only through the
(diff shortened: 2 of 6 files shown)


DavidObando:
**cs2gs-is-in-its-own-corpus check: PASSED.**

`build/run-cs2gs-selfmig-pr-guard.sh` on this branch: **4/4 guarded apps migrated and compiled in 1514s (migrate exit 0)**. That covers `InternalAnalyzers`, `src/Core`, `Cs2Gs.CodeModel` and `Cs2Gs.Translator` — the hot core, including the `CSharpToGSharpTranslator.Suspension.cs` file this PR edits. This is the #3831/#3896/#3905 hazard (a translator change that CI passes but that breaks the gate by existing); it is clear.

Note the guard's own banner: this says the hot core still migrates, not that the PR migrates cleanly. `Gsharp.Runtime.Channels` remains deliberately outside `guard_apps` because it is still red (11 sites).

Issue #1213, opened 2026-06-26
event member cannot be referenced/invoked (evt?.Invoke(...)) — GS0125/GS0158


DavidObando:
### Summary
An `event` member can be **declared** on a class but cannot be **referenced or invoked**. A bare reference reports `GS0125: Variable '<Event>' doesn't exist`; a `this.`-qualified reference reports `GS0158: Cannot find member`. So `MyEvent?.Invoke(this, args)` (the faithful mapping of C# event raising) does not bind.

### Repro
```gsharp
package p
class C {
    event ChapterRead (object?, int32) -> void
    func Fire(x int32) { ChapterRead?.Invoke(this, x) }     // GS0125: Variable 'ChapterRead' doesn't exist
    func Fire2(x int32) { this.ChapterRead?.Invoke(this, x) } // GS0158: Cannot find member ChapterRead
}
```

### Expected
An `event` member should be referenceable in expression position within the declaring type (and invocable via `?.Invoke(...)` / direct invocation), binding as its delegate type so the standard null-conditional invoke pattern works.

Ideally, the contracted form of event invocation `ChapterRead?(this, x)` would also work. This however is deferrable if the complexity is too high.

### Impact (Oahu.Decrypt)
~4 direct GS0125 (`ChapterRead`, `ProgressUpdated`, `ConversionProgressUpdate`, …) + GS0159 `Invoke` ×4 + GS0158 `OnProgressUpdate` ×2 + GS0144 `ReportProgress` — ~11 errors.

### Area
Binder — member lookup must surface `event` symbols (as their delegate type) in expression position; emit must wire `?.Invoke`.

### Environment
gsc at main (b95400fe). `bug`.

Issue #1072, opened 2026-06-24
G#: 'nil!!' (non-null assertion on nil literal) is not accepted as a non-nullable reference value — breaks faithful C# null! migration (GS0154/GS0155/GS0214)


DavidObando:
### Summary

The `!!` non-null assertion operator applied to the `nil` literal does not produce a non-nullable reference value. `nil!!` keeps the type `nil`, so it cannot be assigned to (or passed where) a non-nullable reference type is expected — producing **GS0154 / GS0155** (or, in a base-constructor argument position, a misleading **GS0214** "no constructor takes N arguments"). The `!!` operator works correctly on a nullable *variable* (`x!!`); only the `nil` literal is unaffected.

This is the faithful G# mapping of C#'s `null!` (null-forgiving), so the gap blocks faithful migration of any C# that uses `null!`.

### Repros

`nil!!` assigned to a non-nullable reference:
```gsharp
package p
class Base { }
class C {
    func Make() Base {
        let x Base = nil!!
        return x
    }
}
// GS0155: Cannot convert type 'nil' to 'Base'.
```

`nil!!` as a constructor/method argument (non-nullable param):
```gsharp
package p
class Base(A int32, B Base) { }
class C { func Make() Base { return Base(1, nil!!) } }
// GS0154: Parameter 'B' requires a value of type 'Base' but was given a value of type 'nil'.
```

`nil!!` in a base-initializer argument (misleading error):
```gsharp
package p
open class Base(A int32, B Base) { }
class Derived : Base { init(a int32) : base(a, nil!!) { } }
// GS0214: Class 'Base' has no accessible constructor that takes 2 argument(s).
```

Working for contrast — `!!` on a nullable variable:
```gsharp
package p
class C {
    func Take(b C) { }
    func Make(x C?) { Take(x!!) }   // compiles
}
```

And plain `nil` (no `!!`) to a non-nullable ref correctly errors (this is expected):
```gsharp
Take(nil)   // GS0155: Cannot convert type 'nil' to 'C'  (correct — needs C? or nil!!)
```

### Expected

`nil!!` should be a non-nullable reference value (target-typed to the expected reference type), assignable to a non-nullable reference target and usable as an argument for a non-nullable reference parameter — matching C# `null!`. It should participate in overload/constructor resolution as a value of the target reference type (so the base-initializer case resolves the 2-arg ctor instead of reporting GS0214).

### Impact

[...]

PR #3916, opened 2026-09-04
gsc + cs2gs: four root causes take migrated Gsharp.Runtime.Channels from 11 sites to 3 (#3907)

DavidObando:
**Picks up where #3914 left off.** Based on current `origin/main` (`d75d6c4fc`).

## Re-measured first, as the issue asks

Scale: **deduplicated compile-error artifacts**, the same one #3914 used for its 27/11. Measured end to end on the real pipeline (translate → compile through the repacked Release `Gsharp.NET.Sdk` nupkg), isolated to this one app with `--exclude` for the other 61 projects.

| stage | artifacts |
|---|---:|
| `origin/main` (`d75d6c4fc`) | **11** |
| + this PR | **3** |

**#3912 moved the count neither way** — it is still exactly the 11 the issue triaged, same ids, same locations. The cascade-suppression worry was unfounded here.

## The four root causes

**1. gsc — a SOURCE declaration now outranks an imported homonym for a package-qualified name.**
`Binder`'s dotted-name resolution ran its greedy reflection prefix walk *before* consulting same-compilation declarations, and a source type has no CLR representation while binding — so whenever the reference set also contained `Pkg.Name`, the **metadata** type won and the source one was never considered. C# resolves that collision the other way round (Roslyn reports `CS0436` and uses the source type).

It bites hardest when an assembly is compiled against a build of **itself**, which is the normal state for `Gsharp.Runtime.Channels`: `ReferenceResolver.FindBundledChannelsRuntimePath` appends the bundled channel runtime to *every* reference set, so while compiling that runtime's own sources every `Gsharp.Concurrency.X` in a base clause bound to the previous build's `X`. One wrong base link on `class TaskArm[T] : Gsharp.Concurrency.TaskArm` produced all three of the issue's `TaskArm` diagnostics at once — `GS0214` (no accessible 2-arg base ctor), `GS0185` (override does not match the base), `GS0155` (`TaskArm[T]` is not an `ArmDescriptor`).

This also explains the issue's "**does not reproduce** in a 30-line file with the same shape, dotted package included". It does not, and cannot: the missing ingredient was not in the code at all. `/assemblyname:Gsharp.Runtime.Channels` + `package Gsharp.Concurrency` + the qualified spelling is the whole repro, and it reproduces in 14 lines. The **bare** spelling was always correct, which is why only the qualified one failed.

The new probe is deliberately strict — the qualifier must exactly name a package this compilation declares, the simple name and arity must match, and two candidates yield no match rather than a guess — so `System.Console` and every other purely-imported qualified name reach the existing walk unchanged. There is a test for that.

**2. gsc — static events can be read, raised and subscribed to by bare name (#3911, ADR-0052 amended).**
The issue flagged this as the item most likely to need a language decision. **It did not, and the premise on #3911 was wrong**: static events are not an unimplemented feature. Issue #263 shipped declaration in a `shared` block, `add`/`remove`/`raise` accessors with the #256 `Interlocked.CompareExchange` loop, `EventDef`/`MethodSemantics` metadata, and `Type.Event += handler` subscription. Exactly one lookup step was missing — `Binder`'s static bare-name exposure covered static fields, const fields and static properties but not a field-like static event's **backing field**, and `BindBareEventOrCompoundAssignment` walked instance events only. So a static event could be declared and subscribed to but never read or raised: `Static?(nil, args)` gave `GS0130` while the identical instance event worked.

The fix is the exact static counterpart of the instance exposure (#1213/#1221), plus a static branch in the bare `+=`/`-=` path placed *before* the compound-assignment fallback — without it the new backing-field exposure would capture `+=` and turn a subscription into a delegate compound assignment on the field, silently bypassing the accessors' CAS loop. ADR-0052's "Static events on user types" follow-up entry is struck and replaced with an amendment saying what is supported and what is still deferred.

Both things the coordinator asked me to check while I was in there are covered by tests: `+= nil` is a silent no-op (the static path routes its handler through the same `BindEventSubscriptionHandler` that carries #3775/#3793's nil tolerance), and the generated accessors are the thread-safe #256 ones.

**3. gsc — an `async` function whose return clause spells the void-result envelope may use an arrow body over a void call.**
#1918 lets an `async func` name `Task` / `ValueTask` instead of the bare awaited result, and `DeclarationBinder` unwraps it. The parser's return-vs-statement choice for an arrow body (`asReturn: type != null`) runs long before that, so `async func Run[T](body ValueTask[T]) ValueTask -> Deposit(await …)` lowered to `{ return <void expr> }` and the binder rejected its own body with `GS0122` + `GS0124`. Both neighbouring spellings were already right: `func F() -> voidExpr` and `async func F() -> voidExpr` (no clause) both produce an expression statement. This is the C# shape `public async ValueTask M<T>(…) => VoidCall();`, which is what `AsyncLetCell.Run` is written as. The new test pins all four spellings, including the two that already worked.

**4. cs2gs — an inferred local keeps a delegate's nominal type.**
`var handlers = UnhandledGoroutineException;` is a plain `var`; the emitted type clause exists only to carry the #1072 nullable widening. Mapping it through the structural mapper erased `EventHandler<T>` to `(object?, T) -> void`, and in G# the conversion between a nominal delegate and a structurally equal function type is **explicit** — so the emitted declaration failed with `GS0156`. `CSharpTypeMapper.MapEventType` already exists for precisely this reason ("structurally equivalent delegates are not interchangeable in the CLR"); this routes the inferred-local branch through it. It falls through to `Map` unchanged for every non-delegate type, so only this shape moves.

## Layer judgement

**gsc (1, 2, 3).** The C# is ordinary and legal, and in each case gsc's own neighbouring path already got it right: bare-name base clauses resolve to source; instance events are exposed by bare name; a named async function's envelope is normalized. These are wrong for *every* G# program, not just this app — #1 in particular silently mis-resolves any project compiled against an older build of itself.

**cs2gs (4).** The C# is fine and the emitted G# is not what the author wrote.

**No source edits.** I did not touch #3882's C#, and in particular did not turn the process-wide diagnostic hooks into instance events — which the owner's decision made moot anyway, but which was also unnecessary once the actual hole turned out to be one missing lookup step.

## Evidence

- `Issue3907StaticEventTests` (5 cases) and `Issue3907SourceTypePrecedenceTests` (3 cases) **compile, ILVerify and RUN** the emitted program and assert on observable behaviour: a subscriber actually fires, `-=` actually detaches, a nil handler leaves the invocation list untouched, the base constructor actually ran, the override actually won, each arrow body ran its void call exactly once and in order.
- **Failing on the base branch, honestly labelled.** Re-run with the four gsc files reverted: **7 of 8 fail**, with exactly `GS0130`/`GS0125` (static events), `GS0214` (base link) and `GS0122`+`GS0124` (arrow body). The 8th — `PackageQualifiedName_StillReachesAnImportedTypeWhenSourceDeclaresNoSuchName` — is the deliberate anti-over-reach guard and must pass both ways.
- **Full suites:** `Core.Tests` **8874 passed / 0 failed**; `Compiler.Tests` **4722 passed / 0 failed**. Both matter: #1 changes type resolution and #2 changes bare-name lookup, whose blast radius is the whole binder. Plus **509 targeted `Cs2Gs.Tests`** (golden, nullability, deconstruction, event) — the full Cs2Gs suite has a known non-deterministic host crash, so it is verified by subset as usual.
- **PR guard:** `build/run-cs2gs-selfmig-pr-guard.sh` — result in a comment below.

## What this does NOT do, and the corrected remainder

**`Channels` does not reach zero, so the ten dependents do not clear and the gate stays at 38/53.** The all-or-nothing property is unchanged, and I did not beat it. `Channels` therefore stays **out** of `run-cs2gs-selfmig-pr-guard.sh`'s `guard_apps`.

Three artifacts remain, but only **two** are defects — and this corrects the issue's triage:

**A. `GS0154` at `Chan{T}.gs:173` — gsc ignores `@AllowNull`.** The issue called this "a `T?` in a `T` slot", implying a cs2gs neighbour of root 2. It is not. `ReceiveResult<T>`'s C# ctor is `public ReceiveResult([AllowNull] T value, bool ok)` and cs2gs translates it faithfully to `init(@AllowNull value T, ok bool)`. **gsc has no `AllowNull` support at all** — grep finds it only as a consumer in `Compilation.cs`. `[AllowNull]` is the .NET-standard "this *input* accepts null though its type is not nullable" annotation and is exactly what makes the C# legal. Verified in 30 lines: a `@AllowNull T` parameter and a plain `T` parameter reject a `T?` argument identically. cs2gs is already right on its side — `ObliviousNullabilityAnalyzer.HasAllowNullWriteContract` correctly declines to add a `!!` because the target accepts null. **Fix belongs in gsc**: carry the annotation on `ParameterSymbol` and relax the *argument* conversion only (never the parameter's own type, so reads inside the body still require the author's `!!` — the asymmetry the attribute exists to express). I scaffolded this and backed it out rather than ship it half-done: the conversion must be annotation-only, and getting it wrong on a value-type `T` would emit `Nullable<T>` where `T` is expected — a wrong-IL bug an ILVerify-passing test would not catch.

**B. `GS0156` at `Chan{T}.gs:451`/`452` — a by-ref argument over a same-compilation user class erases to `object&`.** Root-caused precisely. `Interlocked` declares both `Exchange<T>(ref T, T) where T : class` and the non-generic `Exchange(ref object?, object?)`. Issue #658 erases a user class argument to `System.Object` because its CLR type is a surrogate while binding; for a by-value argument that ride-through is harmless and load-bearing, but for a **by-ref** one it makes the non-generic overload an exact *identity* match, so gsc picks it and the call's type becomes `object`. csc never considers that overload at all, because a `ref` argument requires exact type identity. Confirmed by three controls: the same call over a BCL element type resolves correctly; `Exchange[Node?](&slot, nil)` with an explicit type argument resolves correctly; and the defect is independent of the reference set. Note the neighbouring `Volatile.Write(&receiveNodePool, node)` almost certainly binds the wrong overload too and is simply *silent*, because its result is discarded — so this is a latent correctness bug, not only a diagnostic. `ImportedClassSymbol.TryLookupFunction` already has the exact analogue for user **value** types (#1599/#1601, feeding the "matches any by-ref parameter" sentinel); the reference-type case needs the *symbolic* type-argument vector to reach inference (`deferredInferenceArgs` + `recoverTypeArgSymbols`), which that path does not thread today. I attempted both the sentinel widening and a `ref`-identity rule in `ClrOverloadResolution.ClassifyImplicit`, verified neither is sufficient, and reverted both.

**C. `GS0116` at `SelectRandom.gs:59` is not a defect at all — it is a pipeline artifact, and the issue's diagnosis of it is wrong.** The issue calls it "the `!!` lift skips a tuple-deconstruction assignment TARGET". cs2gs's lift is present and correct: translating `SelectRandom.cs` emits `order!![i], order!![j] = order!![j], order!![i]`, verified three ways — standalone, standalone with `#nullable enable`, and with all 41 project files loaded together. The assertions are removed afterwards by `NullAssertionPolishPass`, whose log for this app reads `assertions stripped: 24`. The keep-rule is the bug:

```csharp
if (polished.IsAvailable && (polished.Succeeded || !result.Succeeded))
{
    result = polished;   // kept even if stripping INTRODUCED an error
```

The rollback below it only fires when a strip "regressed a previously passing" build. On an app that is already red for unrelated reasons, `!result.Succeeded` is true, so a strip that introduces a *new* error is kept permanently and invisibly. That is exactly what happens here. **It self-heals the moment A and B land**: with the app otherwise green, the same strip would regress a passing build and the backups would be restored. So `Channels` has **two** remaining root causes, not three — but the polish keep-rule is worth fixing on its own, because while it does not block anything it actively manufactures phantom sites and misdirects triage, which is precisely what it did here.

Refs #3907, #3911, #3910, #3914, #3501, #1918, #263, #658, #1599, #3775, ADR-0052.

### DIFF
--- a/docs/adr/0052-event-declarations.md
+++ b/docs/adr/0052-event-declarations.md
@@ -162,4 +162,64 @@ This matches exactly what a C# `public event Action<object, EventArgs> Click;` p
 - `delegate MyHandler(…) ` syntax for named delegate types;
 - Thread-safe field-like event accessors (Interlocked.CompareExchange pattern)
 - `raise` accessor support (C#-style, rarely used)
-- Static events on user types
+- ~~Static events on user types~~ — **superseded, see the amendment below.**
+
+## Amendment (2026-09-04, issues #3911 / #3907): static events are supported
+
+**Static events on user types are no longer out of scope.** The entry above is
+struck; this section is now authoritative for them.
+
+A static event is declared in a `shared` block, with the same grammar,
+accessibility surface and accessor forms as an instance event:
+
+```gsharp
+class GsharpRuntime {
+    shared {
+        event DeferGraceExpired EventHandler[DeferGraceExpiredEventArgs]?
+
+        func RaiseDeferGraceExpired(budget TimeSpan) {
+            DeferGraceExpired?(nil, DeferGraceExpiredEventArgs(budget))
+        }
+    }
+}
+
+GsharpRuntime.DeferGraceExpired += func(sender object?, e DeferGraceExpiredEventArgs) { … }
+```
+
+### What is supported
+
+Everything the instance form supports, with identical semantics:
+
+- **Declaration** — field-like and explicit-accessor (`add` / `remove` / `raise`), in a `shared` block.
+- **Subscription** — `Type.Event += handler` / `-= handler` from outside, and the bare
+  `Event += handler` spelling inside the declaring type's own members.
+- **Reading and raising** — the event's name resolves inside the declaring type's members
+  to its backing delegate field, so `Event?(sender, args)`, `let h = Event`, and
+  `if Event != nil` all work. This mirrors C#, where a field-like event access
+  *within the program text of the declaring type* is an access to the backing field.
+- **Metadata** — an `EventDefinition` row, `add_`/`remove_`/`raise_` specialname
+  static methods and `MethodSemantics` rows, so C# consumers subscribe with
+  ordinary `+=` / `-=`.
+- **Thread safety** — the generated field-like `add`/`remove` accessors use the
+  same `Interlocked.CompareExchange` loop as the instance form (issue #256).
+- **Nil handlers** — `Event += nil` is a silent no-op, as in C#, because the
+  static path binds its handler through the same conversion the instance path
+  uses (issue #3775 / PR #3793). It is not a conversion error and not a throw.
+
+### History, and why the entry was stale
+
+Most of this shipped with issue #263 (`shared`-block static events: declaration,
+accessors, metadata emission, `Type.Event += handler`). What was never wired up
+was the *bare-name* half — reading, raising, and the bare `Event += handler`
+form inside the declaring type — because the binder's static bare-name exposure
+covered static fields, const fields and static properties but not static events'
+backing fields, and the bare `+=`/`-=` path walked instance events only. So a
+static event could be declared and subscribed to but never read or raised, which
+is what issue #3911 reported: `Static?(nil, args)` failed with `GS0130` while the
+identical instance event worked. Issue #3907 fixed both halves.
+
+### Still deferred
+
+Unchanged from the list above, and orthogonal to static-ness: the `delegate`
+keyword for named handler types, and custom (non-`Action`/`Func`/`EventHandler`)
+delegate declarations. Static events reach these exactly as instance events do.
--- a/src/Core/CodeAnalysis/Binding/Binder.cs
+++ b/src/Core/CodeAnalysis/Binding/Binder.cs
@@ -633,6 +633,52 @@ public sealed class Binder
                         }
                     }
                 }
+
+                // Issue #3907 / #3911: expose a field-like STATIC event of the
+                // enclosing type as a bare name, bound to its backing delegate
+                // field — the exact static counterpart of the instance
+                // exposure above (issues #1213 / #1221), and what makes the
+                // canonical raise pattern
+                //
+                //     shared { event Ticked EventHandler? }
+                //     shared func Raise() { Ticked?(nil, EventArgs.Empty) }
+                //
+                // resolve, precisely as C# compiles a field-like event access
+                // inside its declaring type to a read of the backing field.
+                //
+                // Static events were fully implemented by issue #263 —
+                // declaration, accessor emission with the #256
+                // Interlocked.CompareExchange loop, EventDef/MethodSemantics
+                // metadata, and `Type.Event += handler` subscription all work.
+                // This ONE lookup step was the omission, so a static event
+                // could be declared and subscribed to but never read or raised.
+                // ADR-0052's "Static events on user types" follow-up entry is
+                // amended accordingly.
+                //
+                // Scoped to the enclosing type with no base-class walk, which
+                // is what every other static path here does (see the note on
+                // the static-field block above). A bare `Ticked += handler`
+                // still routes to the event-subscription path, which
+                // BindBareEventOrCompoundAssignment checks before it consults
+                // this variable.
+                if (!ownerStruct.StaticEvents.IsDefaultOrEmpty)
+                {
+                    foreach (var evt in ownerStruct.StaticEvents)
+                    {
+                        if (!evt.IsFieldLike
+                            || evt.BackingField == null
+                            || paramNames.Contains(evt.Name)
+                            || seenMembers.Contains(evt.Name))
+                        {
+                            continue;
+                        }
+
+                        if (seenMembers.Add(evt.Name))
+                        {
+                            scope.TryDeclareVariable(new ImplicitStaticFieldVariableSymbol(ownerStruct, evt.BackingField));
+                        }
+                    }
+                }
             }
 
             // ADR-0089 / issue #1030: expose interface static *state* by bare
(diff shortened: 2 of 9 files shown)


DavidObando:
**PR guard: 4/4 apps green, run PASSED** (`build/run-cs2gs-selfmig-pr-guard.sh`, migrate exit 0).

```
src/Analyzers/InternalAnalyzers/InternalAnalyzers.csproj   PASS  PASS  PASS  PASS
src/Core/Core.csproj                                       PASS  PASS  PASS  PASS
tools/cs2gs/Cs2Gs.CodeModel/Cs2Gs.CodeModel.csproj         PASS  PASS  PASS  PASS
tools/cs2gs/Cs2Gs.Translator/Cs2Gs.Translator.csproj       PASS  PASS  PASS  PASS
```

translate → compile → ilverify → test-parity, all four stages, for the hot core. This matters more than usual for this PR: root causes 1 and 2 change type resolution and bare-name lookup, whose blast radius is the whole binder, and root 4 is in `Cs2Gs.Translator`, which is inside its own migration corpus.

`src/Sdk/Gsharp.Runtime.Channels` is **not** added to `guard_apps` — it is still red (3 artifacts, 2 real root causes), and the script's own note explains why a day-one-red guard is worse than none. The line goes back when A and B in the PR description land.

DavidObando:
## Fixed the `cs2gs-issue34` failure — a real regression from root cause 1, now scoped correctly

`Issue3466NestedHomonymAliasTests.SameNamespaceTopLevelType_InsideNestedHomonymScope_StaysQualified`, reproduced locally in 20 lines:

```
package Demo
class Target { }
class Holder {
    class Target { }                                  // nested homonym
    shared {
        func MakeNestedScope() Demo.Target {
            return Demo.Target()
        }
    }
}
```
```
error GS0155: Cannot convert type 'Holder.Target' to 'Target'.
```

**What went wrong.** My source-over-metadata probe was over-scoped. It fired on *any* package-qualified name the compilation declares — including this one, where the competing declaration is another **source** type (the nested `Holder.Target`) rather than an imported one. That is a source-vs-**source** question, and it is governed by ordinary scope shadowing, not by this rule. The result was that the return-TYPE clause started resolving `Demo.Target` to the top-level type while the construction EXPRESSION `Demo.Target()` beside it still resolved to `Holder.Target` — so the two disagreed and the round-trip stopped binding.

**The fix.** The probe now also requires that an imported homonym actually exist:

```csharp
if (scope.TryLookupSourceTypeInPackage(pkg, name, arity, out var sourceType)
    && TryResolveOuterPrefix(segmentTexts, totalSegments, targetArity) != null)
```

That is exactly C#'s CS0436 rule and nothing more: *when both a source and an imported declaration carry this fully-qualified name, source wins.* `Demo.Target` has no CLR type at all, so the walk and the simple-name fallback under it resolve it exactly as they did on main — #3466's behaviour is byte-identical again. The channels case is unaffected because the bundled `Gsharp.Runtime.Channels.dll` genuinely does supply `Gsharp.Concurrency.TaskArm`.

**Not a silent re-baseline.** No expectation was changed; the test and its assertions are untouched. This was a real binding regression in my change, caught exactly where the standing constraint says it should be — it altered *what binds* in a way I did not intend, so the fix was to narrow the rule rather than move the goalposts.

**A genuine pre-existing bug it surfaced, reported not widened.** The two spellings disagreeing means gsc's **qualified-construction** path resolves a package-qualified name by inner-scope simple name: inside `Holder`, `Demo.Target()` binds to `Holder.Target`, though the C# it came from is `new global::Demo.Target()`. On main both sides were wrong together, so it was invisible. That is a separate defect in expression binding and I have deliberately not pulled it into this PR.

**Re-verified after the narrowing:**
- `Issue3466` 33/33 and `Issue3907` 12/12 pass.
- `Gsharp.Runtime.Channels` still measures **3** artifacts — the narrowing costs nothing there.
- The 14-line `/assemblyname:Gsharp.Runtime.Channels` matrix (`TaskArm`, `SelectWaiter`, `ArmDescriptor` present in the bundled dll; `Widget`, `Zorblatt`, `Box` absent; both assembly names) is clean in all 12 combinations.
- Full `Core.Tests` and `Compiler.Tests` re-running on the amended head; results to follow.

DavidObando:
**CI is green on the amended head (`bfc7655a1`): 37/37 checks pass, 0 failures.**

`tests (cs2gs-issue34)` — the shard that caught the over-scoped probe — now passes, as do `tests (core-binding)`, `tests (cs2gs-remainder)`, `cs2gs-corpus`, `cs2gs-oahu`, `cs2gs-code-exploder`, `ilverify`, `e2e`, `nullable-hygiene`, and the **hot-core translation guard**. The four non-passing entries are `skipping` (publish/bench/pages), not failures.

That is the CI counterpart of the local `run-cs2gs-selfmig-pr-guard.sh` run reported above (4/4 green), so the hot core is confirmed twice on two different machines.

Local re-verification on the amended head: `Core.Tests` **8874 passed / 0 failed**, `Issue3466` 33/33, `Issue3907` 12/12, and `Gsharp.Runtime.Channels` still measures **3** deduplicated artifacts — the narrowing cost nothing on the target app.

Issue #3694, opened 2026-08-30
cs2gs: oblivious test writes null into an ANNOTATED non-nullable property (cross-context write) — GS0155 at an object initializer


DavidObando:
Split out of the [#3501 `test/Core.Tests` triage](https://github.com/DavidObando/gsharp/issues/3501#issuecomment-5469492729) family F3, whose array/slice half is fixed in the PR for #3682. This is the one remaining F3 site whose repair is *not* local to the literal.

`test/Core.Tests` is nullable-**oblivious** (`build/gsharp.build.props` sets `<Nullable>disable</Nullable>` for test projects) while `src/Core` is **annotated**. `Compilation.DebugInformation` is declared non-nullable but its setter normalises `null`:

```csharp
// src/Core/CodeAnalysis/Compilation/Compilation.cs:119
/// Setting <see langword="null"/> is normalised to a fresh default instance.
public DebugInformationOptions DebugInformation
{
    get => debugInformation;
    set => debugInformation = value ?? new DebugInformationOptions();
}
```

The oblivious test writes `null` into it, and C# reports nothing because the *write* crosses a nullable-context boundary — the same cross-context root cause as F6 (fixed in #3687), but in the write direction, where no `!!` bridge exists because the value genuinely **is** `nil`:

```csharp
// test/Core.Tests/CodeAnalysis/Emit/DebugInformationOptionsTests.cs:38
var compilation = new Compilation(SyntaxTree.Parse("package P\n"))
{
    DebugInformation = null,
};
```

```gs
// migrated/test/Core.Tests/CodeAnalysis/Emit/DebugInformationOptionsTests.gs(27,91)
let compilation = Compilation(SyntaxTree.Parse("package P\n")){DebugInformation = nil}
// error GS0155: Cannot convert type 'nil' to 'GSharp.Core.CodeAnalysis.Emit.DebugInformationOptions'.
```

Confirmed still live after the #3682 fix (whole-repo translate + `cs2gs validate --app test/Core.Tests/Core.Tests.csproj`).

**Why it is not in the #3682 fix.** The three tractable repairs all fail here:

- **Promote the value** — impossible, the value is a literal `nil`.
- **`!!` the value** — turns a clean C# assignment (which the setter deliberately accepts) into a runtime throw.
- **Promote the literal's own type** — there is no literal type to promote; the target is a *declaration* in another project.

[...]

Issue #1902, opened 2026-07-03
cs2gs: LINQ query clauses without a lowering: second `from` (SelectMany), `let`, `join`(+`into`), `group`


DavidObando:
Each reports `CS2GS-GAP: query clause/query '<K>' has no canonical G# lowering yet`: `from t in tens from o in ones select t+o`; `let sq = n * n`; `join p in pets on o.Id equals p.OwnerId` (+ `into petGroup`); `group n by n % 3`. Basic from/where/orderby/select/into all lower correctly (grid app G11). Kinds: `FromClause` (2nd), `LetClause`, `JoinClause`, `JoinIntoClause`, `GroupClause`.
## Definition of done
- [ ] Fix + `Issue<N>*.cs` regression test
- [ ] Re-enable the quarantined grid fixture (or add one) and re-capture its baseline
- [ ] `tools/cs2gs/coverage/csharp-construct-inventory.json` row updated

_Discovered by the ADR-0138 conformance-grid build-out (gsc 0.2.585+01ad6ec964); quarantined sources live next to each grid app under `tools/cs2gs/corpus/grid/`._

Issue #658, opened 2026-06-10
Binder: passing a G# class to a constructor parameter typed as a CLR interface fails to bind (especially with default-valued params); workaround requires explicit interface-typed local


DavidObando:
## Summary

When calling a constructor whose first parameter is typed as a CLR interface, passing a G# class that **implements** that interface fails to bind unless the value is first stored in an interface-typed local. The diagnostic is misleading — "Function 'X' doesn't exist" or "Cannot convert type 'MyClass' to 'X'".

This bites particularly hard when the constructor also has optional parameters with defaults — overload resolution evidently doesn't try the user-class → interface implicit conversion when narrowing across the multiple synthesized overloads.

## Repro

C#:
```csharp
namespace Probe.CSharp;
public interface IExecutor { void Run(); }
public sealed class JobScheduler
{
    public JobScheduler(IExecutor executor, int retries = 3) { … }
}
```

G#:
```gsharp
package Probe.Tests
import Probe.CSharp

type ProbeExecutor class : IExecutor {
    func Run() { }
}

type Probe class {
    @Fact
    func Bind_With_User_Iface_Impl() {
        let e = ProbeExecutor()
        // Fails: GS0155 / "Function 'JobScheduler' doesn't exist"
        let s1 = JobScheduler(e, 5)

        // Works: explicit interface-typed local
        let exe IExecutor = ProbeExecutor()
        let s2 = JobScheduler(exe, 5)
    }
}
```

## Workaround

Always pre-bind the value to an interface-typed local:
```gsharp
let exe IExecutor = ProbeExecutor()
let s = JobScheduler(exe, ...)
```

## Context

- `gsc` version: `Gsharp.NET.Sdk/0.1.516`
- The single-argument form `JobScheduler(e)` (using only the default `retries`) sometimes works; the N-argument form usually fails. The trigger seems related to overload selection between the synthesized arity variants for default-valued parameters interacting with the user→interface implicit conversion.
- Hit while porting `Commands/ConvertCommandTests` (CapturingExecutor → JobScheduler) and `App/JobSchedulerTests` (multi-arg ctor calls).

DavidObando:
Followup: tested the two suggested alternative cast spellings in 0.1.516. **Both work**.

```gsharp
// All three produce the same binding need for JobScheduler(IExecutor, int) ctor:

// ✅ Type-function cast
let exe = IExecutor(ProbeExecutor())
let s = JobScheduler(exe, 5)

// ✅ `as` cast
[...]

PR #3927, opened 2026-09-04
gsc: six root causes take migrated Gsharp.Runtime.Channels from 7 compile errors to 0 (#3907)

DavidObando:
**Picks up where #3916 left off.** Based on `64401d0c1`.

## Re-measured first, and the count had moved the wrong way

Scale: **deduplicated compile-error artifacts**, the same one #3914/#3916 used for 27/11/3. Measured end to end on the real pipeline (translate → compile through the repacked Release `Gsharp.NET.Sdk` nupkg), isolated to this one app with `--exclude` for the other 61 projects.

| stage | compile-error artifacts |
|---|---:|
| `origin/main` (`64401d0c1`) | **7** |
| + this PR | **0** |

**#3916 measured 3, and it was right when it measured.** It was based on `d75d6c4fc`; between that base and its merge, **#3915's performance rewrite of the channels runtime landed** — 12 `Gsharp.Runtime.Channels` files, +1510 lines — and added four artifacts nobody saw, because `Channels` is deliberately outside the PR guard. That is the #3831/#3896/#3905 hazard happening a fourth time, on the one app the guard does not cover.

## The six root causes

Three of the four new ones and one of the two carried over are the same family: **issue #658's type-argument erasure deciding a question it cannot answer.** A same-compilation type or an open type parameter has no CLR `Type` while binding, so it rides through as `System.Object`; that is load-bearing and harmless by value, and wrong wherever the symbolic type is the answer.

**1. gsc — a range slice over an open `Span[T]` kept its element type.** (`Chan{T}.Batch.gs`, `Chan{T}.gs` — GS0154 + GS0155.) The slice member is found by reflecting on the receiver's ClrType, so inside `class Holder[T]` a `Span[T]` reports `Span<object>` as `Slice`'s return type, and `destination[first..]` produced a `Span[object]` that no longer fitted the `Span[T]` slot it was sliced from. Every BCL slice shape returns the receiver's own type, so when the reflected result is identical to the receiver's CLR type the answer is the receiver's symbolic type. `destination.Slice(1)` — an ordinary call — was always right, and `Span[int32]` was always right; both are controls in the test.

**2. gsc — a tuple type argument is symbolic when an element is.** (`ParkedNode{T}.gs` — GS0187 + GS0494.) `MemberLookup.IsSymbolicTypeArgument` had no tuple arm, so `class OpReceiveNode[T] : IValueTaskSource[(Value T, Ok bool)]` was verified against the ERASED `IValueTaskSource<ValueTuple<object, bool>>` and could not satisfy its own `GetResult` slot. The emitter's counterpart already recursed into tuples (#1902), so binder and emit disagreed — the doc comment claiming they mirror each other was stale. `SameTypeSymbol` needed the matching arm for the same reason (`TupleTypeSymbol.BuildClrType` returns null as soon as one element cannot be projected — the #3748/#3910 hole, here on the conformance side). The identical declaration over concrete element types always worked, and is the control.

**3. gsc — a by-ref argument over a same-compilation user class is not a `ref object`.** (`Chan{T}.gs` — GS0156.) `Interlocked` declares both `Exchange<T>(ref T, T) where T : class` and the non-generic `Exchange(ref object?, object?)`. Erasure made `&receiveNodePool` an exact IDENTITY match for the non-generic overload, so it won and the call's type became `object`. **csc never considers it: a `ref` argument requires exact type identity.** A dedicated sentinel now marks such an argument, a candidate filter drops any overload whose parameter at that slot is a by-ref over a non-generic type (applied on the OPEN candidates, because a closed generic's `ref T` has already become `ref object`), and the type argument is recovered symbolically.

> **The latent bug #3916 flagged beside this one does not exist, and I can show it.** `System.Threading.Volatile` has **no** non-generic `object` overload — only `Read<T>(ref T)` / `Write<T>(ref T, T)` — so `Volatile.Write(&receiveNodePool, node)` never had a wrong overload to bind to. Enumerated at runtime, and the pool round-trips correctly under ILVerify. Nothing to file.

**4. gsc — `@AllowNull` on a parameter.** (`Chan{T}.gs` — GS0154.) `[AllowNull] T value` is C#'s "this INPUT accepts null though its type is not nullable". cs2gs widens a REFERENCE-typed `[AllowNull]` declaration to `T?` (#3694) so the two contracts agree, but an unconstrained type parameter has no such widening, so `ReceiveResult<T>([AllowNull] T value, bool ok)` translates verbatim to `init(@AllowNull value T, …)` and the annotation is the whole contract. gsc had no `AllowNull` support at all.

> The fix is **annotation-only**, which is the trap #3916 backed out of. The parameter's declared type is never widened — a value-type `T` would then emit a `Nullable<T>` signature slot where `T` is expected, wrong IL that ILVerify does not catch because the assembly is internally consistent and simply describes a different method. `AllowNullParameter_EmitsTheBareTypeNotNullable` reads the ctor signature back out of the emitted metadata and asserts `!0`. The relaxation is also gated on the source nullable being representation-identical to its underlying, so `int32?` at an `@AllowNull int32` slot stays rejected — as it is in C#.

**5. gsc — narrowing lost a nested assignment's value.** (`SelectRandom.gs` — GS0116.) `order = (buffer = new int[…])` types as `buffer`'s declared `[]?int32`, so the outer assignment looked like it stored a nullable and `order` stayed nullable — while the identical `order = [8]int32` narrowed. The value of an assignment expression IS the value assigned, so this is a pure look-through applying the same test to the inner operand.

**6. gsc — an inherited interface slot had no emitted handle.** (`ArmDescriptor.gs` — GS9998, an emitter crash.) ADR-0174's `ISendSelectableCore[T] : ISelectableCore[T]` declares `Deregister` on the BASE; the resolver only searched the receiver interface's own definition, so the call fell through to the bare-MethodDef lookup and threw. Parenting the MemberRef at the derived interface would have been equally wrong, so both halves move together.

**Root 3's premise correction is worth carrying forward** and so is **#3916's third item**: `SelectRandom.gs:59` is a real defect after all, not a polish artifact. The polish pass never touched that file — instrumented and confirmed, 26 stripped sites, none in `SelectRandom.gs`. It did **not** self-heal, and would not have.

## Layer judgement

**All six are gsc.** In each case the C# is ordinary and legal, csc accepts it, and gsc's own neighbouring path already got it right: `.Slice()` keeps `Span[T]` where `[a..]` did not; the emitter recurses into tuples where the binder did not; `ref` identity is C#'s rule; a direct assignment narrows where the nested one did not; a slot the receiver interface declares itself resolves where an inherited one did not. These are wrong for every G# program, not just this app. **No source edits** — I did not touch #3882's or #3915's C#.

## Evidence

- **Nine executing cases** across three new test classes: each **compiles, ILVERIFIES and RUNS** the emitted program and asserts observable behaviour — the slice really is the tail of the same buffer, both interface slots dispatch separately on the same object, the pool actually pools and the rent actually clears the slot, the nil actually reaches the field, the swap actually swaps, the inherited slot actually runs. Both checks are load-bearing on this effort: an executing test has passed a broken build that only ILVerify caught, and an ILVerify-clean build has produced a wrong runtime answer.
- **Controls that passed before the fix** are in every case (ADR-0154 discrimination), plus two negative tests that must fail both ways.
- `Issue3907*`: **22 passed / 0 failed** (10 new + the 12 from #3914/#3916).

## What this does NOT do

**`Channels` is not green.** Its compile stage passes for the first time, and the ILVERIFY stage — never before reached — now reports **19 findings** over roughly eight distinct emitter defects (open-vs-closed generic self-reference in `.ctor`, an erased lambda adapter typed `Task` where `ValueTask` is expected, `ReceiveResult` ref-vs-value on the stack, `Memory<T>` vs `T[]`). Those were invisible while the app could not compile. They are a separate PR.

`src/Sdk/Gsharp.Runtime.Channels` therefore stays **out** of `run-cs2gs-selfmig-pr-guard.sh`'s `guard_apps`: the script's own note explains why a day-one-red guard is worse than none, and the guard runs translate → compile → **ilverify** → test-parity.

Refs #3907, #3501, #3915, #3916, #658, #1902, #3694, #3748, ADR-0155, ADR-0172, ADR-0174.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ConversionClassifier.cs
+++ b/src/Core/CodeAnalysis/Binding/ConversionClassifier.cs
@@ -706,6 +706,15 @@ internal sealed class ConversionClassifier
 
         if (!conversion.Exists)
         {
+            // Issue #3907: the target parameter is annotated `@AllowNull`, so
+            // a `T?` argument is exactly what it declares itself willing to
+            // receive. See AcceptsNilAnnotatedArgument for why this is
+            // annotation-only and why it cannot reshape the emitted call.
+            if (AcceptsNilAnnotatedArgument(callParameter, expression.Type, type))
+            {
+                return expression;
+            }
+
             // Issue #1017: a user-defined conversion operator declared on a
             // same-package struct/class is modelled as a static op_Implicit /
             // op_Explicit FunctionSymbol — those types have no reflectible
@@ -2506,6 +2515,61 @@ internal sealed class ConversionClassifier
     internal static bool IsNumericReturnWidening(TypeSymbol source, TypeSymbol target)
     {
         return IsNumericReturnWideningCore(source, target);
+    }
+
+    /// <summary>
+    /// Issue #3907: true when <paramref name="parameter"/> is annotated
+    /// <c>@AllowNull</c> and <paramref name="sourceType"/> is the nullable
+    /// spelling of the parameter's own <paramref name="targetType"/> — i.e.
+    /// the argument is precisely the <c>nil</c>-carrying input the annotation
+    /// declares the parameter accepts.
+    /// </summary>
+    /// <remarks>
+    /// <para>Two properties make this safe, and both are load-bearing.</para>
+    /// <para><b>It is annotation-only.</b> The parameter's own
+    /// declared type is never widened. Widening it would make a
+    /// value-type <c>T</c> emit a <c>Nullable&lt;T&gt;</c> signature slot where
+    /// <c>T</c> is expected — wrong IL that ILVerify does not catch, because
+    /// the metadata is internally consistent and simply describes a different
+    /// method. Keeping the slot at <c>T</c> also preserves the asymmetry the
+    /// attribute exists for: reads of the parameter inside the body still see
+    /// a non-nullable <c>T</c> and still need the author's <c>!!</c>.</para>
+    /// <para><b>It never reshapes the emitted call.</b> The relaxation is
+    /// gated on <see cref="NullableLifting.IsAnyValueTypeNullable"/> being
+    /// false, so the source nullable is one whose runtime representation is
+    /// identical to its underlying (ADR-0001 §Phase 3.C.1: for a reference
+    /// type or an unconstrained type parameter, <c>T?</c> relays <c>T</c>'s
+    /// <c>ClrType</c> and the IL value is the same reference either way). A
+    /// value-type nullable — <c>int32?</c> into an <c>@AllowNull</c>
+    /// <c>int32</c> slot — really is <c>Nullable&lt;int32&gt;</c> on the
+    /// stack, is rejected by C# too, and stays rejected here. The underlying
+    /// must also be IDENTICAL to the target rather than merely convertible to
+    /// it, so this admits exactly <c>T?</c> at a <c>T</c> slot and no other
+    /// shape; the argument is returned unchanged, with no synthesized
+    /// conversion node.</para>
+    /// </remarks>
+    /// <param name="parameter">The target parameter, or <see langword="null"/> when not binding a call argument.</param>
+    /// <param name="sourceType">The argument's static type.</param>
+    /// <param name="targetType">The parameter's (substituted) declared type.</param>
+    /// <returns><see langword="true"/> when the argument is accepted as-is.</returns>
+    internal static bool AcceptsNilAnnotatedArgument(
+        ParameterSymbol? parameter,
+        TypeSymbol? sourceType,
+        TypeSymbol? targetType)
+    {
+        if (parameter is not { AllowsNullArgument: true }
+            || targetType == null
+            || targetType == TypeSymbol.Error
+            || sourceType is not NullableTypeSymbol nullableSource
+            || NullableLifting.IsAnyValueTypeNullable(nullableSource))
+        {
+            return false;
+        }
+
+        var underlying = nullableSource.UnderlyingType;
+        return underlying != null
+            && underlying != TypeSymbol.Error
+            && Conversion.Classify(underlying, targetType).IsIdentity;
     }
 
     // ----- Private helpers (kept here because they are only used by methods in this class) -----
--- a/src/Core/CodeAnalysis/Binding/DeclarationBinder.Constructors.cs
+++ b/src/Core/CodeAnalysis/Binding/DeclarationBinder.Constructors.cs
@@ -616,6 +616,14 @@ internal sealed partial class DeclarationBinder
                 continue;
             }
 
+            // Issue #3907: an `@AllowNull` base-constructor parameter declares
+            // that it accepts a nil-carrying input.
+            if (ConversionClassifier.AcceptsNilAnnotatedArgument(parameter, argument.Type, parameter.Type))
+            {
+                convertedArgs.Add(argument);
+                continue;
+            }
+
             if (argument.Type != parameter.Type
                 && !Conversion.Classify(argument.Type, parameter.Type).IsImplicit
                 && !ExpressionBinder.IsImplicitConstantNarrowingArgument(argument, parameter.Type))
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs
@@ -3102,7 +3102,7 @@ internal sealed partial class ExpressionBinder
             lengthOf,
             statements);
 
-        var returnType = TypeSymbol.FromClrType(sliceMethod.ReturnType);
+        var returnType = SliceResultType(target, sliceMethod.ReturnType);
         var call = new BoundImportedInstanceCallExpression(
             null,
             srcRef,
@@ -3116,8 +3116,46 @@ internal sealed partial class ExpressionBinder
     private BoundExpression BindRangeIndexerSlice(BoundExpression target, RangeExpressionSyntax range, PropertyInfo indexer)
     {
         var rangeValue = BuildSystemRangeValue(range);
-        var resultType = TypeSymbol.FromClrType(indexer.PropertyType);
+        var resultType = SliceResultType(target, indexer.PropertyType);
         return new BoundClrIndexExpression(range, target, indexer, ImmutableArray.Create(rangeValue), resultType);
+    }
+
+    /// <summary>
+    /// Issue #3907: the static type of a range slice (<c>src[a..b]</c>) over an
+    /// imported slice shape.
+    /// </summary>
+    /// <remarks>
+    /// <para>The slice member is discovered by reflecting on the receiver's
+    /// <see cref="TypeSymbol.ClrType"/>, and issue #658 erases a type argument
+    /// that has no CLR representation while binding — so inside
+    /// <c>class Holder[T]</c> a <c>Span[T]</c> receiver reflects as
+    /// <c>Span&lt;object&gt;</c> and <c>Slice</c> reports that as its return
+    /// type. Taking the reflected answer verbatim threw the symbolic <c>T</c>
+    /// away: <c>destination[1..]</c> produced a <c>Span[object]</c> that no
+    /// longer converted back into the <c>Span[T]</c> slot it was sliced
+    /// from.</para>
+    /// <para>Every slice shape in the BCL — <c>Span&lt;T&gt;</c>,
+    /// <c>ReadOnlySpan&lt;T&gt;</c>, <c>Memory&lt;T&gt;</c>,
+    /// <c>ReadOnlyMemory&lt;T&gt;</c>, <c>string</c> — yields the RECEIVER'S
+    /// OWN type, so when the reflected result is identical to the receiver's
+    /// CLR type the symbolic answer is just the receiver's symbolic type,
+    /// which still carries <c>T</c>. Deliberately guarded on that identity:
+    /// a slice member returning anything else keeps the reflected answer
+    /// unchanged, and a receiver whose type argument is already concrete
+    /// (<c>Span[int32]</c>) gets back the same symbol it had before.</para>
+    /// </remarks>
+    /// <param name="target">The sliced receiver.</param>
+    /// <param name="reflectedResult">The slice member's reflected result type.</param>
+    /// <returns>The slice expression's static type.</returns>
+    private static TypeSymbol SliceResultType(BoundExpression target, Type reflectedResult)
+    {
+        var targetType = target.Type;
+        if (targetType?.ClrType != null && targetType.ClrType.IsSameAs(reflectedResult))
+        {
+            return targetType;
+        }
+
+        return TypeSymbol.FromClrType(reflectedResult);
     }
 
     // Issue #1016/#1022/#1038: construct a `System.Range` value from a range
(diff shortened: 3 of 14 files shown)


DavidObando:
## The dependent cascade clears — measured, not predicted

#3907's most important finding was that the ten apps referencing `Gsharp.Runtime.Channels` carry a **byte-identical** artifact set that is literally `Channels`' own errors, so they clear **if and only if `Channels` compiles clean**. That was a prediction. It is now an observation.

Same isolated pipeline, keeping `Channels` plus a dependent and its ProjectReference closure, `--exclude` for everything else:

| app | on `origin/main` (#3907's cascade proof) | with this PR |
|---|---|---|
| `src/Compiler` | 19 compile-error artifacts, all naming `Gsharp.Concurrency` symbols | **0 — translate / compile / ilverify / test-parity all PASS** |
| `src/Core` | — | **0 — all four stages PASS** |
| `src/Analyzers/InternalAnalyzers` | — | **0 — all four stages PASS** |
| `src/Analyzers/…Analyzers.Testing` | — | **0 — all four stages PASS** |

The dependents auto-reference the migrated channels runtime and recompile its sources, so `Channels` reaching zero compile errors is exactly what they needed. Nothing in this PR touches those apps' own translation.

### One app in that run is NOT a cascade victim, and it is worth being precise about

`test/Runtime.Channels.Tests` still reports 15 compile errors — but every one of them is in **its own** emitted `.gs`, not in `Channels`':

```
GS0159 ChannelBatchExtensionsTests.gs   Cannot find function AsTask.          (×7)
GS0155 SelectWaiterStressTests.gs       Cannot convert 'int32' to 'Xunit.Record'.
GS0122/GS0124 GoroutineWorkItemTests.gs a lambda returning a value from a void body
GS0129 ChannelBatchExtensionsTests.gs   '+=' not defined for 'int32' and '?'
GS0557 SelectWaiterTests.gs             'case cancelled' with no ambient context
GS0558 BlockingTests.gs                 a suspending call from a non-suspending caller
```

Those are that project's own translation gaps (ADR-0174 test surface), not the shared cascade, and they are out of scope here.

### And `Channels` itself is still not green

Its **compile** stage passes for the first time. Its **ilverify** stage — which no run had ever reached — now reports 19 findings over roughly eight distinct emitter defects. Those were invisible while the app could not compile, and they are a separate PR. `Channels` therefore stays out of `run-cs2gs-selfmig-pr-guard.sh`'s `guard_apps`: that guard runs translate → compile → **ilverify** → test-parity, so adding it today would give exactly the day-one-red guard the script's own note warns against. The line goes back when the ilverify findings clear.

DavidObando:
## Failing on the base, honestly labelled

The eleven `src/Core` files reverted to this PR's base (`64401d0c1`), everything else — the tests included — left exactly as it is, rebuilt, re-run:

| case | on base | why |
|---|---|---|
| `RangeSliceOverAnOpenSpan_KeepsItsElementType` | **FAIL** | root 1 |
| `ExplicitInterfaceImplementationOverATupleTypeArgument_BindsAndDispatches` | **FAIL** | root 2 |
| `ByRefArgumentOverAUserClass_PicksTheGenericOverload` | **FAIL** | root 3 |
| `AllowNullParameter_AcceptsANilCarryingArgumentAndRuns` | **FAIL** | root 4 |
| `AllowNullParameter_EmitsTheBareTypeNotNullable` | **FAIL** | root 4 |
| `NestedAssignmentValue_NarrowsTheOuterTargetLikeADirectOne` | **FAIL** | root 5 |
| `InheritedGenericInterfaceSlot_EmitsAndDispatches` | **FAIL** | root 6 |
| `WithoutTheAnnotation_ANilCarryingArgumentIsStillRejected` | pass | guard rail — must pass BOTH ways |
| `AllowNullOnAValueTypeParameter_StillRejectsAValueTypeNullable` | pass | guard rail — must pass BOTH ways |
| `ANullableSlotWithoutTheAssignment_IsStillRejected` | pass | guard rail — must pass BOTH ways |

**7 of 10 fail on base; the 3 that pass are the deliberate anti-over-reach guards.** Each of the seven also carries an in-test CONTROL that passed before the fix — `.Slice(1)` and `Span[int32]`, a named tuple over concrete elements, the same pool over a BCL element type, a slot the receiver interface declares itself, a direct (non-nested) assignment — so a mutant that flips the rule ON unconditionally is caught too.

On the branch: `Issue3907*` **22 passed / 0 failed** (10 new + the 12 from #3914/#3916).

## Full suites

- **`Core.Tests`: 8888 passed / 0 failed** locally. This matters more than usual: five of the six roots change type resolution, overload applicability or narrowing, whose blast radius is the whole binder, and the sixth is in the emitter.
- `Compiler.Tests` is covered by CI's sharded `tests (compiler-*)` jobs rather than a local full run; `tests (core-binding)`, `tests (core-remainder)`, `tests (compiler-conformance)`, `tests (compiler-emit-remainder)`, `ilverify`, `e2e`, `cs2gs-corpus`, `cs2gs-code-exploder` and `tests (cs2gs-issue34)` — the shard that caught #3916's over-scoped probe — are the ones I would look at first.

The head this comment describes differs from the CI-validated tree above by one **comment-only** commit (`6ca69f3b9`), which corrected a stale sentence about where the by-ref sentinel's restriction lives; CI re-runs on it.

DavidObando:
## Correction to the cascade table above, and the before/after I actually measured

The table in my previous comment listed `src/Core`, `src/Analyzers/InternalAnalyzers` and `src/Analyzers/…Analyzers.Testing` alongside `src/Compiler` as apps that "clear". **That was sloppy of me: those three do not reference `Gsharp.Runtime.Channels` at all** — they were green on the base too, and this PR does not change them. Only `src/Compiler` in that set is a cascade dependent.

So I re-ran the same five apps with **only the eleven `src/Core` files this PR touches reverted to `64401d0c1`**, so the two runs differ by exactly the fix and nothing else in the tree:

| app | on the base | with this PR |
|---|---|---|
| `src/Sdk/Gsharp.Runtime.Channels` | 7 compile errors — **compile FAILED** | **0 compile errors — compile PASSED** (ilverify now fails, 19; see below) |
| `src/Compiler` | 6 compile errors — **compile FAILED** | **0 — translate / compile / ilverify / test-parity all PASS** |
| `src/Core` | 0 — green | 0 — green (not a dependent) |
| `src/Analyzers/InternalAnalyzers` | 0 — green | 0 — green (not a dependent) |
| `src/Analyzers/…Analyzers.Testing` | 0 — green | 0 — green (not a dependent) |

### And `src/Compiler`'s six ARE `Channels`' own — verbatim

```
GS0155  Cannot convert type 'System.Span[object]' to 'System.Span[T]'.
GS0154  Parameter 'destination' requires 'System.Span[T]' but was given 'System.Span[object]'.
GS0187  Class 'OpReceiveNode' does not implement 'IValueTaskSource[(Value T, Ok bool)].GetResult(int16)'.
GS0494  'IValueTaskSource`1[(Value T, Ok bool)]' has no member 'GetResult' matching this declaration…
GS0156  Cannot convert type 'object' to 'OpReceiveNode[T]'.
GS0116  Type '[]?int32' is not indexable.
```

Every one names a `Gsharp.Concurrency` symbol, and every one has an empty `gsFile` — they anchor in the sibling project's emitted files, not `src/Compiler`'s own. That is the cascade mechanism #3907 described, caught in the act: `src/Compiler` has no defect of its own, it simply recompiles `Channels`' sources. Roots 1, 2, 3 and 5 of this PR account for all six.

A second dependent (`test/Core.Tests`) is measuring now; I will post it when it lands rather than assert it.

DavidObando:
## CI is green on the pushed head: **37 pass / 0 fail** (4 skipping)

Including the ones that matter most for a six-root binder-and-emitter change:

| job | why it matters here |
|---|---|
| **hot-core translation guard** | the CI counterpart of `build/run-cs2gs-selfmig-pr-guard.sh` — translate → compile → ilverify → test-parity over `src/Core`, `InternalAnalyzers`, `Cs2Gs.CodeModel`, `Cs2Gs.Translator`. Five of the six roots change type resolution, overload applicability or narrowing; this is the job that would catch them taking the gate down by existing (#3831 / #3896 / #3905). |
| `tests (core-binding)`, `tests (core-remainder)` | the binder's own suites |
| `tests (compiler-conformance)`, `tests (compiler-emit-remainder)` | root 6 is in the emitter |
| `ilverify` | the sample corpus re-verified after an emitter change |
| `tests (cs2gs-issue34)` | the shard that caught #3916's over-scoped probe |
| `cs2gs-corpus`, `cs2gs-oahu`, `cs2gs-code-exploder`, `e2e`, `nullable-hygiene` | the rest of the translation surface |

Locally: `Core.Tests` **8888 passed / 0 failed**; `Issue3907*` **22 passed / 0 failed**.

DavidObando:
## Second dependent, measured both ways: `test/Core.Tests` 6 → 1, and the 6 were `Channels`' own

Same method — only the eleven `src/Core` files this PR touches reverted to `64401d0c1`, everything else identical:

| app | on the base | with this PR |
|---|---|---|
| `src/Sdk/Gsharp.Runtime.Channels` | 7 — compile FAILED | **0 — compile PASSED** (ilverify fails, 19) |
| `test/Core.Tests` | **6 — compile FAILED** | **1 — compile still fails, on something else entirely** |
| `src/Core` | 0 — green | 0 — green |

`test/Core.Tests`' six on base are **byte-identical to `src/Compiler`'s six**, and to six of `Channels`' own seven:

```
GS0155  Cannot convert type 'System.Span[object]' to 'System.Span[T]'.
GS0154  Parameter 'destination' requires 'System.Span[T]' … 'System.Span[object]'.
GS0187  Class 'OpReceiveNode' does not implement 'IValueTaskSource[(Value T, Ok bool)].GetResult(int16)'.
GS0494  'IValueTaskSource`1[(Value T, Ok bool)]' has no member 'GetResult' matching …
GS0156  Cannot convert type 'object' to 'OpReceiveNode[T]'.
GS0116  Type '[]?int32' is not indexable.
```

That is the cascade, twice over, from two unrelated dependents. **All six clear.**

### The one that remains is NEW TO VIEW, and I am not going to claim more than that

```
GS9998 NotSupportedException: Conversion from
  'KeyValuePair`2[string, ValueTuple`3[int, int, List`1[ValueTuple`3[int,int,int]]]]' to
  'KeyValuePair`2[string, (Rid int32, CatchHandlerPlusOne int32,
                           Awaits List`1[(Yield int32, Resume int32, ResumeRid int32)])]'
  is not yet supported by the emitter.
```

It comes from `Adr0174AsyncDebugInfoTests`' `Dictionary<string, (int Rid, int CatchHandlerPlusOne, List<(int, int, int)> Awaits)>` — a nested NAMED tuple as a generic argument, where the erased CLR shape and the symbolic named shape have to be reconciled at emit.

**On the base this app's compile failed at the Channels errors, so the emitter never ran and this diagnostic could not appear.** It is therefore newly *visible*; whether it is newly *caused* is not settled by these two runs, and I could not reproduce it in a standalone 15-line program with the same dictionary shape. What I can say precisely:

- none of the six roots touches dictionary, `foreach` or tuple **emit**;
- `MemberLookup.IsSymbolicTypeArgument` — the change closest to tuples — has exactly one caller, `TryGetSymbolicClrGenericInterface`, used only by **interface conformance verification**; `KeyValuePair` is not an interface;
- `src/Core` itself, which contains the emitter, passes translate / compile / **ilverify** / test-parity identically on both sides;
- `test/Core.Tests` is **not** in `selfmig-baseline.json`'s `greenApps`, so it was already red and this moves no banked app either way.

The honest summary is: this is the same pattern as root 6 in the PR description — a defect that only becomes reachable once the ones in front of it are fixed — and it belongs with the 19 `Channels` ilverify findings in the next round of work.

Issue #3933, opened 2026-09-05
gsc: a capture-free lambda inside a GENERIC type is hoisted to <Program> and loses access to the encloser's private members (#1469's skip) — the last Channels ilverify finding


DavidObando:
Root **E** of #3932's triage, and the **last** ilverify finding standing between migrated `src/Sdk/Gsharp.Runtime.Channels` and green (that app is now `FAIL(1)`, down from 19).

```
[FieldAccess]: [Gsharp.Concurrency.<Program>::<lambda_erased2>(object, CancellationToken)]
[offset 0x00000015] Field is not visible.
```

## What happens

`ParkedNode<T>.RegisterCancellation` passes a **capture-free** (`static`) lambda to `token.UnsafeRegister`, and that lambda reads `ParkedNode<T>`'s `private readonly Chan<T> owner`:

```csharp
registration = token.UnsafeRegister(
    static (state, token) =>
    {
        var node = (ParkedNode<T>)state!;
        node.owner.CancelParkedNode(node, token);   // private field
    },
    this);
```

gsc hoists it to a static method on the top-level `<Program>` host:

```
.method ... void '<lambda_erased2>'<T> (object state, valuetype CancellationToken token)
    IL_000e: castclass class Gsharp.Concurrency.ParkedNode`1<!!T>
    IL_0015: ldfld class Gsharp.Concurrency.Chan`1<!!0> class Gsharp.Concurrency.ParkedNode`1<!!T>::owner
```

`<Program>` is outside `ParkedNode<T>`'s accessibility domain, so the `ldfld` is unverifiable and would throw `FieldAccessException` at runtime. csc avoids this by hosting the lambda in a display class **nested inside the declaring type**, which shares its accessibility domain.

## Why it is still open

**This is not an unknown problem — it is a known fix with one case deliberately left out.** #1469 established exactly this rule for capture-free lambdas, and #1512 extended the capturing path to generic enclosers. The remaining hole is in `ClosureEmitter.SynthesizeClosures`:

```csharp
if (literal.Function.IsGeneric
    || literal.Function.LexicalEnclosingType is not StructSymbol zeroCaptureEnclosing
    || !zeroCaptureEnclosing.TypeParameters.IsDefaultOrEmpty)   // <- generic enclosers skipped
{
    continue;   // keep the <Program> static placement
}
```

with the reason stated in the code:

> Generic enclosers are skipped because a nested type would have to re-declare the encloser's type parameters, which this synthesis does not model.

[...]

Issue #3932, opened 2026-09-05
gsc: the newly-reachable ilverify wall in migrated Gsharp.Runtime.Channels — 19 findings, 5 root causes, all generic-context emit sites


DavidObando:
Follow-on to #3907, which closed the compile wall for migrated `src/Sdk/Gsharp.Runtime.Channels` (194 → 0 across #3910/#3914/#3916/#3927) and thereby made the app's **ilverify** stage reachable for the first time. It reports **19 findings**.

**Scale: deduplicated ilverify artifacts** — one triage artifact per distinct error code + failing-method skeleton, the same unit #3914/#3916/#3927 used for their 27/11/3/7. Re-measured on `efd1b1a91` with a single-app targeted migration (`Gsharp.Runtime.Channels` is a ProjectReference leaf, so it reproduces alone): translate **PASS**, compile **PASS**, ilverify **FAIL(19)**.

## Triage: 19 findings, 5 root causes

| # | cause | findings | layer | new/known/cascade | tractability |
|---|---|---:|---|---|---|
| A | a chained `init(...)` inside a **generic** aggregate emits the bare MethodDef, which names the OPEN definition — so `this` is never marked initialized and the argument slots stay uninstantiated | **9** | gsc emitter | new; `newobj` on the same type has always used the self TypeSpec | small |
| B | an argument at an **imported generic ctor's** own type-parameter slot converts to the erased `object`, emitting `box !!T` in front of a `newobj` whose parent TypeSpec is correctly symbolic. Also drags a tuple's element type to `object` | **7** | gsc binder | new; the imported generic **method** slot (`List[T].Add(v)`) has always been correct | small |
| C | a **`shared` (static)** async method on a generic type has no `ReceiverType`, so its state machine reifies at arity **0** with every hoisted field carrying a dangling `!0` | 1 | gsc emitter | new; the instance form and the iterator sibling both consult `StaticOwnerType` | small |
| D | at an **extension-call** argument whose parameter type still *contains* a type parameter, conversion is skipped entirely — right for a bare open `T`, wrong for `Memory[T]`, which emits as a real constructed slot, so `[]T` arrived with no `op_Implicit` | 1 | gsc binder + emitter | new; the same call with a **concrete** element has always emitted the operator | medium |
[...]

Issue #1540, opened 2026-07-01
gsc: generic type parameter not convertible to object (GS0155 'Cannot convert type T to object')


DavidObando:
While migrating **Oahu.Foundation** with cs2gs, delegate-marshalling and boxing code fails with `GS0155: Cannot convert type 'T' to 'object'`. Converting a generic type-parameter value to a genuine `object` slot (return, assignment, explicit cast, argument) should always succeed — it is a boxing conversion for value `T` and a reference conversion for reference `T` (C# `(object)t` / implicit).

## Minimal repro (FAILS today)
```gsharp
package Probe
class Ext {
    shared {
        func BoxImplicit[T](val T) object { return val }        // GS0155
        func BoxExplicit[T](val T) object { return object(val) } // GS0155
        func BoxAssign[T](val T) { let o object = val }          // GS0155
    }
}
```
gsc today: `error GS0155: Cannot convert type 'T' to 'object'` on all three. C#: `object o = val;` / `return (object)val;` are always valid.

## Root cause / design tension
`src/Core/CodeAnalysis/Binding/Conversion.cs` (~lines 99-112). The type-parameter conversion rule (`TypeParameterConvertsTo`, issue #1196) DELIBERATELY excludes `T -> object`:
> "NOTE: this deliberately does NOT cover `T -> object`. Open generic parameter slots (e.g. the `!0` element type of `List[T].Add(!0)`) are erased to the `object` singleton by the binder, so a `T -> object` rule here is indistinguishable from the identity argument conversion `T -> !0` and would make the emitter inject a spurious `box T`, producing invalid IL (the #1196 regression)."

So the difficulty: an ERASED `!0` slot and a GENUINE `System.Object` target are BOTH represented as `TypeSymbol.Object` at conversion-classification time. A blanket `T -> object` implicit rule reintroduces the #1196 spurious-box bug in erased-slot argument passing; but the current total exclusion wrongly rejects genuine `T -> object` conversions.

## Fix direction (investigate)
Distinguish a GENUINE `object` target (user-written `object` return type / local / cast / a parameter whose declared type is truly `System.Object`) from an ERASED type-parameter slot that merely presents as `object`. Options to evaluate:
- Carry a marker/flag on erased `!0` slots so the classifier can tell them apart from genuine `object`, then allow implicit/explicit `T -> object` for genuine object and keep identity (no box) for erased slots.
[...]

PR #3934, opened 2026-09-05
gsc: four generic-context emit defects take migrated Gsharp.Runtime.Channels from 19 ilverify findings to 1 (#3932)

DavidObando:
gsc: four generic-context emit defects take migrated Gsharp.Runtime.Channels from 19 ilverify findings to 1 (#3932)

#3907 closed the compile wall for migrated `src/Sdk/Gsharp.Runtime.Channels`
(194 -> 0 across #3910/#3914/#3916/#3927), which made that app's **ilverify**
stage reachable for the first time. It reported **19 findings**. Triage
(deduplicated ilverify artifacts, the same scale #3914/#3916/#3927 used)
collapsed them to **five** root causes; this fixes four of them, taking the app
to **FAIL(1)**.

All five are gsc, and in every one gsc's own neighbouring path already got it
right. That is the shape of this whole wall: not five unrelated gaps but five
siblings of already-correct code, each missed by one condition. No source edit
to the runtime's C# — it is ordinary, legal C# that csc compiles to verifiable
IL.

| # | cause | findings | layer |
|---|---|---:|---|
| A | a chained `init(...)` inside a GENERIC aggregate emits the bare MethodDef, naming the OPEN definition, so `this` is never marked initialized and the argument slots stay uninstantiated | 9 | emitter |
| B | an argument at an IMPORTED generic ctor's own type-parameter slot converts to the erased `object`, emitting `box !!T` before a `newobj` whose parent TypeSpec is correctly symbolic | 7 | binder |
| C | a `shared` (static) async method on a generic type has no `ReceiverType`, so its state machine reifies at arity 0 with every hoisted field carrying a dangling `!0` | 1 | emitter |
| D | at an extension-call argument whose target still CONTAINS a type parameter, conversion is skipped entirely — right for a bare open `T`, wrong for `Memory[T]`, which emits as a real constructed slot | 1 | binder + emitter |
| E | a capture-free lambda inside a GENERIC type is hoisted to `<Program>` and loses access to the encloser's private members | 1 | **filed as #3933** |

The already-correct siblings, one per cause: `newobj` on the same generic type
has always used the self TypeSpec; the imported generic METHOD slot
(`List[T].Add(v)`) has always taken the raw `T`; the instance async form and the
iterator path both consult `StaticOwnerType`; the same extension call with a
CONCRETE element type has always emitted `op_Implicit`.

B is worth calling out: it accounts for 7 findings across three shapes that
looked unrelated in the raw list (`found ref 'T' / expected value 'T'`,
`found ref 'ReceiveResult`1<T0>' / expected value`, and
`ValueTuple`2<object,bool>` where `ValueTuple`2<T0,bool>` belongs). All one
spurious `box !!T`.

## Evidence

`test/Compiler.Tests/Issue3932GenericEmitSitesTests.cs` — four tests, each of
which COMPILES, ILVERIFIES, RUNS and asserts on behaviour. Both checks are
load-bearing on #3501: an executing test has passed a build only ILVerify
caught, and an ILVerify-clean build has produced a wrong runtime answer.

Two also read the emitted metadata back, because "the token is a MethodDef, not
a TypeSpec-parented MemberRef" and "the state machine has arity 0" are encoding
facts a behavioural assertion cannot name:

- root A reads the chained call's token KIND out of the emitted method body
  (MemberReference on the generic type, MethodDefinition on the non-generic
  control);
- root C reads the state machine's generic-parameter COUNT out of the metadata
  (1 for the `shared` and instance forms on a generic encloser, 0 for the
  non-generic control).

Discrimination (ADR-0154): every test carries a control that passed BEFORE the
fix — the same construct in the non-generic / instance / concrete-element /
bare-`T` form — so a mutant that applies the new path unconditionally is caught
as surely as one that never applies it.

Anti-vacuity: all four FAIL on `origin/main` (verified by reverting only
`src/Core` to `origin/main` and re-running: 4 failed / 0 passed; restored: 4
passed / 0 failed).

## What this does NOT do

**`Channels` is not green** — root E (#3933) still fails ilverify — so
`src/Sdk/Gsharp.Runtime.Channels` stays OUT of
`build/run-cs2gs-selfmig-pr-guard.sh`'s `guard_apps`. That guard runs
translate -> compile -> ilverify -> test-parity, so adding it today gives
exactly the day-one-red guard the script's own note warns against. The line goes
in the moment #3933 lands; nothing else needs to change.

Also checked, per #3907's ask: `test/Core.Tests`' remaining `GS9998` emitter
`NotSupportedException` over a nested named tuple as a generic argument does NOT
share a root with any of the five, and is untouched by these fixes.

Refs #3932, #3933, #3907, #3501, #1469, #1512, #765, #1540, #3222, #671,
ADR-0087.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Nng28yiBdPVdeML7mSZphs

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ConversionClassifier.cs
+++ b/src/Core/CodeAnalysis/Binding/ConversionClassifier.cs
@@ -1351,6 +1351,23 @@ internal sealed class ConversionClassifier
             return true;
         }
 
+        // Issue #3932: an IMPORTED conversion operator whose declaring generic
+        // is closed over an OPEN type parameter — `[]T -> Memory[T]` — reaches
+        // neither branch here. The CLR branch above needs both sides to have a
+        // reflectible `ClrType` and an open `[]T` has none; the symbolic branch
+        // below only knows same-package operators. So the conversion was simply
+        // dropped and the emitter wrote the raw array into a `Memory[T]` slot.
+        // Resolve it off the target's OPEN definition instead, which is where
+        // `Memory`1::op_Implicit(!0[])` actually lives.
+        if (argument.Type != null
+            && expectedType != null
+            && argument.Type != TypeSymbol.Error
+            && TryResolveSymbolicImportedConversion(argument.Type, expectedType, out var openConvMethod))
+        {
+            converted = new BoundClrConversionCallExpression(null, argument, openConvMethod, expectedType);
+            return true;
+        }
+
         // Issue #1017: same-package user-defined implicit conversion operators
         // are modelled as static op_Implicit FunctionSymbols and have no
         // reflectible ClrType during binding, so resolve them symbolically.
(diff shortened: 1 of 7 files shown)


DavidObando:
CI: **37 pass / 0 fail** (3 skipping), including the **hot-core translation guard**, `cs2gs-corpus`, `cs2gs-oahu`, `ilverify`, `nullable-hygiene`, `test-partition` and every test shard (`compiler-*`, `core-binding`, `core-remainder`, `cs2gs-*`).

Local, on top of that: the four new tests **pass on this branch and fail on `origin/main`** — verified by reverting only `src/Core` to `origin/main` and re-running (4 failed / 0 passed), then restoring (4 passed / 0 failed). That is the anti-vacuity guard rail for all four roots.

Targeted migration of `src/Sdk/Gsharp.Runtime.Channels` (a ProjectReference leaf, so it reproduces alone), SDK nupkg repacked immediately before each migrate:

| | translate | compile | ilverify |
|---|---|---|---|
| `efd1b1a91` (base) | PASS | PASS | **FAIL(19)** |
| this branch | PASS | PASS | **FAIL(1)** |

The remaining finding is exactly root E (#3933), byte-identical to the one on base — no new shapes appeared behind the four that fell.

Issue #2668, opened 2026-07-21
gsc emitter: generated async lambdas call inaccessible methods


DavidObando:
Cycle 7 non-unsafe ILVerify MethodAccess in three Cli.Server Program generated async lambda MoveNext methods (lambda16/22/30). Generalize synthesized method accessibility/call target emission; exact fingerprints plus runtime proof.

Issue #2118, opened 2026-07-05
gsc: constrained interface-method call on a type parameter inside a lambda emits invalid IL (missing constrained. prefix, StackUnexpected)


DavidObando:
### Summary

gsc emits **invalid IL** when a lambda/closure body calls a **constrained interface method** on a value of a generic type parameter (`T` constrained to an interface). The `constrained.` prefix is omitted in the closure's method body, so a `T&`/`T` is left on the stack where the interface type is expected. The IL fails `ilverify` with `StackUnexpected [found ref '!0'][expected ref 'System.IComparable`1<!0>']`.

The **same call in a normal (non-lambda) method emits correctly** — the bug is specific to lambdas/closures, where the type-parameter constraint is apparently not propagated to the synthesized closure method body.

### Minimal repro (StackUnexpected)

```gsharp
package R
import System

func Build[T IComparable[T]]() (T, T) -> int32 {
    let f (T, T) -> int32 = (x T, y T) -> x.CompareTo(y)
    return f
}
```

`gsc /target:library` succeeds, but `ilverify`:

```
[StackUnexpected]: [R.<lambda1>(!0, !0)][offset 0x09][found ref '!0'][expected ref 'System.IComparable`1<!0>'] Unexpected type on the stack.
```

### Control (verifies clean) — same call, no lambda

```gsharp
package R
import System
open class C[T IComparable[T]] {
    func Cmp(x T, y T) int32 -> x.CompareTo(y)   // direct method: emits `constrained. !T` — OK
}
```

### Root cause (hypothesis)

Calling an interface method on a type-parameter-typed receiver requires the `constrained. !T` prefix before `callvirt IComparable<T>::CompareTo`. In a normal method gsc emits this correctly. In a lambda, the body is lowered into a synthesized closure method; the receiver's static type (the constrained type parameter) — or the knowledge that a constrained call is needed — is lost, so gsc emits the call without `constrained.`, expecting the receiver to already be the interface type. Fix: propagate the type-parameter constraint context into the closure lowering so the constrained interface call is emitted in the closure body exactly as in a normal method.

### Impact (Oahu.Decrypt)

`EnumerableExtensions` (`Refs #914` migration):

```csharp
var comparer = Comparer<TResult>.Create((x, y) => keySelector(x).CompareTo(keySelector(y)));  // TKey : IComparable<TKey>
```

[...]

PR #3936, opened 2026-09-05
gsc: nest a capture-free lambda's host inside a GENERIC encloser, taking migrated Gsharp.Runtime.Channels to zero ilverify findings (#3933)

### LINKED ISSUE #3933: gsc: a capture-free lambda inside a GENERIC type is hoisted to <Program> and loses access to the encloser's private members (#1469's skip) — the last Channels ilverify finding

DavidObando:
Root E of #3932's triage, and the last finding standing between migrated
`src/Sdk/Gsharp.Runtime.Channels` and green. A capture-free lambda lexically
inside a GENERIC type was hoisted to a static method on `<Program>`, which is
outside the encloser's accessibility domain, so reading the encloser's
`private` member emitted `FieldAccess: Field is not visible` and threw
`FieldAccessException` at the first call.

#1512's reification applies directly. #1469 established this rule (host the
lambda in a fieldless display class nested in the declaring type, as csc does
with `<>c`) and excluded generic enclosers because "a nested type would have
to re-declare the encloser's type parameters, which this synthesis does not
model". #1512 had already taught the CAPTURING sibling to do exactly that via
`SynthesizedClosureReifier` — and that reification is driven from
`SynthesizeDisplayClass`, which BOTH paths share and which is entirely
shape-independent of whether the class carries capture fields. Dropping the
one excluding condition routes the zero-capture host through machinery that
was already there.

One thing #1512's path never had to answer, and the only new code here: a
capturing closure always holds the encloser's type parameters in its own
capture fields, so signature-driven discovery finds them. A capture-free host
has no such guarantee — the lambda may reference only SOME of them, or reach
the encloser only through a path the collector does not walk (a `private
shared` call, which the emitter renders as `Reg`1<!0>`). Reified from
discovery alone that host has arity 0, the `!0` has no slot, and the result
ILVERIFIES CLEAN and throws `BadImageFormatException` at the first call —
measured, not assumed. So the encloser's own type-parameter list is passed as
a required seed. It is a no-op in the common case, because CollectOrdered
already emits class parameters ahead of method parameters by ordinal.

## Evidence

`test/Compiler.Tests/Emit/Issue3933ZeroCaptureGenericEncloserEmitTests.cs` —
four facets, each of which COMPILES, ILVERIFIES, RUNS and asserts on
behaviour, then reads the emitted metadata back (the host is nested in the
encloser at the encloser's arity, and no `<lambda*>` method is left on
`<Program>` — encoding facts no behavioural assertion can name):

- A the reported shape: private instance field of a generic encloser;
- B the encloser's parameters discovery does not find (only `U` of `[T, U]`,
  and #1469 facet C's no-reference lambda);
- C the private-`shared` call — the facet ILVerify cannot see;
- D #2668's async shape on a generic encloser, whose state machine now nests
  one level deeper still.

Discrimination (ADR-0154): every facet carries a control that passed BEFORE
the fix — the same lambda on a NON-generic encloser, which #1469 already
nested at arity 0 — so a mutant that nests unconditionally at the wrong arity
is caught as surely as one that never nests.

Anti-vacuity: all four FAIL on `origin/main` (verified by reverting only
`src/Core/CodeAnalysis/Emit/ClosureEmitter.cs` and re-running: 4 failed / 0
passed; restored: 4 passed / 0 failed). Three fail as ILVerify FieldAccess /
MethodAccess; facet C's discovery hole additionally fails as a runtime
BadImageFormatException with the seeding removed.

## Migration measurement

Targeted four-app PR-guard run (Channels plus three of its dependents),
before = `origin/main`'s ClosureEmitter, after = this change, everything else
identical:

| app | before | after |
|---|---|---|
| src/Sdk/Gsharp.Runtime.Channels | PASS/PASS/**FAIL(1)**/skip | PASS/PASS/**PASS**/**PASS** |
| src/Sdk/Gsharp.NET.Sdk | PASS/PASS/PASS/PASS | PASS/PASS/PASS/PASS |
| src/Sdk/Gsharp.HotReload.Runtime | PASS/PASS/PASS/PASS | PASS/PASS/PASS/PASS |
| bench/concurrency/clr/ClrBaseline | PASS/PASS/PASS/PASS | PASS/PASS/PASS/PASS |

The before-run's single finding is byte-for-byte #3933's:
`[FieldAccess]: [Gsharp.Concurrency.<Program>::<lambda_erased2>(object,
CancellationToken)][offset 0x00000015] Field is not visible.`

Channels reaches ZERO ilverify findings and goes green through test-parity.
No new stage opened behind ilverify — test-parity passed on its first run.

**A premise correction, stated loudly.** #3933/#3932's framing has Channels at
the root of an all-or-nothing cascade in which ten dependents carry a
byte-identical copy of its diagnostics. Measured, the three dependents above
were ALREADY fully green on `origin/main`, with Channels sitting at FAIL(1)
next to them. ilverify verifies an app's own assembly, so a dependent that
merely REFERENCES an unverifiable Channels does not inherit the finding; the
cascade those ten apps suffered was the COMPILE wall, and that closed with
#3907. What this change buys is Channels itself, which is exactly what
`guard_apps` needed.

## The guard line

`src/Sdk/Gsharp.Runtime.Channels` joins `guard_apps` in
`build/run-cs2gs-selfmig-pr-guard.sh`. It was deliberately absent through five
PRs because the guard runs translate -> compile -> ilverify -> test-parity and
the app had never been ilverify-green, so adding it earlier gave exactly the
red-from-day-one guard the script's own note warns against. That reason is now
gone, and the guard covers all three of the incidents it names instead of two.
It is a ProjectReference leaf, so the closure invariant still holds and the
job costs a couple of minutes.

## Follow-up, not done here

#2016's GS0468 ("local function cannot reference an enclosing type
parameter") is now MORE conservative than the emitter requires: its
generic-class-encloser arm exists precisely because such a lambda used to be
hoisted to `<Program>` without the type parameters, which is no longer true.
Relaxing it would rewrite the contract of an existing test
(`NonGenericLocalFunction_ReferencesEnclosingGenericClassTypeParameter_ReportsGS0468`),
so it is left alone and filed as its own question.

Refs #3933, #3932, #3934, #3907, #3501, #1469, #1512, #1477, #2016, #2118,
#2668, #2894, #3836.

Closes #3933.

### DIFF
--- a/build/run-cs2gs-selfmig-pr-guard.sh
+++ b/build/run-cs2gs-selfmig-pr-guard.sh
@@ -35,8 +35,9 @@
 # here means "the hot core still migrates", never "this PR migrates fine".
 # The banner printed at the end says so out loud on purpose.
 #
-# In particular it does NOT cover src/Sdk/Gsharp.Runtime.Channels, i.e. it
-# would NOT have caught #3905 — see the note under guard_apps.
+# As of #3933 it DOES cover src/Sdk/Gsharp.Runtime.Channels, so it now covers
+# all three of the incidents above rather than two — see the note under
+# guard_apps.
 #
 # Usage: run-cs2gs-selfmig-pr-guard.sh
 # Environment:
@@ -64,25 +65,30 @@ repo_root=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
 #   tools/cs2gs/Cs2Gs.CodeModel   Core -> CodeModel -> Translator, the chain
 #                                 #3836 names.
 #   tools/cs2gs/Cs2Gs.Translator  cs2gs is inside its own corpus (#3831).
+#   src/Sdk/Gsharp.Runtime.Channels  #3905's root, and a ProjectReference
+#                                 leaf, so it costs this job only a couple of
+#                                 minutes. ADDED BY #3933, see below.
 #
-# DELIBERATELY ABSENT, and the first thing to add back:
+# ADDED BY #3933. Channels was deliberately absent through five PRs because
+# this guard runs translate -> compile -> ilverify -> test-parity and the app
+# had never been ilverify-green: adding it earlier would have produced exactly
+# the red-from-day-one guard that gets ignored or disabled. That reason is now
+# gone. The app went 194 -> 0 compile artifacts (#3910/#3914/#3916/#3927,
+# closing #3907) and then 19 -> 1 -> 0 ilverify findings (#3934 for #3932's
+# roots A-D, #3933 for root E), and a targeted four-app run measured on
+# f7fde6d2f reads PASS/PASS/PASS/PASS on all four stages. That closes the
+# hazard this whole script exists for on the one app it did not cover — the
+# hazard has now fired four times (#3831, #3896, #3905, and #3915's rewrite
+# adding four unseen artifacts between #3916's base and its merge).
 #
-#   src/Sdk/Gsharp.Runtime.Channels — #3905's root, referenced by eight
-#   projects, and a ProjectReference leaf, so it would cost this job about
-#   four minutes and would make the guard cover all three incidents instead
-#   of two. It is left out only because it is RED ON MAIN TODAY (#3907: the
-#   #3905 crash fix un-blinded 327 diagnostics), and a guard that is red from
-#   day one gets ignored or disabled. Measured on 03b6e1e8d, adding it gives
-#   4/5 rather than 5/5 — a real failure, not a harness artifact, which is
-#   itself evidence this job detects the thing it is for. Add the line back
-#   the moment #3907 closes; nothing else needs to change.
-#
-# The same reasoning governs tools/cs2gs/Cs2Gs.Tests, which #3836 names as the
-# natural second step: it has ~9 known migration failures and would be red by
-# construction until those clear.
+# STILL DELIBERATELY ABSENT: tools/cs2gs/Cs2Gs.Tests, which #3836 names as the
+# natural next step. It has ~9 known migration failures and would be red by
+# construction until those clear — the same reasoning that kept Channels out
+# until today.
 guard_apps=(
   src/Analyzers/InternalAnalyzers/InternalAnalyzers.csproj
   src/Core/Core.csproj
+  src/Sdk/Gsharp.Runtime.Channels/Gsharp.Runtime.Channels.csproj
   tools/cs2gs/Cs2Gs.CodeModel/Cs2Gs.CodeModel.csproj
   tools/cs2gs/Cs2Gs.Translator/Cs2Gs.Translator.csproj
 )
--- a/src/Core/CodeAnalysis/Emit/ClosureEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/ClosureEmitter.cs
@@ -230,9 +230,25 @@ internal sealed class ClosureEmitter
                 // capture-bearing #1335 path does. Issue #2668 extends the same
                 // placement to async lambdas so their MoveNext methods retain
                 // access to private static members of the lexical encloser.
-                // Generic enclosers are skipped because a nested type would have
-                // to re-declare the encloser's type parameters, which this
-                // synthesis does not model.
+                // Issue #3933: generic enclosers used to be skipped here, because
+                // a nested type must re-declare the encloser's type parameters
+                // (ECMA-335 §II.10.3.1) and this synthesis did not model that.
+                // The consequence was the exact bug #1469 exists to prevent, just
+                // one type-parameter away: a capture-free lambda inside a GENERIC
+                // type that reads a `private` member of that type stayed on
+                // `<Program>`, was generic-promoted by #2118 so the field access
+                // read `Node`1<!!T>::secret`, and failed ILVerify with
+                // `FieldAccess: Field is not visible` / threw
+                // `FieldAccessException` at the first call. The capturing sibling
+                // has handled this since #1512 by reifying the display class over
+                // the encloser's type parameters before nesting it, and the
+                // reification (SynthesizedClosureReifier, driven from
+                // SynthesizeDisplayClass) is entirely shape-independent of whether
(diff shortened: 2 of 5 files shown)


DavidObando:
### Follow-on commit, and the guard proving itself on its first run

A second commit landed on this branch: **an ITERATOR lambda now nested in a
generic encloser must not count its type parameters twice**. It was caught by
the FULL `Compiler.Tests` suite, not by the targeted run — the first commit
turned `Issue2957FunctionLiteralIteratorTests.GenericReceiverStateMachineMemberSignatures_UseSourceTypeParameterOrdinals`
red. Reproduced standalone: a `sequence` lambda inside `Factory[T]`'s `make[U]`
built a state machine at arity 3 instead of 2 and threw `TypeLoadException`
("has a field of an illegal type") on the first call. See that commit for the
two causes (a double-counted enclosing chain, then an unaliased original
type parameter) — both were latent for #1512's CAPTURING placement and simply
had no test reaching them.

Worth stating plainly, because it is the evidence standard this effort runs on:
after the arity was corrected the assembly **ILVerified clean and still threw at
load**, and the intermediate state before that **ILVerified with a
`StackUnexpected` that only a run would have explained**. Compile + ILVerify +
run were all three load-bearing here, on the same defect, within one hour.

### Where it stands

| check | result |
|---|---|
| `Compiler.Tests` (full, local) | 4741 / 4741 |
| `Core.Tests` (full, local) | 8900 / 8900 |
| CI on this branch | **37 pass, 0 fail** (4 skipped) |
| **hot-core translation guard** (CI, WITH Channels) | **pass**, 29m19s |

That last row is the one that matters for the `guard_apps` line: the guard job
ran translate -> compile -> ilverify -> test-parity over the five-project hot
core **including `src/Sdk/Gsharp.Runtime.Channels`** and came back green. The
line is not a promise that Channels is ready for the guard; it is a measured
guard run that already covered it.

Issue #3940, opened 2026-09-05
gsc: a nested generic type returned from a generic method keeps its ENCLOSING type argument open at the call site


DavidObando:
Found by the #3939 emitter sweep, and independent of the two defects that sweep fixes.

When a generic method on a generic type returns a **nested** generic type of that same encloser, the call-site token's return signature substitutes the method's own type argument but leaves the **enclosing** type argument open.

```gs
package main
import System

class Outer[T] {
    class Inner[U] {
        var item T = default(T)
        var other U = default(U)
    }

    func Make[U](v T, w U) Inner[U] {
        let i = Inner[U]()
        i.item = v
        i.other = w
        return i
    }
}

let o = Outer[int32]()
let i = o.Make[string](4, "z")
Console.WriteLine("v=" + i.item.ToString() + "/" + i.other)
```

`gsc` compiles this without a diagnostic. Then:

```
[IL]: Error [StackUnexpected]: main.<Program>::<Main>$(string[])][offset 0x0000001A]
  [found ref '[main]main.Outer`1+Inner`1<int32,string>']
  [expected ref '[main]main.Outer`1+Inner`1<!0,string>'] Unexpected type on the stack.
[IL]: Error [StackUnexpected]: main.<Program>::<Main>$(string[])][offset 0x00000034]
  [found ref '!0'][expected ref 'object'] Unexpected type on the stack.
```

and at load:

```
System.BadImageFormatException: An attempt was made to load a program with an incorrect format. (0x8007000B)
   at main.<Program>.<Main>$(String[] args)
```

Note the asymmetry inside the single rendered type: `Inner`1<!0, string>` has the **method** parameter `U` already substituted to `string` while the **enclosing** parameter `T` stayed at `!0`.

## Scope

- Reproduces with **no** `shared`, **no** state machine, **no** lambda, so it is not the #3939 A/B family and not a `StateMachineEmitter` or `ClosureEmitter` concern. It is call-site / signature encoding for a nested generic type's enclosing type arguments.
- The nesting shape itself is fine when the inner type is non-generic: `class Outer[T] { class Inner { … } }` with an async method and an iterator method on `Inner` compiles, ilverifies and runs (verified — it is one of #3939's cleared pairs).
- The shape has prior art in the codebase: #3812 covers `Outer[T].Inner`'s *initializer* scope, including the both-levels-generic case, so binding is right and the gap is downstream of it.

[...]

Issue #2030, opened 2026-07-04
gsc: generic async functions cannot be emitted/run (state machine + hoisting gaps)


DavidObando:
## Summary

While fixing #2026 (generic async function/local-function calls losing `Task<T>` wrapping on the substituted return type, a binder-level bug), found that generic `async func`s cannot actually be compiled and run at all today, due to (at least) two separate, pre-existing gaps in the emit layer. These are unrelated to #2026's binder fix (which only fixes type-checking / the call-site observable return type) and are NOT fixed by that PR.

## Gap 1: state-machine synthesis fails (GS0190) whenever the declared inner return type is an open type parameter

```gsharp
package P
async func Foo[U](x U) U {
    return x
}
```

Even standalone (no caller, no nesting), this reports:

```
error GS0190: Could not synthesize the state machine for this async function; the awaitable's method-builder type could not be resolved.
```

Root cause: `AsyncStateMachineTypeBuilder.ResolveAsyncReturnClrType` (`src/Core/CodeAnalysis/Lowering/Async/AsyncStateMachineTypeBuilder.cs`) needs a real CLR `System.Type` for `kickoff.Type` (here, the open method type parameter `U`) to resolve `Task<T>`/`AsyncTaskMethodBuilder<T>` via reflection. It special-cases same-compilation `struct`/`interface`/`enum` (erasing to `object`), and nullable-of-those, but has no branch for a bare/wrapped `TypeParameterSymbol` — so it returns `null`, and `AsyncEmitPrecheck` (`src/Core/CodeAnalysis/Lowering/Async/AsyncEmitPrecheck.cs`) reports GS0190.

Simply erasing to `object` (like the struct/interface/enum case) is not sufficient here, because the builder type itself needs to be an open-generic-over-`!!0` shape at the IL level (`AsyncTaskMethodBuilder<!!0>`), which requires a real reflection `Type` placeholder from the eventual `MethodBuilder`'s own generic parameters — a chicken-and-egg problem the current builder-then-resolve pipeline isn't set up for.

## Gap 2: hoisting a generic-typed parameter/local into the state machine crashes at runtime (BadImageFormatException) even when the declared return type is concrete

```gsharp
package P
async func Answer() int32 {
    return 42
}
async func Outer[U](seed U) int32 {
    var r = Answer()
    return await r
}
var t = Outer("hi")
t.Wait()
Console.WriteLine(t.Result)
```

[...]

PR #3942, opened 2026-09-05
gsc: the static-call sibling of the instance path — owner TypeSpec AND MethodSpec, not one or the other (#3939)

### LINKED ISSUE #3939: Emitter sweep: inconsistent siblings in generic contexts (#3705 class) — two live defects found, one filed
A systematic sweep of the emitter for the #3705 defect class — *a probe or path omits what a neighbouring probe a few lines away already does correctly* — restricted to paths that must reason about **generic context**.

The motivation: #3934 and #3936 found seven instances of that class in emit in two PRs, every one in a generic context, every one a sibling of an already-correct path, and all seven found only because a migrated app happened to exercise them. This sweep looks for the rest **before** the migration trips over them.

## Method

For each generic-context emit path, find its siblings — the async form next to the iterator form, the capturing closure next to the capture-free one, the **instance form next to the static one**, the imported-generic-type slot next to the imported-generic-method slot — and ask whether they agree. Where they disagree, one of them is wrong.

Grouped by **emitted IL**, not by symptom: three of the shapes below print three different runtime diagnostics (`InvalidOperationException`, `TypeLoadException`, ILVerify `[found ref 'string'][expected ref '!!0']`) and are **one** cause; grouping by the printed shape would have produced four causes and missed that they collapse.

Every row below was measured: compiled with `gsc`, run through `dotnet ilverify`, and executed.

## Findings

| # | site | siblings compared | agree? | which is right, and why | latent or live |
|---|---|---|---|---|---|
| **A** | `MethodBodyEmitter.Expressions.cs`, static-call token selection | STATIC call vs INSTANCE call (`MethodBodyEmitter.Calls.cs`) | **no** | the INSTANCE sibling. Owner-parenting and method-instantiation are independent axes; the instance path resolves a TypeSpec-parented MemberRef and *then* wraps it in a MethodSpec, sequentially. The static path chose between them with an `else if`, so `Box[int32].Make[U](…)` emitted a TypeSpec-parented MemberRef naming the **open** generic method and never got a MethodSpec | **live** |
| **B** | `StateMachineEmitter.EmitAsyncKickoffBody` | ASYNC kickoff's self-instantiation vs the reification in `ReflectionMetadataEmitter.RegisterStateMachineEnclosingGenerics`; and vs the ITERATOR kickoff | **no** | both of the others. #3932 root C taught the *reification* to read `StaticOwnerType` as well as `ReceiverType`; the kickoff body computes the same list independently and was never taught. A `shared async func Fetch[U]()` on a generic `Holder[T]` reifies at arity 2 and then self-instantiates with 1 argument | **live**, and hidden behind A |
| **C** | call-site MemberRef signature for a method returning a NESTED generic type | `Outer[T].Inner[U]` return vs a non-nested generic return | **no** | undetermined — filed separately, not fixed here | **live** |

A and B are fixed in the PR for this issue. C is filed as its own issue.

### A — a generic `shared` method on a generic type or interface loses its MethodSpec

```gs
class Holder[T] {
    shared {
        func Plain[U](v T, other U) string -> v.ToString() + "|" + other.ToString()
    }
}
Console.WriteLine(Holder[int32].Plain[string](41, "b"))
```

ILVerify: `[found ref 'string'][expected ref '!!0'] Unexpected type on the stack.`
Runtime: `InvalidOperationException: Could not execute the method because either the method itself or the containing type is not fully instantiated.`

Four surface shapes, one emitted-IL cause — `call <TypeSpec-parented MemberRef naming an open generic method>`, no MethodSpec:

- plain `shared func Make[U]` on a generic **class**;
- the same on a generic **interface** (`IBox[int32].Make[string](…)`, the #1433 branch, which additionally `break`s past the #3226 nullable-lift unwrap);
- `shared func` **iterator** (`sequence[…]`) generic on a generic type;
- `shared async func` generic on a generic type (which then also hits B).

Every neighbour is correct and stayed correct: the instance form `Holder[int32]().PlainInst[string](…)`, the non-generic `shared` form `Holder[int32].PlainNG(…)` (#1209's TypeSpec MemberRef), the generic `shared` form on a **non-generic** type (a bare MethodDef + MethodSpec), and `EmitMethodGroupTarget` in `MethodBodyEmitter.Closures.cs` — which applies the owner resolution and the MethodSpec **sequentially**, and is the model the fix follows.

### B — the async kickoff's self-instantiation

```gs
class Holder[T] {
    shared {
        async func FetchSharedG[U](v T, other U) ValueTask[string] { … }
    }
}
```

With A fixed, this surfaces as `TypeLoadException: The generic type '<FetchSharedG>d__0`2' was used with the wrong number of generic arguments`. The state-machine struct is reified over class-then-method parameters; `EmitAsyncKickoffBody` re-derived that list from `ReceiverType` alone, so a `shared` kickoff constructed a 2-parameter struct with 1 argument. The **iterator** kickoff has never had this bug because it constructs over the very `scopeTPs` its class was reified from — one list, one decision.

The fix is at the shared primitive (`StateMachineEmitter.GetAsyncKickoffGenericScope`), so the reification and the kickoff cannot drift again.

### C — a nested generic type as a generic method's return

```gs
class Outer[T] {
    class Inner[U] { var item T; var other U }
    func Make[U](v T, w U) Inner[U] { … }
}
let o = Outer[int32]()
let i = o.Make[string](4, "z")
```

ILVerify: `[found ref 'Outer`1+Inner`1<int32,string>'][expected ref 'Outer`1+Inner`1<!0,string>']`, then `BadImageFormatException` at load. The nested type's **enclosing** type argument is not substituted at the call-site token while its own is. Independent of A and B (it reproduces with no `shared`, no state machine and no lambda), so it is filed on its own.

## Pairs checked and CLEARED

Knowing where the emitter is consistent is worth as much as knowing where it is not.

| pair | verdict |
|---|---|
| async vs iterator state machine on a NON-generic type nested in a generic outer (`Outer[T].Inner`) | agree — both arity 1. `RegisterStateMachineEnclosingGenerics` does not walk `CollectEnclosingTypeParameters` while `GetIteratorTypeParametersInScope` does, but the nested definition's own `TypeParameters` already carries the re-declaration, so the two arrive at the same list. A textual asymmetry that is not a defect |
| capture-free vs capturing lambda inside a **generic method** of a generic encloser, reading the encloser's `private` field | agree — #3933's fix generalizes one type parameter deeper |
| iterator vs async **function literal** inside a generic method of a generic type (#3936's double-count shape, one parameter deeper) | agree |
| `EmitMethodGroupTarget` owner-resolution vs MethodSpec wrapping | agree — sequential, and the model finding A's fix follows |
| static field read/write through a constructed generic type (`Box[int32].count`), per-instantiation | agree |
| `MethodBodyPlanner.TryGetUserKickoffReceiverHandle` (async + iterator + async-iterator) vs `TryGetUserKickoffReceiverSymbol` (async only) | agree in effect — its only call site iterates `smStructsOrdered`, which is async SM **structs** only; iterator SMs are classes and take the other loop. The doc comment "the symbol counterpart of" overstates the relationship, but no shape reaches the gap |
| `ReceiverType ?? StaticOwnerType` across `MethodBodyPlanner`, `GetIteratorTypeParametersInScope`, `GetIteratorSourceTypeParameterOrigins`, `RegisterStateMachineEnclosingGenerics` | agree — after B is fixed, every emit-side consultation of the owning type reads both |

## Premise notes

Nothing in the framing of this sweep was contradicted, but two things are worth stating precisely:

- Finding A is **not** a state-machine or closure bug at all. It reproduces with a plain non-async, non-iterator `shared func`. The async and iterator shapes are downstream of it, which is exactly why grouping by symptom would have mis-attributed it to `StateMachineEmitter`.
- The `else if` in A predates both #1209 and #1433 as a *pair*: each branch was correct for the case it was written for, and the gap only exists where both conditions hold at once. Neither issue was wrong; the composition was never considered.

Refs #3705, #3707, #3716, #3741, #3932, #3933, #3934, #3936, #1209, #1433, #1465, #1467, #1512, #2030, #3226, #3501, ADR-0087.

DavidObando:
A **proactive sweep** of gsc's emitter for #3705's defect class — *a path omits what a neighbouring path a few lines away already does correctly* — restricted to paths that must reason about **generic context**. #3934 and #3936 found seven instances of that class in emit in two PRs, all in generic context, all sibling-shaped, and every one of them found only because a migrated app happened to exercise it. This looks for the rest first.

Two live defects found and fixed; one more found and filed as **#3940**; seven sibling pairs checked and **cleared**.

Sweep and findings table: **#3939**.

## A — a generic `shared` method on a generic type loses its MethodSpec

```gs
class Holder[T] {
    shared {
        func Plain[U](v T, other U) string -> v.ToString() + "|" + other.ToString()
    }
}
Console.WriteLine(Holder[int32].Plain[string](41, "b"))
```

ILVerify: `[found ref 'string'][expected ref '!!0']`. Runtime: `InvalidOperationException: Could not execute the method because either the method itself or the containing type is not fully instantiated.`

**This is live and reachable from ordinary hand-written G#, not a migration artifact.** The snippet above is the whole reproducer: a `shared` generic method on a generic type, an unremarkable shape, written directly. gsc accepts it without a diagnostic and emits an assembly that does not verify and throws on the first call. The sweep was motivated by migration work, but this finding is a plain correctness bug for anyone writing G# today, and it should be reviewed as one.

The static-call path in `MethodBodyEmitter.Expressions.cs` chose between two **independent axes** with an `else if`:

- the **owner** token — which type the method belongs to. A `shared` method on a constructed generic type (#1209) or interface (#1433) needs a MemberRef parented at the construction's TypeSpec.
- the **MethodSpec** — how the *method* is instantiated (ADR-0087 §3 R3+R4).

A `shared func Make[U]()` on a generic `Box[T]` needs both, and got only the first: a TypeSpec-parented MemberRef naming the **open** generic method.

Neither #1209 nor #1433 was wrong. Each branch is correct for the case it was written for; the gap exists only where both conditions hold at once, and the composition was never considered.

**The already-correct siblings.** `EmitUserInstanceCall` (`MethodBodyEmitter.Calls.cs`) resolves a TypeSpec-parented MemberRef for a generic receiver and *then* wraps it in a MethodSpec whenever `call.Method.IsGeneric` — sequentially. So does `EmitMethodGroupTarget` (`MethodBodyEmitter.Closures.cs`), which is the model this fix follows. The non-generic `shared` form, the generic `shared` form on a *non*-generic type, and the instance form all stayed correct throughout.

Four surface shapes, **one** emitted-IL cause (`call <TypeSpec-parented MemberRef naming an open generic method>`): the plain class form; the interface form, whose #1433 branch resolved its token and then `break`ed out of the case entirely so it could never reach the MethodSpec *or* #3226's nullable-lift unwrap; the `shared` **iterator** form; and the `shared async` form. Grouping by symptom would have split these across three files and attributed two of them to `StateMachineEmitter` — where nothing is wrong.

## B — the async kickoff's self-instantiation, hidden behind A

```gs
class Holder[T] {
    shared {
        async func FetchSharedG[U](v T, other U) ValueTask[string] { … }
    }
}
```

With A fixed this surfaces on its own: `TypeLoadException: The generic type '<FetchSharedG>d__0`2' was used with the wrong number of generic arguments`.

The state-machine struct is reified over class-then-method parameters by `ReflectionMetadataEmitter.RegisterStateMachineEnclosingGenerics`. `StateMachineEmitter.EmitAsyncKickoffBody` re-derived the same list to self-instantiate it — and its copy read `ReceiverType` alone, so it never picked up the `StaticOwnerType` half **#3932 root C added to the reification one file over**. A `shared` kickoff therefore constructed a 2-parameter struct with 1 argument.

The **iterator** kickoff has never had this bug, and the reason is structural rather than lucky: it constructs over the very `scopeTPs` its class was reified from. One list, one decision. So the fix is at the shared primitive — `StateMachineEmitter.GetAsyncKickoffGenericScope`, which both the reification and the kickoff now read — in the spirit of the #3705 binder funnels (`ClrMemberVisibility`, `ClrLoadContext`) that made those fixes stick.

## Evidence

`test/Compiler.Tests/Emit/Issue3939GenericStaticSiblingEmitTests.cs` — four tests, each of which **COMPILES, ILVERIFIES, RUNS and asserts on behaviour**. All three are load-bearing on #3501, and this week supplied fresh proof of each: #3936's intermediate state ILVerified clean and threw `TypeLoadException` at load, #3900's first cut passed an executing test and was caught only by ILVerify.

Two also read the emitted metadata back, because the facts at issue are ones behaviour cannot name:

- A: the **token kind** in `<Main>$` — a `MethodSpecification` must be present (the fix) *and* a `MemberReference` must still be present (the non-generic-method control, which must not gain a MethodSpec).
- B: the **arity** of each state machine, with all four controls pinning *different* expected arities (2 / 1 / 2 / 1), so a mutant that reifies unconditionally at 2 is caught as surely as one that reverts to `ReceiverType`.

The iterator test asserts the **full ordered** output rather than `Contains`, so an iterator that runs but yields in the wrong order or drops an element is caught.

Discrimination (ADR-0154): every test carries controls that passed **before** the fix — the same construct in its instance form, its non-generic-*method* form, and its non-generic-*encloser* form.

**Anti-vacuity**, and it separates the two fixes:

| src/Core state | Issue3939 result |
|---|---|
| `origin/main` | **4 failed / 0 passed** |
| A only (`MethodBodyEmitter.Expressions.cs`) | 1 failed / 3 passed — the async test still fails, so B is independently load-bearing |
| A + B | 4 passed / 0 failed |

The middle row is the point: **B is independently load-bearing.** With A's fix in and B's reverted, three tests go green and the async one still fails — so B is not a speculative second fix riding along with the first, and "was the second change necessary?" has a measured answer rather than an argued one. B was simply invisible until A stopped failing first.

## Filed, not fixed

**#3940** — a nested generic type returned from a generic method keeps its **enclosing** type argument open at the call site: `Outer`1+Inner`1<!0,string>` where `<int32,string>` belongs, then `BadImageFormatException` at load. It reproduces with no `shared`, no state machine and no lambda, so it is not this family and does not belong in this change.

## Pairs checked and CLEARED

Knowing where the emitter is consistent is worth as much as knowing where it is not. Each of these was compiled, ilverified and run:

- async vs iterator state machine on a non-generic type nested in a generic outer (`Outer[T].Inner`) — both arity 1. `RegisterStateMachineEnclosingGenerics` does not walk `CollectEnclosingTypeParameters` while `GetIteratorTypeParametersInScope` does, but the nested definition's own `TypeParameters` already carries the re-declaration, so they arrive at the same list. **A textual asymmetry that is not a defect.**
- capture-free vs capturing lambda inside a **generic method** of a generic encloser, reading the encloser's `private` field — #3933 generalizes one type parameter deeper.
- iterator vs async **function literal** inside a generic method of a generic type — #3936's double-count shape, one parameter deeper.
- `EmitMethodGroupTarget`'s owner resolution vs its MethodSpec wrapping — sequential, and the model A's fix follows.
- static field read/write through a constructed generic type (`Box[int32].count`), per-instantiation.
- `MethodBodyPlanner.TryGetUserKickoffReceiverHandle` (async + iterator + async-iterator) vs `TryGetUserKickoffReceiverSymbol` (async only) — agree **in effect**: the only call site iterates `smStructsOrdered`, which is async SM *structs* only; iterator SMs are classes and take the other loop. The doc comment "the symbol counterpart of" overstates the relationship, but no shape reaches the gap.
- every emit-side `ReceiverType ?? StaticOwnerType` consultation — after B, they all read both.

## Premise notes

- Finding A is **not** a state-machine or closure bug. It reproduces with a plain non-async, non-iterator `shared func`; the async and iterator shapes are downstream of it. The sweep's framing anticipated `StateMachineEmitter` and `ClosureEmitter` as the likely sites, and the densest cause turned out to be neither.
- `ClosureEmitter` came out **clean**: every sibling pair probed there agreed. #3933/#3936 appear to have closed that area rather than partially closed it.

Refs #3939, #3940, #3705, #3707, #3716, #3741, #3932, #3933, #3934, #3936, #1209, #1433, #1465, #1467, #1512, #2030, #3226, #3501, ADR-0087, ADR-0154.

Closes #3939.

### DIFF
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Expressions.cs
@@ -143,24 +143,35 @@ internal sealed partial class MethodBodyEmitter
                     }
                 }
 
-                // Issue #1433: a `shared` (static) method called on a
-                // constructed generic INTERFACE (`IBox[int32].Make()`) must be
+                // The OWNER token: which entity the call names. Three shapes —
+                // a `shared` method on a constructed generic INTERFACE
+                // (`IBox[int32].Make()`, issue #1433) or on a constructed
+                // generic user TYPE (`Box[int32].Make()`, issue #1209) must be
                 // referenced through a MemberRef parented at the construction's
-                // TypeSpec. The substituted method is not in the handle cache,
-                // so resolve it before the generic-struct/cache lookups below.
+                // TypeSpec; everything else is the bare MethodDef.
+                EntityHandle callTokenToEmit;
                 if (call.StaticGenericInterfaceOwnerType != null
                     && ReflectionMetadataEmitter.IsUserGenericInterfaceReference(call.StaticGenericInterfaceOwnerType))
                 {
-                    this.il.OpCode(ILOpCode.Call);
-                    this.il.Token(this.outer.userTokens.ResolveUserInterfaceStaticMethodToken(call.StaticGenericInterfaceOwnerType, call.Function));
-                    break;
+                    // The substituted method is not in the handle cache, so this
+                    // resolves without the generic-struct/cache lookups below.
+                    callTokenToEmit = this.outer.userTokens.ResolveUserInterfaceStaticMethodToken(
+                        call.StaticGenericInterfaceOwnerType,
+                        call.Function);
                 }
-
-                if (!this.outer.cache.FunctionHandles.TryGetValue(call.Function, out var fnHandle)
-                    && !this.outer.cache.MethodHandles.TryGetValue(call.Function, out fnHandle))
+                else
                 {
-                    throw new InvalidOperationException(
-                        $"Call to function '{call.Function.Name}' has no emitted MethodDef.");
+                    if (!this.outer.cache.FunctionHandles.TryGetValue(call.Function, out var fnHandle)
+                        && !this.outer.cache.MethodHandles.TryGetValue(call.Function, out fnHandle))
+                    {
+                        throw new InvalidOperationException(
+                            $"Call to function '{call.Function.Name}' has no emitted MethodDef.");
+                    }
+
+                    callTokenToEmit = call.StaticGenericOwnerType != null
+                        && ReflectionMetadataEmitter.IsUserGenericTypeReference(call.StaticGenericOwnerType)
+                        ? this.outer.userTokens.ResolveUserStaticMethodToken(call.StaticGenericOwnerType, call.Function)
+                        : fnHandle;
                 }
 
                 // ADR-0087 §3 R3+R4: when the target is a generic function,
@@ -168,25 +179,32 @@ internal sealed partial class MethodBodyEmitter
                 // the substituted type arguments — bare MethodDef tokens
                 // would carry MVAR slots and fail ilverify against the
                 // concrete argument types.
-                EntityHandle callTokenToEmit = fnHandle;
-                if (call.StaticGenericOwnerType != null
-                    && ReflectionMetadataEmitter.IsUserGenericTypeReference(call.StaticGenericOwnerType))
-                {
-                    // Issue #1209: a `shared` (static) method called on a
-                    // constructed generic user type (`Box[int32].Make()`) must
-                    // be referenced through a MemberRef parented at the
-                    // construction's TypeSpec; a bare MethodDef token is invalid
-                    // for a method of a generic type.
-                    callTokenToEmit = this.outer.userTokens.ResolveUserStaticMethodToken(call.StaticGenericOwnerType, call.Function);
-                }
-                else if (call.Function.IsGeneric && !call.Function.TypeParameters.IsDefaultOrEmpty)
+                //
+                // Issue #3939: this MUST be applied on TOP of whatever owner
+                // token was selected above, not INSTEAD of it. The two are
+                // independent axes — the owner token says which type the method
+                // belongs to, the MethodSpec says how the METHOD is
+                // instantiated — and a `shared func Make[U]()` on a generic
+                // `Box[T]` / `IBox[T]` needs both. Selecting one or the other
+                // (which is what an `else if` did here through #1209 and #1433)
+                // emitted `call Box`1<int32>::Make(!!0)` naming the OPEN generic
+                // method: `[found ref 'string'][expected ref '!!0']` from
+                // ILVerify and `InvalidOperationException: Could not execute the
+                // method because either the method itself or the containing type
+                // is not fully instantiated` at the first call. The INSTANCE
+                // sibling (`EmitUserInstanceCall` in MethodBodyEmitter.Calls.cs)
+                // has always done both, sequentially: it resolves a
+                // TypeSpec-parented MemberRef for a generic receiver and THEN
+                // wraps it in a MethodSpec whenever `call.Method.IsGeneric`.
+                // This is the static path catching up.
+                if (call.Function.IsGeneric && !call.Function.TypeParameters.IsDefaultOrEmpty)
                 {
                     // Issue #3226: a lifted call instantiates the MethodSpec at
                     // Nullable<X> so the erased `!!T` slots really carry the
                     // nullable representation the T? contract promises.
                     callTokenToEmit = nullableLift != null
-                        ? this.outer.userTokens.BuildMethodSpecForLiftedGenericCall(fnHandle, nullableLift.TypeArguments)
-                        : this.outer.userTokens.BuildMethodSpecForGenericCall(fnHandle, call);
+                        ? this.outer.userTokens.BuildMethodSpecForLiftedGenericCall(callTokenToEmit, nullableLift.TypeArguments)
+                        : this.outer.userTokens.BuildMethodSpecForGenericCall(callTokenToEmit, call);
                 }
 
                 this.il.OpCode(ILOpCode.Call);
--- a/src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs
@@ -461,13 +461,12 @@ internal sealed class ReflectionMetadataEmitter
         // consulted both — this is the async path catching up. Instance methods
         // were unaffected, which is why the failure needed a `shared async
         // func` on a generic type to surface at all.
-        var receiverDef = (kickoff?.ReceiverType as StructSymbol ?? kickoff?.StaticOwnerType as StructSymbol) is { } receiver
-            ? (receiver.Definition ?? receiver)
-            : null;
-        var classTPs = receiverDef?.TypeParameters ?? ImmutableArray<TypeParameterSymbol>.Empty;
-        var methodTPs = kickoff == null || kickoff.TypeParameters.IsDefaultOrEmpty
-            ? ImmutableArray<TypeParameterSymbol>.Empty
-            : kickoff.TypeParameters;
+        //
+        // Issue #3939: the same decision is needed by the KICKOFF BODY, which
+        // self-instantiates the struct this method reifies, so both now read it
+        // from one primitive on StateMachineEmitter instead of each deriving it
+        // — the kickoff's copy had never picked up the `StaticOwnerType` half.
+        var (receiverDef, classTPs, methodTPs) = StateMachineEmitter.GetAsyncKickoffGenericScope(kickoff);
 
         if (classTPs.IsDefaultOrEmpty && methodTPs.IsDefaultOrEmpty)
         {
--- a/src/Core/CodeAnalysis/Emit/StateMachineEmitter.cs
+++ b/src/Core/CodeAnalysis/Emit/StateMachineEmitter.cs
@@ -476,6 +476,59 @@ internal sealed class StateMachineEmitter
         public AsyncMethodSteppingInfo? AsyncStepping { get; }
     }
 
+    /// <summary>
+    /// Issue #3939: the single place that decides which type parameters are in
+    /// scope at an ASYNC kickoff method — the owning type's, then the kickoff's
+    /// own, in Roslyn order. Two emit paths depend on this list agreeing
+    /// exactly: <c>ReflectionMetadataEmitter.RegisterStateMachineEnclosingGenerics</c>
+    /// reifies the state-machine struct at this ARITY, and
+    /// <see cref="EmitAsyncKickoffBody"/> self-instantiates the struct with
+    /// this INSTANTIATION. They used to compute it independently, and the
+    /// second read only <c>ReceiverType</c> — so a <c>shared async func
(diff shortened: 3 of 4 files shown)


DavidObando:
## Verification runs — all four green

| run | against | result |
|---|---|---|
| full `Compiler.Tests` | `691df2d98` (every `src/Core` change) | **4745 passed / 0 failed**, 42m53s |
| filtered `Issue3939` | `219139d4c` (adds the self-reference facet) | **4 passed / 0 failed** |
| `Core.Tests` | `219139d4c` | **8904 passed / 0 failed**, 11m30s |
| `build/run-cs2gs-selfmig-pr-guard.sh` | `219139d4c` | **5/5 apps green; run PASSED** (migrate exit 0) |

### Why each was worth running rather than assumed

The **full `Compiler.Tests`** run is the one that answers the #3936 exposure directly: that change turned an *unrelated* existing test red (`Issue2957FunctionLiteralIteratorTests`) with a targeted run that was clean, and it only surfaced under the full suite. Nothing went red here, so this change did not repeat that pattern.

**`Core.Tests`** is the run that caught #3936's state-machine arity regression, and this change is in `src/Core` touching the same `StateMachineEmitter` surface. Clean at 8904.

The **selfmig PR guard** is the one that exists for exactly this kind of change — the hazard has fired four times (#3831, #3896, #3905, and #3915's rewrite), and the two most expensive were `src/Core` changes. All four stages green on all five hot-core apps, `Gsharp.Runtime.Channels` included (added to `guard_apps` by #3936, and the app whose ilverify stage produced this whole defect family):

```
gsc: 0.4.474+219139d4c6

app                                                             translate  compile  ilverify  test-parity
src/Analyzers/InternalAnalyzers/InternalAnalyzers.csproj        PASS       PASS     PASS      PASS
src/Core/Core.csproj                                            PASS       PASS     PASS      PASS
src/Sdk/Gsharp.Runtime.Channels/Gsharp.Runtime.Channels.csproj  PASS       PASS     PASS      PASS
tools/cs2gs/Cs2Gs.CodeModel/Cs2Gs.CodeModel.csproj              PASS       PASS     PASS      PASS
tools/cs2gs/Cs2Gs.Translator/Cs2Gs.Translator.csproj            PASS       PASS     PASS      PASS

5/5 apps green; run PASSED.
```

`src/Core` translating and compiling green matters twice over here: the translator is inside its own corpus, so an emitter change that CI passes can still break the gate by existing — and this change is in the emitter that compiles the migrated result.

Per the script's own banner, a green run here means *the hot core still migrates*, not *this PR migrates cleanly*; only the nightly gate is the gate.

### Commit split

`691df2d98` carries all three `src/Core` edits. `219139d4c` is **test-only** — the self-reference facet (a `shared func Plain[U]()` called from inside `Holder[T]`, where the owner instantiation is the open `!0` rather than a concrete `int32`). That facet is covered by the filtered run above; the full suite ran against the commit holding the entire compiler change.

Neither `tools/cs2gs/selfmig-baseline.json` nor `selfmig-test-allowlist.json` was touched. The guard ran with `SELFMIG_PR_GUARD_ROOT` pointed at a scratch directory.

PR #3947, opened 2026-09-05
gsc: a facade call the author wrote is not a lowered channel operator (#3945)

### LINKED ISSUE #3945: gsc: a written ChannelOps facade call colors its caller, adding a hidden Context parameter to [Fact] methods
## Symptom

Migrated `test/Extensions.Tests` fails test-parity with one fingerprint (`sha256:146ec6a1d0e9…`), four tests:

```
Failed Merge_OneInputClosing_DoesNotCloseTheOutput
Failed Chunks_DoesNotWaitToFillABatch
Failed Merge_DeliversEveryValueFromEveryInput_ThenCloses
Failed Chunks_HandsOverWholeBatches_AndEndsWithTheChannel
  System.InvalidOperationException : [Fact] methods are not allowed to have parameters. Did you mean to use [Theory]?

Failed!  - Failed: 4, Passed: 176, Skipped: 0, Total: 180
```

Observed in gate run 33943018295. `test/Extensions.Tests` is a banked `greenApps` entry.

## Root cause

ADR-0174 D4 suspension inference (`SuspensionPointCollector.VisitImportedCallExpression`) marks a body as having a direct suspension point whenever it reaches `ChannelOps.Receive` / `Receive2` / `Send`, via `ChannelRuntimeBinder.IsFacadeCall`, which matches on method name and declaring type only.

That does not distinguish the facade calls the **binder lowers** from a channel operator (`<-ch`, `ch <- v`, a channel `range`, a `select` arm) from the ones the **author writes out by name**. The four failing tests are exactly the four methods in `ConcurrencyTests` that write `ChannelOps.Receive2(reader, CancellationToken.None)`; every other test in the class passes.

Coloring is an ABI change — D7 prepends a hidden leading `Context` parameter and the return is retyped `ValueTask` — so a parameterless method acquires an argument, and xUnit rejects it at discovery.

For a written call the coloring buys nothing: `RetargetFacadeCancellation` only retargets a facade call whose trailing cancellation is a `BoundDefaultExpression`, so a call that already carries the author's token is left untouched. The function pays the ABI cost with no corresponding rewrite.

## Minimal repro

```
class Probe {
    func WrittenFacadeCall() {
        let ch = Chan[int32](1)
        ch.TrySend(7)
        let (value, ok) = ChannelOps.Receive2(ch, CancellationToken.None)
        Console.WriteLine("w=" + value.ToString() + ":" + ok.ToString())
    }
}
```

Emitted signature on `9ec9a4c89`:

```
Probe::WrittenFacadeCall  sigParams=1 [Context] -> ValueTask
```

Expected `sigParams=0 -> Void`. The operator form (`let (value, ok) = <-ch`) must keep `sigParams=1 [Context]`.

## Scope note

This is why `test/Extensions.Tests` regressed. It did **not** regress in the window most recently under suspicion: the four tests were added on 2026-09-03 by #3882 (`5976026ca`, `3f4e939cc`), in the same PR that introduced suspension inference. The last green gate run was 33819616178 (`4ed74c3f4`); the first failing one was 33825780017 (`36a147f0f` = #3882).

DavidObando:
Fixes #3945. Takes migrated `test/Extensions.Tests` back to green, one of the two banked `greenApps` entries the gate reported as no longer green in run 33943018295.

## The premise correction first

The two apps the gate named were investigated together on the assumption that both regressed between `efd1b1a91` and `f1399b202`. Neither did.

- That window contains **six** commits, not the four usually cited: `#3930` (`f42549eff`) and `#3934` (`f7fde6d2f`) are also in it.
- More importantly, the **previous** gate run (33929083318, `efd1b1a91`) already flagged *both* apps as no longer green, at 39/53. It was not a both-green baseline.
- The last actually-green run was **33819616178** (`4ed74c3f4`, 2026-09-03 23:54); the first failing one was **33825780017** (`36a147f0f`). That range is a single merge: **#3882**, ADR-0174 wave 2.

Both apps trace to #3882, for two unrelated reasons.

## `bench/concurrency/clr/ClrBaseline.csproj` — not a defect, a stale ratchet entry

It is **absent from the corpus entirely** (0 of 53 apps are under `bench/`). #3882's commit `ec15e4958` added, deliberately and with a good rationale:

```
  # ADR-0174 D11: the concurrency benchmark's C# and Go sides are measurement
  # apparatus, not migration targets. Translating the CLR baseline would
  # measure the translator rather than the runtime …
  --exclude bench/concurrency/clr
  --exclude bench/concurrency/aot
```

It had been banked into `greenApps` the day before (gate run 33743456719). So the app was banked on 09-03 and excluded on 09-04 by the same wave that added it. **There is no code fix; it can never be green again by design.** The `greenApps` entry has to go.

That edit belongs in `tools/cs2gs/selfmig-baseline.json`, which this PR deliberately does not touch — #3944 owns it. Note that #3944 currently keeps the entry, on the stated grounds that the two apps "are being fixed, not absorbed". Half of that is right; ClrBaseline is not fixable.

**Corrected arithmetic.** With this PR's fix, and #3944's changes, and the phantom dropped:

| | |
|---|---|
| corpus | 53 |
| green now | 43 |
| green with this fix | **44** |
| `greenApps` after #3944 + dropping the phantom | 44 entries, all green |
| `greenFloor` | 44 — exactly met |

So the outcome is 43 → **44**, not 43 → 45.

## `test/Extensions.Tests` — the real defect

Its single test-parity fingerprint is four tests failing at xUnit *discovery*:

```
System.InvalidOperationException : [Fact] methods are not allowed to have parameters.
Failed!  - Failed: 4, Passed: 176, Skipped: 0, Total: 180
```

(A count with its `Failed! … Total:` line, per #3931.)

The four are exactly the four methods in `ConcurrencyTests` that call `ChannelOps.Receive2(reader, CancellationToken.None)`; the other tests in the class pass. Those tests were added by #3882 itself (`5976026ca`, `3f4e939cc`) — the same PR that introduced suspension inference — so the app broke by acquiring tests that exercise a defect the same PR shipped.

**Root cause.** ADR-0174 D4 inference (`SuspensionPointCollector.VisitImportedCallExpression`) treats any reach of `ChannelOps.Receive/Receive2/Send` as a direct suspension point, via `IsFacadeCall`, which matches on name and declaring type only. It does not distinguish the facade calls the **binder lowers** from a channel operator (`<-ch`, `ch <- v`, a channel `range`, a `select` arm) from those the **author writes out**.

Coloring is an ABI change — D7 prepends a hidden leading `Context` parameter and retypes the return as `ValueTask`. That is the whole point for an operator. For a written call it buys nothing: `RetargetFacadeCancellation` only retargets a facade call whose trailing cancellation is a `BoundDefaultExpression`, so a call already carrying the author's token is left untouched. The function pays the ABI cost with no corresponding rewrite — and anything reflecting over the signature breaks.

**The fix** uses the discriminator that path already relies on. A lowered operation is bound with a cancellation the binder supplied (the uncancellable default token, or an enclosing `scope`'s `Context`); a written call passes a `CancellationToken` expression of its own. Only the latter stops being a suspension point. Layer: `src/Core`, binding — not emit, and not cs2gs, whose translation of these calls is faithful.

## Before / after

Minimal repro, emitted signature blob (what `MethodInfo.GetParameters()` reads):

| method | on `9ec9a4c89` | with this PR |
|---|---|---|
| `WrittenFacadeCall` (`ChannelOps.Receive2(ch, CancellationToken.None)`) | `1 [Context] -> ValueTask` | **`0 [] -> Void`** |
| `OperatorReceive` (`let (v, ok) = <-ch`) | `1 [Context] -> ValueTask` | `1 [Context] -> ValueTask` (unchanged) |
| `NoChannelOp` | `0 [] -> Void` | `0 [] -> Void` |

## Test evidence

`test/Compiler.Tests/Issue3945WrittenFacadeCallColoringTests.cs` — compiles, **IL-verifies and runs**, and additionally reads the emitted signature back, because "the method has no parameters" is an encoding fact no behavioural assertion can name and is precisely what xUnit reads.

**Failing-on-main proof, honestly labelled.** With the discriminator reverted to main's behaviour:

```
Failed WrittenFacadeCall_KeepsItsSignature_WhileTheOperatorFormIsStillColored
Failed!  - Failed: 1, Passed: 1, Total: 2
```

The other test, `WrittenFacadeCall_StillReceives_AndTheProgramRuns`, **passes on main** — the program always ran correctly; only the signature was wrong. That is the anti-vacuity guard rail, and it is why a behaviour-only test would have passed a wrong fix here.

**Discrimination (ADR-0154).** `OperatorReceive` must still be colored and sits in the same program, so a mutant that stops coloring facade calls altogether fails as surely as one that colors written ones.

**No regressions**, targeted suites on this branch:

| suite | result |
|---|---|
| `Core.Tests` — Suspension / Adr0174 / Channel / Concurrency | 328 passed, 0 failed |
| `Compiler.Tests` — Adr0174 / Channel / Issue3932 / Issue3907 | 47 passed, 0 failed |
| `Extensions.Tests` (the C# oracle) | 180 passed, 0 failed |
| `Runtime.Channels.Tests` | 188 passed, 0 failed |

The cs2gs PR guard triggers automatically on `src/Core/**`, so it runs here.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ChannelRuntimeBinder.cs
+++ b/src/Core/CodeAnalysis/Binding/ChannelRuntimeBinder.cs
@@ -671,6 +671,36 @@ internal sealed class ChannelRuntimeBinder
             && call.Function.ImportedClass.ClassType.FullName == ChannelOpsTypeName;
 
     /// <summary>
+    /// ADR-0174 D4/D7: distinguishes a facade call the BINDER lowered from a
+    /// channel operator (<c>&lt;-ch</c>, <c>ch &lt;- v</c>, a channel
+    /// <c>range</c>, a <c>select</c> arm) from one the AUTHOR wrote by name.
+    /// </summary>
+    /// <remarks>
+    /// <para>Every lowered operation is bound with a cancellation the binder
+    /// itself supplied — the uncancellable default token when no lexical scope
+    /// encloses it, or that scope's <c>Context</c> when one does — precisely so
+    /// the suspension pass can later park it on the ambient context via
+    /// <see cref="RetargetFacadeCancellation"/>. A call written out as
+    /// <c>ChannelOps.Receive2(ch, token)</c> has already chosen its
+    /// cancellation, so there is nothing for that pass to retarget: it is an
+    /// ordinary blocking call to a public library method.</para>
+    /// <para>That distinction is load-bearing for the COLORING. Marking the
+    /// caller suspending prepends a hidden <c>Context</c> parameter and retypes
+    /// the return as <c>ValueTask</c> — an ABI change — which is the point for a
+    /// channel operator but is pure cost for a written facade call, whose body
+    /// the pass leaves untouched. It also breaks callers that reflect over the
+    /// signature: an xUnit <c>[Fact]</c> that drains a channel through the
+    /// facade is rejected at discovery as "not allowed to have parameters"
+    /// (#3501, migrated <c>test/Extensions.Tests</c>).</para>
+    /// </remarks>
+    /// <param name="call">A facade call, already matched by <see cref="IsFacadeCall"/>.</param>
+    /// <returns><see langword="true"/> when the trailing cancellation argument was written by the author rather than supplied by the binder.</returns>
+    public static bool HasAuthorWrittenCancellation(BoundImportedCallExpression call)
+        => call.Arguments.Length != 0
+            && call.Arguments[^1] is not BoundDefaultExpression
+            && call.Arguments[^1].Type?.ClrType?.FullName == CancellationTokenTypeName;
+
+    /// <summary>
     /// ADR-0174 D7: rebinds a facade call that was bound with the
     /// uncancellable default token so it parks on <paramref name="context"/>
     /// instead. Used by the suspension pass for operations that no lexical
--- a/src/Core/CodeAnalysis/Binding/Suspension/SuspensionPointCollector.cs
+++ b/src/Core/CodeAnalysis/Binding/Suspension/SuspensionPointCollector.cs
@@ -87,11 +87,18 @@ internal sealed class SuspensionPointCollector : BoundTreeWalker
     /// <inheritdoc/>
     protected override void VisitImportedCallExpression(BoundImportedCallExpression node)
     {
-        if (goDepth == 0 && lockDepth == 0
-            && (ChannelRuntimeBinder.IsFacadeCall(node, "Receive")
+        var isFacadeOperation =
+            (ChannelRuntimeBinder.IsFacadeCall(node, "Receive")
                 || ChannelRuntimeBinder.IsFacadeCall(node, "Receive2")
-                || ChannelRuntimeBinder.IsFacadeCall(node, "Send")
-                || LockRegions.IsBlockingBridge(node)))
+                || ChannelRuntimeBinder.IsFacadeCall(node, "Send"))
+
+            // A facade call the AUTHOR wrote with its own token is an ordinary
+            // blocking library call, not a lowered channel operator: the
+            // suspension pass has nothing to retarget on it, so coloring the
+            // caller would change its ABI for no gain (ADR-0174 D4/D7).
+            && !ChannelRuntimeBinder.HasAuthorWrittenCancellation(node);
+
+        if (goDepth == 0 && lockDepth == 0 && (isFacadeOperation || LockRegions.IsBlockingBridge(node)))
         {
             facts.HasDirectPoint = true;
         }
--- a/test/Compiler.Tests/Issue3945WrittenFacadeCallColoringTests.cs
+++ b/test/Compiler.Tests/Issue3945WrittenFacadeCallColoringTests.cs
@@ -1,0 +1,282 @@
+// <copyright file="Issue3945WrittenFacadeCallColoringTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Metadata;
+using System.Reflection.PortableExecutable;
+using System.Text;
+using Xunit;
+
+namespace GSharp.Compiler.Tests;
+
+/// <summary>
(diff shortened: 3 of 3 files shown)


DavidObando:
## Migrated-run proof

Targeted migration of `test/Extensions.Tests` plus its full reference closure, on this branch (gsc `0.4.476+4799d980a5`). Ordering per the guard: Cs2Gs.Cli Release → Debug prerequisites → **Release SDK nupkg repack immediately before migrate**, nothing built in between.

```
app                                                              translate  compile  ilverify  test-parity
src/Analyzers/GSharp.CodeAnalysis.Analyzers.Testing/…            PASS       PASS     PASS      PASS
src/Analyzers/InternalAnalyzers/InternalAnalyzers.csproj         PASS       PASS     PASS      PASS
src/Compiler/Compiler.csproj                                     PASS       PASS     PASS      PASS
src/Core/Core.csproj                                             PASS       PASS     PASS      PASS
src/Sdk/Gsharp.Runtime.Channels/Gsharp.Runtime.Channels.csproj   PASS       PASS     PASS      PASS
test/Extensions.Tests/Extensions.Tests.csproj                    PASS       PASS     PASS      PASS

6/6 apps green
```

### Before / after, with the run summary line in both

| | |
|---|---|
| before (gate run 33943018295, `f1399b202`) | `Failed!  - Failed: 4, Passed: 176, Skipped: 0, Total: 180` |
| after (this branch) | `Passed!  - Failed: 0, Passed: 180, Skipped: 0, Total: 180` |

Both carry the `Failed!`/`Passed! … Total:` line, so neither number is a truncation artifact (#3931). **The total is 180 on both sides** — the four tests were fixed, not lost, which is the #3872/#3869 evidence rule the pipeline also enforces independently before it will report the stage green.

`Extensions.Tests` had exactly one fingerprint (`sha256:146ec6a1d0e9…`) and it is gone. That is the app back to green.

### One honest caveat on the exit code

The run reports `migrate exit 1` despite 6/6 green. The cause is a mirror-fidelity complaint unrelated to this change:

```
cs2gs: Repository migration produced unexpected file 'src/Sdk/Gsharp.Extensions/Gsharp.Extensions.xml'
```

That is an artifact of the hand-rolled narrow `--exclude` set used for this targeted proof (a stray doc-XML from a prior full-solution build in the same tree), not a stage failure and not a parity failure. Every stage of every app reads PASS. The real corpus-wide verdict is the nightly gate; this run is evidence for one app.

Issue #3959, opened 2026-09-05
gsc: the CLR fallback type renderer spells a rank-1 array 'int32[]', not G#'s '[]int32'


DavidObando:
## The gap

When a diagnostic (or hover) renders a type through the **CLR fallback** renderer, a
rank-1 array is spelled in C# suffix form, `int32[]`. That is not valid G# source
spelling, and the very next line of the same function already renders rank ≥ 2 in G#'s
prefix form.

`src/Core/CodeAnalysis/Symbols/Display/SymbolDisplay.cs:985`:

```csharp
if (clrType.IsArray)
{
    return clrType.GetArrayRank() == 1
        ? $"{FormatClrTypeName(clrType.GetElementType(), qualifyNames)}[]"                                  // int32[]   ← C# spelling
        : $"[{new string(',', clrType.GetArrayRank() - 1)}]{FormatClrTypeName(clrType.GetElementType(), qualifyNames)}";  // [,]int32 ← G# spelling
}
```

The symbol path a few hundred lines up is already right — `FormatType`'s
`case SliceTypeSymbol slice: return $"[]{FormatType(slice.ElementType)}"`
(`SymbolDisplay.cs:717`). Only the fallback taken for a CLR-backed
`ImportedTypeSymbol` disagrees with it.

## How it shows up

One diagnostic can print both spellings for the same element type. From #3924, before
that fix (the parameter is a `chan[T]` type clause, the argument a CLR-backed
`Chan<int[]>`):

```
error GS0154: Parameter 'ch' requires a value of type 'out chan[[]int32]'
              but was given a value of type 'Gsharp.Concurrency.Chan[int32[]]'.
```

Same type, two spellings, in one sentence — which reads as a type mismatch that is not
one. #3924 removed *that* particular message by making the conversion bind, but the
renderer is unchanged and still reachable: any genuine mismatch over an array element
prints it, e.g. passing a `chan[[]int32](4)` where `out chan[[]int64]` is wanted.

## Expected

`[]int32`, matching the G# type-clause spelling, the `SliceTypeSymbol` case, and the
rank ≥ 2 branch beside it.

## Notes

- Recorded as ADR-0174 errata 37 ("a separate display gap") while fixing #3924; this is
  the issue that entry means to point at.
- Deliberately kept out of the #3924 PR: it is a one-line change, but it moves every
  golden and diagnostic assertion that spells a rank-1 array the CLR way, so it wants
  its own change with that sweep in it.
[...]

Issue #3958, opened 2026-09-05
gsc: a conversion targeting a nullable directional channel ('in chan[T]?' / 'out chan[T]?') is not supported by the emitter


DavidObando:
## The gap

Every conversion whose **target is a nullable directional channel** — `in chan[T]?` or
`out chan[T]?` — binds and then fails in the emitter.

```gsharp
package P
import System

var l out chan[int32]? = chan[int32](1)
Console.WriteLine("ok")
```

```
error GS9998: NotSupportedException: Conversion from
  'Gsharp.Concurrency.Chan`1[[System.Int32, System.Private.CoreLib, ...]]'
  to 'out chan[int32]?' is not yet supported by the emitter.
```

Isolated against its three neighbours, same compiler build (`4e8ec96d`):

| declaration | assigned | result |
|---|---|---|
| `out chan[int32]` | `chan[int32](1)` | **OK** |
| `chan[int32]?` | `chan[int32](1)` | **OK** |
| `out chan[int32]?` | `chan[int32](1)` | **GS9998** |
| `in chan[int32]?` | `chan[int32](1)` | **GS9998** |

So it is neither the directional head on its own nor the nullable wrapper on its own —
it is the two together. It does not depend on the element type (`int32` and `[]int32`
both fail), on the syntactic form of the source (a `chan[T](n)` expression and a
variable already typed `Chan[T]` both fail), or on the slot (a local, a global and a
`struct` field all fail).

## Why it matters

ADR-0159's own diagnostic recommends this exact spelling. Declare a channel-typed field
with no initializer and GS0520 says:

> Channel-typed declaration 'sink' requires an explicit initializer — an auto-created
> channel has no sensible default (buffer size, ownership); initialize with
> `out chan[int32]()` (rendezvous) or `out chan[int32](capacity)`, **or declare the slot
> as `out chan[int32]?` if the channel is genuinely optional (ADR-0159)**.

Taking that advice produces a slot nothing can be assigned into. An optional send-only
or receive-only endpoint — the ordinary shape for a component that may or may not have
been wired up — is not expressible today.

## Where it likely is

`Conversion.TryClassifyChannelConversion` deliberately declines a nullable operand:

```csharp
// A nullable wrapper on either side is not a channel conversion; the
// lifted rules handle it over the underlying pair.
if (from is NullableTypeSymbol || to is NullableTypeSymbol)
{
    return false;
}
```

The lifted rules then classify it, so binding succeeds. But the emitter's channel arm
[...]

Issue #3876, opened 2026-09-03
gsc: imported-constructor applicability does not see a `chan[T]` argument as `Channel[T]` when the element type is open


DavidObando:
Found while implementing ADR-0174 D10 (`chunks`, PR on `adr-0174/phase1-runtime-scaffold`). Worked around there with a static factory; recorded as ADR-0174 errata 32.

A `chan[T]` **is** `System.Threading.Channels.Channel<T>` — the conversion is identity (`Conversion.cs`). Overload resolution honours that when the element type is closed, and the imported **static-method** path honours it even when the element is open. The imported **constructor** path does not: with an open element it finds no applicable overload.

## Repro (measured on `b20ab60a`)

`ChunkReader<T>` is `Gsharp.Concurrency.ChunkReader<T>`, whose constructors are `(Channel<T>, int)` and `(ChannelReader<T>, int)`.

```gs
package Demo
import System
import Gsharp.Concurrency

func wrap[T](source chan[T]) int32 {
    let reader = ChunkReader[T](source, 64)
    return 0
}

Console.WriteLine(wrap[int32](chan[int32](4)).ToString())
```

```
error GS0267: No overload of 'ChunkReader' is applicable to the given argument list.
```

## Three controls, all of which compile and run

1. **Same constructor, closed element.** Only `T` → `int32` changes:

```gs
func wrapInt(source chan[int32]) int32 {
    let reader = ChunkReader[int32](source, 64)
    return 0
}
```

2. **Same open element, same parameter type, but a static generic method instead of a constructor.** `Chunks.Of<T>(Channel<T>, int)` takes exactly the argument the constructor rejected:

```gs
func wrap[T](source chan[T]) int32 {
    let reader = Chunks.Of[T](source, 64)   // binds
    return 0
}
```

3. **A user-declared generic class with a `chan[T]` constructor parameter and an open element** binds fine, so it is not the constructor form by itself:

```gs
class Holder[T] {
    var source chan[T]?
    init(source chan[T]) { this.source = source }
}

func wrap[T](source chan[T]) int32 {
    let h = Holder[T](source)   // binds
    return 0
}
```

And `List[T](items)` with an open `T` binds, so it is not open generics in general either.

## Expected

The imported-constructor applicability check should classify a `chan[T]` argument against a `Channel[T]` parameter the same way the imported-static-method check already does, whether or not the element type is open.

## Notes

[...]

PR #3960, opened 2026-09-05
gsc: a chan[T] whose element is a structural type converts to in/out chan[T] (#3924)

### LINKED ISSUE #3924: gsc: a chan[T] whose element is an array does not convert to out chan[T] / in chan[T]
## The gap

A `chan[T]` does not convert to `out chan[T]` (or `in chan[T]`) when the element type is an **array**.
The identical conversion works for every scalar element I tried.

```gsharp
package Probe

func take(ch out chan[[]int32]) {
}

func Main() {
    let c = chan[[]int32](4)
    take(c)
}
```

```
error GS0154: Parameter 'ch' requires a value of type 'out chan[[]int32]'
              but was given a value of type 'Gsharp.Concurrency.Chan[int32[]]'.
```

Isolated against the two neighbouring cases, same compiler build:

| declaration | argument | result |
|---|---|---|
| `out chan[int32]` | `chan[int32](4)` | **OK** |
| `chan[[]int32]` | `chan[[]int32](4)` | **OK** |
| `out chan[[]int32]` | `chan[[]int32](4)` | **GS0154** |

So it is neither the array element type on its own nor the directional head on its own — it is the two
together. Note the message's own asymmetry: it prints the parameter as `out chan[[]int32]` and the
argument as `Chan[int32[]]`, i.e. the element renders as `[]int32` in the type-clause spelling and
`int32[]` in the constructed one.

## Why it matters

ADR-0174 D2 makes `in`/`out` the ordinary way to express direction on a channel parameter, and a
producer function taking `out chan[T]` is the idiomatic shape — `bench/concurrency/gsharp/Bench.gs`
already has `func produce(ch out chan[int32], …)`. A producer of *batches* cannot be written that way
today.

It also blocks the honest counterpart to Go's `chan []int` in the D11 benchmark: the array-transport
rows added for issue #3902 S1d had to declare `ch chan[[]int32]` bidirectionally, with a comment
pointing here, because the `out` form does not bind.

## Related

This looks like the same family as ADR-0174 errata 28 and issue #3876 — applicability and view
conversions not firing for a `chan[T]` whose element is not a plain named type. #3876 is the
open-element constructor case; this is the array-element directional case. They may share a root.

Filed while doing #3902 S1d.

DavidObando:
Fixes #3924.

## The gap

ADR-0174 D2 makes `in`/`out` the ordinary way to express direction on a channel
parameter, but the narrowing did not bind when the element was a **structural** G# type
— `[]int32`, `map[string,int32]`, `[][]int32` — while the identical narrowing over
`int32`, `string` or a same-compilation `struct` bound fine. A producer of *batches*
could not be written `out chan[[]int32]`, which is why the D11 array-transport rows
added for #3902 S1d declare their producer bidirectionally with a comment pointing at
the issue.

## Root cause

Errata 3: the value `chan[T](n)` produces is the runtime class, so the conversion's
source element is whatever `TypeSymbol.FromClrType` mints for the **reflected** type
argument — and a structural shape has no CLR counterpart symbol. `[]int32` comes back as
an `ImportedTypeSymbol` over `int32[]`, never the `SliceTypeSymbol` the type clause
bound. The direction lattice compares elements by symbol shape, so it saw two different
types. (It is also where the issue's rendering asymmetry comes from: one side has a
`SliceTypeSymbol` to print, the other only a CLR type.)

## The fix

`Conversion.TryClassifyChannelConversion` now falls back to comparing the elements'
**effective** CLR types, which is the only question `Channel<T>`'s invariance asks.

This is not merely a fix for what `chan[T](n)` builds: a foreign `Channel<int32[]>` from
C# failed identically, and its element exists only as reflected metadata, so no amount
of preserving G# symbols at construction would have reached it. D2's identity
transparency to foreign channels is the reason the fix belongs at the comparison.

Two deliberate narrowings:

- **Effective** CLR type (`NullableTypeSymbol.GetEffectiveClrType`, the same projection
  `ChannelRuntimeBinder.ProjectElement` uses), because a `NullableTypeSymbol` borrows its
  underlying type's `ClrType` — only the lifted form keeps `chan[int32?]` distinct from
  `chan[int32]`.
- The fallback applies **only when one side is the metadata-recovered form**: an imported
  symbol carrying no symbolic type arguments, which is what `FromClrType` mints and what
  the source-constructed `GetConstructed` does not. A CLR type is coarser than the G#
  symbol above it — `[3]int32` and `[4]int32` are both `int32[]`, and `List[Foo]` and
  `List[Bar]` over two same-compilation structs are one `List<object>` — so an element
  that still carries its symbols is left to the finer comparison: `chan[[3]int32]` does
  not narrow to `out chan[[4]int32]`, nor `chan[List[Bar]]` to `out chan[List[Foo]]`.
  Past a construction the distinction is already gone, and was before this change:
  `chan[[3]int32](1)` and `chan[[4]int32](1)` build one `Chan<int32[]>`, which a
  bidirectional `chan[[4]int32]` parameter has always accepted.

One classification-kind change with no IL consequence: a constructed `Chan[T]` to a
bidirectional `chan[T]` with a structural element previously reached `Implicit` through
the general subclass rule and is now `Identity`. `EmitConversion`'s channel arm no-ops on
a same-direction pair either way.

## Discrimination (ADR-0154)

Ten new cases in `test/Compiler.Tests/Issue3924DirectionalChannelStructuralElementTests.cs`.
Every executable case compiles, **IL-verifies and runs**, so `get_Reader`/`get_Writer` is
proven emitted rather than merely bound — a bind-only assertion cannot tell the two apart.

| run | result |
|---|---|
| `Conversion.cs` reverted to the parent commit | **5 failed / 7 passed** — slice, string-slice, nested-slice, map and foreign-BCL-channel cases red with GS0154 |
| with the fix | **12 passed** |
| guard widened to "any imported symbol" | **2 failed** — `source-constructed-generic-elements-are-not-conflated`, `fixed-array-length-inside-a-constructed-generic-element` |
| guard dropped entirely | **1 more failed** — `fixed-array-lengths-still-discriminate` |

Four controls are green on both sides and are there to fail a blanket fix:
`struct-slice-element` (`[]Point`, the symbolic-type-argument branch, which always
bound), `fixed-array-lengths-still-discriminate`, and the two source-constructed-generic
cases added from review feedback. The other rejection cases pin a mismatched element
(`[]int64` for `[]int32`), a nullable element (`int32?` for `int32` — the case that
forces the effective CLR type), and `in` → `out`.

## Also in this PR

- `bench/concurrency/gsharp/Bench.gs`: `produceArrays` restored to `out chan[[]int32]`,
  the workaround comment removed. `python3 build/verify-concurrency-bench.py --smoke`
  passes and `chunk64-arrays` measures 5.70 ns/op on this machine.
- ADR-0174 errata 37 records the refinement and the guard's boundary.

## Test runs

`Compiler.Tests` 4754 passed, `Core.Tests` 8900 passed. The 7 failures across the two
suites share one cause — `targeting pack root '~/.dotnet/packs/NETStandard.Library.Ref'
missing` — and throw in fixture setup before anything is compiled; they are unrelated to
this change and fail identically on the parent commit.

## Related, not touched

- **#3876** — the imported-**constructor** applicability path with an open element. The
  issue wondered whether it shares a root; checked, and it does not.
- **#3959** — filed from this work: the CLR fallback renderer spells a rank-1 array
  `int32[]` rather than G#'s `[]int32`, which is the display half of the asymmetry #3924
  noticed. One line, but it moves every golden asserting the CLR spelling, so it is kept
  separate.
- **#3962** — filed from the review of this PR: a constructed generic over an
  array-shaped type argument compares by its erased CLR type, so `List[[3]int32]`
  converts to `List[[4]int32]`. Pre-existing (the repro has no channel in it and compiles
  on the parent commit) and left alone here; this PR's guard is deliberately strict
  enough not to add a second path to it.
- **#3958** — filed from this work: a conversion targeting `in chan[T]?` / `out chan[T]?`
  is not supported by the emitter. Pre-existing, independent of the element type, and
  ADR-0159's own GS0520 message recommends that spelling.

### DIFF
--- a/bench/concurrency/gsharp/Bench.gs
+++ b/bench/concurrency/gsharp/Bench.gs
@@ -329,10 +329,7 @@ func chunkedArrays(size int32, count int32) TimeSpan {
     return sw.Elapsed
 }
 
-// `chan[[]int32]` rather than `out chan[[]int32]`: the directional view
-// conversion does not apply when the element type is an array, so the `out`
-// form does not bind (issue #3924). Bidirectional is only a workaround here.
-func produceArrays(ch chan[[]int32], count int32, size int32) {
+func produceArrays(ch out chan[[]int32], count int32, size int32) {
     var sent = 0
     while sent < count {
         var chunk = [size]int32{}
--- a/docs/adr/0174-goroutines-and-channels-wave-2.md
+++ b/docs/adr/0174-goroutines-and-channels-wave-2.md
@@ -2665,6 +2665,38 @@ implementation had to refine it.
     reader does not spend the effort on channels that the number appears to
     demand.
 
+37. **The direction lattice compares elements by runtime type, not by symbol
+    shape (issue #3924).** D2 says a `chan[T]` narrows to `in chan[T]` /
+    `out chan[T]`. It did not when `T` was a *structural* G# type — `[]int32`,
+    `map[string,int32]`, `[][]int32` — while the same narrowing over `int32`,
+    `string` or a same-compilation `struct` bound fine. The cause is errata 3:
+    the value a `chan[T](n)` expression produces is the runtime class, so the
+    conversion's source element is whatever `TypeSymbol.FromClrType` mints for
+    the *reflected* type argument, and a structural shape has no CLR
+    counterpart symbol — `[]int32` comes back as an imported `int32[]`, never
+    the `SliceTypeSymbol` the type clause bound. The lattice compared those two
+    symbols and saw two different types. It now falls back to comparing the
+    elements' effective CLR types, which is the only question `Channel<T>`'s
+    invariance actually asks. This is not merely a fix for what `chan[T](n)`
+    builds: a foreign `Channel<int32[]>` from C# failed identically, and its
+    element exists only as reflected metadata, so no amount of preserving G#
+    symbols at construction would have reached it. With it, the D11
+    array-transport rows (`chunk64-arrays`, `chunk1k-arrays`) declare their
+    producer `out chan[[]int32]`, as errata 28's `merge` discussion intends.
+    The fallback applies only when one side IS that metadata-recovered form.
+    A CLR type is coarser than the G# symbol above it — `[3]int32` and
+    `[4]int32` are both `int32[]` — so where both sides still carry their
+    symbols the finer comparison stands: `chan[[3]int32]` still does not
+    narrow to `out chan[[4]int32]`. Past a construction the distinction is
+    already gone and was before this change: `chan[[3]int32](1)` and
+    `chan[[4]int32](1)` build one `Chan<int32[]>`, which a bidirectional
+    `chan[[4]int32]` parameter has always accepted.
+
+    The `Chan[int32[]]` spelling in the pre-fix diagnostic is a separate
+    display gap, filed as issue #3959: the CLR fallback renderer spells a
+    rank-1 array in C# suffix form while rank ≥ 2 already uses G#'s prefix
+    form.
+
 ## Addendum A — The ten patterns, three ways
 
 The pattern study in the Context section gives ratings. This addendum gives
--- a/src/Core/CodeAnalysis/Binding/Conversion.cs
+++ b/src/Core/CodeAnalysis/Binding/Conversion.cs
@@ -3152,7 +3152,8 @@ public sealed class Conversion
             return false;
         }
 
-        if (!IsCrossContextIdenticalElement(fromElement, toElement))
+        if (!IsCrossContextIdenticalElement(fromElement, toElement)
+            && !AreSameRuntimeChannelElement(fromElement, toElement))
         {
             return false;
         }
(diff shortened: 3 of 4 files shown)


DavidObando:
Good catch on the shape — the guard was too loose, and it is tightened in 4d5882eb. Two corrections to the specifics, because the example given and the actual regression are different bugs.

**The cited example is pre-existing, not introduced here.** `chan[List[[3]int32]]` narrowing to `out chan[List[[4]int32]]` compiles on the parent commit too (built `ffa46214`'s `Conversion.cs`, with `chan[[3]int32]` → `out chan[[4]int32]` and the #3924 slice case as controls to confirm the binary was really the pre-fix one). It is not channel-specific either — plain `var l4 List[[4]int32] = l3` binds, no channel in sight, and `Conversion.TryClassifyChannelConversion` is only reached when one side is a `chan[T]` type clause.

Lines 188–192 are also not the guard against it; they are the path that accepts it. `HasSubstitutableTypeArgument` is what escalates to the finer `AreRuntimeEquivalentIgnoringReferenceNullability` comparison — the one that compares `ArrayTypeSymbol.Length` — and an `[N]T` argument does not trip it (it has a CLR type, needs no symbolic projection, is not a nested constructed generic, is not a named tuple). So the escalation is skipped and `AreSame(List<int32[]>, List<int32[]>)` answers `Identity`. Filed as #3962 with that analysis.

**The regression the reasoning does find is worse than the example.** Following it to a *same-compilation* argument:

```gsharp
struct Foo { var X int32 }
struct Bar { var Y int32 }

func wants(w out chan[List[Foo]]) {}

var barChan chan[List[Bar]] = chan[List[Bar]](1)
wants(barChan)          // accepted with the loose guard
```

Two entirely different structs. Both `List[Foo]` and `List[Bar]` close over `object` while binding, so both `ClrType`s are `List<object>` — and unlike the `[N]` case this one was correctly rejected before (ordinary assignment still rejects `List[Bar]` → `List[Foo]` with GS0155). That is the real hole.

**The fix,** which is the remediation suggested, phrased against the boundary the comment states:

```csharp
private static bool IsMetadataRecoveredElement(TypeSymbol type)
    => type is ImportedTypeSymbol { TypeArguments.IsDefaultOrEmpty: true };
```

"Came back from reflection" means an imported symbol carrying **no** symbolic type arguments — precisely what `FromClrType`/`ImportedTypeSymbol.Get` mints, and what the source-constructed `GetConstructed` does not. A source-constructed element still holds its arguments as symbols, so it is left to the symbol comparison that can read them. This is deliberately stricter than "no *substitutable* arguments": it also keeps the fallback away from `List[[3]int32]`, so the PR adds no second path to #3962's answer.

**Tests.** Both shapes are in the rejection matrix — `source-constructed-generic-elements-are-not-conflated` and `fixed-array-length-inside-a-constructed-generic-element`. The `List[[3]int32]`/`List[[4]int32]` case is not, because it compiles identically before and after this PR; a test asserting rejection there would be red on `main`, and it belongs to #3962.

Witness, now twelve cases: 5 red / 7 green on the reverted `Conversion.cs`, 12 green with it. Widening the guard back to "any imported symbol" turns exactly the two new cases red; dropping it entirely turns a third red (`fixed-array-lengths-still-discriminate`).

Issue #2051, opened 2026-07-04
gsc: implicit T -> Nullable<T> wrap dropped for ValueTuple value types (StackUnexpected/ilverify)


DavidObando:
Discovered while validating #1943 (cs2gs typed positional subpattern + nullable tuple guard fix).

Minimal repro (no pattern matching involved):
```
package Repro
var p (int32, int32)? = (0, 0)
```

Compiles successfully, but `ilverify` fails:
```
[IL]: Error [StackUnexpected]: [...][found value 'System.ValueTuple`2<int32,int32>'][expected value 'System.Nullable`1<System.ValueTuple`2<int32,int32>>'] Unexpected type on the stack.
```

Also reproduces when passing a tuple literal as an argument to a `(int,int)?`-typed parameter (`h.Test((0,0))`).

Root cause: gsc's implicit conversion from a non-nullable value-type tuple literal to its `Nullable<T>` form omits the `newobj Nullable<T>::.ctor(T)` wrap at the conversion site — the raw `ValueTuple` is left on the stack where a `Nullable<ValueTuple>` is expected. Verified this is NOT specific to `!!`/pattern-matching: a bare local-variable initializer trips it. Nullable value-type *scalars* (e.g. `int32?`) are unaffected — this appears specific to `ValueTuple`-shaped value types.

This is out of scope for #1943 (cs2gs translator fix) since it's a pre-existing gsc emitter gap unrelated to positional-pattern translation; #1943's corpus fixture intentionally avoids exercising nullable value-type tuples end-to-end to sidestep this bug (the null-guard fix itself is validated at the translator-unit level instead).

Likely fix location: `src/Core/CodeAnalysis/Emit` — wherever implicit `T -> Nullable<T>` conversions are emitted for value-type operands (see `Nullable<T>` wrap handling around issue #504/#1572 in `MethodBodyEmitter.Operators.cs`), needs a ValueTuple-shaped-T case.

PR #3961, opened 2026-09-05
gsc: emit a name-differing tuple instantiation as the identity it is (ADR-0172, #3956)

### LINKED ISSUE #3956: gsc: NotSupportedException converting KeyValuePair[string, (int,int,List[(int,int,int)])] to its named-tuple form — the only compile error left in migrated test/Core.Tests
`test/Core.Tests` is down to a **single** compile fingerprint in the #3501 gate (runs 33968896644 and 33973595210), and it is a gsc internal exception rather than a diagnostic:

```
GS9998: NotSupportedException: Conversion from
  'System.Collections.Generic.KeyValuePair`2[[System.String],
     [System.ValueTuple`3[[System.Int32],[System.Int32],
        [System.Collections.Generic.List`1[[System.ValueTuple`3[[System.Int32],[System.Int32],[System.Int32]]]]]]]]'
to
  'System.Collections.Generic.KeyValuePair`2[string,
     (Rid int32, CatchHandlerPlusOne int32, Awaits System.Collections.Generic.List`1[(Yield int32, Resume int32, ResumeRid int32)])]'
```

Source and target are the **same shape**; they differ only in that the target carries tuple element *names* (`Rid`, `CatchHandlerPlusOne`, `Awaits`, `Yield`, `Resume`, `ResumeRid`), nested two levels deep inside a `KeyValuePair` and a `List`. Per ADR-0172 tuple element names are shape-not-identity, so this conversion should be an identity conversion; instead the conversion walker reaches a case it does not handle and throws.

Worth noting for whoever takes it: this is the *last* compile error for a very large app, so clearing it does not by itself make `test/Core.Tests` green — compile, ILVerify and test-parity all still have to pass behind it, and the parity surface is unmeasured. See also #3684.

Refs #3501.

DavidObando:
Fixes #3956.

## Root cause

`KeyValuePair[string, (int32, int32, List[(int32, int32, int32)])]` recovered from a CLR signature and the same instantiation written with tuple element *names* are **one type** — ADR-0172 makes element names metadata over the positional shape — and `Conversion.Classify` already returns **Identity** for the pair. (I confirmed this by instrumenting the classifier: it was never the culprit, so no classifier change was needed.)

The emitter did not agree. Its identity fast path is `if (from == to)` — symbol **reference** equality — which two independently minted symbols for one CLR type never satisfy. Here `from` arrived as a `NullabilityAnnotatedTypeSymbol` over an `ImportedTypeSymbol` with **no symbolic type arguments** (minted straight from the CLR `Type`), while `to` was an `ImportedTypeSymbol` whose second type argument is a `TupleTypeSymbol` carrying the names. Every structural arm below the fast path keys off a specific symbol shape and declined, and emit fell through to the `GS9998` internal-error channel.

**What was keyed on identity that should have been keyed on shape:** the emitter's `from == to` no-op arm. It is the only identity test on the emit path, and it is a reference test.

## Fix

One arm, immediately before the `NotSupportedException` fallback: when both sides have a CLR backing, `ClrTypeUtilities.AreSame` agrees, **and** the classifier calls the conversion Identity, emit nothing — the conversion is representation-preserving, exactly like `from == to`.

Both halves of the guard are load-bearing. A `NullableTypeSymbol` relays its *underlying* type's `ClrType` (#2051), so the CLR check alone would let a genuine `T -> Nullable<T>` lift through unwrapped; the identity check alone would admit an alias whose backing legitimately differs. Placing the arm **last** means only programs that previously crashed change behaviour.

## Failure mode

The `NotSupportedException` is already contained — `EmitDiagnosticException.Wrap` reports it as `error GS9998` and gsc exits non-zero. It is the emitter's deliberate internal-error channel (there is no `DiagnosticBag` in scope on the emit path), not an unhandled crash, so no separate change was warranted there.

## Tests

`test/Compiler.Tests/Emit/Issue3956NamedTupleGenericArgumentIdentityTests.cs` — three cases, each of which **compiles, ILVerifies AND executes** (an emitter defect of this kind binds clean and still produces a body the verifier or the runtime rejects):

1. the issue's shape verbatim (`Enumerable.Single` over a `Dictionary` whose value is a named tuple holding a `List` of named tuples);
2. the reverse direction — named source into an unnamed target;
3. a **nested-only** name difference, where both sides spell the outer tuple's names identically and differ only inside the `List` element two levels down. A top-level-only fix passes 1 and 2 and fails this one.

All three fail on `main` with the exact `GS9998` from the issue (`Failed! - Failed: 3, Passed: 0`) and pass with the fix (`Passed! - Failed: 0, Passed: 3`).

## Scope

No name-propagation audit gaps surfaced beyond this one — the classifier's ADR-0172 handling was already correct at every level I probed, including nested. I deliberately reverted an earlier, broader experiment (relaxing `ImportedTypeSymbol.HasSubstitutableTypeArgument` for the conversion-identity question) once instrumentation showed classification was not the defect.

### DIFF
--- a/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Conversions.cs
+++ b/src/Core/CodeAnalysis/Emit/MethodBodyEmitter.Conversions.cs
@@ -710,6 +710,33 @@ internal sealed partial class MethodBodyEmitter
             }
         }
 
+        // Issue #3956 / ADR-0172: the `from == to` fast path at the top of this
+        // method is symbol REFERENCE equality, but one CLR type is routinely
+        // minted as several non-reference-equal symbols — a nullability-
+        // annotated wrapper on one side, a `ValueTuple<…>` recovered from a CLR
+        // signature versus a `TupleTypeSymbol` carrying element NAMES on the
+        // other. Element names are metadata over the positional shape and never
+        // change the runtime type, so `KeyValuePair[string, (int32, int32,
+        // List[(int32, int32, int32)])]` and the same shape written with names
+        // are one type, and the binder duly classifies the conversion between
+        // them as IDENTITY — yet every structural arm above keys off a specific
+        // symbol shape and declines, leaving the crash below as the only
+        // outcome. Ask the classifier the SHAPE question the reference check
+        // cannot: when the two symbols denote the same type AND agree on their
+        // CLR backing, the conversion is representation-preserving and needs no
+        // IL at all, exactly like the `from == to` case. Both halves are
+        // load-bearing — a `NullableTypeSymbol` relays its UNDERLYING type's
+        // `ClrType` (#2051), so the CLR check alone would let a genuine
+        // `T -> Nullable<T>` lift through unwrapped, and the identity check
+        // alone would admit an alias whose backing legitimately differs.
+        if (from.ClrType != null
+            && to.ClrType != null
+            && ClrTypeUtilities.AreSame(from.ClrType, to.ClrType)
+            && Conversion.Classify(from, to).IsIdentity)
+        {
+            return;
+        }
+
         EmitDiagnosticException.Wrap(
             conv.Syntax,
             new NotSupportedException(
--- a/test/Compiler.Tests/Emit/Issue3956NamedTupleGenericArgumentIdentityTests.cs
+++ b/test/Compiler.Tests/Emit/Issue3956NamedTupleGenericArgumentIdentityTests.cs
@@ -1,0 +1,199 @@
+// <copyright file="Issue3956NamedTupleGenericArgumentIdentityTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using GSharp.Core.CodeAnalysis.Symbols;
+using Xunit;
+
+namespace GSharp.Compiler.Tests.Emit;
+
+/// <summary>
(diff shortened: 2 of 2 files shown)


DavidObando:
## Verification

**CI: 37/37 checks pass, zero failures** — including `hot-core translation guard` (this change is in `src/Core`, so it must itself be translatable), `ilverify`, `e2e`, `cs2gs-corpus`, `cs2gs-oahu`, `cs2gs-code-exploder`, and every `tests (…)` shard. `tests (interpreter)` and `tests (core-binding)` / `tests (core-remainder)` are green, which is the blast radius worth naming: the REPL enters the binder through a different entry point, and this change sits on a path the interpreter shares.

**Migrated `test/Core.Tests` now COMPILES.** A full local self-migration (`build/run-cs2gs-selfmig-migrate.sh`, whole corpus, 52/53 apps translated) followed by `run-cs2gs-selfmig-validate.sh … test/Core.Tests/Core.Tests.csproj`:

```
    358 Warning(s)
    0 Error(s)
```

`grep -c GS9998` over that build log is **0** — the fingerprint from this issue is gone, and nothing was hiding behind it. (The count reached 0 only after the `!!`-polish loop converged; the intermediate rounds are all `GS0536` redundant-`!!`, which is that loop's own work, not a compile wall.)

Per the issue's own caveat, compiling is not the same as green: ILVerify and test-parity for that app are still unmeasured here, so I am not quoting a parity number — a count without its `Failed! … Total: N` summary line would be a truncation artifact. `tools/cs2gs/selfmig-baseline.json` is deliberately untouched.

PR #3964, opened 2026-09-05
gsc: a variadic carrier passed whole binds in normal form during overload ranking (#3880)

DavidObando:
## Root cause

C# §12.6.4.2: when a single trailing argument already **is** the params carrier, the candidate is applicable in **normal form** — that argument ranks against the carrier type, not the element type.

gsc's user-symbolic overload ranking (`OverloadResolver.SelectBestInstanceOverload`) classified that argument as an *expanded-form tail slot* and compared it to the element type. Two failures compounded:

1. the conversion kind was mis-rated (`[]string` against `string`), and
2. `ParamTypes[i]` is only populated on the non-tail branch, so the `Reference` "better conversion target" tie-break in `CompareUserConversions` was skipped (its `paramA != null && paramB != null` guard failed).

Both siblings landed on `Reference` with null parameter types, tied at 0, and then tied on every later axis — defaulted-parameter count, variadic-ness, genericity — so `GS0266` was reported.

Minimal repro (compiled with a locally built `gsc`):

```gs
class C {
    shared {
        func F(printedSources ...string) int32 { return 1 }
        func F(additionalReferencePaths IReadOnlyList[string]?, printedSources ...string) int32 { return 2 }
    }
}
let arr = []string{"a", "b"}
C.F(arr)   // GS0266 before; binds to overload 1 (as C# does) after
```

This is exactly the migrated `tools/cs2gs/Cs2Gs.Tests` pair `AssertBindsAgainstGsCore` at `Adr0169AnalyzerTranslationTests.gs` lines 788 and 820 — 2 sites, 1 of the 5 gate fingerprints on that app.

## Fix

The applicability filter already recognised this shape via `IsVariadicCarrierPassThrough`. Ranking now uses the same test: when the call binds the carrier whole, the slot is ranked as a normal-form slot against the carrier type.

## Verification

- New `test/Compiler.Tests/Emit/Issue3880VariadicPassThroughOverloadEmitTests.cs` — **executes**, so the *selected* overload is witnessed at runtime, not merely bound. ILVerify runs on the emitted assembly.
- `Core.Tests` filtered `Overload|Variadic|Params`: 479 passed.
- `Compiler.Tests` filtered `Overload|Variadic|Params`: 196 passed.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.Candidates.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.Candidates.cs
@@ -461,6 +461,28 @@ internal sealed partial class OverloadResolver
                 && variadicElement != TypeSymbol.Error
                 ? variadicElement
                 : null;
+
+            // Issue #3880: a single trailing argument that already IS the
+            // variadic carrier (`F(existingArray)` against `F(xs ...string)`)
+            // binds the candidate in NORMAL form, not expanded form — the same
+            // pass-through the applicability filter recognises via
+            // IsVariadicCarrierPassThrough. Ranking used to classify it as an
+            // expanded tail slot against the ELEMENT type, which both mis-rates
+            // the conversion (`[]string -> string`) and leaves ParamTypes[i]
+            // null, so the Reference "better conversion target" tie-break could
+            // not run and two variadic siblings
+            // (`F(xs ...string)` / `F(refs IReadOnlyList[string]?, xs ...string)`)
+            // tied and reported a spurious GS0266 where C# picks the first.
+            var variadicCarrierType = isVariadic
+                ? cand.Parameters[cand.Parameters.Length - 1].Type
+                : null;
+            var bindsCarrierInNormalForm = isVariadic
+                && !HasNamedArguments(argumentNames)
+                && argumentCount - paramCountForScore == 1
+                && paramCountForScore < boundArguments.Count
+                && boundArguments[paramCountForScore]?.Type is { } carrierPassThroughArgType
+                && variadicCarrierType != null
+                && IsVariadicCarrierPassThrough(carrierPassThroughArgType, variadicCarrierType);
             for (var i = 0; i < boundArguments.Count; i++)
             {
                 var slot = MapArgumentIndexToParameterSlot(cand, argumentNames, i, parameterOffset, paramLen);
@@ -486,8 +508,21 @@ internal sealed partial class OverloadResolver
                     // one purely because its tail was compared favourably.
                     // Between two variadic candidates (both expanded), the
                     // element-type kind still decides genuine betterness.
-                    isTailSlot[i] = true;
                     var tailArgType = boundArguments[i]?.Type;
+                    if (bindsCarrierInNormalForm)
+                    {
+                        // Normal form (issue #3880): rank against the carrier
+                        // itself, in a normal-form slot, so an exact `[]string`
+                        // beats a sibling's `IReadOnlyList[string]?` widening.
+                        var normalFormCarrier = Invariant.Required(
+                            variadicCarrierType,
+                            "a carrier pass-through candidate has a variadic carrier parameter");
+                        paramTypes[i] = normalFormCarrier;
+                        kinds[i] = ClassifyUserArgumentConversionKind(tailArgType, normalFormCarrier);
+                        continue;
+                    }
+
+                    isTailSlot[i] = true;
                     kinds[i] = elementType == null
                         ? ClrOverloadResolution.ImplicitConversionKind.Identity
                         : ClassifyUserArgumentConversionKind(tailArgType, elementType);
--- a/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.Candidates.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.Candidates.cs
@@ -473,9 +473,22 @@ internal sealed partial class OverloadResolver
             // not run and two variadic siblings
             // (`F(xs ...string)` / `F(refs IReadOnlyList[string]?, xs ...string)`)
             // tied and reported a spurious GS0266 where C# picks the first.
+            // The carrier must be CLOSED before it is compared to the argument.
+            // A generic variadic candidate (`func Take[T](values ...T)`) whose
+            // inference already ran declares the carrier as the open `[]T`;
+            // ranking `[]string` against `[]T` classifies an identity as a
+            // reference conversion and hands the call to a fixed
+            // `IEnumerable[string]` sibling that C# does not choose. The
+            // non-tail slots below already substitute through candSubstitution
+            // for exactly this reason.
             var variadicCarrierType = isVariadic
                 ? cand.Parameters[cand.Parameters.Length - 1].Type
                 : null;
+            if (variadicCarrierType != null && candSubstitution != null)
+            {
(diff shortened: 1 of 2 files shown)


DavidObando:
## Review finding on `OverloadResolver.Candidates.cs:478/480` — **confirmed, and fixed here**

The claim was that a generic variadic candidate is ranked against the **open** carrier. It is correct, and the stated consequence is correct too: it flips the selection.

### csc oracle (net10.0)

```csharp
static string Take<T>(params T[] values) => "generic-carrier";
static string Take(IEnumerable<string> values) => "sequence";

static string Two<T>(params T[] values) => "generic-carrier";
static string Two(IReadOnlyList<string> refs, params string[] values) => "two-param";

var arr = new string[] { "a" };
Take(arr);   // generic-carrier
Two(arr);    // generic-carrier
```

Both print `generic-carrier`: after inference `T = string`, the carrier pass-through is an **identity** conversion, which beats the sibling's implicit reference conversion before any generic-vs-non-generic tie-break is ever consulted.

### gsc, same shapes

| gsc build | `Take(arr)` | `Two(arr)` |
|---|---|---|
| `origin/main` (no PR) | `sequence` | `two-param` |
| this PR as first pushed | `sequence` | `two-param` |
| this PR + carrier substitution | `generic-carrier` | `generic-carrier` |

Two things worth stating precisely, because they pull in opposite directions:

1. **This is not a regression introduced by the PR.** `origin/main` already picks the wrong overload for both shapes — before the fix the tail slot was classified against the open ELEMENT type `T`, which ranked even worse, so the sibling won there too. The PR neither caused it nor fixed it.
2. **The PR is nonetheless the right place to fix it**, because the PR is what makes `ParamTypes[i]` load-bearing for the tail slot. It introduced the read of the carrier type into ranking; reading it open is a defect in the code this PR adds, whatever the prior behaviour happened to be.

### Fix

The carrier is now closed through the candidate's inferred `candSubstitution` before both the pass-through test and the `paramTypes[i]` store — the same substitution the non-tail slots a few lines below have always applied. Three lines, no change to carrier semantics, so ADR-0173 (`...X[T]` ≡ `params X<T>`) is untouched: this is substitution order, not what a carrier *is*.

Deliberately scoped: the expanded-form `elementType` a few lines up is also open, which is pre-existing behaviour on a path where `ParamTypes` stays null and the reference tie-break therefore never runs. Widening that is a separate ranking change with its own blast radius and does not belong in this PR.

### Verification

`GenericCarrierPassThrough_ClosesTheCarrierBeforeRanking` **executes** the emitted program and asserts the printed overload, because this defect binds clean, ILVerifies clean and returns the wrong answer at runtime — a binding assertion would have passed throughout. It fails on `origin/main` and on the first push of this PR, and passes with the substitution.

Wide suites re-run on the combined change (a ranking difference does not announce itself): results posted below when they finish.

DavidObando:
### Suite results for the substitution fix

Local runs stayed targeted (the machine is shared with other agents' full suites; CI shards run the rest):

- `Core.Tests` filtered `Overload|Variadic|Params`: **479 passed**, 32s — identical count *and* identical duration to the pre-substitution run, so the extra `SubstituteType` / `Conversion.Classify` on the pass-through path costs nothing measurable in the ranking hot path.
- `Compiler.Tests` filtered `Overload|Variadic|Params|Issue3880`: **197 passed** (196 before, +1 new executing test).

Full `Core.Tests` / `Compiler.Tests` are left to CI.

### End-to-end evidence that this PR moves the wall

Ran `run-cs2gs-selfmig-migrate.sh` + `run-cs2gs-selfmig-validate.sh` locally for `tools/cs2gs/Cs2Gs.Tests` on this PR stacked with #3967, against a **repacked** `Gsharp.NET.Sdk` nupkg (`0.4.485-g89ed4b86b3`) so the gate really compiled through this compiler and not a stale one.

`sdk.build.log` for the migrated app went from **7 error sites / 5 gate fingerprints** to:

```
```

One site, one fingerprint — and that one is `GNodeSamples.All[type]()`, a different root cause, fixed in #3966. The two `GS0266 AssertBindsAgainstGsCore` sites this PR targets are gone.

DavidObando:
## The finding predates the fix — but re-reading it found a live twin

**Chronology.** The comment's `original_commit_id` is `4f7bbb577` (11:51:17 PDT), it was posted at 18:55:24Z, and the substitution landed in **`2b80a9794`** at 19:34:18Z — 39 minutes later. GitHub re-anchored `commit_id` to the new head because the lines still exist, which is why it reads as current. Lines 484-490 of the file it points at now *are* the fix:

```csharp
var variadicCarrierType = isVariadic ? cand.Parameters[^1].Type : null;
if (variadicCarrierType != null && candSubstitution != null)
{
    variadicCarrierType = Binder.SubstituteType(variadicCarrierType, candSubstitution);
}
```

and it precedes **both** sites the comment names — the pass-through test and the `paramTypes` store. Verified behaviourally, not just by reading: `Take[T](...T)` vs `Take(IEnumerable[string])` with a `[]string` argument prints `generic-carrier`, matching csc.

**But the principle had one more application, and I had argued it away wrongly.** In scoping the first fix I wrote that the expanded-form `elementType` was harmless because "ParamTypes stays null on that path, so the reference tie-break never runs." That reasoning does not hold: `CompareUserConversions` compares conversion **kinds** first and returns before `ParamTypes` is ever read. An expanded tail slot can misrank with no help from that tie-break.

csc oracle (net10.0):

```csharp
static string Elem<T>(params T[] values) => "generic-elem";
static string Elem(params object[] values) => "object-elem";
Elem("a", "b");   // generic-elem
```

| gsc | `Elem("a","b")` |
|---|---|
| `2b80a9794` (carrier fix only) | `object-elem` ❌ |
| `3a5e59849` (element fix) | `generic-elem` ✅ |

`string` against the open `T` and `string` against `object` tied on kind, then the non-generic-over-generic tie-break handed the call away. Fixed in `3a5e59849` by closing `elementType` through the same `candSubstitution`.

So: the comment as written was already addressed, but it was **not** a duplicate to wave off — taking it seriously surfaced a second wrong-overload selection of the same family.

**Verification.** `Compiler.Tests` `Overload|Variadic|Params|Issue3880`: 198 passed (+1). `Core.Tests` `Overload|Variadic|Params|Generic`: 1519 passed. The new test **executes** and asserts the printed overload — every defect in this PR binds clean and ILVerifies clean, so a binding assertion would have passed through all of them.

DavidObando:
Post-merge follow-up, since this PR merged while I was verifying.

**The finding as written was already fixed here.** Its `original_commit_id` is `4f7bbb577` and it was posted at 18:55:24Z; the carrier substitution landed in `2b80a9794` at 19:34:18Z. Confirmed on merged `main` (`8b41b91ed`) — `OverloadResolver.Candidates.cs:489` closes `variadicCarrierType` through `candSubstitution` before both sites the comment names.

**But the same rule had one more application, and it is still live on `main`.** The expanded-form `elementType`, computed one line above the carrier, is still open:

```gs
func Elem[T](values ...T) string   // "generic-elem"
func Elem(values ...object) string // "object-elem"
Elem("a", "b")
```

csc prints `generic-elem`; a `gsc` built from merged `main` prints `object-elem`. Filed and fixed in **#3969**.

I had explicitly scoped that out on this PR, arguing the expanded path was harmless because its `ParamTypes` stays null so the reference tie-break never runs. That was wrong: `CompareUserConversions` compares conversion **kinds** first and returns before `ParamTypes` is read, so an expanded tail slot misranks without that tie-break participating at all. The reviewer's instinct generalised further than the line it was attached to.

Issue #3954, opened 2026-09-05
gsc/cs2gs: a ValueTask-returning call auto-awaits, so migrated `.AsTask()` has no G# form (7 of the 15 Runtime.Channels.Tests sites)


DavidObando:
Found while triaging the #3501 gate (runs 33968896644 and 33973595210 — `test/Runtime.Channels.Tests` is `compile-error`, 15 deduplicated fingerprint **signatures**, not 15 sites).

Seven of the fifteen signatures are the same shape:

```
GS0159 ChannelBatchExtensionsTests.gs:72,63  let pending = reader.ReceiveBatch(buffer, atLeast: 4).AsTask()
GS0159 ChannelBatchExtensionsTests.gs:133,64 let sending = writer.SendBatch([]int32{1,2,3,4,5}).AsTask()
...  Cannot find function AsTask.
```
plus one cascade (`GS0129 '+=' is not defined for types 'int32' and '?'`).

## Root cause

Not a missing member. `ValueTask[T].AsTask()` resolves fine on its own:

```gs
func main() {
    let vt = ValueTask[int32](5)
    let t = vt.AsTask()          // compiles
}
```

The problem is the *call*. `ChannelBatchExtensions.ReceiveBatch` is declared `public static ValueTask<int> ReceiveBatch<T>(this ChannelReader<T>, Memory<T>, int, Context? = null)`, but under ADR-0174 gsc presents a `ValueTask`-returning import as a **suspending function returning the awaited value**. A 14-line repro (gsc 0.4.479+ffa462149a):

```gs
package Probe
import System
import System.Threading.Channels
import System.Threading.Tasks
import Gsharp.Concurrency

async func main() {
    let ch = Channel.CreateUnbounded[int32]()
    let buffer = [8]int32
    let vt ValueTask[int32] = ch.Reader.ReceiveBatch(buffer, atLeast: 1)
}
```
```
error GS0155: Cannot convert type 'int32' to 'System.Threading.Tasks.ValueTask[int32]'.
```

So the call's type is `int32`, and `.AsTask()` on it is meaningless — as is the whole migrated `.AsTask().WaitAsync(timeout)` idiom the C# tests use to bound a wait.

## What needs deciding

This is a **language-model collision**, not a missing overload, so it is not a one-line fix:

[...]

PR #3965, opened 2026-09-05
gsc: an awaitable keeps its C# meaning; only channel syntax awaits for you (#3954)

DavidObando:
Fixes the 7 `.AsTask()` signatures of #3954.

## What the issue actually is

The issue's root-cause sentence — "gsc presents a `ValueTask`-returning import as a suspending function" — is over-general, and the correction is what makes the fix tractable. The implicit await fires only on `[Gsharp.Concurrency.Suspending]` (`ExpressionBinder.Async.cs:914`); a plain BCL `ValueTask` import keeps C# semantics today (`let vt ValueTask[int32] = ch.Reader.ReadAsync()` compiles). Exactly **four** methods in the repo are hand-marked: D10's `ReceiveBatch`/`SendBatch` on `ChannelReader[T]`/`ChannelWriter[T]` plus the `Channel[T]` overloads. Everything else carrying the attribute is G#-emitted.

So the scope is four methods, not a class of types.

## The decision this implements

An awaitable type should mean in G# what it means in C#, and the compiler should await for you only where the **syntax** is a channel operation — `ch <- v`, `<-ch`, `select`, channel `for..in`. That principle lands on `await`, not on `[Suspending]`, and it needed a gap closed first.

**Half one — D4's `await g()` row, implemented.** ADR-0174 D4's normative table has always said `await g()` colours the CALLER suspending. The implementation said `await` requires `async func` (GS0132). An `await` in a plain `func` is now a suspension point like a channel operation: the binder leaves it in place, `SuspensionPointCollector` reports it, and the inference fixed point colours the function. Where inference may not go, it may not go for an `await` either — a boundary keeps its declared shape and the new **GS0574** asks the author for `async func` (an observable `Task[R]`) or `suspend func` (implicitly awaited, `R`). GS0132 keeps only the no-enclosing-function case.

**Half two — the four batch methods drop `[Suspending]`.** D10's `suspend func (ch in chan[T]) ReceiveBatch(…)` spelling is superseded: the batch surface is an ordinary `ValueTask[int32]` API and the G# caller writes the `await`. `X.AsTask()` works again, which is the idiom the C# tests use to start a batch, cancel it, and await the count with a timeout.

They only work together: without half one, un-marking would make the batch surface unusable from colourless Go-style code (`bench/.../Bench.gs`'s `produceBatched` is a plain `func`).

## What is deliberately NOT changed

`[Suspending]` on **G#-emitted** functions, and the `suspend func` modifier, both stay exactly as they were.

- The attribute is D4's cross-assembly record: it is how a `suspend func`, or one inference coloured, keeps its logical `R` at a G# call site in another assembly. `Adr0174SuspendCrossAssemblyTests.SuspendFunc_InALibrary_IsAwaitedImplicitlyByAConsumer` pins that, and dropping it would mean a function gaining a channel operation forces `await` at every cross-assembly call site — the source break D4 exists to contain.
- The modifier is D4's "declared where it must be" half. Inference stops at boundaries, and a boundary cannot be coloured by an `await` any more than by a channel operation, so `suspend func` remains the only way to pin the ABI there.

What was wrong was applying the attribute **by hand to a C#-authored method**: the API then has two contradictory shapes, `ValueTask<int>` to the C# that calls it (including its own tests) and `int32` to G#.

Nothing was lost on the D7 side because nothing was there: the ambient context reaches only a trailing parameter named `<>ctx` (`SupplyImportedContext`), and these methods take an ordinary `Context? context = null` the caller passes explicitly.

## Discrimination (ADR-0154)

Ten new cases in `test/Compiler.Tests/Issue3954AwaitableSemanticsTests.cs` — five executable, five rejections. Every executable one compiles, **IL-verifies and runs**, because the counts they print are what distinguish a real transfer from a completed placeholder and a wrong state machine IL-verifies clean but deadlocks; the runner reads both streams asynchronously and bounds the wait, so a deadlocking case fails rather than hanging the suite. Each rejection case asserts the absence of GS9998 alongside its expected id, since reaching the emitter is the failure they exist to prevent.

| run | result |
|---|---|
| whole `src/` reverted to the parent commit | **5 failed / 1 passed** |
| with the fix | **10 passed** |
| only the un-marking reverted | **2 failed** — the `.AsTask()` pair, with the issue's own GS0159 |
| only the binder/inference change reverted | **3 failed** — both plain-`func` awaits (GS0132) and the boundary case |

The two half-reverts are the argument for one PR rather than two: each half alone leaves the other's cases red.

The single green in the full revert is `channel-syntax-still-awaits-implicitly`, the control that draws this issue's line — a mutant that "fixes" #3954 by removing the implicit await everywhere fails it, because `<-ch` in a plain `func` needs no `await` and never will. `await-at-a-boundary-is-rejected` holds the new rule to D4's stated limits; the boundary kinds were each probed (`open` method, constructor, iterator, function literal all report GS0574 rather than reaching the emitter).

## Tests replaced, not deleted

`AsyncAwaitTests.Await_OutsideAsync_Diagnoses` pinned ADR-0023's rule and could not survive this decision. It becomes `Await_InAPlainFunc_ColoursTheCaller` (evaluating to 42) plus `Await_AtASuspensionBoundary_Diagnoses`, which pin ADR-0174 D4's. The batch-surface emit tests now say `await`, and gain `ABatchCall_IsAnOrdinaryValueTask` — the point of the issue, that the call has a task a caller can name.

## Also in this PR

- `bench/concurrency/gsharp/Bench.gs`: `await ch.SendBatch(slice)`. Harness smoke-verifies; `chunk64` measures 8.33 ns/op locally.
- ADR-0174 errata 37, `docs/diagnostics.md` (GS0574 added, GS0132 reworded), and release notes.

## Positions where `await` is rejected

Making `await` legal in a plain `func` needed the positions that have nowhere to suspend to be closed, and review found two. Both are now diagnostics rather than reaching the emitter, and each also closes a pre-existing hole:

- **`GS0575` — a `lock` body.** The monitor is thread-affine and reentrant, which is exactly why errata 10 has a channel operation there compile to the BLOCKING form; an explicit await has no blocking form to fall back to, and a continuation resuming on another thread would exit a lock it does not hold. Rejected in every function kind — the same shape was already accepted inside an `async func`, where C# spells the rule CS1996.
- **`GS0576` — nested in a `go` operand's arguments.** `BindGoStatement` strips the operand's own await (the goroutine consumes that task), and `SpillSequenceSpiller` does *not* hoist a nested one out of a `go`. Measured on `go consume(await fetch())`:

  | enclosing function | before this PR | now |
  |---|---|---|
  | plain `func` | GS0132 | GS0576 |
  | `async func` | **GS9998** (emitter) | GS0576 |

  So the diagnostic replaces an internal compiler error as well as covering the shape this PR would otherwise have let through.

**Merge note:** rebased onto `main` after #3960 landed; this PR's ADR entry is errata 38.

## Scope

The 7 `.AsTask()` signatures. The other eight in #3954 — cs2gs `return await Task.Yield()`, the Xunit `Record` collision, a tuple-literal `Assert.Equal`, and the GS0557/GS0558 rule collisions — are four unrelated subsystems and are left to their own work.


DavidObando:
Rebased onto `main` (#3960 landed, so the errata renumbered 37 → 38 and the test's cross-reference is now correct). All seven comments addressed; two of them found real defects, and one of those turned out to predate this PR.

**`await` in a `lock` body — correct, and worse than stated.** Verified: `lock o { v = await answer() }` compiled in a plain `func` coloured by another suspension point, *and* in an `async func`. The second is pre-existing — my change only opened the first — but the hazard is identical and ADR-0174 itself gives the reason: errata 10 says a `lock { ch <- v }` body compiles to the BLOCKING form precisely because Monitor is thread-affine and reentrant. An explicit await has no blocking form to fall back to. New **GS0575** rejects it in every function kind, which is also exactly C#'s CS1996, and closing the pre-existing half is the right call under this PR's own principle that awaitables behave as in C#.

**Nested `await` in a `go` operand — real, but the mechanism is the opposite of what the comment says, and the fix is different.** `SpillSequenceSpiller` does *not* hoist nested awaits out of `BoundGoStatement`. Measured on `go consume(await fetch())`:

| enclosing function | before this PR | on the first revision | now |
|---|---|---|---|
| plain `func` | GS0132 | **GS9998** (emitter) | GS0576 |
| `async func` | **GS9998** | GS9998 | GS0576 |

So counting the await as a suspension point would not have helped — colouring the caller still leaves an await inside the `go` that no function kind can lower. The honest fix is to reject it with new **GS0576**, naming the workaround (bind the value to a local, then `go` the call). That fixes the regression this PR introduced in row 1 *and* replaces a pre-existing internal compiler error in row 2. The collector keeps its `goDepth == 0` guard, now with a comment saying why: colouring a function on an await that is about to be rejected would be a signature change bought for nothing.

**GS0132's descriptor** — fixed in the descriptor, not only the docs. It now reads "'await' needs an enclosing function to make suspending, and there is none here (ADR-0174 D4)."

**GS0574's remedy** — you're right that `async init` does not exist (ADR-0065 defers async initializers) and accessors cannot be redeclared either, so "declare it `async func`" was impossible advice for several boundaries the message itself names. Reworded to lead with the remedy that always works: *move the awaited work into a `suspend func` or `async func` and call that from here*, with declaring this one that way offered only "where the form allows it."

**Test harness hang** — fixed. `RunDotnet` now reads both streams asynchronously and bounds the wait at 60s, killing the process tree and failing the case rather than hanging the run. You are right that this matters specifically here: the remarks say deadlock is a discriminator, so a draining read would have hung the whole suite instead of failing one case.

**Errata number** — resolved by the rebase; the entry is 38 and the test's `<remarks>` already said 38.

**Stale reference docs** — `website/docs/ref/spec.md` and `feature-matrix.md` both updated: `await` added to the list of suspension points that colour a plain `func`, GS0574/GS0575/GS0576 named, and a new paragraph stating the line this PR draws — the compiler inserts an await only where the syntax is a channel operation or the callee is itself suspending, never because a method returns a task.

**Tests:** the rejection cases are now a `[Theory]` of five — boundary, lock in a plain `func`, lock in an `async func`, nested `go` await in a plain `func`, nested `go` await in an `async func` — each asserting the expected id *and* the absence of GS9998, since reaching the emitter is the failure they exist to prevent. 10 green; 5 red / 1 green on the reverted `src/`; the two half-reverts still discriminate as before. Targeted runs: Compiler.Tests #3954 10/10, Core.Tests `Adr0174|Suspension|Async|Await|Lock` 936/936, Interpreter.Tests 35/35, bench harness smoke OK.

PR #3966, opened 2026-09-05
gsc: read `X.Member[i](args)` as indexer-then-invoke when the generic reading fails (#3880)

DavidObando:
## Root cause

ADR-0020's follow-set rule commits `Name[...]` to a **generic call site** whenever the bracket contents scan as type clauses and the next token is one of `(`, `{`, `.`. The `(` marker is arity-agnostic, so a single bracketed *simple name* is committed too — and indexer-then-invoke is never reconsidered afterwards.

```gs
class Samples {
    shared {
        let All Dictionary[Type, Func[string]] = Dictionary[Type, Func[string]]()
    }
}

Samples.All[t]()          // error GS0158: Cannot find member All
let f = Samples.All[t]    // binds fine
f()                       // binds fine
```

Issue #942 already made this call for the `.` follow-set — a single bracketed argument followed by `.` is an indexer, because an indexer can only ever hold one index. The `(` follow-set never got the equivalent treatment.

This is the `Coverage/PrinterExhaustivenessTests.gs:57` site (`GNodeSamples.All[type]()`) — 1 of the 5 gate fingerprints on the migrated `tools/cs2gs/Cs2Gs.Tests`.

## Why the parser cannot fix it

`Map[int](xs)` is a genuine generic instantiation with the identical token shape. The two readings are distinguishable only by what `All` / `Map` turns out to be, which is a binder fact. So the parser keeps its commitment and the **binder** falls back — and only after the generic reading has already failed.

## Fix

Two edits, one rule.

- **`ExpressionBinder.Calls.Arguments`** — inside `TryBindUserCallableMemberInvocation`, once the member turns out not to be callable, bind the bracketed identifier as an **index** against the member load and, when the result is a delegate of matching arity, invoke it. The index bind is speculative (`Diagnostics.TruncateTo` on failure), so a genuine "no such generic method" still reports the diagnostic the caller would have produced.
- **`ExpressionBinder.Calls.Invocation`** — the instance path resolves the type-argument list eagerly and bailed with `GS0113: Type 'key' doesn't exist` long before the member was looked at. It now falls through to the indexed reading, but **only** when the bracket is a bare identifier *and* the receiver really has a readable value member of that name.

Only a bare identifier is treated as ambiguous:

- anything a type clause can spell that an index cannot (`T?`, `[]T`, `List[T]`, a qualified name, a tuple, a function type) stays a type argument;
- anything an index can spell that a type clause cannot (`All[0]()`, `All["k"]()`) never reached the generic reading in the first place.

## Verification

`test/Compiler.Tests/Emit/Issue3880IndexedCallableMemberEmitTests.cs` — **executes** end to end (ILVerify + run), covering static receiver, instance receiver with an argument, an array-typed member, the un-invoked form, and `Generic.Make[int32]()` as the control that the generic reading must still win. A second test pins that `h.Pick[NotAType]()` and `Holder.Make[AlsoNotAType]()` keep reporting their own `GS0113` rather than having it swallowed by the recovery.

Full `Core.Tests`: 8904 passed.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Arguments.cs
+++ b/src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Arguments.cs
@@ -672,6 +672,23 @@ internal sealed partial class ExpressionBinder
 
         if (!MemberLookup.TryGetDelegateFunctionTypeFromSymbol(callableType, out var functionType))
         {
+            // Issue #3880: ADR-0020 commits `X.Member[i](args)` to a GENERIC
+            // call site — a single bracketed simple name scans as a type clause
+            // and `(` is in the follow-set — and indexer-then-invoke was never
+            // reconsidered afterwards. A dictionary of factories
+            // (`GNodeSamples.All[type]()`) therefore reported GS0158 on the
+            // member, while the identical `let f = GNodeSamples.All[type]`
+            // followed by `f()` bound fine: the `.` follow-set already got this
+            // treatment for the single-argument shape in issue #942, the `(`
+            // follow-set did not. The parser cannot decide this — `Map[int](xs)`
+            // is a real generic instantiation with the same token shape — so
+            // the binder decides it, and only where the generic reading has
+            // already failed: the member is not itself callable.
+            if (TryBindIndexedCallableMemberInvocation(ce, memberLoad, arguments, out result))
+            {
+                return true;
+            }
+
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


DavidObando:
### Where this sits in the #3880 wall

Measured from the newest **completed** gate run's `run.merged.json` (`33973595210`), `tools/cs2gs/Cs2Gs.Tests` was **5 fingerprints / 7 error sites** behind **3 root causes**.

A local migrate + validate with the other two fixes (#3967, #3964) applied leaves exactly one site standing:

```
```

which is this PR's. With all three merged the app's compile wall reaches zero.

DavidObando:
All three findings evaluated and addressed in `bf550410c`. Two were real defects, the third was real and trivial.

---

### 1. Reduced argument binding — **REAL, fixed**

Confirmed before changing anything:

```gs
delegate Joiner(sep string, parts ...string) string;
b.Named[k](",", "a", "b", "c")
```

```
error GS0159: Cannot find function Named.
```

My exact-arity check (`ParameterTypes.Length != arguments.Length`) rejected four arguments against arity 2. Everything the canonical binders own — variadic packing, optional/named arguments, ref/out/in kinds — was missing.

Rather than call into the indirect-call binder (which re-binds arguments from syntax and would have double-bound them, since the callers here already have `boundArguments`), I removed the duplication entirely: **the indexed value now replaces the callable and falls through to the tail the non-indexed delegate member already used.** That tail is `TryBindNamedDelegateArguments` for named delegates and the variadic-aware conversion path otherwise. `TryBindIndexedCallableMemberInvocation` is gone; what remains is `TryBindAmbiguousBracketAsIndex`, which only produces the indexed value. Net: one argument path, and ~40 fewer lines than before.

Now prints `a,b,c`, and `b.Named[k]("-")` (zero variadic elements) works too.

---

### 2. Imported CLR receivers — **REAL, fixed**, and you were right to weight it highest

```gs
r.Callbacks[k]()   // Registry from a referenced assembly
```

```
error GS0159: Cannot find function Callbacks.
error GS0113: Type 'k' doesn't exist.        <- k is a local variable
```

Character-identical source compiles when `Registry` is source-declared. Two separate gates excluded imported receivers: `HasCallableIndexableValueMember` recognised only source members, and the bind itself only ran for `StructSymbol`/`InterfaceSymbol`. Both now handle CLR fields/properties, and `TryBindClrDelegateMemberInvocation` grew the same fall-through, reusing its existing `Invoke` dispatch.

**Scope boundary, measured rather than assumed.** The indexed form now works wherever the *non-indexed* delegate-member form works:

| receiver | plain `X.Cb()` | indexed `X.Map[k]()` |
|---|---|---|
| instance, source | ✅ | ✅ |
| instance, imported | ✅ | ✅ **fixed here** |
| static, source | ✅ | ✅ |
| static, imported | ❌ | ❌ |

The last row is **not** a hole in this recovery: `Reg.Cb()` on an imported static fails too, because gsc has no static analogue of `TryBindClrDelegateMemberInvocation`. That is pre-existing and independent of the parser ambiguity, so it is filed as **#3970** rather than folded in here. The indexed form will start working there for free when that lands.

---

### 3. Leaked temp directory — **REAL, fixed**

`CompileExpectingErrors` now uses the same `try/finally` cleanup as `CompileAndRun`.

---

### Verification

- `IndexedNamedDelegate_KeepsVariadicAndNamedArgumentSupport` — executes.
- `IndexedCallableMember_WorksThroughAnImportedReceiver` — compiles a library, compiles a consumer against it, **ILVerifies the consumer against the library**, and runs it. (Note for anyone copying it: ilverify resolves references by simple assembly name, so the library file has to be named for its package.)
- Both existing tests still pass, including the negative one pinning that a genuine bad type argument keeps reporting its own GS0113.
- `Compiler.Tests` filtered `Delegate|Imported|Clr|Issue3880|Accessor|Index` — **989 passed**. That filter deliberately covers the imported-call path, since `TryBindClrDelegateMemberInvocation` is on every imported member call and now builds the member load one step earlier. (`ApplyMemberNarrowing` is a pure read of `binderCtx.NarrowedVariables`, so moving it before the delegate-type check has no side effects.)

DavidObando:
One more consequence of finding 1's fix, worth recording because it is user-visible.

Routing the indexed callable through the canonical argument binders means a wrong-arity call now gets the *arity* diagnostic instead of a lookup failure. Same source, before and after:

```gs
h.Table[k]()      // Table is Dictionary[string, Func[int32, string]]
```

| | diagnostic |
|---|---|
| before | `GS0159: Cannot find function Table.` |
| after | `GS0144: Function 'Table' requires 1 arguments but was given 0.` |

The old message was a masking shape of exactly the kind #3880 is about — it said the member could not be found when the member had been found and only the argument count was wrong. Under the previous design the arity check was mine and its failure meant "give up on the indexed reading", so the call fell through to the generic-reading diagnostic. Now the indexed reading commits and the canonical binder reports what is actually wrong.

The negative test still passes unchanged: a genuine unresolvable type argument (`h.Pick[NotAType]()`, `Holder.Make[AlsoNotAType]()`) still reports its own GS0113, because the recovery only fires when the receiver really has a value member of that name.

PR #3967, opened 2026-09-05
cs2gs tests: build the ADR-0169 parity control from the analyzer's own source (#3880)

DavidObando:
## Root cause — a test premise that stops holding once the repo migrates itself

The two ADR-0169 parity tests named their Roslyn control through the `GSharp.InternalAnalyzers` **project reference**:

```csharp
ImmutableArray.Create<DiagnosticAnalyzer>(new StructFieldDefsReadAnalyzer())
ImmutableArray.Create<DiagnosticAnalyzer>(new RewriterClonePreservationAnalyzer())
```

cs2gs retargets `src/Analyzers/InternalAnalyzers` onto the G# analyzer API (ADR-0169), so in the migrated tree those types are `GSharpDiagnosticAnalyzer`, not Roslyn `DiagnosticAnalyzer`. There is no applicable overload, and the whole fluent chain cascades:

```
GS0159 Cannot find function WithAnalyzers
GS0159 Cannot find function Create
GS0159 Cannot find function GetAnalyzerDiagnosticsAsync
```

None of which mentions the real problem — a masking shape of the #3823/#3842 family. **4 sites, 3 of the 5 `tools/cs2gs/Cs2Gs.Tests` gate fingerprints** on the newest completed nightly (`33973595210`, `run.merged.json`).

This is a **test defect**, not a translation defect and not a compiler defect: `Cs2Gs.Tests` cannot both migrate itself and keep a build-output reference to a project that migration deliberately retargets.

## Fix

New `Adr0169TranslatedAnalyzerHarness.CompileRoslynAnalyzer(analyzerFileName, analyzerTypeName)` compiles the analyzer's **own C# source** with Roslyn (via the loader the harness already uses), emits it to memory, loads it and instantiates the analyzer as a `DiagnosticAnalyzer`.

Nothing about what the test measures changes, and the two halves become symmetric: the G# half already reads the same file off disk, translates it in analyzer mode and compiles it with `gsc`. Both sides now start from the same bytes instead of one side starting from a build output — arguably a better control than the original.

No G# language surface was added to accommodate this; the new code is ordinary reflection that already translates.

## Verification

Both parity suites green in C#: 8 passed (`Adr0169AnalyzerParityTests`, `Adr0169Gsa0005ParityTests`).

### DIFF
--- a/tools/cs2gs/Cs2Gs.Tests/Adr0169AnalyzerParityTests.cs
+++ b/tools/cs2gs/Cs2Gs.Tests/Adr0169AnalyzerParityTests.cs
@@ -11,7 +11,6 @@ using Cs2Gs.CodeModel.Printing;
 using Cs2Gs.Translator;
 using Cs2Gs.Translator.Analyzers;
 using Cs2Gs.Translator.Loading;
-using GSharp.InternalAnalyzers;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Xunit;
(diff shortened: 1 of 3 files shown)


DavidObando:
### End-to-end verification

Ran `run-cs2gs-selfmig-migrate.sh` + `run-cs2gs-selfmig-validate.sh` locally for `tools/cs2gs/Cs2Gs.Tests` with this PR stacked with #3964, against a repacked `Gsharp.NET.Sdk` nupkg.

The migrated app's `sdk.build.log` went from **7 error sites / 5 gate fingerprints** down to a single site:

```
```

All four `WithAnalyzers` / `Create` sites this PR targets — 3 of the 5 fingerprints — are gone. The remaining one is `GNodeSamples.All[type]()`, a separate root cause fixed in #3966.

The new harness code also **survives translation**: it appears in the migrated tree and compiles, so the #3831/#3896/#3905/#3915 hazard (a `tools/cs2gs/**` change that is green in CI and still takes the nightly down purely by existing) is cleared for this change. No G# language surface was added.

PR #3969, opened 2026-09-05
gsc: close the variadic ELEMENT type before ranking a generic candidate (#3880)

DavidObando:
Follow-up to **#3964**, which closed the variadic **carrier**. The expanded-form **element** type, computed one line above it, was left open — and misranks on its own.

Raised by the review thread on #3964 (["Generic variadic candidates are still ranked against the open carrier type"](https://github.com/DavidObando/gsharp/pull/3964#discussion_r0)). That comment's literal subject — the carrier — was already fixed by `2b80a9794` before it was posted; taking it seriously rather than closing it as a duplicate surfaced this second, still-live instance of the same rule.

## Defect

```gs
class Pick {
    shared {
        func Elem[T](values ...T) string { return "generic-elem" }
        func Elem(values ...object) string { return "object-elem" }
    }
}
Pick.Elem("a", "b")
```

csc oracle (net10.0), same shapes:

```csharp
static string Elem<T>(params T[] values) => "generic-elem";
static string Elem(params object[] values) => "object-elem";
Elem("a", "b");   // generic-elem
```

| build | result |
|---|---|
| merged `main` (`8b41b91ed`, includes #3964) | `object-elem` ❌ |
| this PR | `generic-elem` ✅ |

Verified against a `gsc` built from merged `main`, not from a stale worktree.

## Why it goes wrong

After inference the candidate still declares its element type as the open `T`. `ClassifyUserArgumentConversionKind(string, T)` and `ClassifyUserArgumentConversionKind(string, object)` land on the **same** conversion kind, so the two candidates tie on every argument, survive pairwise domination together, tie on defaulted-parameter count and on variadic-ness, and are then separated by the non-generic-over-generic tie-break — which picks the `object` overload. C# never gets there: with `T = string` the element conversion is an identity, which is strictly better, and betterness is decided before that tie-break is consulted.

## Correcting my own reasoning from #3964

When I scoped this out on that PR I claimed the expanded path was harmless because "its `ParamTypes` stays null, so the reference *better conversion target* tie-break never runs." That is true and irrelevant. `CompareUserConversions` compares conversion **kinds** first and returns before `ParamTypes` is read, so an expanded tail slot can misrank without that tie-break participating at all. The scope decision was wrong on the merits, not merely conservative.

## Fix

Three lines: substitute `elementType` through the candidate's inferred `candSubstitution`, exactly as the non-tail slots and (since #3964) the carrier already do. Not a carrier-semantics change — ADR-0173's `...X[T]` ≡ `params X<T>` is untouched; this is substitution order.

## Verification

`GenericExpandedForm_ClosesTheElementTypeBeforeRanking` **executes** the emitted program and asserts the printed overload. This defect binds clean and ILVerifies clean — a binding assertion passes on it, so only running the program shows the wrong method was called.

`Compiler.Tests` `Issue3880VariadicPassThroughOverload`: 4 passed.

### DIFF
--- a/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.Candidates.cs
+++ b/src/Core/CodeAnalysis/Binding/OverloadResolution/OverloadResolver.Candidates.cs
@@ -553,9 +553,31 @@ internal sealed partial class OverloadResolver
                     }
 
                     isTailSlot[i] = true;
-                    kinds[i] = elementType == null
-                        ? ClrOverloadResolution.ImplicitConversionKind.Identity
-                        : ClassifyUserArgumentConversionKind(tailArgType, elementType);
+
+                    // Record the (closed) element type as this slot's ranking
+                    // target. It was left null, which made CompareUserConversions
+                    // unreachable-by-assumption for expanded slots: its
+                    // NumericWidening arm dereferences BOTH parameter types, and
+                    // its Reference / UserDefinedImplicit arms silently tie when
+                    // either is null. Two variadic candidates whose tails
+                    // classify identically therefore either crashed (numeric) or
+                    // lost their "better conversion target" tie-break (reference).
+                    // `G(anchor int64, values ...int64)` vs
+                    // `G(anchor int64, values ...float64)` called with an int32
+                    // tail crashed with GS9998 on this path before this line
+                    // existed — closing the element type above widens the set of
+                    // pairs that reach it, so it has to be populated here.
+                    if (elementType == null)
+                    {
+                        // No element type to rank against: stay Identity and
+                        // leave the slot's target unset. Identity never reaches
+                        // an arm of CompareUserConversions that reads it.
+                        kinds[i] = ClrOverloadResolution.ImplicitConversionKind.Identity;
+                        continue;
+                    }
+
+                    paramTypes[i] = elementType;
+                    kinds[i] = ClassifyUserArgumentConversionKind(tailArgType, elementType);
                     continue;
                 }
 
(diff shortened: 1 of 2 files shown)


DavidObando:
## Reproduced. The finding is correct, and it found more than it claimed.

Fixed in `0fc4b1cc6`.

### The crash reproduces on the reviewer's exact shape

```gs
func F[T](anchor T, values ...T) string { return "generic" }
func F(anchor int64, values ...float64) string { return "float" }

let a int64 = 1
let b int32 = 2
C.F(a, b)
```

```
error GS9998: InvalidOperationException: Invariant violated: 'paramA?.ClrType' was null.
Expected non-null because a numeric conversion target has a CLR type.
```

The mechanism is exactly as described, and it is the other half of the fact I used to overturn my own scope-out on #3964. `CompareUserConversions` returns early when kinds **differ**; when they are **equal** it proceeds — and its `NumericWidening` arm dereferences both parameter types, which expanded tail slots never recorded. Closing the element type is what makes two tails land on the same kind.

### But the hole predates this PR

The same crash reaches plain `main` with **no generic candidate at all**:

```gs
func G(anchor int64, values ...int64) string { return "i64" }
func G(anchor int64, values ...float64) string { return "f64" }
C.G(a, b)     // int64, int32
```

`candSubstitution` is null for non-generic candidates, so this branch is a **no-op** for that source — and it crashes on `main` today. So closing the element type did not create the defect; it widened what reaches it.

### Full before/after, csc as oracle

| shape | pristine `main` | this PR before | now | csc |
|---|---|---|---|---|
| `F` generic numeric tail | `float` — **wrong overload** | 💥 GS9998 | `generic` ✅ | `generic` |
| `G` non-generic numeric tails | 💥 **GS9998 (pre-existing)** | 💥 GS9998 | `i64` ✅ | `i64` |
| `H(...Animal)` vs `H(...object)`, Dogs | **GS0266 spurious ambiguity** | GS0266 | `animal` ✅ | `animal` |

`F` is the genuine regression this PR introduced — and note it was already picking the **wrong overload** before it started crashing, so the element substitution was right; it just landed on an unguarded slot. `H` is the reference half the finding asked me to cover: it was silently tying and reporting a spurious ambiguity, because `CompareReferenceTargets` is guarded on both parameter types being non-null.

### Fix

1. **The one suggested.** The expanded slot now records its closed element type as its ranking target, which feeds both the numeric and the reference better-conversion-target rules. Three defects fall out of that one line: the pre-existing crash, the regression, and the spurious ambiguity.
2. **Defensive.** The `NumericWidening` arm asserted three `Invariant.Required` facts the function cannot actually guarantee — it is handed whatever the ranking loop recorded. It now ties instead of throwing. Recording the element type is the real fix; this exists so a **ranking heuristic can never take a whole compile down again**. A tie is the honest answer when there is nothing to compare, and a crash on source that C# accepts is a strictly worse failure than a suboptimal tie.

### Verification

Three regression tests, all **executing** and asserting the printed overload:

- `ExpandedNumericTails_RankByBetterConversionTargetInsteadOfCrashing` — covers both `F` (fails on this branch pre-fix) and `G` (fails on `main`).
- `ExpandedReferenceTails_RankByBetterConversionTarget` — the reference tie-break.

Suites, re-run wide because populating that slot changes ranking behaviour generally:

- `Core.Tests` `Overload|Variadic|Params|Generic|Conversion` — **1759 passed**
- `Compiler.Tests` `Overload|Variadic|Params|Generic|Conversion|Numeric` — **1118 passed**
- `Issue3880VariadicPassThroughOverload` — 6 passed

Issue #3837, opened 2026-09-03
cs2gs self-migration: test/Interpreter.Tests compile wall — 38 fingerprints, 5 root causes, 32 of them one defect


DavidObando:
## Summary

`test/Interpreter.Tests` is the largest remaining self-migration wall (#3501). It reached the **compile** stage for the first time after #3807 (translate crash), #3817 (linked-source divergence) and #3823 landed. Gate run 33663855851 (`7a6fa9f35`) reports **38 fingerprints**.

Re-measured from that run's shard-3 artifact (`sdk.build.log`, deduplicated — MSBuild repeats every diagnostic):

- **38 fingerprints = 65 unique diagnostic lines** (a fingerprint dedups per file+message).
- **Every one is Interpreter.Tests' own error.** All 65 file paths are `migrated/test/Interpreter.Tests/*.gs`; `src/Repl`, `src/Compiler`, `src/Core` and `test/Shared` are all green in the same run. The earlier claim that these were "a dependency's errors, led by GS0157 ×21" is **wrong on both counts** — the leading family is 21 GS0157 but they are this app's, and the count is not the estimate.
- No `GS0536`.

## Root causes: 5, not 38 — and one is 32 of them

| # | Root cause | Fingerprints | Unique lines | Layer | Status | Tractability |
|---|---|---|---|---|---|---|
| RC1 | T3 flattens `src/Repl`'s entry class, erasing `GSharp.Repl.Program` from the migrated assembly | **32** | 59 | cs2gs translator + pipeline | **known cause, incomplete detector** (#3645) | fixed here |
| RC2 | `typeof(Issue3149Mapper[_])` — unbound-generic `typeof` of a source-declared delegate | 2 | 2 | gsc binder | new | small, needs its own issue |
| RC3 | GS0264 duplicate overload after delegate-type erasure (`Predicate<T>` and `Func<T,bool>` both print `(T) -> bool`) | 2 | 2 | G# type system / ADR | new | design question |
| RC4 | GS0252 expression-bodied `ref`-returning `func` (and `ref` dropped from `prop`/indexer) | 1 | 1 | cs2gs translator | new | small |
| RC5 | GS0125 unqualified member read inside a default interface method body | 1 | 1 | gsc binder (or cs2gs) | new | small |

`38` was never the estimate: **RC1 alone is 84% of the wall.**

### RC1 — the whole GS0157/GS0154 wall is one defect

[...]

Issue #1987, opened 2026-07-04
cs2gs: ref-return corner cases (ref-returning indexer element, return ref F(x) fixture)


DavidObando:
Follow-ups from PR #1986 (#1900) Opus review. Non-blocking:

1. `ElementAccessExpressionSyntax` on a type with a ref-returning indexer (e.g. `ref List<T>.this[]`) emits call-under-index into gsc → gsc compile error (loud, not silent miscompile), but not a clean gap. Consider detecting ref-returning indexers and gapping precisely.
2. Add a fixture/test for `return ref F(x)` where the ref-return operand is a call — path exists (loud gap via TranslateRefExpression fallback) but is untested.

Discovered during autopilot review of #1900.

Issue #490, opened 2026-06-05
ref returns from G# functions (follow-up to ADR-0060)


DavidObando:
## Background

ADR-0060 ships `ref`/`out`/`in` parameter modifiers at call sites and at method definitions for G#-authored functions. It explicitly **forecloses** `ref` returns from the V1 surface:

> `ref` returns from G# functions (`func foo() ref int32 { return ref x }`). Not addressed in this ADR; remains a follow-up gated on ref-safe-to-escape per ADR-0058.

This issue tracks the follow-up.

## Goal

Allow a G# function (free function or member) to return a managed pointer:

```gs
func max(ref a int32, ref b int32) ref int32 {
    return ref (a > b ? a : b)
}
```

Callers consume the result either as a re-bindable alias (`let ref m = max(ref x, ref y); m = 0` to mutate through the returned pointer) or pass it onward to another `ref` parameter.

## Scope

- New return-type syntax `ref T` on `func`-decl `ReturnTypeClauseSyntax` (parser, syntax node)
- `FunctionSymbol.ReturnRefKind` and matching emit (return type encoded as `T&`)
- `BoundReturnStatement` extended to carry a `RefKind` on its expression payload, mirroring §5's `BoundIndirectAssignmentExpression` shape
- Bind-time validation that the returned expression is an lvalue whose ref-safe-to-escape scope outlives the function's call frame (per ADR-0058)
- New diagnostics for: returning a non-lvalue as `ref`, returning a local whose scope dies at function exit, conditional ref-returns where the two branches have incompatible escape scopes
- Override / interface implementation matching must include return ref-kind (extension of GS0240)
- Interpreter parity
- C# consumer interop tests (verify Roslyn sees the return as `ref int32`)

## Dependencies

- **ADR-0058** (ref-safe-to-escape) must have a mature data-flow tracker for this to be safe. Pre-conditions: scope-set propagation through binder, conservative `scoped` enforcement on `ref` parameters.
- Composes with the follow-up issue for `let ref x = expr` aliasing locals (which itself depends on this).

## References

- ADR-0060 §Consequences (Foreclosed), §Follow-ups
- ADR-0058 (ref-safe-to-escape)
- ADR-0039 (managed by-ref pointers — `&x`, `*T`)
- PR #489 (ADR-0060 implementation)

PR #4052, opened 2026-09-07
gsc + cs2gs: a ref-returning property and indexer (#3879)

### LINKED ISSUE #3879: G#: no ref-returning property or indexer, so cs2gs cannot translate C#'s (gap surfaced by #3839)
Split out of **#3839** / PR #3878, which established that this is a language gap rather than a translator bug.

## The gap

C# permits a ref-returning property and indexer:

```csharp
public ref int Value => ref this.slot;
public ref int this[int i] => ref this.items[i];
```

**G# has no such form.** `prop P ref int32` does not parse, and ADR-0060 gives the by-ref return to `func` only. So there is no spelling for cs2gs to translate into.

## Current behaviour after #3878

PR #3878 fixed the **method** half of #3839 (`WrapExpressionBody` built the return without `isRef`, and `TryFoldArrowBody` then folded it into the arrow form — both corrected, so a ref-returning *method* keeps `{ return ref values[index] }`).

For properties and indexers it now **gaps loudly at the declaration site** for both `ref` and `ref readonly`, matching the verdict cs2gs already reached at the *use* site in #1987.

That is the correct interim behaviour — the alternative was the status quo, where the member migrated into something that **silently returned copies**. A visible gap beats a silent wrong answer.

**Concrete consequence, so it is not a surprise:** `test/Interpreter.Tests/Issue3004ManagedByRefAutoDereferenceFixture.cs` now fails migration loudly instead of migrating into a fixture that quietly returns copies.

## The work

Decide whether G# should support a ref-returning property/indexer, and if so add it:

- syntax (`prop P ref int32 { get }`? or restrict to computed properties only, since an auto-property cannot meaningfully return a ref);
- binder and emit (`ReturnType.IsByRef` on the property's getter);
- the ADR-0060 amendment;
- cs2gs translation, which becomes mechanical once the spelling exists.

Worth considering **restricting it to computed properties**: a ref-returning *auto*-property has no coherent meaning, and #3878's GS0219 sweep showed the auto-property backing-field path is exactly where by-ref hazards concentrate.

If the decision is **not** to support it, that is legitimate — but then the gap should be documented in the ADR and the migration consequence accepted explicitly, rather than left as an unexplained failure for whoever next migrates that fixture.

## Note on testing this class

A ref-return aliasing test cannot be written in G# today: the ref-alias binder rejects a call result as an lvalue (the documented #1900 limit). PR #3878's proof compiles a **C# driver** against the emitted G# library, takes `ref int slot = ref holder.GetValue(1)`, writes through it, and reads back via a separate accessor — a copy-returning member reads the old value. Reuse that shape.

Refs #3839, #3878, #1987, #1900, #3004, ADR-0060.

DavidObando:
Fixes #3879.

## The decision

**Support it**, restricted to computed, read-only properties and indexers — the
issue's own recommended direction, and the one the C# surface makes total (C#
forbids a setter on a ref property, CS8147, so every C# ref property is already
this shape).

```gs
class Holder {
    private var values []int32 = []int32{40, 41, 42}

    prop First ref int32 { get { return ref values[0] } }   // block form
    prop Second ref int32 -> values[1]                      // arrow form (ADR-0131)
    prop this[i int32] ref int32 -> values[i]               // indexer form (ADR-0118)
}
```

ADR-0060 gains a **§14 amendment**; ADR-0051's grammar production gains the
optional `ref`, with the rule itself owned by ADR-0060.

### Why "computed only" is a rule, not a shortcut

| Rejected shape | Diagnostic | Reason |
| --- | --- | --- |
| `prop Slot ref int32` (auto-property) | GS0578 | A by-ref return has to *name* the storage it aliases. An auto-property's getter copies out of a compiler-synthesized backing field, so `ref` would alias the field behind the caller's back — and #3878's own GS0219 sweep found that backing-field path is exactly where by-ref hazards concentrate. |
| `prop Slot ref int32 { get }` | GS0578 | Same rule: an abstract slot or a read-only auto-property names no storage. |
| `interface I { prop Slot ref int32 { get } }` | GS0578 | An interface member is a slot, not storage — and G# has no ref-kind matching for property slots (GS0255 covers methods only), so an implementor could satisfy a `ref` requirement with a copy-returning property and nothing would catch it. Rejected rather than accepted-and-unchecked. |
| a `set` / `init` accessor alongside `ref` | GS0579 | The returned reference *is* the write path. C# spells the same rule CS8147. |
| `prop P ref *int32` | GS0250 (reused) | Same redundancy rule a ref-returning `func` gets. |

Everything else is the **existing** GS0248–GS0255 surface, unchanged and
un-duplicated: the accessor body is bound with the getter's `FunctionSymbol` as
its enclosing function, so a plain `return` inside a ref getter is GS0252, a
non-lvalue operand is GS0253, and returning a reference to a getter-local is
GS0254. All verified; no new code for any of them.

### The arrow form is ref-aware; the `func` arrow still is not

`prop P ref T -> e` desugars to `{ get { return ref e } }`. The asymmetry with
#3878's method rule ("a ref return is NOT foldable") is deliberate: a `func`
arrow body *has* a `return` in its surface spelling for `ref` to attach to, so
folding one drops information; a property's arrow has none, so the declaration's
own `ref` is the only place the modifier can live. There is exactly one spelling
— `-> ref e` does not parse.

## What changed

**gsc**

* **Parser** — an optional `ref` between the member name (or the indexer's `]`)
  and the type clause, recognized with the identical lookahead `func` uses, so a
  property literally *named* `ref` (`prop ref int32` — tested) is unaffected. The
  flag threads into the whole-member arrow and into `get -> e`.
* **Symbol** — `PropertySymbol.ReturnRefKind` (`None`/`Ref`), mirroring
  `FunctionSymbol.ReturnRefKind`. `PropertySymbol.Type` stays the **pointee**
  `T`, exactly as ADR-0060 §3 decided for parameters and #490 for function
  returns; carried across generic construction.
* **Binder** — one `ValidatePropertyReturnRefKind` shared by the instance and
  `shared` paths (interfaces reject outright); the getter's accessor symbol
  carries the same ref-kind, which is what makes `return ref` legal in its body
  and makes `FunctionEmitter` encode `T&`.
* **Emit** — a single `EncodePropertyType` funnel for every property-signature
  return slot (instance, static, interface, indexer, and the metadata-only
  reference-assembly fallback), so the implementation and reference assemblies
  cannot disagree. Both the PropertyDef row **and** `get_P` encode `T&`.
  Measured: with the getter by-ref and the PropertyDef row left by-value, Roslyn
  does not merely refuse the alias — it refuses the member outright with
  **CS1546** ("… is not supported by the language; try directly calling accessor
  method `Holder.get_Property()`").

**cs2gs**

* `PropertyDeclaration.IsRefReturn` + printer; `TranslateProperty` /
  `TranslateIndexer` set it instead of gapping; the property/indexer arrow fold
  now accepts a ref return (`allowRefReturn`), the `func` fold still refuses it.
* #1987's use-site gap on a plain-`ref` indexer read is lifted — the declaration
  translates and gsc loads through the pointer at the read.
* **`ref readonly` still gaps**, at the declaration and at the use site. G# has
  no read-only by-ref return; rendering one as a plain `ref` would hand the
  caller a *writable* alias to storage the author declared read-only. That is a
  different silent behaviour change, not a smaller one.
* An **extension property** (which lowers to a receiver-clause `func`, never to a
  `prop`) now gaps too — it never carried `isRefReturn`, so it would otherwise
  have silently started emitting a copy-returning function once the property gap
  was lifted.

## The side fix this forced, stated plainly

#490 deferred "consume a ref return at a G# call site". The deferral was not
free: the emitted call site left the raw `T&` on the stack where the bound tree
said `T`. On `main`, today, before any of my changes:

```
func pick(ref a int32) ref int32 { return ref a }
var n int32 = 7
Console.WriteLine(pick(ref n))     // prints 180863064 — the address
```

and through an instance member it is an outright `InvalidProgramException`. A
ref-returning **property** would have inherited that, and a property read is far
more common than a method-call read — so shipping the feature on top of it would
have meant adding a new silent-wrong-answer surface, which is the outcome #3839
exists to prevent. A call to any G# member whose `ReturnRefKind` is `Ref` now
loads through the returned pointer at the use site (ADR-0056 §1's rule, applied
to source members as it already was to imported ones). Three lines plus a guarded
helper; the `ByRefTypeSymbol` guard is load-bearing, because the imported path
and `BaseCallForwarderRewriter` already dereference and a second load would
corrupt the stack.

This means **`func` ref returns are also fixed**. That is a behaviour change
beyond the issue's scope — flagging it explicitly for review rather than burying
it.

**Writing** through the reference from G# stays out of scope and fails loudly:
`holder.RefProp = 5` reports GS0127 (the property has no setter). Documented in
ADR-0060 §14 as a coherent follow-up — it is a write-through property
assignment, a second feature.

## Evidence: executing, not printed

A printed-shape assertion and a metadata assertion both pass just as happily on a
member that lost its `ref`. The load-bearing test in each project is behavioural,
reusing PR #3878's shape: gsc emits the library, Roslyn compiles a **C# driver**
against it, the driver takes `ref int slot = ref holder.Property` (and
`ref holder[2]`), writes through it, and reads the storage back through a
separate by-value accessor. A copy-returning member reads 40/41/42.

* **`test/Compiler.Tests/Emit/Issue3879RefPropertyEmitTests.cs`** — 9 tests. The
  driver proof; PropertyDef **and** getter `IsByRef` on all three forms, with an
  ordinary property and an ordinary method on the same type as anti-vacuity; the
  G#-side read printing 40/41/42 (the deref guard); GS0578 ×3, GS0579, GS0252;
  and `prop ref int32` still parsing as a property *named* `ref`.
* **`tools/cs2gs/Cs2Gs.Tests/Issue3879RefReturningPropertyTranslationTests.cs`**
  — 7 tests. Translate → gsc → C# driver end to end; the `ref readonly` gaps;
  #1987's read no longer gapping; ordinary members staying by value; and
  `Issue3004ManagedByRefAutoDereferenceFixture.cs` reproduced **verbatim**,
  asserted gap-free.
* `Issue3839RefReturningMemberTranslationTests` keeps its method half and its
  `ref readonly` gap; the two now-obsolete plain-`ref` gap assertions are retired
  with the class doc pointing at #3879.
* `Issue1900RefLocalsTranslationTests.ElementAccess_OnRefReturningIndexer_StaysLoudGap`
  becomes `…_NoLongerGaps`, and gains a `ref readonly` sibling so the change is a
  narrowing, not a deletion.

**ILVerify**: the class-based shapes verify clean. A ref-returning member on a
**struct** trips `ReturnPtrToStack` — but it does so identically for the
pre-existing `func` form on `main` (measured), it is the ADR-0058 struct-`this`
ref-safety hole rather than anything this PR introduces, and `ReturnPtrToStack`
is already in `IlVerifyRunner.KnownIlVerifyFalsePositives`.

## Failing-on-main / anti-vacuity

Measured by reverting production hunks and re-running.

**Whole feature reverted** (`src/Core` + `tools/cs2gs/Cs2Gs.{Translator,CodeModel}`,
tests kept):

```
test/Compiler.Tests    Issue3879        Failed: 8, Passed: 1
tools/cs2gs            Issue3879        Failed: 6, Passed: 1
tools/cs2gs            Issue1900 (2 changed cases)  Failed: 2
```

The two survivors are the deliberate guard rails and are load-bearing:
`PropertyNamedRef_IsStillAnOrdinaryProperty` and
`OrdinaryPropertyAndIndexer_StayByValue` must pass on **both** trees, or "consume
`ref` unconditionally" / "always print `ref`" would satisfy everything else.

**Targeted reverts**, to prove each half is separately guarded:

| reverted hunk | result |
| --- | --- |
| the use-site dereference only | exactly 1 failure — `RefPropertyAndIndexer_ReadFromGSharp_LoadThroughThePointer`. Nothing else covers it. |
| the PropertyDef by-ref encoding only (getter still `T&`) | 2 failures; the C# driver stops compiling with **CS1546** on all three members. This is where the CS1546 claim above comes from. |

**One vacuity bug found and fixed in my own test, worth calling out**: the
artifact directory is named after the test, gsc echoes the source path in every
diagnostic, and the test names contain the diagnostic ids — so
`Assert.Contains("GS0578", output)` was satisfied by the *path* and passed on a
fully reverted tree. The assertions now match `"error GS0578"`, which a path
cannot contain, and the reverted-tree run above is what caught it.

## Migration impact

`test/Interpreter.Tests` was failing the self-migration gate (#3501) at the
**translate** stage with exactly two CS2GS-GAP fingerprints, both in
`Issue3004ManagedByRefAutoDereferenceFixture.cs` (line 15's `ref` property, line
19's `ref` indexer). Both now translate — the fixture is asserted gap-free
verbatim in the cs2gs suite.

**Do not expect that app to go green.** Issue #3837 documents a 38-fingerprint
compile wall sitting behind the translate gap, which this change un-masks. That
is expected and is not a regression from this PR.

`tools/cs2gs/selfmig-baseline.json` is untouched. If the gate's numbers move, let
the ratchet be banked separately (#4047 is open against that file).

## Suites

Run on this commit's tree.

| suite | result |
| --- | --- |
| `test/Core.Tests` | 9019 passed / 2 failed — both `Issue3755PinningShapeTargetFrameworkTests`, `prerequisite missing: no NETStandard.Library.Ref targeting pack`. Environment, not the change. |
| `test/Compiler.Tests` | 5226 passed / 3 failed — all three `Issue3764CrossContextValueTypeCallShapeTests`, same missing targeting pack. |
| `tools/cs2gs/Cs2Gs.Tests` | 2845 passed / 1 failed — `Adr0151…Issue2819_SameVersionPackedToolAndSdk_CompileTranslatedPatternVariable`, which needs `out/bin/Release/nupkgs` (a Release pack) that this Debug-only run never produced. |

None of the four failures touch a file this PR changes; all four reproduce from a
missing local prerequisite rather than from source.

`gsfmt` round-trips the new form with the `ref` intact (checked by hand); its two
spacing quirks around `[]` are pre-existing and reproduce without `ref`.


DavidObando:
All five review findings are fixed in `06f29dfe`. I confirmed each one against the code before changing anything — none was a false positive.

## Per finding

**1 (high) — the dereference reached only two source-call branches.** Correct, and it was worse than described. `base.RefMethod()` / `base.RefProperty` (`BoundBaseClassCallExpression`), `base` through a default-interface implementation (`BoundBaseInterfaceCallExpression`), and a static-interface call (`BoundConstrainedStaticCallExpression`) all returned straight after the call, leaving `T&` where the bound node says `T`. All three now route through `EmitRefReturnDereferenceIfNeeded`. The base-class form carries *either* a method or a property accessor, so it takes the ref-kind from whichever is present and skips the setter arm, which pushes nothing to dereference.

Separately, and worth recording: **the free-function branch was already fixed but was not guarded.** Reverting its hunk passed all nine original tests. Both it and the base forms now have behavioural tests.

**2 (medium) — ref-kind absent from override matching.** Correct. A source base property was found by name alone, and `FindProperty` received only the pointee type, so `override prop P ref T` bound happily to a by-value base slot, and a genuine override of an external `T&` property could never match because `T&` never equals `T`. Both directions now report **GS0185**, on both lookup paths — an explicit comparison on the source path, and `FindProperty` now takes a `returnRefKind`, compares it, and matches the *pointee* against the declared type for external by-ref properties.

I took the matching option rather than the "reject virtual/override ref properties" alternative because there was already a precedent: `FindMethod` has taken a `returnRefKind` and fed it to `ReturnMatches` since #490. Methods already did ref-kind override matching; only properties didn't. This brings them into line rather than inventing a rule. The change is inert for by-value properties — existing callers pass `RefKind.None`, `externalIsByRef` is false, and the comparison short-circuits to the previous behaviour.

**3 + 4 (medium) — cs2gs marked abstract/interface `ref` members translatable.** Correct, and it contradicted this PR's own stated principle. gsc rejects both with GS0578, so cs2gs was emitting G# that failed at compile several steps downstream instead of gapping loudly at the declaration — the exact failure mode #3839/#3878 established the gap to avoid. Both now report `Unsupported`. The comment asserting "the mapping is total" was the claim that was wrong, so it is corrected rather than left standing beside changed code.

**5 (low) — setter-only property reported GS0578.** Correct. The computed-getter check returned before the write accessor was inspected. The write accessor is now answered first: GS0579 names the accessor the author actually wrote, where GS0578 would send them looking for a getter shape. The bare auto-property form has no accessor object to point the message at and still falls through to GS0578, as its comment says.

## Tests

Ten new tests, and **every one was verified by reverting its fix and confirming that specific test fails** — the same anti-vacuity discipline this PR already applies, applied to the new work:

| reverted hunk | test that fails |
| --- | --- |
| `Expressions.cs` free-function deref | `RefReturningFunc_ReadFromGSharp_LoadsThroughThePointer` |
| `Expressions.cs` base-class deref | `BaseClassRefMembers_ReadFromGSharp_LoadThroughThePointer` |
| both override ref-kind guards | `RefOverrideOfByValueBaseProperty_ReportsGS0185`, `ByValueOverrideOfRefBaseProperty_ReportsGS0185` |
| accessor-order fix | `SetterOnlyRefProperty_ReportsGS0579` |
| cs2gs abstract/interface guards | the four `*_StaysLoudGap` tests |

`ConcreteRefMembers_StillTranslate` is the anti-vacuity neighbour for findings 3+4, so the new check cannot have been written as a blanket refusal of every `ref` member.

Full suites, on a tree with these fixes:

| suite | result |
| --- | --- |
| `test/Core.Tests` | 9019 / 9021 |
| `test/Compiler.Tests` | 5231 / 5234 |
| `tools/cs2gs/Cs2Gs.Tests` | 2850 / 2851 |

The 6 failures are unchanged from this branch's pre-existing baseline (2 / 3 / 1 before these fixes, same tests), and the pass counts rose by exactly the 10 tests added. All 6 are environment prerequisites on this machine, each verified by reading its assertion rather than assumed:

- 5 fail with `targeting pack root '/usr/local/share/dotnet/packs/NETStandard.Library.Ref' missing` — the pack is not installed here.
- 1 (`Issue2819_SameVersionPackedToolAndSdk_…`) fails because `ResolveLocalCs2GsPackage` returns null: it needs a locally built `.nupkg` that a Debug run never produces.

None touches a changed file.

## Docs

ADR-0060 §14 now records the override rule and the cs2gs concrete-only gap, and its claim that "G# has no ref-kind matching for property slots" is corrected — still true for interface satisfaction, no longer true for class overrides. `website/docs/ref/spec.md` and both diagnostics tables follow, including the GS0185 row.

## Not changed

The `func` ref-return read-side fix stays in this PR rather than being split out. Finding 1 turned out to be that same fix being incomplete across three further call branches, so the free-function, base-class and base-interface paths are now one coherent change with tests; separating them would carve those branches out of the property work they interleave with. Confirmed with the repo owner.

PR #4105, opened 2026-09-07
ci: distinguish hot-core timeout from cancellation

### LINKED ISSUE #4077: ci: the hot-core translation guard runs at 88-92% of its 60-minute timeout, so it times out on slower runners
The `hot-core translation guard` (`.github/workflows/cs2gs-pr-guard.yml`) now
runs at **88–92% of its 60-minute `timeout-minutes`**, so it times out on
whichever PR happens to land on a slower runner. It is a required check, so
that reads to the author as "my change hung gsc" — which is what this job is
built to detect, and what the timeout headroom exists for.

Found while investigating a timeout on PR #4060, which was attributed to that
PR's overload-resolution changes. It is not that; measurements below.

## Measured, from the Actions API

Successful runs of this job, on branches with no relation to each other:

| run | branch | started | finished | duration | headroom |
| --- | --- | --- | --- | --- | --- |
| 34081775896 | `fix/3888-expanded-params-nullability` | 04:04:30Z | 04:59:39Z | **55m 09s** | 4m 51s |
| 34079309956 | `issue-4037-4042/…` | 03:20:21Z | 04:13:48Z | **53m 27s** | 6m 33s |
| 34081308934 | `issue-4036-4051/…` | 03:56:01Z | 04:58:09Z | **62m** | **timed out** |

The workflow's own comment at `cs2gs-pr-guard.yml:82` says:

> The measured run is well under this; the headroom is for a gsc hang, which is
> a failure mode this job specifically exists to catch (#3905).

That is no longer true. Five minutes on a ~55-minute job is roughly 9%, which
is inside normal GitHub-hosted runner variance.

## The timed-out run was not a hang — measured

Reproduced locally at the same commit (`4c1cd90f`), same script,
`SELFMIG_PR_GUARD_ROOT` outside the corpus:

```
PR guard: 8/8 guarded app(s) migrated and compiled in 2859s (migrate exit 1).
```

**47m 39s, all eight apps green.** A watcher sampling `ps` every 10s and
dumping a managed stack (`dotnet-stack report`) for any `gsc.dll` child alive
past 240s captured **nothing** — no gsc invocation in the whole run came close
to stalling.

And an A/B of the single most binder-intensive unit in the guard — compiling the
migrated `GSharp.Core` (642 `.gs` files) straight from the run's own `.rsp`:

| gsc build | run 1 | run 2 |
| --- | --- | --- |
| parent `3e5103de` | 164s | 164s |
| PR #4060 `4c1cd90f` | 165s | 165s |

**+1s, +0.6%** — noise, and three orders of magnitude short of explaining a
60-minute timeout.

## A second, quieter problem: cancellations look identical to timeouts

`concurrency: cancel-in-progress: true` means a new push cancels the running
job, and that produces the same `##[error]The operation was canceled.` line a
timeout does. Three of the four recent "cancelled" runs on batch-6 PRs were
push-cancellations, not timeouts:

| run | duration | cause |
| --- | --- | --- |
| 34079861399 | 22m | cancelled by the branch's own 03:52:03Z push |
| 34081084876 | 45m | cancelled by the branch's own 04:37:33Z push |
| 34079206606 | 38m | cancelled by the branch's own 03:56:01Z push |
| 34081308934 | **62m** | **genuine timeout** |

Telling them apart requires comparing `run_started_at`/`updated_at` against the
branch's push history, which is not something an author reading a red check
will do. That is how this got attributed to a code change.

## Suggested remedies

1. **Raise `timeout-minutes`** to restore real headroom — 90 would put the
   measured 53–55 minutes back at ~60% of budget and still catch a genuine hang
   in reasonable time. This is the one-line fix.
2. **Emit progress from the migrate stage.** The job is silent for ~50 minutes
   because `cs2gs migrate` prints nothing between `excluded N project(s)` and
   its final summary. A line per app (translate/compile/ilverify/test-parity)
   would make a real hang immediately distinguishable from a slow run, and would
   name the project it hung on — which is the diagnostic this job exists to
   produce.
3. **Consider why the job costs ~55 minutes at all.** `run-cs2gs-selfmig-pr-guard.sh`
   already notes it grew "to a ~30-minute job"; it is now nearly double that.
   Whatever caused that drift is the real regression, and a raised timeout only
   hides it.

Remedy 1 unblocks the batch; 2 and 3 are the durable fixes.

## Notes

- Not channel business, so no ADR-0174 erratum.
- Not a gsc or cs2gs defect: 8/8 apps translate, compile, ILVerify and
  test-parity green locally at the head of #4060.

DavidObando:
## Summary

- impose a 75-minute migrate deadline inside the existing 90-minute job limit
- report that deadline as an explicit `Hot-core migration timeout` annotation and exit 124
- add a lightweight behavioral timeout regression using the existing fake-`dotnet` harness

Fixes #4077
Follow-up to #4104, which merged immediately before this review request.

## Progress regression gap

The merged #4104 head already replaced source-text greps with `MigrationPipelineTests.StageProgress_ReportsStartBeforePassOrFail_WithElapsedTime`. That test runs the real pipeline coordinator with existing pass/fail fake stages, captures stdout, and verifies:

- app and stage names are present
- `started` precedes `passed` / `failed`
- elapsed time has invariant one-decimal-second formatting

No additional test abstraction is needed.

## Cancellation-cause investigation

GitHub's available workflow state does not expose a cancellation-reason field. Both the timed-out run attempt and a push-cancelled run report workflow/job/step conclusion `cancelled`; `github`/`job` contexts expose status, not whether cancellation came from `timeout-minutes` or `concurrency.cancel-in-progress`.

An after-the-fact `if: always()` / `cancelled()` classifier is not reliable: GitHub re-evaluates conditions while cancelling and then signals or kills the runner process tree. In this repository, the existing `if: always()` artifact step ran after both observed causes and still had no distinguishing context.

References:
- https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-cancellation
- https://docs.github.com/en/actions/reference/workflows-and-actions/contexts

## Remedy

Rather than guessing after cancellation, CI now sets `SELFMIG_PR_GUARD_MIGRATE_TIMEOUT=75m`. The guard runs `cs2gs migrate` under GNU `timeout`:

- healthy work keeps 20+ minutes of migrate headroom over measured 50-55 minute slow runs
- a real migrate overrun ends before the 90-minute GitHub job cap, emits an explicit error annotation naming the 75-minute deadline, and exits 124 as a failure
- a superseding push still uses GitHub concurrency cancellation and remains `cancelled`, with no false timeout annotation
- the 90-minute job timeout remains a final backstop for setup or abnormal termination

This sidesteps the platform's missing cancellation-cause field without an external service, API race, or post-cancellation step.

## Validation

- `bash -n build/run-cs2gs-selfmig-pr-guard.sh build/test-cs2gs-selfmig-pr-guard.sh`
- `./build/test-cs2gs-selfmig-pr-guard.sh`
  - fake migrate exceeding a 1-second deadline exits 124
  - log identifies internal timeout and explicitly excludes push/concurrency cancellation
  - generic `no run.json produced` diagnosis is suppressed
- `git diff --check`
- no local `dotnet` restore/build/test, per constraint; GitHub CI is the full validation path

### DIFF
--- a/.github/workflows/cs2gs-pr-guard.yml
+++ b/.github/workflows/cs2gs-pr-guard.yml
@@ -92,14 +92,18 @@ env:
   # Nerdbank.GitVersioning's GitBuildVersion* emission.
   NBGV_SetCloudBuildVersionVars: false
   SELFMIG_PR_GUARD_ROOT: /tmp/gsharp-cs2gs-pr-guard
+  # GitHub reports both job timeouts and concurrency cancellations as
+  # "cancelled". Expire migrate first so the script can identify a real timeout.
+  SELFMIG_PR_GUARD_MIGRATE_TIMEOUT: 75m
 
 jobs:
   guard:
     name: hot-core translation guard
     runs-on: ubuntu-latest
     # Current eight-project runs measure ~53-55 minutes. Ninety minutes restores
-    # runner-variance headroom while keeping a finite whole-job backstop; each
-    # compiler/test child retains its tighter ProcessRunner timeout.
+    # runner-variance headroom while keeping a finite whole-job backstop. The
+    # migrate step has a 75-minute reporting deadline, and each compiler/test
+    # child retains its tighter ProcessRunner timeout.
     timeout-minutes: 90
 
     steps:
--- a/build/run-cs2gs-selfmig-pr-guard.sh
+++ b/build/run-cs2gs-selfmig-pr-guard.sh
@@ -77,6 +77,9 @@
 #                          set can be widened (e.g. to tools/cs2gs/Cs2Gs.Tests
 #                          once #3836's ~9 known failures clear) without
 #                          editing this file.
+#   SELFMIG_PR_GUARD_MIGRATE_TIMEOUT
+#                          optional GNU timeout duration for migrate; CI sets
+#                          75m so it can report timeout before the 90m job cap.
 set -euo pipefail
 
 repo_root=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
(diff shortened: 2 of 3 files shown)


DavidObando:
Superseded by #4106. #4106 keeps the timeout headroom, adds the missing runtime short-circuit, and classifies GitHub's own timeout/concurrency annotations from a separate job that survives cancellation rather than imposing a second migrate deadline.

PR #4109, opened 2026-09-07
cs2gs: exclude validation artifacts from readability counters

### LINKED ISSUE #4085: selfmig: post-validation readability counters include test-generated Snippet.gs artifacts and breach the zero-label ratchet
## What happened

A clean classic full self-migration run on current `origin/main` (`7708be7d9`, run `2026-09-07T10-50-20Z_f80427`) reached the expected **50/56 green** and kept every other ceiling in bounds, but the gate failed:

```
self-migration: 50/56 green (floor 50); labels=4 (ceiling 0); __local_=30 (ceiling 30); lines>300=442 reducible (ceiling 450), 150 single-atom-bounded; bangs=11212 (ceiling 11500)
GATE: synthetic label count 4 exceeded ceiling 0.
```

The four labels are not migrated repository source. They are two `goto`/label pairs in transient files created by migrated `Cs2Gs.Tests` while test-parity runs:

```
migrated/out/bin/Release/Cs2Gs.Tests/issue-2231-e2e/<guid>/Snippet.gs
```

`LocalFunctionHoistTranslationTests.CompileAndRun` deliberately writes each snippet under `AppContext.BaseDirectory/issue-2231-e2e/<guid>`. Because the classic gate calls `selfmig_measure "$migrated_dir"` **after** all validation/test-parity, `find ... -name '*.gs'` counts those test scratch files as corpus output. A second run on `befc1e925` reproduced the same 4-count failure with different GUIDs.

## Root / scope

This is counter-scope contamination, not a returned `__patternGuardEnd` lowering in the migrated repository. The pre-validation tree does not contain these files; test-parity manufactures them beneath the measured tree. Lowering `syntheticLabelCeiling` or excluding the family would hide the ratchet and is not the fix.

The narrow fix should make post-validation readability measurement ignore generated build/test scratch (at least `out/`, `bin/`, and `obj/` trees) while continuing to count translated source files. Add a regression that creates a synthetic-bearing `.gs` file under a test output path and proves it does not affect the counter, with a source-tree control that still does.

## Why this was found

While revalidating #3907/#3932, `src/Sdk/Gsharp.Runtime.Channels` was green through translate, compile, ILVerify, and parity in both the single-app gate and this full run. This is an independent residual split rather than a Channels failure.

DavidObando:
## Summary

- route every readability metric through one translated-source finder
- match `RepositoryFileInventory` by pruning `bin/`, `obj/`, and `TestResults/` segments case-insensitively
- keep legitimate translated source under `out/` measurable
- add shell regressions for test-generated `Snippet.gs` contamination and mixed-case artifact layouts

## Root cause

Classic self-migration measures the migrated tree after validation and test parity. `Cs2Gs.Tests` creates transient `.gs` snippets beneath paths such as `migrated/out/bin/Release/Cs2Gs.Tests/issue-2231-e2e/<guid>/Snippet.gs`; the prior unrestricted `find ... -name '*.gs'` treated those test artifacts as translated corpus source and breached the zero synthetic-label ratchet.

## Fix

`cs2gs_find_translated_sources` is now the shared file-selection point for filtered lines, raw lines, long-line classification, and report file counts. Its `find -iname` segment pruning matches the established `RepositoryFileInventory` generated-path predicate for `bin`, `obj`, and `TestResults`. It deliberately does not prune `out`: self-migration includes legitimate source there, including the green app `out/scratch/ildump/ildump.csproj`. The reported contamination under `out/bin/...` remains excluded by the nested `bin` prune.

## Regression

`build/test-cs2gs-counters.sh` creates identical synthetic-bearing content under `out/bin` and mixed-case `BiN`, `OBJ`, and `TeStReSuLtS` paths, then confirms every readability metric remains zero. It copies the same file to `out/scratch/Translated.gs` and confirms labels, lifted locals, long lines, null assertions, synthetic totals, raw counters, and `.gs` file totals count that legitimate translated source.

## Validation

- `bash -n build/cs2gs-counters.sh build/selfmig-common.sh build/test-cs2gs-counters.sh`
- `build/test-cs2gs-counters.sh`
- `git diff --check`
- all readability metric file traversal re-audited through `cs2gs_find_translated_sources`
- no local dotnet build, restore, or test commands run

Fixes #4085

### DIFF
--- a/build/cs2gs-counters.sh
+++ b/build/cs2gs-counters.sh
@@ -78,7 +78,11 @@ cs2gs_find_translated_sources() {
   local tree=$1
   shift
   find "$tree" \
-    \( -type d \( -name bin -o -name obj \) -prune \) -o \
+    \( -type d \( \
+      -name '[Bb][Ii][Nn]' -o \
+      -name '[Oo][Bb][Jj]' -o \
+      -name '[Tt][Ee][Ss][Tt][Rr][Ee][Ss][Uu][Ll][Tt][Ss]' \
+    \) -prune \) -o \
     \( -type f -name '*.gs' "$@" \)
 }
 
--- a/build/cs2gs-counters.sh
+++ b/build/cs2gs-counters.sh
@@ -70,6 +70,18 @@ cs2gs_synthetic_families=(
 # How many distinct unknown identifiers to name in the table before truncating.
 cs2gs_unknown_sample_limit=${CS2GS_UNKNOWN_SAMPLE_LIMIT:-25}
 
+# Visits translated source files while pruning conventional build-output trees.
+# Validation can create transient .gs test fixtures under these directories;
+# they are artifacts, not translator output, and must not move readability
+# counters. Callers supply the find action so every metric shares this scope.
+cs2gs_find_translated_sources() {
+  local tree=$1
+  shift
+  find "$tree" \
+    \( -type d \( -name out -o -name bin -o -name obj \) -prune \) -o \
+    \( -type f -name '*.gs' "$@" \)
+}
+
 # Emits the CODE lines of a migrated tree: every line of every .gs file minus
 # the ones that are not code for metric purposes.
 #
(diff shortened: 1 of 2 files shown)


DavidObando:
Review corrections pushed in `923a8672e4936eef1ec538e6f09b4689d2772ab8`.

- Preserves legitimate `out/scratch/ildump/Program.gs` source.
- Excludes `bin`, `obj`, and `TestResults` directory segments case-insensitively using portable bracket patterns.
- Covers original-like `out/bin/.../Snippet.gs` plus mixed-case `out/BiN`, `out/OBJ`, and `out/TeStReSuLtS` artifacts.
- Asserts `.gs` file count and every top-level readability metric before and after adding the legitimate source.

Validated with Bash syntax checks, `build/test-cs2gs-counters.sh`, and `git diff --check`. No dotnet/build/restore commands run.

PR #4123, opened 2026-09-08
cs2gs: track raw strings lexically, not by backtick parity (#4082)

### LINKED ISSUE #4082: cs2gs counters: a backtick inside a double-quoted string desyncs raw-string tracking, misfiling 58 reducible long lines as unfixable
Found while building a long-line inventory of the migrated tree for ADR-0179 phase 6. Measured on gate run [34073123904](https://github.com/DavidObando/gsharp/actions/runs/34073123904) (gsc `0.4.531+acb89d8346`, PASSED at 50/56), reproducing the gate's own numbers exactly before correcting them.

## The defect

`cs2gs_long_line_counts` (build/cs2gs-counters.sh:101) decides whether a line sits inside a raw string by counting backticks per line and toggling a flag on odd parity:

```python
if raw_line.count("`") % 2:
    in_raw = not in_raw
```

It counts backticks **inside ordinary double-quoted strings too**. G#'s raw string is backtick-delimited, but a backtick inside a `"…"` literal is just text.

`src/Core/CodeAnalysis/DiagnosticDescriptors.gs` line 155 carries this message:

```
"Unsupported documentation Markdown: {0}. Use ```xmldoc for complex XML-doc constructs."
```

Three backticks, inside a normal string. `in_raw` flips to `true` there and **nothing in the file ever flips it back** — it is the only odd-parity line in the file.

## What that does to the numbers

While `in_raw` is wrongly true the counter takes the widest atom to be the entire line:

```python
if in_raw:
    widest = len(raw_line.lstrip())
```

which makes `indent + 8 + widest > 300` true for essentially any long line, so every one of them is filed as **single-atom-bounded** — the bucket meaning "no formatter can shorten this without changing the token stream." In that one file, 1 long line precedes line 155 and **54 follow it**; all 54 are misfiled. They are ordinary multi-argument `DiagnosticDescriptor(...)` calls that an argument-list wrap would fix.

Tree-wide, on the run above:

| counting method | reducible | single-atom-bounded | total |
|---|---:|---:|---:|
| as the gate counts it today | 438 | 143 | 581 |
| with backticks inside string literals masked | **496** | **85** | 581 |

**58 lines are misclassified**, in the direction that hides work.

## Why this matters more than a reporting nit

1. **`reducible` is the GATED number.** `longLineCeiling` is compared against it, so the gate is enforcing a ceiling on an undercount and reporting less outstanding work than exists.
2. **ADR-0179's floor estimate is wrong.** The ADR states "the floor under this metric is ~27 string lines, not 0", computed from an atomic bucket inflated by this bug.
3. **Phase 6 was sized against it.** 6a's "317 of 537" and 6b's "173" both come from this counter.

## The fix, and the ratchet consequence

Mask double-quoted string literals before counting backticks — the same `strip_dq` shape the classifier needed:

```python
def strip_dq(s):
    out, i, n = [], 0, len(s)
    while i < n:
        if s[i] == '"':
            j = i + 1
            while j < n and s[j] != '"':
                j += 2 if s[j] == '\\' else 1
            i = j + 1
        else:
            out.append(s[i]); i += 1
    return ''.join(out)
```

then toggle on `strip_dq(raw_line).count("`") % 2`. Worth also asking whether a *file*-level scan would be more robust than per-line parity, since any future odd-parity line has the same blast radius.

**`longLineCeiling` must be re-baselined in the same PR.** It is 450 today, set in #4047 against the under-reported 438. A corrected counter measures 496 and breaches it immediately. That is a measurement correction, not a readability regression, and the baseline comment should say so explicitly — otherwise the next reader sees a ceiling raised 450 → ~510 with no incident behind it and reasonably reads it as drift. The other four metrics are unaffected.

A regression test is cheap and worth having: a fixture line containing a backtick inside a double-quoted string, asserting the following long line is still counted reducible.

Part of #3501 (Track C) and ADR-0179.

DavidObando:
Fixes #4082.

## The defect

`cs2gs_long_line_counts` decided whether a line sits inside a backtick raw string by toggling a flag on **per-line backtick parity**:

```python
if raw_line.count("`") % 2:
    in_raw = not in_raw
```

That counts backticks which are not raw-string delimiters at all. Three live sources in the corpus:

| source | example |
|---|---|
| backtick inside a `"…"` literal | `src/Core/CodeAnalysis/DiagnosticDescriptors.gs`: `"Unsupported documentation Markdown: {0}. Use ```xmldoc for complex XML-doc constructs."` |
| backtick inside a `//` comment | `src/Core/CodeAnalysis/Binding/Binder.gs`: ``// imports (mangled name `Name`N`). This lets users write `List[int]`…`` |
| backtick as a character literal | `tools/cs2gs/Cs2Gs.CodeModel/Printing/GSharpPrinter.gs`: ``if literal.Value.IndexOf('`') < 0 {`` |

Each flips the flag **on** with nothing later in the file to flip it back. While it is on, the widest atom is taken to be the *whole line*, so `indent + 8 + widest > 300` is trivially true and every subsequent long line in that file is filed as **single-atom-bounded** — the bucket meaning "no formatter can shorten this". They are ordinary multi-argument calls an argument-list wrap would fix.

## Measurement

Reconstructed the tree of gate run [34232380469](https://github.com/DavidObando/gsharp/actions/runs/34232380469) (gsc `0.4.575+033c38dfb8`) — the `cs2gs-selfmig-migrated` artifact with all six shards' `polished.tar.gz` extracted over it, which reproduces the gate's own numbers — and counted it three ways:

| counting method | reducible | single-atom-bounded | total |
|---|---:|---:|---:|
| as the gate counts it today | 433 | 157 | 590 |
| masking only `"…"` literals (the fix #4082 proposed) | 503 | 87 | 590 |
| **this PR (file-level lexical scan)** | **554** | **36** | 590 |

**121 of 590 long lines were misfiled**, every one of them in the direction that hides work from the *gated* number.

## Why a file-level scan, not a wider per-line mask

The issue proposed a `strip_dq` helper. It recovers 70 of the 121 and leaves 51 to the comment and character-literal cases — and, more importantly, it leaves the fragility in place. Per-line parity fails **unbounded**: one bad line silently corrupts every line after it in the same file, which is exactly why a single message string cost 54 lines in `DiagnosticDescriptors.gs`.

`raw_string_flags` instead tracks the lexical state each line *starts* in — backtick raw strings (no escape, so the next backtick always closes one), `"…"` literals with backslash escapes and `${…}` interpolation holes whose contents are code and may contain nested strings, `'…'` character literals, `//` line comments and `/* */` block comments. The blast radius is bounded too: only raw strings and block comments may span a newline, so every other state is reset at the next line and a future misread costs one line instead of a file.

Spot-checked the result: only 3 long lines in the whole tree are now judged to sit inside a raw string, and all three genuinely do (two begin with the closing backtick of a multi-line raw string, the third is embedded C# inside a `.gs` test fixture). The `${…}`-hole handling is load-bearing — without it `GSharpPrinter.gs:1067` (a nested string literal inside an interpolation hole) desyncs the scanner in the *opposite* direction.

## Re-baselining `longLineCeiling` 450 → 570

**This is a measurement correction, not a readability regression.** The tree did not change: no line got longer, no line got added. 121 lines moved out of a bucket where they were invisible and into the one that says an argument-list wrap would fix them — work made *visible* to ADR-0179 phases 6–7, not work newly created.

570 keeps ~3% margin over the measured 554, the convention this file already uses. The ledger entry in `tools/cs2gs/selfmig-baseline.json` records the mechanism, the before/after numbers, and the fact that only the measurement moved, so a future reader does not have to take a 450 → 570 raise on faith. No other metric changes: `greenFloor` stays 52, `greenApps` and `stageFloor` are untouched, and labels/lifts/bangs never consulted raw-string state.

## ADR-0179 correction

ADR-0179 states "the floor under this metric is ~27 string lines", computed from a `single-atom-bounded` bucket the counter was inflating. The corrected bucket on this tree is **36** — 33 lines that *are* one over-budget string literal, plus 3 raw-string body lines — which is consistent with the ~27 estimate rather than with the 157 the gate has been reporting. Annotated in the same PR so the inflated figure is not reused.

## Tests

`build/test-cs2gs-counters.sh` gains a four-file fixture, one per trigger (a single file mixing them would let two odd-parity lines cancel and hide the defect), plus a fourth file containing a *real* raw string that must still count as single-atom-bounded — so the test fails on an over-correction as well as on the original bug.

**Verified against a reverted counter**: `0 4 4` before the fix, `3 1 4` after.

### Also fixed: the test file asserted nothing on macOS

Checking that last point is how this turned up. bash 3.2 — still the system bash on macOS, and therefore what a contributor validating locally runs — does **not** honour `set -e` for a failing `[[ ]]`. Every bare assertion in `build/test-cs2gs-counters.sh` was a silent no-op locally, while failing correctly on CI's bash 5. Reverting the counter and watching the file print `cs2gs counter contract tests passed` anyway is what exposed it. The assertions now go through an `assert_eq` helper that exits explicitly and names the mismatch.


DavidObando:
Both review findings addressed in c055b29d — thanks, both were real and both were verified against `src/Core/CodeAnalysis/Syntax/Lexer.cs` rather than taken on faith.

## Finding 1 — `$$` escape (`cs2gs-counters.sh:202`)

Confirmed at `Lexer.cs:905-915`: `$$` inside a `"…"` literal is the escape for a literal `$`, so in `"$${"` the brace is an ordinary character and the literal runs on. The scanner read the second `$` as an interpolation opener, entered a false hole, and then took the literal's own closing quote for the **start** of a new literal — so a raw-string opener later on that line was swallowed and its body lines misclassified. Fixed by consuming the escaped pair before testing for `${`.

## Finding 2 — multiline interpolation holes (`cs2gs-counters.sh:218`)

Confirmed at `Lexer.cs:824-830` ("Multiline holes are handled inside the `${ … }` scanner and never reach here") and `ScanInterpolationHole`, whose doc comment says outright that newlines are permitted; `samples/InterpolatedStringRichHoles.gs:17-19` carries a real one. While a hole spans a newline the scanner's state is `code`, so the per-line reset cleared the hole stack, the continuation's `}` could no longer restore `dq`, and the rest of that line was read in the wrong state. Fixed by carrying `holes`/`depth` across the boundary while still resetting the non-multiline literal states (`dq`, `char`) there.

That does not reintroduce an unbounded failure mode: a stale hole entry can only turn a later top-level `}` into `dq`, and `dq` is itself reset at the next line boundary.

## Regression check: the count does not move

Both bugs are **latent** on this corpus, and that gives a strong check. Re-measured the run [34232380469](https://github.com/DavidObando/gsharp/actions/runs/34232380469) tree (`cs2gs-selfmig-migrated` with all six shards' `polished.tar.gz` extracted over it, 4,253 `.gs` files):

| | reducible | single-atom-bounded | total |
|---|---:|---:|---:|
| PR head before this commit | 554 | 36 | 590 |
| **with both fixes** | **554** | **36** | **590** |

69 `$${` sites exist in the tree, including in the three files that dominate the long-line count; neutralizing the `$${` in `DiagnosticDescriptors.gs`, `GoldenTests.gs` and `TypeClauseCompletions.gs` and re-counting moves nothing either, and the one multiline-hole site with a long line after it (`GSharpPrinter.gs`) is likewise inert. `longLineCeiling` stays **570**, still sized against 554.

## Why fix a latent bug

Because the count being right today is luck, not design. This PR's whole thesis is replacing a fragile heuristic with a faithful lexer, and #4082 was expensive precisely because one misread line corrupted every line after it in the same file. Shipping two constructs the real lexer accepts and this scanner does not leaves exactly that class of failure armed, waiting for a `$${` or a multiline hole to land in front of a raw string — a translator change or one new source file away.

## Tests

One fixture per finding in `build/test-cs2gs-counters.sh`, each a genuine raw string whose body line flips from single-atom-bounded to reducible the moment the scanner desyncs. Separate trees, because a shared one would let either fix mask the other's fixture.

**Verified each fails with only its own fix reverted**, not merely green with both applied:

```
FAIL: PR #4123: $$ escape does not open a hole: expected '0 1 1', got '1 0 1'
FAIL: PR #4123: multiline interpolation hole: expected '0 1 1', got '1 0 1'
```

Both go through the `assert_eq` helper this PR added, so they are real assertions on macOS bash 3.2 as well as on CI's bash 5.

Issue #4045, opened 2026-09-06
selfmig: migrated test/Compiler.Tests parity — a nominal delegate erased, `!!` on an expected null, and a run that never finishes


DavidObando:
Measured on gate run [34036106619](https://github.com/DavidObando/gsharp/actions/runs/34036106619) (gsc `0.4.520+a09f4e1b22`, PASSED at 50/56). `test/Compiler.Tests` translates, compiles and ilverifies, then fails at **test-parity**. It has no issue yet.

## First, the measurement is incomplete — and that is its own defect

```
LIBRARY-TESTS-TIMED-OUT: the migrated test run exceeded its 00:10:00 budget and was killed;
the output below is TRUNCATED and carries no run summary, so its [FAIL] list is not a parity count.
```

The harness is explicit that a killed run yields no parity count. **10 distinct failures are visible before the kill; the real number is unknown and can only be larger.** So this app cannot be driven to green by fixing what is listed here — the budget has to come up (or the app has to shard) before anyone can even see the finish line. `test/Compiler.Tests` compiles G# source to assemblies inside its own test bodies, so it is legitimately one of the slowest apps in the corpus; this is not obviously a hang.

The 10 visible failures:

```
BareEventRaise_NullBackingField_IsSafeNoOp
BaselineCompiledCSharpConsumer_BindsAndRunsAgainstMigratedContract
ChangedSettings_Subscription_Forwards_To_Inner
GenericEventHandlerRaise_NoSubscribersAndWithSubscriber_Behaves
NominalEventHandlerRaise_NoSubscribers_IsSafeNoOp
NominalEventHandlerRaise_UnsubscribeThenRaise_IsSafeNoOp
NominalEventHandlerRaise_WithSubscriber_InvokesHandler
Override_Wins_Over_Inner_Quality
StructuralEventArgsRaise_NoSubscribers_IsSafeNoOp
StructuralEventHandlerEvents_ExposeNominalMetadataAndVerify
```

Both cs2gs defects below are in `test/Compiler.Tests/Emit/Issue2796NullConditionalEventRaiseEmitTests.cs`, and between them they account for the seven event-raise failures.

---

## RC1 — a nominal delegate type annotation is erased to G#'s structural arrow type (cs2gs)

C#:

```csharp
EventHandler handler = (_, _) => hits++;
ev.AddEventHandler(instance, handler);
```

Emitted G#:

```gsharp
[...]

Issue #3869, opened 2026-09-03
selfmig: test/Core.Tests' test-parity passes with ZERO tests — cs2gs drops `ref` from a `ref struct`, the assembly fails type-load, and the gate scores it green (#3863)


DavidObando:
Found while clearing the `test/Core.Tests` ILVerify wall (#3863). With #3867/#3868 fixed, `test/Core.Tests` reaches `test-parity` for the first time — and the stage reports **PASS while running zero tests**.

## What actually happens

`dotnet test` on the migrated `GSharp.Core.Tests.dll`:

```
[xUnit.net 00:00:00.08] Exception discovering tests from GSharp.Core.Tests:
  System.TypeLoadException: A ByRef or ByRef-like type cannot be used as the type
  for an instance field in a non-ByRef-like type.
   at System.Reflection.RuntimeAssembly.GetExportedTypes()
   at Xunit.Sdk.ReflectionAssemblyInfo.GetTypes(Boolean includePrivateTypes)
   …
No test is available in …/GSharp.Core.Tests.dll.
```

`dotnet test` still exits **0**, and `TestParityStage.RunMirroredTestProject` passes on `result.ExitCode == 0` alone:

```csharp
ProcessRunResult result = SdkCompileRunner.TestMirroredProject(...);
this.Note(context, result.Output ?? string.Empty);
if (result.ExitCode == 0)
{
    return StageOutcome.Passed();
}
```

So the app is reported fully green — `translate PASS / compile PASS / ilverify PASS / test-parity PASS` — on a migrated assembly that cannot even be enumerated. This is the #1831 class again, one stage further along.

## Defect 1 (cs2gs): the `ref` is dropped from a `ref struct`

`test/Core.Tests/CodeAnalysis/Binding/Issue2852GenericUnmanagedStackAllocTests.cs`:

```csharp
public ref struct Issue2852ImportedRefStruct
{
    public Span<int> Values { get; set; }
}
```

translates to

```go
struct Issue2852ImportedRefStruct {
    prop Values Span[int32]
}
```

— the `ref` modifier is gone. G# **does** have the construct (`ref struct Window {` — see `samples/RefStructGenericField.gs`, issue #367 / ADR-0056 §4), and gsc emits `IsByRefLikeAttribute` for it (`TypeDefEmitter.EmitIsByRefLikeAttribute`), so this is a translator fidelity loss, not a gsc gap.

[...]

PR #4125, opened 2026-09-08
cs2gs: size the mirrored test-parity budget to the suite

DavidObando:
## What is broken

`tools/cs2gs/Cs2Gs.Tests` fails the self-migration gate at test-parity, and it fails by being **killed**:

```
LIBRARY-TESTS-TIMED-OUT: the migrated test run exceeded its 00:10:00 budget and was killed;
the output below is TRUNCATED and carries no run summary, so its [FAIL] list is not a parity count.
```

The app translates, compiles and ILVerifies. The budget already had a per-app shape — `MirroredTestRunTimeoutFor(string appId)` gave `test/Compiler.Tests` 90 minutes (#4045) and everything else the 10-minute default (#3931) — and that precedent is right on both counts: a budget belongs per app, and the tight default must stay so hangs still fail promptly. What it lacks is a rule. `Cs2Gs.Tests` declares **2,579 `[Fact]` methods** (2,891 executed cases; the C# **original** takes 21 minutes), falls to the `else`, and is killed for being large rather than for being stuck — it could not have finished inside ten minutes in any state of health.

This PR keeps the mechanism and replaces its `string.Equals` arm with a rule. The alternative — a second hardcoded arm for `Cs2Gs.Tests` — was rejected for one reason: it fixes the app that already tripped the wall and leaves the next one to trip it silently, which is how this app got here. `Compiler.Tests` still gets exactly 90 minutes, now by rule instead of by name.

The cost is not a slow job. A killed `dotnet test` emits no VSTest summary, so the app produces **no parity count at all**: nobody can see how many tests fail, and the app cannot be driven to green. This is a measurement blocker.

## The fix: a budget derived from the app's own size

`MirroredTestRunTimeoutFor` is the same method in the same place, with the same two call sites; only its input changes. Instead of an app id it takes the number of `[Fact]` methods the app's C# original declares — the number `TestParityStage.CountCSharpFactMethods` already computes and the gate already trusts as the lower bound on executed cases (#3869) — and returns:

```
max(10 min, ceil(5 min + 1.5 s × declaredFacts))   clamped to 90 min
```

* **Floor — 10 min**, unchanged. A 30-test suite that has not finished in ten minutes is stuck, not big, and a hang there must still fail promptly. About fifty corpus apps stay exactly where they are.
* **Rate — 1.5 s per declared `[Fact]`.** Calibrated, not guessed. In the same nightly, migrated `test/Compiler.Tests` (4,103 declared facts) completed its run: `Failed! - Failed: 79, Passed: 5415, Total: 5494, Duration: 1 h`, inside its 90-minute grant, with ~81 minutes of whole-app validation. That is ~0.9–1.0 s per declared fact on a CI runner. The C# originals agree on the shape (~0.45 s per executed case; declared facts undercount executed cases because a `[Theory]` expands). 1.5 s is that observed rate plus ~50% for runner variance — enough to tell "large" from "stuck" without waiting out a real hang.
* **Ceiling — 90 min**, exactly what #4045 granted `Compiler.Tests` by hand. It is kept because a validation shard runs all its apps under one `cs2gs validate` process that writes `shard-run.json` only at the end: a shard that overruns the job's 240-minute limit loses *every* app's verdict — the same "no measurement at all" failure, one level up.
* The `#3931` wording is preserved and now strictly stronger: the budget the killed run was **actually held to** is threaded from `RunMirroredTestProject` into the diagnostic rather than recomputed, and the sizing decision is written to the run log up front (`mirrored test run budget: 01:10:00 for 2579 declared [Fact] method(s) …`), so a future reader of a timeout can tell "starved" from "stuck".

`Cs2Gs.Tests` is the only app the flat budget still kills. In run 34232380469 the two other test-parity reds both COMPLETED and reported real counts — `Compiler.Tests` `LIBRARY-TESTS-FAILED` with 79/5,494, `Interpreter.Tests` likewise — so this is the last app in the corpus whose parity number is invisible.

### Effect, app by app (declared `[Fact]` counts measured with the production counter)

| app | declared `[Fact]` | before | after |
|---|---|---|---|
| `tools/cs2gs/Cs2Gs.Tests` | 2,579 | 10 min (killed) | **70 min** |
| `test/Compiler.Tests` | 4,103 | 90 min (hand-set, #4045) | 90 min (ceiling — formula asks 108) |
| `test/Core.Tests` | 7,419 | 10 min | 90 min (ceiling; red at *compile*, never reaches the run today) |
| `test/Interpreter.Tests` | 817 | 10 min | 26 min |
| `test/LanguageServer.Tests` | 418 | 10 min | 16 min |
| everything else (≤189 facts, incl. all non-test apps) | ≤189 | 10 min | 10 min (unchanged) |

The #4045 special case is deleted: `Compiler.Tests` now reaches the same 90 minutes by the rule instead of by name, so its budget does not regress and no app's worst case rises above what #4045 already established.

## Options rejected

* **A flat raise of the constant.** Sizing one number for the biggest suite makes every genuine hang cost that much everywhere: a 30-test app that deadlocks would burn 70 minutes before anyone learned anything, on ~50 apps that have no need of it.
* **Deriving the budget from `build/selfmig-shard-costs.json`.** Tempting — it is per-app and already measured — but circular. Those numbers are *wall times republished by each nightly*, so a killed run's recorded cost is capped by the very budget being derived, and the number a hang writes would fund the next hang. It is also whole-app validation time (translate-side polish, compile, ILVerify, test), not test-run time. A declared-test count is a property of the source: it moves only when someone writes or deletes a test.
* **Making the run cheaper (parallelism / sharding the mirrored test run).** Real, but a bigger and riskier change to land while the measurement is still blocked: xunit already parallelizes collections in-process, and splitting the run into N processes means N run summaries to merge — precisely the surface #3869/#3872 hardened (a lost test must not be able to hide behind a summary). Worth revisiting once `Compiler.Tests` and `Cs2Gs.Tests` are green and we know what the real runtimes are.

## CI cost

Latest nightly (34232380469): migrate 26 min; validate shards 52 / 85 / 64 / 61 / 52 / 69 min; whole workflow 112 min. The critical path is the shard holding `Compiler.Tests` (4,864 s of app cost).

`Cs2Gs.Tests` currently costs 1,650 s (27.5 min), of which 600 s is the kill. Completing the run at `Compiler.Tests`' observed migrated rate (~1.0 s/declared fact ⇒ ~43 min of tests) puts the app near **60 min**, i.e. **+30 min on its shard**. Expected nightly wall clock: **112 → ~115–135 min**, with aggregate runner time up ~10% (≈ 35 min out of ~6 h). If the suite genuinely needs its full 70-minute budget, that shard lands near 105–120 min — still less than half the 240-minute job limit. Shard packing self-corrects on the next run: the gate republishes `selfmig-shard-costs.json` from what it measured, so the newly-expensive app is packed first rather than landing last on a full shard.

No other budget changes, and the mirrored **build** budget is untouched.

## What this does NOT do

**This does not make `tools/cs2gs/Cs2Gs.Tests` green, and it is very likely to stay red.** The goal is that the run **completes and produces a real parity count**, so the failures the timeout has been hiding become visible and fixable. A still-red `Cs2Gs.Tests` after this lands is the expected outcome, not a failed fix — the difference is that its `[FAIL]` list will finally be a parity result instead of a truncation artifact. Those failures deserve their own issue once the first complete run reports them.

`stageFloor` for `tools/cs2gs/Cs2Gs.Tests` stays `test-parity`: the app still reaches that stage, and nothing here can move it backwards. No gated metric moves — `build/cs2gs-counters.sh` and `tools/cs2gs/selfmig-baseline.json` are deliberately untouched (#4082 is in flight there).

## A complete run, achieved locally — and what it shows

An exact-app self-migration was run locally with this change (whole-repo `cs2gs migrate --translate-only`, then `cs2gs validate --app tools/cs2gs/Cs2Gs.Tests/Cs2Gs.Tests.csproj`, Release):

| stage | result |
|---|---|
| translate | PASS (56/56 apps) |
| compile | PASS (735 s) |
| ILVerify | PASS (0.7 s) |
| test-parity | **FAIL — `LIBRARY-TESTS-FAILED`**, i.e. a real result |

```
Failed!  - Failed: 137, Passed: 2754, Skipped: 0, Total: 2891, Duration: 8 m 20 s
```

That is the whole point of this PR: the app is still red, but its `[FAIL]` list is now a **parity count** instead of a truncation artifact. Two facts fall out of it immediately, neither of which was knowable before:

* **No test is lost.** The migrated assembly executed 2,891 cases — exactly what the C# original runs. This is a pure outcome-parity gap, not a discovery gap.
* **The migrated suite is not slow.** 8 m 20 s in Release locally versus 21 m 46 s for the C# original in Debug; the CI kill at ten minutes was a runner-speed margin, not a hang. The 70-minute budget is therefore mostly margin — which is the correct side to err on, since a budget only costs wall clock when a run does not finish.

The 137 failures are catalogued in **#4127** (50 NREs concentrated in `MigrationPipeline.RunAsync` and `ReportModel.BuildAppReport`, ~36 fixture/IO misses, 35 assertion mismatches, and one real `GS0264` delegate-overload-set collapse).

The run also caught a self-referential trap worth naming: the corpus-measuring test added here read `tools/cs2gs/Cs2Gs.Tests`'s C# sources, and in the migrated mirror that directory holds `.gs` files and no C# — so the test proving this app is no longer starved of budget would have been its 138th parity failure. The second commit guards it, asserting the condition that makes the claim inapplicable rather than swallowing an empty measurement. Found by running the thing, not by reasoning about it.

## Tests

`tools/cs2gs/Cs2Gs.Tests/Issue3501ScaledTestParityBudgetTests.cs` pins the properties that make the sizing safe: small apps keep the tight budget, an uncountable app falls back to the floor rather than below it, a several-thousand-test suite clears its measured C# original, no suite exceeds the ceiling, the budget is monotonic in suite size (a PR that adds tests can never shorten the budget), the value stays whole minutes because it is quoted in the diagnostic, and the timeout artifact names *this app's* budget. One test measures the real repository through the production counter, so the claim "the app this unblocks gets more than the old constant" tracks the corpus instead of restating a constant. `Issue3931KilledTestRunClassificationTests` keeps its two classification directions and re-expresses the #4045 guarantee against the sizing function.

Local verification: `tools/cs2gs/Cs2Gs.Tests` — `Failed: 1, Passed: 2890, Total: 2891, Duration: 21 m 46 s`. The single failure is `Adr0151IfLetExpressionTranslationTests.Issue2819_SameVersionPackedToolAndSdk_CompileTranslatedPatternVariable`, the known pre-existing case that needs a locally built Release `.nupkg`; it fails the same way on an unmodified tree.


DavidObando:
Both review findings addressed in 630bd846. Documentation only — the formula, the constants, the clamps and every expected value are untouched; the budget is still `max(10 min, ceil(5 min + 1.5 s × declaredFacts))` clamped to 90 min.

### 1. ADR-0179 overstated what the sizing retires

The claim "each test added raises its budget with it" is wrong three ways, and I confirmed all three against the code:

* `HasFactAttribute` (`TestParityStage.cs:502-509`) matches only `Fact`/`FactAttribute`, so a **`[Theory]` is not counted at all** — adding one adds executed cases and zero budget.
* The result is `Math.Ceiling(seconds / 60.0)`, so at 1.5 s per fact it takes **~40 added facts to buy one minute**; a single addition usually buys none.
* It clamps at both ends, so inside the 10/90-minute bounds additions can change nothing at all.

The row now says the budget is **non-decreasing** in the declared `[Fact]` count — which is what the code actually guarantees, and what the monotonicity test asserts — and **keeps the residual risk** instead of describing it as retired:

> Reduced, not retired: the mirrored-test budget is now derived from the app's declared `[Fact]` count rather than fixed for the whole corpus (#3501), so `LanguageServer.Tests` is budgeted for the tests it actually has (~16 minutes today) and that budget is **non-decreasing** in the declared count — adding tests can never shrink it. It does not track every addition, though: only `[Fact]` methods are counted (a `[Theory]` is not counted at all), the value is rounded up to whole minutes, and it is clamped to 10 and 90 minutes, so ~40 added facts buy one extra minute and most single additions buy none. Residual risk: a theory-heavy or unusually expensive batch can grow the executed-case count substantially while the budget does not move at all, putting this project back into truncation — check such a batch against the measured run time rather than assuming the sizing absorbed it

### 2. The calibration record cited a number that was not from the run it named

Confirmed from the run's own artifact rather than from the PR text — `cs2gs-selfmig-shard-1` of run 34232380469, `test-parity-051f728eed54.json`:

```
Failed!  - Failed:    79, Passed:  5415, Skipped:     0, Total:  5494, Duration: 1 h - GSharp.Compiler.Tests.dll (net10.0)
```

So the cited **5,234 was never that run's number**. It is a genuine figure from a *different* measurement — the C# **original** timed locally in Debug on 2026-09-06 — that had been attached to the CI citation. Both numbers were real; the attribution was wrong.

The comment now splits the record into two explicitly labelled sources so they cannot be conflated again, and says outright that the rate comes from declared facts over wall time, so an executed-case count is context and not an input:

* **CI**, gate run 34232380469 (`cs2gs-selfmig-nightly` on `main`, Release, GitHub-hosted runner): migrated `test/Compiler.Tests`, 4,103 declared `[Fact]` methods, reported `Failed: 79, Passed: 5415, Skipped: 0, Total: 5494`. Whole-app validation **4,864 s (~81 min)** (`shard-costs.json`), of which the test run was **3,658 s (~61 min)** (stage timing in the job log) — **~0.9 s** per declared `[Fact]` for the run alone, **~1.2 s** with the rest of validation charged to the same count.
* **LOCAL**, C# **originals** in Debug on a developer machine: `Cs2Gs.Tests` 2,891 cases in 21 m 57 s (re-measured 2026-09-08), `Compiler.Tests` 5,234 cases in 41 min (2026-09-06) — ≈ 0.45 s per executed case.

The derived rate is unchanged: 1.5 s still covers the observed ~0.9–1.2 s with margin.

### 3. Found while checking the rest of the record: the same conflation, twice more

`MirroredTestRunTimeoutFor`'s remarks read `Cs2Gs.Tests` "(2,579 declared `[Fact]` methods, 2,851 cases…)" with no source attached — that case count is a local measurement of the **C# original**, and the migrated run has never reported a count at all, which is the whole point of the change. It is now labelled as such.

And **2,851 was itself wrong**. Running the C# original here reports:

```
Failed!  - Failed:     1, Passed:  2890, Skipped:     0, Total:  2891, Duration: 21 m 57 s - GSharp.Cs2Gs.Tests.dll (net10.0)
```

**2,891**, which is the number this PR's own description uses — the code comment had a transposed digit. Corrected in all three places it appears (`SdkCompileRunner` twice, `Issue3501ScaledTestParityBudgetTests` doc comment once); no assertion or expected value depends on it, and ≈ 0.45 s per case still holds.

### Verification

`dotnet test tools/cs2gs/Cs2Gs.Tests/Cs2Gs.Tests.csproj -c Debug`: **2,890 passed, 1 failed** — the single failure is `Issue2819_SameVersionPackedToolAndSdk_CompileTranslatedPatternVariable`, which needs a locally built Release `.nupkg` and is a pre-existing environment quirk, unrelated to this change. All #3501 and #3931 budget tests pass.

DavidObando:
Small correction to the note above, since this PR is specifically about getting provenance right.

`2,851` was **not** a transposed digit. It was an accurate local measurement of the `Cs2Gs.Tests` C# original taken on **2026-09-06**:

```
Failed! - Failed: 1, Passed: 2850, Skipped: 0, Total: 2851, Duration: 21 m 4 s
```

The suite gained 40 tests between then and the re-measurement on 2026-09-08, which reports `Total: 2891`. So the old figure was correct when it was taken and had simply gone stale.

The correction itself stands — 2,891 is the current number and the comment now carries its measurement date, which is exactly what stops this recurring. Only the stated *reason* was wrong, and it is worth fixing here because "the original was a typo" and "the original was right and aged" imply different things about whether the calibration record can be trusted.

No code change; the ≈ 0.45 s per executed case shape is unaffected either way.

Issue #3083, opened 2026-08-02
[cs2gs] with expression inside LINQ lambda fails G# round-trip parsing


DavidObando:
## Observed failure

Migrating `DavidObando/code-exploder` at `b4f6be83be664f02965f0363ec41f8946b56ff1d` with `cs2gs`/`gsc` `0.3.504+e8e6f7c8e8` fails translation of `src/CodeExploder.Pipeline/Generators.cs`.

C# source in `ArchitectureSynthesizer.Sanitize`:

```csharp
var scenarios = (doc.Scenarios ?? [])
    .Select(s => s with
    {
        Steps = s.Steps
            .Where(st => ids.Contains(st.From) && ids.Contains(st.To))
            .ToList(),
    })
    .Where(s => s.Steps.Count >= 2)
    .Take(4)
    .ToList();
```

cs2gs emits:

```gsharp
let scenarios = (doc.Scenarios ?? []ArchScenario{})
    .Select((s ArchScenario) -> s with {
        Steps = s.Steps
            .Where((st ScenarioStep) -> ids!!.Contains(st.From) && ids!!.Contains(st.To))
            .ToList()
    })
    .Where((s ArchScenario) -> s.Steps.Count >= 2)
    .Take(4)
    .ToList()
```

The emitted compilation unit does not round-trip through the G# parser:

```text
CS2GS-ROUNDTRIP
GS0005 Unexpected token <OpenParenthesisToken>, expected <OpenBraceToken>.
```

Fingerprint: `sha256:9dd020ef25f2221f1c4811eab57bfcf6dd9a0bf5103e1e5ea1b98c976bc5151b`.

## Expected behavior

A record/data-class `with` expression used as the expression body of a LINQ lambda must have an unambiguous canonical G# representation that round-trip parses. cs2gs should add grouping or use a block-bodied lambda if required by the grammar.

## How to verify the fix

1. Add a reduced cs2gs translator test for `.Select(x => x with { Value = ... }).Where(...)`.
2. Assert the emitted G# parses a second time and compiles.
3. Re-run:

```bash
dotnet out/bin/Debug/Cs2Gs.Cli/cs2gs.dll migrate \
  --corpus ~/GitHub/DavidObando/code-exploder \
  --app src/CodeExploder.Pipeline/CodeExploder.Pipeline.csproj \
  --out /tmp/code-exploder-gs \
  --artifacts /tmp/code-exploder-gs-runs \
  --config Debug
```

[...]

Issue #2496, opened 2026-07-19
cs2gs: nullable-oblivious argument forgiveness injects !! into expression-tree lambdas


DavidObando:
## Summary

In nullable-oblivious projects, cs2gs's call-argument forgiveness pass can wrap an entire lambda argument in G# `!!`. The printer then renders that assertion on the lambda body. When the imported parameter is `Expression<TDelegate>` (EF Core, `Queryable`, or another LINQ provider), gsc correctly rejects the generated expression tree with GS0473 because null-assertion nodes are not expression-tree-representable.

The same valid C# under nullable-enabled compilation translates without `!!` and passes. A lambda value itself is inherently non-null; argument-position nullable forgiveness must not classify the lambda's synthesized method symbol as a nullable forwarded value.

## Oahu evidence

Full migration with `gsc 0.3.152+b1402faeaa`:

- Canonical pipeline fingerprint: `sha256:183d8b9a47222ade6d8aae4449fd1cfcc8bb965fd034dc2363952e033d17fd2f`
- C# source: `src/Oahu.Data/BooksDbContext.cs:129` (and the remaining EF model selectors through line 205)
- Emitted G#: `corpus_Oahu.Data/BooksDbContext.gs:66-98`
- Raw compiler result: 33 occurrences of `GS0473: An expression-tree lambda may not contain a null-assertion operator (issue #2130).`

Examples:

```gsharp
modelBuilder.Entity[Book]().HasKey((e Book) -> e.Id!!)
modelBuilder.Entity[Book]().HasIndex((e Book) -> e.Asin!!).IsUnique()
modelBuilder.Entity[SeriesBook]().HasKey(
    (e SeriesBook) -> AnonymousType0{SeriesId: e.SeriesId, BookId: e.BookId}!!)
```

The SDK build task currently masks the locationless raw GS0473 lines as a single GS9998, but a diagnostic build shows every underlying error before `Failed.`. This issue concerns the illegal cs2gs output, not the secondary diagnostic-forwarding loss.

## Stable minimal repro

Nullable-oblivious C#:

```csharp
using Microsoft.EntityFrameworkCore;

public sealed class Item
{
    public int Id { get; set; }
}

public sealed class DataContext : DbContext
{
    protected override void OnModelCreating(ModelBuilder modelBuilder)
    {
        modelBuilder.Entity<Item>().HasKey(item => item.Id);
    }
}
```

Project settings:

```xml
<TargetFramework>net10.0</TargetFramework>
[...]

PR #4147, opened 2026-09-08
gsfmt: run the post-pass by default and bank the long-line win (ADR-0179 phase 7b)

DavidObando:
## The finding

Phase 6 shipped a formatter that wraps (#4093, #4094) and phase 7a shipped the
`--format` flag that runs it — but **nothing ever passed the flag**. No build
script, no gate job: `grep -rn -- '--format' build/*.sh` returns nothing, and
`PipelineOptions.FormatOutput` was a plain `bool` defaulting to `false`. The
migrated tree the #3501 ratchet measures was still produced entirely by
`GSharpPrinter`, so every long line the gate has ever counted was emitted by the
*old* printer and phases 6a/6b had bought the corpus nothing.

This is ADR-0179 phase 7b: flip the default on.

## Measured

Two whole-corpus `cs2gs migrate --translate-only` passes over the same commit
(3,905 `.gs` files), scored with `build/cs2gs-counters.sh`'s post-#4123 counter
in the character scale:

| | printer only | + gsfmt post-pass |
|---|---:|---:|
| lines >300 (reducible) | **557** | **1** |
| lines >300 (single-atom-bounded) | 36 | 36 |
| lines >300 (total) | 593 | 37 |
| files gsfmt refused to format | – | **0** |

The 557 by the outermost construct a formatter has to break — if-expression
145, call argument list 120, collection/object literal 112, string
concatenation 73, `&&`/`||`/`??`/`or`/`and` chain 67, single-argument call 24,
switch-arm pattern list 16 — **all go to zero**. The one survivor is a
304-character `description: "…"` whose literal is 279 characters: the
`indent + 8 + widest_atom` heuristic files it as reducible, but the line has no
legal break point in it.

`gsfmt --check` over all 3,905 formatted files exits 0, so ADR-0179 D4's
idempotence holds corpus-wide, and the fail-soft path never fired.

## Ratchets

`longLineCeiling` **570 → 10**. The file's ~3% convention does not survive the
scale change — 3% of 1 is not a margin — so the number is sized to the failure
mode instead: a reducible long line now means gsfmt failed on a file, and 10
allows a handful before the gate goes red. **A run near 10 is a formatter
defect to investigate, never a ceiling to raise.**

`liftedLocalCeiling` **30 → 31** and `nullAssertionCeiling` **11500 → 12100**,
and neither is a regression — not one `__local_` and not one `!!` was added.
`selfmig_code_grep` drops any line containing a `"` before counting (the #3937
undercount this repo already documents), so an occurrence sharing a line with a
string literal is invisible to it; wrapping moves tokens off those lines and
**unmasks** occurrences that were always there. Across the two trees the raw
counts are byte-identical — labels 30, `__local_` 67, `!!` 29,920 — while the
code-filtered views read 30 → 31 and 22,387 → 23,535. The single unmasked lift
is exact and checkable: `ControlFlowGraph.gs` held two `__local_` calls *and* a
`"finallyBody"` literal on one line, and wrapping puts
`__local_…_CollectLabels(root)` on its own quote-free line.
`nullAssertionCeiling` is a projection, stated rather than hidden, because
`--translate-only` does not run stage 2's `!!`-polish; **tighten it from the
first post-flip gate run.**

## Churn

13 `Cs2Gs.Tests` failures, all `Assert.Contains` on emitted layout — no
compile, ILVerify, round-trip or stdout-golden movement, which is what D4
predicts. They land in their own mechanical commit. Eleven are one-line updates
to the canonical spelling the repository's own `.gs` already carries from phase
8's `gsfmt -w` (`var f(int32) -> int32`, `func Gather(ok bool)(Items …)`,
`is {}`, `@assembly: X`, `text[1 .. ^1]`, `{source.Value!!}`); two read a
*line* where the member is now several lines and are re-expressed against the
member.

## Guard failure on the first revision, and the fix

The first revision of this PR went **2/8 on the hot-core translation guard**:
six apps failed to compile with 15 x `GS0227` "documentation comment is not
attached to a declaration". gsfmt was splitting the accessibility modifier off
`init` — `private`, a blank line, then `init(...)` — detaching the `///` block
above it. 36 declarations across 25 files of the migrated tree.

**The defect was not in the formatter.** `ConstructorDeclarationSyntax.Span`
started at `convenience`/`init` and excluded `AccessibilityModifier`, so a
declaration's span did not cover the modifier that introduces it. Two consumers
read `Span.Start` as "where this declaration begins" — `DocumentationAttacher`,
which requires a documented declaration to start on the line after the `///`
block, and the layout builder, which puts its member break at that same
position. The span now starts at
`AccessibilityModifier ?? ConvenienceModifier ?? InitKeyword`, which also fixes
attachment for hand-written G# the formatter never touches.

Scope was probed, not assumed: only two declaration nodes override `Span` by
hand, and `DeinitDeclarationSyntax` has no modifiers to lose. Formatting one
documented member per declaration keyword — `init` with each of
private/internal/public, `convenience init`, `func`, `prop`, `var`, `let`,
nested `class`/`struct`/`interface`/`enum`, `shared`, `event`, `deinit` —
splits on exactly the four `init` shapes and none of the others.

### The lesson, made into a test

`gsfmt --check` over all 3,905 migrated files was **clean** while six guarded
apps did not build. The broken output was a perfectly stable fixed point — it
was just the wrong one. Idempotence cannot see that; only compiling can.

`test/Formatting.Tests/DocCommentAttachmentTests.cs` therefore **binds and
emits** the formatted output and asserts no `GS0227` and no error, across all
13 shapes. It fails on 4 of them without the span fix and passes with it.

Re-formatting the whole corpus with the fix leaves **zero** lone-modifier lines
(was 36 across 25 files) and the measurement above is unchanged at 1 / 36 / 37.

## Deliberately not here

`RenderWrappable`/`RenderWrapped` are **not** deleted. They are still the layout
of record whenever the post-pass fails soft, and removing them re-lays every
printer assertion in `Cs2Gs.Tests`. ADR-0179's phase 7 is split into 7b (this)
and 7c (delete, plus the chain-argument nesting defect this measurement found:
when a member-chain link's argument list breaks, the argument nests against the
chain head rather than the link).

## Verified locally / not verified

- `Cs2Gs.Tests`: green after the churn commit — 2,891/2,892 in the full re-run,
  and the one remaining expectation verified by a targeted re-run after its fix
  (baseline before the flip was 2,892/2,892 green). Many of these translate →
  compile via the SDK → ILVerify → run the migrated G#, so formatted output is
  known to compile and execute.
- `Formatting.Tests`: 59 green. `gsfmt --check` clean over `src/` and
  `samples/`. `build/test-cs2gs-counters.sh` green.
- **Not run locally:** the PR guard and the nightly self-migration gate. The
  ceilings above are the numbers to watch on the first gate run.


DavidObando:
All three review findings addressed in `35e82afb`.

**1 — `--format` docs contradicted the flip.** Confirmed: `website/docs/tooling/cs2gs.md` said `--format` was "opt-in during ADR-0179 phase 7a" and `--no-format` "remains the default until phase 7b". This PR *is* phase 7b, so the public option table stated the opposite of what ships, and no test covers a docs page. Both rows updated; `--no-format` is now described as what it actually is — the A/B escape hatch that emits the same program with the printer's line breaks. I swept the rest of the page for the same claim and found none.

**2 + 3 — two semantic assertions pinned the formatter's layout.** Confirmed, and the mechanism is worth stating precisely: `Compact` collapses runs of whitespace to a single space but does not make whitespace *optional*, so `HasIndex( (item Item) -> item.Name )` and `.Select( (s ArchScenario) -> ( s with{ …` encoded spaces that exist only because a line break sat there. Both are semantic regression tests — #2496 about which arguments carry a `!!`, #3083 about the shape of a `with` expression — and #3083's own comment said the layout "is what the formatter now owns" while the assertion encoded it anyway.

Both now match through a `ContainsIgnoringWrap` helper that builds a regex from an *unwrapped* expectation: whitespace tolerated beside a bracket or comma, where a break can legally land, and **required** between two ordinary tokens. A blunt strip-all-whitespace match would have been simpler and wrong — it also accepts `itemItem`, trading these tests' brittleness for a hole in the thing they exist to discriminate.

Verified in both directions:

| case | result |
| --- | --- |
| current wrapped output | match |
| unwrapped rendering | match |
| differently reflowed | match |
| `(itemItem)` — missing space | no match |
| wrong argument | no match |
| spurious `!!` | no match |

and, the check that actually matters, injecting a bogus `!!` into an expectation still turns the test red — a layout-tolerant assertion that had stopped discriminating would be worse than the brittle one it replaced.

**Tests:** `Cs2Gs.Tests` **2892/2892**, `Formatting.Tests` **85/85**. The PR guard was not re-run — this commit touches only a docs page and two test files, and the previous commit's guard run is unaffected by it.

---

**One thing deliberately left out of scope, flagged rather than silently skipped.** The same coupling exists across the other re-laid test files, in the opposite direction: `Issue2521`, for example, moved from `{ source.Value!! }` to `{source.Value!!}` — baking in the *absence* of a wrap rather than its presence. Equally layout-dependent, equally scheduled to break.

That is a pre-existing class spanning roughly a dozen files, and it belongs with **phase 7c** — which deletes `RenderWrappable`/`RenderWrapped` and re-lays ~4,756 printer assertions, and is therefore the reflow that will actually trigger it. Converting them all here would obscure this PR under churn unrelated to the change being reviewed.

Issue #4129, opened 2026-09-08
selfmig: migrated test/Compiler.Tests reports 79 parity failures behind a single fingerprint — needs triage


DavidObando:
Measured on gate run [34232380469](https://github.com/DavidObando/gsharp/actions/runs/34232380469) (gsc `0.4.575+033c38dfb8`, PASSED at 52/56).

## The count is 79, not 1

The gate reports **one fingerprint** for this app. The completed run reports:

```
Failed! - Failed: 79, Passed: 5415, Skipped: 0, Total: 5494, Duration: 1 h
```

This is [#3853](https://github.com/DavidObando/gsharp/issues/3853) exactly — test-parity fingerprints do not discriminate root causes, so 79 failures across at least a dozen root causes collapse onto a single hash and the ledger reads as if one thing is wrong. **Do not size this work from the fingerprint count.**

Worth recording that the run *completes*: `MirroredTestRunTimeoutFor` already grants this app 90 minutes (`SdkCompileRunner.cs:790`), and it used one hour of it. This app is no longer timeout-blocked — [#4045](https://github.com/DavidObando/gsharp/issues/4045)'s truncation problem is resolved here and now only affects `tools/cs2gs/Cs2Gs.Tests`.

## Clustering

By error message, over the 79:

| count | error head |
|---:|---|
| 48 | `AggregateException : One or more errors occurred. (Object reference not set to an instance of an object.)` |
| 7 | `NullReferenceException : Object reference not set to an instance of an object.` |
| 5 | `Assert.Equal() Failure: Values differ` |
| 4 | `Assert.IsType() Failure: Value is not the exact type` |
| 3 | `compile failed` |
| 2 | `ArgumentException : Object of type 'System.Func`1[System.Int32]' cannot be converted to type 'System.Action'.` |
| 2 | `Assert.DoesNotContain() Failure: Sub-string found` |
| 1 each | `Assert.Collection`, `Assert.Contains`, `gsc failed`, `EventHandler` → `Action<object,EventArgs>`, `GS0157 Cannot find type ConstantNarrowingBase` |

By implicated test file:

| lines | file |
|---:|---|
| 90 | `Emit/Issue2891TryRegionFlowEmitTests.gs` |
| 54 | `Emit/Issue2906ExhaustiveSwitchReturnEmitTests.gs` |
| 9 | `Emit/Issue1489GenericAsyncIteratorEmitTests.gs` |
[...]

Issue #4127, opened 2026-09-08
cs2gs: Cs2Gs.Tests migrated parity — 137 failures on the first complete run


DavidObando:
## What this is

The first **complete** parity run of migrated `tools/cs2gs/Cs2Gs.Tests`. Until now the app was killed on the flat 10-minute mirrored-test budget, so its `[FAIL]` list was a truncation artifact and no failure count existed (that budget is fixed in #4125). This issue records what the run actually reports, so the app can be driven to green.

## The run

Local exact-app self-migration (whole-repo `cs2gs migrate --translate-only`, then `cs2gs validate --app tools/cs2gs/Cs2Gs.Tests/Cs2Gs.Tests.csproj`, Release, macOS, gsc on `15b79938`):

| stage | result |
|---|---|
| translate | PASS (56/56 apps) |
| compile | PASS (735 s) |
| ILVerify | PASS (0.7 s) |
| test-parity | **FAIL — `LIBRARY-TESTS-FAILED`** |

```
Failed!  - Failed: 137, Passed: 2754, Skipped: 0, Total: 2891, Duration: 8 m 20 s
```

Two things worth stating plainly:

* **No test was lost.** The migrated assembly executed **2,891** cases — exactly what the C# original runs locally (`Failed: 1, Passed: 2890, Total: 2891, 21 m 46 s`, the one failure being the known `Issue2819…` case that needs a locally built Release `.nupkg`). So this is a pure outcome-parity gap, not a discovery gap.
* **The migrated run is not slow.** 8 m 20 s in Release here versus 21 m for the C# original in Debug. The CI kill at 10 minutes was a runner-speed margin, not a hang.

## The 137 failures, clustered by error

| count | error |
|---|---|
| 50 | `System.NullReferenceException` |
| 22 | `Assert.Contains() Failure: Sub-string not found` |
| 22 | `System.IO.FileNotFoundException: Could not find file …` |
| 10 | `Assert.Equal() Failure: Values differ` |
| 7 | `System.ArgumentException: GenericArguments[0], 'System.Int32', on 'System.Text.Json.Serialization…'` |
| 5 | `FileNotFoundException: Project file not found …` |
| 5 | corpus fixture not locatable (`L1-Console`, `L4-Console`, missing fixture dirs) |
| 4 | `missing fixture …` |
| 3 | `Assert.DoesNotContain() Failure: Sub-string found` |
| 2 | `Translated G# must bind without errors` |
[...]

PR #4155, opened 2026-09-09
cs2gs: avoid a gsc pattern-match gap that erased delegate identity under self-hosting (#4045 follow-up)

### LINKED ISSUE #4153: gsc: property pattern with an enum sub-pattern silently fails to match an imported interface-typed scrutinee
## What this is

Root-caused while fixing the residual nominal-delegate-identity defect (#4045 follow-up, tracked by #4127/#4129, umbrella #3501). This is a **gsc compiler bug**, not a cs2gs translation bug — the fix for the delegate-identity symptom lives entirely in cs2gs's own source (avoiding the pattern shape below), but the underlying defect is real, general, and needs a gsc-level fix independent of cs2gs.

## The bug

A property pattern with an **enum-constant sub-pattern**, matched against a scrutinee whose **static type is an imported interface**, incorrectly evaluates as non-matching — even when the type test and every individual sub-property genuinely hold.

Minimal repro (`.gs`, compiled and run with the local `gsc.dll`):

```
import Microsoft.CodeAnalysis

func OnePropOnEnumOnly(left ITypeSymbol) string {
    if left is INamedTypeSymbol { TypeKind: TypeKind.Delegate } leftDelegate {
        return "match:" + leftDelegate.Name
    }
    return "no-match"
}
```

Called with a real `Predicate<int32>` `INamedTypeSymbol` (obtained from a live `Microsoft.CodeAnalysis.CSharp` `CSharpCompilation`, passed in as a value statically typed `ITypeSymbol`): **returns `"no-match"`**, even though `left is INamedTypeSymbol` is true and `((INamedTypeSymbol)left).TypeKind == TypeKind.Delegate` is true.

The **decomposed equivalent works correctly**:

```
func TwoPropSeparate(left ITypeSymbol) string {
    if left is INamedTypeSymbol leftNamed && leftNamed.TypeKind == TypeKind.Delegate && leftNamed.DelegateInvokeMethod != nil {
        return "match:" + leftNamed.Name
    }
    return "no-match"
}
```
→ `"match:Predicate"`, correctly.

A property pattern combining an enum sub-pattern with a **null-check** sub-pattern only (no enum) works fine:

```
func OnePropOnNullCheckOnly(left ITypeSymbol) string {
    if left is INamedTypeSymbol { DelegateInvokeMethod: not nil } leftDelegate {
        return "match:" + leftDelegate.Name
    }
    return "no-match"
}
```
→ matches correctly. So the defect is specific to the **enum-constant** sub-pattern, not property patterns or null-check sub-patterns in general.

And the identical shape against a **source-declared** class/enum/interface (not imported) works fine:

```
enum AnimalKind { Canine, Feline }
interface IAnimal2 { }
class Dog2 : IAnimal2 { public var Kind AnimalKind = AnimalKind.Canine }

func EnumPropSourceType(a IAnimal2) string {
    if a is Dog2 { Kind: AnimalKind.Canine } d {
        return "match:" + d.Kind.ToString()
    }
    return "no-match"
}
```
→ matches correctly.

So the defect requires **all** of: (1) a property pattern with an enum-constant sub-pattern, (2) matched against a value whose static type is an interface, (3) where the named type and/or the enum are **imported/metadata** symbols (not source-declared). The three isolated conditions each individually work; only the combination breaks.

## How this was found

cs2gs's own `CSharpTypeMapper.cs` (issue #3841's collision-detection logic, and issue #2835's `MapEventType`) uses exactly this pattern shape against `Microsoft.CodeAnalysis.ITypeSymbol`/`INamedTypeSymbol`/`TypeKind`. Those call sites work correctly when compiled and run as ordinary C# (via `csc`) — all of cs2gs's own C# unit tests for #3841/#4045 pass natively. But cs2gs's self-migration effort (#3501) translates cs2gs's own source to G# and compiles it with `gsc`, and in that self-hosted build the same logic silently stops detecting the colliding `Predicate<T>`/`Func<T, bool>` overload set (`GS0264` in the emitted program) and silently stops preserving `EventHandler`'s nominal identity in locals — because the translated G# is a faithful, semantically-identical rendering of the C#, and gsc mis-evaluates it. Full diagnosis, including a self-hosted end-to-end reproduction (translate `CSharpTypeMapper.cs` with the production translator, compile the result with `gsc`, run it against a real Roslyn compilation, observe the collision silently going undetected) is in the companion cs2gs PR (link added once opened).

## Other call sites that use the same shape and have NOT been individually verified

These all combine an enum sub-pattern with matching against a metadata/interface-typed scrutinee, in `tools/cs2gs/Cs2Gs.Translator/`, and are worth auditing once this is fixed (or worth converting defensively regardless, since cs2gs self-hosts):

- `CSharpToGSharpTranslator.Analyzers.cs:56`
- `CSharpToGSharpTranslator.Declarations.cs:531,1033`
- `CSharpToGSharpTranslator.Invocations.cs:299,347,2329,2390,3615,3724,3725`
- `CSharpToGSharpTranslator.Expressions.cs:105,3490`
- `CSharpToGSharpTranslator.Patterns.cs:2567`
- `CSharpToGSharpTranslator.Members.cs:582`
- `CSharpToGSharpTranslator.Nullability.cs:756`
- `CSharpToGSharpTranslator.Types.cs:1137`
- `ObliviousNullabilityAnalyzer.cs:2312`

Some of these may be benign (e.g. if the scrutinee is never actually an imported/metadata symbol in practice), but each is a candidate for the same silent-non-match failure mode under self-hosting.

## Suggested next step

Find gsc's property-pattern lowering/binder path for a nested pattern combining a type test, an enum-constant sub-pattern, and (optionally) other sub-patterns, when the scrutinee's static type is an interface. Compare against the working decomposed lowering (plain `is T name` + separate `==`/`!=` comparisons) to find where the combined form diverges — likely in how the constant sub-pattern's comparison is emitted/evaluated relative to the type-test's narrowing, or in how metadata enum constants are resolved within a nested pattern context.

DavidObando:
## Summary

- PR #4110 preserved nominal delegate identity at several call sites, but three of its own regression tests still failed under the self-migration gate (#4127, #4129) even though they pass natively as C#. Root-caused: it is not a cs2gs translation defect. cs2gs's own C# at these sites is a faithful, semantically correct rendering of the intended logic — but once cs2gs translates its own source and gsc compiles the result (self-hosting, #3501), a gsc pattern-matcher bug (filed as #4153) silently mis-evaluates a specific pattern shape, and every explicit-delegate-identity check that used that shape silently degrades to structural erasure.
- Fixes it by rewriting the four affected call sites in `CSharpTypeMapper.cs` (`MapEventType`, `AddIfDistinctDelegates`, `TypesEraseAlike`) and `CSharpToGSharpTranslator.Constructors.cs` (`explicitlyNamedDelegate`) from a nested property pattern (`is [not] INamedTypeSymbol { TypeKind: TypeKind.Delegate, DelegateInvokeMethod: not null } name`) to the semantically identical decomposed form (`is INamedTypeSymbol name && name.TypeKind == TypeKind.Delegate && name.DelegateInvokeMethod != null`), which survives self-hosting.
- Files #4153 for the underlying gsc compiler defect (out of scope here), with a minimal repro and a list of ~9 other call sites in cs2gs using the same shape that have not been individually verified.

## The gsc bug (#4153), in one line

A property pattern combining an **enum-constant sub-pattern** with a scrutinee whose **static type is an imported interface** silently evaluates as non-matching, even when the type test and every sub-property individually hold. The decomposed equivalent (designated `is` narrowing + plain `==`/`!=` comparisons) works correctly. Full repro and analysis in #4153.

## Verification

- New regression tests in `tools/cs2gs/Cs2Gs.Tests/Issue4045SelfHostedPatternMatchRegressionTests.cs`:
  - `TranslatedOwnSource_NoLongerEmitsTheNestedEnumPatternAtTheFixedSites` — translates cs2gs's own `CSharpTypeMapper.cs`/`CSharpToGSharpTranslator.Constructors.cs` with the production translator and asserts the emitted G# no longer contains the buggy shape at the four fixed sites. **This is the test that discriminates the fix** — confirmed red (fails) with the production hunks reverted, green with them applied.
  - `SelfHostedCollectIdentityCriticalDelegates_DetectsThePredicateFuncCollision` — a hand-kept decomposed mirror of the fixed logic, compiled by the real `gsc.dll` and run against a real Roslyn compilation containing the exact `Add(Predicate<int>)`/`Add(Func<int, bool>)` collision from #3841, proving the decomposed shape actually detects the collision once compiled by gsc (does not discriminate this diff by itself, since it is a hand-kept mirror rather than a translation of the production file — documented in its own doc comment).
- `dotnet test tools/cs2gs/Cs2Gs.Tests/Cs2Gs.Tests.csproj -c Debug`: 2,904/2,905 passed (before and after this change — the single failure is the pre-existing `Issue2819…` case, which needs a locally built Release `.nupkg`, unrelated and unaffected).
- `dotnet test test/Compiler.Tests/Compiler.Tests.csproj -c Debug`: 5,793/5,796 passed (3 known pre-existing failures — `NETStandard.Library.Ref` targeting pack missing on this machine).
- Targeted issue tests (`Issue3841DelegateOverloadSetTranslationTests`, `Issue4045CompilerTestsParityRegressionTests`, `Issue2835NominalDelegateTypeTranslationTests`) all still pass natively — expected, since this is a self-hosting-only bug and these tests never exercise gsc on cs2gs's own source.

Note on scope: this fix resolves the three specifically-cited symptoms (GS0264 overload collision, lost `Predicate<int32>` constructor-parameter identity, lost `EventHandler` local identity) by removing the ONE gsc-mishandled shape from their exact call path. It does not, and cannot, fix #4153 itself, nor the ~9 other untouched call sites using the same shape — those remain tracked in #4153 for a gsc-level fix.

### DIFF
--- a/tools/cs2gs/Cs2Gs.Tests/Issue4045SelfHostedPatternMatchRegressionTests.cs
+++ b/tools/cs2gs/Cs2Gs.Tests/Issue4045SelfHostedPatternMatchRegressionTests.cs
@@ -1,0 +1,453 @@
+// <copyright file="Issue4045SelfHostedPatternMatchRegressionTests.cs" company="GSharp">
+// Copyright (C) GSharp Authors. All rights reserved.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Cs2Gs.CodeModel.Ast;
+using Cs2Gs.CodeModel.Printing;
+using Cs2Gs.Translator;
+using Cs2Gs.Translator.Loading;
+using Xunit;
+
+namespace Cs2Gs.Tests;
+
+/// <summary>
+/// Issue #4045 follow-up (tracked by #4127/#4129, umbrella #3501): PR #4110
+/// preserved nominal delegate identity at declaration/expression sites, but
+/// three of its own regression tests — <c>Issue3841DelegateOverloadSetTranslationTests</c>
+/// and <c>Issue4045CompilerTestsParityRegressionTests</c> — still failed
+/// under the self-migration gate (gate run 34300556173) even though they
+/// pass when run natively as C#. The root cause is NOT a cs2gs translation
+/// defect: cs2gs's own C# is a faithful, semantically correct rendering of
+/// the underlying logic. It is a gsc compiler defect (filed as #4153): a
+/// property pattern combining an ENUM-constant sub-pattern with a scrutinee
+/// whose static type is an imported interface silently evaluates as
+/// non-matching, even when the type test and every sub-property individually
+/// hold. <c>CSharpTypeMapper.MapEventType</c> (issue #2835),
+/// <c>AddIfDistinctDelegates</c> and <c>TypesEraseAlike</c> (issue #3841),
+/// and <c>CSharpToGSharpTranslator.Constructors.MapParameter</c>'s
+/// <c>explicitlyNamedDelegate</c> check all used exactly this shape
+/// (<c>is [not] INamedTypeSymbol { TypeKind: TypeKind.Delegate, ... } name</c>)
+/// against <c>Microsoft.CodeAnalysis.ITypeSymbol</c> — correct C#, but once
+/// cs2gs translates its OWN source and gsc compiles the result (self-hosting,
+/// #3501), the collision-detection set this logic builds comes back empty
+/// and every explicit-delegate-identity site silently degrades to structural
+/// erasure: the exact shape of all three symptoms (GS0264 overload collision,
+/// a lost <c>Predicate&lt;int32&gt;</c> constructor parameter identity, a
+/// lost <c>EventHandler</c> local identity).
+/// <para>
+/// The fix is a source-level rewrite of those four call sites to the
+/// DECOMPOSED equivalent (a designated <c>is</c> narrowing followed by plain
+/// <c>==</c>/<c>!=</c> comparisons) — semantically identical C#, proven by
+/// the self-hosted regression test below to survive translation + gsc
+/// compilation. It is a workaround for cs2gs's own source, not a fix for
+/// #4153 itself (which needs a gsc-level fix and remains open, tracking the
+/// ~9 other call sites in this project using the same shape that have not
+/// individually been verified).
+/// </para>
+/// </summary>
+public sealed class Issue4045SelfHostedPatternMatchRegressionTests
+{
+    /// <summary>
+    /// Translates cs2gs's own <c>CSharpTypeMapper.cs</c> and
+    /// <c>CSharpToGSharpTranslator.Constructors.cs</c> with the production
+    /// translator (exactly what self-migration does) and asserts the emitted
+    /// G# no longer contains the nested-pattern shape gsc mis-evaluates
+    /// (issue #4153) at the four fixed call sites. This is the cheap,
+    /// syntactic guard: it does not require compiling/running the emitted G#
+    /// (see <see cref="SelfHostedCollectIdentityCriticalDelegates_DetectsThePredicateFuncCollision"/>
+    /// for the full functional proof), so it stays fast and catches anyone
+    /// reintroducing the collapsed shape at these sites.
+    /// </summary>
+    [Fact]
+    public async Task TranslatedOwnSource_NoLongerEmitsTheNestedEnumPatternAtTheFixedSites()
+    {
+        string typeMapperGs = await TranslateOwnFile(
+            "Cs2Gs.Translator", "CSharpTypeMapper.cs");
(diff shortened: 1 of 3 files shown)


DavidObando:
Fixed in `2001e341`.

Confirmed against #4153's own text: `OnePropOnEnumOnly` reproduces with the enum sub-pattern alone, no null check involved. The three fixed-site comments (`MapEventType`, `TypesEraseAlike`, `Constructors.cs`'s `explicitlyNamedDelegate` check) described the trigger as "enum + null-check combined" — not wrong about these specific three sites, since each happens to combine both, but stated as *the* trigger it risks a future audit reading an enum-only pattern as unaffected when #4153 already showed otherwise. All three now say the enum sub-pattern alone is sufficient, with the null check called out as incidental rather than load-bearing.

Comment-only change; no production logic moved.

Issue #4180, opened 2026-09-10
selfmig: Type.FullName throws on a gsc-emitted iterator state-machine's reified interface (AssertImplementsGenericOf)


DavidObando:
Split out of #4129's clustering work. This is Cluster B: **7 of #4129's 79-88 test-parity failures**, all sharing one call site.

## Shared call site

`test/Compiler.Tests/Emit/Issue1489GenericAsyncIteratorEmitTests.cs:465`, `AssertImplementsGenericOf(Type smType, Type openInterface, Func<Type[], bool> argsMatch)`:

```csharp
var ok = smType.GetInterfaces().Any(i =>
    i.IsGenericType
    && i.GetGenericTypeDefinition() == openInterface
    && argsMatch(i.GetGenericArguments()));
Assert.True(
    ok,
    $"State machine '{smType.FullName}' must implement the expected reified " +
    $"{openInterface.Name} row. Actual interfaces: " +
    string.Join(", ", smType.GetInterfaces().Select(i => i.FullName)));
```

`Assert.True(bool, string message)` takes a plain `string`, so the message — including `smType.GetInterfaces().Select(i => i.FullName)` — is evaluated **eagerly on every call**, regardless of whether `ok` is true. The observed failure is a bare `System.NullReferenceException`, and the stack trace shows the crash inside the `i => i.FullName` selector lambda itself (invoked from `ArraySelectIterator.MoveNext()` during `String.Join`), not inside `argsMatch` or the `ok` computation.

`Type.FullName` normally returns `null` (not throw) for open generic / unresolved-parameter types, so a throw here is unusual — it smells like malformed generic metadata on one of `smType`'s reflected interfaces, where `smType` is a state-machine type emitted by gsc for an async iterator / iterator method under self-hosted compilation.

## Failing cases (from gate run 34370082359, gsc 0.4.604+e2913d9b30)

[...]

Issue #4179, opened 2026-09-10
selfmig: Compiler.Tests emitted <Main>$ NREs at runtime for exhaustive-switch/try-region-flow snippets under self-hosted gsc


DavidObando:
Split out of #4129's clustering work. This is Cluster A: **48 of #4129's 79-88 test-parity failures**, all sharing one call site.

## Shared call site

`test/Compiler.Tests/Emit/Issue2891TryRegionFlowEmitTests.cs:608`, `CompileVerifyLoadAndRun(string name, string source)`:

```csharp
var execution = Task.Run(
    () => entry.Invoke(null, entry.GetParameters().Length == 0 ? null : new object[] { Array.Empty<string>() }));
Assert.True(execution.Wait(TimeSpan.FromSeconds(10)), $"{name} execution timed out");
```

By the time this runs, `result.Success` (compile) and `IlVerifier.Verify(assemblyPath)` have already passed — the snippet compiles cleanly and verifies. The failure is `System.AggregateException : One or more errors occurred. (Object reference not set to an instance of an object.)`, thrown by `.Wait()` because the task delegate (`entry.Invoke`, i.e. the emitted `<Main>$` of the *snippet under test*) NREs when actually executed.

## Failing cases (from gate run 34370082359, gsc 0.4.604+e2913d9b30)

All 48 are theory rows of 3 test methods in `Issue2891TryRegionFlowEmitTests.gs` / `Issue2906ExhaustiveSwitchReturnEmitTests.gs` (both use this same helper pattern):
- `AcceptedFiniteShape_VerifiesLoadsAndRuns` — 30 rows
- `ExhaustiveSwitch_VerifiesLoadsAndRuns` — 11 rows
- `ExhaustiveSwitchStatement_UnmatchedValueFallsThrough` — 5 rows
- `ExhaustiveEnum_UnnamedValueWithFixedReturn_ThrowsDefensively`, `ExhaustiveEnum_UnnamedRuntimeValue_ThrowsDefensively` — 1 row each

All are exhaustive-switch-statement / try-finally-goto-return shaped snippets (see `Case(...)`/`OutputCase(...)` calls near these tests for the exact source bodies).

## Framing

[...]

Issue #4112, opened 2026-09-08
selfmig: migrated Compiler.Tests exhaustive-switch emit paths throw NRE (18 parity failures)


DavidObando:
Split from #4045 after a complete migrated `test/Compiler.Tests` run (5,393 passed / 75 failed / 5,468 total).

Eighteen failures are in `Issue2906ExhaustiveSwitchReturnEmitTests`, covering exhaustive enum/sealed-type switches, fallthrough, nested try/select shapes, and defensive unmatched-value paths. The migrated test harness reaches execution but the compile/verify/run task faults with `NullReferenceException`.

Reproduce by migrating `test/Compiler.Tests`, then run the `Issue2906ExhaustiveSwitchReturnEmitTests` class. Root-cause the migrated compiler path and add executing + ILVerify coverage.

Part of #3501. Split from #4045.

DavidObando:
These 18 failures are the exact same bug as #4179 (CLOSED), and #4179's own body names them individually. Its "Failing cases" list enumerates 48 rows across **two** test classes that share one helper:

```
AcceptedFiniteShape_VerifiesLoadsAndRuns                      - 30 rows   <- #4111
ExhaustiveSwitch_VerifiesLoadsAndRuns                         - 11 rows   \
ExhaustiveSwitchStatement_UnmatchedValueFallsThrough          -  5 rows    |  = 18 = THIS ISSUE
ExhaustiveEnum_UnnamedRuntimeValue_ThrowsDefensively          -  1 row     |
ExhaustiveEnum_UnnamedValueWithFixedReturn_ThrowsDefensively  -  1 row    /
```

30 + 18 = 48, and #4179 states plainly that all 48 "are theory rows of 3 test methods in `Issue2891TryRegionFlowEmitTests.gs` / `Issue2906ExhaustiveSwitchReturnEmitTests.gs` (both use this same helper pattern)". #4111 (the 30) was already closed as this duplicate; these 18 are the remainder of the same cluster. All four issues were measured on the same pre-fix run (5,393 passed / 75 failed / 5,468 total), which predates #4182.

## The shared line is byte-identical

`test/Compiler.Tests/Emit/Issue2906ExhaustiveSwitchReturnEmitTests.cs:364-366`
`test/Compiler.Tests/Emit/Issue2891TryRegionFlowEmitTests.cs:644-646` (#4179 cites the helper at `:608`)

```csharp
var execution = Task.Run(
[...]

Issue #4111, opened 2026-09-08
selfmig: migrated Compiler.Tests try-region emit paths throw NRE (30 parity failures)


DavidObando:
Split from #4045 after a complete migrated `test/Compiler.Tests` run (5,393 passed / 75 failed / 5,468 total).

Thirty failures are all `Issue2891TryRegionFlowEmitTests.AcceptedFiniteShape_VerifiesLoadsAndRuns` cases. The migrated test harness reaches execution, then its async compile/run wrapper faults with `NullReferenceException`. Cases span returns, breaks, continues, gotos, catches, rethrows, and finally overrides, so this is a runtime/compiler-emission family rather than one test assertion.

Reproduce by migrating `test/Compiler.Tests`, then run the `Issue2891TryRegionFlowEmitTests` class. Determine the first migrated-only null dereference and add executing + ILVerify coverage.

Part of #3501. Split from #4045.

DavidObando:
This issue's 30 failures (all \`Issue2891TryRegionFlowEmitTests.AcceptedFiniteShape_VerifiesLoadsAndRuns\` rows) are the exact same bug as #4179 (CLOSED) -- specifically, #4179's own "AcceptedFiniteShape_VerifiesLoadsAndRuns -- 30 rows" line item, against the identical \`CompileVerifyLoadAndRun\` helper line. #4179 was root-caused and fixed by #4182 (merged as \`9cafdbd9\`, already on \`main\`): cs2gs's \`IsUnguardedForwardOfTaintedValueAsRuntimeLambdaResult\` heuristic force-asserted the reflection-nullable \`MethodInfo.Invoke(...)\` result with \`!!\` inside an unobserved \`Task.Run\` lambda, throwing a NullReferenceException the instant the void \`<Main>$\` legitimately returned null -- exactly this issue's "async compile/run wrapper faults with NullReferenceException" symptom. #4182's own investigation (a self-hosted gsc/Core.dll running every snippet body from the affected test files directly) disproves the "compiler-emission family" framing in this issue's title: every snippet ran correctly, and the failure is 100% attributable to the shared harness line, independent of try-region shape.

I re-verified independently: a fresh whole-repo \`cs2gs migrate --translate-only\` on top of latest \`main\` shows the migrated \`Issue2891TryRegionFlowEmitTests.gs\`'s \`entry.Invoke(...)\` carries no \`!!\`, and reverting #4182's fix reproduces the exact bug (verified via anti-vacuity test run).

[...]

PR #4202, opened 2026-09-11
cs2gs: record that #4180's fix also closes #4115's 6 selfmig iterator NRE rows

### LINKED ISSUE #4115: selfmig: generic iterator state-machine metadata inspection throws NRE (6 parity failures)
Split from #4045 after complete parity measurement. Six failures across `Issue1481MapFunctionIteratorEmitTests`, `Issue1489GenericAsyncIteratorEmitTests`, and `Issue813TupleSequenceReturnEmitTests` throw `NullReferenceException` while inspecting the generated state-machine interfaces for tuple, function, map, and generic element types.

Reproduce against the migrated compiler and identify whether generic interface type arguments are missing or malformed before the reflection assertion. Add sync/async iterator ILVerify and runtime metadata coverage.

Part of #3501. Split from #4045.

DavidObando:
## Summary

Closes the paper trail for #4115. **There is no defect left to fix** — #4115's 6 test-parity failures are the exact same rows as #4180 (already fixed by #4181, commit `964a50a6`, already on `main`). This PR is a comment-only amendment; no production code and no test behaviour changes.

## Investigation

Per the issue's own instructions, I first checked whether #4115 shared #4111/#4112's root cause (the #4179/#4182 unobserved-`Task.Run` `!!` heuristic on `CompileVerifyLoadAndRun`). It does not: none of `Issue1481MapFunctionIteratorEmitTests.cs`, `Issue1489GenericAsyncIteratorEmitTests.cs`, or `Issue813TupleSequenceReturnEmitTests.cs` use `CompileVerifyLoadAndRun` or an unobserved `Task.Run` — they compile via a direct in-process `Program.Main` call and reflect on `smType.GetInterfaces()` directly.

Reading those three files closely surfaced a sharper match: they are **exactly** the three files and the same shape of failure (`NullReferenceException` "while inspecting the generated state-machine interfaces") already described and fixed by #4180/#4181:

- cs2gs's `LambdaResultFlowsToNullableSink` only recognised a `.Select(...)` result that chains into a further scalar-returning call (e.g. `.Select(...).FirstOrDefault()`). It didn't recognise the sibling shape where the `Select` result flows **whole, as a collection**, into an argument position whose parameter itself accepts a nullable-element sequence — exactly `string.Join(string?, IEnumerable<string?>)`, which every one of these tests' `AssertImplementsGenericOf`/`AssertImplementsEnumerableOf` helpers builds as their `Assert.True` failure message: `string.Join(", ", smType.GetInterfaces().Select(i => i.FullName))`.
- Falling through to the general oblivious-forwarding bridge inserted a G# `!!` on `i.FullName`. `Type.FullName` legitimately returns `null` for a reified generic interface row closed over an unresolved type parameter (`IEnumerable<Dictionary<string,T>>` closed over the SM's own `Var(0)`) — exactly the shape 6 of these 7 reflection tests construct. The 7th, `StateMachineClass_MapOfListElement_StaysErased_AndVerifies`, asserts a fully-closed, erased row (`Dictionary<string, List<object>>`) whose `FullName` is never null — which is why it was never among the 6 failures.
- This is **not a gsc emitter defect**: #4180's investigation (and the assertion messages themselves, which only evaluate lazily-necessary on failure) confirm the emitted state-machine interface metadata is correct under both native and self-hosted gsc. It is purely a self-hosting **translation-fidelity** issue in the test's own reflection code.

## Why no new production fix

The fix for this exact defect (`CSharpToGSharpTranslator.Expressions.cs`'s `LambdaResultFlowsToNullableSink` unwrapping a sink's single generic type argument and comparing element-to-element) already merged in #4181 (commit `964a50a6`), which is already on `main` in this worktree. What was missing is provenance connecting #4115 (split from #4045) to that fix (which was filed as #4180, split from #4129) — the same gap #4199 closed for #4112 against #4179/#4182. This PR adds that note to `tools/cs2gs/Cs2Gs.Tests/Issue4180NullableSelectResultRegressionTests.cs`, in the same shape.

I also checked `tools/cs2gs/selfmig-test-allowlist.json` — none of these test names or #4115/#4180/#4129 appear there, so there is no allow-list entry to remove.

## Verification

**Green, on current `main` (already past `964a50a6`):**

```
dotnet test tools/cs2gs/Cs2Gs.Tests/Cs2Gs.Tests.csproj -c Release --filter "FullyQualifiedName~Issue4180"
Passed! - Failed: 0, Passed: 4, Skipped: 0, Total: 4

dotnet test test/Compiler.Tests/Compiler.Tests.csproj -c Release --filter "FullyQualifiedName~Issue1481MapFunctionIteratorEmitTests|FullyQualifiedName~Issue1489GenericAsyncIteratorEmitTests|FullyQualifiedName~Issue813TupleSequenceReturnEmitTests"
Passed! - Failed: 0, Passed: 25, Skipped: 0, Total: 25
```

(Native tests were never affected — native gsc never hits this translator-only bug, matching #4180's own documentation of re-translating these exact three files with no `!!` left on any `i.FullName` selector.)

**Red, by anti-vacuity** (reverted just #4180/#4181's translator hunk in `CSharpToGSharpTranslator.Expressions.cs`, comparing the commit before and after `964a50a6`):

```
dotnet test tools/cs2gs/Cs2Gs.Tests/Cs2Gs.Tests.csproj -c Release --filter "FullyQualifiedName~Issue4180"
Failed! - Failed: 4, Passed: 0, Total: 4

Found:  "t.FullName!!"

System.NullReferenceException: Object reference not set to an instance of an object.
   at Default.Probe.<lambda_host_<lambda1>_1>.Invoke(Type t)
   at System.Linq.Enumerable.ArraySelectIterator`2.MoveNext()
   at System.String.Join(String separator, IEnumerable`1 values)
```

— reproducing the exact reported crash shape and stack trace. Restored the fix afterward and re-confirmed green (4/4).

Part of #3501. Fixes #4115.

### DIFF
--- a/tools/cs2gs/Cs2Gs.Tests/Issue4180NullableSelectResultRegressionTests.cs
+++ b/tools/cs2gs/Cs2Gs.Tests/Issue4180NullableSelectResultRegressionTests.cs
@@ -52,6 +52,29 @@ namespace Cs2Gs.Tests;
 /// matching when that element is already nullable-ANNOTATED (e.g. the
 /// BCL-declared <c>IEnumerable&lt;string?&gt;</c> <c>string.Join</c> accepts).
 /// </para>
+/// <para>
+/// Issue #4115 ("selfmig: generic iterator state-machine metadata
+/// inspection throws NRE (6 parity failures)", split from #4045) is the
+/// SAME six rows as #4180 (split from #4129) — both parity measurements
+/// independently surfaced the identical
+/// <c>StateMachine_BareElement_IsGeneric_ImplementsAsyncEnumerableOfVar0</c>,
+/// <c>StateMachine_MapElement_ImplementsAsyncEnumerableOfReifiedDictionary</c>,
+/// <c>StateMachine_FunctionElement_ImplementsAsyncEnumerableOfReifiedFunc</c>,
+/// <c>StateMachineClass_MapElement_ImplementsIEnumerableOfReifiedDictionary</c>,
+/// <c>StateMachineClass_FunctionElement_ImplementsIEnumerableOfReifiedFunc</c>,
+/// and <c>StateMachineClass_TupleElement_ImplementsIEnumerableOfValueTuple</c>
+/// failures, in the same three files, with the same
+/// <c>NullReferenceException</c>-inside-<c>.Select(i =&gt; i.FullName)</c>
+/// signature. There is no defect left to fix for #4115 either — this class's
+/// two <c>!!</c>-on-<c>FullName</c> tests already pin the exact translator
+/// defect, and <c>test/Compiler.Tests</c>' 25 native tests across
+/// <c>Issue1481MapFunctionIteratorEmitTests</c>,
+/// <c>Issue1489GenericAsyncIteratorEmitTests</c>, and
+/// <c>Issue813TupleSequenceReturnEmitTests</c> were never affected (native
+/// gsc never hits this bug — it is purely a self-hosting translation-fidelity
+/// issue). This provenance note is the same shape #4199 used to record that
+/// #4195's regression test also pinned #4112's rows.
+/// </para>
 /// </summary>
 public sealed class Issue4180NullableSelectResultRegressionTests
 {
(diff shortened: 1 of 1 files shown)


DavidObando:
Addressed in 3903cf75: distinguished the six `AssertImplementsGenericOf`/`AssertImplementsEnumerableOf` helper-based reflection tests from the seventh test in `Issue813TupleSequenceReturnEmitTests.cs` (`StateMachineClass_TupleElement_ImplementsIEnumerableOfValueTuple`), which performs the equivalent interface scan inline rather than through either shared helper. Re-verified green: `dotnet test tools/cs2gs/Cs2Gs.Tests/Cs2Gs.Tests.csproj -c Release --filter "FullyQualifiedName~Issue4180"` -> 4/4 pass.

Issue #1272, opened 2026-06-27
cs2gs translator emits System.GC.AllocateArray[T](n) for new T[n]


DavidObando:
This should be replaced for the G# `[n]T` form.

PR #4257, opened 2026-09-15
Accept the no-initializer zero-length attribute array constant (#4247)

### LINKED ISSUE #4247: Attribute array constants reject zero-length allocation without an initializer
Found while expanding #4245 review controls; separate from boxed enum-array metadata identity. Roslyn accepts [Value(new Kind[0])] for an object-valued attribute constructor, but translated @Value([0]Kind) is rejected with GS0202. The equivalent explicit empty initializer new Kind[] { } translates and executes successfully after the scoped serialization fix.

The attribute-array binder requires syntax.Elements to be present and rejects the zero-length allocation before serialization. The CLR blob has an ordinary zero-length SZARRAY representation. Add a bounded zero-length allocation constant control and preserve rejection of nonconstant lengths/nonserializable elements. This syntax/binding gap is not hidden by weakening #4245 assertions; that PR tests both empty and nonempty explicitly initialized enum arrays.

DavidObando:
Fixes #4247

## Summary

- `[N]T` written with **no initializer at all** (not even empty braces `{}`) parses as the runtime/zero-initialised allocation form (issue #1272): its length lives in `ArrayCreationExpressionSyntax.LengthExpression`, and `Elements` stays `null`. This is a different syntax shape from `[]T{}` / `[N]T{}` (an explicit, possibly-empty initializer), which DOES populate `Elements` — just with zero items.
- `DeclarationBinder.TryBindAttributeArrayArgument` unconditionally required `syntax.Elements` to be present, so the constant, zero-length spelling `[0]Kind` was rejected with `GS0202` even though Roslyn accepts the equivalent `new Kind[0]`, and G#'s own explicit-empty-initializer spellings already worked.
- Accept the no-initializer form once its length resolves to the literal constant `0` — a zero-length SZARRAY is exactly as constant and serialisable as an explicit empty initializer, and the CLR blob it produces is identical either way. Any other runtime length (a variable, or a nonzero constant with no initializer to supply its elements) still has no constant element list to serialize and correctly keeps reporting `GS0202` — the fix is scoped to the zero-length case, not a general default-value-filled allocation.

## Verification

- `test/Compiler.Tests`: full suite green, including the new `Issue4247ZeroLengthAttributeArrayTests`:
  - `[0]Kind` / `[0]DayOfWeek` (no initializer) compile, ILVerify, run, and read back via `MetadataLoadContext` as an empty array of the correct enum type (`P.Kind[]` / `System.DayOfWeek[]`, not merely "compiles").
  - `[]Kind{}` (explicit empty initializer) parity control still compiles and reifies identically.
  - `[3]Kind` (nonzero constant, no initializer) and `[Length()]Kind` (nonconstant length) both still report `GS0202` — the rejection scope is unchanged for every shape this fix doesn't touch.
- Ran the same 132 attribute/array-related `Compiler.Tests` and 145 attribute/array-related `Core.Tests` as a focused regression pass: all green.
- Witness (ADR-0154): reverted the fix and re-ran `Issue4247ZeroLengthAttributeArrayTests` — the two "must compile clean" cases fail with `GS0202` at exactly the reported line (`@E([0]Kind)` / `@E([0]DayOfWeek)`), reproducing the issue precisely; both pass with the fix applied.

## Checklist

- [x] Behavioral tests carry a witness of discrimination (ADR-0154): confirmed the new tests fail on the pre-fix code (reverted the fix) and pass with it.
- [ ] Self-review verdict recorded when one was run (MERGEABLE / BLOCKER / SHOULD-FIX / NIT), with residuals filed as issues rather than dropped.


DavidObando:
Addressed in d11de03c: added boxed-object enum-array coverage (the same no-initializer zero-length syntax used in ordinary, non-attribute code, boxed to `object` and read back through the reverse cast) and runtime execution assertions on all compiled programs (previously only compiled + read metadata back through `MetadataLoadContext`, never actually run).

DavidObando:
Good catch — that was a genuinely separate gap. The issue's own repro is an object-VALUED attribute constructor parameter (Roslyn's `[Value(new Kind[0])]`), not the directly array-typed constructor my existing tests used. Added `NoInitializerZeroLengthArray_AsObjectValuedConstructorArgument_CompilesAndReifiesAsEmptyArray` in 5ce7928d, confirmed it reproduces the identical GS0202 pre-fix (reverted the product change and reran) and passes post-fix with correct `P.Kind[]` metadata identity.

PR #4258, opened 2026-09-16
cs2gs: fix implicit top-level local wrongly emitted as readonly let (#4239)

### LINKED ISSUE #4239: cs2gs: implicit top-level local mutated by a local function is emitted as readonly let (GS0127)
Found while validating #4212 / PR #4237. Reduced and confirmed on the exact pre-PR baseline `13c96e6e1f18f64886ba291b0b07757f022f2ecb`.

## Repro

An ordinary `net10.0` executable with implicit C# top-level statements:

```csharp
int value = 0;
void Update() { value = 1; }
Update();
System.Console.WriteLine(value);
```

The original C# builds and runs, printing `1`.

cs2gs emits:

```gsharp
import System

let value = 0

let Update = func () {
    value = 1
}

Update()
Console.WriteLine(value)
```

Compilation correctly rejects the assignment with `GS0127: Variable 'value' is read-only and cannot be assigned to.` The translator should emit a mutable binding for a top-level local mutated by its local function, while keeping genuinely immutable bindings as `let`.

## Baseline evidence

- Commit: `13c96e6e` (before the #4212 implementation).
- Compiler: `0.4.688+13c96e6e1f`; .NET SDK `10.0.400`.
- Baseline migration run: `2026-09-14T23-13-15Z_f1e902`.
- Stages: translate PASS, compile FAIL (`GS0127` at emitted `Program.gs:6:11`), ILVerify/test-parity skipped because compilation failed.
- Fingerprint: `sha256:25f8dbeca96809c0ab145582b945093071be32e57d9c1df8259ae247252e7e0f`.
- Native C# executable stdout: `1`.

This repro has **no named arguments, no omitted defaults, no delegate invocation, no recursive group, and no reference-argument spill**. It is independent of #4212's call-argument reassembly and its flattened-entry-point reference-frame fix, so it is tracked separately rather than broadening that PR.

Suggested coverage: implicit entry-point locals mutated by local functions/lambdas and by nested expressions, plus an immutable-binding control.

Part of #3501.

DavidObando:
Fixes #4239

## Root cause

`TranslateTopLevelProgram` (the translation path for a native C# top-level-statements program — `int value = 0;` at the top of `Program.cs`, no explicit `Main`) never set `DocumentTranslationState.CurrentBodyScope` before translating the entry statements. Every other body owner does: `TranslateBody` sets `CurrentBodyScope` to itself for a method, constructor, lambda, or local function, and `IsLocalReassigned` (via `IsSymbolReassigned`) walks that scope's descendants to decide `let` vs `var`.

With `CurrentBodyScope` left `null` for a top-level program, `IsSymbolReassigned` short-circuited to `false` for every top-level local — a write buried inside a sibling local function's (or lambda's) body was never seen, so the local was always bound `let`. gsc then correctly rejected the mutation with `GS0127` once the emitted G# was compiled.

## Fix

`TranslateTopLevelProgram` now sets `CurrentBodyScope` to the compilation-unit root for the whole entry-point translation (restored via `finally`, same pattern `TranslateBody` uses). Every global statement — and everything nested inside them, including a local function's or lambda's own body — is a descendant of that root, so the existing reassignment walk now sees writes anywhere in the top-level program, exactly like it already does for an explicit `Main`.

This is a single scope-setting fix at the point every other body owner already sets it, not a per-call-site patch — it also transparently fixes the same gap for every other `CurrentBodyScope`-based check exercised by a top-level program (deconstruction targets, `if`/pattern smart-cast invalidation, `ref`/`out` argument mutability, etc.), not just the plain local-declaration path the issue reports.

## What I tested

- `dotnet build -c Release`: 0 warnings, 0 errors.
- New regression tests in `tools/cs2gs/Cs2Gs.Tests/Issue4239ImplicitTopLevelLocalMutatedByLocalFunctionTests.cs` (4 tests, all passing), covering:
  - the exact repro from the issue (local function mutation) — compiles and runs end-to-end through the real `gsc`, printing `1`;
  - the same mutation via a lambda instead of a local function;
  - a mutation buried in a nested compound-assignment expression (`total += amount`) inside a local function;
  - an immutable-binding control — a local function that only *reads* the top-level local must keep `let` (the fix must not widen mutability, only correct it).
- Targeted regression filter across every top-level-statement/reassignment/capture test file (`Issue2382`, `Issue4197`, `Issue4212`, `Issue4219`, `Issue1736`, `Issue1886`, `Issue1904`, `Issue2222`, `Issue3399`, `Issue3413`, `Issue3460`, `Issue3645`, `Issue3837`, `Issue3865`, `Issue3935`, `LocalFunctionHoistTranslationTests`): 254 tests, all passing.
- Local equivalent of the "hot-core translation guard": `build/test-cs2gs-selfmig-pr-guard.sh` (fast, mocked) passes; `build/run-cs2gs-selfmig-pr-guard.sh` (the real translate→compile→ilverify→test-parity run over the 8 hot-core projects) run in full locally — see below.
- Checked release-notes convention with `git log -- website/docs/release-notes.md`: only `gsc:`-prefixed (compiler) fixes update that file; recent `cs2gs:`-only fixes (#4200, #4211, #4237) did not. This PR is cs2gs-only, so `website/docs/release-notes.md` is intentionally left unchanged.


DavidObando (commented):
Rubber-ducked strictly against #4239's scope. Verdict: solid, would approve — GitHub blocks self-approval on your own PR, so filing this as a comment review instead.

**Root cause confirmed by reading the code, not just the PR description.** `IsSymbolReassigned` (`CSharpToGSharpTranslator.ControlFlow.cs`) short-circuits to `false` when `scope == null`, and `IsLocalReassigned` always calls it with `this.state.CurrentBodyScope`. `TranslateTopLevelProgram` never set that field, unlike `TranslateBody`, which sets it to `bodyOwner` for every other method/ctor/lambda/local-function body (with the same try/finally restore shape this PR now mirrors). So every top-level local's mutability was decided against a `null` scope — always `false` — always `let`. Confirmed.

**Verified the fix doesn't leak/clobber scope.** `TranslateTopLevelProgram`'s new `try`/`finally` sets `CurrentBodyScope` to the compilation-unit root and restores the previous value (`null`, since it's only ever called once, non-recursively, from `CSharpToGSharpTranslator.cs`). Built gsc from the branch and confirmed empirically with a repro: top-level statements followed by a local `class` whose own method has a nested local function — the class's `Method()` correctly got `var local` (mutated by its own nested local function) with no interference from the top-level program's scope having already been set/restored. No leak.

**Verified the three "for free" claims empirically** (all `CurrentBodyScope`-gated, as the PR claims):
- `ref` argument to a helper → `var value` + `Helper(&value)` — correct.
- Deconstruction assignment from inside a local function (`(a, b) = (1, 2)`) → both targets emitted `var` — correct.
- `if`/pattern smart-cast invalidation: a top-level `if (o is int i) {...} i = 100;` → `i` correctly ends up `var` (reassigned after the pattern), `o` stays `let` — correct.

**Checked for false positives** — none found:
- Multiple local functions each mutating a *different* top-level local: both get `var` independently, no cross-contamination.
- A local only mutated inside a lambda that's never invoked still gets `var` — correct, since mutability is a static/syntactic property, not a reachability one (matches existing non-top-level behavior).
- Same-name shadowing between a top-level local and an unrelated nested local: no interference, since `IsSymbolReassigned`/`BindsTo` compare by `ISymbol` identity via `SymbolEqualityComparer`, never by name.

**Ran the build and the author's targeted regression filter** (Issue2382/4197/4212/4219/1736/1886/1904/2222/3399/3413/3460/3645/3837/3865/3935/LocalFunctionHoistTranslationTests + the 4 new Issue4239 tests): 174/174 passing, 0 warnings/errors on `dotnet build -c Release`.

**Release-notes convention sanity-checked**: `website/docs/release-notes.md`'s git log only has `gsc:`-prefixed entries; `cs2gs:`-only commits (#4200, #4211, #4237) never touch it. Leaving it alone here is correct.

Two things noted, neither blocking, both out of #4239's scope so not fixed here:
1. The open Copilot comment (workspace cleanup for the 4 new e2e tests, pointing at the existing best-effort pattern in `Issue3841DelegateOverloadSetTranslationTests.cs:256-274,315-328`) is a legitimate, pre-existing convention this new file doesn't follow — test-output hygiene only, not a correctness issue.
2. Widening `CurrentBodyScope` to the whole compilation-unit root for a top-level program means two file-scoped-but-otherwise-unrelated helpers (`IteratorForeachReceiverRequiresAssertion` / `ForEachBindingInfersNonNullElement` in `CSharpToGSharpTranslator.Expressions.cs:1056-1088,1864-1873`) can now see a `yield` statement anywhere else in the *same file* (e.g., an unrelated class's iterator method) when reasoning about a top-level `foreach`'s nullable-receiver assertion. I confirmed this can only make the output more conservative (an occasional redundant, harmless `!!`, or a missed non-null-inference optimization) — never unsound or a build break. Narrow edge case (top-level statements + a colocated unrelated iterator method + a specific nullable-foreach pattern); filing separately if worth tightening, not touching it here.

Issue #2350, opened 2026-07-15
gsc: async functions reject ref-struct locals that do not live across await


DavidObando:
## Summary

The binder blanket-rejects `Span<T>`/ref-struct locals anywhere in an async function (GS0219), even when a local is scoped to one loop iteration and is dead before every await. C#/Roslyn permits this with liveness-based state-machine safety.

Real Oahu.Diagnostics occurrence: PipelineProbeCheck.

## Expected fix/tests

Replace the coarse issue-#367 rule with per-local control-flow/liveness analysis: allow ref-struct locals whose lifetime cannot cross suspension; continue rejecting paths that reach await/yield while live. Cover branches, loops, try/finally, and negative escape/capture cases.

PR #4260, opened 2026-09-16
Allow native ref aliases in async/iterator functions when confined to one suspension segment

### LINKED ISSUE #4222: gsc: allow native ref aliases in async and iterator regions that do not cross suspension (GS0258)
## Capability gap

Allow native **`let ref` / `var ref` aliases in async functions and iterators when the alias is not used across a suspension boundary**.

This is a separate composability gap discovered while following #4219's ref analysis. It does not require recursion, function literals or ref-returning calls: the repro aliases an array element directly.

#2350 implemented analogous liveness-based support for ref-struct locals, but native ref-alias declarations still have a blanket context prohibition. Do not confuse this with the existing low-level `&`/`*` pointer path, which already supports some related shapes.

## Reproduced on gsc

Compiler rebuilt from `07d64f9d`: `0.4.635+07d64f9d8c`, SDK `10.0.400`.

```gsharp
package P
import System.Threading.Tasks
async func Run() int32 {
    var values = []int32{10}
    {
        var ref alias = values[0]
        alias = 42
    }
    await Task.Delay(1)
    return values[0]
}
Console.WriteLine(Run().Result)
```

Actual: **GS0258**, `Ref-aliasing local 'alias' cannot be declared as a local in an async function (it would be hoisted into the state machine)`.

The alias has left its lexical block before the await. There is no requirement to store a managed pointer in a state-machine field.

The iterator sibling also fails with GS0258:

```gsharp
package P
func Values() sequence[int32] {
    var values = []int32{10}
    {
        var ref alias = values[0]
        alias = 42
    }
    yield values[0]
}
for value in Values() {
    Console.WriteLine(value)
}
```

### C# and G# controls

The equivalent C# async method was compiled with `-langversion:13.0` and returned **42**:

```csharp
static async Task<int> Run()
{
    int[] values = { 10 };
    { ref int alias = ref values[0]; alias = 42; }
    await Task.Delay(1);
    return values[0];
}
```

The C# iterator equivalent yielded **42**. This is supported beginning with C# 13; no preview-language extension was needed.

Changing only the inner G# block to the following also compiled and returned **42**:

```gsharp
{
    var slot = &values[0]
    *slot = 42
}
```

Thus this is a gap in native alias composition and state-machine treatment, not proof that G# cannot express the underlying operation at all.

A negative C# control that used the ref alias after `await` was rejected with **CS9217**. Preserving an actual managed ref across suspension is not the requested feature.

## Root cause

`BindRefAliasLocalDeclaration` rejects all aliases when the function is async/suspending or iterator-returning, before any per-local liveness question can be answered.

Simply removing that guard is unsafe because the alias and pointer representations differ:

- A native alias keeps the **pointee type** on its `LocalVariableSymbol` and carries `RefKind.Ref`.
- `RefInitializationHoister` recognizes managed-pointer locals through **`ByRefTypeSymbol`** plus an address-of initializer.
- `AsyncCaptureWalker` skips ByRef-typed locals and ref-struct locals, but otherwise broadly hoists referenced locals. A native ref alias with pointee type `int32` does not become safe merely by passing the binder.
- `RefStructAsyncLivenessAnalyzer` provides useful CFG, nested-function, loop and exceptional-flow machinery, but its tracked set is currently by-ref-like locals and its entry points focus on async scopes. Native aliases and ordinary iterator yields require deliberate coverage.

Sources at `07d64f9d`:

- `src/Core/CodeAnalysis/Binding/StatementBinder.Narrowing.cs:1908-1933`
- `src/Core/CodeAnalysis/Lowering/Async/RefInitializationHoister.cs:13-40,69-91`
- `src/Core/CodeAnalysis/Lowering/Async/AsyncCaptureWalker.cs:13-23,50-98`
- `src/Core/CodeAnalysis/Binding/RefStructAsyncLivenessAnalyzer.cs:13-51,72-91,529-549`
- `test/Core.Tests/CodeAnalysis/Emit/AsyncRefLocalEmitTests.cs:24-107` exercises `&`-initialized pointer locals, not native alias declarations.

## Recommended path

1. Decide the common lowered representation of native aliases versus managed-pointer locals. Reuse the existing mechanisms, but do not pretend their current symbol contracts are identical.
2. Track alias uses and referent validity through control flow, recognizing await and yield suspension points, nested scopes and exceptional paths.
3. Keep aliases that are confined to one execution segment in valid method/MoveNext locals, or normalize them through a proven equivalent lowering. Never emit a byref state-machine field.
4. Replace the blanket declaration error with a specific diagnostic for genuinely prohibited cross-suspension use.
5. Preserve one-time evaluation of alias initializers and their null/bounds exception timing. Avoid value-copy substitutes.
6. Apply the same principle to ordinary async functions, async literals/local functions, synchronous iterators and async iterators, validating each backend independently.

**Scope boundary:** do not use this issue to authorize arbitrary refs across await/yield. Existing pointer-reconstruction behavior is a separate established capability and should not be weakened or expanded accidentally. General ref-returning call support and readonly aliases compose with this work later; neither is required to fix the concrete array-element repro.

## Definition of done

- [ ] Both G# repros produce 42; `let ref` and `var ref` are covered.
- [ ] The existing low-level pointer control and ref-struct-local behavior remain correct.
- [ ] Cover aliases before the first suspension, created after a suspension, and used only between successive suspensions.
- [ ] Cover loops, branches, try/catch/finally, nested functions and aliases whose use is hidden in another expression.
- [ ] Prohibited across-suspension uses are rejected precisely, including exceptional-flow paths.
- [ ] No byref state-machine fields or copies that silently replace alias identity are emitted.
- [ ] Receiver/index side effects and declaration-time exceptions occur once at the original point.
- [ ] Validate emitted runtime behavior and IL for async, iterator and async-iterator routes.
- [ ] cs2gs can retain safe C# ref locals in these contexts without forcing source-level pointer rewrites.

C# baseline:
`https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-13#ref-and-unsafe-in-iterators-and-async-methods`.

DavidObando:
Fixes #4222

## Root cause

`BindRefAliasLocalDeclaration` blanket-rejected every `let ref`/`var ref` alias declared in an async or iterator function with `GS0258`, before any per-local liveness question could be asked. Issue #2350 already replaced the analogous blanket rejection for by-ref-like (`ref struct`) locals with a sound dataflow liveness analysis (`RefStructAsyncLivenessAnalyzer`); native ref aliases never got the same treatment because they have a different symbol shape: an alias keeps its **pointee** type on `LocalVariableSymbol.Type` and carries the byref-ness out of band via `RefKind`, whereas a by-ref-like local is recognized by its `Type` directly. Every mechanism keyed on `TypeSymbol.IsByRefLike`/`ByRefTypeSymbol` therefore never saw a native alias at all.

## Fix

- `RefStructAsyncLivenessAnalyzer` now also tracks native ref-alias locals (`LocalVariableSymbol.RefKind != RefKind.None`) and treats `yield` as a suspension point alongside `await`, so it runs for iterator functions too. By-ref-like locals stay async-only (issue #2350's own scope is untouched — a by-ref-like local in a plain iterator remains the pre-existing blanket rejection).
- Assigning *through* an alias (`alias = 42`) is modeled as a **use** of the alias (it needs the address to write through), not a redefinition of it — unlike overwriting a by-ref-like value local outright, which legitimately kills prior liveness.
- The blanket declaration-site rejection in `BindRefAliasLocalDeclaration` for async/iterator functions is removed. A local actually proven live across a suspension still reports `GS0258`, now with a precise reason instead of a blanket one.
- `AsyncCaptureWalker` and `IteratorRewriter`'s hoist-local collector both exclude alias locals from state-machine fields (mirroring the existing by-ref-like exclusion), so a confined alias simply stays a plain state-machine-method local. The pre-existing `RefKind`-driven `ldind`/`stind` emission already handles such a local correctly (confirmed by tracing `MethodBodyEmitter`/`SlotPlanner`) — **no emitter changes were needed**.

The existing low-level `&`/`*` pointer-reconstruction path (`RefInitializationHoister`) and by-ref-like-local liveness are unchanged.

## What was tested

- `dotnet build` (Release) of the full solution: 0 warnings, 0 errors.
- New tests:
  - `test/Core.Tests/CodeAnalysis/Binding/Issue4222RefAliasSuspensionLivenessTests.cs` — binder-level `GS0258` liveness coverage: both repros (async + iterator), `let ref`/`var ref`, an implicit async-iterator (`IAsyncEnumerable[T]` without the `async` keyword), loops, branches, try/catch/finally, nested async lambda, alias created after a suspension, alias confined between two suspensions, negative controls for cross-suspension use (including through `alias = await X()`) and for the pre-existing by-ref-like-in-plain-iterator blanket rejection (unchanged).
  - `test/Core.Tests/CodeAnalysis/Emit/Issue4222RefAliasSuspensionEmitTests.cs` — end-to-end PE round trips for both of the issue's repros plus their `let ref` siblings and an async-iterator route, each asserting the runtime value (`42`) and that no type in the emitted assembly has a by-ref-typed field (the IL-shape requirement from the issue's definition of done).
  - `tools/cs2gs/Cs2Gs.Tests/Issue1900RefLocalsTranslationTests.cs` — a new case translating a confined C# `ref` local inside an `async` method, proving cs2gs's existing (unconditional, no async-specific special-casing) `ref`-local translation now also *binds* against the fixed gsc, satisfying the issue's "cs2gs can retain safe C# ref locals in these contexts" requirement with no translator changes needed.
- Regression sweep, all green:
  - `Core.Tests`: 600 pre-existing tests filtered on `Async|Iterator|RefLocal|RefStruct|RefSafe` + 45 new/updated tests for this issue.
  - `Compiler.Tests`: 467 tests, same filter plus `Issue2350`/`Issue2360`.
  - `Cs2Gs.Tests`: 9 tests in the `Issue1900` ref-local translation suite.
- Hot-core cs2gs self-hosting translation guard (`build/run-cs2gs-selfmig-pr-guard.sh`): `src/Core.csproj` — the project containing every changed file — translated to G# and **passed** translation. The guard's later compile/ilverify/test-parity steps for that project were still running locally when this PR was opened; CI's own `hot-core-translation-guard` job will re-run the full guard on the PR.

## Explicitly not deferred

Every checklist item in the issue's "Definition of done" is covered: both repros produce 42 (`let ref` and `var ref`); the pre-existing pointer (`&`/`*`) and by-ref-like-local behavior is untouched (regression suite green); aliases before/after/between suspensions, loops, branches, try/catch/finally, and nested function literals are all covered by tests; cross-suspension use (including through an assignment) is precisely rejected with `GS0258`; no by-ref state-machine field is ever emitted (asserted directly via reflection); receiver/index side effects still evaluate once (unchanged — this fix only widens *when* a confined alias is allowed, not how one is lowered); async, iterator, and async-iterator routes are all validated at runtime and by IL shape; cs2gs composes with no translator change required.


DavidObando (commented):
Rubber-duck review against issue #4222's scope and DoD. Root cause and overall approach are sound: reusing #2350's RefStructAsyncLivenessAnalyzer machinery for native ref-alias locals (tracked via `RefKind` since an alias keeps its pointee type on `LocalVariableSymbol.Type`), and modeling assignment-through-an-alias as a use rather than a kill, are both correct calls given the symbol-shape difference the PR body describes. `AsyncCaptureWalker` and `IteratorRewriter.LocalCollector` exclusions are correctly wired, including via `AsyncIteratorRewriter`'s reuse of `AsyncCaptureWalker.Analyze` for the async-iterator route (no separate patch needed there). No scope creep: changes stay within the binder/liveness/hoisting files the issue's own root-cause analysis named, `cs2gs` gets a test only (no translator change, as claimed), and `Binder.IsIteratorReturnType`'s private→internal bump is a legitimate, minimal reuse rather than a new abstraction.

Found and fixed one genuine, concrete correctness bug while verifying the liveness analyzer is sound (not just a diagnostic-completeness gap):

`RefStructAsyncLivenessAnalyzer.ApplyExpression` (and the analogous `BoundVariableDeclaration` case) folded an expression's own interesting-variable reads into `live` *after* checking for an unsafe await crossing, not before. That makes the check blind to a read positioned, in evaluation order, strictly *after* a nested `await` within the *same* expression (as opposed to being the `await`'s own assignment target, which was already handled correctly) — unless something later in the function also needed the value.

Concretely, `total = await X() + alias` reads `alias` only once the `await` resumes, so the alias's address (not just its pointee's value) must survive the suspension — but the old ordering let this compile with no GS0258. I verified this is a real, reachable defect, not just theoretical: built gsc from the branch and ran it against that exact shape with a genuinely-suspending `await Task.Delay(...)` — it compiled clean and then threw a `NullReferenceException` at runtime (dereferencing a never-hoisted, reset-to-default alias slot after resume). The equivalent by-ref-like-local shape from #2350 (`await X() + span.Length`, unmodified code) has the identical gap and is worse: it *silently* returns a wrong value (0) instead of erroring, since `span` resets to its CLR default rather than crashing. So this is a pre-existing latent gap in the shared machinery, but #4222 is precisely what makes it newly reachable for aliases, and the issue's own DoD requires "prohibited across-suspension uses are rejected precisely" — this was a real hole in that guarantee.

Fix (pushed to this branch, `29430fd6`): union the expression's reads into `live` before the `ContainsAwait` check, in both call sites. This matches the analyzer's own already-established "conservative, never unsound" stance for self-referential by-ref-like reassignment (`span = span.Slice(await X())`, same file, existing comment: "evaluation order does not change which value is live entering the statement") — extended uniformly rather than left as a special case. Net effect: a same-expression read that provably always precedes the `await` (e.g. `alias + await X()`) is now also conservatively rejected rather than precisely allowed, which I think is the right tradeoff for a crash/silent-wrong-value fix and is a tiny, uncommon pattern (trivially worked around by splitting the `await` onto its own statement).

Verified:
- Full solution `dotnet build -c Release`: 0 warnings, 0 errors.
- Added 4 regression tests to `Issue4222RefAliasSuspensionLivenessTests.cs` covering the after-await miss (both as an assignment RHS and a `var` initializer) and the before-await conservative-reject.
- Re-ran the PR's own regression sweep with the fix applied: `Core.Tests` 602/602 (Async|Iterator|RefLocal|RefStruct|RefSafe filter) + all Issue4222 tests (28/28), `Compiler.Tests` 467/467 (same filter + Issue2350/Issue2360), `Cs2Gs.Tests` Issue1900 suite 9/9 — all green, matching the counts claimed in the PR body.
- Confirmed the hot-core guard concern in the PR body (compile/ilverify/test-parity stages not locally confirmed) is a reasonable thing to leave to CI — didn't re-run the full self-hosting guard myself given the ~hour cost, but nothing in this review suggests it's at risk.

No other defects found. The rest of the liveness CFG (loops, try/catch/finally ambient-live modeling, switch, nested async/iterator lambdas) checked out correctly against both the new tests and manual tracing.

Issue #4259, opened 2026-09-16
gsc: closure captures a ref/out/in parameter silently instead of being rejected


DavidObando:
## Summary

A closure (ordinary non-generic `let X = func(...) {...}`, and equally a generic local function per #4221/#4252) that captures a `ref` parameter of its enclosing function compiles successfully and silently loses the mutation instead of being rejected at compile time, the way a `ref struct` capture (GS0219), a managed-pointer capture (GS9004), or a fixed-pointer capture (GS9008) already are.

## Repro

```gsharp
package P
func Foo(ref x int32) {
    let Bad = func(a int32) int32 {
        x = x + 1
        return a
    }
    Bad(1)
}
var y = 1
Foo(ref y)
Console.WriteLine(y)
```

Expected: either a compile-time diagnostic rejecting the capture of a `ref` parameter (matching the issue #4221 recommended-implementation-path note: "Illegal captures of ref locals, ref/out/in parameters and ref-struct variables must remain rejected"), or, if by-ref parameter capture is meant to work, correct shared-reference semantics.

Actual: exit 0, prints `1` — the write inside the closure does not reach `y`. The mutation is silently lost with no diagnostic.

Confirmed this reproduces identically whether the closure is an ordinary non-generic `let X = func(...) {...}` or a generic local function (`let X[T] = func(...) {...}`), so this is a pre-existing gap in ordinary closure-capture validation, not something introduced by #4221/#4252's generic-local-function capture support.

## Root cause (not investigated in depth)

The capture-legality checks in `LambdaBinder.BindFunctionLiteralBody` (see `src/Core/CodeAnalysis/Binding/LambdaBinder.cs`) currently reject `TypeSymbol.IsByRefLike` (ref-struct), `ByRefTypeSymbol` (managed pointer) and `PointerTypeSymbol` (fixed pointer) captures, but there is no check for a captured variable that is itself a `ref`/`out`/`in` parameter binding.

## Found while

Auditing issue #4221 ("allow generic local functions to capture ordinary outer state") for full DoD coverage after PR #4252. Filed separately because it's out of scope for #4221 — it affects ordinary (non-generic) closures identically and predates that work.

Issue #4223, opened 2026-09-14
gsc: reify enclosing type parameters for generic and non-generic local functions (GS0468)


DavidObando:
## Capability gap

Allow local functions to use **type parameters from enclosing methods and types**, whether the local function has its own type parameters or not. This is a type-environment/reification issue independent of outer-value captures, recursion groups and ref-returning literals.

Follow-up to #4219 and the intentionally diagnostic-based resolutions of #1940 and #2016. Those issues stopped invalid IL by rejecting affected programs; their closure did not implement the capability.

## Reproduced examples

Compiler rebuilt from `07d64f9d`: `0.4.635+07d64f9d8c`, SDK `10.0.400`.

### Generic local combining outer and own type parameters

```gsharp
package P
func Outer[U](seed U) U {
    let Keep[T] = func(ignored T, value U) U { return value }
    return Keep(1, seed)
}
Console.WriteLine(Outer("ok"))
```

Actual: **GS0468**, `Local function 'Keep' cannot reference 'U', a type parameter of an enclosing method or class`.

There is no outer-value capture in the local body: both `ignored` and `value` are its own parameters.

The corresponding C# method compiled with `-langversion:13.0` and returned **"ok"**:

```csharp
static U Outer<U>(U seed)
{
    U Keep<T>(T ignored, U value) => value;
    return Keep(1, seed);
}
```

### Even a non-generic local is affected

```gsharp
package P
func Outer[U](seed U) U {
    let Keep = func(value U) U { return value }
    return Keep(seed)
}
Console.WriteLine(Outer("ok"))
```

Actual: **GS0468**. The equivalent non-generic C# local method returned **"ok"**.

Changing the G# declaration to `let Keep = (value U) -> value` also produced **GS0468**. Arrow syntax is not a workaround for this case.

### Contrasting control: introducing a capture changes the outcome

```gsharp
package P
func Outer[U](seed U) U {
    let Keep = func() U { return seed }
    return Keep()
}
Console.WriteLine(Outer("ok"))
```

This compiled and printed **ok**. The capturing form already uses a reified closure environment. A separate generic local using only its own type parameters also compiled and ran.

**The compiler should not require an otherwise unnecessary capture just to give an enclosing type parameter a valid emitted owner.**

## Impact and root cause

[...]

PR #4261, opened 2026-09-16
Fix #4221 follow-up: generic local function calling a capturing sibling crashed

### LINKED ISSUE #4221: gsc: allow generic local functions to capture ordinary outer state (GS0463 capability gap)
## Capability gap

Allow a **generic local function to capture ordinary outer state**, preserving shared-variable semantics. This fails even with one non-recursive local function, so it is independently useful and not contingent on mutual-recursion groups or ref-returning literals.

Follow-up to #4219; historical foundation: #1886. This issue requests actual support, not merely a diagnostic replacing invalid IL.

## Reproduced on the inspected compiler

Built `src/Compiler/Compiler.csproj` Release from `07d64f9d`; compiler informational version `0.4.635+07d64f9d8c`, .NET SDK `10.0.400`. The following G# program was compiled through the real gsc CLI:

```gsharp
package P
func Run() int32 {
    var count = 0
    let Add[T] = func(value T) T {
        count = count + 1
        return value
    }
    Add(1)
    Add("a")
    return count
}
Console.WriteLine(Run())
```

Actual: exit 1, **GS0463: Generic local function 'Add' cannot capture variables from the enclosing scope.**

The C# equivalent compiled with `-langversion:13.0` and returned **2**:

```csharp
static int Run()
{
    int count = 0;
    T Add<T>(T value) { count++; return value; }
    Add(1);
    Add("a");
    return count;
}
```

Contrasting G# controls also compiled and ran:

- Remove `[T]` and use `int32` for `Add`'s parameter/return, calling `Add(1)` and `Add(2)`: returns **2**. Ordinary closure mutation already works.
- Keep a generic `Keep[T] = func(value T) T { return value }` without a capture: `Keep(42)` returns **42**. Generic local calls already work.

The missing feature is the composition of the two.

## Impact

Generic traversal helpers, generic memoization helpers, local adapters and algorithms cannot retain lexical state without manually moving the function and threading state through its arguments. An enclosing method's counter, cache or instance should not force an otherwise local generic helper out of its lexical scope.

This is not a request for a first-class value representing an unbound generic method. C# local methods can be called directly; a closed callable can be materialized only where a delegate is actually needed.

## Root cause and existing machinery

`LambdaBinder.BindGenericLocalFunctionDeclaration` creates a callable `FunctionSymbol` and a `BoundLocalFunctionDeclaration`, then rejects any nonempty `CapturedVariables` list. That direct-call representation avoids trying to store an open generic method in a delegate.

The existing closure path is not yet an interchangeable replacement: `ClosureEmitter` explicitly keeps generic locals on the direct top-level method path, while its synthesized closure `Invoke` method does not model a local method's own generic parameters. Ordinary direct calls also need access to the proper environment and generated method; removing GS0463 alone is not a fix.

Reuse:

- `CaptureBoxingRewriter` for shared cells and outer/inner mutation.
- `ClosureEmitter` and existing closure reification for lexical ownership and generic environment types.
- Generic direct-call resolution and MethodSpec emission.
- Existing capture discovery, extended for transitive calls where necessary.

Sources at `07d64f9d`:

- `src/Core/CodeAnalysis/Binding/LambdaBinder.cs:556-635`
- `src/Core/CodeAnalysis/Binding/BoundLocalFunctionDeclaration.cs:9-37`
- `src/Core/CodeAnalysis/Emit/ClosureEmitter.cs:251-307,399-469`
- `src/Core/CodeAnalysis/Lowering/CaptureBoxingRewriter.cs:13-46`
- `test/Compiler.Tests/Emit/Issue1886GenericLocalFunctionEmitTests.cs:110-129`

## Recommended implementation path

1. Represent a generic local method's lexical environment explicitly, without converting its open generic declaration into a delegate.
2. Preserve one shared environment/cell identity across different instantiations of that local method. Calling `Add<int>` and `Add<string>` must update the same `count`.
3. Route direct calls to the generated generic method with the correct environment. Do not create a fresh environment on each call.
4. Preserve lexical accessibility for captured `this` and private/protected members.
5. Support closed method-to-delegate use without losing captured-state identity.
6. Compose this with the separate enclosing-type-parameter workstream and, subsequently, recursion-group support under #4219.

**Do not weaken safety restrictions wholesale.** Ordinary value captures are the capability requested here. Illegal captures of ref locals, ref/out/in parameters and ref-struct variables must remain rejected. C# rejects ref-struct captures too; that is not a parity gap.

Definite assignment needs deliberate handling: local-method calls and delegate variables are not the same model. Keep capture reads and writes correct around calls, including transitive local-method calls; extend recursive summaries when recursion groups are introduced.

## Definition of done

- [ ] The reproducer compiles and returns 2; both existing controls remain correct.
- [ ] Multiple generic instantiations share captured state; mutations are visible both inside and outside the local function.
- [ ] Captured parameters, immutable values, mutable reference/value types and `this` are covered.
- [ ] Transitive captures, nested local functions and closed delegate conversions retain environment identity.
- [ ] Enclosing generic parameters and lexical accessibility compose with the corresponding capability workstream.
- [ ] Illegal captures still receive specific diagnostics; no byref fields appear in heap closures.
- [ ] Add emitted-runtime and IL-verification regressions; update the GS0463 blanket-rejection test only for shapes that now work.
- [ ] Retarget affected cs2gs local-function cases and prove behavior, not just improved printed names.

Reference: Microsoft Learn, "Local functions", variable capture and delegate implementation:
`https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions`.

DavidObando:
## Summary

Audited issue #4221 ("allow generic local functions to capture ordinary outer state") against already-merged PR #4252. The core capability works — the issue's own repro now compiles and returns 2, and both its documented controls still pass — but #4252 left one real gap and stayed open because its PR body used "resolving issue #4221" rather than a GitHub auto-close keyword.

**The gap:** #4252 removed two negative test rows from `test/Core.Tests/CodeAnalysis/Issue4219GenericLocalRecursionTests.cs` that used to assert `GS0463` for a generic local function calling a sibling local function that itself captures outer state — without ever adding a positive replacement proving the now-legal shape actually works. It doesn't:

```gsharp
package P
func Run() int32 {
    var outer = 1
    let first[T] = func(x T) int32 { return second(x) }
    let second[U] = func(x U) int32 { return outer }
    return first(9) + first("z")
}
Console.WriteLine(Run())
```

`first` never reads `outer` itself — it only calls `second`. This crashed `gsc` with `GS9998: InvalidOperationException: Variable 'outer' has no local slot or parameter index in the current method.`, regardless of whether `second` was declared before or after `first`, or the two called each other in a cycle.

## Root cause

A generic local function is called directly through its `FunctionSymbol`, not through a delegate value (that's the whole point of #4221/#4252's design — a CLR delegate can't close over an unbound generic method). `LambdaBinder`'s existing transitive-capture folding already walks a *physically nested* function literal's `CapturedVariables` into its enclosing literal's own set (`CapturedVariableCollector.RewriteFunctionLiteralExpression`, issue #503 follow-up), but had no equivalent for a `BoundCallExpression` referencing a *sibling* local function's `FunctionSymbol`. Nothing propagated `second`'s capture of `outer` into `first`'s own environment, so `MethodBodyEmitter.EmitGenericLocalClosureInstance` had no field to read `outer` from when rebuilding `second`'s closure instance at the `second(x)` call site inside `first`'s body.

## Fix

- `CapturedVariableCollector` now consults a per-binder map of every local function's already-known captures (keyed by `FunctionSymbol`) and folds a callee's captures into its caller, exactly like the nested-literal case does.
- `LambdaBinder.ReconcileGenericLocalFunctionGroupCaptures` re-converges a whole consecutive generic-local-function-declaration group to a fixed point after every member's body is bound. The in-order pass alone only sees a callee declared *earlier*; this covers a forward-referenced callee and a call cycle between two members (verified: both terminate and produce correct results — shared-state identity is preserved through the transitive call in every direction, and even a genuine call cycle between two capturing generic locals converges correctly).
- `BoundFunctionLiteralExpression.CapturedVariables` gained an internal setter so the reconciliation pass can widen it after initial binding.

## Also verified (Definition of Done items #4252 left untested)

Added regression coverage for shapes that already worked but had no test locking them in: captured `this` (instance field mutation), a captured parameter, a mutable reference-type capture, a plain local function nested inside a generic local function's own body, and composition with an enclosing generic class's type parameter. Restored the two deleted `Issue4219GenericLocalRecursionTests` rows as positive assertions (no more `GS0463`, no `GS9998`, correct runtime value).

Also updated `docs/generic-local-recursion.md` (was still describing capture as forbidden) and a stale comment in `ClosureEmitter.cs` referencing the removed GS0463 rule.

## Explicitly out of scope

- **Enclosing generic type parameters** (#4223) — only verified composition doesn't crash; actually referencing an enclosing type parameter from inside a capturing generic local remains that separate workstream's job.
- **Closed delegate conversion of a generic local function** (e.g. `var d Func[int32,int32] = Add[int32]` or bare `Add`) — confirmed this doesn't work at all in G# today, with or without a capture, with or without genericity (`'Add' is not a variable`). This is a general, pre-existing language gap, not something #4221 introduced or is positioned to fix as a side effect.
- **A `ref`/`out`/`in` parameter capture silently compiling and losing its mutation instead of being rejected** — reproduces identically for an ordinary non-generic closure, so it predates #4221 entirely. Filed separately as #4259 rather than fixed here.
- No cs2gs corpus fixture exercises a capturing generic local function (checked the translator source and the grid corpus — nothing needed retargeting because nothing previously worked around `GS0463`), so there was nothing to retarget for this issue's last Definition-of-Done item.

## Testing

- `dotnet test test/Core.Tests` (full suite): 9301/9301 passed.
- `dotnet test test/Compiler.Tests --filter "FullyQualifiedName~Lambda|FullyQualifiedName~Local|FullyQualifiedName~Generic"`: 1565/1565 passed.
- `dotnet test test/Compiler.Tests --filter "FullyQualifiedName~Issue1886GenericLocalFunctionEmitTests"` and `...~Issue4219GenericLocalRecursionTests`: all passing, including 8 new tests.
- Manually compiled and ran the issue's exact repro (`gsc`): returns `2`.
- `build/run-cs2gs-selfmig-pr-guard.sh` (the hot-core cs2gs self-hosting guard): `src/Core.csproj` — the project containing every change in this PR — translated cleanly through cs2gs (translate stage passed); the full multi-app guard (compile/ilverify/test-parity across all guarded apps) was still running when this PR was opened due to shared-sandbox contention from concurrent work. Will report back if it surfaces anything; CI runs this guard authoritatively regardless.

Fixes #4221


DavidObando:
Update: the local cs2gs self-hosting guard (build/run-cs2gs-selfmig-pr-guard.sh) has now fully validated src/Core.csproj — the project containing every change in this PR — through all four stages: translate (221.8s), compile (929.2s), ilverify (1.0s), and test-parity (all passed). This resolves the caveat in the PR description, which was written while that run was still in the compile stage. The guard continued on to the other guarded hot-core apps (Formatting, Cs2Gs.CodeModel, Cs2Gs.Translator, etc.), which don't contain any changes from this PR.

DavidObando:
Full guard result: 7/8 guarded apps PASS (translate+compile+ilverify+test-parity), including src/Core.csproj — the only project this PR touches — fully green across all four stages.

The one failure, tools/cs2gs/Cs2Gs.Pipeline.csproj, is not a code defect: it failed at the compile stage with `MSB4236: The SDK 'Gsharp.NET.Sdk/0.4.704-g700a9a84d5' specified could not be found` — an MSBuild SDK-restore error, not a gsc diagnostic against any emitted G# construct. This sandbox had multiple concurrent agent sessions (different worktrees, same base commit) running this same guard simultaneously, all racing to restore/repack the identically-versioned Gsharp.NET.Sdk nupkg into the same shared NuGet cache — a resource-contention artifact of this environment, not something introduced by this PR's diff. Cs2Gs.Pipeline does not reference any file this PR changes.

No action needed on this PR for that failure; flagging for transparency. CI's own guard run (uncontended) is authoritative.

DavidObando (commented):
Rubber-duck review against issue #4221's scope, following up on merged PR #4252.

## What I checked

- Read #4221 in full (Definition of Done, recommended path, explicit non-goals) and PR #4261's diff/body/commits.
- Verified the reconciliation algorithm (`LambdaBinder.ReconcileGenericLocalFunctionGroupCaptures`) is a correct, terminating fixed-point loop: `CapturedVariables` only ever grows (direct captures are fixed per body-walk, callee-folded captures come from a dictionary that itself only grows), bounded by the finite set of in-scope variables, so the loop must converge; `maxPasses = max(8, N²)` is generous headroom over the number of hops actually needed to propagate a capture around an N-member cycle or chain.
- Built `gsc` from the branch and ran the issue's original repro plus the PR's 8 new tests — all produce the documented runtime values, not just "doesn't crash" (confirmed with the debugger of manual arithmetic against a 3-member chain, a mutual-recursion cycle, and a staggered-partial-capture 3-member group — all matched hand-computed expected values).
- Confirmed scoping is clean: no attempt at enclosing-type-parameter reification (#4223) or the ref/out/in capture bug (#4259); `EnclosingTypeParameters_StillRejected` still passes untouched.
- The deferred full multi-app cs2gs self-hosting guard is a reasonable thing to leave to CI per this repo's stated convention (targeted local filters, full suite on PR CI) — the translate-stage pass on `src/Core.csproj`, which is 100% of the changed surface, is adequate local evidence.

## Gap found and fixed (pushed to this branch, commit `0afe6388`)

The reconciliation pass only re-walked the group's own **top-level** members. A plain (or generic) local function declared *inside* a member's own body, that itself calls a sibling capturing outer state, was missed — its own `CapturedVariables` is fixed at its bind time, and nested-literal folding only consults that cached set without re-descending into the nested body. So the same class of crash the PR fixes reappears one level of nesting deeper:

```gsharp
let First[T] = func (v T) int32 {
    let Inner = func (x int32) int32 { return Second(x) }   // Inner never touches `outer` itself
    return Inner(1)
}
let Second[U] = func (v U) int32 { outer = outer + 1; return outer }
```

This crashed with the identical `GS9998: Variable 'outer' has no local slot...` when `Second` is a forward reference (declared after `First`). Order-reversed it happened to work by accident (callee already bound when `Inner` was first bound), which is why none of the 8 new tests caught it — none combine "nested local function" with "calls a sibling declared later / in a cycle."

Fix: `ReconcileGenericLocalFunctionGroupCaptures` now also collects every function literal nested inside a group member's body at any depth (new `NestedFunctionLiteralCollector`, mirroring the existing `NestedCaptureCollector` traversal pattern) and folds them into the same fixed-point loop, so a nested literal's captured set is refreshed before its enclosing member's re-walk reads it. This is idempotent for literals that don't call a group sibling — recomputing produces the same result given the same dictionary state — so it can't perturb an unrelated closure inside a member body.

Added one permanent regression test (`GenericLocalFunction_NestedLocalFunctionCallsForwardSiblingThatCaptures_PropagatesTransitively` in `Issue1886GenericLocalFunctionEmitTests.cs`) and extended the release-notes entry to describe this fourth shape. Verified:
- `dotnet test test/Compiler.Tests --filter "FullyQualifiedName~Issue1886GenericLocalFunctionEmitTests"`: 17/17 (16 existing + 1 new).
- `dotnet test test/Compiler.Tests --filter "FullyQualifiedName~Lambda|FullyQualifiedName~Local|FullyQualifiedName~Generic"`: 1566/1566.
- `dotnet test test/Core.Tests --filter "FullyQualifiedName~Issue4219|FullyQualifiedName~Lambda|FullyQualifiedName~Local|FullyQualifiedName~Closure|FullyQualifiedName~Capture"`: 770/770.
- Manually re-ran the original crash repro plus 3-member chain, mutual-recursion cycle, and nested-calls-nested combinations through the rebuilt `gsc` — all correct.

## One non-defect note for the record

Both the original reconciliation loop and my extension widen `CapturedVariables` after the GS0219 ref-struct-capture check has already run at bind time. This is harmless today — the callee that directly captures a ref-struct already fails GS0219 regardless — but worth flagging as an ordering assumption for whoever touches this next.

I did not run the full cs2gs self-hosting guard locally (consistent with this repo's convention that CI is authoritative there); nothing in this fix touches cs2gs.

Note: I'm the PR author, so I can't `--approve` this myself (GitHub rejects self-approval) — filing this as a review comment instead. With the gap above fixed and pushed, I consider this ready to merge pending CI.

Issue #1900, opened 2026-07-03
cs2gs: `ref` locals/returns (RefExpression/RefType) have no mapping


DavidObando:
`ref int r = ref xs[1];`, `ref Middle(data)`, `return ref values[1];` all report `CS2GS-GAP: expression 'RefExpression' has no canonical G# form yet`. (grid apps G09/G12).
## Definition of done
- [ ] Fix + `Issue<N>*.cs` regression test
- [ ] Re-enable the quarantined grid fixture (or add one) and re-capture its baseline
- [ ] `tools/cs2gs/coverage/csharp-construct-inventory.json` row updated

_Discovered by the ADR-0138 conformance-grid build-out (gsc 0.2.585+01ad6ec964); quarantined sources live next to each grid app under `tools/cs2gs/corpus/grid/`._

Issue #1361, opened 2026-06-28
Split monolithic source files into smaller, focused files


DavidObando:
## Summary

Several source files have grown to extreme sizes, making them difficult to navigate, review diffs for, and maintain. These files should be split into smaller, focused files using partial classes or separate class files.

### 1. \src/Core/CodeAnalysis/Emit/ReflectionMetadataEmitter.cs\ (9,035 lines)

**Suggested splits**: Partition by emit phase or type kind:
- \ReflectionMetadataEmitter.Classes.cs\ — class/struct emission
- \ReflectionMetadataEmitter.Methods.cs\ — method body emission  
- \ReflectionMetadataEmitter.Metadata.cs\ — metadata table emission
- \ReflectionMetadataEmitter.Pdb.cs\ — PDB emission
- \ReflectionMetadataEmitter.Resources.cs\ — resource emission
- \ReflectionMetadataEmitter.Helpers.cs\ — shared utilities

### 2. \src/Core/CodeAnalysis/Syntax/Parser.cs\ (8,721 lines)

**Suggested splits**: Partition by parse target:
- \Parser.Declarations.cs\ — type, function, variable declaration parsing
- \Parser.Expressions.cs\ — expression parsing
- \Parser.Statements.cs\ — statement parsing
- \Parser.Types.cs\ — type syntax parsing
- \Parser.Patterns.cs\ — pattern parsing
- \Parser.Utilities.cs\ — error recovery, token consumption helpers

### 3. \src/Core/CodeAnalysis/Binding/DeclarationBinder.cs\ (7,029 lines)

**Suggested splits**: Partition by declaration kind:
- \DeclarationBinder.Classes.cs\ — class/record/struct binding
- \DeclarationBinder.Interfaces.cs\ — interface binding
- \DeclarationBinder.Functions.cs\ — function binding
- \DeclarationBinder.Variables.cs\ — variable/field binding
- \DeclarationBinder.Enums.cs\ — enum binding

### 4. \src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.cs\ (4,453 lines)

**Suggested splits**:
- \ExpressionBinder.Calls.Regular.cs\ — regular method calls
- \ExpressionBinder.Calls.Constructors.cs\ — constructor/object creation
- \ExpressionBinder.Calls.Operators.cs\ — operator overload resolution
- \ExpressionBinder.Calls.Delegates.cs\ — delegate invocation

[...]

PR #4263, opened 2026-09-16
Make ref-returning calls and properties usable as storage, with alias lifetime propagation

### LINKED ISSUE #4224: gsc: make ref-returning calls and properties usable as storage, with alias lifetime propagation
## Capability gap

Make ref-returning **named methods, properties and indexers usable as storage locations from G#**, not merely as sources of copied values or APIs exported for C# consumers.

This also requires preserving the referent's lifetime through ref aliases. None of the repros below uses a function literal or mutual recursion.

Follow-up to #4219 and the remaining consumer-side limits of #490, #491, #1900, #1987 and #3879. #3879's writable-ref declarations are implemented; this issue is about consuming and forwarding their storage identity. Do not reopen those historical issues as if their delivered subsets were absent.

## Reproduced behavior

Built gsc `0.4.635+07d64f9d8c` from `07d64f9d`, SDK `10.0.400`.

### Alias a named ref-returning call

```gsharp
package P
func At(values []int32, index int32) ref int32 {
    return ref values[index]
}
func Run() {
    var values = []int32{10}
    var ref alias = At(values, 0)
    alias = 42
    Console.WriteLine(values[0])
}
Run()
```

Actual: **GS0256**, the alias initializer must be an lvalue.

Changing only the initializer to `values[0]` compiles, mutates the original element and prints **42**. Calling `At(values, 0)` in ordinary value position also works.

### Forward an existing ref-returning call

```gsharp
func Forward(values []int32) ref int32 {
    return ref At(values, 0)
}
```

Actual: **GS0253**, the operand of `return ref` must be an lvalue.

### Forward an alias to caller-owned storage

```gsharp
package P
func Forward(ref value int32) ref int32 {
    var ref alias = value
    return ref alias
}
var value = 42
Console.WriteLine(Forward(ref value))
```

Actual: **GS0254**, reported as returning a pointer to function-local storage. The storage is actually the caller's `value`; only the alias slot is local.

### Ref properties and assignment are affected too

Using this implemented declaration:

```gsharp
class Holder {
    var slot int32 = 10
    prop Value ref int32 { get { return ref slot } }
}
```

Measured results:

| Use | Result |
|---|---|
| `Console.WriteLine(holder.Value)` | Compiles and prints 10 |
| `var ref alias = holder.Value` | GS0256 |
| `holder.Value = 42` | GS0127, property treated as read-only because there is no setter |
| `At(values, 0) = 42` | GS0005 at `=`, parser does not accept this assignment target |

The analogous C# operations compiled under C# 13. A driver initialized an array to 10, obtained a ref through a forwarding method, assigned 40, incremented through a ref-returning call, and incremented through a forwarded ref alias; the original array element ended at **42**.

## Why this matters

G# can produce byref APIs but cannot naturally consume their principal behavior from G#. This blocks zero-copy/in-place accessors, wrappers around ref-returning methods, and native G# alias-identity tests. Replacing the return with a value copy is not equivalent; forcing the assertion into a C# driver proves interop but leaves the language gap.

An ordinary setter is not an equivalent replacement for a ref-returning getter: an alias can be retained or passed onward, and the getter's receiver/index expressions must be evaluated once at the time the alias is obtained.

## Root cause

1. `ExpressionBinder.IsLvalue` accepts variables, fields, indexes and dereferences, not ref-producing call/property nodes.
2. `StatementBinder.IsLvalueForRefReturn` has a separate similarly restricted classifier.
3. `BindRefAliasLocalDeclaration` marks every alias `IsScoped = true`; `HasFunctionLocalRefScope` treats local-variable aliases as local storage instead of propagating the initializer's ref-safe scope.
4. Source-method/property reads use emitter-side auto-dereference. Address-taking/forwarding must preserve the returned managed pointer instead of taking the address of a copied value or loading through it too early.
5. Property assignment normally follows setter availability; writable-ref getters need a distinct valid store-through path. There is existing imported span/ref-indexer write-through machinery to reuse.
6. cs2gs explicitly rejects a `RefExpression` over a call result.

Sources at `07d64f9d`:

- `src/Core/CodeAnalysis/Binding/ExpressionBinder.cs:1518-1536`
- `src/Core/CodeAnalysis/Binding/StatementBinder.Narrowing.cs:1840-1979`
- `src/Core/CodeAnalysis/Binding/StatementBinder.Jumps.cs:532-642`
- `src/Core/CodeAnalysis/Emit/MethodBodyEmitter.MemberAccess.cs:2241-2269`
- `src/Core/CodeAnalysis/Binding/ExpressionBinder.Access.MemberLookup.cs:2275-2310`
- `tools/cs2gs/Cs2Gs.Translator/CSharpToGSharpTranslator.Types.cs:449-486`
- `tools/cs2gs/Cs2Gs.Tests/Issue1900RefLocalsTranslationTests.cs:148-184`
- `docs/adr/0060-ref-out-in-parameters.md:358-380`

## Recommended implementation path

1. Establish shared semantics for a ref-producing expression: referent type, writable/readonly capability, and ref-safe lifetime. Avoid independently teaching several classifiers incompatible subsets.
2. Preserve a raw address in alias initialization, `ref` arguments and `return ref`; load through it only in value context. Reuse the relevant existing bound nodes/helpers rather than introduce a second pointer runtime.
3. Derive an alias's ref-safe scope from its initializer. Distinguish an alias to caller/heap storage from a true stack local.
4. Use callable contracts and argument scopes to compute ref-call result escape scope. A ref-return flag alone must not authorize returning a reference rooted in a body-local argument.
5. Support store-through assignment and, subsequently, compound assignment/increment through writable-ref calls/properties/indexers. Evaluate receivers and arguments once and preserve exception ordering.
6. Wire source, imported, generic and delegate calls consistently; delegate/literal forms compose with #4219. Readonly reference support must restrict writes rather than erase ref identity.
7. Remove cs2gs gap paths only once native G# behavior is proven.

**Do not relax GS0254 or declare every call an lvalue.** C# also rejects returning references to body-local storage. Fix the lifetime model and preserve specific errors for invalid escapes.

## Definition of done

- [ ] Alias initialization from named ref-returning calls and properties mutates original storage.
- [ ] `return ref At(...)` and the caller-owned alias-forwarding example compile and preserve identity.
- [ ] Reads still yield values; ref contexts retain the original address.
- [ ] Writable-ref getter assignment works without inventing a setter; readonly getters remain protected.
- [ ] Cover class fields, array elements, struct receivers, imported APIs, generic calls, nested aliases and side-effecting receiver/index expressions.
- [ ] Reject stack-local escapes and invalid scoped/readonly conversions, including escapes through forwarding calls.
- [ ] Verify original-storage mutation, single evaluation, exception order, metadata and IL validity.
- [ ] Add native G# regressions alongside the existing C#-consumer interop tests; strengthen the currently loud-gap cs2gs fixtures.

Reference: C# ref-safe scope and reference variables:
`https://learn.microsoft.com/en-us/dotnet/csharp/advanced-topics/performance/`.

DavidObando:
Fixes #4224

## Root cause

A call to a native ref-returning function/method, or a read of a native ref-returning property/indexer, was never recognized as an lvalue: `ExpressionBinder.IsLvalue` and `StatementBinder.IsLvalueForRefReturn` accepted only a variable, field, array element, or `*p` dereference. An imported/CLR ref-returning member had no such gap — it is already wrapped in a `BoundDereferenceExpression` at bind time (`ConversionClassifier.AutoDereferenceRefReturn`), which was already an lvalue — so this was specifically a native-call/property asymmetry.

## Fix

- `RefCapabilities.IsNativeRefReturningCall` / `TryGetRefReturnEscapeSources` are the shared classifiers `ExpressionBinder.IsLvalue`, `StatementBinder.IsLvalueForRefReturn`, and `StatementBinder.HasFunctionLocalRefScope` all consume, so a native `BoundCallExpression`/`BoundUserInstanceCallExpression`/`BoundPropertyAccessExpression` with `ReturnRefKind != None` is a valid alias/`return ref` source, with its caller-side ref-safe-scope computed from its instance receiver and its non-`scoped` ref/in/out argument storage — mirroring C#'s ref-safe-scope rule for a ref-returning call. A genuine escape (`return ref Fwd(ref x)` over a function-local `x`) still reports `GS0254`.
- `MethodBodyEmitter.EmitAddressOf` gained cases for these node kinds (calling `EmitCallExpressionCore`/`EmitUserInstanceCall`/`EmitPropertyAccess(..., addressOnly: true)` — all extracted/adjusted to leave the callee's managed pointer on the stack without the usual dereference), and `EmitInstanceReceiver` routes an addressable ref-returning receiver through it directly instead of the rvalue spill-to-temp path (which never had to handle this receiver shape before, since it was unreachable prior to this fix).
- A writable (`ref`, not `ref readonly`) getter with no setter now supports assignment without inventing one — `holder.Value = 42`, a chained `GetHolder().Value = v` (single receiver evaluation), and a native indexer (`box[i] = v`) all store through the getter, reusing the existing imported-indexer write-through shape (`BoundIndirectAssignmentExpression` over `BoundAddressOfExpression`). A struct receiver needing a defensive copy is rejected for both the alias and the assignment form; a `ref readonly` view remains allowed.
- cs2gs's `ref expr` translator no longer refuses an `InvocationExpressionSyntax` operand outright — it translates it like any other lvalue and lets gsc's binder validate (issue #1900's own deferred limit).

## What was tested

- `dotnet build` (Release), full solution: 0 warnings, 0 errors.
- New `test/Core.Tests/CodeAnalysis/Binding/Issue4224RefReturningCallStorageTests.cs` (16 tests): alias from a top-level ref-returning function call and from a ref-returning property/indexer, mutating original storage; `return ref` forwarding a ref-returning call and a caller-owned alias through a `ref` parameter; rejection of a genuine function-local escape through a forwarding call (`GS0254`, two shapes); a generic ref-returning function; the pre-existing imported `Span[T]` ref-indexer alias (no new code, proves the "imported APIs" coverage); property/indexer write-through including a side-effecting chained receiver (single evaluation asserted); a `ref readonly` getter staying protected (`GS0127`); a struct receiver needing a defensive copy rejected for both alias (`GS9005`) and assignment (`GS0499`), with a `let ref readonly` view still permitted; a plain struct receiver mutating correctly.
- Strengthened `Issue4220ReadOnlyRefTests.DeferredSourceRefResultAliasesNeverBecomeSnapshots` (renamed `SourceRefResultAliasesObserveMutation_NeverBecomeSnapshots`): it previously asserted the GS0256 gap this issue closes; it now proves the six forms compile and are genuine live references (mutating the source after aliasing and observing the mutation through the alias), not snapshots.
- Strengthened `Issue1900RefLocalsTranslationTests`'s two "stays loud gap" tests (renamed "no longer gaps"): cs2gs now translates both shapes and gsc binds them.
- Regression sweep, all green: `Core.Tests` full suite (1315 tests) and a `Ref|Property|Indexer|Issue490|Issue491|Issue3879|Issue4220|Issue4222|Issue1900|Issue4224`-filtered `Compiler.Tests` run (1072 tests); `Cs2Gs.Tests` Issue1900 suite (9 tests).
- Hot-core cs2gs self-hosting translation guard (`build/run-cs2gs-selfmig-pr-guard.sh`): `src/Core.csproj` — the project containing every changed file — translated, compiled, ilverified, and passed test-parity, confirmed against a near-final snapshot (missing three small follow-up patches made afterward: the `EmitInstanceReceiver` addressable-receiver fix, the native-indexer write-through, and the readonly-struct-receiver `IsReadOnlyStorage` check — all in already-guarded files). A second full run hit an unrelated infrastructure collision (a concurrent guard run from another session on the same machine clobbering the shared `/tmp/gsharp-cs2gs-pr-guard` directory mid-run, evidenced by a bare "path not found" I/O error, not a diagnostic). **CI's own `hot-core-translation-guard` job should re-run the full guard against the final commit to confirm the downstream projects past `src/Core.csproj` are unaffected** — locally confirmed clean was only the stage containing every file this PR touches.

## Explicitly deferred (not in this PR)

- **A bare ref-returning call as a direct assignment target** (`At(values, 0) = 42`) is a parser-level generalization (no existing syntax production accepts a call expression as an assignment LHS) that is not enumerated in the issue's Definition of Done — which asks for "writable-ref GETTER assignment" (satisfied via the property/indexer path above) — and the issue's own recommended path calls compound assignment through such a call "subsequent". Aliasing the call result and assigning through the alias (`var ref alias = At(values, 0); alias = 42`) is fully supported and is exactly the workaround the issue's own repro used.
- Ref-returning dispatch through `BoundBaseInterfaceCallExpression`, `BoundBaseClassCallExpression`, and `BoundConstrainedStaticCallExpression` (base/interface/constrained-generic dispatch of a ref-returning member) is not covered by the new lvalue/escape-scope cases — not exercised by the issue's repros or explicitly required by the Definition of Done, and adding it without a concrete test would be undirected surface area.
- A pre-existing, unrelated bug was found and left unfixed (out of scope for this issue): a native STRUCT's own instance member cannot `return ref` its own scalar field at all (`struct Holder { var slot int32 = 10; func GetSlot() ref int32 { return ref slot } }` reports `GS0253` regardless of `this`/no-`this`), reproducing identically on a plain method with no ref-returning-call/property involvement at all — confirmed orthogonal to this diff. The struct-receiver tests in this PR route the getter's `return ref` through an array-backed field instead.


DavidObando (commented):
Rubber-duck review against issue #4224's scope only. Verdict: sound, no blocking defect found. Nothing pushed.

## What I checked

- Read #4224 in full (repros, root causes, DoD) and this PR's diff in full (`gh pr diff 4263`).
- Traced `RefCapabilities.IsNativeRefReturningCall` / `TryGetRefReturnEscapeSources` / `SelectByRefArguments` and their consumers (`ExpressionBinder.IsLvalue`, `StatementBinder.IsLvalueForRefReturn`, `HasFunctionLocalRefScope`) by hand against C#'s ref-safe-scope rules: receiver check skips reference types (heap-safe), recurses into struct receivers via the `default:` branch (which composes correctly through chained ref-returning calls/properties — verified this is what makes nested chains sound), `scoped` ref/in/out args are correctly excluded (symmetric with the callee's own rejection of returning a `scoped` parameter), and unrecognized argument shapes fall back to the pre-existing conservative "unsafe" default rather than silently permitting escape.
- Built gsc from `b17d292` (the exact PR HEAD, in the already-checked-out worktree) and ran adversarial repros beyond the PR's own 16 tests:
  - Function returning ref to its own local, then aliased at the call site → correctly rejected `GS0254`.
  - Chain of two ref-returning calls where the inner call's ref is only safe within a narrower (function-local struct) scope than the outer caller assumes → correctly rejected `GS0254`.
  - Generic struct (`Box[T]`) ref-returning method called on a function-local generic struct instance, then forwarded → correctly rejected `GS0254`; the same shape over a `ref` parameter → correctly compiles and mutates original storage through the alias.
  - Ref-returning call inside a loop, alias re-declared and mutated each iteration → compiles, correctly mutates the backing array each iteration (no stale/retained-pointer issue).
  - A mutating instance method called *through* a writable ref-returning property receiver (`h.Value.Inc()`, not just `h.Value = x`) → compiles and correctly mutates the original field twice — good extra evidence for the disclosed `EmitInstanceReceiver` crash fix, since this shape isn't in the 16 new tests.
  - Same shape through a `ref readonly` receiver and through a `let ref readonly` alias → compiles but the mutation is silently lost on a defensive copy, exactly matching C#'s own (surprising but memory-safe) behavior for a mutating call on a readonly ref. No unsoundness — original storage is never touched.
  - A by-value `Span[int32]` parameter forwarded through a ref-returning call (`return ref s[0]`) → rejected `GS0254` by a *pre-existing*, unrelated-to-this-PR conservative rule in `HasFunctionLocalRefScope` (predates #4224, from the #1361 file split / original issue #490) that treats every by-value parameter as function-local regardless of byref-like-ness. This means the PR description's claim to fully "mirror C#'s ref-safe-scope rule for a ref-returning call" is slightly overstated for this one shape, but it's not reachable as a false negative through this PR's new code — the callee can't even be defined. Filed separately: #4265 (not blocking).
- Ran the actual test suites on the branch: `Issue4224RefReturningCallStorageTests` + `Issue4220ReadOnlyRefTests` (69 passed), `Cs2Gs.Tests` Issue1900 filter (9 passed). Confirmed the 16 new tests assert specific diagnostic IDs (`GS0254`, `GS0127`, `GS9005`, `GS0499`) against `DiagnosticDescriptors.cs`, not generic "compilation failed."
- Confirmed the two "explicitly deferred" items (bare-call assignment target; base/interface/constrained-generic ref dispatch) are genuinely not required by the DoD and fail safe (still rejected, not silently accepted) rather than being secretly load-bearing.
- Reproduced the disclosed "found and left alone" bug independently (`struct { func GetSlot() ref int32 { return ref slot } }` → `GS0253` on `return ref slot` even via implicit `this`) — confirmed real, pre-existing, and orthogonal to this diff; the PR's own struct-receiver tests correctly route around it via an array-backed field.
- Release notes bullet matches the style/detail of the other `gsc:`-shaped entries in the same file (e.g. the #4222 bullet directly above it).

## Follow-up filed (not fixed here, per scope)

#4265 — a native ref-returning function can't take a by-value byref-like (e.g. `Span[T]`) parameter and return a ref derived from it at all (over-conservative `HasFunctionLocalRefScope`, pre-dates this PR). Non-blocking: fails safe, not reachable as a false negative through this PR's new escape-source code.

No changes requested.

Issue #4219, opened 2026-09-14
gsc: C# capability parity for mutual-recursion groups and ref-returning function literals


DavidObando:
## Objective

Make G# more capable as a compiler in its own right by supporting mutual-recursion groups and ref-returning function literals, with a staged path toward C# local-function and ref-return parity.

This is the capability-focused follow-up to #4198. **Do not gate this work on residual `__local_` identifiers in self-migration.** The motivation is language expressiveness and semantic interoperability, not a readability counter.

The assessment below is based on compiler commit `07d64f9d` and the September 13, 2026 self-migration artifacts. It contains no effort or timeline estimates.

## Assessment summary

- **Non-capturing generic recursion groups have a useful existing foundation.** Generic local functions already bind as directly callable symbols, and self-recursion works. The missing mechanism is registering sibling signatures before binding their bodies.
- **Full generic local-function parity needs more than forward declarations.** Captures, mixed generic/non-generic groups, enclosing type parameters, method placement, and definite assignment need coherent treatment.
- **Ref-returning literals are an end-to-end semantic feature, not a parser-only change.** Callable identity, closure methods, delegate metadata, conversions, invocation and lifetime checks must all preserve ref intent.
- **Ref-return consumption belongs in the initial ref workstream.** Merely emitting a ref-returning delegate does not enable alias initialization or `return ref Other(...)`.
- **Reuse existing compiler infrastructure.** Build on direct generic calls, capture boxing, closure reification, synthesized delegates, and named-function ref-return machinery. A new runtime or a delegate encoding of unbound generic methods is not required.

### The remaining migration lift is unrelated

#4197 closed on September 11, 2026; its follow-up #4211 merged later that day. The subsequent September 13 nightly run `34759865620`, at `f24064f675` with gsc `0.4.634`, provides the requested post-fix measurement:

| Evidence | Observation |
|---|---|
| Gate job `103746154046` | `__local_=3 (ceiling 31)` |
| `cs2gs-selfmig-run` artifact `10320421881`, `metrics.json` | `"lifts": 3` |
| `cs2gs-selfmig-migrated` artifact `10319575524` | 3,942 `.gs` files; one emitted `__local_` function declaration found |
| Remaining helper | `__local_BindGlobalScope_FindTopLevelBaseIndex` |
| Original source | Non-generic, ordinary `int` return, optional `visibleNestedTypeNames = null` |

[...]

PR #4266, opened 2026-09-16
Issue #4223: reify enclosing type parameters for local functions

### LINKED ISSUE #4223: gsc: reify enclosing type parameters for generic and non-generic local functions (GS0468)

DavidObando:
Fixes #4223

## Root cause

`GS0468` was reported for *any* local function referencing an enclosing method/type's own type parameter, even when the local function captured no outer variables — a case where a working reification path either already existed (issue #2118, for a non-generic local) or could be extended to exist (for a generic-own-type-parameter local). The diagnostic was a stopgap from issues #1940/#2016 that never got upgraded into the actual capability.

## What changed

- **`UserTokenResolver.TryPromoteNonCapturingGenericLambda`** (issue #2118's promotion of a zero-capture local into a genuine generic method) is extended to also apply when the local already declares its own `[T, ...]` list: every referenced enclosing type parameter is cloned as an *additional* method type parameter appended after the local's own. `BuildMethodSpecForGenericCall` supplies the extra type arguments at call sites (they're the in-scope originals, never inferable from the call's own arguments).
- **Constraint-clause binding scope fix**: a generic local's own type-parameter list (`let Keep[T IComparable[U]] = ...`, `U` owned by the enclosing method) used to bind its constraints against an *empty* scope, reporting `GS0113 "Type 'U' doesn't exist"` before the reference even reached the enclosing-type-parameter machinery. Now seeded with the enclosing type parameters.
- **`LambdaEnclosingTypeParameterCollector` gap** (shared by this promotion and the pre-existing capturing-closure reification, issues #1477/#3933): it never checked a call's static owner type, a method group's owner type, a field/property access's declaring type, or a function-pointer's owner type — shapes `LambdaBinder.EnclosingTypeParameterReferenceWalker` already checked at bind time. A local function calling a `shared` member of the enclosing generic type by its unqualified name (`return Public()` → `Holder[Outer].Public()`) had its `Outer` reference missed entirely, producing `BadImageFormatException` at load.
- **Transitive promotion (fixed point)**: promoting a callee to carry an extra enclosing type parameter can require its *caller* — another zero-capture local in the same pass, textually unrelated to that parameter — to carry it too, purely to supply it at the call site. Missing this didn't crash; it silently produced the **wrong value at runtime** (the caller's call site fell back to whatever type parameter happened to share the same ordinal). `TryPromoteNonCapturingGenericLambda` now runs to a fixed point over every zero-capture local/lambda in a compilation unit.
- **Async state-machine remap fix**: `ReflectionMetadataEmitter.RegisterStateMachineEnclosingGenerics` built its outer-type-parameter remap keyed by the promoted local's own (cloned) type parameters, never consulting the enclosing-TP→clone-ordinal remap — the same fix `RegisterGeneratedGenericRemaps` already applies for the analogous iterator state machine (issue #810). Fixed by aliasing each original onto the same slot its clone occupies.
- Removed the now-redundant `CheckNonGenericLocalFunctionEnclosingTypeParameterReference` diagnostic and its constructor-plumbing (dead code once the zero-capture path is provably correct) — **reinstated in a narrower form** by the review fix below.

### Review fix: async zero-capture local, enclosing type parameter in its OWN parameter

Adversarial review (multiple type instantiations of every new capability, not just "compiles without GS0468") found that unconditionally removing the diagnostic above went one shape too far. A non-generic, zero-capture, **async** local function whose own PARAMETER type (not its return type or body) references an enclosing type parameter is emitted through a separate, pre-existing async "erased delegate" adapter — not through `RegisterStateMachineEnclosingGenerics`'s reification above — and that adapter's parameter-unboxing conversion has a confirmed defect for a value-typed instantiation:

```gsharp
func Outer[U](seed U) U {
    let Local = async func (x U) U { return x }
    return Local(seed).Result
}
Outer("hi")  // succeeds
Outer(42)    // NullReferenceException at the first call
```

This exact shape was unconditionally `GS0468` before #4223. `LambdaBinder.CheckAsyncNonGenericLocalFunctionEnclosingTypeParameterInParameter` restores the diagnostic for exactly this narrow combination (parameter-type reference + async + zero-capture + non-generic-own), matching the *pre-#4223* gate's footprint exactly — including its existing exclusion for a local nested inside a non-generic user type (that path routes through `ClosureEmitter`'s display-class code instead, and was never GS0468-gated either before or after this issue, so it is intentionally left alone). The arrow-lambda form (`async (x U) -> x`) hits the same gate. Return-type-only and body-only references, and the async GENERIC-own-type-parameter case, are unaffected and remain fully supported.

### Deliberately out of scope (still reports GS0468)

Two combinations remain unsupported:

1. A **capturing** generic local function referencing an enclosing type parameter in its own signature or an own-type-parameter constraint — that path is a different emission mechanism (a synthesized closure class, issue #4221/#4252) that hasn't been extended the same way. The issue text itself calls out composing this with capture support as a later concern. Confirmed by direct repro that this combination previously compiled clean and crashed (`TypeLoadException`/invalid IL).
2. A **non-generic, zero-capture, async** local function whose own parameter type references an enclosing type parameter (see "Review fix" above) — retained specifically because relaxing it would trade a compile-time diagnostic for a runtime crash on a previously-safe program.

### Found but explicitly not fixed (pre-existing, unrelated)

An async local function's erased-delegate adapter (used whenever a function-literal's declared type mentions an open type parameter) has a pre-existing defect in its parameter-unboxing conversion for a value-typed parameter. Reproduced identically on an already-legal **capturing** async lambda, and on a zero-capture async local nested inside a non-generic type — both confirmed on `main`, predating #4223 entirely, neither ever gated by `GS0468` — so it's unrelated to this issue. Not fixed here; regression tests for the async route were designed to avoid tripping it where the shape is otherwise legal (reifying via return type / `typeof` / body, not a value-typed parameter).

## What I tested

- `dotnet build` the full solution in Release: 0 warnings, 0 errors.
- Full **Core.Tests** suite (Release): **9299/9299 passed**.
- Full **Compiler.Tests** suite (Release): **5969/5969 passed**.
- Manual repro scripts covering: the issue's exact repros (generic-own-TP + enclosing, non-generic, arrow-lambda form), multi-instantiation runtime correctness (same program, different type arguments, verifying actual per-instantiation values — not just "doesn't crash"), an enclosing type parameter that is *not* ordinal 0, an own-type-parameter constraint referencing an enclosing parameter, two levels of local-function nesting, `typeof`/`is`/`sizeof`/generic-call/constrained-static-call reference shapes, sync + async + iterator routes verified independently, and the transitive-promotion case (caller/callee chain where only the callee needs the enclosing parameter).
- Targeted `Compiler.Tests` filters covering the existing generic-local-function/closure/capture/async/iterator machinery (issues #1886, #2118, #810, #1465, #1489, #3933, #1499, #4219) before and after: no regressions.
- **Independent review pass** (adversarial, multi-instantiation): both enclosing-type and enclosing-method type parameters in scope simultaneously; a local function composing captured outer state with a promoted enclosing type parameter transitively via a sibling call; nested local functions where only a grandparent generic type supplies the needed type parameter; constraint-violation rejection (`GS0152`) still fires correctly; the capturing-generic-with-enclosing-constraint diagnostic still fires correctly. This surfaced the async parameter-type gap above, which is now fixed and covered by new/updated tests in `Issue2016NonGenericLocalFunctionEnclosingTypeParameterTests` and `Issue2026GenericAsyncTaskWrapTests`. Full **Core.Tests** (1962/1962 on the relevant filter) and targeted **Compiler.Tests** (40/40) re-verified green after the fix.

### Hot-core self-hosting guard

Ran `build/run-cs2gs-selfmig-pr-guard.sh` locally. An earlier run hit transient build-output file contention (`GS9997` fatal I/O error, not a code diagnostic) from other concurrent `dotnet build`/`test` processes against the same worktree — not a real code issue. A clean re-run (after shutting down stray MSBuild/VBCSCompiler server processes) was in progress at the time of this PR; CI's `build.yml` guard job is the authoritative gate regardless, per repo convention.

## Not deferred

Every item in the issue's "Definition of done" checklist is covered by the changes and/or the new/updated tests, except the two explicitly-named GS0468 combinations described above, which are follow-up composition/adapter concerns rather than part of this issue's scope.


DavidObando (commented):
## Rubber-duck review against issue #4223 (strict scope)

**Verdict: solid, with one real gap found and fixed during review. Commenting rather than approving self-authored work.**

### Scope discipline (ponytail lens)
This correctly builds on #4252/#4261's closure machinery rather than duplicating it: `UserTokenResolver.TryPromoteNonCapturingGenericLambda` (issue #2118) is *extended*, not reimplemented, and the capturing-closure path (#4221/#4252's synthesized closure class) is deliberately left untouched — GS0468 stays for the capturing+enclosing-TP combination, which is the right call given that composing capture support with enclosing-TP reification is explicitly out of #4223's scope per the issue text.

### What I verified
- Full diff read, issue DoD checklist cross-checked item by item.
- Built `gsc` from the branch; ran adversarial multi-instantiation repros beyond the issue's minimal example:
  - Enclosing CLASS type parameter and enclosing METHOD type parameter both in scope simultaneously (`class Box[V] { func Run[W](...) }`), both non-generic-own and generic-own local shapes — correct per-instantiation values for multiple (V, W) pairs.
  - Composing #4221/#4261 outer-value capture with a *separately* promoted zero-capture sibling call (capturing caller → promoted zero-capture callee, transitive) — correct.
  - Three levels of local-function nesting where only the grandparent (containing generic class) supplies the needed type parameter — correct.
  - Constraint violations (`GS0152`) still correctly rejected after the constraint-scope fix.
  - The one documented unsupported combination (capturing generic local + enclosing-TP constraint/signature) still correctly reports `GS0468`.
  - cs2gs DoD item ("no longer needs to distort lexical structure"): `grep -rn "GS0468" tools/cs2gs` finds nothing — vacuously satisfied, no workaround to remove.
- Ran targeted `Compiler.Tests`/`Core.Tests` filters (Issue4223/1940/2016/2026/4219, plus a broad Lambda/LocalFunction/Closure/Async/Generic/Iterator filter — 1962 Core.Tests + 40 Compiler.Tests) and a full solution build: all green.

### Defect found and fixed (pushed to this branch, commit 837c6707)
Multi-instantiation testing of the ASYNC route (per the issue's own "verify async independently, don't infer from sync" requirement) found that unconditionally relaxing GS0468 opened one previously-safe-by-rejection shape onto a real crash: a non-generic, zero-capture, **async** local function whose own *parameter* type (not return type, not body) references an enclosing type parameter routes through a separate, pre-existing async "erased delegate" adapter whose parameter-unboxing has a confirmed defect for value types —

```gsharp
func Outer[U](seed U) U {
    let Local = async func (x U) U { return x }
    return Local(seed).Result
}
Outer("hi")  // fine
Outer(42)    // NullReferenceException
```

This exact shape was unconditionally `GS0468` before #4223 (nothing crashed pre-PR because it never compiled); the PR's blanket relaxation would have turned that compile-time safety net into a runtime crash for exactly the DoD's own warning ("relax GS0468 only for routes with proven valid emission"). Restored a narrow diagnostic (`LambdaBinder.CheckAsyncNonGenericLocalFunctionEnclosingTypeParameterInParameter`) for exactly this combination, carefully scoped so its rejection footprint is a **subset** of the pre-#4223 gate's (verified against `main`: same exclusion for a local nested in a non-generic user type, same arrow-lambda coverage). Return-type/body-only async references and the async generic-own-TP case are unaffected and remain fully supported — verified with both reference-type and value-type instantiations. Updated/added tests in `Issue2016NonGenericLocalFunctionEnclosingTypeParameterTests` and `Issue2026GenericAsyncTaskWrapTests`, and updated the release note + PR description accordingly.

### Separate issues to file (not fixed here, correctly out of scope)
1. **Pre-existing, unrelated defect**: the async erased-delegate adapter's value-type parameter-unboxing crash is reachable on `main` today via an already-legal *capturing* async local (`let Local = async func(x U) U {...}` with a captured variable) and via a zero-capture async local nested inside a non-generic type — neither was ever GS0468-gated, before or after this PR. Worth its own issue; not touched here per scope.
2. **Pre-existing, unrelated limitation**: a named *generic* local function cannot be referenced as a delegate value at all (`let d = Keep` / `let d Func[...] = Keep` → `GS0126 'Keep' is not a variable`), regardless of whether it references any enclosing type parameter — confirmed identical on a plain generic local with zero enclosing references. Out of #4223's scope.
3. **Minor, non-blocking**: the issue's DoD item "verify ... closed delegate use" has no dedicated repo test for the non-generic enclosing-TP case (`let d Func[U,U] = Keep` compiles and runs correctly — verified manually — but isn't pinned by a test). Not blocking; flagging for a follow-up test addition.

### Current status
`mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` (pending CI on the new commit / review approval — no conflicts with `main`).

Issue #4236, opened 2026-09-14
selfmig: Compiler.Tests channel matrix hits InvalidProgramException in ReceiveValueAsync


DavidObando:
## Exact failing result

On source main `b9cd4a4fe17889b4768ee39b6893335b83ea5971`, the complete migrated Compiler.Tests suite finishes with **5855 passed, 1 failed, 0 skipped, 5856 total**. The one failure is `Issue2965ChannelElementSlotTests.ChannelElementMatrix_LoadsVerifiesAndRuns`:

```text
Gsharp.Concurrency.ScopeException: One or more goroutines owned by a scope failed
 ---> System.InvalidProgramException: Common Language Runtime detected an invalid program.
   at Gsharp.Concurrency.ChannelOps.ReceiveValueAsync[T](Channel`1 channel, CancellationToken cancellationToken)
   at Issue2965Matrix.<Program>.delayedReceive(Channel`1 ch, Context <>ctx)
```

This is a completed failing suite, not a timeout: the parity budget was 90 minutes and execution completed in 3759 seconds. The project itself passed compile (1836.6 seconds) and ILVerify before parity ran.

- Exact validation run: `2026-09-14T21-01-55Z_5cd7fd`.
- gsc: `0.4.686+b9cd4a4fe1`; .NET SDK `10.0.400`; ILVerify `10.0.8`.
- Fingerprint: `sha256:3dc53ea4f337f3a5ff3d6b1ef62863d4ed15c75246d6fefdbf1ce2eda3d1f142`.

## Controls and reproduction

The exact native Debug test passes **1/1**, zero skips; its project build has zero warnings/errors:

```bash
dotnet build test/Compiler.Tests/Compiler.Tests.csproj -c Debug --no-restore -graph
dotnet test test/Compiler.Tests/Compiler.Tests.csproj -c Debug --no-build --no-restore \
  --filter FullyQualifiedName~Issue2965ChannelElementSlotTests.ChannelElementMatrix_LoadsVerifiesAndRuns
```

After full-corpus translation, validate only the app while retaining full discovery:

```bash
bash build/run-cs2gs-selfmig-migrate.sh
bash build/run-cs2gs-selfmig-validate.sh compiler-channel \
  test/Compiler.Tests/Compiler.Tests.csproj
```

An explicit one-test rerun against the produced migrated assembly reproduces the same failure (0 passed, 1 failed, exit 1):

```bash
[...]

PR #4232, opened 2026-09-14
selfmig: close #4198 measurement scope and bank three retained lift uses

DavidObando:
## Decision
Refs #4198 and #4197. Fresh full-corpus translation on b9cd4a4fe17889b4768ee39b6893335b83ea5971: **56/56 projects, 3,911 .gs files, three counted __local_ occurrences, one ordinary non-recursive default-parameter helper**. No generic/ref-returning residual justifies language implementation. Independent September 13 and 14 full nightlies agree. Capability work remains #4219.

## Changes
- Record complete residual catalogue, caller/lowering trace, exact commands, and prior-run evidence in docs/self-migration-4198.md.
- Tighten only liftedLocalCeiling: **31 → 3**. All other ceilings, floors, app identities, exclusions, and allowlists unchanged.
- Add focused shared-counter/gate regression: declaration plus two calls passes; fourth code occurrence fails. Comment is not counted.
- No compiler/translator/parser/emitter changes.

## Verified
- Release CLI and test-project builds: zero warnings/errors after locked solution restore.
- Complete post-fix translation: run 2026-09-14T19-58-29Z_a6ced4, 56/56 passed; labels 0, lifts 3, reducible long lines 1. Pre-polish assertions 22,188 are not claimed as gate-parity counts.
- Native Cs2Gs.Tests: 3,067/3,067 passed, zero skips.
- Exact Core source compile, ILVerify, and migrated test-parity stages passed.
- GitHub CI passed all required build, test, translation, compile, ILVerify, e2e, extension, and hot-core guards.
- Independent rubber-duck review found no in-scope findings.
- Core.Tests MethodAccess ILVerify and Extensions.Tests migrated TypeLoadException were reproduced and filed separately as #4233 and #4234; no unrelated fixes or false-green changes included.
- Shell syntax and diff checks pass.

Ready to merge.

Issue #3750, opened 2026-09-01
cs2gs: references to an elided static extension-holder class dangle (GS0113, from #3745)


DavidObando:
Split out of #3745 (family F6). Found in the migrated `test/Compiler.Tests` compile wall; 1 of that app's 20 errors, and the only cs2gs-layer defect in the set.

cs2gs correctly lowers a C# static extension-method **holder class** to top-level G# receiver-clause funcs and elides the class itself — but any reference to the holder class *by name* is left behind, dangling.

## Evidence

`test/Compiler.Tests/Emit/ClrInteropEmitTests.cs:326`:

```csharp
public static class NamedParamsExtensionFixture
{
    public static string Describe(
        this string source,
        int statusCode,
        Type responseType = null,
        string contentType = null,
        params string[] contentTypes) => …;
}
```

is translated (correctly) to the tail of `Emit/ClrInteropEmitTests.gs`:

```gs
func (source string) Describe(
    statusCode int32,
    responseType Type? = nil,
    contentType string? = nil,
    contentTypes ...string) string { … }
```

…with no `NamedParamsExtensionFixture` anywhere in the migrated tree. But `ClrInteropEmitTests.cs:246` still names it:

```csharp
CompileAndRun(source, typeof(NamedParamsExtensionFixture).Assembly.Location));
```

which migrates verbatim to `Emit/ClrInteropEmitTests.gs:192,42` and reports:

```
GS0113: Type 'NamedParamsExtensionFixture' doesn't exist.
```

## Shape of a fix

Whenever cs2gs elides a static extension-holder class, references to that type (`typeof(T)`, `nameof(T)`, static-member qualification, `T.Something`) must be rewritten to something that still resolves. `typeof(Holder).Assembly.Location` in particular only wants the *declaring assembly*, so rewriting it to any surviving type in the same translated project would preserve the semantics; a general rule is to keep the holder class when it is referenced by name rather than eliding it unconditionally.

Not a regression — this is not one of the families solved so far.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Nng28yiBdPVdeML7mSZphs

PR #4267, opened 2026-09-16
selfmig: preserve extension-method owner-type identity through self-hosted emit

### LINKED ISSUE #4234: selfmig: ChannelExtensions owner-type mismatch breaks Extensions and Runtime.Channels tests
## Confirmed exact-migration failure

On main source `b9cd4a4fe17889b4768ee39b6893335b83ea5971`, exact `test/Extensions.Tests/Extensions.Tests.csproj` migration passes translate, compile, and ILVerify, then its test host aborts:

```text
System.TypeLoadException: Could not load type Gsharp.Concurrency.ChannelExtensions
from assembly Gsharp.Runtime.Channels, Version=0.4.686.0.
   at Gsharp.Concurrency.<Program>.mergeInto[T](Channel`1 merged, Channel`1[] inputs, Context <>ctx)
```

The exception escapes a free goroutine, so the existing ADR-0174 fail-fast behavior terminates the test host. **Do not treat the partial Passed! line as a green suite**: the pipeline run reported 53 passed before the abort, and an isolated rerun reported 63, both below the original 117 declared Fact-method lower bound and neither a completed run. The gate correctly classifies this as test-parity failure. This app is already banked in `greenApps`.

## Controls / reproducibility

- Native Debug `Extensions.Tests`: **180 passed, 0 failed, 0 skipped**, 2 seconds; build has zero warnings/errors.
- Exact local validation run `2026-09-14T21-02-22Z_783bd4`, gsc `0.4.686+b9cd4a4fe1`, .NET SDK `10.0.400`, ILVerify `10.0.8`: compile passed in 859.1 seconds, ILVerify passed, parity aborted in 3 seconds.
- Copying the complete built migrated test output directory and invoking `dotnet vstest GSharp.Extensions.Tests.dll` reproduces the same TypeLoadException with exit 1. That rerun does not share active shard output files.
- Source was translated in one complete 56-project pass, retaining canonical exclusions and linked-source checks; validation selected apps without changing discovery. Two complete pre-validation trees were byte-identical.
- Found while measuring #4198 / PR #4232. That branch has no C#, G#, project, compiler, or translator changes. This is not a generic/ref-returning mutual-recursion residual.

## Bounded investigation evidence

The migrated `src/Sdk/Gsharp.Runtime.Channels/ChannelExtensions.gs` contains native G# extension declarations (`func (channel Channel[T]) Close[T]()`), not a CLR class named `ChannelExtensions`. Meanwhile the failing already-G# Extensions assembly contains a call expecting that C# owner type.

The mirrored `Gsharp.Extensions.csproj` uses the packaged `Gsharp.NET.Sdk/0.4.686-gb9cd4a4fe1`; its build-only runtime project-reference group has been stripped. The exact compile log also reports `MSB3243` conflicts between `Gsharp.Runtime.Channels, Version=0.4.686.0` and `Version=0.4.0.0`, choosing the former arbitrarily. These observations point to incompatible bootstrap/migrated runtime references or extension-owner metadata; the precise binding/packaging fix still needs investigation. Do not paper over it with an exception handler or by removing channel tests.

## Exact commands

After building the Release cs2gs CLI and setting a non-overlapping `SELFMIG_GATE_ROOT`:

```bash
bash build/run-cs2gs-selfmig-migrate.sh
bash build/run-cs2gs-selfmig-validate.sh extensions-runtime \
  test/Extensions.Tests/Extensions.Tests.csproj
```

Native control:

```bash
dotnet build test/Extensions.Tests/Extensions.Tests.csproj -c Debug --no-restore -graph
dotnet test test/Extensions.Tests/Extensions.Tests.csproj -c Debug --no-build --no-restore
```

Preserve the complete pipeline logs (`sdk.build.log`, `ilverify.log`, `test-parity.log`). The current parity fingerprint is explicitly **unclassified**, because a process abort supplies no complete failing-test identity list; fingerprint equality must not be used as proof of a shared cause.

## Acceptance

Keep the banked green app, ceilings, exclusions, test-count checks, and goroutine fail-fast semantics unchanged. Resolve the runtime metadata/reference mismatch, add a focused executing regression, and make the exact migrated Extensions suite complete all tests through compile, ILVerify, and parity. This is separate from the Core.Tests ILVerify blocker #4233 and the older Core.Tests parity work #4214. Part of #3501.

DavidObando:
Fixes #4234

## Root cause

cs2gs translates a C# extension method into a G# receiver-clause `func`, and a receiver-clause func only binds at top level (its receiver is not in scope inside a `shared { }` block). When a C# `static class` holds *only* extension methods, every member gets lifted to a top-level declaration, and the compiler's emitter always hosts a top-level extension function on the package's synthesized `<Program>` TypeDef. That is invisible to ordinary G# code, but not to a self-hosted build of the same method: compiling `src/Sdk/Gsharp.Runtime.Channels` through gsc re-emits native `Gsharp.Concurrency.ChannelExtensions::Close` as `Gsharp.Concurrency.<Program>::Close`, and native `ChannelBatchExtensions::ReceiveBatch` the same way — confirmed via the MemberRef diffs in the issue's comments. That breaks two things that key off the *exact declaring type*, not just the method's behavior:

- Reflection: `test/Runtime.Channels.Tests/ChannelOpsTests.cs` calls `typeof(ChannelExtensions).GetMethods()`.
- `RendezvousBatchAnalyzer`'s owner check (`BatchExtensionsTypeName = "Gsharp.Concurrency.ChannelBatchExtensions"`, `src/Core/CodeAnalysis/Binding/RendezvousBatchAnalyzer.cs:120`) — it only recognizes the batch-extension call shape against that literal type name, with no `<Program>` fallback (unlike the `chunks(...)` free-function check right next to it, which already has one because it's a genuinely top-level G#-authored function). That is exactly the `ARendezvousChannel_ReportsGS0562` / golden-test (`ChannelOwnership.gs`/`FanInMerge.gs`/`Pipeline.gs`) failure the issue's comments isolate.

cs2gs had a partial mitigation (`IsTypeOfReferenced`, issue #3750): it kept the enclosing static class alive, empty, when a `typeof(Holder)` reference survived translation *in the same file*. That only prevents `GS0113`/`GS0159` on a dangling name — it never re-attached the lifted method's `MethodDef` to that class, so the owner-identity break was unaffected, and it couldn't help at all when nothing in the translated file names the holder (the common case for `ChannelExtensions`).

## Fix

Reuses ADR-0053's existing static-owner machinery — no new binding/emission model, just a new way to reach it:

- **New compiler-intrinsic annotation `@ExtensionOwner(typeof(T))`**, recognized purely by name (parallel to `@SuppressDiagnostic`, not a real CLR attribute). On a top-level extension function it resolves `T` to a same-package, non-generic **class**, then does exactly what an ordinary shared-block static method (or the existing receiver-clause-operator path) already gets: sets `FunctionSymbol.StaticOwnerType` and adds the function to that class's `StaticMethods`, so the *existing* TypeDef-placement/row-planning machinery (`PlanClassMethods`, `aggregateMethodHandles`) hosts the `MethodDef` there instead of `<Program>`. Malformed use (missing `typeof(...)`, a generic type, a struct, a cross-package type) reports the new **GS9306**.
- **`TypeDefEmitter`** forces `Abstract | Sealed` on a class that hosts such a method — ECMA-334 §13.6.9's static-class shape, which is already how `<Program>` itself is always emitted, and is required both by C#/F# extension-method discovery and by gsc's own `MemberLookup.IsStaticClass` import check.
- **`ReflectionMetadataEmitter`** stamps `[ExtensionAttribute]` on that TypeDef, mirroring the existing `<Program>` stamp right next to it.
- **`Binder.cs`**: an `@ExtensionOwner`-routed function is *also* registered as an ordinary top-level extension (needed so `x.Close()` still resolves at call sites) and therefore already gets body-bound by the top-level-function pass; the separate shared-block-statics body-binding loop now skips `IsExtension` methods instead of double-binding them into `functionBodies` (which threw an `ImmutableDictionary` duplicate-key `ArgumentException`, GS9998, before this guard was added).
- **cs2gs** no longer drops a static extension class once it has no members left. It stays, empty, and each lifted method now carries `@ExtensionOwner(typeof(OriginalClass))` pointing back to it — regardless of whether anything in the *same translation unit* names the holder, since a single file can never rule out a downstream self-hosted consumer needing the identity. This makes the old `#3750` typeof-only elision unreachable for this shape (the tests it added now assert survival instead of elision — see below).

## Two corpus-caught collisions, fixed in the second commit

An unconditional "always tag the lifted func" first attempt broke two existing corner cases, both caught by the full `Cs2Gs.Tests` run (not by the hot-core PR guard, which doesn't touch these fixtures):

1. **Issue #3413's private-nested-aggregate owners.** When a static extension class's body needs access to a `private` nested type it declares, cs2gs already keeps the method's real implementation as an in-owner shared-block method of the *same name* and emits the lifted top-level func as a thin forwarding wrapper. Tagging that wrapper too routed a *second*, identically-named/signed `MethodDef` onto the same owner — a genuine duplicate (`AmbiguousMatchException` on `GetMethod`), not a fix, since that owner already keeps its identity by construction. Fixed by skipping the tag whenever `RequiresOwnerScopedExtension` is true for the method.
2. **A C#-`@`-escaped owner name.** `public static class @select` (escaping a G# contextual keyword collision) rendered as `@ExtensionOwner(typeof(@select))` — `node.Identifier.Text` carries C#'s verbatim `@` marker through, which isn't valid G# syntax in this position, while the class's own declaration was correctly sanitized to G#'s own escaping (`$select`). Fixed by reusing `this.EmittedName(symbol, ...)` — the same name-allocator call the owner's own `TypeDeclaration` is built from — instead of the raw syntax token text.

## What I verified

- `dotnet build -c Release`: 0 warnings, 0 errors (confirmed via full solution build, several times, at each stage).
- **On the actual migrated `Gsharp.Runtime.Channels.dll` produced by the self-hosting guard** (not just a toy repro), reflected over the output assembly:
  - `Gsharp.Concurrency.ChannelExtensions`: `IsAbstract=True IsSealed=True ExtensionAttr=True`, exactly two methods, `Close[T](Channel<T>)` and `Close[T](ChannelWriter<T>)`, both `ExtensionAttr=True` — byte-for-byte the native C# shape, and the exact MemberRef the issue names (`Gsharp.Concurrency.ChannelExtensions::Close`).
  - `Gsharp.Concurrency.ChannelBatchExtensions`: same Abstract/Sealed/ExtensionAttribute shape, with `TryReceiveBatch`, `TrySendBatch`, `ReceiveBatch`, `SendBatch` all `ExtensionAttr=True` on it (not `<Program>`) — exactly what `RendezvousBatchAnalyzer`'s `BatchExtensionsTypeName` check needs; the two private `*ForeignAsync` helpers correctly stay non-extension, unaffected.
  - The emitted `src/Sdk/Gsharp.Runtime.Channels/ChannelExtensions.gs` shows the mechanism in place: `class ChannelExtensions { }` followed by `@ExtensionOwner(typeof(ChannelExtensions))` on each lifted `Close` func.
  - `test/Runtime.Channels.Tests/ChannelOpsTests.cs:185` (`Close_Extension_BindsOnChannelAndWriter_NotOnReader`) asserts `typeof(ChannelExtensions).GetMethods().FirstOrDefault(m => m.Name == "Close" && m.GetParameters()[0].ParameterType.Name == "ChannelReader`1")` is `null` — satisfied, since the two real methods take `Channel<T>`/`ChannelWriter<T>`, never `ChannelReader<T>`.
  - Also compiled a minimal standalone `.gs` file directly with the built `gsc.dll` for a fast red/green check outside the hour-long corpus (`class ChannelExtensions {}` + `@ExtensionOwner(typeof(ChannelExtensions)) func (sb StringBuilder) Shout() ...`): before the fix `Shout` sat on `<Program>`; after, it's on `ChannelExtensions` with the same Abstract/Sealed/ExtensionAttribute shape, and `ILVerify` (10.0.8) passes clean on the output.
- Malformed-use repro: `@ExtensionOwner("not-a-type")` reports `GS9306: '@ExtensionOwner' requires a single 'typeof(T)' argument naming the CLR owner type.` at the right location.
- `dotnet test test/Core.Tests` (`--filter FullyQualifiedName~Extension` and `~DiagnosticIdUniqueness`): 193 passed, 0 failed — includes the new `Issue4234ExtensionOwnerEmitTests` (owner routing + two GS9306 misuse shapes) and confirms `docs/diagnostics.md`/`website/docs/ref/diagnostics.md` stay in sync with `DiagnosticDescriptors`.
- `dotnet test tools/cs2gs/Cs2Gs.Tests` **full suite, twice**: the first full run (with only the first commit) caught the two collisions below as failures; after the second commit's fix, a complete rerun is **3160 passed, 0 failed, 0 skipped** — the whole cs2gs translator test corpus, not just the extension-related subset.
- `build/run-cs2gs-selfmig-pr-guard.sh` (covers `src/Sdk/Gsharp.Runtime.Channels` since #3933), run twice:
  - First run (before the two-collision fix below, and contaminated by a concurrent `dotnet build` of mine racing the guard's own NuGet package resolution near the end of its ~58-minute run): 7/8 hot-core apps green through translate/compile/ILVerify/test-parity — including `src/Sdk/Gsharp.Runtime.Channels` itself — the 8th (`Cs2Gs.Pipeline`) failed with `MSB4236: SDK ... could not be found`. That's an environment/package-cache race from my own concurrent build (the app's generated `.gsproj` ended up referencing a newer SDK nupkg version than the guard's own NuGet cache had at that point), not a code defect — the real code defects it was masking are the two collisions below, which the full `Cs2Gs.Tests` corpus (not this guard) actually caught and which are now fixed and confirmed by the all-green rerun of that suite above. (A same-session attempt to rebuild `Cs2Gs.Pipeline`'s migrated output standalone, outside the pipeline's own `dotnet build`, hit ~600 `GS0536: Redundant '!!'` diagnostics; that is not a real signal either way — `CompileStage`'s own `!!`-polish pass, which a standalone `dotnet build` bypasses, is what normally strips those, per the pipeline's own comments.)
  - Second run, from the two-commit fixed state, in an isolated `SELFMIG_PR_GUARD_ROOT`, with no concurrent builds this time: `InternalAnalyzers`, `Core`, `GSharp.Formatting`, and **`Gsharp.Runtime.Channels`** (the app this issue is about) all green through translate/compile/ILVerify/test-parity — this is the run the real-artifact reflection dump above came from. [`tools/cs2gs/Cs2Gs.CodeModel`/`Cs2Gs.Translator`/`Cs2Gs.ProjectLoading`/`Cs2Gs.Pipeline` were still finishing at PR-creation time; I'll post the final banner as a follow-up comment on this PR.]

## What was NOT changed / deferred, and why

- The `MSB3243` native-vs-migrated-runtime `Version=0.4.686.0` vs `Version=0.4.0.0` reference-resolution conflict the issue's body mentions is a separate packaging/reference-resolution concern (which DLL a project reference picks when both a native and a migrated `Gsharp.Runtime.Channels.dll` exist), not an owner-type-identity concern. Per the issue's own comment 1, a declared-`AssemblyVersion` fix for this had already landed by the time of that comment (on `c6ba1c17`, ahead of this branch's base). I did not touch it further — it's orthogonal to this issue's root cause.
- No release-notes precedent: I checked recent `selfmig:`-prefixed commits (`#4232`, `#3944`, …) and none touch `website/docs/release-notes.md` — that convention is for `gsc:`-prefixed user-facing language/diagnostic changes. Since `@ExtensionOwner`/GS9306 *is* real, reachable compiler surface (even though only cs2gs is expected to ever emit it), I added a short entry to the "Fixed" section of the Unreleased (0.5) line anyway, to stay consistent with how thoroughly that file documents diagnostic-behavior changes elsewhere.
- A malformed `@ExtensionOwner` placed on a receiver-clause function that resolves to a same-package-struct *instance method* (not a genuine extension) is silently dropped with no diagnostic, rather than reporting GS9306. cs2gs can never produce this combination (a C# `this` receiver on an unrelated foreign/BCL type is always a genuine extension), so this is an intentionally unclosed corner rather than a functional gap.

## Relationship to #4236

I read #4236 (`Compiler.Tests channel matrix hits InvalidProgramException in ReceiveValueAsync`) for awareness, per instructions, but did **not** attempt to fix it here. Its failure is `System.InvalidProgramException` at `Gsharp.Concurrency.ChannelOps.ReceiveValueAsync[T]` — a plain generic instance/static method on `ChannelOps`, never an extension method, never lifted, never touched by this fix's code paths. The stack trace points at a miscompiled generic async-receive IL body, not a missing/renamed owner type. **My assessment: #4236 is very likely a separate root cause and is not resolved by this PR.** The next agent should still independently re-verify against a fresh self-hosted build before starting on it, since #4234's fix does change what compiles inside `Gsharp.Runtime.Channels`, but I see no mechanism by which it would touch `ReceiveValueAsync`'s IL.


DavidObando:
Follow-up: the isolated, no-concurrent-builds rerun of `build/run-cs2gs-selfmig-pr-guard.sh` finished:

```
8/8 apps green; run PASSED.
```

All 8 hot-core apps (`InternalAnalyzers`, `Core`, `GSharp.Formatting`, `Gsharp.Runtime.Channels`, `Cs2Gs.CodeModel`, `Cs2Gs.Pipeline`, `Cs2Gs.ProjectLoading`, `Cs2Gs.Translator`) pass translate/compile/ILVerify/test-parity, including `Cs2Gs.Pipeline`, which failed the first (concurrent-build-contaminated) run with an unrelated `MSB4236` SDK-resolution race — confirming that failure was an environment artifact and not a code defect, as noted in the PR body.

Combined with the full `Cs2Gs.Tests` corpus (3160/3160 passed) and the real-artifact reflection dump of the migrated `Gsharp.Runtime.Channels.dll` in the PR description, this closes out local verification for #4234.

DavidObando:
Rebased onto main (which picked up #4260 in the meantime) to resolve a conflict — both PRs added a bullet to the same release-notes.md section, and both touch `Binder.cs` at non-overlapping locations. Resolved by keeping both release-notes bullets; the `Binder.cs` merge was clean (auto-resolved, no manual edit needed). Rebuilt (0 warnings/errors) and reran the targeted regression tests (including #4222's) post-rebase — all green.

DavidObando (commented):
## Conflict resolution

Resolved CONFLICTING against main in a scratch worktree (not the locked agent worktree). Conflicts were exclusively in `website/docs/release-notes.md`'s Unreleased section, where concurrent PRs each add their own bullet — resolved by keeping every bullet (this PR's #4234 entry, #4222's async ref-alias entry). Main advanced a second time mid-review (PR #4263 landed), which reopened the same release-notes conflict; resolved identically. Rebuilt (`dotnet build -c Release`: 0 warnings/0 errors) and reran the relevant suites after each merge: `Core.Tests` filtered to Extension/DiagnosticIdUniqueness/4234/4224/4220 (262 passed), and the full `Cs2Gs.Tests` corpus twice (3161 passed, 0 failed, both times — no regression from either merge). Pushed as a fast-forward merge commit (no rebase/force-push). PR now reads `mergeable: MERGEABLE` (`mergeStateStatus: BLOCKED` only pending CI re-run on the new commit).

## Review (scoped to #4234)

Root cause matches the issue's diagnostic trail exactly: a top-level receiver-clause extension func always lands on the package's `<Program>` TypeDef, diverging from the native C# static-class owner that reflection (`typeof(ChannelExtensions).GetMethods()`) and `RendezvousBatchAnalyzer`'s literal-type-name owner check both key off.

The fix reuses ADR-0053's existing static-owner machinery (`StaticOwnerType` + `StaticMethods`) rather than inventing a new binding/emission path — correctly scoped, no gold-plating. `@ExtensionOwner(typeof(T))` is bound purely by name (parallel to `@SuppressDiagnostic`), validated to a same-package non-generic class, and reports the new `GS9306` on misuse (missing typeof, generic, struct, cross-package). `TypeDefEmitter` forces Abstract+Sealed and `ReflectionMetadataEmitter` stamps `[ExtensionAttribute]` on the owner TypeDef, matching the native shape (ECMA-334 §13.6.9). `Binder.cs` correctly skips double-binding an `@ExtensionOwner`-routed function in the shared-block-statics loop since it's already bound via the ordinary top-level pass.

Verified independently rather than trusting the PR's own report: re-ran a scoped self-hosting guard (`run-cs2gs-selfmig-pr-guard.sh`, `SELFMIG_PR_GUARD_APPS` limited to `Gsharp.Runtime.Channels`, unique `SELFMIG_PR_GUARD_ROOT` to avoid the shared-/tmp collision the PR's own report flagged) — translate/compile/ILVerify/test-parity all PASS in isolation, confirming the extension-owner mechanism actually produces the claimed metadata shape end to end, not just in the new unit tests. The added `Issue4234ExtensionOwnerEmitTests` directly assert the owner TypeDef is Abstract/Sealed/`[ExtensionAttribute]`-stamped and hosts the MethodDef instead of `<Program>` — good direct coverage of the core claim.

The two corpus-caught collisions in the second commit are genuine fixes, not workarounds: #3413's owner-scoped-extension wrapper case is skipped via the existing `RequiresOwnerScopedExtension` predicate (reused, not reimplemented) to avoid a duplicate MethodDef, and the `@`-escaped C# identifier case is fixed by reusing the same `EmittedName` allocator the owner's own `TypeDeclaration` already uses, rather than the raw syntax token.

No scope creep: both commits touch only files directly load-bearing for owner-identity routing (binder, emitter, cs2gs translator, diagnostics, tests, docs). The `website/docs/release-notes.md` entry and `docs/diagnostics.md`/`website/docs/ref/diagnostics.md` GS9306 rows are appropriate documentation, not scope expansion.

**Verdict: solid, would approve if self-approval were permitted.**

## Assessment of #4236

The PR's own claim — that #4236 (`Compiler.Tests` channel matrix `InvalidProgramException` in `ReceiveValueAsync`) is a separate, unresolved root cause — is credible. `ReceiveValueAsync` is a plain generic instance method on `ChannelOps`, never lifted as an extension, never touched by any code path this PR changes (owner-annotation binding, TypeDef Abstract/Sealed/ExtensionAttribute stamping, or the cs2gs lift-and-tag logic). The failure signature (miscompiled generic async-receive IL body) has no mechanistic connection to owner-type MethodDef placement. Next agent on #4236 should still independently re-verify against a fresh self-hosted build per the PR's own recommendation, but there's no reason to expect this PR incidentally fixes it.

Issue #4212, opened 2026-09-11
cs2gs: named arguments at a delegate-invoke or full-arity claimed-local-function call site fail loudly (gsc rejects, doesn't mis-bind)


DavidObando:
Found and scoped out as a deliberate follow-up while fixing PR #4211's review feedback (commit `ab247e3e`, which fixed the *silent* version of this same class of bug — see #4197 and PR #4211). Two sibling gaps share one root cause: Roslyn's `IInvocationOperation.Arguments` is in **evaluation order**, not parameter order, and two argument-reassembly code paths in `tools/cs2gs/Cs2Gs.Translator/CSharpToGSharpTranslator.Invocations.cs` didn't originally account for that. PR #4211 fixed the worst instance (silent argument mis-binding for an omitted-default call to a claimed mutually-recursive local function). These two remain, but — importantly — both fail **loudly** (gsc rejects the translation) rather than silently producing a wrong answer, which is why they were correctly triaged as lower-severity follow-ups rather than blocking #4211.

## Gap 1: `TranslateDelegateInvokeArgumentsWithDefaults` (issue #1901's delegate-invoke default-argument path)

This method assumes `operationArguments` walks in parameter order and reconstructs the call by consuming syntax arguments positionally, exactly like the bug #4211 fixed — but for delegate invocations rather than claimed local functions. Unlike a lambda/anonymous-function invocation, a *named* delegate declaration keeps its parameter names, so a named argument through a delegate invocation is legal C#:

```csharp
delegate void D(int a = 10, int b = 20);
D d = ...;
d(b: 5);
```

Confirmed via direct Roslyn probing (Microsoft.CodeAnalysis.CSharp 5.6.0, this repo's pinned version): this call's `IInvocationOperation.Arguments` arrives as `[b Explicit, a DefaultValue]` — the same evaluation-order shape #4211 fixed for local functions. `TranslateDelegateInvokeArgumentsWithDefaults` still assumes positional order and still keeps the `name:` wrapper (via `TranslateArgument`), so it emits something like `f(b: 5, 10)` against a structural arrow type with no parameter names — **gsc rejects this loudly** ("Named argument 'b' does not match any parameter") rather than mis-binding. Confirmed unrepresented in the current corpus (no selfmig failure traces to it today), but it's a real translation gap: legal C# that cs2gs cannot currently translate correctly.

[...]

Issue #3627, opened 2026-08-29
cs2gs: params ReadOnlySpan<T> (C#13 params collections) has no lowering — last translate gap for Cs2Gs.Tests selfmig


DavidObando:
## Summary

cs2gs cannot translate a callee declared with **`params ReadOnlySpan<T>`** (or any span-typed C#13 params collection) — declaration and every call site gap with:

```
CS2GS-GAP: params collection of type 'System.ReadOnlySpan<int>' has no gsc construction form.
```

With the AtToken package-escape fixed (#3624), this is the **only remaining translate gap for Cs2Gs.Tests** in the #3501 self-migration wall (fingerprint `translate-9d7404c16439` in the 2026-08-28 nightly, run 33217212584).

## Why it gaps today

C#13 params collections let `params` accept non-array collection types, and `params ReadOnlySpan<T>` is the overload the C# compiler *prefers* at expanded call sites (allocation-free). cs2gs's model:

- Classic `params T[]` maps 1:1 onto gsc's variadic `...T` (ADR-0101/0102) — fine.
- Non-array params collections are lowered **at the call site** into a `List[T]{...}` literal, gated by the allowlist in `IsSupportedParamsCollectionType` (`CSharpToGSharpTranslator.Invocations.cs:1156`): `List<T>`, `IEnumerable<T>`, `ICollection<T>`, `IList<T>`, `IReadOnlyList<T>`, `IReadOnlyCollection<T>`.
- Anything else (e.g. `ReadOnlySpan<T>`, `HashSet<T>`) is gapped at the **declaration** too (`CSharpToGSharpTranslator.Constructors.cs:961`), deliberately, so we never emit a half-translated callee whose call sites all gap.

The underlying blocker is that **G# has no construction form that produces a `ReadOnlySpan<T>`** — no span literal, no stackalloc/inline-array-backed collection expression — so there is nothing for the call-site lowering to emit.

## Repro

In-repo fixtures already pin the shape (they exist precisely to keep this gap honest):

- `tools/cs2gs/Cs2Gs.Tests/Issue3460TranslatorTests.cs:130` — `Issue3460SpanParamsOptions(params ReadOnlySpan<int> values)`.
- `tools/cs2gs/Cs2Gs.Tests/Issue1901LambdaDefaultsTranslationTests.cs:262` — `static int Total(params ReadOnlySpan<int> values)`; also documents the BCL case `Task.WhenAll(params ReadOnlySpan<Task>)` (line 318).

[...]

Issue #812, opened 2026-06-13
Variadic params on methods, interfaces, ctors, lambdas, delegates (follow-up from #799 / ADR-0101)


DavidObando:
## Summary

Follow-up to #799 / ADR-0101. v1 accepts variadic `name ...T` only on top-level `func` declarations. Lift the restriction to class methods, interface methods, constructors, lambdas, and delegates (currently rejected with `GS0146`).

## Context

ADR-0101 §scope ships variadic `name ...T` for top-level functions only. The PR body for #811 explicitly calls this out:

> **Scope:** top-level `func` only; class/interface/ctor/lambda/delegate continue to reject with `GS0146` until a follow-up ADR.

Each of the additional sites needs:

- **Class instance method**: same binder/emit shape as top-level, plus method-table dispatch — straightforward extension.
- **Class static method (inside `shared { }`)**: same as instance, modulo `this`.
- **Interface method (DIM default body)**: needs the variadic flag to flow through `ADR-0085` default-body emit.
- **Constructor**: the implicit `this` and the `ParamArrayAttribute` must coexist on the last parameter.
- **Lambda expression**: G# lambdas use `(p) -> body` (ADR-0074); accepting `(p ...T) -> body` requires parser + lambda-binding-inference (ADR-0076) updates.
- **Delegate type**: the `(T1, T2, ...T3) -> R` function-type clause (ADR-0075) must accept a trailing variadic. Same change on `delegate` declarations.

## Scope

- Lift the `GS0146` rejection on each of the five sites above.
- Per-site tests + cross-language tests (C# consumer calls the variadic).
- ADR addendum to ADR-0101 (or fresh ADR) covering disambiguation rules where the parser needs them.
- Refresh samples / docs once each form is accepted.

## Acceptance criteria

- `GS0146` no longer fires for valid variadic declarations on any of the five sites.
- IL verified.
- C# consumers see `[ParamArrayAttribute]` on every emitted variadic site.

## Related

- #799 (top-level variadic, MERGED).
- ADR-0101 (variadic spec).
- ADR-0074 (arrow lambdas), ADR-0075 (function-type clauses), ADR-0085 (DIM).
- Parent: #706.

Issue #2810, opened 2026-07-26
Implicit Operator Covariance in lambda expression


captkirk88:
I ran the code described here and randomly came across this bug where a Where func taking a lambda arg that should be implicitly cast to IntValue is resulting in evens done wrong.  The compiler should fail if the cast could not be done.

```
package lang_test 


import System
import System.Collections.Generic
import System.Linq


interface Value {
    func Get() int32;
}

class Int(val int32) : Value {
    func Get() int32 {
        return val
    }

    func operator implicit(val Int) int32 {
        return val.Get()
    }

    func operator implicit(val int32) Int {
        return Int(val)
    }
}

let nums = List[Value]()
nums.Add(Int(1))
nums.Add(Int(2))
nums.Add(Int(3))
nums.Add(Int(4))

// Example Working
let evensRight = nums.Where(x -> x.Get() % 2 == 0)
Console.WriteLine("Evens Done Right:")
for v in evensRight {
    Console.WriteLine(v.Get())
}

// Example Not Working
// (x int32) I assumed would implicitly cast Int to int32, it does not and results in all the items in nums being printed
let evensWrong = nums.Where((x int32) -> x % 2 == 0)
Console.WriteLine("Evens Done Wrong:")
for v in evensWrong {
    Console.WriteLine(v.Get())
}
```

Output:
```
Evens Done Right:
2
4
Evens Done Wrong:
1
2
3
4
```

C# Example that fails at compile time:
```csharp
var nums = new List<Value>
{
    new Int(1),
    new Int(2),
    new Int(3),
    new Int(4)
};

var evensRight = nums.Where(n => n.Get() % 2 == 0);
Console.WriteLine("Evens Done Right:");
foreach (var even in evensRight)
{
    Console.WriteLine(even.Get());
}

var evensWrong = nums.Where((int n) => n % 2 == 0);
Console.WriteLine("Evens Done Wrong:");
foreach (var even in evensWrong)
{
    Console.WriteLine(even.Get());
}


interface Value
{
    int Get();

}

class Int(int val) : Value
{
    private int val = val;

    public int Get()
    {
        return val;
    }

    public static implicit operator int(Int i)
    {
        return i.Get();
    }

    public static implicit operator Int(int val)
    {
        return new Int(val);
    }

    public static Int operator %(Int a, Int b) => new(a.Get() % b.Get());
}
```

C# example fails to compile because it can't find a overload that matches which is what I would think would happen with G#.
```bash
[...]

PR #4268, opened 2026-09-16
Fix lambda variadic carrier mis-binding as slice-of-carrier

### LINKED ISSUE #4235: gsc: named delegate params List<T> carrier is treated as []List<T> when assigning matching lambda
Found while testing #4212 on `origin/main` / `13c96e6e1f18f64886ba291b0b07757f022f2ecb`. This is a separate declaration/binding defect: no invocation, optional argument, or named argument is needed. No compiler sources changed in the #4212 work.

## Standalone G# repro

```gsharp
delegate D(a int32, xs ...List[int32]) int32;
let d D = (a int32, xs ...List[int32]) -> a
```

Run the real Release compiler:

```sh
dotnet out/bin/Release/Compiler/gsc.dll /target:exe /out:repro.dll repro.gs
```

Actual, line 3 column 27:

```
GS0155: Cannot convert type 'System.Collections.Generic.List[int32]' to '[]System.Collections.Generic.List[int32]'.
```

Expected: assignment binds; the delegate and lambda have the same declared carrier. Generalized variadic carriers are supported elsewhere (ADR-0173 / #3627).

The C# shape exposing this during cs2gs tests is:

```csharp
public delegate int D(int a = 1, params System.Collections.Generic.List<int> xs);
D d = (int a, params System.Collections.Generic.List<int> xs) => a;
```

Both defaulted `d()` and explicit-carrier calls fail on the lambda assignment before invocation fidelity can be tested. The standalone repro above removes defaults and all calls, so changing #4212's argument-ordinal reassembly cannot fix this. Standard `params int[]` delegates are covered and working in #4212's focused regression suite. Keep this declaration-side issue separate from that invocation-side fix.

DavidObando:
Fixes #4235

## Root cause

ADR-0173 / #3627 generalized variadic carriers: a `...T` parameter's declared type is `T` itself when `T` is a supported single-argument collection shape (`List[T]`, `Span[T]`/`ReadOnlySpan[T]`, or one of the five `IEnumerable`-family interfaces), and only wraps into a slice `[]T` otherwise. `VariadicCarriers.ResolveDeclaredParameterType` implements that rule, and every declaration path (delegates, named functions, constructors, interface members, anonymous function-type clauses) already routes through it.

`LambdaBinder`'s two lambda-parameter binding sites — `PrepareFunctionLiteralSignature` (`func(...)  {}`  lambdas) and the arrow-lambda binder (`(...) -> ...`) — were missed when #3627 landed. They still had the pre-ADR-0173, #812-era unconditional `...T` -> `[]T` wrap, so an explicitly typed `xs ...List[int32]` lambda parameter bound as `[]List[int32]` (a slice OF the carrier) instead of `List[int32]` (the carrier itself). A named delegate declaring the same `...List[int32]` parameter correctly resolved to `List[int32]`, so the two sides disagreed and assigning the lambda to the delegate failed:

```
delegate D(a int32, xs ...List[int32]) int32;
let d D = (a int32, xs ...List[int32]) -> a
```

```
GS0155: Cannot convert type 'System.Collections.Generic.List[int32]' to '[]System.Collections.Generic.List[int32]'.
```

No invocation was needed to trigger it — the mismatch was already in the declared types.

## Fix

Both `LambdaBinder` sites now call `VariadicCarriers.ResolveDeclaredParameterType` instead of unconditionally wrapping in `SliceTypeSymbol`, exactly like every declaration-side binder already does. Target-typed inference (an omitted lambda parameter type filled in from the delegate's slot) was already correct and is untouched — the fix only affects an *explicitly* typed lambda variadic parameter. Comments at both sites were also corrected: they claimed the old wrap was a "no-op" for `xs ...[]T`, which was never true (`SliceTypeSymbol.Get([]T)` is `[][]T`).

## What was tested

- `dotnet build` Release: 0 warnings, 0 errors.
- New `test/Core.Tests/CodeAnalysis/Binding/Issue4235LambdaVariadicCarrierTests.cs` (11 tests, all passing), covering:
  - The issue's exact repro as a **runtime** assertion (delegate + arrow lambda, `List[int32]` carrier, real call, not just "compiles").
  - Zero-argument and pass-through-existing-`List` calls.
  - The `func(...)  {}` function-literal lambda form (the other binding site).
  - An inferred (delegate-less) function type — exercises the `OverloadResolver.Arguments` packing path for a lambda-typed callee.
  - Target-typed inference with omitted parameter types (`(a, xs) -> ...`), confirming that path was already correct and stays so.
  - The real "array of carrier" shape `...[]List[int32]` still binds as an array, unaffected.
  - Nested generics (`List[List[int32]]`).
  - `Dictionary[K, V]`, which is not a single-argument carrier and correctly keeps the ADR-0101 element interpretation.
  - A second carrier family (`IEnumerable[int32]`), proving the fix isn't List-specific.
  - A genuinely mismatched carrier (delegate wants `List[int32]`, lambda declares plain `...int32`) still reports `GS0155` — no widening was introduced.
- Full existing suites, filtered: `Core.Tests` (Variadic/Lambda/Issue812/Issue3627 filter) — 504 passed; `Compiler.Tests` (same filter) — 409 passed; `Cs2Gs.Tests` (Issue3460/Issue1901/Issue3466/Issue4212 filter) — 121 passed; `Interpreter.Tests` (Variadic filter) — 10 passed. No regressions.
- `build/run-cs2gs-selfmig-pr-guard.sh` (hot-core translation guard, since `LambdaBinder.cs` is in `src/Core`): **8/8 apps green; run PASSED** (translate/compile/ilverify/test-parity for `src/Core`, `InternalAnalyzers`, `GSharp.Formatting`, `Gsharp.Runtime.Channels`, and the four `Cs2Gs.*` hot-core projects).
- Release notes: added a `### Fixed` entry to `website/docs/release-notes.md` under "Unreleased (0.5 line)", after re-reading the file fresh.

## Not changed, and why

cs2gs (the C#-to-G# translator) was intentionally left untouched. The issue explicitly scopes this as a declaration/binding defect separate from #4212's cs2gs argument-ordinal work, and the translator already emits `...List[T]`-shaped G# for a C# `params List<int>` parameter (existing `Issue3460TranslatorTests`, `Issue1901LambdaDefaultsTranslationTests`, `Issue3466*` tests for named functions all pass unchanged). There is no existing skipped/xfail cs2gs test for the lambda+delegate shape to un-skip — grepped for `4235`/`4212` across `test/` and `tools/cs2gs`; nothing referencing this defect exists yet.

Nothing from the issue's scope was deferred.


DavidObando (commented):
Rubber-duck review against issue #4235's scope only (not #4212, not a general variadic-carrier audit).

**Root cause and fix verified.** `VariadicCarriers.ResolveDeclaredParameterType` is the single source of truth for what a `...T` parameter's declared type resolves to (carrier itself vs. `[]T` wrap). Every declaration-side binder already routed through it (delegates, named functions, constructors, interfaces, structs). `LambdaBinder.PrepareFunctionLiteralSignature` and the arrow-lambda binder were the two lambda-parameter sites still hardcoding `SliceTypeSymbol.Get`, unconditionally wrapping — exactly as described. The two-line swap to `VariadicCarriers.ResolveDeclaredParameterType` at both sites is the correct, minimal fix.

**Searched for other missed sites.** Grepped every `SliceTypeSymbol.Get`/`IsVariadic` occurrence across `src/Core/CodeAnalysis/Binding`. Confirmed generic local functions (`let f[T] = func(...) {}`) route through `PrepareGenericLocalFunctionDeclaration` → `PrepareFunctionLiteralSignature`, so they're covered by the same fix, not a separate path. Non-generic local functions are ordinary function-literal expressions, same path. No other declaration-side construction of a variadic parameter type was found outside `VariadicCarriers.cs` itself.

**Verified the tests with mutation testing, not just reading them.** Built the branch (0 warnings/errors) and ran the 11 new tests — all green. Then:
- Reverted both `LambdaBinder` call sites back to the pre-fix `SliceTypeSymbol.Get`: 8/11 tests correctly turn red (the exact repro, function-literal form, IEnumerable/nested-generic carriers, pass-through) while the 3 unrelated ones (array-of-carrier, Dictionary control, target-typed inference) correctly stay green — confirming those 8 tests are actually exercising the fix.
- Restored the fix, then injected a deliberate bug into `VariadicCarriers.ResolveDeclaredParameterType` that collapses a slice-of-carrier (`...[]List[T]`) down to the bare carrier: exactly and only `ArrowLambda_ArrayOfListCarrier_StillBindsAsArray` goes red. That guard is real and precisely targeted.
- The Dictionary[K,V] control is correct per ADR-0101 (two type arguments, not in the single-argument carrier allowlist, so it keeps the element interpretation on both sides).
- The mismatched-carrier test (`List[int32]` vs. `...int32`) still reports GS0155 with the fix applied — no accidental widening.

**Adversarial cases beyond the PR's own tests**, all compiled and run with real assertions on output:
- `var d D = (a int32, xs ...List[int32]) -> a + xs.Count` (explicit-typed `var` target, not just `let`) — compiles, `d(1,2,3)` → 3.
- Lambda passed directly as an argument to a higher-order function expecting the carrier-typed delegate parameter (inference through an argument position, not a variable declaration) — `Apply((a int32, xs ...List[int32]) -> a + xs.Count, 5)` → 7.
- Nested lambda-within-lambda, outer parameter itself a carrier-typed delegate (`Outer(g Inner)` where `Inner` has the `...List[int32]` carrier) — compiles and runs correctly, → 3.
- Cross-carrier (delegate wants `List[int32]`, lambda declares `...IEnumerable[int32]`): now binds `IEnumerable[int32]` (not `[]IEnumerable[int32]`) and is accepted via the existing #2810 adapter rule since `List<T>` converts to `IEnumerable<T>` — correct, and not new widening introduced by this PR, just the pre-existing adapter rule now seeing a materially different (correct) declared type on the lambda side.

**cs2gs scope confirmed untouched.** Only 3 files changed (`LambdaBinder.cs`, the new test file, `release-notes.md`) — nothing under `tools/cs2gs`. Ran the cs2gs variadic/params-relevant filtered suite (Issue3460/1901/3466/4212/Variadic/Params): 169/169 pass.

**Full filtered suites reproduced independently**: Core.Tests (Variadic/Lambda/Issue812/Issue3627 filter) 505 passed, Compiler.Tests (same filter) 409 passed, Interpreter.Tests (Variadic filter) 10/10 passed — matching the PR's own reported numbers.

**Release notes**: entry is accurate to the code change, placed at the top of the `### Fixed` section per convention (most-recent-first). Checked for a rebase/merge conflict against current `main` (one commit ahead of this PR's base, `d35ab7e4`, which also touches `release-notes.md`) — that commit's entry is appended at the *end* of the same `### Fixed` list, non-overlapping with this PR's insertion at the top, and `git merge-tree` confirms a clean merge with no conflict markers.

No scope creep, no defects found. Approving in spirit — self-approval is rejected by GitHub for the PR author, so leaving this as a comment instead.

PR #4269, opened 2026-09-16
Fix InvalidProgramException for out/ref T? over an open value-type parameter

### LINKED ISSUE #4236: selfmig: Compiler.Tests channel matrix hits InvalidProgramException in ReceiveValueAsync

DavidObando:
Fixes #4236

## Root cause

`Gsharp.Concurrency.Chan<T>.ReceiveOrPark`'s `out T? value` parameter (`T` unconstrained, no `struct`/`class` clause) compiled cleanly and passed ILVerify, then threw `InvalidProgramException` at runtime, but only once `T` was JIT-instantiated as a real, non-primitive struct — which is why the native (non-self-hosted) test suite never caught it, and why it surfaced specifically through Compiler.Tests' channel matrix test running against a self-hosted-built `Gsharp.Runtime.Channels.dll`.

`MethodBodyEmitter.EmitStoreIndirect`/`EmitLoadIndirect` (`src/Core/CodeAnalysis/Emit/MethodBodyEmitter.MemberAccess.cs`) choose `stobj`/`ldobj` only when the pointee is *provably* a value type (`NullableLifting.IsAnyValueTypeNullable`) or is a bare, bound `TypeParameterSymbol`. A `NullableTypeSymbol` wrapping an *open, unconstrained* type parameter satisfies neither check, so it fell through to the reference-slot `stind.ref`/`ldind.ref` arm — valid only when the pointee is (or erases to) a reference type. Once `T` resolves to a genuine value type at JIT time, that instruction sequence is invalid IL. This is a gap ILVerify's abstract-over-open-`T` checking does not catch, but the CLR's runtime type loader does.

`ImportedMemberRefFactory.GetElementTypeToken` already resolves this exact shape correctly (issue #814 / ADR-0084 §L5: `Nullable<T>` when `T` is struct-constrained, bare `T` otherwise) — `EmitStoreIndirect`/`EmitLoadIndirect` simply never routed into it for an open type parameter's `T?`. The fix adds that missing case to both, routing to `stobj`/`ldobj` with the already-correct token.

I initially suspected an async-state-machine emission bug (per the issue's framing) but disproved that quickly: `ChannelOps.ReceiveValueAsync<T>` is a plain, non-async generic method. Swapping a self-hosted-built `Gsharp.Runtime.Channels.dll` into the native test harness reproduced the exact crash in seconds without any self-hosting turnaround, and bisecting down from there (reflection-invoking the method directly, then progressively simpler stand-in methods with the same signature shape) localized it to `Chan<T>.ReceiveOrPark`'s `out T? value` parameter specifically — a plain `T` parameter or an out-parameter of a *concrete* nullable value type both worked fine; only the open-type-parameter `T?` combination failed. Dumping the IL confirmed the `stind.ref` instruction directly.

Note: issue #4234 (ChannelExtensions owner-type mismatch, PR #4267) is unrelated and was not yet merged to `main` when I investigated — I confirmed independently that #4236's failure has no connection to it, matching the prior analysis referenced in the issue.

## What I tested

- **Regression test**: `Issue4236OpenNullableTypeParameterIndirectTests` (new, `test/Compiler.Tests/Emit/`) — a minimal, non-self-hosted G# program with a generic class method taking `out value T?` and `ref value T?`, instantiated over a user struct and over `int32`. Compiles, runs, and IL-verifies via native `gsc`. Confirmed **red** (`InvalidProgramException`) before the fix and **green** after.
- **Exact issue reproduction**: swapped a self-hosted-built `Gsharp.Runtime.Channels.dll` into `Compiler.Tests`' output and ran the existing `Issue2965ChannelElementSlotTests.ChannelElementMatrix_LoadsVerifiesAndRuns` (the test named in the issue) — reproduced the exact reported stack trace before the fix, passes after.
- **Targeted local suite**: `dotnet test test/Compiler.Tests --filter "FullyQualifiedName~Issue4236|FullyQualifiedName~Issue3501|FullyQualifiedName~Issue2965|FullyQualifiedName~Nullable|FullyQualifiedName~OutVar|FullyQualifiedName~Ref"` — 1421 passed, 0 failed.
- **Full solution build**: `dotnet build GSharp.sln -c Release` — 0 warnings, 0 errors.
- **Self-hosting PR guard** (`build/run-cs2gs-selfmig-pr-guard.sh`), with the fix applied: 8/8 hot-core apps PASS translate/compile/ilverify/test-parity, including a clean `src/Sdk/Gsharp.Runtime.Channels/Gsharp.Runtime.Channels.csproj` self-hosted build.
- `ilverify` on the fixed self-hosted `Gsharp.Runtime.Channels.dll`: no new finding (the one pre-existing, tracked `ReturnPtrToStack` false positive on `SelectRandom.Shuffle` remains, unrelated).

## Release notes

Added a `### Fixed` entry under `## Unreleased (0.5 line)` in `website/docs/release-notes.md`, re-read fresh against current `main` before editing.

Nothing deferred — root cause fully diagnosed and fixed at the shared emitter helper (both `EmitStoreIndirect` and `EmitLoadIndirect`), not patched at the one call site the issue named.


DavidObando (commented):
Rubber-duck reviewed strictly against issue #4236's scope. No scope creep found in the diff (emitter fix + regression test + one release-notes line). Self-approve is rejected by GitHub (I'm the PR author), so filing this as a review comment instead of an approval — verdict is approve.

## Verified independently

**Root-cause claim.** Confirmed by reading `EmitLoadIndirect`/`EmitStoreIndirect` (`MethodBodyEmitter.MemberAccess.cs`) and `ImportedMemberRefFactory.GetElementTypeToken`: a `NullableTypeSymbol` over an open `TypeParameterSymbol` satisfied neither the `IsAnyValueTypeNullable` arm (not provably a value type) nor the bare-`TypeParameterSymbol` arm (the pointee here is the `Nullable` wrapper, not the parameter itself), so it fell through to `ldind.ref`/`stind.ref`. `GetElementTypeToken` already had the correct logic for this exact shape (struct-constrained → `Nullable<T>` TypeSpec, otherwise → bare-`T` TypeSpec) — the new arm just routes into it via `ldobj`/`stobj`, reusing rather than duplicating that logic.

**Red/green, done myself, not just trusted.** Built gsc from the branch, reverted only the emitter hunk in a scratch worktree, rebuilt. Ran the PR's own `Issue4236OpenNullableTypeParameterIndirectTests` against the reverted build via `dotnet test` — reproduces the exact reported `InvalidProgramException` at `Box<T>.TrySet`. Restored the fix, rebuilt, test passes. (One methodological note for the record: my first pass at an independent repro gave a false "doesn't crash" result — turned out to be several differently-built same-named test DLLs sitting in one directory, so the host resolved the wrong assembly by simple name. Isolating each build in its own directory fixed that; not a defect in the PR, whose own `RunBounded` harness already compiles each test into its own subdirectory.)

**Adversarial matrix beyond the PR's own test**, each compiled+run+ILVerify'd in an isolated directory against the fixed build (all correct), and the value-type cases reproduced red against the reverted build:
- `bool`, `decimal`, tuple `(int32,int32)` — primitives/value-type shapes beyond the PR's `Pair`/`int32`, all crash pre-fix, pass post-fix.
- Unconstrained `Box[T]` instantiated over a reference type (`StringBox`) — confirms the new `stobj`/`ldobj` routing (bare-`T` token, since `GetElementTypeToken` only picks `Nullable<T>` for a struct-constrained `T`) still preserves reference identity through both `out` and `ref`; `ReferenceEquals` held both before and after the fix, so no regression on the reference-type path that used to rely on `stind.ref` legitimately.
- `[T class]`-constrained reference type — same, no regression.
- Doubly-nested generic (class type parameter `T` vs. an unrelated method type parameter `U`) — IL dump confirms `Combine` emits `ldobj !0` (class TP / VAR) while `Inner` emits `box !!0` (method TP / MVAR) for the unrelated parameter, i.e. the fix resolves the correct token in the nested context.
- A pure method-type-parameter (MVAR) case (`Util.Echo[U]`) — same arm, same correctness.
- Load/store symmetry confirmed directly from the IL: `Reassign`'s ref parameter emits both `ldobj !0` (the `value!!` read) and `stobj !0` (the `value = replacement` store) — not a load-only or store-only half-fix.

**ILVerify.** Confirmed it passes cleanly on both the pre-fix (buggy) and post-fix builds of my adversarial program — corroborates the PR's claim that this is a real ILVerify static-analysis gap over open, unconstrained `T`, not something a stricter ILVerify run would have caught either way. Only concrete-type JIT execution surfaces it.

**Self-hosting claim.** `Chan<T>.ReceiveOrPark`'s `private ReceiveStart ReceiveOrPark(CancellationToken, out T? value, out bool ok, out OpReceiveNode<T>? node)` on unconstrained `T` is real, confirmed directly in `src/Sdk/Gsharp.Runtime.Channels/Chan{T}.cs`. Didn't re-run the hour-long PR guard, but full PR CI (including `ilverify`, `tests (runtime-channels)`, `cs2gs-oahu`) is green already, which is consistent with the claim.

**Release notes.** Entry is accurate against the actual diff and correctly placed as the last `### Fixed` item under `## Unreleased (0.5 line)`, consistent with the section's existing entries.

No defect found — fix is symmetric, minimal (reuses the existing correct `GetElementTypeToken` logic rather than reimplementing it), and doesn't regress the reference-type path it doesn't touch.

PR #4270, opened 2026-09-16
Reject closures capturing a ref/out/in parameter (GS9010)

### LINKED ISSUE #4259: gsc: closure captures a ref/out/in parameter silently instead of being rejected

DavidObando:
## Summary

Fixes #4259

A closure (lambda or non-generic local function) that captures a `ref`/`out`/`in` PARAMETER of its enclosing function compiled successfully and silently lost the mutation instead of being rejected:

```gsharp
func Foo(ref x int32) {
    let Bad = func(a int32) int32 {
        x = x + 1
        return a
    }
    Bad(1)
}
```

The closure's field snapshots `x`'s value at closure-creation time rather than holding a live reference to the caller's argument, so the write inside `Bad` never reached the caller. Capturing a by-ref parameter into a heap-allocated closure is unsound in general — the referent's lifetime isn't guaranteed to outlive the closure — so this is now rejected at compile time instead of silently miscompiled.

## Root cause

`LambdaBinder`'s capture-legality checks (in the two places that call `CollectCapturedVariables` for an actual closure — the `func`-literal/local-function body binder and the lambda binder) already rejected:
- a `ref struct` capture (`GS0219`)
- a managed-pointer (`*T`/`ByRefTypeSymbol`) capture (`GS9004`)
- a `fixed`-pointer capture (`GS9008`)

but had no check for a captured variable that is itself a by-ref parameter. A `ref`/`out`/`in` parameter's own `Type` carries no by-ref marker (only `ParameterSymbol.RefKind` does), so none of the existing type-based checks caught it, and the capture flowed straight into `ClosureEmitter`, which creates the closure field directly from `captured.Type` — a plain by-value field.

## Fix

Both capture-legality loops in `src/Core/CodeAnalysis/Binding/LambdaBinder.cs` now add a branch: a captured variable that is a `ParameterSymbol` with `RefKind.Ref`/`Out`/`In` reports a new diagnostic, `GS9010` (`RefParameterCannotBeCaptured`), naming the parameter and its ref-kind keyword.

This is deliberately scoped to `ParameterSymbol`, not `LocalVariableSymbol.RefKind` in general, so it does **not** affect a `ref`/`var ref` LOCAL alias capture — a distinct capture kind (plain `LocalVariableSymbol`, not `ParameterSymbol`) with its own, separately handled legality rules (including the async/iterator suspension-confinement rules #4222 added). It also does not affect a struct instance method's implicit `this` parameter, which is constructed with `RefKind.None`.

## What I tested

- `dotnet build -c Release`: 0 warnings, 0 errors (full solution, including the new test file and diagnostic plumbing).
- New regression suite `test/Core.Tests/CodeAnalysis/Binding/Issue4259RefParameterCaptureEscapeTests.cs` (10 tests, all passing):
  - the issue's exact repro (`ref` parameter mutated inside a `let`-bound local function) reports `GS9010`
  - a read-only capture of a `ref` parameter also reports `GS9010`
  - `out` and `in` parameter captures report `GS9010`
  - the arrow-lambda spelling (`(p) -> {...}`, a different binder call site) reports `GS9010`
  - a generic local function (`let Name[T] = func...`) capturing an outer `ref` parameter reports `GS9010`, matching the issue's note that this reproduces identically for #4221/#4252's generic-local-function capture support
  - an ordinary by-value parameter capture, an ordinary local capture, and a `ref` parameter used only locally (never captured) all remain legal — no over-rejection
  - a `ref`/`var ref` LOCAL alias capture does **not** trip the new `GS9010` check
- `docs/diagnostics.md` and `website/docs/ref/diagnostics.md` updated with the new `GS9010` row (kept in sync per `DiagnosticIdUniquenessTests`, which now passes).
- `website/docs/release-notes.md` updated under "Breaking changes" (Unreleased/0.5 line).
- Full `Core.Tests` regression sweep filtered for `Closure|Lambda|LocalFunction|Capture`: 505 passed, 0 failed.
- Targeted regression sweep for the freshly-landed #4221/#4222/#4223/#4224 ref/capture machinery (`RefLocalAliasing`, `Issue4222`, `Issue2329FixedPointerCaptureEscape`, `RefStructTests`, `RefSafeEscapeTests`, `Issue4221`, `Issue4223`, `Issue4224`, `CaptureBoxing`, `LambdaBinder`, `ClosureEmitter`): 134 passed, 0 failed.
- Full `Compiler.Tests` sweep filtered for `Closure|Lambda|Capture|Ref`: 1136 passed, 0 failed.
- Manual `gsc` repros confirming: `ref`, `out`, and `in` parameter captures all report `GS9010`; a struct instance method's closure over `this`/a field still compiles and runs correctly (no false positive); a `var ref` local alias captured in a closure is unaffected by this change.

Nothing deferred — the fix covers both closure-binding call sites (local function / `func` literal body, and arrow lambda), all three by-ref parameter kinds (`ref`/`out`/`in`), the generic-local-function variant called out in the issue, and the documentation/release-notes update convention.


DavidObando (commented):
Rubber-duck review against #4259's scope only. Verdict: solid, would approve (self-approval blocked by GitHub since I'm also the PR author).

**Scope check.** The two capture-legality check sites in `LambdaBinder.cs` (`BindFunctionLiteralBody` and the arrow-lambda binder) are exhaustive — grepped every `CollectCapturedVariables` call site in the binder; the only other two (`CreateClrMethodGroupAdapter`, `CreateUserExtensionMethodGroupAdapter`) synthesize a receiver-capture adapter that only ever captures a synthesized `LocalVariableSymbol` (`receiverTemp`, a byval snapshot), never a `ParameterSymbol` directly, so they're correctly out of scope. `cs2gs` has no independent capture-legality notion — it lowers to G# and goes through the same binder.

**Transitive/nested-capture soundness — the main risk area.** Verified with a built gsc from the branch:
- Lambda-in-lambda two levels deep: both the outer and inner literal correctly report GS9010 (nested-literal capture folding via `RewriteFunctionLiteralExpression` propagates `CapturedVariables` outward correctly).
- The exact #4221/#4261 shape — a **generic** local function calling a **generic sibling** that captures the ref parameter (`let Helper[T] = func(a T){ x = x+1; ... }`, `let Caller[U] = func(b U){ return Helper(b) }`) — reports GS9010 on both Helper (direct) and Caller (folded via `calleeCapturedVariables`/`RewriteCallExpression`). No laundering.
- More generally: every closure literal is bound through one of the two patched call sites unconditionally, so the function that *directly* captures the ref parameter is always checked at its own bind site regardless of whether any caller's folded capture set is complete — compilation fails at the direct-capture site even in shapes where transitive folding doesn't apply (e.g. a non-generic `let X = func(){}` called as a delegate, where the "callee" call isn't `RewriteCallExpression`'s `FunctionSymbol` shape at all). This is structurally sound, not just empirically.
- Forward-referencing a sibling local function before its declaration is independently rejected (GS0130, "Function doesn't exist"), so declaration-order dependence in the folding cache is moot.

**Generic local function + issue's own repro + non-regression:** all reproduced/confirmed via built gsc — issue's verbatim repro → GS9010; `out`/`in` parameters → GS9010; arrow-lambda spelling → GS9010; generic local function capturing outer `ref` → GS9010; ordinary by-value parameter/local capture → compiles and runs correctly (mutation observable, e.g. `total` reaches 2 after two calls through the closure); struct-method closure over `this`/a field (read path) → compiles and runs correctly, RefKind.None confirmed at both `this`-parameter construction sites; `var ref` LOCAL alias capture → still compiles, not rejected.

**Docs/tests:** `docs/diagnostics.md` and `website/docs/ref/diagnostics.md` both updated (GS9001–GS9009 → GS9001–GS9010 header + new row), GS9010 not used anywhere else. `dotnet build -c Release`: 0 warnings/errors. Ran the PR's own claimed sweeps and they match exactly: `Core.Tests` filtered `Closure|Lambda|LocalFunction|Capture` → 505 passed; the #4221/#4222/#4223/#4224 targeted sweep → 134 passed; `DiagnosticIdUniquenessTests` + new `Issue4259RefParameterCaptureEscapeTests` → 14 passed; `Compiler.Tests` filtered `Closure|Lambda|Capture|Ref` → 1136 passed.

**One real defect found, correctly out of scope, filed separately: #4271.** A `ref`/`var ref` LOCAL alias captured by an ordinary (non-async) closure and invoked synchronously also silently loses the mutation — same failure mode as #4259, but for a `LocalVariableSymbol` rather than a `ParameterSymbol`, so it doesn't trip GS9010 and isn't touched by this diff (`CaptureBoxingRewriter.cs`, where `IsBoxable` returns `false` for any `RefKind != None` local, is untouched by this PR). Confirmed with a repro triad: direct alias mutation (no closure) writes through correctly; an ordinary captured local shares mutation correctly across two closure calls; the same alias captured inside a closure and invoked once does not write through. Pre-existing, unrelated to #4222 (which only guards liveness across an `await`/`yield` suspension point, not capture-by-closure), not a regression from this diff. Not a blocker for this PR.

**One prose nit, non-blocking:** the PR description and the code comments in `LambdaBinder.cs`/`DiagnosticDescriptors.cs` say the `ref`/`var ref` LOCAL alias case is "separately handled" with "its own legality rules." Per the above, there isn't a legality rule for the synchronous-closure-capture case — it's simply unchecked, and `CaptureBoxingRewriter.IsBoxable`'s `RefKind != None` branch is what *produces* the (wrong) snapshot, not what validates it. Worth a follow-up wording tweak referencing #4271, but doesn't affect correctness of this fix.

No over-rejection, no scope creep, no soundness gap found within #4259's actual scope. GS9010 fix is sound and complete for ref/out/in PARAMETER capture.

Issue #3888, opened 2026-09-04
cs2gs: a params ELEMENT is never widened, so an expanded `null` argument translates to `nil` against a non-null `...T` carrier


DavidObando:
Found while triaging #3886. Latent — no site in the current migration corpus
reaches it — but it is a translated program gsc rejects, so it will surface.

The declaration side of a `params T[]` never widens its **element** type,
whatever null evidence exists at the call sites. An expanded `null` literal
argument therefore translates to a bare `nil` against a non-null `...T`
carrier.

## Repro (oblivious C#, translator-level)

```csharp
namespace Demo
{
    public static class Fixture
    {
        public static string[] LoadTogether(params string[] assemblyPaths) => assemblyPaths;
    }

    public static class Caller
    {
        public static string[] Go(string output) => Fixture.LoadTogether(null, output);
    }
}
```

translates to

```gs
class Fixture {
    shared {
        func LoadTogether(assemblyPaths ...string) []string {
            return assemblyPaths
        }
    }
}

class Caller {
    shared {
        func Go(output string) []string {
            return Fixture.LoadTogether(nil, output)
        }
    }
}
```

`nil` into a `...string` slot is GS0154. The C# was legal — `LoadTogether(null,
output)` puts a null *element* in the array — so this is a fidelity loss, not
a rejected-input case.

The `!!` bridge cannot help here and should not: `nil!!` is a guaranteed
runtime throw. The correct repair is on the declaration side — an element that
receives direct null evidence should be emitted `...string?`, the same way an
ordinary parameter is promoted under #1072. Note the array-level promotion
that #3886 disconnected from the sink contract is a *different* question
(`paths` itself being null); this one is about the element.

Behaviour is identical before and after #3886's PR — that PR only changed
which sink contract the bridge reads, and `ForgiveNullableReferenceValue`
already declines to assert on a literal `null`.

Issue #3060, opened 2026-08-01
cs2gs regression: nullable-oblivious interface properties diverge from implementations (GS0504)


DavidObando:
### Summary

The nullable-oblivious interface/implementation asymmetry from #2285 still reproduces on current `main`. Migrating Oahu translates implementation properties to nullable reference types while leaving their interface contracts non-nullable, producing `GS0504` errors.

### Version

- gsharp commit: `f5e9d89250fd20debc41a41de3b7dec077265be8`
- cs2gs/gsc version: `0.3.496+f5e9d89250`
- command: `dotnet out/bin/Release/Cs2Gs.Cli/cs2gs.dll migrate --corpus /path/to/Oahu --out /path/to/Oahu-gs --config Release`

### Reproduction

Nullable-oblivious C# input includes contracts such as:

```csharp
public interface IPerson
{
    string Name { get; set; }
    ICollection<Book> Books { get; }
}

public class Author : IPerson
{
    public string Name { get; set; }
    public virtual ICollection<Book> Books { get; } = new List<Book>();
}
```

cs2gs emits an interface with non-null members:

```gsharp
interface IPerson {
    prop Name string
    prop Books ICollection[Book] { get; }
}
```

but emits nullable implementation members:

```gsharp
open class Author : IPerson {
    prop Name string?
    open prop Books ICollection[Book]? { get; init; }
}
```

The same mismatch occurred for `IBookCommon.Conversion`, `IConversion.DownloadFileName`, and `IConversion.DestDirectory`.

### Actual result

Eight errors such as:

```text
error GS0504: Type 'Author' cannot use property 'Name' to implement interface property 'IPerson.Name': expected type 'string', actual type 'string?'.
```

### Expected result

Nullability promotion should propagate consistently across interface/implementation edges, producing a sound, internally consistent G# repository without manual edits.

### Workaround

Manually widening the generated interface properties to nullable (`string?`, `ICollection[Book]?`, `Conversion?`) allows the affected project to compile.

### Related

- #2285 — previously completed report for the same asymmetric nullable-oblivious promotion behavior.

Issue #2412, opened 2026-07-17
cs2gs: oblivious-nullability taint doesn't cross project/compilation boundaries → GS0155/GS0156 (Oahu.Core AaxExporter.Asin)


DavidObando:
### Summary

`cs2gs`'s oblivious-nullability taint analysis (`ObliviousNullabilityAnalyzer`, issues #2113/#2157/#2167/#2259/#2285) is **scoped to a single `CSharpCompilation`** (cached via `ConditionalWeakTable<Compilation, TaintResult>`, and `Compute` only walks `compilation.SyntaxTrees`). `CSharpProjectLoader.LoadProjectWithReferencesAsync` (used by `cs2gs migrate`/`TranslateStage` for any app with `ProjectReference`s) opens the app project **and every transitively-referenced sibling project as separate `LoadedCSharpProject`/`CSharpCompilation` objects** — one Roslyn `Compilation` per project, linked only via ordinary project/metadata references, never merged.

Consequence: when a downstream project's translator asks "is this symbol null-tainted?" for a symbol **declared in a referenced sibling project** (an interface member or a class member that implements it), `ObliviousNullabilityAnalyzer.IsTainted(downstreamCompilation, symbol)` walks the **downstream project's own** syntax trees — which don't contain the sibling's declaration/body — finds no taint evidence, and returns `false`. The sibling project's *own* standalone translation correctly promotes the same symbol to `T?` (its own compilation *does* contain the tainting `?.`/interface-edge evidence). The result is a genuine, self-consistent-per-project but globally-inconsistent translation: the sibling's compiled G# type is `T?`, but the downstream project emits a bare (non-`!!`) read of it wherever it flows into a non-nullable sink (object-initializer member, return, argument, etc.), and `gsc` correctly rejects the mismatch as GS0155/GS0156.

This is a **new, project-boundary-specific** gap distinct from the already-fixed intra-compilation cases (#2157 property/indexer scanning, #2167 interprocedural/transitive, #2259 `?[]`, #2285 interface/impl symmetry) — all of those fixes work correctly *within* one compilation, which is exactly why the isolated repro below shows the referenced project (`LibB`) migrating 100% green while the *consuming* project (`LibA`) fails.

### Real-world diagnostic (round 17 full-corpus run, `gsc 0.3.90+203ecd4a43`, gsharp `origin/main` @ `203ecd4a`)

```
[...]

Issue #2113, opened 2026-07-05
[analysis] cs2gs on nullable-oblivious C#: T?-vs-T conversion cluster (GS0154/0155/0156) needs design decision


DavidObando:
## Context

While migrating **Oahu.Foundation** to G# with cs2gs (issue #914), the remaining compile errors (~58 after the fixes in #2109, #2110, #2112) are dominated by a single **design-level** cluster that I believe needs owner analysis rather than piecemeal patching. Filing this for analysis — **no labels**, not claiming a specific bug.

## The cluster

Current Oahu.Foundation histogram (post #2109/#2110/#2112):

```
  27 GS0155   Cannot convert type 'X?' to 'X' (or 'nil' to 'X')
  11 GS0156   Cannot convert type 'X?' to 'X'. An explicit conversion exists
   3 GS0154   Parameter requires 'X' but was given 'X?'
```

i.e. ~41 of 58 are the same shape: **a nullable value `X?` is being used where a non-null `X` is required.**

Representative messages:

```
Cannot convert type 'IEnumerable<object>?' to 'IEnumerable<object>'        (x5)
Cannot convert type 'string?' to 'string'                                   (x8, GS0156)
Cannot convert type 'nil' to 'string' / 'object' / 'IFileCopyCallout'
Cannot convert type 'System.Version?' to 'System.Version'
Cannot convert type '(string?, string?)' to '(string, string)'
Cannot convert type '(ProgressMessage) -> void?' to '(ProgressMessage) -> void'
Cannot convert type 'T?' to 'T' / 'T?' to 'object'
Cannot convert type '[]?uint64' to '[]uint64'
```

## Root cause (why this is hard)

**Oahu.Foundation is nullable-*oblivious*** — it has no `<Nullable>` setting and no `#nullable` directives, so Roslyn reports every reference type as oblivious (`string`, not `string?` or `string!`), with **no flow state** to distinguish "definitely non-null" from "maybe null".

Meanwhile cs2gs legitimately *introduces* `T?` in several places:
- translating `?.` conditional-access chains (result is `T?`),
- `= null` field/parameter defaults,
- `default(refType)` → `nil`,
- nullable-returning BCL APIs (`FirstOrDefault`, `IEnumerable<object>?`, etc.).

[...]

PR #4272, opened 2026-09-16
cs2gs: fix spurious !! on nullable delegate argument crossing an oblivious project boundary

### LINKED ISSUE #4262: cs2gs inserts spurious non-null assertion (!!) on intentionally-nullable delegate parameter, breaking runtime behavior
## Summary

`cs2gs` inserts a spurious non-null assertion (`!!`) when translating a locally-scoped, intentionally-nullable delegate parameter, breaking runtime behavior that the source C# relied on. Found while migrating [DavidObando/Oahu](https://github.com/DavidObando/Oahu); traced to a specific, reproducible translation defect.

## Reproduction

Source repo: `DavidObando/Oahu`
Pre-migration C# commit (last commit before the G# migration merged): **`1ecfb07d9a2f598129d62d9302f564eb2d840116`**
Migration commit (cs2gs output): `24857282a979c9f631d3b472d78740d07c303555` (#67)

File under test: `src/Oahu.Cli.App/Jobs/AudibleJobExecutor.cs`

```
git clone https://github.com/DavidObando/Oahu.git
cd Oahu
git checkout 1ecfb07d9a2f598129d62d9302f564eb2d840116
# run cs2gs against src/Oahu.Cli.App/Jobs/AudibleJobExecutor.cs (project has <Nullable>enable</Nullable>)
```

## What goes wrong

The C# source (commit `1ecfb07`) passes a nullable local plainly, with **no** null-forgiving operator:

```csharp
// AudibleJobExecutor.cs, ExecuteAsync — convertAction is `ConvertDelegate<CliCancellation>? convertAction = null;`
// and stays null on the plain "download, no --export" path (see doc comment:
// "Phase 4c.1 only handles download + decrypt (convertAction = null).")
Task runTask = Task.Run(async () =>
{
    using var job = new DownloadDecryptJob<CliCancellation>(api, jobSettings, onState);
    await job.DownloadDecryptAndConvertAsync(
        new[] { conversion }, progress, new CliCancellation(linkedToken),
        convertAction).ConfigureAwait(false);   // <-- plain, no `!`
    ...
});
```

The callee's own parameter is nullable by design in both languages:

```csharp
// DownloadDecryptJob.cs
public async Task DownloadDecryptAndConvertAsync(
    IEnumerable<Conversion> selectedConversions,
    IProgress<ProgressMessage> progress,
    T context,
    ConvertDelegate<T>? convertAction)   // nullable — null is an expected, documented input
```

There is no nullability mismatch to reconcile here — passing a nullable local into a nullable parameter needs no assertion in C#, and the compiler emits no CS8604 at this call site.

cs2gs's translated G# output (`src/Oahu.Cli.App/Jobs/AudibleJobExecutor.gs:206`) instead emits:

```g#
await job.DownloadDecryptAndConvertAsync(
    []Conversion{conversion}, progress, CliCancellation(linkedToken),
    convertAction!!          // <-- inserted by cs2gs; not present in source
).ConfigureAwait(false)
```

`convertAction!!` throws at runtime whenever `convertAction` is actually `nil` — which is the *normal*, documented case for a plain download (no `--export` flag). Every `oahu-cli download <asin>` (no `--export`) therefore crashed with `NullReferenceException` ("Object reference not set to an instance of an object.") immediately after the license-request phase update, before any real network I/O. This affected 100% of plain CLI downloads post-migration, not an edge case.

## Likely root cause / where to look

- `src/Oahu.Cli.App/Oahu.Cli.App.csproj` has `<Nullable>enable</Nullable>`, so `ObliviousNullabilityAnalyzer` (documented to only run for non-nullable-enabled/legacy compilations) is not the mechanism — this is a defect in the *normal* nullable-enabled translation path, likely in `tools/cs2gs/Cs2Gs.Translator/CSharpToGSharpTranslator.Nullability.cs`.
- Hypothesis: cs2gs's flow-sensitive null-state tracking for `convertAction` merges the two branches of
  ```csharp
  ConvertDelegate<CliCancellation>? convertAction = null;
  if (request.ExportToAax || request.ExportToM4b) { ... convertAction = (book, ctx, callback) => { ... }; ... }
  ```
  incorrectly at the post-`if` join point (the `if` has no `else`), concluding `convertAction` is "definitely non-null" there when the correct join result is "still nullable" (null when the condition is false).

## Second instance — same file, same root pattern (lower severity)

```csharp
// C# original, in the m4b-export closure:
var src = book.Conversion.DownloadFileName + ".m4b";     // plain, no `!`
```
```g#
// G# migrated output:
let src = book.Conversion!!.DownloadFileName + ".m4b"    // spurious `!!`
```
Traced further: `Book.Conversion` is a **non-nullable** EF Core navigation property in the C# source (`src/Oahu.Data/Entities.cs:82,157` — `public virtual Conversion Conversion { get; set; }`, no `?`), but cs2gs translated it to a **nullable** property in G# (`src/Oahu.Data/Entities.gs:68,169` — `open prop Conversion Conversion?`). That single property-nullability change then forces `!!` at every `.Conversion` access site throughout the migrated codebase — a plausible contributor to a large fraction of the 957 total `!!` occurrences across `src/` post-migration. This instance didn't crash in practice (the property genuinely is always populated at runtime via eager `.Include()` loading), but it's the same class of defect: cs2gs disagreeing with the source's own nullability contract and papering over the disagreement with a runtime-throwing assertion instead of preserving the source's typing.

## Impact

Silent, systemic correctness regression from an otherwise-clean automated migration (PR #67's own validation — "IL verification and test parity", CLI `--version`/`--help`/`auth status` smoke checks — never exercised an actual `download`, so this shipped undetected). Any cs2gs user relying on "null is a legitimate default for an optional delegate/callback parameter" is at risk of the same crash.

## Suggested fix direction

1. For the `convertAction` case: fix the flow-join logic so a variable's nullability after an `if`-without-`else` conditional reassignment is the union of "unchanged" (original nullable state) and "reassigned" (branch's state) — not narrowed to the branch's state alone.
2. For the `Book.Conversion` case: preserve the source C# property's declared nullability (non-nullable) in the migrated G# property instead of defaulting EF navigation properties to nullable.

DavidObando:
Fixes #4262

## Root cause

`cs2gs` inserted a spurious `!!` non-null assertion on an intentionally-nullable
delegate argument at a call site where a nullable-ENABLED consumer project
(e.g. `Oahu.Cli.App`) calls a delegate parameter declared in a nullable-OBLIVIOUS
sibling project it references (e.g. `Oahu.Core`), even when that sibling's own
whole-program taint fixpoint (`ObliviousNullabilityAnalyzer`, issue #2113/#2412)
proves the parameter genuinely accepts `nil` via a transitive
`if (convertAction != null)` guard several calls downstream.

Two defects, both in `tools/cs2gs/Cs2Gs.Translator/ObliviousNullabilityAnalyzer.cs`:

1. **`IsTainted`'s cross-project entry gate.** The three-argument
   `IsTainted(compilation, symbol, siblingCompilations)` overload bailed out
   immediately whenever the ASKING compilation (the consumer's own) was
   nullable-enabled — before its sibling-walk loop, which already correctly
   remaps into and gates on the obliviousness of each sibling independently,
   ever got a chance to run. Issue #2412's cross-project machinery therefore
   only ever fired when the CONSUMER itself was oblivious, never when an
   enabled consumer called into an oblivious sibling — the exact real-world
   Oahu shape. Fixed by bailing only when neither the asking compilation nor
   any sibling is oblivious.

2. **A second, narrower defect surfaced while fixing the first.**
   `IsTaintedCore`'s own whole-program fixpoint (`Compute`) reads raw syntax
   with no nullable-context guard of its own — an oblivious-only heuristic,
   since oblivious code carries no trustworthy annotation to read instead.
   Once the entry gate above stopped preventing it from ever running against
   a nullable-ENABLED compilation, a `T?` local legitimately initialized to
   `null` in the ENABLED consumer's own source could self-taint a
   cross-project call target through an ordinary argument-to-parameter edge —
   an unsound false positive that happened to look correct for a
   genuinely-tainted target, but would have silently suppressed the `!!`
   bridge even for an oblivious sibling parameter with NO null-check evidence
   anywhere (verified with a regression test). Fixed by only ever consulting
   a compilation's own cached fixpoint when that compilation is itself
   oblivious; the sibling walk still always runs.

3. **`IsParamsElementTainted`** (issue #3888's cross-project `params T[]`
   element taint) had the identical entry-gate bug as (1), reachable the same
   way from `TargetWillRemainNonNullableReference`'s variadic-carrier branch.
   Its own candidate loop already gated each candidate (including itself) on
   obliviousness before touching its cache, so only the entry bail needed
   widening there.

`!!` is a real runtime throw in G# (unlike C#'s erased `!`), so this was a
silent, systemic correctness regression: every plain (non-`--export`)
`oahu-cli download` call crashed with `NullReferenceException` post-migration.

## What I tested

- `dotnet build` Release: 0 warnings, 0 errors.
- New regression tests:
  - `tools/cs2gs/Cs2Gs.Tests/Issue4262CrossProjectObliviousDelegateArgumentTests.cs`:
    a nullable-enabled consumer calling a generic delegate parameter in an
    oblivious sibling proven tainted by a transitive null-check (pass-through
    and captured-in-nested-closure shapes), a precision guard confirming a
    sibling parameter with NO taint evidence still gets `!!` (the fix is not a
    blanket disable), and the `params`-element sibling case.
  - `test/Compiler.Tests/Issue4262NullableDelegateArgumentRuntimeTests.cs`:
    actually compiles and RUNS the fixed and buggy shapes end to end via
    `gsc` + `dotnet exec` — confirms the bare argument runs to completion on a
    legitimate `nil`, and that the `!!`-asserted shape genuinely throws
    `NullReferenceException` at runtime (proving `!!`'s real cost and this
    bug's severity, not just a compile-time difference).
  - Verified directly against the real-world repro: cloned
    `DavidObando/Oahu` at the pre-migration commit named in the issue and
    translated `AudibleJobExecutor.cs` through the fixed translator —
    `convertAction` is now passed bare, no `!!`.
- Ran the targeted regression suites locally (per this repo's
  nullability-adjacent test filters): 578 tests across
  `Cs2Gs.Tests` (`Nullab|Oblivious|2412|Issue4262|3888|Params`,
  including the full `Issue2412CrossProjectObliviousNullabilityTranslationTests`
  suite) — all passing, no regressions in existing nullable-handling tests.
  Also ran `Core.Tests` filtered similarly (669 tests, all passing) as a
  sanity check, though the change is entirely within `Cs2Gs.Translator`.
- Launched the self-hosting PR guard
  (`build/run-cs2gs-selfmig-pr-guard.sh`, `SELFMIG_PR_GUARD_ROOT` set to a
  unique path) to translate/compile/ilverify/test-parity the hot-core apps
  (which mix nullable-enabled `src/*` projects and nullable-oblivious
  `tools/cs2gs/*`/`test/*` projects — exactly this fix's affected shape)
  through the fixed translator; it was still running at hand-off time, so its
  final result should be checked before merge if it hasn't reported back
  through other channels.

## Not fixed (out of scope, by design)

The issue's second, lower-severity instance — `Book.Conversion` (a
non-nullable EF navigation property in oblivious `Oahu.Data`) being emitted
as `Conversion?` in G# — is a *different*, deliberate mechanism: cs2gs always
promotes an EF entity navigation property to nullable in an oblivious
compilation (issue #3060), independent of taint evidence, to avoid a
`NullReferenceException` on a not-yet-materialized/lazy-loaded navigation
property. Verified `Oahu.Data.csproj` has no `<Nullable>` element (oblivious)
and no repo-wide override, so this is the intended #3060 policy operating as
designed, not a bug — changing it would reintroduce the failure mode #3060
fixed. Confirmed this instance never crashed in Oahu (per the issue itself).


DavidObando (commented):
Rubber-duck review against issue #4262's scope. Verdict: solid, approving in spirit (self-approval rejected by GitHub, posting as comment) — with a follow-up fix pushed during review.

## PR's own fix (widened entry gates + inner Disable guard)

Verified the PR's core claim directly:
- Reverting the widened `IsTainted`/`IsParamsElementTainted` entry gates reproduces the exact reported symptom (`convertAction!!` reappears).
- Reverting the `IsTaintedCore` inner `Disable` guard breaks the precision-guard negative-control test (an enabled caller's own `T? = null` local self-taints an unrelated, genuinely-untainted sibling parameter) — confirms the second fix is load-bearing and guards the right compilation.
- Reverse direction (oblivious consumer -> enabled sibling's genuinely non-nullable parameter) still gets `!!`; unaffected by this PR since it's governed by the target's own declared annotation, not the taint fixpoint.
- Three-project chain (enabled A -> enabled C, unrelated oblivious B in the sibling list) does not falsely widen — `RemapToCompilation`'s metadata-identity matching correctly finds no match for an unrelated symbol.
- `Book.Conversion`/#3060 EF navigation-property promotion is untouched, confirmed by reading `SeedEfEntityPropertyTaint` — a separate, deliberate, unaffected policy.
- Re-ran `Issue2412CrossProjectObliviousNullabilityTranslationTests` and the full `Nullab|Oblivious` filter (531 tests) — all pass, no regressions.
- Self-hosting PR guard (`build/run-cs2gs-selfmig-pr-guard.sh`, 8 hot-core apps): 8/8 PASS.

## Follow-up defect found during review (fixed, pushed as bdbf0c6c)

The widened gates make `IsTainted`/`IsParamsElementTainted` reachable for a nullable-ENABLED asking compilation. Three call sites consult that cross-project taint evidence without first requiring the queried declaration's own `NullableAnnotation` to be `None` (genuinely oblivious) — unlike `ShouldPromoteToNullableReference`'s ordinary path, which already requires that before ever calling `IsTainted`.

`IsParamsElementTainted`'s candidate loop matches by a compilation-independent documentation-comment ID rather than remapping by symbol identity the way `IsTainted`'s sibling walk does. Confirmed empirically (failing test before fix, passing after, isolated-revert-verified):

1. **Params consumption-side leak**: an unrelated oblivious sibling B, merely by calling the same target `params` method with a literal `null` somewhere in its own unrelated source, suppressed the `!!` bridge for a completely different, genuinely non-nullable enabled caller elsewhere in the run.
2. **Params declaration-side leak**: the same unrelated evidence repainted the enabled target's OWN emitted `params` element type nullable (`...string` -> `...string?`) when translating that project itself — a real signature-level correctness bug, not just a missing bridge.
3. **Async-return declaration-side leak**: `PromoteAwaitedReturnIfTainted` had the identical gap — an oblivious override/interface-implementation of an enabled `virtual async Task<string> M()` returning `null` repainted the enabled base's own declared return type to `Task<string?>`, via #2285's bidirectional interface/override taint edges.

Fixed both `IsParamsElementTainted` (single guard covering both its callers) and `PromoteAwaitedReturnIfTainted` (matching `None`-annotation guard) the same way `ShouldPromoteToNullableReference` already protects its own call. The nested-property-pattern shape was deliberately avoided (decomposed into a plain cast + `!=`) per the existing `#4127/#4129/#4153` self-hosting pattern-matching hazard noted elsewhere in this file.

Also checked and confirmed safe (no fix needed): the sync interface-member return-type path (already routes through `ShouldPromoteToNullableReference`'s `None` gate) and the delegate-forwarder path in `PromoteDelegateParameterInvokedWithNull` (gated by an explicit `IsObliviousCompilation()` check at its top, unreachable for an enabled asking compilation both before and after this PR).

New tests: `Issue4262UnrelatedObliviousSiblingTaintLeakageTests.cs` (3 tests, each verified to fail without its corresponding fix).

Re-verified after the follow-up fix: full `Nullab|Oblivious|Params` filter (571 tests), `2421|2423|3888|4128|914` filter (119 tests), and the self-hosting guard (8/8 apps) — all green.

## Out of scope, correctly left alone

Tuple-element overloads (`IsTupleElementTainted`, `IsDeconstructedForEachElementTainted`) still carry the original narrow (pre-#4262) entry gate — same #4262 shape could in principle apply to a tuple leaf, but that's unexercised by this issue and is a reasonable follow-up rather than something this PR should have touched.

PR #4274, opened 2026-09-16
Allow ref-returning calls through a by-value Span[T]-like parameter

### LINKED ISSUE #4265: gsc: ref-returning call over a by-value byref-like (Span) parameter always rejected, unlike C#
## Capability gap

Found while reviewing PR #4263 / issue #4224's ref-returning-call escape analysis. Out of scope for #4224 (confirmed non-blocking, no false-negative reachable through it today) but a real, separate gap.

## Reproduced behavior

Built gsc from `fix/issue-4224-ref-returning-call-storage` (`b17d2923`).

```gsharp
import System
func First(s Span[int32]) ref int32 {
    return ref s[0]
}
```

Actual: `GS0254`, reported as returning a pointer to function-local storage — even though `s`'s underlying storage is heap/caller-owned, not the parameter slot itself.

The analogous C# compiles and is used throughout the BCL (e.g. `MemoryMarshal.GetReference<T>(Span<T> span)` returns `ref T` derived from a by-value `Span<T>` parameter).

## Root cause

`StatementBinder.HasFunctionLocalRefScope`'s `BoundVariableExpression` case treats every by-value (non-`ref`/`in`/`out`) parameter as function-local for ref-safety purposes:

```csharp
if (v.Variable is ParameterSymbol p)
{
    return p.IsScoped || p.RefKind == RefKind.None;
}
```

This is correct for an ordinary struct parameter (its fields die with the frame), but overly conservative for a byref-like (ref struct, e.g. `Span[T]`) by-value parameter: the struct itself is a copy, but the *referent* it points to is not the parameter's own storage, so C#'s ref-safe-scope rules give such a parameter "return-safe" (caller) scope, not function-local scope, for this purpose. `src/Core/CodeAnalysis/Binding/StatementBinder.Jumps.cs` around line 606-611.

This predates #4224 (present since the file split in #1361 / originally issue #490) and is orthogonal to #4224's new `RefCapabilities.TryGetRefReturnEscapeSources`/`SelectByRefArguments` machinery — confirmed unreachable through that new code path today, since a native ref-returning function taking a byref-like by-value parameter can't even have its own body accepted, so there's no such function to call from #4224's new escape-source computation. No memory-safety issue: this gap only over-rejects valid programs, it does not admit an invalid one.

## Recommended fix

`HasFunctionLocalRefScope`'s parameter case should special-case a by-value parameter whose type `TypeSymbol.IsByRefLike`: such a parameter's *referent* is safe to escape to the caller (matching C#), while its own struct copy still may not be reassigned/aliased as a variable. `RefCapabilities.SelectByRefArguments` would then also need to include by-value byref-like arguments (using the existing `HasFunctionLocalEscapeScope`/STE walk) once a native ref-returning function over such a parameter can compile at all, so a forwarding chain is escape-checked with the same precision as `ref`/`in`/`out` arguments.

## Definition of done

- [ ] `func First(s Span[int32]) ref int32 { return ref s[0] }`-shaped native functions compile.
- [ ] Forwarding such a result further up (a chain of ref-returning calls over a by-value byref-like parameter) is still escape-checked precisely — a `scoped` by-value byref-like parameter, or one constructed over function-local storage, must still report `GS0254` when forwarded.

DavidObando:
Fixes #4265

## Root cause

`func First(s Span[int32]) ref int32 { return ref s[0] }` always reported `GS0254` ("cannot return a managed pointer to function-local storage"), unlike C#, which allows exactly this (`MemoryMarshal.GetReference[T](Span[T])`-shaped BCL APIs rely on it).

The actual root cause is narrower than the issue's own "Recommended fix" section suggested. `StatementBinder.HasFunctionLocalRefScope` had **no case at all** for `BoundClrIndexExpression` — the node a CLR/imported indexer read such as `Span[T]`'s `s[0]` binds to (wrapped in a `BoundDereferenceExpression` by `ConversionClassifier.AutoDereferenceRefReturn`). It fell straight through to the fully-conservative `default:` branch and reported unconditionally, **never even inspecting `s`**. So the issue's suggested fix (special-case the `ParameterSymbol` case for any byref-like by-value parameter) never actually gets reached for this repro.

## What I changed

1. Added a `BoundClrIndexExpression` case to `HasFunctionLocalRefScope` that delegates to the indexed target's scope.
2. Added a narrow `HasFunctionLocalReferentScope` helper, used **only** when the indexed target's type `TypeSymbol.IsByRefLike` — it mirrors the file's existing `BoundDereferenceExpression` pointer-parameter precedent exactly (`expr is BoundVariableExpression { Variable: ParameterSymbol p } ? p.IsScoped : HasFunctionLocalRefScope(expr)`).
3. Split `RefCapabilities`' argument-escape selection (formerly `SelectByRefArguments`, now `SelectEscapeArguments`) into two separate lists: `byRefArguments` (unchanged, checked via `HasFunctionLocalRefScope`) and a new `byValueByRefLikeArguments` (checked via the new `HasFunctionLocalReferentScope`), so a forwarding chain through a by-value byref-like parameter (`return ref First(s)`) is escape-checked with the same precision as an existing `ref`/`in`/`out` forwarding chain.

## Why I diverged from the issue's own recommendation

The issue's "Recommended fix" proposed special-casing the `ParameterSymbol` case in `HasFunctionLocalRefScope` directly: treat *any* non-scoped by-value byref-like parameter as caller-scoped. **I implemented that first, and it is unsound.** It conflates two different C# ref-safe-scope concepts:

- A by-value ref-struct parameter's **own storage** (its fields) is still a function-local copy, exactly like any other by-value struct — this must stay rejected.
- Only the **referent** such a value *encapsulates* (`Span[T]`'s backing pointer, reached only through its indexer/methods, never through a field) has caller scope, because the caller already guaranteed that referent's lifetime by constructing/passing the value.

Concretely, with the naive parameter-case carve-out, this compiled (and is genuinely unsafe — `a.Total` lives in `Bad`'s own stack frame):

```gsharp
ref struct Acc {
    var Total int32
}
func Bad(a Acc) ref readonly int32 {
    return ref a.Total
}
```

I caught this via a second advisor review before committing, reverted the parameter-case carve-out back to its original conservative form, and instead added the narrower `HasFunctionLocalReferentScope` helper described above, applied only where a ref genuinely passes *through* an encapsulated reference (a CLR indexer target, and a by-value byref-like argument being forwarded) — never to an ordinary field/property read off a by-value byref-like parameter's own storage.

A related trap: a `ref`/`in`/`out` argument that happens to be byref-like-typed must stay on the strict `HasFunctionLocalRefScope` check, not the permissive referent-scope one — using the permissive check there would reopen the same hole from a different angle:

```gsharp
ref struct Acc {
    var Total int32
}
func Fwd(ref a Acc) ref int32 {
    return ref a.Total
}
func Bad(a Acc) ref int32 {
    return ref Fwd(ref a)   // must still report GS0254
}
```

This is why `RefCapabilities` now returns two separate argument lists rather than one, and why `HasFunctionLocalReferentScope` is a distinct helper from `HasFunctionLocalRefScope` rather than a tweak to the existing `ParameterSymbol` case.

## What I tested

New test file `test/Core.Tests/CodeAnalysis/Binding/Issue4265SpanByValueRefReturnTests.cs` (8 tests, all binder-level with `EmittedOracle` for the emit+execute cases):

- **Positive**: `Span[int32]` by-value parameter, `return ref s[0]` compiles; alias mutation through the returned ref is observed on the caller's original array.
- **Positive**: a non-`scoped` forwarding chain (`Forward(s) { return ref First(s) }`) compiles and correctly aliases the original storage.
- **Negative**: a `scoped` by-value `Span[int32]` parameter forwarded through another ref-returning call still reports `GS0254`.
- **Positive**: `ReadOnlySpan[int32]` with a `ref readonly` return compiles.
- **Negative**: `ReadOnlySpan[int32]` with a plain (mutable) `ref` return still reports `GS0253` (its indexer only ever yields `ref readonly`).
- **Negative** (the unsafe case this change must NOT relax): an ORDINARY (non-ref-struct) by-value struct parameter returning a ref to one of its own fields still reports `GS0254`.
- **Negative** (caught during review, see above): a ref-struct by-value parameter returning a ref to its OWN field still reports `GS0254`.
- **Negative** (caught during review, see above): the `Fwd(ref a)`-forwarding-a-ref-struct-field chain still reports `GS0254`.

Also ran, all green against the final code:
- `dotnet build` Release (both the touched project and the full `GSharp.sln`): 0 warnings, 0 errors.
- Full `Core.Tests` Binding-namespace suite (6218 tests) — no regressions.
- Targeted `Compiler.Tests` filter (`Span|RefReturn|RefStruct|ByRef`, 92 tests) — no regressions.
- Targeted `Core.Tests` filter covering #4224/#4263's own ref-return/escape-analysis tests plus `RefSafeEscapeTests`, `Issue4220ReadOnlyRefTests`, `ScopedRefKindCompositionTests` (111 tests) — no regressions.

Also manually verified via `gsc` CLI that the two unsound repros above (ref-struct-own-field direct and forwarded-through-`ref`) are correctly rejected with `GS0254` after the fix, and were (wrongly) accepted with the first, reverted implementation.

Release notes updated (`website/docs/release-notes.md`, Unreleased/Fixed section).

Nothing deferred from the issue's Definition of Done: both listed shapes (`First(s Span[int32])` compiling, and a forwarding chain staying precisely escape-checked including the `scoped` negative case) are covered.

## Soundness hardening (found in independent rubber-duck review, fixed before merge)

The `BoundClrIndexExpression` case above treats **any** byref-like CLR indexer target as encapsulating a caller-supplied referent, the way `Span[T]`/`ReadOnlySpan[T]` genuinely do — but nothing in the CLR type system enforces that. An externally-compiled ref struct can legally declare its indexer `[UnscopedRef]` (`System.Diagnostics.CodeAnalysis`) — a real C# 11+ pattern, no `unsafe`/IL authoring required — meaning the indexer returns a `ref` into the **receiver's own storage** instead of an encapsulated referent. That's exactly the `Acc.Total` shape this PR's own guard exists to reject, just authored in a referenced assembly instead of natively.

Confirmed exploitable end-to-end with a real compiled C# library (no hand-authored IL), referenced via `/r:`:

```csharp
public ref struct RingBuffer {
    private int _value;
    public ref int this[int i] {
        [UnscopedRef]
        get { return ref _value; }
    }
}
```

```gsharp
func M(buf RingBuffer) ref int32 {
    return ref buf[0]
}
```

compiled clean before this hardening, and the returned reference pointed at dead stack memory — reading through the alias after an intervening call reused the frame returned an unrelated stack value (`before=7 after=114` in the repro), i.e. genuine memory corruption. Real C# rejects the analogous code with `CS8166` (`[UnscopedRef]` inverts the receiver's contribution to the call's ref-safe-context from safe-to-escape back to function-local for a by-value receiver).

Fix: `RefCapabilities.IsUnscopedRefIndexerGetter` inspects the resolved indexer for `[UnscopedRef]` via `CustomAttributeData` (so it works correctly through gsc's `MetadataLoadContext`-based reference loading), checking **both** the property and its `get` accessor — C# allows the attribute on either, and an expression-bodied indexer (`[UnscopedRef] public ref int this[int i] => ref field;`) emits it on the property rather than `get_Item`, confirmed against a real compiled indexer of each shape. `HasFunctionLocalRefScope`'s `BoundClrIndexExpression` case now falls back to the strict scope check whenever either carries it — matching C#, and still correctly accepting the same member forwarded through a genuine `ref` parameter (verified against `csc` directly), where the referent legitimately is the caller's own storage.

Two new fixture types (`UnscopedRefIndexerFixture`, getter-level attribute; `UnscopedRefIndexerPropertyLevelFixture`, property-level) and three new tests cover both attribute placements plus the `ref`-parameter positive case.


DavidObando (commented):
Rubber-duck review against issue #4265's scope, treating this as memory-safety-critical (ref-safety/escape-analysis).

**Verdict: solid after one soundness fix, which I found during review and pushed to this branch (commit `12afd97`).**

## What I verified

- Read #4265 in full, this PR's diff in full, and applied the `ponytail`/`caveman` review skills.
- Built gsc from the branch and reproduced, independently, both of this PR's own "unsound counter-example" repros:
  - `ref struct Acc { var Total int32 } func Bad(a Acc) ref readonly int32 { return ref a.Total }` — still `GS0254`.
  - `func Fwd(ref a Acc) ref int32 { return ref a.Total }` called as `Fwd(ref a)` from a by-value `a Acc` — still `GS0254`.
- Confirmed the positive cases: `Span[T]` by-value param `return ref s[0]`, a forwarding chain through a non-scoped `Span[T]` param (including a **two-level** forwarding chain, `Level3 -> Level2 -> Level1`), `ReadOnlySpan[T]` with `ref readonly`, and a `scoped`+non-`scoped` mixed-parameter case where only the non-scoped one is forwarded.
- Confirmed `scoped` rejection when forwarded, including a variant where the `scoped` byref-like parameter is a *different declared type* than a sibling non-scoped one in the same call.
- Probed the "encapsulated caller storage vs. own value fields" line hard with a hand-rolled native G# `ref struct` exposing its own indexer (`prop this[i int32] ref int32 { get { return ref Items[i] } }` and a variant returning `ref` to an own scalar field): every native shape is **unconditionally rejected** (`GS0253`/`GS0254`) regardless of what it actually wraps, because a native indexer routes through `BoundUserInstanceCallExpression`'s existing strict receiver-scope check, never the new permissive `BoundClrIndexExpression` path. Over-conservative, not unsound — the permissive path is reachable only through a genuine CLR/imported indexer.
- **That's where I found a real hole.** The permissive `BoundClrIndexExpression` path trusts *any* byref-like CLR indexer target as safe to forward, on the premise that it always encapsulates a caller-owned referent the way `Span[T]`/`ReadOnlySpan[T]` do. That premise doesn't hold for an externally-compiled ref struct whose indexer is legally marked `[UnscopedRef]` — a real, unsafe-free C# 11+ pattern — which instead returns a `ref` into the *receiver's own storage*. I compiled a small real C# library with such a type, referenced it into gsc via `/r:`, and confirmed:
  - `func M(buf RingBuffer) ref int32 { return ref buf[0] }` compiled clean.
  - The returned reference pointed at dead stack memory: reading through the alias after an intervening call reused the frame returned an unrelated value (`before=7`, `after=114` in my repro) — genuine memory corruption, not just a theoretical gap.
  - Real C# rejects the analogous code with `CS8166` (verified against `csc` directly).
  - The hole exists for **both** legal attribute placements (`[UnscopedRef]` on the `get` accessor, and on the property itself via the expression-bodied indexer form) — confirmed each independently reflects the attribute onto different metadata (getter vs. property), and my first fix attempt only checked the getter and missed the property-level placement until I re-verified against a second compiled repro.
- Fixed by checking the resolved indexer (both the property and its getter) for `[UnscopedRef]` via `CustomAttributeData` (`RefCapabilities.IsUnscopedRefIndexerGetter`) and falling back to the strict scope check when present. Verified: both attribute placements now correctly report `GS0254`; the same member forwarded through a genuine `ref` parameter still correctly compiles (matches `csc`); all original 8 PR tests plus 3 new regression tests pass; the full `Core.Tests` Binding-namespace suite (6191 tests) and the `Compiler.Tests` `Span|RefReturn|RefStruct|ByRef` filter (92 tests) pass with no regressions.
- Confirmed `SelectEscapeArguments`'s split of `byRefArguments`/`byValueByRefLikeArguments` is exhaustive: it's a private method with a single call site (`TryGetRefReturnEscapeSources`), which itself has exactly one caller (`HasFunctionLocalRefScope`'s `default:` branch), and both out-lists are consumed there — no other call site to miscategorize.
- Test coverage and release notes are accurate; I updated the release-notes bullet and PR description with the soundness-hardening details above.

Given the severity (a reachable, real-world memory-corruption path via a fully legitimate, unsafe-free C# library — not hand-authored IL), I pushed the fix directly to this branch rather than leaving it as a review comment. Given I authored part of what's being approved here, flagging for a second human pair of eyes is reasonable before merge, but I'm confident in the fix as verified.

PR #4275, opened 2026-09-16
Fix gsfmt splitting a C-style for-clause header across three lines

DavidObando:
## Summary

`gsfmt` was splitting a C-style `for init; cond; post { ... }` loop header across three lines, one per clause, instead of keeping it on one line (matching the ADR-0179 canonical form: one line up to 120 columns, wrap only if genuinely too long).

**Root cause:** `SyntaxFacts.IsBreakRequiredBetween` treats every `;` as a statement terminator that forces a line break — correct for an ordinary statement, but wrong for the two separators inside `ForClauseStatementSyntax`, which are list separators like `,` rather than terminators. `IsBreakRequiredBetween` is a two-token predicate with no access to the owning syntax node, so it can't tell the two cases apart. The fix lives in `GSharpFormatter`'s `LayoutBuilder` instead: a `;` whose direct parent is `ForClauseStatementSyntax` is now treated the way a `,` already is — no forced break, a wrap point (`Doc.Line`) that collapses to one space when the header fits and indents by 4 when it doesn't. The layout also re-derives the break at that boundary rather than preserving an author-written one (`IsContinuationBoundary`), so hand-formatted three-line headers already in the repo collapse too, not just freshly-formatted input.

A related spacing defect in the same code path: an empty initializer clause (`for ; cond; post {`) left the separator glued to the `for` keyword as `for;`, because the pre-existing "no space before `;`" rule assumed a `;` always closes a preceding token. It now special-cases a for-clause separator with nothing to its left.

`samples/Loop.gs`, `samples/VariadicDelegate.gs`, `samples/VariadicMethods.gs`, `samples/WhileAndLabeledLoops.gs`, and `src/Sdk/Gsharp.Extensions/Sequences/Sequences.gs` carried the three-line form (`gsfmt --check` considered them already formatted) and are reformatted in this PR so the repo stays clean under `gsfmt --check .`.

## Other line-breaking issues found and fixed

While investigating, I found one more: ADR-0179's own phase-7b measurement table lists `&&`/`||`/`??`/`or`/`and` chains together as a family that goes fully to zero reducible long lines, but only `&&`, `||`, and the pattern-disjunction `or`/`and` had a wrap point implemented — a `??` chain had none and could overflow the canonical 120-column width with no way to break. Confirmed with a 121-character `??` chain that gsfmt left unbroken. `??` now gets the identical treatment as `&&`/`||`.

I also probed a handful of other binary operators (`-`, `*`, `==`, `=`, `->`) that have no wrap point either, but ADR-0179's own inventory of what a formatter is expected to break never lists them, so leaving them alone is consistent with the design rather than a gap. After the fixes above, a repo-wide scan of every `.gs` file for lines over 120 columns turns up exactly one line, a `//` comment, which ADR-0179 explicitly says gsfmt should not reflow (comment wrapping was considered and rejected in the ADR's open questions).

No other genuine formatting defects were found.

## Testing

- `dotnet build` (Release) across the full solution — 0 warnings, 0 errors.
- New regression tests in `test/Formatting.Tests/GSharpFormatterTests.cs` (join, idempotence-on-already-joined-input, and four empty-clause spacing cases) and `test/Formatting.Tests/WrappingTests.cs` (an over-long for-header that wraps at its separators, an over-long `??` chain, plus both added to the shared wide-shapes property test for width/idempotence coverage).
- Full `Formatting.Tests` suite: 103/103 passing.
- Idempotence verified directly on the repro cases (`gsfmt` run twice produces identical output) and covered by the existing `Format_IsAFixedPointOnItsOwnWrappedOutput`-style property tests.
- `gsfmt --check .` clean across the whole repository.
- Full `Cs2Gs.Tests` suite (cs2gs runs gsfmt as a default post-pass per ADR-0179 phase 7b): 3168/3168 passing, no regressions.
- Release notes updated (`website/docs/release-notes.md`, Unreleased → Fixed).

Nothing was deferred.


DavidObando (commented):
Rubber-duck reviewed (self-approval rejected by GitHub, posting as comment instead — verdict is approve). Built branch, verified every claim independently, not just re-read the diff.

**Core fix.** 3-line C-style for-header collapses to one line, both freshly-authored and hand-broken-then-reformatted. Overlong header (>120 cols) wraps at the separators, stays under 120, threshold verified exact at the 120/121-col boundary. Empty-clause spacing correct in all 4 combos (empty init, empty cond, empty post, all-three-empty) — no more `for;`.

**`??` chain bug.** Reproduced pre-fix by reverting just the two `QuestionQuestionToken` hunks and rebuilding: a 9-link `??` chain sat at 160 cols with zero wrap. Restored, rebuilt: wraps correctly, all lines ≤120. Non-chained short `??` unaffected (stays one line), parenthesized precedence (`a && (b ?? c)`) unaffected.

**Idempotence.** Ran gsfmt twice on all 5 reformatted files plus 10 of my own constructed cases (basic for-loop, hand-broken for-loop, overlong for-loop, 4 empty-clause variants, the `??` chain, a for-header with a trailing line-comment on the init clause, a for-header with a block comment). All byte-identical on the second pass.

**5 reformatted files.** Diffed each against pre-PR content myself — only the for-header whitespace changed, exactly matching the PR diff, nothing semantic.

**Tests.** Formatting.Tests: 103/103. `gsfmt --check .`: clean, exit 0.

**Re-derive-break concern.** Checked whether collapsing hand-authored breaks could clobber a deliberate 3-line header (e.g. for a comment, or right at 120 cols). Non-issue: a line-comment on the init clause still forces the only break it needs (can't do otherwise — it's a line comment), and gsfmt is already unconditionally opinionated about continuation boundaries elsewhere (&&/||/chains), so this isn't new behavior, just the same rule reapplied to `;`.

**Other line-breaking constructs probed** (not in PR scope, sanity check only): long generic-constraint list wraps fine via the parameter list. Long method chain (`.Where(...).Select(...)...`) wraps correctly at each `.`, idempotent. Long import list: not applicable, G# imports are one per line already.

**One real gap found, not fixed:** a long `?:` ternary chain (`ConditionalExpressionSyntax`, distinct from G#'s block-form `IfExpressionSyntax` which the ADR-0179 phase-7b table already covers) has no wrap point at all — a 134-col ternary chain formats unbroken. This is pre-existing, untouched by this PR, and out of its stated scope, but it's the same class of defect as the `??` bug this PR just fixed. Not a quick fix like `??` was: the `:` token already carries a "no space before" hug-left rule the same way the pre-fix `;` did, so adding a wrap point means working out the interaction the same way this PR did for the empty for-clause, plus a nesting/indent decision for chained/nested ternaries. Flagging as a follow-up rather than fixing inline.

**Merge note:** `gh pr view` reports CONFLICTING against main. Verified the only actual conflict is `website/docs/release-notes.md` (expected — release notes churn across merges) — confirmed no `.gs` file changed on main since the PR's base commit, so this isn't masking a formatting regression, just needs an ordinary rebase before merge.

Verdict: approve. Scope matches its stated claims, verification held up under independent testing.

Issue #4220, opened 2026-09-14
gsc: support ref readonly returns and local aliases without weakening readonly contracts


DavidObando:
## Capability gap

Support **readonly managed-reference contracts** on source methods, computed properties/indexers and local aliases, with faithful CLR interop.

This is distinct from writable-ref returns and ref-returning literals: an ordinary named method or property should be able to return a live alias without granting permission to overwrite the referent.

Follow-up to #4219 and the explicitly deferred readonly half of #3879 / ADR-0060. Writable-ref properties already exist and must remain supported.

## Reproduced evidence

Compiler rebuilt from `07d64f9d`: gsc `0.4.635+07d64f9d8c`, SDK `10.0.400`.

The natural extension of the existing named-function spelling is currently not accepted:

```gsharp
// Proposed syntax, not a claim about an already approved spelling.
package P
func View(ref value int32) ref readonly int32 {
    return ref value
}
var value = 42
Console.WriteLine(View(ref value))
```

Actual: **GS0113** (`Type 'readonly' doesn't exist`) and parser diagnostics including **GS0005**. The parser treats `readonly` as the type after `ref`.

More importantly, there is no source return-ref model for readonly refs, and cs2gs explicitly gaps readonly-ref properties/indexers. This is not inferred solely from one speculative spelling failing.

### C# semantic control: a readonly alias is not a snapshot

This fixture was compiled with C# 13:

```csharp
namespace Parity;
public sealed class RefSource
{
    private int value = 10;
    public ref readonly int View => ref value;
    public void Set(int next) => value = next;
}
```

The executed driver obtained `ref readonly int view = ref source.View`, called `source.Set(42)`, and then read `view`: the result was **42**, proving live alias identity. A separate C# compile rejected assignment through a readonly alias.

### Existing G# controls narrow the actual gap

G# **already consumes the value of this imported readonly-ref property**:

```gsharp
package P
import Parity
let source = RefSource()
Console.WriteLine(source.View)
source.Set(42)
Console.WriteLine(source.View)
```

With a reference to the C# fixture assembly, this compiled and printed **10**, then **42**. Existing imported `ReadOnlySpan` handling also has readonly-return checks. Do not describe all readonly-ref interop as absent.

Also executed:

```gsharp
package P
func Run() {
    var value = 10
    let ref alias = value
    alias = 42
    Console.WriteLine(value)
}
Run()
```

[...]

Issue #893, opened 2026-06-19
Invalid program using `func () R` as parameter to method call


DavidObando:
From `/Users/davidobando/GitHub/DavidObando/Oahu/tests/Oahu.Cli.Tests/DoctorServiceTests.gs`
```
package Oahu.Cli.Tests

import System
import System.Collections.Generic
import System.IO
import System.Linq
import System.Net.Http
import System.Threading
import Oahu.Cli.App.Doctor
import Xunit

class DoctorServiceTests {
    @Fact
    async func RunAsync_WithSkipNetwork_DoesNotMakeHttpCalls() {
        var svc = DoctorService(httpClientFactory: func() HttpClient {
            throw InvalidOperationException("HTTP must not be invoked when --skip-network is set")
        })

        var report = await svc.RunAsync(DoctorOptions() { SkipNetwork = true }, CancellationToken.None)

        let net = report.Checks.Single(func(c DoctorCheck) bool { c.Id == "audible-api" })
        Assert.Equal(DoctorSeverity.Ok, net.Severity)
        Assert.Contains("skipped", net!!.Message)
    }
```

Produces this error:
```
  Oahu.Cli.Tests test net10.0 failed with 1 error(s) (7.7s)
    /Users/davidobando/GitHub/DavidObando/Oahu/tests/Oahu.Cli.Tests/bin/Debug/net10.0/Oahu.Cli.Tests.dll : error TESTERROR: 
      Oahu.Cli.Tests.DoctorServiceTests.RunAsync_WithSkipNetwork_DoesNotMakeHttpCalls (2ms): Error Message: System.InvalidProgramException : Common Langua
      ge Runtime detected an invalid program.
      Stack Trace:
         at Oahu.Cli.Tests.App.<Program>.<lambda3>(DoctorCheck c)
         at System.Linq.Enumerable.TryGetSingle[TSource](IEnumerable`1 source, Func`2 predicate, Boolean& found)
         at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source, Func`2 predicate)
         at Oahu.Cli.Tests.DoctorServiceTests.<RunAsync_WithSkipNetwork_DoesNotMakeHttpCalls>d__0.MoveNext()
      --- End of stack trace from previous location ---
      --- End of stack trace from previous location ---
```

Related to #891

PR #4276, opened 2026-09-16
Issue #4219: non-generic local-function recursion groups + ref-returning function literals

### LINKED ISSUE #4219: gsc: C# capability parity for mutual-recursion groups and ref-returning function literals

DavidObando:
## Summary

Progress on #4219 (the umbrella issue's own remaining scope, after its five spun-off sub-issues #4220-#4224 were merged this session). Audited both capability areas against current `main`, confirmed what already worked, and implemented the rest that was reasonably scoped for one PR. Does **not** close #4219 — see "Deferred" below.

### 1. Mutual-recursion groups for local functions

**Generic local functions already fully support this** (`b84bf800` plus the #4221/#4223/#4261/#4266 follow-ups): forward references and call cycles among a consecutive run of `let name[T] = func (...) ... {...}` declarations compile and run correctly, including captures, enclosing type parameters, constraints, async, and mixed instantiations. No change needed there — confirmed by the existing 41 tests in `Issue4219GenericLocalRecursionTests`, all still green.

**Non-generic local functions did not.** `let first = func(x int32) int32 { return second(x) }` / `let second = func(x int32) int32 { return first(x) }` reported `GS0130` ("undefined variable") on the forward reference, because `StatementBinder`'s consecutive-declaration-group mechanism required a `[T, ...]` list to opt in. This PR extends the *same* prepare-then-bind signature-registration machinery to a run of **two or more** consecutive non-generic `let name = func (...) ... {...}` declarations (a *lone* non-generic literal is untouched, keeping its existing delegate-valued-variable representation — this is a deliberate compatibility boundary). Captures compose: the capturing-direct-call-local closure-dispatch path built for #4221's generic case (`ClosureEmitter.GenericLocalClosures`) is widened from `Function.IsGeneric` to `Function.LocalDeclaration != null`, so a capturing non-generic group member dispatches correctly and shared/mutated captured state is visible across sibling calls.

One narrow, disclosed limitation: converting a **capturing** group member's bare name to an explicit delegate value now reports `GS0126` instead of working (no single stable closure instance to bind to — matches the pre-existing behavior of a capturing *generic* local function, but narrows the old non-generic delegate-variable baseline for that one shape). A **non-capturing** member's bare-name conversion is unaffected. A mixed generic/non-generic group remains out of scope (`GS0130`, unchanged from before).

### 2. Ref-returning function literals

Not parseable at all on `main` (`ParseFunctionLiteralExpression` had no `ref` return-modifier handling). Implemented the scoped slice: a non-generic `let name = func (...) ref T { ... }` always routes through the direct-call local-function path (standing alone or grouped with another ref-returning sibling), reusing the named-function ref-return machinery in full — `FunctionSymbol.ReturnRefKind`, the existing `return ref` statement binder, #4224's alias/storage machinery, and the existing escape/lifetime checker (`GS0254` still fires for escaping storage, including a captured outer variable).

```gsharp
let at = func(values []int32, index int32) ref int32 {
    return ref values[index]
}
var ref alias = at(values, 0)
alias = 42          // mutates values[0]
let v = at(values, 0)  // ordinary value read
```

## Deferred (why this PR doesn't close #4219)

- **Converting a ref-returning function to a delegate/function-type value** — `DelegateTypeSymbol`/`FunctionTypeSymbol` carry no return-ref-kind, so the conversion would otherwise silently narrow a by-ref function to a by-value one. Reported as the new `GS0588` instead of crashing or silently accepting it. (Named ref-returning functions already convert to an explicit, genuinely ref-aware *imported* delegate type via #4220/#4224 — confirmed unaffected by `Issue4220ReadOnlyRefInteropTests` — this gap is specifically the "no target type, or G#'s own synthesized delegate shape" case.)
- **Generic ref-returning local functions** (`let f[T] = func (...) ref T { ... }`) — also `GS0588`; the MethodSpec/signature-encoding interaction with a by-ref return has not been verified.
- **Capturing group member → delegate conversion** (workstream A, noted above) — `GS0126`, disclosed and pinned, not silently broken.
- Broader parity items the umbrella issue itself scopes as follow-up work (mixed generic/non-generic groups, `ref readonly` literal spelling, shorter/contextually-inferred ref lambdas) are unchanged.

## Verification

- `dotnet build` (Release, full solution): 0 warnings, 0 errors.
- Full `Core.Tests`: 9411/9411 passed.
- Full `Compiler.Tests`: 5998/5998 passed.
- `Formatting.Tests`: 91/91 passed (the new syntax slots are picked up automatically by the formatter's reflection-based `GetChildren` walk, no printer changes needed).
- Targeted `Interpreter.Tests` (LocalFunction/#3030): 5/5 passed.
- `DiagnosticIdUniquenessTests` (catalog/doc-sync guards for the new `GS0588`): 4/4 passed.
- New regression tests: `Issue4219NonGenericLocalRecursionTests` (18 tests) and `Issue4219RefReturningFunctionLiteralTests` (9 tests), covering forward references, cycles, captures, mutation visibility, self-recursion inside a group, escape safety, the deferred-scope diagnostics, and two correctness gaps caught during review (see below) with regression pins.
- Updated `docs/diagnostics.md` / `website/docs/ref/diagnostics.md` for `GS0588`, and `website/docs/release-notes.md`.

Two real bugs were caught and fixed during self-review before this was pushed, both pinned by tests now:
- A ref-returning literal whose body ends in a bare trailing expression (no explicit `return ref`) crashed the compiler (`GS9999`/NullReferenceException) instead of reporting a diagnostic — the #893 trailing-expression-as-implicit-return rewrite always synthesizes a value return, which is incompatible with a by-ref-returning function. Now skipped for a ref-returning literal, so it correctly reports "not all paths return a value" (`GS0100`).
- A grouped `let` with an explicit, *mismatched* outer type clause (`let f (int32) -> string = func(x int32) int32 {...}`) silently compiled and ran with the wrong apparent type, because the direct-call group's signature comes entirely from the literal and was ignoring the declared outer type. Both new predicates now exclude an explicit outer type clause, falling back to the ordinary path, which type-checks it correctly.


DavidObando (commented):
Rubber-duck review complete. Verdict: **solid, approve-quality** (self-approval blocked by GitHub since this is the PR author's own account — posting as a review comment instead). One defect found and fixed during review (now part of the branch, commit `5f671d54`).

## What I independently verified by building gsc and testing directly

**Scope vs #4219.** Confirmed the umbrella issue's five sub-issues (#4220-#4224) are merged, and this PR addresses the two remaining items: non-generic local-function mutual recursion, and ref-returning function literals.

**Generic recursion pre-existing.** Built `main` (pre-PR) and confirmed a generic recursive pair (`let first[T]`/`let second[U]` calling each other) already compiled and ran correctly with zero code changes needed — matches the PR's claim.

**Non-generic recursion.** On `main`, the same pattern without `[T]` reported GS0130 as expected. On this branch: 2-member cycles work; I additionally tested a hand-written 3-member cycle (not just the PR's own test) — compiles and runs correctly. Mixed generic/non-generic groups correctly remain GS0130 (verified out-of-scope case is genuinely still rejected, not silently broken).

**Capturing-delegate-conversion narrowing (GS0126).** Verified precisely: on `main`, a single capturing non-generic local converting to an explicit delegate type worked. Verified this is preserved post-PR for a *lone* declaration (untouched). For a *genuinely recursive* capturing group member, GS0126 now fires as disclosed — matches pre-existing generic-side behavior, a reasonable, well-tested, clearly-diagnosed tradeoff given the architecture (per-call closure materialization has no single stable instance to bind a delegate to).

**Found and fixed during review: the grouping gate was pure syntactic adjacency, not actual recursion.** Two *unrelated* capturing locals declared back-to-back (`let onClick = func() {count++}`; `let onLog = func() {...}`, with no cross-reference between them — the ordinary callback-registration shape) were being swept into the same direct-call group as a genuinely recursive pair, and lost delegate-conversion capability even though nothing about them is recursive. This is broader and more surprising than the PR's disclosed "capturing group member" narrowing. Confirmed via CLI repro against both `main` and the original PR branch. Fixed by gating group formation on an actual syntactic cross-reference between members (a bare name or direct call naming a sibling); self-reference alone (ordinary self-recursion) doesn't count, since that already works without grouping. Added two regression tests. Full Core.Tests (9411/9411) and Compiler.Tests (5998/5998) confirmed green with the fix applied.

**Ref-returning function literals.** Verified: alias mutation through a call result, plain value reads, forwarding through a sibling (`return ref at(values, 0)`), self-recursion, `ref readonly` literals, and legal forwarding of a `ref` parameter. Escape safety: returning ref to a function-local var and to a captured outer variable both correctly report GS0254 (reusing the existing, unmodified escape checker — confirmed identical diagnostic for the equivalent named-function shape, so this is genuine reuse, not new leniency).

**GS0588.** Fires correctly for a bare-name delegate conversion of a ref-returning local, a ref-returning literal passed as an argument, a `var`-declared ref-returning literal, and a generic ref-returning local. Ran `Issue4220ReadOnlyRefInteropTests` (36 tests, real IL-level interop) directly — confirmed the named ref-returning function → imported delegate path (#4220/#4224) is unaffected by the new guard.

**Self-review bug fixes (all confirmed real, not just claimed).** Trailing-expression ref-returning literal now reports GS0100 instead of crashing. Grouped `let` with a mismatched explicit outer type clause now correctly falls back to the type-checked path (GS0155). Outer-name shadowing correctly reports GS0266, matching the pre-existing generic-group precedent.

**Docs.** GS0588 present and identically worded in `docs/diagnostics.md` and `website/docs/ref/diagnostics.md`. Release notes accurate (and remain accurate after the fix — the "capturing group member" narrowing they describe is now correctly scoped to actual cross-referencing groups).

**Minor nits (non-blocking):**
- PR body test-count claims are off: `Issue4219NonGenericLocalRecursionTests` has 12 cases, not 18 (now 14 with the 2 regression tests I added); `Issue4219RefReturningFunctionLiteralTests` has 10, not 9.
- A capturing ref-returning literal's delegate conversion reports GS0126 rather than GS0588 — inconsistent label, but this combination requires a captured variable that's also legally ref-returnable, which doesn't appear constructible today (GS0254 fires first in every case I could construct), so it's effectively unreachable.

## Full local suites (independently run, not just trusted from the PR report)
- Core.Tests: 9411/9411 (pre-fix and post-fix)
- Compiler.Tests: 5998/5998 (pre-fix and post-fix, ~55min each)
- Formatting.Tests: 91/91
- Interpreter.Tests (targeted): 5/5
- DiagnosticIdUniquenessTests: 4/4

## Self-hosting guard
Ran it (not run by the PR author) with a unique `SELFMIG_PR_GUARD_ROOT`, on the fixed branch. **PASSED** — all 8 hot-core projects (including `src/Core.csproj`, the direct target of this PR's binder/emitter changes) translate, compile, ilverify, and test-parity cleanly under self-migration.

## What I pushed
One commit (`5f671d54`) fixing the syntactic-adjacency grouping regression described above, with two new regression tests. No other changes.

Issue #845, opened 2026-06-14
Gsharp Extensions code docs


DavidObando:
The Gsharp.Extensions.dll that ships with the SDK doesn’t have code docs (/// comments) for the classes and methods shipped. Consumers expect information on them when they hover over these symbols in a code editor.

PR #4281, opened 2026-09-17
Publish GSharp.CodeAnalysis.Analyzers.Testing as a NuGet package

### LINKED ISSUE #3780: ADR-0169: publish GSharp.CodeAnalysis.Analyzers.Testing as a package instead of co-locating it beside gsc (from #3686)

DavidObando:
Fixes #3780

## What this does

ADR-0169 always described `GSharp.CodeAnalysis.Analyzers.Testing` as a "packable" library. PR #3777 (#3686 M5) shipped it beside `gsc` instead — copied into the SDK nupkg's `tools/compiler` and referenced by a migrated analyzer test project through directory adjacency to `$(GsharpCompilerFullPath)` — as an explicitly-stated interim measure, because the package didn't exist yet. This does the real thing:

1. **`src/Analyzers/GSharp.CodeAnalysis.Analyzers.Testing/GSharp.CodeAnalysis.Analyzers.Testing.csproj`** is now packable, following the exact convention already used by `Gsharp.Repl`/`Gsharp.Gsfmt`/`Gsharp.NET.Sdk`: `PackageId`, `IsPackable`, `GeneratePackageOnBuild`, license/readme/icon from `assets/`. No new packaging infrastructure — same NerdBank.GitVersioning-derived version every other package here uses, same build pipeline (`dotnet pack` -> `out/bin/<Config>/nupkgs/`).
2. **`src/Compiler/Compiler.csproj`** drops the `ProjectReference` to the Testing project and its explanatory comment — the co-location is gone.
3. **`tools/cs2gs/Cs2Gs.Pipeline/GSharpProjectTransformer.cs`** (`RewriteAnalyzerProject`) now injects a `<PackageReference Include="GSharp.CodeAnalysis.Analyzers.Testing" Version="…" />` for a migrated analyzer test project instead of a `Reference`/`HintPath` pointed at the compiler's directory. `GSharp.Core` is untouched — it's still compiler-hosted and still resolved the same way, this issue is only about the verifier.
4. **Version resolution**: `Transform` takes a new `analyzerVerifierPackageVersion` parameter. `SdkCompileRunner.ResolveAnalyzerVerifierPackageVersion` resolves it the same way `ResolveSdkMoniker` already resolves the `Gsharp.NET.Sdk` version — from `out/bin/<Config>/nupkgs/`, falling back to the repo's `.nugs` local feed (already wired up in the root `nuget.config` for exactly this local-package-without-a-real-feed scenario) — and stages the resolved nupkg into `.nugs` so restore can find it. `GsharpTestProjectRunner`'s existing SDK-package-resolution helpers (`ResolveNewestSdkPackage`, `ParseVersion`, `ResolveLocalSdkPackage`) are generalized to take a package id instead of duplicating the logic; every existing call site is untouched (the SDK-specific names now delegate to the generalized ones). `MigrationPipeline`/`TranslateStage` thread the resolved version through for repository-layout migrations, the same way `RepositorySdkMoniker` already does; `SdkCompileRunner.CompileMirroredProject` resolves it inline and degrades to `SdkCompileResult.Unavailable(...)` (same as a missing SDK nupkg) when no local nupkg exists yet, rather than failing hard.

## The SDK-nupkg question — decision: remove it entirely

Investigated what actually consumes `GSharp.CodeAnalysis.Analyzers.Testing.dll` via the SDK path vs. the direct-compile path, per #3777's own stated goal of supporting both:

- Nothing in `Gsharp.NET.Sdk`'s props/targets (`Gsharp.NET.Sdk.props`, `Gsharp.NET.Core.Sdk.targets`, `BuildTask.cs`) references it by name — only `GsharpCompilerFullPath` (`gsc.dll`) is resolved there.
- It only ever reached `tools/compiler/` in the SDK nupkg as a side effect of `Compiler.csproj`'s `ProjectReference`: `PackGsharpCompiler` publishes `Compiler.csproj`, and the publish closure pulled the Testing assembly along for the ride.

Once the `ProjectReference` is removed from `Compiler.csproj`, that closure shrinks and the assembly stops appearing in `tools/compiler/` with **no further change needed**. Verified directly: built the SDK nupkg before and after — before, `tools/compiler/` contained `GSharp.CodeAnalysis.Analyzers.Testing.dll`; after, it does not (only `GSharp.Core.dll`/`.pdb`/`.xml` remain, which `gsc` genuinely needs). A migrated analyzer test project now gets the verifier exclusively via its own `PackageReference`, on both the SDK and direct-compile paths — the SDK nupkg does not need to carry it at all.

## A bug found along the way

Marking the Testing project's `ProjectReference` to `Core.csproj` as `Private="false"` was not enough by itself: `dotnet pack` still emitted a nuspec `<dependency id="GSharp.Core" version="{same version as this project}">`, which is not a real NuGet package (`Core.csproj` isn't itself packable) and broke restore for every consumer with `NU1101`. Fixed with `PrivateAssets="all"` on that `ProjectReference`, which stops it from being declared as a package dependency at all — `GSharp.Core` keeps reaching a migrated test project only through the existing `HintPath` `Reference` (compiler/host-supplied, copied into the test project's own output).

## Local package resolution without real publish infrastructure

This repo already has exactly the mechanism needed: `.nugs/` is a local NuGet folder feed wired up in the root `nuget.config` ("Local folder for built-from-source GSharp packages") and already used the same way for `Gsharp.NET.Sdk` (see `e2etests/packageref-e2e.sh`, `e2etests/nuget-pack-e2e.sh`, `SdkCompileRunner.ResolveSdkMoniker`). This change reuses it as-is rather than inventing anything new — `ResolveAnalyzerVerifierPackageVersion` copies the locally-built `GSharp.CodeAnalysis.Analyzers.Testing` nupkg into `.nugs/` exactly like the SDK nupkg already is.

## Docs

`docs/cs2gs-analyzer-translation.md`'s "Detection" section described both `GSharp.Core` and the verifier as "copied to the test output" — updated to describe the verifier as an ordinary `PackageReference` now. ADR-0169 itself needed no change; it already described the library as packable and never documented the interim co-location. No release-notes entry: this is internal migration-tooling/packaging infrastructure, invisible to end users of `gsc` or G# programs (nothing in the "Unreleased" release-notes section documents infrastructure changes like this — it's exclusively language/compiler-diagnostic entries).

## Testing

- `dotnet build GSharp.sln -c Release` — 0 warnings, 0 errors; confirmed `GSharp.CodeAnalysis.Analyzers.Testing.*.nupkg`/`.snupkg` are produced under `out/bin/Release/nupkgs/`.
- Inspected the produced nupkg: `lib/net10.0/GSharp.CodeAnalysis.Analyzers.Testing.dll` + matching `.xml` (fixed a pre-existing double-`GSharp.`-prefixed doc-file name, same fix as issue #845's `Gsharp.Extensions.csproj`), correct nuspec metadata, and — after the `PrivateAssets="all"` fix — no bogus `GSharp.Core` dependency (confirmed the SDK/`Gsharp.Extensions`-style `PackageReference` on `Microsoft.VisualStudio.Validation` is still a real, correct dependency).
- Confirmed the SDK nupkg's `tools/compiler/` no longer carries the Testing assembly (see the SDK-nupkg section above).
- `dotnet test tools/cs2gs/Cs2Gs.Tests` filtered to `Issue3686AnalyzerTestHarnessTests`, `Adr0169AnalyzerProjectTransformTests`, `GSharpProjectTransformerTests`, `Issue3732TransitivePackageReferenceTests`, `Issue3394LocalSdkPackageSelectionTests`, `Issue1752SdkResolutionTests` — 41/41 passed (updated three assertions in `Issue3686AnalyzerTestHarnessTests` that checked for the old `Reference`/`HintPath` shape).
- `dotnet test test/Core.Tests` filtered to `Adr0169AnalyzerVerifierTests` — 15/15 passed.
- **End-to-end via the self-hosting-guard mechanism** (`build/run-cs2gs-selfmig-pr-guard.sh` with `SELFMIG_PR_GUARD_APPS` overridden to just `test/InternalAnalyzers.Tests` + `src/Analyzers/InternalAnalyzers`, its full closure, and a unique `SELFMIG_PR_GUARD_ROOT`): migrated `test/InternalAnalyzers.Tests` end to end against a locally-packed `.nugs`-staged nupkg — translate/compile/ilverify/test-parity all **PASS**, confirming the generated `.gsproj`'s new `PackageReference` actually restores and builds. This is the one app in the repo that exercises `isAnalyzerTestProject`.
- Regenerated `packages.lock.json` for every project whose restore graph changed (`src/Compiler`, `test/Compiler.Tests`, `test/Extensions.Tests`, `test/Interpreter.Tests`, `test/Sdk.Tests` lost the now-removed transitive Testing-project edge; `test/Core.Tests` and `tools/cs2gs/Cs2Gs.Tests`, which reference the Testing project directly, lost the bogus `GSharp.Core` dependency edge) — diffs are exclusively removals, verified with `git diff`.
- Final `dotnet build GSharp.sln -c Release` after all lock-file regeneration — 0 warnings, 0 errors.


DavidObando (commented):
Self-approval is rejected by GitHub (I'm the PR author), so posting as a comment instead — verdict is approve.

Independently re-verified every claim in the PR description; all check out.

**Convention match** — diffed the new `GSharp.CodeAnalysis.Analyzers.Testing.csproj` against `Repl.csproj`/`Gsfmt.Cli.csproj`: `PackageId`/`IsPackable`/`GeneratePackageOnBuild`/license/readme/icon block is byte-for-byte the same shape, same `assets/` relative paths.

**Co-location removal** — `Compiler.csproj`'s ProjectReference + ADR-0169-M5 comment are gone; `Core.csproj` itself is untouched (not in the diff at all), confirming the GSharp.Core scope boundary is deliberate, not missed.

**Transformer** — `RewriteAnalyzerProject` now emits `PackageReference Include="GSharp.CodeAnalysis.Analyzers.Testing" Version="..."`; `GSharp.Core` keeps its HintPath `Reference`. `ResolveAnalyzerVerifierPackageVersion` cleanly reuses the generalized `ResolveNewestPackage`/`ParseVersion`/`ResolveLocalPackage` (SDK call sites unchanged, delegate through).

**SDK nupkg claim** — built the SDK nupkg from this branch myself: `tools/compiler/` contains only `GSharp.Core.dll/.pdb/.xml` + `gsc.*`/`Gsharp.Runtime.Channels.*` + deps, no Testing assembly. Grepped `Gsharp.NET.Sdk`'s props/targets/`BuildTask.cs` — nothing references it by name. Claim holds.

**NU1101 bug fix** — built the Testing nupkg and inspected the actual nuspec: dependency group has only `Microsoft.VisualStudio.Validation 17.8.8`, no `GSharp.Core` entry. `lib/net10.0/` has `GSharp.CodeAnalysis.Analyzers.Testing.dll` + correctly-named `.xml` (not double-prefixed).

**Build/tests**:
- `dotnet build GSharp.sln -c Release` — 0 warnings, 0 errors, nupkg produced.
- `Cs2Gs.Tests` filtered to all 6 named test classes — 41/41 passed.
- `Core.Tests` filtered to `Adr0169AnalyzerVerifierTests` — 15/15 passed.
- Re-ran `build/run-cs2gs-selfmig-pr-guard.sh` myself with a fresh `SELFMIG_PR_GUARD_ROOT` and `SELFMIG_PR_GUARD_APPS` scoped to `test/InternalAnalyzers.Tests` + `src/Analyzers/InternalAnalyzers` (the repo's one analyzer-test app, and a closed set under ProjectReference) — translate/compile/ilverify/test-parity all PASS. This is the real proof the generated `PackageReference` restores and builds end to end, independent of the PR author's own run.

**Lock files** — spot-checked all 6 regenerated `packages.lock.json` diffs (`Core.Tests`, `Sdk.Tests`, `Compiler.Tests`, `Extensions.Tests`, `Interpreter.Tests`, `Cs2Gs.Tests`, `src/Compiler`): every changed file is pure removals, no added lines.

**Missed references** — grepped the whole repo for `Analyzers.Testing`/`AnalyzerVerifier` across docs, website docs, and `.github/workflows`; nothing references the old co-location mechanism or `tools/compiler/`-adjacent resolution. The only doc hits are the ADR itself and usage examples of `GSharpAnalyzerVerifier`, both unaffected by this change.

No defects found. This is exactly the packaging change ADR-0169/issue #3780 called for, with a genuinely-fixed NU1101 landmine along the way.

PR #4282, opened 2026-09-17
selfmig: fix all 21 Cs2Gs.Tests test-parity failures via one CI-only MSBuild bug

### LINKED ISSUE #4278: selfmig: migrated Cs2Gs.Tests still 21/3168 test-parity failures (at floor, non-gating)
## Summary

Investigated the `cs2gs-selfmig-nightly` workflow following the run link in the task
(https://github.com/DavidObando/gsharp/actions/runs/35102316421). That specific run
had already completed successfully by the time it was checked (`conclusion: success`,
single attempt) — but the two threads below are still relevant to the "gate"/"Cs2Gs.Tests"
question the run URL was meant to represent.

### Thread 1 — the actual gate-blocking failure (already fixed, no action needed here)

The three nightlies immediately preceding the given run all failed the `gate` job
(`Apply the baseline` step, `check-selfmig-stage-floor.py`) with `test/Core.Tests/Core.Tests.csproj`
regressing **below its ratchet floor**, `test-parity` → `ilverify`:

- https://github.com/DavidObando/gsharp/actions/runs/34849787331 (2026-09-14 13:31 UTC)
- https://github.com/DavidObando/gsharp/actions/runs/34907577562 (2026-09-14 23:10 UTC)
- https://github.com/DavidObando/gsharp/actions/runs/34975477764 (2026-09-15 13:31 UTC)

Diagnostic (identical fingerprint on all three):
```
ilverify: Error [MethodAccess]: [.../GSharp.Core.Tests.dll :
GSharp.Core.Tests.CodeAnalysis.Emit.Issue4216ReceiverAttributeEmitTests::
NotNullWhen_OnExtensionReceiver_RoundTripsThroughReflection()][offset 0x00000049]
Method is not visible.
fingerprint: sha256:76e60a0828996d39bf043173515d59ced7cf2bf488de715ab06f88f3415c5909
```

This was already tracked as **#4233** and fixed by **#4253** / commit `19dc9e38`
(`src/Core/CodeAnalysis/Symbols/ReferenceResolver.cs`) — `ReferenceResolver.TryResolveType`
was picking an inaccessible type-name duplicate (a private `NotNullWhenAttribute` polyfill
embedded in `Microsoft.TestPlatform.Utilities.dll`) over the real public BCL type when
called with `requireExternalVisibility: false`. That commit landed 2026-09-15 21:07 UTC,
between the last failing nightly and the next nightly
(https://github.com/DavidObando/gsharp/actions/runs/35039622591, 2026-09-16 00:21 UTC),
which is green with `Core.Tests` back at `PPPP`/`test-parity`. **#4233 is closed; no new
issue needed for this thread.**

### Thread 2 — this issue: Cs2Gs.Tests test-parity is still red, but non-gating ("at floor")

In every recent nightly, including the green ones (e.g. the given run
https://github.com/DavidObando/gsharp/actions/runs/35102316421), the self-migration
per-app table reports:

```
| `tools/cs2gs/Cs2Gs.Tests/Cs2Gs.Tests.csproj` | `PPPF` | `test-parity` | at floor (`test-parity`) | `test-parity-failure` |
```

"At floor" means this isn't a ratchet regression (so it doesn't fail the `gate` job), but
the migrated `Cs2Gs.Tests` project genuinely does not reproduce its C# parity oracle:
**21 of 3168** tests fail when the translated-and-recompiled `GSharp.Cs2Gs.Tests.dll` is run
under `dotnet test` (gsc `0.4.717+f5a04750ea`, from run 35102316421, shard `validate (2)`,
artifact `tools_cs2gs_Cs2Gs.Tests_Cs2Gs.Tests.csproj-90f34f7c/test-parity-82e39630ca1e.json`).

Fingerprint: `sha256:82e39630ca1e432ecea1339b2d4a7eabdd44974dcd1f0e2f731c995454e68bd2`

Failing tests (21):
- `Issue2534ImportedBaseCallPipelineTests.Pipeline_ImportedFrameworkBaseCall_EmitsCanonicalSyntaxAndRuns` — `Assert.Contains` can't find `"base.Write(buffer, offset, count)"` in translated output (empty string)
- `Issue2502InheritedStaticPipelineTests.Pipeline_ViaSdkDefault_InheritedGenericStaticCalls_TranslateAndCompile` — can't find `"Product.Create()"`
- `Issue2492NullableBoxedConversionPipelineTests.Pipeline_ViaSdkDefault_NullableObjectCast_TranslatesAndCompiles` — can't find expected `func Convert(...)` signature
- `Issue2519SourceClassConstraintOrderingPipelineTests.Pipeline_ViaSdkDefault_SourceClassConstraintsCompileBeforeTheirDeclarations` — `Assert.Single` on an empty collection
- `Issue2281NonConstantRecordInitializerTests.OahuCliAppShapes_ParameterlessConstruction_CompilesAndPreservesRuntimeDefaults` — translate stage failed
- `Issue2554MarshalFollowUpTranslationTests.ProjectBacked_SystemParametersInfo_RefBool_GainsCanonicalMarshalAs` — `CS2GS0001`: MSBuild workspace load failure (`SourceRoot items must include at least one top-level (not nested) item when DeterministicSourcePaths is true`)
- `Issue2743NullableObjectInitializerWideningPipelineTests.Pipeline_NullableObjectInitializerValues_RemainNullAtRuntime` — translate stage failed
- `Issue2319AppLocalNbgvPackageReferenceTests.Pipeline_AppLocalNbgvFixture_BumpsDeclaredPackageReferenceVersionWithoutDuplication` — `Sequence contains no elements` (`.Single()`)
- `TranslateStageNbgvPackageReferenceTests.TranslateStage_NonCpmProject_LeavesCentralPackageManagementDisabledAndBumpsAppLocalDeclaration` — expected `Passed`, got `Failed`
- `Issue2494EnumLinqExtremaPipelineTests.Pipeline_ViaSdkDefault_SourceEnumMinShapes_TranslateAndCompile` — can't find `"func Pick() Choice"`
- `Issue2517ConstructedBaseOrderingPipelineTests.Pipeline_ViaSdkDefault_EarlyGenericSignatureAndOahuFilterShape_CompileInPathOrder` — `Assert.Single` on an empty collection
- `Issue2538AnonymousConstructionPipelineTests.Pipeline_ViaSdkDefault_AnonymousDelegationAndObjectApi_Compile` — translate stage failed
- `Issue2500NullableExplicitGenericArgumentsPipelineTests.Pipeline_ViaSdkDefault_OahuTaskShape_PreservesNullableTypeArgumentsAndCompiles` — expected `2`, got `0`
- `TestParityStageFalseGreenTests.LibraryParity_BuildFailed_FailsStage_NotPassedOrSkipped` — filter not matched against a live `test/MinimalLibrary` triage artifact (env-dependent fixture data leaking into the assertion)
- `TestParityStageFalseGreenTests.LibraryParity_SdkUnavailable_IsSkipped_NotPassed` — expected `Skipped`, got `Failed`
- `TestParityStageFalseGreenTests.Pipeline_SkippedStage_ReportsSkipped_NotPassed` — expected `"skipped"`, got `"failed"`
- `Issue2646DeclarationPatternPipelineTests.Pipeline_OahuNullableIntDeclarationPattern_Compiles` — nested `CS2GS0001` MSBuild workspace load failure surfacing as the assertion failure message itself (fixture `test/Issue2646`, `DeterministicSourcePaths` project-load error)
- `Issue2726GenericEventPipelineTests.TranslatedGenericEventHandler_LambdaRunsAndIlVerifies` — `Sequence contains no elements` (`.Single()`)
- `Issue2486ImplicitObjectOverridePipelineTests.Pipeline_ViaSdkDefault_ImplicitObjectOverrides_TranslateAndCompile` — can't find `"class Widget"`
- `Issue2490TupleScalarNullabilityPipelineTests.Pipeline_ViaSdkDefault_TupleElementForwarding_TranslatesAndCompiles` — can't find expected tuple-return signature
- `Issue2498NullableLambdaGenericInferencePipelineTests.Pipeline_ViaSdkDefault_NullableLambdaProjection_TranslatesAndCompiles` — can't find `"default(IEnumerator[T]?)"`

## Patterns worth triaging together

1. **A cluster of "assert-contains on translated source" failures return an empty string**
   (`Issue2534`, `Issue2502`, `Issue2492`, `Issue2494`, `Issue2486`, `Issue2490`, `Issue2498`)
   — the translated `.gs` output the test expects to scan appears to come back empty in the
   migrated harness, not just missing the expected snippet. Worth checking whether the
   in-process translation-capture plumbing these tests rely on behaves differently once
   `Cs2Gs.Tests` itself is running as translated G# rather than C#.
2. **`.Single()`/`.Single()`-empty failures** (`Issue2519`, `Issue2517`, `Issue2726`,
   `Issue2319`) suggest a collection the test expects to be populated (diagnostics, triage
   artifacts, package-reference nodes) comes back empty under the migrated harness.
3. **`TestParityStageFalseGreenTests`** (3 failures) is itself a test *about* the test-parity
   harness's stage-classification logic — it may be asserting on `Failed`/`Skipped`
   sentinel values or on live child-process fixture output (`test/MinimalLibrary`,
   `test/Issue2646`) that differs when the outer process is the migrated G# binary
   instead of the original C#.
4. Two failures reference `SourceRoot items must include at least one top-level (not
   nested) item when DeterministicSourcePaths is true` (`Issue2554`, and nested inside
   `Issue2646`'s failure message) — an MSBuild project-load error against a
   sandbox-generated fixture project path; likely a path-shape issue specific to how the
   self-migration sandbox lays out `out/bin/Release/Cs2Gs.Tests/...` fixture directories.

## Where a fix will likely need to look

- `tools/cs2gs/Cs2Gs.Tests/` — the 21 failing test files listed above (fixture/assertion
  source), especially the "assert on translated snippet" and `.Single()` groups.
- `tools/cs2gs/Cs2Gs.Tests/TestParityStageFalseGreenTests.cs` — appears to depend on
  live/child fixture behavior (`test/MinimalLibrary`, `test/Issue2646`) that may behave
  differently once the harness itself runs as migrated G#.
- Possibly the self-migration sandbox project-generation path (`DeterministicSourcePaths`
  / `SourceRoot` MSBuild items) for the two nested-fixture MSBuild-load failures.
- Not a `gsc`/compiler regression by itself — none of these 21 failures moved the app
  below its floor stage; they're pre-existing gaps the ratchet currently tolerates.

## Reference

- Run investigated (per task): https://github.com/DavidObando/gsharp/actions/runs/35102316421 (succeeded; `gate` job https://github.com/DavidObando/gsharp/actions/runs/35102316421/job/104866378757)
- Diagnostic source: shard `validate (2)` (job 104829133967), artifact `cs2gs-selfmig-shard-2`, file `runs/2026-09-16T14-13-04Z_4bf814/tools_cs2gs_Cs2Gs.Tests_Cs2Gs.Tests.csproj-90f34f7c/test-parity-82e39630ca1e.json`
- Related, closed: #4233 / #4253 (separate `Core.Tests` ilverify regression, already fixed — see Thread 1 above)
- Related, open: #3501 (selfmig readability/roadmap tracker — general umbrella, not this specific failure set)

DavidObando:
## Summary

Fixes #4278. Root-caused and fixed all 21 self-hosted `Cs2Gs.Tests` test-parity failures with one change, in one shared function.

### Root cause

`CSharpProjectLoader`'s two `MSBuildWorkspace.Create()` calls left `DeterministicSourcePaths` under ambient MSBuild control. That property turns on whenever `ContinuousIntegrationBuild` is true (which `GITHUB_ACTIONS=true` sets, via `build/gsharp.build.props`) and `Deterministic` is true (the SDK default). Every throwaway fixture `.csproj` this loader opens — via `TranslateStage`/`TestParityStage`, or written directly by dozens of `Cs2Gs.Tests` pipeline tests under their own scratch directories — inherits this repo's root `Directory.Build.props` through MSBuild's ordinary upward directory search, and with it the SourceLink/git-versioning package references `gsharp.build.props` adds project-wide.

On an ordinary CI run those fixtures still sit inside the real git checkout, so `@(SourceRoot)` resolves to one clean top-level root and nothing fails. Under the self-hosted `cs2gs validate` flow, though, the *outer* `Cs2Gs.Tests.dll` itself runs from the self-migration mirror, which deliberately excludes `.git` (`RepositoryFileInventory`). The same design-time project load then fails outright with:

```
CS2GS0001: MSBuild workspace failed to load the project: ... SourceRoot items must
include at least one top-level (not nested) item when DeterministicSourcePaths is true
```

Every one of the 21 reported failures traced back to this one load path, just surfacing through different assertions downstream:
- 7 "assert-contains empty string" failures and 4 ".Single() on empty" failures — the nested `TranslateStage` call that powers dozens of `*PipelineTests` hit the load failure before writing any `.gs`/`.gsproj` output.
- The 3 `TestParityStageFalseGreenTests` failures — `TestParityStage`'s library-parity path loads the sibling `.Tests` project the same way, so a genuine `Skipped`/`Passed` case reported `Failed` instead.
- The 2 `DeterministicSourcePaths`/`SourceRoot` failures (`Issue2554`, `Issue2646`) — these just print the raw MSBuild diagnostic as their own assertion failure; same root cause, no extra layer.
- The remaining one-offs (`Issue2281`, `Issue2743`, `Issue2538`, `Issue2500`, `TranslateStageNbgvPackageReferenceTests`) — all go through the identical nested-project-load path on their own scratch fixtures.

### Fix

`MSBuildWorkspace.Create()` now takes `{"ContinuousIntegrationBuild": "false"}` for every design-time load in this file (both call sites, via a shared `CreateWorkspace()` helper). This loader only ever *reads* a project's syntax and semantics for translation — it never persists a deterministic-PathMap-mapped build output — so forcing the CI flag off here is safe and has no effect on any real build; it only stops an ambient CI setting from reaching a read-only design-time evaluation that never wanted it.

### Reproduction (independent of the self-hosted validate run)

Built a from-scratch mirror of the repo's non-`.cs`/`.csproj`/`.sln` files (matching `RepositoryMirror`'s own skip-list, so no `.git`), dropped `Cs2Gs.Tests`'s own build output into it, and ran it there under `GITHUB_ACTIONS=true`: all 21 named tests failed with exactly the reported messages. After the fix, the same 21 — plus every sibling test in their classes (31 total) — pass.

## Verification

- `dotnet build` (Release, full `GSharp.sln`): 0 warnings, 0 errors.
- Self-hosted reproduction, the real thing: whole-repo `cs2gs migrate` (translate, 56/56 apps green) followed by `cs2gs validate --app tools/cs2gs/Cs2Gs.Tests/Cs2Gs.Tests.csproj` (compile → ilverify → test-parity via `gsc`), run with `GITHUB_ACTIONS=true` to match the nightly's ambient CI environment:
  ```
  app                                         compile       ilverify      test-parity
  tools/cs2gs/Cs2Gs.Tests/Cs2Gs.Tests.csproj  PASS          PASS          PASS
  1/1 apps green; run PASSED.
  ```
  0 of the 21 previously-failing tests fail — full green, not partial.
- Ordinary (non-self-hosted) `dotnet test tools/cs2gs/Cs2Gs.Tests`: **3168/3168 passed**, confirming no regression in the normal C#-compiled run (this loader is shared code both builds compile).
- No release-notes entry: consistent with every other `selfmig:`-prefixed commit (e.g. #4232, #3944, #4267) — `website/docs/release-notes.md` documents user-facing `gsc`/language surface, not the migration tool's own test-loading internals. There is no new diagnostic or language change here.

## Scope note

This closes the issue completely — all 21 failures share the one root cause above, and the self-hosted validate run now reports the app fully green. Nothing is deferred.


DavidObando (commented):
Rubber-ducked this against issue #4278's full 21-failure list, individually. Verdict: the "one root cause, all 21" claim holds. (Self-approval rejected by GitHub since this is my own PR — posting as a comment instead of an approval.)

## What I checked

Traced the mechanism (not just skimmed the diff) for every one of the 5 failure patterns in #4278:

- **Empty-string family (7)** — Issue2534/2502/2492/2494/2486/2490/2498: `TranslateStage.ExecuteAsync` creates `context.ProjectOutputDir` *before* the `WorkspaceLoadFailed` gate, but writes zero `.gs` files if that gate trips. `Directory.GetFiles(..., "*.gs", ...)` then returns `[]`, `string.Join` on an empty array is `""`, and `Assert.Contains("...", "")` fails exactly as reported. Confirmed by reading Issue2534 and Issue2500 (the `Assert.Equal(2, count)` → "expected 2 got 0" variant is the same empty-string mechanism, just a different assertion shape) and grepping the other 5 for the identical `GetFiles(*.gs).Select(ReadAllText)` → `Assert.Contains` shape.
- **`.Single()`-on-empty family (4)** — Issue2519/2319/2517/2726: when TranslateStage fails, `MigrationPipeline` marks every later stage `"skipped"` (never runs `CompileStage`), so no `.gsproj`/`Program.gs` is ever written. `.Single()` on the resulting empty file list throws "Sequence contains no elements" — verified by reading `MigrationPipeline.RunAppAsync`'s short-circuit loop plus Issue2519/Issue2319 source.
- **`TestParityStageFalseGreenTests` (3)** — `RunLibraryParityAsync` calls `TranslateProjectAsync` on the sibling `.Tests` project *before* ever calling the injected fake `GsharpTestProjectRunner`. A `WorkspaceLoadFailed` there returns `StageOutcome.Failed` immediately, so the fake's `Unavailable`/`BuildFailed` result is never consulted — explains all three exact expected/actual mismatches in the issue.
- **Direct `SourceRoot` (2)** — Issue2554/2646: literally the CS2GS0001 message, no indirection.
- **One-offs (5)** — Issue2281 confirmed by reading source (same TranslateStage-gate pattern as above); `TranslateStageNbgvPackageReferenceTests` has *two* facts in one file, and only one (`NonCpmProject`) is in the failing list — I verified why: the CPM fact writes its own `Directory.Build.props` in the fixture tree, which short-circuits MSBuild's upward search before it ever reaches the mirror's real ancestor `Directory.Build.props` (with SourceLink/nbgv), so it never trips `DeterministicSourcePaths`. That's a subtle, correct distinction the PR body doesn't spell out but that holds up under inspection.

## Independent reproduction

Built my own mirror (not the PR's script) by `git ls-files -z | tar ... | tar -x` into a fresh directory — matches `RepositoryFileInventory.Enumerate`'s own skip-list (`.git`/`bin`/`obj`/`TestResults` excluded, tracked `Directory.Build.props`/`build/gsharp.build.props` included) — giving a faithful no-`.git` mirror.

- **Before fix**, filtered run spanning all 6 test classes / 5 patterns, `GITHUB_ACTIONS=true`: **10/16 failed**, every failure message byte-for-byte matching the issue (`Sub-string not found: ""`, `Sequence contains no elements`, `Expected: Skipped / Actual: Failed`, the literal `CS2GS0001 ... SourceRoot items must include...`, `Expected: 2 / Actual: 0`).
- **After fix**, same filter: **16/16 passed**.
- **After fix, full suite in the same hostile mirror** (no `.git`, `GITHUB_ACTIONS=true`): **3168/3168 passed, 0 failures.**
- **Ordinary suite in the real (git-backed) worktree**, fix applied, no env overrides: **3168/3168 passed, 0 failures** — no regression from the change.

Also confirmed: `grep -rn "MSBuildWorkspace" tools/cs2gs` shows only the two call sites in `CSharpProjectLoader.cs`, both now routed through `CreateWorkspace()` — no other unpatched instance exists. And nothing else in the codebase reads the evaluated `ContinuousIntegrationBuild` MSBuild property, so forcing it false has no effect beyond the SourceRoot/`DeterministicSourcePaths` cascade this loader never wanted. `build/gsharp.build.props`'s `GITHUB_ACTIONS` → `ContinuousIntegrationBuild` wiring and unconditional `Deterministic=true` match the PR's description exactly.

Didn't run the full self-hosted `cs2gs migrate`/`cs2gs validate` pipeline (time) — the mirror repro replicates the same mechanism `RepositoryMirror` relies on and is full-suite-clean, so I'm treating that as sufficient rather than load-bearing-required.

No concerns. The "at both call sites, one root cause, all 21" claim is accurate. Recommend merge.

PR #4283, opened 2026-09-17
cs2gs: sound null-conditional-access analyzer mapping, full idiom set (#4173)

### LINKED ISSUE #4173: cs2gs: no analyzer-API mapping for C# null-conditional-access syntax (ConditionalAccessExpressionSyntax / MemberBindingExpressionSyntax) [investigated: needs idiom rewrite, no live consumer, backlogged]
## Update: most of the originally-reported scope is now fixed

This issue originally reported that `cs2gs`'s ADR-0169 analyzer-translation mode had no mapping for C#'s null-conditional-access syntax (`ConditionalAccessExpressionSyntax` / `MemberBindingExpressionSyntax`), discovered while landing GSA0006 (issue #4172).

Investigating further (still as part of #4172's PR) found the gap was much broader than first reported: `IsInsideLoop`/`IsInsideExemptMethod`'s ancestor walks also referenced `ForStatementSyntax`, `WhileStatementSyntax`, `DoStatementSyntax`, `LambdaExpressionSyntax`, `AnonymousMethodExpressionSyntax`, `LocalFunctionStatementSyntax`, and `TypeDeclarationSyntax` — none of which had a mapping either — plus two smaller gaps found once translation got past those (`AssignmentExpressionSyntax.Right` had no member mapping, and `SyntaxKind.SimpleAssignmentExpression`/`SyntaxKind.IdentifierName` had no enum-member renames).

**All of those are now fixed** in `tools/cs2gs/Cs2Gs.Translator/Analyzers/RoslynAnalyzerApiMap.cs` (landed in #4172's PR), following the file's existing conventions — an exact rename where G# has a 1:1 counterpart, an Adapted mapping with an `AdaptationNote` where G#'s shape only partially covers Roslyn's. The `cs2gs-pr-guard` required check (which migrates `InternalAnalyzers.csproj` as part of its hot-core guard set) is confirmed green with these fixes.

## What's still open: `ConditionalAccessExpressionSyntax` / `MemberBindingExpressionSyntax`

This part of the original report is still true and still unfixed. `AssignmentExpressionSyntax.Left` also turned out to have no G# counterpart at all (G#'s assignment targets a single `IdentifierToken`, never an arbitrary expression) — but that one isn't a map-extension candidate the way the others were: no G# type could represent "the LHS as an expression I can type-test", so #4172's PR reworked GSA0006's own LHS/RHS distinction to avoid needing it (test elimination against `.Right`, which does have a real mapping, instead of testing `.Left` directly) rather than teach the translator something structurally false.

The null-conditional case might have a similar answer (rework an analyzer's own detection logic to avoid it) or might genuinely need a new G# `SyntaxKind` member and a decision on what G# syntax construct backs it, since G#'s own `SyntaxKind` enum currently has no member for either C# kind. GSA0006 itself no longer needs this (the `?.` branch was dropped — see #4172's PR, it turned out to be speculative coverage for a shape that never occurred at any real `.BaseClass`-walk site). This issue now just tracks the underlying translator gap for whenever a future `src/Analyzers/InternalAnalyzers` analyzer needs `x?.y` detection.

DavidObando:
Fixes #4173

## Background

An earlier version of this PR implemented two sound idioms (a type-test rewrite and a registration-guard wrapper) but had to remove a third idiom after review found it unsound — it tried to syntactically "prove" single-hop-vs-chain-inner-node from analyzer source text alone, which is fundamentally impossible (both cases satisfy the same syntactic test). This PR completes the fix with a sound version of the rest of the idiom set, driven by a deeper structural investigation, plus fixes a related sibling bug found along the way.

## The core correspondence (and a correction to how it was first understood)

Roslyn's `a?.b?.c?.d` nests via `WhenNotNull`: the outer `ConditionalAccessExpressionSyntax` (`CAE1`) wraps `CAE2` in `WhenNotNull`, which wraps `CAE3`, etc. — each `CAE_k` is the k-th `?.`/`?[` operator, counting left to right, and `CAE1` (outermost) is the FIRST operator.

An earlier design pass assumed G#'s own postfix chain nests the mirror-image way — via `AccessorExpressionSyntax.LeftPart`, prefix/left-growing — reading the parser's `current = new AccessorExpressionSyntax(current, dotToken, rightSide)` pattern at face value. **This turned out to be wrong**, discovered by empirically parsing G# source and inspecting the resulting tree (as the implementation instructions explicitly required before trusting it): because parsing the right-hand side of a dot recurses back into the full postfix grammar (`ParseMemberContinuation` → `ParseNameOrCallExpression` → its own `ParsePostfixChain`) *before* the enclosing loop iteration finishes, that recursive call already swallows every following `.`/`?.`/`[`/`(`. The actual result is that G#'s dotted continuations right-nest too — through `AccessorExpressionSyntax.RightPart` — structurally the same shape as Roslyn's own `WhenNotNull` chain, not a mirror image. `[`/`?[`/`(` continuations still nest the ordinary iterative way (`IndexExpressionSyntax.Target` / `CallExpressionSyntax.Callee`), so the overall shape is hybrid, not uniform. `AssignmentTargetSyntaxFacts.TryLiftTrailingMemberAccess`'s existing (working, independently-authored) recursion over `accessor.RightPart` is a second, independent confirmation of this same right-nesting.

Because of the hybrid nesting, "does node `n` have a further hop wrapping it" is not one simple parent check — a later `.`/`?.` hop is a *descendant* of an earlier one (via `RightPart`), not an ancestor, while a later `[`/`(` hop *is* an ancestor. `NullConditionalChain`'s helpers are built on two primitives that stay correct regardless of which way a given link nests: `ChainTop` (walk up through every kind of chain link to the outermost node) and a left-to-right token-position scan of the hops underneath that top (rather than a literal parent/child walk) for "next hop" queries. Every span/predicate formula was cross-checked empirically against real Roslyn parses of the same shapes before being trusted (see `HopSpan`/`ReceiverSpan`/`WhenNotNullSpan`'s doc comments in `src/Core/CodeAnalysis/Syntax/NullConditionalChain.cs`).

## The fix

**New file**: `src/Core/CodeAnalysis/Syntax/NullConditionalChain.cs` — a `src/Core` utility (not a cs2gs-only synth helper), pure functions over real G# `SyntaxNode`s, usable by any G# analyzer (hand-written or translated) walking a null-conditional access chain at analysis time. Exposes `IsNullConditionalHop`/`AsNullConditionalHop`, `NullConditionalReceiver`, `IsChainRoot`/`ChainTop`, `NextNullConditionalHop`, `TailIsMemberBinding`/`TailIsElementBinding`, `ReceiverIsMemberBinding`/`ReceiverIsElementBinding`/`ReceiverIsPlainMemberAccess`, and Roslyn-exact `HopSpan`/`ReceiverSpan`/`WhenNotNullSpan`/`NullConditionalOperatorToken`.

**cs2gs rewrites** (`CSharpToGSharpTranslator.Analyzers.cs` / `.Patterns.cs`), all gated on `InAnalyzerApiMode`, and — critically — wired through **one shared entry point** (`TryTranslateAnalyzerNullConditionalTypeTest`) reached from **both** C# shapes a type test can take: a designated `is Type x` / pattern-combinator form (`IsPatternExpressionSyntax`, translated via `TranslateIsPattern`), and a **bare** `expr is Type` with no designator — which parses as the older `BinaryExpressionSyntax`/`SyntaxKind.IsExpression` node and never reaches pattern translation at all. Missing the bare form (the commonest real-world spelling of a plain type probe) would have made the new non-discriminating map row below silently vacuous for it.

- **I1** (replaces the type-test rewrite from the earlier idiom): `expr is ConditionalAccessExpressionSyntax [x]` → `NullConditionalChain.AsNullConditionalHop(expr) is ExpressionSyntax [x]` — covers `?.` and `?[` uniformly via one type pattern (closing a real `a?[i]` under-fire the earlier idiom left in place). Refuses when the scrutinee is itself `<x>.Expression`/`<x>.WhenNotNull` on a CAE-typed `x`, routing to I4/I5 instead.
- **I2** (extends the registration-guard rewrite): `RegisterSyntaxNodeAction(handler, SyntaxKind.ConditionalAccessExpression)` → a guarded wrapper registered for **both** `SyntaxKind.AccessorExpression` and `SyntaxKind.IndexExpression` (was accessor-only — a genuine `a?[i]` under-fire, not a disclosed gap). Guard is `NullConditionalChain.IsNullConditionalHop(ctx.Node)`.
- **I3**: `<x>.Expression` (x CAE-typed) → `NullConditionalChain.NullConditionalReceiver(x)` — same value as Roslyn's receiverless `.Expression` at every hop, but only extent/lifted-type-exact at the chain's first hop (documented, disclosed residual).
- **I3a**: `.GetLocation()`/`.Span` on a CAE-typed node, or on `<x>.Expression`, → Roslyn-exact `NullConditionalChain.HopSpan`/`ReceiverSpan`.
- **I4**: `<x>.WhenNotNull is <Type> [designator]` → `TailIsMemberBinding`/`TailIsElementBinding`/`NextNullConditionalHop(x) is ExpressionSyntax [designator]` depending on `<Type>`. A `MemberBindingExpressionSyntax`/`ElementBindingExpressionSyntax` designator is narrowed with `(x as AccessorExpressionSyntax)!!`/`(x as IndexExpressionSyntax)!!` (the same non-smart-castable narrowing convention `BuildPatternNarrowingReplacement` already uses elsewhere), since the predicate proving it safe doesn't itself narrow G#'s static type the way a native pattern test would.
- **I5**: `<x>.Expression is <Type> [designator]` → `ReceiverIsMemberBinding`/`ReceiverIsElementBinding`/`ReceiverIsPlainMemberAccess`, with the same narrowed-designator treatment. `<Type>` = `ConditionalAccessExpressionSyntax` is a loud `CS2GS-GAP` (Roslyn only reaches a genuine CAE there through a parenthesized chain break, which can't be told apart from an ordinary hop without risking a vacuously-true test — the same class of reasoning that sank the earlier idiom).
- **I6**: `MemberBindingExpressionSyntax.Name`/`.OperatorToken` → `RightPart`/`DotToken`; `ElementBindingExpressionSyntax.ArgumentList` → `Indices` (plus a small procedural rule dropping the `.Arguments` wrapper the same way `InvocationExpressionSyntax.ArgumentList.Arguments` already does).
- **I7**: one `TypeMap` row, `ConditionalAccessExpressionSyntax` → G#'s non-discriminating `ExpressionSyntax` supertype — safe because every discriminating use (`is`/`as`, `OfType<>()`, `FirstAncestorOrSelf<>()`, this row's own bare-cast fallback) is intercepted first by I1/I4/I5/I8; a bare cast trusts the I2 registration guard for soundness rather than trying to statically prove a handler is exclusively guard-registered.
- **I8**: `.OfType<ConditionalAccessExpressionSyntax>()` → `.OfType<ExpressionSyntax>().Where(NullConditionalChain.IsNullConditionalHop)`; `.FirstAncestorOrSelf<ConditionalAccessExpressionSyntax>()` → an explicit self-or-ancestors scan filtered the same way.
- **I9**: everything else stays a loud `CS2GS-GAP`, unchanged philosophy.

**Sibling bug (§6, found independently, same class as the `a?[i]` under-fires above)**: `MemberAccessExpressionSyntax` → `AccessorExpressionSyntax` and `SyntaxKind.SimpleMemberAccessExpression` → `SyntaxKind.AccessorExpression` were **both unconditional** — but Roslyn's ordinary `MemberAccessExpressionSyntax` *excludes* a `?.` hop's tail (that's Roslyn's own receiverless `MemberBindingExpressionSyntax`), so `node is MemberAccessExpressionSyntax` and an unguarded `SimpleMemberAccessExpression` registration both silently over-fired on every `?.` hop's tail too. Fixed with the exact inverse of I1/I2 (`IsNullConditional: false`), same code shape, one boolean flipped — for both the designated and bare `is` forms.

**A third, related pre-existing gap surfaced by I7**: before this PR, a *bare* (undesignated) `x is ConditionalAccessExpressionSyntax`/`x is MemberAccessExpressionSyntax` never reached the analyzer-mode type-test rewrite at all (it only ever handled the designated `is Type x` form) — harmless before I7 landed, since the fallback was an honest `CS2GS-GAP` for the CAE case. Once I7 adds a real (non-discriminating) map row, that same bare form would have silently gone through it unguarded. Fixed by wiring the shared entry point through the `BinaryExpressionSyntax`/`IsExpression` path too (see "one shared entry point" above), with dedicated regression tests (`BareCastHandlerTypeTest_...`, `BareReceiverConditionalAccessTest_...`).

**Out of scope, disclosed**: G#'s null-conditional *indirect* invocation `f?(x)` (a G#-only construct, no C# counterpart) does not dispatch through any of this. The `ElementAccessExpressionSyntax`/`SyntaxKind.ElementAccessExpression` analog of the §6 sibling bug (an ordinary `a[i]` type test/registration would likely over-match `a?[i]` the identical way) was noticed while implementing this fix but is **not** fixed here — flagging it for a follow-up rather than expanding this PR's already-large scope further.

## Testing

`tools/cs2gs/Cs2Gs.Tests/Issue4173NullConditionalAnalyzerMappingTests.cs`, 18 test methods (several data-driven with multiple cases), covering every idiom above plus the critical semantic checks:

- **Real Roslyn parity** (translate the analyzer, compile+run it with the real G# compiler over G# source; separately compile+run the analyzer's own C# source with Roslyn over equivalent C# source; compare): `TailMemberBindingTest_MatchesRoslyn_OnChainAndMixedInputs` (the tail-test idiom fires exactly once on `a?.b?.c`, not twice — the earlier removed idiom's exact over-fire repro — and exactly zero times on `a?.b.c`, its second counterexample), `RecursiveChainWalkTest_MatchesRoslyn_OnThreeLevelChain`, `InnerHopReceiverTest_MatchesRoslyn_OnChain`, `OrdinaryMemberAccessTest_MatchesRoslyn_OnMixedChain` (the sibling-bug fix: `a?.b.c` fires exactly once matching Roslyn, not the pre-fix over-fire).
- **Positive/negative parity via `GSharpAnalyzerVerifier`**, extended with `a?[i]` (the I2 index-registration fix) and a negative ordinary-`[i]` companion.
- **Bare-form regression tests** for the third gap above.
- **Gap-preservation tests**: `.WhenNotNull` as a bare value, `.Expression is ConditionalAccessExpressionSyntax`, and multi-kind registration all still surface a loud `CS2GS-GAP`.
- Compile+bind+behavior checks for I3/I3a (receiver span), I6 (name extraction), I7 (bare cast, with the false-negative-proof of firing on ordinary access removed), I8 (descendant walk finds exactly the chain's hops, not every access).

**RED/GREEN verification**: `git worktree add` at this PR's pre-this-commit tip (`d688294c`) with the new/updated test file copied in. 11 of 18 fail there, including the four explicitly-targeted critical rows — the tail-test parity test fails with a translate-time `CS2GS-GAP` (couldn't even compile before), and notably `OrdinaryMemberAccessTest_MatchesRoslyn_OnMixedChain` fails with `Expected: 1, Actual: 2` — a genuine, reproduced-live over-fire from the pre-fix sibling bug. The 7 that pass unchanged are the ones whose behavior legitimately didn't change (pre-existing loud gaps, and the already-correct single-hop/chain cases from the earlier idiom). All 18 pass on this branch.

**Regression suites** (all green): `dotnet build` on `GSharp.sln`, Release, 0 warnings/0 errors. `Adr0169`/`Analyzer`/`Issue3920`/`Issue3686`/`Issue3794`/`Issue4190` filter (118 tests) and `~Pattern` filter (207 tests, since this change touches the shared `TranslateIsPattern`/`TranslateBinaryExpression` entry points used by all translation, not just analyzer mode).

## Residual, disclosed gaps

- `.Expression`/`NullConditionalReceiver` is value-equivalent but not extent/lifted-type-exact at an inner hop (documented in I3's shape warning and `NullConditionalChain.NullConditionalReceiver`'s doc comment).
- `.WhenNotNull` read as a bare value (not the scrutinee of an `is` test) has no G# counterpart — loud gap.
- `.OperatorToken` on a binding designator maps to G#'s `?.`/`?[` token, wider than Roslyn's bare `.`/`[` — noted in the `RoslynAnalyzerApiMap` row's `AdaptationNote`.
- `f?(x)` (G#'s own null-conditional indirect invocation) is a G#-only construct with no C# counterpart and is out of scope.
- The `ElementAccessExpressionSyntax`/`SyntaxKind.ElementAccessExpression` analog of the §6 sibling bug is a known, disclosed, **not-fixed** gap (see above).

## Round 3 update

A round-2 adversarial review found the type-test fix above (I1/I1b/I4/I5, reached through `TryTranslateAnalyzerNullConditionalTypeTest`) only recognized the **top-level** shape of a pattern — a bare/designated `DeclarationPatternSyntax`/`TypePatternSyntax`/`ConstantPatternSyntax`, matched by that one dedicated entry point. Any pattern wrapped in `not`, `and`, `or`, a switch-case label, or a nested property subpattern fell through to the ordinary native-pattern/boolean-lowering translation, which had no CAE/shared-node awareness at all — silently over-matching.

**Reviewer's exact repro**: a designated `is not ConditionalAccessExpressionSyntax cae` registered on `SyntaxKind.IdentifierName`, run over **plain** (non-null-conditional) `box.Name`, reported 2 diagnostics where real Roslyn reports 0.

### The fix (this round)

Fixed at the **leaf**, not by enumerating more pattern shapes (the same enumeration mistake round 1 and round 2 both made in different spots). Every pattern-position type-mapping call site — `TranslatePattern`, `TranslateRecursivePattern`, `BuildNativePattern`, `TranslatePatternTest`, `TranslateNotPatternTest`, `TranslateRecursivePatternTest`, the loop-condition guard emitter (`EmitMustHoldGuards`), and the statement-level `if let`/guard-hoist family (`TryBuildPositiveGuardHoist`, `TryBuildNegatedGuardHoist`, `TryBuildMultipleNegatedGuardHoists`, `TryTranslateIfLetConditional`, `TryBuildIfLetGuard`, `TryBuildWhileLetLoop`, `TryTranslateIfLetBooleanExpression`) — now routes through two shared builders (`MapPatternTypeSyntax`/`BuildPatternTypeTest` for pattern positions, `BuildTypeTestExpression` for boolean positions) instead of calling the general type mapper (`MapTypeSyntax`/`MapTypeReferenceExpression`) directly.

- **`MemberAccessExpressionSyntax`/`ElementAccessExpressionSyntax`** (the latter a previously-unreported sibling of the identical bug — the map row carried no guard of any kind, not even the shape check I1b added for the designated/bare top-level forms): the fix is a pure G# **pattern** leaf swap (`AccessorExpressionSyntax`/`IndexExpressionSyntax` plus an `IsNullConditional: false` suffix), which composes for free under `not`/`and`/`or`/parens/switch/nested subpatterns through the existing, already-tested pattern walkers — no new shape enumeration anywhere.
- **`ConditionalAccessExpressionSyntax`**, whose faithful G# form is the `NullConditionalChain.AsNullConditionalHop(x) != nil` **predicate** (not a pattern — there is no G# node type it can bind to): a shape-agnostic `DescendantNodesAndSelf` scan (`PatternMentionsConditionalAccessType`) — deliberately not a switch over pattern kinds, so a future shape can't defeat it the way both earlier rounds' enumerations did — detects any CAE-mentioning `is`-pattern and routes the **whole** pattern off the native-G#-pattern paths onto the existing boolean-lowering path, where the same leaf hook emits the predicate and `not`/`and`/`or` compose for free. A CAE designator binds by substitution (re-evaluating the receiver at each use, the same convention I4/I5 already use) at the only three positions C#/G# both allow one — top level, under top-level `not`, or under `and` (GS0390/CS8780 forbid it under `or` or nested `not`).
- A position that is G#-pattern-**only** (a switch-case label, a switch-expression arm, or a nested property subpattern that could not be boolean-lowered) has no sound CAE rewrite and now surfaces a loud `CS2GS-GAP` instead of silently falling back to I7's non-discriminating `ExpressionSyntax` supertype.
- The statement-level `if let`/guard-hoist machinery builds its own hoisted local **ahead of** the pattern-test translation ever running, so it needed its own fix, not just a caller of the shared leaf. In `ControlFlow.cs`'s three guard-hoist builders, the hoist is adapted in place: for CAE the initializer becomes `NullConditionalChain.AsNullConditionalHop(receiver)` directly (already nullable, so it drops into the existing `let t T? = …; if t != nil` shape unchanged), and for a plain-access type the existing `as`-cast is kept but the guard gains an extra `IsNullConditional` conjunct. In `IfLet.cs`'s four `if let`/`while let` builders, adapting the hoist inline was not safe to do without risking the same class of bug in a more scope-sensitive construct, so those decline for a shared-node type and fall back to the general lowering instead (verified — see `PositiveDesignatedConditionalAccess_MatchesRoslyn_OnBothShapes` below — the fallback correctly reaches the same fixed leaf).
- Two real, in-repo sites in `src/Analyzers/InternalAnalyzers/RewriterClonePreservationAnalyzer.cs` (a nested property-subpattern designator at line ~300 and a switch-case label at line ~334, both testing `MemberAccessExpressionSyntax`) go from silently over-matching to the faithful discriminated rewrite — an intentional, expected change in the self-hosted migration's own translated output, verified via `Adr0169AnalyzerTranslationTests.RealAnalyzerSources_TranslateWithReviewWarningsOnly`, not a regression. No selfmig baseline needed refreshing for this: `tools/cs2gs/selfmig-baseline.json`/`selfmig-test-allowlist.json` track whole-repo ratcheting code-quality metrics (green-app floor, long-line/`!!`/label/lift ceilings), not per-analyzer diagnostic text, so this change isn't at a granularity selfmig observes.
- `NullConditionalChain.HopSpan`/`ReceiverSpan` now throw `ArgumentException` on a non-hop node instead of silently returning a plausible-but-meaningless span (precondition hardening; the sole existing emitter only ever calls them on an already-proven hop, so this is not a behavior change there).
- `RoslynAnalyzerApiMap.Entry` gained a `SharedGsNode` flag, set on exactly the three shared-node rows; a new test (`SharedGsNodeRegistry_AgreesWithRoslynAnalyzerApiMap`) asserts this set agrees exactly with the translator's own shared-node registry, so a future fourth shared-node row added to one but not the other fails loudly instead of shipping this bug class a fourth time.

### Explicitly out of scope this round: `as`-casts (§4.4)

An `as ConditionalAccessExpressionSyntax`/`as MemberAccessExpressionSyntax`/`as ElementAccessExpressionSyntax` cast is **left exactly as it behaved before this round** — still the warn-and-pass `CS2GS-ANALYZER-SHAPE` treatment (I7's non-discriminating supertype for CAE; the unconditional map row for the other two), not the discriminating rewrite. This is a deliberate scope cut, not an oversight: narrowing an `as`-cast the same way would touch a previously-approved trust decision (I7's own doc comment: "trusting the I2 registration guard for soundness, deliberately NOT trying to statically prove a handler is exclusively guard-registered") and needs repo-owner sign-off before changing. Tracked as a known follow-up, not fixed here.

### Testing (round 3 additions)

New tests in `tools/cs2gs/Cs2Gs.Tests/Issue4173NullConditionalAnalyzerMappingTests.cs`: the reviewer's exact counterexample (both negated and positive designated forms, `ReviewerCounterexample_NegatedDesignatedConditionalAccess_MatchesRoslyn_OnPlainAccess` / `PositiveDesignatedConditionalAccess_MatchesRoslyn_OnBothShapes`), `or`/`and` combinator composition (`OrCombinatorConditionalAccess_TranslatesBothLeavesSoundly`, `AndCombinatorConditionalAccess_MatchesRoslyn_OnMixedAccess`), switch-label and switch-expression-arm gap tests (`SwitchLabelConditionalAccess_IsALoudGap`, `SwitchExpressionArmConditionalAccess_IsALoudGap`), the critical nested-subpattern-with-null-`receiverSyntax` shape (`NestedSubpatternConditionalAccess_MatchesRoslyn_WithNullReceiverSyntax`, via the one real Roslyn shape that puts a bare CAE inside another node's typed property — a parenthesized chain break, `(a?.Next)?.Value`), the previously-unreported `ElementAccessExpressionSyntax` switch-label sibling (`SwitchLabelElementAccess_MatchesRoslyn_OnMixedAccess`), a base-call-check ordering regression (`BaseCallCheck_StillTranslatesToBaseClassCallExpressionSyntax`), the §4.5 map/registry drift test, and `NullConditionalChainTests.HopSpanAndReceiverSpan_ThrowOnNonHopNode`.

**RED/GREEN verification**: `git worktree add` at this round's pre-change tip (`b5ecff0a`), new/updated test files copied in (the drift test commented out there — it references translator-internal registries that don't exist pre-round-3, so leaving it live would hide RED for the rest of the file behind a compile error). 7 of the round-3-added `Issue4173...` tests fail there for the right reasons (including the reviewer counterexample and the nested-subpattern test), plus `NullConditionalChainTests.HopSpanAndReceiverSpan_ThrowOnNonHopNode` fails (no exception thrown, since the guard doesn't exist yet). All pass on this branch.

**Regression suites** (all green on this branch): `dotnet build` on `GSharp.sln`, Release, 0 warnings/0 errors. `Adr0169`/`Analyzer`/`Issue3920`/`Issue3686`/`Issue3794`/`Issue4190` filter (128 tests, up from 118), `~Pattern` filter (208 tests), `Core.Tests` `CodeAnalysis.Syntax` filter (1148 tests), and the full `Issue4173NullConditionalAnalyzerMappingTests` file (28 tests).


DavidObando (commented):
Rubber-duck review against issue #4173's scope, including the prior investigation's technical comment.

**Verdict: found and fixed a genuine soundness defect in the PR's third idiom during review. Recommend owner review of the resulting scope reduction before merge — commenting rather than approving, since I substantially changed the PR's own diff.**

## What I verified as correct (idioms #1 and #2)

- The `is ConditionalAccessExpressionSyntax x` → `is AccessorExpressionSyntax { IsNullConditional: true } x` type-test rewrite. Verified the chain-parsing claim directly by reading `Parser.ParsePostfixChainCore` (the postfix loop wraps the already-built `current` into a new `AccessorExpressionSyntax` on each `?.`/`.`) and by parsing `box?.Next?.Leaf` and inspecting the tree: two nested `AccessorExpressionSyntax` nodes, both `IsNullConditional=true`. Chain-safe as claimed.
- The `RegisterSyntaxNodeAction` wrapper-lambda guard. Verified it forwards the same `ctx` to the original handler (printed output shows `Analyze(ctx)`), and that a registration combining `ConditionalAccessExpression` with another `SyntaxKind` in the same call correctly produces a loud `CS2GS-GAP` instead of a silently-wrong guard (confirmed via the PR's own `MultiKindRegistration_*` test and an independent scratch check).
- `MemberBindingExpressionSyntax`/`ElementBindingExpressionSyntax` stay unmapped, no map-row was added to `RoslynAnalyzerApiMap.cs`.
- `a?[i]` (`IndexExpressionSyntax`) — confirmed the `CS2GS-ANALYZER-SHAPE` warning fires for a real `WhenNotNull is ElementBindingExpressionSyntax`-testing analyzer, and that `ElementBindingExpressionSyntax` itself still surfaces a loud `CS2GS-GAP` rather than a silent miscompile. This matches the file's existing convention (the `TypeDeclarationSyntax` → `StructDeclarationSyntax` carve-out) — a reasonable, honestly-disclosed scope boundary.

## What I found wrong (idiom #3, now removed)

The PR's third idiom rewrote the conjunction `expr is ConditionalAccessExpressionSyntax x && x.WhenNotNull is MemberBindingExpressionSyntax` to just the left-hand type test, treating the `WhenNotNull` check as proof of single-hop-ness and rewriting a later `x.Expression` to `x.LeftPart`.

That proof does not hold. Roslyn's tree for a two-level chain `a?.b?.c` contains an **inner** `ConditionalAccessExpressionSyntax` node (`Expression: MemberBindingExpressionSyntax(.b)`, `WhenNotNull: MemberBindingExpressionSyntax(.c)`) whose own `WhenNotNull` **is** a bare `MemberBindingExpressionSyntax` — confirmed by parsing `a?.b?.c` directly with Roslyn and inspecting `WhenNotNull.Kind()` on each `ConditionalAccessExpressionSyntax` descendant. So the conjunction is true on that inner chain node too, and G#'s per-node `IsNullConditional` flag can't distinguish it from a genuine single-hop node.

Confirmed end-to-end (translate, compile, run against real G# source, compare to Roslyn's own count on equivalent C#):
- `box?.Next?.Leaf`: Roslyn fires **1**; the naively-lowered translated analyzer fired **2**.
- `box?.Next.Leaf`: Roslyn fires **0**; the naively-lowered translated analyzer fired **1**.

This is a silent over-fire — exactly the failure class the issue's own prior investigation warned a naive map-row extension would cause, reintroduced one level down by this idiom.

## Fix applied and pushed

Removed `TryLowerConditionalAccessSingleHopConjunction`, the `.Expression`→`.LeftPart` rewrite it gated, and `ProvenSingleHopConditionalAccessLocals`. `.WhenNotNull`/`.Expression` now stay a loud `CS2GS-GAP` in every case — single-hop-proof conjunction or not — which is the same safe fallback the prior investigation already prescribed for the plain chain case. Updated the affected test (now `SingleHopConjunction_IsALoudGapRatherThanASilentOverfire`, asserting the loud gap instead of the removed rewrite) and the release notes entry to match. Pushed as a new commit to this branch.

## Independent verification performed

- **Red/green**: built the branch, ran all 8 tests in `Issue4173NullConditionalAnalyzerMappingTests.cs` — green. Reproduced RED independently via `git worktree add` at the merge-base commit with the new test file copied in: 7/8 fail (6 as assertion failures, 1 as a translate-time unmapped-type gap that blocks compilation). The chain-shaped companion test passes unchanged at merge-base, at the original PR head, and after my fix — confirmed it really is an invariant unaffected by this work.
- Analyzer-translation regression suite (`Adr0169`/`Analyzer`/`Issue3920`/`Issue3686`/`Issue3794`/`Issue4190`, 108 tests): green, both before and after my fix.
- Pattern/binary-expression-adjacent suite (`~Pattern` filter, 206 tests — the shared `TranslateIsPattern`/`TranslateBinaryExpression` entry points this change touches): green, confirming ordinary non-analyzer `?.` translation is unaffected.
- Full `dotnet build GSharp.sln -c Release`: 0 warnings / 0 errors, both before and after my fix.
- Full `Cs2Gs.Tests` suite: [see thread — running / result below].

Full transcript of the adversarial reproduction (Roslyn ground truth + end-to-end translated-analyzer counts) is available on request; happy to also leave a standalone regression test for the specific 2-vs-1 / 1-vs-0 chain counts if that's useful beyond the loud-gap test already added.

DavidObando:
Follow-up: the full unfiltered `Cs2Gs.Tests` suite did not complete cleanly during review (two attempts, both invalidated by my own concurrent rebuilds of their output directories while investigating). Not counting it either way. The regression evidence is the targeted suites: 108 analyzer-translation tests, 206 pattern/binary-expression tests (both green), plus the fact that every line removed by the fix was gated behind `InAnalyzerApiMode` — non-analyzer `?.` translation is untouched by construction, not just by test coverage. PR body updated to match.

Issue #4167, opened 2026-09-09
cs2gs: RequiresEnumStatementFallback hits the #4153 pattern-match gap under self-hosting — NullReferenceException


DavidObando:
## What this is

A confirmed instance of the gsc pattern-match gap tracked by #4153, at one of that issue's own listed candidate sites. Small, mechanical, same fix shape as #4155 already applied at four other call sites.

## Where

`tools/cs2gs/Cs2Gs.Translator/CSharpToGSharpTranslator.Types.cs:1137`, `RequiresEnumStatementFallback`:

```csharp
private bool RequiresEnumStatementFallback(SwitchStatementSyntax node)
{
    if (this.context.GetTypeInfo(node.Expression).Type is not INamedTypeSymbol { TypeKind: TypeKind.Enum } enumType)
    {
        return false;
    }
    ...
```

## The break

Gate run [34370082359](https://github.com/DavidObando/gsharp/actions/runs/34370082359) — `tools/cs2gs/Cs2Gs.Tests` self-migration now throws a fresh `NullReferenceException` at this exact call site, in the migrated `CSharpToGSharpTranslator.Types.gs:1138`:

```
System.NullReferenceException : Object reference not set to an instance of an object.
   at Cs2Gs.Translator.CSharpToGSharpTranslator.DeclarationVisitor.RequiresEnumStatementFallback(SwitchStatementSyntax node)
      in .../Cs2Gs.Translator/CSharpToGSharpTranslator.Types.gs:line 1138
   at ...DeclarationVisitor.TranslateSwitchStatement(...)
```

Same shape as every other confirmed #4153 site: an enum-constant sub-pattern (`TypeKind: TypeKind.Enum`) nested in a type test against `ITypeSymbol` (an imported interface, from `GetTypeInfo().Type`). Under self-hosting, the pattern silently mismatches — this method takes the wrong branch, and something downstream that assumed the `enumType` binding null-refs.

Newly surfaced by #4155 landing: that PR's own regression test (`Issue4045SelfHostedPatternMatchRegressionTests`) self-migrates `CSharpTypeMapper.cs`/`Constructors.cs`, which pulls this file's `TranslateSwitchStatement` into the compiled closure and exercises it — so #4155 didn't cause this, it made it visible.

## The fix

Decompose, exactly as #4155 did at its four sites:

```csharp
if (this.context.GetTypeInfo(node.Expression).Type is not INamedTypeSymbol enumType
[...]

PR #4308, opened 2026-09-18
Fix #4287: null-check CLR/imported instance method calls on nullable receivers

### LINKED ISSUE #4287: gsc: imported/CLR instance method calls never null-check a nullable receiver
## Summary

A call to an imported/CLR instance method through a nullable receiver (e.g. `string?`) is never null-checked by gsc — it compiles with zero diagnostics and throws an unattributed `NullReferenceException` at runtime if the receiver is actually nil. This is unconditional: it reproduces with **no narrowing, no `goto`, and no control flow at all** — a plain, direct call is enough.

By contrast, a call to a **user-defined** instance method through a nullable receiver is correctly rejected at compile time with `GS0159` ("Cannot call function ... because receiver ... may be nil").

## Reproduction

```gs
package Demo

class Holder {
    let name string?
    init(n string?) { this.name = n }
    func Use() string {
        return this.name.ToUpper()
    }
}

let h = Holder(nil)
Console.WriteLine(h.Use())
```

Compiled and run with the in-tree `gsc`: compiles with **zero diagnostics**, then throws:
```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Demo.Holder.Use()
```

For contrast, the user-defined-method equivalent is correctly rejected:
```gs
package Demo
class Animal { func Speak() string { return "hi" } }
class Holder { let pet Animal? }
func Use(h Holder) string { return h.pet.Speak() }
```
```
error GS0159: Cannot call function Speak because receiver 'h.pet' may be nil. Use '?.' for a null-safe call or bind it with 'if let'.
```

## Where this likely lives

`src/Core/CodeAnalysis/Binding/ExpressionBinder.Calls.Invocation.cs` (~line 3620-3640): the `GS0159`/`ReportUnableToFindFunction` "may be nil" diagnostic is only reached as a *fallback* when ordinary overload resolution against the nullable receiver type fails and `IsApplicableNullableUnderlyingCall` finds a matching member on the underlying (non-nullable) type. For a CLR/imported instance method, member lookup against the nullable receiver apparently succeeds directly (treating `T?` as equivalent to `T` for CLR member-lookup purposes) without ever falling into this fallback path, so the nil-check is skipped entirely. This needs further investigation to confirm the exact binding path (e.g. `ExpressionBinder.Access.MemberLookup.cs`, wherever an imported/CLR instance member is resolved against a `NullableTypeSymbol` receiver) and to design a fix that reports the same class of diagnostic for this path.

## Discovered via

Found while reviewing PR #4286 (fix for #4285, "smart-cast narrowing survives a `goto` that jumps past the nil-guard"). #4285's own filed repro used exactly this shape (`this.name string?` / `this.name.ToUpper()`); a Copilot review of #4286 correctly flagged that the fix does not make the *exact* filed repro pass, because the repro's crash is caused by two independent, overlapping issues — #4285's narrowing-lift defect (now fixed) and this one (still open). Verified directly that this issue reproduces identically with **zero** `goto`/narrowing involvement, so it is unrelated to #4285's mechanism and out of scope for that PR.

## Impact

This is a broader gap than #4285's — every nullable-receiver call to an imported/CLR instance method (not just narrowed ones, not just goto-adjacent ones) skips the null check entirely. Worth prioritizing independently.

DavidObando:
Fixes #4287

## Summary

- **The first revision of this fix was carve-out-INCONSISTENT with the read path, and that single asymmetry is what reddened CI.** `ExpressionBinder.Access.MemberLookup.cs`'s `CanBindClrInstanceMember` has a deliberate, long-standing carve-out: an imported member whose oblivious metadata made it `T?` (issue #1354) still counts as a valid INTERMEDIATE receiver in a member chain. The new call-path check had no equivalent, so a receiver that was fine for a chained READ became invalid the moment the next step was a CALL — `Environment.Version.Major` compiled while `Environment.Version.ToString()` reported `GS0159`.
- **Fix: share one predicate.** `CanBindClrInstanceMember` now delegates its nullable-receiver decision to a new `IsImportedClrChainReceiver`, and `BindAccessorCall` consults that same `CanBindClrInstanceMember` — reused, not duplicated, so read and call cannot drift apart again. The predicate asks one question: did this receiver's nullability come from imported CLR metadata, or from an explicit G# `?`.
- **Widened to call results, on BOTH paths.** Alongside the imported property and field reads the predicate already admitted (a CLR field read binds to a `BoundClrPropertyAccessExpression` too, so fields needed no separate case), it now admits imported call results — `BoundImportedInstanceCallExpression`, `BoundImportedCallExpression`, `BoundClrStaticCallExpression`. The read path had never covered those either; this widens both, not the call path alone.
- **What stays `GS0159` — what the fix actually buys.** A DIRECT receiver whose `?` is an explicit G# annotation: a `string?` field (`this.name.ToUpper()`, the filed repro), a `string?` local or parameter, a G# function's nullable return (`Maybe().ToUpper()`), and a tuple element (`t.Item2.ToString()`).
- **Copilot review finding #2 (real, fixed).** An imported interface is never an `InterfaceSymbol`, so `IsApplicableNullableUnderlyingCall`'s final arm could not see `System.Object`'s members on one and `IEnumerable[int32]?.ToString()` bound unguarded. The probe now mirrors the ordinary path's imported-interface arm (`clrType is { IsInterface: true }`, issue #2304).
- **Copilot review finding #1 (not a bug).** `Issue2193ExtensionOverloadShadowTests`' receiver is a non-nullable `string`, so the nullable branch never runs and its member-beats-extension precedence is untouched. For a receiver that is still nilable the CLR member is BY DESIGN not selectable — that is the fix — so a `string?`-declared extension is the only viable candidate; the alternative is `GS0159`, which would break the `func (s string?) OrEmpty()` shape this PR's own test protects. Pinned by the new `ExtensionOnNullableType_ShadowingSameNamedClrMember_WinsForNullableReceiver`.
- **`!!` edits, re-derived rather than trusted.** Reverted on `bench/concurrency/gsharp/Bench.gs` — `Environment.Version.ToString()` is the carve-out's canonical case and compiles unchanged. Kept on `samples/NullConditionalIndexing.gs` (`d` is an explicitly-nullable local, a direct receiver) and applied to the two `Issue710NullConditionalIndexingEmitTests` setup lines of that identical shape. `Issue3034`'s receivers are direct locals (`var value StringBuilder? = nil`) and stay `GS0159`; `Issue2229`'s are tuple elements and keep their `!!`.
- **The self-migration guard surfaced a pre-existing cs2gs gap, fixed alongside (commit 2).** The guard's first run on this branch went 8/8 to 5/8 with three `GS0159` on `node.FirstAncestorOrSelf<T>()` in the migrated `Cs2Gs.Translator` (`Cs2Gs.ProjectLoading` and `Cs2Gs.Pipeline` cascaded the identical three because they reference it; only `GS0159` appears in any artifact). This is NOT a binder gap: `FirstAncestorOrSelf` is a public *instance* method on `SyntaxNode`, not an extension, and the receivers are direct explicitly-`?` parameters — exactly the shape #4287 exists to reject, so the diagnostic is correct and the carve-out rightly declines them. The defect is that cs2gs's branch for a generic instance call with explicit type arguments (`x.M<T>(...)`) built its target with a bare `TranslateExpression`, while every other receiver position — member-access read, extension helper, method groups, element access, foreach — calls `TranslateReceiverWithNullForgiveness`. A *non*-generic call falls through to the general path and does forgive, so for one promoted receiver `x.Parent` and `x.Plain()` were asserted while `x.Up<T>()` beside them was left bare. Harmless while gsc bound nilable instance-call receivers unguarded; an error once #4287 closed that. Exactly the hazard the guard's own header describes — the C# compiles, the migrated G# does not — not a compiler regression. Fix is the one-line reuse of the helper the sibling branches already call.
- **#4287 was only half fixed until the private `cs2gs-oahu` corpus caught it (commit 5).** The nil check was gated on having syntax to quote the receiver from *in the diagnostic message* — and a chained call has none, because the accessor walker does not thread `receiverSyntax` through an intermediate step and the bound receiver nodes carry a null `Syntax`. So the whole check was skipped and the unsafe call bound through. Measured before the repair: `s.ToUpper()` → GS0159 (correct), but `s.ToUpper().Trim()`, `s.ToUpper().Length`, `s.ToUpper()[0]` and `sb.Append("a").Append("b")` → **no diagnostic at all**, while `(s.ToUpper()).Trim()` reported — parenthesising "fixed" it. The shape the issue was filed about therefore still shipped whenever the call was chained. A message-formatting detail must never decide whether a safety check runs: the check is now unconditional and the message degrades instead, exactly as the user-defined-receiver fallback in the same method already does. Degrading to the bare `Cannot find function X.` would be actively misleading — it claims the member does not exist — and that is precisely the confusing second diagnostic Oahu reported, so this adds a receiver-less companion report carrying the same nil guidance. **Both Oahu diagnostics are this one root cause**: the first call in the chain was silently skipped, the second reported against the error receiver it left behind. **Expect corpora to report MORE sites after this, not fewer** — every one is a latent unattributed `NullReferenceException`.
- **A second cs2gs gap the Oahu corpus caught: a bare member reference is not a path gsc narrows (commit 7).** Oahu also failed on `X != null && X.Contains(i)` where `X` is a member of the enclosing type. cs2gs's stability classification is *right* — translating one class with all four member kinds shows it already asserts exactly the two gsc cannot narrow (mutable field, settable property) and leaves the two it can (readonly field, get-only property) bare. What was wrong is the assumption that the emitted FORM doesn't matter: gsc narrows a stable member path only through its **qualified** form. `this.Stable != nil && this.Stable.Contains(i)` compiles; the identical bare `Stable != nil && Stable.Contains(i)` does not — for a readonly field and a get-only property alike, and for a READ (`.Capacity`) exactly as for a call, because the smart-cast frame is keyed by an `AccessPath` an implicit-`this` reference never produces. The predicate now declines a bare reference to an instance field or property and keeps every other answer, including its stable/unstable distinction for the qualified form. This also reaches `ReceiverNeedsNullForgiveness`'s value-position callers, correctly — gsc won't narrow a bare member there either. The full `Cs2Gs.Tests` run then caught an `is`-pattern against a Roslyn enum constant in the first cut of this predicate (`Issue4167SelfHostedEnumPatternRegressionTests`): per #4167/#4155/#4153 such a pattern silently mismatches when cs2gs translates its own source, so the `Kind` tests are written as `==` comparisons instead — and the explanatory comment avoids the rejected form too, since that guard scans the translated TEXT and comments are translated with it.
- Regenerates `test/Core.Tests/Baselines/refactoring-baseline.json` — exactly one hash, `samples/NullConditionalIndexing.gs`, whose emitted IL moved because of the retained `!!`.

## Verification

- `dotnet build GSharp.sln -c Release -graph`: Build succeeded, 0 errors (warnings are errors in this build).
- **Baseline red list on the previous commit `b7856060`** (re-derived, not taken from the review): `Core.Tests` 2 failed (`Issue2459NestedImportedClrMemberTests.NestedImportedFieldAndProperty_ResolveLikeMethod`, `Adr0177CatchClauseParityTests.DeclinedException_KeepsItsOriginalThrowSite`), `Compiler.Tests` 5 failed (`Issue2891`, `Issue2734`, `Issue2672`, `Issue710` x2), `Interpreter.Tests` 16 failed (`Issue3099TypeArgumentReificationTests`, all parameterized cases). Every one of them except `Issue710`'s two is now green BY CONSTRUCTION — no per-site `!!` was added to any of them.
- `Issue710`'s two sites are a genuinely different case and are the only ones given `!!`: `var d Dictionary[string, int32]? = Dictionary[string, int32]()` then `d.Add(...)` is a DIRECT receiver on an explicitly-nullable local — the same shape as `samples/NullConditionalIndexing.gs`, i.e. the bug being fixed, not a regression. (Note: these two were never touched by this PR's first commit, contrary to the review's categorization.)
- `bench/concurrency/gsharp/Bench.gs` compiled through `gsc` directly after the `!!` revert: exit 0, no diagnostics.
- **Bare-member witness (ADR-0154).** Removing the early return turns `BareMemberGuardedInSameCondition_IsAsserted_WhileTheQualifiedFormIsNot` red (1/3) and restores to 3/3. The test pins both halves: all four bare members asserted, AND the `this.`-qualified form of the same readonly field still bare — which is what distinguishes "decline bare references" from "give up on member narrowing".
- **Chained-call witness (ADR-0154).** Restoring the syntax gate on the condition turns `ChainedCall_NullableReceiver_StillReportsMayBeNil` and `ChainedIntoRead_NullableReceiver_StillReportsMayBeNil` red (2/22), while `ChainedCall_ThroughObliviousClrResult_CompilesCleanly` stays green — a carved-out oblivious call result never entered the block either way, so the carve-out is provably untouched by the repair. Both chained tests assert the "may be nil" wording, so they also pin the new receiver-less report. Targeted after: `Core.Tests` 38/38 (including `RefactoringBaselineTests` — no sample's IL moves), `Compiler.Tests` 93/93, `Interpreter.Tests` 20/20.
- **cs2gs witness (ADR-0154), a minimal oblivious project, eight shapes.** Pre-fix it fails to compile with the same three `GS0159`; post-fix it is green through translate/compile/ilverify/test-parity, and the diff of the emitted G# is **exactly three `!!` insertions and nothing else**. Asserted after the fix: read-first-use (`read!!.Parent`, unchanged), read-then-generic-call (`both!!.Parent` + `both!!.Up[T]()`, the call was bare), generic-call-first-use (`use!!.Up[T]()`, was bare). Correctly still bare: after a real `if x == nil` guard (G# narrows parameters too), a never-null-tainted parameter (never promoted), a nullable-enabled file, and a `this` receiver. Pinned by `tools/cs2gs/Cs2Gs.Tests/Issue4287GenericInstanceCallReceiverForgivenessTests.cs`. Note the two layers: that test exercises the TRANSLATOR alone, which is liberal by design and emits `guarded!!`; `NullAssertionPolishPass` runs only in CompileStage and strips every assertion gsc reports as unnecessary (GS0536 — 832 on one hot-core run), so through the full pipeline the guarded receiver comes back bare. Both outputs were captured from the same snippet to confirm it. That `guarded!!` doubles as a second discrimination witness: pre-fix the branch asserted nothing at all.
- **Self-migration PR guard (`build/run-cs2gs-selfmig-pr-guard.sh`, fresh root): `8/8 apps green; run PASSED.` / `PR guard PASSED.`** Final run is on the complete four-fix tree (fresh root `-r5`). The FIRST run on this branch was 5/8 — one root cause, three `GS0159` in `Cs2Gs.Translator` cascading to `Cs2Gs.ProjectLoading` and `Cs2Gs.Pipeline` — fixed in commit 2; a later run was 8/8 but on a tree that predated the chained-call and bare-member fixes, so it was discarded and re-run. **Assertion churn measured, not assumed**, over the whole 8-app hot core: raw emitted `!!` 5513 → 5516 (**+3**), of which `NullAssertionPolishPass` stripped 4196 → 4197 (**+1**) — a net **+2** surviving assertions corpus-wide, in 1 polish round against a cap of 12, `GS0536 still reported: 0`. The two fixes that could have churned it (the generic-instance-call receiver and the bare-member predicate) cost two assertions between them, because the corpus's guarded receivers are overwhelmingly locals (which gsc does narrow) or already-asserted unstable members.
- Full unfiltered suites on the final tree: `Core.Tests` **9523/9523**, `Cs2Gs.Tests` **3239/3239**, and (pre-chained-fix, unaffected by it) `Core.Tests` **9520/9520**, `Interpreter.Tests` **1524/1524**, `Compiler.Tests` **6006/6006** (includes `LanguageConformance`), `Cs2Gs.Tests` **3237/3238** — the one failure was an over-specified assertion in this PR's own new test (wrong layer: translator-only vs. post-polish), corrected in commit 4 and re-run **2/2**. No pre-existing or unrelated failure in any suite.
- **Witnesses of discrimination (ADR-0154)**, each an applied product mutant, built and run:
  - Drop `&& !CanBindClrInstanceMember(receiver)` from `BindAccessorCall` → `ObliviousClrProperty_ChainedCall_CompilesCleanly` and `ClrCallResult_ChainedCall_CompilesCleanly` fail (2/19); restore → 19/19.
  - Shrink `IsImportedClrChainReceiver` to `BoundClrPropertyAccessExpression` only → `ClrCallResult_ChainedCall_CompilesCleanly` and `ClrCallResult_ChainedRead_CompilesCleanly` fail (2/19). The `_ChainedRead_` failure is the witness for the read-path widening, which is new behavior with no pre-fix commit to be red on.
  - Drop the `|| underlyingType.ClrType is { IsInterface: true }` arm → `ImportedInterfaceReceiver_ObjectMember_ReportsMayBeNil` fails (1/1).
  - The pre-existing `FiledRepro_NullableStringReceiver_ReportsMayBeNil` (the `this.name.ToUpper()` repro, carrying its own witness from this PR's first commit) is green throughout all three mutants and after: the carve-out does not reach it.

## Residual, filed here rather than dropped

- The carve-out predicate is a metadata-ORIGIN proxy, not an oblivious-vs-annotated test, so a genuinely `[Nullable(2)]`-annotated BCL result chains unguarded too: `Type.GetMethod(n).ToString()` and `field.GetRawConstantValue().ToString()` compile, exactly as `DirectoryInfo.Parent.Name` always has on the read path. Pre-existing ceiling, matched deliberately; the approved decision names `GetRawConstantValue()` as in scope for the carve-out.
- **gsc narrows `this.X` but not the bare `X` for the identical stable path** — a readonly field or get-only property, for reads as much as calls. Real, pre-existing, independent of #4287 (before this PR the call was simply never checked, so it was invisible). Two attempts to close it in the binder did not land; it wants its own change rather than a rider here. Repros: the `k.gs`/`l.gs` pair described above.
- `TryGetUserInstanceMemberReceiverType` (`ExpressionBinder.Access.MemberLookup.cs`) still hard-codes the old property-only form of this rule for user-class-typed receivers reached through a CLR property. Out of scope here; named so the next drift bug has a pointer.

### A note on how item 2 was located

Its layer was redirected twice on inference and corrected each time by a falsifying probe *before* any code shipped. First, `Sel != nil && Sel.Count > 0` compiling while `Sel.Contains(i)` did not looked like a read/call asymmetry — but `Count` resolves to the LINQ *extension*, which is legitimately callable on a nilable receiver; an extension-free member (`Capacity`) showed the read fails identically. Second, the remaining failure looked like implicit-`this` narrowing needing to be applied at the use site — but a probe showed the `this.`-qualified form already works end to end, so the gap is the bare form, not the call path. The four-member cs2gs translation matrix then settled it. Both dead ends were reverted; the tree carries only the fix the evidence demanded.

## Checklist

- [x] Behavioral tests carry a witness of discrimination (ADR-0154): three applied product mutants, each listed above with the exact tests that went red and the restored green count.
- [x] Self-review verdict recorded: MERGEABLE, with the two residuals above filed in this body rather than dropped.


DavidObando:
## Closing in favor of a design pivot

This PR grew through many rounds chasing an ever-widening set of narrowing/carve-out edge cases: read-vs-call carve-out consistency, a cs2gs generic-instance-call receiver gap, a chained-call diagnostic-swallowing bug in the original #4287 check, a cs2gs bare-member-narrowing gap, and a silent-miscompile bug in the chained-call fix's extension probe (`List<T>.Reverse` vs `Enumerable.Reverse`), each discovered by CI or adversarial review after the previous round "looked ready."

That pattern — a system that must be statically proven correct for nullable-oblivious CLR interop, discovering a new corner case every time the surface area (chains, extensions, fields vs. properties vs. locals, overload resolution) is probed a different way — is a signal the underlying design is undersized for the problem, not that this PR needs one more round.

We're pivoting to adopt Kotlin's platform-type model for nullability-oblivious CLR/imported members: such values become a distinct type (`T!`, "platform type") assignable to both `T` and `T?` without a static check, with nullability enforced at the point of use via a runtime assertion rather than proven at compile time. This trades a weaker static guarantee for eliminating the entire class of narrowing/carve-out consistency bugs this PR kept finding.

A new ADR is being drafted to specify the change precisely, informed by every failure mode this PR surfaced (they're valuable case studies for what the new design must handle correctly). It will go through its own review before implementation begins. Issue #4287 and the follow-up issues filed during this PR (#4310, and the tracking issues from the earlier ADR-0184 work) will be revisited once the ADR lands, since the platform-type model changes what "fixed" means for this whole class of bug.

Not merging. Branch and commit history preserved on `fix/4287-nullable-clr-instance-call-receiver` for reference.

PR #4326, opened 2026-09-19
Native @GeneratedRegex support (ADR-0187)

### LINKED ISSUE #4301: cs2gs: `__generatedRegex_{method}` backing-cache field name for `[GeneratedRegex]` methods
`CSharpToGSharpTranslator.Members.cs:412-414` synthesizes a `__generatedRegex_{emittedMethodName}` field name to hold the compiled `Regex` instance backing a `[GeneratedRegex]` partial method. Unlike most of the other live families, this one IS source-derived (it's built from the regex-generating method's own translated name, `EmittedName(symbol, node.Identifier.ValueText)`, with collisions against `symbol.ContainingType`'s other members resolved by appending `_`, matching this repo's established collision convention) — only the `__generatedRegex_` PREFIX itself is synthetic, needed because C#'s own `[GeneratedRegex]` source generator creates an unnamed, compiler-only backing cache field with no C# source name to inherit at all.

Confirmed on the CI run 35287979111 migrated tree: 7 code, 24 raw (`src/Core`'s own `[GeneratedRegex]` usages, self-migrated).

## Fix hypothesis

This is the highest-occurrence active family after `__q`. Since the underlying problem (a compiler-synthesized backing field needs *some* name, and the source has none) is structurally similar to `__anon`/`__decon`, a real fix likely needs a language-level answer rather than a translator tweak — e.g. G# supporting a private, source-generator-style backing member convention that doesn't need a user-legible identifier at all (closer to how the field is genuinely invisible/compiler-only in the original C#). Not designed here.

Related: #3501 (definition of done: zero `__`-prefixed synthetic identifiers); #3086 (prior `[GeneratedRegex]` translation work, closed).

DavidObando:
## Summary

- Adds `@GeneratedRegex` as a third attribute discriminator on a bodyless `func` declaration, alongside `@DllImport`/`@LibraryImport` (ADR-0086 §1 / ADR-0092) — the consumer ADR-0092 §4 named as the trigger for revisiting general source-generator support, delivered here **without** a general pipeline: no grammar change (the `;`-body grammar is already attribute-agnostic), one binder helper (`GeneratedRegexBinder`), one emit method (`TypeDefEmitter.EmitStructGeneratedRegexBackingFields`), and small additions to field-row planning and `.cctor` assembly.
- Emission borrows ADR-0051's auto-property backing-field shape instead of ADR-0092's two-method P/Invoke shape: a private `static` `Regex` field, initialized once in the type's `.cctor`, with the annotated function's body reduced to `return <field>`. The field is `static` even behind an **instance** method — real `[GeneratedRegex]` semantics have no per-instance state — verified end-to-end by a cross-instance reference-equality emit test.
- New diagnostics `GS0593`–`GS0596`, following the `GS0360`/`GS9306` "one attribute, many shape rejections" convention.
- Full design rationale, alternatives considered (full partial-method support, a general source-generator pipeline, lazy backing-field init, raw-IL `.cctor` injection — all rejected with reasons), and verification notes: `docs/adr/0187-generated-regex-source-generator.md`.

Fixes #4301 (native-support half; the cs2gs consumer rewrite is a follow-up PR based on this branch).

## Test plan

- [x] `Issue4301GeneratedRegexBinderTests` (15 cases): well-formed static/instance shapes, GS0593 (missing/non-constant pattern), GS0594 (shape rejections), GS0595 (culture-sensitive `IgnoreCase`, explicit and inline `(?i)`, plus the negative control), GS0596 (invalid `RegexOptions` combination, invalid timeout), `-1` timeout binds clean, named-argument binding, and two regressions pinning the pre-existing GS0325/GS0388 fallbacks are unaffected.
- [x] `Issue4301GeneratedRegexEmitTests` (6 cases, real compile-and-run programs verified under `ilverify`): static and **instance** cross-instance cached-identity, explicit options/timeout round-trip, `-1` value-equal to `Regex.InfiniteMatchTimeout`, the `AppContext` process-wide default timeout, and the emitted backing field's metadata shape.
- [x] `DiagnosticIdUniquenessTests` (both diagnostics-reference mirrors updated for GS0593–GS0596).
- [x] Regression: 170 `Core.Tests` (PInvoke/LibraryImport/auto-property/static-field families) + 83 `Compiler.Tests` (static-ctor/auto-property/LibraryImport emit families) all pass unchanged.


DavidObando:
Closing in favor of a real implementation using SDK source generators.

PR #4327, opened 2026-09-19
cs2gs: emit native @GeneratedRegex instead of __generatedRegex_ (ADR-0187)

DavidObando:
## Summary

- Stacked on #4326 (the gsc language feature this depends on).
- Rewrites `TryTranslateGeneratedRegex` to emit the native `@GeneratedRegex(pattern, options, ...)` discriminator on a bodyless `func` directly, instead of synthesizing a `__generatedRegex_{name}` cache field plus a manual forwarding method.
- The `pattern` argument is always the **resolved constant string value**, never an identifier reference to a sibling `const` field — verified empirically that gsc's attribute-argument binder does not resolve a bare name reference in attribute-argument position. Not a fidelity loss: the resolved Roslyn constant was already what this function read; only the *readable cache-field name* it used to synthesize from the member's name no longer exists to name.
- Retires the `__generatedRegex_` family — 7 occurrences per ADR-0185's own count, the second-largest active synthetic-identifier family after `__q` — toward issue #3501's zero-synthetic-identifiers goal.

Part of #4301.

## Test plan

- [x] `Issue3086GeneratedRegexPipelineTests.PartialRecord_GeneratedRegex_TranslatesCompilesVerifiesAndRuns` — the fixture (static `shared`-block case, an **instance**-method case, default/explicit/infinite timeout, `CultureInvariant` inline-`(?i)`) re-translates to the native form, compiles, passes `ilverify`, and matches `baseline.stdout.golden` end to end — including the fixture's own cross-instance `Regex` reference-equality assertions.
- [x] `Issue3086GeneratedRegexPipelineTests.InlineIgnoreCaseWithoutInvariant_ReportsUnsupported` — the culture-sensitive-`IgnoreCase` migration-time rejection is unchanged.
- [x] Confirmed `__generatedRegex_` no longer appears anywhere in the re-translated output.


DavidObando:
Closing in favor of a real implementation using SDK source generators.

